Issue #593, opened 2015-02-03
monadtall two window normalize


zordsdavini:
for monadtall two windows doesn't work lazy.layout.normalize()

test:
open two window, maximize one and try to normalize - should go back to normal state (this doesn't work)

zordsdavini:
two window normalize can be done with cmd_reset

PR #602, opened 2015-02-10
for two windows doesn't work normalize

zordsdavini:
https://github.com/qtile/qtile/issues/593


tych0:
I don't think that's what normalize does; I think it's behavior is correct. It sounds like what you want is really a cmd_reset_ratio or something like that instead. If you switch this patch to do that, I'll be happy to merge it.

zordsdavini:
I can add cmd_reset_ratio as other patch and it could reset all maximizations, but this path is only for two windows as normalization for them doesn't work

zordsdavini:
There is cmd_reset already what reset all layout. For me it is enough, because secondary pane normalize is not so needed. I close this as it is useless

Issue #594, opened 2015-02-03
monadtall maximize should give focus to maximized window


zordsdavini:
This is enhancement.

situation: two open windows (for more window needs more discussion)

with lazy.layout.maximize() you can maximize window, with redo this command it maximize other window and here it would be better to have focus for it

zordsdavini:
I created it in local config and added to qtile-examples
https://github.com/qtile/qtile-examples/blob/master/zordsdavini/qtile_config.py

PR #603, opened 2015-02-10
follow maximizing window

zordsdavini:
Added option "follow_max": False. If it is set to True - if there are only two windows then cmd_maximized will follow maximized window
https://github.com/qtile/qtile/issues/594

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -13,6 +13,7 @@
 # Copyright (c) 2014 dmpayton
 # Copyright (c) 2014 dequis
 # Copyright (c) 2014 Florian Scherf
+# Copyright (c) 2015 Arnas Udovičius
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


tych0:
I'm confused; isn't there logic to do this directly above (i.e. the if len(self.clients) < 3 bit)? It seems as though that is broken?

zordsdavini:
if I add all logic to < 3 then it would be triggered and for == 1. Sure it will not crash qtile, because cmd_next will give same client. I rewrited to separate logic to other function and with monadtall flow (ratio, _med_ratio, focused)

zordsdavini:
buggy explaination about what it does https://www.youtube.com/watch?v=gJKbDxShyUU (follow focused window)

zordsdavini:
I found video about monadTall https://www.youtube.com/watch?v=NE6IQdwjzso
1st 8 seconds shows how maximization works and here sees how toggling leaves focus in small window. In this case focus should follow bigger window.

zordsdavini:
I did rebase for newer code

tych0:
Sorry, I'm confused. It looks like some other commits got in here twice?

tych0:
Ok, I watched the second video, and it seems to me the behavior in the video is correct. Why should it switch focus if all that is being toggled is the maximized flag?

You could implement this in your config with something like:

```python
def mymaximize(qtile):
    qtile.currentLayout.maximize
    if ...:
        qtile.currentLayout.focus_next()
```

And leave the default behavior the same in core, since it seems like quite a special case.

zordsdavini:
this following is additional help to make everything easier. Your config doesn't work for me. I'll try to find some more examples in qtile-examples. But I don't think it is my special case, this is maximization with additional functionality, what can be switched on by config

zordsdavini:
ok, I created it in local config, so I close it and sugest to save it in examples
https://github.com/qtile/qtile-examples/pull/26

Issue #561, opened 2014-12-11
uncontrolled window, qtile.log shouts


atommixz:
Sometimes the window becomes not controlled. It can not be done float / unfloat, you can not change the order. Also it is shows in all groups. Found .qtile.log, who shouts.
In the last qtile update, terminal - Sakura window is bugs all time.

Config almost standard http://ix.io/16m/
.qtile.log (cut) http://ix.io/1F9

zordsdavini:
problem still exist

zordsdavini:
Qtile doesn't crash totally it just becomes unresponsible. My qtile log shows same thing:
```
2015-02-10 16:15:03,018 ERROR _xpoll:668 Got an exception in poll loop
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/libqtile/manager.py", line 657, in _xpoll
    r = h(e)
  File "/usr/lib/python3.4/site-packages/libqtile/manager.py", line 1018, in handle_UnmapNotify
    self.unmanage(e.window)
  File "/usr/lib/python3.4/site-packages/libqtile/manager.py", line 482, in unmanage
    c.group.remove(c)
  File "/usr/lib/python3.4/site-packages/libqtile/group.py", line 215, in remove
    nextfocus = i.remove(win)
  File "/usr/lib/python3.4/site-packages/libqtile/layout/zoomy.py", line 61, in remove
    self.cmd_previous()
  File "/usr/lib/python3.4/site-packages/libqtile/layout/zoomy.py", line 138, in cmd_previous
    self.group.focus(client, False)
  File "/usr/lib/python3.4/site-packages/libqtile/group.py", line 161, in focus
    l.focus(win)
  File "/usr/lib/python3.4/site-packages/libqtile/layout/zoomy.py", line 118, in focus
    format=8
  File "/usr/lib/python3.4/site-packages/libqtile/xcbq.py", line 606, in set_property
    ).check()
  File "/usr/lib/python3.4/site-packages/xcffib/__init__.py", line 305, in check
    self.conn.request_check(self.sequence)
  File "/usr/lib/python3.4/site-packages/xcffib/__init__.py", line 517, in wrapper
    return f(*args)
  File "/usr/lib/python3.4/site-packages/xcffib/__init__.py", line 610, in request_check
    self._process_error(err)
  File "/usr/lib/python3.4/site-packages/xcffib/__init__.py", line 581, in _process_error
    raise error(buf)
xcffib.xproto.WindowError
[...]

PR #604, opened 2015-02-10
zoomy not needed part

zordsdavini:
This gives just error and functionality left working
It should fix https://github.com/qtile/qtile/issues/561


tych0:
If you're going to remove this, I think all the related functionality (e.g. configuration options, other property values, etc.) should be removed as well. But first, is it possible to fix the regression instead?

zordsdavini:
find better solution

PR #606, opened 2015-02-12
Set an eventloop exception handler to log them

dequis:
Instead of silently dropping them.

You wouldn't believe it took me four days to get to this patch.


dequis:
beep

PR #607, opened 2015-02-12
zoomy layout: don't set property if it is None

zordsdavini:
avoid setting property if it doesn't exist. In this case property_small for ZOOM property will not exist untill it will be set with property_big. In zoomy layout there creates big window and if new window appears it can be „iconified“.
This should fix https://github.com/qtile/qtile/issues/561

### DIFF
--- a/libqtile/layout/zoomy.py
+++ b/libqtile/layout/zoomy.py
@@ -138,11 +138,14 @@ class Zoomy(SingleWindow):
         return d
 
     def focus(self, win):
-        if self.focused and self.property_name:
+        if self.focused and self.property_name and self.focused.window.get_property(
+                self.property_name,
+                "UTF8_STRING"
+                ):
             self.focused.window.set_property(
                 self.property_name,
                 self.property_small,
-                "STRING",
+                "UTF8_STRING",
                 format=8
             )
         SingleWindow.focus(self, win)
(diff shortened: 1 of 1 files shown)


libqtile/layout/zoomy.py
```diff
@@ -138,11 +138,14 @@ def info(self):
         return d
 
     def focus(self, win):
-        if self.focused and self.property_name:
+        if self.focused and self.property_name and self.focused.window.get_property(
```
tych0: Can we make this `is None`, that way people can set it to "" if they want? Other than that, this patch looks good to me.

zordsdavini: Ok. Tomorrow in the morning I'll fix this
2015 Vas 12 19:39 "Tycho Andersen" <notifications@github.com> rašė:

zordsdavini:
compared to None as tych0 suggested.

tych0:
Sorry, I think I misread things. Doesn't this change make it so that the property is never set? Because it won't be set the first time, so the last part of the condition is never true.

zordsdavini:
1st time it will be set by property_big as it should. Because with qtile start there is no windows and one window can't be iconified - it will get property_big. Next window will become big and get property_big and then previous window will be iconified with property_small. So, it should be set.

zordsdavini:
confirm, what it still sets property_small:
http://tinypic.com/r/15qem3c/8
http://tinypic.com/r/27wuo1w/8

PR #608, opened 2015-02-13
TaskList: don't leave empty space if there's no icon

dequis:
Might be a slightly opinionated change, but tasklist items have variable space right now so why not.

Definitely not for 0.9.1

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -126,9 +126,12 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             self.layout.width = width
 
     def drawbox(self, offset, text, bordercolor, textcolor, rounded=False,
-                block=False, width=None):
+                block=False, width=None, icon=None):
         self.drawtext(text, textcolor, width)
-        padding_x = [self.padding_x + self.icon_size + 4, self.padding_x]
+
+        icon_padding = (self.icon_size + 4) if icon else 0
+        padding_x = [self.padding_x + icon_padding, self.padding_x]
+
         framed = self.layout.framed(
             self.borderwidth,
             bordercolor,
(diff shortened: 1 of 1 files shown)


Roger:
My idea for this was to add a default icon, but never did

tych0:
Seems like python 2.6 doesn't like this change?

dequis:
Uhhhh... how can i tell those errors apart from the usual "our test suite isn't deterministic"?

tych0:
On Fri, Feb 13, 2015 at 07:37:28AM -0800, dx wrote:
> Uhhhh... how can i tell those errors apart from the usual "our test suite isn't deterministic"?

I've never seen these before :)

dequis:
Re-ran the test and it passed. I also checked the patch for anything specific to python2.7, but the only thing that came close is the ternary operator was introduced in 2.5.

tych0:
Argh. We need to spend some time to figure that crap out. Looks reasonable to me, @Roger any other thoughts?

Roger:
I have no hope of me doing the default icon stuff, so this looks good :)

kynikos:
If you guys are interested in having a default icon, maybe optionally, I've pushed kynikos@f00e066878504744eaef13a8e18f1a647cde80be on top of this patch, although it should be expanded, but I think the biggest part would be done.

Roger:
@kynikos where did you get those icons, and what's the license?

kynikos:
Sorry, I've added it to the comment after saving it, so you haven't received in the email: the SVG file is from http://tango.freedesktop.org/Tango_Desktop_Project released under Public Domain; the PNGs are simply Inkscape exports that I've saved.

tych0:
Ok, so should I take this patch and then wait for the default icon one?

kynikos:
Given your answer I assume you would indeed merge a complete patch for a default icon, so as far as I'm concerned you can merge this patch and I'll finish mine and open another PR for it when ready.

tych0:
On Sun, Feb 15, 2015 at 06:39:02AM -0800, Dario Giovannetti wrote:
> Given your answer I assume you would indeed merge a complete patch for a default icon, so as far as I'm concerned you can merge this patch and I'll finish mine and open another PR for it when ready.

Ok. Can you please add a bit about the licensing to the header of the
svg? This will make it easier for us to package.

kynikos:
I certainly will, I've taken note of that. Don't expect anything soon though, there are other features I'd like to work on first :)

PR #620, opened 2015-02-25
Gentoo installation docs

buzz:
Proposing to reintroduce updated Gentoo install docs removed in e7aeaf71ba0ee442417fbe4de8d5883650dd8e36

### DIFF
--- a/docs/manual/install/gentoo.rst
+++ b/docs/manual/install/gentoo.rst
@@ -1,0 +1,46 @@
+==========================
+Installing on Gentoo Linux
+==========================
+
+The newest version of Qtile is not available in the official Gentoo
+portage tree. But you can use an unofficial `portage overlay`_ to
+install newer versions of Qtile.
+
+Installing using portage
+========================
+
+Currently `x11-wm/qtile-0.8.0`_ is available. Installing is as easy
+as:
+
+.. code-block:: bash
+
+    emerge qtile
+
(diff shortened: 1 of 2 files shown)


ramnes:
We have qtile-9999 on portage, so getting qtile-0.9.1 from a layman seems a bit useless to me. Though, it could be interesting to have qtile-0.9.1 on portage. Ping @radhermit.

fkmclane:
The qtile-9999 ebuild is actually out of date and, while I haven't tried it, probably won't work. It still depends on PyGTK, does not depend on an asyncio implementation for Python <3.4 (such as trollius which is also not in portage), and does not mark support for Python 3.

buzz:
I agree that having qtile-0.9.1 in portage would be optimal. Until that happens people can use the overlay. qtile-9999 didn't work for me. Anyway I'd prefer to use the latest stable release not the dev branch.

ramnes:
Well, I'm using this ebuild actually. But yeah, now I remember that I had to install one or two dependencies myself.

radhermit:
Or you could try this in about an hour or so, emerge --sync && emerge qtile ;)

For future reference, it can be helpful to attach ebuilds or patches to bugs.gentoo.org if you want to speed things up.

ramnes:
Awesome, thanks @radhermit!

tych0:
Going to close this in favor of the ebuild. Thanks!

PR #621, opened 2015-02-25
fix python3 issues for call of drawer.draw that wants only integer

jrazik:


tych0:
Aloha, looks like there is some noise in this commit about the linter errors I just fixed. If you can axe that I'm happy to merge. Thanks!

jrazik:
Hi, it seems that you have fixed the errors and commit it, thank you.

tych0:
On Wed, Feb 25, 2015 at 11:11:05PM -0800, jrazik wrote:
> Hi, it seems that you have fixed the errors and commit it, thank you.

Yep, thanks for the patch.

flacjacket:
It's a very minor point, but the `int()` call in `int(input_width // sp)` is redundant, the double slash divide `//` does integer division for both 2 and 3 (where "normal" division `/` is integer for 2 and floating point for 3)

ixjlyons:
The `int()` call may still be necessary depending on the type of `input_width` and `sp`:

``` python
>>> type(3//2)
<class 'int'>
>>> type(3//2.0)
<class 'float'>
```

jrazik:
Exactly, 
For `input_width // sp` in my case I got 18.0 and not 18

flacjacket:
Ah, right, nevermind

PR #623, opened 2015-02-28
Some fixes and improvements + fix ##619 for Max layout

### LINKED ISSUE #619: TreeTab and Zoomy layouts: changing a window to floating should also show the previous tiled window
EDIT: This bug was originally reported for the Max layout, but it was later discovered to affect other layouts too.

Steps to reproduce:
1. Open 2 windows in a group with Max layout.
2. Change the focused window to floating: the other window is not shown below it.

The expected behavior would be to also show the previous tiled window when one is changed to floating.

kynikos:
The commit messages and diffs should speak for themselves.
75be78388dddbaf7518c9f68848070cbbe1226ed fixes #619 for the Max layout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,6 @@ ckpatch: lint check
 clean:
 	-rm -rf dist qtile.egg-info docs/_build build/
 
-# strip off the leading 'v'
-VERSION=$(shell git describe --tags | cut -c 2-)
-
 # This is a little ugly: we want to be able to have users just run
 # 'python setup.py install' to install qtile, but we would also like to install
 # the man pages. I can't figure out a way to have the 'build' target invoke the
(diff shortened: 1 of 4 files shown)


kynikos:
There's something in the title or commit messages that wants to close #619 automatically if this PR is merged (very smart, GitHub... :P), what can we do?

tych0:
I suppose it's the string "addresses #619". Sorry I haven't taken a closer look at this yet, I've got a max focus bug that I want to look at as well, I suspect it's related to 30d88be2b30013668135ff5b2ced9b4ad9e9dbbd, so I'd like to try and understand all of these changes at the same time.

tych0:
Derp. Shouldn't have closed it.

tych0:
Looks good, thanks! This also fixed my focus bug that I was experiencing too :)

kynikos:
Fantastic! Can you reopen #619? Also TreeTab and Zoomy are affected.

PR #625, opened 2015-03-01
Emacs-style configuration helper for keys and buttons.

gvalkov:
Hello,

This PR adds convenience wrappers around `Key`, `Drag` and `Click` that make key and button bindings a bit more straighforward in my opinion. I hope the following example, adapted from `default_config.py`, serves well in demonstrating the change.

```python
from libqtile.config import EzKey as Key, EzDrag as Drag, EzClick as Click

keys = [
    # Switch between windows in current stack pane
    Key('M-k', lazy.layout.down()),
    Key('M-j', lazy.layout.up()),

    # Move windows up or down in current stack
    Key('M-C-k', lazy.layout.shuffle_down()),
    Key('M-C-j', lazy.layout.shuffle_up()),

    # Switch window focus to other pane(s) of stack
    Key('M-<space>', lazy.layout.next()),

    # Swap panes of split stack
    Key('M-S-<space>', lazy.layout.rotate()),

    # Toggle between different layouts as defined below
    Key('M-<Tab>', lazy.nextlayout()),
    Key('M-w',     lazy.window.kill()),
]

mouse = [
    Drag('M-1',  lazy.window.set_position_floating()),
    Drag('M-3',  lazy.window.set_size_floating()),
    Click('M-2', lazy.window.bring_to_front())
]
```

In other words:

```python
EzKey('M-A-z', ...)   == Key(['mod4', 'mod1'], 'z', ...)
EzKey('M-<Tab>', ...) == Key(['mod4'], 'Tab', ...)
```

Xmonad users will likely recognize the `EZConfig` name from the handy [XMonad.Util.EZConfig](http://xmonad.org/xmonad-docs/xmonad-contrib/XMonad-Util-EZConfig.html) module. On a side note, I've also implemented this for Awesome WM ([ezconfig.lua](https://github.com/gvalkov/awesome-ezconfig) if curious).

Please consider this as a request for comments more than a final PR. 

Kind regards,
Georgi

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,12 @@ class Mock(MagicMock):
     def __getattr__(cls, name):
             return Mock()
 
+    # xcbq does a dir() on objects and pull stuff out of them and tries to sort
+    # the result. MagicMock has a bunch of stuff that can't be sorted, so let's
+    # like about dir().
+    def __dir__(self):
+        return []
+
 MOCK_MODULES = [
     'cairocffi',
     'cffi',
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -11,7 +11,7 @@ Configuration lookup order
 
 Qtile looks in the following places for a configuration file, in order:
 
-* The location specified by the ``-f`` argument.
+* The location specified by the ``-c`` argument.
 * ``$XDG_CONFIG_HOME/qtile/config.py``, if it is set
 * ``~/.config/qtile/config.py``
 * It reads the module ``libqtile.resources.default_config``, included by
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -2,7 +2,48 @@
 Keys
 ====
 
-The ``keys`` variable defines Qtile's key bindings.
+The ``keys`` variable defines Qtile's key bindings. Individual key
+bindings are defined with ``libqtile.config.Key`` as demonstrated in
+the following example. Note that you may specify more than one
+callback functions.
+
+::
+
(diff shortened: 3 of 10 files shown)


tych0:
Hey, this looks great, thanks! Some thoughts:

1. Can you fix the linter error so that travis is happy?
2. Can you write some documentation about how to use it?
3. What happens if people want to bind to -? (if it's too hard to make this possible, don't worry about it; just looks like the parser will break in this case)

gvalkov:
Thanks for the comments, @tych0. I'll certainly fix 1 and 2 - I didn't want to commit to documentation before the idea was verified. As for `M--`, that is simply `M-<minus>` - there doesn't seem to be a  `xcbq.keysyms["-"]` :)

Alternatively, splitting with something like the following and mapping `-` to `<minus>` would do the job. However, I'm not too thrilled by the `M--` syntax, even though emacs's kbd macro supports it. You're right about the current parser breaking though.

```python
>>> re.split(r'(?<!-)-', 'M-A--')
['M', 'A', '-']
```

tych0:
I got it to work with, https://github.com/tych0/qtile/commit/9e931a59d3c203168c7f8c0c1f60182fd335bd67

With this, we only have to change the docs code, not the actual qtile code, which I think is preferable. Anyway, if you want to cherry pick that on top of your branch and drop the other commit, I'll be happy to merge.

gvalkov:
That's a neat fix. Cherry-picked + removed the `try` around sort.

ramnes:
Wow, just seen that PR. Really awesome, thanks @gvalkov!

Issue #573, opened 2015-01-11
Vertical bar


kynikos:
It would be nice if the bar could also be shown vertically at the left/right sides of the screeen.
This is very easy to implement for the bar itself, but of course the problem is that widgets should support this too.
I think however that not all widgets should necessarily support the vertical orientation, and this could be solved widget-by-widget by moving the test in [bar.py#L155](https://github.com/qtile/qtile/blob/develop/libqtile/bar.py#L155) to a new `bar._HWidget` class that inherits from the existing `bar._Widget` class, so that horizontal-only widgets could derive from it, including `bar._TextBox`. Of course a `bar._VWidget` class should be created too, for completeness' sake. As widgets are given the ability to adapt to the bar orientation, they can be made to inherit directly from `bar._Widget`.

tych0:
On Sat, Jan 10, 2015 at 08:40:29PM -0800, Dario Giovannetti wrote:
> It would be nice if the bar could also be shown vertically at the left/right sides of the screeen.
> This is very easy to implement for the bar itself, but of course the problem is that widgets should support this too.
> I think however that not all widgets should necessarily support the vertical orientation, and this could be solved widget-by-widget by moving the test in [bar.py#L155](https://github.com/qtile/qtile/blob/develop/libqtile/bar.py#L155) to a new `bar._HWidget` class that inherits from the existing `bar._Widget` class, so that horizontal-only widgets could derive from it, including `bar._TextBox`. Of course a `bar._VWidget` class should be created too, for completeness' sake. As widgets are given the ability to adapt to the bar orientation, they can be made to inherit directly from `bar._Widget`.

This is a nice idea (and perhaps even useful for people whose native
languages aren't RTL or LTR ones). Does pango support up/down
rendering of glyphs? If so, then it seems we'd want textbox to support
both horizontal and vertical rendering as well.

I'm not convinced that inheritance is the best design for this,
though, since some widgets would need both rendering methods, and to
switch between them at runtime. Perhaps just a set of constants, and
each widget somehow asserts when qtile starts that the bar orientation
requested is something that it supports?

kynikos:
[...]

PR #627, opened 2015-03-02
Gap: fix calculation of y property

kynikos:
While working on #573 I've accumulated these fixes that I'd like to stop rebasing every time I fetch :) They should be all trivial ones, except for 9f687657bae381c43dcb50f6e19a0b0a0cab5efc, which I'd like to discuss here, because I don't think it's tested, and by just looking at the code I can't understand why `y` was calculated differently whether the Gap was on the left or right side of the screen.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -64,10 +64,9 @@ class Gap(command.CommandObject):
             return screen.y
         elif screen.bottom is self:
             return screen.dy + screen.dheight
-        elif screen.left is self:
+        else:
+            # Screen corners are reserved to horizontal gaps, if present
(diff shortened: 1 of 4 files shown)


kynikos:
This is the relevan history of commits: f07d9dbb5231774d48cad0da1a926cedac9c2924 187d066d858170a8d889c0bb915762d9f2cadd3b cbe029cb2c9cdc6e627b458f5675d076d651df53

kynikos:
Update: added test to demonstrate the need for "Gap: fix calculation of y property".

tych0:
Cool, thanks for the test!

PR #628, opened 2015-03-06
Vertical bar

kynikos:
Yes, this is huge :) This patch enables support for vertical orientation to the bar (#573). It also enables vertical orientation for 4 widgets, as a start. `widget.base._TextBox` hasn't been updated yet, i.e. it still only supports horizontal orientation, but its incompatibility, and therefore that of its derived widgets, is properly documented and protected by the `ConfigError` exception, just like it worked before when trying to install a bar in vertical position.

Since the bar can now be either horizontal or vertical, I've had to change some uses of "width" and "height" to respectively "length" and "size", both in attributes and comments, as they are orientation-agnostic terms.

Then there's the "Bar: support vertical orientation; move test to widgets" commit: it adds an `orientation` positional parameter to `widget.base._Widget.__init__()`. All the widgets in the repo have been updated, but this might actually break existing widgets that are developed in external repos. I think for readability it's necessary to have the widgets explicitly define the supported bar orientation at instantiation, but if we instead want to keep backward compatibility, the `orientation` parameter can be made optional with the predefined value of `ORIENTATION_HORIZONTAL`, maybe temporarily, and change it to simple positional in the future? If we decide for the more abrupt change, as it is now, it may be wise to post a warning in the mailing list.

Finally, the Sep, Spacer and Image widgets used to support a `width` parameter for their constructor: consistently with the new ability to be oriented vertically, I've added a `length` parameter, but I've also decided to keep backward compatibility for `width`, warning about the deprecation in the docstrings. What do you think?


tych0:
Sweet! I'll take a closer look this weekend.

kynikos:
I think the adaptation of Systray (acd51e596fb3a35508bfaf4b6f8b399ed30243c4) is incomplete, I've noticed a bug when starting some applications with a tray icon in a vertical bar (restarting qtile fixes it). I think I've also understood why but I don't have the time to push the fix today I'm afraid.

libqtile/widget/base.py
```diff
@@ -29,30 +29,38 @@
...
 import six
 import subprocess
 import logging
 import threading
 import warnings
 
+ORIENTATION_NONE = 0
+ORIENTATION_HORIZONTAL = 1
+ORIENTATION_VERTICAL = 2
+ORIENTATION_BOTH = 3
```
tych0: Can we export these in libqtile.config, so that that's still the module everyone looks in to get configuration things? (We can either move them there or just re-export them, either way is fine with me.)

libqtile/widget/launchbar.py
```diff
@@ -210,10 +213,11 @@ def draw(self):
                 # display an icon
                 self.drawer.ctx.set_source(self.surfaces[name])
                 self.drawer.ctx.paint()
-            self.drawer.draw(self.offset + xoffset, icon_width + self.padding)
+            self.drawer.draw(offsetx=self.offset + xoffset,
+                             width=icon_width + self.padding)
             xoffset += icon_width + self.padding
 
-    def calculate_width(self):
+    def calculate_length(self):
```
tych0: This bit of the diff looks a little funny. I understand why you changed the name, but s/width/length is a bit weird. I don't really have any better suggestions, though, so maybe it's ok.

libqtile/widget/spacer.py
```diff
@@ -30,16 +30,28 @@
...
+    def __init__(self, length=bar.STRETCH, width=None):
         """
-            - width: Width of the widget.
-              Can be either ``bar.STRETCH`` or a width in pixels.
+            - length: Length of the widget.
+              Can be either ``bar.STRETCH`` or a length in pixels.
+            - width: DEPRECATED, same as ``length``.
         """
-        base._Widget.__init__(self, width)
+        # 'width' was replaced by 'length' since the widget can be installed in
```
tych0: Can you add a call to `base.deprecated` here to warn people in their logs?

libqtile/widget/base.py
```diff
@@ -29,30 +29,38 @@
...
+        The offsetx and offsety attributes are set by the Bar after all widgets
+        have been configured.
     """
-    offset = None
+    offsetx = None
+    offsety = None
     defaults = [("background", None, "Widget background color")]
 
-    def __init__(self, width, **config):
+    def __init__(self, length, orientation, **config):
```
tych0: Is it possible to use our configuration framework here for both length and orientation? That way they will automatically get picked up by the doc generation code.

libqtile/widget/base.py
kynikos: Actually I took inspiration from `bar.STRETCH`, `bar.CALCULATED` and `bar.STATIC`, which aren't in libqtile.config either. The `base.ORIENTATION_*` bits are only needed by widget developers to define what orientation their widget supports when they instantiate it, they aren't supposed to be configured by end users... or didn't I get what you meant?

libqtile/widget/launchbar.py
kynikos: The only alternative I see is adding a `calculate_height()` method and similar for the other s/width/length, and use the right one depending on the orientation of the bar, but IMHO this would complicate the code a lot, also adding many tests, only for the sake of avoiding the word "length". If you confirm you accept the new terms ("length" and "size") I'll expand the proper comments to better explain what they represent, also comparing them to the previous "width" and "height".

libqtile/widget/spacer.py
kynikos: Sure! I assume you mean in _addition_ to the comments etc., not in _alternative_, right?

libqtile/widget/base.py
kynikos: But do we want them in the docs? `orientation` is never supposed to be set by end users, and `length` should be settable only in some cases, like the Spacer widget, so I think the configuration framework should be used only in those particular cases. But again, I may be misunderstanding what you mean?
Regarding the configuration framework in general, the options that widgets inherit from their superclasses seem not to end up in the docs, see for example `Battery` and `_Battery`. I think I'll open a bug report for this, after investigating it a little more.

libqtile/widget/base.py
tych0: Ah, ok, I misunderstood that. You've got it the way it should be, then.

libqtile/widget/spacer.py
tych0: Yep, in addition to.

libqtile/widget/base.py
tych0: In that case, can we move orientation to a class property instead of an instance property? That way it won't be in the docs and also won't be confusing about whether or not it is really settable.

r.e. your bug report, thanks for pointing that out, I'll try and look into it. The intent is that they do show up, at least.

libqtile/widget/launchbar.py
tych0: Oh, I was thinking of keeping the code the same, just renaming the method so that it's "obvious" that it's not a purely vertical or horizontal property (indeed, aren't you already just using this method and doing the right thing based on the widget's configured orientation?). The only name I could come up with is `calculate_size()`, and I'm not sure that's really better than this.

kynikos: Actually I've already used "size" to represent the other (minor) dimension of the bar, which previously was also referred to as "height". I think there's not much more we can do than this, I'll make sure to expand the comments describing the two terms.

libqtile/widget/base.py
kynikos: Of course, silly me, orientation has to be a class property! I'll get to that and the rest in the coming weekend.

kynikos:
Here we are, I've amended the patches as discussed. Only the last 6 commits (in sequential order, not chronological as unfortunately GitHub shows) are changed and need review: if you want to see the difference, the previous PR is saved in my [develop#573-PR1](https://github.com/kynikos/qtile/tree/develop%23573-PR1) branch. If you want, you can use 40578ef02020a0638a10536b10bd791f66eae3f1 to amend 0b33094121a9e6623a849b4d844c110d50802cac.
Also, I've removed the commit with the Systray widget's adaptation because it still needs refinement: for the moment the functionality of the vertical bar can be demonstrated only with the Sep, Spacer and Image widgets.

kynikos:
I've just added 4 commits to document the bar orientation supported by each widget automatically, based on the set `orientations` attribute of its class.
I haven't amended any previous commit this time, but the previous PR is saved in my [develop#573-PR2](https://github.com/kynikos/qtile/tree/develop%23573-PR2) branch anyway.
I stopped reordering the commits so you don't have to force-fetch and don't have to review everything from the start, but before you do the actual merge I can reorder the commits so that they follow a more logical sequence.

tych0:
Thanks, and sorry for the delay. I will try to review this in the next few days.

kynikos:
Thank you, I've rebased everything on the new merged commits, and also added another amendment for the recently added `KeyboardKbdd` widget. There weren't any conflicts, so you don't have to review what you had already reviewed.
I know this is a lot of code to review, if you see you're wasting too much time trying to understand something, it means it's not clear or not commented enough, so please ask here and I'll be glad to improve it!

kynikos:
Tests fail because of #634, which is fixed in #635.

docs/sphinx_qtile.py
```diff
@@ -38,6 +38,9 @@
 qtile_class_template = Template('''
 .. autoclass:: {{ module }}.{{ class_name }}
     :members: __init__
+    {% if is_widget %}
+    Supported bar orientations: {{ obj.orientations }}
```
tych0: Does this just render the int (i.e. 0, 1, 2, 3?) or does it actually render HORIZONTAL or whatever? (Not that it's blocking merging this, but we should probably fix it at some point if it's just the int.)

libqtile/bar.py
```diff
@@ -151,14 +157,20 @@ def __init__(self, widgets, size, **config):
...
-        if self not in [screen.top, screen.bottom]:
-            raise confreader.ConfigError(
-                "Bars must be at the top or the bottom of the screen."
-            )
-        if len([w for w in self.widgets if w.width_type == STRETCH]) > 1:
+        Gap._configure(self, qtile, screen)
+
+        stretches = 0
+        for w in self.widgets:
+            # Executing _test_orientation_compatibility later, for example in
```
tych0: Can we do this in __init__? _configure is called every time a screen is resized, and I don't think we need to do it more than once.

tych0: (Also not blocking merging, I can make the change later, just curious if you've thought about it.)

tych0:
Hi, this looks pretty good to me right now. I'll probably merge it as a branch so we can revert it wholesale if it starts causing problems. If you like the commit history now, I can merge it. Otherwise, feel free to rearrange things so they make the most sense, and I'll merge it when you give me the go ahead.

docs/sphinx_qtile.py
kynikos: It prints the strings set at https://github.com/kynikos/qtile/blob/develop%23573/libqtile/widget/base.py#L72 thanks to https://github.com/kynikos/qtile/blob/develop%23573/libqtile/widget/base.py#L59
I took inspiration from https://github.com/kynikos/qtile/blob/develop%23573/libqtile/bar.py#L120 but if you think it's too obscure I can change it to something else, maybe using a dictionary to translate the integers? Or I could expand the comments too.

libqtile/bar.py
kynikos: I'm not sure what you're referring to exactly :) If you're talking about the horizontal/vertical test, if I understand correctly, in `__init__` the bar (and the gap) doesn't know yet where it's been positioned, right? That's also why the previous test was done there too.

kynikos:
Great! Yes, I'll rearrange the commits a bit before merging, but I'll save the current branch into another one so you'll be able to confirm that the result is the same as the one you reviewed :) I'll do that in the next hours and notify you here when ready!

kynikos:
(Not that I'm taking hours to rearrange, it's just that I don't have time right now :D )

kynikos:
Here we are, I've also added the `orientations` attribute to the recently merged Cmus widget.
You can find the PR you reviewed at https://github.com/kynikos/qtile/tree/develop%23573-PR3
The rearranged branch is at https://github.com/kynikos/qtile/tree/develop%23573-PR4
The current branch is the rearranged one, but also rebased on the current develop branch, because I've wanted to make sure that there were no conflicts or other things to adapt.

tych0:
Awesome, it's in! Now we'll see what it breaks :). Thanks for doing the work!

kynikos:
I just hope you're not running qtile as root :P
I'll wait a bit before pushing vertical adaptations for other widgets, so we make sure that the horizontal behavior is unaffected by this patch.

PR #629, opened 2015-03-17
Added a widget to change keyboard layouts per window, using "kbdd".

tuxitop:


libqtile/widget/keyboardkbdd.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+from . import base
+from dbus.mainloop.glib import DBusGMainLoop
+import re
+import subprocess
+import dbus
+
+
+class KeyboardKbdd(base.InLoopPollText):
```
tych0: You don't really need the polling here, you could just use a `base.Textbox` and update `self.text` instead of `self.keyboard`. The polling may give you some logic about when to refresh that you'd have to implement, though, not sure.

libqtile/widget/keyboardkbdd.py
```diff
@@ -0,0 +1,90 @@
...
+        if re.search("kbdd", stdout):
+            return True
+        self.log.error('Please, check that kbdd is running')
+        return False
+
+    def _dbus_init(self):
+        dbus_loop = DBusGMainLoop()
+        bus = dbus.SessionBus(mainloop=dbus_loop)
+        bus.add_signal_receiver(self._layout_changed,
+                                dbus_interface='ru.gentoo.kbdd',
```
tych0: Is this always the dbus interface it's exposed on, or is it just for gentoo? We should make it configurable if it varies across distros.

tych0:
Hi,

Thanks for the patch. A few thoughts (not sure if you want to fiddle with the first one), and also it seems that there is a linter error. Other than that it looks good, thanks!

tuxitop:
Hi
about the polling, I tried self.TextBox at first, but for some reason, it
doesn't work quit right. for example when you are typing in a text box and
switch your layout, you need to switch to another group or another window,
and then get back to the first window for the widget to update itself. I'm
not sure why this happens, but it happens. I might be doning something
wrong though. that's probably why the original KeyboardLayout widget also
uses polling. I'll give it another shot and update it if I could get it
working with self.TextBox.
about gentoo, It's alright. the address is the same in all distros.
actually I'm using Arch. (and you can also check it in kbdd website.)

libqtile/widget/keyboardkbdd.py
tuxitop: I checked again and I still think polling gives better functionality. It acts a little weird when using base._TextBox.

libqtile/widget/keyboardkbdd.py
tuxitop: It's always gentoo. :)

tuxitop:
Hi,
I fixed the linting problem. everything else seems fine. at least for now, polling seems to be the best way.

tych0:
Awesome, thanks for the code!

PR #632, opened 2015-03-21
add wallpaper widget using feh

muhammedabuali:
a widget to change the wallpaper automatically using feh you can either set an image file or an image folder


libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                subprocess.call([
+                    'feh',
+                    '--bg-fill',
+                    self.wallpaper
+                ])
+            return
+        cur_index = self.index % len(self.images)
+        cur_image = self.images[cur_index]
+        self.text = self.images[cur_index].split("/")[-1]
```
tych0: Here I think you can use os.path.basename, which will account for a few other oddities in paths.

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+        ])
+
+    def button_press(self, x, y, button):
+        if button == 1:
+            self.index += 1
+            self.set_wallpaper()
+            self.draw()
+
+    def draw(self):
+        base._TextBox.draw(self)
```
tych0: Do you need this override at all? I don't think so, since you're just calling the superclass method.

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+                        os.listdir(self.directory)
+                           )
+                       )
+            )
+        except IOError as e:
+            print("I/O error({0}): {1}".format(e.errno, e.strerror))
+
+    def set_wallpaper(self):
+        if len(self.images) == 0:
+            if len(self.wallpaper) == 0:
```
tych0: I think it is more idiomatic to use `None` as the sentinel value here, rather than `""`.

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+                self.text = "empty"
+            else:
+                subprocess.call([
+                    'feh',
+                    '--bg-fill',
+                    self.wallpaper
+                ])
+            return
+        cur_index = self.index % len(self.images)
```
tych0: I think it might be better to make sure that the index is always in range, i.e. put this code where you handle the button click and actually increment the index.

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+    def get_path(self, file):
+        return self.directory + file
+
+    def get_wallpapers(self):
+        try:
+            # get path of all files in the directory
+            self.images = list(
+                filter(os.path.isfile,
+                       map(self.get_path,
+                        os.listdir(self.directory)
```
tych0: The single space formatting here is a little funky. I think pep8 says to line this up with the first argument to map, something like:

```
map(self.get_path,
    os.listdir(self.directory)),
```

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+        except IOError as e:
+            print("I/O error({0}): {1}".format(e.errno, e.strerror))
+
+    def set_wallpaper(self):
+        if len(self.images) == 0:
+            if len(self.wallpaper) == 0:
+                self.text = "empty"
+            else:
+                subprocess.call([
```
tych0: There is probably a way you could restructure this so that you avoid both feh calls, by setting self.wallpaper to the correct thing if self.images has values, and then just setting the background to self.wallpaper if it is not none.

tych0:
Hi, thanks for the code! I made a few comments, nothing major though.

libqtile/widget/wallpaper.py
muhammedabuali: but I am using the modulus. how is it possible to exceed the range

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def set_wallpaper(self):
+        if len(self.images) == 0:
+            if len(self.wallpaper) == 0:
+                self.text = "empty"
+            else:
+                self.images.append(self.wallpaper)
+        cur_index = self.index % len(self.images)
+        cur_image = self.images[cur_index]
+        self.text = os.path.basename
```
tych0: Does this work? I was thinking you'd do os.path.basename(self.images[cur_index]) or similar; I guess this just displays the repr() of the basename function.

libqtile/widget/wallpaper.py
tych0: On Sun, Mar 22, 2015 at 9:02 AM, muhammedabuali <notifications@github.com>
wrote:

> In libqtile/widget/wallpaper.py
> <https://github.com/qtile/qtile/pull/632#discussion_r26903870>:
>
> > +            )
> > +        except IOError as e:
> > +            print("I/O error({0}): {1}".format(e.errno, e.strerror))
> > +
> > +    def set_wallpaper(self):
> > +        if len(self.images) == 0:
> > +            if len(self.wallpaper) == 0:
> > +                self.text = "empty"
> > +            else:
> > +                subprocess.call([
> > +                    'feh',
> > +                    '--bg-fill',
> > +                    self.wallpaper
> > +                ])
> > +            return
> > +        cur_index = self.index % len(self.images)
>
> but I am using the modulus. how is it possible to exceed the range
>

It's not, for this usage. The problem is that if someone comes to write
some new code and uses the index, it might not be valid because you only
take the modulus right before use, instead of right after you add one to
the index.

muhammedabuali:
my bad, I pushed a fix for that

libqtile/widget/wallpaper.py
```diff
@@ -0,0 +1,79 @@
...
+        if len(self.images) == 0:
+            if len(self.wallpaper) == 0:
+                self.text = "empty"
+            else:
+                self.images.append(self.wallpaper)
+        cur_index = self.index % len(self.images)
+        cur_image = self.images[cur_index]
+        self.text = os.path.basename(cur_image)
+        subprocess.call([
+            'feh',
```
dequis: maybe a setting for this?

muhammedabuali:
@dequis I agree and it is done

tych0:
Hi,

It looks like the only thing left is moving the modulus to where you increment the index. If you do that, I'll be happy to merge this.

Issue #634, opened 2015-03-22
Test " test.test_manager.test_focus_stays_on_layout_switch" Failing


aniruddhkanojia:
I ran tests after fetching from the latest [commit](https://github.com/qtile/qtile/commit/44d93541551f8a30faab946608adb637e118c8ad) and found this test failing.

```python
======================================================================
ERROR: test.test_manager.test_focus_stays_on_layout_switch
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/samfisher/gsoc/qtile/test/utils.py", line 119, in wrapped_fun
    return function(self)
  File "/home/samfisher/gsoc/qtile/test/test_manager.py", line 733, in test_focus_stays_on_layout_switch
    xephyr.c.nextlayout()
  File "/home/samfisher/gsoc/qtile/libqtile/command.py", line 113, in __call__
    return self.call(self.selectors, self.name, *args, **kwargs)
  File "/home/samfisher/gsoc/qtile/libqtile/command.py", line 246, in call
    raise CommandError(val)
CommandError: No such command.
-------------------- >> begin captured stdout << ---------------------
------------------------ >> begin log file << ------------------------
2015-03-22 20:08:52,757 WARNING init_log:95 Starting Qtile
2015-03-22 20:08:52,826 INFO fire:270 Internal event: addgroup((<libqtile.manager.Qtile object at 0x7f5753b19390>, 'a'), {})
2015-03-22 20:08:52,827 INFO fire:270 Internal event: changegroup((), {})
2015-03-22 20:08:52,828 INFO fire:270 Internal event: addgroup((<libqtile.manager.Qtile object at 0x7f5753b19390>, 'b'), {})
2015-03-22 20:08:52,829 INFO fire:270 Internal event: changegroup((), {})
2015-03-22 20:08:52,830 INFO fire:270 Internal event: addgroup((<libqtile.manager.Qtile object at 0x7f5753b19390>, 'c'), {})
2015-03-22 20:08:52,831 INFO fire:270 Internal event: changegroup((), {})
2015-03-22 20:08:52,833 INFO fire:270 Internal event: addgroup((<libqtile.manager.Qtile object at 0x7f5753b19390>, 'd'), {})
2015-03-22 20:08:52,834 INFO fire:270 Internal event: changegroup((), {})
[...]

PR #635, opened 2015-03-22
Fixes #394 & #634

### LINKED ISSUE #394: layout.Tile border_focus gives NameError if modified
When setting `layout.Tile` with `border_focus` argument, i.e. with:
```python
layout.Tile(ratio=0.5, border_focus="ffffff") 
```
Qtile crashes and says:

```
  File "/usr/lib/python2.7/dist-packages/libqtile/manager.py", line 585, in _xpoll
    r = h(e)
  File "/usr/lib/python2.7/dist-packages/libqtile/manager.py", line 892, in handle_UnmapNotify
    self.unmanage(e.window)
  File "/usr/lib/python2.7/dist-packages/libqtile/manager.py", line 406, in unmanage
    c.group.remove(c)
  File "/usr/lib/python2.7/dist-packages/libqtile/group.py", line 215, in remove
    self.focus(nextfocus, True)
  File "/usr/lib/python2.7/dist-packages/libqtile/group.py", line 156, in focus
    l.focus(win)
  File "/usr/lib/python2.7/dist-packages/libqtile/layout/max.py", line 46, in focus
    self.group.layoutAll()
  File "/usr/lib/python2.7/dist-packages/libqtile/group.py", line 92, in layoutAll
    self.layout.layout(normal, screen)
  File "/usr/lib/python2.7/dist-packages/libqtile/layout/base.py", line 54, in layout
    self.configure(i, screen)
  File "/usr/lib/python2.7/dist-packages/libqtile/layout/tile.py", line 189, in configure
    bc = self.group.qtile.colorPixel(self.border_focus)
  File "/usr/lib/python2.7/dist-packages/libqtile/utils.py", line 99, in wrap
    ret = f(self, *args)
  File "/usr/lib/python2.7/dist-packages/libqtile/manager.py", line 364, in colorPixel
    return self.conn.screens[0].default_colormap.alloc_color(name).pixel
  File "/usr/lib/python2.7/dist-packages/libqtile/xcbq.py", line 288, in alloc_color
    self.cid, len(color), color
BadName: <xcb.xproto.NameError object at 0x319c908>
```

aniruddhkanojia:
This fixes #394 and #634 
Added changes as asked by @tych0 .
Also fixed the tests.

Please review.

### DIFF
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -37,6 +37,7 @@
 from __future__ import print_function, division
 
 import six
+import re
 
 from xcffib.xproto import CW, WindowClass, EventMask
 from xcffib.xfixes import SelectionEventMask
@@ -316,20 +317,20 @@ class Colormap:
         """
             Flexible color allocation.
         """
-        if color.startswith("#"):
-            if len(color) != 7:
+        try:
+            return self.conn.conn.core.AllocNamedColor(
+                self.cid, len(color), color
+            ).reply()
+        except xcffib.xproto.NameError:
+            if len(color) < 6:
                 raise ValueError("Invalid color: %s" % color)
 
             def x8to16(i):
                 return 0xffff * (i & 0xff) // 0xff
-            r = x8to16(int(color[1] + color[2], 16))
-            g = x8to16(int(color[3] + color[4], 16))
-            b = x8to16(int(color[5] + color[6], 16))
+            r = x8to16(int(color[-6] + color[-5], 16))
+            g = x8to16(int(color[-4] + color[-3], 16))
+            b = x8to16(int(color[-2] + color[-1], 16))
             return self.conn.conn.core.AllocColor(self.cid, r, g, b).reply()
-        else:
-            return self.conn.conn.core.AllocNamedColor(
-                self.cid, len(color), color
-            ).reply()
 
 
 class Xinerama:
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -37,7 +37,6 @@
 from __future__ import print_function, division
 
 import six
-import re
 
 from xcffib.xproto import CW, WindowClass, EventMask
 from xcffib.xfixes import SelectionEventMask
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -321,8 +321,6 @@ class Colormap:
                 self.cid, len(color), color
             ).reply()
         except xcffib.xproto.NameError:
-            if len(color) < 6:
-                raise ValueError("Invalid color: %s" % color)
 
             def x8to16(i):
                 return 0xffff * (i & 0xff) // 0xff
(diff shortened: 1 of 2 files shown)


libqtile/xcbq.py
```diff
@@ -37,6 +37,7 @@
 from __future__ import print_function, division
 
 import six
+import re
```
kynikos: This import isn't necessary anymore I guess

libqtile/xcbq.py
```diff
@@ -316,20 +317,20 @@ def alloc_color(self, color):
...
         """
-        if color.startswith("#"):
-            if len(color) != 7:
+        try:
+            return self.conn.conn.core.AllocNamedColor(
+                self.cid, len(color), color
+            ).reply()
+        except xcffib.xproto.NameError:
+            if len(color) < 6:
                 raise ValueError("Invalid color: %s" % color)
```
kynikos: I wonder if we really need this test... there can be more things wrong in the `color` string than its length, and IMHO the built-in exception that would be normally raised in those cases would be enough.

libqtile/xcbq.py
aniruddhkanojia: ok. should I make another PR ?

libqtile/xcbq.py
kynikos: Sorry, I'm referring to the `if len(color) < 6:` test.

aniruddhkanojia: I used it to avoid list index out of error.

aniruddhkanojia: What do you recommend ?

libqtile/xcbq.py
kynikos: As I said, I'm not an official maintainer, however as you can see the new commit is already visible in this PR, so there's no need to open a new one. tych0 will squash the commits as needed if he thinks it's appropriate.

libqtile/xcbq.py
kynikos: The fact is that you're avoiding `IndexError` just to raise `ValueError` :) If it were me, I'd just remove the test and the `ValueError` line, and let Python raise what's to be raised as needed, but you may want to wait for tych0 for an official opinion.

aniruddhkanojia: I see your point.
I think I will wait for tych0 and hope he is ok with it .
I have spent a lot of unnecessary time on this little Bug :stuck_out_tongue:

kynikos: Yep, maybe `except xcffib.xproto.NameError:` could just `pass`, and its block could be enclosed in another `try` that catches IndexError and other possible exceptions caused by bad `color` strings, for example containing non-hexadecimal characters (which would break the `int` calls I guess), but this is all fussy optimization after all, I agree ;)

aniruddhkanojia: Fingers Crossed :sweat_smile:

tych0: I do think the changes @kynikos has suggested would make this a little more readable, if you can make those the rest of it looks good to me.

aniruddhkanojia: @tych0  is it ok  now ?

aniruddhkanojia:
Closing this as is is duplicate of #638 and #640.

PR #636, opened 2015-03-23
Window._reconfigure_floating: Fix UnboundLocalError with height variable

dequis:


dequis:
Tests fail because of #634, which is fixed in #635

PR #639, opened 2015-03-23
fix mute

### LINKED ISSUE #630: sound widget unmute not working
clicking unmute on the sound widget doesn't unmute the sound.

muhammedabuali:
a tiny fix for volume widget as suggested by @zordsdavini fixes #630

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -112,9 +112,7 @@ class Volume(base._TextBox):
                 subprocess.call([
                     'amixer',
                     '-q',
-                    '-c',
-                    str(self.cardid),
-                    'sset',
+                    'set',
                     self.channel,
                     'toggle'
                 ])


zordsdavini:
I don't know do you need it to hardcode here, because you can simple use like:

            widget.Volume(emoji=True,
                mute_command=[
                    'amixer',
                    '-q',
                    'set',
                    'Master',
                    'toggle'],
                **Theme.widget),

muhammedabuali:
I see what you did there but isn't it better that the widget work correctly by default

zordsdavini:
for others it works. It should have some kind of poll to get most working
value and set it as default

On 23 March 2015 at 13:58, muhammedabuali <notifications@github.com> wrote:

> I see what you did there but isn't it better that the widget work
> correctly by default
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qtile/qtile/pull/639#issuecomment-84964651>.
>



-- 
Arns Udovīčė

tych0:
Sorry, I'm confused about this change. It looks like we're just removing the cardid parameter, is that breaking things?

tych0:
I'll go ahead and close this for now. If you have time to explain the change to me, please re-open.

PR #649, opened 2015-03-27
Fixes Log File Size (MAX : 2 GB) 

### LINKED ISSUE #648: log file too big
Today I left my computer idle(not on sleep) and qtile log file filled my entire hdd.

I know that we started truncating log file on restart after this dfa26b0f0c8bad5a558b54021af6f020e419a6b6

Can't we do something  if the log file gets too big while qtile is running ?

aniruddhkanojia:
I recently faced issues with log file size.
I remember someone in IRC mention it also , even after the patch dfa26b0f0c8bad5a558b54021af6f020e419a6b6

This fixes #648 

Also now we don't need to truncate log file at restarts.

### DIFF
--- a/libqtile/log_utils.py
+++ b/libqtile/log_utils.py
@@ -74,11 +74,11 @@ def init_log(log_level=logging.WARNING, logger='qtile', log_path='~/.%s.log'):
         except TypeError:  # Happens if log_path doesn't contain formatters.
             pass
         log_path = os.path.expanduser(log_path)
-        with open(log_path, "a"):
+        with open(log_path, "w"):
             pass
         handler = logging.handlers.RotatingFileHandler(
             log_path,
-            maxBytes=1000000000,
+            maxBytes=10000000,
             backupCount=1
         )
         handler.setFormatter(
--- a/libqtile/log_utils.py
+++ b/libqtile/log_utils.py
@@ -74,7 +74,7 @@ def init_log(log_level=logging.WARNING, logger='qtile', log_path='~/.%s.log'):
         except TypeError:  # Happens if log_path doesn't contain formatters.
             pass
         log_path = os.path.expanduser(log_path)
-        with open(log_path, "aw"):
+        with open(log_path, "a"):
             pass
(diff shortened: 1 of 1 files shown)


kynikos:
I was wondering if 1GB per file is a little too much for a hard-coded default. The log is configured by command line (https://github.com/qtile/qtile/blob/develop/bin/qtile#L104) or by config file (https://github.com/qtile/qtile/blob/develop/libqtile/manager.py#L68): maybe `maxBytes` (and also `backupCount`?) could be made configurable as well?

tych0:
We can put a limit, but it seems better to fix the source of the problem. Can you post the contents of your log, or an excerpt at least?

aniruddhkanojia:
I was not able to open the file at that time , my computer froze.

However I remember that I was running an instance of Qtile in xephyr and I didn't exit it properly , and it usually keeps pushing this in log files endlessly.

```
ConnectionException: xcb connection errors because of socket, pipe and other stream errors.
2015-03-27 20:33:31,896 ERROR <lambda>:694 Got an exception in poll loop
  File "/usr/local/lib/python2.7/dist-packages/trollius/events.py", line 151, in _run
    self._callback(*self._args)
  File "/home/samfisher/gsoc/qtile/libqtile/manager.py", line 661, in _xpoll
  File "/usr/local/lib/python2.7/dist-packages/xcffib/__init__.py", line 516, in wrapper
    self.invalid()
  File "/usr/local/lib/python2.7/dist-packages/xcffib/__init__.py", line 506, in invalid
```

tych0:
I've just pushed d4c450a2fdeb06ef61e0b3ae4e4154fb9435304c, which should address the particular issue you saw. I think we do want to still truncate the log file on reboots, so we should at least revert that part of this patch. I'm not sure about the max file limit, though. Anyone have any other thoughts?

kynikos:
I don't want to block this with my concern about the size limit: I'm just thinking that a 1GB log file is practically unopenable anyway, so maybe it could be truncated to something like 10-20MB? On a second thought, making this value configurable seems a bit pointless.

PR #651, opened 2015-03-28
Prompt history

soulchainer:
I missed this feature, usual in other aplication launchers (like `gmrun` or the integrated in *awesome WM*, my previous favorite tiling), so I wrote it :).  Basically, it kinda mimics (not exactly) the behavior of shell command history: it keep a list of commands executed previously and, if you hit *arrow up*, and *arrow down* keys, you can navigate between previous commands, relaunch or edit and launch a new command from them.    
After thinking a bit about it, I concluded that the best way to do it was to use a `deque` for store the commands, and serialize to a file with `pickle`. Using plain text seemed more complicated and less optimal to me.                             
When I changed `handle_KeyPress` first time, `flake8` complained (correctly) that the function was too much complex, so I changed the implementation, dividing the function in a bunch of little functions. Now, even when is a bit more long, I think it gains in clarity (opinions?).        
I also made key comments in the parts I edited and add a few docstrings. Still, I want to correct the existing docstrings, to follow *PEP 257*, and add missing ones. This general lack of comments/docstrings made me hard to write my previous widget (*Cmus*), so I want to do it, when I can, if it isn't a problem ;).

I made this change an option, so, if I got it right, you can specify if you want to keep a history of commands and how long you want that list to be.
At least, I can say that it works for me, for now (I'm with it a couple of days, now). Surely, it can be done better, but it works :þ. I'm pretty new with this, sorry. I'm trying my best ;). Learning, learning.


libqtile/widget/prompt.py
```diff
@@ -403,36 +431,109 @@ def _update(self):
...
+            else:
+                self.position -= 1
+                self.archivedInput = self.history[self.position]
+
+    def _get_next_cmd(self):
+        # Get the next command in history.
+        # If the last command was already reached, ring system bell.
+        if self.record_history:
+            if self.position == len(self.history):
+                print("\a")
```
tych0: Does this actually beep here? I'd be amazed if it did.

tych0:
Thanks, this looks like a nice improvement. I'm not so sure we want to do a shell beep though, if it does work. If we keep it, I'd at least like to make it configurable. Alternatively, we could blink the text red or something.

libqtile/widget/prompt.py
soulchainer: Of course it bells.

soulchainer:
Well, that's the default shell behavior :þ. But yes, tomorrow (now here is almost tomorrow, I need to sleep xD) I will try to do something about it :).

libqtile/widget/prompt.py
dequis: Only if there's some terminal connected to stdout, which is not always the case

soulchainer: Hi n_n.
Ummm... According [Wikipedia](http://en.wikipedia.org/wiki/Standard_streams#Standard_output_.28stdout.29) (yes, kill me xD):
> Unless redirected, standard output is the text terminal which initiated the program.

So... it should,be connected in most situations, no?
But, well: I will check that before try to beep then. Not a big deal ;).
Thanks for the info, @dequis. I didn't know about this.

tych0: I think it probably depends on how your login manager spawns things; you may not be connected to a tty at all. You can use `os.isatty` to detect this if you really want to keep the beep, though.

soulchainer: @tych0 I was using `sys.stdout.isatty` for this, but, now that you are saying that, I read about `sys.stdout` and:

> Note Under some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None.

So, I figure that if you try `isatty` over `None`, the universe implodes xD. Using `os.isatty(1)` is safer than that, no?
Sorry, I'm a bit lost about this.

soulchainer:
The DDOS attack that GitHub is suffering is painful.
Well, I just can commit the last changes.

I made system bell optional and add several options.
System bell can be now the usual beep or a visual bell (it changes momentarily prompt text color).
Visual bell color and duration is also configurable (for example, someone can have a red taskbar, and red over red is kinda invisible). Also, check if we are connected to a tty or not (I used `os.isatty`, like @tych0 suggested).

BTW: I'm not so sure about that log message for when we can't ring a bell (basically, I write a info level log about that, saying that stdout is not connected to a tty). I don't know if this is appropiate or not, but I feel like that :þ. Keep it? Delete it?

libqtile/widget/prompt.py
dequis: Ehhh, this whole method of beeping seems like a bad idea. You could use `qtile.conn.conn.core.Bell(0)` instead (making a method for it in xcbq might be nicer)

Roger:
what happens if you have more than one completer? ie, find window, move to group?

dequis:
This got lost in one of the outdated diff comment threads above, quoting myself:

>Ehhh, this whole method of beeping seems like a bad idea. You could use `qtile.conn.conn.core.Bell(0)` instead (making a method for it in xcbq might be nicer)

soulchainer:
@Roger What about the completers? This is about this merge request or about the previous prompt behavior? Because I didn't touch an inch about the completers. So, I guess that if we have more than one completer, the behavior will be the same than like before this. Or not? Please, clarify me your question, for my understanding. Thanks.
@dequis Don't worry, I also got the last one comment. And well... Honestly, I'm pretty lost about this. I tried to understand that `qtile.conn.conn.core.Bell(0)` (all the day), checked `xcffib`... and it's far beyond my knowledge, right now. And the obscure/lack of (for me, at least) documentation about it doesn't help :(. I don't know. If this is only about the beep thing, I guess that I can get rid of it (even when I liked the idea) and preserve only the visual alert. Right now, I can't figure this out. Sorry.

Roger:
i mean, the prompt history will be global, not based on the completer attached at the time, right? it would be cool having the history based on the completer type

Roger:
What's the problem with qtile.conn.conn.core.Bell(0)? i works fine here, try eval("self.conn.conn.core.Bell(0)") on qsh, for example.

About the docs of xcffib, you should check the official docs of libxcb, it's an autogenerated binding of that library, also most of the code for xpyb should apply to xcffib as well.

soulchainer:
@Roger It was more that I was overthinking it and I wanted to "find" first where the Bell function was defined/implemented, and I transversed today the basecode, from qtile, to xcffib, later other docs, and I couldn't understand all that code so suddenly, so I felt overwhelmed and helpless xD. And... In fact (even when I knew it was `self.conn.conn.core.Bell(0)`), because of that, I didn't even tried to simply check if it works. Silly, right? Well, I will try again tomorrow, taking my time (after all, there is no hurry really) and checking if I can figure out that pair of things. Gracias.

soulchainer:
Well. I think the discussed changes are done.
Changed the beep method, like @dequis said, using `qtile.conn.conn.core.Bell(0)`.
And changed the history to be completer based (basically, instead of a plain `deque`, now the history is a dictionary with a deque object for each completer history).
Still want to do a few improvements, but for now the thing is working.

tych0:
Thanks!

PR #655, opened 2015-03-31
update cursors faq on docs

Roger:


soulchainer:
`defualt` (now that you're still in time ^^).

PR #656, opened 2015-04-04
Handle #619 for Zoomy

luxrck:
BTW, I think we always put floating window on the top.
But now, for example, in TreeTab, we float a window and then we create a new window. The new created window will cover that float window.
So what's your opinion ?
(please just ignore any errors you found in my content, I'm not good at english...>_<)

### DIFF
--- a/libqtile/layout/zoomy.py
+++ b/libqtile/layout/zoomy.py
@@ -57,21 +57,19 @@ class Zoomy(SingleWindow):
 
     def focus_last(self):
         if self.clients:
-            return self.clients[len(self.clients) - 1]
+            return self.clients[-1]
 
     def focus_next(self, client):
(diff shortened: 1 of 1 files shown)


libqtile/layout/zoomy.py
```diff
@@ -141,7 +141,7 @@ def focus(self, win):
         if self.focused and self.property_name and self.focused.window.get_property(
             self.property_name,
             "UTF8_STRING"
-        ) is not None:
```
tych0: Hi,

Is there a particular reason to drop this `is not None`? Thanks.

luxrck:
I see the code just now and... >_< My fault, I thought they were the same....

tych0:
Thanks!

PR #659, opened 2015-04-05
Inline command edition and a few errors corrected (prompt widget).

soulchainer:
Added support for cursor movement across the input text,
edition in place (add/delete text before the cursor, and
delete in cursor position with *Delete* key).

Added options for show or not the text cursor and for
choose their color. Cursor position is noticeable by
the text cursor and by the character above it in the same
color than the cursor (even if you choose to not show the
text cursor).

Corrected a few errors in previous code:
- When `cursorblink` was set to zero, the cursor simply
dissapear. If this is a feature, is deceiving: If
`cursorblink` is zero it should, you know, just not
blink. That's what it do now.
- The widget didn't clear the text (*"hide the widget"*) in focus change with
`cursorblink` equal to zero. This also happened when you 
choose to not show the cursor. Now it behaves ok in both cases.

Also, function in charge of return the appropiate action to one
key (`_key_handler`) was renamed (`_get_keyhandler`) and simplified,
creating a dictionary on `__init__` (for no rebuilt it every time a new
key is handled), with pairs (keycode, function to execute).

I still want to review this and the rest of this widget code, for check if I can improve/simplify something, but this already works for me.

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -618,8 +618,9 @@ class Prompt(base._TextBox):
         state = e.state & ~(self.qtile.numlockMask)
         keysym = self.qtile.conn.keycode_to_keysym(e.detail, state)
         handle_key = self._get_keyhandler(keysym)
-        handle_key()
-        del self.key
+        if handle_key:
+            handle_key()
+            del self.key
         self._update()
 
     def cmd_fake_keypress(self, key):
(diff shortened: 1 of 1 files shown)


soulchainer:
Just noticed right now: I need to fix the behaviour when `bell_style="visual"` (color from cursor (and text above it) overlap visual bell one (when triggered). Sure it's trivial :þ. I will check it when I have a moment.

soulchainer:
Well, Solved that `bell_style="visual"` thing with that amend. Basically, I «cheated»: Now, instead of change the text color of the prompt when the bell is triggered, the widget background color is changed (thing that, in fact, is more like the shell behavior [at least in xterm]). With this, even then the cursor text is highlighted, it looks good (well, it depends of the user color choices).

tych0:
Thanks, sorry it took me so long to get to this. Looks like a nice improvement!

PR #663, opened 2015-05-02
Improvements to KeyboardKbdd widget.

tuxitop:
Changed parent to ThreadedPollText.
Better way to check for "kbdd" process at startup.


tych0:
Sorry for the delay, and thanks for the patch!

PR #664, opened 2015-05-02
Added CheckUpdates widget.

tuxitop:
CheckUpdates is a widget that shows number of pending updates in different unix systems.
At first I wanted to improve the pacman widget that only used "checkupdates" method (which I don't like.) and add a different command to check for updates in Arch. Then I came up with a better idea. a widget that checks number of outdated packages in different distros. I added some popular distributions, and this way anyone can add more distros.
This widget does everything that the pacman widget does, so maybe the pacman widget should get removed.


tych0:
Hrm, somehow travis didn't detect this was merged. Anyway, it was, thanks! Sorry for the delay.

PR #668, opened 2015-05-15
Cleanup &Speedup travis

Azd325:
* Add pip cache
* fix formatting and order
* fix typo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,22 +27,22 @@ matrix:
     - python: pypy3
     - python: pypy
 
-script:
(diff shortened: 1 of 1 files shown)


tych0:
Hi, does the caching really work? travis' docs say it's only for private repos and stuff that runs on their container infrastructure, which I don't think we do (because we use sudo):

http://docs.travis-ci.com/user/caching/
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

Azd325:
You're right!

PR #672, opened 2015-05-27
Improved Serialization for Layouts

aniruddhkanojia:
At present support for Layouts Max & Zoomy.

### DIFF
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -56,7 +56,6 @@ class QtileState(object):
                     if layout.name in ['max', 'zoomy']:
                         layout.restore_state(self.layoutMap[group.name][layout.name], qtile.windowMap)
         except KeyError:
-            print "ERROR"
             pass  # group missing
 
         for (group, layout) in self.groups.items():
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -55,7 +55,6 @@ class QtileState(object):
         """
         try:
             for group in qtile.groups:
-                group.focus_history = [qtile.windowMap[i.wid] for i in self.focus_history[group.name]]
                 for layout in group.layouts:
                     d = self.layout_map[group.name][layout.name]
                     d.group = layout.group
@@ -80,6 +79,7 @@ class QtileState(object):
                                 setattr(layout, member, x)
                         except AttributeError:
                             pass
+                group.focusHistory = [qtile.windowMap[i.wid] for i in self.focus_history[group.name]]
         except KeyError:
             pass  # group missing
 
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -55,6 +55,7 @@ class QtileState(object):
         """
         try:
             for group in qtile.groups:
+                group.focusHistory = [qtile.windowMap[i.wid] for i in self.focus_history[group.name]]
                 for layout in group.layouts:
                     d = self.layout_map[group.name][layout.name]
                     d.group = layout.group
@@ -79,7 +80,6 @@ class QtileState(object):
                                 setattr(layout, member, x)
                         except AttributeError:
                             pass
-                group.focusHistory = [qtile.windowMap[i.wid] for i in self.focus_history[group.name]]
         except KeyError:
             pass  # group missing
 
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -66,18 +66,20 @@ class QtileState(object):
                                 last=None
                                 for i in x:
                                     if isinstance(i,window.Window):
+                                        print qtile.windowMap[i.Wid],member
                                         tmp.append(qtile.windowMap[i.Wid])
                                     else:
                                         tmp.append(i)
                                     last=i
                                 if isinstance(last,window.Window):
-                                    setattr(d,member,tmp)
-
-                            if isinstance(x,window.Window):
-                                setattr(d,member,qtile.windowMap[x.Wid])
+                                    setattr(layout,member,tmp)
+                            elif isinstance(x,window.Window):
+                                setattr(layout,member,qtile.windowMap[x.Wid])
+                            elif not callable(x) and not str.startswith(member,'_'):
+                                setattr(layout,member,x)
                         except AttributeError as e:
                             pass
-                    layout=d
+                    # layout=d
         except KeyError:
             pass  # group missing
 
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -81,7 +81,7 @@ class QtileState(object):
                                 setattr(layout, member, qtile.windowMap[x.Wid])
                             elif not callable(x) and not str.startswith(member, '_'):
                                 setattr(layout, member, x)
-                        except AttributeError as e:
+                        except AttributeError:
                             pass
         except KeyError:
             pass  # group missing
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -32,17 +32,17 @@ class QtileState(object):
         # configurations.
         self.groups = {}
         self.screens = {}
-        self.focusHistory = {}
+        self.focus_history = {}
         self.current_screen = 0
-        self.layoutMap = {}
+        self.layout_map = {}
 
         for group in qtile.groups:
             self.groups[group.name] = group.layout.name
-            self.layoutMap[group.name] = {}
+            self.layout_map[group.name] = {}
             for layout in group.layouts:
-                self.layoutMap[group.name][layout.name] = layout
+                self.layout_map[group.name][layout.name] = layout
                 layout.group = None
-            self.focusHistory[group.name] = group.focusHistory
+            self.focus_history[group.name] = group.focusHistory
         for index, screen in enumerate(qtile.screens):
             self.screens[index] = screen.group.name
             if screen == qtile.currentScreen:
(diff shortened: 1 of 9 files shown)


tych0:
Hi,

Thanks for the patch. It seems as though the `get_state` and `set_state` functions could be generalized at least to `SingleWindow` in this case since the code is basically identical, but also to other layouts. Did you try (something like) pickling to see if the layout state was preserved correctly? Ideally we'd like to do this automatically without layout implementors having to worry about it.

aniruddhkanojia:
Yeah you are right I could generalize this ,atleast for the `SingleWindow` class. I just wanted to see if I was moving on the right track.

I tried , pickling the `layout` state directly but it can't be pickled ,some changes would be required for the layout class pickleable. Mainly the attributes referencing `window` objects were causing problems as `window` class is not pickelable.  Hence I had to extract their `id` to get pickling working.

tych0:
What if we implement `__getstate__` and `__setstate__` for windows?

aniruddhkanojia:
I Implemented `__getstate__` and `__setstate__` using only window id's . I then successfully restored them to their original state using the windowmap.

The appropriate changes are made to the layout state but still it doesn't reflect in actual.

libqtile/window.py
```diff
@@ -1316,3 +1316,14 @@ def cmd_set_position(self, dx, dy, curx, cury):
...
                 break
+
+    def __getstate__(self):
+        d = {}
+        d['group'] = self.group.name
+        d['window'] = self.window.wid
+        return d
+
+    def __setstate__(self, state):
+        self.Wid = state['window']
```
tych0: Looks like this is a typo and should be lowercase `wid`.

tych0: In fact, since these are coming from self.group.name and going to self.gid, I don't understand how/why this works.

libqtile/state.py
```diff
@@ -45,6 +53,39 @@ def apply(self, qtile):
             Rearrange the windows in the specified Qtile object according to
             this QtileState.
         """
+        try:
+            for group in qtile.groups:
+                x = []
+                for i in self.focusHistory[group.name]:
+                    x.append(qtile.windowMap[i.Wid])
+                group.focusHistory = x
+                for layout in group.layouts:
```
tych0: It looks like all this is some code to pickle by hand. Is there any reason for not just using pickle directly? This is where your `__getstate__` and `__setstate__` methods would be used; I think they're unused now.

libqtile/window.py
aniruddhkanojia: I purposefully made it `Wid` as it was an attribute name no one had , I was earlier using this to differentiate between a pickled window class and non-pickled one.

`self.gid` is an unused atttribute , I will remove it.

libqtile/state.py
aniruddhkanojia: `__getstate__` & `__setstate__` methods made it possible to pickle the layout class and window class. The group class is still *unpickeble* because of recursive attributes.(`group` has a link to `qtile` which has a link back to `group`). 
We don't need to pickle the group class , only one of it attributes `focusHistory` was enough for everything to get working and so I directly pickled it.If we restore the group state back to original , any changes made in configuration would not reflect after restart. Hence I only pickled it's attribute `focusHistory`.

libqtile/window.py
aniruddhkanojia: Also this doesn't restore the window class back to its original form. After unpickling only a single attribute of the window class remains .The `Wid` which I use later in to get the original window state by using `qtile.windowMap`. Pickling by itself cannot make restore the window state.

libqtile/state.py
aniruddhkanojia: If I use pickling only, some of the attributes always get lost. For example in `window` class, it has an attribute group which has a link back to the window. This is a recursive chain , so  during `__getstate__` I pickled every attribute of the window class but not it's group.However , after unpickling many of the attributes of the window class became *unset*. Removing the `group` attribute causes a lot of changes to the `window` state and so as it wasn't possible for me to restore the whole state back to original by just using picklng alone. I therefore only stored the `wid` during pickling and used it to get the window information back later.

libqtile/state.py
```diff
@@ -31,10 +32,17 @@ def __init__(self, qtile):
...
+        self.focusHistory = {}
         self.current_screen = 0
+        self.layoutMap = {}
 
         for group in qtile.groups:
             self.groups[group.name] = group.layout.name
+            self.layoutMap[group.name] = {}
+            for layout in group.layouts:
+                self.layoutMap[group.name][layout.name] = layout
+                layout.group = None
```
flacjacket: Why are you setting this to None? Why not use a group property for Layout? Is it set anywhere else?

libqtile/state.py
```diff
@@ -45,6 +53,39 @@ def apply(self, qtile):
             Rearrange the windows in the specified Qtile object according to
             this QtileState.
         """
+        try:
+            for group in qtile.groups:
+                x = []
+                for i in self.focusHistory[group.name]:
+                    x.append(qtile.windowMap[i.wid])
+                group.focusHistory = x
```
flacjacket: You can do this with list comprehension:
```
group.focusHistory = [qtile.windowMap[i.wid] for i in self.focusHistory[group.name]]
```

libqtile/window.py
```diff
@@ -1316,3 +1316,12 @@ def cmd_set_position(self, dx, dy, curx, cury):
                 self.group.layout.focused = index2
                 self.group.layoutAll()
                 break
+
+    def __getstate__(self):
```
flacjacket: Where are the `__getstate__` and `__setstate__` functions used? I don't see where they'd be called

libqtile/state.py
flacjacket: Tho where exactly do you use the focus history in the group?

libqtile/state.py
```diff
@@ -45,6 +53,39 @@ def apply(self, qtile):
...
+                for layout in group.layouts:
+                    d = self.layoutMap[group.name][layout.name]
+                    d.group = layout.group
+                    members = dir(d)
+                    for member in members:
```
flacjacket: I'll admit, I haven't worked with the state code, but what is it you're trying to pull out here? Quickly searching, I don't see where a Window would end up in a Layout'- dir().

flacjacket:
Is there any way you can add a test for this stuff? Setup a layout, switch it and switch back?

I haven't worked with the state code, but I'm not sure how all the pieces you set up fit together to achieve the goal, I'd appreciate the clarifications I mentioned.

libqtile/state.py
aniruddhkanojia: Class `group` has multiple attributes that refer back to `layout` class and it ends up in recursion.If this attribute is not set to `None`, an endless recursion will be caused. Thus for python pickle to work I had to break this recursive chain and I set this attribute to None.Since the mapping stores the `group.name` , I use this to restore this value after restart. In the `apply` method I restore this attribute in `line 64`.

libqtile/state.py
aniruddhkanojia: Yes, that a nice suggestion .I will add it to the changes. Thanks :)

I add that in `line 45` in `__init__` method

libqtile/state.py
aniruddhkanojia: All layouts are derived classes of the base layout class and they have multple attributes refering to `window` class.
If you were to see [Max Layout]( https://github.com/qtile/qtile/blob/develop/libqtile/layout/max.py) it has multiple attributes that refer to objects of window class like `clients` , `current` etc. Similarly other layouts have different attributes with different names that refer to `window` class objects.

libqtile/window.py
aniruddhkanojia: I don't directly pickle a `window` object. However as there are many attributes of `layout`  class which are `window` class objects.When I pickle a `layout` class , they get pickled as well. 

The `object` class has many links to `group` class and `layout` class. So it is not possible to pickle the window state in its original form. Using the fact that after restart the `qtile` object has a `windowMap` attribute which I can use to restore the window state back to original by just using the wid .Hence I used this set of  `__gestate__` and `__setstate__`  functions.

libqtile/state.py
```diff
@@ -31,10 +32,17 @@ def __init__(self, qtile):
         # configurations.
         self.groups = {}
         self.screens = {}
+        self.focusHistory = {}
```
ramnes: New code should use `snake_case` rather than `mixedCase`: http://docs.qtile.org/en/latest/manual/hacking.html#coding-style

libqtile/state.py
flacjacket: Ah, alright, I see.

flacjacket:
Alright, I think I am getting this more, thanks for the explanations. Do you think you can add any tests for this? It'd be nice for figuring out what works and what doesn't. Have you made any further progress since then? You can keep working on new material in new branches while waiting on and responding to review.

libqtile/state.py
aniruddhkanojia: Thanks for the reminder :smiley:

aniruddhkanojia:
I was travelling so I haven't been able to do so much. 
I will add tests to this ASAP.

aniruddhkanojia:
I couldn't find any test which needed restarts. I used `self.c.restart()` in nosetests but it returns the following error `nosetests: error: no such option: --no-spawn`. I know it's related to `cmd_restart()` function in qtile.

Any help in this regard would be appreciated .
The complete error [output ](http://pastebin.com/JNwZEHqC)

Roger:
Yeah, that's trying to restart the main proccess, nose.

I guess you can simulate restart here, and give the new qtile instance the old state, https://github.com/qtile/qtile/blob/develop/test/utils.py#L68

flacjacket:
I think this looks good, is there anything else?

aniruddhkanojia:
Yeah, I was just waiting for your review

PR #678, opened 2015-06-14
added to .gitignore pycharm ide

llybin:


Kilte:
Why you don't use a global .gitignore?
IMO, such files should be ignored by global .gitignore only.

llybin:
Sorry, I didn't understood what you mean about global gitignore.
I became interested in the project. I plan to send another pull requests in the project. I use PyCharm IDE, it creates the .idea directory in the project. I saw that .gitignore contains a line for sublime-text. I thought: why not. I apologize if I hurried to make the pull request. This is my first pull request in open source project.

Kilte:
@llybin, 

> I didn't understood what you mean about global gitignore.

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

> I saw that .gitignore contains a line for sublime-text. I thought: why not.

Perhaps you're right. But I think it's wrong, because projects' `.gitignore` should not know about IDEs and text editors. There are many editors and it think it's better to assign this responsibility to the users of these editors.

dequis:
It *could* have IDE specific stuff, or maybe it would be better to leave it out - it's a matter of preference (I personally don't care either way)

But it's certainly better to use the global one for stuff like this, otherwise you'll have to submit PRs like this one to every project you want to contribute to.

llybin:
Thanks for the answer. I understood.

PR #682, opened 2015-06-23
Added CurrentScreen widget

alexf91:
A small widget which indicates if the screen the widget is on is active or inactive.

### DIFF
--- a/libqtile/widget/currentscreen.py
+++ b/libqtile/widget/currentscreen.py
@@ -44,23 +44,20 @@ class CurrentScreen(base._TextBox):
 
     def _configure(self, qtile, bar):
         base._TextBox._configure(self, qtile, bar)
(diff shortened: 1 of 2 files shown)


libqtile/widget/currentscreen.py
```diff
@@ -0,0 +1,66 @@
...
+        self.setup_hooks()
+
+    def setup_hooks(self):
+        def hook_response():
+            if self.qtile.currentScreen == self.bar.screen:
+                self.text = self.active_text
+                self.foreground = self.active_color
+            else:
+                self.text = self.inactive_text
+                self.foreground = self.inactive_color
```
dequis: This is copypasted from the configure function - split it to an update function that both call?

flacjacket:
This looks good, thanks for this.

PR #684, opened 2015-06-27
Include python-xcffib warning for arch install

andrewyoung1991:
to build qtile-python3-git on arch linux with pacman i had to add community-testing to my mirrors list, (which contains the correct version of python-xcffib).

### DIFF
--- a/docs/manual/install/arch.rst
+++ b/docs/manual/install/arch.rst
@@ -20,6 +20,8 @@ if you use `yaourt`_:
     # or for develop
     yaourt -S qtile-python3-git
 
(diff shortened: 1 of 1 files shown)


flacjacket:
This would be fine, but I'm pretty sure the docs on RTD aren't live, and by the time the person in charge of building docs synced it, Arch packaging will probably be fixed...

@Roger do you handle xcffib, too? ETA to having time to fix the Arch packaging fixed?

Roger:
well, now that arch updated dependencies qtile-python3-git and qtile-git  should work fine, the problem now is with stable packages.. them should get fixed when we realease a new version of qtile

dequis:
I think this PR could be closed

flacjacket:
Agreed, @andrewyoung1991 let us know if you still have any problems

PR #686, opened 2015-06-28
Spawn multiple applications on one group

apinsard:


libqtile/dgroups.py
```diff
@@ -127,8 +127,13 @@ def _setup_groups(self):
             self.add_dgroup(group, group.init)
 
             if group.spawn and not self.qtile.no_spawn:
-                pid = self.qtile.cmd_spawn(group.spawn)
-                self.add_rule(Rule(Match(net_wm_pid=[pid]), group.name))
+                if isinstance(group.spawn, str):
```
tych0: You may want to use `six.string_types` here instead of `str` to be compatible with py2/3.

apinsard: Right, fixed.

Roger:
Would you mind updating doc/type here? https://github.com/qtile/qtile/blob/develop/libqtile/config.py#L437

apinsard:
Sure. I don't know this doc format, may I set type as "string or iterable" or "string or strings list"?

Roger:
Looking at sphinx docs i guess "string or list of strings" is ok.

flacjacket:
Thanks for this!

PR #689, opened 2015-06-29
Handle _NET_WM_STATE_DEMANDS_ATTENTION 

drunkbear:
Hi there,
I've added handling of the _NET_WM_STATE_DEMANDS_ATTENTION for setting window urgency state.
It is considered weaker than flag in WM_HINTS, but many applications (I believe all the Qt applications) will use it instead.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -290,7 +290,7 @@ class _Window(command.CommandObject):
 
     @property
     def urgent(self):
-        return self.hints['urgent'] | self._demands_attention
+        return self.hints['urgent'] or self._demands_attention
 
     @urgent.setter
     def urgent(self, val):
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -142,6 +142,7 @@ class _Window(command.CommandObject):
         self.state = NormalState
         self.window_type = "normal"
         self._float_state = NOT_FLOATING
+        self._demands_attention = False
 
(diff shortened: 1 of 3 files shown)


libqtile/window.py
```diff
@@ -142,6 +142,7 @@ def __init__(self, window, qtile):
         self.state = NormalState
         self.window_type = "normal"
         self._float_state = NOT_FLOATING
+        self._demands_attention = False
```
flacjacket: I feel like this should be a hint rather than a property

flacjacket: Ah, actually hints means wm_hints

libqtile/window.py
```diff
@@ -275,18 +276,28 @@ def updateHints(self):
...
+
+        if state:
+            self.qtile.log.debug('_NET_WM_STATE: %s' % ','.join(state))
+            for s in triggered:
+                setattr(self, s, (s in state))
 
     @property
     def urgent(self):
-        return self.hints['urgent']
+        return self.hints['urgent'] | self._demands_attention
```
flacjacket: You should use `or` rather than `|`, which is bitwise or

drunkbear: Indeed, fixed it.

flacjacket:
I pushed this in for 0.10.0, thanks!

PR #693, opened 2015-07-09
fixed margin xmonad layout

Ippolo:
the margin between the main pane and secondary panes was double

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -378,7 +378,6 @@ class MonadTall(SingleWindow):
                 self.group.screen.dheight - 2 * self.border_width - 2 * self.margin,
                 self.border_width,
                 px,
-                margin=0,
             )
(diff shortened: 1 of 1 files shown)


flacjacket:
Isn't it supposed to have double the margin between windows? Looking at the other layouts, the use of the `margin` kwarg is used in this way. Note that you also have double the margin vertically between secondary windows when you have >=3 windows open.

flacjacket:
Though it looks like Tree might specify different types of margins (left vs right, etc)

Ippolo:
"This layout attempts to emulate the behavior of XMonad's default tiling scheme."
XMonad's default layout does not have double margin =)

flacjacket:
Ah, well you got us there. I see no harm in this, especially since the default is `margin=0`, I'll merge this in, if anyone complains, we can start bikeshedding it.

PR #694, opened 2015-07-11
Change wait process from xterm to qtile on xephyr script

cocuh:
In current verison, xephyr start script waits xterm. Exiting xterm, the script will end. It should wait qtile one. When qtile shutdown, i consider xephyr also should shutdown just like xorg. i have fixed the xephyr startup script.


flacjacket:
Ah, yes, that's probably right, thanks for the catch!

PR #695, opened 2015-07-12
Fixed logging, when error occurred at reading config file.

himaaaatti:
I missed error log at ".qtile.log", when config file have syntax error. So, I added code for log message about that.


flacjacket:
This is a good change, thanks.

PR #698, opened 2015-07-14
Add doc on Funtoo installation

apinsard:


flacjacket:
Thanks for the updated info!

PR #699, opened 2015-07-14
execute command on push check_update widget

zordsdavini:
pacman widget has additional functionality to execute command on push by mouse. As this widget is depricated I moved this part to check_update widget

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -83,4 +83,4 @@ class CheckUpdates(base.ThreadedPollText):
     def button_press(self, x, y, button):
         base.ThreadedPollText.button_press(self, x, y, button)
         if button == 1 and self.execute is not None:
-            Popen([self.execute], shell=True)
+            Popen(self.execute, shell=True)
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -19,7 +19,7 @@
 # SOFTWARE.
 
 from . import base
-from subprocess import CalledProcessError
+from subprocess import CalledProcessError, Popen
 
 
 class CheckUpdates(base.ThreadedPollText):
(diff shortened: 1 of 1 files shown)


libqtile/widget/check_updates.py
```diff
@@ -78,3 +79,8 @@ def poll(self):
         if not self.cmd:
             return "N/A"
         return self._check_updates()
+
+    def button_press(self, x, y, button):
+        base.ThreadedPollText.button_press(self, x, y, button)
+        if button == 1 and self.execute is not None:
+            Popen([self.execute], shell=True)
```
tych0: You might make this just self.execute here, so that people can pass in a list if their distro update command (i.e. `apt-get deselect-upgrade`) is more than one word.

zordsdavini: for now I use 'urxvt -e bash -c "yaourt -Syua"' and this works. May be I
should check is it list, overwise split it to list?


On 14 July 2015 at 17:27, Tycho Andersen <notifications@github.com> wrote:

> In libqtile/widget/check_updates.py
> <https://github.com/qtile/qtile/pull/699#discussion_r34573256>:
>
> > @@ -78,3 +79,8 @@ def poll(self):
> >          if not self.cmd:
> >              return "N/A"
> >          return self._check_updates()
> > +
> > +    def button_press(self, x, y, button):
> > +        base.ThreadedPollText.button_press(self, x, y, button)
> > +        if button == 1 and self.execute is not None:
> > +            Popen([self.execute], shell=True)
>
> You might make this just self.execute here, so that people can pass in a
> list if their distro update command (i.e. apt-get deselect-upgrade) is
> more than one word.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qtile/qtile/pull/699/files#r34573256>.
>



-- 
Udovīčė Arns

flacjacket:
This looks fine to me now, I can merge this later today.

flacjacket:
This is merged

PR #705, opened 2015-07-17
widget.ThermalSensor regex and py3 compatibility

andrewyoung1991:
cleaning up the ThermalSensor sensors_temp regex and accommodating py3 degrees symbol match.

try in python 3.3+ shell to confirm compatibility:
```python
>>> assert re.match(r"\xc2\xb0", "°") == None
>>> assert re.match(r"\xb0", "°")
```

### DIFF
--- a/libqtile/widget/sensors.py
+++ b/libqtile/widget/sensors.py
@@ -27,7 +27,7 @@
 # coding: utf-8
 
 from . import base
-from six import u
+from six import u, PY2
 
 import re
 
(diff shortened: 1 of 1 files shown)


libqtile/widget/sensors.py
```diff
@@ -88,7 +88,7 @@ def get_temp_sensors(self):
         except OSError:
             return None
         temp_values = {}
-        for value in re.findall(self.sensors_temp, sensors_out):
+        for value in = self.sensors_temp.findall(sensors_out):
```
flacjacket: I think you need to drop the `=` here

andrewyoung1991: whoops yeah (error between computer and chair)

andrewyoung1991:
this commit got really messy, going to close and reopen

PR #706, opened 2015-07-17
Py3 thermalsensor regex

andrewyoung1991:
ThermalSensor widget was not python3 complient in its regex for
capturing the degree symbol. try the following in the python shell to
confirm.
```python
>>> assert re.match(r"\xc2\xb0", "°") == None
>>> assert re.match(r"\xb0", "°")
```
- includes some general cleaning up on the ThermalSensor class
- includes a unittest for ThermalSensor widget
- includes a util `catch_exception_and_warn` that can be used as a
decorator to catch an exeption and issue a warning in its place if a
method raises an exception.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__
 MANIFEST
 /build
 /dist
+/env
 # complied cffi output
 libqtile/_ffi_pango.py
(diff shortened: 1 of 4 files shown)


flacjacket:
Thanks, this looks good

PR #707, opened 2015-07-18
move window to current group

Ippolo:
I need this feature because I'd like to use an extra group for minimized windows.
this way I can move a window to the group and restore it with a simple keyboard shortcut.

I know there's toggle_minimize(), but I'd still prefer to use an extra group. Hope you won't find this commit useless


flacjacket:
This looks good, thanks!

PR #709, opened 2015-07-20
clear windowname if group has no window

Ippolo:
If I close all the windows in a group the widget still displays the name of the latest opened window (I don't understand why looking at the code...), while it should display nothing


flacjacket:
Forgot to close this in the commit log, but this is merged, thanks!

PR #712, opened 2015-07-21
add possibility to not show window state on windowname widget

Ippolo:
I find this feature a bit confusing and also useless since I can see the state of a window by just looking at it =)


flacjacket:
You can make this a widget option by creating a defaults list and adding show_state to it, for example, see the clock widget: https://github.com/qtile/qtile/blob/develop/libqtile/widget/clock.py#L46

Ippolo:
@flacjacket thank you, I created the defaults list as suggested

PR #713, opened 2015-07-22
update ubuntu docs

tych0:
cairocffi and xcffib are in debian unstable and wily, so let's have people
install those if they can.


flacjacket:
Just pushed this in, thanks for the update! A bit strange to see a PR come in from you :)

Issue #716, opened 2015-07-23
Unclosed file in widget/graph.py


kseistrup:
The pseudo file `/proc/net/route` is never closed in the `get_main_iface()` method.  The patch below is one way to remedy the situation:

```diff
diff --git a/libqtile/widget/graph.py b/libqtile/widget/graph.py
index 5ab252c..ff75bef 100644
--- a/libqtile/widget/graph.py
+++ b/libqtile/widget/graph.py
@@ -349,10 +349,11 @@ class NetGraph(_Graph):
 
     @staticmethod
     def get_main_iface():
-        filename = "/proc/net/route"
-        def make_route(line):
-            return dict(zip(['iface', 'dest'], line.split()))
-        routes = [make_route(line) for line in list(open(filename))[1:]]
+        filename = '/proc/net/route'
+        make_route = lambda line: dict(zip(['iface', 'dest'], line.split()))
+        with open(filename, 'r') as fp:
+            lines = fp.readlines()
+        routes = [make_route(line) for line in lines[1:]]
         try:
             return next(
                 (r for r in routes if not int(r['dest'], 16)),
```

dequis:
Ever heard of "pull requests"? They are really cool stuff.

flacjacket:
@kseistrup dx is right, we'd be happy to merge any of your 3 issues if you want to submit pull requests

PR #722, opened 2015-07-23
Unclosed file in line 355

kseistrup:
The method get_main_iface() doesn't close the file it opens.  Merging this pull-request fixes that and closes issue #716.


flacjacket:
Thanks for the fixes, these are all in!

PR #724, opened 2015-07-24
update deps for debian packaging

tych0:
At some point it would be best for us to tag a point release; but let's wait
until cairocffi 0.7 is in debian unstable, and then we can tag a point release
and do a debian upload. For now, anyway, these are the right dependencies.


flacjacket:
Thanks for the update!

PR #731, opened 2015-07-30
groupbox widget - visible_groups

banthonywalker:
added new configuration property to groupbox widget for defining which groups should be visible in groupbox


flacjacket:
Thanks, just merged this in!

PR #738, opened 2015-08-08
use the .0 soname

### LINKED ISSUE #734: lib*.so references in pangocffi.py
When upgrading from 0.9.1 to 0.10.1, I needed to modify the following references for my system (Ubuntu Vivid) in libqtile/pangocffi.py

gobject = ffi.dlopen('libgobject-2.0.so')
pango = ffi.dlopen('libpango-1.0.so')
pangocairo = ffi.dlopen('libpangocairo-1.0.so')

tych0:
On rpm and deb based distros, these libraries (symlinks) are provided by
the -dev version of the packages, where the .0 name is provided by the
actual library package. Let's use the .0 name instead so we don't have an
undocumented dependency on -dev packages.

Closes #734

### DIFF
--- a/libqtile/pangocffi.py
+++ b/libqtile/pangocffi.py
@@ -54,9 +54,9 @@ try:
 except ImportError:
     from libqtile.ffi_build import pango_ffi as ffi
 
-gobject = ffi.dlopen('libgobject-2.0.so')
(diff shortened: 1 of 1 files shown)


flacjacket:
Cool, thanks!

PR #740, opened 2015-08-09
Catch exceptions during serialization

aniruddhkanojia:


flacjacket:
Thanks for fixing this!

PR #742, opened 2015-08-13
Changes error text if no battery is available

Azd325:
I think this is not an error if no a battery is available because if you a work on desktop pc or a laptop where you removed the battery you don't need this kind of info. That will be more like plug and play. :)

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -151,26 +151,22 @@ class Battery(_Battery):
     """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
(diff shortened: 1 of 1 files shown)


flacjacket:
Sorry for not responding earlier, the argument against this is if you are debugging it, you don't want it to silently fail. Would making it a configurable option be ok?

Azd325:
Yes, I think an configurable option will be fine. I will change my code to it.

Azd325:
So I implemented it as configurable option

flacjacket:
Great, thanks!

PR #743, opened 2015-08-15
Pulseaudio widget

azg1966:
Pulseaudio widget copy-pasted from i3pystatus


libqtile/drawer.py
```diff
@@ -105,8 +105,9 @@ def font_family(self):
 
     @font_family.setter
     def font_family(self, font):
-        d = self.fontdescription()
-        d.set_family(font)
+        # d = self.fontdescription()
+        d = pangocffi.FontDescription.from_string(font)
+        # d.set_family(font)
```
dequis: What's up with this?

dequis:
According to travis, flake8 is failing

```
./libqtile/widget/pulse.py:5:1: F403 'from ctypes import *' used; unable to detect undefined names
./libqtile/widget/pulse.py:8:33: W291 trailing whitespace
./libqtile/widget/pulse.py:306:1: E303 too many blank lines (3)
./libqtile/widget/pulse.py:333:9: E303 too many blank lines (2)
./libqtile/widget/pulse.py:339:5: E303 too many blank lines (2)
```

Also this would be way simpler if it used cffi instead of autogenerated ctypes definitions.

libqtile/drawer.py
azg1966: d.set_family(font) do not work then bold font like 'Times Bold' - groupbox widget stays with default font.

flacjacket:
There is currently the problem that not having pulse audio installed will keep qtile from starting, since it will throw an OSError when it tries to open the library, and safe_import only checks for ImportError's.

I'd also definitely like to have cffi definitions rather than ctypes if you can do that.

ramnes:
Why do we even want to talk directly to pulseaudio via ctypes or cffi when there's a plethora of binaries that can give us all the informations needed? I don't get it. Also, we might want to extend our Volume widget with PA stuff rather than having two different widgets.

ramnes:
Going ahead and closing this, it's two years old and doesn't seem to have much support. If anyone wants this in, feel free to re-open a merge request. Thanks!

PR #744, opened 2015-08-16
Prevent UnicodeEncodeError in the __repr__ of windows in python2

dequis:
These would appear in other places, such as the focus function of layouts, which indirectly call `__repr__` somehow.


flacjacket:
Thanks!

PR #748, opened 2015-08-19
Wlan pythonwifi module change to iwlib - python3 and python2 support

farebord:
Sorry, trying to create another pull I have close the other.

### DIFF
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -25,7 +25,7 @@
 
 from . import base
 import logging
-from pythonwifi.iwlibs import Wireless, Iwstats
+import iwlib
 
 
 class Wlan(base.InLoopPollText):
@@ -42,13 +42,12 @@ class Wlan(base.InLoopPollText):
         self.add_defaults(Wlan.defaults)
 
     def poll(self):
-        interface = Wireless(self.interface)
+        interface = iwlib.get_iwconfig(self.interface)
         try:
-            stats = Iwstats(self.interface)
-            quality = stats.qual.quality
-            essid = interface.getEssid()
+            quality = interface['stats']['quality']
+            essid = bytes(interface['ESSID']).decode()
             return "{} {}/70".format(essid, quality)
-        except IOError:
+        except OSError:
             logging.getLogger('qtile').error('%s: Probably your wlan device '
                     'is switched off or otherwise not present in your system.',
-                    self.__class__.__name__)
+                        self.__class__.__name__)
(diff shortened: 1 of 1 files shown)


flacjacket:
I'm fine with this, I'm not sure if we want to break it for current users, tho. What do you think about trying to use iwlib and having a fallback to pythonwifi?

farebord:
@flacjacket I've tried a little bit, it works, exceptly when changing the essid. It updates the quality changes, but not the essid changes (only change the essid if you restart qtile). 
Like I said in the previous pull request (which I accidently close by creating new branchs) the problem with pythonwifi module is that it not ported to python3, at least, when I tried it 1 week ago it wasn't. I tried to use 2to3 but it has a lot of things that can't port. iwlib works right out of the box on py2 and py3.

flacjacket:
I'm saying for people still using Python 2, they may want to stay with pythonwifi, especially if there are some cases where iwlib doesn't work as well. I'm thinking first import pythonwifi in a `try`/`except ImportError` block, with the except being to import iwlib.

farebord:
@flacjacket I add it. It should work now.

libqtile/widget/wlan.py
```diff
@@ -42,13 +44,24 @@ def __init__(self, **config):
...
+            try:
+                stats = Iwstats(self.interface)
+                quality = stats.qual.quality
+                essid = interface.getEssid()
+                return "{} {}/70".format(essid,quality)
+            except IOError:
+                logging.getLogger('qtile').error('%s: Probably your wlan device '
+                'is switched off or otherwise not present in your system.',
+                self.__class__.__name__)
+        except:
```
flacjacket: I think you should define functions that return quality and essid, rather that putting it in a bare try/except. That is, define a function, such as `_get_wifi_stats`, in the `try`/`except ImportError` blocks at the top of the file, then call that here.

libqtile/widget/wlan.py
```diff
@@ -42,13 +44,24 @@ def __init__(self, **config):
...
-            return "{} {}/70".format(essid, quality)
-        except IOError:
-            logging.getLogger('qtile').error('%s: Probably your wlan device '
-                    'is switched off or otherwise not present in your system.',
+            interface = Wireless(self.interface)
+            try:
+                stats = Iwstats(self.interface)
+                quality = stats.qual.quality
+                essid = interface.getEssid()
+                return "{} {}/70".format(essid,quality)
```
flacjacket: Watch the pep8 check here, the Travis tests are failing because you need a space after the comma

Mic92:
see https://github.com/farebord/qtile/pull/1

flacjacket:
@Mic92 Thanks, that looks good, I can merge off of your branch. Just a heads up, you should put `# -*- coding: utf-8 -*-` at the top of files where you put your name, otherwise Python 2 fails.

PR #758, opened 2015-09-16
Add command to switch to layout with given index.

StephenBarnes:
When you have lots of layouts it's easier to have a shortcut to switch to each than to cycle through all of them, especially when you have laggy windows.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -120,7 +120,7 @@ class _Group(command.CommandObject):
         raise ValueError("No such layout: %s" % layout)
 
     def toLayoutIndex(self, index):
-        assert 0 <= index <= len(self.layouts), "layout index out of bounds"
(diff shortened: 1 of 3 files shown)


libqtile/group.py
```diff
@@ -119,21 +119,20 @@ def layout(self, layout):
                 return
         raise ValueError("No such layout: %s" % layout)
 
-    def nextLayout(self):
+    def toLayoutIndex(self, index):
+        assert 0 <= index <= len(self.layouts), "layout index out of bounds"
```
flacjacket: Should be strictly less than `len(self.layouts)`

StephenBarnes: Nice catch, fixed in next commit.

flacjacket:
Great, thanks, looks good!

PR #761, opened 2015-09-21
rework the focus model a bit

### LINKED ISSUE #760: `cursor_warp = True` not warping on some configs
`cursor_warp = True` seems to conflict with something else in some configs.

That config warps correctly: https://github.com/ramnes/qtile-config/blob/master/config.py
But that one doesn't: https://gist.github.com/anonymous/7b1c52b46d60c9e894c7

Strange thing being:
> rogerduran: hallyn: can you try, eval("self.config.cursor_warp"), from qsh?
> hallyn: > eval("self.config.cursor_warp")
> hallyn: (True, 'True')
> rogerduran: at least is set

Some more IRC copy-pastah:
> tych0: hallyn: i'm thinking the real patch should look something like this, https://github.com/tych0/qtile/commit/21ae6d9765d4624f7b19734a31a8f235eaafd5fb

Looked like Tych0 was going to make a PR in the end.

tych0:
Several issues here:

1. most layouts were passing warp=False in a lot of places, so people who
   used cursor_warp effectively couldn't use those layouts. Instead, let's
   only pass warp=False where it makes sense

2. toScreen defaulted warp=True, which happened on EnterNotify, so the
   mouse would move around if you were switching screens with it, which was
   annoying.

Instead, let's make warp optional, and explicitly pass False when we know
we don't want to warp. This patch leaves max (and similar layouts) as they
were, not ever moving the pointer on behalf of the user, since is no real
"movement" when switching windows in these layouts.

Closes #760

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -195,13 +195,17 @@ class _Group(command.CommandObject):
         for i in self.windows:
             i._resetMask()
 
-    def focus(self, win, warp):
+    def focus(self, win, warp=True):
         """
             if win is in the group, blur any windows and call
             ``focus`` on the layout (in case it wants to track
             anything), fire focus_change hook and invoke layoutAll.
 
-            warp - warp pointer to win
+            warp - warp pointer to win. This should basically always be True,
+            unless the focus event is coming from something like EnterNotify,
+            where the user is actively using the mouse or on full screen
+            layouts where only one window is "maximized" at a time, and it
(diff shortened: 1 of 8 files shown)


hallyn:
Seems to do all the right things for me

flacjacket:
Oops, this was merged and I thought it was a ffwd, but it wasn't so this wasn't closed.

PR #762, opened 2015-09-23
don't crash when loading history

tych0:
For whatever reason, my qtile history couldn't be loaded across restarts,
which causes this code to fail, which then causes me to lose my session.
Instead, let's just log an error (which, unfortuantely only goes to stderr
because this likely happens in the user's config before qtile's logging is
initialized) and continue on.

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -36,6 +36,7 @@ import glob
 import os
 import pickle
 import string
+import logging
 from collections import deque
 
 from . import base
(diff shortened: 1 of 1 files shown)


libqtile/widget/prompt.py
```diff
@@ -374,7 +375,15 @@ def __init__(self, name="prompt", **config):
             self.history_path = os.path.expanduser('~/.qtile_history')
             if os.path.exists(self.history_path):
                 with open(self.history_path, 'rb') as f:
-                    self.history = pickle.load(f)
+                    try:
+                            self.history = pickle.load(f)
```
flacjacket: This line is over indented, I'm surprised flake8 didn't flag this.

tych0: On Sat, Sep 26, 2015 at 08:34:37AM -0700, Sean Vig wrote:
> > @@ -374,7 +375,15 @@ def __init__(self, name="prompt", **config):
> >              self.history_path = os.path.expanduser('~/.qtile_history')
> >              if os.path.exists(self.history_path):
> >                  with open(self.history_path, 'rb') as f:
> > -                    self.history = pickle.load(f)
> > +                    try:
> > +                            self.history = pickle.load(f)
> 
> This line is over indented, I'm surprised flake8 didn't flag this.

Derp, my bad. Pushed, a fix, thanks.

flacjacket:
Forgot to edit the commit message to close this, but just merged

PR #763, opened 2015-09-26
Catch setproperty windowerrors

tych0:


flacjacket:
Nice, thanks for the patch

Issue #764, opened 2015-09-26
restart no longer remembers group


tych0:
I just restarted on develop, and it didn't remember the group I was using.

flacjacket:
I've noticed this too, I wasn't sure if it was just me. I've been considering rolling back some of the restart serialization code until it's in a better state to see if that fixes some of the more recent problems.

ramnes:
Closing in favor of #960, that has more background and a workaround.

PR #736, opened 2015-08-07
Recursive serialization + documentation  

aniruddhkanojia:

PR #711, opened 2015-07-20
Fixes slice serialization

aniruddhkanojia:

PR #765, opened 2015-09-26
RFC: Revert layout serialization

flacjacket:
While having layout serialization is nice, the current implementation has been a bit buggy and introduced some bugs of its own on restarts (potentially #764, in addition to unreported issues noted on IRC). Some of the coding practice in this always worried me a bit and it may be good to roll this code back until it can be properly vetted and implemented. Let me know what you guys think about this, ping @tych0 @Roger @dequis (and I think @ramnes was also reporting having restart problems on IRC).


dequis:
Is this really all of it? What about stuff like `test/test_restart.py` or commits like 6f64d135f63d16c51a6fa30a8bd7eb15fea1a171 (with a very vague commit message and a very suspicious looking fix).

These diffs are too noisy... it looked somewhat ok in the pull request diffs but didn't realize until now that it meant a bunch of small commits doing weird looking stuff and then reverting it.

flacjacket:
Relevant PRs #672, #711, #725, #736, #740. I didn't really realize how much stuff was touched, I'll go and try to clear it out.

flacjacket:
Things I've left in:

From manager.py: `Qtile.cmd_get_state`, `Qtile.cmd_get_info`

test/test_serialize.py: This test isn't the greatest, it didn't catch all serialization problems, but I cleaned this up a bit from `test/test_restart.py`, and it's a decent starting point to fixing the serialization problems that the PR's "fixed"

flacjacket:
Actually, test/test_serialize relies on things that should be taken out, so I'm pulling that, too.

flacjacket:
I know where dx stands on this, does anyone else have an opinion?

tych0:
On Thu, Oct 01, 2015 at 02:44:25PM -0700, Sean Vig wrote:
> I know where dx stands on this, does anyone else have an opinion?

I abstain, since I dropped the ball as well :)

Roger:
well, i guess it's the best to revert this until we have a proper implementation.. actually, i'm suffering this issues too

flacjacket:
Reverted.

PR #768, opened 2015-09-29
MonadTall margin fix

xplv:
When MonadTall is used with a single window, the margin on the righte side and the bottom is too big. This should fix it.


flacjacket:
This fix looks ok, but would you be able to add some tests to `test/layouts/test_xmonad.py` with a margin, so this doesn't break in the future?

flacjacket:
Thanks! I just merged this in

PR #777, opened 2015-10-06
add memory widget

Mic92:


libqtile/widget/memory.py
```diff
@@ -0,0 +1,32 @@
+import re
```
flacjacket: Please add a copyright notice at the top of this file (see the other files for an example)

libqtile/widget/memory.py
```diff
@@ -0,0 +1,32 @@
...
+from . import base
+
+
+def get_meminfo():
+    val = {}
+    with open('/proc/meminfo') as file:
+        for line in file:
+            key, tail = line.split(':')
+            uv = tail.split()
+            val[key] = int(int(uv[0]) / 1000)
```
flacjacket: You can do `int(uv[0]) // 1000` if you do a `from __future__ import division` at the top of the file

libqtile/widget/memory.py
```diff
@@ -0,0 +1,32 @@
...
+    val['MemUsed'] = val['MemTotal'] - val['MemFree']
+    return val
+
+
+class Memory(base.InLoopPollText):
+    """Displays a memory usage graph."""
```
flacjacket: It looks like this displays text, not a graph.

libqtile/widget/memory.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, **config):
+        super(Memory, self).__init__(**config)
+        self.add_defaults(Memory.defaults)
+
+    def poll(self):
+        mem = get_meminfo()
+
+        def replace(m):
+            return str(mem.get(m.group(1), "(nil)"))
+        return re.sub("%{([^}]+)}", replace, self.format)
```
flacjacket: What exactly is this doing? Is this not something that you can do with `.format()` more easily? (cf. https://docs.python.org/3/library/string.html#format-examples)

Mic92:
Add proposed changes. Squashed everything into one commit.

flacjacket:
Great, thanks!

PR #779, opened 2015-10-07
Pin requirements with pip-tools

Azd325:
Splits production and development packages in seperate requirements files.

In requirements.txt includes only the packages what are used for a
production enviroment. In requirements-dev.txt are production and development
dependencies unpinned to test on travis for the latest packages if something
breaks.

This commit adds also pip-tools as development tool to pin and sync the
requirements with the local enviroment.

Updates Makefile to add a new command to compile requirements.in to pin
the latest packages in the requirements.txt


ramnes:
:+1:

.travis.yml
```diff
@@ -48,8 +48,8 @@ before_install:
 
 install:
 # Installl requirements
-  - pip install xcffib
```
flacjacket: We should keep this here to be sure xcffib is always installed before cairocffi.

flacjacket:
I agree that it is good to have a pinned set of packages that we know for sure everything works with, but it would be nice to always test against current releases without having to manually keep the requirements up-to-date.

flacjacket:
I've been thinking about this, and here's my proposal:

Have a separate `requirements` and `requirements-dev`, as you've outlined. The `requirements` will be pinned runtime dependencies (as you've implemented) and will follow the release of the master branch. That way there is a stable release with known dependencies that we know work (I know people mentioned this previously when the cffi breakage was happening). The `requirements-dev` will be what requirements is now, and follow the most recent releases of other packages to make sure we keep qtile working with the most up-to-date libraries.

Azd325:
I like your propsal. I change the code to it.

.travis.yml
Azd325: I reverted this change.

flacjacket:
I like this, thanks!

PR #782, opened 2015-10-10
create new windows in the parent's group

### LINKED ISSUE #775: A new window should pop on the group where it was launched, not on the current one
Right now, if I'm in group A, open something a bit heavy to load (i.e. Chromium) and switch to group B, the window will appear in group B, while I would expect it in group A.

alexf91:
This partly fixes #775 (the case where a process forks and creates a new window).
Handling execution from the prompt needs a different approach.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -213,10 +213,16 @@ def get_cache_dir():
 try:
     import psutil
     def parent_pid(pid):
-        return psutil.Process(pid).ppid()
+        try:
+            return psutil.Process(pid).ppid()
+        except:
+            return None
 
 except ImportError:
     import subprocess
     def parent_pid(pid):
         cmd = 'ps -o ppid= -p {}'.format(pid)
-        return int(subprocess.check_output(cmd.split()))
+        try:
+            return int(subprocess.check_output(cmd.split()))
+        except:
+            return None
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -208,3 +208,15 @@ def get_cache_dir():
     if not os.path.exists(cache_directory):
         os.makedirs(cache_directory)
     return cache_directory
+
+
+try:
+    import psutil
+    def parent_pid(pid):
+        return psutil.Process(pid).ppid()
+
+except ImportError:
+    import subprocess
+    def parent_pid(pid):
+        cmd = 'ps -o ppid= -p {}'.format(pid)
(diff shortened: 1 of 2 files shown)


flacjacket:
This all looks good. I'll try installing this and seeing if it works ok and report back on how it works for me.

libqtile/utils.py
```diff
@@ -208,3 +208,21 @@ def get_cache_dir():
...
+            return None
+
+except ImportError:
+    import subprocess
+    def parent_pid(pid):
+        cmd = 'ps -o ppid= -p {}'.format(pid)
+        try:
+            return int(subprocess.check_output(cmd.split()))
+        except:
+            return None
```
Azd325: What yout think to make the header a bit clearer? Like this :wink:

```python
try:
    import psutil

    def pid_method():
        return psutil.Process(pid).ppid()

except ImportError:
    import subprocess

    def pid_method():
        cmd = 'ps -o ppid= -p {}'.format(pid)
        return int(subprocess.check_output(cmd.split()))


def parent_pid(pid):
    try:
        return pid_method()
    except:  # Is there any good exception to catch?
        return None

```

Azd325: Otherwise should we not add psutil to the requirements then we don't need to catch the `ImportError`?

flacjacket:
Since I've been running this I haven't noticed any vast improvements, lots of dialogs still pop up in the wrong screen or on the wrong group. Do you have a particular case that this has fixed a positioning bug?

alexf91:
It only fixes cases where a new process is started from a terminal and the group is switched immediately after starting. If it takes some time until the window appears, it's currently created on the current group. With this, the window is created in the group where the terminal is located.

This doesn't work for some terminals like gnome-terminal and xfce4-terminal, which run a server process from which the clients are forked from. It also won't work if a process has more than one window.

It basically just handles this case:
* start xterm in group 1
* execute sleep 5 && xterm
* switch group
* window now pops up in group 1 instead of the current group

emdete:
how about programs started from the qtile prompt?

what's about programs with a long startup phase which show a splash screen, it would be wonderful if the main screen appears in the group where the spash screen was shown (..i started the program).

i think the parent-pid aproach alone doesn't really implement the desired behaviour.

alexf91:
Ok, I used this for quite some time now and it doesn't really change anything or makes things even worse than before in some cases.
I'm taking it back for now.

PR #783, opened 2015-10-13
Add 'line' highlighting method for GroupBox widget

flix477:
Adds a new highlighting method called **line** for the GroupBox widget. Here are some screenshots:

gradients
material_dark

The thickness of the line is configurable with the **line_thickness** variable, its color is customizable with **this_current_screen_border** and the color of the block on top of the line is also modifiable with **highlight_color**.

### DIFF
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -132,7 +132,7 @@ class TextLayout(object):
         self.drawer.ctx.show_layout(self.layout)
 
     def framed(self, border_width, border_color, pad_x, pad_y,
-               highlight_color, line_thickness):
+               highlight_color=None, line_thickness=None):
         return TextFrame(self, border_width, border_color, pad_x, pad_y,
                          highlight_color, line_thickness)
 
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -131,16 +131,21 @@ class TextLayout(object):
         self.drawer.ctx.move_to(x, y)
         self.drawer.ctx.show_layout(self.layout)
 
-    def framed(self, border_width, border_color, pad_x, pad_y):
-        return TextFrame(self, border_width, border_color, pad_x, pad_y)
+    def framed(self, border_width, border_color, pad_x, pad_y,
+               highlight_color, line_thickness):
+        return TextFrame(self, border_width, border_color, pad_x, pad_y,
+                         highlight_color, line_thickness)
 
 
 class TextFrame(object):
-    def __init__(self, layout, border_width, border_color, pad_x, pad_y):
+    def __init__(self, layout, border_width, border_color, pad_x, pad_y,
+                 highlight_color, line_thickness):
         self.layout = layout
         self.border_width = border_width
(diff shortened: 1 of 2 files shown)


libqtile/drawer.py
```diff
@@ -162,7 +168,27 @@ def draw(self, x, y, rounded=True, fill=False):
...
+                self.layout.width + self.pad_left + self.pad_right,
+                bar_height,
+                self.border_width
+            ]
+            self.drawer.set_source_rgb(self.highlight_color)
+            self.drawer.fillrect(*highlight_opts)
+
+            lineopts = [
+                x,
+                bar_height - self.line_thickness,
```
flacjacket: You should probably check that `bar_height` isn't `None`

libqtile/widget/groupbox.py
```diff
@@ -150,7 +155,7 @@ class GroupBox(_GroupBase):
         (
             "highlight_method",
             "border",
-            "Method of highlighting (one of 'border', 'block' or 'text') "
+            "Method of highlighting ('border', 'block', 'text' or 'line')"
```
flacjacket: Comma after `'text'`

libqtile/drawer.py
```diff
@@ -162,7 +168,27 @@ def draw(self, x, y, rounded=True, fill=False):
             self.layout.height + self.pad_top + self.pad_bottom,
```
flacjacket: This is probably a good fallback if `bar_height` is `None`

libqtile/drawer.py
flix477: Sorry if it looks like a basic question, but what should I do if `bar_height` is `None`? Should I raise an exception?

flacjacket: I'd say probably use the height of the frame, i.e. `self.layout.height + self.pad_top + self.pad_bottom`

flix477:
Can someone with a multi-monitor setup confirm that it works correctly? I don't have one so I can't test it.

flacjacket:
Tried running this today, you'll need to make the new arguments have default values, or go through the widgets and be sure all the relevant calls are changed. For example, I get:
```
  File "/usr/lib64/python3.4/asyncio/events.py", line 120, in _run
    self._callback(*self._args)
  File "/home/sean/qtile/qtile/bin/libqtile/manager.py", line 1156, in f
    func(*args)
  File "/home/sean/qtile/qtile/bin/libqtile/bar.py", line 288, in _actual_draw
    i.draw()
  File "/home/sean/qtile/qtile/bin/libqtile/widget/tasklist.py", line 261, in draw
    icon=self.get_window_icon(w),
  File "/home/sean/qtile/qtile/bin/libqtile/widget/tasklist.py", line 147, in drawbox
    self.padding_y
TypeError: framed() missing 2 required positional arguments: 'highlight_color' and 'line_thickness'
```

flix477:
Good catch! My latest commit should fix it.

flacjacket:
Great, thanks, this looks cool, merging!

PR #789, opened 2015-10-20
Isort

Azd325:
I propose this change to improve the code quality and standard over all. :)

### DIFF
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -20,13 +20,6 @@
 
 from __future__ import division
 
-try:
-    import tracemalloc
-except ImportError:
-    tracemalloc = None
-
-from libqtile.dgroups import DGroups
-from xcffib.xproto import EventMask, WindowError, AccessError, DrawableError
 import io
 import logging
 import os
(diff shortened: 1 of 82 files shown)


flacjacket:
I'm not sure what exactly this adds to readability or code quality. Yes, it alphabetizes the imports, but I'm not sure that's necessary, imports may be better in logical order rather than lexicographic, for example. Some of the places it adds and deletes whitespace and moves things around in ways I can't understand looking at the result, so even when imports look fine to me, I'll need to run isort to check if I got it right. Even PEP8 is more a guideline than a hard rule for code quality, and our enforcement of it is extremely tough. I've made plenty of comments telling people they have to add some spaces here and there even though PEP8 is the standard, I'm not enthusiastic about having to tell people to go back and rearrange imports.

Azd325:
You're right! There is no PEP for the ordering, but its best practice. To have a order like Python Standard Library, Third Party, Current Python Project,... The spaces are there because the imports got sorted by section, what I described above. I thought about it to and I think the linting can be overkill but I would keep isort in the development dependencies to do it on a regular base like before a release.

ramnes:
I'm used to work with `isort`, so I can only support that PR. I really makes the imports more readable and consistent, which is neat.

Unless there are some special imports I don't know that need to be on top of a file for monkey patching or stuff like that, everything looks good to me.

libqtile/manager.py
```diff
@@ -34,24 +27,30 @@
...
-from . import hook
-from . import utils
-from . import window
-from . import xcbq
+
+try:
+    import tracemalloc
+except ImportError:
+    tracemalloc = None
+
```
ramnes: Too much blank lines here.

Roger:
this reminds me.. https://www.youtube.com/watch?v=wf-BqAjZb8M

ramnes:
@flacjacket maybe we could merge or close this one? I can take the rebase on myself if needed.

flacjacket:
@ramnes I'm still +0 on this, but if the consensus is to add this test in, I'd be happy to merge it.

ramnes:
How do we know what the consensus is? :smiley:

tych0:
IMO we should just close this.

ramnes:
:disappointed:

PR #791, opened 2015-10-21
fix #790 and add line urgent method

### LINKED ISSUE #790: GroupBox 'line' highlighting issue with multimonitor
So running multiple monitors, `highlight_color` does not get a value for the other screen's label (I think), and the log will start dumping this:
```
  File "/usr/lib64/python3.4/asyncio/events.py", line 120, in _run
    self._callback(*self._args)
  File "/usr/lib64/python3.4/site-packages/libqtile/manager.py", line 1156, in f
    func(*args)
  File "/usr/lib64/python3.4/site-packages/libqtile/bar.py", line 288, in _actual_draw
    i.draw()
  File "/usr/lib64/python3.4/site-packages/libqtile/widget/groupbox.py", line 318, in draw
    highlight_color,
UnboundLocalError: local variable 'highlight_color' referenced before assignment
```
@flix477 if you want to take a crack at this, you're welcome to, otherwise I can try to fix it.

flix477:
This fixes the problem encountered when using the `line` highlight method in a multimonitor environment. It also adds `line` as a urgent highlighting method.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -168,7 +168,7 @@ class GroupBox(_GroupBase):
             "urgent_alert_method",
             "border",
             "Method for alerting you of WM urgent "
-            "hints (one of 'border', 'text' or 'block')"
+            "hints (one of 'border', 'text', 'block', or 'line')"
         ),
         (
             "disable_drag",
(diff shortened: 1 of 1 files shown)


flacjacket:
I've been meaning to get around to checking that this works, I will have
some time tomorrow, I'll let you know then how this works for me.

flacjacket:
So I was playing with this trying to get it to work, and was having problems where all the non-highlighted groups would have a black background. In retrospect, the `highlight_color = self.bar.background` should have been `highlight_color = self.background or self.bar.background`, but I played around with this to get it to work. I also dropped `line_thickness` and re-used `borderwidth` to keep us from running up too many options, and then made most of the keyword arguments explicit rather than positional. You can check out my recommendations here [1], let me know what you think of them.

[1] https://github.com/flacjacket/qtile/commit/c00130766165ab19f88480c633165c99bfec7e1d

flix477:
This is how it should've been from the beginning! Nice work!

PR #793, opened 2015-10-27
moc.py widget eats cpu

zordsdavini:
For me with this widget cpu usage is increased. So, I added configurable parameter to manage this

### DIFF
--- a/libqtile/widget/moc.py
+++ b/libqtile/widget/moc.py
@@ -20,7 +20,7 @@ import os
 import subprocess
 
 
-class Moc(base.InLoopPollText):
+class Moc(base.ThreadPoolText):
 
     """A simple MOC widget.
 
@@ -41,7 +41,7 @@ class Moc(base.InLoopPollText):
     ]
 
     def __init__(self, **config):
-        base.InLoopPollText.__init__(self, **config)
+        base.ThreadPoolText.__init__(self, "", **config)
         self.add_defaults(Moc.defaults)
         self.status = ""
         self.local = None
--- a/libqtile/widget/moc.py
+++ b/libqtile/widget/moc.py
@@ -16,12 +16,11 @@
 
 from . import base
 
-import time
 import os
 import subprocess
 
 
-class Moc(base.ThreadPoolText):
+class Moc(base.InLoopPollText):
 
     """A simple MOC widget.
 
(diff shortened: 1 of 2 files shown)


flacjacket:
I think the real problem here is that `ThreadedPollText` doesn't use the `update_interval` option introduced in the parent class `InLoopPollText`. I think the right way to do this is to fix `timer_setup` in `ThreadedPollText` to use `update_interval` properly rather than fixing it on the subclasses.

zordsdavini:
I didn't know about InLoopPollText. I remaked to use it

libqtile/widget/moc.py
```diff
@@ -20,7 +20,7 @@
 import subprocess
 
 
-class Moc(base.ThreadPoolText):
+class Moc(base.InLoopPollText):
```
flacjacket: You should probably stay with `ThreadPoolText` (which now that I see it, it is a typo, actually should be `TheradPollText`). The in-loop one should be used for things that are _fast_, since they run in the main event loop, the threaded version is good for things like this.

Rather, you should change the `ThreadPollText` code in `libqtile/widget/base.py` to use the `update_interval` to re-schedule the poll method.

libqtile/widget/moc.py
```diff
@@ -36,11 +36,12 @@ class Moc(base.ThreadPoolText):
     defaults = [
         ('play_color', '00ff00', 'Text colour when playing.'),
         ('noplay_color', 'cecece', 'Text colour when not playing.'),
-        ('max_chars', 0, 'Maximum number of characters to display in widget.')
+        ('max_chars', 0, 'Maximum number of characters to display in widget.'),
+        ('update_interval', 0.5, 'Update Time in seconds.'),
```
flacjacket: Once you've made the changes I suggested, this should then be the only change you need for this widget

zordsdavini:
@flacjacket now I don't understand what should I do. Should I add update_interval functionality to ThreadPollText and rewrite moc.py to work with it?

flacjacket:
Yes,  change the `ThreadPollText` base class to have a delay in it, then when you add the `update_interval` to the moc widget (to change from the default 10 min update time) it will work as expected.

zordsdavini:
added timer to ThreadPoolText class. Default set to None as it is now. 10 min is too much for some widgets, so it is better to add defaults in widgets

zordsdavini:
p.s. @flacjacket : should be Pool or Poll? (I'm not English native speaker). I see variation her.

flacjacket:
Alright, I can look at this later today. It should be pool, see dc8eee936c42fca73fd07bc400ccf70f119c0eb1

flacjacket:
This looks good, I squashed the commits and pushed them in, thanks!

zordsdavini:
@flacjacket should be POOL everywhere? I can create a patch to fix it

flacjacket:
No, it is correct as it is, the InLoopPoll is for polling in the main eventloop, and ThreadPool is for the thread pool executor.

PR #795, opened 2015-11-02
Adding support for CPU graph in FreeBSD

dinapappor:
This commit adds support for CPU graph in FreeBSD.

For this to work in FreeBSD linprocfs needs to be mounted.

Since linprocfs uses an older model of the kernel not all
fields are available, hence removing some of the fields
for FreeBSD.

In FreeBSD /proc stuff is located in
/compat/linux/proc , but it lacks some of the features
of a modern linux kernel. Hence the platform.system(checks).

This should maintain compatibility with Linux, but please test
it so I don't break anything.

Cheers!

### DIFF
--- a/libqtile/widget/graph.py
+++ b/libqtile/widget/graph.py
@@ -200,7 +200,10 @@ class CPUGraph(_Graph):
         self.oldvalues = self._getvalues()
 
     def _getvalues(self):
-        with open('/proc/stat') as file:
+        proc = '/proc/stat'
+        if platform.system() == 'FreeBSD':
+            proc = '/compat/linux'+proc
+        with open(proc) as file:
             lines = file.readlines()
 
             # default to all cores (first line)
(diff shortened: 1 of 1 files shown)


libqtile/widget/graph.py
```diff
@@ -200,7 +200,10 @@ def __init__(self, **config):
         self.oldvalues = self._getvalues()
 
     def _getvalues(self):
-        with open('/proc/stat') as file:
+        proc = '/proc/stat'
+        if platform.system() == 'FreeBSD':
```
Azd325: You forget to import platform on the top. `import platform`.  This was the reason that the test fails.

Azd325: Here the output from travis
```
flake8 --config=./test/flake8.cfg ./libqtile bin/qtile* bin/qsh
./libqtile/widget/graph.py:204:12: F821 undefined name 'platform'
./libqtile/widget/graph.py:205:35: E226 missing whitespace around arithmetic operator
./libqtile/widget/graph.py:221:16: F821 undefined name 'platform'
```

libqtile/widget/graph.py
```diff
@@ -200,7 +200,10 @@ def __init__(self, **config):
...
+            proc = '/compat/linux'+proc
```
Azd325: And here is whitespace missing around the arithmetic operator. :)

dinapappor:
Oooh, forgot the import. Let me fix that. :)

PR #796, opened 2015-11-02
Adding support for CPU graph in FreeBSD

dinapappor:
This commit adds support for CPU graph in FreeBSD.

For this to work in FreeBSD linprocfs needs to be mounted.

Since linprocfs uses an older model of the kernel not all
fields are available, hence removing some of the fields
for FreeBSD.

In FreeBSD /proc stuff is located in
/compat/linux/proc , but it lacks some of the features
of a modern linux kernel. Hence the platform.system(checks).

This should maintain compatibility with Linux, but please test
it so I don't break anything.

Cheers!

### DIFF
--- a/libqtile/widget/graph.py
+++ b/libqtile/widget/graph.py
@@ -222,7 +222,7 @@ class CPUGraph(_Graph):
             if platform.system() == 'FreeBSD':
                 name, user, nice, sys, idle = line.split(None, 4)
             else:
-	        name, user, nice, sys, idle, iowait, tail = line.split(None, 6)
+                name, user, nice, sys, idle, iowait, tail = line.split(None, 6)
 
             return (int(user), int(nice), int(sys), int(idle))
 
@@ -440,4 +440,3 @@ class HDDBusyGraph(_Graph):
 
     def update_graph(self):
         self.push(self._getActivity())
-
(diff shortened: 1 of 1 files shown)


libqtile/widget/graph.py
```diff
@@ -214,8 +218,10 @@ def _getvalues(self):
 
                 if not line.startswith("cpu%s" % self.core):
                     raise ValueError("No such core: %s" % self.core)
-
-            name, user, nice, sys, idle, iowait, tail = line.split(None, 6)
+            if platform.system() == 'FreeBSD':
+                name, user, nice, sys, idle = line.split(None, 6)
```
flacjacket: You should probably change this to `line.split(None, 4)`

dinapappor:
This also as a side-effect fixes swap graph in FreeBSD.

libqtile/widget/graph.py
```diff
@@ -237,12 +244,19 @@ def update_graph(self):
...
         for line in file:
-            key, tail = line.split(':')
-            uv = tail.split()
-            val[key] = int(uv[0])
+            if line.lstrip().startswith("total"):
+                pass
+            else:
+                key, tail = line.strip().split(':')
+                uv = tail.split()
+                val[key] = int(uv[0]) // 1000
```
flacjacket: It looks like you added `// 1000`, is that a BSD specific, or is that also fixing some error?

dinapappor: No, I stupidly copied some code from memory.py. Let me fix that and re-push.

flacjacket:
This looks good, thanks! I squashed the commits to clean up the history and merged this in. Let us know if you hit any other bugs running qtile on BSD.

PR #797, opened 2015-11-04
add configurable text if updates are available

Mic92:


flacjacket:
This looks ok, but it looks like there are some problems flagged by the
pep8 checks

flacjacket:
Just merged this in, thanks!

PR #799, opened 2015-11-07
Fix make check command

Azd325:
'make check' don't find tests. This was introuced in
728fafdc937b56319d24d87bb21bf2f254dda92a

The problem is that the 'libqtile' was the argument of --cov.
That one restricted the coverage output. After this parameter was
removed nosetests was thinking it should search in the folder libqtile for
tests.

I remove 'libqtile' as argument and now it runs smoothly again.
Also, I expand the command line parameter to see what they
actually mean and the 'make check-cov' call has now an echo output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 default:
 	@echo "'make check'" for tests
+	@echo "'make check-cov'" for tests with converage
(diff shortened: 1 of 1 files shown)


flacjacket:
Oops! Thanks for the fix!

PR #802, opened 2015-11-11
Fix build of docs on python3

Azd325:
mock has to be pinned to version 0.8 because otherwise there is a
RuntimeError (Maximum recursion depth) with in newer versions

### DIFF
--- a/docs/BUILDING
+++ b/docs/BUILDING
@@ -1,15 +1,10 @@
-NOTE that the doc build scripts require Python 2, and will *not* work with
-Python 3!!!
-
-===============================================================================
-
-To build in a virtual environment (requires the 'virtualenv' and 'pip' tools
+To build in a virtual environment (requires the 'virtualenv', 'pip' and 'graphiz' tools
 installed in your distro):
 
 $ cd ./docs
-$ virtualenv2 qtile
+$ virtualenv qtile
 $ source qtile/bin/activate
-$ pip2 install -r requirements.txt
+$ pip install -r requirements.txt
 $ make html
 $ deactivate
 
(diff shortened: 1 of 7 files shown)


flacjacket:
Thanks for finding this and fixing this. There are a few problems with the current implementation that I don't think this addresses w.r.t. Python 3, I think the better fix is flacjacket@7344da39eb78124b682a1366ffb267053d525a61, let me know what you think.

@dmpayton ary you in charge of building the docs? is there any way to test these changes before we start messing with them?

Azd325:
Looks good to me.
I test it very simple, I did a diff between the old and new output folder. So I was sure that we didn't messed up. :)
For this purpose I included also the coverage extension to see undocumented files.

FYI: I started for fun a branch where I use tox for testing, there it would be easy to add a new environment to build the docs separately form the other tests.

flacjacket:
Great, thanks for the updates! Merging!

I think dmpayton had to do a bit of maneuvering to get everything working on RTD, I'm not sure if my changes make that easier or harder.

Testing doc building would definitely be a good idea, I've seen it done on Travis before, too. It might even encourage us to update the docs :)

dmpayton:
It looks like a dependency issue is causing the build to fail on RTD. I haven't had a chance to look too deeply into this. but here's the build output if anyone wants to take a look.

http://dpaste.com/2EE46T5

Azd325:
Seems that the new Mock Modules require the pangocffi module
from travis.yml
```bash
# build pangocffi module
  - python -c 'import cffi, sys; sys.exit(cffi.__version_info__[0])' || python libqtile/ffi_build.py
```

Can we add somehow the call this file(libqtile/ffi_build.py) from our config because I think we can not run   so easy python or bash scripts on RTD?

flacjacket:
@Azd325 that in the RTD log was from a known and obnoxious problem with installing cairocffi and xcffib at the same time via another dependency (read: qtile). The problem was fixed by the mocked modules added in this PR, as then qtile, and thus also xcffib/cairocffi, didn't need to be installed on RTD.

Azd325:
Yes, you are right. I found a different way to fix thisin #805

PR #805, opened 2015-11-13
Fixed RTD build

Azd325:
The docs didn't build because the order of the packages was wrong
Related to #802 (qtile#802 (comment))

I test it on my own rtd instance and I got it to run if the setting is disabled
`python setup.py install` for this case we have to load the general project
requirements file inside the docs requirements file

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,5 @@
+-r ../requirements.in
+
 mock # for python <= 3.2
 sphinx
 sphinx_rtd_theme


Azd325:
download 1

@dmpayton these are my settings how i got to install on my rtd instance

flacjacket:
I don't think this is necessary, I am able to get the docs build to pass without this as long as the "install project" box is disabled and the requirements file is set to `docs/requirements.txt`

Azd325:
Cool, good catch. I thought where I saw this traceback that it need these packages, but I rebuilt it like you described and it worked for me, too. So I will  close this PR.
The only thing what is left, I would move the build process to python3 to support it. :)

flacjacket:
Already changed over to Python 3 :)

It'll be the next release before the "latest" docs are updated, but the "develop" docs are building great on Python 3 now.

Issue #650, opened 2015-03-27
Click on KeyboardLayout widget changes layout only once


VFedyk:
Click on widget call this code (through several methods):

```python
def _set_keyboard(self, keyboard):
        command = ['setxkbmap']
        command.extend(keyboard.split(" "))
        try:
            subprocess.check_call(command)
...
```
So if we initialized KeyboardLayout widget with some 2 layouts (['us', 'ua'] in my case) we will switch to second layout. But __get_keyboard__ method uses __xset -q__ to determine current layout. There will be something like this:

```
Keyboard Control:
  auto repeat:  on    key click percent:  0    LED mask:  00000002
  XKB indicators:
    00: Caps Lock:   off    01: Num Lock:    on     02: Scroll Lock: off
    03: Compose:     off    04: Kana:        off    05: Sleep:       off
    06: Suspend:     off    07: Mute:        off    08: Misc:        off
    09: Mail:        off    10: Charging:    off    11: Shift Lock:  off
    12: Group 2:     off    13: Mouse Keys:  off
  auto repeat delay:  260    repeat rate:  25
  auto repeating keys:  00ffffffdffffbbf
                        fadfffefffedffff
                        9fffffffffffffff
                        fff7ffffffffffff
  bell percent:  50    bell pitch:  400    bell duration:  100
Pointer Control:
  acceleration:  2/1    threshold:  4
Screen Saver:
  prefer blanking:  yes    allow exposures:  yes
  timeout:  600    cycle:  600
Colors:
  default colormap:  0x22    BlackPixel:  0x0    WhitePixel:  0xffffff
Font Path:
  /usr/share/fonts/misc/,/usr/share/fonts/TTF/,/usr/share/fonts/OTF/,/usr/share/fonts/Type1/,built-ins
DPMS (Energy Star):
  Standby: 600    Suspend: 600    Off: 600
  DPMS is Enabled
  Monitor is On
```

So widget thinks (because of xset output) that current layout is first, but second is turned on. And it always tries to switch to second layout.

Can we change layout directly in Python with no use of external tools (via xcb binding or something like that)? If you give me right direction for learning more about that I can try to fix that.

tych0:
Yes, I think we can query it. Here's the output from `xtrace xset -q`:
http://paste.ubuntu.com/10716447/

[...]

PR #807, opened 2015-11-14
Improove multiple keyboard layout support

Arherion:
Hi!

Now, keyboardlayout widget determinates current layout from the setxkbmap verbose output, so  it possible to use 3+ layouts, if needed.
Also:
Fixes [#650](https://github.com/qtile/qtile/issues/650).


flacjacket:
I just squashed your commits and pushed this in, thanks for the PR!

PR #811, opened 2015-11-18
Tox for tests

Azd325:
I did this PR to get some practice with TOX. :)

This PR exclude some logic from travis. So it would be possible to run the test also on other CI systems or it can run locally.

* In https://github.com/qtile/qtile/pull/802 we mention already to build docs on travis. So I include these in this change. 
* PEP8 is now running on python 2 and 3.
* check for packaging with help from https://github.com/mgedmin/check-manifest
* more explicit in the MANIFEST.in (Should be recheck)


flacjacket:
Is it possible to keep the nightly test build with tox?

tox.ini
```diff
@@ -0,0 +1,67 @@
...
+
+    # https://github.com/travis-ci/travis-ci/issues/4866
+    py32: coverage<4.0.0
+commands =
+    # xcffib has to be installed before cairo
+    pip install xcffib
+    # Install unpinned requirements from requirements.in
+    pip install -r {toxinidir}/requirements.in
+    # build pangocffi module
+    bash -c "python -c 'import cffi, sys; sys.exit(cffi.__version_info__[0])' || python {toxinidir}/libqtile/ffi_build.py"
```
flacjacket: It's not great to have to do this, especially since tox will install qtile and this should be run then. With cairocffi, a way was found to run from the tox installed package, https://github.com/SimonSapin/cairocffi/blob/master/tox.ini

requirements-dev.txt
```diff
@@ -1,11 +1,3 @@
-# Development requirements
 flake8
 nose-cov
-python-coveralls
```
flacjacket: Why remove this and the other deps?

tox.ini
```diff
@@ -0,0 +1,67 @@
...
+# Install either trollius or tulip for 3.3
+# Asyncio is in the standard library for 3.4+
+deps =
+    nose-cov
+
+    trollius: trollius
+    tulip: asyncio
+
+    # https://github.com/travis-ci/travis-ci/issues/4866
+    py32: coverage<4.0.0
```
flacjacket: I don't think we've run into any problems running python 3.2 with coverage...

requirements-dev.txt
Azd325: I remove python-coveralls because is only used by travis It is directly used now in the travis.yml
The other ones I removed because they already exists in requirements.txt/in

tox.ini
Azd325: Yes, we do, https://travis-ci.org/Azd325/qtile/jobs/92030151
```
py32-trollius runtests: commands[3] | nosetests --verbose --detailed-errors --with-cov --cover-package libqtile
  File ".tox/py32-trollius/bin/nosetests", line 11, in <module>
    sys.exit(run_exit())
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose/core.py", line 121, in __init__
    **extra_args)
  File "/opt/python/3.2.5/lib/python3.2/unittest/main.py", line 123, in __init__
    self.parseArgs(argv)
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose/core.py", line 145, in parseArgs
    self.config.configure(argv, doc=self.usage())
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose/config.py", line 347, in configure
    self.plugins.begin()
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose/plugins/manager.py", line 99, in __call__
    return self.call(*arg, **kw)
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose/plugins/manager.py", line 167, in simple
    result = meth(*arg, **kw)
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/nose_cov.py", line 59, in begin
    import cov_core
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/cov_core.py", line 5, in <module>
    import coverage
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/coverage/__init__.py", line 13, in <module>
    from coverage.control import Coverage, process_startup
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/coverage/control.py", line 15, in <module>
    from coverage.annotate import AnnotateReporter
  File "/home/travis/build/Azd325/qtile/.tox/py32-trollius/lib/python3.2/site-packages/coverage/annotate.py", line 85
    dest.write(u'  ')
                   ^
SyntaxError: invalid syntax
```

tox.ini
Azd325: Tox don't install qtile because I skip it in line 3 with `skipsdist=True`. Otherwise skipsdist would fail

Azd325: Here is a traceback if `skipdist=True` is removed https://travis-ci.org/Azd325/qtile/jobs/92035029#L458

Azd325:
Yes, I think it is possible to add add the nightly version back but first I have to figure out how I add `allow_failures` back to the matrix. That it get accepted from travis.

tox.ini
Azd325: If you have any ideas about it let me know :)

flacjacket: Gah, that stupid bug! We (unfortunately the royal we) really need to figure out how to get that to work...

This is fine for now then.

requirements-dev.txt
flacjacket: I see, that sounds good

tox.ini
flacjacket: Interesting, I don't think we've hit that before with 3.2 on standard travis builds, but it looks like it never installs coverage...

https://travis-ci.org/qtile/qtile/jobs/91530365#L219

Azd325: My guess is that installed directly in travis and it is catch already. If have look more lines down traivs is installing version 3.7

tox.ini
Azd325: It seems that pip install in a order like ASC if it is in a requirements file. It should be more like this in our case. (http://stackoverflow.com/a/26046664/548558)

```
import pip
with open("requirements.txt") as f:
    for line in f:
        pip.main(['install','U', line])
```

Azd325:
Nighly is now added to the build list and is set as `allow_failures` in the travis.yml

flacjacket:
Gah, why does that stupid pypy test keep failing?

.travis.yml
```diff
@@ -1,36 +1,45 @@
...
+          env: TOXENV=docs
+        - python: 3.5
+          env: TOXENV=py3-pep8
+        - python: 2.7
+          env: TOXENV=py2-pep8
+
+    allow_failures:
+        - python: nightly
+          env: TOXENV=py-nightly
+        - python: pypy
```
flacjacket: Yeah, I guess we need this for now, but it is a very consistent bug, and seemed to always be in the same test, but it's hard to say if the bug is in qtile, the tests, or from running on 12.04.

.travis.yml
```diff
@@ -38,23 +47,22 @@ cache:
...
-# build pangocffi module
-  - python -c 'import cffi, sys; sys.exit(cffi.__version_info__[0])' || python libqtile/ffi_build.py
+  - pip install tox python-coveralls
+
+script:
+  - tox
+
 
 after_success:
   - coveralls
```
flacjacket: We shouldn't run coveralls on docs or lint builds

flacjacket: Actually, it's not reporting coverage information at all

https://coveralls.io/builds/4327157

tox.ini
```diff
@@ -0,0 +1,68 @@
...
+commands =
+    # xcffib has to be installed before cairo
+    pip install xcffib
+    # Install unpinned requirements from requirements.in
+    pip install -r {toxinidir}/requirements.in
+    # build pangocffi module
+    bash -c "python -c 'import cffi, sys; sys.exit(cffi.__version_info__[0])' || python {toxinidir}/libqtile/ffi_build.py"
+    nosetests --verbose --detailed-errors --with-cov --cover-package libqtile
```
flacjacket: It looks like tox is just running `nosetests` without all the extra flags:

https://travis-ci.org/qtile/qtile/jobs/94389293#L371

That is causing the coverage reporting problems

flacjacket:
Alright, we don't really use coverage much as an important metric, I think this testing is good, so I'll merge this in and we can figure out coverage later.

PR #812, opened 2015-11-20
Gc widget 2015

rxcomm:
updated gc widget to work with google_api_python_client 1.4.2.


libqtile/widget/google_calendar.py
```diff
@@ -81,11 +95,6 @@ class GoogleCalendar(base.ThreadedPollText):
     defaults = [
         ('calendar', 'primary', 'calendar to use'),
         (
-            'storage_file',
```
flacjacket: It may be good to leave this option in, since it is not dictated by the API

libqtile/widget/google_calendar.py
```diff
@@ -106,32 +115,32 @@ def __init__(self, **config):
...
+        flow.user_agent = APPLICATION_NAME
 
-        # if the credentials don't exist or are invalid, get new ones from FLOW
-        # FLOW must be run in a different thread or it blocks qtile
-        # when it tries to pop the authentication web page
         def get_from_flow(creds, storage):
-            self.credentials = run(FLOW, storage)
+            if flags:
+                self.credentials = tools.run_flow(flow, storage, flags)
+            else: # Needed only for compatibility with Python 2.6
```
flacjacket: You need 2 spaces here before the comment to satisfy the linter

Azd325: If I'm right, we don't support more  2.6 anymore, so we don't need more this lines

libqtile/widget/google_calendar.py
```diff
@@ -62,14 +65,25 @@
...
+import os
 
-from apiclient.discovery import build
-from oauth2client.client import OAuth2WebServerFlow
-from oauth2client.tools import run
-import oauth2client.file
+from apiclient import discovery
+import oauth2client
+from oauth2client import client
+from oauth2client import tools
```
Azd325: Can we merge both lines together like `from oauth2client import client, tools`
Otherwise nice work :)

flacjacket:
Thanks for this! I just cherry-picked the last 3 commits, which look like they have all the relevant changes, and squashed those on top of develop

PR #816, opened 2015-11-21
Allow subprocess.Popen-style argument list in cmd_spawn

encukou:
This is more convenient when an argument might contain
spaces, since the caller can avoid having to escape that
arg (which would be un-escaped by shlex.split straight away).


flacjacket:
Cool, thanks!

PR #818, opened 2015-11-30
Reuse layout_util's asserts in test_verticaltile.py

dsalisbury:
This is just cleaning up some duplicate code I came across while writing some tests for a new layout.


flacjacket:
Thanks! I'll push this in

PR #821, opened 2015-12-02
Check for None before falling back.

### LINKED ISSUE #820: Params with fallbacks (e.g. padding_y) are ignored if they're set to 0
`configurable.ExtraFallback.hidden_attribute()` uses the fallback attribute if the configured value is falsey.

This means that passing `padding_y=0` to widgets like `widget.GroupBox()` has no effect.

mjkillough:
Previously settings like padding_y=0 would be ignored, as the 0
is falsey. Explicitly check for None instead.

This change adds some tests for libqtile.configurable. Only one
of the tests (test_dont_use_fallback_if_set) checks that this
fixes the problem.

Fixes #820


flacjacket:
Great, thanks! If only all our changes were so well tested.

PR #823, opened 2015-12-03
implement wmii layout

hallyn:
Each group starts with one column.  The first window takes up the whole screen.
Next window splits the column in half.  Windows can be moved to the column to
the left or right.  If there is no column in the direction being moved into, a
new column is created.

Each column can be either stacked (each window takes up the whole vertical real
estate) or split (the windows are split equally vertically in the column)
Columns can be grown horizontally (cmd_grow_left/right).  Growing up or down could
be implemented but isn't.  (We're not currently tracking the info needed for that,
and I haven't wanted it)

Stacked columns are blue (dark blue unfocused, bright blue focused) and split
ones are red-ish (dark unfocused, bright focused).  I'm open to better ways
to quickly make clear what columns are what (without taking up much real estate)

Signed-off-by: Serge Hallyn <serge.hallyn@ubuntu.com>

### DIFF
--- a/libqtile/layout/wmii.py
+++ b/libqtile/layout/wmii.py
@@ -1,0 +1,463 @@
+# Copyright (c) 2013 Mattias Svala
+# Copyright (c) 2013 Tao Sauvage
+# Copyright (c) 2014 ramnes
+# Copyright (c) 2014 Sean Vig
+# Copyright (c) 2014 dmpayton
+# Copyright (c) 2014 dequis
+# Copyright (c) 2014 Tycho Andersen
+# Copyright (c) 2015 Serge Hallyn
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


libqtile/layout/wmii.py
```diff
@@ -0,0 +1,458 @@
...
+#        lazy.layout.toggle_split()
+#    ),
+
+# We have an array of columns.  Each columns is a dict containing
+# width (in percent), rows (an array of rows), # and mode, which is
+# either 'stack' or 'split'
+#
+# Each row is an array of clients
+
+def swap(array, a, b):
```
flacjacket: A nice little Python trick:
```
array[a], array[b] = array[b], array[a]
```
works for this

libqtile/layout/wmii.py
```diff
@@ -0,0 +1,458 @@
...
+        self.columns = [{'active': 0, 'width': 100, 'mode': 'split', 'rows': []}]
+
+    def info(self):
+        d = Layout.info(self)
+        d["current_window"] = self.current_window
+        d["clients"] = [x.name for x in self.clients]
+        return d
+
+    def add_column(self, prepend, win):
+        newwidth = (int)(100 / (len(self.columns) + 1))
```
flacjacket: `int` is a function, not a cast, so this can be `int(100 / (len(self.columns) + 1))`

flacjacket:
Just tested this out, this looks great, thanks!

PR #826, opened 2015-12-06
panel offset setting

neutrak:
When configuring qtile I was unable to create an offset to the tiling area.  My configuration includes conky, which I'd like to be always visible.  I tried changing the y value and height, but that did not seem to fix the problem.  I therefore added a new setting called panel_offset, which allows the user to keep a section of the desktop area visible without being covered up by tiles.  

If there is a better way to do this please let me know!  I am new to this code base and I could easily have missed something.  If not then please merge in this change so that others who wish to use utilities such as conky can configure this in config.py.  Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,7 @@ Features
 Current Release
 ===============
 
-The current stable version of qtile is 0.10.3, released 2015-12-25. See the
+The current stable version of qtile is 0.10.4, released 2016-01-19. See the
 `documentation <http://docs.qtile.org/en/latest/manual/install/index.html>`_
 for installation instructions.
 
--- a/bin/libqtile
+++ b/bin/libqtile
@@ -1,1 +1,0 @@
-../libqtile
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,6 @@
-qtile (0.10.3-1) unstable; urgency=low
+qtile (0.10.4-1) unstable; urgency=low
 
   * Initial release.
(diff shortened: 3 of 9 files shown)


flacjacket:
Maybe it'd be better to modify Bar to reserve an extra area that it doesn't draw to, then you can position the area with all of the 4 possible Bar locations.

neutrak:
I agree that having the ability to position the area on any side is helpful, but tying it to Bar means that you can't use an area that's not adjacent to the bar.  If you wanted the Bar at the top and say a docky-style launcher at the bottom, you couldn't do that if the area were tied to the Bar.  I think for flexibility it should be separate.

ramnes:
I'll go ahead and close this. Feel free to ping me if you think `Gap` isn't sufficient, and we'll reopen. Thanks!

neutrak:
I completely agree.  

Please let me know when a build is released that has that issue (#1230) fixed.

ramnes:
It is not a release yet, but please see [the PR I've just opened](https://github.com/qtile/qtile/pull/1239). Some tests do not pass for obscure reasons but it's still functional, so that would be awesome that you test it and report back if it fits use case!

m-col:
This is pretty old but in case this is still wanted, it is now possible to pass `margin` as a parameter to `Bar`s. For the described behaviour you'd want a margin on the South edge of the bar, so you'd use something like this:
```python
Bar([...], margin=[0 0 20 0])
```

PR #828, opened 2015-12-09
Update core dependencies

Azd325:


flacjacket:
I guess you initially introduced this distinction between pinned and development dependencies, I figured we'd update them with new releases, tho I guess there's no reason not to keep them up-to-date other than the hassle of updating the file.

flacjacket:
I just updated the dependencies with the most recent version bump, so this update isn't needed, thanks!

PR #833, opened 2015-12-22
Support of an empty battery and unchanging battery status

Kiwhacks:


flacjacket:
I feel like doing a check to see if the AC is plugged in is better, AC power and UNKNOWN status means Full, no AC power and UNKNOWN status means it's a secondary battery. Then it doesn't have an arbitrary cutoff of 50% for the battery to show up. Tho that would introduce an additional thing that needs to be checked. What do you think?

PR #834, opened 2015-12-24
debian: update packaging

tych0:
* add debian WNPP bug number
* bump versions for dependencies
* bump version number for release (this may need to be corrected)
* bump debian standards version (lintian)
* get rid of separate licensing for debian/ (lintian)
* add extra dependencies for widgets (still missing are google calendar's deps,
  iwlib, and mpd in python3 which aren't in the debian archives; if users
  install these manually they should work)

Note that there is still one (set of) lintian warning(s, for all the images),
which would require code changes to fix:

W: python-qtile: image-file-in-usr-lib usr/lib/python2.7/dist-packages/libqtile/resources/battery-icons/battery-good.png

I've also bumped the version number to 0.10.3, but that may not be the next
version of qtile that is tagged. We should bump this version number to whatever
that next version is when it is tagged. (And it would be handy to tag soon, so
I can start the process of getting a debian upload :)

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,6 @@
-qtile (0.9.1-1) unstable; urgency=low
+qtile (0.10.3-1) unstable; urgency=low
 
   * Initial release.
+  * Closes: #762637
 
- -- Tycho Andersen <tycho@tycho.ws>  Tue, 20 Jan 2015 18:10:02 +1300
+ -- Tycho Andersen <tycho@tycho.ws>  Thu, 24 Dec 2015 04:03:11 +0000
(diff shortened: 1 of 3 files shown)


flacjacket:
I just pushed this and tagged 0.10.3, thanks! What tools do you use to update the debian package? I can try to keep it up-to-date with the most recent version when I tag releases.

tych0:
On Fri, Dec 25, 2015 at 1:19 PM, Sean Vig <notifications@github.com> wrote:

> I just pushed this and tagged 0.10.3, thanks! What tools do you use to
> update the debian package? I can try to keep it up-to-date with the most
> recent version when I tag releases.
>

Basically just "dch -i" and then edit the changelog. For the actual control
file (and other stuff), there isn't really any tooling :(. I have a debian
container that I keep around to build it in periodically, though, and I
don't mind maintaining that bit, so if you are going to tag a release, let
me know a day or so ahead and I can update the packaging.

PR #835, opened 2015-12-29
Comply with XDG Base Directory specification

kynikos:
I think by default the file paths used by Qtile should respect the [XDG Base Directory specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

The specification was only partially followed, and this series of commits tries to complete the implementation.

The optional dependency on PyXDG has been made a requirement, so packagers should be informed and accordingly update their scripts.

As noted in `CHANGELOG`, some files' default paths have changed. Users may be instructed to move `.qtile_history` to `~/.cache/qtile/prompt_history` in order to preserve the history of commands typed in the prompt. They may also be advised to delete the old `~/.qtile.log` file and refer to `~/.local/share/qtile/qtile.log`, or define `log_path` in their `config.py`.

The GoogleCalendar widget seems to be still using a `~/.credentials` folder, but I don't use it and I haven't tried to update it because I wouldn't be able to test it.

Last thing, I'm not sure if `%changelog` in `rpm/qtile.spec` should be updated too.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@ qtile x.x.x, released xxxx-xx-xx:
         - the default log file path changed from ~/.qtile.log to
           ~/.local/share/qtile/qtile.log
         - the cache directory changed from ~/.cache to ~/.cache/qtile
+        - the prompt widget's history file changed from ~/.qtile_history to
+          ~/.cache/qtile/prompt_history
     * features
         - wlan widget shows when you are disconnected and uses a configurable
           format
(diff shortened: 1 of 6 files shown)


flacjacket:
So I've been sitting on this one, and I think I wouldn't mind moving default file locations around, but I feel like having to leverage PyXDG is overkill for as few files as we generate. What do you think about just changing default file locations without adding the dependency?

kynikos:
Thanks for reviewing this patch! No problems with doing without pyxdg of course, I've modified, rebased and repushed the branch, please see if you like it now.
I've also added a small commit to update the docs on the testing dependencies, I didn't want to open an entire PR only for that.

flacjacket:
Great, thanks! Just pushed this batch in!

PR #836, opened 2015-12-30
delete argument

himaaaatti:
I deleted argument of __init__ for dealing same way with other layouts __init__.

And slice layout have default size and width value, so I am thinking these don't need.


flacjacket:
I just pushed this in, thanks for the fix!

PR #837, opened 2015-12-31
More debian fixes

tych0:
Here's some more stuff that some extra linter options brings up (as well as fixing a major issue w.r.t. Architecture: any vs. Architecture: all).


flacjacket:
Great, thanks for the updates!

Issue #841, opened 2016-01-05
some thoughts around logging


emdete:
- the logline should not be formatted like in

```python
 self.qtile.log.info("Unknown window property: %s" % name)
```

 instead it is much better to write

```python
self.qtile.log.info("Unknown window property: %s" % name)
```
 and leave formatting up to the logging library (it will decide if that level will be logged at all before it does the formatting and save resources).

 - i would use loggers per module instead of a single logger for all, each module should contain a

```python
from logging import getLogger
logger = getLogger(__name__)
```

 which makes clear where the logline comes from (and the linenumbers are usefull now).

 - the config should set the level on the root logger, currently this:

```python
log = getLogger(logger)
log.setLevel(log_level)
```

only sets it for the logger `logger` (if foreign modules would use logging this would get lost). this would be better:

```python
log = getLogger()
log.setLevel(log_level)
```

 - i would not doublicate logging into file + stdout but switch by config to either way

 - color logging should be optional dis/enabled by config

 - the log member of qtile should be removed or deprecated

emdete:
see https://github.com/qtile/qtile/pull/842

ramnes:
> the logline should not be formatted like in
>  self.qtile.log.info("Unknown window property: %s" % name)
> instead it is much better to write
> self.qtile.log.info("Unknown window property: %s" % name)

wat?

flacjacket:
I think the last one was supposed to be `self.qtile.log.info("Unknown window property: %s", name)`

ramnes:
> i would use loggers per module instead of a single logger for all, each module should contain a
> from logging import getLogger
> logger = getLogger(__name__)

I really don't like this. Why would we duplicate code rather than having a single properly instanciated object, which is less prone to error and definitely more DRY?

tych0:
On Thu, Jan 07, 2016 at 11:07:50AM -0800, Guillaume Gelin wrote:
> > i would use loggers per module instead of a single logger for all, each module should contain a
> > from logging import getLogger
> > logger = getLogger(__name__)
> 
> I really don't like this. Why would we duplicate code rather than having a single properly instanciated object, which is less prone to error and definitely more DRY?

Doesn't this way prefix each log message with the module it came from?
That seems useful (although marginally, we could grep for error
messages too which are probably mostly unique).

Overall, I like the idea of not passing a logging object around as an
attribute, but instead having a "global" one that is has the same name
in each file, so you can just write logger.error("foo"), instead of
[...]

PR #842, opened 2016-01-06
Logging

emdete:
implement some of the ideas of https://github.com/qtile/qtile/issues/841


flacjacket:
Other than those couple things, you need to rebase this on top of the current develop branch, as it won't apply cleanly.

emdete:
about INFO: this was an error as i prefere INFO, still i assume that most of the messages should go to DEBUG and INFO should be the default level.

about rebase: sorry, i did not pull before, this is now fixed.

about configurability: i will take a look to properly solve this. for now command line is the only option, the other stuff probably never worked anyway. are you OK with this?

emdete:
i struggled a bit over the declaration of the coding of the sources which is set to utf-8 in only some sources (mainly to support the names in the copyright area i assume) and the setting for vim to follow PEP8. because i like the vim settings (i struggled with 4-spaces in some of my commits) i added those to all sources as first line (because i found that in some sources this way) while i know this setting in the last line. anyway i seem to have destroyed the coding settings which i now reverted and put the coding: into the first line, vim in the second. i hope travis and the developers like it this way.

flacjacket:
I'll take one last look through this.

I don't really mind the vim tabstop declarations. I just have that in my vimrc, but it is the convention we use, so I'd say it's ok.

libqtile/window.py
```diff
@@ -1,3 +1,5 @@
+# vim: tabstop=4 shiftwidth=4 expandtab
+# vim: tabstop=4 shiftwidth=4 expandtab
```
flacjacket: Line repeated

libqtile/xcbq.py
```diff
@@ -1,3 +1,5 @@
+# vim: tabstop=4 shiftwidth=4 expandtab
+# vim: tabstop=4 shiftwidth=4 expandtab
```
flacjacket: Also here

libqtile/xcursors.py
```diff
@@ -1,4 +1,7 @@
+# vim: tabstop=4 shiftwidth=4 expandtab
+# vim: tabstop=4 shiftwidth=4 expandtab
```
flacjacket: And here

docs/conf.py
```diff
@@ -1,3 +1,4 @@
+# vim: tabstop=4 shiftwidth=4 expandtab
```
flacjacket: Put below utf-8 line

emdete:
i hope you give it another look when i have a solution for configurable logging 😉

i will develop something to allow reconfiguration from the config.py but had no time till now. pure command line configuration would be easier but not desired as far as i understood, right?

flacjacket:
Oh, I'll definitely keep giving feedback, I was just thinking I would take another look at it and be able to merge it :) There are the couple things I mentioned that I found that should be fixed, but otherwise I'd be happy to merge this as is. The log configuration is likely broken right now I don't mind getting rid of it for now. Command line options would be ok for a start, but you don't always have easy access to command line option when starting a window manager through a login manager, so access to configuration through configuration files is nice to add.

bin/qtile
```diff
@@ -56,13 +59,13 @@ def make_qtile():
     from argparse import ArgumentParser
     parser = ArgumentParser(
         description='A full-featured, pure-Python tiling window manager.',
-        prog='qtile')
-
+        prog='qtile',
+        )
```
ramnes: Why changing the indentation and adding an useless comma?

docs/sphinx_qtile.py
```diff
@@ -1,3 +1,4 @@
+# vim: tabstop=4 shiftwidth=4 expandtab
```
ramnes: Please configure your editor rather than doing that.

docs/conf.py
ramnes: Yes, or just configure your editor rather than doing that.

bin/qtile
emdete: i use that pattern to be able to easier

 - sort parameters

 - add more parameters or comment them

without struggling over missing commas or brackets.

Roger:
about the vim config line, we can use something like this on the root of the repo: http://editorconfig.org

bin/qtile
ramnes: That's ugly.

flacjacket:
From #841, you can drop the `getLogger()` lines for either importing a single logger or using the root logger. If you can drop the vim lines and whitespace changes, this will be good.

PR #843, opened 2016-01-06
widgets: log threaded polling exceptions

tych0:
this conflicts with the other logging branch; i'll be happy to rebase once that one is in, or if this goes in first then someone else has to rebase ;)


flacjacket:
Well this still applies cleanly :) but needs to be updated to use `logger` rather than `self.qtile.log`

tych0:
Right you are, I've just rebased it.

PR #846, opened 2016-01-09
[battery] Fix: Can not show battery's percentage if current_now file not exists

lisongmin:
Some non standard laptop may lost `power_now` and `current_now`
file after unplug the power, And it will come back while plug in.
That may be a mistake in bios, but we should solve the case in my opinion.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -49,6 +49,13 @@ BATTERY_INFO_FILES = {
     'status_file': ['status'],
 }
 
+BATTERY_DEFAULT_INFO = {
+    'status_file': UNKNOWN,
+    'energy_now_file': '0',
+    'energy_full_file': '0',
+    'power_now_file': '0',
+}
+
 
 def default_icon_path():
     # default icons are in libqtile/resources/battery-icons
@@ -126,11 +133,11 @@ class _Battery(base._TextBox):
                     self.filenames[name] = file
                     return value
 
-        # If we made it this far, we don't have a valid file.
-        # Set it to None to avoid trying the next time.
-        self.filenames[name] = None
-
-        return None
+        # Do not remove the files for some non standard laptop
+        # may lost power_now and current_now file after unplug the power,
+        # And it will come back while plug in.
+        # self.filenames[name] = None
+        return BATTERY_DEFAULT_INFO.get(name)
 
     def _get_info(self):
         try:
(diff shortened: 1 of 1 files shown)


lisongmin:
Why it test failed under python3.3 ? It seems not related to this commit.
Should I rebase to master but not develop?
```
======================================================================

FAIL: test.layouts.test_slice.test_command_propagation

----------------------------------------------------------------------

  File "/home/travis/build/qtile/qtile/.tox/py33-tulip/lib/python3.3/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/qtile/qtile/test/utils.py", line 117, in wrapped_fun

    self.startQtile(self.config)

  File "/home/travis/build/qtile/qtile/test/utils.py", line 219, in startQtile

    self._waitForQtile(rpipe)

  File "/home/travis/build/qtile/qtile/test/utils.py", line 256, in _waitForQtile

    raise AssertionError("Error launching Qtile, socket never came up")

nose.proxy.AssertionError: Error launching Qtile, socket never came up

-------------------- >> begin captured stdout << ---------------------

------------------------ >> begin log file << ------------------------

2016-01-09 02:42:44,323 WARNING init_log:107 Starting Qtile

------------------------- >> end log file << -------------------------

Qtile exited with exitcode: -15

--------------------- >> end captured stdout << ----------------------

>>  raise AssertionError("Error launching Qtile, socket never came up")
```

tych0:
On Fri, Jan 08, 2016 at 07:32:31PM -0800, lsm wrote:
> Why it test failed under python3.3 ? It seems not related to this commit.

You're right, travis is slow sometimes and our timeouts aren't long
enough. You can ignore this failure.

libqtile/widget/battery.py
```diff
@@ -205,6 +212,8 @@ def _get_text(self):
             elif info['stat'] == CHARGING:
                 char = self.charge_char
                 time = (info['full'] - info['now']) / info['power']
+            elif info['stat'] == UNKNOWN:
```
flacjacket: Why do you do this? I think usually an `UNKNOWN` status means fully charged and plugged into AC power. Wouldn't this break the formatting on the final return (since `char` isn't defined)?

lisongmin:
Close this PR for #833 is already fix this problem.

PR #850, opened 2016-01-18
Add altcoin support to btc tracker

tych0:


flacjacket:
Nice, thanks for the updates!

PR #857, opened 2016-02-01
Unquote query parameter to aptitude for CheckUpdates widget

tjarvstrand:
This check returns an empty string for me on Ubuntu 15.10 if I don't unquote the query parameter

The current command would be equal to running ```aptitude search \'~U\'``` in the console which also yields empty output.


flacjacket:
Thanks, I just pushed this up!

PR #858, opened 2016-02-01
Correctly call accessor for KeyBoardLayout widget keyboard property

tjarvstrand:


flacjacket:
Great, thanks for the fix!

PR #859, opened 2016-02-01
Bring Arch Linux installation manual up to date

mskuta:
Please review and add my patch.

### DIFF
--- a/docs/manual/install/arch.rst
+++ b/docs/manual/install/arch.rst
@@ -11,7 +11,7 @@ Using an AUR Helper
 ===================
 
 The preferred way to install Qtile is with an `AUR helper`_. For example,
-if you use `yaourt`_:
+if you use yaourt:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


flacjacket:
@Roger how does this look?

zordsdavini:
Shouldn't it be just another block of installation choice?  Pacman is default installer, so removing it is not Arch-way

mskuta:
pacman is called by makepkg to install missing dependencies. This is what the -s option does. There is no need to execute it explicitly.

Mind that Qtile is available through the Arch User Repository (AUR) and not part of the community repository, where it would be directly accessible by pacman.

docs/manual/install/arch.rst
```diff
@@ -11,7 +11,7 @@ Using an AUR Helper
 ===================
 
 The preferred way to install Qtile is with an `AUR helper`_. For example,
-if you use `yaourt`_:
+if you use yaourt:
```
kynikos: Instead of removing the link, it could be updated to https://archlinux.fr/yaourt-en

docs/manual/install/arch.rst
```diff
@@ -20,39 +20,33 @@ if you use `yaourt`_:
...
+or cloning its repository:
 
+.. code-block:: bash
 
-Installing AUR packages without helper
-======================================
+    # snapshot
+    curl -s https://aur.archlinux.org/cgit/aur.git/snapshot/<package-name>.tar.gz | tar -xvzf -
+    # or repository
+    git clone https://aur.archlinux.org/<package-name>.git
```
kynikos: This looks a bit inconsistent with the section above, which uses the actual package names instead of the `<package-name>` placeholder.

docs/manual/install/arch.rst
```diff
@@ -20,39 +20,33 @@ if you use `yaourt`_:
...
 .. code-block:: bash
 
-    tar -xvzf <packagename>-<vernum>.tar.gz
-    cd <packagename>-<vernum>
-    makepkg -s
-    sudo pacman -U <packagename>
+    cd <package-name>
+    makepkg -sri
 
 Please see the Arch Wiki for more information on installing packages from
```
kynikos: While you're there, can you please change "Arch Wiki" to "ArchWiki" (no space)? It is the most common spelling by far.

docs/manual/install/arch.rst
```diff
@@ -20,39 +20,33 @@ if you use `yaourt`_:
...
 the AUR:
-
-http://wiki.archlinux.org/index.php/AUR#Installing_packages
+https://wiki.archlinux.org/index.php/Arch_User_Repository#Installing_packages
```
kynikos: This could be turned into a pretty link like the others in the page, by using the word `AUR` before the colon as the link text.

kynikos:
I'm an ArchWiki admin and I confirm what @mskuta says: makepkg's `-s` flag already installs the needed dependencies; if using pacman, they should be installed with the `--asdeps` flag.
I've added a few comments to the diff, but overall I support this patch.

docs/manual/install/arch.rst
mskuta: Without a placeholder you have to write out four commands (release/snapshot, release/git, develop/snapshot, develop/git) which would be confusing.

It already was inconsistent before my patch. I would prefer changing the yaourt lines to use a placeholder, too.

kynikos: Indeed, it's just nitpicking of course, but using a placeholder also for yaourt would look better to me.

mskuta:
I've incorporated the good suggestions from @kynikos.

flacjacket:
Thanks for this, just pushed it in!

PR #862, opened 2016-02-04
When using layout MonadTall, open new window on main panel if previously focused window was on the main panel.

tjarvstrand:
This mimics the xmonad layout behaviour a bit more closely.

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -774,10 +774,11 @@ class MonadTall(SingleWindow):
 
     def cmd_swap_main(self):
         "Swap current window to main pane."
-        if self.align == self._left:
-            self.cmd_swap_left()
-        elif self.align == self._right:
-            self.cmd_swap_right()
+        if self.focused != 0:
+            if self.align == self._left:
+                self.cmd_swap_left()
+            elif self.align == self._right:
+                self.cmd_swap_right()
 
     def cmd_left(self):
         "Focus on the closest window to the left of the current window."
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -215,7 +215,7 @@ class MonadTall(SingleWindow):
 
     def add(self, client):
         "Add client to layout"
-        self.clients.insert(self.focused + 1, client)
+        self.clients.insert(self.focused, client)
         self.do_normalize = True
 
     def remove(self, client):


tjarvstrand:
Not really sure why the tests fail, is it related to my change?

flacjacket:
There are some tests of how xmonadtall does layouts in `./test/layouts/test_xmonad.py`, since the layout handling is changed, some of the tests are failing. You can change the tests to fit with this new layout scheme.

tjarvstrand:
Thanks, I'll do that.

I've noticed an issue where restarting qtile will now reverse the window ordering because of the order they are added to the layout. Do you have any tips on a nice way of figuring out whether we're doing initial layout or not. 

One way could be to add a boolean member variable that is toggled after the first layout is completed and based on that either insert new clients either at ```this.focused``` or ```this.focused - 1```. How would I find out when all clients have been added to the layout at startup?

flacjacket:
Hmm, that's an interesting issue that this would introduce... There had been some work on getting some nice serialization between restarts, but it was ultimately too fragile and had to be dropped, but having something like that would hopefully fix this. I'll have to think about how we could solve it with the current setup.

tjarvstrand:
I've been looking at adding a startup_complete-hook. It's not tested yet so I haven't opened a pull request, but basically the hook would fire as the last thing done in the __init__ of the manager. What do you think?

flacjacket:
You can add that change as a new PR, or on top of this PR, if you think it's working

tjarvstrand:
Yeah, I'll do it as a separate PR eventually. It seems testing of the hooks is somewhat lacking so I'm looking into that before submitting. It's taking a bit more time since I'm not that experienced with Python yet :)

tjarvstrand:
I seem to be hitting some kind of wall in my understanding of how Python object references work. @flacjacket, if you could offer any deeper understanding of why this testcase fails I'd be grateful:
https://github.com/tjarvstrand/qtile/blob/add-hook-startup-complete/test/test_hook.py#L92

From debugging I've determined:
- calling ``inc_startup_calls``` in the test does increment the ```startup_calls``` variable
- calling ```libqtile.manager.hook.fire("startup")``` in the test also increments the ```startup_calls``` variable
- When calling ```startQtile``` the startup hook fires and ```inc_startup_calls``` gets called.
- When ```inc_startup_calls``` is called as a result of ```startQtile```, then ```self``` points to the same object as ```self``` outside the method (the Xephyr object).

Yet when startQtile returns, the member variable is still 0 (or 0 again)

Very confused right now

PR #865, opened 2016-02-05
New layout: columns

heitzmann:
This layout is a mix of what I liked best from stack, wmii and xmonadtall.
I thought about extending one of them, but that felt too invasive, so I've decided to write a new one.

### DIFF
--- a/libqtile/layout/__init__.py
+++ b/libqtile/layout/__init__.py
@@ -36,3 +36,4 @@ from .zoomy import Zoomy
 from .matrix import Matrix
 from .verticaltile import VerticalTile
 from .wmii import Wmii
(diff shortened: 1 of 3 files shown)


flacjacket:
The pep8 check is complaining a bit, which is failing the tests, but I'll play with this and check it out and can fix that up if you don't get to it first. Thanks!

heitzmann:
I see... I've just installed flake8 locally and fixed those issues.
The file seems to pass the tests at least here on my machine.

flacjacket:
Cool, thanks for this!

heitzmann:
My pleasure! Thanks for qtile!

PR #866, opened 2016-02-08
PyCharm code check

ThomasWaldmann:
i just picked the easy stuff for now. but even here, it found a few real bugs.

thanks for qtile, I like it!

### DIFF
--- a/libqtile/log_utils.py
+++ b/libqtile/log_utils.py
@@ -86,7 +86,7 @@ def init_log(log_level=WARNING, log_path='~/.%s.log', log_truncate=False,
     # If we have a log path, we'll also setup a log file
     if log_path:
         try:
-            log_path = log_path % 'qtile'
+            log_path %= 'qtile'
         except TypeError:  # Happens if log_path doesn't contain formatters.
             pass
         log_path = os.path.expanduser(log_path)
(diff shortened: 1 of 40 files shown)


libqtile/log_utils.py
```diff
@@ -86,7 +86,7 @@ def init_log(log_level=WARNING, log_path='~/.%s.log', log_truncate=False,
     # If we have a log path, we'll also setup a log file
     if log_path:
         try:
-            log_path = log_path % 'qtile'
+            log_path %= 'qtile'
```
flacjacket: Ooh, that's interesting, I didn't know string formatting could be a compound assignment operator

scripts/qtb
```diff
@@ -33,8 +33,8 @@ def main():
             parser.error("Please specify widget name and argument.")
         try:
             client.widget.__getitem__(args[0]).update(" ".join(args[1:]))
-        except command.CommandError, v:
-            print >> sys.stderr, v
+        except command.CommandError as v:
+            print >>sys.stderr, v
```
flacjacket: Oh, interesting, something that got missed in the Python 3 move, I can take care of that seperately

flacjacket:
Merged these commits, thanks for the much needed clean-up!

libqtile/log_utils.py
ThomasWaldmann: Me neither. :)

PR #870, opened 2016-02-15
Add cmd_spawn to logging, when execvp failed.

himaaaatti:
Added logging. Are you ok with this log message format?

### DIFF
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -1393,7 +1393,7 @@ class Qtile(command.CommandObject):
                 try:
                     os.execvp(args[0], args)
                 except OSError as e:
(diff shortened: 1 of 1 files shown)


libqtile/manager.py
```diff
@@ -1392,8 +1392,8 @@ def cmd_spawn(self, cmd):
 
                 try:
                     os.execvp(args[0], args)
-                except OSError:
-                    pass
+                except OSError as e:
+                    logger.warn("failed spawn: \"{0}\"\n{1}".format(cmd, e))
```
flacjacket: It would probably make sense to make this `.error()` or `.exception()`

himaaaatti:
I think so too. I fixed.

flacjacket:
Great, thanks for the fix

PR #871, opened 2016-02-16
Allow override of log level in xephyr script

tjarvstrand:


scripts/xephyr
```diff
@@ -3,12 +3,13 @@
 HERE=$(dirname $(readlink -f $0))
 SCREEN_SIZE=${SCREEN_SIZE:-800x600}
 XDISPLAY=:1
+LOG_LEVEL=${LOG_LEVEL-INFO}
```
flacjacket: Does this need to be `${LOG_LEVEL:-INFO}`?

tjarvstrand: Yeah, you're right that's probably better. The way it is now works for me but I googled a bit and there is a slight difference in that ```${VAR:-VAL}``` will substitute ```VAR``` with ```VAL``` if ```VAR``` is set but null whereas ```${VAR:-VAL}``` will not. The former is probably better. You always learn something new :)

flacjacket: Alright, I wasn't sure, which is why I asked, if you change that I'd be happy to pull it in!

tjarvstrand:
Updated

flacjacket:
Great, thanks!

PR #873, opened 2016-02-19
Add hook startup complete

tjarvstrand:
I was unable to solve the problem of restarting qtile and checking that startup_once doesn't fire more than once, but at least there's a basic test.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -68,8 +68,18 @@ class Subscribe(object):
         return self._subscribe("startup_once", func)
 
     def startup(self, func):
-        """Called each time qtile is started (including the first time qtile starts)"""
+        """
+            Called each time qtile is started (including the first time qtile
+            starts).
+        """
         return self._subscribe("startup", func)
+
+    def startup_complete(self, func):
+        """
+            Called each time qtile is started (including the first time qtile
+            starts), after all resources have been initialized.
+        """
+        return self._subscribe("startup_complete", func)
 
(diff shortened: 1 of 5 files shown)


flacjacket:
What were you wanting to do with the `startup_complete` hook that you couldn't do before? I'm wondering if we can't push the `startup` hook further up, even as far as where we fire `config.main()` ([here](https://github.com/qtile/qtile/blob/develop/libqtile/manager.py#L153)). There had been discussions of getting rid of calling `config.main` and only using hooks, but we'd need to make sure things like [this](https://github.com/qtile/qtile-examples/blob/master/roger/config.py#L434) is possible, so maybe passing the manager to the startup hooks. If we do that, it might even be better to call the first one `startup_early`, since most of what startup is should be starting programs and such.

tjarvstrand:
With the changes I would like to make to the MonatTall layout, new windows will be added to the stack in a sort of reverse order. This means that when restarting qtile, the entire stack will be reversed which is a bit annoying. Thus I've added code so that the MonadTall layout will add windows to the stack according to their current location on the screen so that already existing windows will remain in the same place. Once startup is complete (all existing windows have been added), the regular layout positioning will kick in.

tjarvstrand:
Ping

flacjacket:
@tjarvstrand sorry, I haven't forgotten about this! I'll try this out tonight just to be sure it's working as expected.

flacjacket:
Looks good, thanks!

PR #875, opened 2016-02-22
This commit adds a KhalCalendar widget

rxcomm:
This widget provides the next calendar update on the
status bar using khal https://github.com/geier/khal


flacjacket:
Great, thanks for kicking this back to us!

Issue #877, opened 2016-02-22
Proposition for a new QSH command for display the "keys binding"


minitchoup:
Like this : 
```
> display_kb
KeySym            Mod             Command                                    Desc                 
---------------   -------------   ----------------------------------------   -------------------  
End               mod4            spawn('amixer -q set Master toggle')       Switch Off speakers  
F1                mod4            toscreen()                                                      
F1                mod4, shift     togroup('1')                                                    
F2                mod4            toscreen()                                                      
F2                mod4, shift     togroup('2')                                                    
F3                mod4            toscreen()                                                      
F3                mod4, shift     togroup('3')                                                    
F4                mod4            toscreen()                                                      
F4                mod4, shift     togroup('4')                                                    
F5                mod4            toscreen()                                                      
F5                mod4, shift     togroup('5')                                                    
F6                mod4            toscreen()                                                      
F6                mod4, shift     togroup('6')                                                    
F7                mod4            toscreen()                                                      
F7                mod4, shift     togroup('7')                                                    
F8                mod4            toscreen()                                                      
F8                mod4, shift     togroup('8')                                                    
F9                mod4            toscreen()                                                      
F9                mod4, shift     togroup('9')                                                    
Left              mod4, shift     decrease_ratio()                                                
Page_Down, Next   mod4            spawn('amixer -q -c 0 sset Master 5dB+')   Increase volume      
Page_Up, Prior    mod4            spawn('amixer -q -c 0 sset Master 5dB-')   Descrease volume     
Return            mod4            spawn('rxvt-unicode')                                           
Return            mod4, shift     spawn('rxvt-unicode', False)                                    
Right             mod4, shift     increase_ratio()                                                
Tab               mod4            next_layout()                                                   
b                 mod4            grow()                                                          
e                 control, mod4   spawn('gksudo emacs')                                           
e                 mod4            spawn('emacs')                                                  
e                 mod4, shift     spawn('emacs', False)                                           
f                 mod4            spawn('dolphin')                                                
i                 mod4            spawn('firefox')                                                
j                 control, mod4   shuffle_up()                                                    
j                 mod4            up()                                                            
k                 control, mod4   shuffle_down()                                                  
k                 mod4            down()                                                          
m                 mod4            spawn('thunderbird')                                            
n                 mod4, shift     client_to_next()                                                
q                 control, mod4   shutdown()                                                      
r                 control, mod4   restart()                                                       
r                 mod4            spawncmd()                                                      
[...]

PR #878, opened 2016-02-22
Add the new QSH command 'display_kb'

minitchoup:
See #877

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -48,11 +48,13 @@ class Key(object):
             - *commands: A list of lazy command objects generated with the
             command.lazy helper. If multiple Call objects are specified, they
             are run in sequence.
+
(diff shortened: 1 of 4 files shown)


libqtile/config.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, modifiers, key, *commands):
         self.modifiers = modifiers
         self.key = key
         self.commands = commands
+        self.kwds = kwds
```
flacjacket: This is why keyword only arguments are so nice...

Since you're only using the `**kwds` for `desc`, maybe you can just get that value here.

flacjacket:
This is a cool addition, I played around with it a bit and that is my only comment.

libqtile/config.py
minitchoup: You are right. I will correcting this.

PR #882, opened 2016-02-26
fix typos related to enable/disable maximize

ThomasWaldmann:


flacjacket:
I know there is a lesson here about increasing code coverage in unit tests...

PR #883, opened 2016-02-27
fix x / y / width / height confusion

ThomasWaldmann:


flacjacket:
Gah, more egregious mistakes from us. Thanks!

PR #884, opened 2016-02-27
Rearrange scripts and add IPython qsh kernel

flacjacket:
Submitting this for comments to see if this is something people would be interested in. I moved the scripts around so they are now setuptools entrypoints. I also added an IPython kernel, you can check the docs [here](https://github.com/flacjacket/qtile/blob/entrypoints/docs/manual/commands/iqsh.rst) for information on how to install it and get it running. Currently, there isn't much that it does that isn't done by qsh (tho I think completing may be better), but it has the capability to leverage more IPython machinery if we incorporate this.


ramnes:
Just seeing this, and yeah, we really needed to move scripts to `libqtile`! I actually thought about it something like two weeks ago, great to see it done. Thanks a bunch.

Nice to see more use of ABC too. We definitely need more of those to assure consistency across the whole code base.

PR #885, opened 2016-02-27
Add the new QSH command 'focus_cmd'…

minitchoup:
...  and change the default behaviour of the 'spawn' command.

1) cmd_spawn(cmd) first search if the cmd is already started, and then focus his window. Else a new cmd instance is launched.
2) cmd_spawn can be call consecutively with the same cmd. So all the windows are cyclically focused.
3) cmd_spawn(cmd, focus=False), launch a new instance of cmd.

Example : 
In QSH (>):

> spawn("emacs")  # launch a first instance of Emacs (because Emacs is not already launched)
> spawn("emacs")  # Emacs window is focused (no new instance is started)
> spawn("emacs", False)  # a new Emacs instance is launched.
> spawn("emacs")  # window of the last Emacs instance is focused
> spawn("emacs")  # window of the first Emacs instance is focused


tych0:
On Sat, Feb 27, 2016 at 10:01:40AM -0800, Minitchoup wrote:
> ...  and change the default behaviour of the 'spawn' command.
> 
> 1) cmd_spawn(cmd) first search if the cmd is already started, and then focus his window. Else a new cmd instance is launched.

Please no :). If I want to start another urxvt with this patch, I
can't. Instead, let's add some new command that does what you want (or
perhaps a flag or something), but changing the default behavior is not
a good idea.

minitchoup:
Yes. I have hesitated a long time. Your first commentary confirm my bad choice. I will restore the previous behavior. And I will propose a mean to change this default focus strategy. So, everyone can do as he wishes.

Two equivalent configurations : 

1)
```
set_auto_focus (True)
keys = [
...
     Key ([mod], "Return", lazy.spawn ("rxvt-unicode"))
     Key ([mod, "shift"], "Return", lazy.spawn ("rxvt-unicode", False))
     Key ([mod], "i", lazy.spawn ("firefox"))
     Key ([mod], "m", lazy.spawn ("Thunderbird"))
...
]
```
2) 
```
keys = [
...
     Key ([mod], "Return", lazy.spawn ("rxvt-unicode", True))
     Key ([mod, "shift"], "Return", lazy.spawn ("rxvt-unicode"))
     Key ([mod], "i", lazy.spawn ("firefox", True))
     Key ([mod], "m", lazy.spawn ("Thunderbird", True))
...
]
```

flacjacket:
The way it is currently running, `spawnMap` is never purged of windows that are closed, it would be good if this actively cleaned up this dictionary.

PR #886, opened 2016-02-29
This commit fixes a problem with missing days

rxcomm:
There was a problem where no remaining agenda items
existed in the current day, and the widget would not
gracefully handle the transition to the next day.

This commit fixes that.


flacjacket:
Thanks for the updates!

PR #891, opened 2016-03-05
Fix to many blank lines

Azd325:


flacjacket:
...oops, thanks for the fix!

PR #894, opened 2016-03-12
add backlight scroll support

pandada8:
Using xbacklight to change the backlight
The reason to using `run_in_executor` is calling `xbacklight` needs 200 - 300 ms in my machines, which just block qtile.


flacjacket:
Cool, thanks for the addition

PR #895, opened 2016-03-18
df widget: always use the integer division operator in py2 and py3

dequis:
When I started using python3 for qtile this widget started showing sizes
as floating point numbers with more digits than it would have if it
showed bytes. Given how old this widget is, the intended behavior was
clearly integer division.


flacjacket:
:+1:

PR #902, opened 2016-04-11
Btwe pr

tych0:


flacjacket:
I know dx had said he used restarting to clear out old dgroups, how would that work with this PR? Is there a good way to not re-initialize the empty dgroups?

flacjacket:
I guess I will need to ping btwe on irc

btwe:
Pong from IRC :) Please go ahead.

flacjacket:
I was wondering will this always keep dgroups alive, even when they are empty? I remember @dequis mentioning he'd restart to clear stale dgroups, but maybe we should be better at clearing those out as we go.

btwe:
Yes, empty dgroups will also be re-initialized if those still exist. Right now, qtile clears empty dgroups with a (default one second) delay after the last window left the group. 

More ideas about that:
- Empty dgroups should not exists if not on a screen. MPOV a more intuitive method would be to clear empty dgroups when they are pulled off from a screen. This would make it easier to start e.g. firefox with a ProfileManager ahead on a new dgroup.
- This feature 'remain dgroups during restart' could be made a configuration option set to false by default

dequis:
>I know dx had said he used restarting to clear out old dgroups, how would that work with this PR? 

Wait I thought we had already merged this behavior. I've been using the not-very-convenient global method `delgroup("name")` from qsh to get rid of empty sticky groups. Even with its inconvenience it's actually better than the mess that results after restarting and getting windows reordered in random groups (these days i've gotten myself in a situation where i have ~10 groups, it does get messy)

So regarding that, I'm ok with this new behavior. You could certainly do something cleverer with those, but I wouldn't worry too much. I think the main source for empty sticky groups is doing `togroup` and moving the last window of a group, creating a new group but not removing the old one. That's probably where it makes more sense to handle the issue.

libqtile/state.py
```diff
@@ -45,11 +45,11 @@ def apply(self, qtile):
         Rearrange the windows in the specified Qtile object according to this
         QtileState.
         """
-        for (group, layout) in self.groups.items():
+        for (group, layout) in self.groups:
```
dequis: Wouldn't this break when upgrading from older versions? Not sure what's the best way to handle that though.

flacjacket: I'm not sure how important it is to maintain restart-ability across versions, I'd be ok breaking that to make improvements to the way it runs.

flacjacket:
@btwe Thanks! Just merged this.

PR #908, opened 2016-05-03
Fix: Lenght is unfixed for Net widget's output

lotabout:
The length of the output of Net Widget is unfixed. For example:

`1.12` -> `__1.12` (two space, 6 characters in total)
while `12.12` -> `12.12` (5 characters in total)

unfixed length of output will cause the widget' length to change every interval.


flacjacket:
I just merged this in, thanks for the fix!

PR #911, opened 2016-05-07
Move pytest config to setup.cfg

Azd325:


flacjacket:
I think this would necessitate adding pytest-cov to hacking docs [1]. I'm not entirely sure we want running code coverage to be the default behavior, but we do always need it for unittests.

[1] https://github.com/qtile/qtile/blob/develop/docs/manual/hacking.rst

Azd325:
Right coverage should not run on every test. So I move the coverage part back to the `tox.ini`

flacjacket:
Great, thanks for the update

PR #914, opened 2016-05-13
Change the implementation of cmd_next and cmd_previous to cycle through visible windows

Etn40ff:
This patch changes the behaviour of cmd_next and cmd_previous in the columns layout to allow cycling through all the visible windows.


flacjacket:
Thanks for this change!

Etn40ff:
You are welcom

PR #916, opened 2016-05-20
Put window on their transient_for window's group

nicoe:


nicoe:
I noticed that transient windows appeared in different group then their transient_for window when switching quickly to another group.

This PR aims at fix that.

flacjacket:
Looks good, thanks!

PR #917, opened 2016-05-21
Some fixes

Azd325:


libqtile/utils.py
```diff
@@ -90,7 +90,7 @@ class lru_cache(object):
             Cache data is tracked as attributes on the object itself. There is
             therefore a separate cache for each object instance.
         """
-        def __init__(self, maxsize=128, typed=False):
+        def __init__(self, maxsize=128):
```
flacjacket: I think this commit should be left unchanged. I add the `typed` kwarg to mimic the signature of functools.lru_cache, since I moved the newer Python versions to use that instead. We may want to refactor this compatibility layer to be more like the standard library lru_cache decorator.

flacjacket:
Let us know what you think about that last commit, the rest look good to me.

Azd325:
Updated this PR

flacjacket:
Thanks for updating this, merged now!

PR #918, opened 2016-05-21
Remove trollius tests for py3k

Azd325:
Trollius is deprecated and should be not more not more tested for the
python3 branch but this keeps trollius for python2 because there is no
replacment for.

pypy3 is not more tested because is currently based on python3.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ matrix:
         - python: 2.7  # these are just to make travis's UI a bit prettier
           env: TOXENV=py27-trollius
         - python: 3.3
-          env: TOXENV=py33-trollius
-        - python: 3.3
-          env: TOXENV=py33-tulip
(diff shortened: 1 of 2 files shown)


flacjacket:
I think as long as we allow trollius for Python 3.3, we should keep testing it. I'm not opposed to changing it, especially since I'd be surprised if people that have made the move to 3 are still using 3.3, but we should also update the installation docs, and maybe also the compatibility checks in `qtile/__init__.py` to only allow trollius to be imported in Python 2.

flacjacket:
Pypy 3 is up to 3.5, so it can run against the standard library asyncio, so no need for this anymore :)

flacjacket:
I guess this isn't only pypy, but all of python 3. That said, Python 3.3 EOL is coming up, so we can drop it soon! https://www.python.org/dev/peps/pep-0398/

Issue #920, opened 2016-05-24
cvlc window do not appear


nicoe:
Hello,

I noticed that using cvlc (vlc without the shiny widgets and so on, only the video appears) instead of vlc has the effect of not displaying the window although the video is playing (as I can tell from the sound).

Moreover the window does not appear in qsh. The debug log of cvlc shows nevertheless that he created a window and even give its ID (launch cvlc witch --verbose 3 and look for the xcb string).

nicoe:
Silly me I forgot the traceback:
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/libqtile/manager.py", line 724, in _xpoll
    r = h(e)
  File "/usr/lib/python3/dist-packages/libqtile/manager.py", line 1049, in handle_MapRequest
    c = self.manage(w)
  File "/usr/lib/python3/dist-packages/libqtile/manager.py", line 601, in manage
    c = window.Window(w, self)
  File "/usr/lib/python3/dist-packages/libqtile/window.py", line 716, in __init__
    _Window.__init__(self, window, qtile)
  File "/usr/lib/python3/dist-packages/libqtile/window.py", line 168, in __init__
    self.updateHints()
  File "/usr/lib/python3/dist-packages/libqtile/window.py", line 227, in updateHints
    h = self.window.get_wm_hints()
  File "/usr/lib/python3/dist-packages/libqtile/xcbq.py", line 474, in get_wm_hints
    window_group=l[8]
IndexError: tuple index out of range
```

PR #921, opened 2016-05-24
Ensure that WindowGroupHint is set in get_wm_hints

nicoe:
This is a fix for #920.
I wonder if the other properties are always there also (I don't know anything about X). I also don't know if None is the right default value but it works for me.


flacjacket:
Thanks for figuring this out, I just merged this commit and added some guards for the other hints, just in case, but I guess we haven't gotten a complaint about this until now.

Issue #733, opened 2015-08-04
MPD widget doesn't always connect


ff2000:
It always shows the "not connected" message (Mpd off), regardless of the actual state of mpd.
Configuration should be fine, I run the defaults (localhost:6600).
I see this line in my qtile.log, which seems to be related, as I don't have it without mpd widget added:
```
2015-08-04 07:27:50,365 WARNING on_done:507 poll() returned None, not rescheduling
```
I'm on qtile:develop running python-3.4.1.

ff2000:
Hm, sorry, party my fault. Didn't look at default_config but only the documented options in __init__(). Setting reconnect to True an the mpd widget starts working.
Still there needs to be something fixed as reconnect=False is the default and that case does not work even with mpd running.

flacjacket:
Ah, yeah, that would probably be preferable. I don't use the mpd widget, so I'm not sure why the default is set to not reconnect

ff2000:
Of course I could create a pull request to set reconnect=True, but as I said I think there is another issue, as the widget shows "Mpd off" even if mpd was running while qtile was started. Any idea why that happens? Do you need more information?

mandaltapesh:
@ff2000   i want to work on this . Can you tell me which is the relevant code/file ?

nicoe:
What happens is that the MPD widget initialize ThreadPoolText with the string used when there is no connection, then poll() is called but the widget is not connected and not set to reconnect so poll() returns None and is never called again.

My fix would be to use a different string ('MPD Widget') to initialize ThreadPoolText. Otherwise I don't think there is a bug.

PR #922, opened 2016-05-24
Initialize ThreadPoolText with 'MPD Widget' to not confuse users

nicoe:
This PR aims at fixing #733


flacjacket:
Does this fix the problem? It sounds like from that issue that we need to change the return from `.poll()` to be sure it is re-scheduled.

nicoe:
I think so, the default value of reconnect is False which will indeed prevent `.poll()` to be called again.
But indeed a better default value for reconnect is True as I guess people are expecting the widget to try to reconnect.

flacjacket:
Ah, I see, I re-read that thread and now I see what you mean, I'll merge this in!

PR #923, opened 2016-05-27
(TaskList) highlight_method='text' option, toggle_minimize at curr window press

andering:
I made very simple and fast edit to add 'text' option for highlight_method. (TaskList).
I didn't create separate variables for colors (like GroupBox) but let inherited from border colors instead.
 
Anyway i think it would be nice to create some constant/standard variable set with their roles  to be used through all widgets.


andering:
I also added 'expected' behavior, when you press at active window in task-list you minimize window

andering:
Don't understand.

flacjacket:
This looks good, thanks!

Issue #924, opened 2016-05-27
TreeTab broken in 0.10.6


dvzrv:
It looks like TreeTab layout is broken in 0.10.6.
When switching to it I get no sidebar and this error in qtile.log:

```
2016-05-27 18:28:05,465 ERROR libqtile manager.py:handle_KeyPress():L910 KB command error next_layout: Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/libqtile/command.py", line 97, in call
    return (SUCCESS, cmd(*args, **kwargs))
  File "/usr/lib/python3.5/site-packages/libqtile/manager.py", line 1302, in cmd_next_layout
    group.nextLayout()
  File "/usr/lib/python3.5/site-packages/libqtile/group.py", line 136, in nextLayout
    self.toLayoutIndex((self.currentLayout + 1) % (len(self.layouts)))
  File "/usr/lib/python3.5/site-packages/libqtile/group.py", line 133, in toLayoutIndex
    self.layout.show(screen)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 475, in show
    self._resize_panel(panel)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 672, in _resize_panel
    self.draw_panel()
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 440, in draw_panel
    self._tree.draw(self, 0)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 72, in draw
    top = i.draw(layout, top, level)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 236, in draw
    offsety=top + layout.section_top
TypeError: draw() got an unexpected keyword argument 'offsetx'
```

Later many more of these follow:

```
2016-05-27 18:30:24,863 ERROR libqtile hook.py:fire():L362 Error in hook focus_change
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/libqtile/hook.py", line 360, in fire
    i(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 440, in draw_panel
    self._tree.draw(self, 0)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 72, in draw
    top = i.draw(layout, top, level)
  File "/usr/lib/python3.5/site-packages/libqtile/layout/tree.py", line 236, in draw
[...]

PR #925, opened 2016-05-27
fix-#924

andering:
`layout._layout` is instance of TextLayout class not of Drawer class.

PR #927, opened 2016-05-28
#924 fix

andering:
layout._layout is instance of TextLayout class not of Drawer class.


flacjacket:
This looks good, but I'll merge it through #925

PR #928, opened 2016-05-28
Place transient window above and on the middle of the other window

nicoe:
This patch aims to place the transient window on a more sensible spot than the current implementation.

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -179,6 +179,7 @@ class Floating(Layout):
             bw = self.fullscreen_border_width
         else:
             bw = self.border_width
+        above = False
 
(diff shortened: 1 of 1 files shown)


flacjacket:
This looks good! It would also be nice if floating windows would spawn in the center of the screen if not transient, rather than currently where a lot spawn on the edge, but this is definitely a move in the right direction.

nicoe:
* Sean Vig  [2016-05-30 02:39 +0200]: 
>This looks good! It would also be nice if floating windows would
>spawn in the center of the screen if not transient, rather than
>currently where a lot spawn on the edge, but this is definitely a
>move in the right direction.

I don't use much floating windows but I'll take a look at it when I
have some time.

PR #930, opened 2016-06-08
Update MPD Widget and fix some issues

jeromelebleu:
This makes use of the config framework - as commented as a TODO - at the MPD Widget initialization.
This also fixes some issues and behavior I've encountered when trying to use it with [python-mpd2](https://github.com/Mic92/python-mpd2) - which seems to come in Debian by the way.

*I'm just discovering how qtile is done, so any reviews and suggestions are really welcome! Thanks by the way for this nice window manager!*

### DIFF
--- a/libqtile/widget/mpdwidget.py
+++ b/libqtile/widget/mpdwidget.py
@@ -252,7 +252,7 @@ class Mpd(base.ThreadPoolText):
                     except (ZeroDivisionError, ValueError):
                         playing = pangocffi.markup_escape_text(text)
                     else:
-                        playing = '<span color="%s">%s</span>%s' % (
+                        playing = '<span color="{0}">{1}</span>{2}'.format(
                             utils.hex(self.foreground_progress),
                             pangocffi.markup_escape_text(text[:progress]),
                             pangocffi.markup_escape_text(text[progress:])
--- a/libqtile/widget/mpdwidget.py
+++ b/libqtile/widget/mpdwidget.py
@@ -117,6 +117,7 @@ class Mpd(base.ThreadPoolText):
                     self.client.disconnect()
                 except Exception:
                     pass
+                return False
 
         self.connected = True
         return True
@@ -281,8 +282,7 @@ class Mpd(base.ThreadPoolText):
 
         if was_connected:
             try:
-                self.client.send_idle()
-                self.client.fetch_idle()
+                self.client.ping()
             except mpd.ConnectionError:
                 self.client.disconnect()
                 self.connected = False
--- a/libqtile/widget/mpdwidget.py
+++ b/libqtile/widget/mpdwidget.py
@@ -76,7 +76,7 @@ class Mpd(base.ThreadPoolText):
         self.add_defaults(Mpd.defaults)
         self.inc = 2
         self.client = mpd.MPDClient()
-        self.connected = False
+        self.connected = None
         self.stop = False
 
     def finalize(self):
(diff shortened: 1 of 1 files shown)


libqtile/widget/mpdwidget.py
```diff
@@ -263,17 +244,19 @@ def _get_status(self):
...
-                        utils.hex(self.foreground_progress),
-                        pangocffi.markup_escape_text(text[:progress]),
-                        pangocffi.markup_escape_text(text[progress:])
-                    )
+                if self.do_color_progress and self.status.get('time'):
+                    try:
+                        elapsed, total = self.status['time'].split(':')
+                        percent = float(elapsed) / float(total)
+                        progress = int(percent * len(text))
+                    except Exception:
```
flacjacket: What sort of exception are you catching here, just division by 0? It's best not to have bare `except` clauses (or equivalently `except Exception`

libqtile/widget/mpdwidget.py
```diff
@@ -288,27 +271,26 @@ def poll(self):
         was_connected = self.connected
 
         if not self.connected:
-            if self.reconnect:
+            if self.connected is None or self.reconnect:
                 while not self.stop and not self.connect(quiet=True):
                     time.sleep(self.reconnect_interval)
             else:
-                return
+                return None
```
flacjacket: I think someone has been cleaning up return statements like this to just be `return`, since `return None` is equivalent to `return`

libqtile/widget/mpdwidget.py
jeromelebleu: At least division by 0, and maybe also if `time` is not correctly formated and doesn't contain `:` - which will raise a `ValueError` but I don't know if it can occur. I agree with not having a bare excpet clause... What do you suggest to do instead? Catching only those two exceptions?

Thanks for your review by the way!

flacjacket: You can try catching those 2, if there are any other bugs or things we should catch, they should show up in the log so we can take care of them.

jeromelebleu: Okay, thanks. I've forgot that but there is already a catch all just under...! https://github.com/qtile/qtile/pull/930/files#diff-75182f151c9dbcacbadc4354b44b83afR265

libqtile/widget/mpdwidget.py
```diff
@@ -263,17 +244,19 @@ def _get_status(self):
...
-                    )
+                if self.do_color_progress and self.status.get('time'):
+                    try:
+                        elapsed, total = self.status['time'].split(':')
+                        percent = float(elapsed) / float(total)
+                        progress = int(percent * len(text))
+                    except (ZeroDivisionError, ValueError):
+                        playing = pangocffi.markup_escape_text(text)
+                    else:
+                        playing = '<span color="%s">%s</span>%s' % (
```
Azd325: Could you use python [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format) for this?

jeromelebleu: Sure!

Azd325: Thanks for you work 👍

jeromelebleu: With pleasure! :)

flacjacket:
Great! Thanks for the fixes!

jeromelebleu:
With pleasure! :)

PR #935, opened 2016-07-12
Implement new mpd widget

AlexTalker:
Since previous based on deprecated python-mpd
(support dropped in 2010) new and extensible
widget for mpd will be useful.

Code passes `make lint`(except that iqshell breaks it)
and works with:
```
$ python --version
Python 3.5.2
```
```
$ python2 --version
Python 2.7.12
```

I tried to make it maximally able to change without inheritance, guess that's good idea.


AlexTalker:
@flacjacket thank you too for the amazing wm :+1:

PR #937, opened 2016-07-18
Added watts to be used with format option

andering:
I personally use as main power management attribute watts (consumption),
because that only really tells me how efficient system is in a moment.

Edit which doesn't cost anything, so keep it or remove it :)


flacjacket:
Great, thanks!

PR #938, opened 2016-07-19
Ignore mouse wheel in GroupBox if desired

Etn40ff:
This PR adds the ability to disable wheel scrolling in GroupBox if desired via the flag ignore_mouse_wheel. This feature can be handy when scrolling with two fingers on a touchpad.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -193,7 +193,7 @@ class GroupBox(_GroupBase):
             "Disable dragging and dropping of group names on widget"
         ),
         ("invert_mouse_wheel", False, "Whether to invert mouse wheel group movement"),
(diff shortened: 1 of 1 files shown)


ramnes:
Wouldn't it be more logical to have a `mouse_wheel` parameter that's `True` by default rather than the opposite? :innocent:

Etn40ff:
Either way should be fine, I guess it is more a question of what is the behaviour on similar flags across qtile. I have no idea on this though. Feel free to change this in case.

PS: maybe `use_mouse_wheel` ?

ramnes:
Yeah, fair enough! :+1:

flacjacket:
@Etn40ff if you change to `use_mouse_wheel`, I'd be happy to merge this

Etn40ff:
Done

flacjacket:
Great, thanks for the fix!

PR #939, opened 2016-08-01
support streaming song titles

zordsdavini:
or songs without artist but with title


flacjacket:
Ah, it looks like your only travis failure was a problem with coveralls, I'll get this merged in, thanks!

Issue #68, opened 2011-04-23
Only one window should have focus decoration with multi screens


mattharrison:
When switching focus from one screen to another there ends up being a focused window on each screen (ie both windows have the focus decoration). I would think that since only one window can have focus, there should only be one window with focus decoration.

zordsdavini:
__Test__
I set both monitors on MonadTall layout. I get focus decorations only on one window. I think  this can be closed. 
![2019-10-16-092644_3840x1200_scrot](https://user-images.githubusercontent.com/4191548/66893713-55fb3d80-eff7-11e9-8280-cad3ba38554d.png)

tych0:
Seems like it's layout specific then? For the layouts I use (xmonad and stack) it works.

m-col:
Lots of changes since last October could have fixed this. Still a thing? :)

ramnes:
yes

ramnes:
Oh wait, it only happens when I toggle groups on one screen, not when I jump from a screen to another (with the mouse or with a keybinding). That may explain why you guys couldn't reproduce.

Issue #498, opened 2014-09-08
border_focus_this_screen and border_focus_other_screen for layouts


chetan-reddy:
The GroupBox widget allows for different colors for the active group on a screen based on whether the screen is current or not(this_screen_border and other_screen_border). Can we similarly have options on the layouts to change the border on a window based on whether the screen is current or not?

PR #942, opened 2016-08-05
Widget Groupbox group highligh recognize screens now

andering:
#68 issue

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -285,10 +285,14 @@ class GroupBox(_GroupBase):
             else:
                 text_color = self.inactive
 
-            if g.screen:
+            if g.screen == self.bar.screen:
                 if self.highlight_method == 'text':
                     border = self.bar.background
-                    text_color = self.this_current_screen_border
+                    if self.qtile.currentScreen == self.bar.screen:
+                        text_color = self.this_current_screen_border
+                        to_highlight = True
+                    else:
+                        text_color = self.this_screen_border
                 else:
                     if self.bar.screen.group.name == g.name:
                         if self.qtile.currentScreen == self.bar.screen:


andering:
not sure but maybe it address even this #498 issue

libqtile/widget/groupbox.py
```diff
@@ -285,10 +285,14 @@ def draw(self):
             else:
                 text_color = self.inactive
 
-            if g.screen:
+            if g.screen == self.bar.screen:
```
flacjacket: It might be good to add an `elif g.screen` to configure the color of the group on the other screen, see the `else` on L292 below. Also that logic in the `else` block can then be condensed

andering:
I guess give it try, because i'm not sure if i know what you mean. thx.

andering:
Before it we should make clear color variables:

1. I believe like this should it work now (even not really):

`active`    - Active group font colour      (if there are windows inside of group)
`inactive` - Inactive group font colour   (if there are not windows inside of group)

`background`  - Widget background color
`foreground` - Foreground colour **(no role because active/inactive)**

`highlight_method` - Method of highlighting ‘border’, ‘block’, ‘text’, or ‘line’
`highlight_color` - Active group highlight color when using ‘line’ highlight method. **(why?)**

`this_current_screen_border` - highlight color of group when active (focused).
`this_screen_border` - highlight color of unactive group on active screen
`other_screen_border` - highlight color of group on unactive screen

`urgent_alert_method` - urgent group 
`urgent_border` - Urgent group color
`urgent_text` - Urgent group font color

2. I propose:

`active`  - keep as it is
`inactive` - keep

`background`  - Widget background color
`foreground` - **alias for `inactive` or remove it**

`highlight_method` - keep
`highlight_color` - **remove**

`this_current_screen_border` - keep
`this_current_screen_text` - highlight text color of group when active(focused)
`this_screen_border` - keep
`this_screen_text` - highlight text color of unactive group on active screen
`other_screen_border` - keep
`other_screen_text` - highlight text color of group on unactive screen

`urgent_alert_method` - urgent group 
`urgent_border` - Urgent group color
`urgent_text` - Urgent group font color

3. I propose also renaming:

`this_current_screen_border` -> `highlight_active_group_color`
`this_current_screen_text`  -> `highlight_active_group_text` 

`this_screen_border`  -> `highlight_unactive_group_color`
`this_screen_text`  -> `highlight_unactive_group_text` 

`other_screen_border`  -> `highlight_unactive_screen_color`
`other_screen_text`  -> `highlight_unactive_screen_text`

`urgent_alert_method`  -> `urgent_method`
`urgent_color`  -> `urgent_color`
`urgent_text`  -> `urgent_text`

4. **furthermore**:

I would make `highlight` and `urgent` object/struct/array,
where you could define color and text color for each group [‘border’, ‘block’, ‘text’, ‘line’]

What do you think ? :)

flacjacket:
The renaming looks good to me. I would just say it should be inactive, rather than unactive. The object for defining color may be nice, but we just need to be sure whatever is chosen for that isn't awkward and fits in with the current scheme of configuration by passing kwargs. It may even be cleanest to let dictionaries be passed (e.g. pass in `highlight_active={'text': xxx, 'color': xxx}`)

andering:
So:

```
highlight_active_group={'text': xxx, 'color': xxx}
highlight_inactive_group={'text': xxx, 'color': xxx}
highlight_inactive_screen={'text': xxx, 'color': xxx}
urgent={'text': xxx, 'color': xxx}
```

or even

```
highlight_active_group={'text': xxx, 'color': xxx, 'method': xxx}
highlight_inactive_group={'text': xxx, 'color': xxx, 'method': xxx}
highlight_inactive_screen={'text': xxx, 'color': xxx, 'method': xxx}
urgent={'text': xxx, 'color': xxx, 'method': xxx}

```

flacjacket:
@andering sorry for the long delay. Yes, i think that would be right. I'm not sure if it makes sense to have a different highlighting method based on whether or not the group is highlighted, but for sure the first option would be correct.

ramnes:
This PR has been closed automatically because we removed `develop`, which was targeted by this branch; if anyone wants to implement the modifications proposed here, feel free to open a new PR that targets `master`!

Issue #265, opened 2013-02-01
Prompt widget doesn't support capslock


dequis:
Shift works, capslock is ignored.

aniruddhkanojia:
Cannot reproduce this bug.

flacjacket:
This may be fixed by #943, if it is confirmed we can close it

dequis:
Yep, works! Thanks!

flacjacket:
```
Jan 31, 2013
```
Nice to clear this one out

jiiiijiij:
9 years later and I run to this problem :((

PR #943, opened 2016-08-07
Temporary fix for Capital letters for prompt.py

andering:
Address #265 issue,
but i believe there is more about qtile conversion between keycode and keysyms,
but i will create separate issue for it.


libqtile/widget/prompt.py
```diff
@@ -616,14 +616,18 @@ def _get_keyhandler(self, k):
                 self.completer.reset()
             return self.keyhandlers[k]
 
-    def handle_KeyPress(self, e):
+    def handle_KeyPress(self, e):
```
ramnes: whitespace

flacjacket: Hmm, I'm surprised that doesn't trip our linter, it probably should

Azd325: https://travis-ci.org/qtile/qtile/jobs/150490065#L223

libqtile/widget/prompt.py
```diff
@@ -25,7 +25,7 @@
 #
 # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 # IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVhENT SHALL THE
```
flacjacket: Not sure how this snuck in there :)

andering: I'm very sorry, i believe it has to be very annoying for you to correct those things after me.
I'll try be more careful next-time. Fixed. Thank you.

flacjacket: No worries, we're glad to have people contribute, more developers means more gets done and it's much easier to check for things like this than to do the actual work on the issue.

flacjacket:
Thanks for the fixes, this is in

PR #945, opened 2016-08-12
Fix xcbq.Window.query_tree return value

frostidaho:
xcbq.Window.query_tree currently returns the root window where it should return the parent window


flacjacket:
Sorry for the delay in merging, but thanks for this!

PR #946, opened 2016-08-17
add name of widget to error about orientation support missing

aronj:
this makes the error easier to apply in a configuration fix.


flacjacket:
I've just merged this in 55f90c54aab9afb077a931ef023187ae8d8c9a8f, thanks for the fix!

PR #949, opened 2016-09-05
Fix double braces in docs lazy.rst

ei-grad:


flacjacket:
Thanks for the docs fix!

PR #950, opened 2016-09-06
when qtile can't found key, self.conn.get_modifire return None.

himaaaatti:
When I remove caps_lock key from keyboard by xmodmap command, I couldn't restart qtile.
So I fixed that ModMasks allow to receive None.


flacjacket:
Looks good, thanks for the update

PR #954, opened 2016-09-14
Robustifying default widgets

classner:
Hi qtile team,

this is a small collection of updates to the standard widgets to make them easier to use and more robust (or just work with the current Python version of other libraries). In detail:

**Wallpaper widget**:
* add option to display a fixed string instead of current image name. This can quite clutter the bar.
* remove a bug that required the folder name to end in '/'. Use os.path.join instead.

**khal widget**:
* use `unicode` to make it work with Python2.
* rewrote the event parsing in a more robust way that works with the current version of khal.

**mailfolder widget**:
* did not use user folder expansion for the mail folder.

**Volume widget**:
* expose commands `increase_vol`, `decrease_vol` and `mute` for easy and consistent keybinding in the configs.

### DIFF
--- a/libqtile/widget/khal_calendar.py
+++ b/libqtile/widget/khal_calendar.py
@@ -82,7 +82,7 @@ class KhalCalendar(base.ThreadedPollText):
         # and get the next event
         args = ['khal', 'agenda', '--days', str(self.lookahead)]
         cal = subprocess.Popen(args, stdout=subprocess.PIPE)
-        output = unicode(cal.communicate()[0], 'utf-8')
+        output = cal.communicate()[0].decode('utf-8')
         output = output.split('\n')
         if len(output) < 2:
             return 'No appointments scheduled'
@@ -100,7 +100,7 @@ class KhalCalendar(base.ThreadedPollText):
                 try:
                     if output[i] == 'Today:':
                         date = str(now.month) + '/' + str(now.day) + '/' + \
-                                str(now.year)
+                            str(now.year)
                     elif output[i] == 'Tomorrow:':
                         date = str(tomorrow.month) + '/' + str(tomorrow.day) + \
                             '/' + str(tomorrow.year)
--- a/libqtile/widget/khal_calendar.py
+++ b/libqtile/widget/khal_calendar.py
@@ -82,32 +82,36 @@ class KhalCalendar(base.ThreadedPollText):
         # and get the next event
         args = ['khal', 'agenda', '--days', str(self.lookahead)]
         cal = subprocess.Popen(args, stdout=subprocess.PIPE)
-        output = cal.communicate()[0]
-        output = output.decode()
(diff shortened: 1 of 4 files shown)


libqtile/widget/wallpaper.py
```diff
@@ -44,7 +48,7 @@ def __init__(self, **config):
         self.set_wallpaper()
 
     def get_path(self, file):
-        return self.directory + file
+        return os.path.join(self.directory, file)
```
flacjacket: You can do `expanduser` here, too, then it is not needed in the defaults

libqtile/widget/khal_calendar.py
```diff
@@ -74,49 +75,50 @@ def poll(self):
...
             except ValueError:
-                date = output[i]
-                caldate = output[i]
-                continue
-            if endtime > now:
-                data = caldate.replace(':', '') + ' ' + output[i]
+                try:
+                    if output[i] == 'Today:':
+                        date = str(now.month) + '/' + str(now.day) + '/' + \
+                                str(now.year)
```
Azd325: [str.format](https://pyformat.info/) would be nice too

libqtile/widget/khal_calendar.py
```diff
@@ -74,49 +75,50 @@ def poll(self):
...
         # get reminder time in datetime format
         remtime = datetime.timedelta(minutes=self.remindertime)
 
         # parse khal output for the next seven days
         # and get the next event
         args = ['khal', 'agenda', '--days', str(self.lookahead)]
         cal = subprocess.Popen(args, stdout=subprocess.PIPE)
-        output = cal.communicate()[0]
-        output = output.decode()
+        output = unicode(cal.communicate()[0], 'utf-8')
```
classner: Thanks! Didn't know unicode was not available on Py3... Works!

libqtile/widget/khal_calendar.py
```diff
@@ -74,49 +75,50 @@ def poll(self):
...
                                                   ignoretz=True)
                 endtime = dateutil.parser.parse(date + ' ' + output[i][6:11],
                                                 ignoretz=True)
             except ValueError:
-                date = output[i]
-                caldate = output[i]
-                continue
-            if endtime > now:
-                data = caldate.replace(':', '') + ' ' + output[i]
+                try:
```
classner: The parsing code I submitted is more robust than the old one and should not have trouble with older khal versions.

libqtile/widget/khal_calendar.py
classner: Fixed. Was on my Laptop w/o linting...

libqtile/widget/wallpaper.py
classner: Good suggestion, much cleaner!

flacjacket:
Great thanks for the quick fixes!

PR #957, opened 2016-09-22
Remove margin in between groups of GroupBox.

hgaiser:
This removes the use of margin for spacing the groups in GroupBox from one another.

See attached picture for the effective change (top is with this PR, bottom is without this PR). In my opinion the margin should be around the entire GroupBox widget as a whole, not around each group inside a GroupBox. Alternatively an option could be added, something like "spacing", to determine how far apart the boxes should be.

groupbox-changes

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -49,13 +49,12 @@ class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
         self.add_defaults(base.MarginMixin.defaults)
 
     def box_width(self, groups):
-        width, height = self.drawer.max_layout_size(
+        width, _ = self.drawer.max_layout_size(
             [i.name for i in groups],
             self.font,
             self.fontsize
         )
-        return width + self.padding_x * 2 + self.margin_x * 2 + \
-            self.borderwidth * 2
+        return width + self.padding_x * 2 + self.borderwidth * 2
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
@@ -103,7 +102,7 @@ class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
         framed = self.layout.framed(
             self.borderwidth,
             bordercolor,
-            self.padding_x,
+            0,
             pad_y,
             highlight_color
         )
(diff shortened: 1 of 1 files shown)


hgaiser:
I haven't found the correct fix for this yet .. the way it is now, the box of the last group is slightly cut off (see attachment).
screenshot_2016-09-26_12-05-06

flacjacket:
As it stands, I'm -0.5 to this, I have played with my config several times, and margins between groups is definitely good for some styles, and there are ways to add additional spacing between different widgets other than the margin (tho I guess it would be harder to remove spacing between the groupbox with a large margin and the neighboring widget, but that is the only case I can see for this change). Please let me know what you think, I'd be happy to merge a rework of margins and spacing, but I wouldn't like it to be at the cost of removing inter-group name spacing

hgaiser:
I completely understand :) but for my setup I am missing an option to remove the space between individual group names. This is perhaps a good starting point to figure out what would be desired, if you will discuss with me to figure it out.

For example, how would you feel about adding a parameter to the GroupBox widget called "spacing" or "group_spacing" which determines the distance (in pixels) between individual group names? What should be the default value for this parameter? Should it be the old behaviour (margin_x) or should it be some static value?

If something like that would be accepted, I don't mind working on it at some point.

flacjacket:
Does the x `margin` not set the spacing between the groups? I currently have that in my config and that is what I have. I would be alright with something like that, `spacing` is probably the variable name that I would vote for.

hgaiser:
Yeah I think it does, but it also sets the y_margin and the margin to the left and right of the groupbox. I would like to keep those margins but change the margin in between groups. I will work on adding a `spacing` parameter that defines the gap between groups.

hgaiser:
I modified this PR to match the changes discussed. A `spacing` parameter now exists and determines the distance between two groups in the GroupBox widget. A review of this PR would be appreciated because I'm unsure if the changes I made do not affect other parts that I am not using / seeing.

One thing I have to note is that I removed the padding in [this line](https://github.com/qtile/qtile/pull/957/files#diff-19d627178a983f5e8e50d22919c8a88cL106) because that makes it effectively be applied twice to the GroupBox. The padding is already applied in computing the box width, adding it in that line as well doubles the padding which seemed incorrect.

flacjacket:
Thanks @hgaiser for the changes, everything looks ok, but I'll play with it a bit to make sure the logic changes don't expose any weird issues.

hgaiser:
Awesome, sounds good.

PR #961, opened 2016-10-07
Change «cmus» reference for «MOC» in moc.py widget

soulchainer:
@zordsdavini :wink:


flacjacket:
Thanks for the typo fix!

PR #962, opened 2016-10-09
Close #741 Option to disallow dups in prompt hist

### LINKED ISSUE #741: Prompt widget stores duplicates in history
Simply removing the launched command before appending it should be easy. But I wonder if there are better options. Should this be made optional? If not is there a better storage type in python that doesn't keep duplicates?

soulchainer:
Implement option in `prompt.py` widget to implement the
`HIST_IGNORE_ALL_DUPS` or `HIST_IGNORE_DUPS` option present
in some shells/DEs/WMs, for not duplicating entries into the
history record. Closes #741

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -31,7 +31,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import copy
 import glob
 import os
 import pickle
@@ -381,7 +380,7 @@ class Prompt(base._TextBox):
                                         for x in self.completers if x}
                     if self.max_history != \
                        self.history[list(self.history)[0]].maxlen:
-                        self.history = {x: deque(copy.copy(self.history[x]),
+                        self.history = {x: deque(self.history[x],
                                                  self.max_history)
                                         for x in self.completers if x}
             else:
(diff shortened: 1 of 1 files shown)


libqtile/widget/prompt.py
```diff
@@ -650,3 +661,20 @@ def cmd_info(self):
             text=self.text,
             active=self.active,
         )
+
+    def _dedup_history(self):
+        """Filter the history deque, clearing all duplicate values."""
+        self.history = {x: deque(copy.copy(self._dedup_deque(self.history[x])),
```
flacjacket: why do you need to re-wrap this in a deque? also, why is the copy needed here?

soulchainer: Ummm... You got me xD. You're so right. The part of wrapping this into a deque is certainly redundant. I didn't notice that here I was already receiving a deque from `_dedup_deque`. I copied this from other part from my code (line 384 ~), where I was dealing with a list and is clear that I didn't thought enough about it. My bad :þ.

About the `copy.copy`, is the same reason. In this part I'm using a new deque, generated by `_dedup_deque` for every  completer, so there is no need to `copy.copy` it. In the other place, it was needed, because I was working with one list and... ummm, really in that place (line 384) I think I could simply use an slice of the entire list `[:]` and also get rid of the `copy` dependence. WTF was I thinking? >_<.

Well... This can wait, right? I mean, it has been waiting a year, so...
Two days ago (hence my delay answering you) my PC broke: Motherboard is gone. As soon as I can, I will fix this properly. I could edit it now, but I can't try it with qtile in my PC (because I don't have it working), pass the tests and all that stuff.

flacjacket:
Didn't realize that you'd pushed changes, thanks for fixing that. Merging now.

PR #963, opened 2016-10-09
Fix «Hacking on Qtile» documentation page

soulchainer:
Specifically this

```sh
pytest --tests=test_module
```

I think this is wrong. The `--tests` option doesn't appear in
`pytest --help`, neither in the [official
docs](https://readthedocs.org/projects/pytest/search/?q=--tests).
I think this is a mistake, derived from the previous use of
`nosetests`, where a `--tests` option exists. In `pytest`, there isn't such an option.


flacjacket:
Good catch, thanks!

Issue #854, opened 2016-01-27
LibreOffice menus don't appear or don't stay visible 


jrazik:
Hi,

Since several versions of qtile and libreoffice,  I'm not able to use mouse for menu action.
If I click, the subpanel doesn't seem to appear (like "Files" for example). If I use the keyboard it's ok.
If I try to click on the icon to open a file, the menu disappear as soon as I release the mouse button.
If I try to change the font by clicking on the arrow, nothing appear.
No error shown and nothing in qtile.log.
Currently I'm using qtile 0.10.4, libreoffice 5.0.4.2 and fedora 23. May be it's a wayland related problem.
With awesome it works fine.

zordsdavini:
In Archlinux Qtile 0.10.4, libroffice 5.0.4.2 (and I remember in earlier versions) if I select MonadTall layout I can't get menus, too. But if I select Max layout – it appears.

jrazik:
Yeah !
I just try and yes, with Max layout it works. And also we Treetab but not with the other following ones that I tried: floating, monadtall, tile, and stack.

jrazik:
I just test to maximize the LibreOffice window and in this case, it works also, with all my layout.
Thanks again for the temporary solution until understanding why.

flacjacket:
My LibreOffice menus will stay up, but I always have to right click the menu to get it to show up and I was never able to figure out why.

ThomasWaldmann:
same problem here. totally annoying.

ThomasWaldmann:
With current code from master, a normal (left) click on "File" menu works now as expected in LibreOffice.

jrazik:
Still not working with me, even after a git pull just now.

zordsdavini:
on Qtile 0.10.5 problem still exist

On 21 March 2016 at 11:39, jrazik <notifications@github.com> wrote:

> Still not working with me, even after a git pull just now.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/qtile/qtile/issues/854#issuecomment-199193414>
>



-- 
Udovīčė Arns

ThomasWaldmann:
I'm at 2b6a04de5e3f6d6d0e7d30a92c197f0d417f47bb and it works for me.

jrazik:
I revert to this commit but unfortunately it still didn't work.
LibreOffice 5.0.5.2 00(Build:2) Fedora 23, but the problem appear long ago.
[...]

PR #967, opened 2016-10-15
 LibreOffice menus don't appear or don't stay visible #854

kes31:


flacjacket:
Thanks for the documentation on the workaround for that issue!

PR #968, opened 2016-10-15
Added class for combining Match-objects with &&.

kes31:
The new class AndMatch in config.py gives the user the possibility to
have a class for representing an match on multiple properties at the
same time. This is usefull, if you would like to match on class-name and role at the same time.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -568,6 +568,32 @@ class Match(object):
         return '<Match %s>' % self._rules
 
 
+class AndMatch(object):
+    """Combines multiple Match-Objects with logical &&-operation
+
+    Parameters
+    ==========
+    *matches:
+        Match-objects for combining
+    """
+    def __init__(self, *matches):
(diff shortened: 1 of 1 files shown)


libqtile/config.py
```diff
@@ -568,6 +569,32 @@ def __repr__(self):
...
+    *matches:
+        Match-objects for combining
+    """
+    def __init__(self, *matches):
+        self.matches = []
+        self.matches.extend(matches)
+
+    def compare(self, client):
+        _tmp_bool = map(lambda x: x.compare(client), self.matches)
+        return reduce(lambda x, y: x and y, _tmp_bool )
```
flacjacket: I feel like it may be better to use the `all()` builtin here, rather than `and`ing all of the results together

flacjacket:
There are also a couple things that the linter is flagging from flake8, check the Travis log for those. Otherwise, this looks fine.

PR #969, opened 2016-10-23
allow to use uvloop

gawel:
Quick PR after testing that qtile works fine with uvloop which is faster than asyncio's default event loop

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,9 @@
+qtile 0.10.7, unreleased
+
+   * features
+
+       - Use uvloop if installed
+
 qtile 0.10.6, released 2016-05-24:
     !!! qsh renamed to qshell !!!
         This avoids name collision with other packages
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -94,6 +94,19 @@ Based on your Python version, there are different ways to install this:
 .. _trollius: http://trollius.readthedocs.io/
 .. _trollius is deprecated: http://trollius.readthedocs.io/deprecated.html#deprecated
 
+
+uvloop
+-------
+
+Qtile will use `uvloop`_ to replace asyncio's event loop if available.
+You can install it with:
+
+.. code-block:: bash
+
+    pip install uvloop
+
+.. _uvloop: https://pypi.python.org/pypi/uvloop/
+
 dbus/gobject
 ------------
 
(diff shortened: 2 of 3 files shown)


gawel:
Just realised that this can be done in config.main() so you may forget it, or not.

flacjacket:
I'd say this would be fine, but I'd rather have it controlled by a configuration parameter.

gawel:
What about an explicit hook set_event_loop_policy() similar to main() ? Just to ensure that the policy is set before any call to new_event_loop/get_event_loop

flacjacket:
Yeah, that actually sounds better to me.

gawel:
Btw, can you explain why you are using asyncio.new_even_loop instead of get_event_loop ?
Is there a real reason ?
For me get_event_loop assure that you use the same loop everywhere. And will allow to use asyncio.get_event_loop().call_later() in config.py

flacjacket:
I'm not sure there is a good reason for that, but I forget the issues that we hit when we were trying to get it working. If you find that it is possible to work using `get_event_loop()`, then we can change to that.

gawel:
Found this: https://github.com/qtile/qtile/pull/489#issuecomment-54077028

Seems that the only reason is that you think it will slow everything. I'm not sure about that. 
For my irc client I've made some benches: https://github.com/gawel/irc3/blob/master/examples/benches.py
Got 1 server and 1000 clients flooding the server all in the same event loop. Works fine.
Also I don't know how the GIL will affect performances when using more than one event loop in the same thread

I'm currently using a single event loop in qtile without problems, for now. 

But. As I can see, the client implementation is using loop.run_until_complete(). And more than one time. I thought that this was not possible since it stop the loop. 

Basically it run_forever:
https://github.com/python/cpython/blob/a3ac6202ceeeb02b2cb1127d4f438e57e3db80c3/Lib/asyncio/base_events.py#L436

And stop the loop:
https://github.com/python/cpython/blob/a3ac6202ceeeb02b2cb1127d4f438e57e3db80c3/Lib/asyncio/base_events.py#L171

So... I don't get it.

flacjacket:
Reading that, I think using a new eventloop in `manager.py` and getting the eventloop in `ipc.py` wasn't a speed thing, but rather getting the same eventloop everywhere. But using `get_event_loop` in both places should be fine. It may have originally been using a new eventloop in both places.

The run_until_complete function does exactly that, it runs the loop "forever" (i.e. until it is stopped), and when the future you want to run is complete, it triggers the callback that stops the event loop and returns from the function.

gawel:
Yes, but since I'm using only one event loop, I don't see the need to start it (it was already started by the manager). And I don't see how it still running after it was stopped by run_until_complete. Seems weird.

PR #972, opened 2016-10-29
Fixed focus issues

emesik:
Two issues have been addressed:
1. Broken `VerticalTile` layout which did not switch focus between windows. Now it works and also cycling among windows is enabled.
2. Added `cmd_focus` to `Window`, which allows for tricks like this:

```
keys.append(Key([mod, "shift"], g.name,
        lazy.window.togroup(g.name),
        lazy.group[g.name].toscreen(screen),
        lazy.window.focus()))
```

### DIFF
--- a/libqtile/layout/verticaltile.py
+++ b/libqtile/layout/verticaltile.py
@@ -225,7 +225,7 @@ class VerticalTile(Layout):
             index = self.clients.index(window)
             return self.clients[index + 1]
         except IndexError:
-            pass
+            return self.clients[0]
 
     def focus_previous(self, window):
         if not self.clients:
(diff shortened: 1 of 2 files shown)


flacjacket:
You shouldn't change `focus_next` and `focus_previous` to cycle over the available windows, the floating window focus handling relies on this returning `None` when it hits the end of the focus list. When you really want to just focus within the layout (skipping floating windows), you can set the command to use `self.focus_next(self.focused) or self.focus_first()`, otherwise `group.cmd_next_window` handles cycling window focus correctly including floating windows.

flacjacket:
I don't think the first commit is a bug, I think that is the intended behavior, so I'll cherry-pick the second commit and close this, you can open a new PR if you think the first commit needs addressing.

PR #973, opened 2016-10-31
Multiscreen widgets

dpiegdon:
These changes make multi-screen setups not require multi-bars. Personally I only want one bar for all my screens, not one on each of them.
- WindowName will show name of currently focused window, even if not on the same screen as bar,
- GroupBox will highlight non-focused screens.

For a default config the old behaviour is retained; to obtain the new behaviour options must be passed to the widgets.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -169,9 +169,14 @@ class GroupBox(_GroupBase):
             "Border or line colour for group on this screen when unfocused."
         ),
         (
+            "other_current_screen_border",
+            "404040",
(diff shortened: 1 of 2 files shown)


flacjacket:
I like this idea, but it looks like I messed up the mergability of this with other PRs. If you can rebase or merge the current development branch, I can get this in. Thanks!

dpiegdon:
as per 975, i just removed the client_killed hook. that should do the trick.

flacjacket:
Thanks for fixing this up, sorry for the long delay, merging this in now!

PR #976, opened 2016-11-09
Use `border_focus` only for currently focused window

de-vri-es:
Currently, whichever window is considered 'active' by the layout of each screen uses the `border_focus` property for the border colour. That makes it difficult to see which window is truly focused (i.e: the one that will process input) when you have multiple screens.

This PR adds a read-only property `has_focus` to windows to check if it is the one and only focused window, and adjusts most layouts (see below) to use that to select `border_focus` or `border_normal`.

This requires a re-layout of the previous screen when switching screens to update the border colours on the previous screen, so that is included in this PR.

Note that I skipped the `wmii` layout for now because it uses a third border colour and I don't know how these colours should represent the state of the layout (or even what the layout does).

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -242,7 +242,7 @@ class Columns(Layout):
         else:
             client.hide()
             return
-        if client == self.cc.cw:
+        if client.has_focus:
             color = self.group.qtile.colorPixel(self.border_focus)
         else:
             color = self.group.qtile.colorPixel(self.border_normal)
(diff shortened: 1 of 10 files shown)


de-vri-es:
Small note: old behaviour also shows a `border_focus` on the same screen for both the floating layout and the regular layout, so it also affects single-screen setups when floating windows are involved.

flacjacket:
Hmm, this looks pretty nice, I've had to rely on a groupbox widget with multiple color states (focused/un-focused and this/other screen), but this looks like it definitely helps. I would say it might be nice if a similar idea could make it's way into border coloring (focused current/unfocused current/normal), but this doesn't reduce functionality in that regard. Thanks!

PR #977, opened 2016-11-09
Added CurrentLayoutIcon widget.

and3rson:
So I've created this graphical version of the `CurrentLayout` widget, similar to how AwesomeWM displays current layout. Thought it might be useful for you guys.

It also allows you to customize size of the icon and use custom icon theme.

Here are some demos:

Example 1
Example 2
Example 3

Here's the complete map:

Layout icon map

P. S. Please let me know if some layout icons are not explanatory enough, I'm not an artist and will be glad to hear any ideas on how to display layout as unambiguous image. :)


libqtile/widget/currentlayout.py
```diff
@@ -57,3 +69,118 @@ def button_press(self, x, y, button):
...
+
+class CurrentLayoutIcon(base._TextBox):
+    """
+    Display the icon representing the current layout of the
+    current group of the screen, the bar containing the widget, is on.
+    """
+    orientations = base.ORIENTATION_HORIZONTAL
+
+    defaults = [
+        ('scale', 1, 'Scale factor, defaults to 1'),
```
flacjacket: Maybe note that this is relative to the height of the bar

libqtile/widget/currentlayout.py
```diff
@@ -57,3 +69,118 @@ def button_press(self, x, y, button):
...
+            for layout_class, layout_class_name
+            in map(lambda x: (getattr(layout_module, x), x), dir(layout_module))
+            if isinstance(layout_class, (type, types.ClassType)) and issubclass(layout_class, Layout)
+        ]
+
+    def setup_images(self):
+        """
+        Loads layout icons.
+        """
+        for key in self._get_layout_names():
```
flacjacket: My fear with this is it becomes a bit more difficult to add new layouts to this repo without adding a new icon, and on the flip side, when users create custom layouts in their config, they won't be picked up in _get_layout_names. I'm not sure how big of a problem either of these things are and would be up to suggestions on how to fix them (if they are actually problematic).

libqtile/widget/currentlayout.py
```diff
@@ -57,3 +69,118 @@ def button_press(self, x, y, button):
...
+            path = os.path.join(self.theme_path, 'layout-{}.png'.format(key))
+            try:
+                img = cairocffi.ImageSurface.create_from_png(path)
+            except (cairocffi.Error, IOError) as e:
+                self.icons_loaded = False
+                logger.exception(
```
flacjacket: Re above comment, this should probably be lowered to a warning

and3rson:
@flacjacket thanks for your review!

Here's how I solved those issues:

- Provided better class docstring and description for defaults.
- Added default icon for unknown layouts.
- Added handling of custom icon paths that can be configured via `custom_icon_paths` config argument.
- Cairo should never throw Exception now because a default icon is now always found. I left it at `exception` level because the only circumstance for it is when the icon is not an image _or_ is not readable.

Thanks!

and3rson:
@flacjacket Seems like the builds are failing for some unrelated reason.

ramnes:
Thanks a bunch for that contribution. It's been a long time that I wanted to see that feature in Qtile. Bravo!

ramnes:
Tho, it doesn't work with Python 3. See https://github.com/qtile/qtile/issues/982.

PR #978, opened 2016-11-09
Add window.toscreen command.

de-vri-es:
This PR adds a command to send windows to a specified screen (or the current screen if no screen index is given).

I like to be able to quickly move windows between screens without being aware of which group is active on each screen. This command allows users to do just that.


flacjacket:
Great, thanks for this change.

Issue #851, opened 2016-01-21
qtile-dmenu


zordsdavini:
Maybe it would be cool to have wrapper to dmenu from Qtile. It could use local same config for appearance what should be used in all cases. Other thing is expanding it to use interact with user. One thing is simple dmenu functionality, other you could enter new choice. For example I get some selection for category but I enter not existing and it creates new category. Or it can give simple input (ex. password). 
This is example how I use dmenu with Qtile (window list, moc navigation, app list, clipboard (this doesn't work correctly): https://youtu.be/CQT95lUBk88

zordsdavini:
#979 please label it as work in progress

zordsdavini:
Pull request is ready for merge with development branch

PR #979, opened 2016-11-12
Dmenu wrapper for #851

zordsdavini:
still in progress
issue: https://github.com/qtile/qtile/issues/851

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -90,6 +90,7 @@ class File(object):
             "widget_defaults",
             "bring_front_click",
             "wmname",
+            "extentions",
         ]
 
         for option in config_options:
--- a/libqtile/extention/__init__.py
+++ b/libqtile/extention/__init__.py
@@ -1,0 +1,24 @@
+# Copyright (c) 2016 zordsdavini
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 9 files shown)


zordsdavini:
08:41 < zordsdavini> could someone help a little bit with coding?
08:42 < zordsdavini> I wanted to add dmenu wrapper and add some extention like window list and run commands through dmenu
08:43 < zordsdavini> the thing is I set config from config.py, then adding some stuff and another call to extention will have this added stuff. I don't want that
08:43 < zordsdavini> at first I used configurable
08:44 < zordsdavini> later removed and added simple dict
08:44 < zordsdavini> but this added stuff still exist
08:45 < zordsdavini> I deleting Dmenu class and each time creating new config for separate extention

zordsdavini:
config.py should have something like:

    from libqtile import extention
    mod = 'mod4'
    keys = [
        ...
        Key([mod], 'l', lazy.run_extention(extention.WindowList)),
        Key([mod], 'r', lazy.run_extention(extention.DmenuRun)),
        ...
    ]
    extentions = {
        'dmenu': {
            'prompt': "→",
            'font' : "Andika-8",
            'background' : "#15181a",
            'foreground' : "#00ff00",
            'selected_background' : "#079822",
            'selected_foreground' : "#fff",
            'height' : 24,
        }
    }

zordsdavini:
still work in progress. Needs to be done:
* python2 support (when we will drop that support?)
* glitch in window list extension

flacjacket:
You can add `  # noqa` to the end of the lines to avoid those sorts of problems with flake8

flacjacket:
Thanks for this, I think it looks good! If you are able to add some docs for this, that would be great!

PR #984, opened 2016-11-27
Added patch to discover resources during build.

and3rson:
I noticed a bug when installing `qtile` on a different machine: `setup.py` wasn't properly including the `layout-icons` dir into `resources` dir during build.

So I've made this small fix to discover existing dirs & avoid this issue in future if someone forgets to add his resources into `setup.py`.

Although I prefer having such things hard-coded and don't like having this sort of "magic" discovery code, but I feel like this one is still gonna save someone's time in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import os
 import sys
 import textwrap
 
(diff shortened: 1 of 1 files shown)


ramnes:
I'm not really sure to understand why we want this, and how it fixes your problem.

and3rson:
@ramnes this fixes the bug where `setup.py` does not collect all required resources (PNG images) when installing via it. This breaks `CurrentLayoutIcon` when installing qtile from `AUR`

ramnes:
I can read, no need to repeat your first comment. :smile: 

What I don't understand is how it is possible that `setup.py` doesn't collect the PNG images in the first place, and why those changes fixes it.

flacjacket:
The old setup.py would get the battery-icons, but not the layout-icons. An
alternate change would be to add layout-icons to the script explicitly.

ramnes:
Aaaaaaaah, okay. I didn't get the subtlety between the two directory names, sorry.

I guess adding explicitly `layout-icons` would be simpler to read and understand, and avoid any side effect. My two cents.

PR #988, opened 2016-12-02
Fixed a bug that prevents bar from redrawing until a group is first switched.

and3rson:
@flacjacket Here's a small fix to `CurrentLayoutIcon` which fixes problems on the start.


flacjacket:
Great, thanks!

PR #992, opened 2016-12-18
Mpris improvements

tych0:
I use both spotify and clementine now, depending on what I want to listen to. Here's some improvements to the mpris widgets as a result.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ check-cov:
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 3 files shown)


flacjacket:
Nice, are you saying you can use the mpris widget with clementine or spotify? It would be cool to have docs on how to setup that up if it's possible.

tych0:
Yep; I can send patches for some docs at some point today. I should also PR my updated config to qtile examples, since the one there is ancient :)

PR #996, opened 2017-01-11
added some new features to widgets:

gsaga:
1) volume widget turns red when sound is muted instead of just
displaying an 'M', so that you can see the
change in volume even when it is muted.

2) added a new default interface 'net' which shows the total network
traffic. Modified '_format()' to display 3 significant figures. Replaced '^' and 'v' with up and down arrows.

3) battery widget turns red when the battery level is above 90% and the
charger is plugged in (useful for laptops).

Sorry for some random changes which should have been made in config.py, I copied these files from my qtile install.

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -38,17 +38,17 @@ class Net(base.ThreadedPollText):
 
     def convert_b(self, b):
         # Here we round to 1000 instead of 1024
-        # because of round things
+        #?# because of round things
         letter = 'B'
         # b is a float, so don't use integer division
-        if int(b / 1000) > 0:
-            b /= 1000.0
-            letter = 'k'
-        if int(b / 1000) > 0:
-            b /= 1000.0
+        if int(b / 1024) > 0:
+            b /= 1024.0
+            letter = 'K'
+        if int(b / 1024) > 0:
+            b /= 1024.0
             letter = 'M'
-        if int(b / 1000) > 0:
-            b /= 1000.0
+        if int(b / 1024) > 0:
+            b /= 1024.0
             letter = 'G'
         # I hope no one have more than 999 GB/s
         return b, letter
@@ -57,25 +57,33 @@ class Net(base.ThreadedPollText):
         lines = []  # type: List[str]
         with open('/proc/net/dev', 'r') as f:
             lines = f.readlines()[2:]
-        interfaces = {}
+        interfaces = {'net':{'down': 0, 'up': 0}}
         for s in lines:
             int_s = s.split()
             name = int_s[0][:-1]
             down = float(int_s[1])
             up = float(int_s[-8])
             interfaces[name] = {'down': down, 'up': up}
+            interfaces['net']['down'] += down
+            interfaces['net']['up'] += up
+        # net_up = 0
+        # net_down = 0
+        # for iface in interfaces:
+        #     net_up += interfaces[iface]['up']
+        #     net_down += interfaces[iface]['down']
+        # interfaces['net'] = {'down': down, 'up': up}
         return interfaces
 
     def _format(self, down, up):
         down = "%0.2f" % down
         up = "%0.2f" % up
-        if len(down) > 5:
-            down = down[:5]
-        if len(up) > 5:
-            up = up[:5]
+        if len(down) > 4:
+            down = down[:4]
+        if len(up) > 4:
+            up = up[:4]
 
-        down = " " * (5 - len(down)) + down
-        up = " " * (5 - len(up)) + up
+        down = "0" * (4 - len(down)) + down
+        up = "0" * (4 - len(up)) + up
         return down, up
 
     def poll(self):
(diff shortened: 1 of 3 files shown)


libqtile/widget/net.py
```diff
@@ -57,25 +57,33 @@ def get_stats(self):
...
             up = float(int_s[-8])
             interfaces[name] = {'down': down, 'up': up}
+            interfaces['net']['down'] += down
+            interfaces['net']['up'] += up
+        # net_up = 0
+        # net_down = 0
+        # for iface in interfaces:
+        #     net_up += interfaces[iface]['up']
+        #     net_down += interfaces[iface]['down']
+        # interfaces['net'] = {'down': down, 'up': up}
```
flacjacket: remove these extra lines

libqtile/widget/net.py
```diff
@@ -57,25 +57,33 @@ def get_stats(self):
...
         down = "%0.2f" % down
         up = "%0.2f" % up
-        if len(down) > 5:
-            down = down[:5]
-        if len(up) > 5:
-            up = up[:5]
+        if len(down) > 4:
+            down = down[:4]
+        if len(up) > 4:
+            up = up[:4]
```
flacjacket: Why are you making this change? If you format a number >100 and < 1000, for example 123.456, you'll end up with something like `123.` here, where the trailing period is odd, rather than `123.4`

libqtile/widget/net.py
```diff
@@ -57,25 +57,33 @@ def get_stats(self):
...
 
-        down = " " * (5 - len(down)) + down
-        up = " " * (5 - len(up)) + up
+        down = "0" * (4 - len(down)) + down
+        up = "0" * (4 - len(up)) + up
```
flacjacket: I feel like a better option here is to make the format string configurable. They you can do `%04.2f` if you want leading zeros.

libqtile/widget/net.py
```diff
@@ -93,7 +101,7 @@ def poll(self):
 
             down, up = self._format(down, up)
 
-            str_base = u"%s%s \u2193\u2191 %s%s"
+            str_base = u"%s%s\u2193\u2191%s%s"
```
flacjacket: Again, maybe better configurable

libqtile/widget/volume.py
```diff
@@ -10,7 +10,6 @@
 # Copyright (c) 2014 Adi Sieker
 # Copyright (c) 2014 dmpayton
 # Copyright (c) 2014 Jody Frankowski
-# Copyright (c) 2016 Christoph Lassner
```
flacjacket: I feel like you are accidentally reverting a commit here.

libqtile/widget/volume.py
```diff
@@ -72,13 +68,17 @@ class Volume(base._TextBox):
...
+        self.wasmuted = False
+        # self.volume = 0
+        # self.update()
         self.add_defaults(Volume.defaults)
         if self.theme_path:
             self.length_type = bar.STATIC
             self.length = 0
-        self.surfaces = {}
-        self.volume = None
+        self.volume = 0
```
flacjacket: I think this should be None, so we always do the drawing update on the first update

libqtile/widget/volume.py
```diff
@@ -72,13 +68,17 @@ class Volume(base._TextBox):
     ]
 
     def __init__(self, **config):
-        base._TextBox.__init__(self, '0', width=bar.CALCULATED, **config)
+        base._TextBox.__init__(self, '      ', width=bar.CALCULATED, **config)
```
flacjacket: Why this? It actually may be better to have some error message as the default string.

libqtile/widget/volume.py
```diff
@@ -125,8 +125,10 @@ def button_press(self, x, y, button):
         self.draw()
 
     def update(self):
+        # if self.volume >= 0:
+        #     self.vol_before_mute = self.volume
```
flacjacket: delete this

libqtile/widget/volume.py
```diff
@@ -158,10 +160,11 @@ def _update_drawer(self):
             elif self.volume >= 80:
                 self.text = u'\U0001f50a'
         else:
-            if self.volume == -1:
-                self.text = 'M'
+            if self.muted:
```
flacjacket: You need to make changes to the other if blocks, too

gsaga: fixed the issues

libqtile/widget/volume.py
```diff
@@ -212,8 +215,11 @@ def get_volume(self):
         except subprocess.CalledProcessError:
             return -1
 
+        self.wasmuted = self.muted
         if '[off]' in mixer_out:
-            return -1
+            self.muted = True
```
poulp: self.muted = '[off]' in mixer_out

ramnes:
This MR has been closed because we removed the `develop` branch; if anyone wants to implement the modifications proposed here, feel free to open a new PR that targets `master`!

PR #998, opened 2017-01-14
xcbq: decode WM_NAME as a UTF8 string

dequis:
Fixes unicode in titles in apps like steam, which doesn't have the other properties like _NET_WM_NAME.

_propertyString() decodes as latin1 by default. This was probably left like this because it was a rarely reached path.

### DIFF
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -455,7 +455,7 @@ class Window(object):
             xcffib.xproto.GetPropertyType.Any
         )
         if r:
-            return self._propertyString(r)
+            return self._propertyUTF8(r)
 
     def get_wm_hints(self):
         r = self.get_property("WM_HINTS", xcffib.xproto.GetPropertyType.Any)
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -450,12 +450,16 @@ class Window(object):
         if r:
             return self._propertyUTF8(r)
 
+        r = self.get_property(xcffib.xproto.Atom.WM_NAME, "UTF8_STRING")
+        if r:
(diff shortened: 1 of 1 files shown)


flacjacket:
Interestesting, it looks like `WM_NAME` should be either Latin-1 or "Compound Text" [1], I'm not sure if compound text [2] is the same as utf8. I'm just trying to figure if there is something we could do to detect that, assuming applications are setting the type of the property correctly.

[1] https://tronche.com/gui/x/icccm/sec-4.html#WM_NAME https://tronche.com/gui/x/icccm/sec-2.html#s-2.7.1
[2] https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/CTEXT/ctext.html

dequis:
xprop says `WM_NAME(UTF8_STRING) = "ᴿᴺᴳ togglr - Chat"`

Also it looks like you weren't there when i talked about it, so copying here:

```
< dx> looks like steam sends stuff with the wrong encoding in the window
      title http://dump.dequis.org/hT3wq.png
< dx> either that or our bug
< dx> text is ᴿᴺᴳ togglr
< dx> xprop and xwininfo display the right thing
< dx> beep http://dump.dequis.org/lK5hN.png
< dx> weird how this only fails for steam
< dx> oh it's not weird
< dx> other apps just have _NET_WM_VISIBLE_NAME or _NET_WM_NAME and those
      are read as utf8
< dx> https://github.com/qtile/qtile/pull/998
```

dequis:
Added another commit to request UTF8_STRING explicitly first, then try the other types and decode those as crap-8859-1

flacjacket:
Alright, this should be good then, thanks!

PR #1006, opened 2017-02-01
Add missing import in doc running inside gnome

poulp:


flacjacket:
Ah, thanks for the fix

Issue #1007, opened 2017-02-02
Bug in libqtile.config.Screen.resize


frostidaho:
config.Screen.resize() currently will always raise an exception.

Here is the [relevant code](https://github.com/qtile/qtile/blob/5ec03621f2c755042d07b4e7e1173be49c8c5301/libqtile/config.py#L380).

`group.layoutAll()` always returns `None` and call_soon expects a callable.

PR #1008, opened 2017-02-02
Fix exception in config.Screen.resize

frostidaho:
Issue 1007 - https://github.com/qtile/qtile/issues/1007


flacjacket:
Thanks for the bug report and the quick fix!

Issue #1010, opened 2017-02-17
Bug in test/conftest/_find_display() - can't run test suite


frostidaho:
### Background
With GDM and xserver running I'm using displays `:0` and `:1`.
So the next available display is `:2`.

Neither GDM or the xserver write their lock file as `/tmp/.X0-lock` or `/tmp/.X1-lock`.
Rather they write to `/tmp/.X11-unix/X0` and `/tmp/.X11-unix/X1`.

### Problem description
On my system normal x11 doesn't write a lockfile to `/tmp/.XN-lock` where `N` is the display number. It only creates a file at `/tmp/.X11-unix/XN`. Xephyr creates files at both locations.

[conftest._find_display()](https://github.com/qtile/qtile/blob/9e59c3638a1ff762a8340ce0a5545df2a87e1422/test/conftest.py#L54) currently only looks for files at `/tmp/.XN-lock`.

It incorrectly gives the next display as `:1` since it doesn't look in the `/tmp/.X11-unix` directory

PR #1011, opened 2017-02-17
Fix bug in finding the next available display in conftest.py

frostidaho:
For issue https://github.com/qtile/qtile/issues/1010


flacjacket:
Nice find, thanks!

PR #1012, opened 2017-02-18
qshell path reference / tab completion improvements

dequis:
- Allow root-relative paths
- Actually use the argument passed to ls
- Add a slash when there's only one cd tab completion suggestion to
continue tab completing inner paths


flacjacket:
Cool, this looks good, thanks!

PR #1013, opened 2017-02-25
New monad

ThomasChiroux:
Hi,
After upgraded to the last release, i remembed that I was using my own custom monad layouts since a few years ago. I did not took time to adapt them to the new evolutions and it needed refactor and simplication.

So here is a try:
- made min_ratio, max_ratio and ex min_height, now min_secondary_size configurable.
- added a MonadWide class which inherit from MonadTall and only changes necessary elements to be able to draw a horizontal monad.
- added the monad wide icon.

The MonadWide class could be event shorter, with less code repetition by playing with the two following inversions: 
  self.group.screen.dwidth <=> self.group.screen.dheight
  client.height <=> client.width
by for example storing in __init__ two new properties self.screen_size and self.client_size  and use this properties in _get_relative_size_from_absolute, _get_absolute_size_from_relative, _maximize_secondary and _shrink_secondary. So the only specific method in MonadWide would be the specific configurations.

Personnaly I thought it was a bit less readable, so i opted for a little code repetition, but let me know if you prefer this second solution.

Thomas.


libqtile/layout/xmonad.py
```diff
@@ -367,19 +379,19 @@ def configure(self, client, screen):
...
                 self.group.screen.dy + self.margin,
                 width - self.margin,
-                self.group.screen.dheight - 2 * self.border_width - 2 * self.margin,
+                (self.group.screen.dheight -
+                    2 * self.border_width - 2 * self.margin),
                 self.border_width,
                 px,
             )
-            client.unhide()
+
```
flacjacket: Also, Travis pep8 checking doesn't like having 2 blank lines here. I feel like silly checks like this could be safely ignored...

libqtile/layout/xmonad.py
```diff
@@ -138,9 +138,9 @@ class MonadTall(Layout):
 
     """
 
-    _left = 0
-    _right = 1
-    _min_height = 85
+    _align_min = 0
+    _align_max = 1
```
ThomasChiroux: yep, I totally agree. I was trying to find a generic name for this, but clearly without success :-) will do that.

libqtile/layout/xmonad.py
```diff
@@ -138,9 +138,9 @@ class MonadTall(Layout):
...
+    _min_secondary_size = 85
     _min_ratio = .25
     _med_ratio = .5
     _max_ratio = .75
```
ThomasChiroux: ok. Was not sure about your policy for this because of the ratio parameter which was linked to self._med_ratio

libqtile/layout/xmonad.py
```diff
@@ -787,3 +799,227 @@ def cmd_right(self):
...
+        Key([modkey], "i", lazy.layout.grow()),
+        Key([modkey], "m", lazy.layout.shrink()),
+        Key([modkey], "n", lazy.layout.normalize()),
+        Key([modkey], "o", lazy.layout.maximize()),
+        Key([modkey, "shift"], "space", lazy.layout.flip()),
+
+    """
+
+    def _get_relative_size_from_absolute(self, absolute_size):
+        return float(absolute_size) / self.group.screen.dwidth
```
ThomasChiroux: right, don't know why it's here, perhaps a copy/paste from my old py2.7 layout, sorry for that :-)

ThomasChiroux:
So, here is a new try, with the discussed updates.
I've also updated documentation and unittest (and while porting unittest, I realised I forgot to port swap_left, swap_right and swap_main, so done that also: I choose not to change the api and kept swap_left and right (i could be here swap_up and down), but i inverted the behaviour: i thought it was more 'logical' to swap 'right' to go down :-)

ThomasChiroux:
(problems in docstring, working on it)

libqtile/layout/xmonad.py
flacjacket: Ah, well, there's no guarantee everything in the current code base is a good idea :)

flacjacket:
Ah, you even added the unittests! Thanks, I'll pull this in!

Issue #1015, opened 2017-03-03
Disabling fullscreen bug


frostidaho:
Using xfce4-terminal I run into a bug where when fullscreen mode is enabled and then disabled the terminal is resized / placed as fast as possible.

The bug only happens if `auto_fullscreen = True` in the qtile config.

I'll post a video because the bug is difficult to describe.

frostidaho:
Here's a video of the bug
https://youtu.be/RruN6QujPJk

flacjacket:
Closed by #1016

jwijenbergh:
This bug is yet again present on latest master. Just hold down the fullscreen key with `auto_fullscreen` and xfce-terminal and you will see it never stops getting out of the fullscreen off/on dance

PR #1016, opened 2017-03-03
Fix race condition bug in libqtile.window.Window.fullscreen

frostidaho:
This patch fixes issue #1015


flacjacket:
This is good, thanks!

You don't have to open issues before you submit PR's, feel free to just submit the code with the explanation in the PR if that's easier :)

frostidaho:
Sure np

Issue #1017, opened 2017-03-05
Bug in qtile-top


frostidaho:
There are [string formatting issues](https://github.com/qtile/qtile/blob/4e19f46c7eb660f9a761bea9d073fd6fa6929ed6/libqtile/scripts/qtile_top.py#L88) in qtile-top which will always raise an exception

flacjacket:
Closed by #1018

PR #1018, opened 2017-03-05
Fix string formatting in qtile_top.py - issue #1017

frostidaho:
For issue #1017 

https://github.com/qtile/qtile/issues/1017

### DIFF
--- a/libqtile/scripts/qtile_top.py
+++ b/libqtile/scripts/qtile_top.py
@@ -85,7 +85,7 @@ def get_stats(scr, client, group_by='lineno', limit=10, seconds=1.5,
     (max_y, max_x) = scr.getmaxyx()
     curses.init_pair(1, curses.COLOR_GREEN, curses.COLOR_BLACK)
     while True:
-        scr.addstr(0, 0, "Qtile - Top {0:s} lines".format(limit))
+        scr.addstr(0, 0, "Qtile - Top {} lines".format(limit))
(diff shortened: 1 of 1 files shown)


flacjacket:
This is fine, I'm curious what it is that is causing the errors? What do the exceptions look like?

flacjacket:
I mostly ask so when this goes in, we don't re-break it later

libqtile/scripts/qtile_top.py
```diff
@@ -101,22 +101,22 @@ def get_stats(scr, client, group_by='lineno', limit=10, seconds=1.5,
             line = linecache.getline(frame.filename, frame.lineno).strip()
             if line:
                 code = line
-            mem = "{0:.1f} KiB".format(stat.size / 1024)
-            filename = "{0:s}:{1:s}".format(filename, frame.lineno)
-            scr.addstr(cnt + 1, 0, '{0:<3s} {1:<40s} {2:<30s}'.format(index, filename, mem))
+            mem = "{:.1f} KiB".format(stat.size / 1024.0)
```
flacjacket: Ah, we should probably do the standard `__future__` imports for stuff like division.

frostidaho:
It'd always raise a ValueError. e.g.,

```
In [2]: x = 37

In [3]: '{:s}'.format(x)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-bb5f3dab3df6> in <module>()
----> 1 '{:s}'.format(x)

ValueError: Unknown format code 's' for object of type 'int'
```

flacjacket:
Ah, I see, thanks!

PR #1019, opened 2017-03-05
Tasklist: task names may consume as much space as available

dhannes:
Allows task names in `TaskList` widget to consume as much space as is available to the widget.
And on the other hand shrink task name boxes such that all tasks can be displayed.

Since the TaskList is a stretch widget, there may be more space available than the initial `max_title_width` and task names may also contain useful information which may get truncated otherwise. 
If the available space is not sufficient, only task names wider than the average box width are shortened.
See screenshots below for clarification.

This pull request also
- adds a `spacing` widget argument to be used for spacing between task names (instead of margin_x), like for the GroupBox widget.
- adds an `unfocused_border` widget argument for defining a color for unfocused tasks, used with highlight_method 'border' or 'block'.
- changed default of `max_title_width` to None. However, if it is defined, it is obeyed such as before.
Setting `max_title_width` should lead to the same behaviour as without these changes.
- changed window state characters to some special unicode characters. Mainly 'V' for floating gets the 'overlap' character: '🗗'.

If enough space is available all tasks are displayed to its full extend.
tasklist1

If not enough space is available, boxes wider than the computed average are shortened percentually
tasklist2
tasklist3

However, if all task names are wider than the average,
and there are many short task names and few long ones,
the percentually shrinking currently leads to following unexpected (and unwanted) behaviour:
tasklist4

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -63,6 +63,13 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             "hints (one of 'border' or 'text')"
         ),
         (
+            "unfocused_border",
+            None,
+            "Border color for unfocused windows. "
+            "Affects only hightlight_method 'border' and 'block'. "
+            "Defaults to None, which means no special color."
+        ),
+        (
             "max_title_width",
             None,
             "Max size in pixels of task title."
@@ -322,7 +329,8 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
                 border = self.border
                 text_color = border
             else:
-                border = self.background or self.bar.background
+                border = self.unfocused_border or (self.background or
+                                                   self.bar.background)
                 text_color = self.foreground
 
             if self.highlight_method == 'text':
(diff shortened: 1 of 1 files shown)


libqtile/widget/tasklist.py
```diff
@@ -71,22 +89,90 @@ def __init__(self, **config):
...
+        Depending on its state minimized, maximized and floating
+        appropriate characters are prepended.
+        """
+        state = ''
+        if window is None:
+            pass
+        elif window.minimized:
+            state = '\U0001F5D5 '  # minimize character
+        elif window.maximized:
+            state = '\u0001F5d6 '  # maximize character
```
flacjacket: I think this should be uppercase U

flacjacket:
Other than that, I think this looks good, this is definitely something that has bothered me with long browser widow titles, so I'll definitely try it out.

dhannes:
Yes, you are right about the uppercase U for that character.

However, I did not notice any Problems here, because the maximized window hides the widgets.
Is this a meaningfull case, or should we just skip the 'Maximized' ?

flacjacket:
I guess you are right, it doesn't really matter, but you've pushed the fix so I can pull it in regardless :)

PR #1021, opened 2017-03-08
Add {posargs} to tox.ini

frostidaho:
This lets us pass arguments to pytest while using tox.
http://tox.readthedocs.io/en/latest/example/general.html

e.g., we can run just some of the tests:
    tox -e py36 -- -x test/test_manager.py


flacjacket:
Ah, nice, thanks!

PR #1024, opened 2017-03-14
Changes to default configuration

kynikos:
Note: this PR may be a little controversial, I'm proposing the whole thing, but it's likely that we'll end up cherry-picking only some changes :)

d06a18b and 30b8ffa should be quite safe.
In cddb65c I've replaced the default widget font "Arial" with "sans", which I think is a saner default especially on Linux, however I've also reduced its size, and the size of the bar as well, since I think it would look better to the eyes of new users, but this is only based on personal preference, I'm just wondering how many of us share the same idea.
79d08eb is an idea to introduce a better way to share basic float rules which would make the affected windows/dialogs behave in a more expected way especially for new users who are just testing qtile, and who otherwise would see the default behavior as buggy.

### DIFF
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -102,9 +102,9 @@ screens = [
 # Drag floating layouts.
 mouse = [
     Drag([mod], "Button1", lazy.window.set_position_floating(),
-        start=lazy.window.get_position()),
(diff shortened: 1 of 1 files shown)


flacjacket:
I'm +1 for all of these, I will merge them in, thanks!

PR #1025, opened 2017-03-18
move mpris2 dbus initialization to configure()

tych0:
The problem here is that when the widget constructor is called (in the
config file), the manager's main function hasn't been called yet, so it
hasn't set up our proxy dbus loop. This ends up causing:

[0m[33m2017-03-17 20:58:49,810 [1m[33mlibqtile confreader.py:__init__():L67 [0m Caught exception in configuration:

  File "/home/tycho/.local/lib/python3.5/site-packages/libqtile/confreader.py", line 55, in __init__
    config = __import__(os.path.basename(fname)[:-3])
  File "/home/tycho/.qtile/config.py", line 118, in <module>
    **widget_defaults
  File "/home/tycho/.local/lib/python3.5/site-packages/libqtile/widget/mpris2widget.py", line 68, in __init__
    '/org/mpris/MediaPlayer2')
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 148, in add_signal_receiver
    path, **keywords)
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 400, in add_signal_receiver
    self._require_main_loop()
RuntimeError: To make asynchronous calls, receive signals or export objects, D-Bus connections must be attached to a main loop by passing mainloop=... to the constructor or calling dbus.set_default_main_loop(...)

The solution here is to wait until we have actually initialized the main
loop, so GDbusMainLoop does the right thing and all is happy.

### DIFF
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -61,17 +61,27 @@ class Mpris2(base._TextBox):
         base._TextBox.__init__(self, '', **config)
         self.add_defaults(self.__class__.defaults)
 
-        dbus_loop = DBusGMainLoop()
-        bus = dbus.SessionBus(mainloop=dbus_loop)
-        bus.add_signal_receiver(self.update, 'PropertiesChanged',
-            'org.freedesktop.DBus.Properties', self.objname,
-            '/org/mpris/MediaPlayer2')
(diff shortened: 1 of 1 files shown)


flacjacket:
:shipit:

PR #1027, opened 2017-03-19
Clock widget: properly set timezone

kynikos:
I think this is self-explaining: setting a Clock in a specific timezone was still showing the local time, this fixes the problem for me.


flacjacket:
Nice quick fix

PR #1030, opened 2017-03-28
Rework of Layout classes to fix window focus cycle.

### LINKED ISSUE #1023: window focus cycling (different for certain layouts)
Hi all,
I just stumbled about different behaviour for `cmd_next_window`, `cmd_prev_window` focus cycling of `libqtile.group._Group` for different layouts.

I am using `Key( [ 'mod4' ], 'Tab', lazy.group.next_window() )` in config.py of qtile 0.10.7.
Depending on the current Layout of the group not all windows are cycled or the cycling stops.

For example
- in _Columns_ layout only windows in the current column and floating windows are cycled.
- for _wmii_ the cycling stops at end of windows in layout. No floating windows are focused.
- other layouts enable the group to cycle through all managed windows. (although i did not consider all layouts yet)

If this behaviour is considered a defect, i am willing to fix it, according to comments on #972 i already found. This seem to require changes in `focus_next`, `focus_previous` of certain layouts.

dhannes:
This PR refactors common code of layouts to two new classes in 'libqtile/layout/base'.
This PR fixes #1023.

During fixing I saw a lot of similar or even copied code, thus I tried to rework that common code, adding two new base classes `ClientList` and `SimpleLayoutBase`.
'ClientList' is all about maintaining a list of clients and a currently active/focused client.
It adds methods for arranging (add, remove, shuffle, ...) clients in the list and implements methdos for the window focus cycle.
It is used in a new base class for layouts SimpleLayoutBase, which is now base class for all layouts except Columns and Stack. However Stack and Columns use 'ClientList' as base for their internal helper clases to managing the stacks. 

All existing layout tests passed without modification.
Tests were added to assert the desired focus cycle.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -415,6 +415,13 @@ class _Group(command.CommandObject):
         self.layoutAll()
 
     def cmd_next_window(self):
+        """
+        Focus the next window in group.
+
+        Method cycles _all_ windows in group regardless if tiled in current
+        layout or floating. Cycling of tiled and floating windows is not mixed.
+        The cycling order depends on the current Layout.
+        """
         if not self.windows:
             return
         if self.currentWindow.floating:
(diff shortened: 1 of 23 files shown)


dhannes:
Uhps, I'm sorry for the failed tests.
I just tested the layout subset, which is pretty negligent.
Seems there is still work to do.

ramnes:
I like what I see here!

flacjacket:
Wow, added docs and tests, and still almost even added/removed lines. This is burning down technical debt if I've ever seen it.

PR #1033, opened 2017-04-04
Implement the 'client_name_updated' hook

frostidaho:


dhannes:
(I guess you implemented it because you need the client instance in subscribed method and `window_name_change` has no arguments)
But isn't it confusing having two separate hooks fired for the same action to react on?

frostidaho:
That's right - I want at the client's Window instance.
Yeah it is kinda silly to have two hooks. But for whatever reason this hook was already in the codebase, just not ever used.

If we add this argument to `window_name_change` we could break people's code if their callbacks aren't built with `*args`

If we just replace `window_name_change` with `client_name_updated` you could also have breakage with people's configs.

frostidaho:
I suppose the cleanest way would be to deprecate `window_name_change` and switch everything internally over to `client_name_updated`. But we'd still be left with both hooks for some time.

flacjacket:
I'd be fine to deprecate it. You could stick with the `window_name_change`, wrap the call with the args in a try/except, and in the except fire the deprecation warning and call it without args.

Tho we have that hook in the code already and never have it fire, so it would be fine to actually implement it.

flacjacket:
This is good, and thanks for updating all the `window_name_change` uses in the main codebase, too.

Issue #1035, opened 2017-04-04
obscure unicode characters in tasklist widget


frostidaho:
![printing these characters in the terminal](https://cloud.githubusercontent.com/assets/8061555/24673705/3ac39ce4-1947-11e7-81ce-0b9a4e42ba12.png)

The tasklist widget was recently changed and the window status symbols are now some uncommon unicode characters. It just shows up as a space with my fonts (dejavu sans iirc). It should probably default to using ascii characters and have it be a customizable option.

old code:
```python
elif w.maximized:		
    state = '[] '		
elif w.minimized:		
    state = '_ '		
elif w.floating:		
    state = 'V '
```

new code:
```python
elif window.minimized:
    state = '\U0001F5D5 '  # minimize character
elif window.maximized:
    state = '\U0001F5d6 '  # maximize character
elif window.floating:
    state = '\U0001F5D7 '  # overlap character
```

flacjacket:
Fixed in #1036

PR #1036, opened 2017-04-04
Make floating, minimized, and maximized window states configurable

frostidaho:
For issue #1035

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -84,17 +84,20 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
         (
             'txt_minimized',
             '_ ',
-            'Text representation of the minimized window state.'
+            'Text representation of the minimized window state. '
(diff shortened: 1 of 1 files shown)


dhannes:
It was changed by me in #1019.
I thought of ascii as too much last century. ;)
However, making it configurable is the best way to get around this issue and personal preferences.

flacjacket:
I think it's fine to have this configurable, but it might be good to have the Unicode codes somewhere in the docstring if you have it configurable like this so people know what would be a good set of utf-8 defaults. You might also make it as an `"ascii"`/`"unicode"` toggleable parameter if you think that would be ok.

frostidaho:
Okay sounds good, I updated the docstrings with examples including U0001F5D5..7. I'd prefer to leave it user-configurable as opposed to a switch, but it doesn't really matter to me.

flacjacket:
Thanks for the changes, I'll push this in.

Issue #971, opened 2016-10-29
Cannot set initial layout for Group


emesik:
I have config file where `Group` gets a layout assigned on the start:

```
g = Group('a', layout='VerticalTile')
groups.append(g)
```

qtile crashes on start with exception:

```
Traceback (most recent call last):
  File "/usr/bin/qtile", line 9, in <module>
    load_entry_point('qtile==0.10.6', 'console_scripts', 'qtile')()
  File "/usr/lib/python2.7/site-packages/libqtile/scripts/qtile.py", line 125, in main
    q = make_qtile()
  File "/usr/lib/python2.7/site-packages/libqtile/scripts/qtile.py", line 116, in make_qtile
    state=options.state,
  File "/usr/lib/python2.7/site-packages/libqtile/manager.py", line 160, in __init__
    self.dgroups = DGroups(self, self.config.groups, key_binder)
  File "/usr/lib/python2.7/site-packages/libqtile/dgroups.py", line 94, in __init__
    self._setup_groups()
  File "/usr/lib/python2.7/site-packages/libqtile/dgroups.py", line 127, in _setup_groups
    self.add_dgroup(group, group.init)
  File "/usr/lib/python2.7/site-packages/libqtile/dgroups.py", line 123, in add_dgroup
    self.qtile.addGroup(group.name, group.layout, group.layouts)
  File "/usr/lib/python2.7/site-packages/libqtile/manager.py", line 466, in addGroup
    g._configure(layouts, self.config.floating_layout, self)
  File "/usr/lib/python2.7/site-packages/libqtile/group.py", line 71, in _configure
    self.layout = self.customLayout
  File "/usr/lib/python2.7/site-packages/libqtile/group.py", line 124, in layout
    raise ValueError("No such layout: %s" % layout)
ValueError: No such layout: VerticalTile
```

The same happens when I use `xmonad-tall` or any other layout name.

BTW, perhaps it would be better to pass a `Layout` object here than a name which is hard to find anywhere in the docs.

and3rson:
@emesik You can still check out the source of `VerticalTile` and use the `name` attribute of the class. Another way to do this is to set custom name for a layout in `layouts=[]` definition and use it in your group config.

dhannes:
The default name of layouts is always the class name in lower case.
see:  qtile/libqtile/layout/base.py (Layout.__init__, line 44ff)

You need to either specify a name for the Layout or use the class name in lower case.

[...]

Issue #953, opened 2016-09-11
Separate layout on each group ?


Hammer2900:
And this moment, how i make separate Layout for each group ? or i make only one layout for all groups.
And how switch layout by hot keys ?

flacjacket:
At this point, you could set a particular layout to be on a group, but at a quick glace, I don't see a way to set a particular to a layout on one and only one group.

flacjacket:
As for switching layouts, the easiest thing to do is use the `simple_key_binder` [1], which will generate keybindings to numerical order of the group as `mod + N` to move to the Nth group, but you can also do a custom key binding, as is done in the default config [2] to bind to the letter of each group.

[1] http://docs.qtile.org/en/latest/manual/config/groups.html#example
[2] https://github.com/qtile/qtile/blob/develop/libqtile/resources/default_config.py#L87

dhannes:
According to http://docs.qtile.org/en/latest/manual/config/groups.html#group it is possible to define `Group` argument `layouts` which is "_the group layouts list overriding global layouts_ ".
Thus it is possible to define seperate layouts for single or all groups, and its also documented and working (I tested it).
I guess we can close this.

PR #1042, opened 2017-04-13
Minor docu upgrade about groups and layout default names

dhannes:
This PR addresses #971 and #953

1. slightly enhanced group docstrings about layouts and layout
arguments.
2. Default names of layout is class name in all lower case.
corrected name in layout defaults were defined otherwise,
because never had any effects, except for documentation.


flacjacket:
Great, thanks for the clarifications

PR #1043, opened 2017-04-18
Add net_wm_state descriptors to xcbq.Window and refactor tests

frostidaho:
So I was working on some code for having sticky window support in qtile (_NET_WM_STATE_STICKY) and this patch is in support of that. It lets you set the net_wm_state on the xcbq.Window instance directly. 
e.g.,
```python
window.net_wm_state_sticky = True
window.net_wm_state_modal = True
```
This patch also makes changes to tests and moves all of the sleep and retry code into a central object `conftest.retry`.

The changes to the tests were done, because they were kind of flakey for me in that they'd often fail with what seemed to be a sort of race condition. This is also what prompted me to include the test dependency of xvfbwrapper. xvfbwrapper can be eliminated with a bit more code.

Tests:
Refactor retrying unreliable functions in test-suite
Remove sleep call from test_bar.test_textbox
Replace conftest._find_display with version from xvfbwrapper
Add test_xcbq.py

libqtile/xcbq.py:
Remove xcbq.GC and related code. It was not used anywhere
Add descriptors to xcbq.Window for net_wm_state flags
e.g., `window.net_wm_state_sticky = True`

libqtile/extension/window_list.py:
Remove unused re import


flacjacket:
Ah, this is a nice change, thanks!

PR #1048, opened 2017-04-24
Add common layout tests + respective fixes (+ other improvements)

kynikos:
I've tried to organize the commit history in a logical way to make them easy to understand, including their messages.

The main addition is a new series of tests that apply to all layouts, to ensure consistent behavior in some common scenarios (9f825f394127931bcc3e4385e9f6a80cbb104e21). This revealed some bugs for which I also provide fixes.

This has also given me the opportunity to add some useful commands for testing, and especially a script to generate custom windows using tkinter (f04ab863aeacb0513a8b1eb6c06d805fd9595306).

All this was inspired by a bug, which is not tested directly, that I discovered because I use the Max and Wmii layouts in the same group: I noticed that when opening a window in the Max layout and then closing it, the previously focused window wouldn't be refocused. This was happening because in the background Wmii was reacting to the window closing by firing its own (redundant) focus_change event, which was in turn influencing back the Max layout and changing the focused window. The number of fired focus_change events is now tested (9678a1b3993fc231a3148326519b8d9a4b0fa250), which could be also expanded to count other events, and the fix for Wmii is 892c0ffe21876965708c8271636f6f5d8fd4a613.

I've also slightly tweaked the behavior of Wmii, which now spawns new windows under the current one, and not at the bottom of the stacks (34d5b85fc72ef7e902006e6b8834dcbd4e3ae704).

Totally unrelated, but I've also bundled a small and safe improvement to the WindowList extension (c85cde4e8ec085ac98c9d42fb535f2061fde369a).


kynikos:
(I only tested on Python 3, so I've had to repush some commits to fix some Python 2 issues which has required editing the commit references in my initial post)

flacjacket:
Ah, tests on all of the layouts!

Thanks for doing this, this is great to have!

kynikos:
Thank you for reviewing this :) but can I ask why you left out the first 3 commits 375ea677d3160ff26d4352882e04d5d4df19380b, c85cde4e8ec085ac98c9d42fb535f2061fde369a and 751f6584ddbca2d69ee6a0203424dbd5e7a09e8c?

flacjacket:
Oops! My mistake, a cherry-pick gone bad.

PR #1051, opened 2017-04-25
Rewrite xcbq.Connection.refresh_modmap

frostidaho:
Combine xcbq.ModMasks and xcbq.ModMapOrder into one object


flacjacket:
Nice, thank!

PR #1052, opened 2017-04-25
Fix focused window after making a window floating and closing it

infmagic2047:
With this patch, after making a window floating, the window returned by
layout.remove() will be focused after closing the last floating window
instead of the window focused before it.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -302,25 +302,36 @@ class _Group(command.CommandObject):
             self.layoutAll()
 
     def mark_floating(self, win, floating):
+        hadfocus = win == self.currentWindow
+        nextfocus = None
+
         if floating:
             if win in self.floating_layout.find_clients(self):
                 # already floating
                 pass
             else:
                 for i in self.layouts:
-                    i.remove(win)
+                    if i is self.layout:
+                        nextfocus = i.remove(win)
+                    else:
+                        i.remove(win)
                     if win is self.currentWindow:
                         i.blur()
                 self.floating_layout.add(win)
                 if win is self.currentWindow:
                     self.floating_layout.focus(win)
         else:
-            self.floating_layout.remove(win)
+            nextfocus = self.floating_layout.remove(win)
             self.floating_layout.blur()
             for i in self.layouts:
                 i.add(win)
                 if win is self.currentWindow:
                     i.focus(win)
+
+        if hadfocus and nextfocus:
+            # store nextfocus in focus history, but don't actually focus it
+            self.currentWindow = nextfocus
+            self.currentWindow = win
         self.layoutAll()
 
     def _items(self, name):


kynikos:
Hi, I have another open PR which adds some tests that fail with this patch: can you please write some clear example steps to reproduce the unwanted behavior fixed by this patch, and clarify the expected result?

kynikos:
Thanks, I see the problem, and of course the final decision to merge this patch is up to the repo maintainers, however *from my point of view* in the second scenario I would expect `b` to be focused, since `c` would have become a normal floating window, just like any other (e.g. dialogs), so closing it should influence the underlying layout like a normal floating window, not as if it was still tiled. I think that inserting/moving a window in the `focusHistory` list without actually ever focusing it is a slightly hackish solution that may generate unexpected behavior somewhere else in the future, since contributors would expect that list to reflect the actual focus history, and just switching a tiled window to floating wouldn't be expected to change the focus order.

If the maintaners will merge this, I won't mind updating my tests in the other PR, however I was for example also proposing a `group.focus_back()` function to change the focus to the previously focused window, which would now be broken when toggling a window to float. Also, if this is merged, I would expand the comment at [line 332](https://github.com/qtile/qtile/pull/1052/files#diff-9b4be58b2abb6785913354cc9d1ed7ccR332) of the new `group.py` to better explain why `focusHistory` is hacked like that.

dhannes:
1. I agree to @kynikos in that the fix proposed here is quite hackish and adresses a corner case of making windows floating and kill em afterwards.
2. Furthermore i assume that the first described behaviour is the erroneous one.
--> Killing a window should focus the last focused one, regardless of it was floating or not.

The window returned by `layout.remove` is the window that should be focused from the layouts point of view and depends mainly on the position of windows in the layout (Most often the next or previous in a internal list).
However the `Group` maintains a `focusHistory` which is also used in `Group.remove` to focus the last focused window from group.
I assume that the problem is different to what shall be fixed in this PR.

flacjacket:
Hmm, I agree with the problem, but the feel like there has to be a cleaner way to fix this.

I merged @kynikos PR that added all of the tests, so you'll need to rebase on that and change any tests that fail there. At least then, we can be sure what behavior is changing (at least within the set of tests that were added).

infmagic2047:
Well, making a window floating can be interpreted as "remove it from tiling layout, add it to floating layout, and focus it if it was the current one". The return value of `group.remove()` in the first step should be respected.

And this patch is more important for layouts where not all windows are shown (like fullscreen layouts). The layout may decide what will be focused next, and draw windows according to that. Without this patch the displayed windows may change after closing a floating window.

ramnes:
Based on your two examples, I think we should always focus `b`, not `d`. I feel it would be really cumbersome when using something like LibreOffice or Gimp, that often has "Yes/No" dialogs, and where you would expect to come back on the main application afterwards and not on the next window in your list.

infmagic2047:
@ramnes Your example will continue to work after this patch, because the dialog is created as floating window, and `layout.remove()` is not called here.

kynikos:
Just to clarify my position, in the case of closing a *tiled* window I am one who would expect `d` to be focused in the example above; in general, I'd leave the choice to each layout as it is now (maybe we should make this user-configurable?). About closing a window that has been changed to floating and never lost focus, I like that the focus is returned simply to the previously focused window, like when closing a regular dialog, however I'd also be ok if @infmagic2047's new behavior was implemented; my only problem is that I wouldn't like `focusHistory` to be changed at all by making a tiled window floating or vice versa.

@infmagic2047 Can you make an example, or even better push a proper test, which shows the current unwanted behavior of layouts that don't show all windows when you close a window that has just been floated?

I think a way to implement the proposed behavior that I'd see as equally hacky, but I'd probably be fine with, would be to change your patch at line 333, and store `nextfocus` in a new special `group` attribute, something like `float_close_next_focus`, which should be initialized to `None` in `_Group`'s constructor with a big comment about how it works and why; when `_Group.remove()` is called, `nextfocus` is assigned as `nextfocus or self.float_close_next_focus or self.currentWindow or ...`; when `_Group.focus()` is called, `float_close_next_focus` should be set back to `None` if the focused window is different from `currentWindow`. Finally all this should be properly and thoroughly tested.

dhannes:
This PR "fixes" the focushistorie in the moment a window is switched to floating layout.
But the unwanted behaviour is about closing windows (in Group.remove).

Imho the fix should be done in `Group.remove()`!
Note that all Layouts derived from `_SimpleLayoutBase` use the `_ClientList` collection, which maintains a `current_client`. That `current_client` is also maintained in `Layout.remove()` (and returned by it).
Thus it is possible in `Group.remove` to ask the current tiling Layout about that current_client and focus that instead of the last in focusHistorie. For `Columns` and `Stack` there is something similar and unified access methods can be easily provided.

However, I don't insist on any of the proposed behaviours.
But i would like to inist that not a symptom is fixed but the problem in a general manner.

infmagic2047:
@dhannes I agree with you about asking layout for the current window, but then there will be two ways to indicate what to focus after removing one window from the layout. Maybe we should use this method in `group.remove` as well.

I noticed that `_ClientList` is not returning anything in `remove()`, so most layouts derived from `_SimpleLayoutBase` don't either. This should be fixed.

ramnes:
Hello @infmagic2047 @kynikos @dhannes! Let us try to find a solution to the issues raised by this PR, so that we can move on to something else. :slightly_smiling_face: 

I feel that the simplest solution would be to always focus the previously focused window (at the group level) when closing a window. Do you see any reason not to?

ramnes:
This PR has been closed automatically because we removed `develop`, which was targeted by this branch. Please continue the discussion on #1333 if needed.

PR #1053, opened 2017-04-26
Remove old compatibility code for cffi

frostidaho:
This patch removes code for cffi version < 1.0, since version 1.0 was released in the middle of 2015.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,13 @@ Features
       unit-tested window mangers around.
 """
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -80,25 +80,7 @@ def finalize_options(self):
       unit-tested window mangers around.
 """
 
-if '_cffi_backend' in sys.builtin_module_names:
```
flacjacket: I think that this part should stay, this is more for PyPy to make sure that we don't try to upgrade the cffi version (since cffi is bundled in PyPy)

frostidaho: oh good catch

flacjacket:
Thanks for the quick updates!

PR #1055, opened 2017-04-27
Battery widget: let users decide which battery to be monitored

rosecrown:
Batteries of some laptops (like the one I bought) are not named as BAT0 (mine is BAT1), thus it should be better if users can select the battery themselves.


frostidaho:
Why can't you just call `Battery(battery_name='BAT1')` as it stands right now?

rosecrown:
Sorry, I should read the doc first.

PR #1056, opened 2017-04-27
Simplify confreader.py & config fallback

frostidaho:
* Simplify confreader.py
* Make qtile always fallback to the default config when it raises an exception in loading the config file.

Reasoning:
It's frustrating when using a display manager and you run into an issue with your config file.
It instantly kicks you back to the greeter, and you have no information as to why. You then have to choose another wm or tty to debug, login again, and then look up the logfile saved to disk. Then you make your change, and if it doesn't fix the issue you have to redo these steps.


flacjacket:
Nice change, thanks!

PR #1057, opened 2017-04-28
Modify test assertions for killing windows

frostidaho:
It now tests to see if the wid is present,
instead of counting the number of windows.


flacjacket:
Great, thanks!

PR #1058, opened 2017-04-28
Logging tweaks

frostidaho:
This patch is a couple small changes to logging. It speeds up running tests by a bit.
I did a couple runs of `tox -e py36` and `tox -e py27-trollius` and measured the time. These commits cut off 10-20% of the runtime.

* Remove existing handlers to root logger when calling init_log

During tests init_log ends up being called multiple times
and we end up with n handlers added to the root logger.

* Turn off colorized logs in test suite
* Stop the logger from writing to disk in test_hook.hook_fixture


frostidaho:
figure_2

flacjacket:
Ah, cool, thanks for figuring this out

PR #1060, opened 2017-04-28
Add scratchpad

dhannes:
Add simple scratchpad behaviour to be used in configuration.
Module provides DropDown class to wrap a window of a spawned
process, whose visibilty can be toggled by keystoke afterwards.

Adds code, test and docs

### DIFF
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -88,6 +88,7 @@ mix!)
     mouse
     screens
     hooks
+    scratchpad
 
 In addition to the above variables, there are several other boolean
 configuration variables that control specific aspects of Qtile's behavior:
--- a/docs/manual/config/scratchpad.rst
+++ b/docs/manual/config/scratchpad.rst
@@ -1,0 +1,51 @@
+==========
+Scratchpad
+==========
+
+By the use of the :mod:`~libqtile.scratchpad` module it is possible to define
+some simple scratchpad like behaviour, in your config.py file.
+Instantiate a :class:`~libqtile.scratchpad.DropDown` and define a command to
+execute and bind the `toggle` property (which returns a `lazy.function`)
+to a `Key` in ``keys`` for example:
+
+.. code-block:: python
+
+    from libqtile import scratchpad
+   
+    dropdown_terminal = scratchpad.DropDown('xterm')
+   
+    keys = [ Key( [], 'F12', dropdown_terminal.toggle), ]
+
+The above example creates a DropDown object whose visibility is toggled by
+pressing F12 key. If the DropDown is set to visible the first time, it spawns
+a process by the defined command (here: xterm). The visibility of the
+corresponding window can afterwards be toggled by the same key.
+
+The DropDown appears always on the current screen if toggled to visible,
+regardless in what group it was before. The DropDown can also be configured t
+to vanish if another window is focused.
+The position and opacity of the corresponding window can be configured as argument
+to __init__, and the window is always placed accordingly.
+
+Note that the DropDown window is always shown as floating window and that the
+window is detached from the DropDown if floating state is toggled.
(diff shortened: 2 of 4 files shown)


frostidaho:
I think this'll be useful

dhannes:
I need some help on the tests, cause the failed checks are for the scratchpad.DropDown itself.
_Background:_
The DropDown uses `cmd_spawn` to spawn a process, by given command string.
Thus in tests i can't use any of the testwindow functions in DropDown, instead i configured the DropDown to spawn a xterm on a certain display by the -display option to xterm.
This might not be the best way but at least it work on my local machine.
In Travis CI build these tests are failed, because the expected window did not appear.

From the log I can't figure out what the problem is...
- no xterm available ?
- wrong assumption about the spawning behaviour?

Any help, suggestion or advice welcome.

frostidaho:
How are you running the tests on your machine? The travis-ci config ultimately runs `tox` to run the test-suite.
Does `tox -e py36 -- test/test_scratchpad.py` work for you? It fails on my machine with the same [message](https://travis-ci.org/qtile/qtile/jobs/226991862#L725) that is given by travis-ci.

FWIW I've found it helpful to sign up for travis-ci. After you sign up you can login and enable your fork of the qtile repository, so that every push you do to your repo will be tested on travis-ci.

frostidaho:
I've had some time to read the code, and I think it shouldn't be too hard to fix
the tests as it stands.

* the x-display is set by the xephyr fixture and then given to the qtile instance, so this section seems like it won't work right `[...scratchpad.DropDown('xterm -display :%i' % d, **kwargs).toggle) for d in range(10)]`

* try using the `conftest.retry` object instead of writing loops and sleeping for retrying unreliable code

However, it'd be easier to test and possibly more broadly usable if scratchpad.py was restructured a bit. Right now the `DropDown` class does two main things: it controls an xwindow and spawns a process.

I think it should be broken up into two+ pieces. The first piece would run the code for controlling the window, and its constructor would take an instance of `window.Window` or `xcbq.Window` plus whatever other options. This part could more easily be tested, and it'd be easier to use elsewhere in the codebase.
The second piece would be something that produces a `window.Window` or `xcbq.Window` object from some command.

dhannes:
Great thanks for looking at code and test.

According to the docs i just used 'pytest test/test_scratchpad.py'
Now, using tox I get the same errors.

The loop you mentioned and its usage is of course the crucial part.
However it just creates 10 DropDown objects, each would spawn on a seperate display and is bound to the appropriate key ('control'+'1' for display ':1', and so forth up to 9)
The tests simulate the appropriate keypress determined from the display attribute of the conftest.Qtile instance. Thus, for a test only one DropDown configuration is used, dpending on the display.

Reworking the test to use conftest.retry does not change the test behaviour but the test are more closely to the testwindow functions. Thanks for the hint.

Now the main difference is on how the xterm instance is spawned (for example on display ":1"). 
For conftest.Qtile.testXterm it is like:
`subprocess.Popen('xterm', env={"DISPLAY": ":1"})`
for the DropDown object like
`_cmd_spawn("xterm -display :1")`
where _cmd_spawn uses os.execvp and some _magic_.

This kind of problem is currently beyond my capabilities (and interest).
However, if someone as a clue or solution on this, please provide.

I will focus on frostidaho's other hint to separate the two phases of spawning a process and handling the visibility of its window. This is good idea, regardless of my current problem with test.

frostidaho:
I'll be happy to help you work on the tests.

One potentially important thing to know about launching the xservers for the tests is that the assigned display is essentially a [random 32bit integer](https://github.com/cgoldberg/xvfbwrapper/blob/4b6789ffa2d43772a83d9e18a60663a1c3aa6c33/xvfbwrapper.py#L128). So this display number will most likely be much larger than 10. I changed this recently, so it's probably my fault it broke!

PR #1061, opened 2017-04-30
Update mouse.rst (tiny import mistake)

Lesik:
`EzClick` is imported as `EzClick`, but later used as `Click`.


flacjacket:
Great, thanks for the fix!

Issue #1014, opened 2017-02-28
svg support branch


frostidaho:
## SVG support branch

I've created a svg support for some widgets and 
refactored how images are loaded for widgets in general.

Qtile's widgets deal with images by creating cairo `cairocffi.ImageSurface` and `cairocffi.SurfacePattern` objects which correspond to some image.

This branch refactors loading and resizing images into [libqtile/images.py](https://github.com/frostidaho/qtile/blob/b0f929f76d43d6e23ee5ca893d695165677da53e/libqtile/images.py). The `libqtile.images.Loader` class can load png and optionally svg files.
For svg support you need to have [cairosvg](https://pypi.python.org/pypi/CairoSVG) installed.

The [develop_svg_support](https://github.com/frostidaho/qtile/tree/develop_svg_support) branch also has some new / slightly modified widgets.

[VolumeImg](https://github.com/frostidaho/qtile/blob/develop_svg_support/libqtile/widget/volumeimg.py) is a
widget in this branch which shows how `libqtile.images.Loader` is used. There are also a number of tests
which show how `Loader` works.

* [example bar with png & svg icons](http://i.imgur.com/fkSWRQM.png)
* [example bar code](https://gist.github.com/frostidaho/88778b1926dc44f14c309a01cb849777)

If there is interest in it, I'd be happy to clean it up.

flacjacket:
I think adding support for svgs in widgets would be really great. I have hacked in svg support into some things for myself, but it was really rough, lots of manual svg drawing, and I eventually gave up and found fontawesome icons that did what I needed. This would definitely open up new avenues for building widgets. There are even things that we already have like the layout icons that are png and would be great in svg. I'd definitely be +1

elParaguayo:
SVG images are supported so I think this should be closed.

PR #1064, opened 2017-05-02
Add support for other image formats (svg, jpeg, ...)

frostidaho:
This patch depends on the cairocffi bindings of GDK-PixBuf.

libqtile/images.py::Img is a class for loading and transforming images.
libqtile/widget/image.py::Image has been updated to use the `Img()` class, and can now display svgs.

Other widgets still need to be updated to use `images.Img` in order to display svgs or jpegs.
This code is a bit simpler than the code described in https://github.com/qtile/qtile/issues/1014, and
no new widgets have been added.


frostidaho:
This PR has an issue with properly scaling svgs. If [this patch](https://github.com/Kozea/cairocffi/pull/98) gets into cairocffi, I'm planning on updating this PR like in [feature_svg3](https://github.com/frostidaho/qtile/tree/feature_svg3).

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
+from __future__ import division
+import cairocffi
+import cairocffi.pixbuf
+import io
+import os
+from collections import namedtuple as _namedtuple
```
flacjacket: I don't think you need to import as `_namedtuple`, just import `namedtuple`

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
+from __future__ import division
```
flacjacket: You should add the license notice at the top of the file.

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if width and height are not None scale the pattern
+    to be size width and height.
+
+    theta is in degrees ccw
+    """
+    EPSILON = 1.0e-6
+    from math import pi
+
+    if surface is None:
```
flacjacket: When should this be None? Should this log an error somewhere earlier

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
...
+        return None
+    pattern = cairocffi.SurfacePattern(surface)
+    pattern.set_filter(cairocffi.FILTER_BEST)
+    matrix = cairocffi.Matrix()
+    # TODO cleanup this function
+    tr_width, tr_height = None, None
```
flacjacket: Why not just set them to `1.0` here?

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
...
+        name, file_type = os.path.splitext(name)
+        file_type = file_type.lstrip('.')
+        return cls(bytes_img, name=name, path=image_path)
+
+    @property
+    def file_type(self):
+        try:
+            return self._file_type
+        except AttributeError:
+            self.surface
```
flacjacket: This seems like a bad code smell to have all these properties with side effects.

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
...
+        self.path = path
+
+    @classmethod
+    def from_path(cls, image_path):
+        "Create an Img instance from image_path"
+        with open(image_path, 'rb') as fobj:
+            bytes_img = fobj.read()
+        name = os.path.basename(image_path)
+        name, file_type = os.path.splitext(name)
+        file_type = file_type.lstrip('.')
```
flacjacket: This isn't used anywhere.

libqtile/widget/image.py
```diff
@@ -58,45 +58,26 @@ def _configure(self, qtile, bar):
             raise ValueError("Filename not set!")
 
         self.filename = os.path.expanduser(self.filename)
```
flacjacket: I feel like there should be a check that the file exists and log it and exit nicely if it doesn't, I don't know if that should be in the `Image.from_path`, or before that.

libqtile/images.py
```diff
@@ -0,0 +1,205 @@
...
+            path=self.path,
+            theta=self.theta,
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        s0 = set((self.bytes_img, self.theta, self.width, self.height))
+        s1 = set((other.bytes_img, other.theta, other.width, other.height))
+        return s0 == s1
```
flacjacket: Why are you putting these into sets?

frostidaho: Woops!

libqtile/images.py
frostidaho: Mmm not sure I'll check. I took this function from another branch where it might've been None.

libqtile/widget/image.py
frostidaho: That makes sense. For this widget I really just tried to change the minimum amount from the baseline to get it switched over to use `images.Img()`, so that it can maybe more easily be used as an example for modifying the other widgets.

frostidaho:
So that cairocffi patch got accepted, so I'm going to close this PR and open a PR that will support the current version of cairocffi and the patched version.

PR #1065, opened 2017-05-04
add ratio of disk usage to df widget.

himaaaatti:
This request makes df widget be like image.
disk

### DIFF
--- a/libqtile/widget/df.py
+++ b/libqtile/widget/df.py
@@ -19,6 +19,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from __future__ import division
 import os
 from . import base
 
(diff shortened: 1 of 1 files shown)


libqtile/widget/df.py
```diff
@@ -34,9 +34,9 @@ class DF(base.ThreadedPollText):
         ('warn_space', 2, 'Warning space in scale defined by the ``measure`` option.'),
         ('visible_on_warn', True, 'Only display if warning'),
         ('measure', "G", "Measurement (G, M, B)"),
-        ('format', '{p} ({uf}{m})',
+        ('format', '{p} ({uf}{m}|{r:.0f}%)',
             'String format (p: partition, s: size, '
-            'f: free space, uf: user free space, m: measure)'),
+            'f: free space, uf: user free space, m: measure) r: ratio (uf/s)'),
```
flacjacket: The close paren after `measure` should be the furthest right

libqtile/widget/df.py
```diff
@@ -68,6 +68,7 @@ def poll(self):
             text = ""
         else:
             text = self.format.format(p=self.partition, s=size, f=free,
-                    uf=self.user_free, m=self.measure)
+                    uf=self.user_free, m=self.measure,
+                    r=(size - self.user_free) / size * 100)
```
flacjacket: You should `from __future__ import division` so this always works ok.

himaaaatti:
I fixed those!

flacjacket:
Thanks for the fixes!

PR #1066, opened 2017-05-05
Add support for other image formats (svg, jpeg, ...)

frostidaho:
This PR is a continuation of [this issue](https://github.com/qtile/qtile/issues/1014) and [this PR](https://github.com/qtile/qtile/pull/1064). I've tried to address the concerns raised by @flacjacket in the previous pr, but I'm sure I've missed some. The rotation feature `Img.theta` is only partially working atm, so that'll need to be fixed or axed. If no one wants rotating an image atm, I'm inclined to just drop that part of the code, and reintroduce it later on (if there's a use case).

The graphic showing example usage was made with [this script](https://gist.github.com/frostidaho/1391da4427e54bafe1896fa485c4656e).
example Img usage

* libqtile/images.py::Img is a class for loading and transforming images.
* libqtile/widget/image.py::Image has been updated to use the
* Img() class, and can now display svgs.


flacjacket:
Nice thanks for this! I can take a look at it, but I'd like to hold off on merging this until cairocffi bumps so we don't have to depend on their git version.

frostidaho:
Yeah I agree

ramnes:
@frostidaho @flacjacket cairocffi 0.9.0 was released in august. Can we move forward here? :slightly_smiling_face:

Issue #1067, opened 2017-05-05
qtile restart not releasing resources ?


blakebd:
Hi,

it seems restart code is somehow broken, as none of created objects bar, widgets, groups gets released and `finalize` methods are not called.

I'm still not sure what is the cause, but so far tracked following

``` python
class Qtile(command.CommandObject):
    """This object is the `root` of the command graph"""

    def stop(self):
        logger.info('Stopping eventloop')
        self._eventloop.stop()
    # ^^^ invoked from cmd_execute after restart

    def loop(self):
        self.server.start()
        try:
            self._eventloop.run_forever()
        finally:
            self.finalize()
        # ^^^ finally block not reached after eventloop.stop from above stop method
```

Calling `.stop()` on asyncio eventloop object *should* stop `run_forever` loop and finally block *should* be performed, however it does happen only at quit, not restart.

Each restart causes resource leaks, not mention mangling of old and newly configured objects.
Figured out by testing my new widget, which spawns external program and acquires i/o pipes causing deadlocks in the older instances.

frostidaho:
In looking into it I found the same behavior. The `finalize()` method wasn't being called. I wasn't able to figure out exactly why, but I did notice that after calling  `_eventloop.stop()` the loop wouldn't seem to die. I think it has something to do with the GLib loop/thread, but I'm not sure.

In any event the patch makes it so `finalize()` is also called for restarts.

flacjacket:
I merged @frostidaho's PR, so this should be fixed now!

blakebd:
Thanks, good job !
Running with latest commit and it seems the issue is gone.

Only left a small issue with qshell, where restart command throws a RuntimeError exception in SelectorSocketTransport.write method: `RuntimeError: Cannot call write() after write_eof()`. Anyway, it happens after all finalize methods were already called and does not affect main way of manager restart through `manager.Qtile`.

Happy hacking ;)

PR #1068, opened 2017-05-05
Ensure that Qtile.finalize() is called on restart

frostidaho:
For issue #1067


frostidaho:
The reason I deleted `cmd_execute(self, cmd, args)` was because afaik it could only be used for restarting or replacing the qtile process, and we already have `cmd_restart` and `cmd_shutdown`. Do you think it should stay?

Issue #492, opened 2014-08-31
GroupBox - possibility to substitute group names


FrostyX:
Hello everyone,
I want ask about this cosmetic improvements. It would be pretty easy to do this. I can code it and send PR if you want, but there is two ways how to do it and I don't know which one suits you better.

a) Add `label=name` property to `Group` class and let `GroupBox` print it instead of `name`
b) Add dict `labels={}` to `GroupBox`. When `group.name in labels` print `labels[group.name]` instead of `group.name`

Additionally it could support substituting icons.

iderik:
I would love this as long as you can use a font to draw icons. I think its quite common that people using fonts like FontAwesome to draw icons in other applications too, like terminal or editors. Feels more universal than using images. Not sure how many terminals that can draw images.

Masacaai:
> A caveat is that you can not use a glyph twice, as Qtile collapses duplicate groups into a single group (I had originally wanted to use `fa-star` to label all three of my misc groups, but had to settle for the circle variants instead).

Is there any workaround for this? I want to set shape glyphs as my group names but qtile still keeps collapsing identical groups into one.

ramnes:
Not that I know of, so let's revive this issue. I'm good with giving groups a label, or an icon, or anything that people like really.

@tych0, I do understand your concern of confusion between `name` and `label` though. What do you think of introducing an `id` value, that would take the name of the group by default, or any other value if specified, and would be used in places where the name is used today, except for the widget output? Or just rename `name` as `id`, and add `label`?

tych0:
I don't think `s/name/id/` really fixes the confusion I was worried about, though. Fundamentally, if we care about this, we should change group_map into group_list. Anything else will be a hack.

ramnes:
How does that change the problem? We still won't be able to match a group based on its name if two groups have the same.

Masacaai:
I was able to get identical glyphs after all. Turns out 'labels' are already a part of the code. The documentation needs to be updated. I was able to find labels while digging through the source code for the GroupBox class. 

Identical glyphs were previously not possible because the code only used 'names' to both display and identify individual groups. The addition of a 'labels' attribute allows us to keep the names different for identification and set the same label for each group.

tych0:
> How does that change the problem? We still won't be able to match a group based on its name if two groups have the same.

Right, but what you're suggesting is that we don't want groups to be uniquely indexed by name any more. We can't put that in a map if we want this to be true.

[...]

PR #1070, opened 2017-05-10
Support of group display name (label) other than the name

dhannes:
Adds label attribute to Group which is used in GroupBox widget
as a display name.
This this PR the current displayed label can be different from the
group name and changed at runtime.
If the label is the empty string the group is not displayed in GroupBox widget.
If the label is not set, or set to None, the groups name is used as label.
Furthermore the display names must not be unique.

Originally i wanted an easy way to hide groups in GroupBox
(and not the opposite way to define visible groups, which is already possible) and found #492.
This PRs solves my problem and the originally request of #492.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -427,6 +427,7 @@ class Group(object):
 
     These groups can spawn apps, only allow certain Matched windows to be on
     them, hide when they're not in use, etc.
+    Groups are identified by their name.
 
     Parameters
     ==========
@@ -452,11 +453,17 @@ class Group(object):
         is this group alive when qtile starts?
     position : int
         group position
+    label : string
+        the display name of the group.
+        Use this to define a display name other than name of the group.
+        If set to None, the display name is set to the name.
     """
     def __init__(self, name, matches=None, exclusive=False,
                  spawn=None, layout=None, layouts=None, persist=True, init=True,
-                 layout_opts=None, screen_affinity=None, position=MAXSIZE):
+                 layout_opts=None, screen_affinity=None, position=MAXSIZE,
+                 label=None):
         self.name = name
+        self.label = label
         self.exclusive = exclusive
         self.spawn = spawn
         self.layout = layout
--- a/libqtile/extension/window_list.py
+++ b/libqtile/extension/window_list.py
@@ -48,7 +48,7 @@ class WindowList(Dmenu):
         for win in windows:
             if win.group:
                 item = self.item_format.format(
-                    group=win.group.name, id=id, window=win.name)
+                    group=win.group.label or win.group.name, id=id, window=win.name)
                 self.item_to_win[item] = win
                 id += 1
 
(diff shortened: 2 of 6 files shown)


libqtile/config.py
```diff
@@ -427,6 +427,7 @@ class Group(object):
 
     These groups can spawn apps, only allow certain Matched windows to be on
     them, hide when they're not in use, etc.
+    Groups are identified by its name.
```
kynikos: Typo: its name -> their names

libqtile/config.py
```diff
@@ -452,11 +453,17 @@ class Group(object):
         is this group alive when qtile starts?
     position : int
         group position
+    label : string
+        the display name of the group.
+        Use this to define a display name other than the groups name.
```
kynikos: Typo: groups -> group's

libqtile/group.py
```diff
@@ -492,5 +496,14 @@ def cmd_switch_groups(self, name):
         """Switch position of current group with name"""
         self.qtile.cmd_switch_groups(self.name, name)
 
+    def cmd_set_label(self, label):
+        """
+        set the display name of current group to be used in GroupBox widget..
+        If label is None, the groups name is used as display name.
```
kynikos: Typos:
set -> Set
double dot .. -> .
groups -> group's

libqtile/widget/groupbox.py
```diff
@@ -146,7 +146,11 @@ def draw(self):
 
 
 class GroupBox(_GroupBase):
-    """A widget that graphically displays the current group"""
+    """
+    A widget that graphically displays the current group.
+    All groups are displayed by their label.
+    If a groups label is the empty string that group will not be displayed.
```
kynikos: Typo: groups -> group's

dhannes:
Changed the WindowList extension to use the group's label if present.
(And thanks for proofreading.)

libqtile/extension/window_list.py
```diff
@@ -48,7 +48,7 @@ def list_windows(self):
         for win in windows:
             if win.group:
                 item = self.item_format.format(
-                    group=win.group.name, id=id, window=win.name)
+                    group=win.group.label or win.group.name, id=id, window=win.name)
```
kynikos: Thanks @dhannes, I'm only wondering if this would be safe as simply `group=win.group.label`, since `win.group.label` is already initialized as `self.label = name if label is None else label` in `Group.__init__()`?

dhannes: Yes, but the label may also be the empty string(!)
in that case, the group is not dislayed in GroupBox widget(, and that is my initial requirement.)
However, for WindowList the windows are listed and belong to a group.
If the group has an empty label, the name of the group is used in formatting.

In Group.__init__ we have three possible values for label:
- label = None (default) --> not set, use the group name as label
- label = <empty string> --> the label is set explicitly to an empty string (hidden in GroupBox)
- label = <text> --> a meaningfull text to be displayed

kynikos: Indeed, thanks for clarifying ^^

flacjacket:
This looks good, thanks for this!

PR #1071, opened 2017-05-18
fix cmd_maximize for monadtall

zordsdavini:
lost default value for medium


flacjacket:
Oh, sorry for leaving this so long, not sure how I missed it. Thanks!

PR #1072, opened 2017-06-04
I do not want mapped windows to steal focus in any situation ever

jettero:
This seems like a reasonable approach to me. Code reviews welcome, etc, etc.

### DIFF
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -630,6 +630,10 @@ class Qtile(command.CommandObject):
                 self.windowMap[w.wid] = c
                 # Window may have been bound to a group in the hook.
                 if not c.group:
+                    # NOTE: checking self.config.focus_on_mapped_windows will
+                    # fail tests since most of the test configs don't have this
+                    # set and also aren't real configs. should the test configs
+                    # be read with readconfig like any other config?
                     if getattr(self.config, 'focus_on_mapped_windows', True):
                         steal_focus = c.can_steal_focus()
                     else:
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -630,12 +630,7 @@ class Qtile(command.CommandObject):
                 self.windowMap[w.wid] = c
                 # Window may have been bound to a group in the hook.
                 if not c.group:
-                    # NOTE: checking self.config.focus_on_mapped_windows will
-                    # fail tests since most of the test configs don't have this
-                    # set and also aren't real configs. Should the test configs
-                    # be read with readconfig like any other config?
-                    steal_focus = getattr(self.config,
-                        'focus_on_mapped_windows', True) and c.can_steal_focus()
+                    steal_focus = self.config.focus_on_mapped_windows and c.can_steal_focus()
                     self.currentScreen.group.add(c, focus=steal_focus)
                 self.update_client_list()
                 hook.fire("client_managed", c)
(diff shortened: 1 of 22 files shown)


kynikos:
Hi, I haven't looked into why this change breaks all those tests, but I see that you should also document the new variable in [docs/manual/config/index.rst](https://github.com/qtile/qtile/blob/develop/docs/manual/config/index.rst).

jettero:
Hrm. I'll examine the travis output and document. Thanks.

jettero:
It looks like at least some of the tests fail because of travis config errors (pep8 invocation error in one of them... I can't find it now). Many seem to fail because I missed something somewhere. I'll figure it out eventually.

jettero:
I pushed the same commits, but skipped commit 0805b7e to see if the tests go any differently

jettero:
Yeah, won't make any difference cuz it merges to develop before build.

jettero:
I'm still baffled by the test results. I didn't do anything I'd think would have broken all those tests. I enabled travis-ci for my fork and pushed a bunch of variations (with and without my new option, and having skipped 0805b7e, which doesn't seem to be the problem).

https://travis-ci.org/jettero/qtile/branches

jettero:
I think I see the problem now. The tests don't use a full config built from the default config.

jettero:
some of the xephyr# kill window tests fail sometimes. I can't see any obvious reason why that happens. I guess I'm done messing with this for now. Comments welcome.

libqtile/manager.py
```diff
@@ -630,7 +630,13 @@ def manage(self, w):
                 self.windowMap[w.wid] = c
                 # Window may have been bound to a group in the hook.
                 if not c.group:
-                    self.currentScreen.group.add(c, focus=c.can_steal_focus())
+                    # NOTE: checking self.config.focus_on_mapped_windows will
+                    # fail tests since most of the test configs don't have this
+                    # set and also aren't real configs. Should the test configs
+                    # be read with readconfig like any other config?
+                    steal_focus = getattr(self.config,
```
flacjacket: You shouldn't need a `getattr` here, you should just be able to get the `self.config.focus_on_mapped_windows` property

jettero: I explained in the note... many tests fail if I do that because the tests don't load the real config, they fake it. I could go through the test configs and add the option...

jettero: I'm attempting to figure out which tests those'll be presently.

jettero: OK, I reversed the getattr and instead added the `focus_on_mapped_windows = True` to 30 config files in `test/`. I feel like I have to be missing something though. It seems to me the config files aught to automatically notice the new option from the confreader or default_config.py

kynikos:
To help clarifying the problem, `default_config.py` is normally read around [libqtile/confreader.py#L59](https://github.com/qtile/qtile/blob/develop/libqtile/confreader.py#L59), coming from [libqtile/scripts/qtile.py#L108](https://github.com/qtile/qtile/blob/develop/libqtile/scripts/qtile.py#L108), however the tests use their own config classes, e.g. [test/conftest.py#L426](https://github.com/qtile/qtile/blob/develop/test/conftest.py#L426), so `default_config.py` is never read when testing.

I think there aren't many options, the solution is to either make all test configs read from a default configuration object (which could be `default_config.py` or a separate one), or indeed do what you've done, i.e. add the new attribute to all the test classes that now fail without it.

PR #1074, opened 2017-06-14
Made sticky windows intrusive

JacobEdelman:
Made it so qtile complies with _NET_WM_STICKY from [the EWMH standard](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm140130317598336).


JacobEdelman:
(scratch that doesn't work right now, will fix)

PR #1078, opened 2017-07-27
Add step argument to Volume widget

Terseus:
This new argument allow to choose how much the volume percentage will change when using the `BUTTON_UP` and `BUTTON_DOWN` buttons.

By default it's 2, which were the hardcoded step.


flacjacket:
Nice, thanks for this.

PR #1080, opened 2017-07-31
update btc tracker to use coinbase

tych0:
Seeems btc-e has died, so that was a bad choice. Coinbase is much more
reputable, perhaps they'll stick around for a few more years.

Note that this drops quite a bit of the flexibility of the previous widget
w.r.t. other currencies, but adds support for many more fiat currencies.
This is simply because of the exchange capabilities. However, I've taken
care to make sure that old configs will at least not crash the new widget,
with a best-effort to honor the ones that can be honored.

### DIFF
--- a/libqtile/widget/bitcoin_ticker.py
+++ b/libqtile/widget/bitcoin_ticker.py
@@ -30,52 +30,39 @@ import locale
 
 class BitcoinTicker(GenPollUrl):
     """
-    A bitcoin ticker widget, data provided by the btc-e.com API. Defaults to
+    A bitcoin ticker widget, data provided by the coinbase.com API. Defaults to
     displaying currency in whatever the current locale is. Examples:
 
     ::
-
         # display the average price of bitcoin in local currency
-        widget.BitcoinTicker(format="BTC: {avg}")
+        widget.BitcoinTicker()
 
-        # display the average price of litecoin in local currency
-        widget.BitcoinTicker(format="LTC: {avg}", source_currency='ltc')
-
-        # display the average price of litecoin in bitcoin
-        widget.BitcoinTicker(format="BTC: ฿{avg}", source_currency='ltc', currency='btc', round=False)
+        # display it in Euros:
+        widget.BitcoinTicker(currency="EUR")
     """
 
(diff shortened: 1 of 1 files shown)


libqtile/widget/bitcoin_ticker.py
```diff
@@ -30,52 +30,39 @@
...
-        widget.BitcoinTicker(format="BTC: {avg}")
-
-        # display the average price of litecoin in local currency
-        widget.BitcoinTicker(format="LTC: {avg}", source_currency='ltc')
+        widget.BitcoinTicker()
 
-        # display the average price of litecoin in bitcoin
-        widget.BitcoinTicker(format="BTC: ฿{avg}", source_currency='ltc', currency='btc', round=False)
+        # display it in Euros:
+        widget.BitcoinTicker(currency=EUR)
```
flacjacket: Should this be a string `"EUR"`?

libqtile/widget/bitcoin_ticker.py
```diff
@@ -30,52 +30,39 @@
...
-        if 'error' in body and body['error'] == "invalid pair":
-            locale.setlocale(locale.LC_MONETARY, "en_US.UTF-8")
-            self.currency = locale.localeconv()['int_curr_symbol'].strip()
-            body = self.fetch(self.url)
-        for k, v in body['ticker'].items():
-            formatted[k] = v
-            if self.round:
-                formatted[k] = locale.currency(v)
-        return self.format.format(**formatted)
+        return "BTC: {symbol}{amount}".format(symbol=self.symbol, amount=body['data']['amount'])
```
flacjacket: This should be left as a configurable

flacjacket: Oh, wait, nevermind, I see, it's just for bitcoin to a fiat currency.

libqtile/widget/bitcoin_ticker.py
tych0: Yep, sorry, fixed :)

flacjacket:
Great, thanks for the update!

PR #1081, opened 2017-08-20
Handle 'sensors' command errors

### LINKED ISSUE #1077: Tests should not depend on sensors being present on build host
When building for some architectures, lm-sensors is returning that no sensors are present, which is causing "test_thermalsensor_regex_compatibility" to fail.

Probably catch the CalledProcessError that is raised, sensors returns an exit code of 1 when no sensors are present on the system, this test should then be marked as skipped.

Terseus:
Fixes #1077 

Handles the error returned by the 'sensors' command as if 'sensors' returned no sensor data and issues a warning.

'sensors' returns the code 1 in more situations than [when no sensors are found](https://github.com/groeck/lm-sensors/blob/1b36ea429930b05e1b3cb77cdbc24cb1ffbcac61/prog/sensors/main.c#L352), like if [sensors failed an initialization](https://github.com/groeck/lm-sensors/blob/1b36ea429930b05e1b3cb77cdbc24cb1ffbcac61/prog/sensors/main.c#L112) for whatever reason, so the downside of this change is that it would hide critical 'sensors' errors from the user; the previous behavior was crashing.

I think a warning is better than a crash in this situation.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -176,6 +176,10 @@ class UnixCommandNotFound(Warning):
     pass
 
 
+class UnixCommandRuntimeError(Warning):
+    pass
+
+
 def catch_exception_and_warn(warning=Warning, return_on_exception=None,
                              excepts=Exception):
     """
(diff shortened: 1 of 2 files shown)


flacjacket:
Hi Terseus, thanks for working on this!

PR #1082, opened 2017-08-26
Replace xterm with xcalc in all tests

### LINKED ISSUE #1076: Ability to skip xterm tests
xterm requires that a pty can be created, which isn't possible in the restrictive Debian build environment.

What would be the best way to skip the xterm tests? Can something else (xclock?) be run in place of xterm?

Terseus:
Includes the following changes:
* Replace the function `testXterm` in `test/conftest.py` with `testXcalc`
* Replace all usages of `testXterm` with `testXcalc`
* Replace specific xterm tests with xcalc tests

Closes #1076


flacjacket:
Great! Thanks for taking care of this

PR #1085, opened 2017-09-18
TaskList: allow easy adjusting of list of window by inheritance

tuomas2:
Code improvement.

Rationale: I wanted to create custom widget for Columns layout that showed only the windows in current column like this:

```
class CurrentColumnTaskList(TaskList):
    """
    Works only in Columns layout. Shows
    all tasks in current column
    """
    @property
    def windows(self):
        return self.bar.screen.group.layout.cc.clients

```

It appeared that base code required some small improvements, thus this PR.


flacjacket:
Sorry for not getting to this in almost a month! This change looks alright to me.

PR #1086, opened 2017-10-03
fix cursor warp when using default layout implementation

tych0:
Layouts like xmonad don't manage their own window lists, so the
cursor_warp = True parameter doesn't work for them.

Instead, let's by default on next/prev for these layouts enable warping (of
course, it is disabled if people don't set cursor_warp=True). A better fix
might just be to always default this to true... I'm not sure if there's a
real case where it should be false.

This commit also removes an outdated comment -- cursor warp is hooked up
now :)


flacjacket:
Nice, thanks for the fix!

PR #1090, opened 2017-10-30
Added functions that are creation and removing a section using prompt widget.

himaaaatti:


himaaaatti:
This PR work like this gif.
output

libqtile/layout/tree.py
```diff
@@ -637,6 +642,30 @@ def cmd_decrease_ratio(self):
         self.panel_width -= 10
         self.group.layoutAll()
 
+    def cmd_add_section_using_prompt(self):
+        prompt = self.group.qtile.widgetMap['prompt']
```
flacjacket: I think it would be a good idea to parameterize the widget name here.

libqtile/widget/prompt.py
```diff
@@ -434,11 +434,15 @@ def startInput(self, prompt, callback,
         self.show_cursor = self.cursor
         self.cursor_position = 0
         self.callback = callback
-        self.completer = self.completers[complete](self.qtile)
+        if not complete:
```
flacjacket: Why is this needed? Can you not use the `NullCompleter`?

himaaaatti:
merged a head of develop branch.

himaaaatti:
I'll consider about more generalized method.

PR #1094, opened 2017-11-04
Fix linter errors

tych0:
flake8 got some updates, so let's make sure we are compliant.


flacjacket:
Nice! Thanks for the fixes!

PR #1099, opened 2017-11-27
rework outdated bitcoin ticker widget

Granitosaurus:
btc-e.com has been dead for ages. This rework changes the widget to use coinmarketcap api which supports majority of cryptocurrencies.

It's my first PR, so let me know if I missed something! :)

### DIFF
--- a/libqtile/widget/bitcoin_ticker.py
+++ b/libqtile/widget/bitcoin_ticker.py
@@ -1,0 +1,13 @@
+from libqtile.widget.crypto_ticker import CryptoTicker
+import logging
+
+logger = logging.getLogger('BitcoinTicker')
+
+
+class BitcoinTicker(CryptoTicker):
+    """deprecated, in favor of CryptoTicker"""
+
+    def __init__(self, **config):
+        super().__init__(**config)
+        logger.warning('BitcoinTicker widget is deprecated in favor of '
+                       'CryptoTicker widget and will be removed in the future')
--- a/libqtile/widget/crypto_ticker.py
+++ b/libqtile/widget/crypto_ticker.py
@@ -30,19 +30,19 @@ import locale
 
 class CryptoTicker(GenPollUrl):
     """
-    A bitcoin ticker widget, data provided by the btc-e.com API. Defaults to
-    displaying currency in whatever the current locale is. Examples:
+    A cryptocurrency ticker widget, data provided by the coinmarketcap API. Defaults to
+    displaying bitcoin currency in whatever the current locale is (or usd). Examples:
 
     ::
 
-        # display the average price of bitcoin in local currency
-        widget.BitcoinTicker(format="BTC: {avg}")
+        # display the average price of bitcoin in euros in format like "BTC:9001"
+        widget.CryptoTicker(from_currency='bitcoin', to_currency='eur', format="{symbol}:{price}")
 
         # display the average price of litecoin in local currency
-        widget.BitcoinTicker(format="LTC: {avg}", source_currency='ltc')
+        widget.CryptoTicker(from_currency='litecoin')
 
         # display the average price of litecoin in bitcoin
-        widget.BitcoinTicker(format="BTC: ฿{avg}", source_currency='ltc', currency='btc', round=False)
+        widget.CryptoTicker(format="{symbol}:{price_btc}", from_currency='litecoin')
     """
 
     QUERY_URL = "https://api.coinmarketcap.com/v1/ticker/{from_currency}/?convert={to_currency}"
--- a/libqtile/widget/crypto_ticker.py
+++ b/libqtile/widget/crypto_ticker.py
@@ -36,7 +36,7 @@ class CryptoTicker(GenPollUrl):
     ::
 
         # display the average price of bitcoin in euros in format like "BTC:9001"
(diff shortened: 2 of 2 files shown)


flacjacket:
I know @tych0 moved the widget from btc-e to coinbase a while ago, the docs must have just not been updated. What is the benefit of coinmarketcap over coinbase? Is there any downsides here? I know one thing with the move from btc-e is that coinbase only had bitcoin prices, it looks like coinmarketcap has many more crypto currencies, that may be worth changing for.

Granitosaurus:
Ah, I've only checked master branch, didn't expect a rework to be out on `develop` branch, sorry. 
Nevertheless I think supporting cryptocurrencies other than bitcoin and exchanges other than coinbase (which has relatively small market share).  

`coinmarketcap` aggregates most of the exchange prices and generates a weighted average, i.e. price of bitcoin on bitfinex will have biggest influence since it takes 20% of the market. 
Another benefit of coinmarket cap is that cover pretty much every cryptocurrency and even tokens and provides more info than just price, e.g. `percentage_change_24h`

tych0:
A few things:

* there are some flake errors on this PR
* it looks like the API changes quite a bit here, which is annoying. If nothing else, it should at least be documented in CHANGELOG what the steps are to forward port your config. But it would be nice if we could re-export something for legacy BitcoinTicker users so their configs don't break.
* I chose coinbase because it's reputable and has been around a while, so that we wouldn't have to keep updating the widget. coinmarketcap is much younger, and it's not clear to me that they have a business with customers that will keep them afloat. So, we could move to this, but we may have to update the widget again in the future.

Anyway, no big issues here, just all things to consider.

Granitosaurus:
Hey tych0, thanks for the feedback.

1. There are some flake errors on this PR
	Looking into it

2. It looks like the API changes quite a bit here, which is annoying. If nothing else, it should at least be documented in CHANGELOG what the steps are to forward port your config. But it would be nice if we could re-export something for legacy BitcoinTicker users so their configs don't break. 
	
    AFAIK currently BitcoinTicker in the current release and master branch is broken and has slightly different api to the fixed version on develop branch. I guess I could change `to_currency` to the original `currency` setting and it would work and duplicated BitcoinTicker as CryptoTicker but that seems a bit too conservative for a simple widget. 

3. I chose coinbase because it's reputable and has been around a while, so that we wouldn't have to keep updating the widget. coinmarketcap is much younger, and it's not clear to me that they have a business with customers that will keep them afloat. So, we could move to this, but we may have to update the widget again in the future.

	I like coinbase as well but I feel it's not the greatest source for accurate tick pricing as it has small market share and offers only few cryptocurrencies.

Let me know what you think would be the correct way to proceed with #2 (and if you have any other comments), I'm working on understanding and fixing the current flake/travis errors :)

tych0:
I think if you `s/to_currency/currency/` and re-export the widget under the BitcoinTicker alias, that's enough to preserve the current widget API.

I didn't choose coinbase for accuracy, I mostly chose it so that I wouldn't have to update the widget again when yet another exchange went down (I've already ported it twice, mtgox -> btce -> coinbase). If you don't mind porting it when coinmarketcap dies, I'm happy with this (assuming the above fix) :)

Granitosaurus:
@tych0 I've updated to include a bit of backwards compatibility. For some reason some tests fail on something that looks completely unrelated though.

libqtile/widget/bitcoin_ticker.py
```diff
@@ -1,68 +1,13 @@
...
-        self.symbol = locale.localeconv()['currency_symbol']
-
-    @property
-    def url(self):
-        return self.QUERY_URL % self.currency.lower()
-
-    def parse(self, body):
-        return "BTC: {symbol}{amount}".format(symbol=self.symbol, amount=body['data']['amount'])
+        super().__init__(**config)
+        logger.warning('BitcoinTicker widget is deprecated in favor of '
```
flacjacket: I think you should use `widget.base.deprecated` rather than logging.

libqtile/widget/bitcoin_ticker.py
```diff
@@ -1,68 +1,13 @@
...
-
-    ::
-        # display the average price of bitcoin in local currency
-        widget.BitcoinTicker()
-
-        # display it in Euros:
-        widget.BitcoinTicker(currency="EUR")
-    """
-
-    QUERY_URL = "https://api.coinbase.com/v2/prices/spot?currency=%s"
```
flacjacket: If you leave the QUERY_URL on this class, the users of the previous class will continue to get coinbase prices as they'd expect.

I feel like before this class is removed, we should make the exchange a configurable parameter.

tych0: That's a great idea :) +1

Granitosaurus:
I'm thinking of reworking this to use https://www.cryptocompare.com/api/ as it would allows more settings such as selecting specific exchange.

Granitosaurus:
@ramnes thanks for reminding me. Yeah I do plan to sit down and finish this up once I get a second. Let me set a reminder.

ramnes:
@Granitosaurus This PR has been closed automatically because we removed `develop`, so feel free to open a new PR that targets `master` if you still want to work on this. :+1:

Issue #1100, opened 2017-11-28
Test(Travis-CI) is failed because a py library is not support python 3.3 from v1.5.0.


sux2mfgj:
the py library is required by a pytest-cov. How to resolve this problem?   

- Python 3.3.x has already reached end of life.
https://www.python.org/downloads/release/python-337/

- This is a announcement of the py library.
http://py.readthedocs.io/en/latest/changelog.html#id2

- a failed test 
https://travis-ci.org/himaaaatti/qtile/builds/308500409  
This is a my repository but this branch is same as a head of qtile/qtile develop branch.

numirias:
My PR fails because of that too.

Is Python 3.3 support something that needs to be maintained?

sux2mfgj:
My PR for this problem (#1103 ) is merged !! thanks!

PR #1101, opened 2017-12-03
Enable a custom TaskList icon size

numirias:
This enables setting a custom icon size for icons in the TaskList widget.

I had the experience that the calculated icon sizes didn't match up well with my small font, so I figured a setting would be useful to control the icon size myself. As a side effect, it's now also possible to hide icons completely, if required.

(Travis seems to fail because of #1100, unrelated to the actual changes in this PR.)


flacjacket:
This looks good, thanks for the contribution!

Issue #1096, opened 2017-11-16
Columns and Wmii layouts are almost equivalent


heitzmann:
In their current versions, these 2 layouts are almost identical.
From what I can tell, Columns implements all of Wmii features with a few extra options (`autosplit`, `num_columns` and `fair`).

Is there anything in Wmii that is missing in Columns?
If not, it makes sense to leave only one of them, doesn't it?

flacjacket:
If all of the Columns functionality can be written in terms of Wmii, then yes, we can get rid of columns. I can take a look as well, but feel free to open a pull request deleting one if you think this is the case.

ramnes:
MR in progress: #1107 1107

PR #1102, opened 2017-12-04
Merge layouts wmii and columns

heitzmann:
Columns' default parameters are changed to reflect wmii's.
Rename columns to replace wmii.

Motivation: columns presents all features of wmii plus a few others, making wmii redundant. #1096


ramnes:
Damn. Chances are high that this will break my [config](https://github.com/ramnes/qtile-config/blob/master/config.py), that I changed some days ago to use Columns, and for which I heavily rely on its implementation. :sob: 

From what I seem to see, you're reimplementing Columns inside WMII. What's the point of doing that versus simply removing the WMII layout?

heitzmann:
@ramnes What this PR does is to remove Wmii and rename Columns to Wmii. It also changes the default parameters of Columns to match Wmii.
I chose to leave the Wmii name instead of Columns because it's been around for longer, so I thought it would break fewer configs out there...

If you're using Columns, just change to Wmii (in this PR) and modify any default parameters that might have been changed.

flacjacket:
I haven't yet sat down and looked at the changes here, but my suggestion would be to leave both, and mark one of them as deprecated. For whichever one you choose to remove, you can just subclass the other and set the defaults to match the original behavior.

heitzmann:
Sounds good @flacjacket 
I'll close this PR and open a new one with the deprecation warning.

PR #1103, opened 2017-12-04
abolition python3.3

sux2mfgj:
This PR is one of the resolution of issue #1100.


flacjacket:
Great, thanks for cleaning this all up, 3.3 was EOL back in September, so this change is fine.

PR #1104, opened 2017-12-05
Implemented a general way to use of prompt widget.

sux2mfgj:
This PR has one bug fix and one additional function.

The bug is when I use NullCompleter at the prompt widget, self.history is not created. 

I implemented a general way to use of prompt widget. A behavior is same as #1090.  I also added a document in a source code, however, I think my English is not good. If you think you need fix document,  fix it, please.

thanks.


libqtile/widget/prompt.py
```diff
@@ -661,10 +668,56 @@ def cmd_info(self):
...
+            Key([alt, 'shift'], 'a',
+                lazy.widget['prompt'].exec_general(
+                    'section(add)',
+                    'layout',
+                    'add_section'))
+        """
+        try:
+            obj = self.qtile.select([(object_name, selector)])
+        except _SelectError:
+            logger.info("cannot select a object")
```
flacjacket: I would raise this to `error` log level for all of these errors.

libqtile/widget/prompt.py
```diff
@@ -377,15 +378,21 @@ def __init__(self, name="prompt", **config):
...
+                    # self.history of size does not match.
+                    if len(self.history) != len(self.completers):
+                        self.history = {x: deque(maxlen=self.max_history)
+                                for x in self.completers}
+
                     if self.max_history != \
                        self.history[list(self.history)[0]].maxlen:
                         self.history = {x: deque(self.history[x],
                                                  self.max_history)
-                                        for x in self.completers if x}
```
flacjacket: Why do you remove all of the `if x` from the dict comprehensions?

libqtile/widget/prompt.py
sux2mfgj: I fixed it by cc676ed2141121dbfaeaeb78907084582fc5686e . thank you for your comment.

libqtile/widget/prompt.py
sux2mfgj: This ```if x``` occurred error at https://github.com/qtile/qtile/blob/develop/libqtile/widget/prompt.py#L442.
Is it better to resolve this problem by another changings?

flacjacket:
Thanks for this, I squashed these diffs and push these changes, thanks!

sux2mfgj:
thank you!

PR #1105, opened 2017-12-10
Bsp layout

heitzmann:
This layout is inspired by bspwm, but it does not try to copy its features.

The first client occupies the entire srceen space.  When a new client is created, the selected space is partitioned in 2 and the new client occupies one of those subspaces, leaving the old client with the other.

The partition can be either horizontal or vertical according to the dimensions of the current space: if its width/height ratio is above a pre-configured value, the subspaces are created side-by-side, otherwise, they are created on top of each other.  The partition direction can be freely toggled.  All subspaces can be resized and clients can be shuffled around.


ramnes:
Those features seem pretty similar to Columns ones. Did you try that layout? What was missing? If the difference is not too big, you could maybe try inheriting from it.

I'm saying all this because there's already an issue (#1096) and MR (#1102) about Columns and WMII layouts being pretty-much duplicates; it seems to me that maybe we could avoid doing this a third time here. :smiley:

heitzmann:
@ramnes I'm actually the author of Columns... this is quite different.

In Columns (Wmii, Stack, etc...) it is not possible, for example, to have a window occupying the full width of the top half of the screen and a few columns splitting the bottom half.

This layout is inspired by [bspwm](https://github.com/baskerville/bspwm).

ramnes:
Ah yeah, I failed to realize this because I used to use it just like columns, but you're right, this is much more powerful. I think that kind of "manual" layout basically will deprecate most other layouts, so thank you for doing it. :+1:

flacjacket:
Cool, thanks for this addition!

PR #1107, opened 2017-12-11
Columns merge

heitzmann:
In reply to #1102, this PR keeps both layouts but:

- Implements layout `Wmii` as `Columns` with different default settings
- Deprecates `Wmii` for future elimination


test/layouts/test_wmii.py
```diff
@@ -61,9 +61,9 @@ def test_wmii_window_focus_cycle(qtile):
     qtile.testWindow("three")
 
     # test preconditions
-    assert qtile.c.layout.info()['clients'] == ['one', 'two', 'three']
+    assert qtile.c.layout.info()['clients'] == ['three', 'two', 'one']
```
flacjacket: So I guess the difference that we capture in our tests is that `Columns` inserts new clients above, while `Wmii` inserts new clients below. It looks like the the `_ClientList.add` has an `offset_to_current` arg that could be used to account for this. Do you think it would be possible to add that as a parameter that is used, so `Wmii` behavior doesn't change?

heitzmann:
@flacjacket Sure, it seems like a better default for `Columns` as well.
I exposed the `offset_to_current` (as `insert_position`) to the configuration of `Columns` so that one could, for example, always insert at the top (`insert_position = -10000`).

heitzmann:
@ramnes Nothing to add that I aware. I don't believe there have been any changes to either Wmii or Columns since this PR.

libqtile/layout/columns.py
```diff
@@ -103,23 +111,30 @@ class Columns(Layout):
...
         ("border_focus", "#881111", "Border colour for the focused window."),
         ("border_normal", "#220000", "Border colour for un-focused windows."),
+        ("border_focus_stack", "#881111",
+         "Border colour for the focused window in stacked columns."),
+        ("border_normal_stack", "#220000",
+         "Border colour for un-focused windows in stacked columns."),
         ("border_width", 2, "Border width."),
         ("margin", 0, "Margin of the layout."),
-        ("autosplit", True, "Autosplit newly created columns."),
+        ("split", True, "New columns presentation mode."),
```
ramnes: Is this potential config breakage worth it? What's the rationale?

libqtile/layout/columns.py
```diff
@@ -67,20 +71,24 @@ def remove(self, client):
...
-    space.  Columns and windows can be resized and windows can be shuffled
-    around.  This layout can also emulate "Wmii", "Verical", and "Max",
-    depending on the default parameters.
+    removed.  Each column can present is windows in 2 modes: split or
+    stacked.  In split mode, all windows are presented simultaneously,
+    spliting the column space.  In stacked mode, only a single window is
+    presented from the stack of windows.  Columns and windows can be
+    resized and windows can be shuffled around.
+
+    This layout can also emulate "Wmii", "Verical", and "Max", depending
```
ramnes: Vertical

libqtile/layout/columns.py
```diff
@@ -103,23 +111,30 @@ class Columns(Layout):
...
+        ("border_normal_stack", "#220000",
+         "Border colour for un-focused windows in stacked columns."),
         ("border_width", 2, "Border width."),
         ("margin", 0, "Margin of the layout."),
-        ("autosplit", True, "Autosplit newly created columns."),
+        ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
         ("fair", False, "Add new windows to the column with least windows."),
+        ("insert_position", 1,
```
ramnes: What would you think using a boolean value, with `"insert_after", True` for example, since there are only two possible choices anyway?

ramnes: Also, if I understand correctly the tests diff, shouldn't we insert before by default, so that the current layout behavior is kept?

libqtile/layout/columns.py
heitzmann: This is just to get in line with the Wmii layout naming scheme, but not strictly necessary.
(Also `split` is shorter and describes better the setting than `autosplit`.)

libqtile/layout/columns.py
heitzmann:
```suggestion
    This layout can also emulate "Wmii", "Vertical", and "Max", depending
```

libqtile/layout/columns.py
heitzmann: `insert_position` is a little more general. The int value indicates the position relative to the current client. This argument comes from `base._ClientList.add(client, offset_to_current)`.

For example, it could be set to some large number to ensure that new clients are always inserted at the end of the list. It could also be set to a very small number (large negative) to ensure new clients are always at the beginning.

As for the default value, either way is fine. Inserting after is in line with `Wmii`'s default, which, in hindsight, made more sense to me.

ramnes: Thanks for the explanation. What would you think of using `1` on `Wmii`'s defaults, and `0` here?

ramnes:
Just a quick note to say that I tried your branch to see if there is any regression with `Columns`, and it's all good.

That being said, that new `insert_position` option defaulting to `1` put a bug in evidence: `ClientList.add` wasn't keeping track of the current client so far. The bug wasn't visible without offset because in that case the client is added in place in the list. So yeah, I got that fixed in https://github.com/qtile/qtile/commit/a925295272ff4264a259fda01969ca6583a56a4e.

ramnes:
Also, I've taken the rebase to the latest version of `develop` myself. Please update your branch with something like:
```
git reset --hard develop
git pull https://github.com/ramnes/qtile.git heitzmann-columns-merge
git push --force
```

libqtile/layout/columns.py
heitzmann: Yes, I believe that works. I'll go ahead and make the change.

test/layouts/test_columns.py
```diff
@@ -62,11 +62,11 @@ def test_columns_window_focus_cycle(qtile):
...
-    assert qtile.c.layout.info()['columns'][0]['clients'] == ['one']  # 'three', 'two', 'one']
+    # test preconditions, columns adds clients at pos after current, in two stacks
+    assert qtile.c.layout.info()['columns'][0]['clients'] == ['one']
     assert qtile.c.layout.info()['columns'][1]['clients'] == ['three', 'two']
     # last added window has focus
     assertFocused(qtile, "three")
 
     # assert window focus cycle, according to order in layout
-    assertFocusPath(qtile, 'two', 'float1', 'float2', 'one', 'three')
+    assertFocusPath(qtile, 'float1', 'float2', 'one', 'three', 'two')
```
ramnes: Looks like this is failing: https://travis-ci.org/qtile/qtile/jobs/454683349#L901

heitzmann:
Using `insert_position = 0` as default for Wmii will change it's default behavior.
I believe it's better to fix the tests to reflect that default, no?

ramnes:
The tests are supposed to reflect the current truth, so if we don't want to break configs, we must not change them but rather comply to them. If they were passing so far, it should mean that Wmii default actually isn't `insert_position = 1`, no?

Or is there something that I'm missing here?

heitzmann:
You're right, @ramnes, I've reverted the tests. The issue was something else: `insert_position` was not being taken into account at initialization.
I believe it is fixed now, but I'm not sure the way I'm passing the Wmii defaults is the best way (by updating `config`).

libqtile/layout/wmii.py
```diff
@@ -25,443 +16,34 @@
...
-                    self.focus(client)
-                    self.group.focus(client)
-                return
+        warnings.warn(
+            "Wmii layout is deprecated in favor of Columns.",
+            category=DeprecationWarning,
+            stacklevel=2)
+        for key, value, _ in Wmii.defaults:
+            if key not in config:
+                config[key] = value
```
ramnes: Can't we simply do something like this?

```python
Columns.__init__(self, **config)
self.add_defaults(Wmii.defaults)
```

ramnes:
We're almost there!

libqtile/layout/wmii.py
heitzmann: That's how it was before. The problem is that on initialization, Columns creates a Column with the current values of `split` and `insert_position`. That first instance of Column will incorrectly be created with Column.defaults instead of Wmii.defaults.

ramnes: Ah, yeah, alright!

Let's just replace the `if` block with `dict.setdefault` then? 😉

ramnes:
We *are* there!

Thank you very much @heitzmann for your great contribution as well as for your reactivity. :+1:

heitzmann:
Thank you all for creating and maintaining qtile! For me it's the perfect WM!

Issue #1095, opened 2017-11-16
Group spawned through Matching does not use label


Sweenu:
When creating a group like that: 
`Group('music', layouts=[layout.Max()], label='spotify')`
And having the group spawned when qtile starts, the label is used in the GroupBox widget.

However if I use the matching mechanism, the label isn't used, it's the name of the group that is used.
`Group('music', layouts=[layout.Max()], persist=False, init=False,
          matches=[Match(title=['Spotify'])], label='spotify')`

flacjacket:
This is closed by #1108.

PR #1108, opened 2017-12-11
Fix issue #1095

dhannes:
Use group label also for non persistent groups.
Seems I missed it in #1070.


flacjacket:
Great, thanks for the fix!

PR #1111, opened 2017-12-18
add a stock ticker widget

tych0:
This also happens to do cryptocurrencies, but supports actual stocks as
well.


flacjacket:
Nice, thanks!

PR #1118, opened 2018-01-13
dynamic get battery name

hellwen:


hellwen:
battery name in my thinkpad x260 is BAT1, so i change the default BAT0 to dynamic get from BAT_DIR

libqtile/widget/battery.py
```diff
@@ -61,9 +61,10 @@ class _Battery(base._TextBox):
     """Base battery class"""
 
     filenames = {}
+    # ACPI name of a battery, usually BAT0
+    battery_name = 'BAT0'
 
     defaults = [
-        ('battery_name', 'BAT0', 'ACPI name of a battery, usually BAT0'),
```
tych0: Can we just s/BAT0/get_battery_name()/ here? That way we will still get all the docs autogenerated for this from config.

hellwen: maybe, i will try it

hellwen: _get_battery_name() replaced BAT0

tych0:
LGTM, thanks!

PR #1119, opened 2018-01-15
remove stock ticker's debug spew

tych0:
whoops :)


flacjacket:
Ah yes, thanks for the fix

Issue #1110, opened 2017-12-16
Make using unavailable widgets not crash


flacjacket:
See #1106 for discussion

flacjacket:
I feel like the way to fix this is to have the `safe_import` replace the offending widget with a text widget class that displays an error.

Roger:
I'm working on this, how should we display the error? this is what I got for now

![image](https://user-images.githubusercontent.com/9746/35385857-8dac8700-01a8-11e8-87dd-c5d30394e55f.png)

ramnes:
From my experience other WM handle that by not loading the module and popping a notification. But that works too.

flacjacket:
Without another notification system, I think that is a great solution. Since this is covered by Roger's PR, I'll close this.

PR #1122, opened 2018-01-25
Don't crash qtile if you try to use a widget that has unmet deps

Roger:
Some users don't realize they are trying to use some widgets that needs
optional dependencies, regardless of the fact, the log warns about that
in the startup of qtile.

This replaces the "broken" widgets with a textbox saying the widget has
an import error.

This  fixes https://github.com/qtile/qtile/issues/1110


flacjacket:
Nice, thanks

PR #1124, opened 2018-01-29
Make group labels persist through restarts

### LINKED ISSUE #1116: Groups loses there label when restarting Qtile
Labels are not persistent through Qtile restarts. Every label is changed to the name of the corresponding group instead of keeping the custom value.

Sweenu:
Fix #1116


flacjacket:
Nice, thanks for the fix!

PR #1131, opened 2018-02-27
Typo

nazar-pc:


flacjacket:
Thanks for contributing this, but it looks like has already been fixed on `develop`, our development branch. When we cut a release, we will fast forward `master` to `develop`.

PR #1134, opened 2018-02-28
libxcb-cursor.so.0 is used instead of libxcb-cursor.so on Ubuntu

### LINKED ISSUE #1133: Cursor theme and size are changed by Qtile
I'm running xsettingsd that has Cursor theme and size configured.
Everything worked with custom Compiz session, but with Qtile most apps use some strange back with white border small cursor I don't even seem to have installed on my system (maybe something compiled into Xorg server).

Firefox, Terminator and all Qt apps show correct cursor and size as well as respond immediately to any changes. Other GTK apps like Nemo, GNOME System Monitor and Synaptic show this weird cursor.

The rest of settings like GTK theme are applied correctly.

I see this in Qtile output:
```
2018-02-28 08:12:07,998 libqtile xcursors.py:_setup_xcursor_binding():L117  xcb-cursor not found, fallback to font pointer
```

But I don't understand why it even needs to do something with cursor.

I'm on Ubuntu 18.04 and Qtile 0.10.7 installed using pip.

nazar-pc:
Fixes #1133

### DIFF
--- a/libqtile/xcursors.py
+++ b/libqtile/xcursors.py
@@ -114,8 +114,12 @@ class Cursors(dict):
         try:
             xcursor = ffi.dlopen('libxcb-cursor.so')
         except OSError:
-            logger.warning("xcb-cursor not found, fallback to font pointer")
-            return False
+            try:
+                # libxcb-cursor.so.0 is used on Ubuntu
+                xcursor = ffi.dlopen('libxcb-cursor.so.0')
+            except OSError:
+                logger.warning("xcb-cursor not found, fallback to font pointer")
+                return False
 
         conn = self.conn.conn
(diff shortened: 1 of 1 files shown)


flacjacket:
This is bizzare, it looks like `libxcb-cursor.so` is in `libxcb-cursor-dev`, while `libxcb-cursor.so.0` and `libxcb-cursor.so.0.0.0` are in `libxcb-cursor0`. I would almost be in favor of just switching the first attempt to use `libxcb-cursor.so.0` to avoid the problem from the outset.

nazar-pc:
Makes sense if it is the same in other distributions, feel free to push more commits to this PR, you should be able to do so.

PR #1135, opened 2018-02-28
Fix dbus-send call

nazar-pc:
`=string` causes `dbus-send` to complain and as the result the whole command execution fails.
At least this is what happens on my Ubuntu 18.04.


flacjacket:
Hmm, it looks like we're way behind the curve on fixing this. Shouldn't it be `--print-reply=literal` tho? Does it still work with that parameter?

PR #1143, opened 2018-03-01
Tile layout: add `add_after_last` option

nazar-pc:
Tile is an interesting layout, but there was only 2 choices for new windows positioning:
* at the beginning of the list
* before currently focused window

This PR adds a third one, controlled with `add_after_last` option, which defaults to `False` and doesn't affect behavior of existing users:
* at the end of the list

Also `add()` method signature aligned with signature of base class.


flacjacket:
Thanks for the updates!

PR #1144, opened 2018-03-02
Tile layout: fix master_match

nazar-pc:
I was getting `AttributeError: 'list' object has no attribute 'current_client'` because code was replacing `_ClientList` with plain `list`.

Not sure if this is a correct way to do it since I'm not a Python developer, but it works for me.


flacjacket:
So these are all great fixes, it would be great if you could also add unit tests to these methods so they aren't broken again in the future :)

nazar-pc:
I was thinking about addition of new tests already, will work on that later

PR #1147, opened 2018-03-02
Tile layout: unify commands so that it aligns well with other existing layouts

nazar-pc:
Consider keys config like following:
```
	'M-<Left>'     : lazy.layout.left(),
	'M-<Down>'     : lazy.layout.down(),
	'M-<Up>'       : lazy.layout.up(),
	'M-<Right>'    : lazy.layout.right(),
	'M-S-<Left>'   : [lazy.layout.move_left(), lazy.layout.shuffle_up()],
	'M-S-<Down>'   : [lazy.layout.move_down(), lazy.layout.shuffle_down()],
	'M-S-<Up>'     : [lazy.layout.move_up(), lazy.layout.shuffle_up()],
	'M-S-<Right>'  : [lazy.layout.move_right(), lazy.layout.shuffle_down()],
```

With current Tile layout implementation, shortcuts that on other layouts switch to other windows, in Tile cause windows rearrangement. This PR aligns Tile layout's commands so that they have the same effect as on other layouts.

### DIFF
--- a/libqtile/layout/tile.py
+++ b/libqtile/layout/tile.py
@@ -10,6 +10,7 @@
 # Copyright (c) 2014 dmpayton
 # Copyright (c) 2014 dequis
 # Copyright (c) 2017 Dirk Hartmann.
+# Copyright (c) 2018 Nazar Mokrynskyi
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


flacjacket:
Nice updates! I've updated the `CHANGELOG` as well, feel free to add future changes to the appropriate region of the changelog!

PR #1148, opened 2018-03-02
Make files transfer windows floating out of the box

nazar-pc:
This class is used in Nemo file manager and probably in others. Would be nice to support it in default config.


flacjacket:
Nice addition.

PR #1149, opened 2018-03-02
Two fixes for floating windows

### LINKED ISSUE #1140: Not all floating windows should be centered
OK, there is one regression in 0.11.0 comparing to 0.10.7: floating windows are always in the center.
This might be good for some purposes, but not always.

For instance, I have an app launcher Cerebro that appears slightly above the center of the screen and now it jumps from the center to its usual position each time I activate it.

The other example if IntelliJ IDEA (Java) whose dropdown menus, even nested, are now all exactly in the center of the screen, looks bizarre:

nazar-pc:
Two fixes for floating windows:
* Dropdown (like menu) Java windows are no longer positioned in the center of the screen (expected and correct behavior)
* Added `no_reposition_match` option that allows to specify which floating windows should not be forcibly repositioned to the center of the screen

Fixes #1140


flacjacket:
Can you add some documentation for what `no_reposition_match` does?

flacjacket:
Thanks for the quick fixes!

PR #1151, opened 2018-03-03
Fix exception during `previous()`/`next()` calls when there are no windows opened

### LINKED ISSUE #1146: Exception in focus_* methods
Not really sure when exactly this happened, but I've got following stacktrace (saw it few times in system logs):
```
2018-03-02 06:44:12,971 libqtile manager.py:handle_KeyPress():L933  KB command error next: Traceback (most recent call last):
  File "/web/github/qtile/libqtile/command.py", line 97, in call
    return (SUCCESS, cmd(*args, **kwargs))
  File "/web/github/qtile/libqtile/layout/base.py", line 614, in next
    client = self.focus_next(self.clients.current_client) or self.focus_first()
  File "/web/github/qtile/libqtile/layout/base.py", line 604, in focus_next
    return self.clients.focus_next(window)
  File "/web/github/qtile/libqtile/layout/base.py", line 412, in focus_next
    return self[self.index(win) + 1]
  File "/web/github/qtile/libqtile/layout/base.py", line 547, in index
    return self.clients.index(client)
ValueError: None is not in list
```

I think it might be a good idea to catch `ValueError` in addition to `IndexError` in `focus_*` methods.

nazar-pc:
In order to reproduce this issue use Matrix group with no windows open and call `next()` or `previous()` using hotkey, you'll get a stacktrace from #1146

Fixes #1146

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -410,7 +410,7 @@ class _ClientList(object):
         """
         try:
             return self[self.index(win) + 1]
-        except IndexError:
+        except (IndexError, ValueError):
             return None
 
     def focus_last(self):
@@ -425,7 +425,7 @@ class _ClientList(object):
         """
         try:
             idx = self.index(win)
-        except IndexError:
+        except (IndexError, ValueError):
             return None
         else:
             if idx > 0:


flacjacket:
Thanks for catching and fixing this! Tho I think the better fix for this is in the `previous` and `next` functions, before `focus_next` and `focus_previous` are called. I'll make that change and add the corresponding test for the `Matrix` class.

PR #1157, opened 2018-03-12
Add option to hide full battery status

czerw:
Enhancement: New option hide_full_status was added to battery widget
to hide battery status when is battery fully charged. It is disabled
by default for backward compatibility.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -169,6 +169,7 @@ class Battery(_Battery):
          'Display format'
          ),
         ('hide_threshold', None, 'Hide the text when there is enough energy'),
+        ('hide_full_status', False, 'Hide the text when is battery fully charged'),
         ('low_percentage',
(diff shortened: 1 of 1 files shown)


zordsdavini:
:+1:

libqtile/widget/battery.py
```diff
@@ -169,6 +169,7 @@ class Battery(_Battery):
          'Display format'
          ),
         ('hide_threshold', None, 'Hide the text when there is enough energy'),
```
flacjacket: It might be good to differentiate these two options in the documentation

czerw: @flacjacket I thought that adding to doc section (just bellow hide_threshold) is enough. Now, I'm not sure where to exactly put it, because I don't see any specific section in documentation (only one which is generated from the code) related to widgets. Could you please point me/give me some hint?

PR #1160, opened 2018-03-19
Fix for Tile layout where master windows are reversed on reset

nazar-pc:
This is a followup to #1144.
Since we add master windows to the beginning of the list, we need to go in reverse order, otherwise master windows will end up in the reverse order each time `resetMaster()` is called


flacjacket:
Good catch, thanks!

PR #1165, opened 2018-04-05
bugfix for wallpaper when user use custom wallpaper_command

JackTheMico:
the cur_image should be removed after "subprocess.call" when a user custom wallpaper_command like this:
`wallpaper_command = [
    'feh', '--bg-max'
]`


flacjacket:
Great, thanks for the fix!

PR #1167, opened 2018-04-06
docs: qtile is official package now on Arch Linux

lisongmin:


flacjacket:
This is great! Thanks for the doc update, I've applied it to develop

PR #1173, opened 2018-05-19
Check for existence of BAT_DIR before listing it.

ensra:
Fixes a crash on systems without sysfs.


jhdulaney:
Since the check is failing on unrelated code, merging.

flacjacket:
Thanks for the fix!

PR #1175, opened 2018-05-31
checkupdate widget: colour_no_updates argument was not working 

Sweenu:
The `_set_colour()` function expects an int to work properly. A string was passed instead.


flacjacket:
Good catch, thanks for the fix!

PR #1182, opened 2018-06-15
create cache dir if necessary

tych0:
While this directory will usually exist by virtue of the cache socket, I
recently noticed that if one deletes their qtile cache dir while qtile is
running, the prompt won't work any more, since it gets ENOENT when trying
to write history. Let's fix that by creating the cache dir if necessary.

it would be nicer to use os.makedirs(..., exists_ok=True), but that only
works in python >= 3.2, and since we still support 2.7...

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -34,6 +34,7 @@
 import glob
 import os
 import pickle
+import six
 import string
 from collections import OrderedDict, deque
 
@@ -561,10 +562,13 @@ class Prompt(base._TextBox):
 
                 if self.position < self.max_history:
                     self.position += 1
-                try:
-                    os.makedirs(os.path.dirname(self.history_path))
-                except FileExistsError:
-                    pass
+                if six.PY3:
+                    os.makedirs(os.path.dirname(self.history_path), exist_ok=True)
+                else:
+                    try:
+                        os.makedirs(os.path.dirname(self.history_path))
+                    except OSError:  # file exists
+                        pass
                 with open(self.history_path, mode='wb') as f:
                     pickle.dump(self.history, f, protocol=2)
             self.callback(self.userInput)
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -561,6 +561,10 @@ class Prompt(base._TextBox):
 
                 if self.position < self.max_history:
                     self.position += 1
+                try:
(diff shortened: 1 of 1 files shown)


ramnes:
Don't you want to add both version with a Python version check, so that the day we drop Python 2 we can keep only the right one?

tych0:
Arg, seems python2 doesn't have os.FileExistsError. Ok, fine :)

libqtile/widget/prompt.py
```diff
@@ -561,10 +561,14 @@ def _send_cmd(self):
 
                 if self.position < self.max_history:
                     self.position += 1
-                try:
-                    os.makedirs(os.path.dirname(self.history_path))
-                except FileExistsError:
-                    pass
+                if six.PY3:
+                    os.makedirs(os.path.dirname(self.history_path), exist_ok=True)
+                else
```
flacjacket: `:`
:)

tych0: Derp derp :)

libqtile/widget/prompt.py
```diff
@@ -561,6 +561,14 @@ def _send_cmd(self):
...
                 if self.position < self.max_history:
                     self.position += 1
+                if six.PY3:
+                    os.makedirs(os.path.dirname(self.history_path), exist_ok=True)
+                else:
+                    try:
+                        os.makedirs(os.path.dirname(self.history_path))
+                    except OSError:  # file exists
+                        pass
+                else:
```
flacjacket: Ah, and this `else` is unnecessary

tych0:
On Tue, Jun 19, 2018 at 05:07:04AM -0700, Sean Vig wrote:
> flacjacket commented on this pull request.
> 
> 
> 
> > @@ -561,6 +561,14 @@ def _send_cmd(self):
>  
>                  if self.position < self.max_history:
>                      self.position += 1
> +                if six.PY3:
> +                    os.makedirs(os.path.dirname(self.history_path), exist_ok=True)
> +                else:
> +                    try:
> +                        os.makedirs(os.path.dirname(self.history_path))
> +                    except OSError:  # file exists
> +                        pass
> +                else:
> 
> Ah, and this `else` is unnecessary

heh, yes, it is. how did that get there? :)

flacjacket:
Last but not least, you need to import six :)

flacjacket:
Thanks!

PR #1187, opened 2018-07-18
Fix floating bug in bsp layout

heitzmann:
Bsp layout has a bug in which it loses track of the current client upon setting a window to floating/fullscreen.
This PR fixes that bug.


flacjacket:
Awesome, thanks!

PR #1191, opened 2018-07-22
Fix 0 edge-case for resize screen.

kes31:
The resize code prevented the setting of 0 to x and y.


flacjacket:
Good catch, thanks!

PR #1192, opened 2018-07-22
fix fake_screens setting to work.

kes31:
fixed the reading of the fake_screens setting. This makes it possible to split real screens into virtual screens again.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -66,6 +66,7 @@ class Config(object):
                 value = getattr(self, key, default[key])
             setattr(self, key, value)
         self._init_deprecated(**settings)
+        self._init_fake_screens(**settings)
 
     def _init_deprecated(self, extensions=None, **settings):
         "Initialize deprecated settings."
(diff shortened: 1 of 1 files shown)


libqtile/confreader.py
```diff
@@ -75,6 +76,14 @@ def _init_deprecated(self, extensions=None, **settings):
...
             self.extension_defaults.update(extensions.get('dmenu', {}))
 
+    def _init_fake_screens(self, **settings):
+        " Initiaize fake_screens if they are set."
+        try:
+            value = settings['fake_screens']
+            setattr(self, 'fake_screens', value)
+        except KeyError:
+            pass
+
```
ramnes: Why not just adding `"fake_screens",` in the list of `settings_keys`?

ramnes:
I fail to understand what this PR permits to achieve. You say that it "fixes" fake screens, but there is no mention of `fake_screens` in `libqtile/confreader.py` history. Can you explain with more details, and provide an example configuration and screenshot of what it does?

kes31:
As I'm not using qtile anymore, I cannot provide a screenshots. But you can have a look at your documentation http://docs.qtile.org/en/latest/manual/config/screens.html?highlight=fake_screens
There are fakescreens referenced and also in source code they are prepared. But you'll see, that they do not work without this patch.

libqtile/confreader.py
tych0: +1, I think that should work.

PR #1193, opened 2018-07-25
Add monitor information to the qtile instance

Sweenu:
This PR takes advantage of the `Monitor` information available as of RandR 1.5 in order to facilitate multi-monitor setups by offering a list of active monitor outputs and their characteristics without the need to parse `xrandr`'s output. It complements `qtile.conn.pseudoscreens` with other, useful, information.

Dynamic monitor configuration becomes very easy with the use of the `screen_change` hook. A quick example, for a 2-monitor setup:
```python
@hook.subscribe.screen_change
def configure_monitors(qtile, ev):
    monitors = qtile.conn.monitors
    nb_monitors = len(monitors)
    if nb_monitors == 1:
        subprocess.run('xrandr --auto'.split())
    elif nb_monitors == 2:
        subprocess.run(f'xrandr --output {monitors[0].name} --left-of {monitors[1].name}'.split())
    qtile.cmd_restart()
```

### DIFF
--- a/libqtile/xcbq.py
+++ b/libqtile/xcbq.py
@@ -319,6 +319,22 @@ class PseudoScreen(object):
         self.height = height
 
 
+class Monitor(object):
+    """
+        This represents a monitor.
+    """
+    def __init__(self, name, primary, automatic, x, y, width, height, mwidth, mheight):
+        self.name = name
+        self.primary = primary
+        self.automatic = automatic
+        self.x = x
+        self.y = y
+        self.width = width
+        self.height = height
+        self.mwidth = mwidth
+        self.mheight = mheight
+
+
 class Colormap(object):
     def __init__(self, conn, cid):
         self.conn = conn
@@ -353,11 +369,14 @@ class Xinerama(object):
 
 class RandR(object):
     def __init__(self, conn):
+        self.conn = conn
         self.ext = conn.conn(xcffib.randr.key)
         self.ext.SelectInput(
             conn.default_screen.root.wid,
             xcffib.randr.NotifyMask.ScreenChange
         )
+        version_reply = self.ext.QueryVersion(1, 5).reply()
+        self.version = version_reply.major_version, version_reply.minor_version
 
     def query_crtcs(self, root):
         crtc_list = []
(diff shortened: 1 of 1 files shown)


Sweenu:
If someone could help me figure out why `test_command` has fails, that would be super helpful!

libqtile/xcbq.py
```diff
@@ -353,25 +369,46 @@ def query_screens(self):
...
 
     def query_crtcs(self, root):
         crtc_list = []
         for crtc in self.ext.GetScreenResources(root).reply().crtcs:
             crtc_info = self.ext.GetCrtcInfo(crtc, xcffib.CurrentTime).reply()
             crtc_dict = {
-                "x": crtc_info.x,
-                "y": crtc_info.y,
-                "width": crtc_info.width,
-                "height": crtc_info.height,
```
ramnes: That kind of change is noise to your PR and to the project history.

libqtile/xcbq.py
```diff
@@ -850,8 +887,18 @@ def __init__(self, display):
                 setattr(self, i, self._extmap[i](self))
 
         self.pseudoscreens = []
-        if "xinerama" in extensions:
-            for i, s in enumerate(self.xinerama.query_screens()):
```
ramnes: Here also.

ramnes:
Looks pretty good to me overall, and I'm pretty sure that would be something I'd hack with, although I'm not sure to understand all the ins and outs of the different classes and concepts involved here (what's the difference between a `PseudoScreen` and a `Monitor`? what's a CRTC and why doesn't it also have its own class?).

libqtile/xcbq.py
Sweenu: Here, the idea was that Xinerama is less used nowadays than RandR so it makes sense to only test Xinerama if "randr" is not in extensions. 
But it's details, should I rollback that change?

Sweenu:
The best resource I found was [this documentation of the RandR protocol](https://cgit.freedesktop.org/xorg/proto/randrproto/tree/randrproto.txt). It helped me a lot to understand.

CRTC is an acronym for "Cathode Ray Tube Controller" but was kept for historical reasons I guess.
They do have their own class: `PseudoScreen`. It was called that to be more friendly I suppose but the class could have been called `CRTC`, it's all the same.

I wouldn't want to say something wrong so I'm not gonna elaborate on crtc's (the documentation explains better) but the `Monitor` information is much more useful for the end user for sure. What you'll have in `qtile.conn.monitors` is the list of all monitors connected along with their names, x & y position, size in pixel, size in millimeters, a boolean for if it's automatically configured (with a `xrandr --auto` for example) and a boolean for if it's the primary monitor (set with the `--primary` flag of `xrandr`).

One neat practical thing I did with that for example, is a dynamic configuration of different keybindings for each monitor's backlight. So `XF86MonBrightnessUp` and ``XF86MonBrightnessDown` controls my laptop's backlight and `mod + Up` and `mod + Down` controls my external monitor's backlight.

libqtile/xcbq.py
ramnes: I thought the diff was just about the quotes style, I didn't see the `elif` being added. That makes it a good point to explain why we should only do those style changes in separated commits/PR! :stuck_out_tongue:

flacjacket:
@Sweenu are you up to trying to figure out the test failures? It looks like making this change breaks a lot of the tests as it changes the way the screens are discovered by the tests.

Sweenu:
I see. I will look into that as soon as I get some time!
Do you have any leads on where I should look for the discovering of screens in the tests?

PR #1194, opened 2018-07-28
Indicate reboot required

kyokley:
For Ubuntu distros, this PR adds an optional indicator to the check_updates widget to represent that a system reboot is required.


flacjacket:
This looks good, thanks for the update

PR #1195, opened 2018-07-31
change invocation to pytest

obestwalter:
pytest supports and recommends its invocation via `pytest` since 3.0


obestwalter:
For reference: https://github.com/pytest-dev/pytest/issues/1629#issue-161422224

flacjacket:
Great, thanks for the fix!

PR #1197, opened 2018-08-04
Make it possible to use a custom command for the CheckUpdate widget

Sweenu:
This PR adds a way to use a custom shell command in order to check the number of updates. 
For example, to also check AUR packages, I use this command:
`checkupdates && pikaur -Syu 2>/dev/null | sed "/^\s*$/d"` (pikaur is the AUR helper I use)

I didn't try and use the `self.subtr` logic but rather let the user take care of it in the command itself. Because it can be more subtle than just subtracting a fixed number; for example in the command above, if, and only if, no AUR packages are to be updated, it will still print out an empty line so I needed to take care of this special case with a sed command.

So I made  the concession to use an unfortunate `shell=True` in order to allow more flexibility. Let me know what you think!


flacjacket:
Thanks!

PR #1212, opened 2018-09-29
Make it possible to set layout via cmd_addgroup

AartGoossens:


ramnes:
We might want to keep explicit arguments names for documentation and other introspection stuff. Apart from that, looks good to me!

AartGoossens:
Thanks @ramnes for replying. Took me a while to get back to this PR...

I expected this comment but there is a reason that I did it this way: I understand that for documentation purposes it is good to keep the method signature clear of `*args` and `**kwargs` but then every time the arguments for `Qtile.addGroup` are changed they need to be changed for this method (`cmd_addgroup`) too. That is what went wrong here and what I am fixing with this PR. Fixing it like I did now makes sure this cannot happen in the future.
Either way, I am fine with both approaches and definitely want to go for how it is done elsewhere in this project. Maybe the question is if documentation is generated for this method (`cmd_addgroup`) or the method `addGroup`. Do you know that?

AartGoossens:
No worries @ramnes. Thanks for taking the time to answer.
The links show that I definitely shouldn't go the *args,**kwargs-way. :) I changed my code an pushed again.

libqtile/manager.py
```diff
@@ -1688,9 +1688,9 @@ def f(cmd):
             return
         mb.startInput(prompt, f, "qshell")
 
-    def cmd_addgroup(self, group, label=None):
```
ramnes: While I indeed prefer `name` than `group`, we should keep `group` so that we avoid breaking configs of people using the keyword. :man_shrugging:

ramnes: On the opposite, but for the same reason, we should keep `label` as the second argument!

AartGoossens: Oh that is sloppy work from me, sorry about that. Fixed and pushed again.

ramnes:
Thank you very much @AartGoossens for your contribution and reactive answers. :+1:

ramnes:
I should have check the build before merging, thought!

https://travis-ci.org/qtile/qtile/jobs/458702288#L964

AartGoossens:
Oh shoot. I didn't manage to run the tests locally so missed that. I will create a PR to fix it.

PR #1216, opened 2018-10-21
add a dirty memory widget for watching a progress of disk write.

sux2mfgj:
We can guess the status of disk write. This module is convenient when we use sync, umount and so on.
dirty


ramnes:
No need for a new widget, you can already do this with `Memory`!

```python
        Memory(fmt="Used: {MemUsed}M - Total: {MemTotal}M - Dirty: {Dirty}M")
```

PR #1217, opened 2018-10-22
add widget caps_num_lock_indicator

soulchainer:
This is a really simple widget I did for myself, but now I share it, just in case it could be also useful for someone else.
I bought a laptop this summer and it doesn't have any led indicator to warn if Caps Lock or Num Lock is on or off and it was making me crazy in some situations (like typing passwords in terminal emulator, as example, but not the only case) xD. So I did this as a helper/alternative, to cover the absence of that leds.
It is useful for me, at least.

### DIFF
--- a/libqtile/widget/__init__.py
+++ b/libqtile/widget/__init__.py
@@ -32,8 +32,8 @@ from .windowname import WindowName  # noqa: F401
 
 
 def safe_import(module_name, class_name):
-    safe_import_((".widget", module_name), class_name, globals(),
-                 fallback=make_error)
+    safe_import_(
+        (".widget", module_name), class_name, globals(), fallback=make_error)
 
 
 safe_import("backlight", "Backlight")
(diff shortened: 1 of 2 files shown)


libqtile/widget/caps_num_lock_indicator.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def status(self):
+        """Return a string with the Caps Lock/Num Lock status."""
+        indicators = self.get_indicators()
+        status = " ".join([" ".join(indicator) for indicator in indicators])
+        return status
+
+    def poll(self):
+        """Poll content for the text box."""
+        return self.status()
```
ramnes: Why not put the whole content of the `status` function directly there? :slightly_smiling_face:

libqtile/widget/__init__.py
```diff
@@ -32,17 +32,19 @@
...
 safe_import("battery", ["Battery", "BatteryIcon"])
 safe_import("currentlayout", ["CurrentLayout", "CurrentLayoutIcon"])
 safe_import("currentscreen", "CurrentScreen")
 safe_import("debuginfo", "DebugInfo")
-safe_import("graph", ["CPUGraph", "MemoryGraph", "SwapGraph", "NetGraph",
-                      "HDDGraph", "HDDBusyGraph"])
+safe_import("graph", [
+    "CPUGraph", "MemoryGraph", "SwapGraph", "NetGraph", "HDDGraph",
+    "HDDBusyGraph"
+])
```
ramnes: Please keep the diff related to your feature.

libqtile/widget/caps_num_lock_indicator.py
soulchainer: You are so right.
IDK what I was thinking there. I will fix it and update the PR, thanks.

libqtile/widget/__init__.py
soulchainer: This was my code formatter, I think. I didn't do it on purpose.
I will revert it. Sorry for the inconvenience!.

ramnes:
Thanks for the changes @soulchainer!

Last thing: the license is not the one we use in Qtile codebase. Can you switch to MIT, please?

soulchainer:
> Thanks for the changes @soulchainer!
> 
> Last thing: the license is not the one we use in Qtile codebase. Can you switch to MIT, please?

So, I guess, the same that other widgets has. Already changed, passing CI :).
Mental note: I had to change that in another widget of mine. Will make another PR tonight with that change, I guess :).

ramnes:
Thanks!

PR #1219, opened 2018-10-25
Add a function to rename of current window.

sux2mfgj:
In the current implementation, sometimes, there are windows which are the same names. It makes me difficult to find a window that I want from that reason. so, I added.

we can use the function via following configuration, but it is one of the examples.
```
...
Keys = [
    Key([mod], 'i', lazy.widget['prompt'].exec_general(
        'window name:',
        'layout',
        'rename_window'))
]
...
```
- gif

output

### DIFF
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -637,6 +637,12 @@ class TreeTab(Layout):
         self.panel_width -= 10
         self.group.layoutAll()
 
+    def cmd_rename_window(self, name):
+        win = self._get_window()
+        if win:
+            win.name = name
+            self.group.layoutAll()
+
     def _create_drawer(self):
         if self._drawer is None:
             self._drawer = drawer.Drawer(


libqtile/layout/tree.py
```diff
@@ -637,6 +637,12 @@ def cmd_decrease_ratio(self):
         self.panel_width -= 10
         self.group.layoutAll()
 
+    def cmd_rename_window(self, name):
+        win = self._get_window()
+        if win:
+            win.name = name
+            self.group.layoutAll()
```
ramnes: Why the `self.group.layoutAll()`?

ramnes:
What happens if the window itself change its name? For example, a terminal that shows the current path in the window's name? Shouldn't we force the given name to be kept intact?

sux2mfgj:
I have to consider the case that when the window's' name changed. Thank you for your comment. I'll resolve the problem with any method, and resend a new PR. So I close this PR.

PR #1221, opened 2018-10-31
add a command to obtain a current log level of qtile.

sux2mfgj:
I also add a test for related commands.


libqtile/manager.py
```diff
@@ -1196,6 +1196,9 @@ def cmd_critical(self):
         logger.setLevel(logging.CRITICAL)
         logger.critical('Switching to CRITICAL threshold')
 
+    def cmd_loglevel(self):
+        return (logging.getLevelName(logger.level), logger.level)
```
ramnes: Why do we need both? Can't we just have enough with one?

test/test_manager.py
```diff
@@ -28,6 +28,7 @@
 import pytest
 import subprocess
 import time
+import logging
```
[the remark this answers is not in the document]
sux2mfgj: ok. I fixed it with a new commit.

libqtile/manager.py
sux2mfgj: The logger.level is represented by numerical values. [a link of the reference](https://docs.python.org/3/howto/logging.html#logging-levels).
It is difficult to understand for me, so I wrote both. Is it better a idea to remove the left one (loggin.getLevelName(logger.level))?

ramnes: What would you think of doing two commands, `cmd_loglevel` and `cmd_loglevelname`?

test/test_manager.py
```diff
@@ -39,6 +39,8 @@
 import libqtile.hook
 import libqtile.confreader
 
+import logging
```
ramnes: Haha, it should have been ordered alphabetically but still in the first block, as it is an import from the standard library. I should have mentioned it, sorry! But honestly that is no big deal, I'll take that on myself if you want, as we don't have any clear rule stated so far anyway.

libqtile/manager.py
sux2mfgj: I think it is good. I'll make a commit.

test/test_manager.py
sux2mfgj: oh.. I'm sorry the simple my mistake. The rule is mentioned in pep8. Usually, Qtile developers are following that rule, I think.

sux2mfgj: I fixed it!

libqtile/manager.py
sux2mfgj: Sorry to bother you again. I changed the test.

ramnes:
Thank you!

PR #1223, opened 2018-11-01
Fully rewrite the wallpaper widget

ValentijnvdBeek:
Hello, 

This is a bit of a rewrite of the wallpaper widget. I haven't been able to find a test for this specific widget nor manual documentation so if there or you want me to add some please say so.

It changes the following things:
- Instead of separate conflicting options it now just takes a wallpaper option which can be one of two special values[1].
- Wallpapers can now be changed on a per interval basis as well
- It can now take multiple directories instead of one
- It will recursively look for images within those folders
- It obeys [xdg user dirs when finding the default picture folder](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/)[2]

I hope that the changes to the changelog are also okay and not too verbose.

Thank you for your time,

Valentijn.

[1] I think it would be better for this to be an unnamed required argument instead to avoid it being weirdly named but I couldn't find a precedent for this in the other widgets.

[2] To explain why this code is a bit weird: these values aren't actually loaded into the environment but instead can only be found using a subprocess call to a separate tool or parsing the file yourself. I went for the latter.

### DIFF
--- a/libqtile/widget/wallpaper.py
+++ b/libqtile/widget/wallpaper.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2015 Muhammed Abuali
+# Copyright (c) 2018 Valentijn van de Beek
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+    on an interval or if you press on the widget.
+    """
     defaults = [
-        ("directory", "~/Pictures/wallpapers/", "Wallpaper Directory"),
-        ("wallpaper", None, "Wallpaper"),
-        ("wallpaper_command", None, "Wallpaper command"),
-        ("random_selection", False, "If set, use random initial wallpaper and "
-         "randomly cycle through the wallpapers."),
+        ("wallpaper", "cycle", "the path to a wallpaper or the special value cycle or random"),
+        ("directories", pictures_dir, "Either an absolute or relative path to directory or "
```
flacjacket: This should note that directories are only used for cycle and random.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+        """
+        Return the current wallpaper
+        """
+        return self.wallpaper
+
+    def _get_wallpapers_in_directory(self, directory):
+        """
+        Add all the wallpapers in a given directory to the images list
+        """
+        if directory[0] not in ('~', '/'):
```
flacjacket: This should `os.path.expanduser`, then `os.path.isabs` to check this.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+            directory = pictures_dir + directory
+
+        path = pathlib.Path(directory).expanduser()
+
+        for ext in ('jpg', 'png', 'jpeg'):
```
flacjacket: I'd change this to `.jpg`, `.png`, and `.jpeg`.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+        return self.wallpaper
+
+    def _get_wallpapers_in_directory(self, directory):
+        """
+        Add all the wallpapers in a given directory to the images list
+        """
+        if directory[0] not in ('~', '/'):
+            directory = pictures_dir + directory
+
+        path = pathlib.Path(directory).expanduser()
```
flacjacket: Pathlib is sadly also Python 3 only.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+        """
+        Create a list of all the images in a given directory.
+        """
+        if type(self.directories) is str:
+            self._get_wallpapers_in_directory(self.directories)
+
+        for directory in self.directories:
+            self._get_wallpapers_in_directory(directory)
+
+    def set_wallpaper(self, *args):
```
flacjacket: You should drop the `*args` here.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+            if self.random:
+                self.index = random.randint(0, len(self.images) - 1)
             else:
-                self.images.append(self.wallpaper)
-        cur_image = self.images[self.index]
+                self.index += 1
+                self.index %= len(self.images)
+
+            image = str(self.images[self.index])
+            print(image)
```
flacjacket: It would be better if you logged this, rather than print it.

libqtile/widget/wallpaper.py
```diff
@@ -21,84 +22,128 @@
...
+            self.text = os.path.basename(image)
         else:
             self.text = self.label
-        if self.wallpaper_command:
-            self.wallpaper_command.append(cur_image)
-            subprocess.call(self.wallpaper_command)
-            self.wallpaper_command.pop()
+
+        try:
+            subprocess.Popen([self.command, *self.args, image])
```
flacjacket: I think `call` or `check_call` would be better.

ramnes:
This PR has been closed automatically because we removed `develop`, which was targeted by this branch; feel free to open a new PR that targets `master`!

PR #1229, opened 2018-11-18
handle DPI in a more reasonable way

tych0:
I'm now using a docking station with my laptop, instead of two separate machines. Except that the laptop screen and the monitors have vastly different DPIs, so I can't hardcode it by hostname any more.

Here's some code that computes the right DPI and sets it up globally.

### DIFF
--- a/libqtile/ffi_build.py
+++ b/libqtile/ffi_build.py
@@ -35,6 +35,7 @@ pango_ffi.cdef("""
     typedef ... PangoLayout;
     typedef ... PangoFontDescription;
     typedef ... PangoAttrList;
+    typedef ... PangoCairoFontMap;
     typedef enum {
         PANGO_ALIGN_LEFT,
         PANGO_ALIGN_CENTER,
@@ -139,6 +140,15 @@ pango_ffi.cdef("""
     gchar *
     g_markup_escape_text(const gchar *text,
                          gssize length);
+
+    // https://developer.gnome.org/pango/stable/pango-Cairo-Rendering.html
+    PangoCairoFontMap *
+    pango_cairo_font_map_get_default (void);
+    void
+    pango_cairo_font_map_set_default (PangoCairoFontMap *fontmap);
+    void
+    pango_cairo_font_map_set_resolution (PangoCairoFontMap *fontmap,
+                                         double dpi);
 """)
 
 xcursors_ffi = FFI()
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -317,6 +317,13 @@ class Qtile(command.CommandObject):
             self._glib_loop = None
 
     def configure_dpi(self):
+        # don't set the dpi if it's already been set.
+        resources = self.root.get_property("RESOURCE_MANAGER", type="STRING", unpack=str)
+        if resources:
+            for r in resources.split("\n"):
+                if r.lower().startswith("xft.dpi"):
+                    return
+
         # For whatever reason, self.conn.default_screen.width_in_millimeters is
         # basically just wrong most of the time. Instead, we query xrandr and
         # add up each of the outputs individually.
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -58,6 +58,7 @@ from . import hook
 from . import utils
 from . import window
 from . import xcbq
+from . import pangocffi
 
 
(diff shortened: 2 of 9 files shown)


ramnes:
Looks neat!

A few questions:

* why setting a "scale" rather than DPI directly?
* why not directly using xrandr to set the DPI?
* why not reading Xft settings in `.Xresources` like i3 and  software?

Also, I wonder if we can add tests for this. A "HiDPI" page on the documentation would be great also, but   I guess that can be done later.

tych0:
Yeah, derp on the docs, I just pushed some. Thanks for pointing that out.

libqtile/ffi_build.py
```diff
@@ -139,6 +140,17 @@
...
+    // https://developer.gnome.org/pango/stable/pango-Cairo-Rendering.html
+    PangoCairoFontMap *
+    pango_cairo_font_map_get_default (void);
+    void
+    pango_cairo_font_map_set_default (PangoCairoFontMap *fontmap);
+    void
+    pango_cairo_font_map_set_resolution (PangoCairoFontMap *fontmap,
+                                         double dpi);
+
+    void free(void *);
```
flacjacket: Why is the `free()` needed?

libqtile/manager.py
```diff
@@ -312,6 +315,18 @@ def gobject_thread():
...
             self._glib_loop = None
 
+    def configure_dpi(self):
+        # For whatever reason, self.conn.default_screen.width_in_millimeters is
+        # basically just wrong most of the time. Instead, we query xrandr and
+        # add up each of the outputs individually.
+        dpi = self.conn.randr.find_dpi(self.root.wid)
+        dpi = dpi * self.config.dpi_scale
+
+        self.root.set_property("RESOURCE_MANAGER", "Xft.dpi: %d\n" % dpi,
```
flacjacket: There could be a _slight_ difference between the xft dpi (int) and the pango dpi (float), will this even be noticeable?

libqtile/ffi_build.py
tych0: Oh, this is me testing and being dumb, let me drop that, thanks.

libqtile/manager.py
tych0: I don't think so, unless you have really good eyes. At the end of the day it's all pixel level interfaces anyway, so it really is ints :)

tych0:
Hey guys, it occurred to me that the Right Way to do this might be in cairo itself, so that when we draw images (e.g. for the systray) those are also scaled. I think we can do this with `cairo_scale()`, but cairocffi doesn't have that exposed. I'm going to play with that some more. Anyway, we can either merge this as-is and I can fix it later (maybe by marking the DPI bits as "experimental" for now) or we can just close this. I'm fine with either.

tych0:
So it turns out this works,

https://github.com/tych0/qtile/commit/e07ad9adbd26938fce79718e0910d1d98cf67e05

except that mouse clicks are still calculated based on the original offsets, and not the apparent offsets that are actually displayed. It may be that each widget needs its own pixmap, so we can figure out exactly how things are scaled?

ramnes:
Should we consider this PR as a breaking change?

tych0:
I suppose we could check if Xft.dpi is already set, and if it is not set it based on our calculated DPI. Then it wouldn't break anything.

I've not come up with a good way to do the cairo scaling right now, so I've just punted on it :)

flacjacket:
It looks like this causes considerable changes to the positions reported in the tests, do you know what would be causing this?

tych0:
Huh, I guess the dpi checking didn't solve some of the offset problems. I wonder what's happening there.

libqtile/manager.py
mikelward: Your comment about width in millimeters reminded me of this: https://download.nvidia.com/XFree86/Linux-x86_64/304.137/README/dpi.html

cr1ogen:
I do not know if it will do anything but I use a 4k monitor and so that the application windows and even the Qtile bar are not seen small, I use the following line in the autostart.sh file "xrandr -s 3840x2160 --output DP- 2 --scale 0.5x0.5 & ". Sorry if this does not go with the topic you are dealing with. Thank you

glyh:
What's the status of this issue?

PR #1235, opened 2018-11-23
Fixes failing unittest

AartGoossens:
Fixes typo introduced in https://github.com/qtile/qtile/pull/1212.


flacjacket:
It looks like the tests are still complaining about `ValueError: No such layout: xmonadtall`

flacjacket:
You would have been correct before with `monadtall`, but there is no `MonadTall` layout in the chosen configuration, see the `ManagerConfig` at the top of the file.

ramnes:
Still failing. :disappointed:

I guess it's because MonadTall is not in the list of layouts of the config in use here. Try with `tile`.

ramnes:
That works also!

PR #1239, opened 2018-11-26
Multiple gaps per position

### LINKED ISSUE #1230: Giving a list to Screen(top=...) makes Qtile hang
[The documentation says](http://docs.qtile.org/en/latest/manual/config/screens.html#screen) that we can give a list to `top`, `bottom`, `left` and `right` parameters of `Screen`, but in fact, `qtile` hangs and fail to complete its initialization if we do so.

ramnes:
[The documentation says](http://docs.qtile.org/en/latest/manual/config/screens.html#screen) that we can give a list of `Gap`/`Bar` object to `Screen.top` (and alike), but it looks like it never was the case. So I implemented it.

It fixes #1230 (see #826, also).

To that end, I had to modify everything related to bars in Qtile's IPC land. Since I could not manage to do it with two dimensions (i.e. I could not manage to make `qtile.c.bar["bottom"][0]` work), I hacked something so that it looks like this: `qtile.c.bar["bottom_0"]`. This is indeed suboptimal but I feel that it's the only way without rewriting the whole `CommandTree` logic.

The only thing left is to fix the tests in `test_fakescreen.py`, that are broken but I've no idea why; the error doesn't help at all: `AssertionError: retry failed!`. I need help on this.

### DIFF
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -336,9 +336,8 @@ class Qtile(command.CommandObject):
                 l.finalize()
 
             for screen in self.screens:
-                for bar in [screen.top, screen.bottom, screen.left, screen.right]:
-                    if bar is not None:
-                        bar.finalize()
+                for gap in screen.gaps:
+                    gap.finalize()
 
             logger.info('Removing io watch')
             fd = self.conn.conn.get_file_descriptor()
@@ -1331,10 +1330,10 @@ class Qtile(command.CommandObject):
             width=i.width,
             height=i.height,
             gaps=dict(
-                top=i.top.geometry() if i.top else None,
-                bottom=i.bottom.geometry() if i.bottom else None,
-                left=i.left.geometry() if i.left else None,
-                right=i.right.geometry() if i.right else None,
+                top=[gap.geometry() for gap in i.top],
+                bottom=[gap.geometry() for gap in i.bottom],
+                left=[gap.geometry() for gap in i.left],
+                right=[gap.geometry() for gap in i.right],
             )
         ) for i in self.screens]
         return lst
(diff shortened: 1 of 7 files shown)


tych0:
https://github.com/qtile/qtile/pull/1229/commits/fd0854a5101660a1b23c081be0a6d1a39db2f3d9 might help you debug things. It's in my DPI PR, but perhaps we should just pick it separately?

ramnes:
@tych0, you're my hero. :heart: 

Cherry-picking this into develop, indeed!

ramnes:
Tests turned green. Anyone for a review? :slightly_smiling_face:

libqtile/manager.py
```diff
@@ -1103,7 +1102,12 @@ def _items(self, name):
         elif name == "widget":
             return False, list(self.widgetMap.keys())
         elif name == "bar":
-            return False, [x.position for x in self.currentScreen.gaps]
+            keys = []
+            for gap in self.currentScreen.gaps:
+                position = getattr(self.currentScreen, gap.position)
+                index = position.index(gap)
+                keys.append("{}_{}".format(gap.position, index))
```
tych0: I wonder if we can make "bottom" an alias? or have it be a list of gaps instead of bottom_0?

ramnes: As I said in the original description, I do agree that it's not perfect but I couldn't manage to do this without rewriting the whole `CommandTree` logic.

ramnes:
I've noticed a small bug that prevents this PR to be merged: if you define two bars for a position, the space will be allocated for both, but they will be drawn without offset (i.e. one on top of the other).

For example, with a `size=22` grey bar, and then a `size=10` black bar (the dark blue thing being my background) :
example

ramnes:
Note to myself: try to rebase on top of #1299.

PR #1241, opened 2018-12-02
Initialise fake_screens and add documentation

### LINKED ISSUE #1238: Fake screens
[This part of the documentation](https://github.com/qtile/qtile/blob/ed7198a5fb5438110f81a8c6ecc0e6289676c057/libqtile/config.py#L231-L232) mentions "fake screens", and the term is also found [in the code](https://github.com/qtile/qtile/blob/7c2a88fba68bdcf6f25dfb5494a74afc475d674e/libqtile/manager.py#L357-L373).

What are they? How to use them?

We need to document answers to those questions, and then make sure they work correctly.

See #1192 for this last point.

Melkor333:
closes #1238  and as a replacement of #1192:

Initialize fake_screens if they exist plus some documentation.

I literally just copied the documentation from the file [test/test_fakescreen.py](https://github.com/qtile/qtile/blob/develop/test/test_fakescreen.py) and trimmed it a bit. I also added a short oneline explanation. I hope this is enough :)

### DIFF
--- a/docs/manual/config/screens.rst
+++ b/docs/manual/config/screens.rst
@@ -38,6 +38,96 @@ background="#000000")` will give you a black back ground (the default), while
 :code:`bar.Bar(..., background=["#000000", "#FFFFFF"])` will give you a
 background that fades from black to white.
 
+Fake Screens
+============
+
+instead of using the variable `screens` the variable `fake_screens` can be used to set split a physical monitor into multiple screens.
+They can be used like this:
+
+::
+
+    from libqtile.config import Screen
+    from libqtile import bar, widget
+
+    # screens look like this
+    #     600         300
+    #  |-------------|-----|
+    #  |          480|     |580
+    #  |   A         |  B  |
+    #  |----------|--|     |
+    #  |       400|--|-----|
+    #  |   C      |        |400
+    #  |----------|   D    |
+    #     500     |--------|
+    #                 400
+    #
+    # Notice there is a hole in the middle
+    # also D goes down below the others
+
+    fake_screens = [
+      Screen(
+          bottom=bar.Bar(
+              [
+                  widget.Prompt(),
+                  widget.Sep(),
+                  widget.WindowName(),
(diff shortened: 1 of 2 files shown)


Melkor333:
Sorry I somehow overlooked the notifications on this PR. But it seems to be fixed now, thanks @tych0 for your hint :)

tych0:
LGTM!

flacjacket:
So you need to add `fake_screens` to the default config. I also noticed that I can't launch qtile with an empty fake_screens, as it triggers the fake_screens portion of the manager, even just to have it set. I tried this:
```
diff --git a/libqtile/manager.py b/libqtile/manager.py
index 6ad3a3e6..cd4f4bf2 100644
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -369,7 +369,7 @@ class Qtile(command.CommandObject):
             self.screens.append(s)

     def _process_screens(self):
-        if hasattr(self.config, 'fake_screens'):
+        if self.config.fake_screens is not None:
             self._process_fake_screens()
             return

diff --git a/libqtile/resources/default_config.py b/libqtile/resources/default_config.py
index 4f694cf9..b80bc0a3 100644
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -104,6 +104,7 @@ screens = [
         ),
     ),
 ]
+fake_screens = None

 # Drag floating layouts.
 mouse = [
```
which gets qtile working for me locally, but it causes all of the tests to timeout. If you can fix that, I'd say we can merge this.

Melkor333:
I thought something like that shouldn't happen when I saw the comment on #1192 from ramnes to "just add it to the settings_keys". Maybe the solution from #1192 wasn't all that bad with the separate initialization, I guess it would solve the issue? Or does still something speak against it?

Melkor333:
Should I take the diff from #1192 and push it here, or can you/do you want to merge #1992 directly?

PR #1242, opened 2018-12-05
shutdown: try to be slightly more graceful

tych0:
Right now, qtile just closes the X session, which causes everything in that
session to get a SIGKILL. Instead, let's send a SIGTERM to everyone and
wait a little bit so that maybe they'll be able to save their state in some
nice way.

### DIFF
--- a/libqtile/manager.py
+++ b/libqtile/manager.py
@@ -40,6 +40,7 @@ import xcffib
 import xcffib.xinerama
 import xcffib.xproto
 import six
+import time
 import warnings
 
 from .asyncio_compat import asyncio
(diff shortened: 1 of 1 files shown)


libqtile/manager.py
```diff
@@ -1519,6 +1520,33 @@ def cmd_qtile_info(self):
...
+            except ProcessLookupError:
+                return False
+
+        # give everyone a little time to exit and write their state. but don't
+        # sleep forever (2s).
+        for _ in range(4):
+            pids = list(filter(still_alive, pids))
+            if len(pids) == 0:
+                break
+            time.sleep(0.5)
```
tych0: 0.5 seconds may be a little extreme, probably a time.sleep(0.1) with 10 loop iterations would be better. I can change it depending on what people think :)

ramnes:
What would you think of making an option out of this? For example, `graceful_shutdown`, which value would be the maximum time allowed for the shutdown in seconds. Then you'd just iterate N times where `N * 0.1 = graceful_shutdown`

Also, what about people using `poweroff` or `halt` or stuff like that? Don't we want [those lines](https://github.com/qtile/qtile/blob/7c2a88fba68bdcf6f25dfb5494a74afc475d674e/libqtile/manager.py#L278-L279) to benefit from this also?

libqtile/manager.py
ramnes: I think I'd rather do 0.1 also.

ramnes: Don't we want to use `asyncio.sleep` rather than `time.sleep`?

tych0:
On Wed, Dec 05, 2018 at 02:41:03PM -0800, Guillaume Gelin wrote:
> What would you think of making an option out of this? For example, `graceful_shutdown`, which value would be the maximum time allowed for the shutdown in seconds. Then you'd just iterate N times where `N * 0.1 = graceful_shutdown`
> 
> Also, what about people using `poweroff` or `halt` or stuff like that? Don't we want [those lines](https://github.com/qtile/qtile/blob/7c2a88fba68bdcf6f25dfb5494a74afc475d674e/libqtile/manager.py#L278-L279) to benefit from this also?

Yes, we should probably put this in stop; given that cmd_shutdown()
calls it too, we'll get both for free.

ramnes:
> Yes, we should probably put this in stop; given that cmd_shutdown() calls it too, we'll get both for free.

Eeehhh no, `stop` is also used by `cmd_restart`, sadly. Some little refactoring might be needed here.

tych0:
On Wed, Dec 05, 2018 at 02:56:42PM -0800, Guillaume Gelin wrote:
> > Yes, we should probably put this in stop; given that cmd_shutdown() calls it too, we'll get both for free.
> 
> Eeehhh no, `stop` is also used by `cmd_restart`, sadly. Some little refactoring might be needed here. 

Ah, good catch :). I'll figure something out, thanks.

tych0:
I think making this configurable might be a little *too* much configuration, I just switched it to 0.1/10 iterations, though, hopefully that will work for most people. For me it never slept in either case, so it doesn't seem to matter much.

PR #1250, opened 2018-12-20
Replace /proc hackery with psutil (widget portability)

ekollof:
Start of replacing 'groping for values in /proc' with a cross platform system monitoring interface called psutil.

Already done:

- memory/swap widgets
- CPUGraph/MemoryGraph in graph.py

To be done:
- SwapGraph
- net.py
- NetGraph


ekollof:
Oh, easy fix, hang on

libqtile/widget/memory.py
```diff
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright (c) 2015 Jörg Thalheim (Mic92)
+# -*- coding: utf-8 -*- # Copyright (c) 2015 Jörg Thalheim (Mic92)
```
tych0: ? this seems like noise.

ekollof:
I'll look into thatOp 20 dec. 2018 19:20 schreef Tycho Andersen <notifications@github.com>:@tych0 commented on this pull request.







In libqtile/widget/memory.py:

> @@ -1,5 +1,4 @@

-# -*- coding: utf-8 -*-

-# Copyright (c) 2015 Jörg Thalheim (Mic92)

+# -*- coding: utf-8 -*- # Copyright (c) 2015 Jörg Thalheim (Mic92)



? this seems like noise.

libqtile/widget/memory.py
```diff
@@ -19,25 +18,25 @@
...
     val = {}
-    with open('/proc/meminfo') as file:
-        for line in file:
-            key, tail = line.split(':')
-            uv = tail.split()
-            val[key] = int(uv[0]) // 1000
-    val['MemUsed'] = val['MemTotal'] - val['MemFree']
+    val['MemUsed'] = int(psutil.virtual_memory().used / 1024 / 1024)
+    val['MemTotal'] = int(psutil.virtual_memory().total / 1024 / 1024)
+    val['MemFree'] = val['MemTotal'] - val['MemUsed']
```
ramnes: There's a potential config breakage for people doing stuff [like this](https://github.com/qtile/qtile/pull/1216#issuecomment-438977640) here.

Maybe we could accept that config breakage but then expose all the stuff contained in `psutil.virtual_memory()`, so that people actually can replace what they used to do? We could then drop the useless mapping in names.

For the same reason, I would also add `psutil.swap_memory()` informations here and not in a separate widget (plus I don't feel like the two widgets will benefit from being separated, the code being identical).

libqtile/widget/memory.py
```diff
@@ -19,25 +18,25 @@
...
     return val
 
 
 class Memory(base.InLoopPollText):
     """Displays memory usage"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ("fmt", "{MemUsed}M/{MemTotal}M", "see /proc/meminfo for field names")
+        ("fmt", "{MemUsed}M/{MemTotal}M", "Formatting for field names")
```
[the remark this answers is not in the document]
ekollof: I've made them the same as the /proc interface. I'll put them in the docs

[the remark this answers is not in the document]
ramnes: My sentence was a continuation to:

> Maybe we could accept that config breakage but then expose all the stuff contained in psutil.virtual_memory(), so that people actually can replace what they used to do? We could then drop the useless mapping in names.

Sorry if that wasn't clear!

ramnes:
Any reason that you repeat calls like `psutil.virtual_memory()` rather than keeping and reusing the returned value? I guess it doesn't change much since it's just reading a pseudo-filesystem anyway, but still.

ekollof:
Good point. I guess I yank and paste faster than I think ;)

I'll change it. Thanks for pointing it out

libqtile/widget/memory.py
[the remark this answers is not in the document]
ekollof: Oh yeah that. Well, we could use these as general purpose widgets for portability (most people just want CPU/Mem/Swap usage anyway), and maybe keep the old ones for OS-specific hackery? Maybe rename the original one to linux_memory.py (since it is really OS-specific).

libqtile/widget/memory.py
ekollof: Good point. It would be easy to merge that into memory.py. I basically did the same on graph.py. I'll get on that.

libqtile/widget/memory.py
```diff
@@ -19,25 +19,31 @@
...
+    val['MemTotal'] = int(mem.total / 1024 / 1024)
+    val['MemFree'] = int(mem.free / 1024 / 1024)
     return val
 
 
 class Memory(base.InLoopPollText):
-    """Displays memory usage"""
+    """Displays memory usage
+
+MemUsed: Returns memory in use
```
flacjacket: These indents should be further in, along with other docstrings below.

libqtile/widget/graph.py
```diff
@@ -244,18 +234,16 @@ def update_graph(self):
...
-            if line.lstrip().startswith("total"):
-                pass
-            else:
-                key, tail = line.strip().split(':')
-                uv = tail.split()
-                val[key] = int(uv[0])
-    val['MemUsed'] = val['MemTotal'] - val['MemFree']
+    mem = psutil.virtual_memory()
+    swap = psutil.swap_memory()
+    val['MemUsed'] = int(mem.used / 1024 / 1024)
```
flacjacket: You can do `mem.used // 1024 // 1024` to get the same effect.

libqtile/widget/memory.py
```diff
@@ -19,25 +19,31 @@
...
     val = {}
-    with open('/proc/meminfo') as file:
-        for line in file:
-            key, tail = line.split(':')
-            uv = tail.split()
-            val[key] = int(uv[0]) // 1000
-    val['MemUsed'] = val['MemTotal'] - val['MemFree']
+    val['MemUsed'] = int(mem.used / 1024 / 1024)
+    val['MemTotal'] = int(mem.total / 1024 / 1024)
+    val['MemFree'] = int(mem.free / 1024 / 1024)
```
flacjacket: There are more parameters in psutil that are also in meminfo, we may want to copy more over to limit how much this is a breaking change.

ekollof: I'll pull over whatever I can find :)

libqtile/widget/memory.py
ekollof: I was wondering about that. Will change them.

ramnes:
I took the rebase on myself and merged manually. Thanks!

PR #1254, opened 2018-12-30
Keep track of changing timezones

Etn40ff:
Currently, when using the clock widget without a specified timezone, the
timezone is chosen to be the one currently being used at startup and never
updated. In particular the widget does not reflect changes to the system
timezone. This PS change this behaviour.


ramnes:
Thanks!

PR #1259, opened 2019-01-07
Make widgets.Clock much more robust and simple with pytz

ramnes:
The idea here is mostly to remove the environment's `TZ` hackery (which, I guess, avoid having different clocks on different timezones), and thus making the code way simpler.

It also allows users to pass a `tzinfo` class directly to the widget rather than a string.

### DIFF
--- a/libqtile/widget/clock.py
+++ b/libqtile/widget/clock.py
@@ -38,9 +38,10 @@ class Clock(base.InLoopPollText):
     defaults = [
         ('format', '%H:%M', 'A Python datetime format string'),
         ('update_interval', 1., 'Update interval for the clock'),
-        ('timezone', None, 'The timezone to use for this clock, '
-            'e.g. "US/Central" (or anything in /usr/share/zoneinfo). None means '
-            'the default timezone.')
+        ('timezone', None, 'The timezone to use for this clock, either as'
+         ' string if pytz is installed (e.g. "US/Central" or anything in'
+         ' /usr/share/zoneinfo), or as tzinfo (e.g. datetime.timezone.utc).'
+         ' None means the system local timezone and is the default.')
     ]
     DELTA = timedelta(seconds=0.5)
 
--- a/libqtile/widget/clock.py
+++ b/libqtile/widget/clock.py
@@ -24,6 +24,7 @@
 import time
 from datetime import datetime, timedelta, timezone
 from . import base
+from ..log_utils import logger
 
 try:
     import pytz
@@ -46,8 +47,15 @@ class Clock(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Clock.defaults)
-        if pytz and isinstance(self.timezone, str):
-            self.timezone = pytz.timezone(self.timezone)
+        if isinstance(self.timezone, str):
+            if pytz is None:
+                logger.warning('Clock widget can not infer its timezone from a'
+                               ' string without the pytz library. Install pytz'
+                               ' or give it a datetime.tzinfo instance.')
+            else:
+                self.timezone = pytz.timezone(self.timezone)
+        if self.timezone is None:
+            logger.info('Defaulting to the system local timezone.')
 
     def tick(self):
         self.update(self.poll())
--- a/libqtile/widget/clock.py
+++ b/libqtile/widget/clock.py
@@ -22,25 +22,13 @@
 # SOFTWARE.
 
 import time
-from datetime import datetime, timedelta
-from contextlib import contextmanager
+from datetime import datetime, timedelta, timezone
 from . import base
 
-import os
-
-from datetime import timezone
-
-
-@contextmanager
-def tz(the_tz):
-    orig = os.environ.get('TZ')
-    os.environ['TZ'] = the_tz
-    yield
-    if orig is not None:
-        os.environ['TZ'] = orig
-    else:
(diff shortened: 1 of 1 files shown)


ramnes:
One caveat of that PR though, is that it removes the ability for Python 2 users to set the clock's timezone. The idea is that Python 2 timezones support is pretty bad, so let's not try to deal with it.

ramnes:
Also, if Python 3 users don't have `pytz` installed, they will be forced to use a `tzinfo` directly rather than a timezone string.

libqtile/widget/clock.py
```diff
@@ -24,27 +24,17 @@
...
 from . import base
 
-import os
+try:
+    import pytz
+except ImportError:
+    pytz = False
 
 if six.PY3:
     from datetime import timezone
```
flacjacket: Can you remove this import now?

libqtile/widget/clock.py
```diff
@@ -60,6 +50,8 @@ class Clock(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Clock.defaults)
+        if pytz and isinstance(self.timezone, six.string_types):
```
flacjacket: Maybe use `self.timezone is not None`? Maybe log a warning if pytz is not installed?

libqtile/widget/clock.py
ramnes: I use it line 65

libqtile/widget/clock.py
ramnes: The idea is to let the user pass a tzinfo directly if he has Python 3 but doesn't want to use pytz

tych0:
Can you give a documentation example where users without anything besides the stdlib can use this? I think for a widget so basic as a clock, we shouldn't require any additional plugins or anything.

libqtile/widget/clock.py
```diff
@@ -22,25 +22,13 @@
...
-    yield
-    if orig is not None:
-        os.environ['TZ'] = orig
-    else:
-        del os.environ['TZ']
-    time.tzset()
+try:
+    import pytz
+except ImportError:
+    pytz = False
```
flacjacket: Maybe make this `None` to appease mypy

libqtile/widget/clock.py
```diff
@@ -58,6 +46,8 @@ class Clock(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Clock.defaults)
+        if pytz and isinstance(self.timezone, str):
```
flacjacket: Then here, `pytz is not None`

flacjacket: We may want to log something, one way or another, I'm not familiar with pytz enough to know if it could lead to some unexpected behavior if someone didn't realize that pytz was doing something to the timezone, or if someone expects pytz to resolve the timezone but it silently doesn't because it is not installed.

ramnes: Done!

tych0:
I have to say I still don't love this PR; we break configs that used to work unless people install some other dep, for no real win (in my mind at least) simplicity. If you really want to pass datetime.tzinfo objects to the widget, can't we just add that and keep the old stuff around?

ramnes:
@tych0, I don't particularly want to pass `datetime.tzinfo` objects to the widget. It is just a detail of the implementation so that the feature is available to everyone without making pytz mandatory, thinking of those that won't use this widget in the first place.

What I want is that we do not rely on the system's environment to set the timezone of the clock, but on Python code, so that we make sure that multiple clocks can use multiple timezones, that our code is portable, and safe; having a process modifying and restoring its own environment every second sound like it could lead to bugs due to processes environment inheritance.

As far as I know, Python itself doesn't provide any way to create a `datetime.tzinfo` from a timezone string, so the only way of doing this without relying on the system in use is to use a library. pytz is a super common library that is already pulled in by a lot of Python packages, which is why I choose to use it and not a more powerful one (like Pendulum or Arrow), to minimize config breakage.

Last but not least, we're already doing a big cleaning and breaking a lot of stuff for the next release, so I really don't feel bad adding a change that will only break the configs of those that 1/ use timezones on clock, 2/ don't have pytz already pulled in, and 3/ don't read the changelog while using a software that is not labeled 1.0 yet.

Sorry for the long message, I just hope it will give you a better description of this PR than my first comment which I guess was not clear enough.

ramnes:
Just a quick note to say that [we could use dateutil](https://dateutil.readthedocs.io/en/stable/tz.html) rather than pytz. And both are equally common, I think.

ramnes:
Just took a look at how others (py3status guys) do it, and they use pytz also: https://github.com/ultrabug/py3status/blob/master/py3status/modules/clock.py

tych0:
> What I want is that we do not rely on the system's environment to set the timezone of the clock, but on Python code, so that we make sure that multiple clocks can use multiple timezones, that our code is portable, and safe; having a process modifying and restoring its own environment every second sound like it could lead to bugs due to processes environment inheritance.

Ok, I agree that this is a good thing. I've not seen a problem with the existing code when rendering clocks with multiple timezones, but I do admit that it is a theoretical issue :). I don't want to hold this up any more, so feel free to merge it if you're just waiting on me.

ramnes:
I'll go ahead and merge this. :man_shrugging:

PR #1261, opened 2019-01-10
no margins when single window

uoou:
This is my first time doing a PR so sorry if I did anything wrong.

I just added "single_margin" in very much the same vein as "single_border_width" so that I can have no margin when there's only one window.


ramnes:
Hi @uoou, thanks for the straightforward PR!

PR #1267, opened 2019-01-18
core refactoring: configuration

cortesi:
This PR continues the core refactoring. It includes the following main changes:

- Introduce the `core.base.Core` interface, which all cores will implement. This is a [facade](https://en.wikipedia.org/wiki/Facade_pattern) onto a set of back-end modules that are specific to the back-end. For the moment, we only add the methods needed to make the configuration process core-agnostic.
- Introduce `core.xcore.XCore`, which implements the core interface for X.
- Add a `.validate` method for config objects, which validates the config against a core.

The PR also includes a change to test output. Ordinary events now log with level `debug`, and are not included in test output unless requested. I've added a `--debuglog` flag to pytest to turn enable test output. Previously test output was much too verbose.


tonich-sh:
2019-02-15 11:26:19,779 ERROR libqtile manager.py:_xpoll():L758 Got an exception in poll loop
  File "/home/anton_s/src/qtile/libqtile/core/manager.py", line 736, in _xpoll
    r = h(e)
  File "/home/anton_s/src/qtile/libqtile/core/manager.py", line 1017, in handle_ButtonPress
    if m.start:
AttributeError: 'Drag' object has no attribute 'start'

tonich-sh:
```diff
diff --git a/libqtile/config.py b/libqtile/config.py
index 07f46036..323186e9 100644
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -78,6 +78,10 @@ class Drag(Mouse):
     It focuses clicked window by default.  If you want to prevent it pass,
     `focus=None` as an argument
     """
+    def __init__(self, modifiers, button, *commands, **kwargs):
+        Mouse.__init__(self, modifiers, button, *commands, **kwargs)
+        self.start = kwargs.get("start")
+
     def __repr__(self):
         return "<Drag (%s, %s)>" % (self.modifiers, self.button)
```

PR #1268, opened 2019-01-22
[wip] Shift event reading into Core

cortesi:
- Add `./scripts/run`, a small utility that activates the venv and runs qtile. This is handy for display manager configurations.
- Shift complex initialisation out of the Qtile object constructor.
- Shift the event reading loop into Core. Events are placed on an asynchronous queue, and then consumed by Master. 

One complication with this patch is that it exacerbates the existing raciness of our test suite. Many tests incorrectly rely on synchronous behaviour, and don't succeed reliably. With this patch, these tests are more likely to fail, especially on faster systems. Improving test reliability will require test infrastructure that will be made possible by the core refactoring.


test/conftest.py
```diff
@@ -47,7 +47,7 @@
 SECOND_WIDTH = 640
 SECOND_HEIGHT = 480
 
-max_sleep = 5.0
+max_sleep = 15.0
```
ramnes: Is it really needed? It used to be 20 seconds, but I lowered to 5 as it was super boring to wait more than that in case of real code errors.

ramnes:
> One complication with this patch is that it exacerbates the existing raciness of our test suite.

What is causing this, exactly?

libqtile/core/manager.py
```diff
@@ -718,45 +692,26 @@ def get_target_chain(self, ename, e):
...
+            if ename.endswith("Event"):
+                ename = ename[:-5]
+            logger.debug(ename)
+            try:
+                for h in self.get_target_chain(ename, e):
+                    logger.debug("Handling: %s" % ename)
+                    r = h(e)
+                    if not r:
+                        break
             except Exception:
```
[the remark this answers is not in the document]
cortesi: Right. We'll get rid of these one by one (there are a bunch). At this point I'm being a bit conservative by changing one thing at a time (or trying to at least).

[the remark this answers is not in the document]
tych0: Ok, as long as it's on your radar :)

cortesi:
Regarding the races... these are tricky. On my dev desktop (which is a very fast system) the test suite hangs about 30% of the time, even before any of these refactoring patches. On my laptop, which is a bit slower, the test suite was stable until this async work.

I'm digging in to these, so watch for developments here. We won't proceed until we've solved the test stability issue in one way or another.

libqtile/core/manager.py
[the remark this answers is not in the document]
cortesi: The plan with this bit of the code is to gradually replace the X native events with backend-agnostic pseudo-events, until this bit of code just disappears.

libqtile/core/manager.py
```diff
@@ -707,45 +692,26 @@ def get_target_chain(self, ename, e):
             logger.info("Unhandled event: %r" % ename)
         return chain
 
-    def _xpoll(self):
+    async def process_events(self):
```
flacjacket: Finally, async syntax!

libqtile/core/base.py
```diff
@@ -1,8 +1,27 @@
 from abc import ABCMeta, abstractmethod
+
 import typing
+import asyncio
+
+from . import xcbq
+
+
+class ShutdownEvent:
+    message: str
```
flacjacket: I think this should be removed, I think this defines a class variable `ShutdownEvent.message`, which we don't really want. I feel like this is to appease the mypy linter, if that's the case, you could maybe try the `NamedTuple` in `typing`?

flacjacket: Sounds good!

libqtile/core/xcore.py
```diff
@@ -1,11 +1,71 @@
...
+                if error_code:
+                    error_string = xcbq.XCB_CONN_ERRORS[error_code]
+                    message = "X connection error: %s (%s)" % (error_string, error_code)
+                self.q.put_nowait(base.ShutdownEvent(message))
+                return
+
+            if not e:
+                break
+
+            if e.__class__ not in _IgnoredEvents:
```
flacjacket: Is this check maybe better as `not isinstance(e, _IgnoredEvents)`?

flacjacket:
This seems good, I like the direction this is headed. The test situation is interesting, as this breaks Travis because it slows all of the tests way down. But I think this is great to have, as long as we're careful with manual testing.

libqtile/core/xcore.py
cortesi: I feel `__class__` is slightly more explicit here. That said, for Python 3 these should be more or less equivalent, and I don't have particularly strong feelings about it.

ramnes: I usually try to use the keywords and the language's main structure as much as possible unless I have a good reason not to. I feel that it makes things easier for newcomers.

ramnes: The feeling isn't that strong, haha! It's just some thoughts, you do as you wish.

cortesi: Now we're both being too obliging! Consider it changed! :)

ramnes:
Just reading this PR again and it makes a lot of sense, especially the third commit, but a lot has changed in the meantime and the rebase will probably hurt. What do you guys want to do?

ramnes:
I think it would be almost impossible to rebase at this point, so I'll go ahead and close this. Sorry!

PR #1276, opened 2019-02-02
warp cursor on screen change

tych0:
In the case of a screen change, we're changing the window focus, so we should
warp the cursor to the middle of the newly selected window.


tych0:
Looks like there's some unrelated format errors, I'll send a separate PR for those. But It also seems like the test failures are unrelated? Perhaps this is the raciness that @cortesi mentioned?

flacjacket:
I'm going to guess the other diff will fix this, landing!

Issue #1274, opened 2019-01-30
Laptop battery monitoring on FreeBSD


andkem:
Hello!

The bundled battery monitoring widget only works on Linux. I had a look at adapting it to work on FreeBSD as well, but it was a pain to try and axe support into it because of the lack of a clear interface against batteries.

I ripped most of the code out and defined a well specified battery interface that clearly states what a battery implementation for a specific system is meant to supply. The code written can be found here: https://github.com/andkem/qtile/commit/a17d9cb9e625cfc95b8d147cd67aa2899eee6e75

I have been considering porting the Linux parts of the widgets into this new interface, but since I don't have a Linux laptop at the moment, it would be a bit of work seeing if I can borrow one or boot my own from a USB stick. Hence, I'm checking here to see if there would even be any interest in me porting the Linux parts to my own code base. If no one wants to use my code, there is really no point in doing work porting things I'm never going to use.

Thank you for any response!

flacjacket:
This seems like a good thing to have, particularly with other fixes to our widget compatability in #1250. I'd say open up a PR for it!

ramnes:
Closing since #1277 got merged in! Thanks @andkem.

PR #1277, opened 2019-02-03
Add new BatteryMonitor widget

andkem:
Adds a widget called BatteryMonitor for monitoring the status of the
battery in a laptop. It is written with the aim to provide a structure
that facilitates porting to other systems, such as Linux.

Currently it only supports FreeBSD.

I wrote this for the reasons stated in https://github.com/qtile/qtile/issues/1274. This one only works on FreeBSD because I currently don't have a Linux laptop, but I am willing to work around the issue if no one else feels the call to help port Linux support.

If Linux support is ported to this one, I guess one could rename it and simply replace the old Battery widget.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -29,48 +29,172 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import os
+import re
+import platform
+from abc import ABC, abstractclassmethod
+from enum import Enum, auto, unique
+from subprocess import check_output, CalledProcessError
 from libqtile import bar
 from libqtile.log_utils import logger
 from . import base
-from .. import images
+from .. import images, configurable
 
 from typing import Dict
 
-BAT_DIR = '/sys/class/power_supply'
-CHARGED = 'Full'
-CHARGING = 'Charging'
-DISCHARGING = 'Discharging'
-UNKNOWN = 'Unknown'
 
-BATTERY_INFO_FILES = {
-    'energy_now_file': ['energy_now', 'charge_now'],
-    'energy_full_file': ['energy_full', 'charge_full'],
-    'power_now_file': ['power_now', 'current_now'],
-    'status_file': ['status'],
-}
+@unique
+class State(Enum):
+    CHARGING = auto()
+    DISCHARGING = auto()
+    FULL = auto()
+    EMPTY = auto()
+    UNKNOWN = auto()
 
 
-def default_icon_path():
-    # default icons are in libqtile/resources/battery-icons
-    root = os.sep.join(os.path.abspath(__file__).split(os.sep)[:-2])
-    return os.path.join(root, 'resources', 'battery-icons')
+class _Battery(ABC):
+    """
+        Battery interface class specifying what member functions a
+        battery implementation should provide.
+    """
+    @abstractclassmethod
+    def _update_data(self):
+        """
+            Reads the battery status from the system and stores the
+            result internally for later retrieval. Is called every
+            time battery status is queried.
+
+            No return
+
+            Raises RuntimeError on error.
+        """
+        pass
+
+    @abstractclassmethod
+    def _get_state(self):
+        """
+            Returns the state of the battery.
+
+            Return value is one of:
+                State.CHARGING
+                State.DISCHARGING
+                State.FULL
+                State.EMPTY
+                State.UNKNOWN
+
+            Raises RuntimeError on error.
+        """
+        pass
+
+    @abstractclassmethod
+    def _get_percentage(self):
+        """
+            Returns the percentage remaining in the battery in decimal
+            form.
+
+            Return value:
+                float containing remaining charge percentage in
+                decimal form
+
+            Raises RuntimeError on error.
+        """
+        pass
+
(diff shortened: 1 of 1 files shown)


libqtile/widget/battery_monitor.py
```diff
@@ -0,0 +1,394 @@
...
+import re
+import platform
+from enum import Enum
+from subprocess import check_output, CalledProcessError
+from libqtile import bar
+from . import base
+from .. import images
+
+
+class State(Enum):
```
flacjacket: I'd `@unique` decorate this.

libqtile/widget/battery_monitor.py
```diff
@@ -0,0 +1,394 @@
...
+    CHARGING = 0
```
flacjacket: Might be good to make all of these `auto()`

libqtile/widget/battery_monitor.py
```diff
@@ -0,0 +1,394 @@
...
+    DISCHARGING = 1
+    FULL = 2
+    EMPTY = 3
+    UNKNOWN = 4
+
+
+class _Battery(object):
```
flacjacket: I actually like this a lot more as an interface that we could port the other battery over to. It would be awesome if you could also add some type annotations to it!

It'd be good to make this an `abc.ABCMeta` abstract class and all of the not implemented methods as abstract methods.

libqtile/widget/__init__.py
```diff
@@ -38,6 +38,7 @@ def safe_import(module_name, class_name):
 
 safe_import("backlight", "Backlight")
 safe_import("battery", ["Battery", "BatteryIcon"])
+safe_import("battery_monitor", ["BatteryMonitor", "BatteryMonitorIcon"])
```
flacjacket: Why put this in a different file? Can we just adjust the battery to add this functionality via your FreeBSD implementation?

libqtile/widget/battery_monitor.py
```diff
@@ -0,0 +1,394 @@
...
+
+    def _get_state(self):
+        stat = re.search(r'State:\t+([a-z]+)', self.info)
+
+        if stat is None:
+            raise RuntimeError('Could not get battery state!')
+
+        stat = stat.group(1)
+        if stat == 'charging':
+            result = State.CHARGING
```
flacjacket: No need to assign a variable in all of these, just return the values

libqtile/widget/__init__.py
andkem: I put it in a separate file because it was simpler while working on it. If or when the Linux integration happens, I was planning on putting it back in the old file. I believe I explained it some in the PR or the issue that spawned the PR.

andkem:
I've now rewritten the Linux parts as well as I finally found my USB stick and was able to install Linux on it. I'll get to addressing all the comments and taking a few clean-up sweeps next.

libqtile/widget/battery_monitor.py
andkem: Done

libqtile/widget/battery_monitor.py
andkem: Done

libqtile/widget/battery_monitor.py
andkem: Done

andkem: I wrote the data type in the comment. If you want them in some specific format, please give me an example.

libqtile/widget/battery_monitor.py
andkem: I'm of the opinion that assigning a value to a result and returning it at the end of a function is cleaner and more maintainable as it reduces code complexity. If you feel strongly about it, I can of course change it.

flacjacket:
Awesome, thanks for this! Let us know when you are happy with it and I'll give it a shot!

libqtile/widget/battery_monitor.py
flacjacket: No problem, I'll leave that up to preference.

andkem:
> Awesome, thanks for this! Let us know when you are happy with it and I'll give it a shot!

I'm happy with the code as it is now. I only have a computer that reports current on Linux via sysfs. The part that handles power is thus untested. If someone has a computer reporting power, it would be great if they could try it out.

ramnes:
It works:

ramnes:
Although, there is one thing I miss with this new version of the widget, which is that it used to write `Full` rather than `= 99% 0:00 0.00 W`.

flacjacket:
Check out #1293 and let me know what you think. I started from this PR and built off of that, and maintained exactly the behavior ramnes mentioned, with a config to change it back.

andkem:
> Check out #1293 and let me know what you think. I started from this PR and built off of that, and maintained exactly the behavior ramnes mentioned, with a config to change it back.

I think it looks good. I appreciate making "Full" an option as I prefer to see the text. I've run it on my FreeBSD machine and it still seems to work fine with your changes.

Feel free to refactor or rewrite as you see fit to get the code up to your standard. You can push changes directly to this PR or close it and replace it with your own. I don't mind either way.

flacjacket:
Thanks for this! I opted to just pull in this PR and rebase my tweaks on top of master.

ramnes:
Great, thanks @flacjacket!

PR #1279, opened 2019-02-08
copy default config attributes over to test configs

tych0:
The recent cursor_warp fix meant that now some of the test code configs
need cursor_warp. It's a pain in the ass to add unrelated variables to the
test configs every time we add a new config variable, and it trips a lot of
people up.

Instead, let's just copy everything that doesn't already exist over from
the default config, so nobody has to mess around with this any more.


flacjacket:
This look good, tho it seems a cairocffi bump has now broken all of our tests even more...

PR #1280, opened 2019-02-09
tests: don't leak an fd

tych0:
test/test_hook.py::test_can_call_by_selection_notify
  /home/tycho/packages/qtile/test/conftest.py:195: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/.X358037035-lock' mode='w' encoding='UTF-8'>
    self.display = ":{}".format(_find_display())

The problem here is that Xvfb._get_next_unused_display() opens a lockfile,
but since we're not using Xvfb() in a finally block, we don't close it.
Let's just open-code this, and try to drop the xvfb dep in the future, as
it's only used in one other place, and we could presumably use xcffib's
XvfbTest wrapper instead.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -27,6 +27,7 @@ from libqtile.core import xcore
 from libqtile.log_utils import init_log
 from libqtile.resources import default_config
 
+import fcntl
 import functools
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


tych0:
Hmm, actually we may want to do this slightly differently. This fixes the leak at least, but doesn't really fix any raciness. I'll close this for now.

PR #1284, opened 2019-02-12
custom change_command for backlight widget

zordsdavini:
add possibility to change executed command, ex `light -S %s` as change_command

### DIFF
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -22,6 +22,7 @@
 # SOFTWARE.
 
 import os
+import shlex
 from . import base
 
 try:
@@ -85,7 +86,7 @@ class Backlight(base.InLoopPollText):
         return self.format.format(percent=percent)
 
     def change_backlight(self, value):
-        self.call_process((self.change_command % value).split())
+        self.call_process(shlex.split(self.change_command % value))
 
     def button_press(self, x, y, button):
         if self.future and not self.future.done():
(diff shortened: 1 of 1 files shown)


libqtile/widget/backlight.py
```diff
@@ -84,7 +85,7 @@ def poll(self):
         return self.format.format(percent=percent)
 
     def change_backlight(self, value):
-        self.call_process(["xbacklight", "-set", str(value)])
+        self.call_process((self.change_command % value).split())
```
tych0: Can you use shutil here instead?

zordsdavini: @tych0  why should we use different method as in other places? As I understand you speak about `shutil.run()`

tych0: Oof, sorry. I meant shlex, not shutil. In particular, `shlex.split` instead of `.split()` so that we split things correctly if people use shell syntax in their configs.

libqtile/widget/backlight.py
```diff
@@ -84,7 +86,7 @@ def poll(self):
         return self.format.format(percent=percent)
 
     def change_backlight(self, value):
-        self.call_process(["xbacklight", "-set", str(value)])
+        self.call_process(shlex.split(self.change_command % value))
```
ramnes: Any reason to not use `.format` and its templating rather than `%`?

zordsdavini: refactored to use `format` method

tych0:
Thanks!

PR #1287, opened 2019-02-20
Update check_updates.py

XPOHUK:
Set colour_have_updates only when number of updates > 0
Now this color is set even if the number of updates is 0

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -83,7 +83,7 @@ class CheckUpdates(base.ThreadedPollText):
 
     def _set_colour(self, num_updates):
         # type: (str) -> None
-        if num_updates:
+        if int(num_updates) > 0:
             self.layout.colour = self.colour_have_updates
         else:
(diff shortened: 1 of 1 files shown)


ramnes:
This won't work if we enter the condition in line 78.

flacjacket:
It looks like that is actually the correct behavior, Ramnes, the addition in that block is actually string concatenation.

ramnes:
@flacjacket Because it is a string concatenation, it can't be cast as int.

libqtile/widget/check_updates.py
```diff
@@ -83,7 +83,7 @@ def _check_updates(self):
 
     def _set_colour(self, num_updates):
         # type: (str) -> None
-        if num_updates:
+        if int(num_updates) > 0:
```
ramnes: hackish but works:

```suggestion
        if not num_updates.startswith("0"):
```

ramnes:
Thanks!

XPOHUK:
Not at all!

PR #1290, opened 2019-02-24
broken window list extension fix

zordsdavini:
ah, I've braked window list extension (all dmenu stuff that goes vertically). This is fix

### DIFF
--- a/libqtile/extension/dmenu.py
+++ b/libqtile/extension/dmenu.py
@@ -89,7 +89,7 @@ class Dmenu(base.RunCommand):
 
     def run(self, items=None):
         if items and self.dmenu_lines:
-            lines = min(len(items), self.dmenu_lines)
+            lines = min(len(items), int(self.dmenu_lines))
             self.configured_command.extend(("-l", str(lines)))
 
         proc = super().run()


ramnes:
Can you explain what is wrong without the cast?

zordsdavini:
Searching min between str and int.

ramnes:
@zordsdavini but why is `self.dmenu_lines` a string in the first place? I think this is the real problem.

zordsdavini:
> @zordsdavini but why is `self.dmenu_lines` a string in the first place? I think this is the real problem.

this is because I couldn't save default as int (0) in Dmenu class. Lint problems

ramnes:
I don't understand: the default is `None`, and `None` isn't a string (and also, `int(None)` is not something that works). What default are you talking about? What line?

zordsdavini:
this was from https://github.com/qtile/qtile/pull/1231#issuecomment-462592874 . Maybe I resolved it in wrong way

zordsdavini:
this fixed my issue in command set extension https://github.com/qtile/qtile/pull/1231/commits/1196b9c3db4140d38ffae47329e7010f88ac6ce5 but because of that I got this

flacjacket:
Is this still needed? It looks like it needs a rebase.

zordsdavini:
Will try latest version. If issue still exiats will rebase. But from
discussion earlier I'm not sure have I fixed it in right way. Maybe I
should try to fix typehints and set default value to integer

2019 m. kov. 16 d., št 18:41, Sean Vig <notifications@github.com> rašė:

zordsdavini:
Had the last testings - still can't get window lists. My given patch fixes the issue

flacjacket:
Thanks!

PR #1294, opened 2019-03-16
Bat uses threaded poll

flacjacket:
Builds off of #1293 to have the Battery widget use the threaded poll text base class, rather than doing its own rescheduling and polling.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -298,7 +298,7 @@ class _LinuxBattery(_Battery, configurable.Configurable):
         return BatteryStatus(state=state, percent=percent, power=power, time=time)
 
 
-class Battery(base._TextBox):
+class Battery(base.ThreadedPollText):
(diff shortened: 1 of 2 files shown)


libqtile/widget/battery.py
```diff
@@ -353,6 +343,21 @@ def _get_text(self) -> str:
         except RuntimeError as e:
             return f'Error: {e}'
 
+        return self.build_string(status)
```
ramnes: Why another method? I do not see it in use anywhere else.

ramnes:
LGTM from the diff (didn't try it). :+1:

flacjacket:
Having it like this makes it easy to extend the behavior of widget. You
just have to override this one function to go from status to text, and
don't have to do anything else related to reading the status or handling
polling behavior. I actually do this in my own config to use some Unicode
characters for the battery, haha.

PR #1296, opened 2019-03-16
Change to using setup.cfg

flacjacket:
Change from configuring everything in the setup.py file to using the
newer setup.cfg file for a lot of the metadata associated with the qtile
package.


Ryuno-Ki:
Oh, that's interesting! Do you have a link which describes the difference? (I was asked by the maintainer of another python package whether I'd like to inherit it, but need to learn more about setup.py/setup.cfg and PyPI before I can make a decision).

I can fill another issue for this request if you want.

flacjacket:
It's a feature of newer setuptools, there is some info documenting it over in those docs:

flacjacket:
here: https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

ramnes:
Very nice, I didn't know about that new `setup.cfg` fanciness!

Why not keeping `requirements-dev.txt`, thought? It seems to me than installing a requirements file is the  most common workflow for Python programmers that want to quickly hack stuff (more than `pip install -e .)`.

flacjacket:
That's a good point, added the requirements-dev.txt file back in. It would definitely be nice if there were a way to get just dependencies installed, but it doesn't seem like there is.

PR #1299, opened 2019-03-22
Create new command graph

flacjacket:
The new command graph is completely decoupled from any other
functionality, and can be used to resolve any object or command in the
command graph.  Change over from the current command graph to the new
command graph will occur in additional commits.

Currently I am just putting this up for comments or suggestions.

### DIFF
--- a/libqtile/scripts/qtile_run.py
+++ b/libqtile/scripts/qtile_run.py
@@ -26,7 +26,7 @@ import atexit
 import argparse
 import subprocess
 
-from libqtile import ipc, command, command_graph
+from libqtile import ipc, command_graph
 
 
 def parse_args():
@@ -70,7 +70,7 @@ def parse_args():
 def main() -> None:
     opts = parse_args()
     if opts.socket is None:
-        socket = command.find_sockfile()
+        socket = ipc.find_sockfile()
     else:
         socket = opts.socket
     client = ipc.Client(socket)
--- a/libqtile/scripts/qtile_run.py
+++ b/libqtile/scripts/qtile_run.py
@@ -26,26 +26,39 @@ import atexit
 import argparse
 import subprocess
 
-from libqtile import command
+from libqtile import ipc, command, command_graph
 
 
 def parse_args():
-    parser = argparse.ArgumentParser(description="Run a command applying rules"
-                                     " to the new windows")
-    parser.add_argument('-s', '--socket', type=str, dest="socket",
-                        help='Use specified communication socket.')
-    parser.add_argument('-i', '--intrusive',
-                        dest="intrusive", action='store_true',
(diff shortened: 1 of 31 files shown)


libqtile/scripts/qtile_run.py
```diff
@@ -26,26 +26,39 @@
...
+        action='store_true',
+        help='Do not break on match (keep applying rules).')
+    parser.add_argument(
+        '-g',
+        '--group',
+        help='Set the window group.')
+    parser.add_argument(
+        'cmd',
+        nargs=argparse.REMAINDER,
+        help='Command to execute')
```
ramnes: disturbing indentation

ramnes:
Looks super good with a much cleaner classes API, that might help a lot for #1239! Eager to play with it.

libqtile/scripts/qtile_run.py
flacjacket: Haha, I picked this up as my typical indentation pattern after watching https://www.youtube.com/watch?v=ZsHMHukIlJY, but I'd be happy to change it back.

flacjacket:
Glad you think it is a move in the right direction. I think this should be good to run with, there is still some docs updating and fixing up qtile-cmd and the ipython kernel to work with the new clients and command graph. The next move after that will be to move the lazy infrastructure over to the new command graph (currently that is still on the current command graph, and lazy calls wouldn't be able to use the new APIs). Once that happens, this should be much more fully functional, and we can start really getting rid of some of the old command graph stuff.

flacjacket:
Also, before we land this, we'll probably want to cut a release, since this is going to be a pretty big change that could potentially be pretty breaking.

libqtile/scripts/qtile_run.py
ramnes: I wouldn't mind this type of indentation if the closing parenthesis was on a new line. But to be honest my favorite one is just the one that was there before. :smile: 

And thank you Python for not enforcing an indentation so that we can keep having those passionate discussions. :heart:

flacjacket:
Haha, I typically have the parenthesis on the new line, but thought this
would be a smaller change from before, definitely seems more weird to mix
the styles.

At least we can be glad to not have to deal with opening and closing brace
indentation styles, haha.

ramnes:
I'm really enthusiastic about this but to be honest, I don't think anyone will be able to review the whole thing (either lack of time or knowledge of Qtile's internals), so maybe we should just merge and start again from here? Is there anything specific holding you back?

flacjacket:
Given the breadth of the changes here, I think it'd be good to get a
release out first, then merge this in. I have been running it locally
without issue, but that isn't to say there aren't bugs or behavior changes
others will hit.

PR #1301, opened 2019-03-26
Fix battery widget updating frequency

flacjacket:
When the battery widget was flipped from using manual ticking to using
the threadedpolltext base, the `update_delay` parameter that the battery
widget had been using was not updated to the correct value of
`update_interval`.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -32,6 +32,7 @@
 import os
 import platform
 import re
+import warnings
 
 from abc import ABC, abstractclassmethod
 from enum import Enum, unique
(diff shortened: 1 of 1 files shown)


libqtile/widget/battery.py
```diff
@@ -313,11 +314,15 @@ class Battery(base.ThreadedPollText):
...
         ('low_foreground', 'FF0000', 'Font color on low battery'),
-        ('update_delay', 60, 'Seconds between status updates'),
+        ('update_interval', 60, 'Seconds between status updates'),
         ('battery', 0, 'Which battery should be monitored'),
     ]
 
     def __init__(self, **config) -> None:
+        if "update_delay" in config:
+            warnings.warn("Change from using update_delay to update_interval for battery widget, removed in 0.15",
+                          DeprecationWarning)
```
tych0: Should we patch it in to the right key here as you do below?

flacjacket: Ah! yes totally missed that, thanks for the catch.

Issue #1249, opened 2018-12-20
How to find X,Y position of mouse


ruapotato:
I cannot seem to find the x,y position of the mouse in the qtile scripting commands. Can this be done? 

I found:
`qtile.cmd_get_position()`
But I get the error `qtile has no attribute cmd_get_position`

flacjacket:
What are you looking for with the mouse position? When do you want it? The position at a particular time? The only `cmd_get_position` I see is on the window.

ruapotato:
I'm trying to find the mouse position relative to the screen in a callback function (in this case: handleMouseWheelUP)
```
mouse = [
  Click([mod2], "Button4", lazy.function(handleMouseWheelUp))
]
```

ruapotato:
It looks like I have access to stuff in libqtile.manager
Is it possible to extend libqtile.manager to have a getMousePosition? At the moment I only see: 
mouseMap
grabMouse()

ramnes:
Closing since #1305 got merged. Thanks!

PR #1305, opened 2019-04-07
Track last known mouse XY

ruapotato:
This would solve my issue: 

"How to find X,Y position of mouse"
https://github.com/qtile/qtile/issues/1249

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -80,6 +80,8 @@ class Qtile(command.CommandObject):
 
         self._eventloop = None
         self._finalize = False
+        self.mouseX = 0
+        self.mouseY = 0
 
         if not display_name:
             display_name = os.environ.get("DISPLAY")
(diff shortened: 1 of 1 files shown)


libqtile/core/manager.py
```diff
@@ -80,6 +80,8 @@ def __init__(
 
         self._eventloop = None
         self._finalize = False
+        self.mouseX = 0
+        self.mouseY = 0
```
ramnes:
```suggestion
        self.mouse_position = (0, 0)
```

ramnes:
@flacjacket How would you do this, roughly?

libqtile/core/manager.py
ruapotato: mouse_position is a better name. :)

flacjacket:
That's a good question. It may be that we want to pull x specific handling
out to some other location. I'm not sure how we'd hook that in, tho.

flacjacket:
Thanks! I think we can move this functionality around when we find a better place for all of the other X11 handling stuff in the manager.

PR #1306, opened 2019-04-11
Proposed changes to Mpd2 formatter method

MagusLucius:
The proposed changes fix the following bugs:
    1) logger spamming error messages once per update when tags are missing.
    2) widget crashing (forcing qtile restart to fix) when playlist is empty.

The changes add the following functionality:
    1) a new 'remaining' tag, which is 'fuilltime' - 'elapsed'
    2) a new system for introducing user formatting methods for each tag.

### DIFF
--- a/libqtile/widget/mpd2widget.py
+++ b/libqtile/widget/mpd2widget.py
@@ -246,12 +246,12 @@ class Mpd2(base.ThreadPoolText):
         context = obj or self
         try:
             getattr(context, attr_name)()
-        except (AttributeError, TypeError) as E:
-            if isinstance(E, AttributeError):
+        except (AttributeError, TypeError) as e:
+            if isinstance(e, AttributeError):
                 err = err1.format(Class=type(context).__name__, attr=attr_name)
             else:
                 err = err2.format(Class=type(context).__name__, attr=attr_name)
-            logger.exception(err + " {}".format(E.args[0]))
+            logger.exception(err + " {}".format(e.args[0]))
 
     def toggle(self):
         """toggle play/pause."""
--- a/libqtile/widget/mpd2widget.py
+++ b/libqtile/widget/mpd2widget.py
@@ -1,12 +1,17 @@
+"""
+A widget for Music Player Daemon (MPD) based on python-mpd2.
+
+This widget exists since python-mpd library is no longer supported.
+"""
 from . import base
 from libqtile.log_utils import logger
 
 from socket import error as socket_error
 from mpd import MPDClient, ConnectionError, CommandError
 from collections import defaultdict
+from cgi import escape
 
-
-# Shortcuts
+# Mouse Interaction
 # TODO: Volume inc/dec support
 keys = {
     # Left mouse button
@@ -21,7 +26,6 @@ keys = {
     "command": None
 }
 
-
 # To display mpd state
 play_states = {
     'play': '\u25b6',
@@ -31,6 +35,11 @@ play_states = {
 
 
 def option(char):
+    """
+    old status mapping method.
+
+    Deprecated.
+    """
     def _convert(elements, key, space):
         if key in elements and elements[key] != '0':
             elements[key] = char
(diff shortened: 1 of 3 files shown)


libqtile/widget/mpd2widget.py
```diff
@@ -165,28 +187,59 @@ def __getitem__(self, key):
...
+        song_info = defaultdict(lambda: default)
+        song_info['play_status'] = self.play_states[status['state']]
+
+        for k in current_song:
+            song_info[k] = current_song[k]
+        song_info['fulltime'] = song_info['time']
+        del song_info['time']
+
+        song_info.update(status)
+        if song_info['updating_db'] == default:
```
flacjacket: You can say here `if 'updating_db' not in song_info`

libqtile/widget/mpd2widget.py
```diff
@@ -165,28 +187,59 @@ def __getitem__(self, key):
...
         try:
-            formatted = fmt.format(play_status=play_status, **status)
+            formatted = fmt.format(**song_info)
             return formatted
         except KeyError as e:
             logger.exception("mpd client did not return status: {}".format(e.args[0]))
             return "ERROR"
 
-    def prepare_formatting(self, status, currentsong):
+    # Removed an unused parameter from this function.
```
flacjacket: You can remove this comment

libqtile/widget/mpd2widget.py
```diff
@@ -165,28 +187,59 @@ def __getitem__(self, key):
...
+        else:
+            self.prepare_formatting(song_info)
+
+        # 'remaining' isn't actually in the information provided by mpd
+        # so we construct it from 'fulltime' and 'elapsed'.
+        # 'elapsed' is always less than or equal to 'fulltime', if it exists.
+        # Remaining should default to '00:00' if either or both are missing.
+        if 'remaining' in self.status_format:
+            total = float(song_info['fulltime'])\
+                if song_info['fulltime'] != default else 0.0
```
flacjacket: These as well should use eg `'fulltime' in song_info`

libqtile/widget/mpd2widget.py
```diff
@@ -165,28 +187,59 @@ def __getitem__(self, key):
...
-
-        # Dirty hack to prevent keys conflict
-        currentsong['fulltime'] = currentsong['time']
-        del currentsong['time']
-
-        self.prepare_formatting(status, currentsong)
-        status.update(currentsong)
+    def formatter(self, status, current_song):
+        default = 'Undefined'
+        song_info = defaultdict(lambda: default)
```
flacjacket: It doesn't look like anything here is necessarily needing the defaultdict, rather than a normal dict. What properties do you need to have added to keep the formatting from failing? Maybe you can conditionally add them at the end?

MagusLucius: I apparently have quite a number of corrupted tags in my library, all mp3 files from what I can tell.  MPD refuses to read the tags because of the presence of an invalid frame.  When this happens, the currentsong() method returns an empty dict.  Originally, I was using regex to fish the requested tags out of the format string, but I thought using a defaultdict was a better idea, as it would return a default value for any missing tag, without the overhead of a regular expression.

MagusLucius: I've thought of a few more edits, but I haven't pushed them to my fork yet.
1) For the sake of consistency, I moved the parameters 'prepare_status' and 'status_format' out of the constructor, and put them in the defaults dictionary instead.  I see that's the way the old mpd widget handled things.

2) I added an Idle state message that replaces the regular message when MPD is idle (e.g. status['state'] == 'stop' and current_song == {}) There's a secondary format string that I set to default to "{play_status} {idle_message} [{repeat}{random}{single}{consume}{updating_db}]"

3) I added a default formatting function ('all': lambda s: cgi.escape(s)) which prevents pango from throwing a fit if you use certain characters like "&" (I use this character because I'm using a special font to display the status symbols and the "shuffle" glyph is mapped to "&".  The character also shows up occasionally in artist information for songs. The 'all' function gets applied before any other user formatting.  I'm not sure if this is a great idea or not (partially why I haven't updated the fork) the presence of 'all' in the formatting function dict means that a user would then have to take that function's result into account with every other formatting function in the dictionary. This isn't difficult in the case of a simple escape sequence (I use ast.literal_eval to convert the escaped numbers back into numbers, but it's an extra step).  I did note that in my documentation, however.

let me know what you think.

ramnes:
@MagusLucius can you rebase on master please?

MagusLucius:
> @MagusLucius can you rebase on master please?

Done.  The reason I switched it to develop was that the README said to make PR's against develop.

MagusLucius:
Changed mouse interaction method for more flexibility.
Added configurable Idle message for when playlist is empty
Fixed pep8 errors (including 2247.7 and 2247.8)

MagusLucius:
I edited to add mpd info to the docs and sandboxed the widget internals access for mouse_button, but it isn't showing up here for some reason.

tych0:
Sorry, I misunderstood what was going on here. Can you drop the merge commits, so the history is clean? This LGTM otherwise.

MagusLucius:
> Sorry, I misunderstood what was going on here. Can you drop the merge commits, so the history is clean? This LGTM otherwise.

active_branches

If I understand how to do that.  Can I simply click the trash bin next to #1 and #3?

tych0:
Possibly, I don't know how to do it through the github interface. I'd do a
git rebase -i and squash all the commits together.

tych0:
Ok, I went ahead and squashed this myself. I also added some docs about its deps and deleted the old mpd widget. Thanks for everyone's work on this!

tych0:
Seems like this breaks our doc generations. I've tracked it to these three:

```       ('prepare_status', status_dict,
            'characters to show the status of MPD'),
        ('status_format', status_format, 'format for displayed song info.'),
        ('idle_format', idle_format,
            'format for status when mpd has no playlist.'),
```

Any chance you can take a look at fixing the doc generation?

m-col:
At some point the widget's mpd client has disconnected and it's not getting reconnected, so it's stopped updating. It should probably reconnect upon any action (click or poll) to continue updating.

PR #1307, opened 2019-04-12
Don't make distutils copy man pages to /share/man

rocha:
Running `python setup.py install` fails while trying to copy qtile's man pages to `/share/man`, which requires root permissions. This breaks installing qtile as python package inside a virtualenv for isolation or testing. This was possible before 749dfc9bc


flacjacket:
Thanks!

PR #1308, opened 2019-04-19
Columns layout - add screen focus wrap option and shrink edge columns

mclgn:
These are just some small usability changes to the Columns layout I figured I should give back, as I've gotten more into customising this wm!

First, there are two added options: "wrap_focus_columns" and "wrap_focus_rows". Their default state, enabled, is the current behaviour. When disabled, attempting to focus off-screen will not do anything, rather than focussing the far opposite window.

Second, additional behaviour is added to "cmd_grow_left" and "cmd_grow_right". Previously, if, for example, the left-most column was focussed and "cmd_grow_left" was executed, nothing happened. With these changes, this would instead shrink that column to the left. The result is that if you had two columns, growing left and right would not require you to be focussed on the window you wanted to grow, and you could directly shift the boundary in the centre left/right, which I think feels more natural. With >2 columns, this behaviour is similar for a focussed outside column and the adjacent inside column. When focussed on an inside column, there is the original behaviour of both directions growing that column in the specified direction.

These changes don't change the current behaviour but extend them, so usage wouldn't be disrupted. I'm relatively to new to python so feedback is welcome if there are points it could be cleaner or optimised!

Thanks for the awesome project.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -121,6 +121,10 @@ class Columns(Layout):
         ("insert_position", 0,
          "Position relative to the current window where new ones are inserted "
          "(0 means right above the current window, 1 means right after)."),
+        ("wrap_focus_columns", True,
+         "Wrap the screen when moving focus across columns."),
+        ("wrap_focus_rows", True,
+         "Wrap the screen when moving focus across rows."),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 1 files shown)


flacjacket:
Merging and fixing up style stuff on master. Thanks!
Not sure why those issues aren't getting caught by the linter...

mclgn:
Great, thanks!

PR #1312, opened 2019-04-25
Make the README fancier

ramnes:
See the result here: https://github.com/ramnes/qtile/blob/fancy-readme/README.rst


ramnes:
If you wonder why I use HTML to center everything rather than rST directives, it is because of https://github.com/github/markup/issues/163.

flacjacket:
Do you know if the HTML directives will work on pypi? We use our readme
there as well.

ramnes:
Do we? https://pypi.org/project/qtile/ seems pretty blank compared to the README here.

ramnes:
Oh okay, it's from the latest `setup.cfg` changes. Just tried an upload on PyPI and it works with `master` but not with this branch, indeed…

ramnes:
@flacjacket so yeah, I put back everything as rST, but on the left, and gave up on the centering. Works great on PyPI.

flacjacket:
> Do we? https://pypi.org/project/qtile/ seems pretty blank compared to the README here.

I don't think we've had a release since the setup.cfg changes :)

PR #1313, opened 2019-04-25
Keep layout state to restore its layout if window is readded. Fix #1202

### LINKED ISSUE #1202: Unmaximize is not keeping order for monad layout
This issue is conceptually similar to #1201. Fullscreen toggle should restore exactly the same state. 

### Reproduce

1. set monad layout
2. maximize master window
3. unmaximize it

### Expected

order of windows should be the same

### Actual

the unmaximized window was put to second place

### Visual

Yellow is master window in initial state. After maximize/unmaximize the red one takes position of master

visual

neworld:
Fix of #1202

I assume other layouts might want to save their own state for the same problems. This solution works well for both tests and real testing. Still, I am not sure if it is a good decision or a hacky one. 

BTW. I am amazed at how well the code is written. For a while, I was a bit afraid dig here, but now I see how easy is to contribute.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -251,7 +251,7 @@ class _Window(command.CommandObject):
         return self == self.qtile.current_window
 
     def reset_layout_state(self):
-        self.layout_state = {}
+        self.layout_state = None
 
     def set_layout_state(self, state):
         self.layout_state = state
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -159,7 +159,7 @@ class _Window(command.CommandObject):
         self.hidden = True
         self.group = None
         self.icons = {}
-        self.layout_state = {}
+        self.layout_state = None
         window.set_attribute(eventmask=self._window_mask)
 
         self._float_info = {
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -159,6 +159,7 @@ class _Window(command.CommandObject):
         self.hidden = True
         self.group = None
         self.icons = {}
+        self.layout_state = {}
         window.set_attribute(eventmask=self._window_mask)
 
         self._float_info = {
@@ -248,6 +249,12 @@ class _Window(command.CommandObject):
     @property
     def has_focus(self):
         return self == self.qtile.current_window
+
+    def reset_layout_state(self):
+        self.layout_state = {}
+
+    def set_layout_state(self, state):
+        self.layout_state = state
 
     def update_name(self):
         try:
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -197,22 +197,24 @@ class MonadTall(_SimpleLayoutBase):
 
     def add(self, client):
         "Add client to layout"
-        if "index" in client.layout_state:
-            self.clients.add(client, client.layout_state["index"])
+
+        # Make sure layout wont crash if another layout puts its state
+        if isinstance(client.layout_state, int):
+            self.clients.add(client, client.layout_state)
         else:
             self.clients.add(client, 0 if self.new_at_current else 1)
 
-        "Cleanup layout state"
+        # Make sure this state will be never reused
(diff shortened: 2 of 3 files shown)


ramnes:
Hey @neworld, thanks for the PR and the kind words. :heart:

Check the Travis build, there are some PEP8 errors. Also, please do not put docstrings / anonymous strings in the middle of functions.

Now, I feel that putting the current layout state into the client objects might be a pretty good idea even if it's weird at first glance. I'm not entirely sure yet though; I need to check some things and hear what others think!

libqtile/window.py
```diff
@@ -159,6 +159,7 @@ def __init__(self, window, qtile):
         self.hidden = True
         self.group = None
         self.icons = {}
+        self.layout_state = {}
```
ramnes: Why not let each layout define what it needs, either a dictionary or something else?

```suggestion
        self.layout_state = None
```

Then you could just give the index directly for the xmonad layout, rather than a dictionary with just one key.

neworld: Store object instead of a dictionary is an even better solution.

flacjacket:
The problem I see with this is that only one layout will be able to use this parameter without stepping on another layout, and all of the layouts in a particular group have the windows added and removed as the floating changes. Furthermore, while this seems to work for full screening, it would probably have some strange interactions with floating/unfloating a window, as the offset that is applied may place the window into some strange location. I personally think it would be better if some solution could be worked out that could be used by all of the layouts for handling things that float and come back.

neworld:
> The problem I see with this is that only one layout will be able to use this parameter without stepping on another layout

This is the main reason I user dictionary before. It could be improved, even more, for example, `layout_state[layout.__class__.__name__] = SomeState`. I don't have many experience with python, so I don't know the best practices to deal with it.

Or maybe would be better to call `add(client, client.layout_state)` and manage layout states in caller site. And `remove()` could return the state of `None` if it doesn't allow state restoration. 

> Furthermore, while this seems to work for full screening, it would probably have some strange interactions with floating/unfloating a window, as the offset that is applied may place the window into some strange location.

The current solution puts the window into a strange location as well. This idea works pretty well with 2-3 windows in the layout and it enables me to temporally fullscreen focused window. 

> I personally think it would be better if some solution could be worked out that could be used by all of the layouts for handling things that float and come back.

I totally agree with you, but I hardly imagine a single solution for all layouts. I believe for many layouts the index is enough. But the `column` and some other layouts are tricky.

ramnes:
> This is the main reason I user dictionary before. It could be improved, even more, for example, `layout_state[layout.__class__.__name__] = SomeState`. I don't have many experience with python, so I don't know the best practices to deal with it.

I thought about this already and it sounds good to me, this the main thing I said I wanted to check.

tych0:
So this looks really interesting to me; I think using a dictionary with the layout as the key is a great idea.

I wonder if we can't also solve @flacjacket's complaints by adding a parameter to 'remove', so we could do something like:

```
l.remove(win, reason=window.FULLSCREEN)
l.remove(win, reason=window.FLOATING)
l.remove(win, reason=window.MINIMIZED)
l.remove(win, reason=UNMANAGED)
```

and then let the layouts decide how they want to handle things.

Unfortunately, this means that every layout will have to decide what to do on remove, but maybe we can add this logic to _ClientList instead to capture it for most of them? Thoughts?

This bug also annoys me, so I'd be happy to pick up the PR.

neworld:
Having `reason` sounds a good idea.

> Unfortunately, this means that every layout will have to decide what to do on remove, but maybe we can add this logic to _ClientList instead to capture it for most of them? Thoughts?

I suppose restoring logic should be in the layout itself because it allows custom layouts to use this infrastructure.

> This bug also annoys me, so I'd be happy to pick up the PR.

Would be great if you finish this

ramnes:
Somewhat related to the discussion in #1334

PR #1316, opened 2019-04-30
fix(doc): update gnome-settings-daemon to org.gnome.SettingsDaemon

nicolas-pantel:
Since Gnome 3.23.2, gnome-settings-daemon was split into helper daemons (source: https://gitlab.gnome.org/GNOME/gnome-settings-daemon/blob/master/NEWS).
The old config lead to an immediate logout after login in Qtile.


ramnes:
Thanks!

PR #1317, opened 2019-05-05
add pulseaudio C api based volume control widget

krya:
This is a simple C api based pulseaudio volume control widget.
The widget subscribes for changes of the default output device (sink in pulseaduio terminology) to get notification about volume change outside of the widget itself (e.g. via pavucontrol).
This widget does NOT limit the max volume as most of other DE seem to be doing ( I am not sure whether it should, hence no limitation is hardcoded).
Neither docs nor tests are provided via this PR since I'm not a native speaker and I have no idea how one would test a tiny wrapper over C lib.
Thought ?


ramnes:
Hey @krya, thank you for the great MR!

It is really clean and tidy, and I would probably be in favor of merging as is... but I'm always a bit reluctant with adding more C stuff in the code base. Why not just use `pactl subscribe` (leveraging asyncio) and `pactl`?

@flacjacket thoughts?

ramnes:
Also, I'm realizing that the `Volume` widget already has a lot of useful stuff we shouldn't duplicate. Could you extend `Volume` rather than adding a new widget?

krya:
Hi @ramnes 
> Why not just use pactl subscribe

while it might be ok to shell into pactl and wait one would have to make some regexes (and we all know where that path leads to )

> and pactl?

shelling to pactl on each mouse scroll is cpu intensive and horribly slow not because you need to actually run cmd in shell but also wait for changed to get reflected in that pactl subscribe async task

> Could you extend Volume rather than adding a new widget?

yeah, that could be done and as far as I'm aware the only usefull thing this widget could inherit from a regular volume is the code that does the fancy drawing with e.g. icons but then there some hops have to be made in order to avoid the parent widget to schedule updates.

libqtile/widget/pulse_volume.py
```diff
@@ -0,0 +1,253 @@
...
+
+        self.tick()
+
+    def poll(self):
+        lib.pa_mainloop_iterate(self.loop, 0, ffi.NULL)
+
+    def tick(self):
+        self.poll()
+
+    def update(self):
```
tych0: Seems like this is overriding the InLoopPollText update() method?

tych0:
I don't mind the C necessarily, but it looks like the build is complaining about a few legitimate issues?

libqtile/widget/pulse_volume.py
krya: it is overriding original's widget update as it states in the doc since that widget assumes that the value is stored in self.volume so it seems like a best place to update it

krya:
@tych0 well, only mypy build is failing and I have no idea why. is it because its a C wrapper and travis doesnt build them during that test run ?

libqtile/widget/pulse_volume.py
tych0: I think you just don't want to inherit from InLoopPollText() at all, and manage the event loop on your own (or really, subscribe to the main eventloop somehow).

tych0:
I think you probably need something like:

```
diff --git a/setup.py b/setup.py
index 5bf2f0ea..f2a4a9d5 100755
--- a/setup.py
+++ b/setup.py
@@ -69,5 +69,6 @@ setup(
     cffi_modules=[
         'libqtile/pango_ffi_build.py:pango_ffi',
         'libqtile/core/xcursors_ffi_build.py:xcursors_ffi',
+        'libqtile/widget/pulseaudio_ffi.py:ffibuilder',
     ],
 )
```

setup.cfg
```diff
@@ -120,5 +120,7 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-libqtile.core._ffi_xcursors]
 ignore_missing_imports = True
+[mypy-libqtile.widget._pulse_audio]
+ignore_missing_imports = True
```
tych0: I don't think this is quite right; it really is complaining about a bug, you're changing the signature of InLoopPollText; seems like you can just drop that from the inheritance chain all together.

krya: well, at least now mypy doesnt complain about import error of that module

tych0: Yes, but I don't think we want to merge it like this.

krya: aaaand its failing now

libqtile/widget/pulse_volume.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def get_sinks(self):
+        op = lib.pa_context_get_sink_info_list(
+            self.context,
+            lib.qtile_on_sink_info,
+            self.handle
+        )
+        self.wait_for_operation(op)
+        if not self._subscribed:
+            self.subscribe_to_sink_events()
```
flacjacket: Maybe just move this to `.connect()`?

libqtile/widget/pulse_volume.py
```diff
@@ -0,0 +1,253 @@
...
+        # to wait a callback from pulseaudio
+        self.default_sink['values'] = list(volume.values)
+        op = lib.pa_context_set_sink_volume_by_index(
+            self.context,
+            self.default_sink['index'],
+            volume,
+            ffi.NULL,
+            ffi.NULL
+        )
+        if op:
```
flacjacket: Why do you need the check here and not for the mute?

libqtile/widget/pulse_volume.py
```diff
@@ -0,0 +1,253 @@
...
+            'channels': self.default_sink['channels'],
+            'values': self.default_sink['values'],
+        })
+        lib.pa_cvolume_inc(
+            volume,
+            int(value * base / 100),
+        )
+        # check that we dont go over 100% in case its set in config
+        if self.limit_max_volume:
+            volume.values = [(i if i <= base else base) for i in volume.values]
```
flacjacket: Maybe `max(i, base)`?

setup.py
```diff
@@ -69,5 +69,6 @@ def finalize_options(self):
     cffi_modules=[
         'libqtile/pango_ffi_build.py:pango_ffi',
         'libqtile/core/xcursors_ffi_build.py:xcursors_ffi',
+        'libqtile/widget/pulseaudio_ffi.py:pulseaudio_ffi',
```
flacjacket: I worry here about adding a pulse dependency which is only used for this widget. Is there any way to make this optional?

libqtile/widget/pulse_volume.py
```diff
@@ -0,0 +1,253 @@
...
+        self._subscribed = False
+        self.default_sink_name = None
+        self.default_sink = None
+        self.handle = ffi.new_handle(self)
+
+        # create a loop and api entry point
+        self.loop = lib.pa_mainloop_new()
+        self.api = lib.pa_mainloop_get_api(self.loop)
+        # create context (e.g. connection)
+        self.client_name = ffi.new('char[]', b'Qtile-pulse')
```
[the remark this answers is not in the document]
krya: pulseaudio needs it as a reference to be alive

libqtile/widget/pulse_volume.py
krya: I think you meant `min(i, base)` and thats the main reason its written verbosely

setup.py
krya: I am actually not sure whether there is such a way except maybe compile it during regular qtile run. at least travis needs it to be included so that mypy wont complain. 
and then for a regular user that would install qtile - as far as I am aware qtile only ships libqtile to destination meaning there is no script that could be run in order to compile.

tych0: What about just appending it if pkgconfig says it's present?

krya:
@flacjacket @tych0 what else needs to be changed in order this PR to be merged ?

tych0:
On Sat, Sep 07, 2019 at 11:46:37AM -0700, Denis K. wrote:
> @flacjacket @tych0 what else needs to be changed in order this PR to be merged ?

Seems like it at least needs a rebase, looks like the build also
didn't pass?

Anyway, thanks for working on it!

Tycho

krya:
@tych0 github says its mergeable and the build failed in some unrelateted test which looks like glitch in CI issue itself

tych0:
@krya any chance you can rebase this?

krya:
@tych0 done.

tych0:
Thanks! I went ahead and squashed this all down and merged it. Thanks for your patience.

PR #1333, opened 2019-06-01
Always focus the latest focused window when removing a window

ramnes:
This is an alternative PR to https://github.com/qtile/qtile/pull/1052, that implements my proposition of always focusing the latest focused window at the group level when a window is removed, and not letting the current layout handling that.

It still needs some work to understand why it breaks the BSP layout tests.


ramnes:
Rebased on top of #2058.

Issue #297, opened 2013-04-04
Drag windows within positions of layout using mouse


william-gr:
Hi,

I used awesome wm before, and there was this feature that I could switch window position by holding Mod + MouseButton1 click and move.

For instance, if I have a RadiotTile layout with 6 windows, I can Mod + click1 the 1st window and move it to the 6th position.

Right now there is a command set_position_floating that will set the window as floating and remove it from the Tile, but the desired feature is to just move the window position instead of make it float.

Thank you.

Johnystar:
> Afaik, @blotus already did this. Didn't remember if it was merge or not, though.
> 
> Edit: see https://github.com/blotus/qtile/commits/move_tile.

``move_tile`` doesn't appear to be a thing anymore. Is there any way to still do this today? I'm looking to switch from AwesomeWM, but this could be a deal-breaker for me.

Issue #855 also seems to be related to this issue, but the #1334 pull request mentioned there has since turned stale and has been automatically closed.

jwijenbergh:
I feel like we should close this. It has been recently fixed with https://github.com/qtile/qtile/pull/4166

To support this for arbitrary layouts, implement a "swap" method

PR #1334, opened 2019-06-01
Improve cmd_set_position by asking layouts to implement a swap method

ramnes:
I wrote this code after reading https://github.com/qtile/qtile/issues/297 and https://github.com/qtile/qtile/pull/410 again, and digging a bit on what a user can do with this as of today. Sadly, it looks like it doesn't work with many (any?) layouts right now. As a reminder, it is about changing the placement of a window without making it float (just by swapping in place with other windows)

As all layouts don't handle their windows the same way (some have a `_ClientList`, others have more than one, and others have a `list`), I propose to ask every layout to implement a new `swap` method, that will be used so that `cmd_set_position` really becomes layout-agnostic. The only problem with this method is that it doesn't make sense to swap with the mouse on some layouts (Max or TreeTab for example), and we should find a way to avoid it being tried so that weird things don't happen.

Another way of doing this, which would be a big architecture change and require way more work, would be to keep the list of windows only in the group, not in layouts, so that we could just invert the two windows in that list of group clients. Layouts would just expose "slots" that the group would use to throw windows in. This solution could probably help for our serialization problems, by isolating responsibilities.

This PR also needs tests and documentation if anything.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -205,6 +205,24 @@ class Columns(Layout):
             self.remove_column(c)
         return self.columns[self.current].cw
 
+    def swap(self, client1, client2):
+        column1 = None
+        column2 = None
+        for c in self.columns:
+            if client1 in c:
+                column1 = c
+            if client2 in c:
+                column2 = c
+        if not (column1 and column2):
+            return
+        if column1 is column2:
+            column1.swap(client1, client2)
+        else:
+            column1.add(client2)
+            column2.add(client1)
+            column1.remove(client1)
+            column2.remove(client2)
+
     def configure(self, client, screen):
         pos = 0
         for col in self.columns:
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -261,6 +261,10 @@ class Floating(Layout):
         d["clients"] = [c.name for c in self.clients]
         return d
 
+    def swap(self, client1, client2):
+        # This can't ever be called, but implement the abstract method
+        pass
+
     def cmd_next(self):
         # This can't ever be called, but implement the abstract method
         pass
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -136,6 +136,14 @@ class Layout(CommandObject, configurable.Configurable, metaclass=ABCMeta):
         pass
 
     @abstractmethod
+    def swap(self, client1, client2):
+        """Called to swap a window with another one
+
+        The layout should exchange the position of the two windows.
+        """
+        pass
+
+    @abstractmethod
     def configure(self, client, screen):
         """Configure the layout
 
@@ -629,6 +637,9 @@ class _SimpleLayoutBase(Layout):
 
     def remove(self, client):
         return self.clients.remove(client)
+
+    def swap(self, client1, client2):
+        return self.clients.swap(client1, client2)
 
     def info(self):
         d = Layout.info(self)
--- a/libqtile/layout/stack.py
+++ b/libqtile/layout/stack.py
@@ -205,6 +205,24 @@ class Stack(Layout):
             if n:
                 return n.cw
 
+    def swap(self, client1, client2):
+        stack1 = None
+        stack2 = None
+        for s in self.stacks:
+            if client1 in s:
+                stack1 = s
+            if client2 in s:
+                stack2 = s
+        if not (stack1 and stack2):
+            return
+        if stack1 is stack2:
+            stack1.swap(client1, client2)
+        else:
+            stack1.add(client2)
+            stack2.add(client1)
+            stack1.remove(client1)
+            stack2.remove(client2)
+
     def configure(self, client, screen):
         for i, s in enumerate(self.stacks):
             if client in s:
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -415,6 +415,9 @@ class TreeTab(Layout):
         del self._nodes[win]
         self.draw_panel()
 
+    def swap(self, client1, client2):
+        pass
+
     def _create_panel(self):
         self._panel = window.Internal.create(
             self.group.qtile,
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1425,10 +1425,6 @@ class Window(_Window):
                 continue
             curx, cury = self.qtile.get_mouse_position()
             if self._is_in_window(curx, cury, window):
-                clients = self.group.layout.clients
-                index1 = clients.index(self)
-                index2 = clients.index(window)
-                clients[index1], clients[index2] = clients[index2], clients[index1]
-                self.group.layout.focused = index2
-                self.group.layout_all()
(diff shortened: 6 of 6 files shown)


libqtile/layout/floating.py
```diff
@@ -261,6 +261,10 @@ def info(self):
         d["clients"] = [c.name for c in self.clients]
         return d
 
+    def swap(self, client1, client2):
+        # This can't ever be called, but implement the abstract method
```
ramnes: This is not true actually, we could implement this but... I guess it's not very useful, unless we find a way to swap across layouts.

tych0:
One thing that would be useful for implementing other features is if we did force every layout to keep track of their windows in some standard way. Maybe we can make our own WindowList() class or something? I realize that's a lot more work than maybe you want to do, though :)

ramnes:
Yes, that could work also! But I feel that it was already tried with `_ClientList`, and didn't work that well (some layouts don't use it, or use many). How do you enforce the usage of one and only one instance of that structure? Maybe by stopping layouts to implement methods such as `add`, `remove`, or even `focus_*`, and rather having them registering hooks to such events.

As for what work I want to do, well, if I could I'd rewrite the whole groups/layouts thing, eheh. It's much more a matter of free time, or lack of.

tych0:
Right, I think unless someone bothers to port every layout to the new system, not every layout will use it :).

It might be interesting to make the layouts themselves stateless, and have a ClientState() or something, where you can do state.current_window(), state.set_current_window(), etc. and also have it behave as a list if people want to iterate over it?

heitzmann:
This would be similar to what dwm does: it keeps a single client stack and all layouts use that same data structure, if I'm not mistaken. Although simpler to maintain, I don't think it's possible for dynamic layouts, such as Stack, Columns and Bsp to function as they do today... they would still have to keep internal data about the clients and keep that in sync with the "global stack"... sounds more complicated in the end.

ramnes:
Yes, hence my first proposition that they simply don't care about the windows and just expose slots that are filled by the group. I'm still not sure how that would look like in terms of code, though.

tych0:
I guess I'm sort of suggesting some kind of hybrid approach: every layout would keep the list of windows in this known attribute, so that the qtile core could look into the window list in a reasonable way.

The one remaining feature in qtile I've wanted forever would require some way to ask the layouts "is this the last window in your list"; having everyone use the same list implementation would make this easy. But if we don't have that, then each layout has to implement this hook (and every other hook about any question that the core might want to ask of the layouts).

heitzmann:
I see all three options being possible: the slots idea, the standard attribute that the core can look into, and the implementation of standard hooks. In all cases the layout would keep an internal representation of its state.

It seems to me that using a standard attribute across layouts is more error prone than have them implement the required hooks. This looks more elegant to me and would probably require fewer changes.

A swap hook, for example, would not be hard to implement in most layouts. Others, like max, could fall through a standard base class implementation that essentially does nothing (in fact, this could be the default behavior until the layout implements the hook).

However, in my opinion, the decision should be based on what you be more efficient, since no one really likes a sluggish WM...

tych0:
On Mon, Jun 03, 2019 at 12:07:30PM -0700, Lucas H. Gabrielli wrote:
> I see all three options being possible: the slots idea, the standard attribute that the core can look into, and the implementation of standard hooks. In all cases the layout would keep an internal representation of its state.
> 
> It seems to me that using a standard attribute across layouts is more error prone than have them implement the required hooks. This looks more elegant to me and would probably require fewer changes.

It would require fewer changes initially, but result in more code
ultimately, I think. For example, a canonical "window list" option
would mean that the code to answer the question "is this the last
window in the layout?" would only have to exist once, instead of
having essentially the same code everywhere to interpret the various
implementations of client lists.

The next time someone wants to ask a question like the above, they'd
have to go implement it again for every layout, instead of just using
the existing list. (A corresponding question I'd need to answer is "is
this the first window in your layout.) IMO, answering these sorts of
general questions about layout properties would look better if they
were all computed over the same data structure, vs. custom
implementations.

That said, if someone came along with an implementation using hooks
that made it easy for me to answer the above two questions (without
having to implement it for every layout), I'd be happy to use it :).

tych0:
On Mon, Jun 03, 2019 at 09:04:50PM -0600, Tycho Andersen wrote:
> On Mon, Jun 03, 2019 at 12:07:30PM -0700, Lucas H. Gabrielli wrote:
> > I see all three options being possible: the slots idea, the standard attribute that the core can look into, and the implementation of standard hooks. In all cases the layout would keep an internal representation of its state.
> > 
> > It seems to me that using a standard attribute across layouts is more error prone than have them implement the required hooks. This looks more elegant to me and would probably require fewer changes.
> 
> It would require fewer changes initially, but result in more code
> ultimately, I think. For example, a canonical "window list" option
> would mean that the code to answer the question "is this the last
> window in the layout?" would only have to exist once, instead of
> having essentially the same code everywhere to interpret the various
> implementations of client lists.
> 
> The next time someone wants to ask a question like the above, they'd
> have to go implement it again for every layout, instead of just using
> the existing list. (A corresponding question I'd need to answer is "is
> this the first window in your layout.) IMO, answering these sorts of
> general questions about layout properties would look better if they
> were all computed over the same data structure, vs. custom
> implementations.
> 
> That said, if someone came along with an implementation using hooks
> that made it easy for me to answer the above two questions (without
> having to implement it for every layout), I'd be happy to use it :).

Although, upon thinking about it some more, I'm not sure it's possible
to answer the questions I have with a generic window list. For
example, in stack, what I'm really after is "would a move left move
off the screen?", which is not the same question as "is this the last
window in the layout". So perhaps you should ignore me :)

ramnes:
Okay guys, what do you think of the diff as is? Should we follow that route and discuss that big refactoring again later?

I just thought that also, if we merge this, we should be able to refactor a good bunch of the `cmd_shuffle_*` methods.

libqtile/layout/columns.py
```diff
@@ -205,6 +205,24 @@ def remove(self, client):
...
+                column2 = c
+        if not (column1 and column2):
+            return
+        if column1 is column2:
+            column1.swap(client1, client2)
+        else:
+            column1.add(client2)
+            column2.add(client1)
+            column1.remove(client1)
+            column2.remove(client2)
```
heitzmann: In this case, shouldn't each client be added into the position from where the other was removed? It seems to me that the order will be messed up. It would also reset the relative sizes between clients in each column. The same problem would happen with `Stack`. I believe this should be replaced by (I have not tested it , though):
```python
i1 = column1.index(client1) - column1.current_index
i2 = column2.index(client2) - column2.current_index
h1 = column1.heights[client1]
h2 = column2.heights[client2]
temp = column1.insert_position
column1.insert_position = i1
column1.add(client2, h2)
column1.insert_position = temp
temp = column2.insert_position
column2.insert_position = i2
column2.add(client1, h1)
column2.insert_position = temp
column1.remove(client1)
column2.remove(client2)
```

Issue #1325, opened 2019-05-23
closing windows seems to be broken with xcffib.xproto.WindowError


jagguli:
Closing an window also throws errors 

my env is broken right now, I was using the git version on arch, switched to the release version issue still exists. So I'm guessing its something related to xcffib? 

I'm on archlinux tried upgrading also.

```
File "/home/steven/.config/qtile/extra.py", line 179, in __call__
    qtile.current_screen.cmd_toggle_group(self.name)
  File "/usr/lib/python3.7/site-packages/libqtile/config.py", line 406, in cmd_toggle_group
    self.set_group(group)
  File "/usr/lib/python3.7/site-packages/libqtile/config.py", line 324, in set_group
    old_group._set_screen(None)
  File "/usr/lib/python3.7/site-packages/libqtile/group.py", line 183, in _set_screen
    self.hide()
  File "/usr/lib/python3.7/site-packages/libqtile/group.py", line 191, in hide
    i.hide()
  File "/usr/lib/python3.7/site-packages/libqtile/window.py", line 422, in hide
    self.window.unmap()
  File "/usr/lib/python3.7/site-packages/libqtile/core/xcbq.py", line 744, in unmap
    self.conn.conn.core.UnmapWindowChecked(self.wid).check()
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 338, in check
    self.conn.request_check(self.sequence)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 570, in wrapper
    return f(*args)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 679, in request_check
    self._process_error(err)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 647, in _process_error
    raise error(buf)
xcffib.xproto.WindowError
```

jagguli:
```
2019-06-01 02:22:08,031 libqtile group.py:layout_all():L163  Exception in layout max
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/group.py", line 160, in layout_all
    self.layout.layout(normal, screen)
  File "/usr/lib/python3.7/site-packages/libqtile/layout/base.py", line 57, in layout
    self.configure(i, screen)
  File "/usr/lib/python3.7/site-packages/libqtile/layout/max.py", line 58, in configure
    client.hide()
  File "/usr/lib/python3.7/site-packages/libqtile/window.py", line 422, in hide
[...]

Issue #1150, opened 2018-03-03
Exception in Max layout


nazar-pc:
Found this in my log, not really sure why this happened, but it did:

```
2018-03-02 14:56:35,928 ERROR libqtile group.py:layoutAll():L163 Exception in layout Max
Traceback (most recent call last):
  File "/web/github/qtile/libqtile/group.py", line 160, in layoutAll
    self.layout.layout(normal, screen)
  File "/web/github/qtile/libqtile/layout/base.py", line 57, in layout
    self.configure(i, screen)
  File "/web/github/qtile/libqtile/layout/max.py", line 57, in configure
    client.hide()
  File "/web/github/qtile/libqtile/window.py", line 417, in hide
    self.window.unmap()
  File "/web/github/qtile/libqtile/xcbq.py", line 741, in unmap
    self.conn.conn.core.UnmapWindowChecked(self.wid).check()
  File "/home/nazar-pc/.local/lib/python3.6/site-packages/xcffib/__init__.py", line 328, in check
    self.conn.request_check(self.sequence)
  File "/home/nazar-pc/.local/lib/python3.6/site-packages/xcffib/__init__.py", line 558, in wrapper
    return f(*args)
  File "/home/nazar-pc/.local/lib/python3.6/site-packages/xcffib/__init__.py", line 667, in request_check
    self._process_error(err)
  File "/home/nazar-pc/.local/lib/python3.6/site-packages/xcffib/__init__.py", line 635, in _process_error
    raise error(buf)
xcffib.xproto.WindowError
```

jagguli:
@nazar-pc did you manage to fix or workaround this?

nazar-pc:
No, I didn't

tych0:
I think this basically means that a window was created, and then destroyed all before the configure() method completed; in this case (for the max layout), it looks basically harmless. We could wrap the hide/unhide() in max's configure() with a try/except, or perhaps we should switch unmap() to use the unchecked flavor to avoid the exception all together.

tych0:
Gonna close this per the above fix.

PR #1335, opened 2019-06-02
don't check result of unmap

tych0:
We have reports of stack traces like:

```
File "/home/steven/.config/qtile/extra.py", line 179, in __call__
    qtile.current_screen.cmd_toggle_group(self.name)
  File "/usr/lib/python3.7/site-packages/libqtile/config.py", line 406, in cmd_toggle_group
    self.set_group(group)
  File "/usr/lib/python3.7/site-packages/libqtile/config.py", line 324, in set_group
    old_group._set_screen(None)
  File "/usr/lib/python3.7/site-packages/libqtile/group.py", line 183, in _set_screen
    self.hide()
  File "/usr/lib/python3.7/site-packages/libqtile/group.py", line 191, in hide
    i.hide()
  File "/usr/lib/python3.7/site-packages/libqtile/window.py", line 422, in hide
    self.window.unmap()
  File "/usr/lib/python3.7/site-packages/libqtile/core/xcbq.py", line 744, in unmap
    self.conn.conn.core.UnmapWindowChecked(self.wid).check()
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 338, in check
    self.conn.request_check(self.sequence)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 570, in wrapper
    return f(*args)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 679, in request_check
    self._process_error(err)
  File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 647, in _process_error
    raise error(buf)
xcffib.xproto.WindowError
```

with the latest xcffib. The WindowError here means we passed a bad window id,
which means the window died sometime before configure() can happen. xcffib
0.8.0 included what seems to be a fairly significant optimization, which may
result in these sorts of races popping up more and more.

Instead of getting this error, let's simply ignore (i.e. call the Unchecked()
version of the function) any errors from Unmap(). Note that this is probably
not ok in the general case, but Unmap() takes only one argument, the wid, so
the only error it will ever generate is a WindowError. If we encounter this in
the future with other functions, we may want to keep calling the Checked()
version to validate other inputs, and just catch and ignore WindowError.

See #1325 and #1150

### DIFF
--- a/libqtile/core/xcbq.py
+++ b/libqtile/core/xcbq.py
@@ -741,7 +741,7 @@ class Window:
         self.conn.conn.core.MapWindow(self.wid)
 
     def unmap(self):
-        self.conn.conn.core.UnmapWindowChecked(self.wid).check()
+        self.conn.conn.core.UnmapWindowUnchecked(self.wid)
 
     def get_attributes(self):
         return self.conn.conn.core.GetWindowAttributes(self.wid).reply()


flacjacket:
Nice, thanks

PR #1336, opened 2019-06-03
Drop xvfbwrapper

tych0:
Here's a better version of #1281 where we just drop xvfbwrapper as a dependency all together.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -39,8 +39,8 @@ import time
 import traceback
 
 import xcffib
+import xcffib.testing
 import xcffib.xproto
-from xvfbwrapper import Xvfb
 
 # the default sizes for the Xephyr windows
 WIDTH = 800
@@ -102,12 +102,6 @@ def can_connect_qtile(socket_path):
     if val == 'OK':
         return True
     return False
-
-
-def _find_display():
-    """Returns the next available display"""
-    xvfb = Xvfb()
-    return xvfb._get_next_unused_display()
 
 
 def whereis(program):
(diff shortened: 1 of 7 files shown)


test/conftest.py
```diff
@@ -237,6 +231,12 @@ def stop_xephyr(self):
 
         self.proc = None
 
+        # clean up the lock file for the display we allocated
```
ramnes: why?

ramnes:
Awesome, thanks!

test/conftest.py
ramnes: Haha! More like why do we have to do this while we didn't with xvfbwrapper?

tych0: We didn't have to because we were just leaking things everywhere. If you look at my original patch in the other PR, I added cleanup code in the same place.

xcffib.testing.XvfbTest() will handle this for you, but since we're going out of band to use xcffib.testing.find_display() (for use with xephyr, not xvfb), we have to do manual cleanup. The xcffib test suite uses xvfb exclusively, so it doesn't offer a wrapper for xephyr.

It's a little bit unnecessary to use both xephyr and xvfb, we could probably just switch everything to just xephyr. But xvfb is probably faster for the test cases where it works. Xephyr can handle mouse/keyboard inputs, where xvfb really just handles drawing. So for qtile, we really do need xephyr to fake input, whereas the xcffib tests mostly just try to manipulate windows to make sure the protocol basics work correctly.

ramnes: Thanks!

PR #1340, opened 2019-06-09
add outer_margin option to columns layout

m-col:
This small change adds an 'outer_margin' option to the columns layout. It was inspired by the same setting in bspwm. It works well when I've played around with it but my understanding of qtile isn't complete so hopefully the implementation makes sense. In any case, I'd love to have this setting added so please do let me know if any changes should be made to it and I can try to fix it up.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -114,6 +114,7 @@ class Columns(Layout):
          "Border colour for un-focused windows in stacked columns."),
         ("border_width", 2, "Border width."),
         ("margin", 0, "Margin of the layout."),
+        ("outer_margin", 0, "Outer margin of the layout."),
         ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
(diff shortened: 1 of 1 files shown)


m-col:
To clarify what this does, here is a screenshot. The config here has margin set to 8 and outer_margin set to 50. It pads the perimeter around the screen, rather than the windows, as 'margin' does. They can be used together as shown here because its dope.

ramnes:
Hey @m-col, thanks for this. It's always nice to see some Qtile rice! :sunglasses: 

I really would like such a feature, but I'd prefer it to be available no matter the layout. In that sense, I feel that your PR here is redundant with the work I've started on https://github.com/qtile/qtile/pull/1239, so that we can have multiple empty "gaps" per screen.

Let's keep this open to 1/ see what others think and 2/ see if more people wants it in the tree right now, but otherwise I think the proper way would be to finish #1239 and then write something in `layout.base.Layout` and/or `config.Screen` so that it's easy to add those margins at once for any layout and/or at the screen level directly.

How does that sound to you?

m-col:
A generalised implementation as per #1239 makes way more sense! I initially tried to add outer_margin to to _Window.place() and generalising it for all layouts but I quickly realised how complicated that would become.

Does "Multiple gaps per position" refer to setting different gap sizes to each of the 4 screen edges? That would be awesome if so.

I agree with your suggestion and look forward to the outcome of #1239 !

ramnes:
Yep, that's it exactly. I'm glad that you're interested in this also! Let's keep the PR open for a few days to see if anyone kicks in or close it otherwise. Thanks!

m-col:
It's been a little while so I'll closed this, and look forward to the future changes following #1239.

Thanks again for qtile!

PR #1347, opened 2019-06-18
Add pytype check to travis

flacjacket:
This adds a typing check using the pytype tool.

We still need to fix up a few things before this can be merged.


ramnes:
I didn't know pytype, that's really nice! It seems like there isn't much to fix to have it pass, do you plan to do it?

tych0:
Arg, I has a sad. I was going to play with this again tonight, but it turns out: https://github.com/google/pytype/issues/440

PR #1348, opened 2019-06-18
Move DGroups to group manager in core

flacjacket:
Building off #1299, moves the current dgroup functionality into a class called group manager, move some of the functionality to manage groups off of the main qtile manager class and into the group manager. Maintains backwards compatibility with pass through methods.


ramnes:
That looks interesting, but what's the end goal exactly?

Things I like at first glance:
- `dgroups` and `groups` handled under the same namespace, opens the way for a deeper refactoring to remove the distinction between both
- makes `manager.py` simpler

Things I'm not sure about:
- `qtile._group_manager.groups` and that kind of stuff, which I feel is not intuitive and could end up as an unnecessary architectural complexity (one could ask if we should also have a `WidgetManager` or a `KeyManager`)
- keeping backward compatibility on an internal API vs achieving end goal (and properly documenting breaking-changes)
- the fact that `GroupManager` still depends on a `Qtile` instance being passed in

Overall I'm probably +1 on this, but I'm not sure why we are going down this road. :)

flacjacket:
Yeah, my initial thoughts were to consolidate some of the group handling logic into one place and make it more consistent, I think the current DGroups had been someones custom extension that was merged in, and you can still tell from the split between that and the main manager. I have made the `cmd_`s maintain backwards compatibility, but definitely broke things that directly accessed the manager's internals. Agree that this would need to be a documented change. My initial hope was to not have to keep the `qtile` instance in the group manager, and it could a separate (and separately tested) entity, but there was a lot that the group handling currently does that is dependent on other properties the qtile object had access to. It'd be great if we could refactor more of that out, and suggestions there welcome.

flacjacket:
The dgroup vs group distinction is actually something that I think came about from the fact that there was this extension added in. I think the actual distinction now is between the configuration of the groups, which includes any dynamic properties of the groups and how they spawn, and the currently active groups, which is mostly the name, label, and a container of layouts. It'd be great if we could better clarify the distinction between configuration and the active elements (here and throughout). I think that would help us clean up our act with how we can restart and reload configurations.

PR #1349, opened 2019-06-19
More consistent resize behavior in Columns layout

heitzmann:
In Columns layout, it is possible to shrink the left-most column with
grow_left and the right-most one with grow_right.  This commit extends
the same bahavior for the top and bottom-most clients in a given column
with grow_up and grow_down, respectively.

It also includes a small bugfix for the case when there is only a single
column: grow_left would still try to resize column[1].


ramnes:
Nice, thanks!

PR #1359, opened 2019-06-27
Update volume widget

krive001:
Modify PEP8 code and fix command call.

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -60,17 +60,15 @@ class Volume(base._TextBox):
         ("padding", 3, "Padding left and right. Calculated if None."),
         ("theme_path", None, "Path of the icons"),
         ("update_interval", 0.2, "Update time in seconds."),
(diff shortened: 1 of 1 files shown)


ramnes:
Hello @krive001, thanks for your pull request.

The first lines of your diff actually go against the PEP8, which says that:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

For the rest of your changes, it's just a matter of preference, so I'm not too keen on merging those unless we add a strict style linter like yapf or black and just fix it everywhere, not only in one widget. I hope this make sense.

That being said, it would be nice to merge the `shell=True` arguments. If you could modify your PR to just keep those modifications, that would be great.

ramnes:
Thank you very much for your contribution and understanding! :100:

Issue #1300, opened 2019-03-22
Spacer widget cannot take a background color


Cur10:
In the documentation it says the spacer widget can use a background however when it is added in the bar, it errors out.

ramnes:
It has been fixed by @krive001 pull request. Thank you for reporting this @Cur10!

SpyrosRoum:
Hello, I still get `TypeError: __init__() got an unexpected keyword argument 'background'` if I try to put background in a spacer widget

ramnes:
Hey @SpyrosRoum, thanks for raising this. Did you install Qtile from sources? This patch hasn't been released yet.

SpyrosRoum:
Oh sorry, I installed from pacman and I thought it was out

PR #1364, opened 2019-07-01
Spacer background error

krive001:
Bug #1300

### DIFF
--- a/libqtile/widget/spacer.py
+++ b/libqtile/widget/spacer.py
@@ -57,7 +57,7 @@ class Spacer(base._Widget):
         base._Widget.__init__(self, length)
 
     def draw(self):
-        self.drawer.clear(self.bar.background)
+        self.drawer.clear(self.background)
         if self.bar.horizontal:
             self.drawer.draw(offsetx=self.offset, width=self.length)
         else:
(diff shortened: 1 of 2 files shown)


libqtile/widget/spacer.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, length=bar.STRETCH, width=None):
         base._Widget.__init__(self, length)
 
     def draw(self):
-        self.drawer.clear(self.bar.background)
+        self.drawer.clear(self.background)
```
ramnes: Shouldn't we fallback on `self.bar.background` if `self.background` isn't set?

```suggestion
        self.drawer.clear(self.background or self.bar.background)
```

libqtile/widget/volume.py
```diff
@@ -100,31 +100,31 @@ def create_amixer_command(self, *args):
     def button_press(self, x, y, button):
         if button == BUTTON_DOWN:
             if self.volume_down_command is not None:
-                subprocess.call(self.volume_down_command)
+                subprocess.call(self.volume_down_command, shell=True)
```
ramnes: Please rebase on the master branch so that those changes don't appear here. :wink:

PR #1366, opened 2019-07-02
Small bug fix

heitzmann:
When `qsh` was changed to `qshell`, the number of stripped characters was not updated.


ramnes:
good catch, thanks!

PR #1367, opened 2019-07-04
Implement an fmt option for all textboxes

ramnes:
Right now, if you want to add some stuff around your widgets, you probably do something like this:

```python
widget.TextBox(text="!!! "),
widget.Notify(),
widget.TextBox(text="clock: {}"),
widget.Clock()
```

This PR allows to do stuff like that:

```python
widget.Notify(fmt="!!! {}"),
widget.Clock(fmt="clock: {}")
```

It's better because it re-uses the settings of the widget (font size, color, etc), but also respects the way the widget uses the text (for example, it won't leave `!!!` around when there is no new notification with `Notify`). And it's shorter to write!

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -293,6 +293,7 @@ class _TextBox(_Widget):
             "font shadow color, default is None(no shadow)"
         ),
         ("markup", False, "Whether or not to use pango markup"),
+        ("fmt", "{}", "How to format the text")
     ]  # type: List[Tuple[str, Any, str]]
 
     def __init__(self, text=" ", width=bar.CALCULATED, **config):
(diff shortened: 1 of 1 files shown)


libqtile/widget/base.py
```diff
@@ -301,17 +302,6 @@ def __init__(self, text=" ", width=bar.CALCULATED, **config):
...
-    def text(self):
-        return self._text
-
-    @text.setter
-    def text(self, value):
-        assert value is None or isinstance(value, str)
-        self._text = value
-        if self.layout:
-            self.layout.text = value
-
```
ramnes: I decided to remove this whole part because it felt useless to me to 1/ re-assign `self.layout.text` every time you manipulate `self.text`, even if you only use it when it's drawn and 2/ check the type of what's given in with an `assert` (let the layout fail itself if it can't handle this)

flacjacket:
Nice!

libqtile/widget/base.py
ramnes: This decision ended up being a bit buggy, see https://github.com/qtile/qtile/commit/77b47b07b08bdd38e15468714f3ba2d5de56a78e

PR #1381, opened 2019-07-22
Accept Pango markup by default on all text widgets

ramnes:
I think it's great to be more "powerful" by default. Do you see any reason to limit Pango markup usage?


flacjacket:
I'd be amazed if somebody accidentally used pango markup, looks like a good change.

PR #1382, opened 2019-07-24
columns: add separate wrapping config for stacked mode

hallyn:
When windows are stacked, I want cmd_up/down to wrap.  When they
are not stacked, I want them to stop at top/bottom.  Provide a
separate config.  The default is True so that the defaults are
unchanged.

Signed-off-by: Serge Hallyn <shallyn@cisco.com>


flacjacket:
Looks good, thanks!

PR #1387, opened 2019-08-02
Multiple gaps per position

ramnes:
Continuation of #1239, targeting `master` this time (still needs a rebase though).

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -63,27 +63,27 @@ class Gap(CommandObject):
         # given to the horizontal ones
         if self.position == "top":
             self.x = screen.x
-            self.y = screen.y
+            self.y = screen.y + self.offset
             self.length = screen.width
             self.width = self.length
             self.height = self.size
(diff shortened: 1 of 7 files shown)


ramnes:
Rebased against master.

ramnes:
The last commit fixes the issue I've mentioned in https://github.com/qtile/qtile/pull/1239#issuecomment-448252066.

ramnes:
I'm still trying to get my head around the IPC stuff so that I can replace the second commit (a76089b8a49c8b55a8a89e2fdd60e27f058727a8) with something that would make `bar["bottom"][0]` work, rather than `bar["bottom_0"]`. Any hint for a solution that doesn't imply rewriting all that part of the codebase would be really appreciated!

ramnes:
I'm going to close this for now as I don't see it going anywhere, and since it doesn't add a lot of value anymore with #1438 and #1560 merged in. Of course, free to start from this diff if for some reason you want to push it further.

PR #1389, opened 2019-08-10
Add a quick exit widget and change the default configuration.

sux2mfgj:
Some times I failed to change my configuration file of Qtile. At the time, the default configuration was loaded. but, From the qtile with the default configuration, I couldn't shut down the qtile, because of a key map of the command doesn't find easily. 

In order to resolve the problem, I implemented a quick exit widget using the mouse (GUI) and added the widget to the default configuration loaded on error detected.


ramnes:
Thanks for the PR! Why the countdown?

sux2mfgj:
Thank you for your comment.

I want to prevent qtile shut down accidentally.

sux2mfgj:
I implemented the countdown style, but there are any styles. How do you think about it?

libqtile/widget/quick_exit.py
```diff
@@ -0,0 +1,76 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from enum import Enum
+
+from . import base
+from .. import bar
+
+
+class State(Enum):
```
flacjacket: You can just use `auto()` for the values here.

libqtile/widget/quick_exit.py
```diff
@@ -0,0 +1,76 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from enum import Enum
+
+from . import base
+from .. import bar
```
flacjacket: I know this is done throughout, but I feel it is better to do a fully qualified import from `libqtile`

libqtile/widget/quick_exit.py
```diff
@@ -0,0 +1,76 @@
...
+        self.draw()
+
+    def button_press(self, x, y, button):
+        if self.state == State.Neutral:
+            self.state = State.Counting
+            self.update()
+            return
+
+        if self.state == State.Counting:
+            self.qtile.stop()
```
flacjacket: Shouldn't a button press while it is counting reset it back to neutral?

ramnes: My thoughts also

libqtile/widget/quick_exit.py
sux2mfgj: thanks. I change to use auto.

libqtile/widget/quick_exit.py
sux2mfgj: I fixed it.

libqtile/widget/quick_exit.py
sux2mfgj: I agree with your comments. I changed this behavior.
https://github.com/qtile/qtile/pull/1389/commits/dba37a6bd5a06710f2899d995de5df65f205654e

libqtile/widget/quick_exit.py
sux2mfgj: https://github.com/qtile/qtile/pull/1389/commits/cbee326059f76d30451bc02245a4aa3229fc4ceb

sux2mfgj:
There is no enum.auto at python 3.5, so the test failed. Do we have to use any library that can fill the gap?  e.g. [aenum](https://pypi.org/project/aenum/).

sux2mfgj:
I removed enum and use boolean instead.

flacjacket:
Great, thanks for this!

PR #1390, opened 2019-08-11
Change to run the static checker(mypy) with make.

sux2mfgj:
In the CI, the code is checked by the mypy, but ckpatch command in Makefile has no checking the test. I think it is forgot to add the test. This patch fixes the problem.


flacjacket:
Looks fine, thanks!

PR #1391, opened 2019-08-11
Delete groups if there are not configured any more after restarting. (#1315)

### LINKED ISSUE #1315: GroupBox continues to show old group names
After fixing a ConfigError in config or adding new groups to it and restarting QTile, GroupBox continues to show old tags.

For example, I have 5 tags at start: `1 2 3 4 5`
After I remove "5" and add "6" GroupBox shows it like: `1 2 3 4 6 5`

or after fixing a config error: `1 2 3 4 5 a s d f u i o p`

whatacold:
Hi,

This fixes #1315, but I can't make it to set up the dependencies to run pytest successfully. I'm happy to adjust the code if it doesn't meet you expectation.

P.S. Is there any docker images to run pytest locally?

### DIFF
--- a/libqtile/state.py
+++ b/libqtile/state.py
@@ -46,10 +46,10 @@ class QtileState:
         QtileState.
         """
         for (group, layout, label) in self.groups:
-            try:
+            if group in qtile.groups_map:
                 qtile.groups_map[group].layout = layout
-            except KeyError:
-                qtile.add_group(group, layout, label=label)
+            else:
+                qtile.delete_group(group)
 
         for (screen, group) in self.screens.items():
             try:
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -340,11 +340,17 @@ def test_delgroup(qtile):
     self = qtile
 
     self.test_window("one")
+    self.test_window("two")
+    assert qtile.c.group.info()["name"] == 'a'
     for i in ['a', 'd', 'c']:
         self.c.delgroup(i)
+    # There must be at lease one group
     with pytest.raises(CommandException):
         self.c.delgroup('b')
 
+    # The windows are moved to other existed group
+    assert "one" in self.c.groups()["b"]["windows"]
+    assert "two" in self.c.groups()["b"]["windows"]
 
 @manager_config
 @no_xinerama
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -352,6 +352,7 @@ def test_delgroup(qtile):
     assert "one" in self.c.groups()["b"]["windows"]
     assert "two" in self.c.groups()["b"]["windows"]
 
+
 @manager_config
 @no_xinerama
 def test_nextprevgroup(qtile):


ramnes:
There is no Docker image as of today, but that could be a nice improvement as the test suite is indeed a tad difficult to setup. The other way to check being looking at Travis, which is showing us the green light right here. :wink: 

I don't remember this part of the code, are you sure that we can safely delete rather than add in every scenarios? I'll have to double check this before merging unless someone else that knows this part better steps in.

whatacold:
> The other way to check being looking at Travis, which is showing us the green light right here.

That's why I create this PR first, though I haven't run the test suite locally :)

> I don't remember this part of the code, are you sure that we can safely delete rather than add in every scenarios?

I'm not very sure, as I don't have a deep understanding of the codebase.
I just tested it with these scenarios on my side:

1. Change groups from `groups = [Group(i) for i in "dfui"]` to `groups = [Group(i) for i in "fui"]`, with/without some terminals open in group `d`

   The windows open in group `d`, if any, and their layout are moved to the next group `f`.
   
   But if there are windows in `f` already, they are also moved to the next group, `u`. *This should be a problem*, I guess.

   But, if `f` is the last group, all the windows will be in `f`.

1. Delete all groups, change groups from `groups = [Group(i) for i in "dfui"]` to `groups = None`

   Well, this results in an `IndexError: list index out of range`, in `manager.py`

So, it seems there is some room to improve.

whatacold:
As for the Docker image, which base image do you think will be a good start point? I will try to make one if I have some time at hand.

flacjacket:
> I don't remember this part of the code, are you sure that we can safely delete rather than add in every scenarios? I'll have to double check this before merging unless someone else that knows this part better steps in.

I have the same reservations on this. I feel like the main thing to see would be what happens to windows that may be on the groups that are removed, to be sure they are appropriately moved. This would be a great thing to have a unit test for to be sure this doesn't do anything nasty.

whatacold:
> I feel like the main thing to see would be what happens to windows that may be on the groups that are removed, to be sure they are appropriately moved. This would be a great thing to have a unit test for to be sure this doesn't do anything nasty.

Totally agree.

I'm trying to make a Docker image right now, so that I can test it locally. What distro do you use to setup the test environment?

ramnes:
Travis tests use Ubuntu 16.04 (Xenial), so I guess it would make a safe choice to use the same base until we upgrade both.

whatacold:
I've made an [Dockerfile](https://github.com/whatacold/qtile-docker) based on the official python image, which unfortunately is based on Debian :(

The result of `make check` is `2 failed, 377 passed, 1 skipped, 1 warnings in 2906.42s (0:48:26)`, which is not too bad, except that it take too much time, I'm still trying to figure out why they fails.

Besides that, the container doesn't quit when pytest finishes, also don't know why.

tych0:
So I cherry-picked a bunch of the independent patches in this thread and pushed them to master. Then I took the rest, rebased them, fixed all the new linter stuff that we've added, and squashed them down here: https://github.com/tych0/qtile/commit/af5061bb60b741c8d27cc8562acd05e5a1264343

I'm still not sure I understand the purpose of FloatingTile, since I haven't actually read the code that closely. We at least need a better commit message before this goes in.

tych0:
So based on my read of this, I think we can just drop FloatingTile, which was the only part I really objected to. I don't see that it really does anything?

tych0:
Well it must do something, because the tests fail without it :)

tych0:
Oh, crap. All of these replies are on the wrong thread. Derp derp :)

tych0:
Ok, since I hadn't read this one before: I agree that it's probably not a good idea to delete these groups. The way we store what group a window was in is actually an X atom, so we don't have a way to detect (short of iterating through all the available windows and querying their atoms) whether anything is in the group or not. So I think we should keep these groups around; I'll go ahead and close this.

ramnes:
How would you fix #1315 then?

tych0:
I wouldn't, I don't think it's a bug.

PR #1396, opened 2019-08-19
Fix wallpaper_command commandline building

markhellewell:
Also, refresh wallpaper list when button is pressed

### DIFF
--- a/libqtile/widget/wallpaper.py
+++ b/libqtile/widget/wallpaper.py
@@ -81,8 +81,9 @@ class Wallpaper(base._TextBox):
         else:
             self.text = self.label
         if self.wallpaper_command:
-            self.wallpaper_command.append(cur_image)
-            subprocess.call(self.wallpaper_command)
+            wallpaper_command = self.wallpaper_command.copy()
+            wallpaper_command.append(cur_image)
+            subprocess.call(wallpaper_command)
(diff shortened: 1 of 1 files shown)


libqtile/widget/wallpaper.py
```diff
@@ -81,8 +81,9 @@ def set_wallpaper(self):
         else:
             self.text = self.label
         if self.wallpaper_command:
-            self.wallpaper_command.append(cur_image)
-            subprocess.call(self.wallpaper_command)
+            wallpaper_command = self.wallpaper_command.copy()
+            wallpaper_command.append(cur_image)
+            subprocess.call(wallpaper_command)
             self.wallpaper_command.pop()
```
ramnes: Shouldn't we drop this `pop()`, then?

markhellewell: Sorry, you’re right, have updated

ramnes:
What does it fix, exactly?

markhellewell:
Actually, you know what? I fixed a problem that only exists in the 0.10.7 version which ships with ubuntu 18.04. In that version the commandline grows and grows because the list never gets popped :(

Disregard this PR. I’ll get set up with a way to run a more recent version of qtile!

Issue #1398, opened 2019-08-20
docs: List missing requirements for widgets


pawamoy:
For example, when trying to use the `Wlan` widget, I was getting the `ImportError` kind of widget. I had to look at the source to see that it tries to import
1. `pythonwifi`
2. `iwlib` if the previous failed

On Python 3, `pythonwifi` generates syntax errors (`iwlibs.py, line 113`), so I went with `iwlib` only, installed with `sudo python3 -m pip install iwlib`.

For the MPD widgets, I had to install `python-mpd` and `python-mpd2` with the same method.

Of course, seeing https://github.com/qtile/qtile/issues/587, I'm not asking you to put `sudo python3 -m pip install` in the docs. I used that because I personally installed QTile in my system Python libraries (didn't know a better way to make it accessible when booting, but maybe installed as `pip --user` would have worked as well). Rather, we could just list the required `apt-get` or `pip` packages.

ramnes:
See #1354 for more discussion around that topic.

pawamoy:
Thank you for the link to the previous issue.

Indeed using setuptools' `extras_require` could be nice.

For example:

```python
extras_require={
    "wlan":  ["iwlib"],
    "mpd": ["python-mpd", "python-mpd2"],
}
```
Then one could run `pip install qtile[wlan,mpd]`, even if Qtile has already been installed with `pip`.

I left out the version specifiers, but I guess we should add them if we start doing this.

I think the easiest to start with is to list the requirements in the docs at least, for each widget. This is where people learn about the built-in widgets :slightly_smiling_face:

ramnes:
That'd be great, we have to start somewhere!

PR #1400, opened 2019-08-20
docs: Add missing requirement for Wlan widget

pawamoy:
For issue #1398 

What do you think of this (english is not my main language, so I kept it short)?

### DIFF
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -55,7 +55,14 @@ except ImportError:
 
 
 class Wlan(base.InLoopPollText):
-    """Displays Wifi ssid and quality"""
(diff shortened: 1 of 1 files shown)


ramnes:
LGTM, waiting to see what others think

ramnes:
Okay, let's go ahead.

Could you please take the opportunity to remove all the `pythonwifi` stuff, here or in another PR? The library isn't maintained and we don't support Python 2.7 anymore, so let's get this out.

pawamoy:
Of course, I'll do that in another PR since this one is aimed at the docs.

ramnes:
Alrighty, thanks!

PR #1404, opened 2019-08-22
Fix pytest deprecation of pytest.config global.

whatacold:
Per https://docs.pytest.org/en/latest/deprecations.html#pytest-config-global

> pytest.config global
> Removed in version 5.0.
> 
> The pytest.config global object is deprecated. Instead use request.config (via the request fixture) or if you are a plugin author use the pytest_configure(config) hook. Note that many hooks can also access the config object indirectly, through session.config or item.config for example.

It prevents the WM process to be forked in the subprocess in conftest.py, so I fixed it, and all the tests passed on my side.

Hopefully it won't break the CI.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -250,9 +250,10 @@ class Qtile:
     is done.  Windows can be spawned for the qtile instance to interact with
     with various `.test_*` methods.
     """
-    def __init__(self, sockfile, display):
+    def __init__(self, sockfile, display, debug_log):
         self.sockfile = sockfile
         self.display = display
+        self.log_level = logging.DEBUG if debug_log else logging.INFO
 
         self.proc = None
         self.c = None
@@ -262,10 +263,9 @@ class Qtile:
         rpipe, wpipe = multiprocessing.Pipe()
 
         def run_qtile():
-            llvl = logging.DEBUG if pytest.config.getoption("--debuglog") else logging.INFO
             kore = xcore.XCore()
             try:
-                init_log(llvl, log_path=None, log_color=False)
+                init_log(self.log_level, log_path=None, log_color=False)
                 q = QtileManager(kore, config_class(), self.display, self.sockfile)
                 q.loop()
             except Exception:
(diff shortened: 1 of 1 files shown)


flacjacket:
Have you tried running the tests with the latest pytest version? There are some places we limit that which would be nice to roll back.

whatacold:
I'm sorry, but I don't think I get the point as my English is not very good.

Do you mean that this will fail with the lastest pytest, or that I should run it with the lastest to ensure nothing is break?

pawamoy:
I ran it with latest pytest (5.1.2) and everything passes except for `test_svg_scaling`, but I don't think it's related to this PR.

```
    def assert_distortion_less_than(distortion, bad_distortion, factor=0.3):
        for test_val, bad_val in zip(distortion, bad_distortion):
>           assert test_val < (bad_val * factor)
E           assert inf < (22.414 * 0.3)

test/test_images2.py:112: AssertionError
```

whatacold:
Thanks for running the tests.

And I also encountered this `test/test_images2.py:112: AssertionError` failure before,
I found that it was because the output of below command was `inf` :

    cmd ['convert', '/tmp/pytest-of-root/pytest-0/audio-volume-muted.png', '/qtile/test/data/comparison_images/audio-volume-muted_good.png', '-metric', 'PSNR', '-compare', '-format', '%[distortion]\n', 'info:']

but I failed to fix it as I'm not familiar with image processing :(

flacjacket:
Sorry for the confusion! In c29dfb5714eb7ac794f4938edef9653da3d18d24 we pinned the pytest version used to be `< 5.0.0`, if this change allows us to revert that, that'd be great! Also, yes, I think there is other problems with that svg test, that we just don't ever hit on travis....

whatacold:
All test suites except two (they are already reported as issues) are passed with pytest 4.6.5.

> In [c29dfb5](https://github.com/qtile/qtile/commit/c29dfb5714eb7ac794f4938edef9653da3d18d24) we pinned the pytest version used to be `< 5.0.0`

What's the reason for this limit?

flacjacket:
> All test suites except two (they are already reported as issues) are passed with pytest 4.6.5.
> 
> > In [c29dfb5](https://github.com/qtile/qtile/commit/c29dfb5714eb7ac794f4938edef9653da3d18d24) we pinned the pytest version used to be `< 5.0.0`
> 
> What's the reason for this limit?

Some of the tests stopped passing, I'd have to go back to check which ones, but limiting the pytest version kept the tests green.

Issue #994, opened 2017-01-06
Qtile crashed after installing python 3.6


zordsdavini:
Sadly I never learn lesson not to be quick with updates :( 

> 2017-01-06 13:05:31,930 libqtile qtile.py:make_qtile():L119  Qtile crashed during startup Traceback (most recent call last):
>   File "/home/arnas/Skleps/qtile/libqtile/scripts/qtile.py", line 116, in make_qtiled
>     state=options.state,
>   File "/home/arnas/Skleps/qtile/libqtile/manager.py", line 175, in __init__
>     self._process_screens()
>   File "/home/arnas/Skleps/qtile/libqtile/manager.py", line 383, in _process_screens
>     self.groups[i],
>   File "/home/arnas/Skleps/qtile/libqtile/config.py", line 268, in _configure
>     i._configure(qtile, self)
>   File "/home/arnas/Skleps/qtile/libqtile/bar.py", line 193, in _configure
>     self.height
>   File "/home/arnas/Skleps/qtile/libqtile/drawer.py", line 244, in __init__
>     self.height,
>   File "/usr/lib/python3.6/site-packages/cairocffi/xcb.py", line 37, in __init__
>     p = cairo.cairo_xcb_surface_create(
> AttributeError: cffi library 'libcairo.so.2' has no function, constant or global variable named 'cairo_xcb_surface_create'
> Traceback (most recent call last):
>   File "/home/arnas/Skleps/qtile/scripts/../bin/qtile", line 33, in <module>
>     qtile.main()
>   File "/home/arnas/Skleps/qtile/libqtile/scripts/qtile.py", line 125, in main
>     q = make_qtile()
>   File "/home/arnas/Skleps/qtile/libqtile/scripts/qtile.py", line 116, in make_qtile
>     state=options.state,
>   File "/home/arnas/Skleps/qtile/libqtile/manager.py", line 175, in __init__
>     self._process_screens()
>   File "/home/arnas/Skleps/qtile/libqtile/manager.py", line 383, in _process_screens
>     self.groups[i],
>   File "/home/arnas/Skleps/qtile/libqtile/config.py", line 268, in _configure
>     i._configure(qtile, self)
>   File "/home/arnas/Skleps/qtile/libqtile/bar.py", line 193, in _configure
>     self.height
>   File "/home/arnas/Skleps/qtile/libqtile/drawer.py", line 244, in __init__
>     self.height,
[...]

PR #1406, opened 2019-08-27
Make it explicit to install cairocffi after xcffib.

whatacold:


dev.sh
```diff
@@ -7,6 +7,11 @@ echo "Creating dev environment in ./venv..."
 python3.7 -m venv venv
 . venv/bin/activate
 pip3.7 install -U pip setuptools
+
+# https://github.com/qtile/qtile/issues/994
+echo "Installing xcffib then cairocffi..."
+pip3.7 install --no-cache-dir 'xcffib >= 0.5.0' && pip3.7 install --no-cache-dir 'cairocffi >= 0.9.0'
```
ramnes: `xcffib` can be pulled from cache, it's just `cairocffi` that needs the `--no-cache-dir` so that it's compiled against `xcffib`.

dev.sh
```diff
@@ -7,6 +7,11 @@ echo "Creating dev environment in ./venv..."
 python3.7 -m venv venv
 . venv/bin/activate
 pip3.7 install -U pip setuptools
+
+# https://github.com/qtile/qtile/issues/994
```
ramnes: I'd rather link to the explanations directly: https://github.com/qtile/qtile/issues/994#issuecomment-497984551

dev.sh
```diff
@@ -7,6 +7,11 @@ echo "Creating dev environment in ./venv..."
...
+echo "Installing xcffib then cairocffi..."
+pip3.7 install --no-cache-dir 'xcffib >= 0.5.0' && pip3.7 install --no-cache-dir 'cairocffi >= 0.9.0'
+
```
ramnes: Either add another `echo` here to say that we're now installing other required packages, or remove the previous one.

ramnes:
Thanks!

Issue #1425, opened 2019-09-05
Errors when running scripts/xephyr


pawamoy:
Just pasting the log if it talks to anybody.

```
% ./scripts/xephyr                                                                                           (master|+1…) 20:23:14
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported high keycode 372 for name <I372> ignored
>                   X11 cannot support keycodes above 255.
>                   This warning only shows for the first high keycode.
> Internal error:   Could not resolve keysym XF86MonBrightnessCycle
> Internal error:   Could not resolve keysym XF86RotationLockToggle
Errors from xkbcomp are not fatal to the X server
xterm: cannot load font "-Misc-Fixed-medium-R-*-*-13-120-75-75-C-120-ISO10646-1"
xterm: cannot load font "-misc-fixed-medium-r-semicondensed--13-120-75-75-c-60-iso10646-1"
2019-09-05 20:23:25,443 libqtile utils.py:safe_import():L192  Unmet dependencies for optional Widget: '.widget.launchbar.LaunchBar', No module named 'xdg.IconTheme'
2019-09-05 20:23:25,526 libqtile manager.py:setup_python_dbus():L310  importing dbus/gobject failed, dbus will not work.
Traceback (most recent call last):
  File "/media/data/dev/forks/qtile/scripts/../bin/qtile", line 33, in <module>
    qtile.main()
  File "/media/data/dev/forks/qtile/libqtile/scripts/qtile.py", line 130, in main
    q = make_qtile()
  File "/media/data/dev/forks/qtile/libqtile/scripts/qtile.py", line 124, in make_qtile
    state=options.state,
  File "/media/data/dev/forks/qtile/libqtile/core/manager.py", line 193, in __init__
    self._process_screens()
  File "/media/data/dev/forks/qtile/libqtile/core/manager.py", line 404, in _process_screens
    self.groups[i],
  File "/media/data/dev/forks/qtile/libqtile/config.py", line 255, in _configure
    i._configure(qtile, self)
  File "/media/data/dev/forks/qtile/libqtile/bar.py", line 192, in _configure
    self.height
  File "/media/data/dev/forks/qtile/libqtile/drawer.py", line 241, in __init__
    self.height,
  File "/home/pawamoy/data/isolated/virtualenvs/qtile/local/lib/python3.7/site-packages/cairocffi/xcb.py", line 37, in __init__
    p = cairo.cairo_xcb_surface_create(
[...]

Issue #1420, opened 2019-09-05
Default config uses deprecated lazy


pawamoy:
Seems to be related to #644.

The default config imports `lazy` from `libqtile.command`, which is deprecated.

It should be imported from `libqtile.lazy` instead. Does it work the same way?

ramnes:
Good catch. IIRC it works exactly the same way.

pawamoy:
I just tried to replace `from libqtile.command import lazy` by `from libqtile.lazy import lazy` and got a `ConfigError` :confused:

pawamoy:
Hmm it might just be because my installed version of QTile lags behind master. 59 new commits since 0.14.2, with a lot of structural changes.

PR #1427, opened 2019-09-06
docs: Fix command to build cffi module

pawamoy:
Fix issue #1424

PR #1428, opened 2019-09-06
refactor: Use lazy.lazy in default config instead of command.lazy (deprecated)

pawamoy:
Issue #1420. Behavior should stay exactly the same.

I was not able to run the tests locally (see #1425), let see what Travis says.


pawamoy:
Same as in #1427, timeout. I think you can restart the jobs if have the privilege to do so on Travis.

ramnes:
Thanks!

Issue #1416, opened 2019-08-31
Remove pythonwifi from the Wlan widget


ramnes:
The library isn't maintained and we don't support Python 2.7 anymore, so let's get this out.

cc @pawamoy

PR #1429, opened 2019-09-06
refactor: Remove pythonwifi related code in Wlan widget

pawamoy:
Issue #1416.

If I understand correctly, when `iwlib` is not installed the `ImportError` will be caught a level above, and the widget will render "Import Error", so I don't have to handle the case in `wlan.py`.

I did not find any other occurrence of `pythonwifi` throughout the project.


ramnes:
Ah, great, thank you again!

Issue #1408, opened 2019-08-28
qtile-cmd: don't output help message when running a function


pawamoy:
For some (most?) functions, `qtile-cmd` outputs the list of available function even when successfully calling one.

For example, if I run `qtile-cmd -o group -f setlayout -a max`, I still get this output:
```
-o group -f commands       	  Returns a list of possible commands for this object
-o group -f doc            	* Returns the documentation for a specified command name
-o group -f eval           	* Evaluates code in the same context as this function
-o group -f focus_back     	  Focus the window that had focus before the current one got it.
-o group -f focus_by_name  	* Focus the first window with the given name. Do nothing if the name is
-o group -f function       	* Call a function with current object as argument
-o group -f info           	  Returns a dictionary of info for this group
-o group -f info_by_name   	* Get the info for the first window with the given name without giving it
-o group -f items          	* Returns a list of contained items for the specified name
-o group -f next_window    	  Focus the next window in group.
-o group -f prev_window    	  Focus the previous window in group.
-o group -f set_label      	* Set the display name of current group to be used in GroupBox widget.
-o group -f setlayout      	*
-o group -f switch_groups  	* Switch position of current group with name
-o group -f toscreen       	* Pull a group to a specified screen.
-o group -f unminimize_all 	  Unminimise all windows in this group
```

...which adds a lot of noise in scripts. Of course I can redirect output to `/dev/null` in scripts, but it's cumbersome to do on the command line when testing :slightly_smiling_face: 

Same with `qtile-cmd -o window -f kill` and many others.

`qtile-cmd` should only output this when the selected function (with `-f`) does not exist, or when no arguments are given after `-o TYPE`.

ramnes:
Yep, can confirm.

flacjacket:
Thanks for the fix!

PR #1430, opened 2019-09-06
fix: Don't print help if qtile-cmd function returns nothing

pawamoy:
Issue #1408.

Just removed the `print` when the function returns `None`.


flacjacket:
This looks good, thanks

PR #1432, opened 2019-09-06
docs: Fix missing psutil mock

### LINKED ISSUE #1431: docs: Missing deps when building widget docs gives "alias to ImportErrorWidget"
See http://docs.qtile.org/en/latest/manual/ref/widgets.html#memory for example.

I guess the widget dependencies are not installed while building the docs, resulting in Sphinx telling the widget is an alias to `libqtile.widget.import_error.make_error.<locals>.ImportErrorWidget`.

EDIT: okay I see where the deps are listed: in `docs/conf.py`. Indeed `mpd` is present but `psutil` is not, so the `Memory` widget's docs do not build.

pawamoy:
The `Memory` and `Net` widgets' docs were not build correctly due to `psutil` not being mocked in `docs/conf.py`. Fixes #1431

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -36,6 +36,7 @@ MOCK_MODULES = [
     'iwlib',
     'keyring',
     'mpd',
+    'psutil',
     'trollius',
     'xcffib',
     'xcffib.randr',


ramnes:
Thanks!

PR #1433, opened 2019-09-07
Make an example of how to run Qtile on Fedora. (Fix #1417)

### LINKED ISSUE #1417: docs: how to install Qtile from source in detail?
Hi,

Installing Qtile via a distro's package manager is a no-brainer.

But, If I want to try the source of master, I'm not sure how to do that exactly. There is a [section](http://docs.qtile.org/en/latest/manual/install/index.html#installing-from-source) on the site about how to install dependent packages, but it doesn't mention how to install Qtile itself actually.

My thought is to:
1. create a system-wide virtualenv
2. copy the source to somewhere, e.g. `/usr/local/bin/`
3. create a new file `/usr/share/xsessions/qtile-venv.desktop`

Am I right? How do you do that?

whatacold:
Add an example about how to run Qtile within a virtualenv on Fedora, this fixes #1417 . Correct me if anything is incorrect, be it English grammar or the content.

### DIFF
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -45,14 +45,14 @@ cairocffi
 ---------
 
 Qtile uses cairocffi_ with XCB support via xcffib. You'll need ``libcairo2``,
-the underlying library used by the binding.  You should be sure before you
-install cairocffi that xcffib has been installed, otherwise the needed
+the underlying library used by the binding.  You should **be sure before you
+install cairocffi that xcffib has been installed**, otherwise the needed
 cairo-xcb bindings will not be built.  Once you've got the dependencies
 installed, you can use the latest version on PyPI:
 
 .. code-block:: bash
 
-    pip install cairocffi
+    pip install --no-cache-dir cairocffi
 
 .. _cairocffi: https://pythonhosted.org/cairocffi/overview.html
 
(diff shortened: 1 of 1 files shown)


docs/manual/install/index.rst
```diff
@@ -45,14 +45,14 @@ cairocffi
...
-install cairocffi that xcffib has been installed, otherwise the needed
+the underlying library used by the binding.  You should **be sure before you
+install cairocffi that xcffib has been installed**, otherwise the needed
 cairo-xcb bindings will not be built.  Once you've got the dependencies
 installed, you can use the latest version on PyPI:
 
 .. code-block:: bash
 
-    pip install cairocffi
+    pip install --no-cache-dir cairocffi
```
flacjacket: This is nice, we may want to add this to the error message that we print if cairocffi is not properly installed.

docs/manual/install/index.rst
```diff
@@ -92,3 +92,36 @@ Stable versions of Qtile can be installed from PyPI:
 As long as the necessary libraries are in place, this can be done at any point,
 however, it is recommended that you first install xcffib to ensure the
 cairo-xcb bindings are built (see above).
+
+To recap, let's take it as an example on Fedora distro to run qtile within a virtualenv,
```
flacjacket: `let's` to `lets`

docs/manual/install/index.rst
```diff
@@ -92,3 +92,36 @@ Stable versions of Qtile can be installed from PyPI:
...
+the virtualenv is optional (assume that the git repo is cloned at ``~/local/qtile/`` of user ``foo``):
+
+#. (optional) Create a virtualenv ``~/local/qtile/qtile-env/``, and install the dependencies there.
+#. (optional) Create a glue shell to take advantage of the virtualenv.
```
flacjacket: Is this the `/home/foo/local/qtile/qtile-venv-entry` file? Is that optional? May be good to note where the output should go.

docs/manual/install/index.rst
whatacold: Where would be the best place to print this error message?

flacjacket: Ah, I was thinking in `setup.py`, but it looks like that error message already has this.

whatacold:
@flacjacket, thanks for the review.

I rephrased the steps to make it more clear on how to run it within a virtualenv step by step.

docs/manual/install/index.rst
```diff
@@ -92,3 +92,41 @@ Stable versions of Qtile can be installed from PyPI:
...
 cairo-xcb bindings are built (see above).
+
+The above steps are sufficient to run Qtile directly, but there are some extra works if you want to run it within a virtualenv. Here are the steps on a Fedora system for user ``foo``, it should work on other Linux systems too.
+
+#. Clone the repo as ``~/local/qtile/``.
+
+   .. code-block:: bash
+
+       mkdir -p ~/local/
+       git clone git://github.com/qtile/qtile.git
```
flacjacket: `cd ~/local/`

whatacold: Fixed, thanks!

whatacold:
@flacjacket I would like to do a squashed rebase when you think it's ok to do so, so that the commit will be more relevant.

flacjacket:
This is fine, thanks for the updates!

flacjacket:
Feel free to squash if you would like to, otherwise we can just pull in the commits as-is.

whatacold:
OK, rebased and squashed.

flacjacket:
Sounds good, thanks!

PR #1438, opened 2019-09-08
docs: Fix parameter type of Screen's top/left/right/bottom

pawamoy:
Previously typed as `List of Bar/Gap objects, or None`, which is not correct: these parameter do not accept lists, only a single bar or a single gap. Too bad, I wanted to setup two top bars :wink:

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -216,10 +216,10 @@ class Screen(CommandObject):
 
     Parameters
     ==========
-    top: List of Gap/Bar objects, or None.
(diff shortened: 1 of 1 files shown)


ramnes:
I worked on this with #1387, and it's already functional (but not 100% correct implementation-wise). But fair enough, let's merge this meanwhile and I'll revert it in my PR. :smile:

pawamoy:
Ah! I didn't see that PR. Well, great! I hope you manage to get it ready :slightly_smiling_face:

PR #1451, opened 2019-09-15
Fix #1394: handle empty HOME completion in test_bar.py::test_completion.

### LINKED ISSUE #1394: test_bar.py::test_completion fails if HOME is empty
Hi,

https://github.com/qtile/qtile/blob/master/test/test_bar.py#L106 has the assertion `assert c.complete("~") != "~"`, it fails when there are neither directories nor binaries there.

whatacold:


test/test_bar.py
```diff
@@ -103,7 +105,18 @@ def test_completion():
     c.reset()
     assert c.complete("/bin") != "/bin/"
     c.reset()
-    assert c.complete("~") != "~"
+
+    home_dir = os.path.expanduser("~")
```
flacjacket: Hmm, this is interesting that we would change the tests depending on whether or not there is something is home. Maybe it would be better to instead create a temporary directory and test the different behaviors we expect on files that we have more control over?

whatacold: Good idea, I've update it so that both the cases of empty dir and non-empty dir can be covered.

PR #1452, opened 2019-09-15
Center-align groupboxes by default.

whatacold:
The difference can be noticed by setting `widget.GroupBox(center_aligned=True)` or not,
as illustrated by below screenshots:

1. before setting

2. after setting


ramnes:
Makes sense to me. Next step would be to deprecate that option, I guess.

PR #1457, opened 2019-09-17
added option "option" to keyboardLayout widget

zordsdavini:
setxkbmap has option -option where you can add compose key, make caps lock work as backspace, switch led etc. So, with this you can add this

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -113,7 +113,7 @@ class KeyboardLayout(base.InLoopPollText):
         command = ['setxkbmap']
         command.extend(keyboard.split(" "))
         if self.options:
-            command.extend(['-options', self.options])
+            command.extend(['-option', self.options])
         try:
             self.call_process(command)
         except CalledProcessError as e:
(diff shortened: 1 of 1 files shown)


libqtile/widget/keyboardlayout.py
```diff
@@ -43,6 +44,7 @@ class KeyboardLayout(base.InLoopPollText):
         ("configured_keyboards", ["us"], "A list of predefined keyboard layouts "
             "represented as strings. For example: "
             "['us', 'us colemak', 'es', 'fr']."),
+        ("options", None, "list of setxkbmap options. Ex., 'compose:menu,grp_led:scroll'"),
```
flacjacket: `list` may not be the best description of this, as it is really a string. It may be worth changing this to `additional options passed to setxkbmap via -options`

pawamoy: The option name is `-option`. It's correctly written `-option` in the code though :slightly_smiling_face:

zordsdavini: You are right. At first I thought to add this as list in real, but checked how it is with `setxkbmap` displayed and `options` become string. Fixed. This diff was in my mind for two years, just time is good to do all small stuffs because I had to build config from the begining

flacjacket:
Nope, looks good, thanks! And sorry for the slow reviews :)

PR #1458, opened 2019-09-17
fix: Fix ratiotile shuffle commands

pawamoy:
Just noticed it while automating screenshots.

### DIFF
--- a/libqtile/layout/ratiotile.py
+++ b/libqtile/layout/ratiotile.py
@@ -310,12 +310,12 @@ class RatioTile(_SimpleLayoutBase):
     def cmd_shuffle_down(self):
         if self.clients:
             self.clients.rotate_up()
-            self.self.group.layout_all()
+            self.group.layout_all()
 
     def cmd_shuffle_up(self):
         if self.clients:
             self.clients.rotate_down()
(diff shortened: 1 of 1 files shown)


pawamoy:
I just noticed you request standard commit messages in the hacking document, which I did not follow at all :disappointed_relieved:

Three of my Angular-style commits made their way into master. Sorry :pray: !!

I can amend the rest of the commits in my pending PRs if you prefer!

flacjacket:
Thanks for the fix! Appreciate all of the recent work!

Yeah, full commit messages are nice, particularly for bigger changes, but for minor fixes and changes we'll generally let things like that slide.

pawamoy:
No problem! I love Qtile and hacking on it :wink: 

What I meant was that I use a prefix in the commit summary: the type (`fix`, `feat`, `docs`, `chore`, `tests`, `ci`, etc.) and an optional scope (`docs(screenshots)`, `fix(layout)`, etc.), and it's not the standard way described [here](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), though it's even more informative. My worries were about this prefix.

For bigger commits, yes, I also add a body to the message.

PR #1465, opened 2019-09-30
Improve readability of the Volume widget.

whatacold:
Currently, the text of the volume widget is not very readable, I have to explicitly place a textbox before it in order to work around it, so I think it will be better to let users customize the text format.

    bottom=bar.Bar(
        [
            #...,
            widget.TextBox('Vol:'),
            widget.Volume(),
            #...,
        ],

As for the tests, I don't know if it's acceptable to set the member `self.volume` and to call the "private" method `self._update_drawer()` directly, what do you think?

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -58,10 +58,12 @@ class Volume(base._TextBox):
         ("device", "default", "Device Name"),
         ("channel", "Master", "Channel"),
         ("padding", 3, "Padding left and right. Calculated if None."),
-        ("theme_path", None, "Path of the icons"),
         ("update_interval", 0.2, "Update time in seconds."),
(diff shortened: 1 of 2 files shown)


zordsdavini:
:+1: 
but maybe we should use common flow and add option `fmt` with default `"Vol: {0}"` instead `text_format` and not updatable prefix `Vol: `? For example I don't like any text around, just simple 67% :)

whatacold:
@zordsdavini , Ah, you're right, the `fmt` option in base.py is already sufficient :(

Then there is nothing in this PR except the additional tests, I'll close it if you think it's useless.

zordsdavini:
you can leave tests if they works

flacjacket:
This looks great, thanks!

PR #1466, opened 2019-10-02
Net Widget Overhaul

WhosMyName:
I've added some changes to make the Net Widget more dynamic and useful for multi-NIC devices:  
* takes list, strings and None as param for interfaces  
* allows multiple NICs to be displayed  
* default option will show a summary of all active NICs  
* no changes in config for users as a single specified NIC will be cast to a list  
* connection speed prefixes scale into the next decades  
  
pep8 and flake8 ain't complaining and I've been using these changes for the past week:  
I'd say it's stable.

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -29,30 +29,39 @@ class Net(base.ThreadedPollText):
     """Displays interface down and up speed"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('interface', None, 'List of interfaces to monitor or None to displays all active NICs combined'),
+        ('interface', None, 'List of interfaces or single NIC as string to monitor, None to displays all active NICs combined'),
         ('update_interval', 1, 'The update interval.'),
     ]
 
     def __init__(self, **config):
         base.ThreadedPollText.__init__(self, **config)
         self.add_defaults(Net.defaults)
+        #if not isinstance(self.interface, list):
+        #    if self.interface is None:
+        #        self.interface = "all"
+        #    self.interface = [self.interface]
         if not isinstance(self.interface, list):
             if self.interface is None:
-                self.interface = "all"
-            self.interface = [self.interface]
+                self.interface = ["all"]
+            elif isinstance(self.interface, str):
+                self.interface = [self.interface]
+            else:
+                raise ArgumentError("Invalid Argument passed: %s\nAllowed Types: List, String, None" % self.interface)
         self.stats = self.get_stats()
 
     def convert_b(self, b):
 
-        letters = ["b", "kb", "Mb", "Gb", "Tb", "Pb", "Eb", "Zb", "Yb"]
+        letters = ["kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
         factor = 1000.0
-        i = 0
+        unit = "B"
 
-        while int(b / 1000) > 0:
-            b /= factor
-            i = i + 1
-
-        return b, letters[i]
+        for letter in letters:
+            if b > factor:
+                b /= factor
+                unit = letter
+            else:
+                break
+        return b, unit
 
     def get_stats(self):
         interfaces = {}
@@ -63,10 +72,9 @@ class Net(base.ThreadedPollText):
         else:
             net = psutil.net_io_counters(pernic=True)
             for iface in net:
-                name = iface
                 down = net[iface].bytes_recv
                 up = net[iface].bytes_sent
-                interfaces[name] = {'down': down, 'up': up}
+                interfaces[iface] = {'down': down, 'up': up}
             return interfaces
 
     def _format(self, down, up):
@@ -82,8 +90,8 @@ class Net(base.ThreadedPollText):
         return down, up
 
     def poll(self):
+        ret_stat = []
         try:
-            ret_stat = ""
             for intf in self.interface:
                 new_stats = self.get_stats()
                 down = new_stats[intf]['down'] - \
@@ -96,11 +104,11 @@ class Net(base.ThreadedPollText):
                 down, down_letter = self.convert_b(down)
                 up, up_letter = self.convert_b(up)
                 down, up = self._format(down, up)
-                str_base = "%s: %s%s \u2193\u2191 %s%s "
+                str_base = "%s: %s%s \u2193\u2191 %s%s"
                 self.stats[intf] = new_stats[intf]
-                ret_stat += str_base % (intf, down, down_letter, up, up_letter)
+                ret_stat.append(str_base % (intf, down, down_letter, up, up_letter))
 
-            return ret_stat
+            return " ".join(ret_stat)
         except Exception as excp:
             logger.error('%s: Caught Exception:\n%s',
                          self.__class__.__name__, excp)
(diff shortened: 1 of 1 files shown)


WhosMyName:
It's seems to me that the Docs won't need an update as I couldn't find a detailed usage documentation for specific widgets.
If there is a documentation for specific widgets, I'll update it as soon as possible <3

libqtile/widget/net.py
```diff
@@ -29,43 +29,45 @@ class Net(base.ThreadedPollText):
     """Displays interface down and up speed"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('interface', 'wlan0', 'The interface to monitor'),
+        ('interface', None, 'List of interfaces to monitor or None to displays all active NICs combined'),
```
flacjacket: You should leave in the doc that a string will be interpreted as a single interface.

libqtile/widget/net.py
```diff
@@ -29,43 +29,45 @@ class Net(base.ThreadedPollText):
...
     ]
 
     def __init__(self, **config):
         base.ThreadedPollText.__init__(self, **config)
         self.add_defaults(Net.defaults)
-        self.interfaces = self.get_stats()
+        if not isinstance(self.interface, list):
+            if self.interface is None:
+                self.interface = "all"
+            self.interface = [self.interface]
```
flacjacket: Maybe make this:
```
if self.interface is None:
    self.interface = "all"
elif isinstance(self.interface, str):
    self.interface = [self.interface]
```
to make this a bit more explicit.

libqtile/widget/net.py
```diff
@@ -29,43 +29,45 @@ class Net(base.ThreadedPollText):
...
-            b /= 1000.0
-            letter = 'M'
-        if int(b / 1000) > 0:
-            b /= 1000.0
-            letter = 'G'
-        # I hope no one have more than 999 GB/s
-        return b, letter
 
-    def get_stats(self):
+        letters = ["b", "kb", "Mb", "Gb", "Tb", "Pb", "Eb", "Zb", "Yb"]
```
flacjacket: Does this deal in bits or bytes?

libqtile/widget/net.py
```diff
@@ -29,43 +29,45 @@ class Net(base.ThreadedPollText):
...
 
-        interfaces = {}
-        net = psutil.net_io_counters(pernic=True)
-        for iface in net:
-            name = iface
-            down = net[iface].bytes_recv
-            up = net[iface].bytes_sent
-            interfaces[name] = {'down': down, 'up': up}
+        while int(b / 1000) > 0:
+            b /= factor
```
flacjacket: This should probably check as well that `i < len(letters) - 1`, in case some erroneous case happens where over 1 Yb is reported, at least we won't crash the widget. This should probably also read `b < factor` rather than `int(b / 1000) > 0`.

I'm wondering if we want to get fancy with this. Something like:
```
idx = log(b) / log(1000)
```
to compute the index of the letter. One additional thing that would be needed here is to be sure 0 and (erroneous) negative numbers aren't fed into the log, so maybe that isn't ideal.

libqtile/widget/net.py
```diff
@@ -29,43 +29,45 @@ class Net(base.ThreadedPollText):
...
+            net = psutil.net_io_counters(pernic=False)
+            interfaces["all"] = {'down': net[1], 'up': net[0]}
+            return interfaces
+        else:
+            net = psutil.net_io_counters(pernic=True)
+            for iface in net:
+                name = iface
+                down = net[iface].bytes_recv
+                up = net[iface].bytes_sent
+                interfaces[name] = {'down': down, 'up': up}
```
flacjacket: You can just assign to `interfaces[iface]` and save the `name =` assignment above.

libqtile/widget/net.py
```diff
@@ -81,23 +83,24 @@ def _format(self, down, up):
...
+                down = down / self.update_interval
+                up = up / self.update_interval
+                down, down_letter = self.convert_b(down)
+                up, up_letter = self.convert_b(up)
+                down, up = self._format(down, up)
+                str_base = "%s: %s%s \u2193\u2191 %s%s "
+                self.stats[intf] = new_stats[intf]
+                ret_stat += str_base % (intf, down, down_letter, up, up_letter)
+
+            return ret_stat
```
flacjacket: It is better if we can create a list of strings and then `" ".join()` them.

libqtile/widget/net.py
```diff
@@ -81,23 +83,24 @@ def _format(self, down, up):
...
+                    self.stats[intf]['down']
+                up = new_stats[intf]['up'] - \
+                    self.stats[intf]['up']
+
+                down = down / self.update_interval
+                up = up / self.update_interval
+                down, down_letter = self.convert_b(down)
+                up, up_letter = self.convert_b(up)
+                down, up = self._format(down, up)
+                str_base = "%s: %s%s \u2193\u2191 %s%s "
```
flacjacket: By doing a join with `" "` below, you can drop the trailing space here.

libqtile/widget/net.py
pawamoy: Bytes must be the obvious answer, though I can't say for sure. So it should be a capitalized B. Another question is: does this deal with multiples of `1000` or `1024` (`2**10`)? Former: `B, kB, MB, GB, TB, PB, EB, ZB, YB`. Latter: `B, KiB, MiB, GiB, TiB, PiB, EiB, ZiB, YiB`

libqtile/widget/net.py
pawamoy: This is what I use in another project:

```python
chosen_unit = "B"
for unit in ("KiB", "MiB", "GiB", "TiB"):
    if value > 1000:
        value /= 1024
        chosen_unit = unit
    else:
        break
```

It think it handles the two cases you mentioned. My 2 cents :slightly_smiling_face:

libqtile/widget/net.py
WhosMyName: The doc should be fine now
```python
defaults = [
        ('interface', None, 'List of interfaces or single NIC as string to monitor, None to displays all active NICs combined'),
        ('update_interval', 1, 'The update interval.'),
    ]
```

libqtile/widget/net.py
WhosMyName: As per [psutil.net_io_counters](https://psutil.readthedocs.io/en/latest/#psutil.net_io_counters): it's bytes and not bits.
Currently it's working with multiples of 1000.
I'll have the Bees fixed XD

libqtile/widget/net.py
WhosMyName: May I propose:
```python
def convert_b(self, b):

    letters = ["kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
    factor = 1000.0
    unit = "B"

    for letter in letters:
        if b > factor:
            b /= factor
            unit = letter
        else:
            break
    return b, unit
```
I'm in favor of a solution based on @pawamoy's comment:
* cleaner
* no extra check to make sure the values don't escalate too quickly
* no while loop
* I don't think there's much more to optimize

libqtile/widget/net.py
WhosMyName: Done!
I was pretty blind there...

WhosMyName:
I've added the changes from above in the current commit and will test it now.
I'll do the clean up with the next commit/change(s).

flacjacket:
This looks good, thanks for updating it!

WhosMyName:
Thanks for merging<3

PR #1468, opened 2019-10-09
Allow to move a window to a group, and switch to that group.

whatacold:
At first I just switch to that group always, as I think most of the time it makes sense to switch to the group at the same time. But then I wonder maybe there are folks who are just happy with current behavior of qtile, so I add an additional `switch_group` parameter, which defaults to `False`.

Let me know what you think.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1026,7 +1026,7 @@ class Window(_Window):
             self.height = h
         self._reconfigure_floating(new_float_state=new_float_state)
 
-    def togroup(self, group_name=None, switch_group=False):
+    def togroup(self, group_name=None, *, switch_group=False):
         """Move window to a specified group
 
         Also switch to that group if switch_group is True.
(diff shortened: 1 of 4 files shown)


libqtile/window.py
```diff
@@ -1026,8 +1026,11 @@ def _enablefloating(self, x=None, y=None, w=None, h=None,
             self.height = h
         self._reconfigure_floating(new_float_state=new_float_state)
 
-    def togroup(self, group_name=None):
-        """Move window to a specified group"""
+    def togroup(self, group_name=None, switch_group=False):
```
flacjacket: I think it'd be really cool if we could start to put things like this behind keyword-only arguments, now that we are on python 3. That would make the process of deprecating much easier, as we could easily add/remove named arguments.

whatacold: Good to know keyword-only arguments :)

I leave it as-is for `group_name`, as forcing it to keyword-only will break configurations. Also, I think keyword-only arguments is more readable.

flacjacket:
Looks good, thanks!

PR #1469, opened 2019-10-09
Show current layout and prepare all layouts in the default config.

whatacold:
I had no idea what various layouts look like when I first tried qtile, as I didn't know how to add layouts, which one I was currently using, etc. With this PR, I think it will be a little easier to get started with layouts, hopefully.


ramnes:
Good idea, thanks!

PR #1471, opened 2019-10-10
Changes to make the dev.sh script more generic.

sux2mfgj:
The dev.sh script was specifying the python version to 3.7. This patch makes it more generic.


ramnes:
Thanks!

PR #1475, opened 2019-10-14
Update OS informations

dannycolin:
- Use number instead of codename
- Update Debian information to let the user know that qtile previously in unstable is now in stable since the Buster release

### DIFF
--- a/docs/manual/install/ubuntu.rst
+++ b/docs/manual/install/ubuntu.rst
@@ -2,14 +2,14 @@
 Installing on Debian or Ubuntu
 ==============================
 
-On recent Ubuntu (17.04 Zesty+) and Debian unstable (Sid) versions, there are
+On recent Ubuntu (17.04 or greater) and Debian (10 or greater) versions, there are
 Qtile packages available via:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/manual/install/ubuntu.rst
```diff
@@ -2,14 +2,14 @@
...
-On recent Ubuntu (17.04 Zesty+) and Debian unstable (Sid) versions, there are
+On recent Ubuntu (17.04 or greater) and Debian (10 or greater) versions, there are
 Qtile packages available via:
 
 .. code-block:: bash
 
     sudo apt-get install qtile
 
-On older versions of Ubuntu (Wily+) and Debian testing (Stretch), the
+On older versions of Ubuntu (15.10 or lower) and Debian (9 or lower), the
```
ramnes: The "or lower" isn't appropriate here. `python3-xcffib` and `python3-cairocffi` do not exist in Ubuntu Trusty (14.04), for example.

dannycolin: I've updated the information for both Ubuntu and Debian.

flacjacket:
Thanks!

Issue #1435, opened 2019-09-08
KeyError: 'MemUsed'


pawamoy:
```
Traceback (most recent call last):
  File "/media/data/dev/forks/qtile/scripts/../bin/qtile", line 33, in <module>
    qtile.main()
  File "/media/data/dev/forks/qtile/libqtile/scripts/qtile.py", line 130, in main
    q = make_qtile()
  File "/media/data/dev/forks/qtile/libqtile/scripts/qtile.py", line 124, in make_qtile
    state=options.state,
  File "/media/data/dev/forks/qtile/libqtile/core/manager.py", line 193, in __init__
    self._process_screens()
  File "/media/data/dev/forks/qtile/libqtile/core/manager.py", line 404, in _process_screens
    self.groups[i],
  File "/media/data/dev/forks/qtile/libqtile/config.py", line 255, in _configure
    i._configure(qtile, self)
  File "/media/data/dev/forks/qtile/libqtile/bar.py", line 204, in _configure
    i._configure(qtile, self)
  File "/media/data/dev/forks/qtile/libqtile/widget/base.py", line 440, in _configure
    _TextBox._configure(self, qtile, bar)
  File "/media/data/dev/forks/qtile/libqtile/widget/base.py", line 361, in _configure
    self.formatted_text,
  File "/media/data/dev/forks/qtile/libqtile/widget/base.py", line 317, in formatted_text
    return self.fmt.format(self._text)
KeyError: 'MemUsed'
```

I believe the `Memory` widget is responsible.

pawamoy:
Hmm in fact it was because the xephyr script tries to load my own config file. Trying with the default config this traceback disappear.

However I'm not sure why the exception happens in xephyr and not in my main X display.

pawamoy:
Should have been fixed by #1481. If not, still closing since it was a very specific scenario and had no impact at all. Reopen if needed.

PR #1481, opened 2019-10-20
Fix for memory widget causing crash

b10011:
The memory widget caused a crash of the WM.

### DIFF
--- a/libqtile/widget/cpu.py
+++ b/libqtile/widget/cpu.py
@@ -1,0 +1,54 @@
+# Copyright (c) 2019 Niko Järvinen (b10011)
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from libqtile.widget import base
+import psutil
+
+
+class CPU(base.ThreadedPollText):
+    orientations = base.ORIENTATION_HORIZONTAL
+
+    defaults = [
+        ("update_interval", 1.0, "Update interval for the CPU widget"),
+        (
+            "format",
+            "CPU {freq_current}GHz {load_percent}%",
+            "CPU display format",
+        ),
+    ]
+
+    def __init__(self, **config):
+        base.ThreadedPollText.__init__(self, **config)
+        self.add_defaults(CPU.defaults)
+
+    def tick(self):
+        self.update(self.poll())
+        return self.update_interval
+
+    def poll(self):
+        variables = dict()
+
+        variables["load_percent"] = round(psutil.cpu_percent(), 1)
+        freq = psutil.cpu_freq()
+        variables["freq_current"] = round(freq.current / 1000, 1)
+        variables["freq_max"] = round(freq.max / 1000, 1)
+        variables["freq_min"] = round(freq.min / 1000, 1)
+
+        return self.format.format(**variables)
--- a/libqtile/widget/cpu.py
+++ b/libqtile/widget/cpu.py
@@ -35,7 +35,7 @@ class CPU(base.ThreadedPollText):
     ]
 
     def __init__(self, **config):
-        base.ThreadedPollText.__init__(self, **config)
+        super().__init__(**config)
         self.add_defaults(CPU.defaults)
 
     def tick(self):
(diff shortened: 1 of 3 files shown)


b10011:
Whoops, now the fix for non-graph CPU widget is also in the same pull req.

libqtile/widget/memory.py
```diff
@@ -39,27 +39,33 @@ class Memory(base.ThreadedPollText):
     SwapFree: Returns amount of swap free
     SwapUsed: Returns amount of swap in use
 """
+
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ("fmt", "{MemUsed}M/{MemTotal}M", "Formatting for field names.")
+        ("format", "{MemUsed}M/{MemTotal}M", "Formatting for field names."),
```
pawamoy: I don't think you should rename it `format`. It can shadow the built-in format function, making its use confusing. All other widgets also use `fmt`, this would break consistency. Wouldn't it be a breaking change as well, since users would need to update their code?

pawamoy:
CPU widget, nice!! :rocket: 

Just added comments about `fmt` versus `format` :slightly_smiling_face:

libqtile/widget/cpu.py
```diff
@@ -0,0 +1,54 @@
...
+import psutil
+
+
+class CPU(base.ThreadedPollText):
+    orientations = base.ORIENTATION_HORIZONTAL
+
+    defaults = [
+        ("update_interval", 1.0, "Update interval for the CPU widget"),
+        (
+            "format",
```
pawamoy: Same remark as for the memory widget.

b10011:
@pawamoy You can try it yourself, change the `format` to `fmt`, and it crashes again. There were two problems: usage of `fmt` and how it calls the inherited class.

The reason it crashes is because `libqtile.widget.base._TextBox` has property method `formatted_text` that tries to format the `fmt` string with `self._text`. That will result in `"{MemUsed}M/{MemTotal}M".format("N/A")` which will crash it. If we override that method with one that returns "hello", then our widget only says "hello" and can't be updated using update-method, because it uses `self.text`'s setter, and it calls `formatted_text`. `fmt` works for adding prefixes and postfixes, but it breaks for multiple variables. We could use `self._text` for storing tuple of numbers (cpu/memory data) but that would look really hacky and weird.

And clock-widget also uses `format` instead of `fmt`.

I could add `fmt` as a flag for the constructor, but store the format in `format`. That way it doesn't crash, and it doesn't break previous configs.

libqtile/widget/memory.py
```diff
@@ -39,27 +39,33 @@ class Memory(base.ThreadedPollText):
...
+        ("update_interval", 1.0, "Update interval for the Memory"),
     ]
 
     def __init__(self, **config):
-        super().__init__(**config)
+        base.ThreadedPollText.__init__(self, **config)
```
pawamoy: What is the reason for this change? Isn't it semantically the same?

pawamoy:
Hmmm that's interesting information. Maybe there's some logic to fix in `libqtile.widget.base._TextBox` then (not implying you should change your PR). Thanks for the details.

libqtile/widget/memory.py
b10011: It's not, because `super().__init__` refers to `_TextBox.__init__` (iirc) and we need to init via `ThreadedPollText.__init__` or it crashes. I tried both changes individually and both crashes the WM.

pawamoy:
I take back what I said before about the consistent use of `fmt`: only the widgets gmail, memory,  and pomodoro are using `fmt`, while all the other widgets use `format`. My bad :sweat_smile:!

b10011:
@pawamoy No problem!

libqtile/widget/memory.py
pawamoy: Well, I'm a bit lost here.
- `ThreadedPollText` does not redefine `__init__`, so it must use its parent one, `InLoopPollText.__init__`, which itself calls `_TextBox.__init__`
- `super()` should do the same because it goes up one level, using the direct parent's `__init__` which is `ThreadedPollText.__init__`, etc.

But maybe there's things happening in `super()` that I don't know.

Anyway, what do you mean by both changes crash the WM? Which changes?

b10011: Me too, it raised some error message earlier but now it worked just fine. I added a commit to change it back to `super().__init__`

I meant that changing `format` back to `fmt` crashes and using `super().__init__` crashes bot individually. But I was wrong about the super.

pawamoy: OK thanks for the explanation :slightly_smiling_face:

b10011:
I was checking other issues and noticed that this should also fix #1435

libqtile/widget/memory.py
```diff
@@ -39,27 +39,33 @@ class Memory(base.ThreadedPollText):
...
+        ("update_interval", 1.0, "Update interval for the Memory"),
     ]
 
     def __init__(self, **config):
         super().__init__(**config)
         self.add_defaults(Memory.defaults)
 
+    def tick(self):
+        self.update(self.poll())
+        return self.update_interval
```
flacjacket: What's the rationale for adding this? This seems to be changing the widget to behave like the in-loop poll text base class.

libqtile/widget/memory.py
flacjacket: Hmm, this is interesting. If the old method was broken, doing an api breaking change should be fine to fix it.

libqtile/widget/memory.py
b10011: If I remember correctly, without that it updated really randomly, occasionally it took a bit over second, sometimes over 3 seconds. With that it takes more consistently a bit over 1 second to update.

b10011:
@flacjacket The structure, in general, was looking good but the inconsistent usage of the classes was causing problems.
I'm making custom prompt for myself with support for using xdg desktop entries and I noticed the Prompt is also quite cluttered and should support multiple completers at the same time.
Most of the widgets are quite small and I'd be willing to re-code/fix a bunch of them if there will be a redesign for base class structure and guidelines that must be followed to get the widgets merged.

libqtile/widget/memory.py
flacjacket: I think the reason that is the case is that it will run the update in the main thread, if that is desired it should base class the in-loop poll text widget, rather than subclassing the threaded poll text widget, but only if we know the function calls in the poll function are fast and not going to block.

libqtile/widget/memory.py
ramnes: It might be broken since the implementation of `fmt` in the base class. I'd be in favor of moving it to `format` so that it's consistent with other widgets, `fmt` being the "overall" format of the output and `format` being the internal formatting of the widget.

flacjacket:
These changes seem good, thanks for the fixes! Sorry for the delay!

PR #1485, opened 2019-10-26
Fixes to comments in xcbq, related to now irrelevant xpyb

Darion:
Hi! I found old comments in `xcbq` about `xpyb`, replaced to `xcffib` long time ago. I made some fixes to actualize them.


flacjacket:
Thanks for the updates!

PR #1487, opened 2019-11-02
let maildir widget display inbox grand total

m-col:
I often found myself checking my maildir status on the bar and caring more about the total unread emails I had than which inboxes they were in. I think this must be a work/life balance issue and I shouldn't check emails when only my work inbox has mail.

In any case, to enable this troubling behaviour, I made this change to the maildir widget. Passing `total=True` will sum the number of emails in all specified subfolders and the widget will display only that number. The label assigned to the first subfolder is used as the label.


flacjacket:
This looks good, thanks!

Issue #1483, opened 2019-10-22
QTile Bad Config


dk0104:
QTile Config error with default config.py .Config file location ~/.config/qtile/config.py
QTile installed  on fedora 30 dnf. QTile version : 0.13.0
qtile.log:
```
2019-10-22 23:10:55,246 WARNING libqtile log_utils.py:init_log():L120 Starting logging for Qtile
2019-10-22 23:10:55,260 ERROR libqtile confreader.py:from_file():L87 Could not import config file '/home/dk/.config/qtile/config.py'
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/confreader.py", line 83, in from_file
    config = __import__(os.path.basename(path)[:-3])
  File "/home/dk/.config/qtile/config.py", line 2, in <module>
    from libqtile.lazy import lazy
ModuleNotFoundError: No module named 'libqtile.lazy'
2019-10-22 23:10:55,260 ERROR libqtile qtile.py:make_qtile():L109 Error while reading config file (Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/confreader.py", line 83, in from_file
    config = __import__(os.path.basename(path)[:-3])
  File "/home/dk/.config/qtile/config.py", line 2, in <module>
    from libqtile.lazy import lazy
ModuleNotFoundError: No module named 'libqtile.lazy'
)
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/confreader.py", line 83, in from_file
    config = __import__(os.path.basename(path)[:-3])
  File "/home/dk/.config/qtile/config.py", line 2, in <module>
    from libqtile.lazy import lazy
ModuleNotFoundError: No module named 'libqtile.lazy'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/scripts/qtile.py", line 107, in make_qtile
    config = confreader.Config.from_file(options.configfile)
  File "/usr/lib/python3.7/site-packages/libqtile/confreader.py", line 89, in from_file
    raise ConfigError(tb)
libqtile.confreader.ConfigError: Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/confreader.py", line 83, in from_file
    config = __import__(os.path.basename(path)[:-3])
  File "/home/dk/.config/qtile/config.py", line 2, in <module>
    from libqtile.lazy import lazy
[...]

Issue #1479, opened 2019-10-17
config Err when qtile is installed differently ?


ReductoAdAbsurdum:
when I install qtile using: 
+++
git clone git://github.com/qtile/qtile.git
cd qtile
pip3 install .
+++
everything works fine, but when I install it using:
+++
pip3 install qtile
+++
I get a config Err ...why is that ?

zordsdavini:
Try `pip show qtile` to get version. From pip you installed 0.14.2 I giess. By github ypu installed development version

zordsdavini:
It looks like iy is time to bump version 😀

ramnes:
As @zordsdavini said, `pip3 install qtile` will install the latest release, while cloning the repository and executing `pip3 install .` will install the current development version.

If you want to debug your configuration for the version on which it doesn't work, you can execute `python3 config.py` and/or take a look at the `~/.local/share/qtile/qtile.log` file to hunt any error.

Feel free to ping me here or open another issue if needed. :wink:

ReductoAdAbsurdum:
So the config.py for a version is not compatible with another version ? no backwards compatibility ?

ramnes:
We document the most changes in the [CHANGELOG](https://github.com/qtile/qtile/blob/master/CHANGELOG), especially the breaking ones. We try to not change the API as much as possible (i.e. how a normal config file would look) but the internals might change more regularly, so if you're using them you must keep an eye to the changes.

See this part of the CHANGELOG for an example of what kind of things we allow ourselves:

> Many internal things were renamed from camel case to snake case. If your config uses main(), or any lazy.function() invocations that interact directly with the qtile object, you may need to forward port them. Also note that we do *not* consider the qtile object to be a stable api, so you will need to continue forward porting these things for future refactorings (for wayland, etc.). A better approach may be to add an upstream API for what you want to do ;)

Issue #1273, opened 2019-01-29
config error


cr1ogen:
I follow the instructions and give me this error. Installed with the command sudo python3 setup.py install, if i want to install with sudo python setup.py install give me a  install error 

```
2019-01-28 20:58:45,173 libqtile confreader.py:from_file():L102  Could not import config file '/home/cr1ogen/.config/qtile/config.py'
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/confreader.py", line 98, in from_file
    config = __import__(os.path.basename(path)[:-3])
ModuleNotFoundError: No module named 'config'
2019-01-28 20:58:45,173 libqtile qtile.py:make_qtile():L111  Error while reading config file (Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/confreader.py", line 98, in from_file
    config = __import__(os.path.basename(path)[:-3])
ModuleNotFoundError: No module named 'config'
)
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/confreader.py", line 98, in from_file
    config = __import__(os.path.basename(path)[:-3])
ModuleNotFoundError: No module named 'config'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/scripts/qtile.py", line 109, in make_qtile
    config = confreader.Config.from_file(kore, options.configfile)
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/confreader.py", line 104, in from_file
    raise ConfigError(tb)
libqtile.confreader.ConfigError: Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/qtile-0.13.0-py3.7.egg/libqtile/confreader.py", line 98, in from_file
    config = __import__(os.path.basename(path)[:-3])
ModuleNotFoundError: No module named 'config'
```

flacjacket:
This looks like a problem from running qtile, not installing it, right?

It looks like it is unable to find the `config.py` file, but it should then fall back to a default config. Is there any more context for this error? What is the command you used to get this error?

cr1ogen:
[...]

PR #1489, opened 2019-11-12
Create a initial config for user if not existing yet.

whatacold:
This tries to create an initial config.py for user if not existing yet, which is an exact copy of the default_config.py, so that it will lower the chance of incompatibility between the source code and config, which has been seen in #1483 #1479 #1273 .

And `qtile.1` should be updated too, but I haven't figured out how to generate it from the .rst file :(


ramnes:
Let's go!

PR #1491, opened 2019-11-14
Toggle group if setting a group twice.

whatacold:
Here is a quick fix for this use case: quickly switch to a group and take a look, then switch it back to the previous group.

Assume the current group is group `a`, and I want to take a quick look at group `b`. There are two way with shortcut keys of the default config:
1. Clicking on label b of the groupbox, and then clicking it again.

    One who is used to use mouse to switch group will benefit a bit more from this patch, as he/she doesn't need to know the previous group (hence its shortcut key), and just clicks on the label twice.

2. Hitting `mod+b` twice.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -290,6 +290,7 @@ class Screen(CommandObject):
     def set_group(self, new_group, save_prev=True):
         """Put group on this screen"""
         if new_group.screen == self:
+            self.cmd_toggle_group()
             return
 
         if save_prev:
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -287,10 +287,6 @@ class Screen(CommandObject):
     def get_rect(self):
         return ScreenRect(self.dx, self.dy, self.dwidth, self.dheight)
 
-    @property
-    def current_group(self):
-        return self.group
-
     def set_group(self, new_group, save_prev=True):
         """Put group on this screen"""
         if new_group is None:
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -289,15 +289,15 @@ class Screen(CommandObject):
 
     def set_group(self, new_group, save_prev=True):
         """Put group on this screen"""
+        if new_group is None:
+            return
+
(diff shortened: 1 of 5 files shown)


whatacold:
It seems that there are a few false-positive cases can cause toggling group, such as clicking a link in Chrome, which need further investigation.

whatacold:
I record a screencast gif to demonstrate what it looks like:
1. First `Super + 2` switches to group 2, and second `Super + 2` key bounces back to group 1.
2. Likely, first clicking on group 2 switches to group 2, and second clicking bounces back.

setting-group-twice

pawamoy:
Nice! What software did you use to record with the keystrokes?

libqtile/config.py
```diff
@@ -287,17 +287,21 @@ def dheight(self):
     def get_rect(self):
         return ScreenRect(self.dx, self.dy, self.dwidth, self.dheight)
 
+    @property
+    def current_group(self):
+        return self.group
```
ramnes: Why this property?

ramnes:
LGTM apart from that property that seems to add a small "complexity" for nothing.

libqtile/config.py
whatacold: I think it's better not to expose the member `self.group` directly, just like [current_group()](https://github.com/qtile/qtile/blob/master/libqtile/core/manager.py#L525) in `manager.py`.

whatacold:
> Nice! What software did you use to record with the keystrokes?

It's [screenkey](https://gitlab.com/wavexx/screenkey), which seems to be the only solution in Google as I searched it.

libqtile/config.py
ramnes: Why? `current_group` exists in the manager because the manager doesn't have a direct relation with groups. It's not the case here, a screen has to have a group.

whatacold: You're right, I am removing it now.

ramnes:
This is in! :tada:

PR #1495, opened 2019-11-30
Allow setting wallpaper per Screen in config

m-col:
This commit lets you have something like this in your config to paint your screens with a wallpaper:

```
screens = [
    Screen(wallpaper={'path': '/path/to/image.png'}),
    Screen(wallpaper={'path': '/path/to/another_image.png'}),
]
```

It works with multiple screens and different images, using xcffib, cairocffi and the Qtile object's own X connection.

The wallpaper kwarg is a dict for the purpose of specifying how to paint the wallpaper. By default, the image is positioned at the corresponding screen's origin and retains its dimensions. Alternatively, passing 'fill' or 'stretch' will make the image fill or stretch to fit the screen. I.e.:

```
wallpaper = {
    'path': '/path/to/image.png',
    'option': 'fill',
}
screens = [
    Screen(wallpaper=wallpaper),
    Screen(wallpaper=wallpaper),
]
```

It has been working well for me with single and double screens and removes the need for another tool installed just for setting the wallpaper (like feh). I understand this may be considered out of scope for Qtile, but I figured if Qtile has its own bar then painting the root window might not be too far out!

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -31,6 +31,7 @@ import warnings
 from . import configurable
 from . import hook
 from . import utils
+from .log_utils import logger
 from libqtile.command_object import CommandObject
 
 
@@ -218,19 +219,28 @@ class Screen(CommandObject):
     and ``height`` aren't specified usually unless you are using 'fake
     screens'.
 
+    The screen's wallpaper can be set by passing in a dict as the ``wallpaper``
+    kwarg, which must contain a 'path' key to specify the image to use, and can
+    optionally contain an 'option' key containing the string 'fill' or
+    'stretch' to specify how to fit the image onto the screen. 'fill' will
+    centre the image on the screen and fill it without warping it, 'stretch'
+    will warp the image to fit all of it into the screen. By default the image
+    will be placed at the screens origin and retain its own dimensions.
+
     Parameters
     ==========
     top: Gap/Bar object, or None.
     bottom: Gap/Bar object, or None.
     left: Gap/Bar object, or None.
     right: Gap/Bar object, or None.
+    wallpaper: Dict, or None.
     x : int or None
     y : int or None
     width : int or None
     height : int or None
     """
     def __init__(self, top=None, bottom=None, left=None, right=None,
-                 x=None, y=None, width=None, height=None):
+                 wallpaper=None, x=None, y=None, width=None, height=None):
         self.group = None
         self.previous_group = None
 
@@ -238,6 +248,7 @@ class Screen(CommandObject):
         self.bottom = bottom
         self.left = left
         self.right = right
+        self.wallpaper = wallpaper
         self.qtile = None
         self.index = None
         # x position of upper left corner can be > 0
@@ -257,6 +268,14 @@ class Screen(CommandObject):
         self.set_group(group)
         for i in self.gaps:
             i._configure(qtile, self)
+        if self.wallpaper:
+            try:
+                self.paint(**self.wallpaper)
+            except TypeError:
+                logger.exception("Error in wallpaper configuration.")
+
+    def paint(self, path, option=None):
+        self.qtile.paint_screen(self, path, option)
 
     @property
     def gaps(self):
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -48,6 +48,9 @@ import xcffib.randr
 import xcffib.xinerama
 import xcffib.xproto
 
+import cairocffi
+import cairocffi.xcb
+
 from libqtile import xkeysyms
 from libqtile.log_utils import logger
 from .xcursors import Cursors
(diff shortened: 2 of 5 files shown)


m-col:
I've tried to follow along with the changes to how Qtile manages the backend and x11 and updated this accordingly. The painting is now done in the core's Connection instance, so the Qtile manager simply passes a wallpaper configuration dict to self.core.conn.paint_root, which I hope would also be sufficient for other backends.

flacjacket:
This is very interesting. Have you taken a look at some of the existing wallpaper widgets? I'm wondering if this is something that could replace some of that functionality. By plugging into the widget framework could give some additional features, such as polling intervals that would allow the wallpaper to be changed easily. This is really cool how you have worked this out, I'm just thinking if we can keep some of the core functionality as clean as possible, that will help as we refactor towards wayland support.

ramnes:
It would have sense to me that we implement a `Wallpaper` class in `config.py`, just like screens, which would be handled on one side by the wallpaper widget or directly through the the config, and on the other side by the chosen core.

m-col:
I just pushed some changes following the first reply. How it is organised now (which I think is more clean):

- The Qtile manager has a `paint_screen` method which just passes inputs to the core. This is the clean interface that anything else can use. It just requires the Screen instance of interest and a path to an image.
- `Screen` in config.py has a `paint` method, requiring a path to an image, and passes this and itself to qtile.paint_screen. It makes use of this in `Screen._configure`
- xcbq.py `Connection.paint_screen` does the heavy lifting for the X11 backend. It's now more generalised and clean than initially. Again, all it needs is a `Screen` and image path.
- The Wallpaper widget has been updated to use this interface

Previously, the Wallpaper widget by default used feh, but the default config for it had `wallpaper_command` as `None`, and if it was `None`, then it used feh. Now `wallpaper_command` defaults to the same feh command in the defaults dict, so the widget's behaviour is unchanged for most people. However, anyone who used the 'one_screen' setting would have to update the `wallpaper_command` accordingly i.e. to `['feh', '--bg-fill', '--no-xinerama']`. If `wallpaper_command` is set to `None`, then the widget will use the native interface i.e. `self.qtile.paint_screen(self.bar.screen, cur_image, self.option)` when clicked, and also upon `Wallpaper._configure` only if the Screen did not have the wallpaper attribute set (otherwise two things try to set the wallpaper upon Qtile launch).

Any thoughts on how this could be improved or organised differently? I'm not sure what the benefits of a Wallpaper class would be, other than serving as a slightly different way than a dict to configure the wallpaper (though I guess a class is a bit nicer than a dict!)

libqtile/config.py
```diff
@@ -257,6 +267,14 @@ def _configure(self, qtile, index, x, y, width, height, group):
         self.set_group(group)
         for i in self.gaps:
             i._configure(qtile, self)
+        if self.wallpaper:
+            self.paint(
+                self.wallpaper.get('path'),
+                self.wallpaper.get('option', None),
+            )
```
ramnes: why not just

```python
self.paint(**self.wallpaper)
```
?

ramnes:
LGTM that way!

libqtile/config.py
m-col: Yep that is nicer. Updated!

m-col:
Do I need to do something about the failing Travis check?

ramnes:
It's definitely related to your changes so I'd rather say yes, but then I'm not sure how to fix this since I'm not used to MyPy and stubs.

m-col:
I found a small bug in config.py. If the dict passed to Screen didn't contain only `path` and an optional `option` item then Screen.paint would trigger a mysterious exception. I've imported the logger to config.py and caught it with an error message so it will continue. I also rebased - is this necessary for PRs?

Looking a bit into mypy and stubs, it seems that to pass the check a stub file for cairocffi.xcb needs to be added into the stubs folder. I'll look into this.

m-col:
I've sorted the failed check.

I had to add `self.conn.core.SetCloseDownMode(xcffib.xproto.CloseDown.RetainPermanent)` to the initialisation of xcbq.Connection, so that any pixmaps painted to the root window aren't cleaned up when the Connection disconnects during a Qtile restart. As far as I can tell, this has not and should not have any other side effects.

It looks like that should be it ready to go.

m-col:
I spoke too soon: setting the close down mode to retain permanent makes windows that are not in the current workspace when Qtile is restarted get lost. Perhaps the wallpaper logic should use its own X connection rather than the core connection, in which case a Painter class could be created to handle that rather than touching xcbq.Connection. Any thoughts?

m-col:
OK, think it's sorted. I'm no longer getting any more issues.

m-col:
This has been working perfectly for me the for past couple months, at least within my usage habits (laptop, laptop + monitor, double monitors). Is there anything stopping this from being merged?

tych0:
Looks good to me, thanks! Sorry for the delay.

ramnes:
Mhhh I'm too late but I don't like the dictionary argument, can't we just ask for two arguments directly?

m-col:
Yeah I also wasn't sure if it was the cleanest. We could do something like this instead:
```
screens = [
    Screen(wallpaper='/path/to/image.png',),
    Screen(wallpaper='/path/to/anotherimage.png', option='fill'),
]
```
Though I don't think the word 'option' is very clear. Maybe 'paint_method'? Any other ideas?

ramnes:
`wallpaper_path` and `wallpaper_mode`, maybe?

m-col:
Done in #1551. Thanks for accepting my changes!

PR #1498, opened 2019-12-12
add support for checking updates via yay on Arch

daroczig:
`yay` is a `pacman` wrapper that can also look up updates on AUR


ramnes:
Thanks!

Issue #1482, opened 2019-10-21
Fullscreen-off windows are moved to the tail of the list of windows


whatacold:
It's a bit frustrating that toggling fullscreen of windows will mess up the window layout.
For example, it can be reproduced with a tile layout(configured by `layout.Tile(add_after_last=True)`):
1. Open 3 windows, and they looks like:

        |   | B
        | A |--
        |   | C

2. Toggle A into fullscreen, and toggle it off, then you will find A is moved to the tail of the window list, which makes them look like:

        |   | C
        | B |--
        |   | A

The reason for that, is that the window is removed from the list when it is toggled to fullscreen, as in [_Group.remove()](https://github.com/qtile/qtile/blob/master/libqtile/group.py#L274),

    def remove(self, win, force=False):
        ...
        if win.floating:
            ...
        else:
            for i in self.layouts:
                if i is self.layout:
                    nextfocus = i.remove(win)
                else:
                    i.remove(win)

I think maybe the window is better marked hidden, but is still kept in the list. But I haven't figured out how to fix it properly:(

ramnes:
Can confirm. Super annoying.

ramnes:
Related to #1114 #1201 #1202 and #1313. I think we can confirm that it happens on pretty much all layouts at this point. Let's find a global way to address this.

gmjain:
Aren't we technically cheating by converting the window to floating when trying to make it fullscreen? There are some side-effects to that too. Like, a fullscreen floating window can now be dragged/resized using the click/drag bindings which is not (or shouldn't be) a property of a strictly fullscreen window.

I agree with @whatacold that we shouldn't remove it from the layout client list. I think a potential solution can be to not make it floating but just change its dimensions for rendering, overlapping the other windows in the layout.

jwijenbergh:
Hey I'm still having this issue myself. Is the plan to move forward with #1313 or #1500? I am willing to help with these prs but I have never contributed to qtile myself.

ramnes:
I don't know, I don't feel very enthusiastic about the two RPs; maybe just because they've been abandoned for a while, or maybe because I'm hopeful that we can do something simpler. So feel free to choose one, or to work on your own take.

flexagoon:
Can confirm, the issue is very annoying

VoltaireNoir:
[...]

Issue #1444, opened 2019-09-09
An actual Floating layout?


tamis-laan:
Hey is it possible to have an actual floating layout, as I understand is Floating is not really a layout but configures the floating layout state. 

However I would like to be able to switch to an actual Floating layout where all windows are floating and the position and size of the windows are remembered when switching layouts back and forth. 

Is this possible?

ramnes:
No, it's not possible as of today. You can already add an instance of `libqtile.layout.Floating` in your configuration `layouts` variable, but as you noticed, any modification of a window's size or position will result in marking that window as floating, even when you switch to another layout.

What you describe would be a really nice feature though, and I would love to see a pull request on this.

PR #1500, opened 2019-12-14
Fix windows disorder after toggling fullscreen

whatacold:
Wow, so many fails need to be fixed.

This PR tries to fix issue #1482 , which is still WIP. The rational is to make `_ClientList` only expose non-floating windows to its users(Floating layout uses a plain list), so as to keep the floating/fullscreen ones in it.

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -106,7 +106,7 @@ class Floating(Layout):
                 return True
         return False
 
-    def interested_in_client(self, client):
+    def _interested_in_client(self, client):
         '''Test if a client is interested in.'''
         floating = client.floating
         if (self._floating_clients_only and floating) or \
@@ -118,7 +118,7 @@ class Floating(Layout):
         """Find all clients belonging to a given group"""
         clients = []
         for c in self._clients:
-            if (group is None or c.group is group) and self.interested_in_client(c):
+            if (group is None or c.group is group) and self._interested_in_client(c):
                 clients.append(c)
         return clients
 
@@ -181,7 +181,7 @@ class Floating(Layout):
             return clients[idx - 1]
 
     def focus(self, client):
-        if self.interested_in_client(client):
+        if self._interested_in_client(client):
             self.focused = client
 
     def blur(self):
@@ -253,7 +253,7 @@ class Floating(Layout):
 
     def add(self, client):
         self._clients.append(client)
-        if self.interested_in_client(client):
+        if self._interested_in_client(client):
             self.focused = client
 
     def remove(self, client):
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -289,7 +289,6 @@ class FloatingTile(Floating):
         ("name", "floatingtile", "Name of this layout.")
     ]
 
-
     def __init__(self, **config):
         Floating.__init__(self, **config)
         self._floating_clients_only = False
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -157,10 +157,10 @@ class Floating(Layout):
             return clients[0]
 
     def focus_next(self, win):
-        if win not in self._clients or win.group is None:
+        clients = self.find_clients(win.group)
+        if win not in clients or win.group is None:
             return
 
-        clients = self.find_clients(win.group)
         idx = clients.index(win)
         if len(clients) > idx + 1:
             return clients[idx + 1]
@@ -172,10 +172,10 @@ class Floating(Layout):
             return clients[-1]
 
     def focus_previous(self, win):
-        if win not in self._clients or win.group is None:
+        clients = self.find_clients(win.group)
+        if win not in clients or win.group is None:
             return
 
-        clients = self.find_clients(win.group)
         idx = clients.index(win)
         if idx > 0:
             return clients[idx - 1]
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -89,7 +89,7 @@ class Floating(Layout):
         the screen.
         """
         Layout.__init__(self, **config)
-        self.clients = []
+        self._clients = []
         self.focused = None
         self.group = None
         self.float_rules = float_rules or DEFAULT_FLOAT_RULES
@@ -117,7 +117,7 @@ class Floating(Layout):
     def find_clients(self, group):
         """Find all clients belonging to a given group"""
         clients = []
-        for c in self.clients:
+        for c in self._clients:
             if (group is None or c.group is group) and self.interested_in_client(c):
                 clients.append(c)
         return clients
@@ -157,7 +157,7 @@ class Floating(Layout):
             return clients[0]
 
     def focus_next(self, win):
-        if win not in self.clients or win.group is None:
+        if win not in self._clients or win.group is None:
             return
 
         clients = self.find_clients(win.group)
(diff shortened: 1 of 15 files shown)


whatacold:
There are still some test failures to be fixed, let me know what you guys think :)

ramnes:
See #1313 for an alternative route to fix that problem.

whatacold:
I wasn't aware there is already a proposing solution, thanks for referring to it.

It was quite a different and simple solution(on the client side vs on the layout side), but I think there are some downsides of it:
1. The restored window order may be different if they are unfloating in a different order than they are floating.

   Assume there are 5 windows `[0, 1, 2, 3, 4]` are in `self.clients`, and client 4, 3, 2 are floating and then unfloating, in that order. Then the resulted windows will look like `[0, 1, 2, 4, 3]`, which is slightly different.

   Check out this snippet for it:

        >>> l = [0, 1]
        >>> for i in (4, 3, 2): l.insert(i, i)
        ... 
        >>> l
        [0, 1, 2, 4, 3]

2. The state can be hard to represent.

   Although it's easy to use the index to store the state of clients for simple layouts using lists (`_ClientList`) as their container, it's not easy for others. For example, I can't think of a straight way to represent the state for Bsp, as it organizes all the clients at leaves of a full binary tree.

3. Every layout needs to take care of the states of the client when it is floated/unfloated.

If I understand it correctly, these downsides are due to that, the clients are not the right ones responsible for maintaining the order, the containers(layouts) are.

So in this PR, I kept those floating clients still within their containers, but not to show them. And it's more general as every layout using `_ClientList` as its clients container can work as before, only other layouts such as Bsp/TreeTab need to extra work.

That said, I believe there are downsides that I can't think of too. Among others, `_ClientClient.non_floating_clients`'s performance may be bad, which should be fixed if it's necessarily that bad, although currently, it's acceptable on my box.

whatacold:
It's almost done, and most tests are passed except one failure with Python 3.5, I haven't figure out why it's different yet. And there are also some pep8 and mypy failures.

Besides, I will try to add more tests of fullscreen toggling and add more docs of various layouts as I understand them. And I will recap these modifications so as to help make the code review process a bit easier.

whatacold:
A summary of these changes I've made:
1. group.py

   - Adding/removing a client manipulates both tiled layouts and the floating layout.
   - Toggling floating of clients no long adds or removes the client from its layouts, but just call stubs functions (floating_on/floating_off) instead.
2. base.py: let `_ClientList` hide floating clients

   This makes most layouts using `_ClientList` support floating correctly. For Bsp/TreeTab/Floating layouts, they need extra work.

3. The Bsp layout: make the floating clients never have the chance of appending newly children, and let its sibling take all the screen space if a client is floating.

4. The TreeTab layout: let the first child of preorder tree traverse take the place of a floating client virtually.

5. The Floating layout: a dedicated `FloatingTile` layout is introduced for tiled layout usage since it's impossible to use `Floating` as a tiled layout and a floating layout, because floating clients should not be shown in a tiled layout right now.

pawamoy:
Do I understand correctly if I say that this PR is a step towards supporting floating windows in any layout, without moving them to the special FloatingLayout?

I'm gonna pull this and try it against the screenshot generator to see the real impact 🙂

whatacold:
> Do I understand correctly if I say that this PR is a step towards supporting floating windows in any layout, without moving them to the special FloatingLayout?

No, the `Floating` layout is still needed for containing floating windows.

whatacold:
There are some screencasts recorded by `byzanz-record`.

1. Tile

   tile

2. Bsp

   bsp

3. TreeTab

   treetab

whatacold:
Alright, all tests are passed finally. :) I think it's OK for code review.

libqtile/layout/floating.py
```diff
@@ -268,3 +278,19 @@ def cmd_next(self):
     def cmd_previous(self):
         # This can't ever be called, but implement the abstract method
         pass
+
+
+class FloatingTile(Floating):
+    '''Floating layout for tile layouts config: the `layouts` symbol in config.py.
+
+    It is almost exact with Floating, except that it's intended to be use a tile layout.'''
```
tych0: I'm not clear on what this actually does or why we need it. Is it basically another layout on top of the existing layout?

tych0:
Some of the stuff here looks pretty nice. Do you mind if I cherry pick some of the commits out of here? I generally don't understand the FloatingTile approach or why it's relevant here, but the added tests and stuff seem useful on their own.

whatacold:
> Do you mind if I cherry pick some of the commits out of here?

That's ok.

> I generally don't understand the FloatingTile approach or why it's relevant here
I will see if it can be improved if I have the time.

tych0:
Ok, I've started looking at cherry picking parts of this. But I found a few other oddities in our test suite that I'm going to try and get rid of first :)

tych0:
Ok, so the independent stuff is upstream, and I've rebased, fixed up, and squashed the rest here: https://github.com/tych0/qtile/commit/530c5135339bfd1902b08b6e8906572871eaf1c5

I'd still like to drop FloatingTile if possible, I don't really see what it does. But when I did drop it, the tests didn't pass. So I need to read more thoroughly it seems.

whatacold:
@tych0 Sorry that `FloatingTile` thing confused you.

The reason for it comes from that the original `Floating` has two responsibilities(we can see that from `default_config.py`):
1. Typically it's used to contain floating windows for layouts like tile, bsp, etc., as in `floating_layout = layout.Floating(float_rules=[...])`.
2. It can also be used as a regular layout, as in `layouts = [layout.Max(), layout.Floating, ...]`

So, for the groups using the `Floating` layout, it needs a way to separate these two roles apart, so I introduced `_floating_clients_only` and `_interested_in_client` to do that job.

whatacold:
@tych0 Sorry that `FloatingTile` thing confused you.

The reason for it comes from that the original `Floating` has two responsibilities(we can see that from `default_config.py`):
1. Typically it's used to contain floating windows for layouts like tile, bsp, etc., as in `floating_layout = layout.Floating(float_rules=[...])`.
2. It can also be used as a regular layout, as in `layouts = [layout.Max(), layout.Floating, ...]`

So, for the groups using the `Floating` layout, it needs a way to separate these two roles apart, so I introduced `_floating_clients_only` and `_interested_in_client` to do that job.

tych0:
@whatacold no worries, thanks for the explanation! (and welcome back :)

Do people actually use it how you've described in #2? If so, then I agree this distinction makes sense. I guess I've never thought about using it that way.

A quick look through qtile-examples suggests they don't, but I'm open to keeping it if someone says they do?

ramnes:
What if we just get rid of the floating layout completely? It was never intended to be a real layout, and it only brings confusion. And a second floating layout would bring even more of it...

tych0:
Yeah, that was my worry is more confusion.

Getting rid of it would be interesting; basically just teach layouts to
ignore windows with window.floating == True?

ramnes:
In the grand scheme of things, I really think that layouts should only be given a list of clients to place, and should do nothing else, so they wouldn't even have to worry about the client being floating or not, that would be handled behind the scenes. Then we could also move the window selection logic outside layouts for most layouts, only based on clients coordinates; but that's another thing. Teaching layouts to ignore floating layouts so that we can remove the floating layout might be a good first step, I don't know.

whatacold:
> Do people actually use it how you've described in #2? If so, then I agree this distinction makes sense. I guess I've never thought about using it that way.

I don't use it as a layout personally, but I did see a feature request for a floating layout as in #1444 .

PR #1501, opened 2019-12-19
add ability to use bits as unit in net widget

hlmtre:
I wanted the ability to view the `net.py` output in bits instead of bytes, so here it is.


flacjacket:
This seems good. I'm wondering if it's better to have an option for units, as I remember seeing some discussion of binary vs decimal units (eg KB vs KiB), having units wound disambiguate that.

flacjacket:
Thanks for the changes!

PR #1508, opened 2019-12-27
Move X11 related functionality into the xcore

flacjacket:
Shift some of the X related functionality into the xcore module from the
top level manager.  There isn't yet a great interface around some of
this functionality, it is simply moving what exists currently into the
xcore. As a new backend is brought up, this can be worked out into what
is needed.


ramnes:
Looks great!

PR #1516, opened 2020-01-08
Do not spam logs when the battery device is temporarily inaccessible

ramnes:
With the battery widget enabled, my logs are full of this error:

```
2020-01-08 10:47:52,577 ERROR libqtile base.py:worker():L478 problem polling to update widget battery
  File "qtile/libqtile/widget/base.py", line 474, in worker
    text = self.poll()
  File "qtile/libqtile/widget/battery.py", line 351, in poll
    status = self._battery.update_status()
  File "qtile/libqtile/widget/battery.py", line 276, in update_status
    power_str, power_unit = self._get_param('power_now_file')
  File "qtile/libqtile/widget/battery.py", line 255, in _get_param
    value = self._load_file(filename)
  File "qtile/libqtile/widget/battery.py", line 235, in _load_file
    return f.read().strip(), value_type
OSError: [Errno 19] No such device
```

I'm not sure why this happens, but from what I can see, when the battery is full and the AC takes over, it seems that the kernel stops answering for the battery device but keeps the files there for some reason.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -233,9 +233,14 @@ class _LinuxBattery(_Battery, configurable.Configurable):
         try:
             with open(path, 'r') as f:
                 return f.read().strip(), value_type
-        except FileNotFoundError:
-            logger.debug("Failed to get %s" % name)
-        return None
+        except OSError as e:
+            logger.debug("Failed to read '{}': {}".format(path, e))
+            if isinstance(e, FileNotFoundError):
+                # Let's try another file if this one doesn't exist
+                return None
+            # Do not fail if the file exists but we can not read it this time
+            # See https://github.com/qtile/qtile/pull/1516 for rationale
+            return "-1", "N/A"
 
     def _get_param(self, name) -> Tuple[str, str]:
         if name in self.filenames and self.filenames[name]:


libqtile/widget/battery.py
```diff
@@ -233,9 +233,11 @@ def _load_file(self, name) -> Optional[Tuple[str, str]]:
...
             with open(path, 'r') as f:
                 return f.read().strip(), value_type
-        except FileNotFoundError:
-            logger.debug("Failed to get %s" % name)
-        return None
+        except OSError as e:
+            logger.debug("Failed to read '{}': {}".format(path, e))
+            if isinstance(e, FileNotFoundError):
+                return None
+            return "-1", "N/A"
```
flacjacket: You could do this as:
```
except FileNotFoundError:
    logger.debug("Failed to get %s" % name)
    return None
except OSError:
    return "-1", "N/A"
```

ramnes: Yeah, I just wanted to log in any case without copy pasting the said line.

ramnes:
Yes, it's very possible that it comes from my setup as I'm using Gentoo with an unstable kernel that I heavily tweaked and a pretty recent Dell dock that powers my laptop through USB C, so let's say it's not very standard and proven technology. Still, I don't think it would hurt anyone to have this guardrail there.

flacjacket:
Yeah, that makes sense. The only thing I'd say is that some comment explaining why we are swallowing this type of error would be helpful.

ramnes:
FWIW I also have this error on another Dell laptop which doesn't use a dock this time.

Okay for comments, on their way!

Ahmeed2m:
This actually solves  my semi-related issue I mentioned [here](https://groups.google.com/forum/#!topic/qtile-dev/Al3sSdlnRhU)
A little background here, some laptops don't report current_now/power_now at all while discharging.
I think even one of Microsoft's surface tablets was like this and they said to prevent false estimated time left calculations (lol). Estimating time left doesn't even work in Windows for me. It's something in the firmware.
I found case similar to mine [here](https://bbs.archlinux.org/viewtopic.php?id=186446)
even acpi can't calculate estimated time left
> $ acpi -i
Battery 0: Discharging, 85%, rate information unavailable
Battery 0: design capacity 3282 mAh, last full capacity 3282 mAh = 100%


My laptop (HP) locks down the current_now file and changes it to binary unreadable by any program producing "[Errno 19] No such device" all over the logs too. but the widget wasn't working at all. It was like [this](https://i.imgur.com/imCq2rp.png) while charging, [this](https://i.imgur.com/DvU9X3f.png) while discharging.
Doen't matter having {hour:d}:{min:02d} in my Battery widget config or not it'll always break the widget while discharging.
I was half way into my implementation to solve it and I settled in a logic similar to yours !
Glad I checked the open PRs before making a potato out of myself :')
Using your implementation shows correct estimated while [charging](https://i.imgur.com/m7TSYSm.png) and garbage while [discharging](https://i.imgur.com/Sp3Op0g.png). Doesn't break at all

Will this PR merge ?

ramnes:
I'd like that it doesn't show garbage in that situation, but I guess you or me or anyone else can fix this in another PR. Thanks for trying out the branch and glad it helps! Merging now.

Issue #1515, opened 2020-01-08
Rounded corners


ramnes:
```
03h56 aircraftblues ok so I'm new and have been playing around with the config and widgets. does anybody know where i can find info for rounded window corners?
```

See [here](https://www.reddit.com/r/unixporn/comments/benebi/i3gaps_i3_rounded_corners/) for an example and links to different implementations in bspwm and i3.

Definitely some great rice I'd be glad to see in Qtile.

pawamoy:
(added a [draft](https://github.com/qtile/qtile/wiki/rices) of a "Rices" page in the wiki, would be great to show some)

m-col:
I have a functioning [round corners commit](https://github.com/m-col/qtile/tree/round_corners) that I've used for a couple months. I haven't made a PR for it because it doesn't have anti-aliasing and I don't really know how to add it. Perhaps the composite extension in xcffib could be used but that guess is as far as I've got. This could serve as a starting point for rounded corners :)

ramnes:
Awesome! I think it wouldn't hurt to get that into the codebase and then open another issue for the anti-aliasing, what do you think?

m-col:
It's funny this is brought up now... I've recently been thinking I should open an issue for advice regarding anti-aliasing. I've made the PR #1517

ramnes:
FWIW, it seems that picom can round the windows too, and that it works quite well with Qtile: https://www.reddit.com/r/UsabilityPorn/comments/k8s8ph/qtile_i_showed_off_my_bspwm_rice_earlier_so_heres/

Do you see any advantage of implementing this feature here rather than letting the compositor manage it?

m-col:
Yep - picom cuts into borders, whereas as we have control over the borders we
can draw borders that round on both the inside and outside edge of those
borders.

ramnes:
Are you sure? The borders aren't cut in the screenshot of Reddit post I mentioned, and picom doesn't seem to do borders, nor that fork.

m-col:
[...]

PR #1517, opened 2020-01-08
Add setting rounded corners to windows

m-col:
re: #1515

This adds a `round_corners` method to xcbq.Window in the X11 backend which does some xcffib pixmap manipulation to the window's border. The method is called by `window._Window.place`, if a corner_radius is passed for window placing.

This commit also updates the floating and columns layouts to use this. I only updated these layouts initially because they are the only ones I ever use (and are enough for testing), but figured I shouldn't update the others in case there is a nicer way than adding another setting to all layouts.


ramnes:
I don't think there's a better way to add this to all layouts as not all layouts will want this (i.e. Max or such). I guess we should add this everywhere it makes sense, until we refactor layouts so that layouts with borders all inherit from a same parent class, if we ever.

m-col:
This screenshot illustrates how it works:



It rounds the borders with a fixed radius both on the outside of the border, and the inside. Without clipping the inside too, we get [this problem](https://github.com/baskerville/bspwm/pull/856#discussion_r228241584). This means that it clips the contents of the window. In practice this isn't a problem unless for some reason you wanted a huge radius:



I'm not sure why you'd want that though. The second image also illustrates what it looks like without a border. Of course when the radius is less than around 16 pixels it looks good!

Currently I can see two changes that would be nice: anti-aliasing, and a non-fixed radius. See [here](https://github.com/baskerville/bspwm/pull/856#issuecomment-479091948) for an illustration of the latter.

I'd be happy to update other layouts which the corner_radius setting - just let me know which ones.

ramnes:
Just a gave a try at the branch!

It worked great at first sight, apart from a small bug on the border with Chrome windows: the border fills the empty space between the rounded corner and what the real window is.

Then I tried to switch to the Max layout and things got pretty bad. I ended up with all my windows being stuck to their current size, and restarting `qtile` didn't help, even with `corner_radius=0`, which wasn't applied. I had to kill and relaunch each window to get something usable.

Can you reproduce these two problems?

ramnes:
As for the layouts that should get the update, I think you can at least add these ones:

* bsp
* matrix
* ratiotile
* stack
* tile
* verticaltile
* xmonad

Maybe zoomy too, although I'm not sure how it should be handled there.

m-col:
I see the same with Chromium. The bspwm discussion for rounded corners mentions this too. [https://github.com/baskerville/bspwm/pull/856#issuecomment-431811207](This comment) suggests that it is a feature which can be disabled.

The problem with the Max layout is bizarre. I can't figure it out. If I set Max
as the only layout and the started a new X session, open a window and make it
floating, when making it tile again it stays roughly in position (top left
corner for xterm) but the shadow drawn by compton is in the correct position.
The round_corners method is never called so nothing should have changed. Any
ideas?

m-col:
I added the `corner_radius` setting to all layouts excluding Max. There is an issue with Tree and Max where switching to these layouts from one that has round corners results in correctly positioned windows that retain their old height, width and corner radius.

There also appears to be another (likely related) issue with something like this:
```
layouts = [
    layout.Columns(corner_radius=16),
    layout.Columns(),
]
```
Switching to the first layout rounds all the corners but switching to the second does not un-round them.

m-col:
Some progress... the problem I described in my last comment appears to be because the clip mask applied to the window persists, so Qtile replacing it somewhere else won't undo the round corners and will clip the window if it is larger than it was when the mask was applied, even with `corner_radius == 0`. The last commit 780bfcb adds `xcbq.Window.unround_corners` that is called when a rounded window is reconfigured with no corner radius. This draws it properly. The catch: the new mask itself persists, so the clipping will happen the next time it is moved. For everything to appear to work perfectly, `unround_corners` would have to be called _every_ time a window that was once rounded is reconfigured, which would be far from ideal.

So... does anybody know how to _remove_ a mask applied to a window, as seen [here](https://github.com/qtile/qtile/blob/780bfcbc532626f7e48820776fa85c0f72fa73b6/libqtile/backend/x11/xcbq.py#L884)?

m-col:
Bingo. Got it working perfectly. When placing rounded windows with a new `corner_radius` of 0 (e.g. changing layout), the bounding and clip masks are completely removed and so they return to normal behaviour.

I've squashed so all done for now on this end!

m-col:
I've been testing the last commit this week (just in case!) often going back and forth between layouts with rounded or pointed corners, and can report it's working brilliantly :)

ramnes:
This still happens, eheh:

chrome

Also, I happen to have all the windows that become rectangular again when switching back and forth between groups. Maybe this is because my two screens have different resolutions?

m-col:
I believe the problem with Chromium (and programs that use it, like Slack) is theirs, not ours, so there isn't much we could do about it.

Sorry I don't understand what you're describing. Do you mean that if you focus to a different Screen, the rounded windows on the first screen unround?

ramnes:
> I believe the problem with Chromium (and programs that use it, like Slack) is theirs, not ours, so there isn't much we could do about it.

Well, I did enable "Use system title bars and borders" in Chrome, as advised by the comment you linked, but the problem still happens. Are you sure it's on Chrome side?

> Sorry I don't understand what you're describing. Do you mean that if you focus to a different Screen, the rounded windows on the first screen unround?

Exactly.

m-col:
(Assuming Slack is behaving the same as Chromium) When I change to a rounded group its corners round initially and then the bounding region of the window is removed, but the clip region remains. I think the programs are setting/removing the bounding region themselves, and this is happening after `window.place` has been called. Does toggling that Chrome setting change anything at all? Searching online I only found references to that setting discussing the title bar, and I just noticed that the [following comment](https://github.com/baskerville/bspwm/pull/856#issuecomment-431816464) on the bspwm thread indicates that the setting didn't help.

I cannot reproduce the screen issue, but my monitors are the same resolution. Does having different resolution screens have any effects on how groups/windows are placed?

ramnes:
Most likely yes, as the placement is based on the screen width and height, unless I didn't understand your question?

Do some xrandr to lower the resolution of one of your screens and we'll know!

m-col:
Sorry what I meant was whether Qtile processed windows/groups differently if the screens were the same vs different resolutions. I don't think it does though. Using xrandr to change the resolution (TIL that you could do that!) doesn't change the behaviour for me, it still works fine.

However I noticed that if I have Slack and xterm open on the same screen and change group to round their corners while Slack is _not_ focussed then it rounds just fine. Only when I move the mouse over Slack does it undo the rounding. It rounds again when I click on xterm. I wonder if this is anything to do with [_NET_WM_SYNC_REQUEST](https://specifications.freedesktop.org/wm-spec/1.4/ar01s06.html). Do you see the same behaviour with Chrome?

m-col:
If `send_configure_notify` within `window.place` is commented out then it rounds as you'd expect, so looks like it is to do with the _NET_WM_SYNC_REQUEST. I'll have a look into the TODO at the top of `window.place` that, if resolved, would mean we only send configure_notify messages when we need to.

m-col:
Looked in the [Chromium source](https://github.com/chromium/chromium/blob/55f44515cd0b9e7739b434d1c62f4b7e321cd530/ui/base/x/x11_window.cc#L1640) and can confirm that they reset the clip mask manually if the "Use system title bars and borders" setting is enabled: 

> // If the window has system borders, the mask must be set to null (not a
> // rectangle), because several window managers (eg, KDE, XFCE, XMonad) will
> // not put borders on a window with a custom shape.

I think the only way to force it would be if Qtile was reparenting windows :/

tych0:
Can you add a test for this somewhere?

waltereikrem:
Hi guys! is this part of the master? like is it safe to test out even for a newbie lol?

ramnes:
As you can see, the PR is still opened, so it's not in master yet. That being said, it should be pretty safe to use m-col's branch if you want to try it.

m-col:
Hi @waltereikrem, I maintain this branch separately and rebase it every time it gets broken, and I think some recent changes broke it. I want to check it and fix it this weekend though so if it doesn't work now then it should do soon :) The reason the PR is on hold is because Chromium (and Chromium-based) windows don't  play nice and I haven't figured out a way to get them under control.

ramnes:
Is it? I thought we were waiting for the tests. I feel they are more important than Chromium support. :)

m-col:
Ah yep them too!

PR #1522, opened 2020-01-09
Added missing kwargs argument in EzKey class constructor

### LINKED ISSUE #1510: EzKey does not allow description
I think the [EzKey constructor](https://github.com/qtile/qtile/blob/master/libqtile/config.py#L155) does not allow a description (no `kwds` variable) although [Key constructor](https://github.com/qtile/qtile/blob/master/libqtile/config.py#L53) does. 

Edit: Why do you set the description within a dictionary instead of having a constructor argument for it?

Edit 2: Forgot my versions and stuff:

|    Item   |     Version    |
|:---------:|:--------------:|
|   Qtile (from official repositories)   |     0.14.2     |
| ArchLinux | 5.4.6-arch3-1 |

lulivi:
This fixes #1510 .

### Summary 

`kwargs` dictionary was missing in `EzKey` class constructor, so you couldn't add a description for `EzKey`'s.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -152,9 +152,9 @@ class EzConfig:
 
 
 class EzKey(EzConfig, Key):
-    def __init__(self, keydef, *commands):
+    def __init__(self, keydef, *commands, **kwargs):
         modkeys, key = self.parse(keydef)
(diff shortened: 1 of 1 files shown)


ramnes:
Thanks!

PR #1524, opened 2020-01-12
add display option to Net widget and introduce f-strings

stehessel:
I added a `display_format` option to the Net widget similar to how it is already implemented in the CheckUpdates widget. If the unit string contains one digit (i.e. 'b' and 'B'), an additional space is inserted to make sure the bar is not shifting back and forth. I also replaced string formats with f-strings and did minor formatting to make it more consistent. What do you think?

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -30,8 +30,11 @@ class Net(base.ThreadedPollText):
     """Displays interface down and up speed"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('interface', None, 'List of interfaces or single NIC as string to monitor, \
-            None to displays all active NICs combined'),
+        ('display_format', '{interface}: {down} \u2193\u2191 {up}',
+         'Display format of down-/upload of given interfaces'),
+        ('interface', None,
+         'List of interfaces or single NIC as string to monitor, '
+         'None to display all active NICs combined'),
         ('update_interval', 1, 'The update interval.'),
         ('use_bits', False, 'Use bits instead of bytes per second?'),
     ]
@@ -41,11 +44,13 @@ class Net(base.ThreadedPollText):
         self.add_defaults(Net.defaults)
         if not isinstance(self.interface, list):
             if self.interface is None:
-                self.interface = ["all"]
+                self.interface = ['all']
             elif isinstance(self.interface, str):
                 self.interface = [self.interface]
             else:
-                raise AttributeError("Invalid Argument passed: %s\nAllowed Types: List, String, None" % self.interface)
+                raise AttributeError(
+                    (f'Invalid Argument passed: {self.interface}\n'
+                     f'Allowed Types: List, String, None'))
         self.stats = self.get_stats()
 
     def convert_b(self, num_bytes: float) -> Tuple[float, str]:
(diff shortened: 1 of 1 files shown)


libqtile/widget/net.py
```diff
@@ -41,22 +44,24 @@ def __init__(self, **config):
...
             if self.interface is None:
-                self.interface = ["all"]
+                self.interface = ['all']
             elif isinstance(self.interface, str):
                 self.interface = [self.interface]
             else:
-                raise AttributeError("Invalid Argument passed: %s\nAllowed Types: List, String, None" % self.interface)
+                raise AttributeError(
+                    (f'Invalid Argument passed: {self.interface}\n'
+                     f'Allowed Types: List, String, None'))
```
ramnes: :warning: We're still supporting Python 3.5 and thus can not use f-strings.

libqtile/widget/net.py
```diff
@@ -41,22 +44,24 @@ def __init__(self, **config):
...
         """Converts the number of bytes to the correct unit"""
         factor = 1000.0
 
         if self.use_bits:
-            letters = ["b", "kb", "Mb", "Gb", "Tb", "Pb", "Eb", "Zb", "Yb"]
+            letters = ['b', 'kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Eb', 'Zb', 'Yb']
             num_bytes *= 8
         else:
-            letters = ["B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
+            letters = ['B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
```
ramnes: Let's not pollute your commit with such changes, and rather try to keep it atomic. If you really care about the codebase being consistent on single and double quotes, please do the change across the whole code and open another PR for this.

ramnes:
Hey @stehessel, thanks for the PR!

A `display_format` option is great, although the diff is hard to read due to the other things you've added to the commit. Can you please strip it down to the bare feature you want to add?

Also, I'm not sure about filling with spaces, it might look weird to me whereas I don't mind if the widget "moves", but let's see what others have to say.

pawamoy:
About shifting: I don't really mind shifting within a single widget, though I think having the text *still* looks better. What bothers me is that this shifting will also push other widgets.

Two solutions:
- directly prevent widget shifting in widgets code itself
- or implement some widget-padding logic. It could be "have a fixed width for the widget, allowing it to move inside this box without pushing other widgets", or "have a spacer widget that acts as a buffer between shifting widgets". If I got it right, the latter already exists as `libqtile.widgets.Spacer` and can be initialized with a length in pixels. You would typically give it a length equal to the biggest shift that can happen in the next/previous widget.

I think the Spacer can also be used to prevent "inside-shifting", by using a TextBox followed by a Spacer and then the widget with the value only (no text), though it's complicated and might not be suited for some widgets (dynamic text).

stehessel:
> Can you please strip it down to the bare feature you want to add?

No worries I reverted most of the changes.

Note that the previous implementation already padded/clipped the text to keep the widget size fixed. It just broke when single space units were introduced. Personally I like the padding/clipping this way, but maybe we can make it optional or include different styles.

ramnes:
> Note that the previous implementation already padded/clipped the text to keep the widget size fixed. It just broke when single space units were introduced.

Ah! Then it totally makes sense to do this for single letter units also, thank you for mentioning this.

libqtile/widget/net.py
```diff
@@ -30,6 +30,8 @@ class Net(base.ThreadedPollText):
     """Displays interface down and up speed"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
+        ('display_format', '{interface}: {down} \u2193\u2191 {up}',
```
ramnes: Let's stay consistent with other widgets:

```suggestion
        ('format', '{interface}: {down} \u2193\u2191 {up}',
```

stehessel: Sure I don't mind that, although I did copy the naming from the check_updates widget for consistency, see https://github.com/qtile/qtile/blob/f186e93a6ae5e23d17f1d177e226eaa123b27cdb/libqtile/widget/check_updates.py#L35

ramnes:
LGTM apart from that last comment. I'll go ahead an make the modification myself, thanks!

stehessel:
@ramnes I believe you forgot to change `display_format` to `format` when it is referenced later in the code.

ramnes:
Yes, and then tried to be sneaky by force pushing a minute later but still, you caught me!

PR #1526, opened 2020-01-23
Custom keyboard display

zordsdavini:
the idea is to avoid long naming or do what you want :)
Ex. I have "RU PHONETIC" and this information is needless for me. RU it is enough.

config.py
```
keyboard_widget = widget.KeyboardLayout(
        configured_keyboards=['us', 'lt', 'ru phonetic'],
        display_map={'ru phonetic': 'ru'},
        options='compose:menu,grp_led:scroll',
        foreground=GREEN
        )
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ qtile X.X.X, released XXXX-XX-XX:
     * features
         - added `guess_terminal` in utils
         - added keybinding cheet sheet image generator
+        - custom keyboardlayout display 
 
 qtile 0.15.0, released 2020-04-12:
     !!! Config breakage !!!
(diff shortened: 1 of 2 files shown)


zordsdavini:
will come back to that later, because I have to layouts with variant and one without. Plus I wanted to make mapping like that:
```
        # display_map={
        #     'us': 'us ',
        #     'lt sgs': 'sgs',
        #     'ru phonetic': 'ru ',
        #     },

```
to save spacing, but it fails.
To be continue...

ep12:
I've subclassed that keyboard layout because I wanted to shorten _NO DEAD KEYS_ to _NDK_. My implementation is a bit different, but maybe you want to have look at it [here](https://github.com/ep12/qtileconf/blob/1bc8b08a3eebaf8bca8bf09d4c914474603a1c96/config.py#L32).  
In turn I'll have a look at your pr. Why exactly are you not satisfied with it at the moment?

zordsdavini:
Nice :) When I'll be back to this diff I will look deepier into yours. Additionally I thought about implementation of flags with
giving directory and then look for file RENAMED_LAYOUT.png or let give pattern for file. And the most important part: I want to
move to better setxkbmat quering and not resetting each time.

zordsdavini:
@ep12 I moved all mapping logic in poll and got all working

zordsdavini:
squashed

ramnes:
Thanks!

Issue #1525, opened 2020-01-23
MPRIS2 widget infinitely adds "Paused: " to song name when paused


judemille:
![Peek 2020-01-22 21-51](https://user-images.githubusercontent.com/8127111/72952987-9ccca200-3d61-11ea-8174-f465a55d69c7.gif)
As shown in the image, "Paused: " gets added to the front of the song title four times, staying there a few seconds, before the song disappears entirely.

whatacold:
Can you reproduce the problem? Maybe I could help fix it, although I haven't used this widget yet.

tych0:
@juliandemille any chance you can test the PR above?

juliandemille:
Sorry, I don't use qtile anymore.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

ramnes:
Fixed in 2caeb6d8540813dc60a9d8e36ca62ac77eadcae8. Thanks @judemille for raising the issue!

PR #1527, opened 2020-01-24
Fix mpris2 pause #1525

whatacold:
I think this could solved the issue #1525 , although I haven't tried it on my side. I am happy to hear some feedback!

### DIFF
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -94,10 +94,7 @@ class Mpris2(base._TextBox):
         olddisplaytext = self.displaytext
         self.displaytext = ''
 
-        metadata = None
-        playbackstatus = None
         metadata = changed_properties.get('Metadata')
-        playbackstatus = changed_properties.get('PlaybackStatus')
         if metadata:
             self.is_playing = True
             self.displaytext = ' - '.join([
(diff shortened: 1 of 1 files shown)


libqtile/widget/mpris2widget.py
```diff
@@ -110,9 +110,9 @@ def update(self, interface_name, changed_properties, invalidated_properties):
             if self.stop_pause_text is not None:
                 self.is_playing = False
                 self.displaytext = self.stop_pause_text
-            elif olddisplaytext:
+            elif self.displaytext:
                 self.is_playing = False
-                self.displaytext = 'Paused: {}'.format(olddisplaytext)
+                self.displaytext = 'Paused: {}'.format(self.displaytext)
```
juliandemille: Since `self.displaytext` is cleared earlier in the function, this would simply yield `Paused: `.

juliandemille: It actually may not if the metadata is changed.

juliandemille: Actually, it might work if the metadata is still there.

whatacold: > Actually, it might work if the metadata is still there.

Yeah, and if `self.displaytext` is cleared, this `elif` will not be hit, and it will yield `Paused` in the following `else`.

tych0:
@juliandemille can you confirm this fixes the issue? I'd be happy to merge it if so.

tych0:
Ping, can anyone confirm this fixes the issue?

ramnes:
I can confirm :)

Issue #1243, opened 2018-12-06
GroupBox: Color of highlighted text


baldrianbandit:
Currently, the text color of active and inactive screens is independent from being selected or not. I'd like to have a variable `selected` to give selected text a different color and thus, creating a nice contrast like this:
![screenshot](https://user-images.githubusercontent.com/12493502/49599492-7000d400-f981-11e8-8a14-ab4f58479361.png)

That variable could default to the value of `active` so that there is no change in the out of the box look. In groupbox.py:
`default = ["selected", "FFFFFF", "Active selected group font colour"), ... ]`
and
`text_color = self.selected` within `if g.screen`

osimplex:
You can modify the colors of active and incative screens.
http://docs.qtile.org/en/latest/manual/ref/widgets.html#groupbox

```
widget.GroupBox(
	active = color,
	inactive = color,
	urgent_text = color
	highlight_method = "block",
	urgent_alert_method = "block",
	urgent_border = color,
	this_current_screen_border = color,
	this_screen_border = color,
	other_current_screen_border = color,
	other_screen_border = color,
	foreground = color,
	background = color
)
```

ramnes:
So basically you'd like to have something to complete `this_current_screen_border`, but on the font side, right? Would something like `this_current_screen_font` (or simply `this_current_screen`) answer your need?

baldrianbandit:
Yeah, exactly! Thank you!

ramnes:
So, do we agree that if we want to be 100% complete and consistent, we should add all those?

* `current_screen_active`
* `current_screen_inactive`
* `this_current_screen_active`
* `this_current_screen_inactive`
* `other_current_screen_active`
* `other_current_screen_inactive`
* `other_screen_active`
* `other_screen_inactive`

Am I forgetting something? Or seeing too large?

ramnes:
I really don't like those names, honestly. If someone has an idea of names that would make all of this simpler, even if it's non-backward-compatible, I'm all hear.

osimplex:
Now I understand the point, sorry @baldrianbandit.

baldrianbandit:
Let me try to recap: If you have two monitors, then _this_ means the selected or main monitor while _other_ is the not selected or secondary monitor. _Active_ means that the screen has windows on it, while _inactive_ screens do not have windows. Wouldn't it be, then:

`this_monitor_selected_screen_active`
`this_monitor_selected_screen_inactive`
[...]

PR #1530, opened 2020-01-29
Add groupbox feature: Font colour when box highlighting

baldrianbandit:
This creates a third font colour called "selected", which is set to a groupbox item when highlight method "block" is used.

Background: Usually, the bar is of dark colour while the fonts are of bright colour. When using highlight method "block", the block must be of bright colour, too, to have a contrast to the bar. Then the font inside the block must be dark again, while the (in)active font colour is still bright. Thus, the font clour inside the block - selected - should be configurable.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -155,6 +155,7 @@ class GroupBox(_GroupBase):
     """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
+        ("selected", "000000", "Active selected group font colour"),
         ("active", "FFFFFF", "Active group font colour"),
         ("inactive", "404040", "Inactive group font colour"),
         (
@@ -334,6 +335,7 @@ class GroupBox(_GroupBase):
                     border = self.bar.background
                     text_color = self.this_current_screen_border
                 else:
+                    text_color = self.selected
                     if self.bar.screen.group.name == g.name:
                         if self.qtile.current_screen == self.bar.screen:
                             border = self.this_current_screen_border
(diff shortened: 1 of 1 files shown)


baldrianbandit:
For reference: https://github.com/qtile/qtile/issues/1243

tych0:
Any chance we can change the name to reference the fact that this is only relevant in 'block' highlight mode? `block_highlight_text_color` or something?

baldrianbandit:
Agreed.

tych0:
Unfortunately I don't think we can change the other ones without some deprecation warnings and backwards compatibility code. So can you either add that code, or drop the others from the patch and just change the one you're adding?

baldrianbandit:
Yeah sorry, guess you're right.

tych0:
Merged it manually, thanks!

m-col:
I think it would this would be better if this:
```python
text_color = self.block_highlight_text_color
```
was instead this:
```python
if self.block_highlight_text_color:
    text_color = self.block_highlight_text_color
```
and the default value was None or False. This way it can be left and would yield the original result, which was not colouring the text by whether it is on screen, only by active or inactive. (Sorry for being awkward but I liked it the old way!)

tych0:
On Sun, Apr 05, 2020 at 02:01:09PM -0700, m-col wrote:
> I think it would this would be better if this:
> ```python
> text_color = self.block_highlight_text_color
> ```
> was instead this:
> ```python
> if self.block_highlight_text_color:
>     text_color = self.block_highlight_text_color
> ```
> and the default value was None or False. This way it can be left and would yield the original result, which was not colouring the text by whether it is on screen, only by active or inactive. (Sorry for being awkward but I liked it the old way!)

Oh, I didn't realize this broke the old behavior. Can you send a patch
to fix it?

m-col:
Done #1586.

PR #1532, opened 2020-02-01
Fix switching group failure (fix #755)

### LINKED ISSUE #755: Keybindings are not active after closing emacs
Hi,
I have some strange problem with a dynamic group where emacs is spawned. In my setup I have a few normal groups which are always active and two dynamic groups for emacs and firefox, respectively. When I start emacs from an empty static group, after its closing qtile stops to react on pressed keys. It seems that screen does not get a focus when the 'emacs' dynamic group becomes non-active. Qtile resumes to react on pressed keys again when a group in the groupbox is selected with a mouse.  After starting emacs from a terminal or from a non-empty group, keybindings remain active after exiting emacs. With firefox everything works as advertised, i.e.  keybindings operate without any problem after closing firefox.  
Best regards,
Toller

whatacold:
This fixes #755 , which described a weird behavior that qtile's hotkeys don't work after Emacs closes, and it can work again if clicking on a groupbox label.

### DIFF
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -155,6 +155,8 @@ Qtile's conversations with the X server. To capture one of these, create an
 This will put the xtrace output in Qtile's logfile as well. You can then
 demonstrate the bug, and paste the contents of this file into the bug report.
 
+Note that xtrace may be named ``x11trace`` on some platforms, for example, on Fedora.
+
 Resources
 =========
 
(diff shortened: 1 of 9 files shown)


whatacold:
There are some comment inside `Window.focus()` method saying that why it works around for Java applications, I'm wondering maybe this may fix that too, but I can't confirm it.

ramnes:
Thanks!

Issue #1410, opened 2019-08-28
libqtile.command_object.SelectError: Unable to make selection on current node


ramnes:
```bash
$ qtile-cmd -o group 
Traceback (most recent call last):
  File "/home/ramnes/.local/bin/qtile-cmd", line 11, in <module>
    load_entry_point('qtile', 'console_scripts', 'qtile-cmd')()
  File "/home/ramnes/perso/qtile/libqtile/scripts/qtile_cmd.py", line 194, in main
    obj = get_object(cmd_client, args.obj_spec)
  File "/home/ramnes/perso/qtile/libqtile/scripts/qtile_cmd.py", line 116, in get_object
    client = client[arg]
  File "/home/ramnes/perso/qtile/libqtile/command_client.py", line 232, in __getitem__
    raise SelectError("Unable to make selection on current node", name, self._current_node.selectors)
libqtile.command_object.SelectError: Unable to make selection on current node
```

This shouldn't happen, I guess?

whatacold:
Can confirm, even an example (`qtile-cmd -o group 3 -f focus_back`) printed by `qtile-cmd` itself can raise this exception.

whatacold:
It seems that this problem was introduced by commit a3ae3c623859b247813fc1876b188c4d40e84df0 , as identified by the following helper script for `git bisect` with the starting good endpoint being `v0.14.2`, which is based on the `scripts/xephyr`:

    #!/bin/sh
    
    HERE=$(dirname $(readlink -f $0))
    SCREEN_SIZE=${SCREEN_SIZE:-1000x800}
    XDISPLAY=${XDISPLAY:-:1}
    LOG_LEVEL=${LOG_LEVEL:-DEBUG}
    LOG_LEVEL=INFO
    if [[ -z $PYTHON ]]; then
        PYTHON=python
    fi
    
    ./scripts/ffibuild
    
    CLIENT="urxvt"
    XTRACE=""
    
    Xephyr +extension RANDR -screen ${SCREEN_SIZE} ${XDISPLAY} -ac &
    #Xephyr -screen ${SCREEN_SIZE} ${XDISPLAY} -ac &
    XEPHYR_PID=$!
    sleep 0.5
    source ~/workspace/virtualenv/qtile-devel/bin/activate
    env DISPLAY=${XDISPLAY} $XTRACE ${PYTHON} "${HERE}"/../bin/qtile -l ${LOG_LEVEL} $@ &
    QTILE_PID=$!
    
[...]

Issue #1298, opened 2019-03-19
qtile-cmd -o window <windowid> ... doesn't convert <windowid> to an int


AntonyBlakey:

przymusp:
I can confirm this, I have some temporary workaround (see code bellow), but this definitely is NOT the solution.

The problem with temporary solution is that not all numbers at command line should be treated as integers.

Possible solutions are:

  * predefined list of elements of path or arguments that should be treated as int
  * getting info on types of arguments from function objects.
 
Any suggestions ? 

```diff
110,115c110
<             if arg.isdigit():
<                 obj = obj[int(arg)]  # check if it is an item
<             else:
<                 obj = obj[arg]  # check if it is an item
< 
---
>             obj = obj[arg]  # check if it is an item
```

whatacold:
@przymusp How about trying the `obj[int(arg)]` version first, then trying `obj[arg]` in addition if the it fails? It's a quite simple way in my opinion.

PR #1533, opened 2020-02-01
Fix qtile-cmd failure, #1410.

whatacold:
This fixes the failure of qtile-cmd as described in #1410, it excludes the initial object of `CommandGraphRoot` type, which has no meaningful children in qtile manager.

### DIFF
--- a/libqtile/command_client.py
+++ b/libqtile/command_client.py
@@ -27,7 +27,7 @@ that interacts with qtile objects, it should favor using the command graph
 clients to do this interaction.
 """
 
-from typing import Any, List, Optional, Union
+from typing import Any, List, Optional
 
 from libqtile.command_graph import (
     CommandGraphCall,
@@ -211,7 +211,7 @@ class InteractiveCommandClient:
         next_node = self._current_node.navigate(name, None)
         return self.__class__(self._command, current_node=next_node)
 
-    def __getitem__(self, name: Union[str, int]) -> "InteractiveCommandClient":
+    def __getitem__(self, name: str) -> "InteractiveCommandClient":
         """Get the selected element of the currently selected object
 
         From the current command graph object, select the instance with the
@@ -219,7 +219,7 @@ class InteractiveCommandClient:
 
         Parameters
         ----------
-        name : Union[str, int]
+        name : str
             The name, or index if it's of int type, of the item to resolve
 
         Return
@@ -234,10 +234,10 @@ class InteractiveCommandClient:
 
         if not isinstance(self._current_node, CommandGraphObject):
             raise SelectError("Unable to make selection on current node",
-                              str(name), self._current_node.selectors)
+                              name, self._current_node.selectors)
 
         if self._current_node.selector is not None:
-            raise SelectError("Selection already made", str(name),
+            raise SelectError("Selection already made", name,
                               self._current_node.selectors)
 
         def _check_item(item):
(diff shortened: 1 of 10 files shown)


whatacold:
It seems that changing the type hint of the `name` argument of `InteractiveCommandClient.__getitem__()` from `str` to `Union[str, int]` involves many other signature changes such as those in `def navigate()`, `def has_item()`, etc.

Should I proceed or leave it as is in this PR?

whatacold:
> Should I proceed or just leave it as is in this PR?

Well, I think I'd better keep it as is.

ramnes:
Also related to #1298 right?

libqtile/command_client.py
```diff
@@ -220,23 +220,42 @@ def __getitem__(self, name: str) -> "InteractiveCommandClient":
...
+                raise SelectError("Item not available in object",
+                                  str(item), self._current_node.selectors)
+
+        if isinstance(name, str) and name.isdigit():
+            # Check the item as is, and check its int version once more if it fails
+            try:
+                _check_item(name)
+            except SelectError:
+                name = int(name)
+                _check_item(name)
```
ramnes: I really don't like this. We should have some way to know whether we must send a string or int.

test/test_qtile_cmd.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+server_config = pytest.mark.parametrize("qtile", [ServerConfig], indirect=True)
+
+
+@server_config
+def test_qtile_cmd(qtile):
+
+    def _qtile_cmd(obj_spec, function, args):
+        "Imitate the behavior of qtile_cmd"
```
ramnes: Is it impossible to use `qtile_cmd` directly?

libqtile/command_client.py
whatacold: > Also related to #1298 right?

Yeah!

How about normalizing the item at first? There is no way to determine if an item is int or str from shell arguments.

test/test_qtile_cmd.py
whatacold: You mean by calling the script directly? It's possible but I'm afraid it's a bit slower, and the output need to be parsed again if some fields need to be asserted.

libqtile/command_client.py
ramnes: From shell arguments no, but from functions parameters yes!

test/test_qtile_cmd.py
ramnes: Yes but then we test the whole thing and I feel it has a lot of value, more than the speed gain.

whatacold: Yeah, I agree that the correctness of the WM and also the test is more important, so I changed it to call `qtile-cmd` directly now.

libqtile/command_client.py
whatacold: Now it's normalized in `qtile_cmd.py`, only where it's necessary.

test/test_command.py
```diff
@@ -161,15 +161,6 @@ class ServerConfig:
 server_config = pytest.mark.parametrize("qtile", [ServerConfig], indirect=True)
 
 
-@server_config
-def test_item_types(qtile):
```
tych0: Can we change this to use the new normalize_item() instead of just dropping the test?

flacjacket:
It looks like all of the comments on this have been resolved, and I think this looks great now, thanks!

PR #1537, opened 2020-02-06
Make toggling group optional (#1491)

heitzmann:
I use a multimonitor setup where specific groups are tied to specific screens, but toggling messes up this configuration. I hope this is not too invasive... I'm proposing this as a change because I could not get it to work properly simply by hacking on my config.


tych0:
Seems reasonable to me, thanks!

PR #1543, opened 2020-03-01
WM_TAKE_FOCUS: use the valid timestamp got from X server.

### LINKED ISSUE #1542: Copy from qt5 app fails with "QXcbClipboard::setMimeData: Cannot set X11 selection owner"
I'm using keepassxc as a password manager, and on a new (work) laptop I installed a more recent qtile than on an old one without taking much note of it (just checked out qtile from github and setup the virtualenv).

Suddenly I could not copy passwords in keepassxc any more and noticed that it prints the error message if I start it from a terminal.

QXcbClipboard::setMimeData: Cannot set X11 selection owner

Googling for the error message gives several ancient qt bugs.

(I found an issue that was found to be caused by a window manager, but I can't find it any more at the moment. I think I saved it in a note on the new work laptop.)

Then I reverted to the commit I was using on the old laptop, after having eliminated all other causes I could think of, and the behaviour was fixed.

I have now bisected this on a completely different laptop (the newer and older work laptops I first tried this with are running openSUSE 15.1, the private laptop runs Ubuntu 18.04) and the first bad commit is 28d5c30005b5342db9ea400e638daad15d110bfe, the fix for #755.

If I go back to the commit before it, I can copy passwords in keepassxc.

I can also replicate the problem in qt5ct (go to a number field in the second tab and try to copy it with ctrl-c, same error printed in the shell, nothing is copied).

I suspect it's reproducible with other qt(5?) apps. I generally don't take note which toolkit an application is using, so I'm not aware of any other qt apps I might use.

The behaviour in #755 is annoying too. I see that this was a very hard one to figure out and fix. I wasn't aware that it was specific to emacs and I never have an empty desktop that often. Copy + paste from keepassxc however is more important to me. (So I'm obviously  using 0617235c3e5748de83b2d8d630f766e3cd5682a4 until this issue is closed.)

whatacold:
This tries to fix #1542 and #755 (again) properly.

[The Timestamp definition in X11 protocol](https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#id2616329) is the following:
> A timestamp is a time value, expressed in milliseconds. It typically is the time since the last server reset. Timestamp values wrap around (after about 49.7 days). The server, given its current time is represented by timestamp T, always interprets timestamps from clients by treating half of the timestamp space as being earlier in time than T and half of the timestamp space as being later in time than T. One timestamp value (named CurrentTime) is never generated by the server. This value is reserved for use in requests to represent the current server time.

It seems like that it is a relative time offset since the server reset, so we don't know what it is by WM itself. In this PR, I take the approach of i3wm (grep `last_timestamp` in the i3 codebase), that keep the event timestamp from the X server, and take it in `WM_TAKE_FOCUS` ClientMessage event.

Though it fixes both #1542 and #755 on my side, I'm not sure if it breaks other things, and if there are better solutions, I'm open to hearing your opinions!

### DIFF
--- a/libqtile/backend/x11/xcore.py
+++ b/libqtile/backend/x11/xcore.py
@@ -84,6 +84,8 @@ class XCore(base.Core):
             "_NET_SUPPORTED", [self.conn.atoms[x] for x in xcbq.SUPPORTED_ATOMS]
         )
 
+        self._last_event_timestamp = xcffib.CurrentTime
+
         wmname = "qtile"
         self._supporting_wm_check_window = self.conn.create_window(-1, -1, 1, 1)
         self._supporting_wm_check_window.set_property("_NET_WM_NAME", wmname)
@@ -273,6 +275,9 @@ class XCore(base.Core):
         """
         assert self.qtile is not None
 
+        if hasattr(event, "time") and event.time > 0:
+            self._last_event_timestamp = event.time
+
         handler = "handle_{event_type}".format(event_type=event_type)
         # Certain events expose the affected window id as an "event" attribute.
         event_events = [
@@ -300,6 +305,12 @@ class XCore(base.Core):
         if not chain:
             logger.info("Unhandled event: {event_type}".format(event_type=event_type))
         return chain
+
+    def get_valid_timestamp(self):
+        """Get a valid timestamp, i.e. not CurrentTime, for X server.
+
+        It may be used in cases where CurrentTime is unacceptable for X server."""
+        return self._last_event_timestamp
 
     @property
     def display_name(self) -> str:
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -21,7 +21,6 @@ import array
 import contextlib
 import inspect
 import traceback
-import time
 import warnings
 from xcffib.xproto import EventMask, StackMode, SetMode
 import xcffib.xproto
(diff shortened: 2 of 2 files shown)


TauPan:
I can confirm this fixes #1542 for me. (If that helps.)

TauPan:
(Interesting... in one of the qt bugs I mentioned, namely https://bugreports.qt.io/browse/QTBUG-65888 copy and paste stops working after Qt apps are running for  a while (a few days). And the section of the X11 protocol definition mentions that the time values wrap around after around 49.7 days. So problems like this are likely to arise if the X server and the client disagree on the age of timestamps. See also https://bugreports.qt.io/browse/QTBUG-65145 I totally didn't notice this and so failed to mention it in my bugreport.)

zordsdavini:
I just wanted to write what we have problem with this place. Thanks for this. My problem is not working Jetbrains apps any more

whatacold:
So does this patch fix the problem of Jetbrains apps? @zordsdavini

tych0:
Sounds like this at least fixes some things. I'll go ahead and merge this @zordsdavini if you have more bugs feel free to open another issue.

Thanks for the quick fix!

bernatnan:
Thanks very much! Works perfect now.

PR #1550, opened 2020-03-05
widgets: drop the mpd1 widgeto

tych0:
904640bbe2a5 ("Implement new mpd widget") notes that support for the
original python-mpd package which this widget depends on ended in 2010.

Even ubuntu/debian now ship python-mpd2 (in a package called python-mpd),
so it's unlikely this code even works any more unless people bother to
install the old dependency.

Let's remove it.


tych0:
Closing in favor of #1514

PR #1551, opened 2020-03-06
Specify wallpapers with wallpaper and wallpaper_mode Screen parameters

m-col:
This is the quick update on wallpaper configuration discussed in #1495. 

Rather than using a dict, `config.Screen` has `wallpaper` and `wallpaper_mode` parameters to configure wallpapers. I've propogated the name change from 'option' to 'mode' through all of the wallpaper code too to keep it consistent and clear.


tych0:
LGTM, @ramnes ?

ramnes:
Awesome, thanks!

PR #1552, opened 2020-03-06
Set wmname according to the config.

### LINKED ISSUE #1548: Jetbrain's products loose focus
# 2020-04-28 update

The main problem is that Jetbrain's products after focusing other window and back looses focus in editor frame. You can't enter any text. This even happens then navigating between Jetbrain's tabs. Additionally popups loose focus, too.
Confirmed on phpStorm, pyCharm, AndroidStudio.

What has been tried:
- set `_JAVA_AWT_WM_NONREPARENTING=1` environment variable
- set ``wmname="LG3D"  in Qtile config
 
revision 28d5c30 is the start of this bug
revision before (0617235) works as expected

workarounds: 
- revert to 0.14.2
- use Xephyr

whatacold:
Ah, this tries to fix #1548 .

### DIFF
--- a/libqtile/backend/x11/xcore.py
+++ b/libqtile/backend/x11/xcore.py
@@ -86,9 +86,9 @@ class XCore(base.Core):
 
         self._last_event_timestamp = xcffib.CurrentTime
 
-        wmname = "qtile"
+        self._wmname = "qtile"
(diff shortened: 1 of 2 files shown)


tych0:
Good spot, thanks :)

zordsdavini:
Good fix. This solved :+1:

PR #1559, opened 2020-03-20
core: properly shutdown event loop

tych0:
Per https://stackoverflow.com/questions/34710835/proper-way-to-shutdown-asyncio-tasks

the way to properly shutdown an event loop is to cancel all the tasks, and
then wait for a CancelError.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -32,6 +32,7 @@ import traceback
 import time
 import warnings
 from typing import Optional
+from concurrent.futures import CancelledError
 
 import xcffib
 import xcffib.xinerama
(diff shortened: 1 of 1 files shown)


libqtile/core/manager.py
```diff
@@ -236,7 +237,10 @@ def gobject_thread():
             await self.finalize()
 
     def loop(self):
-        self._eventloop.run_until_complete(self.async_loop())
+        try:
+            self._eventloop.run_until_complete(self.async_loop())
+        except CancelledError:
+            pass
```
ramnes: Why this diff?

tych0: See the SO from the commit message; I guess the expected behavior is that it throws a CancelledError.

flacjacket:
There is a lot more information in the docs about how to use async than the last time I looked, thanks for finding this!

PR #1560, opened 2020-03-23
let Bars have margins for more control over position

m-col:
Hello!

This small commit gives `bar.Bar` a `margin` parameter that as the name suggests can be used to give the bar some margins, making that 'detached' bar style. Margins can either be an `int` or a `list` of 4 `int`s corresponding to [North East South West]. For example:

```
Screen(
    top=bar.Bar(
	[
	    widget.GroupBox(),
	    widget.Spacer(),
	    widget.Clock(),
	],
	40,
	margin=[10, 10, 0, 10],
    ),
)
```

Would give all sides but South a margin of 10px. If margin was instead just `10` then all sides would get the 10px margin (equivalent to `[10, 10, 10, 10]`).

This works with all configurations of single or multiple bars on any edges of the screen.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -153,6 +153,7 @@ class Bar(Gap, configurable.Configurable):
     defaults = [
         ("background", "#000000", "Background colour."),
         ("opacity", 1, "Bar window opacity."),
+        ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
     ]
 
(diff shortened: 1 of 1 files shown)


tych0:
LGTM, thanks!

ramnes:
Works great, thanks!

m-col:
Great! :)

PR #1563, opened 2020-03-25
use shuffle_left/right alias in all

zordsdavini:
we have alias only in wide type. Added in base xmonad class


tych0:
LGTM, thanks!

PR #1569, opened 2020-03-28
Revert "core: properly shutdown event loop"

### LINKED ISSUE #1566: lazy.restart does not restart the qtile
I don't know why but the lazy.restart does a log out instead of reloading config even with the default config.sorry for the lack of info im a newbie in python

tych0:
This reverts commit 8e249464572b34f88dc78ca420f864baf726690d.

Closes #1566

Seems like something with this is wrong, so let's revert it until someone
has time to look at it further.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -32,7 +32,6 @@ import traceback
 import time
 import warnings
 from typing import Optional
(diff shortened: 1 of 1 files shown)


tych0:
Now wait a minute. This actually works for me. So let's drop this for now.

PR #1570, opened 2020-03-28
add an issue template

tych0:
There have always been a lot of questions-via-issues, but recently there
have also been a lot of issues where the first round trip is "please post
your config" or "please elaborate". Let's try to head that off with an
issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,23 @@
+Please do not ask general questions here! There are [community
+contact](https://github.com/qtile/qtile#community) options for that.
+
+# Issue description
+
+A brief discussion of what failed and how it failed. A description of
+what you tried is helpful, i.e. "When I use lazy.kill() on a window I get
+the following stack trace" instead of "Closing windows doesn't work".
+
+# Qtile version
+
+Please include the exact commit hash of the version of Qtile that failed.
+
+# Stack traces
+
+Please attach any stack traces found in:
+
+* `~/.xsession-errors`
+* `~/.local/share/qtile/qtile.log`
+
+# Configuration
+
+Please include a link or attach your configuration to the issue.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+Please do not ask general questions here! There are [comunnity
+contact](https://github.com/qtile/qtile#community) options for that.
```
ramnes: I'm +1 for the template but I really don't mind the questions here. I actually prefer here than on the mailing list, for example.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Issue description
+
+A brief discussion of what failed and how it failed. A description of what you
+tried is helpful, i.e. "When I use lazy.kill() on a window nothing happens"
```
ramnes: Nothing happens is pretty vague too, maybe talk about the logs?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+instead of "Closing windows doesn't work".
+
+# Qtile version
+
+Please include the exact commit hash of the version of Qtile that failed.
```
ramnes: People installing through their distribution package manager won't know the commit hash for sure

ramnes:
LGTM overall, way overdue

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+Please do not ask general questions here! There are [comunnity
```
pawamoy: typo: comunnity

.github/ISSUE_TEMPLATE.md
pawamoy: If properly tagged, questions on the bugtracker are great to feed the wiki :slightly_smiling_face:

tych0: I can drop this, but then it seems like we should dismantle the mailing list all together if nobody wants to use it.

.github/ISSUE_TEMPLATE.md
tych0: Then they need to find it out :). This seems like useful information for debugging. Now, one thing I do notice is that qtile --version doesn't print it, so we should fix that.

.github/ISSUE_TEMPLATE.md
ramnes: I like this idea, but then to each his favorites tools I guess, so I wouldn't mind keeping the current status to reach more people. I didn't find this unmanageable so far. The only thing I don't want is adding more tools than we have currently, e.g. dealing with another chat than IRC unless we all stop using IRC.

.github/ISSUE_TEMPLATE.md
ramnes: Yes, let's fix that first and say to past the output here

.github/ISSUE_TEMPLATE.md
tych0: Ok. I find it annoying that we use issues for people to ask questions. So maybe we can all agree that we can close questions immediately, or something? I'd prefer for the mailing list to be about qustions, issues to be about issues, etc.

tych0:
Merging this for now. If someone has time to look into how to always get the right git version number, that would be handy.

ramnes:
Ah damn, my server got shutdown and thus I got disconnected from IRC, and then taken over by kids, so I'm not sure what the end of our discussion was! I still really think it's not welcoming / user friendly to tell people not to ask questions here. Do you like the idea of using the GitHub action actions/stale so that old issues or PR get closed automatically, and then remove the first paragraph of the template here?

tych0:
Yeah, that sounds great!

PR #1573, opened 2020-03-29
add callback functions to textbox widget

m-col:
This commit lets you pass a dictionary to the textbox widget mapping mouse buttons to callback functions. Callbacks are passed the Qtile instance. Example config:
```
def myfunc(qtile):
    something_useful()

texbox = widget.TextBox('text', callbacks={1: myfunc})
```
It supports any value for `button` that is passed to `TextBox.button_press` because `button` is used directly to get the callback from the dictionary.

An alternative to this might be to add `callbacks` as a default parameter in the base class for all widgets, leaving their use up to the widget itself in its `button_press` method. This could then be used by any widgets that aren't already interactive, or to unused buttons on those that are.

### DIFF
--- a/libqtile/widget/textbox.py
+++ b/libqtile/widget/textbox.py
@@ -23,12 +23,21 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from typing import Any, List, Tuple
+
 from .. import bar
 from . import base
 
 
 class TextBox(base._TextBox):
-    """A flexible textbox that can be updated from bound keys, scripts, and qshell"""
+    """A flexible textbox that can be updated from bound keys, scripts, and qshell
+
+    Callback functions can be assigned to button presses by passing a dict to the
+    'callbacks' kwarg. For example: {'Button1': func} will execute func when the widget
+    receives a button 1 press. The Qtile instance of passed as the only argument to the
+    callback functions.
+
+    """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
         ("font", "sans", "Text font"),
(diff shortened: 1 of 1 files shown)


libqtile/widget/textbox.py
```diff
@@ -28,14 +28,22 @@
...
+
+    """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
         ("font", "sans", "Text font"),
         ("fontsize", None, "Font pixel size. Calculated if None."),
         ("fontshadow", None, "font shadow color, default is None(no shadow)"),
         ("padding", None, "Padding left and right. Calculated if None."),
         ("foreground", "#ffffff", "Foreground colour."),
+        ("callbacks", {}, "Dict of button press callback functions."),
```
tych0: Can we use the same language as for Mouse() configuration, i.e. "Button1" instead of 1, "Button2" instead of 2, etc.?

And maybe name this mouse_callbacks? Other than that lgtm.

m-col: Done!

m-col:
The second commit didn't turn the integer into a string where it needed to. I've fixed and squashed.

tych0:
Looks like we need a `# type: List[Tuple[str, Any, str]]` comment now to keep pytype happy too, other than that lgtm, thanks!

m-col:
Alright, got that in so hopefully the tests will pass.

m-col:
I forgot to update the `if button in` to use the full name `"ButtonX"` that is used for the dict. That's fixed now.

tych0:
One thing I was thinking is that it might be cool to have this on the base widget class, letting things like groupbox etc. override it as they do. I merged it as-is, but if you're interested in doing that, it would be maybe handy for someone :)

Issue #1441, opened 2019-09-09
Feature Request: tooltips when hovering on widgets


pawamoy:
It could be nice to have tooltips showing up when hovering on a widget, be it to display it's mouse bindings or display some information. The text could be static, or dynamic by providing a command from which to get the text. The command could be run at custom intervals. A bit like what Conky offers :slightly_smiling_face:

zordsdavini:
Maybe it can be not as tooltip but change widget text. Example, I have battery widget and display only left percentage but with this hover I can display all information (remaining time, watts etc. as it is in default text). Another example Gmail widget (actually now it is not working for me), I used it to display only not read emails, but with hover it could give full information: total emails and unread.
Additionally it could get some different color

pawamoy:
Hmm, this is interesting as well! And maybe easier to implement (maybe not:  a tooltip would be an addition to the code, whereas changing the text of the widget would require changing the widgets code).

In fact, if #1440 gets implemented, we could have fake tooltips using notifications. Example: when clicking on your Gmail widget, a notification (using your system's notifications) would pop, showing you extra info :slightly_smiling_face:

zordsdavini:
hovering and pushing notifications would give a lot off noise but maybe this could be the option. Even I imagine it could be possible to expand top bar with text bellow temporally (just brainstorming)

m-col:
(How) Is it possible to listen to mouse-enter events at the level of individual widgets?

tych0:
I think you'd need something like, https://github.com/tych0/qtile/commit/c01d4245557918e68cb2ee4a4716000226e5b839

m-col:
Correct me if I'm wrong, but wouldn't that only pass entry/leaving of the entire bar window, but not entry/leave of a widget to/from another widget?

tych0:
Yeah, true. So you'll need the above plus some other code handling MotionNotify too.

ramnes:
Unless each widget is a client? Not sure if that's feasible or if that makes sense though.

tych0:
I think it's probably simpler to just handle MotionNotify.

m-col:
I've got the window side of things working on [this branch](https://github.com/m-col/qtile/tree/widget_tooltips).

[...]

PR #1575, opened 2020-03-30
move mouse_callbacks into base widget

m-col:
Follow on from #1573. I moved the logic for widget mouse_callbacks into the base class and it seems to be working well. The base button_press method calls these callbacks by default and is then overriden by widgets that have pre-existing interactivity. The callbacks dict can be put into the config's `widget_defaults` and it then applies to all widgets that haven't overriden button_press.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -92,11 +92,19 @@ class _Widget(CommandObject, configurable.Configurable):
 
     The offsetx and offsety attributes are set by the Bar after all widgets
     have been configured.
+
+    Callback functions can be assigned to button presses by passing a dict to the
+    'callbacks' kwarg. For example: {'Button1': func} will execute func when the widget
+    receives a button 1 press. The Qtile instance of passed as the only argument to the
+    callback functions.
     """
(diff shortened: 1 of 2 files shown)


ramnes:
I guess we should migrate the widgets that use clicks?

zordsdavini:
We have ticket like https://github.com/qtile/qtile/issues/1441 , will it cover mouse hover?

tych0:
On Mon, Mar 30, 2020 at 04:57:11AM -0700, Guillaume Gelin wrote:
> I guess we should migrate the widgets that use clicks?

Yeah, that seems like a good idea.

tych0:
I'll go ahead and merge this for now, if someone has the bandwidth to port our widgets that use clicks to this, that would be handy.

flacjacket:
I'm trying to think how this could be used in the current state for the widgets which use clicks? In general, they all will need access to `self` or at least some notion of which widget it is being fired on, should that be something that is passed somewhere? Is there any way for the `default_config` dict to be setup with a set of functions? Or is the idea that those widgets would setup the callbacks in `__init__`?

PR #1583, opened 2020-04-04
x11 backend: use extmap for XFixes

tych0:
In case some X server doesn't have support for XFixes, we don't want to set
up the XFixes extension. Let's just use the same mechanism as we do for
randr and xinerama.


tych0:
Same docs failure, merging :)

PR #1585, opened 2020-04-05
replace cgi.escape with html.escape in mpd2 widget

m-col:
`cgi.escape` is [deprecated](https://docs.python.org/3.5/library/cgi.html) and at somepoint (I think Python 3.8) was removed. `html.escape` is recommended as a drop-in replacement.


tych0:
I get this docs failure on master right now, so it's unrelated. I'll try and bisect tonight. In any case, this patch lgtm. Thanks!

PR #1586, opened 2020-04-06
only use groupbox block_highlight_text_color if set

m-col:
This restores the default behaviour of the groupbox text colour, as discussed [here](https://github.com/qtile/qtile/pull/1530#issuecomment-609482070).

The active group box text colour can be set by using the block_highlight_text_color parameter as per that PR.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -155,7 +155,7 @@ class GroupBox(_GroupBase):
     """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ("block_highlight_text_color", "000000", "Selected group font colour"),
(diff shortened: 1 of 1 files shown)


tych0:
Great, thanks!

m-col:
A docs failure... I guess this is what you mentioned over on the [last PR](https://github.com/qtile/qtile/pull/1585#issuecomment-609492878)?

tych0:
Yeah, see https://github.com/qtile/qtile/pull/1306#issuecomment-609559478

Seems like somehow the rendering of those config defaults is screwing with the templates. I tried various ways of escaping the content, nothing worked though.

tych0:
Anyway, this lgtm too, thanks!

m-col:
Ah yep, looks mysterious. No worries!

PR #1587, opened 2020-04-06
restore mpd widget text color progress tracker

m-col:
This restores the setting for the mpd widget that will change the colour of the text by song progress, as discussed [here](https://github.com/qtile/qtile/pull/1514#pullrequestreview-379857766).


tych0:
I'm gonna hold off on merging this for now, we really need to fix the docs. If we can't get them fixed soon, I'll just revert that whole patch.

m-col:
Sure thing.

That second push removed the escaping, as the text was already escaped.

tych0:
This is our same docs failure from before, which should be fixed now. This LGTM, so I'll merge it. Thanks!

PR #1590, opened 2020-04-07
docs to start Qtile without display manager

zordsdavini:


bin/qtile-session
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+while true; do
+    qtile >/dev/null 2>&1
```
tych0, line 4: I wonder if we shouldn't leave the output, since that's what ~/.xsession-errors shows, and we often ask people to look there when debugging stuff?

That said, can we just drop this and added it to the sample .xinitrc you put below?

zordsdavini: I think we can

docs/manual/config/without-dm.rst
```diff
@@ -0,0 +1,64 @@
...
+    #   nm-applet &
+    #   blueman-applet &
+    #   dunst &
+    #
+    # or
+    #
+    #   source ~/.xsession
+
+    while true; do
+        exec qtile
```
m-col, line 63: `exec` exits from the script when it is done, so the loop won't work. But do you even need a loop? Why exit Qtile just to run it again, when you can just restart it? The only reason you'd exit Qtile on a DM-less setup is if you want the TTY, no?

tych0:
On Fri, Apr 10, 2020 at 10:50:41AM -0700, m-col wrote:
> @m-col commented on this pull request.
> 
> 
> 
> > +::
> +
> +    # some apps that should be started before Qtile, ex.
> +    #
> +    #   [[ -f ~/.Xresources ]] && xrdb -merge ~/.Xresources
> +    #   ~/.fehbg &
> +    #   nm-applet &
> +    #   blueman-applet &
> +    #   dunst &
> +    #
> +    # or
> +    #
> +    #   source ~/.xsession
> +
> +    while true; do
> +        exec qtile
> 
> `exec` exits from the script when it is done, so the loop won't work. But do you even need a loop? Why exit Qtile just to run it again, when you can just restart it? The only reason you'd exit Qtile on a DM-less setup is if you want the TTY, no?

I think the idea is that if qtile crashes, you want to re-exec it
instead of losing the current X state and ending the session.

m-col:
Ah.. fair enough. In that case just removing the exec should get it working.

docs/manual/config/without-dm.rst
zordsdavini: you are right about `exec`. Romeved and it worked. tych0 right about purpose of this loop

zordsdavini:
failing on not related stuff :(

zordsdavini:
Ready to merge :)

docs/manual/config/index.rst
```diff
@@ -197,18 +198,24 @@ creating a `qtile.desktop
 
 A second way to start Qtile is a custom X session. This way allows you to
 invoke Qtile with custom arguments, and also allows you to do any setup you
-want (e.g. special keyboard bindings like mapping caps lock to control, setting
+want (e.g. special keyboardy to start Qtile is a custom X session. This way
+allows you to i bindings like mapping caps lock to control, setting
```
ramnes, line 202: I don't get this diff

docs/manual/config/without-dm.rst
```diff
@@ -0,0 +1,64 @@
...
+
+`username` should be changed to current user name.
+
+Check more for other `examples <https://wiki.archlinux.org/index.php/Getty#Automatic_login_to_virtual_console>`_.
+
+Qtile session
+-------------
+
+Starting Qtile directly will not preserve running applications on Qtile
+restarts by `qtile-cmd -o cmd -f restart` or other ways in Qtile itself. This
```
ramnes, line 30: I don't have this problem?

docs/manual/config/without-dm.rst
```diff
@@ -0,0 +1,64 @@
...
+    #   ~/.fehbg &
+    #   nm-applet &
+    #   blueman-applet &
+    #   dunst &
+    #
+    # or
+    #
+    #   source ~/.xsession
+
+    while true; do
```
ramnes, line 62: I'm worried this could have nasty side effects, and I personally did not need this to run Qtile without DM and still have restarts working

tych0:
On Sun, Apr 12, 2020 at 09:06:31AM -0700, Guillaume Gelin wrote:
> @ramnes commented on this pull request.
> 
> 
> 
> > @@ -197,18 +198,24 @@ creating a `qtile.desktop
>  
>  A second way to start Qtile is a custom X session. This way allows you to
>  invoke Qtile with custom arguments, and also allows you to do any setup you
> -want (e.g. special keyboard bindings like mapping caps lock to control, setting
> +want (e.g. special keyboardy to start Qtile is a custom X session. This way
> +allows you to i bindings like mapping caps lock to control, setting
> 
> I don't get this diff

Me either, I dropped it from what I finally merged :)

> +    # some apps that should be started before Qtile, ex.
> +    #
> +    #   [[ -f ~/.Xresources ]] && xrdb -merge ~/.Xresources
> +    #   ~/.fehbg &
> +    #   nm-applet &
> +    #   blueman-applet &
> +    #   dunst &
> +    #
> +    # or
> +    #
> +    #   source ~/.xsession
> +
> +    while true; do
> 
> I'm worried this could have nasty side effects, and I personally did not need this to run Qtile without DM and still have restarts working

As in, qtile automatically restarts? How?

docs/manual/config/index.rst
zordsdavini: sorry, this is accidantly paste, but I understand it is in master now, so, will fix with next diff

docs/manual/config/without-dm.rst
zordsdavini: Are you sure?

docs/manual/config/without-dm.rst
zordsdavini: side effect could be that your qtile will broke but system will try to restart and will break again and so on

docs/manual/config/without-dm.rst
ramnes: Well, yes, unless I don't understand correctly "preserve running applications". Isn't it just related to #1615?

tych0: No, Qtile can crash for many reasons, not just that one :)

pawamoy: FOr me, restarting Qtile also restores my running applications correctly. Sometimes not in the same order/position, but restored anyway :slightly_smiling_face:

docs/manual/config/without-dm.rst
ramnes: Doesn't systemd have a respawn feature that could handle this more properly?

I personally do not use systemd, so I put this into `/etc/inittab` to accomplish the same:

```
c1:12345:respawn:/sbin/agetty --autologin <username> --noclear 38400 tty1 linux
```

(Note the `respawn` part, which will fallback to a simple shell if agetty crash-loops)

ramnes:
As in a crash loop you cannot stop and recover from

docs/manual/config/without-dm.rst
tych0: Is it agetty or qtile that it will respawn on? The problem here is qtile dying and dropping you back to a shell, killing all your windows.

tych0:
Yes, we should probably add some sort of crash-loop autodetection or something to this to make it a bit nicer.

docs/manual/config/without-dm.rst
ramnes: @tych0 it's not about crashes here, but simple restarts

docs/manual/config/without-dm.rst
ramnes: agetty. But I'm not sure to understand: if qtile crashes then it makes sense to me that all applications get killed, and I don't see how this `while true` would change anything to that. Can you explain further?

docs/manual/config/without-dm.rst
tych0: You're right that I think the text is wrong, but the problem still exists: if Qtile crashes unexpectedly, without something to restart it, the X session will end.

docs/manual/config/without-dm.rst
ramnes: Oh sorry, I think I misread. Well, it will just respawn the `agetty --autologin`, so you'll end up in qtile again.

docs/manual/config/without-dm.rst
ramnes: Which is to be expected, isn't it? Do you imply that we have a way to keep running applications even in the case of a qtile crash?

docs/manual/config/without-dm.rst
tych0: Ugh. github destroyed my formatting, but you get the idea hopefully :)

ramnes: Which answers my latest comment in the other thread. Interesting!

tych0: Your point stands, though, that the explanation here is wrong: qtile doing this on lazy.restart() is a bug, because we re-exec and don't actually exit. So the docs here need fixed.

zordsdavini: Hm, I'll recheck is Qtile restarting correctly and maybe it is enough for `exec qtile`. I had it before but I removed dm and got failed restarts. Thats why I created loop

tych0: The failed restarts were a bug, but Qtile can have many bugs like this, so it's a worthwhile addition to the docs.

zordsdavini: But maybe better to add such story in separate block?

zordsdavini: I've have tested now without loop - all works. I'll update documentation to use `exec qtile`. And about loop, maybe I'll just write some sentence what you can do like that, but if we have config crash then we will loop in error

tych0:
On Tue, Apr 14, 2020 at 01:46:36PM -0700, Arnas Udovičius wrote:
> @zordsdavini commented on this pull request.
> 
> 
> 
> > +
> +::
> +
> +    # some apps that should be started before Qtile, ex.
> +    #
> +    #   [[ -f ~/.Xresources ]] && xrdb -merge ~/.Xresources
> +    #   ~/.fehbg &
> +    #   nm-applet &
> +    #   blueman-applet &
> +    #   dunst &
> +    #
> +    # or
> +    #
> +    #   source ~/.xsession
> +
> +    while true; do
> 
> I've have tested now without loop - all works. I'll update documentation to use `exec qtile`. And about loop, maybe I'll just write some sentence what you can do like that, but if we have config crash then we will loop in error

But `exec qtile` doesn't solve the problem...?

zordsdavini:
What is the problem? The PR was to create doc how to start Qtile without DM, directly from booted computer through systemd. The loop was added because my Qtile was loosing applications on simple monitor plug-in because of corrupted restart. Now restart is fixed (or reverted to old flow) and problem solved.
If we want to have example with loop - I added some info in config/index.rst about possibility

PR #1594, opened 2020-04-08
docs: manually pprint instead of using the filter

tych0:
The problem here is when pprint prints dictionaries or strings over
multiple lines, sphinx gets confused and thinks they're template
directives.

Instead, let's pprint everything before, so we can have super *long* lines,
since web browsers will render it with line breaks as necessary anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-sudo: false
+os: linux
 dist: xenial
 language: python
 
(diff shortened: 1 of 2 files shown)


tych0:
I'm going to go ahead and merge this, since it fixes the build. If anyone objects, then can send a patch ;)

Issue #1579, opened 2020-04-01
Widgets common to multiple bars are duplicated


m-col:
[I see there is a new (to me) issue template. Does this issue count as a 'general question' even though it relates to a potential improvement rather than "how do I do X?"]

# Issue description

I often use 2 monitors which each have their own `Bar`, but most of their widgets are the same. Common widgets that execute code non-interactively (such as checking how many emails one has) will run twice as often as needed. Additionally, common widgets that poll to update their text often go out of sync and display different text, when they should always be identical.

# Configuration

I think the neatest way for this to be controlled at the config level would be simply to pass the same instance of a widget to your bars. Internally somewhere in the base widget could redraw any updates to their list of bars rather than their single bar.

I'd be happy to work on this if you think it would be worth doing!

tych0:
Seems fine to me, the question is really just how to go about doing it. Right now, all the widgets add their state to "self", but it seems like instead they should all be stateless, and take some kind of state object? Then you can indicate via the state object that two widgets should share the same state.

Worth noting though that some widgets you really do want two copies of. For example, GroupBox(). So we need a way to accomplish both.

tych0:
I suppose another option would be to introduce a new widget type, which just renders a copy of whatever the other widget's text is. That seems simpler to do?

m-col:
The new widget type was definitely the easier idea! Submitted at #1596.

PR #1596, opened 2020-04-09
use mirror widgets to paint duplicate widgets

m-col:
re: #1579

This adds a `Mirror` class in `widget.base`. Instances take the place of widgets that have already been passed to a `Bar` and configured, and hook into their `draw` operations to copy the contents of the original widget's `drawer`. Mouse clicks are forwarded to the original. This lets you do this:

```python
cpu = widget.CPUGraph()
mpd = widget.Mpd2()
battery = widget.Battery()
clock = widget.Clock()

screens = [
    Screen(
	top=bar.Bar(
	    [
		widget.GroupBox(),
		cpu, mpd, battery, clock,
	    ],
	),
    ),
    Screen(
	top=bar.Bar(
	    [
		widget.GroupBox(),
		cpu, mpd, battery, clock,
	    ],
	),
    ),
]
```
Widgets can be passed to more than one bar, so that there don't need to be any duplicates executing the same code all the time, and they'll always be visually identical.

This works for all widgets that use `drawers` (and nothing else) to display their contents. Currently this is all widgets except for `Systray`, which looks like it actually just manages tiny windows hoving over the bar or something.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -222,7 +222,10 @@ class Bar(Gap, configurable.Configurable):
         qtile.windows_map[self.window.window.wid] = self.window
         self.window.unhide()
 
-        for i in self.widgets:
+        for idx, i in enumerate(self.widgets):
+            if i.configured:
+                i = i.create_mirror()
+                self.widgets[idx] = i
             qtile.register_widget(i)
             i._configure(qtile, self)
         self._resize(self.length, self.widgets)
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -283,6 +283,9 @@ class _Widget(CommandObject, configurable.Configurable):
         except:  # noqa: E722
             logger.exception('got exception from widget timer')
 
+    def create_mirror(self):
+        return Mirror(self)
+
 
 UNSPECIFIED = bar.Obj("UNSPECIFIED")
 
@@ -590,6 +593,40 @@ class MarginMixin(configurable.Configurable):
 
     margin_x = configurable.ExtraFallback('margin_x', 'margin')
     margin_y = configurable.ExtraFallback('margin_y', 'margin')
+
+
+class Mirror(_Widget):
+    def __init__(self, reflection):
+        _Widget.__init__(self, reflection.length)
+        reflection.draw = self.hook(reflection.draw)
+        self.reflects = reflection
+        self._length = 0
+
+    @property
+    def length(self):
+        return self.reflects.length
+
+    @length.setter
+    def length(self, value):
+        self._length = value
+
+    def hook(self, draw):
+        def _():
+            draw()
+            self.draw()
+        return _
+
+    def draw(self):
+        if self._length != self.reflects.length:
+            self._length = self.length
+            self.bar.draw()
+        else:
+            self.drawer.ctx.set_source_surface(self.reflects.drawer.surface)
+            self.drawer.ctx.paint()
+            self.drawer.draw(offsetx=self.offset, width=self.width)
+
+    def button_press(self, x, y, button):
+        self.reflects.button_press(x, y, button)
 
 
 def deprecated(msg):
--- a/libqtile/widget/__init__.py
+++ b/libqtile/widget/__init__.py
@@ -25,6 +25,7 @@ from .import_error import make_error
 
 # only directly import widgets that do not have any third party dependencies
 # other than those required by qtile, otherwise use the same import function
+from .base import Mirror  # noqa: F401
 from .clock import Clock  # noqa: F401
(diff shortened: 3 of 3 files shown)


libqtile/bar.py
```diff
@@ -222,8 +222,12 @@ def _configure(self, qtile, screen):
...
         self.window.unhide()
 
-        for i in self.widgets:
-            qtile.register_widget(i)
+        for idx, i in enumerate(self.widgets):
+            if i.configured:
+                i = i.create_mirror()
+                self.widgets[idx] = i
+            else:
+                qtile.register_widget(i)
```
tych0, line 230: So it looks like we skip registering the widget, which isn't ideal. For example, we call widgets' finalize() methods based on there presence in widgets_map, which means we'll leak X resources if we don't include them here. Can we include them, maybe with "-mirror" tacked on or something?

libqtile/widget/base.py
```diff
@@ -592,5 +595,39 @@ class MarginMixin(configurable.Configurable):
...
+
+    def hook(self, draw):
+        def _():
+            draw()
+            self.draw()
+        return _
+
+    def draw(self):
+        if self._length != self.reflects.length:
+            self._length = self.length
```
tych0, line 621: Do we need this _length member at all? Can we just get rid of it entirely, and have the setter body be `pass`?

tych0:
Neat, this takes surprisingly little code :). Two minor nits, but other than that it seems reasonable to me.

The only other thing I can think is maybe we want the API to be explicit, i.e.:
```
bar.Bar([foo, Mirror(foo)])
```

but I don't really have any good reason for wanting that, and this requires the user to know less, so it seems like a better idea.

libqtile/widget/base.py
m-col: I think so - line 620 detects if the length has changed, so that if it has changed it will re-draw the whole bar. If it hasn't changed length then it will just re-draw itself. Or am I missing something?

libqtile/bar.py
m-col: Good point - always registering and removing the `else` should do it. Looking at the register_widget code though I'm not sure it is registering all widgets every time. If you mirror a widget more than once then the repeated name stops all others being registered. Could qtile.widgets_map not just be a list of all registered widgets instead of a dict using their names?

m-col:
I think the repeat use of an instance seems the most intuitive, what somebody might think to try without any previous knowledge if that is what they wanted
```
bar.Bar([foo])
bar.Bar([foo])
```
However, explicitly using `Mirror` in the config gives the exact same result anyway. I just pushed another commit that exposes Mirror at widget.Mirror so both options are available.

libqtile/bar.py
tych0: Yeah, I think the only reason it's a list now is that everything was unique.

In fact you're right, if people have two bars with a GroupBox() on either of them, it will only register the first one. So we're already leaking this anyway. Switching it to a list would be good.

libqtile/widget/base.py
tych0: I guess I'm wondering why does the length ever diverge from the underlying widget?

libqtile/bar.py
tych0: So to resolve this, can we just call .register_widget() on everything, and later we can fix that code to use a map instead of a list?

tych0:
> I think the repeat use of an instance seems the most intuitive

Yeah, I think you're probably right there. The patch to expose it via widgets is great though, thanks for that!

libqtile/bar.py
m-col: Sure!

libqtile/widget/base.py
m-col: When the original widget changes length, it draws. When it draws, it afterwards run's the mirrow's draw. At this draw execution the lengths mismatch, so the mirror gets the new length from the original and then re-draws the whole bar. Redrawing the whole bar then executes the mirror's draw a second time, but this time the length matches the original's and draws itself. Make sense?

tych0: Yeah, I'm just wondering if we can skip that and just re-draw our own with the new length from the underlying widget? Or do we need to do the bar.draw() to recalculate all the offsets?

m-col: My understanding of this came from [here](https://github.com/qtile/qtile/blob/master/libqtile/widget/base.py#L247), which suggests you ought to redraw the whole bar when the length changes. I guess this is to recalculate the offsets like you say.

tych0: Ah, yeah, makes sense. Can you squash the fixes? This LGTM.

m-col:
All squashed and ready to go, test pending.

tych0:
Sweet! This looks great :)

PR #1598, opened 2020-04-10
use mouse_callbacks in all widgets

m-col:
The sequel to #1575 : mouse_callbacks dict used by all widgets.

The widgets that re-implement `button_press` have been updated to perform the same lookup in `mouse_callbacks`, and if they find a callback then they execute it and return. If they do not find a callback for a given button then they continue as normal. This means that any button can be given a callback for any widget, and any code that would normally be executed upon a button press is overriden if the user provides a callback for that button.

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -97,6 +97,9 @@ class CheckUpdates(base.ThreadedPollText):
 
     def button_press(self, x, y, button):
         # type: (int, int, int) -> None
-        base.ThreadedPollText.button_press(self, x, y, button)
+        name = 'Button{0}'.format(button)
+        if name in self.mouse_callbacks:
+            self.mouse_callbacks[name](self.qtile)
+            return
         if button == 1 and self.execute is not None:
             Popen(self.execute, shell=True)
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -96,6 +96,10 @@ class Backlight(base.InLoopPollText):
         self.call_process(shlex.split(self.change_command.format(value)))
 
     def button_press(self, x, y, button):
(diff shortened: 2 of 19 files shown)


libqtile/widget/check_updates.py
```diff
@@ -97,6 +97,9 @@ def poll(self):
 
     def button_press(self, x, y, button):
         # type: (int, int, int) -> None
-        base.ThreadedPollText.button_press(self, x, y, button)
+        name = 'Button{0}'.format(button)
+        if name in self.mouse_callbacks:
+            self.mouse_callbacks[name](self.qtile)
+            return
         if button == 1 and self.execute is not None:
```
tych0, line 104: Sorry, I was thinking more that we hoist dispatch of all the other calculations into a button press, rather than copy paste this into every widget.

So for this one, we'd register a callback for Button1 that is the body of this function.

It's also a bit awkward that we pass the qtile object to the buttons, since for the most part I assume they'll want to operate on themselves. I wonder if we can change it to pass the button?

m-col: For your first comment - I agree that that would be much better, but figured this provide the functionality for all widgets in the meantime while they each get ported over time. Understandable if you'd rather not add this in this way (it's not particularly pretty code).

For your second comment - do you mean just the button string i.e. `Button1`, or the widget's self?

tych0: I was thinking the widget's self reference. Yeah, I'd rather not merge this, I was thinking a port some/all the widgets would be better.

m-col: No worries at all, I'll add a couple of these widgets to my to-do list!

20jam: #1713

Issue #1449, opened 2019-09-14
To add config check before restart


zordsdavini:
Maybe ir would be good practice to validate user config before restarting Qtile. We even can implement flag to do Hard code restart what because of broken user config should fall back to default config. But Maybe this iš needless

pawamoy:
So, I wrote a snippet that checks if a configuration file is valid. You can put it in the config file itself, in the `if __name__ == "__main__":` conditional block. Here it is:

```python
if __name__ == "__main__":
    from libqtile.confreader import Config, ConfigError
    from libqtile.core.xcore import XCore

    try:
        Config.from_file(XCore(), __file__)
    except ConfigError as error:
        print(
            "The configuration file '{}' generated the following error:\n".format(
                __file__
            ),
            file=sys.stderr,
        )
        print("{}\n".format(error), file=sys.stderr)

        if str(error).startswith("No such key"):
            # there is an invalid key in the config

            def find_similar_keys(key):
                import regex
                from libqtile.xkeysyms import keysyms

                regexp = regex.compile(
                    r"({}{{e<{}}})".format(key, len(key)), regex.IGNORECASE
                )
                similar_keys = []
                for qtile_key in keysyms.keys():
                    if regexp.match(qtile_key):
                        similar_keys.append(qtile_key)
                if not similar_keys:
                    return "No key similar to '{}' could be found in libqtile.xkeysyms.keysyms :/".format(
                        key
                    )
                if len(similar_keys) == 1:
                    return "Maybe you meant '{}'?".format(similar_keys[0])
                return "Maybe you meant '{}', or '{}'?".format(
                    "', '".join(similar_keys[:-1]), similar_keys[-1]
                )

            print(
                find_similar_keys(str(error).replace("No such key: ", "")),
                file=sys.stderr,
            )
        sys.exit(1)
    else:
        print("OK!")
        sys.exit(0)
```

Now just run `python ~/.config/qtile/config.py` to validate it. It's a bit lengthy, but it will print configuration errors as well as erroneous keys and valid similar candidates if any. This last feature requires `regex` to be installed/importable.

Example output when using `left` instead of `Left`:
```
The configuration file 'config.py' generated the following error:

No such key: left

[...]

PR #1603, opened 2020-04-11
Store the config file path in the Config object

pawamoy:
This change is required for the implementation of #1449.

When (re)validating the config before a restart, we need its file path to reload it.

So, when building the Config object, we simply store the file path as an attribute.

`None` is allowed for when there is an exception at startup. In that case, the Config object created is empty, and Qtile adds the "Config err!" widget to the bar.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -54,14 +54,16 @@ class Config:
         "wmname",
     ]
 
-    def __init__(self, **settings):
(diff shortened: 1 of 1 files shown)


pawamoy:
If you'd prefer I send everything for #1449 in a single PR, just tell me, I'll close this one and reopen a new one.

tych0:
Either way works for me, thanks!

PR #1608, opened 2020-04-12
widget.Backlight: Use /sys interface when command=None

m-col:
This change to the Backlight widget makes the backlight widget attempt to change brightness directly to the brightness file available in /sys/class/backlight. To enable this, the `change_command` parameter needs to be set to `None`. The `backlight_name` parameter must be set correctly.

To change the brightness this way, the user may need to be added to the 'video' group (linux). This may be dependent on the correct udev rules also being installed to change the group on the brightness file to 'video'; these are typically installed alongside brightness tools such as brightnessctl so installing that is an easy way to get it working.

### DIFF
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -33,7 +33,12 @@ BACKLIGHT_DIR = '/sys/class/backlight'
 
 
 class Backlight(base.InLoopPollText):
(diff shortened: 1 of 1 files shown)


tych0:
lgtm, I added some of the text from your PR (slightly reworked) into the backlight docs too. Thanks!

m-col:
Great, thanks!

PR #1611, opened 2020-04-13
Validate config before restart to prevent errors

pawamoy:
Follow-up of #1603 to implement #1449.

~~I guess we could also log that we prevented a restart in case there's no notification server listening to notifications on the user system.~~ Done

Also not sure about how to write tests for this. Any advice?

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -63,8 +63,8 @@ from ..backend.x11.xcore import XCore
 
 
 def find_similar_keys(key):
-    # Python package 'regex' is required,
-    # but the import error will be caught in validate_config
+    # Python package 'regex' is optional.
+    # The import error will be caught in validate_config
     import regex
 
     regexp = regex.compile(
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -51,15 +51,14 @@ from ..extension.base import _Extension
 from .. import hook
 from .. import utils
 from .. import window
-from ..backend.x11 import xcbq
-from .. import command_interface
-from ..command_client import InteractiveCommandClient
-from ..command_interface import QtileCommandInterface, IPCCommandServer
-from ..command_object import CommandObject, CommandError, CommandException
-from ..lazy import lazy
-from ..confreader import Config, ConfigError
-from ..xkeysyms import keysyms
-from ..backend.x11.xcore import XCore
+from libqtile.backend.x11 import xcbq
+from libqtile import command_interface
+from libqtile.command_client import InteractiveCommandClient
+from libqtile.command_interface import QtileCommandInterface, IPCCommandServer
+from libqtile.command_object import CommandObject, CommandError, CommandException
+from libqtile.lazy import lazy
+from libqtile.confreader import Config, ConfigError
+from libqtile.backend.x11.xcore import XCore
 
 
 def validate_config(file_path):
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -62,30 +62,13 @@ from ..xkeysyms import keysyms
 from ..backend.x11.xcore import XCore
 
 
-def find_similar_keys(key):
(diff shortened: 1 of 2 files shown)


pawamoy:
Unrelated docs failure, and mypy failure because it cannot find info about the `regex` package. Should I `# type: ignore` it?

tych0:
I don't think the docs error is unrelated: you're adding an import of xcore, which imports xcffib.render. It is easy to fix though: add xcffib.render to the list of mock modules in doc/conf.py.

Is there a reason it can't find regex? Seems like an external dependency we should annotate?

pawamoy:
> I don't think the docs error is unrelated: you're adding an import of xcore, which imports xcffib.render. It is easy to fix though: add xcffib.render to the list of mock modules in doc/conf.py.

Ah, indeed, sorry!

> Is there a reason it can't find regex? Seems like an external dependency we should annotate?

Should I add it to the list of things installed in CI? It's an external dependency indeed: a Python package. How would we annotate it ourselves? Not sure to understand.

tych0:
If it's an external dependency, we need to list it in requirements.txt.

pawamoy:
It is, but it's not required. It simply improves the error message when the user specified a wrong key (e.g. `left` instead of `Left`).

tych0:
Any chance we can port it to re? That's in the python stdlib.

pawamoy:
I don't think so, but I'll see if it's possible :slightly_smiling_face:

pawamoy:
Yeah, no I used `regex` because it allows to do fuzzy-matching, which `re` doesn't.

libqtile/core/manager.py
```diff
@@ -50,12 +51,75 @@
...
-from libqtile.lazy import lazy
+from ..backend.x11 import xcbq
+from .. import command_interface
+from ..command_client import InteractiveCommandClient
+from ..command_interface import QtileCommandInterface, IPCCommandServer
+from ..command_object import CommandObject, CommandError, CommandException
+from ..lazy import lazy
+from ..confreader import Config, ConfigError
+from ..xkeysyms import keysyms
+from ..backend.x11.xcore import XCore
```
ramnes, line 62: Do we really want to favor relative imports over absolute imports? IIRC, PEP8 recommends absolute imports, and I personally prefer them also.

pawamoy: I personally encountered import issues (module not initialized ot similar) with absolute imports, that's why I changed them to relative ones. Maybe it was just misconfiguration on my side. If absolute imports are preferred, lets use them! Either way, we should be consistent.

ramnes:
Thanks for this PR, I think it will help a lot of people!

I'm not sure about the `find_similar_keys` function. I feel that it clutters the code for not much. Also, I doubt anyone will install the dependency, unless pulled on the system for another reason. Overall I'm really not fond of optional dependency inside the core and would be in favor of either making it a hard dependency because we think it adds value, or removing the said function.

libqtile/core/manager.py
pawamoy: I'm using relative imports in all my projects, I thought I had read somewhere they were best-practice... either I misread, or the best-practices have changed :smile: Searched again, and indeed there are more pros than cons for absolute imports. Will fix all my projects haha

ramnes: Let's keep this PR atomic and stick to how it was before, we'll go through all imports at once later. I'm opening an issue for this.

pawamoy:
Fair enough. I don't think it adds so much value. These key errors will not happen often, and all the keys are listed in the docs. Users can even check the code locally to see them. This should be moved as a community-thing in the wiki :slightly_smiling_face:

libqtile/core/manager.py
pawamoy: Changed back to absolute imports

tych0:
Thanks! And thanks @ramnes for the review as well :)

PR #1612, opened 2020-04-13
Exclude new _ffi modules location from flake8

pawamoy:
The configuration was still excluding _ffi modules in `libqtie/core`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,7 +95,7 @@ source-dir = docs
 build-dir = docs/_build
 
 [flake8]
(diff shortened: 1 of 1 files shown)


tych0:
Just for my own edification, it seems like we're not running flake8 in
travis any more, right? Maybe we should start doing that again.

pawamoy:
Sure, this seems like it would be a good thing to do :slightly_smiling_face:

tych0:
Oh, looks like we are actually running it, just not on the generated code, since we don't generate the ffi files in the tox:pep8 env. I think that's fine, so we should just merge this as-is.

PR #1613, opened 2020-04-13
Adjust X and Y when adjusting maximised dimensions using increment hints

m-col:
When a window is maximised and it gives height and width increment hints, the width and height of the window is decreased to the nearest multiple of these increments. This can leave asymmetric gaps at the right and bottom edges that look like an unintentional bug (which I sought out to fix until I discovered that this was intentionally to follow increment hints!). This change adjusts the window's X and Y position to position any maximised windows following these hints so that it is centred, removing the weird cut edge effect.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1007,10 +1007,14 @@ class Window(_Window):
             height = max(self.height, self.hints.get('min_height', 0))
 
             if self.hints['base_width'] and self.hints['width_inc']:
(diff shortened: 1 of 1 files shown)


ramnes:
@m-col @tych0 This commit completely breaks floating windows resizing for me.

m-col:
Oh no! I just checked and it has messed it up for me too.

I will take a look to see if there is a way for the adjustment to be done only
when its a fullscreen action.

PR #1614, opened 2020-04-13
exclude libqtile/backend/x11/_ffi_xcursors.py from flake8

m-col:
I seem to be bombarding you with PRs at the minute... it's the lockdown.

This one just updates the flake8 config to exclude `libqtile/backend/x11/_ffi*.py`. I think this file used to be in `libqtile/core` before the backend stuff was created, so I've changed that path.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,7 +95,7 @@ source-dir = docs
 build-dir = docs/_build
 
 [flake8]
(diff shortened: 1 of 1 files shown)


pawamoy:
Too late :D #1612

m-col:
Oh wow, guess we both got annoyed by those false positives in the flake8 output, huh!

PR #1617, opened 2020-04-14
Only centre fullscreen windows when adjusting to increments

m-col:
This fixes the bug I introduced a few hours ago, doh! re #1613

It only updates x and y if the window that is being configured is fullscreen and not any other type of floating window.


ramnes:
Thanks!

Issue #1625, opened 2020-04-16
Auto-sort the imports


pawamoy:
Quality approach again (#1618, #1623): would you consider using a tool like [isort](https://github.com/timothycrosley/isort) to automatically order the imports?

With the following options:

```toml
[tool.isort]
line_length = 120
not_skip = "__init__.py"
multi_line_output = 3
force_single_line = false
balanced_wrapping = true
default_section = "THIRDPARTY"
known_first_party = "libqtile"
include_trailing_comma = true
```

`isort -y -rc libqtile` will do things like

```diff
diff --git a/libqtile/backend/x11/xcbq.py b/libqtile/backend/x11/xcbq.py
index 10f55586..217d206c 100644
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -34,25 +34,23 @@
     A minimal EWMH-aware OO layer over xcffib. This is NOT intended to be
     complete - it only implements the subset of functionalty needed by qtile.
 """
-from collections import OrderedDict
-from itertools import repeat, chain
-import operator
 import functools
+import operator
 import typing
+from collections import OrderedDict
+from itertools import chain, repeat
 
-from xcffib.xproto import CW, WindowClass, EventMask
-from xcffib.xfixes import SelectionEventMask
-
+import cairocffi
+import cairocffi.xcb
 import xcffib
 import xcffib.randr
 import xcffib.xinerama
 import xcffib.xproto
-
-import cairocffi
-import cairocffi.xcb
-
 from libqtile import xkeysyms
 from libqtile.log_utils import logger
+from xcffib.xfixes import SelectionEventMask
+from xcffib.xproto import CW, EventMask, WindowClass
+
 from .xcursors import Cursors
 
 keysyms = xkeysyms.keysyms
```

From `git diff --stat`: 86 files changed, 210 insertions(+), 214 deletions(-).

Isort organizes the imports into three (or more) categories: standard lib, [known frameworks, ] third-party, current package.

In CI you can check it's properly formatted with `isort -c -rc libqtile`.

tych0:
Yeah, this seems like a no-brainer. I'd say if you want to send a patch for this today that checks it in CI, I'd be happy to merge it.

ramnes:
Merged, thanks!

pawamoy:
Should we add a `format` rule in the Makefile to help contributors format the imports? Doing so manually is not very user-friendly :sweat_smile:

[...]

Issue #1618, opened 2020-04-14
Be consistent on imports style


ramnes:
[PEP8 favors absolute imports](https://www.python.org/dev/peps/pep-0008/#imports), let's use them everywhere in place of the relative ones we have.

The discussion started here: https://github.com/qtile/qtile/pull/1611#discussion_r408289718

tych0:
+1, sounds good to me. And if there's a linter we can use to reject relative imports after we fix this, even better.

pawamoy:
I'm taking care of this right now, before #1625.

pawamoy:
I couldn't find a tool to check for presence/absence of relative imports, but we could write a Python script:

```python
import ast

with open("file.py") as fd:
    code = ast.parse(fd.read())

for node in code.body:
    if isinstance(node, ast.ImportFrom) and node.level > 0:
        print("detected relative import!")
```

We would need something a bit more complex of course, to recurse on nodes' bodies, etc.

ramnes:
Isort doesn't seem to do that, bummer. Just found this library, maybe it's would be worth a try?

https://github.com/adamchainz/flake8-tidy-imports

ramnes:
Oh, pylint seems to check and enforce absolute imports also.

pawamoy:
The flake8 plugin seems like the way to go :slightly_smiling_face:

ramnes:
Thanks @pawamoy!

PR #1626, opened 2020-04-16
Sort imports, add check in CI

pawamoy:
For issue #1625

First commit adds the configuration for isort in setup.cfg, the tox env in tox.ini, and the job in travis config.

Next commit will actually fix the imports, to see if CI passes.

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -8,6 +8,11 @@
 # The above copyright notice and this permission notice shall be included in
 # all copies or substantial portions of the Software.
 
+import io
+import os
+import re
+from collections import OrderedDict, defaultdict, namedtuple
+
 # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 # IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 # FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -17,10 +22,6 @@
 # SOFTWARE.
 import cairocffi
 import cairocffi.pixbuf
-import io
-import os
-import re
-from collections import namedtuple, defaultdict, OrderedDict
 
 
 class LoadingError(Exception):
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -4,15 +4,10 @@
 # to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 # copies of the Software, and to permit persons to whom the Software is
 # furnished to do so, subject to the following conditions:
-
+#
 # The above copyright notice and this permission notice shall be included in
 # all copies or substantial portions of the Software.
-
-import io
-import os
-import re
-from collections import OrderedDict, defaultdict, namedtuple
-
+#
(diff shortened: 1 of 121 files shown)


pawamoy:
I think #1618 should be done first, to avoid ordering imports now and re-ordering them after #1628.

pawamoy:
Job correctly failing, pushing the fix :slightly_smiling_face:

ramnes:
Yisss! I don't like the 120 line length though, but I guess we'll discuss this in the issue about code formatting.

pawamoy:
I can change it, no problem. Thing is, flake8 is configured with a line length of 120 so I used the same length :)

ramnes:
Oh! If flake8 is already configured to 120, then don't mind me.

tych0:
I think I just did that to avoid having to reformat everything. I think we
can definitely have a discussion about shrinking it.

pawamoy:
- should I put the isort command inside of the pep8 tox env instead of its own?
- what line length?

tych0:
Putting it in pep8 is fine. For imports, seems like we should be able to get away with 80 characters pretty easily for a line length.

ramnes:
You might want to use https://github.com/gforcada/flake8-isort

pawamoy:
Alright I'll update this. I should definitely take a look at all those flake8 plugins :smile:

pawamoy:
Some things detected by `flake8-isort` are not fixed by running `isort` itself with the same configuration.

```
./libqtile/widget/battery.py:40:1: I001 isort found an import in the wrong position
./libqtile/widget/battery.py:47:1: I003 isort expected 1 blank line in imports, found 0
./libqtile/widget/battery.py:47:1: I003 isort expected 1 blank line in imports, found 0
./libqtile/widget/generic_poll_text.py:21:1: E303 too many blank lines (3)
```

I fixed them manually.

The first three were due to an import kept apart because of a `noqa` comment.

```python
from typing import Dict  # noqa: F401; noqa: F401
from typing import Any, List, NamedTuple, Optional, Tuple
```

Fixed by

```python
from typing import Any, Dict, List, NamedTuple, Optional, Tuple
```

Not sure what the F401 warning is about.

libqtile/images.py
```diff
@@ -8,6 +8,11 @@
 # The above copyright notice and this permission notice shall be included in
 # all copies or substantial portions of the Software.
 
+import io
+import os
+import re
+from collections import OrderedDict, defaultdict, namedtuple
+
```
pawamoy, line 15: Hmm, this doesn't seem right. I think we should concatenate both comments section to the imports are not inserted between the two.

test/layouts/test_bsp.py
```diff
@@ -17,12 +17,13 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from test.conftest import no_xinerama
+from test.layouts.layout_utils import assert_focus_path, assert_focused
+
```
pawamoy, line 22: I think we should mark `test` as third-party in isort's config. Will fix it.

pawamoy:
I'm not sure why but `flake8` does not detect unused imports in the tests, for example `import libqtile.bar` and `import libqtile.scratchpad` in `test_scratchpad.py`.

tych0:
This all looks great to me, thank you for doing it!

PR #1630, opened 2020-04-16
Add description for Keys in default_config.py

arczan1:
My first pull request, It is possible that I did something wrong


tych0:
Looks great to me, thanks!

PR #1631, opened 2020-04-16
Keybinding image generator

zordsdavini:
Should be run from git root directory (because logo.png is used)
as argument config can be added:
`python scripts/gen-keybinding-img /home/arnas/.config/qtile/config.py`
if no config given - default config is generated
mod4-shift__
mod4-control__
mod4__

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,7 +14,7 @@ ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
-.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext coverage
+.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext coverage genkeyimg
 
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
@@ -38,11 +38,12 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
+	@echo "  genkeyimg  to generate keybindings images. Will save in _static/keybindings"
 
 clean:
 	-rm -rf $(BUILDDIR)/*
 
-html:
+html: genkeyimg
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
@@ -157,3 +158,12 @@ coverage:
 	$(SPHINXBUILD) -b coverage $(ALLSPHINXOPTS) $(BUILDDIR)/coverage
 	@echo "Testing of coverage in the sources finished, look at the " \
 	      "results in $(BUILDDIR)/coverage/python.txt."
+
+genkeyimg:
+	mkdir -p ./_static/keybindings
+	rm -f ./_static/keybindings/*.png
+	../scripts/gen-keybinding-img -o ./_static/keybindings
+	sed -i '/^\.\. LS_PNG/,/^\.\. END_LS_PNG/{/^\.\. image/d}' manual/commands/keybindings.rst
+	sed -i '/^\.\. END_LS_PNG/e find _static/keybindings/* | awk '"'"'{ print length, $$0 }'"'"' | sort -n | cut -d" " -f2- | awk '"'"'{print ".. image:: /" $$1}'"'"'' manual/commands/keybindings.rst
+	@echo
+	@echo "Keybinding images have been generated."
--- a/docs/manual/commands/keybindings.rst
+++ b/docs/manual/commands/keybindings.rst
@@ -1,0 +1,38 @@
+=====================
+Keybindings in images
+=====================
+
+Default configuration
+=====================
+
+.. don't delete LS_PNG and END_LS_PNG (it is used for `make genkeyimg`)
+.. LS_PNG
+.. image:: /_static/keybindings/mod4.png
+.. image:: /_static/keybindings/mod4-shift.png
+.. image:: /_static/keybindings/mod4-control.png
+.. END_LS_PNG
+
+Generate your own images
+========================
+
+Qtile provides a tiny helper script to generate keybindings images from a
+config file. In the repository, the script is located under
+``scripts/gen-keybinding-img``.
+
+This script accepts a configuration file and an output directory. If no
+argument is given, the default configuration will be used and files will be
+placed in same directory where the command has been run.
+
+::
+
+    usage: gen-keybinding-img [-h] [-c CONFIGFILE] [-o OUTPUT_DIR]
+
+    Qtile keybindings image generator
+
+    optional arguments:
+        -h, --help          show this help message and exit
+        -c CONFIGFILE, --config CONFIGFILE
+                            use specified configuration file. If no presented
+                            default will be used
+        -o OUTPUT_DIR, --output-dir OUTPUT_DIR
+                            set directory to export all images to
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,3 +2,5 @@ sphinx
 sphinx_rtd_theme
 sphinxcontrib-seqdiag
 numpydoc
+
+-r ../requirements.txt
--- a/scripts/gen-keybinding-img
+++ b/scripts/gen-keybinding-img
@@ -1,0 +1,430 @@
+#!/usr/bin/env python
+
+#######################################
+#  Qtile keybindings image generator  #
+#######################################
+
+import getopt
+import os
+import sys
+
+import cairocffi as cairo
+from cairocffi import ImageSurface
+
+this_dir = os.path.dirname(__file__)
+base_dir = os.path.abspath(os.path.join(this_dir, ".."))
+sys.path.insert(0, base_dir)
+
+BUTTON_NAME_Y = 65
+BUTTON_NAME_X = 10
+
+COMMAND_Y = 20
+COMMAND_X = 10
+
+LEGEND = ['modifiers', 'layout', 'group', 'window', 'other']
+
+CUSTOM_KEYS = {
+    'Backspace': 2,
+    'Tab': 1.5,
+    '\\': 1.5,
+    'Return': 2.45,
+    'shift': 2,
+    'space': 5,
+    }
+
+
+class Button:
+    def __init__(self, key, x, y, width, height):
+        self.key = key
+        self.x = x
+        self.y = y
+        self.width = width
+        self.height = height
+
+
+class Pos:
+    WIDTH = 78
+    HEIGHT = 70
+    GAP = 5
+
+    def __init__(self, x, y):
+        self.x = x
+        self.row_x = x
+        self.y = y
+        self.custom_width = {}
+        for i, val in CUSTOM_KEYS.items():
+            self.custom_width[i] = val * self.WIDTH
+
+    def get_pos(self, name):
+        if name in self.custom_width:
+            width = self.custom_width[name]
+        else:
+            width = self.WIDTH
+
+        info = Button(name, self.x, self.y, width, self.HEIGHT)
+
+        self.x = self.x + self.GAP + width
+
+        return info
+
+    def skip_x(self, times=1):
+        self.x = self.x + self.GAP + times * self.WIDTH
+
+    def next_row(self):
+        self.x = self.row_x
+        self.y = self.y + self.GAP + self.HEIGHT
+
+
+class KeyboardPNGFactory:
+    def __init__(self, modifiers, keys):
+        self.keys = keys
+        self.modifiers = modifiers.split('-')
+        self.key_pos = self.calculate_pos(20, 140)
+
+    def rgb_red(self, context):
+        context.set_source_rgb(0.8431372549, 0.3725490196, 0.3725490196)
+
+    def rgb_green(self, context):
+        context.set_source_rgb(0.6862745098, 0.6862745098, 0)
+
+    def rgb_yellow(self, context):
+        context.set_source_rgb(1, 0.6862745098, 0)
+
+    def rgb_cyan(self, context):
+        context.set_source_rgb(0.5137254902, 0.6784313725, 0.6784313725)
+
+    def rgb_violet(self, context):
+        context.set_source_rgb(0.831372549, 0.5215686275, 0.6784313725)
+
+    def calculate_pos(self, x, y):
+        pos = Pos(x, y)
+
+        key_pos = {}
+        for c in '`1234567890-=':
+            key_pos[c] = pos.get_pos(c)
+
+        key_pos['Backspace'] = pos.get_pos('Backspace')
+        pos.next_row()
+
+        key_pos['Tab'] = pos.get_pos('Tab')
+        for c in 'qwertyuiop[]\\':
+            key_pos[c] = pos.get_pos(c)
+        pos.next_row()
+
+        pos.skip_x(1.6)
+        for c in 'asdfghjkl;\'':
+            key_pos[c] = pos.get_pos(c)
+        key_pos['Return'] = pos.get_pos('Return')
+        pos.next_row()
+
+        key_pos['shift'] = pos.get_pos('shift')
+        for c in 'zxcvbnm':
+            key_pos[c] = pos.get_pos(c)
+        key_pos['period'] = pos.get_pos('period')
+        key_pos['comma'] = pos.get_pos('comma')
+        key_pos['/'] = pos.get_pos('/')
+        pos.next_row()
+
+        key_pos['control'] = pos.get_pos('control')
+        pos.skip_x()
+        key_pos['mod4'] = pos.get_pos('mod4')
+        key_pos['mod1'] = pos.get_pos('mod1')
+        key_pos['space'] = pos.get_pos('space')
+        key_pos['Print'] = pos.get_pos('Print')
+        pos.skip_x(3)
+        key_pos['Up'] = pos.get_pos('Up')
+
+        pos.next_row()
+        pos.skip_x(12.33)
+        key_pos['Left'] = pos.get_pos('Left')
+        key_pos['Down'] = pos.get_pos('Down')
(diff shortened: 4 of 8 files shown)


zordsdavini:
updated arrow keys, fn keys

mod4__

pawamoy:
Awesome work @zordsdavini! This is a great visual help :slightly_smiling_face:

ramnes:
Can you also add whatever is needed in the documentation build process to generate these images and add them to a relevant page?

zordsdavini:
> Can you also add whatever is needed in the documentation build process to generate these images and add them to a relevant page?

I think I need to add `-c config_path.py` and `-o output_directory`. I'll think about documenting this in http://docs.qtile.org/en/latest/manual/config/index.html#testing-your-configuration or separate block in config/index and it would be awesome to generate/add images to documentation during doc build (but I'm not strong in this part) but I think this part should be done in separate task

zordsdavini:
@ramnes added documentation + included default config generation in docs html build target

tych0:
I think you can add `libqtile.backend.x11._ffi_xcursors` to the list of MockModules in docs/conf.py, and then hopefully it'll work?

zordsdavini:
It looks like this did't help. What about if we put img generation in
seperate process. Docs should already have static images and image
generation should be somewhere else but before deploy?

2020-04-19, sk 00:38, Tycho Andersen <notifications@github.com> rašė:

zordsdavini:
@tych0 the problem is that script calls for `__ffi_cursors`. It looks like I should revert all integration in docs build

zordsdavini:
is therre something more what I should do before it could be merged into master. I'd like to start on next task: default config update to support max and columns layouts

tych0:
Can we avoid committing the PNGs? As I understand it, they're products of the config file. What if we just add them to .gitignore and generate them each time the docs are built?

docs/Makefile
```diff
@@ -14,7 +14,11 @@ ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
-.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext coverage
+GENKEYIMG_SCRIPT = ../scripts/gen-keybinding-img -o ./_static/keybindings
+GENKEYIMG_REMOVE_IN_RST = sed -i '/^\.\. LS_PNG/,/^\.\. END_LS_PNG/{/^\.\. image/d}' manual/commands/keybindings.rst
+GENKEYIMG_INSERT_IN_RST = sed -i '/^\.\. END_LS_PNG/e find _static/keybindings/* | awk '"'"'{ print length, $$0 }'"'"' | sort -n | cut -d" " -f2- | awk '"'"'{print ".. image:: /" $$1}'"'"'' manual/commands/keybindings.rst
```
ramnes, line 19: Why do we need these two?

setup.py
```diff
@@ -64,6 +64,25 @@ def finalize_options(self):
...
+        def run(self):
+            subprocess.call(['scripts/ffibuild'])
+            subprocess.call(['make', '-C', 'docs', 'genkeyimg'])
+            super().run()
+
+    cmdClass['build_sphinx'] = BuildDocWrapper
+
+except ImportError as e:
+    print(e)
+    pass
```
ramnes, line 83: Why do we need all of this?

setup.py
```diff
@@ -89,7 +108,7 @@ def get_cffi_modules():
 
 
 setup(
-    cmdclass={'install': CheckCairoXcb},
+    cmdclass=cmdClass,
```
ramnes, line 111: Why changing this?

scripts/gen-keybinding-img
```diff
@@ -0,0 +1,424 @@
...
+                }
+
+
+        if text not in dictionary:
+            return text
+
+        return dictionary[text]
+
+
+
```
ramnes, line 315:
```suggestion
```

scripts/gen-keybinding-img
```diff
@@ -0,0 +1,424 @@
...
+                'Right': '→',
+                'Up': '↑',
+                'AudioRaiseVolume': 'Volume up',
+                'AudioLowerVolume': 'Volume down',
+                'AudioMute': 'Audio mute',
+                'AudioMicMute': 'Mic mute',
+                'MonBrightnessUp': 'Brightness up',
+                'MonBrightnessDown': 'Brightness down',
+                }
+
```
ramnes, line 307:
```suggestion
```

scripts/gen-keybinding-img
```diff
@@ -0,0 +1,424 @@
...
+        if fn:
+            key = key[4:]
+            context.set_font_size(10)
+        else:
+            context.set_font_size(14)
+
+        context.move_to(x + BUTTON_NAME_X, y + BUTTON_NAME_Y)
+        context.show_text(self.translate(key))
+
+    def show_multiline(self, context, x, y, key):
```
ramnes, line 257: What is this function doing? I think it might help to add a few docstrings when it's not 100% obvious (and most functions are obvious to me, but not this one).

scripts/gen-keybinding-img
```diff
@@ -0,0 +1,424 @@
...
+
+    def skip_x(self, times=1):
+        self.x = self.x + self.GAP + times * self.WIDTH
+
+    def next_row(self):
+        self.x = self.row_x
+        self.y = self.y + self.GAP + self.HEIGHT
+
+
+class kbPngFactory:
```
ramnes, line 77: This class name isn't very Pythonic, what do you think of something like this instead?

```suggestion
class KeyboardPNGFactory:
```

docs/Makefile
zordsdavini: this updates docs to list current images automatically

setup.py
zordsdavini: this during doc buildings add image generation before. As not all have all requirements to build docs - moved under `try` block

setup.py
zordsdavini: because of above - custom Sphinx doc build flow

scripts/gen-keybinding-img
zordsdavini: will refactor

scripts/gen-keybinding-img
zordsdavini: cairo doesn't have multiline, so I split lines to fit in button. Wrapping by words

setup.py
ramnes: I really don't get it; you just made it a variable, how is it changing anything?

docs/requirements.txt
```diff
@@ -2,3 +2,7 @@ sphinx
 sphinx_rtd_theme
 sphinxcontrib-seqdiag
 numpydoc
+
+cairocffi >= 0.9.0
+cffi >= 1.1.0
+xcffib >= 0.5.0
```
ramnes, line 8: IIRC we don't have this here because we have dev.sh

ramnes: (and because they need to be installed in a certain order, which a requirements file or pip itself can not guarantee)

docs/manual/commands/keybindings.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. don't delete LS_PNG and END_LS_PNG (it is used for `make genkeyimg`)
+.. LS_PNG
+.. image:: /_static/keybindings/mod4__.png
+.. image:: /_static/keybindings/mod4-shift__.png
+.. image:: /_static/keybindings/mod4-control__.png
+.. END_LS_PNG
+
+Generate own images
+===================
```
ramnes, line 16:
```suggestion
Generate your own images
========================
```

setup.py
zordsdavini: I expand it with Sphinx. Look at `cmdClass['build_sphinx'] = BuildDocWrapper`

docs/manual/commands/keybindings.rst
```diff
@@ -0,0 +1,37 @@
...
+
+The Qtile repo includes a tiny helper script to generate keybindings images from
+config file. Script located in ``scripts/gen-keybinding-img``.
+
+Script accepts config file or output directory. If nothing is given default
+config will be used and files should be placed in same directory where it has
+been run.
```
ramnes, line 23: Quick proof-reading, might not be perfect:

```suggestion
Qtile provides a tiny helper script to generate keybindings images from a
config file. In the repository, the script is located under
``scripts/gen-keybinding-img``.

This script accepts a configuration file and an output directory. If no
argument is given, the default configuration will be used and files will be
placed in same directory than where the command has been run.
```

docs/manual/commands/keybindings.rst
```diff
@@ -0,0 +1,37 @@
...
+
+::
+
+    usage: gen-keybinding-img [-h] [-c CONFIGFILE] [-o OUTPUT_DIR]
+
+    Qtile keybindings generator to nice files
```
ramnes, line 29: "nice files"? :)

docs/Makefile
ramnes: Are you saying that it adds these lines? https://github.com/qtile/qtile/pull/1631/files#diff-edea8d20248b32a41e85854cd1dc0defR10-R12

If so, why do we need to automate this? Can't we just keep the lines as is and update if/when needed?

docs/Makefile
```diff
@@ -38,11 +42,12 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
+	@echo "  genkeyimg  to generate keybindings images. Will save in _static"
```
ramnes, line 45:
```suggestion
	@echo "  genkeyimg  to generate keybindings images. Will save in _static/keybindings"
```

docs/Makefile
```diff
@@ -14,7 +14,11 @@ ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
-.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext coverage
+GENKEYIMG_SCRIPT = ../scripts/gen-keybinding-img -o ./_static/keybindings
```
ramnes, line 17:
```suggestion
GENKEYIMG_SCRIPT = ../scripts/gen-keybinding-img
```

docs/Makefile
```diff
@@ -157,3 +162,10 @@ coverage:
 	$(SPHINXBUILD) -b coverage $(ALLSPHINXOPTS) $(BUILDDIR)/coverage
 	@echo "Testing of coverage in the sources finished, look at the " \
 	      "results in $(BUILDDIR)/coverage/python.txt."
+
+genkeyimg:
+	${GENKEYIMG_SCRIPT}
```
ramnes, line 167:
```suggestion
	${GENKEYIMG_SCRIPT} -o ./_static/keybindings
```

docs/manual/commands/keybindings.rst
```diff
@@ -0,0 +1,37 @@
...
+=====================
+
+Default configuration
+=====================
+
+.. don't delete LS_PNG and END_LS_PNG (it is used for `make genkeyimg`)
+.. LS_PNG
+.. image:: /_static/keybindings/mod4__.png
+.. image:: /_static/keybindings/mod4-shift__.png
+.. image:: /_static/keybindings/mod4-control__.png
```
ramnes, line 12: Why do the files end with __?

setup.py
ramnes: Alright, I get it now!

Do we really want to build the docs when installing Qtile with pip? I think we should rather exclude them, and only have them installed by proper package managers, and the package maintainers will decide what to do with the docs.

I personally see these docs as something that is meant to be hosted online, not as something that should take bytes on every user's computer disk. What do you think?

ramnes: Also, that would make the pip install longer and more prone to errors in case the doc does not build, and I think we don't want this.

docs/requirements.txt
zordsdavini: but Travis doesn't use `dev.sh` or I miss something?

docs/manual/commands/keybindings.rst
zordsdavini: -> Qtile keybindings image generator

docs/Makefile
zordsdavini: the directory is hardcoded and later it is used in other scripting parts (next lines). Why it would be  better to move? If user will want to have its own images, he will run script and not `make`

docs/Makefile
zordsdavini: same as above

docs/manual/commands/keybindings.rst
zordsdavini: The name is made from modifiers, so if no modifiers - script should add something

docs/requirements.txt
ramnes: Mhhh, well, it uses what you want it to use? Right now it uses tox, which itself handles its dependencies with `deps` in each environment definition. But what is the relation between this PR and Travis?

setup.py
zordsdavini: Tycho asked to include it in docs build. But how documentation building is related to `pip install qtile`? This cmdclass is just commands to build. If you run simple install - it will not generate docs, just with `python setup.py build_sphinx` (if I not mistype something) and I'm fixing that part

docs/Makefile
zordsdavini: yes, it adds these image source lines.
You want to say that manual is better then automatically? :) Then why you are using tiling wm? :D
I think that manual intervention is place to forget.

docs/requirements.txt
zordsdavini: yes, tox. But Travis builds failed without that :)

docs/Makefile
ramnes: It felt more logical to me, but now that I'm looking at it again, why not just put everything into the `genkeyimage` task? I don't see why we would want these three lines to be variables of the Makefile.

docs/Makefile
ramnes: But these lines are never going to change, right? I'm also using a tiling WM because it's minimalist and doesn't do too much. :wink:

docs/manual/commands/keybindings.rst
ramnes: Let's add something like

```
if not modifier:
    filename = "base.png"
else:
    filename = "{}.png".format(modifier)
```

:)

setup.py
ramnes: Alright, I get it now! (second edition)

I didn't know about that `build_sphinx` thing, it's more clear now that I read about it on the Sphinx documentation.

setup.py
```diff
@@ -64,6 +64,25 @@ def finalize_options(self):
...
 
+cmdClass = {'install': CheckCairoXcb}
+
+try:
+    from sphinx.setup_command import BuildDoc
+    import subprocess
+
+    class BuildDocWrapper(BuildDoc):
+        def run(self):
+            subprocess.call(['scripts/ffibuild'])
```
ramnes, line 75: Why do we need this? I guess that the genkeyimg task needs `scripts/ffibuild` to have been executed? If so, I'd rather make this as another task in the Makefile and depend on it there, no?

setup.py
```diff
@@ -64,6 +64,25 @@ def finalize_options(self):
...
+            subprocess.call(['make', '-C', 'docs', 'genkeyimg'])
```
ramnes, line 76: Why not doing everything so that we don't need the `-C docs`?

zordsdavini: We hold `docs` separately. All logics here. I think it should leave here as it is

setup.py
zordsdavini: I don't remember why I chouse that path, but we need working `configreader`. As it doesn't change logics, I'd like faster to deploy and then later refactor if we will see that we need `ffbuild` somewhere else

docs/manual/commands/keybindings.rst
zordsdavini: refactored to `no_modifier.png` and `modifier + '.png'`

docs/requirements.txt
ramnes: Do you have a link to a failed Travis build due to this?

zordsdavini: It is hard to find... Easier to create new build without these lines. Will give you soon. (maybe it is not needed anymore, because I changed a lot)

docs/Makefile
zordsdavini: moved under `genkeyimage:`

docs/requirements.txt
zordsdavini: fail https://travis-ci.org/github/qtile/qtile/jobs/680551947

docs/Makefile
```diff
@@ -157,3 +158,12 @@ coverage:
 	$(SPHINXBUILD) -b coverage $(ALLSPHINXOPTS) $(BUILDDIR)/coverage
 	@echo "Testing of coverage in the sources finished, look at the " \
 	      "results in $(BUILDDIR)/coverage/python.txt."
+
+genkeyimg:
+	mkdir -p ./_static/keybindings
```
tych0, line 163: We have both a .gitkeep for this and this mkdir -p; can we just get rid of the gitkeep?

zordsdavini: removed .gitkeep

docs/requirements.txt
ramnes: So yeah, I guess it should work to just add cairocffi as deps in the tox environment for docs?

zordsdavini: The begining of this history is with some lessons :) I've started building documentation as described in docs/BUILDING. So created that environment, got issues with sphinx doc building because it self searched virtenv and found warnings. Then this cairocffi issue. Added all dependencies and got documentation. I didn't ues tox locally (or I don't know that I use it :D), so these dependencies will not brake something (now)

ramnes: I'm really sorry, this whole discussion is due to a misreading and a bad memory. I'll do a last suggestion and then have some rest, because apparently I need some. :)

ramnes, line 8:
```suggestion
-r ../requirements.txt
```

setup.py
ramnes: Not sure to understand but it's no big deal :)

setup.py
ramnes: sounds fair

docs/requirements.txt
zordsdavini: nice :)

ramnes:
Let's go!

Issue #1409, opened 2019-08-28
libqtile.command.Client doesn't exist anymore


ramnes:
How has it been replaced? The docs [here](http://docs.qtile.org/en/latest/manual/commands/scripting.html) need an update.

pawamoy:
And [here](http://docs.qtile.org/en/latest/manual/commands/index.html) as well.

pawamoy:
I was able to instantiate a client with this code:

```python
from libqtile.ipc import find_sockfile, Client
from libqtile.command_client import InteractiveCommandClient
from libqtile.command_interface import IPCCommandInterface

sock_file = find_sockfile()
ipc_client = Client(sock_file)
cmd_object = IPCCommandInterface(ipc_client)
cmd_client = InteractiveCommandClient(cmd_object)
cmd_client.status()
```

Or as a one-liner:

```python
c = InteractiveCommandClient(IPCCommandInterface(Client(find_sockfile())))
c.status()
```
```

**Comment by pawamoy** [comment]
Maybe a simple fix would be to allow `None` for the `command` argument of `InteractiveCommandClient`:

```python
from libqtile.ipc import find_sockfile, Client
from libqtile.command_interface import IPCCommandInterface

class InteractiveCommandClient:
    def __init__(self, command: CommandInterface, current_node: GraphType = None) -> None:
        if not command:
            command = IPCCommandInterface(Client(find_sockfile()))
        self._command = command
        ...
```

And update the docs to

```python
from libqtile.command_client import InteractiveCommandClient as Client

c = Client()
c.status()
```

ramnes:
You would need to default to `command=None` in your example but apart from that LGTM. @flacjacket any thoughts on this?

pawamoy:
Ah yes, forgot the default to `None`.

tych0:
I think with the two PRs above we can close this, so I'll go ahead and do that. Please re-open (or send patches :D) if we find other stuff that needs updated.

PR #1636, opened 2020-04-18
Fix use of CommandClient in docs

pawamoy:
For issue #1409, part 2.

Not sure if we should tell about `CommandClient` or `InteractiveCommandClient` (I don't know the difference).


tych0:
Thanks for these!

PR #1639, opened 2020-04-18
Makefile: some basic cleanups

tych0:
We've migrated to tox for linting/testing, and now that we do a bunch of
additional linting on imports, etc., we should probably use the same thing
for the make targets that we do for CI. (Am I the only one that uses the
Makefile? :D)

We can drop ckpatch, as that was originally the entry point for travis. I
just tend to run individual things manually.

Also, let's clean up some of the new coverage reports that tox generates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,26 +8,15 @@ default:
 
 .PHONY: check
 check:
-	pytest --verbose
-
-.PHONY: check-cov
-check-cov:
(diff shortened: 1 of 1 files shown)


pawamoy:
I tend to use make over tox in local. I usually avoid duplication, and would typically call tox from the Makefile, and this is exactly what you did here :+1:

tych0:
Cool, I'll keep the makefile around then!

Issue #1352, opened 2019-06-21
0.14.2: test/widgets/test_{battery,volume}.py, test/test_images{,2}.py fail


dvzrv:
During packaging 0.14.2 for Arch Linux, I'm again attempting to run the test suite. While it still hangs (#1130) in general, I have the following failures, some of which I have observed before:

[qtile-0.14.2-check.txt](https://github.com/qtile/qtile/files/3313999/qtile-0.14.2-check.txt)

Quite a few of them fail because of `test/data/svg/audio-volume-muted.svg`, while the other variants of the same test work just fine!

```
test/test_images.py::test_get_cairo_surface[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 16%]
...
test/test_images.py::TestImg::test_init[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 17%]
...
test/test_images.py::TestImg::test_from_path[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 19%]
...
test/test_images.py::TestImg::test_pattern[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 21%]
...
test/test_images.py::TestImg::test_pattern_rotate[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 23%]
...
test/test_images2.py::test_svg_scaling[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] FAILED [ 28%]
...
test/widgets/test_battery.py::test_images_good FAILED                    [ 98%]
...
test/widgets/test_volume.py::test_images_good FAILED                     [100%]
```

This is the failures output:

```
=================================== FAILURES ===================================
_ test_get_cairo_surface[/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg] _

path_n_bytes_image = ('/build/qtile/src/qtile-0.14.2/test/data/svg/audio-volume-muted.svg', b'<svg xmlns="http://www.w3.org/2000/svg" viewB...508 3.508 -3.508 -1.492 -1.492 -3.508 3.508 z" style="fill:#ececec;opacity:0.4;fill-opacity:1;color:#000"/>\n</svg>\n')

    def test_get_cairo_surface(path_n_bytes_image):
[...]

PR #1643, opened 2020-04-19
images: don't swallow image decoding errors

tych0:
In #1352, we see the test failing like:

    def get_cairo_surface(bytes_img, width=None, height=None):
        try:
            surf = cairocffi.ImageSurface.create_from_png(io.BytesIO(bytes_img))
            return _SurfaceInfo(surf, 'png')
        except (MemoryError, OSError):
            pass
        try:
            return _decode_to_image_surface(bytes_img, width, height)
        except cairocffi.pixbuf.ImageLoadingError:
            pass
>       raise LoadingError("Couldn't load image!")
E       libqtile.images.LoadingError: Couldn't load image!

but that exception gives us exactly zero useful info about why image loading
failed. It looks to me like the first try is a try at decoding a PNG before we
decode the SVG, so we could/should probably do that better so we don't swallow
unnecessary exceptions. But at least this will help us not swallow the
exception from that test.

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -53,11 +53,7 @@ def get_cairo_surface(bytes_img, width=None, height=None):
         return _SurfaceInfo(surf, 'png')
     except (MemoryError, OSError):
         pass
-    try:
-        return _decode_to_image_surface(bytes_img, width, height)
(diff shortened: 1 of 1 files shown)


ramnes:
another no brainer here

PR #1644, opened 2020-04-19
session manager: set_event_loop() to our event loop

### LINKED ISSUE #1046: Can't use asyncio event loop in widgets
I am creating a widget that uses asyncio to run some external command (with `asyncio.create_subprocess_exec`). It doesn't work, and raises the `RuntimeError("Cannot add child handler, the child watcher does not have a loop attached")` exception instead.

If my understanding of the code is correct, calling `set_event_loop` after `new_event_loop` should fix this issue, but I'm not sure whether it will cause other problems.

tych0:
This will enable widgets to do things using our event loop if they want.

Closes #1046

### DIFF
--- a/libqtile/core/session_manager.py
+++ b/libqtile/core/session_manager.py
@@ -25,6 +25,7 @@ class SessionManager:
             The state to restart the qtile instance with.
         """
         eventloop = asyncio.new_event_loop()
+        asyncio.set_event_loop(eventloop)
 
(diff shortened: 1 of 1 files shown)


ramnes:
No brainer :)

flacjacket:
Nice! We should see if we can't get a more async driven eventloop threaded throughout to make this even easier.

PR #1645, opened 2020-04-19
libqtile/lazy: fix 'when' filter

### LINKED ISSUE #1629: lazy.when
# Issue description

I found this quite oldy [`config.py`](https://gist.github.com/cjbarnes18/4151805) that uses a `when` operation to limit the keybinding to a layout but it throws an exception. I also couldn't find it in the docs, so is it supposed to exist? Is there another similar intended way to limit key bindings to a layout.

```python
```
...
```python
    Key([mod], 'Left',
        lazy.layout.left().when('xmonadtall'))
```

# Qtile version

v0.15.0

# Stack traces

```
libqtile xcore.py:_xpoll():L277  Got an exception in poll loop
  File "/share/git/qtile/libqtile/backend/x11/xcore.py", line 246, in _xpoll
    ret = target(event)
  File "/share/git/qtile/libqtile/backend/x11/xcore.py", line 494, in handle_KeyPress
    self.qtile.process_key_event(keysym, event.state & self._valid_mask)
  File "/share/git/qtile/libqtile/core/manager.py", line 342, in process_key_event
    if cmd.check(self):
AttributeError: 'NoneType' object has no attribute 'check'
```

Thanks for any help

airwoodix:
Fixes #1629

Was apparently broken during refactoring in 877e684 and 85f95a5. Tested with Xephyr, seems to work. A test case would be nice but I can't do this now. Merging this would at least bring back the functionality.

### DIFF
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -72,6 +72,7 @@ class LazyCall:
     def when(self, layout=None, when_floating=True):
         self._layout = layout
         self._when_floating = when_floating
+        return self
 
     def check(self, q) -> bool:
         if self._layout is not None:


ramnes:
Thanks!

PR #1648, opened 2020-04-19
fix image test

tych0:
330f0434acf5 ("images: don't swallow image decoding errors") changed the
exception this throws, but didn't update the test. Let's do that.


flacjacket:
Nice!

PR #1649, opened 2020-04-20
Move command object functionality from Qtile manager

flacjacket:
Set the command object functionality from the qtile manager into a
separate dedicated file.  This allows for the functionality of the
manager to be separated from the command tree integration.

### DIFF
--- a/libqtile/core/command_root.py
+++ b/libqtile/core/command_root.py
@@ -1,0 +1,805 @@
+# Copyright (c) 2008, Aldo Cortesi. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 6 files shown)


flacjacket:
Just as a note, there are still a few places where `qtile.cmd_*` are called here that will need to be cleaned up before this can land.

libqtile/core/command_root.py
```diff
@@ -0,0 +1,805 @@
...
+    finally:
+        if fp:
+            fp.close()
+    return module
+
+
+class CommandRoot(CommandObject):
+    """This object is the `root` of the command graph"""
+
+    def __init__(self, qtile: "Qtile") -> None:
```
tych0, line 116: What does the string "Qtile" do here?

flacjacket: This is used for type checking, to specify that this is a Qtile manager object. We can't import `Qtile` directly here, because the `Qtile` instance creates this object, so we'd get a circular reference. The type checkers have a way to get around this with the `typing.TYPE_CHECKING` flag to make this import only when running type checking. In older versions, the type annotation objects need to be imported even when running, the way to get around this is making the type annotation a string. In Python 3.8 (and 3.7 with `from __future__ import annotations`), the type annotations are by default interpreted as strings, so this can just specify the type.

PR #1650, opened 2020-04-20
bars should use initial_size when (re)configuring window size

m-col:
I came across a bug where if a bar that has margins is re-configured, the size is increased each time (to add the margins to its `self.size`) and upon reconfiguring, the height takes its value from the now greater size. This creates a black window growing out of the bar that grows each time they are re-configured. These changes make the height get its value from the initial size, which never changes, and stops the size from growing more than once.


ramnes:
Thanks!

Issue #1646, opened 2020-04-19
lazy/check: broken logic


airwoodix:
# Issue description

I was too quick in testing for #1645. The patch fixes the reported issue (#1629) but [the logic](https://github.com/qtile/qtile/blob/8d5c17354cbfa58cd0b4e58b3cbe245e225fb66f/libqtile/lazy.py#L77) in ``LazyCall.check`` seems broken to me. Sorry for the noise.

The code reads:

```python
    def check(self, q) -> bool:
        if self._layout is not None:
            if self._layout == 'floating':
                if q.current_window.floating:
                    return True
                return False
            if q.current_layout.name != self._layout:
                if q.current_window and q.current_window.floating and not self._when_floating:
                    return False
        return True
```

which implies that the call is evaluated even if ``q.current_layout.name != self._layout`` when e.g. the current window is not floating.

The [old logic](https://github.com/qtile/qtile/blob/a3ae3c623859b247813fc1876b188c4d40e84df0/libqtile/command.py#L293) looks better but has an edge case when the current layout is ``floating`` but there's no window any more (access ``.floating`` on ``q.current_window = None``). This is also the case in the code above.

ramnes:
Can you describe why it is broken, exactly? What is it that you try, what is the expected result, and what do you get?

Sorry if that's obvious but we have numerous issues to deal with and it's hard to give time for each one of them when it's not ELI5.

airwoodix:
I updated the the issue description above.

PR #1651, opened 2020-04-20
Fix libqtile/lazy:LazyCall.check, add type hints and doc to LazyCall.when

airwoodix:
Follow-up to: #1645  add type annotations and documentation to ``LazyCall.when``
Fixes: #1646 simplify and fix logic according to issue reported.

Both changes are in the same PR because the documentation for ``LazyCall.when`` depends on the behavior of ``LazyCall.check``.

The proposed logic sticks to the [previously working one](https://github.com/qtile/qtile/blob/a3ae3c623859b247813fc1876b188c4d40e84df0/libqtile/command.py#L293), only simplifying the code. Some discussion points:

  - the check for ``self._floating == 'floating'`` is pretty restrictive. If there are no windows left but the last one was floating, the current layout is still ``floating`` but the call will not be evaluated. This might be surprising because ``lazy.something.when(layout="floating")`` does not seem to assume that there has to be a floating window;

  - the previous point also conflicts with ``when_floating`` (it ignores the latter). This is probably unproblematic since the following is weird: ``lazy.something.when(layout="floating", when_floating=False)``. The only use case is that of floating layout with no window, as previously;

  - AFAICT, there's no support for restricting a command to multiple layouts (chaining ``.when``s would overwrite ``self._layout`` wouldn't it?). Extending ``layout`` to ``Union[List[str], None]`` or even ``Union[List[str], str, None]`` is relatively straightforward (the latter would not break legacy code). One could also stick to ``Optional[str]`` with a comma separated list of layouts as parameter. Relevance? Opinions?

### DIFF
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -69,7 +69,19 @@ class LazyCall:
         """The kwargs to the given call"""
         return self._kwargs
 
-    def when(self, layout=None, when_floating=True):
+    def when(self, layout: Optional[str] = None,
+             when_floating: bool = True) -> 'LazyCall':
+        """Filter call activation per layout or floating state
+
+        Parameters
+        ----------
+        layout : str or None
+            Restrict call to given layout name. If None, call for all layouts.
+            If 'floating', call if, and only if the current window is floating,
+            ``when_floating`` is ignored in this case.
+        when_floating : bool
+            Call if the current window is floating.
+        """
         self._layout = layout
         self._when_floating = when_floating
         return self
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -75,14 +75,17 @@ class LazyCall:
         return self
 
     def check(self, q) -> bool:
-        if self._layout is not None:
-            if self._layout == 'floating':
-                if q.current_window.floating:
-                    return True
-                return False
-            if q.current_layout.name != self._layout:
-                if q.current_window and q.current_window.floating and not self._when_floating:
-                    return False
+        cur_win_floating = q.current_window and q.current_window.floating
+
+        if self._layout == 'floating':  # ignore _when_floating
+            return cur_win_floating
+
+        if cur_win_floating and not self._when_floating:
+            return False
+
+        if self._layout and q.current_layout.name != self._layout:
+            return False
+
         return True
 
 
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -76,18 +76,13 @@ class LazyCall:
         Parameters
         ----------
         layout : str, Iterable[str], or None
-            Restrict call to given layout names. If None, enable call
-            for all layouts.
-            Can be a single name, a string of comma-separated names, or any
-            iterable yielding layout names.
+            Restrict call to one or more layouts.
+            If None, enable the call for all layouts.
         when_floating : bool
             Enable call when the current window is floating.
         """
         if layout is not None:
-            if isinstance(layout, str):
-                self._layouts = set(l.strip() for l in layout.split(","))
-            else:
-                self._layouts = set(layout)
+            self._layouts = {layout} if isinstance(layout, str) else set(layout)
 
         self._when_floating = when_floating
         return self
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import Dict, List, Optional, Tuple, Union  # noqa: F401
+from typing import Dict, Iterable, List, Optional, Set, Tuple, Union  # noqa: F401
 
 from libqtile.command_client import InteractiveCommandClient
 from libqtile.command_graph import (
@@ -46,7 +46,7 @@ class LazyCall:
         self._args = args
         self._kwargs = kwargs
 
-        self._layout = None  # type: Optional[str]
+        self._layouts = set()  # type: Set[str]
         self._when_floating = True
 
     @property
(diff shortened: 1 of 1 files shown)


tych0:
This discussion is great and the fix looks fine, any chance you can put some of the discussion in the commit logs instead of just the PR message?

airwoodix:
Thanks for the feedback.

  - I think that the first point in the discussion should be resolved. I see no reason to handle the ``floating`` layout in a different way.

  - the 2nd point comes from the special handling in the 1st point. Let's remove that.

  - chaining calls to when (e.g. ``lazy.cmd.when(layout="max").when(layout="columns", when_floating=False)``) seems the most pythonic option for me for the 3rd point.

I'd like to suggest the following behavior:

```python
 # evaluate command only in "columns" and "monadtall" layouts
 cmd.when(layout="columns").when(layout="monadtall")

 # same as above, but do not evaluate if the
 # current window is floating in "columns" layout
 cmd.when(layout="columns", when_floating=False)\
     .when(layout="monadtall")

 # never evaluate if the current window is floating
 cmd.when(when_floating=False)

 # evaluate only in group "a"
 cmd.when(filt=lambda q: q.current_group.name == 'a')

 # evaluate only in "columns" layout when in group "a"
 # and in "monadtall" layout (any group)
 cmd.when(layout="columns",
          filt=lambda q: q.current_group.name == 'a') \
     .when(layout="monadtall")
```

Rationale / discussion:

  - the most common use case is to filter per layout. This is exposed directly. Other filters can be implemented using  the ``filt`` predicate on the command object (``Qtile`` I believe);

  - in principle, the whole functionality (including the old behavior) can be obtained with e.g. ``cmd_function``, doing nothing if the predicates are not satisfied. ``LazyCall.when`` is essentially a convenience function (that's probably why it was broken, power-users use ``cmd_function``);

  - this does not provide a way to dispatch between different actions for a given key combination depending on context. This is also feasible with ``cmd_function``. Maybe a ``LazyCall.dispatch`` method would be nice? (but that's definitely out of the scope of this PR)

  - when a layout is specified, ``when_floating`` and ``filt`` become layout specific;

  - except for special handling of ``layout='floating'``, old configs should behave correctly;

  - I have some test code [here](https://github.com/airwoodix/qtile/blob/enh-lazy-when/libqtile/lazy.py#L74) (airwoodix/qtile@4e8e2f0) that implements this above functionality. The implementation is relatively easily extensible to other "easy-access" filters like e.g. groups.

I can add the above mentioned patchset to this PR if wanted. Otherwise, I will at least extend the comments/docstrings to get a trace of the discussion here.

ep12:
I personally think that chaining `.when` is less pythonic than calling it with a `str` or `set` as the layout argument:
```python
def when(self, layout: Optional[Union[Iterable[str], str]] = None, ...):
    self._layout = (({layout} if isinstance(layout, str) else set(layout))
                    if layout is not None else set())
    ...
```
Also, for me it is a bit annoying to always have to use `lazy.function`. I might start a new pr when this one is merged...

ramnes:
I would expect chaining `when` to apply all conditions, not to do an "OR" between them so yeah, I do agree with @ep12.

> Also, for me it is a bit annoying to always have to use lazy.function. I might start a new pr when this one is merged...

Yisss, I've always felt this was something we could eventually get rid of, let's try it!

ep12:
@ramnes I don't want to get rid of `lazy.function` but I want to simplify the configuration of _per program_ keybindings, let's say. Or a bit generalised: conditional actions, so I already had the `.when` in mind. To avoid merge conflicts, I'll wait before actually trying to implement something. But I'll open an issue now...

ramnes:
Ah. :smile:

ep12:
See #1661

airwoodix:
Thanks for the comments. #1661 looks interesting so let's keep this PR short.

Fair enough for the chaining. I implemented the ``Union[Iterable[str], str]`` signature, also supporting comma-separated names. I think that configs benefit a lot from lighter syntax. (i.e. ``when(layout="max,columns")`` vs ``when(layout={"max", "columns"})``).

Is the docstring understandable?

Lint gave a ``I001`` (isort import in wrong position) error for the ``typing`` imports. Any idea what's wrong with that line?

ep12:
[isort](https://pypi.org/project/isort/) gave me this:
```python
from typing import Dict, Iterable, List, Optional, Set, Tuple, Union  # noqa: F401
from libqtile.command_client import InteractiveCommandClient
from libqtile.command_graph import CommandGraphCall, CommandGraphNode, SelectorType
from libqtile.command_interface import CommandInterface
```
I don't get it why the CI wants the imports to be sorted alphabetically...

pawamoy:
Because we recently added  isort checks to the CI.

But I don't see what's wrong here :confused:

ramnes:
Just run isort:

```diff
-from typing import Dict, Iterable, List, Optional, Set, Tuple, Union  # noqa: F401
+from typing import (  # noqa: F401
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
```

ramnes:
> I don't get it why the CI wants the imports to be sorted alphabetically...

When you have a lot of imports it's much simpler to look for a specific import at glance if it's sorted. :wink:

ep12:
Yeah, but maybe a module provides some functions and classes where some of them are related closely to another one (commonly used together). I don't think alphabetic ordering should be enforced or preferred over thematic sorting. It really depends on the programmer... I think alphabetic sorting can be disabled on its own, because the stdlib - third party - custom order is really the important point... However, if you want to enforce a really consistent coding style, you can leave that enabled. But this might also mean that people's PRs fail the checks because of this unspecific lint error and then they have to waste time figuring that one out... But that's just my opinion :sun_with_face:

pawamoy:
@ep12 I'll add a `format` rule to the makefile and its related tox env. There's already a `lint` rule (again, calling a tox env) to check if everything's fine. I guess the use of these rules should be emphasized in the "Hacking on Qtile" or documents about contributing to Qtile :slightly_smiling_face: 

The next step would be enforcing it locally with `pre-commit` though that's another story.

ramnes:
We're off-topic, let's not argue here. Please raise your concerns about imports ordering in #1625 or in a new issue.

airwoodix:
> Just run isort:

Thanks. The problem is that the line is too long with the F401 (module imported but unused) ignoring comment. The imports _were_ sorted alphabetically. For some reason, `isort` doesn't catch this correctly. Since we don't need to disable F401 here (all typing imports are used in type hints), the last commit has the comment removed.

libqtile/lazy.py
```diff
@@ -69,20 +69,38 @@ def kwargs(self) -> Dict:
...
+    def when(self, layout: Optional[Union[Iterable[str], str]] = None,
+             when_floating: bool = True) -> 'LazyCall':
+        """Enable call only for given layout(s) and floating state
+
+        Parameters
+        ----------
+        layout : str, Iterable[str], or None
+            Restrict call to given layout names. If None, enable call
+            for all layouts.
+            Can be a single name, a string of comma-separated names, or any
```
flacjacket, line 81: As long as you are allowing iterables of names, I'd say to drop the parsing of comma separated names.

airwoodix: Is there a drawback supporting both?
The syntax with comma separated names is lighter and can therefore be more attractive to some. Fair enough, people could call ``layout="max,columns".split(",")`` themselves but it doesn't cost much to have this on the library side. Or does it?

ramnes: `layout="max,columns"` is only four characters lighter than `layout=("max", "columns")` (five if you count the space) so it doesn't seem like a useful shortcut. Let's use real Python structures and keep a clean interface.

airwoodix: Fair enough, fixed by a1d2fce.

airwoodix:
@ramnes @tych0 any update on merging this patch? or do you want to hold / discard in favor of other discussions targeting the same subsystem? (e.g. #1661)

tych0:
Can you add a test to make sure it doesn't break again? Otherwise lgtm.

brandonwillard:
Any chance this could get merged soon?  I can't seem to make layout-specific keys work without one of the `check` logic fixes in this PR.

If not, would you folks be willing to merge a small fix for only the `if q.current_layout.name != self._layout` condition?

tych0:
Yeah, a fix is better than nothing. But without a test, no guarantee this keeps working :). I'll merge it for now, if someone can send a test, that would be much appreciated.

PR #1652, opened 2020-04-20
Enh: widget/maildir: conditional colored display, custom subfolder format, hide when empty

airwoodix:
This patch is a proposal for enhancement of the ``Maildir`` widget. It adds support for the following options:

  - ``subfolder_fmt``: custom format string for a single subfolder. Defaults to ``{label}: {value}`` to match current behaviour;
  - ``empty_color`` and ``nonempty_color``: foreground colors applied to the formatted string depending on the whether the subfolder (sum of subfolders if ``total=True``) is empty (i.e. no new mail) or not. Both default to ``None``, using the default;
  - ``hide_when_empty``: flag to disable the display of the subfolder(s) if it is empty.

The patch also adds type hints for methods in ``widget.maildir.Maildir``.

Discussion:
  - ``_TextBox.foreground`` and ``_TextBox.background`` already manage foreground and background color for the whole widget but independently from the maildir state. This patch allows conditional colors depending on the most basic status of the maildir;
  - this is a minimal extension: more advanced/flexible options would be possible. For example: apply color only to value (or defined by ``subfolder_fmt``), define multi-color thresholds (less crude than empty or not), adjusting the background color, font weight, etc. However, all these enter the area of templating and are always possible by subclassing ``widget.Maildir`` in the user's own config. The balance here is to provide more options to the average user;
  - ``subfolder_fmt`` can be used to hide labels if wanted, which is not supported by ``_TextBox.fmt``;
  - a natural extension is to support similar options for ``GmailChecker`` and ``ImapWidget``.

Opiniated feedback is most welcome.

### DIFF
--- a/libqtile/widget/maildir.py
+++ b/libqtile/widget/maildir.py
@@ -61,7 +61,7 @@ class Maildir(base.ThreadedPollText):
                 for folder in self.sub_folders
             ]
 
-    def poll(self) -> str:
+    def poll(self):
         """Scans the mailbox for new messages
 
         Returns
--- a/libqtile/widget/maildir.py
+++ b/libqtile/widget/maildir.py
@@ -26,6 +26,7 @@
 
 import mailbox
 import os.path
+from typing import Dict
 
 from libqtile.widget import base
 
@@ -60,7 +61,7 @@ class Maildir(base.ThreadedPollText):
                 for folder in self.sub_folders
             ]
 
-    def poll(self):
+    def poll(self) -> str:
         """Scans the mailbox for new messages
 
         Returns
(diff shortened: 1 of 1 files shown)


ramnes:
Looking good, but can you fix mypy?

airwoodix:
> Looking good, but can you fix mypy?

Yes/no. Apparently, mypy has a problem with ``mailbox.Maildir``. Removing the return type annotation on ``poll`` (82312e3) makes mypy happy but that's not very satisfactory.

tych0:
Can you put this discussion in the commit message and squash it? lgtm otherwise

ramnes:
I did it myself to move forward. Thanks @airwoodix!

airwoodix:
Thanks @ramnes @tych0

PR #1660, opened 2020-04-21
Remove hooks that aren't fired anywhere

ramnes:


tych0:
+1 for negative diffs :)

PR #1664, opened 2020-04-22
Fix config validation and move the cache invalidation logic

ramnes:
See commit messages for rationale.

Ping @pawamoy


pawamoy:
Looks good :+1:!

PR #1666, opened 2020-04-23
Add format recipe

pawamoy:
- This adds a `format` tox env and makefile rule.
- `make format` actually runs `TOXENV=format tox` to be consistent with the other rules

For #1625

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ check:
 
 .PHONY: lint
 lint:
-	TOXENV=pep8 tox
+	TOXENV=format,pep8 tox
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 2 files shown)


ramnes:
It sounds weird to me to use tox here.

pawamoy:
Using tox makes it [more easily] cross-platform, but this might not be a valid point here indeed.

Makefile
```diff
@@ -10,6 +11,10 @@ default:
 check:
 	TOXENV=py38 tox
 
+.PHONY: format
+format:
+	TOXENV=format tox
```
tych0, line 16: Can we just add this to the "lint" target? In my mind they're basically the same thing.

pawamoy: Yes we could do that. Contributors would just run `make lint` to format the code and check it.

ramnes: Good idea, that would probably be the best of both worlds!

Issue #1521, opened 2020-01-09
Extending notify.py to draw unmanaged windows for notifications


m-col:
[notify.py](https://github.com/qtile/qtile/blob/master/libqtile/notify.py) appears to provide a org.freedesktop.Notifications service, which is only used by the notify widget. It's not clear to me how I can use the service in other ways, or if I can at all. It would be cool if it could show `libqtile.window.Internal` windows or similar to the bar (i.e. unmanaged) as little notification windows. Then notify.py could have a small function to pass the service notifications that could be used in scripting, widgets, error messages and potentially many other places in qtile. What do you think?

ramnes:
Sorry but I'm not sure to understand what you want to achieve. What's the concrete use case?

m-col:
What is the purpose of `libqtile.notify`? I can't find any info in the docs. It is used only by the notify widget, but I figured by it being implemented outside of the widget that it was written with other uses in mind.

My idea is to glue together existing code from `libqtile.notify` and `libqtile.window` to create a graphical notifications service to show pop-up notifications much like dunst or similar.

m-col:
I played around with this and created a 95% fully function graphical notification service. The code for it is [here](https://github.com/m-col/qtools/blob/master/qtools/notification/notification.py). It uses pretty much only libqtile to handle the windows. I've been using it instead of dunst for about a month and its worked really well. Here is a screenshot:

![screenshot](https://user-images.githubusercontent.com/44536113/77566907-4513f980-6ebe-11ea-9e29-12a4dac31e46.png)

For the notification service I did have to slightly [extend libqtile.notifier](https://github.com/m-col/qtile/commit/7ef41cc2123193180118a97c6b0d28a63eb8379b) so it would support all fields that notifications might supply.

It uses a base class called `Popup` ([here](https://github.com/m-col/qtools/blob/master/qtools/__init__.py#L80)), which generalises simple qtile 'internal' windows and could be useful for things like widget tooltips ([discussed here](https://github.com/qtile/qtile/issues/1441)), or maybe even to inform the user of things like config errors. If there would be interest in having this within libqtile so it can be improved and used for other windows let me know and I can create a PR, but understandable if it is outside of scope.

pawamoy:
@m-col this is great!

@ramnes @tych0 what do you think about this? I want to start working on https://github.com/qtile/qtile/issues/1449 and this would allow us to avoid a dependency to a third-party notification program like `notify-send`.

m-col:
[...]

PR #1667, opened 2020-04-23
notifier will keep all information with notifications

m-col:
This extends `libqtile.notifier` slightly to retain all useful information that it receives via dbus for notifications.

The notifier now also keeps track of its capabilities, which notification services should report to dbus when asked; see https://developer.gnome.org/notification-spec. Capabilities can be updated by other code that plugs into the notifier e.g. a graphical notification handler should update this to reflect what it data it is capable of using.

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -48,15 +48,24 @@ if has_dbus:
             bus_name = service.BusName(BUS_NAME, bus=bus)
             service.Object.__init__(self, bus_name, SERVICE_PATH)
             self.manager = manager
+            self._capabilities = {'body'}
 
         @service.method(BUS_NAME, in_signature='', out_signature='as')
         def GetCapabilities(self):  # noqa: N802
-            return ('body')
+            return list(self._capabilities)
+
+        def register_capabilities(self, capabilities):
+            if isinstance(capabilities, str):
+                self._capabilities.add(capabilities)
+            elif isinstance(capabilities, (tuple, list, set)):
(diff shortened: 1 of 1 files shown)


m-col:
Yep so I mentioned this change [here](https://github.com/qtile/qtile/issues/1521#issuecomment-603981088) with regard to the graphical notification server.

The capabilities part makes the notifier in `libqtile.notify` follow the dbus spec more closely by providing more information about what the notifier is offering in terms of notifications - anything registering the notifier (like the notify widget, or the aforementioned graphical server) can update it with more things they are capable of.

The other changes relay all information that `libqtile.notify.notifier` receives via dbus, as seen [in this Notify method](https://github.com/qtile/qtile/pull/1667/files#diff-72fbd9cc5af5b8e5d49b8db8f372c761L57). Currently, most fields are just ignored but they can be passed and used to whatever is using the libqtile notifier in case they can make use of them, e.g. icons, replacing previous notifications, processing any desired actions.

m-col:
The concrete pre-existing case is so that my notification server code (which I'll submit too) can display app icons on notifications, can replace already-present notifications of the same type, knows the app name that sent the notification.

ramnes:
Ping @dequis :)

dequis:
Holy shit I wrote this 10 years ago

ramnes:
Holy shit @dequis, you wrote this 10 years ago, AND IT STILL WORKS!

PR #1668, opened 2020-04-23
Qtile terminal

zordsdavini:
plan to have script that run terminal


libqtile/resources/default_config.py
```diff
@@ -52,7 +52,7 @@
     # Unsplit = 1 window displayed, like Max layout, but still with
     # multiple stack panes
     Key([mod, "shift"], "Return", lazy.layout.toggle_split()),
-    Key([mod], "Return", lazy.spawn("xterm")),
+    Key([mod], "Return", lazy.spawn("qtile-terminal")),
```
tych0, line 55: Can we change this to a python function that looks something like lazy.spawn_reasonable_terminal()? It would be great not to have to distribute another "binary" with everything.

zordsdavini: the first idea was as you suggest. Thou I wanted to give a try. But I agree with you. Other idea is to set default terminal and use `lazy.spawn_terminal()` or ex. `lazy.spawn_terminal('tmux')`, so the command could be executed. Then in config we can add `default_terminal = None` as default and ex. `default_terminal = 'st'` would run in `st`

ramnes:
I think I like this approach more than another function, although I'm not sure about the shell part.

What do you see as a problem with distributing more executables, @tych0?

tych0:
More moving parts for less value. More random stuff in $PATH. It just feels sloppy :)

libqtile/resources/default_config.py
am01k01h3: Really like this idea, over the binary. 
I imagine, most people already have their pick for default terminal and this would support that.

ramnes:
What about a third solution? :)

I'd like to propose to just write a function in libqtile.utils for finding the said reasonable terminal. That would allow to just do this in the default config:

```python
from libqtile.utils import find_terminal

terminal = find_terminal()
...
Key([mod], "Return", lazy.spawn(terminal))
```

That way you don't have the binary, but you don't have the redundant cmd_ function neither.

Sorry if I'm nitpicking!

Issue #1582, opened 2020-04-03
Plans for default config improvements


zordsdavini:
Brainstorming what should be done to have friendly start for newcomers.

- add "Press <M-r> to spawn". Maybe in different color
- detect terminal (not all has xterm)
- review default keybindings (looks like some of them are not working)
- review default layouts, uncommenting should give working layout
- add default keybindings picture to documentation
- add optional dependencies to documentation
- start terminal by launching default Qtile

pawamoy:
I was thinking about adding a shortcut to pop a notification to display the current key bindings and their description.

zordsdavini:
I think we should compare to other WM what they have in default. For exampe DWM comes with almost nothing :) But sucless dwm tries to follow unix filosofy. In our case Qtile is not only WM. In my case it has many integrations and looks more like DE then WM :rofl: 
but @pawamoy I totally agree with you about notifications on failing parts. I remember you did some tricks for keybindings. We should add it. The case was notification app if I am not wrong. Just add fallback to logger.warning()

am01k01h3:
I think it doesn't matter what other WMs do. I assume Our goal is for most people to have a lower barrier of entry to adopt qtile. If yes, then we should do whatever possible to make it easier for people.

For start, it would be great if there is a config.py present in the XDG config dir.
And, if the file has lots of options baked in, even if they are commented out, that would be awesome. So, anyone that bothered to try out qtile, at least has some way to move forward.

On the other hand, if the goal is similar to suckless to keep it elitist with higher barrier of entry, then so be it :+1:

zordsdavini:
> I was thinking about adding a shortcut to pop a notification to display the current key bindings and their description.
@pawamoy in `qshell`  we have nice command `display_kb`. I tried to run from console `qtile-cmd -o cmd -f display_kb` but it gives very ugly print out.

zordsdavini:
@am01k01h3 default config is already copied to users directory on first startup. I don't know about stable Qtile version but dev version has it

zordsdavini:
for picture of default keybindings could be similar to ![xmonad](https://wiki.haskell.org/wikiupload/b/b8/Xmbindings.png)

zordsdavini:
And about default layouts, I'd suggest switch to Max and Columns. Columns has same functionality as Stack but more posibilitites

ramnes:
It's being proposed in #1628 that we had descriptions to Key objects, and I like the idea.

arczan1:
[...]

PR #1669, opened 2020-04-24
Spawn terminal

zordsdavini:
Version 3
`guess_terminal` in utils

Now in config it could be used:
```python
from libqtile.utils import get_duess_terminal
terminal = guess_terminal()
<...>
    Key([mod], "Return", lazy.spawn(terminal), desc="Launch terminal"),
<...>
```
or if you want some app inside
`Key([mod], "t", lazy.spawn([terminal, '-e', 'htop']), desc="Launch htop")`

Part of https://github.com/qtile/qtile/issues/1582

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+qtile X.X.X, released XXXX-XX-XX:
+    * features
+        - added `guess_terminal` in utils
+        - added keybinding cheet sheet image generator
+
 qtile 0.15.0, released 2020-04-12:
     !!! Config breakage !!!
         - removed the mpd widget, which depended on python-mpd.
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -50,7 +50,7 @@ key. The basic operation is:
 * ``mod + <ctrl> + r``: restart qtile with new config
 * ``mod + <group name>``: switch to that group
 * ``mod + <shift> + <group name>``: send a window to that group
-* ``mod + <enter>``: start xterm
+* ``mod + <enter>``: start terminal guessed by ``libqtile.utils.guess_terminal``
 * ``mod + r``: start a little prompt in the bar so users can run arbitrary
   commands
 
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -23,6 +23,7 @@ import importlib
 import os
 import traceback
 import warnings
+from shutil import which
 
 from libqtile.log_utils import logger
 
@@ -209,3 +210,39 @@ def send_notification(title, message, urgent=False, timeout=10000):
         notifier.show()
     except Exception as exception:
         logger.error(exception)
+
+
+def guess_terminal():
+    """Try to guess terminal."""
+    test_terminals = [
+            'x-terminal-emulator',
+            'roxterm',
+            'sakura',
+            'hyper',
+            'alacritty',
+            'terminator',
+            'termite',
+            'gnome-terminal',
+            'konsole',
+            'xfce4-terminal',
+            'lxterminal',
+            'mate-terminal',
+            'kitty',
+            'yakuake',
+            'tilda',
+            'guake',
+            'eterm',
+            'st',
+            'urxvt',
+            'xterm',
+            ]
+
+    for terminal in test_terminals:
+        logger.debug('Guessing terminal: {}'.format(terminal))
+        if not which(terminal, os.X_OK):
+            continue
+
+        logger.info('Terminal found: {}'.format(terminal))
+        return terminal
+
+    logger.error('Default terminal has not been found.')
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -29,8 +29,10 @@ from typing import List  # noqa: F401
 from libqtile import bar, layout, widget
 from libqtile.config import Click, Drag, Group, Key, Screen
 from libqtile.lazy import lazy
+from libqtile.utils import guess_terminal
 
 mod = "mod4"
+terminal = guess_terminal()
 
 keys = [
     # Switch between windows in current stack pane
@@ -59,7 +61,7 @@ keys = [
     # multiple stack panes
     Key([mod, "shift"], "Return", lazy.layout.toggle_split(),
         desc="Toggle between split and unsplit sides of stack"),
-    Key([mod], "Return", lazy.spawn("xterm"), desc="Launch xterm"),
+    Key([mod], "Return", lazy.spawn(terminal), desc="Launch terminal"),
 
     # Toggle between different layouts as defined below
     Key([mod], "Tab", lazy.next_layout(), desc="Toggle between layouts"),


ramnes:
See my comment here: https://github.com/qtile/qtile/pull/1668#issuecomment-618872032

zordsdavini:
@ramnes but this solutions has advantage on adding commands and quickly change terminal app.
From other hand utils choice could eat command and format as list for lazy.spawn.

libqtile/core/manager.py
```diff
@@ -1224,6 +1224,49 @@ def cmd_spawn(self, cmd):
...
+        if isinstance(cmd, str):
+            args = shlex.split(cmd)
+        else:
+            args = list(cmd)
+
+        TEST_TERMINALS = [
+                ['x-terminal-emulator', '-?'],
+                ['xterm', '-help'],
+                ['urxvt', '-help'],
+                ['st', '-v'],
```
ramnes, line 1243: Let's add a few more:
* gnome-terminal
* terminator
* lxterminal
* mate-terminal
* roxterm
* tilda
* guake
* eterm
* sakura
* konsole
* hyper
* alacritty

zordsdavini: I don't think we need so much because we need basic terminal what should be installed by default. But ok - this don't cost much

ramnes: Well if you come from a GNOME environment, i.e. Ubuntu, I think that you might only have gnome-terminal installed. I do agree though that we could remove the terminals that doesn't come by default with any known distribution, à la alacritty or hyper.

zordsdavini: all debian based distributions has `x-terminal-emulator`

ramnes: Alright, Ubuntu was a bad example, but there are other distributions than Debian-based distributions. :stuck_out_tongue:

libqtile/core/manager.py
```diff
@@ -1224,6 +1224,61 @@ def cmd_spawn(self, cmd):
...
+
+        default_terminal = getattr(self.config, "default_terminal", None)
+        if default_terminal:
+            full_cmd = [default_terminal]
+            if len(args):
+                full_cmd += ['-e'] + args
+            return self.cmd_spawn(full_cmd)
+
+        to_null = ['>', '/dev/null']
+        for terminal in test_terminals:
```
ramnes, line 1266: Why not use get_default_terminal here?

zordsdavini: sorry, forgot to remove this command - it is not needed anymore. Will add something to documentation

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+qtile X.X.X, released XXXX-XX-XX:
+    * features
+        - added `get_default_terminal` in utils
+        - added keybinding cheet sheet image generator
```
ramnes, line 4: It's no big deal but this line should rather be in the other PR :)

libqtile/core/manager.py
```diff
@@ -1224,6 +1224,61 @@ def cmd_spawn(self, cmd):
             os.close(r)
             return int(pid)
 
+    def cmd_spawn_terminal(self, cmd=[]):
```
ramnes: Can we make a second PR for this?

libqtile/utils.py
```diff
@@ -209,3 +209,41 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['hyper', '-v'],
+            ['alacritty', '-V'],
+            ]
+
+    to_null = ['>', '/dev/null']
+    for terminal in test_terminals:
+        logger.warning('test '+terminal[0])
+        try:
+            exit_code = os.spawnvp(os.P_WAIT, terminal[0], terminal + to_null)
+            if 127 == exit_code:
```
ramnes, line 240:
```suggestion
            if exit_code == 127:
```
:)

libqtile/utils.py
```diff
@@ -209,3 +209,41 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['eterm', '-h'],
+            ['sakura', '-v'],
+            ['konsole', '--help'],
+            ['hyper', '-v'],
+            ['alacritty', '-V'],
+            ]
+
+    to_null = ['>', '/dev/null']
+    for terminal in test_terminals:
+        logger.warning('test '+terminal[0])
```
ramnes, line 237:
```suggestion
        logger.debug('test '+terminal[0])
```

CHANGELOG
zordsdavini: I've already squashed all changes in that PR and it is waiting for merge. Sorry but I wanted to save somehow ;)
Next time I'll be more accurate

libqtile/utils.py
```diff
@@ -209,3 +209,41 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+        try:
+            exit_code = os.spawnvp(os.P_WAIT, terminal[0], terminal + to_null)
+            if 127 == exit_code:
+                continue
+        except Exception:
```
ramnes, line 242: What kind of exception are we expecting here?

libqtile/core/manager.py
zordsdavini: I've removed that part. Don't me?

libqtile/utils.py
```diff
@@ -209,3 +209,41 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            logger.error('Terminal guesser failed on %s' % terminal[0])
+            continue
+
+        logger.warning('found '+terminal[0])
```
ramnes, line 246:
```suggestion
        logger.info('Terminal found: {}'.format(terminal[0]))
```

libqtile/utils.py
zordsdavini: I forgot the name of that practice but I follow it :)

libqtile/utils.py
zordsdavini: I was not sure for 100% here, so I left catcher here. Should I remove?

libqtile/utils.py
```diff
@@ -209,3 +209,41 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['mate-terminal', '-h'],
+            ['roxterm', '-?'],
+            ['tilda', '--help'],
+            ['guake', '-h'],
+            ['eterm', '-h'],
+            ['sakura', '-v'],
+            ['konsole', '--help'],
+            ['hyper', '-v'],
+            ['alacritty', '-V'],
+            ]
```
pawamoy, line 233: `xfce4-terminal` is quite a common terminal as well.

`]` should be dedented :stuck_out_tongue:

libqtile/utils.py
zordsdavini: ups

libqtile/utils.py
```diff
@@ -209,3 +209,38 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['lxterminal', '-v'],
+            ['mate-terminal', '-h'],
+            ['roxterm', '-?'],
+            ['tilda', '--help'],
+            ['guake', '-h'],
+            ['eterm', '-h'],
+            ['sakura', '-v'],
+            ['konsole', '--help'],
+            ['hyper', '-v'],
+            ['alacritty', '-V'],
```
ramnes, line 233: I think I would revert the list so that the more "modern" and "powerful" terminals can be found first. Because if such a terminal is present, it means that the user installed it for its own use, and thus would probably prefer to see it spawn rather than something like xterm. Is it a fair assumption?

For example, I currently have roxterm, urxvt and xterm installed, and I would like them to be found in that order.

zordsdavini: let's try to build queue

            ['x-terminal-emulator', '-?'],
            ['roxterm', '-?'],
            ['sakura', '-v'],
            ['hyper', '-v'],
            ['alacritty', '-V'],
            ['terminator', '-v'],
            ['gnome-terminal', '-?'],
            ['konsole', '--help'],
            ['xfce4-terminal', '-V'],
            ['lxterminal', '-v'],
            ['mate-terminal', '-h'],
            ['tilda', '--help'],
            ['guake', '-h'],
            ['eterm', '-h'],
            ['st', '-v'],
            ['urxvt', '-help'],
            ['xterm', '-help'],

ramnes: I'd put tilda and guake above gnome-terminal, roxterm under terminator, and eterm under urxvt. It looks pretty good to me otherwise!

zordsdavini: I disagree about guake and tilda - they are not real terminals :D and I don't know how they should look if I press new terminal twice. I don't think that we will find ideal way to represent all people. The main reason for this is to have terminal then you start. I think list should be as it is and maybe from practice we will see what is not in right place. In my case I will not use this utility myself and after installing os I will install my loved terminal and it will be alone (st), other just for testing (xterm because of Qtile dev and Alacritty because I havent removed it yet)

ramnes: I agree, but then let's remove guake and tilda? I don't think anyone has only one of these two installed anyway.

CHANGELOG
tych0: Thanks for adding this; we should probably go back and look at the other stuff we've added, since people asked me about release notes for the last release :)

libqtile/utils.py
```diff
@@ -209,3 +209,38 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['eterm', '-h'],
+            ['st', '-v'],
+            ['urxvt', '-help'],
+            ['xterm', '-help'],
+            ]
+
+    to_null = ['>', '/dev/null']
+    for terminal in test_terminals:
+        logger.debug('Guessing terminal: {}'.format(terminal[0]))
+        exit_code = os.spawnvp(os.P_WAIT, terminal[0], terminal + to_null)
```
tych0, line 239: Might be worth using subprocess here so we don't need the shell intermediary to do the redirection. But also maybe it doesn't matter.

tych0:
This looks good to me, if you can squash it all into consumable commits that would be great! One minor comment, feel free to ignore it or not.

libqtile/utils.py
ramnes: I was expecting this comment :)

zordsdavini: I thought about subprocess, but tried this way

zordsdavini:
squashed

libqtile/utils.py
```diff
@@ -209,3 +211,39 @@ def send_notification(title, message, urgent=False, timeout=10000):
...
+            ['st', '-v'],
+            ['urxvt', '-help'],
+            ['xterm', '-help'],
+            ]
+
+    to_null = ['>', '/dev/null']
+    for terminal in test_terminals:
+        logger.debug('Guessing terminal: {}'.format(terminal[0]))
+        try:
+            call(terminal + to_null)
```
flacjacket, line 242: `subprocess.run()` seems like it'd be useful here. Also, adding a timeout is probably a good idea.

airwoodix: What about ``subprocess.check_call`` on ``which <terminal name>``?
This way one doesn't need to maintain a list of options to call the terminal emulators with immediate return and there's no need for a timeout. Or is ``which`` not portable enough?

libqtile/utils.py
airwoodix: [termite](https://github.com/thestinger/termite) and [kitty](https://github.com/kovidgoyal/kitty) are also [somewhat popular](https://wiki.archlinux.org/index.php/Dotfiles).

libqtile/utils.py
zordsdavini: > `subprocess.run()` seems like it'd be useful here. Also, adding a timeout is probably a good idea.

I missmatched python versions and did not noted `run`. But why timeout is needed?

zordsdavini: > What about `subprocess.check_call` on `which <terminal name>`?
> This way one doesn't need to maintain a list of options to call the terminal emulators with immediate return and there's no need for a timeout. Or is `which` not portable enough?

I tried that but the flow didn't worked. Speaking the true, I used python 2 documentation, so maybe something changed

libqtile/utils.py
zordsdavini: we don't need to cover all terminals. The main idea is to give user some working terminal. If we will see that list should be expanded - then new PRs would be welcome.

libqtile/utils.py
airwoodix: Sorry, my bad, ``subprocess.check_call`` was superseeded by ``subprocess.run`` in Python 3.5.

However:
  - the proper way to redirect standard output to the null data sink is passing ``stdout=subprocess.DEVNULL`` to ``subprocess.run``, not a shell redirection without ``shell=True`` (the call ``subprocess.run(["xterm", "-help", ">", "/dev/null"])`` indeed succeeds with return code 0 when ``xterm`` is installed on the system but the redirection is not effective and ``xterm`` complains with `xterm: No absolute path found for shell: >`). Standard error should also be redirected to ``/dev/null`` since for example ``st -v`` writes on standard error (and exits with status 1).

  - Agreed that it is unlikely that terminal emulators change how to access their help or version number. I find it nevertheless useless to maintain a list of these options together with a list of terminal emulators. Calling ``which`` has the following advantages:
    + no need to check/maintain which options need to be passed to the terminal emulators to silently check their presence;
    + consistent interface: if ``term`` is there, ``which term`` exits with 0, else with 1;
    + no call to the actual terminal emulator, no real need for a timeout (although something like 1 second would still be reasonable);
    + ``subprocess.run(["which", "<terminal>"], check=True, stdout=DEVNULL)`` in a try-except block catching ``CalledProcessError`` is clean and explicit. Catching ``FileNotFoundError`` means that ``which`` is not there and the search can stop immediately.

``which`` seems [widespread enough](https://en.wikipedia.org/wiki/Which_(command)) to me but I would not argue if someone comes with a decent counter-example

IMHO, the first point should be fixed. I'm happily discussing the second one further if wanted.

libqtile/utils.py
airwoodix: > we don't need to cover all terminals.

Why not? To me, everything that is a terminal emulator and is in a non-user contributed repository (i.e. no AUR) of a mainstream distribution could make it in the list. ``termite``, ``kitty``, but also ``yakuake`` are in community or extra repos in Arch. ``yakuake`` is even in Debian stable (Quake-style emulator for KDE, like ``guake`` for Gnome).

This criterion is not satisfied by all programs in this list (e.g. ``roxterm`` (AUR: 10 votes, pop 0.03), ``hyper`` (AUR: 58 votes, pop 1.30), ``eterm`` (AUR: 25 votes, pop 0)) and that's in principle fine, but I see no reason to not extend the search list right now. There will be no better metric in the future to decide if a program is worth adding (telemetrics is not a suitable solution).

libqtile/utils.py
zordsdavini: @airwoodix thanks for such good explaination :) Fixed as you suggested

pawamoy: If using `which`, what about [`shutil.which()`](https://docs.python.org/3/library/shutil.html#shutil.which) then?

libqtile/utils.py
zordsdavini: I loved yakuake ~15 years ago. Ok, added kitty, termite, yakuake

libqtile/utils.py
zordsdavini: :D @pawamoy good command, too. But maybe let's stop here

docs/manual/config/index.rst
```diff
@@ -50,7 +50,7 @@ key. The basic operation is:
 * ``mod + <ctrl> + r``: restart qtile with new config
 * ``mod + <group name>``: switch to that group
 * ``mod + <shift> + <group name>``: send a window to that group
-* ``mod + <enter>``: start xterm
+* ``mod + <enter>``: start terminal guessed by `libqtile.utils.guess_terminal`
```
pawamoy, line 53: I think you want double backticks, not single ones: <code>\`\`libqtile.utils.guess_terminal\`\`</code>

zordsdavini: Yes I do, fixed

libqtile/utils.py
airwoodix: @pawamoy good point, of course I forgot to check ``shutil``... I also think it's a better match than the subprocess call. It removes the dependency to ``which`` and the interface is even cleaner. 
@zordsdavini it doesn't cost much to fix that, right?

zordsdavini: 6 x 5€ = 30€

zordsdavini: pushed

airwoodix: thanks

tych0:
Cool, thanks!

Issue #1657, opened 2020-04-21
Restarting takes two minutes because keyring is blocking


ep12:
# Issue description

# Qtile version
0.15.1-1 (arch community repo)
More version stuff:
```
Name            : gdk-pixbuf2
Version         : 2.40.0-2
Name            : pango
Version         : 1:1.44.7+11+g73b46b04-1
Name            : python-cffi
Version         : 1.14.0-2
Name            : python-xcffib
Version         : 0.9.0-1
Name            : python-cairocffi
Version         : 1.1.0-3
Name            : python-dbus
Version         : 1.2.16-1
Name            : python-gobject
Version         : 3.36.0-2
Name            : python-setuptools
Version         : 1:46.1.3-1
Name            : python-six
Version         : 1.14.0-1
```

# Stack traces
+ Sometimes the following occured in `~/.local/share/qtile/qtile.log`:
```
2020-04-21 11:40:45,004 ERROR libqtile manager.py:cmd_restart():L1141 Preventing restart because of a configuration error: The configuration file
 '~/.config/qtile/config.py' generated the following error:

ConnectionException: xcb connection errors because of socket, pipe and other stream errors.
```
+ Relevant lines from `~/.local/share/qtile/qtile.log`:
```
2020-04-21 13:36:51,182 WARNING libqtile manager.py:loop():L245 Restarting Qtile with os.execv(...)
2020-04-21 13:36:51,791 WARNING libqtile utils.py:safe_import():L191 Unmet dependencies for optional Widget: '.widget.mpd2widget.Mpd2', No module named 'mpd'
2020-04-21 13:36:51,796 WARNING libqtile utils.py:safe_import():L191 Unmet dependencies for optional Widget: '.widget.wlan.Wlan', No module named 'iwlib'
2020-04-21 13:38:59,145 ERROR libqtile manager.py:cmd_spawn():L1163 failed spawn: "a"
[Errno 2] No such file or directory
```
Note that I tried to spawn `a` ***immediately*** after qtile was back.

# Configuration
[most part is here](https://www.github.com/ep12/qtileconf)
Also tried it with the default config (renamed `config.py -> .config.py`). That takes two minutes to restart, too => *not config related!?!*

# Notes
The snippet below also took about two minutes when I (first) executed it. (After sleep and resume it took a fraction of a second, don't ask me why...)
```py
import os
import time
from libqtile.confreader import Config
from libqtile.backend.x11.xcore import XCore
s = time.time() 
try: 
[...]

PR #1670, opened 2020-04-24
Verbose logging for libqtile.widget.__init__

ep12:
(See #1657)
This commit implements very basic and verbose logging for the `libqtile.widget` import. Because python caches the module, this should only pollute the log when qtile starts.
Maybe a time threshold should be introduced because this is now very
verbose...

### DIFF
--- a/libqtile/widget/__init__.py
+++ b/libqtile/widget/__init__.py
@@ -20,6 +20,9 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import logging
+import time
+
 from libqtile.utils import safe_import as safe_import_
 # only directly import widgets that do not have any third party dependencies
 # other than those required by qtile, otherwise use the same import function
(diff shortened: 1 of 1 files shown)


libqtile/widget/__init__.py
```diff
@@ -41,71 +41,92 @@
...
-    )
-    e = time.time()
+def safe_import(imports):
     if any(x.startswith('libqtile') for x in logging.Logger.manager.loggerDict):
         log = logging.getLogger(__name__)
-        log.info('Attempt to safely import .widget.%s as %s took %.3f seconds',
-                 module_name, class_name, e - s)
+        log.info('Importing widgets')
+    else:
+        log = None
```
ep12, line 49: Write a test for that? If so, how?

ramnes:
Thanks for the effort @ep12. 

I'd prefer to keep the safe_import function as bare as possible, and thus I don't really like the switch to a single call. Also I feel this is way too much code for what it does, or at least it should be decoupled (i.e. mesuring the time a function take can be useful in a lot of places, not just here).

Overall I'm pretty much -1 for it, sorry. I don't think we need to clutters the code by timing the imports. As we say in France, it feels like "a band-aid on a wooden leg". It seems to me that the real problem to fix is that a faulty widget will always be imported, and thus will break Qtile, even if the user is not using it.

ep12:
Never mind! What do you think how should the timing be handled? Using a decorator that logs it?

ep12:
Closed in favor of .

PR #1675, opened 2020-04-24
Implement lazy importing of widget submodules

ep12:
See #1657: Lazily import `libqtile.widget` submodules as required. Log the import with appropriate levels depending on the duration.


ep12:
The tests were failing. PEP 562 requires python 3.7 or later. ef2434d fixes that.

tych0:
Can you describe what problem this is solving? It's not clear to me why we need to do this. I'd also prefer to import that copy paste from somewhere, vs. having it live in the tree.

ep12:
@tych0 
The tests for python <3.7 were failing. To fix that I found that [this](https://github.com/facelessuser/pep562) is probably the easiest solution. Alternatively, we could fall back to import all submodules when using older versions of python. But that would contradict the whole point of making imports lazy.
Note that the project explicitly suggests to copy this file to the tree. If anything else should be imported in a lazy fashion, the `pep562` module could be reused.

tych0:
The tests were just failing because of the way this PR was implemented, right? That is not the reason you did it in the first place.

Generally, I feel like if we're going to do anything, we should stop the vanity imports into __init__ and just make everyone do `from libqtile.widgets.widget import Widget`. But it seems better to try and fix the widgets that can sleep for a long period of time and delay things. I don't love a bunch of additional code we have to maintain just for vanity imports.

ramnes:
I'm sorry @ep12, but this is way too much code to maintain for the problem we want to solve. :(

If we cannot find a simpler solution (i.e. a 10 lines max diff), I'd rather go the @tych0's way and break all configs by clearing that __init__ file.

tych0:
Well before we do that, we should debug the widgets a bit more. Do we know
what is hanging? Can we fix it?

Tycho

ep12:
Will continue the discussion where it started: see #1657...

ramnes:
@ep12 Can you update this PR to the "minimized" 17 lines diff version you were talking about, or open a new one? I'd like to see how that shorter version would compare with what I've pushed in #1678 so that we can choose wisely or merge both solutions.

libqtile/widget/__init__.py
```diff
@@ -20,6 +20,9 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import sys
+
+from libqtile.log_utils import logger
```
ep12, line 25: unused import... Forgot about that

pawamoy:
I find the proxy version easier to reason about. The idea of not repeating `safe_import` for each widget is nice though. However maybe this could be made simpler by inverting the keys and values of the lazy spec:

```python
LAZY_IMPORT_SPEC = {
    'Backlight': 'backlight',
    'Battery': 'battery',
    'BatteryIcon': 'battery',
    'CurrentScreen': 'currentscreen',
    'DebugInfo': 'debuginfo',
    'CPUGraph': 'graph',
    'MemoryGraph': 'graph',
    'SwapGraph': 'graph',
    'NetGraph': 'graph',
    'HDDGraph': 'graph',
    'HDDBusyGraph': 'graph',
    ...
}
```

This way you don't have to iterate on the values to find the name: 

```python
    if name in __all__:
        module_name = LAZY_IMPORT_SPEC[name]
        safe_import(module_name, name)
```

...or something similar.

ep12:
This PR has now 17 lines of relevant new code, that is 17 lines more than it previously had (with dozens of `safe_import(x, y)` calls. Those still have to be maintained (as before, so they are not new).
In contrast to #1678 I only changed `libqtile.widget.__init__`, which should be emptied later on afaik. It is therefore not necessary to remember which other files have changed. Just clear `__init__.py` and you're done.
One downside of this is that for python < 3.7 all modules are imported (not lazily -> old behaviour).
If `__init__.py` should be cleared in the near future (breaking change), then #1678 and this PR are equivalent because they are both temporary solutions that will be removed later.
Before merging any of those PRs, a change should be made to emit a `PendingDeprecationWarning` or a `DeprecationWarning` when the import chain hits `libqtile.widget.__init__`.

ep12:
@pawamoy Yes, that is a possibility.
Pro:
+ `__getattr__` would be easier (and faster?)

Contra:
+ with old python versions, every import would happen seperately. Note that python's import machinery would import the module only once, so why not bind all related names at the same time

In the end, it does not really matter (assuming that this is a temporary fix that will be removed soon).

libqtile/widget/__init__.py
```diff
@@ -37,66 +39,85 @@
...
+    module_name, class_names = None, None
+    if name in __all__:
+        for mod, names in LAZY_IMPORT_SPEC.items():
+            if name in names:
+                module_name, class_names = mod, names
+                break
+    if (module_name, class_names) != (None, None):
+        safe_import(module_name, class_names)
+        if name in globals():
+            return globals()[name]
```
ep12, line 116: Maybe these two lines could be moved to the top of the function to make look-ups of already imported objects faster...

libqtile/widget/__init__.py
```diff
@@ -37,66 +39,85 @@
 from libqtile.widget.textbox import TextBox  # noqa: F401
 from libqtile.widget.windowname import WindowName  # noqa: F401
 
+LAZY_IMPORT_SPEC = {
```
flacjacket, line 42: You could setup a
```
LAZY_MODULE_SPEC = {
    widget: module for module, widgets in LAZY_IMPORT_SPEC.items() for widget in widgets
}
```
that would make the look-ups performed in the `__getattr__` a lot easier to resolve.

flacjacket:
That is to say, I don't think we need to be making it deprecated to import directly out of `libqtile.widget`.

libqtile/widget/__init__.py
ep12: https://github.com/qtile/qtile/pull/1675#issuecomment-621884260

ramnes:
Okay so @flacjacket seems to like more lazyness, @tych0 seems to be +0, and I'm definitely +1 for this, so all in all let's not keep this stale. I'll go ahead and merge my PR rather than this one for the sole reason that it's compatible and consistent with all Python versions. Thank you very much for your work here @ep12, for raising the issue in the first place, then investigating and finding the faulty import! We still should try to understand why keyring is blocking and how to fix it, but at least this won't break anymore for people that do not use the Imap widget.

PR #1676, opened 2020-04-25
Provide in depth docs on the structure and design of the command interfaces

flacjacket:
Fix up some of the current docs and provide additional information on
the structure and design of the various components that underlie the
command graph structure.  This is a developer-centric overview of how
these component slot together and how the command graph itself relates
back to the qtile objects.


ramnes:
Awesome! Let's fix the build and merge this ASAP!

Issue #1329, opened 2019-05-30
Improvement : overview of keyboard shortcuts


erikdubois:
I was wondering if something like this already exists and if not ...
Will you consider in making this available?
An overview of all keyboard shortcuts from the config.py.

Something like this on Awesome
![arcolinux-awesome-keyboard-shortcuts](https://user-images.githubusercontent.com/10594806/58653122-554a5f00-8315-11e9-86c8-3cc9475cc154.jpg)

or 

something like this on Ubuntu
![ubuntu](https://user-images.githubusercontent.com/10594806/58653141-5f6c5d80-8315-11e9-953c-94563116267e.jpg)

That would be just great.

ramnes:
Hey @erikdubois :wave: 

That would be very nice, indeed. Contributions welcome!

erikdubois:
Indeed - if someone can contribute that would be great... Not me though. All my time goes into the development of ArcoLinux ;-)

krive001:
Update suggestion from manager.py
cmd_display_kb function.
![ArcoLinuxD_2019_06_01_13_37_51](https://user-images.githubusercontent.com/24889940/58749258-17b81400-8484-11e9-902b-46ce044466f4.jpg)

manager.py lines
def cmd_display_kb(self, *args):
        """Display table of key bindings"""
        class FormatTable(object):
            def __init__(self):
                self.max_col_size = []
                self.rows = []

            def add(self, row):
                n = len(row) - len(self.max_col_size)
                if n > 0:
                    self.max_col_size += [0] * n
                for i, f in enumerate(row):
                    if len(f) > self.max_col_size[i]:
                        self.max_col_size[i] = len(f)
                self.rows.append(row)

            def getformat(self):
                format_string = (" | ".join(
                                 "%-{0:d}s".format(max_col_size)
                                 for max_col_size in self.max_col_size))
                return format_string + "\n", len(self.max_col_size)

            def expandlist(self, list, n):
                if not list:
                    return ["+" * max_col_size
                            for max_col_size in self.max_col_size]
                n -= len(list)
                if n > 0:
                    list += [""] * n
                return list

            def __str__(self):
                format, n = self.getformat()
                return ("".join([format % tuple(self.expandlist(row, n))
                                 for row in self.rows]))

        result = FormatTable()
        result.add(["Mod+" + "KeySym", "Description"])
        result.add([])
        rows = []
        for (ks, kmm), k in self.keyMap.items():
            if not k.commands:
                continue
[...]

PR #1677, opened 2020-04-25
add Popup window class

m-col:
This adds popup windows that can be used for a variety of purposes, including:

- notifications (#1521)
- tooltips (#1441 )
- a keyboard overview window (#1329)

Once this is merged I can use it to add my notification server code into qtile, which should make it clear if `Popup` needs updating e.g. to move some of the setup logic into a `._configure` method. I can then add a small bit in the docstring to show how to use it for any future use.

### DIFF
--- a/libqtile/pangocffi.py
+++ b/libqtile/pangocffi.py
@@ -74,6 +74,13 @@ ELLIPSIZE_END = pango.PANGO_ELLIPSIZE_END
 units_from_double = pango.pango_units_from_double
 
 
+ALIGNMENTS = {
+    'left': pango.PANGO_ALIGN_LEFT,
+    'center': pango.PANGO_ALIGN_CENTER,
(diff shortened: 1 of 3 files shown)


m-col:
Lol, I did that initially but thought I'd get the comment "why not put it inside libqtile.window?". Yeah I think something separate would be nicer to avoid poluting libqtile.window.

ramnes:
Let's wait for a second review, but this looks really good to me. :)

m-col:
I'll have to set up my system for the recent changes made to make lint ASAP to avoid more failing tests. All sorted now :)

tych0:
Can you add a test for this?

m-col:
I've added a test to create a popup and check that it hasn't become part of a group and focussed. I'm not sure what else there would be to test so let me know if there is anything else that would be good to add :)

PR #1678, opened 2020-04-25
Make safe imports lazy to avoid side effects when using vanity imports

ramnes:


tych0:
I still feel like the way to do this is to just fix the widgets. It seems impolite to block for two minutes when your module is imported, so people will probably be willing to take patches to their libraries to fix it.

ramnes:
Yes of course, I also want to fix the widget! I just want to avoid side effects so that the next time we have a faulty widget, it doesn't break qtile even if the user isn't using it. We would have saved a lot of debugging time with this in place for the two minutes problem, as anyone could have isolated the bug to the imap widget.

ramnes:
To be honest, I'd rather remove these vanity imports as you suggested, but I feel this is a good middle ground between the current behavior of importing all widgets for nothing, and breaking all configs.

ramnes:
See https://github.com/qtile/qtile/pull/1675#issuecomment-628468546

Issue #1518, opened 2020-01-08
Qtile doesn't respect WM_SIZE_HINTS


Brn9hrd7:
Qtile by default sets all windows with the WM_STATE "dialog" to floating automatically. But there are also some programs that use WM_STATE "normal" and only set a minimum and maximum window size hint. I think windows with a maximum window size hint should be set to floating as well, otherwise they can look pretty weird because they're not meant to be sized differently. An example for this would be the startup dialog and the info dialog of Steam.

pmkap:
I have a hook for this in my `config.py`:

```@hook.subscribe.client_new
def floating_size_hints(window):
    hints = window.window.get_wm_normal_hints()
    if hints['max_width']:
        window.floating = True
```

Maybe this should be in the docs or in the default config.

ramnes:
Nice! PR welcome!

pmkap:
Alternatively, I can try to make the changes in `floating.py` and `window.py` if this is preferable.

ramnes:
As you feel. Did you ever see anything wrong with it? I'm kinda worried it could have indésirable side effects, i.e. put in floating applications that do not handle their hints correctly but aren't meant to float. I'm adding it to my configuration right now to experiment with it.

ramnes:
So far so good, it makes a lot of things I previously manually handled in my configuration floating by default.

pmkap:
I cannot tell about side effects yet because I also just started using it. I will observe, though. Please note I just made another small change in in my comment above. I made a mistake before.

ramnes:
I modified it a tiny bit to make it (I think) even simpler:

```python
@hook.subscribe.client_new
def set_floating(window):
    if window.window.get_wm_normal_hints().get("max_width"):
        window.floating = True
```

pmkap:
Ok, but this could possibly lead to an AttributeError because `get_wm_normal_hints` either returns a dict or None. I didn't check on which occasions it returns None but it's in the code.

ramnes:
Ah, right! Ignore my previous comment.

pmkap:
Is there a reason why the `class Floating` has different parameters `float_rules` and `auto_float_types`? This is somehow confusing. Even more if we add additional floating properties like the maximum size.
My idea is to have one new parameter for all information about which windows to float. Maybe like this:
```python
float_props = {
    'wmtype': ['toolbar', 'splash'],
    'role': ['About'],
    'wmclass': ['file_progress, classB'],
[...]

Issue #1275, opened 2019-01-31
How to make sure no windows start as floating?


wurosh:
Pretty self explanatory. I tried setting `float_rules=[]`, `float_rules=None`, and removing `layout.Floating()` from my `config.py` altogether. I just don't want to have to keep running `lazy.window.toggle_floating()` all the time because I really dislike handling floating windows, even if they are dialog boxes.

flacjacket:
It looks like any attempt to blank out `float_rules` will result in the float rules being set to the default. If you change:
```
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -91,7 +91,7 @@ class Floating(Layout):
         self.clients = []
         self.focused = None
         self.group = None
-        self.float_rules = float_rules or DEFAULT_FLOAT_RULES
+        self.float_rules = float_rules if float_rules is not None else DEFAULT_FLOAT_RULES
         self.no_reposition_match = no_reposition_match
         self.add_defaults(Floating.defaults)

```
does that fix this behavior?

jjzmajic:
Shoot, sorry for the delay (exam time). I just tried it, no luck.

jjzmajic:
Any updates?

jjzmajic:
Bump in case anyone thought of a solution...

ramnes:
It seems weird to me that @flacjacket diff does not work. Can you confirm that you tried to use the diff with `float_rules=[]` (and not `float_rules=None`)?

jjzmajic:
@ramnes I'm surprised too, but yes, I tried with `[]`

jjzmajic:
Did some more digging, but came out empty-handed. The logic of the patch should indeed fix the issue, but based on the above, it seems that `DEFAULT_FLOAT_RULES` replace `float_rules` some time before that takes effect.

jjzmajic:
Hate to be a nag about this, just wondering if there are any updates? I can't seem to find the place where user vars are overwritten no matter how hard I look.

m-col:
Looks like https://github.com/qtile/qtile/pull/1937 set the default `float_rules` to have no rules, so it can be omitted or set to `None` or `[]` to have no windows starting as floating.

PR #1680, opened 2020-04-27
Restructuring float rules

pmkap:
The Floating Layout has currently two parameters to specify the rules for floating windows. I think this is a bit confusing. Also it is not possible to have no floating windows all (Issue #1275).

This is a proposal to change this and let the user specify the different rules for floating with a single dictionary, e.g. like this.
```python
float_rules = {
    "wmclass": ["Wine", "vlc"],
    "role": ["About"],
    "wmtype": ["notification", "toolbar"],
    "wmname": ["appname"],
}
```
However, the current structure of `float_rules` is different, so this requires extra code to prevent old configs from breaking. If you don't like this, I will reconsider this. This is just the way that I find is the prettiest.

Also, I made another default rule which addresses issue  #1518. This is similar to what i3 uses.

If a user specifies `float_rules`, they are appended to preset rules. But the user can also override these.

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -68,6 +68,7 @@ class Floating(Layout):
             float_rules = {
                 "wmclass": ["Wine", "vlc"],
                 "role": ["About"],
+                "wmtype": ["notification", "toolbar"],
             }
         The follwing keys are possible. (With corresponding X property)
             "wmname"    (WM_NAME)
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -113,7 +113,6 @@ class Floating(Layout):
         """Compares a window's properties with the `float_rules`.
         Returns `True` if window should float.
         """
-        print(win.window.get_wm_type())
         if (win.name in self.float_rules['wmname'] or
                 win.window.get_wm_window_role() in self.float_rules['role'] or
                 set(win.window.get_wm_class()) & self.float_rules['wmclass'] or
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -31,14 +31,16 @@
 from collections import defaultdict
 
 from libqtile.layout.base import Layout
+from libqtile.log_utils import logger
 
 # sensible floating defaults like in i3
 # https://github.com/i3/i3/blob/ae757c6848b49d7a0423973a39791ba6eca29308/src/manage.c#L449
 # except `A__NET_WM_STATE_MODAL` windows (couldn't figure this one out)
-DEFAULT_FLOAT_RULES  = defaultdict(set, {
-    "wmtype": {"dialog" ,"utility", "toolbar", "splash"},
+DEFAULT_FLOAT_RULES = defaultdict(set, {
+    "wmtype": {"dialog", "utility", "toolbar", "splash"},
     "wmhint": {"fixed_size"},
 })
+
 
 class Floating(Layout):
     """
@@ -95,11 +97,11 @@ class Floating(Layout):
             self.float_rules = DEFAULT_FLOAT_RULES
         # provide support for deprecated `auto_float_types` argument
         if self.auto_float_types:
-            # TODO: deprecated warning
+            logger.warning("`auto_float_types` is deprecated. Use `float_rules` for this.")
             self.float_rules["wmtype"].update(self.auto_float_types)
         # provide support for the deprecated list-structure of `float_rules`
         if isinstance(float_rules, list):
-            # TODO: deprecated warning
+            logger.warning("The list structure of `float_rules` is deprecated. Use a dictionary instead")
             for rule_dict in float_rules:
                 key, val = tuple(rule_dict.items())[0]
                 self.float_rules[key].add(val)
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -28,21 +28,17 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from collections import defaultdict
+
 from libqtile.layout.base import Layout
 
-DEFAULT_FLOAT_WM_TYPES = set([
-    'utility',
-    'notification',
-    'toolbar',
-    'splash',
-    'dialog',
-])
-
-DEFAULT_FLOAT_RULES = [
-    {"role": "About"},
-    {"wmclass": "file_progress"},
-]
-
+# sensible floating defaults like in i3
+# https://github.com/i3/i3/blob/ae757c6848b49d7a0423973a39791ba6eca29308/src/manage.c#L449
+# except `A__NET_WM_STATE_MODAL` windows (couldn't figure this one out)
+DEFAULT_FLOAT_RULES  = defaultdict(set, {
+    "wmtype": {"dialog" ,"utility", "toolbar", "splash"},
+    "wmhint": {"fixed_size"},
+})
 
 class Floating(Layout):
     """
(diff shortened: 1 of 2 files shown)


libqtile/layout/floating.py
```diff
@@ -28,20 +28,18 @@
...
+from libqtile.layout.base import Layout
+from libqtile.log_utils import logger
 
-DEFAULT_FLOAT_RULES = [
-    {"role": "About"},
-    {"wmclass": "file_progress"},
-]
+# sensible floating defaults like in i3
+# https://github.com/i3/i3/blob/ae757c6848b49d7a0423973a39791ba6eca29308/src/manage.c#L449
+# except `A__NET_WM_STATE_MODAL` windows (couldn't figure this one out)
```
tych0, line 38: Isn't thinks just "wmclass": "__NET_WM_STATE_MODAL"?

tych0: Or maybe we do some fancy processing, so "wmclass": "modal"?

libqtile/layout/floating.py
```diff
@@ -88,19 +86,41 @@ def __init__(self, float_rules=None, no_reposition_match=None, **config):
...
         Layout.__init__(self, **config)
+        self.add_defaults(Floating.defaults)
         self.clients = []
         self.focused = None
         self.group = None
-        self.float_rules = float_rules or DEFAULT_FLOAT_RULES
         self.no_reposition_match = no_reposition_match
-        self.add_defaults(Floating.defaults)
+
+        if override_default_float_rules:
```
tych0, line 95: Is there a reason to have this flag? It seems sort of clunky.

libqtile/layout/floating.py
```diff
@@ -88,19 +86,41 @@ def __init__(self, float_rules=None, no_reposition_match=None, **config):
...
+            self.float_rules = defaultdict(set)
+        else:
+            self.float_rules = DEFAULT_FLOAT_RULES
+        # provide support for deprecated `auto_float_types` argument
+        if self.auto_float_types:
+            logger.warning("`auto_float_types` is deprecated. Use `float_rules` for this.")
+            self.float_rules["wmtype"].update(self.auto_float_types)
+        # provide support for the deprecated list-structure of `float_rules`
+        if isinstance(float_rules, list):
+            logger.warning("The list structure of `float_rules` is deprecated. Use a dictionary instead")
```
tych0, line 105: Let's add a description of how to update your config to /CHANGELOG

libqtile/layout/floating.py
pmkap: The idea is to provide the user with the possibility to only use her own rules or no floating rules at all. I added this because of issue #1275 .

pmkap: I admit it is sort of clunky. Maybe a shorter name appears less clunky.

tych0: Can't we just always set the floating rules to whatever the config contains? If it's unspecificed, it'll get set to what's in the default config, and if it's specified, we want to respect the user's settings.

The way we did it before was kind of silly, with an out-of-band configuration thing.

libqtile/layout/floating.py
pmkap: I am not sure about this. There is a method `get_net_wm_state` in here https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/xcbq.py#L603 , but it returns a tuple.

libqtile/layout/floating.py
pmkap: Almost all users will want the preset rules because they are so basic. That's why I think it's good to have the user make an extra step to override them. Most users just want to add some specific applications.

Without the override-flag, these basic rules need to be in the users config, so we could add them to the `default_config.py`. But then the behaviour for people with older config files changes.

libqtile/layout/floating.py
pmkap: Will do!

libqtile/layout/floating.py
tych0: > That's why I think it's good to have the user make an extra step to override them.

They already said they want to override them when they set custom rules. Making them set it again is just silly. Inevitably someone will miss it, and they'll ask me about it, and I'll have to grep through the code and say "oh you have to set this other flag too" :). Let's add them to default_config.py, and break older users. As long as there's a clear note in `/CHANGELOG` about how to forward port things, I think it's an improvement.

pmkap: It's not a hard break anyway. And probably not affecting too many.

ramnes: Also, if users want to just extend the rules, they can simply do something in the lines of:

```python
floating = Floating()
floating.float_rules.append({...})
```

or if we prefer your defaultdict API:

```
floating = Floating()
floating.float_rules["wmtype"].append(...)
```

This would requires to use lists rather than sets though, as sets aren't extensible.

libqtile/layout/floating.py
ramnes, line 105: I don't think it's a worthwhile change. It will require users to change their configs while not bringing any value to my eyes.

I do understand that you find this cleaner, but it's a subjective opinion and I need an objective argument to agree to this change and to the request for a config update that we will have to make to all Qtile users.

libqtile/layout/floating.py
```diff
@@ -88,19 +86,41 @@ def __init__(self, float_rules=None, no_reposition_match=None, **config):
...
+        Returns `True` if window should float.
+        """
+        if (win.name in self.float_rules['wmname'] or
+                win.window.get_wm_window_role() in self.float_rules['role'] or
+                set(win.window.get_wm_class()) & self.float_rules['wmclass'] or
+                win.window.get_wm_type() in self.float_rules['wmtype']):
             return True
-        for rule_dict in self.float_rules:
-            if win.match(**rule_dict):
+        if "fixed_size" in self.float_rules['wmhint']:
```
ramnes, line 122: Do I understand correctly that `fixed_size` is just a "special" value that you can give to `float_rules["wmhint"]` to trigger that second check, and not a real hint? If so, this doesn't seem like a good API to me. :(

pmkap: Yes. To check this we need minimum and maximum size in WM_NORMAL_HINTS.

pmkap: If we want to give the user the possibiltiy to set this check, it needs to be `float_rules` somehow. Else we have to hardcode this check.

pmkap: But it is in the spec:
> Windows can indicate that they are non-resizable by setting minheight = maxheight and minwidth = maxwidth in the ICCCM WM_NORMAL_HINTS property. The Window Manager MAY decorate such windows differently. 

https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#NORESIZE
So it's just implementing the API in my opinion.

ramnes: Yeah, it's great that we handle this scenario, but what I don't like is that we need to use a special value with a special case that can not be guessed.

One solution could be that we allow functions inside float_rules, for example. We could have something like:

```python
DEFAULT_FLOAT_RULES = [
    lambda w: w.window.get_wm_type() in ("dialog", "utility", "toolbar", "splash"),
    lambda w: w.hints["max_width"] == w.hints["min_width"],
]
```

That would also have the advantage of just being an addition to the current API, and not a change, as the structure of the rules would stay the same (a list). We would just have to check if the rule is a dict and if so, use `win.match`, or a callable and if so, call it.

pmkap: I think the `fixed_size` value isn't that special as it is mentioned in the X specification. If we put it in the docs, there is no need for guessing.

But I also like the idea with the lambdas because it would allow the user to set very custom rules.

PR #1682, opened 2020-04-27
keyboard: don't set keyboard every second

### LINKED ISSUE #1446: Firefox cpu usage escalates and then stops responding only when run in qtile
Hi, 

When starting firefox (firefox 69.0-1, qtile 0.14.2-2) its cpu usage starts rising to the point were the interface becomes unresponsive, also only parts of firefox interface are drawn, but like after 5 mins the window is not visible at all, may be the screenshots will make things clearer. 

screenshots with description:
https://imgur.com/a/V17wJrM

Nothing is there in the logs (I only checked .local/share/qtile/qtile.log), this does not happen with me under i3 or bspwm. I also use optirun with my browsers but it does not matter if I start firefox without it. this bug happens every time I use firefox, and I use a lot of gtk3, gtk2, qt4, qt5, electron apps and it only happens with firefox on qtile only.

I am not sure what other information I can provide.

### LINKED ISSUE #1616: KeyboardLayout widget re-applies layout every two seconds
# Issue description

When the KeyboardLayout widget is loaded, the command `setxkbmap -verbose 10` is executed every two seconds.This command causes a `MappingNotify` message to be broadcast, which can cause issues in other programs.

A good example for this is the kitty terminal emulator, which tries to releoad the keyboard layout and currently has an issue which doesn't remember the modifier key states when doing so.

To reproduce this problem:
* Start qtile with the keyboard layout widget present
* Start kitty
* Open nano and load a large file
* Press and hold Ctrl and press k (cut line) repeatadly

As soon as the `setxkbdmap` call is run, the ctrl key stops working untill released and re-pressed.

While I'm not exactly sure this is solely a qtile bug, I have a feeling that other applications have similar issues and fixing the issue in Kitty is just going to patch one instance of this. Instead of calling `setxkbdmap` itself, which re-applies the current configuration, we should rather call `setxkbdmap -print` or `-query` which will only report the current settings. This is also preferable from a performance standpoint.

# Qtile version

qtile 0.15.0 as published on pypi.org

tych0:
Without the -query parameter, setxkbmap causes a cascade of X events that
cause problems in other applications. Additionally, it re-sets custom
keyboard layouts, which is annoying :)

Closes #1616
Closes #1446

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -97,7 +97,7 @@ class KeyboardLayout(base.InLoopPollText):
         Examples: "us", "us dvorak".  In case of error returns "unknown".
         """
         try:
-            command = 'setxkbmap -verbose 10'
+            command = 'setxkbmap -verbose 10 -query'
             setxkbmap_output = self.call_process(command.split(' '))
             keyboard = self.get_keyboard_layout(setxkbmap_output)
             return str(keyboard)


ramnes:
Yet another no brainer, thanks! Kinda baffled it took us that much time for such a one line diff. :(

PR #1688, opened 2020-04-28
Move setup of qtile object within try/finally test block

flacjacket:
Occasionally, when ctrl-C'ing the running test suite, it will cause the tests to hang and have to be killed manually. This seems to improve this behavior and better cleans up when the qtile object setup is cancelled.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -944,7 +944,10 @@ class Connection:
         return Window(self, wid)
 
     def disconnect(self):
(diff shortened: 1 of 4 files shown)


ramnes:
I think it's a safe merge, although I'm not seeing any improvement, tests still hang when I try to kill them. What test did you try the C-c on?

flacjacket:
Hmmm, alright, I had just tried it a few random times and didn't see it, so not sure that this totally addresses all of the possible failure conditions.

Issue #1346, opened 2019-06-17
Qtile crash at startup


max-k:
Qtile crash almost every time I restart my computer.

If I restart only Qtile, everything is working properly.

System informations:

* Os: Archlinux
* Qtile : 0.13.0  - also tested with latest git version [commit a83257b7](https://github.com/qtile/qtile/commit/a83257b77af2aad1c263ae177e735ac9798b3938) -
* Python: 3.7.3
* Kernel: 5.1.10-hardened

Here is the log output:
```
2019-06-17 21:43:18,422 libqtile hook.py:fire():L371  Internal event: startup((), {})
2019-06-17 21:43:18,429 libqtile hook.py:fire():L371  Internal event: startup_complete((), {})
2019-06-17 21:43:18,430 libqtile ipc.py:start():L247  Starting server
2019-06-17 21:43:18,506 libqtile qtile.py:main():L131  Qtile crashed
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/libqtile/scripts/qtile.py", line 129, in main
    q.loop()
  File "/usr/lib/python3.7/site-packages/libqtile/manager.py", line 806, in loop
    self.server.start()
  File "/usr/lib/python3.7/site-packages/libqtile/ipc.py", line 248, in start
    self.server = self.loop.run_until_complete(server_coroutine)
  File "/usr/lib/python3.7/asyncio/base_events.py", line 582, in run_until_complete
    raise RuntimeError('Event loop stopped before Future completed.')
RuntimeError: Event loop stopped before Future completed.
2019-06-17 21:43:18,509 libqtile qtile.py:main():L132  Exiting...
```

Symptoms are always the same, I get my screen layout but Qtile isn't responding.

zordsdavini:
it looks the same as in https://github.com/qtile/qtile/issues/1326

max-k:
I'll investigate a few days more but it seems fixed in Qtile 0.14.0.

max-k:
In fact it's still not working.

zordsdavini:
does it crash with default config?

max-k:
Hi, I found where was the problem.

It came from an old hook (still documented [here](http://docs.qtile.org/en/latest/manual/ref/hooks.html#libqtile.hook.subscribe.screen_change)) I took from official documentation.

I think it was triggered at startup preventing Qtile to start properly.

I disabled it and everything is working again.

max-k:
[...]

PR #1689, opened 2020-04-28
Use additional async functionality in setup

flacjacket:
Move more of the setup functionality into async. I think it'd be great if nothing needed to be passed a loop in the constructor and all of the async initialization could be called in async tasks, which would then be able to automatically get the running loop. Some additional refactoring could help with this. This may set us on a path that the Qtile object can be easily re-made from within the session manager if we can pull out enough of the direct loop handling from that.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -236,18 +236,20 @@ class Qtile(CommandObject):
             logger.warning("importing dbus/gobject failed, dbus will not work.")
             self._glib_loop = None
 
-    async def async_loop(self):
+    async def async_loop(self) -> None:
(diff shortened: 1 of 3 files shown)


tych0:
Will stuff like this fix #1346?

flacjacket:
It might, but I can't quite get this to work as expected. I pulled some of the server setup functionality out into #1705, which is a start.

flacjacket:
Alright, I currently have this stacked on top of #1706 and #1705, and it is working properly now! Not totally sure it'd fix that bug, but it feels like a step in the right direction.

PR #1690, opened 2020-04-28
base.deprecated does not come up in the logs

ramnes:
Not sure why, but I figured out that I had deprecation errors due to my config, but they weren't showing up in my logs. @tych0 I know you made the `base.deprecated` function in the first place, do you think it's still needed? Or can we use `logger.warning` directly, as I'm doing here? Another option would be to use `logger.warning` inside `base.deprecated`, but using `logger.warning` directly sounds like a more intuitive option to me.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -31,7 +31,6 @@
 
 import subprocess
 import threading
-import warnings
 from typing import Any, List, Tuple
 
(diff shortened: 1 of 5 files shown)


tych0:
I think the intent was to have something that people could use automated tooling for, but if it doesn't even show up in the logs, that's not useful. This lgtm, thanks!

PR #1692, opened 2020-04-28
Revamp config and restart

ramnes:
The context of this PR is that for some reason, validating the config for each `cmd_restart` isn't a good idea with my config, as it used `cmd_restart` as a screen change hook, and it made qtile enter into an infinite restart loop. The easy way would have been to just merge the last commit of this PR, or to add a `force` argument to `cmd_restart`, but I really wanted also to simplify the validation mechanism and add a `cmd_validate_config` also.

The PR could have been separated in three PR:
* Revamp confreader
* Add cmd_validate_config
* Decouple the cmd_restart from the actual restart

I decided to make it a single PR because the last two points depend on the first one. I highly recommend to any potential reviewer to go through all commits one by one rather than looking at the overall changes.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -24,6 +24,7 @@
 # SOFTWARE.
 import os
 import sys
+from typing import Optional
 
 from libqtile.backend import base
 
@@ -78,7 +79,7 @@ class Config:
             setattr(self, key, value)
 
     @classmethod
-    def from_file(cls, kore: base.Core, path: str):
+    def from_file(cls, path: str, kore: Optional[base.Core] = None):
         "Create a Config() object from the python file located at path."
         cnf = cls(file_path=path, kore=kore)
         cnf.load()
@@ -104,7 +105,8 @@ class Config:
             raise ConfigError(tb)
 
         self.update(**vars(config))
-        self.validate()
+        if self.kore:
+            self.validate()
 
     def validate(self) -> None:
         """
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -53,15 +53,21 @@ class Config:
         "wmname",
     ]
 
-    def __init__(self, file_path=None, **settings):
+    def __init__(self, file_path=None, kore=None, **settings):
         """Create a Config() object from settings
 
         Only attributes found in Config.settings_keys will be added to object.
         config attribute precedence is 1.) **settings 2.) self 3.) default_config
         """
+        self.file_path = file_path
+        self.kore = kore
+        self.update(**settings)
+
+    def update(self, *, fake_screens=None, **settings):
         from libqtile.resources import default_config
 
-        self.file_path = file_path
+        if fake_screens:
+            self.fake_screens = fake_screens
 
         default = vars(default_config)
         for key in self.settings_keys:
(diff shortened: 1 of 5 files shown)


ramnes:
PS: I still don't have an explanation as to why the config validation triggers my screen change hook (maybe it is that fake_screens thing?), but I'd rather open a dedicated issue / PR for this.

pawamoy:
Nice :slightly_smiling_face:  Would it be worth to point to the log in the notification? "Configuration error: ... Details in /path/to/log.qtile", with an approximated time or a string to grep? Just a suggestion!

ramnes:
Yeah, maybe once we get @m-col work on notification popups merged in? 

Meanwhile I quite like that it stays on one line, so that if you use the Qtile notification widget, you get something readable:
image

ramnes:
Or maybe we should modify the notification widget so that it only displays the last line of the notification message? Or remove the line breaks? Because otherwise it simply does not fit in a bar anyway.

tych0:
I think you probably have restart-on-screen-change because we don't have real screen-add so we recommend that people restart to add bars and widgets to new screens.

ramnes:
> I think you probably have restart-on-screen-change because we don't have real screen-add so we recommend that people restart to add bars and widgets to new screens.

I know that, right? :smile:

libqtile/confreader.py
```diff
@@ -53,15 +54,21 @@ class Config:
...
 
-    def __init__(self, file_path=None, **settings):
+    def __init__(self, file_path=None, kore=None, **settings):
         """Create a Config() object from settings
 
         Only attributes found in Config.settings_keys will be added to object.
         config attribute precedence is 1.) **settings 2.) self 3.) default_config
         """
+        self.file_path = file_path
+        self.kore = kore
```
flacjacket, line 64: Having the config reader hold the core object seems a bit heavy handed here. Is this something that can be passed in to `load` and `validate`? That would also give some finer control over which instances the validation is run.

libqtile/confreader.py
```diff
@@ -53,15 +54,21 @@ class Config:
...
+        self.update(**settings)
+
+    def update(self, fake_screens=None, **settings):
```
flacjacket, line 67: It would be nice to guard `fake_screens` as a keyword-only argument.

libqtile/core/manager.py
```diff
@@ -38,7 +38,6 @@
 
 from libqtile import command_interface, hook, utils, window
 from libqtile.backend.x11 import xcbq
-from libqtile.backend.x11.xcore import XCore
```
flacjacket, line 41: This is good to be getting out of here.

libqtile/confreader.py
ramnes: Why do you find this heavy handed? I find it boring to pass it again if we already gave it at instanciation. Also, this is what permits to remove the XCore import from libqtile/core/manager.py :)

libqtile/confreader.py
ramnes: Sounds good!

ramnes:
I'll go ahead and merge this, I use it daily, it works great, and I can't live without cmd_validate_config anymore. :)

libqtile/confreader.py
flacjacket: The config object should not have to know what backend core it is using, if it is needed for the sake of validation that is one thing. The import you mention should never have been there in the first place, that should have been using the core that is attached to the manager to do the verification rather than creating an XCore.

flacjacket: We can see as we pull out X-specific functionality if it makes sense to leave this like this or if it will help for it to be passed in on verification.

PR #1695, opened 2020-04-29
setup: use setuptools_scm to get git hash in version number

### LINKED ISSUE #1681: qtile version
As we ask users to say version during posting new issue, it would be better to have git version included. Especially in latest version.
It could be in format: 
```
$ qtile --version
0.15.0 (3449.01b0bc06)
```

airwoodix:
Fixes #1681.

Rationale:

  - use [setuptools_scm](https://pypi.org/project/setuptools-scm), stick to [PEP-0440](https://www.python.org/dev/peps/pep-0440) format

  - in sphinx, use version number [from metadata](https://github.com/pypa/setuptools_scm#usage-from-sphinx). If this turns out not to be what one wants, we can set the version [in setup.py](https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 [metadata]
 name = qtile
-version = 0.15.0
 url = http://qtile.org
 license = MIT
 license_file = LICENSE
@@ -42,6 +41,7 @@ setup_requires =
   cffi >= 1.1.0
   cairocffi[xcb] >= 0.9.0
   setuptools >= 40.5.0
+  setuptools_scm >= 3.2.0
 install_requires =
   cffi >= 1.1.0
   cairocffi[xcb] >= 0.9.0
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,7 +12,7 @@
 # serve to show the default.
 
 import os
-import pkg_resources
+import setuptools_scm
 import sys
 from unittest.mock import MagicMock
 
@@ -98,7 +98,7 @@ copyright = u'2008-2020, Aldo Cortesi and contributers'
 # built documents.
 #
 # The short X.Y version.
-version = pkg_resources.get_distribution("qtile").version
+version = setuptools_scm.get_version(root="..")
 # The full version, including alpha/beta/rc tags.
 release = version
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,7 @@
 # serve to show the default.
 
 import os
+import pkg_resources
 import sys
 from unittest.mock import MagicMock
 
@@ -97,7 +98,7 @@ copyright = u'2008-2020, Aldo Cortesi and contributers'
 # built documents.
 #
 # The short X.Y version.
-version = '0.15.0'
+version = pkg_resources.get_distribution("qtile").version
 # The full version, including alpha/beta/rc tags.
 release = version
 
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
+setuptools_scm
 sphinx
 sphinx_rtd_theme
 sphinxcontrib-seqdiag
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,7 @@ def get_cffi_modules():
 
 setup(
     cmdclass={'install': CheckCairoXcb},
+    use_scm_version=True,
     cffi_modules=get_cffi_modules(),
     install_requires=["cffi>=1.0.0"],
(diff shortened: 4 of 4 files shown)


zordsdavini:
It looks like docs still have old version

setup.cfg
```diff
@@ -1,6 +1,5 @@
 [metadata]
 name = qtile
-version = 0.15.0
```
ramnes, line 3: Should we replace this by something in setup.py?

zordsdavini: replaced by setuptools_scm
```
HEAD is now at ba771bab Release 0.15.1
 arnas@zordmachine  ~/out/q/scm  ➦ ba771bab  python setup.py --version
0.15.1
```

ramnes:
Let's fix Travis and merge. :)

airwoodix:
Is `build_sphinx` the official way to build the docs? (according to the unit test yes)

Two options to fix CI:
  - install `qtile` in Travis env;
  - override version for sphinx in `setup.py` ([ref](https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html)). The problem with this is that you need sphinx to run `setup.py`

Preferences?

ramnes:
Installing qtile seems fine to me. I guess we could replace the `-r ../requirements.txt` in the `docs/requirements.txt` we just added yesterday with @zordsdavini by a simple `..` or `-e ..`?

airwoodix:
We're touching the limits of ``pip`` here, because ``xcffib`` and ``cairocffi`` (and now ``sphinx``) need to be installed by hand before (as in ``dev.sh``). Somehow `pip` doesn't realize that their installation is pending before evaluating `setup.py` for `qtile`. 
Or am I missing something here?

airwoodix:
> Ah yes, I guess we are getting blocked by CheckCairoXcb?

Indeed.

I have the impression that an attempt at bypassing it will by ugly, or do you have an idea? But making it non fatal is also not very satisfactory. Do you know how to print warnings from the setup? (such that the `CheckCairoXcb`'s output is coloured if it fails)

zordsdavini:
Maybe this would not be very hakish?
https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html
```
command_options={
        'build_sphinx': {
            'version': ('setup.py', version),
``` 
but maybe that should be added under `try` as I've added pre requires for phinx in setup.py

ramnes:
Ahah, that is terrible. But what about using `setuptools_scm.get_version` here?

zordsdavini:
this worked for me https://gist.github.com/zordsdavini/fea21e7f3c896a837f69fa2a06578e20

airwoodix:
But this breaks building the docs with the ``docs/Makefile`` (in: the version number will be wrong). Should we remove the Makefile and `docs/build.sh`, and update 'docs/BUILDING'?

ramnes:
I think it would be better to keep the `docs/Makefile`. Can't we call `setuptools_scm.get_version()` directly inside the doc somehow? Sorry, I can not really test by myself at the moment.

airwoodix:
This is explicitly advised against [in the documentation](https://github.com/pypa/setuptools_scm#usage-from-sphinx) but works in this case. Both Makefile and ``setup.py build_sphinx`` now work.

tych0:
"The underlying reason is, that services like Read the Docs sometimes change the working directory for good reasons and using the installed metadata prevents using needless volatile data there."

But will read the docs? :)

airwoodix:
@tych0 ^^ I guess you don't have that problem and upload the static html content to the machine serving docs.qtile.org directly, correct?

tych0:
No, we don't. It's all built dynamically on read the docs. See https://readthedocs.org/projects/qtile/

tych0:
Although if both this and @zordsdavini's PRs are limited by RTD, it may be time to re-evaluate.

airwoodix:
Any idea if RTD uses/can use ``build_sphinx`` to build the docs? If so, I'd go for overriding the version in `setup.py` and remove the Makefile and build script in `docs`.

zordsdavini:
no, RTD uses direct sphix call but with .readthedocs.yml you can make it to install libqtile.

tych0:
Trying github's fancy new UI to resolve conflicts... :)

zordsdavini:
@airwoodix can you rebase with upstream/master?

ramnes:
Thanks! :rocket:

PR #1697, opened 2020-04-29
Multiple borders

m-col:
This PR lets qtile draw multiple borders on windows. Changes:

`xcbq.Window` does all border handling itself rather than `libqtile.window.Window` in a `paint_borders` method, and it gets a `set_borderpixmap` method which accepts a pixmap which will be used to draw window borders. A fun side effect of this is that the `paint_borders` method can be overriden for alternative border styles ([proof of concept](https://github.com/m-col/qtools/blob/master/qtools/borders/borders.py)).

Multiple borders can be painted by passing a layout with "border_width" parameter a list/tuple of ints and the same number of colours to any colour parameters the layout has (i.e. any other "border_*" parameter). Borders can therefore be different widths. The parameters correspond to borders from outside inward. As before, layouts will accept an int width or str colour so configs are not broken; they are always converted to tuples when the layout is instantiated. Tuples are used so that `Qtile.color_pixel` will accept them (lru_cache needs hashable args).

All layouts that use borders have been updated to allow for multiple borders.

### DIFF
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -100,7 +100,7 @@ for i in groups:
     ])
 
 layouts = [
-    layout.Columns(border_focus_stack='#d75f5f'),
+    layout.Columns(border_focus_stack=['#d75f5f', '#8f3d3d'], border_width=4),
     layout.Max(),
     # Try more layouts by unleashing below layouts.
     # layout.Stack(num_stacks=2),
--- a/test/layouts/test_common.py
+++ b/test/layouts/test_common.py
@@ -452,3 +452,25 @@ def test_cycle_layouts(manager):
         # Use manager.c.layout.info()['name'] in the assertion message, so we
         # know which layout is buggy
         assert manager.c.window.info()['name'] == "three", manager.c.layout.info()['name']
+
+
+class AllLayoutsMultipleBorders(AllLayoutsConfig):
+    """
+    Like AllLayouts, but all the layouts have border_focus set to a list of colors.
+    """
+    layouts = [
+        layout_cls(border_focus=['#000' '#111', '#222', '#333', '#444'])
+        for layout_name, layout_cls in AllLayoutsConfig.iter_layouts()
+    ]
+
+
+@pytest.mark.parametrize("manager", [AllLayoutsMultipleBorders], indirect=True)
+def test_multiple_borders(manager):
+    manager.test_window("one")
+    manager.test_window("two")
+
+    initial_layout_name = manager.c.layout.info()['name']
+    while True:
+        manager.c.next_layout()
+        if manager.c.layout.info()['name'] == initial_layout_name:
+            break
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+Qtile xxxxxx, released xxxxxxxxxx:
+    * features
+        - All layouts will accept a list of colors for border_* options with which
+          they will draw multiple borders on the appropriate windows.
+
 Qtile 0.18.0, released 2021-07-04:
     !!! Config breakage !!!
         - The `qtile` entry point doesn't run `qtile start` by default anymore
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,7 @@ MOCK_MODULES = [
     'xcffib',
     'xcffib.randr',
     'xcffib.render',
+    'xcffib.wrappers',
     'xcffib.xfixes',
     'xcffib.xinerama',
     'xcffib.xproto',
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -271,8 +271,8 @@ class Window(_Window, metaclass=ABCMeta):
     def get_pid(self) -> int:
         """Return the PID that owns the window."""
 
-    def paint_borders(self, color, width) -> None:
-        """Paint the window borders with the given color and width"""
+    def paint_borders(self, color: Union[ColorType, List[ColorType]], width: int) -> None:
+        """Paint the window borders with the given color(s) and width"""
 
     @abstractmethod
     def cmd_focus(self, warp: bool = True) -> None:
--- a/libqtile/backend/wayland/output.py
+++ b/libqtile/backend/wayland/output.py
@@ -160,30 +160,36 @@ class Output(HasListeners):
         transform_matrix = self.transform_matrix
 
         if window.borderwidth:
-            bw = window.borderwidth * scale
+            bw = int(window.borderwidth * scale)
 
             if surface == window.surface.surface:
-                bc = window.bordercolor
-                border = Box(
-                    int(x),
-                    int(y),
-                    int(width + bw * 2),
-                    int(bw),
-                )
-                x += bw
-                y += bw
-                self.renderer.render_rect(border, bc, transform_matrix)  # Top border
-                border.y = int(y + height)
-                self.renderer.render_rect(border, bc, transform_matrix)  # Bottom border
-                border.y = int(y - bw)
-                border.width = int(bw)
-                border.height = int(height + bw * 2)
-                self.renderer.render_rect(border, bc, transform_matrix)  # Left border
-                border.x = int(x + width)
-                self.renderer.render_rect(border, bc, transform_matrix)  # Right border
-            else:
-                x += bw
-                y += bw
+                outer_w = width + bw * 2
+                outer_h = height + bw * 2
+                num = len(window.bordercolor)
+                bws = [bw // num] * num
+                for i in range(bw % num):
+                    bws[i] += 1
+                coord = 0
+                for i, bc in enumerate(window.bordercolor):
+                    border = Box(
+                        int(x + coord),
+                        int(y + coord),
+                        int(outer_w - coord * 2),
+                        int(bws[i]),
+                    )
+                    self.renderer.render_rect(border, bc, transform_matrix)  # Top border
+                    border.y = int(y + outer_h - bws[i] - coord)
+                    self.renderer.render_rect(border, bc, transform_matrix)  # Bottom border
+                    border.y = int(y + coord)
+                    border.width = int(bws[i])
+                    border.height = int(outer_h - coord * 2)
+                    self.renderer.render_rect(border, bc, transform_matrix)  # Left border
+                    border.x = int(x + outer_w - bws[i] - coord)
+                    self.renderer.render_rect(border, bc, transform_matrix)  # Right border
+                    coord += bws[i]
+
+            x += bw
+            y += bw
 
         box = Box(
             int(x),
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -83,7 +83,7 @@ class Window(base.Window, HasListeners):
         self._mapped: bool = False
         self.x = 0
         self.y = 0
-        self.bordercolor: ffi.CData = _rgb((0, 0, 0, 1))
+        self.bordercolor: List[ffi.CData] = [_rgb((0, 0, 0, 1))]
         self.opacity: float = 1.0
         self._outputs: List[Output] = []
 
@@ -294,9 +294,14 @@ class Window(base.Window, HasListeners):
             if switch_group:
                 group.cmd_toscreen(toggle=False)
 
-    def paint_borders(self, color: ColorType, width) -> None:
+    def paint_borders(self, color: Union[ColorType, List[ColorType]], width) -> None:
         if color:
-            self.bordercolor = _rgb(color)
+            if isinstance(color, list):
+                if len(color) > width:
+                    color = color[:width]
+                self.bordercolor = [_rgb(c) for c in color]
+            else:
+                self.bordercolor = [_rgb(color)]
         self.borderwidth = width
 
     @property
@@ -720,7 +725,7 @@ class Static(base.Static, Window):
         self.x = 0
         self.y = 0
         self.borderwidth: int = 0
-        self.bordercolor: ffi.CData = _rgb((0, 0, 0, 1))
+        self.bordercolor: List[ffi.CData] = [_rgb((0, 0, 0, 1))]
         self.opacity: float = 1.0
         self._outputs: List[Output] = []
         self._float_state = FloatStates.FLOATING
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING
 
 import xcffib
 import xcffib.xproto
+from xcffib.wrappers import GContextID, PixmapID
 from xcffib.xproto import EventMask, SetMode, StackMode
 
 from libqtile import hook, utils
@@ -285,7 +286,7 @@ class XWindow:
 
     def configure(self, **kwargs):
         """
-        Arguments can be: x, y, width, height, border, sibling, stackmode
+        Arguments can be: x, y, width, height, borderwidth, sibling, stackmode
         """
         mask, values = xcbq.ConfigureMasks(**kwargs)
         # older versions of xcb pack everything into unsigned ints "=I"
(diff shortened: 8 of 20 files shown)


ramnes:
This seems to be a big breaking change as is. I'd be in favor of supporting either a string/int or a list/tuple everywhere, what do you think? It shouldn't be too heavy with something in the lines of:

```python
if not isinstance(parameter, list):
    parameter = list(parameter)
```

m-col:
If you mean for configuring layouts, then that is implemented here. All layouts still accept an int for border_width and strings for the colors (and also support lists). These are then converted to tuples - see the addition at layout.base.check_config - so it doesn't break configs. Unless you've noticed something else that I haven't?

ramnes:
I went a bit too fast and didn't see that, thanks for mentioning it!

m-col:
Phew! I'm just looking through the failed tests now to figure out what's happened but I'm not too familiar with them (or tests in general).

m-col:
OK got the tests fixed :)

tych0:
One thing that's not clear to me from the patch notes is: why? :). What's the motivation for supporting this?

m-col:
Because it's awesome! However, it was not asked for so I totally understand if you don't think it should be merged. A quick illustration to put forward my case... borders like these are cool:



Further, the bringing together of all border-painting logic and ability to specify multiple border colours and widths makes it easy to create custom border patterns by overriding `xcbq.Window.paint_borders`. For example, I created a custom pattern to paint them like this (fvwm/CDE style):



Like I said no worries at all if you don't think it should be part of Qtile. I tried to figure out a way to use custom border patterns without modifying the Qtile code itself but the limitation was that states indicated by a layout's colours (focussed vs normal) was limited to a single colour, so I needed to pass layouts multiple colours. It didn't make much sense to allow multiple colours without something internal to Qtile to use them so I set up multiple borders (as, essentially, a "default" border pattern) to use them and be consistent with the normal single-border appearance.

tych0:
On Sat, May 02, 2020 at 11:22:22AM -0700, m-col wrote:
> Because it's awesome! However, it was not asked for so I totally understand if you don't think it should be merged. A quick illustration to put forward my case... borders like these are cool:
> 
> 
> 
> Further, the bringing together of all border-painting logic and ability to specify multiple border colours and widths makes it easy to create custom border patterns by overriding `xcbq.Window.paint_borders`. For example, I created a custom pattern to paint them like this (fvwm/CDE style):
> 
> 
> 
> Like I said no worries at all if you don't think it should be part of Qtile. I tried to figure out a way to use custom border patterns without modifying the Qtile code itself but the limitation was that states indicated by a layout's colours (focussed vs normal) was limited to a single colour, so I needed to pass layouts multiple colours. It didn't make much sense to allow multiple colours without something internal to Qtile to use them so I set up multiple borders (as, essentially, a "default" border pattern) to use them and be consistent with the normal single-border appearance.

I don't mind having it in the repo, "because it's awesome" is a
perfectly reasonable answer (we have merged much crazier things, IMO),
I was mostly just curious if I was missing something :)

libqtile/backend/x11/xcbq.py
```diff
@@ -802,6 +803,52 @@ def grab_pointer(self, owner_events, event_mask, pointer_mode,
...
+        for i in range(len(borderwidths)):
+            core.ChangeGC(gc, xcffib.xproto.GC.Foreground, [bordercolors[i]])
+            rect = xcffib.xproto.RECTANGLE.synthetic(
+                coord, coord, outer_w - coord * 2, outer_h - coord * 2
+            )
+            core.PolyFillRectangle(pixmap, gc, 1, [rect])
+            coord += borderwidths[i]
+
+        self.set_borderpixmap(pixmap, gc, outer_w, outer_h)
+        core.FreePixmap(pixmap)
```
tych0, line 833: These will all leak if something in the middle throws an exception; can we wrap this in some kind of try: finally or context handler or something?

libqtile/backend/x11/xcbq.py
```diff
@@ -802,6 +803,52 @@ def grab_pointer(self, owner_events, event_mask, pointer_mode,
...
+        )
+        borderwidth = self.borderwidth
+        most_w = outer_w - borderwidth
+        most_h = outer_h - borderwidth
+        core.CopyArea(pixmap, border, gc, borderwidth, borderwidth, 0, 0, most_w, most_h)
+        core.CopyArea(pixmap, border, gc, 0, 0, most_w, most_h, borderwidth, borderwidth)
+        core.CopyArea(pixmap, border, gc, borderwidth, 0, 0, most_h, most_w, borderwidth)
+        core.CopyArea(pixmap, border, gc, 0, borderwidth, most_w, 0, borderwidth, most_h)
+        core.ChangeWindowAttributes(self.wid, xcffib.xproto.CW.BorderPixmap, [border])
+        core.FreePixmap(border)
```
tych0, line 850: Some similar leakage here. I wonder if we shouldn't add some kind of wrapper for doing this sort of thing to xcffib.

m-col: try-finally might be the most simple approach - just 4 extra lines, 2 for each of these methods. I took a look at xcffib to see if I could submit a context manager PR but I couldn't wrap my head around how it worked. Maybe we could do the try-finally for the time being, and if an xcffib wrapper seems like a good idea in the future for something else then it could be implemented and then applied here too. What do you think?

tych0: Happy to do it whatever way you think is best. We can chat about how xcffib works on IRC if you like, but no worries if you think it's simpler to just do here.

m-col:
OK so I've added the try-finally blocks and everything looks good to go from this end. Been using this for a while with no issues :)

m-col:
Alrighty I have purged the original PR of lots of ugly changes and reduced it down lots. I was testing the changes in the first commit - which are not specifically related to multiple borders - using the changes in the second commit, which is why they are in the same PR. I can of course submit the first commit separately if the second one isn't wanted :)

With these changes, the only thing a user needs to do to enable multiple borders is to supply multiple colours to a layout (list/tuple) instead of a string for just one.

I'd like to add a test for this and for that I'm thinking a pixel comparison must like in test_images2.py so I'll have to figure that out and then test some border changes.

m-col:
I've added a test and updated the layout docstrings ("Border color..." -> "Border color(s)..." in defaults dicts). I also uncommented the Bsp layout in the default config and gave it some multi borders so that people might discover this feature :)

Please could someone review this and see if it's fine to be merged?

The only change I can foresee in the future would be replacing the pixmap and GC try-finally blocks with context managers (https://github.com/tych0/xcffib/issues/109).

m-col:
> The only change I can foresee in the future would be replacing the pixmap and GC try-finally blocks with context managers ([tych0/xcffib#109](https://github.com/tych0/xcffib/issues/109)).

This has been implemented so this PR is finished, pending tests and review :)

m-col:
What do I need to change for the test envs to find `xcffib.wrappers`?? All mentions of xcffib version are set correctly as far as I can tell :|

m-col:
I was mistaken about the current release of xcffib. This PR won't work until the next release of xcffib.

tych0:
Do you need an xcffib release? I know the maintainer :)

m-col:
Oh that is convenient ;) I do, but presumably it would take a while for the
update to trickle down through all of the distros etc?

tych0:
> presumably it would take a while for the update to trickle down through all of the distros etc?

Yes, but we can bump the required version in our requirements.txt and that way at least people won't be able to install without installing the newer version. I also push it to pypi, so it's not that hard. I'll try to tag one in the next few days.

m-col:
Awesome :)

tych0:
Ok, 0.11.0 is up with these wrappers and one other small bugfix.

m-col:
Not sure if I was just slow but the update has trickled down to my system down, so from my end this is good to go. I'm not sure what the state of distros like ubuntu might be regarding the update to xcffib though. Probs fine?

m-col:
An unexpected error in the tests, with the `docs` env. Running it locally I get this error. Any ideas why? xcffib.wrappers is importable just fine.

```
docs create: /home/mcol/git/qtile/.tox/docs
docs installdeps: -r/home/mcol/git/qtile/docs/requirements.txt
WARNING: Discarding $PYTHONPATH from environment, to override specify PYTHONPATH in 'passenv' in your configuration.
docs installed: alabaster==0.7.12,Babel==2.9.0,blockdiag==2.0.1,cairocffi==1.2.0,certifi==2020.12.5,cffi==1.14.4,chardet==4.0.0,docutils==0.16,funcparserlib==0.3.6,idna==2.10,imagesize==1.2.0,Jinja2==2.11.2,MarkupSafe==1.1.1,numpydoc==1.1.0,packaging==20.8,Pillow==8.0.1,pycparser==2.20,Pygments==2.7.3,pyparsing==2.4.7,pytz==2020.4,requests==2.25.1,seqdiag==2.0.0,setuptools-scm==5.0.1,six==1.15.0,snowballstemmer==2.0.0,Sphinx==3.3.1,sphinx-rtd-theme==0.5.0,sphinxcontrib-applehelp==1.0.2,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==1.0.3,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-seqdiag==2.0.0,sphinxcontrib-serializinghtml==1.1.4,urllib3==1.26.2,webcolors==1.11.1,xcffib==0.11.1
docs run-test-pre: PYTHONHASHSEED='3232876511'
docs run-test: commands[0] | python3 setup.py build_sphinx -W
running build_sphinx
Running Sphinx v3.3.1
Failed to import dependencies for notifications: No module named 'gi'

Extension error:
Could not import extension sphinx_qtile (exception: No module named 'xcffib.wrappers'; 'xcffib' is not a package)
ERROR: InvocationError for command /home/mcol/git/qtile/.tox/docs/bin/python3 setup.py build_sphinx -W (exited with code 1)
_________________________________________________________________________________________________________________ summary __________________________________________________________________________________________________________________
ERROR:   docs: commands failed
```

tych0:
I think you need a,

```
diff --git a/docs/conf.py b/docs/conf.py
index d394989d..141e09b4 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,6 +41,7 @@ MOCK_MODULES = [
     'psutil',
     'trollius',
     'xcffib',
+    'xcffib.wrappers',
     'xcffib.randr',
     'xcffib.render',
     'xcffib.xfixes',
```

If you can figure out a way we can get rid of this, that would be great, it's tripped me up before :)

m-col:
NIce, it worked

libqtile/backend/x11/xcbq.py
```diff
@@ -759,9 +760,57 @@ def query_tree(self):
...
+            core.CreatePixmap(
+                self.conn.default_screen.root_depth, border, self.wid, outer_w, outer_h
+            )
+            most_w = outer_w - borderwidth
+            most_h = outer_h - borderwidth
+            core.CopyArea(pixmap, border, gc, borderwidth, borderwidth, 0, 0, most_w, most_h)
+            core.CopyArea(pixmap, border, gc, 0, 0, most_w, most_h, borderwidth, borderwidth)
+            core.CopyArea(pixmap, border, gc, borderwidth, 0, 0, most_h, most_w, borderwidth)
+            core.CopyArea(pixmap, border, gc, 0, borderwidth, most_w, 0, borderwidth, most_h)
+            core.ChangeWindowAttributes(self.wid, xcffib.xproto.CW.BorderPixmap, [border])
```
ramnes, line 813: X11 noob question here: why do we need two pixmaps and copy one into the other? Can't we just make the final one from the start?

ramnes:
Can you do the rebase? :)

libqtile/backend/x11/xcbq.py
```diff
@@ -759,9 +760,57 @@ def query_tree(self):
...
-    def paint_borders(self, color):
-        if color:
-            self.set_attribute(borderpixel=self.conn.color_pixel(color))
+    def paint_borders(self, colors, borderwidth, width, height):
+        if not colors or not borderwidth:
+            return
+
+        if isinstance(colors, str):
+            self.set_attribute(borderpixel=self.conn.color_pixel(colors))
+            return
```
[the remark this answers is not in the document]
m-col: This is a slight optimisation -- using this attribute will be a bit faster than drawing a rect when a window only needs one colour.

libqtile/resources/default_config.py
```diff
@@ -102,9 +102,9 @@
 layouts = [
     layout.Columns(border_focus_stack='#d75f5f'),
     layout.Max(),
+    layout.Bsp(border_focus=['#d75f5f', '#8f3d3d'], border_width=4),
```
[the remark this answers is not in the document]
m-col: Sure thing!

libqtile/backend/x11/xcbq.py
```diff
@@ -759,9 +760,57 @@ def query_tree(self):
             parent = Window(self.conn, q.parent)
         return root, parent, [Window(self.conn, i) for i in q.children]
 
-    def paint_borders(self, color):
-        if color:
-            self.set_attribute(borderpixel=self.conn.color_pixel(color))
+    def paint_borders(self, colors, borderwidth, width, height):
+        if not colors or not borderwidth:
+            return
```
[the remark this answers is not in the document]
m-col: Fair point.

libqtile/backend/x11/xcbq.py
```diff
@@ -759,9 +760,57 @@ def query_tree(self):
             parent = Window(self.conn, q.parent)
         return root, parent, [Window(self.conn, i) for i in q.children]
 
-    def paint_borders(self, color):
-        if color:
-            self.set_attribute(borderpixel=self.conn.color_pixel(color))
+    def paint_borders(self, colors, borderwidth, width, height):
```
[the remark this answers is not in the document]
m-col: I agree, and I have a branch where I am merging these two classes (more specifically, forming a template Window for a base class in the backend base, subclassed in the x11 backend to actually implement the X specific parts). Cleaning up things like this will definitely play a part :)

libqtile/backend/x11/xcbq.py
m-col: It is strange, I know, and is a weird quirk of painting onto X windows whose justification seems to have been lost to the ages. Basically, the pixmap maps onto the window assuming that the window's content (minus borders) is in the top left, the horizontal orders are both right at the bottom, and the vertical bottoms are at the far right. The 4 squares in the corners fit in there somewhere too. So we need to rearrange the contents of the pixmap before painting so that it maps how we want.

test/layouts/test_common.py
```diff
@@ -457,3 +457,25 @@ def test_cycle_layouts(qtile):
...
+@pytest.mark.parametrize("qtile", [AllLayoutsMultipleBorders], indirect=True)
+def test_multiple_borders(qtile):
+    qtile.test_window("one")
+    qtile.test_window("two")
+
+    initial_layout_name = qtile.c.layout.info()['name']
+    while True:
+        qtile.c.next_layout()
+        if qtile.c.layout.info()['name'] == initial_layout_name:
+            break
```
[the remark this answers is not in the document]
m-col: True -- just added another test (to test/test_window.py), that I think you will definitely be happy with!

I have left this one here to go with the other one, as this one ensures that each layout is capable of accepting it without erroring.

m-col:
OK, rebased, added a much improved test that actually scrapes the colours from a drawn window, and addessed the other points :)

ramnes (approved):
LGTM

Barbarossa93:
I'm keenly interested in this PR, along with the rounded corners PR (#1517). A couple of questions to how to go about using this: When specifying a list of colors, does it paint the borders colors in the order of the list? Such that the first color is the innermost border and the last in the list is the outermost? Also, from what I can gather, you can only set a maximum border width and the colors are evenly distributed. I think what might be superior is to be able to specify a list of border widths for each color such that the first item in color list gets the width of the first item in the border width list. I apologize if I misunderstood/missed something; Again, very excited to see this come along!

m-col:
This PR is finished now, so you can give a test and see how it works -- confirmation will be valuable. The colours are drawn in order like you said, but IIRC it is outside first, going in.

About setting border widths per colour, I actually think that my very first draft of this PR did that. It has changed a lot and been hugely simplfied since then (and there have been many other commits to Qtile since then) but in theory it is a simple addition because the window can just use the sum of those values for its border width and the border drawing would only need a tiny change. We removed that to keep the first iteration of this simple, but I'd also be happy to see that addition once this is merged. For the time being I've been specifying a colour more than one time to get the same effect e.g. `[blue, black, black]`.

Barbarossa93:
So I finally had a chance to give it a shot: the multi borders work great! For fun, I tried merging this branch with your round_corners branch and perhaps unsurprisingly, it didn't work well. It's probably my fault for not merging the branches correctly, but if it isn't, I was curious to know if it would be possible to have both? These two branches would significantly increase qtile's ricing game over at [r/unixporn](https://www.reddit.com/r/unixporn/) :)

m-col:
Haha yeah these PRs probably have a bunch of conflicts. I figured I should get
one out of the way (this one) and then adapt the other on top of the changes
introduced first. I have been thinking about how they would interact...
individually they are relatively simple but what if one wanted rounded corners
with multiple borders. Would you round each stripe? It may be a bit complicated
to implement (not impossible!).

zordsdavini:
I'd like to have that PR on master as I see the purpose of it to add some colors to app for some reason, ex. I want to mark the terminal (with red border) of production shell (ssh to server) and not write some stupid things in it :) Was thinking about to do myself but I see we already have it if I understand correctly

m-col:
I also want this on master but other PRs delayed it and then it began to conflict. But don't worry I haven't forgot, so I will rebase these changes on top of master very soon (and I want to add a version for the Wayland backend too) so we can use it! I have missed having my double borders since this broke :/

m-col:
Rebased and made it work in Wayland in case anyone wants to continue using this.

Ready for review if anyone's up for it :)

m-col:
@ramnes @tych0 you guys mind if I merge this?

ramnes:
I already approved it 😉

m-col:
Oh yes, so you did. Well in that case, I'll go ahead and merge it!

tych0:
Fine with me!

PR #1702, opened 2020-04-30
irc actions

zordsdavini:
add github actions to send notifications to IRC
- on push
- on PR
- on tag


tych0:
Can we talk about this? I know we had this before, but I didn't love it
then. Is there a strong demand for it now?

zordsdavini:
As there was just small comunication between me and ramnes, and looked more like test, I remove it

PR #1704, opened 2020-05-01
Adds the ability to bind a key to swap groups among screens.

kovasap:
For example, if you have dual monitors and you want the group on one monitor to
switch to the second monitor (and the group on the second monitor to switch to
the first), this command will let you do that.

### DIFF
--- a/qtile-venv-entry
+++ b/qtile-venv-entry
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 # Add the following...
 # [Desktop Entry]
 # Name=Qtile(venv)
@@ -9,7 +8,6 @@
 # Keywords=wm;tiling
 # ... to ...
 # /usr/share/xsessions/qtile-venv.desktop
-
 
 source ~/.bashrc
 source ~/.xinitrc
--- a/qtile-venv-entry
+++ b/qtile-venv-entry
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+# Add the following...
+# [Desktop Entry]
+# Name=Qtile(venv)
+# Comment=Qtile Session Within Venv
+# Exec=/home/kovas/qtile/qtile-venv-entry
+# Type=Application
+# Keywords=wm;tiling
+# ... to ...
+# /usr/share/xsessions/qtile-venv.desktop
+
+
 source ~/.bashrc
 source ~/.xinitrc
 source ~/.virtualenvs/qtile/bin/activate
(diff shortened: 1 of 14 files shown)


kovasap:
See https://groups.google.com/forum/#!msg/qtile-dev/MnyO4rMo_M0/-FN4g-gCBAAJ

ramnes:
Great, thanks! I would actually use this. Should we add an argument or a second function to swap in the opposite direction?

kovasap:
Added argument!  Note that I have not tested this on more than 2 monitors, so I'm not sure if the rotating behavior will work exactly as expected.

kaistian:
I have tested on 3 monitors and it rotates perfectly both ways.

tych0:
Can you add a test for this to make sure it doesn't break?

kovasap:
Is there a place where such a test should go?  I'm not very familiar with
the codebase yet :(

tych0:
Yeah, somewhere sensible in the test/ directory.

pawamoy:
`test/test_manager.py` seems to be the right place to add a test.

kovasap:
I'm not sure I quite understand how to access the screens on the qtile object in my test to verify everything is working.  Anyone have any pointers?

kovasap:
Yep, that's where my latest code change is (I copied another test).  I'm
still a bit stuck tho.  Maybe I need to read some docs on the qtile object?

kovasap:
Hey sorry for the slowness, I'm still not quite sure what's wrong with my particular test.  If there's anyone who could take a look I would really appreciate it : )

qtile-venv-entry
```diff
@@ -0,0 +1,17 @@
...
+# Type=Application
+# Keywords=wm;tiling
+# ... to ...
+# /usr/share/xsessions/qtile-venv.desktop
+
+
+source ~/.bashrc
+source ~/.xinitrc
+source ~/.virtualenvs/qtile/bin/activate
+python ~/qtile/bin/qtile $*
```
ramnes, line 17: Did you really mean to add this here? Doesn't seem related to that PR.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 cairocffi >= 0.9.0
 cffi >= 1.1.0
 xcffib >= 0.10.1
+iwlib
```
ramnes, line 4: Same here.

ramnes:
Can you rebase on master? The diff is hard to understand with your latest push.

kovasap:
Sorry I'm trying to fix another problem with my qtile that I ran into by
updating my fork.  I didn't mean to pollute this PR!  Still not a git
expert.  Let me see if I can fix...

kovasap:
Continued in https://github.com/qtile/qtile/pull/2381.

PR #1705, opened 2020-05-02
Simplify the IPC client and server

flacjacket:
Change the IPC server over to the simpler higher-level interface. This removes a good amount of complexity here and doesn't seem to have any impact in runtime.


tych0:
Ha, I was just looking at this. LGTM in any case :)

PR #1716, opened 2020-05-09
manager: fire setgroup hook when changing screens

### LINKED ISSUE #1712: Current desktop not properly set when switching to a group on another screen
If i just move the mouse between monitors, the group on the screen from the new monitor does get focus but something's not being set properly and `rofi -modi windowcd -show windowcd` still shows the windows from the previously active group.

However, if I manually click on the currently active group displayed in the GroupBox widget then everything works as expected afterwards and the windows from that group are shown by the same rofi command that I have bound to _alt-`_.

So it seems to me that this is an issue with how qtile handles changing the current desktop when switching focus to a group on another screen and not something wrong with rofi itself.

I'm still on version 0.14.2 because the latest one still doesn't play nice with jetbrains products I use for work everyday.

tych0:
When changing screens, the group has necessarily changed to something else,
since we can't display the same group in two places. So let's fire the
setgroup hook.

We need to do this, since update_net_desktops() is subscribed to setgroup,
and it's the thing that updates the X atom state to indicate to other
clients what group is currently focused. This means we'll get the correct
state on restart after re-focusing with a mouse, and also that clients that
query this state will get the right information.

Closes #1712

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -27,3 +27,4 @@ include bin/iqshell
 
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
+recursive-exclude stubs *
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -822,6 +822,7 @@ class Qtile(CommandObject):
         self.current_screen = self.screens[n]
         if old != self.current_screen:
             hook.fire("current_screen_change")
+            hook.fire("setgroup")
             old.group.layout_all()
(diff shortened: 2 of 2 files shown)


tych0:
I think this *should* fix it, although because I'm a terrible person I haven't actually tested it myself. But it is an obvious omission.

tych0:
I'm gonna go ahead and merge this since someone confirmed it fixed the bug, and this branch also has a build fix for one of the recent build failures.

Issue #1663, opened 2020-04-22
Scratchpads windows are lost when restarting


m-col:
# Issue description

Qtile appears not to keep track of scratchpad windows when restarting so calling the scratchpad windows after restarting creates duplicate instances of these programs. I don't think they are closed, as programs that want only one instance complain that they are already open.

If it is difficult to restart Qtile and keep information about running scratchpads and how they map onto the config (that may have changed), Qtile could potentially add some X property to the window and remove it again after restart. Scratchpads that are no longer in the config could be plunked on the first desktop for the user to deal with.

I'm happy to work on a solution after confirmation :) 

# Qtile version

9ed2449

# Configuration

E.g.:
```
groups.extend([
    ScratchPad("scratchpad", [
        DropDown("tmux", "xterm -e tmux", height=0.4, on_focus_lost_hide=False, opacity=1)
    ]),
])
```

m-col:
I searched to find this issue already, because it surprises me that it hasn't already been posted. Sorry if I missed it!

tych0:
On Wed, Apr 22, 2020 at 07:41:15AM -0700, m-col wrote:
> I searched to find this issue already, because it surprises me that it hasn't already been posted. Sorry if I missed it!

Yeah, scratchpad should probably set some atom and look for that atom
on restart.

m-col:
Alright cool - I'll look into this then.

zordsdavini:
@m-col can you do something like common state storage, because there are a lot of objects what could use that, ex. pomodori widget current state, I have bar show/hide state. Maybe layout window positions could be saved, too

tych0:
We still have QtileState right? Sorry, I forgot about it :-)

m-col:
Yeah that sounds useful

m-col:
#1717 closes this

PR #1717, opened 2020-05-09
Track DropDown windows across Qtile restarts

m-col:
re #1663

These changes are most of the way to sorting this out I think. State is stored and then restored to/from the ScratchPad object, with only a little added to QtileState.

Currently I'm facing a problem where, upon a restart while a dropdown window is present in the current group, the tiled group below things it owns the window and so reserves space for it, while the dropdown window (correctly) takes its floating position above. I'm not sure which group to set for the windows upon restoration - normally, do visible dropdown windows belong to the current floating group?

### DIFF
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -215,6 +215,7 @@ class ScratchPad(group._Group):
         self._dropdownconfig = {dd.name: dd for dd in dropdowns}
         self.dropdowns = {}
         self._spawned = {}
+        self._to_hide = []
 
     def _check_unsubscribe(self):
         if not self.dropdowns:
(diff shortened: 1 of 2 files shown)


m-col:
I've updated the commit so it is now working well with no issues. Existing dropdown windows remain on the scratchpad group if both the group's name and dropdown's name are found in the new config. Otherwise they are added to the current group.

I checked out the test but I'm not sure exactly what is happening, and running the same test on my computer doesn't give me the same failure :(

m-col:
I'm getting inconsistent test results, sometimes passing and sometimes failing a random test. The error is always the same as the one from #1716. Is this unrelated to this PR?

tych0:
Yeah, @flacjacket was in here recently, does that look related to you?

PR #1720, opened 2020-05-10
let PulseVolume widget controls accessible via lazy interface

m-col:


tych0:
Thanks! Failure should be fixed on master, it was unrelated.

Issue #899, opened 2016-03-29
Yahooweather widget stopped to work


tollerm:
Yahooweather widget does not produce output anymore.

tych0:
On Tue, Mar 29, 2016 at 02:01:33PM -0700, tollerm wrote:
> Yahooweather widget does not produce output anymore.

Yes, there was some discussion on IRC a few days ago that yahoo just
started requiring oauth2 for their API, so until someone has time to
patch it it'll stay broken :(

PR #1724, opened 2020-05-12
Fixed YahooWeather-widget.

tcld:
This is a rework of the YahooWeather-widget. Since I had never interacted with Yahoos API before, I used the Cinnamon weather-applet by [@mockturtl](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/weather%40mockturtl) as a reference.

The format options have mostly changed since the widget has last been worked on, but I didn't want to map old and new but instead just have the keys be auto-generated from the received json-data in the hopes of keeping the widget work as long as possible.

There is a caveat here: I am new to Qtiles code (I have only switched to Qtile four days ago) *and* PyCharm doesn't really work with the current release (#1548), so there will be code conventions I haven't fully adhered to. Also, I saw that the latest changes to the file included a reordering of the imports - it is very likely that I have ordered them wrong, but I didn't see any guidelines for this.

Any kind of feedback is welcome!

Fixes: #899

### DIFF
--- a/libqtile/widget/yahoo_weather.py
+++ b/libqtile/widget/yahoo_weather.py
@@ -160,14 +160,10 @@ class YahooWeather(GenPollUrl):
     def __init__(self, **config):
         GenPollUrl.__init__(self, **config)
         self.add_defaults(YahooWeather.defaults)
-        self._url = None
         self.headers.update(HEADER)
 
     @property
     def url(self):
-        if self._url:
-            return self._url
-
         if not self.woeid and not self.location and not self.coordinates:
             return None
 
@@ -194,8 +190,7 @@ class YahooWeather(GenPollUrl):
         }
         params.update(oauth)
 
-        self._url = QUERY_URL + urlencode(params) + '%26'
-        return self._url
+        return QUERY_URL + urlencode(params) + '%26'
 
     def flatten_json(self, obj):
         out = {}
--- a/libqtile/widget/yahoo_weather.py
+++ b/libqtile/widget/yahoo_weather.py
@@ -33,12 +33,13 @@ from urllib.parse import urlencode
 
 from libqtile.widget import base
 
-# See documentation here: https://developer.yahoo.com/weather/documentation.html
+# See documentation: https://developer.yahoo.com/weather/documentation.html
 QUERY_URL = 'https://weather-ydn-yql.media.yahoo.com/forecastrss'
 APP_ID = 'xSqyTW54'
-CONSUMER_KEY = 'dj0yJmk9R0RwZ3dveWEwTHdWJmQ9WVdrOWVGTnhlVlJYTlRRbWNHbzlNQS0tJnM9Y29uc3VtZXJzZWNyZXQmc3Y9MCZ4PTVi'
+CONSUMER_KEY = ('dj0yJmk9R0RwZ3dveWEwTHdWJmQ9WVdrOWVGTnhlVlJYTlRRb'
+                'WNHbzlNQS0tJnM9Y29uc3VtZXJzZWNyZXQmc3Y9MCZ4PTVi')
 CONSUMER_SECRET = '83ea8fbd202ea06cd57fc01268139601bf966b47'
-HEADER = { 'X-Yahoo-App-Id': APP_ID }
+HEADER = {'X-Yahoo-App-Id': APP_ID}
 
 
 class YahooWeather(base.ThreadedPollText):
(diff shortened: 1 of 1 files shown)


libqtile/widget/yahoo_weather.py
```diff
@@ -84,60 +156,78 @@ class YahooWeather(GenPollUrl):
         ('steady', 's', 'symbol for steady atmospheric pressure'),
     ]
 
-    json = False
-
     def __init__(self, **config):
-        GenPollUrl.__init__(self, **config)
+        base.ThreadedPollText.__init__(self, **config)
```
tych0, line 160: Is there a reason to make this switch? It looks like you're manually making the request using the requests module, which is a dependency we currently do not require.

tych0:
You should be able to do all the import sorting and other formatting using the 'make lint' target on master.

libqtile/widget/yahoo_weather.py
```diff
@@ -25,53 +26,124 @@
...
-from libqtile.widget.generic_poll_text import GenPollUrl
 
-QUERY_URL = 'http://query.yahooapis.com/v1/public/yql?'
-WEATHER_URL = 'http://weather.yahooapis.com/forecastrss?'
-WEATHER_NS = 'http://xml.weather.yahoo.com/ns/rss/1.0'
+# See documentation here: https://developer.yahoo.com/weather/documentation.html
+QUERY_URL = 'https://weather-ydn-yql.media.yahoo.com/forecastrss'
+APP_ID = 'xSqyTW54'
+CONSUMER_KEY = 'dj0yJmk9R0RwZ3dveWEwTHdWJmQ9WVdrOWVGTnhlVlJYTlRRbWNHbzlNQS0tJnM9Y29uc3VtZXJzZWNyZXQmc3Y9MCZ4PTVi'
+CONSUMER_SECRET = '83ea8fbd202ea06cd57fc01268139601bf966b47'
```
tych0, line 40: It would be nice if there was a better way to distribute this, but I'm not sure there is.

tych0:
FWIW, I think this has been broken so long nobody is really using it, so I think you were right to ignore backward compatibility.

libqtile/widget/yahoo_weather.py
tcld: That's a good point; I really only made the switch because at the time this was easier than setting up the url for GenPollUrl. I'll have another look.

libqtile/widget/yahoo_weather.py
tcld: I would have no idea how. Although the app_id was specifically set up for the "qtile weatherwidget", and at the very least this way everything is transparent and editable.

libqtile/widget/yahoo_weather.py
tcld: I am having difficulties setting up the url; the example-code on the Yahoo site is for Python 2 and doesn't work anymore - even if it did work, though, it would require another ten imports.
I'll give this some more thought tomorrow.

tcld:
Made the switch to GenPollUrl, although I wasn't able to do it using the request-header (the only way I found would have been a little dirty).

I removed a few dependencies, but had to add `time` and `uuid` for generating the OAuth-data; I think those could be fixed and things would still work, but I am not fully sure whether that would be a longterm solution.
The generated `url` is non-staic for the same reasons. Let me know how bad those extra dependencies are in your opinion.

Also ran `make lint`.

tych0:
On Wed, May 13, 2020 at 11:06:47AM -0700, tcld wrote:
> Made the switch to GenPollUrl, although I wasn't able to do it using the request-header (the only way I found would have been a little dirty).
> 
> I removed a few dependencies, but had to add `time` and `uuid` for generating the OAuth-data; I think those could be fixed and things would still work, but I am not fully sure whether that would be a longterm solution.
> The generated `url` is non-staic for the same reasons. Let me know how bad those extra dependencies are in your opinion.

Those are fine, they're both in the standard library. It's really just
making people run "pip install something-else" that I think it's good
to avoid.

> Also ran `make lint`.

Cool, thanks!

tcld:
Is there anything I can do to make Travis happy?

tych0:
On Wed, May 13, 2020 at 12:55:13PM -0700, tcld wrote:
> Is there anything I can do to make Travis happy?

I think this is unrelated to your patches; there's a bug in some of
the recent exit code that got merged that nobody has looked at yet :(

tcld:
I added a few more unicode-symbols to cover the basic weather-conditions (in case people prefer that over text, like me). Sadly there aren't enough unicode-symbols (yet) to cover everything.

The only points left that I think could be improved have to do with the secret now (it is out in the open and not encrypted when sent). I wouldn't know how to improve on that or whether it is even worth bothering about.

On a sidenote, Yahoo sent me a notice today that the requests coming from qtile are officially white-listed now. So is there anything else I should do? (The documentation is updated automatically, as far as I can see, right?)

tych0:
I squashed things down and made a bit nicer commit message, but otherwise LGTM, thanks!

PR #1726, opened 2020-05-13
Adds .vscode/ to the .gitignore file.

oboingo:
Adds .vscode/ to the .gitignore file.  This would ignore vscode specific config files (similar to .idea/ for intellij IDEs).

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,7 +1,7 @@
 name: "Close stale issues"
 on:
   schedule:
-  - cron: "0 * * * *"
+  - cron: "0 */4 * * *"
 
 jobs:
   stale:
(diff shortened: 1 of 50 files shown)


tych0:
Can you drop the other git merge commits?

tych0:
And FWIW, I would add this type of stuff to your global gitignore: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

But I'm happy to have it in the tree too.

oboingo:
Oh!  Well you learn something everyday!  I would say that we don't need this PR then!

ramnes:
Merging, it's still great for new contributors that do not have git configured, and also we already have some other editor directories ignored so why not more. :)

PR #1727, opened 2020-05-13
Adds group...toscreen() toggle documentation.

oboingo:
Adds a small bit of text to the lazy.group["group_name"].toscreen() toggle documentation, explaining the toggle.

### DIFF
--- a/docs/manual/config/lazy.rst
+++ b/docs/manual/config/lazy.rst
@@ -73,7 +73,10 @@ Group functions
     * - ``lazy.screen.toggle_group()``
       - Move to the last visited group
     * - ``lazy.group["group_name"].toscreen()``
-      - Move to the group called ``group_name``
+      - Move to the group called ``group_name``.
+        Takes an optional ``toggle`` parameter (defaults to True).
+        If this group is already on the screen, then the group is toggled 
+        with last used
     * - ``lazy.layout.increase_ratio()``
       - Increase the space for master window at the expense of slave windows
     * - ``lazy.layout.decrease_ratio()``
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -364,7 +364,8 @@ class _Group(CommandObject):
         screen :
             Screen offset. If not specified, we assume the current screen.
         toggle :
-            If this group is already on the screen, then toggle group.
+            If this group is already on the screen, then the group is toggled
+            with last used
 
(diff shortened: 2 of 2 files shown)


docs/manual/config/lazy.rst
```diff
@@ -73,7 +73,9 @@ Group functions
     * - ``lazy.screen.toggle_group()``
       - Move to the last visited group
     * - ``lazy.group["group_name"].toscreen()``
-      - Move to the group called ``group_name``
+      - Move to the group called ``group_name``.
+        Takes an optional ``toggle`` parameter (defaults to True).
+        If this group is already on the screen, then the group is toggled
```
zordsdavini, line 78: Maybe better:
If this group is already on the screen, then the group is toggled with last used

tcld:
I might be overlooking something, but does `toscreen()` actually exist or is it just `cmd_toscreen()` we have? This has confused me more than once already.

Either way, this is good to know. I kind of assumed that `cmd_toscreen()` was broken.^^

docs/manual/config/lazy.rst
oboingo: I just copied it straight from the source code comments.  But I am up for whatever.

oboingo:
> I might be overlooking something, but does `toscreen()` actually exist or is it just `cmd_toscreen()` we have? This has confused me more than once already.
> 
> Either way, this is good to know. I kind of assumed that `cmd_toscreen()` was broken.^^

I use toscreen in my config file for my keys like this:
`Key([mod], k[index], lazy.group[i.name].toscreen(g[index]), lazy.to_screen(g[index])),`

Using mod + number I can switch to a certain group on a certain screen (my groups are "fixed" to a certain screen).  So... yeah...?  It exists and seems to work?

tych0:
The client code allows any method that starts with cmd_ to be called, and strips the prefix.

Looks like there's a bunch of git goop (merge commits and other things), can you drop those? LGTM otherwise.

tych0:
Seems like the rebase still got kind of messed up; the author/commit message is from a different commit, and the actual commit just has a small whitespace fix?

ramnes:
Fixed the commits myself and merged. Thanks!

oboingo:
Sorry about the messed up commits/rebase.  Thanks for merging.

Issue #1713, opened 2020-05-08
Example Please : mouse_callbacks{}


20jam:
https://github.com/qtile/qtile/blob/1d7b4c5a35310c6fdaa5239f7edf48588769be53/libqtile/widget/base.py#L95-L98

Say I want to spawn a terminal window with a command or dmenu

Thanks

tcld:
Is this just about adding a few lines to the doc string?

I have the following in my config, would that do?

```
def open_calendar(qtile):  # spawn calendar widget
    qtile.cmd_spawn('gsimplecal next_month')

def close_calendar(qtile):  # kill calendar widget
    qtile.cmd_spawn('killall -q gsimplecal')

# add widget with callbacks somewhere

widget.Clock(format='%Y-%m-%d %a %H:%M', mouse_callbacks={'Button1': open_calendar, 'Button2': close_calendar})
```

m-col:
I'll update the docstring to show an example. @tcld is it OK if I use your spawn example?

tcld:
Sure, feel free, then I won't have to add it myself. :)

m-col:
We can close this (#1728)

20jam:
Thanks for the help, It is very appreciated. Sometime is difficult for us non python developer to look for answer in source code, examples are very useful.

m-col:
No worries at all :)

PR #1728, opened 2020-05-14
add config example on how to use widget mouse_callbacks

m-col:
re: #1713 

This just expands the Widget docstring slightly to show an example of how to use `mouse_callbacks` in one's config.py.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -105,7 +105,6 @@ class _Widget(CommandObject, configurable.Configurable):
     When the clock widget receives a click with button 1, the ``open_calendar`` function
     will be executed. Callbacks can be assigned to other buttons by adding more entries
     to the passed dictionary.
-
     """
     orientations = ORIENTATION_BOTH
     offsetx = None
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -93,9 +93,19 @@ class _Widget(CommandObject, configurable.Configurable):
     have been configured.
 
     Callback functions can be assigned to button presses by passing a dict to the
-    'callbacks' kwarg. For example: {'Button1': func} will execute func when the widget
-    receives a button 1 press. The Qtile instance of passed as the only argument to the
-    callback functions.
+    'callbacks' kwarg.
+
+    For example::
+
+        def open_calendar(qtile):
+            qtile.cmd_spawn('gsimplecal next_month')
+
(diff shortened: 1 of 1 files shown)


libqtile/widget/base.py
```diff
@@ -93,9 +93,19 @@ class _Widget(CommandObject, configurable.Configurable):
...
+
+        def open_calendar(qtile):
+            qtile.cmd_spawn('gsimplecal next_month')
+
+        clock = widget.Clock(mouse_callbacks={'Button1': open_calendar})
+
+    When the clock widget receives a click with button 1, the open_calendar function
+    will be executed. Callbacks can be assigned to other buttons by adding more entries
+    to the passed dictionary.
+
```
ramnes, line 108:
```suggestion
```

libqtile/widget/base.py
```diff
@@ -93,9 +93,19 @@ class _Widget(CommandObject, configurable.Configurable):
     have been configured.
 
     Callback functions can be assigned to button presses by passing a dict to the
-    'callbacks' kwarg. For example: {'Button1': func} will execute func when the widget
-    receives a button 1 press. The Qtile instance of passed as the only argument to the
-    callback functions.
+    'callbacks' kwarg.
+
+    For example::
```
ramnes, line 98: I'm not a rST expert but we'll miss Python syntax highlighting, nah?

```suggestion
    For example:

    .. code-block:: python
```

libqtile/widget/base.py
```diff
@@ -93,9 +93,19 @@ class _Widget(CommandObject, configurable.Configurable):
...
+
+        def open_calendar(qtile):
+            qtile.cmd_spawn('gsimplecal next_month')
+
+        clock = widget.Clock(mouse_callbacks={'Button1': open_calendar})
+
+    When the clock widget receives a click with button 1, the open_calendar function
```
ramnes, line 105:
```suggestion
    When the clock widget receives a click with button 1, the ``open_calendar`` function
```

libqtile/widget/base.py
m-col: I wasn't sure either so I coped it from the Columns layout because I remembered it had a config example too. Looks like you're right, and there are a few other instances of the first form in the dostrings.

m-col:
Alright I've updated with the review changes so should be good to go.

ramnes:
Thanks!

PR #1729, opened 2020-05-14
Revert "Simplify the IPC client and server (#1705)"

tych0:
This reverts commit 934301e6df361c9e529cb460df994fcc984141d8.

We have seen a lot of racy test failures recently like:
https://travis-ci.org/github/qtile/qtile/jobs/686770119

This commit seems the most smoke-y gun-y, so I reverted it and ran it on
travis 4 times and got no failures. So let's revert this for now until
someone has time to debug it so we can get our tests passing reliably
again.

/cc @flacjacket

### DIFF
--- a/libqtile/core/session_manager.py
+++ b/libqtile/core/session_manager.py
@@ -40,7 +40,7 @@ class SessionManager:
 
         if os.path.exists(fname):
             os.unlink(fname)
-        self.server = ipc.Server(fname, self.qtile.server.call)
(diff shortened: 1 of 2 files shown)


flacjacket:
Certainly unfortunate, thanks for keeping the tests green. I can see if we can't get the simpler ipc client/server implementation working without some of the extra async, which is likely where the flakiness is coming in.

PR #1732, opened 2020-05-16
Fix docs

### LINKED ISSUE #1731: Missing content on docs.qtile.org
The documentation for extensions and widgets seems to be missing from http://docs.qtile.org/en/latest/manual/ref/extensions.html since today.

# Qtile version

I am fairly sure that the documentation was still working yesterday, at the very least the day before.

Can andybody confirm this?

ramnes:
The first commit fixes the `:baseclass:` logic, which I think never actually worked.

The second one fixes #1731: since we're now importing widgets (and extensions) lazily through a function proxy, Sphinx can not inspect the classes behind the proxies. I'm not fond of looking into `sys.modules`, but I couldn't find a simpler solution. If anyone has a better idea, I'm all ears!

### DIFF
--- a/docs/sphinx_qtile.py
+++ b/docs/sphinx_qtile.py
@@ -182,8 +182,8 @@ class QtileModule(SimpleDirectiveMixin, Directive):
 
         for item in dir(module):
             obj = import_object(self.arguments[0], item)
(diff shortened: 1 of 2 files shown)


tych0:
LGTM, thanks for figuring this out.

Issue #1707, opened 2020-05-02
review default layout


zordsdavini:
This is part of Plans for default config improvements #1582
_Just wanted to separate for clear discussion._
I'd suggest switch to Max and Columns. Columns has same functionality as Stack but more possibilities.
But only difference is that stack layout with filled 1 stack, other will leave empty. The problem is for idea to start default config with open terminal. If we will have columns/max layout then all screen will be filled. With stack/max you can get filled half of window. Solution could be start new terminal as floated.

ramnes:
#1741 got merged :)

Issue #1595, opened 2020-04-09
qtile base16 support


aMOPel:
Hey, i dont know if you want to implement something rather specific like color themes for this 
self tinker wm, but i made an integration for base16 https://github.com/chriskempson/base16 color themes into qtile.

I thought maybe you can use it to make something more generic and better integrated out of it and possibly push that to the base16 repo.

my repo : https://github.com/aMOPel/base16-qtile-test

i really like qtile so far ;) thanks for that

zordsdavini:
I have "theme" idea in my Qtile future list :) I thought about this like separate project with many themes to select. And integration into config maybe could be with less user manual work (all these **params['widget']['groupbox'])
I was thinking is it possible to change colors and other stuff like to find object, ex. widget/groupbox and assign theme. Then in user config it could be like few lines at the end 
```
import ThemeProvider from qtile.theme
 ThemeProvider('base-16')
```
Bang and you have everething in place.

But your work is awesome and as you can see - gives some thoughts

aMOPel:
Base16 has like a 300 themes, wasnt sure if u knew that from what you wrote.

Anyway, they have a standard for themes which is about which color goes on which number from 1 to 16, originally for syntax highligthing.
Since it is not easy to have rather generic color themes make sense with the elements in qtile i thought you might have use at least for my linking from [color numbers like `base00` to qtile elements](https://github.com/aMOPel/base16-qtile-test/blob/master/colors.py), since i dont hardcode a specific theme into my config but only the base16 interface for colors.
And i believe if you want to use base16 you have to do this link anyway, with some thought spend on which color goes where.

Or maybe i didnt get quite where you were going with your idea :smile:

zordsdavini:
I didn't new about Base16, but now I understand themes list in some other apps, ex. pywal :) Thanks

Maybe you didn't understand me correctly, I'll try explain more.
My idea was how colors can be applied to Qtile. You took the way by giving dictionary of all objects and then each object (ex. widget, layout) will use this dict but assignment should be done manually. 
I thought that maybe there is the way to assign without such manual work. Some wrapper to set colors to objects. Even with some config like layout configs take from one them and colors from other. Just brainstorming.

BTW. if we will have clear interface to do such things we could look for pywal integration.

My config use colors from Pywal and font from Xresources, so actually I use themes, too. You can look at my config https://github.com/zordsdavini/dotfiles/blob/master/.config/qtile/config.py .

m-col:
[...]

PR #1741, opened 2020-05-21
#1707: default layouts to columns and max

zordsdavini:
This is suggestion how default config should be adopted to default columns and max layouts for #1707

### DIFF
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -35,25 +35,34 @@ mod = "mod4"
 terminal = guess_terminal()
 
 keys = [
-    # Switch between windows in current stack pane
-    Key([mod], "k", lazy.layout.down(),
-        desc="Move focus down in stack pane"),
-    Key([mod], "j", lazy.layout.up(),
-        desc="Move focus up in stack pane"),
-
-    # Move windows up or down in current stack
-    Key([mod, "control"], "k", lazy.layout.shuffle_down(),
-        desc="Move window down in current stack "),
-    Key([mod, "control"], "j", lazy.layout.shuffle_up(),
-        desc="Move window up in current stack "),
-
-    # Switch window focus to other pane(s) of stack
+    # Switch between windows
+    Key([mod], "h", lazy.layout.left(), desc="Move focus to left"),
+    Key([mod], "l", lazy.layout.right(), desc="Move focus to right"),
+    Key([mod], "j", lazy.layout.down(), desc="Move focus down"),
+    Key([mod], "k", lazy.layout.up(), desc="Move focus up"),
     Key([mod], "space", lazy.layout.next(),
-        desc="Switch window focus to other pane(s) of stack"),
+        desc="Move window focus to other window"),
 
-    # Swap panes of split stack
-    Key([mod, "shift"], "space", lazy.layout.rotate(),
-        desc="Swap panes of split stack"),
+    # Move windows between left/right columns or move up/down in current stack.
+    # Moving out of range in Columns layout will create new column.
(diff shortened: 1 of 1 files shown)


libqtile/resources/default_config.py
```diff
@@ -35,25 +35,33 @@
...
-
-    # Swap panes of split stack
-    Key([mod, "shift"], "space", lazy.layout.rotate(),
-        desc="Swap panes of split stack"),
+    # Switch between windows
+    Key([mod], "h", lazy.layout.left(), desc="Move focus to left"),
+    Key([mod], "l", lazy.layout.right(), desc="Move focus to right"),
+    Key([mod], "j", lazy.layout.down(), desc="Move focus down"),
+    Key([mod], "k", lazy.layout.up(), desc="Move focus up"),
+    Key([mod], "space", lazy.layout.next(), desc="Move window focus to other"),
```
ramnes, line 43: To other what? :)

libqtile/resources/default_config.py
```diff
@@ -91,11 +99,11 @@
     ])
 
 layouts = [
+    layout.Columns(border_focus_stack='#d75f5f'),
```
ramnes, line 102: Why overriding the default color?

libqtile/resources/default_config.py
```diff
@@ -35,25 +35,33 @@
...
-
-    # Switch window focus to other pane(s) of stack
-    Key([mod], "space", lazy.layout.next(),
-        desc="Switch window focus to other pane(s) of stack"),
-
-    # Swap panes of split stack
-    Key([mod, "shift"], "space", lazy.layout.rotate(),
-        desc="Swap panes of split stack"),
+    # Switch between windows
+    Key([mod], "h", lazy.layout.left(), desc="Move focus to left"),
```
ramnes, line 39: To **the** left, no?

libqtile/resources/default_config.py
zordsdavini: updated

libqtile/resources/default_config.py
zordsdavini: because we have same color on unsplit column and simple window borders. I found it useful to set different colors to have some kind of indicator to see is it splitted or unsplit column.

libqtile/resources/default_config.py
zordsdavini: updated

libqtile/resources/default_config.py
zordsdavini: there was thoughts update floating window border, too. Because colors what we have in default config looks to different. That dark blue could be light blue (cyan) as we have current group border or Qtile logo.
P.S. maybe we should have Qtile logo in bar as image?

zordsdavini:
Floating window color preview. But Qtile logo I think I will not add, but as example

libqtile/resources/default_config.py
ramnes: Maybe we should keep the default here and do another branch for saner default colors across all layouts?

zordsdavini: will wait a little bit for more comments on other stuff (CurrentLayoutImage widget instead text) and can revert. I agree it is better to select default pallete of colors and set in all project defaults. @m-col suggested to have some theme.

zordsdavini: and for colors we need https://github.com/qtile/qtile/issues/1595 to start for this `themer`. Then we could use that themer to set defaults in widgets, layouts etc.

zordsdavini:
removed all additional changes. We will discuss separately on widgets and colors

zordsdavini:
@ramnes thanks for merged this. Will look for next step

PR #1746, opened 2020-05-25
Focus testing

tych0:
Here's some tests for the recent focus patches, to make sure things stay the way they are.

It's not clear to me that this behavior is entirely correct, but as we get more testing, we can update these as well.


tych0:
Arg. Seems like my refactoring of the tests was racy. I wonder if we can just get rid of all this exec nonsense.

tych0:
I'm just gonna close this for now in favor of a larger test re-work.

PR #1747, opened 2020-05-25
Focus testing

tych0:
Ok, some more tests, hopefully this time that actually globally pass :)


tych0:
I'm also going to merge almost every patch in this branch, besides the one that enshrines the WM_TAKE_FOCUS/input hint conversation, since I believe that's wrong and I'm working on a patch for it.

PR #1760, opened 2020-05-31
Focus fixen

tych0:
Here's some commits that try and fix/test the current focus problems in master.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -523,62 +523,63 @@ class _Window(CommandObject):
     def can_steal_focus(self):
         return self.window.get_wm_type() != 'notification'
 
+    def _do_focus(self):
+        """
+        Focus the window if we can, and return whether or not it was successful.
+        """
+
+        # don't focus hidden windows, they should be mapped. this is generally
+        # a bug somewhere in the qtile code, but some of the tests do it, so we
+        # just have to let it slide for now.
+        if self.hidden:
+            return False
+
+        # if the window can be focused, just focus it.
+        if self.hints['input']:
+            self.window.set_input_focus()
+            return True
+
+        # does the window want us to ask it about focus?
+        if "WM_TAKE_FOCUS" in self.window.get_wm_protocols():
+            data = [
+                self.qtile.conn.atoms["WM_TAKE_FOCUS"],
+                # The timestamp here must be a valid timestamp, not CurrentTime.
+                #
+                # see https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7
+                # > Windows with the atom WM_TAKE_FOCUS in their WM_PROTOCOLS
+                # > property may receive a ClientMessage event from the
+                # > window manager (as described in section 4.2.8) with
+                # > WM_TAKE_FOCUS in its data[0] field and a valid timestamp
+                # > (i.e. not *CurrentTime* ) in its data[1] field.
+                self.qtile.core.get_valid_timestamp(),
+                0,
+                0,
+                0
+            ]
+
+            u = xcffib.xproto.ClientMessageData.synthetic(data, "I" * 5)
+            e = xcffib.xproto.ClientMessageEvent.synthetic(
+                format=32,
+                window=self.window.wid,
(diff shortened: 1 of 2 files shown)


m-col:
Can confirm - now I can focus and type into keepassxc, which was broken before.

Jamim:
*Note: My previous message is [here](https://github.com/qtile/qtile/issues/1548#issuecomment-637674737).*

> it would be good to move this discussion to the actual PR.

ok

> can you post an xprop for Alacritty?

Sure, here it is:
```
WM_HINTS(WM_HINTS):
_NET_WM_DESKTOP(CARDINAL) = 9
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING
_NET_WM_ICON(CARDINAL) =
		program specified size: 800 by 600
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
WM_CLIENT_MACHINE(STRING) = "irma-sun"
_NET_WM_PID(CARDINAL) = 5912
WM_CLASS(STRING) = "Alacritty", "Alacritty"
XdndAware(ATOM) = BITMAP
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x2, 0x0, 0x1, 0x0, 0x0
_NET_WM_NAME(UTF8_STRING) = "fish /home/mim"
WM_NAME(STRING) = "fish /home/mim"
```

> and how do you switch from one screen to another?

By pressing `mod1`+<number of a *group* which is on an *inactive screen* currently> (please see my [config file](https://github.com/Jamim/dotfiles/blob/6e976b69cc9047805cb524a90c896095a4c9e1e3/qtile/config.py#L126)) or by clicking on a window on that screen.

tych0:
> By pressing mod1+<number of a group which is on an inactive screen currently> (please see my config file) or by clicking on a window on that screen.

For posterity, this is `qtile.cmd_to_screen(group.screen.index)`. I think this is probably a different bug that exists without these focus issues, right?

zordsdavini:
@tych0 your diff is perfect in the case what it solves. Other issues can be from other scopes. Please, merge it into master and we will test in live :)

Jamim:
> I think this is probably a different bug that exists without these focus issues, right?

I'm not sure. It looks pretty similar to the bug I had with KeePassXC previously.
Also, I don't think it's specific for `qtile.cmd_to_screen(group.screen.index)` since it doesn't work when I'm clicking directly on the window on the inactive screen as well.

serzhshakur:
Same issue with alacritty!

Jamim:
> I'm not sure. It looks pretty similar to the bug I had with KeePassXC previously.

I just checked the master branch once again and now I can say that the alacritty issue is already there, so it's a different bug.

zordsdavini:
Just installed alacritty to check on latest dev version (`0.15.1.dev131+g599e0b6a`). It doesn't focus. Same way as viber, teams. If I columns layout to max - I can get focus. So, this is not related to `focus-fixen` branch.

tych0:
If you alacritty users revert https://github.com/qtile/qtile/commit/d58cb2dd67fbef9e6053ca658d3bf485df284a8e does that help?

Can you xtrace the xprop command? That may also be useful.

In any case, yes, sounds like a different bug, so we should move that discussion to a different thread.

But I don't think we should merge this quite yet, sounds like there's still some issues with focus on JetBrains.

tych0:
Well, or we can merge this and I can work on fixing the jetbrains stuff on top. Sounds like we probably do need some of the unfortunate java hacks we had before.

serzhshakur:
Let's concentrate on focus issue in Jetbrains products. We already confirmed this is fixed.

For this bug I can open a separate issue https://github.com/qtile/qtile/issues/1548#issuecomment-637278825

tych0:
Ok, a new bug would be great. I'll go ahead and merge this. Thanks everyone :)

PR #1763, opened 2020-06-04
Fix unsetting hints

tych0:
See patch notes for details.


zordsdavini:
As PR is approved and looks like it is working better, I'm gonna merge it. 
@tych0 thank you for awesome job :)

PR #1770, opened 2020-06-06
confreader: do basic type checking of top level items

tych0:
I still think that getting type checking for config object arguments (like
those annotated in 98927b7b9f1f532d35f4e6b9abf80802d4a4d24b) is a good path
forward, but because of the way we do configs (via __import__), there's no
good way to type check top level variables.

So let's add some code that does a little bit of that. This doesn't recurse
into anything (so we don't get deep checking of e.g. lists of groups or
screens), but we do get checking of the top level boolean/string flags at
least, and basic sanity "this is a list" checking.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -74,6 +74,10 @@ class Config:
         for key in self.settings_keys:
             try:
                 value = settings[key]
+                if not isinstance(value, type(default[key])):
+                    from libqtile.log_utils import logger
+                    logger.error("bad type for {}: got {}, expected {}".format(key,
+                                 type(value), type(default[key])))
             except KeyError:
                 value = getattr(self, key, default[key])
             setattr(self, key, value)


ramnes:
Mhh, I'm not sure. I think I'd rather add more type checks where the variable are used. For example, do we really want to never allow None? Or do we really want to enforce lists over tuples?

tych0:
I want to get less bug reports :-)

I'm happy for this code to be improved with deep inspection, or cleverer
type matching on itetables or whatever. But for now, this seems like an
improvement over nothing.

tych0:
Maybe we can do it a different way though, can we look into the type annotations? Then we could annotate the default config with e.g. `Opetional[str]` or whatever where appropriate?

ramnes:
I think it's a more direct approach but then it would clutter the default config file for people that are not used to Python. Maybe we could create a second file, empty_config.py or something, with the type annotations and only empty values, that we'd use for the whole validation and to get the available configs keys?

ramnes:
Or to define a dict of available keys and type hints directly in confreader?

tych0:
I don't know, type hints are increasingly becoming part of the language. People who are new to python probably want to learn about modern python, so it doesn't seem like that big of a deal to include them.

That said, I couldn't figure out how to read them anyway, so maybe it doesn't matter.

pawamoy:
If you're after validation of a class and its fields, Pydantic might be a good candidate: https://pydantic-docs.helpmanual.io/

ramnes:
Yeah I thought about Pydantic myself but then it would be overkill to add it as a dependency just for the config. Hence my last comment.

pawamoy:
True. Hmmm. I thought about [cluegen](https://github.com/dabeaz/cluegen), which is very lightweight and fast, but it does not enforce types, so we would need to extend it.

tych0:
I agree that I think adding a dependency for this is overkill.

It looks like there is: https://docs.python.org/3/library/typing.html#typing.get_type_hints

But that doesn't work for strings or bools to me, it looks like:

```
Python 3.8.2 (default, Apr 27 2020, 15:53:34) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> foo: str = "foo"
>>> import typing
>>> typing.get_type_hints(foo)
  File "/usr/lib/python3.8/typing.py", line 1252, in get_type_hints
    raise TypeError('{!r} is not a module, class, method, '
TypeError: 'foo' is not a module, class, method, or function.
>>> bar: bool = False
>>> typing.get_type_hints(bar)
  File "/usr/lib/python3.8/typing.py", line 1252, in get_type_hints
    raise TypeError('{!r} is not a module, class, method, '
TypeError: False is not a module, class, method, or function.
```
So I think something like the above is best for "basic" things like bools and strs (for which I don't think `None` is or should be a valid value).

What if I change this to do a type check using `get_type_hints()` where possible, and fall back to using `type(defaults[key])` where not?

pawamoy:
`typing.get_type_hints` is best used on classes and module. It returns what is stored in the modue/class `__annotations__` attribute. But I don't see how it would help type checking the config, because it would return the annotations that the *user* added to their config file. If they don't add any, `get_type_hints` will return nothing.

tych0:
No, I'm suggesting that we do `get_type_hints()` on the default config, and check the user's config against that somehow.

pawamoy:
Oooooh okay, that makes more sense.

ramnes:
Yeah, `get_type_hints` is great! While I get your points Tycho, I'm standing by the idea to not add the hints in the default config to keep it as simple as possible. And  since we already have a `Config.settings_keys` variable, we could simply replace it with a class like this

```python
class Config:
    class SettingKeys:
        cursor_warp: bool
        ...
    ...
```

and then use `get_type_hints(ConfigKeys)` in the validation.

tych0:
> While I get your points Tycho, I'm standing by the idea to not add the hints in the default config to keep it as simple as possible.

But isn't it then annoying to then have to go look somewhere else to figure out what types are allowed for a particular config option? The python zen says: "Explicit is better than implicit.".

I also don't agree that using a basic feature of the language (and indeed, a required feature of most other programming langauges) is complicated.

ramnes:
I don't think it's complicated, I think it clutters the config file and might lead to incomprehensions like "why is it here?" or "does it mean I can change the types?". Type hints are a great thing when you expose an API and the users uses it, but I don't think it makes sense to have the user have the type hints for your API in his config file.

tych0:
> ask the users to have the type hints for your API in his config file.

We're not. Nobody will need to add type annotations for this to work correctly.

ramnes:
> > ask the users to have the type hints for your API in his config file.
> 
> We're not. Nobody will need to add type annotations for this to work correctly.

You're not doing it explicitly but users might feel it that way since it will be there when they will open their config file the first time, so they could end up thinking it's necessary for the config to work correctly.

> But isn't it then annoying to then have to go look somewhere else to figure out what types are allowed for a particular config option? The python zen says: "Explicit is better than implicit.".

You don't have to look elsewhere, you just have to modify the existing configuration which is already using the right types. And now you'll even have a nice error notification telling you that a type is incorrect if you try to reload with a wrong configuration, eh!

> The python zen says: "Explicit is better than implicit.".

It also says that simple is better than complex, although I'm not sure the Zen of Python is worth quoting at all. :)

tych0:
> It also says that simple is better than complex

You yourself said

> I don't think it's complicated

So which is it? :)

Anyway, I'll close this. Feel free to send a patch to do it that way if you want.

ramnes:
Why closing? :(

You know very well that words can have different meanings in different situation. I was just implying that type hints are not complicated per se but would still make the config less simple (more complex) by cluttering it. I'm sorry if you felt I was arguing for the sake of it, that's simply not the case, I just want the best for both the users and developers.

tych0:
> Why closing? :(

Because we can't agree on an approach that I want to implement :)

> type hints are not complicated per se but would still make the config less simple (more complex) by cluttering it.

Yes, and I disagree. But that's ok!

pawamoy:
Maybe there's a way to use a stub file? To add annotations for the default config file in a separate file?

ramnes:
I don't think so, stubs are not meant to be used at runtime.

pawamoy:
We could also imagine writing the type annotations at the beginning of the config, in a single block, so we can easily strip it when copying the file in the user config.

PR #1772, opened 2020-06-08
Config type checking

ramnes:
Continuation of #1770 

Check the commits messages. I think the first commit should be merged in master anyway.

@tych0 @pawamoy

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -24,9 +24,12 @@
 # SOFTWARE.
 import os
 import sys
-from typing import Optional
+from typing import Callable, Optional
 
 from libqtile.backend import base
+from libqtile.layout.base import Layout
+
+NoneType = type(None)
 
 
 class ConfigError(Exception):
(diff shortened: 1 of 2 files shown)


tych0:
All that and we can't use type annotations anyway :). lgtm.

ramnes:
Yeah :(

And I need to fix the tests.

tych0:
So what if we instead made it optional? If one of the things was present, we'd use it to check the type annotations, if not, then we wouldn't do anything.

The reason I ask is because it's part of a larger idea I have, which I'll post about shortly.

ramnes:
Ah, this is why you put that into load() and not validate() in the first place? Sounds good to me, I think I can revert to that or add a few bits to implement that behavior in validate(). What a cliffhanger though!

tych0:
Read the above and see what you think; maybe we can do this with real typing anyway?

ramnes:
I don't really like what I wrote here, let's close in favor of #1773.

PR #1774, opened 2020-06-10
accept any number of spacer widgets

m-col:
This lets a bar accept any number of `widget.Spacer` widgets, allowing for any number of blocks of widgets in the bar. These blocks are distributed so that the first (left) and last (right) are up against the ends of the bar (as usual) and the rest of these blocks have their centrepoint at regular intervals across the bar. E.g. 2 spacers will put the centre of the second block right in the centre of the bar.


tych0:
I'd say test whatever you want to keep working :). That looks reasonable, but more tests won't hurt. I'll merge this for now.

PR #1776, opened 2020-06-13
handle mpd widget disconnections cleanly

m-col:
#1306 had a small problem where the widget failed a connection attempt and an exception froze the widget. This PR handles the connection attempts a bit more cleanly so that failed reconnections don't freeze up the widget and it can continue when it reconnects later.


tych0:
Both of these LGTM; and your writeups in the PR message are good too. Can you include those in the commit message as well in the future? (github will automatically make the PR comments your commit message if there's just a single patch)

m-col:
Absolutely! Thanks for the feedback.

Issue #1208, opened 2018-09-13
Key Sequences / Key Chords


jjzmajic:
I was just wondering if there was a way to get Qtile to respond to key sequences /chords such as

- Emacs style `C-c C-c`
- Vim style `d i w` 

Further, is it possible to use any of the mod keys alone? E.g. using the `Super` key as a leader key that makes Qtile listen to the sequence to follow until it recognizes one / can't match it to any?

If not, I expect I'll finally get down to the "Hacking on Qtile" section of the manual, and start doing so. If that's the case, I'd really appreciate any pointers you guys might have.

Also thinking of implementing an emacs-which-key (https://github.com/justbur/emacs-which-key) kind of popup for when you bug out and forget shortcuts. Any pointers on that are more than welcome as well as I'll probably start working on that regardless of the answer to the above. The same goes for any preferences anyone might have, e.g. curses over GUI (and the case for either).

Melkor333:
Look at my issue #1190, I started doing this. Since opening the issue, I have hacked a bit more to have a widged display when I press something, but since I am about to build my own nixos desktop condiguration, I couldn't document it yet, open a PR here or something like that. But I have created a dotfiles repo containing [the qtile config](https://gitlab.com/Melkor333/dotfiles/tree/master/dotfiles/.config/qtile).

One thing to remember: I haven't found a way to assign both vim style AND normal/emacs style bindings to the same key. (like using `super` in `super d w` AND `super+h`). If you've got any question to my code, you can gladly open an issue on my gitlab or ask here..

Jfyi the way keychains are stored is a bit ugly right now. I couldn't come up with a good tree-like container to store the variables yet, but I will look into it again when I have a better working desktop.. My goal is to be able to create keychains in the config.py with minimal effort & knowledge, but right now it still includes setting up a root key, a home key, defining the home, etc. Just look at the files KeyChain.py for the feature and keys.py for how to use it in the config.

jjzmajic:
Oh, awesome! Thanks, @Melkor333! will take a look once hell (exam) week is over! Just as a first though, have you considered using a `dict` to store they key sequences instead of the objects themselves (and obviously writing up something pretty to access and modify it)? But I guess that would involve reworking the `Key` class completely... I guess it will have to wait a few days, but definitely expect me to be in touch.

Melkor333:
[...]

PR #1777, opened 2020-06-14
Keybinding chords

Miziak:
PR implements key chords(discussed here https://github.com/qtile/qtile/issues/1208) as core qtile functionality. Chords are implemented in two ways:
1. "immediate chord" after enter chord its any subbinding or Escape cause leave chord
2. "mode chord" work more like VIM modes we stay in this mode till Escape key is hit

PR implements also  widget indicator for "mode chord".

In my day job I'm not code python so please review PR in details, I'm open for any changes.


libqtile/core/manager.py
```diff
@@ -376,6 +382,25 @@ def ungrab_key(self, key) -> None:
         if key is not None:
             self.core.ungrab_key(keysym, modmask)
 
+    def clear_chords(self) -> None:
+        self.core.ungrab_keys()
+        self.keys_map.clear()
+
+    def grab_chord(self, chord) -> None:
+        self.current_chord = chord.mode if chord.mode != "" else True
+        if self.current_chord and self.current_chord is not True:
```
tych0, line 391: Can we drop the `is not True` here?

libqtile/widget/chord.py
```diff
@@ -0,0 +1,63 @@
...
+
+from libqtile import bar, hook
+from libqtile.widget import base
+
+
+class Chord(base._TextBox):
+    """Display current key chord"""
+    orientations = base.ORIENTATION_HORIZONTAL
+    defaults = [
+        ("chords_colors", {},
```
tych0, line 32: Is there a reason to have this config value? People should be able to set foreground and background themselves right?

libqtile/core/manager.py
Miziak: Right , changed

libqtile/widget/chord.py
Miziak: Ya, it's inspired by Vim statusbars - in theme we can change colors of mode(in here chord) indicators eg. Insert is orange and normal is blue. This value let exactly do that.

tych0: But I don't think this code is necessary; can't people set `background=...` when configuring their widget?

Miziak: Yes, they can but case here is to have ability to assign different colors to different modes, on your suggestion(and current behavior) colors are static and applied to all modes. This add instant visual clue about mode without even need to read text

libqtile/widget/chord.py
```diff
@@ -0,0 +1,67 @@
...
+    def _setup_hooks(self):
+        def hook_enter_chord(chord_name):
+            if chord_name is True:
+                self.text = ""
+                return
+
+            self.text = self.name_transform(chord_name)
+            if chord_name in self.chords_colors:
+                colors = self.chords_colors.get(chord_name)
+                if type(colors) is tuple and len(colors) == 2:
```
tych0, line 56: Can we just access colors[0] and colors[1] here without the if? That way if people get it wrong, it will fail loudly instead of silently.

libqtile/widget/chord.py
tych0: Oh, I see! Yes, thank you.

libqtile/widget/chord.py
Miziak: Sure I can omit this condition

tych0:
I'll go ahead and merge this now, but can you send a documentation patch about this? People ask about this feature a lot, so it would be nice to have a writeup about how to use it.

tych0:
Ping, any chance you can send a docs patch? :)

PR #1781, opened 2020-06-16
simple index page refactor

fdev31:
Lightweight change on the index page.
Looks more readable to me while improving the access to the essential items.


tych0:
For everyone else who wasn't on IRC: https://pasteboard.co/JdxkDxY.png is the screenshot of the index page now. Seems like a massive improvement to me.

PR #1782, opened 2020-06-17
add notify_below option to battery widget

m-col:
The notify_below option passed to the Battery widget will make the
widget send a system notification during `poll()` if the battery
percentage is below the notify_below value.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -333,6 +333,7 @@ class Battery(base.ThreadedPollText):
         self.add_defaults(self.defaults)
 
         self._battery = self._load_battery(**config)
+        self._has_notified = False
 
     @staticmethod
     def _load_battery(**config):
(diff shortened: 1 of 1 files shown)


libqtile/widget/battery.py
```diff
@@ -353,6 +355,11 @@ def poll(self) -> str:
         except RuntimeError as e:
             return 'Error: {}'.format(e)
 
+        if self.notify_below:
+            percent = int(status.percent * 100)
+            if percent < self.notify_below:
+                send_notification("Warning", "Battery at {0}%".format(percent), urgent=True)
```
ramnes, line 361: So it's going to send a notification every time the widget is polled? Shouldn't we keep a state to know if it has been sent already?

m-col: That was the idea but your way would also work. My logic was that if you have low battery you probably don't want to forget! But yes, we could remember if we've sent it, then reset that state when we go back above the threshold.

tych0: Code looks fine to me, feel free to merge when you guys agree on this :)

m-col: It's updated so it will only notify the first time.

PR #1785, opened 2020-06-18
Revert "Revert "send MapNotify to auto positioned windows too""

### LINKED ISSUE #1784: Leaving group with fullscreen window and then returning to it is broken
# Description

When I create a fullscreen window in a group, and then change to another group
and back to the original group, the fullscreen window is no longer rendered. Functionality
for the window is still there, however. E.g. when running a zoom ;) conversation in
fullscreen mode, the audio part of the conversation continues whereas window containing
the video does not render. Clearly the window (and it's functionality) is still someplace,
just not rendered.

Other applications also exhibit this behavior when used in fullscreen mode. Another example
is chromium (enter fullscreen mode with F11).

# Commit hash

This behavior manifests itself with current git, master branch. I bisected the issue,
and the last good commit that does not exhibit this behavior is: 52d787944ebb669d5e90d7d372648e539034c271.

# Stack traces
No stack traces were thrown.

tych0:
This reverts commit df621b9a8ec19e7207b61789bb58274e8ba9e639.

It seems my analysis in the above commit is wrong: we *do* need this for
some applications. In particular, since we unmap things when changing to a
different group, we do need to send a remap hint when it is floating.

Now, this may cause a flicker the *first* time a user-positioned window
appears, because it'll map twice: once the first user time, and once when
we do this map. This is better than leaving it unmapped, and I think is
less confusing than adding a boolean that is a once-mapped flag. If it
really bothers someone, they can fix it (and indeed, some windows may even
be smart enough to ignore it).

Closes #1784

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -209,25 +209,24 @@ class Floating(Layout):
         return above
 
     def configure(self, client, screen):
-        # After this, the client will be mapped. Either this will do it, or the
-        # client has already done it.
-        client.hidden = False
-
         # 'sun-awt-X11-XWindowPeer' is a dropdown used in Java application,
         # don't reposition it anywhere, let Java app to control it
         cls = client.window.get_wm_class() or ''
         is_java_dropdown = 'sun-awt-X11-XWindowPeer' in cls
         if is_java_dropdown:
+            client.unhide()
             return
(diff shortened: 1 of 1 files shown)


rxcomm:
This works for me. Thanks!

PR #1788, opened 2020-06-19
Option to warp cursor to bottom right before drag

TheUnknownCylon:
Added option to warp cursor to bottom right of window before drag. This
allows one to fully shrink a floating window, even when the mouse was
positioned in the top left of the window before dragging.

Rationale:
I made the transition from Xmonad to qtile a short while ago. I also muddle around sometimes with floating windows (shame). The the default window dragging behavior of qtile seems a bit odd to me compared to XMonad. When resizing a floating window, XMonad warps the mouse cursor to the bottom right of the window that is being warped. In this way the entire window can be shrunk entirely (down to a 1x1 window).
In Qtile, the mouse position is not warped to the bottom right of the window. Therefore, if the mouse reaches the top left of the screen when dragging, the window can not be shrunk any more, since the cursor can not leave the top left of the screen. This is annoying to me, since I don't want to move the mouse manually to the bottom right to shrink windows to my liking... :(

It was not possible for me to configure the Drag-directive to my liking, so I added a few lines of code. Feel free to reject this PR if it is possible to configure Qtile to warp the mouse before dragging (I would be happy to get a configuration hint).

I can make a screencast if the above explanation is not clear enough.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ qtile X.X.X, released XXXX-XX-XX:
         - added `guess_terminal` in utils
         - added keybinding cheet sheet image generator
         - custom keyboardlayout display 
+        - added option to warp cursor to bottom right of window before drag
 
 qtile 0.15.1, released 2020-04-14
     * bugfixes
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -114,9 +114,10 @@ class Drag(Mouse):
     To position the mouse cursor at the bottom right of the window before
     dragging, pass `cursor_warp=True` as argument
     """
-    def __init__(self, *args, start=False, **kwargs):
+    def __init__(self, *args, start=False, cursor_warp=False, **kwargs):
         super().__init__(*args, **kwargs)
         self.start = start
+        self.cursor_warp = cursor_warp
 
     def __repr__(self):
         return "<Drag (%s, %s)>" % (self.modifiers, self.button)
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -110,6 +110,9 @@ class Drag(Mouse):
 
     It focuses clicked window by default.  If you want to prevent it pass,
     `focus=None` as an argument
+
+    To position the mouse cursor at the bottom right of the window before
+    dragging, pass `cursor_warp=True` as argument
     """
     def __init__(self, *args, start=False, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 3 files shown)


tych0:
Code looks good to me, can you add a test for this?

m-col:
This sounds cool and I'll definitely use it :)

There was discussion a while back about allowing resizing from all 4 corners of a window, which would be an awesome thing to have. I guess there wouldn't be any conflict with this behaviour if it was implemented?

TheUnknownCylon:
Sure, I'll add some tests later this week.

> This sounds cool and I'll definitely use it :)
> 
> There was discussion a while back about allowing resizing from all 4 corners of a window, which would be an awesome thing to have. I guess there wouldn't be any conflict with this behaviour if it was implemented?

Is someone actually going to implement this? If not, I would be happy to add support for this as well. What would be the expected behavior then? Does the user configure a fixed corner to resize, or should qtile automatically detect the corner closest to the current mouse position, and start the resize from this corner? We'll have to configuration options then:
* `cursor_warp` with possible values `True` or `False`, to move the cursor to the correct corner before the resize, and
* `corner`, with possible values `'auto'`, `'top-left'`, `'top-right'`, `'bottom-left'`, or `'bottom-right'`.

(Perhaps it's best to create a separate PR for the `corner` option, if you prefer small PRs (I do))

m-col:
I think resizing from the bottom right is the most familiar for most people so i dont think it would be worth implementing the warp for the other corners... at least until someone asks for it. I guess somebody would either use the warp, and always resize from the bottom right, or they'd resize using all corners if they could all be moved.

tych0:
Ping, any chance we can get some tests? :)

TheUnknownCylon:
Jeah, I'll submit some tests. I can not find any related tests to mouse handling / drags, so my initial estimate (adding some tests in the upcoming weekend) was a bit optimistic. I'll look into it, if I can't decide what the best approach is I'll report back :)

TheUnknownCylon:
I am a bit in the dark. What kind of (unit) tests do you expect? Should it simulate mouse behavior? My initial thought was to unit test the behavior by calling `process_button_click` on the `QTile` object and asserting some state (such as mouse position) , but the QTile object available in the tests in `test_manager.py` does not expose the QTile object methods that I find in `manager.py`. Any hints would be welcome :)

tych0:
Yeah, the object in the tests is unfortunately poorly named and is not the same. The tests don't really have access to the actual qtile object, since it is run in a different process.

Generally we interact with it by testing inputs and testing that we get the expected output. I'm imagining you could accomplish this with some invocations of `xdotool mousemove` and friends?

PR #1789, opened 2020-06-20
fix qtile randomly not able to start

fdev31:
corrupted pickle prevents startup
leading to a core dump

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -179,11 +179,15 @@ class Qtile(CommandObject):
         hook.fire("startup")
 
         if state:
-            st = pickle.load(io.BytesIO(state.encode()))
             try:
-                st.apply(self)
+                st = pickle.load(io.BytesIO(state.encode()))
             except:  # noqa: E722
                 logger.exception("failed restoring state")
+            else:
+                try:
+                    st.apply(self)
+                except:  # noqa: E722
+                    logger.exception("failed applying state")
 
         self.core.scan()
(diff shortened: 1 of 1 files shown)


tych0:
Can you post the exact stack trace? I'm curious what this error is, it really "shouldn't" happen...

fdev31:
I run my qtile in the background, mate-session is my DE, and I'm used to kill my X with Ctrl+Alt+Bkspace, maybe that provides more context to understand how that happens...
I guess some pickle saving function runs partially & get killed.
If you tell me which file I can corrupt (I don't know what this state file is) I'll reproduce the bug quickly.
Else I would have to run without my patch & wait for it to happen...

tych0:
There is no file that this particular state is saved to; it's passed on the command line when running with --with-state as you note as a literal argument.

Can you show the argument that's breaking, maybe? This seems like an underlying bug somehow, maybe with argparse or something.

fdev31:
I'm confused, since I wasn't using this argument as far as I remember... at some point I tried adding it since the error was about some state and this unpickle line and I got similar stack trace using:
`with-state=/tmp/foo`
But when this happened I had no arguments passed...

tych0:
Right, generally there's no need to manually use this argument. If you weren't restarting and you got this stack trace, then something went *really* wrong, and we should try and figure that out.

tych0:
I'm going to close this for now; we should figure out whatever the underlying pickling error was and fix that.

Issue #1751, opened 2020-05-25
Thoughts on configuration and remote scriptability


ramnes:
Difficulties to play with the different Qtile objects such as in #1748 make me think that we could greatly improve the user experience by having a global `qtile` object and play with it, à la `current_app` in Flask. Something like:

```python
from libqtile import qtile

qtile.keys = [...]
```

This would make the configuration file radically different and a lot of things would have to be changed in the codebase, but I think it would increase the hackability of Qtile by an order of magnitude. And we could drop that `main()` hack.

If we push this one step further, we could make this `qtile` object remotely available and this would make the remote scriptability much simpler, because you could then import that global object in any Python shell and play with it, and you wouldn't have to use the command clients anymore, which I feel are an unnecessary abstraction over the manager from the user perspective, and yet another thing to understand and master.

I guess we will come back to a chicken and eggs problem on the implementation, but before we get into that I'd love to hear everyone's thoughts on the observation and idea.

zordsdavini:
I'd like to see no differencies between internal and external scripting. Maybe `qshell` will become needless, because in `ipython` I could just import `qtile` and play around.
I support this for simplicity.

zordsdavini:
And it is still time to do that before 1.0 Qtile release ;)

Craksy:
This would be great. i'll admit, i'm a bit of a noob. I'm sort of getting the hang of how the qtile instance is being exposed by now, but it is all quite confusing if you're just starting out.
when i was first exploring and playing around, directly importing qtile was one of the first things i tried.
for what it's worth i think this and #1791 would make things much more intuitive and approachable, especially for beginners.

ramnes:
[rlpython](https://github.com/fscherf/rlpython) looks like a good pretender for a future qtile-shell that looks a bit more like a normal shell

Issue #1719, opened 2020-05-10
Make hooks arguments more sensible


tych0:
A lot of the hooks fire and don't have any context for what actually changed when they fire. We have various commits (ac2346a26b39ccff69af13b80def7e94857f9e8f for #1341 or example) that add relevant context, we should do that everywhere.

m-col:
Many of the hooks currently pass context-specific arguments, some of which couldn't otherwise be determined (e.g. `selection_change`, `screen_change`). I guess this means that all hooks would receive one argument (`qtile`) and some would have a second, context-specific argument. Is this the desired way to approach this? I don't think there would be a way to send all hooked funcs the same number of arguments, which would be super neat.

tych0:
Yeah, I think "do whatever makes sense" should be the guiding principle. It seems reasonable to *always* pass the qtile object just because people may want to do arbitrary stuff, but any additional arguments based on context could/should also be passed. I don't think there's a reason to try and unify the arguments passed, as they may differ based on context.

m-col:
[Looks like](https://github.com/qtile/qtile/blob/master/libqtile/hook.py#L45) the qtile object is available at `hook.qtile` as an importable. I can't find any reference to `hook.qtile` in libqtile... is it old, vestigial code? We could either use that copy, fire hooked functions with `i(qtile, *args, **kwargs)` and remove the qtile argument from any hooks that currently pass the qtile object, or alternatively we could remove that global (unless it's used somewhere else) and pass `qtile` manually to all `hook.fire` calls. I like the idea of the qtile being 'implied' and passed as part of the hook mechanism automatically, but like many I've had a dislike for "global" drilled into me. What do you think?

tych0:
Oh, interesting. Maybe we should just rely on that vs. arguments? People could also keep track of it via `main()` if they wanted. (Maybe the above is a reason to get rid of `main()`?)

m-col:
Depending on how/if #1751 progresses `main` might not be required and `qtile` might be accessible from somewhere that isn't `libqtile.hook.qtile`. `libqtile.hook` seems like an odd place to import qtile from. Although perhaps that is a different conversation... not having to pass `qtile` around when firing hooks would be nice so I'll make change and we can then reassess.

tych0:
On Mon, Jun 22, 2020 at 09:16:58AM -0700, m-col wrote:
[...]

PR #1791, opened 2020-06-22
Pass qtile manager to all hooks

m-col:


m-col:
Changes thus far:

- passing `qtile` every time a hook is fired, as the first argument, automatically
- updating `hook.fire` instances which manually passed `qtile`
- updating hooked functions' arguments

We could also:

- tidy any hooked functions within libqtile that use `qtile` with a nested reference, where we could use the passed `qtile` directly
- remove the deprecated `window_name_change` hook

tych0:
I thought per https://github.com/qtile/qtile/issues/1719#issuecomment-647624197 that we were going to move libqtile.hook.qtile to somewhere more sensible instead?

m-col:
My bad, I must have missed that last comment. If we moved that qtile global somewhere else, would we then want to access it within libqtile.hooks to pass it to hooked functions, or actually just remove the qtile instance from all hook arguments and let the hooked functions deal with getting the qtile instance themselves (if they need it)?

tych0:
On Mon, Jun 29, 2020 at 09:41:01AM -0700, m-col wrote:
> My bad, I must have missed that last comment. If we moved that qtile global somewhere else, would we then want to access it within libqtile.hooks to pass it to hooked functions, or actually just remove the qtile instance from all hook arguments and let the hooked functions deal with getting the qtile instance themselves (if they need it)?

It seems nicer to just have it be global? But globals are also
terrible. I could be convinced either way :)

m-col:
I think this usage of global is pretty restricted anyway, so it seems like an easy way to get an importable reference to `qtile`. I'd argue the neatest place would be at `libqtile.qtile` by moving the `hooks.init` function into libqtile's `__init__.py`. This might be a nice way to get the first suggestion in #1751 working. I think that #1751 might have to be completed before updating the hooks, so that the hooks can use that new API.

tych0:
> I think that #1751 might have to be completed before updating the hooks, so that the hooks can use that new API.

I don't think #1751 is coming any time soon; I don't even really understand how to implement it :). Let's not block any hooks fixups on that.

If you think a global is better, do you want to switch this PR to a global (and maybe add a doc somewhere? :D)

m-col:
Actually yeah - if/when #1751 is worked on, the API would presumably be
unchanged i.e. `import qtile from libqtile`.

So before I make some updates, would this be the plan for this PR:

  - move hook.init and hook.qtile so qtile is available at libqtile.qtile
  - update any hook.fire calls to remove qtile from the arguments
  - import qtile to any modules with hooked functions that currently get qtile
    from hook.fire arguments

Perhaps we should also add a deprecation warning to the use of `main` in
configs?

ramnes:
Yep, I'm in line with everything in these last two comments. :)

tych0:
> Perhaps we should also add a deprecation warning to the use of `main` in configs?

Yes, and a bit of documentation on how to access the qtile object on the hooks page would be helpful too. Thanks!

m-col:
Test failed with "libqtile/hook.py:38:1: I004 isort found an unexpected blank line in imports"

Is it meant to be 1 or 2 blank lines after imports? I thought it was 2.

ramnes:
It's 2 blank lines around functions and classes, and 1 otherwise.

https://www.python.org/dev/peps/pep-0008/#blank-lines

m-col:
Huh, TIL, I am ashamed.

m-col:
These are the changes:

- moved hook.init and hook.qtile so qtile is available at libqtile.qtile
- updated hook.fire calls to remove qtile from the arguments
- import qtile to any modules with hooked functions that currently get qtile from hook.fire arguments
- add a deprecation warning to the use of main in configs
- added documentation on how to access the qtile object on the hooks page
- removed `main` from the `configuration variables` docs

Using a main function will still work but you'll get a deprecation warning in the log. Anything done in `main` can just be done directly in the config with `libqtile.qtile`.

Config breakage: hooked functions will not get passed the `qtile` object if they previously did.

ramnes:
Can you add this to the CHANGELOG?

m-col:
Done. I also noticed `main` in the test configs and removed those. However, there was one thing in the tests that I was unsure how to update. [Here](https://github.com/qtile/qtile/blob/master/test/test_manager.py#L1134), [here](https://github.com/qtile/qtile/blob/master/test/test_manager.py#L1160) and [here](https://github.com/qtile/qtile/blob/master/test/layouts/test_common.py#86) there are main functions defined in these test configs, though I'm not familiar enough with pytest to know exactly how to have the equivalent functionality outside of a main function. Any thoughts?

tych0:
Can you remove the commits that cancel each other out (adding qtile to hooks, then removing it) and rebase? The other stuff lgtm.

libqtile/core/manager.py
```diff
@@ -108,6 +109,7 @@ def __init__(
 
         self.core.wmname = getattr(self.config, "wmname", "qtile")
         if config.main:
+            warnings.warn("Defining a main function is deprecated, use libqtile.qtile", DeprecationWarning)
```
ramnes, line 112: Should we mention `@hook.subscribe.startup_complete` also?

ramnes:
Or more simply, let's squash everything down to one commit. :)

libqtile/core/manager.py
ramnes: (here and in the changelog)

m-col: To remind people that `qtile` can also be modified after startup?

ramnes: If you try to play `libqtile.qtile` inside the config directly, IIRC most things won't be initiated yet, so you will very quickly end up with errors. But if you play with it inside `@hook.subscribe.startup_complete`, you have the guarantee that everything is instantiated.

m-col:
How is the updated changelog entry?

Tests passed and rebased.

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+qtile X.X.X, released XXXX-XX-XX:
+    !!! Config breakage !!!
+	- hooks that pass the qtile object as an argument no longer will. It
+	  can be accessed directly at libqtile.qtile.
+    !!! deprecation warning !!!
+	- defining a main function in your config is deprecated. Instead the
+	  qtile object can be accessed before setup by importing libqtile.qtile
+	  into your config, or after setup by using
+	  hook.subscribe.startup_complete.
```
ramnes, line 9:
```suggestion
	- defining a main function in your config is deprecated. You should use
	  @hook.subscribe.startup_complete instead. If you need access to the
	  qtile object, import it from libqtile directly.
```

ramnes: Also, it's a config breakage rather than a deprecation now that you removed it from configreader.py, isn't it?

tych0:
I still see a whole bunch of commits that cancel each other out. Can you drop those?

CHANGELOG
m-col: I jumped the gun a bit with that one, and misunderstood how the confreader works. You're right. The intended behaviour was:
```
if the config defines main:
    save it as qtile.main, and run it upon start with the deprecation warning
else:
    do nothing with main
````
We don't want `main` in the default config if it's being deprecated, but still want it to be allowed. Maybe we could just add a comment for now in the default config saying it will soon be deprecated, do not use?

**Comment by m-col** [comment]
OK hopefully that is it :)

File: CHANGELOG [hunk 3]
```diff
@@ -1,3 +1,8 @@
+qtile X.X.X, released XXXX-XX-XX:
+    !!! Config breakage !!!
+	- hooks that pass the qtile object as an argument no longer will. It
```
tych0, line 3: Can you add the list of the hooks here so people can grep?

m-col: Ah yep good idea

PR #1793, opened 2020-06-28
Move un/grab methods to core

flacjacket:
Setup the core with abstract (un)grab methods for the keys, buttons, and
pointer.  Move the functionality which performed this grabbing from the
xcbq window into the X core.  The use of a root window is somewhat
X-specific, so exposing this instead in the core with the calls made
using the X root window id is easier to extend to Wayland.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -1,6 +1,8 @@
 import typing
 from abc import ABCMeta, abstractmethod
 
+from libqtile import config
+
 
 class Core(metaclass=ABCMeta):
     @property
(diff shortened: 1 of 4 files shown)


tych0:
LGTM. Also, if you want to start upstreaming any of your wayland stuff even if it's not complete, that seems fine to me. Anything that people can start playing with :)

flacjacket:
Yep! I'm just starting to play around with trying to integrate something and this is the first steps in that. Over in https://github.com/flacjacket/pywlroots there is a working simple compositor now, so just now turning to getting something up and running here.

PR #1795, opened 2020-06-29
Fix floating no frames

tych0:


tych0:
Seems like this doesn't work for fixing the bug, but something *like* this should.

tych0:
Ok, the above fixes it for me. Can anyone confirm?

PR #1798, opened 2020-06-30
Allow to selectively synchronize fullscreen

fdev31:
I wondered if it wasn't simpler to re-use the auto_fullscreen variable and allow it to be polymorphic, but I kept it separated.

config.py to avoid browsers to get fullscreen automatically:

```python
no_auto_fullscreen_windows = (
     Match(wm_class=['firefox']),
     Match(wm_class=['google-chrome']),
)
```

### DIFF
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -171,7 +171,7 @@ floating_layout = layout.Floating(float_rules=[
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
-no_auto_fullscreen_windows = False # Set a list of windows ignored by auto fs
+no_auto_fullscreen_windows = False  # Set a list of windows ignored by auto fs
 focus_on_window_activation = "smart"
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -27,7 +27,7 @@
 from typing import List  # noqa: F401
 
 from libqtile import bar, layout, widget
-from libqtile.config import Click, Drag, Group, Key, Screen
+from libqtile.config import Click, Drag, Group, Key, Screen, Match
 from libqtile.lazy import lazy
 from libqtile.utils import guess_terminal
 
@@ -171,7 +171,7 @@ floating_layout = layout.Floating(float_rules=[
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
-auto_fullscreen_exceptions = []  # Set a list of Match to be ignored
+auto_fullscreen_exceptions = []  # type: List[Match]  # Set a list of Match to be ignored
 focus_on_window_activation = "smart"
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -171,7 +171,7 @@ floating_layout = layout.Floating(float_rules=[
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
-no_auto_fullscreen_windows = False  # Set a list of windows ignored by auto fs
+auto_fullscreen_exceptions = None  # Set a list of Match to be ignored
 focus_on_window_activation = "smart"
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -171,7 +171,7 @@ floating_layout = layout.Floating(float_rules=[
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
-auto_fullscreen_exceptions = None  # Set a list of Match to be ignored
+auto_fullscreen_exceptions = []  # Set a list of Match to be ignored
 focus_on_window_activation = "smart"
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -171,6 +171,7 @@ floating_layout = layout.Floating(float_rules=[
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
+no_auto_fullscreen_windows = False # Set a list of windows ignored by auto fs
 focus_on_window_activation = "smart"
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -27,7 +27,7 @@
 from typing import List  # noqa: F401
 
 from libqtile import bar, layout, widget
-from libqtile.config import Click, Drag, Group, Key, Screen, Match
+from libqtile.config import Click, Drag, Group, Key, Match, Screen
 from libqtile.lazy import lazy
 from libqtile.utils import guess_terminal
 
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -321,7 +321,12 @@ class _Window(CommandObject):
         triggered = ['urgent']
 
         if self.qtile.config.auto_fullscreen:
-            triggered.append('fullscreen')
+            if self.qtile.config.no_auto_fullscreen_windows:
+                if not any(match.compare(self) \
+                        for match in self.qtile.config.no_auto_fullscreen_windows):
+                    triggered.append('fullscreen')
+            else:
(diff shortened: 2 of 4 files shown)


fdev31:
To make it for a release if it has some interest, it could be a bit more flexible by making it a list of exception over the general rule, eg:
```python
auto_fullscreen = True
auto_fullscreen_exceptions = (Match(wm_class['firefox']),)
```
Would apply auto fullscreen to vlc and others but not firefox.

```python
auto_fullscreen = False
auto_fullscreen_exceptions = (Match(wm_class['firefox']),)
```
Would apply auto fullscreen to firefox but not vlc and others.

ramnes:
What is the use case for this feature? And yes, I like this second approach better.

fdev31:
My own use case (not having it in browser) is probably a matter of taste, I like to have a "fullscreen video in a tile" while I'm doing something else.
I can imagine a few scenarios where this is useful:

- a group used to show many "fullscreen images" at once : one image viewer is configured as "no auto fullscreen" so it can extend to the tile instead of the screen when run in fullscreen mode.
- watching several video streams at once with no widgets or so displayed
- any app which doesn't behave the way you like regarding fullscreen (eg: citrix)

libqtile/resources/default_config.py
```diff
@@ -171,6 +171,7 @@
     {'wmclass': 'ssh-askpass'},  # ssh-askpass
 ])
 auto_fullscreen = True
+auto_fullscreen_exceptions = None  # Set a list of Match to be ignored
```
ramnes, line 174: Let's put an empty list instead.

libqtile/window.py
```diff
@@ -320,7 +320,10 @@ def update_hints(self):
     def update_state(self):
         triggered = ['urgent']
 
-        if self.qtile.config.auto_fullscreen:
+        is_exception = any(match.compare(self)
+                           for match in self.qtile.config.auto_fullscreen_exceptions or [])
```
ramnes, line 324: The let's assume we always have a list here.

libqtile/resources/default_config.py
fdev31: ok

fdev31:
Thanks!
About the documentation, should I inline it after the "auto_fullscreen" part or maybe make a separate page for auto fullscreen & the exceptions then add it after "hooks" ?

ramnes:
Have a look here: https://github.com/qtile/qtile/blob/master/docs/manual/config/index.rst#configuration-variables

fdev31:
> Have a look here: https://github.com/qtile/qtile/blob/master/docs/manual/config/index.rst#configuration-variables

Yes, my comment was referring to that, rather in the toctree in a sub document (after hooks) or inline just after like the auto_fullscreen ?

ramnes:
Ah sorry, I didn't understand. Yeah, inline after auto_fullscreen is what makes the most sense to me.

tych0:
I would almost lean towards letting auto_fullscreen be polymorphic in this case, but then I wonder if this isn't better just done as a user hook for client_new somehow. Is there a way to selectively set fullscreen from the client hook?

fdev31:
Making it polymorphic limits the flexibility compared to the current implementation having "exceptions" over the base flag (you lose the capability to have it only for some of the clients). I liked the polymorphic version too since it's "good for me", but this version may fit more usages.
For the other question I don't think it's possible without changing the base code, have you something in mind?

tych0:
I guess this really feels like policy for me. For example, you want some windows not to be automatically fullscreened. The next person wants some groups to not have automatically fullscreen windows, so she implements: `no_auto_fullscreen_groups`, the person after that wants it only on one screen (`no_auto_fullsreen_screen`), etc.

What I'm wondering is, can we make it so that this can be done from a hook? Right now people control floating from client new with various incantations of:

```
@hook.subscribe.client_new
def my_floating_policy(client):
    if some_condition(client):
        client.floating = True
```

I'm wonder if we can't do the same with `client.fullscreen = True`, to avoid getting a bunch of different "config as policy" settings, and instead push the policy into users' hands.

fdev31:
I don't think it can really work, I added:
```diff
diff --git a/libqtile/hook.py b/libqtile/hook.py
index 416eb093..e9ed58e2 100644
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -149,6 +149,15 @@ class Subscribe:
         """
         return self._subscribe("float_change", func)
 
+    def window_state_change(self, func):
+        """Called when an in window state is made
+
+        **Arguments**
+
+        None
+        """
+        return self._subscribe("window_state_change", func)
+
     def group_window_add(self, func):
         """Called when a new window is added to a group
 
diff --git a/libqtile/window.py b/libqtile/window.py
index 473a02f7..5d2db14e 100644
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -320,10 +320,7 @@ class _Window(CommandObject):
     def update_state(self):
         triggered = ['urgent']
 
-        is_exception = any(match.compare(self)
-                           for match in self.qtile.config.auto_fullscreen_exceptions)
-
-        if self.qtile.config.auto_fullscreen != is_exception:
+        if self.qtile.config.auto_fullscreen:
             triggered.append('fullscreen')
 
         state = self.window.get_net_wm_state()
@@ -332,6 +329,8 @@ class _Window(CommandObject):
         for s in triggered:
             setattr(self, s, (s in state))
 
+        hook.fire('window_state_change', self)
+
     @property
     def urgent(self):
         return self.hints['urgent'] or self._demands_attention
 ```

With various implementations such as:
```python
auto_fullscreen = True
@hook.subscribe.window_state_change
def conditional_fullscreen(client):
    for name in client.window.get_wm_class():
        if name == 'firefox':
            client.fullscreen = False
```

It doesn't really work as expected and it's extremely easy to completely make qtile unusable with this hook.

One other option could be to make the `auto_fullscreen` a function, which is called in that case...

tych0:
What if you let it modify the wm state? Something like:

```
diff --git a/libqtile/window.py b/libqtile/window.py
index df1b6281..afb6ef52 100644
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -325,6 +325,8 @@ class _Window(CommandObject):
 
+        hook.fire('window_state_change', self, state)
+
         logger.debug('_NET_WM_STATE: %s', state)
         for s in triggered:
             setattr(self, s, (s in state))
```

The problem you're hitting is that `self.fullscreen` is actually a property, which triggers a bunch of code when it is set. So you have to either decide to set it or not before hand, setting it and then unsetting it will trigger weird and racy behavior. FWIW, I wish we could get rid of these kinds of properties and go back to assignment is just assignment everywhere, so if you want to do a larger refactoring and do that instead, that seems reasonable too.

fdev31:
I don't see how it will prevent the auto fullscreen from being triggered, we would need some "feedback" and conditional auto_fullscreen hook (returning true/false) to make it work.

tych0:
Right, the feedback would be changing the values in the wm state before it's applied.

fdev31:
So in my case, it would work with "my own" auto_fullscreen implementation in the custom hook and the standard auto_fullscreen set to False. Isn't it? It's slightly more code to maintain in the config.py but it's still fine for me this way.

tych0:
It depends on exactly how it's implemented, I could see it working regardless of auto_fullscreen setting.

fdev31:
I tried something from what I understood about this change.
After some discussion on IRC here is some kind of summary, but I'll first explain the behavior that was achievable with the first versions of the PR and that I can't reproduce now with some more flexible solution:
explain

I am not certain about the behavior I get or know the code enough to fully understand the impact, so I'll paste some code and explain the behavior.
- The hook in `config.py`
```python
@hook.subscribe.window_state_change
def conditional_fullscreen(client, state):
    for name in client.window.get_wm_class():
        if name == 'firefox':
            if 'fullscreen' in state:
                if not client.fullscreen:
                    state.remove('fullscreen')
```
- the update_state method in window.py:
```python
    def update_state(self):
        triggered = ['urgent']

            triggered.append('fullscreen')

        logger.debug('_NET_WM_STATE: %s', state)
        hook.fire('window_state_change', self, state)

        for s in triggered:
            setattr(self, s, (s in state))
```

With this code, the client loses the fullscreen internal state when the WM state change (client.fullscreen):
- if I'm on the bottom-right state in the grid above
- when I press a key to toggle the fullscreen state of the client I expect to be in the bottom-left state.
- instead I return to the top-left state... so the two states are still "linked" in this case while I like them independent

tych0:
> With this code, the client loses the fullscreen internal state when the WM state change (client.fullscreen):

Doesn't this mean your predicate for state change is wrong in your hook? I'm imagining something like:

```
@hook.subscribe.window_state_change
def conditional_fullscreen(client, state):
    for name in client.window.get_wm_class():
        if name == 'firefox':
            if 'fullscreen' in state:
                if not client.fullscreen and i_really_dont_want_fullscreen():
                    state.remove('fullscreen')
```

Only you can define that second part of the condition though, I'm not sure I totally understand the requirements.

PR #1799, opened 2020-06-30
Match: Allow any class of a client to match

fdev31:
I wondered why `Match` was considering only the 2nd WM_CLASS if there are many.
Why not all of them?
To me it looks more convenient, but maybe it was like this for a good reason.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -610,32 +610,26 @@ class Match:
 
     def compare(self, client):
         for _type, rule in self._rules:
-            if _type == "net_wm_pid":
-                def match_func(value):
-                    return rule == value
-            else:
-                match_func = getattr(rule, 'match', None) or \
-                    getattr(rule, 'count')
-
-            if _type == 'title':
+            # get value
+            if _type == "title":
                 value = client.name
-            elif _type == 'wm_class':
-                value = None
-                _value = client.window.get_wm_class()
-                if _value and len(_value) > 1:
-                    value = _value[1]
-            elif _type == 'wm_instance_class':
-                value = client.window.get_wm_class()
-                if value:
-                    value = value[0]
-            elif _type == 'wm_type':
-                value = client.window.get_wm_type()
-            elif _type == 'net_wm_pid':
-                value = client.window.get_net_wm_pid()
-            else:
+            elif _type == "wm_instance_class":
+                value = client.window.get_wm_class() and client.window.get_wm_class()[0]
+            elif _type == "role":
                 value = client.window.get_wm_window_role()
+            else:
+                value = getattr(client.window, 'get_' + _type)()
 
-            if value and match_func(value):
+            # compare
+            if _type == "net_wm_pid":
+                if rule == value:
+                    return True
+            match = getattr(rule, 'match', None) or getattr(rule, 'count')
+
+            if _type == "wm_class":
+                if any(match(v) for v in value):
+                    return True
+            elif match(value):
                 return True
         return False
 


tych0:
I don't understand your PR message here, it seems unrelated to the commit?

libqtile/config.py
```diff
@@ -620,10 +620,8 @@ def match_func(value):
             if _type == 'title':
                 value = client.name
             elif _type == 'wm_class':
-                value = None
-                _value = client.window.get_wm_class()
-                if _value and len(_value) > 1:
-                    value = _value[1]
+                value = client.window.get_wm_class()
+                match_func = utils.make_many(match_func, any)
```
tych0, line 624: I think this seems simpler as `lambda values: any(match_func(val) for val in values)`, but other than that this seems very reasonable to me.

fdev31: Do you mean you preferred the first version? I changed this code because I had issues with the linting rules.

tych0: Probably worth just fixing the linter issue, IMO. I like the lambda version better.

tych0: Or, well, it'll be named after you fix the issue :)

fdev31: So you prefer a nested function definition ?

tych0:
On Mon, Jul 06, 2020 at 02:36:29PM -0700, Fabien Devaux wrote:
> @fdev31 commented on this pull request.
> 
> 
> 
> > @@ -620,10 +620,8 @@ def match_func(value):
>              if _type == 'title':
>                  value = client.name
>              elif _type == 'wm_class':
> -                value = None
> -                _value = client.window.get_wm_class()
> -                if _value and len(_value) > 1:
> -                    value = _value[1]
> +                value = client.window.get_wm_class()
> +                match_func = utils.make_many(match_func, any)
> 
> So you prefer a nested function definition ?

Yeah, that would be great, thanks :)

libqtile/config.py
```diff
@@ -620,10 +620,11 @@ def match_func(value):
...
                 value = client.name
             elif _type == 'wm_class':
-                value = None
-                _value = client.window.get_wm_class()
-                if _value and len(_value) > 1:
-                    value = _value[1]
+                value = client.window.get_wm_class()
+                _match_func = match_func
+
+                def match_func(values):
```
tych0, line 626: I'm confused, doesn't seem like this would work? Don't you want to do something like `def _match_func(...)...; match_func = _match_func`?

fdev31: I tested it again and it seems to work, with it I can match "firefox" or "Navigator" and both work, what is the concern with it?

libqtile/config.py
```diff
@@ -630,8 +619,20 @@ def _comparator(_type, rule, value):
                 }
 
         for _type, rule in self._rules:
-            if _comparator(_type, rule, prop_getters[_type](client)):
```
fdev31: I'm not sure which version I prefer... (with or without the comparator function)

libqtile/config.py
```diff
@@ -609,34 +609,28 @@ def __init__(self, title=None, wm_class=None, role=None, wm_type=None,
...
+                'title': lambda c: c.name,
+                'wm_class': lambda c: c.window.get_wm_class(),
+                'wm_instance_class': lambda c: c.window.get_wm_class() and c.window.get_wm_class()[0],
+                'wm_type': lambda c: c.window.get_wm_type(),
+                'net_wm_pid': lambda c: c.window.get_net_wm_pid(),
+                'role': lambda c: c.window.get_wm_window_role(),
+                }
+
+        for _type, rule in self._rules:
+            if _comparator(_type, rule, prop_getters[_type](client)):
```
fdev31: Maybe I should get rid of the "client" parameter here to not add any limitation (just use client from the scope and have no parameter for the lambdas)

libqtile/config.py
```diff
@@ -610,32 +610,26 @@ def __init__(self, title=None, wm_class=None, role=None, wm_type=None,
...
-                value = client.window.get_wm_class()
-                if value:
-                    value = value[0]
-            elif _type == 'wm_type':
-                value = client.window.get_wm_type()
-            elif _type == 'net_wm_pid':
-                value = client.window.get_net_wm_pid()
-            else:
+            elif _type == "wm_instance_class":
+                value = client.window.get_wm_class() and client.window.get_wm_class()[0]
```
tych0, line 617: Doesn't this just take the first character of "wm_instance_class"?
```
>>> a = "foo"
>>> a and a[0]
'f'
```

fdev31: But there is no such thing as "wm_instance_class", it's the first class from the list, isn't it ?

tych0: Right, thanks. I didn't realize this had just been changed from above.

PR #1801, opened 2020-06-30
it happended I have windows with no group...

fdev31:
That was to avoid some trace:
`None has no attribute "screen"`

As far as I remember it happened during reload scenarios, but not only.

- are reload scenarios to be considered ?
  - then we can consider having a bit more tolerant code while reloading
- else I'll run without the patch & try to figure what is triggering it

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -865,7 +865,7 @@ class Window(_Window):
                 self.window.set_property('_NET_WM_STATE', list(new_state))
 
         if do_full:
-            screen = self.group.screen or \
+            screen = (self.group and self.group.screen) or \
                 self.qtile.find_closest_screen(self.x, self.y)
 
             self._enablefloating(


ramnes:
Yeah I'd prefer that we find the root cause on why a window doesn't have a group, that doesn't sound good.

tych0:
+1, if you have the stack traces that would be helpful!

fdev31:
I undid the patch and I can't see it in the logs, I'm looking from time to time, maybe something changed in my config or maybe that was really a specific case :(
Maybe we add some logging in the case the screen is not defined, hopping it happens again... or we wait... it wasn't making anything noticeably wrong as far as I remember...

ramnes:
I'll go ahead and close this, but please reopen an issue as many details as you can if you encounter this again.

PR #1805, opened 2020-07-08
Implement AND-match semantics for Match objects

cryzed:
Currently values in the float_rules for the Float layout are matched using "any" (OR) semantics, while it is more sensible to match them as "all" (AND). Creating a float rule that requires multiple properties to match at the same time is currently impossible, while recreating the old "any"-behavior is still possible, by simply creating a new rule for each property that used to be in the dict after this change.

Window.match now uses config.Match instead of only partially reimplementing its logic. This change also allows to match against three new properties: wm_type, wm_instance_class and net_wm_pid in the float rules, and anywhere else that Window.match might be used. The changes should hopefully all be backwards-compatible, except obviously the changed semantics of Window.match.

I tested this locally with the use-case I mentioned on IRC ({"wname": "win0", "wmclass": "jetbrains-pycharm-ce"}), i.e. matching the PyCharm splash dialog more specifically, which worked. Negating the match by using a bogus value for some property also seemed to work successfully. Otherwise Qtile seems to behave pretty normally while using it, so I am somewhat optimistic that I didn't break things too badly. Please look this over with a critical eye, I'm not very familiar with Qtile's code otherwise.


tych0:
Can you add some docs about this?

cryzed:
Yes, I will do so in the future, however I think we should talk about some details first. I feel like we should try to unify the attribute names between window.Match and Match.compare and keep them internally consistent. Why for example do we call "wname" "title" in Match, and why do we strip the "wm_" prefix from role. I'd feel better about this if we either gave all the attributes "friendly names" or sticked strictly to the technical X11 property names. What are your thoughts on this?

tych0:
Yeah, I think we should definitely synchronize the names where possible. I think we also probably don't need to pass lists to Match() any more, since it's conjunctive and therefore can't match two different roles/types/classes etc.

As for whether to use the raw x11 properties, it seems like having some abstraction over that might be nice, if for no other reason then peoples' rules might still work when wayland lands. But maybe wayland only has names, and not roles or classes? @flacjacket probably knows more here.

cryzed:
> I think we also probably don't need to pass lists to Match() any more, since it's conjunctive and therefore can't match two different roles/types/classes etc.

Well current semantics to config.Match follow the docstring, i.e. you can pass lists to it to get an or-match on a single property between all values, but one has to much. So...: `(A1 v A2 v A3) ^ (B1 v B2 v B3)` where `A` and `B` are property names.

tych0:
> Well current semantics to config.Match follow the docstring, i.e. you can pass lists to it to get an or-match on a single property between all values, but one has to match.

Right, I'm suggesting we get rid of this behavior. Every Match() represents exactly one set of properties.

> I'm personally for keeping technical names here, to prevent naming conflicts with Wayland-specific properties (which I hope Qtile will support in addition to X11).

Possibly, although this represents a larger migration for people. I don't really have a strong opinion either way. But given that I'm the guy who answers all these "why did my config break" questions, I'd prefer something more discoverable for people :)

cryzed:
> Possibly, although this represents a larger migration for people. I don't really have a strong opinion either way. But given that I'm the guy who answers all these "why did my config break" questions, I'd prefer something more discoverable for people :)

So keep the keys in float_rules for layout.Floating as they are, but change them to the more technical names in config.Match possibly? Also I saw a different pull request, where you said that you preferred closure functions to get the "matcher/predicate" function -- should I do the same here?

tych0:
> So keep the keys in float_rules for layout.Floating as they are, but change them to the more technical names in config.Match possibly?

Everything we can synchronize I think is good. So if there's a sane way to make all the names the same, I'd prefer that over anything else.

> Also I saw a different pull request, where you said that you preferred closure functions to get the "matcher/predicate" function -- should I do the same here?

I don't really care about closures vs. not, it's just that the code in that particular PR was much harder to follow the way it was abstracted. Simpler is better :)

libqtile/config.py
```diff
@@ -557,87 +557,95 @@ def __repr__(self):
...
+    matches all specified values, it is considered a match.
 
     Parameters
     ==========
     title:
-        things to match against the title (WM_NAME)
-    wm_class:
-        things to match against the second string in WM_CLASS atom
+        matches against the title (WM_NAME)
+    wmclass:
```
tych0, line 571: Is there a reason to drop the underscore here?

libqtile/layout/floating.py
```diff
@@ -28,20 +28,9 @@
...
-    'notification',
-    'toolbar',
-    'splash',
-    'dialog',
-])
-
-DEFAULT_FLOAT_RULES = [
-    {"role": "About"},
-    {"wmclass": "file_progress"},
-]
```
tych0, line 44: Nice that we can get rid of both of these, thank you!

tych0:
Overall this looks really great, thank you. One small nit and when the tests pass I'm +1. Can you also add a blurb to /CHANGELOG describing the API changes?

libqtile/config.py
cryzed: No this happened by accident, I'll revert that.

cryzed:
> Overall this looks really great, thank you. One small nit and when the tests pass I'm +1.

It looks like a majority of tests currently fail because of other code on the master branch (or so I hope), otherwise I'll have to spend some more time figuring out what actually broke. I'll also have to adjust the documentation for `layout.slice.Slice` still.

> Can you also add a blurb to /CHANGELOG describing the API changes?
Yes, I will do this soon.

libqtile/window.py
```diff
@@ -1081,39 +1082,19 @@ def toscreen(self, index=None):
                 raise CommandError('No such screen: %d' % index)
         self.togroup(screen.group.name)
 
-    def match(self, wname=None, wmclass=None, role=None):
+    def match(self, match):
```
tcld, line 1085: I already like this a lot, but in my opinion we could go further: Maybe it would be nice to move catching the exceptions into `Match.compare` itself. `Window.match` could then be deprecated, unless the ability to match things in both directions is useful for purposes I am overlooking right now.

tcld:
I really like this PR! @cryzed : Is there a chance that you will find the time to take care of the tests? I would be willing to help, as long as you at least find the time to take care of PRs in your fork etc. (I assume that's the most efficient way of collaborating)

cryzed:
@tcld Sorry, but I've since stopped using qtile and can't find the motivation to keep working on this.

tcld:
That's a shame to hear. I would at least like to somehow help finish this PR, but I don't just want to steal your code. I'm not sure how to do that, though. The rebase might already change the commit history enough, and if the commits need to be cleaned up in the end, the original author is probably lost.

cryzed:
That's really okay with me, you can just take my code and adapt it if you want, I don't really care about having my name appear anywhere.

tcld:
OK, I'll give it a shot. Thanks for the work - I'll try to keep the original author somewhere in there, I just don't know what git is going to do in the end.

cryzed:
As I said, no worries -- it's fine either way with me. Thanks for picking this up, I feel bad about just abandoning this.

pawamoy:
You can add co-authors to a commit by adding this to its message:

```
Co-authored-by:  Original Author Fullname <original.author@email.com>
```

tcld:
Thank you for the hint. I am currently rebasing and I noticed that the author stays @cryzed, while the committer is me. I never noticed those could be different people. I'll see how that looks in the actual history later.

tcld:
Since I think there is no way to change the branch for this PR, I created another PR: #1937 
This one can be closed.

PR #1807, opened 2020-07-09
core: fix multiscreen restart

### LINKED ISSUE #1790: Qtile screens not updating on restart after changing number of monitors
# Issue description

When I unplug a third monitor from my laptop (and have xrandr run automatically via udev) and then restart qtile, I still see in my status bar the third screen being present (3 of my groups are highlighted as if they are on a screen).  Also, my hotkeys for switching screens still behave as if this third screen exists.  

This happens even with a default qtile config.  I don't see any relevant lines in the qtile logs that would suggest a problem with the reset; it's as if this is intended behavior?

I was also using this tip: http://docs.qtile.org/en/latest/manual/ref/hooks.html#libqtile.hook.subscribe.screen_change to have qtile update to the correct number of screens on unplug, but was having a strange issue where every time a monitor was plugged in/unplugged my qtile would freeze for 1-3 minutes before resetting and working properly.  I stopped using this because of this slowness, but now I have this current issue.  Is there a difference between the restart command here and the lazy.restart() that is used by the default config's keybindings?

# Qtile version

commit b82c7c36b8e4ffe5ccb009c22f8c197ec5532dfe 
Author: Sean Vig <sean.v.775@gmail.com>
Date:   Mon Apr 27 22:31:53 2020 -0400

    Use additional async functionality
    
    Move more of the setup functionality into async. I think it'd be great if
    nothing needed to be passed a loop in the constructor and all of the async
    initialization could be called in async tasks, which would then be able to
    automatically get the running loop. Some additional refactoring could help
    with this. This may set us on a path that the Qtile object can be easily
    re-made from within the session manager if we can pull out enough of the
    direct loop handling from that.
    
    Closes #1689

tych0:
We have a bug where windows not on the screen with focus don't get drawn
correctly after multiscreen restart. This is because they don't realize
they're mapped any more. When qtile does a restart, it does:

1. apply all the state (i.e. which groups are in which screens)
2. focus the right screen
3. grab all the existing windows and throw them in a layout

The problem is that screens which aren't focused don't get another layout
event which means they don't realize they're being displayed.

Since state now relies on this ordering (viz. the scratchpad code, which
expects windows to be added after it has run), let's keep this ordering but
just add a:

4. for all screens, make sure they do one final layout so that groups know
   to map their windows if they're visible

Note that we explicitly don't have to do any focus dancing, since we don't
warp the cursor or otherwise change focus here, just map the right windows
at the right places.

Closes #1790

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -186,6 +186,9 @@ class Qtile(CommandObject):
                 logger.exception("failed restoring state")
 
         self.core.scan()
+        if state:
+            for screen in self.screens:
+                screen.group.layout_all()
         self.update_net_desktops()
         hook.subscribe.setgroup(self.update_net_desktops)
 


tych0:
Gonna merge this since there are no objections and it fixes the bug for me.

PR #1808, opened 2020-07-10
fix ci

tych0:
This pins isort to a < 5.0 version, since it's no longer API compatible and
flake8 is broken with >= 5.0.

See https://github.com/timothycrosley/isort/issues/1273


tych0:
I'm gonna go ahead and merge this since it does indeed fix CI.

PR #1812, opened 2020-07-13
added window_state_changed hook and fullscreen_no_border on xmonad layout

lKk2:


libqtile/layout/xmonad.py
```diff
@@ -299,6 +305,19 @@ def configure(self, client, screen):
         self._configure_specific(client, screen, px, cidx)
         client.unhide()
 
+    def _fullscreen_no_border(self, *args):
+        client = self.clients[self.focused]
+        if client.window.get_net_wm_state() == ['fullscreen']:
```
tych0, line 310: I'm not sure I understand why this is xmonad specific; shouldn't we remove the border for all fullscreened windows?

libqtile/window.py
```diff
@@ -325,6 +325,8 @@ def update_state(self):
 
         state = self.window.get_net_wm_state()
 
+        hook.fire("window_state_changed", self)
```
tych0, line 328: We're working on a very similar hook in parallel in #1798, but that has more arguments; can you add the state as an argument to this one if we end up adding it here?

lKk2: Sure, just added that.

libqtile/layout/xmonad.py
lKk2: I dont think it is xmonad specific, but by default xmonad does have border even on fullscreen windows, the wierd part is, right now its a bit bugged, when you go fullscreen it doesnt show up 'all' borders, let me bring a screenshot so you can see what im saying.


but yeah i hate those borders even on default xmonad

m-col: I get the fullscreen borders sometimes (not always) in the Columns layout, though this only started happening in the last couple months

tych0: I wonder if the "not always" bug is related to the user-specified position fixes.

But anyway, this all seems like a bug to me in window.place(); if a window is fullscreen, we should just set the border width to 0 there, right? No need for layout specific behavior.

lKk2:
@tych0 updated a lil bit, how do you think about this approach instead?

tych0:
Can we drop the hook entirely and just do the if fullscreen test in window.place()? That seems the least roundabout way to me.

PR #1814, opened 2020-07-13
Use python3 rather than python

mikelward:
On Ubuntu 20.04 /usr/bin/python is a symlink to python2, not python3.

PEP 394 suggests the "python3" binary should be available on any system with
Python 3. Verified for Ubuntu and Arch Linux.

Using "python" rather than "python3" should not usually be a problem, since venv
may make "python" Python 3 even if the system version is Python 2. But this way
seems more correct, and allows developing and running Qtile outside of a virtual
env if desired.

### DIFF
--- a/bin/qshell
+++ b/bin/qshell
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (c) 2008, Aldo Cortesi. All rights reserved.
 #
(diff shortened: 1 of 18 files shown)


mikelward:
Whoops! I resent this because I thought I'd accidentally deleted the PR somehow. But it was merged. :)

Closing this duplicate PR.

PR #1815, opened 2020-07-13
Use /usr/bin/env bash in scripts that rely on bash features

mikelward:
And fix some minor issues in `docs/build.sh`.

Without this, a default Ubuntu system prints

    scripts/xephyr: 8: [[: not found

when trying to run `scripts/xephyr`.

Normally I'd just use `#!/bin/bash` but I guess using env is more portable, and is the approach already used in `bin/dqtile-cmd`.

### DIFF
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 # Copyright (c) 2012 dmpayton
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


mikelward:
The only thing that doesn't work in `/bin/sh` in these scripts is use of the `[[ ... ]]` (double bracket) form of conditionals.

Alternatively, it would be trivial to update them to use `[` (or `test`), and maintain compatibility with `/bin/sh`.

tych0:
Up to you, I don't see it as a big deal to require bash, but I know some people are purists :)

tych0:
Anyway, this LGTM

mikelward:
If there's no preference, how about we use this as-is? That's less work. ;)

tych0:
Works for me, thanks for the patch!

PR #1816, opened 2020-07-14
Make MonadTall, MonadWide, and TreeTab work together with Slice

### LINKED ISSUE #1811: MonadTall and Slice don't work together
# Issue description

If I create a Slice layout with MonadTall as the delegate/fallback layout, the MonadTall layout uses the entire screen width, instead of just the width remaining after Slice.

It seems to be something to do with `self.group.screen` being the whole screen instead of just the screen passed to `clone`?

# Qtile version

3c46be326e7c1f14d24e836b40dcefac443d9478

# Stack traces

# Configuration

Like https://github.com/mikelward/conf/blob/main/config/qtile/config.py, with

```
layouts = [
    ...
    layout.Slice(width=400, name='sidebrowser', role='browser', fallback=layout.MonadTall()),
    ...
]
```

mikelward:
Using Slice means the delegate layout's available ScreenRect is smaller
than the Group's ScreenRect.

MonadTall and MonadWide store the size of the screen. Get the ScreenRect
from the call to configure, and store that instead.

TreeTab assumes its panel should be placed at (0, 0). Get the ScreenRect
from the call to show, and use it as the origin instead.

Slice was also not forwarding focus, show, and hide calls to it sub
layouts. TreeTab needs to receive show and hide events to show and hide
its panel. And without receiving the focus call, it would never place
or unhide its clients.

Fixes #1811

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -270,7 +270,7 @@ class Delegate(Layout):
     def delegate_layout(self, windows, mapping):
         """Delegates layouting actual windows
 
-        Parameterer
+        Parameters
         ===========
         windows:
             windows to layout
@@ -297,6 +297,22 @@ class Delegate(Layout):
                 idx += 1
                 focus = layouts[idx].focus_first()
         return focus
+
+    @abstractmethod
+    def show(self, screen):
+        """Tells the layout to show itself on the given ScreenRect"""
+        pass
+
+    def hide(self):
+        for lay in self._get_layouts():
+            lay.hide()
+
+    def focus(self, win):
+        self.layouts[win].focus(win)
+
+    def blur(self):
+        for lay in self._get_layouts():
+            lay.blur()
 
     def focus_first(self):
         layouts = self._get_layouts()
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -176,22 +176,24 @@ class MonadTall(_SimpleLayoutBase):
         if self.single_margin is None:
             self.single_margin = self.margin
         self.relative_sizes = []
+        self.screen_rect = None
 
     @property
     def focused(self):
         return self.clients.current_index
 
     def _get_relative_size_from_absolute(self, absolute_size):
-        return absolute_size / self.group.screen.dheight
+        return absolute_size / self.screen_rect.height
 
     def _get_absolute_size_from_relative(self, relative_size):
-        return int(relative_size * self.group.screen.dheight)
+        return int(relative_size * self.screen_rect.height)
 
     def clone(self, group):
         "Clone layout for other groups"
         c = _SimpleLayoutBase.clone(self, group)
         c.sizes = []
         c.relative_sizes = []
+        c.rect = group.screen.get_rect() if group.screen else None
         c.ratio = self.ratio
         c.align = self.align
         return c
@@ -210,7 +212,7 @@ class MonadTall(_SimpleLayoutBase):
         "Evenly distribute screen-space among secondary clients"
         n = len(self.clients) - 1  # exclude main client, 0
         # if secondary clients exist
-        if n > 0 and self.group.screen is not None:
+        if n > 0 and self.screen_rect is not None:
             self.relative_sizes = [1.0 / n] * n
         # reset main pane ratio
         if redraw:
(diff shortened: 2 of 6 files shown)


mikelward:
I think the fix looks something like this. Still needs tests.

tych0:
This looks like the right thing to me, thanks! Tests would be great :)

mikelward:
If my test is right, TreeTab may have a similar issue. Taking a look at that.

mikelward:
I don't think this fully fixes TreeTab, but it seems like an improvement?

mikelward:
`Slice` wasn't delegating the `focus` or `show` calls, which seems to have been the cause of issues with `TreeTab`. (`TreeTab` didn't think it was focused, so it didn't call `configure`.)

I'm not sure if there are other things it should delegate, but this fixes the thing I care about, and tests are passing locally.

Can look at doing more stuff as a follow-up.

mikelward:
Is CI broken?

```
PermissionError: [Errno 13] Permission denied: '/tmp/.X10-lock'
```

Doesn't sound related to my change?

tych0:
Yep, it is, I pushed a broken xcffib. I just pushed a fixed one, hopefully this works now? I'll restart it.

mikelward:
> Yep, it is, I pushed a broken xcffib. I just pushed a fixed one, hopefully this works now? I'll restart it.

Still seems to be failing?

libqtile/layout/base.py
```diff
@@ -298,6 +298,20 @@ def remove(self, win):
                 focus = layouts[idx].focus_first()
         return focus
 
+    def show(self, screen):
+        raise NotImplementedError
```
ramnes, line 302: Should it has `@abstractmethod`? If so, we should add a docstring to explain to layout developers what they are meant to implement here, and maybe pass rather than breaking, like others abstract methods above.

libqtile/layout/xmonad.py
```diff
@@ -267,6 +269,8 @@ def cmd_maximize(self):
 
     def configure(self, client, screen):
         "Position client based on order and sizes"
+        self.rect = screen
```
ramnes, line 272: Is `screen` an instance of `ScreenRect` here? Shouldn't we use `screen.get_rect()`?

mikelward: The existing naming is confusing.

`Layout.layout` takes a `screen`, which is really a ScreenRect.
https://github.com/qtile/qtile/blob/master/libqtile/layout/base.py#L54

It's called from `delegate_layout`, where it's clearer that it's passing a ScreenRect.
https://github.com/qtile/qtile/blob/master/libqtile/layout/base.py#L288

It would be nice to rename the `screen` parameter of `configure` and `show` to be called `rect` or `screen_rect`, but that would break the API.

libqtile/layout/base.py
mikelward: Yes. Thanks for catching that. How's this?

ramnes: Great, thanks!

libqtile/layout/xmonad.py
ramnes: Let's break the API if some other name makes much more sense and avoids confusion. Or maybe we should break the API on a type perspective and give a Screen instead of a ScreenRect?

mikelward: I would lean towards only passing a ScreenRect, since that's all the receiver needs.

Should I rename it in this PR, or as a follow-up?

mikelward: This touches a lot of things. I'd like to make it a separate PR if that's OK with you?

mikelward: Looks like there was one more place we need to pass the `rect` down: `TreeTab._create_drawer`. Doing that in this PR too...

mikelward: I made the rename a separate PR. I don't think I can base it on my other branch/PR, but here's a diff until we merge the first PR: https://github.com/mikelward/qtile/pull/1/commits/a93e070aa226452784fc468d753c62fde5cd40b2

mikelward: https://github.com/qtile/qtile/pull/1838 for renaming `screen` to `screen_rect`.

ramnes: Awesome, thanks!

PR #1819, opened 2020-07-17
copy default args for Window.static to Window.cmd_static

m-col:
`Window.static` has default values of `None` for x, y, width and height, so it makes sense that `Window.cmd_static` would also pass these as defaults and only require the index of the screen to make a window static.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -582,7 +582,7 @@ class Qtile(CommandObject):
                     return
 
                 if w.get_wm_type() == "dock" or c.strut:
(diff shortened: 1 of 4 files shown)


tych0:
Can we just s/static/cmd_static/ everywhere instead, and get rid of the proxy?

m-col:
Fixing the failed test was a real head-scratcher. The issue was that the cmd_
was returning a `Static` object which isn't compatible with the IPC. It doesn't
need to be returned anyway so removing that fixes it.

tych0:
Thanks!

Issue #1810, opened 2020-07-12
Turning any window static causes widgets to not render.


Craksy:
# Issue description
Making any window `static` either from an InteractiveCommandInterface or through a window object passed by the client_new hook causes several widgets, namely (A)Groupbox, Tasklist, and Prompt to not render(probably others too). CPUGraph, Clock and Volume seem to be rendering just fine.
Perhaps it's also worth noting that i can still click these widgets, and they react as they are supposed to. they just don't visually update. I've tried on Konsole, D-Feet popups, and a small app of my own.

# Qtile version
Qtile 0.15.1.dev160+gc427c9ef.d20200712 (fresh download)

# Stack traces
There are not really any errors or stack traces. The only thing i could find out of the ordinary in the logs is a couple of entries with an "Unhandled event: Expose" which were recorded whenever another window was opened in front of the static window, and then removed.
I attached the log output from a single session showcasing the error below.
[qtile-issue.log](https://github.com/qtile/qtile/files/4909474/qtile-issue.log)

and here's the xprop output for the static window that caused it (actually not from the same session, but produced in exactly the same fassion except in my main setup)
`_NET_WM_USER_TIME(CARDINAL) = 7766934
_NET_WM_DESKTOP(CARDINAL) = 4
WM_STATE(WM_STATE):
                window state: Normal
                icon window: 0x0
WM_WINDOW_ROLE(STRING) = "MainWindow#1"
_NET_WM_ICON_NAME(UTF8_STRING) = 
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) = "~ : ipython"
_NET_WM_NAME(UTF8_STRING) = "~ : ipython — Konsole"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x3e, 0x7e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x1400009
WM_HINTS(WM_HINTS):
                Client accepts input or input focus: True
                window id # of group leader: 0x1400009
WM_CLIENT_MACHINE(STRING) = "amnesia"
_NET_WM_PID(CARDINAL) = 73162
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 20971528
[...]

Issue #1654, opened 2020-04-20
qtile.update_gaps isn't cleanly reconfiguring bars


m-col:
# Issue description

If I create a window that says its window type is `_NET_WM_WINDOW_TYPE_DOCK`, qtile.update_gaps is called. My qtile bar then reacts strangely: clicks still work, a CPU graph continues to update, but the groupbox doesn't update when I change groups. The compton shadows around the bar get darker, so I suspect that what is happening is a new bar is being created (in the _configure calls), or at least the bar is creating a new window for itself, but the original isn't being cleaned away.

I'd like to take a look and fix the behaviour, but first I wanted to clarify what the full aims of qtile.update_gaps are. This is what I could interpret from the code:

1. if a dock is spawned with the window type above or it has strut hints and wants reserved edge space, qtile turns its window into a `Static` window
2. `Static` replaces qtile's struts with its own (maybe qtile should be summing them? What if there are two windows with struts?)
3. With the new struts, qtile should update current_screen.{top,bottom,left,right} to make space. This bit seems to be the broken bit. 

I think the solution might involve slightly changing how the bars are configured so that they don't need to create a whole new window upon reconfiguring, and just need to increase their outside margin.

# Qtile version

5bd8e57

m-col:
Looking into this, I can't see what benefit there is to Qtile handling struts at all. If a user wants to make space for a third party bar/dock then they are able to easily by adding a Gap, or if there is a Bar on that edge, then by setting the margin. With Bars having margins now, do we need [window.Static.update_strut](https://github.com/qtile/qtile/blob/master/libqtile/window.py#L736), [qtile.reset_gaps](https://github.com/qtile/qtile/blob/master/libqtile/core/manager.py#L494) and [qtile.update_gaps](https://github.com/qtile/qtile/blob/master/libqtile/core/manager.py#L498)?

tych0:
On Mon, Apr 27, 2020 at 01:40:43PM -0700, m-col wrote:
> Looking into this, I can't see what benefit there is to Qtile handling struts at all. If a user wants to make space for a third party bar/dock then they are able to easily by adding a Gap, or if there is a Bar on that edge, then by setting the margin.

Right, but they have to do that manually. IIUC the struts spec is so
that they don't. i.e. the bar can tell the WM itself "hey, WM_STRUTS
extension me off 30px at the top", so the user needn't intervene.

m-col:
This is being addressed by #1820

m-col:
[...]

PR #1820, opened 2020-07-17
Adjust Bar margin/Gap size for struts without duplicating windows (#1654, #1810)

m-col:
New static windows make bars reconfigure; currently reconfiguring bars
creates a new window each time, and makes widgets re-register with qtile
and erroneously get turned into Mirror widgets. Instead, this just
adjusts the bar's outside margin with the new strut value, which can be
negative in the case that static windows are becoming unmanaged. In the
absence of a bar on the corresponding screen edge, a Gap is either added
or adjusted in size.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -162,6 +162,7 @@ class Bar(Gap, configurable.Configurable):
         self.widgets = widgets
         self.saved_focus = None
         self.cursor_in = None
+        self.window = None
 
         self.queued_draws = 0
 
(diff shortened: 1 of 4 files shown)


m-col:
Xephyr seems to have gotten a segfault during `test/test_manager.py::test_rotate` and `test/test_manager.py::test_resize`. On my system I ran these tests a bunch of times and got the segfault inconsistently around 1/5 of tests for rotate, and not at all for resize.  I'm not sure what to make of this. Have there been any changes to these tests recently?

tych0:
I'm in the process of doing `for i in $(seq 10); do TOXENV=py38 tox; done` on master and I haven't hit it so far... I also haven't seen this crash before :(

m-col:
I just ran `for i in $(seq 10); do pytest test/test_manager.py::test_resize; done` and `for i in $(seq 10); do pytest test/test_manager.py::test_rotate; done` on both master and this PR's branch and they both passed every time. Lol I have no idea what that means. I'd love to suggest it's an anomaly but it happened on my machine at least twice (for test_rotate) with the exact same error and message as travis. Could you try it also on this branch?

tych0:
Yeah, I'm running it on this branch now... no failures so far :(

tych0:
See #1823, maybe that will help?

m-col:
I ran the tests again with tox 10 times and got no errors (Xephyr 1.20.8-2 on Arch). Are you able to re-trigger to travis run?

tych0:
Huh, weird, that looks like roughly the newest version. Anyway, yeah, I'll re-run them.

tych0:
Failed again... and looking at the tests it does have to do with bar resizing :(. Any chance you can install the debug symbols so we can get more info r.e.. the crash?

m-col:
The debug symbols for xephyr?

I could but I am unable to re-trigger the error on my computer :/

m-col:
I rebased and it ran the checks again, with interesting outcomes:

- TOXENV=py35 and py36: test_rotate and test_resize both fail
- TOXENV=py37: just test_resize fails
- TOXENV=py38: every passes

I don't know what to make of this, as I don't think anything should be different between them :(

tych0:
So it turns out that both test_rotate and test_resize_ are incorrect tests (see #1848), so maybe this won't be a problem after that's merged :)

m-col:
That would be fantastic news if that is the case

ramnes:
Can you try a rebase on master?

m-col:
Looks like those changes you made fixed it :)

ramnes:
Great!

PR #1831, opened 2020-07-21
Fix transparent borders

### LINKED ISSUE #692: different window border on tiled focused terminal windows with compositing
I'm running debian stretch amd_64.
if I enable compositing (I tried xcompmgr, compton and unagi) I see that the border of focused tiled terminal windows is of a sort of yellow, while should be red according to the layout settings.

I have an other machine which is running debian jessie x86, and I'm not able to reproduce this issue there...
I thought it could be a video driver issue, but it's not because I still have the problem while using the vesa driver.

the issue afflicts all terminal emulators I tried with transparency support (even if it's not enabled) [terminator, lxterm, roxterm, sakura and others...].
with urxvt is present only if I enable true transparency, while if I use fake transparency there's no issue. So it must be something related with transparent windows...

I have the issue also with xmonad wm, so I don't think this is a qtile specific problem since everything works on debian jessue x86, but if you please try to reproduce this issue on your desktop we could get an idea of where the problem is...

ramnes:
Dumb copy-pasta of https://github.com/szatanjl/dwm/commit/1529909466206016f2101457bbf37c67195714c8

Closes #692

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -497,7 +497,8 @@ class Qtile(CommandObject):
 
     @functools.lru_cache()
     def color_pixel(self, name):
-        return self.conn.screens[0].default_colormap.alloc_color(name).pixel
+        pixel = self.conn.screens[0].default_colormap.alloc_color(name).pixel
+        return pixel | 0xff << 24
 
     @property
     def current_layout(self):


ramnes:
@Ippolo @tjarvstrand @akemrir @nemya9066 @waltereikrem Works for you?

akemrir:
Works very well:

pic-window-200721-2005-47

nemya9066:
can confirm as well work as intended.

PR #1833, opened 2020-07-22
Export QTILE_XEPHYR so configs can tell when they're run under Xephyr

mikelward:
For example, I plan to do:

    if os.environ.get("QTILE_XEPHYR"):
        width = 480
        mod = "mod1"

So I can more easily test under Xephyr using only part of the screen,
and using a different mod key.

### DIFF
--- a/scripts/xephyr
+++ b/scripts/xephyr
@@ -13,7 +13,7 @@ Xephyr +extension RANDR -screen ${SCREEN_SIZE} ${XDISPLAY} -ac &
 XEPHYR_PID=$!
 (
   sleep 1
-  env DISPLAY=${XDISPLAY} ${PYTHON} "${HERE}"/../bin/qtile -l ${LOG_LEVEL} $@ &
+  env DISPLAY=${XDISPLAY} QTILE_XEPHYR=1 ${PYTHON} "${HERE}"/../bin/qtile -l ${LOG_LEVEL} $@ &
   QTILE_PID=$!
   env DISPLAY=${XDISPLAY} ${APP} &
   wait $QTILE_PID


ramnes:
Can't you already tell with XEPHYR_PID?

ramnes:
Can't you pass the environment variable by yourself when running Xephyr? I don't use Xephyr myself, but when I do run a debug session, I do pass an environment variable: `DEBUG=1 startx`. I do it for the exact same reason than you, i.e. enabling different things in my config.

mikelward:
> Can't you already tell with XEPHYR_PID?
> 
> Edit: ah sorry, misread, it's not exported.

We could export that instead if you prefer?

mikelward:
> Can't you pass the environment variable by yourself when running Xephyr? I don't use Xephyr myself, but when I do run a debug session, I do pass an environment variable: `DEBUG=1 startx`. I do it for the exact same reason than you, i.e. enabling different things in my config.

Yes, of course. It's just more work. It would be nice to do less typing, and have a common approach so it's easier to search for and share config snippets with each other.

ramnes:
Well, if we want to have a common approach, we should use something else than XEPHYR_PID because not everyone uses Xephyr. :)

Maybe we could rather add a `qtile --debug` flag that does things like setting the appropriate log level and storing a `debug=True` somewhere?

mikelward:
> Maybe we could rather add a `qtile --debug` flag that does things like setting the appropriate log level and storing a `debug=True` somewhere?

That would work for me too. No preference.

Ideally we'd ensure it worked predictably with the default `LOG_LEVEL` that's set in `scripts/xepyhr`.
https://github.com/qtile/qtile/blob/master/scripts/xephyr#L6

ramnes:
Yeah. @tych0 thoughts?

tych0:
I use,

```
# If we're running in debug mode, it's for development. Make sure the
# hotkeys don't clash, only start one window, etc.
if '-d' in sys.argv:
    hostname = 'xephyr'
    mod = "mod4"
```

Will that work?

mikelward:
If we did that, would it be OK for `scripts/xephyr` to pass `-d` by default?

tych0:
Yeah, looks like we don't have `-d` any more (@ramnes pointed out we only had it in v0.1 :D), but you could do something similar with -l DEBUG.

tych0:
Where are we at with this? Is it still important?

mikelward:
I would still like something like this.

I don't think we should overload log level to mean two different things. It
should be possible to run a regular Qtile session with debug logs without
it also changing layouts, modifier keys, etc.

Calling it DEBUG has similar issues, but I can work with that if there's
consensus around that?

tych0:
This seems fine, I just wanted to make sure you still needed it. Cheers!

Issue #1490, opened 2019-11-12
Exactly once ScreenChangeNotifyEvent


ramnes:
IRC log (modified for readability):

```
10h38 ramnes tych0: any idea why plugging/unplugging a screen generates three different xcffib.randr.ScreenChangeNotifyEvent objects every time?
10h39 ramnes 2019-11-12 10:17:43,209 WARNING libqtile config.py:set_screens():L113 {'response_type': 88, 'sequence': 1239, 'rotation': 1, 'timestamp': 681475813, 'config_timestamp': 681528745, 'root': 448, 
             'request_window': 448, 'sizeID': 0, 'subpixel_order': 0, 'width': 1920, 'height': 1080, 'mwidth': 506, 'mheight': 285, 'bufsize': 32}
10h39 ramnes 2019-11-12 10:17:43,631 WARNING libqtile config.py:set_screens():L113 {'response_type': 88, 'sequence': 1239, 'rotation': 1, 'timestamp': 681475813, 'config_timestamp': 681528772, 'root': 448, 
             'request_window': 448, 'sizeID': 0, 'subpixel_order': 0, 'width': 1920, 'height': 1080, 'mwidth': 506, 'mheight': 285, 'bufsize': 32}
10h39 ramnes 2019-11-12 10:17:43,656 WARNING libqtile config.py:set_screens():L113 {'response_type': 88, 'sequence': 1239, 'rotation': 1, 'timestamp': 681475813, 'config_timestamp': 681528875, 'root': 448, 
             'request_window': 448, 'sizeID': 0, 'subpixel_order': 0, 'width': 1920, 'height': 1080, 'mwidth': 506, 'mheight': 285, 'bufsize': 32}
15h13 @tych0 ramnes: yeah, i think X gonna X there
15h13 @tych0 a while ago i was going to write a better hook
15h14 @tych0 which would send along one notification when the monitors were actually ready
15h14 @tych0 but i never got around to it
16h28 ramnes tych0: how did you plan to do that?
16h29 ramnes get the event, see if a previous event is being handled and if not, spawn something that polls for the monitor status?
16h38 @tych0 ramnes: yeah, you'd have to keep some sort of local state
16h38 @tych0 or fire off a thread to do the polling, or whatever
17h23 ramnes let's make this an issue
```

and here we are

ramnes:
[...]

PR #1848, opened 2020-07-28
core: don't resize 0th screen on ConfigureNotify

tych0:
First, the comment """ Handle xrandr events """ is incorrect, since
ConfigureNotify doesn't come from the randr extension. (People can handle
randr events by subscribing to the ScreenChangeNotify hook.)

Second, it is incorrect to change the 0th screen's size on ConfigureNotify
for the root window, since the root window spans all current screens, not
just the 0th one. When we do this, it makes the one group span both
screens, which is a bug. This event happens on e.g. fullscreen exit
from some games, resume from suspend or hibernate, etc.

This does mean that we don't really handle the root window changing sizes
(e.g. when someone adds or deletes a monitor), but we already didn't do
that, so let's not try and pretend and introduce bugs.

### DIFF
--- a/libqtile/backend/x11/xcore.py
+++ b/libqtile/backend/x11/xcore.py
@@ -558,13 +558,6 @@ class XCore(base.Core):
 
         self.qtile.process_button_motion(event.event_x, event.event_y)
 
-    def handle_ConfigureNotify(self, event) -> None:  # noqa: N802
-        """Handle xrandr events"""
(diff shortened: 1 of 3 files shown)


ramnes:
Oh, may it fix #1490 also? I'll try to play with it tomorrow.

tych0:
I don't think so, but maybe :-)

PR #1850, opened 2020-07-29
Ensure that libqtile.qtile is never None from the user perspective

ramnes:
`libqtile.qtile` is `None` until `libqtile.init` is called, which is made in `Qtile.__init__`. Let's defer configuration loading after that step to make sure that an user importing `libqtile.qtile` will never get `None` instead of a manager instance.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -81,9 +81,7 @@ class Config:
     @classmethod
     def from_file(cls, path: str, kore: Optional[base.Core] = None):
         "Create a Config() object from the python file located at path."
-        cnf = cls(file_path=path, kore=kore)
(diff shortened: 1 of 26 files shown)


tych0:
Code lgtm, but not to travis :)

ramnes:
Yeah, right. It's because custom configs used in test do not inherit from `libqtile.confreader.Config` and thus do not have the `load` method. I just pushed a quick fix, but another route would be to have all these test configs inherit from `libqtile.confreader.Config`. What do you prefer?

tych0:
Yeah, subclassing Config seems reasonable if there's now required functionality in it.

ramnes:
Done!

PR #1852, opened 2020-07-31
Make cmd_spawn easier by using a shell

ramnes:
Rationale: https://groups.google.com/forum/#!topic/qtile-dev/5EZyuJZZtRs


tych0:
In the specific example you cited, is there a reason not to just use time.sleep() instead? I guess I don't think we should default this to true, if we merge this.

ramnes:
Probably? But the point to me is about making bash operators available.

Another related discussion in the mailing list, about aliases this time: https://groups.google.com/forum/#!topic/qtile-dev/85wOh0BRdc0

ramnes:
> I guess I don't think we should default this to true, if we merge this.

What's your reasoning behind this? I think the overhead is irrelevant these days, and it makes things easier for everyone as it gives the same context that the one you're used to work with (your shell).

tych0:
> it makes things easier for everyone

Until someone comes along and wants to pass a "&" to a command and has to figure out why it's not working :). Generally with APIs like this, I wouldn't assume that it's using a shell underneath. But if everyone disagrees with me, I'll be quiet.

tych0:
I do think it's worth exposing shell=True, just not defaulting it to true.

ramnes:
@tych0 See the last commit, related to our discussion on IRC. :)

tcld:
When starting out with Qtile, I had a number of issues related to missing aliases and environment variables. From the point of view of somebody coming in and just wanting to execute a command, expecting those things to work seems the natural thing to me.

Also, I played with `Popen` a bit and it seems that the only issues I was able to cause when using an ampersand were when `shell` was set to `False`. Decoupling via & or concatenating commands via && worked fine with `shell` set to `True`.

Unless I am overlooking something, I would very much agree that defaulting this to `True` is a great idea!

ramnes:
Just FYI, Tycho rightly told me that aliases still won't work, because bash only source the `.bashrc` file in interactive mode. But that definitely helps for the shell syntax. In the end we agreed that it was probably a good idea to make `shell=True` the default for `cmd_spawncmd`, because it's mostly used as a visual tool (the widget), and `shell=False` the default for `cmd_spawn`, which is a lower level function. I'll go ahead and merge this, feel free anyone to open another PR if you want to change the defaults.

tcld:
It all sounds reasonable to me, although a little documentation would be greatly appreciated. I wasn't able to find anything in the `subprocess`-documentation so far. For example, what environment variables would the shell inherit, and would `Popen` always spawn bash or the users default shell?

ramnes:
Actually I just updated the branch to not use `Popen`, as we're already forking and doing fancy process management stuff, so the change should be minimal on the behavior.

tcld:
I thought I understood where this was supposed to go, but I really don't. Sorry for interrupting!

PR #1854, opened 2020-08-02
Allow any class of a client to match

fdev31:
This is a slightly modified version, hopefully not breaking the logic anymore, sorry for the delay (I was off traveling).


fdev31:
I tested with the sample file I got in the meantime and also fixed another potential error

tych0:
Take two :). In the future, can you put a reference to the old reverts there so people can track all this if they want? But this looks fine for now, thanks!

PR #1860, opened 2020-08-10
Allow selecting layouts using negative indices

mikelward:
I'd like to be able to create shortcuts like

```
Key([mod], "grave", lazy.to_layout_index(-1), desc="Switch to layout -1"),
```

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -126,9 +126,9 @@ class _Group(CommandObject):
         logger.error("No such layout: {}".format(layout))
 
     def use_layout(self, index):
-        assert 0 <= index < len(self.layouts), "layout index out of bounds"
+        assert -len(self.layouts) <= index < len(self.layouts), "layout index out of bounds"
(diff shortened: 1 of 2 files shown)


mikelward:
Any thoughts?

Personally I think it would be nice to support any valid array index, i.e. `-len(self.layouts) <= index < len(self.layouts)`.

Let me know if you'd rather that?

tych0:
Yep, that is in keeping with python's list semantics so that sounds fine to me. If you push that it looks good.

mikelward:
> Yep, that is in keeping with python's list semantics so that sounds fine to me. If you push that it looks good.

Done.

tych0:
Can you add a small test for this too somewhere? The test failure you just hit should be fixed if you rebase.

mikelward:
> Can you add a small test for this too somewhere? The test failure you just hit should be fixed if you rebase.

Sure! Thanks!

Issue #1863, opened 2020-08-13
Static does not care about screen


superpupervlad:
# Issue description
So static function does not consider screen variable, all this functions will be working identically
```
window.static(screen=0, x=0, y=0, width=0, height=0)
window.static(screen=1, x=0, y=0, width=0, height=0)
window.cmd_static(screen=0, x=0, y=0, width=0, height=0)
window.cmd_static(screen=1, x=0, y=0, width=0, height=0)
```

# Qtile version
0.15.1-1

# Simple example
This code will place window in (10,20) on 0 screen, not on 1.
```
@hook.subscribe.client_new
def func(new_window):
	if new_window.match(wmclass="pavucontrol"):
		new_window.static(screen=1, x=10, y=20, width=100, height=200)
```
# What I wanted to do
If think that will be good if default screen will be current screen
```
@hook.subscribe.client_new
def func(new_window):
	if new_window.match(wmclass="pavucontrol"):
		new_window.static(screen=current_screen, x=10, y=20, width=100, height=200)
```

# What I had to do
```
@hook.subscribe.client_new
def func(new_window):
	if new_window.match(wmclass="pavucontrol"):
		qtile = new_window.qtile
		current_screen = qtile.current_screen.cmd_info()
		x = current_screen['x'] + 10
		y = current_screen['y'] + 20
		new_window.static(screen=0, x=x, y=y, width=100, height=200)

```
# Why this happens
Because in `libqtile.window` when `Static` class inits it use `self.place(x, y, width, height, 0, 0)` without participation of screen

tych0:
Seems like a reasonable fix; can you send a patch?

m-col:
This can be a pretty annoying big. https://github.com/qtile/qtile/pull/1865 fixes it but is just waiting for a test.

m-col:
Fixed with #1865

PR #1865, opened 2020-08-13
Static now consider screen

superpupervlad:
Now static consider screen and possible to call static function without screen, default screen is current screen.
#1863

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -44,6 +44,8 @@ Qtile X.X.X, released XXXX-XX-XX:
           AND-semantics, allowing for more complex and specific rules
         - Python 3.9 support
         - switch to Github Actions for CI
+        - Columns layout has new `margin_on_single` option to specify margin
+          size when there is only one window (default -1: use `margin` option).
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -118,7 +118,7 @@ class Columns(Layout):
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw a border when there is one only window."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
-        ("margin_on_single", False, "Add margin to the layout for only one window."),
+        ("margin_on_single", -1, "Margin when only one window. `-1` means use `margin`."),
         ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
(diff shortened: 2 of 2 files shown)


libqtile/window.py
```diff
@@ -955,7 +956,7 @@ def toggleminimize(self):
         warnings.warn("toggleminimize is deprecated, use toggle_minimize", DeprecationWarning)
         self.toggle_minimize()
 
-    def cmd_static(self, screen, x=None, y=None, width=None, height=None):
+    def cmd_static(self, screen=None, x=None, y=None, width=None, height=None):
```
tych0, line 959: Is this part of the patch necessary? It seems like it is not.

superpupervlad: This is not necessary, but without this it will be required to specify screen.

tych0: Right, but it wasn't before, so is there a reason to relax this requirement?

superpupervlad: Well, in my view if someone will want to use it(place static window not on a certain screen, but on current screen) he have to write extra lines of code, on other if someone want to use specific screen he already specify it.

tych0: Yep, fair point. So this looks fine, can you add a test for the whole thing?

superpupervlad: Um, I don't think I can handle it :disappointed_relieved:

libqtile/window.py
```diff
@@ -715,13 +715,14 @@ def __init__(self, win, qtile, screen,
         self.conf_y = y
         self.conf_width = width
         self.conf_height = height
-        self.x = x or 0
-        self.y = y or 0
+        x = x or 0
+        y = y or 0
+        self.x = x + screen.cmd_info()['x']
+        self.y = y + screen.cmd_info()['y']
```
ramnes, line 721: Can't we get the screen position in a simpler way?

libqtile/window.py
```diff
@@ -715,13 +715,14 @@ def __init__(self, win, qtile, screen,
         self.conf_y = y
         self.conf_width = width
         self.conf_height = height
-        self.x = x or 0
-        self.y = y or 0
+        x = x or 0
+        y = y or 0
+        self.x = x + screen.cmd_info()['x']
```
superpupervlad, line 720:
```suggestion
        self.x = x + screen.x
```

libqtile/window.py
superpupervlad: Of course
```suggestion
        self.y = y + screen.y
```
Also maybe to get rid of this checks for None
```
        self.x = x or 0
        self.y = y or 0
```
we should change default values of x and y in line 711 to
`                 x=0, y=0, width=None, height=None):`

m-col:
re testing: this patch should probably do it:

```
index 41d1c42a..cfaa49d1 100644
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -530,7 +530,9 @@ def test_static(qtile):
 
     self.test_xeyes()
     self.test_window("one")
-    self.c.window[self.c.window.info()["id"]].static(0, 0, 0, 100, 100)
+    self.c.window[self.c.window.info()["id"]].static(width=100, height=100)
+    self.test_window("two")
+    self.c.window[self.c.window.info()["id"]].static(screen=0, x=1, y=1)
 
 
 @manager_config
```

m-col:
re testing: this patch should probably do it:

```
index 41d1c42a..cfaa49d1 100644
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -530,7 +530,9 @@ def test_static(qtile):
 
     self.test_xeyes()
     self.test_window("one")
-    self.c.window[self.c.window.info()["id"]].static(0, 0, 0, 100, 100)
+    self.c.window[self.c.window.info()["id"]].static(width=100, height=100)
+    self.test_window("two")
+    self.c.window[self.c.window.info()["id"]].static(screen=0, x=1, y=1)
 
 
 @manager_config
```

m-col:
Hmm actually that isn't enough. We should stick a static window onto screen number 2 and test it's geometry.

m-col:
Hmm actually that isn't enough. We should stick a static window onto screen number 2 and test its geometry.

libqtile/window.py
m-col: Yeah that sounds good

m-col: After this change is made, I can merge it and try again to make a test. IMO a squashed bug without a test is better than a present bug.

ramnes:
It looks like I found a bug on GitHub as it closed the PR right after I rebased the branch. :(

Merged manually in f271f13f, thanks!

m-col:
Awesome thanks. I pinky swear to put a test on my To Dos 😉

Issue #753, opened 2015-09-08
Z axis


ramnes:
It would be nice to have some sort of Z axis information, which is completely missing today as far as I know.

That would fix at least #415 and #676, I guess.

ramnes:
Would also fix #1260.

ramnes:
One implementation could be to keep a z-ordered list of clients and then make `layout_all` iterate through that list to apply `bring_to_front` on each one. @flacjacket what do you think?

ramnes:
Yeah, looks like it's how it's meant to be done: https://stackoverflow.com/questions/40257320/how-to-define-window-stacking-order/40259106

tcld:
To get more into the code, I'll try my hands at #415 or #1260 and do what you suggested, @ramnes: Add in the facilities for keeping track of z-layers within a group and then draw them accordingly. It sounds suspiciously simple, so it is probably going to cause a million issues.

I am a little unsure how the z-layering would interact with the `focus_history` right now. If the lists were coupled (i.e. focusing a window brings it to the front, which might be expected behaviour), the entire idea of keeping track of the z-layers might be rendered moot, since the `focus_list` already does it.
A less automatic approach would require a few ways for the user to interact with the z-layering, and the result might be very much unintuitive to use.
It might even be best to just extend the `focus_history` list so it contains windows that were never focused. Then that list could be used for the z-layering and nobody would be confused.

Does anybody happen to have any input on this?

tcld:
After some experimentation, I have a little code to show by now - any feedback would be welcome (the general approach might be completely off):
https://github.com/tcld/qtile/tree/feature_zlayering

There is still a lot to learn (utilizing tests, create an environment to test for #415, minimize flickering due to redrawing in `layout_all`, and possibly more).
I have a few ideas for how to improve the flickering, but my understanding of Qtile and especially the X11-backend are too limited right now.

Overall I wanted to add screen-layering (not window-layering) for fake-screens, so I have tried to set up the new class in a way that might allow to reuse it for that purpose later. The placement of the class as well as the formatting of the code are not final in the linked branch.

ramnes:
Thanks for the work, glad to see it's moving forward!

[...]

Issue #415, opened 2014-04-27
Max layout: don't just display the active window


terminalmage:
In the Max layout, when using an application that is not floating, but has alpha transparency, the wallpaper/background is shown instead of the application under it in the stack. This appears to be so because only the active window is displayed in the Max layout, and thus there is no application under it to _be_ displayed.

I often use a transparent terminal, and like being able to see through to the application "underneath", for instance a web browser. It would be nice to be able to do this with qtile.

ramnes:
Fixed by 3ce1a785ee81b8f8b4486391f9a92e19a2de70d2

Issue #1260, opened 2019-01-09
Always on top floating windows


yousufinternet:
So I want to start by saying that I am sorry if this is not the right place for this post, I tried asking my question on the IRC channel (couldn't find any active sub reddit or forum), but I don't have constant access to internet/electricity (I live in Iraq) so I can't remain connected all the time and I might miss answers. 

My problem is that I want floating windows to remain always on top of tiled windows, the current behavior is that floating windows get covered by newly spawned windows. 
next I want to be able to apply this rule to some windows only (for example if I want to make mpv window playing youtube video always on top), and I want them to be sticky (always appear on the current active group). 

I couldn't figure out how to use the scripting commands on certain windows, groups, screens, root obj ... etc from the documentation. and the qtile-examples repository only contained very simple configurations with nothing to help me with my problem. 

So, to conclude I think the related hooks are: 
- client_new and client_focus, to run cmd_bring_to_front on a specific window whenever a new client is spawned or whenever the focus is changed 
- setgroup or changegroup (I don't know what the difference is exactly), to run cmd_togroup on the specified window to move it to the current group whenever the active group is changed 

But I don't know how to put everything together, how can I get the windows objects and run commands on them.. etc 

Thanks in advance for help, and sorry for my English, it is not my native language

flacjacket:
I can't think of a simple way that we could do this, but this is definitely something that we can look into improving with our current floating layout.

yousufinternet:
Thanks for the reply.. I thought that something could be setup with hooks, but I guess I just have to wait for future versions.

yousufinternet:
I know it has been a long time since this was posted, but I have solved this in somewhat weird way: 

in qtile\config.py:
`@libqtile.hook.subscribe.focus_change
def float_mpv():
    subprocess.Popen(os.path.expanduser('~/.config/qtile/floating_mpv.py'))`

floating_mpv.py:
`def float_mpv():
    c = Client()
    floating_mpvs = [(w['id'], w['name'], w['group']) for w in c.windows()
                     if 'mpv' in w['name'] and w['floating']]
    if len(floating_mpvs) == 0:
        return
    for floating_mpv in floating_mpvs:
        c.window[floating_mpv[0]].bring_to_front()
        c.window[floating_mpv[0]].togroup()

    # This code runs only when the current group is changed, so that
    # correct window is focused instead of mpv window
    prv_grp = floating_mpvs[0][2]
    cur_grp = c.group.info()['name']
[...]

Issue #1145, opened 2018-03-02
Tint2 is covered by other apps


nazar-pc:
I'm using Tint2 as my panel of choice and configured it to show on top of all windows except fullscreen.
However, this only works for windows that were opened before Tint2, everything opened later covers it.

Tint2 doesn't occupy separate area on the screen, but rather act like an overlay, so I'd really like to see it all the time.

`xprop` output:
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: -1175627520, 22095
		program specified minimum size: 564 by 24
		program specified maximum size: 564 by 24
WM_CLASS(STRING) = "tint2", "Tint2"
XdndAware(ATOM) = ATOM
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x2, 0x0, 0x0, 0x0, 0x0
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
_NET_WM_STATE(ATOM) = _NET_WM_STATE_SKIP_PAGER, _NET_WM_STATE_SKIP_TASKBAR, _NET_WM_STATE_STICKY, _NET_WM_STATE_ABOVE
_NET_WM_DESKTOP(CARDINAL) = 4294967295
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DOCK
_NET_WM_PID(CARDINAL) = 16986
_NET_WM_ICON_NAME(UTF8_STRING) = "tint2"
_NET_WM_NAME(UTF8_STRING) = "tint2"
WM_ICON_NAME(STRING) = "tint2"
WM_NAME(STRING) = "tint2"
```

nazar-pc:
Wrote a workaround, would be nice to have something like this in simpler form out of the box:
```python
from libqtile import hook
from xcffib.xproto import StackMode

tint2_instance = None

@hook.subscribe.client_managed
def register_and_bring_to_front_tint2_instance (window):
	global tint2_instance
	if window.name == 'tint2':
		tint2_instance = window
	elif tint2_instance is not None:
		tint2_instance.window.configure(stackmode=StackMode.Above)

@hook.subscribe.client_killed
def unregister_tint2_instance (window):
	global tint2_instance
	if window.name == 'tint2':
		tint2_instance = None
```

soulchainer:
[...]

PR #1870, opened 2020-08-17

tcld:
As discussed in #753 , this is an attempt at providing an ability to manage the z-layering of windows.

Addresses #415, #753, #1260 and #1893 in its current form.
Also touches on #676 and #1145, but I haven't taken an active look at the effects in those cases.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,8 @@ qtile X.X.X, released XXXX-XX-XX:
           AND-semantics, allowing for more complex and specific rules
         - Python 3.9 support
         - switch to Github Actions for CI
+        - added `only_focused` setting to Max-layout, allowing to draw multiple
+          clients on top of each other when set to False
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -41,6 +41,9 @@ qtile X.X.X, released XXXX-XX-XX:
         - switch to Github Actions for CI
         - added `only_focused` setting to Max-layout, allowing to draw multiple
           clients on top of each other when set to False
+        - added `cmd_keep_above` and `cmd_keep_below` to Windows so they can be
+          kept above/below other windows permanently; calling the corresponding
+          function twice removes the flag
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -150,7 +150,9 @@ net_wm_states = (
 WindowStates = {
     None: 'normal',
     '_NET_WM_STATE_FULLSCREEN': 'fullscreen',
-    '_NET_WM_STATE_DEMANDS_ATTENTION': 'urgent'
+    '_NET_WM_STATE_DEMANDS_ATTENTION': 'urgent',
+    '_NET_WM_STATE_BELOW': 'below',
+    '_NET_WM_STATE_ABOVE': 'above',
 }
 
 # Maps property names to types and formats.
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -752,17 +752,30 @@ class Qtile(CommandObject):
 
         # Additional option for config.py
         # Brings clicked window to front
+        client = self.windows_map.get(wnd)
         if self.config.bring_front_click:
-            self.conn.conn.core.ConfigureWindow(
-                wnd,
-                xcffib.xproto.ConfigWindow.StackMode,
-                [xcffib.xproto.StackMode.Above]
-            )
+            if not client or not client.group or not client.floating:
+                # TODO
+                # In the case of a non-floating client, we can't change any of the z-properties:
+                #   z.layout: Too volatile. Would be reset by the layout if it manages the z-layer (e.g. Max).
+                #   z.group: Too sticky. Would make it impossible to switch windows in Max-layout. The same goes for the
+                #            other z-properties.
+                # So we can't do this by incrementing a z-property. Maybe we should make the window floating instead?
+                # It needs to stick to the top of the groups window-list, but not so much that navigating
+                # through a Max-layout won't work anymore.
+                self.conn.conn.core.ConfigureWindow(
+                    wnd,
+                    xcffib.xproto.ConfigWindow.StackMode,
+                    [xcffib.xproto.StackMode.Above]
+                )
+            elif client.floating:
+                windows = list(client.group.windows)
+                windows.remove(client)
+                client.z.layout = max([c.z.layout + 1 for c in windows], default=0)
 
-        window = self.windows_map.get(wnd)
-        if window and not window.window.get_property('QTILE_INTERNAL'):
-            self.current_group.focus(self.windows_map.get(wnd), False)
-            self.windows_map.get(wnd).focus(False)
+        if client and not client.window.get_property('QTILE_INTERNAL'):
+            self.current_group.focus(client, False)
+            client.focus(False)
 
         self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayPointer, e.time)
         self.conn.conn.flush()
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -117,10 +117,10 @@ class _Group(CommandObject):
         self._previous_layering = windows
 
         index = windows.index(win)
-        if index > len(windows) // 2:
+        if True:  # index > len(windows) // 2:
             for w in windows[index:]:
                 w.window.configure(stackmode=StackMode.Above)
-        else:
+        else:  # right now this can't be used, since we wouldn't be able to draw over the bar etc.
             for w in reversed(windows[:index+1]):
                 w.window.configure(stackmode=StackMode.Below)
 
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -30,6 +30,7 @@ import contextlib
 
 import xcffib
 import xcffib.xproto
+from xcffib.xproto import StackMode
 
 from libqtile import hook, utils, window
 from libqtile.command_object import CommandObject
@@ -60,6 +61,7 @@ class _Group(CommandObject):
         self.focus_history = []
         self.screen = None
         self.current_layout = None
+        self._previous_layering = []
 
     def _configure(self, layouts, floating_layout, qtile):
         self.screen = None
@@ -90,16 +92,37 @@ class _Group(CommandObject):
             pass
         self.focus_history.append(win)
 
-    def _remove_from_focus_history(self, win):
-        try:
-            index = self.focus_history.index(win)
-        except ValueError:
-            # win has never received focus
-            return False
+    def _order_clients(self):
+        """
+        Orders our clients by their layouts z-value, then by the clients z-values.
+        If z-values are equal, the window ids are used for reproducibility.
+        The z-values themselves are managed by the layouts.
+        """
+        return sorted(
+            [w for w in self.windows if not w.minimized],
+            key=lambda c: (c.z, c.window.wid)
+        )
+
+    def assure_correct_layer(self, win):
+        """
+        Make sure that the window is in its designated z-position by moving other
+        windows around as necessary.
+        """
+        if win is None or win.minimized:
+            return
+
+        windows = self._order_clients()
+        if self._previous_layering and self._previous_layering == windows:
+            return
+        self._previous_layering = windows
+
+        index = windows.index(win)
+        if index > len(windows) // 2:
+            for w in windows[index:]:
+                w.window.configure(stackmode=StackMode.Above)
         else:
-            del self.focus_history[index]
-            # return True if win was the last item (i.e. it was current_window)
-            return index == len(self.focus_history)
+            for w in reversed(windows[:index+1]):
+                w.window.configure(stackmode=StackMode.Below)
 
     @property
     def layout(self):
@@ -141,19 +164,17 @@ class _Group(CommandObject):
         self.use_layout((self.current_layout - 1) % (len(self.layouts)))
 
     def layout_all(self, warp=False):
-        """Layout the floating layer, then the current layout.
+        """Layout the normal layer, then the floating layer on top.
 
         If we have have a current_window give it focus, optionally moving warp
         to it.
         """
         if self.screen and self.windows:
             with self.disable_mask(xcffib.xproto.EventMask.EnterWindow):
-                normal = [x for x in self.windows if not x.floating]
-                floating = [
-                    x for x in self.windows
-                    if x.floating and not x.minimized
-                ]
                 screen_rect = self.screen.get_rect()
+                windows = self._order_clients()
+                normal = [w for w in windows if not w.floating and not w.fullscreen]
+                floating = [w for w in windows if w.floating or w.fullscreen]
                 if normal:
                     try:
                         self.layout.layout(normal, screen_rect)
@@ -165,6 +186,7 @@ class _Group(CommandObject):
                 if self.current_window and \
                         self.screen == self.qtile.current_screen:
                     self.current_window.focus(warp)
+                self.assure_correct_layer(self.current_window)
 
     def _set_screen(self, screen):
         """Set this group's screen to screen"""
@@ -271,8 +293,10 @@ class _Group(CommandObject):
             self.focus(win, warp=True, force=force)
 
     def remove(self, win, force=False):
+        hadfocus = (len(self.focus_history) > 0 and self.focus_history[-1] == win)
+        if win in self.focus_history:
+            self.focus_history.remove(win)
         self.windows.remove(win)
-        hadfocus = self._remove_from_focus_history(win)
         win.group = None
 
         if win.floating:
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -98,10 +98,28 @@ class _Group(CommandObject):
         If z-values are equal, the window ids are used for reproducibility.
         The z-values themselves are managed by the layouts.
         """
-        return sorted(
+        clients = sorted(
             [w for w in self.windows if not w.minimized],
             key=lambda c: (c.z, c.window.wid)
         )
+        # check if there are child-windows that need to be moved above their parents
+        toplevel = [w for w in clients if not w.minimized and not w.z.ontopwid]
+        children = [w for w in clients if not w.minimized and w.z.ontopwid]
+        while len(children) > 0:
+            child = children[0]
+            parent = next((p for p in toplevel if p.window.wid == child.z.ontopwid), None)
+            if parent:
+                idx = clients.index(parent)
+                if idx > clients.index(child):
+                    clients.remove(child)
+                    clients.insert(idx, child)
+                toplevel.append(child)
+            else:
+                parent = next((p for p in children if p.window.wid == child.z.ontopwid), None)
+                if parent:
+                    children.append(child)
+            children.pop(0)
+        return clients
 
     def assure_correct_layer(self, win):
         """
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -191,18 +191,20 @@ class Floating(Layout):
         return True
 
     def compute_client_position(self, client, screen_rect):
-        """ recompute client.x and client.y, returning whether or not to place
-        this client above other windows or not """
+        """ recompute client.x, client.y and, if a parent window exists,
+        client.z """
         above = True
 
+        transient_for = client.window.get_wm_transient_for()
+        win = client.group.qtile.windows_map.get(transient_for)
+        if win is not None:
+            client.z.layout = win.z.layout + 1
         if client.has_user_set_position() and not self.on_screen(client, screen_rect):
             # move to screen
             client.x = screen_rect.x + client.x
             client.y = screen_rect.y + client.y
         if not client.has_user_set_position() or not self.on_screen(client, screen_rect):
             # client has not been properly placed before or it is off screen
-            transient_for = client.window.get_wm_transient_for()
-            win = client.group.qtile.windows_map.get(transient_for)
             if win is not None:
                 # if transient for a window, place in the center of the window
                 center_x = win.x + win.width / 2
@@ -273,10 +275,12 @@ class Floating(Layout):
                 bw,
                 bc,
                 above,
+                z=None
             )
         client.unhide()
 
     def add(self, client):
+        client.z.layout = max([c.z.layout + 1 for c in self.clients], default=0)
         self.clients.append(client)
         self.focused = client
 
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -193,8 +193,6 @@ class Floating(Layout):
     def compute_client_position(self, client, screen_rect):
         """ recompute client.x, client.y and, if a parent window exists,
         client.z """
-        above = True
-
         transient_for = client.window.get_wm_transient_for()
         win = client.group.qtile.windows_map.get(transient_for)
         if win is not None:
@@ -209,7 +207,6 @@ class Floating(Layout):
                 # if transient for a window, place in the center of the window
                 center_x = win.x + win.width / 2
                 center_y = win.y + win.height / 2
-                above = False
             else:
                 center_x = screen_rect.x + screen_rect.width / 2
                 center_y = screen_rect.y + screen_rect.height / 2
@@ -228,7 +225,6 @@ class Floating(Layout):
 
             client.x = int(round(x))
             client.y = int(round(y))
-        return above
 
     def configure(self, client, screen_rect):
         if client.has_focus:
@@ -257,15 +253,13 @@ class Floating(Layout):
             client.cmd_bring_to_front()
 
         else:
-            above = False
-
             # We definitely have a screen here, so let's be sure we'll float on screen
             try:
                 client.float_x
                 client.float_y
             except AttributeError:
                 # this window hasn't been placed before, let's put it in a sensible spot
-                above = self.compute_client_position(client, screen_rect)
+                self.compute_client_position(client, screen_rect)
 
             client.place(
                 client.x,
@@ -274,7 +268,6 @@ class Floating(Layout):
                 client.height,
                 bw,
                 bc,
-                above,
                 z=None
             )
         client.unhide()
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -197,6 +197,7 @@ class Floating(Layout):
         win = client.group.qtile.windows_map.get(transient_for)
         if win is not None:
             client.z.layout = win.z.layout + 1
+            client.z.ontopwid = transient_for
         if client.has_user_set_position() and not self.on_screen(client, screen_rect):
             # move to screen
             client.x = screen_rect.x + client.x
@@ -274,8 +275,8 @@ class Floating(Layout):
 
     def add(self, client):
         client.z.layout = max([c.z.layout + 1 for c in self.clients], default=0)
-        if client.group.floating_layout == self:
-            client.z.group = 1  # make sure we are no "manual" floating layout, since that shares its clients
+        if client.group and client.group.floating_layout == self:  # "manual" floating layouts share their clients
+            client.z.group = 1
         self.clients.append(client)
         self.focused = client
 
@@ -283,7 +284,10 @@ class Floating(Layout):
         if client not in self.clients:
             return
 
-        client.z.group = 0
+        if client.group and client.group.floating_layout == self:  # "manual" floating layouts share their clients
+            client.z.group = 0
+        client.z.ontopwid = None
+
         next_focus = self.focus_next(client)
         if client is self.focused:
             self.blur()
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -274,6 +274,8 @@ class Floating(Layout):
 
     def add(self, client):
         client.z.layout = max([c.z.layout + 1 for c in self.clients], default=0)
+        if client.group.floating_layout == self:
+            client.z.group = 1  # make sure we are no "manual" floating layout, since that shares its clients
         self.clients.append(client)
         self.focused = client
 
@@ -281,6 +283,7 @@ class Floating(Layout):
         if client not in self.clients:
             return
 
+        client.z.group = 0
         next_focus = self.focus_next(client)
         if client is self.focused:
             self.blur()
--- a/libqtile/layout/max.py
+++ b/libqtile/layout/max.py
@@ -31,7 +31,8 @@ class Max(_SimpleLayoutBase):
     commands to switch to next and previous windows in the stack.
     """
 
-    defaults = [("name", "max", "Name of this layout.")]
+    defaults = [("name", "max", "Name of this layout."),
+                ("only_focused", True, "Only draw the focused window")]
 
     def __init__(self, **config):
         super().__init__(**config)
@@ -44,7 +45,7 @@ class Max(_SimpleLayoutBase):
         return super().add(client, 1)
 
     def configure(self, client, screen_rect):
-        if self.clients and client is self.clients.current_client:
+        if not self.only_focused or (self.clients and client is self.clients.current_client):
             idx = self.clients.current_index
             order = (self.clients[idx+1:] + self.clients[:idx+1])
             client.place(
--- a/libqtile/layout/max.py
+++ b/libqtile/layout/max.py
@@ -45,16 +45,20 @@ class Max(_SimpleLayoutBase):
 
     def configure(self, client, screen_rect):
         if self.clients and client is self.clients.current_client:
+            idx = self.clients.current_index
+            order = (self.clients[idx+1:] + self.clients[:idx+1])
             client.place(
                 screen_rect.x,
                 screen_rect.y,
                 screen_rect.width,
                 screen_rect.height,
                 0,
-                None
+                None,
+                z=order.index(client)
             )
             client.unhide()
         else:
+            client.z.layout = 0
             client.hide()
 
     cmd_previous = _SimpleLayoutBase.previous
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -497,7 +497,7 @@ class _Window(CommandObject):
         )
 
     def place(self, x, y, width, height, borderwidth, bordercolor,
-              above=False, margin=None, z=NORMAL_LAYER):
+              margin=None, z=NORMAL_LAYER):
         """Places the window at the specified location with the given size.
         """
 
@@ -540,8 +540,6 @@ class _Window(CommandObject):
             width=width,
             height=height,
         )
-        if above:
-            kwarg['stackmode'] = StackMode.Above
 
         self.window.configure(**kwarg)
         self.paint_borders(bordercolor, borderwidth)
@@ -1060,7 +1058,6 @@ class Window(_Window):
                 width, height,
                 self.borderwidth,
                 self.bordercolor,
-                above=True,
                 z=None,
             )
         if self._float_state != new_float_state:
@@ -1414,10 +1411,14 @@ class Window(_Window):
         self.toggle_minimize()
 
     def cmd_bring_to_front(self):
-        if self.floating:
-            self.window.configure(stackmode=StackMode.Above)
+        if not self.floating:
+            self._reconfigure_floating()  # automatically above
+        elif self.group:
+            self.z.layout = max([c.z.layout + 1 for c in self.group.floating_layout.clients], default=0)
+        if self.group:
+            self.group.assure_correct_layer(self)
         else:
-            self._reconfigure_floating()  # atomatically above
+            self.window.configure(stackmode=StackMode.Above)
 
     def cmd_match(self, *args, **kwargs):
         return self.match(*args, **kwargs)
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -396,6 +396,7 @@ class _Window(CommandObject):
         _type = self.window.get_wm_type()
         logger.debug('_NET_WM_STATE (%s): %s', self.name, state)
         logger.debug('_NET_WM_WINDOW_TYPE (%s): %s', self.name, _type)
+        old = self.z.wm
         if _type == 'desktop':
             self.z.wm = DESKTOP_LAYER
         elif 'below' in state:
@@ -406,6 +407,8 @@ class _Window(CommandObject):
             self.z.wm = FULLSCREEN_LAYER
         else:
             self.z.wm = NORMAL_LAYER
+        if self.group and old != self.z.wm:
+            self.group.layout_all()
 
         for s in triggered:
             setattr(self, s, (s in state))
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1420,6 +1420,20 @@ class Window(_Window):
         else:
             self.window.configure(stackmode=StackMode.Above)
 
+    def cmd_keep_above(self):
+        if self.z.interactive == ABOVE_LAYER:
+            self.z.interactive = NORMAL_LAYER
+        else:
+            self.z.interactive = ABOVE_LAYER
+        self.group.assure_correct_layer(self)
+
+    def cmd_keep_below(self):
+        if self.z.interactive == BELOW_LAYER:
+            self.z.interactive = NORMAL_LAYER
+        else:
+            self.z.interactive = BELOW_LAYER
+        self.group.assure_correct_layer(self)
+
     def cmd_match(self, *args, **kwargs):
         return self.match(*args, **kwargs)
 
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -93,6 +93,13 @@ _NET_WM_STATE_REMOVE = 0
 _NET_WM_STATE_ADD = 1
 _NET_WM_STATE_TOGGLE = 2
 
+# z layers
+DESKTOP_LAYER = -1000000
+BELOW_LAYER = -1000
+NORMAL_LAYER = 0
+ABOVE_LAYER = 1000
+FULLSCREEN_LAYER = 1000000
+
 
 def _geometry_getter(attr):
     def get_attr(self):
@@ -138,6 +145,39 @@ def _float_setter(attr):
     return setter
 
 
+class _Z():
+    def __init__(self):
+        self.override = 0  # user level; can be used for user-defined overrides; won't be touched by anything
+        self.wm = NORMAL_LAYER  # wm level; for docks, above, below etc.
+        self.interactive = NORMAL_LAYER  # interactive level; should be one of BELOW_LAYER, NORMAL_LAYER or ABOVE_LAYER
+        self.group = 0  # group level; used to separate "normal" and floating layout
+        self.layout = 0  # layout level; usage optional, defaults to 0
+
+    def __lt__(self, other):
+        if self.override != other.override:
+            return self.override < other.override
+        if self.wm != other.wm:
+            return self.wm < other.wm
+        if self.interactive != other.interactive:
+            return self.interactive < other.interactive
+        if self.group != other.group:
+            return self.group < other.group
+        return self.layout < other.layout
+
+    def __eq__(self, other):
+        return self.override == other.override and \
+               self.wm == other.wm and \
+               self.interactive == other.interactive and \
+               self.group == other.group and \
+               self.layout == other.layout
+
+    def get(self):
+        return self.override, self.wm, self.interactive, self.group, self.layout
+
+    def __str__(self):
+        return str(self.get())
+
+
 class _Window(CommandObject):
     _window_mask = 0  # override in child class
 
@@ -158,6 +198,7 @@ class _Window(CommandObject):
             g = self.window.get_geometry()
             self._x = g.x
             self._y = g.y
+            self._z = _Z()
             self._width = g.width
             self._height = g.height
             self._float_info['width'] = g.width
@@ -167,6 +208,7 @@ class _Window(CommandObject):
             # values on demand.
             self._x = None
             self._y = None
+            self._z = _Z()
             self._width = None
             self._height = None
 
@@ -197,6 +239,7 @@ class _Window(CommandObject):
 
     x = property(fset=_geometry_setter("x"), fget=_geometry_getter("x"))
     y = property(fset=_geometry_setter("y"), fget=_geometry_getter("y"))
+    z = property(fget=_geometry_getter("z"))
 
     @property
     def width(self):
@@ -321,9 +364,22 @@ class _Window(CommandObject):
         if self.qtile.config.auto_fullscreen:
             triggered.append('fullscreen')
 
+        # https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
         state = self.window.get_net_wm_state()
+        _type = self.window.get_wm_type()
+        logger.debug('_NET_WM_STATE (%s): %s', self.name, state)
+        logger.debug('_NET_WM_WINDOW_TYPE (%s): %s', self.name, _type)
+        if _type == 'desktop':
+            self.z.wm = DESKTOP_LAYER
+        elif 'below' in state:
+            self.z.wm = BELOW_LAYER
+        elif 'above' in state or _type == 'dock':
+            self.z.wm = ABOVE_LAYER
+        elif 'fullscreen' in state:
+            self.z.wm = FULLSCREEN_LAYER
+        else:
+            self.z.wm = NORMAL_LAYER
 
-        logger.debug('_NET_WM_STATE: %s', state)
         for s in triggered:
             setattr(self, s, (s in state))
 
@@ -347,6 +403,7 @@ class _Window(CommandObject):
             name=self.name,
             x=self.x,
             y=self.y,
+            z=self.z.get(),
             width=self.width,
             height=self.height,
             group=group,
@@ -440,7 +497,7 @@ class _Window(CommandObject):
         )
 
     def place(self, x, y, width, height, borderwidth, bordercolor,
-              above=False, margin=None):
+              above=False, margin=None, z=NORMAL_LAYER):
         """Places the window at the specified location with the given size.
         """
 
@@ -472,6 +529,8 @@ class _Window(CommandObject):
 
         self.x = x
         self.y = y
+        if z is not None:
+            self.z.layout = z
         self.width = width
         self.height = height
 
@@ -733,7 +792,8 @@ class Static(_Window):
             self.width,
             self.height,
             self.borderwidth,
-            self.bordercolor
+            self.bordercolor,
+            z=None
         )
         return False
 
@@ -817,6 +877,7 @@ class Window(_Window):
 
     @floating.setter
     def floating(self, do_float):
+        self.z.group = 1 if do_float else 0
         if do_float and self._float_state == NOT_FLOATING:
             if self.group and self.group.screen:
                 screen = self.group.screen
@@ -931,7 +992,7 @@ class Window(_Window):
         self.qtile.windows_map[self.window.wid] = s
         hook.fire("client_managed", s)
 
-    def tweak_float(self, x=None, y=None, dx=0, dy=0,
+    def tweak_float(self, x=None, y=None, z=None, dx=0, dy=0, dz=0,
                     w=None, h=None, dw=0, dh=0):
         if x is not None:
             self.x = x
@@ -940,6 +1001,10 @@ class Window(_Window):
         if y is not None:
             self.y = y
         self.y += dy
+
+        if z is not None:
+            self.z.interactive = z
+        self.z.interactive += dz
 
         if w is not None:
             self.width = w
@@ -968,7 +1033,7 @@ class Window(_Window):
         return (self.width, self.height)
 
     def getposition(self):
-        return (self.x, self.y)
+        return (self.x, self.y, self.z.get())
 
     def _reconfigure_floating(self, new_float_state=FLOATING):
         if new_float_state == MINIMIZED:
@@ -996,8 +1061,10 @@ class Window(_Window):
                 self.borderwidth,
                 self.bordercolor,
                 above=True,
+                z=None,
             )
         if self._float_state != new_float_state:
+            self.z.group = 1
             self._float_state = new_float_state
             if self.group:  # may be not, if it's called from hook
                 self.group.mark_floating(self, True)
@@ -1092,6 +1159,7 @@ class Window(_Window):
                 x, y,
                 width, height,
                 self.borderwidth, self.bordercolor,
+                z=None
             )
         self.update_state()
         return False
@@ -1299,17 +1367,17 @@ class Window(_Window):
         """
         self.toscreen(index)
 
-    def cmd_move_floating(self, dx, dy):
+    def cmd_move_floating(self, dx, dy, dz=0):
         """Move window by dx and dy"""
-        self.tweak_float(dx=dx, dy=dy)
+        self.tweak_float(dx=dx, dy=dy, dz=dz)
 
     def cmd_resize_floating(self, dw, dh):
         """Add dw and dh to size of window"""
         self.tweak_float(dw=dw, dh=dh)
 
-    def cmd_set_position_floating(self, x, y):
+    def cmd_set_position_floating(self, x, y, z=None):
         """Move window to x and y"""
-        self.tweak_float(x=x, y=y)
+        self.tweak_float(x=x, y=y, z=z)
 
     def cmd_set_size_floating(self, w, h):
         """Set window dimensions to w and h"""
@@ -1379,9 +1447,9 @@ class Window(_Window):
         return (window.edges[0] <= x <= window.edges[2] and
                 window.edges[1] <= y <= window.edges[3])
 
-    def cmd_set_position(self, dx, dy):
+    def cmd_set_position(self, x, y, z=None):
         if self.floating:
-            self.tweak_float(dx, dy)
+            self.tweak_float(x=x, y=y, z=z)
             return
         for window in self.group.windows:
             if window == self or window.floating:
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -147,12 +147,19 @@ def _float_setter(attr):
 
 
 class _Z():
+    '''A class keeping track of the z-level of a single window. There are multiple mechanisms that can influence the
+    z-level of a window, and thus there are multiple properties provided here to separate those concerns. Ultimately,
+    a list of windows will be ordered by their z-level, so eventually all the different properties will be combined
+    depending on their priority.
+    A window can be kept on top of another by setting `ontopwid`, which should be used for blocking child-windows.
+    '''
     def __init__(self):
-        self.override = 0  # user level; can be used for user-defined overrides; won't be touched by anything
+        self.override = 0  # user level; can be used for user-defined overrides; only touched by user and ScratchPad
         self.wm = NORMAL_LAYER  # wm level; for docks, above, below etc.
         self.interactive = NORMAL_LAYER  # interactive level; should be one of BELOW_LAYER, NORMAL_LAYER or ABOVE_LAYER
         self.group = 0  # group level; used to separate "normal" and floating layout
         self.layout = 0  # layout level; usage optional, defaults to 0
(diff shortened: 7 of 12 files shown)


libqtile/group.py
```diff
@@ -460,12 +540,12 @@ def cmd_prev_window(self):
             return
         if self.current_window.floating:
             nxt = self.floating_layout.focus_previous(self.current_window) or \
-                self.layout.focus_last() or \
-                self.floating_layout.focus_last(group=self)
+                  self.layout.focus_last() or \
```
tych0, line 543: Can you drop all the reformatting bits?

libqtile/group.py
```diff
@@ -30,12 +30,90 @@
...
+
+    #     def insert(self, index, client):
+    #         index = max(len(self.clients), min(0, index))
+    #         if client in self.clients:
+    #             already_at = self.clients.index(client)
+    #             self.clients.remove(client)
+    #             if already_at < index:
+    #                 index = index - 1
+    #         self.clients.insert(index, client)
+    #         return index
```
tych0, line 60: I guess we can drop this?

libqtile/group.py
```diff
@@ -30,12 +30,90 @@
...
 import xcffib.xproto
+from xcffib.xproto import StackMode
 
+from enum import Flag
 from libqtile import hook, utils, window
 from libqtile.command_object import CommandObject
 from libqtile.log_utils import logger
 
 
+class _History:
```
tych0, line 41: Is it possible to add these smarts to ClientList somehow?

tych0:
Overall looks fairly reasonable to me. Needs some tests, though :)

libqtile/group.py
tcld: Sorry for those, that seems to be something Pycharm does automatically. I'll get rid of those when cleaning up the formatting.

libqtile/group.py
tcld: Don't forget this is WIP. :) I hope that the entire commented method can be removed, but for now I just wanted to keep it around.

libqtile/group.py
tcld: That sounds very reasonable, I'll have to check the code of `_ClientList` in detail. Also, my initial intention was to clear the way for layering fake screens, which could make use of `_History` but not `_ClientList`. But that's for a future PR anyway, so maybe I'll not worry too much about it yet.

ramnes:
> One issue I still have, and I hope somebody can give me a pointer here, is that popups (e.g. context menus) on fullscreen windows are currently spawned underneath the window (or not at all?). These popups don't seem to be managed by the group their parent window belongs to - can anybody tell me where they are managed?

IIRC we're not managing them at all but rather let them be handled by the Xorg directly, and yeah, we probably want to change that now.

> I also removed the ability to hide windows from the max-layout so transparent windows can show the windows behind. 

Cool, this has been an issue for quite some time: #415

> That can probably get expensive if a user goes crazy with his transparency settings, and I don't really know if I should add a cutoff (e.g. show a maximum of two windows behind the top one) or just leave the responsibility with the user entirely.

Maybe this should be an option of the layout?

tcld:
Regarding the problems I described before (certain popups seem to spawn below their parent windows, and sometimes child-windows do the same), I just noticed that I get those same problems when just running the master branch (although I haven't merged the commits of this week yet). So I'll probably go and open an issue for that. So far I haven't had sufficient time to properly read up on the topic.

> Maybe this should be an option of the layout?
That's probably a good idea; if it turns out to be an easy addition, I'll definitely add it in. If it becomes too complicated, I'll focus on the main goal first.

tcld:
@ramnes: I took a look at your PR #416 and readded the setting you were using to make the stacking of windows in the max layout configurable (on/off only, since a full "depth"-setting would have been more difficult). I really didn't see why it was reverted, though. Do you happen to remember? It might provide me with some more test cases.

I added a few mentions at the top of this PR due to the amount of issues that are related to this.

I'll also add a short TODO-list here:

- [ ] identify overlap with `ClientList` and see whether `_History` could be merged into it
- [ ] start managing popups so they can't get lost behind their parents
- [ ] try to reduce the frequency at which windows are actually redrawn, since otherwise there is some flicker whenever multiple floating windows overlap and focus is changed
- [ ] allow to refocus fullscreen windows after moving focus to a window behind them (or at least disallow moving the focus back, I guess)

libqtile/group.py
tcld: The classes do similar things, but their scope seems entirely different: One is used by a group (and later, hopefully, by whatever manages screens) and allows only minimal interaction, the other is used by specific layouts and allows for much more control.

If anything, it looks like `_ClientList` could derive from `_History`. "current" in `_ClientList` would be translated to "top" in `_History`, I guess. But overall that would require an almost complete reimplementation of the class.

ramnes:
> @ramnes: I took a look at your PR #416 and readded the setting you were using to make the stacking of windows in the max layout configurable (on/off only, since a full "depth"-setting would have been more difficult). I really didn't see why it was reverted, though. Do you happen to remember?

Yep, have a look at the commit message here: https://github.com/qtile/qtile/commit/61fd091b21633b8035d209ebf29388f54491a800

tcld:
Thank you for the link. I am also using the `Above`-flag to raise windows, although I haven't really observed any problems with that yet.

I just pushed another commit for this. I think things can be tested now - maybe somebody wants to give me a hand? Any feedback on the code would be welcome, too. There is definitely a bit of cleanup to do, and I haven't attempted to merge the `_History`- and the `_ClientList`-classes yet.

The hidden popups from #1893 are not yet addressed (maybe I'll leave that for another PR); the child-windows might be, but I have seen some of the floating windows pop up on unexpected screens (I do not yet know whether that is caused by any of my code - might have to checkout the main branch again). So that's definitely something to keep an eye on.
Also, I really don't know how to test #676 and #1145 . I don't really have any applications that open up notification windows and I am not using tint2.

libqtile/window.py
```diff
@@ -93,6 +93,13 @@
 _NET_WM_STATE_ADD = 1
 _NET_WM_STATE_TOGGLE = 2
 
+# z layers
+DESKTOP_LAYER = -9999
+BELOW_LAYER = -99
+NORMAL_LAYER = 0
+ABOVE_LAYER = 99
+FULLSCREEN_LAYER = 9999
```
ramnes, line 101: Any reason you picked these values rather than others? I feel it's weird that we jump from 99 to 9999.

Also, what would you think of handling fullscreen windows outside of the Z axis? i.e. whatever the Z-index a window have, if there's a fullscreen window, it will be drawn upon it. Then we wouldn't have to need an upper bound, and thus a limit in the number of windows we can order.

That may be the opposite of what I said previously, and I'm sorry if that's the case! But I think we should ensure we only ever have one fullscreen window, and hide all other windows, so it's quite a special case.

ramnes (commented):
I'm so glad this is going forward, keep up the good work!

Are there general ideas that you still would like to change on this PR?

libqtile/backend/x11/xcbq.py
```diff
@@ -622,6 +624,10 @@ def configure(self, **kwargs):
         # since 1.12, uses switches to pack things sensibly
         if float(xcffib.__xcb_proto_version__) < 1.12:
             values = [i & 0xffffffff for i in values]
+        try:  # if we are a window, tell our group to move us (if StackMode is Above or Below)
+            self.group.windows.to(self, kwargs["stackmode"])
+        except (KeyError, AttributeError):
+            pass
```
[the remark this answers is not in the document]
tcld: I just did it here because I wasn't sure this method wouldn't be called for things other than windows. Is that even a possibility?
I guess a type check would achieve the same goals without hiding exceptions.

libqtile/group.py
```diff
@@ -30,11 +30,120 @@
...
+# TODO: 1. I cannot move focus from a "fullscreen"-window to a "normal" window and back.
+#          NOTE: That might be a "feature" of the floating layout.
+#       2. A fullscreen window can spawn its child-popups behind itself.
+#          NOTE: According to @ramnes we aren't managing popups yet; plus, same behaviour happens in the master branch.
+#       3. After restarting qtile, only the topmost window in a max layout is visible.
+#          NOTE: There are more weird things caused by restarts, and according to tych0 the root problem is that
+#                restarts should only be needed during development/testing.
+
+
+class _History(list):
```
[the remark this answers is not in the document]
tcld: I am working on removing this and using `_ClientList` (in layout/base.py) instead. It does feel hacky, though, since that class approaches the "management" of clients very differently and the concepts clash a little.
Do you generally think that the clients should be managed through a bare list, or just that this class shouldn't be added, if possible?

libqtile/group.py
```diff
@@ -75,31 +185,37 @@ def _configure(self, layouts, floating_layout, qtile):
...
     def current_window(self, win):
-        try:
-            self.focus_history.remove(win)
-        except ValueError:
-            # win has never received focus before
-            pass
-        self.focus_history.append(win)
+        self.focus_history.to_top(win)
+        self.windows.to_top(win)  # setting win as current_window moves it to the top of its stack, too
+        self._arrange_z(win)
```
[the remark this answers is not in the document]
tcld: I started in `layout_all`, but it was not the place for it. Here we have the opportunity to just shift the windows that need shifting, exactly when it's necessary. This avoids all the flickering I mentioned in the PR before. (`layout_all` is called for numerous reasons and therefore the windows get unselectively redrawn a lot.)
@tych0 gave me a hint to move away from `layout_all` when I was asking in IRC how to get rid of the flickering.

libqtile/group.py
```diff
@@ -75,31 +185,37 @@ def _configure(self, layouts, floating_layout, qtile):
...
+
+    def _arrange_z(self, win):
```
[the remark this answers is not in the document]
tcld: Both are in order, I guess.

libqtile/backend/x11/xcbq.py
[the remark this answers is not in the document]
ramnes: > PS: The general idea with this addition was that other places moving windows to "Above" or "Below" wouldn't circumvent the z-layering of the group.

Oh, right, then maybe we should rather modify the places where we used the `stackmode` argument of `configure` and replace these with a new function call like `window.above()`? That would make the code more straightforward, with less "spaghettis".

libqtile/window.py
tcld: There is not much reasoning behind the specific numbers.

I wanted to handle all of these cases to adhere to https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER . There might be cases where multiple fullscreen applications exist (or not?), and even cases where they have floating child windows (which get a "+1") or where the user *wants* windows above a fullscreen window. I will think about this for a bit, but currently I think this is a good approach.

libqtile/group.py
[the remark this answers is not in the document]
ramnes: Yeah, I was thinking of reusing `ClientList` as well but then read the code again and it's indeed very hacky as it doesn't serve the same purpose. And I think this will stay hacky as long as we'll let the layouts handle focus and other similar stuff themselves. We could merge these two logics by extracting a lot of functionalities from the layouts, as I already proposed here: https://github.com/qtile/qtile/pull/1500#issuecomment-629820596. But this is definitely outside the scope of this PR so I'd rather just manage the clients through a simple list meanwhile.

tcld:
> Are there general ideas that you still would like to change on this PR?

There are a few that I think need to be adressed or at least thought through:
1. The `_History` class should either be replaced by/merged with `_ClientList` or find a proper home. (Whatever the outcome, I plan to reuse the results for the layering of fake screens later [different PR]).
2. `group._arrange_z` is kind of related to `window.place` and I suspect that there is a prettier solution that maybe combines the two, but so far I haven't found one.
3. Using `StackMode.Above` or `Below` in `_arrange_z` can lead to a shifting focus (a window raised to the top seems to grab focus automatically); it doesn't seem to cause any noticable issues, but it would be nice to disallow the focus shifts while the windows are being reordered.

libqtile/group.py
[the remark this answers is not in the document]
ramnes: Hmm, interesting. From what I recalled, the point of `layout_all` was that we could make a batch of changes on multiple windows and then apply everything at once. If it's flickering, maybe that means that layouts are using it too much and incorrectly? Maybe we should remove `layout_all` completely and do everything on the fly? It looks like I need to dig into that part of the code again. :)

libqtile/window.py
ramnes: > There might be cases where multiple fullscreen applications exist (or not?)

What's the point?

> and even cases where they have floating child windows (which get a "+1") or where the user wants windows above a fullscreen window.

Well, right, you might want your notifications over a movie, for example. But we should not always try to handle all the edge cases, and rather have a strong opinion when it makes sense. I think it would be simpler and quite safe to take the decision that a fullscreen window is above everything, always. That would adhere to the said specification and I don't think anyone would ever complain about this, especially since we have the Max layout that could be used to circumvent this. What do you think?

libqtile/group.py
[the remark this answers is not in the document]
tcld: There might be more code with the class included, but the related code inside `Group` is easy to read as a result of it. I could move the relevant functions into `Group`, but then I can't reuse any of the code for the fake screens later. So I am a little torn on this.

Thanks for sharing your opinion on `_ClientList`. It looks like, if there weren't any layouts using it yet, it would be a good idea to merge the classes or at least have a common base. Either way, I'll drop trying to unite the two for now.

libqtile/window.py
tcld: > What's the point?
Well, a max layout with a hidden bar has the same effect, so there isn't too much of a point.

Initially I actually started by splitting the windows three ways - "normal", "floating" and "fullscreen". The 9999 wouldn't be required in that case, since fullscreen windows will stay on top of others anyway. It did make things a bit easier, but I could go back to that behaviour.

libqtile/group.py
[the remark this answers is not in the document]
ramnes: Well, if you really think that a class is mandatory for this and for the future improvements you have in mind, we have to work with `_ClientList`. While it serves a different purpose (handling the clients in a layout and the layout's focus), it is still a list of clients, and it would be a shame to reimplement half the methods on both sides. What do you think of extracting what can be useful for us here into another class, and make `_ClientList` inherit from this newer class? We could have `ClientList` and `LayoutClientList` or `layouts.ClientList`, for example. It's not optimal but still better than two different classes that have half the methods in common, I think. I wish this was in another PR though.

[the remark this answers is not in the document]
tcld: I'll try to remove the `_History`-class then, but I'll leave at least two of the methods around - likely in `Group`. One change this will lead to overall, though, is that `Group.windows` won't be a set any longer (I have to rely on the ordering) but a proper list. That does open up the possibility of accidentally "misusing" it in the future, since there won't be any protection against duplicate entries. Do you think that's acceptable?

I agree that an extracted base class is the most attractive option here, and that it should be handled in a different PR.

libqtile/group.py
tcld: As discussed with @ramnes, unifying the classes might be possible (although not trivial) and should be handled in a future PR.

libqtile/group.py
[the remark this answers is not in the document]
tcld: The class is removed. Let me know what you think.

libqtile/group.py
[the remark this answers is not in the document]
tcld: Renamed and a short docstring added.

libqtile/window.py
tcld: I removed the 9999-value; fullscreen-windows now have an initial z-value of 0, but they are kept above everything else.

libqtile/backend/x11/xcbq.py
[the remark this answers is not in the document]
tcld: That sounds reasonable. Accessing `StackMode` from various places does feel a little wrong anyway, that kind of stuff should be hidden away in my opinion. We do have a `cmd_bring_to_front` on the window already, although I think it makes normal windows float. Any idea why that is the case? Could I maybe remove the floating-part?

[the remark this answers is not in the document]
tcld: The lines have been removed, and `StackMode` is now mostly set from within `Group`. `Window` has the new functions `to_relative_top` and `to_relative_bottom` (names very much discussable), which should be called when the StackMode needs to be changed on a window.

libqtile/window.py
tcld: I was just testing a few things and opened up a VirtualBox-window in the process. Put it into fullscreen, opened a settings-window and... I couldn't see it, since the fullscreen window stays on top of everything. I think that's not enough of an edge case to ignore.
In addition I installed Tint2 to check out #1145 (I assume that was about floating windows covering the bar), and that isn't fixed yet.

So I actually think we should include the 9999, plus maybe change the behaviour of `group.assure_correct_layer` to allow for any window to be on top of any other window, if the z-value demands it. That would mean no more priorization just by type/state of the window alone.

ramnes: > I was just testing a few things and opened up a VirtualBox-window in the process. Put it into fullscreen, opened a settings-window and... I couldn't see it, since the fullscreen window stays on top of everything. I think that's not enough of an edge case to ignore.

Right. But then what Z value would these window have?

tcld: How about 1000000 (10^6) for fullscreen? Any child-window will get a plus one, respectively. I am using it like that right now, and it works well. Do you see the problem with a fixed value in general or with the specific value of 9999 (as it was before)?

We could go back and remove the section that assigns hard values for specific windows (above, below, etc. - X11 specs), with every window just starting with a 0. But then there would be some more logic needed to handle child-windows or they could be hidden by their parents again. And handling those X11-flags was probably a good idea to begin with.

tcld:
Right now there are two open points I see:
1. Popups aren't handled (and thus #1893 is only partially addressed), but that might be better left for another PR.
2. A way to disallow focus changes while reordering windows would be great. It would also make it possible to use `StackMode.Below` in more cases to reorder less windows.

Since those are more "nice to have", I'll put this up for review now. I'm pretty sure there are still things in the code that aren't up to standard (the layering constants are likely among them again).

ramnes:
Can you rework the commits into logical changes to ease the review and so that we have a nicer log? i.e. merging each commit and its fixes together, and removing commits that were later reverted, to begin with.

tcld:
I cleaned up the commits. Never had to use `git add --interactive` before, but it is even more awesome than `git rebase -i` - so thanks for making me learn something new. :)

libqtile/window.py
```diff
@@ -101,6 +109,7 @@ def get_attr(self):
             # trigger the geometry setter on all these
             self.x = g.x
             self.y = g.y
+            self.z = g.z
```
ramnes, line 112: How is this supposed to work?

libqtile/core/manager.py
```diff
@@ -749,14 +749,17 @@ def cmd_focus_by_click(self, e):
...
+        window = self.windows_map.get(wnd)
         if self.config.bring_front_click:
-            self.conn.conn.core.ConfigureWindow(
-                wnd,
-                xcffib.xproto.ConfigWindow.StackMode,
-                [xcffib.xproto.StackMode.Above]
-            )
+            if window:
+                window.to_relative_top()
+            else:
```
ramnes, line 756: When do we enter in that block?

libqtile/window.py
tcld: I guess you mean that `get_geometry()` wouldn't return a z-property? I hadn't noticed that before, should definitely use the default then.

libqtile/window.py
```diff
@@ -785,6 +811,7 @@ def __init__(self, window, qtile):
             win = qtile.windows_map.get(transient_for)
             if win is not None:
                 group = win._group
+                self.z = (win.z or NORMAL_LAYER) + 1  # stay above parent
```
ramnes, line 831: Why the `or`? `win.z` will always return an integer value, right?

libqtile/window.py
```diff
@@ -936,7 +963,7 @@ def cmd_static(self, screen, x=None, y=None, width=None, height=None):
         hook.fire("client_managed", s)
 
     def tweak_float(self, x=None, y=None, dx=0, dy=0,
-                    w=None, h=None, dw=0, dh=0):
+                    w=None, h=None, dw=0, dh=0, z=None, dz=0):
```
ramnes, line 983: Let's stay consistent and put `z=None` just after `y=None`, and same for `dz`. A quick grep in the codebase told me that we always call the function with the keywords.

libqtile/core/manager.py
tcld: When `window` is at all managed by qtile. If it isn't, it doesn't matter that we can't sync its z-property with a group and we can just do the raw call to `ConfigureWindow`.

libqtile/window.py
tcld: Yes, it will be an integer. The user _could_ set it to `None`, though. Should I rather expect a reasonable user and simplify the line?

libqtile/window.py
```diff
@@ -1403,9 +1435,9 @@ def _is_in_window(self, x, y, window):
         return (window.edges[0] <= x <= window.edges[2] and
                 window.edges[1] <= y <= window.edges[3])
 
-    def cmd_set_position(self, dx, dy):
+    def cmd_set_position(self, dx, dy, dz=0):
         if self.floating:
-            self.tweak_float(dx, dy)
+            self.tweak_float(dx, dy, dz=dz)
```
ramnes, line 1457: I think the keyword you want to use here is `z` actually, not `dz`. Also, let's put the keywords on all arguments (this is the only exception in the codebase to what I just said in my previous comment).

libqtile/window.py
```diff
@@ -1403,9 +1435,9 @@ def _is_in_window(self, x, y, window):
         return (window.edges[0] <= x <= window.edges[2] and
                 window.edges[1] <= y <= window.edges[3])
 
-    def cmd_set_position(self, dx, dy):
+    def cmd_set_position(self, dx, dy, dz=0):
```
ramnes, line 1455: Should be None by default, not 0 (see [previous comment](https://github.com/qtile/qtile/pull/1870/commits/af47ffeb4273da1c0d8f9f6ed5982e22ffc1db4e#r493904171)).

libqtile/window.py
tcld: That sounds good. I didn't want to break anything.

libqtile/core/manager.py
ramnes: Does it make sense that the function works with a window that is not handled by Qtile?

libqtile/window.py
ramnes: Yeah let's keep it simple. I wonder if we really need the +1 also, I think the X server should handle that for us by default.

libqtile/window.py
ramnes: Yes, that's what I meant.

ramnes: As discussed in the other comment, I think we can pretend _z will never be None and just drop the addition.

libqtile/core/manager.py
tcld: `window = self.windows_map.get(wnd)` was already there, I just moved it up a little. I honestly don't know whether a non-managed window makes sense here, but somebody else did.

libqtile/window.py
```diff
@@ -151,13 +160,15 @@ def __init__(self, window, qtile):
         self._float_info = {
             'x': None,
             'y': None,
+            'z': NORMAL_LAYER,
```
ramnes, line 163: Why do you set this if you do not add a `float_z` property and play with it? I think we do not need a `float_z`, and thus do not need this addition. Unless we have a use case where a dedicated Z-value when the window is floating while keeping the original value aside makes sense.

libqtile/window.py
```diff
@@ -602,6 +629,22 @@ def _items(self, name):
     def _select(self, name, sel):
         return None
 
+    def to_relative_top(self, **kwargs):
```
ramnes, line 632: Why "relative"? :)

libqtile/core/manager.py
ramnes: Right! :)

libqtile/window.py
```diff
@@ -602,6 +629,22 @@ def _items(self, name):
...
+        if self.group:
+            self.group.windows.remove(self)
+            self.group.windows.append(self)
+            self.group.assure_correct_layer(self, **kwargs)
+        else:
+            self.window.configure(stackmode=StackMode.Above, **kwargs)
+
+    def to_relative_bottom(self, **kwargs):
```
ramnes, line 640: Let's drop it if we don't use it?

libqtile/window.py
tcld: That's a leftover from before, where the windows were separated into "normal", "floating" and "fullscreen" before any z-values were taken into account. Back then the windows were only moved to the top of their cohort.
It is probably time to rename this to something else now - and think about `cmd_bring_to_front`, where this method could be called instead of doing other things.

libqtile/window.py
tcld: I though it was fitting to have it. It's gone.

tcld: I'll leave the conversation open in case you actually wanted me to remove the `else`, too. That wouldn't be a good idea, though. Windows without groups exist (I might have seen that in some test).

tcld: If you are going through these changes commit by commit, the `to_relative_top`-method will be used quite a bit in later commits. I considered it "setup" and thus included it here. `to_relative_bottom`, however, isn't used anywhere.

tcld:
While looking for a way to add some tests, @m-col suggested to query `_NET_CLIENT_LIST_STACKING` to find out how windows are actually ordered. I thought it looked promising at first, but the list isn't updated when windows are moved across the z-axis.
I stumbled upon an issue in the IceWM repository regarding this topic and in their code they update that list manually. So even though it might not be useful for new tests, that list and `_Group.windows` do have the same purpose, so it would probably be a good idea to keep them in sync somehow. (Maybe in another PR, maybe here.)

libqtile/window.py
ramnes: Then it should be implemented in the commit that actually uses it :)

libqtile/window.py
ramnes: Because I think the arguments are actually misnamed here. They're passed without keyword, so they end up being `x` and `y` according to `tweak_float` signature.

tcld: You are absolutely right! Since the name of the method is `cmd_set_position` and not `cmd_change_position`, I would agree that the variables are misnamed.

tcld: Changed it to `(x, y, z)` (and renamed the corresponding variables, of course).

libqtile/window.py
tcld: I get it. It is now `None`.

libqtile/window.py
tcld: I had that coming.^^ I'll move it when I merge in all the small fixes that resulted from your review.

libqtile/window.py
tcld: I added a short docstring.

libqtile/window.py
m-col: In addition to notifications and child windows, people also use utilities such as dmenu/rofi that create windows that we would expect to be drawn at the very top. This might be a non-issue though, as tools such as those are usually untouched by the WM.

Do we need to handle fullscreen windows differently? What are the issues with the current behaviour? Maybe all they would need is to bring_to_front when fullscreen is enabled, on the occasion when a window that is covered is in focus and fullscreened.

m-col:
> Also, I really don't know how to test #676 and #1145 . I don't really have any applications that open up notification windows and I am not using tint2.

As far as I can tell, #676 is was an issue with the screensaver, rather than Qtile. I think #1145 isn't specific to Tint2, but rather would be about some kind of always-on-top functionality. I think that would be a great addition (in a different PR) where, for example, we could use some kind of matching criteria for setting the always-on-top flag.

libqtile/window.py
tcld: I use dmenu and I have never had any issues. Can't say the same about general popups (#1893).

The only thing that separates a fullscreen window from the common window in this PR is a somewhat higher starting z-value. I wouldn't consider that "handling them differently". And if they had the same z-value as other windows, this PR would be kind of obsolete, wouldn't it?

tcld:
> As far as I can tell, #676 is was an issue with the screensaver, rather than Qtile.

As I mentioned in #676, a test would be great either way. Just to make sure that any future (or current) tinkering doesn't lead to a regression, if that is at all possible.

> I think #1145 isn't specific to Tint2, but rather would be about some kind of always-on-top functionality. I think that would be a great addition (in a different PR) where, for example, we could use some kind of matching criteria for setting the always-on-top flag.

I think I could test #1145 in a VM with a modified config (changing the z-value of Tint2 when it is spawned), just so I know if this PR provided a base for a solution.

m-col:
Right - I think I am misunderstanding the full mechanics of placing windows in
z. So these values are somewhat "default" values for when a new window is
managed? If we set our current window to fullscreen with the Qtile keybinding,
we first update _NET_WM_STATE in the window.fullscreen setter, and then
window.update_state detects this and sets the window's z to FULLSCREEN_LAYER.
This is pretty much equivalent to the current behaviour where fullscreening the
current window brings it to the front, right? Does this PR change any behaviour
for fullscreen windows?

>I use dmenu and I have never had any issues. Can't say the same about general popups (#1893).

OK cool :)

tcld:
> So these values are somewhat "default" values for when a new window is managed?

Yes. If windows with specific WM_STATEs or WM_HINTs show up, they get those default values. Just to make sure that they win out against other windows with lower z. (although there is a point to still be addressed here where custom z-values aren't just overwritten by this mechanism)

> This is pretty much equivalent to the current behaviour where fullscreening the current window brings it to the front, right? Does this PR change any behaviour for fullscreen windows?

Yes, this is pretty much equivalent for fullscreen windows. I don't think anything changes. Even the popups are still broken.^^

m-col (changes requested):
This is looking great! I am very excited for this PR :)

I guess using some kind of algorithm with `StackMode.Above` and `Below` is the only way to order a list of windows? It is a shame that XCB doesn't provide a nicer way.

m-col:
> Yes. If windows with specific WM_STATEs or WM_HINTs show up, they get those default values. Just to make sure that they win out against other windows with lower z. (although there is a point to still be addressed here where custom z-values aren't just overwritten by this mechanism)

OK makes sense!

> Yes, this is pretty much equivalent for fullscreen windows. I don't think anything changes. Even the popups are still broken.^^

Cool :) One thing at a time...

libqtile/window.py
tcld: I would be fine with that, too.

libqtile/window.py
```diff
@@ -93,6 +93,14 @@
 _NET_WM_STATE_ADD = 1
 _NET_WM_STATE_TOGGLE = 2
 
+# z layers
+NORMAL_LAYER = 0
+ABOVE_LAYER = 1000
+FULLSCREEN_LAYER = ABOVE_LAYER*ABOVE_LAYER
+BELOW_LAYER = -ABOVE_LAYER
+DESKTOP_LAYER = -FULLSCREEN_LAYER
+FLOAT_LAYER = (NORMAL_LAYER + ABOVE_LAYER) / 2
```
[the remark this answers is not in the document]
tcld: I would have to check again whether there is a WM_STATE or WM_TYPE hint for that. Currently `FLOAT_LAYER` is just used dynamically when sorting windows, since the fixed value is never assigned. Maybe, at the end of the day, there needs to be a second z-property to differentiate between WM-z and user-defined z.

libqtile/window.py
```diff
@@ -321,9 +332,20 @@ def update_state(self):
...
         state = self.window.get_net_wm_state()
+        _type = self.window.get_wm_type()
+        if True:  # TODO: Do not change z if it was manually set.
+            logger.debug('_NET_WM (%s): %s %s', self.name, _type, state)
+            if _type == 'desktop':
+                self.z = DESKTOP_LAYER
+            elif 'below' in state:
+                self.z = BELOW_LAYER
+            elif 'above' in state or _type == 'dock':
+                self.z = ABOVE_LAYER
```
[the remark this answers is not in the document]
tcld: The bar is not a `Window`, as far as I can tell and, more importantly, it is not managed by a group, so the bar behavior will be untouched by this. With the current implementation z-values are only enforced within groups; I guess to influence a bar (or other docks?) that logic might have to be moved to the screen.

Generally I think it is a good idea to provide these default layers; eventually somebody should be able to tell his dock to hide behind other windows by setting its z - or would that be a bad default?

libqtile/window.py
```diff
@@ -321,9 +332,20 @@ def update_state(self):
         if self.qtile.config.auto_fullscreen:
             triggered.append('fullscreen')
 
+        # https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
         state = self.window.get_net_wm_state()
+        _type = self.window.get_wm_type()
+        if True:  # TODO: Do not change z if it was manually set.
+            logger.debug('_NET_WM (%s): %s %s', self.name, _type, state)
```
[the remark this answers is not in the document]
tcld: Good idea.

libqtile/window.py
```diff
@@ -472,6 +495,7 @@ def place(self, x, y, width, height, borderwidth, bordercolor,
 
         self.x = x
         self.y = y
+        self.z = z
```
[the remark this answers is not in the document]
tcld: I like the suggestion of setting the default to `None` (in `place()`). The current behavior is prone to eliminating z-values that have been set by WM or the user.

libqtile/window.py
```diff
@@ -485,9 +509,10 @@ def place(self, x, y, width, height, borderwidth, bordercolor,
             borderwidth=borderwidth,
         )
         if above:
-            kwarg['stackmode'] = StackMode.Above
+            self.to_relative_top(**kwarg)
+        else:
+            self.window.configure(**kwarg)
```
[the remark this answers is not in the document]
tcld: I guess `to_relative_top` is doing more than it says on the tin, you are right. The purpose was to replace as many calls to `configure` as possible with it, since only the way through `to_relative_top` guarantees that the group notices changes to the window-stack.

The method is also called in `core/manager.py:755`. Passing in `kwargs` is not exactly pretty, but otherwise we would be calling `configure` twice - I don't know how it works in the backend, maybe there is a downside to that?

libqtile/group.py
```diff
@@ -83,23 +85,54 @@ def current_window(self):
...
-            return False
+    def _order_clients(self):
+        """
+        Orders our clients by their z-value, leaving the order as is in case of equal z.
+        """
+        def sort_by(client):
+            if client.floating and not client.fullscreen:
+                return client.z + window.FLOAT_LAYER
+            return client.z
+        return sorted(filter(lambda w: not w.minimized, self.windows), key=sort_by)
```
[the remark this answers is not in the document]
tcld: Right now `FLOAT_LAYER` is not assigned anywhere (I mentioned that in another comment). I'll take a look at that again.

Thanks for the hint about `filter` and the list comprehension, I'll change it and hopefully remember it for the future.

libqtile/group.py
```diff
@@ -83,23 +85,54 @@ def current_window(self):
...
+        windows = self._order_clients()
+
+        # try to avoid moving other windows at all
+        if not win.floating and not win.fullscreen and self.layout.name not in stacked:
+            win.window.configure(stackmode=StackMode.Below, **kwargs)
+            return
+
+        index = windows.index(win)
+        # decide in which direction we have less windows to draw
+        if True:  # index >= len(windows) / 2:
```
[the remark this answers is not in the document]
tcld: I was hoping to find a way to disallow focus changes while this code is being run, since sometimes the `else`-path will be faster. I haven't, so I'll remove the other path.

libqtile/group.py
```diff
@@ -83,23 +85,54 @@ def current_window(self):
...
+    def assure_correct_layer(self, win, **kwargs):
+        """
+        Make sure that the window is in its designated z-position by moving other windows around as necessary.
+        We assume that win is already at the correct position in our list of windows. Nobody else should be
+        calling xcbq.configure with a StackMode (except Popups).
+        """
+        if win.minimized:
+            return
+
+        stacked = ['max']  # z-order in these layouts does matter
```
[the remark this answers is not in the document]
tcld: This is quite the question. I started by doing the layering in `Group.layout_all` following a suggestion by ramnes . To avoid some flickering I switched to reacting to changes of the `current_window` later. But there are points to be made to move the z-management up and/or down the chain of responsibility: The layout should handle its windows, but then you can't force a normal window to stay on top of a floating window, only the group can do that. The group really can't access the bar or other windows outside its realm of responsibility, so maybe the z-layering should be done on a screen-level.

I'll have to think about how to properly "layer" this entire concept of z-values, ideally everybody can take care of its scope itself, but a group should be able to overrule a layout and a screen to overrule a group, if needed. I am fairly sure there is a lot of room for improvement, but it would also mean a lot more changes to the current code.

libqtile/window.py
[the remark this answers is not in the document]
tcld: I made the values constant, most importantly because I agree that the ordering looks much nicer.

About changing focus from normal to floating window:
I mostly use the `Bsp` layout, and it doesn't work there, but that might be because that layout implements commands for all the focus changes itself. I did see, however, that the layouts are actually encouraged to specifically allow switching focus into another layout, which would then have to be handled by the group. So there might be something wrong with this either way.
Also, one of the linked issues is about *keeping* floating windows on top, since it is otherwise very easy to "lose" them - which is related to the first issue of not really being able to switch focus between layouts.
There is quite a bit to this topic, but most of it should probably be handled elsewhere.

libqtile/group.py
```diff
@@ -83,23 +85,54 @@ def current_window(self):
...
             self.focus_history.remove(win)
-        except ValueError:
-            # win has never received focus before
-            pass
+        self.windows.remove(win)
+
         self.focus_history.append(win)
+        self.windows.append(win)
+        if old != self._order_clients():
+            self.assure_correct_layer(win)
```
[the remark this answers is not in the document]
tcld: Done.

libqtile/group.py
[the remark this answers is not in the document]
tcld: After having learned quite a bit about layouts, I have thought about approaching this a little differently to involve the layouts more:

If the layouts got their own z-level (keeping floating above everything else might be sufficient for now), and a way to track the z-levels of clients, they could pass a client's current z-level to `Window.place()`, which every layout calls when asked to `configure` a client. The group would then realize the suggestions of its layouts.
This would remove the ability to set custom z-levels, be it by the user or through `update_state` - those could maybe be addressed by another z-related property on windows (e.g. `z_offset_user` or `z_offset_wm`?).

The group would still be the executive force, but the layouts could give an opinion on where windows should go. The group wouldn't really have to track z-levels itself anymore, just inquire the suggested order from its layouts.

I'll try to implement that approach in the next few days to see where it leads.

libqtile/window.py
m-col: Fair enough :+1:

libqtile/window.py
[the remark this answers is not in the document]
m-col: I was thinking if, for example, the user is focussed on a floating window and then clicks (and focusses) a tiled window, presumably the tiled window now appears in front of the floating window?

libqtile/window.py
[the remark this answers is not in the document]
m-col: The Qtile bar does not get a `Window` but if somebody runs a third party status bar then that will get a `Window` (more specifically, a `Static` one) and get this method. However, like Qtile bars they won't be managed by a group. I'm not sure what implication that might have for z axis control, but maybe that is something for future work (if an issue appears).

In any case, any 'dock' windows will be automatically given `Static` windows and won't be handled by groups, so this if statement will never be met.

libqtile/group.py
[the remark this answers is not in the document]
m-col: OK I've read these comments over and over trying to organise my thoughts :P 

I think the screen shouldn't have anything to do with it. Screens are somewhat above groups and and considering a group _is_ is the collection of windows, it does make sense for Z control to be handled (mostly) at the group level. Consider if you were to move a group to a different screen - you would expect the relative Z positions of its windows to be unchanged.

So that takes us to groups and layouts. I agree with your idea: layouts can have Z control over the tiled windows that belong to that layout. The layout is, in a sense, a tool used by the group to handle a subset of the groups windows. The remainder are floating windows handled directly by the group. I think that ideally a layout would have absolute control over its windows' Z position, much like they are over their windows' X and Y positions. If a user modifies a tiled window's X or Y (by dragging, for example), that window is yanked out of the layout and becomes floating, so the same thing should happen when the Z value is modified by a user.

Does that make sense?

libqtile/group.py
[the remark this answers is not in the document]
m-col: :+1:

libqtile/window.py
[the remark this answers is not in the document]
tcld: That's explicitly forbidden. In the `Max`-layout (and on occasion every other layout) if a floating window were moved behind a tiling window, there would be no way to get it back to the top. Navigating windows over layout boundaries doesn't work for me (although maybe I have to change my configuration for that?), so I think it is crucial never to hide floating windows.

libqtile/window.py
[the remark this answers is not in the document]
tcld: Thank you for the information. Yes, I guess this should be left for the future, then. Some entity beyond the group would have to take care of this, which might even make it necessary to move the entire layering-code there.

libqtile/group.py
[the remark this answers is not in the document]
tcld: Thanks for your input. This is a bit of a complex topic. :) 

> it does make sense for Z control to be handled (mostly) at the group level

You are leaving a bit of room here: If not for `Screens`, then what should manage the layering for windows outside of groups in your opinion? Should it be managed at all?

> The remainder are floating windows handled directly by the group. [...]
> If a user modifies a tiled window's X or Y (by dragging, for example), that window is yanked out of the layout and becomes floating, so the same thing should happen when the Z value is modified by a user.

Well, there is a floating layout managing the groups floating windows. Continuing your thoughts, that would mean that the floating layout mustn't manage z-layers (we have to assume z to be user-modified), right?
For the user, there is not much of a way to fine-control z (as opposed to x and y) - even if additional functions were provided. So maybe some distinction between (x, y) and (z) is advisable.

libqtile/window.py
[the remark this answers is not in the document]
m-col: Hmm. I use a keybinding that moves all floating windows to the front so I can continue using them. That seems more flexible than always keeping them in front, which seems opinionated. Should perhaps floating and tiled windows be treated "equal" (and have potentially mixed Z positions), but the behaviour you want be specified using the "always-on-top" functionality we've discussed before? What do you think @tych0 @ramnes ?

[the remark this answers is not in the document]
tcld: Even if we wanted to treat them equally, we would have two separate layouts, each suggesting z-levels for its clients but unable to take clients of the other layout into account. So the management done by the layouts would be rendered useless and the group would have to take control again as soon as there is a floating window.

There might be some nuggets in the related issue linked above (#1260 ). A layering mechanism is mentioned in there, and I would favor that.
Since you have a keybinding to work around this: Do you deliberately put your floating windows in the background for later use or is it just something that happened and you needed to come up with a workaround?

libqtile/layout/max.py
```diff
@@ -44,7 +45,7 @@ def add(self, client):
         return super().add(client, 1)
 
     def configure(self, client, screen_rect):
-        if self.clients and client is self.clients.current_client:
+        if not self.only_focused or (self.clients and client is self.clients.current_client):
```
tych0, line 48: Can you add some reasoning for this change? Or better yet, can we drop it all together?

tcld: That's basically the change that allows the `Max` layout to display multiple windows on top of each other. That's really nothing I would want to remove!

tych0: Ok, but why is that necessary? Max should only ever display one window, maximized. It's very unintuitive.

tcld: See #415 . The added code gives the user the option to display more than one window - only really useful when using transparency, but then it is very neat.

libqtile/window.py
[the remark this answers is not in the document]
m-col: Looks like I actually got my keybinding from that issue! I deliberately put my floating windows in the background (by focussing another window that happens to be in front of it) and then use the keybinding to get it back.

> we would have two separate layouts, each suggesting z-levels for its clients but unable to take clients of the other layout into account
Right yeah that sounds tricky. I wonder if the behaviour I'm after (and currently use) is simply achieved using the already available `bring_front_click` setting anyway. With that disabled, tiled windows will never move in front of floating windows by following your rules. So maybe this is a non-issue and your way works best :)

libqtile/layout/max.py
tych0: Let's default it to off, then.

libqtile/window.py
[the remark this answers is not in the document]
tcld: I wasn't really aware of that setting. Without it, clicked (and/or focused) windows would never change their layer (even within a layout)?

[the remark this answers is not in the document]
m-col: Not by clicks, at least, no.

libqtile/layout/max.py
tcld: Ok.

libqtile/window.py
tcld: I removed this and some other stuff in an attempt to shift responsibility for the layering to the layouts (a little).

libqtile/window.py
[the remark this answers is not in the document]
tcld: That's a bit of a shame. My floating layout raises the focused window at the moment, I'll have to check whether that behaviour is canonical. If not, I'll have to stop it from doing that.

libqtile/window.py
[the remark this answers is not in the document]
tcld: I removed this. And I stopped worrying about calling `configure` to often, I'll leave that to #1924.

libqtile/group.py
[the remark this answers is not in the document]
tcld: I removed `FLOAT_LAYER` and shortened the function.

libqtile/group.py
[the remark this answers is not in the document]
tcld: I added a number of changes, some of which reverting previous changes. Therefore I also cleaned up the commit history again, followed by a forced push. I hope that's ok, since it "detaches" a lot of comments in here.

The relevant layouts (`Max` and `Floating`) now suggest a z-level for their clients. The group is then asked to enforce that in `layout_all`. The code has gotten simpler because of it, so I have a strong feeling that this is the right direction.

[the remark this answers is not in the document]
tcld: Also, the hard reference to `Max` has been removed. I'll mark this conversation resolved in a few days, unless something new comes up.

tcld:
I reworked a lot of the changes. The intention was to give more responsibility to the layouts when z-levels for clients are decided. This resulted from one of the discussions with @m-col above, and I believe the changes are a nice improvement.

libqtile/group.py
```diff
@@ -90,16 +93,33 @@ def current_window(self, win):
             pass
         self.focus_history.append(win)
 
-    def _remove_from_focus_history(self, win):
```
tcld, line 93: I could have added this back in after the most recent changes (and I still can, if it is wanted). But I really don't like this function - it seems overly long, is only used once, and the name doesn't reflect what actually needs to be done that one time.

libqtile/window.py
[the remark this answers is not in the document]
tcld: I have just been playing around with `bring_front_click` in this branch. With this PR, the setting was useless within a floating layout, since the focused window would always also rise to the top of the float stack. One could, however, click on any tiled window to raise it to the top.

So I tested another approach, where the floating windows are not rearranged along z on focus change (there are two additional commits for this). Now one has to click to raise floating windows again (if `bring_front_click` is true). The only issue is that, as soon as `layout_all` is called, the original z-order is restored (e.g. whenever a new window is added or removed). Fixing that isn't trivial, since the call to raise the window happens in a place where communication with the layout would seem a bit out of place.

tcld:
I added some tests to include the z-properties of windows set by `Max` and `Floating`. Also, `bring_front_click` is now handled almost as expected. I'm not quite sure what _is_ expected, I guess, since there are a few weird cases (rasing a `Max`-window above a floating window, for example).

It would be great if somebody found the time to give this another look. The approach for handling the z-layers has changed a bit. Except for very few places that raise windows without changing their z (in window.py and manager.py), the solution seems very solid to me. I have been running the PR for about a week now.

libqtile/layout/max.py
```diff
@@ -44,17 +45,21 @@ def add(self, client):
...
+            order = (self.clients[idx+1:] + self.clients[:idx+1])
             client.place(
                 screen_rect.x,
                 screen_rect.y,
                 screen_rect.width,
                 screen_rect.height,
                 0,
-                None
+                None,
+                z=order.index(client)
```
m-col, line 58: Won't `order.index(client)` always equal `len(self.clients)`, which is easier to use than line 50?

m-col: `len(self.clients) - 1`, in fact

m-col: Oh, no, I totally misread the code. Never mind - this makes sense.

libqtile/layout/floating.py
```diff
@@ -259,10 +259,12 @@ def configure(self, client, screen_rect):
             bw,
             bc,
             above,
+            z=None
         )
         client.unhide()
 
     def add(self, client):
+        client.z = 0 if len(self.clients) == 0 else max(self.clients, key=lambda c: c.z).z + 1
```
m-col, line 267: I wasn't sure what was fastest so I tested this keyed max() call against list comprehension:
```
import random
import timeit

class A:
    def __init__(self, z):
        self.z = z

clients = [A(i) for i in range(10)]
random.shuffle(clients)

print(timeit.timeit(
    'max(clients, key=lambda c: c.z).z', globals={'clients': clients}, number=1000000
))
print(timeit.timeit(
    'max([c.z for c in clients])', globals={'clients': clients}, number=1000000
))
```
Which gave (in seconds):
```
1.3131903419998707
0.8532288009992044
```
So it's close, but looks like list comprehension here wins the speed test.

libqtile/window.py
```diff
@@ -785,6 +810,7 @@ def __init__(self, window, qtile):
             win = qtile.windows_map.get(transient_for)
             if win is not None:
                 group = win._group
+                self.z = win.z + 1  # stay above parent
```
m-col, line 813: Something similar to this is being done with `Floating.compute_client_position`, I think it would be a bit neater to remove this, but do the equivalent in that function in the floating layout, replacing what it currently does with the `above` variable, i.e. `Floating.compute_client_position` ought to set `client.z` as it does with X and Y. All the `above` stuff can be removed, (including from `Window.place`), as this whole PR supercedes it.

libqtile/layout/floating.py
tcld: I'll change it, thanks for the test. Although we are talking about 0.5 microseconds whenever a window is added.^^

m-col:
lol yeah. But these microseconds all add up... to a millisecond!

libqtile/window.py
tcld: Thanks for spotting this, seems a very good idea.
I'll read through the code for the floating layout. Kind of neglected that so far.

tcld: So I have read through floating.py (I was surprised it isn't using `_ClientList`) and I followed your suggestions. This was indeed a very good thing to spot, things seem more natural now.

tcld: I have yet to fully grasp what removing the `above`-flag from `Window.place` would mean. Will read a little more code for that.

libqtile/group.py
```diff
@@ -90,16 +93,33 @@ def current_window(self, win):
...
+            return
+
+        windows = self._order_clients()
+        if self._previous_layering and self._previous_layering == windows:
+            return
+        self._previous_layering = windows
+
+        index = windows.index(win)
+        for w in windows[index:]:
+            w.window.configure(stackmode=StackMode.Above)
```
m-col, line 122: Does this reduce calls? (untested, may not actually make sense):
```
index = windows.index(win)
if index > len(windows) // 2:
    for w in windows[index:]:
        w.window.configure(stackmode=StackMode.Above)
else:
    for w in windows[:index]:
        w.window.configure(stackmode=StackMode.Below)
```

tcld: That's actually what I initially tried to do, back when the z-updates where tied to changing the `current_window`. In there, however, moving the current window down led to focus changes and thus to an infinite recursion. I'll check how things look now.

libqtile/window.py
tcld: It seems *almost* possible to remove the `above`-flag, only the popup-class would still require it. Since these popups aren't managed by a group, I don't see a way right now to fully remove it.

libqtile/group.py
tcld: Other than some trouble with my picture-in-pciture fake screen (which never really worked anyway, which is how this entire z-business started), it seems valid to do this now. At least I didn't observe any problems. I'll put this in the next commit.

libqtile/window.py
m-col: Maybe just remove the `above=True` bit from the Popup class, as it's likely that it is redundant anyhow.

tcld: But then who is going to raise that window to the top?

m-col:
I guess when `Popup.unhide` is called it gets manually sent to the top of the
stack anyway, so it probably doesn't need the `above=True` anyway.

libqtile/window.py
tcld: > I guess when `Popup.unhide` is called it gets manually sent to the top of the
> stack anyway, so it probably doesn't need the `above=True` anyway.

I'll give it a shot. Just a little afraid that I won't notice if anything breaks; I'm not sure if there is anything actually using the `Popup`-class.

m-col:
It was added with some other features in mind, such as notification windows and
widget tooltips, but currently nothing within libqtile uses it so no worries.

tcld:
Locally I have been playing around with using an actual class for the z-property (instead of just a single int). It has the following properties:

```python
self.override = 0  # user level; can be used for user-defined overrides; won't be touched by anything
self.wm = NORMAL_LAYER  # wm level; for docks, above, below etc.
self.interactive = NORMAL_LAYER  # interactive level; should be one of BELOW_LAYER, NORMAL_LAYER or ABOVE_LAYER; used by cmd_keep_above and cmd_keep_below
self.floating = False  # group level; used to separate "normal" and floating layout
self.layout = 0  # layout level; usage optional, defaults to 0
```

This does work well, but I can't figure out how to integrate `Window.cmd_bring_to_front` and `Qtile.focus_by_click`. Those functions just raise a window, which can't really be done through a simple change in z. The functions would have to communicate with _all_ layouts to make sure that the windows place is secured. Maybe even with the group.

tcld:
As described in my last comment, I replaced the z-property with an actual class. This now allows to handle (almost) all the cases where windows are moved along z. The only outstanding issue I know of is that `Qtile.focus_by_click` is currently not allowed to raise normal windows. Even raising floating windows seems a bit dirty, but it can be done (as long as their `floating`-property is set, I guess - which won't be the case for a manually configured floating layout).

I would appreciate if somebody could have a look at the code. I have the previous versions of the code backed up in local branches, of course. But I really wasn't getting anywhere with a single scalar property.

libqtile/window.py
```diff
@@ -556,6 +554,7 @@ def user_placed_window_setup(self, borderpixel, borderwidth):
         self.borderwidth = borderwidth
         self.bordercolor = borderpixel
         self.window.configure(borderwidth=borderwidth, stackmode=StackMode.Above)
+        # TODO: Could we handle this with a change of the z-layer? Is the window even part of a group?
```
tcld, line 557: @m-col mentioned PR #1697 on IRC, which would remove this entire function anyway. I didn't bother inventing a fix because of that.

tcld:
The Python 3.5-test keeps failing because of a list that's in a different order. I'm going to have to look into what changed with Python 3.6 in that regard. Although, I read somewhere that we would only have to support the last four versions, and since Python 3.9 just came out, maybe the need to fix that test will go away soon?

tcld:
I moved this back to a draft. It works really well for me, but #1943 also touches the floating code, so it has become difficult to separate which bugs are in master, which ones might be fixed by #1943 and what should be adressed here. One thing at a time for now. :) (Also, maybe support for Python 3.5 will be successfully dropped before merging this, then I won't have to fix the failing tests.)

tcld:
With some other PRs recently merged and Python 3.5 support dropped now, I would consider this "done". I have been running this PR for a few months now, without any issues to speak of. The only change I made in recent weeks is that final commit, taking care of keeping child-windows above their parents in a floating layout. It is essential for fullscreen windows with children (but can still be overruled if `bring_front_click` is set to `True`).

tcld:
Still having problems handling `bring_front_click`, since it leads to modifying the z-property in a way that isn't really reversible. So I added an extra commit for that topic; I don't like it one bit but I just can't seem to come up with anything better. The entire approach with the z-property seems to overcomplicate that one situation way too much.

libqtile/window.py
```diff
@@ -239,7 +239,7 @@ def __init__(self, window, qtile):
 
     x = property(fset=_geometry_setter("x"), fget=_geometry_getter("x"))
     y = property(fset=_geometry_setter("y"), fget=_geometry_getter("y"))
-    z = property(fset=_geometry_setter("z"), fget=_geometry_getter("z"))
+    z = property(fget=_geometry_getter("z"))
```
tych0, line 269: I think you can fold this commit into the one that adds it

CHANGELOG
```diff
@@ -25,6 +25,8 @@ qtile X.X.X, released XXXX-XX-XX:
           AND-semantics, allowing for more complex and specific rules
         - Python 3.9 support
         - switch to Github Actions for CI
+        - added `only_focused` setting to Max-layout, allowing to draw multiple
+          clients on top of each other when set to False
```
tych0, line 29: Can you add a "Fixes #415" tag to the commit message here?

libqtile/window.py
```diff
@@ -321,9 +391,25 @@ def update_state(self):
...
             triggered.append('fullscreen')
 
+        # https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
         state = self.window.get_net_wm_state()
+        _type = self.window.get_wm_type()
+        logger.debug('_NET_WM_STATE (%s): %s', self.name, state)
+        logger.debug('_NET_WM_WINDOW_TYPE (%s): %s', self.name, _type)
+        old = self.z.wm
+        if _type == 'desktop':
+            self.z.wm = DESKTOP_LAYER
```
tych0, line 401: Might be nice to add some tests for this behavior.

libqtile/group.py
```diff
@@ -44,6 +45,7 @@ class _Group(CommandObject):
 
     A group is identified by its name but displayed in GroupBox widget by its label.
     """
+
```
tych0, line 48: ?

tych0:
Overall this feels pretty close to being done. There are a few places where it interacts with X properties that would be nice to have tests, but other than that (and a few minor commit/message nits and random whitespace hunks we can drop), it looks pretty good to me.

libqtile/window.py
tcld: Done.

CHANGELOG
tcld: You mean the commit message or also the CHANGELOG?

I also added a "Closes #753" to the first commit message.

libqtile/group.py
tcld: Removed.

libqtile/window.py
tcld: I'll have a look.

tcld:
Thanks for checking out the code! I would agree that this is close to being done, the only detail I have been going over again and again is a part of that final commit. Applying `bring_front_click` to tiled windows is something I can't seem to find a satisfactory solution for.

tcld:
After thinking about this a lot, I have come to the conclusion that using a fully fletched z-property is cumbersome in many ways. There is quite a bit of code here that only achieves very little in my opinion. Therefore I will put this on hold (I'll keep the branch around for now) and move to a different approach (PR incoming).

Issue #1868, opened 2020-08-16
Changing the value of the Backlight widget sets it to 100%


qwenger:
# Issue description

When changing the backlight value e.g. by scrolling on the widget the value gets set to 100% no matter what.

I think that it is caused by https://github.com/qtile/qtile/commit/b524e4427ca43f77d022aa05d85dc16a1aef7215#diff-49f3d455c9019a091f0f74c63fb8aff6: `xbacklight` expects a percentage, not the actual value found in the ACPI file (see https://linux.die.net/man/1/xbacklight). That commit removed the conversion to percentages.

For reference, on my computer:
```
$ cat /sys/class/backlight/intel_backlight/max_brightness
24242
```

# Qtile version

Version 0.16.1 (current from https://www.archlinux.org/packages/community/x86_64/qtile/). Used to work but I'm not sure of which version.

m-col:
Can you post your config for the widget?

I think it might be that [this line](https://github.com/qtile/qtile/blob/b524e4427ca43f77d022aa05d85dc16a1aef7215/libqtile/widget/backlight.py#L79)
should be executed only if `self.change_command` is None.

qwenger:
Nothing fancy really:

```
screens = [
    Screen(
        bottom=bar.Bar(
            [
                ...
                widget.Backlight(backlight_name="intel_backlight", step=5),
                ...
            ]
        )
    )
]
```

No that line is not problematic, or only as a secondary problem. In `button_press` the absolute value is calculated (the value w.r.t. the maximum value from the ACPI file) and passed to `change_backlight` and from there to the `change_command`. However the default `change_command` uses `xbacklight`, which expects a percentage value rather than absolute.

Please compare with the previous implementation: https://github.com/qtile/qtile/blob/5ea22c655b4591c1603724e28431293306679c5d/libqtile/widget/backlight.py#L105

m-col:
Please could you try #1871 and see if that commit fixes it?

qwenger:
Thank you very much, yes that does the trick!

ramnes:
Fixed in 4d4636efe9a8771e6272c64d90e7e22fbdf4e095

PR #1871, opened 2020-08-17
Clean up logic in Backlight widget

### LINKED ISSUE #1866: Missing file for backlight should not crash whole setup
# Issue description

After upgrade qtile 1.5.1 -> 1.6.1, the configuration crashed. The problem is I have same setup for two PC. Both have different GPU: intel and Nvidia. The older version just printed an error message inside the plugin. Not cool, but at least the environment worked. 

I am not sure if it is a bug or a feature, but I prefer to have a more resilient config. I do not see a point of crashing for such situations. 

# Qtile version

1.6.1

# Stack traces
```
  File "/usr/lib/python3.8/site-packages/libqtile/confreader.py", line 92, in load
    config = __import__(name)  # noqa: F811
  File "/home/neworld/.config/qtile/config.py", line 363, in <module>
    widget.Backlight(
  File "/usr/lib/python3.8/site-packages/libqtile/utils.py", line 213, in class_proxy
    return cls(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/libqtile/widget/backlight.py", line 96, in __init__
    self.max_value = self._load_file(self.max_brightness_file)
  File "/usr/lib/python3.8/site-packages/libqtile/widget/backlight.py", line 105, in _load_file
    raise RuntimeError(
RuntimeError: Unable to read status for max_brightness
```

# Configuration

```
                widget.Backlight(
                    background=color_red,
                    backlight_name = 'intel_backlight',
                ),
```

m-col:
Move button_press logic into mouse_callbacks and fix #1866 and #1868.

### DIFF
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -85,7 +85,7 @@ class Backlight(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Backlight.defaults)
-        self.future = None
+        self._future = None
 
         self.brightness_file = os.path.join(
             BACKLIGHT_DIR, self.backlight_name, self.brightness_file,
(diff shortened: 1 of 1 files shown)


libqtile/widget/backlight.py
```diff
@@ -108,24 +113,22 @@ def _load_file(self, path):
...
     def _get_info(self):
         brightness = self._load_file(self.brightness_file)
-
-        info = {
-            'brightness': brightness,
-            'max': self.max_value,
-        }
-        return info
+        max_value = self._load_file(self.max_brightness_file)
+        return 100 * brightness / max_value
```
ramnes, line 117: Should we keep the dict format here?

m-col: What for? _get_info is only used when we want the percentage

ramnes: Ah sorry, I thought it was related to the IPC.

neworld:
Works fine as before :+1:

m-col:
Looks like something weird happened with the py35 test run. Could someone check it out and see if it was unexpected?

ramnes:
Let's move on and assume the fault is on Travis. :)

PR #1872, opened 2020-08-18
Added memory percentage from psutil to memory widget

lKk2:
Motivation: i love to use percentage on my widgets and i saw there was no built-in in our memory widget so i just added from the psutil itself.


ramnes:
Thanks!

PR #1875, opened 2020-08-19
add traverse cmds to focus windows based on position

m-col:
Posting these to see if there is interest in these focussing functions that I've been using for some time. It provides `qtile.cmd_traverse_{up,down,left,right}` functions that move focus to another visible window based entirely on position.

Pros:

- it doesn't care about screens, so you can focus between them easily
- it doesn't care if windows are tiled or floating, it just works

Caveats:

- Only really makes sense in layouts where tiles do not overlap (but in that case don't use them)

The algorithm to choose which window to focus to might seem a bit weird, but it has been refined over months of use ;) The basic idea is that it considers the centre point of windows that are visible (current groups of all screens), then tries to find the first one in the desired direction within a span of the currently focussed window e.g. if we focus to the right, we first look for those centre points that fall with y values between the top and bottom y values of the current window. If there is none, then rather than looking in a band, we just look at any windows in that direction. The aim is for the focussing to feel the most intuitive and natural.

This would also be a good focus mechanism for a true floating layout (#1444).

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1474,6 +1474,74 @@ class Qtile(CommandObject):
         """Delete a group with the given name"""
         return self.delete_group(group)
 
+    def cmd_traverse_up(self):
+        """ Traverse focus to a window upward """
+        self.traverse(-1, 'y')
+
+    def cmd_traverse_down(self):
+        """ Traverse focus to a window downward """
+        self.traverse(1, 'y')
+
+    def cmd_traverse_left(self):
+        """ Traverse focus to a window leftward """
+        self.traverse(-1, 'x')
+
+    def cmd_traverse_right(self):
+        """ Traverse focus to a window rightward """
+        self.traverse(1, 'x')
+
+    def traverse(self, direction, axis):
+        """ Focus to a visible window in a given direction based on position. """
+        win = None
+        win_wide = None
+        dist = 10000
+        dist_wide = 10000
+        cur = self.current_window
+        if not cur:
+            cur = self.current_screen
+        if axis == 'x':
+            dim = 'width'
+            band_axis = 'y'
+            band_dim = 'height'
+            cur_pos = cur.x
+            band_min = cur.y
+            band_max = cur.y + cur.height
+        else:
+            dim = 'height'
+            band_axis = 'x'
+            band_dim = 'width'
+            band_min = cur.x
+            cur_pos = cur.y
+            band_max = cur.x + cur.width
+
+        cur_pos += getattr(cur, dim) / 2
+        windows = [w for g in self.groups if g.screen for w in g.windows]
+        windows.extend([s for s in self.screens if not s.group.windows])
+        if cur in windows:
+            windows.remove(cur)
+        for w in windows:
+            if isinstance(w, Screen) or not w.minimized:
+                pos = getattr(w, axis) + getattr(w, dim) / 2
+                gap = direction * (pos - cur_pos)
+                if gap > 5:
+                    band_pos = getattr(w, band_axis) + getattr(w, band_dim) / 2
+                    if band_min < band_pos < band_max:
+                        if gap < dist:
+                            dist = gap
+                            win = w
+                    else:
+                        if gap < dist_wide:
+                            dist_wide = gap
+                            win_wide = w
+
+        if not win:
+            win = win_wide
+        if win:
+            self.focus_screen(win.group.screen.index)
+            win.group.focus(win, True)
+            if not isinstance(win, Screen):
+                win.focus(False)
+
     def cmd_add_rule(self, match_args, rule_args, min_priorty=False):
         """Add a dgroup rule, returns rule_id needed to remove it
 


ramnes:
YES! I've been thinking of doing this for a very long time. I'm happily trying the branch right now.

m-col:
:D

ramnes:
It's very nice, thank you so much for bringing this!

One thing I would change from a user point of view (and it's very subjective) is that I would prefer to go to the previous or next group when focusing out of a screen, rather than focusing the other screen. I don't mind implementing this myself in my configuration, although I'm wondering if we could put this logic at the screen or group level, and add another logic at the manager level that doesn't care about screens? That way I could hook myself on the screen/group commands.

Or to write it simpler, what about a `screen.cmd_traverse` that doesn't leave screen boundaries, and `qtile.cmd_traverse` that would use `screen.cmd_traverse` but across screens.?

m-col:
Oh I'd never thought about making it go between groups on the same screen, kind
of like wrapping the groups... that sounds like it could be very useful so I
agree we should add it rather than just in your config. So yep I agree with your
last sentence completely :)

Should we do all that in this PR or merge this and then add the screen side
separately? Perhaps you could do the screen side so you can design exactly the
focus logic you are imagining when going between groups?

m-col:
Actually it should be relatively straightforward, we just need a mechanism to start the "band" at the side of a given group for all 4 sides, and then the inter-group traversal just selects the next/prev group, whereas the inter-screen traversal selects the next visible group in that direction. Then once they've selected the group it works the same. 

Then again, with inter-screen we'd want to look in the band on all screens before then moving to outside the band for all screens, but with inter-group we would want to do in-band then out-band for group 1, then group 2 sequentially. So when searching on a group we need to tell it to either look in a band or in the entire height/width of the screen then the object that is doing the searching can either do band->band->screen->screen for inter-screen or band->screen->band->screen for inter-group. Does that description make any sense? :P

ramnes:
I think the behavior you've implemented is pretty sane (and would fit nicely in the default config), so I think we should start by having this merged. But I also think we should make many more behaviors available. I'm opening an issue to discuss this. :)

ramnes:
Any reason you called this `traverse_{direction}` rather than `focus_{direction}`? I think it would be easier to understand, what do you think?

m-col:
I was about to say that now is the time to rename it before more work on it is done! I chose traverse to distinguish it from the other focus cmds as its behaviour is different to how the layouts generally work, but i think focus is more clear if you're not worried about that.

ramnes:
I am not, but let's hear what others think.

One thing I noticed in this implementation is that if two windows are equidistant from the current window, the same window will always be chosen. I would prefer that we look at the focus history to choose between the two.

ramnes:
Also, I'm seeing a weird behavior: it jumps from a screen to another with `traverse_up`, but not with `traverse_down`, with the opposite on the other screen. I wouldn't expect that the focus moves across screens vertically since my screens are horizontally aligned:

```
DP2-2 connected primary 2560x1440+0+0 (normal left inverted right x axis y axis) 710mm x 400mm
eDP1 connected 1920x1080+2560+0 (normal left inverted right x axis y axis) 290mm x 170mm
```

m-col:
> One thing I noticed in this implementation is that if two windows are equidistant from the current window, the same window will always be chosen. I would prefer that we look at the focus history to choose between the two.

Yeah that would be great.

> it jumps from a screen to another with traverse_up, but not with traverse_down, with the opposite on the other screen

I think this would be because the centre points of the two screens differ in y value, so its considering them as above and below as a fallback to no screen being directly below. I've been thinking that maybe instead of the in-band, out-band search method it could search outward in like a funnel shape from the side of the current window:

```
          ___
         | A |
         |___|
        /  |  \
       /   v   \
      /         \
     /           \
     ___
    | B |  ___
    |___| | C |
          |___|

```
Focussing down from A would focus B rather than C, whereas with the current behaviour C would be focussed. With this behaviour, the screen issue would also disappear.

ramnes:
Ah and, mhh... That doesn't work well with Max, which is quite obvious, but raises the question of how users are expected to handle this in their config.

m-col:
I'm not sure I follow. Surely the current implementation also doesn't work with Max? It is targeted more for layouts with non-overlapping windows.

ramnes:
Yes, as I said it's obvious that it doesn't work. My point is that people will ask how to use these functions and still use the Max layout, and we should have an answer.

tcld:
This seems very interesting, I very much hope it'll be available sooner rather than later. :)

@m-col: Since you mentioned that the algorithm has been refined over several months, have you ever considered to extract a function that could "measure" windows and screens, something to be fed into e.g. a sorting algorithm to rank things from "best next focus candidate" to worst?

I expect that looking at the focus-history (or, rather, the "focus distance" between two windows) should help with overlapping windows within a group. Although it is certainly not trivial, especially when jumping between groups, since right now we can't compare focus-histories.

tych0:
Can you add some tests for this? I'd like to merge it and use it :)

PR #1877, opened 2020-08-20
Widgets A-C: move button_press logic into mouse_callbacks

m-col:
These changes show a way we could hook up the current button_press behaviour of widgets into the newer mouse_callbacks behaviour.

The currentlayout widget uses methods on the Qtile object so it sets the mouse_callbacks in _configure; the Qtile object is `None` during __init__ (currently - I thought there was an issue about this but I cannot find it).

Do you guys think this is a reasonable way to do the widget updates?


libqtile/widget/cmus.py
```diff
@@ -45,6 +46,12 @@ def __init__(self, **config):
         self.status = ""
         self.local = None
 
+        mouse_callbacks = {
+            'Button1': self.play,
+            'Button4': partial(subprocess.Popen, ['cmus-remote', '-n']),
+            'Button5': partial(subprocess.Popen, ['cmus-remote', '-r']),
+        }
```
tych0, line 53: This seems like an unassigned local that never goes anywhere? `self.mouse_callbacks` maybe?

tych0:
Seems reasonable, although one small nit.

Can you also put this reasoning in the body of the commit message?

PR #1881, opened 2020-08-22
Move border painting with X pixel cache into X backend

m-col:


tych0:
Doesn't the client return false or something if the call fails? Perhaps we should just check the return value?

m-col:
You're absolutely right. Changed it so that it will fail if `color_pixel()` fails.

ramnes:
I just reverted this in d50bfde012f0d9182f5b00e1e3938d17a64959d4, as it completely breaks Qtile when making any window fullscreen. See the traceback in the commit message.

m-col:
Sad times. It looks like the problem was more X-specific code outside of the X backend so I can submit a new PR and hunt around for some other bits that could be moved there too.

m-col:
It's strange the tests didn't catch it... it isn't clear to me why it passed test_toggle_fullscreen

PR #1884, opened 2020-08-26
Move _all_ border painting into X backend

m-col:
A previous commit moved most border painting into the backend, but did
not update Window.user_placed_window_setup in window.py, resulting in
errors.


ramnes:
Thanks!

PR #1887, opened 2020-08-28
Update geometry properties when user-specified placing is used

m-col:
When windows are positioned using user-specified x, y, width and height
by the X server, these properties of the corresponding
libqtile.window.Window object are not updated.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -19,8 +19,6 @@
 # SOFTWARE.
 import array
 import contextlib
-import inspect
-import traceback
 import warnings
 
 import xcffib.xproto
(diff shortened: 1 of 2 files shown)


ramnes:
I'm not sure this is the most reasonable place in the execution flow to do this. Can we make it sooner, i.e. when the window is being created or handled?

ramnes:
Also maybe we could set a variable in the window like `user_placed` at the same time and remove that function.

m-col:
I've updated it with a WIP commit that will make lots of tests fail because I can't figure out how it's all working :P

The primary issue is that `window.Window`'s x, y, widith and height properties become incorrect when a window is positioned with "user specified" geometry, causing issues when custom scripting use these properties.

This led me to try to merge `Window.user_placed_window_setup` with `Window.place` so have just one entrypoint to placing windows, but it doesn't appear to be as simple as that.

I'm at a loss currently so if anybody has any ideas that is great, otherwise I will take a break from this and try again with fresh eyes at some point :)

libqtile/window.py
```diff
@@ -197,24 +165,45 @@ def __init__(self, window, qtile):
         }
         self.update_hints()
 
-    x = property(fset=_geometry_setter("x"), fget=_geometry_getter("x"))
-    y = property(fset=_geometry_setter("y"), fget=_geometry_getter("y"))
+    @property
+    def x(self):
```
tych0, line 169: Can we make these function calls? I know qtile uses properties a lot, but I find it quite hard to follow, and I'd like to avoid more of it.

ramnes: FWIW it's not "more" of it since these were properties already. I'd rather not change the API here, although I do agree that we should probably pay more attention to the properties we add in the future.

m-col:
I've done some bad PR hygiene and solved the problem that this was trying to solve in another PR. This commit: https://github.com/qtile/qtile/pull/1697/commits/42b59fa5d5011d89f214c2d30822554099c84b5e which part of the multiple-borders PR, because it was a necessary precursor (although it should have been its own PR!).

libqtile/window.py
tych0: Ah, yes, I missed that they were properties before. So ignore me :)

tcld:
I just checked out this branch. Can't seem to put a tiled window into fullscreen mode anymore (well, the window seems to grow accordingly, but it isn't moved on top of the bar). But this does seem to fix #1893.

tych0:
Ok, so should we close this and can you send one that just fixes the single bug?

m-col:
I'm happy to close this. The problem this solves is in my other PR but should be another. As other work is currently being done that might affect this behaviour anyway I will hold off for a bit.

PR #1895, opened 2020-09-03
Add optional preferences to utils.guess_terminal

CharString:
This makes it a tiny bit easier to have a portable config that works on
machines where your favourite terminal isn't available.


tych0:
Thanks!

CharString:
I'm just sharing alike... Thanks for qtile and the quick merge.

PR #1897, opened 2020-09-05
Let widgets send notifications during cmd_s with notify=True

m-col:
This introduces two options for all widgets, 'notify' and
'notify_timeout', which will be used by supporting widgets to send
notifications upon status changes. For example, currently the backlight
and volume widgets will, if notify=True is passed, send notifications
when the backlight and volume changes. The send_notifications has been
updated to allow for optional passing of an 'id', which by the
notification service spec should make notifications replace visible
notifications of the same id, so e.g. if you bind
Backlight.cmd_increase_backlight to a key and hit it many times, only
one notification should appear, and it should get replaced/updated
instead of loads of separate notifications appearing.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -25,9 +25,21 @@ import sys
 import traceback
 import warnings
 from collections.abc import Sequence
+from random import randint
 from shutil import which
 
 from libqtile.log_utils import logger
+
+_can_notify = False
+try:
+    import gi
+    gi.require_version("Notify", "0.7")  # type: ignore
+    from gi.repository import Notify  # type: ignore
+    Notify.init("Qtile")
+    if Notify.get_server_info()[0]:
+        _can_notify = True
+except Exception:
+    pass
 
 
 class QtileError(Exception):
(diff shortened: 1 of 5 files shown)


m-col:
The failed check looks to be a problem with how mypy does the checks. An exception is being raised inside a module-level try-except statement and it doesn't like it. Is there a good reason why it might be bad to do the imports like that?

tych0:
Yeah, it does seem like a mypy bug. I'd just annotate it somehow with an ignore and add a little note saying why.

ramnes:
Looks interesting, but what's the use case? I'm not sure to understand why I'd want a notification if I already have the widget text.

m-col:
The use case is visual feedback when using cmd_s bound to keys. E.g. I use emoji
volume indicators which don't aren't very precise; the notification gives me the
percentage.

However, my setup might be unusual and so actually maybe it is best we don't
merge this unless it is explicitly asked for by someone else. I can implement it
easily in my config, and would prefer to err on the side of not adding more code
to qtile :)

ramnes:
Let's see what others think but if you can have the same behavior in your config without pain, I think it's better not to add this, and rather put a snippet in the docs or wiki.

m-col:
I implemented this in my config and it is easy to do, just adding `send_notification` into the widgets that I want, so I think adding this would be feature creep tbh. I will close.

Issue #1120, opened 2018-01-23
Unicode character in group name breaks qtile restart


doenerkebap:
Having a group name with a unicode character produces the following error when trying to restart qtile:

```
2018-01-23 02:44:33,242 libqtile manager.py:handle_KeyPress():L910  KB command error restart: Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/libqtile/command.py", line 97, in call
    return (SUCCESS, cmd(*args, **kwargs))
  File "/usr/local/lib/python3.5/dist-packages/libqtile/manager.py", line 1387, in cmd_restart
    argv.append('--with-state=' + buf.getvalue().decode())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xdc in position 218: invalid continuation byte

```

Sweenu:
I'm not answering the issue but you can use labels instead and have a unicode-free name:
```python
Group('music', label='Åll the Ünicøde ¢haracters yøu wañt')
```

doenerkebap:
As far as I can see, the label feature was introduced only recently in https://github.com/qtile/qtile/commit/31abc0f2dac95b270351eabb56d9d8071a78a60b . I installed version 10.7 via pip, so it is not available to me, right?
Is there an easy way to upgrade to the latest github version. Can the develop branch recommended to be used by end users, or will i run into further trubles?

Sweenu:
The 10.7 version is almost a year old. In my opinion, you're better off using the develop branch. You'll get the fixes and features and you shouldn't have more problems because each pull request is tested against the whole test suite. 
You can build it from source or, if you're on Arch, install the  qtile-python3-git AUR package.

flacjacket:
This seems like a shortcoming of pickling everything to serialize between restarts. In general, I think we need a more robust way to serialize and deserialize state, there are several issues related to the way that we restart and how we restore the previous state.

Also, I can work on getting a release out this weekend so you can get caught up :)

tcld:
There has been a recent change to the serialization code by @m-col: #1898 fixes this issue.

m-col:
Great, thanks. Looks like these are (almost) duplicates, so makes sense.

PR #1898, opened 2020-09-06
Store state as bytes in a tmp file (fixes #1888)

### LINKED ISSUE #1888: lazy.restart() fails when a group has a non-ASCII name
# Issue description

After adding middle-dots in my group names, lazy.restart() fails. It does the same with a set of non-ASCII characters I tested: "àç".

# Qtile version

0.16.2.dev21+ge71fae80

Using qtile-git from the AUR.

# Stack traces

Here is the error that seems relevant to the issue

```
2020-08-29 02:09:05,707 ERROR libqtile manager.py:process_key_event():L374 KB command error restart: Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/libqtile/command_interface.py", line 312, in call
  File "/usr/lib/python3.8/site-packages/libqtile/core/manager.py", line 1147, in cmd_restart
    self.restart()
  File "/usr/lib/python3.8/site-packages/libqtile/core/manager.py", line 290, in restart
    argv.append('--with-state=' + buf.getvalue().decode())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe0 in position 163: invalid continuation byte
```

And the entire log : [qtile.log](https://github.com/qtile/qtile/files/5144542/qtile.log)

# Configuration

A 1-line config is sufficient to produce the issue on my laptop

```python
from libqtile.config import Group

groups = [Group(i) for i in "azà"]
```

m-col:


tych0:
Why does this fix it?

m-col:
It avoids any decoding/encoding steps and saves the state directly as pickled data. Any state that works fine with Qtile will not be modified in any way (beside pickling) across restarts.

It works fine with the minimum-breaking config in the issue, but it is possible that I misunderstand encodings :)

libqtile/core/manager.py
```diff
@@ -189,8 +190,9 @@ def noop(qtile):
         hook.fire("startup")
 
         if state:
-            st = pickle.load(io.BytesIO(state.encode()))
             try:
+                with open(state, 'rb') as f:
+                    st = pickle.load(f)
```
tych0, line 195: I think we need to remove the file here.

tych0:
Ok - can you put that analysis in the commit message? Also, we need to remove the file once we're done with it. But other than that, LGTM, thanks!

libqtile/core/manager.py
m-col: I forgot to add this to the update I just pushed... sigh. I omitted it from the initial PR on the basis that "it will be removed when the system shuts down" and "maybe someone will want to save it elsewhere for usage", but I guess not all systems use a tmpfs for /tmp, and I doubt anybody would want it. I'll update that now.

libqtile/core/manager.py
```diff
@@ -189,8 +190,10 @@ def noop(qtile):
...
+                os.remove(state)
```
tych0, line 196: Can we put this in a finally: block maybe? Just to be really sure it does get deleted.

tcld:
This also fixes #1120, I just tested it on my machine.

PR #1899, opened 2020-09-07
Reconfigure screens upon screen_change events

### LINKED ISSUE #1874: Old groups remain after lazy.restart()
# Issue description

When I change group names and do ` lazy.restart() ` I have both old and new groups. Plus, if I have an error in config.py I get "asdfuiop" groups from default config and after fixing errors and another ` lazy.restart() ` I have my groups and additional 8 groups. They are not only in **GroupBox** widget, ` qtile-cmd -o cmd -f groups ` also shows old group on the list.

# Qtile version

0.15.1 from Manjaro distro, also have this issue with 0.16.2.dev8+gbe769d82

# Stack traces

Restart gives no errors stack traces, only a warning.
`2020-08-19 00:16:37,217 WARNING libqtile manager.py:maybe_restart():L269 Restarting Qtile with os.execv(...)`

# Configuration 

I set groups like this:

```
group_names = [
    ("Term ", {'layout': 'matrix'}),
    ("Dev ", {'layout': 'max'}),
    ...
]

groups = [Group(name, **kwargs) for name, kwargs in group_names]

for i, (name, kwargs) in enumerate(group_names, 1):
    keys.append(Key([mod], str(i), lazy.group[name].toscreen()))
    keys.append(Key([mod, "shift"], str(i), lazy.window.togroup(name)))
```

m-col:
This makes Qtile internally respond to changes in physical monitor setup
by reconfiguring screens, making it unnecessary to fully restart Qtile
to match the setup. Fixes #1874.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -779,28 +779,6 @@ class Connection:
             if i in self._extmap:
                 setattr(self, i, self._extmap[i](self))
 
-        self.pseudoscreens = []
-        if "xinerama" in extensions:
-            for i, s in enumerate(self.xinerama.query_screens()):
-                scr = PseudoScreen(
-                    self,
-                    s.x_org,
-                    s.y_org,
-                    s.width,
-                    s.height,
-                )
-                self.pseudoscreens.append(scr)
-        elif "randr" in extensions:
-            for i in self.randr.query_crtcs(self.screens[0].root.wid):
-                scr = PseudoScreen(
-                    self,
-                    i["x"],
-                    i["y"],
-                    i["width"],
-                    i["height"],
-                )
-                self.pseudoscreens.append(scr)
-
         self.atoms = AtomCache(self)
 
         self.code_to_syms = {}
@@ -809,6 +787,31 @@ class Connection:
 
         self.modmap = None
         self.refresh_modmap()
+
+    @property
+    def pseudoscreens(self):
+        pseudoscreens = []
+        if hasattr(self, "xinerama"):
+            for i, s in enumerate(self.xinerama.query_screens()):
+                scr = PseudoScreen(
+                    self,
+                    s.x_org,
+                    s.y_org,
+                    s.width,
+                    s.height,
+                )
+                pseudoscreens.append(scr)
+        elif hasattr(self, "randr"):
+            for i in self.randr.query_crtcs(self.screens[0].root.wid):
+                scr = PseudoScreen(
+                    self,
+                    i["x"],
+                    i["y"],
+                    i["width"],
+                    i["height"],
+                )
+                pseudoscreens.append(scr)
+        return pseudoscreens
 
     def finalize(self):
         self.cursors.finalize()
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -195,6 +195,9 @@ class Qtile(CommandObject):
         self.update_net_desktops()
         hook.subscribe.setgroup(self.update_net_desktops)
 
+        if self.config.reconfigure_screens:
+            hook.subscribe.screen_change(self.cmd_reconfigure_screens)
+
         hook.fire("startup_complete")
 
     def _prepare_socket_path(
@@ -299,6 +302,7 @@ class Qtile(CommandObject):
             self.screens.append(s)
 
     def _process_screens(self) -> None:
+        self.screens = []
         if hasattr(self.config, 'fake_screens'):
             self._process_fake_screens()
             return
@@ -324,6 +328,23 @@ class Qtile(CommandObject):
 
             scr._configure(self, i, x, y, w, h, grp)
             self.screens.append(scr)
+
+    def cmd_reconfigure_screens(self, ev=None):
+        """
+        This can be used to set up screens again during run time. Intended usage is to
+        be called when the screen_change hook is fired, responding to changes in
+        physical monitor setup by configuring qtile.screens accordingly. The ev kwarg is
+        ignored; it is here in case this function is hooked directly to screen_change.
+        """
+        logger.info("Reconfiguring screens.")
+        self._process_screens()
+
+        for group in self.groups:
+            if group.screen:
+                if group.screen in self.screens:
+                    group.layout_all()
+                else:
+                    group.hide()
 
     def paint_screen(self, screen, image_path, mode=None):
         self.core.painter.paint(screen, image_path, mode)
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -174,6 +174,7 @@ floating_layout = layout.Floating(float_rules=[
 ])
 auto_fullscreen = True
 focus_on_window_activation = "smart"
+reconfigure_screens = True
 
 # XXX: Gasp! We're lying here. In fact, nobody really uses or cares about this
 # string besides java UI toolkits; you can see several discussions on the
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,11 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - memory widget can now be displayed with decimal values
         - new "qtile migrate" command, which will attempt to upgrade previous
           configs to the current version in the case of qtile API breaks.
+        - A new `reconfigure_screens` config setting. When `True` (default) it
+          hooks `Qtile.reconfigure_screens` to the `screen_change` hook,
+          reconfiguring qtile's screens in response to randr events. This
+          removes the need to restart qtile when adding/removing external
+          monitors.
     * bugfixes
         - fix KeyboardLayout widget (xkb-switch is new dependency)
 
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -164,6 +164,10 @@ configuration variables that control specific aspects of Qtile's behavior:
              fontsize=12,
              padding=3)
       - Default settings for bar widgets.
+    * - reconfigure_screens
+      - True
+      - Controls whether or not to automatically reconfigure screens when there
+        are changes in randr output configuration.
     * - wmname
       - "LG3D"
       - Gasp! We're lying here. In fact, nobody really uses or cares
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -58,6 +58,7 @@ class Config:
         ("widget_defaults", "Dict[str, Any]"),
         ("extension_defaults", "Dict[str, Any]"),
         ("bring_front_click", "bool"),
+        ("reconfigure_screens", "bool"),
         ("wmname", "str"),
     ]
 
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -295,21 +295,10 @@ class Subscribe:
     def screen_change(self, func):
         """Called when a screen is added or screen configuration is changed (via xrandr)
 
-        Common usage is simply to call ``qtile.cmd_restart()`` on each event
-        (to restart qtile when there is a new monitor):
-
         **Arguments**
 
             * ``xproto.randr.ScreenChangeNotify`` event
 
-        Examples
-        --------
-
-        ::
-
-            @libqtile.hook.subscribe.screen_change
-            def restart_on_randr(ev):
-                libqtile.qtile.cmd_restart()
         """
         return self._subscribe("screen_change", func)
 


m-col:
Correction - this does not _fix_ #1874, as that would be resolved by completely re-configuring groups across restarts rather than keeping around groups that aren't in the config.

m-col:
Should we merge this? I've been using it since I posted it here and it works really well. Anyone else tested it?

m-col:
I tell a lie - I actually have a minor change that I never pushed to github, and I've been using that commit. I'll push that now and then it should be good to go.

m-col:
Scratch that - I already pushed it 18 days ago...

tcld:
I have a question about this:
On my setup I frequently swap out display one with a projector; both have the same resolution, so after a while of using Qtile I noticed that I didn't require a restart at all and everything would still work fine after a switch. With this PR, however, Qtile would react to the switch in some manner. Do you have an idea what that would entail? (If you rebase this, I may give it a go on my system.)

I took a look at the code, by the way. Looks good to me, if that counts for anything. :+1:

m-col:
Thanks for the heads up, just rebased.

The `screen_change` hook is fired when there is an change in the geometry of the X screen. This is more of a software thing, i.e. if you enable/disable/reposition external monitors using xrandr/arandr, then it is fired. It is _not_ fired when there is only a hardware change, so your disconnecting and reconnecting would be unaffected if you aren't touching the randr configuration.

Please do try it! I think it would be really valuable, as it would start to move qtile restarts into the realm of "only when you change configuration", which would help close a few open issues.

tcld:
I'll check this out immediately. Any suggestions what I should test? I can swap monitors, and maybe move screens in arandr a little. But that's all I can come up with.

m-col:
I guess that is all really, as well as enabling/disabling them of course. None
of it should require a restart, and everything should continue to work as
normal.

Some things might need manual setup to work nicely though. For example, I
configure 2 `Screens`s each their a bar, each with a GroupBox widget. So when I
have 1 screen I want the first widget to show all of my groups (6 of them),
whereas if I have 2 screns then I split them 3 and 3 across the 2 screens. The
2nd groupdbox widget always shows the last 3 groups, but it isn't drawn if there
is only 1 screen. 

```
def _refresh_screens():
     if len(qtile.screens) > 1:
         groupboxes[0].visible_groups = ['1', '2', '3']
     else:
         groupboxes[0].visible_groups = ['1', '2', '3', 'q', 'w', 'e']
     groupboxes[0].bar.draw()

hook.subscribe.screen_change(lambda *_: qtile.call_later(0.2, _refresh_screens))
hook.subscribe.startup(lambda *_: qtile.call_later(0.2, _refresh_screens))
```

tcld:
Playing with arandr seems to work pretty well. I moved around screens and swapped the primary and secondary one. Haven't disabled anything, since my config right now sets up two screens and I don't know what Qtile would do if there was only one.

The only issue I have observed is that two of my widgets are kind of drawn on top of each other: `YahooWeather` and `Systray`. I think the latter is partially drawn on top of the former. But I also think that I have observed this behaviour once before, so the cause is likely outside of this PR.

tcld:
I noticed one more thing: When moving a screen that displays a window, it seems the window coordinates aren't updated until I focus the window again. (That's for tiled windows, haven't tested floating ones yet.)
Since the screen_rect changed, maybe a `layout_all` should be triggered after changing anything about the screens?

m-col:
Ah yeah good idea.

tcld:
`layout_all` fixes the problem for tiled windows. Floating windows aren't moved, however, which can technically leave them entirely off-screen when screens are repositioned. I assume that's a problem with the floating layout rather than your PR, though.

tcld:
I've been using this for a few days now. There are some occasions where my screens are re-detected even though I wouldn't want it, however. It seems that randr is very trigger-happy when it comes to swapping displays or launching fullscreen games, which often times leads to my displays switching places. (I have already gotten used to my systray being messed up most of the time.)
Coming from cinnamon, I am very familiar with these exact issues. I was surprised to see that Qtile actually handled these edge cases much better by just doing nothing and assuming nothing.

I am still very much for including this PR, but maybe there is a way to make the functionality a little less intrusive?

m-col:
What do you mean your screens are redetected? Qtile (with this PR) doesn't 'do'
anything with randr, only responding to the randr events emitted by the X server
and enabling/disabling Screens accordingly. Your displays 'switching places': is
this the content (e.g. windows, bar), or completely switching so that the cursor
has to go off the opposite edge to what you'd expect to jump to the other
monitor? Also, presumably you didn't get these issues before trying out this
commit, right? I find it very strange you'd get these issues... if that is the
case, could you put something in the screen_change hook on master and see if it
is still being triggered lots when it shouldn't be?

tcld:
> only responding to the randr events emitted by the X server
That's what I mean: randr is quite trigger-happy, Qtile is not at fault.

> Your displays 'switching places': is this the content ... or completely switching so that the cursor has to go off the opposite edge ...
The latter.

> presumably you didn't get these issues before trying out this commit
That is true, but only because Qtile didn't react to any of the randr events before.

> could you put something in the screen_change hook on master and see if it is still being triggered lots when it shouldn't be
The assumption being that I see the same amount of events? I would hope so, but what would it tell us?

m-col:
> That's what I mean: randr is quite trigger-happy, Qtile is not at fault.

I see.

>The assumption being that I see the same amount of events? I would hope so, but what would it tell us?

Yeah, don't worry about it, that was just me thinking about how to confirm the
above. But I do wonder if useful information could be extract from the events
that are generated. randr shouldn't be trigger-happy. Indeed I do not see the
same. What X version are you using (mine is 1.20.9-2)? Do you have anything like
auto-randr running?

As the main intention of this PR is to remove the need to restart when
reconfiguring monitors, we could move the hook from libqtile into the default
config. This way people who have issues with it can bind it to something else,
like a key combo, rather than upon screen_change firing.

tcld:
Same version of `xorg-server` here. I used autorandr in the past, but it isn't automatically called by anything at the moment. I'll uninstall it and make sure it isn't being used anywhere. Other than that I have xrandr and arandr installed. But they shouldn't act on their own.

Moving the hook might be a good idea, but I also wouldn't want my possibly esoteric use-case to stand in the way of a cool feature.

m-col:
I have moved to hook subscription to the default config with a description of what it does, and added a note to the changelog for discoverability. This can help us avoid situtations where people have a trigger-happy randr for some reason!

Beside that, tests are all good, so this is good to go if someone could do a quick review.

libqtile/core/manager.py
```diff
@@ -360,6 +360,32 @@ def _process_screens(self) -> None:
             )
             self.screens.append(scr)
 
+    def reconfigure_screens(self, ev):
+        """
+        This configures screens upon firing of the screen_change hook.
```
tcld, line 365: Maybe the comment should be changed or removed, since the connection between `reconfigure_screens` and the mentioned hook is now optional.

libqtile/resources/default_config.py
```diff
@@ -196,3 +196,10 @@
 # We choose LG3D to maximize irony: it is a 3D non-reparenting WM written in
 # java that happens to be on java's whitelist.
 wmname = "LG3D"
+
+
+# This hook causes qtile to reconfigure its screens when it receives randr
+# screen change events, for example when reconfiguring monitors with xrandr.
+@hook.subscribe.screen_change
+def reconfigure_screens(e):
+    qtile.reconfigure_screens
```
tcld, line 205: This doubles as an example of how to use hooks, which is nice. :+1:

The call should probably be `qtile.reconfigure_screens(e)`, or is there some Python magic going on that I don't know about?
Also, would you mind calling the parameter `ev` or even `event`? For me personally, the biggest challenge with Python continues to be finding out what type an object has. More elaborate names help a little.

libqtile/core/manager.py
m-col: Good call. Done.

libqtile/resources/default_config.py
m-col: Nope, that was me making a dumb typo when I wrapped `qtile.reconfigure_screens` inside the top-level function.

Thanks for the feedback, I've updated the commit.

libqtile/core/manager.py
```diff
@@ -361,6 +361,35 @@ def _process_screens(self) -> None:
...
+            if i < len(self.screens):
+                scr = self.screens[i]
+            elif i < len(self.config.screens):
+                scr = self.config.screens[i]
+            else:
+                scr = Screen()
+
+            scr._configure(self, i, x, y, w, h, self.groups[i])
+            scr.group.layout_all()
+            screens.append(scr)
```
ramnes, line 385: Can we use `_process_screen` here instead of repeating the code?

libqtile/core/manager.py
```diff
@@ -361,6 +361,35 @@ def _process_screens(self) -> None:
...
+
+        for group in self.groups:
+            if group.screen and group.screen not in screens:
+                group.screen = None
```
ramnes, line 389: Can you tell me more about this? In particular, why are we not assigning the leftovers to one of the new screens?

ramnes (changes requested):
Looks pretty good overall, and most importantly it works well! That's quite simple and still, it's a big improvement for Qtile. :)

libqtile/resources/default_config.py
```diff
@@ -196,3 +196,10 @@
 # We choose LG3D to maximize irony: it is a 3D non-reparenting WM written in
 # java that happens to be on java's whitelist.
 wmname = "LG3D"
+
+
+# This hook causes qtile to reconfigure its screens when it receives randr
+# screen change events, for example when reconfiguring monitors with xrandr.
+@hook.subscribe.screen_change
+def reconfigure_screens(event):
+    qtile.reconfigure_screens()
```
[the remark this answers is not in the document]
m-col: 👍🏽

libqtile/backend/x11/xcbq.py
```diff
@@ -779,13 +779,26 @@ def __init__(self, display):
...
+        self.refresh_keymap()
+
+        self.modmap = None
+        self.refresh_modmap()
+
+    def finalize(self):
+        self.cursors.finalize()
+        self.disconnect()
+
+    def get_pseudoscreens(self):
```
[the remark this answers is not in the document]
m-col: Yep!

libqtile/backend/x11/xcbq.py
```diff
@@ -779,13 +779,26 @@ def __init__(self, display):
         extensions = self.extensions()
         self.screens = [Screen(self, i) for i in self.setup.roots]
         self.default_screen = self.screens[self.conn.pref_screen]
+        self.atoms = AtomCache(self)
```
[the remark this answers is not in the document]
m-col: Yep!

libqtile/core/manager.py
m-col: This is more for when there are now fewer screens present: any groups that were previously on a screen that is now not there will have have their `screen` reset to `None`. When there are more screens available, then nothing will happen because `group.screen` will be in `screens`.

That does make me wonder -- what happens (on master) when there are more screens than there are groups? I have not accounted for that possibility in this PR.

libqtile/core/manager.py
m-col: I believe so! Just pushed.

m-col:
What do you think about changing the method to `cmd_reconfigure_screens` and exposing it to the command interface?

libqtile/core/manager.py
ramnes: Yeah, my question was badly formulated but that was my concern as well.

Also, I'm thinking that we probably hold a lot more references to the removed screens than just in groups. What should we do with them?

ramnes:
Yeah I think we should have it as a command anyway.

libqtile/backend/x11/xcbq.py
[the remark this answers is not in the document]
ramnes: It's still not where I wanted to try to place it. :sweat_smile: I meant above `self.code_to_syms = {}` (and a blank line), just as in the master branch.

libqtile/core/manager.py
```diff
@@ -326,39 +329,66 @@ async def finalize(self):
...
             if not self.current_screen:
                 self.current_screen = scr
 
             scr._configure(
                 self, i, x, y, w, h, self.groups[i],
             )
-            self.screens.append(scr)
+            screens.append(scr)
+
+        return screens
```
ramnes, line 372: Why not modifying `self.screens` directly?

libqtile/core/manager.py
```diff
@@ -326,39 +329,66 @@ async def finalize(self):
...
+        screens = []
         screen_info = self.core.get_screen_info()
 
         for i, (x, y, w, h) in enumerate(screen_info):
-            if i + 1 > len(self.config.screens):
-                scr = Screen()
-            else:
+            if i < len(self.screens):
+                scr = self.screens[i]
+            elif i < len(self.config.screens):
```
ramnes, line 359: I feel this might be prone to bugs, wouldn't it be safer to use the screen names instead?

ramnes (approved):
LGTM overall :)

libqtile/core/manager.py
m-col: I just checked all references to screens the other things that save screens directly (rather than accessing via a group or `qtile`) are: `Gap`s/`Bar`s, `widget.Systray`, and `window.Static`. I guess that sounds about right, too.

- Static windows can just be re-positioned to be on the first screen (TODO).
- Gaps and Bars do not need to be handled further; they exist within `qtile.config.screens` and are simply reconfigured when their screen (index) is reconfigured, and this does very little as bars are already set up for multiple calls to `Bar._configure`. Widgets on bars from dormat screens continue to live (and poll) but cannot be accessed until their screen is restored.
- `widget.Systray` _may_ show possibly unexpected behaviour, but I'm not familiar enough with X's systrays to know for sure. I do however use this widget and PR together (and have done for months) and have had no problem. The widget saves an atom telling the server which screen it is on. The screen number doesn't really change (it just disappears) so it will be correct when the screen is restored.

libqtile/core/manager.py
m-col: Hmm it made sense while I was hooking it up with _process_screens but rethinking it, some of these changes might not be necessary.

libqtile/core/manager.py
m-col: What bug risk do you mean? And screen names?

m-col: The `len(self.screens)` test might be redundant.

ramnes: I'm worried that we use the existing screens on one side of the condition, and config screens on the other side. My brain is quite dead at this time of the day so please bear with me if I'm saying anything dumb, but I have the impression the user change the order of the screens in its configuration, we may end up in a situation where a same Screen object is used for two different physical screens.

m-col: Got it. Using `self.screens` here was redundant anyway, as the same objects are available in the config. Removing it also will prevent that issue you describe :)

m-col:
Alright, I think that's it done now.

`window.Static` will likely need a bit of modification (e.g. giving it a `set_screen` method to change its screen and redraw it), and `Qtile.cmd_reconfigure_screens` will need updating in order to call that method on all static windows that are stuck to now absent screens. However, static windows and their interactions with screens are already broken, but once fixed by #1865, we can make these updates too.

m-col:
(The tests have each passed individually during separate runs.)

ramnes:
So do you want to wait on #1865 or merge this now? What about the things you listed in https://github.com/qtile/qtile/pull/1899#discussion_r549893128?

m-col:
Merge this now. Static windows still work as they always have done and aren't made worse by this, so I don't think we need to wait. The things in the comment are all (generally) accounted for. Bars and Gaps get `_configure`-ed again and work as one would expect. I have no had problems with the systray, but think that it ought to be something to keep an eye out for, rather than something that holds this back, as it may not be an issue. In particular the systray tells the X server which screen it belongs to, but I don't know how/where this is used, and it hasn't been an issue for me running a systray anyway. If an issue does arise, then the first place to look would be to see if that information needs to be accurate at all, if not we can just put screen 0 always.

m-col:
I've been thinking about better ways to handle bars that are "removed" than just ignoring them and letting them continue any activity. One option might be to completely tear down bars along with their screen, and then when we are adding screens to read the config again to see if bars need to be (re)created. I think this would be cleaner in that there is no difference in state between having a single screen after removing your second screen, and only starting with a single screen. The main caveat I see with this approach is that the config will be executed during runtime. However, compare that to the current approach to re-configure screens, which is completely restarting Qtile and reading the config anyway.

tych0:
Ping on this, that PR was merged, any chance you can rebase?

m-col:
Alrighty, just rebased and fortunately it still works lol.

> I've been thinking about better ways to handle bars that are "removed" than just ignoring them and letting them continue any activity. One option might be to completely tear down bars along with their screen

I think stopping old bars is a good idea, but not re-reading the config. Perhaps we can stop bars but keep them around (in `qtile.config.screens`) in a state where we can start them up again if needed. Bars themselves are already capable of being `_configure`ed multiple times (this is used when adjusting margin sizes to allow for struts), but we'd have to shut down or pause widgets. Perhaps a `_finalize` followed by a `_configure` later already works.

m-col:
Actually maybe it's fine to just leave bars to continue running. Looking at the `finalize` methods of the widgets, all it would do is finalise the `Drawers`, which accomplishes nothing really. Arguably _that_ is a waste of the computer's time! So I guess widgets just aren't that much of a burden that shutting them down will be an improvement. If you guys agree then this is ready for review and/or merge :)

tcld:
That sounds like the remaining bars and widgets are more of a theoretical problem. Do you know whether they can still count as windows that will then have to be tracked by `core.manager` and thus X? Could they spontaneously show up in unexpected places?

m-col:
As it is, these windows will only be visible if they are within a remaining
screen's dimensions. It might be a good idea to simply hide the window then
actually. But doing that, although they would remain as tracked windows, I don't
expect any problems would arise.

tcld:
It's all edge cases anyway. Probably not worth thinking about overly much until there is a concrete issue that arises because of it.

m-col:
Yep, agreed :)

tych0:
Agreed. Let's see what bugs people complain about :)

PR #1901, opened 2020-09-09
Set up notifications once upon import and implement notification IDs

m-col:
A bool flag is used to track whether notifications are possible (i.e. if
gi cannot be imported), and send_notification will silently do nothing
if that are not. Notification IDs can be passed as a kwarg (id=4), which
the notification service can use to replace existing notifications
with the same ID.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -25,9 +25,20 @@ import sys
 import traceback
 import warnings
 from collections.abc import Sequence
+from random import randint
 from shutil import which
 
 from libqtile.log_utils import logger
+
+_can_notify = False
+try:
+    import gi
+    gi.require_version("Notify", "0.7")  # type: ignore
+    from gi.repository import Notify  # type: ignore
+    Notify.init("Qtile")
+    _can_notify = True
+except ImportError as e:
+    logger.warning("Failed to import dependencies for notifications: %s" % e)
 
 
 class QtileError(Exception):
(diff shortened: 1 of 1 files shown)


libqtile/utils.py
```diff
@@ -25,10 +25,22 @@
...
+_can_notify = False
+try:
+    import gi
+    gi.require_version("Notify", "0.7")  # type: ignore
+    from gi.repository import Notify  # type: ignore
+    Notify.init("Qtile")
+    if Notify.get_server_info()[0]:
+        _can_notify = True
+except Exception:
+    pass
```
ramnes, line 42: Let's log something here?

ramnes:
Although it didn't look great, I think that the previous setup implementation was a great way to ensure we would always send a notification if possible. Why changing this?

m-col:
My motive was to reduce some function calls. Do you mean that the current implementation deals with the edge case where somebody installs `python-gobject` during qtile runtime and then it will work? Because why would we support that here and not in, for example, all the widgets that have external dependencies?

tych0:
For what it's worth, I think it's totally fine to make people restart when they've installed new dependencies. I see little value and lots of code in doing it dynamically.

ramnes:
> Do you mean that the current implementation deals with the edge case where somebody installs `python-gobject` during qtile runtime and then it will work?

No, what worries me is what happens if the notification server is not up and running at the time of import, but shortly afterwards.

m-col:
Oh, I had not considered that. Good point.

I've updated the commit so that availability of a service is tested when sending a notification, so this should not be an issue. The imports can still be done ahead of time. I also added the logging call as you suggested.

libqtile/utils.py
```diff
@@ -219,22 +230,19 @@ def class_proxy(*args, **kwargs):
...
-                notifier.set_urgency(Notify.Urgency.CRITICAL)
-            notifier.show()
-    except Exception as exception:
-        logger.error(exception)
+    if _can_notify and Notify.get_server_info()[0]:
+        notifier = Notify.Notification.new(title, message)
+        notifier.set_timeout(timeout)
+        if urgent:
+            notifier.set_urgency(Notify.Urgency.CRITICAL)
+        notifier.set_timeout(timeout)
```
tcld, line 240: Why are you setting the timeout twice (four lines above it is also set)? If there is a reason, a comment would be important.

libqtile/utils.py
```diff
@@ -219,22 +230,19 @@ def class_proxy(*args, **kwargs):
...
+        if id is None:
+            id = randint(10, 1000)
```
tcld, line 242: I'm probably overthinking this (especially since we didn't have an id before^^), but maybe a uuid or a timestamp would be nicer here?

m-col: This is per the spec (see https://developer.gnome.org/notification-spec/). The Replaces ID is a UINT32.

libqtile/utils.py
m-col: Nope, that is a total typo! Thanks for spotting it.

libqtile/utils.py
tcld: You always have those links at the ready.^^ Maybe add it in a comment?

m-col: Sure, and I've explained how this can be used :)

tcld: That's great stuff, thanks. :)

libqtile/utils.py
```diff
@@ -25,10 +25,21 @@
...
 from libqtile.log_utils import logger
 
+_can_notify = False
+try:
+    import gi
+    gi.require_version("Notify", "0.7")  # type: ignore
+    from gi.repository import Notify  # type: ignore
+    Notify.init("Qtile")
+    _can_notify = True
+except Exception as e:
```
tych0, line 40: Can we use ImportError here if we only expect import errors?

m-col: yep done.

tych0:
lgtm, thanks!

PR #1906, opened 2020-09-13
don't anti-alias box-type graph widgets

m-col:


m-col:
I realised that this might not be obvious as to the why... The "box" graph type is all square shapes, like a bar-chart, with no curves, and so the anti-aliasing makes it look worse rather than being sharply defined edges that align with the pixels anyway.

tych0:
Can you add that to the commit message? lgtm otherwise

PR #1918, opened 2020-09-26
Documentation/small fixes Tile-layout (and other minor things)

tcld:
I am currently reading through the layout code to get new ideas of how to improve #1870 (if I come up with too much of an idea, I'll do that in a separate PR, of course).

While doing so I found a few minor inconsistencies or mistakes; please see the individual commit messages for explanations of the various details.
After understanding the `Tile`-layout in detail, I thought that I might as well add some documentation for it. This is by far the biggest commit here.

I have yet to understand `VerticalTile` in detail, but so far both of them seem awfully similar and different at the same time. In my opinion they should share a common ancestor so all the features are available to both while the only difference is the orientation, but that's a different topic for a different time.

### DIFF
--- a/libqtile/layout/tile.py
+++ b/libqtile/layout/tile.py
@@ -34,35 +34,57 @@ from libqtile.layout.base import _SimpleLayoutBase
 
 
 class Tile(_SimpleLayoutBase):
+    """A layout with two stacks of windows dividing the screen
+
+    The Tile layout divides the screen_rect horizontally into two stacks. The
+    maximum amount of "master" windows can be configured; surplus windows will
+    be displayed in the slave stack on the right.
+    Within their stacks, the windows will be tiled vertically.
+    The windows can be rotated in their entirety by calling up() or down() or,
+    if shift_windows is set to True, individually.
+    """
+
     defaults = [
         ("border_focus", "#0000ff", "Border colour for the focused window."),
         ("border_normal", "#000000", "Border colour for un-focused windows."),
         ("border_width", 1, "Border width."),
         ("name", "tile", "Name of this layout."),
         ("margin", 0, "Margin of the layout"),
+        ("ratio", 0.618,
+            "Width-percentage of screen size reserved for master windows."),
+        ("master_length", 1,
+            "Amount of windows displayed in the master stack. Surplus windows "
+            "will be moved to the slave stack."),
+        ("expand", True,
+            "Expand the master windows to the full screen width if no slaves "
+            "are present."),
+        ("ratio_increment", 0.05,
+            "By which amount to change ratio when cmd_decrease_ratio or "
+            "cmd_increase_ratio are called."),
+        ("add_on_top", True,
+            "Add new clients before all the others, potentially pushing other "
+            "windows into slave stack."),
+        ("add_after_last", False,
+            "Add new clients after all the others. If this is True, it "
+            "overrides add_on_top."),
+        ("shift_windows", False,
+            "Allow to shift windows within the layout. If False, the layout "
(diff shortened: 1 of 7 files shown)


libqtile/layout/tile.py
```diff
@@ -34,35 +34,57 @@
...
+
     defaults = [
         ("border_focus", "#0000ff", "Border colour for the focused window."),
         ("border_normal", "#000000", "Border colour for un-focused windows."),
         ("border_width", 1, "Border width."),
         ("name", "tile", "Name of this layout."),
         ("margin", 0, "Margin of the layout"),
+        ("ratio", 0.618,
+            "Width-percentage of screen size reserved for master windows."),
+        ("masterWindows", 1,  # TODO: masterWindows should be named master_windows
```
tcld, line 55: Would renaming this be ok?

libqtile/layout/tile.py
```diff
@@ -105,43 +126,42 @@ def add(self, client, offset_to_current=0):
...
+            super().add(client, offset_to_current)
         self.reset_master()
 
     def configure(self, client, screen_rect):
+        if client not in self.clients:
+            return
         screen_width = screen_rect.width
         screen_height = screen_rect.height
         border_width = self.border_width
-        if self.clients and client in self.clients:
```
tcld, line 115: As mentioned in the commit message, I don't see why this layout should hide foreign clients, while `Matrix` doesn't. If there is a good reason for hiding them, then I would have to make this change in `Matrix` instead.

libqtile/layout/tile.py
ramnes: It wasn't even documented, so let's rename it and add a line in the change log.

libqtile/layout/tile.py
ramnes: Maybe we should ask @cortesi, as it dates from [the original commit of this layout](https://github.com/qtile/qtile/commit/c45d2976a9742fb70ff2e1883e3a946a05fc78dc#diff-4adc59572ff217731264602873f7084fR114), 10 years ago. :)

I think it's safe to remove.

ramnes (approved):

libqtile/layout/tile.py
tcld: Some of the other layouts check for foreign clients (and hide them), others don't. I guess there are multiple opinions on this.^^

Since the group is the one calling `Layout.layout()`, the group would have to be responsible for feeding the layouts with a wrong client. I don't think that can happen at the moment; and because of that, if it did, an exception might be preferrable over hiding a foreign client.
Maybe I should leave this section untouched?

libqtile/layout/tile.py
tcld: That sounds good.

libqtile/layout/tile.py
ramnes: Yeah, let's keep it that way for now and make it consistent for all layouts in another PR if we think it's worth it.

tcld: Ok, I removed that section.

ramnes: Don't forget to update the commit message. :)

tcld: Oops. Also done.

tcld: I just noticed that there is already a `@property` named `master_windows`. I'll have to come up with a different name then.

tcld: Went for `master_length` since `master_size` seemed ambiguous and I found `master_stack_size` too long.

ramnes:
Looking good, thanks!

PR #1919, opened 2020-09-27
remove already defined `install_requires` from setup.py

DavHau:
The `install_requires` definition inside setup.py is redundant and also differs from the one in `setup.cfg`.


ramnes:
Thanks!

Issue #1470, opened 2019-10-09
Slice layout not work


whatacold:
As @pawamoy noticed in #1423 , the Slice layout just doesn't work. If I add it to layouts in the config like:

    layouts = [
        layout.Max(),
        layout.Stack(num_stacks=2),
        layout.Slice(),
    ]

And test it with xephyr: `./scripts/xephyr -c libqtile/resources/default_config.py`, qtile will freeze, then I have to switch to other consoles to kill it manually to recover it.

ramnes:
Slice without fallback doesn't really make sense. Does it work if you add one?

pawamoy:
Oooooh, I completely missed how slice was working before, so I'm gonna run some tests in my screenshots PR to see if it, in fact, works properly :slightly_smiling_face: 

Other than that, isn't it supposed to use the Max layout as fallback by default?

whatacold:
> Slice without fallback doesn't really make sense. Does it work if you add one?

Nope, it still doesn't work explicitly setting it:

    layout.Slice(fallback=layout.Max()),

But I think it will default to `Max` as in its definition if not setting it, right?:

    class Slice(Delegate):
        """Slice layout
    
        This layout cuts piece of screen and places a single window on that piece,
        and delegates other window placement to other layout
        """
    
        defaults = [
            ("width", 256, "Slice width"),
            ("side", "left", "Side of the slice (left, right, top, bottom)"),
            ("name", "max", "Name of this layout."),
            ("wname", None, "WM_NAME to match"),
            ("wmclass", None, "WM_CLASS to match"),
            ("role", None, "WM_WINDOW_ROLE to match"),
            ("fallback", Max(), "Fallback layout"),
        ]

ramnes:
My bad, I didn't look at the code. Looks like some investigation is needed. :wink:

tcld:
The problem, I think, comes from the mechanism to automatically assign certain windows to the "slice" (i.e. a `Single`[WindowLayout]). Without setting at least one of `wname`, `wmclass` or `role`, the `add`-method for the layout will fail, so at the very least the layout never can contain any windows. It might even mess up the entire group, since the group would expect all its layouts to know the existing windows.

So, to just fix this issue, set up your `Slice` like this: `Slice(wname='Firefox')`

[...]

PR #1925, opened 2020-10-04
Fix Slice.add and make it use Match-object

tcld:
Inspired by #1470, since failing to pass in any matching criteria caused problems. As mentioned in that issue, I would prefer to use a proper `Match`-object here, but I could remove it again if it seemed unsuitable.

Replaces wname, wmclass and role properties with "match" (Match object).
Change a few comments.
Make `add` check whether `match` has been set up. (this is the "fix" for #1470)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@ qtile X.X.X, released XXXX-XX-XX:
     !!! Config breakage !!!
 	- Pacman widget has been removed. Use CheckUpdates instead.
 	- property "masterWindows" of Tile layout renamed to master_length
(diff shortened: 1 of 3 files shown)


tcld:
I think #1805 includes these changes, too. But it has a much different scope.

tych0:
Can you add a doc about API breakage in /CHANGELOG? But otherwise this seems resonable to me.

tcld:
Thank you for the reminder. `CHANGELOG` is now updated, too.

tych0:
Thanks!

Issue #1920, opened 2020-09-30
Dropdown misplacement


tamis-laan:
I have the following scratchpad configuration:

```python
scratchpad = ScratchPad("scratchpad", [
    DropDown("jit", "qutebrowser -T -s statusbar.show never -s tabs.show never meet.jit.si",
                x=0.7, y=0.0, width=0.29, height=0.3, on_focus_lost_hide=False)
])
keys = [
    Key([mod1], 'v', lazy.group['scratchpad'].dropdown_toggle('jit')),
]
```
When pressing `mod+v` the dropdown is misplaced on the left side of the screen instead of the right side. The `x` `y` `width` `height` parameters have no impact on the dropdown position and size. When moving the window it does jump to the right position.

tamis-laan:
For reference: 
```
> qtile --version
0.16.2.dev43+g2f5da5ca
```

m-col:
Just so we can more easily test this, do you see the same behaviour if you use a plain xterm instead of qutebrowser?

tamis-laan:
My other drop downs work and they are terminal based, I use the termite terminal.  However both `qutebrowser` and `jitsi-meet-desktop` which is an electron application have the above stated problem.

m-col:
Right. I tested it with qutebrowser and what I notice is that if I resize and reposition qutebrowser and then close it with Ctrl-Q, the old position and size is retained when it is opened up again. qutebrowser is remembering the user-specified geometry and requesting it to be restored. Normally Qtile respects this for floating windows, so it looks like we are respecting these requests for windows in the scratchpad too, when instead we should be forcing a geometry onto them. Shouldn't be too hard to track this down.

tamis-laan:
I would definitely expect forced geometry in case of dropdowns. This will also make it less likely for the user to accidentally move or change the geometry of the dropdown when using a mouse.

m-col:
Does it get the correct geometry after you have hidden the window and shown it again, after it has been spawned? i.e. is just the very first time it appears where the geometry is wrong?

tamis-laan:
Nope, it keeps its geometry and location. However when I move the window it jumps to the geometry specified in the config.

tamis-laan:
Although I'm not familiar with the `qtile` code base, looking at `scratchpad.py` I can see that when the dropdown is toggled the geometry is set every time before the window is unhidden: 
[...]

Issue #1787, opened 2020-06-19
Hide bar when using Max layout? 


tamis-laan:
When using the max layout to go to full screen I would like to hide the bar. The only way to hide the bar as far as I could find is the cmd_hide_show_bar(). However this only toggles the bar, one would need to explicitly hide or show hooking into the layout_change group_change. Any way to make this work?

tamis-laan:
Managed to hack it together:
```python 
def main(qtile):
    # Show/Hide bar based on layout
    def _bar(qtile):
        # Get the bar 
        bar = qtile.current_screen.bottom
        # Check the layout and hide bar accordingly
        if(qtile.current_layout.info()['name'] == 'max'):
            bar.show(False)
        else:
            bar.show(True)

    @hook.subscribe.layout_change
    def layout_change(layout,group):
        _bar(qtile)

    @hook.subscribe.changegroup
    def group_change():
        _bar(qtile)

    @hook.subscribe.client_focus
    def focus_change(window):
        _bar(qtile)
```
However getting strange behavior where the bar content is hidden but the bar window itself remains.

tamis-laan:
It seams as if:
```python
    @hook.subscribe.changegroup
    def group_change():
        _bar(qtile)
```
is never triggered. Could it be that changegroup hook is bugged?

tych0:
On Sat, Jun 20, 2020 at 01:30:43AM -0700, Tamis wrote:
> It seams as if:
> ```python
>     @hook.subscribe.changegroup
>     def group_change():
>         _bar(qtile)
> ```
> is never triggered. Could it be that changegroup hook is bugged?

Possibly, I think someone mentioned this before.

tamis-laan:
I'm having doubts, seams as if the hook is triggered however the bar is not being drawn causing the bar to behave like a gap. By changing the layout back and forth i.e. `layout_change` the bar is properly drawn again. For me not very easy to debug.

ramnes:
Fixed in #1926

PR #1926, opened 2020-10-05
Fix issues #1920 and #1787

tamis-laan:
Fixed the following issues:
https://github.com/qtile/qtile/issues/1920 
https://github.com/qtile/qtile/issues/1787

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -365,11 +365,9 @@ class Screen(CommandObject):
 
         hook.fire("setgroup")
         hook.fire("focus_change")
-        hook.fire(
-            "layout_change",
-            self.group.layouts[self.group.current_layout],
-            self.group
-        )
+        hook.fire("layout_change",
+                  self.group.layouts[self.group.current_layout],
+                  self.group)
 
     def toggle_group(self, group=None):
         """Switch to the selected group or to the previously active one"""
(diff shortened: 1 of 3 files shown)


libqtile/scratchpad.py
```diff
@@ -184,18 +184,25 @@ def show(self):
...
+            height = int(screen.dheight * self.height)
+            # Set the geometry
+            win.x = x
+            win.y = y
+            win.float_x = x
+            win.float_y = y
+            win.width = width
+            win.height = height
+            # Configure the new geometry
+            win._reconfigure_floating()
```
ramnes, line 204: This line here is the only real change of the whole diff, right? Can we remove all the noise?

ramnes:
Thanks for the PR! Please see my comment on the diff. How is that helping for #1787?

libqtile/scratchpad.py
tamis-laan: Sure the only required part is `win._reconfigure_floating()`, I will quickly revert the rest.

tamis-laan:
> Thanks for the PR! Please see my comment on the diff. How is that helping for #1787?
When you move another group to the current screen then the that group is first drawn and then the hooks are called. 

The problem is that in my hook I'm hiding the bar, however because the group is drawn before hiding the bar there a gap will appear where the bar was. 

By first executing the hooks and then drawing the group this problem is resolved. It does result in some duplicate code which is a bit ugly though.

libqtile/config.py
```diff
@@ -348,6 +348,7 @@ def set_group(self, new_group, save_prev=True):
             g2 = new_group
             s2 = new_group.screen
 
+
```
ramnes, line 351: This is going to break the CI I guess! :)

ramnes:
Thanks!

PR #1927, opened 2020-10-07
Scrolling text widget and improved MPRIS2 widget

Bauthe:
I've been annoyed by how the MPRIS2 widget works. The character-wise text scrolling is ugly, the text cannot be formatted and it can't interact with the player.
I've solved these issues with a new MPRIS2 widget. I added a _ScrollText widget to create it, which can be very useful for other purposes.

This is obviously not an urgent change but I think it's a cool addition.


Bauthe:
The MPRIS2 widget adds a dependency to markupsafe

tych0:
Can you rebase this and fix the CI issues? Also -- is there any reason to have two widgets, or can we integrate the pixel scrolling and other stuff to the main mpris widget?

elParaguayo:
Should we split this out into different requests? Having a new base scrolling widget sounds good.

I'm already updating the MPRIS2 widget to use dbus_next so it may be worth leaving that bit out for now. I can flip it over to the new scroll base if it gets merged.

elParaguayo:
The pixel scrolling on this looks really nice. I do think there's a fair amount of redundant code in here (or I'm just not understanding it properly) so I'll have a play with it.

Bauthe:
Sorry everyone for the late update, I'm having a lot of work lately.
The redundant code was just for backwards compatibilty, as my new widget has a very different signature from the first one.
Of course old code needs to be deprecated. I will rebase and separate both features in different merge requests if it's more convenient, but for now I don't have time.
Count on me before the end of the month.

m-col:
I haven't ran it myself but I am imagining that the pixel scrolling behaviour is super nice and I want it. It could definitely be really cool in other widgets, for instance the notify widget. So +1 on splitting it into one commit to provide that a simple base class we can mix into others flexibly and another for the MPRIS stuff.

Bauthe:
#2039

PR #1930, opened 2020-10-13
Implemented stop qtile hook

tamis-laan:
A hook that get's triggered when stopping qtile.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 qtile X.X.X, released XXXX-XX-XX:
     !!! Config breakage !!!
 	- Pacman widget has been removed. Use CheckUpdates instead.
+	- property "masterWindows" of Tile layout renamed to master_length
 
 qtile 0.16.1, released 2020-08-11:
     !!! Config breakage !!!
--- a/libqtile/layout/tile.py
+++ b/libqtile/layout/tile.py
@@ -34,35 +34,57 @@ from libqtile.layout.base import _SimpleLayoutBase
 
 
 class Tile(_SimpleLayoutBase):
+    """A layout with two stacks of windows dividing the screen
+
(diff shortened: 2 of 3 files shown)


ramnes:
Let's call it "shutdown" and also implement "shutdown_once", as opposed to the existing "startup" and "startup_once". What do you think?

Also, what's your usecase behind this?

tamis-laan:
On IRC there was a need for it, also there is a startup hook so why not a stop or shutdown hook. 

I chose the name stop instead of shutdown because the hook should also be triggered when restarting qtile, not just on shutdown. We can always change the name to shutdown if that is what you prefer.

tych0:
I'm fine with adding this, but I agree we probably need two: one that runs on restart and one that doesn't. The _once mnemonic is what we have for startup, so it seems reasonable to use this nomenclature.

tamis-laan:
What would shutdown_once look like? Is this a case that gets triggered on shutdown but not on restart?

tych0:
shutdown_once would only trigger when not `self._restart is not None`, i.e. when we're not passing `--with-state` and the user has asked qtile to fully shut down, instead of just restarting to reload a new config.

tych0:
I could also be convinced to call it `hook.restart` and `hook.shutdown`

tamis-laan:
hm... yeah I don't know `shutdown_once` sounds rather confusing and isn't very obvious. However `restart` and `shutdown` breaks the already established convention. They are way more descriptive however.

tamis-laan:
I have implemented the startup and shutdown hooks however when I `SIGTERM` qtile the shutdown routine is not called? Not sure how qtile deals with signals???

tych0:
If you can squash all these into one commit, lgtm, thanks!

PR #1933, opened 2020-10-15
Adding a weather widget which uses Open Weather Map services

hschauhan:
API Key is required to be generated before this widget can be used.


tych0:
Can you squash these to one commit? lgtm otherwise

hschauhan:
> Can you squash these to one commit? lgtm otherwise

Sure.

tcld:
I have been in the same situation with the `YahooWeather`-widget: Is there a chance you could provide an `app_key` for the end-user? With Yahoo, I had to file a registration form, estimating how many requests they should be expecting through from qtile. A day later they informed me that qtile had been white-listed, so now we have a public app-key for that widget.

hschauhan:
> I have been in the same situation with the `YahooWeather`-widget: Is there a chance you could provide an `app_key` for the end-user? With Yahoo, I had to file a registration form, estimating how many requests they should be expecting from qtile. A day later they informed me that qtile had been white-listed, so now we have a public app-key for that widget.

I looked at their website. They are mostly paid subscription. For free also, mandatory registration is required? Is it too much for a user to acquire a key and then use it life long?

libqtile/widget/openweather.py
```diff
@@ -0,0 +1,215 @@
...
+            if type(_json) is dict:
+                for key, value in _json.items():
+                    __inner(value, name + key + '_')
+            elif type(_json) is list:
+                for i in range(len(_json)):
+                    __inner(_json[i], name + str(i) + '_')
+            else:
+                out[name[:-1]] = _json
+        __inner(obj)
+        return out
```
ramnes, line 58: Why do we need to flatten?

libqtile/widget/openweather.py
```diff
@@ -0,0 +1,215 @@
...
+# SOFTWARE.
+
+import time
+from urllib.parse import urlencode
+
+from libqtile.widget import base
+from libqtile.widget.generic_poll_text import GenPollUrl
+
+# API access details at https://openweathermap.org/current
+QUERY_URL = "http://api.openweathermap.org/data/2.5/weather?"
```
ramnes, line 29: Looks like it's used only once, so why making it a constant? I'd rather have this as a widget option or hard-coded.

libqtile/widget/openweather.py
```diff
@@ -0,0 +1,215 @@
...
+
+        data['location'] = opwr.get_location_name()
+        data['temp'] = opwr.get_actual_temperature()
+        data['humidity'] = str(opwr.get_humidity()) + "%"
+        data['weather_details'] = opwr.get_weather_details()
+        data['sunrise'] = "\u2600 {}".format(opwr.get_sunrise_time())
+        data['sunset'] = "\u26ed {}".format(opwr.get_sunset_time())
+        data['weather_concise'] = opwr.get_weather_title()
+        data['wind_speed'] = str(opwr.get_wind_speed())
+        data['wind_direction'] = opwr.get_wind_direction()
```
ramnes, line 213: Can we make all of this simpler and assign directly the values from the answer of OWM? I think we would save a lot of bytes overall, feels kinda over-engineered here. :)

e.g.:
```python
data["location"] = response["name"]
```

tych0:
Ping, any thoughts r.e. @ramnes' feedback?

libqtile/widget/openweather.py
hschauhan: Just for an easier look ups. The json output is nested. so "main::temp" becomes "main_temp" in dictionary lookup.

libqtile/widget/openweather.py
hschauhan: Doesn't make sense as an option. Hardcoding directly in function can be done. This is mostly for beautification so that that lines don't go beyond 80/90 char limit. Easier on eyes.

libqtile/widget/openweather.py
hschauhan: There is a separate class to parse the json output. If I have to assign directly, I will have to parse these inline. That would be become messy and unmaintainable.

libqtile/widget/openweather.py
```diff
@@ -0,0 +1,215 @@
...
+        ),
+        ('metric', True, "True to use metric (Default). Imperial if false"),
+        ('language', 'en',
+         """Language of response. List of languages supported can
+         be seen at: https://openweathermap.org/current under
+         Multilingual support""")
+    ]
+
+    def __init__(self, **config):
+        GenPollUrl.__init__(self, **config)
```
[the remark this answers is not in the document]
hschauhan: Yes. Will do so.

tych0:
Ping, looks like htis is pretty close?

tych0:
Ping

hschauhan:
I am sorry. Does anyone want to take this up? Ever since I switched to Ubuntu I can't get qtile working. So I have switched to mate for a while.

tcld:
I want to see this merged, so I'll try to find the time and energy to pick it up. Thanks for your work, I will leave you in as the author. :)

libqtile/widget/openweather.py
tcld: The YahooWeather-widget does this, too. But to be fair, there is no other layer for reading out the values, the flattened json is made available to the user directly.

libqtile/widget/openweather.py
tcld: Again, the YahooWeather-widget does the same thing. I guess it could make sense to have this if a common base class would be extracted or anything like that. An option to the widget might be a good idea, too, since the URL could change in the future.

libqtile/widget/openweather.py
tcld: I don't know which direction to lean towards. The YahooWeather-widget basically just flattens the returned json and then the user can choose the keys he wants to access. This keeps things much more flexible in case the provider changes his format or anything, but of course the access strings might be a bit technical.

PR #1935, opened 2020-10-16
Add a shutdown and restart hook

tamis-laan:
Getting the hang of this, please review.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -87,6 +87,24 @@ class Subscribe:
         """
         return self._subscribe("startup_complete", func)
 
+    def shutdown(self, func):
+        """Called before qtile is shutdown
+
+        **Arguments**
+
+        None
+        """
+        return self._subscribe("shutdown", func)
+
+    def restart(self, func):
+        """Called before qtile is restarted
+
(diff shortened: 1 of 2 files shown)


ramnes:
I think that intuitively I would have expected a "restart" hook to happen after a restart, not before. @tych0 what do you think? If you're good with the current naming, feel free to merge.

tamis-laan:
That does not sound very intuitive to me, usually you want to do some work on startup and then do some work on shutdown/restart. 

If the restart hook was triggered after the restart it would be triggered at the same time as the startup hook!?

m-col:
What about "before_restart"?

libqtile/hook.py
```diff
@@ -87,6 +87,24 @@ def startup_complete(self, func):
         """
         return self._subscribe("startup_complete", func)
 
+    def shutdown(self, func):
+        """Called when qtile is shutdown
```
tych0, line 91: Maybe change "when" to "before"? That clarifies most of the discussion r.e. names, I think.

ramnes:
Alright let's gooooo

Issue #1915, opened 2020-09-19
Match windows that update their class or names themselves after a slight delay


Ayush1325:
Okay, So I am using trying to use Godot in qtile, but for some reason, I can not make only my game window float. The floating works fine if the workspace is in floating mode, but using the client_new hook doesn't work, with either the wm_class or the wm_name. 
I still haven't been able to determine the cause of it so any help is appreciated.

Here is my floating function
```python
@hook.subscribe.client_new
def float_steam(window):
    float_class = [
        ("lxpolkit", "Lxpolkit"),
        ("evelauncher.exe", "steam_app_8500"),
        ("pavucontrol", "Pavucontrol"),
        ("gw2-64.exe", "gw2-64.exe"),
        ("glyphclientapp.exe", "steam_app_743090"),
        ("War Thunder", "War Thunder"),
        ("totono_en.exe", "totono_en.exe"),
        ("launcher.exe", "steam_app_230410"),
        ("saya_en.exe", "saya_en.exe"),
        ("Godot_Engine", "ActionRPG"),                                 # Godot game wm_class
    ]
    float_name = [
        "JetBrains Toolbox",
        "Heartache 101 v2.5",
        "Android Emulator - Pixel_3a:5554",
        "Welcome to Android Studio",
    ]
    wm_class = window.window.get_wm_class()
    w_name = window.window.get_name()
    if (
        wm_class == ("Steam", "Steam")
        and (
            w_name != "Steam"
            # w_name == "Friends List"
            # or w_name == "Screenshot Uploader"
            # or w_name.startswith("Steam - News")
            or "PMaxSize" in window.window.get_wm_normal_hints().get("flags", ())
        )
    ) or (
        w_name in float_name
    ) or (
        wm_class in float_class
    ) or (
        wm_class == ("Places", "firefox") and w_name == "Library"
    ):
        window.floating = True

```

Qtile version: 0.16.1
OS: Arch Linux

m-col:
Are you trying to float it just using those rules in `float_steam`, or do you
also float manually using a keybind (i.e. `lazy.window.toggle_floating`)?

Could you post the output from an `xprop` on the window?

Ayush1325:
> Are you trying to float it just using those rules in `float_steam`, or do you also float manually using a keybind (i.e. `lazy.window.toggle_floating`)? Could you post the output from an `xprop` on the window?

[...]

PR #1937, opened 2020-10-17
Continued: Implement AND-match semantics for Match objects

tcld:
This is (at the moment) just a rebase of PR #1805, since the author stopped working on it:

@cryzed originally wrote:

>Currently values in the float_rules for the Float layout are matched using "any" (OR) semantics, while it is more sensible to match them as "all" (AND). Creating a float rule that requires multiple properties to match at the same time is currently impossible, while recreating the old "any"-behavior is still possible, by simply creating a new rule for each property that used to be in the dict after this change.

>Window.match now uses config.Match instead of only partially reimplementing its logic. This change also allows to match against three new properties: wm_type, wm_instance_class and net_wm_pid in the float rules, and anywhere else that Window.match might be used. The changes should hopefully all be backwards-compatible, except obviously the changed semantics of Window.match.

>I tested this locally with the use-case I mentioned on IRC ({"wname": "win0", "wmclass": "jetbrains-pycharm-ce"}), i.e. matching the PyCharm splash dialog more specifically, which worked. Negating the match by using a bogus value for some property also seemed to work successfully. Otherwise Qtile seems to behave pretty normally while using it, so I am somewhat optimistic that I didn't break things too badly. Please look this over with a critical eye, I'm not very familiar with Qtile's code otherwise.

I have yet to fix some tests and clean up the commit history without changing the original author.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,8 @@ qtile X.X.X, released XXXX-XX-XX:
           were made explicit and added to the default config
         - using `dict`s for `layout.Floating`'s `float_rules` is now deprecated, please
           use `config.Match` objects instead
+        - `no_reposition_match` in `layout.Floating` has been removed; use the list of
+          `config.Match`-objects `no_reposition_rules` instead
     * features
 	- new restart and shutdown hooks
         - rules specified in `layout.Floating`'s `float_rules` are now evaluated with
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,20 +4,20 @@ qtile X.X.X, released XXXX-XX-XX:
 	- property "masterWindows" of Tile layout renamed to master_length
 	- properties wname, wmclass and role of Slice-layout replaced by Match-
 	  type property "match"
-        - rules specified in `layout.Floating`'s `float_rules` are now evaluated with
-          AND-semantics instead of OR-semantics, i.e. if you specify 2 different
-          property rules, both have to match
-        - check the new `float_rules` for `floating_layout` in the default config and
-          extend your own rules appropriately: some non-configurable auto-floating rules
-          were made explicit and added to the default config
-        - using `dict`s for `layout.Floating`'s `float_rules` is now deprecated, please
-          use `config.Match` objects instead
-        - `no_reposition_match` in `layout.Floating` has been removed; use the list of
-          `config.Match`-objects `no_reposition_rules` instead
+	- rules specified in `layout.Floating`'s `float_rules` are now evaluated with
+	  AND-semantics instead of OR-semantics, i.e. if you specify 2 different
+	  property rules, both have to match
+	- check the new `float_rules` for `floating_layout` in the default config and
+	  extend your own rules appropriately: some non-configurable auto-floating rules
+	  were made explicit and added to the default config
+	- using `dict`s for `layout.Floating`'s `float_rules` is now deprecated, please
+	  use `config.Match` objects instead
+	- `no_reposition_match` in `layout.Floating` has been removed; use the list of
+	  `config.Match`-objects `no_reposition_rules` instead
     * features
 	- new restart and shutdown hooks
-        - rules specified in `layout.Floating`'s `float_rules` are now evaluated with
-          AND-semantics, allowing for more complex and specific rules
+	- rules specified in `layout.Floating`'s `float_rules` are now evaluated with
+	  AND-semantics, allowing for more complex and specific rules
 
 qtile 0.16.1, released 2020-08-11:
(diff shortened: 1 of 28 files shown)


libqtile/window.py
```diff
@@ -1053,39 +1054,19 @@ def toscreen(self, index=None):
                 raise CommandError('No such screen: %d' % index)
         self.togroup(screen.group.name)
 
-    def match(self, wname=None, wmclass=None, role=None):
+    def match(self, match):
```
tcld, line 1057: Since this barely does more than `match.compare(win)`, we could maybe deprecate the entire function.

tcld:
Tests are fixed. Commits still have to be cleaned up, I'll probably find some time for that tomorrow.

tcld:
The commit history is cleaned. I also found a few minor bugs I fixed. I kept @cryzed as the author of the major commits (for the record: `git commit --author="name <address@mail.com>" can be used to change the primary author) and only tried to add myself as the co-author (add `Co-authored-by: name <address@mail.com>" after two blank lines in the commit message).
Thanks again to @cryzed for putting in most of the work. :+1: 

It would be nice if somebody could have a look at the code now; my local tests went fine. Working on this even made me aware of another enhancement recently done to the `Match`-class that allows it to check more than one `wm_class`.

It should be noted (again), that this PR will require changes to configuration-files, mostly due to the removal of the list-support of the `Match`-class.

tcld:
One thing I was wondering is whether the standard floating rules should maybe be moved to somewhere else, so they don't _need_ to be copied around so much and can just be referenced when needed. In the test-code I am referring to `default_config.py` for those rules, but in a live environment that can lead to conflicting imports in case there are multiple versions of Qtile available.

Maybe somebody has an opinion on that. Otherwise this is good to go in my opinion - I have been running this branch all day today and not encountered any issues. Actually even a few improvements I didn't expect (I saw some behaviour similar to #1915 before that is now gone).

tcld:
I think a small addition to the `Rule` class should be added (can be done in another PR), since due to the changes to the `Match`-class, rules can't makes use of the OR-semantics anymore and thus one might need a lot of similar rules.

libqtile/window.py
tych0: +1, does the internal API use it at all? I'd say just drop it if not. But not necessary for this PR.

CHANGELOG
```diff
@@ -4,8 +4,20 @@ qtile X.X.X, released XXXX-XX-XX:
 	- property "masterWindows" of Tile layout renamed to master_length
 	- properties wname, wmclass and role of Slice-layout replaced by Match-
 	  type property "match"
+        - rules specified in `layout.Floating`'s `float_rules` are now evaluated with
```
tych0, line 7: Some whitespace issues here.

tych0:
This seems like a nice API consistency improvement, thanks for rebasing it. I agree you can drop match(), if you want to add a patch to this branch for that.

One small whitespace thing, but other than that LGTM.

libqtile/window.py
tcld: I didn't find any places using the function (other than `Window.cmd_match`). I'll do this in another PR, then.

CHANGELOG
tcld: The entire CHANGELOG seems to use a mix of tabs and spaces. I switched to using tabs now.

tcld:
Whitespaces are fixed, `Window.match` will be handled in another PR, since it is really not that pressing.

What do you think about making `Rule`s use multiple `Match`es, as mentioned in my last comment? (That, too, should probably happen in another PR. I'd like to get this merged sooner rather than later. :) )

CHANGELOG
tych0: So it is, we should probably fix that :)

tych0:
I'm not sure I understand your last comment,

> I think a small addition to the Rule class should be added (in another PR), since due to the changes to the Match-class, rules can't make use of the OR-semantics anymore and thus one might need a lot of similar rules.

What small addition?

tcld:
To set up a `config.Rule` at the moment, one needs to pass in a single `Match`-object. Before, those `Match`-objects could be set up to match multiple clients (OR), now they can only match one client. Therefore I think that the constructor of the `config.Rule`-class should allow to pass in multiple `Match`-objects at once, otherwise rules might have to be duplicated.

tych0:
Yep, sounds reasonable to me. It may also be nice to add an example of how to use Rule/Match to the docs to get various semantics (a contrived example would be fine). Thanks for working on this!

cryzed:
Agreed, thanks again @tcld!

PR #1938, opened 2020-10-19
Add support for Python 3.9, drop 3.5 and 3.6

tcld:
Support for Python 3.5 had actually been dropped before, but the tests were still available.

I also found a doc string for funtoo which mentioned some Python versions.

@ramnes wants to add more stuff to this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,21 @@ jobs:
           env: TOXENV=py37
         - python: 3.8
           env: TOXENV=py38
+        - python: 3.9
+          env: TOXENV=py39
         - python: pypy3
           env: TOXENV=pypy3
         - python: nightly
           env: TOXENV=py-nightly
-        - python: 3.8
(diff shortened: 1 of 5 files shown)


tych0:
Can we hold off on this until travis actually supports 3.9? No point in adding a test that we expect to always fail, IMO.

tcld:
This isn't just about adding tests for Python 3.9, more importantly the support for 3.5 and 3.6 would be dropped.

@ramnes wanted to add in some things anyway after the discussion on IRC a few days ago. Maybe Travis will have been updated by the time that's done.

ramnes:
https://travis-ci.community/t/python-3-9-0-build/10091/7

tych0:
Ok. Let's just add a patch for dropping 3.6 if that's the goal. 3.9 can go in when travis has support for it.

tcld:
I removed the Python 3.9 test again.

Issue #1773, opened 2020-06-08
idea: qtile-config {check,migrate}


tych0:
For a long time we've had issues with configs:

1. people have syntax/logic errors and have to restart a bunch
2. we want to make changes to the internals and people are using them
3. people use the wrong kind of values for config variables, but it works "well enough", except doesn't get the desired behavior
4. etc.

In particular, we've long struggled with number 2, and have no real conclusions about what to support for a 1.0 API. For example, we (rightly so) got rid of a bunch of camel case in our codebase, but it broke everyone's configs who was messing with the internals. I've argued that everything that isn't a `cmd_` is not a public API and we reserve the right to change, but it's still annoying for people to have to update their configs.

Further, r.e. numbers 1 & 3, we've often told people "just run your config via python3 config.py" or (when we get type annotations working) "run pytype on your config". But there are additional checks we could/should have people do (see e.g. #1772) that neither of those will accomplish.

So, what if we introduce a new command `qtile-config`. Initially I'm imagining two subcommands, `check` and `migrate`. `check` would run a type checker, our checks from #1772, and anything else we want (deprecation warnings, etc.). `migrate` would do a best-effort migration of code. In general re-writing code is undecidable, but most of our changes are things like "add another argument to this function" or "rename fooBar to foo_bar", which we could do with reasonable accuracy, and if your config doesn't get auto-migrated, well, too bad.

Then, after 1.0, we allow any API changes where someone can write a migrate blurb that would be reasonably successful at migrating most configs. This means that 1. users don't have do this themselves and 2. we can keep refactoring our code for e.g. wayland.

Once we had this code, we could do other things too, like run `migrate` on qtile start so users' configs are automatically upgraded.

There is an issue of dependencies, but we can make the type checking deps for `qtile-config check` optional, so people don't have to install them if they don't want, and just print a big warning if they're not available.

I'm happy to work on an implementation of this, but I'd just like to see if anyone else has any good ideas before starting :)

ramnes:
I really like the idea! 

We said a few times that we have too many entrypoints, so maybe we should merge all these into qtile beforehand? Something like

```
qtile
qtile help
qtile shell
qtile check
qtile migrate
qtile top
...
```

Also, we could have an automigrate flag in config so that users can optin for us to migrate their config automatically when we detect that qtile version has changed (with the version written in a cache dir or something).

I'll try to think a bit more about this and report back if I have anything else. :)

zordsdavini:
[...]

PR #1939, opened 2020-10-19
Move to subcommand

tych0:
We'd like to move to a `qtile <subcommand>` type command architecture instead of just dumping tons of stuff in /usr/bin. Here's a start on that.

The motivation for this for me was really https://github.com/qtile/qtile/issues/1773


tych0:
FWIW, I think this no longer a WIP and is ready to go.

libqtile/scripts/main.py
```diff
@@ -0,0 +1,35 @@
...
+    start.add_subcommand(subparsers)
+    shell.add_subcommand(subparsers)
+    top.add_subcommand(subparsers)
+    run_cmd.add_subcommand(subparsers)
+    cmd_obj.add_subcommand(subparsers)
+    options = parser.parse_args()
+    if not hasattr(options, "func"):
+        parser.print_help()
+        print("running qtile has moved to subcommand 'qtile start'")
+        os.exit(1)
```
m-col, line 34: `AttributeError: module 'os' has no attribute 'exit'`. Did you mean sys.exit?

tych0: I sure did; now why didn't our CI catch this?

m-col:
The most obvious problem we'll face is people updating their package, not updating their .xinitrc or similar (i guess it's fine if they use a login manager that uses .desktop files), and not being sure as to what is going on. Is it reasonable to default `qtile` to `qtile start`?

tych0:
> Is it reasonable to default qtile to qtile start?

That's a great question -- generally I feel like if you run a program without arguments, it shouldn't do anything. What if I print a big deprecated warning so we can make it not do something later, but default it to `qtile start` for now?

m-col:
Yep I agree that is the cleanest way to go.

m-col:
FWIW I just played around with all of the commands and they all seem to be working fine.

m-col:
Maybe also update the xephyr script?

tych0:
Yeah, and the .desktop file too.

I spent a few minutes playing around with argparse and a default subcommand last night and couldn't get it to work :(

tych0:
Ok, I think this has all of the updates we talked about. @ramnes @m-col

m-col:
Running `qtile start` inside my running session stills starts the new instance :(

tych0:
Hrm, do you have all the x properties? It worked for me.

tych0:
Also - it works in the tests :)

m-col:
Would it be possible to print the help message when `qtile help` is called?

Other than that, all else seems good :)

tych0:
Ok - did you get it sorted out? It is totally possible there's bugs here somewhere.

m-col:
Ah yes sorry, forgot to say, I discovered the cause. You've moved the `sys.path`
lines to after the import of libqtile. This makes running `./bin/qtile` from the
git repository import the installed version, rather than the local git copy.
This was the cause of my error, and is problemetic for when testing things
within the git copy, so the `sys.path` prepend needs to be moved up to before
the import. This issue won't affect the installed `qtile` executable though.

tych0:
Ah ha, that's totally unintentional. Let me revert it. Thanks :)

PR #1941, opened 2020-10-22
Pass qtile object to mouse callbacks

hschauhan:
The mouse callbacks need qtile object to do anything
worth while. This seems to work in current release.
But 1.6 seems to throw exceptions.

Signed-off-by: Himanshu Chauhan <hschauhan@nulltrace.org>

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -229,7 +229,7 @@ class _Widget(CommandObject, configurable.Configurable):
     def button_press(self, x, y, button):
         name = 'Button{0}'.format(button)
         if name in self.mouse_callbacks:
-            self.mouse_callbacks[name]()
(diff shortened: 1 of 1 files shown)


tych0:
I think we can just use from libqtile import qtile now thanks to some
patches from @m-col.

m-col:
For posterity: what you said on IRC is correct, the qtile object did used to be passed to callback functions. However, because most widgets manually implemented their own callbacks prior to commit 55f92b22b346ccd8a4c3e1867080bc82dbcaf33b, only a small number of widgets actually used the base class's mouse_callbacks functionality, which passed `qtile`. That commit made all widgets conform to the same implementation (going by majority, _not_ passing `qtile`) to make it more consistent and predictable (and as tych0 states, it can be imported anyway).

Issue #1913, opened 2020-09-17
Floating windows spawn on screen 1


tcld:
# Issue description

I have two screens attached to my machine. When opening certain floating windows, they appear on the wrong screen. Specifically imagine the following situation:

Group A is active on screen 1, group B is active on screen 2. Group B/screen 2 has focus. I launch a program and it shows up on screen 1, but it is managed by group B. If there are any windows in group A, it is then not possible to move the floating window around since the windows in group A grab the focus instead.

From what I have seen, this only applies to windows defined as floating through the `floating_layout` variable (windows put into floating mode manually don't show this problem). After digging in the code a bit it looked as if the screen offset wasn't added to the `float_x` and `float_y`-variables (like it is done in window.place()), but I didn't see an easy way to fix or even confirm that.

# Qtile version

https://github.com/qtile/qtile/commit/f7c212db2abce06f78c23f3aaa4e2fa03affcdb1

# Configuration

Just my floating configuration here. If anybody thinks that another section might be helpful, I could add it.

```python
    floating_layout = layout.Floating(
        float_rules=[
            # Run the utility of `xprop` to see the wm class and name of an X client.
            {'wmclass': 'confirm'},
            {'wmclass': 'dialog'},
            {'wmclass': 'download'},
            {'wmclass': 'error'},
            {'wmclass': 'file_progress'},
            {'wmclass': 'notification'},
            {'wmclass': 'splash'},
            {'wmclass': 'toolbar'},
            {'wmclass': 'confirmreset'},  # gitk
            {'wmclass': 'makebranch'},  # gitk
            {'wmclass': 'maketag'},  # gitk
            {'wname': 'branchdialog'},  # gitk
            {'wname': 'pinentry'},  # GPG key password entry
            {'wmclass': 'ssh-askpass'},  # ssh-askpass
            {'wmclass': 'lxpolkit'}, # credential management by authentication agent
            {'wmclass': 'Xephyr'},  #window manager testing w/ Xephyr
            {'wname': 'Compress'}, # compression extension for file browsers
            {'wmclass': 'veracrypt'},
        ],
        no_reposition_match=Match(wm_class=['Gsimplecal']),  # do not reposition windows to center of screen (leave it to them instead)
        **layout_theme
    )
```

tcld:
I might have found the cause for this:
[...]

PR #1943, opened 2020-10-25
Floating layout lays out windows with USPosition/PPosition

### LINKED ISSUE #1893: Popups and child-windows are drawn behind parent windows in certain situations
# Issue description
In some situations I noticed that popups or child-windows are drawn (or moved) behind their respective parents; for child-windows this can be quite a problem in case the child is blocking, since then the software might be completely unresponsive.

Places where I have noticed this and that I remember:
1. Virtualbox (fullscreen) (the main window, not a client): Context menus on VM-list open in the background.
2. Veracrypt (floating) "mount" child-window: The window either spawns in the back or is moved there once it loses focus.

There have been some other cases, but due to testing PR #1870, which I initially expected to be the cause for this, I haven't taken note of all of them.
Some software definitely has working popup-windows, so Virtualbox might be doing something weird.

# Qtile version

https://github.com/qtile/qtile/commit/c6b80e9444e9b4e5dc2f52ca1954dc502c45f2eb
(newer versions not tested yet)

# Configuration

I'll only include one snippet here, since verything else (setting up Keys and Widgets) most likely has nothing to do with the problem.

```python
@hook.subscribe.client_new
def modify_window(client):
    if (client.window.get_wm_transient_for() or client.window.get_wm_type() in floating_types):
        client.floating = True
```

On top of this I am currently using fake screens instead of normal screens.

It would be nice if somebody could confirm this, since I am not entirely sure this isn't all caused by my own PR.

Related to #1870, which could fix at least the described issue with child-windows.

tcld:
Most floating windows (that were not manually put into floating mode)
have either USPosition or PPosition already set (sometimes both).
Therefore we cannot reliably check these flags to see whether the
floating layout should lay out a window.

Most naturally floating windows already come with a valid position,
which will be respected. The floating layout will lay out the windows
anyway. Windows with invalid coordinates are repositioned first.

Fixes #1893 and #1913.

Every software I have tested worked great, most importantly the above issue have been fixed. It is possible, though, that some other pieces of software might still show weird behaviour.

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -179,35 +179,53 @@ class Floating(Layout):
     def blur(self):
         self.focused = None
 
+    def on_screen(self, client, screen_rect):
+        if client.x < screen_rect.x:  # client's left edge
+            return False
+        if screen_rect.x + screen_rect.width < client.x + client.width:  # right
+            return False
+        if client.y < screen_rect.y:  # top
+            return False
+        if screen_rect.y + screen_rect.width < client.y + client.height:  # bottom
+            return False
+        return True
+
     def compute_client_position(self, client, screen_rect):
         """ recompute client.x and client.y, returning whether or not to place
         this client above other windows or not """
-        above = False
-        transient_for = client.window.get_wm_transient_for()
-        win = client.group.qtile.windows_map.get(transient_for)
-        if win is not None:
-            # if transient for a window, place in the center of the window
-            center_x = win.x + win.width / 2
-            center_y = win.y + win.height / 2
-        else:
-            center_x = screen_rect.x + screen_rect.width / 2
-            center_y = screen_rect.y + screen_rect.height / 2
-            above = True
+        above = True
 
-        x = center_x - client.width / 2
-        y = center_y - client.height / 2
+        if client.has_user_set_position() and not self.on_screen(client, screen_rect):
+            # move to screen
+            client.x = screen_rect.x + client.x
+            client.y = screen_rect.y + client.y
+        if not client.has_user_set_position() or not self.on_screen(client, screen_rect):
+            # client has not been properly placed before or it is off screen
+            transient_for = client.window.get_wm_transient_for()
+            win = client.group.qtile.windows_map.get(transient_for)
+            if win is not None:
+                # if transient for a window, place in the center of the window
+                center_x = win.x + win.width / 2
+                center_y = win.y + win.height / 2
+                above = False
+            else:
+                center_x = screen_rect.x + screen_rect.width / 2
+                center_y = screen_rect.y + screen_rect.height / 2
 
-        # don't go off the right...
-        x = min(x, screen_rect.x + screen_rect.width)
-        # or left...
-        x = max(x, screen_rect.x)
-        # or bottom...
-        y = min(y, screen_rect.y + screen_rect.height)
-        # or top
-        y = max(y, screen_rect.y)
+            x = center_x - client.width / 2
+            y = center_y - client.height / 2
 
-        client.x = int(round(x))
-        client.y = int(round(y))
+            # don't go off the right...
+            x = min(x, screen_rect.x + screen_rect.width)
+            # or left...
+            x = max(x, screen_rect.x)
+            # or bottom...
+            y = min(y, screen_rect.y + screen_rect.height)
+            # or top
+            y = max(y, screen_rect.y)
+
+            client.x = int(round(x))
+            client.y = int(round(y))
         return above
 
     def configure(self, client, screen_rect):
@@ -231,14 +249,7 @@ class Floating(Layout):
             client.paint_borders(bc, bw)
             client.cmd_bring_to_front()
 
-        # similar to above but the X11 version, the client may have already
-        # placed itself. let's respect that
-        elif client.has_user_set_position():
-            client.paint_borders(bc, bw)
-            client.cmd_bring_to_front()
-
-        # ok, it's not java and the window itself didn't position it, but users
-        # may still have asked us not to mess with it
+        # alternatively, users may have asked us explicitly to leave the client alone
         elif any(m.compare(client) for m in self.no_reposition_rules):
             client.paint_borders(bc, bw)
             client.cmd_bring_to_front()
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -216,11 +216,11 @@ class Floating(Layout):
             y = center_y - client.height / 2
 
             # don't go off the right...
-            x = min(x, screen_rect.x + screen_rect.width)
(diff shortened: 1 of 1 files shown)


libqtile/layout/floating.py
```diff
@@ -181,31 +181,35 @@ def compute_client_position(self, client, screen_rect):
...
+            else:
+                center_x = screen_rect.x + screen_rect.width / 2
+                center_y = screen_rect.y + screen_rect.height / 2
+                above = True
+
+            x = center_x - client.width / 2
+            y = center_y - client.height / 2
+
+            # don't go off the right...
+            x = min(x, screen_rect.x + screen_rect.width - client.width)
```
tcld, line 203: I made a small change here, since it seems more sensible to have the entire window on screen than just requiring it to be adjacent to it.

tych0:
Huh? We don't want to just delete this code... perhaps we should add logic to figure out the right screen. But ignoring USPosition is not correct either.

tcld:
The function in the `Window`-class isn't used by anything else. Or are you talking about the lines I removed from the floating layout?

Also, this is not just about the screen. The other issue about child-windows and popups seems to be solved by this, too (I haven't performed a proper test comparing the master branch and just this PR yet).

tych0:
> Or are you talking about the lines I removed from the floating layout?

Yes :). We should not ignore USPosition.

tcld:
Well, this PR doesn't entirely ignore the flags - there are two differences from before:
1. Instead of checking whether the flags are set, _the coordinates themselves_ are checked. If they don't make sense, they are overwritten. Otherwise we trust the software that set them. This fixes #1913. 
2. For every window with USPosition and/or PPosition set, `_Window.place` is called instead of `_Window.user_placed_window_setup`. This fixes #1893. 

There is probably an approach where we could still check the flags - but the only thing I could see be done differently, is how the windows _without_ either of the flags set would be handled (i.e. always instead of just when the given coordinates are weird). So instead of [the checks here](https://github.com/qtile/qtile/pull/1943/files#diff-475097a9172bb7cb3cc2bc9fe378a383acf1231b6a6bc1ec048e764af439eddcR186) we could do something like this:
```python
if not client.has_user_set_position:
```

What do you think?

tych0:
> Instead of checking whether the flags are set, the coordinates themselves are checked. If they don't make sense, they are overwritten.

Right, I think this is incorrect. If a window specified its position, we should do our best to put it there.

What I'm suggesting is that the coordinates are perhaps relative to the screen the window is on, and the bug is that we're assuming they're global. But I don't think overwriting the coordinates is correct.

tcld:
That's a good idea, it seems that some software provides "global" coordinates, while other software provides local coordinates - that would explain some of the problems I noticed. So overwriting the nonsensical coordinates would be correct after all, it's just the way it is done that should be changed. I'll add another commit tomorrow.

tcld:
I reintroduced the `has_user_set_position`-method. If it is `True` *and* the client is on screen, we don't mess with its coordinates. In case it is off screen, the coordinates need to be changed, however.

Either way, the layout will actuall call `place()` on the client, but maybe without changing the coordinates first.

@tych0 : Is this more what you imagined?

libqtile/layout/floating.py
```diff
@@ -231,12 +249,6 @@ def configure(self, client, screen_rect):
             client.user_placed_window_setup(bc, bw)
             return
 
-        # similar to above but the X11 version, the client may have already
-        # placed itself. let's respect that
-        if client.has_user_set_position():
-            client.user_placed_window_setup(bc, bw)
```
tych0, line 237: Is there a way we can avoid a place() call if has_user_set_position() is true and we don't have to change the coordinates? It does avoid some flicker.

tcld: Not calling `_Window.place()` seems to be what is causing #1893. I guess that means that the call to `_Window.send_configure_notify`is the important part? Other than that `place` does what `user_placed_window_setup` does, as far as I can see. I'll try to take a closer look.

@m-col : Your PR #1887 touches a lot of this code, too. I'll try to see whether that, too, solves #1893.

tcld: It seems that #1887 also solves #1893. It doesn't adress the cases where floating windows spawn on the wrong screen, however. So I think #1887 would be preferrable over this PR, I could then later try to fix the problem with spawning on the wrong screen separately.

tcld: I put some time into investigating why calling `place` for windows having `USPosition` and/or `PPosition` solves having invisible popups/children (#1893): The critical call needed is that to `xcbq.Window.configure(x, y, width, height)`.

This seems very confusing to me, since the window should already be positioned and visible without that call, right? For the cases described in #1893, however, only _every other_ instance appears visible. There might be an underlying bug in how we grab and manage windows, but I don't think I can find it on my own.

Here is the change that would fix #1893 - maybe somebody else knows why this works and what underlying issue it might be pointing to? (@m-col , @ramnes ?)
```python
# inside layout.floating.Floating.configure
if client.has_user_set_position():
    client.user_placed_window_setup(bc, bw)
    client.window.configure(x=client.x, y=client.y, width=client.width, height=client.height)
```

Any help would be very much appreciated!

tcld:
Rebased to current master. Also moved the `on_screen`-method so it can be called externally (could be useful, since the recent changes to the screen change-hook can leave floating windows off screen in some edge cases).

The intended fix with this PR is #1913, #1893 is just an accident.

libqtile/layout/floating.py
tych0: Can you add this reasoning to the commit message itself?

libqtile/layout/floating.py
```diff
@@ -231,12 +249,6 @@ def configure(self, client, screen_rect):
             client.paint_borders(bc, bw)
             client.cmd_bring_to_front()
 
-        # similar to above but the X11 version, the client may have already
-        # placed itself. let's respect that
-        elif client.has_user_set_position():
-            client.paint_borders(bc, bw)
-            client.cmd_bring_to_front()
-
         # ok, it's not java and the window itself didn't position it, but users
```
tych0, line 252: This comment is now wrong, IIUC

libqtile/layout/floating.py
tcld: OK, I reworded the commit message.

libqtile/layout/floating.py
tcld: You are correct, thanks for spotting that. I changed the comment, too.

tych0:
Great, thanks for working on this!

tcld:
I mixed up the issue numbers: This definitely fixes #1913 and works around #1893, not the other way around. But since both issues should be gone, we can get away with closing both for now. I went ahead and closed #1913.

PR #1949, opened 2020-11-06
WindowList should not include ScratchPad-windows

tcld:
I noticed that when switching windows via `WindowList`, possibly existing `ScratchPad`-windows are included in the shown list as well. When selected, the corresponding `ScratchPad`-group is opened up with all the usual clients layed out at once. Interaction now happens in the same way as with normal groups, not `ScratchPad`-windows.

I am not entirely sure whether this is an actual bug or a feature, but since I found the experience very much confusing (it took me a minute to realize that Qtile was indeed still running normally), I wanted to get rid of those scratchpad-windows. There doesn't seem to be much point in having them in a quick-switch list anyway.

### DIFF
--- a/libqtile/extension/window_list.py
+++ b/libqtile/extension/window_list.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 
 from libqtile.extension.dmenu import Dmenu
+from libqtile.scratchpad import ScratchPad
 
 
 class WindowList(Dmenu):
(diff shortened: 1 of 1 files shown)


tych0:
Sounds reasonable to me, @m-col may have thoughts?

m-col:
I suppose the "bug" here is that WindowList should try to move scratchpad windows to the current group when selected, rather than trying to trying to move to that group, which should be hidden. It does sound reasonable to exclude those windows from the quick switcher though so that also solves the problem. Unless someone disagrees to that, LGTM.

PR #1951, opened 2020-11-06
Allow Match-list to be passed into config.Rule

tcld:
As mentioned [here](https://github.com/qtile/qtile/pull/1937#issuecomment-719271950), `config.Rule` needed a little change to make better use of the recent changes to `config.Match`.

@tych0 : You also mentioned that a bit of documentation would be nice. As far as I can see, `config.Rule` is exclusively used in conjunction with `DGroups`, which aren't documented at all right now (as far as I can see). I haven't played with adding documentation outside of what's usually written in a classes header, and I haven't used `DGroups` so far (although that's partially due to the lacking documentation).
I could extend the documentation for `config.Match` a little so it includes an example or two, but that doesn't really fit this PR, I think. Also, it isn't too difficult to figure out how `config.Match` works - especially since it is actually used in various places.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -644,15 +644,15 @@ class Match:
 
 
 class Rule:
-    """How to act on a Match
+    """How to act on a match
 
-    A Rule contains a Match object, and a specification about what to do when
-    that object is matched.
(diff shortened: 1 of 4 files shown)


tych0:
Great, thanks for this!

Issue #1867, opened 2020-08-16
Mapping XF86CD disables all input to windows


tcld:
#  Issue description

I have connected a Bluetooth-device to my machine and it comes with a play/pause-button I wanted to make use of.

The following is in my config.py right now:

```python
Key([], 'XF86AudioPlay', lazy.spawn('playerctl play-pause'), desc='Play/pause playback.'),
# Key([], 'XF86CD', lazy.spawn('playerctl play-pause'), desc='Play/pause playback.'),
```
The first of those lines refers to a button on my keyboard, which does work perfectly fine. As soon as I remove the comment from the second line (and restart), however, any input done by my keyboard is not passed to the applications I am running anymore (if there is a cursor visible, it might blink when I hit a key, though). The only thing that the keyboard can still be used for is to enter my Qtile-shortcuts.
For this problem to occur I don't even have to connect the Bluetooth-device or press the button, there just appears to be something wrong with binding that particular key.

# Qtile version

0.15.1

# Stacktrace

A bit of the log (looks like something hangs in an infinite loop since the output adds up very quickly):

```log
2020-08-16 12:43:00,902 ERROR libqtile xcore.py:_xpoll():L277 Got an exception in poll loop
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/libqtile/backend/x11/xcore.py", line 246, in _xpoll
    ret = target(event)
  File "/usr/lib/python3.8/site-packages/libqtile/backend/x11/xcore.py", line 494, in handle_KeyPress
    self.qtile.process_key_event(keysym, event.state & self._valid_mask)
  File "/usr/lib/python3.8/site-packages/libqtile/core/manager.py", line 329, in process_key_event
    key = self.keys_map[(keysym, mask)]
KeyError: (65515, 0)
2020-08-16 12:43:00,961 ERROR libqtile xcore.py:_xpoll():L277 Got an exception in poll loop
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/libqtile/backend/x11/xcore.py", line 246, in _xpoll
    ret = target(event)
  File "/usr/lib/python3.8/site-packages/libqtile/backend/x11/xcore.py", line 494, in handle_KeyPress
    self.qtile.process_key_event(keysym, event.state & self._valid_mask)
  File "/usr/lib/python3.8/site-packages/libqtile/core/manager.py", line 329, in process_key_event
    key = self.keys_map[(keysym, mask)]
KeyError: (65513, 64)
[...]

PR #1952, opened 2020-11-06
Allow keysym to be mapped to multiple keycodes

tcld:
Each keycode can be mapped to multiple keysyms by X11, and qtile could
already handle the lookup in that direction. However, looking up a
keysym only ever returned the first keycode it had been mapped to,
which made it difficult to map buttons of e.g. different devices to the
same function.

This is related to #1867. The main effect of this PR is that one can now make use of buttons with different keycodes/same keysyms to trigger stuff in qtile - in particular I can now use a bluetooth-device to play/pause media, while my keyboard has a similar button that does the same.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -809,7 +809,7 @@ class Connection:
         self.atoms = AtomCache(self)
 
         self.code_to_syms = {}
-        self.first_sym_to_code = None
+        self.sym_to_codes = None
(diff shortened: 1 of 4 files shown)


tych0:
Looks like the linter is angry, but the patch itself looks good to me.

tcld:
I made a small change to pass the tests, but it seems starting them takes a bit.

tych0:
Yeah, travis has been super backed up lately :(

PR #1955, opened 2020-11-07
Refresh update count after execute

elParaguayo:
Based on discussion in qtile-dev group, this pull request allows an update of the update count immediately after updates have run.

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -36,7 +36,8 @@ class CheckUpdates(base.ThreadedPollText):
         ("display_format", "Updates: {updates}", "Display format if updates available"),
         ("colour_no_updates", "ffffff", "Colour when there's no updates."),
         ("colour_have_updates", "ffffff", "Colour when there are updates."),
-        ("restart_indicator", "", "Indicator to represent reboot is required. (Ubuntu only)")
+        ("restart_indicator", "", "Indicator to represent reboot is required. (Ubuntu only)"),
+        ("refresh_after_execute", True, "Immediately refresh the widget after executing updates.")
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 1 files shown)


libqtile/widget/check_updates.py
```diff
@@ -104,4 +105,19 @@ def button_press(self, x, y, button):
...
-            Popen(self.execute, shell=True)
+            self.do_execute()
+
+    def do_execute(self):
+        self._process = Popen(self.execute, shell=True)
+
+        if self.refresh_after_execute:
+            self.timeout_add(1, self.check_updates)
+
+    def check_updates(self):
```
elParaguayo, line 116: Just realised that this method is not brilliantly named. Doesn't clash but is too close to existing code.

tcld:
Looks more well-rounded than my approach, although I assume you have based your code on version 0.16.1? The code looks [a bit different now](https://github.com/qtile/qtile/blob/master/libqtile/widget/check_updates.py).

elParaguayo:
Yes. Based on 0.16.1 because that's what I'm running and wanted to test it.

If it's changed then we can ignore this and I'll redo whenever 0.16.2 comes
out.

tych0:
This conflicts. Can you send a PR against master?

libqtile/widget/check_updates.py
```diff
@@ -36,7 +36,8 @@ class CheckUpdates(base.ThreadedPollText):
         ("display_format", "Updates: {updates}", "Display format if updates available"),
         ("colour_no_updates", "ffffff", "Colour when there's no updates."),
         ("colour_have_updates", "ffffff", "Colour when there are updates."),
-        ("restart_indicator", "", "Indicator to represent reboot is required. (Ubuntu only)")
+        ("restart_indicator", "", "Indicator to represent reboot is required. (Ubuntu only)"),
+        ("refresh_after_execute", True, "Immediately refresh the widget after executing updates.")
```
tcld, line 40: I thought about this for a bit. And do we *really* expect somebody to set `refresh_after_execute` to `False` because he doesn't want to see the widget update right after a system update? I can't see a realistic use-case for it.

@tych0 : What do you say? You seem to be in favor of adding options most of the time, so you probably have a better perspective. :)

elParaguayo:
I've now got qtile in a virtual environment so I'm happy to rework against
the current master.

elParaguayo:
Updated request with changes to current master: https://github.com/qtile/qtile/pull/1957

Closing this one.

libqtile/widget/check_updates.py
tych0: I agree that there's no reason to add an option here.

PR #1963, opened 2020-11-14
Make detection of __xcb_proto_version__ more robust.

### LINKED ISSUE #1962: configure() in libqtile.backend.x11.xcbq is not robust
# Issue description

The function is not robust, as it breaks on xcffib declaring a version with patch level:
https://github.com/qtile/qtile/blob/464bf7f191e800bcb8a075d465841cde81bc15f8/libqtile/backend/x11/xcbq.py#L619-L628

# Qtile version

0.16.1

# Stack traces

Testing breaks on this:

```
>           raise AssertionError("Error launching Qtile, traceback:\n%s" % error)
E           AssertionError: Error launching Qtile, traceback:
E           Traceback (most recent call last):
E             File "/build/qtile/src/qtile-0.16.1/test/conftest.py", line 286, in run_qtile
E               q = SessionManager(kore, config_class(), fname=self.sockfile)
E             File "/build/qtile/src/qtile-0.16.1/libqtile/core/session_manager.py", line 31, in __init__
E               self.qtile = Qtile(kore, config, self.eventloop, no_spawn=no_spawn, state=state)
E             File "/build/qtile/src/qtile-0.16.1/libqtile/core/manager.py", line 161, in __init__
E               self._process_screens()
E             File "/build/qtile/src/qtile-0.16.1/libqtile/core/manager.py", line 352, in _process_screens
E               scr._configure(
E             File "/build/qtile/src/qtile-0.16.1/libqtile/config.py", line 289, in _configure
E               i._configure(qtile, self)
E             File "/build/qtile/src/qtile-0.16.1/libqtile/bar.py", line 208, in _configure
E               self.window = window.Internal.create(
E             File "/build/qtile/src/qtile-0.16.1/libqtile/window.py", line 688, in create
E               i.place(x, y, width, height, 0, None)
E             File "/build/qtile/src/qtile-0.16.1/libqtile/window.py", line 492, in place
E               self.window.configure(**kwarg)
E             File "/build/qtile/src/qtile-0.16.1/libqtile/backend/x11/xcbq.py", line 623, in configure
E               if float(xcffib.__xcb_proto_version__) < 1.12:
E           ValueError: could not convert string to float: '1.14.1'
```
# Configuration

n/a

dvzrv:
libqtile/backend/x11/xcbq.py:
Only use major and minor version of `xcffib.__xcb_proto_version__` for
comparison in `Window.configure()`.
Previously the comparison would raise `ValueError` on a version string
also including a patch level.

Fixes #1962

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -623,7 +623,7 @@ class Window:
         mask, values = ConfigureMasks(**kwargs)
         # older versions of xcb pack everything into unsigned ints "=I"
         # since 1.12, uses switches to pack things sensibly
-        if float(xcffib.__xcb_proto_version__) < 1.12:
+        if float(".".join(xcffib.__xcb_proto_version__.split(".")[0: 2])) < 1.12:
             values = [i & 0xffffffff for i in values]
         return self.conn.conn.core.ConfigureWindow(self.wid, mask, values)
 


tych0:
I don't think this is quite right: the check is really about the xcb-proto version, not the xcffib version (0.12 doesn't exist). Can we just fix the float cast to be a real version parsing function?

Also, where did you get xcb-proto of 1.14.1? I only see 1.14 released on their web page.

dvzrv:
> I don't think this is quite right: the check is really about the xcb-proto version, not the xcffib version (0.12 doesn't exist). Can we just fix the float cast to be a real version parsing function? Also, where did you get xcb-proto of 1.14.1? I only see 1.14 released on their web page.

Yes, I only realized this after submitting my first draft. The fix now is extracting the first two integers and uses that for the float comparison.

I'm not entirely sure about the xcb-proto version, but it seems to be the one currently in Arch Linux' staging environment (might have been upgraded to be able to build against python3.9, I don't know the details).

All in all, this should now make the version detection more robust while not  changing anything else.

ramnes (approved):
LGTM

tych0:
Ok, I mostly just wanted to make sure I wasn't missing something in my test matrix for xcffib :). Patch looks good to me, thanks!

PR #1965, opened 2020-11-14
Replace Window.user_placed_window_setup() with paint_borders()

m-col:
Window.user_placed_window_setup implies an alternative route to similar
functionality seen within Window.place. It is neater and more clear if
Window.place is seen as The One True way to place and show windows.
Where Window.user_placed_window_setup is used, the only functionality
needed is to reconfigure the window borders. This indicates the need to
set window borders without touching the window's geometry (i.e. when
window set it directly), so a Window.paint_borders method can be used
instead where user_placed_window_setup was.


tcld:
Thanks for separating this, I'm looking forward to rebase my own stuff on these changes. :)

PR #1966, opened 2020-11-18
add warning to changelog about new state format during restarts

m-col:


tych0:
Heh, this conflicts now :(. But lgtm

m-col:
that's good because mine has tabs and not spaces ;)

tych0:
your heathenry knows no bounds!

PR #1972, opened 2020-11-21
drop various python < 3.7 version tests

tych0:
Now that we only support >= 3.7, let's drop a bunch of version tests.

### DIFF
--- a/libqtile/core/session_manager.py
+++ b/libqtile/core/session_manager.py
@@ -1,6 +1,5 @@
 import asyncio
 import os
-import sys
 
 from libqtile import ipc
 from libqtile.backend import base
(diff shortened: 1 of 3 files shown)


tych0:
Not quite clear to me why this fails, I tried the latest version of mypy and no luck.

tych0:
When I pass an explicit `--python-version 3.7` to mypy, it works just fine. But all the docs say I shouldn't have to do that...

flacjacket:
We set the minimum Python version that mypy uses for checking in `setup.cfg`, should just need to update that.

tych0:
Ah ha, derp derp. Thanks :)

PR #1975, opened 2020-11-24
WidgetBox widget

elParaguayo:
I've made a new widget that can take other widgets as children. The widgets are hidden by default but shown when the widget is clicked.

The idea was to have a means to hide non-essential widgets by default to reduce clutter on the bar.

The gif below shows how it looks:
widgetbox

I'm still relatively new when looking at qtile's inner workings but I think I've got this set up correctly. I made sure the child widget are still passed to `qtile.register_widget` so they're still accessible to command interfaces.

Button clicks should also be passed to the appropriate child widget.

### DIFF
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -1,0 +1,189 @@
+# Copyright (c) 2020 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from collections import namedtuple
+
+from libqtile import bar
+from libqtile.log_utils import logger
+from libqtile.widget import base
+
+BoxedWidget = namedtuple("Widget", ["widget", "draw"])
+
+
+def _no_draw(*args, **kwargs):
+    pass
+
+
+class WidgetBox(base._Widget):
+    """A widget to declutter your bar.
+
+    WidgetBox is a widget that hides widgets by default but shows them when
+    the box is opened.
+
+    Widgets that are hidden will still update etc. as if they were on the main
+    bar.
+
+    Button clicks are passed to widgets when they are visible so callbacks will
+    work.
+
+    Widgets in the box alsos remain accessible via command interfaces.
+
+    Widgets can only be added to the box via the configuration file. The widget
+    is conigured by adding widgets to the "widgets" parameter as follows:
+
+        widget.WidgetBox(widgets=[
+            widget.TextBox(text="This widget is in the box"),
+            widget.Memory(),
+            ]),
+    """
+    orientations = base.ORIENTATION_HORIZONTAL
+    defaults = [
+        (
+            "font",
+            "sans",
+            "Text font"
+        ),
+        (
+            "fontsize",
+            None,
+            "Font pixel size. Calculated if None."
+        ),
+        (
+            "fontshadow",
+            None,
+            "font shadow color, default is None(no shadow)"
+        ),
+        (
+            "foreground",
+            "#ffffff",
+            "Foreground colour."
+        ),
+        (
+            "close_button_location",
+            "left",
+            "Location of close button when box open ('left' or 'right')"
+        ),
+        (
+            "text_closed",
+            "[<]",
+            "Text when box is closed"
+        ),
+        (
+            "text_open",
+            "[>]",
+            "Text when box is open"
+        ),
+    ]
+
+    def __init__(self, widgets=list(), **config):
+        base._Widget.__init__(self, bar.CALCULATED, **config)
+        self.add_defaults(WidgetBox.defaults)
+        self.box_is_open = False
+        self._widgets = widgets
+        self.add_callbacks({"Button1": self.cmd_toggle})
+
+        if self.close_button_location not in ["left", "right"]:
+            val = self.close_button_location
+            msg = "Invalid value for 'close_button_location': {}".format(val)
+            logger.warning(msg)
+            self.close_button_location = "left"
+
+    def _configure(self, qtile, bar):
+        base._Widget._configure(self, qtile, bar)
+
+        self.layout = self.drawer.textlayout(
+            self.text_closed,
+            self.foreground,
+            self.font,
+            self.fontsize,
+            self.fontshadow,
+            markup=False,
+        )
+
+        for idx, w in enumerate(self._widgets):
+            if w.configured:
+                w = w.create_mirror()
+                self._widgets[idx] = w
+            self.qtile.register_widget(w)
+            w._configure(self.qtile, self.bar)
+
+            # In case the widget is mirrored, we need to draw it once so the
+            # mirror can copy the surface but draw it off screen
+            w.offsetx = self.bar.width
+            self.qtile.call_soon(w.draw)
+
+        # We need to stop hidden widgets from drawing while hidden
+        # (e.g. draw could be triggered by a timer) so we take a reference to
+        # the widget's drawer.draw method
+        self.widgets = [BoxedWidget(w, w.drawer.draw) for w in self._widgets]
+
+        # # Overwrite the current drawer.draw method with a no-op
+        for w in self.widgets:
+            w.widget.drawer.draw = _no_draw
+
+    def calculate_length(self):
+        return self.layout.width
+
+    def set_box_label(self):
+        self.layout.text = (self.text_open if self.box_is_open
+                            else self.text_closed)
+
+    def toggle_widgets(self):
+        for item in self.widgets:
+            try:
+                self.bar.widgets.remove(item.widget)
+                # Override drawer.drawer with a no-op
+                item.widget.drawer.draw = _no_draw
+            except ValueError:
+                continue
+
+        index = self.bar.widgets.index(self)
+
+        if self.close_button_location == "left":
+            index += 1
+
+        if self.box_is_open:
+
+            # Need to reverse list as widgets get added in front of eachother.
+            for item in self.widgets[::-1]:
+                # Restore the original drawer.draw method
+                item.widget.drawer.draw = item.draw
+                self.bar.widgets.insert(index, item.widget)
+
+    def draw(self):
+        self.drawer.clear(self.background or self.bar.background)
+
+        self.layout.draw(0,
+                         int(self.bar.height / 2.0 -
+                             self.layout.height / 2.0) + 1)
+
+        self.drawer.draw(offsetx=self.offsetx, width=self.width)
+
+    def button_press(self, x, y, button):
+        name = "Button{}".format(button)
+        if name in self.mouse_callbacks:
+            self.mouse_callbacks[name]()
+
+    def cmd_toggle(self):
+        """Toggle box state"""
+        self.box_is_open = not self.box_is_open
+        self.toggle_widgets()
+        self.set_box_label()
+        self.bar.draw()
--- a/test/widgets/test_widgetbox.py
+++ b/test/widgets/test_widgetbox.py
@@ -1,0 +1,81 @@
+import pytest
+
+from libqtile.backend.x11 import xcbq
+from libqtile.bar import Bar
+from libqtile.widget import TextBox, WidgetBox
+from test.conftest import BareConfig
+
+
+def no_op(*args, **kwargs):
+    pass
+
+
+class FakeWindow:
+    class _NestedWindow:
+        wid = 10
+
+    window = _NestedWindow()
+
+
+widget_config = pytest.mark.parametrize("qtile", [BareConfig], indirect=True)
+
+
+@widget_config
+def test_widgetbox_widget(qtile):
+    qtile.conn = xcbq.Connection(qtile.display)
+
+    # We need to trick the widgets into thinking this is libqtile.qtile so
+    # we add some methods that are called when the widgets are configured
+    qtile.call_soon = no_op
+    qtile.register_widget = no_op
+
+    tb_one = TextBox(name="tb_one", text="TB ONE")
+    tb_two = TextBox(name="tb_two", text="TB TWO")
+
+    # Give widgetbox invalid value for button location
+    widget_box = WidgetBox([tb_one, tb_two],
+                           close_button_location="middle",
+                           fontsize=10)
+
+    # Create a bar and set attributes needed to run widget
+    fakebar = Bar([widget_box], 24)
+    fakebar.window = FakeWindow()
+    fakebar.width = 10
+    fakebar.height = 10
+    fakebar.draw = no_op
+
+    # Configure the widget box
+    widget_box._configure(qtile, fakebar)
+
+    # Invalid value should be corrected to default
+    assert widget_box.close_button_location == "left"
+
+    # Check only widget in bar is widgetbox
+    assert fakebar.widgets == [widget_box]
+
+    # Open box
+    widget_box.cmd_toggle()
+
+    # Check it's open
+    assert widget_box.box_is_open
+
+    # Default text position is left
+    assert fakebar.widgets == [widget_box, tb_one, tb_two]
+
+    # Close box
+    widget_box.cmd_toggle()
+
+    # Check it's closed
+    assert not widget_box.box_is_open
+
+    # Check widgets have been removed
+    assert fakebar.widgets == [widget_box]
+
+    # Move button to right-hand side
+    widget_box.close_button_location = "right"
+
+    # Re-open box with new layout
+    widget_box.cmd_toggle()
+
+    # Now widgetbox is on the right
+    assert fakebar.widgets == [tb_one, tb_two, widget_box]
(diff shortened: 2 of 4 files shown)


libqtile/widget/widgetbox.py
```diff
@@ -0,0 +1,144 @@
...
+        offset += self.layout.width
+
+        self.drawer.draw(offsetx=self.offsetx, width=self.width)
+
+        for i in self.widgets:
+            i.offsetx = offset
+            offset += i.width
+            i.draw()
+
+    def button_press(self, x, y, button):
```
tych0, line 112: Can you use mouse_callbacks for this, or will that not work?

tych0:
Looks reasonable overall, needs some unit tests though :)

libqtile/widget/widgetbox.py
elParaguayo: This is just for the WidgetBox. I need the x value to work out which widget has been clicked. My thinking was that, by then calling the child's button_press method, it would still trigger the mouse_callbacks.

I don't think using mouse_callbacks for the WidgetBox itself makes sense because we just want to pass these events. I can be persuaded otherwise if you think there's a better approach.

elParaguayo:
I've never written unit tests before so I don't know how to go about it. If there's a good guide (ideally specific to qtile) then I'm very happy to write these too.

tych0:
The best guide is probably the existing unit tests for widgets, see
/test/widgets for some samples.

elParaguayo:
Great - leave it with me.

Also, I note that the checks have failed however I don't think those are to do with my code (could be wrong) but seems to be some 404 errors when installing dependencies.

elParaguayo:
OK, so silly question: once I've written a unit test, how do I actually run it?

tych0:
Agreed r.e. CI, I think #1976 will fix it, thanks!

tych0:
The easiest way is probably just with tox, e.g. TOXENV=py38 tox (or whatever your local python is). The Makefile has examples of how to run other tox stuff.

m-col:
Or: `python -m pytest test/test_widgets.py::my_test` so you don't have to run the whole suite.

libqtile/widget/widgetbox.py
m-col: When you click on a child widget, does the click event not go directly to that widget rather than into the WidgetBox widget?

elParaguayo: Interesting question. I don't _think_ so but it's probably worth checking. 

I assumed that, given the WidgetBox's width is calculated as the total width of the widgets inside it, the bar would assume that any click in that area was just passed to the WidgetBox.

elParaguayo: Checked. If I remove the button handling then the widgets in the box don't register and clicks. Passing the events the way I do seems to work but I may be missing a better way?

I also haven't passed mouse_enter/leave events currently. That seems harder to implement here so probably need to include something in the docs.

m-col: I think it is fine how you override `button_press` because it does need unique logic, but I think it should still funnel into `mouse_callbacks`. Consider something like this:
```
        widget, x_pos = self.find_widget(x)

        if widget == self:
            name = 'Button{0}'.format(button)
            if name in self.mouse_callbacks:
                self.mouse_callbacks[name]()

        elif widget is not None:
            widget.button_press(x_pos, y, button)
```
Then you can get rid of the `clicked` method, and add `self.add_callbacks({'Button1': self.toggle_box})` to `__init__`. This way, more callbacks can be added, or the button1 callback can be overriden.

m-col: As for `mouse_enter` and `mouse_leave`, you can use `find_widget` the same:
```
    def mouse_enter(self, x, y):
        widget, x_pos = self.find_widget(x)
        widget.mouse_enter(x_pos, y)
```

m-col:
BTW I think this is an awesome widget and I will almost certainly use it.

One general question regarding configurability: would it be possible to specify the position of the "tab", or whatever you call it... the "[<]" part? In your gif it is at the right when closed and at the left when open, but it might be cool and straightforward to let both open and close sides to be configurable.

libqtile/widget/widgetbox.py
elParaguayo: @m-col : yes, that could work for `mouse_callbacks`. Happy to make that change.

As for `mouse_enter` the above idea only when you enter the widget box. The problem I have is when you're already in the box but move horizontally over different widgets in the box. In that scenario, the mouse_enter wouldn't get triggered more than once. If we can't sort that then `mouse_leave` would potentially fire in a different widget to the one that was entered.

We would constantly need to know mouse_position in the widget to implement this effectively.

elParaguayo:
Yes - I had the same thought too and have started updating the code for that.

Also, still trying to get my head round unit tests. Don't understand the scope. If I run one test, which other tests are being pulled in? e.g. by running `test_chord.py` the code which defines `fake_bar` is also being run. I find that odd but only because I've never done testing before!

elParaguayo:
Sorry. Making no progress on unit test. Really not sure how to write it. 

The draw and _configure methods are quite important here. Can those be run in a test?

libqtile/widget/widgetbox.py
m-col: Ah yes....  Is there any way you think the child widgets can be handled in a way that they are actually part of the bar's normal widget list, rather than a 'sub'-widget? That would also simplify the button press logic because everything would be direct. Maybe to hide and show the widgets all you need to do is move them in and out of the bar's `self.widgets` list.

m-col:
Yes, but testing the result might have the same effect i.e. the widget will have a fixed length when it is closed, and will be wider when it is open, so you can test those and you can reasonably conclude it works. It would probably be worth testing x positions too of the widget and the widgets that it controls, particularly if you do modify it as per my other comment from a few minutes ago where the handled widgets are actually on the bar's widget list. That way you can just test the bar's widgets' x positions and widths, toggle the WidgetBox, then test them again. If they change as you expect, then bingo.

libqtile/widget/widgetbox.py
elParaguayo: That has a horrible sound of being a far better way for this widget to work! Will investigate.

elParaguayo: Genius. So much simpler. Thanks.

elParaguayo:
The change recommended by @m-col means that unit testing is far easier now. Should have these done today.

tych0:
Can you fold the PEP8 fixes into their original commits? Other than that (and a simple rebase), lgtm

elParaguayo:
Sorry, dumb question as I'm not a git expert. Not sure what the difference is between the fold and rebase in your comment. I can squash the pep8 fixes into their previous commits but would you rather I just changed this into a single commit?

tych0:
Multiple commits is fine as long as they're logical entities, it's just the linting fixes you can fold into their original commits.

elParaguayo:
Is the quickest fix here for me just to merge current master, resolve the conflict and push a new commit?

tych0:
I would rebase on master, as merges in feature branches can get kind of confusing. But yeah, rebase and squash as you see fit!

elParaguayo:
I find git very confusing sometimes. I've rebased onto the latest master but some commits are listed twice now. Is that what's meant to happen?

m-col:
Nope, that's not meant to happen. The easiest way to fix it would be to:
1. copy the files that you have changed for the PR (`libqtile/widget/__init__.py`, `libqtile/widget/widgetbox.py`, and `test/widgets/test_widgetbox.py `) out of the git folder
2. remove those 29 commits using `git reset --hard HEAD~29`
3. copy those three files back into place, so they appear as changes in `git diff`.
4. commit those changes into your 'final' PR commit
5. rebase to master with `git pull --rebase upstream master` assuming 'upstream' points to https://github.com/qtile/qtile
6. the push your changes to your own branch here with `git push -f origin hide_widgets` assuming 'origin' points to https://github.com/elParaguayo/qtile

elParaguayo:
Thanks m-col. I need to do some reading on git usage. Still came up with merge conflicts doing those steps. However, the PR looks neater now!

m-col:
Is this all ready to go from your end?

libqtile/widget/widgetbox.py
```diff
@@ -0,0 +1,169 @@
...
+        self.toggle_widgets()
+
+    def toggle_box(self):
+        self._change_state()
+        self.set_box_label()
+        self.bar.draw()
+
+    def cmd_toggle(self):
+        """Toggle box state"""
+        self.toggle_box()
```
m-col, line 169: Could you combine these 3 functions into a single function?
```
def cmd_toggle(self):
        self.box_is_open = not self.box_is_open
        self.toggle_widgets()
        self.set_box_label()
        self.bar.draw()
```

libqtile/widget/widgetbox.py
```diff
@@ -0,0 +1,169 @@
...
+    def set_box_label(self):
+        self.layout = self.drawer.textlayout(
+            self.text_open if self.box_is_open else self.text_closed,
+            self.foreground,
+            self.font,
+            self.fontsize,
+            self.fontshadow,
+            markup=False,
+        )
+
```
m-col, line 123: Currently you're creating a new textlayout each time `self.toggle` is called. If the only thing that needs updating is the text, you can update that directly by doing `self.layout.text = self.text_open if self.box_is_open else self.text_closed`

libqtile/widget/widgetbox.py
```diff
@@ -0,0 +1,169 @@
...
+    def toggle_widgets(self):
+        for widget in self.widgets:
+            try:
+                self.bar.widgets.remove(widget)
+            except ValueError:
+                continue
+        try:
+            index = self.bar.widgets.index(self)
+        except ValueError:
+            logger.warning("Can't locate item in bar")
```
m-col, line 133: I think this try block can be removed, just leaving the `index =` line. I don't think this error can get raised (under normal usage), and if it does, then we want whatever is trying to toggle the widgets to stop trying, rather than to simply log and carry on. Indeed, as is, line 136 will error because `index` doesn't exist if this error is raised.

elParaguayo:
@m-col many thanks for taking a look at this. Let me go over these and update as appropriate.

libqtile/widget/widgetbox.py
m-col: Does that approach work completely? Widgets that are hidden will continue doing their polling or other behaviour in the background, which is fine, but some will modify their `self.drawer` directly even if they are not within a bar's widget list. Does this not cause issues? The `Drawer`s will paint onto the bar's window, no?

m-col: I should also add that mirrored widgets are not working for me. I.e. if I do this:
```
clock = widget.Clock()
widgetbox = widget.WidgetBox(widgets=[clock])
mybar = bar.Bar([widgetbox, clock])
```
`clock` within the bar's widget list is replaced with a `Mirror` widget, copying the pixels from the `clock` widget that is held by `widgetbox`. For some reason, likely related to how drawing is being handled, the mirror widgets aren't updating unless the widgetbox is visible.

elParaguayo: > Does that approach work completely? Widgets that are hidden will continue doing their polling or other behaviour in the background, which is fine, but some will modify their `self.drawer` directly even if they are not within a bar's widget list. Does this not cause issues? The `Drawer`s will paint onto the bar's window, no?

Interesting. I hadn't experienced this but I can see your point. Not sure what the fix is for this because I think the widgets should continue to poll.

elParaguayo: > I should also add that mirrored widgets are not working for me. I.e. if I do this:
> 
> ```
> clock = widget.Clock()
> widgetbox = widget.WidgetBox(widgets=[clock])
> mybar = bar.Bar([widgetbox, clock])
> ```
> 
> `clock` within the bar's widget list is replaced with a `Mirror` widget, copying the pixels from the `clock` widget that is held by `widgetbox`. For some reason, likely related to how drawing is being handled, the mirror widgets aren't updating unless the widgetbox is visible.

I'm not familiar with mirror widgets. What's the expected behaviour here?

m-col: > Interesting. I hadn't experienced this but I can see your point. Not sure what the fix is for this because I think the widgets should continue to poll.

Can you confirm with a bunch of different widgets whether or not this is actually a problem? The issue isn't that they continue to poll (that is to be expected), but that they will by themselves try to draw pixels onto the bar's window. A potential solution would be to hijack each widget's `draw` method and substitute it when hidden with a no-op. Then revert it to show the widget. You may or may not need to do the bar list manipulation... not sure.

> I'm not familiar with mirror widgets. What's the expected behaviour here?

You can see the implementation [here](https://github.com/qtile/qtile/blob/master/libqtile/widget/base.py#L621), and where it is used within `Bar._configure`, including the block that you've copied here. They let users use a single widget instance (and therefore only one object doing tasks) in multiple places on the same bar, or on multiple bars. Any time a widget gets `_configure`-ed a further time after the first time, it gets replaced by a mirror widget that copies the original's pixels. They get their own `Drawer` instance that hooks into the drawing of the widget that they mirror's drawer, but I suspect that no-oping their `draw` method as above will probably fix any potential issues.

elParaguayo: Yes, can do a lot more testing over the weekend. As for hijacking the draw function, the was the approach that I was considering so I'm glad to see I'm on the right track!

elParaguayo: > I should also add that mirrored widgets are not working for me. I.e. if I do this:
> 
> ```
> clock = widget.Clock()
> widgetbox = widget.WidgetBox(widgets=[clock])
> mybar = bar.Bar([widgetbox, clock])
> ```
> 
> `clock` within the bar's widget list is replaced with a `Mirror` widget, copying the pixels from the `clock` widget that is held by `widgetbox`. For some reason, likely related to how drawing is being handled, the mirror widgets aren't updating unless the widgetbox is visible.

OK - I understand what you mean now. That's an interesting issue because, based on the other issue you identified, we need to suppress the hidden widget's draw function. We could prevent mirroring by forcing the widgets in the box to have `configured = False` but that feels a bit ugly and might cause unexpected issues if widgets rely on that flag for other purposes.

Will do some more thinking about this.

elParaguayo: Ok... so I _think_ I've got this worked out. For the `Mirror` to work, the widget still needs to prepare its surface so we can't just detach the `draw` function. Instead, I've detached the widget's `drawer.draw` function.

This seems to work (mirror updates when widget is hidden in box, hidden widgets don't display when polled). Only downside is that it's broken my unittest so I'll need to rework that!

test/widgets/test_widgetbox.py
```diff
@@ -0,0 +1,78 @@
...
+
+@widget_config
+def test_widgetbox_widget(qtile):
+    qtile.conn = xcbq.Connection(qtile.display)
+    qtile.test_xeyes()
+    qtile.windows_map = {}
+
+    # Disconnect functions called by widget but aren't needed here
+    qtile.call_soon = no_op
+    qtile.register_widget = no_op
```
m-col, line 31: Perhaps confusingly, the `qtile` object here isn't actually `libqtile.qtile`, it's an instance of the `Qtile` class defined within "test/conftest.py". With this we interact with the running qtile process using a command client at `qtile.c`. In any case, leaving functions like these intact should be fine (indeed, that is probably preferable so that the test objects are as close as possible to the normal ones).

elParaguayo: Maybe the comment about "disconnecting" the functions is inappropriate. Without those lines I get an error `AttributeError: 'Qtile' object has no attribute 'call_soon'`. Your explanation that it's a different object would seem to explain why that happens.

I take your point about tests being as close as possible to normal operation but I'm not very familiar with pytest or how you can use qtile objects outside of qtile itself.

test/widgets/test_widgetbox.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+widget_config = pytest.mark.parametrize("qtile", [BareConfig], indirect=True)
+
+
+@widget_config
+def test_widgetbox_widget(qtile):
+    qtile.conn = xcbq.Connection(qtile.display)
+    qtile.test_xeyes()
+    qtile.windows_map = {}
```
m-col, line 27: I don't think these 3 lines are needed

libqtile/widget/widgetbox.py
m-col: Awesome!

test/widgets/test_widgetbox.py
elParaguayo: The first one is. The widgets can't create their `Drawer` objects without it. The second and third line aren't needed.

test/widgets/test_widgetbox.py
m-col: No worries at all, tests do take a bit of getting used to.

OK that makes sense then. Maybe just update the comment so it's clear that we're tricking the widget into thinking `qtile` is `libqtile.qtile`. The rest of the testing looks good!

libqtile/widget/widgetbox.py
```diff
@@ -0,0 +1,189 @@
...
+            if w.configured:
+                w = w.create_mirror()
+                self._widgets[idx] = w
+            self.qtile.register_widget(w)
+            w._configure(self.qtile, self.bar)
+
+            # In case the widget is mirrored, we need to draw it once so the
+            # mirror can copy the surface but draw it off screen
+            w.offsetx = self.bar.width
+            self.qtile.call_soon(w.draw)
```
m-col, line 130: Does this continue to allow mirror widgets that mirror a widget that is hidden inside the widgetbox to update when the hidden one tries to update?

test/widgets/test_widgetbox.py
m-col: Ah I see, OK

libqtile/widget/widgetbox.py
elParaguayo: It should do, yes.

The initial `draw` is called here because, without it, the mirror widget is blank until the box is opened for the first time. Calling `draw` means that the widget's drawer surface is drawn so the mirror has something to copy.

If we use the clock widget as an example and put the original in the box and the widget outside it, the mirror continues to show the correct time even if you don't open the box. Is that the issue you're concerned about or have I missed a different scenario?

m-col:
No that is exactly what I was getting at, and if that works then that is great!
I'm not sure I fully understand how that works yet lol but I will have a play
around with it now!

m-col:
That comment above was meant to be a reply to [this](https://github.com/qtile/qtile/pull/1975#discussion_r546246476) thread, which I replied to by email.... not sure why it went to the main thread.

Anyway this is looking great, and I just had a play around with it and it works perfectly!

Please could you be squash all the commits, and then LGTM!

elParaguayo:
You need the initial `draw` because, otherwise, the widgets are only drawn once the bar is configured and there's a `bar.draw()`. However, as the box is closed by default, anything in the box is not in bar's list of widgets. As a result, those widgets aren't drawn and the mirror has nothing to copy. Therefore anything in the box needs to be drawn _before_ the `bar.draw()`. Took me a while to work that one out.

All the timers etc. continue to work as they're not called by the bar. The widgets' `draw` methods are still called, the surface (which is copied by the mirror) is updated, but the final `drawer.draw()` isn't called so they don't draw themselves on the bar.

m-col:
Ah OK yeah that makes sense! It's good to have the explanation here for posterity, in case at some point someone gets confused, they can look here :)

m-col:
Oh and maybe add an entry to the changelog mentioning the new widget?

elParaguayo:
Ah - changelog! Let me add that.

m-col:
Brilliant thanks for this :)

PR #1979, opened 2020-11-26
Add WindowCount widget

elParaguayo:
Following [a request on the qtile-dev group](https://groups.google.com/g/qtile-dev/c/froEabtHAb4), here's a very basic widget that displays number of windows in the current group.

I've also added a unit test which is, putting it mildly, ugly!


tych0:
Less ugly than not having a test, thanks!

PR #1983, opened 2020-11-29
Qtile check

tych0:
This adds a new subcommand that checks various correctness elements of configs, aka the first bit of #1773.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -15,14 +15,6 @@ class Core(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def get_keys(self) -> typing.List[str]:
-        pass
-
-    @abstractmethod
-    def get_modifiers(self) -> typing.List[str]:
-        pass
-
-    @abstractmethod
     def grab_key(self, key: config.Key) -> typing.Tuple[int, int]:
         """Configure the backend to grab the key event"""
 
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -1,0 +1,63 @@
+# Copyright (c) 2020, Tycho Andersen. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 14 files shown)


libqtile/backend/base.py
```diff
@@ -10,14 +10,6 @@ class Core(metaclass=ABCMeta):
...
         pass
 
-    @abstractmethod
-    def get_keys(self) -> typing.List[str]:
-        pass
-
-    @abstractmethod
-    def get_modifiers(self) -> typing.List[str]:
-        pass
-
```
m-col, line 20: It makes sense for these to be in the base backend, so maybe instead of moving them out of the core into standalone functions, maybe (abstract) static methods would do the job?

tych0:
Either that or we start passing around the module object instead of a
Core object. Static methods generally feel ugly to me, but maybe I've
been out of python too long :)

test/test_check.py
```diff
@@ -0,0 +1,72 @@
...
+def check_literal_config(config):
+    with tempfile.NamedTemporaryFile(suffix=".py") as temp:
+        temp.write(textwrap.dedent(config).encode('utf-8'))
+        temp.flush()
+        return run_qtile_check(temp.name)
+
+
+def test_check_bad_syntax():
+    assert not check_literal_config("""
+        meshuggah rocks!
```
m-col, line 56: nice

libqtile/scripts/check.py
```diff
@@ -0,0 +1,63 @@
...
+        print("mypy not found, can't type check config file\n"
+              "install it and try again")
+    else:
+        try:
+            subprocess.check_call(["mypy", args.configfile])
+            print("config file type checking succeeded")
+        except subprocess.CalledProcessError:
+            print("config file type checking failed")
+            sys.exit(1)
+
```
m-col, line 45: Could the error message be printed as well? There are a few reasons this could fail e.g. 'mypy' not found, or mypy has some kind of bug/error due to a user's installation for whatever reason. Might be useful.

tych0: The `shutil.which()` above should prevent it from being not found, and the actual invocation is hooked up to the same stdout/stderr here, so I think it should mostly be irrelevant. But yeah, probably a good idea not to ignore the error.

m-col:
In general, I would agree that static methods are ugly, but this seems like the
one type of use case where it makes most sense i.e. some kind of utility
function that "belongs" to a base class that acts as a template. Passing around
the module object sounds even worse... maybe we can just stick to what you have
done (standalone functions) and remember to implement those again with any new
backends. It isn't like new backends are created with great frequency :)

tych0:
Well, you're not wrong that this way is ugly too. Let me play around
with it and see what I can come up with.

ramnes:
Let's not overthink this. I think we can keep it that way and refactorize as needed when we add a second core.

tych0:
> Let's not overthink this. I think we can keep it that way and refactorize as needed when we add a second core.

I was leaning toward that too since I couldn't figure out a nice way to do it. But first I have to get CI to pass on my branch :+1: 

Of course, it works locally...

PR #1985, opened 2020-11-29
start: only log in color mode when stdout is a tty

tych0:
For the most part, people are looking at logs in ~/.xsession-errors, which
is not a tty. let's not spew a bunch of color changing control codes there.


m-col:
LGTM

Issue #1419, opened 2019-09-03
Rofi spawns on the monitor that has the mouse, not focus


tamis-laan:
I trigger rofi with the command `rofi -monitor -1` this tells rofi to open it's window on the currently focused monitor. However rofi opens on the monitor where my mouse cursor is placed and not on the monitor where the focus is. 

This is a fallback mechanism of rofi, thus the information about what monitor is in focus is not being passed allong. How can circumvent this issue?? Is this a bug, or is this feature lacking in qtile?

tych0:
I believe this is fixed by #1716

owarai:
Same problem here. 
monitor A, B. 
I write something in vim on monitor B which focused on group 2, same time mouse pointer is stil located on monitor A which focused on group 1. Then if I trigger rofi with shortcuts(corresponding command: `rofi show drun -monitor -1`), monitor B auto unfocused and rofi window show up on monitor A. I also set `m: "-1";` in ~/.config/rofi/config.rasi to ensure the global behavior.

> -1 means show rofi on focused monitor. 
> Also try this keybindings on sxhkd, same result.

Maybe this issue should take another look?

kaistian:
It's probably because Qtile doesn't support _NET_DESKTOP_VIEWPORT ref this issue davatorium/rofi/issues/1158

owarai:
> It's probably because Qtile doesn't support _NET_DESKTOP_VIEWPORT ref this issue [davatorium/rofi/issues/1158](https://github.com/davatorium/rofi/issues/1158)

Yes, output of `xprop -root` lack this property. 

@tych0 . Sorry to bother, could you take a look to confirm this?
If so, is there any plan to support this EWMH property?

tych0:
Yeah, looks like that's the issue. I have no plans to implement it, but would be happy to accept a patch with tests that does so.

m-col:
The commit in #1991 may fix this, but I am having trouble confirming. When I run `rofi -monitor -1 -show combi -no-config` it tells me it cannot find the viewport information for this desktop, number 3, when I only have two monitors (laptop and one external). Anyone else see this? Could well be a rofi bug but I'm erring more towards me doing something wrong.

kaistian:
I'm using an older Qtile #1500 (because of annoying reordering of windows) but don't this it matters for testing this so I patch my files manually.
[...]

PR #1991, opened 2020-12-02
set _NET_DESKTOP_VIEWPORT

m-col:
This is an attempt at dealing with #1419, but simply setting _NET_DESKTOP_VIEWPORT does not fix the problem annoyingly.

With these changes, the viewport is being changed (correctly) each time I focus a different screen both by moving the mouse or by focussing to a window on that screen with a keybinding.

However, when I run `rofi -monitor -1 -show combi -no-config`, focus moves to the screen where the mouse is and the rofi window opens there, which is fallback behaviour, but the rofi window should instead open on the current screen. I'll continue debugging.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Qtile x.x.x, released xxxx-xx-xx:
     * features
         - Add `place_right` option in the TreeTab layout to place the tab panel on the right side
+        - X11: Add support for _NET_DESKTOP_VIEWPORT. E.g. can be used by rofi to map on current output.
     * bugfixes
         - Remove non-commandable windows from IPC. Fixes bug where IPC would fail when trying to get info
           on all windows but Systray has icons (which are non-commandable `_Window`s.)
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -429,6 +429,10 @@ class Core(base.Core):
         self._root.set_property("_NET_NUMBER_OF_DESKTOPS", len(groups))
         self._root.set_property("_NET_DESKTOP_NAMES", "\0".join(i.name for i in groups))
         self._root.set_property("_NET_CURRENT_DESKTOP", index)
+        viewport = []
+        for group in groups:
+            viewport += [group.screen.x, group.screen.y] if group.screen else [0, 0]
+        self._root.set_property("_NET_DESKTOP_VIEWPORT", viewport)
 
     def lookup_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
         """Find the keysym and the modifier mask for the given key"""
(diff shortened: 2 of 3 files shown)


m-col:
The [spec](https://specifications.freedesktop.org/wm-spec/1.3/ar01s03.html) and [this](https://github.com/davatorium/rofi/blob/5d7bf1cfcf53f3f56a63ef5daf714e82c30adbbf/source/x11-helper.c#L258) snippet from rofi's source suggest that the _NET_DESKTOP_VIEWPORT data should be a list of pairs per virtual desktop i.e. `[[0, 0], [1920, 0], ...]`, but that is given me some `struct` packing errors from xcffib so I'm probably doing the `set_property` call wrong there.

tych0:
Yeah, the xcffib/xpyb api for this is not very good. I think you want something like (totally untested): https://github.com/tych0/qtile/commit/7ee0c837ae92a0344c53ccc5006ed607be324976

elParaguayo:
Needs to be rebased against master and then
``` diff
diff --git a/libqtile/backend/x11/xcbq.py b/libqtile/backend/x11/xcbq.py
index 5cb20f87..e1d4bf67 100644
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -193,6 +193,7 @@ SUPPORTED_ATOMS = [
     "_NET_CLIENT_LIST",
     "_NET_CLIENT_LIST_STACKING",
     "_NET_CURRENT_DESKTOP",
+    "_NET_DESKTOP_VIEWPORT",
     "_NET_ACTIVE_WINDOW",
     "_NET_SUPPORTING_WM_CHECK",
     # From http://standards.freedesktop.org/wm-spec/latest/ar01s05.html
```

elParaguayo:
Does it work for you?

m-col:
Yes - interesting. Did something in rofi change?

m-col:
Test seems unrelated:
```
_______ test_khal_calendar_upcoming_and_live_event[1-x11-khal_manager1] ________
  
  khal_manager = <test.helpers.TestManager object at 0x7f5a1d088410>
  
      @pytest.mark.parametrize("khal_manager", [(50, dict()), (70, dict())], indirect=True)
      def test_khal_calendar_upcoming_and_live_event(khal_manager):
          """Check foreground colour changes when within reminder period."""
          widget = khal_manager.c.widget["khalcalendar"]
          info = widget.info()
  >       assert info["text"] == "Today 2021-01-01 10:00-11:00 Write Qtile tests"
  E       AssertionError: assert 'Calendar not initialized.' == 'Today 2021-0...e Qtile tests'
  E         - Today 2021-01-01 10:00-11:00 Write Qtile tests
  E         + Calendar not initialized.
  
  test/widgets/test_khal_calendar.py:122: AssertionError
```

rofi was almost certainly looking in `_NET_SUPPORTED` to see if `_NET_DESKTOP_VIEWPORT`, and because we hadn't done that it wasn't working. Which is a pretty dumb mistake on my part lol - thanks for pointing that out!

m-col:
I pushed again with a changelog entry for the atom. Maybe tests will flake and pass.

elParaguayo:
I'll revert that khal calendar test. Clearly too flaky.

m-col:
Tests passing this time. We happy to merge this?

elParaguayo:
Absolutely.

Issue #1982, opened 2020-11-29
Moving from dbus to dbus_next


elParaguayo:
This isn't an "issue" as such but is just somewhere for me to set out progress and ask questions related to moving from dbus to dbus_next.

Tasks to do:
- [X] Strip dbus/glib mainloop out of core/master.py
- [X] Notify widgets when asyncio's eventloop has started
- [X] Update mpris2 widget (I'm aware there's a pull request for an updated version of this widget. If that gets merged I'll need to redo this one)
- [x] Update keyboardkbdd widget - Tricky. I can't test this one and I don't really understand what it does!
- [x] Look at base widgets and see if anything should be updated
- [x] Update notify.py
_List last updated: 4 January 2021_

Removed tasks:
- Update mpris widget. Widget will be deleted (see #2066)

Happy to add to the list if anyone has any comments.

elParaguayo:
OK - first question.

I'm looking at widget/base.py. 

The ThreadedPollText class uses threading for the poll
``` python
    def tick(self):
        def worker():
            try:
                text = self.poll()
                if self.qtile is not None:
                    self.qtile.call_soon_threadsafe(self.update, text)
            except:  # noqa: E722
                logger.exception("problem polling to update widget %s", self.name)
        # TODO: There are nice asyncio constructs for this sort of thing, I
        # think...
        threading.Thread(target=worker).start()
```

But the ThreadPoolText uses asyncio
``` python
    def timer_setup(self):
        def on_done(future):
            try:
                result = future.result()
            except Exception:
                result = None
                logger.exception('poll() raised exceptions, not rescheduling')

            if result is not None:
                try:
                    self.update(result)

                    if self.update_interval is not None:
                        self.timeout_add(self.update_interval, self.timer_setup)
                    else:
                        self.timer_setup()

                except Exception:
                    logger.exception('Failed to reschedule.')
            else:
                logger.warning('poll() returned None, not rescheduling')

        future = self.qtile.run_in_executor(self.poll)
        future.add_done_callback(on_done)
```
It's not immediately clear to me what the differences are which means we have two classes here. One of the docstrings talks about pushing vs pulling.

I'd assume we'd look to move the threading with an ayncio coroutine but would appreciate any thoughts.

tych0:
On Sun, Nov 29, 2020 at 06:37:40AM -0800, elParaguayo wrote:
> OK - first question.
> 
> I'm looking at widget/base.py. 
> 
[...]

PR #1994, opened 2020-12-05
Dbus_next migration

elParaguayo:
**Warning: there are a lot of commits in this PR so it will be necessary to rebase**

This is a work in progress with further commentary in https://github.com/qtile/qtile/issues/1982

This is not ready for merging but I've created this PR so people can test if interested. Feedback greatly appreciated.


libqtile/notify.py
```diff
@@ -29,29 +30,24 @@
...
 try:
-    import dbus
-    from dbus import service
-    from dbus.mainloop.glib import DBusGMainLoop
-    DBusGMainLoop(set_as_default=True)
-    has_dbus = True
+    from dbus_next.aio import MessageBus  # noqa: I001
+    from dbus_next.service import method, signal, ServiceInterface  # noqa: I001
+    has_dbus = True  # noqa: I005S
 except ImportError:
```
tych0, line 36: You can probably just dispense with this. dbus_next looks lightweight enough that we can just require it.

elParaguayo:
Apologies - last bunch of commits is very ugly. Used wrong command from history by mistake which mucked a few things up! Can tidy this all up when it's ready to rebase.

tych0:
Yeah, no worries. Feel free to rewrite it and force push as you go along. I do that all the time for branches in progress :-)

ramnes:
I really like what I'm seeing here at a quick glance, keep up the good work!

elParaguayo:
@ramnes - thanks for the encouragement!

dbus doesn't seem to be too deeply integrated in the existing code (which is quite a relief).

Looking at my to do list for this (copied from #1982):
- [X] Strip dbus/glib mainloop out of core/master.py
- [X] Notify widgets when asyncio's eventloop has started
- [X] Update mpris2 widget (I'm aware there's a pull request for an updated version of this widget. If that gets merged I'll need to redo this one)
- [ ] Update mpris widget (is this necessary? If mpris has been deprecated in favour of mpris2, should we just remove this widget?)
- [ ] Update keyboardkbdd widget - Tricky. I can't test this one and I don't really understand what it does!
- [X] Look at base widgets and see if anything should be updated
- [X] Update notify.py

I'm not sure what else there is to do or whether I should now focus on testing this. Would appreciate thoughts on where I could best direct my efforts here.

One thought, I've seen some issues and PRs about the notification code. Given those discussions, it may be worth stripping out any changes related to dbus_next until the other points have been resolved.

DorianGray:
This PR is super important to me as well, and is more important than any new features imo.

elParaguayo:
Which bits are critical for you? Want to make sure I don't miss anything!

DorianGray:
I use qtile with pypy and pygobject/dbus doesn't build on pypy...so all of it!

elParaguayo:
@DorianGray I've got a dbus battery widget already (https://github.com/elParaguayo/qtile-widget-laptopbattery) which I've updated to use dbus_next. Once we've got dbus_next into the main branch then I'll look at submitting a pull request for it. 

I've also got a wifi widget using NetworkManager's dbus interface but that one still needs a lot of work!!

DorianGray:
mpris is gone =) Also, CI works reliably now if you rebase.

DorianGray:
btw, thanks for the link to your widget! It helped me in testing https://github.com/qtile/qtile/pull/2046

elParaguayo:
@DorianGray I certainly need to rebase against master as there have been some changes which, I suspect, will cause some conflicts with my code. I should probably do some tidying up/squashing of the various commits too.

And, yes, happy to help on a wifi widget.

ramnes:
Uh, I'm not sure everything in a single commit is a good thing either. :)

DorianGray:
Would you mind breaking out the ThreadedPollText removal into a separate PR? That would allow us to close https://github.com/qtile/qtile/pull/2076

elParaguayo:
Sorry @ramnes i got a bit overexcited with the squashing! 

Happy to fix but might need to wait until the weekend.

elParaguayo:
@DorianGray Sorry, I'm not a pro on github. If you want the removal of ThreadedPollText in a new PR, I can do a new branch and cherry pick that commit but it does impact some of my other changes here. If I rebase this branch after that PR is merged, am I correct in thinking that the commit will just disappear from here?

DorianGray:
yeah, although if it's a different sha due to rebase at the time you may need to `rebase -i master` and then `d` the threadedpolltext removal commit.

DorianGray:
also, if this is nearing getting ready to merge don't worry about it too much =) love these changes

elParaguayo:
@DorianGray PR for removing ThreadedPollText is up. See #2084.

DorianGray:
You are the best <3

ramnes:
Can you do the rebase? And then what's left, only the keyboardkbdd widget?

libqtile/notify.py
```diff
@@ -22,63 +23,55 @@
...
+from dbus_next.service import ServiceInterface, method, signal  # noqa: I001
 
-try:
-    import dbus
-    from dbus import service
-    from dbus.mainloop.glib import DBusGMainLoop
-    DBusGMainLoop(set_as_default=True)
-    has_dbus = True
-except ImportError:
-    has_dbus = False
```
ramnes, line 38: I think we should keep dbus as an optional dependency, and thus keep that kind of stuff. That would also resolve your problem with package maintainers. :)

ramnes: Forget my comment, Tycho is right, a pure-python dependency is not a big deal.

docs/manual/install/index.rst
```diff
@@ -67,14 +67,12 @@ You'll also need ``libpangocairo``, which on Ubuntu can be installed via ``sudo
...
 ------------
 
-Until someone comes along and writes an asyncio-based dbus library, qtile will
-depend on ``python-dbus`` to interact with dbus. This means that if you want
-to use things like notification daemon or mpris widgets, you'll need to
-install python-gobject and python-dbus. Qtile will run fine without these,
-although it will emit a warning that some things won't work.
+Qtile requires ``dbus_next`` which is a lightweight, asyncio-based dbus library
+to interact with dbus and provide things like notification daemon or mpris
+widgets.
```
ramnes, line 75: I guess we should just include dbus_next in setup.cfg and remove that part of the doc.

libqtile/notify.py
elParaguayo: I think that's right. 

My question is more about etiquette/procedure. I assume it's the package maintainers' duty to identify dependency changes but wanted to check.

libqtile/widget/base.py
```diff
@@ -203,6 +202,18 @@ def _configure(self, qtile, bar):
         if not self.configured:
             self.configured = True
             self.qtile.call_soon(self.timer_setup)
+            self.qtile.call_soon(self._eventloop_started)
+
+    def _eventloop_started(self):
```
ramnes, line 207: I guess we probably want to do async stuff most of the time with this method, so why not making it async in the first place? That would make the diff simpler on the widgets you used it for. :)

libqtile/notify.py
ramnes: I guess we'll see if people complain or not. :)

ramnes:
Let's not forget to update the change log.

libqtile/widget/base.py
elParaguayo: D'oh! Yes, much better idea. 

My expectation was that this method would just be for asyncio stuff so this makes much more sense than the approach I adopted.

elParaguayo: Might change the name back to `_config_async` as that would make more sense with this change.

libqtile/widget/mpris2widget.py
```diff
@@ -97,39 +130,45 @@ def update(self, interface_name, changed_properties, invalidated_properties):
...
-        if playbackstatus == 'Paused':
-            if self.stop_pause_text is not None:
-                self.is_playing = False
-                self.displaytext = self.stop_pause_text
-            elif self.displaytext:
-                self.is_playing = False
-                self.displaytext = 'Paused: {}'.format(self.displaytext)
+        if playbackstatus:
+            playbackstatus = playbackstatus.value
+            if playbackstatus == 'Paused':
```
ramnes, line 151: Can we make the diff simpler by not adding that extra level of indentation?

elParaguayo: Will add to the "to do" list.

libqtile/widget/base.py
elParaguayo: I now remember why I didn't do it this way: you can't do `asyncio.create_task` in `_config` as the eventloop may not have started so you get an error. However, I do think having an async method makes sense so I'll fix that.

docs/manual/install/index.rst
elParaguayo: I've added dbus_next to setup.cfg but have left this bit in the docs as that's in the "installing from source" section so it's correct to show the dependency.

elParaguayo:
Have made all the changes suggested. Just need to rebase but there are some conflicts that I need to fix before uploading. Ran out of time to do this before work so will do later today.

docs/manual/install/index.rst
ramnes: The `pip install .` in the next section will already install it if it's in the setup.cfg.

libqtile/widget/mpris2widget.py
```diff
@@ -68,25 +76,50 @@ def __init__(self, **config):
...
+        self.bus = await MessageBus().connect()
+        rule = ",".join("{}='{}'".format(k, v)
+                        for k, v in self.match_args.items())
+        subscribe = await self.bus.call(
+            Message(destination='org.freedesktop.DBus',
+                    interface='org.freedesktop.DBus',
+                    path='/org/freedesktop/DBus',
+                    member='AddMatch',
+                    signature='s',
+                    body=[rule]))
```
ramnes, line 111: Is there anything higher level than this in dbus-next? Something closer to the original `add_signal_receiver` method?

elParaguayo: Excellent question and I asked something very similar on the dbus_next repo: https://github.com/altdesktop/python-dbus-next/issues/75 and https://github.com/altdesktop/python-dbus-next/issues/76

The second link shows that there's no higher level method currently but this might change in the future.

ramnes: Then we should probably add a small function in `libqtile.utils` that does just this, that we could propose upstream later. What do you think?

elParaguayo: That would be a neater way of doing it, especially as it's code that could be used by other widgets. 

I'll knock something up and add it to this PR.

docs/manual/install/index.rst
elParaguayo: Ah. I understand now. Will fix.

libqtile/widget/mpris2widget.py
elParaguayo: I've left these as separate commits for now so you can see the changes I've made. Can squash them later.

libqtile/utils.py
```diff
@@ -30,6 +30,10 @@
 from random import randint
 from shutil import which
 
+from dbus_next import Message  # type: ignore
+from dbus_next.aio import MessageBus  # type: ignore
+from dbus_next.constants import BusType, MessageType  # type: ignore
+
 from libqtile.log_utils import logger
 
 _can_notify = False
```
elParaguayo, line 39: I've just noticed this section and the `send_notification` method further down. I assume we want to strip this out to use dbus_next again, rather than gi?

libqtile/utils.py
```diff
@@ -310,3 +314,84 @@ def scan_files(dirpath, *names):
...
+                                       bus_name=bus_name,
+                                       path=path)
+
+    return success
+
+
+async def remove_signal_handler(session_bus=False, signal_name=None,
+                                dbus_interface=None, bus_name=None, path=None):
+    """
+    Helper function to remove presvious subscribed dbus message handlers.
```
ramnes, line 383: previous

ramnes: previously

libqtile/widget/mpris2widget.py
```diff
@@ -68,25 +75,38 @@ def __init__(self, **config):
...
+    async def _config_async(self):
+        await self._connect_dbus()
+
+    async def _connect_dbus(self):
+        if self.connect_count < self.MAX_RETRIES:
+            self.connect_count += 1
+            await self._setup()
+        else:
+            logger.warning("Exceeded max retries. "
+                           "Could not subscribe to dbus signals.")
```
ramnes, line 90: Why do we need that reconnect logic when it wasn't here before?

libqtile/utils.py
```diff
@@ -310,3 +314,84 @@ def scan_files(dirpath, *names):
...
+                                       signal_name=signal_name,
+                                       dbus_interface=dbus_interface,
+                                       bus_name=bus_name,
+                                       path=path)
+
+    return success
+
+
+async def remove_signal_handler(session_bus=False, signal_name=None,
+                                dbus_interface=None, bus_name=None, path=None):
```
ramnes, line 381: Looks like it isn't used anywhere?

libqtile/widget/mpris2widget.py
elParaguayo: It was there before. https://github.com/qtile/qtile/pull/1994/commits/a039b2e3fc494a37c9480eaf6181c7d1453c341f#diff-8606537352d6a3e9599d52fbe08da22e73f39c568bcbff00b56cef16416333c0L97

libqtile/utils.py
elParaguayo: Will fix.

libqtile/utils.py
elParaguayo: Correct. It's not currently used but if we're making a helper function to add the handler I thought it appropriate to provide a mechanism to remove it. python-dbus has `add_signal_receiver` and `remove_signal_receiver`.

The mistake I've made here is that I do need to pass the callback function as this needs to be removed too.

However, if you would prefer not to include it here, I can just delete it.

libqtile/utils.py
elParaguayo: My thinking here is just to have the `send_notification` make a dbus method call to `org.freedesktop.Notifications.Notify` as, that way, it should be able to work with any notification server that the user has running. Was hoping to sort tonight.

m-col: Isn't that what it does already? (Assuming `_can_notify` is true). Given that we're making `dbus_next` a hard dependency, i agree it makes sense to strip out that `_can_notify` import test.

elParaguayo: Yes, I think it does but it does it via gi.repository.Notify and we're catching the ImportError. Given the dependency, I can remove the import test completely and rework the notification to use dbus_next instead.

libqtile/utils.py
```diff
@@ -243,16 +234,41 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None):
...
+    except RuntimeError:
+        logger.warning("Eventloop has not started. Cannot send notification.")
+
+    return id
+
+
+async def _notify(title, message, urgency, timeout, id):
+    bus = await MessageBus(bus_type=BusType.SESSION).connect()
+
+    msg = Message(destination="org.freedesktop.Notifications",
```
elParaguayo, line 251: Just realised it might be worth refactoring this as it shares some code with the `_send_dbus_message` code for subscribing to signals. No time today to do it.

libqtile/widget/mpris2widget.py
ramnes: I meant in the python-dbus version of the code. :)

libqtile/utils.py
ramnes: Yeah, don't include dead code, we'll add it later if needed.

libqtile/widget/mpris2widget.py
ramnes: Thanks!

libqtile/widget/mpris2widget.py
elParaguayo: Fair point. Will remove.

libqtile/utils.py
elParaguayo: No problem. Will strip out.

ramnes (approved):
LGTM

libqtile/notify.py
```diff
@@ -22,63 +23,55 @@
...
 # SOFTWARE.
 
-"""
-    If dbus is available, this module implements a
-    org.freedesktop.Notifications service.
-"""
-from libqtile.log_utils import logger
+# type: ignore
+
+from dbus_next.aio import MessageBus  # noqa: I001
```
DorianGray, line 28: Instead of noqaing these, would it be possible to run `tox -e format`? It should automatically re-order the imports to make flake8 happy. Otherwise, they just need to be in alphabetical order I think.

DorianGray (approved):
just nits, looks great!

DorianGray:
Ah, looks like mpris needs asyncio import removed as well to make flake8 happy.

elParaguayo:
Sure. Can sort nits later.

One bigger question, looking at that last commit, I wonder if it's ok to create a new bus connection each time a notification is sent. The old code had `Notify.init` outside the `send_notification` method. However, I'm not sure I can do that here because the eventloop would likely not be running. Maybe it would be better to close the bus connection after each notification as I'm not sure if this is garbage collected or not (my knowledge of garbage collection in python is limited/non-existent).

This is different to the widgets that need dbus (Mpris2, KeyboardKbdd) as they're listening for signals so those connections need to be retained (and they need to be separate connections otherwise the messages would get sent to both widgets).

I might test this later today.

ramnes:
You could call this in `Manager._configure`, which is called once the eventloop is up and running.

Python's garbage collection is just reference counting, and dbus-next being pure-Python, we most likely don't need to do anything other than not holding a reference, unless we specifically want the bus connection to be closed right now.

DorianGray:
Is there a way to use a behind-the-scenes connection pool by chance?

elParaguayo:
@ramnes the reference count bit is what I find interesting here. Take the Mpris2 widget, it asks the bus to send it a copy of any `PropertiesChanged` signal from the media player which it then passes to a callback function. I don't do anything with the bus after adding the callback so I'm not sure how python knows to keep that object, which it must do as the widgets continue to get the callbacks.

As for the `Manager._configure` idea, that's good to know but probably makes the code a bit harder to follow.

I think we need to be careful to not share connections as that would cause issues for the widgets (they'd need to add code to check that messages received were meant to be handled by the widget). 

I don't know if this is a real issue or not. May only matter if there's something that uses `send_notification` a lot. Disconnecting the bus after sending is as simple as adding one line of code so I'm inclined to go that route!

@DorianGray Sorry, not sure what you mean by the pool.

elParaguayo:
Also worth adding that the Mpris2 widget has issues that aren't directly dbus related. It doesn't show text the first time you play something on the media player. It's to do with the logic used when reading the messages. I'm pretty sure this was an issue with the old widget too (not that I used it) so I'll pick this up separately.

DorianGray:
I looked into it(not deeply), we'd have to implement it ourselves using https://pypi.org/project/connection_pool/ probably

libqtile/notify.py
elParaguayo: I think these might have been ok. Neither tox or isort seemed to change them. I've removed the noqa.

libqtile/notify.py
```diff
@@ -22,63 +23,55 @@
...
-            pass
+    @signal()
+    def NotificationClosed(self, id_in, reason_in):  # noqa: N802
+        pass
 
-        @service.method(BUS_NAME, in_signature='', out_signature='ssss')
-        def GetServerInformation(self):  # noqa: N802
-            return ("qtile-notify-daemon", "qtile", "1.0", "1")
+    @method()
+    def GetServerInformation(self) -> 'ssss':  # noqa: N802, F821
```
ramnes, line 73: What is `'ssss'`? :D

elParaguayo: It's the output signature for that dbus method. `ssss` means that the method returns 4 strings (being `"qtile-notify-daemon"`, `"qtile"`, `"1.0"` and `"1"`).

While dbus_next's use of python's type hinting for the purposes of defining the signatures is pretty neat/clever. It does cause problems with mypy's type checking. Hence the need for extra `noqa`.

libqtile/notify.py
```diff
@@ -22,63 +23,55 @@
...
+
+    @method()
+    def CloseNotification(self, id: 'u'):  # noqa: N802, F821
+        pass
 
-        @service.signal(BUS_NAME, signature='uu')
-        def NotificationClosed(self, id_in, reason_in):  # noqa: N802
-            pass
+    @signal()
+    def NotificationClosed(self, id_in, reason_in):  # noqa: N802
```
elParaguayo, line 69: Looks like I've missed the type hinting on the arguments.

elParaguayo: Nope. Docs show signature only needed if the signal has an out argument.

elParaguayo:
Need to remove dbus section from this page: http://docs.qtile.org/en/latest/manual/hacking.html

Might then look to do some squashing of commits to tidy up this PR.

elParaguayo:
Some comments on the docs:

_docs/conf.py_
`dbus` and `dbus.mainloop.glib` are listed in `MOCK_MODULES`. I have no idea what this does so I've left as is for now but my suspicion is they can be removed. Do I need to add `dbus_next` here?

_docs/manual/hacking.rst_
I've removed the dbus errors section. I note that this did contain advice about how to install `requirements{-dev}.txt` but I've removed as the section was just on how to install `pygobject`. Also removed some unnecessary whitespace.

_docs/manual/install/funtoo.rst_
Left reference to `dbus` userflag. Don't know what it does but have assumed it's still relevant.

tych0:
> Do I need to add dbus_next here?

Yeah; this is how we do the auto-generation of the Foo.defaults docs.

>  Don't know what it does but have assumed it's still relevant.

This is probably an ebuild flag for enabling libdbus; since you're dropping the hard dependency on libdbus in this PR, I think you can drop that too. (but also the funtoo docs were last updated in 2016, so who knows if this even still works... https://www.funtoo.org/index.php?title=Package:Qtile&action=history )

elParaguayo:
> > Do I need to add dbus_next here?
> 
> Yeah; this is how we do the auto-generation of the Foo.defaults docs.

OK. Presumably I can remove `dbus` and `dbus.mainloop.glib` too as all those dependencies have been removed?

tych0:
Yep, I think so.

elParaguayo:
OK. So adding `dbus_next` to `docs\conf.py` breaks the docs build. Have I missed an import somewhere else?

libqtile/widget/mpris2widget.py
```diff
@@ -97,16 +101,26 @@ def update(self, interface_name, changed_properties, invalidated_properties):
...
-            self.displaytext = ' - '.join([
-                metadata.get(x)
-                if isinstance(metadata.get(x), dbus.String)
-                else ' + '.join(y for y in metadata.get(x) if isinstance(y, dbus.String))
-                for x in self.display_metadata if metadata.get(x)
-            ])
+
+            meta_list = []
+            for key in self.display_metadata:
+                val = metadata.get(key)
```
DorianGray, line 109: `val = getattr(metadata.get(key), 'value', None)` would save you 2 lines here but :shrug:

The only reason I even mention it is that these lines confused me on first read.

libqtile/widget/mpris2widget.py
```diff
@@ -97,16 +101,26 @@ def update(self, interface_name, changed_properties, invalidated_properties):
...
+                    meta_list.append(val)
+                elif isinstance(val, list):
+                    val = " - ".join((y for y in val if isinstance(y, str)))
+                    meta_list.append(val)
+
+            self.displaytext = ' - '.join(meta_list)
             self.displaytext.replace('\n', '')
 
         playbackstatus = changed_properties.get('PlaybackStatus')
+        if playbackstatus:
```
DorianGray, line 122: might be good to give this the same treatment.

libqtile/utils.py
```diff
@@ -239,16 +234,43 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None):
...
-        if id is None:
-            id = randint(10, 1000)
-        notifier.set_property('id', id)
-        notifier.show()
-        return id
+    id = randint(10, 1000) if id is None else id
+    urgency = 2 if urgent else 1
+
+    try:
+        asyncio.create_task(_notify(title, message, urgency, timeout, id))
```
DorianGray, line 241: You might want to `asyncio.get_running_loop()` here instead so if `create_task` throws an unrelated `RuntimeError` it isn't swallowed.

DorianGray (approved):
Just some nits really, looks great!

libqtile/widget/mpris2widget.py
elParaguayo: Yes - not the easiest bit to read. Thanks for the suggestion. Will fix.

libqtile/utils.py
elParaguayo: Interesting. `asyncio.get_running_loop()` also throw `RuntimeError` if there's no eventloop but  I guess your point is that it's calling an internal `asyncio` method so it only throws that error if there's no eventloop. 

Makes sense. Will tweak the code.

elParaguayo:
@DorianGray I've fixed the nits with the mpris2 widget. I've left most of the original logic in place as this PR is just for the dbus-next migration. However, I do want to look at how the widget responds to signals as I get the impression it could be better. I'll pick that up in a separate PR though.

elParaguayo:
Looks like the pypy test has got stuck...

DorianGray:
> Looks like the pypy test has got stuck...

Yeah, that's the hang that https://github.com/qtile/qtile/pull/2107 fixes :/ logger mutex and forking

tych0:
> Yeah, that's the hang that #2107 fixes :/ logger mutex and forking

Yeah, I haven't had a chance to read that one closely again since I last looked at it. Should have some time next week, though.

elParaguayo:
Wondering if it would be better to use `self.qtile.run_in_executor` instead of `asyncio.create_task`. Would look neater to use methods from `core.manager` as using `asyncio.create_task` has no direct link to our eventloop. 

Will test.

CHANGELOG
```diff
@@ -1,3 +1,9 @@
+Qtile X.X.X, released XXXX-XX-XX:
+    * !!! Dependency update !!!
+        - New dependency needs to be installed for qtile to work: dbus_next.
+          Replaces previous reliance on dbus/Glib and allows qtile to use async
```
tych0, line 4: dbus/glib were previously optional dependencies, but this series adds a requirement that dbus_next be installed, even for people who aren't using dbus at all for any widgets. Is there any chance we can make dbus_next optional?

elParaguayo: Apologies, I may have misremembered, I thought we said it was ok to add as a dependency given it was so light.

I don't have strong feelings one way or the other on this (although I would like Arch to put `dbus-next` in their main repository) so happy to rework if that's your preferred route.

tych0: Yep, so feel free to ignore me here :)

tych0: Although it might be nice to make it optional... I'll leave it up to you.

tych0: FWIW, we can merge this as-is just to get it in so, too. Looks like it just needs a rebase for the changelog?

elParaguayo: I agree. Probably politer to make it optional. Will update.

elParaguayo:
OK - I've made it optional. I've broken out these changes into separate commits to help review but I can squash them into their parents before we merge.

libqtile/notify.py
```diff
@@ -23,109 +23,128 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-# type: ignore
```
elParaguayo: I'd removed the blanket mypy ignore but the result is that we now need lots of `# type: ignore` lines. Might be neater to go back to ignoring the whole file.

libqtile/notify.py
```diff
@@ -23,109 +23,128 @@
...
-
-    def show(self, *args, **kwargs):
-        notif = Notification(*args, **kwargs)
-        return (notif, self.add(notif))
 
+        async def register(self, *args, **kwargs):
+            pass
 
-notifier = NotificationManager()
+    notifier = FakeManager()
```
elParaguayo, line 150: Just to explain this: The `Notify` widget imports `notifier` from this module. In addition, the `Notify` widget does not use `dbus-next`. So, instead of testing whether `dbus-next` is installed in the widget, we import a dummy object that allows the widget to be initialised and configured. A warning message is sent to the log file as well.

libqtile/notify.py
tych0: I think this is preferable -- might as well type check as much as we can.

libqtile/notify.py
tych0: Might be nice to see if there's another way we can do this (or at least add a comment somewhere with this text in it). However, I don't want to hold this PR up any longer, so we can do it separately.

elParaguayo: Sure - I'll take a look.

And thanks for merging - looks like you're on a mission today!

tych0:
I'm seeing some warnings when running the tests with this now:

```
test/widgets/test_widgetbox.py::test_widgetbox_widget[BareConfig]
  /home/tycho/packages/qtile/libqtile/widget/base.py:206: RuntimeWarning: coroutine '_Widget._config_async' was never awaited
    self.qtile.call_soon(asyncio.create_task, self._config_async())

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

elParaguayo:
Feels like I'm having a bad day now with all my PRs causing issues!

elParaguayo:
I suspect that warning is because there's no a coroutine in all widgets which may not get run depending on the test. For most widgets it doesn't matter because the `_config_async` method is essentially a no-op but we don't want the warnings.

I can run the tests locally and see which tests are impacted. I can then see how best to deal with this.

tych0:
> Feels like I'm having a bad day now with all my PRs causing issues!

Eh, shit happens. We'll get it fixed.

elParaguayo:
> I'm seeing some warnings when running the tests with this now:
> 
> ```
> test/widgets/test_widgetbox.py::test_widgetbox_widget[BareConfig]
>   /home/tycho/packages/qtile/libqtile/widget/base.py:206: RuntimeWarning: coroutine '_Widget._config_async' was never awaited
>     self.qtile.call_soon(asyncio.create_task, self._config_async())
> 
> -- Docs: https://docs.pytest.org/en/stable/warnings.html
> ```

I need to rework this test - it oversimplifies things and is unhappy that a coroutine is being called outside of the eventloop. Will take a look and send a new PR.

tcld:
I couldn't help but notice that `python-dbus-next` is another package I have to build myself. I assume the assumption is that it will become an official package in the near future?

elParaguayo:
Yes - that is the assumption. It's an optional dependency but it would be good to have in an official package.

tcld:
I haven't checked all the code, but it appears that a large number of other widgets have been changed in response to this. So I have to wonder: Is `dbus_next` really optional?

elParaguayo:
The only bits that have changed are where the code used `dbus` before. Two widgets used it (MPRIS2 and a keyboard widget) and the notification service used it. If `dbus` wasn't installed, those items wouldn't have worked. Now they don't work if `dbus-next` is not installed.

I don't think it's as significant as you fear.

tcld:
OK. :) Let's still hope that `dbus_next` quickly shows up in official repositories, otherwise those widgets might not be usable for many people.

elParaguayo:
I've submitted a request on the Arch forum to have it moved to the main repo.

tcld:
That sounds great.

I installed `dbus_next` this morning, but I am under the impression that the `Notify`-widget doesn't work anymore. (I got a few messages today, none of which showed up in the bar, like they usually do.) Am I right in assuming that this should "just work" without any further configuration?

tych0:
> I installed dbus_next this morning, but I am under the impression that the Notify-widget doesn't work anymore. (I got a few messages today, none of which showed up in the bar, like they usually do.) Am I right in assuming that this should "just work" without any further configuration?

Definitely, dbus-next should just be a drop in replacement for users, and shouldn't require any additional changes besides installing it.

elParaguayo:
Yes. Should work fine (I'd tested this multiple times). Anything in your logs?

elParaguayo:
OK - tracked this one down. Should be fixed in the above PR.

Weird one as not picked up in my tests - notifications displayed perfectly. Tested sending notifications via a different app while looking at dbus-monitor and saw the error. 

Good catch and thanks for flagging.

tcld:
Thank you for the quick investigation and solution! I'm looking forward to the merge (although I am a bit hindered at the moment since the current master crashes on my system).

PR #1998, opened 2020-12-07
Fix GmailChecker widget

elParaguayo:
GmailChecker widget is broken.

Issue is that current widget puts display format into a variable called "fmt" but this clashes with the internal "fmt" in _TextBox. That's not going to cause an issue in every instance but does here because the format string needs two variables.

This means that when the _TextBox tries to run [this line](https://github.com/qtile/qtile/blob/master/libqtile/widget/base.py#L350) you get an error because the format string needs two variables but only gets one.

We could also do away with `status_only_unseen` and just use named values e.g.
``` python
("display_fmt", "inbox[{total}],unseen[{unseen}]", "Display format"),
```

Happy to add that to the same pull request if you want.

### DIFF
--- a/libqtile/widget/gmail_checker.py
+++ b/libqtile/widget/gmail_checker.py
@@ -29,14 +29,17 @@ from libqtile.widget import base
 
 
 class GmailChecker(base.ThreadedPollText):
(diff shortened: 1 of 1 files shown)


tych0:
This is a great analysis, can you put it in the commit message?

PR #2002, opened 2020-12-10
Highlight urgent windows

Etn40ff:
Highlight urgent windows in the TreeTab layout.

### DIFF
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -263,6 +263,9 @@ class Window(TreeNode):
         if self.window is layout._focused:
             fg = layout.active_fg
             bg = layout.active_bg
+        elif self.window.urgent:
+            fg = layout.urgent_fg
+            bg = layout.urgent_bg
         else:
             fg = layout.inactive_fg
             bg = layout.inactive_bg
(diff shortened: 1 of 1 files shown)


tych0:
Can you add a test for this? Patch itself lgtm. (I wonder if we shouldn't have a more layout-generic way to do this...)

Etn40ff:
I am not sure which one is the right way to add tests in qtile. Can you point me to an example?

tych0:
I'd put it in test/layouts/test_treetab.py along side the other tests for this layout.

ramnes:
LGTM

> I wonder if we shouldn't have a more layout-generic way to do this...

So many things that should be more layout-generic... :)

ramnes (approved):

Etn40ff:
I had a look at  test/layouts/test_treetab.py but I did not find a way to test colors in a sensible way: there no property of the layout that one can look at easily. Any suggestion?

tych0:
Yeah, it doesn't look like it's super well organized for testing right
now.

It seems like we could take all the text/border color/format
calculation stuff and turn it into its own function, and then test
various arguments to that. That's not exactly a refactoring we should
ask you to do in this patch though. Unless you're interested of course
:)

Etn40ff:
In normal circumstances I would take care of it but right now there is way too much on my plate. Feel free to ignore my pull request if testing is a blocker.

tych0:
Nah, that's ok. We should really reorganize this code if we want people to test it. Thanks for the patch!

Issue #1685, opened 2020-04-27
Ampersands need to be escaped in WindowName widget


sffjunkie:
# Issue description
Ampersands in window names need to be changed in WindowName widget to "&amp;" before being passed to Pango/Cairo

# Qtile version
0.15.1

# Stack traces
Exception: parse_markup() failed for b'Seth Lakeman - King & Country'
2020-04-27 19:12:00,744 ERROR libqtile hook.py:fire():L373 Error in hook focus_change
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/libqtile/hook.py", line 371, in fire
    i(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/libqtile/widget/windowname.py", line 67, in update
    self.text = "%s%s" % (state, w.name if w and w.name else " ")
  File "/usr/lib/python3.8/site-packages/libqtile/widget/base.py", line 323, in text
    self.layout.text = self.formatted_text
  File "/usr/lib/python3.8/site-packages/libqtile/drawer.py", line 70, in text
    attrlist, value, accel_char = pangocffi.parse_markup(value)
  File "/usr/lib/python3.8/site-packages/libqtile/pangocffi.py", line 173, in parse_markup
    raise Exception("parse_markup() failed for %s" % value)
Exception: parse_markup() failed for b'Seth Lakeman - King & Country'

# Configuration
N/A

ramnes:
Fixed, thanks for raising this issue!

PR #2010, opened 2020-12-15
Add TVHeadend widget

elParaguayo:
Not sure how much of a demand there is for this but I thought I'd share as I like it!

It's a small icon widget that shows when your TVHeadend server is recording and also lists the upcoming recordings in a popup window.

tvh_widget

### DIFF
--- a/libqtile/widget/tvheadend.py
+++ b/libqtile/widget/tvheadend.py
@@ -1,0 +1,332 @@
+# Copyright (c) 2020 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import os
+from datetime import datetime
+from pathlib import Path
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+from libqtile import bar, images, pangocffi
+from libqtile.log_utils import logger
+from libqtile.popup import Popup
+from libqtile.widget import base
+
+
+def icon_path():
+    """Get the path to tv icon"""
+    dir_path = Path(__file__).resolve() / ".." / ".." / "resources" / "tvheadend-icons"
+    return str(dir_path.resolve())
+
+
+class TVHJobServer:
+
+    def __init__(self, host=None, auth=None, timeout=5):
+        self.host = host
+        self.auth = auth
+        self.timeout = timeout
+
+    def _send_api_request(self, path, args=None):
+        url = self.host + path
+
+        r = requests.post(url,
+                          data=args,
+                          auth=self.auth,
+                          timeout=self.timeout)
+        return r.json(strict=False)
+
+    def _tidy_prog(self, prog, uuid=None):
+
+        x = prog
+
+        return {"subtitle": x["disp_subtitle"],
+                "title": x["disp_title"],
+                "start_epoch": x["start"],
+                "stop_epoch": x["stop"],
+                "start": datetime.fromtimestamp(x["start"]),
+                "stop": datetime.fromtimestamp(x["stop"]),
+                "filename": x["filename"],
+                "basename": os.path.basename(x["filename"]),
+                "creator": x["creator"],
+                "channelname": x["channelname"],
+                "error": x["errorcode"],
+                "uuid": uuid if uuid else x["uuid"],
+                "duplicate": x.get("duplicate", 0) > 0}
+
+    def get_upcoming(self, path, hide_duplicates=True):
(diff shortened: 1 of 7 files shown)


libqtile/widget/tvheadend.py
```diff
@@ -0,0 +1,327 @@
...
+
+class TVHWidget(base._Widget, base.MarginMixin):
+    """
+    A widget to show whether a TVHeadend server is currently recording or not.
+
+    The widget will also show a popup displaying upcoming recordings.
+
+    NB if you use a username and password, these are stored in plain text. You
+    may therefore wish to create an unprivileged user account in TVHeadend that
+    only has access to scheduled recordings data.
```
tych0, line 99: Can you add a note here that this depends on the third party module requests? We've had complaints about this before...

libqtile/widget/tvheadend.py
```diff
@@ -0,0 +1,327 @@
...
+    """Get the path to tv icon"""
+    dir_path = Path(__file__).resolve() / ".." / ".." / "resources" / "tvheadend-icons"
+    return str(dir_path.resolve())
+
+
+class TVHTimeOut(Exception):
+    pass
+
+
+class TVHJobServer(object):
```
tych0, line 46: No need to inherit from object here.

libqtile/widget/tvheadend.py
```diff
@@ -0,0 +1,327 @@
...
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+from libqtile import bar, images
+from libqtile.log_utils import logger
+from libqtile.popup import Popup
+from libqtile.widget import base
+
+UPCOMING_RECORDINGS = "/dvr/entry/grid_upcoming"
```
tych0, line 33: Any chance this needs to be configurable?

libqtile/widget/tvheadend.py
```diff
@@ -0,0 +1,327 @@
...
+    def _send_api_request(self, path, args=None):
+        url = self.host + path
+        try:
+            r = requests.post(url,
+                              data=args,
+                              auth=self.auth,
+                              timeout=self.timeout)
+            return r.json(strict=False)
+        except (requests.exceptions.Timeout, requests.ConnectionError):
+            raise TVHTimeOut()
```
tych0, line 62: This re-raising swallows the cause for the original exception, can we just drop this?

libqtile/widget/tvheadend.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if not self.data:
+            lines.append("No upcoming recordings.")
+
+        else:
+            lines.append("Upcoming recordings:")
+            for rec in self.data:
+                lines.append(self.popup_format.format(**rec))
+
+        self.popup = Popup(self.qtile,
```
tych0, line 298: By my count this is the first in-tree use of the popup code, nice :)

tych0:
Overall looks good to me, modulo a few small things. Any chance there's a nice way to write one simple test? :)

elParaguayo:
Wow. Quick reviews @tych0 !

All perfectly valid points so let me update for them later.

As for the popup, it's fantastic. My Strava widget uses it too but that one's waaaay too messy to upload!

libqtile/widget/tvheadend.py
elParaguayo: Old habits die hard...

tych0:
You made a strava widget?! Sounds awesome :)

libqtile/widget/tvheadend.py
elParaguayo: _Shouldn't _be__ as it's behind the scenes and I don't think users can amend the path but I guess TVH could change their API so I can shift it to being configurable.

elParaguayo:
Yes, you can see the Strava icon (or my attempt at drawing it) in the screenshot next to the TV. Clicking on it shows my running stats. It can also display text in the bar. It only works for running but it could probably be made to work for cycling too.

elParaguayo:
As for writing a test, won't this fail as `requests` is not a required library?

tych0:
You can put it in requirements-dev.txt; everyone should have that when running the tests.

elParaguayo:
The test will only make sense if I can get the popup to work in the test. I've tried running the `test_popup.py` file locally with `pytest` but I get errors so I can't adapt it for my needs here. I get this message:
`ERROR test/test_popup.py::test_focus[BareConfig] - FileNotFoundError: [Errno 2] No such file or directory: 'Xvfb'`

Any ideas what I need to do to get that to work for me?

tych0:
Install Xvfb :-)

elParaguayo:
Ah. Assumed all packages would have been installed as part of dev requirements. Never heard of xvfb before. Learning lots as I go along...

elParaguayo:
Sorry. I must have missed a step. It looks like test failed because it couldn't import requests but I did include it in the `requirements-dev.txt` file.

tych0:
Huh, I thought tox would have installed that. Maybe try it with,

```
index 16fd6616..3fed157d 100644
--- a/tox.ini
+++ b/tox.ini
@@ -25,6 +25,7 @@ deps =
     setuptools >= 40.5.0
     xcffib >= 0.8.1
 commands =
+    pip install -r requirements-dev.txt
     python3 setup.py pytest --addopts "-Wall --cov libqtile --cov-report term-missing {posargs}"
 
 [testenv:packaging]
```

elParaguayo:
Thanks. That fixed it. Tests running successfully now.

tych0:
lgtm, although there's a small conflict; can you rebase and squash in the review changes?

elParaguayo:
Don't _think_ that check error is to do with my code.

tych0:
Yeah, agreed. We've seen this failure on master just today. Something changed recently and is causing a lot of issues...

elParaguayo:
Is there any way to re-run the tests?

tych0:
Yes, although the failure rate looks to be ~50%, so without a lot of babysitting I'm not sure it'll do much good.

elParaguayo:
Ah, ok. That's a pretty bad rate!

tych0:
Yeah, I'll probably wait another week or two to debug. Hopefully the cairo people figure it out by then :)

elParaguayo:
Added escaping of popup text to prevent pango parse_error (as per #1685)

tych0:
Can you rebase this? I think CI is fixed and the code itself lgtm.

elParaguayo:
Looks like my test has broken. Has something changed in the way tests are run?

elParaguayo:
My test seems to pass so I'm assuming something else is going wrong with the tests again.

elParaguayo:
Might be worth holding off merging this until #2072 is merged.

Currently, I have to create a Popup that's big and then shrink it to the necessary size as the Drawer isn't resized. 2072 should fix that.

elParaguayo:
The old version of this PR inlcuded `requests` in `requirements-dev.txt` as that was needed to run the test. It looks like `tox.ini` has changed slightly since then. Am I right in saying I should now just add `requests` to the `deps` section under `[testenv]`?

PR #2016, opened 2020-12-17
Wallpaper widget add a few functions. Please refer to CHANGELOG for details

tramhao:
1.Add update_interval in seconds.
2.Filter *.jpg, *.jpeg, *.png files only.
3.Scan the folder recursively.
I'm new to Python and qtile so please let me know if the code is all right. For me personally I'm using python 3.8 and it works.

### DIFF
--- a/libqtile/widget/wallpaper.py
+++ b/libqtile/widget/wallpaper.py
@@ -20,10 +20,10 @@
 #
 # To use this widget, you will need to install feh wallpaper changer
 
+import fnmatch
 import os
 import random
 import subprocess
-import fnmatch
 
 from libqtile import bar
 from libqtile.log_utils import logger
--- a/libqtile/widget/wallpaper.py
+++ b/libqtile/widget/wallpaper.py
@@ -23,13 +23,14 @@
 import os
 import random
 import subprocess
+import fnmatch
 
 from libqtile import bar
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
 
-class Wallpaper(base._TextBox):
+class Wallpaper(base.ThreadPoolText):
     defaults = [
         ("directory", "~/Pictures/wallpapers/", "Wallpaper Directory"),
         ("wallpaper", None, "Wallpaper"),
@@ -40,6 +41,7 @@ class Wallpaper(base._TextBox):
         ("label", None, "Use a fixed label instead of image name."),
         ("option", "fill", "How to fit the wallpaper when wallpaper_command is"
             "None. None, 'fill' or 'stretch'."),
+        ("update_interval", 5, "Update interval in seconds, if none, it won't update"),
     ]
 
     def __init__(self, **config):
@@ -53,10 +55,12 @@ class Wallpaper(base._TextBox):
 
         self.add_callbacks({'Button1': self.set_wallpaper})
 
-    def _configure(self, qtile, bar):
-        base._TextBox._configure(self, qtile, bar)
-        if not self.bar.screen.wallpaper:
-            self.set_wallpaper()
+    def poll(self):
+        """
+        Return the current wallpaper
+        """
+        self.set_wallpaper()
+        return self.text
 
     def get_path(self, file):
         return os.path.join(os.path.expanduser(self.directory), file)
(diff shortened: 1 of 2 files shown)


libqtile/widget/wallpaper.py
```diff
@@ -40,6 +41,7 @@ class Wallpaper(base._TextBox):
         ("label", None, "Use a fixed label instead of image name."),
         ("option", "fill", "How to fit the wallpaper when wallpaper_command is"
             "None. None, 'fill' or 'stretch'."),
+        ("update_interval", 5, "Update interval in seconds, if none, it won't update"),
```
tych0, line 44: Doesn't this change the widget to update every five seconds by default? That feels quite jarring. Can we make this `None` to match the old behavior?

libqtile/widget/wallpaper.py
```diff
@@ -53,22 +55,26 @@ def __init__(self, **config):
...
 
     def get_wallpapers(self):
         try:
             # get path of all files in the directory
-            self.images = list(
-                filter(os.path.isfile,
-                       map(self.get_path,
-                           os.listdir(
-                               os.path.expanduser(self.directory)))))
+            for root, dirnames, filenames in os.walk(os.path.expanduser(self.directory)):
```
tych0, line 71: What about svg, xcf, otf, etc? Doesn't this break people who were using those formats? I don't see any reason to change this behavior in this patch.

m-col: Similar scanning-folder-for-files logic has been written in `libqtile.images` (function `get_matching_files`). Perhaps we should move that more general logic into `libqtile.utils`, and then the widget can use that. Indeed, if whichever 'general' implementation returns something simple like a list of paths, then the widget only needs to accept a list of paths. No need to hard-code any extensions, as `libqtile.images.get_matching_files` already uses a regex, and the user can then decide what to use, and we can just log errors.

libqtile/widget/wallpaper.py
tramhao: That is because when you use this widget, you want to know if the auto change is working or not.

libqtile/widget/wallpaper.py
tramhao: For the extension filter, normally when we pull some wallpaper from website, there will be files like README, and it'll not be rendered, thus we need to filter them. svg, xcf are just vector files and of course you can use it as wallpaper, but 99% of people will just stick with jpg and png afaik.
For the common functions, I didn't know it exists...will check later.

tramhao: The libqtile.images module is not on documentation...
http://docs.qtile.org/en/latest/_modules/index.html
That's why I didn't know it.

tramhao: For get_matching_files,
after reading the code, currently it's used only by getting the icons. So maybe stick to my implementation for the widget now?

libqtile/widget/wallpaper.py
tych0: Can you write a test for that? I'd prefer to keep the widget's behavior as it is now.

tramhao: I'm sorry I don't know how to write a test. I didn't find any topic on writing tests on the website.
And a little complain here if you don't like you can ignore.
I saw the same widget was committed 2 years ago but didn't get merged. This is not how an open source project should be running.  I guess it's the same reason why openweather widget is on github, but not merged also, despite no one will use yahoo weather over open weather map. Anyway, as a irrelevant member of a small community, I don't have the right to criticize anything.

m-col: > I didn't find any topic on writing tests on the website.

Trying putting `python tests` into your favourite search engine, and consider reading some of the tests in our 'tests' folder. We use [pytest](https://docs.pytest.org/en/latest/) for testing.

>  saw the same widget was committed 2 years ago but didn't get merged.

Why wasn't it merged? Was it waiting for something? Feel free to post a comment on it if you want to draw attention to it.

> I guess it's the same reason why openweather widget is on github, but not merged

Looks to me like we're waiting for the submitter to respond. Sometimes people go for weeks/months without working on qtile, because it's a hobby project and people have lives. What would you have us do? Anyway that seems completely unrelated to your PR and writing a test.

> Anyway, as a irrelevant member of a small community, I don't have the right to criticize anything.

I don't think you are irrelevant, and indeed you do have the right to criticise anything we do, but your comment sounded more like frustration taken out on us for some reason that I cannot get from your rant. If you do have general problems with the way that Qtile is managed then please feel free to post an issue expressing your concerns and what we can do to improve.

libqtile/widget/wallpaper.py
m-col: > The libqtile.images module is not in the documentation

I think this is because it is used only internally, so isn't a 'user-facing' module. No probs though. I implemented the more general file-scanning utility function so you can hook into that, which might keep things simple. If you check out the `scan_files` function in `libqtile.utils` from PR #2022, you can see how to use it. What this would allow for is receiving a list of glob strings as arguments to the widget that would be used to find the appropriate image files. This could then replace the hard-coded `fnmatch.filter` lines so users can use whatever image type they use. Any thoughts? :)

m-col: That scan_files function has been merged so you could use it for your widget. It is [here](https://github.com/qtile/qtile/blob/master/libqtile/utils.py#L301)

PR #2024, opened 2020-12-19
fix commands not in subcommands

DorianGray:
```
14:39 < mcol> looks like qtile --version is broken
14:40 < mcol> perhaps we should test all the new commands :)
```

### DIFF
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -11,6 +11,14 @@ except (pkg_resources.DistributionNotFound, ImportError):
 
 
 def main():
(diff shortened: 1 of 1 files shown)


m-col:
LGTM! Probably best for @tych0 for go over this one as he put together this arg parsing recently.

tych0:
Derp derp, thanks!

tych0:
One point of note -- in the future can you put stuff like that IRC discussion in the commit message?

DorianGray:
Absolutely

PR #2025, opened 2020-12-19
Focus should not follow mouse across screens

m-col:
Currently, enter events on non-focussed screens cause the focus to shift
to that screen. This behaviour is similar in behaviour to what is
expected with the follow_mouse_focus config option, but happens in this
situation even when that is set to false. Instead, when
follow_mouse_focus is False, a different screen should only be focussed
upon a click of a window in that screen, or the root window within that
screen.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,7 @@ qtile X.X.X, released XXXX-XX-XX:
           Running `qtile` without arguments will continue to work for the
           forseeable future, but will be eventually deprecated. qtile prints a
           warning when run in this configuration.
+        - Qtile.cmd_focus_by_click is no longer an available command.
     * features
         - new WidgetBox widget
         - new restart and shutdown hooks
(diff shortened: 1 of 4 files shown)


m-col:
This also turns `Qtile.cmd_focus_by_click` into `Qtile._focus_by_click`. I'm not sure why that was an exposed command, as it is used exclusively by the following method (`Qtile.process_button_click`), and takes an `xcffib.xproto.KeyPressEvent` as an argument. If this was a mistake then I can change the name back.

tych0:
I would at least note it in the changelog in case someone was using it.

m-col:
Done.

tych0:
This seems like the sort of thing a test might be nice for :)

mikelward:
It looks like this broke the `follow_mouse_focus` case. https://github.com/qtile/qtile/issues/2098

PR #2029, opened 2020-12-20

tcld:
A completely different approach to #1870, since I wasn't able to iron out the remaining issues with that due to the nature of the approach.

This PR extends the Qtile-object by adding a few functions:
```python
change_layer(wid, up)  # move a client to the top or bottom of the stack
pin(wid, up)  # keep a client above or below all others
unpin(wid)  # undo the changes done by pin
```
The names of the functions (as well as their scope) are open for discussion, of course.

I'm sure this approach isn't perfect either, but it is much simpler and integrates better with X11.

On top of the advantages already mentioned, this should also allow picture-in-picture functionality with fake screens, since the window list is now global instead of group specific. :) 

TODO:
- [x] fix tests (no idea how to do that at the moment)
- [x] add some tests

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 
 import asyncio
+import enum
 import io
 import logging
 import os
@@ -79,6 +80,12 @@ def handle_exception(loop, context):
         logger.error(context["exception"], exc_info=True)
     else:
         logger.error("exception in event loop: %s", context)
+
+
+class LayerChange(enum.Enum):
+    Deny = -1
+    Auto = 0
+    Force = 1
 
 
 class Qtile(CommandObject):
@@ -463,7 +470,7 @@ class Qtile(CommandObject):
         windows = [wid for wid, c in self.windows_map.items() if c.group]
         self.core.update_client_order(windows)
 
-    def change_layer(self, wid, up=True, force=False):
+    def change_layer(self, wid, up=True, do=LayerChange.Auto):
         """Raise a window above its peers or move it below them, depending on 'up'.
         Raising a normal window will not lift it above pinned windows etc.
         """
@@ -504,7 +511,8 @@ class Qtile(CommandObject):
 
         for i, w in items:
             self.windows_map.move_to_end(i, last=up)
-        if force or order != list(self.windows_map.keys()):
+        if do == LayerChange.Force or \
+                (do == LayerChange.Auto and order != list(self.windows_map.keys())):
             client.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
             for i, w in items:
                 w.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
@@ -692,10 +700,7 @@ class Qtile(CommandObject):
                     c.window.net_wm_state_above = self.windows_map[parent].window.net_wm_state_above
                     c.window.net_wm_state_below = self.windows_map[parent].window.net_wm_state_below
                 self.update_client_list()
-                if assume_correct_layering:
-                    self.update_client_order()
-                else:
-                    self.change_layer(w.wid, force=True)
+                self.change_layer(w.wid, do=LayerChange.Deny if assume_correct_layering else LayerChange.Force)
                 hook.fire("client_managed", c)
             return c
         else:
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -652,7 +652,7 @@ class Qtile(CommandObject):
                 pass
         self.unmanage(window_id)
 
-    def manage(self, w):
+    def manage(self, w, assume_correct_layering=False):
         try:
             attrs = w.get_attributes()
             internal = w.get_property("QTILE_INTERNAL")
@@ -692,7 +692,10 @@ class Qtile(CommandObject):
                     c.window.net_wm_state_above = self.windows_map[parent].window.net_wm_state_above
                     c.window.net_wm_state_below = self.windows_map[parent].window.net_wm_state_below
                 self.update_client_list()
-                self.change_layer(w.wid, force=True)
+                if assume_correct_layering:
+                    self.update_client_order()
+                else:
+                    self.change_layer(w.wid, force=True)
                 hook.fire("client_managed", c)
             return c
         else:
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -31,10 +31,12 @@ import tempfile
 import time
 import traceback
 import warnings
+from collections import OrderedDict
 
 import xcffib
 import xcffib.xinerama
 import xcffib.xproto
+from xcffib.xproto import StackMode
 
 import libqtile
 from libqtile import command_interface, confreader, hook, utils, window
@@ -98,7 +100,7 @@ class Qtile(CommandObject):
         self.config = config
         libqtile.init(self)
 
-        self.windows_map = {}
+        self.windows_map = OrderedDict()
         self.widgets_map = {}
         self.groups_map = {}
         self.groups = []
@@ -444,13 +446,103 @@ class Qtile(CommandObject):
         self.core.update_net_desktops(self.groups, index)
 
     def update_client_list(self) -> None:
-        """Updates the client stack list
+        """Update X11 client list
 
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
         """
         windows = [wid for wid, c in self.windows_map.items() if c.group]
         self.core.update_client_list(windows)
+
+    def update_client_order(self) -> None:
+        """Update X11 client order
+
+        This is needed for third party tasklists and drag and drop of tabs in
+        chrome
+        """
+        windows = [wid for wid, c in self.windows_map.items() if c.group]
+        self.core.update_client_order(windows)
+
+    def change_layer(self, wid, up=True, force=False):
+        """Raise a window above its peers or move it below them, depending on 'up'.
+        Raising a normal window will not lift it above pinned windows etc.
+        """
+        client = self.windows_map.get(wid, None)
+        if client is None:
+            return
+        order = list(self.windows_map.keys())
+        self.windows_map.move_to_end(wid, last=up)
+
+        items = []
+        above = client.window.net_wm_state_above
+        below = client.window.net_wm_state_below
+        # we have three flags to check (up, above and below), so there are 8 combinations in total
+        if (up and above) or (not up and below):  # 2 out of 8
+            # if we're at the top, nobody needs to be raised above us; same in the other direction
+            pass
+        elif up and below:  # 2 out of 8 (ignore above)
+            # if we're at the bottom, almost everybody needs to be put above us
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if not w.window.net_wm_state_below
+            ]
+        elif not up and above:  # 2 out of 8 (ignore below)
+            # if we're at the top, almost everybody needs to be put below us
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if not w.window.net_wm_state_above
+            ]
+        elif not below and not above:  # 2 out of 8
+            # we're just a normal window; make sure we don't interfere with pinned windows
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if (w.window.net_wm_state_above if up else w.window.net_wm_state_below)
+            ]
+
+        if not up:
+            items = reversed(items)
+
+        for i, w in items:
+            self.windows_map.move_to_end(i, last=up)
+        if force or order != list(self.windows_map.keys()):
+            client.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
+            for i, w in items:
+                w.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
+            self.update_client_order()
+
+    def pin(self, wid, up=True):
+        """Pin a window to the top or the bottom of the stack, depending on 'up'."""
+        client = self.windows_map.get(wid, None)
+        if client is None:
+            return
+        order = list(self.windows_map.keys())
+        self.windows_map.move_to_end(wid, last=up)
+        client.window.net_wm_state_above = up
+        client.window.net_wm_state_below = not up
+        if order != list(self.windows_map.keys()):
+            client.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
+            self.update_client_order()
+
+    def unpin(self, wid):
+        client = self.windows_map[wid]
+        if client is None:
+            return
+        order = list(self.windows_map.keys())
+        items = []
+        if client.window.net_wm_state_above:
+            client.window.net_wm_state_above = False
+            items = [(i, w) for i, w in self.windows_map.items() if w.window.net_wm_state_above]
+            up = True
+        if client.window.net_wm_state_below:
+            client.window.net_wm_state_below = False
+            items = reversed([(i, w) for i, w in self.windows_map.items() if w.window.net_wm_state_below])
+            up = False
+        for i, w in items:
+            self.windows_map.move_to_end(i, last=up)
+        if order != list(self.windows_map.keys()):
+            for i, w in items:
+                w.window.configure(stackmode=StackMode.Above if up else StackMode.Below)
+            self.update_client_order()
 
     def add_group(self, name, layout=None, layouts=None, label=None):
         if name not in self.groups_map.keys():
@@ -595,7 +687,12 @@ class Qtile(CommandObject):
                 # Window may have been bound to a group in the hook.
                 if not c.group:
                     self.current_screen.group.add(c, focus=c.can_steal_focus())
+                parent = w.get_wm_transient_for()
+                if parent:
+                    c.window.net_wm_state_above = self.windows_map[parent].window.net_wm_state_above
+                    c.window.net_wm_state_below = self.windows_map[parent].window.net_wm_state_below
                 self.update_client_list()
+                self.change_layer(w.wid, force=True)
                 hook.fire("client_managed", c)
             return c
         else:
@@ -611,6 +708,7 @@ class Qtile(CommandObject):
                 c.group.remove(c)
             del self.windows_map[win]
             self.update_client_list()
+            self.update_client_order()
         if self.current_window is None:
             self.conn.fixup_focus()
 
@@ -739,11 +837,7 @@ class Qtile(CommandObject):
             wid = e.child
 
             if self.config.bring_front_click:
-                self.conn.conn.core.ConfigureWindow(
-                    wid,
-                    xcffib.xproto.ConfigWindow.StackMode,
-                    [xcffib.xproto.StackMode.Above]
-                )
+                self.change_layer(wid)
 
             window = self.windows_map.get(wid)
             try:
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1346,10 +1346,8 @@ class Window(_Window):
         self.toggle_minimize()
 
     def cmd_bring_to_front(self):
-        if self.floating:
-            self.qtile.change_layer(self.window.wid)
-        else:
-            self._reconfigure_floating()  # automatically above
+        self.qtile.change_layer(self.window.wid)
+        self.floating = True
 
     def cmd_keep_above(self):
         if self.window.net_wm_state_above:
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -23,7 +23,7 @@ import inspect
 import traceback
 
 import xcffib.xproto
-from xcffib.xproto import EventMask, SetMode, StackMode
+from xcffib.xproto import EventMask, SetMode
 
 from libqtile import hook, utils
 from libqtile.command_object import CommandError, CommandObject
@@ -481,11 +481,11 @@ class _Window(CommandObject):
             width=width,
             height=height,
         )
-        if above:
-            kwarg['stackmode'] = StackMode.Above
 
         self.window.configure(**kwarg)
         self.paint_borders(bordercolor, borderwidth)
+        if above:
+            self.qtile.change_layer(self.window.wid)
 
         if send_notify:
             self.send_configure_notify(x, y, width, height)
@@ -962,7 +962,7 @@ class Window(_Window):
             screen.group.add(self, force=True)
             self.qtile.focus_screen(screen.index)
 
-        self._reconfigure_floating()
+        self._reconfigure_floating(above=False)
 
     def getsize(self):
         return (self.width, self.height)
@@ -970,7 +970,7 @@ class Window(_Window):
     def getposition(self):
         return (self.x, self.y)
 
-    def _reconfigure_floating(self, new_float_state=FLOATING):
+    def _reconfigure_floating(self, new_float_state=FLOATING, above=True):
         if new_float_state == MINIMIZED:
             self.state = IconicState
             self.hide()
@@ -995,7 +995,7 @@ class Window(_Window):
                 width, height,
                 self.borderwidth,
                 self.bordercolor,
-                above=True,
+                above=above,
             )
         if self._float_state != new_float_state:
             self._float_state = new_float_state
@@ -1347,9 +1347,9 @@ class Window(_Window):
 
     def cmd_bring_to_front(self):
         if self.floating:
-            self.window.configure(stackmode=StackMode.Above)
+            self.qtile.change_layer(self.window.wid)
         else:
-            self._reconfigure_floating()  # atomatically above
+            self._reconfigure_floating()  # automatically above
 
     def cmd_match(self, *args, **kwargs):
         return self.match(*args, **kwargs)
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1379,9 +1379,9 @@ class Window(_Window):
         return (window.edges[0] <= x <= window.edges[2] and
                 window.edges[1] <= y <= window.edges[3])
 
-    def cmd_set_position(self, dx, dy):
+    def cmd_set_position(self, x, y):
         if self.floating:
-            self.tweak_float(dx, dy)
+            self.tweak_float(x, y)
             return
         for window in self.group.windows:
             if window == self or window.floating:
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1351,6 +1351,24 @@ class Window(_Window):
         else:
             self._reconfigure_floating()  # automatically above
 
+    def cmd_keep_above(self):
+        if self.window.net_wm_state_above:
+            self.qtile.unpin(self.window.wid)
+        else:
+            self.qtile.pin(self.window.wid)
+
+    def cmd_keep_below(self):
+        if self.window.net_wm_state_below:
+            self.qtile.unpin(self.window.wid)
+        else:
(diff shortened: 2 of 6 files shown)


tcld:
I am having some serious trouble with the failing tests - any hints would be appreciated:

1. `test_popup_focus` (*old*): This popup creates a window that it passes its "fake" qtile-object to, and then the window tries to use it due to a change I made in `def _Window.place()`. This leads to a crash, since the fake object isn't up to it.
2. `test_layered_max_window_focus_cycle` (*new*): During this test a window is removed from a `Max`-layout, after which two windows ('one' and 'two') are left. Directly after the removal either one of the windows can receive an `EnterNotify`, which leads to a differing focus chain. So sometimes the test fails, sometimes it passes.

While 1 isn't directly related to the code in this PR, it still needs to be fixed alongside of it. 2, however, seems to be related to the PR, but I really don't understand where the erratic results are coming from.

tcld:
I removed the commits in question, both were about #415. I think that should be easy enough to add now by laying out the clients in order from top (possibly focussed) to bottom, calling `lower_client` on each of them. But I would rather try my hands at that at a later point, since the actual layering code is just a tool for that.

That leaves only one failing test: `test_popup_focus` - mcol mentioned on IRC that he would take care of that, which is great, since I still have no clue. :)

libqtile/core/manager.py
```diff
@@ -444,14 +446,96 @@ def update_net_desktops(self) -> None:
...
+            items = [(i, w) for i, w in self.windows_map.items() if not w.window.net_wm_state_below]
+        elif not up and client.window.net_wm_state_above:  # 2 out of 8 (ignore net_wm_state_below)
+            # if we're at the top, almost everybody needs to be put below us
+            items = [(i, w) for i, w in self.windows_map.items() if not w.window.net_wm_state_above]
+        elif not client.window.net_wm_state_below and not client.window.net_wm_state_above:  # 2 out of 8
+            # we're just a normal window; make sure we don't interfere with pinned windows
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if (w.window.net_wm_state_above if up else w.window.net_wm_state_below)
+            ]
```
m-col, line 492: Any chance you could reformat this a bit? It's quite hard to read, and the lines are very long :)

m-col:
Summarising our discussion on IRC:

- I'm getting strange behaviour when dragging a window: window movement is very delayed and laggy, as if some intense computation is happening upon every call of Qtile.process_button_motion. @tcld cannot reproduce, deepening this mystery. I get this even on the default config, and it happens with or without picom. ([Here](https://f.mcol.xyz/1608742209.mp4) is a video - github was giving me errors putting it here directly).
- I mentioned that doing `lower_client()` on a window that has had `keep_above()` enabled will send it back, but not disable the `keep_above()`. We agree this behaviour is useful in case one has multiple pinned windows.

Other than the formatting comment I made and the drag issue I'm getting, I think this implementation is super nice, and its interface is simple :)

As you mention, I can deal with the Popup test when the main code is done.

libqtile/core/manager.py
m-col: Looks like there might be some repeat calls to `client.window.X` getters that could be prevented by storing the result in a variable

m-col:
Here's a patch for test_popup:
```
diff --git a/test/test_popup.py b/test/test_popup.py
index b4b09088..6576ebef 100644
--- a/test/test_popup.py
+++ b/test/test_popup.py
@@ -20,6 +20,7 @@
 
 
 import pytest
+from xcffib.xproto import StackMode
 
 from libqtile.backend.x11 import xcbq
 from libqtile.popup import Popup
@@ -29,13 +30,20 @@ from test.conftest import BareConfig
 @pytest.mark.parametrize("qtile", [BareConfig], indirect=True)
 def test_popup_focus(qtile):
     qtile.test_xeyes()
-    qtile.windows_map = {}
 
-    # we have to add .conn so that Popup thinks this is libqtile.qtile
-    qtile.conn = xcbq.Connection(qtile.display)
+    class FakeQtile:
+        def __init__(self):
+            self.conn = xcbq.Connection(qtile.display)
+            self.windows_map = {}
+
+        def change_layer(self, wid):
+            mask, values = xcbq.ConfigureMasks(stackmode=StackMode.Above)
+            self.conn.conn.core.ConfigureWindow(wid, mask, values)
+
+    fake = FakeQtile()
 
     try:
-        popup = Popup(qtile)
+        popup = Popup(fake)
         popup.width = qtile.c.screen.info()["width"]
         popup.height = qtile.c.screen.info()["height"]
         popup.place()
@@ -46,4 +54,4 @@ def test_popup_focus(qtile):
         popup.hide()
     finally:
         popup.kill()
-        qtile.conn.finalize()
+        fake.conn.finalize()
```

libqtile/core/manager.py
tcld: Are those expensive? I guess you are talking about the `client.window.net_wm_state`, not the stuff in the list comprehensions, right?

m-col: They aren't particularly expensive, so it's fine. Just figured that it might also make it less verbose a bit. Perhaps you could do two levels of `if`s i.e. `if up:` and break it down. I know this is being picky, sorry! It will just be helpful in the future when glancing through.

m-col: If there is no improvement then it's fine as is.

tcld: I tried to improve it a bit. Pulling the cheks out of the `if` helped quite a bit, I think. Let me know what you think.

m-col:
Could you squash those last 3 commits (and the popup test patch) into the first commit? (Or all into one)

libqtile/core/manager.py
```diff
@@ -444,14 +446,104 @@ def update_net_desktops(self) -> None:
...
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if not w.window.net_wm_state_above
+            ]
+        elif not below and not above:  # 2 out of 8
+            # we're just a normal window; make sure we don't interfere with pinned windows
+            items = [
+                (i, w) for i, w in self.windows_map.items()
+                if (w.window.net_wm_state_above if up else w.window.net_wm_state_below)
+            ]
```
m-col, line 500: Yeah this is much more readable. Thanks for that!

libqtile/scratchpad.py
```diff
@@ -285,6 +287,7 @@ def on_float_change(self, *args, **kwargs):
         for name, dd in self.dropdowns.items():
             if not dd.window.floating:
                 if dd.window.group is not self:
+                    dd.window.qtile.unpin(dd.window.window.wid)
```
m-col, line 290: What's this for?

libqtile/window.py
```diff
@@ -1347,9 +1347,27 @@ def cmd_toggle_minimize(self):
 
     def cmd_bring_to_front(self):
         if self.floating:
-            self.window.configure(stackmode=StackMode.Above)
+            self.qtile.change_layer(self.window.wid)
+        else:
+            self._reconfigure_floating()  # automatically above
```
m-col, line 1352: If `self.qtile.change_layer(self.window.wid)` was called when the window is _not_ floating, would it have the same effect?

libqtile/scratchpad.py
tcld: I decided to "pin" Scratchpads to the top; that (and the other change in that file) basically from the other z-PR, where it was because you at some point mentioned that scratchpads didn't work correctly.

I should test whether that line is even necessary!

libqtile/window.py
tcld: It would still raise the window. But `_reconfigure_floating` also *makes* the window floating (which I find weird, but people might expect it by now).

tcld: If `bring_front_click` is `True`, it shows a different behaviour, since it does **not** make windows floating. Maybe this would be a good point to unify that behaviour and just drop a few lines of code here?

m-col: That seems like odd behaviour, but if that's the case, then fair enough. Maybe we should make that more explicit in the code by changing this to:
```
    def cmd_bring_to_front(self):
        self.qtile.change_layer(self.window.wid)
        self.floating = True
```

libqtile/scratchpad.py
tcld: I removed the changes in this file, since scratchpads seem to work fine without them. Might have to think about this, though, since I didn't expect that.

libqtile/window.py
tcld: I would like that, but calling `_reconfigure_floating` also could change the dimensions of the window.

m-col: I was just trying to figure out some of the floating stuff, and it seems that the "proper" way to trigger floating is to use `_enablefloating`, which is used exclusively by some of the setters (floating, maximised, fullscreen). I think these setters are therefore the "proper" way to change those settings. `self.floating = True` uses the dimensions in `self._float_info`, which I reckon are the "proper" dimensions to use when triggering floating, so it seems that using `_reconfigure_floating` here to enable floating was an oversight. I'd vote for using `self.floating = True`, but interested to hear other comments in case I've misinterpreted things here.

m-col (approved):
Just playing around with it, and it works beautifully!

Thank you very much for this work! LGTM.

Please could you confirm which issues would be fixed by this, and which not?

libqtile/window.py
tcld: I've added it in an extra commit. Let's see how it fares.

m-col: 👍🏽

tcld:
Closes #753 (mentioned in the first commit message).

Leaves #415 open (but provides a basis to solve it), enables #1260 (but does not explicitly address it).

No idea what it does for #676 and #1145.

m-col:
I think this PR would be sufficient to close #1260 (if it wasn't already closed) and #1145. #1145 can be resolved in the config, using the `client_new` hook to call `cmd_keep_above` on a tint2 window. Something similar could presumably be used for #676 too. So great! :D

tcld:
Ok, I guess #1145 could be closed with this, but pinned windows will likely still be on top of Tint - but since the creator of the issue didn't want to use a `Gap`, that is probably by design.

#1260 can be achieved, but if somebody "pins" a tiled window, then he is back to square one, since it can overlay floating windows again. So I would say this provides a workaround, but it isn't a full solution.

As for #676, my impression was that it kind of fixed itself. Again, this might provide a workaround for such a situation, but not a true solution, since manually pinned windows and automatically pinned ones are of equal importance. Maybe a few more of the X-flags besides `_NET_WM_STATE_ABOVE`/`BELOW` could be used to simulate more layers, but I don't really like that idea, and it would have to be done in a future PR anyway.

ramnes:
I quickly tried the branch, and I can confirm that it does not fix #415, #676, nor #1260. For the most part I didn't see any difference with the current implementation, which is great.

But then I tried to trigger a restart by unplugging my second monitor while i3lock is active; I already found out a few months ago than doing so would make the bar visible, but still hide everything else (yeah, I should open an issue for this). With that branch, everything becomes visible and i3lock vanishes (while still catching all input), which is really not good. :(

libqtile/core/manager.py
```diff
@@ -595,7 +687,12 @@ def manage(self, w):
                 # Window may have been bound to a group in the hook.
                 if not c.group:
                     self.current_screen.group.add(c, focus=c.can_steal_focus())
+                parent = w.get_wm_transient_for()
+                if parent:
+                    c.window.net_wm_state_above = self.windows_map[parent].window.net_wm_state_above
+                    c.window.net_wm_state_below = self.windows_map[parent].window.net_wm_state_below
                 self.update_client_list()
+                self.change_layer(w.wid, force=True)
```
ramnes, line 695: This is the faulty line for the behavior I'm seeing when I restart under i3lock. What's the rationale behind it?

tcld:
Thanks for testing!

I am working on a separate PR to address #415, but I'm not yet satisfied with the solution. The other two issues aren't really addressed, as I mentioned, but they could be worked around with this.

Do you think we could create a test to simulate the situation you described?

libqtile/core/manager.py
tcld: Once a new window is added, it has to be moved to its place in the stack, otherwise we might violate the "pinned" state of other windows or even fall out of sync with the stacking order we report to X11.

Do you know whether i3lock has _NET_WM_STATE_ABOVE or anything similar set?

ramnes:
It can be triggered very easily actually: `i3lock & qtile cmd-obj -o cmd -f restart`. And I just figured out that i3lock doesn't vanish away, it's just sent below every other window.

Screenshot from 2020-12-28 14-21-09

So yeah, I guess we could spawn any screen locker, call `cmd_restart`, and have a look at the stack.

libqtile/core/manager.py
ramnes: It looks like it doesn't set anything but WM_NAME? And most importantly we don't manage its window (it's not listed in `qtile.windows_map` when running).

tcld: If so, then that likely means that layering *any* window will lead to the unmanaged windows falling by the wayside? We could maybe work around that by using `StackMode.Below` exclusively (if nothing is ever raised, the unmnaged windows won't be buried), but that feels like a bit of a hack.

tcld:
I was trying to figure out why certain windows aren't managed by Qtile, and the search seemingly lead into `xcore._xpoll()`. But I have no idea where to go from there or whether that's even the right place.

The assumption, however, stands: Unmanaged windows would be buried by any calls to `xcbq.Window.configure(stackmode=StackMode.Above)`, since nobody takes care of their place in the stack. For short-lived windows that really shouldn't matter (if they were spawned "above"), but I had no idea that even lockscreens could fall into that category - who knows what else?

The solution seems to be managing all windows, at least in a rudimentary way. I don't even understand yet what's causing some windows to be managed and others not. Going even further, I would assume that this issue has always existed, but currently the calls that actually raise a window are limited.

I'm officially stuck now.

libqtile/core/manager.py
tcld: After adding that most recent commit, the behaviour isn't perfect, but it is back to how the master-branch is currently behaving (which I just now noticed). So unless the final commit is too ugly, I would suggest that to be tackled in a separate PR, likely by someone more knowledgable than me.

tcld:
Would it be ok to explicitly check for unmanaged windows and raise them (after a restart or even always)? This is of course assuming that we are able to identify such windows. And I'm not sure we are, since we don't tell X about them through `_NET_CLIENT_LIST`.

ramnes:
I don't know why we don't manage all windows in the first place. @tych0 any idea?

ramnes:
If we merge #2122, do you think we'll be able to implement it in a reasonable way here? Or are we always going to draw floating windows over tiled ones?

tcld:
What #2122 does is check the type of window before raising it. I don't think that would interfere with this PR at all; a tiled window could still be raised or pinned using one of the methods this PR provides (above floating windows), but I think those mechanisms complement each other nicely.

I'll have to rebase this, of course, but that's necessary anyway since I haven't done that in almost two weeks.

I am mostly waiting for a resolution to the question of how/whether to handle unmanaged windows; there is a bit of randomness after restarts as long as not all windows are layered.

senthilbaboo:
IMO, this seems to be re-inventing the wheel. The stacking order is already taken care by X server. The  isssue we have is that we are messing up the stacking order by calling ConfigureWindow with StackMode.Above, whenever we want to bring windows to the front. This is bad. All we have to do is to replace such calls with restack calls, and only bring windows on top in relation to the window in the same group and leave other windows as it is. And during restart, don't mess the stack order and everything will be fine.

senthilbaboo:
Also for pinning windows, rather than introducing a new API, I think we should try implement EWMH correctly. In this particular case, if a window has _NET_WM_STATE_ABOVE, it should be above all other windows.

tcld:
I definitely don't know too much about X, so I can't fully evaluate what you are saying. Could you back it up with some links? From all I have seen so far, we *tell* X what the stacking order is, not the other way around.

ramnes:
Yeah I'd love to read more about this as well, please show us the way to X11 holiness!

Also, thank you for reminding us about EMWH, we should definitely work on implementing it better, and it starts here.

senthilbaboo:
Disclaimer:- I am myself not a guru in X11.

X itself maintains a stacking order of windows. We can query the stacking order using [XQueryTree](https://tronche.com/gui/x/xlib/window-information/XQueryTree.html) on the root window. We are already doing this in qtile in xcbq.query_tree to scan for existing windows.

We can restack windows using [XRestackWindows](https://tronche.com/gui/x/xlib/window/XRestackWindows.html). Note, to restack, we don't need to specify all the windows, we just need to specify a sibling window, in relation to which, we would like to restack. I did not find a restack API with XCB, but we could achieve restacking using xcb_configure_window. Here is a [stack overflow answer](https://stackoverflow.com/questions/60612753/how-to-restack-all-windows-properly-in-xcb-like-xrestackwindows-in-xlib) that I found which is used to restack a window in relation to another window.

Let's say, we want to bring a window to the top, we are currently making a xcb_configure_window call without specifying a sibling. Rather we should find the window that is on top in the current group (we can call query_tree for this and we can also cache) and raise the window in relation to the top window, instead of raising it all the way up in the stack (which is what causes issues when we restart qtile, IMO). Also, during restart, we don't have to restack, as the stacking order should not change between restarts.

There are a few layering rules that we might have to implement in the window manager. The layering rule should be something like (windows with _NET_WM_STATE_BELOW, tiled windows, floating window, followed by windows that have _NET_WM_STATE_ABOVE. But IMO, we don't have to maintain the stacking order in the WM itself and leave it to be maintained by the X server.

And windows with override_redirect should not be managed by window manager and if I am right, if we don't alter absolute stacking order window, windows with override redirect, will place themselves correctly in the stack.

tcld:
> X itself maintains a stacking order of windows.

I am fairly sure that's wrong. X provides a list of windows in their current order, but the wm is the one that needs to provide that list - I remember experimenting with this in the past by raising windows through `StackMode.Above`, but that list never changed. This is why [this code](https://github.com/qtile/qtile/pull/2029/files#diff-7a2ab1986389e2f2762fd6089b3e8008bfe2a11d5ba11e27b978126d9455a5d2R391) exists. But I will have another look at `QueryTree` to be sure. Maybe raising relative to siblings has an influence on things, too.

> Note, to restack, we don't need to specify all the windows, we just need to specify a sibling window, in relation to which, we would like to restack.

The StackOverflow-answer you linked to does something similar to what this PR is doing when it comes to layering; the sibling-part I don't fully get yet, though. I'll have to read up on that a bit - it might make calls to the layering-code cheaper and help simplify the code a bit, which would be great.
By the way, I don't like the idea of just looking at windows within a group, since there are scenarios where windows from other groups (or outside of groups) interfere sufficiently to be taken into account, too.

> The layering rule should be something like (windows with _NET_WM_STATE_BELOW, tiled windows, floating window, followed by windows that have _NET_WM_STATE_ABOVE.

There is another, now closed PR about layering windows where that approach was taken (even a bit more granular by taking other flags into account), whereas this PR only distinguishes between below, "normal" and above to leave control with the user. Maybe we can get closer to [these specifications](https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER) again by making use of the sibling-functionality.

> And windows with override_redirect should not be managed by window manager and if I am right, if we don't alter absolute stacking order window, windows with override redirect, will place themselves correctly in the stack.

That could be an extremely valuable point, since it addresses the last big issue I was having with this PR.

I hope that wasn't too critical an answer, I am very grateful for any input I can get. I'll try to incorporate as much of it as possible; and if it turns out that X can keep track of the layering entirely on its own, then that's all the better for us. :)

There is one more thing from before that I have to add in:
> Also for pinning windows, rather than introducing a new API, I think we should try implement EWMH correctly.

It's probably a good idea, but I don't really see how that get's rid of the layering methods. From all I know so far, the window manager provides and honors (or ignores) those hints, so wouldn't the methods introduced here just become more complicated instead of going away?

senthilbaboo:
> > X itself maintains a stacking order of windows.
> 
> I am fairly sure that's wrong. X provides a list of windows in their current order, but the wm is the one that needs to provide that list - I remember experimenting with this in the past by raising windows through `StackMode.Above`, but that list never changed. This is why [this code](https://github.com/qtile/qtile/pull/2029/files#diff-7a2ab1986389e2f2762fd6089b3e8008bfe2a11d5ba11e27b978126d9455a5d2R391) exists. But I will have another look at `QueryTree` to be sure. Maybe raising relative to siblings has an influence on things, too.
> 

I am not sure what do you mean the list does not change. I wrote a test to check if that list changes in one of my PR, to bring the window in front after clicking [here](https://github.com/qtile/qtile/pull/2122/commits/9b8043635a0730ff26489dc927432bbfa7b6c206). That list or tree kept in X server, is the ultimate source of truth. If that does not change, my test would fail terribly.

> > Note, to restack, we don't need to specify all the windows, we just need to specify a sibling window, in relation to which, we would like to restack.
> 
> The StackOverflow-answer you linked to does something similar to what this PR is doing when it comes to layering; the sibling-part I don't fully get yet, though. I'll have to read up on that a bit - it might make calls to the layering-code cheaper and help simplify the code a bit, which would be great.
> By the way, I don't like the idea of just looking at windows within a group, since there are scenarios where windows from other groups (or outside of groups) interfere sufficiently to be taken into account, too.
> 
We either can maintain a stacking order within a group or for all the windows that we have. Both can be done, without us explicitly maintaining a data structure for the same. As I said, it is already maintained by X.
> > The layering rule should be something like (windows with _NET_WM_STATE_BELOW, tiled windows, floating window, followed by windows that have _NET_WM_STATE_ABOVE.
> 
> There is another, now closed PR about layering windows where that approach was taken (even a bit more granular by taking other flags into account), whereas this PR only distinguishes between below, "normal" and above to leave control with the user. Maybe we can get closer to [these specifications](https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER) again by making use of the sibling-functionality.
> 
> > And windows with override_redirect should not be managed by window manager and if I am right, if we don't alter absolute stacking order window, windows with override redirect, will place themselves correctly in the stack.
> 
> That could be an extremely valuable point, since it addresses the last big issue I was having with this PR.
> 
> I hope that wasn't too critical an answer, I am very grateful for any input I can get. I'll try to incorporate as much of it as possible; and if it turns out that X can keep track of the layering entirely on its own, then that's all the better for us. :)
> 
> There is one more thing from before that I have to add in:
> 
> > Also for pinning windows, rather than introducing a new API, I think we should try implement EWMH correctly.
> 
> It's probably a good idea, but I don't really see how that get's rid of the layering methods. From all I know so far, the window manager provides and honors (or ignores) those hints, so wouldn't the methods introduced here just become more complicated instead of going away?

senthilbaboo:
Sorry, hit send, before I completed.
> > Also for pinning windows, rather than introducing a new API, I think we should try implement EWMH correctly.
> 
> It's probably a good idea, but I don't really see how that get's rid of the layering methods. From all I know so far, the window manager provides and honors (or ignores) those hints, so wouldn't the methods introduced here just become more complicated instead of going away?

Yes, window manager is the one which has to respect the hints and stack the windows correctly. What all I am saying is that we don't have to maintain additional data structures for stacking, we can always get the current stack order from X and determine the stack order needed, based on hints that exists on the windows and their states (floating/tiled) etc.

tcld:
> I wrote a test to check if that list changes in one of my PR, to bring the window in front after clicking here. That list or tree kept in X server, is the ultimate source of truth.

Then the code [here](https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/core.py#L390) should have never been needed in the first place? If I understand this correctly, our interfering with that list would have undermined X's ability to properly keep track of the layers. Unless X completely repopulates the list itself, i.e. ignoring us. But if it did, then why would we even have to tell X about what windows we currently know of? Shouldn't it already know?

This certainly leads to a lot of questions on my end, and I am a bit terrified of learning more about X. But your input is very much appreciated!

I really like your test, by the way. Writing tests often feels very cumbersome to me, and your code gives me a lot of ideas for the future.

senthilbaboo:
> > I wrote a test to check if that list changes in one of my PR, to bring the window in front after clicking here. That list or tree kept in X server, is the ultimate source of truth.
> 
> Then the code [here](https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/core.py#L390) should have never been needed in the first place? If I understand this correctly, our interfering with that list would have undermined X's ability to properly keep track of the layers. Unless X completely repopulates the list itself, i.e. ignoring us. But if it did, then why would we even have to tell X about what windows we currently know of? Shouldn't it already know?
> 
X knows about all the windows. The property that we set here is never read by X. But is read by other third party application (like chrome) to find other chrome windows that are currently open (I guess). Also this list would only contain windows that are managed by WM.

> This certainly leads to a lot of questions on my end, and I am a bit terrified of learning more about X. But your input is very much appreciated!
> 
For this particular case, IMO, the issue we have is that we are messing up with the stacking order too much. We should never call configure with just StackMode.Above without specifying a sibling.

bring_to_front should be something like this:
1. get the current top window that is visible in the screen (this should look into the state of window, attributes of window etc).
2. call configure with StackMode.Above but with the sibling as we determined in the previous step.

Right now, I see the floating windows (when I drag) hides the bar, which is usually not a behavior we would expect and this happens because we send that window all the way to the top.

Also regarding windows showing up when the screen is locked and qtile restarts, happens exactly because of the same reason. Some lock screen application would not allow that and re place themselves on top of the stack, but window manager should not keep messing up with the stack.

> I really like your test, by the way. Writing tests often feels very cumbersome to me, and your code gives me a lot of ideas for the future.
Thanks!

tcld:
> X knows about all the windows. The property that we set here is never read by X. But is read by other third party application (like chrome) to find other chrome windows that are currently open (I guess). Also this list would only contain windows that are managed by WM.

So then we *do* have to keep our own data structure to keep X "informed" for certain applications to work properly.

I'll think about all this and put some more work into this PR in the near future.

senthilbaboo:
I am still not convinced maintaining our own data structure to track window stacking.

[Here](https://github.com/i3/i3/blob/dcd6079c9bb5e06716a53b90f8539715bb194eeb/src/x.c#L1228), we can see how i3 does it. If you look closely they don't keep track of the stack by themselves at all, but only rely of X. I have other issues with i3's internal design. But this bit, they got right.

Keeping track of stacking order by ourselves will only make the design complicated and will become hard to maintain.  My two cents.

senthilbaboo:
> I am still not convinced maintaining our own data structure to track window stacking.
> 
> [Here](https://github.com/i3/i3/blob/dcd6079c9bb5e06716a53b90f8539715bb194eeb/src/x.c#L1228), we can see how i3 does it. If you look closely they don't keep track of the stack by themselves at all, but only rely of X. I have other issues with i3's internal design. But this bit, they got right.
> 
> Keeping track of stacking order by ourselves will only make the design complicated and will become hard to maintain.  My two cents. 
> 

Never mind. Actually they do keep track of the stack. But still, I think that is unnecessary. The responsibility of keeping track of window stack should be with X and the responsibility of updating the stack correctly should be with WM.

senthilbaboo:
I just did some quick hack in #2144 on how we should approach this. Basically, we never maintain the stack order in WM, but we keep the stack order updated in X and query the same whenever we need to access it.

tcld:
Closed in favor of #2194.

PR #2031, opened 2020-12-21
Global qtile

m-col:
This puts `qtile = Qtile()` directly at the module level of
libqtile/__init__.py. This means that the value of libqtile.qtile will
always by the instantiated qtile object. This solves a problem where any
module that imports it with `from libqtile import qtile` before the
instance calls `libqtile.init` to stick itself at `libqtile.qtile` will
retain a module-level variable called `qtile` with the value None. With
these changes, import `libqtile.qtile` in any way and at any time will
always yield the qtile object.

As creation of the Qtile instance happens very early on, its old
__init__() that sets it up has been replaced with an explicit init(),
which can be used when the process is going to become a server (i.e.
when run with the `qtile start` command. As some of the instantiation
logic has moved into libqtile/__init__.py, and as much of the scope of
the SessionManager overlapped with that of the Qtile class, the latter
has absorbed the former. This also justified moving manager.py out of
the 'core' folder.

This also lays the groundwork for setting up libqtile.qtile differently
when the process will be a qtile client rather than server. The test
changes were required due to conflict between the two variables named
`qtile` in conftest.py.


m-col:
These changes are pretty bold so feedback and comments welcome!

It looks a bit like the test has hung again... this has happened a few times recently.

m-col:
I think this might be going in the wrong direction. What I'm trying to accomplish is the ability to do `from libqtile import qtile` from a module within libqtile. The issue currently as I mention above is that `qtile` will reference None, except within the config. With these changes, adding the import to e.g. window.py then errors due to importing of an impartial module, due to the circular dependency (window.py wants \_\_init\_\_.py wants manager.py wants window.py). I think that that might be getting into bad design territory. I even looked into getting around it by defining a class in \_\_init\_\_.py, using that for `qtile`, then during its init method, injecting the `Qtile` class into its base classes so that it magically transformed into a `Qtile` instance. Super hacky. Not sure what alternative solutions there might be that are less ugly, though the solution might be to just leave it, as passing `qtile` to every object for all classes and saving it at `self.qtile` on all of them might be the least ugly solution.

tych0:
> What I'm trying to accomplish is the ability to do from libqtile import qtile from a module within libqtile.

What I wonder is: why? Generally people frown upon globals; is there a reason to prefer one in this case?

m-col:
This PR was kinda attempting to do 2 things simultaneously and that wasn't my initial intention. Firstly there is working towards using `libqtile.qtile` for both the server and client interface as we discussed in https://github.com/qtile/qtile/issues/1751. Then I got derailed by my dislike that we have to pass `qtile` to everything (such as `Window`, `Widget`, `Screen`, `Group` etc) so that they can access, which led me to change how `libqtile.qtile` is created initially (trying to set it as the qtile manager from the start, rather than changing it later when it is initialsied).  If nobody shares that dislike then that basically ends the discussion, which is fine :) In any case, the only solution I could think of was this, which basically turns libqtile.qtile into a global and takes a step into circular dependency hell, so all the more reason to leave it as is! That endeavor also led me to absorb SessionManager into Qtile, which I don't have particularly strong feeling toward anyway (although I do think it simplifies things marginally). Either way, none of things actually help in the path towards getting `libqtile.qtile` to also provide a client interface, because we can do usually a function similar to `libqtie.init` to modify the global, after determining that the process is a client process. So I shouldn't do such late night coding :)

tych0:
> Firstly there is working towards using libqtile.qtile for both the server and client interface as we discussed in #1751.

I sort of suspected this might be it :). Do you have a plan for how you would implement object sharing between the client and the server? I suspect this might inform how to restructure things for a global qtile.

FWIW, tcl mentioned on IRC that this PR does some other useful things (like renaming `qtile` in the test suite; although I don't love the name `self` either because it implies the thing is a member function, it is almost certainly better than qtile).

If you want to split that one and maybe the SessionManager bits you like into separate patches/PRs those seem reasonable.

PR #2034, opened 2020-12-22
Allow resizing of Drawer() internal pixmaps

DorianGray:
proto docs, so we lazily recreate the pixmap on any geometry change.

I ran into this while making a generic Tooltip(Popup) class. After adjusting the size of the text to be displayed, the window would only draw up to the previous text size.

### DIFF
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -10,6 +10,7 @@
 # Copyright (c) 2014 Nathan Hoad
 # Copyright (c) 2014 dequis
 # Copyright (c) 2014 Tycho Andersen
+# Copyright (c) 2020 Robert Andrew Ditthardt
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


libqtile/drawer.py
```diff
@@ -244,11 +228,74 @@ def __init__(self, qtile, wid, width, height):
...
+                self.width,
+                self.height
+            )
+            # surface is not guaranteed to be available before pixmap is
+            if self.surface is not None:
+                self.surface.drawable = self._pixmap
+
+        return self._pixmap
+
+    @pixmap.deleter
```
tych0, line 275: The idea itself looks good to me, but can we get rid of some of these properties and just call the actual code we want to call? e.g. here we could just call this:

```
def free_pixmap(self):
    ...
```

and then at user sites we just do ```self.free_pixmap()``` instead of `del self.pixmap`? Feels a little less magical to me that way.

DorianGray: Done.

DorianGray:
I rewrote things to take advantage of some changes on xcffib. The test failures are a direct result of those missing changes.

DorianGray:
This now depends on https://github.com/tych0/xcffib/pull/114

DorianGray:
Also, I just found a bug with this. Will re-open when it's fixed.

PR #2035, opened 2020-12-23
conftest: Qtile -> TestManager

m-col:
The Qtile class in test/conftest.py and the fixture 'qtile' that yields
it to tests share a name with the Qtile class in libqtile and the
'qtile' name that its instance takes. This has lead to confusion as they
do not represent the same thing, and are used in different ways. To make
the distinction more clear, this commit simply renames the testing
Qtile class to TestManager, and renames the fixture to 'm', so that
the manager passed to tests is referenced by the variable 'm'.

### DIFF
--- a/docs/manual/contributing.rst
+++ b/docs/manual/contributing.rst
@@ -61,9 +61,9 @@ documentation, but there are tutorials online that explain how it is used.
 
 Our tests are written inside the ``test`` folder at the top level of the
 repository. Reading through these, you can get a feel for the approach we take
-to test a given unit. Most of the tests involve an object called ``qtile``
-(note that this is distinct from ``libqtile.qtile``).  This exposes a command
-client at ``qtile.c`` that we use to test a Qtile instance running in a
+to test a given unit. Most of the tests involve an object called ``manager``.
+This is the test manager (defined in test/conftest.py), which exposes a command
+client at ``manager.c`` that we use to test a Qtile instance running in a
 separate thread as if we were using a command client from within a running
 Qtile session.
 
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -259,7 +259,7 @@ class Xephyr:
             pass
 
 
-class Qtile:
+class TestManager:
     """Spawn a Qtile instance
 
     Setup a qtile server instance on the given display, with the given socket
@@ -485,7 +485,7 @@ def xephyr(request, xvfb):
 
 
 @pytest.fixture(scope="function")
-def qtile(request, xephyr):
+def manager(request, xephyr):
     config = getattr(request, "param", BareConfig)
 
     for attr in dir(default_config):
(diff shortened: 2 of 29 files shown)


test/test_popup.py
```diff
@@ -26,24 +26,24 @@
...
-    qtile.test_xeyes()
-    qtile.windows_map = {}
+@pytest.mark.parametrize("m", [BareConfig], indirect=True)
+def test_popup_focus(m):
+    m.test_xeyes()
+    m.windows_map = {}
 
-    # we have to add .conn so that Popup thinks this is libqtile.qtile
-    qtile.conn = xcbq.Connection(qtile.display)
+    # we have to add .conn so that Popup thinks this is libqtile.m
```
tcld, line 34: Search and replace worked too well on this one. :)

tcld:
Two small things:
  1. I myself prefer longer names for variables, especially if you ever plan to do a search (and replace). Thus I would choose something like `testmanager` instead of `m`.
  2. I would kind of like to see `conftest.py` renamed: We can't call it `test_manager.py` after its central class, but maybe `base.py`? Next to `test_config.py` it doesn't stand out as much as it should, I think.

test/test_manager.py
```diff
@@ -87,111 +86,102 @@ class ManagerConfig(Config):
...
+    assert m.c.screen.info()["x"] == 0
+    assert m.c.screen.info()["width"] == 800
+    assert m.c.group.info()["name"] == 'a'
+    assert m.c.group.info()["focus"] == 'xclock'
 
 
 @pytest.mark.parametrize("xephyr", [{"xoffset": 0}], indirect=True)
 @manager_config
-def test_clone_dim(qtile):
-    self = qtile
```
tcld, line 123: Thanks for removing **that**!

ramnes:
`conftest.py` is a standard file used by pytest, you can't really change its name. I do agree on the longer name for the fixture, and I'd vouch for "manager".

tcld:
I didn't know about `conftest.py`, I'm very new to pytest. Just forget about that part, then.

`manager` sounds ok to me. With a class whose purpose mostly is to communicate with a running Qtile instance, the name could even be `QtileClient` or something - as long as it isn't `Qtile` anymore. :)

m-col:
Lol, nice spot.

test/test_popup.py
m-col: Lol yep

m-col:
Alright how's that for you fellas

docs/manual/contributing.rst
```diff
@@ -61,11 +61,11 @@ documentation, but there are tutorials online that explain how it is used.
 
 Our tests are written inside the ``test`` folder at the top level of the
 repository. Reading through these, you can get a feel for the approach we take
-to test a given unit. Most of the tests involve an object called ``qtile``
-(note that this is distinct from ``libqtile.qtile``).  This exposes a command
-client at ``qtile.c`` that we use to test a Qtile instance running in a
-separate thread as if we were using a command client from within a running
-Qtile session.
+to test a given unit. Most of the tests involve an object called ``m``. This
```
tcld, line 64: `manager`, right?^^

m-col: Ah yep, forgot to update this, thanks!

test/conftest.py
```diff
@@ -495,23 +495,23 @@ def qtile(request, xephyr):
...
-            yield q
+            yield manager
         finally:
-            q.terminate()
+            manager.terminate()
 
 
 @pytest.fixture(scope="function")
-def qtile_nospawn(request, xephyr):
+def m_nospawn(request, xephyr):
```
tcld, line 507: Anoter `m` - honestly can't tell whether this one is inentional, though.

m-col: oh, no, that was a total miss by my using \<m\> as my search and replace. This means that this exists elsewhere too! Man, really not on the ball today. I will replace m_nospawn everywhere.

tych0:
Thanks folks, I think this is a nice improvement.

PR #2036, opened 2020-12-23
Add OpenWeather-widget

tcld:
Following up on #1933, since @hschauhan didn't have the time to polish it.

I restructured the code a little and extracted a `WeatherResponseParser`-class, which is now also used by `YahooWeather`. I also tried to synchronize as many of the keywords as possible, while still allowing access to the more technical terms for (hopefully) longterm compatibility with the services.
The `app_key` is now optional again; I created an account with OpenWeather specifically for this purpose and made the key public, like I did with YahooWeather. Not sure whether that's the best approach, but as an end-user I wouldn't use the widget if I had to create an account first.

For now I kept the original commit, but once everything looks good I would probably merge the first two commits (while keeping the original author, of course).


tych0:
IIUC, the free yahoo weather API is going away fairly shortly. Can we just delete that and merge the base class into the open weather widget? Code all looks reasonable to me.

ramnes:
That would be great. I don't really like that base class.

And I still think that the whole parsing thing is unnecessary, as said in #1933. Let's just give to the user whatever OWM exposes, so that users can rely on OWM documentation directly. I don't think there's a value in us maintaining a middle layer.

tych0:
Ping, would be nice to get this in.

tcld:
@ramnes: To some degree the parsing is necessary, since the user won't have the ability to modify the received data anymore; that's where the methods in that extra class come in. The remapping of certain properties could probably be dropped, but then the user has to be directed towards the documentation. I just thought it would be nice to have a bit of a common interface across the different weather providers, but I understand that that makes it more likely for this stuff to break again without anybody taking care of it for a while.

So if I drop the Yahoo-widget (which is still working fine - do you have a link to share about the removal, @tych0 ?) and move the extracted class into the OpenWeather-file (while stripping it down to the bare essentials), would that be ok? I do like the two-class approach, since in my opinion it separates fetching and processing the data very nicely and might be a good example if anybody ever wanted to add another provider.

ramnes:
I don't think it's worth maintaining a middle layer between OWM and us and I don't think it's worth maintaining a common interface between this widget and another weather widget, let alone building stuff in anticipation for a second weather widget that may never happen if we delete the Yahoo one. And even if it ever happens, again, let's just really on the provider's documentation, that's the best way for us to ensure less maintenance and more quality (i.e. widgets that actually work).

ramnes:
> do you have a link to share about the removal, @tych0 ?

It's written as a header here: https://developer.yahoo.com/weather/

> As of Monday, February 1, 2021, the free Yahoo Weather API will be retired.

tych0:
Closing this one for now in favor of the rebase above.

PR #2037, opened 2020-12-24
Allow resizing of Drawer() internal pixmaps

DorianGray:
proto docs. Cairo to the rescue! Cairo implements a RecordingSurface
that we can use locally while the underlying Drawable may not be
accessible. These changes make the Drawer write to a RecordingSurface
and only copy to the pixmap on Drawer.draw.

This also fixes our random segfaults in CI since the XCBSurface is
created on draw instead of on __init__

### DIFF
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -10,6 +10,7 @@
 # Copyright (c) 2014 Nathan Hoad
 # Copyright (c) 2014 dequis
 # Copyright (c) 2014 Tycho Andersen
+# Copyright (c) 2020, 2021 Robert Andrew Ditthardt
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


DorianGray:
This ended up much more involved than my original PR, but I am confident this is much closer to the right solution.

Turns out that while surfaces and contexts allow changing their drawables, there isn't a way to manage the drawable lifecycle outside of . Luckily I came across RecordingSurface which is meant exactly for this. It stages cairo operations in memory and can be replayed on another surface. This allows us to manage a pixmap(drawable) lifecycle that works for us while maintaing ease of use for clients of Drawer.

DorianGray:
This now also depends on https://github.com/qtile/qtile/pull/2046

DorianGray:
This seems to also fix the random segfaults by proxy, since creating an xcbsurface too early was the problem.

ramnes (approved):
LGTM

ramnes:
Just reverted it, it breaks my bar:

broken-bar

It should look like this:

normal-bar

@m-col seems to have problems as well:

```
12h05 mcol i think that last commit (the Drawer resize one) might have some problems
12h05 mcol i just had to remove it from my live branch
12h06 mcol completely slowed down everything, and pretty sure it pumped my CPU a lot too for some reason
12h06 mcol could it be blocking qtile at times?
```

ramnes:
```
2021-01-05 15:13:20,236 ERROR libqtile manager.py:handle_exception():L64 'Drawer' object has no attribute 'pixmap'
  File "/usr/lib/python3.7/asyncio/events.py", line 88, in _run
  File "libqtile/core/manager.py", line 905, in f
    logger.error('Switching to ERROR threshold')
  File "libqtile/bar.py", line 376, in _actual_draw
    i.draw()
  File "libqtile/widget/systray.py", line 189, in draw
    icon.window.set_attribute(backpixmap=self.drawer.pixmap)
AttributeError: 'Drawer' object has no attribute 'pixmap'
```

PR #2038, opened 2020-12-24
Add natural resizing mechanism to resize from any window corner

m-col:
This lets users resize a window from any corner (that closest to the
cursor when the drag starts), holding the opposite corner in place.

The default approach using `lazy.window.set_size_floating()` and
`start=lazy.window.get_size()` cannot be modified to perform this
without modifying the core drag mechanism, as the initial drag x-y
position (stored during the start= function) is required to determine
the direction of resizing (i.e. a rightward drag from the left side
shrinks in x). Going down the route of modifying the core drag mechanism
would likely lead to config breakages.

Instead, this introduces two similar methods to be used as:

Drag(..., ..., lazy.window.drag_resize_floating(), start=lazy.window.start_resize())

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -947,7 +947,8 @@ class Window(_Window):
         hook.fire("client_managed", s)
 
     def tweak_float(self, x=None, y=None, dx=0, dy=0,
-                    w=None, h=None, dw=0, dh=0):
+                    w=None, h=None, dw=0, dh=0,
+                    from_left=False, from_top=False):
         if x is not None:
             self.x = x
         self.x += dx
@@ -977,7 +978,7 @@ class Window(_Window):
             screen.group.add(self, force=True)
             self.qtile.focus_screen(screen.index)
 
-        self._reconfigure_floating()
+        self._reconfigure_floating(from_left=from_left, from_top=from_top)
 
     def getsize(self):
         return (self.width, self.height)
@@ -985,7 +986,8 @@ class Window(_Window):
     def getposition(self):
         return (self.x, self.y)
 
-    def _reconfigure_floating(self, new_float_state=FLOATING):
+    def _reconfigure_floating(self, new_float_state=FLOATING,
+                              from_left=False, from_top=False):
         if new_float_state == MINIMIZED:
             self.state = IconicState
             self.hide()
@@ -996,12 +998,16 @@ class Window(_Window):
             if self.hints['base_width'] and self.hints['width_inc']:
                 width_adjustment = (width - self.hints['base_width']) % self.hints['width_inc']
                 width -= width_adjustment
+                if from_left:
+                    self.x += width_adjustment
                 if new_float_state == FULLSCREEN:
                     self.x += int(width_adjustment / 2)
 
             if self.hints['base_height'] and self.hints['height_inc']:
                 height_adjustment = (height - self.hints['base_height']) % self.hints['height_inc']
                 height -= height_adjustment
+                if from_top:
+                    self.y += height_adjustment
                 if new_float_state == FULLSCREEN:
                     self.y += int(height_adjustment / 2)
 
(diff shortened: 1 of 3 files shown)


tych0:
Any chance you can add some tests for this? :)

m-col:
I'd like to. What would you like to test? Maybe introduce xdotool and some mouse movements to the test suite?

tych0:
Anything that tests the major codepaths of the two new functions works for me; I suppose that will require introducing some mouse movement primitives to the test suite via xdotool or perhaps just direct use of xcb if you're feeling adventurous.

I don't think it's a problem to add an xdotool (or any other similar X utility) to the test suite, though, so if that route is simpler that's fine.

m-col:
Hmm, the new `test/test_manager.py::test_drag_resize` failed in the CI, but it works for me every time on my own machine. Can anyone else test it and see if they get failures? Any thoughts as to why? I had more success with `xdotool` actually having an effect when I split each command into its own process, so it does seem like a temperamental tool.

libqtile/resources/default_config.py
```diff
@@ -151,8 +151,8 @@
 mouse = [
     Drag([mod], "Button1", lazy.window.set_position_floating(),
          start=lazy.window.get_position()),
-    Drag([mod], "Button3", lazy.window.set_size_floating(),
-         start=lazy.window.get_size()),
+    Drag([mod], "Button3", lazy.window.drag_resize_floating(),
+         start=lazy.window.start_resize()),
```
ramnes, line 155: I don't see a point in keeping both implementations alive, since this one is strictly superior in terms of user experience. I think we should modify `set_position_floating` and `get_position` directly (and whatever mechanism used in `Drag`) rather than implementing new methods.

libqtile/window.py
```diff
@@ -1307,6 +1314,45 @@ def cmd_resize_floating(self, dw, dh):
...
+    def cmd_start_resize(self):
+        start_x = None
+        start_y = None
+        x, y = self.qtile.mouse_position
+        if x < self.x + self.width // 2:
+            start_x = self.x
+        if y < self.y + self.height // 2:
+            start_y = self.y
+        self._resize_start = (self.width, self.height, start_x, start_y)
+        return (0, 0)
```
ramnes, line 1326: Returning a fake value while putting the real stuff in the window object sounds nasty. Can't we respect the current `Drag` API or extend it?

libqtile/resources/default_config.py
m-col: Hopefully this isn't easier said than done. I'll dive right in!

m-col:
Alrighty I gave it another shot and modified the Drag API slightly to make it work.

- The Drag mechanism now does not pre-sum the starting coordinates with the current coordinates before passing to any Drag commands. Instead, the `val` returned by any `start=` command passed to `Drag` is simply forwarded to whatever the main `Drag` command is. `Window.cmd_set_position_floating` has been updated so the default respositioning `Drag` configuration is not broken.
- `Window.cmd_set_size_floating` which does the resizing has been updated to do the natural resizing but to do it it requires the window's original x and y coords. I haven't come up with a way to calculate these with the data available (original dimensions and total `dx` and `dy` since the drag began, as well as current state). My solution for this was the following...
- `Window.cmd_get_size` has been changed so that instead of returning `(self.width, self.height)` it returns `(self.x, self.y, self.width, self.height)`. I think this likely breaks considerably fewer configs because it means nobody has to change their `Drag`s to use this i.e. we don't have to tell people to change their `start=` kwarg to `Window.cmd_get_geometry` (doesn't yet exist). Those who use `Window.cmd_get_size` in other things will of course have problems, but I would think those people are fewer in number and more likely to not be bothered by the changes / read the changelog.

I don't really like that last point so I'm interested in hearing other solutions, particularly if you think the original coordinates can be calculated within `Window.cmd_set_size_floating` without passing them through like that.

In any case, I haven't added tests yet so i've made it a draft.

m-col:
Don't look into this yet, it has a big bug.

m-col:
I fixed the bug I mentioned so it now works as expected. It isn't very efficient because upon every motion step the cmd needs to calculate which corner of the original window dimensions the original mouse position was within to know how to resize. This of course can be avoided, as can the change to `Window.cmd_get_size`, if instead we use a new `start=` cmd to pass to the `Drag` that has all of this required information. This would lead to immediate breakage for everyone who upgrades unless we use a new `cmd` for the resizing itself too (like in the original commit). An ugly compromise might be to have a temporary patch to allow either to work in the next release (specifically looking to see if `cmd_get_size` is being used within a Drag), with noisy logs when people use the old method, then remove it. Any thoughts?

PR #2039, opened 2020-12-24
Add scroll text widget

Bauthe:
Adds a widget that can scroll text.
Here is a config to showcase this widget: https://gist.github.com/Bauthe/258371d2734f6d5d9f9ae38ad5cc3617

### DIFF
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -22,158 +22,154 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from dbus_next.constants import MessageType  # type: ignore
+import asyncio
+from collections import defaultdict
+from string import Formatter
+from html import escape
+
+from dbus_next import MessageType
+from dbus_next.aio import MessageBus
 
 from libqtile.log_utils import logger
 from libqtile.utils import add_signal_receiver
 from libqtile.widget import base
 
+play_states = {
+    'play': '\u23f5',
+    'pause': '\u23f8',
+}
 
-class Mpris2(base._TextBox):
+class Mpris2(base._ScrollText):
     """An MPRIS 2 widget
 
     A widget which displays the current track/artist of your favorite MPRIS
     player. It should work with all MPRIS 2 compatible players which implement
-    a reasonably correct version of MPRIS, though I have only tested it with
-    audacious.  This widget scrolls the text if neccessary and information that
-    is displayed is configurable.
+    a reasonably correct version of MPRIS. This widget scrolls the text if
+    necessary and information that is displayed is configurable.
     """
-    orientations = base.ORIENTATION_HORIZONTAL
-    defaults = [
-        ('name', 'audacious', 'Name of the MPRIS widget.'),
 
+    defaults = [
         ('objname', 'org.mpris.MediaPlayer2.audacious',
             'DBUS MPRIS 2 compatible player identifier'
             '- Find it out with dbus-monitor - '
             'Also see: http://specifications.freedesktop.org/'
             'mpris-spec/latest/#Bus-Name-Policy'),
-
-        ('display_metadata', ['xesam:title', 'xesam:album', 'xesam:artist'],
-            'Which metadata identifiers to display. '
-            'See http://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index5h3 '
-            'for available values'),
-
-        ('scroll_chars', 30, 'How many chars at once to display.'),
-        ('scroll_interval', 0.5, 'Scroll delay interval.'),
-        ('scroll_wait_intervals', 8, 'Wait x scroll_interval before'
-            'scrolling/removing text'),
-
-        ('stop_pause_text', None, "Optional text to display when in the stopped/paused state"),
+        ('format', '{play_status}  <b>{title}</b> - {artist}',
+            'How to format the status metadata.'
+            'Accepted keys are: "play_status" and Xesam properties.'
+            'https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/'),
+        ('play_states', play_states, 'dict of `play_status` strings.'
+            'Accepted keys are "play" and "pause".'),
+        ('idle_message', '', 'Text to display when the player is idle'),
+        ('enable_scrolling', True, 'Wether to scroll the text.'),
     ]
 
     def __init__(self, **config):
-        base._TextBox.__init__(self, '', **config)
+        base._ScrollText.__init__(self, **config)
         self.add_defaults(Mpris2.defaults)
 
-        self.scrolltext = None
-        self.displaytext = ''
-        self.is_playing = False
-        self.scroll_timer = None
-        self.scroll_counter = None
-        self.count = 0
+        self.player = None
+        self.add_callbacks({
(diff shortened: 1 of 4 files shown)


ramnes:
Interesting! But please:
* remove all the noise related to reformatting
* add a widget (or modify an existing one) so that this base class is actually used somewhere and do not just end up being dead code
* ideally, add tests

m-col:
General question: would it make sense and be possible to add the scrolling functionality to the existing `_TextBox` or even `_Widget` base classes?

Bauthe:
@ramnes
> * remove all the noise related to reformatting

Done

> * add a widget (or modify an existing one) so that this base class is actually used somewhere and do not just end up being dead code

I modified the Mpris2 widget in a second commit. What I've added:
- Smooth text scrolling
- Control of the player through a dbus interface
- More formatting options

> * ideally, add tests

I can help with that when I have some feedback about the functionnality


@m-col 
> General question: would it make sense and be possible to add the scrolling functionality to the existing `_TextBox` or even `_Widget` base classes?

It is possible on `_TextBox`, but I don't think we want it because it adds a lot of overhead/clutter to the drawing function. I would suggest to keep it separate for readability/performance.

Bauthe:
@elParaguayo @tych0 mentioning you because you were interested in my first PR #1927

elParaguayo:
Thanks. I was going to make some changes to this widget myself as I hated the metadata/playback handling section of the code. However, I can see you're way ahead of me which is great!

I also like the fact that you've added some basic controls - (shameless plug alert) that's what I had in mind as part of #2251.

I'll take a look later this week and let you know if I have any comments.

elParaguayo:
The scrolling looks really nice and a big improvement on the character scrolling.

A few comments on this:
1) Is MarkupSafe really needed as a dependency?
2) There are a couple of glitches but I haven't dug into the code to see what's causing them:

i) When scrolling there's a small artefact on both the left and right sides, like those pixels aren't part of the scroll.

ii) Sometimes the widget changes size (could due to some empty characters at the end of the string). This results in the widget being too long to start with but resetting itself when the scroll loop restarts.

However, metadata, playback status and controls all seem to work.

libqtile/widget/mpris2widget.py
```diff
@@ -22,158 +22,152 @@
...
-
-        ('stop_pause_text', None, "Optional text to display when in the stopped/paused state"),
+        ('format', '{status_char}  <b>{title}</b> - {artist}',
+            'How to format the data.'
+            'Accepted keys are: "status_char" and Xesam properties.'
+            'https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/'),
+        ('play_char', '', '"status_char" to display a song is playing.'),
+        ('pause_char', '', '"status_char" to display when a song is paused.'),
+        ('txt_inactive', '', 'Text to display when the player is inactive'),
+        ('scrolling_enabled', True, 'Wether to scroll the text.'),
```
tych0, line 61: This looks like a bunch of breaking changes. Please either 1. implement a migration for these changes via qtile-migrate or 2. figure out how to make them non-breaking.

Bauthe:
@elParaguayo thanks for the input
> 1. Is MarkupSafe really needed as a dependency?

Not at all, after looking it up it seems feasible to escape them manually: https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

> i) When scrolling there's a small artefact on both the left and right sides, like those pixels aren't part of the scroll.
>
> ii) Sometimes the widget changes size (could due to some empty characters at the end of the string). This results in the widget being too long to start with but resetting itself when the scroll loop restarts.

I don't encounter these issues, so I don't see how to solve them. If you use a particular config, could you share it ? Or send a gif of these glitchy behaviours ? Here is how it behaves on my side:

MPRIS2widget

libqtile/widget/mpris2widget.py
Bauthe: I've adapted the signature so that it fits as much as possible the `Mpd2` widget syntax. I will take a look at `qtile migrate`, it seems awesome.

tych0: The problem is that people who are existing users of the mpris widget will not necessarily be familiar with the mpd2 widget. It is best to make changes non-breaking if possible.

libqtile/widget/mpris2widget.py
```diff
@@ -22,158 +22,154 @@
...
-        if len(self.scrolltext) >= self.scroll_chars:
-            self.scrolltext = self.scrolltext[1:]
-            if len(self.scrolltext) == self.scroll_chars:
-                self.scroll_counter += self.scroll_wait_intervals
-            self.timeout_add(self.scroll_interval, self.scroll_text)
-            return
-        self.text = ''
-        self.bar.draw()
-
-    def cmd_info(self):
```
tych0, line 174: This is a public API, so we should avoid dropping it.

libqtile/widget/mpris2widget.py
```diff
@@ -22,158 +22,154 @@
...
 
     async def _config_async(self):
         subscribe = await add_signal_receiver(
-                        self.message,
-                        session_bus=True,
-                        signal_name="PropertiesChanged",
-                        bus_name=self.objname,
-                        path="/org/mpris/MediaPlayer2",
-                        dbus_interface="org.freedesktop.DBus.Properties")
+            self.message,
```
tych0, line 85: This looks like a bunch of whitespace changes; can we drop these whitespace changes (and all others, e.g. the `"".format() -> f""`) from this patch?

elParaguayo:
You can see both issues in the gif below:
mpris


1) Artefacts at either end of the scroll
2) After stopping and starting, the widget has moved to the left but resets correctly after the first scroll. (This is when you see the pause symbol, then the text disappears as I stop the track. When I restart the track, the text is in the wrong place).

Bauthe:
I've managed to reproduce the "widget moving to left behaviour" by putting a `bar.STRETCH` spacer before my widget, so I might be able to debug it easily. I think it is because I don't redraw the bar everytime I should. For the artefacts on each end, I can't reproduce it, but I think it is due to imprecisions in floating point values, e.g. in `_ScrollText.draw`: `padding = int(min(self.actual_padding, self.width / 2))`.

libqtile/widget/mpris2widget.py
Bauthe: Yes I get it. I will try to keep the old names when it can still make sense. However, what should I do to deprecate certain parameters smoothly ? Is there a guideline ?

elParaguayo:
If I set `padding` to 0 in that file the artefacts disappear so you are right about that being the cause.

I'd like to take a look at the `_ScrollText` class some more as I'm not sure I follow all the logic at the moment.

m-col:
> It is possible on _TextBox, but I don't think we want it because it adds a lot of overhead/clutter to the drawing function. I would suggest to keep it separate for readability/performance.

To me it seems an arbitrary decision for some text-based widgets to be scollable and others not. What kind of performance hit does a scrollable widget really experience, even when not scrolling? I foresee once there is one scrollable widget people will want to scroll other ones. Indeed I have a few text-based widgets I'd like this feature on. Can we work around any expected performance hit when scrolling is not enabled?

libqtile/widget/mpris2widget.py
tych0: Yes, you'll need to implement a `qtile migrate` migration for anything you'd like to change.

PR #2041, opened 2020-12-24
Add window margins for different directions

arczan2:
Changed window margins behavior in place() method. Now it can also be passed as list of ints [N E S W] like Bar margins.
I think It might be helpful when creating some complex Layouts.

### DIFF
--- a/libqtile/layout/bsp.py
+++ b/libqtile/layout/bsp.py
@@ -152,7 +152,7 @@ class Bsp(Layout):
         ("border_focus", "#881111", "Border colour for the focused window."),
         ("border_normal", "#220000", "Border colour for un-focused windows."),
         ("border_width", 2, "Border width."),
-        ("margin", 0, "Margin of the layout."),
+        ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
         ("ratio", 1.6,
          "Width/height ratio that defines the partition direction."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -117,7 +117,7 @@ class Columns(Layout):
          "Border colour for un-focused windows in stacked columns."),
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw a border when there is one only window."),
-        ("margin", 0, "Margin of the layout."),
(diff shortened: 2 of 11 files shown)


tych0:
Do you have a sample layout that would benefit from this? Right now I think this would be dead code in the tree.

arczan2:
> Do you have a sample layout that would benefit from this? Right now I think this would be dead code in the tree.

I got this idea when I was modifying the Columns layout. I wanted to get different gaps between columns and rows:
`layout.Columns(row_margin=10, col_margin=2, border_focus='#777777'),`

tych0:
Ok, I'd like to see some in-tree code that uses this before we land it, and ideally some tests as well.

arczan2:
I've used this margins in xmonad layouts. In my opinion It is more readable now

ramnes (approved):
LGTM! Maybe we could go a step further and modify all layouts that take a `margin` argument so that they accept such a list? It can be done in another PR. :)

tych0:
Can you show an example of how you've used it in the commit message?

arczan2:
> Can you show an example of how you've used it in the commit message?

Do you want me to create a new empty commit with massage: "Make xmonad layout use list of window margins instead of calculating x, y, width and height values"?
Did I understand you correctly?

arczan2:
> LGTM! Maybe we could go a step further and modify all layouts that take a `margin` argument so that they accept such a list? It can be done in another PR. :)

I can take care of it :)

arczan2:
So I've checked all layouts. Most of them work with margins list without any code changes :) The only thing I would do is to update the documentation:
`("margin", 0, "Margin of the layout(int or list of ints [N E S W])")`

The only one that need special attention is xmonad (Tall and Wide). The problem is that this layout does not add margin from one side of main window, to avoid "double gap"  in the middle. This can be solved by adding new parameter `main_margin` but I don't know if it is a good idea. That will lead us to the situation where xmonad layout will have parameters: `single_margin, margin, main_margin`.`

ramnes:
Cool! Can you edit the docstrings then? Let's ignore the Monad layouts for now. :)

arczan2:
> Cool! Can you edit the docstrings then? Let's ignore the Monad layouts for now. :)

Done! 
It looks like failure is unrelated :(

ramnes:
Merged! Thank you @arczan2. :)

Issue #2027, opened 2020-12-20
KeyboardLayout widget doesn't work as expected


k4leg:
# Issue description
In `libqtile.widget.keyboardlayout.KeyboardLayout` viewing the current layout is implemented through parsing `setxkbmap` (`setxkbmap -verbose 10 -query`) which is wrong (it doesn't give information about the current layout).
You can check with this example:
```
$ setxkbmap 'us,ru' -option 'grp:caps_toggle'
$ setxkbmap -verbose 10 -query > english # Current layout — English.
$ setxkbmap -verbose 10 -query > русский # Current layout — Russian.
$ diff -Naurp0 english русский # Without output, which means it doesn't change when you change the keyboard layout.
```

`setxkbmap` doesn't let you know the current layout.

Switching layouts also doesn't work as expected. Take a look at an example:
```
# `libqtile.widget.keyboardlayout.KeyboardLayout()`
$ setxkbmap 'us,ru' -option 'grp:caps_toggle,compose:menu' # My setup.
$ setxkbmap us # `libqtile.widget.keyboardlayout.KeyboardLayout` actions when clicking on it.
$ setxkbmap us # `libqtile.widget.keyboardlayout.KeyboardLayout` actions when clicking on it.
# My keyboard layout is now only 'us' I cannot change it back to 'ru' even through the widget, through Caps Lock.

# `libqtile.widget.keyboardlayout.KeyboardLayout(configured_keyboards=['us', 'ru'])`
$ setxkbmap 'us,ru' -option 'grp:caps_toggle,compose:menu' # My setup.
$ setxkbmap ru # `libqtile.widget.keyboardlayout.KeyboardLayout` actions when clicking on it.
$ setxkbmap us # `libqtile.widget.keyboardlayout.KeyboardLayout` actions when clicking on it.
# Thus, I cannot switch the keyboard layout via Caps Lock. Only through the widget even if I add an `option='grp:caps_toggle,compose:menu'` parameter.
```

Also, the keyboard layout doesn't change automatically in the widget if I change it through Caps Lock.

# Qtile version
Since [9bb78e80 ("Adding keyboard layout widget", 2013-08-17)](https://github.com/qtile/qtile/commit/9bb78e80).

tych0:
Can you propose a patch to fix it? I'm not sure I understand all the nuances from your report.

k4leg:
> Can you propose a patch to fix it? I'm not sure I understand all the nuances from your report.

[...]

Issue #1130, opened 2018-02-18
Packaging: Tests fail in headless chroot


dvzrv:
I'm currently moving qtile to the Arch Linux [community] repository. Packages are built in a clean chroot to ensure sane environments and working dependencies.

However, when trying to run the tests during packaging

```
python libqtile/ffi_build.py
py.test

```

a lot of them fail, because Xephyr can not be started properly (amongst other issues):
[qtile-build.log](https://github.com/qtile/qtile/files/1735157/qtile-build.log)

According to #906 Xephyr is supposed to run in xvfb. Both are available in the test environment, but it still fails.
AFAICS, I'm not supposed to start Xephyr myself, as this is taken care of by the test suite, right?

When not building in a clean chroot environment (this I can not do for packaging!) the tests still fail:
[qtile-build-user.log](https://github.com/qtile/qtile/files/1735168/qtile-build-user.log)

Does the test setup assume, that Xorg in the environment is run by root (I run X as my user)?

In any case: I have to disable tests (and this is bad) to build qtile as a package.

Maybe you have any thoughts on this or any suggestions as to how the tests can be used.

flacjacket:
Hmmm, I had seen similar errors running on my machine, but figured it was
an issue on my side since the tests have continued to pass on Travis. It
must be something with a newer library that isn't on the old Travis image.
Xephyr and xvfb should be started up by the test runner. I have also had
the same issues just running xephyr not inside xvfb. We'll need to figure
out what is causing this, I had seen this behavior even just connecting and
disconnecting from xephyr with xpyb, so I don't suspect it is an issue
necessarily on our side. The next steps will be likely gathering the
information needed for a bug report upstream.

dvzrv:
On Arch Linux currently the following things are in use (build from [this crazy split-package](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/xorg-server)):
xorg-server{,xephyr,xvfb} 1.19.6+13+gd0d1a694f

flacjacket:
Looking into this further, rolling back my xorg-server to 1.19.5 seems to fix this issue.

flacjacket:
Doing a little extra digging, this is what the backtrace looks like:
```
#0  0x0000000000000000 in ?? ()
#1  0x00005555555e7872 in doFreeResource (res=0x5555559ed360, skip=skip@entry=0)
[...]

PR #2042, opened 2020-12-24
Fix and remove unused code in KeyboardLayout widget

k4leg:
#2027

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -22,116 +22,269 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import re
+import asyncio
 from subprocess import CalledProcessError
+from typing import NoReturn, Union
 
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
-kb_layout_regex = re.compile(r'layout:\s+(?P<layout>\w+)')
-kb_variant_regex = re.compile(r'variant:\s+(?P<variant>\w+)')
+
+class NoDependencyFoundError(Exception):
+    pass
 
 
 class KeyboardLayout(base.InLoopPollText):
-    """Widget for changing and displaying the current keyboard layout
-
-    To use this widget effectively you need to specify keyboard layouts you want to use (using "configured_keyboards")
-    and bind function "next_keyboard" to specific keys in order to change layouts.
-
-    For example:
-
-        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
-
-    It requires setxkbmap to be available in the system.
     """
+    Widget for changing and displaying the current keyboard layout group
+
+    It requires xkb-switch to be available in the system.  setxkbmap is
+    optional dependency (to configure the keyboard locally).
+    """
+
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ("update_interval", 1, "Update time in seconds."),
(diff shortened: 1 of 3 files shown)


libqtile/widget/keyboardlayout.py
```diff
@@ -22,107 +22,245 @@
...
+        (
+            'variant',
+            None,
+            "Specifies which variant of the keyboard layout should be used to"
+            " determine the components which make up the keyboard"
+            " description.",
+        ),
     ]
+    dependencies = {
+        'setxkbmap': ['setxkbmap', '-version'],
```
tych0, line 148: Can we depend on either xkb-switch or setxkbmap, but not both?

k4leg: Yes, we can make setxkbmap an optional dependency (then the user will have to configure it via `/etc/X11/xorg.conf.d/00-keyboard.conf`).

k4leg:
Looks like issue #1130. Tests fail with `TOXENV=py39 tox` in qtile/master.

libqtile/widget/keyboardlayout.py
```diff
@@ -22,107 +22,274 @@
...
 
-    def __init__(self, **config):
-        base.InLoopPollText.__init__(self, **config)
+    def __init__(self, **config) -> None:
+        base._TextBox.__init__(self, **config)
         self.add_defaults(KeyboardLayout.defaults)
+        self.raise_for_dependencies()
+        self.use_optional_dependencies_if_available()
+        event_loop = asyncio.get_event_loop()
+        event_loop.create_task(self.loop())
```
tych0, line 157: Is there a reason to switch to this vs. just continuing to use InLoopPollText? Maybe we should just switch InLoopPollText to use this?

k4leg: Okay, there is no particular reason to switch to `_TextBox`.

tcld:
For me `xkb-switch` appears to not be part of any official repositories. This means I have to remove the keyboard widget to keep using Qtile or stay on an older commit for now. Is `xkb-switch` expected to become a part of official repositories for arch and other distros?

tych0:
> I have to remove the keyboard widget to keep using Qtile or stay on an older commit for now

Or send a patch to fix the bug via setxkbmap :)

> Is xkb-switch expected to become a part of official repositories for arch and other distros?

No idea. I don't use this area of the code myself, so independent reviews are helpful. I think if you'd like this reverted, open a new issue and tag @k4leg in it and let's see what happens.

tcld:
I honestly don't think I ever understood the issue with the previous implementation, so I have no base at all to fix anything.

I can't tell without going back and figuring out what the issue was, but I hope that either `xkb-switch` is going to eventually become official or - at least - this new implementation brings a ton of value with it. (Does it?)

tych0:
> I honestly don't think I ever understood the issue with the previous implementation, so I have no base at all to fix anything.

I didn't either; there are many areas of Qtile that I do not personally use. I rely on community reviews to vet pull requests against these areas.

> I hope that either xkb-switch is going to eventually become official 

What does "become official" mean, exactly? Become more widely used? It can only become more widely used if people adopt it :). As I said, I'm open to alternative implementations which use more mainstream tools.

> this new implementation brings a ton of value with it.

It fixed a bug someone reported, and sat on the PR tracker for two months with no negative reviews. In my mind, that says it adds value. If you don't like it, please open a new issue with your reasoning why, and we can discuss it there.

In general, I appreciate the feedback, but it is more useful *before* stuff lands :)

tych0:
I can appreciate the "this is an obscure dependency" argument, so thanks for raising this issue. If you have this complaint, I'm sure others do too. I filed #2265 for further discussion.

PR #2046, opened 2020-12-27
Refactor Qtile LifeCycle

DorianGray:
This PR refactors the process lifecycle related parts of Qtile to properly manage resource finalization and just generally cleans up the code.

Instead of running os.execv before python's builtin finalization
routines, this uses those same finalization routines to move os.execv to
the last thing done by the interpreter before termination.

### DIFF
--- a/bin/qtile
+++ b/bin/qtile
@@ -28,7 +28,9 @@ this_dir = os.path.dirname(__file__)
 base_dir = os.path.abspath(os.path.join(this_dir, ".."))
 sys.path.insert(0, base_dir)
 
-from libqtile.scripts.main import main
+# import lifecycle early so no atexit handlers are lost on restart
+import libqtile.core.lifecycle  # noqa: F401, E402
+from libqtile.scripts.main import main  # noqa: E402
 
 if __name__ == '__main__':
     main()
--- a/libqtile/core/lifecycle.py
+++ b/libqtile/core/lifecycle.py
@@ -1,0 +1,43 @@
+import atexit
+import enum
+import os
+import sys
+from typing import Optional
+
+from libqtile.log_utils import logger
+
+__all__ = [
+    'lifecycle',
+]
+
+Behavior = enum.Enum('Behavior', 'NONE TERMINATE RESTART')
+
+
+class LifeCycle:
+    # This class exists mostly to move os.execv to the absolute last thing that
+    # the python VM does before termination.
+    # Be very careful about what references this class owns. Any object
+    # referenced here when atexit fires will NOT be finalized properly.
+    def __init__(self) -> None:
+        self.behavior = Behavior.NONE
+        self.state_file: Optional[str] = None
+        atexit.register(self._atexit)
+
+    def _atexit(self) -> None:
+        if self.behavior is Behavior.RESTART:
+            argv = [sys.executable] + sys.argv
+            if '--no-spawn' not in argv:
+                argv.append('--no-spawn')
+            argv = [s for s in argv if not s.startswith('--with-state')]
+            if self.state_file is not None:
+                argv.append('--with-state=' + self.state_file)
+            logger.warning('Restarting Qtile with os.execv(...)')
+            # No other code will execute after the following line does
+            os.execv(sys.executable, argv)
+        elif self.behavior is Behavior.TERMINATE:
+            logger.warning('Qtile will now terminate')
+        elif self.behavior is Behavior.NONE:
+            pass
+
+
+lifecycle = LifeCycle()
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -26,8 +26,6 @@ import pickle
 import shlex
 import signal
 import subprocess
-import sys
-import tempfile
 import time
 import warnings
 
@@ -59,29 +57,21 @@ from libqtile.utils import get_cache_dir, send_notification
 from libqtile.widget.base import _Widget
 
 
-def handle_exception(loop, context):
-    if "exception" in context:
-        logger.error(context["exception"], exc_info=True)
-    else:
-        logger.error("exception in event loop: %s", context)
-
-
 class Qtile(CommandObject):
     """This object is the `root` of the command graph"""
     def __init__(
         self,
         kore,
         config,
-        eventloop,
         no_spawn=False,
         state=None
     ):
-        self._restart = False
         self.core = kore
         self.no_spawn = no_spawn
         self._state = state
 
         self._stopped_event = None
+        self.should_restart = False
 
         self._drag = None
         self.mouse_map = None
@@ -102,57 +92,10 @@ class Qtile(CommandObject):
 
         libqtile.init(self)
 
+        self._eventloop = None
         self.server = IPCCommandServer(self)
         self.config = config
         self.load_config()
-
-        self._eventloop = eventloop
-        self.setup_eventloop()
-
-        self._process_screens()
-        self.current_screen = self.screens[0]
-        self._drag = None
-
-        self.conn.flush()
-        self.conn.xsync()
-        self.core._xpoll()
-
-        # Map and Grab keys
-        for key in self.config.keys:
-            self.grab_key(key)
-
-        self.mouse_map = {}
-        for i in self.config.mouse:
-            if self.mouse_map.get(i.button_code) is None:
-                self.mouse_map[i.button_code] = []
-            self.mouse_map[i.button_code].append(i)
-
-        self.grab_mouse()
-
-        # no_spawn is set when we are restarting; we only want to run the
-        # startup hook once.
-        if not no_spawn:
-            hook.fire("startup_once")
-        hook.fire("startup")
-
-        if state:
-            try:
-                with open(state, 'rb') as f:
-                    st = pickle.load(f)
-                st.apply(self)
-            except:  # noqa: E722
-                logger.exception("failed restoring state")
-            finally:
-                os.remove(state)
-
-        self.core.scan()
-        if state:
-            for screen in self.screens:
-                screen.group.layout_all()
-        self.update_net_desktops()
-        hook.subscribe.setgroup(self.update_net_desktops)
-
-        hook.fire("startup_complete")
 
     def load_config(self):
         try:
@@ -210,6 +153,63 @@ class Qtile(CommandObject):
             pass
         self.config.mouse += (Click([], "Button1", lazy.function(noop), focus="after"),)
 
+    def dump_state(self, buf):
+        try:
+            pickle.dump(QtileState(self), buf, protocol=0)
+        except:  # noqa: E722
+            logger.exception('Unable to pickle qtile state')
+
+    def _configure(self):
+        """
+        This is the part of init that needs to happen after the event loop is
+        fully set up. asyncio is required to listen and respond to backend
+        events.
+        """
+        self._process_screens()
+        self.current_screen = self.screens[0]
+
+        self.conn.flush()
+        self.conn.xsync()
+        self.core._xpoll()
+
+        # Map and Grab keys
+        for key in self.config.keys:
+            self.grab_key(key)
+
+        self.mouse_map = {}
+        for i in self.config.mouse:
+            if self.mouse_map.get(i.button_code) is None:
+                self.mouse_map[i.button_code] = []
+            self.mouse_map[i.button_code].append(i)
+
+        self.grab_mouse()
+
+        # no_spawn is set when we are restarting; we only want to run the
+        # startup hook once.
+        if not self.no_spawn:
+            hook.fire("startup_once")
+        hook.fire("startup")
+
+        if self._state:
+            try:
+                with open(self._state, 'rb') as f:
+                    st = pickle.load(f)
+                st.apply(self)
+            except:  # noqa: E722
+                logger.exception("failed restoring state")
+            finally:
+                os.remove(self._state)
+
+        self.core.scan()
+        if self._state:
+            for screen in self.screens:
+                screen.group.layout_all()
+        self._state = None
+        self.update_net_desktops()
+        hook.subscribe.setgroup(self.update_net_desktops)
+
+        hook.fire("startup_complete")
+
     @property
     def root(self):
         return self.core._root
@@ -222,97 +222,42 @@ class Qtile(CommandObject):
     def selection(self):
         return self.core._selection
 
-    def setup_eventloop(self) -> None:
-        self._eventloop.add_signal_handler(signal.SIGINT, self.stop)
-        self._eventloop.add_signal_handler(signal.SIGTERM, self.stop)
-        self._eventloop.set_exception_handler(handle_exception)
-
-        logger.debug('Adding io watch')
-        self.core.setup_listener(self, self._eventloop)
-
-        self._stopped_event = asyncio.Event()
-
-        # This is a little strange. python-dbus internally depends on gobject,
-        # so gobject's threads need to be running, and a gobject "main loop
-        # thread" needs to be spawned, but we try to let it only interact with
-        # us via calls to asyncio's call_soon_threadsafe.
-        try:
-            # We import dbus here to thrown an ImportError if it isn't
-            # available. Since the only reason we're running this thread is
-            # because of dbus, if dbus isn't around there's no need to run
-            # this thread.
-            import dbus  # noqa
-            from gi.repository import GLib  # type: ignore
-
-            def gobject_thread():
-                ctx = GLib.main_context_default()
-                while not self._stopped_event.is_set():
-                    try:
-                        ctx.iteration(True)
-                    except Exception:
-                        logger.exception("got exception from gobject")
-            self._glib_loop = self.run_in_executor(gobject_thread)
-        except ImportError:
-            logger.warning("importing dbus/gobject failed, dbus will not work.")
-            self._glib_loop = None
-
     async def async_loop(self) -> None:
         """Run the event loop
 
         Finalizes the Qtile instance on exit.
         """
+        self._eventloop = asyncio.get_running_loop()
+        self._stopped_event = asyncio.Event()
+        self.core.setup_listener(self)
+        self._configure()
         try:
             await self._stopped_event.wait()
         finally:
-            await self.finalize()
-
-        self._eventloop.stop()
-
-    def maybe_restart(self) -> None:
-        """If set, restart the qtile instance"""
-        if self._restart:
-            logger.warning('Restarting Qtile with os.execv(...)')
-            os.execv(*self._restart)
+            self.finalize()
+            self.core.remove_listener()
 
     def stop(self):
-        # stop gets called in a variety of ways, including from restart().
-        # let's only do a real shutdown if we're not about to re-exec.
-        if not self._restart:
-            hook.fire("shutdown")
-            self.graceful_shutdown()
-
-        logger.debug('Stopping qtile')
-        self._stopped_event.set()
+        hook.fire("shutdown")
+        self.graceful_shutdown()
+        self._stop()
 
     def restart(self):
         hook.fire("restart")
-        argv = [sys.executable] + sys.argv
-        if '--no-spawn' not in argv:
-            argv.append('--no-spawn')
-        state_file = os.path.join(tempfile.gettempdir(), "qtile-state")
-        try:
-            with open(state_file, 'wb') as f:
-                pickle.dump(QtileState(self), f, protocol=0)
-        except:  # noqa: E722
-            logger.error("Unable to pickle qtile state")
-        argv = [s for s in argv if not s.startswith('--with-state')]
-        argv.append('--with-state=' + state_file)
-        self._restart = (sys.executable, argv)
-        self.stop()
+        self.should_restart = True
+        self._stop()
 
-    async def finalize(self):
-        self._eventloop.remove_signal_handler(signal.SIGINT)
-        self._eventloop.remove_signal_handler(signal.SIGTERM)
-        self._eventloop.set_exception_handler(None)
+    def _stop(self):
+        logger.debug('Stopping qtile')
+        if self._stopped_event is not None:
+            self._stopped_event.set()
 
-        if self._glib_loop:
-            try:
-                from gi.repository import GLib
-                GLib.idle_add(lambda: None)
-                await self._glib_loop
-            except ImportError:
-                pass
+    def is_stopped(self):
+        if self._stopped_event is not None:
+            return self._stopped_event.is_set()
+        return False
 
+    def finalize(self):
         try:
             for widget in self.widgets_map.values():
                 widget.finalize()
@@ -326,7 +271,6 @@ class Qtile(CommandObject):
                         bar.finalize()
         except:  # noqa: E722
             logger.exception('exception during finalize')
-            self.core.remove_listener()
 
     def _process_fake_screens(self):
         """
@@ -1542,7 +1486,7 @@ class Qtile(CommandObject):
     def cmd_get_state(self):
         """Get pickled state for restarting qtile"""
         buf = io.BytesIO()
-        pickle.dump(QtileState(self), buf, protocol=0)
+        self.dump_state(buf)
         state = buf.getvalue().decode(errors="backslashreplace")
         logger.debug('State = ')
         logger.debug(''.join(state.split('\n')))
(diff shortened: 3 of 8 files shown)


DorianGray:
CI failure seems to be unrelated :/

ramnes:
yeah, let's run them again

DorianGray:
ugh mypy, I'll need to get that working locally tomorrow

DorianGray:
Added the appropriate mypy hints

ramnes:
Why not just `atexit.register(qtile.maybe_restart)`?

DorianGray:
maybe_restart is a bound method. I wanted to give all objects, especially qtile and the things it owns, a chance to be finalized before os.execv.

ramnes:
Okay, that's what I thought. :)

You made a perfect transition to what I was going to say next with your last sentence, eh: I'd prefer everything in lifecyle.py to be in a class so that we could avoid the setters and getters that are just proxies over global objects. I think that we don't care having such a class instance still alive at exit. I was thinking of something along these lines:

```python
class Lifecyle:
    def __init__(self):
        self.behavior = Behavior.TERMINATE
        ...

    def teardown(self):
        if self.behavior = Behavior.RESTART:
            ...

lifecycle = Lifecycle()
atexit.register(lifecycle.teardown)
```

Then we could probably manage other things from Qtile's lifecycle... but I feel that's a bit redundant with SessionManager, so maybe we should just merge both features here?

PS: I do have toddlers yelling around as well, so I feel you!

DorianGray:
Fair enough! I'll make that change in a bit.

DorianGray:
Done!

ramnes:
Cool, but did you see the last part of my previous comment? :eyes:

> but I feel that's a bit redundant with SessionManager, so maybe we should just merge both features here?

libqtile/core/lifecycle.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self) -> None:
+        self.behavior: Behavior = Behavior.TERMINATE
+        self._execv_args: ExecVArgs = (
+            sys.executable,
+            tuple([sys.executable] + sys.argv),
+        )
+        atexit.register(self._finalize)
+
+    def set_execv_args(self, exe: str, args: Tuple[str, ...]) -> None:
+        self._execv_args = (exe, args,)
```
ramnes, line 28: I don't think we need this setter neither :P

m-col:
I agree with the merge with SessionManager. It is managing the session after all! This may involve some playing around with conftest.py.

DorianGray:
I'll see about merging lifecycle + session manager either tonight or tomorrow. I'll also go ahead and get rid of that setter... my reasons for keeping it were just to make the method signature match execv so it was kind of like a delayed call but that's probably not that valuable anyways.

DorianGray:
Done!

DorianGray:
segfault :/

m-col:
Is it feasible to have the new session manager logic contained within the session manager itself, rather than having this bi-directional relationship between it and the qtile object?

DorianGray:
qtile needs some way to signal the session manager that it would like to restart instead of terminate, what did you have in mind?

m-col:
Relevant IRC logs:
```
14:30 < dditthardt> reworking stop/restart in qtile looks to be very invasive
14:31 < dditthardt> would involve moving all the event loop code out of qtile and into session manager
14:31 < dditthardt> but yeah, that does seem like the way it should eventually be
14:32 < dditthardt> it'd be super nice to have the qtile object care about how it's started/stopped
...
14:35 < dditthardt> termination signal handling and things
14:36 < dditthardt> is in qtile right now
14:36 < dditthardt> belongs in session_manager imo
14:36 < mcol> yeah, that sounds right
14:36 < dditthardt> then we won;t have to pass the event loop around anymore
14:37 < dditthardt> I'm kind of excited to work on that, the question is do that now or later?
14:37 < mcol> OK, looking at it I definitely agree with you
14:38 < mcol> it makes sense as a precursor to the atexit stuff
14:38 < dditthardt> the dbus/glib event loop binding code belongs...somewhere for now too
14:39 < mcol> or perhaps alongise the atexit stuff - could just put it in the same PR
14:39 < mcol> it doesn't look like it would be that much work, right?
14:39 < mcol> there aren't many use of self._eventloop in manager.py
14:40 < dditthardt> would you mind posting that reasoning on the github issue? I'll give it a shot later today
14:40 < mcol> sure thing
14:40 < dditthardt> ty
14:41 < ramnes> dditthardt: did you see svig comment in the code? :)
14:41 < ramnes> https://github.com/qtile/qtile/blob/master/libqtile/core/session_manager.py#L48
14:42 < ramnes> and we're 3.7+ now :)
14:42 < dditthardt> yeah
14:42 < ramnes> I'd love to see this
14:42 < ramnes> would make the architecture much better imo
```

DorianGray:
planning on breaking out loop setup/teardown and atexit stuff into different files

DorianGray:
segfault but tests pass! invasive work done, now to make it prettier

DorianGray:
I should also mention that this slightly changes how different config hooks are fired.

For instance, in my config I was using asyncio.run as a crude method of having async handlers for hooks like startup and startup_once. With this change that no longer works, it errors informing the user that they cannot use asyncio.run inside of an existing event loop. Since my use case did not require awaiting a result, I simply changed asyncio.run to asyncio.create_task

DorianGray:
> I should also mention that this slightly changes how different config hooks are fired.
> 
> For instance, in my config I was using asyncio.run as a crude method of having async handlers for hooks like startup and startup_once. With this change that no longer works, it errors informing the user that they cannot use asyncio.run inside of an existing event loop. Since my use case did not require awaiting a result, I simply changed asyncio.run to asyncio.create_task

I am planning on adding support for async hooks in another PR

ramnes:
Can you please remove all the reformatting stuff and keep consistent with the existing style to ease the review? :)

DorianGray:
Done...I think? Please let me know if there's any reformatting that you want gone still, specifically.

DorianGray:
Tests pass in 3.7 and 3.8, segfault in 3.9 so it looks like my gitfu didn't break anything

libqtile/core/manager.py
```diff
@@ -152,16 +154,28 @@ def __init__(
...
+            pass
+        self.config.mouse += (Click([], "Button1", lazy.function(noop), focus="after"),)
 
-        self.current_screen = None
-        self.screens = []
+    def dump_state(self, buf):
+        try:
+            pickle.dump(QtileState(self), buf, protocol=0)
+        except:  # noqa: E722
+            logger.exception('Unable to pickle qtile state')
```
m-col, line 167: Can we catch this `Exception` and put it in the log message?

DorianGray: That's exactly what this code already does.

bin/qtile
```diff
@@ -28,7 +28,9 @@ this_dir = os.path.dirname(__file__)
 base_dir = os.path.abspath(os.path.join(this_dir, ".."))
 sys.path.insert(0, base_dir)
 
-from libqtile.scripts.main import main
+# import lifecycle early so no atexit handlers are lost on restart
+import libqtile.core.lifecycle  # noqa: F401, E402
+from libqtile.scripts.main import main  # noqa: E402
```
m-col, line 33: Are there any atexit handlers to be lost?

libqtile/backend/x11/xcore.py
```diff
@@ -206,7 +206,7 @@ def remove_listener(self) -> None:
     def _remove_listener(self) -> None:
         if self.fd is not None:
             logger.debug("Removing io watch")
-            loop = asyncio.get_event_loop()
+            loop = asyncio.get_running_loop()
```
m-col, line 209: I guess this is just following along with the changes in the asyncio API?

DorianGray: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

yeah, get_running_loop is preferred over get_event_loop

bin/qtile
DorianGray: not now, but potentially later there could be.

DorianGray: atexit handlers are run in FILO order

DorianGray: https://docs.python.org/3/library/atexit.html#atexit.register

libqtile/backend/x11/xcore.py
m-col: Great, thanks.

bin/qtile
m-col: Sorry you already mentioned this! I need to re-read PR threads before reviewing!

libqtile/core/manager.py
```diff
@@ -72,24 +71,17 @@ def _import_module(module_name, dir_path):
...
 
 
-def handle_exception(loop, context):
-    if "exception" in context:
-        logger.error(context["exception"], exc_info=True)
-    else:
-        logger.error("exception in event loop: %s", context)
-
-
 class Qtile(CommandObject):
```
m-col, line 74: It looks like most of the changes to this class are to do the way it sets up its attributes. I am 100% down with these changes but would prefer they be separate from the other changes. Not necesarilly a separate PR, but maybe a preceding commit. Does that work with the other changes, or do some changes only make sense done in paralle with these?

DorianGray: Almost all of them need to be done with the refactor since we can only add attributes to the class in __init__.

libqtile/core/manager.py
```diff
@@ -102,11 +94,22 @@ def __init__(
...
+        self.current_screen = None
+        self.screens = []
+        self._drag = None
+        self.mouse_map = None
+        self._state = state
+        self.server = IPCCommandServer(self)
+
+        self.run_config()
+
+    def run_config(self):
```
m-col, line 112: We really needed something like this, so glad it's here! Maybe change it to `load_config` or `read_config`? I think they might be more clear to a user, and I expect we will soon turn this into a `cmd_` function for live config reloading.

DorianGray: I'm gonna see about making another change I was thinking about along with this.

```
self.config = self.load_config(config)
```

however, it's possible that we might have some parts of init outside of qtile that depend on qtile.config being there when libqtile.init is called. It's possible I can reorder this to be MUCH cleaner, but unsure about how much breakage will occur.

libqtile/core/manager.py
m-col: OK, no worries then.

m-col:
I made a couple comments but in general this is all really good, so thanks.

The one thing that I'm hesitant about is how many different parts there are to just setting up a Qtile instance. SessionManager, Qtile, LifeCycle, and QtileLoop. LifeCycle cannot reference the qtile object so I understand why that would be separate, but I do wonder if there is an elegant way to have it merged with SessionManager without that issue. Likewise, I'd be in favour of moving the QtileLoop login into SessionManager. Thoughts?

DorianGray:
I'm definitely in favor of having eventloop setup separate from SessionManager mostly because I like lots of small, digestable, classes that do one thing only. Aside from that my reasoning was that since there is both setup and teardown involved, a context manager made the most sense to model that behavior. I suppose I could make session manager a context manager and have it's associated enter and exit functions set up the loop but even so I'd still personally like the logic in a separate class/file.

libqtile/core/manager.py
```diff
@@ -220,97 +230,37 @@ def conn(self):
...
-        except ImportError:
-            logger.warning("importing dbus/gobject failed, dbus will not work.")
-            self._glib_loop = None
-
     async def async_loop(self) -> None:
         """Run the event loop
 
         Finalizes the Qtile instance on exit.
         """
+        self.configure()
```
m-col, line 244: Is this the only place `configure` is called? If so, can you privatise it?

m-col:
Could you swap the changes you make in the two commits around? It seems odd to create `run_config` and then immeditately remove it. Can't you do the re-order first, without doing any lifecycle stuff, and then after those changes add the lifecycle commit?

libqtile/core/manager.py
DorianGray: fair enough

DorianGray:
Updated to separate the load_config breakout and __init__ reordering from the rest of the changes

m-col:
The `load_config` will (eventually) be converted into a method for live config reloading, and therefore having `self.config` set prior to running it is fine, as this will be the case mid session. Apart from removing that disclaimer beside the load_config call and the failing test (which is real), this all looks really good :)

DorianGray:
Yeah, looks like my last history modification lost my fix for that test :/ will push it back momentarily

m-col:
I'm getting 3 new exceptions when I SIGINT Qtile (using xephyr):
```
--- Logging error ---
  File "/usr/lib/python3.9/logging/handlers.py", line 73, in emit
    if self.shouldRollover(record):
  File "/usr/lib/python3.9/logging/handlers.py", line 191, in shouldRollover
    msg = "%s\n" % self.format(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 923, in format
    return fmt.format(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 659, in format
    record.message = record.getMessage()
  File "/usr/lib/python3.9/logging/__init__.py", line 363, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "/home/mcol/git/qtile/scripts/../bin/qtile", line 36, in <module>
    main()
  File "/home/mcol/git/qtile/libqtile/scripts/main.py", line 46, in main
    options.func(options)
  File "/home/mcol/git/qtile/libqtile/scripts/start.py", line 92, in start
    q.loop()
  File "/home/mcol/git/qtile/libqtile/core/session_manager.py", line 69, in loop
    asyncio.run(self.async_loop())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 629, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 596, in run_forever
    self._run_once()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1890, in _run_once
    handle._run()
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
  File "/home/mcol/git/qtile/libqtile/core/session_manager.py", line 76, in async_loop
    await self.qtile.async_loop()
  File "/home/mcol/git/qtile/libqtile/core/loop.py", line 48, in __aexit__
    logger.exception('exception in event loop:', *args)
Message: 'exception in event loop:'
Arguments: (None, None, None)
2021-01-02 23:45:11,440 libqtile start.py:start():L94  Qtile crashed
  File "/home/mcol/git/qtile/libqtile/scripts/start.py", line 92, in start
    q.loop()
  File "/home/mcol/git/qtile/libqtile/core/session_manager.py", line 69, in loop
    asyncio.run(self.async_loop())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/mcol/git/qtile/libqtile/core/session_manager.py", line 76, in async_loop
    await self.qtile.async_loop()
  File "/home/mcol/git/qtile/libqtile/core/loop.py", line 58, in __aexit__
    loop.run_until_complete(self._glib_loop)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 618, in run_until_complete
    self._check_running()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 578, in _check_running
    raise RuntimeError('This event loop is already running')
RuntimeError: This event loop is already running
2021-01-02 23:45:11,441 libqtile lifecycle.py:_atexit():L38  Qtile will now terminate
Future exception was never retrieved
future: <Future finished exception=AttributeError("'NoneType' object has no attribute 'is_set'")>
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
  File "/home/mcol/git/qtile/libqtile/core/loop.py", line 35, in gobject_thread
    while not self.qtile.is_stopped():
  File "/home/mcol/git/qtile/libqtile/core/manager.py", line 266, in is_stopped
    return self._stopped_event.is_set()
AttributeError: 'NoneType' object has no attribute 'is_set'
```

DorianGray:
These look like simple fixes \o/

DorianGray:
Not quite as simple as I thought. I ended up needing to add some additional loop cleanup routines because we were silently swallowing exceptions before. The loop context manager now gracefully terminates all remaining coroutines before shutdown.

DorianGray:
Confirmed dbus is working normally after these changes. Was able to use both the notification widget as well as elparaguayo's upower/dbus based battery widgets successfully with no errors. I was able to send notifications and see them displayed in my bar with both `notify-send` and `libqtile.utils::send_notification`.

libqtile/core/manager.py
```diff
@@ -59,61 +57,67 @@
...
+        self.current_screen = None
+        self.screens = []
+
+        libqtile.init(self)
+
+        self._event_loop = None
+        self.server = IPCCommandServer(self)
+        self.config = config
+        # config can potentially be replaced by the default config
+        self.config = self.load_config(config)
```
ramnes, line 99: I'd suggest any of the two:

```suggestion
        self.config = self.load_config(config)
```

```suggestion
        self.config = config
        self.load_config()
```

libqtile/core/manager.py
```diff
@@ -59,61 +57,67 @@
...
             logger.exception('Error while reading config file (%s)', e)
-            self.config = confreader.Config()
+            config = confreader.Config()
             from libqtile.widget import TextBox
-            widgets = self.config.screens[0].bottom.widgets
+            widgets = config.screens[0].bottom.widgets
             widgets.insert(0, TextBox('Config Err!'))
 
-        self.core.wmname = getattr(self.config, "wmname", "qtile")
+        self.core.wmname = getattr(config, "wmname", "qtile")
```
ramnes, line 111: I thought that you wanted to replace `self.config` by `config` everywhere to make the function pure, but you still left side effects behind, so what's the point?

libqtile/core/manager.py
```diff
@@ -131,23 +135,35 @@ def __init__(
...
+        config.mouse += (Click([], "Button1", lazy.function(noop), focus="after"),)
 
-        self.current_screen = None
-        self.screens = []
+        return config
+
+    def dump_state(self, buf):
+        try:
+            pickle.dump(QtileState(self), buf, protocol=0)
+        except:  # noqa: E722
```
ramnes, line 156: Why ignoring E722 exactly?

libqtile/core/manager.py
```diff
@@ -222,97 +219,42 @@ def conn(self):
...
-        self._eventloop.stop()
+            self.finalize()
+            self.core.remove_listener()
 
-    def maybe_restart(self) -> None:
-        """If set, restart the qtile instance"""
-        if self._restart:
-            logger.warning('Restarting Qtile with os.execv(...)')
-            os.execv(*self._restart)
+    def shutdown(self):
```
ramnes, line 237: I'd rather not rename this function if we can.

libqtile/core/manager.py
```diff
@@ -919,26 +860,26 @@ def call_soon(self, func, *args):
...
             func(*args)
             self.conn.flush()
-        return self._eventloop.call_later(delay, f)
+        return self._event_loop.call_later(delay, f)
 
     def run_in_executor(self, func, *args):
         """ A wrapper for running a function in the event loop's default
         executor. """
-        return self._eventloop.run_in_executor(None, func, *args)
+        return self._event_loop.run_in_executor(None, func, *args)
```
ramnes, line 882: I'm good with the renaming but please put it in another comment, it hurts the readability of the diff.

libqtile/core/session_manager.py
```diff
@@ -1,14 +1,24 @@
...
 class SessionManager:
     def __init__(
-        self, kore: base.Core, config, *, fname: str = None, no_spawn=False, state=None
+        self,
+        kore: base.Core,
+        config,
+        *,
+        fname: str = None,
+        no_spawn: bool = False,
+        state: bool = None,
```
ramnes, line 21: Please don't change code if not necessary, it clutters your commit and makes the history harder to navigate.

libqtile/core/session_manager.py
```diff
@@ -24,35 +34,43 @@ def __init__(
...
+        if path is None:
             # Dots might appear in the host part of the display name
             # during remote X sessions. Let's strip the host part first
-            display_name = kore.display_name
+            display_name = self.qtile.core.display_name
             display_number = display_name.partition(":")[2]
             if "." not in display_number:
                 display_name += ".0"
-            fname = ipc.find_sockfile(display_name)
+            path = ipc.find_sockfile(display_name)
```
ramnes, line 53: Again, I'm good with the renaming but please put it in another comment, it hurts the readability of the diff.

ramnes (changes requested):
Thank you so much for your work, it looks like it really improves the overall code workflow! That being said, and maybe because I find this PR really important, the diff is still a bit too complicated to get my approval for two reasons:

1. I feel that the interaction between the three classes LifeCycle, QtileLoop and SessionManager complicates the overall architecture for a small benefit, since both the LifeCycle and QtileLoop classes are only used in SessionManager. What kind of advantage do you see in the current implementation? Could we merge LifeCycle into QtileLoop and include it in session_manager.py, for example? 

2. Your small unnecessary modifications here and there really make it tedious to read the commits and will clutter the history if we merge them. I know it takes time on your side to change and I'm sorry I have to ask you to work on this since it makes me feel like I'm nitpicking, but with a codebase of this size it's really important that everyone does atomic commits or digging through history becomes a nightmare when we need to.

I really want us to get this done and I'm sure it's going to be a big improvement for Qtile overall, so feel free to come bug me on IRC if you want a faster feedback than here. :)

libqtile/core/manager.py
DorianGray: I didn't add this, just moved it. Bare except is almost never acceptable... unless you're restoring a pickle =) E722 is that linting error.

ramnes: Oh right, didn't see it was there already. Why do you say it's acceptable for pickles?

libqtile/core/manager.py
ramnes: Or this:

```suggestion
        self.load_config(config)
```

libqtile/core/manager.py
DorianGray: They can throw literally anything
https://docs.python.org/2/library/pickle.html#pickle.UnpicklingError

libqtile/core/lifecycle.py
```diff
@@ -0,0 +1,43 @@
...
+]
+
+Behavior = enum.Enum('Behavior', 'NONE TERMINATE RESTART')
+
+
+class LifeCycle:
+    '''
+    Be very careful about what references this class owns. Any object
+    referenced here when atexit fires will NOT be finalized properly.
+    '''
```
ramnes, line 20: This should be a comment, not a docstring. :)

libqtile/core/manager.py
ramnes: Yeah but they still only raise BaseException children, don't they? They wont send us a KeyboardInterrupt or alike, and catching these is probably one of the things that currently prevents us from quitting the tests, for example. :)

ramnes (approved):
LGTM :rocket:

tych0:
After thinking about this some more, I'm still not convinced that this is the right way to solve the problem. It seems the argument in favor of it was "we run the python garbage collector this way, which fixes things". However, we shouldn't *need* to run the garbage collector: reclaiming memory isn't necessary before exec().

So, this suggests that whatever bug this patch was fixing is a poor use of `__del__` or a missing finalize() call somewhere. From the other issue,

> I believe this would even cause, for example, an async with statement to not have it's context cleaned up

Shouldn't we be closing this manually on shutdown via the code in the finally block of `SessionManager.loop`?

tych0:
Yes, but read the code -- the finally block happens before the exec().

DorianGray:
Now I have no idea what you're talking about. Can you provide some context? snippets?

tych0:
Sorry, I'd hoped, 

>  in the finally block of SessionManager.loop

was enough:

https://github.com/qtile/qtile/blob/f3add880b5b6271b7a57931d8ebc6f0c094fd54b/libqtile/core/session_manager.py#L50

PR #2047, opened 2020-12-27
Remove a deprecated workaround

ramnes:
We don't support Python 3.5 nor 3.6 anymore. :)


ramnes:
Not so deprecated after all :)

PR #2050, opened 2020-12-28
Allow notifications to be deleted

tchar:
Hello,

I updated the NotificationManager class to allow Notification deletion. I tried to be consinstent with the current code.
Here is a brief overview of what is changed/added:

- Inner changes:
    - Notifications are held under an OrderedDict for fast retrieval during delete.
    - self.notifications is now a property that returns a list as previously to allow backwards compatibility
    - introduce a dict to allow indexing of previous/next notifications for fast retrieval of prev and next.
    - introduce a lock to provide concurrency when deleting/adding notifications when multiple subscribers are registered.
- Methods:
    - register_delete: Same as register but for delete.
    - delete: Deletes the notifications provided a Notification object or notification ids and calls the delete callbacks.
    - delete_all: Deletes all the notifications and calls the delete callbacks.
    - prev/next: Since now notification index in the list cannot be infered from the notification id (may be deletions), these methods fetch the next/prev based on a Notification object or a notification id.
 
Notification:
- I added an Urgency Enum class inside the Notification class to get the Urgency from a Notification object (e.g Notification.Urgency(notif) == Notification.Urgency.LOW)

Notify widget:
- I updated the Notify widget to comply with Notification deletion with methods similar to update and real_update (delete and real_delete)
- Right click deletes current notification as previously.
- I enabled Left click to delete all notifications.

Existing API does not change, only inner working/logic.

Please let me know if you want any changes on this, if you need extra tests and if yes where is best to declare them.

Cheers.

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -26,6 +26,10 @@
     If dbus is available, this module implements a
     org.freedesktop.Notifications service.
 """
+from collections import OrderedDict
+from enum import Enum
+from threading import RLock, Thread, current_thread, main_thread
+
 from libqtile.log_utils import logger
 
 try:
@@ -36,6 +40,7 @@ try:
     has_dbus = True
 except ImportError:
     has_dbus = False
+
 
 BUS_NAME = 'org.freedesktop.Notifications'
 SERVICE_PATH = '/org/freedesktop/Notifications'
@@ -82,6 +87,20 @@ if has_dbus:
 
 
 class Notification:
+
+    class Urgency(Enum):
+        LOW = 0
+        NORMAL = 1
+        CRITICAL = 2
+
+        def _missing_(self):
+            try:
+                # Default priority is NORMAL
+                priority = self.hints.get('urgency', 1)
+                return Notification.Urgency(int(priority))
+            except (AttributeError, ValueError):
+                return Notification.Urgency.NORMAL
+
     def __init__(self, summary, body='', timeout=-1, hints=None, app_name='',
                  replaces_id=None, app_icon=None, actions=None):
         self.summary = summary
@@ -96,9 +115,16 @@ class Notification:
 
 class NotificationManager:
     def __init__(self):
-        self.notifications = []
-        self.callbacks = []
+        # Use OrderedDict to maintain order
+        self._notifications = OrderedDict()
+        # Use a dict for indexing for quick finding prev/next notification
+        self._notifications_idx = dict()
+        self._current_id = 0
+        self.callbacks_add = []
+        self.callbacks_delete = []
         self._service = None
+        # Use a lock to protect notification concurrency.
+        self._lock = RLock()
 
     @property
     def service(self):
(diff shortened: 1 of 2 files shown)


libqtile/notify.py
```diff
@@ -26,6 +26,10 @@
     If dbus is available, this module implements a
     org.freedesktop.Notifications service.
 """
+from collections import OrderedDict
```
DorianGray, line 29: As of python 3.7 dict is ordered by order of insertion. It looks like this code relies on insertion order in the same way so OrderedDict might not be required. Python 3.7 is the minimum python version supported by qtile.

The only thing you would need to change about the code is your use of reversed. values will need to be converted to a list before it can be reversed. In python 3.8 that isn't the case, so a comment for removal later might be nice =)

tchar: Oh here is something I didn't know about dict in newer versions.

I will drop the OrderedDict.

What I did not get is the comment about reversing. I am using reverse only to get the last key (which is a notification id). I don't actually care about the value (now that I see it line [217](https://github.com/qtile/qtile/blob/4127cf35bcce2a989647cfade159c1571edf6c51/libqtile/notify.py#L217) is obsolette.

Also did not get the part about python 3.8 (maybe because I did not get the comment about the values in reverse).

Thanks for the comment.

m-col: I also did not know this 👀

libqtile/notify.py
```diff
@@ -110,22 +136,174 @@ def service(self):
...
+            notifs_deleted = self._notifications.values()
+            self._notifications = OrderedDict()
+            self._notifications_idx = {}
+
+        for callback in self.callbacks_delete:
+            # Don't let the subscriber crash us
+            try:
+                callback(*notifs_deleted)
+            except Exception as e:
+                logger.error(e)
```
m-col, line 306: Could you change this to `logger.exception` so that the traceback is logged?

libqtile/notify.py
```diff
@@ -96,9 +115,16 @@ def __init__(self, summary, body='', timeout=-1, hints=None, app_name='',
...
+        # Use OrderedDict to maintain order
+        self._notifications = OrderedDict()
+        # Use a dict for indexing for quick finding prev/next notification
+        self._notifications_idx = dict()
+        self._current_id = 0
+        self.callbacks_add = []
+        self.callbacks_delete = []
         self._service = None
+        # Use a lock to protect notification concurrency.
+        self._lock = RLock()
```
m-col: Do we need all the `threading` stuff? Won't it all be fine without, running within qtile's async loop?

libqtile/notify.py
```diff
@@ -213,19 +212,20 @@ def add(self, notif):
         with self._lock:
             # Find the id of the previous notification
             if self._notifications:
-                prev_id = next(reversed(self._notifications))
-                prev_id = self._notifications[prev_id].id
+                # Python 3.7 compatibility. For python 3.8 and later ommit
+                # the list since dict and dict_keys are reversible.
+                prev_id = next(reversed(list(self._notifications.keys())))
```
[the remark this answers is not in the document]
tchar: > list-ing a dictionary already yields just the keys:
 Yeah you are right, I already knew that I wanted to put the keys to basically make it explicit that they key is the id. I will fix it.

> You can also simplify the next-reversed just by indexing:
When I wrote this i was between what you wrote and this. I just wanted to leave it like this because creating a list just to take the last element of a dict is not actually the most efficient thing (but required for Python 3.7). It is just there to indicate that in case Qtile changes minimum Python version requirement, list should be dropped and use next(reversed()) for efficiency. I could also make version checking but I think it is an overkill since notifications are not going to be coming bursting.

I will make the changes and update the comment.

libqtile/notify.py
```diff
@@ -82,6 +87,20 @@ def GetServerInformation(self):  # noqa: N802
...
+        CRITICAL = 2
+
+        def _missing_(self):
+            try:
+                # Default priority is NORMAL
+                priority = self.hints.get('urgency', 1)
+                return Notification.Urgency(int(priority))
+            except (AttributeError, ValueError):
+                return Notification.Urgency.NORMAL
+
```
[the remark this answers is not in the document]
tchar: I initially had it in the top level. I decided to put it under Notification to follow the same structure as [this](https://github.com/qtile/qtile/blob/1b0406b472c9285978e7b57c451a9438c5070a00/libqtile/utils.py#L240)

Also this is being used in the Notify widget like `Notification.Urgency(notification) == Notification.Urgency.CRITICAL`. The `Notification` itself has the information about the Urgency inside the `hints` so I thought there was no point adding it explicitly inside the `Notification` object, I just wrote this so you can extract the Urgency from the Notification from an API like interface. If your only concern is the Error silencing I can remove that.

If we add the urgency attribute in the Notification I will make changes to [this](https://github.com/qtile/qtile/blob/ef6e3e30d5de29ef9ea5efc85183bae280065c62/libqtile/notify.py#L66) also.

Let me know which of the two you prefer.

libqtile/notify.py
tchar: Sure!

libqtile/notify.py
tchar: Oh I knew that I was going to have this conversation :) . It is the only way I see that it will be concurrent. Let me elaborate a bit.
- add: Runs on a separate thread than `main_thread` by default. So you cannot avoid concurrency issues even with one subscriber.
- delete: Should run on a separate thread than `main_thread` for the same reasons as `add`.

Example:

Widget clicks delete and at the same time new notification arrives. Even if delete runs on main thread (which it should not), add runs on a separate thread so there should be concurrency so that variables/attributes do not get messed up.

And this a concurrency issue with just the widget as a subscriber. If you have more than one subscriber created by a user, that potentially itself runs on a separate thread so it does not block Qtile, things get more complicated without locking.

The implementation will not have a performance impact whatsoever with any Use Case, unless you benchmark it (which is not the point anyways) with 1000 notification access/add/delete per second with multiple subscribers.

Now one comment I have on this is that reads block when they don't have to be necessarily. We could 
- use a read write lock like [this](https://pypi.org/project/readerwriterlock/) to improve performance but it adds a  dependency on qtile.
- or I could even write it on my own (maybe not the best idea) using something like Semaphores. But in the end. this will provide slight to almost no performance boost in Use Cases where notifications are not bursting.

Let me know what you think.

Cheers

m-col:
Please could you elaborate on why these changes are necessary? What issue were you finding with the current code?

tchar:
> Please could you elaborate on why these changes are necessary? What issue were you finding with the current code?

Ok let me elaborate. Current code does has two main problems.

1.  Does not allow notifications to be deleted from memory once the user dismisses them.
  - Just as an example. A notification object from the e-mail client I use is about 14-15 Mb in memory. Now if I get about 20 e-mails and dismiss them from the widget, this is about 300Mb that stays in memory (about half of the memory that Qtile itself occupies in my machine).
2. Does not allow other Notification subscribers to communicate dismisses. With the current code:
  - I use the widget on the bar.
  - User writes some code that subscribes to the NotificationManager (so NotificationManager has 2 subscribers) for himself.
  - Dismiss a notification from widget > Notification still appears in user defined subscriber
  - Dismiss a notification from user defined subscriber > Notification still appears in the widget

ramnes:
Can you please split 1 and 2 into two different pull requests? I'm definitely interested in 1, not so sure about 2, and mixing both makes it harder to review (and thus approve). Also by reading the diff I feel that there is more involved than these two things.

tchar:
> Can you please split 1 and 2 into two different pull requests? I'm definitely interested in 1, not so sure about 2, and mixing both makes it harder to review (and thus approve). Also by reading the diff I feel that there is more involved than these two things.

Hello,

Well `1` and `2` are not 2 different implementations combined into the same Pull Request. By implementing `1`, you almost automatically get `2`. Let me explain:

 - `1` is: If you want to allow notification deletion you basically need all the concurrency protection in the `NotificationManager` that this code implements because you don't know when notifications are added/deleted.
 - `2` is: Just `1` with callback calling for deletion (last 5 lines of code in `NotificationManager.delete`) .

What I could split is basically:
- `1` and `2` into the first Pull Request
- `widget.Notify` class polishing I did and the new `Notification.Urgency` class into another (which is maybe why you see a lot of potentially unrelated things changed).

Would you mind splitting it that way?

ramnes:
Thank you for the explanation, it's much more clear now. But I'm still interested in 1 and not so sure about 2. :)

The point is not the number of lines of code but the amount of thoughts and discussions we'll have to go through before merging. Keeping both features in the same pull request is not going to help — if you want to get active and complete reviews from the benevolent people that run Qtile, it's in your interest to make the work easier for them. I strongly suggest you to open a PR for 1, then another for 2 even if that's just five lines of code, then another one for any other atomic problem you'd like to solve.

And I tend to think that the concurrency protection you're mentioning should not be in a first PR, but in another one that showcase the problem with a failing test. Qtile's codebase is getting really big and we're trying hard to keep it as simple as possible, so please understand that lines of code that may look like premature optimization are a red flags for us at the moment, especially without tests alongside.

I don't want to be rough and close the discussion, but I'd still like to avoid that we go back and forth too much on this and not spend hours debating unnecessarily. Please come discuss on IRC if you feel there's a need, that would be simpler. :)

tchar:
I see. I already have the tests tha fail and I thought that it would be stupid to make a PR with a failing code.
I understad you need small incremental changes to verify them.
I'll see what I can do when I have the time. Closing this since it's not gonna be merged.

Cheers

PR #2052, opened 2020-12-29
Make the Makefile self-documented

ramnes:
I found by accident that the default rule of the Makefile was outdated and listing some rules that do not exist anymore, so rather than updating it, I thought we could have it self-documented.

This hack is based on this article (and its comments): https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html


ramnes:
image

ramnes:
image

m-col:
Nice, I like it. LGTM.

PR #2055, opened 2020-12-31
add ci for pypy3.7

DorianGray:


DorianGray:
Tests pass...but segfault :/

DorianGray:
finally have real test failures here

DorianGray:
Turns out github's pypy is built against old tkinter - the reason for the test failures. Not sure of a workaround, so skip relevant tests.

LMK if there's a better replacement for tkinter we can use in place...something that we can control the version of.

m-col:
There you go https://github.com/qtile/qtile/pull/2108

DorianGray:
> There you go #2108

You are the best

DorianGray:
This now depends on https://github.com/qtile/qtile/pull/2108

DorianGray:
Added documentation on python VM differences in install guide.

docs/manual/install/index.rst
```diff
@@ -27,8 +27,18 @@ Installing From Source
 First, you need to install all of Qtile's dependencies (although some are
 optional/not needed depending on your Python version, as noted below).
 
-Note that Python 3 versions 3.5 and newer are currently supported and tested,
-including corresponding PyPy3 versions.
+Note that Python 3 versions 3.7, 3.8, and 3.9 are currently supported and
+tested, including corresponding PyPy3 versions(currently only PyPy3.7). There
```
tych0, line 31: I wonder if we shouldn't just put a link to the right place in `.github/workflows/ci.yaml` vs. explicitly listing versions here since things will inevitably get out of date again.

tych0:
One minor comment, but other than that LGTM. I merged the other PR so you can rebase on that as well.

docs/manual/install/index.rst
DorianGray: That's fair, I'll do it in a bit.

DorianGray: done!

DorianGray: I linked to the folder, since the multiple ci workflows PR changes this.

PR #2058, opened 2021-01-01
Remove useless base classes

ramnes:
`SingleWindow` and `Delegate` are only used by the Slice layout for years, so let's simplify everything and remove them while moving their methods to where they really belong.


tcld:
For `SingleWindow` I wholeheartedly agree, but I did like the idea of `Delegate` and always saw `Slice` as more of a demo-implementation, just waiting for somebody with some really creative layouting ideas to come along. You are probably right in removing it, however.

ramnes:
Yeah I really don't mind if we revert this commit when needed. Should I make one commit per class?

libqtile/layout/slice.py
```diff
@@ -50,6 +50,18 @@ def remove(self, window):
...
+                screen_rect.x, screen_rect.y,
+                screen_rect.width, screen_rect.height,
+                0,
+                None,
+            )
+            window.unhide()
+        else:
+            window.hide()
+
     def _get_window(self):
```
tcld, line 65: While you're at it...do we really need this method?

tcld:
One commit per class sounds like a good middleground, I think. Although wouldn't that mean that, if Mr. Layout came along in seven years, somebody would have to remember about that `Delegate`-commit?

I might be overthinking this. I like `Delegate` as a concept, but less code is less code.

m-col:
In my humble opinion we shouldn't keep code 'just in case' nor should we count
on remembering. So +1 to removing any way!

libqtile/layout/slice.py
ramnes: Nop. :)

ramnes:
Edited the branch to split in two commits and remove the useless _get_window method.

tcld:
Thanks for taking the time, looks good to me. :)

PR #2061, opened 2021-01-02
module imp is deprecated as of py3.5.

DorianGray:
This changes libqtile/core/manager.py::_import_module to use a full path
instead of dir path as well as switches from imp to importlib.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 
 import asyncio
+import importlib.util
 import io
 import logging
 import os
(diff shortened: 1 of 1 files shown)


m-col:
Nice, thanks. Pls could you include that message in the commit message?

m-col:
Looks like it is used in one place only, and if it's 3 lines, then those lines may as well just be moved into that method `Qtile.cmd_run_external`.

m-col:
Also, `imp` is deprecated in favour of `importlib`. I'm happy to merge this anyway and change that myself, unless you fancy it?

DorianGray:
will fix!

m-col:
Awesome thanks!

m-col:
I've been trying to create a test for `Qtile.cmd_run_external` but, at least with my python installation, it doesn't work because `sys.exc_clear` isn't a thing. Investigating to see if that means it doesn't work for anything, and therefore nobody is using it, and therefore we could consider just removing it. It was added in 2014 with no meaningful additions since.

m-col:
According to [this](https://docs.python.org/3/whatsnew/3.0.html?highlight=sys%20exc_clear), `sys.exc_clear` was removed in Python 3.0 (!!!) and therefore `Qtile.cmd_run_external` hasn't been used (successfully) since Qtile dropped Python 2 support.

Maybe we should just drop that method altogether @tych0 @ramnes?

m-col:
The qshell appears to use `run_external` when giving a custom file to run, but it gives me an `Invalid command` message when trying to use that argument.

m-col:
This would look like this: https://github.com/qtile/qtile/pull/2064

ramnes:
Yeah I'd rather remove the method completely. @DorianGray, do you have a use case that really needs to revive this command or did you just fix that while looking around the code?

DorianGray:
Just ran across it while refactoring manager init. Will close this

PR #2062, opened 2021-01-02
Add generalised get_core function to backends

m-col:
This begins the move towards compartmentalising the backends, necessary
to have them configurable. It adds a hidden argument to `qtile start`
that allows for setting the backend to something other than x11,
removing the hard-code of the x11 backend in the script. This option is
then used to get the `Core` from the specified backend, and uses this
for the new session.


libqtile/backend/x11/xcore.py
```diff
@@ -55,7 +55,7 @@ class ExistingWMException(Exception):
     pass
 
 
-class XCore(base.Core):
+class Core(base.Core):
```
ramnes, line 58: Then let's rename xcore.py to core.py as well?

libqtile/scripts/start.py
```diff
@@ -133,4 +134,11 @@ def add_subcommand(subparsers):
         dest='state',
         help='Pickled QtileState object (typically used only internally)',
     )
+    parser.add_argument(
+        '-b', '--backend',
+        default='x11',
+        dest='backend',
+        choices=libqtile.backend.CORES,
+        help=argparse.SUPPRESS,
```
ramnes, line 142: Why the SUPPRESS? Because we only have one core at the moment? I think it still have value to show the argument in the help, maybe it would encourage people to implement other backends. :)

m-col:
I wanted to, but figured there could be people who may have imports that'll
break with that. But it would be nicer and would remove the need for the import
in `__init__.py` if `get_core` imports from `core.py` in all backends.

m-col:
That reply was to the `core.py` comment -- email replies to inline comments clearly don't end up where they should!

libqtile/scripts/start.py
m-col: Yep that was my thinking, and that it might add a bit of noise, but, I guess you're right!

ramnes:
What's your point? Aren't you breaking imports by renaming the class anyway?

m-col:
🤔 yes...

m-col:
Alrighty, both those changes have been added.

ramnes:
Thanks, we just miss another core now. 😄

PR #2063, opened 2021-01-02
Implement async hook subscriptions

DorianGray:
This allows us to @hook.subscribe to coroutines as well as normal
functions. I've modified the hook.fire routine to handle this as well as
allow for scheduling coroutines on the event loop if one exists in the
current thread while creating, running, and destroying one if it
doesn't. This allows firing hooks from both sync and async contexts.

Async hooks are useful because it removes the headache of writing the
async boilerplate themselves for our users. For instance, in the case
where someone wants to sleep before running some code when a hook fires,
they would merely `await asyncio.sleep(time)` and not block the event
loop, whereas to do so without these changes they would need to add the
boilerplate that is now a part of hook.fire to do so.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -30,6 +30,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import asyncio
+import contextlib
 from typing import Dict, Set
 
 from libqtile import utils
@@ -373,6 +375,17 @@ class Unsubscribe(Subscribe):
 unsubscribe = Unsubscribe()
 
 
+def _fire_async_event(co):
+    loop = None
+    with contextlib.suppress(RuntimeError):
+        loop = asyncio.get_running_loop()
+
+    if loop is None:
+        asyncio.run(co)
+    else:
+        asyncio.ensure_future(co)
+
+
 def fire(event, *args, **kwargs):
     if event not in subscribe.hooks:
         raise utils.QtileError("Unknown event: %s" % event)
@@ -380,7 +393,12 @@ def fire(event, *args, **kwargs):
         logger.debug("Internal event: %s(%s, %s)", event, args, kwargs)
     for i in subscriptions.get(event, []):
         try:
-            i(*args, **kwargs)
+            if asyncio.iscoroutinefunction(i):
+                _fire_async_event(i(*args, **kwargs))
+            elif asyncio.iscoroutine(i):
+                _fire_async_event(i)
+            else:
+                i(*args, **kwargs)
         except:  # noqa: E722
             logger.exception("Error in hook %s", event)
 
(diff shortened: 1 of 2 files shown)


ramnes (approved):
LGTM

DorianGray:
added tests

DorianGray:
tests are breaking all the tests for some reason.

DorianGray:
This requires lifecycle because of it's improved event loop cleanup and error handling.

DorianGray:
added the ability to fire async hooks before and after the event loop is started/terminated. Makes using them less brittle.

libqtile/core/manager.py
```diff
@@ -195,6 +157,59 @@ def noop(qtile):
...
+        for grp in config.groups:
+            if isinstance(grp, ScratchPadConfig):
+                sp = ScratchPad(grp.name, grp.dropdowns, grp.label)
+                sp._configure([config.floating_layout],
+                              config.floating_layout, self)
+                self.groups.append(sp)
+                self.groups_map[sp.name] = sp
+
+        # It fixes problems with focus when clicking windows of some specific clients like xterm
+        # 2020-12-31: unsure if this still fixes anything
```
tych0: Can we just drop it and see if people complain? Either that or drop this additional comment. I don't think it adds any value.

DorianGray: wrong PR =)

tych0:
Can you talk a bit about the motivation for this in the commit messages? It's not clear why we want this.

DorianGray:
absolutely! Soon™

ramnes:
Here is a good example (I think?) of why we'd need this: https://groups.google.com/g/qtile-dev/c/DYVYuAzYbG4/m/qkx7wU9WAgAJ

tych0:
Can you spell it out for me?

ramnes:
Some applications like Spotify do not have a their WM_CLASS or WM_NAME set properly when they're spawned, only a few milliseconds later after when they update themselves. So if you want to match against these, you have to wait a bit, and doing so synchronously will block Qtile. Running the hooks in the event loop (and thus supporting async functions) is a good way to not block the whole window management for a matching rule.

ramnes (approved):

tych0:
Seems like that would be good analysis for the commit message :)

DorianGray:
Yeah, I haven't rewritten the messages yet. I have a habit of pushing often so interested parties can follow along. I think that's all that's left but I need to test more, see if it needs more tests.

DorianGray:
There, minus the spurious CI failure this should be ready for merge.

ramnes (approved):

ramnes:
Approved for the third time and re-ran tests for the Nth time.

tych0:
Generally I feel like "fixing race conditions with sleeps" is not a good motivating use case nor something we should advertise (perhaps we should create a hook that fires when wm class is changed, so people can move windows to where they should go when these things are set).

However, this seems useful to me in other contexts, so LGTM.

ramnes:
We're on the exact same page. :)

That's exactly what I suggested here as well, and I do agree that it would be a much cleaner solution: https://github.com/qtile/qtile/issues/1915#issuecomment-697553164

But yes, being able to run async functions is most likely going to be useful for many other things.

PR #2064, opened 2021-01-02
Drop run external

m-col:
This command has not worked for some time, so instead of fixing it we
can just remove it. Similar functionality can be achieved by
implementation desired code in one's config.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -29,7 +29,6 @@ import subprocess
 import sys
 import tempfile
 import time
-import traceback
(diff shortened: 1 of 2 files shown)


ramnes:
Wow, I didn't know about this command. It's quite nice actually, but if it's untested an undocumented, let's focus on the other supported way to extend Qtile, we have too many. LGTM.

ramnes (approved):

DorianGray:
Closed the other PR. Kill it with fire!

Issue #2059, opened 2021-01-01
KeyChord documentation


ramnes:
KeyChord really needs a few lines in the docs! Please help if you're using it.

Relevant issues and PR:
* #1190 
* #1208 
* #1777

elParaguayo:
@ramnes I can probably draft something for this.

tcld:
I'm looking forward to some documentation. Tried my hands on the chords, but the handling was quite awkward, so I likely missed something. (I noticed we have a widget related to chords - maybe I should have used it?)

PR #2065, opened 2021-01-03
Add KeyChord documentation

elParaguayo:
Addresses issue #2059

### DIFF
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -45,6 +45,82 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
        'C': 'control',
     }
 
+KeyChords
+=========
+
+Qtile also allows sequences of keys to trigger callbacks. In Qtile, these
+sequences are known as chords and are defined with
+:class:`libqtile.config.KeyChord`. Chords are added to the ``keys`` section of
+the config file.
+
+::
+
+    from libqtile.config import Key, KeyChord
+
+    keys = [
+        KeyChord([mod], "z", [
+            Key([], "x", lazy.spawn("xterm"))
+        ])
+    ]
+
+The above code will launch xterm when the user presses Mod + z, followed by x.
+
+.. warning::
+    Users should note that key chords are aborted by pressing <escape>. In the
+    above example, if the user presses Mod + z, any following key presses will
+    still be sent to the currently focussed window. If <escape> has not been
+    pressed, the next press of x will launch xterm.
+
+Modes
+-----
+
+Chords can optionally specify a "mode". When this is done, the mode will remain
+active until the user presses <escape>. This can be useful for configuring a
+subset of commands for a particular situations (i.e. similar to vim modes).
+
+::
+
+    from libqtile.config import Key, KeyChord
+
+    keys = [
+        KeyChord([mod], "z", [
+            Key([], "g", lazy.layout.grow()),
+            Key([], "s", lazy.layout.shrink()),
+            Key([], "n", lazy.layout.normalize()),
+            Key([], "m", lazy.layout.maximize())],
+            mode="Windows"
+        )
+    ]
+
+In the above example, pressing Mod + z triggers the "Windows" mode. Users can
+then resize windows by just pressing g (to grow the window), s to
+shrink it etc. as many times as needed. To exit the mode, press <escape>.
+
+.. note::
+    If using modes, users may also wish to use the Chord widget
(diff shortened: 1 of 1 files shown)


docs/manual/config/keys.rst
```diff
@@ -45,6 +45,82 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
...
+        ])
+    ]
+
+The above code will launch xterm when the user presses Mod + z, followed by x.
+
+.. warning::
+    Users should note that key chords are aborted by pressing <escape>. In the
+    above example, if the user presses Mod + z, any following key presses will
+    still be sent to the currently focussed window. If <escape> has not been
+    pressed, the next press of x will launch xterm.
```
m-col, line 72: So with this configuration, if I hit `mod+z` and then `y`, the mode remains until `escape` or `x` is hit, passing anything else through?

docs/manual/config/keys.rst
```diff
@@ -45,6 +45,82 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
...
+            Key([], "plus", lazy.layout.grow()),
+            Key([], "minus", lazy.layout.shrink()),
+            Key([], "n", lazy.layout.normalize()),
+            Key([], "m", lazy.layout.maximize())],
+            mode="Windows"
+        )
+    ]
+
+In the above example, pressing Mod + z triggers the "Windows" mode. Users can
+then resize windows by just pressing g (to grow the window), s to
```
m-col, line 96: I think `g` and `s` here are meant to be `plus` and `minus`, right?

docs/manual/config/keys.rst
```diff
@@ -45,6 +45,82 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
...
+    from libqtile.config import Key, KeyChord
+
+    keys = [
+        KeyChord([mod], "z", [
+            KeyChord([], "x", [
+                Key([], "c", lazy.spawn("xterm"))
+            ])
+        ])
+    ]
+
```
m-col, line 120: Jesus lol, this is hilarious. I guess this is a side effect of how they are implemented, rather than this being specifically desired from the beginning? What use cases do people have for this?

docs/manual/config/keys.rst
elParaguayo: Sigh... Thought I'd fixed that before pushing. The plus didn't work when testing the code so I changed to "'s" and "g". Will fix.

docs/manual/config/keys.rst
m-col: Commenting again after reading the modes bit below. What I understood from the description is that in the above (without a `mode`), when the first key inside the mode is hit, the mode is exited. And below (with a `mode`) the mode remains until `escape` is hit, even if you hit some of the nested key bindings. Is that right?

elParaguayo: Yes. See [this thread](https://groups.google.com/u/1/g/qtile-dev/c/r87i-fdIW2o/m/dpy8bowIAQAJ) on the qtile-dev group for more.

I might try and take a look at fixing the bug later but, for now, I've put the warning in the docs.

docs/manual/config/keys.rst
elParaguayo: The question came up on the [qtile-dev group](https://groups.google.com/u/1/g/qtile-dev/c/DxbRCqtw-s8/m/nHTZWZXhCQAJ) so I included an example here.

I don't think it was envisaged but, as you say, it works because of the way the code is implemented.

docs/manual/config/keys.rst
elParaguayo: Just say your second comment (we posted at the same time). 

With modes, once you enter the mode, you stay in it until you leave by pressing escape. While in the mode, you can then trigger subcommands numerous times.

Without a mode, the chord ends once a valid combination has been pressed. However, the bug is that the chord does not end with an invalid key press.

m-col: Is that a bug? Sounds like a feature!

Thanks for the clarification -- if I was right then the docs are clear enough :D

docs/manual/config/keys.rst
m-col: I like it.

m-col:
Could you squash the two commits into one please? Then LGTM.

m-col:
Awesome, thank you for this!

ramnes:
Thanks!

PR #2069, opened 2021-01-04
Added option to change empty group string for the window name widget

Barbarossa93:
Hello,

In this PR, I've added an option for the window name widget to change the string to display when the desktop/group is empty. Previously, it was simply " ". The default is the same, but specifying empty_group_string in the config.py will allow the user to change it to whatever they wish, e.g. "Desktop".

This is my second ever PR so bear with me if I take some time to fix any needed changes.

As an aside, thanks for creating qtile! It's a great opportunity to learn/play around with python!

### DIFF
--- a/libqtile/widget/windowname.py
+++ b/libqtile/widget/windowname.py
@@ -30,10 +30,20 @@ from libqtile.widget import base
 
 class WindowName(base._TextBox):
     """Displays the name of the window that currently has focus"""
+
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('show_state', True, 'show window status before window name'),
-        ('for_current_screen', False, 'instead of this bars screen use currently active screen')
+        ("show_state", True, "show window status before window name"),
+        (
+            "for_current_screen",
+            False,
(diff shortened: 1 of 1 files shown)


libqtile/widget/windowname.py
```diff
@@ -56,14 +66,17 @@ def update(self, *args):
...
             elif w.minimized:
-                state = '_ '
+                state = "_ "
             elif w.floating:
-                state = 'V '
-        unescaped = "%s%s" % (state, w.name if w and w.name else " ")
+                state = "V "
+        unescaped = "%s%s" % (
+            state,
+            w.name if w and w.name else self.empty_group_string,
```
tych0, line 79: Can you drop all the other formatting changes and just do this?

Barbarossa93: I believe I just fixed it. Let me know if I missed anything!

tych0:
Looks good to me; I did a manual squash and added a bit of your explanation to the commit message. Thanks for the patch!

ramnes:
TIL we had such a widget. Thanks for the patch!

PR #2072, opened 2021-01-05
Allow resizing of Drawer() pt2, electric boogaloo

DorianGray:
proto docs. Cairo to the rescue! Cairo implements a RecordingSurface
that we can use locally while the underlying Drawable may not be
accessible. These changes make the Drawer write to a RecordingSurface
and only copy to the pixmap on Drawer.draw.

This also fixes our random segfaults in CI since the XCBSurface is
created on draw instead of on __init__

### DIFF
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -10,6 +10,7 @@
 # Copyright (c) 2014 Nathan Hoad
 # Copyright (c) 2014 dequis
 # Copyright (c) 2014 Tycho Andersen
+# Copyright (c) 2020, 2021 Robert Andrew Ditthardt
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


DorianGray:
This should handle 2 problems with the previous attempt.
1. drawer.pixmap now exists again since the systray widget requires it. We should investigate ways to remove that requirement.
2. performance problems in draw

tych0:
Can you put your second comment in the commit message itself?

DorianGray:
of course, once it's ready for merge I will.

m-col:
To summarise what we were trying to troubleshoot on IRC: when running with this commit and doing a bunch of operations that involve a drawer, Qtile becomes slow and laggy. Specifically I was changing group lots of times and gradually the delay between hitting the keybinding and the group and GroupBox change increased to quite long, in some tests getting up to a second before I stopped.

DorianGray:
After a few passes on this, I've come up with a nice way to allow drawer's to be resizeable AND to ensure that our external resources are ready before we try to draw to them. This now uses a cairo RecordingSurface to stage all drawing operations in local memory in qtile and only writes them to the underlying XCBSurface on draw.  After draw, the RecordingSurface is recreated so we do not leak these drawing operations between draws.

We must keep a surface available at all times for drawing since that's how the users of Drawer expect it. RecordingSurface are unbounded, meaning they can be literally ANY size, and store operations in a very memory efficient way. The drawing operations are stored as calls to the high level surface functions with arguments, and can be replayed/painted to any other surface type.

I decided against an approach where I did a direct copy between 2 pixmaps to preserve the drawing operations because this required I destroy the XCBSurface regardless. The memory used by a pixmap + xcbsurface is more than the memory used by a recordingsurface. If we need to keep the drawing operations somewhere, might as well use the most efficient means. The code required to juggle 2 pixmaps was unwieldy and error prone as well.

DorianGray:
In trying to fix the CI errors, I determined that using a RecordingSurface to stage drawing operations is also the fix for that...therefore https://github.com/qtile/qtile/pull/2077 is unecessary.

DorianGray:
If you wouldn't mind, please pull these changes to your local and smoke test, especially @m-col since he had trouble with the last pass. I have tested as well as I can locally.

DorianGray:
I should also note that this PR will allow us to draw to drawers without X with a few small modifications. I would imagine that is extremely useful for unit testing.

DorianGray:
@tych0 I made the changes you requested to the commit message. Please let me know if you would like anything else included.

m-col:
> I decided against an approach where I did a direct copy between 2 pixmaps to preserve the drawing operations because this required I destroy the XCBSurface regardless. The memory used by a pixmap + xcbsurface is more than the memory used by a recordingsurface. If we need to keep the drawing operations somewhere, might as well use the most efficient means.

Maybe I have misunderstood, so bare with me in case I have. Currently we use a pixmap, controlled by an XCBSurface. If we need resize with those, we need to create another pixmap and XCBSurface, CopyArea between them (or something between the two surfaces), then destroy the old pixmap and XCBSurface. Then we continue on as usual. With the RecordingSurface, we have to destroy the pixmap and XCBSurface and create new ones, then later let the RecordingSurface deal with recreating its contents. I'm not sure I see the advantage here, because the latter approach has one extra object. The latter approach as implemented also involves destroying and creating a new cairo contex upon every draw, right? What is this for? Also, what stops the RecordingSurface from growing and growing in memory?

DorianGray:
> > I decided against an approach where I did a direct copy between 2 pixmaps to preserve the drawing operations because this required I destroy the XCBSurface regardless. The memory used by a pixmap + xcbsurface is more than the memory used by a recordingsurface. If we need to keep the drawing operations somewhere, might as well use the most efficient means.
> 
> Maybe I have misunderstood, so bare with me in case I have. Currently we use a pixmap, controlled by an XCBSurface. If we need resize with those, we need to create another pixmap and XCBSurface, CopyArea between them (or something between the two surfaces), then destroy the old pixmap and XCBSurface. Then we continue on as usual. With the RecordingSurface, we have to destroy the pixmap and XCBSurface and create new ones, then later let the RecordingSurface deal with recreating its contents. I'm not sure I see the advantage here, because the latter approach has one extra object. The latter approach as implemented also involves destroying and creating a new cairo contex upon every draw, right? What is this for? Also, what stops the RecordingSurface from growing and growing in memory?

Look for calls to _reset_surface, we finish and recreate the recording surface on each draw.

This way we can keep just a single pixmap around instead of 2 for when we have to copy. Copying the pixmap when width or height is changed is not feasible, and we must keep at least one "live" pixmap around as the backing for the surface. In one of my previous comments I also mention that because of the recordingsurface, we can make this code easily draw to devnull for unit tests as well.

m-col:
> Look for calls to _reset_surface, we finish and recreate the recording surface on each draw.

This seems like unnecessary additional work, and consider that `draw` is called very frequently.

> Copying the pixmap when width or height is changed is not feasible

Why not?

> We must keep at least one "live" pixmap around as the backing for the surface

We only need one pixmap total, except during the (infrequent as far as I can tell, in fact never required so far) resizing operations, where we can create a new one, copy contents, then destory the old one.

> In one of my previous comments I also mention that because of the recordingsurface, we can make this code easily draw to devnull for unit tests as well.

Tests are obviously important but here I think performance is critical.

DorianGray:
> > Look for calls to _reset_surface, we finish and recreate the recording surface on each draw.
> 
> This seems like unnecessary additional work, and consider that `draw` is called very frequently.
> 
It's a simple object,think of a RecordingSurface as a list of operations. We're simply appending operations to the list and then painting them when we actually want to, all at once, on draw.
We then forget about that list when draw is finished and make a new list, the old list is gced at python's leisure.

You can create and dispose of literally millions of recordingsurfaces a second without a sweat.
> > Copying the pixmap when width or height is changed is not feasible
> 
> Why not?
> 
Because you'd end up copying twice whenever width and height was changed together, and in the case of a large internal window being resized that copy operation would be extremely heavy. I run qtile on a 4k screen... very very heavy

> > We must keep at least one "live" pixmap around as the backing for the surface
> 
> We only need one pixmap total, except during the (infrequent as far as I can tell, in fact never required so far) resizing operations, where we can create a new one, copy contents, then destory the old one.
> 
Right, this is mostly for future use(tooltip windows with changing content,  dynamic menus, etc) but it fixes our CI issues now and allows the Popup class to be used for things that aren't just static content.

> > In one of my previous comments I also mention that because of the recordingsurface, we can make this code easily draw to devnull for unit tests as well.
> 
> Tests are obviously important but here I think performance is critical.
I agree, but I think we can have both

DorianGray:
So, before our drawing pipeline looked like this:

fill pixmap with whatever, whenever.
on draw, copy the pixmap to the window

Now it looks like this
fill recordingsurface with whatever, whenever
write drawing operations to pixmap
copy pixmap to window

once the systray widget doesn't require direct access to drawer.pixmap anymore, that changes to
fill recordingsurface wherever, whenever
on draw, write drawing operations to window

DorianGray:
@m-col made a good point about only drawing when there are operations to draw. I've modified the code so that we check the recording surface for operations before we attempt to paint them. Now our drawing pipeline is nearly the same for all existing use cases and it also works for future ones.

DorianGray:
here is a screenshot of demonstrating some of the use cases for a resizeable drawer.
https://imgur.com/a/o4jkahI

ramnes (approved):
LGTM

@m-col, did you try this latest version of the branch?

m-col:
I get this error when I restart (restarting does succeed):
```
2021-01-13 19:42:56,203 ERROR libqtile base.py:on_done():L542 Failed to reschedule.
  File "/home/mcol/git/qtile/libqtile/widget/base.py", line 534, in on_done
    self.update(result)
  File "/home/mcol/git/qtile/libqtile/widget/base.py", line 550, in update
    old_width = self.layout.width
  File "/home/mcol/git/qtile/libqtile/drawer.py", line 80, in width
  File "/home/mcol/git/qtile/libqtile/pangocffi.py", line 132, in get_pixel_size
    pango.pango_layout_get_pixel_size(self._pointer, width, height)
```

m-col:
And `qtile` seems to be running around ~40% CPU according to htop

m-col:
Oh actually I'm getting high CPU on master now. Will investigate...

m-col:
The CPU change was 34a3b2193498366ce19e4f4d57aa6b0d99922cde.

I did notice a couple of bugs though:
image

The top is master, and the bottom is this branch, showing my bar (ignore the thinner lines, they are `imv`'s window borders). There are two differences:

1. the `CPUGraph` widget has a strange glitch that has caused vertical lines appearing
2. the top has 4 dark purple trims: 2 at the ends, and 2 just inside the ends. This is just a single texbox widget (`trim = widget.TextBox(' ', background=xcolors[8])`) which I place into the widget list in 4 places, causing the other 3 to be `Mirror` widgets. So it seems that mirror widgets are not being handled correctly.

AFAICT the lag-causing issue isn't here anymore which is great. I'll keep running this branch for a couple hours just in case but that's probs fine.

DorianGray:
Thanks! I'll take a look

DorianGray:
Ah, I see the issue with mirror. It's not tied to the original drawer's draw call so the surface is empty when it's used. I'm going to see about moving that logic into drawer(seems to belong there anyway).

DorianGray:
Fixed Mirror widgets. @m-col , would you mind sending me exactly what `colors[8]` is? I think this is related to alpha.

m-col:
Their working fine!

DorianGray:
Updated to fix _Graph rendering. We now support proper vector rendering through cairo as a side effect of this PR, so floating point pixels actually matter.

m-col:
Can confirm, works great, LGTM.

DorianGray:
@ramnes @tych0 I know you have already taken a look at this but if it's not too much trouble please take one more look and also if possible smoke test this with your configs?

tych0:
I don't love setters generally since it makes control flow hard to reason about, but I don't think that's enough to block this PR. It seems like it Mostly Works, so let's let-er-rip and see what people complain about :)

PR #2073, opened 2021-01-05
Columns: make margin_on_single accept ints

m-col:
This makes the `margin_on_single` Columns option, rather than accepting
a boolean to specify whether it should have margins or use those set my
`margin`, accept an `int` so that different margins can be set on single
windows. `-1` can be used to tell the layout to use that specified by
`margins`, which until the previous commit would be default behaviour,
so this restores that.

This commit also adds the setting to the changelog.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -118,7 +118,7 @@ class Columns(Layout):
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw a border when there is one only window."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
-        ("margin_on_single", False, "Add margin to the layout for only one window."),
+        ("margin_on_single", -1, "Margin when only one window. `-1` means use `margin`."),
         ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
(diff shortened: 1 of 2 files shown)


m-col:
You can almost tell that my commit message was a stream of consciousness

leschalu:
it's even better like that, if it had uniform margin like xmonad does it would be the best layout

libqtile/layout/columns.py
```diff
@@ -118,7 +118,7 @@ class Columns(Layout):
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw a border when there is one only window."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
-        ("margin_on_single", False, "Add margin to the layout for only one window."),
+        ("margin_on_single", -1, "Margin when only one window. `-1` means use `margin`."),
```
ramnes, line 121: Let's keep consistent for users that already use a margin:

```suggestion
        ("margin_on_single", 0, "Margin when only one window. `-1` means use `margin`."),
```

ramnes:
Why would you want a margin other than your normal margin when there is only one window? Why not doing the same for the border?

m-col:
I like having a larger margin when I have only one window because I hardly ever need the single window to take up the whole screen (minus regular margins), so it centres it and makes it a more appropriate size and is way more pretty doncha think?

As for the border -- yep, i suppose the border could be changed to do the same but nobody has asked for it! I will happily make that change too for you if you want it ;)

libqtile/layout/columns.py
m-col: Well the setting was only introduced to master 30 minutes before this PR was submitted, and that commit itself changed the default setting for everyone by making `margin_on_single` equal false, where before single windows used `margin` and therefore did have margins if there were any. So `-1` restores the default of using `margin`'s size for single windows.

ramnes:
Nah right, let's wait for someone that really wants it.

libqtile/layout/columns.py
ramnes: Wut? If I have margin=10 from an older version of Qtile and get this with -1 as default, I'll have a margin on single windows when I didn't have one before.

ramnes: My point being that no, Columns didn't use the margin for single windows before.

ramnes: My bad, it did. :)

ramnes:
Thanks!

PR #2075, opened 2021-01-07
Rename fname to socket_path in SessionManager and ipc.py

DorianGray:


libqtile/ipc.py
```diff
@@ -107,18 +107,18 @@ def pack(msg: Any, *, is_json: bool = False) -> bytes:
...
         ----------
-        fname : str
+        socket : str
             The file path to the file that is used to open the connection to
             the running IPC server.
         is_json : bool
             Pack and unpack messages as json
         """
-        self.fname = fname
+        self.socket = socket
```
tych0, line 121: How about socket_path instead, since this isn't the actual socket?

DorianGray: fair, will just require some reformatting

DorianGray:
done

tych0:
Thanks!

PR #2076, opened 2021-01-07
ThreadedPollText: suppress exception when no loop

DorianGray:
Reported by ramnes: this exception can happen when the thread attempt to
call qtile.call_soon_threadsafe after the event loop has already been
finalized. This was always happening, but the exception was being lost
before the lifecycle changes.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -511,6 +511,12 @@ class ThreadedPollText(InLoopPollText):
                 text = self.poll()
                 if self.qtile is not None:
                     self.qtile.call_soon_threadsafe(self.update, text)
+            except RuntimeError:
+                logger.debug(
+                    'problem polling to update widget %s (ignoring, this is '
+                    'most likely because the event loop has finalize already)',
+                    exc_info=True,
+                )
             except:  # noqa: E722
                 logger.exception("problem polling to update widget %s", self.name)
         # TODO: There are nice asyncio constructs for this sort of thing, I


ramnes:
Mhh, I think I'd prefer a try-except block with at least a debug log.

DorianGray:
done.

libqtile/widget/base.py
```diff
@@ -510,7 +510,16 @@ def worker():
             try:
                 text = self.poll()
                 if self.qtile is not None:
+                    # This can throw a runtime error if the event loop is
+                    # not available. This would only happen before qtile is
+                    # started or after it is finalized. The thread here is
+                    # fire and forget, so cannot be signaled.
                     self.qtile.call_soon_threadsafe(self.update, text)
+            except RuntimeError:  # gets raised when the event loop is down
+                logger.debug(
```
tych0, line 519: Can we make this `logger.exception` instead? Then we'll get a stack trace and we can track it down.

DorianGray: This only happens when the Thread started on like 527 outlives the event loop, which happens on shutdown. It's debug just to quiet the logs until ThreadedPollText goes away.

DorianGray: I mean...the right fix is to make ThreadedPollText go away but that's already done in the dbus PR.

DorianGray: I should also note that this is an exception only seen since the lifecycle PR. It's one of the ones we were swallowing before.

tych0: So if we don't care about these logs, is there any point in logging them? @ramnes?

libqtile/widget/base.py
```diff
@@ -510,7 +510,16 @@ def worker():
...
+                    'Thread tried to access event loop while it was unavailable',
```
ramnes, line 520: We don't know this for sure, a RuntimeError could happen for other reasons, and this is why I'd prefer a debug log in the first place. I'd rather write something like: `problem polling to update widget %s (ignoring, this is most likely because the event loop has finalized already)`. And then you could probably remove the other comments. :)

libqtile/widget/base.py
ramnes: https://github.com/qtile/qtile/pull/2076/files#r553595736

libqtile/widget/base.py
ramnes: Or we can match against the error message and re-raise if it's different than what we expect.

tych0: Agreed; IMO a stack trace is useful to track these sorts of things down, so let's include that if we can.

DorianGray:
closed in favor of https://github.com/qtile/qtile/pull/2084

PR #2080, opened 2021-01-07
Remove Session Manager

DorianGray:


ramnes:
LGTM to me but do you have any specific reason, other than Dependency Injection is Great™?

DorianGray:
Not really, just cleanup that I had thought about doing while tinkering with the lifecycle changes.

Session manager has no business looking at the config, kore, no_spawn, or state parameters. I'm planning on moving _should_restart into qtile after, have it talk directly to lifecycle instead of a flag. Should I include that in this PR to make it make more sense?

DorianGray:
There, went ahead and finished the related changes.

DorianGray:
got rid of 12 lines =)

libqtile/core/manager.py
```diff
@@ -197,6 +198,16 @@ def _configure(self):
...
         hook.fire("startup_complete")
 
+    @property
+    def display_name(self):
+        # Dots might appear in the host part of the display name
+        # during remote X sessions. Let's strip the host part first
+        name = self.core.display_name
+        display_number = name.partition(":")[2]
+        if "." not in display_number:
+            name += ".0"
```
m-col, line 208: Could you move this part into the core? It is X specific

m-col:
LGTM

libqtile/core/manager.py
DorianGray: I'll need to move it and add an ipc_socket_path property to core as well, is that ok? My reasoning is that when we have a separate core there will be a separate way of naming the socket for the ipc server now managed by Qtile.

m-col: Actually thinking about it again, this string manipulation is only required when finding the socket file, _not_ when we are using the display of `Qtile` (necessarily). I.e. it is also used by command clients to communicate with `Qtile`. So maybe just put it as a standalone function like it was originally but inside `libqtile/backend/x11/__init__.py` or somewhere like that.

DorianGray: RE irc: I'll take a look at how the command clients do things and try to emulate them while I try to move this logic into the backend.

DorianGray:
This should fix https://github.com/qtile/qtile/issues/2087 as well

libqtile/core/manager.py
m-col: Actually I think you can completely remove this property and just use `self.core.display_name` on line 211. The lines 223-225 are redundant as they are also within `ipc.find_sockfile`. If we need to change this in the future due to new cores then we can deal with it if/when we need to.

DorianGray:
strangely, CI seems more stable on this branch but I have no idea why.

DorianGray:
not just more stable...passes every time. wat

libqtile/core/manager.py
```diff
@@ -701,8 +724,7 @@ def _focus_by_click(self, e):
 
     def process_button_click(self, button_code, state, x, y, event) -> None:
         self.mouse_position = (x, y)
-        k = self.mouse_map.get(button_code)
-        for m in k:
+        for m in self.mouse_map.get(button_code, []):
```
DorianGray, line 727: For anyone wondering why this change is here... mypy found a bug!

ramnes (approved):
LGTM

libqtile/scripts/start.py
```diff
@@ -89,7 +90,7 @@ def start(options):
     rename_process()
     q = make_qtile(options)
     try:
-        q.loop()
+        asyncio.run(q.loop())
```
ramnes, line 93: Can we make the API simpler and keep `loop` as a synchronous function that calls `asyncio.run` itself, as `SessionManager` used to?

DorianGray:
Updated based on @ramnes feedback

PR #2084, opened 2021-01-08
Remove ThreadedPollText

elParaguayo:
ThreadedPollText and ThreadPoolText are very similar. The latter
uses asyncio eventloop rather than threading and so is preferred
here.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -30,7 +30,6 @@
 # SOFTWARE.
 
 import subprocess
-import threading
 from typing import Any, List, Tuple
 
 from libqtile import bar, configurable, confreader, drawer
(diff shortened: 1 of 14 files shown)


ramnes:
Great, thanks!

m-col:
This commit seems to shoot up CPU usage attribute to `qtile`, according to `htop`.

elParaguayo:
That's odd. Not aware of any reason that should be the case. Which widgets (that were updated via this PR) are you running? Will see if I can replicate here.

m-col:
`Maildir`, and `Battery`

m-col:
No sorry, just `Battery` I think

m-col:
`Battery` normally, but I just removed it and the problem persists. `GenPollText`, when removed, appears to be the culprit!

m-col:
lmao i got it. I was using `GenPollText` to check that a server I run is up by making and closing a quick connection. It looks like now `GenPollText` will continuously run the `func` it is given nonstop over and over with no delay. Previously it must have had some update interval, whose default value I was relying. I've added an `update_interval` setting now.

elParaguayo:
OK - sounds like something that should be added as a default... If that's the case, I'll sort in a new PR.

m-col:
Yep just confirming in the code, previously it's superclass had `update_interval` set to `600` by default, and now it is nothing. So if it is possible then it make sense to add that as the default with the new version too

elParaguayo:
OK - will do. Sorry for missing that.

elParaguayo:
See #2105

PR #2086, opened 2021-01-09
Add new groups when too few for number of screens

### LINKED ISSUE #475: Don't crash if there are less groups than screens
[insert traceback here]

m-col:
This makes Qtile create more groups when setting up if there are fewer
groups than there are screens. Fixes #475.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -112,9 +112,7 @@ class Qtile(CommandObject):
             warnings.warn("Defining a main function is deprecated, use libqtile.qtile", DeprecationWarning)
             self.config.main(self)
 
-        if self.config.groups:
-            self.dgroups = DGroups(self, self.config.groups,
-                                   self.config.dgroups_key_binder)
+        self.dgroups = DGroups(self, self.config.groups, self.config.dgroups_key_binder)
 
         if self.config.widget_defaults:
             _Widget.global_defaults = self.config.widget_defaults
@@ -288,9 +286,15 @@ class Qtile(CommandObject):
             if not self.current_screen:
                 self.current_screen = scr
 
-            scr._configure(
-                self, i, x, y, w, h, self.groups[i],
-            )
+            if len(self.groups) < i + 1:
+                name = f"autogen_{i + 1}"
+                self.add_group(name)
+                grp = self.groups[i]
+                logger.warning(f"Too few groups in config. Added group: {name}")
+            else:
+                grp = self.groups[i]
+
+            scr._configure(self, i, x, y, w, h, grp)
             self.screens.append(scr)
 
     def paint_screen(self, screen, image_path, mode=None):
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -215,6 +215,17 @@ def test_keypress(manager):
     assert manager.c.groups()["a"]["focus"] == "one"
 
 
+class TooFewGroupsConfig(ManagerConfig):
+    groups = []
+
+
+@pytest.mark.parametrize("manager", [TooFewGroupsConfig], indirect=True)
+@pytest.mark.parametrize("xephyr", [{"xinerama": True}, {"xinerama": False}], indirect=True)
+def test_too_few_groups(manager):
+    assert manager.c.groups()
+    assert len(manager.c.groups()) == len(manager.c.screens())
+
+
 class _ChordsConfig(Config):
     groups = [
         libqtile.config.Group("a")


libqtile/core/manager.py
```diff
@@ -288,9 +286,15 @@ def _process_screens(self) -> None:
             if not self.current_screen:
                 self.current_screen = scr
 
-            scr._configure(
-                self, i, x, y, w, h, self.groups[i],
-            )
+            try:
```
DorianGray, line 289: mind if this is an `if len(self.groups) > i:` instead of a `try ... except`? exceptions in python are relatively expensive compared to a branch.

DorianGray (changes requested):

libqtile/core/manager.py
m-col: Good point, cheers

libqtile/core/manager.py
```diff
@@ -288,9 +286,15 @@ def _process_screens(self) -> None:
             if not self.current_screen:
                 self.current_screen = scr
 
-            scr._configure(
-                self, i, x, y, w, h, self.groups[i],
-            )
+            if len(self.groups) < i + 1:
+                name = f"group_{i + 1}"
+                self.add_group(name)
```
DorianGray, line 291: Now that I think about this more, this only solves the case when there is one more screen than groups. I think we should either add enough groups to match the number of screens or we should allow screens to be groupless.

m-col: > this only solves the case when there is one more screen than groups

does it?

DorianGray: ...am I missing something? It doesn't iterate and only calls self.add_group once O.o

DorianGray: nvm I am a dumb and apparently cannot read

DorianGray (approved):

libqtile/core/manager.py
```diff
@@ -288,9 +286,15 @@ def _process_screens(self) -> None:
             if not self.current_screen:
                 self.current_screen = scr
 
-            scr._configure(
-                self, i, x, y, w, h, self.groups[i],
-            )
+            if len(self.groups) < i + 1:
+                name = f"group_{i + 1}"
```
ramnes, line 290: Why not just `str(i)`? Because numbers are common names and you don't want collision? Feels like a hack to me, but I don't have much better to offer. :(

m-col: Yep exactly, just reducing the change of a collision. I figure as well that if someone is naming their groups "group_1, group_2" etc, then they certainly will not have a "group_5" when there are fewer than 5 groups, so that is much less likely than someone having a group just named "5". So maybe bit of a hack, buuuuuuuuuut :)

libqtile/core/manager.py
m-col: :D honestly i read it over a few times trying to figure out what mistake i had made lol

ramnes:
What would you think of implementing groups at the screen level, that could only be used on one screen? We could then name these default groups after the screen (for example "eDP1"), and could guarantee that we don't collide depending on how we design things. That would also help for all the people coming from other WM and asking for sticky groups. But that's much more work. :)

DorianGray:
That makes a ton of sense, plus this default group per screen could be used for ungrouped windows, giving all windows a group and fixing issues like https://github.com/qtile/qtile/issues/2092

libqtile/core/manager.py
tych0: What about "autogen_1"?

m-col: sure thing, changed

ramnes (approved):

m-col:
I thought I replied to this, woops. This was waiting for me to hit 'comment':

I functionally use groups this way, but I don't think it is something we should implement for these reasons:

- it is a big paradigm change from decoupled groups and screens, which itself is a legit setup used by some other popular WMs (and so implementing it is opinionated)
- it would be disruptive; many people ask for this behaviour but presumably many more people do not.
- As many people ask for this behaviour, I often point people towards a gist of the configuration that I use and no doubt there are many other implementations, customised to exactly the behaviour that the author desires.
- ^ that last point hints at the customisability of the current implementation, which we'd be sacrificing. It is pretty easy to get this behaviour already, but implementing it within libqtile would likely make it more difficult for people who don't want this to get the current default behaviour.

My suggest instead that we add little snippets that we often discuss on IRC or in GH threads to the wiki or somewhere else that is easily modified by anyone. I guess the examples repo exists for this purpose but the wiki would be well suited to much small snippets of code so might work well.

tych0:
Yes, I agree with @m-col here, this is not something we should implement at the qtile level. I use things the other way now, and IMO it's one of the major selling points.

PR #2090, opened 2021-01-11
Move state.py to core

DorianGray:
QtileState is only used by Qtile and Qtile is in core. I don't see a
reason why a normal user would want to tinker with QtileState.


tych0:
Can you describe why this is a good move?

Also - I see some other refactoring PRs. What's the state of your CI stuff? I'd really like to get back to reliable CI before we embark on many more refactorings.

DorianGray:
https://github.com/qtile/qtile/pull/2072 - fixes CI

DorianGray:
added a small comment to the commit message explaining why this is a good move.

ramnes:
Seems kinda obvious to me. Thanks!

ramnes:
Just for information, this will break Qtile when restarting from a previous commit to upgrade:

```
2021-01-13 17:05:44,293 WARNING libqtile manager.py:maybe_restart():L273 Restarting Qtile with os.execv(...)
2021-01-13 17:05:45,163 ERROR libqtile manager.py:_configure():L183 failed restoring state
  File "qtile/libqtile/core/manager.py", line 180, in _configure
    st = pickle.load(f)
ModuleNotFoundError: No module named 'libqtile.state'
```

This was to be expected, but maybe we should add a note in the changelog for the next release.

PR #2091, opened 2021-01-11
Move command_*.py into new command module

DorianGray:
It kind of defeats the purpose of the module system to prefix module
names.


DorianGray:
@m-col suggested I add a deprecation warning for the old imports in irc.

flacjacket:
This is a much nicer structure here, thanks!

flacjacket (approved):

tych0:
Can you also add a note to the CHANGELOG about breakage?

DorianGray:
Done

m-col:
I think only the `command_client` and `command_interface` modules need to continue working, the others are just used internally.

Could you also update the docs? Specifically [this page](http://docs.qtile.org/en/latest/manual/commands/index.html) and [this page](http://docs.qtile.org/en/latest/manual/commands/scripting.html)?

DorianGray:
Done =)

ramnes (approved):

m-col:
Thanks!

PR #2093, opened 2021-01-11
pytest-runner is deprecated, remove.

DorianGray:
pytest-runner depends on deprecated features in setuptools per
https://pypi.org/project/pytest-runner/. This removes it without any
other effects.


ramnes (approved):
LGTM

tych0:
Huh, I didn't realize using `setup_requires` was bad. Seems like we should figure out how to drop it from xcffib too, then.

Issue #2089, opened 2021-01-10
Dgroups broken in current master


dequis:
# Issue description

I tried git master to see if another thing is already fixed there, but i found that dgroups doesn't work anymore. It just sends everything to the current group and doesn't create new ones.

v0.16.1 works, master doesn't. Did bisect, fce22e59fceb23310ea03feaa13a0aa5b4f802a9 is the first bad commit

# Qtile version

fce22e59fceb23310ea03feaa13a0aa5b4f802a9 (first bad commit)
33ef59f88fe9655b82b4b3b232ae1f8252f2cd65 (current master)

# Stack traces

 It shows this when starting a client with rules (gimp in this case)

```
2021-01-10 21:20:07,198 libqtile hook.py:fire():L385  Error in hook client_new
Traceback (most recent call last):
  File "/home/dx/test/gits/qtile/libqtile/hook.py", line 383, in fire
    i(*args, **kwargs)
  File "/home/dx/test/gits/qtile/libqtile/dgroups.py", line 169, in _add
    if rule.matches(client):
  File "/home/dx/test/gits/qtile/libqtile/config.py", line 672, in matches
    return self.match.compare(w)
  File "/home/dx/test/gits/qtile/libqtile/config.py", line 631, in compare
    if not match(rule_value):
  File "/home/dx/test/gits/qtile/libqtile/config.py", line 603, in predicate
    return value and any(match(v) for v in value)
  File "/home/dx/test/gits/qtile/libqtile/config.py", line 603, in <genexpr>
    return value and any(match(v) for v in value)
  File "/home/dx/test/gits/qtile/libqtile/config.py", line 602, in <lambda>
    match = getattr(other, 'match', lambda v: other in v)
TypeError: 'in <string>' requires string as left operand, not list
```

# Configuration

```
dgroups_app_rules = [
    Rule(Match(
        wm_class=['MPlayer', 'Exe', 'feh', 'Pinentry', 'Plugin-container',
                  'Gtk-recordMyDesktop', 'Beep.py']),
        float=True, intrusive=True, break_on_match=False),
    Rule(Match(title=['Event Tester', 'Firefox Preferences']),
        float=True, intrusive=True),
    Rule(Match(wm_class=['Gnome-terminal', 'st-256color', 'st']), group='term'),
    Rule(Match(wm_class=['fontforge']), group='fontforge', float=True),
    Rule(Match(wm_class=['Gimp', 'gimp']), group='gimp', float=True),
[...]

PR #2094, opened 2021-01-11
Fix issue #2089

DorianGray:
issue caused in https://github.com/qtile/qtile/pull/2084
simple logic error, fixes https://github.com/qtile/qtile/issues/2089

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -605,7 +605,7 @@ class Match:
         elif name == 'wm_class':
             def predicate(other):
                 # match as an "include"-match on any of the received classes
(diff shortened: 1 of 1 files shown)


m-col:
Cool thanks. Any chance we could get a test so it won't break again?

DorianGray:
As soon as I can get the tests running locally, absolutely!

ramnes:
Let's make @dequis happy and merge this as is. A test in another PR would still be welcome. :)

dequis:
<3

DorianGray:
I think there are more bugs in this code. I have additional commits related on my local including tests! Should be ready soonish

PR #2105, opened 2021-01-14
Fix default update_interval in ThreadPoolText

elParaguayo:
Previous default of "None" caused widgets to constantly update
resulting in high CPU usage. New default is update interval of
1 minute.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -445,7 +445,7 @@ class _TextBox(_Widget):
 class InLoopPollText(_TextBox):
     """ A common interface for polling some 'fast' information, munging it, and
     rendering the result in a text box. You probably want to use
-    ThreadedPollText instead.
+    ThreadPoolText instead.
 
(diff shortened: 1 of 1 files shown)


tych0:
I think the previous default was 600, any reason to diverge from that?

elParaguayo:
10 mins seemed long to me but, if that's the case, I'm free to change it as a user! 

Have changed to old number.

ramnes (approved):

elParaguayo:
Just thinking about this - let me check the old widgets that used `ThreadPoolText` to make sure this change doesn't break them. I think it's sensible to have a default timer as opposed to None to prevent the CPU issue but some of the widgets may not have that problem.

elParaguayo:
All ok. 5 widgets used `ThreadPoolText` before change and all of these set a default `update_interval` so nothing to change.

tych0:
Great, thanks!

PR #2106, opened 2021-01-15
Multiple CI Workflows

DorianGray:
On a test run, this dropped our total CI time from 7m20s to 4m55s an gives us the ability to re-run individual segments.


DorianGray:
depends on https://github.com/qtile/qtile/pull/2102

ramnes (approved):

.github/workflows/test_3-8.yml
```diff
@@ -0,0 +1,23 @@
...
+        steps:
+            - uses: actions/checkout@v2
+            - name: Set up python 3.8
+              uses: actions/setup-python@v2
+              with:
+                  python-version: "3.8"
+            - name: install dependencies
+              run: |
+                  sudo apt update
+                  sudo apt install graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors
```
tych0, line 20: This seems to duplicate a lot of stuff for not a lot of value IMO (CI being too slow is not a big blocker for this project since we switched to github actions). Is there any way we can at least share what packages need to be installed?

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
                   python-version: ${{ matrix.python-version }}
             - run: |
                 sudo apt update
-                sudo apt install graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors
+                sudo apt install libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors
```
tych0: +1 to this, if you make a separate PR I'm happy to take it.

.github/workflows/test_3-8.yml
DorianGray: unfortunately no, github actions is working on features to make this setup less bloated. This is the recommended way to both increase parallelization and allow re-running of individual segments :/

.github/workflows/ci.yml
DorianGray: yeah, we weren't actually doing static analysis on large sections of the code because of these missing libraries.

DorianGray: I can make this a separate PR I thnk

.github/workflows/test_3-8.yml
DorianGray: related discussion : https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/35

ramnes: We can share the package that needs to be installed with a base Docker image I think.

DorianGray: That's a good idea, want me to build one? It'll need a special place to be hosted on dockerhub.

tych0: I guess what I'm wondering is: is there a reason to merge anything like this at all? What does it achieve, and is that worth the additional complexity?

tych0: (These are all great things to put in commit messages :)

ramnes: It is in the commit message actually:

> This breaks out python versions into their own workflows to allow re-running only a single version.

tych0: I saw that, but that seems like a solution in search of a problem. I currently do not have the need to be able to run the tests individually. Adding lots of copypasta or another point of failure in the form of a dependency on a docker registry is not worth it to solve a problem we don't have.

.github/workflows/ci.yml
DorianGray: As I noted in IRC, actually this is in another PR: https://github.com/qtile/qtile/pull/2102

tych0: But that has the composable requirements as well :). Any chance we can get this alone?

.github/workflows/test_3-8.yml
ramnes: The tests are somehow reliable since the last fixes, but just before that it was a nightmare to have them complete. We're doing fancy stuff with processes and alike, so the likelihood of race problems is pretty high. When it's broken, it's nice to be able to run the tests individually as a last resort while the tests are being fixed.

tych0: > When it's broken, it's nice to be able to run the tests individually as a last resort while the tests are being fixed.

I'd prefer not to merge stuff for this case. Re-running the tests to get things green is not a habit I think we want to get into. We should fix the tests instead.

tych0: If that blocks merging things because merging things is painful, so be it. That will encourage more people to look at fixing things :)

.github/workflows/ci.yml
DorianGray: done https://github.com/qtile/qtile/pull/2162

.github/workflows/test_3-8.yml
DorianGray: I'm definitely in the "lets not slow down all development because one person made a mistake" boat.

ramnes: @tych0 I agree with you to some degree, but look how long it took for someone to find the problem last time. Not everyone has the time to investigate such obscure errors and yet could help on other subjects meanwhile.

tych0: It's not about mistakes. If there is clearly a bad commit, we can revert it and move on. The problem is when there *isn't* a clear bad commit, and people just re-run CI until it passes and merge. Everyone gets complacent, and CI becomes less and less reliable as more and more semi-reliable tests are introduced. I am not interested in going down this road, or any patches whose sole motivation is to make it easier to go down this road, especially if they introduce additional complexity.

tych0: > but look how long it took for someone to find the problem last time.

Yes, but merging more potentially broken code is not a solution. It is a path to madness.

ramnes:
I'm +0 and @tych0 is -1, so let's save everyone's time and close this for now. Thank you for the effort @DorianGray.

.github/workflows/test_3-8.yml
DorianGray: If your concern is test reliability you might take a look at https://github.com/qtile/qtile/pull/2107, although it's become quite big.

tych0: Yeah, I think there's maybe some rebase issues with the most important patch there; I haven't looked too closely at it though.

DorianGray: I don't see any...O.o

Issue #2068, opened 2021-01-04
CI hang is probably the result of our use of multiprocess' default fork behavior


DorianGray:
# Issue description

Sometimes CI will hang forever for no apparent reason. I believe this is because of how we use multiprocessing for tests.

We initialize logging at import time. Logging uses a lock internally to manage actually writing messages so they don't clobber eachother from different threads. In our tests we use multiprocessing to create a new process to run qtile in. The default behavior of multiprocessing is to fork and use the target function as the insertion point in the new process. Relevant documentation follows.

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
```
spawn
The parent process starts a fresh python interpreter process. The child process will only inherit those resources necessary to run the process object’s run() method. In particular, unnecessary file descriptors and handles from the parent process will not be inherited. Starting a process using this method is rather slow compared to using fork or forkserver.

Available on Unix and Windows. The default on Windows and macOS.

fork
The parent process uses os.fork() to fork the Python interpreter. The child process, when it begins, is effectively identical to the parent process. All resources of the parent are inherited by the child process. Note that safely forking a multithreaded process is problematic.

Available on Unix only. The default on Unix.

forkserver
When the program starts and selects the forkserver start method, a server process is started. From then on, whenever a new process is needed, the parent process connects to the server and requests that it fork a new process. The fork server process is single threaded so it is safe for it to use os.fork(). No unnecessary resources are inherited.

Available on Unix platforms which support passing file descriptors over Unix pipes.
```

The problem is, if we fork after logging has been initialized, it's possible that the lock will be copied while locked...and will never unlock itself since the thread that would have unlocked it is not running in the new forked process. Any call to log will hang forever at this point.

# Qtile version

all afaik

# Possible solutions:

1. We hoist logging init up out of import time and make sure we fork in tests before we init logging in the forks. We would need to stop sharing loggers between instances of objects and instead create a logger per instance.
2. We move away from multiprocessing altogether and instead use threads.
3. We stick with multiprocessing but we switch to the `spawn` start method. This would require us to rework how we do tests since our config objects are not picklable and spawn requires that any arguments to the new process or objects in the target functions scope are picklable. `spawn` uses fork followed by execve so the log lock is obliterated.

DorianGray:
I have seen this issue in the Real World™ on a uwsgi app in Another Lifetime™.

DorianGray:
[...]

PR #2107, opened 2021-01-16
Fix test hangs. Improve test resource management. Parallelize tests.

DorianGray:
This should fix our hanging tests as documented in https://github.com/qtile/qtile/issues/2068

While refactoring tests to allow `multiprocess.set_start_method("spawn")`, I realized why we had sporadic failures and were unable to run our tests in parallel. By changing some finalization procedures, we now no longer leak resources between tests.


ramnes:
Why switching to `multiprocess` if the only thing we need is `multiprocessing.set_start_method('spawn')`?

DorianGray:
that's not the only thing we need, we also need `dill` instead of `pickle`, which multiprocess provides.

DorianGray:
Anyways this is a draft, not even sure this approach is going to work until I work through some of the remaining pickling errors.

DorianGray:
We might be able to get by without dill, will test once I get things passing with(it's much more robust than pickle)

DorianGray:
Definitely cannot get away with not using multiprocess/dill...would require significant gymnastics to get it to pickle a bunch of things that should be simple...it's just a test dependency anyways :/

DorianGray:
https://github.com/qtile/qtile/pull/2119/checks?check_run_id=1724637109 is an example of the hangs that this PR fixes.

DorianGray:
So, this successfully parallelizes tests but I can't seem to get much above 2 workers. Things start timing out and causing problems.

DorianGray:
now depends on https://github.com/qtile/qtile/pull/2142

PR #2108, opened 2021-01-17
Tests: Replace tkinter windows with xcffib windows

m-col:
This extends test/scripts/window.py so that it will set the wm_type of
the window that it spawns, allowing for creation of e.g. dialogs, docks,
notifications. This makes the tkinter-based test/scripts/tkwindow.py
redundant, removing the tkinter dependency and simplifying the code ever
so slightly. Also solves a problem where github's pypy is built against
an old tkinter version, causing problems with pypy CI.


tych0:
Thanks, I've been meaning to do this for a while.

PR #2109, opened 2021-01-17
Bunch of enhancements for scratchpad

senthilbaboo:
Use wids instead of pids to save/restore states
We can use an arbitrary matcher instead of PIDs for scratchpad matching. This is useful, when the launch process, creates a sub process that creates a window (when scripting). Also helpful with chrome/firefox where the same process is re-used for multiple windows.
We can hide all scratchpads if needed in a group
We can set a scratchpad group to exclusive so that only one of the
scratchpad is displayed at a given time.
We have a WindowTranform config that can be used to place the scratchpad
relative to the screen, which provides more capabilities.
Don't hide scratchpads when saving states. Here the Xlib connection seems to be closed already and causing failures when saving states. Instead I am moving the scratchpad to the appropriate scratchpad group upon spawning.
_NET_WM_WINDOW_TYPE takes an ATOM

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -34,6 +34,7 @@ from libqtile import configurable, hook, utils
 from libqtile.bar import BarType
 from libqtile.command.base import CommandObject
 from libqtile.lazy import lazy
+from libqtile.window import WindowTransform
 
 
 class Key:
@@ -533,11 +534,15 @@ class ScratchPad(Group):
     label : string
         The display name of the ScratchPad group. Defaults to the empty string
         such that the group is hidden in ``GroupList`` widget.
+    single : Boolean
+        Only one of the window among the specified dropdowns will be
+        visible at a time.
     """
-    def __init__(self, name, dropdowns=None, position=sys.maxsize, label=''):
+    def __init__(self, name, dropdowns=None, position=sys.maxsize, label='', single=False):
         Group.__init__(self, name, layout='floating', layouts=['floating'],
                        init=False, position=position, label=label)
         self.dropdowns = dropdowns if dropdowns is not None else []
+        self.single = single
 
     def __repr__(self):
         return '<config.ScratchPad %r (%s)>' % (
@@ -571,7 +576,7 @@ class Match:
         rule)
     """
     def __init__(self, title=None, wm_class=None, role=None, wm_type=None,
-                 wm_instance_class=None, net_wm_pid=None):
+                 wm_instance_class=None, net_wm_pid=None, wid=None):
         self._rules = {}
 
         if title is not None:
@@ -584,6 +589,8 @@ class Match:
             self._rules["wm_type"] = wm_type
         if wm_instance_class is not None:
             self._rules["wm_instance_class"] = wm_instance_class
+        if wid is not None:
+            self._rules["wid"] = wid
         if net_wm_pid is not None:
             try:
                 self._rules["net_wm_pid"] = int(net_wm_pid)
@@ -594,7 +601,7 @@ class Match:
 
     @staticmethod
     def _get_property_predicate(name, value):
-        if name == 'net_wm_pid':
+        if name == 'net_wm_pid' or name == 'wid':
             return lambda other: other == value
         elif name == 'wm_class':
             def predicate(other):
@@ -620,6 +627,8 @@ class Match:
                 value = wm_class[0]
             elif property_name == 'role':
                 value = client.window.get_wm_window_role()
+            elif property_name == "wid":
+                value = client.window.wid
             else:
                 value = getattr(client.window, 'get_' + property_name)()
 
@@ -729,7 +738,7 @@ class DropDown(configurable.Configurable):
         ),
     )
 
-    def __init__(self, name, cmd, **config):
+    def __init__(self, name, cmd, match=None, transform=None, **config):
         """
         Initialize DropDown window wrapper.
         Define a command to spawn a process for the first time the DropDown
@@ -741,11 +750,24 @@ class DropDown(configurable.Configurable):
             The name of the DropDown configuration.
         cmd : string
             Command to spawn a process.
+        match : Match
+            A match object to identify the window instead of the pid.
+        transform: WindowTransform
+            Transform config for the window
         """
         configurable.Configurable.__init__(self, **config)
         self.name = name
         self.command = cmd
+        self.match = match
+        self.transform = transform
         self.add_defaults(self.defaults)
+        if self.transform is None:
+            self.transform = WindowTransform(
+                x=self.x,
+                y=self.y,
+                w=self.width,
+                h=self.height,
+            )
 
     def info(self):
         return dict(name=self.name,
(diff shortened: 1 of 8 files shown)


libqtile/config.py
```diff
@@ -533,11 +535,15 @@ class ScratchPad(Group):
     label : string
         The display name of the ScratchPad group. Defaults to the empty string
         such that the group is hidden in ``GroupList`` widget.
+    exclusive : Boolean
+        Only one of the window among the specified scratchpad should be
+        visible at a time.
     """
-    def __init__(self, name, dropdowns=None, position=sys.maxsize, label=''):
+    def __init__(self, name, dropdowns=None, position=sys.maxsize, label='', exclusive=False):
```
m-col, line 542: This is nice, but I think `exclusive` could be replaced with a more intuitive name. Perhaps `single`?

libqtile/scratchpad.py
```diff
@@ -294,12 +313,23 @@ def cmd_dropdown_toggle(self, name):
...
             if name in self._dropdownconfig:
                 self._spawn(self._dropdownconfig[name])
 
+    def cmd_hide_all(self):
+        """
+        Hide all scratchpads.
+        """
+        for _, d in self.dropdowns.items():
+            d.hide()
+
```
m-col, line 332: How about:
```
        for d in self.dropdowns.values():
            d.hide()
```

test/test_scratchpad.py
```diff
@@ -65,6 +70,68 @@ def is_killed(manager, name):
     raise ValueError('not yet killed')
 
 
+@scratchpad_config
+def test_sratchpad_with_matcher(manager):
+    # adjust command for current display
+    manager.c.group["SCRATCHPAD"].dropdown_reconfigure('dd-e', command='xterm -T dd-e -class dd-e -display %s sh' % manager.display)
```
m-col, line 76: Is manually specifying the display necessary? I think spawning commands in the tests should have the display present in their environment. If not then making that the case might make more sense than having to do this manul specification.

m-col:
There is a lot of nice stuff here. Couple things:

What changes are there (if any) in behaviour when using `wid`s rather than `pid`s to keep track of a scratchpad's windows? Was this change primarily to enable the use of a `Match`?

> We can use an arbitrary matcher instead of PIDs for scratchpad matching. This is useful, when the launch process, creates a sub process that creates a window (when scripting). Also helpful with chrome/firefox where the same process is re-used for multiple windows.

I'm just trying to understand this... please could you elaborate on what the current behaviour is and its issues and the new behaviour with these changes?

> We have a WindowTranform config that can be used to place the scratchpad
> relative to the screen, which provides more capabilities.

I think the idea is good but the implementation is a bit over-engineered, specifcally giving it a whole new class and argument to `DropDown`. I think it could be simpler by slightly extending the existing geometry arguments for `DropDown`. As they all take fractional values between 0 and 1 currently, they could also accept integers, and then place the window based on whether the arg is `int` or `float`. It may also make sense to consider negative values as offset from the other side of the screen, which is a common approach used by status bars, and this would provide the same behaviour you're trying to achieve using the pivot points. This implementation would likely move the logic out of `window.py`. I'm not sure I see the value of a `Window.cmd_transform`, largely because there are a few other `cmd_`s that do similar things and the behaviour can be achieved with a little bit of logic in your config.

All in all I think these are great ideas, so thanks!

senthilbaboo:
> There is a lot of nice stuff here. Couple things:
> 
> What changes are there (if any) in behaviour when using `wid`s rather than `pid`s to keep track of a scratchpad's windows? Was this change primarily to enable the use of a `Match`?

Yes this change is to primarily support general matches and make the states independent of PIDs.

> 
> > We can use an arbitrary matcher instead of PIDs for scratchpad matching. This is useful, when the launch process, creates a sub process that creates a window (when scripting). Also helpful with chrome/firefox where the same process is re-used for multiple windows.
> 
> I'm just trying to understand this... please could you elaborate on what the current behaviour is and its issues and the new behaviour with these changes?

There are some programs like chrome/firefox where there is just one main process and any subsequent windows created will share the same process id. For instance 'chrome --new-window' will create a new window but the PID of the new window won't be the same as PID returned by the command. This limits us from using scratchpads for such applications. Also, let's say I have a script that runs another application which creates the window, the PID of the script and the PID of the application that creates the window are different and cannot be used for matching.

> 
> > We have a WindowTranform config that can be used to place the scratchpad
> > relative to the screen, which provides more capabilities.
> 
> I think the idea is good but the implementation is a bit over-engineered, specifcally giving it a whole new class and argument to `DropDown`. I think it could be simpler by slightly extending the existing geometry arguments for `DropDown`. As they all take fractional values between 0 and 1 currently, they could also accept integers, and then place the window based on whether the arg is `int` or `float`. It may also make sense to consider negative values as offset from the other side of the screen, which is a common approach used by status bars, and this would provide the same behaviour you're trying to achieve using the pivot points. This implementation would likely move the logic out of `window.py`. I'm not sure I see the value of a `Window.cmd_transform`, largely because there are a few other `cmd_`s that do similar things and the behaviour can be achieved with a little bit of logic in your config.
> 
I don't think negative values can be used to achieve what pivot points can do. pivot points are mainly used for scaling. Let's say I want to resize the window to 1000 x 500 and place it at the center of the screen (being agnostic to screen size), with pivot points I can do (x=0.5, y=0.5, w=1000, h=500, px=0.5, py=0.5) and I don't think we can do this with negative numbers. Another example I could give is to resize the window from the bottom right edge. I can have something like (dw=10, dh=10, px=1.0, py=1.0). I don't think can be achieved by negative numbers. I primarily started fiddling with this to move my netflix/youtube/vlc windows to the bottom right corner (with some margin), but still have the ability to resize them, without changing their corners in the bottom right.

The reason, I introduced a WindowTransform config as opposed to adding a bunch of parameters to existing Scratchpad is just a design consideration, so that the WindowTransform object can be re-used just like Match.

> All in all I think these are great ideas, so thanks!

test/test_scratchpad.py
senthilbaboo: Without specifying the display, test fails. Other tests in the samefile too specifies display.

ramnes:
I'll leave the appreciation of this PR to other people since I have never used scratchpads. But I don't like adding a new core structure for one feature. I'd prefer a PR that adds WindowTransform and uses it where it can to replace current code and demonstrate its value, or not having this class at all.

m-col:
Could you split the changes into separate commits? Then it will be easier to see each change more easily. It would also be nice to merge the other changes while we discuss the window transform code.

m-col:
Do you hope to continue on this or would you be happy for someone else to continue where you left off?

jwijenbergh:
I don't completely understand every change in here currently. If I have the time I can pick this up but only focus on the WID matching part.

senthilbaboo:
I am so sorry for not responding here for a long time. I got busy with a few other things. I can put out another pull request without Window Transform part. Let me know, if you already have a pull request for WID part, in which case, I can put another for other features.

senthilbaboo:
FYI, I made [PR](#2801) for this.

PR #2110, opened 2021-01-17
CHANGELOG: add a note about ThreadedPollText removal

tych0:
34a3b2193498 ("Remove ThreadedPollText") broke my out of tree widgets, and
I'm probably not the only one who has them. Let's add a note about this in
the CHANGELOG.


tych0:
@DorianGray are these the timeouts you were talking about fixing?

tych0:
(CI timeouts, that is)

DorianGray:
no, I don't see a hang here. When the hangs occur, they timeout after ~3 hours

tych0:
Hrm. Seems like we may have some other bug in our CI then :(

DorianGray:
Yeah, I have seen that error before but I assumed it was related to the segfaults.

PR #2111, opened 2021-01-17
executable: make --log-level a "global" argument

### LINKED ISSUE #2099: `qtile --log-level=INFO` no longer works
My `.xsession` runs qtile using `qtile --log-level=INFO`.

This no longer works.

```
qtile: error: unrecognized arguments: --log-level=INFO
```

I'm guessing due to 908b910d00087ece13bb576f672c94bcf9e6fc43?

No big deal, but the changelog says

```
          Running `qtile` without arguments will continue to work for the
          forseeable future, but will be eventually deprecated.
```

tych0:
All commands use our logging framework, so it seems reasonable to want to
set the log level for all of them with the same code. Let's just move the
bits from `qtile start` into the main function.

This patch also allows case insensitive log levels, so we can do
--log-level=debug too.

Closes #2099

### DIFF
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -1,6 +1,8 @@
 import argparse
+import logging
 import sys
 
+from libqtile.log_utils import init_log
 from libqtile.scripts import cmd_obj, run_cmd, shell, start, top
 
 try:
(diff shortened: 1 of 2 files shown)


m-col:
Nice, just tested, works as expected. LGTM.

ramnes:
Cool!

mikelward:
Thanks!

PR #2113, opened 2021-01-17
Fix references to `tick` in ThreadLoopPool widgets

elParaguayo:
ThreadLoopPool does not have a tick method.

CheckUpdates widget called `tick` to update text after installing
updates so this functionality is currently broken.

Memory widget has redundant tick method which can be removed.


tych0:
Thanks!

PR #2119, opened 2021-01-18
Fix todos

DorianGray:
Grepped the codebase for TODO... found some I thought I could tackle!

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -148,6 +148,8 @@ class _Window(CommandObject):
         self.icons = {}
         window.set_attribute(eventmask=self._window_mask)
 
+        self._old_geometry = None
+
         self._float_info = {
             'x': None,
             'y': None,
@@ -458,20 +460,6 @@ class _Window(CommandObject):
             space around window as int or list of ints [N E S W]
         """
 
-        # TODO: self.x/y/height/width are updated BEFORE
-        # place is called, so there's no way to know if only
-        # the position is changed, so we are sending
-        # the ConfigureNotify every time place is called
-        #
-        # # if position change and size don't
-        # # send a configure notify. See ICCCM 4.2.3
-        # send_notify = False
-        # if (self.x != x or self.y != y) and \
-        #    (self.width == width and self.height == height):
-        #       send_notify = True
-        # #for now, we just:
-        send_notify = True
-
         # Adjust the placement to account for layout margins, if there are any.
         if margin is not None:
             if isinstance(margin, int):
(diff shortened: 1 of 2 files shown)


libqtile/core/manager.py
```diff
@@ -399,6 +399,7 @@ def update_net_desktops(self) -> None:
         # (via current_group), and when we set up the initial groups, there
         # aren't any screens yet. This can probably be changed when #475 is
         # fixed.
+        # #475 is fixed, this is still required.
```
m-col, line 402: Is this missing a word?

DorianGray: lemme rephrase

libqtile/window.py
```diff
@@ -195,8 +197,21 @@ def __init__(self, window, qtile):
...
+    def x(self, value):
+        _geometry_setter("x")(self, value)
+
+    @property
+    def y(self):
+        return _geometry_getter("y")(self)
+
+    @y.setter
+    def y(self, value):
+        _geometry_setter("y")(self, value)
```
m-col, line 214: You prefer it this way!?

DorianGray: ack, sorry, remnant of another attempt at solving this

m-col: Doesn't this way require resolving the {s,g}etter every time, where before it was resolved upon class definition?

DorianGray: yes

DorianGray: I'm going to add a commit that gives width and height the same treatment

libqtile/core/manager.py
```diff
@@ -399,6 +399,7 @@ def update_net_desktops(self) -> None:
         # (via current_group), and when we set up the initial groups, there
         # aren't any screens yet. This can probably be changed when #475 is
         # fixed.
+        # #475 is fixed, catching AttributeError is still necessary.
```
m-col, line 402: Ah I see. I misunderstood the first form lol. Thanks for updating it.

libqtile/window.py
```diff
@@ -199,22 +199,14 @@ def __init__(self, window, qtile):
...
-    def height(self, value):
-        _geometry_setter("height")(self, value)
+    width = property(
+        fset=_geometry_setter("width"),
+        fget=_geometry_getter("width"),
+    )
+    height = property(
+        fset=_geometry_setter("height"),
+        fget=_geometry_getter("height"),
+    )
```
m-col, line 209: I for one am glad to see this change too.

m-col (approved):
Nice, thanks for these, LGTM.

libqtile/core/manager.py
DorianGray: Thanks for helping me get it right!

libqtile/window.py
DorianGray: \o/

libqtile/core/manager.py
tych0: Any chance we can just delete the bits about #475 instead? I haven't evaluated whether this TODO can be fixed, but if the part about 475 is no longer relevant, let's just remove it.

ramnes:
I just reverted dc95d81a as it broke my ROXTerm (the window wouldn't be visible and show other parts of the screen instead), so it's probably not good for other things as well.

DorianGray:
I'll take a closer look at the icccm docs and figure out exactly what cases should get exactly what kind of configurenotify event. It must not have been as simple as just implementing 4.2.3

Barbarossa93:
I noticed that the same thing happens to termite, another vte based terminal emulator in https://github.com/qtile/qtile/commit/99c944e3f3b0e7062ed0d6e59aff95a06bbc7d29. Perhaps vte has specific window requests? Not sure if this helps, but  I thought it might help pinpoint where the issue is

PR #2120, opened 2021-01-19
Make ffibuild optional in tox

DorianGray:
On my machine, missing pulse audio makes the tests not run in tox despite pulse not being required. This makes ffibuild run and ignore if it fails

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,7 +27,7 @@ deps =
     xcffib >= 0.8.1
 commands =
     python3 setup.py install
-    {toxinidir}/scripts/ffibuild
+    - {toxinidir}/scripts/ffibuild
     python3 -m pytest -Wall --cov libqtile --cov-report term-missing {posargs}
 
 [testenv:packaging]


DorianGray:
I also introduced this problem when I ripped out pytest-runner :/

tych0:
I'm not so sure this is a good idea, what happens if it starts failing in CI?

ramnes:
I guess the tests would fail because they can't use the foreign functions, and we'll realize that there's a problem with the C interfaces. What's your concern?

tych0:
My concern is the "we'll realize..." step requires someone to remember that this script allows failure. Any time spent not realizing that is wasted.

The commit message is essentially "X fails on my machine, ignore it", but it describes neither why X fails or why it is safe to ignore. IMO ignoring failures should be a high bar, and it's not clear to me that this meets that.

DorianGray:
The errors from ffibuild are still logged so it's pretty obvious what happened in the event of failure.

That said...why was this merged? Am I missing some of the conversation?

DorianGray:
right, but this returns us to the same behavior as pytest-runner.
I don't think that tests should be gated behind optional dependencies that we don't have tests for anyways.
Would you say that I should make ffibuild accept arguments as to exactly what to build to fix this instead?

tych0:
>  I don't think that tests should be gated behind optional dependencies that we don't have tests for anyways.

They're not; this is at least a build test of the binding, which we don't get if we ignore failures.

>  Would you say that I should make ffibuild accept arguments as to exactly what to build to fix this instead?

Can you clarify how things are failing and why we would need this? IMO, requiring developers install libpulse-dev is fine.

DorianGray:
1. we're building things unnecessary to the tests
2. not everyone wants pulseaudio on their system, but they would certainly still like to just run `tox`
3. This isn't a big deal for me as I am running tests in an ubuntu VM now, but it does return us to the same behavior as before I removed pytest-runner. It was ignoring these failures for me before.

tych0:
> we're building things unnecessary to the tests

Well, but we're doing a build test of the thing itself right? That seems useful.

> not everyone wants pulseaudio on their system, but they would certainly still like to just run tox

IMO, the answer to this is "that's life".

> it does return us to the same behavior as before I removed pytest-runner. It was ignoring these failures for me before.

I prefer the behavior where we explicitly fail, so thanks for getting us there :)

DorianGray:
Great! Glad this is unnecessary.

PR #2121, opened 2021-01-19
Chord Enhancements

senthilbaboo:
Name a key_chord which are not modes, so that we can display the name in chord widget when key chord is active.
Block keyboard when a key chord is active as it is confusing to see keys forwarded to user application when a key chord is active.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -49,6 +49,7 @@ class Config:
         "extension_defaults",
         "bring_front_click",
         "wmname",
+        "block_keyboard_on_active_chord",
     ]
 
     def __init__(self, file_path=None, kore=None, **settings):
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -84,7 +84,7 @@ class Qtile(CommandObject):
         self.dgroups: Optional[DGroups] = None
 
         self.keys_map: Dict[Tuple[int, int], Key] = {}
-        self.current_chord = False
+        self.current_chord = None
         self.numlock_mask, self.valid_mask = self.core.masks
 
         self.current_screen: Optional[Screen] = None
@@ -327,12 +327,14 @@ class Qtile(CommandObject):
     def process_key_event(self, keysym: int, mask: int) -> None:
         key = self.keys_map.get((keysym, mask), None)
         if key is None:
-            logger.info("Ignoring unknown keysym: {keysym}, mask: {mask}".format(keysym=keysym, mask=mask))
+            logger.warning("Ignoring unknown keysym: {keysym}, mask: {mask}".format(keysym=keysym, mask=mask))
             return
 
         if isinstance(key, KeyChord):
             self.grab_chord(key)
         else:
+            if self.current_chord and (self.current_chord.mode == "" or key.key == "Escape"):
+                self.ungrab_chord()
             for cmd in key.commands:
                 if cmd.check(self):
                     status, val = self.server.call(
@@ -340,10 +342,6 @@ class Qtile(CommandObject):
                     )
                     if status in (interface.ERROR, interface.EXCEPTION):
                         logger.error("KB command error %s: %s" % (cmd.name, val))
-            else:
-                if self.current_chord is True or (self.current_chord and key.key == "Escape"):
-                    self.ungrab_chord()
-                return
 
     def grab_keys(self) -> None:
         """Re-grab all of the keys configured in the key map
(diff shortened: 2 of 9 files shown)


elParaguayo:
I haven't tested any of this and will leave to the devs to comment. However, one request from me: can you include updates to the docs as well please as this changes some of the behaviours described in the currents docs.

senthilbaboo:
Can someone take a look?

m-col:
Is the pass-through of the keys a feature or a bug? It sounds to me (and I don't use Chords (yet... really should)) like grabbing the keyboard when using a Chord is a no-brainer. So the question is, does it not make sense to just make that always the case?

As for the naming -- is that simply to give the Chord widget something to display? What did it display before, just nothing?

elParaguayo:
Chord widget only displayed a name when using a `mode` which is when you stay in the chord until pressing escape. I guess it's a bit like vim modes.

I assume this change means when you press the keys that starts the non-mode chord, the widget would now display the name of this chord if the user has given it a name. 


As for your first question, it feels like a bug to me as it could result in some unexpected behaviour where some keys are caught by the chord and some are passed to underlying applications. It feels better to catch everything until the chord mode has finished.

senthilbaboo:
Just to add to the above comment, the only reason I added that as an option is to make it backward compatible and I agree that is should be the default behavior. As grabbing keyboard is a bit risky, I don't want to make it default yet. If you think it's fine, I can remove the option to control it.

libqtile/confreader.py
```diff
@@ -49,6 +49,7 @@ class Config:
         "extension_defaults",
         "bring_front_click",
         "wmname",
+        "block_keyboard_on_active_chord",
```
tych0, line 52: Patch itself looks fine, but I wonder if we can make this a property of the particular key chord vs. a global one?

senthilbaboo: I think that this should be True by default globally. The current behavior of passing keys to user applications when a key chord is active seems weird.

We could argue for a use case where a user would still want to send keys to other applications while a chord is active. But that is a niche case and why to pollute the interface for that?

ramnes: > As grabbing keyboard is a bit risky, I don't want to make it default yet. If you think it's fine, I can remove the option to control it.

What risks are you thinking of?

ramnes:
> As grabbing keyboard is a bit risky, I don't want to make it default yet. If you think it's fine, I can remove the option to control it.

What risks are you thinking of?

senthilbaboo:
> What risks are you thinking of?

Recently I found a case where if the function that is mapped to a key is blocking, the keyboard is not released until the function is done. I fixed it by ungrabbing the keyboard before running the function. However this work only for key chords that are not modes. For modes, the keyboard won't be released until the user hits escape. This is fine for trivial functions such as moving, resizing windows, but if they had mapped a function that make take non trivial time to complete, the keyboard will not be responsive.

On top of that, I can't think of all the ways that something may go wrong. I just wanted to have a fail safe option so that one need not have to go into the source to turn off this feature.

Having said all that, I am in favor of removing this option and make it default to grab the keyboard.

libqtile/confreader.py
tych0: This is the way it works today, and we try not to break existing behaviors if possible. Let's just make it a configuration option on KeyChord itself.

m-col:
Perhaps we can hook into all of the async stuff to alleviate that problem?

libqtile/core/manager.py
```diff
@@ -327,23 +327,21 @@ def paint_screen(self, screen, image_path, mode=None):
...
         if key is None:
-            logger.info("Ignoring unknown keysym: {keysym}, mask: {mask}".format(keysym=keysym, mask=mask))
+            logger.warning("Ignoring unknown keysym: {keysym}, mask: {mask}".format(keysym=keysym, mask=mask))
             return
 
         if isinstance(key, KeyChord):
             self.grab_chord(key)
         else:
+            if self.current_chord and (self.current_chord.mode == "" or key.key == "Escape"):
+                self.ungrab_chord()
```
ep12, line 337: Question: What is the benefit of this being above the key handling? I think this would trigger `hook.fire("leave_chord")` before the actual key actions are triggered?

ep12:
+1 for the `name` option
+1 for making the keyboard grab optional:
> This is the way it works today, and we try not to break existing behaviors if possible. Let's just make it a configuration option on KeyChord itself.

Maybe this would be useful for [my config](https://www.github.com/ep12/qtileconf) in the `xournalpp` mode to still be able to use some of the default xournal keybindings.

PR #2123, opened 2021-01-19
Revert "Make ffibuild optional in tox"

tych0:
This reverts commit 7489b44d12af55d1a86ce219a0436e404002478c.

Per discussion in #2120, if we allow build failures it may make things
harder to track down in the future. Additionally, for things that don't
themselves have tests (e.g. the pulseaudio widget right now) this serves as
at least a build test of that binding.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,7 +27,7 @@ deps =
     xcffib >= 0.8.1
 commands =
     python3 setup.py install
-    - {toxinidir}/scripts/ffibuild
+    {toxinidir}/scripts/ffibuild
     python3 -m pytest -Wall --cov libqtile --cov-report term-missing {posargs}
 
 [testenv:packaging]


ramnes:
I don't mind much with or without it, I think there are good arguments both sides. My only problem is that we have a lot of tests that are optional already, that only get executed if a dependency is there. So failing if pulseaudio isn't available doesn't seem fair to me as it's not one of our hard dependencies. Can we keep this one optional but make the other FFI compilations mandatory?

tych0:
> Can we keep this one optional but make the other FFI compilations mandatory?

I'd prefer not to, and I'd prefer to make the other test deps mandatory. More test coverage is better, IMO

ramnes:
I'm good with everything mandatory if we provide some kind of isolation à la Docker to run the tests. I don't want all these things to be on my system just to run the tests (especially pulseaudio), and most importantly, I don't want the other contributors to stop trying to run the tests on their machine because they see how many dependencies they have and would rather rely on the CI (which already happens to some degree, we've been asked a Docker image a few times already).

On a side note, a Docker image would also benefit to the CI so that we don't have to reinstall everything at each run. And help for reproducibility. But that's more work than simply making pulseaudio optional.

tych0:
If we don't run (build the pulse bindings, etc.) in CI, they will eventually break. Feel free to propose patches to make them easier to run locally however you want, but IMO we should test everything we can in CI.

PR #2127, opened 2021-01-20
Implement Drawer modes

DorianGray:
This allows us to more finely control how widgets draw themselves to the
screen. We can either TEST by not sending anything to X at all, draw
DIRECTly to the window(good for graphs and other animations whose contents
change between every draw() call), or draw the old way with PM_COPY(good
for things that do not change their contents much).

Please let me know if you think an approach that uses different Drawer subclasses would be better...but imo there's so much shared code and such small differences it might be messier.

### DIFF
--- a/libqtile/drawer.py
+++ b/libqtile/drawer.py
@@ -30,12 +30,22 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import collections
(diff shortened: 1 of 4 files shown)


DorianGray:
In my testing, this resulted in a 40% cpu usage reduction compared to master when using a `CPUGraph` with `frequency` set to `0.001`

DorianGray:
Cancelling this since I'm doing this a different much cleaner way while abstracting drawer away from X

PR #2131, opened 2021-01-21
Guess terminal to use as default app when using scripts/xephyr

SirWrexes:


scripts/guess_terminal
```diff
@@ -0,0 +1,30 @@
...
+mate-terminal
+kitty
+yakuake
+tilda
+guake
+eterm
+st
+urxvt
+xterm
+x-terminal-emulator"
```
tych0, line 22: Can we make this a python script that uses our existing guess_terminal() instead of copy/pasting the list?

m-col:
Nice, great idea.

m-col:
You could also just stick this where you have `$(./guess_terminal)`:
```
$(python -c "from libqtile.utils import guess_terminal; print(guess_terminal()))"
```

SirWrexes:
Huh. I guess I should have thought about that.
This might as well just be in the script rather than a separate file since it fits in one line. 🤔

m-col:
Please could you squash the commits down into one? Otherwise looks good.

SirWrexes:
Oh I'm sorry I forgot to remove the guess_terminal script. Hold on.

m-col:
Awesome, thanks for this!

Issue #2130, opened 2021-01-21
Broken Xephyr script (qtile unrecognised arguments)


SirWrexes:
# Issue description

When trying to launch `qtile` in a `xephyr` container using the provided script, I get this error :
![image](https://user-images.githubusercontent.com/5567164/105354341-b6acf900-5bf0-11eb-9bae-2574423b9883.png)
Changing the log level doesn't solve the issue.
However, replacing `"${HERE}"/../bin/qtile` in the script with `$(which qtile)` to invoke the one installed via `pacman`, this happens :
![image](https://user-images.githubusercontent.com/5567164/105355177-fde7b980-5bf1-11eb-960a-e78b01b65ca6.png)
Removing `start` from the line where `qtile` is called fixes this, and the container launches, effectively spawning a new X server with `qtile` as the window manager.

Note : I see the default terminal called is `xterm`, which I did not install (I only have `kitty`).
I made a simple bash script inspired by your utils.guess_terminal() to fix this issue and will propose a pull request to implement it in the `xephyr` script soon.

# Qtile version
Repo : dc95d81a39b64be9d452c88df1e2849cb78fc85b
Arch package : 0.16.1-2

elParaguayo:
Sorry if I'm not completely understanding your issue here.

However, regarding using `start` or not: this is a change that was added after 0.16.1 (the one in the Arch repos). For 0.16.1 you launch via `qtile` but, if you're using git version, you need `qtile start`.

elParaguayo:
Ah, sorry, I think I'm with you now.

Looking at the error, it looks like `qtile` expects `start` after the log level argument but the script file places it before (https://github.com/qtile/qtile/blob/master/scripts/xephyr#L16).

If that's right, can you try just moving start to before the `&` in the xephyr script and see if it works.

Barbarossa93:
Changing it to this:
```
  env DISPLAY=${XDISPLAY} QTILE_XEPHYR=1 ${PYTHON} "${HERE}"/../bin/qtile -l ${LOG_LEVEL} start $@ &
```
worked for me

elParaguayo:
OK - I can submit a pull request for this later if no one else does.

Barbarossa93:
No worries, I got it @elParaguayo  :)

elParaguayo:
Thanks - can you test passing other arguments that would be expanded via the `$@` part? My suspicion is the `start` argument is the only positional one and needs to come last so could break if `$@` is expanded into some other arguments.

[...]

PR #2122, opened 2021-01-19
Inhibit tiled windows from hiding floating windows

senthilbaboo:
When clicked on windows that are not floating, don't bring them to front.

PR #2133, opened 2021-01-21
Fixed xephyr script

Barbarossa93:
Update xephyr script to run without `qtile: error: unrecognized arguments: -l INFO` error. Should address https://github.com/qtile/qtile/issues/2130

### DIFF
--- a/scripts/xephyr
+++ b/scripts/xephyr
@@ -13,7 +13,7 @@ Xephyr +extension RANDR -screen ${SCREEN_SIZE} ${XDISPLAY} -ac &
 XEPHYR_PID=$!
 (
   sleep 1
-  env DISPLAY=${XDISPLAY} QTILE_XEPHYR=1 ${PYTHON} "${HERE}"/../bin/qtile start -l ${LOG_LEVEL} $@ &
+  env DISPLAY=${XDISPLAY} QTILE_XEPHYR=1 ${PYTHON} "${HERE}"/../bin/qtile -l ${LOG_LEVEL} start $@ &
   QTILE_PID=$!
   env DISPLAY=${XDISPLAY} ${APP} &
   wait $QTILE_PID


tych0:
Seems like 00ae5bd5bc0f5fc139a68bd308843c79366c23a8 broke this, lgtm.

m-col:
Shouldn't the argument work there too?

Barbarossa93:
@m-col Not according to the commit @tych0 linked. Looks like that commit deliberately moved the `-l` flag to be in the main function from the start function.

Should I be worried that some checks were not successful? Literally just changed an itty bit of shell script, not sure why the github checks would fail lol

tych0:
Looks like a racy test from #2122

m-col:
What I meant was it would be nice if `-l $LOGLEVEL` can be put anywhere, but it doesn't matter so much. Just figure that if it causes this problem then it could well cause issues if people have that argument entered into their normal qtile script come next release time.

ramnes:
Yeah, it seems really counter intuitive to put optional arguments in front of positional arguments, so I guess we'll get complaints if we leave it that way. Ideally we'd support both, but I don't think there's a way to do that without making the code redundant with argparse... What would you think of reverting 00ae5bd and adding the `--log-level` option to all sub-commands, maybe with a base parser or something?

ramnes:
Well, actually we have to have both if we still want to address #2099.

tych0:
I think this should either be fixed in argparse or not at all. We're definitely not the only ones with this problem in a subcommand structure, so people are mostly used to it by now.

ramnes:
argparse does implement base ("parent") parsers, so have a look at #2143 for an implementation example that seems more appropriate to me for what we're trying to achieve with the log level.

tych0:
I'm going to close this in favor of #2143

PR #2139, opened 2021-01-21
Warp to screen

senthilbaboo:
When switching to a screen, and when there are no windows, warp to
the center of the screen.


tych0:
Thanks!

PR #2142, opened 2021-01-22
Fix a flaky test

senthilbaboo:
Make sure the windows are created in the intended order by calling sync after creating every window.


tych0:
The mypy failure was fixed, so I think this is good to land.

tych0:
Thanks!

PR #2143, opened 2021-01-22
Give a common parent parser to the subparsers (and main parser)

ramnes:
By doing so, we can give the --logging flag to any sub-command in a more intuitive way, i.e. after the sub-command and not before it (see the discussion in #2133), while keeping `qtile -l DEBUG` functional (see #2099) as a backward compatibility hack.

This approach has a downside though: `qtile -l DEBUG start` will be a valid command but won't act as expected (the log level will stay the default one, INFO). But this command will get invalid as soon as we remove that backward compatibility hack.

Note: we can probably make `qtile -l DEBUG start` invalid with yet another hack but I thought it was acceptable as-is since the backward compatibility with no sub-command is not meant to stay. Feel free to tell me if you think the opposite.

### DIFF
--- a/libqtile/scripts/top.py
+++ b/libqtile/scripts/top.py
@@ -175,9 +175,10 @@ def top(opts):
         raw_stats(c, limit=lines, force_start=force_start)
 
 
-def add_subcommand(subparsers):
-    parser = subparsers.add_parser("top", help="resource usage information")
-    parser.add_argument('-l', '--lines', type=int, dest="lines", default=10,
+def add_subcommand(subparsers, parents):
+    parser = subparsers.add_parser("top", parents=parents,
+                                   help="resource usage information")
+    parser.add_argument('-L', '--lines', type=int, dest="lines", default=10,
                         help='Number of lines.')
     parser.add_argument('-r', '--raw', dest="raw", action="store_true",
                         default=False, help='Output raw without curses')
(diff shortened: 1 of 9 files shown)


libqtile/scripts/top.py
```diff
@@ -175,9 +175,10 @@ def top(opts):
         raw_stats(c, limit=lines, force_start=force_start)
 
 
-def add_subcommand(subparsers):
-    parser = subparsers.add_parser("top", help="resource usage information")
-    parser.add_argument('-l', '--lines', type=int, dest="lines", default=10,
+def add_subcommand(subparsers, parents):
+    parser = subparsers.add_parser("top", parents=parents,
+                                   help="resource usage information")
+    parser.add_argument('-L', '--lines', type=int, dest="lines", default=10,
```
ramnes, line 181: I changed from -l to -L here so that it doesn't conflict with the logging.

tych0:
I think this would be nice to have. I wonder if this means we shouldn't just drop the backwards compat for `qtile start`. Generally, I worry about subtly breaking people with compat stuff, but not even starting seems like a fairly obvious breakage :).

m-col:
> I wonder if this means we shouldn't just drop the backwards compat for qtile start

With the new release out, now is as good a time as any to drop backwards compatibility!

ramnes:
> With the new release out, now is as good a time as any to drop backwards compatibility!

Yeah right, done. :)

PR #2144, opened 2021-01-23
Experiment bring to front

senthilbaboo:
This is not for review, but only to show, how I think we can implement the window stack handling. I have not even tested it completely, but this will give a basic idea, i hope.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -924,6 +924,10 @@ class Connection:
                 xcffib.xproto.Time.CurrentTime,
             )
 
+    def query_win_stack(self):
+        q = self.conn.core.QueryTree(self.default_screen.root.wid).reply()
+        return list(q.children)
+
     @functools.lru_cache()
     def color_pixel(self, name):
         pixel = self.screens[0].default_colormap.alloc_color(name).pixel
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -410,7 +410,9 @@ class Qtile(CommandObject):
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
         """
+        stack = self.conn.query_win_stack()
         windows = [wid for wid, c in self.windows_map.items() if c.group]
+        windows.sort(key=lambda wid: stack.index(wid))
         self.core.update_client_list(windows)
 
     def add_group(self, name, layout=None, layouts=None, label=None):
@@ -700,13 +702,6 @@ class Qtile(CommandObject):
             wid = e.child
             window = self.windows_map.get(wid)
 
-            if self.config.bring_front_click and (
-                self.config.bring_front_click != "floating_only" or getattr(window, "floating", False)
-            ):
-                self.conn.conn.core.ConfigureWindow(
-                    wid, xcffib.xproto.ConfigWindow.StackMode, [xcffib.xproto.StackMode.Above]
-                )
-
             try:
                 if window.group.screen is not self.current_screen:
                     self.focus_screen(window.group.screen.index, warp=False)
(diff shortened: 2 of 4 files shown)


tcld:
Thank you very much for the example. I'm not sure I'll find the time this weekend to rework #2029, next week is more likely. One detail I would like to improve upon is that you are just checking windows within the current group - ideally that would work by checking a screen region instead (i.e. usually a full screen, but we do have the concept of fake screens, too).

On a sidenote: Are you just insterested in the topic or do you actually want to implement it yourself?

senthilbaboo:
> Thank you very much for the example. I'm not sure I'll find the time this weekend to rework #2029, next week is more likely. One detail I would like to improve upon is that you are just checking windows within the current group - ideally that would work by checking a screen region instead (i.e. usually a full screen, but we do have the concept of fake screens, too).

Looking into screen region for overlaps seems overkill to me and adds unnecessary complexity. As far as the fake screens are concerned, I believe we still would have only one active group. So I don't think, we need any special handling for that.

Things we may have to add is actually handling EWMH hints like above, below, dock, panel, sticky etc. and probably handler codes to check when they change. 

Hence functionality like pinning or sticking windows is just about changing those hints. 

> 
> On a sidenote: Are you just insterested in the topic or do you actually want to implement it yourself?

We can always share work here. There will be a lot of things to do here. And in this example, I could probably push some basic implementations and then we can start gluing things together. 

I just want to understand the general consensus and the opinions on the approach here before I spend more time.

I also have a lot of other ideas but we have to get this stacking right for most of those other things to work, like popup windows, semi transparent sticky windows, built in key logger etc.

tcld:
> Looking into screen region for overlaps seems overkill to me and adds unnecessary complexity. As far as the fake screens are concerned, I believe we still would have only one active group. So I don't think, we need any special handling for that.

No, every (fake-)screen has it's own group. On top of that floating windows can technically exist outside of their groups screen. The latter case can probably ignored, but the former I consider important.

> We can always share work here. There will be a lot of things to do here.

I think there should be a basic implementation that can be merged in a reasonable amount of time, with further refinements being left for future PRs. I could create a draft-PR in the next few days and, hopefully, you can take a look, maybe add your own commits to it - how das that sound?

> Hence functionality like pinning or sticking windows is just about changing those hints.

I really hope so, that would make it a lot easier to handle.

> I just want to understand the general consensus and the opinions on the approach here before I spend more time.

I believe the consensus will be "less code is better" and "leveraging X is better".^^ About those further ideas you might have, there is probably a chance for a more balanced discussion. :)

senthilbaboo:
> > Looking into screen region for overlaps seems overkill to me and adds unnecessary complexity. As far as the fake screens are concerned, I believe we still would have only one active group. So I don't think, we need any special handling for that.
> 
> No, every (fake-)screen has it's own group. On top of that floating windows can technically exist outside of their groups screen. The latter case can probably ignored, but the former I consider important.
> 
In this case, instead of just looking into windows or current screen or group, we could just look into all the windows that are currently visible on any of the screen. I think that should be fairly easy to implement as well. 

> > We can always share work here. There will be a lot of things to do here.
> 
> I think there should be a basic implementation that can be merged in a reasonable amount of time, with further refinements being left for future PRs. I could create a draft-PR in the next few days and, hopefully, you can take a look, maybe add your own commits to it - how das that sound?
> 
Sounds good. 
> > Hence functionality like pinning or sticking windows is just about changing those hints.
> 
> I really hope so, that would make it a lot easier to handle.
> 
> > I just want to understand the general consensus and the opinions on the approach here before I spend more time.
> 
> I believe the consensus will be "less code is better" and "leveraging X is better".^^ About those further ideas you might have, there is probably a chance for a more balanced discussion. :)
👍

PR #2145, opened 2021-01-23
Fix mypy check error on widgetbox

elParaguayo:
I've noticed a few PRs recently failing with a mypy error due to widgetbox.py.

Not sure what's changed in mypy as I don't get this error on my system but I think this fixes it.


ramnes:
Thanks!

PR #2156, opened 2021-01-27
Fix the names of the XMonad layouts (closes #2152)

### LINKED ISSUE #2152: Minor errors in documentation for MonadTall and MonadWide
# Issue description
The MonadTall and MonadWide documentation has some minor errors:
- For MonadTall the default name is documented as 'xmonadtall' when it actually is called just 'monadtall' (at least in my version of Qtile)
- For MonadWide the default name is also documented as 'xmonadtall' when it actually is called 'monadwide' (at least in my version of Qtile)

[http://docs.qtile.org/en/latest/manual/ref/layouts.html#monadtall](http://docs.qtile.org/en/latest/manual/ref/layouts.html#monadtall)
[http://docs.qtile.org/en/latest/manual/ref/layouts.html#monadwide](http://docs.qtile.org/en/latest/manual/ref/layouts.html#monadwide)

# Qtile version
0.16.1

# Stack traces
N/A

# Configuration
Not relevant

ramnes:


libqtile/layout/xmonad.py
```diff
@@ -150,7 +150,7 @@ class MonadTall(_SimpleLayoutBase):
         ("border_width", 2, "Border width."),
         ("single_border_width", None, "Border width for single window"),
         ("single_margin", None, "Margin size for single window"),
-        ("name", "xmonadtall", "Name of this layout."),
+        ("name", "monadtall", "Name of this layout."),
```
tych0, line 153: Can we just delete this line since it's autogenerated in the base Layout class?

libqtile/layout/xmonad.py
```diff
@@ -846,6 +846,10 @@ class MonadWide(MonadTall):
     _up = 0
     _down = 1
 
+    def __init__(self, **config):
+        config.setdefault("name", "monadwide")
```
tych0, line 850: Same here, I wonder if we can't just drop this?

tych0:
Ping?

ramnes:
This attribute is used by the currentlayout widget and other things, so we can remove it but it's more work as we'd have to remove it for all layouts and modify where it's used. It could also break configs if anyone use that to rename their layout for some reason (it's a configurable attribute). I'm definitely open to this but frankly this was just a quick fix to have one less issue open and I don't plan to put more energy on this matter. :)

tych0:
But doesn't just deleting the code fix it? It should still be configurable, since it's generated in the base layout class...

ramnes:
Nah it shows None as a default value instead, and would probably make the currentlayout widget crash or show None as well.

tych0:
But other layouts don't hard code their names, right? There must be something else going on here...

ramnes:
Yes they do :)

tych0:
So they do. Seems like we can either drop that or the code in __init__ that sets them. What do you think of https://github.com/qtile/qtile/pull/2365 ?

ramnes:
Yeah just found that as well. Sounds good!

PR #2158, opened 2021-01-28
Update volume widget doc

### LINKED ISSUE #2157: Missing requirement in documentation for Volume widget
# Issue description
The Volume widget by default needs access to amixer, which is part of the package alsa-utils, unless the `get_volume_command` is manually set to something else by the user.

This is not mentioned in the documentation [http://docs.qtile.org/en/latest/manual/ref/widgets.html#volume](http://docs.qtile.org/en/latest/manual/ref/widgets.html#volume)

# Qtile version
0.16.1

# Stack traces
N/A

# Configuration
Irrelevant

elParaguayo:
Include reference to default dependency on `amixer` for volume
commands.

Fixes #2157

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -46,6 +46,10 @@ re_vol = re.compile(r'\[(\d?\d?\d?)%\]')
 class Volume(base._TextBox):
     """Widget that display and change volume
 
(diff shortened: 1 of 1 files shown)


ramnes:
Thanks!

PR #2159, opened 2021-01-28
Smarter floating

ramnes:
[As noted by @pmkap in #1518](https://github.com/qtile/qtile/issues/1518#issuecomment-619381672), i3 floating defaults are quite good and the main difference with ours is a check of the size hints to see if the window has a fixed size. This is a great way to match windows that are complete desktop applications but still meant to float, like gnome-screenshot, so let's do it too. :)

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -28,9 +28,9 @@
 import os.path
 import sys
 import warnings
-from typing import List, Optional
+from typing import Callable, List, Optional
 
-from libqtile import configurable, hook, utils
+from libqtile import configurable, hook, utils, window
 from libqtile.bar import BarType
 from libqtile.command.base import CommandObject
 from libqtile.lazy import lazy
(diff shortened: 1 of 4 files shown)


libqtile/config.py
```diff
@@ -569,9 +569,12 @@ class Match:
     net_wm_pid:
         matches against the _NET_WM_PID atom (only int allowed for this
         rule)
+    func:
+        matches against the given function, which receives the client as only
+        argument
```
tych0, line 574: Patch itself seems fine, but can we give this an explicit doc about what args the function takes and what it should return? And maybe a `Callable[[window.Window], bool]` annotation.

ramnes: > Patch itself seems fine, but can we give this an explicit doc about what args the function takes and what it should return?

Here or somewhere else?

ramnes: > And maybe a `Callable[[window.Window], bool]` annotation.

Added

tych0: > Here or somewhere else?

Here, the current docstring for the function is mostly useless, IMO :). Before the type annotation, I had to read the source to figure out what it got as an argument and what it should return.

ramnes: done

ramnes: done

PR #2160, opened 2021-01-28
Respect max size

ramnes:


tych0:
A test for this would be nice :)

ramnes:
Yeah, and for #2159 as well. I tried to modify `tests/scripts/window.py` to spawn a window with such hints but couldn't find how to do that with xcffib. Can you help? Or I can add gnome-screenshot as a dependency for both tests. :)

tych0:
See stuff like test_only_one_focus(), it sets WM_HINTS like you want here.

ramnes:
test added :)

PR #2161, opened 2021-01-29
Fix WM_NORMAL_HINTS unpacking

ramnes:
https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#wm_normal_hints_property

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -38,7 +38,7 @@ import functools
 import operator
 import typing
 from collections import OrderedDict
-from itertools import chain, repeat
+from itertools import chain, islice, repeat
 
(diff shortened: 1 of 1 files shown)


tych0:
Can you describe what was incorrect about the previous approach in the commit message?

ramnes:
*  Only atom 11 is assigned to `min_aspect` while it should be a `x, y` tuple.
* `max_aspect` gets atom 12 while it should get 13 and 14 for the same reason.
* `base_width`, `base_height` and `win_gravity` all get the same atom, which simply doesn't make sense. Each has its own value, and they have to default to other values when empty, as mentioned in the specification.

ramnes:
(and added to the commit message)

PR #2170, opened 2021-01-31
Some big changes to the backend

m-col:
I'm starting work on moving window.Window into the x11 backend and merging it with xcbq.Window, while forming a Window abstract base class for use in other backends.

I'm going to try to drip feed the commits in otherwise it may get a bit whelming. Here are a couple.

The second one, which removes the `_add_net_wm_state` decoration from xcbq.Window seems to be unused in-tree so I've removed it, but let me know if there is a good reason to keep it.


m-col:
Just added another commit, moving the window.py contents into the x11 backend. The commit message:

    This commit is pretty much a cut-and-paste of the contents of window.py
    into xcbq.py of the x11 backend, without doing any kind of merging of
    the two Window classes. All libqtile.window imports are updated to point
    to xcbq for now, though non-x11-specific code that is now in xcbq will
    be moved into the base backend. It also adds base backend classes for
    Window, Internal and Static windows, which the existing ones now
    subclass and are accessed as attributes of the respective backend's
    Core.

m-col:
Rationale for merging window.Window into the x11 backend:

- unclear boundaries in the scope of window.Window and xcbq.Window
- many/most of the code in window.Window is X specific and operates on the window's xcbq.Window instance
- This will lead to creation of a base Window class in backend/base/window.py that will have all of the appropriate methods and `cmd_` commands, which will need backend-specific implementations (e.g. `map`)

m-col:
The core is now responsible for providing `Window`, `Internal` and `Static` to the `Qtile` manager. The types don't match up yet between the base Core and the x11 Core so mpyp complains. I've added some ignore comments to those lines temporarily, which will be removed when I merge the two Window classes so that the x11 Window is in fact a child class of base.Window.

tych0:
We'll still need a wrapper class for window at the qtile level right? Or are you thinking about getting rid of it entirely?

tych0:
I guess I'd rather move everything in to xcbq.Window, versus having two different classes with the same name. It feels like if we're going to refactor things in preparation for wayland, we should try and re-factor them into a more reasonable position, not just a different one.

m-col:
> It feels like if we're going to refactor things in preparation for wayland, we should try and re-factor them into a more reasonable position, not just a different one.

For sure. The commits here are only the first of a few, where the next ones will be piece-by-piece merging the two Window classes together. I wanted to submit them a few commits at a time so to get feedback as I go, in case any disgreements about the approach mean that I would have wasted a lot of time in a full refactor. The last commit (actually moving window.Window) in particular is the one that can cause headaches will rebasing on top of other changes so I was eager to get that in while I had the chance, then any following commits will have much smaller diffs (and any other PRs that modify Window will be contained to that file).

If you like I can add a few more commits soon so you can see the next few steps and the merge taking shape (hopefully nobody modifies window.py in the meantime...)

ramnes:
> * This will lead to creation of a base Window class in backend/base/window.py that will have all of the appropriate methods and `cmd_` commands, which will need backend-specific implementations (e.g. `map`)

I would have started with that to avoid messing up the git history too much. Why going the other way around? The project sounds good to me otherwise. :)

m-col:
I tried that initially but there are loads of things that will need to go on the
base class, not just the cmd_s but lots of properties, for example. I paused for
a couple weeks and I got out of date so it was easier to start again, and found
this way was easier as I can go through each property/method on each Window
class and merge them, adding it to the base class if necessary as I go :)

m-col:
I've just pushed another couple commits, the last one being the actual merge of the two classes. I tried to keep the changes minimal i.e. simply moving things around and updating references in other files (think `win.window.wid` -> `win.wid`). A couple of changes that were required to allow this:

- `qtile` has to be passed to windows after their creation, so they've been given their own `_configure` method.
- `Connection.create_window` has been converted to `Connection.create_wid` where it does what it did before but without instantiating a window class. This has then been used for example by `Bar` to create its `Internal` window. This also means there were some updates to tests that use this mechanism to make windows; hopefully those changes are clear.

All tests are passing on my local machine, and I am running it live ATM and so far so good. Unfortunately today's commits have made rebasing non-trivial so I'll have to find some more time at some point soon to deal with that, but in the meantime it would be great if people can test it out live because these changes seem quite big. I'll let you know if there I have any problem while running it.

m-col:
I've rebased on top of master. I can also confirm that nothing unexpected has manifested while running this for the past two days. Anyone happy to review?

tych0:
I would still really like to avoid https://github.com/qtile/qtile/pull/2170/commits/f0bd08e877468d227fcf4b5596346c861b2d327c if possible. It leaves the tree in a messy state while not really achieving much towards abstracting things. Plus, you're already going through method by method anyway, so I don't see much value in a wholesale move.

m-col:
Can you clarify what you mean "It leaves the tree in a messy state"? Just so I know what to avoid next time.

tych0:
It leaves things in a state where we have two different classes that effectively represent the same thing, xcbq.XWindow and xcbq.Window. People looking at the tree will wonder "why are these things different?". There were already two different objects which were kind of the same, and having them in the same file doesn't really fix that.

But the other part is just it being a huge commit that doesn't really do anything. It does, however, break rebases for everyone with patches in flight (as you've noticed), and cause issues with navigating git history (although I think this is mostly fixed with newer gits). We've had problems with big no-op commits like this in the past ("fix formatting to be pep8 compliant", etc.), so I'd like to avoid them.

m-col:
OK, sounds good, thanks for the clarification!

libqtile/backend/x11/core.py
```diff
@@ -103,7 +107,7 @@ def __init__(self, display_name: str = None) -> None:
         )
 
         self._wmname = "qtile"
-        self._supporting_wm_check_window = self.conn.create_window(-1, -1, 1, 1)
+        self._supporting_wm_check_window = xcbq.Window(self.conn, self.conn.create_wid(-1, -1, 1, 1))
```
DorianGray, line 110: Should these be `self.Window`? Or am I missing the point of the class level class properties added to core?

DorianGray (changes requested):
Overall I really like these changes, especially the fact that I never have to type `window.window.wid` ever again.

I would strongly prefer that the various Window related classes be in separate files instead of all being in `backend/x11/xcbq.py`. In the current state things are very difficult to navigate. Of course the base classes are good in `backend/base/window.py`. It also might be better to move the core abstract bases into `backend/core.py` and `backend/window.py` instead of having a base module, since the backend module IS the base module. Each of the subsequent cores could just replicate that organization.

For example,
`libqtile/backend/core.py::Core`
`libqtile/backend/window.py::Window`
`libqtile/backend/window.py::Static`
`libqtile/backend/window.py::Internal`

then

`libqtile/backend/x11/core.py:(X)Core`
`libqtile/backend/x11/window.py::(X)Window`
etc

I don't have strong feelings about the class name being `Core` or `XCore` in the individual backend since we can get around that name collision easily by using module imports.

libqtile/backend/x11/core.py
m-col: The idea is that the core needs to provide the Window class defined by that backend to the rest of libqtile in a general way. Class attributes seem like a convenient way to do that. A direct import e.g. using `xcbq.Window` is specific to the x11 backend so cannot be used outside of the x11 backend to get the Window. In the case of the core's definition itself (like this line) it is equivalent though. However, I think the final form of the backends might not need to provide a Window instance at all, but time will tell.

m-col:
> I would strongly prefer that the various Window related classes be in separate files instead of all being in `backend/x11/xcbq.py`

Yep I agree, that is my eventual intention.

m-col:
I'm closing this in favour of submitting smaller chunks at a time.

Issue #2173, opened 2021-01-31
KeyboardLayout widget doesn't set setxkbmap option.


MatejaMaric:
To recreate this issue simply include KeyboardLayout widget in your Bar and specify `option` variable. I found it in the most recent version of Qtile.

For example:
```
    widget.KeyboardLayout(option='caps:swapescape'),
```

If you run this example you will realize that Escape and Caps Lock are not swapped.

I managed to solve this issue myself, so I will link it in my pull request.

PR #2174, opened 2021-01-31
Fix issue #2173

MatejaMaric:
Fixes: #2173 
When I was looking through the code I realized that setxkbmap options
and even keyboard layouts themselves are not set until the first run of `next_keyboard` function which calls keyboard setter.

I solved this by simply setting the keyboard to the first one from the list of predefined `configured_keyboards`, after the initialization of widget.

I also added one newline in widget documentation in hope to format code example as code block.

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -39,8 +39,8 @@ class KeyboardLayout(base.InLoopPollText):
     and bind function "next_keyboard" to specific keys in order to change layouts.
 
     For example:
-        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
 
+        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
 
     It requires setxkbmap to be available in the system.
     """
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -60,6 +60,8 @@ class KeyboardLayout(base.InLoopPollText):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(KeyboardLayout.defaults)
 
+        self.keyboard = self.configured_keyboards[0]
+
         self.add_callbacks({'Button1': self.next_keyboard})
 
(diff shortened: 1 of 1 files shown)


libqtile/widget/keyboardlayout.py
```diff
@@ -39,8 +39,8 @@ class KeyboardLayout(base.InLoopPollText):
     and bind function "next_keyboard" to specific keys in order to change layouts.
 
     For example:
-        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
 
+        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
```
tych0, line 43: Seems like you can drop this from the diff?

tych0:
Your analysis in the PR message is great; can you put that in the commit message?

libqtile/widget/keyboardlayout.py
MatejaMaric: That diff actually puts a new line between:
```
For example:
```
and
```python
Key([mod], “space”, lazy.widget[“keyboardlayout”].next_keyboard(), desc=”Next keyboard layout.”),
```

---

If you go to the [widget documentation](http://docs.qtile.org/en/latest/manual/ref/widgets.html#keyboardlayout) you will see that
```python
Key([mod], “space”, lazy.widget[“keyboardlayout”].next_keyboard(), desc=”Next keyboard layout.”),
```
is not treated as a code block. That is why I did this.
Diff looks kind of weird because I actually borrowed a new line from one below, which isn't showed here.

MatejaMaric:
> Your analysis in the PR message is great; can you put that in the commit message?

Thanks @tych0, I just did.

libqtile/widget/keyboardlayout.py
tych0: Ok, can you put this in a separate patch?

MatejaMaric: @tych0 Sure, I moved it to a separate commit.

ramnes (approved):

tych0:
Thanks!

Issue #2154, opened 2021-01-27
[FEATURE REQUEST] The option to use Window Class instead of name in WindowName widget


Masacaai:
# Background

When opening windows with large names , the WindowName widget becomes congested with text, making it look rather ugly. 

# Suggestion

Allow the user to choose between using Window Name and Window Class. This would function much like an Application Name widget we see in KDE Plasma, thus allowing the WindowName widget to only display the class of the window in focus, not it's actual name.

I don't know if implementing this would be as easy as just adding an attribute to use w.class instead of w.name, but I'll try to look into this when I can.

mbenford:
I'm not fond of long window names as well, and I "fixed" it by extending the `WindowName` widget in order to truncate long titles:

```python
from libqtile import widget
from libqtile.widget import base

class WindowName(widget.WindowName):
    orientations = base.ORIENTATION_HORIZONTAL
    defaults = [
        ('max_chars', 0, 'max chars before truncating with ellipis'),
    ]

    def __init__(self, width=bar.STRETCH, **config):
        widget.WindowName.__init__(self, width=width, **config)
        self.add_defaults(WindowName.defaults)

    def truncate(self, text):
        if self.max_chars == 0:
            return text

        return (text[:self.max_chars - 3].rstrip() + "...") if len(text) > self.max_chars else text

    @widget.WindowName.text.setter
    def text(self, value):
        super(WindowName, type(self)).text.fset(self, self.truncate(value))
```

Result: 
![image](https://user-images.githubusercontent.com/446867/106395938-b7f5d700-63e3-11eb-87c3-3dac9344a53b.png)

But if you really want to replace the window name with the window class name, all you have to do is override the `update` method of `WindowName` to your liking.

Masacaai:
How do I implement your code? Do I just copy/paste it into my config file? That doesn't seem to work for me? Do I need to do anything else?

Masacaai:
Where can I find the available attributes for current_window?
I tried extending the update function to use w.class instead of w.name, but it didn't work. 
I assumed 'class' would be an attribute just like 'name' but it doesn't seem to accept that.
I tried searching through the source code, but I wasn't able to find it.
Sorry, bit of a newbie at this :)

mbenford:
Take a look at the [window.py](https://github.com/qtile/qtile/blob/master/libqtile/window.py) file and search for `wm_class`. You will find `self.window.get_wm_class()` in the `cmd_inspect` method. You can use that function in your `update` method:

[...]

PR #2177, opened 2021-02-01
Adds the option for WindowName widget to use class instead of name as addressed in #2154

Masacaai:
This is my first PR, so do let me know if I'm supposed to do something differently. Really thankful to @mbenford  for helping me out!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -51,6 +51,7 @@ Qtile X.X.X, released XXXX-XX-XX:
         - switch to Github Actions for CI
         - Columns layout has new `margin_on_single` option to specify margin
           size when there is only one window (default -1: use `margin` option).
+        - WindowName widget now allows you to use class instead of name.
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
--- a/libqtile/widget/windowname.py
+++ b/libqtile/widget/windowname.py
@@ -35,6 +35,7 @@ class WindowName(base._TextBox):
         ('show_state', True, 'show window status before window name'),
         ('for_current_screen', False, 'instead of this bars screen use currently active screen'),
         ('empty_group_string', ' ', 'string to display when no windows are focused on current group'),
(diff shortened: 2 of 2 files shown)


Masacaai:
I'm sorry if the code feels a bit newbie-ish. If there's a better way to do what I did, please let me know!

elParaguayo:
Could we do this by letting the user define the output format as a parameter? By using python's string formatting, we could limit long strings. You could make the `state`, `window.name` and `w.window.get_wm_class()[0]` values available to the formatter for the user to use as they wish.

e.g. `"{state}{name:.20}"` to only show max 20 characters of the window name.

That way you have much more flexibility over the display.

elParaguayo:
Don't think you need to close this! Was just suggesting an alternative approach.

Masacaai:
Oh, I know! I just felt that your method was better. I'm going to see if there's a way to implement what you suggested and reopen my pull request

elParaguayo:
Great.

Have a look at the [memory widget](https://github.com/qtile/qtile/blob/master/libqtile/widget/memory.py). It's a good example of how the user can specify a format string.

mbenford:
If I may weigh in, I tried to truncate and show ellipsis by using only Python's string formatting, but I failed (not a Python expert myself, so it could totally be my own fault). Perhaps `format` could be either a string or a function so one could freely decide the final value for the widget to show, but that would be the case for all widgets, not only `WindowName`.

elParaguayo:
Not your fault - I'm pretty sure python's string format doesn't have the ability to put in an ellipsis. However, the `TextLayout` used by Qtile to display text will add an ellipse if the text is longer than the size of the layout (and not set to wrap) so, if you set a max size for a task item, you should get your ellipse.

https://github.com/qtile/qtile/blob/c355e2c8fe419f00ae7284dcdbe7032a9728d5fe/libqtile/drawer.py#L48

Masacaai:
I've opened up a new PR with a format attribute for the WindowName widget.  Check #2178

PR #2178, opened 2021-02-01
Add a format attribute to WindowName widget that addresses #2154

Masacaai:
This is my first PR, so do let me know if I'm supposed to do something differently. Really thankful to @mbenford and @elParaguayo for helping me out!

I'm sorry if the code feels a bit newbie-ish. If there's a better way to do what I did, please let me know!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -52,6 +52,7 @@ Qtile X.X.X, released XXXX-XX-XX:
         - Columns layout has new `margin_on_single` option to specify margin
           size when there is only one window (default -1: use `margin` option).
         - new format attribute for WindowName widget
+        - new max_chars attribute for WindowName widget
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -43,6 +43,8 @@ Qtile X.X.X, released XXXX-XX-XX:
         - libqtile.widget.base.ThreadedPollText has been removed; out of tree
           widgets can use ThreadPoolText in the same package instead.
         - show_state attribute from WindowName widget has been removed. Use format attribute instead.
+                show_state = True  --> format = '{state}{name}'
+                show_state = False --> format = '{name}'
     * features
         - new WidgetBox widget
         - new restart and shutdown hooks
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,6 +42,7 @@ Qtile X.X.X, released XXXX-XX-XX:
           libqtile.command.*
         - libqtile.widget.base.ThreadedPollText has been removed; out of tree
           widgets can use ThreadPoolText in the same package instead.
+        - show_state attribute from WindowName widget has been removed. Use format attribute instead.
     * features
         - new WidgetBox widget
         - new restart and shutdown hooks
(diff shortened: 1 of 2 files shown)


CHANGELOG
```diff
@@ -51,6 +51,7 @@ Qtile X.X.X, released XXXX-XX-XX:
         - switch to Github Actions for CI
         - Columns layout has new `margin_on_single` option to specify margin
           size when there is only one window (default -1: use `margin` option).
+        - new format attribute for WindowName widget
```
m-col, line 54: Can you add the 'max_chars' option to this too? It's a pretty nice addition in its own right!

libqtile/widget/windowname.py
```diff
@@ -35,6 +35,8 @@ class WindowName(base._TextBox):
         ('show_state', True, 'show window status before window name'),
         ('for_current_screen', False, 'instead of this bars screen use currently active screen'),
         ('empty_group_string', ' ', 'string to display when no windows are focused on current group'),
+        ('max_chars', 0, 'max chars before truncating with ellipsis'),
+        ('format', '{state}{class}:{name}', 'format of the text'),
```
m-col, line 39: Could we set the default format to the format that it was previously, so that people who are happy with the old form don't have to go in and change their format? (ignore me if this is already the case)

libqtile/widget/windowname.py
```diff
@@ -52,6 +54,12 @@ def on_screen_changed():
             if self.for_current_screen:
                 self.update()
 
+    def truncate(self, text, char_limit):
```
m-col, line 57: There's no need to pass `char_limit` through if it is available at `self.char_limit`

m-col:
Looks great overall, so thanks for this! Just a couple comments inline. Once you've made those changes and made the checks happy, could you squash the commits down into a single commit?

Masacaai:
Wait, so were the PEP8 checks failing because I didn't have enough comments in the code?

Masacaai:
I've done the changes you requested for, but I'm still failing the PEP8 commands check. Admittedly, I didn't know PEP8 was a thing until about last week, so I might be missing something....

elParaguayo:
Your PEP8 errors seem to be mainly from you using tabs and spaces to indent (use spaces).

See the errors here: https://github.com/qtile/qtile/pull/2178/checks?check_run_id=1812641064#step:5:1405

Masacaai:
Yeah I think I mixed it up while typing. How do I distinguish tabs from spaces though? Do I just rewrite the whole thing using spaces? Or is there an easier way?

Masacaai:
There, it's finally done. I made one final change and switched the values of 'unescaped' and 'text' to be more consistent with the original code. I think only those with write access to this repository can squash and merge PRs. I'm aware there's a way for me to do it on my end, but that requires the usage of git commands, something I'm not super-familiar with as of now and I don't really have the time to learn them right now. I'll be sure to learn them for future usage sometime later. I hope you don't mind squashing the commits on your end.

Masacaai:
I've also removed the 'show_state' attribute as it's function is now handled by the 'format' attribute.

CHANGELOG
```diff
@@ -42,6 +42,7 @@ Qtile X.X.X, released XXXX-XX-XX:
           libqtile.command.*
         - libqtile.widget.base.ThreadedPollText has been removed; out of tree
           widgets can use ThreadPoolText in the same package instead.
+        - show_state attribute from WindowName widget has been removed. Use format attribute instead.
```
tych0, line 45: Can you put an example of how people would achieve the old behavior?

tych0:
I went ahead and cleaned up the history and merged this. Thanks!

PR #2179, opened 2021-02-02
Handle PAspect

ramnes:
To do:
* [x] limit resize
* [x] make float
* [x] tests

### DIFF
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -53,7 +53,8 @@ class Floating(Layout):
         Match(wm_class='notification'),
         Match(wm_class='splash'),
         Match(wm_class='toolbar'),
-        Match(func=lambda c: c.has_fixed_size())
+        Match(func=lambda c: c.has_fixed_size()),
+        Match(func=lambda c: c.has_fixed_ratio())
     ]
 
     defaults = [
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -227,6 +227,15 @@ class _Window(CommandObject):
     def has_focus(self):
         return self == self.qtile.current_window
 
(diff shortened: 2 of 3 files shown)


ramnes:
One thing I'm wondering is how to center a fixed ratio window after it gets floated when it doesn't have a fixed ratio from the start. For example, mplayer doesn't set a fixed ratio at launch, but then loads the given video and update its hints; so with this patch, it gets tiled, then it's floated after the hints update. But since it already has a position set so we're not centering it. Any idea on how to change this?

tych0:
The actual centering logic happens in `Floating.compute_client_position()` IIUC. So maybe add a flag that says "force recompute this client's position" and then retrigger a floating.layout() call?

ramnes:
Yeah, that's what I tried, without success so far. `window.place()` is called a lot (most likely more than necessary), and for some reason it blinks in the center and then is moved back.

tych0:
I'd be curious what each of the two stack traces are; once the first place() is called from compute_client_position() it does seem like it should stick, at least based on my read of the code.

ramnes:
I'm removing the centering of floating windows from the to do list here because it's actually outside the scope of this PR: the same thing happens already when you put a terminal floating, or basically anything that has been tiled and has set its position based on that. It seems to me that we should only respect PPosition and alike if the window has never been tiled before.

ramnes:
test added :)

ramnes:
Yeah, it's probably too agressive to check all float rules at each `update_hints` call. I'm not 100% sure why I went that route, but I think you can safely replace it with something lighter like `if self.has_fixed_size() or self.has_fixed_ratio()`. But then the problem is that users can't control that behavior anymore.

Issue #2103, opened 2021-01-13
test_qtile_cmd appears to be flaky


vrusinov:
# Issue description

test_qtile_cmd.py seems to be flaky. Once in a while I get an error when running `pytest . -x` from the fresh checkout. The issue pops randomly and non-determenistically. I attempted to reproduce it running `while pytest test/test_qtile_cmd.py ; do :; done` but it never failed after 10-15 minutes of running.

# Qtile version

commit cae69ccb9dd572abf5c3fe41a32c23ead36a7089
Date:   Tue Jan 12 14:06:08 2021 -0500

```
$ python3 --version
Python 3.8.6
```

xorg-server 1.20.10

# Stack traces

```
<...>
test/test_popup.py::test_popup_focus[BareConfig] PASSED       [ 45%]
test/test_qtile_cmd.py::test_qtile_cmd[ServerConfig] Fatal Python error: Aborted

Thread 0x00007f5accdf3640 (most recent call first):
  File "/usr/lib/python3.8/site-packages/gi/overrides/GLib.py", line 507 in iteration
  File "/home/rvs/workspace/qtile/libqtile/core/loop.py", line 68 in gobject_thread
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57 in run
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 80 in _worker
  File "/usr/lib/python3.8/threading.py", line 870 in run
  File "/usr/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Current thread 0x00007f5ad6601740 (most recent call first):
  File "/usr/lib/python3.8/site-packages/cairocffi/xcb.py", line 37 in __init__
  File "/home/rvs/workspace/qtile/libqtile/drawer.py", line 236 in __init__
  File "/home/rvs/workspace/qtile/libqtile/bar.py", line 214 in _configure
  File "/home/rvs/workspace/qtile/libqtile/config.py", line 289 in _configure
  File "/home/rvs/workspace/qtile/libqtile/core/manager.py", line 288 in _process_screens
  File "/home/rvs/workspace/qtile/libqtile/core/manager.py", line 152 in _configure
  File "/home/rvs/workspace/qtile/libqtile/core/manager.py", line 217 in async_loop
  File "/home/rvs/workspace/qtile/libqtile/core/session_manager.py", line 75 in async_loop
[...]

PR #2181, opened 2021-02-03
Finalize core

DorianGray:
Broke these out from https://github.com/qtile/qtile/pull/2107. This improves our finalization process by cleaning up things like window properties and hooks on shutdown. It allowed me to make xephyr a session level fixture, which seems to fix the spurious xephyr xcb connection issues that occur sometimes.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -146,6 +146,14 @@ class Core(base.Core):
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
         self._valid_mask = ~(self._numlock_mask | xcbq.ModMasks["lock"])
 
+    def finalize(self) -> None:
+        self.conn.conn.core.DeletePropertyChecked(
+            self._root.wid,
+            self.conn.atoms["_NET_SUPPORTING_WM_CHECK"],
+        ).check()
+        self.qtile = None
+        self.conn.finalize()
+
     def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
         """Get the screen information for the current connection"""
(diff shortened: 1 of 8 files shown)


DorianGray:
https://github.com/qtile/qtile/issues/2103

tych0:
I suppose it doesn't really fix things it just starts a lot less xephyrs and thus makes the race less likely. But still, quite an improvement, thanks!

libqtile/backend/x11/core.py
```diff
@@ -146,6 +146,14 @@ def __init__(self, display_name: str = None) -> None:
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
         self._valid_mask = ~(self._numlock_mask | xcbq.ModMasks["lock"])
 
+    def finalize(self) -> None:
+        self.conn.conn.core.DeletePropertyChecked(
+            self._root.wid,
+            self.conn.atoms["_NET_SUPPORTING_WM_CHECK"],
+        ).check()
+        self.qtile = None
+        self.conn.finalize()
```
ramnes, line 155: Don't we need to call `remove_listener` here?

ramnes:
Nice, thanks!

libqtile/backend/x11/core.py
DorianGray: no, the listener logic and finalization are now separate. We call remove_listener elsewhere.

DorianGray: (basically, we don't call setup_listener in init because we don't want to require an eventloop whenever we create a core. Because setup_listener is not called in init, it's not a good idea to call it during finalization)

PR #2184, opened 2021-02-04
Manage Qtile subprocess better

DorianGray:
Gather logs from subprocess, don't blindly try to retry connect if process is not running, move process target function to module scope for easier pickling.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -22,6 +22,8 @@
 # SOFTWARE.
 
 import functools
+import inspect
+import io
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


test/conftest.py
```diff
@@ -260,6 +247,45 @@ def stop_xephyr(self):
...
+        self._queue = queue
+
+    def write(self, s):
+        # This will block forever if the queue is full. Luckily, we constantly
+        # empty the queue in the main process until this process dies.
+        self._queue.put(s)
+
+
+def run_qtile(
+    *,
```
tych0, line 261: Seems like you can drop this?

tych0:
Can you put your writeup of what changed in the commit message?

DorianGray:
updated

PR #2185, opened 2021-02-04
Fix graph rendering

DorianGray:
This makes our graph rendering much more robust and efficient by skipping
"values" that wouldn't be rendered anyways. We use itertools.islice to
slice the values iterator into just the values that will actually be
drawn.

It also changes rendering so that antialiasing will not occur between steps, but only on the edges of the entire graph(as it used to be).

This was broken a bit by the drawer changes, but really these things were just being done incorrectly before. Each segment of the graphs used to be separate paths in cairo, now the graph is just a single path.

### DIFF
--- a/libqtile/widget/graph.py
+++ b/libqtile/widget/graph.py
@@ -31,6 +31,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import itertools
 import operator
 import time
 from os import statvfs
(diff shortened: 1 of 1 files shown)


libqtile/widget/graph.py
```diff
@@ -91,34 +92,43 @@ def graphwidth(self):
     def graphheight(self):
         return self.bar.height - self.margin_y * 2 - self.border_width * 2
 
+    @property
+    def step(self):
+        return max(int(self.graphwidth / float(self.samples)), 1)
```
elParaguayo, line 97: If you set the widget width to, say, 200 and leave samples at 100 then this doesn't work. `graphwidth` will be just less than 200, samples is 100 so step is 1. This means the graph only draws halfway across the widget.

DorianGray: You're right. Any suggestions? I don't even use graphs :/

The original problem was that when `cairo` draws multiple operations together(like when we stage all our drawing operations to `RecordingSurface` in `Drawer`) it does some sort of subpixel rendering when 2 vectors touch on the same pixel. The result of that was that `Graph`s suddenly had weird rendering artifacts in the form of blending at the edges of each individual sample.

DorianGray: I'll look into if there's some way I can change the code drawing on the surface to handle this better. The only other alternative is to keep track of the rounding errors and add/remove them from a sample at the end to make everything even.

elParaguayo: I would do it the other way round. Let the user define `step` and calculate the `width` and number of `samples` needed based on that. 

It's the only way I can see that you'll always get the right fit.

elParaguayo: Re-reading this - my last comment doesn't quite make sense!

elParaguayo: The options I can see are:
1) Let the user define `width` and `step` and we calculate the number of `samples` needed off the back of that; or 
2) Define `step` and `samples` and calculate the `width`.

Either of those options will result in an integer `step`. The third option, defining `width` and `samples`, will never guarantee this unless there is some manipulation of those values behind the scenes which effectively makes it the same as 1 or 2.

Based on that, option 1 would be my preferred route.

My guess is that `step` will almost always be 1. Removing it as a configurable option would simplify the issue but it would be nicer to leave for users to change if they want.

elParaguayo: I don't use `Graph` either but I like fixing problems.

DorianGray: I'd like to avoid a changelog entry for this if possible... I would prefer if I can fix it without changing the graph configuration

DorianGray: of course if it's impossible then 🤷

DorianGray:
Sorry this took so long, this PR should be ready to go now. The subpixel rendering problem is solved properly by disabling antialiasing altogether.

libqtile/widget/graph.py
```diff
@@ -91,34 +92,46 @@ def graphwidth(self):
...
+    def _prepare_context(self):
+        self.drawer.ctx.set_line_join(cairocffi.LINE_JOIN_ROUND)
         self.drawer.set_source_rgb(self.graph_color)
-        for val in values:
+        self.drawer.ctx.set_antialias(cairocffi.ANTIALIAS_NONE)
+        self.drawer.ctx.set_operator(cairocffi.OPERATOR_SOURCE)
+        self.drawer.ctx.set_line_width(self.line_width)
+
+    def draw_box(self, x, y, values):
+        self._prepare_context()
```
elParaguayo, line 115: I'll test the PR later today but am just curious whether the `_prepare_context` method needs to be called each time in the relevant `draw_` method, or whether we could just call it once in `_configure`.

DorianGray: each time... actually, we should be creating fresh contexts for each draw too but :shrug:

If the drawer is resized or otherwise modified, the context is recreated and loses it's state.

elParaguayo: Great. Thanks for explaining.

elParaguayo:
This is looking much better. Only comment (and you can choose to ignore this) is that, depending on the size or number of samples there's a small gap between the end of the graph and the border.  Maybe that's just an unavoidable consequence of using floats.

DorianGray:
could it be an off-by-one somewhere?

elParaguayo:
Could be. Don't have time to check today but will investigate more later.

elParaguayo:
Don't think it's an "Off-by-one" as it only appears some of the time. For example, width of 100 draws ok but 200 gives me a small gap.

DorianGray:
Ah, sounds like antialiasing being disabled is preventing the extra space from being blended. I'll look into making the entire graph a single draw operation so AA can still be used on the edges if you think it's worth it. Not sure how hard it will be.

DorianGray:
I need to add some logic around not setting colors if they are undefined then this should be good to go. I should have an update here tonight(8 hours or so.)

DorianGray:
Updated. Fixes for checking if a color is a str before using it(still supports lists of colors for gradients) and antialiasing between cairo paths.

PR #2186, opened 2021-02-04
Make Drawer support multiple backends

DorianGray:
This breaks out all the calls to X from drawer into a backend object
, DrawerCore, that is created by qtile's backend. When we implement a
wayland backend, we can simply make a facsimile DrawerCore. It would
provide, for example, buffers backed by EGL FBOs.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -2,12 +2,17 @@ import typing
 from abc import ABCMeta, abstractmethod
 
 from libqtile import config
+from libqtile.drawer import DrawerBackend
 
 
 class Core(metaclass=ABCMeta):
     @abstractmethod
     def finalize(self):
         """Destructor/Clean up resources"""
+
+    @abstractmethod
+    def get_drawer_backend(self, wid, *args, **kwargs) -> DrawerBackend:
+        pass
 
     @property
     @abstractmethod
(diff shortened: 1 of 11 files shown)


tych0:
Can you show some example code for what a wayland (or EGL, vulkan, or whatever else) implementation of this would look like? For example, we implement double buffering manually via copy_area(), but I can imagine other frameworks have better APIs for that type of thing and we'd need to factor things differently for those.

DorianGray:
Yeah, so, the reason the API for `DrawerBackend` looks the way it does is because of wayland terminology(except copy area).

https://wayland.freedesktop.org/docs/html/ch03.html - in the hardware enabling section at the end they give an overview of EGL
https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglSwapBuffers.xhtml - vulkan reference for the same EGL buffer swap method of sharing buffers between a client and display server.
https://emersion.fr/blog/2018/wayland-rendering-loop/ - guide on how to actually build a rendering system for wayland

I haven't written any example code yet, but this architecture matches what eveything expects...a way to allocate and swap buffers. On wayland we would call `DrawerBackend.create_buffer` and get some sort of frame buffer object, depending. When we're done working with the buffer, we can call `DrawerBackend.copy_area` which would copy the area via a buffer swap in wayland

DorianGray:
I should also note that I have multiple Drawer implementations on my local(Direct, Buffered, and Test) which can all use the same rawerBackend, meaning we can even implement multiple drawing strategies using this simple architecture

m-col:
#2485 included changes to achieve the same thing, so I'll close this. Your input is welcome (and encouraged) on how that was done! :)

PR #2187, opened 2021-02-04
Remove OrderedDict

DorianGray:
Since our minimum python is 3.7 and in python 3.7 all dicts are
ordered...remove the cruft.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -21,7 +21,6 @@
 
 import asyncio
 import os
-from collections import OrderedDict
 from typing import TYPE_CHECKING, Callable, Iterator, List, Optional, Tuple
 
 import xcffib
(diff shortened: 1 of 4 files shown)


libqtile/backend/x11/core.py
```diff
@@ -163,7 +162,7 @@ def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
         # important, because it indicates what people have chosen via xrandr
         # --primary or whatever. So we need to alias screens that should be
         # aliased, but preserve order as well. See #383.
-        xywh = OrderedDict()  # type: Dict[Tuple[int, int], Tuple[int, int]]
+        xywh = {}
```
tych0, line 165: seems like we're losing the type hint here? rest of it looks good to me

DorianGray: yeah, didn't think we needed it anymore(we don't with pytype). We still do!

tych0: We don't with pytype? not sure I understand what you mean there. if we really can drop it, that's fine.

DorianGray: mypy requires this type hint here. pytype does not, since it does type inference.

It's still needed as long as we use mypy, I've just been tinkering with getting pytype working on my local(it provides better hints)

tych0: great, thanks for that. pytype would be great if it can do more of this for us. maybe i should look into it for qtile-check too...

PR #2188, opened 2021-02-05
Remove deprecated hook

ramnes:


tych0:
Nice! Github has a new automerge thingy. Looks like there's a conflict with this, but it should be simple to resolve.

ramnes:
Yeah I activated that feature like a week ago in the project settings, hehe. :)

PR #2191, opened 2021-02-05
Deprecate Qtile.conn

DorianGray:
Qtile.conn will need to go away to make qtile with multiple backends
work.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -210,13 +210,6 @@ class Qtile(CommandObject):
         return self.core._root
 
     @property
-    def conn(self):
-        logger.warning(
-            'Qtile.conn is deprecated. Please use Qtile.core.conn instead.',
-        )
-        return self.core.conn
-
-    @property
     def selection(self):
         return self.core._selection
 
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -156,10 +156,6 @@ class Qtile(CommandObject):
         self._process_screens()
         self.current_screen = self.screens[0]
 
-        self.conn.flush()
-        self.conn.xsync()
-        self.core._xpoll()
-
         # Map and Grab keys
         for key in self.config.keys:
             self.grab_key(key)
@@ -215,6 +211,9 @@ class Qtile(CommandObject):
 
     @property
     def conn(self):
+        logger.warning(
+            'Qtile.conn is deprecated. Please use Qtile.core.conn instead.',
+        )
         return self.core.conn
 
     @property
@@ -578,7 +577,7 @@ class Qtile(CommandObject):
             del self.windows_map[win]
             self.update_client_list()
         if self.current_window is None:
-            self.conn.fixup_focus()
+            self.core.fixup_focus()
 
     def graceful_shutdown(self):
         """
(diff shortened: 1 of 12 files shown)


DorianGray:
This will be a draft until I can drop the popup skip commit. That will happen when #2186 is merged

ramnes:
Does it really make sense to do all these changes if we still use `core.conn` outside `core`? Shouldn't we make more abstract methods for cores instead?

libqtile/core/manager.py
```diff
@@ -214,6 +210,9 @@ def root(self):
 
     @property
     def conn(self):
+        logger.warning(
+            'Qtile.conn is deprecated. Please use Qtile.core.conn instead.',
```
tych0, line 214: Can we assert False in our tests some how to catch any internal uses?

tych0:
> Shouldn't we make more abstract methods for cores instead?

Let's not, for now. When we have actual working wayland code we can try to figure out how to abstract things. Without that, I don't think anyone has any idea what reasonable primitives will look like. For now, most of the methods on core are just abstraction for the sake of abstraction, which is not very useful IMO.

DorianGray:
> Does it really make sense to do all these changes if we still use `core.conn` outside `core`? Shouldn't we make more abstract methods for cores instead?

Absolutely, this is a first step. I've made a couple here already, but I am not familiar enough with wayland to know how these methods will be used in the future to design an api.

DorianGray:
> > Shouldn't we make more abstract methods for cores instead?
> 
> Let's not, for now. When we have actual working wayland code we can try to figure out how to abstract things. Without that, I don't think anyone has any idea what reasonable primitives will look like. For now, most of the methods on core are just abstraction for the sake of abstraction, which is not very useful IMO.

lol, missed this when replying to @ramnes. Agreed.

libqtile/core/manager.py
DorianGray: I can make a test fixture that patches this property to do this, maybe? It would be temporary I guess.

I think I got all the internal uses anyways...is it worth it?

ramnes:
Yeah I do agree with that, of course. But how does replacing `conn` with `core.conn` helps if we're not actually working on Wayland?

tych0:
> But how does replacing conn with core.conn helps if we're not actually working on Wayland?

I would think of it as "getting rid of aliases to the same object", vs "prep for wayland". Having a canonical reference to an object seems like an improvement, IMO.

FWIW, this is essentially the same concern I outlined in https://github.com/qtile/qtile/pull/2186#issuecomment-774320963 that we don't know enough about anything to be merging these abstractions. I don't like that PR either for this same reason.

libqtile/core/manager.py
tych0: If you think you got all the internal users, I'd just remove this. I doubt people are using the xcffib connection directly in their configs.

ramnes:
Fair enough!

DorianGray:
Hey if I can just delete conn I'm happier =)

ramnes:
I think you can

DorianGray:
updated

m-col:
Can I ask for this to not be merged? I'd like to make these changes as part of https://github.com/qtile/qtile/pull/2376 and I think it will be easier for that PR if I can have some control over where these changes fall in the git history :)

m-col:
Hey thanks for this! The other PR has been merged so `qtile.conn` is gone :)

PR #2192, opened 2021-02-06
xcbq.Window: remove _add_net_wm_state decorator

m-col:
_add_net_wm_state and the attributes that it provides to xcbq.Window are
not used within libqtile so are no longer necessary.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -433,14 +433,6 @@ class NetWmState:
         return
 
 
-def _add_net_wm_state(cls):
-    for name in net_wm_states:
-        lower_name = name.lstrip('_').lower()
-        setattr(cls, lower_name, NetWmState(name))
-    return cls
-
-
-@_add_net_wm_state
 class Window:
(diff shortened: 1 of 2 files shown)


tcld:
The corresponding class `NetWmState` in x11/xcbq.py doesn't appear to be used anymore after this change.

Also, do we still have a way to get/set these kinds of properties comfortably?

tcld:
I am currently thinking about a way to access a few of these properties in an elegant manner for the current z-PR (#2194). But, to be honest, I can't do it in six lines of code or less (even if you factor in the entire `NetWmState`-class [35 lines] it's really not that much code for the functionality it provides).

The z-PR has a few sections of code that could be shortened, I could get rid of at least 16 lines of code.

@ramnes, @m-col :
What do you think of reverting this PR? Or would it overall be more beneficial to provide ways to access only specific flags instead of having this "multitool" available?

tcld:
I want with an implementation on the `Window`-class for now. Still not sure it's the best solution, but that discussion can be led somewhere else. So please ignore my above comment for now.

PR #2193, opened 2021-02-06
move base backend into folder, add base window classes

m-col:


libqtile/backend/base/__init__.py
```diff
@@ -0,0 +1,2 @@
+from libqtile.backend.base.core import Core  # noqa: F401
```
tych0, line 1: I don't know that I love this move, it mostly just amounts to more cd-ing around when trying to reason about the code. Is there a technical reason to need this base class package?

libqtile/backend/base/window.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABCMeta
+
+
+class Window(metaclass=ABCMeta):
```
tych0, line 4: If we're introducing these, it seems like we should make window.Window a subclass and change the tests for window.Window. Otherwise it's just dead code.

libqtile/backend/base/__init__.py
m-col: A base `window.py` file will also be added, so this was just to keep them together. We could instead just have the base files within base/ and not a nested class. Also happy to do it that way!

libqtile/backend/base/window.py
m-col: For now, it would be. I'm finding it difficult to identify the best order of changes to make to move `window.Window` into the backend.

Moving method at a time from `window.Window` to `xcbw.Window` could work but would be ugly -- any usage of a given method would be converted from `window.method` to `window.window.method` until they are all moved and then they are merged properly.

Alternatively there is moving it all at once, which I submitted previously and was a big diff.

Either way, introduction of the base classes as per this addition isn't necessary until after the merge, so I'd be fine with not introducing it just yet.

Issue #676, opened 2015-06-11
Windows spawned after certain screen lockers wrongly drawn on top


TauPan:
When I lock my screen with xscreensaver-command -lock and then send myself a notification, the notification will appear over the locked screen.

Note that I do not use the notification factilities provided with qtile but run my own notification daemon from ~/.xsession (xfce4-notifyd being started on demand by notify-send).

```
xscreensaver-command -lock; sleep 3; notify-send Hullo 'I am on top'                                                                                
```

This is also true for notifications from chromium, and might pose a mild security issue because of potential sensitive contents (i.e. from google mail), depending on user setup.

I've verified that the notification windows are _not_ override-redirect windows (if they were, it appears that the only fix would be to stop using those programs, as the window manager can't do anything about them).

We've discussed this on irc before:

``` #!irc
12:32       TauPan > Hm, I just noticed that I'm affected by this: http://www.jwz.org/xscreensaver/faq.html#popup-windows
12:48       TauPan > I've just checked, and the chrome notifications that appear over xscreensaver are *not* override-redirect windows, just _NET_WM_STATE(ATOM) = _NET_WM_STATE_SKIP_TASKBAR, 
                     _NET_WM_STATE_ABOVE
12:49       TauPan > So the fix proposed by jwz to map windows with XRestackWindows instead of XRaiseWindow might apply here.
12:49       TauPan > (but I didn't inspect qtile code yet.)
[...]
17:32        tych0@> TauPan: interesting, i didn't see either in our code, actually
[...]
09:42       TauPan > tych0: you mean, xtile uses neither XRestackWindows nor XRaiseWindow (or whatever their equivalents in xcb are)?
[...]
18:01        tych0@> TauPan: well, we don't call their xcb equivalents, we might stack the windows differently, though
18:01        tych0@> now that i think abou tit, we probably do
[...]
09:29       TauPan > tych0: it's easy to reproduce for me... I lock my screen with xscreensaver and then send myself a notification via Android: Do Button -> Pushbullet and the notification 
                     will appear on top of the screensaver. It's probably even easier with xscreensaver-command -lock; sleep 3; notify-send Hullo 'I am on top' 
09:30       TauPan > tych0: That works... I just verified it.
09:30       TauPan > tych0: (note that I do not use qtiles notification widget but run xfce4-notifyd
09:30       TauPan > )
[...]
16:50        tych0@> TauPan: yeah, i don't doubt that it's a bug :)
[...]

PR #2194, opened 2021-02-06

tcld:
After a great example by @senthilbaboo (#2144), I reworked the (second) attempt at layering code (#2029) to make better use of X-functionality. This is the first implementation I finished and tested, so there will be some rough edges and room for improvement. I also have yet to see how the tests fare.

The `core.manager.change_layer()`-method is at the heart of this PR; things seems to work as well as the previous implementation so far (even a little better, since #676 seems to be fixed by this), but so far I don't really like the spaghettis in that method, especially considering that there are more X-flags we should be taking care of.

- [x] convert `_Window.cmd_keep_above()` and `_Window.cmd_keep_below()` to properties (keep the methods for easy access, though)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,8 @@ Qtile 0.18.1, released 2021-09-16:
           a further time removes the flag
         - added `cmd_move_above` and `cmd_move_below` to Windows so they can be
           shifted along z
+        - added `only_focused` setting to Max-layout, allowing to draw multiple
+          clients on top of each other when set to False
 
 Qtile 0.18.0, released 2021-07-04:
     !!! Config breakage !!!
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -34,6 +34,11 @@ Qtile 0.18.1, released 2021-09-16:
     * features
         - All layouts will accept a list of colors for border_* options with which
           they will draw multiple borders on the appropriate windows.
+        - added `cmd_keep_above` and `cmd_keep_below` to Windows so they can be
+          kept above/below other windows permanently; calling the functions
+          a further time removes the flag
+        - added `cmd_move_above` and `cmd_move_below` to Windows so they can be
+          shifted along z
 
 Qtile 0.18.0, released 2021-07-04:
     !!! Config breakage !!!
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -441,41 +441,6 @@ class XFixes:
                                                   self.selection_mask)
 
 
-class NetWmState:
-    """NetWmState is a descriptor for _NET_WM_STATE_* properties"""
-    def __init__(self, prop_name):
-        self.prop_name = prop_name
-
-    def __get__(self, xcbq_win, cls):
-        try:
-            atom = self.atom
-        except AttributeError:
-            atom = xcbq_win.conn.atoms[self.prop_name]
-            self.atom = atom
-        reply = xcbq_win.get_property('_NET_WM_STATE', 'ATOM', unpack=int)
-        if atom in reply:
-            return True
-        return False
-
-    def __set__(self, xcbq_win, value):
-        try:
-            atom = self.atom
-        except AttributeError:
-            atom = xcbq_win.conn.atoms[self.prop_name]
-            self.atom = atom
-
-        value = bool(value)
-        reply = list(xcbq_win.get_property('_NET_WM_STATE', 'ATOM', unpack=int))
-        is_set = atom in reply
-        if is_set and not value:
-            reply.remove(atom)
-            xcbq_win.set_property('_NET_WM_STATE', reply)
-        elif value and not is_set:
-            reply.append(atom)
-            xcbq_win.set_property('_NET_WM_STATE', reply)
-        return
-
-
 class Connection:
     _extmap = {
         "xinerama": Xinerama,
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -610,7 +610,7 @@ class Qtile(CommandObject):
             # Window may have been bound to a group in the hook.
             if not win.group and self.current_screen.group:
                 self.current_screen.group.add(win, focus=win.can_steal_focus)
-        self.core.update_client_list(self.windows_map)
+        self.core.update_client_list()
         self.core.update_client_stack()
         win.change_layer()
         hook.fire("client_managed", win)
@@ -626,7 +626,7 @@ class Qtile(CommandObject):
                 if c.group:
                     c.group.remove(c)
             del self.windows_map[wid]
-            self.core.update_client_list(self.windows_map)
+            self.core.update_client_list()
             self.core.update_client_stack()
 
     def find_screen(self, x: int, y: int) -> Optional[Screen]:
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -611,6 +611,8 @@ class Qtile(CommandObject):
             if not win.group and self.current_screen.group:
                 self.current_screen.group.add(win, focus=win.can_steal_focus)
         self.core.update_client_list(self.windows_map)
+        self.core.update_client_stack()
+        win.change_layer()
         hook.fire("client_managed", win)
 
     def unmanage(self, wid: int) -> None:
@@ -625,6 +627,7 @@ class Qtile(CommandObject):
                     c.group.remove(c)
             del self.windows_map[wid]
             self.core.update_client_list(self.windows_map)
+            self.core.update_client_stack()
 
     def find_screen(self, x: int, y: int) -> Optional[Screen]:
         """Find a screen based on the x and y offset"""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -99,6 +99,9 @@ class Core(CommandObject, metaclass=ABCMeta):
     def update_client_list(self, windows_map: Dict[int, WindowType]) -> None:
         """Update the list of windows being managed"""
 
+    def update_client_stack(self) -> None:
+        """Update the list of windows currently stacked by Z axis"""
+
     @contextlib.contextmanager
     def masked(self):
         """A context manager to suppress window events while operating on many windows."""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -96,7 +96,7 @@ class Core(CommandObject, metaclass=ABCMeta):
     def warp_pointer(self, x: int, y: int) -> None:
         """Warp the pointer to the given coordinates relative."""
 
-    def update_client_list(self, windows_map: Dict[int, WindowType]) -> None:
+    def update_client_list(self) -> None:
         """Update the list of windows being managed"""
 
     def update_client_stack(self) -> None:
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -218,6 +218,18 @@ class _Window(CommandObject, metaclass=ABCMeta):
         """Return a dictionary of info."""
         return self.info()
 
+    def cmd_keep_above(self) -> None:
+        """Keep this window above all others"""
+
+    def cmd_keep_below(self) -> None:
+        """Keep this window below all others"""
+
+    def cmd_move_above(self) -> None:
+        """Move this window above the next window along the z axis"""
+
+    def cmd_move_below(self) -> None:
+        """Move this window below the previous window along the z axis"""
+
 
 class Window(_Window, metaclass=ABCMeta):
     """
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -415,7 +415,7 @@ class Core(base.Core):
         """The name of the connected display"""
         return self._display_name
 
-    def update_client_list(self, windows_map: Dict[int, base.WindowType]) -> None:
+    def update_client_list(self) -> None:
         """Updates the client list
 
         This is needed for third party tasklists and drag and drop of tabs in
@@ -423,7 +423,7 @@ class Core(base.Core):
         """
         # Regular top-level managed windows, i.e. excluding Static, Internal and Systray Icons
         wids = [
-            wid for wid, c in windows_map.items() if isinstance(c, window.Window)
+            wid for wid, c in self.qtile.windows_map.items() if isinstance(c, window.Window)
         ]
         self._root.set_property("_NET_CLIENT_LIST", wids)
 
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -233,7 +233,7 @@ class Core(base.Core):
             return
 
         # Qtile just started - scan for clients
-        children = self._root.query_tree()
+        children = [window.XWindow(self.conn, i) for i in self._root.query_tree()]
         for item in children:
             try:
                 attrs = item.get_attributes()
@@ -416,7 +416,7 @@ class Core(base.Core):
         return self._display_name
 
     def update_client_list(self, windows_map: Dict[int, base.WindowType]) -> None:
-        """Updates the client stack list
+        """Updates the client list
 
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
@@ -426,7 +426,14 @@ class Core(base.Core):
             wid for wid, c in windows_map.items() if isinstance(c, window.Window)
         ]
         self._root.set_property("_NET_CLIENT_LIST", wids)
-        # TODO: check stack order
+
+    def update_client_stack(self) -> None:
+        """Set the current stacking order of clients to the given list of windows"""
+        stack = self._root.query_tree()
+        wids = [
+            wid for wid, c in self.qtile.windows_map.items()
+            if isinstance(c, window.Window) and c.group and wid in stack
+        ]
         self._root.set_property("_NET_CLIENT_LIST_STACKING", wids)
 
     def update_desktops(self, groups, index: int) -> None:
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -233,7 +233,7 @@ class Core(base.Core):
             return
 
         # Qtile just started - scan for clients
-        _, _, children = self._root.query_tree()
+        children = self._root.query_tree()
         for item in children:
             try:
                 attrs = item.get_attributes()
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -420,13 +420,7 @@ class XWindow:
 
     def query_tree(self):
         q = self.conn.conn.core.QueryTree(self.wid).reply()
-        root = None
-        parent = None
-        if q.root:
-            root = XWindow(self.conn, q.root)
-        if q.parent:
-            parent = XWindow(self.conn, q.parent)
-        return root, parent, [XWindow(self.conn, i) for i in q.children]
+        return [XWindow(self.conn, i) for i in q.children]
 
     def paint_borders(self, depth, colors, borderwidth, width, height):
         """
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -285,6 +285,28 @@ class XWindow:
         if r:
             return r[0]
 
+    def get_layering_information(self) -> Tuple[bool, bool, bool, bool]:
+        """
+        Get layer-related EMWH-flags
+        https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
+        """
+        state = self.get_net_wm_state()
+        _type = self.get_wm_type() or []
+
+        desktop = '_NET_WM_TYPE_DESKTOP' in _type
+        below = '_NET_WM_STATE_BELOW' in state
+        dock = '_NET_WM_TYPE_DOCK' in _type
+        above = '_NET_WM_STATE_ABOVE' in state
+        full = '_NET_WM_STATE_FULLSCREEN' in state
+
+        # sort the flags from bottom to top, True meaning further below than False at each step
+        return (
+            desktop,
+            below,
+            not (above or (dock and not below)),
+            not full  # not (c.has_focus and full)
+        )
+
     def configure(self, **kwargs):
         """
         Arguments can be: x, y, width, height, borderwidth, sibling, stackmode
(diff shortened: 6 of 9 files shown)


libqtile/backend/x11/xcbq.py
```diff
@@ -927,6 +927,10 @@ def fixup_focus(self):
                 xcffib.xproto.Time.CurrentTime,
             )
 
+    def query_tree(self):
+        q = self.conn.core.QueryTree(self.default_screen.root.wid).reply()
+        return list(q.children)
+
```
m-col, line 933: We already have a similar `query_tree` method on the Window class, used by the root window's instance of that class. Would it be feasible to adapt that one for your purpose, or your version here for the other one's purpose? Otherwise there are two apparent ways of doing what appears to be the same thing.

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
+    def update_client_order(self) -> None:
+        """Update X11 client order
+        This is needed for third party tasklists and drag and drop of tabs in
+        chrome
+        """
+        windows = [wid for wid, c in self.windows_map.items() if c.group]
+        stack = self.conn.query_tree()
+        # TODO: there appear to be some windows (starting VirtualBox creates one) that might not be properly unmanaged
+        windows = sorted([w for w in windows if w in stack], key=lambda wid: stack.index(wid))
+        self.core.update_client_order(windows)
```
m-col, line 428: Could you move the X-specific logic into the core itself, keeping only general high-level logic in the manager?

Also with the stuff below.

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
+            wnd = self.windows_map[wid].window
+            if above:
+                return wnd.net_wm_state_above
+            elif below:
+                return wnd.net_wm_state_below
+            else:
+                return not wnd.net_wm_state_above and not wnd.net_wm_state_below
+
+        stack = [w for w in self.conn.query_tree() if w in self.windows_map]
+        if len(stack) == 1:
```
m-col, line 453: Should this also return if `len(stack)` equals 0?

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
 
+    def update_client_order(self) -> None:
+        """Update X11 client order
+        This is needed for third party tasklists and drag and drop of tabs in
+        chrome
+        """
+        windows = [wid for wid, c in self.windows_map.items() if c.group]
+        stack = self.conn.query_tree()
+        # TODO: there appear to be some windows (starting VirtualBox creates one) that might not be properly unmanaged
+        windows = sorted([w for w in windows if w in stack], key=lambda wid: stack.index(wid))
```
m-col, line 427: You might be able to do `key=stack.index` instead of `key=lambda wid: stack.index(wid)`

m-col (changes requested):
This looks super nice (haven't tested yet). Most of it looks fine, I'd just suggest keeping things in the `Qtile` class high-level and moving as much X-specific logic into the X core as possible.

From your point of view, what stage is this at? By the looks of the commits it looks "done".

libqtile/backend/x11/xcbq.py
tcld: Yeah, I wasn't sure about this. The other method wraps the window ids with newly created windows, which seems useless overhead for the purposes of this PR. If that wasn't the case, one could just call the existing method on the root window and we wouldn't need a second implementation.

The original `query_tree` is only called in one place, I think (haven't looked through the tests, though). So it should be possible to find some common ground.

libqtile/core/manager.py
tcld: So... just call `query_tree` in the core and directly feed the result into the stacking order? Without anything in between? That would probably be much easier, actually.

libqtile/core/manager.py
tcld: It should, but since we should have at least one window (which this method is being called from), an exception might be interesting to notice.

tcld:
About what stage this is at: Since I had a pretty solid template both with the previous PR (which worked nicely for months) and the code by @senthilbaboo, this should just need a little bit of polish - thank you very much for the feedback. :)

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
+                up = False
+            else:
+                peer = next((w for w in stack if self.windows_map[w].window.net_wm_state_above), None)
+                up = peer is None
+                if peer is None:
+                    peer = stack[-1]
+        elif up:
+            peer = peers[-1]
+        else:
+            peer = peers[0]
```
ramnes, line 473: Why not moving that logic into each of the methods rather than handling it in a single method?

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
+        This is needed for third party tasklists and drag and drop of tabs in
+        chrome
+        """
+        windows = [wid for wid, c in self.windows_map.items() if c.group]
+        stack = self.conn.query_tree()
+        # TODO: there appear to be some windows (starting VirtualBox creates one) that might not be properly unmanaged
+        windows = sorted([w for w in windows if w in stack], key=lambda wid: stack.index(wid))
+        self.core.update_client_order(windows)
+
+    def change_layer(self, wid, up=True):
```
ramnes, line 430: Why adding this method to the manager rather than the Window object? Especially if its main argument is a window id.

libqtile/core/manager.py
```diff
@@ -408,14 +408,78 @@ def update_net_desktops(self) -> None:
...
+            peer = peers[-1]
+        else:
+            peer = peers[0]
+
+        # place the window
+        self.conn.conn.core.ConfigureWindow(
+            wid,
+            xcffib.xproto.ConfigWindow.StackMode | xcffib.xproto.ConfigWindow.Sibling,
+            [peer, xcffib.xproto.StackMode.Above if up else xcffib.xproto.StackMode.Below]
+        )
```
ramnes, line 480: We should not play with xcffib outside the X11 backend files. I think this belongs to `xcbq.Window.configure`, feel free to modify it to your needs.

libqtile/window.py
```diff
@@ -465,12 +465,11 @@ def place(self, x, y, width, height, borderwidth, bordercolor,
...
             height=height,
         )
-        if above:
-            kwarg['stackmode'] = StackMode.Above
-
         self.window.configure(**kwarg)
         self.paint_borders(bordercolor, borderwidth)
 
+        if above:
+            self.qtile.change_layer(self.window.wid)
```
ramnes, line 472: We're already doing two ConfigureWindow in two lines, let's not add a third one. :)

libqtile/window.py
```diff
@@ -990,7 +989,7 @@ def _reconfigure_floating(self, new_float_state=FLOATING):
                 width, height,
                 self.borderwidth,
                 self.bordercolor,
-                above=True,
+                above=above,
```
ramnes, line 992: Why adding the `above` argument to the function rather than just switching to False here?

libqtile/window.py
```diff
@@ -1387,9 +1400,9 @@ def _is_in_window(self, x, y, window):
         return (window.edges[0] <= x <= window.edges[2] and
                 window.edges[1] <= y <= window.edges[3])
 
-    def cmd_set_position(self, dx, dy):
+    def cmd_set_position(self, x, y):
         if self.floating:
-            self.tweak_float(dx, dy)
+            self.tweak_float(x, y)
```
ramnes, line 1405: I'm cherry picking this into master, please rebase.

test/test_manager.py
```diff
@@ -946,6 +946,77 @@ def test_focus_stays_on_layout_switch(manager):
...
 
+@manager_config
+@no_xinerama
+def test_window_stacking_order(manager):
+    conn = xcbq.Connection(manager.display)
+
+    def get_all_windows():
+        root = conn.default_screen.root.wid
+        q = conn.conn.core.QueryTree(root).reply()
+        return list(q.children)
```
ramnes, line 957: I guess we don't need this anymore :)

test/test_manager.py
```diff
@@ -946,6 +946,77 @@ def test_focus_stays_on_layout_switch(manager):
     assert manager.c.window.info()['name'] == 'one'
 
 
+@manager_config
+@no_xinerama
+def test_window_stacking_order(manager):
+    conn = xcbq.Connection(manager.display)
```
ramnes, line 952: You can use the `display` fixture instead, now.

libqtile/core/manager.py
ramnes, line 427: Why doing this rather than mapping into query_tree directly? Or writing another method for this specifically, that could almost replace _clients() in the tests if I'm understanding it correctly.

ramnes (changes requested):
This is getting much more understandable and clean than the two previous PR, thank you so much for all the work you're putting into this, and thank you @senthilbaboo for putting us back on the right track. :)

Let's do it!

libqtile/core/manager.py
tcld: Since the logic shuffles all windows around, I wouldn't really expect it in the `Window`-class. It also requires access to `windows_map`.  I'll think about it some more.

libqtile/window.py
tcld: I don't know what happens in the backend - is this worth thinking about?

In general I agree, but the only solution I see at the moment would be that `change_layer` just helps gather parameters for the eventual call to `configure` instead of actively changing the layer. Do you have something better in mind?

libqtile/window.py
tcld: As far as I remember, the overall reasoning for adding the parameter at all was that without it moving floating windows around got very expensive (they were layered over and over for every pixel they had been moved).
At that point I didn't dare to mess with this too much, since the effects are a bit unclear. I'll give it a go locally.

libqtile/core/manager.py
tcld: Thank you!

CHANGELOG
```diff
@@ -61,6 +61,11 @@ Qtile X.X.X, released XXXX-XX-XX:
         - new OpenWeather widget to replace YahooWeather
         - new format attribute for WindowName widget
         - new max_chars attribute for WindowName widget
+        - added `cmd_keep_above` and `cmd_keep_below` to Windows so they can be
+          kept above/below other windows permanently; calling the functions
+          twice removes the flag
+        - added `cmd_raise_client` and `cmd_lower_client` to Windows so they
```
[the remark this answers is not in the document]
tcld: I chose `cmd_move_up` and `cmd_move_down` now. I assume `cmd_keep_above` etc.

test/test_manager.py
tcld: Sorry, I just borrowed the setup lines in here from other places in this file. I am having a somewhat hard time with the tests, the typeless-ness of Python makes understanding of some of these lines impossible outside of a debugger (and I have no setup at the moment to debug tests).

So your two comments on the test confuse me. Could you maybe elaborate a bit?

libqtile/core/manager.py
tcld: To keep the functionality as is (I'm not sure it would be a good idea to just feed the full stack back into X's stacking order), I just moved the three middle lines into the called core function with the latest change to the first commit. Let me know if you agree or disagree with it.

CHANGELOG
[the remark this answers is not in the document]
tcld: Maybe `cmd_lift` and `cmd_lower` might be even better?

libqtile/backend/x11/xcbq.py
tcld: I merged the functions. Next push will contain the changes.

libqtile/core/manager.py
tcld: I reworked the function completely, now this comparison is a `< 2`.

libqtile/core/manager.py
tcld: I reworked the entire function. If I am not overlooking anything, the current solution is much more elegant.

libqtile/core/manager.py
tcld: I'm still not sure about this. We need access to potentially all other windows to find our new place in the stack. Is that something I should be accessing from a `Window`? If you think so, I'll move the method.

libqtile/core/manager.py
tcld: I agree, I'll see if I can modify `configure` to be suitable for both purposes.

libqtile/window.py
tcld: Done.

libqtile/core/manager.py
tcld: After a comment by @m-col I moved the core of this function into `backend/x11/core.py` while also simplifying it to

```python
_, _, stack = self._root.query_tree()
windows = [w for w in stack if w in windows]
```

I assume that covers your comment, too, since the entire sorting was superfluous and has been removed.

CHANGELOG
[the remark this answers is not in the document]
ramnes: I like move_down and move_up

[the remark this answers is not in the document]
ramnes: Wait, no, this might conflict with the logic behind #1875 some day. What do you think of move_below and move_above? That would be consistent with the two other methods.

tcld:
This is currently entirely broken, since I was making use of the `net_wm_...`-flags a lot, but they have been removed recently. Haven't had the time yet to find out how to set the flags now.

m-col:
Huh, sorry about that. They haven't yet been used in-tree which is why they were removed. Feel free to add a way to access the ones you need for the PR.

Just reading through that code again I realised that (NetWmState)[https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/xcbq.py#L401] is not used elsewhere in libqtile either, so should probably have been removed at the same time. Reading through that class will probably give you a good idea how to access those properties. Indeed all you need to do is do `win.window.get_property('_NET_WM_STATE_X', 'ATOM', unpack=int)` for whichever properties you are interested in.

tcld:
I think I noted that `NetWmState` isn't used anymore in a comment to your PR, did I not?

m-col:
Oh yes, you did!

libqtile/core/manager.py
tcld: I modified the call and it seems there were no further modifications needed in `xcbq.Window.configure`. So I'm going to mark this resolved.

libqtile/window.py
tcld: The current implementation avoids the third call. Please have a look and let me know whether you really prefer it - I'm not decided on it yet.^^

tcld:
I still have a few of your comments to consider, but at least the PR is now working again. I have been sorely missing the pinning functionality on my machine.

As far as I see it, the only real difficulty about this entire implementation is inside `core.manager.change_layer`. Climbing up (or down) the window tree while deciding which windows one can move beyond and which not has proven far more complex than I thought. The different EWMH-flags seem managable, but child-windows complicate the situation quite a bit.

Maybe I should just provide the most basic implementation (so the overall behaviour stays as it currently is) and then the method can be fleshed out in another PR.

kamiox:
@tcld I'm getting this crash when using your PR.
I never had this issue on the upstream Qtile but I need to investigate what are the steps to reproduce it.

>  2021-03-16 19:08:50,320 ERROR libqtile core.py:_xpoll():L305 Got an exception in poll loop
> Traceback (most recent call last):
>   File "~/qtile/libqtile/backend/x11/core.py", line 273, in _xpoll
>     ret = target(event)
>   File "~/qtile/libqtile/window.py", line 1202, in handle_PropertyNotify
>     self.update_hints()
>   File "~/qtile/libqtile/window.py", line 282, in update_hints
>     self.group.layout_all()
>   File "~/qtile/libqtile/group.py", line 164, in layout_all
>     self.floating_layout.layout(floating, screen_rect)
>   File "~/qtile/libqtile/layout/base.py", line 57, in layout
>     self.configure(i, screen_rect)
>   File "~/qtile/libqtile/layout/floating.py", line 288, in configure
>     client.place(
>   File "~/qtile/libqtile/window.py", line 469, in place
>     self.qtile.change_layer(self.window.wid, True, kwarg)
>   File "~/qtile/libqtile/core/manager.py", line 467, in change_layer
>     if stack.index(wid) == pos:
> ValueError: 107045453 is not in list

tcld:
Thank you for testing. That looks like you are trying to raise a window that X doesn't know about. Does the crash appear randomly or can you relate it to a specific program you are using?

kamiox:
> Thank you for testing. That looks like you are trying to raise a window that X doesn't know about. Does the crash appear randomly or can you relate it to a specific program you are using?

These crashes have happened when I was using Android Studio. But I don't know a particular way to reproduce it.

CHANGELOG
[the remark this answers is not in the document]
tcld: Renamed.

libqtile/window.py
tcld: I am setting this to `False` now. Have to test whether `ScratchPad`s still work properly, though. But they seem to have a few other issues at the moment anyway, so spending some time might be worth it.

tcld: This seems fine. A pinned window stayed above my `ScratchPad`, in all other cases the `ScratchPad` seemed to win out.

tcld:
I made a few small changes to the layering function while
a) implementing a few of the above suggestions and
b) splitting up the commits a little further to separate the foundation code from the stuff building on it.

@kamiox : If you could find the time to test this again, I would highly appreciate it. I couldn't reproduce your issue, but at least "non existing" windows shouldn't be raised anymore.

tcld:
I finally got around to fixing the tests again. For now I left out moving child-windows around when a parent is moved, I'll leave that for a potential future PR. It might even be better to not handle that situation at all. (The current code does, however, disallow to lower child windows.)

@ramnes: I never really understood what you meant with the comments on the test-code.

kamiox:
>  If you could find the time to test this again, I would highly appreciate it. I couldn't reproduce your issue, but at least "non existing" windows shouldn't be raised anymore.

@tcld  I have tested yesterday's changes and it worked quite well without any crashes and child windows were properly raised above their parents. I haven't tested it extensively but the overall experience was very good, it worked for me much better than the version from the master branch or even when compared to other tiling window managers.

However, after pulling today's changes something go wrong. Most of the child windows (for instance file chooser dialogs) are hidden below their parents.

tcld:
Thanks for the feedback. I had only tested the layering with a few applications and found no issues. After I read your message I tried the file chooser dialog in VSCode, and it is indeed opened behind the parent window (when opened for the first time). I'll have another look.

tcld:
Issue found and fixed.

tcld:
I (re-)added the ability to display multiple windows on top of each other in the `Max`-layout.

The code can probably be made prettier, but the functionality has proven very solid over the past few days.

tcld:
Here are the parts of my config that handle the PIP-functionality, in case somebody wants to try that out. This is particularly useful for video windows that run in "full screen" (i.e. within the fake screen):

```python
fake_screens = [  # we have to use fake_screens instead of screens
  Screen(x=0, y=0, width=1920, height=1080), # actual screen
  Screen(x=1280, y=720, width=640, height=360)  # overlayed fake screen (bottom right)
]

def shift_to_pip(group):
    def f(qtile):
        _from = qtile.current_group
        _to = qtile.groups_map[group]
        if _from == _to:
            return
        client = _from.current_window
        for c in list(_to.windows):  # empty out target group
            c.kept_above = False
            c.togroup(_from.name, switch_group=False)

        if not client:
            return
        client.togroup(group, switch_group=False)
        client.kept_above = True

    return f

def remove_all_pip(group):
    def f(qtile):
        _to = qtile.current_group
        _from = qtile.groups_map[group]
        if _from == _to:
            return
        for c in list(_from.windows):
            c.kept_above = False
            c.togroup(_to.name, switch_group=False)

    return f

groups.extend([
    Group('pip', position=10, label='PIP', layout='max')
])

keys.extend([
    Key([mod], 'p', lazy.function(shift_to_pip('pip')), desc='Move focused window to group PIP.'),
    Key([mod, alt], 'p', lazy.function(remove_all_pip('pip')), desc='Move all PIP-windows to current group.'),
])

# rough example of a GroupBox-widget that excludes the PIP-group
widget.GroupBox(visible_groups=[g.name for g in groups if g.name not in ['pip']])
```

m-col:
Could you rebase? `libqtile/window.py` has moved now into the x11 backend but I would imagine that many of the changes don't need much editing to work again.

Don't worry about implementing the Wayland side of Z stacking if you don't want to; I'm not completely sure how it would look, but it would probably make the most sense and be the easiest approach to get it working in X and then iterate on top of that implementation to get it working for Wayland. Indeed proper Z layering is very important for the Wayland backend's usability (I mean, it is for x11, but more so for wayland because at least in X the X server does some of it), so I'm eager to put more focus on this if you're up for it!

tcld:
I'll have to take some time looking at the conflicts, they don't appear to be trivial this time.

tcld:
A lot of things have been moved, so it took me almost two hours just to rebase the first commit. Maybe dropping this is the smarter approach; I'll just keep running my local version for the time being.

m-col:
OK, no worries. Apologies! If you'd like I can take a shot at sticking your changes on top of the current head and re-submit?

tcld:
Nothing to apologize for, I am very happy to see us moving towards Wayland. :)

And yes, please have a go! You likely know where to put the code, so you can probably do it in a reasonable timeframe. Although I think there will be some additional work needed to separate the `change_layer`-method to really keep all the X11-related stuff out of `core/manager.py`. This wasn't much of an issue for the first commit, but it will likely get more complicated for the later ones.

m-col:
What was the rationale for commit `Move keep_{above,below}-code into properties `?

tcld:
It adds the ability to read and explicitly set those states, which the simple functions `keep_above()` etc. won't do.

m-col:
I was playing with pushing to a fork but looks like I don't have write permission for your branch. I have rebased this PR on top current master and I'm pretty sure everything survived unscathed. No problem at all if you aren't, @tcld, but would be interested in testing out the rebased version and seeing how it is? I can submit it as a separate PR or push to this PR if you give me access. I'm also happy to open as a separate PR and work on this independently.

tcld:
I actually don't know how to give you access to my branch. But feel free to create your own branch, it's likely all your work anyway. :)

I would be happy to give your changes a test run, although I'd have to update Qtile first (couldn't live without the layering, so I haven't updated for several months now). And I'd need at least a few days for it, since I'm going to be busy at least all of tomorrow.

m-col:
> I actually don't know how to give you access to my branch. But feel free to create your own branch, it's likely all your work anyway. :)

I actually didn't need to make many changes, just move things around! I've never done the push-to-PR thing before so kinda want to try it out. I found this at bottom of the side bar on a PR I made on another repository (where I'm not a maintainer). Maybe you see the same on this page, but unchecked?

image

> I would be happy to give your changes a test run, although I'd have to update Qtile first (couldn't live without the layering, so I haven't updated for several months now). And I'd need at least a few days for it, since I'm going to be busy at least all of tomorrow.

Brilliant! I'm very eager to get this working as I don't want to put you out again, but there is no rush at all so that's all good.

tcld:
The description says that you are allowed to add commits - although I guess you want to do a forced push, right? Maybe that's not possible.

ramnes:
You can force push it it's checked. @m-col, while it's very possible to do this by hand, I recommend using GitHub's CLI for this. Just run `gh co 2194`, throw your modifications and push, the tool will do all the heavy work for you.

m-col:
Cool, got it, thanks.

tcld:
Raising and lowering windows doesn't seem to work at the moment:

```log
2021-09-05 14:38:18,414 ERROR libqtile manager.py:process_key_event():L361 KB command error keep_above: Traceback (most recent call last):
  File "/home/tcl/Software/qtile/libqtile/command/interface.py", line 313, in call
  File "/home/tcl/Software/qtile/libqtile/backend/x11/window.py", line 1944, in cmd_keep_above
    self.kept_above = not self.kept_above
  File "/home/tcl/Software/qtile/libqtile/backend/x11/window.py", line 1456, in kept_above
    atom = self.qtile.conn.atoms['_NET_WM_STATE_ABOVE']
AttributeError: 'Qtile' object has no attribute 'conn'
```

Not sure what has changed about the `Qtile`-object.

m-col:
Ah, the `conn` was removed as it's X11-specific, but it can be accessed at `self.qtile.core.conn`, so just that `core` bit needs to be added.

tcld:
With that I stop getting an exception, but nothing is raised or lowered anyway.

m-col:
I fixed that conn reference and rebased to fix that conflict. Also getting no working behaviour :(

m-col:
It's because the line in `change_layer` does `query_tree()` on the window rather than root, so that bit should be `self.qtile.core._root.query_tree()` rather than `self.window.query_tree()`.

tcld:
Yes, that seems to have fixed it. :)

tcld:
Switching layers (in `Max`-layout) and pinning windows to the bottom of the stack doesn't work. The latter one does _something_, but I can't really tell what it is.

m-col:
It seems the `kept_above` but not the `kept_below` property call `self.change_layer()`. Should the latter call it with `self.change_layer(up=False)`? How should `change_layer` be interpreted - e.g. is it "update position in response to a change in layering?", or telling the window to change layer (which would then be similar to `move_{above,below}`)?

tcld:
It's more the latter (there is a bit of documentation belonging to `change_layer`). So if you say `change_layer(up=true)` you are asking the WM to move you to the top with respect to windows that live in the same layer as you. So you will become the one-eyed king of the one-eyed people, but you won't touch the blind or the two-eyed ones. `change_layer(up=false)` does the opposite.

The doc string mentions a number of edge cases.

m-col:
When a windows WM_STATE changes to specify that the window should be kept above other windows with the hint, we are then calling `change_layer`, but should we not instead be calling a method that re-sorts the windows according to their states?

tcld:
I guess it does both. You can modify flags or not, and when you call `change_layer`, everything is going to be put into place properly.

m-col:
Huh, this PR says both "changes requested" but also that it is waiting for review. Which is it, github?

I'm going to assume that it is the latter and that I should give this another review and test. I'll make this top priority and give it a look over in the next couple of days @tcld! Any additional testing would be really beneficial as we should really get this merged, so @elParaguayo if you'd be happy to have this in your day-to-day setup as you're running X then that would be great!

elParaguayo:
OK - I'll take a look.

elParaguayo:
Yeah. None of those commands are doing anything for me.

elParaguayo:
I think this might have broken my scratchpads. They won't open on top of my browser and, instead, open behind it.

elParaguayo:
So I've actually had this PR running on my setup for a while now and I haven't seen any day-to-day issues which is great.

I don't use floating windows much but I have played around a bit with the various commands and have a few minor observations:
`cmd_move_above` - this moves to the top rather than moving above the next window in the stack. I need to look at the logic for this a bit more closely. Same issue for `cmd_move_below`.
`kept_above` currently overrides `move_below` so moving has no impact. That's a design choice rather than a bug but I was just wondering if users might expect the window to move if they do call the `move_below` command (i.e. we deactivate `kept_above`).

I'm happy to make changes. Just looking for some input.

elParaguayo:
This PR fixes #1145.

m-col:
> Also, should floating windows default to being above tiled windows?

If you mean having something in the default config then yeah - like mentioned in https://github.com/qtile/qtile/issues/3120#issuecomment-1019288990

Even if this just exists as a hook defined in the default confit then that would be useful.

> kept_above currently overrides move_below so moving has no impact. That's a design choice rather than a bug but I was just wondering if users might expect the window to move if they do call the move_below command (i.e. we deactivate kept_above).

Hmm, perhaps we can use a `force=True` optional argument to do the override? I can see how both behaviours might be desirable.

elParaguayo:
Closing this in favour of #3409

PR #2196, opened 2021-02-08
fix 'Window' object has no attribute 'window_get_wm_class'

leschalu:


libqtile/widget/windowname.py
```diff
@@ -65,7 +65,7 @@ def update(self, *args):
         else:
             w = self.bar.screen.group.current_window
         state = ''
-        if w and (w.name or w.window_get_wm_class()[0]):
```
tych0, line 68: I think it should be `w.window.get_wm_class()[0]` instead of just deleting it (i.e. it's just a `_ -> .` typo)

leschalu: It's 1st thing I tried, unfortunately w.window.get_wm_class()[0] throws same error in the statement.

tych0: Can you paste the error?

leschalu: AttributeError: 'Window' object has no attribute 'window_get_wm_class'
2021-02-08 22:56:21,108 ERROR libqtile hook.py:fire():L391 Error in hook client_name_updated

leschalu: calling info() on current window has everything but wm_class that is why I deleted it

leschalu: format works as expected eg. '{state}{name} {format}'

tych0: That suggests your fix of `_ -> .` didn't get installed correctly; I'd double check that you're running the right version.

leschalu: > That suggests your fix of `_ -> .` didn't get installed correctly; I'd double check that you're running the right version.

Correct, I had to do relog. Now I see `IndexError: tuple out of range.`
Calling without index solves it, until you add {class} to format option when there are no windows and `IndexError: tuple out of range.` will be again.

```python
if w and (w.name or w.window_get_wm_class()):
```
Simply using:
```python
if w and w.name:
```
like in pr just solves it. Is there any reason to check wm_class in statement?

tych0: Yes, if we don't check it people can't match on class any more, which would be a regression.

Looks like we also need to test that the window has a class; the error you're hitting is because it doesn't have one.

elParaguayo: Do you even need `w.name` in that line? Your just testing to see if you've got a window. Unless `w.name` is somehow linked to the ability to get the window state then I'm not sure it adds anything.

You could then check window name and class individually in lines 77-78 (i.e. when you try to retrieve those attributes) and handle errors at that point.

tych0: Yes, agreed. I hadn't actually read the code in question, thanks. @elParaguayo's suggestion seems like the correct fix.

leschalu:
```python
    def update(self, *args):
        if self.for_current_screen:
            w = self.qtile.current_screen.group.current_window
        else:
            w = self.bar.screen.group.current_window
        state = ''
        try:
            if w:
                if w.maximized:
                    state = '[] '
                elif w.minimized:
                    state = '_ '
                elif w.floating:
                    state = 'V '
                var = {}
                var["state"] = state
                var["name"] = w.name
                var["class"] = w.window.get_wm_class()[0]
                text = self.format.format(**var)
                unescaped = self.truncate(text)
            else:
                unescaped = self.empty_group_string
            self.bar.draw()
            self.text = pangocffi.markup_escape_text(unescaped)
        except IndexError:
            pass
```
I also added exception because sometimes I see IndexError: tuple out of range when {class} added to format. If that's looks good I can update pr.

tych0: I would switch the line to `var["class"] = w.window.get_wm_class() if len(w.window.get_wm_class()) > 0 else ""`; skipping the whole block means the window name won't render if there's not a class, which I think will break your use case.

tych0:
Great, thanks for the fix!

PR #2201, opened 2021-02-11
remove rpm/

tych0:
It looks like this is being maintained elsewhere: https://src.fedoraproject.org/rpms/qtile.git

The last version in our tree was 0.11.0 which is two years old. Fedora has
0.14.2 which is newer, and has corresponding changes for it. Let's drop
this from our tree.


m-col:
Awesome.

PR #2202, opened 2021-02-11
Enable column swapping with swap_column_left and swap_column_right methods

mbenford:
This PR adds two new methods to the Columns layout: `cmd_swap_column_left` and `cmd_swap_columns_right`. They swap the whole column left/right - cycling when appropriate - and it's useful when you have split windows in a column and want to move them all at once to a new position.

qtile-column-swapping

(The missing window at the end of the GIF was just a glitch in Peek)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -77,6 +77,7 @@ Qtile 0.17.0, released 2021-02-13:
               - validates that qtile can import the config file (e.g. that
                 syntax is correct, ends in a .py extension, etc.)
               - validates Key and Mouse mod/keysym arguments are ok.
+        - Columns layout now enables column swapping by using swap_column_left and swap_column_right
     !!! warning !!!
         - When (re)starting, Qtile passes its state to the new process in a
           file now, where previously it passed state directly as a string. This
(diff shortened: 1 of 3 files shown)


m-col:
Very cool. I wonder if `swap_left` and `swap_right` might be misleading and make people think it swaps windows rather than columns. That itself could be a good utility. Maybe we could rename this to `swap_column_left` and right?

mbenford:
@m-col I agree. Will rename them as `swap_column_left` and `swap_column_right`.

mbenford:
BTW, you mentioned having `swap` for windows, as other layouts do. I don't think I fully understand the differences between `swap` and `shuffle`. Could you enlighten me?

m-col:
> I don't think I fully understand the differences between swap and shuffle. Could you enlighten me?

I suppose it might depend on the layout. For instance, shuffling in Columns doesn't swap any windows if you shuffle left or right, it just moves it between columns. If you go up or down then I guess it looks like swapping. So it's probably implemented sufficiently :)

I just tested it out and it works perfectly.

Could you rebase please? Then LGTM!

mbenford:
Done! :)

mbenford:
It bothers me that my MR adds new functionality but no new documentation. On the other hand, I didn't do it in the first place because no existing `cmd_` methods in `columns.py` had any doc. Are these methods supposed to be discovered by using `qtile shell` or something?

tych0:
Might be nice to render the methods for a particular layout here: http://docs.qtile.org/en/latest/manual/ref/layouts.html I think that would help discoverability.

mbenford:
@tych0 thanks for the suggestion! Done!

m-col:
Awesome thank you @mbenford!

PR #2207, opened 2021-02-11
xcursors: drop a "warning" to "info"

tych0:
This is not really "warning" material -- only people who are using custom
cursors even care about this, which is not may people based on
qtile-examples. Let's make the log 'info'.


tych0:
I'm gonna close this. I was being too lazy to rebase https://github.com/qtile/qtile/pull/1983 where it is originally from, but then I got un-lazy and rebased it :)

Issue #2198, opened 2021-02-09
Improper behaviour of Tile layout with add_on_top=False


Tokariew:
# Issue description

With focus on master window, newly created layout will push current master to slave stack, and on slave stack newly created window will push window with focus below it.

Focus on A, new window D, push A to slave
```
|   | B    |   | A
| A | --   | D | B
|   | C    |   | C
```
second example, focus on B, new window D, created on top of it
```
|   | B    |   | D
| A | --   | A | B
|   | C    |   | C
```
# Qtile version

0.16.1

# Configuration

layout.Tile(add_on_top=False),

# fix?
0 should be 1 in the line below in file https://github.com/qtile/qtile/blob/master/libqtile/layout/tile.py
```
def add(self, client, offset_to_current=0):
```

With fix
```
|   | B    |   | D
| A | --   | A | B
|   | C    |   | C
```
```
|   | B    |   | B
| A | --   | A | D
|   | C    |   | C
```

tych0:
This is a great writeup; any chance you can send this as a patch? Including a test would be appreciated as well.

Tokariew:
I will try in the weekend to send a patch with test.
Test should be added here https://github.com/qtile/qtile/blob/master/test/layouts/test_tile.py ?

tych0:
Yep, that would be great!

PR #2212, opened 2021-02-13
Fix and test issue #2198

Tokariew:
Fix for issue #2198 

Test is passed. I tested for two cases which I described in issue.

### DIFF
--- a/libqtile/layout/tile.py
+++ b/libqtile/layout/tile.py
@@ -120,7 +120,7 @@ class Tile(_SimpleLayoutBase):
         c = _SimpleLayoutBase.clone(self, group)
         return c
 
-    def add(self, client, offset_to_current=0):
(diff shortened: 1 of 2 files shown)


tych0:
Looks like there's a pep8 error in here, but the patch itself looks good to me.

Can you also include your writeup from the issue in your commit message?

Thanks!

Tokariew:
Not sure exactly how should I do it?
Ammending message to commit will require force push?

tych0:
Yep, force pushing to this branch would be totally fine, thanks!

Tokariew:
Done, I will close my issue.

tych0:
Looks like there's still a pep8 failure,

>   test/layouts/test_tile.py:158:5: E265 block comment should start with '# '

tych0:
Great, thanks!

PR #2214, opened 2021-02-14
Update lazy objects documentation

elParaguayo:
Add section on lazy.function to show how user-defined functions can be bound to key presses.


tych0:
Thanks!

PR #2146, opened 2021-01-23
Add page on creating custom widgets to docs

elParaguayo:
Here's a first draft of documentation on how to create a widget. I hope it's of use/interest!

You may have seen the discussion (#2138) about where's the best place for documentation like this. My feeling is that it should be in the official docs (you'll see I've created a section called `Tips & Tricks`) to separate it from the more formal section.

This is a draft request for now as it refers to my dbus_next PR #1994 for instructions on how widgets can use dbus.

PR #2215, opened 2021-02-14
Add bluetooth widget

Graeme22:
Adds a new polling widget to display bluetooth status which I needed for my latest rice.
Current implementation shows one of three text outputs:
- off (adapter is currently turned off)
- on (adapter is currently turned on, but no connection is established)
- [device name] (a connection is currently established to the device)

In order to provide the functionality, the user must supply the system bus address of their hci0 device, which can be found with d-feet or similar.

Formatting of output text should work like the other widgets.

### DIFF
--- a/libqtile/widget/bluetooth.py
+++ b/libqtile/widget/bluetooth.py
@@ -1,0 +1,69 @@
+# Copyright (c) 2021 Graeme Holliday
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


elParaguayo:
You may want to take a quick look at #1994. Depending on if/when that gets merged then this will impact your widget (however, it's pretty quick to migrate from `dbus` to `dbus-next`).

Some other thoughts/suggestions:
- Is it possible to subscribe to signals and update on something like `PropertiesChanged`? I haven't looked at the bluez dbus interface for a while but would expect this to be possible.
-  You could have a mouse click turn the device on/off

Please don't take any of this as criticism; I think a bluetooth widget would be useful addition to qtile and the devs may well have different suggestions to mine.

Graeme22:
> You may want to take a quick look at #1994. Depending on if/when that gets merged then this will impact your widget (however, it's pretty quick to migrate from `dbus` to `dbus-next`).

What's the time frame on that? Will it be in the next release?

> * Is it possible to subscribe to signals and update on something like `PropertiesChanged`? I haven't looked at the bluez dbus interface for a while but would expect this to be possible.
> * You could have a mouse click turn the device on/off

Good points, I'll take a look at these when I have time.

elParaguayo:
Don't know about timings. Up to the devs. However, they are keen to include it.

It's missed the the 0.17 release so maybe it'll be in 0.18...

tych0:
> It's missed the the 0.17 release so maybe it'll be in 0.18...

I'm happy to merge it assuming you're happy with the docs. Looks like there's one small conflict?

elParaguayo:
> > It's missed the the 0.17 release so maybe it'll be in 0.18...
 
> I'm happy to merge it assuming you're happy with the docs. Looks like there's one small conflict?

Docs should be ok. Conflict is just the changelog so I can fix that.



@Graeme22 if my PR is going to be merged then I suggest you hold off this one and then update it to use dbus-next. 

Also, have a look at #2146 for some draft documentation I've written on creating widgets. There's some stuff in there about using dbus-next and listening to signals. 

Happy to answer any questions you may have too.

Graeme22:
Great. I'll close this for now and re-open with dbus-next.

> Also, have a look at #2146 for some draft documentation I've written on creating widgets. There's some stuff in there about using dbus-next and listening to signals.

Will do.

PR #2217, opened 2021-02-17
fix gmail_checker.py

zordsdavini:
widget doesn't work. There was some king of refactoring and this was not fixed


elParaguayo:
Dammit. That one's my fault. Weird because I had tested it and thought I'd caught this. 

@Tycho - sorry - that's going to be broken in 0.17

tycho:
@elParaguayo  I don't think I'm the one you intended to ping.

elParaguayo:
Nope - not having a good day obviously...

Issue #2155, opened 2021-01-27
[FEATURE REQUEST] Add measure parameter to the Memory widget


OlaRune:
_Note: I might to this myself in the future if someone else doesn't do it first, but I don't have time right now..._

# Background
When using the DF (Disk Free) widget there is a parameter called `measure` which can take the values `G`, `M` or `B`. This parameter controls in which measurement unit the data is output when using the various format variables. There is also a format variable called `m` that outputs the chosen measurement unit.

See [http://docs.qtile.org/en/latest/manual/ref/widgets.html#df](http://docs.qtile.org/en/latest/manual/ref/widgets.html#df)

# Suggestion
Implement the same behaviour for the Memory widget, with RAM and SWAP units independently selectable. Possible values are `G`, `M` or `B` with default `M`. This measure should affect all format variables outputting size, see below.

New parameters:
- `measure_mem` to set the unit for memory
- `measure_swap` to set the unit for swap

New format variables:
- `mm` to output the selected unit for memory
- `ms` to output the selected unit for swap

Format variables calculated in chosen unit:
- `MemUsed`
- `MemTotal`
- `MemFree`
- `Buffers`
- `Active`
- `Inactive`
- `Shmem`
- `SwapTotal`
- `SwapFree`
- `SwapUsed`.

# Subtasks

- [ ] Add new parameters listed above
- [ ] Add new format variables listed above
- [ ] Add calculations for each measure inspired by the DF widget source and make sure they affect all format variables listed above
- [ ] Update widget documentation

OlaRune:
Maybe also add recommendations on how to select number of decimal digits in the format string, for example (`'{MemUsed:.2f}'`).

elParaguayo:
I'd actually already had a go at doing something like this. Have a look at this branch on my fork: https://github.com/elParaguayo/qtile/tree/widget-memory

OlaRune:
> I'd actually already had a go at doing something like this. Have a look at this branch on my fork: https://github.com/elParaguayo/qtile/tree/widget-memory
> 
> This was a first go so there's probably lots that needs tidying up but you're welcome to build on it or do it your own way!

Cool!

Masacaai:
I believe this issue has been resolved. Check #2218.

OlaRune:
> I believe this issue has been resolved. Check #2218.

Cool! Well done.

Masacaai:
Thanks!

PR #2218, opened 2021-02-17
Adds the option to choose measurement units in memory widget

Masacaai:
This was done as per #2155
 I've also added the option to display decimal values.


Masacaai:
Okay, I'm confused. Why did the check take 6h to complete? And why did it fail?

tych0:
We have a race in our tests right now; I just re-ran them for you. Can you squash the pep8 fixes into their respective commits?

ramnes:
Thanks!

PR #2224, opened 2021-02-18
Add basic test for all widgets

elParaguayo:
This is a very _very_ simple test that runs on every widget and checks it initialises ok. There's a bit of hackery involved as some widgets won't initialise without some specific settings (question: should those be set by default?).

One issue that won't work here is that, where widgets have an ImportError, they are replaced by an ImportErrorWidget. Assuming that widget initialises ok the widget will still pass. What's the right approach for testing widgets with third party modules?

I've marked this as draft for now as I will see what other tests need to be drafted and may include them.

@tych0 You mentioned also configuring the widgets. I'm not sure how best to do that (i.e. giving them a proper qtile object). I'd like to do it as generically as possible to avoid writing specific tests for too many widgets while also not excluding them from tests in this file.

### DIFF
--- a/test/widgets/test_widget_init_configure.py
+++ b/test/widgets/test_widget_init_configure.py
@@ -1,0 +1,102 @@
+import pytest
+
+import libqtile.bar
+import libqtile.config
+import libqtile.confreader
+import libqtile.layout
+import libqtile.widget as widgets
+from libqtile.widget.crashme import _CrashMe
+
+# This file runs a very simple test to check that widgets can be initialised
+# and that keyword arguments are added to default values.
+#
+# This test is not meant to replace any widget specific tests but should catch
+# any mistakes that inadvertently breakag widgets.
+#
+# By default, the test runs on every widget that is listed in __init__.py
+# This is done by building a list called `parameters` which contains a tuple of
+# (widget class, kwargs).
+#
+# Adjustments to the tests can be made below.
+
+# Some widgets may require certain parameters to be set when initialising.
+# Widgets listed here will replace the default values.
+# This should be used as a last resort - any failure may indicate an
+# underlying issue in the widget that should be resolved.
+overrides = [
+]
+
+# Some widgets are not included in __init__.py
+# They can be included in the tests by adding their details here
+extras = [
+    (_CrashMe, {}),  # Just used by devs but no harm checking it works
+]
+
+# To skip a test entirely, list the widget class here
+no_test = [
+    widgets.Mirror  # Mirror requires a reflection object
+]
+
+################################################################################
+# Do not edit below this line
+################################################################################
+
+# Build default list of all widgets and assign simple keyword argument
+parameters = [
+    (getattr(widgets, w), {"dummy_parameter": 1}) for w in widgets.__all__
+]
+
+# Replace items in default list with overrides
+for ovr in overrides:
+    parameters = [ovr if ovr[0] == w[0] else w for w in parameters]
+
+# Add the extra widgets
+parameters.extend(extras)
+
+# Remove items which need to be skipped
+for skipped in no_test:
+    parameters = [w for w in parameters if w[0] != skipped]
+
+
+def no_op(*args, **kwargs):
+    pass
+
+
+class MinimalConf(libqtile.confreader.Config):
+    auto_fullscreen = False
+    keys = []
+    mouse = []
+    groups = [libqtile.config.Group("a")]
+    layouts = [libqtile.layout.stack.Stack(num_stacks=1)]
+    floating_layout = libqtile.resources.default_config.floating_layout
+    screens = []
+
+
+@pytest.mark.parametrize("widget_class,kwargs", parameters)
+def test_widget_init_config(manager_nospawn, widget_class, kwargs):
+    widget = widget_class(**kwargs)
+    widget.draw = no_op
+
+    # If widget inits ok then kwargs will now be attributes
+    for k, v in kwargs.items():
+        assert getattr(widget, k) == v
+
+    # Test configuration
+    config = MinimalConf
+    config.screens = [
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar([widget], 10)
+        )
+    ]
+
+    manager_nospawn.start(config)
+
+    i = manager_nospawn.c.bar["top"].info()
+
+    # Check widget is registered by checking names of widgets in bar
+    allowed_names = [
+        widget.name,
+        "importerrorwidget",  # Allow widgets to be replaced on import error
+        "<no name>"  # systray is called "<no name>" as it subclasses _Window
+    ]
+    assert i["widgets"][0]["name"] in allowed_names


test/widgets/test_widget_init_only.py
```diff
@@ -0,0 +1,70 @@
...
+# (widget class, kwargs).
+#
+# Adjustments to the tests can be made below.
+
+# Some widgets require certain parameters to be set when initialising.
+# Widgets listed here will replace the default values
+overrides = [
+    (
+        widgets.BitcoinTicker,
+        {"currency": "USD"}  # Set the currency to prevent locale lookup
```
tych0, line 24: Is there a reason to want to prevent the locale lookup?

elParaguayo: It made the test fail on my computer. Admittedly, that could just a problem with my locale set up rather than anything else!

tych0: would be interesting to see the stack trace. I can imagine if it's just LANG_C or something it'll fail, but so will the widget itself

test/widgets/test_widget_init_only.py
```diff
@@ -0,0 +1,69 @@
...
+    ),
+    (
+        widgets.Maildir,  # Sub-folders needs to defined in correct format
+        {"sub_folders": [{"path": "INBOX", "label": "Home mail"}]}
```
tych0, line 27: seems like you can just move this to Maildir, everyone is defining their own already anyway

elParaguayo: Yeah - that was my feeling. Will do that in a separate PR.

test/widgets/test_widget_init_only.py
elParaguayo:
```
____________________________________________________________________________________ test_widget_init[BitcoinTicker-kwargs4] _____________________________________________________________________________________

Widget = <class 'libqtile.widget.bitcoin_ticker.BitcoinTicker'>, kwargs = {'dummy_parameter': 1}

    @pytest.mark.parametrize("Widget,kwargs", parameters)
    def test_widget_init(Widget, kwargs):
        # Simple test to check whether widget initialises and sets user-defined args
>       widget = Widget(**kwargs)

test/widgets/test_widget_init_only.py:72:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
libqtile/widget/bitcoin_ticker.py:61: in __init__
    locale.setlocale(locale.LC_MONETARY, "en_US.UTF-8")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

category = 4, locale = 'en_US.UTF-8'

    def setlocale(category, locale=None):

        """ Set the locale for the given category.  The locale can be
            a string, an iterable of two strings (language code and encoding),
            or None.

            Iterables are converted to strings using the locale aliasing
            engine.  Locale strings are passed directly to the C lib.

            category may be given as one of the LC_* values.

        """
        if locale and not isinstance(locale, _builtin_str):
            # convert to string
            locale = normalize(_build_localename(locale))
>       return _setlocale(category, locale)
E       locale.Error: unsupported locale setting

/usr/lib/python3.9/locale.py:610: Error
```

tych0: I guess you don't have en_US.UTF-8 installed and don't have a default locale?

elParaguayo: My locale is:
```
LANG=en_GB.UTF-8
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC="en_GB.UTF-8"
LC_TIME="en_GB.UTF-8"
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY="en_GB.UTF-8"
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER="en_GB.UTF-8"
LC_NAME="en_GB.UTF-8"
LC_ADDRESS="en_GB.UTF-8"
LC_TELEPHONE="en_GB.UTF-8"
LC_MEASUREMENT="en_GB.UTF-8"
LC_IDENTIFICATION="en_GB.UTF-8"
LC_ALL=
```

elParaguayo: My concern with this is, depending on the setup of the CI machine (and any users running tests at home), it could result in failed tests when the widget itself is fine.

test/widgets/test_widget_init_only.py
elParaguayo: #2227

test/widgets/test_widget_init_only.py
tych0: I suspect the test framework clears locale settings before actually executing the tests, so we probably need to figure out how to propagate that setting into the tests. IMO it's fine to run with whatever locale users have configured.

elParaguayo: Sorry - don't think I was clear enough. I inserted the currency line as the test was broken on my laptop and so I assumed it would be broken on CI. I've installed "en_US.UTF-8" locally so the test works on my machine without the override and it appears to work on CI too.

tych0:
I just had a thought - it might be nice to instantiate peoples' widgets on their bar for qtile-check using the same function. Not necessarily for this PR, but maybe if we can factor this in some way that we can use it from qtile check, that might be handy.

elParaguayo:
OK - not familiar with qtile-check so will take a look. I agree it may be nice if we can factor it in.

For this test, the bit that's missing is `_configure`. Is there an easy way to get a `qtile` object that can be used to configure the widget? `bar` is easier.

elParaguayo:
Have added the configure test too.

This test will fail for widgets that would otherwise be handled by #2236 because it checks the name of the widget in the bar and doesn't (currently) allow for `configerrorwidget`.

tych0:
Thanks!

tych0:
Seems like maybe this one also causes some issues,

```
______________________________ test_widget_init_config[WindowTabs-kwargs61] ______________________________

manager_nospawn = <test.conftest.TestManager object at 0x7fa3d94c4e20>
widget_class = <class 'libqtile.widget.windowtabs.WindowTabs'>, kwargs = {'dummy_parameter': 1}

    @pytest.mark.parametrize("widget_class,kwargs", parameters)
    def test_widget_init_config(manager_nospawn, widget_class, kwargs):
        widget = widget_class(**kwargs)
        widget.draw = no_op
    
        # If widget inits ok then kwargs will now be attributes
        for k, v in kwargs.items():
            assert getattr(widget, k) == v
    
        # Test configuration
        config = MinimalConf
        config.screens = [
            libqtile.config.Screen(
                top=libqtile.bar.Bar([widget], 10)
            )
        ]
    
        manager_nospawn.start(config)
    
>       i = manager_nospawn.c.bar["top"].info()

test/widgets/test_widget_init_configure.py:94: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <libqtile.command.client.InteractiveCommandClient object at 0x7fa3d95f15e0>, name = 'top'

    def __getitem__(self, name: Union[str, int]) -> "InteractiveCommandClient":
        """Get the selected element of the currently selected object
    
        From the current command graph object, select the instance with the
        given name.
    
        Parameters
        ----------
        name : str
            The name, or index if it's of int type, of the item to resolve
    
        Return
        ------
        InteractiveCommandClient
            The current client, navigated to the specified command graph
            object.
        """
        if isinstance(self._current_node, CommandGraphRoot):
            raise KeyError("Root node has no available items",
                           name, self._current_node.selectors)
    
        if not isinstance(self._current_node, CommandGraphObject):
            raise SelectError("Unable to make selection on current node",
                              str(name), self._current_node.selectors)
    
        if self._current_node.selector is not None:
            raise SelectError("Selection already made", str(name),
                              self._current_node.selectors)
    
        # check the selection is valid in the server-side qtile manager
        if not self._command.has_item(self._current_node.parent,
                                      self._current_node.object_type, name):
>           raise SelectError("Item not available in object",
                              str(name), self._current_node.selectors)
E           libqtile.command.base.SelectError: Item not available in object, name: top, selectors: [('bar', None)]

libqtile/command/client.py:251: SelectError
------------------------------------------ Captured stderr call ------------------------------------------
```

elParaguayo:
No problem. Revert that one too and I'll take another look.

elParaguayo:
Easiest fix here after (#2243) is to skip the configuration for `ImportErrorWidget` as the wrapper seems to cause issue with pickling.

I've no idea why pickling is needed in the subprocess part though!

I'll try a bit longer to see if there's a neater fix.

elParaguayo:
It also fails on `PulseVolumeWidget` with
```
TypeError: cannot pickle '_cffi_backend.__CDataOwnGC' object
```

PR #2227, opened 2021-02-19
Fix Maildir widget defaults

elParaguayo:
Previous default `subfolders` was an empty list. If this was not
overriden then widget would crash as __init__ checks the type of
the first item in the list.

Users would be expected to set this anyway but providing a sane
default value will be useful for unit tests (e.g. #2224)


tych0:
Thanks!

PR #2231, opened 2021-02-19
Add window at the bottom of the stack option for monadX layout.

Fabien-Bernard:
Edit libqtile/layout/xmonad.py to manage the addition at the bottom of
the stack via new_at_current key.

Edit test/layout/test_xmonad.py to add 4 tests covering this option for
monadtall and monadwide layout.

In conjunction of [https://groups.google.com/g/qtile-dev/c/-cvMDny-zmo/m/1Sd5jkFgBgAJ](url)

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -293,19 +293,32 @@ class _ClientList:
             return self[idx - 1]
         return None
 
-    def add(self, client, offset_to_current=0):
+    def add(self, client, offset_to_current=0, client_position=None):
         """
         Insert the given client into collection at position of the current.
 
         Use parameter 'offset_to_current' to specify where the client shall be
         inserted. Defaults to zero, which means at position of current client.
         Positive values are after the client.
+
+        Use parameter 'client_position' to insert the given client at 4 specific
+        positions : top, bottom, after_current, before_current.
         """
-        pos = max(0, self._current_idx + offset_to_current)
-        if pos < len(self.clients):
-            self.clients.insert(pos, client)
+        if client_position is not None:
+            if client_position == "after_current":
+                return self.add(client, offset_to_current=1)
+            elif client_position == "before_current":
+                return self.add(client, offset_to_current=0)
+            elif client_position == "top":
+                self.append_head(client)
+            else:  # ie client_position == "bottom"
+                self.append(client)
         else:
-            self.clients.append(client)
+            pos = max(0, self._current_idx + offset_to_current)
+            if pos < len(self.clients):
+                self.clients.insert(pos, client)
+            else:
+                self.clients.append(client)
         self.current_client = client
 
     def append_head(self, client):
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -167,8 +167,12 @@ class MonadTall(_SimpleLayoutBase):
             "(one of ``MonadTall._left`` or ``MonadTall._right``)"),
         ("change_ratio", .05, "Resize ratio"),
         ("change_size", 20, "Resize change in pixels"),
-        ("new_at_current", False,
-            "Place new windows at the position of the active window."),
+        ("new_client_position", "after_current",
+            "Place new windows : "
+            " after_current - after the active window."
+            " before_current - before the active window,"
+            " top - at the top of the stack,"
+            " bottom - at the bottom of the stack,"),
     ]
 
     def __init__(self, **config):
@@ -203,7 +207,7 @@ class MonadTall(_SimpleLayoutBase):
 
     def add(self, client):
         "Add client to layout"
-        self.clients.add(client, 0 if self.new_at_current else 1)
+        self.clients.add(client, client_position=self.new_client_position)
         self.do_normalize = True
 
     def remove(self, client):
--- a/test/test_migrate.py
+++ b/test/test_migrate.py
@@ -167,3 +167,25 @@ def test_main():
     """)
     with pytest.raises(FileNotFoundError):
         check_migrate(noop, noop)
+
+
+def test_new_at_current_to_new_client_position():
+    orig = textwrap.dedent("""
+        from libqtile import layout
+
+        layouts = [
+            layout.MonadTall(border_focus='#ff0000', new_at_current=False),
+            layout.MonadWide(new_at_current=True, border_focus='#ff0000'),
+        ]
+    """)
+
+    expected = textwrap.dedent("""
+        from libqtile import layout
+
+        layouts = [
+            layout.MonadTall(border_focus='#ff0000', new_client_position='after_current'),
+            layout.MonadWide(new_client_position='before_current', border_focus='#ff0000'),
+        ]
+    """)
+
+    check_migrate(orig, expected)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ Qtile 0.x.x, released xxxx-xx-xx:
           dbus calls within asyncio's eventloop.
         - widget.BatteryIcon no longer has a fallback text mode; use
           widget.Battery instead
+        - MonadX layout key new_at_current is deprecated, use new_client_position.
     !!! deprecation warning !!!
         - 'main' config functions, deprecated in 0.16.1, will no longer be executed.
     * features
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -89,12 +89,39 @@ def hook_main_function(config):
     )
 
 
+# Deprecated new_at_current key replaced by new_client_position.
+# In the node, we want to change the key name
+# and adapts its value depending of the previous value :
+#   new_at_current=True => new_client_position=before_current
+#   new_at_current<>True => new_client_position=after_current
+def update_node_nac(node, capture, filename):
+    key = capture.get("k")
+    key.value = "new_client_position"
+    val = capture.get("v")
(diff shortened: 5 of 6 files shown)


ramnes:
I'm not too fond of 1/ reusing `new_at_current` for something different and 2/ using integers as values. I think we should rather make a new option and deprecate `new_at_current`, that would either take strings, or constants, or both. How does something like `new_client_position` sound, with standard values like "top", "bottom", "after_current", "before_current"?

If we implement it that way, it could be used in a lot of other layouts as well and remove quite a bunch of inconsistency between the layouts. Do you want to try standardizing these things? :)

Fabien-Bernard:
I'll try to keep in mind your good practices and I would be happy to do more.
For the moment, I have only been using monadTall and max. I will take a look at the other layouts to get a taste of what standardization means.

Fabien-Bernard:
Except for the tree-structured layout like bsp, your proposition seems to cover all the quite various cases of the different layouts. 
I will adapt my first attempt to stick to it and propose a new commit for the monadX layout.

libqtile/layout/xmonad.py
```diff
@@ -167,19 +168,40 @@ class MonadTall(_SimpleLayoutBase):
             "(one of ``MonadTall._left`` or ``MonadTall._right``)"),
         ("change_ratio", .05, "Resize ratio"),
         ("change_size", 20, "Resize change in pixels"),
-        ("new_at_current", False,
-            "Place new windows at the position of the active window."),
+        ("new_client_position", "after_current",
```
tych0, line 171: We need a migration for this if we're going to change its semantics.

tych0:
Code itself looks fine, thanks for the tests. Can you implement a migration for the config breakage?

Fabien-Bernard:
Ok, let me read about migration and I'll try to jump on the bandwagon.
Therefore I think the precautions I added in init method are superfluous.

--
(Off topic : as I remember reading something similar with qtile check command, how is the migration doing with splitted config file?)

Fabien-Bernard:
The whole migration thing is fun, remind me my young years with lex and yacc. 
Sorry for stacking avoidable commits, I was no victorious in my fight with an indomitable git.

tych0:
Just pushed a rebase, let's see if that passes CI. Usually when I am confused about how git is doing things, I do something like:

```
git co master # get to something sane
git co -b pending-patches
git cherry-pick <all the non-merge patches from your branch, fixing up conflicts>
```

Then you can do:

```
git co monadtall_add_client_bottom
git reset --hard pending-patches # make this look like the pending-patches branch we just created
git push -f <my remote> monadtall_add_client_bottom
```

That's essentially how I created the branch above (although with an additional `git rebase -i` to squash out the intermediate commits). There are other ways to fix things up, but that is generally the "nuclear option" I use when I'm confused :)

libqtile/layout/xmonad.py
```diff
@@ -167,19 +168,40 @@ class MonadTall(_SimpleLayoutBase):
...
+            "Place new windows : "
+            " after_current - after the active window."
+            " before_current - before the active window,"
+            " top - at the top of the stack,"
+            " bottom - at the bottom of the stack,"),
     ]
 
-    def __init__(self, **config):
+    def __init__(self, new_at_current=None, **config):
```
tych0, line 179: I think we should drop this, since it's unused now.

Fabien-Bernard:
Thank you really much for the explanation on git, really appreciate it. I vaguely remember that the whole nuclear thing is about deterrence, so I hope to use it not too frequently!

libqtile/layout/xmonad.py
```diff
@@ -176,32 +176,21 @@ class MonadTall(_SimpleLayoutBase):
...
-                    self.new_client_position = "before_current"
-                else:
-                    self.new_client_position = "after_current"
-            else:  # new_at_current value will not be considered.
-                logger.warning('Please use only new_client_position.')
+        try:
+            if hasattr(self, 'new_at_current'):
+                logger.warning('Monad<x> layout key new_at_current is deprecated.'
+                               ' Please use new_client_position instead.')
+        except AttributeError:
```
tych0, line 192: You don't need the `except` bit here, hasattr() doesn't throw if an attribute isn't present, it just returns false. That said, you might be able to drop this whole bit and just rely on the `migrate` code to handle this for people. Hopefully we'll have a better story about automatic migration for the next release.

test/test_migrate.py
```diff
@@ -137,3 +137,25 @@ def test_pacman():
...
+            layout.MonadTall(border_focus='#ff0000', new_at_current=False),
+            layout.MonadWide(new_at_current=True, border_focus='#ff0000'),
+        ]
+    """)
+
+    expected = textwrap.dedent("""
+        from libqtile import layout
+
+        layouts = [
+            layout.MonadTall(border_focus='#ff0000', new_client_position=after_current),
```
tych0, line 156: Just reading this test -- shouldn't "after_current" be a string?

Fabien-Bernard: Outch! 
You are right, the migration will certainly generate a non loadable config without adding quote around the value of the key.
I will push a new commit modifying the migration and the test.

tych0:
Any chance you can rebase vs. merge this? :)

Fabien-Bernard:
Ok, maybe an opportunity to experiment git rebase -i : would it be acceptable or too savage to reorder and squash the commits to left only two of them, one for the layout modification, one for the migration?

tych0:
Yep, that sounds good to me!

libqtile/layout/base.py
```diff
@@ -320,6 +320,19 @@ def append(self, client):
         """
         self.clients.append(client)
 
+    def standard_add(self, client, position="after_current"):
```
ramnes, line 323: Why another method and not merged into `add`?

Fabien-Bernard: The add method with its numeric argument is used by several layouts.
We want to change the type of this argument for 1 layout (for now).
I think as we are creating something new and different, it is better to separate the old from the new way.
Plus, by adding a new method, it reduces the risk that I insert regressions in the existing method.
TBH, it's also possible that my mind is too set by statically-typed languages.
Let me know if it's not the python way, I'm good to change it.

ramnes: I feel it makes the API more complex (i.e. "why is there two methods for the same thing?") for little gain, when it could be just another argument to the existing method. I'm sure we can make it without regression. :)

libqtile/layout/base.py
```diff
@@ -293,19 +293,34 @@ def focus_previous(self, win):
...
+            else:  # ie client_position == "bottom"
+                self.append(client)
         else:
-            self.clients.append(client)
+            pos = max(0, self._current_idx + offset_to_current)
+            if pos < len(self.clients):
+                self.clients.insert(pos, client)
+            else:
+                self.clients.append(client)
         self.current_client = client
```
ramnes, line 324: May I suggest something like this?

```python
        if client_position == "after_current":
            return self.add(client, offset_to_current=1)
        if client_position == "before_current":
            return self.add(client, offset_to_current=0)
        if client_position == "top":
            return self.append_head(client)
        if client_position == "bottom":
            return self.append(client)
            
        pos = max(0, self._current_idx + offset_to_current)
        ...
```

It makes the diff simpler and reuse the existing `offset_to_current`, which I think is good to ensure we don't break compatibility in the future. What do you think?

libqtile/layout/base.py
```diff
@@ -293,19 +293,34 @@ def focus_previous(self, win):
...
+    def add(self, client, offset_to_current=0, client_position=None):
+        """
+        If client_position is set by user, insert the given client depending of
+          its value : top, bottom, after_current, before_current.
+        Else:
+          Insert the given client into collection at position of the current.
+
+          Use parameter 'offset_to_current' to specify where the client shall be
+          inserted. Defaults to zero, which means at position of current client.
+          Positive values are after the client.
```
ramnes, line 305: It's not a big deal but I don't like the if-else structure and indentation. Let's write plain English, not pseudo-code. :)

(also, let's avoid modifying a line if we can)

ramnes (approved):
LGTM, just needs a rebase now :)

libqtile/layout/base.py
ramnes: Nah, that won't work; `append` and `append_head` do not set the current client. :(

libqtile/layout/base.py
```diff
@@ -293,19 +293,34 @@ def focus_previous(self, win):
...
+        """
+        if client_position is not None:
+            if client_position == "after_current":
+                pos = max(0, self._current_idx + 1)
+                self.clients.insert(pos, client)
```
ramnes, line 310:
```suggestion
                return self.add(client, offset_to_current=1)
```

libqtile/layout/base.py
```diff
@@ -293,19 +293,34 @@ def focus_previous(self, win):
...
+            elif client_position == "before_current":
+                pos = max(0, self._current_idx)
+                self.clients.insert(pos, client)
```
ramnes, line 313:
```suggestion
                return self.add(client, offset_to_current=0)
```

Fabien-Bernard:
Comment has been updated to limit diff.
The 2 suggested code change have been taken into account.
Rebased and force-pushed twice, because of a recalcitrant changelog combined with a half-blind conflicts resolver!

ramnes:
Thanks!

PR #2233, opened 2021-02-20
check: copy the entire config dir contents

### LINKED ISSUE #2223: Copy entire config folder for qtile check
# Issue description
My config is split across two files, `config.py` (main config) and `custom.py` (custom widgets), but `qtile check` only copies the `config.py` file to the `/tmp/tmp…/` folder; therefore, the `mypy` check fails because it cannot find the `custom` module. I can work around this by quickly copying the `custom.py` file to the folder, but this gets annoying.

This could be fixed by automatically copying the entire `~/.config/qtile` folder to the temporary folder, instead of just the `config.py`. It might also be necessary to remove the `__pycache__` folder if it is present.

# Qtile version
The latest version on the Archlinux repos, `v0.17.0-1` (not sure about exact commit hash, sadly).

# Configuration
https://github.com/MithicSpirit/dotfiles/tree/master/.config/qtile

tych0:
a fair number of people have their configs organized in multiple files;
let's support this in qtile-check by copying the entire tree instead of
just the entrypoint.

Closes #2223

### DIFF
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -111,11 +111,12 @@ def check_config(args):
     # need to do all the checking in a tempdir because we need to write stuff
     # for stubtest
     with tempfile.TemporaryDirectory() as tempdir:
-        tmp_path = path.join(tempdir, "config.py")
-        shutil.copy(args.configfile, tmp_path)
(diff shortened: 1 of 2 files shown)


ramnes (approved):
LGTM

PR #2235, opened 2021-02-20
Fix crash in image widget with missing/invalid file

### LINKED ISSUE #2226: No filename provided Image widget causes QTile to crash.
# The issue
When no filename argument, OR an invalid filename is provided for the Image widget, Qtile seems to crash, and needs to be killed to restart. You are obviously not supposed to provide a non-existant image, but I have doubts that it crashing is intended behavior. 

What I am describing here as a "crash" is no keyboard input being accepted, and windows from *all* other workspaces being displayed on the workspace you are currently on. If this is not actually a crash, I apologize, but regardless, Qtile becomes unusable until the process is killed and I am kicked back to my Display Manager.

# Steps to reproduce
In your bar, create a new ``Image`` widget somewhere inside. Either provide a path to an image that does not exist, or do not provide one period.

# Qtile version
This is the commit hash of the version I am running.
6c4d0557124989d46ffb2bb24f4468db687fcdb2

# Stack traces
No stack traces from xsession-errors, or the Qtile log are produced, however I have traced the error (through using the logger provided in the module's file) to the ``_configure`` method of the Image widget, and it seems to be coming the line: ``base._Widget._configure(self, qtile, bar)``

# Configuration
https://pastebin.com/qxBq6yPn

If there is any information I got wrong here, or some other bit of information I can provide that will help this issue get solved, I will try my best.

elParaguayo:
Fixes #2226

This should fix the immediate problem reported in the issue.

However, should we also include better error handling in `bar._configure` e.g. if there's an exception when configuring the widget, log it, remove the widget from the bar and continue?

### DIFF
--- a/libqtile/widget/image.py
+++ b/libqtile/widget/image.py
@@ -53,14 +53,17 @@ class Image(base._Widget, base.MarginMixin):
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
+        self.img = None
 
         if not self.filename:
-            raise ValueError("Filename not set!")
+            logger.warning("Image filename not set!")
+            return
 
         self.filename = os.path.expanduser(self.filename)
 
         if not os.path.exists(self.filename):
-            raise ValueError("File does not exist: {}".format(self.filename))
+            logger.warning("Image does not exist: {}".format(self.filename))
+            return
 
         img = Img.from_path(self.filename)
         self.img = img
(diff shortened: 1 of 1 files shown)


tych0:
> However, should we also include better error handling in bar._configure e.g. if there's an exception when configuring the widget, log it, remove the widget from the bar and continue?

Yeah, that would be great!

PR #2236, opened 2021-02-20
Improve error handling when widgets configured

elParaguayo:
Rather than crashing qtile completely, the bar will now replace a broken widget with a warning for the user.

This is messy/proof of concept* so interested in thoughts here.

*There will be errors in tests.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -22,6 +22,7 @@ from typing import Union
 
 from libqtile import configurable, drawer, window
 from libqtile.command.base import CommandObject
+from libqtile.log_utils import logger
 
 
 class Gap(CommandObject):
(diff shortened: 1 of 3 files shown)


tych0:
Any chance you can add a test for this?

libqtile/bar.py
```diff
@@ -201,8 +203,18 @@ def _configure(self, qtile, screen):
 
         if self.window:
             self.window.place(self.x, self.y, self.width, self.height, 0, None)
+            crashed_widgets = []
             for i in self.widgets:
-                i._configure(qtile, self)
+                try:
```
elParaguayo, line 208: There's some duplication here and below. Will look to refactor.

tych0:
Thanks!

tych0:
Hrm, seems like something about this made CI unhappy?

elParaguayo:
Really? Tests all ran ok when I pushed it.

Do you want to revert it and I can take a look?

tych0:
Yeah, although I merged a bunch of stuff all before this one, so it's probably the interactions between separate PRs. It happens, I'm looking at it now. If I can't figure it out shortly I'll revert. The tests seem to be running fine (although I'm only 50% of the way through) with things reverted, so it's only this one patch.

elParaguayo:
No problem. I can rebase onto latest master and rework it.

elParaguayo:
So this will be fixed by #2243 so nothing more to do here.

PR #2237, opened 2021-02-21
qtile: add migrate command

### LINKED ISSUE #1773: idea: qtile-config {check,migrate}

tych0:
Here's an implementation of the proposed `qtile migrate` command.

For now, I've used pybowler. While they say in their docs that their API
may not be stable, it seems mostly unchanged since 2016 when it was
introduced. I played around with other refactoring packages like rope, and
pybowler ends up being much simpler for the most part.

I tried to write a variety of migrations from our last release that people
complained about. One I looked at briefly but couldn't figure out was how
to automatically remove the qtile arg mouse_callbacks in widget
constructors. I think it's probably reasonably easy if anyone is
interested.

One thing I didn't do was write any sort of policy about what kinds of
requirements we want to place on API breaking changes, or what we want to
state as our official API. For now, I'd like to require any API breaking
changes at least do a good faith effort at a migration, and let's see how
much of a burden it places. But others' thoughts are welcome on this policy
:)

I'm going to say this:

Closes #1773

although the conversation about API stability continues...


libqtile/scripts/migrate.py
```diff
@@ -0,0 +1,116 @@
...
+            bowler.Query(config)
+            .select_module(fro)
+            .rename(to)
+        )
+    MIGRATIONS.append(f)
+
+
+def do_migrate(args):
+    if "bowler" not in sys.modules:
+        print("bowler can't be found, not mirating config file")
```
ramnes, line 88:
```suggestion
        print("bowler can't be found, not migrating config file")
```

ramnes (approved):
Awesome!

libqtile/scripts/migrate.py
tych0: Good catch, thanks!

libqtile/scripts/migrate.py
```diff
@@ -0,0 +1,116 @@
...
+            bowler.Query(config)
+            .select_module(fro)
+            .rename(to)
+        )
+    MIGRATIONS.append(f)
+
+
+def do_migrate(args):
+    if "bowler" not in sys.modules:
+        print("bowler can't be found, not mirgating config file")
```
m-col, line 88: Third time's a charm?

m-col:
Admittedly I have been a bit skeptical about how feasible this was but this looks really cool!

The specific changes that we add to command only make sense for migrating from one Qtile release to another (which I assume is ultimately the purpose of the command, less so for those git-tracking users). Therefore it might be worth labelling each migration with the relevant release (e.g. `"for 0.17.0"` in the docstring). This way we can remove them in a release or two.

> For now, I'd like to require any API breaking
changes at least do a good faith effort at a migration, and let's see how
much of a burden it places.

This seems like a sensible approach. I just had a quick read about bowler and the API looks pretty nice, but having examples within our own code will definitely help others who need to add them for their own API-breaking changes. The migrate command itself that you have implemented looks pretty straightforward and understandable.

Are the pyi files for mypy or for bowler?

libqtile/scripts/migrate.py
tych0: bleh!

tych0:
> Are the pyi files for mypy or for bowler?

For our mypychecks, bowler doesn't export type info unfortunately.

tych0:
I'm gonna go ahead and merge this for now. We should figure out if/how we want to make this automatic (only when config fails to load, every time, etc) and also it would be great if other folks play around with bowler. I'm definitely not married to it as a library to use for this (and perhaps there's some things we could build our own crude parsers for, etc). But no way to find out answers to these questions until people start playing with it.

PR #2240, opened 2021-02-21
Add wttr widget

akkaky:
Hi!
I made a simple weather widget for wttr.in service. Maybe it would be interested.
Sorry, if I made something wrong. This is my first commit ever. Waiting for yours advises and reviews.

### DIFF
--- a/libqtile/widget/wttr.py
+++ b/libqtile/widget/wttr.py
@@ -41,7 +41,6 @@ class Wttr(GenPollUrl):
 
     orientation = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('json', False, "Don't set ``True``"),
         (
             'format', '3',
             'Display text format. Choose presets in range 1-4 (Ex. ``"1"``) '
@@ -66,7 +65,7 @@ class Wttr(GenPollUrl):
     ]
 
     def __init__(self, **config):
-        GenPollUrl.__init__(self, **config)
+        GenPollUrl.__init__(self, json=False, **config)
         self.add_defaults(Wttr.defaults)
         self.url = self._get_url()
 
--- a/libqtile/widget/wttr.py
+++ b/libqtile/widget/wttr.py
@@ -62,6 +62,12 @@ class Wttr(GenPollUrl):
             "``'m'`` - metric, ``'M'`` - show wind speed in m/s, "
             "``'s'`` - imperial"
         ),
+        (
+            'update_interval', 600,
+            'Update interval in seconds. Recommendation: if you want to '
+            'display multiple locations alternately, maybe set a smaller '
+            'interval, Ex. 30.'
+        )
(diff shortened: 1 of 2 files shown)


libqtile/widget/wttr.py
```diff
@@ -0,0 +1,88 @@
...
+    location: ``~Vostok Station`` or ``~Eiffel Tower``.
+
+    Also can use IP-addresses (direct) or domain names (prefixed with @) to
+    specify a location:
+    ``@github.com``, ``123.456.678.123``
+    """
+
+    orientation = base.ORIENTATION_HORIZONTAL
+    defaults = [
+        ('json', False, "Don't set ``True``"),
```
elParaguayo, line 44: If this should never be set to `True` then can you just remove it from `defaults` entirely and hard code it into L69:
``` python
GenPollUrl.__init__(self, json=False, **config)
```
or just do:
``` python
self.json = False
```
at the end of `__init__`?

akkaky: OK. I will do it.

libqtile/widget/wttr.py
```diff
@@ -0,0 +1,87 @@
+from urllib.parse import quote, urlencode
+
+from libqtile.widget import GenPollUrl, base
+
+
+class Wttr(GenPollUrl):
```
elParaguayo, line 6: `GenPollUrl` inherits `ThreadPoolText` which has a default `update_interval` of 600 seconds: https://github.com/qtile/qtile/blob/7ec0bdd002ea44417bb2c439aca466f382ce80a5/libqtile/widget/base.py#L515-L518

Is this something that you want your users to be able to update? If so, add it to your `defaults` so users know they can change it (it'll still work if they set it in their config but this would make it clearer in the docs).

akkaky: I forgot about it. I add it now.
And something else. Are docstrings without mistakes? My English is not very good.

elParaguayo: Sorry - my advice has been terrible! I've just generated the docs for your PR and can see that it adds all the defaults from inherited classes, so `json` is listed etc.

That's a little bit frustrating because you won't want users to change most of those.

Let me ask a question to the devs on this.

akkaky: Will waiting an answer.)

tych0: LGTM, I don't think the json-rendering-in-docs issue is a problem you need to solve for this widget. Thanks!

elParaguayo: Sounds good. 

Also, would just add that I like this widget. Very neat. Thanks @akkaky

akkaky: Wow! Thanks! My first approved PR!

PR #2243, opened 2021-02-22
tests: fix a couple of config instantiations

tych0:
622b69428c3b ("Manage Qtile subprocess better") rearranged the way configs
get instantiated, and run_qtile() now automatically instantiates them if
the class object is passed.

However, these two tests were making modifications to the class object,
which were then getting discarded by the instantiation in run_qtile().

We can't just drop this instantiation, because Config.load expects a self
argument:

  File "/home/tycho/packages/qtile/libqtile/core/manager.py", line 104, in load_config
    self.config.load()
TypeError: load() missing 1 required positional argument: 'self'

which means if we're going to make local modifications to config objects, we
need to instantiate them before we do so.

### DIFF
--- a/libqtile/widget/import_error.py
+++ b/libqtile/widget/import_error.py
@@ -21,9 +21,14 @@
 from libqtile.widget.base import _TextBox
 
 
+class ImportErrorWidget(_TextBox):
+    def __init__(self, class_name, **config):
+        _TextBox.__init__(self, **config)
+        self.text = "Import Error: %s" % class_name
+
+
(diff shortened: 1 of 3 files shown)


tych0:
/cc @elParaguayo 

this fixes some of the test failures, but there's a few more it looks like.

elParaguayo:
> /cc @elParaguayo
> 
> this fixes some of the test failures, but there's a few more it looks like.

Thanks - it's just failing on the import error modules. Investigating now.

PR #2245, opened 2021-02-22
Fix widgetbox test

elParaguayo:
This should remove the warning about not awaiting a coroutine which was introduced when #1994 was merged.

### DIFF
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -100,11 +100,7 @@ class WidgetBox(base._Widget):
         self._widgets = widgets
         self.add_callbacks({"Button1": self.cmd_toggle})
 
-        if self.close_button_location not in ["left", "right"]:
-            val = self.close_button_location
-            msg = "Invalid value for 'close_button_location': {}".format(val)
-            logger.warning(msg)
-            self.close_button_location = "left"
+        self.cmd_set_location(self.close_button_location)
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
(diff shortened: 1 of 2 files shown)


tych0:
Hrm, I ran with this (on my fix-some-ci branch @ 97f229480b80e2ac640b2f60c51978542901fdfd) and I'm still seeing:

```
============================ warnings summary ============================
test/widgets/test_check_updates.py::test_update_available
test/widgets/test_check_updates.py::test_no_update_available_without_no_update_string
test/widgets/test_check_updates.py::test_no_update_available_with_no_update_string_and_color_no_updates
  /home/tycho/packages/qtile/libqtile/widget/base.py:206: RuntimeWarning: coroutine '_Widget._config_async' was never awaited
    self.qtile.call_soon(asyncio.create_task, self._config_async())

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

elParaguayo:
This is going to be an issue for any widget that is `_configure`d manually in a test (as opposed to via a manager).

This is because the `base._Widget` uses `call_soon` to start the `_config_async` in the eventloop. However, here, `call_soon` is a no_op which is receiving a coroutine as one of its args.

If we can have the `no_op` run the coroutine in short loop then that might be enough to stop this.

Let me take a look later today.

elParaguayo:
Quick bit of googling and pretty sure I've got a fix for this. Might be worth making a standard `FakeQtile` for widgets that deals with this.

Would also mean all of the changes in this PR aren't actually needed!!

elParaguayo:
Closed in favour of #2248.

PR #2246, opened 2021-02-22
Fix generic widget test

elParaguayo:
Fixes residual errors in #2224 by moving ImportErrorWidget out of wrapper (I'm not familiar with pickling so very happy to change this if there's a better way to handle the problem).

Removes PulseVolume widget from test for now as it fails with 
```
TypeError: cannot pickle '_cffi_backend.__CDataOwnGC' object
```


tych0:
I'm gonna close this, as I merged it as part of #2243.

PR #2248, opened 2021-02-23
Fix runtime warnings in widget tests

elParaguayo:
Creates a fake_qtile fixture for widget tests that safely handles new _config_async coroutine that was introduced in PR #1994.

### DIFF
--- a/test/widgets/conftest.py
+++ b/test/widgets/conftest.py
@@ -25,3 +25,29 @@ def svg_img_as_pypath():
     )
     audio_volume_muted = py.path.local(audio_volume_muted)
     return audio_volume_muted
+
+
+@pytest.fixture(scope='module')
+def fake_qtile():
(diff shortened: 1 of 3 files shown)


elParaguayo:
Still a couple of warnings here but I don't think they're related to #1994.

However, one warning references asyncio so I'll take a quick look:
```
  test/widgets/test_misc.py::test_textbox_color_change[WidgetTestConf]
  test/widgets/test_misc.py::test_textbox_color_change[WidgetTestConf]
  test/widgets/test_misc.py::test_textbox_color_change[WidgetTestConf]
    /opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/asyncio/base_events.py:626: ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
      source=self)
```

PR #2249, opened 2021-02-24
Add tooltip mixin

elParaguayo:
I've been working on adding a mixin which allows any widget to have the ability to display a tooltip. Looks like this:

widget

The tooltip is a just a `Popup` window. Having it as a mixin means you can have consistently styled tooltips across widgets which users can theme as appropriate.

This is draft for now but wanted to show it here in case people had any thoughts on this.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -36,6 +36,7 @@ from typing import Any, List, Tuple
 from libqtile import bar, configurable, confreader, drawer
 from libqtile.command.base import CommandError, CommandObject
 from libqtile.log_utils import logger
+from libqtile.popup import Popup
 
 
 # Each widget class must define which bar orientation(s) it supports by setting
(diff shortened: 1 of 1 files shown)


m-col:
I'm glad to see this being worked on.

I think tooltips would be the most useful if they could be given widgets. This opens up a lot more possible use cases for tooltips (e.g. hover over an MPD widget showing the current song, see a tooltip of another MPD widget showing controls, or more info). I imagine that doing so is also relatively straight forward, as all of the code that handles widget dimensions and interactivity is already written. Perhaps widgets don't need to be aware they are on a tooltip and will use `self.bar` as they would do on the actual bar. What do you think?

m-col:
The approach above might not fit with the use of a Mixin, as it might require an option for `_Widget` that will accept a widget. This makes it possible on all widgets rather than just those with a mixin.

elParaguayo:
Let's move this to a discussion #2251

PR #2253, opened 2021-02-25
Warnings errors2

tych0:
Let's make warnings errors (again) :)


tych0:
```
  No sensors found!
  Make sure you loaded all the kernel drivers you need.
  Try sensors-detect to find out which these are.
```

Heh, looks like github actions don't have a real /sys :)

ramnes:
Hell yeah, kill the bloat

PR #2258, opened 2021-02-26
ci: add dead code analysis

tych0:
Not clear if this is really that useful, since it didn't catch a ton of
actually useful stuff. But I did it, so let's see. Using confidence < 100
is more interesting (patches to come), but it also has a ton of false
positives.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -74,9 +74,15 @@ commands =
 deps = -r{toxinidir}/docs/requirements.txt
 commands = python3 setup.py build_sphinx -W
 
+[testenv:vulture]
+deps =
+    vulture
+commands =
+    vulture --min-confidence=100 --exclude test/configs/syntaxerr.py libqtile test
+
 [gh-actions]
 python =
     pypy-3.7: pypy3
     3.7: py37
     3.8: py38
-    3.9: py39, packaging, format, pep8, mypy, docs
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -74,9 +74,15 @@ commands =
 deps = -r{toxinidir}/docs/requirements.txt
 commands = python3 setup.py build_sphinx -W
 
+[testenv:vulture]
+deps =
+    vulture
+commands =
+    vulture --min-confidence=100 --exclude test/configs/syntaxerr.py libqtile test
```
m-col, line 81: So this will fail tests only when it finds something it is 100% sure about?

m-col:
What does prepending the underscores to kwargs do?

ramnes (approved):

tox.ini
tych0: Yes, exactly. Unfortunately due to the nature of python, that is not very much stuff.

tych0:
> What does prepending the underscores to kwargs do?

It's a convention to say "this is an unused parameter", which causes vulture to ignore the fact that it's unused. This IMO is better than `#noqa` comments where we can't delete args (e.g. test fixtures or because they're arguments defined by a superclass or whatever). I did have to use `noqa` in a couple of places, because changing the names of arguments results in failures due to pytest's crazy introspection :(

PR #2259, opened 2021-02-26
sh: clean up termwidth logic

tych0:
...and also drop the update_termwidth=True thing, which was always true.
This allows us to lower the self.termwidth attribute to a local, since it
is only used locally.

(Yes, it's inefficient, but nobody has complained since it's been this way
for a long time, so let's just live with it.)


tych0:
Argh. DI for tests, whee...

PR #2260, opened 2021-02-26
More dead attrs

tych0:
Some more stuff that vulture found with 60% confidence.


m-col:
Nice. Kinda surprised these kinds of things aren't picked up by some of the tools we run in the tests. Vulture is good :)

PR #2263, opened 2021-02-26
Fix for (possibly) unwanted logging behaviour

### LINKED ISSUE #2242: logging: possibly unwanted behaviour: 'smart' is reported as invalid for `focus_on_window_activation`
I stumbled across an entry in my log: `Invalid value for focus_on_window_activation: smart`.
This piece of code is responsible for that:
https://github.com/qtile/qtile/blob/3bee7be1713e3970e2c5f03a840b355048f6c1e4/libqtile/window.py#L1163-L1177
If `focus_behavior == "smart" and self.group.screen and self.group.screen != self.qtile.current_screen` then the final `else` case is the only one matching and hence `smart` is logged although it is a valid option.

Suggestion: Change
```py
elif focus_behavior == "smart" and self.group.screen and self.group.screen == self.qtile.current_screen: 
     logger.info("Focusing window") 
     self.qtile.current_screen.set_group(self.group) 
     self.group.focus(self)
```
to
```py
elif focus_behavior == "smart" and self.group.screen:
     if self.group.screen == self.qtile.current_screen: 
         logger.info("Focusing window") 
         self.qtile.current_screen.set_group(self.group) 
         self.group.focus(self) 
    else:
        logger.info("Ignoring focus request")
```
or change
```py
elif focus_behavior == "never":
```
to
```py
elif focus_behavior == "never" or (focus_behavior == "smart" and self.group.screen and self.group.screen != self.qtile.current_screen):
```
The first option is longer, but more readable...

Version: newest,
config: `focus_on_window_activation = 'smart'`

ep12:
Fixes #2242:
The focus request handler has/had a not very readable collection of cases. One case fell through and triggered a warning that shouldn't have been emitted. This pr fixes this fall-through and makes the corresponding section more readable.

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1164,14 +1164,18 @@ class Window(_Window):
                     logger.info("Focusing window")
                     self.qtile.current_screen.set_group(self.group)
                     self.group.focus(self)
-                elif focus_behavior == "smart" and self.group.screen:
+                elif focus_behavior == "smart":
+                    if not self.group.screen:
+                        logger.info("Ignoring focus request")
+                        return
                     if self.group.screen == self.qtile.current_screen:
                         logger.info("Focusing window")
                         self.qtile.current_screen.set_group(self.group)
                         self.group.focus(self)
-                    else:
-                        logger.info("Ignoring focus request")
-                elif focus_behavior == "urgent" or (focus_behavior == "smart" and not self.group.screen):
+                    else:  # self.group.screen != self.qtile.current_screen:
+                        logger.info("Setting urgent flag for window")
+                        self.urgent = True
+                elif focus_behavior == "urgent":
                     logger.info("Setting urgent flag for window")
                     self.urgent = True
                 elif focus_behavior == "never":
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1164,10 +1164,13 @@ class Window(_Window):
                     logger.info("Focusing window")
                     self.qtile.current_screen.set_group(self.group)
                     self.group.focus(self)
-                elif focus_behavior == "smart" and self.group.screen and self.group.screen == self.qtile.current_screen:
-                    logger.info("Focusing window")
-                    self.qtile.current_screen.set_group(self.group)
-                    self.group.focus(self)
+                elif focus_behavior == "smart" and self.group.screen:
+                    if self.group.screen == self.qtile.current_screen:
+                        logger.info("Focusing window")
+                        self.qtile.current_screen.set_group(self.group)
+                        self.group.focus(self)
+                    else:
+                        logger.info("Ignoring focus request")
                 elif focus_behavior == "urgent" or (focus_behavior == "smart" and not self.group.screen):
                     logger.info("Setting urgent flag for window")
                     self.urgent = True


libqtile/window.py
```diff
@@ -1164,10 +1164,13 @@ def handle_ClientMessage(self, event):  # noqa: N802
...
-                    self.qtile.current_screen.set_group(self.group)
-                    self.group.focus(self)
+                elif focus_behavior == "smart" and self.group.screen:
+                    if self.group.screen == self.qtile.current_screen:
+                        logger.info("Focusing window")
+                        self.qtile.current_screen.set_group(self.group)
+                        self.group.focus(self)
+                    else:
+                        logger.info("Ignoring focus request")
                 elif focus_behavior == "urgent" or (focus_behavior == "smart" and not self.group.screen):
```
tych0, line 1174: It looks like this is another case in this if-else chain that would break; can you fix this one up too?

This whole thing might be made cleaner by having a `needs_warning = True` that gets set to `False` in the various cases, vs. nesting things more. Whatever way you think looks best is fine with me.

tych0:
Also, can you update the commit message with some of discussion from your bug report?

libqtile/window.py
ep12: I am not quite sure if I understand you correctly? What would be messed up?
See #2242: I found a warning and traced it back to this piece of code. The only way I could figure out why this message was shown is the setup in the issue description (`focus_behavior == "smart" and self.group.screen and self.group.screen != self.qtile.current_screen` as this fell through before.
Could you clarify what might be broken and needs fixing?

tych0: It looks like the next line in this if-else chane (line 1174, the one this comment thread is on) has the same bug as the one you fixed: it has additional tests in the if condition that would cause extra logging messages to be emitted when they shouldn't.

I think your fix is correct, and that there are additional instances of this same class of bug.

ep12: Oh I see. The `focus_behavior == "smart"` branch is split into three parts which is quite unlucky: The third piece is on line 1174:
Assuming `focus_behavior == "smart"`, there are two cases:
Case A: `self.groups.screen == true`
  + checked: line 1167
  + Two sub-cases:  
    Case A.i: `self.groups.screen == self.qtile.current_screen` (line 1168)
    Case A.II: `self.groups.screen != self.qtile.current_screen` (line 1172)
Case B: `self.groups.screen == false`
  + checked: line 1174
So I don't think the setting "smart" should ever allow a fall-through, right?

ep12: And "focus", "urgent" and "never" also seem to be fine <=> impossible to fall through

tych0: Yes, exactly. It may be cleaner to have a `needs_warning` bool vs. rearranging all the if conditions, whatever looks best to you is fine with me.

ep12: Hm. We want to throw a warning like [this](https://github.com/qtile/qtile/blob/9ffc1c6b8e540b56d9297ea3b19ec76fa0fcde0f/libqtile/window.py#L1177) one if `focus_behavior not in {"smart", "focus", "urgent", "never"}`.
Since we want to deal with all those cases to act appropriately and throw debugging messages, the `if`-`elif` block is necessary. I don't see how another local variable could be cleaner than just using the `else` case. The only thing I could imagine is to really group by `focus_behavior`, e. g. truly separate cases for "smart" etc. that would require one more level of nesting or a more complicated condition so that we would get away with the same level (also indentation level in this case) as in the pr at its current state.
So should I group the "smart" stuff together then?

tych0: Sure, however it looks best to you sounds good.

ep12: Okay, done in 8261115fad2335dfd7ddf794afbdf0321aae68db

ep12:
@tych0 I updated the pr description. Are there more to-do's?

tych0:
Nope, looks good to me, thanks!

One note -- I generally am not that worried about PR descriptions, and more interested in git commit messages. I took your PR description and put it in the content of the git commit message, and reworded the title of the commit slightly. Anyway, thanks for the patch!

Issue #2241, opened 2021-02-21
RFC/enhancement: Key sequence mode feature: mode stack to keep outer mode active


ep12:
# Issue description
I want to create a mode for `xournalpp` because I want to control it with one hand. With the other hand I am drawing with a graphics tablet.  
KeyChords are all fine, but one thing is just not possible with them: to return to an outer level:
+ I want to configure qtile in such a way that `mod+x` enters the `xournalpp` mode. In this mode I'd like to have
    ```
    t p: pen
    t e: eraser
    t h: highlighter
    ....
    ```
+ I could easily configure it to stay in the `tools` or `recogniser` level, but not in the main `xournalpp` level
+ A work-around is not possible:
    https://github.com/qtile/qtile/blob/7ec0bdd002ea44417bb2c439aca466f382ce80a5/libqtile/core/manager.py#L385-L391
    the hook is called too early to implement a work-around because `self.ungrab_keys()` cannot be inhibited.
+ I think this should really be a feature instead of a user hack:  
+ Suggestion based on my attempt to create a work-around:
   + Qtile keeps a mode stack (list) with all the modes that have been activated.
   + In `grab_chord()`: `chord` is pushed onto the stack
   + In `ungrab_chord()`: the last value is popped off the stack, the new last value `stack[-1]` is used to find a higher-level chord that is then restored using `grab_chord()`
        ```py
	 def ungrab_chord(self) -> None:
		    self.current_chord = False
		    hook.fire("leave_chord")
		    self.ungrab_keys()
		    while self.chord_stack:
		        chord = self.chord_stack.pop()
		        if chord.mode != "":
		            self.grab_chord(chord)
		            break
		    else:
		        for key in self.config.keys:
		            self.grab_key(key)
        ```
        Note that this is just a guess, I haven't tested this.

I hope this short text was enough to explain the idea and I look forward to reading your comments.

# Qtile version
newest: 7ec0bdd002ea44417bb2c439aca466f382ce80a5

# Configuration
```py
...
keys = [
	...
	KeyChord([mod], 'x', [
		KeyChord([], 't', [
			Key([], 'p', send_key('Control+Shift+P')),  # pen
			Key([], 'e', send_key('Control+Shift+E')),  # eraser
			Key([], 'h', send_key('Control+Shift+H')),  # highlighter
			...
		]),  # xournalpp: tools
		KeyChord([], 'r', [
			Key([], 's', send_key('Control+1')),  # shape recogniser
[...]

Issue #1661, opened 2020-04-22
Feature: Simplify the creation of keybindings with custom conditions (i. e. per program keybindings)


ep12:
Qtile makes it relatively hard to have key combinations trigger different action depending on the active window (or the qtile state in general).
It would be nice to make simple filters like so:
```py
import re
from libqtile.lazy import lazy
from libqtile.config_utils import program_filter
from libqtile.... import Key

PROG_FF = program_filter(title=re.compile('.* - Mozilla Firefox))

keys = [
    ...
    Key([mod], 'x',
        lazy.layout.shuffle().when(filter_func=PROG_FF),
        ...
    ),
    ...
]
```
# Details
+ `when` should take a `Callable[[libqtile.manager.Qtile], bool]` as an argument (`filter_func`)
+ `program_filter` is a helper function that generates such a filter func
+ this approach should be modular: instead of adding a `program_title_re` argument to when, it is more pythonic to pass a custom callable. Why? Because then the user has the choice, we don't limit users just to check the window titles. We avoid that the user has to implement similar features using `lazy.function` that do basically the same thing...

# Possible problems
+ What if an action should only be triggered when all the other actions weren't triggered? (-> python is missing a switch-case/match statement / functionality, really) That is a bit annoying... Maybe a magic argument `default_action=True`???
+ What about the description argument to `Key` et al? Should it be `desc: Union[str, List[str]]`?

Opinions? More examples and ideas for features that this should have? More state variables that might be checked in `filter_func`s regularly (-> more helper functions)?

airwoodix:
What about leaving ``when`` as a simple layout filter and calling this functionality ``dispatch``?

One could then recycle keybindings depending on context, something like:

```python
Key([mod], 'x', lazy.dispatch({
     program_filter("firefox"): cmd1,
     custom_filter: cmd2,
     # ...
    })
```
with the keys being ``Callable[['Qtile'], bool]`` as you proposed.

Discussion:

  - didn't think too much about the implementation details. Not sure how non-trivial this is but looks feasible as a wrapper around ``function``;

  - iteration order might be crucial. For Py3.5 users, passing an ``OrderedDict`` is a bit verbose but would work. IIRC, since Py3.6 (effectively) and Py3.7 (guaranteed), the insertion order is preserved for dicts;

  - it is trivial to define a default action;

[...]

PR #2264, opened 2021-02-26
Implement chord_stack

ep12:
Fixes: #2241.
I started to hack on #2241.
# Rationale:
+ emacs-like shortcut sequences have been requested: #1208, #470 (opened 2014, closed in 2020. Other projects would've closed this earlier and then forgot about it, so a huge thanks at this point)
+ they are available since #1777 (mid-2020), but probably a niche application. Let's see how this develops.
+ people who wanted this feature (including me) probably gave up and maybe switched to another WM with this functionality, [e.g. this comment](https://github.com/qtile/qtile/issues/1208#issuecomment-455323244). I was really glad when I came across this feature and tried to use it.
+ The functionality this PR provides cannot be achieved with user extensions, as stated in #2241. The hooks are not called in the right places to create a work-around.
+ The pr was developed and tested extensively with [my config](https://github.com/ep12/qtileconf) (especially `xournal_bindings.py`). I was able to do everything I wanted. Before that I filtered the actions by program, see #1661. With the `xournal` chord/mode I don't need all that complicated stuff.
+ Combining normal key bindings with custom modes is really quite ideal: the important stuff can be controlled with one keystroke, special stuff is in a separate mode. 
+ The performance impact of this feature enhancement is ~0.
+ The config breakage impact of this feature is minimal: having `lazy.ungrab_chord()` bound to a key inside two nested `KeyChord`s with `mode` set behaves now differently; a suitable replacement (`lazy.ungrab_all_chords()`) is provided.

# This pr:
+ [x] implements a `chord_stack` onto which the `KeyChord` objects are pushed when a chord mode is entered.
+ [x] when a chord mode is left (`cmd_ungrab_chord`), the next chord that has `mode != ""` is enabled using `grab_chord`
+ [x] if no chord with `mode != ""` can be found, the root key bindings are restored
+ [x] fixes a typo in `libqtile.config.KeyChord`
+ [x] fixes the `KeyChord.__init__(..., submappings, ...)` type annotation such that `qtile check` does not fail anymore when nesting `KeyChord`s
+ [x] fixes all type annotations where `Key` should be replaced with `Union[Key, KeyChord]`
+ [x] adds tests to check if everything is working correctly
+ [x] adds an example + additional text to the documentation
+ [x] removes `current_chord` as it is not required anymore.

### DIFF
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -118,8 +118,33 @@ Chords can also be chained to make even longer sequences.
         ])
     ]
 
-Modes can also be added to chains if required.
+Modes can also be added to chains if required. The following example
+demonstrates the behaviour when using the ``mode`` argument in chains:
 
+::
+
+    from libqtile.config import Key, KeyChord
+
+    keys = [
+        KeyChord([mod], "z", [
+            KeyChord([], "y", [
+                KeyChord([], "x", [
+                    Key([], "c", lazy.spawn("xterm"))
+                ], mode="inner")
+            ])
+        ], mode="outer")
+    ]
+
+After pressing Mod+z y x c, the "inner" mode will remain active. When pressing
+<escape>, the "inner" mode is exited. Since the mode in between does not have
+``mode`` set, it is also left. Arriving at the "outer" mode (which has this
+argument set) stops the "leave" action and "outer" now becomes the active mode.
+
+.. note::
+    If you want to bind a custom key to leave the current mode (e.g. Control +
+    G in addition to ``<escape>``), you can specify ``lazy.ungrab_chord()``
+    as the key action. To leave all modes and return to the root bindings, use
+    ``lazy.ungrab_all_chords()``.
 
 Modifiers
 =========
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -24,12 +24,17 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+#
+#
+# required for lazy type annotations
+# can be removed when python 3.7...3.9 support is dropped (see PEP 563)
+from __future__ import annotations
 
 import contextlib
 import os.path
 import sys
 import warnings
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Union
 
 import xcffib.xproto
 
@@ -76,12 +81,14 @@ class KeyChord:
     key:
         A key specification, e.g. "a", "Tab", "Return", "space".
     submappings:
-        A list of Key declarations to bind in this chord
+        A list of Key or KeyChord declarations to bind in this chord.
     mode:
-        A string with vim like mode name if it's set chord not end
-        after use one of submapings or Esc key
+        A string with vim like mode name. If it's set, the chord mode will
+        not be left after a keystroke (except for Esc which always leaves the
+        current chord/mode).
     """
-    def __init__(self, modifiers: List[str], key: str, submapings: List[Key], mode: str = ""):
+    def __init__(self, modifiers: List[str], key: str,
+                 submapings: List[Union[Key, KeyChord]], mode: str = ""):
         self.modifiers = modifiers
         self.key = key
 
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -88,14 +88,14 @@ class KeyChord:
         current chord/mode).
     """
     def __init__(self, modifiers: List[str], key: str,
-                 submapings: List[Union[Key, KeyChord]], mode: str = ""):
+                 submappings: List[Union[Key, KeyChord]], mode: str = ""):
         self.modifiers = modifiers
         self.key = key
 
         def noop(qtile):
             pass
-        submapings.append(Key([], "Escape", lazy.function(noop)))
-        self.submapings = submapings
+        submappings.append(Key([], "Escape", lazy.function(noop)))
+        self.submappings = submappings
         self.mode = mode
 
     def __repr__(self):
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -85,7 +85,7 @@ class Qtile(CommandObject):
         self.dgroups: Optional[DGroups] = None
 
         self.keys_map: Dict[Tuple[int, int], Key] = {}
-        self.current_chord = False
+        self.chord_stack = []
         self.numlock_mask, self.valid_mask = self.core.masks
 
         self.current_screen: Optional[Screen] = None
@@ -366,10 +366,9 @@ class Qtile(CommandObject):
                     )
                     if status in (interface.ERROR, interface.EXCEPTION):
                         logger.error("KB command error %s: %s" % (cmd.name, val))
-            else:
-                if self.current_chord is True or (self.current_chord and key.key == "Escape"):
-                    self.cmd_ungrab_chord()
-                return
+            if self.chord_stack and (self.chord_stack[-1].mode == "" or key.key == "Escape"):
+                self.cmd_ungrab_chord()
+            return
 
     def grab_keys(self) -> None:
         """Re-grab all of the keys configured in the key map
@@ -396,19 +395,40 @@ class Qtile(CommandObject):
         self.keys_map.clear()
 
     def grab_chord(self, chord) -> None:
-        self.current_chord = chord.mode if chord.mode != "" else True
-        if self.current_chord:
-            hook.fire("enter_chord", self.current_chord)
+        self.chord_stack.append(chord)
+        if self.chord_stack:
+            hook.fire("enter_chord", self.chord_stack[-1].mode)
 
         self.ungrab_keys()
         for key in chord.submapings:
             self.grab_key(key)
 
     def cmd_ungrab_chord(self) -> None:
-        self.current_chord = False
+        """Leave a chord mode"""
         hook.fire("leave_chord")
 
         self.ungrab_keys()
+        if not self.chord_stack:
+            logger.debug("cmd_ungrab_chord was called when no chord mode was active")
+            return
+        # The first pop is necessary: Otherwise we would be stuck in a mode;
+        # we could not leave it: the code below would re-enter the old mode.
+        self.chord_stack.pop()
+        # Find another named mode or load the root keybindings:
+        while self.chord_stack:
+            chord = self.chord_stack.pop()
+            if chord.mode != "":
+                self.grab_chord(chord)
+                break
+        else:
+            for key in self.config.keys:
+                self.grab_key(key)
+
+    def cmd_ungrab_all_chords(self) -> None:
+        """Leave all chord modes and grab the root bindings"""
+        hook.fire("leave_chord")
+        self.ungrab_keys()
+        self.chord_stack.clear()
         for key in self.config.keys:
             self.grab_key(key)
 
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -400,7 +400,7 @@ class Qtile(CommandObject):
             hook.fire("enter_chord", self.chord_stack[-1].mode)
 
         self.ungrab_keys()
-        for key in chord.submapings:
+        for key in chord.submappings:
             self.grab_key(key)
 
     def cmd_ungrab_chord(self) -> None:
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -30,7 +30,7 @@ import subprocess
 import tempfile
 import time
 import warnings
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Union
 
 import xcffib
 import xcffib.xinerama
@@ -84,8 +84,8 @@ class Qtile(CommandObject):
         self.groups: List[_Group] = []
         self.dgroups: Optional[DGroups] = None
 
-        self.keys_map: Dict[Tuple[int, int], Key] = {}
-        self.chord_stack = []
+        self.keys_map: Dict[Tuple[int, int], Union[Key, KeyChord]] = {}
+        self.chord_stack: List[KeyChord] = []
         self.numlock_mask, self.valid_mask = self.core.masks
 
         self.current_screen: Optional[Screen] = None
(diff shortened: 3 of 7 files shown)


ep12:
I had no trouble running the branch in production, so I think this pr is now ready for comments. It should be easier to play around with it than having a look at the theoretical discussion in #2241. Maybe somebody finds it useful...

libqtile/core/manager.py
```diff
@@ -142,6 +143,7 @@ def noop(qtile):
         self.config.mouse += (Click([], "Button1", lazy.function(noop), focus="after"),)
 
     def dump_state(self, buf):
+        del(self.chord_stack)  # TODO: is there a better way to avoid errors?
```
m-col, line 146: What is this for? Could you elaborate on what error you get and need to avoid?

libqtile/core/manager.py
```diff
@@ -386,10 +386,18 @@ def grab_chord(self, chord) -> None:
...
         self.current_chord = False
         hook.fire("leave_chord")
 
         self.ungrab_keys()
+        if self.chord_stack:
+            # The first pop is necessary: Otherwise we would be stuck in a mode;
+            # we could not leave it: the code below would re-enter the old mode.
+            self.chord_stack.pop()
+        else:
+            logger.debug("cmd_ungrab_chord was called when no chord mode was active")
```
m-col: I think it's clear enough why it'd be needed -- you're just removing the old (current) mode, right? If so then you'd be fine with just:
```py
if not self.chord_stack:
    logger.debug("cmd_ungrab_chord was called when no chord mode was active")
    return
self.chord_stack.pop()
```
You also don't return following the debug log, and continue to `grab_key` the config keys. Is this necessary? Wouldn't the config keys already be grabbed at this point always?

I think it'd be nice if you could merge these 2 commits (the first two, implementation and bug fix for it).

libqtile/config.py
```diff
@@ -73,12 +75,14 @@ class KeyChord:
...
     mode:
-        A string with vim like mode name if it's set chord not end
-        after use one of submapings or Esc key
+        A string with vim like mode name. If it's set, the chord mode will
+        not be left after a keystroke (except for Esc which always leaves the
+        current chord/mode).
     """
-    def __init__(self, modifiers: List[str], key: str, submapings: List[Key], mode: str = ""):
+    def __init__(self, modifiers: List[str], key: str,
+                 submapings: List[Union[Key, KeyChord]], mode: str = ""):
```
m-col: Forgive me as I am not very knowledgable about Python's type checking (something I avoided for the longest time...). I read through PEP 563 but it still isn't clear to me why it is needed for this use case. Why wouldn't the change you have made to this annotation work without the added futures import?

An unrelated comment -- what's up with the `submapings` spelling? I assume this is an error that predates this PR. Should we fix that?

m-col: And as I move on to the next commit I see you have fixed the spelling! Ignore that comment :D

libqtile/config.py
```diff
@@ -82,14 +82,14 @@ class KeyChord:
         current chord/mode).
     """
     def __init__(self, modifiers: List[str], key: str,
-                 submapings: List[Union[Key, KeyChord]], mode: str = ""):
+                 submappings: List[Union[Key, KeyChord]], mode: str = ""):
```
m-col, line 88: When was the spelling error introduced? As this is part of the config API, unless it was introduced super recently we should probably add a changelog and `qtile migrate` entry for this change.

libqtile/core/manager.py
```diff
@@ -407,7 +407,7 @@ def cmd_ungrab_chord(self) -> None:
             for key in self.config.keys:
                 self.grab_key(key)
 
-    def cmd_leave_chord_mode(self) -> None:
+    def cmd_ungrab_all_chords(self) -> None:
```
m-col, line 410: Please could you merge this with the commit that added this function?

docs/manual/config/keys.rst
```diff
@@ -118,8 +118,33 @@ Chords can also be chained to make even longer sequences.
...
+            KeyChord([], "y", [
+                KeyChord([], "x", [
+                    Key([], "c", lazy.spawn("xterm"))
+                ], mode="inner")
+            ])
+        ], mode="outer")
+    ]
+
+After pressing Mod+z y x c, the "inner" mode will remain active. When pressing
+<escape>, the "inner" mode is left. Since the mode in between does not have
```
m-col, line 139: Could you replace "left" with "exited" or something similar? I actually got confused because I read "left" as "remains" lol

m-col:
This is awesome, thanks! I really appreciate the small and simple commits, makes reviewing a breeze :) Overall I think the feature and approach is super nice but I've left some comments and questions.

Does Qtile's chord_stack` attribute make the `current_chord` attribute redundant?

libqtile/core/manager.py
ep12: When qtile is restarted, the current state is dumped. But the `KeyChord` on the stack do not support serialisation such that restoring the state would actually work. And, in fact, if the user changes keybindings in the config and then restarts qtile, he/she expects to be in the root mode (not in the chord mode they might have been in previously - that would be unintuitive, complicated to implement correctly and not an important feature at all). Additionally, `current_chord` is an internal shortcut to check if we stay in a mode or if we need to call `cmd_ungrab_chord()`. Persisting one of `chord_stack` and `current_chord` would result in inconsistencies. See also: [my comment here](https://github.com/qtile/qtile/issues/2241#issuecomment-786790819).
Actually, you made me think about it so that I might want to revisit the state topic: If `current_chord` is dumped/restored, I might need to un-persist that one too.

libqtile/core/manager.py
ep12: Okay, didn't think about that. I should definitely rewrite that.
Regarding the commits: Isn't that called squashing or something? I need to find out how that works, still relatively new to using git. I'd also like to rename the branch since the `persistent` idea was dropped...
One general question: Should prs be always be just a single commit? I came across some merged prs from tych0, ramnes, you etc. so is there a silent agreement that prs should be one commit only?

libqtile/config.py
ep12: So before PEP 563 or whatever, I mentioned in the commit message, using a class inside itself as a type annotation was only possible with a really ugly hack:
```py
class MyClass:
    pass

class MyClass:
    ...
    def myFunc(self, other: MyClass):
         ...
```
because while defining `MyClass`, the `MyClass` doesn't yet exist. With that PEP / future import we can enable lazy type annotation evaluation so that we don't get a `NameError` or what this thing actually throws without that import.
From 3.10 onwards this behaviour will be default, so the import can be removing when python 3.9 support will be eventually dropped in the distant future...

libqtile/config.py
ep12: Didn't think about that, it might be a good idea. Although I doubt that the literally like 3 people that already started using `KeyChord`s since they became available recently instantiate the `KeyChord` objects with `submapings=[...]` explicitly.
But your point is completely fine, I have to find out how that `migrate` thing works under the bonnet.

ep12:
> This is awesome, thanks! I really appreciate the small and simple commits, makes reviewing a breeze :) Overall I think the feature and approach is super nice but I've left some comments and questions.
> 
> Does Qtile's chord_stack`attribute make the`current_chord` attribute redundant?

Actually, it could probably. But then we might need to check for `self.chord_stack and self.chord_stack[-1].mode != ""` which is less readable than `self.current_chord is False`. It would also naturally resolve [my comment](https://github.com/qtile/qtile/pull/2264#discussion_r588906667) above. I'd love to get more opinions regarding that!

libqtile/config.py
ep12: Behaviour in python 3.9.2:
```py
~ via 🐍 v3.9.2 took 16s 
❯ python
>>> class MyClass:
...     ...
...     def myFunc(self, other: MyClass):
...          ...
... 
  File "<stdin>", line 3, in MyClass
NameError: name 'MyClass' is not defined
>>> ^D
~ via 🐍 v3.9.2 took 37s 
❯ python
>>> from __future__ import annotations  # with this import everything works fine
>>> class MyClass:
...     ...
...     def myFunc(self, other: MyClass):
...          ...
... 
>>> 
```

libqtile/core/manager.py
m-col: Do we need to store either? Currently neither is dumped, and I wouldn't expect either to persist across restarts.

ep12: Oh, I see. This line is really unnecessary, right? :rofl:

libqtile/core/manager.py
m-col: Although PRs are often a single commit, there are also plenty where multiple commits make more sense. Each commit should represent one logical change to the tree and (ideally) the tree would make sense at each point. The way you've committed everything is actually super nice and friendly, and by having more small commits where each one is one change (like you've done) it makes the PR very easy to follow -- so thanks for that!

The git-related changes that I have asked for are simply due to you changing something that you had already changed in a previous commit (e.g. adding `cmd_leave_chord_mode` then in a later commit renaming it). This results in the upstream git history remaining more clean.

You can merge and reorder commits using `git rebase --interactive`, this article might be useful: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html. It doesn't talk about re-ordering but if you need to do that you can just rearrange the lines in the interactive git file that it opens in an editor.

libqtile/config.py
m-col: OK, awesome, thanks a lot for the explanation! All good :)

libqtile/core/manager.py
m-col: I believe so :)

ep12:
Okay, I rewrote the history, made the commit messages a bit more verbose and squashed a few commits. Nearly expect this thing to break now :smile: :cry:

libqtile/config.py
ep12: The commit that fixes this typo is the very last making it easy to `commit --amend` if the migrate thing is decided to be worth it. Leaving this unresolved for further opinions.

m-col:
Brilliant thanks! I'll leave it for another pair of eyes just in case I missed something, but it looks good to me as is :)

libqtile/core/manager.py
ep12: @m-col Okay, thanks a lot for the explanation. I guess you'll like the history rewrite then :smile:
I played through https://learngitbranching.js.org/ so I've heard of the powers of `rebase -i`. But with great power... So naturally I was a bit scared at first (making backup branches just to be sure). But some tests have passed, maybe I'll loose my fear. Marking this as resolved.

libqtile/core/manager.py
ep12: Removed that dummy line, too...

ep12:
Thank you for your feedback, thoughts and explanations @m-col! I am looking forward to reading more comments 😁.

tcld:
This looks great, I didn't even find a single spot to put a comment on. :)

I hope I find some time to play around with this; I tried my hands at chords a few months back but for me it felt too cumbersome for daily use. It really is time for another go.

ep12:
Thanks for your feedback, @tcld! Glad to hear that!  
What do you think about replacing `current_chord`? What are your thoughts regarding the `migrate` utility and the typo fix?

tcld:
About the typo: I don't think a migration is necessary, but it might be an opportunity to learn how to use the new tool. In my opinion it is up to you to decide, since this is hardly going to affect anybody.

About `current_chord`: It looks redundant in the code, so I think removing it might be the better choice. You could implement a read-only property, if you really wanted to preserve the "shortcut". If you went that route, maybe it would be good to rename the property, too. `current_chord` doesn't seem to describe it too well.

ep12:
Okay, thanks for your feedback!
Regarding the `migrate` utility: I think it would be completely analogous to https://github.com/qtile/qtile/blob/b4dd1cbbe47f91510fdad3ad2c7159a668f01e6f/libqtile/scripts/migrate.py#L46-L51
so rather trivial and for a good style I'd have to write a test. I've decided that this is not worth the effort, especially since this would be removed soon, I guess.

I do think that `current_chord` should be removed, so I'll do that now...

ep12:
@tcld @m-col One more question regarding that change:
Currently, the `enter_chord` hook is fired with `True` if `self.chord_stack[-1].mode == ""`. Do you think I can break that and just pass  `self.chord_stack[-1].mode`? That way I could avoid an if statement that isn't worth anything IMO.

tcld:
I don't see any value in passing a constant `True` back into the `enter_chord` hook, so I agree with your change. Reacting to the mode name definitely opens up some possibilities for the user.

Yes, you should probably mention that change in the changelog, even though I think it would be surprising if anybody found a use for the passed "boolean".^^

ep12:
> I don't see any value in passing a constant `True` back into the `enter_chord` hook, so I agree with your change. Reacting to the mode name definitely opens up some possibilities for the user.
> 
> Yes, you should probably mention that change in the changelog, even though I think it would be surprising if anybody found a use for the passed "boolean".^^
> 
> Two more things:
> 
> 1. Are the stacked chords included in the images generated by our `gen-keybinding-img`? (Just curious, this PR would be valuable either way.)
> 2. Are you using chords heavily? I would love to see a configuration that uses the chords in a creative manner. I lack such creativity, but it feels like there would be near infinite use cases for the chords (especially with this PR).
> 
> Oh, and another thing:
> Would it maybe make sense to pass back the "ungrabbed" chord with the `leave_chord`-hook, too?



Okay, a lot of things to react to:
+ yep, `chord_stack[-1].mode` is definitely simpler than `chord_stack[-1].mode if chord_stack[-1].mode != "" else True` which previously would've caused strange-looking hooks: `if isinstance(arg, str) and ...: ...` and `if isinstance(arg, bool) and arg: ...`. The new way seems a lot more consistent...
+ okay, I'll add an entry to the changelog then
+ See #2247: This a different beast. This pr should probably be merged before working on that problem.
+ [No, not really (yet)](https://github.com/ep12/qtileconf). But as stated in the rationale, this opens up a lot of possibilities! So far I only have the xournalpp mode (and unnamed sub-modes). Maybe I'll do more fancy stuff if I've time to play...
+ That would break the existing hook, I'd rather treat that separately as well. It might be useful to know which chord is popped, but then I don't even use the chord hooks after all. Open up an issue if you like.

ep12:
And I just stumbled across [this distrotube video](https://www.youtube.com/watch?v=9NO2DJARbU0&t=640s). Nice to see that he started using them, too.

ep12:
Furthermore, there seem to be a lot of places where the typing annotations are incomplete, e.g. they say `Key` and not `Union[Key, KeyChord]`. I have to add a to-do for that.

tych0:
Can you add your analysis from the original PR message to the commit log somewhere? Looks like the main commit is missing a bunch of it.

tych0:
Code itself looks fine to me though, it comes with test :)

ep12:
Okay, I added the rationale to the d92aa40 commit message, squashed two commits (typing annotations A+B, documentation+CHANGELOG) and resolved the merge conflict (CHANGELOG). I really hope I picked all the :cherries: :rofl:.

tych0:
Awesome, that's a great commit message. Thank you!

PR #2269, opened 2021-02-27
change location attr to dict

akkaky:
Location attribute changed from string to dict. This allows the GPS coordinates to be used as the name of the place and to replace the coordinates with the preferred name.
Also returned the json field to `defaults` for correct display in the documentation.

### DIFF
--- a/libqtile/widget/wttr.py
+++ b/libqtile/widget/wttr.py
@@ -47,6 +47,7 @@ class Wttr(GenPollUrl):
             'or build your own custom output format, use the special '
             '%-notation. See https://github.com/chubin/wttr.in#one-line-output'
         ),
+        ('json', False, 'Is Json?'),
         (
             'lang', 'en',
             'Display text language. List of supported languages '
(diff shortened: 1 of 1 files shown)


tych0:
Can you separate these two things into two logical commits, and include your reasoning for the change in the PR notes in the commit message body as well?

akkaky:
Ok! Will make new PR's.

akkaky:
Have I to close this PR?

tych0:
No, you can just force push with `git push -f` to this branch when you've updated the commits. Thanks!

akkaky:
Like this or wrong?

tych0:
That looks good to me, but can you add your reasoning about *why* you changed the list to a dict in that commit? e.g.:

> This allows the GPS coordinates to be used as the name of the place and to replace the coordinates with the preferred name.

akkaky:
Add to commit text? Sorry, I just study work with git.

libqtile/widget/wttr.py
```diff
@@ -5,9 +5,12 @@
 
 class Wttr(GenPollUrl):
     """Display weather widget provided by wttr.in_.
+
```
tych0: It looks like this is a commit full of whitespace only changes; if you want these can you squash them (and the message) into the previous commit?

akkaky:
@tych0 Thanks for the advice. Now I figured out how to make clear commits. Before I made a scrapyard.

tych0:
Sure thing, thanks for the patches!

PR #2270, opened 2021-02-27
Widget finalize fixes

tych0:
See patch notes for details.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,8 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - New optional dependency for dbus related features: dbus-next.
           Replaces previous reliance on dbus/Glib and allows qtile to use async
           dbus calls within asyncio's eventloop.
+        - widget.BatteryIcon no longer has a fallback text mode; use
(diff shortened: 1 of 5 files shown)


m-col:
Looks good.

It might be a bit hacky (and I expect boos) but a simple solution for not drawing widgets with scheduled tasks might be to override `self.draw` with a no-op when finalising. This could be implemented at the level of `_Widget.finalize` and propagate to all widgets, even if they implement a custom `self.draw` method.

tych0:
> It might be a bit hacky (and I expect boos)

boo :)

> This could be implemented at the level of _Widget.finalize and propagate to all widgets, even if they implement a custom self.draw method.

Ideally we wouldn't have to resort to hacks like this; it seems reasonable to want widgets themselves to know when to stop trying to draw. Since the vast majority of widgets are text based, hopefully this will catch most of these kinds of bugs. I'd say let's play whack-a-mole for the remaining ones as we see them, and if we still can't figure it out then we can resort to more hacks.

m-col:
Alrighty, sounds good :)

PR #2271, opened 2021-02-28
tests: make sure logs from qtile processes are exposed

tych0:
This is a combo revert:

Revert "Fix tests hanging due to logging lock on fork"
This reverts commit 57f0963fbd63f1c32b383412bc9bbc093b38dc22.

Revert "Manage Qtile subprocess better"
This reverts commit 622b69428c3bdfad0ca533301f55231c3cdb2061.

The first one seems not to have actually fixed the hang (viz.
https://github.com/qtile/qtile/runs/1986400750 which still has the hang in
pypy).

The second one swallows qtile log output:

on v0.17.0:

diff --git a/libqtile/config.py b/libqtile/config.py
index 847021dc..ff9186da 100644
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -416,6 +416,8 @@ class Screen(CommandObject):

     def cmd_info(self):
         """Returns a dictionary of info for this screen."""
+        from libqtile.log_utils import logger
+        logger.error("screen info() called")
         return dict(
             index=self.index,
             width=self.width,
index cfe95b66..ec718d7d 100644
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -1568,6 +1568,7 @@ def test_cursor_warp(manager):
     root = conn.default_screen.root.wid

     assert manager.c.screen.info()["index"] == 0
+    assert False

     manager.test_window("one")
     manager.c.window.set_position_floating(50, 50)

TOXENV=py38 tox -- test/test_manager.py::test_cursor_warp

shows me:

------------------------------------------ Captured stdout call ------------------------------------------
2021-02-28 07:40:22,686 ERROR libqtile config.py:cmd_info():L420 screen info() called

there's no such logging on current, but I do see:

---------------------------------------- Captured stderr teardown ----------------------------------------
Error in atexit._run_exitfuncs:

----------- coverage: platform linux, python 3.8.6-final-0 -----------
Name                                         Stmts   Miss  Cover   Missing
--------------------------------------------------------------------------

Both of these add a dependency on something that's not in the standard library.
Let's just drop them for now since they don't solve the issue and swallow
useful logs from the tests.


tych0:
This makes the tests noticeably faster on my machine as well.

PR #2277, opened 2021-03-06
Do not try to focus upon _NET_WM_USER_TIME property change event

### LINKED ISSUE #2172: PyCharm popup loses focus on keypress
I've been using qtile as my daily driver for a few weeks without major issues, but yesterday I noticed that the New file popup in PyCharm isn't working properly. When it is open and focused and I press any key, the main window regains focus while the popup is still shown. The following GIF shows the problem:

qtile-pycharm

FWIW, it works just fine in i3, which makes me think that it might be an issue in qtile.

# Qtile version

0.16.2.dev200+ge6daea51

# Stack traces

The xprop output of the popup window is as follows:

```
_NET_WM_DESKTOP(CARDINAL) = 0
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_USER_TIME(CARDINAL) = 148265765
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG
_MOTIF_DRAG_RECEIVER_INFO(_MOTIF_DRAG_RECEIVER_INFO) = 0x6c, 0x0, 0x5, 0x0, 0x3a, 0xa, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x10, 0x0, 0x0, 0x0
XdndAware(ATOM) = BITMAP
_NET_WM_STATE(ATOM) = _NET_WM_STATE_SKIP_TASKBAR
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
		window id # of group leader: 0x100004c
WM_TRANSIENT_FOR(WINDOW): window id # 0x100004c
_NET_WM_PID(CARDINAL) = 192251
WM_PROTOCOLS(ATOM): protocols  
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x0, 0x0, 0x0, 0x0
		program specified location: 1722, 797
		program specified size: 392 by 54
		window gravity: NorthWest
WM_CLASS(STRING) = "jetbrains-pycharm", "jetbrains-pycharm"
WM_CLIENT_LEADER(WINDOW): window id # 0x1000008
_NET_WM_NAME(UTF8_STRING) = "win90"
WM_NAME(STRING) = "win90"
```

# Configuration

https://github.com/mbenford/dotfiles/blob/master/.config/qtile/config.py

m-col:
When this property is changed on a window and `follow_mouse_focus` is
False, that window is focussed by Qtile. However, this behaviour is not
described in the specification and causes unexpected side effects.
Fixes #2172

### DIFF
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -1230,10 +1230,6 @@ class Window(_Window):
             # are set when the property is emitted
             # self.update_state()
             self.update_state()
-        elif name == "_NET_WM_USER_TIME":
-            if not self.qtile.config.follow_mouse_focus and \
-                    self.group.current_window != self:
-                self.group.focus(self, False)
         else:
             logger.info("Unknown window property: %s", name)
         return False


m-col:
I want to run this for a few days to confirm that there are no unexpected side effects of removing it. I'll update then with how that goes.

tych0:
+1, if this fixes the bug it seems very reasonable to me.

ramnes:
cc @Roger

PR #2278, opened 2021-03-06
Widget update fixes

tych0:
Here's some more widget update fixes, which should ultimately result in fewer bar draws. As a bonus, the code is simplified, and all _TextBox based widgets get truncation functionality.


libqtile/widget/currentscreen.py
```diff
@@ -45,15 +45,11 @@ def _configure(self, qtile, bar):
...
-            self.text = self.inactive_text
-            self.foreground = self.inactive_color
+            if self.qtile.current_screen == self.bar.screen:
+                self.foreground = self.active_color
+                self.update(self.active_text)
+            else:
+                self.foreground = self.inactive_color
+                self.update(self.inactive_text)
+
+        hook.subscribe.current_screen_change(self.hook_response)
```
m-col, line 55: Is there any reason `hook_response` is dynamically constructed within `setup_hooks` and not a class method? Looks like you hooked `self.hook_response` as if it was.

tych0: Oh, I think I had it that way before and didn't fix this. I can make it a class method, thanks for catching this.

test/widgets/test_window_count.py
```diff
@@ -1,70 +1,38 @@
...
-    fakebar = Bar([wc], 24)
-    wc.bar = fakebar
-    qtile = FakeQtile()
-    wc.qtile = qtile
-    wc._setup_hooks()
+    manager_nospawn.start(WindowCountConf)
 
     # No windows opened
-    assert wc._count == 0
+    assert int(manager_nospawn.c.widget["windowcount"].get()) == 0
```
m-col, line 17: do these need the `int`s?

m-col (approved):
Nice, looks good, especially this: 
image

test/widgets/test_window_count.py
tych0: I think so; the return value is a string.

m-col: Ah yep, of course.

m-col:
Cool cool, LGTM!

ramnes:
Excellent, thank you!

PR #2284, opened 2021-03-10
Fix #2282 : displaying bar with margin properly after hide/show toggle.

### LINKED ISSUE #2282: Gaps on bar toggle
When you use a keybind to toggle a bar qtile does not correctly put the gaps in when the bar returns to the screen. 

I noticed that it seems to have issue with existing windows on the screen so the gaps worked fine around the top and side but on the bottom underneath where I had the bar return, my gap settings were ignored.

I guess that maybe the existing windows on the screen need to be redrawn to take in the bar again and have the correct gaps

Prior to toggle
21-03-09_08-12_366x94

After toggling the bar
21-03-09_08-12_358x113

The only way to get it back to normal is to restart qtile

Fabien-Bernard:
Modification of libqtile/bar.py to save and restore the calculated size of a bar during toggle.

Modification of test/test_bar.py by adding 1 test to check the size of a bar :
- before hidding,
- after hidding,
- and after unhidding.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -164,6 +164,7 @@ class Bar(Gap, configurable.Configurable):
         self.saved_focus = None
         self.cursor_in = None
         self.window = None
+        self.size_calculated = 0
 
         self.queued_draws = 0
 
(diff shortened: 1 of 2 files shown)


tych0:
Thanks!

LamprosPitsillos:
I still have my margin ignored after unhiding... I'm asking around before i open an issue.
Do i have to do something specific for my margin to be persistent?

tych0:
This patch is not in any released version of qtile; I'd check that you've got it as a first step.

LamprosPitsillos:
Oh you're right! I'll start using the git version probably

LamprosPitsillos:
Wasn't this in the 0.18.0 update? I still have this issue sadly.

Fabien-Bernard:
Yes, it's part of 0.18.0 version.
May you share your config?

PR #2286, opened 2021-03-10
Revert "Fix and remove unused code in KeyboardLayout widget"

### LINKED ISSUE #2265: KeyboardLayout widget uses obscure dependency
xkb-switch is not necessarily mainstream, although it was used in the fix for #2027 in #2042. It would be nice to fix the bug using setxkbmap instead (or perhaps just revert #2042 if enough people don't like this change).

/cc @tcld @k4leg

tych0:
This reverts commit 089103720c9bbb2d81a737b116e82f0c53c4ae33.

We have two votes for reverting this and keeping the setxkbmap-based
version, and no votes for keeping this. Let's revert it.

Fixes #2265

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,8 +21,6 @@ Qtile 0.x.x, released xxxx-xx-xx:
           A new command, `lazy.ungrab_all_chords`, was introduced to return to the root bindings.
           The `enter_chord` hook is now always called with a string argument.
           The third argument to `KeyChord` was renamed from `submaping` to `submapping` (typo fix).
-    * bugfixes
-        - fix KeyboardLayout widget (xkb-switch is new dependency)
 
 Qtile 0.17.0, released 2021-02-13:
     !!! Python version breakage !!!
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -22,269 +22,116 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import asyncio
+import re
 from subprocess import CalledProcessError
-from typing import NoReturn, Union
 
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
-
-class NoDependencyFoundError(Exception):
-    pass
+kb_layout_regex = re.compile(r'layout:\s+(?P<layout>\w+)')
+kb_variant_regex = re.compile(r'variant:\s+(?P<variant>\w+)')
 
 
 class KeyboardLayout(base.InLoopPollText):
-    """
-    Widget for changing and displaying the current keyboard layout group
+    """Widget for changing and displaying the current keyboard layout
 
-    It requires xkb-switch to be available in the system.  setxkbmap is
-    optional dependency (to configure the keyboard locally).
-    """
+    To use this widget effectively you need to specify keyboard layouts you want to use (using "configured_keyboards")
+    and bind function "next_keyboard" to specific keys in order to change layouts.
 
+    For example:
+
+        Key([mod], "space", lazy.widget["keyboardlayout"].next_keyboard(), desc="Next keyboard layout."),
+
+    It requires setxkbmap to be available in the system.
+    """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        (
-            'compact',
-            None,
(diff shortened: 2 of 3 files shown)


CHANGELOG
```diff
@@ -11,8 +11,6 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - memory widget can now be displayed with decimal values
         - new "qtile migrate" command, which will attempt to upgrade previous
           configs to the current version in the case of qtile API breaks.
-    * bugfixes
-        - fix KeyboardLayout widget (xkb-switch is new dependency)
```
ramnes, line 15: we probably want to keep this

ramnes:
I have to admit that I wasn't confortable with the original PR; I should have expressed my thoughts earlier, sorry. Reverting it is rough though; @k4leg wouldn't have spent time working on this if it didn't resolve a problem they had with the previous implementation.

I seem to remember that @k4leg originally implemented support for both setxkbmap and xkb-switch and you asked to change it, but maybe that would make everyone happy and is the good way forward?

Another option could be to bundle xkb-switch, either here or in another Python package, especially since it doesn't seem to have any dependency other than X.

zordsdavini:
For me new version didn't worked completed. So, I wrote simple GenPollText to show current layout and moved all setting to .xsession with setxkbmap. With old widget there is problems with options, especially toggle the keyboard. That's why I needed to add
```
    EzKey(
        "A-S-<space>",
        lazy.widget['keyboardlayout'].next_keyboard(),
        desc='switch keyboard layout'),
```
My suggestion would be to have both versions if new version is working for somebody

tych0:
> I seem to remember that @k4leg originally implemented support for both setxkbmap and xkb-switch and you asked to change it, but maybe that would make everyone happy and is the good way forward?

That wasn't my understanding of the code, but I'm happy to be corrected :)

> My suggestion would be to have both versions if new version is working for somebody

No. We can barely get people to take time to review and maintain one keyboard layout widget. Adding another will not fix this problem.

tych0:
So is there any conclusion on this? Seems like basically nobody wants to work on it, so perhaps we should just delete it from the tree if we can't agree on what an in-tree implementation would look like.

ep12:
> So is there any conclusion on this? Seems like basically nobody wants to work on it, so perhaps we should just delete it from the tree if we can't agree on what an in-tree implementation would look like.

I had a look at the original issue #2027 and I could reproduce that `setxkbmap` didn't print the current layout correctly. So there definitely is an issue, although quite a niche one it seems. Since @m-col pointed us to the [setxkbmap](https://github.com/qtile/qtile/issues/2265#issuecomment-786931988) source, it should be possible to write a c/cython extension that interfaces with the xkb extension until `xcffib` supports that.
I also had a look at the pr (#2042, not back then, but now) and found that it removes a `cmd_next_keyboard` which might break the config besides the new dependency `xkb-switch` that prevents qtile from starting (at least it did when I restarted qtile after a checkout back then, I am using the widget). Furthermore, it actually uses both `setxkbmap` and `xkb-switch`, where `setxkbmap` seems to be optional.

To be honest, I am not too actively checking the prs so that's why I didn't comment on that one. Sorry to be so harsh, but IMO this should be fixed in a cleaner way than #2042 did. Maybe this commit should only be reverted when a better option (pr) is available that fixes #2027. I'm kind of undecided on that; It fixes @k4leg's issue but it probably causes minor annoyances for a lot of other people. I'd definitely recommend to open #2027 again hoping that somebody comes up with a better solution.

MatejaMaric:
I think #2042 should be reverted and #2027 reopened. 
Today I updated Qtile and KeyboardLayout widget doesn't work properly (or is not backwards compatible) even if I install `xkb-switch`. 
I am kind of busy for the next few days but after that I can try fixing #2027. I would also use `xkb-switch` but I would make it an optional dependence.

tych0:
@ramnes thoughts? IMO, we should revert, given that several people have complained and there's no movement on fixing things in a way that is more palatable.

CHANGELOG
tych0: What do you mean by this? If we drop the fix, shouldn't we drop the changelog note about it?

ramnes: Oh sorry, I thought it was released already. Makes much more sense to revert then.

ramnes:
I thought it was released already and it's not, so let's just revert and see if someone comes with an alternative fix. 👍

PR #2288, opened 2021-03-12
core: better logging when spawn() fails

tych0:
If someone sends us an invalid thing to run, right now we just silently
swallow it, because the logging module doesn't work after we've forked.

Instead, let's just keep around what we're going to exec (being smart about
shell=True invocations so we don't just do a lookup of "/bin/sh") and do a
path lookup on it. If it doesn't exist, we can fail without doing all the
forking and give the user a real error message.

While arguably not logging anything is fine (someone typed an invalid
thing), I've seen a lot of confusion more recently about spawn not working
because of typos and such. Hopefully this will fix things.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -24,6 +24,7 @@ import logging
 import os
 import pickle
 import shlex
+import shutil
 import signal
 import subprocess
 import tempfile
@@ -1138,14 +1139,19 @@ class Qtile(CommandObject):
 
             spawn(["xterm", "-T", "Temporary terminal"])
         """
-        if shell:
-            if not isinstance(cmd, str):
-                cmd = subprocess.list2cmdline(cmd)
-            args = ["/bin/sh", "-c", cmd]
-        elif isinstance(cmd, str):
+        if isinstance(cmd, str):
             args = shlex.split(cmd)
         else:
             args = list(cmd)
+            cmd = subprocess.list2cmdline(args)
+
+        to_lookup = args[0]
+        if shell:
+            args = ["/bin/sh", "-c", cmd]
+
+        if shutil.which(to_lookup) is None:
+            logger.error("couldn't find `{}`".format(to_lookup))
+            return -1
 
         r, w = os.pipe()
         pid = os.fork()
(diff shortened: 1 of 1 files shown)


libqtile/core/manager.py
```diff
@@ -1138,14 +1139,24 @@ def cmd_spawn(self, cmd, shell=False):
 
             spawn(["xterm", "-T", "Temporary terminal"])
         """
+        to_lookup = ""
         if shell:
             if not isinstance(cmd, str):
+                to_lookup = cmd[0]
                 cmd = subprocess.list2cmdline(cmd)
+            else:
+                to_lookup = cmd
```
ramnes, line 1148:
```suggestion
                to_lookup = cmd.split(" ", 1)[0]
```

libqtile/core/manager.py
```diff
@@ -1138,14 +1139,24 @@ def cmd_spawn(self, cmd, shell=False):
...
             args = ["/bin/sh", "-c", cmd]
         elif isinstance(cmd, str):
             args = shlex.split(cmd)
+            to_lookup = args[0]
         else:
             args = list(cmd)
+            to_lookup = args[0]
```
ramnes, line 1155: maybe worth refactoring to look for the type of `cmd` before checking `shell`

libqtile/core/manager.py
tych0: Really it should be shlex.split(cmd)[0], right?

libqtile/core/manager.py
tych0: Yeah, that might be a bit better, I'll take a look.

libqtile/core/manager.py
ramnes: yeah preferably I guess, although it probably doesn't change anything because we're looking for the binary name

tych0: It does if the binary name has a space in it :)

libqtile/core/manager.py
tych0: So, I couldn't figure out how to do this nicely. Turns out that if `shell=True`, we need to render `cmd` as a string, but if `shell=False`, we want it as a list. I'm open to other implementations, though.

ramnes: This is the best I could come with:

```python
if isinstance(cmd, str):
    args = shlex.split(cmd)
    cmd = subprocess.list2cmdline(args)
else:
    args = list(cmd)

to_lookup = args[0]
if shell:
    args = ["/bin/sh", "-c", cmd]
```

tych0: Cool, thanks, I updated it to this.

PR #2304, opened 2021-03-17
Fix import_error widget

elParaguayo:
ImportError widget crashes if user has configured widget with positional arguments. This can happen if users don't use the relevant keyword but pass the arguments in the correct order.

For example, if a user does not have `xdg` installed and configures `Launchbar` as follows:

``` python
widget.LaunchBar([      
    ("google-chrome", "google-chrome-stable", "Launch Google Chrome"),
    ("firefox", "firefox", "Launch Google Firefox")
    ])
```
`ImportError` will fail as there is a positional argument. However, if the widget had been configured with a keyword argument:

``` python
widget.LaunchBar(progs=[      
    ("google-chrome", "google-chrome-stable", "Launch Google Chrome"),
    ("firefox", "firefox", "Launch Google Firefox")
    ])
```
then there would be no error.

This PR should allow both scenarios to be addressed.

### DIFF
--- a/libqtile/widget/import_error.py
+++ b/libqtile/widget/import_error.py
@@ -22,13 +22,13 @@ from libqtile.widget.base import _TextBox
 
 
 class ImportErrorWidget(_TextBox):
-    def __init__(self, class_name, **config):
+    def __init__(self, class_name, *args, **config):
         _TextBox.__init__(self, **config)
         self.text = "Import Error: %s" % class_name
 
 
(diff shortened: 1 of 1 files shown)


tych0:
Feels like progs should be in the "defaults" list, since it's a user configurable setting. That said, this patch also seems reasonable to me.

elParaguayo:
I don't disagree! This just gives us an extra layer of protection.

PR #2305, opened 2021-03-17
NvidiaSensors: Add check for correct sensors names in format string, small refactoring

akkaky:


libqtile/widget/nvidia_sensors.py
```diff
@@ -48,17 +52,27 @@ def _get_sensors_data(self, command):
...
+                for gpu in self._get_sensors_data(command)
+            ]
+            for gpu in sensors_data:
+                if gpu.get('temp') and int(gpu['temp']) > self.threshold:
+                    self.foreground = self.foreground_alert
+            return ' - '.join(
+                [self.format.format(**gpu) for gpu in sensors_data]
+            )
+        except Exception as e:
+            return e
```
tych0, line 78: Let's not catch and return a string, it is easier to debug if we just leave the exception as is and let qtile's framework log it.

tych0:
Again, would be great to get some test data for this...

akkaky:
Maybe I didn't understand right. Write tests for code?

tych0:
> Maybe I didn't understand right. Write tests for code?

Yeah, for the parsing code specifically. People like to refactor these kinds of things, and it's nice to have some test data to validate that it still works. For example, just grabbing some output and putting it in a hard coded string in a test, then run your parsing function over it and see if it produces what you expect. The tests you've added here are a good start, though, thanks for those!

Issue #2247, opened 2021-02-23
AttributeError: KeyChord breaks cmd_display_kb


ep12:
# Issue description
qtile shell excerpt:
```
 > display_kb()
Command exception: Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/libqtile/command/interface.py", line 312, in call
    return SUCCESS, cmd(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/libqtile/core/manager.py", line 1001, in cmd_display_kb
    if not k.commands:
AttributeError: 'KeyChord' object has no attribute 'commands'
```
# Configuration
At least one `KeyChord` in `keys`. My config: https://www.github.com/ep12/qtileconf

elParaguayo:
Yes, that makes sense as a KeyChord doesn't have that attribute. That line could be changed to:
``` python
if not getattr(k, "commands", None):
```
So it checks if the attribute exists and returns None if not.

Separate question is if/how should KeyChords be handled in this function? You've got nested KeyChords too in your config which would be harder still for this function.

ep12:
The `display_kb` function already contains a local class, why not define a function that handles the `Key` and `KeyChord` elements using recursion? `KeyChord` would be translated into a "enter ... mode" row and then the function would be called on all the child elements... Maybe another column `mode` should be prepended such that keybindings are sorted by mode first, then as before?

ep12:
> Separate question is if/how should KeyChords be handled in this function? You've got nested KeyChords **too** in your config which would be harder still for this function.

@elParaguayo If you want to nest `KeyChord`s too, are you interested in #2241 and #2264. I know that there is #2121 but it implements different things...
It seems only very few people care about `KeyChord`s, and even fewer about nested ones... I would be glad to get some feedback and maybe you could test the branch?

elParaguayo:
I don't use chords much at all and don't currently have any nested ones. My interest in this is primarily as a coding problem that can be fixed.

PR #2308, opened 2021-03-19
Fix `cmd_display_kb` in conjunction with `KeyChord`.

ep12:
Fixes: #2247.
+ [x] This pr fixes an `AttributeError` when calling `display_kb` while using `KeyChords`. It does this by adding a new `mode` column to the table which `display_kb` creates. All the bindings are then added to the table
recursively.
+ [x] Support for named, non-persistent modes (e. g. in #2121 or a similar implementation) should be easily implementable.
+ [x] Adds tests for `display_kb` (which previously didn't have one)
+ [x] ***RFC*** optional new feature: add a new output format option to `display_kb` to enhance fiddling with the output. See the commit message for details. This is definitely optional, so I don't bother if this last commit is removed before merging. It wasn't hard to implement at that low level, so it didn't cost much energy. Now that I've implemented it, it is easier for you to try out. _I am looking forward to reading the discussion about this though_ :smile:.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1036,19 +1036,42 @@ class Qtile(CommandObject):
                 return "".join([format % tuple(self.expandlist(row, n)) for row in self.rows])
 
         result = FormatTable()
-        result.add(["KeySym", "Mod", "Command", "Desc"])
+        result.add(["Mode", "KeySym", "Mod", "Command", "Desc"])
         result.add([])
         rows = []
+
+        def walk_binding(ks: Union[int, str], kmm: Union[int, List[str]], k: Union[Key, KeyChord],
+                         mode: str) -> None:
+            nonlocal rows
+            if not (isinstance(ks, (int, str)) and isinstance(kmm, (int, list))):
+                return
+            name = (", ".join(xcbq.rkeysyms.get(ks, ("<unknown>", )))
+                    if isinstance(ks, int) else ks)
+            modifiers = (", ".join(xcbq.translate_modifiers(kmm))
+                         if isinstance(kmm, int) else ", ".join(kmm))
+            if isinstance(k, Key):
+                if not k.commands:
+                    return
+                allargs = ", ".join(
+                    [value.__name__ if callable(value) else repr(value)
+                     for value in k.commands[0].args] +
+                    ["%s = %s" % (keyword, repr(value)) for keyword, value in k.commands[0].kwargs.items()]
+                )
+                rows.append((mode, name, modifiers,
+                             "{:s}({:s})".format(k.commands[0].name, allargs), k.desc))
+                return
+            if isinstance(k, KeyChord):
+                new_mode_s = k.mode if k.mode else "<unnamed>"
+                new_mode = (k.mode if mode == "<root>" else
+                            "{}>{}".format(mode, k.mode if k.mode else "_"))
+                rows.append((mode, name, modifiers, "", "Enter {:s} mode".format(new_mode_s)))
+                for s in k.submappings:
+                    walk_binding(s.key, s.modifiers, s, new_mode)
+                return
+            raise TypeError("Unexpected type: {}".format(type(k)))
+
         for (ks, kmm), k in self.keys_map.items():
-            if not k.commands:
-                continue
-            name = ", ".join(xcbq.rkeysyms.get(ks, ("<unknown>", )))
-            modifiers = ", ".join(xcbq.translate_modifiers(kmm))
-            allargs = ", ".join(
-                [repr(value) for value in k.commands[0].args] +
-                ["%s = %s" % (keyword, repr(value)) for keyword, value in k.commands[0].kwargs.items()]
-            )
-            rows.append((name, str(modifiers), "{0:s}({1:s})".format(k.commands[0].name, allargs), k.desc))
+            walk_binding(ks, kmm, k, "<root>")
         rows.sort()
         for row in rows:
             result.add(row)
--- a/test/test_qtile_cmd.py
+++ b/test/test_qtile_cmd.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 
 import os
+import re
 import subprocess
 
 import pytest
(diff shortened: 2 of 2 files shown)


libqtile/core/manager.py
```diff
@@ -1036,19 +1036,42 @@ def __str__(self):
                 return "".join([format % tuple(self.expandlist(row, n)) for row in self.rows])
 
         result = FormatTable()
-        result.add(["KeySym", "Mod", "Command", "Desc"])
+        result.add(["mode", "KeySym", "Mod", "Command", "Desc"])
         result.add([])
         rows = []
+
+        def recurse(ks: Union[int, str], kmm: Union[int, List[str]], k: Union[Key, KeyChord],
```
tych0, line 1043: Can we give this a better name? walk_key or something?

ep12: Sure, no problem. But how on earth? I mean, this is a draft. Why are you reviewing that sooo early? :rofl:

tych0: You made the PR, I'll review it if/when I have time. In my experience, early reviews are better than late reviews...

m-col:
In general this look good, and I'm also in favour of a "machine-readable" option if the main output format is tricky to manipulate. Do you also see this kind of output? I don't find this readable by anything:
image

An option might be to have a very easily parsed human-readable output format like a very simple table delimited by whitespace.

ep12:
> In general this look good, and I'm also in favour of a "machine-readable" option if the main output format is tricky to manipulate. Do you also see this kind of output? I don't find this readable by anything:
> image
> 
> An option might be to have a very easily parsed human-readable output format like a very simple table delimited by whitespace.

Yep, we have all have that problem and it has been reported elsewhere. If I recall it correctly, I pointed to the source where the result of an ipc call is pretty-printed. That causes this weird output. Run it through `print(eval(...))` and you get the table. Same for the "machine-readable" format which also is a pretty-printed list

m-col:
I think I might be missing something... why would we want it printing out like this? Shouldn't it _always_ be printed nicely? And what would the difference be between the machine-readable version and the regular one?

ep12:
> I think I might be missing something... why would we want it printing out like this? Shouldn't it _always_ be printed nicely? And what would the difference be between the machine-readable version and the regular one?

@m-col Okay, it is [here](https://github.com/qtile/qtile/blob/0444d047c9ac0606fcbb3aa78d2ac0750df99bca/libqtile/scripts/cmd_obj.py#L182-L184). The result is pretty-printed which is appropriate most of the time, e. g. for `qtile cmd-obj -o cmd -f commands` and the like (` -f windows`, ...). `display_kb` returns a string by default which is then pretty-printed and loses its painfully calculated format. I don't know if the calls to `run_function` result in lists and dicts that get printed nicely (the output looks like that is the case :zap: but [the signature](https://github.com/qtile/qtile/blob/0444d047c9ac0606fcbb3aa78d2ac0750df99bca/libqtile/scripts/cmd_obj.py#L138) indicates a `str` return value :zap:). Actually, the `str` result seems more plausible because the IPC works using nothing but strings I think. But the pretty-printed output looks really more like `pprint(<list object>)`.

----------------

The `list` output format is a (pretty-printed) list which might be easier for post-processing (but this feature is still optional, the most important bit of this pr are the first two commits). This is the default format:
default_format_table
An this is the list format (not really, see the second command in the pipe:
list_format_image
The unprocessed output of `print(eval(sys.stdin.read()))` looks horrible (because _not_ pretty-printed) so I printed the type of `l` and then the individual items.
As you can see, the output data is the same, the output format however is easy to manipulate right away. Still, doesn't mean that this is of any use to anybody. Parsing of the default output can also be done, so `no bloat` is definitely an argument here. It's up to you to decide if you want this small additional feature.

m-col:
Cool, thanks for the clarification. What I don't understand is why `print(eval(sys.stdin.read()))` is needed at all. Should we not simply adjust the return value of `cmd_display_kb` such that, when pretty-printed via the IPC (which seems to me the only way anyone is accessing this data), it is already formatted into the table in your first image? Perhaps we are just familiar with different tools, but piping that first table to awk & pals looks much more convenient than trying to parse that second table.

ep12:
The `print(eval(...))` is solely required because of `pprint.pprint` (which, as stated above, is a good choice in many cases). Either this would have to be changed to _pretty-print if not running `display_kb`, in that case just print_ or the `eval` is necessary. It is (nearly or completely) impossible to make the argument to `pprint` be printed as a table (first screenshot) because it already knows how to deal with lists, strings, dictionaries... And having that extra effort just to catch the call to `display_kb` and then `print` instead of `pprint` is really quite ugly I think. 
In that case it might be better to introduce a new sub-command `qtile display-kb` that sits in-between and directly gets an object from the ipc. The advantage of that idea I just had would possibly even enable us to move the table formatting code out of `manager.py` (where it feels really weird tbh). If we can send JSON data through the ipc wire, we can do the formatting in that sub-command if the user wants us to. Otherwise we could also output the JSON data directly as this is what the new `list` format (optional feature of this pr) does. 


Should I open an issue about that and remove that last feature commit?

m-col:
> It is (nearly or completely) impossible to make the argument to `pprint` be printed as a table (first screenshot) because it already knows how to deal with lists, strings, dictionaries

Isn't it pprint that created that table in the first screenshot?

ep12:
> Perhaps we are just familiar with different tools, but piping that first table to awk & pals looks much more convenient than trying to parse that second table.

The second "table" is not the raw output. It really is a pretty-printed list of lists of strings that is evaluated back to a list of list of strings that can be easily turned into JSON or csv both of which are IMHO easier to deal with (`xsv`, `nushell` and other nifty tools) than raw output with regular expressions, aka `awk` and the like. No matter what the user prefers, if both formats are available (or even csv as support for that inside a new sub-command is easy) the user can freely choose which one requires less piping. Or more, if you like shell golf ⛳ :rofl:

ep12:
> Isn't it pprint that created that table in the first screenshot?

No, sadly not: it is painfully calculated inside `cmd_display_kb` and then returned as a string. Have a look at that very method:  
Inside it, we define a class (and with pr an additional function) that is then used to get a nicely formatted string that `prints` as a table (and `pprints` as garbage).

ep12:
Should I remove that last commit then?

ramnes:
Please move that last commit in another PR and I'll merge the rest. We need  to have this discussion about display_kb and IPC in general, it's been a burden for too long.

ep12:
Done. I don't think I'll open another pr though, because I think the discussion will result in a change of some sort, so it's not worth it working on `display_kb` without having decided on how to improve ipc.

ramnes:
Sounds good. Thanks!

PR #2310, opened 2021-03-20
Calling config main functions can now be fully deprecated

m-col:
Defining a 'main' function was deprecated in version 0.16.1. This change
finishes off its deprecation so it is no longer executed.

A couple of tests that used a main function were removed rather than
replaced because the functionality that they test is already tested
elsewhere.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,8 @@ Qtile 0.x.x, released xxxx-xx-xx:
           dbus calls within asyncio's eventloop.
         - widget.BatteryIcon no longer has a fallback text mode; use
           widget.Battery instead
+    !!! deprecation warning !!!
+        - 'main' config functions, deprecated in 0.16.1, will no longer be executed.
     * features
         - added measure_mem and measure_swap attributes to memory widget to allow user to choose measurement units.
         - memory widget can now be displayed with decimal values
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -29,7 +29,6 @@ import signal
 import subprocess
 import tempfile
 import time
-import warnings
 from typing import Dict, List, Optional, Tuple, Union
 
 import xcffib
@@ -112,9 +111,6 @@ class Qtile(CommandObject):
             widgets.insert(0, TextBox('Config Err!'))
 
         self.core.wmname = getattr(self.config, "wmname", "qtile")
-        if self.config.main:
-            warnings.warn("Defining a main function is deprecated, use libqtile.qtile", DeprecationWarning)
-            self.config.main(self)
 
         self.dgroups = DGroups(self, self.config.groups, self.config.dgroups_key_binder)
 
--- a/test/layouts/test_common.py
+++ b/test/layouts/test_common.py
@@ -20,6 +20,7 @@
 
 import pytest
 
+import libqtile
 import libqtile.config
 import libqtile.hook
 from libqtile import layout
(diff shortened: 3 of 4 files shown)


m-col:
A couple of tests are using `main` to subscribe a hook. I tried moving the subcribe line into the main body of the test but then it doesn't seem to get fired. The tests in question are checking the `focus_change` hook is firing as expected during layout operations. It seems we test hooks in isolation and never during a running Qtile instance anywhere else. Perhaps we need a better way to subscribe hooks from tests where they will get fired correctly?

tcld:
Should/could there be a migration for this?

Also, what is the recommended way of handling the "main"-code? I assume putting your code under `if __name__ in ['config', '__main__']:` will still work?

m-col:
> Should/could there be a migration for this?

Yeah good idea.

> Also, what is the recommended way of handling the "main"-code? I assume putting your code under if __name__ in ['config', '__main__']: will still work?

`main` was originally used to access the `qtile` object in configs but is no longer needed as `qtile` can be directly imported. I'm not sure I understand that `if` statement -- why not just have it at the top level? (Maybe you/others use their configs in other ways, whereas my config is only ever used by Qtile itself so `__name__` will always be `'config'` for me.)

For migration we could either remove the `def main` and dedent that block, or we could replace it with your `if` statement. Or even stick a `@hook.subscribe.startup` before it.

tcld:
I guess I just like to have my code in some block, so I can fold it away easily. Before we had `qtile check` (even though I haven't gotten it to work properly for me yet), I sometimes executed my config with python to see if there were any major mistakes. That's why it checks for `main`.

A hook actually sounds like a really clean idea to me. Maybe other people agree with that?

tych0:
Yeah, making it a hook seems fine with me.

ramnes:
Yes. `qtile` has a few attributes that are not instantiated yet when the config is imported IIRC, in which case `@hook.subscribe.startup` is a strictly superior solution.

m-col:
I fixed the failed test so just have to implement the migration.

What if someone already updated their config simply by hooking their pre-existing `main` function to the startup_once hook? A migration based on hooking (but not editing) the `main` function would cause it to be called twice. A migration based on editing the `def main` to hook it with the leading `@` syntax may introduce `NameError`s if `main` is refrenced later (e.g. if hooked later).

This leads to a kinda general question about migration - are we happy to have some level of error, on the basis that migrating is an interactive process, and the migration script is a convenience as part of an interactive process (rather than a complete solution)?

tcld:
You could _attemp_ a migration followed by a `qtile check` and, if it fails, report the errors back to the user (maybe including some hints so that user can fix things himself). So at the very least no migration should introduce errors.

Any migration you come up with would likely only affect some people, since others might not use a `main` function at all or use the syntax I mentioned above, etc. Unless you can at least isolate the straight forward cases, maybe this isn't a good candidate for a migration after all?

ep12:
That sounds quite similar to the #2325 [migration](https://github.com/qtile/qtile/pull/2325#discussion_r602343394) that was unreadable, unreliable and is avoided by not deprecating. This is not an option here, of course. But bowler seems to be quite young and operates on a very low level. More like token instead of AST. So the most reasonable option I see is to accept erroneous migrations in case of such major changes because the user will probably be intelligent enough to tell that duplicating the decorator line is bogus. The process is interactive, so the user should see the changes when running `qtile migrate` in the terminal. Or is the output suppressed sometimes?
In general, I do have to admit that the situation is quite unfortunate that it is so easy to come up with examples where the migration fails.

ramnes:
I'm fine with not handling that kind of situation. That being said, maybe `qtile migrate` should by default print the differences interactively and ask for each block if it should be merged, edited, or ignored, unless a `-y` option or alike is given.

m-col:
Currently it prints every change it makes, but makes the change without confirmation. That might be enough, but I'd also be in favour of making `qtile migrate --interactive` the default, adding a `--non-interactive` to replace it, and if possible changing the default confirmation to 'y' in `Apply this hunk [y,N,q,a,d,?]?` (though that's minor). If people agree I can open a PR for that.

Either way, I think simply hooking the `main` will do and will likely cover most people. Those whose configs are not migrated correctly will have the output of `qtile migrate` for confirmation.

ep12:
+1 for making `migrate` interactive by default

PR #2314, opened 2021-03-21
fix: code quality issues

withshubh:
### Description
Hi :wave: I work at [DeepSource](https://deepsource.io), I ran DeepSource analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/qtile/issues/?category=recommended) in the codebase, opening this PR so you can assess if our platform is right and helpful for you.

### Summary of changes

- Used literal syntax instead of function calls to create data structure
- Removed length check in favour of truthiness of the object
- Removed unnecessary use of comprehension
- Used shallow copy of dictionary for iteration
- Used `sys.exit()` calls
- Removed methods with unnecessary super delegation
- Removed unnecessary generator
- Used `is` to compare type of objects
- Iterated dictionary directly
- Added .deepsource.toml to continuously analyze the repo for code quality issues

### DIFF
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -258,7 +258,7 @@ class ScratchPad(group._Group):
         If the associated window is killed, reset internal state.
         """
         name = None
-        for name, dd in self.dropdowns.items():
+        for name, dd in self.dropdowns.copy().items():
             if dd.window is client:
                 dd.unsubscribe()
                 del self.dropdowns[name]
@@ -273,7 +273,7 @@ class ScratchPad(group._Group):
         will spawn a new process.
         """
         name = None
-        for name, dd in self.dropdowns.items():
+        for name, dd in self.dropdowns.copy().items():
             if not dd.window.floating:
                 if dd.window.group is not self:
                     dd.unsubscribe()
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -309,7 +309,7 @@ class ScratchPad(group._Group):
         If name is None, a list of all dropdown names is returned.
         """
         if name is None:
-            return {'dropdowns': [ddname for ddname in self._dropdownconfig]}
+            return {'dropdowns': list(self._dropdownconfig)}
         elif name in self.dropdowns:
             return self.dropdowns[name].info()
         elif name in self._dropdownconfig:
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,16 @@
+version = 1
+
+test_patterns = ["test/**/test_*.py"]
+
+exclude_patterns = [
+    "bin/**",
+    "docs/**",
+    "scripts/**"
+]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -553,7 +553,7 @@ class Window:
         if r:
             s = self._property_string(r)
             return tuple(s.strip("\0").split("\0"))
-        return tuple()
+        return ()
 
     def get_wm_window_role(self):
         r = self.get_property("WM_WINDOW_ROLE", "STRING")
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -903,10 +903,8 @@ class Connection:
         return self.conn.get_setup()
 
     def extensions(self):
-        return set(
-            i.name.to_string().lower()
-            for i in self.conn.core.ListExtensions().reply().names
-        )
+        return {i.name.to_string().lower()
+            for i in self.conn.core.ListExtensions().reply().names}
 
     def fixup_focus(self):
         """
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -264,7 +264,7 @@ class AtomCache:
         self.reverse = {}
 
         # We can change the pre-loads not to wait for a return
-        for name in WindowTypes.keys():
+        for name in WindowTypes:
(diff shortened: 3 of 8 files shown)


m-col:
Pretty nifty. There seems to be a lot of nice suggestions on the website. I have a couple of questions:

- An example code quality issue on your link is this: https://deepsource.io/gh/withshubh/qtile/issue/PYL-W0612/occurrences. Is there an automated way to get PRs submitted to correct these e.g. setting unused return values to `_` or something?
- Is there a way to get a description/explanation for suggested changes if it wasn't obvious why the proposed change is better?*
-

* E.g. "Use shallow copy of dictionary for iteration" - why is this better?

m-col:
I hit 'Comment' before finishing my comment... just letting you know in case you'd reply by email and miss the rest of it.

I'd also like to know out of interest -- what's DeepSource's business model?

tych0:
Hi,

I have to say I generally don't love PRs like this. There's a bunch of churn and no reasoning about *why* that churn is good. Clearly, this fixes bugs like `s/exit/sys.exit/` which would otherwise generate name errors (although that commit doesn't mention that it fixes the bug, it just describes the change visible in the diff).

Other stuff like iterating over copies of dictionaries or using set literal notation are not "code quality issues" so much as style fixes. If we're going to introduce a bunch of style fixes to the tree, we need to add things to CI to keep them from re-appearing, otherwise this just introduces noise into git log for no real benefit.

ep12:
@tych0 an option to catch those style errors is to use `pylint` and `pycodestyle` (and even `pydocstyle`?). There are indeed quite a lot of things pylint is reporting (I use it all the time: `flycheck` runs it while coding in `emacs`). One particular style issue I fixed in my last pr that got merged was an unnecessary `else` after a `for` block. `for-else` already is a combination the normal programmer needs to recall what it actually does. Hence it is super annoying to find out that those thoughts (maybe even a quick web search) where completely unnecessary as the `else` block is called always. If you like, I'll play around with it and create a pr to see what those tools are able to detect...

tych0:
I would say that I don't love style fixes in general, because everyone has a different idea of what good style is. I have had problems in the past with too many style fixes making the tree incomprehensible for me when I'm trying to go through the git log to find bugs.

That said, while style fixes are annoying, those that make the code conform to some PEP and come with a CI fix are fine with me. In cases where there isn't a PEP, the commit message should have a description of why something is preferred, not just a description of what's in the diff.

ep12:
> I would say that I don't love style fixes in general, because everyone has a different idea of what good style is. I have had problems in the past with too many style fixes making the tree incomprehensible for me when I'm trying to go through the git log to find bugs.
> 
> That said, while style fixes are annoying, those that make the code conform to some PEP and come with a CI fix are fine with me. In cases where there isn't a PEP, the commit message should have a description of why something is preferred, not just a description of what's in the diff.

Okay @tych0 I'll open a draft pr soon. First of all, we already have a `.pylintrc`, we don't use it yet. Secondly, I agree that style might differ, e. g. naming so I'll disable most of the warnings right away. Then we can start with the critical things (mutable default values etc.) and re-enable stuff as we go. I think it is better to have a strict CI telling me that I shouldn't use `eval` or `exec`. If it is necessary at some point, I can disable the warning for just one line instead of ignoring it always.

tych0:
Yep, that sounds great, thanks!

libqtile/scratchpad.py
```diff
@@ -258,7 +258,7 @@ def on_client_killed(self, client, *args, **kwargs):
         If the associated window is killed, reset internal state.
         """
         name = None
-        for name, dd in self.dropdowns.items():
+        for name, dd in self.dropdowns.copy().items():
```
ep12, line 261: I don't think `self.dropdowns` changes during runtime.

ep12:
@tych0 @ramnes @m-col I had a look at the profile of the OP and found jacebrowning/memegen/pull/598, fluent/fluentd/pull/3286, microsoft/qlib/pull/311, CN-UPB/DeepCoMP/pull/3. All those pull requests are incredibly similar (even the choice of words) to this one. I can't get rid of the feeling that this is just some kind of campaign to add a `.deepsource.toml` to the project. Those commits are trivial, don't change much and only fix one or two occurrences of the same problem each. Only one commit 5cba2f2 changes something that has the potential to be an issue (though probably not here). I don't know what the motivation behind this is (apart from using DeepSource) but I doubt that it's the love for qtile...

tych0:
I agree, though two of the patches do seem useful. I've picked those up in #2317. I'll close this for now.

withshubh:
Hi @m-col :wave: 

> * An example code quality issue on your link is this: [deepsource.io/gh/withshubh/qtile/issue/PYL-W0612/occurrences](https://deepsource.io/gh/withshubh/qtile/issue/PYL-W0612/occurrences). Is there an automated way to get PRs submitted to correct these e.g. setting unused return values to `_` or something?
- As of now there is no autofix available for this issue since this issue is sometimes opinionated. We are constantly working to increase the number of autofix for the issues. Read more [here](https://deepsource.io/python-static-code-analysis/).

> * Is there a way to get a description/explanation for suggested changes if it wasn't obvious why the proposed change is better?* 
> * E.g. "Use shallow copy of dictionary for iteration" - why is this better?
- Yes, There is a description available for every issue caught. For the issue that you mentioned read the description [here](https://deepsource.io/gh/withshubh/qtile/issue/PTC-W0056/description).

> * Is there a way to permanently ignore one type of suggestion if required (not that I have seen one)?
- You can silence issues directly from the DeepSource dashboard or use `skipcq`. Read more [here](https://deepsource.io/docs/how-to/silence-issues.html).

> I'd also like to know out of interest -- what's DeepSource's business model?
- DeepSource is Free for Open Source projects forever and has premium plans for enterprise, deployment and more.

withshubh:
Thanks for your reviews everyone! :)

PR #2316, opened 2021-03-22
Add CI checks for code and documentation "style"

ep12:
See #2314.

This pr adds two new test environments `codestyle` and `docstyle` to the `tox` configuration.  
Check your code locally with `tox -e codestyle,docstyle` if you like.

The `codestyle` environment runs `pylint` and `pycodestyle`. The current HEAD already satisfies `pycodestyle` and hence the test will fail if `pycodestyle` does.
The most important bit is `pylint`: It will help us to find not only real _style_ issues, but also basic programming mistakes (mutable default arguments, ...). Right now it complains about A LOT of things although I have initially disabled dozens of warnings to help find critical insuffiencies.

The goal is to either fix the problems to disable the messages or, if absolutely necessary, disable the warnings on a local basis. By that I mean: for single lines, functions or modules.  
The more localised, the better: Let's give an example here: We don't want to disable the warning regarding the use of `eval` or `exec` for a module, probably not even for a function, but instead for a single line where we need to make sure that it is safe to use.

We can disable messages locally with a comment `# pylint: disable=exec-used,some-other-warning,...`.  
Depending on where that comment is placed, it disables a warning for a line, function or module.  
It should be made clear why a message is disabled. If it isn't obvious, better write a few comment lines explaining the choice.

Currently, only the main code is checked with the tools mentioned above.  
Later on we might also want to check the tests which can easily be done, of course. Until then there remains a lot to do.

The `docstyle` environment runs `pydocstyle` to check if we did our homework and documented out functions well enough.  Currently, it is ignored but can be used to find undocumented places in the code or to check if your documentation is verbose enough.

+ [x] add CI configuration
+ [x] deal with `dangerous-default-value`, `undefined-loop-variable`, `deprecated-method`, `useless-else-on-loop` and `pointless-string-statement`
+ [x] enable `consider-using-sys-exit`, `useless-super-delegation`
+ [x] add a note to `CONTRIBUTING.md` and `docs/manual/contributing.rst`
+ [x] deal with `redefined-builtin`: Not done, but postponed for another pr.

### DIFF
--- a/docs/manual/contributing.rst
+++ b/docs/manual/contributing.rst
@@ -74,3 +74,37 @@ Qtile session.
 For any Qtile-specific question on testing, feel free to ask on our `issue
 tracker <https://github.com/qtile/qtile/issues>`_ or on IRC (#qtile on
 irc.oftc.net).
+
+Running tests locally
+---------------------
+
+This section gives an overview about ``tox`` so that you don't have to search
+`its documentation <https://tox.readthedocs.io/en/latest/>`_ just to get
+started.
+Checks are grouped in so-called ``environments``. Some of them are configured to
+check that the code works (the usual unit test, e.g. ``py39``, ``pypy3``),
+others make sure that your code conforms to the style guide (``pep8``,
+``codestyle``, ``mypy``). A third kind of test verifies that the documentation
+and packaging processes work (``docs``, ``docstyle``, ``packaging``).
+
+The following examples show how to run tests locally:
+   * To run the functional tests, use ``tox -e py39`` (or a different
+     environment). You can specify to only run a specific test file or even a
+     specific test within that file with the following commands:
+
+     .. code-block:: bash
+
+        tox -e py39 # Run all tests with python 3.9 as the interpreter
+        tox -e py39 -- -x test/widgets/test_widgetbox.py  # run a single file
+        tox -e py39 -- -x test/widgets/test_widgetbox.py::test_widgetbox_widget
+
+   * To run style and building checks, use ``tox -e docs,packaging,pep8,...``.
+     You can use ``-p auto`` to run the environments in parallel.
+
+     .. important::
+
+        The CI is configured to run all the environments. Hence it can be time-
+        consuming to make all the tests pass. As stated above, pull requests
+        that don't pass the tests are considered incomplete. Don't forget that
+        this does not only include the functionality, but the style, typing
+        annotations (if necessary) and documentation as well!
--- a/.pylintrc
+++ b/.pylintrc
@@ -141,6 +141,7 @@ disable=print-statement,
         deprecated-sys-function,
         exception-escape,
         comprehension-escape,
+        not-callable,  # TEMP: has false-positives currently. Re-enable later.
         protected-access,  # we need this a lot, so it will stay here!
         # all messages disabled above were suggested by
         # `pylint --generate-rcfile`
--- a/.pylintrc
+++ b/.pylintrc
@@ -143,6 +143,7 @@ disable=print-statement,
         comprehension-escape,
         not-callable,  # TEMP: has false-positives currently. Re-enable later.
         protected-access,  # we need this a lot, so it will stay here!
+        c-extension-no-member,  # this is probably staying here, too.
         # all messages disabled above were suggested by
         # `pylint --generate-rcfile`
         # maybe some of them can be re-enabled later, but only after we have
@@ -162,7 +163,6 @@ disable=print-statement,
         broad-except,
         bare-except,
         assignment-from-no-return,
-        c-extension-no-member,
         redefined-outer-name,
         unused-argument,
         unused-variable,
@@ -183,6 +183,9 @@ disable=print-statement,
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once). See also the "--disable" option for examples.
 # enable=c-extension-no-member
+# Time to play Whac-A-Mole:
+enable=consider-using-sys-exit,
+       useless-super-delegation,
 
 
 [REPORTS]
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,19 +1,640 @@
 [MASTER]
 
-# files that cause recursion depth exceeded errors
-ignore=graph.py,manager.py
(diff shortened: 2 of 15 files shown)


ep12:
@tych0 Here it is. Currently configured such that it should pass the tests and ignore the failures of `pylint` and `pydocstyle`.

ep12:
Looks like we are indeed running tests with `Bar([])`. That probably causes error messages when `Bar.draw()` is called.

ep12:
> Looks like we are indeed running tests with `Bar([])`. That probably causes error messages when `Bar.draw()` is called.

It does _not_ cause error messages in the log because [`self.qtile.call_soon(self._actual_draw)`](https://github.com/qtile/qtile/blob/0444d047c9ac0606fcbb3aa78d2ac0750df99bca/libqtile/bar.py#L405) suppresses them:
image
And indeed, I configured an empty `Bar([], 10, background='#ffffff')` that showed up as a black bar. Not a relevant security issue and under normal circumstances not relevant, but definitely unexpected behaviour and hence a bug.

I'll fix the test and force-push later...

ep12:
RFC: @tych0, @m-col, ...: The message `W0622: Redefining built-in %r (redefined-builtin)` pops up quite a lot. Do we want to fix this (and rename the variables) or disable the warning?

ramnes:
I personally don't care. Let's disable it for a start so that the PR doesn't get too big?

ep12:
> I personally don't care. Let's disable it for a start so that the PR doesn't get too big?

I can always open new pull requests if you want to keep this one simple. But do you prefer to fix those messages or to accept snippets like
```py
def some_func(a, b, type):
    if type is 'abc':
        ...
    ...
```
Obviously code that then contains type checks like `type(a) == list` (we currently have checks like this!) would not work in that function. This could slip through if the functions are not covered by tests. More critical are imports that redefine built-ins at the module level (we redefine `hex` somewhere). It is a style choice, so it's up to you to decide.

Btw. I planned to just add some text to the docs and then mark the pr ready for review. The changes in 4b61c04 were mainly to fix the most critical things and to demonstrate how to disable messages locally after making sure that it is safe to do so (in other words: _prove_ (check) that implicit requirements are fulfilled).
Oh and the next thing is to play Whac-A-Mole, i.e. to enable the messages that check for `consider-using-sys-exit` and `useless-super-delegation`.

ramnes:
Okay, so let me rephrase my previous comment.

It's obviously not a good practice to reuse built-ins names, so I'm not against a change in that respect. But I also assume that developers (and reviewers) are adults who only do (or accept) this when it provides clarity and when the built-in is not likely to be used in the logic of the function, which means that fixing this all around the code will probably make it less readable and mostly mess up the git history. Therefore, given the pros and cons of each alternative, I'm deciding not to decide.

If you want me or other developers to merge pull requests, you'll have better luck if they only contain things we want to decide on, and thus if the difference is small. Assuming we know our basics like the drawbacks of reusing build-ins names will probably help saving time as well. :)

ep12:
@ramnes First of all: could you kill the checks, please? `pypy` is running since a couple of hours straight. Something's clearly wrong there.

ramnes:
done

ep12:
> It's obviously not a good practice to reuse built-ins names, so I'm not against a change in that respect. But I also assume that developers (and reviewers) are adults who only do (or accept) this when it provides clarity and when the built-in is not likely to be used in the logic of the function, which means that fixing this all around the code will probably make it less readable and mostly mess up the git history. Therefore, given the pros and cons of each alternative, I'm deciding not to decide.

Okay, I get that. I don't have sleepless nights because we redefine `ConnectionError`, `hex` or `id`. But redefining `type` is something I think we should avoid. Also, redefining things on a module level (we do that with `hex` afaik) is not good. There is no option to tell pylint to report only those cases (except for locally disabling the other ones with a comment on every line), so if we want to catch this practice in the CI at some point, we need to deal with this message. If some messages always plop up we ignore the outcome of the pylint run and hence could skip the test. It would be of no use whatsoever.
There are 16 occurrences of `redefined-builtin` which shouldn't introduce too much noise. Maybe I should just open a new pr when this one is merged to see how much noise it is exactly.

ep12:
> done

Is there any indication as to why it took so long? The changes in 7428c43 aren't even related to what the pypy env does...

ramnes:
> Is there any indication as to why it took so long? The changes in [7428c43](https://github.com/qtile/qtile/commit/7428c43bd7b1a27eef8c3a4abe4118e008d5112d) aren't even related to what the pypy env does...

Probably just a random race condition unrelated to your changes.

> There are 16 occurrences of redefined-builtin which shouldn't introduce too much noise. Maybe I should just open a new pr when this one is merged to see how much noise it is exactly.

That would be great, thanks for your understanding! :)

ep12:
a17dc2c: I added a few lines to the documentation. Feel free to suggest changes, reword things etc.

ep12:
> [a17dc2c](https://github.com/qtile/qtile/commit/a17dc2cd3ba6e409753afbdf8f7ae6fa701cb309): I added a few lines to the documentation. Feel free to suggest changes, reword things etc.

Wait: there is an `-x` missing from the examples. I'll have to fix this.

docs/manual/contributing.rst
```diff
@@ -74,3 +74,37 @@ Qtile session.
 For any Qtile-specific question on testing, feel free to ask on our `issue
 tracker <https://github.com/qtile/qtile/issues>`_ or on IRC (#qtile on
 irc.oftc.net).
+
+Running tests locally
```
tych0, line 78: Let's not copy paste this, please link from one to the other.

ep12: Didn't completely copy-paste :P. Which way around? The `contributing.rst` is better, right? Then I'd remove stuff from `CONTRIBUTING.md` and link to the docs.qtile.org page instead.

tych0: Either way is fine, I just don't want to have to maintain two copies of basically the same text.

ep12: :heavy_check_mark:

tych0:
LGTM, thanks!

PR #2317, opened 2021-03-23
Exit fixes

tych0:
These are the patches from #2314 that make sense IMO, plus one additional one to make the exit code of `qtile top` make a bit more sense.


ramnes:
Thanks everyone

PR #2318, opened 2021-03-23
test_migrate: ensure migrated config file is valid

tych0:
The result of qtile migrate should be a valid python file which should
parse and type check correctly. If mypy is present (which it is in CI),
let's check this.

### DIFF
--- a/test/test_migrate.py
+++ b/test/test_migrate.py
@@ -6,6 +6,7 @@ import tempfile
 import textwrap
 
 from libqtile.scripts.migrate import BACKUP_SUFFIX
+from test.test_check import have_mypy, run_qtile_check
 
 
 def run_qtile_migrate(config):
@@ -46,8 +47,9 @@ def check_migrate(orig, expected):
             f.write(orig.encode('utf-8'))
 
         run_qtile_migrate(config_path)
+        if have_mypy():
+            assert run_qtile_check(config_path)
 
-        config_path = os.path.join(tempdir, "config.py")
         assert expected == read_file(config_path)
         assert orig == read_file(config_path+BACKUP_SUFFIX)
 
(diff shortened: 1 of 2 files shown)


ep12:
That force-push you did to make the checks pass wasn't a clever idea I think. `tox.ini` tells me that `mypy` is only available in the `mypy` testenv and there we are running only `pytest test/test_check.py`. So this does not enable `mypy` checking in the CI (which is what you're aiming for).

tych0:
> That force-push you did to make the checks pass wasn't a clever idea I think.

Well it fixed a bug (testing the function for truthiness vs. the function's result), but,

> tox.ini tells me that mypy is only available in the mypy testenv and there we are running only pytest test/test_check.py. So this does not enable mypy checking in the CI (which is what you're aiming for).

Yeah, right you are. I guess maybe we need to run the migrate checks in the mypy environment as well.

test/test_migrate.py
```diff
@@ -46,8 +47,9 @@ def check_migrate(orig, expected):
             f.write(orig.encode('utf-8'))
 
         run_qtile_migrate(config_path)
+        if have_mypy():
+            run_qtile_check(config_path)
```
Fabien-Bernard, line 51: Doesn't seem to stop me doing my bad thing. Shouldn't we assert it?

tych0: Yes, looks like we should. Also looks like the mypy env doesn't have bowler... bleh.

ep12: Should we assert it? Wouldn't it be enough to run this check in one environment? I mean, we would be using the same bowler version in different test envs, that shouldn't make a difference and just makes the test runs slower...

tych0: In theory environments shouldn't make a difference at all; in practice they often do :). Yes, we should put an assert here, though.

m-col:
Nice, LGTM.

PR #2320, opened 2021-03-23
Strut compliance to ewmh specifications

### LINKED ISSUE #1947: External bar creating huge gap
## What is happening

As soon as I initialize polybar on my main screen a huge gap appears between the bar and the applications.

Main Screen

I use two monitors one of which is rotated. The issue only appeared on the main monitor up to now affecting one bar at a time. However both, top and bottom, have shown this effect.
Disabling the second monitor or just rotating it to a normal position seemed to makes the gap disappear the times I tried it. Also worth noting: the builtin bar works as expected. 

## Qtile-Version
0.16.1

## Stack Trace
qtile.log:
>2020-11-01 21:46:11,538 ERROR libqtile xcore.py:_xpoll():L277 Got an exception in poll loop
>Traceback (most recent call last):
>  File "/usr/lib/python3.8/site-packages/libqtile/backend/x11/xcore.py", line 230, in _xpoll  
>     event = self.conn.conn.poll_for_event()
>  File "/usr/lib/python3.8/site-packages/xcffib/__init__.py", line 571, in wrapper
>    return f(*args)
>  File "/usr/lib/python3.8/site-packages/xcffib/__init__.py", line 612, in poll_for_event
>    return self.hoist_event(e)
>  File "/usr/lib/python3.8/site-packages/xcffib/__init__.py", line 685, in hoist_event
>    return self._process_error(ffi.cast("xcb_generic_error_t *", e))
>  File "/usr/lib/python3.8/site-packages/xcffib/__init__.py", line 648, in _process_error
>    raise error(buf)
>xcffib.xproto.ValueError

.xsession-errors:
>RuntimeError: Cannot call write() after write_eof()
>Task exception was never retrieved
>future: <Task finished name='Task-5' coro=<Server._server_callback() done, defined at /usr/lib/python3.8/site-packages/libqtile/ipc.py:185> exception=RuntimeError('Cannot call write() after write_eof()')
>Traceback (most recent call last):
>  File "/usr/lib/python3.8/site-packages/libqtile/ipc.py", line 212, in _server_callback
>    writer.write(result)
>  File "/usr/lib/python3.8/asyncio/streams.py", line 341, in write
>    self._transport.write(data)
>  File "/usr/lib/python3.8/asyncio/selector_events.py", line 893, in write
>    raise RuntimeError('Cannot call write() after write_eof()')
>RuntimeError: Cannot call write() after write_eof()

SpringerJack:
According to [this part](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45841379520816) of the ewmh specification the strut values are meant to be used as absolute coordinates and not relative to any borders of physical screens. Fixes #1947

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -527,25 +527,24 @@ class Qtile(CommandObject):
     def current_window(self):
         return self.current_screen.group.current_window
 
-    def add_strut(self, strut):
+    def reserve_space(self, reserved_space, screen):
         from libqtile.bar import Bar, Gap
 
         for i, pos in enumerate(["left", "right", "top", "bottom"]):
-            if strut[i]:
-                bar = getattr(self.current_screen, pos)
+            if reserved_space[i]:
+                bar = getattr(screen, pos)
                 if isinstance(bar, Bar):
-                    bar.adjust_for_strut(strut[i])
+                    bar.adjust_for_strut(reserved_space[i])
                 elif isinstance(bar, Gap):
-                    bar.size += strut[i]
+                    bar.size += reserved_space[i]
                     if bar.size <= 0:
-                        setattr(self.current_screen, pos, None)
+                        setattr(screen, pos, None)
                 else:
-                    setattr(self.current_screen, pos, Gap(strut[i]))
+                    setattr(screen, pos, Gap(reserved_space[i]))
+        screen.resize()
 
-        self.current_screen.resize()
-
-    def remove_strut(self, strut):
-        self.add_strut([-i for i in strut])
+    def free_reserved_space(self, reserved_space, screen):
+        self.reserve_space([-i for i in reserved_space], screen)
 
     def map_window(self, window: xcbq.Window) -> None:
         c = self.manage(window)
@@ -589,7 +588,7 @@ class Qtile(CommandObject):
                 except (xcffib.xproto.WindowError, xcffib.xproto.AccessError):
                     return
 
-                if w.get_wm_type() == "dock" or c.strut:
+                if w.get_wm_type() == "dock" or c.reserved_space:
                     c.cmd_static(self.current_screen.index)
                     return
 
@@ -613,8 +612,8 @@ class Qtile(CommandObject):
         c = self.windows_map.get(win)
         if c:
             hook.fire("client_killed", c)
-            if c.strut:
-                self.remove_strut(c.strut)
+            if isinstance(c, window.Static):
+                self.free_reserved_space(c.reserved_space, c.screen)
             if getattr(c, "group", None):
                 c.group.remove(c)
             del self.windows_map[win]
--- a/libqtile/window.py
+++ b/libqtile/window.py
@@ -173,7 +173,7 @@ class _Window(CommandObject):
         self.borderwidth = 0
         self.bordercolor = None
         self.name = "<no name>"
-        self.strut = None
+        self.reserved_space = None
         self.state = NormalState
         self._float_state = NOT_FLOATING
         self._demands_attention = False
@@ -727,17 +727,45 @@ class Static(_Window):
             "_NET_WM_STRUT_PARTIAL",
             unpack=int
         )
-        strut = strut or self.window.get_property(
-            "_NET_WM_STRUT",
-            unpack=int
-        )
         if strut:
-            self.qtile.add_strut(strut)
-        self.strut = strut
+            x_screen_dimensions = self.qtile.root.get_geometry()
+            if strut[0]:    # left
+                x = strut[0]
+                y = (strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2
+            elif strut[1]:  # right
+                x = x_screen_dimensions.width - strut[1]
+                y = (strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2
+            elif strut[2]:  # top
+                x = (strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2
+                y = strut[2]
+            else:           # bottom
+                x = (strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2
+                y = x_screen_dimensions.height - strut[3]
+            self.screen = self.qtile.find_screen(x, y)
+
+            if self.screen is None:
+                logger.error("No screen at target")
+                return
+            elif None in [self.screen.x, self.screen.y, self.screen.height, self.screen.width]:
+                logger.error("Missing screen information")
+                return
+
+            empty_space = [
+                self.screen.x,
+                x_screen_dimensions.width - self.screen.x - self.screen.width,
+                self.screen.y,
+                x_screen_dimensions.height - self.screen.y - self.screen.height
+            ]
+            self.reserved_space = [strut[i] - empty if strut[i] else 0 for i, empty in enumerate(empty_space)]
+
+            self.qtile.reserve_space(self.reserved_space, self.screen)
+
+        else:
+            self.reserved_space = None
 
     def handle_PropertyNotify(self, e):  # noqa: N802
         name = self.qtile.conn.atoms.get_name(e.atom)
-        if name in ("_NET_WM_STRUT_PARTIAL", "_NET_WM_STRUT"):
+        if name == "_NET_WM_STRUT_PARTIAL":
             self.update_strut()
 
     def __repr__(self):
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -201,7 +201,6 @@ SUPPORTED_ATOMS = [
     '_NET_WM_DESKTOP',
     '_NET_WM_WINDOW_TYPE',
     '_NET_WM_STATE',
-    '_NET_WM_STRUT',
     '_NET_WM_STRUT_PARTIAL',
     '_NET_WM_PID',
 ]
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -1469,17 +1469,33 @@ def test_hints_setting_unsetting(manager):
 
 @manager_config
 def test_strut_handling(manager):
-    w = None
+    w = []
     conn = xcbq.Connection(manager.display)
 
     def has_struts():
         nonlocal w
-        w = conn.create_window(0, 0, 10, 10)
-        w.set_property("_NET_WM_STRUT", [0, 0, 0, 10])
-        w.map()
+        w.append(conn.create_window(0, 0, 10, 10))
+        w[-1].set_property("_NET_WM_STRUT_PARTIAL", [0, 0, 0, 10, 0, 0, 0, 0, 0, 0, 0, 800])
+        w[-1].map()
+        conn.conn.flush()
+
+    def with_gaps_left():
+        nonlocal w
+        w.append(conn.create_window(800, 0, 10, 10))
+        w[-1].set_property("_NET_WM_STRUT_PARTIAL", [820, 0, 0, 0, 0, 480, 0, 0, 0, 0, 0, 0])
+        w[-1].map()
+        conn.conn.flush()
+
+    def with_gaps_bottom():
+        nonlocal w
+        w.append(conn.create_window(800, 0, 10, 10))
+        w[-1].set_property("_NET_WM_STRUT_PARTIAL", [0, 0, 0, 130, 0, 0, 0, 0, 0, 0, 800, 1440])
+        w[-1].map()
         conn.conn.flush()
 
     def test_initial_state():
+        while manager.c.screen.info()["index"] != 0:
+            manager.c.next_screen()
         assert manager.c.window.info()['width'] == 798
         assert manager.c.window.info()['height'] == 578
         assert manager.c.window.info()['x'] == 0
@@ -1488,13 +1504,22 @@ def test_strut_handling(manager):
         bar = manager.c.window[bar_id].info()
         assert bar["height"] == 20
         assert bar["y"] == 580
+        manager.c.next_screen()
+        assert manager.c.window.info()['width'] == 638
+        assert manager.c.window.info()['height'] == 478
+        assert manager.c.window.info()['x'] == 800
+        assert manager.c.window.info()['y'] == 0
 
+    manager.test_xcalc()
+    manager.c.next_screen()
     manager.test_xcalc()
     test_initial_state()
 
     try:
+        while manager.c.screen.info()["index"] != 0:
+            manager.c.next_screen()
         manager.create_window(has_struts)
-        manager.c.window.static(0, None, None, None, None)
+        manager.c.window.static(None, None, None, None, None)
         assert manager.c.window.info()['width'] == 798
         assert manager.c.window.info()['height'] == 568
         assert manager.c.window.info()['x'] == 0
@@ -1504,8 +1529,18 @@ def test_strut_handling(manager):
         assert bar["height"] == 20
         assert bar["y"] == 570
 
+        manager.c.next_screen()
+        manager.create_window(with_gaps_bottom)
+        manager.c.window.static(None, None, None, None, None)
+        manager.create_window(with_gaps_left)
+        manager.c.window.static(None, None, None, None, None)
+        assert manager.c.window.info()['width'] == 618
+        assert manager.c.window.info()['height'] == 468
+        assert manager.c.window.info()['x'] == 820
+        assert manager.c.window.info()['y'] == 0
     finally:
-        w.kill_client()
+        for window in w:
+            window.kill_client()
         conn.finalize()
 
     test_initial_state()


tych0:
Thanks for the patch; can you add some tests for this behavior?

SpringerJack:
Yes I'll add some. Should I be worried about the last check failing?

tych0:
Yes, looks like all real errors:

```
   libqtile/core/manager.py:333: error: Function "builtins.chr" is not valid as a type
  libqtile/core/manager.py:333: note: Perhaps you need "Callable[...]" or a callback protocol?
  libqtile/core/manager.py:335: error: Value of type variable "SupportsLessThanT" of "max" cannot be "Union[int, None, Any]"
  libqtile/core/manager.py:335: error: Incompatible types in assignment (expression has type "Union[int, None, Any]", target has type "int")
  libqtile/core/manager.py:336: error: Unsupported operand types for + ("int" and "None")
  libqtile/core/manager.py:336: error: Unsupported operand types for + ("None" and "int")
  libqtile/core/manager.py:336: error: Unsupported left operand type for + ("None")
  libqtile/core/manager.py:336: note: Both left and right operands are unions
  libqtile/core/manager.py:337: error: Value of type variable "SupportsLessThanT" of "max" cannot be "Union[int, None, Any]"
  libqtile/core/manager.py:337: error: Incompatible types in assignment (expression has type "Union[int, None, Any]", target has type "int")
  libqtile/core/manager.py:338: error: Unsupported operand types for + ("int" and "None")
  libqtile/core/manager.py:338: error: Unsupported operand types for + ("None" and "int")
  libqtile/core/manager.py:338: error: Unsupported left operand type for + ("None")
  libqtile/core/manager.py:338: note: Both left and right operands are unions
  ```

File: libqtile/core/manager.py [hunk 1]
```diff
@@ -530,19 +537,51 @@ def current_window(self):
...
+                y = strut[2]
+            elif strut[3]:
+                y = self.x_screen_dimensions.get('y') - strut[2]
+            else:
+                y = ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
+            screen = self.find_screen(x, y)
+            assert (screen is not None)
+        except AssertionError:  # strut Values are defined but the coordinates are not to be found on any screen
+            logger.error("Relevant screen could not be found")
+            return
```
ep12, line 557: Maybe it is better to use a normal 
```py
if screen is None:
    logger.error("...")
    return
```
because it is shorter and also works with `python`s `-O` flag: [`assert` statements have no effect when running with `-O`](https://docs.python.org/3/reference/simple_stmts.html?highlight=assert#the-assert-statement). I don't know why python does it, but apparently it does.

ep12: Forget that and mark it as resolved...

tych0: I don't know, seems like a reasonable fix to me. That lets us drop a whole level of indentation too...

libqtile/core/manager.py
```diff
@@ -530,19 +539,56 @@ def current_window(self):
...
+                   screen.height is not None and screen.width is not None
+        except AssertionError:  # strut Values are defined but the coordinates are not to be found on any screen
+            logger.error("Relevant screen could not be found")
+            return
+        except IndexError:  # if _NET_WM_STRUT_PARTIAL is not set, we need to default to something
+            logger.warning("Not enough information to calculate the relevant screen. Defaulting to current.")
+            screen = self.current_screen  # defaulting to current_screen in lack of better options?
+        except AttributeError:
+            screen = self.current_screen
+            logger.warning("A")
```
ep12, line 567: May I suggest something like
```py
except AttributeError as e:
    logger.warning("%s. Defaulting to current screen.", e)
    screen = self.current_screen
```

SpringerJack: One might even go further and remove those lines again which seem to have forgotten. I thought about using this clause instead of the second assert statement. However I feel like the screen should not have None in any of those attributes so considering it just another failure to compute the screen seemed more reasonable to me.

libqtile/core/manager.py
```diff
@@ -530,19 +539,53 @@ def current_window(self):
...
+                y = strut[2]
+            elif strut[3]:
+                y = self.x_screen_dimensions.get('y') - strut[2]
+            else:
+                y = ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
+            screen = self.find_screen(x, y)
+            assert screen is not None
+            assert screen.x is not None and screen.y is not None and \
+                   screen.height is not None and screen.width is not None
+        except AssertionError:  # strut Values are defined but the coordinates are not to be found on any screen
```
tych0, line 559: I still think we can drop all the asserts in favor of if. It'll let you drop a level of indentation here.

libqtile/core/manager.py
```diff
@@ -530,19 +539,52 @@ def current_window(self):
...
+            elif strut[1]:
+                x = self.x_screen_dimensions.get('x') - strut[1]
+            else:
+                x = ((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)
+            if strut[2]:
+                y = strut[2]
+            elif strut[3]:
+                y = self.x_screen_dimensions.get('y') - strut[2]
+            else:
+                y = ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
```
SpringerJack, line 554: I am not too fond of this bulky statement, would you consider something like
```suggestion
            x = strut[0] or \
                (strut[8] + strut[9]) / 2 or \
                (strut[10] + strut[11]) / 2 or \
                self.x_screen_dimensions.get('x') - strut[1]
            y = strut[2] or \
                (strut[4] + strut[5]) / 2 or \
                (strut[6] + strut[7]) / 2 or \
                self.x_screen_dimensions.get('y') - strut[2]
```

to be more readable or do all of those or connections make things worse?

ep12: You suggested to switch the order of the `elif` and the `else` block. This changes the behaviour! I don't know the spec, but this exchange in the execution order looks suspicious, so I thought it'd be better to point it out. However, I have no clue which of the two (different!) implementations is in agreement with the spec. If it is the original bulky block, an alternative would be
```py
x = (strut[0] if strut[0] else
     (self.x_screen_dimensions.get('x') - strut[1] if strut[1] else
      ((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)))
y = (strut[2] if strut[2] else
     (self.x_screen_dimensions.get('y') - strut[3] if strut[3] else
      ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)))
```
or 
```py
x = (strut[0] if strut[0]
     else (self.x_screen_dimensions.get('x') - strut[1] if strut[1]
           else ((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)))
y = (strut[2] if strut[2]
     else (self.x_screen_dimensions.get('y') - strut[3] if strut[3]
           else ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)))

```
with the indentation shifted to the right of course.
Also, in line 545-546 you use index `1` two times, and in 551-552 you use `3` and `2` where the rest of the lines is similar. Is that correct?

SpringerJack: As far as I can see the order should not matter as any of the values, if set to something other than zero, give an idea of on-screen coordinates. A problem would be one window setting struts on different screens but if that happens neither implementation is correct.
Good catch with the indices, I agree line 551 should be `y = self.x_screen_dimensions.get('y') - strut[2]`

libqtile/core/manager.py
```diff
@@ -330,6 +330,15 @@ def _process_screens(self) -> None:
...
             self.screens.append(scr)
 
+        self.x_screen_dimensions: Dict[str, int] = {'x': 0, 'y': 0}
+        for screen in self.screens:
+            if screen.x is not None is not screen.width and \
+                    screen.y is not None is not screen.height:
+                self.x_screen_dimensions['x'] = max(self.x_screen_dimensions.get('x') or 0,
+                                                    screen.x + screen.width)
+                self.x_screen_dimensions['y'] = max(self.x_screen_dimensions.get('y') or 0,
+                                                    screen.y + screen.height)
```
m-col, line 340: This calculates the X screen to be the smallest rectangle that fits all of the enabled outputs inside it. Is that always the case?

Also perhaps you can ask the X server for this information direction... try `qtile.root.get_geometry()`.

m-col: Another small comment - could you change "x" and "y" to width and height, if that is what they represent? This would keep it consistent with other uses of these word in the code and avoid confusion

m-col: If we only need this information ready when `add_strut` is called, maybe it can be move into that methods local scope

libqtile/core/manager.py
```diff
@@ -530,19 +539,52 @@ def current_window(self):
...
+            else:
+                x = ((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)
+            if strut[2]:
+                y = strut[2]
+            elif strut[3]:
+                y = self.x_screen_dimensions.get('y') - strut[2]
+            else:
+                y = ((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
+            screen = self.find_screen(x, y)
+        except IndexError:  # if _NET_WM_STRUT_PARTIAL is not set, we need to default to something
```
m-col, line 556: Is the idea that we are just falling back to the short version upon any indexing failure into `strut`? If so, it might be logically neater if the `try` block is replaced with a test of `len(strut)`, and may be easier to debug if needed in the future

libqtile/core/manager.py
SpringerJack: Using root.get_geometry() seems to be the smarter way, I just did not know it existed. I moved it into add_strut(), otherwise I'd have preferred it to remain in _process_screens() to not reiterate again and again.

libqtile/core/manager.py
```diff
@@ -530,19 +530,54 @@ def current_window(self):
...
+                y = abs(strut[2])
+            else:  # bottom
+                x = abs((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)
+                y = x_screen_dimensions.height - abs(strut[3])
+            screen = self.find_screen(x, y)
+        else:
+            logger.warning("Not enough information to calculate the relevant screen. Defaulting to current.")
+            screen = self.current_screen
+
+        if screen is None or screen.x is None or screen.y is None or \
```
m-col, line 553: Looks like a job for `not all()` :)

m-col:
The changes look great, thanks! For the test, you could probably just edit `test_strut_handling` in test_manager.py to also try struts with the full 12-element strut data. Other than that, please could you squash the commits (including the test changes) into a single commit to keep the git history clean?

SpringerJack:
Unit tests now consider the most interesting cases. I'm no expert on this, but maybe one of you knows some more: How widely used is NET_WM_STRUT on it's own? I realized that always defaulting to current_screen in that case can really lead to trouble as soon as a static window is killed as the struts may get "removed" on the wrong screen. If it still is commonly used we might want to set the windows screen attribute correctly and then pass that information on.

libqtile/core/manager.py
SpringerJack: Python really has some neat features. Going with `any()` even allows to avoid a `not not ` construction.

ep12:
> Unit tests now consider the most interesting cases. I'm no expert on this, but maybe one of you knows some more: How widely used is NET_WM_STRUT on it's own? I realized that always defaulting to current_screen in that case can really lead to trouble as soon as a static window is killed as the struts may get "removed" on the wrong screen. If it still is commonly used we might want to set the windows screen attribute correctly and then pass that information on.

Thanks for pointing that out. Actually, I have no clue how common that is. When it is merged we'll see if that current behaviour leads to trouble :)

tych0:
> How widely used is NET_WM_STRUT on it's own?

What do you mean "on it's own"? My understanding are that STRUTs are mostly used by third party bars like polybar or taffybar, but not generally by X programs.

> If it still is commonly used we might want to set the windows screen attribute correctly and then pass that information on.

Seems like this would be easy enough to do, or am I missing something?

SpringerJack:
The spec for NET_WM_STRUT states:

>  [...] clients MAY set this property in addition to _NET_WM_STRUT_PARTIAL to ensure backward compatibility with Window Managers supporting older versions of the Specification. 

thus it would seem to me that _NET_WM_STRUT_PARTIAL should always be set alongside with it.  can only arise, if a client chooses to only adhere to an older version and not specify  _NET_WM_STRUT_PARTIAL. 
Using the screen attribute adds a risk for unwanted side effects, that's why I was reluctant.

m-col:
TBH any software worth anything will definitely use the mots recent version. Also it looks neither i3 nor bspwm support `_NET_WM_STRUT`. Maybe we should just drop all support for `_NET_WM_STRUT` and only respond to `_NET_WM_STRUT_PARTIAL`?

SpringerJack:
I'd prefer dropping it. If at any point only `_NET_WM_STRUT` is set the user can still add a gap manually which would still make more sense to me than defaulting to a screen.

tych0:
Sounds good to me.

libqtile/window.py
```diff
@@ -727,12 +727,23 @@ def update_strut(self):
...
+                x = abs(strut[0])
+                y = abs((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
+            elif strut[1]:  # right
+                x = x_screen_dimensions.width - abs(strut[1])
+                y = abs((strut[4] + strut[5]) / 2 or (strut[6] + strut[7]) / 2)
+            elif strut[2]:  # top
+                x = abs((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)
+                y = abs(strut[2])
+            else:           # bottom
+                x = abs((strut[8] + strut[9]) / 2 or (strut[10] + strut[11]) / 2)
```
m-col, line 742: What happens when a window sets the property incorrectly and doesn't provide enough values here? I'm guessing that the `IndexError` is caught by the eventloop, logged and qtile continues on as before. Do you know?

SpringerJack: I just forced an `IndexError` to occur and for me that results in qtile continuing, obviously without the static window, and logging the exception.

m-col: Ah, nice.

libqtile/core/manager.py
```diff
@@ -527,25 +527,40 @@ def current_group(self):
...
+            logger.error("Missing screen information")
+            return
+
+        x_screen_dimensions = self.root.get_geometry()
+        gaps = [
+            screen.x,
+            x_screen_dimensions.width - screen.x - screen.width,
+            screen.y,
+            x_screen_dimensions.height - screen.y - screen.height,
+        ]
```
m-col, line 543: As you've moved some of the calculations into `class Static`, maybe calculate `gaps` there too and pass `screen` and `gaps` to `Qtile.add_strut`. This avoid doing `x_screen_dimensions = self.root.get_geometry()` in two places.

It also appears that if `qtile.find_screen` returns `None` then line 533 will get an `AttributeError` when trying to get `screen.x`. Maybe upon return of that function only continue the calculations and call to `add_strut` if `screen is not None`.

A nice side benefit of this is that these calculations, which are specific to X's strut hints, are separated from the more general "use `gaps` to update this screen's `Gap`s" which in theory could be used as-is with any backend (just thinking long-term with the come of Wayland).

libqtile/core/manager.py
```diff
@@ -527,25 +527,40 @@ def current_group(self):
     def current_window(self):
         return self.current_screen.group.current_window
 
-    def add_strut(self, strut):
+    def add_strut(self, strut, screen):
         from libqtile.bar import Bar, Gap
 
+        if any(e is None for e in [screen, screen.x, screen.y, screen.height, screen.width]):
```
m-col, line 533: This is just a bit easier to read:
```suggestion
        if not all([screen, screen.x, screen.y, screen.height, screen.width]):
```
or even:
```suggestion
        if None in [screen, screen.x, screen.y, screen.height, screen.width]:
```

m-col:
This might resolve https://github.com/qtile/qtile/issues/1902 as well, which may be a duplicate of #1947

libqtile/core/manager.py
ep12: `if None in [...]` is probably more explicit. If `screen.x == 0`, then `not all([...])` would be `True = not False` whereas `None` might not be in `[...]`. We probably don't want to return if one item is `0`, right?

m-col: Ah yep good spot, `if None in` is definitely the way to go.

SpringerJack: Yes, the second one looks better.

libqtile/core/manager.py
SpringerJack: Moving the gaps calculations into `class Static`, would you suggest storing the list in a new attribute `self.gaps` for instances of Static? We will need the values again to remove struts and this call is not done from within `update_strut`.

Is there a way to stop evaluating the list as soon as a `None` value is found?

m-col: The `self.strut` attribute already stores the window property from `_NET_WM_STRUT_PARTIAL` so maybe it makes sense to store the calculated version there instead of the window property. We'd need to re-process it when this attribute is used for `remove_strut` anyway so this way it's already done. I think that makes sense?

> Is there a way to stop evaluating the list as soon as a None value is found?

Which bit are you referring to?

m-col:
Please could you also update these lines to remove the non-partial form:
 - https://github.com/qtile/qtile/blob/303e120ff2ab5519264866170e81ab3e5e5a04cf/libqtile/window.py#L751
 - https://github.com/qtile/qtile/blob/16fe9ce0d7b4216fb00a7b39a87331a9a6741482/libqtile/backend/x11/xcbq.py#L204

libqtile/core/manager.py
SpringerJack: Yes, storing it probably would make sense. Do we then actually need the pure values of `_NET_WM_STRUT_PARTIAL` in `self.strut`? I think I'd like to store just one global attribute, something like `self.reserved_space`, which we generate in `update_strut` from the local `strut` and `gaps`? This way we'd only need to pass two arguments.

SpringerJack: > Is there a way to stop evaluating the list as soon as a None value is found?
I'm referring to the `AttributeError`. I am used to such evaluations being lazy and therefore not checking as far as `screen.x` if `screen` itself is None.

m-col: > Do we then actually need the pure values of _NET_WM_STRUT_PARTIAL in self.strut?

I don't think so. I think your approach with saving gaps in `self.reserved_space`\* and removing `self.strut` is a good idea. I guess that and `self.screen` is sufficient for removal, right?

For the `AttributeError` -- I think dropping `screen` from that list altogether would be fine. Instead you could check that the return value of `qtile.find_screen` is not `None` and only call `qtile.add_strut` if it is alright. This way, `qtile.add_strut` can just always except a Screen object.

\* Calling it reserved_space is actually really nice, maybe we should rename the methods that refer to "strut" to something similar, which would probably make things more explicit and clear. For instance `Qtile.add_strut` could definitely be improved by renaming it to `Qtile.reserve_space` or something. Likewise `Qtile.unreserve_space` or something to remove them. Anyway this is unnecessary so don't feel like you need to make those changes!

SpringerJack: > This way, `qtile.add_strut` can just always except a Screen object.

Moving the `gaps` logic to `class Static` introduces the need to already check the whole list `update_strut`.

m-col: Ah sorry, bad wording, that's what I meant :)

libqtile/window.py
```diff
@@ -727,17 +727,45 @@ def update_strut(self):
...
+                logger.error("Missing screen information")
+                return
+
+            empty_space = [
+                self.screen.x,
+                x_screen_dimensions.width - self.screen.x - self.screen.width,
+                self.screen.y,
+                x_screen_dimensions.height - self.screen.y - self.screen.height
+            ]
+            self.reserved_space = [strut[i] - empty if strut[i] else 0 for i, empty in enumerate(empty_space)]
```
m-col, line 759: I'm finding it a bit hard to see what's going on here. What are the values in `empty_space`?

The rest of the PR look great, thanks

SpringerJack: `empty_space` is what used to be called `gaps` in earlier commits, I changed the name as the similarity with `bar.Gap` started to bug me. It contains dimensions of empty areas on the logical screen as strut values use absolute coordinates, but to set the size of `bar.Gap` correctly we need coordinates relative to physical screen edges. Basically those values are the sole reason for this PR :). Maybe there is a more suggestive name for this?

m-col: Sorry if I'm being a bit slow here :p  I had to read it a few times to understand what it's doing but I think I get it. So you're starting off with just the x and y values of the edges and then adjusting them using `strut`. I wonder if we can write it so that is more obvious just from a glance, just to avoid confusion in the future. Was it different before? I don't remember getting tripped up on something like this before lol

m-col: So what I came up with is this. If I understood it correct, then I think we don't have to handle a strut value being 0 any differently. This means we can just do this:
```suggestion
            self.reserved_space = [
                strut[0] - self.screen.x,
                strut[1] - (x_screen_dimensions.width - self.screen.x - self.screen.width),
                strut[2] - self.screen.y,
                strut[3] - (x_screen_dimensions.height - self.screen.y - self.screen.height),
            ]
```
The reason for this is that `strut[0]` will be `0` when a window wants its left edge to be at x == 0 in X screen coods, which I guess could still happen if that window "belongs" to a physical screen whose X coord is not 0. This would be like reserving negative space (if it was attached to that other output... in reality the X server doesn't really bind windows to a specific output). Another example would be if a window that is on an output with its X coordinate at 1920 set `strut[0]` to 1920, then it wants to reserve no extra space i.e. it would be up against that output's left physical edge. In that case, if `strut[0]` actually equalled 0 then it actually wants to be on a different screen than it is.

Anyway I hope my description make sense and I hope that I have understood things correctly! Please correct me if I have misunderstood how the data is encoded here.

SpringerJack: > strut[0] will be 0 when a window wants its left edge to be at x == 0 in X screen coods

I'm afraid this interpretation is at odds with the ewmh spec. `strut[0:4]` contains
> 4 cardinals specifying the width of the reserved area at each border of the screen

A value of `0` would thus just mean "no reservation for this cardinal". 
In the examples only one cardinal gets specified, which would be insufficient otherwise. Also I don't think the ending remark 

> A "corner" panel that does not extend for the full length of a screen border SHOULD only set one strut.

would not make sense if an unset strut had a meaning apart from "nothing to do here".

SpringerJack: > Was it different before?

During this PR the list has only changed it's name, but as for the latest release the pure strut values are passed on to be reserved possibly resulting in a too large `bar.Gap`

m-col: > A value of 0 would thus just mean "no reservation for this cardinal".

so a value of 0 for the left hand side would say "put my left edge against the far left end of the x screen", right? Or is this not the same as "don't reserve me space" (I guess this only holds true if the remaining "partial" cardinals aren't factored in i.e. the bar reserves the entire edge).

m-col: ^ This would be for a horizontal bar on the top or bottom edge

SpringerJack: Ah, wait. Are we talking about the first 4 cardinals or the last 8? I've been referring to the first 4 the whole time. 

> if the remaining "partial" cardinals aren't factored in i.e. the bar reserves the entire edge

Those partial cardinals are only considered when looking for the correct physical screen. For all I know there is no `bar.Gap` covering only parts of an edge, right?

m-col: Yeah sorry that's what I meant -- I also meant the first four, just in that hypothetical where a window reserves a band of space on one edge (a band that covers the entire length of that edge), the ends of the reserved area will touch the two adjecent perpendicular edges.

Let's say we have three outputs side-by-side (1, 2 & 3; each are 1000 pixels wide) and we open a window on the middle output and it wants to reserve 20 pixels at the top of that middle output, to look like this:
```
+-------+-------+-------+
|       +-------+       |
|       |       |       |
|       |       |       |
|   1   |   2   |   3   |
+-------+-------+-------+

```
In output-relative coordinates, the window wants to reserve `[0, 0, 20, 0]` (L, R, T, B). What would the strut values provided by the window be?

SpringerJack: In your example 1, 2 & 3 all have the same height either, right? That would then result in `[0, 0, 20, 0, 0, 0, 0, 0, 1000, 2000, 0, 0]`.
But suppose that 1 & 3 both are 2000px tall, screen 2 just 1800. The three of them are aligned on their bottom edges. Here we'd get [0, 0, 220, 0, 0, 0, 0, 0, 1000, 2000, 0, 0] for a third party bar with the size of 20 pixels.

m-col: Ahhhh OK, I see, gotcha. Great, thanks for bearing with me!

In that case it all looks good. Was there anything else left to do on the PR?

SpringerJack: Glad I could help. No, I don't think I had anything else in mind.

m-col:
Brilliant, thanks again for this.

PR #2322, opened 2021-03-24
Add `xkb-switch` as optional dependency to KeyboardLayout widget.

### LINKED ISSUE #2027: KeyboardLayout widget doesn't work as expected

MatejaMaric:
I made `use_xkb_switch` option that when enabled should fix #2027 and also makes KeyboardLayout widget easier to use.
When disabled everything should work as it did in previous versions of KeyboardLayout widget, making it backwards compatible.

Changes:

- Use `xkb-switch` in `next_keyboard`, if `use_xkb_switch` is set to True.
- Use `xkb-switch` in `keyboard` getter and sanitize it's output, if `use_xkb_switch` is set to True.
- `keyboard` setter is never used if `use_xkb_switch` is set to True.
- Lower default `update_interval` for better responsiveness.
- Rewrote widget documentation.
- Updated CHANGELOG.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,7 @@ Qtile 0.x.x, released xxxx-xx-xx:
           A new command, `lazy.ungrab_all_chords`, was introduced to return to the root bindings.
           The `enter_chord` hook is now always called with a string argument.
           The third argument to `KeyChord` was renamed from `submaping` to `submapping` (typo fix).
+        - Add `xkb-switch` as optional dependency to KeyboardLayout widget for improved functionality.
 
 Qtile 0.17.0, released 2021-02-13:
     !!! Python version breakage !!!
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2014-2015 Sean Vig
 # Copyright (c) 2014 Tycho Andersen
 # Copyright (c) 2019 zordsdavini
+# Copyright (c) 2021 Mateja Maric
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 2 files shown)


tych0:
Is there a way we can accomplish this same thing just using setxkbmap?

MatejaMaric:
@tych0 Not that I know of, `setxkbmap -query` can show configured keyboard layouts, but not one currently in use.

tych0:
Urgh, yeah, https://unix.stackexchange.com/questions/12072/how-do-i-get-current-keyboard-layout

Seems like there is not a good way besides xkb directly.

ep12:
If I knew how to use `cffi` I would give it a shot. I mean, we have the pulse volume widget that glues C and python together. Shouldn't it be possible to use the X extension directly?

MatejaMaric:
I think #2323 is a much better fix for #2027 than this pull request.

tych0:
Can you explain why? :-)

MatejaMaric:
@tych0 Status bars that most people use have keyboard widget that don't require any configuration or dependencies like `xkb-switch`.
They auto refresh and are implemented in compiled languages making them quite performant.
#2323 has potential to provide this "plug and play" experience.
In my opinion, it is a better solution even if it has to use xlib instead of xcb, given that xlib is available everywhere. Of course if you can implement xkb generated wrappers for xcffib, you should, but if not I don't think it is a big deal if xlib is used.

ep12:
> I think #2323 is a much better fix for #2027 than this pull request.

> Can you explain why? :-)

Also, it isn't a complete fix yet... Apart from using Xlib instead of xcb the actual work has yet to be done, but this should not be too far away. Except: xcb.

tych0:
> In my opinion, it is a better solution even if it has to use xlib instead of xcb

Unfortunately, it also introduces a completely new dependency and way of communicating with the X server, including all the traditional perils of using xlib. This will only work if we can actually maintain it, which I'm not convinced we can.

MatejaMaric:
@tych0 Back when I was using DWM, I used a shell script for displaying keyboard layout that would run [this program](https://github.com/bblackcblackc/getxkblayout/blob/master/getxkblayout.c) 5 times per second. I don't know hard would it be to implement or maintain Python interface for such a small program.

tych0:
> I don't know hard would it be to implement or maintain Python interface for such a small program.

It's not really about the smallness or largeness of the code. It's more about an additional dependency and the cognitive load (and user complaints about additional linking requirements, etc.) that comes with it. Let's just fix xcffib if people want to do this kind of thing directly, it will ultimately be less work in the long run.

ep12:
> @tych0 Back when I was using DWM, I used a shell script for displaying keyboard layout that would run [this program](https://github.com/bblackcblackc/getxkblayout/blob/master/getxkblayout.c) 5 times per second. I don't know hard would it be to implement or maintain Python interface for such a small program.

This is approximately what #2323 is doing except that [the `x.split(',')` is done in python](https://github.com/qtile/qtile/pull/2323/files#diff-6dd43400199825bb5fc799621b975877a337ee3974fbeca84827ff08ab09a029R116-R120). I think @tych0 would be happy if we could do the same with xcb instead of xlib until xcffib supports the keyboard extension right away if I understood him correctly. Sadly, I don't think I can work on xcffib as I couldn't get it to work and then writing haskell code (which I have no idea about) to generate the bindings to a library (xcb) I know nothing about for a module xcffib I don't know how to use is probably a bit too much.

PR #2323, opened 2021-03-24
Add a keyboard helper `cffi` module to fix #2027.

### LINKED ISSUE #2027: KeyboardLayout widget doesn't work as expected

ep12:
Fixes: #2027, #2265.
Alternative to #2322, not ready yet.  [_Relatively dependency-free_](https://github.com/qtile/qtile/issues/2265#issuecomment-786931988) as @m-col would put it.

This pr is just a first draft that adds the required `cffi` module.
The actual fix has yet to be done.
Remaining to-dos are:
+ [ ] unit tests
+ [x] setup/packaging integration (`setup.py`) (sort of done??)
+ [x] make sure that the function to enable listening for `Xkb` events
      actually works.
+ ...

To be honest, I am not a C programmer, have absolutely no clue about how X works or even Xkb and have never used `cffi` in my life until now. Most of the code is _inspired_ by [`setxkbmap`](https://gitlab.freedesktop.org/xorg/app/setxkbmap/-/blob/master/setxkbmap.c) and [`xkb-switch`](https://github.com/grwlf/xkb-switch/blob/master/src/XKeyboard.cpp) (to avoid the word _stolen_).
But I _think_ that the first commit already adds a lot of the functionality we need to actually get the currently active keyboard layout in a setup like in #2027 with `setxkbmap us,ru,...`.
The `keyboard_helper.py` module provides `connect_to_display`, which must be called first, e.g. `connect_to_display(b':0')`. Then we can use `get_configured_layouts()` to get a list of layouts. Finally, the active layout is the one with the index `get_group_index()`. We can also call `set_group_index(group_num: int)` to switch to the next layout. There seems to be a C function to enable listening to Xkb events (like when switching the layout externally). We want to listen for those events to update the widget when that happens (see #2027). It would be good if somebody else who actually knows C and X11 and [Xkb](https://www.x.org/releases/X11R7.7/doc/libX11/XKB/xkblib.html) could say something about that.

### Rationale 
`xkb-switch` is not in the official arch repos, so on other distros it probably needs to be installed from source. But since we already use `cffi` and Xkb is probably also available, it would make sense not to have an additional dependency for that.
Side remark: I don't fully understand how `setxkbmap` sets the layouts, so I can't help to make that dependency-free.


tych0:
I would really prefer to avoid introducing xlib code to our code base, in favor of sticking with xcb. I realize that means more low level binding work up front, but I think it will be much easier to maintain going forward.

ep12:
> I would really prefer to avoid introducing xlib code to our code base, in favor of sticking with xcb. I realize that means more low level binding work up front, but I think it will be much easier to maintain going forward.

Okay, whatever you prefer. Do you think it would be easy to rewrite this with `xcb` instead of `xlib`? Where would I find documentation for that? Would [this](https://xcb.freedesktop.org/manual/group__XCB__xkb__API.html) the page to start looking? Or are there more practical examples? Just to clarify: I don't really have a clue how it all works. Bare with me if I cannot say a single thing about `xlib` vs `xcb`...

tych0:
Currently xcffib doesn't have xkb generated wrappers. I forget exactly what the problems were, but the first step would be to hack the xcffib generator code to generate the xkb bindings and go from there. I forget exactly what the problems were, but there may need to be some fixes in xcb-types as well.

ep12:
> Currently xcffib doesn't have xkb generated wrappers. I forget exactly what the problems were, but the first step would be to hack the xcffib generator code to generate the xkb bindings and go from there. I forget exactly what the problems were, but there may need to be some fixes in xcb-types as well.

Okay, it would be great if the necessary interface would be right through `xcffib` without having to write a separate module for the `Xkb` extension! I tried to get my clone of `xcffib` to work but it threw weird errors at me. No luck so far.
What about this pr? Should I close it?

tych0:
Up to you. We generally don't close PRs in case people find them useful, but I don't think this will go in as-is.

ep12:
> Up to you. We generally don't close PRs in case people find them useful, but I don't think this will go in as-is. 

Okay, I'll leave it open for now and then implement the actual fix for #2027 to get a working proof of concept with Xlib. After that I might attempt to figure out how xcb works and then port everything...

PR #2324, opened 2021-03-25
Add a new command to label the current group on the flight

ramnes:
Entering an empty value will restore the original name of the group.


tych0:
Any chance you can add a test for this? :)

ramnes:
Yeah right, done!

I also added a second commit, but this one is way harder to test since qtile restarts are not correctly handled by our test manager at the moment. I tried to change that but it's quite the spaghettis and I can't take more time right now. We don't test anything related to restarts so far, so maybe it's acceptable to let this one line diff pass this time? :)))

m-col (approved):

Issue #2311, opened 2021-03-20
CheckUpdate widget add line strip to custom_command? [feature request]


yobleck:
# Issue description
The custom_command doesn't have the line strip option that the built in cmd_dict has.

self.sutr at:https://github.com/qtile/qtile/blob/master/libqtile/widget/check_updates.py#L62 reads from the dict,
but on:https://github.com/qtile/qtile/blob/master/libqtile/widget/check_updates.py#L79 it just says zero.

It would be nice if the custom_command allowed for a tuple format. 
For example on Manjaro `custom_command=("pamac checkupdates", 3),` would make sense.

example code change starting at:https://github.com/qtile/qtile/blob/master/libqtile/widget/check_updates.py#L77
```
else:
    updates = self.call_process(self.custom_command[0], shell=True);
    self.subtr = self.custom_command[1];
```

some error checking to ensure compatibility with previous versions might be necessary. I'm not comfortable making the pull request my self since I'm new to git, but hopefully I'm not out of line to post here.

# Qtile version
0.17.0

ep12:
Thanks for opening that issue! It is definitely an option that should be configurable since `custom_command` already exists. If you really don't want to create a pr on your own, I could do that. But feel free to use this small issue as a chance to learn git! I am relatively new, too, but the most important bits are learned quickly. I personally played through https://learngitbranching.js.org which still sometimes comes in handy to check the commands in a sandbox before doing anything wrong. Tell us/me how you like to proceed and we'll help you with that.

yobleck:
I guess I'll take a crack at making a pr.

My first question his how should I handle backwards compatibility?
If I make custom_command take a two tuple as an argument I could add a check to convert strings into the first item in the tuple with the second value defaulting to 0.
Or I could make a new command like custom_command_line_strip which takes integer values (and defaults to 0).
The latter option would be easier to implement and have better backwards compat but would be messier for the user.
The former option would look cleaner but would require another update at some point to remove backwards compat.

yobleck:
whoops didn't mean to do that

ramnes:
> If I make custom_command take a two tuple as an argument I could add a check to convert strings into the first item in the tuple with the second value defaulting to 0.

LGTM

ep12:
> The former option would look cleaner but would require another update at some point to remove backwards compat.

Maybe not even that, because the default value of `0` might be fine for many users. I agree with you and ramnes that the first option is probably preferable.  
[...]

PR #2325, opened 2021-03-25
Add custom_command output line count modification argument to CheckUpdate widget

yobleck:
Changes the custom_command argument input from string to tuple or list ("string", int)
mimicking the format of the distro argument allowing for line stripping of package manager update commands.
For example: `custom_command = "pacman -Qu"` (while still acceptable for backwards compatibility)
can now be written as `custom_command = ("pacman -Qu", 0)` where 0 can be any integer of lines to be removed.

checking of whether to use distro or custom_command and setting of internal values was also moved to \_init\_ so that the custom command doesn't have to be reparsed every time it is called by poll().

instigating issue: #2311 

Edit: `custom_command` reverted to string format and new `custom_command_modify` argument added that allows user to input lambda function for greater flexibility. Note new argument is overridden by `distro` argument

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -31,6 +31,7 @@ class CheckUpdates(base.ThreadPoolText):
     defaults = [
         ("distro", "Arch", "Name of your distribution"),
         ("custom_command", None, "Custom shell command for checking updates (counts the lines of the output)"),
+        ("custom_command_modify", (lambda x: x), "Lambda function to modify line count from custom_command"),
         ("update_interval", 60, "Update interval in seconds."),
         ('execute', None, 'Command to execute on click'),
         ("display_format", "Updates: {updates}", "Display format if updates available"),
@@ -56,15 +57,20 @@ class CheckUpdates(base.ThreadPoolText):
                          "Mandriva": ("urpmq --auto-select", 0)
                          }
 
-        # Check if distro name is valid.
-        try:
-            self.cmd = self.cmd_dict[self.distro][0].split()
-            self.subtr = self.cmd_dict[self.distro][1]
-        except KeyError:
-            distros = sorted(self.cmd_dict.keys())
-            logger.error(self.distro + ' is not a valid distro name. ' +
-                         'Use one of the list: ' + str(distros) + '.')
-            self.cmd = None
+        if self.custom_command:
+            # Use custom_command
+            self.cmd = self.custom_command
+
+        else:
+            # Check if distro name is valid.
+            try:
+                self.cmd = self.cmd_dict[self.distro][0]
+                self.custom_command_modify = (lambda x: x - self.cmd_dict[self.distro][1])
+            except KeyError:
+                distros = sorted(self.cmd_dict.keys())
+                logger.error(self.distro + ' is not a valid distro name. ' +
+                             'Use one of the list: ' + str(distros) + '.')
+                self.cmd = None
 
         if self.execute:
             self.add_callbacks({'Button1': self.do_execute})
@@ -72,14 +78,10 @@ class CheckUpdates(base.ThreadPoolText):
     def _check_updates(self):
         # type: () -> str
         try:
-            if self.custom_command is None:
-                updates = self.call_process(self.cmd)
-            else:
-                updates = self.call_process(self.custom_command, shell=True)
-                self.subtr = 0
+            updates = self.call_process(self.cmd, shell=True)
         except CalledProcessError:
             updates = ""
-        num_updates = len(updates.splitlines()) - self.subtr
+        num_updates = self.custom_command_modify(len(updates.splitlines()))
 
         if num_updates == 0:
             self.layout.colour = self.colour_no_updates
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,9 @@ Qtile 0.x.x, released xxxx-xx-xx:
           A new command, `lazy.ungrab_all_chords`, was introduced to return to the root bindings.
           The `enter_chord` hook is now always called with a string argument.
           The third argument to `KeyChord` was renamed from `submaping` to `submapping` (typo fix).
+        - added new argument for CheckUpdates widget: `custom_command_modify` which allows user to modify the
+          the line count of the output of `custom_command` with a lambda function (i.e. `lambda x: x-3`).
+          Argument defaults to `lambda x: x` and is overridden by `distro` argument's internal lambda.
 
 Qtile 0.17.0, released 2021-02-13:
     !!! Python version breakage !!!


libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,49 @@ def __init__(self, **config):
...
+                self.cmd = None
+                self.subtr = 0
+        else:
+            # Otherwise use custom_command
+            try:
+                # logger.warning("checking updates with custom_command")
+                self.cmd = self.custom_command[0]  # .split()
+                self.subtr = self.custom_command[1]
+                # logger.warning(self.cmd)
+            except:
```
tych0, line 85: Let's not use bare except:s here, and just let the stack trace happen. The qtile logging framework will take care of catching it and making sure it doesn't cause any problems.

libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,49 @@ def __init__(self, **config):
...
-        try:
-            self.cmd = self.cmd_dict[self.distro][0].split()
-            self.subtr = self.cmd_dict[self.distro][1]
-        except KeyError:
-            distros = sorted(self.cmd_dict.keys())
-            logger.error(self.distro + ' is not a valid distro name. ' +
-                         'Use one of the list: ' + str(distros) + '.')
-            self.cmd = None
+        # Backwards compatibility for custom_command. Convert string to tuple.
+        if type(self.custom_command) is not tuple:  # Maybe remove this if statement after deprecation time period expires?
```
tych0, line 60: Please introduce a migration for this.

yobleck: Sorry I'm new to git related stuff so I'm sure what you mean by migration. Are you referring to the 'libqtile.widget.base.deprecated' mentioned in the docs?

ep12: We now have a command [`qtile migrate`](https://github.com/qtile/qtile/blob/master/libqtile/scripts/migrate.py) which attempts to fix a config that got broken by changes like this. You'd want to do something like the migration for `Tile` which looks like this:
```py
def tile_master_windows_rename(config):
    return (
        bowler.Query(config)
        .select_function("Tile")
        .modify_argument("masterWindows", "master_length")
    )
```
Unfortunately, we don't have an example of how to modify an argument to a tuple where the first item is the original. Hence it is time to search for this in the `bowler` documentation.
Also, I am not sure if we want to make this mandatory. For many users the default `subtr` has not been a problem (I think?) so I think a check if it is a tuple and then adding the 0 in the `__init__` would be enough.

libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,44 @@ def __init__(self, **config):
...
-        try:
-            self.cmd = self.cmd_dict[self.distro][0].split()
-            self.subtr = self.cmd_dict[self.distro][1]
-        except KeyError:
-            distros = sorted(self.cmd_dict.keys())
-            logger.error(self.distro + ' is not a valid distro name. ' +
-                         'Use one of the list: ' + str(distros) + '.')
-            self.cmd = None
+        # Backwards compatibility for custom_command. Convert string to tuple.
+        if type(self.custom_command) is not tuple:
```
ep12, line 60: `isinstance(self.custom_command, (list, tuple))` might be better here. If the user passes `['pamac checkupdates', 42]` we would get `self.custom_command = (['pamac checkupdates', 42], 0)` in the next line.

libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,44 @@ def __init__(self, **config):
...
+            logger.warning("string inputs to custom_command are deprecated." +
+                           "please use tuple: (\"command\",int(lines to strip from output))")
+
+        if not self.custom_command[0]:
+            # Check if distro name is valid.
+            try:
+                # logger.warning("checking updates with distro")
+                self.cmd = self.cmd_dict[self.distro][0]  # .split()
+                self.subtr = self.cmd_dict[self.distro][1]
+                # logger.warning(self.cmd)
```
ep12, line 71: Can you remove those comment comment lines starting with `# logger.*` later on?

libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,44 @@ def __init__(self, **config):
...
-                updates = self.call_process(self.custom_command, shell=True)
-                self.subtr = 0
-        except CalledProcessError:
-            updates = ""
+            # logger.warning("running check update cmd")
+            updates = self.call_process(self.cmd, shell=True)
+            # logger.warning(str(updates))
+        except CalledProcessError as e:
+            # logger.warning(e.output)
+            updates = e.output.decode()
```
ep12, line 96: I do have to pick up the discussion from #2311 here:
`pamac checkupdates` _always_ has a return code of 100. This is annoying because for all other _check for updates_ commands a non-zero exit code indicates an error. This way we might actually count the line length of error messages (don't know if one of the programs emits an error message on `stdout` though). The only alternative I could think of would be to add yet another argument, the expected return code, defaulting to 0. Then we could just do
```py
res = subprocess.run(self.cmd, shell=True, capture_output=True)
if res.returncode == self.expected_return_code:
    updates = result.stdout.decode()
else:
    logger.error('checking for updates with %r failed:\n%s\n%s', self.cmd, res.stderr.decode(), res.stdout.decode())
    updates = ''
```
What do you guys think?

libqtile/widget/check_updates.py
ep12: Okay @tych0 @yobleck. Here is a minimal partially working migration:
<details><summary><code>test_class.py</code>, approx. <code>libqtile.widget.checkupdates</code></summary>
<pre>
class CheckUpdates:
    def __init__(self, *, custom_command = (None, 0)):
        self.custom_command = custom_command
</pre>
</details>
<details><summary><code>test.py</code>, approx. <code>~/.config/qtile/config.py</code></summary>
<pre>
from test_class import CheckUpdates
CheckUpdates(custom_command='pamac checkupdates')
CheckUpdates(custom_command=' '.join(['pamac', 'checkupdates']))
CheckUpdates(custom_command=('pamac checkupdates' if HOSTNAME == 'xy' else None))
CheckUpdates(custom_command=('pamac checkupdates', 42))
</pre>
</details>

<details><summary><code>fix.py</code>, approx. <code>qtile migrate</code></summary>
<pre>
import bowler
from bowler.types import Leaf, Node, SYMBOL
from fissix.pygram import token
def my_modifier(node, capture, filename):
    if 'class_call' not in capture:
        return
    # Sadly, bowler only provides FunctionSpec, not ClassSpec.
    capture['function_name'] = capture['class_name']
    capture['function_arguments'] = capture['class_arguments']
    capture['function_parameters'] = node.children[1]
    spec = bowler.FunctionSpec.build(node, capture)
    for i, argument in enumerate(spec.arguments):
        if argument.name != 'custom_command':
            continue
        if not argument.value.children:
            new_node = Node(SYMBOL.atom, [
                Leaf(token.LPAR, '('),
                Node(SYMBOL.testlist_gexp, [
                    Leaf(argument.value.type, argument.value.value),
                    Leaf(token.COMMA, ','),
                    Leaf(token.NUMBER, '0', prefix=' ')]),
                Leaf(token.RPAR, ')')
            ])
            spec.arguments[i] = bowler.FunctionArgument(name=argument.name, value = new_node,
                                                        annotation=argument.annotation,
                                                        star=argument.star, prefix=argument.prefix)
            break
    spec.explode()
if __name__ == '__main__':
    (bowler.Query('test.py')
     .select_class('CheckUpdates')
     .modify(my_modifier)
     .execute(write=True, interactive=True))
</pre>
</details>

Although I am not a `bowler` expert and haven't used it really, I think it is waaaaaaaaaaaaay easier to maintain a short `if not isinstance(self.custom_command, (list, tuple)): ...` than this bulky migration which doesn't even always work:
```sh
$ python fix.py
--- test.py
+++ test.py
@@ -1,6 +1,6 @@
 from test_class import CheckUpdates
-CheckUpdates(custom_command='pamac checkupdates')
+CheckUpdates(custom_command=('pamac checkupdates', 0))
 CheckUpdates(custom_command=' '.join(['pamac', 'checkupdates']))
 CheckUpdates(custom_command=('pamac checkupdates' if HOSTNAME == 'xy' else None))
 CheckUpdates(custom_command=('pamac checkupdates', 42))
Apply this hunk [y,N,q,a,d,?]?
```
The second version is not caught. This is definitely a pathological case, but it would also fail if it was
`CheckUpdates(custom_command=('x' if HOSTNAME == 'MACHINE_A' else 'y'))`. This is definitely realistic as only one machine would need to run on an os that is not in the `self.cmd_dict`.

tych0: > I think it is waaaaaaaaaaaaay easier to maintain a short if not isinstance(self.custom_command, (list, tuple)): ... than this bulky migration which doesn't even always work:

Seems like perhaps we should invent a different API for this, then. Continually asking our users to change their configs is not sustainable, and I'd like to move away from it if possible.

More generally, I don't really understand the motivation for the feature. In particular, neither this patch or the original issue have examples of problematic output, which might be useful to if we want to design alternative solutions.

ep12: I think the check that is currently meant to be the backwards compatibility layer could just be a normal part such that the argument type becomes `custom_command: Union[str, Tuple[str, int]]` where the default 0 is added if using the `str` variant. I don't think we need to deprecate the `str` variant as this makes everything (the migration, config breakage, ...) harder.
Maybe @yobleck can add a rationale to the pr description answering your second point.

yobleck: > More generally, I don't really understand the motivation for the feature

Personally I just wanted to use pamac instead of pacman and it outputs some text even if there are no updates so subtr = 0 wasn't helpful. I later discovered a -q option that removes extra lines but this still seemed like a better idea than trying to to include every possible package manager in the cmd_dict.

> Continually asking our users to change their configs is not sustainable

Agreed. I think the string to list/tuple converter should be good enough in the short term that people won't have to update their configs if they don't want to or are unaware of the change.

libqtile/widget/check_updates.py
yobleck: Should I take out the deprecation message?

libqtile/widget/check_updates.py
ep12: Ah I also didn't see `-q`. So the original main aspect of this pr --- changing the type of `custom_command` is rather a question to make it consistent with the `cmd_dict`. The refactoring to `__init__` also seems a good idea instead of having checks everywhere. But this really does not belong here since the original comment was about a migration for a breaking change that just makes things hard.

libqtile/widget/check_updates.py
ep12: Yeah, I'd suggest to drop that and the comment `# Backwards compatibility for custom_command. Convert string to tuple.` a few lines above.

libqtile/widget/check_updates.py
ep12: @yobleck Could you mark this as resolved, please?

libqtile/widget/check_updates.py
ep12: @tych0 what do you think about the retcode?

libqtile/widget/check_updates.py
```diff
@@ -56,29 +56,37 @@ def __init__(self, **config):
...
+            except KeyError:
+                distros = sorted(self.cmd_dict.keys())
+                logger.error(self.distro + ' is not a valid distro name. ' +
+                             'Use one of the list: ' + str(distros) + '.')
+                self.cmd = None
+                self.subtr = 0
+        else:
+            # Otherwise use custom_command
+            self.cmd = self.custom_command[0]
+            self.subtr = self.custom_command[1]
```
ep12, line 79: Technically, you could simplify this to
```py
self.cmd, self.subtr = self.custom_command
```
if you like.

yobleck: I'll add `[:2]` to custom_command in case the user inputs a list with more than two elements.

libqtile/widget/check_updates.py
yobleck: I just realized that the pamac specific issue can be bypassed with a simple shell script.
```
#!/bin/bash
OUT=$(pamac checkupdates -q)
echo $out
exit 0
```
This should bypass pamac's error code and still get the output. I don't have any updates right now but a test script that just echos a couple of lines and then exits 1 worked but the output was all on one line. With some tweaking it could work better.

With that in mind I don't think its worth adding exit code handling for what seems to be one outlier (thoughts?). Unless there are multiple other package managers that also output non 0 codes. I think this pr should focus on just giving custom_command feature parity with cmd_dict. At the very least even if we don't go through with it, it's still worth moving the custom_command check to \_init\_

yobleck:
I'm sorry that I don't really know what to do here (again new to git stuff and collaborating with other people). 
Have we made a decision on how to handle backwards compatibility?
As for the CalledProcessError, is it worth trying to grab the output or should we just revert back to the old update = ""?
Then once we've made those decisions do I just mark the PR as ready for review?

ep12:
> I'm sorry that I don't really know what to do here (again new to git stuff and collaborating with other people).

Getting started is the hardest part, we all know that! No need for a _sorry_!

> Have we made a decision on how to handle backwards compatibility?

You have a check in the `__init__` so we don't have to worry about migrations or backwards compatibility anymore. It is guaranteed.

> As for the CalledProcessError, is it worth trying to grab the output or should we just revert back to the old update = ""?

If you are fine with a custom wrapper script around `pamac` and don't want to work on the exit code problem, then you probably should do revert it to `update = ""` just to behave as before. That's also what I'd go for, this extra work for exit codes really isn't worth it if three lines of bash can fix it. And actually, because we use `shell=True` on that, you could easily do 
```py
custom_command=("pamac; echo -n", 42)
```
in your configuration. The `echo -n` does nothing (doesn't even print a newline) but returns 0, so that there is no `CalledProcessError`. That idea just came to my mind, I wish it was earlier.

> Then once we've made those decisions do I just mark the PR as ready for review?

Yes, definitely mark that as ready. Others directly start PRs, I also like to start with a draft if I know that there will still be something to do.

One thing you should do if you can is to squash the commits into one. This can also be done when merging, but the authorship might get lost. The easy route is to wait for that to happen. The longer route where you learn something about git is to rebase your work interactively. (And this would be easier if you were working on a feature branch.) So I guess that for now you don't want to rebase and squash :).

ep12:
[Seems like `git` can indeed preserve ownership.](https://github.com/ep12/qtile/commit/284a2eb578ae9432e5e6d3b5de81916d93421001) So there shouldn't be a problem...

libqtile/widget/check_updates.py
m-col: > With that in mind I don't think its worth adding exit code handling for what seems to be one outlier 

I agree, and I think that suggesting small custom scripts like yours above is a valid solution to users whose distro's update checker needs a little wrapping.

m-col:
Rather than having a bit of an awkward `(command, int)` format for a parameter, could we not just have an option that accepts a function that can be used to adjust the output? e.g.:
```py
    adjust=lambda value: value - 1
```
where value is the number of lines and `value - 1` is the number of updates that are actually there?

This removes any need to change the parameters as they come in the widget, and opens up any additional flexibility if required by anyone else who comes along in the future saying their check-updates script outputs 2 lines per package (i.e. you could pass `lambda v: v / 2`.

ep12:
> Rather than having a bit of an awkward `(command, int)` format for a parameter, could we not just have an option that accepts a function that can be used to adjust the output? e.g.:
> 
> ```python
>     adjust=lambda value: value - 1
> ```
> 
> where value is the number of lines and `value - 1` is the number of updates that are actually there?
> 
> This removes any need to change the parameters as they come in the widget, and opens up any additional flexibility if required by anyone else who comes along in the future saying their check-updates script outputs 2 lines per package (i.e. you could pass `lambda v: v / 2`.

Huh? Uh. Passing a function broadens the possibilities a bit, but what if the command outputs a header that might have varying length, e.g. missing and outdated packages from the AUR? This is getting complicated. If we'd want to be able to deal with that, it'd be easier to make the user implement what they need.
I don't think it would help to make the second argument a function. The only thing that would be meaningful would be to move the number to subtract to yet another argument, e.g. `custom_command_subtract: int = 0`. How should the argument be handled if it is set but `custom_command` isn't? Should that override the distro-specific value? Should it be ignored? The way it currently is is explicit about that, although I have to admit that passing a tuple never looks sexy. But most users would pass a string, so I am definitely fine with how it is. The user can always write a custom wrapper script to preprocess the output to ignore headers, footers, additional lines with package descriptions, packages that are held and will not be updated, ...

m-col:
>  Passing a function broadens the possibilities a bit, but what if the command outputs a header that might have varying length, e.g. missing and outdated packages from the AUR? This is getting complicated. If we'd want to be able to deal with that, it'd be easier to make the user implement what they need.

Then that would be very simple! `lambda v: (v - 10) // 2` if there was a header of 10 lines, and 2 lines per package. This _does_ make the user implement what they need, no?

> I don't think it would help to make the second argument a function. The only thing that would be meaningful would be to move the number to subtract to yet another argument, e.g. custom_command_subtract: int = 0

Sorry, that's actually what I meant. Remove the whole `(command, int)` thing so it always just takes a string or None, and then provide the needed update as a stand-alone option. My suggestion is then similar to adding your `custom_command_subtract` except it's slightly more flexible in that it can do a little bit more arithmetic but taking a function.

> How should the argument be handled if it is set but custom_command isn't?

Like you say it doesn't make so much sense on the built-in checks so maybe the name of the parameter should be clear about that, much like your `custom_command_subtract`.

> The user can always write a custom wrapper script to preprocess the output to ignore headers, footers, additional lines with package descriptions, packages that are held and will not be updated

Sure, they could do all that in a relatively small shell script and plug it into `widget.GenPollText`, but this function just makes things a bit more convenient :)

> Oh, and you probably meant lambda v: v // 2 or is the number of updates rational? 2.5 updates? rofl

You never know what strange systems people are running these days!

yobleck:
> How should the argument be handled if it is set but custom_command isn't?

I was thinking replace `self.subtr` with `self.modify_output` where modify points to the lambda function

If custom_command is a string 
(random tangent. For our purposes whats the difference between the default 
`custom_command = None` and `custom_command = ""` ?)
then modify_output will just be what he user inputs or the default `lambda x: x`
and for the built in options do `self.modify_output = lambda x: x - self.cmd_dict[self.distro][1]`
which would override what the user puts in (maybe with a warning?) and avoids changing too much other code.

Then in _check_updates() after the try block, replace num_updates with:
`num_updates = self.modify_output(len(updates.splitlines()))`

The name modify_output is up for debate. Is something like custom_command_modify_output
or custom_command_modify_line_count too long?

ep12:
Looks good!

m-col:
Yeah the changes look pretty good. Please could you squash them all into a single commit?

yobleck:
> Yeah the changes look pretty good. Please could you squash them all into a single commit?

I've never done that before and I think I just messed it up.
I ran `git rebase -i HEAD~12`
changed the last commit from pick to squash
then ran `git push origin`.
should I have used a different name that origin?
I wasn't able to find anything like this (https://i.stack.imgur.com/8t1OJ.png) on github's website, presumably because I don't have write permissions.
I'm not sure what to do next.

m-col:
You've mostly got it. `git rebase -i` is right, you just need to use `squash` for all of the commits except the first one ("use commit, but meld into previous commit"). Then you're left with a single commit, whose git message you may need to amend - I can't remember if you can do this when rebasing, if not you may need to do `git commit --amend`. This basically lets you edit the commit message just so it is summarising the PR's addition to the code (it's useful when looking through the git history). Unless you've set up additional remotes then `origin` should work fine (`git pull` alone will use that). However github will complain because the history you're pushing contradicts the history that it has. To overwrite github's old version of the branch you add a 'force' flag: `git push -f`. That should do it :)

m-col:
Brilliant, thanks for this!

PR #2328, opened 2021-03-27
[RFC] Fix pylint warnings and enable it as a style guard in the CI config

ep12:
This continues my previous work to get pylint working (#2316).
More specifically, the goal is to deal with all the warnings that pylint throws at us. Once it does not complain anymore, we can finally change `tox.ini` such that tests fail if a change introduces bad code again.
This pr fixes:
+ [x] `redefined-builtin`
+ [x] `unnecessary-pass`
+ [ ] `eval-used` and `exec-used`
+ [ ] `tox.ini`: When this change is made, tests can fail if pylint fails. Until now, the return code of the pylint run is ignored.

Some stats for @ramnes about `redefined-builtin`
| file(\.function)?::line                                | builtin           | lines changed (approx.) | strategy        | reason                                                       |
| ------------------------------------------------------ | ----------------- | ------------- | --------------- | ------------------------------------------------------------ |
| `libqtile.config.Rule.__init__::686`                   | `float`           | -2+3          | disable locally | is just an argument, maintain compatibility.                 |
| `libqtile.scripts.start::25`                           | `exit`            | -2+2          | rename          |                                                              |
| `libqtile.widget.mpd2widget::11`                       | `ConnectionError` | -1+3          | disable locally | `ConnectionError` is rarely used.                            |
| `libqtile.core.manager.cmd_display_kb::1026`           | `list`            | -5+5          | rename          |                                                              |
| `libqtile.core.manager.cmd_display_kb::1035`           | `format`          | -2+2          | rename          |                                                              |
| `libqtile.layout.stack::157`                           | `next`            | -3+3          | rename          |                                                              |
| `libqtile.layout.stack::172`                           | `next`            | -3+3          | rename          |                                                              |
| `libqtile.layout.stack::330`                           | `next`            | -3+3          | rename          |                                                              |
| `libqtile.utils::104`, `libqtile.backend.x11.xcbq::55` | `hex`             | -7+7          | rename          |                                                              |
| `libqtile.utils::187`                                  | `id`              | -4+4          | rename          |                                                              |
| `libqtile.utils::215`                                  | `id`              | -2+2          | rename          |                                                              |
| `libqtile.extension.window_list::44`                   | `id`              | -3+3          | rename          |                                                              |
| `libqtile.backend.x11.xcbq.set_property::638`          | `type`, `format`  | -1+4          | disable locally | are just arguments, maintain compatibility. Also, 20 test lines use this. |
| `libqtile.backend.x11.xcbq.get_property::688`          | `type`            | -1+2          | disable locally | naming consistency with `set_property`                       |

Why _approx._? 'Cause I didn't want to push 14 commits for `redefined-builtin` and might have missed a line here or there.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -52,7 +52,7 @@ from xcffib.xproto import CW, EventMask, WindowClass
 from libqtile import xkeysyms
 from libqtile.backend.x11.xcursors import Cursors
 from libqtile.log_utils import logger
-from libqtile.utils import hex
+from libqtile.utils import color_hex
 
 keysyms = xkeysyms.keysyms
 
@@ -342,7 +342,7 @@ class Colormap:
             def x8to16(i):
                 return 0xffff * (i & 0xff) // 0xff
 
-            color = hex(color)
+            color = color_hex(color)
             r = x8to16(int(color[-6] + color[-5], 16))
             g = x8to16(int(color[-4] + color[-3], 16))
             b = x8to16(int(color[-2] + color[-1], 16))
@@ -635,7 +635,10 @@ class Window:
             self.wid, mask, values
         )
 
-    def set_property(self, name, value, type=None, format=None):
+    def set_property(self, name, value,
+                     type=None, format=None):  # pylint: disable=redefined-builtin
+        # builtins type and format are redefined. As they are only arguments to a method
+        # which is used regularly, we maintain compatibility here.
         """
         Parameters
         ==========
@@ -685,7 +688,8 @@ class Window:
                 'X error in SetProperty (wid=%r, prop=%r), ignoring',
                 self.wid, name)
 
-    def get_property(self, prop, type=None, unpack=None):
+    def get_property(self, prop, type=None, unpack=None):  # pylint: disable=redefined-builtin
+        # builtin type is redefined. We allow this here to be consistent with self.set_property.
         """Return the contents of a property as a GetPropertyReply
 
         If unpack is specified, a tuple of values is returned.  The type to
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1023,17 +1023,17 @@ class Qtile(CommandObject):
                 format_string = " ".join("%-{0:d}s".format(max_col_size + 2) for max_col_size in self.max_col_size)
                 return format_string + "\n", len(self.max_col_size)
 
-            def expandlist(self, list, n):
-                if not list:
+            def expandlist(self, row, n):
+                if not row:
                     return ["-" * max_col_size for max_col_size in self.max_col_size]
-                n -= len(list)
+                n -= len(row)
                 if n > 0:
-                    list += [""] * n
-                return list
+                    row += [""] * n
+                return row
 
             def __str__(self):
-                format, n = self.getformat()
-                return "".join([format % tuple(self.expandlist(row, n)) for row in self.rows])
+                fmt, n = self.getformat()
+                return "".join([fmt % tuple(self.expandlist(row, n)) for row in self.rows])
 
         result = FormatTable()
         result.add(["KeySym", "Mod", "Command", "Desc"])
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1474,7 +1474,7 @@ class Qtile(CommandObject):
                     logger.info('No command entered.')
                     return
                 try:
-                    result = eval(u'c.{0:s}'.format(cmd))
+                    result = eval(u'c.{0:s}'.format(cmd))  # pylint: disable=eval-used
                 except (CommandError, CommandException, AttributeError) as err:
                     logger.error(err)
                     result = None
--- a/libqtile/command/base.py
+++ b/libqtile/command/base.py
@@ -189,9 +189,10 @@ class CommandObject(metaclass=abc.ABCMeta):
         """
         try:
             try:
-                return True, str(eval(code))
+                # The method announces the powers of eval, so we definitely need it:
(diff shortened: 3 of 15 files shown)


libqtile/backend/x11/xcbq.py
```diff
@@ -52,7 +52,7 @@
 from libqtile import xkeysyms
 from libqtile.backend.x11.xcursors import Cursors
 from libqtile.log_utils import logger
-from libqtile.utils import hex
+from libqtile.utils import color_hex
```
ep12, line 55: Is that name okay for you?

ep12:
There are still some `eval`s and `exec`s left:

+ [ ] This one had `eval` until ac3906750cd changed that to `exec`:
    https://github.com/qtile/qtile/blob/8358d6d8b7641fdb81c867dbc87d12a36c07c09c/libqtile/widget/launchbar.py#L191
    Is this use of `exec` justified and the message should be locally disabled?
+ [ ] The second one is
    https://github.com/qtile/qtile/blob/8358d6d8b7641fdb81c867dbc87d12a36c07c09c/libqtile/widget/prompt.py#L154-L164
    Can't this be done using `hasattr`? Do we really need `eval` for this one? Or at least the first occurrence in L154?

libqtile/backend/x11/xcbq.py
ramnes: I guess it could be more descriptive like "as_hex" or something but I don't think it's very important at this point. :)

ramnes:
Is there really a point in avoiding eval and exec in a WM? 

> Is this use of exec justified and the message should be locally disabled?

I don't like this feature but now that it's in the code, I don't see how we could implement it differently without breaking the API.

> Can't this be done using hasattr? Do we really need eval for this one? Or at least the first occurrence in L154?

It probably can I guess?

ep12:
> Is there really a point in avoiding eval and exec in a WM?

I already marked some `eval`s and `exec`s as fine, because the user expects an `eval` or `exec` there (`cmd_qtilecmd` and `cmd_eval`, using `eval` is definitely not surprising there). Those two functions that are left over are not obviously related to `eval` and `exec`. I don't like that. We could either clarify that we are using that here, or (if possible), get rid of it and do it differently.

> I don't like this feature but now that it's in the code, I don't see how we could implement it differently without breaking the API.

Didn't try to rewrite it, I was asking for opinions first. So that might be another `pylint: disable=...` for now?
 
> > Can't this be done using hasattr? Do we really need eval for this one? Or at least the first occurrence in L154?
> 
> It probably can I guess?

The I probably should try to rewrite it.

tych0:
r.e. the launchbar exec, I was going to suggest we just delete it, because it seems like kind of an obtuse API. However, it looks like as recently as 96ac866e54500836f29d2d756391a59d88a4fd97 it was being used by someone, so maybe best just to ignore it for now. I agree that it doesn't really seem like a big problem: we're exec()ing user-defined strings. Would be nice to get rid of it, but I think leave it for now. Basically exactly what @ramnes said:

> I don't like this feature but now that it's in the code, I don't see how we could implement it differently without breaking the API.

libqtile/core/manager.py
```diff
@@ -1474,7 +1474,7 @@ def f(cmd):
                     logger.info('No command entered.')
                     return
                 try:
-                    result = eval(u'c.{0:s}'.format(cmd))
+                    result = eval(u'c.{0:s}'.format(cmd))  # pylint: disable=eval-used
```
tych0, line 1477: This one it feels like we can just delete (i.e. delete `cmd_qtilecmd`). Nobody is (successfully) typing the qtile API directly into their prompt widget on their bar...

flacjacket:
> r.e. the launchbar exec, I was going to suggest we just delete it, because it seems like kind of an obtuse API. However, it looks like as recently as [96ac866](https://github.com/qtile/qtile/commit/96ac866e54500836f29d2d756391a59d88a4fd97) it was being used by someone, so maybe best just to ignore it for now. I agree that it doesn't really seem like a big problem: we're exec()ing user-defined strings. Would be nice to get rid of it, but I think leave it for now. Basically exactly what @ramnes said:
> 
> > I don't like this feature but now that it's in the code, I don't see how we could implement it differently without breaking the API.

Another option, I think we should be able to parse the input from the user and properly navigate to the object that is requested rather than just throwing it at an exec and letting that try to sort it out. This is somewhat related to my comments on the command client vs the interactive command client, and should be able to be done a bit more cleanly (or at least without an exec) using the regular (non-interactive) command client in the backend.

ep12:
> > r.e. the launchbar exec, I was going to suggest we just delete it, because it seems like kind of an obtuse API. However, it looks like as recently as [96ac866](https://github.com/qtile/qtile/commit/96ac866e54500836f29d2d756391a59d88a4fd97) it was being used by someone, so maybe best just to ignore it for now. I agree that it doesn't really seem like a big problem: we're exec()ing user-defined strings. Would be nice to get rid of it, but I think leave it for now. Basically exactly what @ramnes said:
> > > I don't like this feature but now that it's in the code, I don't see how we could implement it differently without breaking the API.
> 
> Another option, I think we should be able to parse the input from the user and properly navigate to the object that is requested rather than just throwing it at an exec and letting that try to sort it out. This is somewhat related to my comments on the command client vs the interactive command client, and should be able to be done a bit more cleanly (or at least without an exec) using the regular (non-interactive) command client in the backend.

@flacjacket You seem to be very familiar with all that. Could you make those changes because I'll be running short of time very soon and I don't think I can acquire the necessary skills in the near future.

ep12:
I think this should still be worked on, right?

ramnes:
The question isn't if it should still be worked on but if someone actually works on it. :)

PR #2330, opened 2021-03-27
CONTRIBUTING: add more docs about what we look for in PRs

tych0:
We often say the same things over and over on reviews, so I tried to write
them all down here.

The biggest thing is requiring a `qtile migrate` for all breaking changes,
which we've been experimenting with and seems like maybe it'll work. Let's
see how it goes :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,10 +40,24 @@ Pull requests are not considered complete until they include all of the
 following:
 
 1. Code: conforms to PEP8 and passes `make lint`.
-2. Unit tests: CI tests pass. Adding new tests to verify that your code works is recommended.
+2. Unit tests: CI tests pass. Please add new unit tests to ensure your code works and stays working.
    See [our website](http://docs.qtile.org/en/latest/manual/contributing.html#running-tests-locally)
    on how to run the tests locally.
 3. Documentation: Should get updated if it needed.
+4. A ``qtile migrate`` config migration is necessary for any config-breaking change.
+5. Please do not include unrelated formatting changes.
+6. Widgets don't need to catch their own exceptions, or introduce their own
+   polling infrastructure. The code in ``libqtile.widget.base.*`` does all of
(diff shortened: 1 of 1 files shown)


ep12:
What about the _contributing_ section in the docs? The _Running tests locally_ section was added to the docs. You are adding details to the markdown file. To avoid confusion, it would probably best to decide for one of those places and the other one being a very short document with not more than a link?

m-col:
Yeah... perhaps we could just add a small note and link to the website into the markdown file, or alternatively put everything into the mardown file but then parse it when generating the docs to create the website. Hacky, in true Qtile style ;)

tych0:
>  What about the contributing section in the docs?

Yeah, that sounds maybe reasonable. I don't know where people look for this kind of information when contributing to a project.

ep12:
> Yeah, that sounds maybe reasonable. I don't know where people look for this kind of information when contributing to a project.

I think this depends on where we point them.  If we had a link _contribution_ on pypi right where the link to the homepage is (and the link to the source code and issue tracker could be) and a detailed section on the website (docs), most people would probably find that. Some might come from the _Hacking on qtile_ page which is right below _Contributing_. (Could me move that above contributing?) Those few people reading the CONTRIBUTING.md on GitHub probably wouldn't mind if they were pointed to the docs. And those who already have contributed wouldn't need to read this section again. So I think it is all about where we point them.

tych0:
Sure, feel free to send a PR to arrange the docs however you want. Should I close this and you'll incorporate it to your PR, or should we merge this?

ep12:
> Sure, feel free to send a PR to arrange the docs however you want. Should I close this and you'll incorporate it to your PR, or should we merge this?

Would make sense to do things together and not start shifting things around. I could open a pr but I definitely don't mind if you want to do it :) Or which pr do you mean? Because my #2316 is already merged.

tych0:
> I could open a pr but I definitely don't mind if you want to do it :)

I don't want to do it. Please include the documentation from this PR in your rework.

Issue #2329, opened 2021-03-27
Random (?) `pyp3` CI hangs


ep12:
# Issue description
Two CI runs (6e05a3e2b and 7428c43bd7b1a) failed: both times it was the tox environment `pypy3` and both times the last tests that were run are `test/layouts/test_common.py::test_desktop_notifications[...]` tests. Sadly, the CI does not show how long each test took, so I am not sure if those tests hang or if previous ones are the culprit.
I am running `tox -e <env> -- --durations 0` to get some timings and will add some stats soon.

# Qtile version
[fix_pylint_warnings](https://github.com/ep12/qtile/tree/fix_pylint_warnings) or equivalently `upstream/master`, because I haven't changed the tests.

# Cancelled test runs
+ https://github.com/qtile/qtile/runs/2206210095?check_suite_focus=true
+ https://github.com/qtile/qtile/runs/2177893963?check_suite_focus=true
+ https://github.com/qtile/qtile/runs/2206210095?check_suite_focus=true
+ https://github.com/qtile/qtile/runs/2159766343?check_suite_focus=true
+ https://github.com/qtile/qtile/runs/2004857936?check_suite_focus=true
+ https://github.com/qtile/qtile/runs/1986400750?check_suite_focus=true (before `test_desktop_notifications`!)
+ https://github.com/qtile/qtile/runs/1912859219?check_suite_focus=true (!)
+ https://github.com/qtile/qtile/runs/1880398025?check_suite_focus=true (!)

ep12:
Timing stats for `py39` and `pypy3` on my machine:
<details><summary><code>py39</code></summary>
<pre>
                                                                          mean rel. std   max         call        setup     teardown len
name                                                                                                                                    
test_qtile_cmd.py::test_qtile_cmd                                        4.410    0.00% 4.410      4.410±0      0.110±0      0.030±0   1
test_fakescreen.py::test_hammer_ratio_tile                               1.280    0.00% 1.280      1.280±0      0.250±0      1.090±0   1
test_fakescreen.py::test_hammer_tile                                     1.260    0.00% 1.260      1.260±0      0.270±0      1.090±0   1
layouts/test_treetab.py::test_window                                     0.850    0.00% 0.850      0.850±0      0.110±0      0.030±0   1
layouts/test_common.py::test_desktop_notifications                       0.749    2.64% 0.800  0.749±0.020  0.112±0.004  0.025±0.005  13
[...]

PR #2331, opened 2021-03-27
Add `pytest-timeout` plugin to fail when tests hang

ep12:
Related: #2329 
This pr adds `pytest-timeout` (with a timeout of 30min = 1800s for a test to complete) to the test suite.
Although timings can be obtained with `tox -e <ENVNAME> -- --durations=0`, this only prints them after all tests succeed or fail in a nominal way. If the process is killed (Ctrl-C, GitHub Actions execution time limit exceeded), the acquired duration data will not be printed. Hence it makes sense to print a rough estimation (`%.3f`, in seconds) right after each test ends, resulting in either `PASSED (0.123)` or `FAILED (6.283)` to be printed. In case of a timeout, `TIMEOUT` will be printed instead. This should help identify how long each test takes to execute.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,6 +91,9 @@ ban-relative-imports = true
 python_files = test_*.py
 testpaths = test
 addopts = --verbose
+# 10 minutes should be enough for a single test to complete:
+timeout = 600
+timeout_func_only = True
 
 [check-manifest]
 ignore =
--- a/setup.cfg
+++ b/setup.cfg
@@ -93,7 +93,8 @@ testpaths = test
 addopts = --verbose
 # 10 minutes should be enough for a single test to complete:
 timeout = 600
-timeout_func_only = True
+timeout_func_only = False
+timeout_method = thread
 
 [check-manifest]
 ignore =
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -61,6 +61,41 @@ def pytest_addoption(parser):
     )
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """Make timeouts appear as such in the logs.
+
+    The pytest-timeout plug-in messes with the durations that are reported for tests.
+    Fix that by marking the report such that we can print `TIMEOUT` instead of
+    `FAILED (incorrect duration)`. See
+    https://github.com/pytest-dev/pytest/blob/38d8deb74d95077ebf189440ca047e14f8197da1/src/_pytest/runner.py#L366
+    """
+    r = yield
+    if not hasattr(r, 'get_result'):
+        return
+    report = r.get_result()
+    timed_out = False
+    if hasattr(call.excinfo, 'value'):
+        msg = getattr(call.excinfo.value, 'msg', None)
+        if isinstance(msg, str) and msg.startswith('Timeout >'):
+            timed_out = True
+    report.timed_out = timed_out
+
+
+def pytest_report_teststatus(report, **kwargs):
+    """Report test run duration, adapted from
+    https://github.com/pytest-dev/pytest/blob/38d8deb74d95077ebf189440ca047e14f8197da1/src/_pytest/runner.py#L202
+    """
+    d = report.duration
+    if report.passed:  # category, shortletter, verbose-word
+        return 'passed', 'P', 'PASSED (%.3f)' % d
+    if report.timed_out:
+        return 'failed', 'T', 'TIMEOUT'
+    if report.failed:
+        return 'failed', 'F', 'FAILED (%.3f)' % d
+    return None
+
+
 class Retry:
     def __init__(self, fail_msg='retry failed!', ignore_exceptions=(),
                  dt=sleep_time, tmax=max_sleep, return_on_fail=False):
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -35,6 +35,7 @@ import pytest
 import xcffib
 import xcffib.testing
 import xcffib.xproto
+from _pytest.runner import runtestprotocol
 
 import libqtile.config
 from libqtile import command, ipc
@@ -86,14 +87,38 @@ def pytest_report_teststatus(report, **kwargs):
     """Report test run duration, adapted from
     https://github.com/pytest-dev/pytest/blob/38d8deb74d95077ebf189440ca047e14f8197da1/src/_pytest/runner.py#L202
     """
-    d = report.duration
+    if report.skipped:
+        return 'skipped', 'S', 'SKIPPED'
+    timings = getattr(report, 'timings', None)
+    d = ('%s=%s' % ('+'.join('%.3f' % t if t > 1e-3 else '0' for t in timings),
+                    '%.3f' % sum(timings) if sum(timings) > 1e-3 else '0')
+         if timings else '%.3f' % getattr(report, 'duration', 0.0))
     if report.passed:  # category, shortletter, verbose-word
-        return 'passed', 'P', 'PASSED (%.3f)' % d
-    if report.timed_out:
-        return 'failed', 'T', 'TIMEOUT'
+        return 'passed', 'P', 'PASSED (%s)' % d
+    if getattr(report, 'timed_out', False):
+        return 'failed', 'T', 'TIMEOUT (%s)' % d
     if report.failed:
-        return 'failed', 'F', 'FAILED (%.3f)' % d
+        return 'failed', 'F', 'FAILED (%s)' % d
     return None
+
+
+def pytest_runtest_protocol(item, nextitem) -> bool:
+    """Report test run duration, adapted from
+    https://github.com/pytest-dev/pytest/blob/9653a0e9f47ad2ae5135a974db52ddeb5bfcf5d9/src/_pytest/runner.py#L110
+    """
+    ihook = item.ihook
+    ihook.pytest_runtest_logstart(nodeid=item.nodeid, location=item.location)
+    reports = runtestprotocol(item, log=False, nextitem=nextitem)
+    if reports:
+        timings = []
+        for report in reports:
+            timings.append(getattr(report, 'duration', 0.0))
+            if any([report.failed, report.skipped, getattr(report, 'timed_out', False)]):
+                break
+        report.timings = timings
+        ihook.pytest_runtest_logreport(report=report)
+    ihook.pytest_runtest_logfinish(nodeid=item.nodeid, location=item.location)
+    return True
 
 
 class Retry:
--- a/tox.ini
+++ b/tox.ini
@@ -22,9 +22,14 @@ setenv = LC_CTYPE = en_US.UTF-8
 passenv = DISPLAY TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 whitelist_externals=convert
 # xcffib has to be installed before cairocffi
+#
+# NOTE: pytest-timeout might trigger a `Timeout` while the coverage trace
+# function is running. This causes a warning about probably incorrect coverage
+# stats. The timeout makes the CI tests fail, so we don't care about coverage.
 deps =
     pytest >= 6.2.1
     pytest-cov >= 2.10.1
+    pytest-timeout >= 1.4.1
     setuptools >= 40.5.0
     xcffib >= 0.8.1
     bowler
@@ -81,6 +86,7 @@ deps =
     bowler
     xcffib >= 0.8.1
     pytest >= 6.2.1
+    pytest-timeout >= 1.4.1
 commands =
     pip install -r requirements.txt
     mypy -p libqtile


test/conftest.py
```diff
@@ -31,6 +31,7 @@
 import time
 import traceback
 
+import _pytest
```
m-col, line 34: I'm guessing this is a private module for a reason... what's the situation with `_pytest`? Should we avoid it if we can?

test/conftest.py
```diff
@@ -61,6 +62,45 @@ def pytest_addoption(parser):
...
 
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtest_makereport(item, call):
+    """Make timeouts appear as such in the logs.
+
+    The pytest-timeout plug-in messes with the durations that are reported for tests.
+    Fix that by marking the report such that we can print `TIMEOUT` instead of
+    `FAILED (incorrect duration)`. See
+    https://github.com/pytest-dev/pytest/blob/38d8deb74d95077ebf189440ca047e14f8197da1/src/_pytest/runner.py#L366
+    """
```
m-col, line 73: If this is the case, how do others who use pytest-timeout handle this?

setup.cfg
```diff
@@ -91,6 +91,9 @@ ban-relative-imports = true
 python_files = test_*.py
 testpaths = test
 addopts = --verbose
+# 30 minutes should be enough for a single test to complete:
+timeout = 1800
+# timeout_method = thread does not work!
```
m-col, line 96: looks like pytest-timeout will choose a good timeout_method for us, so this line could be dropped

test/conftest.py
```diff
@@ -61,6 +62,45 @@ def pytest_addoption(parser):
...
+    timed_out = False
+    try:
+        if call.excinfo is not None:
+            msg = call.excinfo.value.msg
+            if isinstance(msg, str) and msg.startswith('Timeout >'):
+                timed_out = True
+    except Exception:  # pylint: disable=broad-except
```
m-col, line 80: I'm not sure i understand what this block is doing. Please could you explain a little bit? Also, could exceptions do you see that get risen here? Perhaps they should be coded in to avoid the broad-except.

m-col (commented):
I'm a bit on the fence about adding pytest-timeout to the suite because it seems like a workaround for problems that would need actual fixing. However, as you mention in #2329 it may not be Qtile that is causing these problems, so perhaps it's a nice thing to have the tests cancelled early if they are going to fail. Either way, I'd lean more towards adding it if it is optional, so systems that do not have it installed would work as they originally do. Is that possible (or perhaps already how it works)?

I have a couple other comments inline, the only other thing I'd suggest (if we go ahead with it) is lowering the timeout. atm it is 30 minutes per test, but you could probably set it to 30 seconds per test. Your testing in the issue show that the longest is 4s, so we may as well reduce it just because it might keep us focussed if we are pushing a faulty commit and it doesn't take 30 minutes to timeout :)

test/conftest.py
ep12: We cannot. The report class any almost everything is in `_pytest`, the "public" `pytest` module is nearly empty.

test/conftest.py
ep12: Haven't had a deep look round, but most people just use it to fail tests after a certain period. Most users don't care about timings AND time-outs. Printing TIMEOUT was the only solution that didn't involve printing incorrect durations. Even with `time.time` in the exception handler I got weird results that were below the threshold.

setup.cfg
ep12: I think it always uses `signal` by default. I left that as a warning that everything breaks if we use the `thread` method and a timeout is actually triggered.

test/conftest.py
ep12: Have a look at the default implementation (url): it is a one-liner that creates and returns the report. We introduce additional code here, so we definitely do not want weird exceptions cause more problems. Exceptions should not occur, and when the test has timed out (or the test failed, which is actually a caught `AssertionError` most of the time) everything proceeds as pytest does it normally. The only purpose of this block is to add a magic attribute to the report so that we can later tell that it was a timeout. The exception info that is part of the `call` is dropped (see the constructor/class method) when creating the report, I think. So our only chance is to catch that here and store our findings. Then we can indeed use that later on in the `pytest_report_teststatus` function.

ep12: I know that this is a broad except and right now we are not even checking our tests for stuff like that, but we really want this here. Hence I already added the `pylint: disable=` comment :)

ep12:
> I'm a bit on the fence about adding pytest-timeout to the suite because it seems like a workaround for problems that would need actual fixing. However, as you mention in #2329 it may not be Qtile that is causing these problems, so perhaps it's a nice thing to have the tests cancelled early if they are going to fail.

Yep, I didn't mark this pr as a fix for that issue for exactly that reason. But until now there is no indication why those hangs occur. Maybe some polling inside qtile or async libraries randomly takes longer in `pypy`. Whatever the cause is, we probably want to find that. I also stumbled across [this](https://docs.pytest.org/en/stable/usage.html#modifying-python-traceback-printing), more specifically the `--full-trace` argument for pytest. This might come in handy that we get some insight when a timeout occurs. If GitHub kills the tests, we get nothing at all. With `pytest-timeout` we at least trigger pytest to stop and print a (full) traceback before the CI run fails. The most annoying thing about this is that all the pypy tests worked on my machine. Maybe it really is random, but until now it hasn't happened to me locally.

> Either way, I'd lean more towards adding it if it is optional, so systems that do not have it installed would work as they originally do. Is that possible (or perhaps already how it works)?

If you use `tox` then it'll automatically be installed. Use `tox -e <ENV> -- --timeout=299792458` (meters per second :P) or so to disable the timeout. If you use pytest directly, it'll probably just ignore that unknown timeout option in setup.cfg. All the other stuff in conftest.py does not require `pytest-timeout` so everything will work just fine I guess.

> I have a couple other comments inline, the only other thing I'd suggest (if we go ahead with it) is lowering the timeout. atm it is 30 minutes per test, but you could probably set it to 30 seconds per test. Your testing in the issue show that the longest is 4s, so we may as well reduce it just because it might keep us focussed if we are pushing a faulty commit and it doesn't take 30 minutes to timeout :)

Yep, I also find that quite high. _I think_ I saw a timeout of 10000 somewhere in the codebase. Not sure if that is in seconds or milliseconds. So for a start I chose something very conservative. Actually, some tests passed just under 20 seconds, so it is probably a good idea to keep the timeout quite high, at least 5 minutes. If the machine has a hickup and continues nominally afterwards, even a test that _randomly_ takes some minutes is acceptable if the tests pass all in all.
Also, most faulty commits fail after a few minutes, so timeouts shouldn't be the reason for failure that often :P

test/conftest.py
ep12: Actually, I probably should have a second look round. Maybe pytest exposes the report class somewhere else, but it is definitely defined there.

ep12: The `pytest.__init__.py` does not expose the `TestReport` class. So sadly the import of `_pytest` is necessary.

tych0:
> it seems like a workaround for problems that would need actual fixing.

I agree, I think it's best not to merge this. If the pypy failures are getting annoying and nobody has time to fix them, let's disable pypy and stop claiming that we support it.

So far (to me at least), they don't seem *that* frequent, so there's some hope that someone will come along and fix them. But I agree, we don't want to be in this state forever.

ep12:
> > it seems like a workaround for problems that would need actual fixing.
> 
> I agree, I think it's best not to merge this. If the pypy failures are getting annoying and nobody has time to fix them, let's disable pypy and stop claiming that we support it.
> 
> So far (to me at least), they don't seem _that_ frequent, so there's some hope that someone will come along and fix them. But I agree, we don't want to be in this state forever.

Yep. It wasn't my intent to say that we should ignore pypy. What I would like to express is
+ I'd like to get some timing data for tests. Directly, not in the report because tox might be killed before it gets to print it.
+ If tests time out, tox should handle that and print a summary including a (full) stack trace where exactly the thread was when it caught the signal to time out. That's more graceful than GitHub killing the process without getting a report / summary.
+ Those stack traces _might_ help to identify the causes of those hangs. If the stack traces all have the same line somewhere in between, then this is a place to start investigating.
+ The pypy tests have not hung on my machine, so until now I have no idea where to start searching for the root. A random guess would be some kind of `poll` or `wait_for_event` or something. But until I can't get pypy to fail, ..., you know, no place to start.
+ I don't think of this as a workaround but more as an addition to the test suite that helps us investigate anomalies _after_ they've happened. E.g. I can't reproduce the pypy hangs and they do occur randomly. I listed some in #2329 and I am not the only one experiencing them, but they are definitely random, hence an anomaly.

test/conftest.py
ep12: I `rg`'d the source code of pytest for `TestReport` and it only appeared in `src/_pytest` and the docs and test directories, but not in `src/pytest`. So it is a must.

m-col:
I the hangs are not reproducible on your own machine, I can see why having the timing data could be invaluable in tracking down the hangs. It is a bit of a hammer approach but in the absence of alternative approaches I guess it might be a good idea, for as long as it's seen as a temporary thing.

Then again... if you've got it set up on your fork then using it there might be sufficient to be helpful.

ramnes:
What happens if we just add pytest-timeout without the code you've added in conftest?

ep12:
Then the execution will still time out as expected, but we don't get the timings in-line with the PASSED|FAILED|TIMEOUT status. If you don't want the timings, then I can throw that out of the window.

ramnes:
Do we care about the timing of tests that do not timeout? And it it does timeout, won't the timing be the same as the timeout?

ep12:
General aside: This pr is not about pypy, it is about timeout limits (and timing data) for tests. So if you reject it, please reject it because you don't think it is a helpful addition to the test/test output and ***not*** because it doesn't fix pypy.

-----------------

> Then again... if you've got it set up on your fork then using it there might be sufficient to be helpful.

My fork is just for the development of pull requests, I don't plan to let this diverge. Apart from that, random hangs already imply that the machinery to catch those must be running with all tests to ever get something out of it. I can't see how it could be sufficient on my fork alone where I only do tests locally. And they always passed, even on pypy.

ep12:
> Do we care about the timing of tests that do not timeout?

That is your decision :) I would. Then it is easier to see if some tests take very long.

> And it it does timeout, won't the timing be the same as the timeout?

No, unfortunately. That's why I had to introduce the TIMEOUT status. Without that it said `FAILED (3.827)` and then later on `Timeout >4s`. Some people might read that as _the tests are doing weird things_, so it probably isn't my fault if the tests fail. But of course it should be interpreted as _I introduced code that timed out, so I have to fix this_.

ramnes:
Okay, thanks for the explanation, I get it now. I'd say let's try to have this merged upstream first. :)

I personally like the timeouts and the TIMEOUT status (although I'd prefer this not to be in our codebase), but I don't mind if we don't have the duration of each test.

@mcol @tych0 I don't think it's "a workaround for problems that would need actual fixing" if we keep the tests failing, it will just fail quicker and that's great.

ep12:
> I personally like the timeouts and the TIMEOUT status (although I'd prefer this not to be in our codebase), but I don't mind if we don't have the duration of each test.

Actually, I'd prefer `pytest-timeout` to be built into `pytest` with correct calculation of test run durations :P but that's not the way it currently is and right now I don't have the time to open a discussion about that in a new issue over on `pytest-dev/pytest`.

So should I remove the timing stuff?

ramnes:
I meant pytest-timeout when saying "upstream". Getting this into pytest may be a little too ambitious. :)

No rush, let's see what other think.

ep12:
Oh hang on. The weird times that got reported are because the setup and the teardown count by default. There is an option to only cause a timeout on the real run. The downside would be that setup and teardown cannot time out. They shouldn't cause trouble but as our tests may be doing something during the setup phase (connecting to X or something) we might also set a time limit there. Just because the real-world behaviour of the CI sometimes differs from the theoretical expectations. tych0 mentioned this while running mypy checks under all versions (`check´ for the `migrate` result). So even if we get correct timings that way, we might see tox being killed when it hangs in setup or teardown. And the configuration code wouldn't even be that much shorter when we want the TIMEOUT and/or the duration of the test run. Apart from that:
> No rush, let's see what others think.

test/conftest.py
m-col: What exceptions do you expect would be risen here?

test/conftest.py
m-col: OK 👍🏽

m-col:
Yeah OK, +1 from me, and I agree it'd be nice if the custom timeout indicators were merged into pytest-timeout.

test/conftest.py
ep12: Could be `AttributeError`s if `call.excinfo` is not of the type it should be according to `pytest`. Or if `call.excinfo.value` is not an Exception and hence doesn't have a `msg` attribute. So there shouldn't be any errors here. I still don't want to risk removing the `try: ... except`.

ep12:
It is fairly interesting to see that even the setup phase sometimes takes quite a long time on `pypy`.

test/conftest.py
```diff
@@ -61,6 +61,41 @@ def pytest_addoption(parser):
...
+        if isinstance(msg, str) and msg.startswith('Timeout >'):
+            timed_out = True
+    report.timed_out = timed_out
+
+
+def pytest_report_teststatus(report, **kwargs):
+    """Report test run duration, adapted from
+    https://github.com/pytest-dev/pytest/blob/38d8deb74d95077ebf189440ca047e14f8197da1/src/_pytest/runner.py#L202
+    """
+    d = report.duration
```
ep12, line 89: Although all CI tests completed successfully, another
```py
if not hasattr(report, 'duration'):
    return None
```
might be necessary here?

ramnes:
Thank you for opening the issue on the pytest-timeout repository. Let's see what these good folks say. LGTM otherwise.

ramnes (approved):

test/conftest.py
ramnes: Why?

ep12: `pypy3` failed locally while collecting tests. I tried to fix it using the additional check, now it's complaining about `pickle` that there is no protocol 5. `py39` collecting phase is okay, but the second test doesn't even start and doesn't time out neither. Something is badly wrong on my system. But that's not the point. The point is that I once saw an `AttributeError` (`CollectionReport` has no ... or something?) and this shouldn't cause any trouble. It's really weird, because I never had anything like this previously, even without the proposed check...

ramnes:
Mhh, looks like there's something wrong with the implementation actually. :)

Every test status line is outputted three times:

```
  test/test_bar.py::test_completion PASSED (0.001)                         [  0%]
  test/test_bar.py::test_completion PASSED (0.071)                         [  0%]
  test/test_bar.py::test_completion PASSED (0.000)                         [  0%]
```

Skipped tests get two lines:

```
  test/test_check.py::test_check_default_config SKIPPED (needs mypy)       [  3%]
  test/test_check.py::test_check_default_config PASSED (0.000)             [  3%]
```

ep12:
> Mhh, looks like there's something wrong with the implementation actually. :)
> 
> Every test status line is outputted three times:
> 
> ```
>   test/test_bar.py::test_completion PASSED (0.001)                         [  0%]
>   test/test_bar.py::test_completion PASSED (0.071)                         [  0%]
>   test/test_bar.py::test_completion PASSED (0.000)                         [  0%]
> ```
> 
> Skipped tests get two lines:
> 
> ```
>   test/test_check.py::test_check_default_config SKIPPED (needs mypy)       [  3%]
>   test/test_check.py::test_check_default_config PASSED (0.000)             [  3%]
> ```

That's a feature, not a bug! Actually, the first line is the `setup` line, the second line is `call`, number three is `teardown`. That wasn't obvious for me neither, but you can actually do `'PASSED (%.3f, %s)'` % (d, report.when) and then you see it.

test/conftest.py
ep12: Recreating the `tox -e pypy3` environment fixed that `pickle` error, but the second test doesn't start (same behaviour as with `tox -e py39`). This is also true for `upstream/master` and `upstream/master^`. Have to find out what's wrong with my setup.

ramnes:
Uh oh, okay. The multiple statuses make me confused. Do you think we could have a single line with the overall status and total duration?

test/conftest.py
tych0: Let's not use `except Exception` here. Please make it `AttributeError` if that's all you expect.

ep12: That piece of code is outdated. The latest version of the commit doesn't even have a `try` in it...

tych0: Ah ok :)

test/conftest.py
m-col: OK, no probs

tych0:
I agree, if we can get everything summarized in one line that would be ideal. Three lines is confusing.

ep12:
Good news: one line with the overall duration is possible!
Bad news: we need `_pytest` internals because the `log` option of some functions that are called is not configurable. This method relies on the internals so that we don't implement everything ourselves again. However, a minimal wrapper is a must.

tych0:
To save everyone else a few clicks, the output is:

```
  test/test_scratchpad.py::test_toggling[ScratchPadBaseConfic-xephyr0] PASSED (1.192+2.364+0.806=4.361) [ 32%]
```

Seems maybe reasonable to me.

ep12:
I switched to `method = thread` because this kills the tests more reliably (with `method = signal` the signal handler might not be called in really bad situations so the tests might still hang. In that case we want to get a traceback. Ideally, `pytest-timeout` would provide a `method = both` to `signal.raise_signal(SIGALRM)` hoping that the test times out gracefully. If that doesn't help, after a certain time the thread-based method could be used to kill the test run and print a traceback. Unfortunately, `pytest-timeout` doesn't provide the best of both worlds at the moment. As long as we use the basic `method = thread` the `TIMEOUT` doesn't get printed, only when switching back to `method = signal` we'd be able to spot it... But then we don't get a traceback in case of very serious hangs...

ep12:
Would you like to give some feedback on #2380? Is that what we'd like? @tych0 is there a way to manually cause a hang in xcffib? Seems like a recent update has fixed the problems on my system and now the tests don't hang on my sys anymore. I'd like to test the new `both` method which should result in a hard kill when a C extension hangs. Otherwise it should be soft so the test suite continues even after one test timed out.

tych0:
> @tych0 is there a way to manually cause a hang in xcffib?

Sure, put a sleep somewhere :). It's not even clear that it needs to be in xcffib...

ep12:
> > @tych0 is there a way to manually cause a hang in xcffib?
> 
> Sure, put a sleep somewhere :). It's not even clear that it needs to be in xcffib...

No. I wasn't really clear on that. The `time.sleep` will be interrupted by the `signal.raise_signal`. But I want to test something that can only be interrupted by the thread method (like C extensions that hang). It would good to see that the work I did to combine `method = signal` with `method = thread` indeed works out.

tych0:
>  No. I wasn't really clear on that. The time.sleep will be interrupted by the signal.raise_signal.

Why not install a signal handler that swallows the signal?

ep12:
> > No. I wasn't really clear on that. The time.sleep will be interrupted by the signal.raise_signal.
> 
> Why not install a signal handler that swallows the signal?

Good news: Thanks to that hint above I managed to add two tests to [ep12/pytest-timeout@timing_enhancements](https://www.github.com/ep12/pytest-timeout/tree/timing_enhancements) that indeed seem to confirm that `method = both` works as expected, i.e. gracefully *and* reliably. (pytest-dev/pytest-timeout#87)

ep12:
The pytest-timeout prs are still wip.

PR #2332, opened 2021-03-28
migrate: print message and remove backup file if no changes made

m-col:
Currently `qtile migrate` makes a backup of the config and prints
nothing if no changes were required. Let's make it clear that there were
no changes by printing out a message saying so, and don't leave the
backup around if it is identical to the main file.

### DIFF
--- a/test/test_migrate.py
+++ b/test/test_migrate.py
@@ -27,12 +27,13 @@ def test_migrate_default_config_noop():
         default_config = os.path.join(os.path.dirname(__file__), '..', 'libqtile', 'resources', 'default_config.py')
         config_path = os.path.join(temp, "config.py")
         shutil.copyfile(default_config, config_path)
+
+        before = hash_file(config_path)
         run_qtile_migrate(config_path)
+        after = hash_file(config_path)
 
-        original = hash_file(config_path)
-        migrated = hash_file(config_path+BACKUP_SUFFIX)
-
-        assert original == migrated
+        assert before == after
+        assert not os.path.exists(config_path + BACKUP_SUFFIX)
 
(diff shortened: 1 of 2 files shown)


test/test_migrate.py
```diff
@@ -15,24 +14,13 @@ def run_qtile_migrate(config):
...
-
-
 def test_migrate_default_config_noop():
     with tempfile.TemporaryDirectory() as temp:
         default_config = os.path.join(os.path.dirname(__file__), '..', 'libqtile', 'resources', 'default_config.py')
         config_path = os.path.join(temp, "config.py")
         shutil.copyfile(default_config, config_path)
         run_qtile_migrate(config_path)
-
-        original = hash_file(config_path)
```
tych0, line 32: I wonder if we should hash it before we migrate it, so we can make sure it's still a no-op?

m-col: [filecmp](https://docs.python.org/3/library/filecmp.html#filecmp.cmp) compares the contents of the files, so I'm not sure what hashing would also provide. Unless I misunderstand what you mean?

m-col: Oh, I see what you mean actually, OK, ignore the above

m-col: I just added it back, but reading through it on here again I think it's actually redundant, but perhaps I can make the test check for the existence of the backup file. If it's there, then `filecmp.cmp` returned `False` -> the files differ.

m-col: Alright try this out for size

tych0: LGTM, thanks!

PR #2333, opened 2021-03-28
CONTRIBUTING: add more docs about what we look for in PRs

ep12:
@tych0 I was lazy and even copied your commit message :rofl: from #2330.
@tych0 @m-col @flacjacket @ramnes whoever pushes releases to [pypi](https://pypi.org): We have `project_urls` in our `setup.cfg`. They have been there for two years I think, git blame it yourself if in doubt. But they don't show up on the pypi website. I am not sure why exactly that is, but maybe this is related: [project_urls is being ignored on PyPI](https://github.com/pypa/warehouse/issues/4809). So please use `twine` in the future.


ep12:
Maybe it would also be nice to configure the github issue and pull request templates. I skipped over the docs and I think it should be possible to add different templates for bug reports and feature requests. And to add a fancy YAML header they call it so that issues started with the feature request template get the `feature request` label and the bug reports `bug` and `unconfirmed` correspondingly. Also, the pr template could point at the contributing page to make sure people read it. However, I don't know if people actually get to see that if they go via their own repos -> create pull request where sometimes the commit message is used as the pr description.

tych0:
> @tych0 I was lazy and even copied your commit message rofl from #2330.

Ok, can you add some other notes about the rest of the changes and rearrangements?

docs/manual/contributing.rst
```diff
@@ -25,14 +25,21 @@ has two qualities:
...
 ============
 
 To get started writing code for Qtile, check out our guide to :ref:`hacking`.
-
 A more detailed page on creating widgets is available :ref:`here <widget-creation>`.
 
+.. important::
+
+    Use a separate **git branch** to make rebasing easy. Ideally, you would
+    ``git checkout -b <my_feature_branch_name>`` before starting your work.
```
tych0, line 41: I wonder if we shouldn't just link to a guide about git branching here, talking about git rebase -i, that git push -f is ok, etc?

tych0:
Overall this looks pretty good. I think you could e.g. split the https:// changes into separate commits, but it's minor. Noting that everything you did in the commit was intentional in the commit log would be good though.

ep12:
> Noting that everything you did in the commit was intentional in the commit log would be good though.

Do you mean it'd be okay to keep the protocol changes in the commit and mention them? Or do you mean the other points that were added? `git` stuff etc.?

tych0:
> Do you mean it'd be okay to keep the protocol changes in the commit and mention them?

Generally I would separate them, since they're unrelated.

> Or do you mean the other points that were added? git stuff etc.?

Every hunk in the commit should have some motivation for *why* it is necessary in the commit message.

ep12:
> > Do you mean it'd be okay to keep the protocol changes in the commit and mention them?
> 
> Generally I would separate them, since they're unrelated.
> 
> > Or do you mean the other points that were added? git stuff etc.?
> 
> Every hunk in the commit should have some motivation for _why_ it is necessary in the commit message.

Okay, will do that.

ep12:
@tych0 Should I change to `https` everywhere or should I omit links in python files? In the docs it is fine I think, but I don't really care about the links in docstrings in the source code.

tych0:
> Should I change to https everywhere or should I omit links in python files?

I'd probably leave the source files alone, yes.

docs/manual/contributing.rst
m-col: I had a look around and although there are lots of guides for various combinations of git commands, it might be more straightforward (and less confusing to new users) to just mention a few commands to start them on searching around themselves. We can add a sentence saying when `git rebase -i` might be useful and that `git push -f` can be used to update the commits on their PR. From there they might be good to go.

docs/manual/contributing.rst
```diff
@@ -48,7 +55,35 @@ to our `issue tracker <https://github.com/qtile/qtile/issues>`_ on GitHub.
...
+      *Please add unit tests* to ensure that your code works and stays working!
     * **Documentation** updates on an as needed basis.
+    * A ``qtile migrate`` **migration** is required for config-breaking changes.
+      See `migrate.py <https://github.com/qtile/qtile/blob/libqtile/scripts/migrate.py>`_
+      for examples and consult the `bowler documentation <https://pybowler.io>`_
+      for detailed help and documentation.
+    * **Code** that does not include *unrelated changes*. Examples for this are
+      formatting changes, replacing quotes or whitespace in other parts of the
+      code or "fixing" linter warnings popping up in your editor on existing
+      code. *Do not include anything like the above!*
```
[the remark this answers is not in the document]
ep12: Yeah I thought about that but wasn't quite sure if it would be too obvious. Also, this might encourage people to create pull requests with all sorts of linter fixes. It was decided that we don't want to pollute the git history. So we might want to leave that out for now and then put it in once the style guard (#2328) is doing its job.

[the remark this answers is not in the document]
m-col: Fair enough :)

m-col:
This is great. It all looks great, lgtm for once we decide on what git-related advice to include.

ep12:
I wrote some stuff about `git`, but totally unsure if that is helpful at all.

docs/manual/howto/git.rst
```diff
@@ -0,0 +1,158 @@
...
+  git checkout master
+  git pull upstream/master
+  git checkout feature-branch
+  git rebase upstream/master
+
+You will be asked to solve conflicts where your diff cannot be applied with
+confidence to the work that was pushed upstream. If that is the case, open the
+files in your text editor and resolve the conflicts manually. You possibly need
+to ``git rebase --continue`` after you have resolved conflicts for one commit if
+you are rebasing multiple commits.
```
ep12, line 131: Add a note that rebasing is harder when working off master without a feature branch?

m-col:
Sorry for the u-turn but actually I think having a git guide like this will likely lead to us having to maintain this as a bit of extra work, and perhaps it would be better to forward readers to other sources online. I like how you have organised it (like with the "I've been told to..." type of section headers) and I think it might fit quite well with linking to other places. So rather than one link to a general guide to git, we could have like "I've beent old I need to rebase" then maybe one sentence for a super light-touch description of what that is (without going into details of _how_) and then a link to elsewhere. Kind of like an FAQ at that point. What do you think? I'm just worried that if there are any points of confusion on it then we will be responsible for ensuring it is super clear and to maintain it and update it.

ep12:
Yeah, I get your point. Some things to take into consideration:
+ If we link to other pages, we have to still maintain something if the page is 404 at some point. If that happens, users would probably the first ones to notice and then would open issues about that.
+ If we have a very short practical guide (last commit) the information is stripped down. The point is: we would have to choose pages where the information is also to the point. We cannot just link to online versions of the `git` man page, that is too much information. Users wouldn't be able to find the essential bits and would be more frightened and confused.
+ What if we make the disclaimer really pop out? Like a big red box with _this is not meant to be a complete guide, if in doubt search the web_? That way we push the responsibility away, we don't claim anything. Yet we'd still have something short that would help users who know the vocab.

Just some ideas. Totally up to you.

ep12:
Opinions?

m-col:
Fair points, maybe links are a bad idea. Alternatively if we provide enough information to get users knowing what they'd need to search for themselves, then it'd be future proof. I guess that's what you've put in that guide.

> Like a big red box with this is not meant to be a complete guide, if in doubt search the web?

I think this is a good idea.

However, in its current form I think the 'short practical guide' in the last commit is quite long and has a wider scope than I think is necessary. I think it ought to be focussed exclusively on the common tasks users are asked to do as put of PRs. I think it would be enough to strip out the "What do I need to do to get a local copy of the qtile source code?" section and the " I want to change something and create a pull request" section.

ep12:
> Fair points, maybe links are a bad idea. Alternatively if we provide enough information to get users knowing what they'd need to search for themselves, then it'd be future proof. I guess that's what you've put in that guide.
> 
> > Like a big red box with this is not meant to be a complete guide, if in doubt search the web?
> 
> I think this is a good idea.

Done.
 
> However, in its current form I think the 'short practical guide' in the last commit is quite long and has a wider scope than I think is necessary. I think it ought to be focussed exclusively on the common tasks users are asked to do as put of PRs. I think it would be enough to strip out the "What do I need to do to get a local copy of the qtile source code?" section and the " I want to change something and create a pull request" section.

Done.

docs/manual/howto/git.rst
m-col: Nah you already say to use a feature branch elsewhere, should be fine

m-col:
This looks great as is, but there is a conflict in the readme. Please could you resolve and rebase? Then good to go.

m-col (approved):
LGTM, but I'll it to another pair of eyes to merge. Thanks again @ep12 :)

ep12:
Anyone out there reviewing this PR as @m-col suggested?

docs/manual/howto/git.rst
```diff
@@ -0,0 +1,104 @@
...
+
+  user wants to merge n commits into qtile:master from user:branch
+
+Right now you probably have one *remote* from which you can fetch changes, the
+``origin``. If you cloned ``qtile/qtile``, ``git remote show origin`` will spit
+out the *upstream* url. If you cloned your fork, ``origin`` points to it and you
+probably want to ``git remote add upstream https://www.github.com/qtile/qtile``.
+To try out somebody's work, you can add his fork as a new remote::
+
+  git remote add user https://www.github.com/user/qtile
```
elParaguayo, line 36: Interesting. I don't do it this way, I do:
```
git fetch upstream pull/<id>/head:<new-branch-name>
git checkout <new-branch-name>
```
That way I don't add additional remotes.

docs/manual/howto/git.rst
```diff
@@ -0,0 +1,104 @@
...
+
+I committed changes and the tests failed
+========================================
+
+You can easily change your last commit: After you have done your work,
+``git add`` everything you need and use ``git commit --amend`` to change your
+last commit. This causes the git history of your local clone to be diverged from
+your fork on GitHub, so you need to force-push your changes with::
+
+  git push -f your_username
```
elParaguayo, line 54: I find this confusing as I never include my username when pushing because it's set via `git config` so I end up doing `git push -f origin <branch name>`.

elParaguayo:
I've added a couple of review comments which are probably pretty minor. Agreed that it looks good otherwise.

docs/manual/howto/git.rst
m-col: Oooh! This maybe seems nicer... perhaps we should describe this approach too? Not sure I understand how it works actually. If the upstream is to full qtile/qtile URL, how does the ID get resolved? Is this GitHub being clever, or does this work with many/all git servers?

elParaguayo: So, yes, `upstream` is full qtile/qtile. ID doesn't get resolved, you enter it. So, to test this PR I did:
```
git fetch upstream pull/2333/head:pr2333
git checkout pr2333
cd docs
make html
```

m-col: Huh, well this is way better. @ep12 what do you think about using this instead for the docs?

docs/manual/howto/git.rst
```diff
@@ -0,0 +1,104 @@
...
+will find guides elsewhere to fix this problem, ideally by creating a branch and
+resetting your master branch to where it should be.
+
+I was told to squash some commits
+=================================
+
+If you introduce changes in one commit and replace them in another, you are told
+to squash these changes into one single commit without the intermediate step::
+
+  git rebase -i
```
elParaguayo, line 85: Running this command actually gives me an error. I think you usually have to say what you're rebasing against.

Typically, when squashing commits, I'd run something like:
```
git rebase -i <branch-name>~4
```
to show me the last 4 commits on my branch.

ep12: It probably depends on what you have checked out. Wildly guessing I'd say `git rebase -i` is a shortcut for `git rebase -i master` assuming you're on a feature branch. Not completely certain about that though...

docs/manual/howto/git.rst
ep12: Haven't used it, but sure if is better then this should be changed of course! Thanks for the hint btw. 😁

docs/manual/howto/git.rst
ep12: Actually, the username is the origin, i.e. I've set origin and upstream to qtile/qtile I think. Yeah, should definitely add an explanation to this...

ep12:
> @ep12 Is this still alive? I'd like get this merged soon but I think there are still some open comments.

@elParaguayo Yes, indeed. Have a look around if you're happy with these changes. Feel free to request more changes or merge it away from my long-term to-do list :smile:

elParaguayo:
@ep12 This is looking really nice. I've just made 2 small comments then I think it's good to go.

PR #2340, opened 2021-04-01
`CommandClient` refactoring of `cmd-obj` command

flacjacket:
The primary change here is in the 4th commit, which is to change the command client in the `cmd-obj` script to use the `CommandClient` rather than the `InteractiveCommandClient`. This allows the graph navigation to use the named functions `navigate()` and `call()` rather than using `__getattr__`, `__getitem__`, and `__call__`. There are also a couple clean ups in the implementation, for example, all command objects have a `cmd_doc` and `cmd_commands`, so just unconditionally call these commands, rather than checking if they exist.

### DIFF
--- a/libqtile/command/client.py
+++ b/libqtile/command/client.py
@@ -46,7 +46,7 @@ from libqtile.ipc import Client, find_sockfile
 class CommandClient:
     """The object that resolves the commands"""
 
-    def __init__(self, command: CommandInterface = None, *, current_node: GraphType = None) -> None:
+    def __init__(self, command: CommandInterface = None, *, current_node: Optional[CommandGraphNode] = None) -> None:
         """A client that resolves calls through the command object interface
 
         Exposes a similar API to the command graph, but performs resolution of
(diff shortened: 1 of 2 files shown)


ramnes:
> This allows the graph navigation to use the named functions `navigate()` and `call()` rather than using `__getattr__`, `__getitem__`, and `__call__`

Why do we want this?

flacjacket:
> > This allows the graph navigation to use the named functions `navigate()` and `call()` rather than using `__getattr__`, `__getitem__`, and `__call__`
> 
> Why do we want this?

I think one example is being better able to distinguish selecting an object from selecting a function on that object. I think some of the checking that is currently done (like is the `doc` command defined, or is the `commands` command defined) may exist because of the ambiguity in the interactive command client. (Maybe this is a failing of the interactive command client and we can change this... but that would probably break `lazy`) At least IMO, it also makes it a bit more clear where the navigation is being done, using the named functions, rather than indirectly via the dunder methods.

flacjacket:
I rebased this on top of #2349, since there are some extra changes that need to be made with that in place, and that seems like a more straightforward PR.

ramnes:
I don't have much opinion about the command graph in general, so I'll let your pull-requests to others, but still I cherry-picked one of the commits from this branch (see bd6e8f3) because I need it right now. :)

PR #2341, opened 2021-04-01
Do more aggressive type checks and apply fixes

flacjacket:
Setup the mypy testing to be a bit more aggressive in testing types,
including checking for unreachable code and non-sensical type
assertions.  Make changes based on this analysis to make the code pass
the type checks, including both getting rid of some of this dead code
and fixing the type annotations.  This also changes over to using the
__future__ annotations functionality to not have to handle the
not-yet-defined types as strings.

### DIFF
--- a/libqtile/sh.py
+++ b/libqtile/sh.py
@@ -23,12 +23,13 @@
 
 import fcntl
 import inspect
+import pprint
 import re
 import readline
 import struct
 import sys
 import termios
-from typing import List, Optional, Tuple
+from typing import Any, List, Optional, Tuple
 
 from libqtile.command import graph
 from libqtile.command.interface import (
@@ -295,7 +296,7 @@ class QSh:
     do_quit = do_exit
     do_q = do_exit
 
-    def process_line(self, line: str) -> str:
+    def process_line(self, line: str) -> Any:
         builtin_match = re.fullmatch(r"(?P<cmd>\w+)(?:\s+(?P<arg>\S+))?", line)
         if builtin_match:
             cmd = builtin_match.group("cmd")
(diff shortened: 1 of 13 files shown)


libqtile/sh.py
```diff
@@ -344,7 +343,4 @@ def loop(self) -> None:
                 continue
 
             val = self.process_line(line)
-            if isinstance(val, str):
-                print(val)
-            elif val:
-                pprint.pprint(val)
```
tych0, line 350: We lose json pretty printing with this, right?

tych0: Or, well, python object pretty printing.

flacjacket: Right, that might be the case, this is something that was flagged as dead because of the `-> str` annotation on `process_line`. But, we can relax the `process_line` type annotation if we think we can get Python objects returned here.

tych0: I suspect it's bogus, I'm getting pretty printing from somewhere at least,

```
screen.layout > info()
{'clients': ['Mozilla Firefox', 'urxvt', 'urxvt'],
 'current_stack': 1,
 'group': 'www',
 'name': 'stack',
 'stacks': [{'clients': ['Mozilla Firefox'], 'current': 0, 'split': False},
            {'clients': ['urxvt', 'urxvt'], 'current': 0, 'split': False}]}
```

flacjacket: That's good to know. I relaxed the type on `process_line` to return `Any` so the mypy test doesn't complain here, and added the pprint back in.

tych0:
Great, thanks!

tych0:
And welcome back :)

PR #2346, opened 2021-04-01
Add PaddingMixin to _TextBox

elParaguayo:
A user on qtile-dev group wanted ability to adjust vertical alignment of text widgets (see https://groups.google.com/g/qtile-dev/c/CYgAOQ3kXsc).

This PR uses the PaddingMixin to provide horizontal and vertical padding options.

NB this potentially breaks configs as, before the PR, users set horizontal padding with "padding" argument. Once the PaddinMixin is subclassed, "padding" impacts both horizontal and vertical padding.

A migrate function has been added to automatically adjust configs.

Hopefully I haven't over-complicated this too much!

Also, I haven't updated changelog yet but would appreciate any thoughts on this first.


elParaguayo:
I also need to write a test but I can see it's failing CI checks already. Doesn't like my migrate function by the look of it. Strange, as it works well on my machine.

m-col:
I've always been a little confused as to why those mixins weren't just part of the base widget. I'd be in favour of removing them as mixins and just let any widget accept any paddings. If we did go that way, it might be less painful (WRT config breakage) to make `padding` apply to the bar's direction (horizontal on horizontal bars) and not the other direction, and `padding_{x,y}` apply to either specifically (of course `padding_x` would be redundant on horizontal bars). Not sure what's best in that respect but breaking `padding` will affect probably every qtile user. Thoughts?

elParaguayo:
Interesting... I don't see any immediate issue with having the mixin be part of the base. I'm not sure I quite understand the need for both a Padding and Margin mixin...

I do think having "padding" being horizontal or vertical depending on the bar's orientation may be a bit confusing. However, you're right that the alternative could be that users' configs result  in ugly layouts to the extent they've set "padding".

That also reminds me that I haven't checked if any of the widgets set "padding" in their defaults.

tych0:
I also have never understood the difference between margin and padding (in any practical sense, I get that one is "outside the border" and the other "inside"), but if we could hoist the implementations to the widget base, that would be great.

elParaguayo:
Ok. Do we need both though? If we're doing an "outside the border" then that, to me, suggests the background shouldn't be painted in that area.

I can add both margin and padding to the base widget. It may take some time to work through the widgets to make sure they handle it correctly.

Also, thinking about a migrate function here, it may be a bad idea. For example: in this PR, I made a migrate function that changes `padding` to `padding_x` in users' config files. However, after this change, some people may validly want to use `padding` so the migrate would break that. There's no way to distinguish.

ramnes:
> Also, thinking about a migrate function here, it may be a bad idea. For example: in this PR, I made a migrate function that changes `padding` to `padding_x` in users' config files. However, after this change, some people may validly want to use `padding` so the migrate would break that. There's no way to distinguish.

Uh, that's a funny one. Should we add dates (or any other increasing id) to the migrations, and keep the last one applied in the configuration itself?

tych0:
> Ok. Do we need both though?

I don't think we need either, so I'm probably the wrong person to ask :)

> in this PR, I made a migrate function that changes padding to padding_x in users' config files

Can't you just leave the padding values as they are? I don't understand why a migration is necessary at all.

elParaguayo:
The problem I'm trying to deal with is that, in `_TextBox` widgets, `padding` sets the horizontal padding and there's no option for vertical padding as the text is centred vertically. However, there are instances when it seems to be necessary to add vertical padding (see the link in the first post for an example).

I thought I should use the `PaddingMixin` but, if I do that and try to use the `padding_y` attribute for vertical alignment, this falls back to whatever is in `padding`. So, without the migrate, users who have set `padding` would see the vertical alignment changed too.

The alternative would just be to add a some sort of `vertical_padding` parameter to `_TextBox` and leave the original `padding` untouched.

tych0:
Ah, I see. So it seems correct to do the `padding -> padding_x` migrate initially, but in future instances that may not be the case?

I'd say add the migration for now, and we'll figure out a way to handle this before the next release.

elParaguayo:
I'd expect the number of instances of people using `padding` in `_TextBox` widgets _after_ this PR is merged would be small as I doubt many people would want to touch the vertical alignment.

I'm also reluctant to merge `MarginMixin` and `PaddingMixin` into the base widget. Doing so would result in those attributes appearing in the docs when most widgets don't actually use them at all. That would just add confusion for users. If they're going to be added to the base class then widgets needed to be updated to use them.

If that's a route we want to go down then we can close this PR and I'll work on that separately. Will just take a bit more time!

Currently, `AGroupBox`, `GroupBox` and `Tasklist` use both `PaddingMixin` and `MarginMixin`. `Image` just uses `MarginMixin`.

elParaguayo:
Hmm. I had hoped to dynamically build a list `_TextBox` widgets but that fails where widgets have third party module dependencies such as `iwlib`. I may need to hardcode the list instead.

Also, this migrate function adds about 30 seconds to the tests which seems excessive. Need to do some more digging.

elParaguayo:
Also struggling with `bowler`'s selectors.

If users add widgets via `widget.TextBox` then it seems I need to use `select_method`. However, if they use `TextBox` (i.e. they import it directly) then I need `select_class`.

I don't currently understand the syntax well enough to make a custom selector so I'll keep looking but any pointers would be great.

libqtile/scripts/migrate.py
```diff
@@ -115,6 +119,54 @@ def new_at_current_to_new_client_position(config):
...
+                trailer< '(' class_arguments=any* ')' >
+            >
+            |
+            class_call=power<
+                'widget'
+                trailer< '.' '{widget}' >*
+                trailer< '(' class_arguments=any* ')' >
+                any* 
+            >
+        """.format(widget=widget)
```
elParaguayo, line 155: There's probably a neater way of doing this selector but I'm very new to `bowler`!

libqtile/scripts/migrate.py
```diff
@@ -115,6 +119,54 @@ def new_at_current_to_new_client_position(config):
...
+    def update_padding(ln, cap, fn):
+        for c in cap.get("class_arguments"):
+            for leaf in c.leaves():
+                if leaf.value == "padding":
+                    leaf.value = "padding_x"
+
+    migrations = []
+
+    textwidgets = [getattr(libqtile.widget, w) for w in libqtile.widget.__all__]
+    textwidgets = [widget.__name__ for widget in textwidgets if inspect.isclass(widget) and issubclass(widget, _TextBox)]
```
elParaguayo, line 138: Not convinced this is best approach. As noted in comments, this will skip widgets that have third party dependencies.

Is it better to have hard-coded list of `_TextBox` widgets?

tych0: It'll only skip things that have third party dependencies that aren't currently installed, right? I think that's fine - people won't have widgets in their configs that they don't have dependencies for.

elParaguayo: D'oh! Of course.

libqtile/scripts/migrate.py
m-col: We all are :)

PR #2347, opened 2021-04-01
Write that we aim to always support the last three versions of Python

ramnes:


docs/manual/install/index.rst
```diff
@@ -27,16 +27,18 @@ Installing From Source
 First, you need to install all of Qtile's dependencies (although some are
 optional/not needed depending on your Python version, as noted below).
 
-All of Qtile's supported versions of python can be found in the yml files Here_.
-There are not many differences between versions aside from python features you
+We aim to always support the last three versions of CPython, the reference
```
tych0, line 30: My only question is: how realistic of a goal is this, and why three? In particular, we didn't support three versions between 3.6 and 3.9 because we dropped support for 3.6 so we could get lazy imports. This discussion arose because we want to use the new walrus thingy. It seems kind of arbitrary to choose three, and perhaps two would be a better number, given that this is what our desired features list has included in the past?

The other thing is, we could also say a *minimum* of the last two versions. There's no reason to remove anything from CI if it's working and there's no shiny new language features we want to use.

m-col: > how realistic of a goal is this, and why three?

That's the big question :) Our worry is of course that some users will get left behind, so identifying how many people would be is important to knowing how far back we ought to support. As you said on IRC we'd still be supporting 3.6 if we tracked ubuntu LTS versions, and AFAICT nobody has complained about us dropping 3.6. So maybe nobody uses ubuntu LTS for their desktop? Maybe that's not that surprising... maybe the most "outdated" desktop that is reasonable to support is ubuntu proper, as we definitely should support that, and anyone else using other distros probably has the same Python version or a more recent version. Based on that I'd maybe be in favour of tracking Ubuntu and whatever Python version it uses for its default Python package. I think that is slightly more pragmatic than following Python versions because it directly follows the userbase.

ramnes: > In particular, we didn't support three versions between 3.6 and 3.9 because we dropped support for 3.6 so we could get lazy imports.

I'm not sure to get what you mean here. We do support at least three versions for quite some time and waited for 3.9 before dropping 3.6 to get the lazy imports. We could have dropped it day 1 if we didn't mind less than three versions.

> It seems kind of arbitrary to choose three, and perhaps two would be a better number, given that this is what our desired features list has included in the past?

Two is just as arbitrary as three, he! :) I think three is a good balance between new features and compatibility for most people. Three makes everyone happy: the "cutting-edge" guys, the "normal" people (from my experience I suspect that most people are one version behind) and those a bit late. I'm using Gentoo and still, I've been two versions behind quite often since the beginning of Python 3.x.

> The other thing is, we could also say a minimum of the last two versions. There's no reason to remove anything from CI if it's working and there's no shiny new language features we want to use.

I don't think the uncertainty brings any value. What we want to know is when we can allow ourselves to drop an old version, and users want to know which version of Python is guaranteed to work.

tych0: > So maybe nobody uses ubuntu LTS for their desktop?

I do :). focal (20.04) is on 3.8: https://packages.ubuntu.com/focal/python3

21.04 will have 3.9: https://packages.ubuntu.com/hirsute/python3

>  I'm not sure to get what you mean here. We do support at least three versions for quite some time and waited for 3.9 before dropping 3.6 to get the lazy imports.

Ah ha, I think I had re-written my memory. You're quite right.

> I don't think the uncertainty brings any value. What we want to know is when we can allow ourselves to drop an old version, and users want to know which version of Python is guaranteed to work.

Fair point. I'm happy to stick with three, then.

ramnes:
Let's go then :)

docs/manual/install/index.rst
ramnes: FWIW :)

image

Source: https://www.jetbrains.com/lp/python-developers-survey-2020/

I think we made the good choice with three versions.

m-col: Interesting, good to know!

PR #2349, opened 2021-04-02
`CommandClient` immediately invokes requested calls

flacjacket:
Rather than first resolving the call in `call()` and then invoking it is
`__call__`, just always immediately invoke the call when `call()` in
run.  To do this, add args and kwargs to the `call` function.  With this
change, the `CommandClient` will only hold command graph nodes, and not
the command graph calls.  This allows a few `asserts` to be dropped that
had been required to narrow the types.


flacjacket:
Currently the `CommandClient` is unused, but I think this will help clean up its use in #2340.

PR #2356, opened 2021-04-03
Unify `._items()` behavior

flacjacket:
Set a common return type for the `._items()` method and add the
annotation to the various command objects.  Some of the command objects
had been raising errors or returning `None` rather than an empty list
when there are no specific items to select.

This is built on top of #2340.


tych0:
I just merged #2357 which has this, looks like this one got stuck in the pypy ether in CI.

PR #2359, opened 2021-04-04
Remove `main` variable from default config

m-col:


tych0:
I guess we need to remove it from `libqtile/confreader.py` too :)

m-col:
Ah yip, added :)

Issue #2348, opened 2021-04-02
unit testing harness for restarts


tych0:
right now basically *all* of the restart code is untested, because we don't have a great harness for restarts. Might be nice to cook something up.

https://github.com/qtile/qtile/pull/2324#issuecomment-809271440

m-col:
If we dump state, restart then dump state again, would the two states be identical?

Issue #932, opened 2016-06-23
Special window to display errors


manipuladordedados:
If you use a bad config.py file, qtile will restart and reset to the default config. This can maybe cause confusion on the user, I know, qtile uses a system for saving what's wrong on a log file, but an simple window showing where the bad config is located on the config.py is not bad, right? Can we build it up?

gsaga:
`$ python path/to/qtile/config.py` should do it

himaaaatti:
I think we you should run below command.
$ python -m py_compile path/to/qtile/config.py

zordsdavini:
tail -f ~/.local/share/qtile.log

this should show init errors. (That's how I test my ```config.py```). I sugest to close this bug

PR #2360, opened 2021-04-04
send_notification: create popup when dbus is not available

m-col:
Currently send_notification will try to send a notification but fail
quietly if there is no notification server running. To get the message
to the user's eyes, it can instead fall back to a transient Popup
window.

This also changes the config-reading code to use this mechanism to
inform the user of a config error, rather than creating a new Config
object just so we can stick a TextBox in it to notify them.


libqtile/core/manager.py
```diff
@@ -104,11 +104,8 @@ def load_config(self):
             self.config.load()
             self.config.validate()
         except Exception as e:
-            logger.exception('Error while reading config file (%s)', e)
-            self.config = confreader.Config()
-            from libqtile.widget import TextBox
-            widgets = self.config.screens[0].bottom.widgets
-            widgets.insert(0, TextBox('Config Err!'))
+            msg = "Woops! Error in config: " + e.__class__.__name__
+            send_notification("Qtile Config", msg, timeout=20000)
```
ramnes, line 108: Let's make this consistent with the error in cmd_validate_config maybe?

libqtile/core/manager.py
```diff
@@ -104,11 +104,8 @@ def load_config(self):
             self.config.load()
             self.config.validate()
         except Exception as e:
-            logger.exception('Error while reading config file (%s)', e)
```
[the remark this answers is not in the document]
m-col: Ah yeah, good idea

libqtile/popup.py
```diff
@@ -166,3 +171,22 @@ def hide(self):
...
+    popup.draw_text()
+    popup.place()
+    popup.unhide()
+    popup.draw()
+
+    if timeout:
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            return
```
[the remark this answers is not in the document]
m-col: 👍🏽

libqtile/core/manager.py
m-col: Just to confirm, you mean the message itself, right?

If so, do you prefer this form (providing the error class) or the cmd_validate_config form (sending the full trace)? I'm easy either way, though I think the former results in nicer notifications and I assume users of cmd_validate_config aren't using the notification window itself for thorough debugging.

ramnes: Yeah just the message. Personally I like to have the full trace directly so that I precisely know what I have to debug when it happens and do not need to look somewhere else, but I mostly want consistent stuff. :)

m-col: Well your actual use case is higher priority than my hypothetical so I'll go that way ;)

ramnes (approved):
LGTM

m-col:
Just had a popup not go away so converting to draft until it gets resolved.

libqtile/utils.py
```diff
@@ -194,9 +194,8 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None):
     https://developer.gnome.org/notification-spec/
     """
     if not has_dbus:
-        logger.warning(
-            "dbus-next is not installed. Unable to send notifications."
-        )
+        from libqtile.popup import send_popup
+        send_popup(f"<b>{title}</b>\n{message}", timeout=timeout)
```
tych0, line 198: Can we make this configurable? I basically never want popups from any application ever :)

m-col: Do you ever get notifications through `send_notification`? You'll likely only ever see this if your config is wrong when you try to load it. I don't think you'll be using `send_notification` at all currently unless you have a notification server running, in which case you'll not get any popups.

tych0: No, I don't run any libnotify servers or anything like that. I guess what I'm worried about is people starting to use send_notification() for other stuff and that I'll start getting popups all of the sudden. But I can complain then, instead of now :)

m-col: Please do -- I do not want `send_notification` to become more than something explicitly triggered by custom user code. E.g. in config functions or the one or two `Qtile.cmd_` methods that use it.

m-col:
There was no bug, I was just being a bit dim and didn't notice the event loop warning printed to the log. Basically if the event loop isn't running yet then a timer cannot be set to hide the window, so in some instances you need to click it to hide it.

Anyway I rebased and tidied a couple things, so I've un-drafted.

ramnes:
Maybe worth a test or two?

m-col:
I added a test for `send_popup`. It would be nice to also add a test in the future when trying to restart with a faulty config, but for that proper restart handling in the tests is required #2348

libqtile/utils.py
m-col: I'm changing this so rather than send_notification falling back to send_popup, Qtile.load_config will itself deal with the fallback, so there is no chance of popups appear on account of having no libnotify server running. Otherwise like you hint at, popups will be seen as a fallback for these but it's only for this one case :)

m-col:
I wasn't completely happy with the way I'd done this so I've re-written it and I think it's better. Two commits: one just adding `send_popup`, and _not_ adding it to `send_notification`, and one using these two functions to notify the user upon config errors, instead of the current approach of overwriting the current `qtile.config` with the default and inserting a TextBox into its bar.

By not having `send_popup` used by `send_notification`, one can easily choose to never see popup windows (unless they get config errors).

There is one unresolved issue. We need to check that `send_notification` fails before falling back to `send_popup`, but in the case that `dbus_next` is available and the event loop is running, we can't know if the notification will fail due to there being no server because of the async `_notify` approach. Is it possible to check the presence of a server, or instead notify synchronously? cc @elParaguayo

elParaguayo:
This is ugly and untested but something like this?
``` diff
diff --git a/libqtile/utils.py b/libqtile/utils.py
index 2d6c7003..661583cd 100644
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -23,6 +23,7 @@ import asyncio
 import glob
 import importlib
 import os
+import time
 import traceback
 from collections import defaultdict
 from collections.abc import Sequence
@@ -212,7 +213,11 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None):
         logger.warning("Eventloop has not started. Cannot send notification.")
         return -1
     else:
-        loop.create_task(_notify(title, message, urgency, timeout, id))
+        task = loop.create_task(_notify(title, message, urgency, timeout, id))
+        while not task.done():
+            time.sleep(0.1)
+
+        return task.result()

     return id

@@ -239,11 +244,14 @@ async def _notify(title, message, urgency, timeout, id):
     if msg.message_type == MessageType.ERROR:
         logger.warning("Unable to send notification. "
                        "Is a notification server running?")
+        id = -1

     # a new bus connection is made each time a notification is sent so
     # we disconnect when the notification is done
     bus.disconnect()

+    return id
+
```

It's ugly as hell, though.

m-col:
That is pretty ugly ;) With the old `gi` version of the `send_notification`, we could do this:
```py
gi.require_version("Notify", "0.7")
from gi.repository import Notify
Notify.init("Qtile")
info = Notify.get_server_info()
```
and `info` would us information about any running notification server. This is great for knowing whether to continue with sending it. I'm slowly looking around dbus_next to see how to do the equivalent and then see where/if it makes sense to go, but any pointers would be appreciated as I've not done any dbus work.

elParaguayo:
You can still use dbus-next to get the server info, but it would be done asynchronously. Is that really a problem though? You can check the status/try to send a notification and attach a callback to send the popup if the server's not available.

m-col:
Looks like I could use `utils._send_dbus_message` which is actually ideal. Just gotta figure out all these weird arguments :)

elParaguayo:
You've lost me. What can you do by using that function directly (which is still asynchronous) that you can't do by trying to send the notification and testing for success there?

m-col:
Not a big difference, just that `send_notification` creates a task and returns, whereas I can await `_send_dbus_message` and see the result directly so that I don't have to do a polling loop until the task is done :) I don't want to change the behaviour of `send_notification`, I just want to check it works to know whether to popup or not.

elParaguayo:
OK - but you can only await from an async function so it still sounds like you need some extra steps.

That said, you know what you're doing so I'll stop replying unless you actually need my help!

m-col:
Ha, I actually do not know what I'm doing, I've never done any dbus or even
asyncio coding before. If you are willing to help then that would actually be
much appreciated :)

So my await approach is off the table then. Do you have any other ideas that
ideally don't involve changing (much of) send_notification?

elParaguayo:
OK. I think there needs to be some compromise in the code somewhere.

Again, this is untested:
``` diff
index 770a4b4d..eb9c2689 100644
@@ -116,8 +116,10 @@ class Qtile(CommandObject):
         except Exception as e:
             logger.exception('Error while reading config file (%s)', e)
             msg = str(error.__context__)
-            if send_notification("Configuration error", msg) < 0:
+            def fallback():
                 send_popup(f"<b>Configuration error</b>\n{msg}")
+            send_notification("Configuration error", msg, on_fail=fallback)
+

         if hasattr(self.core, "wmname"):
             self.core.wmname = getattr(self.config, "wmname", "qtile")  # type: ignore
diff --git a/libqtile/utils.py b/libqtile/utils.py
index 2d6c7003..69751ec6 100644
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -188,7 +188,7 @@ def lazify_imports(registry, package, fallback=None):
     return __all__, __dir__, __getattr__


-def send_notification(title, message, urgent=False, timeout=10000, id=None):
+def send_notification(title, message, urgent=False, timeout=10000, id=None, on_fail=None):
     """
     Send a notification.

@@ -210,9 +210,16 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None):
         loop = asyncio.get_running_loop()
     except RuntimeError:
         logger.warning("Eventloop has not started. Cannot send notification.")
+        if on_fail:
+            on_fail()
         return -1
     else:
-        loop.create_task(_notify(title, message, urgency, timeout, id))
+        def callback(task):
+            if task.result() < 0 and on_fail:
+                on_fail()
+
+        task = loop.create_task(_notify(title, message, urgency, timeout, id))
+        task.add_done_callback(callback)

     return id

@@ -239,11 +246,14 @@ async def _notify(title, message, urgency, timeout, id):
     if msg.message_type == MessageType.ERROR:
         logger.warning("Unable to send notification. "
                        "Is a notification server running?")
+        id = -1

     # a new bus connection is made each time a notification is sent so
     # we disconnect when the notification is done
     bus.disconnect()

+    return id
+

 def guess_terminal(preference=None):
     """Try to guess terminal."""
```
At least there are no `sleep` calls in this one!

m-col:
Ooh yes this is actually really nice! I'll stick it in and see how it fares. Thanks!

m-col:
OK I think this is pretty good now and is working quite well. There is a problem with popup windows not showing text under Wayland but that's a separate issue that this PR identified rather than introduced, so I'll leave fixing that for another PR. Thanks @elParaguayo for the help!

Ready for review I think!

m-col:
Looks like this has been asked for before https://github.com/qtile/qtile/issues/932

m-col:
I started working on config reloading and it's working pretty good. I've added the commit here for it ("wip") because it's upon this PR as well as the commit from #2530. To get finalizing configurables working cleanly I've made widgets cancel any scheduled tasks in `finalize` which fixes #2509 and (potentially other issues discussed in https://github.com/qtile/qtile/issues/2557).

One issue is left remaining which is why I've pushed, in case anyone has any bright ideas. The layouts do not seem to be changing when reloading the configs.

I'll also have to make some tests for this to make sure it does exactly what is expected.

m-col:
I've added a big test and updated the docs so this PR is ready to go now if anybody is up for reviewing it. It seems to be working pretty well too!

tych0:
One question I have is: is there some specific set of things that I as a developer need to watch out for when adding new features in the face of config reloading? I guess it's really "have a way to clear your state, and put that hook in reload_config()"?

m-col:
Yeah, that's pretty much it. If it's something that one would expect to be
reload-able then they'd have to make sure that it is. Many of the basic settings
such as `auto_fullscreen` are very simple, as when used they are always
referenced via e.g `self.qtile.config.auto_fullscreen`. More complicated
configurable things such as key bindings would need a way to clear their state
and do so in reload_config, like you describe.

m-col:
Just pushed a fix for a small problem with the scratchpad commit, but other than that this has been working great during the past few days of running with this patch applied and some config reloading through the day. It's actually really nice.

elParaguayo:
I get blank popups with this one. Have tested via  the `send_popup` function as well as creating a `Popup` manually. In both cases I just get a black window with no text.

m-col:
Did you get blank popups in wayland before or did this make them blank?

elParaguayo:
I noticed it when working on the StatusNotifier widget. Originally, I had errors which I "fixed" by making the same change you did here:
https://github.com/qtile/qtile/blob/cdfeb91bb6c0263faf6c5d67b96024fbc8c9c874/libqtile/backend/wayland/window.py#L634-L639

At that point I had blank popups so I don't think it's an issue caused by this PR but I haven't dug into where it's going wrong yet,

m-col:
OK cool thanks for the info. In that case I don't think we should let it hold this back, as long as the window does appear then these changes are working.

elParaguayo:
Sorry - should have been really clear - the issue I was referring to was on Wayland and is the point you made in one of your earlier comments. 

The popups work fine on X11.

m-col:
There is some potentially strange behaviour that I'm not certain how to deal with. What's the expected behaviour if you change your groups?

I don't use dynamic groups at all so when I reload I don't want any old groups around. Currently, through `QtileState` old groups are kept around as the restoration of state is done the same way as when restarting. This seems unexpected. What was the use case for keeping old groups around when restarting?

elParaguayo:
This is looking nice. Not seeing any particular issues in testing. Reloading works nicely.

I agree, the question about what to do with groups is interesting. If I remove groups from my config and reload the config, the old groups are still there. I understand why that's the case but some users may not. 

I use the basic default for my groups "123456". If I put windows in groups 1-4, remove groups 4-6 from my config and reload, this  is the result:
- Groupbox widget still shows groups 1-6
- Group 4 is shown as having no windows and I can't access it via any existing group

Once I add groups 4-6 back to the config and reload then group 4 is accessible again and the window is still there.

I'm not sure what the best answer is here but here are some thoughts:
- Users will probably expect that, if they remove groups from the config, they won't still see those groups in the groupbox widget
- If there is a window in a removed group then we could either move window to a different group (but how do we pick which one?) or alert the user that there's an "orphaned" window (could we do this _before_ the config is actually reloaded? Probably a lot of effort to implement)

elParaguayo:
Also, this doesn't solve the bug in #2620

m-col:
I've resolved all of the groups problems, which ultimately were odd side-effects of the way groups were being persisted across reloads. I don't think it makes sense to persist groups across reloads the way they persist across restarts, as the implication of this PR is that cmd_reload_config is used for reloading the config and cmd_restart is used for updating the runtime libqtile following package/code update. Windows that are on groups that are removed are moved to the new current group so it is very clear for the user that they may need to place it to whatever group they want.

If people agree with that approach (groups are refreshed completely from the config when you reload that config), then this PR is ready to go!

elParaguayo:
As per our chat, here's a diff to solve the delay when waiting to see if there's an available notification server:
``` diff
diff --git a/libqtile/utils.py b/libqtile/utils.py
index 0c3203b4..34ff8020 100644
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -264,6 +264,11 @@ def send_notification(title, message, urgent=False, timeout=10000, id=None, on_f

 async def _notify(title, message, urgency, timeout, id) -> bool:
     """Returns True upon success, False upon failure."""
+    can_notify = await _has_notification_server()
+
+    if not can_notify:
+        return False
+
     notification = ["qtile",  # Application name
                     id,  # id
                     "",  # icon
@@ -292,6 +297,27 @@ async def _notify(title, message, urgency, timeout, id) -> bool:
     return msg.message_type != MessageType.ERROR


+async def _has_notification_server():
+    """Checks for the existence of 'org.freedesktop.Notifications'."""
+    bus, msg = await _send_dbus_message(True,
+                                        MessageType.METHOD_CALL,
+                                        "org.freedesktop.DBus",
+                                        "org.freedesktop.DBus",
+                                        "/org/freedesktop/DBus",
+                                        "ListNames",
+                                        "",
+                                        [])
+
+    bus.disconnect()
+
+    if msg.message_type != MessageType.METHOD_RETURN:
+        return False
+
+    names = msg.body[0]
+
+    return "org.freedesktop.Notifications" in names
+
+
 def guess_terminal(preference=None):
     """Try to guess terminal."""
     test_terminals = []
```

m-col:
Brilliant, thank you for that, I've folded it in.

libqtile/core/manager.py
```diff
@@ -106,28 +107,23 @@ def __init__(
...
             self.config.load()
             self.config.validate()
         except Exception as e:
             logger.exception('Error while reading config file (%s)', e)
-            self.config = confreader.Config()
-            from libqtile.widget import TextBox
-            widgets = self.config.screens[0].bottom.widgets  # type: ignore
-            widgets.insert(0, TextBox('Config Err!'))
+            on_fail = partial(send_popup, f"<b>Configuration error</b>\n{str(e)}")
+            send_notification("Configuration error", str(e), on_fail=on_fail)
```
elParaguayo, line 118: Should we use the same approach in `cmd_restart`? There's only a notification in that scenario with no fallback to a popup.

m-col: Oh yeah, good point.

m-col: Actually it's beginning to look like `send_popup` should just be part of `send_notification` as a general fallback in the case where a dbus notification server is not present. In which case, it might be worth adding a config option `disable_popups` for people like @tych0 who don't want any feedback if a config change triggers an error. Thoughts?

elParaguayo: Hmm. Yeah, I can see the reasoning for that. I would think that `disable_popups` (might need to change the name as some `Popup`s are still ok) should be `True` by default. If a user starts with an invalid config they'll get dropped into the default config and that way it's very obvious that something's wrong (i.e. no need for a popup). However, if they make changes to their config and try restarting, the popup would be helpful.

m-col: How about `disable_notifications`, and it be used exclusively by `send_notification` (whether it uses `send_popup` or dbus)?

elParaguayo: Would that turn off notifications and popups? I think Tych0's objection was to the popups but a notification may still be acceptable.

m-col: In that case it could be set to `False`, if the user wants them and has a notification server running.

elParaguayo: Ah, yes. Fair enough.

elParaguayo:
Only one comment from me: When I put a bug in my config and restart, the notification I get (via Notify) says "Configuration Error - None".

This is because you're using `str(error.__context__)` for the notification content. `__context__` seems to be empty so you may want to use `str(error)` instead.

elParaguayo:
Hmm. Still some bugs here.
`reload_config` didn't prevent a reload with a bad config (just lost my bar contents) and I ended up with 2 of the same scratchpad on top of eachother.

m-col:
> reload_config didn't prevent a reload with a bad config (just lost my bar contents) and I ended up with 2 of the same scratchpad on top of eachother.

elParaguayo:
Awesome work on this one. Thanks.

PR #2362, opened 2021-04-04
Some improvements to the config-reading code

m-col:
I'm working on live reloading of the config and this was some of the prep work. I've paused here so that I can continue with both these commits and https://github.com/qtile/qtile/pull/2360 merged together. Also I can get some comments on these in the mean time :)


tych0:
All the patches seem fine besides `Config: validate using Qtile.core`, which seems like abstraction for abstraction's sake. Has anyone verified that there's any similar concepts in wayland that we need to future proof for, and what the wayland implementation would look like? If not, I'd prefer less indirection.

m-col:
Yeah fair point. Many (most?) desktop wayland compositors use xkbcommon for key mapping and tbh I would expect our own wayland backend to use `libqtile.xkeysyms` too, so the available keys will be common between that and the x11 backend. Definitely unnecessary.

ramnes (approved):
LGTM

PR #2364, opened 2021-04-06
Add a restart arg to restart qtile from terminal.

lkjell:
Making restart qtile from terminal a bit easier than to dig into the documents.

### DIFF
--- a/docs/manual/commands/shell/index.rst
+++ b/docs/manual/commands/shell/index.rst
@@ -10,6 +10,7 @@ repository are also documented below.
     :maxdepth: 1
 
     qtile start <qtile-start>
+    qtile restart <qtile-restart>
     qtile shell <qshell>
     qtile cmd-obj <qtile-cmd>
     qtile run-cmd <qtile-run>
--- a/docs/manual/commands/shell/qtile-restart
+++ b/docs/manual/commands/shell/qtile-restart
@@ -1,0 +1,10 @@
+.. _qtile-restart:
+
+===========
+qtile restart
+===========
+
+This will restart qtile. 
+
+This is an alias to ``qtile cmd-obj -o cmd -f restart``.
+Check ``qtile cmd-obj --help`` or ``qtile restart --help`` for more information.
--- a/docs/manual/commands/shell/qtile-restart.rst
+++ b/docs/manual/commands/shell/qtile-restart.rst
@@ -1,8 +1,8 @@
 .. _qtile-restart:
 
-===========
+=============
 qtile restart
-===========
+=============
 
 This will restart qtile. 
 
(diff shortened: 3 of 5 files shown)


tych0:
I think if we want to support this, let's implement more general alias functionality a la. git. I'd rather not have a bunch of hard coded aliases in the upstream implementation.

lkjell:
I cannot think of another cmd that needs an alias. Perhaps `qtile shutdown` or `qtile logout` in the future. A possibility is to have a tip on how to restart qtile when issuing `qtile --help`.

tych0:
> I cannot think of another cmd that needs an alias.

But someone else can :)

lkjell:
Fair but until somebody want to make that happen this PR fixes a problem. The scope of this PR fixes right now is: How to restart qtile from the terminal in an intuitive way without having to read the docs?

I believe with a general alias functionality you still need to know the full command to restart qtile. And you need to configure it somewhere.

tych0:
> Fair but until somebody want to make that happen this PR fixes a problem. The scope of this PR fixes right now is: How to restart qtile from the terminal in an intuitive way without having to read the docs?

It also adds an API that we have to maintain going forward, which is not easily covered by qtile-migrate. I appreciate that restarting is important, but there is ongoing work to make it less important, and I'd rather spend energy moving that forward. There are several examples for `qtile cmd-obj`, perhaps we should add an example of restarting there?

ramnes:
It's funny because I was thinking about this exact feature like two days ago. I'm definitely interested in making more Qtile commands accessible as subcommands of our entrypoint but I'd want something much more generic that we can replicate for other things. I think most of the functions listed in `qtile cmd-obj -o cmd` (i.e. first level / manager's cmd_ functions) would make sense as subcommands.

m-col:
> It's funny because I was thinking about this exact feature like two days ago.

Actually same! That, and having nice command completion for `zsh` (and whatever else), would make using the cli much nicer.

> I think most of the functions listed in qtile cmd-obj -o cmd (i.e. first level / manager's cmd_ functions) would make sense as subcommands.

I agree.

General +1 for this discussion. If we do go down the route of revamping the cli then another idea: we could drop the flags and do everything positionally i.e. `qtile cmd-obj -o window -f bring_to_front` could just be `qtile cmd-obj window bring_to_front` (or even `qtile window bring_to_front`). Hell, a simple dict of aliases could get us to `qtile w bring_to_front`. Just spit-balling here.

ramnes:
I like this

tych0:
I'm not opposed to exposing this stuff in a nicer way. However, I'd caution against having everything be a top level command. You don't want the output of `qtile --help` to show `qtile start` at the same level as `lazy.toscreen` or whatever, they are different levels of importance.

That said, I don't think cmd-obj is particularly intuitive (WTF is a cmd-obj? :D), so adding some new subcommand to invoke things in the lazy tree that has a much better UX is a +1 from me. We could even mark cmd-obj as hidden to sort of "deprecate" it once we have this new thing.

With that said, I'll close this, since we all seem to want something more general.

PR #2365, opened 2021-04-06
layouts: drop explicit names

tych0:
We have code that names these things in Layout.__init__, and have gotten
these wrong a few times (e.g. 9a947d5181f491979758ee22943127e844129c40 or
the xmonadtall vs. MonadTall discrepancy in this commit).

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -29,16 +29,7 @@ from libqtile.command.base import CommandObject, ItemT
 
 class Layout(CommandObject, configurable.Configurable, metaclass=ABCMeta):
     """This class defines the API that should be exposed by all layouts"""
-    @classmethod
-    def _name(cls):
-        return cls.__class__.__name__.lower()
-
-    defaults = [(
-        "name",
-        None,
-        "The name of this layout"
-        " (usually the class' name in lowercase, e.g. 'max')"
-    )]  # type: List[Tuple[str, Any, str]]
+    defaults = []  # type: List[Tuple[str, Any, str]]
 
     def __init__(self, **config):
         # name is a little odd; we can't resolve it until the class is defined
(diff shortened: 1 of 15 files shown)


libqtile/layout/base.py
```diff
@@ -29,10 +29,6 @@
 
 class Layout(CommandObject, configurable.Configurable, metaclass=ABCMeta):
     """This class defines the API that should be exposed by all layouts"""
-    @classmethod
-    def _name(cls):
-        return cls.__class__.__name__.lower()
-
     defaults = [(
         "name",
```
ramnes, line 33: I guess we should drop this as well?

tych0: I think this gets it in the defaults generated by the docs (albeit with a value of "None", so maybe not all that useful), I could go either way.

ramnes: I think it's just noise in the docs

tych0: Always happy to delete more code :)

ramnes:
Should we add a migration? We're basically removing a "feature" here, although I'm not sure it's used by anyone.

tych0:
It should still work fine, since the names are autogenerated in layout.__init__. we use exactly this same strategy for widgets, IIUC

ramnes:
Nah, my point is about users that would have something like this in their config:

```python
layouts = [
    Max(name="fullscreen"),
    ...
]
```

tych0:
Right, won't that still work?

ep12:
> Right, won't that still work?

I think so. You don't seem to be intending to drop custom names, right? I need them, so unless that is removed, everything looks fine.

tych0:
> I think so. You don't seem to be intending to drop custom names, right? I need them, so unless that is removed, everything looks fine.

Right, this just moves to auto-generating the name by default (similar to how we do for widgets).

ramnes:
Okay, I don't understand how but it still works. :)

PR #2367, opened 2021-04-07
Add an example of how to restart qtile

lkjell:
Example on how to restart qtile under `qtile cmd-obj --help`.

As commented at https://github.com/qtile/qtile/pull/2364#issuecomment-814323846 .
Agree that `cmd-obj` should be renamed as something else. Unfortunately `run_cmd` is used.


tych0:
Great, thanks!

PR #2369, opened 2021-04-09
Enable live loading of config via Qtile.cmd_load_config

m-col:
This converts Qtile.load_config to Qtile.cmd_load_config to that the
configuration file can be reloaded during runtime. Many things work
right away without much/any adapting, with the exception of the DGroups
which needs a bit of toying with to be happy being reconfigured.

(draft)

This sits atop #2360, as that PR stops `Qtile.load_config` from overwriting and losing the config instance when it fails to load/validate the config. The second commit is the changes I've made so far towards this goal.

I'm posting this a bit early to get some feedback before I delve much deeper. I'm not very familiar with DGroups; I've always run with a fixed number of groups and it's always 'just worked' so I've never had to read into the group code. It seems like they are the only thing left that need some adjusting so that reloading the config will yield a non-surprising result. So, first question, what would you expect to happen if you reload your config under different group conditions (static groups, dynamic but not changed, dynamic but changed)? Should we clear groups completely and start fresh every time the config is loaded? This might be simple, but we may have to then deal with moving windows from old ones to new ones. Then again, we may have to do that anyway. I'm a tad stuck here so open to ideas and suggestions!

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -36,7 +36,7 @@ import xcffib.xinerama
 import xcffib.xproto
 
 import libqtile
-from libqtile import confreader, hook, ipc, utils, window
+from libqtile import hook, ipc, utils, window
 from libqtile.backend.x11 import xcbq
 from libqtile.command import interface
 from libqtile.command.base import CommandError, CommandException, CommandObject
@@ -103,12 +103,9 @@ class Qtile(CommandObject):
         try:
             self.config.load()
             self.config.validate()
-        except Exception as e:
-            logger.exception('Error while reading config file (%s)', e)
-            self.config = confreader.Config()
-            from libqtile.widget import TextBox
-            widgets = self.config.screens[0].bottom.widgets
-            widgets.insert(0, TextBox('Config Err!'))
+        except Exception as error:
+            logger.exception('Error while reading config file (%s)', error)
+            send_notification("Configuration error", str(error.__context__))
 
         self.core.wmname = getattr(self.config, "wmname", "qtile")
 
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -81,7 +81,7 @@ class Qtile(CommandObject):
         self.widgets_map: Dict[str, _Widget] = {}
         self.groups_map: Dict[str, _Group] = {}
         self.groups: List[_Group] = []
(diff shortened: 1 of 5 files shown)


lkjell:
Need some clarification on what live loading means.

1) Automatic reloading the config file when saved.
2) Reload config files without restarting qtile

I am not a fan of 1 since I want to control when to reload the config.
Beside there are other tools better suited for this e.g. hooks from editor.

Not sure what reloading config files without restarting qtile solves.
Right now I am restarting qtile to reset layout.

m-col:
Definitely not 1 -- I agree, that'd be super annoying.

It would be 2. Restarting Qtile and completely closing down and starting up a
new Qtile process is overkill and clunky if all you want to do is add a widget
or remove a key binding or whatever. This approach to reloading your config
would replace restarting the whole process so that much less has been done, and
the update process can be more easily controlled. It's also much nicer on the
eyes (we can avoid all the windows popping out of place for instance) and faster
(restarting for me on one machine takes a few seconds). I'd counter with: Not
sure what restarting the entire process solves, when we just want to reload the
config file.

lkjell:
Restarting ensure that the state is predictable. There is also less maintenance. With live reloading you need to unload then load. 

To make it clean and without any side effects you need to unload everything.

ramnes:
> So, first question, what would you expect to happen if you reload your config under different group conditions (static groups, dynamic but not changed, dynamic but changed)? Should we clear groups completely and start fresh every time the config is loaded? This might be simple, but we may have to then deal with moving windows from old ones to new ones. Then again, we may have to do that anyway. I'm a tad stuck here so open to ideas and suggestions!

I'd expect some best effort to make the windows move only if something has changed in their group configuration making them irrelevant here.

m-col:
> Restarting ensure that the state is predictable. There is also less maintenance. With live reloading you need to unload then load.
> 
> To make it clean and without any side effects you need to unload everything.

I totally hear you, all good points. I also have an eye towards wayland (#2376) and exec-ing into a whole new process won't be feasible with that backend because the clients would not persist. Rather than in X, where Qtile is "just" the window manager but the client connections are made to the X server, in wayland Qtile will act as both server and window manager, so closing Qtile means leaving the full session.

m-col:
Closing as this is superceded by  #2360

PR #2371, opened 2021-04-10
Fix BatteryIcon reconfigure issue

elParaguayo:
The BatteryIcon widget's icon changes size if the widget is reconfigured (e.g. via `cmd_reconfigure_screens`). This is because
the icon size is adjusted by the value of `image_padding`. This is zero when the widget is initialised but has a larger value when the
widget is reconfigured.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -478,6 +478,7 @@ class BatteryIcon(base._Widget):
 
     def _configure(self, qtile, bar) -> None:
         base._Widget._configure(self, qtile, bar)
+        self.image_padding = 0
         self.setup_images()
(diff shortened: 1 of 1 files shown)


m-col:
Ah yeah makes sense. Thanks for this! LGTM pending tests.

elParaguayo:
Ideally, I'd remove the line `self.image_padding = 0` from `__init__` but it breaks the tests because the tests don't call `_configure`!

m-col:
I was just thinking something similar -- as reconfiguring screens has become a thing, and in the not-so-distant future reloading of the config file might become a thing, `_configure` is becoming more involved. We should probably add some tests for that at all levels. I'll add it to the tests discussion I made and nobody has used yet!

PR #2372, opened 2021-04-10
Fix Systray reconfigure bug

### LINKED ISSUE #2313: Turning screen off and on again creates some issues
# Issue description

I have observed an **issue with my systray and the assigment of groups to specific screens after turning off all my screens and then turning one or more back on**. I don't know whether this issue only shows up when using multiple screens.

*A bit more detail:*
After Qtile starts up, I assign a specific group to each screen through a function in my config. This works fine. After turning screens off and on, however, I end up with the standard groups being selected - i.e. group 1 for screen 1, group 2 for screen 2, etc.
I have set up a fake screen, so I am using `qtile.config.fake_screens` for my setup, not `qtile.config.screens`. I have not tested the other variable so far.

On top of that, the layout of my systray icons is broken after I turn on the screens again. I have observed this before when testing #1899 after restarting Qtile. @m-col : Do you think the issues could be related to that PR? (I have nothing tied to the `screen_change`-hook.)

# Qtile version

Most recent master.

# Stack traces

There actually is a bit of a stacktrace that might be related to this:

```log
2021-03-21 08:16:29,679 ERROR libqtile core.py:_xpoll():L313 Got an exception in poll loop
  File "/home/tcl/Software/qtile/libqtile/backend/x11/core.py", line 281, in _xpoll
    ret = target(event)
  File "/home/tcl/Software/qtile/libqtile/widget/systray.py", line 88, in handle_DestroyNotify
    del(self.systray.icons[wid])
KeyError: 23068737
2021-03-21 08:16:30,406 ERROR libqtile core.py:_xpoll():L313 Got an exception in poll loop
  File "/home/tcl/Software/qtile/libqtile/backend/x11/core.py", line 281, in _xpoll
    ret = target(event)
  File "/home/tcl/Software/qtile/libqtile/widget/systray.py", line 88, in handle_DestroyNotify
    del(self.systray.icons[wid])
KeyError: 23068745
```

# Configuration

```python
@hook.subscribe.screen_change
def restart_on_add_screen(ev):
    # reload_screen()
    # libqtile.qtile.cmd_restart()
    pass

fake_screens = init_fakescreens()
```

### LINKED ISSUE #2363: CPUGraph is covered by other widgets
Brief description:
The CPUGraph widget borders aren't respected by other widgets. They overlay on top of the graph. See the attached image.

abc

# Qtile version

Commit at which I noticed the issue: 0566d1111ec7a602e7e45792b48317e715b01c70
Commit at which I did not have the issue: 79d480b85a935986514b9aa8a2e0bdd9e8ca790f
I didn't bisect the issue, but could if that would be helpful.

# Stack traces
This didn't create a stack trace

# Configuration
    widget.CPUGraph(**widget_defaults),

elParaguayo:
Systray widget icons are displayed in incorrect location if widget is reconfigured (e.g. via `cmd_reconfigure_screens`). This
PR addresses that by only allowing the X selection configuration to occur once.

Fixes #2363
Fixes #2313

I don't have a multi-monitor setup so I can't check if any other issues arise on a `reconfigure_screens` call. Will need someone to test.

This sits on top of #2373.

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -118,6 +118,10 @@ class Systray(window._Window, base._Widget):
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
+
+        if self.configured:
+            return
+
         win = qtile.conn.create_window(-1, -1, 1, 1)
         window._Window.__init__(self, xcbq.Window(qtile.conn, win.wid), qtile)
         qtile.windows_map[win.wid] = self
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -250,6 +250,7 @@ class Bar(Gap, configurable.Configurable):
         configured = True
         try:
             widget._configure(self.qtile, self)
+            widget.configured = True
         except Exception as e:
             logger.error(
                 "{} widget crashed during _configure with "
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -201,7 +201,6 @@ class _Widget(CommandObject, configurable.Configurable):
             self.bar.height
         )
         if not self.configured:
-            self.configured = True
             self.qtile.call_soon(self.timer_setup)
             self.qtile.call_soon(asyncio.create_task, self._config_async())
 


libqtile/widget/systray.py
```diff
@@ -110,6 +110,7 @@ def __init__(self, **config):
         self.add_defaults(Systray.defaults)
         self.icons = {}
         self.screen = 0
+        self.tray_configured = False
```
m-col, line 113: I think you can use `self.configured` already, as this is part of `_Widget` and is set to `True` the very first time `_Widget._configure` is called.

elParaguayo: I thought about that but wouldn't `self.configured` be set after this line `base._Widget._configure(self, qtile, bar)`? In that case, all that X window stuff wouldn't even run the first time. That's why I used a different variable.

elParaguayo: Alternatively, we could move `base._Widget._configure(self, qtile, bar)` to the end of `_configure`. I don't think that would cause too many issues.

m-col: Oh yeah! I think instead we should set it to `True` within `bar.py` then, when the bar actually does the configuring. This way, all levels of `_configure` across all classes on widgets will have configured correctly set at all times.

elParaguayo: Sorry - not sure I understand what you're proposing there. Can you explain again?

m-col: We can stick `widget.configured = True` after this line: https://github.com/qtile/qtile/blob/master/libqtile/bar.py#L253

elParaguayo: Ah, I see. And remove this line: https://github.com/qtile/qtile/blob/master/libqtile/widget/base.py#L204 ?

I can put that in a separate PR as it would prevent the need for the hack here.

m-col: Yeah that should work I think. Maybe stick it in a commit before this commit here, might be easier to then check it's working fine in the systray widget (updated to use `self.configured` obvs)

elParaguayo: I've put that commit in separately (and pulled it out into it's own PR)

libqtile/widget/systray.py
```diff
@@ -149,6 +153,8 @@ def _configure(self, qtile, bar):
         )
         qtile.root.send_event(event, mask=EventMask.StructureNotify)
 
+        self.tray_configured = True
```
m-col, line 156: You left his one in here

elParaguayo: Indeed...

rxcomm:
This works. The systray does not cover up the other widgets in the bar when _e.g._ I type text in lazy.systemcmd(). However, there is an odd behavior when I plug in another (hdmi) display, _even before running xrandr._

This is the original bar layout before plugging in the hdmi screen:
xample2

After I plug in the hdmi screen, the bar layout changes to this:
xample

Note the change in size for the battery icon. Not sure why this is, but I didn't see this behavior before commit 9e6af506cd714fc993c5b7f534c689b80cc21793. I can certainly live with this behavior, but it seem strange.

elParaguayo:
Yeah. I spotted that too when working on this. It's fixed in #2371 which has already been merged.

ramnes:
Doesn't seem to work here. I have two monitors of different resolution (big screen, smaller laptop screen), and when I plug the big screen, the systray ends up in the middle of the bar, apparently with the same offset than on the laptop screen.

elParaguayo:
Thanks for testing. Is that SysTray moving from your small screen to your big screen? I don't think this PR should impact the offsets so I can't immediately understand why this would happen. What happens for you when running without this PR?

Unfortunately, without a second screen this it going to be hard for me to test further so someone else may need to take this on.

ramnes:
Yes, the bar is on the main screen, which is the big one when jt's plugged. The same happens without the PR. I guess I misunderstood what it's supposed to fix. :)

elParaguayo:
Well this is supposed to stop the tray icons being drawn over other widgets but if they're being drawn in the wrong place then it would be nice to fix that at the same time.

m-col:
We got confirmation on the issue, so this is ready to merge, right?

tych0:
> We got confirmation on the issue, so this is ready to merge, right?

Yes, although I wonder if we can't develop a more general solution to this problem somehow. IIUC, this patch only fixes the systray widget from stepping on itself, but if any other widget does similar things, they will also need an `if self.configured` check.

I'll merge this for now since it fixes a bug for somebody, but I would appreciate a more general solution if anyone has one :)

PR #2373, opened 2021-04-10
Change widget configured attribute timing

elParaguayo:
Currently, widgets have their `configured` flag set by the `base._Widget._configure` method. The result of this is that widgets
are shown as configured before any additional `_configure` code is run by the widgets. Moving this to `bar` means widgets are only configured once the whole `_configure` method has completed.

This removed the need for an additional flag (see #2372) as discussed here: https://github.com/qtile/qtile/pull/2372#discussion_r611069633


elParaguayo:
Don't think we need this as #2372 has been merged.

PR #2376, opened 2021-04-10
Add wayland backend

m-col:
This adds a standalone wayland core to the backends. Though the core is
not integrated into the rest of libqtile yet, it can be run directly
(either from TTY or within a running x11/wayland session) and will
provide a base wayland compositor:

    >> from libqtile.backend.wayland.core import Core
    >> Core()

This backend uses pywlroots by @flacjacket, which depends on pywayland,
python-xkbcommon, and wlroots.

The code is heavily based on pywlroots' tiny example, which is in turn
based on the tinyWL example that is part of wlroots.

Some todos (if anybody wants to help out these are good places to start 😉):

- [x]  Fix mypy dependencies so type checking works
- [x]  Move X-related code from `Qtile.process_button_click` to backend
- [x]  Move backend-agnostic Window methods to a base Window class for backends to build on. Or instead do we keep Window's `window` attribute and have two levels of windows?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,35 @@ jobs:
               uses: actions/setup-python@v2
               with:
                   python-version: ${{ matrix.python-version }}
-            - run: |
+            - name: Install dependencies
+              run: |
                 sudo apt update
-                sudo apt install libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors
+                sudo apt install libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 \
+                  graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors libwlroots-dev git ninja-build
                 pip install tox tox-gh-actions
-              name: install dependencies
+                # we want wlroots' dependencies only, so remove the packages (these are outdated in ubuntu's repos)
+                sudo dpkg -r --force-depends wlroots libwayland-dev
+            - name: Build wayland
+              env:
+                WAYLAND: 1.19.0
+              run: |
+                wget --no-check-certificate https://wayland.freedesktop.org/releases/wayland-$WAYLAND.tar.xz
+                tar -xJf wayland-$WAYLAND.tar.xz
+                cd wayland-$WAYLAND
+                ./configure --disable-documentation
+                make
+                sudo make install
+            - name: Build wlroots
+              env:
+                WLROOTS: 0.13.0
+              run: |
+                wget --no-check-certificate https://github.com/swaywm/wlroots/archive/$WLROOTS.tar.gz
+                tar -xzf $WLROOTS.tar.gz
+                sudo pip install meson
+                cd wlroots-$WLROOTS
+                meson build
+                ninja -C build
+                sudo ninja -C build install
             - name: run tests
               run: |
                 [ "$(grep -c -P '\t' CHANGELOG)" = "0" ]
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -116,6 +116,20 @@ class Window(metaclass=ABCMeta):
         """Is it OK for this window to steal focus?"""
         return True
 
+    @property
+    def floating(self) -> bool:
+        """Whether this window should be floating."""
+        return False
+
+    @property
+    def wants_to_fullscreen(self):
+        """Does this window want to be fullscreen?"""
+        return False
+
+    def match(self, match: config.Match) -> bool:
+        """Match window against given attributes."""
+        return False
+
 
 class Internal(Window, metaclass=ABCMeta):
     pass
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -5,6 +5,8 @@ import enum
 import typing
 from abc import ABCMeta, abstractmethod
 
+from libqtile.command.base import CommandObject, ItemT
+
 if typing.TYPE_CHECKING:
     from typing import Dict, List, Tuple, Union
 
@@ -102,7 +104,7 @@ class FloatStates(enum.Enum):
     MINIMIZED = 6
 
 
-class Window(metaclass=ABCMeta):
+class Window(CommandObject, metaclass=ABCMeta):
     def __init__(self):
         self.borderwidth: int = 0
         self.name: str = "<no name>"
@@ -162,6 +164,16 @@ class Window(metaclass=ABCMeta):
     def place(self, x, y, width, height, borderwidth, bordercolor,
               above=False, margin=None):
         """Place the window in the given position."""
+
+    def _items(self, name: str) -> ItemT:
+        return None
+
+    def _select(self, name, sel):
+        return None
+
+    @abstractmethod
+    def cmd_focus(self, warp: typing.Optional[bool] = None) -> None:
+        """Focuses the window."""
 
 
 class Internal(Window, metaclass=ABCMeta):
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -91,6 +91,14 @@ class Window(metaclass=ABCMeta):
     def wid(self) -> int:
         """The unique window ID"""
 
+    @abstractmethod
+    def hide(self) -> None:
+        """Hide the window"""
+
+    @abstractmethod
+    def unhide(self) -> None:
+        """Unhide the window"""
+
 
 class Internal(metaclass=ABCMeta):
     pass
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -23,6 +23,10 @@ class Core(metaclass=ABCMeta):
         """Setup a listener for the given qtile instance"""
 
     @abstractmethod
+    def update_desktops(self, groups, index: int) -> None:
+        """Set the current desktops of the window manager"""
+
+    @abstractmethod
     def grab_key(self, key: typing.Union[config.Key, config.KeyChord]) -> typing.Tuple[int, int]:
         """Configure the backend to grab the key event"""
 
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -194,6 +194,9 @@ class Window(_Window, metaclass=ABCMeta):
         """What window is this window a transient windor for?"""
         return None
 
+    def paint_borders(self, color, width) -> None:
+        """Paint the window borders with the given color and width"""
+
     @abstractmethod
     def cmd_focus(self, warp: Optional[bool] = None) -> None:
         """Focuses the window."""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -5,7 +5,7 @@ import typing
 from abc import ABCMeta, abstractmethod
 
 if typing.TYPE_CHECKING:
-    from typing import List, Tuple, Union
+    from typing import Dict, List, Tuple, Union
 
     from libqtile import config
     from libqtile.core.manager import Qtile
@@ -70,6 +70,9 @@ class Core(metaclass=ABCMeta):
 
     def warp_pointer(self, x: int, y: int) -> None:
         """Warp the pointer to the given coordinates relative."""
+
+    def update_client_list(self, windows_map: Dict[int, WindowType]) -> None:
+        """Update the list of windows being managed"""
 
 
 @enum.unique
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -136,6 +136,19 @@ class Window(metaclass=ABCMeta):
         """Match window against given attributes."""
         return False
 
+    @abstractmethod
+    def focus(self, warp: bool):
+        """Focus this window and optional warp the pointer to it."""
+
+    @property
+    def has_focus(self):
+        return self == self.qtile.current_window
+
+    @abstractmethod
+    def place(self, x, y, width, height, borderwidth, bordercolor,
+              above=False, margin=None):
+        """Place the window in the given position."""
+
 
 class Internal(Window, metaclass=ABCMeta):
     pass
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -67,6 +67,10 @@ class Core(metaclass=ABCMeta):
     def ungrab_pointer(self) -> None:
         """Release grabbed pointer events"""
 
+    @abstractmethod
+    def focus_by_click(self, event: typing.Any) -> None:
+        """Focus a window by clicking on it."""
+
     def scan(self) -> None:
         """Scan for clients if required."""
 
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -117,6 +117,10 @@ class Window(metaclass=ABCMeta):
     def unhide(self) -> None:
         """Unhide the window"""
 
+    @abstractmethod
+    def kill(self) -> None:
+        """Kill the window"""
+
     @property
     def can_steal_focus(self):
         """Is it OK for this window to steal focus?"""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -27,6 +27,10 @@ class Core(metaclass=ABCMeta):
         """Set the current desktops of the window manager"""
 
     @abstractmethod
+    def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
+        """Get the screen information"""
+
+    @abstractmethod
     def grab_key(self, key: typing.Union[config.Key, config.KeyChord]) -> typing.Tuple[int, int]:
         """Configure the backend to grab the key event"""
 
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -80,7 +80,10 @@ class FloatStates(enum.Enum):
 
 
 class Window(metaclass=ABCMeta):
-    pass
+    @property
+    @abstractmethod
+    def wid(self) -> int:
+        """The unique window ID"""
 
 
 class Internal(metaclass=ABCMeta):
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -85,6 +85,9 @@ class Core(metaclass=ABCMeta):
         """A context manager to suppress window events while operating on many windows."""
         yield
 
+    def flush(self) -> None:
+        """If needed, flush the backend's event queue."""
+
     def graceful_shutdown(self):
         """Try to close windows gracefully before exiting"""
 
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import enum
 import typing
 from abc import ABCMeta, abstractmethod
@@ -74,6 +75,11 @@ class Core(metaclass=ABCMeta):
 
     def update_client_list(self, windows_map: Dict[int, WindowType]) -> None:
         """Update the list of windows being managed"""
+
+    @contextlib.contextmanager
+    def masked(self):
+        """A context manager to suppress window events while operating on many windows."""
+        yield
 
 
 @enum.unique
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -63,3 +63,6 @@ class Core(metaclass=ABCMeta):
     @abstractmethod
     def ungrab_pointer(self) -> None:
         """Release grabbed pointer events"""
+
+    def scan(self) -> None:
+        """Scan for clients if required."""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -81,6 +81,9 @@ class Core(metaclass=ABCMeta):
         """A context manager to suppress window events while operating on many windows."""
         yield
 
+    def graceful_shutdown(self):
+        """Try to close windows gracefully before exiting"""
+
 
 @enum.unique
 class FloatStates(enum.Enum):
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -8,7 +8,7 @@ from abc import ABCMeta, abstractmethod
 from libqtile.command.base import CommandObject, ItemT
 
 if typing.TYPE_CHECKING:
-    from typing import Dict, List, Tuple, Union
+    from typing import Dict, List, Optional, Tuple, Union
 
     from libqtile import config
     from libqtile.core.manager import Qtile
@@ -133,6 +133,9 @@ class Window(CommandObject, metaclass=ABCMeta):
     def kill(self) -> None:
         """Kill the window"""
 
+    def get_wm_class(self) -> Optional[str]:
+        """Return the class of the window"""
+
     @property
     def can_steal_focus(self):
         """Is it OK for this window to steal focus?"""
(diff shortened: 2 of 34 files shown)


m-col:
I'm posting this as a draft to open discussion while I continue to work on it in parallel.

m-col:
Correction, `pywlroots` itself doesn't depend on `python-xkbcommon` (just the C lib) but realistically it is required for any kind of keymap/layout configuration (generally set by environmental variables e.g. I have `XKB_DEFAULT_OPTIONS` and `XKB_DEFAULT_LAYOUT` set which is read by various wayland compositors that also use xkbcommon).

m-col:
One thing that has popped up that I'm not sure how to deal with is incorporating the dependencies into the mypy tox run. I added them to `tox.ini` with this diff:
```diff
diff --git a/tox.ini b/tox.ini
index 1058d922..a25abbbc 100644
--- a/tox.ini
+++ b/tox.ini
@@ -81,6 +81,9 @@ deps =
     bowler
     xcffib >= 0.8.1
     pytest >= 6.2.1
+    xkbcommon
+    pywayland
+    pywlroots
 commands =
     pip install -r requirements.txt
     mypy -p libqtile
```
It looks like the `ffi_build.py` script of pywlroots (possibly the others?) need to be run as this error is being raised when mypy is starting up:
```
...
  File "mypy/lookup.py", line 47, in lookup_fully_qualified
AssertionError: Cannot find component '_ffi' for 'wlroots._ffi.lib'
ERROR: InvocationError for command /home/mcol/git/qtile/.tox/mypy/bin/mypy -p libqtile (exited with code 1)
```
Perhaps @flacjacket you might be have an idea as to what might need doing? Maybe pywlroots' setup.py needs to run ffi_build.py when building for install?

tych0:
Nice, this is a lot less code than I imagined :)

libqtile/backend/wayland/keyboard.py
```diff
@@ -0,0 +1,100 @@
...
+        logger.debug("Signal: keyboard modifier")
+        self.seat.keyboard_notify_modifiers(self.keyboard.modifiers)
+
+    def _on_key(self, _listener, event: KeyboardKeyEvent):
+        logger.debug("Signal: keyboard key")
+        handled = False
+
+        if event.state == KeyState.KEY_PRESSED:
+            # translate libinput keycode -> xkbcommon
+            keycode = event.keycode + 8
```
tych0, line 91: Can you add a link or something to describe the magic number 8 here?

libqtile/core/manager.py
```diff
@@ -153,10 +154,6 @@ def _configure(self):
         self._process_screens()
         self.current_screen = self.screens[0]
 
-        self.conn.flush()
-        self.conn.xsync()
-        self.core._xpoll()
```
tych0, line 158: I wonder if this means that the manager shouldn't have the connection any more, since it's really an x11 thing?

libqtile/backend/wayland/keyboard.py
m-col: I'd love to but I can't find anything that expands on it. It's lifted from here: https://github.com/swaywm/wlroots/blob/master/tinywl/tinywl.c#L191-L192. I searched libxkbcommon's github and the Xorg gitlab for libinput and libinput's gitlab for xkbcommon, and none point to anything that elaborates on this. I guess it was just a decision to make it easy for developors, but they offset by a byte for some reason.

libqtile/core/manager.py
m-col: Yes for sure. There are _lots_ of things that need to be changed both in the manager and diffused throughout the code.

My approach for resolving all of these is following chronologically what happens when the WM starts up, and so the commits loosely follow that logic. Currently at setting up mouse button grabs and listeners (`Qtile._configure`). I think many things will simply require moving some of the logic from `manager.py` into the X backend and adding an equivalent method to the cores, but some things like `window.Window`'s hard-coded x window references are going to be a pain.

libqtile/backend/wayland/keyboard.py
tych0: Ha, fair enough :)

m-col:
Does anyone foresee any issues if I were to move the `self.core.scan` call so that it is hooked into `startup_complete`? I'm thinking that it doesn't matter if it happens right at the end because every time a window is managed the layouts will be reconfigured so everything will organised as expected.

tych0:
My only complaint about using hooks is that they're hard to follow and debug.

The concept of scanning in this way doesn't really exist in wayland, right? So the function is just a no-op?

m-col:
It doesn't, as clients can't exist without the server (and in wayland Qtile _is_ the server). I suppose adding a no-op method would do nicely!

m-col:
Update: the backend is now setup to be started the "proper" way (with `qtile start`) so this will no longer work:
```
>> from libqtile.backend.wayland.core import Core
>> Core()
```
However, this will work and create a running session:
```sh
qtile start -b wayland
```
Lots of things should work, such as listening and responding to key bindings etc. If a client creates a window it will be rendered (though not positioned in any useful way). However the next thing that needs doing is refactoring `window.Window` and pals so that windows are managed in a more consistent and useful way. Until this is done, creating a bar will error so for testing I've been running the default config with the `bar` bit removed.

tych0:
Nice, that's awesome :)

m-col:
Another update: I think the most difficult tasks have been completed, namely updating any code around libqtile that should be in the x11 backend but ain't. It is now. Most of the work ahead on the wayland backend is limited to actually implementing it, and fortunately wlroots makes that really easy. I've tried to keep the git history understandable so although there are a lot of commits in the PR, hopefully it should be pretty straight forward to read through.

Given the complexity of the proposed changes I'm obviously eager to merge it before we get too many new commits to master, especially ones that require me to rebase and go through the new commits to update them. For this reason I've left many things not implemented, things that can be added in future standalone PRs. I wanted to get to the stage where Qtile will start up, render things, and correctly manage and place windows. This means that the base Window classes and their backend-agnostic handling is implemented around the rest of libqtile.

The main thing holding back merging (beside review and discussion, of course) is fixing up the type checking as pulling in the extra dependencies has made that somewhat more complicated so I'm going to dedicate a bit of time to that next.

In the meantime, if anybody has some time for reviewing the commits then that would be much appreciated :)

libqtile/backend/wayland/core.py
```diff
@@ -0,0 +1,346 @@
...
+            self.desktops += 1
+        while new_count < self.desktops:
+            self.desktops -= 1
+        self.current_desktop = index
+
+    def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
+        """Get the screen information"""
+        xywh = {}  # type: Dict[Tuple[int, int], Tuple[int, int]]
+        for screen in self.outputs:
+            sx, sy, sw, sh = screen.get_geometry()
```
flacjacket, line 302: It may make sense for the screen info to just return this screen geometry information, and the logic around combining screens can be kept in a single shared location between backends.

libqtile/backend/base.py
```diff
@@ -41,3 +65,95 @@ def grab_pointer(self) -> None:
...
         """Release grabbed pointer events"""
+
+    def scan(self) -> None:
+        """Scan for clients if required."""
+
+    def warp_pointer(self, x: int, y: int) -> None:
+        """Warp the pointer to the given coordinates relative."""
+
+    def update_client_list(self, windows_map: Dict[int, WindowType]) -> None:
+        """Update the list of windows being managed"""
```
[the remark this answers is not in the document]
m-col: Yeah I set these as no-ops as their behaviour/return values are not required for Qtile to continue, so in theory a backend could not implement them with no consequences.

libqtile/backend/wayland/core.py
```diff
@@ -0,0 +1,346 @@
...
+        os.environ["WAYLAND_DISPLAY"] = self.socket.decode()
+        logger.info("Starting core with WAYLAND_DISPLAY=" + self.socket.decode())
+        self.backend.start()
+
+    def finalize(self):
+        self._on_new_xdg_surface_listener.remove()
+        self._on_request_cursor_listener.remove()
+        self._on_new_output_listener.remove()
+        self._on_new_input_listener.remove()
+        self._on_request_set_selection_listener.remove()
```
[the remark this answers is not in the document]
m-col: Looks like the ones missing are ones I added second and forgot to remove here, so a helper to avoid that is probably a good idea. To be honest removing listeners when finalizing the core isn't as important as removing listeners when closing a window or removing an input or output. Either way, I might be able to put something together to make this more fool proof.

libqtile/backend/base.py
```diff
@@ -41,3 +65,95 @@ def grab_pointer(self) -> None:
...
+
+    @contextlib.contextmanager
+    def masked(self):
+        """A context manager to suppress window events while operating on many windows."""
+        yield
+
+
+class FloatStates:
```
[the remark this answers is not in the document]
m-col: What's the advantage to using that?

libqtile/backend/wayland/core.py
m-col: Yeah good call, updated.

libqtile/backend/wayland/keyboard.py
```diff
@@ -0,0 +1,100 @@
...
+        self.keyboard.modifiers_event.add(self._on_modifier_listener)
+        self.keyboard.key_event.add(self._on_key_listener)
+        self.keyboard.destroy_event.add(self._on_destroy_listener)
+
+    def finalize(self):
+        self._on_modifier_listener.remove()
+        self._on_key_listener.remove()
+        self._on_destroy_listener.remove()
+        self.core.keyboards.remove(self)
+        if self.core.keyboards:
```
flacjacket, line 74: Would we first want to check if it is the current keyboard that is the active keyboard before setting `keyboards[-1]`?

libqtile/backend/wayland/core.py
```diff
@@ -0,0 +1,339 @@
...
+            previous_xdg_surface = xdg_shell.XdgSurface.from_surface(previous_surface)
+            previous_xdg_surface.set_activated(False)
+
+        # roll the given surface to the front of the list, copy and modify the
+        # list, then save back to prevent any race conditions on list
+        # modification
+        windows = self.windows[:]
+        windows.remove(win)
+        windows.append(win)
+        self.win = windows
```
flacjacket, line 260: I think a lot of this logic in the tiny compositor is just to facilitate being able to do simple focus prev/focus next like operations. I would think that would be handled somewhere else once we are in qtile. Does that sound right to you?

libqtile/backend/x11/core.py
```diff
@@ -169,27 +170,19 @@ def finalize(self) -> None:
...
+        if not info:
+            info.append((
+                0, 0,
                 self.conn.default_screen.width_in_pixels,
                 self.conn.default_screen.height_in_pixels,
-            )
+            ))
+
+        if self.qtile:
+            self._xpoll()
```
flacjacket, line 183: Might be worth changing the name of `_xpoll`?

flacjacket (commented):
Taking a quick pass through, I don't see anything obvious that would need to be changed. I think it makes sense to try to get this functionality in and iterate on it there, rather than causing a maintenance headache for this branch. The only thing might be if we want to get out a more stable release before going ahead with this change, since it is pretty big.

libqtile/backend/x11/core.py
m-col: Can do. What were you thinking instead?

libqtile/backend/wayland/core.py
m-col: Yep I think so too. The Core is maintaining a list of windows that is used only to know what order to render them and currently has no awareness of groups or anything like that. The way it manages windows will have to change (whether this is per group or not is too early to tell). I have a working setup for "virtual desktops" on another project (https://github.com/m-col/wimp) which I'll try to mirror as the PR as outgrown tinywm.

libqtile/backend/wayland/keyboard.py
m-col: Ah yep, that makes sense.

m-col: I've added the check but using `_ptr` directly for now.

m-col:
> The only thing might be if we want to get out a more stable release before going ahead with this change, since it is pretty big.

Is it a problem if a non-functioning backend is packaged alongside releases? `qtile start --help`, when listing backends, already has a note for wayland saying it is for development only. We could additionally add a note to the issue template saying we are not interested in bugs when you are using this backend. And of course, there is no mention of the backend in the docs.

flacjacket:
I wouldn't worry about the new backend, more about if the other changes to
the x11 backend and the core cause some other issues. If we are diffident
in those changes then that may not be an issue. We (read: I) have
definitely been bitten by making too many big changes right before a
release and having to do something quick to fix it.

tych0:
I was thinking about this the other day, and I think it's fine to list the backend in `qtile start --help` with the note that's there. You could also add a note that the configuration for the backend is subject to change. I don't see any need to stabilize it at this point. However,

> We could additionally add a note to the issue template saying we are not interested in bugs when you are using this backend

don't you want people to file bugs if something is busted? It would be more like "make sure you've updated your config and are running git master, we are not interested in wayland bugs of released versions of qtile since velocity is high at this time".

libqtile/backend/base.py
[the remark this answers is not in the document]
flacjacket: It can help to convey intent and semantics a bit. If you have a type that accepts or returns an `int`, you need to figure out what that represents, but an annotation of `FloatStates` is more descriptive. Similarly, if you are trying to debug something, printing out the value will be more descriptive as the enum rather than as the int. You also can't do things like accidentally do some arithmetic with the values of an enum like you can with an int.

m-col:
> I wouldn't worry about the new backend, more about if the other changes to
> the x11 backend and the core cause some other issues.

Good point. We have had some nice bug fixes recently so maybe we should get out a minor release or something before merging this.

> don't you want people to file bugs if something is busted?

I was thinking that we'll get issues saying things are missing just because we haven't yet implemented that thing, however if I make an issue to organise a to-do list for wayland then we can close any issues that point out a bug/omission that is listed there and we're aware about.

m-col:
The CI is giving me `I001 isort found an import in the wrong position` isort errors but I'm not getting the same errors locally, even when run with tox and tox prints out the exact same isort version number. Pretty weird. Anyone seen something like that?

m-col:
Specifically these ones, can anyone see what's wrong with these imports?
```
  libqtile/backend/wayland/keyboard.py:25:1: I001 isort found an import in the wrong position
  libqtile/backend/wayland/core.py:27:1: I001 isort found an import in the wrong position
  libqtile/backend/wayland/core.py:28:1: I001 isort found an import in the wrong position
  libqtile/backend/wayland/core.py:29:1: I001 isort found an import in the wrong position
  libqtile/backend/wayland/window.py:25:1: I001 isort found an import in the wrong position
  libqtile/backend/wayland/window.py:26:1: I004 isort found an unexpected blank line in imports
  libqtile/backend/wayland/output.py:25:1: I001 isort found an import in the wrong position
```

m-col:
So I have no clue what's up with those isort messages... I match the versions of every single python package used during the CI on my local machine and still I get no messages. I tested to check it was otherwise working and the CI isn't using some cached version of qtile or something by swapping some imports in lazy.py.... these were detected as expected!

ramnes:
What you're doing in ci.yml is really hacky (especially cloning the repositories and installing them by hand), and with isort being ran by tox in its virtualenv, it' s possible that you end up confusing isort, making it think that some of the libraries are first party, or something like this. You probably want to set the dependencies properly in setup.cfg to fix this. And it' s how we'll have to package things anyway.

.github/workflows/ci.yml
```diff
@@ -21,8 +21,46 @@ jobs:
...
+                meson build
+                ninja -C build
+                sudo ninja -C build install
+                cd ..
+                # and we want git versions of pywayland
+                git clone https://github.com/flacjacket/pywayland
+                cd pywayland
+                pip install -r requirements.txt
+                python pywayland/ffi_build.py
+                python setup.py build
```
flacjacket, line 53: I _think_ these should be run automatically when `setup.py install` is run, is that not the case?

flacjacket: After these are installed, maybe you need to `rm -r` the cloned git directories? I'm not sure if having these packages still available locally is somehow throwing off isort?

flacjacket:
At least the pywlroots and pywayland packages have fresh releases on the tip of master, so those should be usable from the current releases, and I am happy to cut patch releases as needed for this.

.github/workflows/ci.yml
m-col: Ah yeah that would make sense. Hopefully I can avoid this commit altogether though :)

m-col:
> What you're doing in ci.yml is really hacky

Yes :')

> You probably want to set the dependencies properly in setup.cfg to fix this. And it's how we'll have to package things anyway.

While this would avoid this whole issue, it would require a new release from pywlroots or pywayland upon any commits that are added simply to meet needs of Qtile. Then again...

> At least the pywlroots and pywayland packages have fresh releases on the tip of master, so those should be usable from the current releases, and I am happy to cut patch releases as needed for this.

If that is alright, then I can go ahead and use the pip versions and not do this hoop jumping.

As for the C libs, are we happy to download and build these in the CI? They can of course stick to releases rather than git versions.

ramnes:
Yeah I would just go with the releases and kindly ask developers to release if needed. Worst case scenario, pip can install zips, and GitHub has zip archives by commit. :)

m-col:
OK, so, I've updated tox to use release versions. I added xkbcommon and pywayland to the 'deps' in `tox.ini` but `pip install pywlroots` as the first line in the 'commands' section, because it needs to be installed after pywayland. Similar thing to the xcffib-cairocffi thing, though I can't tell from the file how that is handled. In any case, it's running fine and I'm just left with some mypy errors (extract of many similar lines):
```
 libqtile/backend/wayland/wlrq.py:27: error: Cannot find implementation or library stub for module named 'wlroots.wlr_types.keyboard'
libqtile/backend/wayland/window.py:25: error: Cannot find implementation or library stub for module named 'pywayland.server'
...
```
Given these deps are installed and type annotated, I'm not sure why mypy is having trouble following the import. Any ideas?

m-col:
Sorry to get carried away and pile a few more commits on, but now it's actually at a really good point. Here is a preview of what's implemented. Qtile is launched from a running Qtile X session so it opens up in a window much like Xephyr does:

https://user-images.githubusercontent.com/44536113/115909497-82e9a580-a463-11eb-9a20-085c707b9cd3.mp4

You'll notice that the floating windows appear behind the others (and the opacity works great!). Some proper Z axis handling will be required to make that a bit nicer.

setup.cfg
```diff
@@ -127,3 +127,9 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-libqtile.widget._pulse_audio]
 ignore_missing_imports = True
+[mypy-xkbcommon.*]
+ignore_missing_imports = True
+[mypy-pywayland.*]
+ignore_missing_imports = True
+[mypy-pywlroots.*]
```
flacjacket, line 134: This should just be `wlroots`

stubs/wlroots/__init__.pyi
```diff
@@ -0,0 +1 @@
+from ._ffi import ffi as ffi, lib as lib
```
flacjacket, line 1: I wonder if having this stub file here is throwing things off? I think we can get rid of this, if it is needed, it should be able to pull it from the actual package.

m-col:
I've finished up all of it I think, and the CI is mostly working. The error of the CI is this, which I also see (locally at least) on master:
```
  libqtile/widget/pulse_volume.py:4: error: Module 'libqtile.widget._pulse_audio' has no attribute 'ffi'
  libqtile/widget/pulse_volume.py:4: error: Module 'libqtile.widget._pulse_audio' has no attribute 'lib'
```

m-col:
But I don't get it on master on github actions. Guess I'll have to bisect using github actions, how fun!

m-col:
The problem is (fortunately) caused by that last commit that plays with the CI. Here it is by itself on top of master with the rest of the PR: https://github.com/m-col/qtile/runs/2423161831

Could anyone take a look to see what's causing this error? I guess it's something with the way qtile is being installed.

tox.ini
```diff
@@ -82,10 +86,11 @@ deps =
     xcffib >= 0.8.1
     pytest >= 6.2.1
 commands =
-    pip install -r requirements.txt
+    pip install -r requirements.txt pywayland xkbcommon
+    pip install pywlroots
+    python3 setup.py install
     mypy -p libqtile
     # also run the tests that require mypy
-    python3 setup.py install
```
flacjacket, line 88: Is there something happening here that is odd with the order of `python3 setup.py install` and `mypy -p libqtile`?

.github/workflows/ci.yml
```diff
@@ -19,11 +19,35 @@ jobs:
...
+            - name: Install dependencies
+              run: |
                 sudo apt update
-                sudo apt install libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors
+                sudo apt install libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 \
+                  graphviz imagemagick x11-apps xserver-xephyr xterm xvfb libpulse-dev lm-sensors libwlroots-dev git ninja-build
                 pip install tox tox-gh-actions
-              name: install dependencies
+                # we want wlroots' dependencies only, so remove the packages (these are outdated in ubuntu's repos)
+                sudo dpkg -r --force-depends wlroots libwayland-dev
```
flacjacket, line 29: Not related to the mypy changes, but do you also need to remove `libwlroots-dev` and all the `libwayland-*` packages?

tox.ini
m-col: I don't think had an effect on the issue that I'm seeing. However doesn't it make more sense first, because the setup.py call runs ffi_build.py, right?

.github/workflows/ci.yml
m-col: I believe so. (lib)wayland is built manually because it is also packaged by ubuntu with an outdated version that is incompatible with wlroots 0.13

tox.ini
tych0: FWIW, I see this all the time locally and I just ignore it, so perhaps it's something with 20.04's setuptools or something too. I have no idea why it works in CI right now, in theory I should be doing exactly the same tox stuff...

m-col:
I added some basic protocols (for another PR) to support screen copying so now I can take images or videos from my TTY session rather than a nested session within X:

20210423_23h19m02s_grim

Also, gamma control!

tox.ini
flacjacket: I am thinking there is something with doing the build and install that is causing this to fire. I can reproduce this with this change: https://github.com/flacjacket/qtile/commit/1eb2c4291e29c67111b723d278a82c02394ba553 see https://github.com/flacjacket/qtile/runs/2423680304

m-col: Huh, thought I had tested it without that change too and still got it. Well I hope I was wrong. Just pushed with that bit reverted.

m-col: brilliant, that fixed it!

m-col:
I think it's at a good place for reviewing/merging if people are happy with that :) I'm just a bit aware that each merge to master potentially gives me a rebasing nightmare (yes, yes, I know it's my fault for even having a 50 commit PR...)

libqtile/backend/wayland/core.py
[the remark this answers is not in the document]
m-col: I'm going to add a reminder to the To Do list to go over the listener handling code and see if there is a tidyer way to handle all of this, rather than add more to this PR.

m-col:
Just pushed a tiny fix. Should be good to go.

ramnes:
Looks like a very solid base to me, going through each of these commits has been a real pleasure. The work you've put into this is incredible, thank you so much. :)

m-col:
Thanks! :D

PR #2377, opened 2021-04-11
Move _TextBox max_chars handling

### LINKED ISSUE #2374: The max_chars key of CurrentLayout widget has no effect.
# Issue description

I'd like to prevent my bar to adjust when the layout changes in the CurrentLayout widget.
Since ae282e38, the max_chars key is available for CurrentLayout widget.
So I tried to use 'max_chars=2' but the layout name is not truncated as expected.
I'll try to fix my bar with the fmt key for the moment, but I guess the base._TextBox.update method should be called by the hook_response method to take into account max_chars.

# Qtile version

git-version (90a07097 at the moment)

# Stack traces

Nothing relevant.

# Configuration

widget.CurrentLayout(max_chars=2),

elParaguayo:
Users can specify number of `max_chars` for `_TextBox` based widgets. Where text exceeds this length, it is truncated and an ellipsis is appended.

However, the truncation was only applied when the `update` method was called. Widgets that set the `text` attribute directly would not have the text truncated if necessary.

This PR ensures that the truncation is applied in both scenarios.

Fixes #2374.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -357,6 +357,8 @@ class _TextBox(_Widget):
 
     @text.setter
     def text(self, value):
+        if len(value) > self.max_chars > 0:
+            value = value[:self.max_chars] + "…"
         self._text = value
         if self.layout:
             self.layout.text = self.formatted_text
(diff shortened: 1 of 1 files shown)


ramnes:
Thanks!

PR #2381, opened 2021-04-13
Swap groups between screens

kovasap:
Continuing work for https://github.com/qtile/qtile/pull/1704.


kovasap:
At this point I'm still blocked on understanding the tests, and don't have bandwidth to work on it too much now.  I may pick it up later, or if someone else wants to see this merged I'm happy to take their guidance to finish up this PR.

tych0:
The git history for this is kind of a mess, and I suspect it will prevent others from picking it up directly. Any chance you can clean that up?

kovasap:
Yeah good point.  I guess that would involve removing commits from the branch via rebasing?

kovasap:
I'd like to finally merge this.  @ramnes you mentioned in the original PR that you're interested in this feature (https://github.com/qtile/qtile/pull/1704#issuecomment-622625967).  I'm still not sure how to fix up the tests,  could you help or point me to some docs?

ramnes:
What is it that you need help with more precisely? FYI I'm not really involved in Qtile these days, alas.

kovasap:
Ah, I'm just a bit lost trying to make sure my tests are sufficient and
will pass.

elParaguayo:
I'll take a look later.

kovasap:
Gentle ping on this

elParaguayo:
I'm not convinced by this one. Seems like some code that could be kept in qtile-examples rather than being core functionality.

Also, what's the behaviour when you've got more than 2 screens. Are all groups being changed in that scenario?

ramnes:
Seems to be quite a popular thing that people implement in their config: https://github.com/search?q=qtile+AND+%28%22def+switch_screen%22+OR+%22def+swap_screen%22%29+language%3APython&type=code&l=Python

elParaguayo:
Great. Then we should include it.

kovasap:
I fixed a merge conflict but am still not sure how to proceed with the tests.

test/test_manager.py
```diff
@@ -99,7 +99,19 @@ class ManagerConfig(Config):
...
+            bottom=libqtile.bar.Bar(
+                [libqtile.widget.GroupBox()], 20
+            ),
+        )
+        for _ in range(3)
+    ]
+
+
+manager_config = pytest.mark.parametrize("qtile", [ManagerConfig], indirect=True)
+multi_screen_manager_config = pytest.mark.parametrize("qtile", [MultiScreenManagerConfig], indirect=True)
```
elParaguayo, line 114:
```suggestion
manager_config = pytest.mark.parametrize("manager", [ManagerConfig], indirect=True)
multi_screen_manager_config = pytest.mark.parametrize("manager", [MultiScreenManagerConfig], indirect=True)
```
This is the cause of all the test failures. The first argument is the name of the fixture that you're parametrizing so it has to be `manager` here.

libqtile/core/manager.py
```diff
@@ -1412,6 +1412,17 @@ def prev_screen(self) -> None:
...
+    def swap_screens(self, reverse = False):
+        """Switch groups between active screens"""
+        prev_group = None
+        screens = self.screens[::-1] if reverse else self.screens
+        for screen in screens:
+            if prev_group is None:
+                prev_group = screen.group
+                continue
+            screen.set_group(prev_group)
+            self.cmd_next_screen()
```
elParaguayo, line 1424:
```suggestion
            self.next_screen()
```
The `cmd_` prefix is deprecated.

Also, this changes the actual screen that's focused. Do you need to do that?

test/test_manager.py
```diff
@@ -165,6 +177,66 @@ def test_to_screen(manager):
     manager.c.prev_screen()
     assert manager.c.window.info()["name"] == "one"
 
+@multi_screen_manager_config
+def test_swap_screens(qtile):
+    self = qtile
```
elParaguayo, line 182:
```suggestion
def test_swap_screens(manager):
```
`manager` is a defined fixture name. It needs to be used here. Please don't use the `self` alias here, that's inconsistenc across tests so those references need to be changed back to `manager`.

test/test_manager.py
```diff
@@ -165,6 +177,66 @@ def test_to_screen(manager):
...
+    assert ga["windows"] == ["two"]
+
+    gb = self.c.groups()["b"]
+    assert gb["windows"] == ["one"]
+
+    self.c.swap_screens()
+    print(self)
+    print(self.__dict__)
+    print(self.screens)
+    print(self.screens[0].__dict__)
```
elParaguayo, line 201:
```suggestion

```
Delete debugging statements.

test/test_manager.py
```diff
@@ -165,6 +177,66 @@ def test_to_screen(manager):
...
+    assert gb["windows"] == ["two"]
+    assert self.c.window.info()["name"] == "one"
+    self.c.swap_screens()
+    assert ga["windows"] == ["one"]
+    assert gb["windows"] == ["two"]
+    assert self.c.window.info()["name"] == "one"
+
+
+@multi_screen_manager_config
+def test_swap_screens(qtile):
```
elParaguayo, line 212: Is this exactly the same test?!

libqtile/core/manager.py
```diff
@@ -1412,6 +1412,17 @@ def prev_screen(self) -> None:
         self.focus_screen((self.screens.index(self.current_screen) - 1) % len(self.screens))
 
     @expose_command()
+    def swap_screens(self, reverse = False):
+        """Switch groups between active screens"""
```
elParaguayo, line 1416:
```suggestion
        """Rotate groups between active screens"""
```

elParaguayo (changes requested):
Thanks for this.

The commit history is also a bit of a mess. Once you've updated for my changes, can you rebase and squash the commit history to a single commit.

kovasap:
Thanks for the notes!  I actually attempted to do that squashing earlier at the suggestion of tych0, but failed.  Is there a particular git command you have in mind to do this?  I also know that for some github repos when a pull request is merged there is an option to do an automatic squash.

ervinpopescu:
@kovasap Probably something along the lines of:
- https://docs.qtile.org/en/latest/manual/howto/git.html#i-was-told-to-rebase-my-work
- https://docs.qtile.org/en/latest/manual/howto/git.html#i-was-told-to-squash-some-commits

kovasap:
Ok, I tried the rebase method multiple times, but it seems like this branch just has more commits in it now :cry: , and the new commits i created don't look like my net changes.  Any ideas of another way I could accomplish this?

kovasap:
I created https://github.com/qtile/qtile/pull/5195 as the successor to this with a cleaned up log.

PR #2382, opened 2021-04-13
window: make WM_CHANGE_STATE code configurable

### LINKED ISSUE #2149: Regression when starting full screen steam games
f0a67e8bd4ebda97f310de6c235f1cbd72a0dc7d introduces a regression where, when I start a steam game on one monitor (e.g. dota) and then go click on the other monitor, the window disappears and there's not a way to get it back.

I assume the window goes "somewhere" (it is still using cpu and doing i/o, etc.), but we don't have a way to get windows *out* of IconicState as far as I can tell.

/cc @m-col

tych0:
This behavior really annoys me, so I'd like to make it configurable that we
do this at all.

As an alternative, I'm open to some code that "resurrects" these windows,
but frankly I don't love that either, and would rather they just stick
around when not focused.

Closes #2149

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -59,6 +59,7 @@ class Config:
         ("bring_front_click", "bool"),
         ("reconfigure_screens", "bool"),
         ("wmname", "str"),
+        ("respect_minimize_requests", "bool"),
     ]
 
     def __init__(self, file_path=None, **settings):
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -175,6 +175,10 @@ auto_fullscreen = True
 focus_on_window_activation = "smart"
 reconfigure_screens = True
 
+# If things like steam games want to auto-minimize themselves when losing
+# focus, should we respect this or not?
+respect_minimize_requests = True
+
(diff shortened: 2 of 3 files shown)


m-col (approved):
Seems reasonable, LGTM.

ramnes:
docs + changelog pls <3

ramnes:
also, I would have called this `auto_minimize` to be consistent with `auto_fullscreen` which I think is quite a similar feature

tych0:
Well it's technically not auto_minimize, since it ignores *any* state change the window wants to perform (e.g. if the window wants to map itself, this will ignore that as well). The example in the bug report was minimization, but if some impolite window wants to start mapping itself, I'd like to ignore it too until I want to map it :). I'm happy to change the name to whatever, though. I just want to be able to disable this behavior.

I will send a changelog though.

ramnes:
Then it's technically not respect_minimize_requests neither, right? :P

tych0:
Right :). Please suggest a name, and I'll change it to that.

ramnes:
Well if we really want to be technically correct that would make it something like `auto_state` or `auto_change_state` but meh. I can't find a good one.

PR #2383, opened 2021-04-13
migrate: work with multiple files in a directory

tych0:
Much like when I wrote `qtile check`, `qtile migrate` only supported
configs that were a single file, because that's how mine is. However, I
know it is reasonably common for people to have configs that are spread
across files, so this implements support for that. The patch itself is
fairly simple, because bowler by default if you pass it a directory selects
everything with *.py in that directory to refactor, so we end up
refactoring everything in the directory the config is in.

This isn't completely accurate for a couple of reasons: people may have
non-qtile files in the same directory as their config, and people may have
configs spread out across subdirectories using submodules. I have to say I
really can't be bothered to support the second case :)

The first case is potentially problematic, as there isn't a good way to
address it, besides perhaps passing migrate an explicit list of files that
you'd like to migrate. Perhaps we could go that route, depending on what
people think.

### DIFF
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -21,6 +21,7 @@ import os
 import os.path
 import shutil
 import sys
+from glob import glob
 
 BACKUP_SUFFIX = ".migrate.bak"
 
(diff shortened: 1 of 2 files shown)


elParaguayo:
So, your first scenario is if a user has non-config files in their config folder? Wouldn't seem entirely unreasonable not to cater for that scenario either. 

I certainly agree with your view on the second scenario.

tych0:
> So, your first scenario is if a user has non-config files in their config folder? Wouldn't seem entirely unreasonable not to cater for that scenario either.

Right, exactly. Perhaps we should just ship it this way and see if anyone complains :)

Fabien-Bernard:
To ease the complains, if I understand right, we may want to change the backup system, perhaps the entire config directory instead of only the main config file.

tych0:
> To ease the complains, if I understand right, we may want to change the backup system, perhaps the entire config directory instead of only the main config file.

Yes, that's a great point, thanks.

tych0:
Ok, this should do a reasonable job of handling backing up multiple files as well too, take a look.

m-col:
> perhaps passing migrate an explicit list of files that
> you'd like to migrate. Perhaps we could go that route, depending on what
> people think.

Just thinking about this idea again, and it might work quite well. It's very flexible and it's clear which files could be modified. Implementing it would be necessary for migrating files not in the default config folder (right?), so we could maybe get files from `sys.argv` if there are any, otherwise glob the config folder.

tych0:
What does the API look like? `-c config.py -c other_import.py -c other_other_import.py`?

m-col:
I was thinking more like `qtile migrate file.py file.py file.py`. Then shell
globbing would work: `qtile migrate **/*.py`

m-col:
> What does the API look like? -c config.py -c other_import.py -c other_other_import.py?

Wait... doesn't this already work, making this PR redunant?

tych0:
> Wait... doesn't this already work, making this PR redunant?

I don't think so, we'd at least need another for loop in do_migrate() to make it work.

The thing I like about the current `-c` API is that it looks exactly the same as `qtile start`. The fact that we can't/don't parse imports that the config file does is mostly a technical limitation on our end, and surfacing it to users feels kind of ugly.

m-col:
OK yeah good points, I agree :)

Issue #2054, opened 2020-12-30
Fix coveralls


ramnes:
https://coveralls.io/github/qtile/qtile

Last update was on November 18th, so I guess something's broken.

ramnes:
We probably want to use this: https://github.com/coverallsapp/github-action

And I want to setup notifications on pull requests when coverage decreases.

ramnes:
Or maybe we should make the tests fail with something like `pytest --cov-fail-under=$(curl -s https://coveralls.io/github/qtile/qtile.json | jq -r '.covered_percent')`

m-col:
I'm not very familiar with coveralls. Am I interpreting correctly that the github action will basically do the coverage test via coveralls for PRs, whereas the pytest will simply do the test itself? And then once a commit has been added to master, coveralls updates?

ramnes:
It will actually just send the coverage results written by pytest-cov during the tests to coveralls servers but yes, you have the overall idea.

m-col:
If we use `pytest --cov-fail-under=$(curl -s https://coveralls.io/github/qtile/qtile.json | jq -r '.covered_percent')` does that basically mean that it will never decrease? That sounds pretty nice.

ramnes:
Yeah, but that can be cumbersome sometime. Ideally you want to do this only when you have 100% coverage to ensure you keep it that way, but alas I don't see us coming close to 100% without something like this.

Also we should rather have this as another test in the CI than hidden in a pytest option, if we can. The result would be the same, a big red thing on PRs that make the coverage decrease, but it would be more explicit as it would have its own line and description.

elParaguayo:
I'm new to this. Is "coverage" essentially having tests that cover every part of the codebase?

If so, how important/realistic is it to have 100% coverage?

I'm happy to spend a bit of time writing some tests but this would probably be centred around the widgets but it would be good to understand how meaningful an impact that exercise would have.

ramnes:
Yeah but here it's about the connection with the service we use to monitor our code coverage (i.e. Coveralls) that seems broken. :)

You can see which parts of the code are untested at the end of a pytest run if you want to work on improving the coverage.

elParaguayo:
Thanks will add to my list!

PR #2386, opened 2021-04-14
Increasing widget test coverage

elParaguayo:
As per #2054, additional tests are required to improve code coverage. This commit adds tests for a number of widgets.

### DIFF
--- a/libqtile/widget/crashme.py
+++ b/libqtile/widget/crashme.py
@@ -62,4 +62,3 @@ class _CrashMe(base._TextBox):
             1 / 0
         elif button == 3:
             self.text = '<span>\xC3GError'
-            self.bar.draw()
--- a/test/widgets/conftest.py
+++ b/test/widgets/conftest.py
@@ -2,6 +2,11 @@ import os
 
 import pytest
 
+import libqtile.bar
+import libqtile.config
+import libqtile.confreader
+import libqtile.layout
+
 
 @pytest.fixture(scope='function')
 def fake_bar():
@@ -51,3 +56,21 @@ def fake_qtile():
                 loop.close()
 
     return FakeQtile()
+
+
+# Fixture that defines a minimal configurations for testing widgets.
+# When used in a test, the function needs to receive a list of screens
+# (including bar and widgets) as an argument. This config can then be
+# passed to the manager to start.
+@pytest.fixture(scope='function')
+def minimal_conf_noscreen():
+    class MinimalConf(libqtile.confreader.Config):
+        auto_fullscreen = False
+        keys = []
+        mouse = []
+        groups = [libqtile.config.Group("a")]
+        layouts = [libqtile.layout.stack.Stack(num_stacks=1)]
+        floating_layout = libqtile.resources.default_config.floating_layout
+        screens = []
+
+    return MinimalConf
--- a/test/widgets/test_countdown.py
+++ b/test/widgets/test_countdown.py
@@ -1,0 +1,37 @@
+# Copyright (c) 2015 Tycho Andersen
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Widget specific tests
+
+from datetime import datetime, timedelta
+
+from libqtile import widget
+
+td = timedelta(days=10, hours=10, minutes=10, seconds=10)
+
+
+def test_countdown_formatting():
+    # Create widget but hide seconds from formatting to allow for
+    # timing differences in test environment
+    countdown = widget.Countdown(date=datetime.now() + td,
+                                 format="{D}d {H}h {M}m")
+
+    output = countdown.poll()
+    assert output == "10d 10h 10m"
--- a/test/widgets/test_countdown.py
+++ b/test/widgets/test_countdown.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2015 Tycho Andersen
+# Copyright (c) 2021 elParaguayo
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
--- a/test/widgets/test_crashme.py
+++ b/test/widgets/test_crashme.py
@@ -1,0 +1,75 @@
+# Copyright (c) 2015 Tycho Andersen
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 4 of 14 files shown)


test/widgets/test_countdown.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2015 Tycho Andersen
```
tych0, line 1: Looks like some copypasta errors here, don't give me credit for your work :)

elParaguayo: Haha. Dammit. Copied from a previous one, amended it, but obviously forgot to copy again!

No doubt you'll have other comments so I'll add this to my "to fix" list for the commit.

elParaguayo:
> ...holy shit this is awesome!

Thanks. These are the "easy" wins in the widgets. Other widgets either have third party library dependencies, connect to external servers etc. so I'm not sure how we can hit 100% on those (or tell the system to just ignore coverage on those).

test/widgets/test_quickexit.py
```diff
@@ -0,0 +1,93 @@
...
+
+import libqtile.bar
+import libqtile.config
+import libqtile.confreader
+import libqtile.layout
+from libqtile import widget
+from libqtile.ipc import IPCError
+
+
+class MinimalConf(libqtile.confreader.Config):
```
tych0, line 33: I wonder if we can just have one copy of this and import it everywhere?

elParaguayo: Yup. I do use it a lot and, unfortunately, it's a bit hacky as I add the screen/bar/widget in the test function.

There's probably a much neater way of doing this but I'm still very new to pytest.

tych0: I don't even think you need to do it with pytest magic, just put it in conftest and then import it from there. I agree there's probably a way to do it using fixtures, but IMO less smoke and mirrors is better.

elParaguayo: Sure. Will take a look at that.

elParaguayo: Yup. Quick test and that will work fine. Will update them later. Need some sleep!

ramnes:
Great work, thanks! After a quick glance, I'd just make the MinimalConfig a fixture in conftest but LGTM overall.

> Other widgets either have third party library dependencies, connect to external servers etc. so I'm not sure how we can hit 100% on those

For the libraries, feel free to add them to the test environment. For the external servers, your choices are either to mock the responses, to change the widgets so that they use dependency injection, or to setup a local test instance with docker-compose or alike if there's any open source server available. Because of our big number of tests and our wish to parallelize them, mocks are probably the easier path.

elParaguayo:
Great. Thanks for the tips.

I'll tidy up these ones for this PR and will then look at other widgets as part of "vol 2".

elParaguayo:
Made MinimalConf into a fixture. Can squash commits when ready to merge.

Will leave a couple of review comments for points I wanted to flag to you to check that you're ok with them.

test/widgets/test_crashme.py
```diff
@@ -0,0 +1,75 @@
...
+    assert w["text"] == "Crash me !"
+
+    # Testing errors. Exceptions are wrapped in CommandException
+    # so we catch that and match for the intended exception.
+
+    # Left click generates ZeroDivisionError
+    with pytest.raises(CommandException) as e_info:
+        topbar.fake_button_press(0, "top", 0, 0, button=1)
+
+    assert e_info.match("ZeroDivisionError")
```
elParaguayo, line 60: I couldn't think of a better way to do this given that the Manager wraps the exception into a CommandException. This should be robust enough but happy to listen to alternatives if there's a better way to do this.

test/widgets/test_sep.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize("screen,location,attribute", parameters)
+def test_orientations(manager_nospawn, screen, location, attribute):
+    config = MinimalConf
+    config.screens = [screen]
+
+    manager_nospawn.start(config)
+    bar = manager_nospawn.c.bar[location]
+
+    w = bar.info()["widgets"][0]
+    assert w[attribute] == 3
```
elParaguayo, line 47: I got a bit confused by the sep code. `length` seems to be set twice with different values:
https://github.com/qtile/qtile/blob/48f7c5e499fed64913732896a5fdc65fa921d2e8/libqtile/widget/sep.py#L52-L55

I would have thought the length of the widget should be `2x padding + line width`, but it's actually just `padding + line width`.

test/widgets/test_windowtabs.py
```diff
@@ -0,0 +1,137 @@
...
+def test_multiple_windows(manager):
+
+    def widget_text():
+        return manager.c.bar["top"].info()["widgets"][0]["text"]
+
+    window_one = manager.test_window("one")
+    assert widget_text() == "<b>one</b>"
+
+    window_two = manager.test_window("two")
+    assert widget_text() in ["<b>two</b> | one", "one | <b>two</b>"]
```
elParaguayo, line 114: I did it this way as the order seemed to change randomly so test would pass some times but not others. Is it ok to do this way or should the order be fixed?

elParaguayo:
Just seen some of these widgets still aren't at 100%. More to do...

elParaguayo:
WindowTabs's test is not complete because of these lines:
https://github.com/qtile/qtile/blob/48f7c5e499fed64913732896a5fdc65fa921d2e8/libqtile/widget/windowtabs.py#L58-L59

I'm guessing this is necessary to prevent `AttributeErrors` for the subsequent tests but can `w` ever be `None`?

libqtile/widget/crashme.py
```diff
@@ -62,4 +62,3 @@ def button_press(self, x, y, button):
             1 / 0
         elif button == 3:
             self.text = '<span>\xC3GError'
-            self.bar.draw()
```
elParaguayo, line 65: I don't think this line is neede to trigger the error.

test/widgets/test_sep.py
```diff
@@ -57,3 +57,10 @@ def test_padding_and_width(manager_nospawn, minimal_conf_noscreen):
 
     w = topbar.info()["widgets"][0]
     assert w["width"] == 12
+
+
+def test_deprecated_config(caplog):
+    sep = widget.Sep(height_percent=80)
+
+    assert sep.size_percent == 80
+    assert "deprecated" in caplog.text
```
elParaguayo, line 66: This works on my system but it's not working here. Can anyone think why the log would not be captured here?

elParaguayo: Will just remove log text assertion. Don't think it impacts coverage as line is still run.

test/widgets/test_crashme.py
ramnes: LGTM

test/widgets/test_windowtabs.py
ramnes: Yeah I think we should have this fixed. :)

test/widgets/conftest.py
```diff
@@ -51,3 +56,27 @@ def call_soon(self, func, *args):
...
     return FakeQtile()
+
+
+# Fixture that defines a minimal configurations for testing widgets.
+# When used in a test, the function needs to receive a list of screens
+# (including bar and widgets) as an argument. This config can then be
+# passed to the manager to start.
+@pytest.fixture(scope='function')
+def minimal_conf_noscreen():
+    def add_screens_to_conf(screens):
```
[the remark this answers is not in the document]
elParaguayo: Because this way saves me one line of code in each test ;)

If it's overly complicated, I'll revert to the above.

test/widgets/test_countdown.py
elParaguayo: You're kidding? I was sure I got all of them...

test/widgets/test_windowtabs.py
elParaguayo: OK. May leave this as is for now because it's not the widget's fault. I'd need to do some more digging to work out why this is happening.

elParaguayo:
> Great work, thanks! After a quick glance, I'd just make the MinimalConfig a fixture in conftest but LGTM overall.
> 
> > Other widgets either have third party library dependencies, connect to external servers etc. so I'm not sure how we can hit 100% on those
> 
> For the libraries, feel free to add them to the test environment. For the external servers, your choices are either to mock the responses, to change the widgets so that they use dependency injection, or to setup a local test instance with docker-compose or alike if there's any open source server available. Because of our big number of tests and our wish to parallelize them, mocks are probably the easiest path.

@ramnes - would appreciate your thoughts on #2389.

test/widgets/test_sep.py
tych0: It would not surprise me if there's a bug in the calculations here.

test/widgets/test_windowtabs.py
```diff
@@ -0,0 +1,162 @@
...
+        return manager.c.bar["bottom"].info()["widgets"][0]["text"]
+
+    window_one = manager.test_window("one")
+    assert widget_text() == "!!one!!"
+
+    manager.kill_window(window_one)
+    assert widget_text() == ""
+
+
+# This test is needed to hit 100% coverage
```
tych0, line 140: Feels like you're right and we can just drop the "None" case, a window in the windows list should never be none.

test/widgets/test_sep.py
elParaguayo: OK - will take a look separately and can submit a PR if it needs fixing.

test/widgets/test_windowtabs.py
elParaguayo: As above - will take a look separately and can submit a PR if it needs fixing.

elParaguayo:
Anything outstanding on this one from your perspectives?

elParaguayo:
Do you want to remove the `if None` bit in a separate PR?

tych0:
Separate patch in this PR is fine with me, thanks!

elParaguayo:
@tych0 I know this is merged but have a quick look at #2405. That seems to be an example of where a window is `None`.

tych0:
As ramnes points out in that thread, it's not the window being none, but the window class. The window should never be None here.

PR #2389, opened 2021-04-16
Increase test coverage vol2

elParaguayo:
This is draft for now as I want to get opinions on the approach I've adopted here - see specific review comments below.

This is built on top of #2386 but is separated so we can merge in phases.

### DIFF
--- a/test/widgets/conftest.py
+++ b/test/widgets/conftest.py
@@ -68,7 +68,7 @@ def minimal_conf_noscreen():
         auto_fullscreen = False
         keys = []
         mouse = []
-        groups = [libqtile.config.Group("a")]
+        groups = [libqtile.config.Group("a"), libqtile.config.Group("b")]
         layouts = [libqtile.layout.stack.Stack(num_stacks=1)]
         floating_layout = libqtile.resources.default_config.floating_layout
         screens = []
--- a/test/widgets/test_check_updates.py
+++ b/test/widgets/test_check_updates.py
@@ -1,5 +1,6 @@
+import libqtile.config
 from libqtile.bar import Bar
-from libqtile.widget import CheckUpdates
+from libqtile.widget.check_updates import CheckUpdates, Popen  # noqa: F401
 
 
 def no_op(*args, **kwargs):
@@ -16,6 +17,7 @@ wrong_distro = "Barch"
 good_distro = "Arch"
 cmd_0_line = "export toto"   # quick "monkeypatch" simulating 0 output, ie 0 update
 cmd_1_line = "echo toto"     # quick "monkeypatch" simulating 1 output, ie 1 update
+cmd_error = "false"
 nus = "No Update Avalaible"
 
 
@@ -72,3 +74,119 @@ def test_no_update_available_with_no_update_string_and_color_no_updates(fake_qti
     text = cu4.poll()
     assert text == nus
     assert cu4.layout.colour == cu4.colour_no_updates
+
+
+def test_update_available_with_restart_indicator(monkeypatch, fake_qtile):
+    """ test output with no indicator where restart needed """
+    cu5 = CheckUpdates(distro=good_distro,
+                       custom_command=cmd_1_line,
+                       restart_indicator="*",
+                       )
+    monkeypatch.setattr("os.path.exists", lambda x: True)
+    fakebar = Bar([cu5], 24)
+    fakebar.window = FakeWindow()
+    fakebar.width = 10
+    fakebar.height = 10
+    fakebar.draw = no_op
+    cu5._configure(fake_qtile, fakebar)
+    text = cu5.poll()
+    assert text == "Updates: 1*"
+
+
+def test_update_available_with_execute(manager_nospawn, minimal_conf_noscreen, monkeypatch):
+    """ test polling after executing command """
+
+    # Use monkeypatching to patch both Popen (for execute command) and call_process
+
+    # This class returns None when first polled (to simulate that the task is still running)
+    # and then 0 on the second call.
+    class MockPopen:
+        def __init__(self, *args, **kwargs):
+            self.call_count = 0
+
+        def poll(self):
+            if self.call_count == 0:
+                self.call_count += 1
+                return None
+            return 0
+
+    # Bit of an ugly hack to replicate the above functionality but for a method.
+    class MockSpawn:
+        call_count = 0
+
+        @classmethod
+        def call_process(cls, *args, **kwargs):
+            if cls.call_count == 0:
+                cls.call_count += 1
+                return "Updates"
+            return ""
+
+    cu6 = CheckUpdates(distro=good_distro,
+                       custom_command="dummy",
+                       execute="dummy",
+                       no_update_string=nus,
+                       )
+
+    # Patch the necessary object
+    monkeypatch.setattr(cu6, "call_process", MockSpawn.call_process)
+    monkeypatch.setattr("libqtile.widget.check_updates.Popen", MockPopen)
+
+    config = minimal_conf_noscreen
+    config.screens = [
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar([cu6], 10)
+        )
+    ]
+
+    manager_nospawn.start(config)
+
+    topbar = manager_nospawn.c.bar["top"]
+
+    assert topbar.info()["widgets"][0]["text"] == "Updates: 1"
+
+    # Clicking the widget triggers the execute command
+    topbar.fake_button_press(0, "top", 0, 0, button=1)
+
+    # The second time we poll the widget, the update process is complete
+    # and there are no more updates
+    _, result = manager_nospawn.c.widget["checkupdates"].eval("self.poll()")
+    assert result == nus
+
+
+def test_update_process_error(fake_qtile):
+    """ test output where update check gives error"""
+    cu7 = CheckUpdates(distro=good_distro,
+                       custom_command=cmd_error,
+                       no_update_string="ERROR",
+                       )
+    fakebar = Bar([cu7], 24)
+    fakebar.window = FakeWindow()
+    fakebar.width = 10
+    fakebar.height = 10
+    fakebar.draw = no_op
+    cu7._configure(fake_qtile, fakebar)
+    text = cu7.poll()
+    assert text == "ERROR"
+
+
+def test_line_truncations(fake_qtile, monkeypatch):
+    """ test update count is reduced"""
+
+    # Mock output to return 5 lines of text
+    def mock_process(*args, **kwargs):
+        return "1\n2\n3\n4\n5\n"
+
+    # Fedora is set up to remove 3 from line count
+    cu8 = CheckUpdates(distro="Fedora")
+
+    monkeypatch.setattr(cu8, "call_process", mock_process)
+    fakebar = Bar([cu8], 24)
+    fakebar.window = FakeWindow()
+    fakebar.width = 10
+    fakebar.height = 10
+    fakebar.draw = no_op
+    cu8._configure(fake_qtile, fakebar)
+    text = cu8.poll()
+
+    # Should have 2 updates
+    assert text == "Updates: 2"
--- a/test/widgets/test_clock.py
+++ b/test/widgets/test_clock.py
@@ -1,0 +1,298 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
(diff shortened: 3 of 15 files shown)


libqtile/widget/generic_poll_text.py
```diff
@@ -52,8 +52,6 @@ def __init__(self, **config):
         self.add_defaults(GenPollUrl.defaults)
 
     def fetch(self, url, data=None, headers=None, is_json=True, is_xml=False):
-        if headers is None:
```
elParaguayo, line 55: `headers` can never be None as `fetch` is only called from `poll` and it always has `User-Agent` defined by default.

test/widgets/conftest.py
```diff
@@ -1,4 +1,6 @@
 import os
+import contextlib
```
elParaguayo, line 2: ignore these 2 lines. They shouldn't be here!

test/widgets/test_generic_poll_text.py
```diff
@@ -0,0 +1,118 @@
...
+# Widget specific tests
+
+import contextlib
+import sys
+from importlib import reload
+from types import ModuleType
+
+import pytest
+
+from libqtile.widget import generic_poll_text
```
elParaguayo, line 30: I have to import this way as I need to access all attributes of the `generic_poll_text` module in order to apply the monkeypatch.

test/widgets/test_generic_poll_text.py
```diff
@@ -0,0 +1,118 @@
...
+def test_gen_poll_url_not_configured():
+    gpurl = generic_poll_text.GenPollUrl()
+    assert gpurl.poll() == "Invalid config"
+
+
+def test_gen_poll_url_text(monkeypatch):
+    gpurl = generic_poll_text.GenPollUrl(json=False, parse=lambda x: x, url="testing")
+    monkeypatch.setattr(generic_poll_text, "Request", MockRequest)
+    monkeypatch.setattr(generic_poll_text, "urlopen", Mockurlopen)
+    generic_poll_text.Request.return_value = b"OK"
```
elParaguayo, line 83: monkeypatching the web requests and setting a desired return value.

test/widgets/test_generic_poll_text.py
```diff
@@ -0,0 +1,118 @@
...
+def test_gen_poll_url_xml(monkeypatch):
+    gpurl = generic_poll_text.GenPollUrl(json=False, xml=True, parse=lambda x: x, url="testing")
+    monkeypatch.setattr(generic_poll_text, "Request", MockRequest)
+    monkeypatch.setattr(generic_poll_text, "urlopen", Mockurlopen)
+    generic_poll_text.Request.return_value = b"OK"
+    with pytest.raises(Exception):
+        gpurl.poll()
+
+    with inject_mock_module(Mockxml, "xmltodict"):
+        reload(generic_poll_text)
```
elParaguayo, line 104: I've inkected my fake `xmltodict` module but I have to reload the widget module as the `ImportError` test is only run once when the module is loaded.

test/widgets/test_generic_poll_text.py
```diff
@@ -0,0 +1,118 @@
...
+    generic_poll_text.Request.return_value = b'{"test": "OK"}'
+    assert gpurl.poll()["test"] == "OK"
+
+
+def test_gen_poll_url_xml(monkeypatch):
+    gpurl = generic_poll_text.GenPollUrl(json=False, xml=True, parse=lambda x: x, url="testing")
+    monkeypatch.setattr(generic_poll_text, "Request", MockRequest)
+    monkeypatch.setattr(generic_poll_text, "urlopen", Mockurlopen)
+    generic_poll_text.Request.return_value = b"OK"
+    with pytest.raises(Exception):
```
elParaguayo, line 100: Might need to fix this part of the widget. Raising an Exception if the `xmltodict` module is not installed is a bit unnecessary.

test/widgets/test_check_updates.py
```diff
@@ -72,3 +76,115 @@ def test_no_update_available_with_no_update_string_and_color_no_updates(fake_qti
...
+    # The ONLY way the text can change here is if the check_updates command
+    # is triggered after the completion of the execute command (the default
+    # polling interval is too large to trigger an update here)
+    retries = 10
+    while topbar.info()["widgets"][0]["text"] != nus and retries > 0:
+        sleep(0.05)
+        retries -= 1
+
+    # Assert that we didn't time out and the execute command has been run
+    assert retries > 0
```
elParaguayo, line 151: This works on my system but not in the CI testing environment.  The test is failing because it times out. 

Any suggestions on how to redo/fix this test?

elParaguayo: Actually, I've got another idea. Will test later.

libqtile/widget/generic_poll_text.py
tych0: Can you make a separate commit for this change with just this as the message?

test/widgets/test_generic_poll_text.py
tych0: Can you make this a comment in the code?

test/widgets/test_generic_poll_text.py
tych0: What else could you do? The user has requested people parse xml and there's no library to parse it.

tych0: One way to fix it would just be to add a parser using the stdlib. But the stdlib is a lot more painful to use :)

libqtile/widget/generic_poll_text.py
elParaguayo: Of course.

test/widgets/test_generic_poll_text.py
elParaguayo: What I meant was, in the widget, rather than raising an exception, we could log an error but have the widget display an error message. Raising an Exception seems overkill.

tych0: Doesn't raising an exception log an error message?

elParaguayo: I'd expect you to get the full traceback message and the user wouldn't know there's an issue.

Admittedly, I haven't tested this theory...

test/widgets/test_clock.py
```diff
@@ -0,0 +1,332 @@
...
+    fakebar.height = 10
+    fakebar.draw = no_op
+    clk2._configure(fake_qtile, fakebar)
+
+    # An invalid timezone current causes a TypeError
+    with pytest.raises(TypeError):
+        clk2.poll()
+
+    # Default time is 10:20 and we add 1 hour for the timezone
+    assert "can not infer" in caplog.text
```
elParaguayo, line 119: Is it not possible to use `caplog` to check log output? Works on my machine but fails here.

I'll remove this line and just rely on the `TypeError`. That'll give us full coverage but it feels a little inelegant as I think the widget should handle the invalid timezone more gracefully than it does.

libqtile/widget/generic_poll_text.py
elParaguayo: #2396

test/widgets/test_clock.py
```diff
@@ -0,0 +1,330 @@
...
+            return cls
+
+        @classmethod
+        def gettz(cls, val):
+            return None
+
+    # Stop system importing these modules in case they exist on environment
+    monkeypatch.setitem(sys.modules, "pytz", None)
+    monkeypatch.setitem(sys.modules, "dateutil", None)
+    monkeypatch.setitem(sys.modules, "dateutil.tz", None)
```
elParaguayo, line 86: I suspect a lot of this patching could be put into a fixture as it's repeated in tests.

test/widgets/test_gmail_checker.py
```diff
@@ -0,0 +1,112 @@
...
+
+# This test is only required because the widget is written
+# inefficiently. display_fmt should use keys instead of indices.
+def test_gmail_checker_only_unseen(fake_qtile, monkeypatch):
+    monkeypatch.setitem(sys.modules, "imaplib", FakeIMAP("imaplib"))
+    reload(gmail_checker)
+
+    gmc = gmail_checker.GmailChecker(
+        display_fmt="unseen[{0}]",
+        status_only_unseen=True,
```
elParaguayo, line 101: This is silly. `display_fmt` should use keys e.g.
``` python
display_fmt="inbox[{messages}],unseen[{unseen}]"
```
we can then do away with the `status_only_unseen` flag.

Would need a `migrate` function to do this. One for a separate PR...

test/widgets/test_keyboardkbdd.py
```diff
@@ -0,0 +1,157 @@
...
+    window = _NestedWindow()
+
+
+@pytest.fixture
+def patched_widget(monkeypatch):
+    monkeypatch.setitem(sys.modules, "dbus_next.constants", Mockconstants("dbus_next.constants"))
+    from libqtile.widget import keyboardkbdd
+
+    # The next line shouldn't be necessary but I got occasional failures without it when testing locally
+    monkeypatch.setattr("libqtile.widget.keyboardkbdd.MessageType", Mockconstants.MessageType)
```
elParaguayo, line 75: This `keyboardkbdd` test is killing me. Runs fine when running just the test file but occasionally fails when running whole `test/widgets` folder. Makes no sense (or, rather, I'm too dumb to understand why).

Patching a patched import seems to fix it but is ugly as hell!

elParaguayo:
That's probably enough for this PR.

ramnes (approved):
LGTM

Can you just drop the `return ...; else:` blobs? :)

test/widgets/test_check_updates.py
```diff
@@ -72,3 +76,130 @@ def test_no_update_available_with_no_update_string_and_color_no_updates(fake_qti
...
+    class MockPopen:
+        def __init__(self, *args, **kwargs):
+            self.call_count = 0
+
+        def poll(self):
+            if self.call_count == 0:
+                self.call_count += 1
+                return None
+            else:
+                return 0
```
ramnes, line 114: Here for example:

```suggestion
                return None
            return 0
```

test/widgets/test_clock.py
ramnes: Yeah. Why not do it?

I use `@pytest.mark.usefixtures(​...)` for this kind of fixture that doesn't need to pass an argument.

test/widgets/test_clock.py
```diff
@@ -0,0 +1,330 @@
...
+    config.screens = [
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar([clk6], 10)
+        )
+    ]
+
+    manager_nospawn.start(config)
+
+    topbar = manager_nospawn.c.bar["top"]
+    sleep(0.1)
```
tych0, line 329: Let's not fix problems with sleeps. It's just a recipe for unreliable CI in the future.

I see this and one other sleep in the code; I wonder if we can do some synchronization via the IPC interface somehow? Like, having it do callbacks when particular hooks are fired or whatever, so you could fire up IPC and wait for the startup hook to be called.

More magically if we could inspect the qtile object from the test process, that would be handy. I fear that will bring lots of demons, so the above may be easier.

elParaguayo: Agreed. I was uncomfortable with the `sleep` call too. I actually raised this concern in my vol3 PR (https://github.com/qtile/qtile/pull/2414#discussion_r618102262) and have implemented the fix suggested there.

I can use:
``` python
manager_nospawn.c.widget["clock"].eval("self.tick()")
```

I'm also not a big fan of `eval`!

Your last suggestion would be the nicest but I don't know enough about IPC here. I had hoped a simple:
``` python
def cmd_get_object(self):
    return self
```
would work but it throws out errors and, if it were that easy, it would probably be implemented already,

elParaguayo: I've fixed this in the `test_check_updates.py` test too. A couple more things to update and then I'll push changes.

I note some other non-widget tests use `sleep` too. I'll add to my to do list and see if we can change those too.

elParaguayo:
Pushed these changes as separate commits so they can be seen easily. I'll squash them into the main tests when we're ready to merge.

test/widgets/test_clock.py
elParaguayo: Ah. Hadn't understood how `usefixtures` works. That will make things much neater/more readable. Will update.

test/widgets/test_clock.py
tych0: > I note some other non-widget tests use sleep too. I'll add to my to do list and see if we can change those too.

Awesome, thanks!

> would work but it throws out errors and, if it were that easy, it would probably be implemented already,

Yeah, the problem is that you need to share the memory across processes, which is not so simple. The qtile process is where the cmd_* stuff gets run, then it gets pickled and sent over a socket to the client process (which is what the test code is).

> I'm also not a big fan of eval!

eval() bothers me less than sleep() since it won't break when the host is under load, but doing some kind of real synchronization would be best. eval() is fine for now, though.

tych0:
Overall looks good to me if you want to squash, thanks!

elParaguayo:
Do you want it squashed into one or keep one commit per test?

tych0:
Whatever makes the most sense to you. Seems like hoisting stuff like `self.execute_polling_interval = 1` could be its own commit, but one commit per widget seems fine too.

elParaguayo:
Ok. Will sort later today.

elParaguayo:
Looks like that test timed out.

ramnes:
Thanks! :rocket:

PR #2391, opened 2021-04-17
Delete deprecated

tych0:
Let's delete a bunch of deprecated stuff (and in some cases, add type enforcement for things that have been deprecated).

Note that this does not deprecate command_, since that makes the auto-migration tests fail, which means that nobody would ever get the benefit of those migrations (which is maybe ok...). A commit to do that is here, if we want to add it: https://github.com/tych0/qtile/commit/5e058270c4770774240e99022dc0da8432b6760d

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -33,7 +33,6 @@ from __future__ import annotations
 import contextlib
 import os.path
 import sys
-import warnings
 from typing import TYPE_CHECKING, Callable, List, Optional, Union
 
 import xcffib.xproto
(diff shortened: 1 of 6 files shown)


m-col:
Looks like all of the deprecations in the PR commits are older than 0.17, so are all good. Perhaps we should hold on the command_ one for a bit?

ramnes:
No migration for these? :)

tych0:
> No migration for these?

If we want to add a migration, we have to leave them in the tree, because the config files have to parse both ways. See the CI failures on the command_ patch for why. These are all old enough that we can just delete them without a migration IMO.

tych0:
Did we come to a conclusion on whether we want to keep this stuff?

m-col:
Considering we've just released, now might be a good time to tear these things out!

PR #2392, opened 2021-04-17
README: drop coveralls stat

tych0:
1. I think coverage is kind of a silly metric
2. the coveralls stats haven't been updated since we switched to github
   actions.

Further, I don't really think our coverage stats are actually that
accurate. Most of the testing happens in a process that's forked from the
main process, and I don't think the coverage stuff quite handles this.

Perhaps we should delete the coverage infrastructure all together, but we
should definitely at least remove this until someone bothers to fix it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 
 **A full-featured, hackable tiling window manager written and configured in Python**
 
-|website| |pypi| |ci| |rtd| |coveralls| |license|
+|website| |pypi| |ci| |rtd| |license|
 
 Features
 ========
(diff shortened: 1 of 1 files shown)


elParaguayo:
I assume you still want more tests though!

tych0:
Yeah, definitely. Mostly just don't need this broken readme badge :-)

m-col (approved):
Obviously you're right but sad :( it'd be nice to have a high number to feel good about.

tych0:
> it'd be nice to have a high number to feel good about.

In my experience high numbers are meaningless; I've experienced numerous projects with 100% test coverage that didn't actually work.

This is because people who are obsessed with test coverage also love to monkey patch stuff to get the output they expect, instead of running an actual integration test to see if the thing really works. </rant>

Anyway, I'm happy to put the badge back if/when it works again.

elParaguayo:
> > it'd be nice to have a high number to feel good about.
> 
> In my experience high numbers are meaningless; I've experienced numerous projects with 100% test coverage that didn't actually work.
> 
> This is because people who are obsessed with test coverage also love to monkey patch stuff to get the output they expect, instead of running an actual integration test to see if the thing really works. </rant>


Ha. There's a fair bit of monkeypatching in my tests but I don't _think_ I'm overdoing it!

PR #2394, opened 2021-04-17
config: rename and document respect_window_state_requests

tych0:
This isn't just minimization requests, but *all* state changes that people
can optionally ignore.

### DIFF
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -178,6 +178,10 @@ configuration variables that control specific aspects of Qtile's behavior:
         we're a working one by default. We choose LG3D to maximize irony:
         it is a 3D non-reparenting WM written in java that happens to be
         on java's whitelist.
+    * - auto_minimize
+      - True
+      - If things like steam games want to auto-minimize themselves when losing
+        focus, should we respect this or not?
 
 
 Testing your configuration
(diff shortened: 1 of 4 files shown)


m-col:
Other than fullscreen and minimisation requests, what other requests do people commonly see that this would block?

tych0:
https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/XIconifyWindow.html says the only valid thing *is* IconicState, so maybe it really is just minimization? Something the X11 people eventually thought would be extended but never was?

m-col:
But more info here too: https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.4

I'm assuming that the mapping which you mention on the original PR https://github.com/qtile/qtile/pull/2382 refers to minimised windows un-minimising to the "normal" state? If so, @ramnes's `auto_minimise` suggestion might be the most appropriate.

m-col:
Or `auto_minimize` for you yanks ;)

ramnes:
Yeah I think auto_minimize is a better one. It may not be the most technically correct name but at least it's understandable. We could also change the code to only drop requests that ask for the iconic state so that there's no surprise at all.

tych0:
> We could also change the code to only drop requests that ask for the iconic state so that there's no surprise at all.

Yeah, I think that's the right move, then the naming makes sense. Take a look at the patch now.

m-col (approved):
Yep looks good :)

ramnes:
Cool!

PR #2396, opened 2021-04-17
Refactor generic_poll_text widget and add test

elParaguayo:
`headers` can never be None as `fetch` is only called from `poll` and it always has `User-Agent` defined by default.

### DIFF
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -51,10 +51,15 @@ class GenPollUrl(base.ThreadPoolText):
         base.ThreadPoolText.__init__(self, "", **config)
         self.add_defaults(GenPollUrl.defaults)
 
-    def fetch(self, url, data=None, headers=None, is_json=True, is_xml=False):
-        if headers is None:
-            headers = {}
-        req = Request(url, data, headers)
+        self.headers["User-agent"] = self.user_agent
+        if self.json:
+            self.headers['Content-Type'] = 'application/json'
+
+        if self.data and not isinstance(self.data, str):
+            self.data = json.dumps(self.data).encode()
+
+    def fetch(self):
+        req = Request(self.url, self.data, self.headers)
         res = urlopen(req)
         charset = res.headers.get_content_charset()
 
@@ -62,10 +67,10 @@ class GenPollUrl(base.ThreadPoolText):
         if charset:
             body = body.decode(charset)
 
-        if is_json:
+        if self.json:
             body = json.loads(body)
 
-        if is_xml:
+        if self.xml:
             body = xmlparse(body)
         return body
 
@@ -73,16 +78,7 @@ class GenPollUrl(base.ThreadPoolText):
         if not self.parse or not self.url:
             return "Invalid config"
 
-        data = self.data
-        headers = {"User-agent": self.user_agent}
-        if self.json:
-            headers['Content-Type'] = 'application/json'
-
-        if data and not isinstance(data, str):
-            data = json.dumps(data).encode()
-
-        headers.update(self.headers)
-        body = self.fetch(self.url, data, headers, self.json, self.xml)
+        body = self.fetch()
 
         try:
             text = self.parse(body)
--- a/test/widgets/test_generic_poll_text.py
+++ b/test/widgets/test_generic_poll_text.py
@@ -1,0 +1,132 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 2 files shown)


m-col:
Looks like it was to avoid having `{}` as a default kwarg value. Why are those even optional arguments? `fetch` is used only in one place and always has all arguments passed.

tych0:
This is one of the reasons 100% coverage is silly IMO: default args that are dicts are bad, but having named arguments isn't inherently a bad thing. So we have to write 10 lines of coverage to do this two line thing which just deals with a shortcoming in the language.

I agree, maybe we remove the default too?

elParaguayo:
The other way to think about is that writing tests forces you to really understand the code and can help identify redundant code, as is the case here.

I'll remove all defaults from `fetch` and push the update now.

m-col:
TBH some of those arguments appear completely unnecessary. `url`, `is_json` and `is_xml` are all attributes available directly and don't need to be passed around like that.

elParaguayo:
Not sure how I missed this reply. 

Yes, good point, can tidy this up more.

libqtile/widget/generic_poll_text.py
```diff
@@ -51,38 +51,34 @@ def __init__(self, **config):
...
-    def fetch(self, url, data=None, headers=None, is_json=True, is_xml=False):
-        if headers is None:
-            headers = {}
-        req = Request(url, data, headers)
+        self.headers["User-agent"] = self.user_agent
+        if self.json:
+            self.headers['Content-Type'] = 'application/json'
+
+        if self.data and not isinstance(self.data, str):
+            self.data = json.dumps(self.data).encode()
```
elParaguayo, line 59: All these bits only need to be set once so we can do it during `__init__`.

libqtile/widget/generic_poll_text.py
```diff
@@ -51,38 +51,34 @@ def __init__(self, **config):
...
-        headers = {"User-agent": self.user_agent}
-        if self.json:
-            headers['Content-Type'] = 'application/json'
-
-        if data and not isinstance(data, str):
-            data = json.dumps(data).encode()
-
-        headers.update(self.headers)
-        body = self.fetch(self.url, data, headers, self.json, self.xml)
+        body = self.fetch()
```
elParaguayo, line 81: Much neater!

elParaguayo:
Probably makes sense for me to extract the test for this from my other PR...

tych0:
I have to admit that making things more implicit by using object attributes instead of function arguments does not look that great to me. In this case it seems maybe ok because things like is_json are really properties of the object and won't change, but it does make this code less flexible for any future users.

It also changes the semantics of fetch(): users can't dynamically pass in post data, they have to set self.data. We don't have any internal users of this functionality, but I can imagine some external users of that. Might be worth a note in the changelog (or just dropping that piece).

elParaguayo:
The reason I did it is because all values seem to be static. I agree that it changes `fetch` but, as this can only be called from `poll`, that didn't seem problematic. 

Calling `fetch` externally doesn't really make sense. Even before this change, there was still no real way to change `data` other than setting `self.data`.

tych0:
> Calling fetch externally doesn't really make sense. Even before this change, there was still no real way to change data other than setting self.data.

Couldn't you just pass it as an argument, i.e. `fetch(data=...)`?

I disagree that it doesn't make sense: suppose I've written a widget that needs to make multiple different http requests to the same API to render its content (like finding the weather in two cities or whatever), presumably I would want to call fetch() twice with different city names in the post data.

elParaguayo:
I don't disagree with that last part at all. My point was that this is not how this widget is set up to work. 

It is a widget that that just updates text based on a defined interval. Text is updated by calling `poll` and it's only `poll` that's calling `fetch`. If you want to call `fetch` with different parameters, then you've got to reconnect the response to `parse` and then update the text. It's just not expecting parameters to change.

Your scenario, which is a good example, needs some more fundamental changes.

tych0:
But suppose I have an out of tree widget that overrides poll:

```
def poll(self):
    r1 = self.fetch(data=d1)
    r2 = self.fetch(data=2d)
    return combine_responses(r1, r2)
```

wouldn't that work with the previous code and be broken with this?

elParaguayo:
Nice. Yup, that would do it.

So, if we don't want to break anything for people who might have out of tree widgets, I can just change `fetch` to have keyword arguments with the defaults being the instance attributes.

tych0:
> I can just change fetch to have keyword arguments with the defaults being the instance attributes.

I don't think that will compile, unfortunately. Best is to just leave data as a kwarg that poll() passes to fetch(). I agree that the other args can be removed.

ramnes:
LGTM as is, and LGTM with data as a kwarg as well. I think you're being a bit overzealous here @tych0, though. :)

test/widgets/test_generic_poll_text.py
```diff
@@ -0,0 +1,132 @@
...
+    monkeypatch.setattr(generic_poll_text, "urlopen", Mockurlopen)
+    generic_poll_text.Request.return_value = b"OK"
+    with pytest.raises(Exception):
+        gpurl.poll()
+
+
+def test_gen_poll_url_xml_has_xmltodict(monkeypatch):
+    # injected fake xmltodict module but we have to reload the widget module
+    # as the ImportError test is only run once when the module is loaded.
+    monkeypatch.setitem(sys.modules, "xmltodict", Mockxml("xmltodict"))
```
ramnes, line 117: Have a look at pytest-mock, I think it can make things cleaner and simpler

ramnes (approved):

tych0:
> I think you're being a bit overzealous here @tych0, though. :)

Fair enough, I'll merge as-is.

test/widgets/test_generic_poll_text.py
elParaguayo: Don't make me change my `monkeypatch`es... They're all over #2389!!

elParaguayo:
> > I think you're being a bit overzealous here @tych0, though. :)
> 
> Fair enough, I'll merge as-is.

You can blame me when the bug reports come in!

Issue #2397, opened 2021-04-17
`display_kb` doesn't always display all keybindings


ep12:
As originally stated in #2393:
If a `KeyChord` has been activated, `qtile cmd-obj -o cmd -f display_kb` does not print all the keybindings. This  is due to https://github.com/qtile/qtile/blob/cefd75854b0e5d42de41f53bb89462c30deaf62e/libqtile/core/manager.py#L1068
where `self.keys_map` is not a map of all keybindings, but rather of the currently bound chord. This should be changed to `self.config.keys`. A direct consequence is that some code in `walk_binding` can be dropped since we don't need to handle `ks: Union[int, str]` and `kmm: Union[int, List[str]]` anymore, but `ks: str` and `kmm: List[str]` instead.

Issue #2393, opened 2021-04-17
walk_binding seems unnecessary according to the type annotations


tych0:
```
$ mypy --python-version=3.8 -p libqtile
libqtile/core/manager.py:1041: error: Left operand of 'and' is always true
# remove left operand test and we get:
$ mypy --python-version=3.8 -p libqtile
libqtile/core/manager.py:1046: error: If condition is always true
```

which means the whole function is a no-op since we return if that is true. seems like the type annotations are probably wrong at the very least, or this function doesn't do anything. Introduced in 16643034381c9c8ab95864905df09c8a46963c1a

/cc #2247 @ep12

ep12:
I guess mypy assumes that the function is always called with arguments adhering to the specification, e.g. https://github.com/qtile/qtile/blob/48f7c5e499fed64913732896a5fdc65fa921d2e8/libqtile/core/manager.py#L1038-L1039
The first check in line 1041 is just a sanity check to ensure that the `ks` and `kmm` values are of the correct type: https://github.com/qtile/qtile/blob/48f7c5e499fed64913732896a5fdc65fa921d2e8/libqtile/core/manager.py#L1041-L1042
The condition in parens should always be true. If not, just return. Maybe an error could be risen here... Theoretically, it shouldn't even make a difference if those two lines were removed. But just in case the type isn't quite right, it is better to deal with it early rather than wait for cryptic messages like `TypeError: can only join an iterable` to plop up.

The second one in https://github.com/qtile/qtile/blob/48f7c5e499fed64913732896a5fdc65fa921d2e8/libqtile/core/manager.py#L1045-L1046
is less obvious to me. If we have a `KeyChord` object, the `modifiers` attribute is a list (as per configuration). So if we encounter a `KeyChord` inside the root `KeyChord` (that got its modifier and key translated to integers), then the condition should not be true.

Aside: How exactly did you run `mypy`? If I run it with the arguments you passed, I get `Success: no issues found in 141 source files`. With `--strict` I get a lot of messages, but not the ones you saw.

Another thing: run `qtile cmd-obj -o cmd -f display_kb`. If that function was a no-op, you'd only see a header and none of your key bindings should be printed.

ep12:
[...]

PR #2398, opened 2021-04-18
Make `display_kb` output consistent

ep12:
Fixes: #2393.
Fixes: #2397.
Previously, not all key bindings were printed if a `KeyChord` mode was active (#2397). This change also simplifies the helper function `walk_binding`, in turn fixing #2393.


tych0:
Thanks!

PR #2399, opened 2021-04-18
Avoid changing groups when reconfiguring screens

m-col:


m-col:
Closes https://github.com/qtile/qtile/issues/2313

PR #2400, opened 2021-04-18
Fix keyboardkbdd widget colour bug

elParaguayo:
KeyboardKbdd takes an optional list of colours that can be used for different layouts. If the list is the wrong length, the code tries a previous index until it finds a colour.

However, the code listens for a ValueError rather than the correct IndexError so the widget crashes if the list is too short.


m-col:
Is there a test for that parameter, either already or in your big widget test PR?

elParaguayo:
I spotted this while writing a test. Didn't see the point in writing a test for the bug but the rest of the widget should be covered (or will be when I can get it to run in the testing environment - it keeps failing...)

m-col:
OK cool :)

PR #2402, opened 2021-04-18
Fix keyboardkbdd widget colour error (pt 2)

elParaguayo:
Following on from #2400 there's another bug in the `_set_colour` method. 

Just goes to show why we need tests!

ps @m-col sorry, didn't realise there were two bugs in this method. I've tested this change against my test and it does work.


m-col:
Nice :)

Issue #1792, opened 2020-06-25
Spotify problem


day4me:
The problem is that you can't open the dropdown menu in Spotify/Settings by clicking it with mouse or touchpad, but you can open it with keyboard. It works in different wm's(like awesome).

[https://streamable.com/hr2fg4](url)

You can see in the video, I tried to open dropdown menu by clicking it with my mouse, but it only worked when i opened it with keybord.

qtile version - 0.15.1
OS: arch linux
kernel: 5.7.5
spotify ver - 1:1.1.10.546-4 (aur)
x.org x server ver - 1.20.8

tych0:
On Thu, Jun 25, 2020 at 11:37:38AM -0700, day4me wrote:
> The problem is that you can't open the dropdown menu in Spotify/Settings by clicking it with mouse or touchpad, but you can open it with keyboard. It works in different wm's(like awesome).
> 
> [https://streamable.com/hr2fg4](url)
> 
> You can see in the video, I tried to open dropdown menu by clicking it with my mouse, but it only worked when i opened it with keybord.
> 
> qtile version - 0.15.1
> OS: arch linux
> kernel: 5.7.5
> spotify ver - 1:1.1.10.546-4 (aur)
> x.org x server ver - 1.20.8

Yeah, I see the same behavior. It's odd too, since the dropdown isn't
a second window, so it doesn't seem like qtile should be interfering,
but somehow it is.

ramnes:
Ah, funny, I've always thought this was a problem from the Linux version of the application itself, not Qtile. The workaround is to put your cursor on the menu item you want to click and then press Enter.

tych0:
On Fri, Jun 26, 2020 at 12:00:49AM -0700, Guillaume Gelin wrote:
> Ah, funny, I've always thought this was a problem from the Linux version of the application itself, not Qtile. The workaround is to put your cursor on the menu item you want to click and then press Enter.

Yeah, I sort of thought the same thing since it wasn't a separate
window. I never bothered to try it in other WMs though, so if it works
there, there's some qtile bug...

day4me:
> On Fri, Jun 26, 2020 at 12:00:49AM -0700, Guillaume Gelin wrote: Ah, funny, I've always thought this was a problem from the Linux version of the application itself, not Qtile. The workaround is to put your cursor on the menu item you want to click and then press Enter.
[...]

Issue #2404, opened 2021-04-18
Click-through drop-down menus on the web


ArtemSmaznov:
# Issue description
Items in hover/drop-down menus on the web do not register click and instead the page underneath the menu gets clicked.
This happens on a number of web sites but easiest I can think of is duolingo as there is a language menu that does not require an account. I am not sure if drop-down menus on all sites are behaving like this.

- I have reproduced it using the Brave / Chrome / Vivaldi browsers while on qtile.
- Reproduced it on both my PCs as well as with the default config on a new test user on the system.
- Reproduced it after disabling my picom compositor and rebooting.
- I was not able to reproduce the issue on AwesomeWM nor Cinnamon which makes me believe that this is caused by qtile

# Repro Steps
1. Open a GUI web browser
2. Navigate to https://www.duolingo.com/
3. Click "Site language: English" link on the top right
4. Select any other language

# Expected behavior
Click should be registered on the drop-down menu and the new language should be selected

# Actual result
The language selection menu closes and the new language is not selected. Instead the click registered on the page behind the menu (if there was another link under the menu where the cursor was it would have been clicked)

The only way to navigate these menus is to middle click (or right click and select open in new tab) them to open in new tab and close the old one which is quite annoying.

# Reproduction rate
20/20

# Qtile version
0.17.0-1

# Stack traces
n/a (blank logs)

# Configuration
Reproducible on default config

# System info
Arch Linux x86_64
Kernel: Linux 5.11.15-arch1-2

elParaguayo:
Yes - I have this bug too. See #1809 for an example.

ArtemSmaznov:
Thanks, that does seem exactly the same as described in the comments there. The title threw me off

ramnes:
Yeah, this bug is incredibly boring, especially since it doesn't happen on all drop-down menus.

Workarounds so far are either to middle-click (and thus open in a new tab), or right-click then left-click on the link (yes, that's weird, but it works). Or navigating the menu with the keyboard when it works.

Related issues (or possibly related): #1792 #1809 #1824

elParaguayo:
Any ideas how we'd approach debugging this? Is this something to do with how we handle mouse clicks in windows?

I'd love to help but not sure I have the skills.

ramnes:
I usually start by looking at the logs. This is what happens when I click on the Duolingo menu:

```
2021-04-21 10:31:12,227 libqtile core.py:_xpoll():L276  EnterNotify
[...]

Issue #1824, opened 2020-07-18
Focus Issues in chrome during tiling


piyushk:
# Issue description

When chrome is not floating in a tiling layout (MonadTall/Tile), I can't click on various interface boxes. Here are 2 examples:
- Create a bookmark by clicking star. The Bookmark Added/Edit Bookmark popup appears. You cannot click on any option in this popup. All clicks are captured by the underlying chrome window instead.
- If I type in a query in the chrome bar, I can't click on any of the autocomplete suggestions. I can hit the down arrow to select them though

If chrome is floating and not tiled, then these problems do not occur.

# Qtile version

Tested on latest master: https://github.com/qtile/qtile/commit/e87b32b7c5a8652f077f09441705e20286f2d6f7
Present in 0.15.1

I noticed this problem a long time ago, so it's been around for a few versions.

# Configuration

Same as https://github.com/qtile/qtile/blob/master/libqtile/resources/default_config.py, except with MonadTall.
```
layouts = [
    # layout.Max(),
    # layout.Stack(num_stacks=2),
    # Try more layouts by unleashing below layouts.
    # layout.Bsp(),
    # layout.Columns(),
    # layout.Matrix(),
    layout.MonadTall(),
    # layout.MonadWide(),
    # layout.RatioTile(),
    # layout.Tile(),
    # layout.TreeTab(),
    # layout.VerticalTile(),
    # layout.Zoomy(),
]
```

ramnes:
Yes, I can confirm, thank you for writing it down, super annoying. Can you also confirm that it does not happen with the Max layout? @tych0, that's what I was telling you about during last FOSDEM.

piyushk:
@ramnes I can confirm it doesn't happen with the max layout.

tych0:
Yep, I can reproduce this on other layouts. Never noticed because I use the max layout for my browser :)

tych0:
I spent some time thinking about this today. It turns out that the bookmarks window is not actually an X window, so qtile isn't mis-focusing or anything.

That it only works in Max suggests that there's something about the offset calculation in chrome that's weird. One thing of note is that it has:

```_NET_WM_STATE(ATOM) = _NET_WM_STATE_MAXIMIZED_VERT, _NET_WM_STATE_MAXIMIZED_HORZ```

[...]

Issue #1809, opened 2020-07-12
Brave "edit bookmark" menu unclickable


dodopod:
When bookmarking a page in the Brave browser, left clicks pass through the "Bookmark added"/"Edit bookmark" menu. So, if there is a link under the menu, trying to click on the menu will end up clicking on the link instead. Middle and right clicks still work as expected.

The layout also seems to matter. This issue occurs in the MonadTall and Floating layouts, but Max seems to work fine.

I'm running QTile v0.15.1, and Brave 1.10.97 on Arch Linux.

tych0:
Can you post an xprop for the window?

dodopod:
Here it is:

```
XdndTypeList(ATOM) = STRING, UTF8_STRING, TEXT, _NETSCAPE_URL, chromium/x-renderer-taint, text/html, text/plain, text/x-moz-url
_NET_WM_DESKTOP(CARDINAL) = 0
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_STATE(ATOM) = _NET_WM_STATE_MAXIMIZED_VERT, _NET_WM_STATE_MAXIMIZED_HORZ
_NET_WM_USER_TIME(CARDINAL) = 19911973
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: 315, 262
		program specified minimum size: 500 by 100
WM_NAME(UTF8_STRING) = "xprop(1): property displayer for X - Linux man page - Brave"
_NET_WM_NAME(UTF8_STRING) = "xprop(1): property displayer for X - Linux man page - Brave"
XdndAware(ATOM) = BITMAP
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x2, 0x0, 0x1, 0x0, 0x0
_NET_WM_ICON(CARDINAL) = 	Icon (128 x 128):
	                                            ░▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒░                                           
	                                           ░▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒░                                          
[...]

PR #2413, opened 2021-04-21
Cleaner clicks

ramnes:
Here are various bits I've cleaned while investigating for #2404. Have a look at the commit messages for more details.


ramnes:
The last commit I've added fixes  #1792 #1809 #1824 and #2404 for users of `follow_mouse_focus = True` (that is the default value).

m-col:
Are they still a problem when users have follow_mouse_focus as `False`?

elParaguayo:
This doesn't actually work for me. Tried with `follow_mouse_focus` as `True` and `False` and it fails both times.

elParaguayo:
Ignore that. Didn't save config after changing `follow_mouse_focus` to `True`. Works when `True` fails when `False` but I think that was expected.

ramnes:
> Are they still a problem when users have follow_mouse_focus as False?

> Works when True fails when False but I think that was expected.

Yes. It's just a partial fix so that it stops annoying most people, but more work is needed to have it right in all use cases.

ramnes:
Alright, I'm removing that last commit for now, because it has a pitfall: focus on click is actually useful when we switch between layouts and that the mouse ends up on a window that doesn't have focus yet. That will probably help the rest of this PR to be merged with less discussion as well. 

If anyone wants to use this commit, it's still available here: https://github.com/ramnes/qtile/tree/remove-click-focus

libqtile/backend/x11/core.py
```diff
@@ -577,6 +577,8 @@ def handle_ButtonPress(self, event) -> None:  # noqa: N802
         self.qtile.process_button_click(
             button_code, state, event.event_x, event.event_y, event
         )
+        self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayPointer, event.time)
+        self.conn.conn.flush()
```
tych0, line 581: Overall this looks good to me, though I wonder what this flush is for. Seems like it should work without this?

tych0: (I understand that you copy/pasted it, I'm wonder if we can just drop it)

ramnes: Probably useless, yeah.

PR #2414, opened 2021-04-21
Widget tests - vol 3

elParaguayo:
Carrying on my process of working through the widgets and creating tests.

For the music lovers, there are some real crowd-pleasers in the `moc` test!

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -82,7 +82,7 @@ class Net(base.ThreadPoolText):
         interfaces = {}
         if self.interface == ["all"]:
             net = psutil.net_io_counters(pernic=False)
-            interfaces["all"] = {'down': net[1], 'up': net[0]}
+            interfaces["all"] = {'down': net.bytes_recv, 'up': net.bytes_sent}
             return interfaces
         else:
             net = psutil.net_io_counters(pernic=True)
@@ -102,8 +102,8 @@ class Net(base.ThreadPoolText):
     def poll(self):
         ret_stat = []
         try:
+            new_stats = self.get_stats()
             for intf in self.interface:
-                new_stats = self.get_stats()
                 down = new_stats[intf]['down'] - \
                     self.stats[intf]['down']
                 up = new_stats[intf]['up'] - \
--- a/test/widgets/test_moc.py
+++ b/test/widgets/test_moc.py
@@ -1,0 +1,193 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Widget specific tests
+
+import subprocess
+
+import pytest
+
+import libqtile.config
+from libqtile.bar import Bar
+from libqtile.widget import moc
+
+
+class MockMocpProcess:
+    info = {}
+    is_error = False
+    index = 0
+
+    @classmethod
+    def reset(cls):
+        cls.info = [
+            {
+                'State': "PLAY",
+                'File': "/playing/file/rickroll.mp3",
+                'SongTitle': "Never Gonna Give You Up",
+                'Artist': "Rick Astley",
+                'Album': "Whenever You Need Somebody"
+                },
+            {
+                'State': "PLAY",
+                'File': "/playing/file/sweetcaroline.mp3",
+                'SongTitle': "Sweet Caroline",
+                'Artist': "Neil Diamond",
+                'Album': "Greatest Hits"
+                },
+            {
+                'State': "STOP",
+                'File': "/playing/file/itsnotunusual.mp3",
+                'SongTitle': "It's Not Unusual",
+                'Artist': "Tom Jones",
+                'Album': "Along Came Jones"
+                }
+        ]
+        cls.index = 0
+
+    @classmethod
+    def run(cls, cmd):
+        if cls.is_error:
+            subprocess.call(["exit", "1"])
+
+        arg = cmd[1]
+
+        if arg == "-i":
+            output = "\n".join("{k}: {v}".format(k=k, v=v) for k, v in cls.info[cls.index].items())
+            return output
+
+        elif arg == "-p":
+            cls.info[cls.index]['State'] = "PLAY"
+
+        elif arg == "-G":
+            if cls.info[cls.index]['State'] == "PLAY":
+                cls.info[cls.index]['State'] = "PAUSE"
+
+            elif cls.info[cls.index]['State'] == "PAUSE":
+                cls.info[cls.index]['State'] = "PLAY"
+
+        elif arg == "-f":
+            cls.index = (cls.index + 1) % len(cls.info)
+
+        elif arg == "-r":
+            cls.index = (cls.index - 1) % len(cls.info)
+
+
+def no_op(*args, **kwargs):
+    pass
+
+
+class FakeWindow:
+    class _NestedWindow:
+        wid = 10
+    window = _NestedWindow()
+
+
+@pytest.fixture
+def patched_moc(fake_qtile, monkeypatch):
+    widget = moc.Moc()
+    MockMocpProcess.reset()
+    monkeypatch.setattr(widget, "call_process", MockMocpProcess.run)
+    monkeypatch.setattr("libqtile.widget.moc.subprocess.Popen", MockMocpProcess.run)
+    fakebar = Bar([widget], 24)
+    fakebar.window = FakeWindow()
+    fakebar.width = 10
+    fakebar.height = 10
+    fakebar.draw = no_op
+    widget._configure(fake_qtile, fakebar)
+    return widget
+
+
+def test_moc_poll_string_formatting(patched_moc):
+
+    # Both artist and song title
+    assert patched_moc.poll() == "♫ Rick Astley - Never Gonna Give You Up"
+
+    # No artist
+    MockMocpProcess.info[0]["Artist"] = ""
+    assert patched_moc.poll() == "♫ Never Gonna Give You Up"
+
+    # No title
+    MockMocpProcess.info[0]["SongTitle"] = ""
+    assert patched_moc.poll() == "♫ rickroll"
+
+
+def test_moc_state_and_colours(patched_moc):
+
+    # Initial poll - playing
+    patched_moc.poll()
+    assert patched_moc.layout.colour == patched_moc.play_color
+
+    # Toggle pause
+    patched_moc.play()
+    patched_moc.poll()
+    assert patched_moc.layout.colour == patched_moc.noplay_color
+
+    # Toggle pause --> playing again
+    patched_moc.play()
+    patched_moc.poll()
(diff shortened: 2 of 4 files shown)


test/widgets/test_moc.py
```diff
@@ -0,0 +1,191 @@
...
+
+    manager_nospawn.start(config)
+
+    # When started, we have the first item playing
+    topbar = manager_nospawn.c.bar["top"]
+    assert topbar.info()["widgets"][0]["text"] == "♫ Rick Astley - Never Gonna Give You Up"
+
+    # Trigger next item and wait for update poll
+    topbar.fake_button_press(0, "top", 0, 0, button=4)
+    sleep(0.06)
```
elParaguayo, line 174: Not a big fan of using sleep here but I don't think I can force a `poll` when the widget is handled by `manager_nospawn`. `ThreadPollText` doesn't have a `cmd_poll` to be able to force this.

m-col: Would `manager_nospawn.c.sync()` not work?

m-col: Oh actually I misunderstood. Maybe it won't. Perhaps using `manager_nospawn.c.eval()`?

elParaguayo: I was hoping to use `manag_nospawn.c.widget["moc"].function()` but don't know what to pass as an argument as I can't seem to access the widget's internals once it's inside the manager.

Alternatively, I just as a `cmd_poll` to the base widget class which would allow users to force a poll whenever they wanted.

m-col: It seems all command objects have `cmd_eval` from the base, so you could do `manager_nospawn.c.widget["moc"].eval('self.poll()')`

elParaguayo: Interesting. I didn't think that would work as the `eval` wasn't called a method of the widget and so wouldn't pass `self` but it looks like it does work.

I can change this to: `success, result = manager_nospawn.c.widget["moc"].eval('self.poll()')` and then test `result`.

Thank you.

test/widgets/test_mpris2widget.py
```diff
@@ -0,0 +1,264 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Widget specific tests
+
+# The test_widget_init_config will cover the scenario where `kbdd` is not
+# running and will also run the asyncio _config_async method.
```
elParaguayo, line 24: (sigh) Need to delete.

test/widgets/test_mpris2widget.py
```diff
@@ -0,0 +1,264 @@
...
+
+class MockMessage:
+    def __init__(self, is_signal=True, body=None):
+        self.message_type = 1 if is_signal else 0
+        self.body = body
+
+
+class obj:  # noqa: N801
+    def __init__(self, value):
+        self.value = value
```
elParaguayo, line 62: `dbus_next` message data is stored in a `Variant` object. The widget extracts the information from the `value` attribute so that's all we need to mock here.

libqtile/widget/net.py
```diff
@@ -82,7 +82,7 @@ def get_stats(self):
         interfaces = {}
         if self.interface == ["all"]:
             net = psutil.net_io_counters(pernic=False)
-            interfaces["all"] = {'down': net[1], 'up': net[0]}
+            interfaces["all"] = {'down': net.bytes_recv, 'up': net.bytes_sent}
```
elParaguayo, line 85: Using attributes to be consistent (and it's easier to read).

elParaguayo:
Need to sort out a venv here. Test works fine on my machine. Probably because I have dbus_next installed.

test/widgets/test_moc.py
```diff
@@ -0,0 +1,193 @@
...
+    # Both artist and song title
+    assert patched_moc.poll() == "♫ Rick Astley - Never Gonna Give You Up"
+
+    # No artist
+    MockMocpProcess.info[0]["Artist"] = ""
+    assert patched_moc.poll() == "♫ Never Gonna Give You Up"
+
+    # No title
+    MockMocpProcess.info[0]["SongTitle"] = ""
+    assert patched_moc.poll() == "♫ rickroll"
```
tych0, line 128: :thumbsup:

test/widgets/test_moc.py
```diff
@@ -0,0 +1,193 @@
...
+    topbar.fake_button_press(0, "top", 0, 0, button=4)
+    _, result = manager_nospawn.c.widget["moc"].eval("self.poll()")
+    assert result == "♫ Neil Diamond - Sweet Caroline"
+
+    # Trigger next item and wait for update poll
+    # This item's state is set to "STOP" so there's no track title
+    topbar.fake_button_press(0, "top", 0, 0, button=4)
+    _, result = manager_nospawn.c.widget["moc"].eval("self.poll()")
+    assert result == "♫"
+    # assert topbar.info()["widgets"][0]["text"] == "♫"
```
tych0, line 183: Should we uncomment this?

tych0:
One small comment, but this looks good to me otherwise.

test/widgets/test_moc.py
elParaguayo: No, we should delete it. `poll` returns the text value, rather than updating the display. So, we can check the text by looking at the return value but we can't check the `text` value. 

Well, that's not quite true, I could ask the manager to eval `self.update(self.poll())`... Maybe that's better.

test/widgets/test_notify.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.fixture
+def notify_utility(monkeypatch):
+    monkeypatch.setitem(sys.modules, "dbus_next.aio", MockDBNAio("dbus_next.aio"))
+    monkeypatch.setitem(sys.modules, "dbus_next.service", MockDBNService("dbus_next.service"))
+    import libqtile.notify
+    reload(libqtile.notify)
+    return libqtile.notify
+
+
+# To send notifications via the manager object we need to send a string to eval
```
ramnes, line 80: Why can't we just send the notifications with dbus?

test/widgets/test_net.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def no_op(*args, **kwargs):
+    pass
+
+
+class FakeWindow:
+    class _NestedWindow:
+        wid = 10
+    window = _NestedWindow()
```
ramnes, line 39: Let's make this a fixture to avoid repetition

test/widgets/test_notify.py
elParaguayo: That would certainly be neater... However, what tools are available to send notifications with dbus? I think we've excluded `dbus_next` from the building environment. What did you have in mind?

test/widgets/test_net.py
elParaguayo: I'll look at that as a separate PR. I agree, there's definitely some tidying up that can be done with this.

test/widgets/test_notify.py
ramnes: notify-send or alike? But I guess isolation is the next problem, as we don't want to spam the host machine with notifications.

ramnes: We should probably just write a helper function that makes our notification server think there is a new notification coming in.

ramnes: Ah, but our notification service only exist if `dbus_next` is installed.

> I think we've excluded dbus_next from the building environment as it's only an optional dependency.

What is the rationale?

Worst case scenario, we could probably mock a notification manager, if I understand the code in `notify.py` correctly. That would be much cleaner than here, I think.

elParaguayo: If you can't isolate then the problem works the other way too e.g. if some other process on the host machine sends a notification when we're running our test, we'll get unexpected results.

ramnes: Yeah right. :)

What do you think of going with a custom notification manager then?

elParaguayo: I did think of that as an approach to use here and decided against it as I wanted to use the existing manager,

However, this test should really just check that the widget displayed notifications received from the attached manager so no reason not to use a custom one. It would probably make this one easier to read!

I was planning to do a separate test for the `NotificationService` and `NotificationManager` (not sure how, yet...).

m-col: Would it not be strictly superior as a test to install `dbus-next` and run `dbus` proper, establish the notification server and send it a real notification?

elParaguayo: Depends... do you want to test everything (service, manager and widget) in one test or do you test them separately?

Also, I don't know how the testing environment works. Is the notification service isolated from other processes (not just our various tests but any other processes on the server)?

m-col: I think test them separately or together where appropriate. I'm assuming that installing dbus and starting it with `dbus-launch` (https://dbus.freedesktop.org/doc/dbus-launch.1.html) before starting the tests would be enough, but I can't say I'm super familiar with it so not sure. I think you can get the dbus address from dbus-launch as it starts up and stick that in the environmental variables so that any dbus-interfacing code knows what to do. I guess `dbus-next` will "just work" as normal if it can get the address from the environment.

elParaguayo: Interesting. Let me do some investigating. As you say, `dbus_next` looks up the bus address from the environment so this may work.

elParaguayo: Any suggestions as to how I could test this locally (i.e. setting up a minimal environment and running dbus inside it etc?).

Alternative is likely to be lots of different commits being pushed and waiting for CI to run all the tests (the widgets are done last!)

m-col: I think something kinda like this might do it (there is probably a neater way):
```diff
diff --git a/test/conftest.py b/test/conftest.py
index 76c49923..8ce88410 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -487,6 +487,27 @@ def xephyr(request, xvfb):  # noqa: F841
         yield x
 
 
+@pytest.fixture(scope="function")
+def dbus(request):  # noqa: F841
+    use = getattr(request, "param", False)
+    if use:
+        launcher = whereis("dbus-launch")
+        if launcher is None:
+            yield None  # should probably raise an exception
+        else:
+            result = subprocess.run(launcher, stdout=subprocess.PIPE)
+            if result.returncode or not result.stdout:
+                yield None  # likewise
+            else:
+                env = {}
+                for line in result.stdout.strip():
+                    var, val = line.split("=")
+                    env[var] = val
+                yield env
+    else:
+        yield None
+
+
 @pytest.fixture(scope="function")
 def manager(request, xephyr):
     config = getattr(request, "param", BareConfig)
```
This can be parametrised with `True` to enable it for a test. I guess it should probably be used by `manager` or maybe even `xephyr` rather than tests directly so that the `env` that it yields can be put in the right place where downstream code (i.e. `dbus-next` can find the right environmental variables. I think that should work fine.

elParaguayo: Thanks. I was more concerned about testing locally (where I have `dbus_next` and `dbus` installed already) and being able to check whether this solution works before pushing to GitHub for the CI checks. i.e. just because a test works on my local machine, it doesn't guarantee it will work in CI.

m-col: I see... can `dbus_next` use dbus even if the environmental has been cleared of dbus-related variables?

ramnes: > I was planning to do a separate test for the `NotificationService` and `NotificationManager` (not sure how, yet...).

This is probably what we want to start with here. :)

Can you remove the notifications part of this PR so that I can merge the rest?

elParaguayo: Done.

ramnes:
Cool, thanks!

m-col:
Locally I'm getting these vulture failures when doing `make lint`:
```
/home/mcol/git/qtile/test/widgets/test_imapwidget.py:61: unused variable 'app' (100% confidence)
/home/mcol/git/qtile/test/widgets/test_net.py:50: unused variable 'nowrap' (100% confidence)
```
I was confused as to how these weren't picked up in the CI but turns out the vulture call used by `make lint` isn't part of the CI.

Should we add it? Was there a reason we hadn't already? In this case it has picked up unused variables missed by flake8.

ramnes:
Yeah I guess so

elParaguayo:
Both of those are where I've mocked a function and kept the same signature but we don't need to do anything with them. I could replace them with `*args` if that would get rid of the error but that doesn't seem ideal as we should check that the widget calls a function correctly.

While the error/warning is correct, I'm not convinced it needs fixing here. Can we tell the linter to skip these lines?

test/widgets/test_notify.py
elParaguayo: Update for the both of you: I stripped down `ci.yml` and `tox.ini` on a fork on my clone so that it only runs the `test_notify` and removes all other stuff to keep the testing really quick. I added some new dependencies and I can the widget to test successfully using `dbus_next`.

Lots more to do before I can submit a PR but it looks promising.

m-col: Awesome!

PR #2416, opened 2021-04-22
WIP: define our logging principles and apply them

ramnes:


ramnes:
Related discussion: #2408

m-col:
Do we really need to spell it out like that in the docs?

docs/manual/hacking.rst
```diff
@@ -87,6 +87,30 @@ to ensure your patch complies with reasonable formatting constraints. We also
...
+
+Logs are important to us because they are our best way to see what Qtile is
+doing when something abnormal happens. However, our goal is not to have as many
+logs as possible, as this hinders readability. What we want are relevant logs.
+
+To decide which log level to use, refer to the following scenarios:
+
+* ERROR: a problem affects the behavior of Qtile in a way that is noticeable to
+  the end user, and we can't work around it.
+* WARNING: a problem causes Qtile to operate in a sub optimal manner.
```
flacjacket, line 101: `sub optimal` -> `suboptimal`

tych0:
> Do we really need to spell it out like that in the docs?

It seems silly, but it's probably nice to have a record of this conversation. Otherwise in three year's we'll be like "I know we had this conversation and there was some conclusion but I don't remember what it was" :)

I agree nobody will read it before deciding what to log, though. :D

libqtile/backend/x11/core.py
```diff
@@ -427,7 +427,7 @@ def grab_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
 
         for code in codes:
             if code == 0:
-                logger.warning(
+                logger.debug(
```
tych0, line 430: This still feels like a warning, the user asked for something and they didn't get it, right?

ramnes:
Yeah I don't really expect people to read it, but it's needed as a reference for further nitpicking. :)

m-col:
OK fair enough :)

libqtile/backend/x11/core.py
ramnes: > If it can't make sense to your grandma, it's probably meant to be a DEBUG message.

ramnes: > I don't know, that's kind of a nonsense heuristic. My grandmother has late stage dementia, most of the world doesn't make sense to her.

And I don't have a grandma anymore. That was a metaphor. :)

> How will people find out their keys weren't correctly mapped without this message? I'm imagining they'll file a bug, we'll tell them to run it with --debug, and then they'll see this message. Seems like we could short circuit that process by leaving it as a warning.

Nobody knows what a keysym and a modmask are, especially when printed as hex, so I don't think it would short circuit anything. That being said I don't mind much because I'm not even sure it's possible to enter that branch ; invalid keys will return a ConfigError and fallback to the default config long before we end up here. So I went back to a warning message, but tried to make it a bit more comprehensive. Does it look good to you?

ramnes: I'm not 100% sure it's dead though. :)

ramnes: Actually I do trigger it when I bind `` ` `` but... strangely enough, the key still works. :D

```
Can't grab <Key (['mod4'], dead_grave)> (unknown keysym: 0xfe50)
Can't grab <Key (['mod4', 'shift'], dead_grave)> (unknown keysym: 0xfe50)
```

libqtile/backend/x11/core.py
```diff
@@ -363,9 +362,6 @@ def _get_target_chain(self, event_type: str, event) -> List[Callable]:
 
         if hasattr(self, handler):
             chain.append(getattr(self, handler))
-
-        if not chain:
-            logger.info("Unhandled event: {event_type}".format(event_type=event_type))
```
m-col, line 368: This is my favourite bit

libqtile/backend/x11/core.py
m-col: Maybe there are 2 codes that that key, and one is working, one is hitting the log?

elParaguayo:
I'm still in favour of merging this one. Can we fix the conflict and get this done?

ramnes:
Ask and receive. :)

PR #2418, opened 2021-04-25
wayland docs, changelog, couple extras

m-col:
Here are a couple extra commits I had finished, as well as a migration for `libqtile.window` to `libqtile.backend.x11.window` and a doc entry for installing the deps required for using the backend.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - widget.BatteryIcon no longer has a fallback text mode; use
           widget.Battery instead
         - MonadX layout key new_at_current is deprecated, use new_client_position.
+        - `libqtile.window` has been moved to `libqtile.backend.x11.window`; a migration has been added for this.
     !!! deprecation warning !!!
         - 'main' config functions, deprecated in 0.16.1, will no longer be executed.
     * features
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,10 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - added new argument for CheckUpdates widget: `custom_command_modify` which allows user to modify the
           the line count of the output of `custom_command` with a lambda function (i.e. `lambda x: x-3`).
           Argument defaults to `lambda x: x` and is overridden by `distro` argument's internal lambda.
+        - A Wayland backend has been added which can be used with `qtile start -b wayland`. It is undergoing heavy development
+          and bugs are to be expected, but if you find any bugs or conspicuously missing features that are not listed on our
+          to do list at https://github.com/qtile/qtile/discussions/2409 please do let us know. The backend requires pywlroots
+          and its dependencies python-xkbcommon, pywayland and wlroots 0.13.0.
 
 Qtile 0.17.0, released 2021-02-13:
     !!! Python version breakage !!!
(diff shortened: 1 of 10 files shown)


docs/manual/install/index.rst
```diff
@@ -111,3 +111,32 @@ Stable versions of Qtile can be installed from PyPI:
...
 however, it is recommended that you first install xcffib to ensure the
 cairo-xcb bindings are built (see above).
+
+Wayland
+-------
+
+Qtile can also be run as a Wayland compositor rather than an X11 window
+manager. This does not require an X server or ``xcffib`` to be installed.
+Instead Qtile uses pywlroots_, a Python binding around the wlroots_ library,
+pywayland_ and python-xkbcommon_. First, wlroots needs to be installed, and
```
flacjacket, line 121: We could add a `wayland` dependency install in `options.extras_require` in `setup.cfg`. Really, I think `pywlroots` will pull in `pywayland` and `python-xkbcommon`, so for now that should be the only dependency we need to add there.

docs/manual/install/index.rst
```diff
@@ -111,3 +111,32 @@ Stable versions of Qtile can be installed from PyPI:
...
+must be the latest release. Be aware that some distributions package outdated
+versions of wlroots. The three Python libraries can then be installed using
+``pip install`` as above.
+
+Qtile can then be run directly from your TTY, specifying the backend:
```
flacjacket, line 126: It might be good to note that this can also be run within a running X11 session, and will launch a nested window.

docs/manual/install/index.rst
```diff
@@ -111,3 +111,32 @@ Stable versions of Qtile can be installed from PyPI:
 As long as the necessary libraries are in place, this can be done at any point,
 however, it is recommended that you first install xcffib to ensure the
 cairo-xcb bindings are built (see above).
+
+Wayland
```
flacjacket, line 115: It may be good to note the current key limitations if there are things we know will be very broken at the moment.

CHANGELOG
```diff
@@ -27,6 +28,10 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - added new argument for CheckUpdates widget: `custom_command_modify` which allows user to modify the
           the line count of the output of `custom_command` with a lambda function (i.e. `lambda x: x-3`).
           Argument defaults to `lambda x: x` and is overridden by `distro` argument's internal lambda.
+        - A Wayland backend has been added which can be used with `qtile start -b wayland`. It is undergoing heavy development
+          and bugs are to be expected, but if you find any bugs or conspicuously missing features that are not listed on our
+          to do list at https://github.com/qtile/qtile/discussions/2409 please do let us know. The backend requires pywlroots
+          and its dependencies python-xkbcommon, pywayland and wlroots 0.13.0.
```
tych0, line 34: I would mention that any wayland-specific configuration is not intended to be stable at this time.

m-col:
The last release of pywlroots changes the interface for `Output`s so currently the backend won't work. I have a fix so I'll add that to this PR while I make the review changes tonight

docs/manual/install/index.rst
m-col: Yep sounds good. The main ones are Z positoning and the status bar, so I've added those.

docs/manual/install/index.rst
m-col: Yeah good idea

ramnes:
Moaaar

PR #2425, opened 2021-04-28
Also find wayland socket files with ipc.find_sockfile

m-col:
This makes `ipc.find_sockfile` also find Wayland socket files so that
IPC can work when Qtile is being run with the Wayland backend. This
algorithm is used to identify a socket file when the `display` is not
specified.

    - If WAYLAND_DISPLAY is set, use it.
    - else if DISPLAY is set, use that.
    - else check for the existence of a socket file for
      WAYLAND_DISPLAY=wayland-0 and if it exists, use it.
    - else check for the existence of a socket file for DISPLAY=:0.0 and
      if it exists, use it.
    - else raise IPCError

### DIFF
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -42,17 +42,49 @@ HDRLEN = struct.calcsize(HDRFORMAT)
 SOCKBASE = "qtilesocket.%s"
 
 
-def find_sockfile(display: str = None):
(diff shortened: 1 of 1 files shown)


tych0:
Seems reasonable to me, thanks.

ramnes:
Thanks!

PR #2426, opened 2021-04-28
Use shutil.which instead of manually searching PATH in conftest.py

m-col:
shutil is already used elsewhere in the tests so we may as well import
it and use it here too.


tych0:
I think it's in the stdlib in python3 right? But yeah, agreed.

m-col:
Yep it is

PR #2429, opened 2021-04-30
Wayland: implement support for wlr-layer-shell, xdg-decorations, and clean up listener code

m-col:
This adds support for the layer shell protocol. This allows for "layer"
    surfaces to be created by clients, used for applications such as status
    bars, overlays, third-party wallpapers (such as swaybg), and other tools
    like dmenu, rofi etc (their wayland clones that is).
    
    These clients are managed by Qtile as Static windows and are bound to an
    output rather than a group, and are not tiled.
    
    Part of layer shell protocol allows clients to request "exclusive
    zones", however this is not implemented in this commit.
    
    This depends on pywlroots>=0.2.7, which is reflected in tox.ini and
    setup.cfg.

--

This is another commit on top of #2418 but figured I should open a new PR for it as the other commits have already been reviewed once.

This also depends on a commit just merged into pywlroots that exposes this protocol which isn't available yet in a release. @flacjacket do you think it would be OK to make a new release with the layer shell commit?

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -39,20 +39,22 @@ from wlroots.wlr_types import (
     XCursorManager,
     XdgOutputManagerV1,
     input_device,
+    layer_shell_v1,
     pointer,
     seat,
-    xdg_shell,
 )
 from wlroots.wlr_types.cursor import WarpMode
 from wlroots.wlr_types.virtual_keyboard_v1 import (
     VirtualKeyboardManagerV1,
     VirtualKeyboardV1,
 )
+from wlroots.wlr_types.xdg_shell import XdgShell, XdgSurface, XdgSurfaceRole
 from xkbcommon import xkb
 
 from libqtile import hook
 from libqtile.backend import base
-from libqtile.backend.wayland import keyboard, output, window, wlrq
+from libqtile.backend.wayland import keyboard, window, wlrq
+from libqtile.backend.wayland.output import Output
 from libqtile.log_utils import logger
 
 if typing.TYPE_CHECKING:
@@ -91,7 +93,7 @@ class Core(base.Core):
 
         # set up outputs
         self.output_layout = OutputLayout()
-        self.outputs: List[output.Output] = []
+        self.outputs: List[Output] = []
         self._on_new_output_listener = Listener(self._on_new_output)
         self.backend.new_output_event.add(self._on_new_output_listener)
 
@@ -112,9 +114,12 @@ class Core(base.Core):
         self.cursor.motion_absolute_event.add(self._on_cursor_motion_absolute_listener)
 
         # set up shell
-        self.xdg_shell = xdg_shell.XdgShell(self.display)
+        self.xdg_shell = XdgShell(self.display)
         self._on_new_xdg_surface_listener = Listener(self._on_new_xdg_surface)
         self.xdg_shell.new_surface_event.add(self._on_new_xdg_surface_listener)
+        self.layer_shell = layer_shell_v1.LayerShellV1(self.display)
+        self._on_new_layer_surface_listener = Listener(self._on_new_layer_surface)
+        self.layer_shell.new_surface_event.add(self._on_new_layer_surface_listener)
 
         # Add support for additional protocols
         XdgOutputManagerV1(self.display, self.output_layout)
@@ -138,6 +143,7 @@ class Core(base.Core):
         self._on_new_input_listener.remove()
         self._on_request_set_selection_listener.remove()
         self._on_new_virtual_keyboard_listener.remove()
+        self._on_new_layer_surface_listener.remove()
 
         for kb in self.keyboards:
             kb.finalize()
@@ -187,7 +193,7 @@ class Core(base.Core):
             wlr_output.enable()
             wlr_output.commit()
 
-        self.outputs.append(output.Output(self, wlr_output))
+        self.outputs.append(Output(self, wlr_output))
         self.output_layout.add_auto(wlr_output)
 
     def _on_request_cursor(self, _listener, event: seat.PointerRequestSetCursorEvent):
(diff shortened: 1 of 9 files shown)


m-col:
Have a screenshot for a little demonstration:

image

The status bar at the top is [Waybar](https://github.com/Alexays/Waybar), and the notification window was created by the notification server [Mako](https://github.com/emersion/mako). Notice how Waybar has not been allocated dedicated space and so is occluding the top of the terminal window.

m-col:
I added a couple more commits because otherwise working on them was gonna conflict. The first one just tides up the way `Listener`s are used and finalised. The second one creates a manager for the xdg_decoration protocol so that client windows can request client- or server-side window decorations, and Qtile can tell windows to always use server-side decorations. This last one depends on https://github.com/flacjacket/pywlroots/pull/23

m-col:
Tests finally seem happy about this, and i learned more about mypy along the way. Good to go.

libqtile/backend/wayland/core.py
```diff
@@ -253,6 +242,30 @@ def _on_cursor_motion_absolute(self, _listener, event: pointer.PointerEventMotio
...
+        decoration.set_mode(xdg_decoration_v1.XdgToplevelDecorationV1Mode.SERVER_SIDE)
+
+    def _get_new_wid(self):
+        wid = 0
+        wids = self.qtile.windows_map.keys()
+        while True:
+            if wid not in wids:
+                break
+            wid += 1
+        return wid
```
ramnes, line 267: Why not just `max(self.qtile.windows_map.keys()) + 1`?

m-col: Yeah that'd work. Could any problem arise by having an ever-increasing number?

ramnes: I think the performance problem you'll get by iterating through all windows ids before creating a new window will hit you before any problem with a number that's too big, but I'm just guessing.

m-col: You don't need to have `INT_MAX` windows existing to have a window with ID `INT_MAX`. With the current version, you can have two windows with IDs 0 and 2, and the next window will get ID 1. But the every-increasing version can do:

1. have window with ID 0
2. next window gets ID 1
3. close window 0
4. next window gets ID 2
5. next window only needs to iterate through 2 windows

You could get a too big number with never opening more than 2 windows. Unless I'm misunderstanding what you meant?

I'm more wondering out of curiousity... tbh I would assume it would take literally years to get an error by incrementing the int too much :stuck_out_tongue:

Then again, doesn't `max(self.qtile.windows_map.keys()) + 1` iterate through them all anyway?

m-col: OK im gonna go with your version, it makes the code easier to understand and the difference either way is gonna be like microseconds or something lol

ramnes: > Then again, doesn't max(self.qtile.windows_map.keys()) + 1 iterate through them all anyway?

Yes but it only does it once. With `if wid not in wids`, you're doing it N times here, where N is the number of attempts at brute-forcing the first available wid. Not exactly the same time complexity. :)

> You don't need to have INT_MAX windows existing to have a window with ID INT_MAX.

No need to explain, I had it already. :) 

> the difference either way is gonna be like microseconds or something lol

Yeah I think we're miles ahead any problem both ways.

m-col:
I'm not going to add anything more to this PR as the next things I have worked on require changes not yet available in pywlroots, including that damage-tracking PR. So ready for review :)

m-col:
Correction, we got the damage tracking interface merged and released in pywlroots so I've appended that extra commit here.

m-col:
Mypy doesn't seem happy with something but isn't very clear as to what exactly. These are its complaints:
```
 libqtile/backend/wayland/window.py:144: error: Cannot determine type of 'damage'
  libqtile/backend/wayland/window.py:470: error: Cannot determine type of 'damage'
  libqtile/backend/wayland/window.py:474: error: Cannot determine type of 'damage'
  libqtile/backend/wayland/output.py:230: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:230: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:230: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:230: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "width"
  libqtile/backend/wayland/output.py:230: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "width"
  libqtile/backend/wayland/output.py:230: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "width"
  libqtile/backend/wayland/output.py:232: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "y"
  libqtile/backend/wayland/output.py:232: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "y"
  libqtile/backend/wayland/output.py:232: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "y"
  libqtile/backend/wayland/output.py:232: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "height"
  libqtile/backend/wayland/output.py:232: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "height"
  libqtile/backend/wayland/output.py:232: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "height"
  libqtile/backend/wayland/output.py:236: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:236: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:236: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "x"
  libqtile/backend/wayland/output.py:238: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "y"
  libqtile/backend/wayland/output.py:238: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "y"
  libqtile/backend/wayland/output.py:238: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "y"
```
The first 3 lines is refering to `output.damage`, where `output` is an item from `core.outputs`, and `core` and `core.outputs` are both typed, and `Output` has a `damage` attribute initialised in `__init__` so I don't see anything special or absent that is making it difficult for mypy to determine type.

The rest.... are straight up not true? `x` and `y` are set manually in `__init__` and `width` and `height` are properties of all 3 window types so not sure what that's all about. Anyone with superior mypy skillz able to give any insight?

m-col:
Awesome, looks like I figured it out. mypy is happy now. Good to go on my end.

ramnes:
Booyah!

PR #2431, opened 2021-05-01
Update wlroots compositor setup

flacjacket:
Modify the wayland core initialization to update the way the wlroots
compositor and backend are initialized based on the current pywlroots
master.


m-col:
Awesome thanks!

PR #2432, opened 2021-05-01
Check if the surface is an XDG surface before converting it

flacjacket:
This can cause runtime errors in cases where the previous surface is not an XDG surface.


m-col:
:+1:  thanks

PR #2434, opened 2021-05-01
Add test for notify widget

elParaguayo:
Following the discussion in #2414, I've added `dbus-next` to the test environment and added the packages needed to provide `dbus-launch` and `notify-send`.

The test for the notify widget now activates a dbus session, sends notifications over the bus and checks the output in the widget.

This PR also fixes two bugs that arise in two other dbus widgets (mpris2 and keyboardkbdd) that arise as a result of `dbus-next` now being installed in the testing environment.

Would be good to get feedback on the `dbus` fixture in `conftest.py` as I haven't accounted for the scenario where `dbus-launch` is not available as it should always be on CI but, theoretically, may be unavailable on people's systems which could create issues when running tests locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,8 @@ jobs:
                 sudo apt install --no-install-recommends \
                   libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz \
                   imagemagick libpulse-dev lm-sensors git xserver-xephyr xterm xvfb x11-apps ninja-build libegl1-mesa-dev \
-                  libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev libpixman-1-dev libpciaccess-dev
+                  libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev libpixman-1-dev libpciaccess-dev \
+                  dbus-x11 libnotify-bin
                 sudo pip -q install tox tox-gh-actions meson
             - name: Build wayland
               run: |
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,8 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - `libqtile.window` has been moved to `libqtile.backend.x11.window`; a migration has been added for this.
     !!! deprecation warning !!!
         - 'main' config functions, deprecated in 0.16.1, will no longer be executed.
+    !!! Notice for packagers - new dependencies !!!
+        - Tests now require the 'dbus-next' python module plus 'dbus-launch' and 'notify-send' applications
     * features
         - added measure_mem and measure_swap attributes to memory widget to allow user to choose measurement units.
         - memory widget can now be displayed with decimal values
--- a/test/widgets/conftest.py
+++ b/test/widgets/conftest.py
@@ -1,4 +1,7 @@
 import os
+import shutil
+import signal
+import subprocess
 
 import pytest
 
@@ -74,3 +77,44 @@ def minimal_conf_noscreen():
         screens = []
 
     return MinimalConf
+
+
+@pytest.fixture(scope='function')
+def dbus(monkeypatch):
+    # for Github CI/Ubuntu, dbus-launch is provided by "dbus-x11" package
+    launcher = shutil.which("dbus-launch")
+
+    # If dbus-launch can't be found then tests will fail so we
+    # need to skip
+    if launcher is None:
+        pytest.skip("dbus-launch must be installed")
+
+    # dbus-launch prints two lines which should be set as
+    # environmental variables
+    result = subprocess.run(launcher, capture_output=True)
+
+    pid = None
+    for line in result.stdout.decode().splitlines():
+
+        # dbus server addresses can have multiple "=" so
+        # we use partition to split by the first one onle
+        var, _, val = line.partition("=")
+
+        # Use monkeypatch to set these variables so they are
+        # removed at end of test.
+        monkeypatch.setitem(os.environ, var, val)
+
+        # We want the pid so we can kill the process when the
+        # test is finished
+        if var == "DBUS_SESSION_BUS_PID":
+            try:
+                pid = int(val)
+            except ValueError:
+                pass
+
+    # Environment is set and dbus server should be running
+    yield
+
+    # Test is over so kill dbus session
+    if pid:
+        os.kill(pid, signal.SIGTERM)
--- a/test/widgets/test_notify.py
+++ b/test/widgets/test_notify.py
@@ -1,0 +1,171 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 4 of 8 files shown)


test/widgets/conftest.py
```diff
@@ -74,3 +77,39 @@ class MinimalConf(libqtile.confreader.Config):
         screens = []
 
     return MinimalConf
+
+
+@pytest.fixture(scope='function')
+def dbus(monkeypatch):
+    # dbus-launch is provided by "dbus-x11" package
```
m-col, line 84: *on Ubuntu. On Arch it's provided by `dbus` which fortunately has no X dependencies.

m-col:
Awesome, looks like it wasn't that painful in the end to get dbus working in the CI, pretty cool. I think it would be fine for people running tests locally to skip the dbus and notify-send dependent tests if the dependencies cannot be found. You can use `pytest.mark.skipif` to mark those tests. There is an example here: https://github.com/qtile/qtile/blob/master/test/test_images2.py#L48

test/widgets/conftest.py
elParaguayo: Indeed. This was more to explain why the dbus-x11 package was added to CI.

elParaguayo:
> Awesome, looks like it wasn't that painful in the end to get dbus working in the CI, pretty cool. I think it would be fine for people running tests locally to skip the dbus and notify-send dependent tests if the dependencies cannot be found. You can use `pytest.mark.skipif` to mark those tests. There is an example here: https://github.com/qtile/qtile/blob/master/test/test_images2.py#L48

Amazing. Let me update that later.

test/widgets/conftest.py
m-col: Yeah good to keep that

elParaguayo:
`pytest.mark.skipif` is great. I was hoping to put a couple of functions in `conftest.py` for this so I could create a decorator for skipping tests. e.g.
``` python
@needs_app("dbus-launch")
def test_notification():
    ...
```
But it looks like you can only put fixtures in `conftest.py`. Am I missing something obvious?

Would be nice to have a `needs_app` and `needs_external_module` decorators for tests.

test/widgets/conftest.py
```diff
@@ -74,3 +77,44 @@ class MinimalConf(libqtile.confreader.Config):
...
+
+@pytest.fixture(scope='function')
+def dbus(monkeypatch):
+    # for Github CI, dbus-launch is provided by "dbus-x11" package
+    launcher = shutil.which("dbus-launch")
+
+    # If dbus-launch can't be found then tests will fail so we
+    # need to skip
+    if launcher is None:
+        pytest.skip("dbus-launch must be installed")
```
elParaguayo, line 90: Added skip: Skip any test that uses this fixture if `dbus-launch` is unavailable.

test/widgets/test_notify.py
```diff
@@ -0,0 +1,173 @@
...
+                                         timeout=10000)
+MESSAGE_3, NOTIFICATION_3 = notification("Low priority",
+                                         "Windows closed unexpectedly",
+                                         urgency=0)
+
+
+@pytest.mark.skipif(
+    shutil.which("notify-send") is None,
+    reason="notify-send not installed."
+    )
```
elParaguayo, line 91: As per earlier comment, I'd prefer a custom decorator rather than doing in full each time but it works.

CHANGELOG
```diff
@@ -10,6 +10,8 @@ Qtile 0.x.x, released xxxx-xx-xx:
         - `libqtile.window` has been moved to `libqtile.backend.x11.window`; a migration has been added for this.
     !!! deprecation warning !!!
         - 'main' config functions, deprecated in 0.16.1, will no longer be executed.
+    !!! Notice for packagers - new dependencies !!!
+        - Tests now require the 'dbus-next' python module plus 'dbus-launch' and 'notify-send' applications
```
elParaguayo, line 14: Is it worth leaving this in if we've now got ability to skip tests where they're not installed?

m-col: Maybe just add it to the `tox.init` test dependencies? I think it's all then automatic when using tox anyway.

m-col:
> But it looks like you can only put fixtures in conftest.py. Am I missing something obvious?

You can put functions, but make sure to import them as you would do normally i.e. `from test`.

CHANGELOG
elParaguayo: `dbus-next` is in the tox dependencies. My concern was more around the `dbus-launch` and `notify-send` apps. As you said in an earlier comment, these apps are packaged different in different distros. 

For example, on Arch, the PKGBUILD file for qtile runs the test suite in its `check` function (see https://github.com/archlinux/svntogit-community/blob/packages/qtile/trunk/PKGBUILD). There are two options for the Arch packagers: 1) add the relevant dependencies to `checkdepends`; or 2) do nothing and the test will now be skipped if those requirements aren't installed.

Option 2 doesn't seem too bad.

m-col: Ah I see ok. In that case it's probs best to keep this changelog entry so they can update checkdepends (if they're reading it) and no issues if they do not

m-col:
What's the status of this PR? Ready for review + merge? (After the small conflict gets resolved)

elParaguayo:
I think it's done other than that conflict which I'll do now.

.github/workflows/ci.yml
```diff
@@ -31,7 +31,8 @@ jobs:
                 sudo apt install --no-install-recommends \
                   libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz \
                   imagemagick libpulse-dev lm-sensors git xserver-xephyr xterm xvfb x11-apps ninja-build libegl1-mesa-dev \
-                  libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev libpixman-1-dev libpciaccess-dev
+                  libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev libpixman-1-dev libpciaccess-dev \
+                  dbus-x11 libnotify-bin
```
m-col, line 35: In what way is `dbus-x11` required, compared to other packages, presumably `dbus`? I haven't changed anything in my installation and notifications through dbus work normally in Wayland.

elParaguayo: It provides `dbus-launch` which the test uses to create a separate dbus session.

test/widgets/conftest.py
```diff
@@ -74,3 +77,44 @@ class MinimalConf(libqtile.confreader.Config):
         screens = []
 
     return MinimalConf
+
+
+@pytest.fixture(scope='function')
+def dbus(monkeypatch):
+    # for Github CI, dbus-launch is provided by "dbus-x11" package
```
m-col, line 84: Could you change this to say Ubuntu? That way if packagers are trying to figure out why their test is failing they'll find it easier to figure out how they can fix it rather than also figuring out how we configure the CI

test/widgets/test_notify.py
```diff
@@ -0,0 +1,173 @@
...
+    if urg_level != "normal":
+        text += '</span>'
+    text += '</span> - {body}'
+
+    text = text.format(subject=subject, body=body)
+
+    return text, cmds
+
+
+# "notify-send" is provided by libnotify-bin package
```
m-col, line 69: Same here, with 'Ubuntu'. E.g. on Arch it is provided by `libnotify`.

.github/workflows/ci.yml
m-col: RIght yeah, strange that it isn't provided by dbus directly considering it's also applicable to wayland environments. Fair enough.

test/widgets/test_notify.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # Clear the dispay
+    obj.clear()
+    assert obj.info()["text"] == ""
+
+    # Show the display
+    obj.display()
+    assert obj.info()["text"] == MESSAGE_3.format(colour=LOW)
+
+# untested: 92 - play sound
```
m-col, line 173: What does the 92 here mean?

elParaguayo: It's to do with the test coverage. It means this line is showing as untested:
https://github.com/qtile/qtile/blob/5ea61eb817791500851b30108e5696631cffc5ec/libqtile/widget/notify.py#L92

I was aiming for 100% coverage where possible.

m-col: Ah cool. If that line of code moves to a different line if notify.py gets edited, will the change in coverage result be picked up as a test failure?

elParaguayo: No. It was a note for me as I was writing the test (I started with all the lines that pytest said was untested and removed them as I incorporated them into the test).

Line numbers changing will have no impact on the test.

m-col: Do we leave it in?

elParaguayo: Not if you think it's confusing.

m-col:
Nice, this is awesome, thanks

PR #2435, opened 2021-05-01
qtile cmd-obj: convert integer arguments into ints

### LINKED ISSUE #2433: Error using qtile cmd-obj with an argument meant to be an int
Hello,

I'm trying to write a simple script to switch focus between screens and found that using
```
qtile cmd-obj -o cmd -f to_screen -a 0
```
in the terminal  causes a TypeError where the arg "0" can be any integer.

I'm guessing that this can be fixed with a simple "int()" somewhere, right? I'd make the pr myself but I'm new to github and need more experience with issues first.

---
```
$ qtile --version
0.17.1.dev166+g2f36ad91
```
---

```
$ qtile cmd-obj -o cmd -f to_screen -a 0

error: Sorry cannot run function 'to_screen' with arguments ['0']: Traceback (most recent call last):
  File "/home/andy/.local/lib/python3.8/site-packages/libqtile/command/interface.py", line 313, in call
  File "/home/andy/.local/lib/python3.8/site-packages/libqtile/core/manager.py", line 1293, in cmd_to_screen
    return self.focus_screen(n)
  File "/home/andy/.local/lib/python3.8/site-packages/libqtile/core/manager.py", line 851, in focus_screen
    if n >= len(self.screens):
TypeError: '>=' not supported between instances of 'str' and 'int'

```

Thanks!

m-col:
This checks through the argument list as passed from the shell when
using `qtile cmd-obj ... -a <arguments>` and converts any arguments that
appear to be integers into `int`s. This makes the commands that expect
`int`s receive `int`s, rather than `str`s. Fixes #2433.

### DIFF
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -125,8 +125,12 @@ def get_object(client: CommandClient, argv: List[str]) -> CommandClient:
     return client
 
 
-def run_function(client: CommandClient, funcname: str, args: List[str]) -> str:
+def run_function(client: CommandClient, funcname: str, args: List[str, int]) -> str:
     "Run command with specified args on given object."
+    for i, arg in enumerate(args):
+        if arg.lstrip("-").isdigit():
+            args[i] = int(arg)
+
     try:
         ret = client.call(funcname, *args)
     except SelectError:


ramnes:
I'd rather inspect type hints. This will probably break functions taking a string as argument and receiving a legit string made of digits. Groups, for example, are often called 1, 2, 3...

m-col:
Oh yeah good point. How can type hints be accessed?

elParaguayo:
> Oh yeah good point. How can type hints be accessed?

Well `dbus-next` inspects type hints as it uses custom hints to match dbus signatures. I think it uses the `inspect` module.

There's this: https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object

m-col:
> There's this: https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object

Great, thanks for this, looks useful

m-col:
The inspect stuff looks pretty straightforward. Where in the whole IPC communication makes the most sense to cast the arguments? Ideally it would be in the client side before sending the arguments to the server, and not somewhere where it is attempted during normal server-side operation (e.g. when resolving keybindings). It only would be needed when using a client from a shell too, not in Python clients. It seems that there isn't a client-side way to get a handle on the command that the client then wants to call. Looking into the IPC stuff I'm beginning to think this wasn't as simple as it seemed at first :sweat_smile:

m-col:
It makes sense for this to come after https://github.com/qtile/qtile/pull/2727 is merged

elParaguayo:
We should keep this open as it could be improved!

elParaguayo:
Taking a look at this.

Could we do something like this:
- create a method that returns the argument signature for the method (like an introspect method)
- have cmd_obj call this first and then try type conversions based on the signature

It would mean two calls instead of one but at least we can keep it client side.

Alternatively, if we want something on the server, we need a client to flag when everything is sent as a string so the server knows to do some type conversion.

@m-col any thoughts/preference?

m-col:
> Alternatively, if we want something on the server, we need a client to flag when everything is sent as a string so the server knows to do some type conversion.

I think this would be easier, just send the arguments as strings to the server, which does a quick lookup on the command's type signature and casts any non-strings. The more complicated part will be dealing with lists of things (e.g. `cmd_spawn`).

I have some ideas of how to implement this but I think it will be best to hold off until the command decorators PR is merged.

elParaguayo:
I had started playing with that idea too but don't remember where I got to.

Agreed with holding off for now.

atagen:
Can this be reopened/reconsidered? Even an ugly fix is better than not being able to use `cmd-obj`.

tych0:
Just looking through old PRs. Now that the command bits have landed, what about something like this? https://github.com/qtile/qtile/commit/3fb2cd3d04044511c82d84049a417079b09be0ab

It will require us to annotate types in function arguments in order to hoist correctly, but that doesn't seem like a bad thing IMO.

elParaguayo:
> Just looking through old PRs. Now that the command bits have landed, what about something like this? [3fb2cd3](https://github.com/qtile/qtile/commit/3fb2cd3d04044511c82d84049a417079b09be0ab)
> 
> It will require us to annotate types in function arguments in order to hoist correctly, but that doesn't seem like a bad thing IMO.

That's an elegant approach. I'm going to test it a bit.

elParaguayo:
How would that work where we have params that could take different types? or is that more a flaw in our code?

tych0:
I guess we could do something for option types (i.e. optional None), but yeah, for everything else it seems like a code smell to me at least.

tych0:
I guess the other problem is for compound types, e.g. List[int] or something. But for basic types it should work, which is maybe better than nothing?

tych0:
Ok, it's a little (well, a lot) more verbose when trying to handle union types: https://github.com/tych0/qtile/tree/type-conversions

this is all still untested. let me know if this looks interesting and i'll try to make it work for real.

fiorematteo:
I experimented a bit with https://github.com/tych0/qtile/commit/7f8aef7a432b11b83bd321feed49e54994dd1a1c, and came up with this:
```py
def lift_arg(typ: Union[type, str], arg):
    # annotations can be strings with the type name, or actual types
    if isinstance(typ, str):
        typ = eval(typ)
    if not isinstance(typ, type):
        raise TypeError("Could not evaluate to an actual type")

    # for stuff like int | None, allow either
    if get_origin(typ) is Union:
        for t in get_args(typ):
            if t == NoneType:
                # special case None? I don't know what this looks like
                # coming over IPC
                if arg == "":
                    return None
                if arg is None:
                    return None
                continue

            try:
                return lift_arg(t, arg)
            except TypeError:
                pass
        # uh oh, we couldn't lift it to anything
        raise TypeError(f"{arg} is not a {typ}")

    if get_origin(typ) is list:
        # i am explicitly too lazy to support heterogeneous lists, even
        # though it would be easy to do here. that would be an awful
        # user api.
        for t in get_args(typ):
            out = []
            try:
                for elt in arg:
                    out.append(lift_arg(t, elt))
            except TypeError:
                pass
            return out
        # uh oh, we couldn't lift it to anything
        raise TypeError(f"{arg} is not a {typ}")
```

From my testing this can handle builtin type as well as unions and lists.
I don't think it's a good idea to try and support more complex types.

The most annoying part is that annotations are saved as strings so we need to call eval to get the actual type object.
I'm not sure this is a good approach, handling the string case in each command seems simpler.
Also because for any of this to work we need to first annotate all commands.

tych0:
Yeah, I don't want to do eval; we can get rid of the strings if we get rid of all the `from __future__ import annotation`, which I believe is okay now that we only support 3.9? But maybe I'm missing something there.

elParaguayo:
I thought there was another reason we needed to use that line but can't remember what it was. I agree it would be nice to get rid of it.

Also, a few other ideas:
1) Should/could we limit argument types for exposed commands to basic types? If they're designed to be used with an external command interface, they shouldn't use complex types in the first place.
2) We could add a parameter to `@expose_command` to take a custom function used to parse more complex types
3) Ideally, we should limit any sort of type conversion activity so it only happens when calls are made via IPC i.e. not called directly or via `lazy` calls.

tych0:
I believe (3) is happening with my patch https://github.com/tych0/qtile/commit/7f8aef7a432b11b83bd321feed49e54994dd1a1c although I didn't update the commit message.

I wonder if we could invent a convention and have a `@staticmethod` that's `def from_ipc(data: str) -> ...`? Then we can do some kind of test in the lifting code and people still only have to define it once per-type, instead of at every call site.

tych0:
Ok, reading https://peps.python.org/pep-0563/ ; it seems like we should not drop `from __future__ import annotations` since that is what's going to happen in the future, i.e. delayed type evaluation, so things are `str`s and not their actual type. That seems... unfortunate to me.

they talk about `eval()` there, but then this one interface.py file has to import everything that could possibly ever be considered an arg type, which is also annoying.

tych0:
Ok; here is a maybe-not-embarassing PR: https://github.com/qtile/qtile/pull/4741

PR #2437, opened 2021-05-02
More widget tests

elParaguayo:
Memory widget: Added a test and also fixed a bug in the widget ("K" unit missing).

I think I've covered all the "easy wins". The other widgets are a bit more complicated so I might slow down a bit here!

### DIFF
--- a/test/widgets/test_memory.py
+++ b/test/widgets/test_memory.py
@@ -1,0 +1,101 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


test/widgets/test_memory.py
```diff
@@ -0,0 +1,98 @@
...
+        config = minimal_conf_noscreen
+        config.screens = [
+            libqtile.config.Screen(
+                top=libqtile.bar.Bar([widget], 10)
+            )
+        ]
+        manager_nospawn.start(config)
+
+        return manager_nospawn
+    return build_widget_in_manager
```
ramnes, line 79: I'm not fond of fixtures that return a function. I don't think the fixture has a value here, because you have to read the fixture to understand what the test really do. I would prefer the lines above being repeated in the two tests because it really is part of the test logic.

elParaguayo: No problem. Happy to amend.

elParaguayo: Sorry that took a while. Wanted to finish another test first. Done now.

ramnes:
Thanks a lot for your work on tests, that's really awesome. :)

PR #2439, opened 2021-05-03
Fix systray icon_size bug

### LINKED ISSUE #2438: Bad icon scaling in Systray widget
# Issue description
1. I added the system tray widget to my qtile bar with an icon size of 20.
2. I currently use the icon set Papirus Dark.
3. I launched the `nm-applet` through the autostart script and I obtained a really small icon for the `nm-applet` as you can see in the following image: 

image

The previous small icon also happened if I launched `nm-applet` through the terminal. 

One way I found to correct it was to setting the **icon size for the system tray to 16**. With this icon size, I obtained the following good results:

image

# Qtile version
`qtile 0.17.0`

# Configuration

My configuration can be found in the following [dotfiles](https://github.com/juanscr/dotfiles/tree/master/.config/qtile)

elParaguayo:
If icon has min_width/height hints, the icon size is not increased if the user specifies an icon greater than this size.

Fixes #2438

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -55,18 +55,14 @@ class Icon(window._Window):
         icon_size = self.systray.icon_size
         self.update_hints()
 
-        try:
-            width = self.hints["min_width"]
-            height = self.hints["min_height"]
-        except KeyError:
-            width = icon_size
-            height = icon_size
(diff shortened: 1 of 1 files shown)


m-col:
Cool cool, thanks.

PR #2442, opened 2021-05-04
Explain the Mirror widget.

tinruufu:
Most of this explanation is taken from the description of #1596, which is the only explanation I could find.


ramnes:
Cool, thanks!

m-col:
Awesome, thanks, I definitely should have added this to the original PR!

PR #2446, opened 2021-05-06
Wayland: implement damage tracking of outputs

m-col:
```
This implements damage tracking so that new frames are only rendered
when there has been a change of pixel contents, and only the regions
with changes are rendered.
```

--

This PR is sat atop #2429. I'm posting it as a draft because I could do with a little help if anyone's interested (with the damage tracking commit).

I'm pretty certain that the code that sets damage is perfectly fine, but I'm less certain that the rendering code is a fine. If the entire output is re-rendered when there is any damage (still a significant improvement from rendering any time a new frame can be rendered) then everything looks good, which is why I specifically think the code that selects only damaged regions to be re-rendered is the problem.

As far as I can tell, the rendering code mirrors the steps sway uses to select damaged regions and only render them. However, I see rendering artifacts where the damage is e.g. when I drag a window around. The damaged regions are cleared with black pixels and should immediately (before the frame is committed) be regenerated with the contents, but I still see black shapes when moving the window around. The window itself looks good, just not some areas around the edges it's moving to/from.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -42,6 +42,7 @@ from wlroots.wlr_types import (
     layer_shell_v1,
     pointer,
     seat,
+    xdg_decoration_v1,
 )
 from wlroots.wlr_types.cursor import WarpMode
 from wlroots.wlr_types.virtual_keyboard_v1 import (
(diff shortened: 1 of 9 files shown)


m-col:
If anyone does want to take a kick at it, it also requires this PR in pywlroots: https://github.com/flacjacket/pywlroots/pull/25

m-col:
I read a bunch of implementations and some useful webpages such as: https://emersion.fr/blog/2019/intro-to-damage-tracking/ about this topic.

The all-or-nothing approach is a great improvement, and going down the path of only rendering damaged regions could be another improvement (less significant), but at the cost of much more complicated code. I'm actually beginning to think more about the performance of Python for this type of job, too. If we were constantly (every time there is damage that trigger a new frame to be rendered) creating new classes (e.g. for `pixman_region32_t` wrappers) and going in and out of big nested function calls with lots of Python-C context switches.... is it really worth it? It's possible I'm over-worrying and there isn't much of an issue; I haven't benchmarked or anything. Or perhaps we might want the niceness of only rendering parts, but would want to write that all in C.

Anyway, I think the all-or-nothing approach is great to have, and we can keep more sophisticated damage re-rendering as a low priority if-someone-wants-to-check-for-improvements idea. I'll revise the current draft to make sure it's got all it should have for now and un-draft it (or just append the commit to the already open wayland PR).

PR #2450, opened 2021-05-07
Fix widgets missing from docs

### LINKED ISSUE #2449: Widgets missing from docs
Some widgets seem to have disappeared from the [official docs](http://docs.qtile.org/en/latest/manual/ref/widgets.html) (specifically `Mpris2` and `KeyboardKbdd`).

Given that these are two widgets that I've done some work on, it may be linked to that (but not immediately sure how).

I can do some bisecting and report back.

# Qtile version

Latest (3fb1e46 at time of posting)

EDIT: I can't replicate this on my system. However, given the identity of the two widgets, my guess is that this is related to `dbus-next` and that the modules need to be included in the `docs\conf.py` file here:
https://github.com/qtile/qtile/blob/3fb1e4645910c67bf6d302d59302dfb88f10c4ad/docs/conf.py#L27-L48

I'll submit a PR on that basis.

elParaguayo:
Two widgets are missing from the generated docs. This is because sphinx can't find `dbus-next` modules. This PR adds the modules
to the MOCK_MODULES list.

Fixes #2449


ramnes:
Thanks!

PR #2452, opened 2021-05-08
Stop dragging only when mouse button is released

### LINKED ISSUE #2440: Qtile Drag to Float using mod+super having issues
There seems to be some really weird issue:  when I use super + mouse1 to float a window , if I release super first then mouse1, qtile behaves as if the mod key is still pressed which means if I click on mouse1, the window also moves. I am on the latest qtile git
Also happens on the default config
```
qtile --version
0.17.1.dev230+gd9f7b281
```
Feel free to ask any logs

m-col:
This lets drags continue for as long as the mouse button is held down,
even if modifier keys are released. This is old behaviour that was
broken by 373ad55. Fixes #2440.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -110,6 +110,7 @@ class Mouse:
         self.button_code = int(self.button.replace('Button', ''))
         for k, v in kwargs.items():
             setattr(self, k, v)
+        self.modmask: int = 0
 
 
 class Drag(Mouse):
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -68,7 +68,7 @@ class Qtile(CommandObject):
         self.socket_path = socket_path
 
         self._drag: Optional[Tuple] = None
-        self.mouse_map: Dict[Tuple[int, int], List[Union[Click, Drag]]] = {}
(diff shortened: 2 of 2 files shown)


m-col:
Previously the modmask was inside the key of the dictionary:
```python
self.mouse_map: Dict[Tuple[int, int], List[Union[Click, Drag]]] = {}
```
but for mouse button releases we don't care about the modmask, we can keep the drag running even if the modifier state changes, only stopping the drag when the button is released. So instead of accessing `self.mouse_map` with `(button_code, modmask)` we instead just use `button_code`:
```python
self.mouse_map: Dict[int, List[Union[Click, Drag]]] = {}
```
But we still need the modmask around so we know what to do when a button is pressed, hence:
```python
        for m in self.mouse_map.get((button_code, modmask), []):
```
becomes:
```python
        for m in self.mouse_map.get(button_code, []):
            if not m.modmask == modmask:
                continue
```
The reason why it seems like I'm adding `Drag.modmask` when that wasn't there before I broke this behaviour is because part of that same breaking commit pre-calculated all of the modmasks once when the keys/buttons are grabbed during `_configure`, rather than on the fly every time a key/button binding is being called.

PR #2458, opened 2021-05-13
Adds temporary proxy module libqtile.window

m-col:
This adds a `libqtile.window` module that forwards all imports to
`libqtile.backend.x11.window` so that configs that import from the
former do not break upon next release.


ramnes:
This is quite some internal stuff, do we really have to care if it breaks?

m-col:
Hmm no not necessarily, if you think the risk is very low then we could do
without.

ramnes:
Well, is there any popular config use case of libqtile.window that we're aware of?

m-col:
Not that I'm aware of.... maybe you're right.

Issue #2457, opened 2021-05-13
cmd_reset function is not using my ratio value


RielPi:
The function `cmd_reset(self, redraw=True)` in [xmonad.py](https://github.com/qtile/qtile/blob/master/libqtile/layout/xmonad.py) supposed to reset all windows to a ratio of `0.5` or the ratio I set in the config file?

I don't use the default ratio of `0.5` which is used in `_med_ratio = 0.5`.
I added a variable under `_med_ratio`  called `reset_ratio` with the value I use for ratio and replaced in `cmd_reset()` function, `self.ratio = self._med_ratio` for `self.ratio = self.reset_ratio`.

I don't know how to grab the ratio I set in the config file and use the value for `reset_ratio`, my Python knowledge is a bit limited for this. :sweat_smile:

elParaguayo:
Are you working with a modified version of `xmonad.py`? If so, can you paste it here?

If you've modified the code so that `cmd_reset` is just looking for the `self.reset_ratio` value then you don't need to define the attribute in `xmonad.py`, you can just define `reset_ratio=...` in your config e.g. in your layouts `layout.MonadTall(reset_ratio=...)`. However, please note, if `reset_ratio` is not defined in the `defaults` list in `xmonad.py` you'll get an error if you try to reset but haven't set the attribute via your config.

Alternatively, `med_ratio` is just a class attribute so you should be able to change it in your config by having a line `layout.MonadTall._med_ratio = ...`. You could add this before `layouts` is defined.

RielPi:
I am using the default version of `xmonad.py`. I was able to workaround this, bare in mind I just omit other lines with `...` which they remain the same:

- I added `_reset_ratio` variable to a "default value"
- if statement in `__init__` to check if `ratio` value is different from the default one
- assign the value of `self._reset_ratio` to `self.ratio` in `cmd_reset`

```python
...
    _left = 0
    _right = 1
    _med_ratio = 0.5
    _reset_ratio = 0.5

...

    def __init__(self, **config):
        _SimpleLayoutBase.__init__(self, **config)
        self.add_defaults(MonadTall.defaults)
        if self.single_border_width is None:
            self.single_border_width = self.border_wi>
        if self.single_margin is None:
            self.single_margin = self.margin
        if self.ratio != 0.5:
            self._reset_ratio = self.ratio
        self.relative_sizes = []
        self.screen_rect = None

...

    def cmd_reset(self, redraw=True):
        "Reset Layout."
[...]

PR #2459, opened 2021-05-13
add `_reset_ratio` variable and if statement in `__init__` to fix the behavior in `cmd_reset`

RielPi:
`cmd_reset` was using default value of `0.5` from `_med_ratio`.

If `ratio` value is modified from the config file, `cmd_reset` kept going to a `ratio` of `0.5` and it didn't follow the new `ratio` value from the config file.

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -143,6 +143,7 @@ class MonadTall(_SimpleLayoutBase):
     _left = 0
     _right = 1
     _med_ratio = 0.5
+    _reset_ratio = 0.5
 
     defaults = [
         ("border_focus", "#ff0000", "Border colour for the focused window."),
@@ -181,6 +182,8 @@ class MonadTall(_SimpleLayoutBase):
             self.single_border_width = self.border_width
         if self.single_margin is None:
             self.single_margin = self.margin
+        if self.ratio != 0.5:
+            self._reset_ratio = self.ratio
         self.relative_sizes = []
         self.screen_rect = None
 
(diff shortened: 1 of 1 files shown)


libqtile/layout/xmonad.py
```diff
@@ -143,6 +143,7 @@ class MonadTall(_SimpleLayoutBase):
     _left = 0
     _right = 1
     _med_ratio = 0.5
+    _reset_ratio = 0.5
```
ramnes, line 146: Why another variable?

elParaguayo: @ramnes - the background to this is #2457.

You can probably get away with one variable if `_med_ratio` becomes configurable but there needs to be a check that it's between `min_ratio` and `max_ratio` (or it's automatically calculated to be the midpoint of these).

libqtile/layout/xmonad.py
```diff
@@ -227,7 +230,7 @@ def cmd_normalize(self, redraw=True):
 
     def cmd_reset(self, redraw=True):
         "Reset Layout."
-        self.ratio = self._med_ratio
+        self.ratio = self._reset_ratio = 0.5
```
RielPi, line 233: ops this should be `self.ratio = self._reset_ratio` instead of `self.ratio = self._reset_ratio = 0.5`.
I don't remember adding ` = 0.5`

libqtile/layout/xmonad.py
RielPi: When I use `cmd_reset` it reset all the layout including the master window but if I edit the `ratio` in the config file it keeps going back to `0.5`, I added this extra variable to avoid "issues" from the other functions variables "`min_ratio` and `max_ratio`" because they use `_med_ratio`

PR #2461, opened 2021-05-15
Remove unnecessary Qtile.mouse_position

m-col:
The `Qtile` object currently keeps track of mouse position in an
attribute at `self.mouse_position`, which is updated upon every mouse
event. However, it is only ever accessed in one place, in the x11
backend's `Window.cmd_set_position` to find a window under the pointer
to operate manipulate. Instead of constantly tracking and updating this
attribute we can query the pointer coordinates on demand from the X
server.


ramnes:
Alleluia

PR #2464, opened 2021-05-15
Remove Qtile.list_wids and client_from_wids

m-col:
This methods are also dead code. They are used in one place (which is
within `Qtile` itself) and due to changes in the format of
`Qtile.windows_map` the methods are unnecessary.


ramnes (approved):
LGTM, just waiting for the CI.

PR #2465, opened 2021-05-16
Complete type annotations for Qtile class 

m-col:
Three commits here. The third is the main one; the first two were actually the result of adding type annotations to `Qtile` and finding a couple things that needed changing because mypy was complaining so I separated those out before committing all the little annotations in the main commit. Together these commits add full type annotations to `Qtile` and a few superficial changes to make mypy happy about them. The first two commits are self-explanatory but the third one has a bit more meat. See its commit message.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -93,6 +93,13 @@ class Core(metaclass=ABCMeta):
     def graceful_shutdown(self):
         """Try to close windows gracefully before exiting"""
 
+    def simulate_keypress(self, modifiers: List[str], key: str) -> None:
+        """Simulate a keypress with given modifiers"""
+
+    def change_vt(self, vt: int) -> bool:
+        """Change virtual terminal, returning success."""
+        return False
+
 
 @enum.unique
 class FloatStates(enum.Enum):
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -189,6 +189,15 @@ class Window(_Window, metaclass=ABCMeta):
     def focus(self, warp: bool):
         """Focus this window and optional warp the pointer to it."""
 
+    @abstractmethod
+    def togroup(
+        self, group_name: Optional[str] = None, *, switch_group: bool = False
+    ) -> None:
+        """Move window to a specified group
+
+        Also switch to that group if switch_group is True.
+        """
+
     @property
     def has_focus(self):
         return self == self.qtile.current_window
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -5,17 +5,20 @@ import enum
 import typing
 from abc import ABCMeta, abstractmethod
 
-from libqtile.command.base import CommandObject, ItemT
+from libqtile.command.base import CommandObject
 
 if typing.TYPE_CHECKING:
-    from typing import Dict, List, Optional, Tuple, Union
+    from typing import Any, Dict, List, Optional, Tuple, Union
 
     from libqtile import config
+    from libqtile.command.base import ItemT
     from libqtile.core.manager import Qtile
     from libqtile.group import _Group
 
 
 class Core(metaclass=ABCMeta):
+    painter: Any
+
     @abstractmethod
     def finalize(self):
         """Destructor/Clean up resources"""
@@ -115,7 +118,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
     def __init__(self):
         self.borderwidth: int = 0
         self.name: str = "<no name>"
-        self.reserved_space: List = None
+        self.reserved_space: Optional[Tuple[int, int, int, int]] = None
         self.defunct: bool = False
 
     @property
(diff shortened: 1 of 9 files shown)


m-col:
I'm very interested in hearing arguments _against_ increasing our type annotation coverage if anyone is particularly against type checking in Python. I had made the assumption that as we have some, we'd probably want to increase this coverage. Indeed working on the Wayland backend I've found myself needing to figure out types by the code where there weren't annotations and I've made some type mistakes due to uncertain types. As a result I've grown to appreciate type hinting more, hence this PR. But anyway as I'm only recently a fan of Python typing I'm still eager to hear arguments against :)

ramnes:
I'm far to be a type hints fan but that's the way it is so let's push it. :)

libqtile/backend/base.py
```diff
@@ -93,6 +96,13 @@ def flush(self) -> None:
     def graceful_shutdown(self):
         """Try to close windows gracefully before exiting"""
 
+    def simulate_keypress(self, modifiers: list[str], key: str) -> None:
```
flacjacket, line 99: I think this will need to use `typing.List`, using the built-in type is supported in [Python 3.9](https://www.python.org/dev/peps/pep-0585/), but is not supported before that. Similarly for some of the other `[]` types annotations.

m-col: Ah yes. This was actually an intentional decision, though I neglected to bring it up for discussion in the post. As you said, built-in types are support for 3.9+ only. The rationale behind making this change was that the CI for mypy uses 3.9, so we could start to phase out the legacy types.

Then again, this might become problematic if people are trying to run mypy locally with older versions of Python. I guess that is the case?

m-col: I've updated these to use the `typing` versions for maintain compability with older versions.

m-col:
Rebased.

libqtile/command/base.py
```diff
@@ -83,7 +83,7 @@ def select(self, selectors: List[SelectorType]) -> CommandObject:
             if items is not None and selector and selector not in items:
                 raise SelectError("", name, selectors)
 
-            obj = obj._select(name, selector)
+            obj = obj._select(name, selector)  # type: ignore
```
flacjacket, line 86: Could the `type: ignore` be dropped if you annotate `obj: CommandObject = self` above? It might need a bit more tweaks, like doing:
```
maybe_obj = obj._select(name, selector)
if maybe_obj is None:
    ...
obj = maybe_obj
```

libqtile/core/manager.py
```diff
@@ -441,13 +465,13 @@ def add_group(self, name, layout=None, layouts=None, label=None):
...
 
-    def delete_group(self, name):
+    def delete_group(self, name: str) -> None:
         # one group per screen is needed
         if len(self.groups) == len(self.screens):
             raise ValueError("Can't delete all groups.")
         if name in self.groups_map.keys():
             group = self.groups_map[name]
-            if group.screen and group.screen.previous_group:
+            if group.screen and hasattr(group.screen, "previous_group"):
```
flacjacket, line 474: This looks like a bit odd, is there a reason the old way of having `previous_group` be an optional would not work with the mypy tests?

libqtile/core/manager.py
```diff
@@ -483,18 +507,25 @@ def register_widget(self, w):
...
 
     @property
-    def current_window(self):
+    def current_window(self) -> Optional[base.Window]:
         return self.current_screen.group.current_window
 
-    def reserve_space(self, reserved_space, screen):
+    def reserve_space(
+        self,
+        reserved_space: List[int],  # [left, right, top, bottom]
```
flacjacket, line 523: It might be good to have this be `Tuple[int, int, int, int]`, since it is required to have length 4. Similarly below.

flacjacket (commented):
This looks great, thanks for getting all of this put in place! Just a couple minor comments, overall I think this is alright.

libqtile/command/base.py
m-col: That little hack worked, thanks

libqtile/core/manager.py
m-col: Yep sounds good. I've updated these tuples in a few places, but had to add an ignore comment due to https://github.com/python/mypy/issues/7509. I've put an inline comment beside it so that in the future it can be checked and removed when mypy resolves this.

libqtile/core/manager.py
m-col: It does look a bit odd, but it's either this or peppering `assert`s throughout the code, which IMO makes things unclear as to _why_ asserts are lying around the code, as well as messy and executing unnecessary code. Same with `self.qtile: Qtile = None` that we have on a number of classes, which results in us having lots of lines like here: https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/core.py#L274

The `hasattr` lines are only required in the few places where these attributes can be accessed before getting populated, which in the case of `self.qtile` on things like windows pretty much never happens, but it can happen on `Screen` as you see here with `self.previous_group`.

To be honest I'm not sure if this is just a weakness in Python, or mypy, and it's probably just down to preference, so I'm happy to do whatever with these ones.

flacjacket (approved):
Nice!

PR #2467, opened 2021-05-16
Atomically apply new Qtile.screens when reconfiguring screens

m-col:
Currently `Qtile._process_screens` can be racy as it can be called a
second time when the X server is polled half through the function via
`self.core.get_screen_info()`. The two simultaneous calls then append to
`self.screens` resulting in duplicate items within `self.screens`. By
creating a function-local list of screens and appending to that, and
then setting it as `self.screens = screens` when it is complete, the
resulting `self.screens` will always be correct.

Ideally this would be avoided altogether but for some reason the X
server likes to emit two screen change notify events when the
configuration is changed with arandr.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -261,7 +261,7 @@ class Qtile(CommandObject):
 
     def _process_screens(self) -> None:
         current_groups = [screen.group for screen in self.screens if screen.group]
-        self.screens = []
+        screens = []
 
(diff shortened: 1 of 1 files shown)


ramnes:
> Ideally this would be avoided altogether but for some reason the X
> server likes to emit two screen change notify events when the
> configuration is changed with arandr.

Yeah I already noticed this. I wonder if it happens with another WM, or if we're triggering that second event somehow.

PR #2468, opened 2021-05-16
Wayland

m-col:
Some more nice Wayland developments.


m-col:
My next Wayland PR will have... Qtile bars!

20210520_11h34m54s_grim

PR #2476, opened 2021-05-20
Fix the PulseVolume widget disappearance from the doc.

Fabien-Bernard:
Add the _pulse_audio module imported by the widget to the mocked modules
of sphinx.

Corrects 2 typos.

--------------------------
**To be clear**, as I do not reproduce locally, I was not able to check the efficiency of this fix.
I still submit this PR based on a similar issue (#2449) which was fixed this way.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,6 +27,7 @@ class Mock(MagicMock):
 MOCK_MODULES = [
     'libqtile._ffi_pango',
     'libqtile.backend.x11._ffi_xcursors',
+    'libqtile.widget._pulse_audio',
     'cairocffi',
     'cairocffi.xcb',
     'cairocffi.pixbuf',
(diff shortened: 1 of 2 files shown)


m-col:
I also can't reproduce the issue locally. Does the system need to not have pulse installed or something?

Fabien-Bernard:
I generated the documentation using :
- my machine with pulseaudio package installed,
- a virtual machine without pulseaudio package. 
Both generations included the PulseVolume widget.

I still don't know how to reproduce the official documentation with the missing widget.

tych0:
I wonder if it's because the Read the Docs environment in which things are built doesn't have the dependencies. You can see various failures here complaining about pulseaudio headers:

https://readthedocs.org/projects/qtile/builds/14253498/

Maybe we need a .readthedocs.yaml that installs the dependencies?

elParaguayo:
It seems unlikely that this would break our docs and may well fix the issue as it did for the `dbus-next` widgets. Tych0's link shows there's an import error when reading the widget source so I'd be optimistic that this will work.

I'd therefore be inclined to merge this.

elParaguayo:
Nice. It worked.

PR #2485, opened 2021-05-24
Last big Wayland jump

m-col:
This is my final big chunk of work on the Wayland backend, and after this it will be back to smaller, more incremental changes and PRs (phew!). Note that the first 3 commits are https://github.com/qtile/qtile/pull/2465 because the typing helped a lot with identifying issues.

Pretty much everything on the to do list that I made (https://github.com/qtile/qtile/discussions/2409) has been completed. Some final things that are pretty important are:

- getting `Match` to work with Wayland equivalents (should be simple) (EDIT: this is added)
- creating a version of `Systray` (or modifying the existing one) for the backend (will have to research how swaybar does this)
- libinput configuration (requires some C-level code either within libqtile or pywlroots)

but after this PR my main focus will be working on making the tests work with the new backend.

The state of the Wayland backend at this point is completely usable! I've actually been running it for a few days just as my normal backend. I have a bool `IS_WAYLAND` in my config to e.g. choose what terminal I use, but this is made by importing one of the `Core`s and doing an `isinstance(qtile.core, Core)` which is kinda ugly... might be nice to add `Core.name` to get `"x11"` or `"wayland"` in the future (EDIT: this is added). I'm encouraging of testing and bug reports from others if anybody is interested in giving it a go!

Oh and all of the tests except those for `qtile check` will work (EDIT: this is fixed) -- I'm still trying to resolve some circular imports triggered by the type checking, but beside some minor fiddling to get mypy happy the code is ready for review.

@DorianGray part of this PR moves the `Drawer` into the backend with a small amount of abstracting. I didn't break it out fully because many of the methods can be re-used in their entirety by the Wayland `Drawer`. I'm afraid this makes https://github.com/qtile/qtile/pull/2186 redundant; however, I'm very interested if you have any feedback on the general rendering and cairocffi code used.


m-col:
Just fixed the mypy problem I had, it was due to mypy on my machine ignoring `PYTHONPATH` and using an older installed qtile version. Also added another commit to update the docs a bit.

> might be nice to add Core.name to get "x11" or "wayland" in the future

And added this while I was there, it's pretty useful.

m-col:
OK this is rebased with the updated typing changes and testing perfectly so ready for review :)

Graeme22:
Excited to try this out!

tych0:
Looks fairly reasonable to me, thanks!

Graeme22:
Got the chance to test this out. Seems to work great! Only noticed two things:
- KeyboardLayout widget didn't seem to work
- My alacritty font size dropped

m-col:
> KeyboardLayout widget didn't seem to work

Trying to get this fixed in https://github.com/qtile/qtile/pull/2511

> My alacritty font size dropped

That's funny. Do you know if, in other Wayland compositors, this also happens? I don't know if this is general X vs Wayland environment thing or if it is something specifically that we do that we shouldn't, or don't do that we should. Could you check it out in sway maybe?

Thanks for the feedback!

Graeme22:
> That's funny. Do you know if, in other Wayland compositors, this also happens? I don't know if this is general X vs Wayland environment thing or if it is something specifically that we do that we shouldn't, or don't do that we should. Could you check it out in sway maybe?

Looks like it may just be a Wayland thing, same thing happens in Sway. It's an easy fix of course, I can just up the font size in the config. So sounds like not something that needs to be fixed.

Also, the install process would be greatly improved if we could get `python-xkbcommon` published to PyPi and the dependencies straightened out for your fork. I believe I had to install `pywlroots` and `wlroots` separately, install `python-xkbcommon` from source, then install your fork. I imagine this is already on the todo list but thought I'd mention it.

m-col:
I absolutely agree, we need to trickle these projects to packagers and set all that up.

> Looks like it may just be a Wayland thing

OK cool, thanks for checking.

Issue #2473, opened 2021-05-19
How to get padding around whole widget?


clawbhaiya:
I want to get padding around the module as a whole.
Not only from other modules but also from top and bottom, so that modules appears to be kinda floating.
Is there anything from which I can achieve this?

I didn't find any documentation on this. That's why I opened an issue.

elParaguayo:
I'm not 100% certain on the end result you hope to get so my comments below could be way off...

Firstly, if you set the background colour for a widget, it paints the whole background. There's (currently) no way to set an area of the widget background that isn't painted. The bar does have a `margin` parameter so this could give you something at the top and bottom but my guess is this would look a bit odd.

As for having the widgets "floating", that seems to suggest you want a transparent bar. As far as I'm aware, that doesn't work at the moment as setting the opacity of the bar also impacts the widgets themselves i.e. a transparent bar = transparent widgets. There are existing open issues about adding the ability to have a transparent bar.

ayush7788:
![IMG_20210520_152433](https://user-images.githubusercontent.com/57079688/118959051-96633000-b97f-11eb-8582-20a6342cc909.png)
I am talking of something like this. See the date module in bar. Is there any way to achieve it in QTile?

ramnes:
I don't think we have anything at the moment to do that. Have a look at @Barbarossa93 config, that apparently used eww to that end: https://github.com/Barbarossa93/Genome

elParaguayo:
Key words being "at the moment"!

I was doing some playing around with widget backgrounds (e.g. linear gradients) and it wouldn't be too hard to do custom backgrounds like the above image but I'm reluctant to add code that makes widgets more complicated unless there's demand for it.

Barbarossa93:
Indeed, I was able to get that effect by overlaying eww bars on top of the qtile bar. A hacky, cheap solution; but aesthetically it works. I was trying to hack a border into the qtile bar itself but I couldn't figure it out and pretty much gave up on it.

@elParaguayo, that sounds like an interesting/complicated way to go about it, but possibly the better way as it can allow for further customization beyond a simple border!

elParaguayo:
OK - I may try a quick proof of concept. Was thinking of some sort of `decoration` object that draws a layer between the background and the widget contents.

elParaguayo:
OK - playing around with this a bit. Currently, can do things like this:
[...]

PR #2489, opened 2021-05-29
Add widget decorations

elParaguayo:
Adding background decorations to widgets as per #2473. This commit adds a base `_Decoration` class and two implementations. Method to draw decorations is added to all widgets and documentation also updated.

This is just eye-candy, no functional improvements. Pushing this for discussion.

Probably needs a test...

This is what this PR does:

Combining this with transparency in #2513 and you can do this:

### DIFF
--- a/docs/manual/ref/widgets.rst
+++ b/docs/manual/ref/widgets.rst
@@ -7,3 +7,53 @@ Built-in Widgets
 .. qtile_module:: libqtile.widget
     :baseclass: libqtile.widget.base._Widget
     :no-commands:
+
+Widget decorations
+==================
+
+Widgets can also be configured with ``Decoration`` objects to add additional
+eye-candy.
+
+The classes need to be imported into the user's config and can then be added to
+widgets as follows:
+
+::
+
+    from libqtile.widget.decorations import RectDecoration
+
+
+
+    decoration_config = {
+        "decorations": [
+            BorderDecoration(
+                padding=2,
+                padding_y=5,
+                filled=True,
+                dec_colour="#005555"
+            )
+        ],
+        "padding": 6, 
+    }
+
+
+    screens = [
+        Screen(
+            bottom=bar.Bar(
+                [
+                    widget.Prompt(),
+                    widget.Sep(),
+                    widget.WindowName(),
+                    widget.Sep(),
+                    widget.Clock(format='%H:%M:%S %d.%m.%Y', **decoration_config),
+                    widget.QuickExit(**decoration_config)
+                ],
+            24,
+            ),
+        )
+    ]
+
+The following decorations are available:
+
+.. qtile_module:: libqtile.widget.decorations
+    :baseclass: libqtile.widget.decorations._Decoration
+    :no-commands:
--- a/docs/manual/ref/widgets.rst
+++ b/docs/manual/ref/widgets.rst
@@ -21,15 +21,27 @@ widgets as follows:
 
     from libqtile.widget.decorations import RectDecoration
 
+    decorated_widget = widget.WindowName(
+        decorations=[
+            RectDecoration(
+                colour="#660066",
+                radius=2,
+                filled=True
+            )
+        ]
+    )
 
+Alternatively, the same instance of the decoration can be passed to multiple widgets:
+
+::
+
+    from libqtile.widget.decorations import BorderDecoration
 
     decoration_config = {
         "decorations": [
             BorderDecoration(
-                padding=2,
-                padding_y=5,
-                filled=True,
-                dec_colour="#005555"
+                colour="#660000",
+                border_width=[0, 0, 2, 0]
             )
         ],
         "padding": 6, 
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -193,6 +193,15 @@ class _Widget(CommandObject, configurable.Configurable):
         self.bar = bar
         self.drawer = bar.window.create_drawer(self.bar.width, self.bar.height)
         if not self.configured:
+
+            # Give each widget a copy of the decoration objects
+            temp_decs = []
+            for i, dec in enumerate(self.decorations):
+                cloned_dec = dec.clone()
+                cloned_dec._configure(self)
+                temp_decs.append(cloned_dec)
+            self.decorations = temp_decs
+
             self.qtile.call_soon(self.timer_setup)
             self.qtile.call_soon(asyncio.create_task, self._config_async())
 
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -116,6 +116,7 @@ class _Widget(CommandObject, configurable.Configurable):
     defaults = [
         ("background", None, "Widget background color"),
         ("mouse_callbacks", {}, "Dict of mouse button press callback functions."),
+        ("decorations", [], "Decorations for widgets")
     ]  # type: List[Tuple[str, Any, str]]
 
     def __init__(self, length, **config):
@@ -269,6 +270,15 @@ class _Widget(CommandObject, configurable.Configurable):
             changed. If it has, you must call bar.draw instead.
         """
         raise NotImplementedError
+
+    def draw_decorations(self):
+        """
+            Method that draws decorations underneath widget. This should be
+            called in a widget's `draw` method immediately after painting the
+            background but before the widget's contents.
+        """
+        for decoration in self.decorations:
+            decoration.draw(self)
 
     def calculate_length(self):
         """
@@ -427,6 +437,8 @@ class _TextBox(_Widget):
         if not self.can_draw():
             return
         self.drawer.clear(self.background or self.bar.background)
+        self.draw_decorations()
+
         self.layout.draw(
             self.actual_padding or 0,
             int(self.bar.height / 2.0 - self.layout.height / 2.0) + 1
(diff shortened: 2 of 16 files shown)


libqtile/widget/base.py
```diff
@@ -279,6 +280,15 @@ def draw(self):
         """
         raise NotImplementedError
 
+    def draw_decorations(self):
```
elParaguayo, line 283: Should this also be in `Bar`? Some users have asked for single borders on the bar and this would be a way to achieve that.

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,226 @@
...
+        _Decoration.draw(self, parent)
+
+        top, bottom, left, right = self.borders
+
+        self.drawer.set_source_rgb(self.colour)
+
+        if top:
+            offset = top // 2
+            self._draw_border(
+                offset + self.padding_x,
```
elParaguayo, line 185: Think my maths is right but still getting slight gaps between widgets.

docs/manual/ref/widgets.rst
```diff
@@ -7,3 +7,53 @@ Built-in Widgets
...
+                ],
+            24,
+            ),
+        )
+    ]
+
+The following decorations are available:
+
+.. qtile_module:: libqtile.widget.decorations
+    :baseclass: libqtile.widget.decorations._Decoration
```
elParaguayo, line 58: I'd like to hide the base class from the docs but not sure how to do it.

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,227 @@
...
+        multi-coloured effects.
+
+        Rectangles can be drawn as just the the outline or filled. Curved corners
+        can be obtained by setting the ``radius`` parameter.
+    """
+    defaults = [
+        ("filled", False, "Whether to fill shape"),
+        ("radius", 4, "Radius for corners (0 for square)"),
+        ("dec_colour", "#000000", "Colour for decoration"),
+        ("linewidth", 2, "Line width for decoration")
```
m-col, line 75: `line_width`? or are you trying to be consistent with another place where there is `linewidth`?

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,227 @@
...
+    @property
+    def drawer(self):
+        return self.parent.drawer
+
+    @property
+    def ctx(self):
+        return self.parent.drawer.ctx
+
+    def draw(self, parent):
+        self.parent = parent
```
m-col, line 58: Is it necessary to set the parent each time? Could it change?

I wonder if the base class shouldn't have a `_configure` method, called by the `Bar` along with the widgets, to give the decoration the parent widget. Then the draw method here can be an abstractmethod, ensuring nobody instantiates this base class directly. And also means child classes don't need to call `_Decoration.draw`.

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,227 @@
...
+        box_height = self.height - 2 * self.padding_y
+        box_width = self.width - 2 * self.padding_x
+
+        self.drawer.set_source_rgb(self.dec_colour)
+
+        degrees = math.pi / 180.0
+
+        self.ctx.new_sub_path()
+
+        delta = self.radius + self.linewidth / 2
```
m-col, line 94: Could this be optimised in the case that radius is 0, and just draw a rectangle directly?

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,227 @@
...
+            self.ctx.set_line_width(self.linewidth)
+            self.ctx.stroke()
+
+
+class BorderDecoration(_Decoration):
+    """
+        Widget decoration that draws a straight line on the widget border.
+        Padding can be used to adjust the position of the border further.
+
+        Only one colour can be set but decorations can be layered to achieve
```
m-col, line 141: nice, another score for "colour" over "color", our most inconsistently used word!

libqtile/widget/decorations.py
```diff
@@ -0,0 +1,227 @@
...
+        multi-coloured effects.
+    """
+    defaults = [
+        ("colour", "#000000", "Border colour"),
+        (
+            "border_width",
+            2,
+            "Border width. Single number is all edges, or [top/bottom, left/right] or [top, bottom, left, right]"
```
m-col, line 149: Upon first read I got a bit confused. Perhaps to be consistent with other things we can specify in all four sides (like Bar margins) we should stick to either a single number or all 4? Otherwise people might try doing the `[top/bottom, left/right]` version where it isn't supported. Along the same lines, can this be changed to `[N, E, S, W]`, to stay consistent with similar options around the codebase?

libqtile/widget/decorations.py
m-col: Could it be due to the `//` operators and a rounding error?

m-col (changes requested):
This is looking really nice!

I've added a couple comments inline but overall it's looking great. Please could you add type annotations to method arguments and class attributes? We should start doing this as standard.

libqtile/widget/decorations.py
elParaguayo: Yeah, tried that. Will do some more digging.

docs/manual/ref/widgets.rst
```diff
@@ -7,3 +7,53 @@ Built-in Widgets
...
+Widgets can also be configured with ``Decoration`` objects to add additional
+eye-candy.
+
+The classes need to be imported into the user's config and can then be added to
+widgets as follows:
+
+::
+
+    from libqtile.widget.decorations import RectDecoration
+
```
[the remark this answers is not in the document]
elParaguayo: All good points. Will fix.

libqtile/widget/decorations.py
elParaguayo: I think `drawer` and `Sep` both use `linewidth` but `graph` uses `line_width`. I'll change it as I think the latter is more readable.

libqtile/widget/decorations.py
elParaguayo: So... the way it works currently is that the decoration is instantiated once in the config but can be attached to multiple widgets (with the benefit of preserving the configuration of the decoration). However, at that point, it's not aware of any of the widgets' properties as they haven't been configured. When the widgets draw themselves, the decoration needs to know how big the parent widget is and so needs a reference to that parent. This means the same `draw` method is called for different widgets, hence the need to set the `parent`.

I thought about having a `_configure` method but I think that means each widget needs a separate instance of the decoration.

libqtile/widget/decorations.py
elParaguayo: Yes.

libqtile/widget/decorations.py
elParaguayo: Ha. Despite my username, I am British. `color` just feels dirty to me!

libqtile/widget/decorations.py
m-col: +1 for line_width where possible, hopefully no user-facing configuration options use linewidth!

libqtile/widget/decorations.py
elParaguayo: Yeah. I did a bad job of describing this. Will adjust as you suggest.

elParaguayo:
> This is looking really nice!
> 
> I've added a couple comments inline but overall it's looking great. Please could you add type annotations to method arguments and class attributes? We should start doing this as standard.

Sorry, that's me being lazy. Not familiar with type annotations. Should make an effort to learn them.

libqtile/widget/decorations.py
m-col: From a config perspective, these aren't mutually exclusive ideas; a `_Decoration.copy()` could be used internally to give each widget its own instance so that the decorations each have exactly one parent and configs can still pass one instance to widgets.

So I guess the decision then comes down to performance or code clarity... not really sure what's the best there so happy for you to choose whichever.

m-col:
No worries at all, I've only in the past few months gained an appreciation of
them. They are very good for spotting hidden bugs and working out what code that
was written years ago does :)

libqtile/widget/base.py
elParaguayo: I think the answer to this is that it can't be done yet. The reason being is that all widgets (including spacers) are drawn over the bar so, without transparency, there's no point decorating the bar.

I do want to look at the bar transparency point separately...

m-col: I don't it's important to do it in this PR so don't sweat it.

elParaguayo: Sorry, yes, I meant I'm hoping to look at it in a different branch but there's a bit more of a learning curve on that one for me.

elParaguayo:
Added ability to round individual corners, meaning you can do effects like this:


WIll be even better when the transparency is done...

elParaguayo:
This will clash with #2675 so shouldn't be merged before that one.

Tokariew:
I think that docs should have example with layered decorations. for someone who don't look into implementation it can be confusing how to set them

elkrien:
I'm using RectDecorations with Qtile-extras and works really great without any problems. Would be nice to see it in Qtile.

elParaguayo:
Closing this as it's in `qtile-extras`.

PR #2490, opened 2021-05-30
Fix for checkupdates widget on Fedora & disqualify negative numbers o…

Alteus:
issue #2488 widget.CheckUpdates on Fedora: show the correct number of updates

On Fedora `dnf list updates -q` returns no output when there aren't updates available.
If there are updates, "available updates" text added on top of the package list.

This should fix that. And as a side effect, rules out negative update numbers on all distros.

closes #2488

PR #2492, opened 2021-05-30
Fix CheckUpdates widget test

elParaguayo:
PR #2490 fixed a bug with the update count for Fedora but the test was not updated at the same time. This PR fixes that.


m-col:
Cool thanks

PR #2496, opened 2021-06-01
Clear startup_once hooks once fired

m-col:
May as well clean up the subscriptions a bit


ramnes:
What's the rationale? I feel bad playing seeing `hook.subscriptions` played with directly, it doesn't seems right for a hook system. :)

m-col:
Over-optimisation no doubt :) I just wanted to get rid of a list that after
initial startup is never touched again, and it seemed overkill to implement a
`hook.subscribe.first.startup_once` or something just for this one thing. This
is super minor so in the interest of less ugly code I will close.

PR #2493, opened 2021-06-01
Update BitcoinTicker widget to handle other cryptocurrencies 

Graeme22:
Very simple changes to allow the BitcoinTicker widget to become more generic and support a host of other cryptocurrencies.

Edit: Also fixes #2470 by allowing users to input their currency symbol instead of trying to guess it.

This *clearly* needs to be squashed.

PR #2497, opened 2021-06-01
Update Bitcoin ticker to handle other cryptocurrencies.

### LINKED ISSUE #2470: Bitcoin Ticker keeps showing locale currency sign
My locale is GB, so if I don't specify a currency, it will show BTC in £. But when I specify, say,"USD", it shows the USD **numbers** but still shows the "£" sign in front of it.

So let's say BTC is "$44,000" in USD and "£31,000" in GBP. If I don't specify, it will say "£31,000". But if I write "USD", it says "**£**44,000" (the USD value, but the GBP sign).

Graeme22:
Redo of #2493. Fixes #2470.

~~Still needs a test.~~


m-col:
Looks good, thanks for this. It might be good for us to let BitcoinTicker still work for the next release, logging a DeprecationWarning and mentioning `qtile migrate` and forwarding to CryptoTicker, as some users will not migrate (or not be aware of they can) and get broken.

ramnes:
> It might be good for us to let BitcoinTicker still work for the next release, logging a DeprecationWarning and mentioning `qtile migrate` and forwarding to CryptoTicker, as some users will not migrate (or not be aware of they can) and get broken.

Mhhh, I hesitated to say this in the previous PR but ended up thinking that the main point of `qtile migrate` is that we could make breaking changes without a deprecation warning. Do we really want to maintain both migrations and a deprecated stuff?

m-col:
I figure it's pretty easy to keep the deprecation warning while encouraging
migrate, and my worry is that `qtile migrate` is still quite young so it's
possible many users don't know of its existence yet. But I'm happy either way,
they'll just drop into the default config and investigate anyway!

ramnes:
@tych0 as the author of qtile-migrate, what are your thoughts?

libqtile/widget/crypto_ticker.py
```diff
@@ -31,40 +32,56 @@
...
-    QUERY_URL = "https://api.coinbase.com/v2/prices/spot?currency=%s"
+    QUERY_URL = "https://api.coinbase.com/v2/prices/{}-{}/spot"
 
     orientations = base.ORIENTATION_HORIZONTAL
 
     defaults = [
         ('currency', _DEFAULT_CURRENCY.strip(),
-            'The currency the value that bitcoin is displayed in'),
+            'The baseline currency that the value of the crypto is displayed in.'),
+        ('symbol', '$',
```
tych0, line 60: Can we derive this from the locale too, as we do the currency? seems like yes,

```
>>> locale.localeconv()
{'int_curr_symbol': 'USD ', 'currency_symbol': '$', 'mon_decimal_point': '.', 'mon_thousands_sep': ',', 'mon_grouping': [3, 3, 0], 'positive_sign': '', 'negative_sign': '-', 'int_frac_digits': 2, 'frac_digits': 2, 'p_cs_precedes': 1, 'p_sep_by_space': 0, 'n_cs_precedes': 1, 'n_sep_by_space': 0, 'p_sign_posn': 1, 'n_sign_posn': 1, 'decimal_point': '.', 'thousands_sep': ',', 'grouping': [3, 3, 0]}```

tych0:
Migration stuff lgtm (though a test would be good), one small comment about currency symbol resolution.

Graeme22:
Should be good. Whoever merges this may want to squash it.

tych0:
You can also squash it yourself and `git push -f`, which may make it less work to merge since it's just a button press vs. a bunch of the same git commands :)

Graeme22:
@elParaguayo the tests are somewhat simplistic, so feel free to have a go at updating them or adding new ones.

tych0:
I fixed up the commit message here, once CI passes it LGTM, thanks!

Graeme22:
> I fixed up the commit message here, once CI passes it LGTM, thanks!

Should be good to go. @elParaguayo feel free to add tests after this is merged.

PR #2498, opened 2021-06-01
Don't warp cursor when changing group via GroupBox

### LINKED ISSUE #2474: feature request: cursor_warp only on keyboard commands
# Issue description

Make all actions of `cursor_warp` only occur when triggered from keyboard commands. ie. not from mouse actions.

The idea being that if you change focus using your mouse, you probably don't want your mouse to teleport across the screen. However, when you're changing focus using keyboard commands, it is probably desirable to have the cursor warp locations. i3 worked in a very similar way.

Note this behavior somewhat already works when using the mouse to change focus on windows in the same group:
qtile_focus_change_20210519

Excuse my ignorance on how exactly hardware inputs are handled on linux, but I think a way to do this would be to check/create a "log" of all keyboard and mouse inputs into the machine. When changing window focus, qtile could check the log to see whether the keyboard was just used, or the mouse and then choose whether to warp the cursor.

Some specific examples:
- Changing Group using the mouse by clicking the bar
    - The cursor movement is quite jarring as you just clicked something and then your mouse has suddenly disappeared
- Issues with Chromium refocusing too aggressively
    - See #990 
    - I believe this would be fixed by the proposed feature (though I'm not entirely sure)
- Closing a floating window on top of a different window
    - The change in focus after closing the floating window (with mouse) causes the mouse to warp. Again, fairly jarring experience

# Qtile version
0.17.0 (git hash not given, maybe should add to `qtile --version` output)

# Configuration
`cursor_warp = True`

m-col:
Partially fixes #2474. Disables cursor warp when the user clicks on the
groupbox widget the change the group by providing a warp=True kwarg to
some of the Screen and Group methods for group changing.

Also group._set_screen -> group.set_screen because it isn't being used
as a private method.

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -444,11 +444,9 @@ class Window(base.Window, HasListeners):
                 self.group.mark_floating(self, True)
             hook.fire('float_change')
 
-    def cmd_focus(self, warp=None):
+    def cmd_focus(self, warp: bool = True) -> None:
         """Focuses the window."""
-        if warp is None:
-            warp = self.qtile.config.cursor_warp
-        self.focus(warp=warp)
+        self.focus(warp)
 
     def cmd_info(self) -> Dict:
         """Return a dictionary of info."""
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -252,7 +252,7 @@ class Window(_Window, metaclass=ABCMeta):
         """Paint the window borders with the given color and width"""
 
     @abstractmethod
-    def cmd_focus(self, warp: Optional[bool] = None) -> None:
+    def cmd_focus(self, warp: bool = True) -> None:
         """Focuses the window."""
 
     @abstractmethod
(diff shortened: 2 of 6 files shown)


libqtile/backend/wayland/window.py
```diff
@@ -444,11 +444,9 @@ def _reconfigure_floating(self, x, y, w, h, new_float_state=FloatStates.FLOATING
                 self.group.mark_floating(self, True)
             hook.fire('float_change')
 
-    def cmd_focus(self, warp=None):
+    def cmd_focus(self, warp: bool = True) -> None:
         """Focuses the window."""
-        if warp is None:
-            warp = self.qtile.config.cursor_warp
-        self.focus(warp=warp)
```
ramnes, line 451: I'm not sure to understand this diff and the next one. Isn't setting warp to `True` by default and removing the `config.cursor_warp` check going to make cursor warping enabled for everyone?

m-col: Nope - `warp` isn't so much "warp" as it is "now would be a good time to warp if `cursor_warp` is enabled". See line 922 - `if warp and self.qtile.config.cursor_warp` - warping will _never_ happen if `config.cursor_warp` is False. The current version is checking `config.cursor_warp` two times.

PR #2500, opened 2021-06-02
Added Gentoo support in CheckUpdates Widget

ZaheenJ:
Pretty much the title.
The output of `emerge -upvDN world` is 6 lines when there is no update:
```

These are the packages that would be merged, in order:

Calculating dependencies  . ....... done!

Total: 0 packages, Size of downloads: 0 KiB
```
The output is one line longer per update:
```

These are the packages that would be merged, in order:

Calculating dependencies  .... done!
[ebuild     U  ] dev-lang/python-3.9.5_p2:3.9::gentoo [3.9.5_p1:3.9::gentoo] USE="bluetooth gdbm ipv6 ncurses readline sqlite ssl xml -build -examples -hardened -test -tk -verify-sig -wininst" 15 KiB
[ebuild     U  ] dev-libs/libinput-1.18.0:0/10::gentoo [1.17.3:0/10::gentoo] USE="-doc -test" INPUT_DEVICES="-wacom" 603 KiB

Total: 2 packages (2 upgrades), Size of downloads: 617 KiB
```

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -54,7 +54,7 @@ class CheckUpdates(base.ThreadPoolText):
                          "Arch_Sup": ("pacman -Sup", 1),
                          "Arch_yay": ("yay -Qu", 0),
                          "Debian": ("apt-show-versions -u -b", 0),
-                         "Gentoo" : ("eix -u# --world", 0),
+                         "Gentoo_eix" : ("eix -u# --world", 0),
                          "Ubuntu": ("aptitude search ~U", 0),
                          "Fedora": ("dnf list updates -q", 1),
                          "FreeBSD": ("pkg_version -I -l '<'", 0),
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -54,7 +54,7 @@ class CheckUpdates(base.ThreadPoolText):
                          "Arch_Sup": ("pacman -Sup", 1),
                          "Arch_yay": ("yay -Qu", 0),
                          "Debian": ("apt-show-versions -u -b", 0),
-                         "Gentoo_eix" : ("EIX_LIMIT=0 eix -u# --world", 0),
+                         "Gentoo_eix": ("EIX_LIMIT=0 eix -u# --world", 0),
                          "Ubuntu": ("aptitude search ~U", 0),
                          "Fedora": ("dnf list updates -q", 1),
                          "FreeBSD": ("pkg_version -I -l '<'", 0),
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -54,7 +54,7 @@ class CheckUpdates(base.ThreadPoolText):
                          "Arch_Sup": ("pacman -Sup", 1),
                          "Arch_yay": ("yay -Qu", 0),
                          "Debian": ("apt-show-versions -u -b", 0),
-                         "Gentoo_eix" : ("eix -u# --world", 0),
+                         "Gentoo_eix" : ("EIX_LIMIT=0 eix -u# --world", 0),
                          "Ubuntu": ("aptitude search ~U", 0),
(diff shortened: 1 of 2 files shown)


tych0:
Is there perhaps a quiet flag that can get rid of that output as well? But in any case, lgtm.

ZaheenJ:
There infact is! I just always used a similar command for updating my system and when I modified it to work with qtile I never thought about it, thanks.

Etn40ff:
I am not sure on how things works in other distros but on gentoo, if you do not
perform a sync of the local package database first, say by `eix-sync`, you do not get
meaningful informations out of `emerge`. Also the command is quite
computationally expensive, not the kind of task I would run every 60 seconds.

ZaheenJ:
I actually just have a cronjob that runs `emaint sync -a` that runs daily and I also have the command run only once per hour. I believe that this is an equivalent to `pacman -Qu` for Arch, which also doesn't sync the local package database. I don't know anything about how computationally expensive this is although I have noticed that it takes a bit for it to run. I guess if this doesn't fit the default 60 second timer then it should be a custom_command.

ramnes:
Stuff like `The following packages are causing rebuilds: ...` will break this. :)

And yes, the dependency resolution of Portage is CPU-heavy and can take more than 60 seconds. I think it would make sense to make the default update interval different for Gentoo specifically.

Etn40ff:
A possible alternative: `eix -uc#`

ramnes:
Or `eix -u# --world` then? LGTM :)

Etn40ff:
wait, I thought this was going to be modified.

ZaheenJ:
Actually I think it might be good to change the name of the "distro" to Gentoo_eix, similar to Arch_yay to make sure that anyone running gentoo knows they need app-portage/eix installed. @Etn40ff I was just installing eix on my system and making sure it worked.

Etn40ff:
@ZaheenJ sorry I think I pulled the trigger too early on you :)

ZaheenJ:
@Etn40ff Hah no worries. Thanks for suggesting the eix command, avoided having to make a separate default timer for Gentoo or anything like that. It also just seems so much faster than emerge I was just never aware of it, I'll start using this now.

libqtile/widget/check_updates.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, **config):
                          "Arch_Sup": ("pacman -Sup", 1),
                          "Arch_yay": ("yay -Qu", 0),
                          "Debian": ("apt-show-versions -u -b", 0),
+                         "Gentoo_eix" : ("eix -u# --world", 0),
```
ramnes, line 57:
```suggestion
                         "Gentoo_eix" : ("EIX_LIMIT=0 eix -u# --world", 0),
```

ramnes: (capped to 50 otherwise)

libqtile/widget/check_updates.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, **config):
                          "Arch_Sup": ("pacman -Sup", 1),
                          "Arch_yay": ("yay -Qu", 0),
                          "Debian": ("apt-show-versions -u -b", 0),
+                         "Gentoo_eix" : ("EIX_LIMIT=0 eix -u# --world", 0),
```
ramnes, line 57: Let's make CI happy again

```suggestion
                         "Gentoo_eix": ("EIX_LIMIT=0 eix -u# --world", 0),
```

ZaheenJ:
Is it okay if I also add support for [paru](https://github.com/Morganamilo/paru) in this same pull request? I've added Arch_paru and Arch_paru_Sup which do the same things as Arch and Arch_Sup respectively. I also have no idea how to fix the mypy commands failing for the CI (I have never used one before and actually don't know how to code).

ramnes:
The mypy errors are unrelated. Let's merge this first!

Issue #2483, opened 2021-05-23
Can't drag scroll bar in Code editor


elParaguayo:
Just noticed this so posting now but will try to do some more debugging...

# Issue description

I can't drag the scroll bar in VS Code (using `code` package on Arch). The bar changes colour when mousing over it but click and drag doesn;t work

# Qtile version

0.17.0 

# Stack traces

Debug log:
```
2021-05-30 21:26:13,266 DEBUG libqtile manager.py:cmd_debug():L918 Switching to DEBUG threshold
2021-05-30 21:26:13,268 DEBUG libqtile ipc.py:_server_callback():L201 Sending result on receive EOF
2021-05-30 21:26:13,268 DEBUG libqtile ipc.py:_server_callback():L203 Closing connection on receive EOF
2021-05-30 21:26:18,094 DEBUG libqtile core.py:_xpoll():L269 EnterNotify
2021-05-30 21:26:18,096 DEBUG libqtile core.py:_xpoll():L272 Handling: EnterNotify
2021-05-30 21:26:18,096 DEBUG libqtile hook.py:fire():L381 Internal event: client_mouse_enter((Window('decorations.py - qtile - Code - OSS'),), {})
2021-05-30 21:26:19,187 DEBUG libqtile core.py:_xpoll():L269 EnterNotify
2021-05-30 21:26:19,188 INFO libqtile core.py:_get_target_chain():L348 Unhandled event: EnterNotify
2021-05-30 21:26:19,189 DEBUG libqtile core.py:_xpoll():L269 ButtonPress
2021-05-30 21:26:19,190 DEBUG libqtile core.py:_xpoll():L272 Handling: ButtonPress
2021-05-30 21:26:19,191 INFO libqtile manager.py:process_button_click():L743 Ignoring unknown button: 1
2021-05-30 21:26:19,192 DEBUG libqtile interface.py:call():L310 Command: function((<function Qtile.load_config.<locals>.noop at 0x7f5abe1e8700>,), {})
2021-05-30 21:26:19,204 DEBUG libqtile hook.py:fire():L381 Internal event: client_focus((Window('decorations.py - qtile - Code - OSS'),), {})
2021-05-30 21:26:19,211 DEBUG libqtile hook.py:fire():L381 Internal event: client_focus((Window('decorations.py - qtile - Code - OSS'),), {})
[...]

PR #2503, opened 2021-06-02
x11: Always focus upon button 1 click in window

m-col:
We need to grab button 1 so that we can focus the window below the
pointer. Currently we do this by appending a `Click` to `config.mouse`
internally and using the root window's button grab (for the `Click`) to
focus the window below the pointer upon button event. This causes
problems with where the events are propagated. This commit replaces this
mechanism to instead grab button 1 manually when a window is managed,
using that window as the grabbing window, rather than the root window.
EnterNotify events are created upon such clicks and their `mode`
attribute tells us which buttons were involved, so we can check for
`e.mode == 1` and focus the window if needed. Simple!

(possibly) related issues: #2404, #2483, #1809, #1824, #1792.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -498,6 +498,20 @@ class _Window:
         }
         self.update_hints()
 
+        # Grab button 1 to focus upon click
+        for amask in self.qtile.core._auto_modmasks():
+            self.qtile.core.conn.conn.core.GrabButton(
+                True,
+                self.window.wid,
+                EventMask.ButtonPress,
+                xcffib.xproto.GrabMode.Sync,
+                xcffib.xproto.GrabMode.Async,
+                xcffib.xproto.Atom._None,
+                xcffib.xproto.Atom._None,
+                1,
+                amask,
+            )
+
     x = property(fset=_geometry_setter("x"), fget=_geometry_getter("x"))
     y = property(fset=_geometry_setter("y"), fget=_geometry_getter("y"))
     width = property(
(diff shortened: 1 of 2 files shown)


m-col:
The test failure is because clicking needs to also raise the window to the front of the stack if bring_front_click option is set, so I'll add that in if the fix works for the issues, because if it doesn't then... back to the drawing board.

elParaguayo:
@m-col great work - fingers crossed but this seems to fix the issues I've been having. Hopefully others will report similar results.

ramnes:
Doesn't seem to work for me? It removes the bugs related to the `Click` object, but reintroduces #125. I can't see any different behavior than if we just remove the `Click`.

elParaguayo:
Sorry. Should have clarified that it fixes the issues that I was experiencing (click-through in Brave and not dragging scrollbar in Code). 

I haven't tested against other issues.

m-col:
> reintroduces #125.

That is weird, it works perfectly for me: 

https://user-images.githubusercontent.com/44536113/120613670-58b9d900-c44e-11eb-85d0-da7b11781319.mp4

m-col:
Oh but that is only with xterm, it doesn't work with firefox!

m-col:
OK, how about now? :)

ramnes:
Yay!

ramnes:
IIRC merging this would basically make the `focus` argument of `Click`'s constructor (and all the logic behind it) irrelevant - let's remove it if I'm not mistaken?

m-col:
Yeah I guess that makes sense... And I guess the whole focus-by-click behaviour
would then be implemented entirely within the backend? Is there any use-case for
`focus="after"` that we'd be removing?

ramnes:
I'm not aware of any, and a quick search on GitHub doesn't seem to return anything else than the code here.

m-col:
@ramnes could you check if the `or e.mode == 1` added to line 1468 is necessary? I suspect it isn't now that it's using ButtonPress instead of EnterNotify

ramnes:
Closed in favor of #2506

PR #2504, opened 2021-06-02
Add get_wm_type as base Window method

m-col:
This lets us keep the wm_type Match entries in the default float rules,
which makes sense, while making these matches a no-op under other
backends.


m-col:
@flacjacket does this fix the issue you came across?

flacjacket:
Just pulled this on master and can confirm this addresses the problem I was seeing, thanks!

m-col:
ok great!

PR #2505, opened 2021-06-02
WL: add wlr_primary_selection_v1 manager

m-col:
This exposes wlroots' implementation of a primary selection clipboard,
used by clients (such as some terminals) to set a basic text-only
clipboard.

Requires pywlroots 0.13.3


ramnes:
Nothing more to do? Neat. :)

m-col:
Yeah, it's pretty nice, wlroots is a great library.

PR #2506, opened 2021-06-03
Always focus upon button 1 click in window

m-col:
We need to grab button 1 so that we can focus the window below the
pointer. Currently we do this by appending a `Click` to `config.mouse`
internally and using the root window's button grab (for the `Click`) to
focus the window below the pointer upon button event (in x11). This
causes problems with where the events are propagated. This commit
replaces this mechanism in the x11 backend to instead grab button 1
manually when a window is managed, using that window as the grabbing
window, rather than the root window. Then the window's own ButtonPress
events are used to focus the window.

This replaces the `focus` option in `Mouse`/`Click`/`Drag` config
classes, which are no longer used. The backend is responsible for
focussing windows upon clicks.

(possibly) related issues: #2404, #2483, #1809, #1824, #1792.

--

This is a more fleshed-out replacement for https://github.com/qtile/qtile/pull/2503, though I wanted to keep both around in case for some annoying reason the other one fixes things but this doesn't.

Please could people test this one too, to check it's all good too?

I'm probably going to have to update the bring_to_front_click tests, and im gonna append a commit for migrating `Click` and `Drag`s.


m-col:
Tests seem happy so I don't have to do anything there. Not sure a migration is necessary - those config classes don't actually define a `focus` argument, they just accept `**kwargs`, and bowler says this doesn't make sense to migrate. On the plus side, there is no config breakage. If people are aware of `migrate` then they presumably read the changelog, so they can read the change there.

ramnes:
Currently smoke testing the branch, looking good to me so far. :)

ramnes:
There's a bug when you click in the groupbox. Everything stays grabbed and thus you can't click anything nor use any keyboard binding.

m-col:
Possibly because `Internal` also has a `handle_ButtonPress` and it doesn't
replay the event again (because previously it was using the root window's grab).
I've just stuck a replay events line at the end of that method. Pls could you
try that?

ramnes:
> I've just stuck a replay events line at the end of that method. Pls could you try that?

Did you forget to push?

m-col:
Lol, yes.

ramnes:
Now it's impossible to switch to another group by clicking on the widget, it switches and instantly come back to the previous grroup (because it clicks twice on the same group twice and gets toggled?)

dbistriceanu:
It looks like now when I click on a different item in the TaskList widget it "minimizes/maximizes" that window instead of just focusing it, as if it would maybe register double-clicks? Unless that's just me somehow and no one else can replicate the issue?


https://user-images.githubusercontent.com/1913782/120821212-efc18680-c55d-11eb-9e00-99b62a77f0f5.mp4

ramnes:
Double click confirmed. :)

m-col:
Ah my bad -- we shouldn't be replaying those events. AllowEvents should just be unfreezing with `AsyncPointer`. Try this change:
```diff
diff --git a/libqtile/backend/x11/window.py b/libqtile/backend/x11/window.py
index caee5026..37ae0cc4 100644
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1051,7 +1051,7 @@ class Internal(_Window, base.Internal):
 
     def handle_ButtonPress(self, e):  # noqa: N802
         self.process_button_click(e.event_x, e.event_y, e.detail)
-        self.qtile.core.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayPointer, e.time)
+        self.qtile.core.conn.conn.core.AllowEvents(xcffib.xproto.Allow.AsyncPointer, e.time)
 
     def handle_ButtonRelease(self, e):  # noqa: N802
         self.process_button_release(e.event_x, e.event_y, e.detail)
```

dbistriceanu:
Changing windows and groups by clicking on those widgets seems fine for me now,

dbistriceanu:
Hmm... click seems broken on Systray icons now :(

m-col:
Alrighty, the button grabbing isn't necessary on `Internal` windows so I've removed it from those and their behaviour is unchanged. Anyone happy to test the updated version?

dbistriceanu:
So far so good. 🤞🏻

libqtile/backend/x11/window.py
```diff
@@ -1077,6 +1077,20 @@ def __init__(self, win, qtile, screen,
...
+                self.window.wid,
+                EventMask.ButtonPress,
+                xcffib.xproto.GrabMode.Sync,
+                xcffib.xproto.GrabMode.Async,
+                xcffib.xproto.Atom._None,
+                xcffib.xproto.Atom._None,
+                1,
+                amask,
+            )
+
```
ramnes, line 1093: Why not move this into `_Window.__init__` rather than copy paste it?

libqtile/backend/wayland/core.py
```diff
@@ -492,7 +493,7 @@ def focus_window(self, win: window.WindowType, surface: Surface = None):
             win.surface.set_activated(True)
         self.seat.keyboard_notify_enter(surface, self.seat.keyboard)
 
-    def focus_by_click(self, event) -> None:
+    def _focus_by_click(self) -> None:
```
ramnes, line 496: Why the rename?

libqtile/backend/x11/window.py
m-col: Because we don't want it on `Internal` windows

libqtile/backend/wayland/core.py
m-col: It used to be called by `Qtile` but now it's only called by `Core`, so it's private.

libqtile/backend/x11/window.py
ramnes: Wouldn't it make more sense if internal windows worked the same way than the other windows? Do you know why it's not the case at the moment?

libqtile/backend/x11/core.py
```diff
@@ -612,6 +605,10 @@ def handle_ButtonPress(self, event) -> None:  # noqa: N802
         state = event.state
         state |= self._numlock_mask
         state &= self._valid_mask
+
+        if not event.child:  # The client's handle_ButtonPress will focus it
+            self.focus_by_click(event)
+
```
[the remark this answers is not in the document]
m-col: Is there a benefit to do so? The implementation is backend-specific, e.g. the Wayland one doesn't pass around an `event`, and the X one we only want to call `focus_by_click` when it was not on a window, so that it can focus the root window instead. That last bit is what the comment is refering to: a click on a window will result on both `Core.handle_ButtonPress` and `Window.handle_ButtonPress` being called, so we only want the latter to focus that window. Hence the `if not event.child`. I can clarify that comment if needed!

libqtile/backend/x11/window.py
m-col: They mostly do, it's just that this new button grabbing is _only_ required so that we can focus the window and pass the event down to the window cleanly, but we don't want to focus internal windows so don't need this implemented on those windows. The cost to doing so is some hackery involving the xembed stuff around the Systray, which seems unnecessary given that this solution works anyway.

libqtile/backend/wayland/core.py
ramnes: Yeah but then why not rename the X11 one? :D

libqtile/backend/x11/core.py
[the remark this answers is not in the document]
ramnes: Oh right, I didn't notice the different signature. :)

ramnes:
Such a relief, thanks!

ramnes:
Alas it didn't fix #1792. :(

neworld:
Updated Spotify works fine even with the current 0.17.0.

ramnes:
@neworld Have a look at the video, it's not about clicking on the main menu (which had a problem as well previously, and is fixed  now), but configuration menus in the settings. And this still doesn't work on my side.

neworld:
Sorry, my bad. Alas, spotify settings are not working for me too.

PR #2510, opened 2021-06-05
Move window __repr__ methods to base classes

m-col:
This also slghtly changes the string returned by `Internal.__repr__`,
because previously it used `self.name` but `Internal` windows don't have
names.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -204,6 +204,10 @@ class _Window(CommandObject, metaclass=ABCMeta):
 
 
 class Window(_Window, metaclass=ABCMeta):
+    """A regular Window belonging to a client."""
+    def __repr__(self):
+        return "Window(name=%r, wid=%i)" % (self.name, self.wid)
+
     @property
     def floating(self) -> bool:
         """Whether this window is floating."""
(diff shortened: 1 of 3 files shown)


ramnes (approved):
LGTM, although I wonder if the wid will be useful to anyone

m-col:
It's been helpful for me while debugging as name often will be the same across
clients (like testing with a bunch of terminals).

On 210605 12:37, Guillaume Gelin wrote:
***@***.*** approved this pull request.

libqtile/backend/base.py
```diff
@@ -204,6 +204,10 @@ def info(self) -> Dict[str, Any]:
 
 
 class Window(_Window, metaclass=ABCMeta):
+    """A regular Window belonging to a client."""
+    def __repr__(self):
+        return "Window(%r, %i)" % (self.name, self.wid)
```
ramnes, line 209: Might be easier to read this way?

```suggestion
        return "Window(name=%r, wid=%i)" % (self.name, self.wid)
```

m-col: I like it. Updated.

PR #2511, opened 2021-06-05
Add Wayland support to KeyboardLayout widget

m-col:
This makes the KeyboardLayout widget support multiple backends, and adds
an implementation for the Wayland backend.

This requires the latest release of python-xkbcommon, 0.3, so this is
explicitly stated in setup.cfg and tox.ini.

--

I think most of the work on this is done. However, it only seems to work for me when Qtile (Wayland) is run within a running X or Wayland session inside a window, but not when run as a full session. Under X it all works as usual. I'm not sure if anyone has any experience with libxkbcommon but maybe they'd be able to take a look and debug?

cc @flacjacket

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -637,3 +637,12 @@ class Core(base.Core, wlrq.HasListeners):
 
         assert len(matched) == 1
         return matched[0]
+
+    def set_keymap(self, layout: Optional[str], options: Optional[str]) -> None:
+        """
+        Set the keymap for the current keyboard.
+        """
+        if self.keyboards:
+            self.keyboards[-1].set_keymap(layout, options)
+        else:
+            logger.warning("Could not set keymap: no keyboards set up.")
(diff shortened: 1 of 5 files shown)


Fabien-Bernard:
Maybe a clue :
I launch qtile with wayland backend via tty.
If I change the keyboard layout via the wigdet, the layout is not changed.
The log indicates during the core init that 2 keyboards were detected (i got only one).
The core.set_keymap method only set keyboard[0].
If I also set keyboard[1] in this method, then the widget changes the layout correctly.

m-col:
Oh, great spot! Not sure why we're getting 2 keyboards but this helps a lot!

m-col:
Cheers @Fabien-Bernard it was a matter of using `core.keyboards[-1]` as this is the keyboard that is currently being used by the seat. 

Working perfectly for me now!

Ready for review :)

m-col:
Oh nice (or not), the mypy errors are unrelated. In that case if another person can confirm this is working as intended then this is good to go.

Fabien-Bernard:
Tested b3b7ed1 with wayland backend via tty, good to be able to change layout, thanks :)
Nothing to report either with x11 backend test.

I reproduce locally the mypy errors with the `make lint` command. Also with master. I don't understand how some PR have successfull checks if this is blocking.

m-col:
Great, thanks for confirming!

re mypy, I think it must have updated or something and we'll have to adjust some things to make the new version happy. PRs that passed probably had their test run before it updated.

Issue #1934, opened 2020-10-16
Is there still no way to make the bar background transparent?


vaishnavmk:
# Issue description

Is there a supported option to make the bar background transparent, without affecting the opacity of bar icons? I've searched around the documentation and didn't find anything. I don't want switch from the qtile bar.

# Qtile version

0.16.1

wetw3rx:
I would love to know this as well. Played around with the "opacity" setting with less than optimal effect, usually with change to the bar icons as well.

ramnes:
I've spend a few hours on this the other day and yeah, we're not quite there yet. The opacity argument controls the opacity for the window of the bar, so it's normal that everything becomes transparent. The proper way to make the background transparent would be to use a transparent color for the background ([see the formats we support here](https://github.com/qtile/qtile/blob/a987366cd58191eb6495e07469d0c698597c0ed1/libqtile/utils.py#L62-L68)), but it doesn't work because of #900 and #1059. Closing in favor of these issues.

Issue #1722, opened 2020-05-11
Bar Background Opacity


am01k01h3:
Using version 0.15.1,
Bar Opacity applies to the whole bar including the objects which are on the bar. If you set opacity to 0, the whole bar would become completely invisible. 

The more common/expected behavior for opacity is to apply it only to the background and not to the objects.  So, if Opacity was set to 0, it would result in a fully transparent bar, but the objects are still visible.

ramnes:
Yes. IIRC we already have some discussion on this issue somewhere but I could not find right now.

am01k01h3:
After I posted this, I did find some old threads, let me link them here.

Completely transparent bar background ? #1320
https://github.com/qtile/qtile/issues/1320

Bar with transparency redraw issue #1059
https://github.com/qtile/qtile/issues/1059

Incorrect behaviour of alpha channel in Bar background #900
https://github.com/qtile/qtile/issues/900

jackiboi307:
this still happens for me 2021. just installed arch and qtile so im not on an outdated package

Zeus07298101:
> this still happens for me 2021. just installed arch and qtile so im not on an outdated package

Yeah, I'm also still experiencing it too, I can only go down to about opacity = 0.8 otherwise widgets start getting hard to read.

ramnes:
Have a look at #2513

elParaguayo:
#2513 merged. Closing.

Issue #1320, opened 2019-05-13
Completely transparent bar background ? 


g0josh:
Is it possible to have a completely transparent background with only widgets visible? I have been playing with Qtile for a few weeks now (i3 + polybar user previously), and being into python I am happy how flexible it is. Although I was able to set opacity for the whole bar (including widgets), I could not find a way to make just the background completely transparent. The bar draws `#000000` even though I set  it to `ff0000.00`. 
PS: I read about a closed issue(#853 ) about the same matter, the response said it is not possible as of 2016, wondering if there is any update since then!

tych0:
It might be, with some combination of clever setting of the alpha with what's in `libqtile/drawer.py`. See https://cairocffi.readthedocs.io/en/stable/api.html#cairocffi.Context.set_source_rgba ; I think basically if you set the background alpha to 0 and run a compositor, you might (?) get it to work.

g0josh:
Thanks @tych0, I guess that's how it's done currently,

This is from `drawer.py`
```python
def set_source_rgb(self, colour):
        if type(colour) == list:
            if len(colour) == 0:
                # defaults to black
                self.ctx.set_source_rgba(*utils.rgb("#000000"))
            elif len(colour) == 1:
                self.ctx.set_source_rgba(*utils.rgb(colour[0]))
            else:
                linear = cairocffi.LinearGradient(0.0, 0.0, 0.0, self.height)
                step_size = 1.0 / (len(colour) - 1)
                step = 0.0
                for c in colour:
                    rgb_col = utils.rgb(c)
                    if len(rgb_col) < 4:
                        rgb_col[3] = 1
                    linear.add_color_stop_rgba(step, *rgb_col)
                    step += step_size
                self.ctx.set_source(linear)
        else:
            self.ctx.set_source_rgba(*utils.rgb(colour))

def clear(self, colour):
      self.set_source_rgb(colour)
      self.ctx.rectangle(0, 0, self.width, self.height)
      self.ctx.fill()
      self.ctx.stroke()
```
But does not work as expected

tych0:
@g0josh I don't think that's how it's done currently; we don't pass an alpha anywhere as far as I can tell, so it uses the default of 1. You'd have to change the code to pass 0 somehow.

g0josh:
Thanks for taking the time @tych0. I have set my bar background to `#ff0000.0` my bar config. But the background renders black with some symbols and glyphs (probably remains of the previous render). I am running compton as well.
[...]

Issue #900, opened 2016-04-02
Incorrect behaviour of alpha channel in Bar background


jmi2k:
I want to make a status bar with a nice shadow-like gradient background, but I'm limited by the fact that I can't set a gradient with alpha values.

I don't know if this issue is applicable to flat colors too, because of the 'opacity' property of Bar.

flacjacket:
When we set the color to draw [1], the function is called `set_source_rgb`, but it takes and uses RGBA colors, see [2]. Does it work if you try to set an opacity?

[1] https://github.com/qtile/qtile/blob/develop/libqtile/drawer.py#L327

[2] https://github.com/qtile/qtile/blob/develop/libqtile/utils.py#L121

jmi2k:
Oh, I didn't know about that! But there's more: when I set the opacity value of the color to something, it looks like the color were translucent, but there were a black solid background behind it. So, for example: #ff0000 results in a bright red background, but #ff0000.2 gives a very dark red (as opposed to the almost invisible transparent red I want).

flacjacket:
@jmi2k if you have a black background, wouldn't you expect `#ff0000.2` to give a very dark red (since it's seeing through to the black underneath)? You might need to reduce the red-ness to get a lighter red. What if you have a patterned background, do you see it then?

jmi2k:
I'll try to explain better: I have a picture as my background (i use a photo loaded with feh as my background). In the version I'm using (0.10.15, from Arch Linux AUR), if I set an opacity to the bar background, let's say #ff0000.2, what I expect is a red overlay over my image (a reddish background). But instead, I get a solid and dark red, without any transparency. I'll attach some screenshots later today, as well as make some tests with the latest version of this repo, but now this is all I can say:

```
$ qtile --version
0.10.5
$ cat .xinitrc
xrdb -load .Xresources
feh --bg-fill ~/.local/wallpaper
compton -b
exec qtile
```

jmi2k:
I changed the title to describe the *actual* problem.

flacjacket:
I see, I misunderstood you. Yes, that is a problem then, tho I'm not sure where the problem lies, since it may have to do with compositing in order to get transparency working right.

jmi2k:
Some pictures of the bug:

![2016-04-21-152813_1366x768_scrot](https://cloud.githubusercontent.com/assets/17623703/14711492/c73830c0-07d9-11e6-87d6-c17a3721e21a.png)
[...]

PR #2513, opened 2021-06-06
Transparency in X11 and Wayland

elParaguayo:
This PR adds 32-bit colour depth for X11 which allows transparent windows (e.g. the bar). This PR aims to address some existing issues (#900, #1320, #1722, #1934 and probably some others) but will need more work before it can be merged. I'm uploading it now in case people want to test it.

It does look pretty good though:

There's a discussion about getting transparency to work here: https://github.com/qtile/qtile/discussions/2508

This is based on the work of frostidaho from a number of years ago - so big kudos to them for doing most of the legwork.

I'm assuming the system tray will be an issue so I'll look at that next...

Wayland would also need to be looked at separately.

EDIT: systray may be ok:

EDIT 2: To do list

- [x] Handle redrawing of semi-opaque drawers. Overlay mode results in increasing opacity.
- [x] X11 - check if 32 bit enabled and adjust transparency if not (getting some rendering issues with 24-bit depth)
- [x] Testing
  - [x] Check systray widget
    - [x] Confirm transparency for icons
    - [x] Note: SysTray only works for fully opaque/fully transparent bar. Won't fix for semi-opaque.
  - [x] Check more widgets (graphs, images etc.)
    - [x] Fix groupbox widget 
    - [x] Check tasklist (this is the only widget that looks like it may have a similar issue to groupbox)
  - [x] Check whether gradient fill still works
  - [x] Add tests!
- [x] Wayland
  - [x] Check that Wayland backend not broken by these changes 
  - [x] Get transparency working in Wayland too!

### DIFF
--- a/docs/manual/config/screens.rst
+++ b/docs/manual/config/screens.rst
@@ -38,6 +38,17 @@ background="#000000")` will give you a black back ground (the default), while
 :code:`bar.Bar(..., background=["#000000", "#FFFFFF"])` will give you a
 background that fades from black to white.
 
+Bars (and widgets) also support transparency by adding an alpha value to the
+desired color. For example, :code:`bar.Bar(..., background="#00000000")` will
+result in a fully transparent bar. Widget contents will not be impacted i.e.
+this is different to the ``opacity`` parameter which sets the transparency of the
+entire window.
+
+.. note::
+    In X11 backends, transparency will be disabled in a bar if the ``background``
+    color is fully opaque.
+
+
 Fake Screens
 ============
 
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -551,7 +551,6 @@ class Drawer:
         self.set_source_rgb(colour)
         self.ctx.rectangle(0, 0, self.width, self.height)
         self.ctx.fill()
-        self.ctx.stroke()
 
     def textlayout(
         self, text, colour, font_family, font_size, font_shadow, markup=False, **kw
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -417,6 +417,10 @@ class Drawer:
     We stage drawing operations locally in memory using a cairo RecordingSurface before
     finally drawing all operations to a backend-specific target.
     """
+    # We need to track extent of drawing to know when to redraw.
+    previous_rect: Tuple[int, int, Optional[int], Optional[int]]
+    current_rect: Tuple[int, int, Optional[int], Optional[int]]
+
     def __init__(self, qtile: Qtile, win: Internal, width: int, height: int):
         self.qtile = qtile
         self._win = win
@@ -427,12 +431,28 @@ class Drawer:
         self.ctx: cairocffi.Context
         self._reset_surface()
 
-        self.clear((0, 0, 1))
+        self.mirrors: Dict[Drawer, bool] = {}
+
+        self.current_rect = (0, 0, 0, 0)
+        self.previous_rect = (-1, -1, -1, -1)
 
     def finalize(self):
         """Destructor/Clean up resources"""
         self.surface = None
         self.ctx = None
+
+    def add_mirror(self, mirror: Drawer):
+        """Keep details of other drawers that are mirroring this one."""
+        self.mirrors[mirror] = False
+
+    def reset_mirrors(self):
+        """Reset the drawn status of mirrors."""
+        self.mirrors = {m: False for m in self.mirrors}
+
+    @property
+    def mirrors_drawn(self) -> bool:
+        """Returns True if all mirrors have been drawn with the current surface."""
+        return all(v for v in self.mirrors.values())
 
     @property
     def width(self) -> int:
@@ -458,8 +478,27 @@ class Drawer:
         )
         self.ctx = self.new_ctx()
 
+    @property
+    def needs_update(self) -> bool:
+        # We can't test for the surface's ink_extents here on its own as a completely
+        # transparent background would not show any extents but we may still need to
+        # redraw (e.g. if a Spacer widget has changed position and/or size)
+        # Check if the size of the area being drawn has changed
+        rect_changed = self.current_rect != self.previous_rect
+
+        # Check if draw has content (would be False for completely transparent drawer)
+        ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
+
+        return ink_changed or rect_changed
+
     def paint_to(self, drawer: Drawer) -> None:
-        """Paint to another Drawer instance"""
+        drawer.ctx.set_source_surface(self.surface)
+        drawer.ctx.paint()
+        self.mirrors[drawer] = True
+
+        if self.mirrors_drawn:
+            self._reset_surface()
+            self.reset_mirrors()
 
     def _rounded_rect(self, x, y, width, height, linewidth):
         aspect = 1.0
@@ -526,13 +565,17 @@ class Drawer:
     def new_ctx(self):
         return pangocffi.patch_cairo_context(cairocffi.Context(self.surface))
 
-    def set_source_rgb(self, colour: Union[ColorType, List[ColorType]]):
+    def set_source_rgb(self, colour: Union[ColorType, List[ColorType]], ctx: cairocffi.Context = None):
+        # If an alternate context is not provided then we draw to the
+        # drawer's default context
+        if ctx is None:
+            ctx = self.ctx
         if isinstance(colour, list):
             if len(colour) == 0:
                 # defaults to black
-                self.ctx.set_source_rgba(*utils.rgb("#000000"))
+                ctx.set_source_rgba(*utils.rgb("#000000"))
             elif len(colour) == 1:
-                self.ctx.set_source_rgba(*utils.rgb(colour[0]))
+                ctx.set_source_rgba(*utils.rgb(colour[0]))
             else:
                 linear = cairocffi.LinearGradient(0.0, 0.0, 0.0, self.height)
                 step_size = 1.0 / (len(colour) - 1)
@@ -543,9 +586,9 @@ class Drawer:
                         rgb_col[3] = 1
                     linear.add_color_stop_rgba(step, *rgb_col)
                     step += step_size
-                self.ctx.set_source(linear)
+                ctx.set_source(linear)
         else:
-            self.ctx.set_source_rgba(*utils.rgb(colour))
+            ctx.set_source_rgba(*utils.rgb(colour))
 
     def clear(self, colour):
         self.set_source_rgb(colour)
--- a/libqtile/backend/wayland/drawer.py
+++ b/libqtile/backend/wayland/drawer.py
@@ -9,7 +9,7 @@ from libqtile import utils
 from libqtile.backend import base
 
 if TYPE_CHECKING:
-    from typing import Optional
+    from typing import Optional, Tuple
 
     from libqtile.backend.wayland.window import Internal
     from libqtile.core.manager import Qtile
@@ -23,6 +23,10 @@ class Drawer(base.Drawer):
     2. Then apply these operations to our ImageSurface self._source
     3. Then copy the pixels onto the wlr_texture self._target
     """
+
+    # We need to track extent of drawing to know when to redraw.
+    previous_rect: Tuple[int, int, Optional[int], Optional[int]]
+
     def __init__(self, qtile: Qtile, win: Internal, width: int, height: int):
         base.Drawer.__init__(self, qtile, win, width, height)
 
@@ -35,6 +39,8 @@ class Drawer(base.Drawer):
             # Initialise surface to all black
             context.set_source_rgba(*utils.rgb("#000000"))
             context.paint()
+
+        self.previous_rect = (-1, -1, -1, -1)
 
     def paint_to(self, drawer):
         drawer.ctx.set_source_surface(self._source)
@@ -50,12 +56,22 @@ class Drawer(base.Drawer):
         if offsetx > self._win.width:  # type: ignore
             return
 
-        # Only attempt to run RecordingSurface's operations if it actually has
-        # some. self.surface.ink_extents() computes the bounds of the current
-        # list of operations. If any of the bounds are not 0.0 then the surface
-        # has operations and we should paint them.
-        if not any(not math.isclose(0.0, i) for i in self.surface.ink_extents()):
+        # We can't test for the surface's ink_extents here on its own as a completely
+        # transparent background would not show any extents but we may still need to
+        # redraw (e.g. if a Spacer widget has changed position and/or size)
+
+        # Check if current rect has changed since previous draw
+        current_rect = (offsetx, offsety, width, height)
+        rect_changed = current_rect != self.previous_rect
+
+        # Check if draw has content (would be False for completely transparent drawer)
+        ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
+
+        if not ink_changed and not rect_changed:
             return
+
+        # Keep track of latest rect covered by this drawwer
+        self.previous_rect = current_rect
 
         # Make sure geometry doesn't extend beyond texture
         if width is None:
@@ -85,3 +101,24 @@ class Drawer(base.Drawer):
 
         # Clear RecordingSurface of operations
         self._reset_surface()
+
+    def clear(self, colour):
+        # Some hacks needed for transparency
+
+        # Check if any colour is not fully opaque
+        if utils.has_transparency(colour):
+
+            # RecordingSurface won't write clear operation to surface so we
+            # need to clear that surface directly.
+            if getattr(self, "_source", None) is not None:
+                ctx = cairocffi.Context(self._source)
+                ctx.save()
+                ctx.set_operator(cairocffi.OPERATOR_CLEAR)
+                ctx.paint()
+                ctx.restore()
+
+        self.ctx.save()
+        self.ctx.set_operator(cairocffi.OPERATOR_SOURCE)
+        self.set_source_rgb(colour)
+        self.ctx.paint()
+        self.ctx.restore()
--- a/libqtile/backend/wayland/drawer.py
+++ b/libqtile/backend/wayland/drawer.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import math
 from typing import TYPE_CHECKING
 
 import cairocffi
@@ -9,7 +8,7 @@ from libqtile import utils
 from libqtile.backend import base
 
 if TYPE_CHECKING:
-    from typing import Optional, Tuple
+    from typing import Optional
 
     from libqtile.backend.wayland.window import Internal
     from libqtile.core.manager import Qtile
@@ -24,9 +23,6 @@ class Drawer(base.Drawer):
     3. Then copy the pixels onto the wlr_texture self._target
     """
 
-    # We need to track extent of drawing to know when to redraw.
-    previous_rect: Tuple[int, int, Optional[int], Optional[int]]
-
     def __init__(self, qtile: Qtile, win: Internal, width: int, height: int):
         base.Drawer.__init__(self, qtile, win, width, height)
 
@@ -40,12 +36,6 @@ class Drawer(base.Drawer):
             context.set_source_rgba(*utils.rgb("#000000"))
             context.paint()
 
-        self.previous_rect = (-1, -1, -1, -1)
-
-    def paint_to(self, drawer):
-        drawer.ctx.set_source_surface(self._source)
-        drawer.ctx.paint()
-
     def draw(
         self,
         offsetx: int = 0,
@@ -56,22 +46,15 @@ class Drawer(base.Drawer):
         if offsetx > self._win.width:  # type: ignore
             return
 
-        # We can't test for the surface's ink_extents here on its own as a completely
-        # transparent background would not show any extents but we may still need to
-        # redraw (e.g. if a Spacer widget has changed position and/or size)
+        # We need to set the current draw area so we can compare to the previous one
+        self.current_rect = (offsetx, offsety, width, height)
+        # rect_changed = current_rect != self.previous_rect
 
-        # Check if current rect has changed since previous draw
-        current_rect = (offsetx, offsety, width, height)
-        rect_changed = current_rect != self.previous_rect
-
-        # Check if draw has content (would be False for completely transparent drawer)
-        ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
-
-        if not ink_changed and not rect_changed:
+        if not self.needs_update:
             return
 
         # Keep track of latest rect covered by this drawwer
-        self.previous_rect = current_rect
+        self.previous_rect = self.current_rect
 
         # Make sure geometry doesn't extend beyond texture
         if width is None:
@@ -99,26 +82,17 @@ class Drawer(base.Drawer):
         )
         self._win.damage()  # type: ignore
 
-        # Clear RecordingSurface of operations
-        self._reset_surface()
+        # If the widget is not being reflected then clear RecordingSurface of operations
+        # If it is, we need to keep the RecordingSurface contents until the mirrors have
+        # been drawn
+        if not self.mirrors:
+            self._reset_surface()
 
     def clear(self, colour):
-        # Some hacks needed for transparency
-
-        # Check if any colour is not fully opaque
-        if utils.has_transparency(colour):
-
-            # RecordingSurface won't write clear operation to surface so we
-            # need to clear that surface directly.
-            if getattr(self, "_source", None) is not None:
-                ctx = cairocffi.Context(self._source)
-                ctx.save()
-                ctx.set_operator(cairocffi.OPERATOR_CLEAR)
-                ctx.paint()
-                ctx.restore()
-
-        self.ctx.save()
-        self.ctx.set_operator(cairocffi.OPERATOR_SOURCE)
-        self.set_source_rgb(colour)
-        self.ctx.paint()
-        self.ctx.restore()
+        # Draw background straight to ImageSurface
+        ctx = cairocffi.Context(self._source)
+        ctx.save()
+        ctx.set_operator(cairocffi.OPERATOR_SOURCE)
+        self.set_source_rgb(colour, ctx=ctx)
+        ctx.paint()
+        ctx.restore()
--- a/libqtile/backend/x11/drawer.py
+++ b/libqtile/backend/x11/drawer.py
@@ -30,6 +30,7 @@ class Drawer(base.Drawer):
         self._xcb_surface = None
         self._pixmap = None
         self._gc = None
+        self._depth, self._visual = qtile.core.conn.default_screen._get_depth_and_visual(win._depth)  # type: ignore
 
     def finalize(self):
         self._free_xcb_surface()
@@ -89,7 +90,7 @@ class Drawer(base.Drawer):
         surface = cairocffi.XCBSurface(
             self.qtile.core.conn.conn,
             self._pixmap,
-            self.qtile.core.conn.default_screen.default_visual,
+            self._visual,
             self.width,
             self.height,
         )
@@ -103,7 +104,7 @@ class Drawer(base.Drawer):
     def _create_pixmap(self):
         pixmap = self.qtile.core.conn.conn.generate_id()
         self.qtile.core.conn.conn.core.CreatePixmap(
-            self.qtile.core.conn.default_screen.default_depth,
+            self._depth,
             pixmap,
             self._win.wid,
             self.width,
@@ -176,7 +177,7 @@ class Drawer(base.Drawer):
     def clear(self, colour):
         # If the drawer is 32 bit then we need to do some extra work to clear it
         # before drawing semi-opaque content
-        if utils.has_transparency(colour) and self.qtile.core.conn.default_screen.default_depth == 32:
+        if utils.has_transparency(colour) and self._depth == 32:
 
             # RecordingSurface won't write clear operation to surface so we
             # need to clear that surface directly.
@@ -193,7 +194,7 @@ class Drawer(base.Drawer):
 
     def set_source_rgb(self, colour):
         # Remove transparency from non-32 bit windows
-        if utils.has_transparency(colour) and self.qtile.core.conn.default_screen.default_depth != 32:
+        if utils.has_transparency(colour) and self._depth != 32:
             colour = utils.remove_transparency(colour)
 
         base.Drawer.set_source_rgb(self, colour)
--- a/libqtile/backend/x11/drawer.py
+++ b/libqtile/backend/x11/drawer.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 import cairocffi
 import xcffib.xproto
 
+from libqtile import utils
 from libqtile.backend import base
 
 if TYPE_CHECKING:
@@ -88,7 +89,7 @@ class Drawer(base.Drawer):
         surface = cairocffi.XCBSurface(
             self.qtile.core.conn.conn,
             self._pixmap,
-            self._find_root_visual(),
+            self.qtile.core.conn.default_screen.default_visual,
             self.width,
             self.height,
         )
@@ -102,7 +103,7 @@ class Drawer(base.Drawer):
     def _create_pixmap(self):
         pixmap = self.qtile.core.conn.conn.generate_id()
         self.qtile.core.conn.conn.core.CreatePixmap(
-            self.qtile.core.conn.default_screen.root_depth,
+            self.qtile.core.conn.default_screen.default_depth,
             pixmap,
             self._win.wid,
             self.width,
@@ -171,3 +172,28 @@ class Drawer(base.Drawer):
             for v in i.visuals:
                 if v.visual_id == self.qtile.core.conn.default_screen.root_visual:
                     return v
+
+    def clear(self, colour):
+        # If the drawer is 32 bit then we need to do some extra work to clear it
+        # before drawing semi-opaque content
+        if utils.has_transparency(colour) and self.qtile.core.conn.default_screen.default_depth == 32:
+
+            # RecordingSurface won't write clear operation to surface so we
+            # need to clear that surface directly.
+            if getattr(self, "_xcb_surface", None) is not None:
+                ctx = cairocffi.Context(self._xcb_surface)
+                ctx.save()
+                ctx.set_operator(cairocffi.OPERATOR_CLEAR)
+                ctx.paint()
+                ctx.restore()
+
+        self.set_source_rgb(colour)
+        self.ctx.rectangle(0, 0, self.width, self.height)
+        self.ctx.fill()
+
+    def set_source_rgb(self, colour):
+        # Remove transparency from non-32 bit windows
+        if utils.has_transparency(colour) and self.qtile.core.conn.default_screen.default_depth != 32:
+            colour = utils.remove_transparency(colour)
+
+        base.Drawer.set_source_rgb(self, colour)
--- a/libqtile/backend/x11/drawer.py
+++ b/libqtile/backend/x11/drawer.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import math
 from typing import TYPE_CHECKING
 
 import cairocffi
@@ -117,26 +116,27 @@ class Drawer(base.Drawer):
             self.qtile.core.conn.conn.core.FreePixmap(self._pixmap)
             self._pixmap = None
 
-    def paint_to(self, drawer):
-        # If XCBSurface has been invalidated, we need to draw now to create it
+    def _check_xcb(self):
+        # If the Drawer has been resized/invalidated we need to recreate these
         if self._xcb_surface is None:
-            self.draw()
-        drawer.ctx.set_source_surface(self._xcb_surface)
-        drawer.ctx.paint()
+            self._pixmap = self._create_pixmap()
+            self._xcb_surface = self._create_xcb_surface()
 
     def _paint(self):
-        # Only attempt to run RecordingSurface's operations if it actually has
-        # some. self.surface.ink_extents() computes the bounds of the current
-        # list of operations. If any of the bounds are not 0.0 then the surface
-        # has operations and we should paint them.
-        if any(not math.isclose(0.0, i) for i in self.surface.ink_extents()):
+        # Only attempt to run RecordingSurface's operations if ie actually need to
+        if self.needs_update:
             # Paint RecordingSurface operations to the XCBSurface
             ctx = cairocffi.Context(self._xcb_surface)
             ctx.set_source_surface(self.surface, 0, 0)
             ctx.paint()
 
-            # Clear RecordingSurface of operations
-            self._reset_surface()
+            # If the widget is not being reflected then clear RecordingSurface of operations
+            # If it is, we need to keep the RecordingSurface contents until the mirrors have
+            # been drawn
+            if not self.mirrors:
+                self._reset_surface()
+
+            self.previous_rect = self.current_rect
 
     def draw(
         self,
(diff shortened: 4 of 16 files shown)


ramnes:
Nice. Can you please share the to do list? Maybe you could get some help. :)

elParaguayo:
Hopefully it's just a case of testing now. I thought that the systray may be a problematic because that seemed to be an issue when frostidaho originally worked on this. It may be working now by my sheer dumb luck!

m-col:
The Wayland `Window`s are all rendered as 32bit so should work fine without much changing.

Do we need to make the depth configurable? Perhaps we could try to use 32bit but `get_depth_and_visual` could fallback to the root visual if a 32bit depth isn't available?

Could you elaborate on the custom colormap stuff and the issues with RecordingSurface? Just so I understand what's going on :)

m-col:
Just seen https://github.com/qtile/qtile/discussions/2508#discussioncomment-831152 which helped. What does `cairocffi.OPERATOR_CLEAR` do? Is it the case the `RecordingSurface` is incompatible with using alpha?

elParaguayo:
> The Wayland `Window`s are all rendered as 32bit so should work fine without much changing.

Well, it didn't work when I tried it but I don't know if I've got Wayland working or not (got a solid magenta background when I started it).

> Do we need to make the depth configurable? Perhaps we could try to use 32bit but `get_depth_and_visual` could fallback to the root visual if a 32bit depth isn't available?

Currently, you can set the depth for the bar but, other than that, the changes default to requesting 32 bit windows for internal windows. It should fallback to root visual if 32 bit isn't available.

> Could you elaborate on the custom colormap stuff and the issues with RecordingSurface? Just so I understand what's going on :)

Custom colormap was shamelessly copied from frostidaho's work (https://github.com/qtile/qtile/issues/1059#issuecomment-298073985).

As for the issue with the RecordingSurface, [this post](https://github.com/qtile/qtile/discussions/2508#discussioncomment-831152) on our discussions page should explain the issue. In short, the clear operator is not applied to the XCBSurface. My "solution" was to write the clear direct to the XCBSurface.

elParaguayo:
Ah - our posts overlapped.

[This page](https://www.cairographics.org/operators/) is good at explaining the operators but, in short, `OPERATOR_CLEAR` removes content from the area being painted. The alternative is to use `OPERATOR_SOURCE` which doesn't do any blending and just puts the new source onto the surface. The default operator is `OPERATOR_OVER` which does blending.

I haven't tested more but it may be the case that `RECORDING_SURFACES` aren't handling changing operators correctly. However, if we wrote directly to an `XCBSurface` we wouldn't have this issue.

m-col:
> Well, it didn't work when I tried it but I don't know if I've got Wayland working or not (got a solid magenta background when I started it).

The magenta is actually currently the default "you have no background set" colour, which I was using to help with damage tracking because it's so high contrast :p I should probably change that to black like the X server.... As for `Internal`, they themselves are represented by a 32 bit texture to which pixels are copied but `Drawer` will need updating in the same way as the x11 `Drawer`, and if some parts overlap then they can be implemented in base.py as is currently the case.

> Custom colormap was shamelessly copied from frostidaho's work (#1059 (comment)).

OK thanks, think I'm following now.

> I haven't tested more but it seems that RecordingSurfaces aren't handling changing operators correctly. However, if we wrote directly to an XCBSurface we wouldn't have this issue.

IIRC RecordingSurfaces were introduced so that the underlying pixmaps can be resized (i.e. destroyed and then replaced with different one) while keeping the drawing operations around and usable across pixmaps. Maybe we could find an alternative solution for that, otherwise we'll have to figure out a way to accomplish using alpha while continuing to use RecordingSurfaces.

elParaguayo:
> The magenta is actually currently the default "you have no background set" colour, which I was using to help with damage tracking because it's so high contrast :p I should probably change that to black like the X server.... As for `Internal`, they themselves are represented by a 32 bit texture to which pixels are copied but `Drawer` will need updating in the same way as the x11 `Drawer`, and if some parts overlap then they can be implemented in base.py as is currently the case.

Good to know. I would like to inlcude Wayland in this PR if I can (will need to rename it...) and it sounds like it might be easy to do so I'll add to the list at the top.

> IIRC RecordingSurfaces were introduced so that the underlying pixmaps can be resized (i.e. destroyed and then replaced with different one) while keeping the drawing operations around and usable across pixmaps. Maybe we could find an alternative solution for that, otherwise we'll have to figure out a way to accomplish using alpha while continuing to use RecordingSurfaces.

As far as I can tell, it's only going to be an issue when we need to clear the background which should be solved by my updated `clear` code. If that's the case, then I don't see a need to necessarily move away from using a `RecordingSurface`. For the rest of the time, if you're drawing semi transparent content over the background in the `Drawer` then the `RecordingSurface` should be fine as it will be using `OPERATOR_OVER` which is exactly the behaviour we'd want at that point.

elParaguayo:
@m-col where is the magenta set in the wayland backend?

ramnes:
> Hopefully it's just a case of testing now.

You probably want to take a look at @m-col's work on #1697 for this, the tests should be quite similar I think?

elParaguayo:
Will do. Although I was thinking of more "hands-on" tests and seeing if things still display properly. A test script won't necessarily do that.

elParaguayo:
Good news is that transparency is working in wayland (much less effort than x11).

However, I'm getting some interesting differences between x11 and wayland with same config.

wayland

More digging to do!

m-col:
Very cool. What differences are you seeing?

elParaguayo:
If you look at the left-hand side you can see that the layout icon is transparent and shows the bar colour in x11 but, in wayland, you see the wallpaper instead. x11 is correct here.

However, also in x11, the group box has lines underneath the inactive groups whereas wayland doesn't. wayland is correct on this.

m-col:
Nice, nice. re to do list entry `Check multiple systray widgets`, what's the worry there? I always found that having multiple systrays was buggy so I put one only on my first screen.

elParaguayo:
To be very honest: that's a typo.

I was thinking about checking multiple widgets (as in I want to check whether transparency threw out any issues in individual widgets, like the GroupBox in my earlier post). I was also thinking about the system tray as that had been flagged earlier as not working with transparency. My brain just squashed those ideas together!

m-col:
Lol fair. The systray is an interesting case because the pixels are provided by clients. Should transparency even apply to its icons?

elParaguayo:
Systray is one of the widgets I'm least familiar with but I plan to look at it later today as it's crashing on wayland at the moment but I suspect that's a quick fix. Once it's running I can then test the transparency point.

In terms of your question, I would hope that the icons are in 32 bit windows so transparency is available to them. That seems to be the case with one of my earlier screenshots showing vlc in the tray but I haven't dug in to the code to see how it all fits together.

m-col:
> Systray is one of the widgets I'm least familiar with but I plan to look at it later today as it's crashing on wayland at the moment but I suspect that's a quick fix. 

I could be wrong, but my understanding of the systray is that it is inherently tied to X (https://specifications.freedesktop.org/systemtray-spec/systemtray-spec-0.2.html), and therefore will never work under Wayland. For this reason I haven't looked at it at all as part of the wayland work. However, that is not to say that a system tray cannot be used under Wayland, it would just use a different protocol, e.g. https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/ which is what Swaybar uses.

elParaguayo:
Good to know. So my `quick fix` will be to remove it from my wayland config!

Can look at making a wayland systray another time.

m-col:
> Good to know. So my `quick fix` will be to remove it from my wayland config!

:D yep!

elParaguayo:
Interesting bug in `SysTray` with these changes. First icon displays fine but subsequent icons copy the drawing from the first icon and so on.

On Master, it should look like this:

But on this branch, it looks like this:


Need to dig in to the `SysTray` code and see how the drawers work there.

elParaguayo:
I think the bug has something to do with this error in the logs:
```
2021-06-11 08:23:23,950 ERROR libqtile core.py:_xpoll():L329 Got an exception
in poll loop
  File
"/home/elp/Documents/scripts/github/qtile-dev/qtile/libqtile/backend/x11/core.py",
line 282, in _xpoll
  File
"/home/elp/Documents/scripts/github/qtile-dev/qtile/venv/lib/python3.9/site-packages/xcffib/__init__.py",
line 571, in wrapper
    return f(*args)
  File
"/home/elp/Documents/scripts/github/qtile-dev/qtile/venv/lib/python3.9/site-packages/xcffib/__init__.py",
line 612, in poll_for_event
    return self.hoist_event(e)
  File
"/home/elp/Documents/scripts/github/qtile-dev/qtile/venv/lib/python3.9/site-packages/xcffib/__init__.py",
line 686, in hoist_event
    return self._process_error(ffi.cast("xcb_generic_error_t *", e))
  File
"/home/elp/Documents/scripts/github/qtile-dev/qtile/venv/lib/python3.9/site-packages/xcffib/__init__.py",
line 649, in _process_error
    raise error(buf)
xcffib.xproto.MatchError
```
I don't get this when the bar's window depth is set to 24 bit so I'm assuming that there's some mismatch between the bar and the SysTray in terms of the depth or visual but I'm having difficulties tracing it down.

m-col:
Anything interesting here? https://github.com/polybar/polybar/blob/master/src/x11/tray_manager.cpp

elParaguayo:
Certainly some interesting bits in there but there's also this:
```
TODO: 32-bit visual
_NET_SYSTEM_TRAY_VISUAL visual_id VISUALID/32

The property should be set by the tray manager to indicate the preferred visual for icon windows.

To avoid ambiguity about the colormap to use this visual must either be the default visual for the screen 
or it must be a TrueColor visual. If this property is set to a visual with an alpha channel, the tray manager 
must use the Composite extension to composite the icon against the background using PictOpOver.
```

m-col:
How will that affect what you're doing?

elParaguayo:
Hopefully it doesn't but, given I've been adding 32 bit visuals to get transparency in the bar, I was a bit disconcerted to see references to transparency in a "To do" section.

I'll take a proper look through that code and compare to our tray's code later.

m-col:
Oh right lol yeah, not very helpful. It seems that they/we wouldn't want to
apply the transparency to the icons too which i guess makes sense.

libqtile/backend/wayland/drawer.py
```diff
@@ -55,13 +57,23 @@ def draw(
...
-        # list of operations. If any of the bounds are not 0.0 then the surface
-        # has operations and we should paint them.
-        if not any(not math.isclose(0.0, i) for i in self.surface.ink_extents()):
+        # We can't test for the surface's ink_extents here on it's own as a completely
+        # transparent background would not show any extents but we may still need to
+        # redraw (e.g. if a Spacer widget has changed position and/or size)
+
+        # Check if current rect has changed since previous draw
+        current_rec = (offsetx, offsety, width, height)
+        rect_changed = any([p != c for p, c in zip(self.previous_rect, current_rec)])
```
m-col, line 68: if the rect has changed but the window is all transparent, does it really need rendering?

libqtile/backend/wayland/drawer.py
```diff
@@ -55,13 +57,23 @@ def draw(
...
+        
+        # Check if draw has content (would be False for completely transparent drawer)
+        ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
+
+        if not rect_changed and not ink_changed:
```
m-col, line 73: If we only need one of these to be `True` then we could check the one least likely to be `False` first and if it's `True` then we don't need to calculate the other

libqtile/backend/wayland/drawer.py
elParaguayo: Yes, because if the window has changed size then you need to overwrite the area that may have had content from a different widget in it.

libqtile/backend/x11/xcbq.py
```diff
@@ -619,6 +676,9 @@ def paint(self, screen, image_path, mode=None):
             logger.error('Wallpaper: %s' % e)
             return
 
+        max_depth = max(self.default_screen.visuals)
+        _, root_visual = self.default_screen.visuals[max_depth]
```
m-col, line 680: It's this bit that errors for me, because `self.default_screen.visuals` is just a list of visuals and so it can't be unpacked into two parts

elParaguayo: Yup - this is wrong. Fixed in latest commit.

libqtile/backend/wayland/drawer.py
elParaguayo: Ok, so if users typically have fixed size widgets then `ink_changed` should go first.

m-col: yeah makes sense

libqtile/backend/wayland/drawer.py
m-col: ah yeah of course

libqtile/backend/x11/drawer.py
```diff
@@ -163,11 +171,43 @@ def draw(
...
+                    ctx.save()
+                    ctx.set_operator(cairocffi.OPERATOR_CLEAR)
+                    ctx.paint()
+                    ctx.restore()
+
+                # May not be needed due to the above...
+                self.ctx.save()
+                self.ctx.set_operator(cairocffi.OPERATOR_CLEAR)
+                self.ctx.paint()
+                self.ctx.restore()
```
m-col, line 206: With `background='ffffffff'` added to the default config's Bar I'm not seeing any transparency. This is with a plain red background:
image

elParaguayo: ffffffff is solid white, there should be no transparency with that.

elParaguayo: for alpha, "ff" is opaque and "00" is transparent.

elParaguayo:
Tray looking better now...

elParaguayo:
On a related topic: there's [an interesting thread/issue on the i3 respository about system tray and StatusNotifier](https://github.com/i3/i3/issues/2088). Looks like some similar issues to what we're seeing and a reluctance to implement StatusNotifier in their codebase.

elParaguayo:
I think the key to fixing this is this part of the system tray spec regarding transparency: 

> If this property is set to a visual with an alpha channel, the tray manager must use the Composite extension to composite the icon against the background using PictOpOver.

That makes sense as it suggest that the icon needs to be overlaid over the parent window rather than replacing it but I can't see much documentation about doing this in xcb.

libqtile/backend/x11/drawer.py
m-col: Lol yes of course

m-col:
> "I can't see much documentation about doing this in xcb."

Classic XCB.

With XCB, the only feasible way to figure out how to do something is to read the code of some other software that already does it :)

libqtile/widget/systray.py
```diff
@@ -150,6 +150,25 @@ def _configure(self, qtile, bar):
...
         qtile.core._root.send_event(event, mask=EventMask.StructureNotify)
 
+        # Bit of an ugly hack here
+        # Don't immediately see why this shouldn't work for 24 bit depth visuals...
+        if self.drawer.depth == 32:
+
+            # If we're using transparency then We need tray to tell icons which
+            # visual to use
+            tray_attrs = self.conn.conn.core.GetWindowAttributes(win.wid).reply()
+            tray_visual = tray_attrs.visual
```
m-col, line 160: What value does `tray_visual` have? Will it always be whatever we specified when the window was created?

elParaguayo: Are you trying to point out that I could have just pulled that value from the `drawer`...?

elParaguayo:
> > "I can't see much documentation about doing this in xcb."
> 
> Classic XCB.
> 
> With XCB, the only feasible way to figure out how to do something is to read the code of some other software that already does it :)

Yes - that's what I've been doing. The bit I'm most confused about (so far) is that the `composite` method seems to be part of `xcb.render` and the documentation seems to refer to an `xcb.render.Connection` but I don't think our `Connection` is.

elParaguayo:
The other idea I was toying with was weather we could extract the icon's pixmap and then use `cairo` to paint this over the bar background and then use `CopyArea` to draw it all back to the icon.

I think this is essentially what `composite` does, but I can't find how to access it yet.

libqtile/widget/systray.py
m-col: Only if the answer is yes :laughing:

m-col:
> I think this is essentially what composite does, but I can't find how to access it yet.

Check out how the randr extension is accessed, might help: https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/xcbq.py#L361

m-col:
> Still some small issues with the background (you'll see the square by the icon is a slightly different colour - it's not a good image to show the issue) but I've got a good idea how to fix that bit.

My vote goes to not fixing this and considering it a caveat of using the systray and transparency together. Getting them to work together and not having these boxes will almost certainly be a big chunk of work with not very huge gains. GNOME and KDE have moved away from the old systray spec, and it is largely getting replaced by StatusNotifierItem spec as far as I can tell. Indeed other WMs have come to the same conclusion:

- https://github.com/awesomeWM/awesome/issues/363
- https://github.com/Airblader/i3/issues/90#issuecomment-225995287 <- Read comment for more explanation of why
- https://github.com/taffybar/taffybar/issues/177

elParaguayo:
As much as I hate to admit defeat, this definitely feels like the right answer.

I've got one last avenue to try but, given that more experienced people than me have have not implemented a solution, I'm pretty realistic about my chance of success.

Once I've failed, I'll go back over the changes and do some tidying up.

Massive thanks for all your help and pointers along the way.

elParaguayo:
Admitted defeat on the last part of SysTray but, given all the other people who've abandoned this, I feel no shame!

This is very close now. Just some more testing to do.

libqtile/bar.py
```diff
@@ -160,7 +160,7 @@ class Bar(Gap, configurable.Configurable):
         ("background", "#000000", "Background colour."),
         ("opacity", 1, "Bar window opacity."),
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
-        ("depth", 32, "Desired colour depth for bar")
+        ("depth", 24, "Desired colour depth for bar. 32 bit depth needed for transparency.")
```
elParaguayo, line 163: I've defaulted to 24 bit as the default background is "#000000".

We could check if background has requested transparency and flip this to "32" when the bar is initialised. Any objections/concerns?

m-col: Is there any reason to have this configurable at all? We can default to 32 in all cases (don't bother checking background colours - the underlying window doesn't know what's going on it), only falling back to 24 if for whatever reason the system/X server doesn't support 32 bit. Saying that, I don't think implementing the fallback makes sense until someone reports it is needed -- maybe everyone running X these days can use 32 bit windows. So maybe hard-code 32bits everywhere?

elParaguayo: I did think about that and I agree it's probably the right answer but I was taking a cautious approach to avoid any breakages.

As for removing fallbacks, we can keep this bit in and just log a message if there isn't a 32 bit visual available:
https://github.com/qtile/qtile/blob/9f513d157dcd4edac49cdc56ad2260eb28e42406/libqtile/backend/x11/xcbq.py#L315-L322

elParaguayo: Thinking about it some more, maybe a better approach is to set a `default_depth` and `default_visual` and have these be 32 bit unless it's not available and then it falls back to `root_depth` and the equivalent visual.

That should probably help remove some code.

m-col: Whatever you think will be the nicest!

elParaguayo:
OK - this is probably ready for review.

It's pretty light on tests (just a couple of utils that I added) but can add more if anyone thinks I've missed something obvious.

libqtile/backend/x11/xcbq.py
```diff
@@ -541,18 +588,24 @@ def keycode_to_keysym(self, keycode, modifier):
 
     def create_window(self, x, y, width, height):
         wid = self.conn.generate_id()
+
+        value_mask = CW.BackPixmap | CW.BorderPixel | CW.EventMask | CW.Colormap
+        values = [
+            xcffib.xproto.BackPixmap._None,
+            0,
```
m-col, line 595: Why do we need to specify the BorderPixel now?

libqtile/utils.py
```diff
@@ -110,6 +110,43 @@ def hex(x):
     return '#%02x%02x%02x' % (int(r * 255), int(g * 255), int(b * 255))
 
 
+def has_transparency(colour):
```
m-col, line 113: These can be typed using `ColorType`

libqtile/backend/x11/xcbq.py
elParaguayo: See: https://lists.freedesktop.org/archives/xcb/2017-October/011047.html

libqtile/backend/base.py
```diff
@@ -539,7 +543,6 @@ def clear(self, colour):
         self.set_source_rgb(colour)
         self.ctx.rectangle(0, 0, self.width, self.height)
         self.ctx.fill()
-        self.ctx.stroke()
```
m-col, line 542: Was this doing nothing?

elParaguayo: I think so. `fill` fills the entire path and, as the `rectangle` is the entire drawer, there shouldn't be any pixels that haven't been painted by this.

libqtile/utils.py
elParaguayo: Indeed. Need to get better at typing.

libqtile/backend/x11/xcbq.py
m-col: Interesting. Thanks!

m-col:
Overall this is looking really good.

I don't think we need to make any of the depth-related code visible from outside of a given backend. The `is_24_bit` property is there as an implementation detail of the x11 backend, and I think all checks using it should be restricted to that backend. Similarly for the `depth` property -- I guess this was useful for debugging the x11 backend, as you're printing it in `Bar.info`. Is this still useful? As for the groupbox widget, I think it is fine to just assume (from the perspective of code outside of the backends) that 32 bit stuff will work; the backend is then responsible for dealing with it, if it doesn't support it. Is there a problem with that approach?

What would you think about pre-calculating things like whether a colour has transparency, to avoid doing stuff repeatedly in `Drawer.clear`? Maybe `_Widget._configure` could go through all of its attributes looking for things that look like colours (or those that match `"color" in attr` maybe) then converting them all into the tuple form, possibly?

libqtile/backend/x11/xcbq.py
elParaguayo: I did also check that this was needed and it does stop working if you remove it.

libqtile/utils.py
elParaguayo: I think it's actually typed with `Union[ColorType, List[ColorType]]` as it can also accept list of colours (e.g. where a user has specified multiple colours for a gradient)

elParaguayo:
> What would you think about pre-calculating things like whether a colour has transparency, to avoid doing stuff repeatedly in `Drawer.clear`? Maybe `_Widget._configure` could go through all of its attributes looking for things that look like colours (or those that match `"color" in attr` maybe) then converting them all into the tuple form, possibly?

Starting with your last question first as it's relevant for your other questions. Pre-calculating would be good but may be overkill. As far as I can tell, it's only relevant for widgets that use the `background` property for something other than drawing a background. Both `TaskList` and `GroupBox` use it to paint a border when the border is meant to be invisible. That's fine for opaque drawings but, with any semit-transparent colour, you get alpha blending. So, we don't want to change the `background` property but we just need to change it for the border. This could be done once in `_configure` rather than on every `draw` but it's not a case of changing every colour.  

> I don't think we need to make any of the depth-related code visible from outside of a given backend. The `is_24_bit` property is there as an implementation detail of the x11 backend, and I think all checks using it should be restricted to that backend.

We need to be able to check the colour depth. If this property isn't available in wayland then we'd have to check the backend first and then look at the colour dept if we're running x11. That felt like overkill so I went with this option instead. Happy to listen to alternatives if there's a better way of doing this.

> Similarly for the `depth` property -- I guess this was useful for debugging the x11 backend, as you're printing it in `Bar.info`. Is this still useful? 

Was going to use it in tests but not sure it's meaningful. You're right that it's not used anywhere else (other than the `is_24_bit` property).

> As for the groupbox widget, I think it is fine to just assume (from the perspective of code outside of the backends) that 32 bit stuff will work; the backend is then responsible for dealing with it, if it doesn't support it. Is there a problem with that approach?

Yes. As above, if we don't change the widget code, it will not draw as expected in certain cases in 32-bit mode. We have to replace a colour with a fully transparent one. The backend has no way to determine when this replacement needs to happen. The widget has to be adjusted for this. It was written with fully opaque colours in mind.

m-col:
> Both TaskList and GroupBox use it to paint a border when the border is meant to be invisible.

Could we not just update these so that they don't?

> We need to be able to check the colour depth.

Sorry if I'm being dense, why do widgets need to know the depth?

>  We have to replace a colour with a fully transparent one. The backend has no way to determine when this replacement needs to happen.

So you're saying that we can't leave it to the Drawers to know when to change the colours, so the widgets have to do it. The problem is with using opaque colours on a transparent window?

> Was going to use it in tests but not sure it's meaningful. You're right that it's not used anywhere else (other than the is_24_bit property).

I think to reduce noise on the base class "API" of Drawers it'd be best to stick the depth as an attribute only on the x11 Drawers. You'll still be able to access it in tests.

elParaguayo:
> > Both TaskList and GroupBox use it to paint a border when the border is meant to be invisible.
> 
> Could we not just update these so that they don't?

That would be nicer. The problem is that there's a `drawbox` method which needs a border value and the border width is not configurable for individual calls. So I could either re-write the widget to not draw a border in certain situations or I can set the colour based on the backend and keep the existing method.

I went with the latter because it was quicker but, if you don't want to expose the colour depth issues to the widgets, I can change to the former (which is fine by me).

> > We need to be able to check the colour depth.
> 
> Sorry if I'm being dense, why do widgets need to know the depth?

They don't _need_ to but it was the option I chose. If we rewrite as above then I don't think it's necessary.

> > We have to replace a colour with a fully transparent one. The backend has no way to determine when this replacement needs to happen.
> 
> So you're saying that we can't leave it to the Drawers to know when to change the colours, so the widgets have to do it. The problem is with using opaque colours on a transparent window?

Not quite. I was saying that, in GroupBox and TaskList, the same colour value was being used for two different aspects. Only one needed to be changed but the drawer couldn't know that as it just gets a colour value.

> > Was going to use it in tests but not sure it's meaningful. You're right that it's not used anywhere else (other than the is_24_bit property).
> 
> I think to reduce noise on the base class "API" of Drawers it'd be best to stick the depth as an attribute only on the x11 Drawers. You'll still be able to access it in tests.

Sounds good.

m-col:
It's this bit that I think I'm still not understanding:

> The problem is that there's a drawbox method which needs a border value and the border width is not configurable for individual calls. So I could either re-write the widget to not draw a border in certain situations

(And sorry I know you will have explained before, please bear with me!) This "situations" you're referring to are when the window is 24bit, right? What happens when you specify a 32bit colour on a 24bit window and vice versa?

elParaguayo:
OK - I've probably done a bad job of explaining. So let me try again.

The issue with these two widgets arises with 32-bit windows when the user has a semi-transparent background colour. In that scenario, the widget uses that colour to draw its background. So far so good. The problem comes because the widget then uses that same colour to draw a border around inactive items. There's no issue if the background is fully transparent (as drawing fully transparent on fully transparent is still fully transparent). Same goes for fully opaque. However, for semi-transparent, drawing the same colour on top of itself results in alpha blending and you get a darker (i.e. visible) border when there should be none.

Taking `GroupBox` as an example, the draw method loops over each group and runs a `drawbox` method on them. The border colour is one of the arguments required for this method so, when we don't want a visible border, the widgets passes the background colour (L377 below):
https://github.com/qtile/qtile/blob/59653ab8642c14fce5aa0a65d06c95f5b3cc8ea2/libqtile/widget/groupbox.py#L376-L390

A better approach would be to say that no border should be drawn in that situation.

I was about to say the issue shouldn't arise with 24 bit windows but I think it actually might in certain scenarios (which may be a side effect of using my `remove_transparency` utility in `drawer.clear` when it may be better placed in `set_source_rgb`).

m-col:
> A better approach would be to say that no border should be drawn in that situation.

Yeah this is definitely the best solution! :smile:

elParaguayo:
Will do a squash once we're happy that this is done.

elParaguayo:
Have squashed everything down but have also realised there is still a bug with `SysTray`.

Now that the colour depth for the bar is not configurable it will almost certainly default to 32 bit depth. This means that the icons will always appear with a transparent background, even if the bar is fully opaque. So, there will always be rendering issues except for when the user has a fully transparent bar.

We can only get the solid background colour behind the icon where the bar is 24 bit.

I expect this could annoy a few people. Maybe we could add a `disable_transparency` option in the bar which is `False` by default but, when `True`, it creates a 24 bit drawer. Thoughts?

ramnes:
> We can only get the solid background colour behind the icon where the bar is 24 bit.

Why?

> I expect this could annoy a few people. Maybe we could add a `disable_transparency` option in the bar which is `False` by default but, when `True`, it creates a 24 bit drawer. Thoughts?

Such a flag doesn't make sense to me, it should just work as expected. Could we not just use 24 bits if the background color doesn't have an alpha?

elParaguayo:
> > We can only get the solid background colour behind the icon where the bar is 24 bit.
> 
> Why?

Good question! The way it worked in the past was by putting the icon in a windon and setting the `backpixmap` of that window to be the same as the bar/widget background. This doesn't work with 32 bit windows and we'd need to use compositing features of xcb which seem incredibly complicated and other WMs have not looked to implement either.

I think other implementations may have relied on pseudo-transparency i.e. accessing the root pixmap beneath the window overlaying on top of this, rather than using 32 bit windows. I didn't know this when I started...

> > I expect this could annoy a few people. Maybe we could add a `disable_transparency` option in the bar which is `False` by default but, when `True`, it creates a 24 bit drawer. Thoughts?
> 
> Such a flag doesn't make sense to me, it should just work as expected. Could we not just use 24 bits if the background color doesn't have an alpha?

That's one approach, yes. I was trying to think through if there were any scenarios where that wouldn't work. The only time I can think if is if someone sets the bar background to a solid colour but also sets a widget background to a transparent/semi-transparent background. Any attempt to do transparency at the widget level would not render as expected if the bar is 24-bit. However, this does seem to be a fairly remote case so I'm happy to try on that basis.

ramnes:
> That's one approach, yes. I was trying to think through if there were any scenarios where that wouldn't work. The only time I can think if is if someone sets the bar background to a solid colour but also sets a widget background to a transparent/semi-transparent background. Any attempt to do transparency at the widget level would not render as expected if the bar is 24-bit. However, this does seem to be a fairly remote case so I'm happy to try on that basis.

Yes, I had this scenario in mind as well.

We should probably:
* add this sentence, shamelessly adapted from i3's documentation, to our own docs: "Please note that due to the way the tray specification works, any transparent background color will cause all tray icons to have a transparent background."
* throw a `logger.warning` if we detect both a transparent background color in the bar and a solid background color for the systray widget.

The combination of the two should be good enough to prevent people thinking that's a bug (I rather see that as a complexity we don't want to add), or at least give us a good reference point in case they open an issue. :)

m-col:
To do this, would the implementation be:  `Bar`, when `qtile.core.name == "x11"`, checks the `background` attribute of itself and all widgets and if none of them have transparency to request a 24 bit window? If any of them _do_ have transparency, carry on with the default behaviour of using a 32 bit window. Is that right?

elParaguayo:
That is one way but I really don't like it as it's incredibly messy.

The problem is that the depths and visuals are established when the `Screen`s are initialised. However, the bar hasn't been initialised at that point so we'd need to set up both 32 and 24 bit. This is basically where we were at the beginning of this PR but we've since unwound the need for having multiple depths.

The approach I'm looking at now is a bit more crude: read the config before setting up X connection and pass a variable saying if we want 24 bit or 32 bit depth.

This is simpler to implement but I'm just having a couple of issues reading the config at the moment...

m-col:
> read the config before setting up X connection and pass a variable saying if we want 24 bit or 32 bit depth.

Hmm I think this sounds precarious and like it might lead to some problems. We don't know what things people will be doing in their configs, and all of that code would be executed. The obvious one being that the config is read when the X connection is established - any code that depends on that will break. Another big one would be anyone doing anything with asyncio; currently the config is read for the first time after the event loop has started, and this changes sounds like it'd be executing that code before it has started, potentially causing lots of issues.

> The problem is that the depths and visuals are established when the Screens are initialised.

What does it mean for them to be established?

elParaguayo:
Fair enough. I'll ditch that approach - sounds like a nasty idea. Got a new, better idea, which is working.

elParaguayo:
Can someone with multiple screens test this latest commit for me. Ideally I want a solid bar on one screen and transparent on another? Try a `SysTray` on each screen too.

My fear is that this will break because references to depths and visual are linked to the `default_screen`.

libqtile/backend/wayland/drawer.py
```diff
@@ -55,13 +61,23 @@ def draw(
...
-        # list of operations. If any of the bounds are not 0.0 then the surface
-        # has operations and we should paint them.
-        if not any(not math.isclose(0.0, i) for i in self.surface.ink_extents()):
+        # We can't test for the surface's ink_extents here on its own as a completely
+        # transparent background would not show any extents but we may still need to
+        # redraw (e.g. if a Spacer widget has changed position and/or size)
+
+        # Check if current rect has changed since previous draw
+        current_rec = (offsetx, offsety, width, height)
+        rect_changed = any([p != c for p, c in zip(self.previous_rect, current_rec)])
```
m-col, line 70: You can just do `rect_changed = self.previous_rect == current_rec`

elParaguayo: what the hell was I thinking when I wrote that line?

elParaguayo:
Still a couple of bugs to iron out. But we're close now!

m-col:
Reports from testing with combinations of: `'#ff0000'`, `'#ff000000'`, `'#ff0000ff'`, `'#ff000033'`:

- [x] single screen X
- [x] double screen X
- [x] single screen Wayland
- [ ] double screen Wayland

Double screen Wayland gives me weird artifacts on the second screen when the backgrounds are '#ff000033' and '#ff000000': 
20210618_11h36m17s_grim

elParaguayo:
Interesting. Thanks for reporting.

Are those widgets `Mirror` objects? I could see that being a problem if you try to mirror an object in both an opaque bar and a transparent one but I would have thought that's more problematic on x11 due to the different colour depths.

I suspect this is something to do with how the surfaces are cleared in the `draw` method.

Also, did you get to the bottom of your groupbox and graph issues? I couldn't replicate them.

m-col:
Yes! They are all mirror widgets, and the groupbox which works fine is not.

> Also, did you get to the bottom of your groupbox and graph issues? I couldn't replicate them.

Actually no but I can't replicate it myself now so gonna assume I made some config or git-related error that caused it.

elParaguayo:
OK - let me look closer at how the `Mirror`s work and see if there's something that can be done here. 

I'm surprised x11 was ok here. I would have thought we need some logic to only mirror if the bars have the same depth (although that's not relevant for wayland). but there may be a better approach here.

libqtile/backend/wayland/drawer.py
```diff
@@ -90,8 +90,11 @@ def draw(
 
         # Paint RecordingSurface operations our window's ImageSurface
         with cairocffi.Context(self._source) as context:
+            context.save()
+            context.set_operator(cairocffi.OPERATOR_SOURCE)
```
elParaguayo, line 94: By using `OPERATOR_SOURCE` in the `drawer.draw` method, we overwrite any existing contents in the drawer. This should stop the overlay/alpha blending issues seen in the screenshot above.

I'm not aware of any instance when we actually want `drawer.draw` to overlay previously rendered contents.

libqtile/bar.py
```diff
@@ -249,7 +249,15 @@ def _configure(self, qtile, screen):
             self.crashed_widgets = []
             for idx, i in enumerate(self.widgets):
                 if i.configured:
-                    i = i.create_mirror()
+                    # To prevent rendering issues, we only create a mirror if
+                    # the bar backgrounds are identical
+                    if i.bar.background == self.background:
```
elParaguayo, line 254: This limits creation of `Mirror` instances to situations where the bar colour is identical, otherwise the widget is cloned.

elParaguayo: This is a bit of a simplification. Arguably, we could mirror if the widget has set it's own `background` colour as this suggests the user wants to override the bar background. However, I think we'd have issues with this if the mirror has a (semi-)transparent background on a solid (24 bit bar). The conditional testing starts to get a bit complicated at that point!

ramnes: > However, I think we'd have issues with this if the mirror has a (semi-)transparent background on a solid (24 bit bar).

How can it happen if the widget background overrides the bar background?

elParaguayo: I was worried that a mirror could have a 32 bit drawer but be drawn on a 24 bit bar, but maybe that's not actually possible. Let me test it.

elParaguayo: I hadn't appreciate the mirror creates a new `drawer` when it's configured. That `drawer` will be linked to the parent bar so the colour depth issue will be handled at that point.

elParaguayo:
Re `Mirror`s: @m-col and I were discussing this on chat. We both agree that, ideally, a mirror without a `background` should display it's own `bar.background`.

This is actually fairly complicated to do, based on how the `drawer` works. Currently, everything is painted to the `drawer`, including any `background`. So, when the mirror tries to copy the `drawer` it picks up the source `background`.

One approach would be to paint the `background` direct to the relevant `surface` and then keep the widget contents in the `RecordingSurface` and have the mirror copy this which it can then draw over it's own `background`. The difficult here is that the `RecordingSurface` is cleared after every `draw()` so the mirror has nothing to copy. You could fix this by not clearing the `RecotdingSurface` but then that's going to continue filling up endlessly which doesn't sound efficient to me.

Typing this out has given me an idea though...

elParaguayo:
Almost got the `Mirror` working as above. It works fine on wayland but has a weird bug on x11: The background is getting cleared but the contents is not immediately drawn. So, if you mirror a `Clock` the text appears after 1 second (i.e. after the first refresh) but if you mirror a static `TextBox` the text is never drawn.

Can't immediately see where I've gone wrong and why it's working on wayland...

@m-col would be good if you could take a look at some point.

elParaguayo:
OK, think this is fixed. The issue is do with the way the hook works in the `Mirror` and how this interacts with the timing of the drawing of each widget during configuration of the bar.

We can fix this by making sure the "parent" is always drawn before the `Mirror` is drawn.

docs/manual/config/screens.rst
```diff
@@ -38,6 +38,17 @@ background="#000000")` will give you a black back ground (the default), while
...
+desired color. For example, :code:`bar.Bar(..., background="#00000000")` will
+result in a fully transparent bar. Widget contents will not be impacted i.e.
+this is different to the ``opacity`` parameter which sets the transparency of the
+entire window.
+
+.. note::
+    In X11 backends, transparency will be disabled in a bar if the ``background``
+    color is fully opaque.
+
+
```
m-col, line 51: Maybe add a note about Systray?

elParaguayo: I added a SysTray comment in its own docstring so it's visible on the Widgets page.

m-col:
The only issue I have is that the very first draw of mirror widgets (or, at least, mirrors of the CPUGraph widget) appear as black and then subsequent draws show what they are meant to. It's a small issue, but fixing it would make this "complete" IMO. If fixing that is not straight forward, I'd rather this PR be in the next rlease as is than not at all.

elParaguayo:
Great - let me take a look at that.

libqtile/backend/wayland/drawer.py
```diff
@@ -50,13 +56,23 @@ def draw(
...
+        # We can't test for the surface's ink_extents here on its own as a completely
+        # transparent background would not show any extents but we may still need to
+        # redraw (e.g. if a Spacer widget has changed position and/or size)
+
+        # Check if current rect has changed since previous draw
+        current_rect = (offsetx, offsety, width, height)
+        rect_changed = current_rect != self.previous_rect
+
+        # Check if draw has content (would be False for completely transparent drawer)
+        ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
```
m-col: It's interesting that `ink_extents` doesn't take into account transparent painted things. I'm guessing you checked this before doing it this way? I'd have expected them to be handled in the same way as things without 100% transparency.

elParaguayo: Correct. Anything with 100% transparency doesn't register with `ink_extents()`. However, I think that makes sense as, if it's fully transparent, there's nothing to draw.

m-col (approved):
Great! Everything looks good to me, so if there is nothing left you had in mind to change then I guess this is ready for merge :tada:

elParaguayo:
It's light on tests (just checking some additional utilities) but, other than that, there's nothing more that I think needs to be added here.

PR #2515, opened 2021-06-07
Don't do anything involving groups with Static windows and add Wayland Window.cmd_static

m-col:
Static windows shouldn't be treated the same as regular Windows in
places that involve groups, as the former is not bound to a group.

This fixes a bug where focus is unexpectedly taken from a focussed
scratchpad window when a Static window is destroyed.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -550,15 +550,10 @@ class Qtile(CommandObject):
         # mypy can't work out that the new tuple is also length 4 (see mypy #7509)
         self.reserve_space(tuple(-i for i in reserved_space), screen)  # type: ignore
 
-    def map_window(self, win: base.WindowType) -> None:
-        self.manage(win)
-        if not win.group or not win.group.screen:
-            return
-        win.unhide()
-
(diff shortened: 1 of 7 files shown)


m-col:
A few pushes there but hopefully mypy is pleased. This cleans up some of the static window code and adds cmd_static to the wayland backend so that regular windows can be made into static windows.

libqtile/core/manager.py
```diff
@@ -550,15 +550,10 @@ def free_reserved_space(
         # mypy can't work out that the new tuple is also length 4 (see mypy #7509)
         self.reserve_space(tuple(-i for i in reserved_space), screen)  # type: ignore
 
-    def map_window(self, win: base.WindowType) -> None:
-        self.manage(win)
-        if not win.group or not win.group.screen:
-            return
-        win.unhide()
```
m-col, line 557: i should clarify this has been (re)moved because it was only used in one place, by Core.handle_MapNotify, so as part of the mypy-related changes the logic has just moved to that method.

PR #2528, opened 2021-06-09
Some multi-monitor fixes

m-col:
Three multi-monitor related commits, see commit messages.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -171,6 +171,7 @@ class Bar(Gap, configurable.Configurable):
         self.cursor_in = None
         self.window = None
         self.size_calculated = 0
+        self._configured = False
 
         self.queued_draws = 0
 
@@ -208,19 +209,17 @@ class Bar(Gap, configurable.Configurable):
             w._test_orientation_compatibility(self.horizontal)
 
         if self.window:
+            # We get _configure()-ed with an existing window when screens are getting
+            # reconfigured but this screen is present both before and after
             self.window.place(self.x, self.y, self.width, self.height, 0, None)
-            self.crashed_widgets = []
-            for i in self.widgets:
-                self._configure_widget(i)
-
-            self._remove_crashed_widgets()
-            self.draw()
-
         else:
+            # Whereas we won't have a window if we're startup up for the first time or
+            # the window has been killed by us no longer using the bar's screen
             self.window = self.qtile.core.create_internal(
                 self.x, self.y, self.width, self.height,
             )
             self.window.opacity = self.opacity
+            self.window.unhide()
 
             self.drawer = self.window.create_drawer(self.width, self.height)
             self.drawer.clear(self.background)
@@ -231,9 +230,12 @@ class Bar(Gap, configurable.Configurable):
             self.window.process_pointer_enter = self.process_pointer_enter
             self.window.process_pointer_leave = self.process_pointer_leave
             self.window.process_pointer_motion = self.process_pointer_motion
-            self.window.unhide()
 
-            self.crashed_widgets = []
+        self.crashed_widgets = []
+        if self._configured:
+            for i in self.widgets:
+                self._configure_widget(i)
+        else:
             for idx, i in enumerate(self.widgets):
                 if i.configured:
                     i = i.create_mirror()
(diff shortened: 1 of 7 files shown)


m-col:
Rebased as master got a conflict. Anyone happy to review this?

ramnes:
Wayland commits LGTM, but I'm not sure about fd2ea0b0511f35525e0704b03c7daee1bcb6fc4a. Shouldn't we just delete the bar instead?

m-col:
Sounds good - the `Bar` itself will stick around in `qtile.config` but I've updated that commit so that the window and its `Drawer` are killed completely. These will then be regenerated when/if the `Screen` is used again.

libqtile/bar.py
```diff
@@ -275,6 +278,12 @@ def _remove_crashed_widgets(self):
     def finalize(self):
         self.drawer.finalize()
 
+    def kill_window(self):
+        """Kill the window when the bar's screen is no longer being used."""
+        self.drawer.finalize()
+        self.window.kill()
+        self.window = None
```
ramnes, line 285: Might be a dumb question, but why not put these in `finalize` and call that instead?

libqtile/bar.py
```diff
@@ -242,9 +243,11 @@ def _configure(self, qtile, screen):
                 if success:
                     qtile.register_widget(i)
 
-            self._remove_crashed_widgets()
-
+        self._remove_crashed_widgets()
+        self.window.unhide()
```
ramnes, line 247: Why do we still need this?

libqtile/core/manager.py
```diff
@@ -312,6 +311,12 @@ def _process_screens(self) -> None:
             scr._configure(self, i, x, y, w, h, grp)
             screens.append(scr)
 
+        for screen in self.screens:
+            if screen not in screens:
+                for gap in screen.gaps:
+                    if isinstance(gap, bar.Bar) and gap.window:
+                        gap.kill_window()
+
```
ramnes, line 319: Isn't ref counting supposed to remove the screens and associated bar if a screen is not in use anymore? Do we still hold a reference to unused screens, and if so, why?

ramnes (approved):
LGTM overall, I just have the weird feeling that it may be simpler if we don't keep references to stuff we don't use but I'm not the one writing the code and I guess I'm missing something completely, so feel free to merge. :D

libqtile/core/manager.py
m-col: No because the screens and bars persist within `qtile.config` and so rather than being made and destroyed they are applied or removes to physical outputs. The `Internal` window itself however can be destroyed and recreated.

libqtile/bar.py
m-col: We only need it with a newly created window, so I've just pushed it up to right after it is made.

libqtile/bar.py
m-col: Finalize is currently only for very final finalising, rather than shutting things down. Although it'd be easy to argue "well we don't need both of those functionalites here, so why not just use finalize?" but this functionality is currently needed in another PR https://github.com/qtile/qtile/pull/2360/commits/8c231cbd44d8d527623415718ccb1531fbd83384#diff-effdc1f9379ed88ce401e42f5852cc8db7444a9b885995477d68709069ca4977R276-R280

libqtile/core/manager.py
ramnes: Shouldn't we destroy stuff and reread config when appropriate?

m-col: I don't think so - and this would be a significant change in itself. Think about how one might configure two `Screen`s in their config because they occasionally add a second screen. The second one waits around some of the time, ready to be used, rather than us deleting it right after `_process_screens` if there is only one screen. If we wanted to go down the route of trimming any unnecessary memory usage from `qtile.config` (which might be a good idea eventually), this would require the config-reloading mechanism to be solid and trusted as there are many factors that need to be taken into account (one example being preventing hooks from being subscribed more than once).

ramnes: We're already protected against duplicate hooks, IIRC. :P

Anyway, I see your point. I still think we shouldn't keep references for stuff we don't use, but practicality beats purity so let's move on!

PR #2529, opened 2021-06-09
Fixes for mypy 0.901

m-col:
mypy just updated and we need to update to match it.

1. They no longer ship third party stubs so for three packages
   (dateutil, pytz & pkg_resources) we need to install their respective
   types-* package for the mypy tests. These are added to tox.ini::mypy
   as well as requirements-dev.txt

2. Improvements to type checking found a error; ColorType does not
   include List and so a check for `type(color) == list` is adjusted
   accordingly.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -521,8 +521,8 @@ class Drawer:
     def new_ctx(self):
         return pangocffi.patch_cairo_context(cairocffi.Context(self.surface))
 
-    def set_source_rgb(self, colour: ColorType):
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thought I was going mad - couldn't work out why the tests were failing on my latest push!

m-col:
Yeah looks like a whole bunch of PRs have scary red crosses they don't deserve :smile:

tych0:
Thanks!

ramnes:
thanks :)

PR #2530, opened 2021-06-09
Don't AttributeError when set_group-ing Screen with no group

m-col:
`cmd_reconfigure_screens` can error:

```
    File "/home/mcol/git/qtile/libqtile/core/manager.py", line 325, in cmd_reconfigure_screens
	self._process_screens()
    File "/home/mcol/git/qtile/libqtile/core/manager.py", line 312, in _process_screens
	scr._configure(self, i, x, y, w, h, grp)
    File "/home/mcol/git/qtile/libqtile/config.py", line 290, in _configure
	self.set_group(group)
    File "/home/mcol/git/qtile/libqtile/config.py", line 349, in set_group
	g1 = self.group
AttributeError: 'Screen' object has no attribute 'group'
```

This happens if `Screen.set_group` is called when that screen has no
group, as rather than the group being `None` it is absent if the screen
has no group, and we need to handle that slightly differently.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -346,13 +346,16 @@ class Screen(CommandObject):
             # g2 (new_group) <-> s2 to
             # g1 <-> s2
             # g2 <-> s1
-            g1 = self.group
+            g1 = getattr(self, "group", None)
             s1 = self
             g2 = new_group
             s2 = new_group.screen
 
-            s2.group = g1
-            g1._set_screen(s2)
+            if g1 is None:
+                del s2.group
+            else:
+                s2.group = g1
+                g1._set_screen(s2)
             s1.group = g2
             g2._set_screen(s1)
         else:


ramnes:
How can a screen not have groups? Shouldn't we fix this instead?

m-col:
As of 396b71e1d00840542d8491a6587dec3fc1f940fd `Screen.group` has type `Group` rather than `Optional[Group]` which helps us avoid a lot of `is not None` checks to please mypy. The caveat of this is that we cannot make it `None`, so if a screen doesn't have a group (i.e. very early in setup, or if that `Screen` instance is not assigned an output e.g. if I declare 2 `Screen`s in my config but only have one output) then it actually doesn't have that attribute (yet).

This approach - having or not having a `Group` rather than always having a `Optional[Group]` feels kinda hacky but the case where the screen has no group only needs to be handled within `Screen` (`Qtile` and other parts of the code will never try to access the group of a screen that isn't in use), and I guess I prefer this over having to check/assert for `None` everywhere `screen.group` is accessed. I guess I don't like either solution :laughing:

m-col:
Closed as this is contained with  #2360

PR #2537, opened 2021-06-11
Build and push packages for tagged releases

flacjacket:
When creating tags and publishing releases, push packages up to PyPI.
When creating a tag, the packages are pushed to Test PyPI, which allows
for the release process to be checked and the packages to be verified
before tagging the release in GitHub.  If any changes are needed, they
can be adjusted by changing the tag, which will re-run the push to Test
PyPI.  The action to do the push to PyPI is triggered by authoring a
release on GitHub.

This adds the functionality to push not only the source release, but
also to build and push wheels for CPython versions 3.7 through 3.10 as
well as PyPy 3.7.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,162 @@
+name: Publish to PyPI
+on:
+  push:
+    branches-ignore: ['*']
+    tags: ['*']
+  release:
+    types: [published]
+jobs:
+  build-wheel:
+    name: Build wheels
+    runs-on: ubuntu-latest
+    container: quay.io/pypa/manylinux_2_24_x86_64
+    strategy:
+      matrix:
+        python-version: ["cp3.7", "cp3.8", "cp3.9", "cp3.10", "pp3.7"]
+    steps:
+      - name: Install dependencies
+        run: |
+          apt update
+          apt-get install -y --no-install-recommends \
+            libcairo-dev \
+            libffi-dev \
+            libpulse-dev
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


flacjacket:
This new actions are not going to run here until we make a new tag, but on my fork, I made some local tweaks to run this on the test pypi instance. That run can be seen here: https://github.com/flacjacket/qtile/actions/runs/928498672

tych0:
For my own edification, the secret.* namespace comes from manual config in the GitHub repo, but how did you generate that secret?

tych0:
And this makes it so the release process is effectively "push a new tag"? Does it need to be annotated?

flacjacket:
Right, the secrets are generated on pypi.org and test.pypi.org. Those are API tokens generated from the "account settings" page, you can create a token with permissions scoped to be able to publish qtile packages. I already generated and saved the secrets we'd need for this on this repo on the secrets page https://github.com/qtile/qtile/settings/secrets/actions. I couldn't find any of the current maintainers with test.pypi.org accounts, I'm happy to add people to that if you let me know your username there (it is separate from the normal pypi accounts, but can be the same username).

Yeah, exactly, pushing a tag will effectively do a test run of the package build and push to test.pypi.org. What actually generates the package is when you create a github release (just through the standard https://github.com/qtile/qtile/releases), the package will be built and pushed to the normal pypi. I think the way we've tagged and created the releases so far is compatible with this and wouldn't need any particular changes.

tych0:
Excellent, this seems like a much better process. Thanks!

PR #2539, opened 2021-06-11
Some floating improvements

m-col:
1. Implement Window.is_transient for for wayland windows. This method is changed slightly to return the parent `Window` rather than just that window's WID (in fact the base class signature already said this, but the x11 implementation just return WID... just goes to show what mypy might catch if the x11 backend was typed). This change is made to reduce a couple lines of extra code if the backend needs to get the Window anyway. This change needs the next pywlroots release from yesterday.
2. Adds wlr_data_control_manager_v1. This is unrelated to floating but i added it here because it also needs the new pywlroots and so the PRs would conflict :smile: 
3. Fix a docstring error in group_window_add hook
4. Unify float_x and float_y usage between backends and clean it up a bit

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -201,6 +201,11 @@ class _Window(CommandObject, metaclass=ABCMeta):
 
 class Window(_Window, metaclass=ABCMeta):
     """A regular Window belonging to a client."""
+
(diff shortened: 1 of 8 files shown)


ramnes:
Nice

PR #2548, opened 2021-06-13
WL: Make input devices configurable

m-col:
This adds a small CFFI wrapper around functions from libinput that can
be used to configure input devices. When input devices are detected by
wlroots, user configuration is checked for a configuration matching the
device, and if one is found then it is applied to that device. The
config API is largely based on how Sway does it: a set of options is
assigned to either all devices, a device type (e.g. pointer), or a
specific device model based on its unique identifier. The most specific
set of options to match a device is applied to it. This makes it quite
straight forward to have matching configuration between Sway and Qtile,
and any other compositor that might follow Sway's lead.

Example use in a config:

```py

wayland_libinput_config = [
    "*": InputConfig(left_handed=True),
    "type:pointer": InputConfig(left_handed=True, tap=True),
    "1267:12377:ELAN1300:00 04F3:3059 Touchpad"": InputConfig(left_handed=True, tap=True, dwt=True),
]
```

This would make this ELAN touchpad model take the third set of options
(only; only one matching `InputConfig` is applied). Other pointer
devices would take the second set of options, and all other
libinput-handled devices would take the wildcard set of options.

The model-specific identifier is determined the same way as Sway:
`"<vendor>:<product>:<name>"`. The device-type matching takes the form
`"type:<device type>"`. Details about the configuration options can be
found in Sway's input man page:
https://man.archlinux.org/man/sway-input.5#LIBINPUT_CONFIGURATION

And the libinput docs are also quite useful:
https://wayland.freedesktop.org/libinput/doc/latest/configuration.html

-- 

The above is the commit message with everything at its current form, but I'm very open to suggestions on alternative ways expose the configuration of devices to users as well as how everything is implemented (particularly the cffi module; it's the first one I've written).


m-col:
The CI is failing when trying to run `./scripts/ffibuild` during setup:
```
      import wlroots.ffi_build as wlr
    File "/home/runner/work/qtile/qtile/.tox/py38/lib/python3.8/site-packages/wlroots/__init__.py", line 5, in <module>
      from ._ffi import ffi, lib  # noqa: F401
  ImportError: /home/runner/work/qtile/qtile/.tox/py38/lib/python3.8/site-packages/wlroots/_ffi.abi3.so: undefined symbol: wlr_renderer_get_shm_texture_formats
  ERROR: InvocationError for command /home/runner/work/qtile/qtile/.tox/py38/bin/python3 setup.py install (exited with code 1)
```
The libinput ffibuild script is trying to import `wlroots.ffi_build` and getting this undefined symbol error. Oddly, I cannot reproduce locally. The function `wlr_renderer_get_shm_texture_formats` is present in the wlroots and pywlroots used (I'm using the same versions locally as the CI). Anyone got any ideas what might cause this error? I

ramnes:
I'm not sure I like it inside Qtile. I guess it's just how Wayland works and that it expects compositors to handle inputs configuration? Would there be any alternative, i.e. any way to configure libinput outside Qtile?

m-col:
> I'm not sure I like it inside Qtile. 

Same :laughing: 

> I guess it's just how Wayland works and that it expects compositors to handle inputs configuration? 

Pretty much - compositors are responsible for everything that the X server is responsible for, including input configuration and keymaps etc.

Ideally there would be some kind of common mechanism for compositors (or at least wlroots-based compositors) to configure input devices and share configuration across compositors. IMO this would be great to have within wlroots itself, or perhaps if they had a basic protocol for this that compositors could use with a little boilerplate, and then some clients (user's choice) to do the actual configuration. I remember seeing that idea shot down in swaywm's issues but alas now I can't find that issue (and there are hundreds matching "libinput"). It's possible that this appears in the future, but at the minute the only way to configure inputs is for the compositor to do it :(

ramnes:
What about another tool / process that Qtile would run at startup and that would read its own configuration file?

m-col:
Unfortunately other processes won't be able to set options on the input devices. Currently there is no standard for the configuration of input devices, and consequently no protocol that can be taken and trivially used from wlroots to configure input devices as they get detected and handled by the compositor. ATM it seems as though all compositors, if they want to support configuration of these devices, must do it themselves. It seems as though one or two people have in the last few years tried to put together some kind of standard and/or protocol that would maybe make it feasible for wlroots to have some kind of plug-n-play libinput config manager, but unfortunately this seems to be one thing that people are not quickly agreeing on, and so we're here.

m-col:
That's not to say something can't come along in the next few years that will allow us to replace all of this with a couple lines of wlroots/some-other-lib code!

jwijenbergh:
> OK, sounds good. In that case, let's try to ensure the code is as minimal as it can be to reduce that load, especially the way it is configured. I was thinking to change the configuration variable to `wayland_libinput_rules` to somewhat follow `floating_rules` but that's a trivial cosmetic change.
> 
> > Unless some kind of wayland protocol is made that will be supported by wlroots (don't think there is one right now?).
> 
> Currently there is no such protocol. I think the reason why there is not one is because it is specific to libinput and most of the existing protocols try to abstract things at a level higher than the implementation details. That said, pretty much every single (desktop) compositor will be using libinput and therefore it's lots of unnecessary code duplication _and_ user config duplication without one, so there is certainly a motive for developing a protocol. When I discussed this on the wlroots IRC channel they said that if one were to be proposed they would welcome it and review it, but it is currently low priority so unlikely for them to work on it themselves. I do hold out hope that in the future we can remove all of this though!
> 
> > Also, is there a way to configure keyboard repeat delay and repeat rate,
> 
> This can't be configured yet but of course it could be. Perhaps exposing it as a cmd_ on `Keyboard` would be enough, like the keyboard layouts?

I forgot to reply :sweat_smile:. Thanks for the background info on the wlroots state of things of such a protocol!

> This can't be configured yet but of course it could be. Perhaps exposing it as a cmd_ on `Keyboard` would be enough, like the keyboard layouts?

That would work for me!

m-col:
I reworked it a bit and I think it's bit tidier. The existing solution for configuring keyboards is suboptimal because it not stateless. You currently need to set xkbcommon environmental variables to set defaults and then use `core.cmd_set_keymap` to change it. It made sense to include keyboards here and make it a more generic solution for input devices. This actually made a lot of sense on the wayland side because that's actually how wlroots does it... we receive generic input devices when connected to the seat and we then determine whether they are keyboards or pointers.

Anyway, this should now allow for configuration of libinput-managed pointer devices, as well as setting keyboard layout, option, variant, and key repeat rate and delay, all from the config. The config variable is `wl_input_rules`. Any feedback and live-testing appreciated!

m-col:
@elParaguayo the docs seem to be failing as wlroots/pywlroots are not present in the build system, and as sphinx wants to read from all modules including `libqtile.backend.wayland.inputs` it's failing as the libinput CFFI module is not getting build. Is there a way around this, or if not, is it possible to just install wlroots on the system?

elParaguayo:
We don't include any `wlroots` in the docs build. All the relevant modules need to be inlcuded in `MOCK_MODULES` section in `docs/conf.py`:
https://github.com/qtile/qtile/blob/7f451f901f92bdeca2ce186e3e3f3054c35e832f/docs/conf.py#L27-L75

Would that work here or is there something special about `libqtile.backend.wayland.inputs`?

m-col:
Maybe! I've added that change so we'll see in a few mins

elParaguayo:
The docs error message is about `libqtile.backend.wayland.input` not `inputs`.

elParaguayo:
Let's take this to IRC so as not to pollute this thread

m-col:
The `packaging` CI is failing:
```
    File "libqtile/backend/wayland/libinput_ffi_build.py", line 1, in <module>
      import wlroots.ffi_build as wlr
  ModuleNotFoundError: No module named 'wlroots'
```
Having a build-type dependency of wlroots is not ideal but I wouldn't want to make it a hard dependency. I'm thinking that the best approach is to do what `cairocffi` does and build with libinput support if wlroots is available at build time, but otherwise build successfully when wlroots is not available. If users want to configure libinput devices in the wayland backend then they will need to install pywlroots first, and then qtile. Any thoughts? Maybe there is a better way?

elParaguayo:
Yeah, that's the approach I was thinking about when this first stopped the docs building.

Seems sensible to me.

m-col:
OK this is all building just fine now. If `wlroots` is not importable during build then the libinput CFFI module is not built. If this happens, libinput configuration can be handled, but the wayland backend will still run if you later install the wayland deps. 

A last once-over before merging perhaps?

setup.py
```diff
@@ -85,6 +85,13 @@ def get_cffi_modules():
             cffi_modules.append(
                 'libqtile/widget/pulseaudio_ffi.py:pulseaudio_ffi'
             )
+    try:
+        import wlroots.ffi_build
+        cffi_modules.append(
+            'libqtile/backend/wayland/libinput_ffi_build.py:libinput_ffi',
+        )
+    except ImportError:
+        pass
```
elParaguayo, line 94: Do you want to include a message here (similar to pulse) that libinput support is not available (similar to the pulseaudio message)?

m-col: Good idea - added

m-col:
Ah apologies I pushed some output fixes to the wrong branch. They're all small and in the wayland theme so I'll just leave them. Hope that's alright. The libinput commit hasn't changed.

m-col:
Scratch that, it just made it complicated. I've split them out into a different PR: https://github.com/qtile/qtile/pull/3320

dr-carlos:
Is there any progress on implementing this PR?

m-col:
The PR is pretty much finished, but my worry is the changes it makes to the build logic. I'll give this a good testing today to be extra sure that there are no problems with building and installing, and if all goes well I will merge it.

PR #2553, opened 2021-06-14
Update widget keyboard grabbing and prompt widget to be backend-agnostic

m-col:
This updates the code that handles keyboard-grabbing by widgets (and Internal windows in general) so that any backend can do it. The `Bar` and prompt widget are updated to use this approach, and the prompt widget uses a non-x11 specific way to ask for the keysyms to which it binds its methods when it has grabbed the keyboard.

Everything appears to be working as expected under both Wayland and X11 backends, but I'm not a user of the prompt widget (though I've been considering adding it to my bar recently!) so if anyone uses it lots maybe they could give it a test run for a day or so just to check it works as before?

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -109,6 +109,10 @@ class Core(metaclass=ABCMeta):
     def simulate_keypress(self, modifiers: List[str], key: str) -> None:
         """Simulate a keypress with given modifiers"""
 
+    def keysym_from_name(self, name: str) -> int:
+        """Get the keysym for a key from its name"""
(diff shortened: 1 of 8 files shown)


m-col:
Can confirm this is working perfectly under both Wayland and X.

elParaguayo:
I'm seeing some slightly odd behaviour where the alt key is not being respected. For exampled `alt`+`.` in a terminal would give me the last argument from the previous command and `alt` + number gives a different screen in `irssi`.

I've seen it in a scratchpad where the window is visible but doesn't have focus.

PR #2519, opened 2021-06-09
Add paru support for CheckUpdates widget

ZaheenJ:
[Paru](https://github.com/Morganamilo/paru) is another AUR helper. This pull request adds Arch_paru and Arch_paru_Sup which do what Arch and Arch_Sup do respectively. I am also unsure why Arch_Sup has 1 line being removed:
```
https://archlinux.thaller.ws/extra/os/x86_64/alsa-card-profiles-1:0.3.30-1-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/extra/os/x86_64/aom-3.1.1-1-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/core/os/x86_64/libcap-2.50-2-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/extra/os/x86_64/pipewire-1:0.3.30-1-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/extra/os/x86_64/pipewire-media-session-1:0.3.30-1-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/extra/os/x86_64/pipewire-alsa-1:0.3.30-1-x86_64.pkg.tar.zst
https://archlinux.thaller.ws/extra/os/x86_64/pipewire-pulse-1:0.3.30-1-x86_64.pkg.tar.zst
```
this is the output of `pacman -Sup` which clearly shows 7 packages that need to be updated but the widget only shows 6.

PR #2559, opened 2021-06-18
Add paru support in CheckUpdates widget

ZaheenJ:
This is actually the same pull request as #2519, I just don't know how to squash commits, now there are only two, is that good enough?


ramnes:
Yep, thanks!

PR #2561, opened 2021-06-19
Forward unhandled keybindings to client window

### LINKED ISSUE #324: Pass on ignored key events to clients
Currently, you can make a keyboard shortcut layout specific by calling .when(layout_name) on it. 

However, qtile still swallows the keypress even when the layout does not match (it call self.root.grab_key for all registered key codes). It would be nice if the keypress could be passed on to the focused client if no handlers handle it.

This would make it possible to rebind commonly used keys for specific layouts that are used with particular applications, without affecting other layouts/applications.

Two possible implementations (I'm no Xlib expert so I could be talking nonsense):

1) Ungrab layout specifc keys codes whenever a layout that oes not match is used

2) If no handler claims the key press, resend it to the client application, in a manner similar to xte.

m-col:
Keybindings can be conditional, e.g. upon the current layout:

```
    lazy.function(func).when(layout='max')
```

We grab the appropriate keys as normal but we swallow these
unconditionally. Instead, if Qtile does not execute the binding because
the conditions were not met, we should not be swallowing these key press
events.

Fixes #324

### DIFF
--- a/libqtile/backend/wayland/keyboard.py
+++ b/libqtile/backend/wayland/keyboard.py
@@ -105,7 +105,7 @@ class Keyboard(HasListeners):
             mods = self.keyboard.modifier
             for keysym in keysyms:
                 if (keysym, mods) in self.grabbed_keys:
-                    self.qtile.process_key_event(keysym, mods)
-                    return
+                    if self.qtile.process_key_event(keysym, mods):
+                        return
 
         self.seat.keyboard_notify_key(event)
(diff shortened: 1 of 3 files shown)


tych0:
Any chance you can add a test for this? Patch itself lgtm.

m-col:
No luck testing this. The difficulty is with creating fake key press events that  can be forwarded in a usable form to a focussed client. I think on X creating an XTest-based approach might work, but if I have understood the XTest interface of xcffib correctly, this would require further implementation. Unless the FakeInput method works for keys too? On Wayland things appear less likely; there does not appear to be an interface yet in wlroots for simulating keypresses, though there is a TODO in the code referencing it :laughing:

tych0:
Roger that :)

PR #2565, opened 2021-06-19
Fix margin_on_single so it works with [N E S W] coords

mbenford:
Fix Columns layout's `margin_on_single` so it accepts a list of values for each margin, instead of just a single integer for all sides. This is based on how monadtall's `single_margin` works. In fact, I considered renaming `margin_on_single` as `single_margin` so there would be one name for the same behavior across layouts, but it would be a breaking change and I wasn't sure if it would be appropriate. I still can do it, however, if a maintainer green-lights it.

I also tried to add new tests to cover the changes, but I couldn't figure out how to do it. Any pointers would be appreciated.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -120,7 +120,7 @@ class Columns(Layout):
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw a border when there is one only window."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
-        ("margin_on_single", -1, "Margin when only one window. `-1` means use `margin`."),
+        ("margin_on_single", None, "Margin when only one window. (int or list of ints [N E S W])"),
         ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
         ("grow_amount", 10, "Amount by which to grow a window/column."),
@@ -235,7 +235,7 @@ class Columns(Layout):
         if len(self.columns) == 1 and (len(col) == 1 or not col.split):
             if not self.border_on_single:
                 border = 0
-            if self.margin_on_single > -1:
+            if self.margin_on_single is not None:
                 margin_size = self.margin_on_single
         width = int(
             0.5 + col.width * screen_rect.width * 0.01 / len(self.columns))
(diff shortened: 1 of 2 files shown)


m-col:
Cool thanks. I think the name you've used is fine, it's probably best to avoid the config breakages. For the test, something like this will hopefully get you started, I think it would be good to just test that the window has the expected geometry for the 3 different `margon_on_single` forms (None, int, or list), for one and two windows:
```diff
diff --git a/test/layouts/test_columns.py b/test/layouts/test_columns.py
index 1a43062e..9962d8a9 100644
--- a/test/layouts/test_columns.py
+++ b/test/layouts/test_columns.py
@@ -36,6 +36,8 @@ class ColumnsConfig(Config):
     ]
     layouts = [
         layout.Columns(num_columns=3),
+        layout.Columns(margin_on_single=10),
+        layout.Columns(margin_on_single=[1, 2, 3, 4]),
     ]
     floating_layout = libqtile.resources.default_config.floating_layout
     keys = []
@@ -139,3 +141,14 @@ def test_columns_swap_column_right(manager):
     assert columns[0]['clients'] == ['2']
     assert columns[1]['clients'] == ['1']
     assert columns[2]['clients'] == ['4', '3']
+
+
+@columns_config
+def test_columns_margins(manager):
+    manager.test_window("1")
+
+    info = manager.c.window.info()
+    assert info['x'] = 
+    assert info['y'] = 
+    assert info['width'] = 
+    assert info['height'] = 
```

I think I added the parameter and therefore added the "Use -1 to use 'margin'" thing, but now I can't tell why it was like that. I guess if somebody wanted no margins they'd set it to `0`, and `None` would use the `margin` option's value, right?

mbenford:
@m-col Thanks for the pointers. I managed to write some tests covering all cases (`None`, `int` and `list`). 

> I guess if somebody wanted no margins they'd set it to 0, and None would use the margin option's value, right?

That's right.

I also ensured that the old default value (`-1`) would become `None` in order to not break any existing configurations. But since `-1` was the default value and it's almost certain than no one would set it manually, do you think that check is actually necessary?

m-col:
> it's almost certain than no one would set it manually, do you think that check is really necessary?

Yeah you're probably right, I doubt that too, no problem if you want to remove it.

Test looks great, thanks for that. LGTM with or without check.

mbenford:
All checks are failing, and I don't think it has anything to do with my latest changes. I've re-run all jobs a few times but they keep on failing.

ramnes:
Thanks!

PR #2567, opened 2021-06-20
migrate: add a migration_count to config files

### LINKED ISSUE #2353: migrations need version annotations
We have cases where migrations are non-idempotent, so we should only run them when upgrading from a particular version to a different one.

My suggestion is to write a `qtile_version = "$whatever` into users' configs after they've done a migration, and then we can parse that and only run migrations newer than that.

We'll need to annotate the migrations we have somehow, but we can do that reasonably easily (a decorator, make them callable classes, whatever).

Thoughts?

tych0:
Here's some code implementing the migrate_count discussion from #2353.
We write a variable into users' configs with the number of total
migrations performed. This number is structured so that it always
monotonically increases, so we never run the same message twice.

There are two interesting scenarios here: when we add a new migration for
the next version, and when we want to remove old migrations.

The first scenario looks something like this:

```diff
diff --git a/libqtile/scripts/migrate.py b/libqtile/scripts/migrate.py
index 157a2d74..daa3c47a 100644
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -117,6 +117,10 @@ def new_at_current_to_new_client_position(query):
     )

+def fancy_new_migration(query):
+    ...
+
+
 MIGRATIONS = [
     # from 0.17.0
     client_name_updated,
@@ -125,6 +129,9 @@ MIGRATIONS = [
     pacman_to_checkupdates,
     hook_main_function,
     new_at_current_to_new_client_position,
+
+    # from 0.18.0
+    fancy_new_migration,
 ]
```
and the second something like this (though obviously we'd remove all the
migrations from 0.17.0 at the same time, I've just removed one here for
brevity):

```diff
diff --git a/libqtile/scripts/migrate.py b/libqtile/scripts/migrate.py
index 157a2d74..d7aa2924 100644
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -41,11 +41,6 @@ def rename_hook(query, fro, to):
     return q.rename(to)

-def client_name_updated(query):
-    """ Rename window_name_change -> client_name_updated"""
-    return rename_hook(query, "window_name_change", "client_name_updated")
-
-
 def tile_master_windows_rename(query):
     return (
         query
@@ -118,8 +113,6 @@ def new_at_current_to_new_client_position(query):

 MIGRATIONS = [
-    # from 0.17.0
-    client_name_updated,
     tile_master_windows_rename,
     threaded_poll_text_rename,
     pacman_to_checkupdates,
@@ -148,7 +141,7 @@ for (fro, to) in MODULE_RENAMES:

 # Number of migrations that have been removed.
-PREVIOUS_MIGRATIONS = 0
+PREVIOUS_MIGRATIONS = 1

 MIGRATION_COUNT_PATTERN = re.compile(r"^migration_count\s*=\s*(\d+)\s*")
```

It's not clear to me that we necessarily want this complexity, since the
motivating use case from #2353 never arrived, but I'll leave that decision up
to reviewers :)

Closes #2353

### DIFF
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -19,6 +19,7 @@
 import filecmp
 import os
 import os.path
+import re
 import shutil
 import sys
 from glob import glob
@@ -117,6 +118,7 @@ def new_at_current_to_new_client_position(query):
 
 
 MIGRATIONS = [
+    # from 0.17.0
     client_name_updated,
     tile_master_windows_rename,
     threaded_poll_text_rename,
@@ -127,6 +129,7 @@ MIGRATIONS = [
 
 
 MODULE_RENAMES = [
+    # from 0.17.0
     ("libqtile.command_graph", "libqtile.command.graph"),
     ("libqtile.command_client", "libqtile.command.client"),
     ("libqtile.command_interface", "libqtile.command.interface"),
@@ -144,6 +147,47 @@ for (fro, to) in MODULE_RENAMES:
     MIGRATIONS.append(f)
 
 
+# Number of migrations that have been removed.
+PREVIOUS_MIGRATIONS = 0
+
+MIGRATION_COUNT_PATTERN = re.compile(r"^migration_count\s*=\s*(\d+)\s*")
+
+
+def get_migration_count(config):
+    # this is fairly horrible, but there's no guarantee we can __import__() a
+    # config file in the current version, since it may not have been migrated
+    # yet. so we parse it manually :)
+    with open(config) as f:
+        for line in f.readlines():
+            match = MIGRATION_COUNT_PATTERN.fullmatch(line)
+            if match is not None:
+                return int(match.group(1))
+
+    # didn't find it, so the file hasn't been migrated before
+    return 0
+
+
+def set_migration_count(config):
+    new_migration_count = "migration_count = {}".format(PREVIOUS_MIGRATIONS + len(MIGRATIONS))
+    written = False
+    with open(config, "r+") as f:
+        lines = f.readlines()
+        f.seek(0)
+        for line in lines:
+            match = MIGRATION_COUNT_PATTERN.fullmatch(line)
+            if match is not None:
+                written = True
+                f.write(new_migration_count)
+                f.write("\n")
+            else:
+                f.write(line)
+
+    if not written:
+        with open(config, "a") as f:
+            f.write(new_migration_count)
+            f.write("\n")
+
+
 def file_and_backup(config_dir):
     for py in glob(os.path.join(config_dir, "*.py")):
         backup = py + BACKUP_SUFFIX
@@ -156,13 +200,21 @@ def do_migrate(args):
         print("install it and try again")
         sys.exit(1)
 
+    migration_count = get_migration_count(args.config)
+
     config_dir = os.path.dirname(args.config)
     for py, backup in file_and_backup(config_dir):
         shutil.copyfile(py, backup)
 
-    for m in MIGRATIONS:
+    # we only want to do migrations that haven't been done before
+    needed_migrations = MIGRATIONS[migration_count-PREVIOUS_MIGRATIONS:]
+    for m in needed_migrations:
         q = bowler.Query(config_dir)
         m(q).execute(interactive=args.interactive, write=True)
+
+    # if we actually did some migrations, let's update the migration_count
+    if len(needed_migrations) > 0:
+        set_migration_count(args.config)
 
     changed = False
     for py, backup in file_and_backup(config_dir):
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -28,6 +28,7 @@ from sys import exit
 import libqtile.backend
 from libqtile import confreader
 from libqtile.log_utils import logger
+from libqtile.scripts.migrate import set_migration_count
 
 
 def rename_process():
@@ -58,6 +59,7 @@ def make_qtile(options):
                                             "resources",
                                             "default_config.py")
             copyfile(default_config_path, options.configfile)
+            set_migration_count(options.configfile)
             logger.info('Copied default_config.py to %s', options.configfile)
         except Exception as e:
             logger.exception('Failed to copy default_config.py to %s: (%s)',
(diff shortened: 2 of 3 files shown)


m-col:
> It's not clear to me that we necessarily want this complexity, since the
motivating use case from #2353 never arrived, but I'll leave that decision up
to reviewers :)

It would be easy to say maybe we should hold off on something like this until we actually get a migration conflict like in https://github.com/qtile/qtile/pull/2346#issuecomment-812575166, but I guess this is inevitable. Having each migration performed only once can also make the experience a bit more clean.

What migration policy would we set? Clear all migrations following a major release? Minor release? or even 2 releases? If we remove them following a release then we don't need to specify the version from which new ones are added, like in your scenario 1.

On the implementation side - why can't we use a different file that isn't the config? We can put another file in `utils.get_cache_dir()` and track it there, no need to modify or parse the config, and keeps it out of view.

tych0:
> What migration policy would we set? Clear all migrations following a major release? Minor release? or even 2 releases?

I'm not married to any particular policy, keeping them for a couple of releases seems nice but is not strictly necessary IMO. I'm happy to add a comment somewhere in the file if we decide on a policy in this thread.

> On the implementation side - why can't we use a different file that isn't the config?

I worry about users who are keeping their configs in git. How will they know to git add this file and keep it up to date? This is also the reason we can't use something like xattrs, because by default git doesn't track those.

libqtile/scripts/migrate.py
```diff
@@ -144,6 +147,47 @@ def f(query, fro=fro, to=to):
...
 
+# Number of migrations that have been removed.
+PREVIOUS_MIGRATIONS = 0
+
+MIGRATION_COUNT_PATTERN = re.compile(r"^migration_count\s*=\s*(\d+)\s*")
+
+
+def get_migration_count(config):
+    # this is fairly horrible, but there's no guarantee we can __import__() a
+    # config file in the current version, since it may not have been migrated
```
m-col, line 158: Could you make it clear that this woud be due to moved imports within libqtile? I was about to comment saying "why not just make them fix it?" before I remembered the mention of this in the issue.

libqtile/scripts/migrate.py
```diff
@@ -144,6 +147,47 @@ def f(query, fro=fro, to=to):
...
+    with open(config, "r+") as f:
+        lines = f.readlines()
+        f.seek(0)
+        for line in lines:
+            match = MIGRATION_COUNT_PATTERN.fullmatch(line)
+            if match is not None:
+                written = True
+                f.write(new_migration_count)
+                f.write("\n")
+            else:
```
m-col, line 182: What is this doing? Looks like it is prepending the migration count but that's done in the next block after the loop, so I think I'm misunderstanding this

m-col:
I think keeping migrations around for longer is better if it doesn't make things difficult.

> I worry about users who are keeping their configs in git. How will they know to git add this file and keep it up to date? 

Then perhaps we can put it in ~/.config/qtile right beside the config, so that it will be displayed by git status? Otherwise, I'm happy with your current approach.

Reading through it I think the approach is quite sensible :) Maybe add a small section to the docs about all of this?

libqtile/scripts/migrate.py
tych0: There are two cases: 1. the user has an existing `migration_count` that needs to be updated (i.e. the regex matches, but has an old value from the last time we did migrations) which is here. Another case is that a user hasn't used `migrate` before, and `migration_count` doesn't exist in their config file; that's what the case below handles.

libqtile/scripts/migrate.py
tych0: Yes, I'll elaborate a bit here.

tych0:
> Then perhaps we can put it in ~/.config/qtile right beside the config, so that it will be displayed by git status? Otherwise, I'm happy with your current approach.

That would make it less of a nasty regex and more just a number in a file, which is nice. I really just worry that people won't pay attention to that. And if the argument is "it doesn't matter that much", then maybe we should just close this PR and forget about this code entirely, since "it doesn't matter that much" :). In case it's not clear, I'm really on the fence about this... it adds some complexity for questionable value.

I'm happy to switch the code to do whatever we decide, do others have thoughts? @ramnes @elParaguayo ?

elParaguayo:
It's a tricky one. 

I think that, from the discussion in #2353, we're focussing on users who are moving from release to release rather than cloning the latest git version. If that's the case then we need to make this as smooth as we can for them. Having a separate file could introduce another point of failure.

I would think the priority should be to change things that would cause their config to break. For those edge cases (e.g. the example I gave where the migration could cause issues if run twice) I think we should make sure that

1. The migrate function is only run once on any config file (I would keep this in the config file)
2. Any migrate functions that would break configs on a second run either:
    - are removed from migrate.py for the next release (the potential flaw in this is that it assumes people install each release); 
    - include a visual warning for the user as to what changes are/are not made so they can take steps to deal with the relevant part (and then do a `check` afterwards).

The migrate script is a brilliant concept and could help make qtile a bit more accessible but I also don't think we should be afraid to just say where the user needs to take actions themselves if we can't be sure to guarantee the right outcome.

tych0:
So I don't hear a ton of arguments in favor of this approach :)

> Any migrate functions that would break configs on a second run ... are removed from migrate.py for the next release

it really feels like we should either remove all of the migrate stuff from a release, or none of it. I'm not sure selectively removing non-idempotent things will really be useful, because it means people can't trust the output.

There is an interactive mode, and generally `migrate.py` is best effort anyways, so maybe we should just ignore this problem and close this PR?

ramnes:
Let's make the interactive mode default, at least?

I like the feature, and I'm sure we'll be forced to implement it at some point, either from a performance standpoint or because of the case we originally mentioned, although I can't help myself but prefer the approach with another file than the config. All in all I'm +0.5 on this PR. :)

m-col:
I agree with everything @ramnes said too!

tych0:
Ok, I've opened #2592 instead, so I'll close this.

ramnes:
Thank you for taking the time!

tych0:
Sure thing. Sometimes you have to write code to realize how truly awful it's going to end up :)

PR #2574, opened 2021-06-23
Note that the example autostart.sh script should be executable

m-col:
Also removes `feh` because I want people to use our internal wallpaper setting :)

### DIFF
--- a/docs/manual/config/hooks.rst
+++ b/docs/manual/config/hooks.rst
@@ -42,13 +42,13 @@ want to look at the ``startup`` and ``startup_once`` hooks. ``startup`` is
 emitted every time Qtile starts (including restarts), whereas ``startup_once``
 is only emitted on the very first startup.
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Looks good to me and also looks like a good candidate for my first merge!

elParaguayo:
Should I have merged this differently? Changed the title? Also, why doesn't it say you "authored" this one?

tych0:
I try to use fast-forward (aka rebase) merges instead of actual merge commits, since it keeps the history slightly cleaner. Looks like

0c049edd96069a7030c4895e9832711c07bb0bfa

is your merge commit and

e65aa66c4efba30f8ecab42ffbff1b5bfe2aa00a

is the actual commit in the PR from m-col. So the authorship is as I would expect, the only difference between how we usually do it is the merge commit. It's not the end of the world though :)

elParaguayo:
Thanks. Will get it right next time!

PR #2575, opened 2021-06-23
Update link for complete lists of special keys in docs

Syakhisk:
The old one was `https://github.com/qtile/qtile/blob/master/libqtile/xkeysyms.py` which didn't exist now.


m-col:
Ah yep I missed this. Thanks!

PR #2576, opened 2021-06-23
Better handling of incompatible widgets (e.g. Systray on Wayland)

elParaguayo:
Now that we've got a Wayland backend we should remove Systray from the default config as it doesn't run on Wayland.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,9 @@ Qtile x.xx.x, released XXXX-XX-XX:
     * features
         - Add ability to draw borders and add margins to the `Max` layout.
     * bugfixes
+        - Widgets that are incompatible with a backend (e.g. Systray on Wayland) will no longer show 
+          as a ConfigError in the bar. Instead the widget is silently removed from the bar and a message
+          included in the logs.
 
 Qtile 0.21.0, released 2022-03-23:
     * features
--- a/docs/sphinx_qtile.py
+++ b/docs/sphinx_qtile.py
@@ -54,6 +54,10 @@ qtile_class_template = Template('''
     .. compound::
 
         Supported bar orientations: {{ obj.orientations }}
+
+        {% if supported_backends %}
+        Only available on the following backends: {{ ", ".join(obj.supported_backends) }}
+        {% endif %}
     {% endif %}
     {% if is_widget and screen_shots %}
     .. raw:: html
@@ -132,6 +136,7 @@ class QtileClass(SimpleDirectiveMixin, Directive):
         obj = import_class(module, class_name)
         is_configurable = ':no-config:' not in arguments
         is_commandable = ':no-commands:' not in arguments
+        is_widget = issubclass(obj, widget.base._Widget)
         arguments = [i for i in arguments if i not in (':no-config:', ':no-commands:')]
 
         # build up a dict of defaults using reverse MRO
@@ -178,7 +183,8 @@ class QtileClass(SimpleDirectiveMixin, Directive):
             'commandable': is_commandable and issubclass(obj, command.base.CommandObject),
             'is_widget': is_widget,
             'extra_arguments': arguments,
-            'screen_shots': widget_shots
+            'screen_shots': widget_shots,
+            'supported_backends': is_widget and obj.supported_backends
         }
         if context['commandable']:
             context['commands'] = [
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -327,6 +327,15 @@ class Bar(Gap, configurable.Configurable):
 
     def _configure_widget(self, widget):
         configured = True
+
+        if widget.supported_backends and (self.qtile.core.name not in widget.supported_backends):
+            logger.warning(
+                f"Widget removed: {widget.__class__.__name__} does not support "
+                f"{self.qtile.core.name}."
+            )
+            self.crashed_widgets.append(widget)
+            return False
+
         try:
             widget._configure(self.qtile, self)
 
@@ -352,13 +361,19 @@ class Bar(Gap, configurable.Configurable):
 
         for i in self.crashed_widgets:
             index = self.widgets.index(i)
-            crash = ConfigErrorWidget(widget=i)
-            crash._configure(self.qtile, self)
-            if self.horizontal:
-                crash.offsety = self.border_width[0]
-            else:
-                crash.offsetx = self.border_width[3]
-            self.widgets.insert(index, crash)
+
+            # Widgets that aren't available on the current backend should not
+            # be shown as "crashed" as the behaviour is expected. Only notify
+            # for genuine crashes.
+            if not i.supported_backends or (self.qtile.core.name in i.supported_backends):
+                crash = ConfigErrorWidget(widget=i)
+                crash._configure(self.qtile, self)
+                if self.horizontal:
+                    crash.offsety = self.border_width[0]
+                else:
+                    crash.offsetx = self.border_width[3]
+                self.widgets.insert(index, crash)
+
             self.widgets.remove(i)
 
     def _items(self, name: str) -> ItemT:
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -138,6 +138,8 @@ screens = [
                 ),
                 widget.TextBox("default config", name="default"),
                 widget.TextBox("Press &lt;M-r&gt; to spawn", foreground="#d75f5f"),
+                # NB Systray is incompatible with Wayland, consider using StatusNotifier instead
+                # widget.StatusNotifier(),
                 widget.Systray(),
                 widget.Clock(format="%Y-%m-%d %a %I:%M %p"),
                 widget.QuickExit(),
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -131,6 +131,11 @@ class _Widget(CommandObject, configurable.Configurable):
     """
 
     orientations = ORIENTATION_BOTH
+
+    # Default (empty set) is for all backends to be supported. Widgets can override this
+    # to explicitly confirm which backends are supported
+    supported_backends: set[str] = set()
+
     offsetx: int = 0
     offsety: int = 0
     defaults = [
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -118,6 +118,8 @@ class Systray(window._Window, base._Widget):
 
     orientations = base.ORIENTATION_BOTH
 
+    supported_backends = {"x11"}
+
     defaults = [
         ("icon_size", 20, "Icon width"),
         ("padding", 5, "Padding between icons"),
--- a/test/test_bar.py
+++ b/test/test_bar.py
@@ -34,7 +34,7 @@ import libqtile.confreader
 import libqtile.layout
 import libqtile.widget
 from libqtile.command.base import CommandError
-from test.helpers import Retry
+from test.helpers import BareConfig, Retry
 
 
 class GBConfig(libqtile.confreader.Config):
@@ -606,3 +606,22 @@ def test_bar_border_vertical(manager_nospawn):
 
     _, yoffset = manager_nospawn.c.bar["right"].eval("self.widgets[0].offsety")
     assert xoffset == "0"
+
+
+def test_unsupported_widget(manager_nospawn):
+    """Widgets on unsupported backends should be removed silently from the bar."""
+
+    class UnsupportedWidget(libqtile.widget.TextBox):
+        if manager_nospawn.backend.name == "x11":
+            supported_backends = {"wayland"}
+        elif manager_nospawn.backend.name == "wayland":
+            supported_backends = {"x11"}
+        else:
+            pytest.skip("Unknown backend")
+
+    class UnsupportedConfig(BareConfig):
+        screens = [libqtile.config.Screen(top=libqtile.bar.Bar([UnsupportedWidget()], 20))]
+
+    manager_nospawn.start(UnsupportedConfig)
+
+    assert len(manager_nospawn.c.bar["top"].info()["widgets"]) == 0


m-col:
Yeah, I guess this makes sense, though the `Systray` is arguably an important widget and likely one of the most used (based on no data). It could be good to show how to do things in a config with a specific backend, so something like this could be quite nice:
```py
if qtile.core.name == "x11":
    tray = widget.Systray()
else:  # Wayland
    tray = widget.StatusNotifier()
```
Though perhaps in the grand scheme of things few people will use more than one backend often enough to configure two. Thoughts? IDK, maybe it's best to just remove it.

elParaguayo:
I did think about some backend check but thought that may be over-complicating it.

Maybe we could leave the line in but comment it out with a note saying it only works on x11.

The default config is just a starting point and we expect users to edit them so this seems like a good compromise.

m-col:
Yeah, sounds good

m-col:
I'm happy to do that (or the original change) but will leave it for someone else to check and merge in case there are strong opinions about keeping it :)

elParaguayo:
OK - Let's see if there are any other views first.

ramnes:
I also think that the systray widget is quite an important one for X11 users. Does it break Qtile on Wayland? If so I'd rather have it not break but just log a warning. If not, I wouldn't bother removing it, or I would add a condition similar to what @m-col proposed.

elParaguayo:
Let me check but I think it gives a ConfigError warning in the bar in wayland. That's why I was removing it as it doesn't look great to have an error in the default!

I'll have a rethink and update later.

elParaguayo:
So, yes, the SysTray widget does crash under Wayland. The "Widget Crashed" text box as a result is not a good look for a "default" config. There are a few options:
1) ~~Delete the Systray~~ We discounted this above
2) Update the Systray widget so it doesn't crash on `_configure`. Could give a log message but this wouldn't be visible to user who may not understand why they don't see a systray.
3) Replace the widget on wayland with a text box saying that the Systray is not available.
4) Replace the widget with the SystemNotifier widget.

Option 4 may be the best here but I'll need a bit more time to finish the widget (should have something to share soon). However, once we've got the SN widget, do we keep the SysTray too?

None of this is a breaking issue so I'm happy to park this for now.

m-col:
Or fully replace the Systray widget with the SN widget in the default config
once that is complete? In theory it should supercede it and is in use by many
applications. Can't speak about in practice though :D

elParaguayo:
Yes. I haven't tried running both side by side yet...

tych0:
I wonder if we can set something like `libqtile.active_backend` so people can do these kinds of tests in their config?

Or maybe more generally we should export a `libqtile.system` abstraction over both backends which would describe things like number of monitors, geometry, active backend, and such so people can make decisions accordingly.

tych0:
(FWIW, I agree with @ramnes we should keep the Systray widget in X11 mode; it's fairly basic functionality)

m-col:
> I wonder if we can set something like libqtile.active_backend so people can do these kinds of tests in their config?

You mean like the `qtile.core.name` test above? https://github.com/qtile/qtile/pull/2576#issuecomment-867170474

elParaguayo:
Not entirely sure why (yet) but trying to test `qtile.core.name` in the default_config stops qtile from loading.

m-col:
Ah, I bet it's because it is being imported before `libqtile.init`, directly
with an `import` statement rather than by `Config`.

elParaguayo:
OK - well that test is available when we `_configure` the widgets so we could test for x11 in the backend and only configure everything if it's on x11, otherwise it's just an empty widget.

ramnes:
For once we have a real reason to log something, so let's do it! :P

elParaguayo:
Ha. Nooo! I debated this and decided against it because I didn't think it was necessary for a default config i.e. there's no error, the widget just isn't available on wayland.

If we do log it then it feels like `INFO` to me.

m-col:
Looking at confreader.py, the import actually _is_ deferred until it is needed -
the `import` statement is within `Config.update`, so I don't think that was the
issue.

ramnes:
> I debated this and decided against it because I didn't think it was necessary for a default config i.e. there's no error, the widget just isn't available on wayland.

I think a warning should be issued when users try to use a widget that isn't compatible for the backend they use, to increase our chances of them understanding what's wrong with their configuration before opening an issue. :)

In combination with a backend condition in the configuration as proposed by @m-col, I think it would be ideal.
 
Related to #2416.

elParaguayo:
OK - looks like I need to go back and find out why the backend test isn't working in the default config then...

elParaguayo:
> Looking at confreader.py, the import actually _is_ deferred until it is needed - the `import` statement is within `Config.update`, so I don't think that was the issue.

That's not right. `default_config` is read as soon as a `confreader.Config` option is created (as `update` is called during `__init__`). 

This means we can't currently check the backend as, at the point `Config.update` is first called, there's no qtile instance. 

I'll have a think as to how best to deal with this.

elParaguayo:
Maybe overkill but...

We could add a `backends` attribute to widgets, similar to `orientations` e.g. `BACKEND_X11`, `BACKEND_WAYLAND`, `BACKEND_ALL`.

When the bar is being configured, if the widget is incompatible with the backend, the bar could just remove it.

m-col:
I think it would be overkill to set up a whole structure for it. The widgets can individually test for a backend if they need to and raise a `ConfigError` if they aren't happy with the running conditions, then the `Bar` can handle it as it does currently by placing a config error widget in its place. I think hiding the error isn't a good idea - that is, dropping the widget quietly or setting its width to 0 - otherwise people will be confused and wonder what's happening.

elParaguayo:
Fine. My only concern is that having a `default_config` generate a `ConfigError` looks odd. We can hide the crashed notification on the bar but include a message in the logs to explain what's happened.

If you disagree, I'm on IRC and we can discuss more there rather than having too much more on this page.

m-col:
On second thought it might not be so bad. Only widgets that are restricted to one backend need to say so, and then perhaps a logged message is sufficient if the message is clear:
```py
class Bar:
...
    def _configure_widget(self, widget):
        configured = True
        if hasattr(widget, "_supported_backends"):
            if self.qtile.core.name not in widget._supported_backends:
                logger.warning(f"Widget not supported by backend: {widget}"
                self.crashed_widgets.append(widget)
                return False
        ...
```
Then it probably would be fine to do it quietly. Maybe that's what you meant originally but I was picturing adding info to all widgets but i guess that isn't necessary :stuck_out_tongue:

docs/sphinx_qtile.py
```diff
@@ -51,6 +51,8 @@
     .. compound::
 
         Supported bar orientations: {{ obj.orientations }}
+
+        Supported backends: {{ ", ".join(obj.supported_backends) }}
```
elParaguayo, line 55: This would show for every widget but we could add a condition so it's only shown if the widget has a different value for `supported_backends`.

m-col:
Nice, looks good. Then we keep Systray in the default config? Would we later replace it once the SNI is finished? I guess for that we might need to see how many programs actually use it. Perhaps both work well together?

docs/sphinx_qtile.py
elParaguayo: Should also remove the text in the widget docstring if we keep this.

The advantage of having sphinx do it is that widgets only need to add the `supported_backends` value and the rest is automatic.

elParaguayo:
Having both Systray and SNI will probably give some duplication (e.g. vlc). I am also a bit concerned that SNI seems to be more KDE focussed while Gnome developed their own implementation called Ayatana which is not compatible with the widget...

docs/sphinx_qtile.py
m-col: Yeah I think it would be nice to avoid the noise of having it displayed for all widgets. Is it straight forward to do with sphinx?

elParaguayo: It's already done.

m-col: Ah, I re-read like 3 times trying to figure out, completely missed line 146. Trying out the new "high contrast" github dark mode and ironically it has made the background of the green lines very low contrast, lol.

m-col:
LGTM like this!

m-col:
> Having both Systray and SNI will probably give some duplication (e.g. vlc). I am also a bit concerned that SNI seems to be more KDE focussed while Gnome developed their own implementation called Ayatana which is not compatible with the widget...

I agree having both is not ideal. Given we can rely on clients using the X systray, It is likely sufficient for X. As it is unavailable for use in wayland, the SNI is good to have for that, even if it does only show some clients. For that reasoning I re-propose the backend conditional I put in my first comment above. What do you think? It would be nice give the SNI widget the attention it deserves!

elParaguayo:
The backend conditional is useful because it will catch scenarios where people try to use `SysTray` in their own Wayland configs, even if they're not using `default_config`.

However, it doesn't solve the issue of just getting SNI to be default for Wayland as we're not saying it's incompatible on x11.

Bit more work to do.

elParaguayo:
I think the problem is here:
https://github.com/qtile/qtile/blob/f73371457720be20e2c23184d221837d05afd641/libqtile/scripts/start.py#L49-L77

`kore` is defined before reading the config, but `Qtile` isn't created until after the config has been read so, in that initial read of the config, we can't parse the backend.

elParaguayo:
Ah, I'm with you. Yes, it's that `self.update` call that causes the problem because that's what imports `default_config`.

Given that `qtile start` doesn't pass `settings`, we can just add a test to see if `settings` has been set and, if it has, call `self.update`.

elParaguayo:
How about a new widget called `Tray` that replaces itself in the Bar when it's being configured and knows what backend it's on.

Or something like that...

elParaguayo:
That won't fix the problem with directly importing `default_config` in tests but maybe there's no easy fix for that.

elParaguayo:
We could wrap the backend test in a try...except... block. That should mean tests run ok.

m-col:
Are all uses of default_config in tests to populate `Config` classes used for tests? If so, we could probably just make a `test/config.py` with a base config used for these tests, and remove the use of default_config.py from tests altogether.

elParaguayo:
Pretty much, yes.

However, some tests rely on certain parts of `default_config` e.g. just copying `floating_layout`. I'd rather not split this into two configs as some bugs could slip through the cracks.

elParaguayo:
We should try to get this one merged so we can stop people reporting SysTray "crashing" on Wayland.

libqtile/widget/base.py
```diff
@@ -123,6 +123,11 @@ def open_calendar():
     """
 
     orientations = ORIENTATION_BOTH
+
+    # Default (empty list) is for all backends to be supported. Widgets can override this
+    # to explicitly confirm which backends are supported
+    supported_backends: List[str] = []
```
m-col, line 129: Could this be a `set`?

libqtile/confreader.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, file_path=None, **settings):
         config attribute precedence is 1.) **settings 2.) self 3.) default_config
         """
         self.file_path = file_path
+        # if settings:
```
m-col, line 83: Was this left in by accident?

libqtile/bar.py
```diff
@@ -330,13 +339,19 @@ def _remove_crashed_widgets(self):
...
-            if self.horizontal:
-                crash.offsety = self.border_width[0]
-            else:
-                crash.offsetx = self.border_width[3]
-            self.widgets.insert(index, crash)
+
+            # Widgets that aren't available on the current backend should not
+            # be shown as "crashed" as the behaviour is expected. Only notify
+            # for genuine crashes.
+            if not i.supported_backends or (self.qtile.core.name in i.supported_backends):
```
m-col, line 346: So widgets that are incompatible with the backend are just invisibly removed?

That's pretty nice - no need for backend checks or whatever in your config, just keep incompatible widgets defined and Qtile will deal with it.

CHANGELOG
```diff
@@ -7,6 +7,8 @@ Qtile 0.19.0, released 2021-12-22:
         - Add ability to draw borders to the Bar. Can customise size and colour per edge.
         - Add `StatusNotifier` widget implementing the `StatusNotifierItem` specification.
           NB Widget does not provide context menus.
+        - `StatusNotifier` widget added to default config for Wayland backends only
+          (SysTray remains default for X11 backends)
         - Add `total` bandwidth format value to the Net widget.
```
m-col, line 12: Could you add another entry saying that defined widgets that are not compatible with the running backend will invisibly be removed? And perhaps a line in the docs below the widgets saying the same thing, so that anybody reading the "Only available on the following backends:" note on a widget see it and search the page to see what this means/what will happen, they'll see that they can just leave it in with no issue.

m-col: Also, the diff doesn't actually make the change that this changelog entry describes - the default config is unchanged. Are we going with the backend check to set the systray to `SysTray` or `StatusNotifier`?

m-col (changes requested):
Agreed - and looking good. Couple comments added inline.

libqtile/bar.py
elParaguayo: Exactly.

libqtile/confreader.py
elParaguayo: Um... maybe...

libqtile/widget/base.py
elParaguayo: Yes.

CHANGELOG
elParaguayo: Oops. Forgot to change the changelog.

elParaguayo: > Also, the diff doesn't actually make the change that this changelog entry describes - the default config is unchanged. Are we going with the backend check to set the systray to `SysTray` or `StatusNotifier`?

No. We're not doing this because the config is loaded before the backend is started so we can't check the backend. However, I will include `StatusNotifier` as a commented line instead,

elParaguayo:
Mustn't forget about this one!

I'll add a test to check the handling of incompatible backends.

CHANGELOG
```diff
@@ -43,6 +50,9 @@ Qtile 0.20.0, released 2022-01-24:
           managers to request the name.
         - Fix bug where `Battery` widget did not retrieve `background` from `widget_defaults`.
         - Fix bug where widgets in a `WidgetBox` are rendered on top of bar borders.
+        - Widgets that are incompatible with a backend (e.g. Systray on Wayland) will no longer show 
+          as a ConfigError in the bar. Instead the widget is silently removed from the bar and a message
+          included in the logs.
```
m-col, line 56: I think this can be dropped?

libqtile/resources/default_config.py
```diff
@@ -138,6 +138,8 @@
                 ),
                 widget.TextBox("default config", name="default"),
                 widget.TextBox("Press &lt;M-r&gt; to spawn", foreground="#d75f5f"),
+                # NB Systray is incompatible with Wayland, consider using StatusNotifier instead
+                # widget.StatusNotifier(),
                 widget.Systray(),
```
m-col, line 143: Should we add an if block above this block with something like this?:
```py
from libqtile import qtile  # at the top

if qtile.core.name == "x11":
    tray = widget.Systray()
elif qtile.core.name == "wayland":
    tray = widget.StatusNotifier()
```

What do you think? 

and then just `tray` here, obviously.

m-col (approved):
Looking good! Just one discussion point

CHANGELOG
elParaguayo: No. That's exactly what this PR does now (the title here is old)

libqtile/resources/default_config.py
m-col: Oh yes.................... ignore me :sweat_smile:

CHANGELOG
m-col: Yep, but the changelog entry is duplicated :smile: 
image

elParaguayo: Whoops!

PR #2577, opened 2021-06-24
x11: ignore mouse button modifiers when handling bindings

### LINKED ISSUE #2525: [Enhancement] Execute Keybindings while holding mouse buttons down
# Issue description
Imagine this situation. I have groups 1 containing a browser window and 2 containing a file manager. I want to drag a file from the file manager to the browser across windows.  Or even on the same group switching between windows with alt+tab. Considering all those groups have the Max layout.

Currently that is not possible because it seems that holding Button1. Button2 or Button3 down blocks any shortcut I have set on my config. It would be ideal to be able to execute those shortcuts or at least enable ignoring the mouse buttons for them, while holding the mouse button. Then I guess that would be enough to drag files across groups or different windows in max layout ?

# Qtile version

``` 
$ qtile --version
0.17.1.dev315+g67f97604

```
OS: Arch Linux x86_64

m-col:
Currently we ignore numlock as a modifier, and remove its mask from key
and button events before processing the event and handling bindings. We
need to do the same thing for mouse buttons, which also have modmasks
associated with them. Not doing so causes an issue where keybindings
cannot be executed when a mouse button is held down.

Closes #2525

This commit also removes a redundant `|= self._numlock_mask` from these
event handlers.

--

@matheusfillipe please could you run with this patch to confirm it's all working as expected and that all key and mouse bindings continue to work fine? In particular, I'd like confirm that the keypad bindings still work.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -161,7 +161,7 @@ class Core(base.Core):
 
         numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["Num_Lock"])[0]
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
-        self._valid_mask = ~(self._numlock_mask | xcbq.ModMasks["lock"])
+        self._valid_mask = ~(self._numlock_mask | xcbq.ModMasks["lock"] | xcbq.AllButtonsMask)
 
     @property
     def name(self):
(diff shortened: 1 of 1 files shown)


matheusfillipe:
The keypad bindings still work both with numlock on and off. My mouse bindings also still work. Keyboard bindings are also still ok. And most importantly! I can now drag files between groups. The mouse buttons are ignored for the keyboard bindings.

Didn't find any issue yet, I think that is good to go.

PR #2578, opened 2021-06-24
Added group function to switching windows

ArtemSmaznov:
I have been looking for a way to switch focus to non-tiling windows (Floating, Maximized, Minimized, Full-screen) as lazy.layout.up/down/left/right/next/prev() only target windows that have not had their layouts changed and I count not find a function for that anywhere in the documentation.
Did manage to find it in an 11 year old issue so decided to propose to have it added to the documentation
Apologies in advance if this is not the best page to add it to or if the description is lacking.


ramnes:
Thanks!

PR #2579, opened 2021-06-25
Add text parsing/modifying option for window name/tab, task list widgets

yobleck:
Adds `text_filter` as a config option for the windowname, windowtabs and tasklist widgets.
The option takes a list of strings that will be removed from the displayed version of window names.
for example the window name:
`"Comparing qtile:master...yobleck:text_filter · qtile/qtile - Chromium"`
has the text `" - Chromium"` which takes up space and isn't necessary for many users (ms windows doesn't even display text anymore, only the program icon).

This allows users to remove text they don't want or need to see without relying on the arbitrary cut off ellipses caused by running out of screen space.

Other examples of text that users might want to remove includes "\u2014 " + Konsole, Dolphin, Kate or any other program that tasks its name onto the end of the windows wm_name. This option can also be used to hide sensitive information, enhancing privacy in public settings.

Edit: changed to more general user defined function. Above is one example of how it can be used

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,12 @@ jobs:
     build:
         runs-on: ubuntu-20.04
         name: "python ${{ matrix.python-version }}"
+        env:
+          WAYLAND: 1.19.0
+          WAYLAND_PROTOCOLS: 1.21
+          WLROOTS: 0.14.0
+          SEATD: 0.5.0
+          LIBDRM: 2.4.105
         strategy:
             matrix:
                 # if you change one of these, be sure to update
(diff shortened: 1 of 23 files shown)


m-col:
Would it make sense for this functionality to be provided by a more general option, e.g. `parse`, which would accept a function to which the string would be passed? Then the specific case you mention would be:
```py
def parse(text):
    for string in [" - Chromium", " - SomethingElse", " dolphin"]:
        text = text.replace(string, "")
    return text
```
Then pass `parse=parse` to the widget. This would allow people to use regexes if they want, or replace the strings with something other than `""`, for example `text.replace("password1234", "************")`. Or do something for _aesthetics_ like `text = text.lower()`. Thoughts?

yobleck:
> Would it make sense for this functionality to be provided by a more general option

That might be better, but if we go that general would it make more sense to move the option to the base _TextBox() widget
and put it in the update function like:

https://github.com/qtile/qtile/blob/50a4adf3066d38a725c1e9c66b861eec73e56bac/libqtile/widget/base.py#L456
```
def update(self, text):
    if self.text == text:
        return
    if text is None:
        text = ""
   
    if parse_text: # maybe check that parse is a function that can accept inputs here?
        text = parse_text(text)

    old_width = self.layout.width
    self.text = text

    # If our width hasn't changed, we just draw ourselves. Otherwise,
    # we draw the whole bar.
    if self.layout.width == old_width:
        self.draw()
    else:
        self.bar.draw()
```
so that all text based functions can benefit from the option, or would that clash with the already existing `fmt` option? (what does the fmt option even do? it's not clear to me from the code)
I'm hesitant to add such a general option to base without knowing how it will affect all widgets.

Also how would error handling work when there is a lot of room for potential user error. Would there just be a general try/except block that logs an error and then allows the widget to function as though the parse option never happened?

yobleck (commented):
okay so mypy doesn't like empty lists. good to know.

yobleck:
> That might be better, but if we go that general would it make more sense to move the option to the base _TextBox() widget?

nvm that doesn't make sense cause that method just gets overwritten.

Should the documentation include an explanation or example of how to use the option?

m-col:
> Should the documentation include an explanation or example of how to use the option?

That might go good to have, yeah.

> also should there be a try except block to catch errors from bad user functions and allow the widget to run as though the function never happened?

This too - it'll definitely happen! Might be good to wrap it in a method and log an exceptions raised.

With those in, please could you squash down the commits?

yobleck:
> Might be good to wrap it in a method and log an exceptions raised.

Just to clarify, by method do you mean moving the code over to a method, outside of self.get_taskname and self.update, called self.run_parser() or something and then calling that method inside of update/get_taskname?

m-col:
Sorry actually, for some reason I read the three file changes to be in the same class, not three separate classes. Don't bother with sticking it in a method - just wrap what you have already in the try-except block, should be good.

yobleck:
That's only my second time squashing/rebasing so hopefully I did that right. Let me know if I screwed it up.

m-col:
Looks like something did go wrong with the rebase. It looks as though all of the new commits that you were merging before your commit were incorporated into your own, which is why the `Files Changed` tab at the top shows all of those changes. It might help to check out `git pull --rebase` if you haven't already. For now though I have fixed it up and merged it for you, so thanks for the changes!

yobleck:
Thanks for the help :)

Issue #2566, opened 2021-06-20
qtile-cmd : unkown command


Rockwell1799:
# Issue description
I came back to qtile after few monts of Awesomewm and before I was able to run this commad in my terminal:
qtile-cmd
now, It says that this command is unkown!
I tried also:
qtile run-cmd and it gives this
```
Traceback (most recent call last):
  File "/usr/bin/qtile", line 33, in <module>
    sys.exit(load_entry_point('qtile==0.17.0', 'console_scripts', 'qtile')())
  File "/usr/lib/python3.9/site-packages/libqtile/scripts/main.py", line 59, in main
    options.func(options)
  File "/usr/lib/python3.9/site-packages/libqtile/scripts/run_cmd.py", line 41, in run_cmd
    proc = subprocess.Popen(opts.cmd)
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1691, in _execute_child
    executable = args[0]
IndexError: list index out of range
```
os:Archlinux

How can I fix this issue, please?

# Qtile version
Qtile 0.17.01 (official)

# Stack traces

# Configuration

hackedWifi:
Here is some information on these commands.
http://docs.qtile.org/en/latest/manual/commands/shell/qtile-run.html

elParaguayo:
I can confirm I get the same error where it's run with no parameters. This shouldn't happen and, instead, it should provide more info on how to use the command.

Will mark as a bug for now.

nazar-pc:
This command no longer exist, you need to use `qtile cmd-obj`, see changelog for 0.17.0: https://github.com/qtile/qtile/blob/master/CHANGELOG

elParaguayo:
Indeed, the changes were as follows:
```
                qtile     -> qtile start
                qtile-cmd -> qtile cmd-obj
                qtile-run -> qtile run-cmd
                qtile-top -> qtile top
                qshell    -> qtile shell
```
However, running `qtile run-cmd` should provide some more useful output when run, rather than crashing.

elParaguayo:
Closed by #2583

PR #2583, opened 2021-06-27
Fix 'qtile run-cmd' bug with no command

elParaguayo:
Executing `qtile run-cmd` with no arguments results in an ugly exception with no explanation of the problem. This is because
the script tries to execute the `cmd` argument and errors when there's no command.

The problem is that the script uses `argparse.REMAINDER` to catch the command and appropriate arguments but `REMAINDER` allows no matches whereas we need at least one.

This PR tries to address that by splitting the `cmd` and `args` element to force the `cmd` to be required.

The only downside here is that, when run with no arguments, the script produces the following output:
```
usage: qtile run-cmd [-h] [-l {DEBUG,INFO,WARNING,ERROR,CRITICAL}] [-s SOCKET] [-i] [-f] [-b] [-g GROUP] cmd ...
qtile run-cmd: error: the following arguments are required: cmd, [args ...]
```
`args` is not shown in the first line but is shown as a "required" argument in the second. I've tried to address this by putting `args` in square brackets.

This should close bug in #2566 (although that wasn't primary issue reported).

### DIFF
--- a/libqtile/scripts/run_cmd.py
+++ b/libqtile/scripts/run_cmd.py
@@ -38,7 +38,11 @@ def run_cmd(opts) -> None:
     client = ipc.Client(socket)
     root = graph.CommandGraphRoot()
 
-    proc = subprocess.Popen(opts.cmd)
+    cmd = [opts.cmd]
+    if opts.args:
+        cmd.extend(opts.args)
+
+    proc = subprocess.Popen(cmd)
     match_args = {"net_wm_pid": proc.pid}
     rule_args = {"float": opts.float, "intrusive": opts.intrusive,
                  "group": opts.group, "break_on_match": not opts.dont_break}
@@ -86,6 +90,11 @@ def add_subcommand(subparsers, parents):
         help='Set the window group.')
     parser.add_argument(
         'cmd',
+        help='Command to execute.'),
+    parser.add_argument(
+        'args',
         nargs=argparse.REMAINDER,
-        help='Command to execute')
+        metavar='[args ...]',
+        help='Optional arguments to pass to command.'
+    )
     parser.set_defaults(func=run_cmd)


m-col:
Arguments _are_ required right? If so, why the square brackets?

elParaguayo:
I don't think so.

The script uses `argparse.REMAINDER` to capture the command and any additional arguments that should be passed to that command. So the command is required but the arguments are optional.

The previous code expected at least one value (ie the command to execute) but `argparse.REMAINDER` allows 0 values.

m-col:
Right, my bad, i was reading "cmd" as "run-cmd". OK, cool.

m-col (approved):

libqtile/scripts/run_cmd.py
```diff
@@ -86,6 +90,11 @@ def add_subcommand(subparsers, parents):
         help='Set the window group.')
     parser.add_argument(
         'cmd',
+        help='Command to execute.'),
```
ramnes, line 93: It doesn't work to just add `required=True` here?

elParaguayo: No. That was my first try too. If you do that you get `TypeError: 'required' is an invalid argument for positionals`.

I also tries using `nargs='+'` but that didn't work as it tries to consume any additional arguments for itself, rather than passing them to `cmd`. For example, running `qtile run-cmd xeyes -fg "#ff0000"` will give the following when parsing the command:
``` python
Namespace(log_level='WARNING', socket=None, intrusive=False, float=True, dont_break=False, group='#ff0000', cmd=['xeyes'], func=<function run_cmd at 0x7f478331aca0>)
```
Note that the `fg` argument which should set xeyes's foreground colour is now the `group` argument.

The solution I've posted should avoid those issues.

ramnes:
Thanks!

PR #2589, opened 2021-06-29
Small fixes for Popups and Internals I found while debugging

m-col:
I was updating my notification server which I hadn't used since I moved to Wayland and came across some bugs. Here are 3 commits:

 - minor Wayland `Internal` improvements
 - libqtile.notify.notifier should catch and log exceptions raised by callbacks without stopping further callbacks from being called
 - minor `Popup` improvements

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -125,7 +125,10 @@ if has_dbus:
             self.notifications.append(notif)
             notif.id = len(self.notifications)
             for callback in self.callbacks:
-                callback(notif)
+                try:
+                    callback(notif)
+                except Exception:
+                    logger.exception("Exception in notifier callback")
             return len(self.notifications)
 
         def show(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


libqtile/notify.py
```diff
@@ -125,7 +125,10 @@ def add(self, notif):
             self.notifications.append(notif)
             notif.id = len(self.notifications)
             for callback in self.callbacks:
-                callback(notif)
+                try:
+                    callback(notif)
+                except Exception:
```
tych0, line 130: Odd to me that this doesn't trigger some sort of linter error (we have noqa E722 for this elsewhere it looks like), but overall the series looks good.

m-col: I thought the same when I wrote it and read up that it's only bare `except:`that gives the error, and that `except Exception:` suppresses it because it's explicit that you want all to be caught.

elParaguayo:
Thanks for this. I'll rebase my StatusNotifier work onto this and will let you know what other wayland issues I run into!

PR #2592, opened 2021-06-30
migrate: always be interactive

### LINKED ISSUE #2353: migrations need version annotations

tych0:
This way people can reject changes to their configs that they don't want
(or are potentially incorrect, viz. the discussion in #2567).

Closes #2353 (we're relying on people to reject incorrect migrations for
now)

### DIFF
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -162,7 +162,7 @@ def do_migrate(args):
 
     for m in MIGRATIONS:
         q = bowler.Query(config_dir)
-        m(q).execute(interactive=args.interactive, write=True)
+        m(q).execute(interactive=not args.yes, write=True)
 
     changed = False
     for py, backup in file_and_backup(config_dir):
@@ -193,8 +193,8 @@ def add_subcommand(subparsers, parents):
         help="Use the specified configuration file (migrates every .py file in this directory)",
     )
     parser.add_argument(
-        "--interactive",
+        "--yes",
         action="store_true",
-        help="Interactively apply diff (similar to git add -p)",
+        help="Automatically apply diffs with no confirmation",
     )
     parser.set_defaults(func=do_migrate)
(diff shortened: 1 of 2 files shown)


m-col (approved):
LGTM

libqtile/scripts/migrate.py
```diff
@@ -193,8 +193,8 @@ def add_subcommand(subparsers, parents):
         help="Use the specified configuration file (migrates every .py file in this directory)",
     )
     parser.add_argument(
-        "--interactive",
-        action="store_true",
-        help="Interactively apply diff (similar to git add -p)",
+        "--non-interactive",
```
ramnes, line 196: What do you think of --yes or --quiet?

libqtile/scripts/migrate.py
```diff
@@ -162,7 +162,7 @@ def do_migrate(args):
 
     for m in MIGRATIONS:
         q = bowler.Query(config_dir)
-        m(q).execute(interactive=args.interactive, write=True)
+        m(q).execute(interactive=not args.non_interactive, write=True)
```
ramnes, line 165: Isn't the double negation wrong if store_false?

tych0: Yeah, derp :). It should be store_true still.

libqtile/scripts/migrate.py
tych0: `--yes` is ok, I can switch it to that. I think `--quiet` is kind of ambiguous.

ramnes: Yes please ❤️

PR #2594, opened 2021-07-01
Update docs for conditional Key callbacks

### LINKED ISSUE #2571: Documentation for layout-specific keybindings
I noticed at https://qtile-dev.narkive.com/KB0yShYE/keyboard-action-based-on-the-current-layout-possible that there is a relatively-easy way to set up keybindings on a per-layout basis.  Is there any official documentation about this that can be linked to from the layouts page?  I feel like this would be a pretty common sticking point for users of multiple layouts like MonadTall/Wide + Stacks (like it is for me).  

Also, is that link the best way to do this?

elParaguayo:
Closes #2571

### DIFF
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -45,6 +45,26 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
        'C': 'control',
     }
 
+Callbacks can also be configured to work only under certain conditions by using
+the ``when()`` method. Currently, two conditions are supported:
+
+::  
+
+    from libqtile.config import Key
+
+    keys = [
+        # Only trigger callback for a specific layout
+        Key(
+            [mod, 'shift'],
+            "j",
+            lazy.layout.grow().when(layout='verticaltile'),
+            lazy.layout.grow_down().when(layout='columns')
+        ),
+
+        # Limit action to when the current window is not floating (default True)
+        Key([mod], "f", lazy.window.toggle_fullscreen().when(when_floating=False))
+    ]
+
 KeyChords
 =========
 


docs/manual/config/keys.rst
```diff
@@ -45,6 +45,24 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
...
+
+    def myfunc():
+        pass
+
+    keys = [
+        # Only trigger callback for a specific layout
+        Key([mod], "a", lazy.function(myfunc).when(layout='max'))
+
+        # When the current window is floating (default True)
+        Key([mod], "b", lazy.function(myfunc).when(when_floating=False))
```
m-col, line 63: Could we instead use other `lazy` paths, and maybe two different ones? From seeing other people's configs I get the impression `lazy.function` is less widely used than e.g. `lazy.layout.next`, and having the `def` draws attention to that. With two different ones, it is more clear that `when` is applicable not just to that example.

docs/manual/config/keys.rst
```diff
@@ -45,6 +45,24 @@ The :class:`EzKey` modifier keys (i.e. ``MASC``) can be overwritten through the
...
+::  
+
+    from libqtile.config import Key
+
+    def myfunc():
+        pass
+
+    keys = [
+        # Only trigger callback for a specific layout
+        Key([mod], "a", lazy.function(myfunc).when(layout='max'))
```
m-col, line 60: Maybe for the layout example we could put two commands into the `Key`, with different layouts, e.g.:
```py
Key(
    [mod, 'shift'],
    "j",
    lazy.layout.grow().when(layout='verticaltile'),
    lazy.layout.grow_down().when(layout='columns'),
)
```
This seems like it would be the main use-case.

elParaguayo: Sounds good. Would still be good to include an example for the `when_floating` condition.

m-col: > Would still be good to include an example for the when_floating condition.

for sure

> Not sure what use case was envisaged for that.

Yeah me neither - guess it doesn't really matter then what the lazy thing is

PR #2596, opened 2021-07-01
WL: more minor Internal fixes

### LINKED ISSUE #2585: Wayland: incorrect mouse pointer
I've noticed that, when using the Wayland backend, I get the incorrect mouse pointer. It shows the "select text" pointer even when mousing over the bar.

Flagging this now but I'll see if I can I identify some steps to recreate this reliably.

# Qtile version

Seen on 50a4adf but was present before too.

m-col:
The previous commit caught the pointer motion coordinates but we missed
the pointer leave coordinates :D

This also resets the cursor image if the pointer moves from
Window->Internal like we do for Window->Desktop.

Fixes #2585

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -386,6 +386,12 @@ class Core(base.Core, wlrq.HasListeners):
                         self.cursor.y - self._hovered_internal.y
                     )
                 else:
+                    if self._hovered_internal:
+                        self._hovered_internal.process_pointer_leave(
+                            self.cursor.x - self._hovered_internal.x,
+                            self.cursor.y - self._hovered_internal.y
+                        )
+                    self.cursor_manager.set_cursor_image("left_ptr", self.cursor)
                     self.seat.pointer_clear_focus()
                     win.process_pointer_enter(self.cursor.x, self.cursor.y)
                     self._hovered_internal = win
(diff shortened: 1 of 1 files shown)


ramnes (approved):

libqtile/backend/wayland/core.py
```diff
@@ -386,6 +386,12 @@ def _process_cursor_motion(self, time):
                         self.cursor.y - self._hovered_internal.y
                     )
                 else:
+                    if self._hovered_internal:
+                        self._hovered_internal.process_pointer_leave(
+                            self.cursor.x - self._hovered_internal.x,
```
elParaguayo, line 391: I wondered about this. Do you end up with a negative value if the mouse leaves via the left or top edge of the window?

Do the co-ordinates really matter here? We need to know the mouse left the window but does it matter where the mouse is at that point?

m-col: You do get a negative value. I just tested on X11 and you see the same there with leave-notify events. This makes sense as it's the pointer coords at the event, and the event can't be raised before it leaves the window. As for whether they matter, that depends on what the window is doing, but they have the information there to work out what to do, if they care about where the pointer left the window.

elParaguayo:
Thanks. Confirmed that it fixes #2585.

PR #2599, opened 2021-07-02
Add actions and NotificationClosed method to notifier and notify widget

m-col:
This adds action handling to the notifier, which uses the
`ActionInvoked` signal on the notification service. This is bound to
button3 on the Notify widget. This will do things such as take Firefox
to the tab that sent a notification.

This also fixes the `NotificationClosed` signal which needs to return
the values it is passed to work. This signal is emitted when a
notification is closed and tells listeners of the reason, which can be
one of: expired (due to timeout), dismissed (the user cleared it) or
method (it was cleared by the dbus client via the CloseNotification
method). The CloseNotification method is given a basic implementation
whereby code registered with the notifier can pass an on_close callback
to `notifier.register` which will get executed when a dbus client calls
this method.

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -40,6 +40,13 @@ SERVICE_PATH = '/org/freedesktop/Notifications'
 
 notifier: Any = None
 
+
+class ClosedReason:
+    expired = 1
+    dismissed = 2
+    method = 3  # CloseNotification method
+
+
 if has_dbus:
     class NotificationService(ServiceInterface):
         def __init__(self, manager):
@@ -69,12 +76,16 @@ if has_dbus:
             return self.manager.add(notif)
 
         @method()
-        def CloseNotification(self, _id: 'u'):  # type:ignore  # noqa: N802, F821
-            pass
+        def CloseNotification(self, nid: 'u'):  # type:ignore  # noqa: N802, F821
+            self.manager.close(nid)
 
         @signal()
-        def NotificationClosed(self, _id_in, _reason_in):  # noqa: N802
-            pass
+        def NotificationClosed(self, nid: 'u', reason: 'u') -> 'uu':  # type:ignore  # noqa: N802, F821
+            return [nid, reason]
+
+        @signal()
+        def ActionInvoked(self, nid: 'u', action_key: 's') -> 'us':  # type:ignore  # noqa: N802, F821
+            return [nid, action_key]
 
         @method()
         def GetServerInformation(self) -> 'ssss':  # type:ignore  # noqa: N802, F821
@@ -96,6 +107,7 @@ if has_dbus:
         def __init__(self):
             self.notifications = []
             self.callbacks = []
+            self.close_callbacks = []
             self._service = None
 
         async def service(self):
@@ -110,7 +122,7 @@ if has_dbus:
                     self._service = None
             return self._service
 
-        async def register(self, callback, capabilities=None):
+        async def register(self, callback, capabilities=None, on_close=None):
             service = await self.service()
             if not service:
                 logger.warning(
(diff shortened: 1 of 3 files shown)


libqtile/widget/notify.py
```diff
@@ -47,6 +47,7 @@ class Notify(base._TextBox):
         ),
         ("audiofile", None, "Audiofile played during notifications"),
     ]
+    capabilities = {'body', 'actions'}
```
elParaguayo, line 50: Are we really supporting actions? Aren't actions supposed be displayed in the notification with the ability to select them? There can be multiple actions per notification.

If we don't display the action then if you `invoke` it, you're doing it blindly.

m-col: The first action provided in the actions list is the 'default' action, which is typically invoked by clicking anywhere on a notification and does not necessarily need displaying the message.

Perhaps we can put an option to disable actions, in case a user often uses a program that sends multiple actions and would prefer they can see them all?

elParaguayo: Yes. I don't have any issues with the approach for the default action but that's really all we're supporting.

We _could_ do it with some extra code but I'm not sure there's a neat way of doing it if we want to keep the text in the bar.

libqtile/notify.py
```diff
@@ -69,12 +76,16 @@ def Notify(self,  # noqa: N802, F722
...
         @method()
-        def CloseNotification(self, _id: 'u'):  # type:ignore  # noqa: N802, F821
-            pass
+        def CloseNotification(self, nid: 'u'):  # type:ignore  # noqa: N802, F821
+            self.manager.close(nid)
 
         @signal()
-        def NotificationClosed(self, _id_in, _reason_in):  # noqa: N802
-            pass
+        def NotificationClosed(self, nid: 'u', reason: 'u') -> 'uu':  # noqa: N802
```
elParaguayo, line 83: Is this `noqa`s sufficient? I think you need `F821` too (for `undefined name`) and you may even need `type: ignore` for mypy.

libqtile/notify.py
```diff
@@ -135,6 +149,13 @@ def show(self, *args, **kwargs):
             notif = Notification(*args, **kwargs)
             return (notif, self.add(notif))
 
+        def close(self, nid):
+            for callback in self.close_callbacks:
+                try:
+                    callback(notif)
```
elParaguayo, line 155: Don't think `notif` is defined. You're passing `nid` as an argument.

libqtile/widget/notify.py
```diff
@@ -98,21 +100,32 @@ def real_update(self, notif):
...
         self.bar.draw()
         return True
 
     def display(self):
         self.set_notif_text(notifier.notifications[self.current_id])
         self.bar.draw()
 
-    def clear(self):
+    def clear(self, reason=ClosedReason.dismissed):
+        notifier._service.NotificationClosed(self.current_id, reason)
```
elParaguayo, line 118: This doesn't actually remove the notification from the widget does it? It looks like it just hides the text and moves the display to the previous notification. Presumably we could make it re-appear by scrolling back and forwards again.

Should a dismissed notification be removed entirely?

Couldn't we also end up with an `IndexError` if we

m-col: I think it should be considered a 'dismiss' rather than a 'delete'. It sets the current ID to the final one, not to the previous.

elParaguayo:
LGTM. Is there anything else that you think needs to be added before we can merge?

m-col:
I don't think so - unless there is something I forgot that we discussed 
and I haven't updated? :D

elParaguayo:
Great. I'll take another look at the end of the day and then merge it.

PR #2601, opened 2021-07-02
Add StatusNotifier widget

elParaguayo:
This is probably in a state when it can be shared here so people can have a play if they want and feed back any comments.

There's a bit to do still, but hopefully nothing too massive:
- [ ] Checking which icon to use (normal vs "attention")
- [x] When to close menu if no item selected (e.g. clicking elsewhere)
- [x] Customising appearance of context menus
- [x] Tests
- [x] Add to default_config.py
- [x] Logging if no icons available
- [ ] Type hinting

Worth pointing out that, as well as the widget, this PR includes my popup toolkit (see #2251) which lets you do some fun stuff with popups. It's all in a single commit for now but I can split out the toolkit if we need.


libqtile/backend/wayland/window.py
```diff
@@ -641,7 +641,7 @@ def kill(self) -> None:
 
     def place(self, x, y, width, height, borderwidth, bordercolor,
               above=False, margin=None, respect_hints=False):
-        if above and self._mapped:
+        if above and self.mapped:
```
elParaguayo, line 644: Need to remove this. This was fixed in #2589.

elParaguayo:
Test is failing because `dbus-next` is not in the test environment. Without `dbus-next` the widget raises a `ConfigError` during `_configure` which the test is not expecting.

Could add it to `tox.ini` but it's already in #2434 which could probably get merged...

libqtile/popup/menu.py
```diff
@@ -0,0 +1,174 @@
...
+            row += 1 if sep else 2
+
+        return cls.generate(qtile, menuitems, **config)
+
+    @classmethod
+    def generate(cls, qtile, menuitems, **config):
+        rowcount = sum(x.row_span for x in menuitems)
+        menu_config = {
+            "width": 200,
+            "height": 12 * rowcount + 10,
```
elParaguayo, line 168: Still hardcoded. This should vary based on fontsize. Is there a formula for this?

libqtile/popup/toolkit.py
```diff
@@ -0,0 +1,626 @@
...
+        self.draw()
+
+    def hide(self):
+        """Hide the popup."""
+        self.popup.hide()
+
+    def kill(self):
+        self.popup.kill()
+
+    # The below methods are lifted from `bar`
```
m-col, line 135: Perhaps it's worth subclassing Bar? I suppose there is a lot of similarity between the two classes.

elParaguayo: Yes - may need to look at how feasible that is.

libqtile/widget/statusnotifier.py
```diff
@@ -0,0 +1,881 @@
...
+    from dbus_next.constants import PropertyAccess
+    from dbus_next.errors import DBusError
+    from dbus_next.service import (
+        ServiceInterface,
+        dbus_property,
+        method,
+        signal,
+    )
+    has_dbus = True
+except ImportError:
```
m-col, line 44: If the widget can't be used without `dbus_next` then it's fine to raise the `ImportError`. `__init__.py` will handle catching it and replacing the widget with an `ImportErrorWidget`.

libqtile/popup/toolkit.py
elParaguayo: Subclassing `Bar` would be overkill. An alternative approach would be to have some sort of mixin that contains these methods and then have both `Bar` and the toolkit inherit from this.

elParaguayo:
Closing in favour of #2750

PR #2604, opened 2021-07-04
Refactor tests

m-col:
This is a series of commits working towards getting tests running for both the x11 and wayland backends. The final state of the tests after these commits is largely unchanged -- by default only x11 tests are run, and none have been added or removed. The backends can be manually specified on the command line (first one is redundant):
```sh
pytest --backend=x11 ...
pytest --backend=wayland ...
pytest --backend=x11 --backend=wayland ...
```

I have to work my way through the tests and make fixes to the wayland backend (or the tests) before we enable the wayland tests in the CI.

One point of note in the tests is that the `test/scripts/window.py` test window and x11-apps clients that are used for tests have all been replaced by a new `test/scripts/window.py`. The new version uses PyGTK to create a simple window that works identically for X and Wayland, and we can more easily hook callbacks to key/button press, which I expect will be useful for testing.

There is one outstanding issue with the tests, which is a failure of `test/backend/x11/test_window::test_struct_handling` (previously in `test_manager.py`), but I'm having trouble identifying why this is newly being triggered, and it is consistently being triggered on github but always passing on my local machine. Anyone got any ideas how I could debug this, or perhaps see any potential causes?


elParaguayo:
How frustrating. It also passes on my local machine but I am confused as to why as well.

If I output the value of `xmanager.c.window.info()['width']` right before the failing assertion, I don't get a width of 798. Mine says 800.

m-col:
I found the issue - one of the tests in `test_manager.py` was doing `config = ManagerConfig` and then modifying its attributes, which of course mutates the original `ManagerConfig`. It was not an issue previously because the tests executed after this test happened to not mind, but the rearranging of tests from that file made a test that _did_ mind get executed after it, and so it failed. The fix is subclassing a new class for the test. The reason for it not being reproducible locally is because in the CI everything is executed, and is executed via tox and pytes, whereas locally I was doing `python -m pytest ...` and I guess the test order was slightly different... bit weird, but now the test is reliably passing.

So... ready for review!

test/backend/x11/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+            yield x
+
+
+@pytest.fixture(scope="function")
+def xmanager(request, xephyr):
+    """
+    This replicates the `manager` fixture except that the x11 backend is hard-coded. We
+    cannot simply parametrize the `backend_name` fixture module-wide because it gets
+    parametrized by `pytest_generate_tests` in test/conftest.py and only one of these
+    parametrize calls can be used.
```
ramnes, line 165: Should we refactorize and add a `tests.helpers:create_manager(backend)` function (not fixture) that is called by the different fixtures that create a manager? IIRC we already have that code replicated a few times.

test/backend/x11/test_window.py
```diff
@@ -0,0 +1,744 @@
...
+    # Click on window one
+    fake_click(conn, xtest, 10, 10)
+    xmanager.c.sync()
+    wins = get_all_windows(conn)
+    if bring_front_click == "floating_only":
+        assert wins.index(wids[0]) < wins.index(wids[2]) < wins.index(wids[1])
+    elif bring_front_click:
+        assert wins.index(wids[2]) < wins.index(wids[1]) < wins.index(wids[0])
+    else:
+        assert wins.index(wids[0]) < wins.index(wids[1]) < wins.index(wids[2])
```
ramnes, line 508: This function (and a lot of others too) don't feel X11-specific to me, even if the implementation is quite tied to X11 currently. :/

test/test_manager.py
```diff
@@ -1059,42 +1059,6 @@ def test_dheight():
...
-        bottom=libqtile.bar.Bar(
-            [
-                libqtile.widget.GroupBox(),
-            ],
-            20
-        ),
-    )]
-    auto_fullscreen = True
-
-
```
ramnes, line 1205: nice, lol

test/backend/wayland/conftest.py
```diff
@@ -16,8 +17,24 @@
...
-    os.environ.update(wlr_env)
-    os.environ["WLR_HEADLESS_OUTPUTS"] = str(outputs)
-    yield os.environ
-    os.environ = old_env
+    env = wlr_env.copy()
+    env["WLR_HEADLESS_OUTPUTS"] = str(outputs)
+    yield env
+
+
+class WBackend(Backend):
```
ramnes, line 25: WaylandBackend please :P

test/helpers.py
```diff
@@ -0,0 +1,330 @@
...
+        proc = None
+
+        def spawn():
+            nonlocal proc
+            # Ensure the client only uses the test display
+            env = os.environ.copy()
+            if "DISPLAY" in os.environ:
+                del env["DISPLAY"]
+            if "WAYLAND_DISPLAY" in os.environ:
+                del env["WAYLAND_DISPLAY"]
```
ramnes, line 259:
```suggestion
            env.pop("DISPLAY", None)
            env.pop("WAYLAND_DISPLAY", None)
```

test/test_fakescreen.py
```diff
@@ -171,7 +171,7 @@ def test_basic(manager):
     assert manager.c.screen.info() == {
         'y': 480, 'x': 0, 'index': 2, 'width': 500, 'height': 400}
     manager.c.to_screen(3)
-    manager.test_xclock()
+    manager.test_window('xclock')
```
ramnes, line 174: Let's not call it xclock if we don't use xclock anymore. Also it is mentioned in the docs that xclocks is required for the tests, so we should remove that too.

ramnes (approved):
LGTM overall, thanks!

test/backend/x11/test_window.py
m-col: I see what you mean - the behaviour itself is not specific to the X11 _backend_, but the implementation of Z stacking (in this case) and therefore our testing tools are specific to the environment. In the Wayland equivalent, much of the code will be similar, but our determination of whether the outcomes are correct will be specific to that environment.

test/helpers.py
m-col: Nice, much better

test/backend/x11/test_window.py
ramnes: What testing tools? `get_all_windows` and `fake_click`? `get_all_windows` could be a function inside the manager and `fake_click` a backend-agnostic function in tests helpers.

m-col: ooh nice, i like it. Will give that a shot.

test/test_fakescreen.py
m-col: Cool - fixed these.

test/backend/x11/test_window.py
m-col: I set these two functions and managed to move `test_popup` in addition to this `bring_front_click` test back into the main set of tests for use with both backends. Works great! I think the `cursor_warp` tests might also be able to be re-used but I'm less certain there so I've left it until there is actually more setup on the Wayland test side.

test/backend/x11/conftest.py
m-col: Good idea - I've significantly reduced down the code in the `manager` and co fixtures, moving the temporary file code inside `TestManager`'s own context manager, and deleting some unnecessary lines assigning attributes to the test `config` from the default config, which is redundant as this happens anyway within the `Qtile` manager's thread.

m-col:
A current limitation of this PR which I think we should fix before merging it that the dependencies of both backends are required even when just testing for one. The `backend` fixture in `test/conftest.py` currently takes both the `xephyr` and `wayland_session` fixtures but only uses the one that it needs for that test. Really it should only take the fixture when it needs it, which would mean that you don't need the deps for one backend if you're only ever testing the other. The pytest lazy fixture would be fine for this if those two fixtures weren't themselves parametrized (e.g. no_xinerama). I've hit a bit of a brick wall with this, so if anyone has any smart ideas on how this could work then please let me know!

m-col:
I've made some nice new changes particularly with the way the number of outputs is configured for tests. There is an `outputs` fixture that can be parametrised with an `int` to specify the number of outputs. This replaces the `no_xinerama` stuff and makes things work quite well with wayland.

I don't think I will add anything else to this PR, because now what is left to do isn't modifying the tests to work with wayland, it's modifying the wayland backend to pass the tests! The one thing I _do_ need/want to change in this PR is remove the requirement to have all of the dependencies for a given backend when you're only testing the other. Currently the `backend` fixture unconditonally uses both a x11 and wayland session fixture, even if the tests that will be ran only use one. We'll need some way of conditionally creating fixtures or something I think.

ramnes:
Would in-function imports help?

```diff
diff --git a/test/backend/wayland/conftest.py b/test/backend/wayland/conftest.py
index ba1922a2..bab455ec 100644
--- a/test/backend/wayland/conftest.py
+++ b/test/backend/wayland/conftest.py
@@ -2,7 +2,6 @@ import contextlib
 import os
 import textwrap
 
-from libqtile.backend.wayland.core import Core
 from test.helpers import Backend
 
 wlr_env = {
@@ -23,6 +22,7 @@ def wayland_environment(outputs):
 
 class WaylandBackend(Backend):
     def __init__(self, env, args=()):
+        from libqtile.backend.wayland.core import Core
         self.env = env
         self.args = args
         self.core = Core
diff --git a/test/conftest.py b/test/conftest.py
index 82b94b37..fe816bfb 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -24,8 +24,6 @@
 import pytest
 
 from libqtile.backend import base
-from test.backend.wayland.conftest import WaylandBackend, wayland_environment
-from test.backend.x11.conftest import XBackend, x11_environment
 from test.helpers import BareConfig, TestManager
 
 
@@ -58,6 +56,8 @@ multimonitor = pytest.mark.parametrize("outputs", [1, 2], indirect=True)
 
 @pytest.fixture(scope="session")
 def xephyr(request, outputs):  # noqa: F841
+    from test.backend.x11.conftest import x11_environment
+
     kwargs = getattr(request, "param", {})
 
     with x11_environment(outputs, **kwargs) as x:
@@ -66,6 +66,8 @@ def xephyr(request, outputs):  # noqa: F841
 
 @pytest.fixture(scope="session")
 def wayland_session(outputs):  # noqa: F841
+    from test.backend.wayland.conftest import wayland_environment
+
     with wayland_environment(outputs) as w:
         yield w
 
@@ -73,11 +75,11 @@ def wayland_session(outputs):  # noqa: F841
 @pytest.fixture(scope="function")
 def backend(request, backend_name, xephyr, wayland_session):
     if backend_name == "x11":
-        b = XBackend({"DISPLAY": xephyr.display}, args=[xephyr.display])
+        from test.backend.x11.conftest import XBackend
+        yield XBackend({"DISPLAY": xephyr.display}, args=[xephyr.display])
     elif backend_name == "wayland":
-        b = WaylandBackend(wayland_session)
-
-    yield b
+        from test.backend.wayland.conftest import WaylandBackend
+        yield WaylandBackend(wayland_session)
 
 
 @pytest.fixture(scope="function")
```

This makes the tests run without Wayland dependencies, although they don't pass for some reason.

m-col:
Unfortunately due to this line:
```py
 def backend(request, backend_name, xephyr, wayland_session):
```
The `xephyr` and `wayland_session` fixtures are going to be executed anyway. We need to conditionally use these fixtures, while still maintain the ability to parametrize their own fixtures (i.e. `outputs`).

ramnes:
Then why do I have timeout errors and no import errors with such a diff?

m-col:
I don't get import errors (when removing the wayland stuff from my pythonpath) unless I do `--backend=wayland`, which is good but only works because `wayland_session` only does some environmental variable setup and doesn't require any imports. The same logic doesn't work for the x11 stuff because the `x11_environment` starts up Xephyr and is called whether or not it is used. Perhaps this can be a temporary half-solution but ideally someone testing and using the Wayland stuff only shouldn't need Xephyr and co installed.

Not sure what your timeout error is - perhaps you don't have PyGObject installed so `test/scripts/window.py` isn't running successfully?

ramnes:
> Not sure what your timeout error is - perhaps you don't have PyGObject installed so `test/scripts/window.py` isn't running successfully?

Bingo. :)

> The same logic doesn't work for the x11 stuff because the x11_environment starts up Xephyr and is called whether or not it is used.

Alright. Let's add safeguards then!

```diff
diff --git a/test/conftest.py b/test/conftest.py
index 82b94b37..3b83b2a6 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -24,8 +24,6 @@
 import pytest
 
 from libqtile.backend import base
-from test.backend.wayland.conftest import WaylandBackend, wayland_environment
-from test.backend.x11.conftest import XBackend, x11_environment
 from test.helpers import BareConfig, TestManager
 
 
@@ -36,14 +34,26 @@ def pytest_addoption(parser):
     parser.addoption(
         "--backend",
         action="append",
-        default=[],
-        help="Specify a backend to test. Can be passed more than once.",
+        choices=("x11", "wayland"),
+        help="Test a specific backend. Can be passed more than once.",
     )
 
 
+def pytest_cmdline_main(config):
+    if not config.option.backend:
+        config.option.backend = ["x11"]
+
+    ignore = config.option.ignore or []
+    if "wayland" not in config.option.backend:
+        ignore.append("test/backend/wayland")
+    if "x11" not in config.option.backend:
+        ignore.append("test/backend/x11")
+    config.option.ignore = ignore
+
+
 def pytest_generate_tests(metafunc):
     if "backend" in metafunc.fixturenames:
-        backends = metafunc.config.getoption("backend") or ["x11"]
+        backends = metafunc.config.option.backend
         metafunc.parametrize("backend_name", backends)
 
 
@@ -58,6 +68,12 @@ multimonitor = pytest.mark.parametrize("outputs", [1, 2], indirect=True)
 
 @pytest.fixture(scope="session")
 def xephyr(request, outputs):  # noqa: F841
+    if "x11" not in request.config.option.backend:
+        yield
+        return
+
+    from test.backend.x11.conftest import x11_environment
+
     kwargs = getattr(request, "param", {})
 
     with x11_environment(outputs, **kwargs) as x:
@@ -65,7 +81,13 @@ def xephyr(request, outputs):  # noqa: F841
 
 
 @pytest.fixture(scope="session")
-def wayland_session(outputs):  # noqa: F841
+def wayland_session(request, outputs):  # noqa: F841
+    if "wayland" not in request.config.option.backend:
+        yield
+        return
+
+    from test.backend.wayland.conftest import wayland_environment
+
     with wayland_environment(outputs) as w:
         yield w
 
@@ -73,11 +95,11 @@ def wayland_session(outputs):  # noqa: F841
 @pytest.fixture(scope="function")
 def backend(request, backend_name, xephyr, wayland_session):
     if backend_name == "x11":
-        b = XBackend({"DISPLAY": xephyr.display}, args=[xephyr.display])
+        from test.backend.x11.conftest import XBackend
+        yield XBackend({"DISPLAY": xephyr.display}, args=[xephyr.display])
     elif backend_name == "wayland":
-        b = WaylandBackend(wayland_session)
-
-    yield b
+        from test.backend.wayland.conftest import WaylandBackend
+        yield WaylandBackend(wayland_session)
 
 
 @pytest.fixture(scope="function")
```

m-col:
Seems to work brilliantly, thanks! That should be it for this PR then, assuming the tests pass.

ramnes:
Mhhh, the CI only runs the tests for X11 for now, right? If so, maybe we don't want to install all Wayland dependencies?

test/conftest.py
```diff
@@ -21,496 +21,99 @@
...
+    return request.param
+
+
+dualmonitor = pytest.mark.parametrize("outputs", [2], indirect=True)
+multimonitor = pytest.mark.parametrize("outputs", [1, 2], indirect=True)
 
 
 @pytest.fixture(scope="session")
-def xephyr(request, xvfb):  # noqa: F841
+def xephyr(request, outputs):  # noqa: F841
```
ramnes, line 70: I think we can drop the noqa comment here and in the next fixture as well.

test/scripts/window.py
```diff
@@ -1,105 +1,52 @@
...
+
+    python window.py --name <class> <title> <type>
+
+where <type> is "normal" or "notification"
+
+The window will close itself if it receives any key or button press events.
 """
-import sys
-import time
+# flake8: noqa
```
ramnes, line 14: Why tho? :D

ramnes (approved):

test/scripts/window.py
m-col: The imports need to be in a non-standard order, and there doesn't appear to be a way to ignore the error more specifically.

test/conftest.py
m-col: :+1:

m-col:
> Mhhh, the CI only runs the tests for X11 for now, right? If so, maybe we don't want to install all Wayland dependencies?

For now the dependencies only help mypy, but after this PR I will be working on fixing up the Wayland backend so that it passes the tests, at which point we can enable it. Running it locally it seems to be only a few tests that fail, and only for two or three bugs, so should be a small job.

ramnes:
Big props to you!

PR #2605, opened 2021-07-04
Add default wallpaper

elParaguayo:
Trawling through old issues and I noticed a comment about needing a default wallpaper to replace the black screen that accompanies the default config. 

Here's my submission.

Don't hold back on critiquing the artwork!


libqtile/resources/default_config.py
```diff
@@ -122,6 +123,9 @@
 )
 extension_defaults = widget_defaults.copy()
 
+wallpaper_path = Path(__file__).resolve() / ".." / "qtile-wallpaper.png"
```
m-col, line 126: Maybe stick it in resources/ alongside the default config?

m-col:
I love the wallpaper, so +1 from me.

libqtile/resources/default_config.py
elParaguayo: That's exactly where it is.

m-col: Oh yeah, brainfart :D

ramnes:
Won't this wallpaper confuse users as to what is a real window or not? :D

m-col:
> Won't this wallpaper confuse users as to what is a real window or not? :D

Lol really lacking faith in humanity huh? I think it's fine: 
image

Of course I think everyone has to be completely happy with it if we're going to have a default wallpaper so no worries if you'd prefer not to :)

Issue #2607, opened 2021-07-05
Simplifying base widget call_process method


yobleck:
# Issue description
The current version of the base widget's call_process method:
https://github.com/qtile/qtile/blob/225cca581a1ec83a89f0e9bf5ed8d5dcca723750/libqtile/widget/base.py#L291-L299
could be simplified as:
```
  def call_process(self, command, **kwargs):
      """
          This method uses `subprocess.check_output` to run the given command
          and return the string from stdout, which is decoded when using
          Python 3.
      """
      return subprocess.check_output(command, **kwargs, encoding="utf-8")
```
The encoding option for subprocess was added in python 3.6: https://docs.python.org/3/library/subprocess.html#frequently-used-arguments

and appears to be nearly identical to the decode built in:
https://docs.python.org/3/library/stdtypes.html#bytes.decode

Is it worth making this change to eliminate the temporary variable and the extra method call?

## Qtile version
0.17 and current dev version (225cca581a1ec83a89f0e9bf5ed8d5dcca723750)

m-col:
Sure! Could you send a patch? :smiley:

ramnes:
Please save your time and don't open issues for minor code changes but PRs directly. :)

PR #2608, opened 2021-07-05
simplify base widget call_process method

yobleck:
removes unnecessary temp variable and method call
see issue https://github.com/qtile/qtile/issues/2607

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -294,9 +294,7 @@ class _Widget(CommandObject, configurable.Configurable):
             and return the string from stdout, which is decoded when using
             Python 3.
         """
-        output = subprocess.check_output(command, **kwargs)
-        output = output.decode()
-        return output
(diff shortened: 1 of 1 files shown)


m-col:
This kinda makes the `check_output` method redundant. Maybe we should just remove it altogether?

yobleck:
> This kinda makes the `check_output` method redundant. Maybe we should just remove it altogether?

I don't understand. Redundant to what? call_process is used by a bunch of widgets and I can't find any other method that accomplishes the same thing.

m-col:
> I can't find any other method that accomplishes the same thing.

`subprocess.check_output` :)

yobleck:
So you mean we just get rid of call_process() in the base widget class and manually implement subprocess.check_output() in every widget that needs it? That seems like it could lead to fragmentation among how the widgets work.

ramnes:
Let's not debate on this, both work. :)

Thanks for the PR!

PR #2617, opened 2021-07-11
Create a PWVolume widget

flexagoon:
The current Volume widget only works with PulseAudio. This PR adds a
modified version of it which works with Pipewire through pamixer.

### DIFF
--- a/libqtile/widget/pipewire_volume.py
+++ b/libqtile/widget/pipewire_volume.py
@@ -1,0 +1,186 @@
+import subprocess
+
+from libqtile import bar
+from libqtile.widget import base
+
+__all__ = [
+    'PWVolume',
+]
+
+class PWVolume(base._TextBox):
+    """Widget that can display and change volume when using pipewire
+
+    By default, this widget uses ``pamixer`` to get and set the volume so users
+    will need to make sure this is installed. Alternatively, users may set the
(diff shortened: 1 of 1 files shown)


ramnes:
Before anything, please inherit from Volume so that most of the code isn't duplicated, and do that we can see what the difference is. Also, did you try the PulseVolume widget?

flexagoon:
What's PulseVolume? It's not in the documentation

ramnes:
> What's PulseVolume?

https://github.com/qtile/qtile/blob/master/libqtile/widget/pulse_volume.py

> It's not in the documentation

https://github.com/qtile/qtile/pull/2476

grtcdr:
Any progress on this feature? I've had trouble in the past getting pamixer to display volume in the bar.

elParaguayo:
Now in the docs: http://docs.qtile.org/en/latest/manual/ref/widgets.html#pulsevolume

grtcdr:
> Now in the docs: http://docs.qtile.org/en/latest/manual/ref/widgets.html#pulsevolume

Great stuff! 🚀

ramnes:
Alright, I'll go ahead and close this assuming that the PulseVolume widget works with Pipewire. Feel free to comment if it doesn't!

PR #2618, opened 2021-07-11
Get Wayland backend passing all remaining tests and enable in CI

m-col:
These are the remaining changes to the tests and Wayland backend that get them all working together. Most of them are just some small miscellaneous changes that were needed to pass a test. The final state is a fully passing Wayland backend that is tested alongside the X backend when run via tox, i.e. in github actions.


ramnes (approved):
LGTM

PR #2621, opened 2021-07-13
X11 _NET_CLIENT_LIST should list all existing windows

m-col:
This hint should list all windows whether they are hidden or not. The
information is used for pagers and other tools, so let's not tell them
about Internal windows.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -418,7 +418,9 @@ class Core(base.Core):
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
         """
-        wids = [wid for wid, c in windows_map.items() if not c.hidden]  # type: ignore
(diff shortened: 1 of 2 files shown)


ramnes:
What problem does that solve exactly?

m-col:
It fixes a bug that I introduced the other day.... :| 

This change: https://github.com/qtile/qtile/commit/9a363900ead14845a2ca3bbd154058c3acc583bc#diff-71a17283e4aa06f2d83c5904b5e7f1689c7e4f6437178ead9d3e3c79a65f7280R421

I misread `if c.group` as `if c.group is self.current_group` i.e. if the window is visible, but actually it should be all normal and static windows, whether visible or not. This is used e.g. by `wmctrl`, `rofi -show window` etc to get information on windows.

PR #2623, opened 2021-07-15
Set the default value of json key to False for wttr widget.

Fabien-Bernard:
At the moment, the default value of the json key is `True` and with this value the widget cannot operate correctly without the user setting the json key to `False` in its config. This is not intuitive as the documentation indicates that the default value of this key is `False`.
The first commit set the default value to False for this key.
Originated from : https://groups.google.com/g/qtile-dev/c/ErfiUphv-So

The second commit just corrects a not related typo of the documentation.


ramnes:
Thanks!

PR #2630, opened 2021-07-17
Allow keyboard layout widget to specify variant under Wayland

flacjacket:
When running setxkbmap under the X11 backend, the keyboard variant is
passed as the argument to the binary after the layout name, and so is
handled by splitting the layout name on the space and passing these
along to the binary.  For the `keymap_new_from_names` call, the variant
needs to be explicitly passed in as a separate argument.  With this
change, perform the equivalent operation for the layout name string
passed to the wayland backend of the keyboard layout widget to parse out
the variant when it is specified and forward it through to the xkbcommon
call.

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -99,7 +99,7 @@ class _X11LayoutBackend(_BaseLayoutBackend):
 
 
 class _WaylandLayoutBackend(_BaseLayoutBackend):
-    def __init__(self, qtile: Qtile):
+    def __init__(self, qtile: Qtile) -> None:
         self.set_keymap = qtile.core.set_keymap  # type: ignore
         self._layout: str = ""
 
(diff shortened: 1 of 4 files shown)


flacjacket:
I realize this also brought in a lot of changes from `black` that just ran in my editor, I'm happy to roll those out if that is helpful.

libqtile/widget/keyboardlayout.py
```diff
@@ -99,15 +99,20 @@ def set_keyboard(self, layout: str, options: Optional[str]) -> None:
...
         self._layout: str = ""
 
     def get_keyboard(self) -> str:
         return self._layout
 
     def set_keyboard(self, layout: str, options: Optional[str]) -> None:
-        self.set_keymap(layout, options)
+        maybe_variant: Optional[str] = None
+        if " " in layout:
+            layout_name, _, maybe_variant = layout.partition(" ")
```
m-col, line 112: Why `partition` over `split`?

m-col (approved):
Awesome, LGTM, especially black's will :)

libqtile/widget/keyboardlayout.py
flacjacket: Right, initially I had always done `layout.partition(" ")`, which is nice in that it always returns a 3-tuple, unlike split which would just return the input as a 1-tuple is the separator character isn't present, but then I had to transform the empty string to `None` to pass to `set_keymap`, which the type checker doesn't like, and you can't annotate a single part of the tuple when descructuring, like `layout_name, _, (maybe_variant: Optional[str]) = layout.partition(" ")`. Ultimately this felt like the better logical layout, but now that we know `" "` is in the layout, using `split` makes sense here.

m-col: Fair - when reading this my next question was going to be "ooh how are we going to limit the splits to 2 items?". I see now there is an argument for that.

m-col:
Cool thanks!

PR #2635, opened 2021-07-18
Document dependencies better

### LINKED ISSUE #2466: Clarify docs on what programs are required for tests
# Issue description

It isn't very clear what programs are required for running the tests. 

The top section of this page seems to be the closest to listing what is required, though it doesn't expliclity say that these are needed for tests (in fact it's not really correct saying they are required for hacking on Qtile): http://docs.qtile.org/en/latest/manual/hacking.html

Perhaps we should have a section on that specifically listing dependencies for tests and whether they are optional (tests will be skipped, not failed, if they are not available).

# Qtile version

0.18.0

# The deps

* `pprint` is needed for test_display_kb in test_qtile_cmd.py
* PyGObject, gtk3?

m-col:
Add dependency tables to docs
    
    This adds two tables to the docs that list out in one place (well, two
    places) the dependencies needed for 1. general runtime usage and 2.  the
    tests. These have replaces some more verbose explanation of how to
    install individual things, the assumptions being that people know how to
    install from pip, and people know how to find packages for their own
    distro if they know the name of the package for Ubuntu.
    
    The first table is on the installation page and lists core dependencies,
    whereas the table listing tests dependencies is on the hacking page and
    lists additional dependencies.
    
    What is *not* included is widget dependencies. There are many, all are
    optional, and they change more quickly than the rest of the deps, so
    splitting where they are documented between two places is probably a bad
    idea. However, many of the dependencies are monkeypatched for testing
    and so fortunately most are not needed for tests. Those are are needed
    for tests are in the test deps table.
    
    Closes #2466.

--

also two more small docs commits

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,6 +10,7 @@ Features
 * Simple, small and extensible. It's easy to write your own layouts,
   widgets and commands.
 * Configured in Python.
+* Runs as an X11 WM or a Wayland compositor.
 * Command shell that allows all aspects of Qtile to be managed and
   inspected.
 * Complete remote scriptability - write scripts to set up workspaces,
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -55,11 +55,10 @@ Miscellaneous
     manual/faq
     manual/license
 
-How To
-======
+Tips & Tricks
+=============
 
 .. toctree::
-    :caption: Tips & Tricks
     :maxdepth: 1
 
     manual/howto/widget
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -7,22 +7,39 @@ Hacking on Qtile
 Requirements
 ============
 
-All tests require `pytest <https://docs.pytest.org>`_ and ``PyGObject``. These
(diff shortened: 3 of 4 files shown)


elParaguayo:
Thanks for doing this. I'll take a proper look later as I've got a container set up at home to run builds on Arch so I can compare your list to what I need to get a successful build (including tests).

m-col:
Might be worth clarifying what's up with `libpulse-dev`, which is installed in the CI, and is required for the `scripts/ffi_build` script. I guess as it's a shell script the PulseVolume widget compilation will fail but the script will continue, however the traceback might cause confusion. It looks like there aren't any tests for that widget, and in fact it is explicitly excluded from `test/widgets/test_widget_init_configure.py`. Perhaps we should catch the compilation error and print a warning instead, and then remove `libpulse-dev` from the CI environment. It will need to be included as a build requirement for packaging, though.

tych0:
Additional clarifications welcome, seems like there's no reason not to merge this now so I'll just go ahead and do it :)

Issue #2622, opened 2021-07-14
Mouse clicks get Lost


mrossinek:
# Issue description

I am experiencing a very odd issue: randomly, mouse clicks appear to get lost..
After a random while, a mouse click will not get registered. This can result (e.g.) in a link in the browser not being clicked properly. I have to press `Esc` multiple times or restart qtile to get the mouse to register clicks again.

I can see this issue on two different machines:

1. Running Arch Linux:
    ```
    % uname -r
    5.12.14-arch1-1
    % pacman -Q xorg-server
    xorg-server 1.20.11-1
    ```

2. Running Fedora 33 (also using the X11 backend) [I don't currently have the detailed data here but it also uses the 5.12.* kernel]

# Qtile version

Compiled from master @ 839771fb0f0a281ac7c766bf2ab81a72241a7b43

# Stack traces

Unfortunately, I am unable to get any log messages from Qtile..
However, I found a more or less reliable way to test this issue is to run `xev` and continuously click. At some point, no more `ButtonPress` or `ButtonRelease` events will get registered.

I have bisected between v0.17.0 and latest master and found the first bad commit to be b2b1ed83e66ee093b8df9104df829b4e27528e42. Reverting it does indeed seem to fix things for me.

# Configuration

https://gitlab.com/mrossinek/dotfiles/-/blob/master/qtile/.config/qtile/config.py

elParaguayo:
There's nothing in that commit about mouse clicks so I really don't see how that could cause the issue you're seeing it (and I'm not just saying that because it's my commit!).

Let me see if I can recreate the issue you're seeing.

elParaguayo:
How many times do you have to click? I've been doing it for about the last 5 minutes and I'm not seeing any clicks that aren't being registered.

mrossinek:
Usually the issue appears after less than a minute. But it being random it is really hard to say.
I agree that the commit is unrelated from mouse clicks (hence I am also confused) but I found it suspiciously reproducible during bisecting.. :/
I still need to check whether reverting this commit works in the long term. I will update this issue in a few days after daily driving but I wanted to open the issue in case anyone else sees something similar

elParaguayo:
OK - let us know how you get on as I can't reproduce this issue.

mrossinek:
[...]

PR #2640, opened 2021-07-20
Changed ThermalSensor to use psutil

comfix:
ThermalSensor widget now uses psutil instead of the sensors command via
subprocess.
This also fixes a bug on gentoo which rendered the widget unusable due
to a utf-8 decoding error.

### DIFF
--- a/libqtile/widget/sensors.py
+++ b/libqtile/widget/sensors.py
@@ -24,10 +24,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import re
-from subprocess import CalledProcessError
+import psutil
 
-from libqtile.log_utils import logger
 from libqtile.widget import base
 
 
@@ -38,6 +36,10 @@ class ThermalSensor(base.InLoopPollText):
     You can get a list of the tag_sensors executing "sensors" in your terminal.
     Then you can choose which you want, otherwise it will display the first
     available.
+
+    Widget requirements: psutil_.
+
+    .. _psutil: https://pypi.org/project/psutil/
     """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
@@ -58,16 +60,6 @@ class ThermalSensor(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(ThermalSensor.defaults)
-        self.sensors_temp = re.compile(
-            (r"\n([\w ]+):"  # Sensor tag name
-             r"\s+[+|-]"     # temp signed
-             r"(\d+\.\d+)"   # temp value
-             "({degrees}"   # degree symbol match
-             "[C|F])"       # Celsius or Fahrenheit
-             ).format(degrees=u"\xb0"),
-            re.UNICODE | re.VERBOSE
-        )
-        self.value_temp = re.compile(r"\d+\.\d+")
         temp_values = self.get_temp_sensors()
         self.foreground_normal = self.foreground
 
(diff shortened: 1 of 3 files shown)


tych0:
Can you add a doc to the docstring similar to other widgets which have external dependencies that mentions psutil?

comfix:
Yes of course.
Totally forgot about that.

m-col:
I guess this means we don't need `lm_sensors` in the CI then? Could you remove it from .github/workflows/ci.yml?

elParaguayo:
If you're removing sensors then I think you need to remove this test too:
https://github.com/qtile/qtile/blob/466022c91e1ec42e1f3c21103b35906624dc9ac2/test/widgets/test_misc.py#L60-L77

Ideally, we'd have a new test for the changes to the widget.

elParaguayo:
The docstring still refers to `lm-sensors` and identifying tags via `sensors`. This probably needs to be updated and you may want to explain how to find tag names using `psutil`.

libqtile/widget/sensors.py
```diff
@@ -81,32 +73,19 @@ def __init__(self, **config):
...
-            name = name.strip()
-            temperature_values[name] = temp, symbol
-        return temperature_values
+
+        temperature_list = {}
+        temps = psutil.sensors_temperatures(fahrenheit=not self.metric)
+        unit = '°C' if self.metric else '°F'
+        for kernel_module in temps:
+            for sensor in temps[kernel_module]:
+                temperature_list[sensor.label] = (str(round(sensor.current, 1)), unit)
```
elParaguayo, line 86: On my laptop, there are multiple sensors with an empty label. This only saves one of them in the `temperature_list` dict and it's not clear which module it's attached too.

elParaguayo:
Can you also rebase onto the latest master as that should fix your test failures.

m-col:
Do these changes change the way the widget is used inside configs? And if so, do we need a migration for those changes?

elParaguayo:
I don't think so. The config is the same. What I don't know is whether the tag names are the same for sensors and psutil.

libqtile/widget/sensors.py
comfix: I added an indexed named generation for this case.
The problem should have been occurring with the old implementation as well as it is using the same back-end for its values.

libqtile/widget/sensors.py
```diff
@@ -81,9 +81,14 @@ def get_temp_sensors(self):
         temperature_list = {}
         temps = psutil.sensors_temperatures(fahrenheit=not self.metric)
         unit = '°C' if self.metric else '°F'
+        empty_index = 0
         for kernel_module in temps:
             for sensor in temps[kernel_module]:
-                temperature_list[sensor.label] = (str(round(sensor.current, 1)), unit)
+                label = sensor.label
+                if not label:
+                    label = 'UNKNOWN-' + str(empty_index)
```
elParaguayo, line 89: Could you use `kernel_module` rather than `UNKNOWN` (but fallback to `UNKNOWN` if `kernel_module` is also blank)?

comfix:
@m-col @elParaguayo No this change does not break existing config as it supplies the values in the same manner as the old one did. I only changed the gathering of sensor values from running an external process to reading the sysfs (which is uses by `sensors` as well as it's backend)
And as other widgets also utilize `psutil` I don't think anybody will notice this change.

libqtile/widget/sensors.py
comfix: Sure thing.

elParaguayo:
Not sure what's happened to this PR but it's pulled in a lot of extra commits which shouldn't be here. Doesn't look like a clean rebase. What command did you use to rebase?

comfix:
I just used `git rebase upstream/master` but you are right it asked me to resolve a merge conflict.
I cant say what happened exactly, I do not rebase that often ^^'

mrossinek:
A friendly ping on this PR :innocent: This is the last piece missing to close #2622. Are there any plans to get this merged any time soon?

elParaguayo:
Friendly ping received. 

Let me run CI on this. If it passes, can you squash this into a single commit and I'll merge.

elParaguayo:
The CI error has been fixed by a recent commit so you'll need to rebase this onto the latest `master` and push it again. You may as well squash all your commits into a single one at the same time.

mrossinek:
@comfix this is your PR so I'll leave that up to you. If you don't have the time to handle this I can take over, but I won't unless you want me to :+1:

PR #2641, opened 2021-07-21
Fix moving bar when screens reconfigured

### LINKED ISSUE #2639: Bar moves when screen wakes up
# Issue description

When my screen goes to sleep (screen blank) and I wake it up by moving the mouse, my bar has moved up a few pixels.
This only happens when the bar is configured to have a gap `margin=[15, 0, 0, 0]`

IMPORTANT: The screen has to actually go to sleep for this to happen. (give it some time after screen blank)

# Qtile version

0.18.1.dev0+g8e7ecc0a.d20210719

# Stack traces

Nothing is written to `qtile.log` or `.xsession-errors`.

# Configuration

This can be reproduced with a minor change to the default config.
Just add some margin to the default bottom bar `margin=[15, 0, 0, 0]`

Reload the configuration.
To force a screen blank run `xset dpms force off`.
Wait until the screen actually goes to sleep.
Move the mouse to wake it up again.

The bar should now have moved up a few pixels.

elParaguayo:
Bug where Bar has a defined margin and screen is reconfigured resulting in bar being moved. Issue arises because when a Gap
is reconfigured, the size is adjusted by the margins but these adjustments should be made to the initial size, not the previously
adjusted size.

Fixes #2639

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -65,6 +65,7 @@ class Gap(CommandObject):
     def _configure(self, qtile, screen):
         self.qtile = qtile
         self.screen = screen
+        self.size = self.initial_size
         # If both horizontal and vertical gaps are present, screen corners are
         # given to the horizontal ones
         if screen.top is self:


elParaguayo:
Fix confirmed: https://github.com/qtile/qtile/issues/2639#issuecomment-884687178

tych0:
Great, thanks!

PR #2645, opened 2021-07-22
Ensure test_check runs on CI

elParaguayo:
`test_check` was being skipped because `mypy` was not installed in the test environment. Once `mypy` is installed, there was an error in `backend/base.py` which was identified.

`test_check` takes 5.5 mins to complete on my system. It's slooow! (EDIT: seems to run quicker on CI)

Also, I've probably used the wrong type hinting here but I want to see if the test runs successfully on CI (EDIT: It does!).

Happy to correct the hinting afterwards.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -428,6 +428,10 @@ class Internal(_Window, metaclass=ABCMeta):
 class Static(_Window, metaclass=ABCMeta):
     """A window bound to a screen rather than a group."""
     screen: config.Screen
+    x: Any
+    y: Any
+    width: Any
+    height: Any
 
     def __repr__(self):
         return "Static(name=%r, wid=%s)" % (self.name, self.wid)
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,7 @@ deps =
     pytest-cov >= 2.10.1
     setuptools >= 40.5.0
     xcffib >= 0.10.1
+    mypy
     bowler
     xkbcommon >= 0.3
     pywayland >= 0.4.4


m-col:
They are meant to be `int`s I'm afraid

elParaguayo:
I tried that. Got this:
```
libqtile/backend/x11/window.py:1092: error: Definition of "width" in base class "_Window" is incompatible with definition in base class "Static"
libqtile/backend/x11/window.py:1092: error: Definition of "x" in base class "_Window" is incompatible with definition in base class "Static"
libqtile/backend/x11/window.py:1092: error: Definition of "y" in base class "_Window" is incompatible with definition in base class "Static"
libqtile/backend/x11/window.py:1092: error: Definition of "height" in base class "_Window" is incompatible with definition in base class "Static"
libqtile/backend/x11/core.py:706: error: Subclass of "Static" and "_Window" cannot exist: would have incompatible method signatures
```

m-col:
Ah, right. I think `width` and `height` are properties defined on `_Window` and so they are inherited - only `x` and `y` need annotations.

elParaguayo:
That's not it either I'm afraid. I don't know how mypy is determining those types at the `_Window` level but it won't let me set the types in that class definition either.

tych0:
> test_check takes 5.5 mins to complete on my system. It's slooow! (EDIT: seems to run quicker on CI)

Yeah, this is because stubtest is fairly slow :)

Anyway, thanks for this. We used to run it in the mypy test, but looks like that got dropped at some point,

Issue #2481, opened 2021-05-22
Feature request for `master_match` improvements


nazar-pc:
# Feature request

After upgrade to 0.17.0 `Match` can only accept one string as `wm_class` and it causes inconveniences in my setup since I had to create extra layouts just to make it work the way I need.

Since these are related, I'd like to request 2 updates:
* [ ] make `master_match` an array of `Match`, so more complex setups are possible, similarly to previously existing functionality
* [ ] make it possible to specify `master_match` on per-group basis, so that it is possible to have single `Tile` layout with different `master_match` for different groups

# Qtile version

0.17.0

Fabien-Bernard:
I think the Match restriction to wm_class is already fixed by 7cc2f54

nazar-pc:
Not sure how that is related, also changelog says that `wm_class` can only be a single string now, not list.

Fabien-Bernard:
Forget my remark, you're right, my head mixed up the issues, sorry.

ramnes:
> make master_match an array of Match, so more complex setups are possible, similarly to previously existing functionality

LGTM, PR welcome. :)

nazar-pc:
One PR is #2659, please take a look once you have a free minute

PR #2659, opened 2021-07-28
Support a list of master matches in Tile layout

nazar-pc:
This implements the first half of #2481, allowing to restore pre-0.17.0 behavior with multiple matches

### DIFF
--- a/libqtile/layout/tile.py
+++ b/libqtile/layout/tile.py
@@ -30,6 +30,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from libqtile.config import Match
 from libqtile.layout.base import _SimpleLayoutBase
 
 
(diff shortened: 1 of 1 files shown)


nazar-pc:
Would be great if someone can finally review this, I have been using this patch since PR was submitted and didn't notice any issues.

libqtile/layout/tile.py
```diff
@@ -115,9 +115,16 @@ def down(self):
...
         if self.clients:
-            masters = [c for c in self.clients if match.compare(c)]
+            if type(master_match) is list:
+                masters = []
+                for c in self.clients:
+                    for match in master_match:
+                        if match.compare(c):
+                            masters.append(c)
+            else:
+                masters = [c for c in self.clients if master_match.compare(c)]
```
m-col, line 127: Perhaps if we make `self.master_match` a list if it isn't already in `__init__()`, then we can do something like this:
```suggestion
        if self.clients:
	    master_match = match or self.master_match
            if isinstance(master_match, Match):
		mastermatch = [mastermatch]
	    masters = []
	    for c in self.clients:
		for match in master_match:
		    if match.compare(c):
			masters.append(c)
```

Which i think is a bit easier to read

m-col:
Just a small comment on the code. Beside that, please could you update the doc for that option, and add a changelog entry for it? Then LGTM!

m-col:
Looks great, thanks!

PR #2661, opened 2021-07-28
Fix CI Test errors

elParaguayo:
dbus-next released a new version which is PEP561 compliant meaning we shouldn't be using "# type: ignore" for dbus-next imports.


m-col (approved):
LGTM pending tests

Issue #2656, opened 2021-07-27
Crash when font used by qtile is modified


cadadr:
# Issue description

When the font file that corresponds to the font used by Qtile is modified (in my case with `cp` from GNU coreutils 8.32, on Arch Linux), this leads to a crash. Logs are somewhat cryptic.

```
$ ls ~/.fonts
[ many fonts, including fonts from my Iosevka build]
$ cd /path/to/iosevka/checkout
$ # rebuilt font
$ cp dist/**/*/*.ttf ~/.fonts
$ # qtile crashes with SIGSEGV
```

# Qtile version

0.18.1.dev0+g8e7ecc0a.d20210719 (Arch repos)

# Stack traces
## `~/.xsession-errors`

I think this is the relevant bit:

```
/home/g/cf/systems/midori/desktop-setup.bash: line 101:   756 Segmentation fault      (core dumped) qtile start
[ 27-07-2021 10:11:24.121 x_create_picture_with_pictfmt_and_pixmap ERROR ] failed to create picture (X error 9 DRAWABLE request 139 minor 4 serial 901466)
xinit: connection to X server lost
```

## `~/.local/share/qtile/qtile.log`
This is the only log in there from today but feels like irrelevant:

```
2021-07-27 10:11:29,859 ERROR libqtile core.py:_xpoll():L330 Got an exception in poll loop
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/libqtile/backend/x11/core.py", line 298, in _xpoll
    ret = target(event)
  File "/usr/lib/python3.9/site-packages/libqtile/backend/x11/core.py", line 688, in handle_DestroyNotify
    if self.qtile.current_window is None:
  File "/usr/lib/python3.9/site-packages/libqtile/core/manager.py", line 522, in current_window
    return self.current_screen.group.current_window
AttributeError: 'Qtile' object has no attribute 'current_screen'
```

## backtrace
This is an excerpt from the top of the gdb stack trace that looks relevant. The whole thing is a bit long so I'll add it as a separate comment.

```
#0  0x00007ff6002deee9 in hb_ot_layout_table_find_feature_variations () from /usr/lib/libharfbuzz.so.0
#1  0x00007ff60030da4b in ?? () from /usr/lib/libharfbuzz.so.0
#2  0x00007ff600315699 in hb_shape_plan_create_cached2 () from /usr/lib/libharfbuzz.so.0
[...]

PR #2662, opened 2021-07-28
Document font update pitfall

cadadr:
If a font file associated to a font used by Qtile is updated, Qtile may crash.  The resulting logs might be somewhat confusing.  This commit documents this pitfall in three places, including the default Qtile configuration file so that users are aware of it and can avoid it.

This adresses #2656.

### DIFF
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -163,7 +163,10 @@ configuration variables that control specific aspects of Qtile's behavior:
       - dict(font='sans',
              fontsize=12,
              padding=3)
-      - Default settings for bar widgets.
+      - Default settings for bar widgets.  Note: if the font file
+        associated with the font selected here is modified while Qtile
+        is running, Qtile may segfault (for details see `issue #2656
+        <https://github.com/qtile/qtile/issues/2656>`_).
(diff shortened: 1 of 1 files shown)


docs/manual/config/index.rst
```diff
@@ -163,7 +163,10 @@ configuration variables that control specific aspects of Qtile's behavior:
       - dict(font='sans',
              fontsize=12,
              padding=3)
-      - Default settings for bar widgets.
+      - Default settings for bar widgets.  Note: if the font file
+        associated with the font selected here is modified while Qtile
+        is running, Qtile may segfault (for details see `issue #2656
+        <https://github.com/qtile/qtile/issues/2656>`_).
```
tych0, line 169: Can we just leave this one? Doesn't feel like we need to say the same thing three times.

cadadr: Sure :+1:  updated the patch.

ramnes:
Took the rebase on myself and merged with adeccf9a56478044fb3e23d9f55fc409ff96c7bf. Thanks!

cadadr:
You're welcome and thanks a lot!

PR #2675, opened 2021-08-02
Add border to bar

### LINKED ISSUE #2115: [Request] Add option to draw a border around the qtile bar
# Issue description

As per our reddit [discussion](https://www.reddit.com/r/unixporn/comments/kzd6zt/qtile_genome/gjpj43c?utm_source=share&utm_medium=web2x&context=3) @ramnes , I wanted to make a request to add a new option for the qtile bar to have a border drawn around it. I imagine this would work similar to polybar in that you can specify the border width, color, and sides to draw a border. To take it a little further, being able to draw multiple borders would be _fantastic_, but not necessary. From the screenshots in the linked post, I'm able to create a border effect by using three seperate [eww](https://github.com/elkowar/eww) windows; One for the top and bottom of the bar to create the bubble effect on the widgets, as well as a gray bottom border to outline the bar itself. While this works, it's not an ideal solution. If I need to restart qtile, then the eww windows that are creating padding end up underneath the bar and then I need to refresh them to get them back on top of the qtile bar. For those trying to get a similar effect, see my [autostart](https://github.com/Barbarossa93/Genome/blob/main/.config/qtile/autostart.sh) and eww [config](https://github.com/Barbarossa93/Genome/blob/4a08d3cfd0900807aefaa9f9241a6dbf926c549b/.config/eww/eww.xml#L149) for how I did it. You also don't need eww for this, you could also achieve the same effect with empty polybars.

I'd be happy to take a stab at this myself, but I'm no pythonista and fully expect to not be able to figure it out. If someone can tell me the relevant files to look into, I'll give it a shot anyway.

Alternatively (or in addition), creating padding/margins around the widgets such that the background color of the bar can "peek" out from underneath the widget (such that the widget total height is smaller than the full height of the bar) would also get the same effect and would be a fantastic addition to the bar config options.

# Qtile version
0.16.2.dev189+g18ff9899

elParaguayo:
Adds ability to draw border around bar. Border can be customised (size and colour) per edge.

Closes #2115

Anyone interested in trying this out?

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -183,49 +183,55 @@ class Bar(Gap, configurable.Configurable):
         if isinstance(self.border_width, int):
             self.border_width = [self.border_width] * 4
 
+        self._initial_margin = self.margin[:]
+        self.struts = [0, 0, 0, 0]
+        self._add_strut = False
+
         self.queued_draws = 0
         self.future = None
 
     def _configure(self, qtile, screen):
-        Gap._configure(self, qtile, screen)
+        # We only want to adjust margin sizes once unless there's a new strut
+        if not self._configured or self._add_strut:
+            Gap._configure(self, qtile, screen)
 
-        if sum(self.margin) or sum(self.border_width):
+            if sum(self._initial_margin) or sum(self.border_width) or self._add_strut:
 
-            try:
-                # Check if colours are valid but don't convert to rgba here
-                if isinstance(self.border_color, list) and len(self.border_color) == 4:
-                    [rgb(col) for col in self.border_color]
+                try:
+                    # Check if colours are valid but don't convert to rgba here
+                    if isinstance(self.border_color, list) and len(self.border_color) == 4:
+                        [rgb(col) for col in self.border_color]
+                    else:
+                        rgb(self.border_color)
+                        self.border_color = [self.border_color] * 4
+                except (ValueError, TypeError):
+                    logger.warning("Invalid border_color specified. Borders will not be displayed.")
+                    self.border_width = [0, 0, 0, 0]
+
+                # Increase the margin size for the border. The border will be drawn
+                # in this space so the empty space will just be the margin.
+                self.margin = [m + b + s for m, b, s in zip(self._initial_margin, self.border_width, self.struts)]
+
+                if self.horizontal:
+                    self.x += self.margin[3] - self.border_width[3]
+                    self.width -= (self.margin[1] + self.margin[3])
+                    self.length = self.width
+                    if self.size == self.initial_size:
+                        self.size += (self.margin[0] + self.margin[2])
+                    if self.screen.top is self:
+                        self.y += self.margin[0] - self.border_width[0]
+                    else:
+                        self.y -= self.margin[2] + self.border_width[2]
+
                 else:
-                    rgb(self.border_color)
-                    self.border_color = [self.border_color] * 4
-            except (ValueError, TypeError):
-                logger.warning("Invalid border_color specified. Borders will not be displayed.")
-                self.border_width = [0, 0, 0, 0]
-
-            # Increase the margin size for the border. The border will be drawn
-            # in this space so the empty space will just be the margin.
-            self.margin = [m + b for m, b in zip(self.margin, self.border_width)]
-
-            if self.horizontal:
-                self.x += self.margin[3] - self.border_width[3]
-                self.width -= (self.margin[1] + self.margin[3])
-                self.length = self.width
-                if self.size == self.initial_size:
-                    self.size += (self.margin[0] + self.margin[2])
-                if self.screen.top is self:
                     self.y += self.margin[0] - self.border_width[0]
-                else:
-                    self.y -= self.margin[2] + self.border_width[2]
-
-            else:
-                self.y += self.margin[0] - self.border_width[0]
-                self.height -= (self.margin[0] + self.margin[2])
-                self.length = self.height
-                self.size += (self.margin[1] + self.margin[3])
-                if self.screen.left is self:
-                    self.x += self.margin[3]
-                else:
-                    self.x -= self.margin[1]
+                    self.height -= (self.margin[0] + self.margin[2])
+                    self.length = self.height
+                    self.size += (self.margin[1] + self.margin[3])
+                    if self.screen.left is self:
+                        self.x += self.margin[3]
+                    else:
+                        self.x -= self.margin[1]
 
         width = self.width + (self.border_width[1] + self.border_width[3])
         height = self.height + (self.border_width[0] + self.border_width[2])
@@ -289,6 +295,7 @@ class Bar(Gap, configurable.Configurable):
         self.draw()
         self._resize(self.length, self.widgets)
         self._configured = True
+        self._add_strut = False
 
     def _configure_widget(self, widget):
         configured = True
@@ -576,16 +583,11 @@ class Bar(Gap, configurable.Configurable):
     def adjust_for_strut(self, size):
         if self.size:
             self.size = self.initial_size
-        if not self.margin:
-            self.margin = [0, 0, 0, 0]
-        if self.screen.top is self:
-            self.margin[0] += size
-        elif self.screen.bottom is self:
-            self.margin[2] += size
-        elif self.screen.left is self:
-            self.margin[3] += size
-        else:  # right
-            self.margin[1] += size
+        for i, gap in enumerate(["top", "right", "bottom", "left"]):
+            if getattr(self.screen, gap) is self:
+                self.struts[i] += size
+
+        self._add_strut = True
 
     def cmd_fake_button_press(self, screen, position, x, y, button=1):
         """
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -25,7 +25,7 @@ import typing
 from libqtile import configurable
 from libqtile.command.base import CommandObject, ItemT
 from libqtile.log_utils import logger
-from libqtile.utils import has_transparency
+from libqtile.utils import has_transparency, rgb
 
 if typing.TYPE_CHECKING:
     from libqtile.widget.base import _Widget
@@ -162,6 +162,8 @@ class Bar(Gap, configurable.Configurable):
         ("background", "#000000", "Background colour."),
         ("opacity", 1, "Bar window opacity."),
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
+        ("border_color", "#000000", "Border colour as str or list of str [N E S W]"),
+        ("border_width", 0, "Width of border as int of list of ints [N E S W]")
     ]
 
     def __init__(self, widgets, size, **config):
@@ -175,34 +177,58 @@ class Bar(Gap, configurable.Configurable):
         self.size_calculated = 0
         self._configured = False
 
+        if isinstance(self.margin, int):
+            self.margin = [self.margin] * 4
+
+        if isinstance(self.border_width, int):
+            self.border_width = [self.border_width] * 4
+
         self.queued_draws = 0
         self.future = None
 
     def _configure(self, qtile, screen):
         Gap._configure(self, qtile, screen)
 
-        if self.margin:
-            if isinstance(self.margin, int):
-                self.margin = [self.margin] * 4
+        if sum(self.margin) or sum(self.border_width):
+
+            try:
+                # Check if colours are valid but don't convert to rgba here
+                if isinstance(self.border_color, list) and len(self.border_color) == 4:
+                    [rgb(col) for col in self.border_color]
+                else:
+                    rgb(self.border_color)
+                    self.border_color = [self.border_color] * 4
+            except (ValueError, TypeError):
+                logger.warning("Invalid border_color specified. Borders will not be displayed.")
+                self.border_width = [0, 0, 0, 0]
+
+            # Increase the margin size for the border. The border will be drawn
+            # in this space so the empty space will just be the margin.
+            self.margin = [m + b for m, b in zip(self.margin, self.border_width)]
+
             if self.horizontal:
-                self.x += self.margin[3]
-                self.width -= self.margin[1] + self.margin[3]
+                self.x += self.margin[3] - self.border_width[3]
+                self.width -= (self.margin[1] + self.margin[3])
                 self.length = self.width
                 if self.size == self.initial_size:
-                    self.size += self.margin[0] + self.margin[2]
+                    self.size += (self.margin[0] + self.margin[2])
                 if self.screen.top is self:
-                    self.y += self.margin[0]
+                    self.y += self.margin[0] - self.border_width[0]
                 else:
-                    self.y -= self.margin[2]
+                    self.y -= self.margin[2] + self.border_width[2]
+
             else:
-                self.y += self.margin[0]
-                self.height -= self.margin[0] + self.margin[2]
+                self.y += self.margin[0] - self.border_width[0]
+                self.height -= (self.margin[0] + self.margin[2])
                 self.length = self.height
-                self.size += self.margin[1] + self.margin[3]
+                self.size += (self.margin[1] + self.margin[3])
                 if self.screen.left is self:
                     self.x += self.margin[3]
                 else:
                     self.x -= self.margin[1]
+
+        width = self.width + (self.border_width[1] + self.border_width[3])
+        height = self.height + (self.border_width[0] + self.border_width[2])
 
         for w in self.widgets:
             # Executing _test_orientation_compatibility later, for example in
@@ -214,7 +240,7 @@ class Bar(Gap, configurable.Configurable):
         if self.window:
             # We get _configure()-ed with an existing window when screens are getting
             # reconfigured but this screen is present both before and after
-            self.window.place(self.x, self.y, self.width, self.height, 0, None)
+            self.window.place(self.x, self.y, width, height, 0, None)
         else:
             # Whereas we won't have a window if we're startup up for the first time or
             # the window has been killed by us no longer using the bar's screen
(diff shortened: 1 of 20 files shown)


mbenford:
Nice! This would be particularly useful to highlight which display has focus in a more obvious way than, say, using the GroupBox widget.

elParaguayo:
Has anyone tried this? Would be good to get some feedback!

elParaguayo:
This one would be a bit harder to include in my `qtile-extras` repo. It would be possible to provide a subclassed `Bar` but the issue is, with the current implementation, I need to edit widgets so they also include `offsety` in their `drawer.draw` call. 

I can take a look to see if I can get this all contained in the `Bar` anyway as that would be neater.

libqtile/bar.py
```diff
@@ -175,34 +177,50 @@ def __init__(self, widgets, size, **config):
...
     def _configure(self, qtile, screen):
         Gap._configure(self, qtile, screen)
 
-        if self.margin:
-            if isinstance(self.margin, int):
-                self.margin = [self.margin] * 4
+        if sum(self.margin) or sum(self.border):
+
+            if isinstance(self.border_colour, str):
+                self.border_colour = [self.border_colour] * 4
```
m-col, line 194: Perhaps this should accept other forms i.e. `Union[str, Tuple[int, int, int], Tuple[int, int, int, float]]`

libqtile/bar.py
```diff
@@ -175,34 +177,50 @@ def __init__(self, widgets, size, **config):
...
                 self.length = self.height
-                self.size += self.margin[1] + self.margin[3]
+                self.size += (self.margin[1] + self.margin[3])
                 if self.screen.left is self:
                     self.x += self.margin[3]
                 else:
                     self.x -= self.margin[1]
 
+        width = self.width + (self.border[1] + self.border[3])
+        height = self.height + (self.border[0] + self.border[2])
```
m-col, line 222: These brackets seem out of place :stuck_out_tongue:

m-col (commented):
This all looks great. My main question is why do the border drawing with cairo as opposed to with X11/wlroots like regular windows?

elParaguayo:
> This all looks great. My main question is why do the border drawing with cairo as opposed to with X11/wlroots like regular windows?

Because I _thought_ (and I'm really hoping I was right) that you can't draw individual borders, different colours, different thicknesses etc with those methods.

If I'm wrong then I've obviously wasted some time...

m-col:
I see - like the second bar in your screenshot. You are right - you can't do that in Qtile. OK, no worries then. I'll give it a test drive!

elParaguayo:
> I see - like the second bar in your screenshot. You are right - you can't do that in Qtile. OK, no worries then. I'll give it a test drive!

It does make things more complicated but I think it was something that people had requested.

Would be interested in your thoughts about whether we can move most of the logic to `Bar`. I'd prefer not to have to edit any widgets for this.

libqtile/bar.py
elParaguayo: Um... yeah. Not sure what I was thinking there.

libqtile/bar.py
```diff
@@ -162,6 +162,8 @@ class Bar(Gap, configurable.Configurable):
         ("background", "#000000", "Background colour."),
         ("opacity", 1, "Bar window opacity."),
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
+        ("border_colour", "#000000", "Border colour as str or list of str [N E S W]"),
+        ("border", 0, "Width of border")
```
m-col, line 166: Maybe also mention that `border` can take a list?

m-col:
I'm just investigating configurable option names with the hope of ensuring consistency and errr... well, we aren't _too_ far off I guess:
```
~/git/qtile/libqtile ▋▌▍ grep -r '("border'
bar.py:        ("border_colour", "#000000", "Border colour as str or list of str [N E S W]"),
bar.py:        ("border", 0, "Width of border")
widget/tasklist.py:        ("`", 2, "Current group border width"),
widget/tasklist.py:        ("border", "215578", "Border colour"),
widget/graph.py:        ("border_color", "215578", "Widget border color"),
widget/graph.py:        ("border_width", 2, "Widget border width"),
widget/groupbox.py:        ("borderwidth", 3, "Current group border width"),
widget/groupbox.py:    defaults = [("border", "000000", "group box border color")]
layout/xmonad.py:        ("border_focus", "#ff0000", "Border colour(s) for the focused window."),
layout/xmonad.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/xmonad.py:        ("border_width", 2, "Border width."),
layout/ratiotile.py:        ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
layout/ratiotile.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/ratiotile.py:        ("border_width", 1, "Border width."),
layout/tree.py:        ("border_width", 2, "Width of the border"),
layout/floating.py:        ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
layout/floating.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/floating.py:        ("border_width", 1, "Border width."),
layout/bsp.py:        ("border_focus", "#881111", "Border colour(s) for the focused window."),
layout/bsp.py:        ("border_normal", "#220000", "Border colour(s) for un-focused windows."),
layout/bsp.py:        ("border_width", 2, "Border width."),
layout/matrix.py:        ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
layout/matrix.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/matrix.py:        ("border_width", 1, "Border width."),
layout/stack.py:        ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
layout/stack.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/stack.py:        ("border_width", 1, "Border width."),
layout/columns.py:        ("border_focus", "#881111", "Border colour(s) for the focused window."),
layout/columns.py:        ("border_normal", "#220000", "Border colour(s) for un-focused windows."),
layout/columns.py:        ("border_focus_stack", "#881111",
layout/columns.py:        ("border_normal_stack", "#220000",
layout/columns.py:        ("border_width", 2, "Border width."),
layout/columns.py:        ("border_on_single", False, "Draw a border when there is one only window."),
layout/tile.py:        ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
layout/tile.py:        ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
layout/tile.py:        ("border_width", 1, "Border width."),
```
The layouts are consistent with each other. `TaskList` and `GroupBox` use `border` for colours whereas `Graph` uses `border_color`, consistent with the layouts. I'd be in favour of renaming the formers' options to `border_color`. Similarly those two also use `borderwidth` instead of `border_width` like the others. Anyway, tangents. For the above findings may I suggest `border_color` (urgh, I know) and `border_width` instead of `border_colour` and `border`?

elParaguayo:
Are we going to have the `color` vs `colour` debate too?

m-col:
Just testing it with `wephyr` and `border=[2, 4, 6, 8], border_colour="#ff0000"`: 
image

It's mostly doing what it's told, but there seem to be two glitches: the `offsety` of the "Widget crashed" widget by the looks of it, and the black line at the far left - perhaps the border is being drawn one pixel too low?

> Are we going to have the color vs colour debate too?

Hey, I'd love for it all to be `colour` but I think we lost that one :crying_cat_face:

elParaguayo:
OK - let me test more in Wayland.

m-col:
Just tested with `xephyr` in X and I get the exact same results (raising an exception manually in Systray's _configure to make it "crash").

elParaguayo:
Awesome. Thanks for those pointers. Some work to do...

elParaguayo:
OK - the `ConfigError` and border gaps should both be fixed.

elParaguayo:
Thinking out loud: Could we remove all the `offsety` stuff from the widgets and then just do some translation of the drawer in `Bar`? Would be neater!

I'll take a look.

elParaguayo:
> Thinking out loud: Could we remove all the `offsety` stuff from the widgets and then just do some translation of the drawer in `Bar`? Would be neater!
> 
> I'll take a look.

So, you can do a `translate` when doing `self.bar.draw()`but when the widget just draws itself because the size is unchanged then you need the offset. Not sure this can be fixed without editing the widgets.

m-col:
Changelog entry pls :)

elParaguayo:
The one bit I don't love about this is I have to add `offsety` to the `draw` calls. This is because the bar needs a bigger `Drawer` to include the borders but the widgets are only drawn in the middle bit and so they need to offset accordingly.

I wonder if there's a better way. This is why I'm hesitating at this stage.

Any custom widgets that have their own `draw` method will not render correctly without changes.

As for being "kinda hacky", I subclassed `Bar` (fine) but needed to inject code into widgets:
https://github.com/elParaguayo/qtile-extras/blob/3a8bb1a0894f877680b31e032a2bd9469dceac5a/qtile_extras/bar.py#L256-L288

m-col:
I think it's fine to expect users who have defined custom widgets to simply make that adjustment. At least with this change Qtile won't crash and they'll see immediately the problem. Also, they will only see this at all if they give the bar some borders, in which case they are already aware! Perhaps add a comment to the changelog about this too, and perhaps also the docstring of widgets if necessary.

Also for the sake of user awareness, maybe add `border_width` and `borde_color` arguments on commented out lines to the default config?

jwijenbergh:
My icons in the systray and statusnotifier are placed incorrectly:

image

The following patch seems to fix it for both (note I am only using a horizontal bar so I'm not sure what happens with a vertical one):
```diff
diff --git a/libqtile/widget/statusnotifier.py b/libqtile/widget/statusnotifier.py
index e3d3a26d..035cc32d 100644
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -642,7 +642,7 @@ class StatusNotifier(base._Widget):
 
                 xoffset += self.icon_size + self.padding
 
-            self.drawer.draw(offsetx=self.offset, width=self.length)
+            self.drawer.draw(offsetx=self.offset, offsety=self.offsety, width=self.length)
 
         else:
             xoffset = (self.bar.width - self.icon_size) // 2
diff --git a/libqtile/widget/systray.py b/libqtile/widget/systray.py
index f52a8dfb..e6fb4f4c 100644
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -213,7 +213,7 @@ class Systray(window._Window, base._Widget):
             icon.window.set_attribute(backpixmap=self.drawer.pixmap)
             icon.place(
                 self.offset + xoffset,
-                self.bar.height // 2 - self.icon_size // 2,
+                self.bar.height // 2 - self.icon_size // 2 + self.offsety,
                 icon.width, self.icon_size,
                 0,
                 None
```

elParaguayo:
Thanks. Missed StatusNotifier because I made this PR before that was merged into the master. Also, StatusNotifier can actually be oriented vertically too so there's an additional line that needs changing.

No excuses on Systray though. Just missed that one entirely!

jwijenbergh:
Works fine, thank you! And no problem, I'm just glad we found it before this gets merged :+1:

jwijenbergh:
I seem to have a bug with this. When I plug in a new monitor, set it with xrandr, the bar will shift to the bottom a bit. This does not happen when I set the border to 0. I have to restart qtile with the keybinding to make it go back to normal.

My normal setup:

qtile 1

After screen has been disconnected or added and configured with arandr/xrandr:

qtile 2

`Reconfigure_screens` is set to `True`

elParaguayo:
Thanks. Will see if I can recreate.

elParaguayo:
The issue is caused by `cmd_reconfigure_screens` triggering `Bar._configure` which, in turn, adds borders to the margin size. So multiple calls to this continuously impact the margins.

Will push a fix for you to test shortly.

m-col:
All looks good. Does the documentation page on creating custom widgets need to be updated or does it still apply as is? In particular the `offsetx` and `offsety` stuff. If there are rules people need to be aware of it'd also be good to add those into the changelog.

elParaguayo:
> All looks good. Does the documentation page on creating custom widgets need to be updated or does it still apply as is? In particular the `offsetx` and `offsety` stuff. If there are rules people need to be aware of it'd also be good to add those into the changelog.

That's in there already:
https://github.com/qtile/qtile/pull/2675/files#diff-f375f929704fd307705f114f56bbac4aad669c679bb3fe5b202c442e2f2f52f4R204-R207

jwijenbergh:
> The issue is caused by `cmd_reconfigure_screens` triggering `Bar._configure` which, in turn, adds borders to the margin size. So multiple calls to this continuously impact the margins.
> 
> Will push a fix for you to test shortly.

It is now fixed with https://github.com/qtile/qtile/pull/2675/commits/cdfc3e46ef2f1be0f9d7fe315823317a52f5edbf

Thank you

libqtile/bar.py
```diff
@@ -175,34 +177,64 @@ def __init__(self, widgets, size, **config):
...
+
+                try:
+                    # Check if colours are valid but don't convert to rgba here
+                    if isinstance(self.border_color, list) and len(self.border_color) == 4:
+                        [rgb(col) for col in self.border_color]
+                    else:
+                        rgb(self.border_color)
+                        self.border_color = [self.border_color] * 4
+                except (ValueError, TypeError):
+                    logger.warning("Invalid border_color specified. Borders will not be displayed.")
```
m-col, line 208: Maybe add "bar" somewhere in this warning.

What's the `TypeError` risk here?

I wonder if `rgb` shouldn't raise `ConfigError` instead and we catch that... Or we could add a new error.. Just thinking out loud, ignore this :)

I wonder if we could `rgb` these colours in `__init__` so that it only needs to be done once. Would that work, or maybe introduce potential problems?

libqtile/bar.py
```diff
@@ -289,6 +326,10 @@ def _remove_crashed_widgets(self):
             index = self.widgets.index(i)
             crash = ConfigErrorWidget(widget=i)
             crash._configure(self.qtile, self)
+            if self.horizontal:
+                crash.offsety = self.border_width[0]
+            else:
+                crash.offsetx = self.border_width[3]
```
m-col, line 332: We mentioned this already in another PR - `ConfigErrorWidget` doesn't support vertical bars, does it? Does that same issue affect this? FWIW I don't think that should hold back any PR as the PRs are still improvements, but I'll bump up the priority of getting vertical text boxes working. My last attempt didn't turn out well...

m-col (commented):
Overall this looks good - just one or two minor comments. I'll test-drive it for a day or two to see how it goes.

m-col:
I'm getting strut problems but they appear to have been introduced by e676819a43c371d53e44b7285c36f730c89883f2. Guess my strut test wasn't good enough :crying_cat_face: The way reverved space is handled could likely be improved as the approach prior to that commit was to update the `Gap`'s size if more or less space is needed and then calling `Screen.size()` does `Gap._configure()`. Before, this would update the gap size, reserving space, but that commit made it so that the gap's size always reset to its initial size.

libqtile/bar.py
elParaguayo: I think you're right, `rgb` only raises `ValueError`s.

As for the last point, yes, this check could be done in `__init__`.

libqtile/bar.py
elParaguayo: These lines aren't affected by that issue (it's actually going to fail on the `crash._configure` line before this. We can leave the lines in as they'll then be ready for when we have vertical text boxes!

Which branch is your vertical textbox on? I can take a look at some point if you want.

elParaguayo:
Yeah, that's an issue that needs fixing. Might be better to do that before merging this one.

I'm not an expert with struts so might need your help with that as I'm not clear how they're meant to work. That previous commit tried to stop margins getting applied multiple times but it sounds like you may need that for struts.

m-col:
I think your idea of adding a `struts` attribute to the bar/gap is good for dealing with that issue and https://github.com/qtile/qtile/issues/2639. Then during `Gap._configure`/`Bar._configure`, instead of:
```
self.size = self.initial_size
```
it would be something like:
```
self.size = [sum(side) for side in zip(self.initial_size, self.struts)]
```
As the `self.strut` attribute is added as part of this PR, could you add this change here too?

elParaguayo:
OK. Do you want me to add that here?

Feels like that should be a separate PR as it's a different issue.

m-col:
> Feels like that should be a separate PR as it's a different issue.

Yeah, guess so.

As the issue is independent if this PR, I'm happy to merge this now and then I can fix the struts handling?

elParaguayo:
> > Feels like that should be a separate PR as it's a different issue.
> 
> Yeah, guess so.
> 
> As the issue is independent if this PR, I'm happy to merge this now and then I can fix the struts handling?

No arguments from me!

m-col:
It is done!

elParaguayo:
Yes!

m-col:
Would it be weird if we allowed clicks on the border that is parallel with a bar's orientation to be forwarded to the widget at that position? E.g. a click on (100, 1) on a horizontal bar with a 3 pixel border could pass the click to the widget at x==100, at it's offsety for y. Thoughts? Too weird and hacky?

elParaguayo:
I thought about that. Didn't like it so didn't implement it.

m-col:
Fair enough!

m-col:
I've been getting a lot of these errors since using this:
```
2021-11-14 20:47:20,355 ERROR libqtile loop.py:_handle_exception():L64 'NoneType' object has no attribute 'ink_extents'
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
  File "/home/mcol/git/qtile/libqtile/core/manager.py", line 829, in f
    func(*args)
  File "/home/mcol/git/qtile/libqtile/bar.py", line 546, in _actual_draw
    self.drawer.draw(0, 0)
  File "/home/mcol/git/qtile/libqtile/backend/base.py", line 708, in draw
    self._draw(offsetx, offsety, width, height)
  File "/home/mcol/git/qtile/libqtile/backend/wayland/drawer.py", line 53, in _draw
    if not self.needs_update:
  File "/home/mcol/git/qtile/libqtile/backend/base.py", line 620, in needs_update
    ink_changed = any(not math.isclose(0.0, i) for i in self.surface.ink_extents())
AttributeError: 'NoneType' object has no attribute 'ink_extents'
```

elParaguayo:
I didn't get them but happy if you want to revert for now.

m-col:
Nah it's OK, i'll just try to track it down. The logging seems to be the only symptom - everything is rendering fine.

elParaguayo:
It's a weird error. Sounds like there's a draw call before the surface has been created, otherwise `self.surface` shouldn't be `None`.

PR #2678, opened 2021-08-02
Add test for stock ticker widget

elParaguayo:


elParaguayo:
Looking for some quick wins!

PR #2679, opened 2021-08-03
Update notify widget test

elParaguayo:
There were some recent changes to the `Notify` widget (emitting some signals per the dbus spec).

This commit adds tests for those changes.

@m-col  - I had to be a bit creative to test these but let me know if you see a better way.


libqtile/widget/notify.py
```diff
@@ -130,7 +130,10 @@ def display(self):
         self.bar.draw()
 
     def clear(self, reason=ClosedReason.dismissed):
-        notifier._service.NotificationClosed(self.current_id, reason)
+        notifier._service.NotificationClosed(
+            notifier.notifications[self.current_id].id,
```
elParaguayo, line 134: Changed this as we were sending the widget's internal reference rather than the notification's id.

test/widgets/test_notify.py
```diff
@@ -169,3 +170,147 @@ def test_notifications(manager_nospawn, minimal_conf_noscreen, monkeypatch):
...
+    # notification server. This needs to be created within the manager
+    # object so we rely on "eval" applying "exec".
+    handler = textwrap.dedent(
+        """
+        import asyncio
+
+        from libqtile.utils import add_signal_receiver
+
+        class SignalListener:
+            def __init__(sl):
```
elParaguayo, line 203: Wasn't sure if using `self` here would be problematic given that `self` refers to the qtile instance!

libqtile/widget/notify.py
m-col: Cool, thanks

libqtile/widget/notify.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, width=bar.CALCULATED, **config):
         if self.action:
             default_callbacks['Button3'] = self.invoke
         else:
-            self.capabilities = Notify.capabilities.difference({"action"})
+            self.capabilities = Notify.capabilities.difference({"actions"})
```
m-col, line 76: Woops!

test/widgets/test_notify.py
```diff
@@ -169,3 +170,147 @@ def test_notifications(manager_nospawn, minimal_conf_noscreen, monkeypatch):
...
+                sl.notification_closed = msg.body
+
+        self.signal_listener = SignalListener()
+        """
+    )
+
+    # Create and send a custom notification with a list of actions.
+    # `utils.send_notfication` is not an option here as it does not
+    # expose actions so we need a lower-level call
+    notification_with_actions = textwrap.dedent(
```
m-col, line 237: Does this need to be run in the Qtile process or can it be run from the test process?

test/widgets/test_notify.py
```diff
@@ -169,3 +170,147 @@ def test_notifications(manager_nospawn, minimal_conf_noscreen, monkeypatch):
...
+    shutil.which("notify-send") is None,
+    reason="notify-send not installed."
+    )
+@pytest.mark.usefixtures("dbus")
+def test_invoke_and_clear(manager_nospawn, minimal_conf_noscreen):
+
+    # We need to create an object to listen for signals from the qtile
+    # notification server. This needs to be created within the manager
+    # object so we rely on "eval" applying "exec".
+    handler = textwrap.dedent(
```
m-col, line 196: Same question as below - why does this stuff need to run in the Qtile process, if it's just listening to dbus signals? (Assuming that is what it is doing, perhaps I've got that wrong)

test/widgets/test_notify.py
elParaguayo: That was my first idea. However, to listen for signals, I need the code to run in a loop. I didn't know how to run an asyncio loop in a test so I did it this way!

test/widgets/test_notify.py
elParaguayo: Well it sends a notification too but that  could be done from the test if we can run an asyncio loop in a test. Let me investigate more.

test/widgets/test_notify.py
elParaguayo: If I run an event loop in my test, then I can't use `manager` objects as they can't do `asyncio.run` in an active loop. So, I think the option is either to use manager and my ugly hack, or run the test in a loop and try to execute the widget in that loop.

m-col: Whichever is easiest :)

elParaguayo: Well, the easiest would be not to rewrite my code!

m-col: Lol, then that! Overall LGTM

test/widgets/test_notify.py
m-col: It would work fine as long as you don't need to access the qtile instance, as it would just be masked in this scope

elParaguayo: OK - let me change it as it would be easier to read that way.

Issue #1679, opened 2020-04-26
Plans for documentation


zordsdavini:
As newcomer or even simple user (as me) documentations is not very intuitive in some parts. I'm not good in documenting but will help where I can. So, notes what looks bad:
- [ ] `Get started`: place to store starting pages about Qtile, install, configuration (this should go into deeper), how to use (this part should be big, with images). _All these sections can be as top level in menu_
- [ ] `Tools` - we should place qshel, iqshel, qtile-top etc. under one menu entry
- [ ] `Command API` and `scripting` (this should have more examples) should go into `Haking on Qtile`
- [ ] `Hacking` content is more about contributing and should be moved here
- [ ] `FAQ` is to small. All answered questions can go here

- [ ] qtile.desktop https://github.com/qtile/qtile/issues/1679#issuecomment-619538578

pawamoy:
+1 on this, except for putting "Command API" and "Scripting" inside of "Hacking on Qtile". Hacking on Qtile is like the common CONTRIBUTING or DEVELOPERS file, it's audience is users who wish to fix bugs, add features, or improve the documentation: forks, development setup and pull requests. Command API and Scripting's audience is users (playing with Qtile features).

zordsdavini:
@pawamoy I disagree with you :) Hacking is for playing with Qtile and what you are saying is hacking should go into Contributing, I think

pawamoy:
I was indeed talking about the contents of "Hacking", not the name :stuck_out_tongue: It seems we agree to say that "Hacking" is not the best name for the current contents (which are more about contributing than hacking/playing).

zordsdavini:
sorry, I didn't read it cearfully. Then I agree with you

PeterSR:
> Get started

I agree. I found this project a couple of days ago. As there was no official package for Ubuntu 20.04, I followed the "Installing From Source", but it wasn't clear how to actually install the window manager such that it could be chosen from the login screen. The answer of course was to add the `qtile.desktop` to `/usr/share/xsessions`, but `pip install qtile` does not do this. 

There is a mention of a `qtile-venv.desktop` at the bottom, but it is stated that it is only if one wants to run Qtile in a virtualenv, so initially I ignored that part. 

It might be pretty easy to piece together the puzzle of finding the `qtile.desktop` in the git repo and adding it, but some newcomers might not find it so easy.

PeterSR:
[...]

PR #2682, opened 2021-08-04
Remove quotation marks from docs descriptions

elParaguayo:
There have been several comments about single and double quotation marks in our docs and how it looks a bit untidy (e.g. #1679). This commit removes them.


m-col:
Cool thanks

PR #2685, opened 2021-08-04
Tidy up configuration documentation

elParaguayo:
This adds formatting to variable names and default values so as to be more consistent with the style used for widgets, layouts etc.


m-col:
Cool, thanks!

Issue #2657, opened 2021-07-27
Moving the cursor to another screen doesn't always cause that screen to focus


ParaSait:
# Issue description
When I move the mouse cursor to another screen, the behavior that I expect is that that screen always becomes focused. However, the screen will now only become focused when you move it onto a window that's on it, not when you move it onto a blank area or a bar. Clicking on the blank area or bar doesn't trigger the focus either.
(By "focused", I mean the screen that any commands apply to, such as spawning a new window.)

# Qtile version
v0.18.0 (in v0.17.0, this was still working)

**Correction:** actually, in 0.17.0, it seems only clicking will focus it, not just moving the mouse over...

# Configuration
I've reproduced it with my own configuration and with the default fallback configuration, so I don't believe it's related to it.

yobleck:
Looks similar to issue #2098
Personally I'd be fine with a qtile constantly watching the root window. Could be a config option with a warning about increased cpu usage

ParaSait:
I'd just like to see the v0.17.0 functionality repaired, i.e., focus on the window & its screen when hovering the mouse over a window, and focus on the screen when clicking on a blank space on the screen. I got it wrong when I implied that just hovering a blank space would change the focus in v0.17.0, but the fact remains that clicking blank space doesn't do anything in v0.18.0, which I believe must be a bug.
Of course, splitting the `follow_mouse_focus` option into two separate options (or something to that effect) would be nice to have.

m-col:
> Personally I'd be fine with a qtile constantly watching the root window. Could be a config option with a warning about increased cpu usage

I'd be in favour of this too, as it is consistent with `follow_mouse_focus` but with screens. I don't think it would affect CPU usage noticably and would be pretty easy to add in if anybody wants to look into it.

jas3333:
I'm having the same issue. Windows kept opening on my main monitor even after clicking on the second screen. Was driving me crazy so I went back to version 0.17.

yobleck:
It's taking me awhile to look into this but there's a temporary work around with the CurrentScreen widget.
put:
`widget.CurrentScreen(mouse_callbacks={"Button1": lambda: qtile.focus_screen(0)}),`
in the bar on every screen (replace `0` with the screen number). This requires you to move the mouse to the bar as you switch screens and click the widget, which is a hassle but serviceable

yobleck:
Another janky workaround:
Add:
```
try:
    if self.qtile.config.follow_mouse_focus:
[...]

PR #2687, opened 2021-08-05
Add ButtonPress event mask to root window

yobleck:
See issue: #2657
Add button press event mask so that clicking on the root window will change screen focus.

Code for screen change was already there, it just wan't activating without a event listener:
https://github.com/qtile/qtile/blob/7537ab744324ec3ac2d2c3493e7d767e4a3d8de9/libqtile/backend/x11/core.py#L763-L767

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -113,6 +113,7 @@ class Core(base.Core):
                 | EventMask.SubstructureRedirect
                 | EventMask.EnterWindow
                 | EventMask.LeaveWindow
+                | EventMask.ButtonPress
             )
         )
 
--- a/test/backend/x11/test_window.py
+++ b/test/backend/x11/test_window.py
@@ -11,6 +11,7 @@ from libqtile import layout, utils
 from libqtile.backend.x11 import window, xcbq
 from test.conftest import dualmonitor
 from test.helpers import (
+    HEIGHT,
     SECOND_HEIGHT,
     SECOND_WIDTH,
     WIDTH,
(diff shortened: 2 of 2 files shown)


m-col:
Any chance you could add a test? Perhaps just warping the pointer to the second output and running doing a fake click, and checking the screen changes. grep the tests for "fake_click", it'll be useful. This pertains specifically to the x11 backend too, so it'd go inside that test folder. Thanks for debugging this!

m-col:
This kinda idea (untested):
```py
@dualmonitor
def test_click_focus_screen(xmanager):
    screen1 = (WIDTH // 2, HEIGHT // 2)
    screen2 = (SECOND_WIDTH // 2, SECOND_HEIGHT // 2)
    xmanager.c.eval(f"self.core.warp_pointer{screen1}")
    assert xmanager.c.screen.info()["index"] == 0

    # Warping alone shouldn't change the current screen
    xmanager.c.eval(f"self.core.warp_pointer{screen2}")
    assert xmanager.c.screen.info()["index"] == 0
    # Clicking should
    xmanager.backend.fake_click(*screen2)
    assert xmanager.c.screen.info()["index"] == 1

    xmanager.c.eval(f"self.core.warp_pointer{screen1}")
    assert xmanager.c.screen.info()["index"] == 1
    xmanager.backend.fake_click(*screen1)
    assert xmanager.c.screen.info()["index"] == 0
```

yobleck:
*facepalm
Here I go again. Git won't let me squash.
When I run `git rebase -i HEAD~2` it instantly deletes the git-rebase-todo file and says:
 `Successfully rebased and updated refs/heads/screen_change.` but then nothing happens. IDK whats going on :P

m-col:
Fantastic, thanks for investigating this!

PR #2688, opened 2021-08-06
Extend Object Graph

elParaguayo:
This commit allows you to select widgets from the parent `Bar` or `Screen` object i.e. show widgets in a selected `Bar` or on a selected `Screen` (which could contain multiple `Bar`s).

It never made sense to me that you couldn't access the widgets from their parents. 

I also wasted a good 45 mins re-drawing the object graph diagram for the docs and then saw that it's generated automatically!

I should add a test for this...

### DIFF
--- a/docs/_static/diagrams/object-graph-orig.dot
+++ b/docs/_static/diagrams/object-graph-orig.dot
@@ -32,6 +32,7 @@ digraph G {
     root -> window;
 
     bar -> screen;
+    bar -> widget;
 
     group -> layout;
     group -> screen;
@@ -44,6 +45,7 @@ digraph G {
     screen -> bar;
     screen -> layout;
     screen -> window;
+    screen -> widget;
 
     widget -> bar;
     widget -> group;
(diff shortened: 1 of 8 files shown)


ramnes (approved):
LGTM

elParaguayo:
One more change - have tried to simplify the diagram in the docs using bi-directional arrows where appropriate.

flacjacket (approved):
Nice!

elParaguayo:
Widget's graph was wrong (it shouldn't be able to access `group` - that makes no sense). Also, despite the graph showing it could access `screen` the code didn't allow it.

elParaguayo:
Should we add the ability for a `screen` to access its current `group`? Seems reasonable to me.

elParaguayo:
In case anyone was wondering about these lines:
https://github.com/qtile/qtile/blob/f86170fb192371328652b34eba5c8f36ab9a656e/libqtile/command/client.py#L272-L278

These are needed because, if you try to do something like `qtile cmd-obj -o screen group`, you get a `ValueError` because the code is trying to run `int("group")`. `cmd-obj` doesn't catch this so it fails.

PR #2690, opened 2021-08-07
Widget to show camera or/and mic active status in statusbar

desdemona2:
There are many big companies which have done this. I noticed it in firefox a long time ago, then iOS14 and now Android 12 all have implemented this. This is my try for linux system.


libqtile/widget/__init__.py
```diff
@@ -27,6 +27,7 @@
     "AGroupBox": "groupbox",
     "Backlight": "backlight",
     "Battery": "battery",
+    "AccessDots": "accessDots",
```
elParaguayo, line 30: The first string is the name of the widget class so needs to match what you call with the widget.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# modification suggested by elParaguayo (https://github.com/elParaguayo)
```
elParaguayo, line 20: Don't need this line.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+# This feature is also included in iOS 14 and Android 12 and inspired from an app AccessDots for Android.
```
elParaguayo, line 21: Any description of the widget should be in its docstring.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+
+from libqtile.widget import base
+import os
+
+
+class active(base.ThreadPoolText):
```
elParaguayo, line 27: Needs a better name and should be capitalised. Personally, I don't like AccessDots as that's just copying the Android app name and it gives the user no idea what it does.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+    """widget to show indication if
```
elParaguayo, line 28: Can you give a better description of what the widget does and capitalise, punctuate etc.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+    camera and mic are active in status bar"""
+
+    defaults = [
+        (
+            "update_interval",
+            1,
+            "Update interval in seconds, if none, the "
+            "widget updates whenever it's done'.",
+        ),
+        # ("fmt", "{mic_str} {cam_str}", "Display format for output"),
```
elParaguayo, line 38: I have a strong preference to include this so, if there's a problem with it, let's fix it.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+        ("camDev", "/dev/video0", "Path to camera device"),
```
elParaguayo, line 39: No camelcase please (yes, my username is a bad example!)

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+        ("micDev", "/dev/snd/pcmC0D0c", "Path to Microphone device"),
+        ("camAct", "📸", "Indication when camera active"),
+        ("micAct", "📢", "Indication when Microphone active"),
```
elParaguayo, line 42: We should add option for inactive string as some users may want a visual indicator when the device is inactive, not just when it's active.

libqtile/widget/accessDots.py
elParaguayo: I would also add a warning that this will only tell you the status of the device when the widget polls the device (default once a second). It does not guarantee that your device is not being accessed in between updates.

libqtile/widget/accessDots.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, **config):
+        super().__init__("", **config)
+        self.add_defaults(active.defaults)
+
+    def poll(self):
+
+        mic = os.system(f"fuser {self.micDev}")
+        camera = os.system(f"fuser {self.camDev}")
+
+        mic_str = "" if mic == 256 else self.micAct
```
elParaguayo, line 54: As above, inactive text should not be hardcoded.

libqtile/widget/accessDots.py
desdemona2: Can u please suggest a good name. I am very bad when it comes to naming.

ramnes:
Let's make a base class and two different subclasses? I think this idea could be applicable to a lot of other hardware: bluetooth, wifi, etc. Also you probably want different mouse_callbacks based on the type of hardware.

libqtile/widget/accessDots.py
LamprosPitsillos: **IOaccess** sounds good!

desdemona2:
@ramnes i tried creating a base widget indicator and inherited MicIndicator and CamIndicator from it and now i am having Metaclass conflict and i have no idea about this.
I was also thinking of creating base class as abstract because i don't see any use of it elsewhere but i am not sure about doing this.

PR #2691, opened 2021-08-08
Add keybinding images generator for FR-BE keyboards

gaelj:
no_modifier


elParaguayo:
How much of this is duplicated from `gen-keybinding-img`

It feels like we should really have a single script with the main methods in it and the ability to specify a keyboard layout as an option.

gaelj:
There are 2 kinds of modifications : key locations and key geometries. I will look into it

gaelj:
Layout country is now a script argument, with only 1 script handling default en-US and the new fr-BE

gaelj:
@elParaguayo what do you think about the latest changes ? Anything more to improve ?

elParaguayo:
Sorry - been looking at lots of different things, will try to get to this one soon. Will take a bit more time as it's a part of the code that I'm less familiar with.

m-col:
Sorry for letting this get stale, I've re-opened it!

m-col:
@zordsdavini would you happy to take a look at this?

PR #2699, opened 2021-08-14
Update multi-screen configuration docs

elParaguayo:
There was a question on chat about how to configure multiple screens and the docs are a bit quiet on this so I've tried to add something short and sweet.

The PR is in draft for now as I'm not 100% sure how screen order is set/resolved on Wayland so need some help on that bit (@m-col - when you've got a minute. Not urgent!)

### DIFF
--- a/docs/manual/config/screens.rst
+++ b/docs/manual/config/screens.rst
@@ -3,9 +3,8 @@ Screens
 =======
 
 The ``screens`` configuration variable is where the physical screens, their
-associated ``bars``, and the ``widgets`` contained within the bars are defined.
-
-See :ref:`ref-widgets` for a listing of available widgets.
+associated ``bars``, and the ``widgets`` contained within the bars are defined
+(see :ref:`ref-widgets` for a listing of available widgets). 
 
 Example
 =======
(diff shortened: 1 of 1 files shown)


docs/manual/config/screens.rst
```diff
@@ -49,6 +48,25 @@ entire window.
...
+
+You can view the current order of your screens by running ``xrandr --listmonitors``.
+
+For examples of how to set the order of your screens please refer to the
+`Arch wiki <https://wiki.archlinux.org/title/Multihead>`_.
+
+Wayland
+~~~~~~~
+
+
```
m-col, line 69:
```suggestion
The Wayland backend supports the wlr-output-management protocol for configuration of outputs by tools such as `Kanshi <https://github.com/emersion/kanshi>`_.
```

m-col (approved):
Nice, good idea. I've proposed an entry for the Wayland section. LGTM

docs/manual/config/screens.rst
elParaguayo: Thanks.

ramnes (approved):

PR #2701, opened 2021-08-15
For tasklist widget added skiptaskbar filter and options to mimic win…

yobleck:
…downame/tabs widgets(to be deprecated)

The primary purpose of this PR is to add a filter to tasklist that checks for windows with the `_NET_WM_STATE_SKIP_TASKBAR`
WM property and hides them from the taskbar. [see this google groups thread](https://groups.google.com/g/qtile-dev/c/NW06IxHBDKc)

However I didn't see the point of continuously add these sorts of features to all three widgets (tasklist, windowname, windowtabs) and having to maintain them, so I added options to tasklist to mimic the behavior of the other two widgets so that they can be deprecated (with qtile migrate) ~~in a future update. I figured I wouldn't remove them right away to give people time to test the new options for bugs/suggest new ideas etc~~.

new tasklist options are current_win_only
and mouse_callbacks now excepts `None` to disable all callbacks including builtins

also fixed a bare except


yobleck:
Am I not doing deprecation warnings correctly? tests pass locally for me (with warnings)

libqtile/widget/tasklist.py
```diff
@@ -238,7 +254,12 @@ def get_taskname(self, window):
 
     @property
     def windows(self):
-        return self.bar.screen.group.windows
+        if self.current_win_only:
+            wins = self.bar.screen.group.current_window
+            wins = set([wins]) if wins else set([])  # Sets are necessary here to ensure consistent formatting
+        else:
+            wins = self.bar.screen.group.windows.copy()  # set2 = set1 uses copy by reference
```
elParaguayo, line 261: Why does this need to be copied?

libqtile/widget/tasklist.py
```diff
@@ -327,6 +348,14 @@ def update(self, window=None):
         if not window or window in self.windows:
             self.bar.draw()
 
+    def skip_taskbar(self, windows):
+        """
+        Check every window's _NET_WM_STATE property for _NET_WM_STATE_SKIP_TASKBAR
+        and remove that window from the list to be shown on the taskbar.
+        This is mostly for GTK programs. Qt doesn't seem to support this
+        """
+        return {w for w in windows if "_NET_WM_STATE_MAXIMIZED_VERT" not in w.window.get_net_wm_state()}
```
elParaguayo, line 357: I thought we were filtering out windows that had `_NET_WM_STATE_SKIP_TASKBAR`. Why are you checking for `_NET_WM_STATE_MAXIMIZED_VERT`?

yobleck: lol my bad. I dont have any programs that use skip task bar so I used the other one for testing and forgot.

libqtile/widget/tasklist.py
yobleck: The original grabs the set by reference so when tasklist.skip_taskbar edits the set it edits the object in memory that all of Qtile uses if we don't copy it. This causes the window to enter a half managed half not state if it has the skiptaskbar property.

elParaguayo:
> Am I not doing deprecation warnings correctly? tests pass locally for me (with warnings)

CI runs pytest with the `-W error` arguments which, I think, means warnings are treated as errors.

You would normally tell pytest to ignore warnings by using [@pytest.mark.filterwarnings](https://docs.pytest.org/en/latest/how-to/capture-warnings.html#pytest-mark-filterwarnings). However, the problem here is that the warning comes as a result of an import so it may need a different approach.

libqtile/widget/tasklist.py
elParaguayo: No problem.

Would be good if you could check it with a program that does set that property. I think, in the google group chain, someone mentioned a terminal that lets you set it.

libqtile/widget/tasklist.py
elParaguayo: Interesting. I assumed `skip_taskbar` was creating a new set rather than editing the existing one as it's not changing `windows` but making a new set from items in `windows`.

elParaguayo:
I'd like to think about the deprecation of those two widgets a bit more and see if there's a nice way to handle it (e.g. can we use those old widget names as a wrapper to return the modified TaskList?)

Would also be interesting to if other members of the team have any views.

libqtile/widget/tasklist.py
yobleck: confirmed working by running 
`xprop -id "$windowid" -f _NET_WM_STATE 32a -set _NET_WM_STATE _NET_WM_STATE_SKIP_TASKBAR `

libqtile/widget/tasklist.py
yobleck: You're right. I forgot about that when I switched to set comprehension from an earlier clunkier version that modified the set in place.

ramnes (approved):
LGTM

libqtile/widget/windowtabs.py
```diff
@@ -48,6 +50,7 @@ class WindowTabs(base._TextBox):
     ]
 
     def __init__(self, **config):
+        warnings.warn("Use tasklist(interactive=False) instead", DeprecationWarning)
```
ramnes, line 53: Why not just 

```suggestion
        warnings.warn("Use tasklist(mouse_callbacks=None) instead", DeprecationWarning)
```

and not add an extra option for this?

ramnes:
Also, just as in #2497, I'd say we should add the migration script and remove the old widgets directly, that's why `qtile migrate` exists. But either works.

libqtile/widget/windowtabs.py
yobleck: That does make more sense. I had to add a check to the button_press function so it doesn't run and throw errors when there are no callbacks. I also added a tasklist docstring for mouse_callbacks that overrides the original and explains setting the value to none.

ramnes: Uh I thought None was already supported for all widgets. If not we should probably do this instead of here only.

ramnes: Also I don't understand why we need the extra check and that seems weird... but I need to sleep right now. :)

yobleck: have a good night.
When I set `mouse_callbacks=None` I get this error
```
  File "/home/yobleck/qtile_dev/update_tasklist/qtile/libqtile/backend/x11/core.py", line 298, in _xpoll
    ret = target(event)
  File "/home/yobleck/qtile_dev/update_tasklist/qtile/libqtile/backend/x11/window.py", line 1070, in handle_ButtonPress
    self.process_button_click(e.event_x, e.event_y, e.detail)
  File "/home/yobleck/qtile_dev/update_tasklist/qtile/libqtile/bar.py", line 362, in process_button_click
    widget.button_press(
  File "/home/yobleck/qtile_dev/update_tasklist/qtile/libqtile/widget/tasklist.py", line 430, in button_press
    base._Widget.button_press(self, x, y, button)
  File "/home/yobleck/qtile_dev/update_tasklist/qtile/libqtile/widget/base.py", line 236, in button_press
    if name in self.mouse_callbacks:
TypeError: argument of type 'NoneType' is not iterable
```

So it seems that `None` isn't supported for the mouse_callback argument?
perhaps changing :https://github.com/qtile/qtile/blob/e928649da5919f30cfa4db9348f51e7d2e851e7b/libqtile/widget/base.py#L236
to
`if self.mouse_callbacks and name in self.mouse_callbacks:`
would be a solution (it fixes the error but IDK if it's the right way to do it).

ramnes: Yeah, that would work I think!

libqtile/widget/base.py
```diff
@@ -116,7 +116,8 @@ def open_calendar():
     offsety: int = 0
     defaults = [
         ("background", None, "Widget background color"),
-        ("mouse_callbacks", {}, "Dict of mouse button press callback functions."),
+        ("mouse_callbacks", {},
+         "Dict of mouse button press callback functions. Set to `None` to disable"),
```
ramnes, line 120: We probably want to modify `add_callbacks` to reflect that for the current widgets? i.e. not adding callbacks if mouse_callbacks is None.

libqtile/widget/tasklist.py
```diff
@@ -167,8 +172,11 @@ def __init__(self, **config):
         self.clicked = None
         if self.spacing is None:
             self.spacing = self.margin_x
+        if self.current_win_only:
+            self.is_interactive = False  # Minimized windows are hard to get back
 
-        self.add_callbacks({'Button1': self.select_window})
+        if self.mouse_callbacks is not None:
```
ramnes, line 178: ... then we wouldn't need this. (related to previous comment)

libqtile/widget/base.py
```diff
@@ -233,7 +234,7 @@ def add_callbacks(self, defaults):
 
     def button_press(self, x, y, button):
         name = 'Button{0}'.format(button)
-        if name in self.mouse_callbacks:
+        if self.mouse_callbacks and name in self.mouse_callbacks:
```
ramnes, line 237: nor this, actually (continuation of previous comment)

yobleck: As in putting:

`if self.mouse_callbacks is not None:`

In the add_callbacks func of the base widget so none of its code runs?

yobleck: I think we still need this one because `name in self.mouse_callbacks:`
throws `TypeError: argument of type 'NoneType' is not iterable`.
without it.

ramnes: Yeah right, this one is still needed. For my defense it was like 4 AM when I wrote the comment, sorry! :D

yobleck: no prob :)

yobleck:
I'm working on migrating the old widgets to the tasklist but I'm running into some issues.

<details>
  <summary>code in scripts/migrate.py</summary>

```
def windowname_to_tasklist(query):
    query.select_attribute("WindowName").rename("TaskList")
    query.select_method("TaskList").modify_argument("markup", "markup_normal")
    query.select_method("TaskList").remove_argument("empty_group_string")
    query.select_method("TaskList").remove_argument("fmt")
    query.select_method("TaskList").remove_argument("for_current_screen")
    query.select_method("TaskList").remove_argument("format")
    return query

def windowtabs_to_tasklist(query):
    query.select_attribute("WindowTabs").rename("TaskList")
    query.select_method("TaskList").modify_argument("markup", "markup_normal")
    query.select_method("TaskList").remove_argument("selected")
    query.select_method("TaskList").remove_argument("fmt")
    query.select_method("TaskList").remove_argument("separator")
    return query
```

</details>

modify_argument seems to work but remove_argument throws 
`ValueError: remove_argument requires passing original function`
and won't seem to work unless the source file is imported [see issue](https://github.com/facebookincubator/Bowler/issues/84)

Also I when changing the widget name from windowname to tasklist I had to use `select_attribute` instead of `select_class`
Which all the other widget name change examples use. Am I doing something wrong there?

ramnes:
Don't we want to keep the arguments anyway, and replace them by an equivalent option rather than removing them?

> Also I when changing the widget name from windowname to tasklist I had to use `select_attribute` instead of `select_class`
> Which all the other widget name change examples use. Am I doing something wrong there?

Why did you "have to"?

yobleck:
Is it worth adding some arguments to tasklist if it already handles them? Like the separator argument for windowtabs which is already handled by tasklists borders.

When I used set_class nothing happened. It returned config unchanged. Not sure why.

ramnes:
> Is it worth adding some arguments to tasklist if it already handles them? Like the separator argument for windowtabs which is already handled by tasklists borders.

Depends on the argument I'd say. For `separator` yeah, I'd say we can drop it. Let's see what others think.

yobleck:
Looking at test_migrate it looks like the reason it does select_class instead of select_attribute is because the test is doing
from `libqtile.widget import class` and then adding `class`to the bar

Whereas I'm doing `from libqtile import widget` then adding `widget.class`.

Not sure which is preferable.

elParaguayo:
> Looking at test_migrate it looks like the reason it does select_class instead of select_attribute is because the test is doing
> from `libqtile.widget import class` and then adding `class`to the bar
> 
> Whereas I'm doing `from libqtile import widget` then adding `widget.class`.
> 
> Not sure which is preferable.

That's interesting. I imagine people's configs could use either approach so really migrate should cover both. I might look at that as a separate issue.

yobleck:
> I can pick this up.

This pr is getting a bit cluttered so any help is appreciated. :)

yobleck:
windowname/tabs get their padding arg from _TextBox but tasklist gets it from PaddingMixin which doesn't allow for none types.
PaddingMixin didn't seem to have a place to handle this so I put it in tasklist.

yobleck:
replaced WindowName with TaskList in default_config.

Added migration for windowname and tabs to tasklist.
I decided to just ignore old options that don't do anything for tasklist 
because python isn't bothered by useless keyword args.

also added test.

yobleck:
seems bowler's add_argument doesn't work either


<details>
<summary>preliminary custom version</summary>
<br>

```
# This nasty workaround handles appending new arguments to a widget
# because bowler's add_argument() doesn't work.
# All the .children calls are traveling through a fissix parse tree
# and the appropriate leaves and nodes are being added.
# Assumes bar.Bar([widget.WidgetName()]) format in config
#TODO: new modify_argument and remove_argument
try: #TODO: add check to do_migrate()
    from fissix.pytree import Node, Leaf
except ImportError:
    pass
def add_argument(node, capture, filename):

    arg_dict = {
        "WindowName": [" current_win_only", "True"],
        "WindowTabs": [" mouse_callbacks", "None"],
        }
    new_arg, new_value = arg_dict[node.children[1].children[1].value]  # widget name

    temp = capture.get("function_arguments")  # check if widget has args
    if temp:
        if node.children[-1].children[-2].children[-1] != Leaf(12, ','):
            node.children[-1].insert_child(-1, Leaf(12, ','))  # insert trailing comma

        n = Node(261, [Leaf(1, new_arg), Leaf(22, '='), Leaf(1, new_value), Leaf(12, ',')])
        node.children[-1].insert_child(-1, n)

    elif temp == []:  # no args
        n = Node(261, [Leaf(1, new_arg), Leaf(22, '='), Leaf(1, new_value), Leaf(12, ',')])
        node.children[-1].insert_child(1, n)
```

</details>

yobleck:
Quick git question. How do I squash my commits after merging other peoples PR's?
I ran `git reset --soft HEAD~11` with the intention of recommitting all my changes at once and force pushing.
But I noticed some of the recent docs changes got caught up in the reset.
I reset the reset with `git reset HEAD@{1}` but I'm not sure what to do next (git rebase tends to screw thing up for me).

ramnes:
`git rebase -i master` and then you follow the instructions written in comments

PR #2705, opened 2021-08-18
Give Notify a parse_text option like other widgets

ramnes:


elParaguayo:
Good idea. LGTM.

Any chance of a test ;)

elParaguayo:
> Good idea. LGTM.
> 
> Any chance of a test ;)

Joking aside, let me have a quick test of this PR tonight. If it's ok, I'll merge it and then update my open PR #2679 to add an extra test for this.

ramnes:
No need to joke, I definitely should have added a test in the first place! But if you want to do it in #2679 I wouldn't say no either. :)

libqtile/widget/notify.py
```diff
@@ -57,6 +58,13 @@ class Notify(base._TextBox):
         ),
         ("audiofile", None, "Audiofile played during notifications"),
         ("action", True, "Enable handling of default action upon right click"),
+        (
+            "parse_text", None, "Function to parse and modify notifications. "
+            "e.g. function in config that removes line returns:"
+            "def my_func(text)"
```
elParaguayo, line 64: Two things:
1) Missing a `:` after the definition
2) This formatting won't be preserved in our docs - it will just look like:

> Function to parse and modify notifications. e.g. function in config that removes line returns:def my_func(text) return text.replace(’n’, ‘’)then set option parse_text=my_func

elParaguayo: Other than that, code runs fine so I can merge once this is fixed. I'll do the test for it.

ramnes: I just copy pasted it from another widget so I'd say let's stay consistent for now and fix both in a later PR. :)

elParaguayo: Fine. I'll add to my list!

PR #2707, opened 2021-08-20
Fix #2706: Migrate script doesn't convert BitcoinTicker

### LINKED ISSUE #2706: BitcoinTicker migration does not work
```
/tmp/crypto cat config.py 
from libqtile.widget import BitcoinTicker

test = BitcoinTicker()
/tmp/crypto qtile migrate -c config.py 
Config unchanged.
/tmp/crypto cat config.py 
from libqtile.widget import BitcoinTicker

test = BitcoinTicker()
```

/cc @Graeme22

Graeme22:
This is what I get for not actually running the migrate script!

### DIFF
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -129,6 +129,7 @@ MIGRATIONS = [
     tile_master_windows_rename,
     threaded_poll_text_rename,
     pacman_to_checkupdates,
+    bitcoin_to_crypto,
     hook_main_function,
     new_at_current_to_new_client_position,
 ]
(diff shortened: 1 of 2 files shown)


elParaguayo:
Can you add a test to `test_migrate.py`?

elParaguayo:
Also, can you rename the commit as something more descriptive so it's clear what it does without referring to an issue number.

elParaguayo:
Thanks. I was hoping we could change the title of the commit itself, rather than the PR title as it's the commit that will show in the history. Just something like "Fix BitcoinTicker migration".

PR #2709, opened 2021-08-21
Clean log generated by cmus widget.

Fabien-Bernard:
Since 5f7e1436, the cmus widget pollutes my log when cmus is not running.
This PR adapts the widget and its test to the subprocess.CalledProcessError.output type induced by this commit.

Snippet of log generated :
```
2021-08-21 17:26:00,935 ERROR libqtile base.py:on_done():L553 poll() raised exceptions, not rescheduling
  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/cmus.py", line 58, in get_info
    output = self.call_process(['cmus-remote', '-C', 'status'])
  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/base.py", line 301, in call_process
    return subprocess.check_output(command, **kwargs, encoding="utf-8")
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['cmus-remote', '-C', 'status']' returned non-zero exit status 1.

  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/base.py", line 550, in on_done
    result = future.result()
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/cmus.py", line 116, in poll
    return self.now_playing()
  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/cmus.py", line 83, in now_playing
    info = self.get_info()
  File "/home/bernard/.local/lib/python3.9/site-packages/qtile-0.18.1.dev92+g6f5581eb-py3.9-linux-x86_64.egg/libqtile/widget/cmus.py", line 60, in get_info
    output = err.output.decode()
AttributeError: 'str' object has no attribute 'decode'
2021-08-21 17:26:00,936 WARNING libqtile base.py:on_done():L567 poll() returned None, not rescheduling
```

### DIFF
--- a/libqtile/widget/caps_num_lock_indicator.py
+++ b/libqtile/widget/caps_num_lock_indicator.py
@@ -40,7 +40,7 @@ class CapsNumLockIndicator(base.ThreadPoolText):
         try:
             output = self.call_process(['xset', 'q'])
         except subprocess.CalledProcessError as err:
-            output = err.output.decode()
+            output = err.output
         if output.startswith("Keyboard"):
             indicators = re.findall(r"(Caps|Num)\s+Lock:\s*(\w*)", output)
             return indicators
--- a/libqtile/widget/caps_num_lock_indicator.py
+++ b/libqtile/widget/caps_num_lock_indicator.py
@@ -41,6 +41,7 @@ class CapsNumLockIndicator(base.ThreadPoolText):
             output = self.call_process(['xset', 'q'])
         except subprocess.CalledProcessError as err:
             output = err.output
+            return []
         if output.startswith("Keyboard"):
             indicators = re.findall(r"(Caps|Num)\s+Lock:\s*(\w*)", output)
             return indicators
--- a/libqtile/widget/cmus.py
+++ b/libqtile/widget/cmus.py
@@ -57,7 +57,7 @@ class Cmus(base.ThreadPoolText):
         try:
             output = self.call_process(['cmus-remote', '-C', 'status'])
         except subprocess.CalledProcessError as err:
-            output = err.output.decode()
+            output = err.output
         if output.startswith("status"):
             output = output.splitlines()
             info = {'status': "",
--- a/libqtile/widget/moc.py
+++ b/libqtile/widget/moc.py
@@ -57,7 +57,7 @@ class Moc(base.ThreadPoolText):
         try:
             output = self.call_process(['mocp', '-i'])
         except subprocess.CalledProcessError as err:
-            output = err.output.decode()
+            output = err.output
         if output.startswith("State"):
             output = output.splitlines()
             info = {'State': "",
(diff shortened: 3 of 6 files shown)


elParaguayo:
Nice catch. Fix looks good to me (but why didn't the test fail?)

I think there are a few other widgets that have a similar set up (catching `CalledProcessError` and then `decode`ing the result). Can you fix those in the same PR.

Fabien-Bernard:
The widget don't crash and we are only checking the value of the text variable, so it goes silently under the carpet i think.

I found 2 other widgets with the same set up : Moc and CapsNumLockIndicator. Let me know if I missed one.

elParaguayo:
Thanks. Agreed, think those are the only other two.

The test was passing because I mocked `call_process` and so the change in that commit didn't impact the test. However, the test would fail after your change to the widget as the widget now expects a string but the test was providing bytes. Thanks for fixing that too.

Fabien-Bernard:
I added a basic test file for Caps_Num_Lock_Indicator widget, I was out of imagination to test the exception manually.
I tried to stay close to the cmus test file. More tests could be added later, but for now, the exception is tested.
It makes me update the widget again, tests are usefull ;)

elParaguayo:
Thanks. This all looks really good.

One question: we're not doing anything with the errors at the moment. Should we be logging them/alerting the user? Need to think how best to do that if these widgets are polling frequently. However, that's something we can fix in a separate commit.

Fabien-Bernard:
Imho :
As a cmus widget user, I'm fine with the current situation. I don't need a warning to tell me that cmus is not running, which will be the cause of the exception most of the time. Same for moc widget.
It's different for the CapsNumLockIndicator widget. If an exception occurs, it  means surely that something is wrong with the xset command. An error should be logged and maybe even a message displayed in the bar by the widget instead of the empty string.
In this case, the poll method of the widget could return `None` to simply avoid rescheduling.

elParaguayo:
Makes sense to me. I may take a look at that later.

PR #2714, opened 2021-08-25
Fix nvidia widget temperature color not changing. Add test

yobleck:
The NvidiaSensors changes to an alert color (red by default) when the GPU goes over the set threshold temperature.
However, when the temperature drops back down below this threshold the color does not reset to normal (white by default).
This PR adds a fix for that, as well as a test based off the moc widget color test.

I'm a bit worried that the logic for my test is circular (as in `assert 1==1`) and doesn't actually do anything.
Any advice on this would be appreciated.


elParaguayo:
Sorry - I appear to have commented on the PR on your fork rather than on here! Let me re-post them.

test/widgets/test_nvidia_sensors.py
```diff
@@ -12,3 +15,44 @@ def test_nvidia_sensors_input_regex():
     assert correct_sensors == {'temp', 'fan_speed', 'perf'}
     assert _all_sensors_names_correct(correct_sensors)
     assert not _all_sensors_names_correct(incorrect_sensors)
+
+
+def no_op(*args, **kwargs):
+    print("no_op")
```
elParaguayo, line 21: You can delete this line. Appreciate you copied it from one of the tests I wrote but it's not needed!

test/widgets/test_nvidia_sensors.py
```diff
@@ -12,3 +15,44 @@ def test_nvidia_sensors_input_regex():
...
+    return (n, request.param)
+
+
+def test_nvidia_sensors_foreground_colour(fake_nvidia):
+    fake_nvidia[0].poll()
+
+    if fake_nvidia[1] == return_below_threshold:
+        assert fake_nvidia[0].layout.colour == fake_nvidia[0].foreground_normal
+    elif fake_nvidia[1] == return_above_threshold:
+        assert fake_nvidia[0].layout.colour == fake_nvidia[0].foreground_alert
```
elParaguayo, line 58: This section isn't really testing the issue that your modified code is fixing i.e. you changed the code to check that the colour reverts to normal if the temperature drops back to normal having been in the alert range.

However, your test is just creating a new widget and checking the colour once, depending on the current temperature.

elParaguayo (changes requested):

PR #2718, opened 2021-08-25
WL: let wlroots internal signals destroy resources

m-col:
Many wlroots managers used by the wayland `Core` respond to destruction
of the display so let's let them handle destruction themselves. This
results in a cleaner shutdown.

Seems we're still getting a segfault right at the end of shutdown though.


m-col:
Gonna close this for now because I need to re-write it and fully debug it and it's low on my prorities.

PR #2719, opened 2021-08-26
Add test for restoration of state after restart

elParaguayo:
As per #2348 we need a test for `restart`. This test:
 - checks restored state (groups and layouts) matches previous state
 - confirms that state files are the same after restarting

This is a first go at this issue. Happy to take suggestions for additional tests.

This test works locally for me but don't know how it will behave on CI.

### DIFF
--- a/test/test_restart.py
+++ b/test/test_restart.py
@@ -1,0 +1,190 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import pickle
+import shutil
+import textwrap
+from multiprocessing import Value
+
+import pytest
+
+import libqtile.bar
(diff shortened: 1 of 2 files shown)


elParaguayo:
Following on from my last post, if you just run a short test like this:
``` python
def test_os_exec():
    os.execv(sys.executable, sys.argv)
    assert True
```
The test does not complete and pytest just seems to exit.

I think that must mean that, in this test, the `_atexit` method in `LifeCycle` is not triggered as, otherwise, this line would cause problems:
https://github.com/qtile/qtile/blob/673b40d945dd166b37903ad3032fd72c9aea7db3/libqtile/core/lifecycle.py#L36

m-col:
> I think that must mean that, in this test, the _atexit method in LifeCycle is not triggered

https://stackoverflow.com/questions/34506638/how-to-register-atexit-function-in-pythons-multiprocessing-subprocess

elParaguayo:
OK. So we are testing that method in this test. Given that it really just recreates the command string, that's probably ok to skip.

Is there anything else connected to restarting that you want to test?

test/test_restart.py
```diff
@@ -0,0 +1,159 @@
...
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar([TextBox("Qtile Test")], 10),
+            x=0, y=0, width=400, height=600
+        ),
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar([TextBox("Qtile Test")], 10),
+            x=400, y=0, width=400, height=600
+        )
+    ]
+    screens = []
```
m-col, line 60: What parts of this are needed for the tests in this module, and which are there because they have to? Perhaps it'd be neater to import `BareConfig` from `helpers` and just explicitly have those defined here that are used here?

test/test_restart.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def test_qtile_state(manager_nospawn, request, backend, backend_name):
+
+    if backend_name == "wayland":
+        pytest.skip("Skipping test on Wayland.")
+
+    # This injection allows us to capture the lifecycle state filepath before
+    # restarting Qtile
+    inject = textwrap.dedent(
```
m-col, line 70: So we're delaying `Qtile._stop`, getting the file path, then letting it continue?

m-col:
Looks great, I can follow along and understand everything :smile: the only other test I can think of is testing the restart hook, and of course testing that the X backend has the same set of windows before and after a restart (not sure how feasible that is).

test/test_restart.py
elParaguayo: Yes. If I don't delay the stop then I can no longer send any `eval` commands to the manager so I can't retrieve the filepath for the state file.

test/test_restart.py
elParaguayo: Yes, that would be neater.

elParaguayo:
> Looks great, I can follow along and understand everything  the only other test I can think of is testing the restart hook, and of course testing that the X backend has the same set of windows before and after a restart (not sure how feasible that is).

I'll have a look at the hook.

As for the windows, the test has two windows which it places on different screens/groups. It then checks that these are returned to the same screens/groups when restarting.

m-col:
Ah, gotcha. Great!

elParaguayo:
Weird that this is now failing on CI. Need to work out why.

elParaguayo:
Ah. It's because I'm editing `BareConfig` which isn't a fixture so the changes persist between tests.

PR #2720, opened 2021-08-27
just to play safe

MohneeshSharma:
I think it should be this way to keep it safe if anything doesnt works


MohneeshSharma:
oh.. i am just getting started with all this opensource contribution. I am feeling nice as I learnt something today :)

ramnes:
Cool! If you want to help, have a look at the issues with the label "easy", they should get you into Qtile with useful contributions. :)

PR #2722, opened 2021-08-28
Store windows in list instead of set

### LINKED ISSUE #1541: How are the tasks in TaskList sorted?
It seems that the task in the TaskList are sorted differently than what the `lazy.layout.next()` thinks. As  when I use the `lazy.layout.next()` the focus in the TaskList jumps around but seldom on the task I would expect...

kapilpokhrel:
Since set is unordered, output of TaskList like widget was unpredictable.
So, keeping windows in list keeps them in order by which they are created.

Output of TaskList Before the fix:
window 4 | window 1 | window 3|  window 2 |
After the fix:
window 1 | window 2 | window 3 | window 4 |

Fixes #1541

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -44,7 +44,7 @@ class _Group(CommandObject):
         self.name = name
         self.label = name if label is None else label
         self.custom_layout = layout  # will be set on _configure
-        self.windows = set()
+        self.windows = []
         self.qtile = None
         self.layouts = []
         self.floating_layout = None
@@ -61,7 +61,7 @@ class _Group(CommandObject):
         self.screen = None
         self.current_layout = 0
         self.focus_history = []
-        self.windows = set()
+        self.windows = []
         self.qtile = qtile
         self.layouts = [i.clone(self) for i in layouts]
         self.floating_layout = floating_layout
@@ -235,7 +235,8 @@ class _Group(CommandObject):
 
     def add(self, win, focus=True, force=False):
         hook.fire("group_window_add", self, win)
-        self.windows.add(win)
+        if win not in self.windows:
+            self.windows.append(win)
         win.group = self
         if self.qtile.config.auto_fullscreen and win.wants_to_fullscreen:
             win._float_state = FloatStates.FULLSCREEN
--- a/test/test_group.py
+++ b/test/test_group.py
@@ -1,0 +1,66 @@
+# Copyright (c) 2011 Florian Mounier
+# Copyright (c) 2012, 2014 Tycho Andersen
+# Copyright (c) 2013 Craig Barnes
+# Copyright (c) 2014 Sean Vig
+#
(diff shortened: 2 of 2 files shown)


tych0:
Can you add a test for window ordering so we preserve this property?

kapilpokhrel:
> Can you add a test for window ordering so we preserve this property?

Sorry, i don't think i can do this

tych0:
Why not?

kapilpokhrel:
I am not quite familiar with it enough yet.
Sorry, once again.

tych0:
That's fine, we can hold this PR until you are familiar with it. Generally I don't like to merge bug fixes without tests, since we have nothing preventing us from breaking this code again. Thanks for looking at it!

kapilpokhrel:
Ok, thanks

kapilpokhrel:
> Can you add a test for window ordering so we preserve this property?

You mean, i need to write a test to check weather the windows are stored in correct order by which they are created?

tych0:
Yes, whatever ordering you're trying to enforce with this PR, which I think is creation order.

kapilpokhrel:
Yes, ok

kapilpokhrel:
```
def test_window_order(manager):
	windows_name = ["one","two","three","four","five","six","seven","eight","nine","ten"] # windows to add
	windows = {}	
	# Add windows one by one
	for win in windows_name:
		windows[win] = manager.test_window(win)

	# Windows must be sotred in the same order as they were created
	assert windows_name == manager.c.group.info()["windows"]

	# Randomly remove 5 windows and see if orders remains persistant
	for i in range(5):
		win_to_remove = random.choice(windows_name)
		windows_name.remove(win_to_remove)
		manager.kill_window(windows[win_to_remove])
		del windows[win_to_remove]
		assert windows_name == manager.c.group.info()["windows"]
```
Is this ok?

tych0:
Looks good to me!

On Sun, Aug 29, 2021, at 2:06 AM, Kapil Pokhrel wrote:
> 
> 
> `def test_window_order(manager):
> 	windows_name = ["one","two","three","four","five","six","seven","eight","nine","ten"] # windows to add
> 	windows = {}	
> 	# Add windows one by one
> 	for win in windows_name:
> 		windows[win] = manager.test_window(win)
> 
> 	# Windows must be sotred in the same order as they were created
> 	assert windows_name == manager.c.group.info()["windows"]
> 
> 	# Randomly remove 5 windows and see if orders remains persistant
> 	for i in range(5):
> 		win_to_remove = random.choice(windows_name)
> 		windows_name.remove(win_to_remove)
> 		manager.kill_window(windows[win_to_remove])
> 		del windows[win_to_remove]
> 		assert windows_name == manager.c.group.info()["windows"]
`

tych0:
LGTM, I just squashed the CI fixes into their respective commits, and when this passes CI it's good to go. Thanks!

PR #2723, opened 2021-08-28
Allow KeyChord mode names without staying in mode

### LINKED ISSUE #2647: Allow defining named KeyChords for Chord widget without setting it up as a mode
# Issue description (Feature request)

When creating a KeyChord, specifically with use of the Chord widget, I've noticed that the only way to display something with the Chord widget is to set the KeyChord up as a mode. It would be ideal to be able to name a KeyChord for display in that widget, without having the "Keep mode until Escape is pressed" functionality. 

(In my case, I would find this useful for distinguishing if I have accidentally entered a key chord, or to signify a more 'dangerous' key chord such as power management)

elParaguayo:
This PR allows all KeyChords to be named (via `mode`) but gives the user the option to switch off the mode persistence then the key sequence is complete.

The benefit of this is that the mode name can be displayed in the Chord widget for all KeyChords, not just those with persistent modes.

Closes #2647

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -35,6 +35,7 @@ from typing import TYPE_CHECKING
 from libqtile import configurable, hook, utils
 from libqtile.bar import Bar
 from libqtile.command.base import CommandObject
+from libqtile.log_utils import logger
 
 if TYPE_CHECKING:
     from typing import Any, Callable, ContextManager, Iterable
@@ -93,10 +94,12 @@ class KeyChord:
     submappings:
         A list of :class:`Key` or :class:`KeyChord` declarations to bind in this chord.
     mode:
-        A string with Vim-like mode name. If set, the chord mode will not be left after
-        a keystroke (except for Esc which always leaves the current chord/mode).
-        (Optional)
-
+        Boolean. Setting to ``True`` will result in the chord persisting until
+        Escape is pressed. Setting to ``False`` (default) will exit the chord once
+        the sequence has ended.
+    name:
+        A string to name the chord. The name will be displayed in the Chord
+        widget.
     """
 
     def __init__(
@@ -104,14 +107,26 @@ class KeyChord:
         modifiers: list[str],
         key: str,
         submappings: list[Key | KeyChord],
-        mode: str = "",
-    ) -> None:
+        mode: bool | str = False,
+        name: str = "",
+    ):
         self.modifiers = modifiers
         self.key = key
 
         submappings.append(Key([], "Escape"))
         self.submappings = submappings
         self.mode = mode
+        self.name = name
+
+        if isinstance(mode, str):
+            logger.warning(
+                "The use of `mode` to set the KeyChord name is deprecated. "
+                "Please use `name='%s'` instead. "
+                "'mode' should be a boolean value to set whether the chord is persistent (True) or not.",
+                mode,
+            )
+            self.name = mode
+            self.mode = True
 
     def __repr__(self) -> str:
         return "<KeyChord (%s, %s)>" % (self.modifiers, self.key)
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -394,7 +394,7 @@ class Qtile(CommandObject):
                     )
                     if status in (interface.ERROR, interface.EXCEPTION):
                         logger.error("KB command error %s: %s", cmd.name, val)
-            if self.chord_stack and (self.chord_stack[-1].mode == "" or key.key == "Escape"):
+            if self.chord_stack and (not self.chord_stack[-1].mode or key.key == "Escape"):
                 self.cmd_ungrab_chord()
             return
 
@@ -425,7 +425,7 @@ class Qtile(CommandObject):
     def grab_chord(self, chord: KeyChord) -> None:
         self.chord_stack.append(chord)
         if self.chord_stack:
-            hook.fire("enter_chord", chord.mode)
+            hook.fire("enter_chord", chord.name)
 
         self.ungrab_keys()
         for key in chord.submappings:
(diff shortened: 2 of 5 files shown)


Infinidoge:
Seems a little odd to keep the name `mode` for something that would be more fundamentally a `name`, and `mode` would make more sense as the bool that `keep_mode` is, but it makes sense to not have a breaking change.

elParaguayo:
> Seems a little odd to keep the name `mode` for something that would be more fundamentally a `name`, and `mode` would make more sense as the bool that `keep_mode` is, but it makes sense to not have a breaking change.

Agreed. It's not really a `mode` when used like this but didn't want to break existing configs. As you say, the alternative would be to have a `name` argument.

elParaguayo:
I think there are 3 options we could take here:

1. Follow the approach I've used in the first commit: use `mode` as the variable that sets the chord name. Persistence is turned off by setting a new attribute `keep_mode` to `False`.
2. Add a new `name` argument. Setting this is gives chords a non-persistent name, `mode` retains its original behaviour.
3. Change `mode` to `name` and have a new attribute `keep_named_mode` (or something like that) to enable persistence and which is `True` by default. This would require a `migrate` script.

Infinidoge:
I am personally in favor of either 2 or 3, as 1 feels the least intuitive. 2 would probably be the easiest and most intuitive, though.

libqtile/config.py
```diff
@@ -83,18 +83,25 @@ class KeyChord:
...
-        not be left after a keystroke (except for Esc which always leaves the
-        current chord/mode).
+        A string with vim like mode name. The name will be displayed in the
+        Chord widget when entered. By default, when the mode is set, the chord
+        mode will not be left after a keystroke (except for Esc which always
+        leaves the current chord/mode).
+    name:
+        A string to name the chord. The name will be displayed in the Chord
+        widget but, unlike "mode", the chord will be exited after the final
+        keystroke.
```
m-col, line 93: How is the final keystroke "final"? What does this mean?

test/widgets/test_chord.py
```diff
@@ -25,3 +28,68 @@ def test_chord_widget(fake_bar):
...
+        groups = [libqtile.config.Group("a"), libqtile.config.Group("b")]
+        layouts = [libqtile.layout.stack.Stack(num_stacks=1)]
+        floating_layout = libqtile.resources.default_config.floating_layout
+        screens = [
+            libqtile.config.Screen(
+                top=libqtile.bar.Bar([Chord()], 10)
+            )
+        ]
+
+    manager_nospawn.start(ChordConf)
```
m-col, line 61: Defining a test class and then immediately passing it to `manager_nospawn.start()` is equivalent to defining the class at the module level and using it to parameterise the `manager` fixture... right? Is there any reason why we would favour one over the other? (Or have I missed something?)

elParaguayo: You've not missed anything!

libqtile/config.py
elParaguayo: It's the last key in the sequence e.g. if you have a chord that is `Mod+x` and `z` then `z` is the final keystroke.

m-col: Gotcha... guess I need to read up on the key chord docs

libqtile/core/manager.py
```diff
@@ -388,7 +388,7 @@ def ungrab_keys(self) -> None:
     def grab_chord(self, chord: KeyChord) -> None:
         self.chord_stack.append(chord)
         if self.chord_stack:
-            hook.fire("enter_chord", chord.mode)
+            hook.fire("enter_chord", chord.mode or chord.name)
```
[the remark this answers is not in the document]
m-col: I agree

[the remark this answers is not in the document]
m-col: Although it could just as well be the whole `Chord`, and provide all information?

m-col:
Reading about `KeyChord`s fresh faced having never used them before, it seems confusing that both `mode` and `name` would take strings that name the chord. I agree with @Infinidoge, that `name` should be a string that names the chord, and that `mode` should be a boolean that determines whether or not the chord is a mode.

elParaguayo:
> Reading about `KeyChord`s fresh faced having never used them before, it seems confusing that both `mode` and `name` would take strings that name the chord. I agree with @Infinidoge, that `name` should be a string that names the chord, and that `mode` should be a boolean that determines whether or not the chord is a mode.

If we do that, we'd need a `migrate` script as that would be a breaking change.

Infinidoge:
I'm honestly surprised this has stayed open for this long.
Is there anything left that needs to be done before this can be merged?

elParaguayo:
Apologies. This one dropped off my list.

I think I need to rework it so that `mode` is a boolean and add some deprecation warnings for people using the old syntax.

Shouldn't take too long to do.

libqtile/config.py
```diff
@@ -93,25 +94,40 @@ class KeyChord:
...
     """
 
     def __init__(
         self,
         modifiers: list[str],
         key: str,
         submappings: list[Key | KeyChord],
-        mode: str = "",
-    ) -> None:
+        mode: bool | str = False,
```
m-col, line 110: Should the type just be `bool`, rather than `bool | str`?

m-col: Oh I see - this is just for gradual deprecation. Nvm!

libqtile/config.py
```diff
@@ -93,25 +94,40 @@ class KeyChord:
...
+
+        if isinstance(mode, str):
+            logger.warning(
+                "The use of `mode` to set the KeyChord name is deprecated. "
+                "Please use `name='%s'` instead. "
+                "'mode' should be a boolean value to set whether the chord is persistent (True) or not.",
+                mode,
+            )
+            # mypy doesn't recognise that mode must be a string here.
+            self.name = self.mode  # type: ignore
```
m-col, line 129: Would it, if you used `mode` instead of `self.mode`? I suspect it's inferring the type `bool | str` for `self.mode` but only `mode` has the isinstance check

m-col (approved):
LGTM!

libqtile/config.py
elParaguayo: Good point. Let me check.

libqtile/config.py
elParaguayo: Yes. That's it.

libqtile/config.py
elParaguayo: Seems to work.

Infinidoge:
:tada:

Issue #2552, opened 2021-06-14
Exposing core cmd_s?


m-col:
# Issue description

For some backend-specific functionality the cores currently can implement a method of the base class that is added to the base class just for that one backend. An example of this is `Core.change_vt`, which is used exclusively by the Wayland core. Similarly the Wayland core would likely need a way to change the keymap, kinda like setxkbmap in X (currently this is already implemented internally, it would just be useful to expose it as a cmd_). Currently the only way to change keymap under Wayland is clicking the KeyboardLayout widget to switch between maps, or if someone was so inclined they could do:
```
qtile cmd-obj -o cmd -f eval -a "self.core.set_keymap('gb')"
```
Perhaps a nicer approach is to just make `Core` one of the command objects that exposes cmd_ methods:
```
qtile cmd-obj -o core -f set_keymap -a gb
```
config example:
```
    lazy.core.change_vt(2)
```
This would let us remove base methods for these methods from the base Core, which is a bit nicer if they are specific to a single core. It would also let use remove wrappers from `Qtile` e.g. `Qtile.change_vt` which basically just proxies to the core. It also has the advantage that it would use existing infrastructure for error handling when a cmd_ method is not present on the target object.

elParaguayo:
What are the disadvantages (if any) to doing this?

m-col:
Increased API surface area I suppose

m-col:
`libqtile.command` is pretty nebulous to me. Any pointers for how I can make this happen, beyond adding cmd_ methods?

elParaguayo:
@m-col having played around with the object graph a bit, I'm happy to take a look at adding `core` to the graph.

As for the `cmd_` bit, I don't love this approach and have seen a few issues where users have got confused. I would have liked a decorator approach to expose methods e.g.
``` python
@command
def restart(self):
    ...
```
This would remove confusion over the naming.

m-col:
Interesting! I like this idea if it can be done nicely.

Would the decorator be applied to the base class in the case of `Window` and others that are subclassed? That way there is a central point where commands are defined (and can be documented from), with optional additional methods appearing on subclasses (e.g. on layouts, widgets, which might have unique commands).

It also means we can remove having both an implementation (e.g. `Window.kill`) and command (`Window.cmd_kill`), using only the former within libqtile.

elParaguayo:
[...]

PR #2725, opened 2021-08-28
Add core/backend to object graph

elParaguayo:
As per #2552, this adds `core` to the object graph which will allow us to expose backend commands to shell commands etc.

Currently, the node is called `core` but happy to rename to `backend` if that's better.

Also, I've only linked the node to root but, in theory, it could probably be linked to all the other nodes. However, I don't really see a reason to do that. Again, happy to listen to differing opinions.

### DIFF
--- a/docs/_static/diagrams/object-graph-orig.dot
+++ b/docs/_static/diagrams/object-graph-orig.dot
@@ -24,12 +24,16 @@ digraph G {
     node [style="filled", color=Purple, fillcolor=Violet, label="window"];
     window;
 
+    node [style="filled", color=SlateBlue, fillcolor=SlateBlue1, label="core"];
+    core;
+
     root -> bar;
     root -> group;
     root -> layout;
     root -> screen;
     root -> widget;
     root -> window;
+    root -> core;
 
     bar -> screen [dir=both];
     bar -> widget [dir=both];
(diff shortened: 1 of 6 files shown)


m-col:
Nice. 

> Currently, the node is called core but happy to rename to backend if that's better.

Hmm, not sure actually, though I suppose backend might make things more obvious if there is only one name? Same thing could be said to argue changing "core" to "backend" everywhere in the code. What do you think?

> Also, I've only linked the node to root but, in theory, it could probably be linked to all the other nodes. However, I don't really see a reason to do that. Again, happy to listen to differing opinions.

I like what you've done, no point linking it to anything else.

elParaguayo:
I'm ok with keeping it as `core`.

m-col (approved):
Great, LGTM!

elParaguayo:
Cool. If you merge this, I'll leave you to add whichever methods you wanted to expose.

m-col:
Noice thanks

ramnes:
Can we replace root by core?

m-col:
> Can we replace root by core?

Isn't root the `Qtile` manager instance?

elParaguayo:
Yes `root` is the `Qtile` object. `core` is the backend.

PR #2726, opened 2021-08-29
Max layout : Add support for margins

ImAReplicant:
Adds the ability to have a margin to Max layout for a nicer look when there are margins to bar.


tych0:
Can you add a test for this?

Issue #1360, opened 2019-06-27
Generate a list of all lazy commands


ramnes:
See #1358 for rationale.

We should have all the available lazy commands listed in the documentation, either on the ["Lazy objects"](http://docs.qtile.org/en/latest/manual/config/lazy.html) or directly along the resource that provide them (for example on the ["Built-in Layouts"](http://docs.qtile.org/en/latest/manual/ref/layouts.html) page).

elParaguayo:
I'm looking at lazy docs and widget commands.

PR #2727, opened 2021-08-29
Using decorators to expose commands

elParaguayo:
Currently, commands that wish to be exposed via IPC need to be prefixed with `cmd_` but this prefix is then removed when calling commands. This can cause confusion for users.

This PR replaces the prefix mechanism with a new decorator: `@CommandObject.expose`. Methods decorated with this will be exposed via IPC but the method name is consistent.

This is draft for now as I'm interested in opinions here and there's a lot more work to do to completely convert it. 

Personally, I've never particularly liked the `cmd_` prefix approach and I think decorators are a clearer solution but I appreciate there may be some differing views.

Pros:

- Consistent naming of methods
- Removes some unnecessary code (e.g. `cmd_info()` calling `info()`. We can remove `cmd_info` and just decorate `info`)

Cons:

- Might be a breaking change for users (e.g. configs may have scripts making references to `cmd_` methods). May be able to mitigate with `migrate` scripts.
- Underlying decorator code is slightly harder to follow than the code used to process `cmd_` methods. I've added fairly detailed comments to explain how it works.

As this PR is not complete, I've not touched the tests yet so it will be interesting to see how many I've broken at this stage!!

EDIT:
```
2021-08-29T17:38:38.2216360Z ====== 225 failed, 189 passed, 1 skipped, 610 errors in 416.32s (0:06:56) ======
```
:thumbsup:

### DIFF
--- a/docs/manual/commands/index.rst
+++ b/docs/manual/commands/index.rst
@@ -178,10 +178,11 @@ which will lead to the way the commands are dispatched.
 All of the configured objects setup by Qtile are ``CommandObject`` subclasses.
 These objects are so named because we can issue commands against them using the
 command scripting API.  Looking through the code, the commands that are exposed
-are commands named ``cmd_*``.  When writing custom layouts, widgets, or any
-other object, you can add your own custom ``cmd_`` functions and they will be
-callable using the standard command infrastructure.  An available command can
-be extracted by calling ``.command()`` with the name of the command.
+are commands that are decorated with the ``@expose_command()`` decorator.
+When writing custom layouts, widgets, or any other object, you can add your own
+custom functions and, once you add the decorator, they will be callable using the
+standard command infrastructure. An available command can be extracted by calling
+``.command()`` with the name of the command.
 
 In addition to having a set of associated commands, each command object also
 has a collection of items associated with it.  This is what forms the graph
--- a/docs/sphinx_qtile.py
+++ b/docs/sphinx_qtile.py
@@ -19,8 +19,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import builtins
-import functools
 import importlib
 import inspect
 import json
@@ -28,6 +26,7 @@ import os
 import pprint
 from pathlib import Path
 from subprocess import CalledProcessError, run
+from unittest.mock import MagicMock
 
 from docutils import nodes
 from docutils.parsers.rst import Directive
@@ -188,7 +187,10 @@ class QtileClass(SimpleDirectiveMixin, Directive):
         }
         if context['commandable']:
             context['commands'] = [
-                attr for attr in dir(obj) if attr.startswith('cmd_')
+                # Command methods have the "_cmd" attribute so we check for this
+                # However, some modules are Mocked so we need to exclude them
+                attr.__name__ for _, attr in inspect.getmembers(obj)
+                if hasattr(attr, "_cmd") and not isinstance(attr, MagicMock)
             ]
 
         rst = qtile_class_template.render(**context)
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -9,7 +9,7 @@ from abc import ABCMeta, abstractmethod
 import cairocffi
 
 from libqtile import drawer, pangocffi, utils
-from libqtile.command.base import CommandError, CommandObject
+from libqtile.command.base import CommandError, CommandObject, expose_command
 from libqtile.log_utils import logger
 
 if typing.TYPE_CHECKING:
@@ -123,7 +123,8 @@ class Core(CommandObject, metaclass=ABCMeta):
         """Get the keysym for a key from its name"""
         raise NotImplementedError
 
-    def cmd_info(self) -> dict:
+    @expose_command()
+    def info(self) -> dict[str, Any]:
         """Get basic information about the running backend."""
         return {"backend": self.name, "display_name": self.display_name}
 
@@ -143,7 +144,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
         self.borderwidth: int = 0
         self.name: str = "<no name>"
         self.reserved_space: tuple[int, int, int, int] | None = None
-        # Window.cmd_static sets this in case it is hooked to client_new to stop the
+        # Window.static sets this in case it is hooked to client_new to stop the
         # Window object from being managed, now that a Static is being used instead
         self.defunct: bool = False
 
@@ -160,11 +161,13 @@ class _Window(CommandObject, metaclass=ABCMeta):
     def unhide(self) -> None:
         """Unhide the window"""
 
-    def cmd_is_visible(self) -> bool:
+    @expose_command()
+    def is_visible(self) -> bool:
         """Is this window visible (i.e. not hidden)?"""
         return False
 
     @abstractmethod
+    @expose_command()
     def kill(self) -> None:
         """Kill the window"""
 
@@ -209,6 +212,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
         self._opacity = opacity
 
     @abstractmethod
+    @expose_command()
     def place(
         self,
         x,
@@ -230,6 +234,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
         return None
 
     @abstractmethod
+    @expose_command()
     def info(self) -> dict[str, Any]:
         """
         Return information on this window.
@@ -246,10 +251,6 @@ class _Window(CommandObject, metaclass=ABCMeta):
 
         """
         return {}
-
-    def cmd_info(self) -> dict:
-        """Return a dictionary of info."""
-        return self.info()
 
 
 class Window(_Window, metaclass=ABCMeta):
@@ -325,20 +326,9 @@ class Window(_Window, metaclass=ABCMeta):
         return match.compare(self)
 
     @abstractmethod
-    def focus(self, warp: bool) -> None:
+    @expose_command()
+    def focus(self, warp: bool = True) -> None:
         """Focus this window and optional warp the pointer to it."""
-
-    @abstractmethod
-    def togroup(
-        self, group_name: str | None = None, *, switch_group: bool = False, toggle: bool = False
-    ) -> None:
-        """Move window to a specified group
-
-        Also switch to that group if switch_group is True.
-
-        If `toggle` is True and and the specified group is already on the screen,
-        use the last used group as target instead.
-        """
 
     @property
     def has_focus(self):
@@ -360,86 +350,91 @@ class Window(_Window, metaclass=ABCMeta):
         """Paint the window borders with the given color(s) and width"""
 
     @abstractmethod
-    def cmd_focus(self, warp: bool = True) -> None:
-        """Focuses the window."""
-
-    def cmd_match(self, *args, **kwargs) -> bool:
-        return self.match(*args, **kwargs)
-
-    @abstractmethod
-    def cmd_get_position(self) -> tuple[int, int]:
+    @expose_command()
+    def get_position(self) -> tuple[int, int]:
         """Get the (x, y) of the window"""
 
     @abstractmethod
-    def cmd_get_size(self) -> tuple[int, int]:
+    @expose_command()
+    def get_size(self) -> tuple[int, int]:
         """Get the (width, height) of the window"""
 
     @abstractmethod
-    def cmd_move_floating(self, dx: int, dy: int) -> None:
+    @expose_command()
+    def move_floating(self, dx: int, dy: int) -> None:
         """Move window by dx and dy"""
 
     @abstractmethod
-    def cmd_resize_floating(self, dw: int, dh: int) -> None:
+    @expose_command()
+    def resize_floating(self, dw: int, dh: int) -> None:
         """Add dw and dh to size of window"""
 
     @abstractmethod
-    def cmd_set_position_floating(self, x: int, y: int) -> None:
+    @expose_command()
+    def set_position_floating(self, x: int, y: int) -> None:
         """Move window to x and y"""
 
     @abstractmethod
-    def cmd_set_position(self, x: int, y: int) -> None:
+    @expose_command()
+    def set_position(self, x: int, y: int) -> None:
         """
         Move floating window to x and y; swap tiling window with the window under the
         pointer.
         """
 
     @abstractmethod
-    def cmd_set_size_floating(self, w: int, h: int) -> None:
+    @expose_command()
+    def set_size_floating(self, w: int, h: int) -> None:
         """Set window dimensions to w and h"""
 
     @abstractmethod
-    def cmd_place(
-        self, x, y, width, height, borderwidth, bordercolor, above=False, margin=None
-    ) -> None:
-        """Place the window with the given position and geometry."""
-
-    @abstractmethod
-    def cmd_toggle_floating(self) -> None:
+    @expose_command()
+    def toggle_floating(self) -> None:
         """Toggle the floating state of the window."""
 
     @abstractmethod
-    def cmd_enable_floating(self) -> None:
+    @expose_command()
+    def enable_floating(self) -> None:
         """Float the window."""
 
     @abstractmethod
-    def cmd_disable_floating(self) -> None:
+    @expose_command()
+    def disable_floating(self) -> None:
         """Tile the window."""
 
     @abstractmethod
-    def cmd_toggle_maximize(self) -> None:
-        """Toggle the maximize state of the window."""
-
-    @abstractmethod
-    def cmd_toggle_minimize(self) -> None:
-        """Toggle the minimize state of the window."""
-
-    @abstractmethod
-    def cmd_toggle_fullscreen(self) -> None:
+    @expose_command()
+    def toggle_maximize(self) -> None:
         """Toggle the fullscreen state of the window."""
 
     @abstractmethod
-    def cmd_enable_fullscreen(self) -> None:
+    @expose_command()
+    def toggle_minimize(self) -> None:
+        """Toggle the minimized state of the window."""
+
+    @abstractmethod
+    @expose_command()
+    def toggle_fullscreen(self) -> None:
+        """Toggle the fullscreen state of the window."""
+
+    @abstractmethod
+    @expose_command()
+    def enable_fullscreen(self) -> None:
         """Fullscreen the window"""
 
     @abstractmethod
-    def cmd_disable_fullscreen(self) -> None:
+    @expose_command()
+    def disable_fullscreen(self) -> None:
         """Un-fullscreen the window"""
 
     @abstractmethod
-    def cmd_bring_to_front(self) -> None:
+    @expose_command()
+    def bring_to_front(self) -> None:
         """Bring the window to the front"""
 
-    def cmd_togroup(
+    @abstractmethod
+    @expose_command()
+    def togroup(
         self,
         group_name: str | None = None,
         groupName: str | None = None,  # Deprecated  # noqa: N803
@@ -455,13 +450,29 @@ class Window(_Window, metaclass=ABCMeta):
 
         `groupName` is deprecated and will be dropped soon. Please use `group_name`
         instead.
+
+        Examples
+        ========
+
+        Move window to current group::
+
+            togroup()
+
+        Move window to group "a"::
+
+            togroup("a")
+
+        Move window to group "a", and switch to group "a"::
+
+            togroup("a", switch_group=True)
         """
         if groupName is not None:
-            logger.warning("Window.cmd_togroup's groupName is deprecated; use group_name")
+            logger.warning("Window.togroup's groupName is deprecated; use group_name")
             group_name = groupName
         self.togroup(group_name, switch_group=switch_group, toggle=toggle)
 
-    def cmd_toscreen(self, index: int | None = None) -> None:
+    @expose_command()
+    def toscreen(self, index: int | None = None) -> None:
         """Move window to a specified screen.
 
         If index is not specified, we assume the current screen
@@ -486,11 +497,9 @@ class Window(_Window, metaclass=ABCMeta):
                 raise CommandError("No such screen: %d" % index)
         self.togroup(screen.group.name)
 
-    def cmd_opacity(self, opacity: float) -> None:
-        """Set the window's opacity.
-
-        The value must be between 0 and 1 inclusive.
-        """
+    @expose_command()
+    def set_opacity(self, opacity: float) -> None:
+        """Set the window's opacity"""
         if opacity < 0.1:
             self.opacity = 0.1
         elif opacity > 1:
@@ -498,27 +507,19 @@ class Window(_Window, metaclass=ABCMeta):
         else:
             self.opacity = opacity
 
-    def cmd_down_opacity(self) -> None:
+    @expose_command()
+    def down_opacity(self) -> None:
         """Decrease the window's opacity by 10%."""
-        if self.opacity > 0.2:
-            # don't go completely clear
-            self.opacity -= 0.1
-        else:
-            self.opacity = 0.1
+        self.set_opacity(self.opacity - 0.1)
 
-    def cmd_up_opacity(self) -> None:
+    @expose_command()
+    def up_opacity(self) -> None:
         """Increase the window's opacity by 10%."""
-        if self.opacity < 0.9:
-            self.opacity += 0.1
-        else:
-            self.opacity = 1
+        self.set_opacity(self.opacity + 0.1)
 
     @abstractmethod
-    def cmd_kill(self) -> None:
-        """Kill the window. Try to be polite."""
-
-    @abstractmethod
-    def cmd_static(
+    @expose_command()
+    def static(
         self,
         screen: int | None = None,
         x: int | None = None,
@@ -532,7 +533,8 @@ class Window(_Window, metaclass=ABCMeta):
         """
         self.defunct = True
 
-    def cmd_center(self) -> None:
+    @expose_command()
+    def center(self) -> None:
         """Centers a floating window on the screen."""
         if not self.floating:
             return
@@ -542,16 +544,16 @@ class Window(_Window, metaclass=ABCMeta):
 
         screen = self.group.screen
 
-        x = (screen.width - self.width) // 2  # type: ignore
-        y = (screen.height - self.height) // 2  # type: ignore
+        x = (screen.width - self.width) // 2
+        y = (screen.height - self.height) // 2
 
         self.place(
             x,
             y,
-            self.width,  # type: ignore
-            self.height,  # type: ignore
+            self.width,
+            self.height,
             self.borderwidth,
-            self.bordercolor,  # type: ignore
+            self.bordercolor,
             above=True,
             respect_hints=True,
         )
@@ -601,6 +603,7 @@ class Static(_Window, metaclass=ABCMeta):
     def __repr__(self):
         return "%s(name=%r, wid=%i)" % (self.__class__.__name__, self.name, self.wid)
 
+    @expose_command()
     def info(self) -> dict:
         """Return a dictionary of info."""
         return dict(
@@ -614,7 +617,8 @@ class Static(_Window, metaclass=ABCMeta):
         )
 
     @abstractmethod
-    def cmd_bring_to_front(self) -> None:
+    @expose_command()
+    def bring_to_front(self) -> None:
         """Bring the window to the front"""
 
 
--- a/libqtile/command/base.py
+++ b/libqtile/command/base.py
@@ -28,8 +28,10 @@ import abc
 import inspect
 import sys
 import traceback
+from functools import partial
 from typing import TYPE_CHECKING
 
+from libqtile.configurable import Configurable
 from libqtile.log_utils import logger
 
 if TYPE_CHECKING:
@@ -38,6 +40,50 @@ if TYPE_CHECKING:
     from libqtile.command.graph import SelectorType
 
     ItemT = Optional[tuple[bool, list[str | int]]]
+
+
+def expose_command(name: Callable | str | list[str] | None = None) -> Callable:
+    """
+    Decorator to expose methods to the command interface.
+
+    The exposed command will have the name of the defined method.
+
+    Methods can also be exposed via multiple names by passing the names to this
+    decorator.
+
+    e.g. if a layout wants "up" and "previous" to call the
+    same method:
+
+    @expose_command("previous")
+    def up(self):
+        ...
+
+    `up` will be exposed as `up` and `previous`.
+
+    Multiple names can be passed as a list.
+    """
+
+    def wrapper(func: Callable):
+        setattr(func, "_cmd", True)
+        if name is not None:
+            if not hasattr(func, "_mapping"):
+                setattr(func, "_mapping", list())
+            if isinstance(name, list):
+                func._mapping += name  # type:ignore
+            elif isinstance(name, str):
+                func._mapping.append(name)  # type:ignore
+            else:
+                logger.error("Unexpected value received in command decorator: %s", name)
+        return func
+
+    # If the decorator is added with no parentheses then we should treat it
+    # as if it had been i.e. expose the decorated method
+    if callable(name):
+        func = name
+        name = None
+        return wrapper(func)
+
+    return wrapper
 
 
 class SelectError(Exception):
@@ -60,10 +106,73 @@ class CommandException(Exception):
 class CommandObject(metaclass=abc.ABCMeta):
     """Base class for objects that expose commands
 
-    Each command should be a method named `cmd_X`, where X is the command name.
+    Any command to be exposed should be decorated with
+    `@expose_command()` (classes that are not explicitly
+    inheriting from CommandObject will need to import the module)
     A CommandObject should also implement `._items()` and `._select()` methods
     (c.f. docstring for `.items()` and `.select()`).
     """
+
+    def __new__(cls, *args, **kwargs):
+
+        # Check which level of command object has been parsed
+        # This test ensures inherited classes don't stop additional
+        # methods from being exposed.
+        # For example, if widget.TextBox has already been parsed, a subsequent
+        # call to initialise a new TextBox will return here. However, if a user
+        # subclasses TextBox for a new widget then that new widget will still
+        # be parsed here to check for new commands.
+        if getattr(cls, "_command_object", "") == cls.__name__:
+            super().__new__(cls)
+
+        commands = {}
+        cmd_s = set()
+
+        # We need to iterate over the class's inherited classes in reverse order
+        # We reverse the order so the exposed command will always be the latest
+        # definition of the method.
+        for c in reversed(list(cls.__mro__)):
+
+            for method_name in list(c.__dict__.keys()):
+                method = getattr(c, method_name, None)
+
+                if method is None:
+                    continue
+
+                # If the command has been exposed, add it to our dictionary
+                # If the method name is already in our dictionary then bind the
+                # latest definition to that command
+                if hasattr(method, "_cmd") or method_name in commands:
+                    commands[method_name] = method
+                # For now, we'll accept the old format `cmd_` naming scheme for
+                # exposing commands.
+                # NOTE: This will be deprecated in the future
+                elif method_name.startswith("cmd_"):
+                    cmd_s.add(method_name)
+                    commands[method_name[4:]] = method
+
+                # Expose additional names
+                for mapping in getattr(method, "_mapping", list()):
+                    setattr(cls, mapping, method)
+                    commands[mapping] = method
+
+        if cmd_s:
+            names = ", ".join(cmd_s)
+            msg = (
+                f"The use of the 'cmd_' prefix to expose commands via IPC "
+                f"is deprecated. Methods should use the "
+                f"@expose_command() decorator instead. "
+                f"Please update: {names}"
+            )
+            logger.warning("Deprecation Warning: %s", msg)
+
+        # Record the object as being parsed.
+        cls._command_object = cls.__name__
+
+        # Store list of exposed commands
+        cls._commands = commands
+
+        return super().__new__(cls)
 
     def select(self, selectors: list[SelectorType]) -> CommandObject:
         """Return a selected object
@@ -92,8 +201,12 @@ class CommandObject(metaclass=abc.ABCMeta):
             obj = maybe_obj
         return obj
 
+    @expose_command()
     def items(self, name: str) -> tuple[bool, list[str | int] | None]:
-        """Build a list of contained items for the given item class
+        """
+        Build a list of contained items for the given item class.
+
+        Exposing this allows __qsh__ to navigate the object graph.
 
         Returns a tuple `(root, items)` for the specified item class, where:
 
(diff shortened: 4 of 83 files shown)


libqtile/command/base.py
```diff
@@ -27,12 +27,42 @@
...
+        # modules are fully imported. So, we first register a tuple of
+        # (the module containing the class, the class name, the method).
+        # Once qtile is starting we call a separate method to generate
+        # references to the class type containing the function.
+        register.append(
+            (
+                inspect.getmodule(func),
+                # This takes the dotted name for the method and trims it
+                # to give the module.
+                func.__qualname__.split('.<locals>', 1)[0].rsplit('.', 1)[0],
```
elParaguayo, line 60: Got this line from StackOverflow. I'm pretty sure we don't need the `.split('.<locals>', 1)[0]` part as none of these methods should be in `locals`.

libqtile/command/base.py
```diff
@@ -27,12 +27,42 @@
...
 import traceback
-from typing import Callable, List, Optional, Tuple, Union
+from functools import partial
+from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
 
 from libqtile.command.graph import SelectorType
 from libqtile.log_utils import logger
 
 ItemT = Optional[Tuple[bool, List[Union[str, int]]]]
+AnyClass = TypeVar("AnyClass")
```
elParaguayo, line 37: I'm still not 100% on type hinting. Basically, I want to allow any class type.

m-col:
+1 for the change in approach.

`@command` might be better than `@expose` so that it's more clear that it comes from `cmd_` and represents the same thing.

> Might be a breaking change for users (e.g. configs may have scripts making references to cmd_ methods). May be able to mitigate with migrate scripts.

We can work around this by checking for `cmd_` prefixes in `__getattr__` and stripping it before logging a deprecation message and continuing. Along with a migration of course.

tych0:
I generally like the idea, so thank you for doing this. A couple of thoughts:

1. it would be great if we could get rid of "aliases", e.g. _Window() has `def cmd_info(self): self.info()`.
2. I think we probably also need a migration just to `sed s/cmd_//` for people who are calling `cmd_` functions via lazy.function() or whatever don't break their configs.

flacjacket:
I think this is a great idea! I took a stab at trying to make this a bit cleaner, here is what I came up with: https://github.com/flacjacket/qtile/commit/82be2d2a4f1f963be6baac04038d385b0417f3c5
This is able to get up to 12 failed, 985 passed on the unit tests, so a nice step in the right direction there.

elParaguayo:
@flacjacket did you have to make your way so much neater than mine?!

If people like the idea, I'm happy to push on with it.

elParaguayo:
> `@command` might be better than `@expose` so that it's more clear that it comes from `cmd_` and represents the same thing.

I agree, this would be nicer, the problem is that there is a `command` method already in `CommandObject`.

I'll apply the patch from flacjacket's commit as it's _waaay_ better than my approach and work through the test errors. We can do a search and replace on the decorator name if people don't like `expose`.

elParaguayo:
I may have been too quick to praise flacjacket!

The simplification is too simple. It's not enough to store the method name in the class attribute, you've got to know which module it belongs to. If not, the commands are available to all `CommandObject`s, whether the method belongs to them or not.

flacjacket:
Hmm, do you know how that is happening? This seems to work as expected and stores a different set of commands for the `X` and `Y` classes:
```
import inspect


def command(func):
    func._cmd = True
    return func


class Base:
    @classmethod
    def get_commands(cls):
        cls._commands = {}
        for method_name, method in inspect.getmembers(cls):
            if getattr(method, "_cmd", False):
                cls._commands[method_name] = method

    @property
    def commands(self):
        if not hasattr(self, "_commands"):
            self.get_commands()

        return self._commands


class X(Base):
    @command
    def cmdfunc(self):
        pass


class Y(Base):
    @command
    def cmdfunc2(self):
        pass


x = X()
y = Y()
print(hasattr(x, "_commands"))
print(x.commands)
print(hasattr(x, "_commands"))
print(y.commands)

x2 = X()
print(hasattr(x2, "_commands"))
print(x2.commands)
```
I guess if a super class has the commands pulled for it, then the child class wouldn't regenerate the dictionary of `_commands`. A simple fix for this would be to not cache the commands in `_commands` and always regenerate the set (might be a bit expensive, but I think should only be costs paid when using the command interface via callbacks in a config or when used interactively, which seems like it should be ok). A more complex solution would be to create the '_commands' via a metaclass or in `__new__`. An example of the metaclass is here: https://github.com/flacjacket/qtile/commit/c3ebae736f06cb1f52ebdd64620ca2b736b22689

flacjacket:
> You'll also run in to problems where commands have repeated names (e.g. `info()`) as this will only catch one.

Ahhhh! That explains why some of the tests were having problems with `layout.info()` invoking that command. So we would probably want to cut over a good portion of the code base at once, at least cut over all of the cases where the command is overridden like this.

elParaguayo:
Interesting. I know nothing about metaclasses. However, if looks like a neater solution than my ugly parsing of `__qualname__` which is what I really wanted to avoid. It seems to work very nicely.

Thanks.

m-col:
I wonder if this will make it easier to get a list of all lazy commands for the docs (https://github.com/qtile/qtile/issues/1360) and better handle argument types when args are passed from a command line (https://github.com/qtile/qtile/issues/2433).

elParaguayo:
Quite possibly, yes, but let's get this done first!

As an update, I've got it working and 100% of tests are completing. 

I'll try to push an update to this PR later today but I also need to set up a discussions page as there were a few interesting bits to deal with along the way. I'd like to highlight what decisions I made and get people's thoughts on them.

elParaguayo:
Let's continue discussion here: #2774.

elParaguayo:
This change breaks the docs: https://docs.qtile.org/en/latest/manual/ref/commands.html

Bit more work to do still...

elParaguayo:
> I wonder if this will make it easier to get a list of all lazy commands for the docs (#1360) and better handle argument types when args are passed from a command line (#2433).

The first bit should be fairly easy. I'll fix broken docs here first (fairly quick, I hope) and will then look to add layouts.

elParaguayo:
Are we still keen on this one? Personally, I like it!

Rebasing this one's going to be fun: 19 conflicting files...

m-col:
> Are we still keen on this one? Personally, I like it!

Absolutely!

elParaguayo:
Oh dear... rebasing this has gone very wrong. Could take a while.

libqtile/backend/base.py
```diff
@@ -212,10 +213,6 @@ def info(self) -> Dict[str, Any]:
         """Return information on this window."""
         return {}
 
-    def cmd_info(self) -> Dict:
-        """Return a dictionary of info."""
-        return self.info()
-
```
m-col, line 216: We are removing `cmd_info` but keeping `info` here. In the case where a command was required by a base class, does the abstract method need decorating, or the implemention methods, or both?

m-col: If I've understood correctly the abstract method doesn't need it; the `expose` decorator is strictly providing functionality for the IPC. I wonder if we can make it so that any implementations of cmd_ abstract methods are necessarily exposed, otherwise the classes run the risk of not always matching each other. This *is* ensured by using `cmd_` prefix. Perhaps this could be accomplished by keep track of exposed methods on the object, rather than the methods themselves. I'm just investigating this.

elParaguayo: I think that's correct. The decorator is only relevant for the method that is ultimately exposed by the class. If a subclass has a decorated method but this is overriden by an undecorated method then the command will no longer be exposed.

Your idea sounds good.

m-col:
Just rebased against master and force pushed, leaving the code otherwise unchanged.

As discussed I appended a new (untested) commit trying to describe what I think might be an improved approach for collecting all commands for a given class. The two changes from the original approach are:

- Only collect commands once per class i.e. if `hasattr(cls, "_commands")` then skip all that code, saving the result as a class attribute rather than instance attribute.
- Also scan base classes when collecting commands on a child class. The way this is implemented means that the decorator is only required on base methods, and is redundant (and does nothing) on child methods that override decorator base classes. The intention here is that on a base class we would decorate base commands, and child classes only need to decorate commands that they define but are *not* part of their base class (e.g. custom layout/widget commands). This avoids the need to manually keep sibling classes and base and child classes in all sync.

A couple other comments on the PR:

- It might be cleaner to merge the `expose` and `expose_add` decorators (i.e. with the command aliases being a kwarg to expose).
- Do we need to rename `Layout.add` to `Layout.add_client`? This will cause trouble with anybody who defines any custom layout logic.
- Do we need to rename `Qtile.screens` and `Qtile.groups`? This will similarly cause a lot of problems for almost everybody using a custom function in their config!

elParaguayo:
Thanks. Busy day today but I'll pull down your changes and take a look when I can.

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
...
+            names = " ".join(cmd_s)
+            msg = (
+                f"The use of the 'cmd_' prefix to expose commands via IPC "
+                f"is deprecated. Methods should use the "
+                f"@CommandObject.expose decorator instead. "
+                f"Please update: {names}"
+            )
+            logger.warning(f"Deprecation Warning: {msg}")
+
+        return instance
```
elParaguayo, line 96: I can see what this is trying to do (and the code looks nice).

The obvious issue I can see is that we're not assigning the `commands` object to any class or instance attribute. That being said, if I add a line like `cls._commands = commands`, the object seems to be empty so maybe it's going wrong somewhere else...

Will do some more digging as it feels like this should be the approach we use.

elParaguayo: Partly my mistake, should have been `instance._commands = commands` not `cls` (`cls` is the `_CommandMeta` class...). Still getting some empty `commands` objects here which I wouldn't expect.

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
     """Error raised while executing a command"""
 
 
-class CommandObject(metaclass=abc.ABCMeta):
+class _CommandMeta(abc.ABCMeta):
+    def __new__(cls, name, bases, dct):
+        instance = super().__new__(cls, name, bases, dct)
+
+        if hasattr(cls, "_commands"):
```
elParaguayo, line 63: This should probably be `instance` not `cls`.

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
...
+            return instance
+
+        commands = {}
+        cmd_s = set()
+
+        for c in bases + (cls,):
```
elParaguayo, line 69:
```suggestion
        for c in bases + (instance,):
```

libqtile/command/base.py
elParaguayo: However, this cause a problem because it sets `CommandObject._commands` which means every object that inherits `CommandObject` will trigger this return.

Do we need to keep a dict of names that have had commands added?

libqtile/command/base.py
elParaguayo: Also looks like `bases` is only the directly inherited class, doesn't seem to be a full MRO.

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
...
+            for method_name in dct.keys():
```
elParaguayo, line 70: `dct` only seems to be methods defined in the class, not inherited ones.

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
...
+                    continue
+
+                if hasattr(method, "_cmd") or method_name in commands:
+                    commands[method_name] = method
+                elif method_name.startswith("cmd_"):
+                    cmd_s.add(method_name)
+                    commands[method_name[4:]] = method
+
+                for mapping in getattr(method, "_mapping", list()):
+                    setattr(cls, mapping, method)
```
elParaguayo, line 83:
```suggestion
                    setattr(instance, mapping, method)
```

libqtile/command/base.py
```diff
@@ -54,14 +56,85 @@ class CommandException(Exception):
...
+        if cmd_s:
+            names = " ".join(cmd_s)
+            msg = (
+                f"The use of the 'cmd_' prefix to expose commands via IPC "
+                f"is deprecated. Methods should use the "
+                f"@CommandObject.expose decorator instead. "
+                f"Please update: {names}"
+            )
+            logger.warning(f"Deprecation Warning: {msg}")
+
```
elParaguayo, line 95:
```suggestion

    instance._commands = commands

```

libqtile/core/manager.py
```diff
@@ -311,7 +311,7 @@ def finalize(self) -> None:
         self.core.finalize()
 
     def _process_screens(self, reloading=False) -> None:
-        current_groups = [s.group for s in self.screens if hasattr(s, "group")]
+        current_groups = [s.group for s in self.screens() if hasattr(s, "group")]
```
elParaguayo, line 314: This was the point you were referring to with `qtile.screens` and `qtile.groups`.

If you don't want to change the property name then we can change the exposed method name. I would assume this is only really relevant for tests.

elParaguayo:
Still some commands missing (`window.kill` for example) and some exposed commands not working (e.g. `window.info`)...

libqtile/command/base.py
```diff
@@ -54,14 +56,102 @@ class CommandException(Exception):
...
+class _CommandMeta(abc.ABCMeta):
+    _updated = set()
+
+    def __new__(cls, name, bases, dct):
+        # Get the class definition
+        instance = super().__new__(cls, name, bases, dct)
+
+        # If we've already processed this class, we can pass this to __init__ now
+        if name in cls._updated:
+            return instance
```
elParaguayo, line 68: This actually causes problems. Looks like we need to do the parsing of commands for each `instance`.

elParaguayo:
> A couple other comments on the PR:
> 
> * It might be cleaner to merge the `expose` and `expose_add` decorators (i.e. with the command aliases being a kwarg to expose).

This is now done.

> * Do we need to rename `Layout.add` to `Layout.add_client`? This will cause trouble with anybody who defines any custom layout logic.

Let me check again.

> * Do we need to rename `Qtile.screens` and `Qtile.groups`? This will similarly cause a lot of problems for almost everybody using a custom function in their config!

Yes, I think so. Exposing commands means they have to be called so people either need to change `qtile.screens` to `qtile.screens()` or `qtile._screens`. However, we do need to split the attribute name from the method name.

libqtile/core/manager.py
m-col: > I would assume this is only really relevant for tests.

I agree and hope so. I would guess that more people would be accessing `qtile.screens` in custom code than would be doing `qtile cmd-obj -o cmd -f screens` in some scripts.

What would you think about naming the commands `get_screens` etc?

elParaguayo: Yeah - this is probably a better approach. Can do the same for groups.

elParaguayo: Have pushed an update for that.

docs/manual/commands/index.rst
```diff
@@ -178,10 +178,11 @@ which will lead to the way the commands are dispatched.
 All of the configured objects setup by Qtile are ``CommandObject`` subclasses.
 These objects are so named because we can issue commands against them using the
 command scripting API.  Looking through the code, the commands that are exposed
-are commands named ``cmd_*``.  When writing custom layouts, widgets, or any
-other object, you can add your own custom ``cmd_`` functions and they will be
-callable using the standard command infrastructure.  An available command can
-be extracted by calling ``.command()`` with the name of the command.
+are commands are decorated with the ``@CommandObject.expose()`` decorator.
```
m-col, line 181: "are commands are" -> "are commands that are"?

libqtile/command/base.py
```diff
@@ -55,11 +57,114 @@ class CommandException(Exception):
...
+    inheriting from CommandObject will need to import the module)
     A CommandObject should also implement `._items()` and `._select()` methods
     (c.f. docstring for `.items()` and `.select()`).
     """
 
+    def __new__(cls, *args, **kwargs):
+
+        # Check which level of command object has been parsed
+        # This test ensures inherited classes don't stop additional
+        # methods from being exposed
```
m-col, line 71: Please could you reword this for clarity? I'm getting a bit confused at "inherited classes". Are you referring to parent classes?

libqtile/command/base.py
```diff
@@ -55,11 +57,114 @@ class CommandException(Exception):
...
+            msg = (
+                f"The use of the 'cmd_' prefix to expose commands via IPC "
+                f"is deprecated. Methods should use the "
+                f"@CommandObject.expose() decorator instead. "
+                f"Please update: {names}"
+            )
+            logger.warning(f"Deprecation Warning: {msg}")
+
+        # Record the object as being parsed.
+        cls._command_object = cls.__name__
```
m-col, line 125: Does this make the parsing only happen once per class definition per Qtile instance?

libqtile/command/base.py
```diff
@@ -55,11 +57,114 @@ class CommandException(Exception):
...
+
+    def expose(  # type: ignore
+        name: Optional[Union[str, List[str]]] = None  # noqa: N805
+    ) -> Callable:
+        """
+        Decorator to expose commands to IPC.
+
+        The exposed command will have the name of the defined method.
+
+        Methods can also be exposed via multuiple names by passing the names of
```
m-col, line 140: "multuiple" typo

m-col:
The big question: what migrations scripts are necessary, and, are you sure? I suspect there will be a few. The s/cmd_/cmd is one thing, but also any command renames (`screens` -> `get_screens`), as well as changes to make way for commands (e.g. the `opacity` property is no longer available AFAICT, so needs to be migrated to the command).

libqtile/command/base.py
elParaguayo: I'll add some additional text.

libqtile/command/base.py
elParaguayo: That's the idea, yes. So, once we've parsed, say, `widget.TextBox` any other instantiations of the same class will not be parsed. But, if we create a widget that subclasses `TextBox` it will be parsed.

elParaguayo:
> The big question: what migrations scripts are necessary, and, are you sure? I suspect there will be a few. The s/cmd_/cmd is one thing, but also any command renames (`screens` -> `get_screens`), as well as changes to make way for commands (e.g. the `opacity` property is no longer available AFAICT, so needs to be migrated to the command).

Good question. Let me break out a list of changes so we can see it clearly. `cmd_` naming will still work, the user will just get deprecation warnings but we can include as a migrate script.

However, this migration will be ugly!

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+
+        The exposed command will have the name of the defined method.
+
+        Methods can also be exposed via multiple names by passing the names of
+        the additional method.
+
+        e.g. if a layout wants "up" and "previous" to call the
+        same method:
+
+        @CommandObject.expose("previous")
```
flacjacket, line 150: If people accidentally use `@expose` rather than `@expose()`, this may cause some odd errors. It should be possible with some introspection to determine if we are in the former case. We could allow that, and just adjust the logic to set the right values on the function, or we should try to raise an error.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+
+        def wrapper(func: Callable):
+            setattr(func, "_cmd", True)
+            if name is not None:
+                if not hasattr(func, "_mapping"):
+                    setattr(func, "_mapping", list())
+                if isinstance(name, list):
+                    func._mapping += name  # type:ignore
+                elif isinstance(name, str):
+                    func._mapping.append(name)  # type:ignore
```
flacjacket, line 167: An `else` here with an error would probably be helpful.

flacjacket (commented):
This is very neat to see. Some thoughts and questions in the comments.

docs/sphinx_qtile.py
```diff
@@ -141,7 +142,10 @@ def make_rst(self):
         }
         if context['commandable']:
             context['commands'] = [
-                attr for attr in dir(obj) if attr.startswith('cmd_')
+                # Command methods have the "_cmd" attribute so we check for this
+                # However, some modules are Mocked so we need to exclude them
+                attr.__name__ for _, attr in inspect.getmembers(obj)
+                if getattr(attr, "_cmd", False) and not isinstance(attr, MagicMock)
```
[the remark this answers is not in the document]
elParaguayo: Um... yes... :)

libqtile/backend/base.py
```diff
@@ -214,13 +216,9 @@ def _select(self, name, sel):
         return None
 
     @abstractmethod
+    @CommandObject.expose()
     def info(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
elParaguayo: We're working on the basis that if an inherited class exposes a method and a child overrides that method, the method is still exposed whether the child decorates the method or not. This should stop objects breaking if users forget to decorate methods.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+
+        # Check which level of command object has been parsed
+        # This test ensures inherited classes don't stop additional
+        # methods from being exposed.
+        # For example, if widget.TextBox has already been parsed, a subsequent
+        # call to initialise a new TextBox will return here. However, if a user
+        # subclasses TextBox for a new widget then that new widget will still
+        # be parsed here to check for new commands.
+        if getattr(cls, "_command_object", "") == cls.__name__:
+            return super(CommandObject, cls).__new__(cls)
```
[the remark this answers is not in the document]
elParaguayo: Let me try again. I thought I tried it but may be wrong.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+    def __new__(cls, *args, **kwargs):
+
+        # Check which level of command object has been parsed
+        # This test ensures inherited classes don't stop additional
+        # methods from being exposed.
+        # For example, if widget.TextBox has already been parsed, a subsequent
+        # call to initialise a new TextBox will return here. However, if a user
+        # subclasses TextBox for a new widget then that new widget will still
+        # be parsed here to check for new commands.
+        if getattr(cls, "_command_object", "") == cls.__name__:
```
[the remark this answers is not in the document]
elParaguayo: Your understanding of metaclasses is likely to be far better than mine.

Very simple to test this. If it doesn't fire, I'll remove the code.

libqtile/command/base.py
elParaguayo: Nice idea.

elParaguayo:
@flacjacket Thanks for the review comments. Some really good tips in there.

libqtile/backend/wayland/output.py
```diff
@@ -148,15 +148,15 @@ def _on_frame(self, _listener, _data):
                             box, wlr_output.transform, 0, wlr_output.transform_matrix
                         )
                         renderer.render_texture_with_matrix(
-                            window.texture, matrix, window.opacity
+                            window.texture, matrix, window._opacity
```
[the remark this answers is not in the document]
elParaguayo: I'm inclined to change `opacity` back to being a property and changing the bound command to be `set_opacity()`.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+        @CommandObject.expose("previous")
+        def up(self):
+            ...
+
+        `up` will be exposed as `up` and `previous`.
+
+        Multiple names can be passed as a list.
+        """
+
+        def wrapper(func: Callable):
```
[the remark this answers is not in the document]
elParaguayo: Actually, do we need this? The wrapper returns the original function.

libqtile/command/base.py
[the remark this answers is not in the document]
elParaguayo: I added a logging line after this and it was called multiple times (for the layouts) so it looks like the `__new__` method _is_ being invoked each time.

libqtile/command/base.py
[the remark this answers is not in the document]
elParaguayo: I was wrong ;)

libqtile/backend/base.py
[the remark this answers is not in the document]
elParaguayo: This is how we deal with it:
``` python
# looping over the MRO

                # If the command has been exposed, add it to our dictionary
                # If the method name is already in our dictionary then bind the
                # latest definition to that command
                if hasattr(method, "_cmd") or method_name in commands:
                    commands[method_name] = method
```

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+            for method_name in list(c.__dict__.keys()):
+                method = getattr(c, method_name, None)
+
+                if method is None:
+                    continue
+
+                # Ignore methods that were defined by subclasses
+                for b in base_classes:
+                    if method == getattr(b, method_name, None):
+                        continue
```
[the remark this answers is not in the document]
elParaguayo: You're right - I don't think this is helpful.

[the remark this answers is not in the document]
elParaguayo: Agreed.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+        if getattr(cls, "_command_object", "") == cls.__name__:
+            return super(CommandObject, cls).__new__(cls)
+
+        commands = {}
+        cmd_s = set()
+
+        # We need to iterate over the class's inherited classes in reverse order
+        # We reverse the order so the exposed command will always be the latest
+        # definition of the method.
+        for c in reversed(list(cls.__mro__)):
```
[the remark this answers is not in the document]
elParaguayo: I think reverse order is correct because we want the child's methods to be the last ones we parse. If this is an exposed method name then that name is bound to the child's definition, not the base class's.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+        # call to initialise a new TextBox will return here. However, if a user
+        # subclasses TextBox for a new widget then that new widget will still
+        # be parsed here to check for new commands.
+        if getattr(cls, "_command_object", "") == cls.__name__:
+            return super(CommandObject, cls).__new__(cls)
+
+        commands = {}
+        cmd_s = set()
+
+        # We need to iterate over the class's inherited classes in reverse order
```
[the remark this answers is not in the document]
elParaguayo: That's pretty similar to what we've done. We're iterating over `c.__dict__.keys()` rather than `dir(c)` but the rest of your code looks the same. As above, I think we want to keep the reverse MRO loop.

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+                    continue
+
+                # Ignore methods that were defined by subclasses
+                for b in base_classes:
+                    if method == getattr(b, method_name, None):
+                        continue
+
+                # If the command has been exposed, add it to our dictionary
+                # If the method name is already in our dictionary then bind the
+                # latest definition to that command
```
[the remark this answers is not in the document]
elParaguayo: Sorry if I'm missing something but I don't think that's a bug. I think the reverse MRO parsing is appropriate. Let's say a widget defines an `info` method but forgets to add the decorator. As `info` is decorated in the base class, the command is still exposed but, because of the reverse MRO parsing, the method that's bound to the command is the child widget's definition.

elParaguayo:
@m-col: here's a summary of changes

`cmd_method()` --> `method()`
We can do a migrate for this but it's not a breaking change. `cmd_` calls still work but with logger messages.

Window:
`opacity` (property)  -> `_opacity` (attribute): should we change command to `set_opacity()` and migrate?
(X11) `hints` (attribute) -> `_hints`: as above, should we change the command to `get_hints()`?
(X11) `cmd_get_position` called `self.getposition`: could add additional exposure so both names available
Same for `cmd_get_size`/`getsize`

Manager:
`groups()` -> `get_groups()`
`screens()` -> `get_screens()`
Can do migrate for both of these.

Layouts:
`add` -> `add_client`
Still checking why I needed to do this!

libqtile/command/base.py
```diff
@@ -55,11 +57,118 @@ class CommandException(Exception):
...
+        # Record the object as being parsed.
+        cls._command_object = cls.__name__
+
+        # Store list of exposed commands
+        cls._commands = commands
+
+        return super(CommandObject, cls).__new__(cls)
+
+    def expose(  # type: ignore
+        name: Optional[Union[str, List[str]]] = None  # noqa: N805
```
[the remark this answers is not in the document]
elParaguayo: Actually, that doesn't quite work (pytest seems to have issues with that).

In fact, there's no need for this even to be a method of `CommandObject` as it does nothing to the class. We can take it outside of the class and define it as a function.

elParaguayo:
@m-col: Re `add` -> `add_client` - I think this was because two layouts (matrix and stack) had a `cmd_add` method so, changing this to `add` clashed with the existing method.

My thinking was that people may have bindings for `lazy.layout.add()` but were less likely to be accessing layout internals so I decided to preserve the exposed command name. Happy to discuss if you have an alternative idea.

elParaguayo:
Still keen to get this one in so rebased it again so it doesn't get too far behind.

Expecting a good number of test failures!

libqtile/command/base.py
```diff
@@ -39,6 +41,50 @@
     ItemT = Optional[tuple[bool, list[str | int]]]
 
 
+def expose_command(name: Callable | str | list[str] | None = None) -> Callable:
```
ramnes, line 44: I guess you can remove that `name` argument if it's never used. :P

ramnes: Uh, it is actually used, sorry for the noise. I'm going back to my cave!

m-col:
I'm also eager for this one :) let me know if you'd like me to rebase it to take some of the load off you! From what I can tell in the discussion most things are covered and there are migrations for most config-breaking changes. Is there anything that still needed discussing?

elParaguayo:
Awesome. I've been waiting for this one for a while!

I'm happy to rebase it. Once I've done that we can discuss any relevant concerns.

elParaguayo:
@m-col These are key points: https://github.com/qtile/qtile/pull/2727#issuecomment-1001717652

Also need to do CHANGELOG change.

One question for you, in the Wayland backend there's a `Layer` class. Does that need any exposed commands?

m-col:
> One question for you, in the Wayland backend there's a Layer class. Does that need any exposed commands?

Only `bring_to_front` as defined on the base `Static` class.

The changes in the summary all look fine. Just going through the diff itself. Is there anything I should be looking out for?

libqtile/scripts/migrate.py
```diff
@@ -125,6 +125,59 @@ def windowtogroup_groupName_argument(funcname, query):  # noqa: N802
...
+        .rename("get_screens")
+        .select_method("opacity")
+        .rename("set_opacity")
+        .select_method("cmd_opacity")
+        .rename("set_opacity")
+        .select_method("hints")
+        .rename("get_hints")
+        .select_method("cmd_hints")
+        .rename("get_hints")
+    )
```
m-col, line 151: Could we also add these? They weren't commands but were also updated.
```suggestion
        .select_method("getsize")
        .rename("get_size")
        .select_method("getposition")
        .rename("get_position")
    )
```

test/layouts/test_matrix.py
```diff
@@ -95,6 +95,7 @@ def test_matrix_add_remove_columns(manager):
     manager.test_window("three")
     manager.test_window("four")
     manager.test_window("five")
+    print(f"{manager.c.layout.commands()}")
```
m-col, line 98: Left this in from debugging?

test/layouts/test_slice.py
```diff
@@ -150,3 +150,15 @@ def test_command_propagation(manager):
...
+
+@slice_config
+def test_command_propagation_direct_call(manager):
+    manager.test_window("slice")
+    manager.test_window("one")
+    manager.test_window("two")
+    info = manager.c.layout.info()
+    assert info["name"] == "slice"
+    org_height = manager.c.window.info()["height"]
+    manager.c.layout.eval("self.toggle_split()")
```
m-col, line 163: `toggle_split` is exposed as a command by `Stack`. Should we also test with `manager.c.layout.toggle_split()`?

libqtile/widget/mpris2widget.py
```diff
@@ -275,18 +269,31 @@ def _task_callback(self, task: asyncio.Task) -> None:
...
+    def stop(self) -> None:
         """Stop playback."""
         self._player_cmd("Stop")
+
+    @expose_command()
+    def info(self):
+        """What's the current state of the widget?"""
+        d = base._TextBox.info(self)
+        d.update(
+            dict(displaytext=self.displaytext, isplaying=self.is_playing, player=self.player)
```
m-col, line 297: How come `displaytext` was added to this info? `base._TextBox.info(self)` still provides `text` but the test for this widget has been updated to check `displaytext` instead of `text`

libqtile/command/base.py
```diff
@@ -40,6 +42,50 @@
...
+    """
+    Decorator to expose commands to IPC.
```
m-col, line 47: Maybe to clarify it's not just IPC but also the lazy interface?
```suggestion
    Decorator to expose methods to the command interface.
```

libqtile/command/base.py
```diff
@@ -40,6 +42,50 @@
...
+
+    The exposed command will have the name of the defined method.
+
+    Methods can also be exposed via multiple names by passing the names of
+    the additional method.
```
m-col, line 52:
```suggestion
    Methods can also be exposed via multiple names by passing names to
    this decorator.
```

libqtile/command/base.py
```diff
@@ -40,6 +42,50 @@
...
+        setattr(func, "_cmd", True)
+        if name is not None:
+            if not hasattr(func, "_mapping"):
+                setattr(func, "_mapping", list())
+            if isinstance(name, list):
+                func._mapping += name  # type:ignore
+            elif isinstance(name, str):
+                func._mapping.append(name)  # type:ignore
+            else:
+                logger.error("Unexpected value received in command decorator: %s", name)
```
m-col, line 76: I don't think this is necessary - the type signature states that only `str | list[str]` can be passed so this should never be reached
```suggestion
            else:
                func._mapping.append(name)  # type:ignore
```

libqtile/command/base.py
```diff
@@ -40,6 +42,50 @@
...
+        return func
+
+    # If the decorator is added with no parentheses then we should treat it
+    # as if it had been i.e. expose the decorated method
+    if callable(name):
+        func = name
+        name = None
+        return wrapper(func)
```
m-col, line 84: Why not just `return wrapper(name)`?

m-col (commented):
I've reviewed everything except most of the core changes in `libqtile.command`
(big diff!).

We're changing `Qtile.cmd_groups` to `Qtile.get_groups`, and same for screens.
Should we also do it for `Qtile.windows` and `Qtile.internal_windows`? It is
not necessary, but it would keep the names consistent and predictable. As all
these commands just return info, they seem primarily of use for the CLI and
Python interfaces, which fortunately makes it unambiguous that accessing any of
groups/screens/windows/internal_windows commands through these interfaces
should be prepended with 'get_'. We can even check for those client-side, print
a message, and prepend it ourselves before doing IPC. 

While we're still allowing code to _call_ commands with the `cmd_` prefix, are
we also still allowing code to _define_ methods with the `cmd_` prefix? It's
likely some users will have custom widgets or layouts with commands defined.
And these methods would be intended to override parent methods that do not have
the prefix, so when a class is defined we could check for `cmd_` methods and
copy them minus the prefix.

Messages for the changelog:
- The `cmd_` prefix has been dropped from all commands.
- Layouts need to rename some methods:
  - `add` to `add_client`
  - `cmd_next` to `next`
  - `cmd_previous` to `previous`
- Layouts or widgets that redefine the `commands` property need to update the signature.
- Some commands have been renamed (in addition to dropping the 'cmd_' prefix):
  - `hints` -> `get_hints`
  - `groups` -> `get_groups`
  - `screens` -> `get_screens`
- `Window.getsize` has been renamed `Window.get_size` (i.e. merged with the `get_size` command).
- `Window.getposition` has been renamed `Window.get_position` (i.e. merged with the `get_position` command).
- The `StockTicker` widget `"function"` option is being deprecated: rename it to `"func"`.

elParaguayo:
> > One question for you, in the Wayland backend there's a Layer class. Does that need any exposed commands?
> 
> Only `bring_to_front` as defined on the base `Static` class.

What about `kill`?

libqtile/backend/base.py
```diff
@@ -325,21 +325,10 @@ def match(self, match: config.Match) -> bool:
         return match.compare(self)
 
     @abstractmethod
+    @expose_command()
     def focus(self, warp: bool) -> None:
```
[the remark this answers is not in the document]
elParaguayo: should `Static.focus` be exposed?

test/layouts/test_matrix.py
elParaguayo: Maybe...

m-col:
> What about kill?

Interesting - it looks like the base `Static` class doesn't have that command. I guess because static windows aren't really that dynamic? E.g. users wouldn't be starting and killing polybar during regular usage. I guess you're asking because it will decide whether `kill` on `_Window` should get decorated - `Layer` already has the method, and I don't see any issue with exposing it.

Also `cmd_is_visible` looks like wrong for most wayland window classes. Really it should be as:
```diff
diff --git a/libqtile/backend/wayland/window.py b/libqtile/backend/wayland/window.py
index a289f61c..00c8b679 100644
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -497,6 +497,9 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
             return self.group.screen if self.group else None
         return None
 
+    def cmd_is_visible(self) -> bool:
+        return self._mapped
+
     def cmd_focus(self, warp: bool = True) -> None:
         """Focuses the window."""
         self.focus(warp)
diff --git a/libqtile/backend/wayland/xdgwindow.py b/libqtile/backend/wayland/xdgwindow.py
index 42ee528d..e54b5d69 100644
--- a/libqtile/backend/wayland/xdgwindow.py
+++ b/libqtile/backend/wayland/xdgwindow.py
@@ -174,9 +174,6 @@ class XdgWindow(Window[XdgSurface]):
         if not self.mapped:
             self.surface.map_event.emit()
 
-    def cmd_is_visible(self) -> bool:
-        return self._mapped
-
     def kill(self) -> None:
         self.surface.send_close()
 
```

libqtile/scripts/migrate.py
elParaguayo: We can but I don't like these migrate scripts. What if someone has defined their own `getsize` function?

libqtile/backend/base.py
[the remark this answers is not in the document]
m-col: At the minute they can't be focussed (i.e. `qtile.current_window` will never return a static window). They need some work so that they can be focussed (and also made un-static, among other things). But ATM no

test/layouts/test_slice.py
elParaguayo: I'm confused. Why is this a question for `test_slice`?

libqtile/widget/mpris2widget.py
elParaguayo: This was old. Good spot!

libqtile/command/base.py
elParaguayo: What if someone writes a custom widget and doesn't run `mypy` on their code?

It shouldn't happen with our codebase but other users may get this wrong.

libqtile/command/base.py
elParaguayo: if `name` is not None then what happens to `wrapper` (lines 66-77). I had assumed I needed to clear `name` so as not to trigger an error in `wrapper`.

libqtile/scripts/migrate.py
elParaguayo: The `cmd_` ones will still work. Users will just get a warning telling them to update their code.

libqtile/command/base.py
m-col: Yeah, fair enough

libqtile/command/base.py
m-col: Oh, right you are, I had incorrectly thought `wrapper` was using a different `name`. Ignore :)

test/layouts/test_slice.py
m-col: Because line 163 is calling `Slice.toggle_split()`, which falls through `Slice.__getattr__` to `Stack.toggle_split()`

libqtile/scripts/migrate.py
m-col: OK, nvm my suggested ones

test/layouts/test_slice.py
elParaguayo: Ah. OK. We do actually test `manager.c.layout.toggle_split()` in line 151 (we have a test which checks if it works via the command interface and another to check if it works with a direct call).

libqtile/command/base.py
m-col: Sorry for getting hung up on this, I found it a bit difficult to reason with and I think I figured out why :smile: 

Where I've seen this logic in other places (e.g. the Python Cookbook), they do something like this:
```py
    if callable(name):
        return expose_command()(name)
```
This does create a fresh scope of locals so no need to reset `func` and `name`, which I think is a bit clearer to think about. If that's at the top of the method then the internal function doesn't get defined unnecessarily and means when reading the method you can follow the type of `name` and it's clear when you get to `def wrapper` that it isn't a callable.

Happy to go with what you've put already if you prefer that, I'm just putting this out there just in case :)

m-col:
This all LGTM, happy for this to get merged :)

I haven't put much thought into the `libqtlie.command` mechanism yet and I'd like to (if just to fully understand how it works) but as it's already providing the intended functionality then any changes to that won't make any difference to the rest of the changes. Also, if there are any issues that arise due to the large diff (and the long time this PR has been sitting!) I'm sure users (and ourselves) running the git version will pick those up and we can fix them as they come. I'm loving this improvement!

elParaguayo:
Cool. I've actually got a couple of things I need to do tonight so may now merge tomorrow evening instead.

Am expecting plenty of bug reports to come in after this one!

elParaguayo:
Fixed the commit message and this should be good to go.

PR #2729, opened 2021-08-30
x11: Include only regular windows in _NET_CLIENT_LIST

m-col:
In addition to internal windows, dock-like windows and systray icons
should not be included in the _NET_CLIENT_LIST and
_NET_CLIENT_LIST_STACKING root properties as these windows should not be
visible on pagers. I have included `Static` windows here in general,
though this may include "regular" (non-dock) windows too, because they
do not obey the same rules as regular managed windows (for example you
cannot move them to a different virtual desktop).

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -415,8 +415,9 @@ class Core(base.Core):
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
         """
+        # Regular top-level managed windows, i.e. excluding Static, Internal and Systray Icons
         wids = [
-            wid for wid, c in windows_map.items() if not isinstance(c, window.Internal)
+            wid for wid, c in windows_map.items() if isinstance(c, window.Window)
         ]
         self._root.set_property("_NET_CLIENT_LIST", wids)
         # TODO: check stack order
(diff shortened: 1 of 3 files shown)


m-col:
The intent of this PR was originally to exclude systray icons from the client list. However I'm finding some inconsistent or at least difficult-to-track behaviours with the client list, so sticking it to draft for now.

pepper-jelly:
Before this PR **rofi** showed ghost window. An unnamed one with LG3D class.  Should clients list be additionally updated in libqtile/backend/x11/window.py? I tried just this portion.

``` diff
index 22f19bca..c2508943 100644
@@ -416,7 +416,7 @@ class Core(base.Core):
         chrome
         """
         wids = [
-            wid for wid, c in windows_map.items() if not isinstance(c, window.Internal)
+            wid for wid, c in windows_map.items() if isinstance(c, window.Window)
         ]
         self._root.set_property("_NET_CLIENT_LIST", wids)
         # TODO: check stack order
```

m-col:
> Before this PR rofi showed ghost window. An unnamed one with LG3D class.

Does it go away with these changes? This is a window created by Qtile internally but it should not appear in this list.

> Should clients list be additionally updated in libqtile/backend/x11/window.py? 

What part of that file should be updated?

m-col:
Did you mean the other change in the commit that adds this line to that file?:
```
self.qtile.core.update_client_list(self.qtile.windows_map)
```
In which case - this is there so that windows that become static windows (with lazy.static()) are removed from the list, because the functionality provided by pagers only applies to "regular" windows that follow standard window manager actions, and static windows do not.

m-col:
The behaviour appears to work perfectly. I can't reproduce the unreliable behaviour I mentioned so I'm concluding that I messed up my testing that time. This is ready for review.

PR #2731, opened 2021-08-30
Allow help() on InteractiveCommandClient

### LINKED ISSUE #1421: Python help broken on instance of InteractiveCommandClient
```python
from libqtile.ipc import find_sockfile, Client 
from libqtile.command_client import InteractiveCommandClient 
from libqtile.command_interface import IPCCommandInterface                                                                    

client = InteractiveCommandClient(IPCCommandInterface(Client(find_sockfile())))

help(client)
```
```
SelectError                               Traceback (most recent call last)
<ipython-input-4-0055de52face> in <module>
----> 1 help(client)

/usr/lib/python3.7/_sitebuiltins.py in __call__(self, *args, **kwds)
    101     def __call__(self, *args, **kwds):
    102         import pydoc
--> 103         return pydoc.help(*args, **kwds)

/usr/lib/python3.7/pydoc.py in __call__(self, request)
   1891     def __call__(self, request=_GoInteractive):
   1892         if request is not self._GoInteractive:
-> 1893             self.help(request)
   1894         else:
   1895             self.intro()

/usr/lib/python3.7/pydoc.py in help(self, request)
   1950             else: doc(str, 'Help on %s:', output=self._output)
   1951         elif isinstance(request, Helper): self()
-> 1952         else: doc(request, 'Help on %s:', output=self._output)
   1953         self.output.write('\n')
   1954 

/usr/lib/python3.7/pydoc.py in doc(thing, title, forceload, output)
   1670     try:
   1671         if output is None:
-> 1672             pager(render_doc(thing, title, forceload))
   1673         else:
   1674             output.write(render_doc(thing, title, forceload, plaintext))

/usr/lib/python3.7/pydoc.py in render_doc(thing, title, forceload, renderer)
   1645     if renderer is None:
   1646         renderer = text
-> 1647     object, name = resolve(thing, forceload)
   1648     desc = describe(object)
   1649     module = inspect.getmodule(object)

/usr/lib/python3.7/pydoc.py in resolve(thing, forceload)
   1637         return object, thing
   1638     else:
-> 1639         name = getattr(thing, '__name__', None)
   1640         return thing, name if isinstance(name, str) else None
   1641 

/media/data/dev/forks/qtile/libqtile/command_client.py in __getattr__(self, name)
    205             # we are going to resolve a command, check that the command is valid
    206             if not self._command.has_command(self._current_node, name):
--> 207                 raise SelectError("Not valid child or command", name, self._current_node.selectors)
    208             call_object = self._current_node.call(name)
    209             return self.__class__(self._command, current_node=call_object)

SelectError: Not valid child or command
```

Maybe it's not fixable without writing bad looking special cases, so feel free to close as `won't fix`.

elParaguayo:
Overriding `__getattr__` in `InteractiveCommandClient` causes python's built-in `help` command to fail when called on instances of `InteractiveCommandClient`. This is fixed by explicitly handling the attributes queried by `help`.

Closes #1421

### DIFF
--- a/libqtile/command/client.py
+++ b/libqtile/command/client.py
@@ -206,6 +206,13 @@ class InteractiveCommandClient:
             a command graph node (if the name is a valid child) or a command
             graph call (if the name is a valid command).
         """
+
+        # Python's help() command will try to look up __name__ and __origin__ so we
+        # need to handle these explicitly otherwise they'll result in a SelectError
+        # which help() does not expect.
+        if name in ["__name__", "__origin__"]:
+            raise AttributeError
+
         if isinstance(self._current_node, CommandGraphCall):
             raise SelectError("Cannot select children of call", name, self._current_node.selectors)
 


m-col:
Instead of raising an error, could we not just implement one of those attributes so the help message is helpful? It seems like the help text from most classes shows the docstring and methods (and their docstrings?). I don't know how much this stuff needs explicit coding, but I wonder if it might only need a simple solution to allow that information to be printed when using `help`.

elParaguayo:
`help` needs those AttributeErrors - it's what makes the command work.

elParaguayo:
With this PR, running help on an `InteractiveCommandClient` instance will now successfully print the docstring rather than raising a `SelectError`.

m-col:
Gotcha, TIL :) Are there any other places that it would make sense to do this?

elParaguayo:
I don't think so, you'd only call `help` in an interactive python session and none of our other classes are likely to be called directly.

We could make the docstring for the `InteractiveCommandClient` a bit more helpful e.g. include examples, but we can do that in a separate PR.

m-col:
OK great, thanks for this

PR #2734, opened 2021-08-31
Add mcol to sponsors button

m-col:


ramnes:
\o/

m-col:
I thought about that but figured this way it can be a list of multiple, whereas the liberapay item at the top level only accepts one AFAICT

PR #2746, opened 2021-09-03
x11: toggle window button grabbing with focussing

### LINKED ISSUE #2744: Mouse click is not working on browser's auto suggestions
A mouse click is not working in most of the browser suggestions. If i click on the address bar and hit any suggestions it is working fine but when I fill a forum, log in or search for anything in sites then I have to do it with the keyboard's navigation key.

**Bug:**
[https://giphy.com/gifs/C1xq9ILSfXuzsPdRmF](https://giphy.com/gifs/C1xq9ILSfXuzsPdRmF)

**Qtile version :**  _0.18.1.dev0+g8e7ecc0a.d20210719_

**Configuration:** [https://www.dropbox.com/s/6hdq7mugrsthqud/config.py?dl=0](https://www.dropbox.com/s/6hdq7mugrsthqud/config.py?dl=0)

m-col:
This changes how the x11 backend handles grabbing of button events on
windows. Currently, button events are grabbed for button 1 when new
windows are created so that we can listen to these and focus an
unfocussed window upon click. This works for focussing, but the grabbing
of buttons even on the focussed window can cause some problems, such as
that seen in #2744. This new approach changes this so that the button
events are grabbed only when a window is unfocussed, and ungrabs them
when the window gains focus. This follows the approach used by i3, and
also reduces some unnecessary button event processing as we no longer
respond to button events on the focussed window, whose button events we
do not care about.

Fixes #2744

--

I'd like some x11 uses to test this out during daily use for a bit just to make sure it's working just fine, particularly anyone who uses any chromium-based software. I did some general testing and so far it seems to work pretty well without issues.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -764,6 +764,29 @@ class _Window:
             eventmask=self._window_mask
         )
 
+    def _grab_click(self):
+        # Grab button 1 to focus upon click when unfocussed
+        for amask in self.qtile.core._auto_modmasks():
+            self.qtile.core.conn.conn.core.GrabButton(
+                True,
+                self.window.wid,
+                EventMask.ButtonPress,
+                xcffib.xproto.GrabMode.Sync,
+                xcffib.xproto.GrabMode.Async,
+                xcffib.xproto.Atom._None,
+                xcffib.xproto.Atom._None,
+                1,
+                amask,
+            )
+
+    def _ungrab_click(self):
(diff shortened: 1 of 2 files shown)


elParaguayo:
I'm happy to test this. Will add to my fork.

m-col:
Confirmed on IRC to be good and have no issues.

elParaguayo:
For those not on IRC, I did, indeed, confirm that I've seen no issues while testing this!

PR #2747, opened 2021-09-03
x11: fix up focus when focus goes to None

tych0:
We have a bug where sometimes we lose track of focus all together, and
hotkeys don't work, etc.

I haven't been able to track down where this bug comes from, but it is
*always* a bug to focus None, since it breaks our hotkeys. See 7b43c9f8af83
("manager: focus root if nothing is focused") for a more detailed
explanation.

Since it is always a bug, let's listen for focus None events and fix up the
focus in this case.

Per: https://tronche.com/gui/x/xlib/events/input-focus/normal-and-grabbed.html

we'll get a FocusOut event with detail set to NotifyDetail#None. Let's test
for this and fixup the focus as we do in other cases.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -53,7 +53,6 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
-    xcffib.xproto.FocusOutEvent,
     xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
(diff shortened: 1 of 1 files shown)


tych0:
Note that I put a log in this if statement and haven't hit this condition yet, but I also haven't hit the bug yet either. Perhaps someone has a better idea of how to accomplish this :)

m-col:
I've not come across this bug myself, but I think it's a pretty neat solution!

elParaguayo:
Haven't seen any downsides of this one. I'm happy to merge unless anyone has any lingering concerns.

tych0:
Yeah, I've been running with this for the last month and haven't run into the issue again, so hopefully it fixes it? I haven't been good about checking logs to see if it actually triggered, but I was getting the bug fairly regularly before this, so it seems like maybe it does.

PR #2748, opened 2021-09-04
CryptoTicker migrate format reset (fix #2736)

### LINKED ISSUE #2736: CryptoTicker migration is incomplete
I previously had:

```
widget.CryptoTicker(format="BTC: {avg}", **widget_defaults)
```

but it seems avg is not an allowed key any more per:

```
2021-09-01 14:42:17,968 ERROR libqtile generic_poll_text.py:poll():L90 got exception polling widget
  File "/home/tycho/.local/lib/python3.9/site-packages/libqtile/widget/generic_poll_text.py", line 88, in poll
    text = self.parse(body)
  File "/home/tycho/.local/lib/python3.9/site-packages/libqtile/widget/crypto_ticker.py", line 90, in parse
    return self.format.format(**variables)
```

/cc @Graeme22

Graeme22:
I haven't used bowler before so hopefully this is the best way to do this...

Fixes #2736 by eliminating the `format` argument from outdated configs using `BitcoinTicker`, which causes the default formatting to be used, thereby preventing breakage.

The tests contain examples of what this can look like.

### DIFF
--- a/test/test_migrate.py
+++ b/test/test_migrate.py
@@ -174,14 +174,30 @@ def test_crypto():
         from libqtile import bar
         from libqtile.widget import BitcoinTicker
 
-        bar.Bar([BitcoinTicker()], 30)
+        bar.Bar(
+            [
+                BitcoinTicker(crypto='BTC', format='BTC: {avg}'),
+                BitcoinTicker(format='{crypto}: {avg}', font='sans'),
+                BitcoinTicker(),
+                BitcoinTicker(currency='EUR', format='{avg}', foreground='ffffff'),
+            ],
+            30
+        )
     """)
 
     expected = textwrap.dedent("""
         from libqtile import bar
         from libqtile.widget import CryptoTicker
 
-        bar.Bar([CryptoTicker()], 30)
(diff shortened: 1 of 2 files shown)


test/test_migrate.py
```diff
@@ -174,14 +174,30 @@ def test_crypto():
...
 
     expected = textwrap.dedent("""
         from libqtile import bar
         from libqtile.widget import CryptoTicker
 
-        bar.Bar([CryptoTicker()], 30)
+        bar.Bar(
+            [
+                CryptoTicker(crypto='BTC'),
+                CryptoTicker( font='sans'),
```
tych0, line 195: Can we drop this whitespace too or is that hard to do?

Graeme22: I honestly have no idea, like I said haven't used bowler and the documentation is a bit sparse.

tych0: Ok, I think this is reasonable for now; if we decide we want to fix up formatting maybe we should just run flake8 or something at the end. Seems much simpler than wading through bowler, I agree.

PR #2750, opened 2021-09-04
Add StatusNotifier widget

elParaguayo:
Add a widget implementing the StatusNotifierItem specification. Should work on both x11 and Wayland backends.

NB Currently, the widget does not implement context menus for the icons.

This replaces #2601. The only difference is that context menu support has been removed/disabled so as not to include popup toolkit in main codebase. I can host on my `qtile-extras` repo.

NB - I've left in the underlying methods to generate the menus from DBus. However, the widget does not access these and there is not code to turn a `DBusMenuItem` into a visible menu.


libqtile/widget/statusnotifier.py
```diff
@@ -0,0 +1,897 @@
...
+        return (
+            len(self.available_icons) * (self.icon_size + self.padding)
+            + self.padding
+        )
+
+    def _configure(self, qtile, bar):
+        if self.configured:
+            return
+
+        self.host = host
```
m-col, line 833: This seems redundant as `host` can be accessed everywhere already. Is this necessary?

elParaguayo: I think we both know the answer to that is 'no'!

libqtile/widget/statusnotifier.py
```diff
@@ -0,0 +1,897 @@
...
+        return None
+
+    def button_press(self, x, y, button):
+        icon = self.find_icon_at_pos(x, y)
+        if not icon:
+            return
+
+        name = 'Button{0}'.format(button)
+        if name in self.mouse_callbacks:
+            self.mouse_callbacks[name](icon)
```
m-col, line 864: Can we keep the arguments passed to the mouse callbacks consistent with other widgets? The standard is not to have any arguments, and this inconsistency will confused anyone who adds a callback to this widget. Alternatively the `icon` could be saved to `self.clicked_icon`, or like the `GroupBox` widget the `x` value could be saved to `self.click` so that the calbacks can use `self.find_icon_at_pos(x)` if they need to - meaning that it can be avoided if it isn't needed for this click.

elParaguayo: Yes. Makes sense.

m-col:
There is a lot of code here for supporting menus, when menus are ultimately not supported. How much of that needs to be in libqtile? Ideally we'd have a version that is similar to the `Systray` in behaviour. Could the extra dbus menu functionality also be packaged as part of the full bells-and-whistles widget?

elParaguayo:
`Systray` supports context menus because they're provided by the window. Not having them is a step backwards but understandable since it's now on us to implement.

Add for the menu stuff. I'll take another look and see if we can strip it out and find a simple way to hook it back in without significant rewrites.

m-col:
Also - where does the XML originally come from? I guess you didn't type it out :p

elParaguayo:
Can't remember. Some Google search. No, didn't type it but could have also dumped it from an introspection call.

m-col:
Is that introspection call possible during runtime? I guess it's very unlikely, but could that interface change?

elParaguayo:
Yes, it's possible at runtime. I avoided it as this is meant to be a standard (although the code is already catering for non-standard icons)

m-col:
Do you know how I could get that xml myself via introspection? Just interested in how this works - my dbus is lacking.

elParaguayo:
I'm flipping this in to draft as I may make some changes as I'm working on the menu version in qtile-extras in parallel.

elParaguayo:
This should be good now. I've got menus working in my repo with not too much additional code required to hook it into this one.

libqtile/resources/status_notifier/dbusmenu.py
```diff
@@ -0,0 +1,71 @@
+SCHEMA_DBUS_MENU = """
```
m-col, line 1: As this is no longer used in this version, could it also be moved to the other repo?

elParaguayo: Dammit. Yes.

judemille:
Any updates on this? I'm in the process of switching to Wayland and would like to use qtile -- if this won't be done in the next week is there a guide for setting up waybar with qtile?

m-col:
> is there a guide for setting up waybar with qtile?

waybar (and other bars) should work out of the box as you'd expect, so please post an issue if you do come across any issues with doing so!

@elParaguayo can comment on the widget.

judemille:
> waybar (and other bars) should work out of the box as you'd expect, so please post an issue if you do come across any issues with doing so!

My understanding is that won't do things like show workspaces and layouts? Perhaps I'm mistaken.

elParaguayo:
Given that we've carved out the the context menus, this should be done and, ideally, just need testing by volunteers.

Once it's committed, I'll push the context menus to my repo.

judemille:
> Given that we've carved out the the context menus, this should be done and, ideally, just need testing by volunteers.
> 
> Once it's committed, I'll push the context menus to my repo.

Do you have an ETA on those context menus? I don't poke at tray icons too much but I don't want to run into a situation where I'm screwed without the menus.

m-col:
> My understanding is that won't do things like show workspaces and layouts? Perhaps I'm mistaken.

I don't think third party bars would display desktops in any wayland compositor as there isn't (yet?) a standard way to communicate "virtual desktops" as X11 knows them. Interesting point that I hadn't considered. There _is_ a protocol for third party tools to control (i.e. close, focus) windows but that isn't implemented yet in Qtile. This is definitely on my list so it's good to have some reinforcement that there is demand for such tools.

judemille:
> I don't think third party bars would display desktops in any wayland compositor as there isn't (yet?) a standard way to communicate "virtual desktops" as X11 knows them. Interesting point that I hadn't considered. There _is_ a protocol for third party tools to control (i.e. close, focus) windows but that isn't implemented yet in Qtile. This is definitely on my list so it's good to have some reinforcement that there is demand for such tools.

Perhaps something that works similar to swaymsg? I'm not too involved in the development but having a protocol with some of the same abilities as swaymsg or even EWMH would be extremely useful.

m-col:
> Perhaps something that works similar to swaymsg? 

Ah yes - you can use Qtile's IPC for this, if the bar can execute arbitrary commands for things. See: http://docs.qtile.org/en/latest/manual/commands/shell/qtile-cmd.html

> E: Regardless, that isn't really a discussion for this PR lol. Perhaps set up a tracking issue or other method of conversation?

Feel free to open an issue or discussion, or chat on IRC :)

m-col:
This is working great! I remember asking this before but I can't find it, and maybe things have changed, but is it possible to load the data in `SCHEMA_STATUS_NOTIFIER_ITEM` during runtime? I remember that you said it isn't likely to change, so I know that won't be an issue, but it just seems odd to check data like that into source control.

elParaguayo:
Ah, yeah. We were going to replace it with the introspection call. I can do that.

elParaguayo:
> > Given that we've carved out the the context menus, this should be done and, ideally, just need testing by volunteers.
> > Once it's committed, I'll push the context menus to my repo.
> 
> Do you have an ETA on those context menus? I don't poke at tray icons too much but I don't want to run into a situation where I'm screwed without the menus.

@judemille Sorry. Missed this. In short, the code for the context menus is pretty much all done. It was originally part of this PR but we cut it back as it adds too much extra code to the codebase. The additional code and the full StatusNotifier widget will be available from my ['qtile-extras' repo](https://github.com/elParaguayo/qtile-extras).

I'd be very surprised if the menus provide any functionality that's not available from the app itself. The menus should really just be there for convenience.

elParaguayo:
Schema stuff has been removed. Fixed the docs.

NB this PR also adds the widget to the default config but maybe we should think about that a bit.

elParaguayo:
Added changelog entry and removed from `default_config` (we can look at that separately).

This should be good to merge now.

m-col:
Beautiful! Thanks for working on this!

PR #2753, opened 2021-09-04
[very-WIP] Trying to figure out how to set low_background

LamprosPitsillos:
# Not Tested

Trying for the first time ever to contribute , trying to solve this #2668.
I just couldn't find how to pass in arguments, and generally sicking some guidance if that's ok.
ex.
`widgets.Battery(low_background="222222")`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,5 @@ docs/screenshots/take_all.log
 
 # ViM tags
 /tags
+
+config.py
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -318,7 +318,7 @@ class Battery(base.ThreadPoolText):
         ('show_short_text', True, 'Show "Full" or "Empty" rather than formated text'),
         ('low_percentage', 0.10, "Indicates when to use the low_foreground color 0 < x < 1"),
         ('low_foreground', '000000', 'Font color on low battery'),
-        ('low_background', 'FF0000', 'Font color on low battery'),
+        ('low_background', 'FF0000', 'Background color on low battery'),
         ('update_interval', 60, 'Seconds between status updates'),
         ('battery', 0, 'Which battery should be monitored (battery number or name)'),
         ('notify_below', None, 'Send a notification below this battery level.'),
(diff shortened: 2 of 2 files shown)


libqtile/widget/battery.py
```diff
@@ -496,9 +497,14 @@ def update(self) -> None:
...
-            self.draw()
+            self.draw(status)
 
-    def draw(self) -> None:
+    def draw(self,status: BatteryStatus) -> None:
+        if self.layout is not None:
+            if status.state == BatteryState.DISCHARGING and status.percent < self.low_percentage:
+                self.background = self.low_background
+            else:
+                self.layout.colour = self.background
```
elParaguayo, line 507: Why are you setting `self.background` in line 505 but `self.layout.colour` here?

Also, you've not got a line to reset `self.background` to a different value, once it's changed to `low_background` you can't change it back.

m-col:
Hey @LamprosPitsillos and thanks very much for the contribution!

I see that the default value of `low_foreground` has changed. Would it be possible to keep this unchanged? The `low_background` default will likely need to be changed so that the default looks alright. My concern is that users who like the default value and therefore don't specify `low_foreground` explicitly will have their widget color suddenly change.

I suggest giving your updated widget a test drive. For this you can use the script in the git repository at `scripts/xephyr` which will open a new window with Qtile running inside, which is useful for quick testing. For more info see http://docs.qtile.org/en/latest/manual/hacking.html#using-xephyr. Doing so will make it clear if the widget is doing what you expect - and I think it actually is not.

.gitignore
```diff
@@ -57,3 +57,5 @@ docs/screenshots/take_all.log
 
 # ViM tags
 /tags
+
+config.py
```
elParaguayo, line 61: You shouldn't need to edit our `.gitignore` file. You should keep your config file out of the cloned repository.

LamprosPitsillos: I'm sorry , I should have made a new branch and commited there. Am slowly learning through trial and error. I will delete this and try again. Thanks!

PR #2755, opened 2021-09-04
Some Wayland input fixes

m-col:
3 misc commits that fix some input-handling for the Wayland backend:

1.  Consider window under pointer when over its border:  This restores the behaviour where moving the pointer over a window border considers that window under the pointer (rather than needing to enter the main content rectangle).
2. Set seat keyboard upon modifier press: this is needed so that multiple keyboards can be used without issues 
3. Don't send clicks used for mouse bindings to focussed window: currently the keys for keybindings are still send into the window but they should be swallowed.


tych0:
Needs a rebase?

PR #2758, opened 2021-09-05
Floating windows snap to windows/screens borders option

yobleck:
Similar to the KDE and Gnome (and other WM/DEs I assume) feature.
Intended to be enable in the config via:
```
mouse = [
    Drag([mod], "Button1", lazy.window.set_position_floating(border_snapping=True, snap_dist=10),
]
```
disabled by default.

Notice how the mouse keeps moving while the window pauses on the border.
border_snapping

some squashed commit stuff:
Add test for window border snapping
Add changes to docs and changelog
Fix pep8 issues

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@ Qtile xxxxxx, released xxxxxxxxxx:
     * features
         - All layouts will accept a list of colors for border_* options with which
           they will draw multiple borders on the appropriate windows.
+        - lazy.window.set_position_floating() now has the options
+          `border_snapping=bool`(False by default) and `snap_dist=int`(in pixels)
+          which allows floating window's borders to snap/stick to screen and other window's borders
+          similar to features seen in KDE and Gnome.
 
 Qtile 0.18.0, released 2021-07-04:
     !!! Config breakage !!!
--- a/docs/manual/config/lazy.rst
+++ b/docs/manual/config/lazy.rst
@@ -107,6 +107,9 @@ Window functions
       - Put the focused window to/from floating mode
     * - ``lazy.window.toggle_fullscreen()``
       - Put the focused window to/from fullscreen mode
+    * - ``lazy.window.set_position_floating(x, y, border_snapping=False, snap_dist=10)``
+      - Move the focused floating window to x,y with optional border snapping for
+        for easy alignment with other windows and screen edges
 
 ScratchPad DropDown functions
 -----------------------------
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1772,9 +1772,55 @@ class Window(_Window, base.Window):
         """Add dw and dh to size of window"""
         self.tweak_float(dw=dw, dh=dh)
 
-    def cmd_set_position_floating(self, x, y):
-        """Move window to x and y"""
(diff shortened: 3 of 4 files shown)


m-col:
Sorry, looks like this PR flew under the rader a bit. This looks cool. My first thought is that as this is a nicety that it might be better to add it to qtile-examples, which we are now also using for small plugin-type snippets. For this it might need to be made into its own command, rather than an argument for the existing one. While it isn't so big and would benefit from being integrated directly, I'm concerned that it will directly lead to the need for more work, such as snapping window edges during resizing, which from what I can tell follows naturally from this. @elParaguayo what do you think?

elParaguayo:
Tricky one. 

My initial reaction (although I've given it very little thought) is that it may be a bit harder to have this as a separate piece of code because it's changing `Window` code. While I could do code injection to change widgets in `qtile-extras`, that only worked because those objects were being created in the user's config. The problem here is that the creation of `Window` objects is being created by the `Qtile` object which is a bit harder to patch with plugins.

However, I do share the concerns about the impact on managing the code and the impact on other areas which may or may not be immediately foreseeable.

It is a nice piece of work so I'd like to find some solution here!

m-col:
It doesn't necessarily need injection - it can be a `lazy.function` and get the current window from `qtile.current_window`.

elParaguayo:
Nice, if it can be extracted into standalone code (which, now that you said it, seems obvious) then I agree qtile-examples would be a good spot for this.

You could even do `lazy.window.function` to get the window passed as the first argument, rather than qtile, which may make it a bit easier too.

m-col:
> You could even do lazy.window.function to get the window passed as the first argument, rather than qtile, which may make it a bit easier too.

Brilliant - much better. I always forget about `lazy.<path>.function`.

How does that sound @yobleck?

yobleck:
I agree that it would make more sense/require less maintenance to have it in qtile-examples.
Conveniently I already have a config code injection version [here](https://github.com/yobleck/configs/blob/41dc2fa4014a4fd937a7931f3ff82f015a85ea5f/qtile/internal_modifications.py#L45-L104)
which is added to the config in the `@hook.subscribe.client_managed` [here](https://github.com/yobleck/configs/blob/41dc2fa4014a4fd937a7931f3ff82f015a85ea5f/qtile/config.py#L178) (`int_mod.` is just because I have the code in a separate file for readability) 

However the way I wrote my code injection still requires adding options to `lazy.window.set_position_floating()`. Would It make sense to just eliminate that (or move it to the hook?) since the user would be explicitly installing the plugin? And since the code's being injected at the hook would `lazy.window.function` still be necessary?

elParaguayo:
You don't need code injection. You can actually just do it like this: https://gist.github.com/elParaguayo/05dd342326a106ebe45c20d0ff309a47

Basically, `lazy.window.function` calls a function with the window as the first argument. This gives you access to the window object so you can just replace `self` with `window` and get the same result. I've just tested it and it seems to work.

Feel free to tweak as you see fit but would be great if this could be submitted to qtile-examples.

yobleck:
submitted to qtile-examples, so I'll close this PR.

nonetrix:
> submitted to qtile-examples, so I'll close this PR.

Doesn't work for me but thanks

PR #2761, opened 2021-09-05
resources: add wayland session desktop entry

ljmf00:

---

This is needed in order to distributions install the correct session to `/usr/share/wayland-sessions/` to provide Wayland support.


m-col:
Ooh great idea, thank you for this!

PR #2766, opened 2021-09-07
x11: Improve UnmapNotify handling to follow spec

### LINKED ISSUE #1232: Floating window always shows in the desktop whare they are opened first time
Floating window, albert or kcalc, in this case, always open in the desktop where they was opened first.
For example if I open a float in desktop 1 and then switch to desktop 4 and try to reopen that float it doesn't display but if I switch back to desktop 1 I can see that floating window.

m-col:
According to
https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45643490113536,
Qtile should remove the _NET_WM_DESKTOP (and _NET_WM_STATE) property
from windows when they unmap. This lets clients reset this property with
a new value so that when they re-use a window it can be re-managed on a
different desktop.  This only applies to windows unmapped by a client,
and not windows unmapped by Qtile (e.g. when switching groups). Both
actions trigger UnmapNotify events on the root window so currently
cannot be distinguished. This commit disables UnmapNotify events
triggered by internal logic by removing SubstructureNotify events from
the root window's event mask when hiding a window. We can then pass all
windows that emit UnmapNotify events through the current unmanaging
code, and also remove the _NET_WM_DESKTOP property in the process. This
lets clients who re-use windows reset the property and have it respected
when their window gets managed again.

Fixes #1232

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -106,16 +106,15 @@ class Core(base.Core):
                 logger.error("not starting; existing window manager {}".format(existing_wmname))
                 raise ExistingWMException(existing_wmname)
 
-        self._root.set_attribute(
-            eventmask=(
-                EventMask.StructureNotify
-                | EventMask.SubstructureNotify
-                | EventMask.SubstructureRedirect
-                | EventMask.EnterWindow
-                | EventMask.LeaveWindow
-                | EventMask.ButtonPress
-            )
+        self.eventmask = (
+            EventMask.StructureNotify
+            | EventMask.SubstructureNotify
+            | EventMask.SubstructureRedirect
+            | EventMask.EnterWindow
+            | EventMask.LeaveWindow
+            | EventMask.ButtonPress
         )
+        self._root.set_attribute(eventmask=self.eventmask)
 
         self._root.set_property(
             "_NET_SUPPORTED", [self.conn.atoms[x] for x in xcbq.SUPPORTED_ATOMS]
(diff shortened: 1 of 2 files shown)


tych0:
CI is a segfault, which seems probably (?) unrelated to this PR? Have you seen it locally?

tych0:
(FWIW, this LGTM modulo the segfault thing, which maybe we can ignore)

m-col:
I hadn't noticed the segfault! I'm not getting it locally. There are also failures for all tests in `test_check.py` and `test_migrate.py` from code that isn't touch in this PR. Interestingly mypy hasn't updated since July so I have no idea how the test results could have changed.

elParaguayo:
I've been seeing inconsistent errors with those tests too, related to two lines in `backend/x11/drawer.py`. Same for you?

m-col:
I've been seeing it recently on a few files, but in this case it is:
```
  libqtile/backend/x11/core.py:707: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:707: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:707: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:708: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:708: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:708: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:710: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:710: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:710: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:711: error: Item "Window" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:711: error: Item "Internal" of "Union[Window, Internal, Static]" has no attribute "window"
  libqtile/backend/x11/core.py:711: error: Item "Static" of "Union[Window, Internal, Static]" has no attribute "window"
```

m-col:
Segfault seems to be unrelated, it hasn't appeared again since. The mypy error was related, just needed an assert, so all good.

elParaguayo:
I'm getting errors which I assume result from this:
```
2021-09-12 16:13:11,011 ERROR libqtile core.py:_xpoll():L336 Got an exception in poll loop
  File "/usr/lib/python3.9/site-packages/libqtile/backend/x11/core.py", line 304, in _xpoll
    ret = target(event)
  File "/usr/lib/python3.9/site-packages/libqtile/backend/x11/core.py", line 707, in handle_UnmapNotify
    assert isinstance(win, (window.Window, window.Static))
AssertionError
```

m-col:
Oh, what? Am I being an idiot? That assert should always be True as far as I can tell

m-col:
Oh, right, this happens when `win` is `None`: https://github.com/qtile/qtile/pull/2790

PR #2767, opened 2021-09-07
widget: net: add support for 'total' bandwidth value

ljmf00:

---

Some users (like me) would like to have, instead of up and down values separately, sum it up altogether. This PR adds 'total' value for that purpose.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Qtile x.x.x, released xxxx-xx-xx:
+    * features
+        - Add `total` bandwidth format value to the Net widget.
+
 Qtile 0.18.1, released 2021-09-16:
     * features
         - All layouts will accept a list of colors for border_* options with which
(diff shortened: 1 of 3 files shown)


elParaguayo:
This looks good to me.

Can you add something to the `defaults` section to show that the `total` key is also available for the `format` string:
https://github.com/qtile/qtile/blob/ea7216d9f4bf7dcce111c71d1a254bee7d8cd4ef/libqtile/widget/net.py#L40-L41

Don't change the default value though.

ljmf00:
> This looks good to me.
> 
> Can you add something to the `defaults` section to show that the `total` key is also available for the `format` string:
> https://github.com/qtile/qtile/blob/ea7216d9f4bf7dcce111c71d1a254bee7d8cd4ef/libqtile/widget/net.py#L40-L41
> 
> Don't change the default value though.

I added `/total` to the `format` defaults description :)

m-col (approved):
Looks great, thanks! Maybe add a line to the changelog so that others might discover your addition? (You may need to add a new section for the next release in the changelog file)

ljmf00:
> Looks great, thanks! Maybe add a line to the changelog so that others might discover your addition? (You may need to add a new section for the next release in the changelog file)

Added.

m-col:
Awesome, thanks for this :smile:

PR #2768, opened 2021-09-07
widget: groupbox: add support for show_screen_only flag

ljmf00:

---

This PR enables support for `show_screen_only` flag that allows you to filter groups per screen.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -230,6 +230,11 @@ class GroupBox(_GroupBase):
             "Hide groups that have no windows and that are not displayed on any screen."
         ),
         (
+            "show_screen_only",
+            False,
+            "Show groups that are currenly only available to the attached screen"
+        ),
+        (
             "spacing",
             None,
             "Spacing between groups"
@@ -347,6 +352,10 @@ class GroupBox(_GroupBase):
 
         offset = self.margin_x
         for i, g in enumerate(self.groups):
+
+            if self.show_screen_only and self.qtile.current_screen != self.bar.screen:
+                continue
+
             to_highlight = False
             is_block = (self.highlight_method == 'block')
             is_line = (self.highlight_method == 'line')


ljmf00:
Eventhough this is fairly trivial, I didn't tested it, as I don't know how to run `./scripts/xephyr` with two fake screens.

yobleck:
> Even though this is fairly trivial, I didn't test it, as I don't know how to run `./scripts/xephyr` with two fake screens.

Yeah, xephyr multiscreen support is not fun to deal with. I just use one xephyr screen and a qtile config with two [fake screens](https://docs.qtile.org/en/latest/manual/config/screens.html#fake-screens)

libqtile/widget/groupbox.py
```diff
@@ -347,6 +352,10 @@ def draw(self):
 
         offset = self.margin_x
         for i, g in enumerate(self.groups):
+
+            if self.show_screen_only and self.qtile.current_screen != self.bar.screen:
```
elParaguayo, line 356: I'm confused. This test doesn't look up anything to do with the groups that you're looping over so it's either true or false for all groups.

What are you actually trying to do?

ljmf00:
Well, I found out that this is way harder to implement with the current architecture since Qtile doesn't save the previous screen state. This can be done externally but it is not an ideal solution. The idea would be to mimic the same behaviour as i3 or bspwm do.

m-col:
I think the difference between Qtile and those others is that in our model the groups that are not currently visible are not actually associated with a screen, so I don't think this makes much sense conceptually.

What you could do (and perhaps this is what you meant with "this can be done externally") is set the `visible_groups` property of the groupboxes and then use hooks to update them when a group is focussed, popping that group out of whichever groupbox it was on before and adding it to the groupbox on the screen that the group is now on. Alternatively if you want to keep groups 'stuck' to a given screen then you'd need to write key bindings that move between a screen's set of groups only - in this case `visible_groups` doesn't need to change.

I suppose this is an example of where qtile provides more flexibility with the cost of more work for configuration.

ljmf00:
> I think the difference between Qtile and those others is that in our model the groups that are not currently visible are not actually associated with a screen, so I don't think this makes much sense conceptually.
> 
> What you could do (and perhaps this is what you meant with "this can be done externally") is set the `visible_groups` property of the groupboxes and then use hooks to update them when a group is focussed, popping that group out of whichever groupbox it was on before and adding it to the groupbox on the screen that the group is now on. Alternatively if you want to keep groups 'stuck' to a given screen then you'd need to write key bindings that move between a screen's set of groups only - in this case `visible_groups` doesn't need to change.
> 
> I suppose this is an example of where qtile provides more flexibility with the cost of more work for configuration.

Yes, that's it. I thought that information was saved by Qtile, but I was wrong. Maybe this can be optionally added to cover a great chunk of the userbase coming from other window managers. I didn't find any configuration with that and for some people that is a deal-breaker when choosing Qtile, IMO. I'm slowly changing my i3wm config to Qtile, but some things I find quite different that some extra work needs to be done on either my side or the upstream.

I would suggest, if this is willing to be accepted, to add an issue to implement this as a feature or, at least, add some guidance in that regard. What do you think?

m-col:
I hear your woes, and I agree this appears to be a very common desire. Indeed, this is how my own configuration works :smile: However, it would require substantial changes to the core codebase to represent groups this way, and I don't think it could be done without impinging upon the current representation or maintaining two independent representations. There are a number of things that would need to be considered, such as different ways to change groups (from `Qtile`, `Window`, `Screen`, `GroupBox`), same for focussing screens, binding keys these things etc, and I wouldn't want to assume everybody wants it exactly the same way (which is the advantage of having the most flexible representation in the core).

I do think it would be really good to have some examples on the config side for users such as yourself to start from to then define the behaviour that they want. This would make an excellent addition to the `qtile-examples` repository as a place to direct users who ask for this kind of behaviour. If you would be happy to put something together, I'd be happy to help out with it! I do have a couple of gists that might make a good starting point: https://gist.github.com/m-col/bb5630f17ed2bddef82bc78a1926c4d5 and https://gist.github.com/m-col/4f96e9c1b417574a68c6787e1f825857

PR #2784, opened 2021-09-11
X11: Fix double border by using the depth of the client

### LINKED ISSUE #2781: Multiple borders not drawn on Firefox in X
# Issue description

Using the x11 backend, multiple borders fail to get drawn on Firefox windows, instead keeping whatever borders they had before or no borders.

# Qtile version

Current master add6e136ce5d7f046d560979bd58a7f3b085e4d2

# Stack traces

`MatchError`s get logged when the border drawing is attempted e.g. when moving focus to a FF window. This makes me suspect it's a mismatch between window and border depth/visual or one of those other X things (haven't looked deeper).

```
2021-09-08 23:02:21,385 ERROR libqtile core.py:_xpoll():L330 Got an exception in poll loop
  File "/home/mcol/git/qtile/libqtile/backend/x11/core.py", line 284, in _xpoll
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 571, in wrapper
    return f(*args)
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 612, in poll_for_event
    return self.hoist_event(e)
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 686, in hoist_event
    return self._process_error(ffi.cast("xcb_generic_error_t *", e))
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 649, in _process_error
    raise error(buf)
xcffib.xproto.MatchError
2021-09-08 23:02:21,402 ERROR libqtile core.py:_xpoll():L330 Got an exception in poll loop
  File "/home/mcol/git/qtile/libqtile/backend/x11/core.py", line 284, in _xpoll
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 571, in wrapper
    return f(*args)
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 612, in poll_for_event
    return self.hoist_event(e)
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 686, in hoist_event
    return self._process_error(ffi.cast("xcb_generic_error_t *", e))
  File "/usr/lib/python3.9/site-packages/xcffib/__init__.py", line 649, in _process_error
    raise error(buf)
xcffib.xproto.MatchError
```

# Configuration

Default config, just open up firefox and either switch the `Column`s column to stack mode, or just rename the default `border_focus_stack` option to `border_focus`.

jwijenbergh:
This fixes #2781 by not using the root depth for the borders but instead query the client for the specified depth. I already came across this very issue in chwb2 from wmutils and fixed it like this. See: https://github.com/wmutils/opt/pull/47

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -443,11 +443,12 @@ class XWindow:
         core = self.conn.conn.core
         outer_w = width + borderwidth * 2
         outer_h = height + borderwidth * 2
+        depth = self.get_geometry().depth
 
         with PixmapID(self.conn.conn) as pixmap:
             with GContextID(self.conn.conn) as gc:
                 core.CreatePixmap(
-                    self.conn.default_screen.root_depth, pixmap, self.wid, outer_w, outer_h
+                    depth, pixmap, self.wid, outer_w, outer_h
                 )
                 core.CreateGC(gc, pixmap, 0, None)
                 borders = len(colors)
@@ -464,15 +465,15 @@ class XWindow:
                     )
                     core.PolyFillRectangle(pixmap, gc, 1, [rect])
                     coord += borderwidths[i]
-                self._set_borderpixmap(pixmap, gc, borderwidth, width, height)
+                self._set_borderpixmap(depth, pixmap, gc, borderwidth, width, height)
 
-    def _set_borderpixmap(self, pixmap, gc, borderwidth, width, height):
+    def _set_borderpixmap(self, depth, pixmap, gc, borderwidth, width, height):
         core = self.conn.conn.core
         outer_w = width + borderwidth * 2
         outer_h = height + borderwidth * 2
         with PixmapID(self.conn.conn) as border:
             core.CreatePixmap(
-                self.conn.default_screen.root_depth, border, self.wid, outer_w, outer_h
+                depth, border, self.wid, outer_w, outer_h
             )
             most_w = outer_w - borderwidth
             most_h = outer_h - borderwidth
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -89,6 +89,7 @@ def _geometry_getter(attr):
             self.y = g.y
             self.width = g.width
             self.height = g.height
+            self.depth = g.depth
         return getattr(self, "_" + attr)
     return get_attr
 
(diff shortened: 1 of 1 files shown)


jwijenbergh:
Oh by the way, maybe I should set the depth [here](https://github.com/qtile/qtile/blob/886bf89df70ce0e45ba6d4a7ed4f069e98554430/libqtile/backend/x11/window.py#L503) and pass it to the paint_borders function? However, what to do when a DrawableError occurs because we're too early? 

It's always a good idea to minimise the queries needed I guess.

m-col:
Cool, nice and simple! Maybe the depth can be saved using the same approach as the other geometry properties, using the geometry getter/setters at the top i.e. `depth = property(fset=_geometry_setter("depth"), fget=_geometry_getter("depth"))`?

elParaguayo:
Looks like I can take this one off my list (thanks @jwijenbergh !).

jwijenbergh:
> Cool, nice and simple! Maybe the depth can be saved using the same approach as the other geometry properties, using the geometry getter/setters at the top i.e. `depth = property(fset=_geometry_setter("depth"), fget=_geometry_getter("depth"))`?

Makes sense,I had this exact thing but then wasn't sure if it was correct because it was too late :sweat_smile: .

> Looks like I can take this one off my list (thanks @jwijenbergh !).

No problem, it's one of those weird cases in X11 that I've seen happen in multiple window managers/tools

libqtile/backend/x11/window.py
```diff
@@ -427,7 +428,7 @@ def query_tree(self):
             parent = XWindow(self.conn, q.parent)
         return root, parent, [XWindow(self.conn, i) for i in q.children]
 
-    def paint_borders(self, colors, borderwidth, width, height):
+    def paint_borders(self, depth, colors, borderwidth, width, height):
```
m-col, line 431: Perhaps instead of passing it, just get it from `self.depth`?

jwijenbergh: I thought about this, but depth is a property for Window, not XWindow: `AttributeError: 'XWindow' object has no attribute 'depth'`. That's why the width and height is also passed right?

m-col: Ah of course - ignore me :)

m-col:
Works perfectly, thank you for this! I'll wait for the tests just in case :)

m-col:
Awesome thanks again!

Issue #2778, opened 2021-09-08
Can't set a keybinding to only use the MOD/SUPER key


lominoss-git:
**Key([mod], "Return", lazy.spawn("rofi -show"))**, this is my current keybinding for launching the Rofi app launcher. But what I would like to be this way: **Key([mod], lazy.spawn("rofi -show"))**, but unfortunately it doesn't work.
Thanks in advance!

elParaguayo:
OK. Firstly, I assume your other key bindings are using a different `mod` as  you can't use it as a modifier for some commands and call it separately (or, rather, I don't know what happens if you try to do that).

Your issue is that, you want to capture the keypress but, here, it's not being used as a modifier so try `Key([], "Super_L", lazy.spawn("rofi -show"))`

It's untested so let me know if that works.

lominoss-git:
**Man/woman**, I just looooove you. Thank you so much **man/woman**!!

m-col:
Huh, it works, I didn't think it would. Unfortunately it seems that it is triggered right away, meaning that if you have a binding for Super+d then the binding for Super is always triggered before you hit the d. It would be nice if it didn't trigger until released so it knows if it's a modifier for another binding or not.

elParaguayo:
> Huh, it works, I didn't think it would. Unfortunately it seems that it is triggered right away, meaning that if you have a binding for Super+d then the binding for Super is always triggered before you hit the d. It would be nice if it didn't trigger until released so it knows if it's a modifier for another binding or not.

Agreed - one for the "ideas"/"enhancements" list?

m-col:
certainly

ep12:
I have a working prototype for on-release key bindings. Let's see if everything works out and becomes a pr.

TooManyHobbies:
> I have a working prototype for on-release key bindings. Let's see if everything works out and becomes a pr.

I too am looking forward to this feature being merged. It looks like there hasn't been any activity for a few months.  Has any progress been made?

TooManyHobbies:
@ep12 I understand, thank you.  For now I think I'll manually merge your fix into my own branch and try it out.  I'm still on X11 and won't use the mouse drag functionality.  Thank you for what you have don so far, it's appreciated!

PR #2785, opened 2021-09-11
Add support for on-release key bindings.

ep12:
See #2778: When using keys like alt/meta, control, super etc. both as
modifiers for bindings and as bindings, the on-key-down policy
interferes, resulting in unwanted behaviour.

This commit adds the possibility to only trigger actions on release.

+ Configuration is simple: `Key([], "Super_L", ..., on_release=True)`.
+ The backend has to call `Qtile.process_key_event` with two additional
  arguments:
  1. `is_release: bool`: if the event corresponds to the release of a
     key.
  2. `key_as_modifier: int`: a mask representing the key, e.g. control
     would be 0b00000100 = 0x04, shift would be 0b00000001 = 0x01 and so
     on. For most (normal) keys this should be 0x00 and have no effect (XOR).
+ The mapping from key (syms) to masks (in the case of backend.x11) is
  calculated in `x11.core.Core._make_keysym_to_modmask_dict`, hacked
  together and heavily inspired by `xmodmap -pm` and its source code:
    https://github.com/freedesktop/xmodmap/blob/6d5aa481669cdbf2f0cb83bb0b8f142ba94307ae/exec.c#L216
  Remark:  This could also be achieved by chaining
  `keysym_to_keycode | get_modifier | ModMasks | maskname2mask`
  where get_modifier might return None. A cached dictionary is more
  efficient in the long run.

-------------------

Progress:

+ [x] Configuration option is implemented
+ [x] X11 backend changes implemented
+ [ ] wayland implementation (@m-col?)

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -48,8 +48,6 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
-    xcffib.xproto.KeyReleaseEvent,
-    # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
     xcffib.xproto.NoExposureEvent,
     xcffib.xproto.ReparentNotifyEvent,
@@ -162,9 +160,32 @@ class Core(base.Core):
             | xcbq.PointerMotionHintMask
         )
 
+        self._mod_ks_to_mask = self._make_keysym_to_modmask_dict()
+
     @property
     def name(self):
         return "x11"
+
+    def _make_keysym_to_modmask_dict(self) -> dict[int, int]:
+        """Make a dictionary mapping modifier keysyms to modifer masks.
+
+        This essentially gives the same information as `xmodmap -pm`, but
+        with keysyms mapping to modifiers instead of the reverse.
+        """
+        modmap_reply = self.conn.conn.core.GetModifierMapping().reply()
+        modmap_keycodes = modmap_reply.keycodes.raw
+        keycodes_per_modifier = modmap_reply.keycodes_per_modifier
+        n_modifiers = modmap_reply.length
+        assert len(modmap_keycodes) >= n_modifiers * keycodes_per_modifier
+        ks_to_mask: Dict[int, int] = {}
+        for mod_num in range(n_modifiers):  # usually 8
+            for i_keycode in range(keycodes_per_modifier):  # usually 4
+                kc = modmap_keycodes[keycodes_per_modifier * mod_num + i_keycode]
+                if kc == 0x00:
+                    continue
+                ks = self.conn.keycode_to_keysym(kc, 0x00)
+                ks_to_mask[ks] = 1 << mod_num
+        return ks_to_mask
 
     def finalize(self) -> None:
         self.conn.conn.core.DeletePropertyChecked(
@@ -362,6 +383,7 @@ class Core(base.Core):
             "ButtonPress",
             "ButtonRelease",
             "KeyPress",
+            "KeyRelease",
         ]
         if hasattr(event, "window"):
             window = self.qtile.windows_map.get(event.window)
@@ -614,6 +636,19 @@ class Core(base.Core):
 
         keysym = self.conn.code_to_syms[event.detail][0]
         self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+
+    def handle_KeyRelease(self, event) -> None:  # noqa: N802
+        assert self.qtile is not None
+
+        keysym = self.conn.code_to_syms[event.detail][0]
+        # We need to fix the mask if we use a modifier key like Control_R as a
+        # regular key. Otherwise it would be reported as Ctrl+Control_R and thus
+        # be ignored as an "unknown keysym"
+        modifier_as_mask = self._mod_ks_to_mask.get(keysym, 0x00)
+        mask = (event.state & self._valid_mask) ^ modifier_as_mask
+        self.qtile.process_key_event(
+            keysym, mask, is_release=True, key_as_modifier=modifier_as_mask
+        )
 
     def handle_ButtonPress(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -176,8 +176,12 @@ class Core(base.Core):
         modmap_keycodes = modmap_reply.keycodes.raw
         keycodes_per_modifier = modmap_reply.keycodes_per_modifier
         n_modifiers = modmap_reply.length
-        assert len(modmap_keycodes) >= n_modifiers * keycodes_per_modifier
-        ks_to_mask: Dict[int, int] = {}
+        ks_to_mask: dict[int, int] = {}
+
+        if not len(modmap_keycodes) >= n_modifiers * keycodes_per_modifier:
+            logger.error("Couldn't create keysym->modmask mapping dict. Please report.")
+            return ks_to_mask
+
         for mod_num in range(n_modifiers):  # usually 8
             for i_keycode in range(keycodes_per_modifier):  # usually 4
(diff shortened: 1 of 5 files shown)


libqtile/backend/x11/core.py
```diff
@@ -54,8 +54,6 @@
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
     xcffib.xproto.FocusOutEvent,
-    xcffib.xproto.KeyReleaseEvent,
-    # DWM handles this to help "broken focusing windows".
```
ep12, line 58: Is this comment line meant to be deleted with `KeyReleaseEvent` or does it refer to the `MapNotifyEvent` line below? Couldn't really tell from the `git blame`...

m-col: Eh, remove it, lots of our own MapNotify-related code has grown a bit so that we are handling things our own way.

ep12: Okay, thanks.

ep12:
@m-col: The wayland part turns out to be a bit tricky, at least for me. Since you did all the wayland stuff, could you have a look?
I think the starting point would be in https://github.com/qtile/qtile/blob/f4d1ff7cfafe435081189033aa97d7ced53a567f/libqtile/backend/wayland/keyboard.py#L96-L126
but there are some things I couldn't quite figure out:
+ the early `return`s are there for a reason, I guess. It seems we return early if the event has been processed. Now with on-release-only key bindings, maybe we need `<backend core>.qtile.process_key_event` to return a `bool` if the event has triggered any action or not?? (remark: `process_key_event` needs to be called even if a binding is pressed instead of released in the case of `on_release` bindings)
+ I've found that X11 does weird things: When releasing the `Control_L` key, it is reported as `Ctrl+Control_L` (the mask is not `0x00`). How does wayland handle this? This is important for the way the detection in `manager.py::Qtile.process_key_event` actually works. Other attempts were flawed (e.g. check that some time has passed) and were much less reliable. (Just keep the key pressed long enough before releasing it).
+ If wayland translates the events to X11 `(keysym, mods)`, then maybe the `_mod_ks_to_mask` (in `x11/core`) should be moved to the general `Qtile` side? Any ideas?

libqtile/backend/x11/core.py
```diff
@@ -165,10 +163,33 @@ def __init__(self, display_name: str = None) -> None:
...
     def name(self):
         return "x11"
 
+    def _make_keysym_to_modmask_dict(self) -> Dict[int, int]:
+        """Make a dictionary mapping modifier keysyms to modifer masks.
+
+        This essentially gives the same information as `xmodmap -pm`, but
+        with keysyms mapping to modifiers instead of the reverse.
+        """
+        modmap_reply = self.conn.conn.core.GetModifierMapping().reply()
```
m-col, line 178: I think the keysyms are likely to be constant across all users, so would it make sense to just add the dictionary directly, into `xcbq.py`, kinda like the `ModMasks` dictionary?

libqtile/backend/x11/core.py
```diff
@@ -596,6 +618,18 @@ def handle_KeyPress(self, event) -> None:  # noqa: N802
         keysym = self.conn.code_to_syms[event.detail][0]
         self.qtile.process_key_event(keysym, event.state & self._valid_mask)
 
+    def handle_KeyRelease(self, event) -> None:  # noqa: N802
+        assert self.qtile is not None
+
+        keysym = self.conn.code_to_syms[event.detail][0]
+        # We need to fix the mask if we use a modifier key like Control_R as a
+        # regular key. Otherwise it would be reported as Ctrl+Control_R and thus
+        # be ignored as an "unknown keysym"
```
m-col, line 627: Does this not happen with key-down bindings that use modifiers as the main key?

libqtile/core/manager.py
```diff
@@ -352,6 +355,14 @@ def process_key_event(self, keysym: int, mask: int) -> None:
...
             self.grab_chord(key)
         else:
+            if key.on_release != is_release:
+                return
+            if key_as_modifier and key.on_release and last_key_mask & key_as_modifier:
+                # quick fix for on_release keys:
+                # When using a modifier key like Super_L as a regular key,
+                # it should not be triggered when released as the last key of a
+                # sequence of modifiers / keys.
+                return
```
m-col, line 365: This solution feels like it's the business of the backend to ensure that it isn't sending the qtile manager binding events that weren't asked for by the manager. I can see that it might make sense to be here if all backends would require the same solution. I'll start exploring how the wayland backend can implement all of this to see if it makes sense to keep this here or to keep it within the x11 backend.

libqtile/backend/x11/core.py
ep12: Actually, they can be configured using `xmodmap`. I bet the intersection of qtile users and xmodmap users configuring their keymap is not empty... I'll still have a look though, maybe it turns out that `ModMasks` somehow contains the same information although I doubt it. It would be much easier if the dictionary was static....

libqtile/backend/x11/core.py
ep12: I am not sure if I get your point there. I think that `Control_L` alone is reported as `Ctrl+Control_L` both when pressing as well as releasing it.

libqtile/core/manager.py
ep12: One point to keep in mind is that `process_key_...` must be called even if the action is not triggered. X11 sends `KeyRelease` events for every key that is released, e.g. `Ctrl+Meta+X ~> Ctrl+Meta ~> Meta`. If `Meta` itself is bound to an on-release action, it is triggered even if I'm just releasing the modifiers of another binding I just used. To detect that, the `last_key_mask` is necessary. This makes the `process_key_event` thing a non-pure function. Another consequence in this case is that the manager needs both the symbol as well as the modmask to be X11 values.  
Not sure if that helps at all, but I still wanted to mention it.

m-col:
> How does wayland handle this? This is important for the way the detection in manager.py::Qtile.process_key_event actually works

The wayland backend receives events when a key or modifier state is changed i.e. every time a key or modifier is pressed or released, and the state can be read directly from the keyboard object if needed, though i think the event usually has all of the needed information. It'll probably be easier than for the X backend.

> If wayland translates the events to X11 (keysym, mods), then maybe the _mod_ks_to_mask (in x11/core) should be moved to the general Qtile side? Any ideas?

It's using xkbcommon so all of the keys and mods are the same (basically), but that's doesn't have to be the case for other backends that might exist in the future, which is why it's good to keep the code on the Qtile side free from backend implementation details.

I haven't tested this yet, but if you define a binding for a release, do the button-down events for that binding get sent to the focussed client? Should they be? I don't know if it's possible, but can the x11 `Core` grab key press/release events specifically?

ep12:
@m-col Could you please remove the stale label?
Also, do you have an idea how the issue with on-release key bindings and mouse button bindings can be resolved? If mouse events and key events would be handled as one event class this would probably have a more obvious solution...

m-col:
Apologies for this getting forgotten about for some time.

re your discussion about how far to separate backends: while the most 'extreme'
separation would be always working at the level of `str`s, as you say, I would
imagine that even mac OS and windows use `int`s to represent key codes.

re getting release bindings triggered after use in mouse bindings, let me try
to figure out a working solution for this.

I'm also try to put together a solution for wayland.

I've rebased and linted, and added a doc to `Key` for the new option.

m-col:
I made a couple of minor changes, I hope you don't mind. I made it possible to bind a modifier as a key to both key presses and release, so you can do hold-on like bindings. Then I added dropdown show/hide commands as a kind of example, it's pretty cool. I spent some time trying to get a wayland version working but I'm not familiar enough with xkbcommon and how all this works so I haven't been able to. There are a few outstanding issues relating to keyboard handling so at some point(TM) I hope to learn.

Regardless, it would be fine it it only worked for X11. If somebody else is compelled and wants to put in the work to implement the feature, as you have done, they can try to put together a patch. For now, marking it has "X11 only" in the documentation should be sufficient. Better to get this merged, even if there are unexpected issues, than let it get forgotten about again.

I'll monitor the CI and check for any issues that have arisen in the tests, and probably add an entry to the changelog, unless you beat me to it :)

PR #2786, opened 2021-09-12
docs: fix rendering of types

### LINKED ISSUE #2390: Formatting issue in docs
# Issue description

As reported in qtile-dev group, there's a formatting bug in the docs where the parameter name and type are joined into a single word.

See example here: https://docs.qtile.org/en/latest/manual/config/groups.html#group

I'm not familiar with sphinx so I'm not immediately sure how to fix this.

# Qtile version

Latest

EDIT:
On the docs site, in v0.16.1 there was a colon between the parameter and its type. However, testing locally I have that colon showing even with the latest version.

GeoCHiP:
Deleting the space before a colon in type annotations fixes
an issue when parameter name and type are joined after
building docs by sphinx.

Fixes #2390.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -737,7 +737,7 @@ class Core(base.Core):
 
         Parameters
         ==========
-        e : xcb event
+        e: xcb event
             Click event used to determine window to focus
         """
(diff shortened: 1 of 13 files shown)


ramnes:
Thanks!

PR #2787, opened 2021-09-12
Move cmd_change_vt to Wayland Core

m-col:
This is currently at `Qtile.cmd_change_vt` but really only applies to
the wayland `Core`. Now that the core can expose commands itself let's
move it there and remove it from the base `Core` class.


elParaguayo:
Will need to remember this one if it gets merged before my decorators!

PR #2789, opened 2021-09-12
Fix memory leak with widgetbox

elParaguayo:
See: https://github.com/qtile/qtile/issues/2622#issuecomment-912946655

b2b1ed8 changed the way that surfaces are drawn. When the widgetbox hides widgets, it replaces `self.drawer.draw` with a no-op. However, this prevents the drawer from clearing its `RecordingSurface` object. As a result, the `RecordingSurface` will continue to accumulate details of all draws made by the widget while it is hidden.

This PR clears the RecordingSurface while widgets are hidden.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -487,6 +487,7 @@ class Drawer:
 
         self.current_rect = (0, 0, 0, 0)
         self.previous_rect = (-1, -1, -1, -1)
+        self._enabled = True
 
     def finalize(self):
         """Destructor/Clean up resources"""
@@ -529,6 +530,14 @@ class Drawer:
             None,
         )
         self.ctx = self.new_ctx()
+
+    def _check_surface_reset(self):
+        """
+        Checks to see if the widget is not being reflected and
+        then clears RecordingSurface of operations.
+        """
+        if not self.mirrors:
+            self._reset_surface()
 
     @property
     def needs_update(self) -> bool:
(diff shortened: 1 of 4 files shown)


libqtile/widget/widgetbox.py
```diff
@@ -27,8 +28,12 @@
 BoxedWidget = namedtuple("BoxedWidget", ["widget", "draw"])
 
 
-def _no_draw(*args, **kwargs):
-    pass
+def _no_draw(self, *args, **kwargs):
```
m-col, line 31: This looks a lot like a method now! Perhaps we should move it to the base widget class, then no need for the `partial` stuff?

elParaguayo: Agreed. Have updated.

elParaguayo:
Ignore last commit - that will break on wayland...

libqtile/backend/x11/drawer.py
```diff
@@ -130,11 +130,7 @@ def _paint(self):
             ctx.set_source_surface(self.surface, 0, 0)
             ctx.paint()
 
-            # If the widget is not being reflected then clear RecordingSurface of operations
-            # If it is, we need to keep the RecordingSurface contents until the mirrors have
-            # been drawn
```
m-col, line 135: You've removed these 3 lines but not the 3 identical lines from `../wayland/drawer.py`. Maybe also remove them from there?

libqtile/widget/widgetbox.py
```diff
@@ -138,7 +134,7 @@ def _configure(self, qtile, bar):
 
         # # Overwrite the current drawer.draw method with a no-op
         for w in self.widgets:
-            w.widget.drawer.draw = _no_draw
+            w.widget.drawer.draw = w.widget.drawer._check_surface_reset
```
m-col, line 137: Hmm this has turned out a bit uglier than i expected. Maybe we should move this internal-to-Drawer logic within Drawer? i.e. here we could have:
```py
w.widget.draw.disable()  # or pause?
```
Then re-enable/unpause it when it gets shown. It would be a bit neater and more clear i reckon.

Then `Drawer` would, inside these two methods, override `self.draw` how it wants.

elParaguayo: Is there another use case for that? 

I agree it would be neater and maybe that's a good enough reason on its own!

m-col: Not really, I was just triggered by the accessing of a private method of another class. I'm also thinking about long-term readability - the name of `_check_surface_reset` describes what it does but it isn't clear of why it's needed unless you go and find where it is used here.

elParaguayo: Well, I've changed it for you anyway!

m-col:
Great, LGTM!

libqtile/backend/base.py
```diff
@@ -530,6 +531,17 @@ def _reset_surface(self):
         )
         self.ctx = self.new_ctx()
 
+    def _check_surface_reset(self, *args, **kwargs):
+        """
+        Checks to see if the widget is not being reflected and
+        then clears RecordingSurface of operations.
+
+        arg and kwarg handling is needed as WidgetBox replaces `draw` with a
```
elParaguayo, line 539: Don't need this anymore.

m-col (approved):
> 00:00 < mcol> elParaguayo: are you waiting for another review on https://github.com/qtile/qtile/pull/2789?

Nevermind, you got one. LGTM again! :)

PR #2790, opened 2021-09-12
x11: Fix wrong assert in handle_UnmapNotify

m-col:
Follows on from https://github.com/qtile/qtile/pull/2766


elParaguayo:
Thanks. This gets rid of the errors.

m-col:
Great, thanks for checking. Clearly I gotta be more careful when I have to make small edits on otherwise complete PRs!

PR #2801, opened 2021-09-18
Enhancements for scratchpad.

senthilbaboo:
- Use wids instead of pids to save/restore states
- We can use an arbitrary matcher instead of PIDs for scratchpad matching.
- We can hide all scratchpads if needed in a group
- We can set a scratchpad group to single so that only one of the
  scratchpad is displayed at a given time.


senthilbaboo:
This is a continuation of #2109. Sorry I was not able to look into this for a long time. Let me know, if you have comments.

senthilbaboo:
Could someone take a look? Thanks!

m-col:
I wonder if the `single` option is redundant. Assuming a dropdown is being shown with a keybinding such as:
```py
Key([mod], 'm', lazy.group['scratchpad'].dropdown_toggle('tmux'))
```
Then the same functionality can be achieved (I think) with:
```py
Key([mod], 'm', lazy.group['scratchpad'].hide_all(), lazy.group['scratchpad'].dropdown_toggle('tmux'))
```
What do you think?

libqtile/core/state.py
```diff
@@ -42,8 +42,6 @@ def __init__(self, qtile):
         for group in qtile.groups:
             if isinstance(group, ScratchPad):
                 self.scratchpads[group.name] = group.get_state()
-                for dd in group.dropdowns.values():
-                    dd.hide()
```
m-col, line 46: What's the rationale for removing this?

libqtile/scratchpad.py
```diff
@@ -159,6 +160,8 @@ def __init__(self, window, scratchpad_name, ddconfig):
         self.y = ddconfig.y
         self.width = ddconfig.width
         self.height = ddconfig.height
+        # Let's add the window to the scratchpad group.
+        window.togroup(scratchpad_name)
```
m-col, line 164: What effect does this have on behaviour?

libqtile/scratchpad.py
```diff
@@ -289,12 +303,23 @@ def cmd_dropdown_toggle(self, name):
         """
         Toggle visibility of named DropDown.
         """
+        if self._single:
+            for n, d in self.dropdowns.items():
+                if n != name:
+                    d.hide()
```
m-col, line 309: This can be avoided if the dropdown specified by `name` is visible and being hidden, as no other window can be visible.

m-col:
It's really great to see the scratchpads getting some loving, so thanks again for this! Mostly looks good, I've just made a couple comments.

libqtile/core/state.py
senthilbaboo: These lines were hiding all windows that were visible when the state was saved. Setting the state should be enough here so that windows that were hidden are still hidden and windows that were visible are visible.

libqtile/scratchpad.py
senthilbaboo: Ideally this code is called, when the scratchpad is first created or when the state is restored. We explicitly show/hide based on the state in both the cases. So without these lines, it would still work. But, I thought this is good initial state when we initialize a dropdown.

libqtile/scratchpad.py
senthilbaboo: Checking for visibility is a bit convoluted right now. I don't want to get visibility information for WindowVisibiityToggler in DropdownToggler. I don't this, this would affect any performance, as hide() will just be a no-op, if the window is already hidden.

senthilbaboo:
> I wonder if the `single` option is redundant. Assuming a dropdown is being shown with a keybinding such as:
> 
> ```python
> Key([mod], 'm', lazy.group['scratchpad'].dropdown_toggle('tmux'))
> ```
> 
> Then the same functionality can be achieved (I think) with:
> 
> ```python
> Key([mod], 'm', lazy.group['scratchpad'].hide_all(), lazy.group['scratchpad'].dropdown_toggle('tmux'))
> ```
> 
> What do you think?

This won't work as expected, if the window is visible and user wants to hide it with toggle. In the above example, it tmux is already visible, the key press would hide it and make it visible again. XMonad has a feature ExclusiveScratchpad, I just implemented a similar thing here. I use it quiet extensively in my workflow.

libqtile/scratchpad.py
m-col: Without it, does it mean that `window.group` is `None` until the window is shown for the first time? And then from then on, `window.group` is either the scratchpad group or the current visible group? (Just trying to understand). If that's the case then yeah I agree, it makes more sense for it to never be `None` and to get the correct initial state.

m-col (commented):
> This won't work as expected, if the window is visible and user wants to hide it with toggle.

Ah yeah, of course.

jwijenbergh:
Sorry for taking so long to respond to this (I responded on the original pr a while back). This works great! This lets me finally have `emacsclient` (emacs client for daemon mode) as a scratchpad as the PID matching was not working for this. A simple `wm_class` match did the trick :+1:. Would be great if we could get this merged as this can counter lots of issues with scratchpads

```python
DropDown("emacs", "emacsclient -c", match=Match(wm_class='emacs'), x=0.2, y=0.25, width=0.5, height=0.6, opacity=1, on_focus_lost_hide=True),
```

PR #2802, opened 2021-09-18
WL: wlroots.wlr_types.box -> wlroots.util.box

m-col:
This moves the import for pywlroots' `Box`, which has moved with a
deprecation warning.

### DIFF
--- a/libqtile/backend/wayland/output.py
+++ b/libqtile/backend/wayland/output.py
@@ -23,9 +23,10 @@ from __future__ import annotations
 import os
 import typing
 
+from wlroots.util.box import Box
 from wlroots.util.clock import Timespec
 from wlroots.util.region import PixmanRegion32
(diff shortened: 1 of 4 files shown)


elParaguayo:
Will this break anything for users e.g. if they don't update `wlroots`?

flacjacket:
You could bump the `pywlroots` required version in `setup.cfg`. That might help somewhat in signaling the change and at least enforce the requirement at install time.

m-col:
Good idea.

elParaguayo:
If this fixes the deprecation warning, shouldn't the CI tests complete successfully?

flacjacket:
There was a fix in pywlroots that was needed to avoid the deprecation warning coming from valid imports. That has been merged and a new release cut, so the tests should pass now, just re-ran.

PR #2812, opened 2021-09-22
WL: Add support for pointer-constraints-unstable-v1 and relative-pointer-unstable-v1 protocols

m-col:
This PR has two commits to support two pointer-related protocols. These protocols allow clients to 1. request pointer constraints and 2. receive pointer movement events even if the pointer cannot move due to pointer constraints (requested constraints, or monitor edges). These are used e.g. by games to "grab" the pointer, keep it constrained, and always receive events from it.


libqtile/backend/wayland/core.py
```diff
@@ -56,6 +57,10 @@
     OutputConfigurationV1,
     OutputManagerV1,
 )
+from wlroots.wlr_types.pointer_constraints_v1 import (
+    PointerContraintsV1,
```
elParaguayo, line 61: Is this a typo in the `wlroots` module? I would have thought this should be `PointerConstraintsV1`...

m-col: Brilliant, the typo is also in pywlroots

m-col: https://github.com/flacjacket/pywlroots/pull/59

elParaguayo:
Is there a way to test this? Either with a `test_` file or hand on testing?

m-col:
I've been using it to play Morrowind with OpenMW (https://openmw.org/en/) which uses the constraints protocol to keep the pointer within the window's rectangle, and then uses the relative-pointer protocol for movement (because otherwise there is no pointer movement when the pointer is being moved into a constraint). That's an example use-case (and AFAICT the reason these protocols exist) and helped me a lot with getting the logic right. So more testing is obviously welcome but I'm confident it works as expected. As for actual unit tests, I'm not actually sure how to test these! There are a few things like this in the wayland backend that are untested because it's a 'lower level' than the general window management logic of Qtile, but I kinda don't mind because it's usually blaringly obvious if something isn't right because those things form the foundation of how the compositor works.

PR #2815, opened 2021-09-24
Add new hook: screens_reconfigured

elParaguayo:
Users trying to use the `screen_change` hook will have their callback called _before_ `qtile.cmd_reconfigure_screens` is run.
This new hook is called when that process is complete.

Background:
My [WordClock widget](https://qtile-extras.readthedocs.io/en/latest/manual/ref/widgets.html#wordclock) works by painting the clock to root window every 5 minutes. If a new screen is plugged in then `cmd_reconfigure_screens` is run and the screen is refreshed, restoring the original wallpaper. The clock won't repaint the clock until the next 5 minute change.

I tried using the `screen_change` hook but my callback is run before the `cmd_reconfigure_screens` one, presumably because it is added to the list first. I could add a timer but that's hacky.

My solution was to add a new hook that is run once `cmd_reconfigure_screens` is complete. I've tested it with the clock and it works perfectly.

I've not passed any arguments with the hook as `qtile.screens` is already accessible.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -395,6 +395,7 @@ class Core(base.Core, wlrq.HasListeners):
             config.send_failed()
         config.destroy()
         hook.fire("screen_change", None)
+        hook.fire("screens_reconfigured")
 
     def _process_cursor_motion(self, time):
         self.qtile.process_button_motion(self.cursor.x, self.cursor.y)
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -717,6 +717,7 @@ class Core(base.Core):
 
     def handle_ScreenChangeNotify(self, event) -> None:  # noqa: N802
         hook.fire("screen_change", event)
+        hook.fire("screens_reconfigured")
 
     @contextlib.contextmanager
     def disable_unmap_events(self):
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -304,6 +304,19 @@ class Subscribe:
         """
         return self._subscribe("screen_change", func)
 
+    def screens_reconfigured(self, func):
+        """Called when all ``screen_change`` hooks have fired.
+
+        This is primarily useful where you want a callback to be triggered once
+        ``qtile.cmd_reconfigure_screens`` has completed (e.g. if
+        ``reconfigure_screens`` is set to ``True`` in your config).
+
+        **Arguments**
+
+        None
+        """
+        return self._subscribe("screens_reconfigured", func)
+
     def current_screen_change(self, func):
         """Called when the current screen (i.e. the screen with focus) changes
 


m-col:
If you use an async hook, won't it be run when the async tasks are done, i.e. at the end?

elParaguayo:
Use a what now?

m-col:
You can define async hooks like:
```py
@hook.subscribe.something
async def ():
    ...
```
Check out: https://github.com/qtile/qtile/blob/master/libqtile/hook.py#L375

Guess we need to document this better.

elParaguayo:
Saw that but assumed it was to give flexibility of you wanted a callback which needed to leverage asyncio stuff. That's not the issue I had here, it was just sequencing.

I can try your idea and will report back.

m-col:
Also bear in mind that hooked functions are called in the order they are hooked, so if you want something called last then you can just hook it last. In the case of the X11 screen change hook, there is nothing done by qtile after the hook is fired by the core. `cmd_reconfigure_screens` is hooked itself when the config option is set, so as it is your PR would lead to an infinite loop of the hook firing itself. By the sounds of it, you want to hook `cmd_reconfigure_screens` and then hook another function, so instead of letting `Qtile` hook the former via the config option, you can set `reconfigure_screens` to False and set the hooks manually in your config, with your post-`cmd_reconfigure_screens` function hooked last.

elParaguayo:
> Also bear in mind that hooked functions are called in the order they are hooked, so if you want something called last then you can just hook it last. 

That's the point. I can't. `cmd_reconfigure_screens` is hooked automatically if `reconfigure_screens` is set to `True` in your config. If you add a hook in your config, it's added before this one. 

> In the case of the X11 screen change hook, there is nothing done by qtile after the hook is fired by the core. `cmd_reconfigure_screens` is hooked itself when the config option is set, so as it is your PR would lead to an infinite loop of the hook firing itself. 

Not sure I follow. My hook fires after screens are reconfigured but does nothing to reconfigure the screens again. I'm just painting the screen.

> By the sounds of it, you want to hook `cmd_reconfigure_screens` and then hook another function, so instead of letting `Qtile` hook the former via the config option, you can set `reconfigure_screens` to False and set the hooks manually in your config, with your post-`cmd_reconfigure_screens` function hooked last.

Yes. That is one option but it's messy and involves users understanding what `reconfigure_screens` does in their config. Or we have a new hook :wink:

If it's just me with this issue then I can certainly do the approach that you set out above.

m-col:
>  If you add a hook in your config, it's added before this one.

An alternative might be to prepend it to the list... not sure if that just makes things inconsistent though.

> Not sure I follow. My hook fires after screens are reconfigured but does nothing to reconfigure the screens again. I'm just painting the screen.

Ignore me - brain fart. Either way, I think the new hook shouldn't fire within `cmd_reconfigure_screens` because it may not be called - perhaps one wants to reconfigure_screens via the `screen_change` hook but some other way. In fact if I hook something else to `screen_change` at some point post-config then it would be called after the `screens_reconfigured` hook, which is unintuitive. Perhaps the new hook should be fired on the next line right after the `screen_change` hook is fired, in the `Core`s.

> Yes. That is one option but it's messy and involves users understanding what reconfigure_screens does in their config. Or we have a new hook wink

You make a compelling point, consider me swayed.

elParaguayo:
> Ignore me - brain fart. Either way, I think the new hook shouldn't fire within `cmd_reconfigure_screens` because it may not be called - perhaps one wants to reconfigure_screens via the `screen_change` hook but some other way. In fact if I hook something else to `screen_change` at some point post-config then it would be called after the `screens_reconfigured` hook, which is unintuitive. Perhaps the new hook should be fired on the next line right after the `screen_change` hook is fired, in the `Core`s.

Nice. I can work with this. Agreed on the location of the hook firing call (if it works). Can also update docs to explain that this is essentially the same as `screen_change` unless you have `reconfigure_screens` set to `True` in which case this is fired after screens have been reconfigured.

m-col:
nice, lgtm!

m-col:
Cool, thanks!

PR #2820, opened 2021-09-25
WindowCount widget per screen

phil0x2e:
I've got an issue with the WindowCount widget in multi monitor setups. I think it's easiest to show my problem with some screenshots.

In both Screenshots I have one window in group 2 on screen 1 and 3 windows on group 3 on screen 2. The Number in square brackets right of CurrentLayoutIcon widget is the number of of the WindowCount widget.

The first Screenshot is the way it currently works. It displays the number of windows on the currently focused screen on all screens. So all the WindowCount widgets always display the same number. This is not the way, that I at least expected it to work.
Issue

So I've created a small fix:
Issue_fix
Now the windowCount shows the number of windows of the screen the widget lives on.

~~Because some people may want the widget to behave the way it does currently I introduced a new optional flag, which when left out behaves the same way as before. But I think most people would expect it to work as I did.~~

If this is a feature that could be added, I of course could also edit the docs on this branch to reflect the introduction of the new optional parameter.

~~Also suggestions for a better name for the `this_screen` parameter are welcome ;)~~

EDIT:
Removed option, the behaviour from second screenshot is now how it works.

### DIFF
--- a/libqtile/widget/window_count.py
+++ b/libqtile/widget/window_count.py
@@ -25,7 +25,10 @@ from libqtile.widget import base
 
 
 class WindowCount(base._TextBox):
-    """A simple widget to show the number of windows in the current group."""
+    """
+    A simple widget to display the number of windows in the
+    current group of the screen on which the widget is.
+    """
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
         ("font", "sans", "Text font"),
@@ -55,7 +58,7 @@ class WindowCount(base._TextBox):
 
     def _wincount(self, *args):
         try:
-            self._count = len(self.qtile.current_group.windows)
+            self._count = len(self.bar.screen.group.windows)
         except AttributeError:
             self._count = 0
 
@@ -63,12 +66,11 @@ class WindowCount(base._TextBox):
 
     def _win_killed(self, window):
         try:
-            self._count = len(self.qtile.current_group.windows)
+            self._count = len(self.bar.screen.group.windows)
+            if window.group == self.bar.screen.group:
+                self._count -= 1
         except AttributeError:
             self._count = 0
-
-        if self._count and getattr(window, "group", None):
-            self._count -= 1
 
         self.update(self.text_format.format(num=self._count))
 
--- a/test/widgets/test_window_count.py
+++ b/test/widgets/test_window_count.py
@@ -1,5 +1,58 @@
+import pytest
+
 import libqtile
+from libqtile.confreader import Config
 from libqtile.widget import WindowCount
+
+
+class DifferentScreens(Config):
+    groups = [
+        libqtile.config.Group("a"),
+        libqtile.config.Group("b"),
+    ]
+    layouts = [
+        libqtile.layout.Stack(num_stacks=1),
+    ]
+    floating_layout = libqtile.resources.default_config.floating_layout
+    fake_screens = [
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar(
+                [
+                    WindowCount(),
+                ],
+                20
+            ),
+            x=0, y=0, width=300, height=300
+        ),
+        libqtile.config.Screen(
+            top=libqtile.bar.Bar(
+                [
+                    WindowCount(),
+                ],
+                20
+            ),
+            x=0, y=300, width=300, height=300
+        )
+    ]
+    auto_fullscreen = True
+
+
+different_screens = pytest.mark.parametrize("manager", [DifferentScreens], indirect=True)
+
+
+@different_screens
+def test_different_screens(manager):
+    # Put one window on screen 0
+    manager.c.to_screen(0)
+    manager.test_window("one")
+
+    # Put two windows on screen 1
+    manager.c.to_screen(1)
+    manager.test_window("two")
+    manager.test_window("three")
+
+    assert manager.c.screen[0].widget["windowcount"].get() == "1"
+    assert manager.c.screen[1].widget["windowcount"].get() == "2"
 
 
 def test_window_count(manager_nospawn, minimal_conf_noscreen):


elParaguayo:
Thanks.

I wrote the original widget. I think your expectation is correct i.e. it should really show the window count for the screen it's on.

I'd prefer that to be the default option.

libqtile/widget/window_count.py
```diff
@@ -37,10 +37,11 @@ class WindowCount(base._TextBox):
         ("show_zero", False, "Show window count when no windows")
     ]  # type: List[Tuple[str, Any, str]]
 
-    def __init__(self, text=" ", width=bar.CALCULATED, **config):
+    def __init__(self, text=" ", width=bar.CALCULATED, this_screen=False, **config):
```
elParaguayo, line 40: This shouldn't be an option in `__init__`. Any additional parameters should be added to the `defaults` list.

libqtile/widget/window_count.py
```diff
@@ -55,19 +56,28 @@ def _setup_hooks(self):
 
     def _wincount(self, *args):
         try:
-            self._count = len(self.qtile.current_group.windows)
+            if this_screen:
```
elParaguayo, line 59: I think most of the if statements can go and this should just be updated to show the count on the relevant screen,

libqtile/widget/window_count.py
elParaguayo: That said, I don't think we need an option here as we should just fix the behaviour.

phil0x2e: Okay, thanks for the feedback, then I'll remove the option and the if statements.

elParaguayo: Can you also expand the docstring for the widget to explain it shows the window count for the relevant screen too,

Also, ideally, we'd add a test for this on multiple screens. Let me know if you need help with that.

phil0x2e: I improved the docstring.
I tried to implement the tests, but somehow I always get the error `test_window_count[1-x11] - AssertionError: Window never appeared...`
I don't know if I made some mistake in setting up my test environment.
But the other test with the Attribute Errors works and I already fixed an uncaught exception bug there.

elParaguayo: Thanks. I'm out for a bit but will take a look later.

phil0x2e: I gave you permission to push directly on the branch of my fork. I think this will make it much easier, if you want to contribute to it.

libqtile/widget/window_count.py
```diff
@@ -55,21 +58,20 @@ def _setup_hooks(self):
...
         except AttributeError:
             self._count = 0
 
         self.update(self.text_format.format(num=self._count))
 
     def _win_killed(self, window):
         try:
-            self._count = len(self.qtile.current_group.windows)
+            self._count = len(self.bar.screen.group.windows)
+            if window.group == self.bar.screen.group:
```
elParaguayo, line 70: Which one of these gives rise to an `AttributeError`?

phil0x2e: The `if window.group == self.bar.screen.group:` initially gave an error in the `test_attribute_errors` testcase.
But as you can see I already moved it into the try/catch block, so now it works as expected. When there is no bar, screen or group the number of windows is 0 and therefore self._count gets set to 0.

So now you only have to write some new tests for the multi monitor setup.

elParaguayo: Ah, yes. `window` can be `None` so `window.group` raises the exception.

elParaguayo:
Here's the test:
``` diff
diff --git a/test/widgets/test_window_count.py b/test/widgets/test_window_count.py
index fac71136..43ba6bed 100644
--- a/test/widgets/test_window_count.py
+++ b/test/widgets/test_window_count.py
@@ -1,7 +1,60 @@
+import pytest
+
 import libqtile
 from libqtile.widget import WindowCount


+class DifferentScreens(Config):
+    groups = [
+        libqtile.config.Group("a"),
+        libqtile.config.Group("b"),
+    ]
+    layouts = [
+        libqtile.layout.Stack(num_stacks=1),
+    ]
+    floating_layout = libqtile.resources.default_config.floating_layout
+    fake_screens = [
+            top=libqtile.bar.Bar(
+                [
+                    WindowCount(),
+                ],
+                20
+            ),
+            x=0, y=0, width=300, height=300
+        ),
+            top=libqtile.bar.Bar(
+                [
+                    WindowCount(),
+                ],
+                20
+            ),
+            x=0, y=300, width=300, height=300
+        )
+    ]
+    auto_fullscreen = True
+
+
+different_screens = pytest.mark.parametrize("manager", [DifferentScreens], indirect=True)
+
+
+@different_screens
+def test_different_screens(manager):
+    # Put one window on screen 0
+    manager.c.to_screen(0)
+
+    # Put two windows on screen 1
+    manager.c.to_screen(1)
+    manager.test_window("three")
+
+    assert manager.c.screen[0].widget["windowcount"].get() == "1"
+    assert manager.c.screen[1].widget["windowcount"].get() == "2"
+
+
 def test_window_count(manager_nospawn, minimal_conf_noscreen):
     config = minimal_conf_noscreen
     config.screens = [
```

Can you add that, check it works and, if so, add to your PR and squash into a single commit.

phil0x2e:
> Here's the test:
> 
> ```diff
> diff --git a/test/widgets/test_window_count.py b/test/widgets/test_window_count.py
> index fac71136..43ba6bed 100644
> --- a/test/widgets/test_window_count.py
> +++ b/test/widgets/test_window_count.py
> @@ -1,7 +1,60 @@
> +import pytest
> +
>  import libqtile
> +from libqtile.confreader import Config
>  from libqtile.widget import WindowCount
> 
> 
> +class DifferentScreens(Config):
> +    groups = [
> +        libqtile.config.Group("a"),
> +        libqtile.config.Group("b"),
> +    ]
> +    layouts = [
> +        libqtile.layout.Stack(num_stacks=1),
> +    ]
> +    floating_layout = libqtile.resources.default_config.floating_layout
> +    fake_screens = [
> +        libqtile.config.Screen(
> +            top=libqtile.bar.Bar(
> +                [
> +                    WindowCount(),
> +                ],
> +                20
> +            ),
> +            x=0, y=0, width=300, height=300
> +        ),
> +        libqtile.config.Screen(
> +            top=libqtile.bar.Bar(
> +                [
> +                    WindowCount(),
> +                ],
> +                20
> +            ),
> +            x=0, y=300, width=300, height=300
> +        )
> +    ]
> +    auto_fullscreen = True
> +
> +
> +different_screens = pytest.mark.parametrize("manager", [DifferentScreens], indirect=True)
> +
> +
> +@different_screens
> +def test_different_screens(manager):
> +    # Put one window on screen 0
> +    manager.c.to_screen(0)
> +    manager.test_window("one")
> +
> +    # Put two windows on screen 1
> +    manager.c.to_screen(1)
> +    manager.test_window("two")
> +    manager.test_window("three")
> +
> +    assert manager.c.screen[0].widget["windowcount"].get() == "1"
> +    assert manager.c.screen[1].widget["windowcount"].get() == "2"
> +
> +
>  def test_window_count(manager_nospawn, minimal_conf_noscreen):
>      config = minimal_conf_noscreen
>      config.screens = [
> ```
> 
> Can you add that, check it works and, if so, add to your PR and squash into a single commit.

I added the test and even got it to pass now.
I squashed the commits, but because I upstreamed the changes on the master branch before squashing the changes to the master, that have already been made now show up in the diff of this PR.
If that's an issue, I may have to create a new pull request from a new branch or remove the changes manually and rebase again.

elParaguayo:
If you make sure your master is up to date and then rebase your branch onto master, it _should_ get rid of those. You can then force push it again.

phil0x2e:
> If you make sure your master is up to date and then rebase your branch onto master, it _should_ get rid of those. You can then force push it again.

@elParaguayo 
Thank you very much, that did it!

test/widgets/test_window_count.py
```diff
@@ -1,10 +1,66 @@
...
+    assert manager.c.screen[0].widget["windowcount"].get() == "1"
+    assert manager.c.screen[1].widget["windowcount"].get() == "2"
+
+
 def test_window_count(manager_nospawn, minimal_conf_noscreen):
     config = minimal_conf_noscreen
     config.screens = [
+            libqtile.config.Screen(
+                top=libqtile.bar.Bar([WindowCount()], 10)
+            ),
```
elParaguayo, line 63: Don't think we need these 3 lines. Unless I'm missing something.

phil0x2e: Oh, yes I added them when I was trying to write tests and forgot them.
I removed them.

PR #2839, opened 2021-09-30
Fix rendering wallpapers on X11

### LINKED ISSUE #2836: Wallpaper not drawing on all screens after `cmd_screens_reconfigure` triggered by xrandr.
# Issue description

When calling `xrandr` to change screen positions, `cmd_reconfigure_screens` is triggered. This should repaint the wallpapers on all screens as per the config but one of my screens is black.

# Qtile version

Latest master: c9cc86dcff89d1b04ccf13626929538fbd21ae4a

# Stack traces

None

# Configuration

Default config with `wallpaper` and `wallpaper_mode` set.

# Likely cause

My working hypothesis is this:
* When my qtile starts with two screens, they are currently set up side by side
* The root window therefore has dimensions of 3840 x 1080 (i.e. 2x 1920 x 1080)
* If I move one screen above the other then the screen dimensions should now be 1920 x 2160
* The root screen dimensions are unchanged so, when the manager tries to paint a wallpaper on the bottom screen. The co-ordinates (x: 0, y: 1080) are outside of the root window so nothing is drawn.

I tested this by reverting to a side-by-side layout and both screens successfully paint the wallpaper.

# Update

Have identified the issue and am working on a fix!

elParaguayo:
Currently, when the root window dimensions change (e.g. after changing the position of a monitor from side-by-side to above-below), the pixmap for the toot window is not recreated. The result is that images cannot be rendered to the new area.

This commit addresses the issue by checking dimensions before updating the wallpaper.

Closes #2836

@m-col I've left `Painter` as a class for now as there's a slight benefit to retaining the previous state (on X11 at least) as this will prevent unnecessary creations of pixmaps.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -674,14 +674,25 @@ class Connection:
 
 class Painter:
     def __init__(self, display):
-        self.conn = xcffib.connect(display=display)
-        self.setup = self.conn.get_setup()
-        self.screens = [Screen(self, i) for i in self.setup.roots]
-        self.default_screen = self.screens[self.conn.pref_screen]
+        self.display = display
+        self._connect()
         self.conn.core.SetCloseDownMode(xcffib.xproto.CloseDown.RetainPermanent)
         self.atoms = AtomCache(self)
+        self.default_screen = None
+
+    def _connect(self):
+        self.conn = xcffib.connect(display=self.display)
+        self.setup = self.conn.get_setup()
 
     def paint(self, screen, image_path, mode=None):
+        """Paints an image to the screen."""
+
+        # We need to refresh the connection to the X server to check the root
+        # window dimensions have not changed.
+        self._connect()
+        screens = [Screen(self, i) for i in self.setup.roots]
+        default_screen = screens[self.conn.pref_screen]
+
         try:
             with open(image_path, 'rb') as f:
                 image, _ = cairocffi.pixbuf.decode_to_image_surface(f.read())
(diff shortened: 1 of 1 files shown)


m-col:
You know, I only own a laptop and I move between single screen, having an extra screen above it vertically, or having two externals horizontally (laptop screen disabled), and I never came across this issue! Anyway, LGTM :)

> the toot window

nice

elParaguayo:
> the toot window

Oops. 

Seriously though, you haven't seen this issue? If you can't recreate it then we shouldn't merge this.

Setup is config with 2 screens. Both with a `wallpaper`. Start qtile with x11 backend. Screens are side by side by default (I don't have an `xorg.conf` file setting for my screens). Run `xrandr` to place them one above the other. Bottom screen does not have wallpaper.

m-col:
> Screens are side by side by default (I don't have an xorg.conf file setting for my screens).

If you don't have a config file to set this, are you setting it with an `xrandr` call in `.xinitrc`, or just X just default to sticking them horizontally? FWIW I have `randr` firing off from `.xinitrc` via a script, and then use the same script when I (un)plug monitors to set randr setup again. I'll have a play around with this next time I'm by a monitor (tomorrow) and check to see if I can reproduce.

elParaguayo:
Currently (and this will change), I'm not calling `xrandr` from any init file so I assume X is just defaulting to side by side. However, I would assume that the same issue would arise regardless as the issue was that the X connection wasn't refreshed on subsequent calls to the `Painter`.

m-col:
Hmm, yeah that doesn't seem right. Maybe I'm full of it. Hopefully!

m-col:
OK sorry for taking ages, just testing this out now.

> Setup is config with 2 screens. Both with a wallpaper. Start qtile with x11 backend. Screens are side by side by default (I don't have an xorg.conf file setting for my screens). Run xrandr to place them one above the other. Bottom screen does not have wallpaper.

1. 2 `Screen`s, both with wallpapers :heavy_check_mark: 
2. If I remove any randr commands from `.xinitrc` then my screens do default to side-by-side :heavy_check_mark:
3. Use `arandr` to move one below the other triggers Qtile's internal code for `reconfigure_screens` :heavy_check_mark:
4. Bottom screen _does_ have wallpaper :x: 

It seems to work perfectly for me :/

elParaguayo:
Wow. That's so weird. I'll test again. I swear I wasn't making this up!

Issue #2848, opened 2021-10-05
Wrong number of positional arguments in lazy.function()


karthink:
Following the documentation for user-defined functions, 

> lazy.function(func, *args, **kwargs):
> Calls func(qtile, *args, **kwargs). NB. the qtile object is automatically passed as the first argument.

I defined the following:

```python
def window_to_previous_screen(qtile, switch_group=False):
    i = qtile.screens.index(qtile.current_screen)
    if i != 0:
        group = qtile.screens[i - 1].group.name
        qtile.current_window.togroup(group, switch_group)

def window_to_next_screen(qtile, switch_group=False):
    i = qtile.screens.index(qtile.current_screen)
    if i + 1 != len(qtile.screens):
        group = qtile.screens[i + 1].group.name
        qtile.current_window.togroup(group, switch_group)

keys.extend([
    Key([mod,"shift"],"comma",lazy.function(window_to_next_screen, switch_group=False)),
    Key([mod,"shift"],"period",lazy.function(window_to_previous_screen, switch_group=False)),
    Key([mod,"control"],"comma",lazy.function(window_to_next_screen, switch_group=True)),
    Key([mod,"control"],"period",lazy.function(window_to_previous_screen, switch_group=True)),
])
```

When I try to run one of these commands, I get the following error:

```
2021-10-04 22:30:19,848 ERROR libqtile base.py:cmd_function():L211 Exception calling "<function window_to_next_screen at 0x7f6894a5e940>":
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/libqtile/command/base.py", line 208, in cmd_function
    function(self, *args, **kwargs)
  File "/home/karthik/.config/qtile/config.py", line 852, in window_to_next_screen
    qtile.current_window.togroup(group, switch_group)
TypeError: togroup() takes from 1 to 2 positional arguments but 3 were given

```

The documentation for `togroup` says:

```
cmd_togroup(groupName=None, *, switch_group=False)

If groupName is not specified, we assume the current group.
If switch_group is True, also switch to that group.
```

I don't get Python, so I must be doing something wrong?

elParaguayo:
Try changing your line to
``` python
qtile.current_window.togroup(group, switch_group=switch_group)
```
The asterisk is used to force users to use named arguments.

karthink:
Thank you, that fixed it.

So would the function definition be the same if it looked like

```python
def self.togroup(groupName=None, *args, switch_groups=False)
```

[...]

PR #2849, opened 2021-10-05
Add ability to pass parameters to lazy calls

elParaguayo:
Where users create functions in their config which are decorated with `@lazy.function`, there is no easy way to allow arguments to be passed to the decorated function when attaching to a keybinding.

One alternative is to use a closure to wrap the decorated function but this is more complicated for users.

The alternative, which is presented by this PR, is to create a `__call__` method in the `LazyCall` class so that, when a user binds
a key in their config by doing `Key(..., decorated_function(args))` those arguments are added to the `LazyCall` object and passed to the decorated function when the key is pressed.

See https://github.com/qtile/qtile/issues/2848#issuecomment-934093883 for more detail on this.

### DIFF
--- a/docs/manual/config/lazy.rst
+++ b/docs/manual/config/lazy.rst
@@ -139,3 +139,67 @@ User-defined functions
     * - ``lazy.function(func, *args, **kwargs)``
       - Calls ``func(qtile, *args, **kwargs)``. NB. the ``qtile`` object is
         automatically passed as the first argument.
+
+Examples
+--------
+
+``lazy.function`` can also be used as a decorator for functions.
+
+::
+
+    from libqtile.config import Key
+    from libqtile.command import lazy
+
+    @lazy.function
+    def my_function(qtile):
+        ...
+
+    keys = [
+        Key(
+            ["mod1"], "k",
+            my_function
+        )
+    ]
+
+Additionally, you can pass arguments to user-defined function in one of two ways:
+
+1) In-line definition
+
+Arguments can be added to the ``lazy.function`` call.
+
+::
+
+    from libqtile.config import Key
+    from libqtile.command import lazy
+    from libqtile.log_utils import logger
+
+    def multiply(qtile, value, multiplier=10):
+        logger.warning(f"Multiplication results: {value * multiplier}")
+
+    keys = [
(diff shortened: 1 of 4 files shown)


elParaguayo:
Still needs a test and some documentation but posting here to gauge reactions.

m-col (approved):
Very neat, I prefer this over the closure approach. +1 from me.

elParaguayo:
Sure.

The bit that people may find confusing is that their method needs to accept `qtile` as the first object but this also needs to be omitted from the keybinding as it's added when the call is executed i.e. the signatures look different.

elParaguayo:
I wondered that too. There's no need for `lazy` objects to receive `qtile` now given that we can just import it. Also, we removed the same "feature" from widget mouse callbacks.

The relevant section is here:
https://github.com/qtile/qtile/blob/296d4c467c4b6d242e39b95faa22abb1ed3a2cb4/libqtile/command/base.py#L205-L208

The simplest solution would be to stop the function passing `self` but I don't know a) if this would break anything or b) if there's a time when this would be useful (I'd imagine, most times, we can access the relevant object by different means).

m-col:
Perhaps we can continue to pass qtile the current approach but let this approach not take one. You could prepend it to `args` when calling the function via the normal method, then when using this new fancy way it isn't passed?

elParaguayo:
Not sure about that. `qtile` gets passed with either method. It's not added to `args`, it's added _before_ `args` so it's always the first argument. It's also added by `cmd_function` which is not part of `lazy`.

As far as I can tell, there's no good reason for lazy functions to receive `qtile` at all but, in theory, you could call `function` on any other `CommandObject` when it may be useful to get a copy of the object (although `eval` should provide you with access to the object already). I'd prefer to remove the reference to the object being passed but that will break configs and it's probably too hard for a migrate script too.

m-col:
> Yes, it's like a partial. The first argument is pre-defined and you're just adding any additional arguments.

Then is there any benefit to having this implemented?

elParaguayo:
If the change to the lazy page in the docs is ok, this should be good to merge.

m-col:
Looks great, thanks!

PR #2852, opened 2021-10-06
[Wayland] Add input inhibitor to enable screen locking

Graeme22:
~~**This is a draft PR only!** It is dependent on https://github.com/flacjacket/pywlroots/pull/60 being merged into `pywlroots` and will thus also necessitate the latest version of `pywlroots`.~~ That was fast!

Adds support for the Wayland input inhibitor (`wlr_input_inhibit_manager`), allowing programs such as swaylock or waylock to prevent sneaky input sequences that could get around the screenlocker program.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -34,6 +34,7 @@ from wlroots.wlr_types import (
     DataControlManagerV1,
     DataDeviceManager,
     GammaControlManagerV1,
+    InputInhibitManager,
     OutputLayout,
     PrimarySelectionV1DeviceManager,
     ScreencopyManagerV1,
@@ -156,6 +157,7 @@ class Core(base.Core, wlrq.HasListeners):
         XdgOutputManagerV1(self.display, self.output_layout)
         ScreencopyManagerV1(self.display)
         GammaControlManagerV1(self.display)
+        self.input_inhibit_manager = InputInhibitManager(self.display)
         PrimarySelectionV1DeviceManager(self.display)
         self._virtual_keyboard_manager_v1 = VirtualKeyboardManagerV1(self.display)
         self.add_listener(
--- a/libqtile/backend/wayland/keyboard.py
+++ b/libqtile/backend/wayland/keyboard.py
@@ -52,6 +52,7 @@ class Keyboard(HasListeners):
         self.seat = core.seat
         self.keyboard = device.keyboard
         self.grabbed_keys = core.grabbed_keys
+        self._inhibitor = core.input_inhibit_manager
 
         self.keyboard.set_repeat_info(25, 600)
         self.xkb_context = xkb.Context()
@@ -99,7 +100,8 @@ class Keyboard(HasListeners):
             self.qtile = self.core.qtile
             assert self.qtile is not None
 
-        if event.state == KEY_PRESSED:
+        # don't handle shortcuts if screen is locked
+        if event.state == KEY_PRESSED and self._inhibitor.is_inactive():
             # translate libinput keycode -> xkbcommon
             keycode = event.keycode + 8
             layout_index = lib.xkb_state_key_get_layout(
@@ -114,6 +116,7 @@ class Keyboard(HasListeners):
             )
             keysyms = [xkb_keysym[0][i] for i in range(nsyms)]
             mods = self.keyboard.modifier
+
             for keysym in keysyms:
                 if (keysym, mods) in self.grabbed_keys:
                     self.qtile.process_key_event(keysym, mods)
--- a/setup.cfg
+++ b/setup.cfg
@@ -69,7 +69,7 @@ ipython =
   ipykernel
   jupyter_console
 wayland =
-  pywlroots>=0.14.4
+  pywlroots>=0.14.7
   xkbcommon>=0.3
 
 [options.package_data]
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps =
     PyGObject
 # pywayland has to be installed before pywlroots
 commands =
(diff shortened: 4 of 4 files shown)


Graeme22:
> What else is needed for swaylock and similar to work as expected? I assume they use an overlay layer shell surface to draw to, but that other externally-triggered events can cause other windows to steal focus, for example.

Yes, according to swaylock docs the following are required:
- wlr-layer-shell
- wlr-input-inhibitor
- xdg-output
- xdg-shell

The input inhibitor was the only thing missing.

However, I don't think this is quite complete. Looking at the [sway implementation](https://github.com/swaywm/sway/blob/master/sway/input/input-manager.c#L279), it looks like it goes through all of the existing seats and locks them to the current client. Which also has to be undone afterwards.

Graeme22:
> I think we do need the `activate` and `deactivate` events to listen to on the inhibit manager, as this is what is listened to (at least by sway) to know when to set the client as the exclusive client. Then we just have to figure out a neat way to respond to those events and restrict other actions from stealing focus.

I agree on using the activate and deactivate events, the way I was doing was definitely a bit hacky. However I'm not sure if we need to go the client route. From what I can see the current qtile wayland backend has no concept of an active client. Is that something we should add? Or could everything just be done in the core class?

Turning off keyboard shortcuts is of course just one part of the equation; the other is making sure that anything else listening to input other than the screenlocker stops receiving it (though that's certainly more of an edge case).

Thoughts?

m-col:
> From what I can see the current qtile wayland backend has no concept of an active client. Is that something we should add?

Not at the level of the backend as you've seen, but wlroots internally tracks which surface (not client) has keyboard focus and which has pointer focus, and then of course at the `Qtile` level there is the active (toplevel) window. `if qtile.current_window is self` might be sufficient for checking that any requests are coming from the current client, and that inputs are going into only that client.

What would need to be added is just a boolean flag on the core that itself and `Keyboard` can use to know when to stop input processing for windows other than `qtile.current_window` when the flag is `True` (when input is being inhibited by `qtile.current_window`).

That's what I was picturing, but I may have overlooked something in those steps that is important!

m-col:
We'll also need to identify other places where external code can make Qtile do things, such as the command interface (e.g. do we block scripts asking to change focus?) or group matches requesting group changes when new windows open. Then also the foreign-toplevel interface (not yet merged).

arjan-s:
Hi, I would like to inquire what the status is on this PR? I'd love to use Qtile on wayland, but not having a screen locker is a deal breaker for me...

arjan-s:
I just merged this branch and current master and built it, so I'm now testing with screen locker support. This works mostly fine, so thank you for that! As you write above, there isn't very good input blocking yet, as I found after unlocking after some time there were semi-random characters printed to one of my open terminals.

Further more, what's still missing is the idle protocol, to support automatic locking after a specific time (as used by swayidle for instance). Running swayidle currently gives just the error `Display doesn't support idle protocol`.

jwijenbergh:
It might be better to support the new `ext-session-lock` protocol instead?

arjan-s:
Yes, I'm still running my systems with it. It's far from perfect (it doesn't inhibit mouse actions for instance, and incidentally keyboard actions seem to fall through too), see also my previous post. But ever since the idle inhibitor was merged, I'm running swayidle&swaylock successfully enough for my usage.

There is still quite some work to do for it to be an actually secure locking mechanism. See the previous posts by @m-col and @Graeme22 .

BTW, I've had to modify this patch a bit to get it to compile on recent Qtile master:
```
index 5c7589ce..f447e63b 100644
@@ -36,6 +36,7 @@ from wlroots.wlr_types import (
     DataDeviceManager,
     ForeignToplevelManagerV1,
     GammaControlManagerV1,
+    InputInhibitManager,
     OutputLayout,
     PrimarySelectionV1DeviceManager,
     RelativePointerManagerV1,
@@ -172,6 +173,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.idle = Idle(self.display)
         idle_ihibitor_manager = IdleInhibitorManagerV1(self.display)
         self.add_listener(idle_ihibitor_manager.new_inhibitor_event, self._on_new_inhibitor)
+        self.input_inhibit_manager = InputInhibitManager(self.display)
         PrimarySelectionV1DeviceManager(self.display)
         self._virtual_keyboard_manager_v1 = VirtualKeyboardManagerV1(self.display)
         self.add_listener(
diff --git a/libqtile/backend/wayland/inputs.py b/libqtile/backend/wayland/inputs.py
index a5c70156..510af0c2 100644
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -152,6 +152,7 @@ class Keyboard(HasListeners):
         self.keyboard = device.keyboard
         self.keyboard.data = self
         self.grabbed_keys = core.grabbed_keys
+        self._inhibitor = core.input_inhibit_manager
 
         self.keyboard.set_repeat_info(25, 600)
         self.xkb_context = xkb.Context()
@@ -197,7 +198,8 @@ class Keyboard(HasListeners):
 
         self.core.idle.notify_activity(self.seat)
 
-        if event.state == KEY_PRESSED:
+        # don't handle shortcuts if screen is locked
+        if event.state == KEY_PRESSED and self._inhibitor.is_inactive():
             # translate libinput keycode -> xkbcommon
             keycode = event.keycode + 8
             layout_index = lib.xkb_state_key_get_layout(self.keyboard._ptr.xkb_state, keycode)
```

Graeme22:
I'm not currently running it, I think it would still work but I'm not currently able to test it unfortunately.

m-col:
Thank you again for working on this @Graeme22. Closing in favour of https://github.com/qtile/qtile/pull/3663.

PR #2862, opened 2021-10-08
Add ColorsType type hint to specify union of colour or list thereof

m-col:
Just reduce some visual noise when a list of colours is also accepted
somewhere.


m-col:
I think this also revealed some hidden bugs in the base `Drawer`.

PR #2868, opened 2021-10-09
Docs: add Wayland page, use base Window class to list commands, fix main contents formatting issue

m-col:
The title summarises the 3 commits. Please see their individual commit messages for deets.


elParaguayo:
All looks pretty good.

A few things that will conflict with decorators but should be a quick enough fix for me.

I'll try building docs a bit later to have a check.

elParaguayo:
`core.cmd_info` needs a docstring too.

m-col:
Tests seem to be hanging in CI. When I run them locall with `python -m pytest` they work fine, but if I run them with `tox` then they hang, even locally. Hmm, not really sure what I'm looking at here.

m-col:
Finally passing tests!

Issue #2871, opened 2021-10-11
Mpd2 widget throw errors when clicked.


Tokariew:
# Issue description

Mpd2 widget throw bunch of errors after bc84aca1de486894e275dafa933cc0bb94aabe13
and don't reacts to clicks.
`self.mouse_buttons` is empty dict, and probably should have some default values, especially that keys still exist

# Qtile version

0.18.2.dev24+g72799dcc

# Stack traces

```
2021-10-11 06:27:35,844 ERROR libqtile core.py:_xpoll():L328 Got an exception in poll loop
Traceback (most recent call last):
  File "/home/tokariew/.local/lib/python3.10/site-packages/libqtile/backend/x11/core.py", line 296, in _xpoll
    ret = target(event)
  File "/home/tokariew/.local/lib/python3.10/site-packages/libqtile/backend/x11/window.py", line 1108, in handle_ButtonPress
    self.process_button_click(e.event_x, e.event_y, e.detail)
  File "/home/tokariew/.local/lib/python3.10/site-packages/libqtile/bar.py", line 378, in process_button_click
    widget.button_press(
  File "/home/tokariew/.local/lib/python3.10/site-packages/libqtile/widget/mpd2widget.py", line 235, in button_press
    m_name = self.mouse_buttons[button]
KeyError: 1
```

PR #2872, opened 2021-10-11
fix issue #2871

Tokariew:
fix issue in mpd2 widget #2871

### DIFF
--- a/libqtile/widget/mpd2widget.py
+++ b/libqtile/widget/mpd2widget.py
@@ -18,15 +18,13 @@ from libqtile.widget import base
 # TODO: Volume inc/dec support
 keys = {
     # Left mouse button
-    "toggle": 1,
+    1: "toggle",
     # Right mouse button
-    "stop": 3,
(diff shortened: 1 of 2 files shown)


elParaguayo:
These lines:
https://github.com/qtile/qtile/blob/4a85fc667c0403727f848cf19919ee7d0d5cc51a/libqtile/widget/mpd2widget.py#L235-L237
should be changed too. Something like:
``` python
 m_name = self.mouse_buttons.get(button, False)
  
 if self.connected and m_name:
```
The first line will stop us getting `KeyError`s and the second will only try calling the method if we've clicked on a mapped button.

m-col:
The Mpd widget always confused me with what the intent of these keys was! I even have this in my config doing exactly what your change suggests:
```py
mpd2.mouse_buttons = {v:k for k, v in widget.mpd2widget.keys.items()}
```

elParaguayo:
I've added a test for this. If it completes ok, we should merge this.

PR #2873, opened 2021-10-11
Using window list form the layout in tasklist widget

kapilpokhrel:
This uses the list of client's window returned from selected layout's info in the output of tasklist widget. When we go to next or previous windows, it uses a window list stored in layout, so it would be nice and easy to have tasklist show the window in the same order.


m-col:
It might be reasonable to expect all layouts to expose an ordered list 
of windows? For many it'll just be renaming an attribute, one or two 
others might need a property getter (like Slice). Would that work?

PR #2878, opened 2021-10-11
Using window list form the layout in tasklist widget

kapilpokhrel:
This uses the list of client's window returned from selected layout's info in the output of tasklist widget. When we go to next or previous windows, it uses a window list stored in layout, so it would be nice and easy to have tasklist show the window in the same order.

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -496,7 +496,7 @@ class _SimpleLayoutBase(Layout):
     def remove(self, client):
         return self.clients.remove(client)
 
-    def clients_window(self):
+    def get_windows(self):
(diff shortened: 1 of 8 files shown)


kapilpokhrel:
> I'm a bit aware that changes to any layout might break this, so perhaps it would be best to include a test too? It might only need to be simple, say something in `test/layouts/test_common.py` that cycles through all of the layouts and checks that the window orders are preserved (for that layout) in the output of `layout.get_windows`.

There are already tests for testing the order of windows from the client's name and get_windows() is just returning the list of window object instead of window's name. Shouldn't it be enough?

PR #2881, opened 2021-10-11
Centre align logo and buttons

m-col:
A subjective improvement in making the readme look nice. Only if others
agree of course :)

GitHub rst doesn't seem to like rst's actual ways of centreing things so
it seems putting stuff in html is the way to do this.

--

Thoughts? See https://github.com/m-col/qtile/tree/readmecentre for it rendered.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,33 @@
-|logo|
+.. raw:: html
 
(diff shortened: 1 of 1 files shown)


jwijenbergh:
Thanks! Qtile is finally usable :laughing:

PR #2882, opened 2021-10-11
Fix regression that made alpha get ignored in Drawers

### LINKED ISSUE #2879: Bar transparency broken with latest days commits
# Issue description

Just updated qtile today and the latest commit seems to have broken my bars transparency. Everything else is working as expected with picom, it's just my bar that has become solid black  since updating today. It was fine before the update.

It was fine with the 8th commit but the 10th has no bar transparency. Nothing new in my log file either ???

# Qtile version

4d9210224e4eca23c0e461f5ec64392e4bf60e83

# Stack traces

# Configuration

m-col:
4d92102 attempted to fix mypy complaints but in the result was not
passing alpha values to cairocffi when setting a drawer's sourge colour.

Fixes #2879


elParaguayo:
Transparency looks good. Let's see how the mypy tests do...

m-col:
It worked!!

PR #2895, opened 2021-10-16
Allow multiple highlight methods in groupbox widget

CSpencerND:
If implemented, will add a new default parameter and change the `draw()` method of the `GroupBox` widget to allow different highlight methods for groups on different screens.  

By default, groups on unfocused screens will be highlighted with `border`, and the user can define `unfocused_highlight_method` in their `config.py` if they would like to change it, just like the current `highlight_method`.


elParaguayo:
Thanks for this.

Have you successfully managed the border to draw the full height of the widget? I don't think it's a bug in your code but I'm curious if you can make it work as I can't. I'd like to fix that too.

I have some minor comments on your code too but will share those later.

CSpencerND:
The redundant conditionals have been removed. 
I will attempt a fix for the height. TaskList has the same issue that always annoyed me, so if I can do this, I'll mess with that at some point.

elParaguayo:
I was going to make the same comment about `TaskList` too!

elParaguayo:
They're not defined in the widget so don't change them. You can assume the logic in those methods is correct.

elParaguayo:
I don't love the idea of a new parameter. `margin` and `padding` are already in there and should be all that's needed to do any size border/position so I'd rather use them.

elParaguayo:
Agreed. Love it when people want to help. Don't worry about making mistakes, it's how we all learn.

CSpencerND:
I just thought of something. You want me to squash this?

CSpencerND:
Hey guys... I banged my head against the wall, but I finally did it. That was so much more difficult and confusing that it needed to be. It's all cleaned up. Rebased. Squashed. Please tell me I did it correctly 😖

CSpencerND:
@elParaguayo I've been wondering. Should I make a pr for setting the border to the height of the bar, or should I wait until I know what's going on with this one?

elParaguayo:
Let me take another look at this but I'd think we'd put that in a separate PR.

elParaguayo:
Sorry, yes. That would make sense.

elParaguayo:
I haven't fully tested this yet but I've noticed a couple of issues:
1) Using "text" for the `unfocused_highlight_method` still draws a border
2) Using "line" has an issue with the vertical text alignment. Text is pushed up by the line. This actually happens for focused and unfocused.

CSpencerND:
> I haven't fully tested this yet but I've noticed a couple of issues:
> 
>     1. Using "text" for the `unfocused_highlight_method` still draws a border
> 
>     2. Using "line" has an issue with the vertical text alignment. Text is pushed up by the line. This actually happens for focused and unfocused.

Oh damn. I thought I tested those. I'll get back to you when it's fixed.

CSpencerND:
I solved the issue with the text highlight method, but the line pushing the text up has me stumped. I haven't changed anything about the drawbox method, but maybe changing the conditionals the way I did in the draw method sort of implicitly affects how the drawbox method works.

What I noticed with the original version is the entire bar looks like it adjusts to accommodate the line. As I increase the border width, the padding seems to increase and the boxes get wider, but with my version, none of this happens. The text above the line just goes higher. 

These are the observations I've made with my experimentation. I hope with this information you can guide me just a little further (if you have a chance).

elParaguayo:
I think the issue is here:
https://github.com/qtile/qtile/blob/master/libqtile/widget/groupbox.py#L97-L103

The problem is you end up with different padding values. If any highlight method is `line` then I think you need to make sure all groups always use that padding.

CSpencerND:
> I think the issue is here:
> 
> https://github.com/qtile/qtile/blob/da96b93d4efa09005a434e80be9ee3a2208a2f17/libqtile/widget/groupbox.py#L97-L103
> 
> The problem is you end up with different padding values. If any highlight method is `line` then I think you need to make sure all groups always use that padding.

Aaahh yes, thank you!

CSpencerND:
I tested many combinations of highlight methods, border widths, and paddings. Everything seems to be working as intended. 
I originally forgot to add a check for `block_highlight_text_color` which has now been implemented. 
Additionally, none of the changes should affect the original behavior of the widget. The default value for `unfocused_highlight_method` is `None` which means that no one should have to change their config if they don't know about or don't want to use the new functionality. They need to opt-in, rather than opt-out.

Here's a screenshot of my widget using `line` and `text` for different screens.
image

elParaguayo:
That looks nice however, looking at the code, I do feel like there must be opportunities to cut it down as there's a good amount of duplication. I'll take a look when I'm not feeling so tired!

CSpencerND:
> That looks nice however, looking at the code, I do feel like there must be opportunities to cut it down as there's a good amount of duplication. I'll take a look when I'm not feeling so tired!

I completely agree. I added a lot more code than I would have liked to. Especially all the nested if statements.

m-col:
What's the status of this PR? Looks like you guys might investigate if you can cut down some of the code duplication, but then otherwise it's all good?

elParaguayo:
It just feels like this could be much neater (and that's not a criticism of the PR) and the widget could possibly benefit from being refactored. 

I haven't had a chance to take a proper look.

Issue #2846, opened 2021-10-05
Multiple callbacks in Key using when


karthink:
I'm trying to assign multiple functions to a single key depending on whether the window is floating or tiled:

```python
    Key([mod, mod3], "h",
        lazy.window.move_floating(-32, 0).when(when_floating=True),
        lazy.layout.shuffle_left().when(when_floating=False),
        desc="Move window to the left"),
    Key([mod, mod3], "l",
        lazy.window.move_floating(32, 0).when(when_floating=True),
        lazy.layout.shuffle_right().when(when_floating=False),
        desc="Move window to the right"),
    Key([mod, mod3], "j",
        lazy.window.move_floating(0, 32).when(when_floating=True),
        lazy.layout.shuffle_down().when(when_floating=False),
        desc="Move window down"),
    Key([mod, mod3], "k",
        lazy.window.move_floating(0, -32).when(when_floating=True),
        lazy.layout.shuffle_up().when(when_floating=False),
        desc="Move window up"),
```

There are two problems:

- It runs the first callback only, the second is never run.
- It's ignoring the when condition: It runs the `move_floating` command for both tiled and floating windows.

Am I doing something wrong? Is there a cleaner way to do this?

karthink:
Here is another example:

```python
    Key([mod1], "Tab", lazy.group.next_window(), lazy.window.bring_to_front().when(when_floating=True),
        desc="Move window focus to other window"),
```

This should only activate bring_to_front() when the window is floating, but it activates for all windows, including tiled ones. This makes the tiled windows float when I press mod1+tab.

m-col:
OK, so I was able to reproduce too, and then got confused because the code looked fine, but then I discovered that `when_floating` isn't "limit this to when the current window is floating" but rather "also call this when the current window is floating". This means that we're actually seeing the intended behaviour. I can see why this would be desired, but it obviously isn't being communicated clearly enough in the docs!

To get the behaviour that you're after we'll need a way to specify *only* floating, perhaps by letting `when_floating` accept the string `"only"` in addition to `True` or `False`, and then _not_ executing the call when the current window is tiled. If I've got this all right, would you like to take a stab at implementing that? Otherwise I am also happy to do so!

PR #2900, opened 2021-10-16

karthink:
when_floating is None by default
when set to True, apply rule to floating windows only
when set to False, apply rule to tiled windows only

As per discussion in #2846.

I haven't been able to test this code. When I start Qtile with startx with the path to this directory, my Xserver crashes. Is there an easier way to test Qtile?

### DIFF
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -47,7 +47,7 @@ class LazyCall:
         self._kwargs = kwargs
 
         self._layouts: Set[str] = set()
-        self._when_floating = True
+        self._when_floating = None
 
     def __call__(self, *args, **kwargs):
         """Convenience method to allow users to pass arguments to
@@ -91,7 +91,7 @@ class LazyCall:
         return self._kwargs
 
     def when(self, layout: Optional[Union[Iterable[str], str]] = None,
-             when_floating: bool = True) -> 'LazyCall':
+             when_floating: Optional[bool] = None) -> 'LazyCall':
         """Enable call only for given layout(s) and floating state
 
         Parameters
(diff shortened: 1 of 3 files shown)


m-col:
Awesome, thanks!

>  Is there an easier way to test Qtile?

In the source directory, check out `scripts/xephyr`. It uses `Xephyr` to run a nested X session in a window within your main X session. For this you will need `Xephyr` installed. Check out http://docs.qtile.org/en/latest/manual/hacking.html#using-xephyr, and dependencies are also listed at the top of that page.

m-col:
Are there any docs that need updating to reflect this?

Also it might be a good idea to add an entry to CHANGELOG about this so others can find the feature!

m-col:
Looks like your branch has conflicts with master so you'll need to rebase. If needed, there is some information on how to do this at: http://docs.qtile.org/en/latest/manual/howto/git.html#i-was-told-to-rebase-my-work

It might be good to also squash your commits down into a single commit while you're there - there's also a section on that page for that in case you find it useful.

karthink:
I added a line in the example code illustrating the difference.

m-col:
The changes look good but the CI is complaining about inconsistent indentation in the changelog entry. Make sure that the indentation of your entry matches the rest of the file.

elParaguayo:
Can we fix the indentation in changelog, resolve the conflict and rebase into a single commit so we can get this merged?

karthink:
I fixed the indentation and rebased on the latest master.

elParaguayo:
Can you squash everything into a single commit (see http://docs.qtile.org/en/latest/manual/howto/git.html#i-was-told-to-squash-some-commits).

karthink:
OK, I think I managed to rebase on master and squash commits. Hope it looks right now.

libqtile/lazy.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, call: CommandGraphCall, args: Tuple, kwargs: Dict) -> None:
         self._kwargs = kwargs
 
         self._layouts: Set[str] = set()
-        self._when_floating = True
+        self._when_floating = None
```
elParaguayo, line 50:
```suggestion
        self._when_floating: Optional[bool] = None
```
This is the initial declaration of `when_floating` so you need to declare the type here.

This should fix the mypy error you're seeing.

elParaguayo (changes requested):

PR #2902, opened 2021-10-17
WL: implement drag and drop functionality

### LINKED ISSUE #2531: No drag-and-drop functionality with Wayland backend
# Issue description

Drag-and-drop functionality needs explicit implementation (from my very limited look into it, some data handling IPC and some surface renderering is required), at least for dragging e.g. files from one client to another.

I also can't seem to drag and drop _within_ Firefox. This may be unrelated - I would have thought this would be simple button event-based activity within a web page but maybe it uses the desktop's DnD functionality too (e.g.: dragging text around a text input box).

# Qtile version

master (152a1012bef9afb76ba7b03d6ee211e78b1da80d)

m-col:
This uses wlroots' API for getting drag-and-drop to work in the Wayland
backend. wlroots does most of the work here; we just need to implement
responding to the appropriate events (client requests) to render a
client-provided icon while dragging, and then wlroots handles data
transfer between clients at the drop.

This is used for dragging within clients (for example re-ordering browser
tabs in firefox; moving files into folders in a file manager) and
between clients (e.g. dragging files from a file manager into firefox).

Requires pywlroots 0.14.8.

Closes #2531

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -82,6 +82,7 @@ if typing.TYPE_CHECKING:
     from typing import List, Optional, Sequence, Tuple, Union
 
     from wlroots.wlr_types import Output as wlrOutput
+    from wlroots.wlr_types.data_device_manager import Drag
 
     from libqtile import config
     from libqtile.core.manager import Qtile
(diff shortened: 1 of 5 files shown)


elParaguayo:
I've got pywlroots 0.14.8 installed but, when I run this, the bar doesn't register any clicks at all. I'm also not seeing any drag and drop working (firefox register clicks but I can't reorder tabs).

Guessing I've got a dependency issue here.

PR #2905, opened 2021-10-17
Add low_background option to Battery widget

### LINKED ISSUE #2683: battery 'Low_Background' argument [ Suggestion ]
### Discussed in https://github.com/qtile/qtile/discussions/2668

<div type='discussions-op-text'>

<sup>Originally posted by **LamprosPitsillos** July 31, 2021</sup>
The battery low_background should be present, unlike low_foreground which is a difference in color of a few pixels , low_background is much better in letting you know that the battery is getting low.
I would love to see this simple feature implemented please.
Thanks!</div>

m-col:
This lets us change the background of widget when on low battery, in
addition to the foreground colour.

Closes #2683


m-col:
Done. It's actually the only test to check any of the runtime configurable options for `Battery` - all of the existing tests just check the internal mechanisms concerning battery levels. Might be nice to include a few more options in that test/other tests. Another time.

PR #2907, opened 2021-10-18
Update Import collections as of Python3.10

segatrooper:
Due to an update for Python3.10, Iterables and isinstance() has moved to collection.abc. Here is the doc refering to this change: https://docs.python.org/3/library/collections.abc.html


m-col:
We supported 3.10 now so looking into this again. It looks like the two uses of `collections` in that `drawer.py` use a full path to `collections.abc.Iterable`. This means that your change actually breaks it, because it then becomes `collections.abc.abc.Iterable`. Perhaps it should be `import collections.abc` rather than `import collections`.

PR #2903, opened 2021-10-17
WL: Set tiled edge xdg_surface state when mapping

m-col:
We currently tell a window's xdg_surface to tile its edges immediately
when it is created. This causes problems with some clients (e.g.
signal-desktop) which do need appear to be ready for being configured.
Instead let's delay setting this state until the client is ready and
emits the map event. This is also what sway does.

Fixes #2835

Issue #2544, opened 2021-06-11
Imperfect rendered frames due to no configure syncing with clients


m-col:
# Issue description

One of the advantages of Wayland is the ability to have perfect frames, but we don't do much to this end yet. The most noticable imperfection in the current backend is windows rendered with a default (client specified) width and height for a fraction of a second before responding to the dimensions we have sent them. This makes windows flash quickly as one size before changing to their final size.

To improve the experience we need to have a little configure-ackconfigure dance with clients before rendering their new state, or something along those lines. I haven't looked much into this (focussing on the functional essentials first!) but this would be great to have.

# Qtile version

current master 66341fc72551caae5e5ea93868cf5499ae1ee385

CharString:
When using floating windows with launchers like Kupfer or Ulauncher, I've noticed that the floating window only resizes when I click on it with the mouse. Is that this same issue?

m-col:
It could be but I am unsure.

I just gave kupfer a test and it seems to close the window when I try resizing it. I can see in the terminal that the process stays around and it'll happily open new windows if I run `kupfer` again, but I notice plenty of these messages which don't inspire hope:
```
(kupfer.py:55435): Wnck-WARNING **: 22:53:34.582: libwnck is designed to work in X11 only, no valid display found
```

ulauncher gives me stranger behaviour but I think it is what you describe. If focus moves to other windows (when I move the cursor, as I use `follow_mouse_focus`), it closes itself, but I think that is it's decision. If I keep the mouse over its own window I do have to click it for it to update. However, I don't think it is this issue because it _does_ update itself, e.g. when typing more keys, but it just doesn't resize its surface. I'm testing it further to see what's up with it, and how it behaves in under WCs.

m-col:
Unfortunately it works fine on sway so it's definitely a qtile bug. I'll investigate further and post an issue if i can't find an immediate solution.

CharString:
Kupfer shows the resizing behaviour when you need a third column. When you use the "Open with..." action, for example.

PR #2911, opened 2021-10-18
WIP: atomically apply window state changes

m-col:
This is a work in progress draft for some collaborative work with @jwijenbergh.

There are two main changes across these 5 commits:
 - a refactor of `Window.place()` and its use within the codebase. This becomes the single, exclusive point of entry for making geometric state changes on windows. `Window.{x,y,width,height}` are no longer intended to be used as direct setters, though they will continue to work (logging deprecation warnings). To reduce code duplication, all of its arguments have become kwargs so you can do e.g. `window.place(borderwidth=4)`, and that would be the canonical way to change borderwidth. This makes it much easier to control state changes on windows and group them for...
 - atomic state changes on windows and groups of windows. Higher-level operations such as changing groups are performed within the `Core.masked()` context manager so that configure requests sent to windows can be saved in a pending queue, and only when the final window is acked are the state changes applied simultaneously on all involved windows.

The (intended) effect is that every frame is rendered perfectly, addressing https://github.com/qtile/qtile/issues/2544. Noticably this means windows will not appear with the wrong geometry for a frame before fixing themselves, and adding/removing windows from a tiled layout results in the tiled layout changing atomically without any windows lagging their geomtry change for very short (but noticable) amount of time.

Current issues that need resolving:
 - when changing to a group, it seems as though the old buffer is being rendered for windows on the now-current group. This is the most noticable in terminals which might only commit new buffers when things are printed, so until a key is pressed or they are clicked, they render old content.
 - even before these changes, firefox displays a single pixel window before fixing itself (that's their doing, I believe). Since https://github.com/qtile/qtile/pull/2903 was merged, it doesn't seemt to fix itself with this patch also applied.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -196,8 +196,10 @@ class _Window(CommandObject, metaclass=ABCMeta):
         return False
 
     @abstractmethod
-    def place(self, x, y, width, height, borderwidth, bordercolor,
-              above=False, margin=None, respect_hints=False):
+    def place(
+        self, x=None, y=None, width=None, height=None, borderwidth=None,
+        bordercolor=None, above=False, margin=None, respect_hints=False
+    ):
         """Place the window in the given position."""
 
     def _items(self, name: str) -> ItemT:
@@ -342,10 +344,14 @@ class Window(_Window, metaclass=ABCMeta):
     def cmd_set_size_floating(self, w: int, h: int) -> None:
         """Set window dimensions to w and h"""
 
-    @abstractmethod
-    def cmd_place(self, x, y, width, height, borderwidth, bordercolor,
-                  above=False, margin=None) -> None:
-        """Place the window with the given position and geometry."""
+    def cmd_place(
+        self, x=None, y=None, width=None, height=None, borderwidth=None,
+        bordercolor=None, above=False, margin=None, respect_hints=False
+    ):
+        self.place(
+            x=x, y=y, width=width, height=height, borderwidth=borderwidth,
+            bordercolor=bordercolor, above=above, margin=margin
+        )
 
     @abstractmethod
     def cmd_toggle_floating(self) -> None:
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import os
 import time
 import typing
@@ -29,6 +30,7 @@ import wlroots.helper as wlroots_helper
 from pywayland import lib
 from pywayland.protocol.wayland import WlSeat
 from pywayland.server import Display
+from wlroots.util.clock import Timespec
 from wlroots.wlr_types import (
     Cursor,
     DataControlManagerV1,
@@ -79,7 +81,7 @@ from libqtile.backend.wayland.output import Output
 from libqtile.log_utils import logger
 
 if typing.TYPE_CHECKING:
-    from typing import List, Optional, Sequence, Tuple, Union
+    from typing import Dict, List, Optional, Sequence, Tuple, Union
 
     from wlroots.wlr_types import Output as wlrOutput
     from wlroots.wlr_types.data_device_manager import Drag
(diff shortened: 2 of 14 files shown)


jwijenbergh:
> This is a work in progress draft for some collaborative work with @jwijenbergh.
> 
> There are two main changes across these 5 commits:
> 
>     * a refactor of `Window.place()` and its use within the codebase. This becomes the single, exclusive point of entry for making geometric state changes on windows. `Window.{x,y,width,height}` are no longer intended to be used as direct setters, though they will continue to work (logging deprecation warnings). To reduce code duplication, all of its arguments have become kwargs so you can do e.g. `window.place(borderwidth=4)`, and that would be the canonical way to change borderwidth. This makes it much easier to control state changes on windows and group them for...
> 
>     * atomic state changes on windows and groups of windows. Higher-level operations such as changing groups are performed within the `Core.masked()` context manager so that configure requests sent to windows can be saved in a pending queue, and only when the final window is acked are the state changes applied simultaneously on all involved windows.
> 
> 
> The (intended) effect is that every frame is rendered perfectly, addressing #2544. Noticably this means windows will not appear with the wrong geometry for a frame before fixing themselves, and adding/removing windows from a tiled layout results in the tiled layout changing atomically without any windows lagging their geomtry change for very short (but noticable) amount of time.
> 

Thanks for the detailed pr! I am currently running this and it's really nice to see my terminals not have the dimension 'flicker'. 

> Current issues that need resolving:
> 
>     * when changing to a group, it seems as though the old buffer is being rendered for windows on the now-current group. This is the most noticable in terminals which might only commit new buffers when things are printed, so until a key is pressed or they are clicked, they render old content.

Could you give me a reliably way to reproduce this (if there is one). I am currently trying it out and I can't seem to get in this state

>     * even before these changes, firefox displays a single pixel window before fixing itself (that's their doing, I believe). Since [WL: Set tiled edge xdg_surface state when mapping #2903](https://github.com/qtile/qtile/pull/2903) was merged, it doesn't seemt to fix itself with this patch also applied.

Firefox works for me. I observe that firefox still goes to a default dimension for a second (like half my screen) and then goes to the whole screen in a layout with it as the only window. Is this something you observer or is firefox completely broken for you now?
The same holds with emacs for me. If I run emacs with emacsclient (making startup very fast as compared to normal emacs) it still happens but it goes to the correct dimension much quicker.
I will see if I can correct this behaviour.

jwijenbergh:
> Here is a video of it happening. Note the firefox window opens on a different group from the current, and it is launched via a keybinding. Then I change to that group, show the small window highlighted by some double borders, and then when I open a foot window firefox fixes itself. Now, the interesting thing is that it only happens to this firefox profile, and if I launch firefox with a different profile it works fine! No clue what is up with that, or if it's their or our bug.

Now that is really weird indeed. I do not seem to have this issue :(. I've had weird firefox profiles before. Does it work with that profile on e.g. sway?

> I am getting occasional problems with dragging floating windows around, and also sometimes segfaults that crash Qtile back to TTY. Nothing reliable yet though so not much else to report.

Hmmmmmm, this is funny because I just had a segfault suddenly when reading that and opening a foot terminal, are you hacking me :eyes:

> The problem with this is that it doesn't seem very reliable, so I will play around with things to see if I can find a pattern. The 200ms cut-off thing is literally a race so maybe it happens when foot is too slow or something. Here is another video showing this. Note the block cursor at the bottom left corner of the window. When changing group goes fine, it's filled (foot is correctly responding to the activated state), but sometimes it is empty. Near the end of the video I type a key, and foot receives this correctly and the next rendered buffer has this key and the block cursor, so it does have keyboard focus correctly.

Interesting thanks, I hope I can get to the bottom of this. I will do some debugging and try various things this week and I will report back when I have some news.

m-col:
Something that we do not do at all, which we could if it helps, is to remove listeners when unmapping and return them again upon mapping, rather than just setting them up at the first map event and then leaving them.

jwijenbergh:
Sorry for taking so long on this. I have actually been making a toy wayland compositor in c++ in my free time as I want to more thoroughly understand wayland. I will get back to this when I have a good understanding of what the best way to do this is

m-col:
No worries at all :smile: sounds great!

m-col:
Another relevant issue: https://github.com/qtile/qtile/pull/2911

jwijenbergh:
> This PR is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 30 days.

go away bot, I want to pick this up eventually :D

jwijenbergh:
@m-col should we close this or do you think we can rebase this on top of 0.16 pr?

PR #2912, opened 2021-10-19
Fix scratchpad focus issue in X11

### LINKED ISSUE #2588: Scratchpad alternates opening with and without focus
I'm seeing some slightly odd behaviour in my scratchpad (running Alacritty) where the alt key is not being respected. For exampled `alt`+`.` in a terminal would give me the last argument from the previous command and `alt` + number gives a different screen in `irssi`.

This occurs when the scratchpad window is visible but doesn't have focus.

If I toggle the scratchpad, it alternates having focus and not having focus. There needs to be another window open on the active group for this to happen.

This bug is present on latest master: d5425eeb

elParaguayo:
There is an issue with some scratchpads where the window will alternate opening with and without focus which can cause issues
for certain keybindings.

The issue seems to be that `set_input_focus` tries to focus the root window rather than the scratchpad window.

This PR prevents focussing of the root window. This may or may not cause other issues so needs testing!

Fixes #2588

@m-col - I'm a bit concerned about preventing focus of the root window so this may be more of a plaster than finding the root cause.

@jwijenbergh - can you have a test and see if this works (and if you notice any other bugs)

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -125,11 +125,12 @@ class XWindow:
         self.conn.conn.core.KillClient(self.wid)
 
     def set_input_focus(self):
-        self.conn.conn.core.SetInputFocus(
-            xcffib.xproto.InputFocus.PointerRoot,
-            self.wid,
-            xcffib.xproto.Time.CurrentTime
-        )
+        if self.wid != self.conn.default_screen.root.wid:
+            self.conn.conn.core.SetInputFocus(
+                xcffib.xproto.InputFocus.PointerRoot,
+                self.wid,
+                xcffib.xproto.Time.CurrentTime
+            )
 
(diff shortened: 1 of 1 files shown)


m-col:
> @m-col - I'm a bit concerned about preventing focus of the root window so this may be more of a plaster than finding the root cause.

I think so too. The root window does not get focussed much, only when the current screen has no mapped windows, but I think this does need to happen otherwise key presses might get sent to unexpected places. In a multi-monitor setup, they'd continue to get sent to the previously focussed window.

I suggest giving `Window` a warp_pointer method (again) and then the x11 backend can use its original behaviour. Wayland `Window`s can just use `self.core.pointer`. Coordinates might need changing though because it would make sense to make this method relative to the window, and I think that's what the scratchpad and x11 windows used to do.

PR #2914, opened 2021-10-19
Persist ScratchPad groups across config reloads

m-col:
As part of the config reload functionality, we decided to not persist
groups across reloads as they do across restarts. This was because it
becomes difficult to consolidate existing groups that may or may not be
reflected in the config with new groups added to the config and groups
removed from the config.

However, this also applied to scratchpads, which can not be dynamically
changed during normal runtime and so do not pose the same issues. The
effect of doing this is that following a config reload each
scratchpad groups is fresh and old scratchpad windows get orphaned to
the current group. Instead, let's persist scratchpad groups across
reloads to prevent this and prevent this unexpected behaviour.

--

I have also added a scratchpad with one or two dropdowns to the config-reload test, however I couldn't get dropdowns to be toggled correctly. I'm not exactly sure what's going on there, so perhaps some fresh eyes will provide some insight. Basically the idea is it'd be good to also spawn a test dropdown, reload, check the window persists in the scratchpad, toggle the second dropdown, reload again, check the first window persists in the scratchpad, and check the second window (from a now defunct dropdown) has successfully been orphaned to the current group.

### DIFF
--- a/libqtile/core/state.py
+++ b/libqtile/core/state.py
@@ -90,7 +90,7 @@ class QtileState:
         """
         Remove any windows from now non-existent scratchpad groups.
         """
-        client_wid = client.window.wid
+        client_wid = client.wid
         if client_wid in self.orphans:
             self.orphans.remove(client_wid)
             client.group = None
--- a/libqtile/core/state.py
+++ b/libqtile/core/state.py
@@ -30,6 +30,7 @@ class QtileState:
 
     If `restart` is True, the current set of groups will be saved in the state. This is
     useful when restarting for Qtile version updates rather than reloading the config.
+    ScratchPad groups are saved for both reloading and restarting.
     """
     def __init__(self, qtile, restart=True):
         self.groups = []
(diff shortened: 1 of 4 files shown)


elParaguayo:
Weird. I can't get that dropdown to work either. I can see it listed in `dropdown_info` but using `dropdown_toggle` doesn't seem to create a window.

Is that the issue you had?

m-col:
> Is that the issue you had?

I think so yeah

> EDIT: If I try a manager_nospawn.c.spawn and the dropdown's command, I still don't get a window so something's stopping spawn working here.

Should a `cmd_spawn()` work if the manager hasn't forked and created its own event loop? The backend won't have initialised.

> EDIT 2: It look like the window.py script is exiting immediately (I added a print statement to the end of the file and this is printed before I get the pid from spawn().

Interesting - any ideas as to what might be causing it? We might be able to get error messages printed from the script, if GTK/GDK is printing anything

elParaguayo:
Haven't found any clues as to why it's not working but still looking.

elParaguayo:
@m-col I think you found the fix for the test. Are you able to push so we can get this merged?

m-col:
... did I?

m-col:
The PR has the most up-to-date version

elParaguayo:
> ... did I?

Yes. You said you needed to use the `@Retry` decorator when checking for the scratchpads.

Or maybe I've muddled PRs!

m-col:
:exploding_head: that was for another PR but I bet that is the solution here!!

m-col:
Actually i wouldn't trust my memory, it probably was for this one

PR #2915, opened 2021-10-19
WL: Fix no Internal clicks when bring_front_click is floating_only

### LINKED ISSUE #2793: Wayland: mouseclicks not received by Internal windows
# Issue description

Mouse clicks are not received by Internal windows. e.g. clicking on widgets in the bar does not trigger mouse callback events.

# Qtile version

Latest master: 76b3962d

# Stack traces

No errors shown in log on debug mode. However, there were no events at all when the mouse entered the bar.

# Configuration

I haven't tried with default_config as Wayland had worked before with this config.

m-col:
This error is caused by `win.floating` raising an exception because
`Internal` doesn't have a float state. This commit re-writes the
bring-to-fron conditions to be more easy to read and explicitly check
that the window is not internal. It now passes this exception just fine
for all 3 window classes. As a result, the code continues and forwards
clicks to internal windows.

Closes #2793

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -671,11 +671,10 @@ class Core(base.Core, wlrq.HasListeners):
         if found:
             win, surface, _, _ = found
 
-            if self.qtile.config.bring_front_click:
-                if (
-                    self.qtile.config.bring_front_click != "floating_only"
-                    or win.floating
-                ):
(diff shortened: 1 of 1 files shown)


elParaguayo:
Cool. I can test this later today.

elParaguayo:
Nice. This fixes the issue.

PR #2924, opened 2021-10-20
Fix build errors in docs

### LINKED ISSUE #2923: Docs are failing to build (again)
See: https://readthedocs.org/projects/qtile/builds/15011707/

Looks like this is a dependency issue related to pywlroots.

elParaguayo:
Adding Wayland compositor page which references backend class object introduced build errors as pywayland and wlroots are not
in the readthedocs environment. We therefore need to add the modules to the list of classes to be mocked.

Fixes #2923

I think we also need to look at how we test docs buillding as the readthedocs environment is not the same as the `tox` environment.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -46,7 +46,6 @@ MOCK_MODULES = [
     'pywayland',
     'pywayland.protocol.wayland',
     'pywayland.server',
-    'trollius',
     'wlroots',
     'wlroots.helper',
     'wlroots.util',


elParaguayo:
> I think the problem here is that we install wlroots etc in tox so our docs test completes ok but readthedocs doesn't have those modules installed.

As per our IRC chat, it looks like the actual issue is CI/tox running the wrong environments. Have opened a new issue (#2927) for that one.

Issue #2587, opened 2021-06-28
Replay mouse/key bindings?


Barbarossa93:
# Issue description
In sxhkd, there's an option to add a tilde in front of a mouse/keybinding. This allows the binding to be replayed to the other clients, i.e. allows the captured event to not only trigger the binding set in the config, but also use that binding on the focused client. For me, this is very useful for a root menu script that is bound to the mouse right click that only activates when the mouse is above the root window and not a client window. I was wondering if there was a way to achieve this in qtile? The root menu in question is not only capable of launching applications, but can also manipulate windows as well (check out jgmenu).

# Qtile version
0.17.1.dev347+g0c049edd

PR #2925, opened 2021-10-20
Make keybinding swallow configurable

### LINKED ISSUE #324: Pass on ignored key events to clients

jwijenbergh:
Key and KeyChord now get another argument: A swallow boolean

With "swallow", I mean whether or not the focused client still gets the keybinding passed to it. When it is set to False, the focused window will get the keybinding passed after the command has been executed. True is the default. Key bindings that are not executed (so also keybindings that fail the `when(...)` check) are never swallowed

Fixes #324 (needs to be re-opened as it was reverted) and implements the feature request in #2587 (as of now partially, but I plan to do the same for mouse keybindings).

This seems to properly handle all cases I tested. The problem we had before (https://github.com/qtile/qtile/pull/2625) is that we did not set handled to `False` when the keybinding is an instance of `KeyChord`. This needs a ton of testing though to confirm everything works correctly. You can test with something like 

```python

# Make sure it's imported
def keychordfunc(qtile):
    print("EXECUTED KEYCHORD")

keys = [
    KeyChord([mod], "y", [
        KeyChord([], "z", [
            KeyChord([], "x", [
                Key([], "c", lazy.function(keychordfunc).when(layout='max'))
            ], mode="inner")
        ])
    ], mode="outer", swallow=False),
...
```

 Added @m-col to co-authors as most of it is based on the previous pr (https://github.com/qtile/qtile/pull/2561).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,6 +22,11 @@ Qtile x.x.x, released xxxx-xx-xx:
           `lazy.group.toscreen(toggle=True)`
         - Tile layout has new `margin_on_single` and `border_on_single` option to specify
           whether to draw margin and border when there is only one window.
+        - `Key`, `KeyChord`, `Click` and `Drag` bindings now have another argument `swallow`.
+          It indicates whether or not the pressed keys should be passed on to the focused client.
+          By default the keys are not passed (swallowed), so this argument is set to `True`.
+          When set to `False`, the keys are passed to the focused client. A key is never swallowed if the
+          function is not executed, e.g. due to failing the `.when()` check.
 
 Qtile 0.18.1, released 2021-09-16:
     * features
--- a/libqtile/backend/wayland/keyboard.py
+++ b/libqtile/backend/wayland/keyboard.py
@@ -114,8 +114,8 @@ class Keyboard(HasListeners):
             mods = self.keyboard.modifier
             for keysym in keysyms:
                 if (keysym, mods) in self.grabbed_keys:
-                    self.qtile.process_key_event(keysym, mods)
-                    return
+                    if self.qtile.process_key_event(keysym, mods):
+                        return
 
             if self.core.focused_internal:
                 self.core.focused_internal.process_key_press(keysym)
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -53,7 +53,6 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
-    xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
     xcffib.xproto.NoExposureEvent,
@@ -358,6 +357,7 @@ class Core(base.Core):
             "ButtonPress",
             "ButtonRelease",
             "KeyPress",
+            "KeyRelease",
         ]
         if hasattr(event, "window"):
             window = self.qtile.windows_map.get(event.window)
@@ -465,7 +465,7 @@ class Core(base.Core):
                     modmask | amask,
                     code,
                     xcffib.xproto.GrabMode.Async,
-                    xcffib.xproto.GrabMode.Async,
+                    xcffib.xproto.GrabMode.Sync,
                 )
         return keysym, modmask & self._valid_mask
 
@@ -514,7 +514,7 @@ class Core(base.Core):
                 True,
                 self._root.wid,
                 eventmask,
-                xcffib.xproto.GrabMode.Async,
+                xcffib.xproto.GrabMode.Sync,
                 xcffib.xproto.GrabMode.Async,
                 xcffib.xproto.Atom._None,
                 xcffib.xproto.Atom._None,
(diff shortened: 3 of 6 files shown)


jwijenbergh:
On second thought, it's probably too complicated to have 3 values. I will set it to a boolean and if they really want a binding to always be swallowed, then they can probably hack something together in their function instead of using `.when`.

jwijenbergh:
I have added mouse bindings. This needed changes in the x11 backend by setting `grab_button` `pointer_mode` to sync, same with the keyboard changes by m-col. @m-col can you expand a bit why this is needed (Async -> Sync) in x11 and if it looks okay this way? I am reading the x11 documentation and it's quite vague.

You can test with

```python
def nevergonna(qtile):
    print("give you up")

# Drag floating layouts.
mouse = [
    Click([], "Button1", lazy.function(nevergonna), swallow=False)
...
```

jwijenbergh:
> > Yeah that makes sense. I will see what the best way to do this is later. Sync was giving me situations where it didn't respond to my keybindings any more.
> 
> This sounds like the keyboard was still grabbed... can't remember the exact logic but I think you have to ungrab to allow key events to continue as normal, and I think there does have to be an explicit call of something to say "reply key to client" or "do not replay". But yeah, docs are bad, everything is trial and error.

Indeed, hmmm.... When I handle KeyRelease events and do SyncKeyboard there, everything works fine so far.


```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index d5a6fa75..c107dcf5 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -53,7 +53,6 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
-    xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
     xcffib.xproto.NoExposureEvent,
@@ -358,6 +357,7 @@ class Core(base.Core):
             "ButtonPress",
             "ButtonRelease",
             "KeyPress",
+            "KeyRelease",
         ]
         if hasattr(event, "window"):
             window = self.qtile.windows_map.get(event.window)
@@ -617,6 +617,11 @@ class Core(base.Core):
             # Forward the event to any focused client
             self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayKeyboard, event.time)
 
+    def handle_KeyRelease(self, event) -> None:  # noqa: N802
+        assert self.qtile is not None
+
+        self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.SyncKeyboard, event.time)
+
     def handle_ButtonPress(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
```

This also seems what sxhkd is doing:

https://github.com/baskerville/sxhkd/blob/fe241b0d2d70c9c483b23cf3cd14f1383f0953a2/src/sxhkd.c#L241

m-col:
Is that necessary? I don't think listening for key release events should be.

jwijenbergh:
> Is that necessary? I don't think listening for key release events should be.

I can't seem to get it to unfreeze without this, will need to try later with a fresh mind.

jwijenbergh:
> Is that necessary? I don't think listening for key release events should be.

Without it when you press a key 'simultaneously' it will get frozen.

Every wm I've seen that handles keybinds in a sync manner all define a keyrelease event, e.g. i3 and the sxhkd tool. IMO it might be better to continue with the keyrelease event as it's very easy to just add key release bindings now. But if you have a better idea just let me know. In general key release events are very easy by just adding an `is_release` to `process_key_event` that influences the swallow variable together with key.is_release (e.g. https://github.com/qtile/qtile/pull/2785). For now, in keyrelease we should just be doing replaykeyboard.

jwijenbergh:
Also as the commands are a list, when do we swallow? Do we not swallow when one of the commands has failed or do we swallow when one of the commands has succeeded? The former is implemented currently, but the latter is maybe wanted here?

jwijenbergh:
Steps taken:
- Fixed double clicking when swallow was set to False and clicking on the root window
- Added tests for mouse bindings (these were quite tricky to test)
- Rebased with master
- Marked ready for review

m-col:
All looks great! Could you add an entry to the changelog for this, then good to merge!

jwijenbergh:
> All looks great! Could you add an entry to the changelog for this, then good to merge!

Done! Let me know if this is okay.

m-col (approved):
Awesome, looks great, will merge when we see the green. Thanks again!

jwijenbergh:
Thank you!

PR #2929, opened 2021-10-21
fixed _get_closest function in xmonad, added default value to handle empty candidates list

fiorematteo:
Without this fix `min()` would throw an error every time you would call `cmd_swap_right`, `cmd_swap_left,`  `cmd_right` and `cmd_left` with no available client to move.
```
2021-09-05 08:38:14,313 ERROR libqtile manager.py:process_key_event():L359 KB command error right: Traceback (most recent call last):
  File "/home/matteo/.local/lib/python3.9/site-packages/libqtile/command/interface.py", line 313, in call
  File "/home/matteo/.local/lib/python3.9/site-packages/libqtile/layout/xmonad.py", line 736, in cmd_right
    self.clients.current_client = self._get_closest(x, y, candidates)
  File "/home/matteo/.local/lib/python3.9/site-packages/libqtile/layout/xmonad.py", line 688, in _get_closest
    target = min(
ValueError: min() arg is an empty sequence
```

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -687,7 +687,8 @@ class MonadTall(_SimpleLayoutBase):
         """Get closest window to a point x,y"""
         target = min(
             clients,
-            key=lambda c: math.hypot(c.info()["x"] - x, c.info()["y"] - y)
+            key=lambda c: math.hypot(c.x - x, c.y - y),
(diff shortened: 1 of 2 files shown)


fiorematteo:
I added the tests, I'm not sure they are correct. This is my first time using pytest.

PR #2941, opened 2021-10-23
Fix up some bar and widget related code

m-col:
These are just some smallish fixes for the bar and widget code:

1. Move the widget orientation check into `_configure` so that failures do not fail config loading, only widget setup, so the rest of the config continues to work.
2. Give a more useful exception when a widget is not given an `int` for its size.
3. Handle consecutivete `STRETCH` widgets a bit differently, where instead they are grouped together and each group of consecutive stretch widgets is given the same amount of space. Lots of info on this, the issue, the solution etc in the commit message.
4. Add a `FakeBar` test class for the widget tests to reduce lots of code duplication that was collecting there.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -238,13 +238,6 @@ class Bar(Gap, configurable.Configurable):
         width = self.width + (self.border_width[1] + self.border_width[3])
         height = self.height + (self.border_width[0] + self.border_width[2])
 
-        for w in self.widgets:
-            # Executing _test_orientation_compatibility later, for example in
-            # the _configure() method of each widget, would still pass
-            # test/test_bar.py but a segfault would be raised when nosetests is
-            # about to exit
-            w._test_orientation_compatibility(self.horizontal)
-
         if self.window:
             # We get _configure()-ed with an existing window when screens are getting
             # reconfigured but this screen is present both before and after
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import typing
+from collections import defaultdict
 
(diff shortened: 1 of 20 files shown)


elParaguayo:
This is looking really nice. I'd like to play around with the spacers a bit to give them a little "stress test".

Anything else you need to add here?

m-col:
Nope, all good

elParaguayo:
So, if you you use two spacers to centre something on the bar it works very nicely. However, if one side becomes bigger than half of the bar then you get some issues. The "centred" widget is still centred but it's drawn on top of the other widgets.



Not sure how to best handle that but I would have thought we try to use available space (e.g. the gap on the left here).

Thoughts?

m-col:
I think that should be the expected behaviour.

elParaguayo:
OK. Any sort of warning about overlapping widgets? How about handling mouse clicks?

m-col:
Hmm I think in practice it's pretty obvious - you get an overlap and you think "oh i have too much on my bar" and fix it.

elParaguayo:
I'd like to get this one merged. How are you doing on vertical text widgets?

m-col:
> I'd like to get this one merged. How are you doing on vertical text widgets?

I haven't taken a look at it yet I'm afraid.

elParaguayo:
Do you want me to have a look then?

m-col:
That would be brilliant, thank you!

m-col:
@elParaguayo happy for this to be merged now the vertical bar textbox code is merged?

elParaguayo:
If I can avoid pressing the wrong button on my phone this time...

Yes, happy to merge.

PR #2942, opened 2021-10-23
New Net widget option to display the current speed with a static prefix

lock-3:
Within the net widget it would be nice to have the option to always show the same unit (prefix) of the current traffic speed. I never liked the fact, that the prefix jumps between "kB" and "MB" and so on. It makes it much harder for me to read the output of the widget with a fast look.

With this change the user can set the widget option "prefix" to one of the
`allowed_prefixes = ["", "k", "M", "G", "T", "P", "E", "Z", "Y"]`
The default setting is "None" which keeps the unit switching depending to the actual traffic (old behaviour).

**This is my first contribution to any project. So please double-check and tell me if i did something wrong in the process.**

One last question: Should the documentation be updated after such a change? If yes, how?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ Qtile x.x.x, released xxxx-xx-xx:
         - Add ability for widget `mouse_callbacks` to take `lazy` calls (similar to keybindings)
         - Add `aliases` to `lazy.spawncmd()` which takes a dictionary mapping convenient aliases
           to full command lines.
+        - Add a new 'prefix' option to the net widget to display speeds with a static unit (e.g. MB).
 
 Qtile 0.18.1, released 2021-09-16:
     * features
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -40,14 +40,28 @@ class Net(base.ThreadPoolText):
         ('format', '{interface}: {down} \u2193\u2191 {up}',
          'Display format of down/upload/total speed of given interfaces'),
         ('interface', None, 'List of interfaces or single NIC as string to monitor, \
-            None to displays all active NICs combined'),
+            None to display all active NICs combined'),
         ('update_interval', 1, 'The update interval.'),
         ('use_bits', False, 'Use bits instead of bytes per second?'),
+        ('prefix', None, 'Use a specific prefix for the unit of the speed.'),
     ]
 
     def __init__(self, **config):
         base.ThreadPoolText.__init__(self, "", **config)
         self.add_defaults(Net.defaults)
+
+        self.factor = 1000.0
+        self.allowed_prefixes = ["", "k", "M", "G", "T", "P", "E", "Z", "Y"]
+
+        if self.use_bits:
+            self.base_unit = "b"
+            self.byte_multiplier = 8
+        else:
+            self.base_unit = "B"
+            self.byte_multiplier = 1
+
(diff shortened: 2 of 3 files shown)


m-col:
Apologies for the delay. The code itself looks great. Looks like there is a conflict in the changelog, so please could you rebase and resolve that, then this would be good to merge!

> I also think the max(*) operation in line 82 power = max(min(power, len(units) - 1), 0) is not necessary. I'm not 100% sure and don't know the reasoning behind this, so I'll leave the code as is.

Yeah I think you're right - if `num_bytes` is always positive, `power` is initially at least 0, so the `max()` call is redundant. It would be great to remove that too :)

lock-3:
Ok, I removed the `max()` call.

The conflict in the CHANGELOG I've already fixed yesterday. At least i think so. The problem was the tab indentation, but now i used whitespaces.

CHANGELOG
```diff
@@ -13,6 +13,10 @@ Qtile x.x.x, released xxxx-xx-xx:
           `group_name`. For the time being a log warning is in place and a
           migration is added. In the future `groupName` will fail.
         - Add `min/max_ratio` to Tile layout and fix bug where windows can extend offscreen.
+        - Add ability for widget `mouse_callbacks` to take `lazy` calls (similar to keybindings)
+        - Add `aliases` to `lazy.spawncmd()` which takes a dictionary mapping convenient aliases
+          to full command lines.
+        - Add a new 'prefix' option to the net widget to display speeds with a static unit (e.g. MB).
 
 Qtile 0.18.1, released 2021-09-16:
```
m-col, line 21: Looks like something didn't quite go right when you rebased - there are some things that are appearing as added here which were already there.

lock-3: I must confess that git is quit new to me. I tried a `git rebase origin/master` and get
`Current branch Net_Widget_Option is up to date.`.

So the rebase did not fix the conflict. As a workaround i removed the change i made in the `CHANGELOG`. Currently I do not have the time to take a closer look at the `rebase` functionality. If you some easy commands for me to start with i may give it another try. Otherwise its fine for me to not mention it in the Changelog.

m-col: No worries at all! Removing it from the changelog works - I can merge the main change and then add your changelog entry to the changelog in a follow-up commit :)

What would work to resolve the conflict is `git pull --rebase origin master` (or "upstream" if you have configured "upstream" to point to qtile/qtile, as "origin" will likely point to your own fork!). This will then update your local branch with your PR's commit removed initially, and then will try to put it back on that new base. It will conflict doing so, and prompt you to edit the conflicting files. The changelog would then have an annotation in the file indicating the conflict to make it easy to resolve. It can be pretty confusing for sure!

m-col:
@elParaguayo Is this the kind of thing you would advocate for a test for, or is it too small a feature for that?

elParaguayo:
> @elParaguayo Is this the kind of thing you would advocate for a test for, or is it too small a feature for that?

Well, the dream/goal is 100% coverage so, yes, I'd want a test.

There's already a test for the widget (https://github.com/qtile/qtile/blob/master/test/widgets/test_net.py) so we're just talking about adding a test to cover the new code. Should be quick.

m-col:
If amended the commit to resolve the changelog conflict and add a test

elParaguayo:
Thanks!

PR #2947, opened 2021-10-25
2935 System tray icons overlap when changing status

krive001:
Bar background set.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -159,7 +159,7 @@ class Bar(Gap, configurable.Configurable):
         width of a vertical bar.
     """
     defaults = [
-        ("background", "#000000", "Background colour."),
+        ("background", "#000000.0", "Background colour."),
         ("opacity", 1, "Bar window opacity."),
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
     ]


krive001:
Not because you can not paint because it is transparent, but because here (e3487bafc1df5a0e47d207b28182f0d7232b352e) who was led from the internal window to the bar binding.

elParaguayo:
That commit is for Wayland which should have no impact at all on `SysTray` (which only runs on an X11 backend) . Any file in `libqtile/backend/wayland` will not be run on any X11 backend. 

Can you highlight a particular line that you think's the issue.

krive001:
xprop to the bar or `qtile cmd-obj  -o cmd -f windows`
before:
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
QTILE_INTERNAL(CARDINAL) = 1
```

After:
```
QTILE_INTERNAL(CARDINAL) = 1
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
```

krive001:
The wayland is a rudimentary thing you may never mind. X has been in development for decades.

krive001:
That's why the sway was created from the i3. The two cannot be considered as one thing.

elParaguayo:
I'm sorry, I just have no idea what you are trying to say. You've said there's an issue with e3487ba but that commit is Wayland only and should have no impact on SysTray which will only ever run on X11.

As for the `xprop` output, you've pasted that before, what's the issue? They say the same thing just in a different order.

Yes, I know sway is different from i3, what is your point?

krive001:
The fact that it is not the bar that handles the window, but the X from here.
There are settings that qtile cannot access.
It would have been more noble to branch qtile before the introduction of wayland.

krive001:
Because X and Wayland do not treat windows the same way.
There are completely different API calls.

m-col:
They are _not_ treated the same way.

X11 internal windows: https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/window.py#L1078
Wayland internal windows: https://github.com/qtile/qtile/blob/master/libqtile/backend/wayland/window.py#L691

The two backends are split into separate folders and only one of these is used at a time: https://github.com/qtile/qtile/tree/master/libqtile/backend

krive001:
And the event mask?
They were thus ignored.

m-col:
Are you missing events?

krive001:
From this commit, it all entrusts window operation to X11, not the bar.

m-col:
Are you missing events?

krive001:
My problem is that from here all the windows on the bar will get out. Which I don't think is right.
From now on, the bar will have no effect on the window

krive001:
I don't think visual_id should be changed to qtile,

elParaguayo:
What do mean the windows "will get out"?

What effect should the bar have on the windows? (which windows?)

elParaguayo:
> I don't think visual_id should be changed to qtile,

Fine. Then you're welcome to play around with it and see what works for you.

krive001:
Why should the systemtray appear on root abels?

elParaguayo:
Why do you say they're on the root window? SysTray creates a new window.

From my testing: if you set a bar with a 32 bit colour depth then the system tray icon windows have a transparent background and you can't set this to the bar background. The fix for this was, where the user wanted an opaque bar, to set the bar colour depth to 24 bit and tell the SysTray to advertise that visual_id so windows know what depth to use. Replacing the backpixmap on 24 bit windows does seem to work.

krive001:
Because he was taken out with this commit. Therefore, it is not inky to paint. Nert is not part of the bar.

krive001:
```qtile cmd-obj  -o cmd -f windows | grep "'id'" |  awk '{print $2}'```

**Comment by elParaguayo** [comment]
If "this commit" is that Wayland one, it has **nothing** to do with the issue here. It is a completely separate backend.

**Comment by elParaguayo** [comment]
`qtile cmd-obj -o cmd -f windows` will not show you the SysTray window.

See this comment for how to get details on it: https://github.com/qtile/qtile/issues/2935#issuecomment-965083993

**Comment by krive001** [comment]
There are the facts. Why don't you want to believe it?

**Comment by krive001** [comment]
It affects X11 in the same way.

**Comment by elParaguayo** [comment]
What facts? You've shown us nothing useful. I asked you to show me the line in that commit that you think is the problem and you haven't.

That Wayland commit does not impact X11.  If it does then that would definitely be a bug but, again, be specific in pointing to the error rather than just pasting a link to an entire commit.

**Comment by krive001** [comment]
That the windows are getting out of the bar.
From now on, it has no effect.

**Comment by krive001** [comment]
So you can't even paint.

**Comment by krive001** [comment]
I hope you understand me.

**Comment by m-col** [comment]
Like this? https://github.com/qtile/qtile/issues/1757

**Comment by elParaguayo** [comment]
Apologies - I do appreciate there's a language barrier here and that English probably isn't your first language, but I don't understand what you mean by "the windows are getting out of the bar" and "so you can't even paint".

The bar does not paint the icons in the System Tray. Never has. Never will.

It works out how much space to keep free and then tells the apps where to place their icons but it does not draw the icons.

**Comment by krive001** [comment]
No applet ids can be listed here.
```
qtile cmd-obj  -o cmd -f windows | grep "'id'" |  awk '{print $2}'
```

Only for bar id and open windows.

**Comment by krive001** [comment]
Because it has to be part of the bar.

**Comment by elParaguayo** [comment]
It's not part of the bar. It's a separate window. That's the whole reason it's a problem and why other window managers also have issues.

**Comment by krive001** [comment]
```
Add Core.create_internal as single point to create Internal windows
This adds a method to the base `Core` class that can be implemented in
backends to create and return `Internal` windows. The method is not
abstract, instead raising `NotImplementedError` when called, so backends
could hypothetically not bother with `Internal`s but otherwise work.

All instances within libqtile of creating `Internal` windows (`Popup`,
`Bar` & `TreeTab`) manually put the window ID into `qtile.windows_map`.
Instead, `Internal` has been updated to do a `qtile.manage` with the new
window so that there is no need, and `Qtile.manage` remains The One Way
of managing windows.

`Core.create_internal` can be accessed the same independently on which
backend is being used, whereas the class method `Internal.create`
requires importing `Internal` directly, which is not backend-agnostic.
This commit replaces the latter with the former.

A nice side-effect of this is that the super ugly hack we used to test
`Popup` has been changed so that it is eval-ed (so still kinda ugly, but
not _as_ bad imo) within the running manager using the proper interface
for it.
```

**Comment by krive001** [comment]
Maybe @Tych0 understands.

**Comment by elParaguayo** [comment]
We're going round in circles. 

I need you to explain what issue you think that causes. You've said twice that the windows "get out" of the bar but you haven't yet explained what that means (at least, not in a way that I understand).

Also, note that the SysTray window is not an `Internal` window. It is an instance of `libqtile.backend.x11.window._Window`.

The Bar _is_ an `Internal` window but that commit doesn't really change anything. Whether or not the id is in `windows_map` is not relevant to the System Tray.

**Comment by krive001** [comment]
After:

QTILE_INTERNAL(CARDINAL) = 1
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0

Before:

WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
QTILE_INTERNAL(CARDINAL) = 1

**Comment by elParaguayo** [comment]
You've shown me that twice and neither time have you explained what the problem is. 

The two properties are the same. The order doesn't matter.

**Comment by krive001** [comment]
Which do you think is the bar and the internal window?
Why did it work well in the meantime?
It doesn't matter.
It belongs to Bar or not.

**Comment by krive001** [comment]
Everything has gone wrong since then.

**Comment by elParaguayo** [comment]
> Which do you think is the bar and the internal window? 

All of that is the bar which is an `Internal` window. The `QTILE_INTERNAL` property is just a flag that we add to windows to identify them as ones that Qtile has created. It has no relevance to how that window is rendered.

> Why did it work well in the meantime? 

That's a good question but it depends what you mean by "works well".

I believe the Systray worked before we merged the commit which introduced transparency for the Bar. With that commit, SysTray works with a fully transparent or fully opaque bar (at least it did in my testing). However, please note that I cannot replicate the overlapping icons issue so I can't verify the cause which is very frustrating.

> Everything has gone wrong since then.

What is "everything"? 

Show me something working before that commit and after it and I'll have something to try and work with.

**Comment by krive001** [comment]
That the internal window should continue to be handled by the bar.

**Comment by krive001** [comment]
And not the root window.

**Comment by elParaguayo** [comment]
> That the internal window should continue to be handled by the bar.

The `Internal` window (that you're seeing with `xprop`) ***is*** the Bar. The window that SysTray creates is handled by the widget. What do you want the Bar to do to it?


One other question for you: have you done a `git bisect` (or other test) where you can show the issue (whatever it is, I'm still not 100% certain) started? If so, post much, _much_ more detail (screenshots, logs,  explanations about what you're seeing etc.). One sentence replies aren't sufficient.

If you can't do that for me then, unfortunately, there's not much more I can do for you. I can't keep replying to this.

**Comment by krive001** [comment]
If you do not have a systemtray widget, you still have an internal window.

**Comment by krive001** [comment]
Commit:
```
Add Core.create_internal as single point to create Internal windows
This adds a method to the base `Core` class that can be implemented in
backends to create and return `Internal` windows. The method is not
abstract, instead raising `NotImplementedError` when called, so backends
could hypothetically not bother with `Internal`s but otherwise work.

All instances within libqtile of creating `Internal` windows (`Popup`,
`Bar` & `TreeTab`) manually put the window ID into `qtile.windows_map`.
Instead, `Internal` has been updated to do a `qtile.manage` with the new
window so that there is no need, and `Qtile.manage` remains The One Way
of managing windows.

`Core.create_internal` can be accessed the same independently on which
backend is being used, whereas the class method `Internal.create`
requires importing `Internal` directly, which is not backend-agnostic.
This commit replaces the latter with the former.
```

**Comment by elParaguayo** [comment]
I cannot say this enough times: the `Internal` window is the ***Bar***. It is nothing to do with the System Tray.

**Comment by krive001** [comment]
```
```
All instances within libqtile of creating `Internal` windows (`Popup`,
`Bar` & `TreeTab`) manually put the window ID into `qtile.windows_map`.
Instead, `Internal` has been updated to do a `qtile.manage` with the new
window so that there is no need, and `Qtile.manage` remains The One Way
of managing windows.

```

elParaguayo:
Yes. The bar is created via `create_internal` which calls `qtile.manage`.
https://github.com/qtile/qtile/blob/de39c977ead474a1300505b10f93a7c1d275478b/libqtile/backend/x11/core.py#L547-L555

The system tray window is not an internal window and its id is added to the map by the widget:
https://github.com/qtile/qtile/blob/802b82e5ab7b6cbad40d92996455c6fb2ea70c82/libqtile/widget/systray.py#L139

I'm still not seeing why you think any of this is relevant to how the SysTray is rendered. Perhaps you overestimate what `qtile.manage` does.
https://github.com/qtile/qtile/blob/3ac7442fad231cb6b94c2488ee4d7ec526b4cfe2/libqtile/core/manager.py#L582-L602

Which bit of that do you think is relevant for the System Tray?

krive001:
It doesn't matter if it's the root child or the bar manufacturer.
root-> bar-> window
root-> window

elParaguayo:
> It doesn't matter if it's the root child or the bar manufacturer. root-> bar-> window root-> window

Right. I'm stopping here. We're getting nowhere and I've taken too much time trying to get the information I need.

I appreciate that you're trying to help fix and issue but you're just not giving me enough information. Maybe someone else in the team can understand what you're trying to say.

Thank you for your effort and I'm sorry I can't help more at this point.

krive001:
It turned out to me here that the applet running on the systray doesn’t even belong to the bar.

```
 root = self.qtile.core._root.wid
        for wid in self.icons:
            self.conn.conn.core.ReparentWindow(wid, root, 0, 0)
        self.conn.conn.flush()

        del self.qtile.windows_map[self.wid]
        self.conn.conn.core.DestroyWindow(self.wid)
```

m-col:
Yes, there are multiple windows involved:

- the bar (`Internal`)
- the systray
- many icons

krive001:
The systray applet is below the bar and not on the bar.

m-col:
Could you post a screenshot?

m-col:
Is this the systray? 
image

krive001:
Yes.

m-col:
Just so I'm understanding correctly - is the screenshot with your change, or without? What difference does the change make?

krive001:
This main ..
I didn't change anything.

m-col:
What difference does the change make? The one from the PR

PR #2954, opened 2021-10-26
group.cmd_toscreen - do not toggle by default

m-col:
Many, many people complain about this behaviour, so let's return it back
to the old behaviour before then parameter was added. The docs can be
used to find it. Changelog mentions to change in case people come
searching for the behaviour again.


elParaguayo:
Balls. I guess this one conflicts because of the net widget merge I just did.

m-col:
Updated.

PR #2957, opened 2021-10-27
Use github's new issue form feature

m-col:
GitHub has a new feature for using a form as an issue template. This
takes advantage of that for our issue tracker.

Here is a link to the feature:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

--

For a live preview I put it on my fork's master branch. Feel free to make any issue to test it out: https://github.com/m-col/qtile/issues


m-col:
Yeah, makes sense. Perhaps we merge this as is and then we can add more inputs if we find the current configuration isn't good enough?

PR #2960, opened 2021-10-29
Add ref in docs to modded version of SNI widget

elParaguayo:
Should reduce questions on whether we support menus.

Also, a shameless bit of self promotion!


m-col:
Sounds good :)

PR #2961, opened 2021-10-29
Remove duplicate TOC entries from docs

elParaguayo:
Now that the widgets, layouts etc. have their own heading in the TOC, we don't need a specific "ref/index.rst" page.

Leaving it in results in duplicate entries when viewing the widgets page (for example).


m-col:
Thanks!

PR #2962, opened 2021-10-29
_SimpleLayoutBase: Pass client_position to _ClientList

jwijenbergh:
I noticed this while creating a custom layout with `_SimpleLayoutBase`. As I see it, `_SimpleLayoutBase` is just a wrapper around `_ClientList` so it should support the `client_position` argument too.

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -490,8 +490,8 @@ class _SimpleLayoutBase(Layout):
         client = self.focus_next(self.clients.current_client) or self.focus_first()
         self.group.focus(client, True)
 
-    def add(self, client, offset_to_current=0):
-        return self.clients.add(client, offset_to_current)
+    def add(self, client, offset_to_current=0, client_position=None):
+        return self.clients.add(client, offset_to_current, client_position)
 
     def remove(self, client):
         return self.clients.remove(client)
--- a/test/layouts/test_base.py
+++ b/test/layouts/test_base.py
@@ -1,0 +1,94 @@
+# Copyright (c) 2021 Jeroen Wijenbergh
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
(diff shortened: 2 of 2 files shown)


elParaguayo:
Thanks. Seems reasonable to me.

Can we add a test for this?

elParaguayo:
Also, what's your custom layout? We'd happily look at a PR if you think it's something worth sharing.

jwijenbergh:
> Thanks. Seems reasonable to me.
> 
> Can we add a test for this?

No problem :+1:  I will add one later

jwijenbergh:
> Also, what's your custom layout? We'd happily look at a PR if you think it's something worth sharing.

It's a bit niche so I'm not sure it would work inside Qtile. It's essentially recreating [exwm](https://github.com/ch11ng/exwm/wiki/Screenshots) in Qtile:

- The layout starts a maximized emacs instance
- Every window that is created is embedded into the emacs window
- The window positions are obtained from emacs itself
- Essentially windows are now just emacs buffers, it uses emacs lisp code to interact with qtile

jwijenbergh:
I have added some position tests to a new file called `tests_base.py` in layouts tests. I didn't want to duplicate a ton of tests of the position tests already done in the xmonad tests so these two should do for a simple change like this.

elParaguayo:
> I'm not completely sure what you expected tests of (as `_ClientList` positions is already tested with e.g. xmonad test code), but...
> 
> I have added some position tests to a new file called `tests_base.py` in layouts tests. I didn't want to duplicate a ton of tests of the position tests already done in the xmonad tests so these two should hopefully do for a simple change like this.

You've exposed an extra argument in `_SimpleLayoutBase.add` so I'd like to see that method being tested successfully.

Would expect it to be a very short test!

jwijenbergh:
> > I'm not completely sure what you expected tests of (as `_ClientList` positions is already tested with e.g. xmonad test code), but...
> > I have added some position tests to a new file called `tests_base.py` in layouts tests. I didn't want to duplicate a ton of tests of the position tests already done in the xmonad tests so these two should hopefully do for a simple change like this.
> 
> You've exposed an extra argument in `_SimpleLayoutBase.add` so I'd like to see that method being tested successfully.
> 
> Would expect it to be a very short test!

Is that not covered now by the test I added? 😅

elParaguayo:
Yes, it is - and it looks good.

I was going to ask you to rebase on to latest master to ensure all tests are run but you've done that too!

jwijenbergh:
Ah okay good. I was just making sure I wrote the correct tests here haha. Thanks for the merge.

PR #2967, opened 2021-10-30
Add aliases feature to lazy.spawncmd

### LINKED ISSUE #2966: Lazy spawncmd alias support
Hello, i had recently started using Qtile and i've stubled upon an issue that is really bothering me a lot.

Is there any chance i could make lazy.spawncmd run commands in my zshell so it would support all my precious aliases i spent hours to set up?

m-col:
This lets the user pass 'aliases' to `lazy.spawncmd()` which can be used
to let the command accept shortened strings and map those into larger
command lines before passing those into `Qtile.cmd_spawn()`.

Closes #2966

### DIFF
--- a/test/test_bar.py
+++ b/test/test_bar.py
@@ -24,6 +24,7 @@
 
 import os
 import tempfile
+from pathlib import Path
 
 import pytest
 
@@ -32,6 +33,8 @@ import libqtile.config
 import libqtile.confreader
 import libqtile.layout
 import libqtile.widget
+from libqtile.command.base import CommandError
+from test.helpers import Retry
 
 
 class GBConfig(libqtile.confreader.Config):
@@ -133,7 +136,7 @@ def test_draw(manager):
 
 
 @gb_config
-def test_prompt(manager):
+def test_prompt(manager, monkeypatch):
     assert manager.c.widget["prompt"].info()["width"] == 0
     manager.c.spawncmd(":")
     manager.c.widget["prompt"].fake_keypress("a")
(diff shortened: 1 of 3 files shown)


elParaguayo:
Is it possible to add a test e.g. an alias to `window.py`?

m-col:
Yup, test added

test/test_bar.py
```diff
@@ -133,7 +137,7 @@ def test_draw(manager):
 
 
 @gb_config
-def test_prompt(manager):
+def test_prompt(manager, monkeypatch):
```
elParaguayo, line 140: Why do we need `monkeypatch` here? It doesn't seem to be in your test.

m-col: Ah yes, good spot. I added it for a first draft as I was changing $PATH but then realised i didn't need to. Will remove!

elParaguayo:
LGTM.

Changelog entry and a rebase onto latest so we can run full test suite? Then should be good to merge.

sakarimov:
great feature for more integrated pipeline, waiting for implementation

elParaguayo:
Thanks. I'll merge ~if~ when these tests pass.

sakarimov:
how can we make this feature possible in our system?, do we have to make a custom Dict with aliases and command inside? or is it enough to make an aliases in our dotfiles?

m-col:
> how can we make this feature possible in our system?, do we have to make a custom Dict with aliases and command inside? or is it enough to make an aliases in our dotfiles?

You'll need to create a dict of aliases and pass that to `spawncmd`:
```py
aliases = {"ff": "firefox", "al": "some long command with args"}
...
lazy.spawncmd(aliases=aliases)
```

euliveiras:
the autocompletion for the added aliases is not working

m-col:
@euliveiras good spot, I added that here: https://github.com/qtile/qtile/pull/4465

PR #2973, opened 2021-11-01
Fix INCR clipboard transfers.

Golui:
Qtile did not comply with https://tronche.com/gui/x/icccm/sec-2.html#s-2.7.2
When transferring large files, this sometimes caused tools such as xclip
transferring in incremental mode (INCR) to stop responding to transfer
requests from other applications.

For completeness, proper target querying was implemented.

What is more, `XWindow.get_property` mistakenly claims a property does not exist if its value is empty, which seems to be the case for an uninitiated INCR transfer. It also does not support getting a property with deletion. As such, the method was slightly extended.

I believe further testing is needed to determine if no other clipboard managers break. I roughly tested with xclip, xsel and copyq and found no issues.

Obviously standard security considerations are still in place. No work on safe removal of sensitive data was carried out.

This still stands to be extended. for example target selection is very simple. Perhaps adding a hook to let users set a preference for the targets/MIME types.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -21,6 +21,7 @@
 
 import asyncio
 import contextlib
+import enum
 import os
 import signal
 import time
@@ -71,6 +72,39 @@ def get_modifiers() -> List[str]:
 
 class ExistingWMException(Exception):
     pass
+
+
+class SelectionStatus(enum.Enum):
+    empty = enum.auto()
+    targets = enum.auto()
+    awaiting = enum.auto()
+    incremental_transfer = enum.auto()
+    available = enum.auto()
+
+
+class Selection:
+    def __init__(self):
+        self.owner = None
+        self.binary_selection = b""
+        self.selection = ""
+        self.status = SelectionStatus.empty
+        self.targets = []
+        self._cur_tar = -1
+
+    @property
+    def current_target(self):
+        return self.targets[self._cur_tar] if self.targets and self._cur_tar >= 0 else "UTF8_STRING"
+
+    # For backwards compatibility
+    def __index__(self, idx):
+        return self.__dict__[idx]
+
+    def __repr__(self):
+        bin_sel = (self.binary_selection[:29] + b"...") if len(self.binary_selection) > 32 else self.binary_selection
+        sel = (self.selection[:29] + "...") if len(self.selection) > 32 else self.selection
+        return "<x11.Selection owner=({}) bin_sel=({}) sel=({}) status=({}) targets=({}) current_target=({})>".format(
+            self.owner, bin_sel, sel, self.status, self.targets, self.current_target
+        )
 
 
 class Core(base.Core):
@@ -130,8 +164,8 @@ class Core(base.Core):
         )
 
         self._selection = {
-            "PRIMARY": {"owner": None, "selection": ""},
-            "CLIPBOARD": {"owner": None, "selection": ""},
+            "PRIMARY": Selection(),
+            "CLIPBOARD": Selection()
         }
         self._selection_window = self.conn.create_window(-1, -1, 1, 1)
         self._selection_window.set_attribute(
@@ -143,11 +177,11 @@ class Core(base.Core):
 
         primary_atom = self.conn.atoms["PRIMARY"]
         reply = self.conn.conn.core.GetSelectionOwner(primary_atom).reply()
-        self._selection["PRIMARY"]["owner"] = reply.owner
+        self._selection["PRIMARY"].owner = reply.owner
 
         clipboard_atom = self.conn.atoms["CLIPBOARD"]
         reply = self.conn.conn.core.GetSelectionOwner(primary_atom).reply()
-        self._selection["CLIPBOARD"]["owner"] = reply.owner
+        self._selection["CLIPBOARD"].owner = reply.owner
 
         # ask for selection on start-up
         self.convert_selection(primary_atom)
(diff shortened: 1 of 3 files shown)


libqtile/backend/x11/core.py
```diff
@@ -73,6 +74,39 @@ class ExistingWMException(Exception):
...
+    # For backwards compatibility
+    def __index__(self, idx):
+        return self.__dict__[idx]
+
+    def __repr__(self):
+        bin_sel = (self.binary_selection[:29] + b"...") if len(self.binary_selection) > 32 else self.binary_selection
+        sel = (self.selection[:29] + "...") if len(self.selection) > 32 else self.selection
+        return "<x11.Selection owner=({}) bin_sel=({}) sel=({}) status=({}) targets=({}) current_target=({})>".format(
+            self.owner, bin_sel, sel, self.status, self.targets, self.current_target
+        )
```
m-col, line 107: Is this required, or was this just helpful for debugging?

libqtile/backend/x11/core.py
```diff
@@ -73,6 +74,39 @@ class ExistingWMException(Exception):
...
+        self.targets = []
+        self._cur_tar = -1
+
+    @property
+    def current_target(self):
+        return self.targets[self._cur_tar] if self.targets and self._cur_tar >= 0 else "UTF8_STRING"
+
+    # For backwards compatibility
+    def __index__(self, idx):
+        return self.__dict__[idx]
```
m-col, line 100: What kind of backwards compatibility is this needed for? Code people may have hooked to the two selection hooks?

libqtile/backend/x11/core.py
```diff
@@ -559,30 +593,89 @@ def handle_FocusOut(self, event) -> None:
...
+                sel.status = SelectionStatus.available
+        elif getattr(event, "owner", None) and sel.status in [SelectionStatus.empty, SelectionStatus.available]:
+            sel = self._selection[name] = Selection()
+            sel.status = SelectionStatus.targets
+            self.convert_selection(event.selection, "TARGETS")
+
+        if sel.status == SelectionStatus.available:
+            hook.fire("selection_change", name, sel)
+
+        hook.fire("selection_notify", name, sel)
```
m-col, line 651: I realised we passed equivalent args to the hooks even before this PR, but are they really useful? My concern with them is that they are not portable to other backends.

m-col (changes requested):
Overall this looks great, thank you for this.

Do you have a way to test that it is work e.g. using `xclip`? And if so, would it be feasible to use that in the test suite (i.e. in `test/backend/x11/test_xcore.py`) to make sure it keeps working? `xclip` and similar tools are generally available in the test environment so they can be used via subprocesses if necessary.

The CI is complaining as you have used tab characters in the changelog instead of spaces. Note that you may get a conflict from the changelog too as there have been some changes to the changelog on master.

Golui:
@m-col Thanks for the reply. I'm currently away, I'll fix it up during the next week. I'll comment on what I can right now.

libqtile/backend/x11/core.py
Golui: This is required for debugging, in my view. Without it, the whole selection as a binary string will get put in the log with the log level set to DEBUG. Clogs the console for large selections (such as images, etc.). This is due to hook.fire automatically logging the event.

libqtile/backend/x11/core.py
Golui: Yes, that is correct.

libqtile/backend/x11/core.py
Golui: I have yet to implement similar behaviour to wayland (first have to set it up, though). It may be useful if people want to make their own clipboard manager widgets that show rich previews. I suspect the targets (MIME types of the selection) will remain the same. The only thing would likely be that some statuses would become obsolete - but I would have to investigate this further. I am not familiar with wayland clipboard at all.

libqtile/backend/x11/core.py
```diff
@@ -73,6 +74,39 @@ class ExistingWMException(Exception):
...
+class SelectionStatus(enum.Enum):
+    empty = enum.auto()
+    targets = enum.auto()
+    awaiting = enum.auto()
+    incremental_transfer = enum.auto()
+    available = enum.auto()
+
+
+class Selection:
+    def __init__(self):
```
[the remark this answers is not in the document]
Golui: Just SelectionStatus or Selection as well?

libqtile/backend/x11/core.py
m-col: Alright, perhaps we should log a deprecation warning so that we can remove this in the future.

libqtile/backend/x11/core.py
m-col: OK sounds good

libqtile/backend/x11/core.py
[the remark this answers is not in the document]
m-col: Both - generally the core file is just for the core, and then xcbq is the misc helpers.

libqtile/backend/x11/core.py
m-col: I guess what we could do is try to pass the content of the selection in a generic way. What would that contain? Like you mentioned, I guess there would be a MIME type, and then the data. Currently the Wayland backend never fires these hooks but ironing this out would make it a good opportunity to do so.

Golui: This is a somewhat difficult question to answer. Under X, the contents of a selection might change significantly depending on the required MIME type. Simplest example to try is copy something from a spreadsheet editor (such as LibreOffice) and paste it into a plain text editor. As expected, you get a text representation of the table. However, fire up GIMP and paste that - you get an image. My understanding is that it's the application providing the clipboard that handles the conversion (as it's called in the X protocol) process.

Just getting all the possible MIME types is out of the question. Applications such as firefox provide ~4 different image types alone.

The laziest solution would be to provide a hook that fires when a selection is about to be updated (would be fired at a similar stage as the current `selection_notify`) that provides the MIME types as parameters. The receiver would then specify what MIME types they require and a callback. After that, we'd somehow pass a dictionary with MIME type - value pairs back to the requesting object. This has the advantage of remaining fairly similar to the current API.

Also, now that I mention it, I should probably cap the size of the selection. As it stands, qtile just happily takes the whole selection, essentially causing the memory footprint of the operation to double. (clipboard provider + qtile) Correct me if I'm wrong, but it doesn't seem that X allows one to check the size of the selection ahead of time. As such, I'd have to just cancel the INCR transfer. Of course, the size limit could be made user configurable. 256MB default should be more than enough.

A more flexible approach, though requiring a bit more work and design, would be just passing a list of the MIME types along with a backend independent "handle" object via a new hook. Perhaps a sort of `ClipboardManager` integrated directly into the `Qtile` class? That manager object given a MIME type would then make the actual request on demand, rather than storing the clipboard globally as it is done now. The advantage is that this is the lightest and probably the most universal solution (I glanced at the wayland clipboard; should work, but I have no solid proof as of yet). The caveat is that the API would essentially totally change.

PR #2976, opened 2021-11-02
Allow widgets to accept LazyCalls as callbacks

elParaguayo:
Users are already familiar with `lazy` calls from their keybindings. This PR allows the `lazy` calls to be bound to mouse callbacks for widgets.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           `group_name`. For the time being a log warning is in place and a
           migration is added. In the future `groupName` will fail.
         - Add `min/max_ratio` to Tile layout and fix bug where windows can extend offscreen.
+        - Add ability for widget `mouse_callbacks` to take `lazy` calls (similar to keybindings)
 
 Qtile 0.18.1, released 2021-09-16:
     * features
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -35,7 +35,9 @@ import subprocess
 from typing import Any, List, Tuple
 
 from libqtile import bar, configurable, confreader
+from libqtile.command import interface
 from libqtile.command.base import CommandError, CommandObject, ItemT
+from libqtile.lazy import LazyCall
 from libqtile.log_utils import logger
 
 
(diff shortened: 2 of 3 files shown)


m-col:
Nice, perhaps some docs + a changelog entry to go with it?

This also makes me wonder if the inverse should also be possible: using a regular function for a keybinding. `lazy.function` is lazy in that it evaluates the path and passes that. Regular functions would allow for no `qtile` arg, and simple partials could be used.

elParaguayo:
> Nice, perhaps some docs + a changelog entry to go with it?

Yes.

> This also makes me wonder if the inverse should also be possible: using a regular function for a keybinding. `lazy.function` is lazy in that it evaluates the path and passes that. Regular functions would allow for no `qtile` arg, and simple partials could be used.

That would probably do away with the need for `lazy.function` entirely, wouldn't it?

m-col:
> That would probably do away with the need for lazy.function entirely, wouldn't it?

No because it can be used for things other than `Qtile`, for example if you want to update a widget or something on a keybind with `lazy.widget['name'].function(func)`.

elParaguayo:
Might be better moving that to a discussion. 

I've never really seen the point of that last example. `func` is not a method of the widget, it just calls `func` with the widget as the first argument but we can already access the widget from `qtile.widgets_map`.

m-col:
While true, the latter implies internal code is part of the API, which is undocumented (and would be a _huge_ API surface) and we try at least a bit harder to keep the lazy API stable

elParaguayo:
> While true, the latter implies internal code is part of the API, which is undocumented (and would be a _huge_ API surface) and we try at least a bit harder to keep the lazy API stable

Excellent point.

jwijenbergh:
Does this also support the ```.when(...)``` check? I see no checking for cmd.check in this pr https://github.com/qtile/qtile/blob/20c1d03171faa95fab03d470fe5c39ddbeb16f8f/libqtile/core/manager.py#L380

elParaguayo:
It doesn't but no reason it couldn't.

jwijenbergh:
Looks good now thank you!

m-col:
Awesome idea. I see you've updated the widget docstring to reference lazy; do we need to update the lazy doc page or do you think it's too tangential?

m-col:
Cool sounds good, thanks for this!

PR #2978, opened 2021-11-03
Add coveralls to CI

### LINKED ISSUE #2054: Fix coveralls

elParaguayo:
This PR submits test results to coveralls.io. 

The job will only run on python 3.9 and only if the previous tasks (`pep8`, `docs` etc.) have all completed successfuly.

Fixes #2054

Pinging @ramnes and @m-col as I suspect we want to discuss this a bit.

This is just the basic uploading of test results but doesn't have any conditions set for when the test should report as failed.

FYI - the job ran successfully on my fork so you can see the results here: https://coveralls.io/builds/43991594

EDIT:
The result of this build is here: https://coveralls.io/jobs/89478238

Some points to note:
- The commit message is messed up
- It doesn't compare the coverage to the `master` branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,4 +79,22 @@ jobs:
               run: |
                 [ "$(grep -c -P '\t' CHANGELOG)" = "0" ]
                 tox
+            - name: Upload coverage data to coveralls.io
+              run: |
+                pip -q install coveralls
+                coveralls --service=github
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                COVERALLS_FLAG_NAME: ${{ matrix.python-version }}
+                COVERALLS_PARALLEL: true
 
+    coverage:
+        name: Finalize Coverage
+        needs: build
+        runs-on: ubuntu-20.04
+        steps:
+            - name: Coveralls Finished
+              uses: coverallsapp/github-action@master
+              with:
+                github-token: ${{ secrets.GITHUB_TOKEN }}
+                parallel-finished: true


elParaguayo:
Trying a second commit using the built in action, rather than installing coveralls from pip.

flacjacket:
I setup something similar for pywayland that would report coverage on all of the python versions that are being tested, and would use the built in action to finalize the coverage report once everything completed successfully: https://github.com/flacjacket/pywayland/blob/main/.github/workflows/ci.yml#L89 Not sure if that is something that is worth setting up for this or just the python 3.9 coverage is sufficient.

elParaguayo:
Thanks.

My thinking was that the same test suite is being run for all environments so capturing it once would be sufficient but happy to take a look at your version too.

elParaguayo:
@flacjacket I've added uploads for all python versions. 

One question for you: have you had any submit a PR from a fork yet? The two issues I can see here are the same as in my first post.

I'm not sure what "new coverage" means compared to "coverage rose" but I would have expected a "coverage rose" here given that we're now at 70% vs the previous 59% but I could be misunderstanding the labels.

ramnes (approved):
Cool! LGTM

ramnes:
> EDIT: it looks like we can get JSON data for the build and the last master so we could write a script that compares these coverage amounts and have it fail if there is a decrease of more than x%.

Codecov does this automatically IIRC

elParaguayo:
> > EDIT: it looks like we can get JSON data for the build and the last master so we could write a script that compares these coverage amounts and have it fail if there is a decrease of more than x%.
> 
> Codecov does this automatically IIRC

I've been searching about this issue and have seen people saying they've moved to Codecov for this reason. Worth a look?

elParaguayo:
Had a bit of a play with Codecov this morning and it looks quite nice.

I've got an example PR on my qtile-extras repo here: https://github.com/elParaguayo/qtile-extras/pull/4

You can see that Codecov is failing the PR because the PR does not meet the coverage level for the project as a whole and also decreases the overall coverage for the project. Limits can be defined for both of these tests, or they can be disabled.

I need to do a bit of reading on getting it to work with forked PRs next.

elParaguayo:
So Codecov does work with forked PRs. See this PR for an example: https://github.com/elParaguayo/qtile-extras/pull/6

You'll see that Codecov's checks currently fail as the PR reduces overall coverage.

elParaguayo:
Slightly annoying bug/design feature with Codecov. It seems to apply commits based on their date, rather than the order they're applied to a branch. So, a commit that pre-dates another commit but was merged afterwards will be placed before that commit.

Have posted on their support forum about this.

elParaguayo:
Support from Codecov hasn't been quick (don't blame them - I'm not a paying user). I'm also not sure how much I like the "full write access to the repo" permission.

elParaguayo:
I'm inclined to go back to coveralls. While CodeCov looks nice, there are some weird issues which don't seem right to me and, unfortunately, I've not had an answer to my query in their forums.

I'm using coveralls for my qtile-extras now. Admittedly, I won't get much in the way of PRs but it's really useful for helping me check my own coverage.

Coveralls also rejects PRs if they fail certain thresholds. For example:
https://github.com/elParaguayo/qtile-extras/pull/11

m-col:
Looks good. 

> This is just the basic uploading of test results but doesn't have any conditions set for when the test should report as failed.

Is there still this latter part that needs to be done? When this is merged, what changes will we see, if any, other than any rejected PRs?

elParaguayo:
If you merge this, you won't see any changes other than PRs and commits being uploaded to Coveralls.

We can then make settings on the coveralls site to allow their bot to post summaries on PRs plus have the PR be rejected if the overall coverage is < x% or if the coverage is reduced by y%.

We'd need to think about those thresholds and if they're even worth setting. While it would be nice to reject anything that reduces coverage, I'm not sure we're in a position where we can expect 100% coverage on all new additions.

m-col:
OK, sounds good. Yeah, maybe we could start off with a light touch, perhaps something that appears in the PR (bot message, or even better an indicator appearing in the CI panel above), that informs us of the net change in coverage.

elParaguayo:
These are the settings we can change:


We could turn "Leave comments" and try the "detailed" format but leave the thresholds blank for now.

m-col:
That status API-based indicator looks perfect https://github.com/lemurheavy/coveralls-public/issues/415

elParaguayo:
It does but I'm obviously missing something - how do we implement it?

When I do a PR on qtile-extras, the coveralls results are listed under the tests. Is that what we're talking about here?

m-col:
> When I do a PR on qtile-extras, the coveralls results are listed under the tests. Is that what we're talking about here?

Yeah that's what I meant - it looks pretty helpful. That said, I wonder if we can get the information without it necessarily passing/failing the PR, so that we can respond to it or ignore it how we see fit. What do you think would be a good configuration?

elParaguayo:
If we don't set thresholds then it will just show as passing which is probably fine.

m-col:
OK cool. We get the ` Coverage decreased (-4.2%) to 83.208% ` message anyway so as long as we get that information then we can act on it. I guess we can see how that goes?

Issue #2983, opened 2021-11-05
Widget.wttr: wrong option value `units='s'`


schragge:
Both [libqtile/widget/wttr.py](https://github.com/qtile/qtile/blob/7df60591fe6f8d2d9709a699856587ce96d76fc1/libqtile/widget/wttr.py#L70) and  [the widget documentation](http://docs.qtile.org/en/latest/manual/ref/widgets.html#wttr) reference `units='s'`  for imperial units. In fact, [wttr.in](https://github.com/chubin/wttr.in) doesn't have this option. Instead, it should be `units='u'` for USCS.

elParaguayo:
Thanks for this. I assume you're the same person that posted the question in the qtile-dev group.

First question: how did you manage to post an issue without using our template?!
Second question: are you able to submit a PR for this?

m-col:
> First question: how did you manage to post an issue without using our template?!

There is this link below the template list:

![image](https://user-images.githubusercontent.com/44536113/140506003-bc9d1a94-59da-4369-94c8-8beb4668a0ca.png)

We should probably rename the template to make it more clear that it should always be used

schragge:
> Thanks for this. I assume you're the same person that posted the question in the qtile-dev group.

No, I'm not the same person. I guess you mean Richard Wallace who's asking this question at [linux.org](https://www.linux.org/threads/37547) and at [LQ](https://www.linuxquestions.org/questions/linux-newbie-8/qtile-weather-widget-4175703117). I'm the person who answered him at LQ.

> First question: how did you manage to post an issue without using our template?! 

I'm not using Qtile at all. Moreover, I've never heard of Qtile until today. I just opened an issue here, on GitHub.

> Second question: are you able to submit a PR for this?

Done (#2984).

elParaguayo:
> > Thanks for this. I assume you're the same person that posted the question in the qtile-dev group.
> 
> No, I'm not the same person. I guess you mean Richard Wallace who's asking this question at [linux.org](https://www.linux.org/threads/37547) and at [LQ](https://www.linuxquestions.org/questions/linux-newbie-8/qtile-weather-widget-4175703117). I'm the person who answered him at LQ.

Ha ok. He also posted on qtile-dev and I replied there with the same solution. 

Thanks very much for helping him. You're more than welcome to push qtile questions this way next time.

> > Second question: are you able to submit a PR for this?
> 
> Done (#2984).

Amazing. Thank you.

PR #2984, opened 2021-11-05
wttr.py: Fix docstring with a wrong units value

schragge:
Change `units='s'` to `units='u'` according to [wttr.in](https://github.com/chubin/wttr.in) docs (issue #2983).

### DIFF
--- a/libqtile/widget/wttr.py
+++ b/libqtile/widget/wttr.py
@@ -67,7 +67,7 @@ class Wttr(GenPollUrl):
         (
             'units', 'm',
             "``'m'`` - metric, ``'M'`` - show wind speed in m/s, "
-            "``'s'`` - imperial"
+            "``'u'`` - USCS"
         ),
         (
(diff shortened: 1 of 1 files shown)


libqtile/widget/wttr.py
```diff
@@ -67,7 +67,7 @@ class Wttr(GenPollUrl):
         (
             'units', 'm',
             "``'m'`` - metric, ``'M'`` - show wind speed in m/s, "
-            "``'s'`` - imperial"
+            "``'u'`` - USCS"
```
elParaguayo, line 70: I don't know how familiar people are with the `USCS` acronym (I've not heard it before). Perhaps change it to "United States units".

elParaguayo: Last request: Can you squash those into a single commit and force push it? I can then take it from there.

Appreciate you doing this given you're not a Qtile user.

PR #2986, opened 2021-11-05
Increase coverage

elParaguayo:
As I've been looking at Coveralls/CodeCov I noticed lots of red highlighting where we don't have test coverage so I thought I'd fill in a few gaps!

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -766,6 +766,7 @@ class Internal(base.Internal, Window):
         self._outputs: List[Output] = []
         self._find_outputs()
         self._reset_texture()
+        self._group = None
 
     def finalize(self):
         self.hide()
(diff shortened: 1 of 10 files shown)


elParaguayo:
Hmm. So...
- All tests pass when I run `pytest` locally but not in a virtual environment
- Running on CI I get failures (which I suspect is because `extension_defaults` is somehow being picked up by these tests
- Running `tox` locally gives me different errors

:angry:

elParaguayo:
I'm inclined to go back to coveralls. While CodeCov looks nice, there are some weird issues which don't seem right to me and, unfortunately, I've not had an answer to my query in their forums.

I'm using coveralls for my qtile-extras now. Admittedly, I won't get much in the way of PRs but it's really useful for helping me check my own coverage.

elParaguayo:
Should me merged after #2978 (and once I've resolved the conflict!)

m-col:
Very nice, guess you want to rebase with coveralls enabled so you can see some nice big :heavy_check_mark: s ;)

m-col (approved):

m-col:
Do you want to merge this now or rebase and see what the increase in coverage is?

elParaguayo:
It'd be fun to see what the increase is!

Issue #2631, opened 2021-07-17
Key bindings in Wayland do not respect keyboard variant


flacjacket:
# Issue description

When running with a keyboard variant, in my case US Dvorak, even though other applications receive the correct key presses, qtile still continues to use the US keyboard layout. This includes all things I've tested so far that take keyboard input, including the keybindings as well as the prompt widget.

# Qtile version

Current git head 4f398189420b5020a4403a3556d4e4a98db38982

# Stack traces

N/A

# Configuration

N/A

yellowhat:
May I ask how did you configure the keyboard layout?

I have added the following to `config.py`:
```python
from libqtile import qtile
qtile.core.set_keymap(layout="gb", options="", variant="")
```
and seems to work.

The following should set up more options for the keyboard:
```python
qtile.core.set_keymap(layout="gb", options="caps:escape,altwin:swap_lalt_lwin", variant="")
```
It is swapping caps with escape but not alt and win.

Exporting these variables seems to work as expenced:
```console
export XKB_DEFAULT_LAYOUT="gb"
export XKB_DEFAULT_OPTIONS="caps:escape,altwin:swap_lalt_lwin"
```

m-col:
I use the exported variables, which is a more general way to configure keyboard inputs (the same variables will work in all wlroots-based compositors) and it means it's the original xkbcommon devs in control of how it works. The `set_keymap` method is there for use if you want (and like you've noticed, it works) but the reason it was added was for the `KeyboardLayout` widget.

Of course, there are still some improvements we need to make for keyboard layout handling to be finished, such as resolving this issue!

yellowhat:
I was starting `qtile` from the `tty`:

```console
export XKB_DEFAULT_LAYOUT="gb"
export XKB_DEFAULT_OPTIONS="caps:escape,altwin:swap_lalt_lwin"
qtile start -b wayland
```

Do this answer you question?

sant0s12:
Using

```python
from libqtile import qtile
qtile.core.cmd_set_keymap(layout="ch", options="", variant="")
```
doesn't work for me. Setting the environment variables does seem to work tho.

m-col:
Ah, not good. I've changes the way this could be configured (or more accurately, actually made it configurable) over on this PR: https://github.com/qtile/qtile/pull/2548

If anybody would be able to give that a try and see if everything works using the configuration approach it introduces, the feedback would be very helpful.

sant0s12:
[...]

PR #2990, opened 2021-11-06
Wayland: Multilanguage keybindings support

dydyamotya:
I'm from Russia, so I needed keybindings on two key layouts. Cause I'm trying to use Wayland, I've written code for Wayland. Look, pls, if it satisfy your requirements. I think these should work for X11 also. If so, I can do this work.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -135,6 +135,7 @@ class Core(base.Core, wlrq.HasListeners):
         self._pointers: list[inputs.Pointer] = []
         self.grabbed_keys: list[tuple[int, int]] = []
         self.grabbed_buttons: list[tuple[int, int]] = []
+        self._multilang_mapping = MultilanguageMapping()
         DataDeviceManager(self.display)
         self.live_dnd: wlrq.Dnd | None = None
         DataControlManagerV1(self.display)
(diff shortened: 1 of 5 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -119,6 +119,7 @@ def __init__(self):
         self.keyboards: List[keyboard.Keyboard] = []
         self.grabbed_keys: List[Tuple[int, int]] = []
         self.grabbed_buttons: List[Tuple[int, int]] = []
+        self.multilang_mapping = MultilanguageMapping()
```
m-col, line 122: Please could you rename this `self._multilang_mapping` as it's only used by this class.

libqtile/backend/wayland/core.py
```diff
@@ -864,3 +869,22 @@ def cmd_change_vt(self, vt: int) -> bool:
...
+                keysyms = tuple(keymap.key_get_syms_by_level(keycode, layout, 0)[0] for layout in range(layouts))
+                for keysym in keysyms:
+                    self.mapping[keysym] = keysyms
+
+    def get_keysyms(self, key: Union[config.Key, config.KeyChord]):
+        keysym = xkb.keysym_from_name(key.key, case_insensitive=True)
+        try:
+            return self.mapping[keysym]
+        except KeyError:
+            return (keysym, )
```
m-col, line 890:
```suggestion
self.mapping.get(keysym, (keysym, ))
```

libqtile/backend/wayland/core.py
```diff
@@ -864,3 +869,22 @@ def cmd_change_vt(self, vt: int) -> bool:
         if not success:
             logger.warning(f"Could not change VT to: {vt}")
         return success
+
+class MultilanguageMapping:
+    def __init__(self):
```
m-col, line 874: Please could you move this class to `keyboard.py`

libqtile/backend/wayland/core.py
```diff
@@ -748,17 +749,21 @@ def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
...
+        #keysym = xkb.keysym_from_name(key.key, case_insensitive=True)
+        keysyms = self.multilang_mapping.get_keysyms(key)
         mask_key = wlrq.translate_masks(key.modifiers)
-        self.grabbed_keys.append((keysym, mask_key))
-        return keysym, mask_key
+        for keysym in keysyms:
+            self.grabbed_keys.append((keysym, mask_key))
+        return keysyms, mask_key
 
     def ungrab_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
```
m-col, line 759: Please update the return type annotation here to reflect the multiple keysyms that could be returned. Same above with `grab_key`

m-col (changes requested):
Nice, this is looking great so thank you for the contribution! I've added some review comments inline but overall it's looking really good.

dydyamotya:
I've changed everything as you suggested.
Have looked also on X11 grab_key implementation, don't know if I'd like to implement it (If someone needs it on X11). Only if you insist :D
Also it should be noted, that my implementation works if XKBCOMMON_DEFAULT variables defined before qtile launching.

There was a discussion about shortcuts. Is it possible to link it here? Or link this PR to it? Sorry, I'm noob at these things.

ohadeytan:
Any news about this? I'm using a Hebrew layout. Currently I'm using [this](https://github.com/qtile/qtile/discussions/2453#discussioncomment-1555818) as a workaround.

m-col (commented):
Hey, sorry for forgetting about this. I've rebased and updated it all for you so you don't have to.

I also added another change to convert the style from a class to a function, because we don't need to keep any state around from this code so we can save some space by only getting the mappings as required (i.e. upon key grab).

Overall this looks good but unfortunately I can't test it as I don't have different keyboard layouts. Is that what is required to test it? @ohadeytan would you be willing to test it out too?

ohadeytan:
Hi @m-col, thanks for taking care of that. I tried that out and it didn't work for me, but I'm using X11, so it shouldn't work, isn't it? If it should, I will try to look into it more, maybe I missed some needed configuration.

m-col:
No I'm afraid this applies to Wayland only, sorry

m-col (approved):

m-col:
No reply on the related issue https://github.com/qtile/qtile/issues/2631, so happy to merge it as is. I've rebased it again and I'll enable it to auto-merge. Thanks for the submission! (and sorry for taking almost a year to merge it!)

Freemason-12:
Hey, why is this not merged? It seems like there's no issue with this PR.

PR #2991, opened 2021-11-06
Add Fedora dependencies for x11

dannycolin:


docs/manual/install/index.rst
```diff
@@ -49,19 +49,19 @@ Here are Qtile's core runtime dependencies and where available the package name
...
+CFFI_             python3-cffi        python3-cffi         Both backends, bars and popups
+X server          xserver-xorg        xorg-x11-server-Xorg X11 backend
+xcffib_           python3-xcffib      python3-xcffib       X11 backend
+wlroots_          libwlroots-dev                           Wayland backend (see below)
+pywlroots_        --                                       Wayland backend
+pywayland_        --                                       Wayland backend
+python-xkbcommon_ --                                       Wayland backend
+cairocffi_        python3-cairocffi   python3-cairocffi    Drawing on bars and popups (see below)
+libpangocairo     libpangocairo-1.0-0 pango                Writing on bars and popups
+dbus-next_        --                  python3-dbus-next    Sending notifications with dbus (optional)
```
elParaguayo, line 64: This doesn't build because you need to fix the following line. It should be:
```
================= =================== ==================== ==========================================
```

elParaguayo:
Thanks for submitting this.

@m-col: Interested in your thoughts here. We can't fit too many distros on this page so is it right to add another like this, or should we do a different page (e.g. on the distro specific page)?

m-col:
This table is more for illustration i.e. from the Ubuntu package name, one should be able to find the corresponding package they need for the their own distro. As Fedora doesn't have a Qtile package (as I understand it) then it would be good to add these dependencies to a distro-specific installation page for Fedora, replacing the incorrect(?) information on that page.

Does Fedora package the Wayland dependencies?

dannycolin:
> Does Fedora package the Wayland dependencies?

It most likely does. I was planning to add it after I had a x11 setup ready on my machine.

> This table is more for illustration i.e. from the Ubuntu package name, one should be able to find the corresponding package they need for the their own distro.

"should" is the keyword here. It wasn't obvious what was the equivalent of libpangocairo on Fedora that's why I opened this PR so others don't have to waste time searching for the right package especially that since there's no qtile package for Fedora they will need to install those dependencies manually.

I agree that it's not ideal to have all the distro in the table. Maybe a solution could be to have a table in an appendix that list packages for the different distros. Of course, adding a distro to the table is up to the contributors and not to you to maintain :)

m-col:
> It most likely does. I was planning to add it after I had a x11 setup ready on my machine.

OK, sounds good.

> It wasn't obvious what was the equivalent of libpangocairo on Fedora that's why I opened this PR so others don't have to waste time searching for the right package especially that since there's no qtile package for Fedora they will need to install those dependencies manually.

Cool, yeah it's great to have these listed. We do have a fedora-specific installation page: http://docs.qtile.org/en/latest/manual/install/fedora.html. I think it would be best to list the dependencies there for other fedora users, kinda like you said already:

> Maybe a solution could be to have a table in an appendix that list packages for the different distros. 

So perhaps mirror the table over on that page with fedora's deps? Of course, replacing the outdated information that is there currently.

dannycolin:
> e do have a fedora-specific installation page: http://docs.qtile.org/en/latest/manual/install/fedora.html. I think it would be best to list the dependencies there for other fedora users,

I agreed. I'll make the appropriated changes and send a new request for a review when it's ready ;)

Issue #2935, opened 2021-10-23
System tray icons overlap when changing status


hamensman:
Let's take the wifi (`nm-applet` ) and bluetooth (`blueman-applet`) system tray icons.

When I am connected to wifi and bluetooth is on, they look like this:
![Screenshot_2021-10-23_00-54-29](https://user-images.githubusercontent.com/76445071/138534476-1809b11e-930a-4034-ac9f-7e5029bbf876.png)
If I disconnect wifi and disable bluetooth, they **should** look like this:
![Screenshot_2021-10-23_00-54-58](https://user-images.githubusercontent.com/76445071/138534502-6908b925-bedc-4862-8562-2543bf0dbe94.png)
But what actually happens is they end up overlapping on top of each other, looking like this:
![Screenshot_2021-10-23_00-55-21](https://user-images.githubusercontent.com/76445071/138534522-cb6c390c-ae35-4408-9e42-b63c245205c2.png)

Right now, I would have to do a qtile restart if I wanted the icons to look as they should.

Using latest Qtile on Arch.

elParaguayo:
I've lost track of this one.

Can anyone confirm if the overlapping is still an issue as I can't recreate it.

ervinpopescu:
Yes, it still is.

CSpencerND:
Happens to me with volumeicon as well. I have my Systray inside a WidgetBox. Interestingly, collapsing and opening the WidgetBox corrects it without having to reload qtile.

CSpencerND:
Check this out. I just woke my laptop from sleep and the Systray inside the WidgetBox is overlapping the other widgets.
![2021-11-07_06-09](https://user-images.githubusercontent.com/61397897/140642635-8aa01478-cdfe-4cd2-b5eb-6f6350358e59.png)

elParaguayo:
Is volumeicon a systray app?

Can you post your config and a screenshot of the issue.

CSpencerND:
I have some errors in the logs.
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/libqtile/hook.py", line 380, in fire
    i(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/libqtile/core/manager.py", line 332, in cmd_reconfigure_screens
    self._process_screens()
  File "/usr/lib/python3.9/site-packages/libqtile/core/manager.py", line 313, in _process_screens
    scr._configure(self, i, x, y, w, h, grp)
[...]

PR #2992, opened 2021-11-08
#2935 System tray icons overlap when changing status

krive001:
Modify bar.py

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -223,7 +223,7 @@ class Bar(Gap, configurable.Configurable):
             # To preserve correct display of SysTray widget, we need a 24-bit
             # window where the user requests an opaque bar.
             if self.qtile.core.name == "x11":
-                depth = 32 if has_transparency(self.background) else self.qtile.core.conn.default_screen.root_depth
+                depth = 32 # if has_transparency(self.background) else self.qtile.core.conn.default_screen.root_depth
 
                 self.window = self.qtile.core.create_internal(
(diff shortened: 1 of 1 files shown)


elParaguayo:
That doesn't work on my system. It gives me a fully transparent background for icons.

The reason I wrote the line you've changed as it is to make the icons work with an opaque bar.

From what I've read elsewhere, this is an issue with the Xembed specification and other window managers have encountered the same issue.

I agree it's not great as it is, but I don't think this will get merged.

Also, FYI, if you remove the conditionality then you don't need the X11 block at all as the default depth for windows in X11 is 32 bit in our code.

krive001:
Because the icon is transparent, it should appear otherwise it will overlap. If there is alpha on the bar it is still transparent.

elParaguayo:
I'm not sure what you're trying to say. 

The "icon" is a window. The tray tells the apps which visual/depth they should provide. If the systray is using 24 bit colour depth then the apps should provide 24 bit windows. We are then able to paint the background to match the bar background. When using 32 bit colours, the painting of the background doesn't work which is why we get a fully transparent background for the icons. If you search for system tray transparency you'll see other window managers grappling with this issue and deciding not to implement overly complicated solutions.

The "solution" we implemented meant that icons should display correctly when bar was fully transparent or fully opaque. It sounds like some people still have issues with opaque bars.

I'm by no means an expert on this so please correct me if I'm wrong but we've looked at this for quite a while.

PR #3002, opened 2021-11-10
Add test for undocumented commands

elParaguayo:
To ensure that our docs are as helpful as possible, all exposed commands (e.g. those available via `lazy` etc.) should have a docstring.

This script will be run as part of the tox suite and will list any undocumented commands.

Currently, the script will not cause CI to fail (as there are a lot of undocumented commands). Once this is fixed, the `-w` flag should be removed so any new, undocumented commands will result in a failure.

This will be very helpful once I finish my fix for the `lazy` docs as this will generate list of commands automatically.

### DIFF
--- a/scripts/check_command_docs
+++ b/scripts/check_command_docs
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


scripts/check_command_docs
```diff
@@ -0,0 +1,106 @@
...
+                qtile = f.index("libqtile")
+            except ValueError:
+                continue
+
+            modules[f] = ".".join(path[qtile:])[:-3]
+
+        self.modules = modules
+
+    def check(self):
+        for path, modulename in self.modules.items():
```
m-col, line 64: It isn't immediately obvious what these methods do without really reading the code - perhaps add some docstrings so that in a year's time someone reading this understands what this is for?

m-col:
Just a thought - could the decorator that will be used to declare commands in your other PR simply raise an error when a docstring is absent? Then the tests will fail as a direct result.

elParaguayo:
Yes - but you don't get a warning of every undocumented command in one place. Plus, I didn't want tests to immediately fail as there are a lot of undocumented commands currently!

m-col:
Yeah, fair enough. I'm just a bit uneasy about using a script external to the main test suite for testing.

Another potential benefit of the runtime decorator error would be that it wouldn't be after-the-changes like the tests, but actually would be needed as part of the initial development of a new command. Anyway I'm probably worrying without cause so I'm happy for you to merge this :)

elParaguayo:
Alternatively, we don't merge this. I'll use it to document commands and, once, the decorators are done, we have them raise an error if they're undocumented. 

I'm fine with that approach.

m-col:
That would be great :)

PR #3003, opened 2021-11-10
Add default config to docs

elParaguayo:
Adds the `default_config.py` source to our docs so it _should_ display the right version depending on the version of the docs being browsed.

Haven't been able to confirm that though as building locally doesn't give me the option to select versions!

### DIFF
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -23,8 +23,7 @@ config, if it doesn't exist yet.
 Default Configuration
 =====================
 
-The `default configuration
-<https://github.com/qtile/qtile/blob/master/libqtile/resources/default_config.py>`_
+The :ref:`default configuration<default_config>`
 is invoked when qtile cannot find a configuration file. In addition, if qtile
 is restarted or the config is reloaded, qtile will load the default
 configuration if the config file it finds has some kind of error in it. The
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,10 +13,18 @@ Getting started
     manual/troubleshooting
     As a Wayland Compositor <manual/wayland>
     manual/commands/shell/index
+
+Reference
+=========
+
+.. toctree::
+    :maxdepth: 1
+
     manual/ref/extensions
     manual/ref/hooks
     manual/ref/layouts
     manual/ref/widgets
+    manual/config/default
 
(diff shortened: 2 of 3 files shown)


docs/manual/config/index.rst
```diff
@@ -234,3 +233,5 @@ session manager and use gnome as usual.
 
     without-dm
     gnome
+
+
```
m-col, line 237: Looks like you left in some empty lines

m-col:
Looks neat, I like it

docs/manual/config/index.rst
elParaguayo: Ha. Wha? That was too quick!

elParaguayo:
> Looks neat, I like it

I'm assuming the `literalinclude` directive picks the right file. Can't see why it wouldn't.

docs/manual/config/index.rst
m-col: I was curious to see if the PR was indeed resolving what was discussed in ORC earlier, and it is :)

m-col:
How come you can't tell locally?

elParaguayo:
Because I don't get the option to select docs version for different tagged releases.

m-col:
But don't you get it for master, with the literal?

elParaguayo:
I'm assuming that the docs are built by checking out the master and tagged versions and building against this. If that's correct then this should link to the correct file.

m-col:
Right, guess we can see how RTC does it!

elParaguayo:
Exactly, at the very worst we've still got a nice link to the config.

elParaguayo:
Doesn't work. Should have realised this too.

The docs for the tagged versions are built at that time so the new link to `default_config` won't appear in that version of the docs.

We could add links to the default config file for the latest tagged release to the default config page.

EDIT: ha. you just merged it!

m-col:
Yes, poor timing. Maybe it's fine for now, and we just release soon?

elParaguayo:
It's certainly fine for now. I can submit a separate PR with a link to the latest release (I'll see if there's a way to get the latest tagged version).

elParaguayo:
> Yes, poor timing. Maybe it's fine for now, and we just release soon?

This is the better answer!

PR #3010, opened 2021-11-12
Rename widgets with duplicate names by appending numeric suffixes

infmagic2047:
Widgets with duplicate names are mostly "forgotten" by Qtile, i.e. they do not appear in the list of widgets. The widgets can continue to work normally, but there is one caveat: qtile.finalize() only calls finalize on named widgets, so some widgets can be left un-finalized.

This patch makes sure every widget, registered in qtile.widgets_map or not, is finalized on shutdown.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -526,20 +526,51 @@ class Qtile(CommandObject):
             self.update_desktops()
 
     def register_widget(self, w: _Widget) -> None:
-        """Register a bar widget
-
-        If a widget with the same name already exists, this will silently
-        ignore that widget. However, this is not necessarily a bug. By default
-        a widget's name is just ``self.__class__.lower()``, so putting multiple
-        widgets of the same class will alias and one will be inaccessible.
-        Since more than one groupbox widget is useful when you have more than
-        one screen, this is a not uncommon occurrence. If you want to use the
-        debug info for widgets with the same name, set the name yourself.
         """
-        if w.name:
-            if w.name in self.widgets_map:
-                return
-            self.widgets_map[w.name] = w
+        Register a bar widget
+
+        If a widget with the same name already exists, the new widget will be
+        automatically renamed by appending numeric suffixes. For example, if
+        the widget is named "foo", we will attempt "foo_1", "foo_2", and so on,
+        until a free name is found.
+
+        This naming convention is only used for qtile.widgets_map as every widget
+        MUST be registered here to ensure that objects are finalised correctly.
+
+        Widgets can still be accessed by their name when using
+        lazy.screen.widget[name] or lazy.bar["top"].widget[name] unless there are
+        duplicate widgets in the bar/screen.
+
+        A warning will be provided where renaming has occurred.
+        """
+        # Find unoccupied name by appending numeric suffixes
+        name = w.name
+        i = 0
+        bars = set()
+        screens = set()
+        while name in self.widgets_map:
+            bars.add(self.widgets_map[name].bar)
+            screens.add(self.widgets_map[name].bar.screen)
+
+            i += 1
+            name = f"{w.name}_{i}"
+
+        if name != w.name:
+            msg = (
+                f"Widget was renamed to {name} in qtile.widgets_map. "
+                f"To bind commands, rename the widget or use lazy.widget['{name}']."
+            )
+
+            if w.bar in bars or w.bar.screen in screens:
+                msg += (
+                    " A duplicate widget was also found in the same bar or screen. "
+                    "If you wish to access the widget via lazy.bar[position].widget or lazy.screen.widget, "
+                    "you should rename the widget in your config."
+                )
+
+            logger.warning(msg)
+
+        self.widgets_map[name] = w
 
     @property
     def current_layout(self) -> Layout:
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks. The missing widgets in `widgets_map` always bugged me. 

I'll take a look at this.

libqtile/core/manager.py
```diff
@@ -92,6 +92,7 @@ def __init__(
 
         self.windows_map: Dict[int, base.WindowType] = {}
         self.widgets_map: Dict[str, _Widget] = {}
+        self._widgets: List[_Widget] = []
```
elParaguayo, line 95: I'd have a preference not to create a new variable. Instead, a better outcome would be to make sure `widgets_map` includes everything. This would have the added benefit of making every widget more easily available via IPC.

elParaguayo:
I've added some thoughts for handling this a bit differently. Let me know your thoughts.

infmagic2047:
I modified the patch to rename widgets automatically instead. The new names are generated by appending numeric suffixes like `foo_1`, `foo_2`, etc. I also set the name property on widgets to the new name, which is needed to make the IPC interface consistent.

elParaguayo:
We should also add a test for this.

m-col:
Another way to do this might be to keep lists in `widgets_map`, similar to how `mouse_map` works:
```py
self.mouse_map: Dict[int, List[Union[Click, Drag]]] = {}
```
Any operation that accesses it just iterates over whatever it finds.

This feels a bit less hacky IMO - widgets keep their name without any unexpected changes in behaviour. If a user wants to access a widget by name then it kinda does make sense for whatever command they are calling to be called on all matching widgets. If not, then they should explicitly rename their widgets themselves.

Thoughts?

elParaguayo:
Going to disagree here. I think you need a one to one match.

With your suggestion, you wouldn't be able to do `lazy.widget["textbox"].update("new text")` for example, as we wouldn't know which widget to update.

m-col:
The approach may cause other problems. Consider passing beyond `widget` in a command path: `lazy.widget["name"].bar`. The node becomes a list at `widget["name"]` and we can't un-list after that point because the `bar` node on those widgets might not be the same (if there is more than one "name" widget), so we end up with an implicit list of nodes... that is arguably unintuitive too :laughing:

elParaguayo:
I should remove the strikethrough on my earlier post: one-to-one is needed then!

m-col:
So @infmagic2047 ignore all that :)

+1 from me for the changes in their current form (plus a test).

infmagic2047:
I added a test for this.

elParaguayo:
The failed test throws out an interesting issue.

The test is failing because there are two `windowcount` widgets. With this PR, the second one is being renamed. However, the test is trying to access the widget through the parent bar which only contains one of those widgets.

What happens if we don't rename the widget but just use the `name_1` value as the key in the map? Would that work?

infmagic2047:
Hmm, I did not think about this use case.

> What happens if we don't rename the widget but just use the `name_1` value as the key in the map? Would that work?

Technically it is simple. Everything other than the qtile object uses `w.name` as the key in RPC interface, so such a change means these new names are only usable under the root object and nowhere else. While this can be considered an inconsistency, I believe it is not as bad as it seems. We can consider these generated names as "provided for convenience only", and users should provide unique names by themselves if they really need it.

If you are fine with the idea, I'll do the implementation soon.

elParaguayo:
Just dumping some thoughts here. Don't have an answer yet!

Looking at the CommandGraph map
graphviz
We can see widgets can be selected from root, screen and bar.

Root looks up the key in `widgets_map`:
https://github.com/qtile/qtile/blob/da93518b2ab924e803552decdb364d64de46088d/libqtile/core/manager.py#L811-L812

Screen looks up the widget name in the bars on that screen:
https://github.com/qtile/qtile/blob/da93518b2ab924e803552decdb364d64de46088d/libqtile/config.py#L413-L419

Bar looks up the widget name in the bar:
https://github.com/qtile/qtile/blob/da93518b2ab924e803552decdb364d64de46088d/libqtile/bar.py#L346-L349

Preventing duplicate names _within a bar_ is definitely required. As is having different keys in `widgets_map`.

Need to think about the best approach for that.

elParaguayo:
@m-col given that this PR should fix #3330, would be good to get your thoughts on my changes to this. In short, we ensure every widget is added to `widgets_map` (as these means every widget will be finalised). However, I'm not renaming the actual widget (i.e. we just give a unique key in the dict).

m-col:
Yeah I think that all makes sense.

> We can consider these generated names as "provided for convenience only", and users should provide unique names by themselves if they really need it.

Agreed.

elParaguayo:
I would like to do a test of the logging but I can't get `caplog` to work with our `TestManager` objects.

elParaguayo:
> I would like to do a test of the logging but I can't get `caplog` to work with our `TestManager` objects.

Think I know how to do this. Will try something later today. If it doesn't work, we should merge this as is.

elParaguayo (approved):
I've made edits directly and pushed these to the PR.

m-col (approved):
LGTM!

PR #3012, opened 2021-11-13
X11: set _NET_FRAME_EXTENTS window property

### LINKED ISSUE #3011: _NET_FRAME_EXTENTS not implemented
According to window manager [specification](https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html#idm45623293938880) `_NET_FRAME_EXTENTS` MUST be implemented, but it is not in Qtile 0.18.1. This might cause issues for some applications.

m-col:
This sets the _NET_FRAME_EXTENTS window property as per
https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html#idm45623293938880.

Fixes #3011

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -432,6 +432,8 @@ class XWindow:
         """
         This method is used only by the managing Window class.
         """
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks for doing this so quickly.

PR #3019, opened 2021-11-14
WL: outputs can skip to current screen if there's only one

m-col:
May as well save those function calls, considering single-output is
likely the most common setup.


elParaguayo (approved):
LGTM

m-col:
> Any reason why this shouldn't just get merged?

Nope this is fine as is

PR #3020, opened 2021-11-14
Add option to draw border/margin in Tile layout if there is only one window

Schroedingersdoraemon:
Dear qtile develop,  this patch added two options which could determine whether to draw border and margin when there is only one window in Tile layout

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -183,49 +183,55 @@ class Bar(Gap, configurable.Configurable):
         if isinstance(self.border_width, int):
             self.border_width = [self.border_width] * 4
 
+        self._initial_margin = self.margin[:]
+        self.struts = [0, 0, 0, 0]
+        self._add_strut = False
+
         self.queued_draws = 0
         self.future = None
 
     def _configure(self, qtile, screen):
-        Gap._configure(self, qtile, screen)
+        # We only want to adjust margin sizes once unless there's a new strut
+        if not self._configured or self._add_strut:
+            Gap._configure(self, qtile, screen)
 
-        if sum(self.margin) or sum(self.border_width):
+            if sum(self._initial_margin) or sum(self.border_width) or self._add_strut:
 
-            try:
-                # Check if colours are valid but don't convert to rgba here
-                if isinstance(self.border_color, list) and len(self.border_color) == 4:
-                    [rgb(col) for col in self.border_color]
+                try:
+                    # Check if colours are valid but don't convert to rgba here
+                    if isinstance(self.border_color, list) and len(self.border_color) == 4:
+                        [rgb(col) for col in self.border_color]
+                    else:
+                        rgb(self.border_color)
+                        self.border_color = [self.border_color] * 4
+                except (ValueError, TypeError):
+                    logger.warning("Invalid border_color specified. Borders will not be displayed.")
+                    self.border_width = [0, 0, 0, 0]
+
+                # Increase the margin size for the border. The border will be drawn
+                # in this space so the empty space will just be the margin.
+                self.margin = [m + b + s for m, b, s in zip(self._initial_margin, self.border_width, self.struts)]
+
+                if self.horizontal:
+                    self.x += self.margin[3] - self.border_width[3]
+                    self.width -= (self.margin[1] + self.margin[3])
+                    self.length = self.width
+                    if self.size == self.initial_size:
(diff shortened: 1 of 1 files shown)


libqtile/layout/tile.py
```diff
@@ -47,8 +47,10 @@ class Tile(_SimpleLayoutBase):
     defaults = [
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
+        ("border_on_single", 1, "Draw border if there is only one window."),
         ("border_width", 1, "Border width."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
+        ("margin_on_single", 1, "Draw margin if there is only one window."),
```
m-col, line 53:
```suggestion
        ("margin_on_single", True, "Whether to draw margin if there is only one window."),
```

And a similar change for `border_on_single`.

This will make it more clear that the option takes `True` or `False` rather than an `int`, which may make people think that the option configures the border size for when there is one window.

libqtile/layout/tile.py
```diff
@@ -155,14 +157,18 @@ def configure(self, client, screen_rect):
...
+                border_width = self.border_width
             client.place(
                 x,
                 y,
                 w - border_width * 2,
                 h - border_width * 2,
                 border_width,
                 bc,
-                margin=self.margin,
+                margin = 0 if (not self.margin_on_single and len(self.clients)==1) else self.margin,
```
m-col, line 171: Spaces are needed around `==` here and on line 160 - this will get caught by the CI which I have just enabled

m-col (changes requested):
Awesome, thanks for the submission!

I've put a couple comments inline and activated the CI so that should run now.

m-col:
One thing I forgot to say - please could you add an entry in the changelog for your new options. This way people can find out about the changes and use the feature!

Schroedingersdoraemon:
> One thing I forgot to say - please could you add an entry in the changelog for your new options. This way people can find out about the changes and use the feature!

Dear developer,  is that CHANGELOG in the repository? If so, does that mean adding an entry in the `qtile x.x.x, released xxxx-xx-xx`?

m-col:
Yep you got it exactly right. Those are the changes that will be part of the next release.

m-col (approved):

m-col:
Feel free to add a new answer to https://github.com/qtile/qtile/discussions/3014 referencing this PR in case somebody else finds your question seeking the same functionality :)

libqtile/layout/tile.py
Schroedingersdoraemon: Dear developer,  I was careless and added space near both `==` and `=`, which lead to the previous failed action.  I just found a tool named autopep8 in the gentoo repository, which helped me spot the space problem.  I guess there would be no problem here this time= =

m-col:
I didn't close this... not sure why github is saying I did

Schroedingersdoraemon:
> I didn't close this... not sure why github is saying I did

Thanks a lot dear developer, this is actually the first PR to my interested project.  Thank you again for your patient review : )

m-col:
The PR is showing no commits. Please could you push your changes back to your github? You may need to do `git push -f` to force it if it complains.

There is a conflict in CHANGELOG as a result of a different merge that was made this morning, so you will also need to rebase and resolve that conflict. To do that, make sure you have qtile/qtile as a remote i.e. "upstream" and then do `git pull --rebase upstream master`. You'll then need to edit CHANGELOG - make sure to keep both entries that are annotated, and then push those resolved changes. Let me know if you need any help!

Schroedingersdoraemon:
> The PR is showing no commits. Please could you push your changes back to your github? You may need to do `git push -f` to force it if it complains.
> 
> There is a conflict in CHANGELOG as a result of a different merge that was made this morning, so you will also need to rebase and resolve that conflict. To do that, make sure you have qtile/qtile as a remote i.e. "upstream" and then do `git pull --rebase upstream master`. You'll then need to edit CHANGELOG - make sure to keep both entries that are annotated, and then push those resolved changes. Let me know if you need any help!

Dear developer, I have resolved the conflict. Should I new a PR or just operate on this one?

m-col:
Great, you can just push to this one

m-col:
Just seen the other PR, we can use that one!

PR #3023, opened 2021-11-14
Add option to draw border/margin in Tile layout if there is only one window

Schroedingersdoraemon:
Dear developer, this patch adds two options which could determine whether to draw border and margin when there is only one window in Tile layout


m-col (approved):

m-col:
Great, thanks for this!

Schroedingersdoraemon:
Thanks a lot for your patient reply！(｡ì _ í｡)

Schroedingersdoraemon:
> Great, thanks for this!


Thanks a lot for your kindly patient reply！(｡ì _ í｡)

PR #3026, opened 2021-11-14
WL: support foreign toplevel management protocol

m-col:
This adds supports for the wlr_foreign_toplevel_management_v1 protocol.
This provides an interface through which clients such as status bars and
other WM utilities can control windows belonging to regular clients.
This provides similar functionality as seen in X via e.g. xdotool,
wmctrl to manipulate windows in some ways.

Requires pywlroots 0.14.10

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ MOCK_MODULES = [
     'wlroots.util.region',
     'wlroots.wlr_types',
     'wlroots.wlr_types.cursor',
+    'wlroots.wlr_types.foreign_toplevel_management_v1',
     'wlroots.wlr_types.keyboard',
     'wlroots.wlr_types.layer_shell_v1',
     'wlroots.wlr_types.output_management_v1',
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -33,6 +33,7 @@ from wlroots.wlr_types import (
     Cursor,
     DataControlManagerV1,
     DataDeviceManager,
+    ForeignToplevelManagerV1,
     GammaControlManagerV1,
     OutputLayout,
     PrimarySelectionV1DeviceManager,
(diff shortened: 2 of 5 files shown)


m-col:
Cool, my earlier CI failures were because I was trying to use the walrus operator and 3.7 is still supported. sad face.

Working well now though.

elParaguayo:
If I want to test this, what should I be looking for?

Also, any way we can actually add tests for this?

m-col:
We could use an existing tool such as https://git.sr.ht/~brocellous/wlrctl or a test client that just uses this protocol to listen to state changes and confirm that it's seeing the right information, and to make requests and confirm they have been made by Qtile. Ironically I've never written a wayland client and have no idea how to do so but pywayland [has an example](https://github.com/flacjacket/pywayland/blob/main/example/surface.py). I was considering adding one but thought I'd wait until something like wlrctl gets packaged for ubuntu.

m-col:
I played around with creating a test client using this protocol to make requests and to check window state. I think it could be useful for testing purpose beyond this protocol itself. I think some work in pywlroots is needed to make this more simple as my current client requires fetching the protocol's XML file from the wlroots-protocols gitlab instance which is pretty hacky. This is because there isn't a true client interface for this protocol, only wrappers around some wlroots functions. It's possible I've misunderstood client-side waylad code though.

Anyway, this is something I'm going to continue working on because an important project I have for the not-so-distant future is having proper testing of the wayland backend -- currently it's only tested using general backend-agnostic window management tests.

In the meantime, I'm happy for this to be merged as it is.

elParaguayo:
Tried this got an error:
```
>  wlrctl pointer move 100 100
Virtual Pointer interface not found!
```

`wlrctl keyboard type` and `wlrctl toplevel` seemed to work fine.

m-col:
Yep - we have not (yet) implemented the virtual pointer protocol. That latter one that does work is using the foreign toplevel protocol.

elParaguayo:
OK - so I can merge this then if everything else works?

m-col:
Yep :)

PR #3030, opened 2021-11-15
Add more configuration to the Max Layout

Borisd93:
Add more configuration's to the Max QTile Layout.

### DIFF
--- a/libqtile/layout/max.py
+++ b/libqtile/layout/max.py
@@ -31,13 +31,30 @@ class Max(_SimpleLayoutBase):
     commands to switch to next and previous windows in the stack.
     """
 
+    defaults = [
+        ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
+        ("border_focus", "#0000ff", "Border colour(s) for the window when focused"),
+        ("border_normal", "#000000", "Border colour(s) for the window when not focused"),
+        ("border_width", 0, "Border width."),
+    ]
+
+    def __init__(self, **config):
+        _SimpleLayoutBase.__init__(self, **config)
+        self.add_defaults(Max.defaults)
+
     def add(self, client):
         return super().add(client, 1)
 
     def configure(self, client, screen_rect):
         if self.clients and client is self.clients.current_client:
             client.place(
-                screen_rect.x, screen_rect.y, screen_rect.width, screen_rect.height, 0, None
+                screen_rect.x,
+                screen_rect.y,
+                screen_rect.width - self.border_width * 2,
+                screen_rect.height - self.border_width * 2,
+                self.border_width,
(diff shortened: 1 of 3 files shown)


libqtile/layout/max.py
```diff
@@ -31,6 +31,15 @@ class Max(_SimpleLayoutBase):
     commands to switch to next and previous windows in the stack.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if hasattr(self, "margin") is False:
```
elParaguayo, line 36: Properties should be set via `defaults` not here.

elParaguayo:
See #2726 - this is a better approach (but doesn't have borders included). You might want to start with that one and include borders too.

libqtile/layout/max.py
```diff
@@ -31,6 +31,16 @@ class Max(_SimpleLayoutBase):
     commands to switch to next and previous windows in the stack.
     """
 
+    __defaults__ = [
```
elParaguayo, line 34: This should be `defaults` not `__defaults__`.

libqtile/layout/max.py
```diff
@@ -31,6 +31,16 @@ class Max(_SimpleLayoutBase):
...
+        ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
+        ("border_color", "#0000ff", "Border colour(s) for the window"),
+        ("border_width", 1, "Border width."),
```
elParaguayo, line 37: Given that the code currently has `0` hard coded for the value, we should use that as the default here.

elParaguayo:
We need a test for this i.e. if the margin is set, the window size should reduce.

m-col:
Hola, could you add an entry to the changelog so that other users find your new features?

elParaguayo:
The default `border_width` is still showing as `1`. This needs to be changed to `0`. As it is, this change will add a border for users who just have `layout.Max()` in their configs which should not happen.

elParaguayo (approved):
Thanks. LGTM.

elParaguayo:
One small change needed. Can you then squash all the commits and force push this branch again.

Borisd93:
> One small change needed. Can you then squash all the commits and force push this branch again.

Clear the commit's?

elParaguayo:
There's still a typo that needs fixing and the commit history looks a bit off. The `SIGUSR1` commit shouldn't be part of this.

hamensman:
If this is of any use, here is a discussion I made here about adding border focus to the max layout: https://github.com/qtile/qtile/discussions/2891

elParaguayo (approved):
Made some changes and pushed to Borisd93's fork.

m-col (approved):
LGTM :)

PR #3044, opened 2021-11-21
Add ability for multiple timers in widgets

### LINKED ISSUE #3042: Cannot cancel multiple timers in widgets
The `timeout_add` method only expects one timer to be used by a widget at any point as, instead of returning a cancellable object, the method just assigns this to `self.future`:
https://github.com/qtile/qtile/blob/169842da985d21cfefeb1ed67013222c83407db0/libqtile/widget/base.py#L312-L318

This means that:
 - multiple timers cannot be cancelled
 - any widget with multiple timers will therefore not `finalize` correctly.

Posting this to remind me to make a fix!

elParaguayo:
`timeout_add` will now return a handle to the timer so that they can be cancelled individually. Timers are added to a list and
all scheduled timers in the list will be cancelled when `finalize` is called.

Closes #3042

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -152,6 +152,7 @@ class _Widget(CommandObject, configurable.Configurable):
             self.length_type = bar.STATIC
             self.length = length
         self.configured = False
+        self._futures: List[asyncio.TimerHandle] = []
 
     @property
     def length(self):
@@ -221,8 +222,8 @@ class _Widget(CommandObject, configurable.Configurable):
         pass
 
     def finalize(self):
-        if hasattr(self, 'future'):
-            self.future.cancel()
+        for future in self._futures:
+            future.cancel()
         if hasattr(self, 'layout') and self.layout:
             self.layout.finalize()
         self.drawer.finalize()
@@ -311,11 +312,14 @@ class _Widget(CommandObject, configurable.Configurable):
 
     def timeout_add(self, seconds, method, method_args=()):
         """
-            This method calls either ``.call_later`` with given arguments.
+            This method calls ``.call_later`` with given arguments.
         """
-        self.future = self.qtile.call_later(
+        future = self.qtile.call_later(
             seconds, self._wrapper, method, *method_args
         )
+
+        self._futures.append(future)
+        return future
 
     def call_process(self, command, **kwargs):
         """
@@ -325,7 +329,15 @@ class _Widget(CommandObject, configurable.Configurable):
         """
         return subprocess.check_output(command, **kwargs, encoding="utf-8")
 
+    def _remove_dead_timers(self):
+        """Remove completed and cancelled timers from the list."""
+        self._futures = [
+            timer for timer in self._futures
+            if not (timer.cancelled() or timer.when() < self.qtile._eventloop.time())
+        ]
+
     def _wrapper(self, method, *method_args):
+        self._remove_dead_timers()
         try:
             method(*method_args)
         except:  # noqa: E722
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -58,6 +58,7 @@ class PulseVolume(Volume):
         lib.pa_mainloop_quit(self.loop, 1)
         lib.pa_context_unref(self.context)
         lib.pa_mainloop_free(self.loop)
+        Volume.finalize(self)
 
     def connect(self):
         """
--- a/libqtile/widget/quick_exit.py
+++ b/libqtile/widget/quick_exit.py
@@ -42,7 +42,6 @@ class QuickExit(base._TextBox):
         self.is_counting = False
         self.text = self.default_text
         self.countdown = self.countdown_start
-        self.__call_later_funcs = []
 
         self.add_callbacks({'Button1': self.cmd_trigger})
 
@@ -50,8 +49,7 @@ class QuickExit(base._TextBox):
         self.is_counting = False
         self.countdown = self.countdown_start
         self.text = self.default_text
-        for f in self.__call_later_funcs:
-            f.cancel()
+        self.timer.cancel()
 
     def update(self):
         if not self.is_counting:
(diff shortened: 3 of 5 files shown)


m-col:
Was the change to `PulseVolume` only possible with the rest of the changes, or did you add that just while you were there?

It isn't clear to me why `QuickExit` tracks multiple countdowns. Could you comment to that?

elParaguayo:
> Was the change to `PulseVolume` only possible with the rest of the changes, or did you add that just while you were there?

I just noticed it wasn't calling the main `finalize` method but it also had `timeout_add` so it wouldn't always get cleaned up properly.

> It isn't clear to me why `QuickExit` tracks multiple countdowns. Could you comment to that?

I was thinking about that too. I can't immediately see a good reason for it. I think we can probably get away without the multiple ones so let me have a play.


I'm really noticing this more on some of my custom widgets. Trying to write some tests but don't have the control I need to cancel individual timers.

elParaguayo:
Added a test for this.

libqtile/widget/base.py
```diff
@@ -311,12 +312,17 @@ def calculate_length(self):
...
-            This method calls either ``.call_later`` with given arguments.
+            This method calls ``.call_later`` with given arguments.
         """
-        self.future = self.qtile.call_later(
+        future = self.qtile.call_later(
             seconds, self._wrapper, method, *method_args
         )
 
+        # Remove completed timers from the list and add this timer
+        self._futures = [timer for timer in self._futures if timer._scheduled]
```
m-col, line 322: Is there a way we can avoid using the private `_scheduled` property?

elParaguayo: Ha. Yeah, I didn't love that either. There are very few public properties. I don't think `cancelled` works.

elParaguayo: You could compare `timer.when()` with `self.qtile._eventloop.time()` and, if the timer has expired, remove it.

elParaguayo: @m-col any preference on this one?

elParaguayo:
Worth adding that, currently, `timeout_add` doesn't return anything but just sets the `self.future` attribute.

This means that I think there are some bugs in any code that does something like this:
``` python
timer = self.timeout_add(delay, func)
...
# This looks like it should generate an error
timer.cancel()
```
I suspect this is why `QuickExit` did that `__call_later_funcs` list.

elParaguayo:
@m-col Let me know your thoughts on this. Would be nice to merge this as timers are a bit broken without it.

libqtile/widget/base.py
m-col: Why doesn't `cancelled` alone work?

I remember there being a way to get the return value of scheduled functions, though the docs for this aren't revealing this to me now. Perhaps it wasn't `TimerHandle` or `Handle` but another way to schedule things. I'll keep looking.

m-col:
Just so I understand the use case properly - the idea is that, while almost all widgets will have some kind of polling mechanism that uses timers, some widgets (currently just custom ones) might benefit from on-the-fly timers being used? I do something similar for a widget in my config: https://github.com/m-col/qtile-config/blob/master/config.py#L398-L407

I'm wondering why this works just fine when maybe it shouldn't? Perhaps the timer is being replaced upon `button_press` but then the original timer is being re-setup when that new timer ends? In which case this just *happens* to work because the timer is only 1s long, but widgets might want e.g. a timer for a few hours but also wants the original timer working as normal in the interim. Is that right? And if so, is there any problem with using one of `Qtile`'s call later methods? I realise the widget timers essentially wrap those methods, but perhaps the wrapping is not necessary for these additional timers. Thoughts?

elParaguayo:
There are two reasons for this:
1) To fix the ugly hack in `QuickExit` and code like this which will never work:
https://github.com/qtile/qtile/blob/b19b3596c12d8f8c57a0d31eafffa29ea24ba5c6/libqtile/widget/mpris2widget.py#L148-L149

2) Yes, so you can have multiple timers.

I'm not sure exactly what happens when `self.future` is overwritten with a new timer but at least this change means `self.future` is less obfuscated than it currently is.

libqtile/widget/base.py
m-col, line 322: Rather than clearing a timer when the next is set up, could we clear it when it finishes i.e. as part of the `timer_setup` logic? Or would this require all users (widgets, including custom widgets) to remember and do it manually?

m-col:
> I'm not sure exactly what happens when self.future is overwritten with a new timer but at least this change means self.future is less obfuscated than it currently is.

I suppose my widget actually *is* broken because the previously timer is lost, but the errors appear on exit would only appear if one of these timers hasn't completed and I very recently clicked it, triggering the overlap. In the case where timers span a longer time period it would become an issue.

OK all understood :)

libqtile/widget/base.py
elParaguayo: This is more about stopping a memory leak. For example, the `Clock` widget creates a timer every second. Without this clean up, the `_futures` list would grow and grow.

m-col: Why could this not be done at the end of a task, rather than the start of the next?

elParaguayo: We could include it in the wrapper for the callback. Happy to make that change.

m-col: I think that would be better - it would ensure that futures are removed from the list even if there is no task following it.

elParaguayo: `cancelled` is only `True` if the timer is cancelled, not if it has been run.

m-col: Oh yeah, du doi. In that case yeah this works.

elParaguayo: The only downside to this is that if every timer is cancelled, the `_futures` list will grow and grow. Pretty unlikely scenario though!

m-col: What can cause a timer to get cancelled, other than when Qtile exits?

elParaguayo: Badly written code! 

I'm just pushing the change now.

m-col:
All looks great, good to merge pending CI :)

m-col:
Thanks!

PR #3045, opened 2021-11-21
Not a well-defined alpha color.

krive001:
100% — FF
99% — FC
98% — FA
97% — F7
96% — F5
95% — F2
94% — F0
93% — ED
92% — EB
91% — E8
90% — E6
89% — E3
88% — E0
87% — DE
86% — DB
85% — D9
84% — D6
83% — D4
82% — D1
81% — CF
80% — CC
79% — C9
78% — C7
77% — C4
76% — C2
75% — BF
74% — BD
73% — BA
72% — B8
71% — B5
70% — B3
69% — B0
68% — AD
67% — AB
66% — A8
65% — A6
64% — A3
63% — A1
62% — 9E
61% — 9C
60% — 99
59% — 96
58% — 94
57% — 91
56% — 8F
55% — 8C
54% — 8A
53% — 87
52% — 85
51% — 82
50% — 80
49% — 7D
48% — 7A
47% — 78
46% — 75
45% — 73
44% — 70
43% — 6E
42% — 6B
41% — 69
40% — 66
39% — 63
38% — 61
37% — 5E
36% — 5C
35% — 59
34% — 57
33% — 54
32% — 52
31% — 4F
30% — 4D
29% — 4A
28% — 47
27% — 45
26% — 42
25% — 40
24% — 3D
23% — 3B
22% — 38
21% — 36
20% — 33
19% — 30
18% — 2E
17% — 2B
16% — 29
15% — 26
14% — 24
13% — 21
12% — 1F
11% — 1C
10% — 1A
9% — 17
8% — 14
7% — 12
6% — 0F
5% — 0D
4% — 0A
3% — 08
2% — 05
1% — 03
0% — 00

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -66,7 +66,7 @@ def shuffle_down(lst):
         lst.append(c)
 
 
-ColorType = Union[str, Tuple[int, int, int], Tuple[int, int, int, float]]
+ColorType = Union[str, Tuple[int, int, int], Tuple[int, int, int, int]]
 ColorsType = Union[ColorType, List[ColorType]]
 
 
@@ -84,26 +84,27 @@ def rgb(x: ColorType) -> Tuple[float, float, float, float]:
 
         Which is returned as (1.0, 0.0, 0.0, 0.5).
     """
+    # logger.warning(f"x: {x}")
     if isinstance(x, (tuple, list)):
         if len(x) == 4:
             alpha = x[-1]
         else:
-            alpha = 1.0
-        return (x[0] / 255.0, x[1] / 255.0, x[2] / 255.0, alpha)
+            alpha = 255
+        return (int(x[0]), int(x[1]), int(x[2]), int(alpha))
     elif isinstance(x, str):
         if x.startswith("#"):
             x = x[1:]
         if "." in x:
             x, alpha_str = x.split(".")
-            alpha = float("0." + alpha_str)
+            alpha = 255 if int(alpha_str) >= int(100) else int(float("0." + alpha_str) * 255)
         else:
-            alpha = 1.0
+            alpha = 255
         if len(x) not in (6, 8):
             raise ValueError("RGB specifier must be 6 or 8 characters long.")
         vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
         if len(x) == 8:
-            alpha = int(x[6:8], 16) / 255.0
-        vals += (alpha,)
+            alpha = int(x[6:8], 16)
+        vals += (int(alpha),)
         return rgb(vals)  # type: ignore
     raise ValueError("Invalid RGB specifier.")
 
@@ -137,7 +138,7 @@ def remove_transparency(colour: ColorsType):
     Returns a tuple of (r, g, b) with no alpha.
     """
     def remove_alpha(col):
-        stripped = tuple(x * 255.0 for x in rgb(col)[:3])
+        stripped = tuple(int(x) for x in rgb(col)[:3])
         return stripped
 
     if isinstance(colour, (str, tuple)):


m-col:
Does this just convert the internal representation of colours from decimal to uint8? What effect does this have on Qtile's behaviour?

Do either of the backends need updating to reflect this change, or will they work as normal?

elParaguayo:
@krive001 You need to help us out if you want us to review this. What issue have you encountered that this PR solves?

krive001:
I reduced unnecessary operations in the function.
No other changes are required.
I haven't tried the python body.

krive001:
The alpha must not float.

m-col:
> The alpha must not float.

Why?

krive001:
The alpha can be an integer from 0 to 255.

krive001:
Opacity and alpha are not the same.
Opacity also affects the child, alpha only for that element.

m-col:
We use cairocffi which asks for floats for alpha: https://cairocffi.readthedocs.io/en/stable/api.html?highlight=set_source_rgba#cairocffi.Context.set_source_rgba

We also use wlroots, which does the same.

elParaguayo:
Correct. Opacity is a window setting, alpha is a colour value for anything drawn with that colour. 

What is the issue you're facing?

At the moment you haven't described a bug that needs fixing.

krive001:
255.0 not same 255
8bit =  256 not 256.0
Color bug.

m-col:
When does the function return 255.0?

krive001:
e.g. : return (x[0], x[1], x[2], alpha)

m-col:
Where? https://github.com/qtile/qtile/blob/master/libqtile/utils.py#L87-L108

krive001:
or  
```
       stripped = tuple(x * 255.0 for x in rgb(col)[:3])
```

m-col:
Oh I see - @elParaguayo this code does seem to convert the rgb back up to int8s. How come this is needed here but not in other uses of colours returned by rgb? https://github.com/qtile/qtile/blob/master/libqtile/utils.py#L139-L141

m-col:
It looks like `remove_transparency` is only used in one place in the codebase: https://github.com/qtile/qtile/blob/169842da985d21cfefeb1ed67013222c83407db0/libqtile/backend/x11/drawer.py#L188-L191

And it's providing int8s instead of floats to set_source_rgb: https://cairocffi.readthedocs.io/en/stable/api.html?highlight=set_source_rgba#cairocffi.Context.set_source_rgba

Shouldn't they remain as floats?

elParaguayo:
Because, when you pass that colour to `set_source_rgb` the values get divided by 255 so, if we don't do this, the colour will always be black.

I agree, it's a bit ugly. We should have some sort of structure for storing colours so it's not possible to parse it twice.

m-col:
Do they also divide it by 255 when they are decimals?

elParaguayo:
I think so - that's why I did it this way.

If you had some form of structure then we could check if colour was an instance of that structure and not process it any more if it was.

m-col:
Wouldn't the values be decimals when the transparency is _not_ removed, as that's the default output of `rgb`? But we don't have problems with that

elParaguayo:
No. It takes the user defined colour value, not the floats returned by `rgb`.

elParaguayo:
@krive001 again, can you please tell me what problem you think this is causing. 

I have no issue in saying my code's a bit ugly but it's used in very limited scenarios (to prevent colours having an alpha value when used in a 24 bit window as this can cause rendering issues) and I'm not sure what you're trying to fix.

I'm not saying there's not a bug, I just need your help in seeing it!

krive001:
It also deletes unnecessary.
There will be no alloc_color.
```
Error: rgb("#000000") -> (0, 0, 0, 1) .> #00000003
Good: rgb("#000000") -> (0, 0, 0, 255) -> #000000ff

```
depth can remain 32

elParaguayo:
Not quite. Alpha value is between 0 and 1.

krive001:
No alpha 0 to 255.

elParaguayo:
> No alpha 0 to 255.

No.
https://github.com/qtile/qtile/blob/169842da985d21cfefeb1ed67013222c83407db0/libqtile/utils.py#L105

That value is divided by 255 to give a value between 0 and 1. That is how the alpha value is stored here.

krive001:
%0% It needs to be rounded up

krive001:
I don't know what python does.

elParaguayo:
What needs to be rounded up? Are you saying if someone puts in an alpha value that you want to round down to 0 or up to 1 (yes, I'm using 0-1 range as that's what `rgb` outputs)?

elParaguayo:
> I don't know what python does.

With the greatest respect, if you don't know what python does, why are you submitting a pull request for us to review? That feels like a waste of my time.

m-col:
Your changes break borders and the bar under wayland: 
image

m-col:
Also X11: 
image

krive001:
`https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4`

elParaguayo:
Ok. Now we're getting somewhere - you're pointing out an an actual issue. That's something I'm happy to look at.

I'm interested though as I've tested on Wayland and not had any issues with that function (which was committed quite some time ago now).

Have you changed anything else?

elParaguayo:
> https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

That has absolutely nothing to do with qtile or cairocffi. Why do you think it's relevant here? m-col has shown you why we use 0-1 range.

m-col:
Are 8-digit hex colours not working?

elParaguayo:
Share your config as I'm using transparency on x11 and it works well for me. 

I also can't tell (from either screenshot) what issue you're referring to.

elParaguayo:
> Are 8-digit hex colours not working?

They should work fine. You can do `rgb("#00000080")` which will give you `(0.0, 0.0, 0.0, 0.5019607843137255)` which is close enough to 50% on alpha given floating point calculations.

elParaguayo:
Feels like we should close this. 

From what I can tell `rgb` is working correctly as it returns an alpha value in the 0-1 range which is what is required by cairocffi.

krive001:
~~~
(255, 255, 255) -> depth 24
(255, 255, 255, 255) -> depth 32
~~~
But I see you don't believe me.

elParaguayo:
> ```
> (255, 255, 255) -> depth 24
> (255, 255, 255, 255) -> depth 32
> ```
> 
> But I see you don't believe me.

What do you mean, "don't believe" you?!

We've told you that qtile/cairocffi needs colour and alpha values in the 0-1 range (see the earlier link) that is why `rgb` returns these values.

Is your concern that you can't pass a value in the range 0-255 for the alpha value?

If you do `rgb((255, 255, 255, 255))` you'll get `(1.0, 1.0, 1.0, 255)` which is invalid. We can introduce some error handling, or assume any alpha value greater than 1 should be treated as being in 0-255 range but your proposed changes are wrong.

krive001:
There opacity thinks, not alpha.

krive001:
Two alpga. #00000000, #000000.0 same

elParaguayo:
> There opacity thinks, not alpha.

Alpha value doesn't affect window opacity, just the amount of transparency/opacity of that colour.

> Two alpga. `#000000ff = #000000.100` same

No, they're not the same.
```
>>> rgb("#000000ff")
(0.0, 0.0, 0.0, 1.0)
>>> rgb("#000000.100")
(0.0, 0.0, 0.0, 0.1)
>>>
```

krive001:
And PR?

elParaguayo:
Your PR causes tests to fail and rendering issues (see m-col's screenshots).

It won't get merged. We've explained why.

I think we're done here.

krive001:
`100% = ff = 255`
Good night.

elParaguayo:
> `100% = ff = 255` 

Correct. But `#000000.100` is not 100%. It is 0.100 which is 0.1 i.e. 10%.

The code is working as it should.

krive001:
PR good color.

PR #3046, opened 2021-11-21
fix(layouts): Fix deletion of TreeTab Sections

muemmel:
- Instead of always deleting the first section, the targeted section is deleted
- Targeted section is also deleted from Root.sections so another section with the same name can be added

### DIFF
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -205,13 +205,15 @@ class Root(TreeNode):
         sec = self.sections[name]
         # move the children of the deleted section to the previous section
         # if delecting the first section, add children to second section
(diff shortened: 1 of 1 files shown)


m-col:
Will this change the behaviour for current users - and if they like the current behaviour will they need to change their configuration to retain the old behaviour? Or is the old behaviour inherently broken and therefore it's likely nobody uses it?

muemmel:
It will change the behavior in so far as make it work as advertised. I don't think it will affect configurations, since I'm not aware of workarounds for the unintended behavior.
In other words: As far as I can tell the old behavior was inherently broken.

m-col:
Thanks for the confirmation, I suspected as much. Thanks for the fix!

PR #3057, opened 2021-11-28
Keep a fullscreen window in the layout

jwijenbergh:
This is the most simple fix to #1482 I could come up with

Instead of removing a fullscreen window from the layout it keeps it both in the floating layout and tiled layout. This needs group to keep track of the windows that are in the tiled layouts as these layouts do not have a way to get the clients in the layout (`.clients` does not work for all layouts)

Note that this does not fix the behaviour that when a floating window is set to fullscreen and then non fullscreen again it will be tiled again.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -46,6 +46,7 @@ class _Group(CommandObject):
         self.label = name if label is None else label
         self.custom_layout = layout  # will be set on _configure
         self.windows = []
+        self.tiled_windows = set()
         self.qtile = None
         self.layouts = []
         self.floating_layout = None
@@ -217,6 +218,7 @@ class _Group(CommandObject):
             name=self.name,
             label=self.label,
             focus=self.current_window.name if self.current_window else None,
+            tiled_windows={i.name for i in self.tiled_windows},
             windows=[i.name for i in self.windows],
             focus_history=[i.name for i in self.focus_history],
             layout=self.layout.name,
@@ -236,7 +238,8 @@ class _Group(CommandObject):
             win._float_state = FloatStates.FLOATING
         if win.floating:
             self.floating_layout.add(win)
-        else:
+        if not win.floating or win.fullscreen:
+            self.tiled_windows.add(win)
             for i in self.layouts:
                 i.add(win)
         if focus:
@@ -256,7 +259,8 @@ class _Group(CommandObject):
                 or self.layout.focus_first()
                 or self.floating_layout.focus_first(group=self)
             )
-        else:
+        # Remove from the tiled layouts if it was not floating or fullscreen
+        if not win.floating or win.fullscreen:
             for i in self.layouts:
                 if i is self.layout:
                     nextfocus = i.remove(win)
@@ -269,6 +273,8 @@ class _Group(CommandObject):
                 or self.current_window
                 or self.layout.focus_first()
             )
+
+            self.tiled_windows.remove(win)
 
         # a notification may not have focus
         if hadfocus:
@@ -285,19 +291,27 @@ class _Group(CommandObject):
                 # already floating
                 pass
             else:
-                for i in self.layouts:
-                    i.remove(win)
-                    if win is self.current_window:
-                        i.blur()
+                # Remove from the tiled windows list if the window is not fullscreen
+                if not win.fullscreen:
+                    self.tiled_windows.remove(win)
+                    # Remove the window from the layout if it is not fullscreen
+                    for i in self.layouts:
+                        i.remove(win)
+                        if win is self.current_window:
+                            i.blur()
                 self.floating_layout.add(win)
                 if win is self.current_window:
                     self.floating_layout.focus(win)
         else:
             self.floating_layout.remove(win)
             self.floating_layout.blur()
-            for i in self.layouts:
-                i.add(win)
-                if win is self.current_window:
+            # A window that was fullscreen should only be added if it was not a tiled window
+            if win not in self.tiled_windows:
+                for i in self.layouts:
+                    i.add(win)
+                self.tiled_windows.add(win)
+            if win is self.current_window:
+                for i in self.layouts:
                     i.focus(win)
         self.layout_all()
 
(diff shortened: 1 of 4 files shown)


jwijenbergh:
Steps taken:

- Rebased with master
- Added tests for the tiled windows list and re-ordering of fullscreen windows that is now fixed
- Marked ready for review

libqtile/group.py
```diff
@@ -285,20 +292,29 @@ def mark_floating(self, win, floating):
...
         else:
             self.floating_layout.remove(win)
             self.floating_layout.blur()
             for i in self.layouts:
-                i.add(win)
+                # A window that was fullscreen should only be added if it was not a tiled window
+                if win not in self.tiled_windows:
+                    i.add(win)
                 if win is self.current_window:
                     i.focus(win)
```
m-col, line 315: We can save more checks here too (especially as `win not in self.tiled_windows` is O(n)):
```suggestion
            # A window that was fullscreen should only be added if it was not a tiled window
            if win not in self.tiled_windows:
		for i in self.layouts:
                    i.add(win)
	    if win is self.current_window:
		for i in self.layouts:
                    i.focus(win)
```

libqtile/group.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, name, layout=None, label=None):
         self.label = name if label is None else label
         self.custom_layout = layout  # will be set on _configure
         self.windows = []
+        self.tiled_windows = []
```
m-col, line 49: Looks like we don't need to preserve order of this list, so let's make it a `set` so that duplicate items are impossible.

m-col (changes requested):
Overall this looks great, thanks for this! I've added some review comments, just some tiny changes.

Could you also add a note to the changelog? We used to keep a bugfixes section there but I see now it hasn't been included since 0.16.0! Might be good to keep that up.

libqtile/group.py
```diff
@@ -285,20 +292,29 @@ def mark_floating(self, win, floating):
...
+                    self.tiled_windows.remove(win)
                 for i in self.layouts:
-                    i.remove(win)
-                    if win is self.current_window:
-                        i.blur()
+                    # Remove the window from the layout if it is not fullscreen
+                    if not win.fullscreen:
+                        i.remove(win)
+                        if win is self.current_window:
+                            i.blur()
```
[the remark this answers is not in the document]
jwijenbergh: makes sense!

libqtile/group.py
```diff
@@ -256,7 +259,8 @@ def remove(self, win, force=False):
                 or self.layout.focus_first()
                 or self.floating_layout.focus_first(group=self)
             )
-        else:
+        # Remove from the tiled layouts if it was not floating or in the tiled windows list
+        if not win.floating or win in self.tiled_windows:
```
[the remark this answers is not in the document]
jwijenbergh: It is, done.

libqtile/group.py
jwijenbergh: Also makes sense, this is much better

libqtile/group.py
jwijenbergh: yep!

jwijenbergh:
@m-col Thanks for the nice review. I have implemented all your suggestions. Could you give this a manual test? E.g. removing fullscreen windows, moving them to different groups and closing them, testing if they're always re-ordered correctly. This all worked for me but maybe you can find an edge case.

m-col:
It seems to work perfectly for all layouts! It should be possible to actually test every layout in one go. In `test/layouts/test_common.py` check out `test_cycle_layouts`. Something similar could be done to check that the window order stays fixed even when cycling across layouts and toggling fullscreen each time.

jwijenbergh:
> It seems to work perfectly for all layouts! It should be possible to actually test every layout in one go. In `test/layouts/test_common.py` check out `test_cycle_layouts`. Something similar could be done to check that the window order stays fixed even when cycling across layouts and toggling fullscreen each time.

Great idea. done!

m-col:
Perfect, thanks! It will be auto-merged if the CI is happy :)

jwijenbergh:
> Perfect, thanks! It will be auto-merged if the CI is happy :)

Thanks! I think you have to approve first though?

m-col:
Looks like this caused a test failure on another PR: https://github.com/qtile/qtile/runs/4872898400?check_suite_focus=true#step:10:1337

I'll restart those tests and see if they are flaky or something.

jwijenbergh:
> Looks like this caused a test failure on another PR: https://github.com/qtile/qtile/runs/4872898400?check_suite_focus=true#step:10:1337
> 
> I'll restart those tests and see if they are flaky or someth

> Looks like this caused a test failure on another PR: https://github.com/qtile/qtile/runs/4872898400?check_suite_focus=true#step:10:1337
> 
> I'll restart those tests and see if they are flaky or something.

Did restarting them work? I think it did right?

m-col:
The test error happened again on master (following an unrelated PR that passed): https://github.com/qtile/qtile/runs/4914974040?check_suite_focus=true#step:10:1333

If you run it a bunch of times locally, maybe you can get it to fail?

jwijenbergh:
> The test error happened again on master (following an unrelated PR that passed): https://github.com/qtile/qtile/runs/4914974040?check_suite_focus=true#step:10:1333
> 
> If you run it a bunch of times locally, maybe you can get it to fail?

I could try later but I'm not sure how this can be the case? There is no randomness involved.

m-col:
Lol, no idea. Sometimes things that require clients to update their own state can be a bit racey, and those are often fixed by doing `manager.c.sync()`. However, I don't think that would be the case here. Perhaps debug logs would be needed... I'm not actually sure how to get those from tests.

jwijenbergh:
> Lol, no idea. Sometimes things that require clients to update their own state can be a bit racey, and those are often fixed by doing `manager.c.sync()`. However, I don't think that would be the case here. Perhaps debug logs would be needed... I'm not actually sure how to get those from tests.

Sorry for the delay on this. I got to reproduce this locally (like 1-3 tests out of 300 sometimes) using (run in `test` directory):

```bash
#!/usr/bin/env bash
# Counter for how many times the test has failed
COUNTER=0

# Run tests 300 times
for i in {1..300}
do
    echo "RUNNING TEST $i"
    pytest -s test_group.py -k 'test_tiled_window_set' >logtemp 2>&1
    if [ $? -ne 0 ]; then
	((COUNTER=COUNTER+1))
	cat logtemp
    fi
done

echo "Failed $COUNTER times"
```

Not sure how to fix the test as of yet. I have asserted that the correct window is focused which it is, so it seems to be a problem only sometimes in `toggle_floating` during testing

PR #3060, opened 2021-11-29
Testing StatusNotifier widget

elParaguayo:
Adds two basic tests for the StatusNotifier widget:
1) Check that icon is added and removed from widget
2) Check that left-clicking the widget triggers the
   item's "Activate" method.

There's probably more parts of the widget that should be tested but this is a good start.

### DIFF
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -563,7 +563,7 @@ class StatusNotifier(base._Widget):
     defaults = [
         ('icon_size', 16, 'Icon width'),
         ('icon_theme', None, 'Name of theme to use for app icons'),
-        ('padding', 5, 'Padding between icons'),
+        ('padding', 3, 'Padding between icons'),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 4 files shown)


elParaguayo:
@m-col - while it's great that these tests work on CI, they don't work locally for me so I might play around a bit more.

It's to do with the `dbus` fixture...

libqtile/widget/statusnotifier.py
```diff
@@ -563,7 +563,7 @@ class StatusNotifier(base._Widget):
     defaults = [
         ('icon_size', 16, 'Icon width'),
         ('icon_theme', None, 'Name of theme to use for app icons'),
-        ('padding', 5, 'Padding between icons'),
+        ('padding', 3, 'Padding between icons'),
```
elParaguayo, line 566: This value wasn't overriding the `widget_defaults` value so I've made this match that value. Users shouldn't notice any difference.

libqtile/widget/statusnotifier.py
```diff
@@ -611,15 +611,17 @@ def draw(x=None):
 
     def find_icon_at_pos(self, x, y):
         """returns StatusNotifierItem object for icon in given position"""
-        xoffset = self.padding
-        if x < xoffset:
+        offset = self.padding
+        val = x if self.bar.horizontal else y
```
elParaguayo, line 615: These changes were needed as clicks weren't registering correctly on vertical bars.

m-col:
Very nice, thanks!

PR #3061, opened 2021-11-29
Thermal zone

x61s:
This is simple thermal zone monitoring widget. It allows you to get CPU temperature in Celsius degrees from /sys/class/thermal/ sensors.

### DIFF
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -14,7 +14,8 @@ class ThermalZone(base.ThreadPoolText):
     defaults = [
         ('update_interval', 2.0, 'Update interval'),
         ('zone', '/sys/class/thermal/thermal_zone0/temp', 'Thermal zone'),
-        ('format', '{temp}°C', 'Thermal zone display format'),
+        ('format', '{temp}°C', 'Display format'),
+        ('format_crit', '{temp}°C CRIT!', 'Critical display format'),
         ('hidden', False, 'Set True to only show if critical value reached'),
         ('fgcolor_crit', 'ff0000', 'Font color on critical values'),
         ('fgcolor_high', 'ffaa00', 'Font color on high values'),
@@ -31,18 +32,18 @@ class ThermalZone(base.ThreadPoolText):
         try:
             with open(self.zone) as f:
                 value = round(int(f.read().rstrip()) / 1000)
+                variables = dict()
+                variables['temp'] = str(value)
+                output = self.format.format(**variables)
                 if value < self.high:
                     self.layout.colour = self.fgcolor_normal
                 elif value < self.crit:
                     self.layout.colour = self.fgcolor_high
                 elif value >= self.crit:
                     self.layout.colour = self.fgcolor_crit
-                variables = dict()
-                variables['temp'] = str(value)
+                    output = self.format_crit.format(**variables)
             if self.hidden and value < self.crit:
                 output = ""
-            else:
-                output = self.format.format(**variables)
             return output
         except OSError:
             logger.exception('{} does not exist'.format(self.zone))
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -25,9 +25,9 @@ class ThermalZone(base.ThreadPoolText):
                 value = round(int(f.read().rstrip()) / 1000)
                 if value < self.high:
                     self.layout.colour = self.fgcolor_normal
-                elif value in range(self.high, self.crit):
+                elif value < self.crit:
                     self.layout.colour = self.fgcolor_high
-                elif value > self.crit:
+                elif value >= self.crit:
                     self.layout.colour = self.fgcolor_crit
                 variables = dict()
                 variables['temp'] = str(value)
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -5,9 +5,14 @@ from libqtile.log_utils import logger
 class ThermalZone(base.ThreadPoolText):
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ('update_interval', 2.0, 'Update interval for the thermal zone sensor'),
-        ('zone', '/sys/class/thermal/thermal_zone0/temp', 'Default thermal zone'),
+        ('update_interval', 2.0, 'Update interval'),
+        ('zone', '/sys/class/thermal/thermal_zone0/temp', 'Thermal zone'),
         ('format', '{temp}°C', 'Thermal zone display format'),
+        ('fgcolor_crit', 'ff0000', 'Font color on critical values'),
+        ('fgcolor_high', 'ffaa00', 'Font color on high values'),
+        ('fgcolor_normal', 'ffffff', 'Font color on normal values'),
+        ('crit', 70, 'Critical temperature level'),
+        ('high', 50, 'High themperature level'),
     ]
 
     def __init__(self, **config):
@@ -17,8 +22,15 @@ class ThermalZone(base.ThreadPoolText):
     def poll(self):
         try:
             with open(self.zone) as f:
+                value = round(int(f.read().rstrip()) / 1000)
+                if value < self.high:
+                    self.layout.colour = self.fgcolor_normal
+                elif value in range(self.high, self.crit):
+                    self.layout.colour = self.fgcolor_high
+                elif value > self.crit:
+                    self.layout.colour = self.fgcolor_crit
                 variables = dict()
-                variables['temp'] = str(round(int(f.read().rstrip()) / 1000))
+                variables['temp'] = str(value)
             return self.format.format(**variables)
         except OSError:
             logger.exception('{} does not exist'.format(self.zone))
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -20,6 +20,6 @@ class ThermalZone(base.ThreadPoolText):
                 value = str(round(int(f.read().rstrip()) / 1000)) + '°C'
             return value
         else:
-            logger.debug('zone {} does not exist'.format(zone))
+            logger.exception('{} does not exist'.format(zone))
             return 'err!'
 
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -3,6 +3,13 @@ from libqtile.log_utils import logger
 
 
 class ThermalZone(base.ThreadPoolText):
+    """Thermal zone widget.
+
+    This widget was made to read thermal zone files and transform values to
+    human readable format. You can set zone parameter to any standard thermal
+    zone file from /sys/class/thermal directory.
+    """
+
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
         ('update_interval', 2.0, 'Update interval'),
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -1,4 +1,3 @@
-from os import path
 from libqtile.widget import base
 from libqtile.log_utils import logger
 
@@ -15,11 +14,12 @@ class ThermalZone(base.ThreadPoolText):
         self.add_defaults(ThermalZone.defaults)
 
     def poll(self):
-        if path.isfile(self.zone):
+        try:
             with open(self.zone) as f:
                 variables = dict()
                 variables['temp'] = str(round(int(f.read().rstrip()) / 1000))
             return self.format.format(**variables)
-        else:
+        except OSError:
             logger.exception('{} does not exist'.format(self.zone))
             return 'err!'
+
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -23,4 +23,3 @@ class ThermalZone(base.ThreadPoolText):
         else:
             logger.exception('{} does not exist'.format(self.zone))
             return 'err!'
-
--- a/libqtile/widget/thermal_zone.py
+++ b/libqtile/widget/thermal_zone.py
@@ -32,19 +32,20 @@ class ThermalZone(base.ThreadPoolText):
         try:
             with open(self.zone) as f:
                 value = round(int(f.read().rstrip()) / 1000)
-                variables = dict()
-                variables['temp'] = str(value)
-                output = self.format.format(**variables)
-                if value < self.high:
-                    self.layout.colour = self.fgcolor_normal
-                elif value < self.crit:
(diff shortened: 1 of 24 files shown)


elParaguayo:
Thank you for the submission.

I'll be honest, I'm not 100% convinced by the need for this one. You could achieve something very similar by just doing this:
``` python
def get_temp(zone):
    def inner():
        with open(f'/sys/class/thermal/thermal_zone{zone}/temp') as f:
            value = str(round(int(f.read().rstrip()) / 1000)) + '°C'
        return value
    return inner

...
widget.GenPollText(func=get_temp(0))
```
I'll add some comments on the code separately.

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,29 @@
...
+        ('zone', '0', 'Default thermal zone'),
+    ]
+
+    def __init__(self, **config):
+        super().__init__("", **config)
+        self.add_defaults(ThermalZone.defaults)
+
+    def get_temp(self):
+        with open('/sys/class/thermal/thermal_zone' + self.zone + '/temp') as f:
+            value = str(round(int(f.read().rstrip()) / 1000)) + '°C'
```
elParaguayo, line 23: There needs to be some error handling, e.g. if the path is invalid or if there's an invalid value.

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,29 @@
...
+        return value
+
+    def poll(self):
```
elParaguayo, line 26: You could just put your `get_temp` code in `poll`.

test/widgets/test_thermal_zone.py
```diff
@@ -0,0 +1,14 @@
+import os.path
+from libqtile import widget
+
+def test_thermal_zone_getting_value():
+    # If thermal zone zero exists - get value
+    zone_zero = '/sys/class/thermal/thermal_zone0/temp'
+    zone_exists = os.path.isfile(zone_zero)
+    
+    if zone_exists:
```
elParaguayo, line 9: This test only works if there is a `thermal_zone0` on the CI machine.

It would be better to mock the contents of that file so that you can check the widget parses the contents correctly.

You could also test the error handling in your code (once you've added it).

test/widgets/test_thermal_zone.py
```diff
@@ -0,0 +1,14 @@
...
+        thermal_zone = widget.ThermalZone(zone='0')
+        output = thermal_zone.poll()
+        print(output)
```
elParaguayo, line 12: Remove this line. Presumably you had it for your own debugging.

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,22 @@
...
+        ('zone', '0', 'Default thermal zone'),
+    ]
+
+    def __init__(self, **config):
+        super().__init__("", **config)
+        self.add_defaults(ThermalZone.defaults)
+
+    def get_temp(self):
+        with open('/sys/class/thermal/thermal_zone' + self.zone + '/temp') as f:
+            value = str(round(int(f.read().rstrip()) / 1000)) + '°C'
```
elParaguayo, line 23: You might also want to allow users more customisation over the output of the widget, rather than hard coding the format.

libqtile/widget/thermal_zone.py
x61s: Done

libqtile/widget/thermal_zone.py
x61s: Done

test/widgets/test_thermal_zone.py
x61s: Done

test/widgets/test_thermal_zone.py
x61s: Done

libqtile/widget/thermal_zone.py
x61s: Done

test/widgets/test_thermal_zone.py
```diff
@@ -0,0 +1,16 @@
...
+    zone_file = tmp + '/sys/class/thermal/thermal_zone0/temp'
+    os.makedirs(os.path.dirname(zone_file), exist_ok=True)
+
+    with open(zone_file, "w") as f:
+        f.write("22000")
+
+    thermal_zone = widget.ThermalZone(zone=zone_file)
+    output = thermal_zone.poll()
+    assert output == '22°C'
+
```
m-col, line 16: Please could you remove the empty trailing line?

libqtile/widget/thermal_zone.py
m-col: Could you perhaps catch an exception and handle the error that way? The rationale of that approach is that exceptions will be rare, so we can get away with skipping the `path.isfile()` check because it will almost always be fine. If we wrap the `with open` block in a `try...catch` then we remove that (admittedly minor) overhead of the check every 2 seconds.

test/widgets/test_thermal_zone.py
x61s: Done

m-col:
I've commented a couple of nits but overall this looks good!

I feel an obvious feature would be to change the foreground/background colour of the widget when the temperature exceeds a user-configured threshold. Do you think this would be good to add? I would probably use this but i think my ideal configuration would be so that the background goes red when it goes above a certain temperate. Also possibly having a different string format for when it goes above the threshold e.g. `""` for normal (i.e. it's invisible) and the default when it goes above threshold.

What do you think?

elParaguayo:
> I feel an obvious feature would be to change the foreground/background colour of the widget when the temperature exceeds a user-configured threshold. Do you think this would be good to add? I would probably use this but i think my ideal configuration would be so that the background goes red when it goes above a certain temperate. Also possibly having a different string format for when it goes above the threshold e.g. `""` for normal (i.e. it's invisible) and the default when it goes above threshold.

I agree. I think that would be a really good addition here.

libqtile/widget/thermal_zone.py
x61s: Done

x61s:
> I've commented a couple of nits but overall this looks good!
> 
> I feel an obvious feature would be to change the foreground/background colour of the widget when the temperature exceeds a user-configured threshold. Do you think this would be good to add? I would probably use this but i think my ideal configuration would be so that the background goes red when it goes above a certain temperate. Also possibly having a different string format for when it goes above the threshold e.g. `""` for normal (i.e. it's invisible) and the default when it goes above threshold.
> 
> What do you think?

Done

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,37 @@
...
+        super().__init__("", **config)
+        self.add_defaults(ThermalZone.defaults)
+
+    def poll(self):
+        try:
+            with open(self.zone) as f:
+                value = round(int(f.read().rstrip()) / 1000)
+                if value < self.high:
+                    self.layout.colour = self.fgcolor_normal
+                elif value in range(self.high, self.crit):
```
m-col, line 28: A bit simpler perhaps:
```suggestion
                elif value < crit:
```

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,37 @@
...
+                    self.layout.colour = self.fgcolor_high
+                elif value > self.crit:
```
m-col, line 30: I think the case where `value == self.crit` is not covered - maybe change this `>` to `>=`

libqtile/widget/thermal_zone.py
x61s: Done

libqtile/widget/thermal_zone.py
x61s: Done

m-col:
Sorry for all the request! When it comes to new widgets its good to cover all bases so that people are interested in using it :)

The last thing I think would be great to have is a different `format` for the different temperature levels. This would be useful because it lets you functionally hide the widget normally, but then show it (with a format string that is not empty) when the temperature is high enough to possibly demand attention.

And lastly - perhaps add it the changelog? Then other users can be aware of the widget and use it too :)

m-col:
Actually I missed something - docs. The widget needs a docstring so that it can be rendered in the docs. @elParaguayo do you know if it's enough to add a docstring or does it need to be listed anywhere else to be included in the docs?

elParaguayo:
Ah, great point. As well as the docstring, you need to add an entry to the `widgets` dict in `libqtile.widget.__init__.py`,

x61s:
> Sorry for all the request! When it comes to new widgets its good to cover all bases so that people are interested in using it :)
> 
> The last thing I think would be great to have is a different `format` for the different temperature levels. This would be useful because it lets you functionally hide the widget normally, but then show it (with a format string that is not empty) when the temperature is high enough to possibly demand attention.
> 
> And lastly - perhaps add it the changelog? Then other users can be aware of the widget and use it too :)

No problem. I take it as a chance to practice in a Python and Qtile development. Thanks for helping with the process.

m-col:
Sorry, I wasn't clear about what I meant with the second `format` option. I meant that rather than the `hidden` option that you have added, there could just be something like this:
```py
         ('format_crit', '{temp}°C', 'Thermal zone display format when the temperature is critical'),
```
This gives users the flexibility to hide it when non-critical by using `format=""` (which is what I would do) but others may want to format it a different way. Does that make sense?

x61s:
> Sorry, I wasn't clear about what I meant with the second `format` option. I meant that rather than the `hidden` option that you have added, there could just be something like this:
> 
> ```python
>          ('format_crit', '{temp}°C', 'Thermal zone display format when the temperature is critical'),
> ```
> 
> This gives users the flexibility to hide it when non-critical by using `format=""` (which is what I would do) but others may want to format it a different way. Does that make sense?

I think hidden option is better and direct way to set widget visibility logic. And now user can set custom normal and critical formats to emphasize the situation. Or user can imitate the hidden option functionality with setting empty normal format if needed.

m-col:
> I think hidden option is better and direct way to set widget visibility logic. And now user can set custom normal and critical formats to emphasize the situation. Or user can imitate the hidden option functionality with setting empty normal format if needed.

Yeah good idea.

All looks great so once the tests have passed I will merge. Thanks again!

m-col (approved):

m-col:
You've got a failed test that might need updating:
```
  =================================== FAILURES ===================================
  _______________________ test_thermal_zone_getting_value ________________________
  
      def test_thermal_zone_getting_value():
          # Create temporary zone file
          tmp = '/var/tmp/qtile/test/widgets/thermal_zone'
          zone_file = tmp + '/sys/class/thermal/thermal_zone0/temp'
          os.makedirs(os.path.dirname(zone_file), exist_ok=True)
      
          with open(zone_file, "w") as f:
              f.write("22000")
      
          thermal_zone = widget.ThermalZone(zone=zone_file)
  >       output = thermal_zone.poll()
  
  test/widgets/test_thermal_zone.py:15: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  self = <libqtile.widget.thermal_zone.ThermalZone object at 0x7f4ed8de24c0>
  
      def poll(self):
          try:
              with open(self.zone) as f:
                  value = round(int(f.read().rstrip()) / 1000)
                  variables = dict()
                  variables['temp'] = str(value)
                  output = self.format.format(**variables)
                  if value < self.high:
  >                   self.layout.colour = self.fgcolor_normal
  E                   AttributeError: 'NoneType' object has no attribute 'colour'
  
  libqtile/widget/thermal_zone.py:39: AttributeError
```

libqtile/widget/thermal_zone.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, **config):
+        super().__init__("", **config)
+        self.add_defaults(ThermalZone.defaults)
+
+    def poll(self):
+        try:
+            with open(self.zone) as f:
+                value = round(int(f.read().rstrip()) / 1000)
+                variables = dict()
```
elParaguayo, line 35: Shouldn't the `except OSError` line be here? we're only reading the file once so none of the rest of this needs to be part of the `with` block.

elParaguayo (changes requested):

test/widgets/test_thermal_zone.py
```diff
@@ -0,0 +1,16 @@
...
+    # Create temporary zone file
+    tmp = '/var/tmp/qtile/test/widgets/thermal_zone'
+    zone_file = tmp + '/sys/class/thermal/thermal_zone0/temp'
+    os.makedirs(os.path.dirname(zone_file), exist_ok=True)
+
+    with open(zone_file, "w") as f:
+        f.write("22000")
+
+    thermal_zone = widget.ThermalZone(zone=zone_file)
+    output = thermal_zone.poll()
```
elParaguayo, line 15: This fails because `poll` tries to set `self.layout.colour` but `self.layout` is `None` as the widget has not been `_configure`d.

You could define a dummy class and set `widget.layout` to be that e.g.:
``` python
    class FakeLayout:
        pass

    with open(zone_file, "w") as f:
        f.write("22000")

    thermal_zone = widget.ThermalZone(zone=zone_file)
    thermal_zone.layout = FakeLayout()
```

libqtile/widget/thermal_zone.py
elParaguayo: Something like:
``` python
        try:
            with open(self.zone) as f:
                value = round(int(f.read().rstrip()) / 1000)
        except OSError:
            logger.exception('{} does not exist'.format(self.zone))
            return 'err!'

        variables = dict()
        variables['temp'] = str(value)
        output = self.format.format(**variables)
        if value < self.high:
            self.layout.colour = self.fgcolor_normal
        elif value < self.crit:
            self.layout.colour = self.fgcolor_high
        elif value >= self.crit:
            self.layout.colour = self.fgcolor_crit
            output = self.format_crit.format(**variables)
        if self.hidden and value < self.crit:
            output = ""
        return output
```

x61s: Done

test/widgets/test_thermal_zone.py
x61s: Done

test/widgets/test_thermal_zone.py
```diff
@@ -0,0 +1,20 @@
+import os
+from libqtile import widget
```
elParaguayo, line 2: You need a gap between these two imports.

x61s: Done

elParaguayo (approved):

PR #3067, opened 2021-12-01
Fix Bar.draw glitches

### LINKED ISSUE #3062: Widgets are causing bar to glitch when text updates
It's been a while I didn't pull qtile and I noticed this constant flicking caused but widget updates on my bars. I tried to record it but the flickering doesn't get recorded, it might be too fast. 

I don't have enough time to bisect this now and I believe someone might know what has changed. This flickering is very distracting and bad for my eyes. It seems that it only happens when there is a text resize, doesn't happen when just the text changes without the need of resizing. It looks like a glitch.

Happens on:
da93518b2ab924e803552decdb364d64de46088d

Didn't back happen on:
d8c00aae90d551fa2c326b2bd07d331d66ded04e

Tested under X11, wayland not tested.

elParaguayo:
Bug in code meant the the border drawing code was running every time `Bar.draw` was called. This PR fixes that bug and adds a flag so that borders are only drawn after the bar is configured.

Fixes #3062

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -189,11 +189,13 @@ class Bar(Gap, configurable.Configurable):
 
         self.queued_draws = 0
         self.future = None
+        self._borders_drawn = False
 
     def _configure(self, qtile, screen):
         # We only want to adjust margin sizes once unless there's a new strut
         if not self._configured or self._add_strut:
             Gap._configure(self, qtile, screen)
+            self._borders_drawn = False
 
             if sum(self._initial_margin) or sum(self.border_width) or self._add_strut:
 
@@ -498,7 +500,7 @@ class Bar(Gap, configurable.Configurable):
         self._resize(self.length, self.widgets)
 
         # We draw the border before the widgets
-        if self.border_width:
+        if any(self.border_width) and not self._borders_drawn:
 
             # The border is drawn "outside" of the bar (i.e. not in the space that the
             # widgets occupy) so we need to add the additional space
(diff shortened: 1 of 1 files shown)


libqtile/bar.py
```diff
@@ -189,11 +189,13 @@ def __init__(self, widgets, size, **config):
 
         self.queued_draws = 0
         self.future = None
+        self._borders_drawn = False
 
     def _configure(self, qtile, screen):
         # We only want to adjust margin sizes once unless there's a new strut
         if not self._configured or self._add_strut:
```
m-col, line 196: Would `self._configured` work or is the logic a bit different?

elParaguayo: Different. `_configured` is set after `self._configure` completes but we do need to draw the borders once after that point.

m-col (approved):
Great, LGTM

matheusfillipe:
This resolves the glitching for both using borders and not using the borders but i believe the borders should also be redrawn after showing them.
Like this if i use something like `lazy.hide_show_bar()` the borders are gone.

elParaguayo:
> This resolves the glitching for both cases of using borders and not using the borders but I think the borders should be also redrawn after showing them.
> 
> For example if I use something like `lazy.hide_show_bar()` the borders are gone.

Yes - I didn't cater for that scenario. Good spot!

matheusfillipe:
Ok, that fixed the borders disappearing.

elParaguayo:
From chat:
```
[19:43] < elParaguayo> | Any glitching with the borders?
[19:44] <      mattf> no glitching
[19:44] <      mattf> it looks really fine now both with and without them
```
If tests pass we should be good to merge.

PR #3071, opened 2021-12-06
Vertical textboxes

elParaguayo:
First attempt at vertical textboxes. Will keep as draft for now.

NB rotation is currently automatic i.e. you can't force the direction.

This is what the PR can do:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           When set to `False`, the keys are passed to the focused client. A key is never swallowed if the
           function is not executed, e.g. due to failing the `.when()` check.
         - Thermal zone widget.
+        - Allow TextBox-based widgets to display in vertical bars.
 
 Qtile 0.18.1, released 2021-09-16:
     * features
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -61,8 +61,6 @@ class Backlight(base.InLoopPollText):
 
     filenames = {}  # type: Dict
 
-    orientations = base.ORIENTATION_HORIZONTAL
-
     defaults = [
         ('backlight_name', 'acpi_video0', 'ACPI name of a backlight device'),
         (
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -31,6 +31,7 @@
 
 import asyncio
 import copy
+import math
 import subprocess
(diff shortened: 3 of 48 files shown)


m-col:
I did something similar but it didn't work :(

Looks awesome, you really are the MVP

m-col:
FWIW I don't think we need to make rotation direction configurable.... maybe we can if somebody asks for it but I can't imagine anybody would want anything other than what you've done

elParaguayo:
> I did something similar but it didn't work :(

I thought I was going crazy for a while. Went through so many failed attempts. Hilarious that the actual code need ended up being so short!

elParaguayo:
Let me add a test for some text widgets in a vertical bar (including import and config error widgets).

elParaguayo:
> FWIW I don't think we need to make rotation direction configurable.... maybe we can if somebody asks for it but I can't imagine anybody would want anything other than what you've done

I think that's right. The only scenario I wondered about was if someone had a very wide vertical bar, they might want horizontal text. Feels like an edge case so I didn't implement it.

elParaguayo:
If you're happy with the changes to the widgets (I removed the `orientations` line completely as this is inherited from the base classes) then I'll squash this into a single commit.

m-col:
Just given it a test drive and it works great! Feel free to squash-merge in the UI if that's easier :)

elParaguayo:
Needed to change commit message and add changelog entry.

Queued for auto-merge now.

gustavosbarreto:
> > FWIW I don't think we need to make rotation direction configurable.... maybe we can if somebody asks for it but I can't imagine anybody would want anything other than what you've done
> 
> I think that's right. The only scenario I wondered about was if someone had a very wide vertical bar, they might want horizontal text. Feels like an edge case so I didn't implement it.

That would be awesome, I could finally have the clock widget in the left sidebar without having to twist my head to look at it 🤣 

@elParaguayo Is there any possibility of this being implemented?

elParaguayo:
@gustavosbarreto are you asking about having a horizontal widget in a vertical bar?

gustavosbarreto:
> @gustavosbarreto are you asking about having a horizontal widget in a vertical bar?

Yes

elParaguayo:
Ok. I don't know about this one. There's a fair bit of work to implement this as this changes the logic that's used to layout widgets.

I suspect this won't be something that I work on at the moment.

PR #3072, opened 2021-12-06
Add `os.path.expanduser` to `utils.scan_files`

elParaguayo:
`images.Loader` uses `utils.scan_files` to locate files in folders.

If users have a theme_path including `~` then `scan_files` will not convert this to the correct path and the `Loader` will fail.

See https://github.com/elParaguayo/qtile-extras/issues/12#issuecomment-987137291 for an example.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -335,6 +335,7 @@ def scan_files(dirpath, *names):
     ['/wallpapers/w2.jpg', '/wallpapers/w3.jpg']})
 
     """
+    dirpath = os.path.expanduser(dirpath)
     files = defaultdict(list)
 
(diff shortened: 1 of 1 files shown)


libqtile/utils.py
```diff
@@ -338,6 +338,7 @@ def scan_files(dirpath, *names):
     files = defaultdict(list)
 
     for name in names:
+        dirpath = os.path.expanduser(dirpath)
```
m-col, line 341: Maybe stick it outside the loop so it's just done once?

elParaguayo: Sigh. This is why I shouldn't do this when I'm tired!

PR #3073, opened 2021-12-06
Fix double parsing of 'fmt' in Pomodoro widget

### LINKED ISSUE #3070: Pomodoro widget duplicates the prefix and suffix texts of the fmt key
```
widget.Pomodoro(
    fmt='prefix{}suffix',
),

widget.TextBox(
    fmt='prefix{}suffix',
    text='TEXTBOX'
),
```

### 1. Pomodoro widget

_expected to display:_

> **prefix**POMODORO**suffix**

_actual:_

> **prefixprefix**POMODORO**suffixsuffix**

### 2. TextBox widget (for example)

_expected to dispay = actual:_

>**prefix**TEXTBOX**suffix**

## Description:

- If there is prefix or suffix in **fmt** key at Pomodoro widget, It doubles the text of prefix or suffix

`qtile -v: 0.18.1`
`logs: nothing relevant`
- [x]  I have searched past issues to see if this bug has already been reported. *

elParaguayo:
By using a variable called 'fmt' which is also used by `base._TextBox`, the string formatting happense twice.

Closes #3070

Test didn't pick it up as it wasn't being run through a `manager` instance.

### DIFF
--- a/libqtile/widget/pomodoro.py
+++ b/libqtile/widget/pomodoro.py
@@ -29,7 +29,6 @@ class Pomodoro(base.ThreadPoolText):
     """Pomodoro technique widget"""
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
-        ("fmt", "{}", "fmt"),
         ("num_pomodori", 4, "Number of pomodori to do in a cycle"),
         ("length_pomodori", 25, "Length of one pomodori in minutes"),
         ("length_short_break", 5, "Length of a short break in minutes"),
@@ -171,4 +170,4 @@ class Pomodoro(base.ThreadPoolText):
         send_notification("Pomodoro", message, urgent=urgent)
 
     def poll(self):
-        return self.fmt.format(self._get_text())
+        return self._get_text()


libqtile/widget/pomodoro.py
```diff
@@ -171,4 +171,4 @@ def _send_notification(self, urgent, message):
         send_notification("Pomodoro", message, urgent=urgent)
 
     def poll(self):
-        return self.fmt.format(self._get_text())
+        return self.display_format.format(self._get_text())
```
m-col, line 174: Would this work without changing behaviour or configuration options?:
```suggestion
        self._text = self._get_text()
        return self.formatted_text
```
And dropping the redundant `fmt` option altogether from the child class.

elParaguayo: Yes - I realised this late last night (but only after I'd submitted this).

m-col:
Just trying to understand where the problem was coming from - before this change, one `str.format()` is happening in the widget's custom `poll()` method, but when is the other one happening?

elParaguayo:
The text is set via `self.update(text)`. This does `self.text = text` which calls this:
https://github.com/qtile/qtile/blob/ca1921ab7b5fa8eda6283f0bd437e9a9ac1c7aca/libqtile/widget/base.py#L380-L390

You can see that last call to `formatted_text` runs `.format` again.

m-col:
Ah yeah, I missed the `tick` part, thanks!

PR #3075, opened 2021-12-07
WL: ensure Drawers use Internal's current texture

m-col:
Wayland backend internal windows use a fixed-size buffer that gets
replaced when the window is resized. Drawers are saving the window's
texture upon creation but are not using new buffers when the window
creates a new one. Instead, this makes Drawers always get the buffer
from the window so that it's always drawing to the correct buffer.

This fixes the TreeTab panel which appears to be broken under Wayland.
It will also be causing problems where internal window resizing is
required.


m-col:
@elParaguayo Happy for me to merge this?

elParaguayo:
> @elParaguayo Happy for me to merge this?

No. _I_ want to do it!

PR #3076, opened 2021-12-07
Fix StatusNotifier reconfigure bug

elParaguayo:
StatusNotifier._configure returns immediately if the widget is configured, but this stops a new drawer object being created.

This, in turn, creates issues in `qtile-extras` when injecting decoration code.

See: https://github.com/elParaguayo/qtile-extras/issues/12

### DIFF
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -590,9 +590,6 @@ class StatusNotifier(base._Widget):
         )
 
     def _configure(self, qtile, bar):
-        if self.configured:
-            return
-
         if has_xdg and self.icon_theme:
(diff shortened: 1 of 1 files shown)


m-col (approved):
Yeah looks fine

m-col:
@elParaguayo @ramnes @tych0 FYI - I enabled auto-merge for the repository which requires setting a branch protection rule so that PRs can only be merged when the tests pass. This lets us select auto-merge while the tests are running so that the PR will merge if it remains unchanged and the tests all pass (like this PR). We just have to remember to update the rule when we change the CI i.e. drop py7 and add py10. It's in settings -> branches.

PR #3077, opened 2021-12-08
Add 'icon_only' shortcut in TaskList widget

elParaguayo:
Based on discussion in #3065.

Setting 'icon_only=True' disables all text output for TaskList widget. X11 only.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           When set to `False`, the keys are passed to the focused client. A key is never swallowed if the
           function is not executed, e.g. due to failing the `.when()` check.
         - Thermal zone widget.
+        - Add 'icon_only' option to TaskListwidget.
 
 Qtile 0.18.1, released 2021-09-16:
     * features
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -154,6 +154,11 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             'Icon size. '
             '(Calculated if set to None. Icons are hidden if set to 0.)'
         ),
+        (
+            'icon_only',
+            False,
+            'Hide all text and just show icons (X11 only)'
+        )
     ]
 
     def __init__(self, **config):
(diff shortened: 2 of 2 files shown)


m-col:
I'm getting a weird appearance from the widget, where the icons are cut off and some icons have a large space to the right hand side, as if `" "` is being rendered beside the icon. Do you see the same? VLC is doing this (it isn't very clear from the image).
out

elParaguayo:
No. Didn't get that at all. Let me try with VLC. 

Did you pass any parameters other than `icon_only=True`?

m-col:
Nope just that one, default config

m-col:
Oops, I wasn't using the default config. Perhaps it's the bar's height? Mine was with 28 height. What's yours?

elParaguayo:
Was 33 in that previous screenshot. 28 in this one:

m-col:
I'm actually seeing this same thing anyway, where the widget is trying to be too tall for the bar.

out

But that's unrelated to this PR.

However I am seeing something weird when just using icons. Windows for which icons cannot be determined (xterm here) show a small gap, which I guess is fine (though perhaps we could show a placeholder or something), but then windows that *do* have icons have a gap to the right hand side of the icon so the icon looks off centre. Here is a video trying to make this more clear:

https://user-images.githubusercontent.com/44536113/145276376-65989381-3c95-484a-85d6-bdf6352a94fd.mp4

elParaguayo:
Maybe `icon_only` isn't a great idea. How should we deal with apps that have no icon?

m-col:
Maybe we can just use the first character of the window name? It might not be very pretty but users can create more desktop files to give programs icons as they see fit.

elParaguayo:
Converted to draft. This needs work!

m-col:
This is some good work - do you want to keep this alive?

elParaguayo:
Yes. We should.

Crystal4276:
Bump to make this alive.

I would be able to remove this silly function:
```python
# Parsing : remove all text.
def txt_remove(text):
    return "
```

ervinpopescu:
Bump, really wanna remove this function too:
```py
def no_text(text):
    return ""
```

PR #3078, opened 2021-12-08
Add coveralls badge to readme

elParaguayo:
Draft for now. Let's see if the level is something we're happy to advertise! Think it should be around 70% now.


m-col:
74%, pretty good, happy to add this

PR #3080, opened 2021-12-09
Increase coverage (test status)

elParaguayo:
Same commits as per #2986 - just want to see if coveralls status message is triggered for new PRs.


elParaguayo:
Apparently resyncing coveralls can get the status working. Not seeing it listed here still. Must be missing something else.

PR #3083, opened 2021-12-09
More tests

elParaguayo:
Gradually adding more tests. Uploading bit by bit to see if the tests work.

Will mark it for merging once I've added a few more!

### DIFF
--- a/test/widgets/test_clipboard.py
+++ b/test/widgets/test_clipboard.py
@@ -1,0 +1,186 @@
+# Copyright (c) 2021 elParaguayo
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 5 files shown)


test/widgets/test_clipboard.py
```diff
@@ -0,0 +1,176 @@
...
+    config = minimal_conf_noscreen
+    config.screens = [
+        Screen(
+            top=Bar([widget], 10)
+        )
+    ]
+    manager_nospawn.start(config)
+
+    if manager_nospawn.backend.name != "x11":
+        pytest.skip("Test only available on X11.")
```
elParaguayo, line 48: I've assumed this widget should be X11 only but the tests would probably still work on Wayland as I'm just mocking the events.

m-col: Currently yeah

elParaguayo:
This khal_calendar test is very flaky. Need to reliably monkeypatch `dateutil.parser`...

If that doesn't work, I may try adding the module to the test environment.

PR #3088, opened 2021-12-11
Get coveralls status reports for PRs

elParaguayo:
Coveralls is not currently providing status reports for our PRs. Adding the `flag-name` may fix this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,14 +79,12 @@ jobs:
               run: |
                 [ "$(grep -c -P '\t' CHANGELOG)" = "0" ]
                 tox
-            - name: Upload coverage data to coveralls.io
-              run: |
-                pip -q install coveralls
-                coveralls --service=github
-              env:
-                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-                COVERALLS_FLAG_NAME: ${{ matrix.python-version }}
-                COVERALLS_PARALLEL: true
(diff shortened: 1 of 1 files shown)


m-col:
Sorry, I'm just lightly following along... just to confirm - coveralls is still looking good but you're just trying to get the right setup?

elParaguayo:
Coveralls is building fine - so we can view coverage on the site.

The issue is that it's supposed to be providing status messages in the PR (i.e. alongside the CI status) but it's not.

I don't know if this is a limitation of coveralls or not.

m-col:
Is there something very different between the way you're trying to set it up here vs on qtile-extras?

elParaguayo:
No. They're the same.

I wonder if the limitation is if we can't get status messages for PRs from forks?

Can you try another quick PR to qtile-extras, based on latest main branch and let's see if we get the status there.

elParaguayo:
That worked.

elParaguayo:
I think I've added the coveralls bot as a "collaborator" on qtile-extras but, from the docs, I think that's only needed for private repos.

m-col:
Could it be a permissions/secrets thing?

elParaguayo:
It looks like it's to do with parallel builds. 

I was just running py3.9 CI tests on qtile-extras and got the coveralls status on PRs. Having just updated CI to use 3.8, 3.9 and 3.10, the status updates have disappeared.

PR #3089, opened 2021-12-11
docs: Fix imports of autostart hook example

sealor:
I had problems to get the autostart example running because of a missing import statement. I hope my documentation improvement safes time for others.

http://docs.qtile.org/en/latest/manual/config/hooks.html?highlight=autostart#examples


m-col:
Great, thanks for this! It's exactly the kind of thing we miss because we're *too* familiar :laughing:

PR #3095, opened 2021-12-14
Two fixes for extensions

### LINKED ISSUE #3093: DmenuRun doesn't take colors from extension_defaults
This is split from https://github.com/qtile/qtile/issues/2730

Qtile version: 0.18.2.dev130+g3fe5e0d1
The problem starts in commit: https://github.com/qtile/qtile/commit/d7016e1dc27680437d89efb322d5e7db03e6b8fe

## test case
```
    EzKey("M-r", lazy.run_extension(extension.DmenuRun()), desc='dmenu run'),
    ...
extension_defaults = dict(
    background='#000000',
    foreground='#00ff00',
    selected_background='#00ff00',
    selected_foreground='#000000'
)
```

No errors

elParaguayo:
More like one fix and one improvement...

First commit changes timing of colour checks to ensure we don't set a value before `extension_defaults` from the config have been applied. This will close #3093.

Second commit replaces `CommandSet`'s use of `os.system` with `qtile.cmd_spawn`.

### DIFF
--- a/test/extension/test_base.py
+++ b/test/extension/test_base.py
@@ -35,7 +35,27 @@ parameters = [
 @pytest.mark.parametrize("value,expected", parameters)
 def test_valid_colours(value, expected):
     extension = _Extension(foreground=value)
+    extension._configure(None)
     assert extension.foreground == expected
+
+
+def test_valid_colours_extension_defaults(monkeypatch):
+    defaults = {
+        "foreground": "00ff00",
+        "background": "000000",
+        "selected_foreground": "000000",
+        "selected_background": "00ff00",
+    }
+    extension = _Extension(foreground="0000ff")
+
+    # Set defaults after widget is created to mimic behaviour of extension being
+    # initialised in config.
+    monkeypatch.setattr(_Extension, "global_defaults", defaults)
+    extension._configure(None)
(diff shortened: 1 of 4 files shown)


test/extension/test_base.py
```diff
@@ -35,6 +35,7 @@
 @pytest.mark.parametrize("value,expected", parameters)
 def test_valid_colours(value, expected):
     extension = _Extension(foreground=value)
+    extension._configure(None)
     assert extension.foreground == expected
```
m-col, line 39: Perhaps this test should also test that `extension_defaults` are being used?

elParaguayo: Yes, but there's no really nice way to do this.

Ideally, you'd run it in a manager instance but extensions aren't `CommandObject` so they're not visible via IPC.

We could just do a case where colours are set via `_Extension.global_defaults` (which is what `extension_defaults` are ultimately set).

m-col: OK yeah it looks fiddly without a huge amount of benefit. I'm pretty happy with how this is now.

m-col (approved):
LGTM, if you don't have anything else to add then feel free to merge :)

PR #3096, opened 2021-12-14
Add Python 3.10 to tox and github actions

m-col:
This adds a py310 tox env which is used for all test envs (i.e. mypy, pep8 etc). py39 is then only used for the pytest run. No envs are removed, because pypy only supports Python 3.7 so in the meantime it might be best to continue support for pypy and py37 while we can.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -60,7 +60,7 @@ deps =
     flake8-tidy-imports
     pep8-naming
 commands =
-    flake8 {toxinidir}/libqtile {toxinidir}/bin/ {toxinidir}/test
+    flake8 {toxinidir}/libqtile {toxinidir}/bin/ {toxinidir}/test --exclude=test/configs/syntaxerr.py,**/_ffi*.py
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -93,7 +95,8 @@ commands =
     mypy -p libqtile
     # also run the tests that require mypy
     python3 setup.py -q install
-    python3 -m pytest -W error -- test/test_check.py test/test_migrate.py
+# py310 currently fails with -W error, see: https://gitlab.gnome.org/GNOME/pygobject/-/issues/476
+    python3 -m pytest -- test/test_check.py test/test_migrate.py
```
elParaguayo, line 99: Do we get the same `ImportWarning` here, even if we're only running `test_check` and `test_migrate`?

m-col: Yeah it happens at the very beginning during the imports. I did test it without this change so can confirm.

PR #3099, opened 2021-12-15
fix #3094

### LINKED ISSUE #3094: Widget for updatable image
I don't want to reinvent the wheel, so will check before.
I use GenPollText for my keyboard layout indicator but instead I want to see the flag (image). As I change layout the image should be changed. Do Qtile has such widget or proper way to do that?

zordsdavini:
Added to Image widget update functionality. Should fix #3094 issue

### DIFF
--- a/libqtile/widget/image.py
+++ b/libqtile/widget/image.py
@@ -46,6 +46,9 @@ class Image(base._Widget, base.MarginMixin):
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
+        self._update_image()
+
+    def _update_image(self):
         self.img = None
 
         if not self.filename:
@@ -94,3 +97,8 @@ class Image(base._Widget, base.MarginMixin):
             return self.img.width + (self.margin_x * 2)
         else:
             return self.img.height + (self.margin_y * 2)
+
+    def cmd_update(self, filename):
+        self.filename = filename
+        self._update_image()
+        self.draw()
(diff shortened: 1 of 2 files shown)


libqtile/widget/image.py
```diff
@@ -94,3 +97,8 @@ def calculate_length(self):
             return self.img.width + (self.margin_x * 2)
         else:
             return self.img.height + (self.margin_y * 2)
+
+    def cmd_update(self, filename):
+        self.filename = filename
+        self._update_image()
+        self.draw()
```
elParaguayo, line 104: If your new image is not the same size as the previous one then you should do a `self.bar.draw()` to ensure the other widgets' positions are also adjusted.

zordsdavini: added update for length check and redraw but got some check issue what I don't understand

elParaguayo: Those check issues aren't caused by this. Still trying to track that one down...

elParaguayo: If you rebase this branch onto the latest master then those failing tests should disappear.

zordsdavini: rebased

m-col:
Works great, thanks! Could you add a line in the changelog so that others can find the new feature?

zordsdavini:
@m-col added info to CHANGELOG

m-col:
Great, thanks, auto-merge is queued!

zordsdavini:
@m-col automatical merge failed because last time I did merge with master (old habit). Did rebase

PR #3102, opened 2021-12-16
Grab keys in async mode when we swallow

### LINKED ISSUE #3092: Can't run CommandSet extension
This is split from https://github.com/qtile/qtile/issues/2730

Qtile version: 0.18.2.dev130+g3fe5e0d1
Commit where problem starts: https://github.com/qtile/qtile/commit/4c0988c2e155ff33b0179d22db653454635be814
Xephyr log `cannot grab keyboard`

### Test case
```
    EzKey("M-C-q", lazy.run_extension(extension.CommandSet(
        commands={
            'a': 'xterm'
            },
        foreground='#FF0000', selected_background='#FF0000')),
        desc='test'),
```

jwijenbergh:
Clients such as dmenu immediately grab the keyboard on opening. Dmenu
is used for extensions such as CommandSet.
This is a problem as when we grab the key in sync mode, it is only
ungrabbed on key release. Grabbing the key in sync mode is only needed
when we want the key to not be swallowed (passed to the client), so
let's grab in Async mode when it should be swallowed.

Co-authored-by: m-col as he came up with the idea, I only implemented it.

Fixes #3092 

Note that I can't add a test for this, manual testing is needed. My initial testing is that it works perfectly :+1: 

Not sure if we want to add somewhere in the docs about the fact that CommandSet or similar functions wont work with `swallow=False`? Not sure why someone would even want to set `swallow=False` for commands like these though. If an user really wants to set swallow to `False` with commands that immediately grab the keyboard, we need to implement on release keybindings.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -454,6 +454,11 @@ class Core(base.Core):
         keysym, modmask = self.lookup_key(key)
         codes = self.conn.keysym_to_keycode(keysym)
 
+        grab_mode = xcffib.xproto.GrabMode.Async
+        # If the key needs to be passed, we need to grab sync
+        if not key.swallow:
+            grab_mode = xcffib.xproto.GrabMode.Sync
+
         for code in codes:
             if code == 0:
                 logger.warning(f"Can't grab {key} (unknown keysym: {hex(keysym)})")
@@ -465,7 +470,7 @@ class Core(base.Core):
                     modmask | amask,
                     code,
                     xcffib.xproto.GrabMode.Async,
-                    xcffib.xproto.GrabMode.Sync,
+                    grab_mode,
                 )
         return keysym, modmask & self._valid_mask
 
@@ -509,12 +514,17 @@ class Core(base.Core):
         if isinstance(mouse, config.Drag):
             eventmask |= EventMask.ButtonRelease
 
+        grab_mode = xcffib.xproto.GrabMode.Async
+        # If the key needs to be passed, we need to grab sync
+        if not mouse.swallow:
+            grab_mode = xcffib.xproto.GrabMode.Sync
+
         for amask in self._auto_modmasks():
             self.conn.conn.core.GrabButton(
                 True,
                 self._root.wid,
                 eventmask,
-                xcffib.xproto.GrabMode.Sync,
+                grab_mode,
                 xcffib.xproto.GrabMode.Async,
                 xcffib.xproto.Atom._None,
                 xcffib.xproto.Atom._None,


elParaguayo:
I'm happy to test this later.

jwijenbergh:
> I'm happy to test this later.

Thanks. Do you happen to know why CI is failing here? AFAICT it's not related to this pr is it?

jwijenbergh:
Now that I think about it and test it some more, this does not correctly handle the behaviour that keys that fail the `when` check have to be passed to the client. This will not work now because we're grabbing the keys async only based on the key's swallow parameter, not on the `lazy.function` check. Any idea how we can fix this @m-col?

To test, e.g. do this


```python
def handle_key(self):
    print("EXECUTED KEY")

Key([mod], "x", lazy.function(handle_key).when(layout='idonotexist')),
```

This is now not executed (which is correct) but the key is not passed to the focused client as it is grabbed async.

m-col:
I think the ungrabbing is meant to be handled differently for sync and async grabs. So the current ungrabbing (part your original swallowing PR) works for the sync grabs because only sync grabs were used. Before then, grabs were async and they were ungrabbed differently. So I think when ungrabbing the code needs to check if it's a sync or async grab and then ungrab accordingly.

jwijenbergh:
> I think the ungrabbing is meant to be handled differently for sync and async grabs. So the current ungrabbing (part your original swallowing PR) works for the sync grabs because only sync grabs were used. Before then, grabs were async and they were ungrabbed differently. So I think when ungrabbing the code needs to check if it's a sync or async grab and then ungrab accordingly.

I'm not sure I completely understand.
I think actually ungrabbing is not really the issue is it here? The issue is that we can't replay async events, ReplayKeyboard exists in XAllowEvents but it only obviously works for sync grabs.

elParaguayo:
> > I'm happy to test this later.
> 
> Thanks. Do you happen to know why CI is failing here? AFAICT it's not related to this pr is it?
> 
> I only found this
> 
> ```
> libqtile/command/base.py:144: error: Incompatible return value type (got "Optional[Any]", expected "Callable[..., Any]")
> libqtile/utils.py:106: error: Unsupported operand types for + ("Tuple[int, ...]" and "Tuple[float]")
> ```
> 
> But I haven't changed anything here

If you rebase onto latest master that should remove these errors (but we're reverting the prior keyboard swallow commit so we may not need this for now...)

m-col:
The original commit was reverted so when rebasing you'll have to include include that commit here too. Still thinking about the fix...

jwijenbergh:
> The original commit was reverted so when rebasing you'll have to include include that commit here too. Still thinking about the fix...

We could resort to also syncing grabs when lazy functions have a `.when()` attached to it? I doubt any user has a dmenu keybind that uses a `.when()` check.

m-col:
> Find a way to replay keyboard events with async mode (something like mimicking xdotool)

This seems like it would give us the ideal final result and usability. This is possible with the xtest extension, which I assume is what xdotool uses. If we went down this route it might be good to stick the implementation into `Core.simulate_keypress` and use that, enabling its use for clients rather than it just being for internal use.

> Or reintroduce a swallow enum with the following values:
> 
>     Swallow.ALWAYS (default): We always swallow so we grab async. This is where dmenu bindings work.
>     Swallow.CHECK: We only swallow when the when() check has succeeded. We grab sync.
>     Swallow.NEVER: We never swallow the binding so we grab sync

Could you expand on how this might be implemented? Is this what you were picturing:

```
======= ===== ========== =========== ========
Setting Grab  when fails when passes w/ dmenu
======= ===== ========== =========== ========
ALWAYS  Async do nothing do nothing  Works
CHECK   Sync  Replay     Sync        Unsupported
NEVER   Sync  Replay     Replay      Unsupported
======= ===== ========== =========== ========
```

jwijenbergh:
> > Find a way to replay keyboard events with async mode (something like mimicking xdotool)
> 
> This seems like it would give us the ideal final result and usability. This is possible with the xtest extension, which I assume is what xdotool uses. If we went down this route it might be good to stick the implementation into `Core.simulate_keypress` and use that, enabling its use for clients rather than it just being for internal use.

Indeed, the code for xdotool seems to be here https://github.com/jordansissel/xdotool/blob/735e301665e7f9b8fe850588e88a3a0973695eec/xdo.c#L1539

> > Or reintroduce a swallow enum with the following values:
> > ```
> > Swallow.ALWAYS (default): We always swallow so we grab async. This is where dmenu bindings work.
> > Swallow.CHECK: We only swallow when the when() check has succeeded. We grab sync.
> > Swallow.NEVER: We never swallow the binding so we grab sync
> > ```
> 
> Could you expand on how this might be implemented? Is this what you were picturing:
> 
> ```
> ======= ===== ========== =========== ========
> Setting Grab  when fails when passes w/ dmenu
> ======= ===== ========== =========== ========
> ALWAYS  Async do nothing do nothing  Works
> CHECK   Sync  Replay     Sync        Unsupported
> NEVER   Sync  Replay     Replay      Unsupported
> ======= ===== ========== =========== ========
> ```

That is indeed what I was picturing. I guess that should work right? But I don't like the complexities that it brings for the user. I will try the xdotool approach.

jwijenbergh:
It seems like this has already been tried? https://github.com/qtile/qtile/blob/b19b3596c12d8f8c57a0d31eafffa29ea24ba5c6/libqtile/backend/x11/core.py#L748

m-col:
`sendevent` could also work, though I *think* I may have tried to implement that one time unsuccessfully. I do think xtest would work though. Either is good, but sendevent is probably the 'right' approach, if it can do this.

jwijenbergh:
Superseded by #3113

PR #3106, opened 2021-12-17
Fix mypy errors

elParaguayo:
A couple of mypy failures have cropped up recently. This addresses both:
1) CommandObject.command can return None rather than Callable
2) alpha value in rgb gives a tuple of (int, int, int, float)

### DIFF
--- a/libqtile/command/base.py
+++ b/libqtile/command/base.py
@@ -128,7 +128,7 @@ class CommandObject(metaclass=abc.ABCMeta):
         Return None if no such object exists
         """
 
-    def command(self, name: str) -> Callable:
+    def command(self, name: str) -> Optional[Callable]:
         """Return the command with the given name
 
         Parameters
@@ -170,6 +170,7 @@ class CommandObject(metaclass=abc.ABCMeta):
         """
         if name in self.commands:
             command = self.command(name)
+            assert command
             signature = self._get_command_signature(command)
             spec = name + signature
             htext = inspect.getdoc(command) or ""
--- a/libqtile/command/interface.py
+++ b/libqtile/command/interface.py
@@ -148,10 +148,13 @@ class QtileCommandInterface(CommandInterface):
             The keyword arguments to pass into the command graph call.
         """
         obj = self._command_object.select(call.selectors)
-
+        cmd = None
         try:
             cmd = obj.command(call.name)
         except SelectError:
+            pass
+
+        if cmd is None:
             return "No such command."
 
         logger.debug("Command: %s(%s, %s)", call.name, args, kwargs)
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -103,7 +103,7 @@ def rgb(x: ColorType) -> Tuple[float, float, float, float]:
         vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
         if len(x) == 8:
             alpha = int(x[6:8], 16) / 255.0
-        vals += (alpha,)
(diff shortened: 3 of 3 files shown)


libqtile/command/base.py
```diff
@@ -128,7 +128,7 @@ def _select(self, name: str, sel: Optional[Union[str, int]]) -> Optional[Command
         Return None if no such object exists
         """
 
-    def command(self, name: str) -> Callable:
+    def command(self, name: str) -> Optional[Callable]:
```
elParaguayo, line 131: The alternative to returning None would be to define a `no_op` command and return that.

libqtile/command/base.py
```diff
@@ -170,6 +170,7 @@ def cmd_doc(self, name) -> str:
         """
         if name in self.commands:
             command = self.command(name)
+            assert command
```
elParaguayo, line 173: Ugly hack but it's here because `self.command` can return `None`.

libqtile/command/interface.py
```diff
@@ -148,10 +148,13 @@ def execute(self, call: CommandGraphCall, args: Tuple, kwargs: Dict) -> Any:
...
         """
         obj = self._command_object.select(call.selectors)
-
+        cmd = None
         try:
             cmd = obj.command(call.name)
         except SelectError:
+            pass
+
+        if cmd is None:
```
elParaguayo, line 157: As above, you could do away with this if we had a `no_op` function returned.

libqtile/command/base.py
m-col: I think `None` is the right call because it *is* right.

libqtile/command/base.py
m-col: Hmm, I think mypy is right to complain here because we are running `self.command(name)` knowing that the return type is not going to be `None` (right?). In which case the assert makes sense to clarify this (to mypy as well as the reader). Alternatively it could `if None: continue` or something.

libqtile/command/interface.py
m-col: Would that fix it? Is the problem with this code (before your addition) not that `cmd` might never be set if a `SelectError` is raised?

m-col (approved):
LGTM. Do you know if mypy updated recently or something?

This has caused problems before - we could consider pinning the mypy version in the CI so we can update mypy in our own time instead of getting sudden failures.

libqtile/command/interface.py
elParaguayo: You're right. That wouldn't fix the SelectError issue so maybe it's better as I've got it here.

elParaguayo:
> LGTM. Do you know if mypy updated recently or something?
> 
> This has caused problems before - we could consider pinning the mypy version in the CI so we can update mypy in our own time instead of getting sudden failures.

Yeah. New version came out on 15 Dec.

I don't mind it not being pinned as this shows that there _were_ errors in our code that should be fixed.

libqtile/command/base.py
elParaguayo: It shouldn't be `None` as we've done `if name in self.commands:`. I think the `assert` is needed or we do a `# type: ignore`.

m-col: Right - that's kinda what I was alluding to. *We* know that this cannot be `None` because we know we are passing in a `name` that will give us a non-`None` value. But the type system is saying "you're getting an `Optional[T]` here". Our assert is (rightfully) telling mypy "nah mate it's not a `None` don't worry", so I think it's fine with the assert.

PR #3109, opened 2021-12-18
Fix max_chars in TextBox widget

elParaguayo:
`max_chars` was only being applied on a subsequent change of text. This change allows `max_chars` to be applied to the initial text set in the widget.


m-col:
Noice thanks!

PR #3113, opened 2021-12-19
Re-introduce configurable key swallowing

jwijenbergh:
Continuation of #2925 and #3102

~Instead of grabbing keys sync, which has problems in Dmenu (as it grabs the keyboard). Let's use XSendEvent. Note that mouse buttons are still grabbed sync, but this does not interfere with dmenu. If someone thinks this is unreasonable I could also use XSendEvent here.~

In the discussion here, we have figured out that XSendEvent would not work for every client. In this pr I'm using XTest

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -621,7 +621,7 @@ class Core(base.Core):
         )
         self.flush()
 
-    def fake_keyPress(self, event) -> None:
+    def fake_KeyPress(self, event) -> None:  # noqa: N802
         xtest = self.conn.conn(xcffib.xtest.key)
         # First release the key as it is possibly already pressed
         self.fake_xtest(xtest, event, 3)
@@ -649,7 +649,7 @@ class Core(base.Core):
             # We need to ungrab the key as otherwise we get an event loop
             self.ungrab_key(key)
             # Modifier is pressed, just repeat the event with xtest
-            self.fake_keyPress(event)
+            self.fake_KeyPress(event)
             # Grab the key again
             self.grab_key(key)
 
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -612,14 +612,38 @@ class Core(base.Core):
     def fake_KeyPress(self, event) -> None:
         xtest = self.conn.conn(xcffib.xtest.key)
         # First release the key as it is possibly already pressed
-        xtest.FakeInput(3, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        xtest.FakeInput(
+            3,
+            event.detail,
+            xcffib.xproto.Time.CurrentTime,
+            event.root,
+            event.root_x,
+            event.root_y,
+            0,
+        )
         self.flush()
         # Fake input by...
         # Presssing the key
-        xtest.FakeInput(2, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        xtest.FakeInput(
+            2,
+            event.detail,
+            xcffib.xproto.Time.CurrentTime,
+            event.root,
+            event.root_x,
+            event.root_y,
+            0,
+        )
         self.flush()
         # And then releasing again
-        xtest.FakeInput(3, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        xtest.FakeInput(
+            3,
+            event.detail,
+            xcffib.xproto.Time.CurrentTime,
+            event.root,
+            event.root_x,
+            event.root_y,
+            0,
+        )
         self.flush()
 
     def handle_KeyPress(self, event, *, simulated=False) -> None:  # noqa: N802
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -609,42 +609,27 @@ class Core(base.Core):
             except IndexError:
                 logger.debug("Invalid desktop index: %s", index)
 
+    def fake_xtest(self, xtest, event, input_type) -> None:
+        xtest.FakeInput(
+            input_type,
+            event.detail,
+            xcffib.xproto.Time.CurrentTime,
+            event.root,
+            event.root_x,
+            event.root_y,
+            0,
+        )
+        self.flush()
+
     def fake_KeyPress(self, event) -> None:
         xtest = self.conn.conn(xcffib.xtest.key)
         # First release the key as it is possibly already pressed
-        xtest.FakeInput(
-            3,
-            event.detail,
-            xcffib.xproto.Time.CurrentTime,
-            event.root,
-            event.root_x,
-            event.root_y,
-            0,
-        )
-        self.flush()
+        self.fake_xtest(xtest, event, 3)
         # Fake input by...
         # Presssing the key
-        xtest.FakeInput(
-            2,
-            event.detail,
-            xcffib.xproto.Time.CurrentTime,
-            event.root,
-            event.root_x,
-            event.root_y,
-            0,
-        )
-        self.flush()
+        self.fake_xtest(xtest, event, 2)
         # And then releasing again
-        xtest.FakeInput(
-            3,
-            event.detail,
-            xcffib.xproto.Time.CurrentTime,
-            event.root,
-            event.root_x,
-            event.root_y,
-            0,
-        )
-        self.flush()
+        self.fake_xtest(xtest, event, 3)
 
     def handle_KeyPress(self, event, *, simulated=False) -> None:  # noqa: N802
         assert self.qtile is not None
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -31,6 +31,7 @@ from typing import TYPE_CHECKING
 import xcffib
 import xcffib.render
 import xcffib.xproto
+import xcffib.xtest
 from xcffib.xproto import EventMask, StackMode
 
 from libqtile import config, hook, utils
@@ -48,6 +49,7 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
+    xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
     xcffib.xproto.NoExposureEvent,
@@ -360,7 +362,6 @@ class Core(base.Core):
             "ButtonPress",
             "ButtonRelease",
             "KeyPress",
-            "KeyRelease",
         ]
         if hasattr(event, "window"):
             window = self.qtile.windows_map.get(event.window)
@@ -468,7 +469,7 @@ class Core(base.Core):
                     modmask | amask,
                     code,
                     xcffib.xproto.GrabMode.Async,
-                    xcffib.xproto.GrabMode.Sync,
+                    xcffib.xproto.GrabMode.Async,
                 )
         return keysym, modmask & self._valid_mask
 
@@ -608,28 +609,40 @@ class Core(base.Core):
             except IndexError:
                 logger.debug("Invalid desktop index: %s", index)
 
+    def fake_KeyPress(self, event) -> None:
+        xtest = self.conn.conn(xcffib.xtest.key)
+        # First release the key as it is possibly already pressed
+        xtest.FakeInput(3, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        self.flush()
+        # Fake input by...
+        # Presssing the key
+        xtest.FakeInput(2, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        self.flush()
+        # And then releasing again
+        xtest.FakeInput(3, event.detail, xcffib.xproto.Time.CurrentTime, event.root, event.root_x, event.root_y, 0)
+        self.flush()
+
     def handle_KeyPress(self, event, *, simulated=False) -> None:  # noqa: N802
         assert self.qtile is not None
 
         keysym = self.conn.code_to_syms[event.detail][0]
-        handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+        key, handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
 
         if simulated:
-            # Simulated key press are currently for internal use only
+            # Even though simulated keybindings could use a proper X11 event, we don't want do any fake input
+            # This is because it needs extra code for e.g. pressing/releasing the modifiers
+            # This we don't want to handle and instead leave to external tools such as xdotool
             return
 
-        if handled:
-            # Swallow the event
-            self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.SyncKeyboard, event.time)
-        else:
-            # Forward the event to any focused client
-            self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayKeyboard, event.time)
-
-    def handle_KeyRelease(self, event) -> None:  # noqa: N802
-        assert self.qtile is not None
-
-        # We just forward the event to any focused client as we do not handle on release key bindings yet
-        self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayKeyboard, event.time)
+        # As we're grabbing async we can't just replay it, so...
+        # We need to forward the event to the focused window
+        if not handled and key:
+            # We need to ungrab the key as otherwise we get an event loop
+            self.ungrab_key(key)
+            # Modifier is pressed, just repeat the event with xtest
+            self.fake_KeyPress(event)
+            # Grab the key again
+            self.grab_key(key)
 
     def handle_ButtonPress(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
@@ -785,7 +798,6 @@ class Core(base.Core):
 
     def simulate_keypress(self, modifiers, key):
         """Simulates a keypress on the focused window."""
-        # FIXME: This needs to be done with sendevent, once we have that fixed.
         modmasks = xcbq.translate_masks(modifiers)
         keysym = xcbq.keysyms.get(key.lower())
 
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -520,7 +520,7 @@ class Core(base.Core):
                 True,
                 self._root.wid,
                 eventmask,
-                xcffib.xproto.GrabMode.Sync,
+                xcffib.xproto.GrabMode.Async,
                 xcffib.xproto.GrabMode.Async,
                 xcffib.xproto.Atom._None,
                 xcffib.xproto.Atom._None,
@@ -658,26 +658,12 @@ class Core(base.Core):
 
         button_code = event.detail
         state = event.state & self._valid_mask
-        root = False
 
         if not event.child:  # The client's handle_ButtonPress will focus it
-            root = True
             self.focus_by_click(event)
 
-        handled = self.qtile.process_button_click(
-            button_code, state, event.event_x, event.event_y
-        )
-
-        # Prevent double clicks on the root window if handled was False
-        if root:
-            handled = True
-
-        if handled:
-            # Swallow the event
-            self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.SyncPointer, event.time)
-        else:
-            # Forward the event to any focused client
-            self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayPointer, event.time)
+        self.qtile.process_button_click(button_code, state, event.event_x, event.event_y)
+        self.conn.conn.core.AllowEvents(xcffib.xproto.Allow.ReplayPointer, event.time)
 
     def handle_ButtonRelease(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -153,6 +153,7 @@ class Core(base.Core):
 
         self.qtile = None  # type: Qtile | None
         self._painter = None
+        self._xtest = self.conn.conn(xcffib.xtest.key)
 
         numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["num_lock"])[0]
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
@@ -609,28 +610,6 @@ class Core(base.Core):
             except IndexError:
                 logger.debug("Invalid desktop index: %s", index)
 
-    def fake_xtest(self, xtest, event, input_type) -> None:
-        xtest.FakeInput(
-            input_type,
-            event.detail,
-            xcffib.xproto.Time.CurrentTime,
-            event.root,
-            event.root_x,
-            event.root_y,
-            0,
-        )
-        self.flush()
-
-    def fake_KeyPress(self, event) -> None:  # noqa: N802
-        xtest = self.conn.conn(xcffib.xtest.key)
-        # First release the key as it is possibly already pressed
-        self.fake_xtest(xtest, event, 3)
-        # Fake input by...
-        # Presssing the key
-        self.fake_xtest(xtest, event, 2)
-        # And then releasing again
-        self.fake_xtest(xtest, event, 3)
-
     def handle_KeyPress(self, event, *, simulated=False) -> None:  # noqa: N802
         assert self.qtile is not None
 
@@ -649,7 +628,7 @@ class Core(base.Core):
             # We need to ungrab the key as otherwise we get an event loop
             self.ungrab_key(key)
             # Modifier is pressed, just repeat the event with xtest
-            self.fake_KeyPress(event)
+            self._fake_KeyPress(event)
             # Grab the key again
             self.grab_key(key)
 
@@ -778,6 +757,27 @@ class Core(base.Core):
 
     def handle_ScreenChangeNotify(self, event) -> None:  # noqa: N802
         hook.fire("screen_change", event)
+
+    def _fake_input(self, input_type, detail, x=0, y=0) -> None:
+        self._xtest.FakeInput(
+            input_type,
+            detail,
+            0,  # This is a delay, not timestamp, according to AwesomeWM
+            xcffib.XCB_NONE,
+            x,  # x: Only used for motion events
+            y,  # y: Only used for motion events
+            0,
+        )
+        self.flush()
+
+    def _fake_KeyPress(self, event) -> None:  # noqa: N802
+        # First release the key as it is possibly already pressed
+        for input_type in (
+            xcbq.XCB_KEY_RELEASE,
+            xcbq.XCB_KEY_PRESS,
+            xcbq.XCB_KEY_RELEASE,
+        ):
+            self._fake_input(input_type, event.detail)
 
     @contextlib.contextmanager
     def disable_unmap_events(self):
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -48,7 +48,6 @@ if TYPE_CHECKING:
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
-    xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
     xcffib.xproto.MapNotifyEvent,
     xcffib.xproto.NoExposureEvent,
@@ -361,6 +360,7 @@ class Core(base.Core):
             "ButtonPress",
             "ButtonRelease",
             "KeyPress",
+            "KeyRelease",
         ]
         if hasattr(event, "window"):
             window = self.qtile.windows_map.get(event.window)
@@ -468,7 +468,7 @@ class Core(base.Core):
                     modmask | amask,
                     code,
                     xcffib.xproto.GrabMode.Async,
-                    xcffib.xproto.GrabMode.Async,
+                    xcffib.xproto.GrabMode.Sync,
                 )
         return keysym, modmask & self._valid_mask
 
@@ -519,7 +519,7 @@ class Core(base.Core):
                 True,
                 self._root.wid,
                 eventmask,
-                xcffib.xproto.GrabMode.Async,
+                xcffib.xproto.GrabMode.Sync,
                 xcffib.xproto.GrabMode.Async,
                 xcffib.xproto.Atom._None,
                 xcffib.xproto.Atom._None,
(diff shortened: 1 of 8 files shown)


jwijenbergh:
Simulated key presses are now proper X events.

simulatekeypress

They still go through the qtile keybindings, so if an user wants to ignore the swallowing when simulating keypresses, they can make their own function that ungrabs and then grabs the binding again. Or just mark the keybind non-swallowing with `swallow=False`.

libqtile/backend/x11/core.py
```diff
@@ -609,6 +609,9 @@ def handle_KeyPress(self, event) -> None:  # noqa: N802
         # As we're grabbing async we can't just replay it, so...
         # We need to forward the event to the focused window
         if not handled and self.qtile.current_window:
+            # Shush mypy
+            # X11 windows always have a window attribute
+            assert hasattr(self.qtile.current_window, "window")
```
jwijenbergh, line 614: I've added this to shush mypy but it still says Window has no attribute window? Any reason why?

m-col: `hasattr` here will evaluate to `True` even if `self.qtile.current_window.window` is `None`. Maybe you were meaning to use `getattr`? You could do `assert self.qtile.current_window.window`. Or alternatively `if self.qtile.current_window.window:`

libqtile/backend/x11/core.py
```diff
@@ -744,18 +766,29 @@ def painter(self):
...
+        ev = KeyPressEvent.synthetic(
+            child=child,
+            detail=keysym_code,
+            event=self._root.wid,
+            event_x=0,
+            event_y=0,
+            root=self._root.wid,
+            root_x=0,
+            root_y=0,
+            same_screen=1,
```
m-col, line 787: Perhaps it would be better to use the real pointer coordinates? Otherwise windows will always receive the events in the top left corner.

test/test_swallow.py
```diff
@@ -0,0 +1,141 @@
...
+from libqtile import config
+from libqtile.backend.x11 import xcbq
+from libqtile.backend.x11.core import Core
+from libqtile.confreader import Config
+from libqtile.lazy import lazy
+
+
+# Function that does nothing
+def swallow_nop(qtile):
+    pass
```
m-col, line 29: Idea: you could use a global variable as a counter and increment it every time this function is called. That way we can confirm that the bound function is being called when expected.

libqtile/backend/x11/core.py
jwijenbergh: It is but they're keypresses not mouse presses so does this make a difference?

test/test_swallow.py
jwijenbergh: Good idea!

test/test_swallow.py
```diff
@@ -0,0 +1,141 @@
...
+# Helper to send process_key_event to the core manager
+# It also looks up the keysym and mask to pass to it
+def send_process_key_event(manager, key):
+    keysym, mask = Core.lookup_key(None, key)
+    output = manager.c.eval(f"self.process_key_event({keysym}, {mask})")
+    # Assert if eval successful
+    assert output[0]
+    # Convert the string to a bool
+    return output[1] == 'True'
+
```
m-col, line 94: I wonder if we could replace this with `cmd_simulate_buttonpress` (once the wayland implementation is added)

jwijenbergh: Probably indeed! The way I did it here is really hacky

libqtile/backend/x11/core.py
m-col: Ah yeah, my mistake. I assume it doesn't make a difference, and perhaps the arguments are there because the signature might match a generic synthetic event signature or something.

libqtile/backend/x11/core.py
jwijenbergh: Oh that makes sense, I will change this

test/test_swallow.py
m-col: Ha, we have lots of these hacky evals in the tests - they are very useful!

jwijenbergh: True but it's especially hacky because I used None for the self argument in lookup key :laughing:

libqtile/backend/x11/core.py
jwijenbergh: Well x11 does actually give the pointer x and y on real events when I logged them, so your idea is not wrong

test/test_swallow.py
jwijenbergh: On second thought, the `cmd_simulate_keypress` function does not return the handled handled boolean which is needed here. So that will have to be added there

m-col: The test window gets closed if it receives a key event, so if it's swallow vs non-swallow that you need to check then that might be a useful way to detect if a key event was not swallowed.

jwijenbergh: TIL that is useful!

libqtile/backend/x11/core.py
jwijenbergh: Still doesn't work :(, I wanted to do hasattr as window is not defined as a member in the backend window base

m-col: Maybe `assert getattr(self.qtile.current_window, "window", False)`? Then it's saying that the attribute exists and is truthy

jwijenbergh: Good idea!, trying...

jwijenbergh: Still the same, maybe there is some ignore comment we can do?

m-col: Maybe `assert isinstance(self.qtile.current_window, (window.Window, window.Static))`? mypy is likely inferring that the window is the base window class. Otherwise `# type: ignore` on the offending lines, presumably where `current_window.window` is accessed.

jwijenbergh: `assert isinstance(self.qtile.current_window, (window.Window, window.Static))` worked, thanks a lot!

test/test_swallow.py
jwijenbergh: I couldn't get a global variable working :/ I don't think pytest likes that, I tried a bunch of things. But the global variable is not correctly updated, it stays at 0. Even though the function is executed

libqtile/backend/x11/core.py
jwijenbergh: Implemented.

test/test_swallow.py
elParaguayo: Can you make a fixture and extend the scope of the fixture to `module` - your noop function could then increment a value within that fixture.

Feel free to ignore this as I'm coming at it with virtually no context!

m-col: Oh yeah nice, good idea

jwijenbergh: Thanks. I tried something with fixtures but I hardly understood them. I will have another look at it. I guess the same could be done for the .when() Match pr, but there I just used `toggle_floating` with a test window.

m-col:
Unfortunately this isn't working for me as expected. With this config to call rofi:
```py
keys.extend([
    Key(["shift"], "f", lazy.spawn("rofi -show"), swallow=True),
    Key(["control"], "f", lazy.spawn("rofi -show"), swallow=False),
])
```
The keys are not being sent to the focused window (xterm) when I call the second binding.

jwijenbergh:
> :

Hmmm and does it work without a binding that uses rofi? Isn't this the expected behaviour because rofi grabs the keyboard? Because when I try to e.g. switch groups in dmenu (have to check rofi) it won't allow me to.

m-col:
Oh, I'm being an idiot, aren't I? We decided that `swallow=False` is unsupported when the binding triggers a keyboard grab, didn't we? Please could you add a note about that to the changelog entry and the docstrings in config.py? People are almost certainly going to be confused by this!

jwijenbergh:
> Oh, I'm being an idiot, aren't I? We decided that `swallow=False` is unsupported when the binding triggers a keyboard grab, didn't we? Please could you add a note about that to the changelog entry and the docstrings in config.py? People are almost certainly going to be confused by this!

Now that I think about it, if it doesn't work for you then that is actually not correct. `swallow=False` with dmenu is only a problem with sync grabs, as we're still grabbing async in this pr and just sending an event this should work. Furthermore, as you're using `lazy.spawn`, rofi/dmenu grabbing the keyboard shouldn't even be a problem so not what sure is going on for you

For me all seems to work fine.

## Starting rofi with lazy.spawn
Key:
```python
Key([], "b", lazy.spawn("rofi -show"), swallow=False),
```
Gif:
swallowfalserofi

## CommandSet
This also works and forwards the key to the focused window *after* dmenu is closed, which makes sense.
```python
Key([], "b", lazy.run_extension(extension.CommandSet(
    commands={
        'a': 'xterm'
    },
foreground='#FF0000', selected_background='#FF0000')),
swallow=False),
```

swallowfalsecommandset


Can you do this keybinding with the focused window being "xev" ? To see if the event is correctly forwarded.

test/test_swallow.py
jwijenbergh: @elParaguayo do you have a simple example how can I implement this with fixtures? I can't seem to access it inside the lazy function. How would the lazy function and fixture itself look like?

m-col:
Using the same keybinding as you, `Key([], "b", lazy.spawn("rofi -show"), swallow=False),`, these are the events printed by xev:

```
FocusOut event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyGrab, detail NotifyAncestor

KeyPress event, serial 34, synthetic YES, window 0x1a3,
    root 0x1a3, subw 0x1c00001, time 23816063, (1495,633), root:(1495,633),
    state 0x0, keycode 56 (keysym 0x62, b), same_screen YES,
    XLookupString gives 1 bytes: (62) "b"
    XmbLookupString gives 1 bytes: (62) "b"
    XFilterEvent returns: False

LeaveNotify event, serial 34, synthetic NO, window 0x1c00001,
    root 0x1a3, subw 0x0, time 23816163, (530,102), root:(1495,633),
    mode NotifyGrab, detail NotifyAncestor, same_screen YES,
    focus YES, state 0

FocusIn event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyUngrab, detail NotifyAncestor

KeymapNotify event, serial 34, synthetic NO, window 0x0,
    keys:  2   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
           0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   

FocusOut event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyGrab, detail NotifyAncestor

FocusIn event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyUngrab, detail NotifyAncestor

KeymapNotify event, serial 34, synthetic NO, window 0x0,
    keys:  2   0   0   0   0   0   0   0   4   0   0   0   0   0   0   0   
           0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   

EnterNotify event, serial 34, synthetic NO, window 0x1c00001,
    root 0x1a3, subw 0x0, time 23817793, (530,102), root:(1495,633),
    mode NotifyUngrab, detail NotifyAncestor, same_screen YES,
    focus YES, state 0

KeymapNotify event, serial 34, synthetic NO, window 0x0,
    keys:  0   0   0   0   0   0   0   0   4   0   0   0   0   0   0   0   
           0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   

KeyRelease event, serial 34, synthetic NO, window 0x1c00001,
    root 0x1a3, subw 0x0, time 23817872, (530,102), root:(1495,633),
    state 0x0, keycode 66 (keysym 0xff1b, Escape), same_screen YES,
    XLookupString gives 1 bytes: (1b) "
    XFilterEvent returns: False

<< The events below are me focussing back to the terminal that called xev >>

KeyPress event, serial 34, synthetic NO, window 0x1c00001,
    root 0x1a3, subw 0x0, time 23819378, (530,102), root:(1495,633),
    state 0x0, keycode 64 (keysym 0xffeb, Super_L), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

FocusOut event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyGrab, detail NotifyAncestor

PropertyNotify event, serial 34, synthetic NO, window 0x1c00001,
    atom 0x19a (_NET_FRAME_EXTENTS), time 23819452, state PropertyNewValue

ConfigureNotify event, serial 34, synthetic YES, window 0x1c00001,
    event 0x1c00001, window 0x1c00001, (960,526), width 950, height 516,
    border_width 5, above 0x0, override NO

PropertyNotify event, serial 34, synthetic NO, window 0x1c00001,
    atom 0x198 (WM_STATE), time 23819453, state PropertyNewValue

FocusOut event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyWhileGrabbed, detail NotifyNonlinear

FocusOut event, serial 34, synthetic NO, window 0x1c00001,
    mode NotifyUngrab, detail NotifyPointer
```

The second event looks like a 'b' is being sent to xev, which made me think that I *may* have read that xterm does not support send-event! However, testing with urxvt, kitty, and firefox, none of these show a 'b'. Here is a clip of me testing with firefox, which I include because you can actually see when I press the 'b' key that the focus to the text input kinda flashes off-on-off before rofi appears... not sure if that is relevant:

https://user-images.githubusercontent.com/44536113/146844275-73ded2b7-ecba-44c6-959c-371ac224eb9e.mp4

What terminal are you testing with?

jwijenbergh:
> Using the same keybinding as you, `Key([], "b", lazy.spawn("rofi -show"), swallow=False),`, these are the events printed by xev:
> 
> ```
> FocusOut event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyGrab, detail NotifyAncestor
> 
> KeyPress event, serial 34, synthetic YES, window 0x1a3,
>     root 0x1a3, subw 0x1c00001, time 23816063, (1495,633), root:(1495,633),
>     state 0x0, keycode 56 (keysym 0x62, b), same_screen YES,
>     XLookupString gives 1 bytes: (62) "b"
>     XmbLookupString gives 1 bytes: (62) "b"
>     XFilterEvent returns: False
> 
> LeaveNotify event, serial 34, synthetic NO, window 0x1c00001,
>     root 0x1a3, subw 0x0, time 23816163, (530,102), root:(1495,633),
>     mode NotifyGrab, detail NotifyAncestor, same_screen YES,
>     focus YES, state 0
> 
> FocusIn event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyUngrab, detail NotifyAncestor
> 
> KeymapNotify event, serial 34, synthetic NO, window 0x0,
>     keys:  2   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
>            0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
> 
> FocusOut event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyGrab, detail NotifyAncestor
> 
> FocusIn event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyUngrab, detail NotifyAncestor
> 
> KeymapNotify event, serial 34, synthetic NO, window 0x0,
>     keys:  2   0   0   0   0   0   0   0   4   0   0   0   0   0   0   0   
>            0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
> 
> EnterNotify event, serial 34, synthetic NO, window 0x1c00001,
>     root 0x1a3, subw 0x0, time 23817793, (530,102), root:(1495,633),
>     mode NotifyUngrab, detail NotifyAncestor, same_screen YES,
>     focus YES, state 0
> 
> KeymapNotify event, serial 34, synthetic NO, window 0x0,
>     keys:  0   0   0   0   0   0   0   0   4   0   0   0   0   0   0   0   
>            0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
> 
> KeyRelease event, serial 34, synthetic NO, window 0x1c00001,
>     root 0x1a3, subw 0x0, time 23817872, (530,102), root:(1495,633),
>     state 0x0, keycode 66 (keysym 0xff1b, Escape), same_screen YES,
>     XLookupString gives 1 bytes: (1b) "
>     XFilterEvent returns: False
> 
> << The events below are me focussing back to the terminal that called xev >>
> 
> KeyPress event, serial 34, synthetic NO, window 0x1c00001,
>     root 0x1a3, subw 0x0, time 23819378, (530,102), root:(1495,633),
>     state 0x0, keycode 64 (keysym 0xffeb, Super_L), same_screen YES,
>     XLookupString gives 0 bytes: 
>     XmbLookupString gives 0 bytes: 
>     XFilterEvent returns: False
> 
> FocusOut event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyGrab, detail NotifyAncestor
> 
> PropertyNotify event, serial 34, synthetic NO, window 0x1c00001,
>     atom 0x19a (_NET_FRAME_EXTENTS), time 23819452, state PropertyNewValue
> 
> ConfigureNotify event, serial 34, synthetic YES, window 0x1c00001,
>     event 0x1c00001, window 0x1c00001, (960,526), width 950, height 516,
>     border_width 5, above 0x0, override NO
> 
> PropertyNotify event, serial 34, synthetic NO, window 0x1c00001,
>     atom 0x198 (WM_STATE), time 23819453, state PropertyNewValue
> 
> FocusOut event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyWhileGrabbed, detail NotifyNonlinear
> 
> FocusOut event, serial 34, synthetic NO, window 0x1c00001,
>     mode NotifyUngrab, detail NotifyPointer
> ```
> 
> The second event looks like a 'b' is being sent to xev, which made me think that I _may_ have read that xterm does not support send-event! However, testing with urxvt, kitty, and firefox, none of these show a 'b'. Here is a clip of me testing with firefox, which I include because you can actually see when I press the 'b' key that the focus to the text input kinda flashes off-on-off before rofi appears... not sure if that is relevant:
> out.mp4
> 
> What terminal are you testing with?

I am using "st" and there it works but not indeed on e.g. kitty, thanks for this 😅. Maybe XTest needs to be used instead? I will do some digging

m-col:
> Maybe XTest needs to be used instead?

I think this is the likely solution. Clients would need to explicitly support send-event, adding a handler for those messages to their event loops. XTest on the other hand is 'invisible' to the clients and appears as if it is a real event, so will likely be the most seamless for its usage. Hopefully it also is on our implementation side.

m-col:
You didn't actually need `simulate_keypress` in the end for the core code, right? I've been working on a wayland implementation and it is non-trivial I think. I'd be happy to drop that change if it is not needed for the swallowing feature. In case we do continue down this route, here is my current version in case it's of interest. It is completely untested and unfinished but you can see it gets a bit hacky:
```py
    def simulate_keypress(self, modifiers: List[str], key: str) -> None:
        """Simulates a keypress on the focused window."""
        if self.keyboards:
            #keysym = xkb.keysym_from_name(key, case_insensitive=True)
            #mods = wlrq.translate_masks(modifiers)
            event = wlr_ffi.new("struct wlr_event_keyboard_key *")
            event.time_msec = int(Timespec.get_monotonic_time().time * 1000)
            event.keycode
            event.update_state = True
            event.state = 1  # pressed
            self.keyboards[1].keyboard.key_event.emit(event)
            self.flush()
            event.state = 0  # released
            self.keyboards[1].keyboard.key_event.emit(event)
            self.flush()
```
The more "correct" alternative would be to create a virtual keyboard client and use that, which would work even if a keyboard wasn't currently attached to the seat (because it makes one). I haven't used client-side virtual keyboard code but I'm sure wlrctl would be handy for that.

jwijenbergh:
> > Maybe XTest needs to be used instead?
> 
> I think this is the likely solution. Clients would need to explicitly support send-event, adding a handler for those messages to their event loops. XTest on the other hand is 'invisible' to the clients and appears as if it is a real event, so will likely be the most seamless for its usage. Hopefully it also is on our implementation side.

Yeah I'm implementing it atm, it's a bit more annoying but possible

jwijenbergh:
> You didn't actually need `simulate_keypress` in the end for the core code, right? I've been working on a wayland implementation and it is non-trivial I think. I'd be happy to drop that change if it is not needed for the swallowing feature. In case we do continue down this route, here is my current version in case it's of interest. It is completely untested and unfinished but you can see it gets a bit hacky:
> 
> ```python
>     def simulate_keypress(self, modifiers: List[str], key: str) -> None:
>         """Simulates a keypress on the focused window."""
>         if self.keyboards:
>             #keysym = xkb.keysym_from_name(key, case_insensitive=True)
>             #mods = wlrq.translate_masks(modifiers)
>             event = wlr_ffi.new("struct wlr_event_keyboard_key *")
>             event.time_msec = int(Timespec.get_monotonic_time().time * 1000)
>             event.keycode
>             event.update_state = True
>             event.state = 1  # pressed
>             self.keyboards[1].keyboard.key_event.emit(event)
>             self.flush()
>             event.state = 0  # released
>             self.keyboards[1].keyboard.key_event.emit(event)
>             self.flush()
> ```
> 
> The more "correct" alternative would be to create a virtual keyboard client and use that, which would work even if a keyboard wasn't currently attached to the seat (because it makes one). I haven't used client-side virtual keyboard code but I'm sure wlrctl would be handy for that.

I indeed do not need it. It's just if the user wants to call it from the command line using `qtile cmd-obj -o cmd -f simulate_keypress --args "" "a"`. And indeed that looks not really trivial, but pretty similar to xtest actually.

m-col:
I don't think we need to have a feature to simulate key presses, so if that command is not available (like currently) then that is fine. Keys can be simulated with external tools like `xdotool`, `wlrctl` etc.

jwijenbergh:
> I don't think we need to have a feature to simulate key presses, so if that command is not available (like currently) then that is fine. Keys can be simulated with external tools like `xdotool`, `wlrctl` etc.

Agreed! However, I could easily add it for the x backend with xtest now. Can you test the latest commit? It is a bit WIP but it seems to work for me. I can't test extensively though right now. I will be able to work on it some more later. In the meantime let me know if there are any issues. It's maybe a bit hacky

jwijenbergh:
Marked as draft up until we both find a solution that works for every situation. Hopefully the xtest approach I just pushed works :crossed_fingers:

libqtile/backend/x11/core.py
```diff
@@ -600,25 +601,60 @@ def handle_ClientMessage(self, event) -> None:  # noqa: N802
             except IndexError:
                 logger.debug("Invalid desktop index: %s" % index)
 
+    def fake_KeyPress(self, keycode, needs_release=True) -> None:
+        xtest = self.conn.conn(xcffib.xtest.key)
+        # The key needs to be released as it was previously pressed
+        if needs_release:
+            xtest.FakeInput(3, keycode, xcffib.xproto.Time.CurrentTime, 0, 0, 0, 0)
```
jwijenbergh, line 608: Note to self, in `FakeInput` don't input all zeros but use the event settings. For DeviceId, we need to pass 0 though (last argument).

libqtile/backend/x11/core.py
```diff
@@ -600,25 +601,60 @@ def handle_ClientMessage(self, event) -> None:  # noqa: N802
...
         keysym = self.conn.code_to_syms[event.detail][0]
-        self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+        key, handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+
+        # As we're grabbing async we can't just replay it, so...
+        # We need to forward the event to the focused window
+        if not handled and key:
+            # We need to ungrab the key as otherwise we get an event loop
+            self.ungrab_key(key)
+            # Modifier is already pressed, just repeat the keycode
```
jwijenbergh, line 628: This is not the case when it is send from `simulate_keypress`, handle this or leave this to external tools such as xdotool? If we want to leave it to xdotool I would need to add a `simulated` argument again (with default `False`)

libqtile/backend/x11/core.py
```diff
@@ -600,25 +601,60 @@ def handle_ClientMessage(self, event) -> None:  # noqa: N802
...
+            self.fake_KeyPress(event.detail)
+            # Grab the key again
+            self.grab_key(key)
```
jwijenbergh, line 631: @m-col could it be the case that we're re-grabbing the key too early? Or does the `flush()` call already take care of it? I'm not sure if this is right what I'm doing.

libqtile/backend/x11/core.py
m-col: Yeah, i think leaving that simulated keypress code the way it was is fine

libqtile/core/manager.py
```diff
@@ -371,25 +371,33 @@ def paint_screen(
     ) -> None:
         self.core.painter.paint(screen, image_path, mode)
 
-    def process_key_event(self, keysym: int, mask: int) -> None:
+    def process_key_event(self, keysym: int, mask: int) -> Tuple[Optional[Union[Key, KeyChord]], bool]:
```
jwijenbergh, line 374: I had to return a tuple with the key as I need to ungrab and grab it!

libqtile/backend/x11/core.py
```diff
@@ -600,25 +601,60 @@ def handle_ClientMessage(self, event) -> None:  # noqa: N802
...
 
         keysym = self.conn.code_to_syms[event.detail][0]
-        self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+        key, handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
+
+        # As we're grabbing async we can't just replay it, so...
+        # We need to forward the event to the focused window
+        if not handled and key:
+            # We need to ungrab the key as otherwise we get an event loop
+            self.ungrab_key(key)
```
jwijenbergh, line 627: Note to self: test with KeyChords

libqtile/backend/x11/core.py
jwijenbergh: I have done a middle ground, I have made simulate keypress closer to an x11 event so that we can in the future maybe support this. But for now simulated keypresses will only elicit qtile functions through the bindings, but will not actually forward it to a window. Xdotool is fine for this use case

jwijenbergh:
Sorry for taking so long on this, was busy

Steps taken:
- Rebased with master
- Squashed commits to use xtest
- Simplified dummy events with simulate_keypress because we just want to advise users to use xdotool/ydotool
- Tested manually with keychord and keys, seems to work fine for me

elParaguayo:
You probably need to add `xcffib.xtest` to `MOCK_MODULES` in `docs/conf.py`.

jwijenbergh:
> You probably need to add `xcffib.xtest` to `MOCK_MODULES` in `docs/conf.py`.

Thanks! Was wondering why they were failing

libqtile/core/manager.py
```diff
@@ -405,25 +405,35 @@ def reconfigure_screens(self, ev: Any = None) -> None:
     def paint_screen(self, screen: Screen, image_path: str, mode: str | None = None) -> None:
         self.core.painter.paint(screen, image_path, mode)
 
-    def process_key_event(self, keysym: int, mask: int) -> None:
+    def process_key_event(
+        self, keysym: int, mask: int
+    ) -> Tuple[Optional[Union[Key, KeyChord]], bool]:
```
m-col, line 410: We no longer need Optional, Tuple or Union:
```suggestion
    ) -> tuple[Key | KeyChord | None, bool]:
```

m-col (commented):
Excellent work as always. For the pep8 complaint feel free to `# noqa: N802` it or change the name of the method, up to you. Then LGTM!

libqtile/core/manager.py
jwijenbergh: Ah yes thanks, that is really neat

jwijenbergh:
Fixes pushed, I have named it `fake_KeyPress` again with the noqa command to be in line with `handle_KeyPress`

m-col:
Do you mind if I push some commits? I've been playing around with some of the event simulation for the tests

jwijenbergh:
> Do you mind if I push some commits? I've been playing around with some of the event simulation for the tests

Go ahead, that would be great! Do I need to enable that somewhere?

jwijenbergh:
In the meantime let me figure out if it's maybe possible to do pointer grabs async as well and use xtest too. I think maybe grabbing things sync is not the best idea

jwijenbergh:
Steps taken:

- Rebased with master
- Fixed changelog entry (it was in the wrong release lol)
- Removed ability to swallow for mouse bindings as grabbing sync is not really nice IMO and I did not get it working with xtest. On top of that I think that not swallowing mouse bindings is less useful. So for now it's only Key and KeyChord.

m-col:
> Go ahead, that would be great! Do I need to enable that somewhere?

Yes, I think you will have something in the sidebar that will enable this?

m-col:
I didn't want to push to this directly as I'm unsure about all the changes I made. Two commits here: https://github.com/m-col/qtile/commits/3113

The first commit exposes commands for `warp_pointer` and `simulate_click` which makes it easier to perform these behaviours in tests in a backend-agnostic way.

The second commit doesn't work and doesn't fix anything, it was a bit of a work in progress before I got distracted. There were two things I had in mind with the tests. Less important was using `test_data` to count swallows to give a bit more of a quantitative assessment of what's going on in the tests. More important was removing the x11-specific imports in `test_swallow.py`. Qtile should be able to run under Wayland without xcffib installed, and the same for running the tests with the wayland backend only (`--backend=wayland`). Currently xcffib is required unconditionally because xcbq is imported in `confreader.py` but I have some work in the pipeline to make xcffib unnecessary for systems that are wayland only. Either way we should merge this asap, so I'm happy for us to either move the tests into `test/backend/x11/` or remove them completely in the mean time (probably keeping them for x11 only is preferable though!).

jwijenbergh:
> > Go ahead, that would be great! Do I need to enable that somewhere?
> 
> Yes, I think you will have something in the sidebar that will enable this?

Ah yes this should be on

jwijenbergh:
Steps taken:
- Implemented some changes by @m-col to cleanup code and improve tests
- Rebased with master

libqtile/core/manager.py
```diff
@@ -405,25 +405,33 @@ def reconfigure_screens(self, ev: Any = None) -> None:
     def paint_screen(self, screen: Screen, image_path: str, mode: str | None = None) -> None:
         self.core.painter.paint(screen, image_path, mode)
 
-    def process_key_event(self, keysym: int, mask: int) -> None:
+    def process_key_event(self, keysym: int, mask: int) -> tuple[Key | KeyChord | None, bool]:
         key = self.keys_map.get((keysym, mask), None)
         if key is None:
             logger.debug("Ignoring unknown keysym: %s, mask: %s", keysym, mask)
-            return
+            return (None, False)
```
m-col, line 412: The bool here specifies whether the binding was handled, and if it wasn't then it couldn't be swallowed, right?

What I'm thinking is that the `tuple[Key | KeyChord | None, bool]` return type implies that `(None, True)` is a valid return value, but from what I can tell it wouldn't be valid. Could it be `tuple[Key | KeyChord, bool] | None` instead? This would also mean the change in `libqtile/backend/wayland/inputs.py` doesn't need the `[1]` index, making it a bit more clear when just reading the code in that file.

m-col (approved):
Just one comment but other than that, LGTM :)

libqtile/core/manager.py
jwijenbergh: But if we return ```(key, False)``` then before it would return ```False```, whereas with the change it would return ```True``` if I don't index by 1. This would only work for the ```None, False``` case where we can now just return ```None```.

 I could also make it ```Key | KeyChord | None``` but then I would have to set key to None conditionally on the ```key.swallow``` parameter which is weird when reading ```process_key_event```. What do you think?

jwijenbergh:
Rebased with master.

libqtile/core/manager.py
m-col: OK, nvm my suggestion actually. I think it's fine as is :)

m-col:
Happy to merge this after rebasing (assuming that fixes the CI) :)

libqtile/core/manager.py
jwijenbergh: Great, thanks for thinking with me :+1:

Issue #3040, opened 2021-11-19
Sending an existing Qtile keybinding to another program using qtile.cmd_simulate_keypress() fails


raj-magesh:
I use a terminal (kitty) with tiling-like support and I'd like to reuse the same keybindings I use in Qtile with the kitty panes. However, Qtile has priority over the keybindings. I've managed to disable certain Qtile keybindings whenever a terminal window is focused using this minimal example:

```
def move_focus(qtile, direction):
    if qtile.current_window.get_wm_class()[0] != 'kitty':
        if direction == 'left':
            qtile.current_group.layout.cmd_left()
    else:
        qtile.cmd_simulate_keypress([MOD], 'h')

keys = [Key([MOD], 'h', lazy.function(move_focus, direction='left'), desc='move focus left')]
```

However, the key combination still doesn't get passed through to the terminal. How can I achieve that?

I've found other people with a similar issue:
- https://groups.google.com/g/qtile-dev/c/cp83eFM-JwQ
- https://groups.google.com/g/qtile-dev/c/e25bgej9Ses/m/pAczjD9lmEoJ

_Originally posted by @raj-magesh in https://github.com/qtile/qtile/discussions/3037_

EDIT:

Okay, going through old Qtile issues, the `kitty` developer also raised this issue ~8 years ago: https://github.com/qtile/qtile/issues/324

elParaguayo:
Have a look at #2925 

I think that's fixing the issue that you're looking at as it references the same old issue! 

The PR is in draft but, if you're interested in testing, I'm sure they'd appreciate feedback and/or suggestions.

raj-magesh:
That's awesome, thank you for letting me know! I'm not sure exactly how I'll be able to help out testing, and I'm really busy for the next few weeks, but once I'm free I can download that version and see if my use case works. I'm new to Qtile and collaborative coding, so I don't have much expertise either.

elParaguayo:
No problem at all. 

Help is always appreciated if you're able (there's help in the docs about how to collaborate if you need some tips) but don't worry if you don't have the time.

As you can see, the issue is being looked at so it should get resolved at some point!

jwijenbergh:
[...]

PR #3114, opened 2021-12-20
Add focused window Match to lazy function's .when() check

jwijenbergh:
This adds the focused attribute to `.when()` checks which uses `config.Match` objects.
So this can be used to limit a keybinding on a certain type of focused window.
When swallowing is complete, #3040 can be done with:

```python

Key([MOD], 'h', lazy.layout.left().when(focused=Match(wm_class=re.compile('(?!kitty)'))), desc='move focus left when kitty is not focused')

### DIFF
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -17,8 +17,21 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+from __future__ import annotations
 
-from typing import Dict, Iterable, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import (
+        Dict,
+        Iterable,
+        List,
+        Optional,
+        Set,
+        Tuple,
+        Union,
+    )
+    from libqtile.config import Match
 
 from libqtile.command.client import InteractiveCommandClient
 from libqtile.command.graph import (
@@ -46,6 +59,7 @@ class LazyCall:
         self._args = args
         self._kwargs = kwargs
 
+        self._focused: Optional[Match] = None
         self._layouts: Set[str] = set()
         self._when_floating = True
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: libqtile/lazy.py [hunk 1]
```diff
@@ -18,7 +18,19 @@
...
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
+
+if TYPE_CHECKING:
+    from libqtile.config import Match
```
m-col, line 33: While you're here, could you put all of the other `typing` imports within the `if TYPE_CHECKING` block?

libqtile/lazy.py
```diff
@@ -90,18 +103,23 @@ def kwargs(self) -> Dict:
         """The kwargs to the given call"""
         return self._kwargs
 
-    def when(self, layout: Optional[Union[Iterable[str], str]] = None,
+    def when(self, focused: Optional['Match'] = None, layout: Optional[Union[Iterable[str], str]] = None,
```
m-col, line 106: IIRC the preferred way for Python 3.10+ is to state the objects themselves directly in type annotations, i.e. `Optional[Match]` with the conditional import. In previous versions this would fail with a `NameError` but from py310 type annotation evaluation is delayed, so the exception is not raised during runtime. When type checking the objects are imported and it works. Before py310 the exception is avoided by using `from __future__ import annotations` which many of our modules have too. You may need to add that to the top of this file too to avoid the exception in the runtimes for py <3.10.

m-col:
A couple small comments but overall this looks great, thanks for implementing this! You're making many people happy with your PRs I am sure :smile:

libqtile/lazy.py
jwijenbergh: Ah yeah will do, I needed it for Match to prevent the circular dependency

jwijenbergh:
> A couple small comments but overall this looks great, thanks for implementing this! You're making many people happy with your PRs I am sure 😄

It's never too early for xmas! happy to do them

libqtile/lazy.py
jwijenbergh: Ah makes sense, thanks for the detailed info.

libqtile/lazy.py
jwijenbergh: Implemented.

libqtile/lazy.py
jwijenbergh: I used `from __future__ import annotations` which worked perfectly. Thanks again

elParaguayo:
> A couple small comments but overall this looks great, thanks for implementing this! You're making many people happy with your PRs I am sure smile

Agreed!

m-col:
Awesome, thanks!

PR #3115, opened 2021-12-20
Adding widget screenshots to docs

elParaguayo:
This is draft for now as I'm interested in feedback. I could have put this in discussions but it's easier to comment on code here!

The idea of the PR is to add screenshots of widgets to the docs so users can understand what certain configuration options do. Like this:

The screenshots are generated by using our existing testing framework i.e. we can leverage existing fixtures to start a Qtile instance, generate the widget and save its contents to a file. 

Crucially, adding screenshots for widgets is then very simple.

To do:
- [ ] Tidy up sphinx/ReadTheDocs requirements
- [x] Add screenshots for all widgets for which we currently have tests (current progress can be seen on my fork: https://qtile-elp-fork.readthedocs.io/en/latest/manual/ref/widgets.html)
- [x] Update widget how to page to explain how to create screenshots

I have no idea how this will fare in tests as a successful CI build is no guarantee of a success on ReadTheDocs.

I'll add some comments to the code below to flag some bits that are interesting or where I'd like some suggestions.

I'm also fine if we think this shouldn't be implemented!

Also, looking at that screenshot, it looks like the default colours in `block` mode aren't a great combination. You can't see group 2 at all!


.readthedocs.yaml
```diff
@@ -0,0 +1,42 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+# Required
+version: 2
+
+build:
```
elParaguayo, line 7: We need ReadTheDocs to install the same packages that we do in the first stage of CI. I suspect some of these can be removed.

docs/Makefile
```diff
@@ -167,3 +167,11 @@ genkeyimg:
 	sed -i '/^\.\. END_LS_PNG/e find _static/keybindings/* | awk '"'"'{ print length, $$0 }'"'"' | sort -n | cut -d" " -f2- | awk '"'"'{print ".. image:: /" $$1}'"'"'' manual/commands/keybindings.rst
 	@echo
 	@echo "Keybinding images have been generated."
+
+genwidgetscreenshots:
+	rm -rf screenshots/widgets
+	@echo "Generating screenshots for widgets using pytest fixtures."
+#    pytest -o python_files="ss_*.py" -o python_functions="ss_*" -q --no-header --no-summary ../test
+	pytest -o python_files="ss_*.py" -o python_functions="ss_*" --backend x11 ../test
```
elParaguayo, line 175: Screenshots are generated by only running pytest on files with the `ss_` prefix.

docs/conf.py
```diff
@@ -200,6 +200,8 @@ def __dir__(self):
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+html_extra_path = ['_static/screenshots']
```
elParaguayo, line 203: We need to copy the contents of this folder to the build directory. This is one way to do it but we could also do it in the Makefile.

NB sphinx will copy any image using the `.. image:` directive automatically but, here, the table is actually a raw html table so the copying doesn't happen.

docs/requirements.txt
```diff
@@ -4,5 +4,7 @@ sphinx_rtd_theme
 funcparserlib==1.0.0a0
 sphinxcontrib-seqdiag
 numpydoc
-
--r ../requirements.txt
```
elParaguayo, line 8: This is removed as it's now called separately in the yaml file. This is to preserve ordering of the installation so we can get `xcffib` before `cairocffi`.

docs/sphinx_qtile.py
```diff
@@ -52,6 +55,27 @@
 
         Supported bar orientations: {{ obj.orientations }}
     {% endif %}
+    {% if is_widget and screen_shots %}
+    .. raw:: html
```
elParaguayo, line 59: I really struggled with any of the helper ways to generate a table so resorted to raw html.

docs/sphinx_qtile.py
```diff
@@ -52,6 +55,27 @@
...
+
+        <table class="docutils">
+        <tr>
+        <td width="50%"><b>example</b></td>
+        <td width="50%"><b>config</td>
+        </tr>
+    {% for sshot, conf in screen_shots.items() %}
```
elParaguayo, line 66: The table has the image and config. Should we add a column for notes/comments? e.g. for `GroupBox` (screenshot in original post) it's not obvious that there's a window open in group 1.

docs/sphinx_qtile.py
```diff
@@ -195,12 +236,21 @@ def make_rst(self):
...
-    call(['make', '-C', this_dir, 'genkeyimg'])
+    run(['make', '-C', base_dir, 'run-ffibuild'])
+    run(['make', '-C', this_dir, 'genkeyimg'])
+
+
+def generate_widget_screenshots():
+    this_dir = os.path.dirname(__file__)
+    try:
+        run(['make', '-C', this_dir, 'genwidgetscreenshots'], check=True)
+    except CalledProcessError:
```
elParaguayo, line 247: If the screenshots fail to build, the docs will also fail.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 cairocffi >= 0.9.0
 cffi >= 1.1.0
-xcffib >= 0.10.1
```
elParaguayo, line 3: This is a bit naughty. I removed `xcffib` from here as it needs to be installed before `cairocffi`. I'm not sure this file would work anyway if anyone tried to install this file.

test/backend/wayland/conftest.py
```diff
@@ -2,7 +2,11 @@
 import os
 import textwrap
 
-from libqtile.backend.wayland.core import Core
+try:
```
elParaguayo, line 5: Very hacky!

The conftest files all get imported when pytest is initialised. I'm not installing wayland in ReadTheDocs so I need to stop any `ImportError`s due to missing files/modules. Installing Wayland on ReadTheDocs doesn't seem like a viable option (can only use apt sources for Ubuntu 18.04).

test/widgets/docs_screenshots/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+    shutil.copytree(folder, docs_folder)
+    with open(log, "w") as f:
+        json.dump(key, f)
+
+    # Clear up the tests folder
+    shutil.rmtree(folder)
+
+
+@pytest.fixture
+def screenshot_manager(widget, request, manager_nospawn, minimal_conf_noscreen, target):
```
elParaguayo, line 85: The `widget` fixture is defined in each of the individual `ss_widgetname.py` file.

test/widgets/docs_screenshots/ss_clock.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.fixture
+def widget():
+    yield Clock
+
+
+@pytest.mark.parametrize("screenshot_manager", [{}, {"format": "%d/%m/%y %H:%M"}], indirect=True)
+def ss_clock(screenshot_manager):
+    screenshot_manager.take_screenshot()
```
elParaguayo, line 32: At its simplest a screenshot file is:

1. A `widget` fixture returning the widget class (can be monkeypatched as necessary)
2. A `ss_` function which is parametrized with the various configs.
3. A call to `screenshot_manager.take_screenshot()`

test/widgets/docs_screenshots/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+        Screen(top=Bar([wdgt], 32), x=0, y=0, width=300, height=300),
+        Screen(top=Bar([], 32), x=0, y=300, width=300, height=300),
+    ]
+
+    manager_nospawn.start(minimal_conf_noscreen)
+
+    # Add some convenience attributes for taking screenshots
+    manager_nospawn.target = filename
+    ss_widget = manager_nospawn.c.widget[name]
+    manager_nospawn.take_screenshot = lambda f=filename: ss_widget.take_screenshot(f())
```
m-col, line 126: re the `cmd_take_screenshot` commands above, they could be moved into this file and added to `minimal_conf_noscreen`.

Also - as it's all being run with X11 only, you can probably remove the `qtile.core.name` check from those functions and just hard-code x11 in there.

m-col (changes requested):
This is pretty cool. I'll admit I was a bit hesitant about it - fear of "bloat" and all that - but I think this will be valued quite a lot, and it's a relatively simple implementation (to read... not necessarily to write, I'm sure :wink:).

I've left some comments inline but overall it looks great.

In this review I've read through everything except the individual widget files. At a glance they look pretty neat. Would it be possible to merge them with the respective widget test modules, rather than having them split out into new files?

.readthedocs.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - libglvnd-dev
+    - libqt5gui5
+    - libegl1
+    - libegl-mesa0
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+# Optionally build your docs in additional formats such as PDF
+formats:
+- pdf
```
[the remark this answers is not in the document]
elParaguayo: I'm not aware of any quotas. Also, our current docs are downloadable as pdf, html and epub so I probably need to add those options.

NB screenshots aren't actually appearing in the pdf which I suspect is because I'm using html code rather than rst for the table. This is because I haven't found another way to get images in a table when iterating in a template.

.readthedocs.yaml
```diff
@@ -0,0 +1,51 @@
...
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  install:
+  - requirements: docs/requirements.txt
+  - requirements: requirements.txt
```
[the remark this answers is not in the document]
elParaguayo: This yaml file pins python to 3.10 (line 10) but good idea to add a comment.

docs/manual/howto/widget.rst
```diff
@@ -551,6 +551,90 @@ tests that test specific functionality of the widget (e.g. reaction to hooks).
 
 See :ref:`unit-testing` for more.
 
+Documentation
+=============
+
+It is really important that we maintain good documentation for Qtile. Any new widgets must
+therefore include sufficient documentation in order for users to understand how to
+use/configure the widget.
+
```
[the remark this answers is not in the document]
elParaguayo: Yes. Excellent point.

test/widgets/docs_screenshots/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+from libqtile.bar import Bar
+from libqtile.config import Group, Screen
+
+
+@pytest.fixture(scope="session")
+def target():
+    folder = Path(__file__).parent / "screenshots"
+    docs_folder = (
+        Path(__file__).parent
+        / ".."
```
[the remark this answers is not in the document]
elParaguayo: `__file__` is the full file path including the file name so the `parent` here gets you to the directory at which point you do the `..` syntax. To be honest, I copied it from somewhere else in the tests!

test/widgets/docs_screenshots/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+        / "docs"
+        / "_static"
+        / "screenshots"
+        / "widgets"
+    )
+    log = os.path.join(docs_folder, "shots.json")
+    if os.path.isdir(folder):
+        shutil.rmtree(folder)
+    os.makedirs(folder)
```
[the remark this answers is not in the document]
elParaguayo: No "might" about it. Will fix.

elParaguayo:
> This is pretty cool. I'll admit I was a bit hesitant about it - fear of "bloat" and all that - but I think this will be valued quite a lot, and it's a relatively simple implementation (to read... not necessarily to write, I'm sure 😉).

Actually, the screenshots are really easy to write. That's why I like this implementation.

> In this review I've read through everything except the individual widget files. 

Yeah, you don't really need to look at them. If they don't work, the build will fail. If they do work, you can see the output on that link in the first post.

> At a glance they look pretty neat. Would it be possible to merge them with the respective widget test modules, rather than having them split out into new files?

You could add them to the test file. The only essential part is that they don't have a `test_` prefix as building the screenshots during the tests won't work as readthedocs has to build the images itself so building them during the tests is wasted effort. Similarly, we don't want readthedocs to run the full test suite.

Personally, I prefer splitting them out as it's just a bit neater. Yes, we're occasionally importing things from the test file (e.g. fake modules for monkeypatching) but I still think it's a better approach given that they're run separately from the tests.

requirements.txt
elParaguayo: I don't remember the exact message but it was to do with cairo not having xcb support.

test/backend/wayland/conftest.py
elParaguayo: Seems to be that all conftest files are parsed when pytest scans tests.

As for installing dependencies, no, I think you're ok. This is only an issue for running tests which most people aren't going to be doing.

m-col:
> You could add them to the test file. The only essential part is that they don't have a test_ prefix as building the screenshots during the tests won't work as readthedocs has to build the images itself so building them during the tests is wasted effort. Similarly, we don't want readthedocs to run the full test suite.
> 
> Personally, I prefer splitting them out as it's just a bit neater. Yes, we're occasionally importing things from the test file (e.g. fake modules for monkeypatching) but I still think it's a better approach given that they're run separately from the tests.

Yeah fair enough - sounds good!

docs/Makefile
m-col, line 175: Accidentally left in comment?

docs/sphinx_qtile.py
m-col: No that should be fine. Does anybody receive any notification, or would this become apparent as part of the CI?

m-col: No I think that's fine. Will we get the red in CI on master, or any other notification?

elParaguayo:
> No I think that's fine. Will we get the red in CI on master, or any other notification?

The screenshots are generated as part of CI's `docs` test. While that doesn't guarantee success on ReadTheDocs, it is a good indicator as we're using the same dependencies.

However, there's also this which might be pretty cool: https://docs.readthedocs.io/en/stable/pull-requests.html

m-col:
OK cool sounds good.

> However, there's also this which might be pretty cool: https://docs.readthedocs.io/en/stable/pull-requests.html

Yeah, very cool!

elParaguayo:
OK. That's enabled but I suspect it may only show for new PRs.

elParaguayo:
Nice! The RTD build check works and gives a link to the built docs at the end so we can view the output.

test/widgets/docs_screenshots/conftest.py
elParaguayo: Thinking about this a bit. We need the `take_screenshot` commands to be exposed via IPC. When we move the decorators, commands can't be added to existing objects as the parsing is done at creation. 

However, I think we can still make this work by creating new classes in the fixture which inherit from `Bar` and the relevant `widget`. Putting decorators on these should work fine.

elParaguayo:
Docs seems to be building ok on RTD. Should we get this merged?

m-col (approved):

m-col (approved):
LGTM :)

I see some widgets don't have images yet. I guess they can be given an image if/when someone feels like adding one?

elParaguayo:
> LGTM :)
> 
> I see some widgets don't have images yet. I guess they can be given an image if/when someone feels like adding one?

I think that correlates with widgets that don't have tests yet. On my to do list but would be good to get this lot merged.

PR #3119, opened 2021-12-21
Lazy when: Solve focused matching when there are no windows

jwijenbergh:
With #3114 we can match on focused windows in keybindings. However, if there are no windows there is a problem as a `NoneType` does not have e.g. a `wm_class`. We need to extend the `Match` class to not error out when we match against a `NoneType`.

So we need to add a configurable empty parameter to lazycalls, to then match if there are no focused windows:

```python
Key([MOD], 'h', lazy.function(somefunction).when(focused=Match(wm_class=re.compile('(?!kitty)')), if_no_focused=True), desc='execute somefunction when the focused client is not kitty or when there are no focused windows')
```
This needs to be configurable (default to `False`) as these negative regexes are possible.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,9 @@ Qtile x.x.x, released xxxx-xx-xx:
         - Fix `Systray` crash on `reconfigure_screens`.
         - Fix bug where widgets can't be mirrored in same bar.
         - Fix various issues with setting fullscreen windows floating and vice versa.
+        - Fix a bug where a .when() check for lazy functions errors out when matching
+          on focused windows when none is focused. By default we do not match on focused windows,
+          to change this set `if_no_focused` to True.
 
 Qtile 0.20.0, released 2022-01-24:
     * features
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -50,6 +50,7 @@ class LazyCall:
         self._kwargs = kwargs
 
         self._focused: Match | None = None
+        self._if_no_focused: bool = False
         self._layouts: set[str] = set()
         self._when_floating = True
 
@@ -93,6 +94,7 @@ class LazyCall:
     def when(
         self,
         focused: Match | None = None,
+        if_no_focused: bool = False,
         layout: Iterable[str] | str | None = None,
         when_floating: bool = True,
     ) -> "LazyCall":
@@ -101,15 +103,24 @@ class LazyCall:
         Parameters
         ----------
         focused: Match or None
-            Match criteria to enable call for the current window
+            Match criteria to enable call for the current window.
+        if_no_focused: bool
+            Whether or not the `focused` attribute should also
+            match when there is no focused window.
+            This is useful when the `focused` attribute is e.g. set
+            to a regex that should also match when there is
+            no focused window.
+            By default this is set to `False` so that the focused
+            attribute only matches when there is actually a focused window.
         layout: str, Iterable[str], or None
             Restrict call to one or more layouts.
             If None, enable the call for all layouts.
         when_floating: bool
             Enable call when the current window is floating.
         """
-        if focused is not None:
-            self._focused = focused
+        self._focused = focused
+
+        self._if_no_focused = if_no_focused
 
         if layout is not None:
             self._layouts = {layout} if isinstance(layout, str) else set(layout)
@@ -120,8 +131,12 @@ class LazyCall:
     def check(self, q) -> bool:
         cur_win_floating = q.current_window and q.current_window.floating
 
-        if self._focused and not self._focused.compare(q.current_window):
-            return False
+        if self._focused:
+            if q.current_window and not self._focused.compare(q.current_window):
+                return False
+
+            if not q.current_window and not self._if_no_focused:
+                return False
 
         if cur_win_floating and not self._when_floating:
             return False
--- a/test/test_when.py
+++ b/test/test_when.py
@@ -19,7 +19,7 @@
 # SOFTWARE.
 import pytest
 
-from libqtile import config
+from libqtile import config, layout
 from libqtile.confreader import Config
 from libqtile.lazy import lazy
 
@@ -35,14 +35,27 @@ class WhenConfig(Config):
         config.Key(
             ["control"],
             "j",
-            lazy.window.toggle_floating().when(config.Match(wm_class="TestWindow")),
+            lazy.window.toggle_floating().when(focused=config.Match(wm_class="TestWindow")),
         ),
         config.Key(
             ["control"],
             "h",
-            lazy.window.toggle_floating().when(config.Match(wm_class="idonotexist")),
+            lazy.window.toggle_floating().when(focused=config.Match(wm_class="idonotexist")),
+        ),
+        config.Key(
+            ["control"],
+            "n",
+            lazy.next_layout().when(focused=config.Match(wm_class="TestWindow")),
+        ),
+        config.Key(
+            ["control"],
+            "m",
+            lazy.next_layout().when(
+                focused=config.Match(wm_class="TestWindow"), if_no_focused=True
+            ),
         ),
     ]
+    layouts = [layout.MonadWide(), layout.MonadTall()]
 
 
 when_config = pytest.mark.parametrize("manager", [WhenConfig], indirect=True)
@@ -51,7 +64,7 @@ when_config = pytest.mark.parametrize("manager", [WhenConfig], indirect=True)
 @when_config
 def test_when(manager):
     # Check if the test window is alive and tiled
-    manager.test_window("one")
+    one = manager.test_window("one")
     assert not manager.c.window.info()["floating"]
 
     # This sets the window to floating as there is no when
@@ -65,3 +78,15 @@ def test_when(manager):
     # This sets the window tiled as the class does match
     manager.c.simulate_keypress(["control"], "j")
     assert not manager.c.window.info()["floating"]
+
+    # Kill the window to create an empty group
+    manager.kill_window(one)
+    prev_layout_info = manager.c.layout.info()
+
+    # This does not go to the next layout as empty is not matched
+    manager.c.simulate_keypress(["control"], "n")
+    assert manager.c.layout.info() == prev_layout_info
+
+    # This does go to the next layout as empty is matched
+    manager.c.simulate_keypress(["control"], "m")
+    assert manager.c.layout.info() != prev_layout_info


jwijenbergh:
I should have added this to the original pr, but I forgot :sweat_smile:

m-col:
I know I'm just being dumb but I don't fully understand what the issue is. Could you elaborate a bit on what the expected behaviour would be for certain `when(focused=Match...)` setups? I think a "not-X" `Match` regex should give a `True` only in the presence of a "not-X" window, and not in the absence of any window - as no windows match the regex. And therefore a `focused=` when should always result in `False` if there is no focussed window.

jwijenbergh:
> I know I'm just being dumb but I don't fully understand what the issue is. Could you elaborate a bit on what the expected behaviour would be for certain `when(focused=Match...)` setups? I think a "not-X" `Match` regex should give a `True` only in the presence of a "not-X" window, and not in the absence of any window - as no windows match the regex. And therefore a `focused=` when should always result in `False` if there is no focussed window.

I agree that there should be no extra parameter ideally. But what would the default value then be for empty windows? Match also matches substrings so we can't just use an empty string. I'm not really that familiar with regexes so there is probably a nice way to do this.

It probably could be implemented very easily, I just couldn't see a nice way to do it

m-col:
> But what would the default value then be for empty windows? Match also matches substrings so we can't just use an empty string.

You mean when there is no window, and the `when()` has accepted a `Match`?

jwijenbergh:
> > But what would the default value then be for empty windows? Match also matches substrings so we can't just use an empty string.
> 
> You mean when there is no window, and the `when()` has accepted a `Match`?

Yeah exactly

m-col:
I think a `Match` should always only match when a window matches the conditions, which means if there are no windows then it would always be `False`. So this:
```py
Key([MOD], 'h', lazy.layout.left().when(focused=Match(wm_class=re.compile('(?!kitty)'))), desc='move focus left when kitty is not focused')
```
Would only fire when a (present) focused window matches. If there is no window, then no window matches. That's what regex would do too.

jwijenbergh:
> I think a `Match` should always only match when a window matches the conditions, which means if there are no windows then it would always be `False`. So this:
> 
> ```python
> Key([MOD], 'h', lazy.layout.left().when(focused=Match(wm_class=re.compile('(?!kitty)'))), desc='move focus left when kitty is not focused')
> ```
> 
> Would only fire when a (present) focused window matches. If there is no window, then no window matches. That's what regex would do too.

Yeah that's what I want too. But how could this be implemented? When there is no window then we still need a default value to match against right? How is this implemented in python? We can't just pass an empty string

m-col:
Just something like this I think:
```diff
diff --git a/libqtile/lazy.py b/libqtile/lazy.py
index 38850e18..ea117a6e 100644
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -130,8 +130,9 @@ class LazyCall:
     def check(self, q) -> bool:
         cur_win_floating = q.current_window and q.current_window.floating
 
-        if self._focused and not self._focused.compare(q.current_window):
-            return False
+        if self._focused:
+            if not (q.current_window and self._focused.compare(q.current_window)):
+                return False
 
         if cur_win_floating and not self._when_floating:
             return False
```

`focused` is saying "only execute this when there is a matching window" so if there is no window then it never matches and is not executed.

jwijenbergh:
> Just something like this I think:
> 
> ```diff
> diff --git a/libqtile/lazy.py b/libqtile/lazy.py
> index 38850e18..ea117a6e 100644
> --- a/libqtile/lazy.py
> +++ b/libqtile/lazy.py
> @@ -130,8 +130,9 @@ class LazyCall:
>      def check(self, q) -> bool:
>          cur_win_floating = q.current_window and q.current_window.floating
>  
> -        if self._focused and not self._focused.compare(q.current_window):
> -            return False
> +        if self._focused:
> +            if not (q.current_window and self._focused.compare(q.current_window)):
> +                return False
>  
>          if cur_win_floating and not self._when_floating:
>              return False
> ```
> 
> `focused` is saying "only execute this when there is a matching window" so if there is no window then it never matches and is not executed.

Sorry I read wrong. I don't agree that it always needs to return False. We can also match with regexes negatively. See the kitty keybind in the pr description

m-col:
Hmm, I see what you're saying but a regex is a whitelist selection so a negative "not X" regex will whitelist things that 1. exist and 2. are not X. For example, `grep` will fail to match on an empty file, and `grep -v` will also. What differences would there be in the functionality of `when()` between our two ideas? Would what I am saying be limiting in some way? I think we wouldn't be able to have `when` *not* match when a window exists and matches, I guess. Is that right?

jwijenbergh:
> Hmm, I see what you're saying but a regex is a whitelist selection so a negative "not X" regex will whitelist things that 1. exist and 2. are not X. For example, `grep` will fail to match on an empty file, and `grep -v` will also. What differences would there be in the functionality of `when()` between our two ideas? Would what I am saying be limiting in some way? I think we wouldn't be able to have `when` _not_ match when a window exists and matches, I guess. Is that right?

Okay it makes sense what you're saying. In this case the `empty` parameter stands for "do not fail if grep finds an empty file". 

I added the configurable empty parameter such that the user can instruct the match to also match when there are no windows. A default return False in this scenario wouldn't make much sense because I can imagine someone wants to make a keybinding that is executed if kitty is not focused. 

We would still need to pass None to the match even if we add explicit negative matches to the match class e.g. Match(wm_class != ...) would match if the wm_class is not ... (So also a NoneType). So an early return without even querying the match is not the right idea right?

I think the empty parameter is nice because it makes it really configurable if the user wants to match on NoneTypes

m-col:
> I can imagine someone wants to make a keybinding that is executed if kitty is not focused.

I agree it's something we need to handle, and similarly the inverse - a keybinding to *only* be executed when kitty is focussed - seems totally reasonable.

I'm not totally convinced it needs to be a change to `Match` to support matching `None`. Would it not be sufficient to pass an option to `when()` to set whether or not it should continue its checks beyond `self._focused` if there is no focused window?

jwijenbergh:
> > I can imagine someone wants to make a keybinding that is executed if kitty is not focused.
> 
> I agree it's something we need to handle, and similarly the inverse - a keybinding to _only_ be executed when kitty is focussed - seems totally reasonable.
> 
> I'm not totally convinced it needs to be a change to `Match` to support matching `None`. Would it not be sufficient to pass an option to `when()` to set whether or not it should continue its checks beyond `self._focused` if there is no focused window?

Yeah that could also be done. Maybe just have the same empty variable in when? 
```python
Lazy.function.when(focused=Match(...), empty=True)
```

```diff
diff --git a/libqtile/lazy.py b/libqtile/lazy.py
index 38850e18..ea117a6e 100644
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -130,8 +130,9 @@ class LazyCall:
     def check(self, q) -> bool:
         cur_win_floating = q.current_window and q.current_window.floating
 
-        if self._focused and not self._focused.compare(q.current_window):
-            return False
+        if self._focused:
+            if not (q.current_window and self._focused.compare(q.current_window)):
+                return self._empty
 
         if cur_win_floating and not self._when_floating:
             return False
```

And then set `self._empty` to false by default? Something like this or maybe something else?

jwijenbergh:
> > > I can imagine someone wants to make a keybinding that is executed if kitty is not focused.
> > 
> > 
> > I agree it's something we need to handle, and similarly the inverse - a keybinding to _only_ be executed when kitty is focussed - seems totally reasonable.
> > I'm not totally convinced it needs to be a change to `Match` to support matching `None`. Would it not be sufficient to pass an option to `when()` to set whether or not it should continue its checks beyond `self._focused` if there is no focused window?
> 
> Yeah that could also be done. Maybe just have the same empty variable in when?
> 
> ```python
> Lazy.function.when(focused=Match(...), empty=True)
> ```
> 
> ```diff
> diff --git a/libqtile/lazy.py b/libqtile/lazy.py
> index 38850e18..ea117a6e 100644
> --- a/libqtile/lazy.py
> +++ b/libqtile/lazy.py
> @@ -130,8 +130,9 @@ class LazyCall:
>      def check(self, q) -> bool:
>          cur_win_floating = q.current_window and q.current_window.floating
>  
> -        if self._focused and not self._focused.compare(q.current_window):
> -            return False
> +        if self._focused:
> +            if not (q.current_window and self._focused.compare(q.current_window)):
> +                return self._empty
>  
>          if cur_win_floating and not self._when_floating:
>              return False
> ```
> 
> And then set `self._empty` to false by default? Something like this or maybe something else?

@m-col what do you think? I'd like this fixed. The problem with that approach I think is that empty now is not closely related to the focused variable, that's why I think adding the empty flag to Match was a neat idea.

m-col:
> @m-col what do you think? I'd like this fixed. The problem with that approach I think is that empty now is not closely related to the focused variable, that's why I think adding the empty flag to Match was a neat idea.

Outside of usage for `when`, what other effects can that `empty` parameter have when part of `Match`? When matching against a specific window with `window.match(match)`, it will never be used because there is a window. I think that's also the case for its other uses e.g. auto-floating windows, matching new windows to groups etc. The docstring for `Match` would presumably indicate that that option only applies to `when` then right? If there is other use for it outside of `when` then I think it's fine having it as part of `Match`, otherwise keeping it in `when` only makes more sense. Of course we'd have to name the parameter to make it super clear - maybe `if_no_focused: bool` or something

jwijenbergh:
> > @m-col what do you think? I'd like this fixed. The problem with that approach I think is that empty now is not closely related to the focused variable, that's why I think adding the empty flag to Match was a neat idea.
> 
> Outside of usage for `when`, what other effects can that `empty` parameter have when part of `Match`? When matching against a specific window with `window.match(match)`, it will never be used because there is a window. I think that's also the case for its other uses e.g. auto-floating windows, matching new windows to groups etc. The docstring for `Match` would presumably indicate that that option only applies to `when` then right? If there is other use for it outside of `when` then I think it's fine having it as part of `Match`, otherwise keeping it in `when` only makes more sense. Of course we'd have to name the parameter to make it super clear - maybe `if_no_focused: bool` or something

Hmmm yeah I don't think `None` is ever passed to match as in most of the code an explicit `None` test is done. We could also just eliminate the `None` checks and let them be handled by match or maybe that doesn't make sense?

`if_no_focused` indeed for `when` could also be done.

m-col:
> Hmmm yeah I don't think None is ever passed to match as in most of the code an explicit None test is done. We could also just eliminate the None checks and let them be handled by match or maybe that doesn't make sense?

I think the intent is a bit clearer the way it's currently being done, when reading the code. I'd prefer keeping it in `when` if you think that's fine.

jwijenbergh:
Steps taken:

- Rebased with master
- Implemented your suggestions @m-col 
- Manually tested it

jwijenbergh:
CI failures seem to be unrelated

```python
  ERROR: Could not install packages due to an EnvironmentError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/3b/02/8d4d27b1cacaac2d129a27d17a22d92a2a5eedcb7817d4ed8ab0d4daf5c4/setuptools-60.9.3-py3-none-any.whl (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7fe4ef9b5a30>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
```

elParaguayo:
Yup. Re-run the tests for you. Seem to be running ok now.

elParaguayo:
Yup. Re-run the tests for you. Seem to be running ok now.

jwijenbergh:
> Yup. Re-run the tests for you. Seem to be running ok now.

Thanks!

m-col:
The code itself looks great. The documentation in the docstring could be a bit clearer, trying to imagine reading it as someone who isn't familiar with the PR. @elParaguayo what do you think?

jwijenbergh:
> The code itself looks great. The documentation in the docstring could be a bit clearer, trying to imagine reading it as someone who isn't familiar with the PR. @elParaguayo what do you think?

Agreed. is this better?

m-col:
Yeah, that's much clearer, thanks! Auto-merge enabled :)

PR #3124, opened 2021-12-22
Fix Imap widget

elParaguayo:
`poll()` method was modifying `self.text` in place before returning the value. However, `self.text` is a property setter which sets the layout text. The widget should only update the text value by returning the value from `poll()`.

### DIFF
--- a/libqtile/widget/imapwidget.py
+++ b/libqtile/widget/imapwidget.py
@@ -83,11 +83,11 @@ class ImapWidget(base.ThreadPoolText):
     def poll(self):
         im = imaplib.IMAP4_SSL(self.server, 993)
         if self.password == 'Gnome Keyring Error':
-            self.text = 'Gnome Keyring Error'
+            text = 'Gnome Keyring Error'
         else:
             im.login(self.user, self.password)
             status, response = im.status(self.mbox, '(UNSEEN)')
-            self.text = response[0].decode()
-            self.text = self.label + ': ' + re.sub(r'\).*$', '', re.sub(r'^.*N\s', '', self.text))
+            text = response[0].decode()
+            text = self.label + ': ' + re.sub(r'\).*$', '', re.sub(r'^.*N\s', '', text))
             im.logout()
-        return self.text
+        return text


m-col:
Makes sense, thanks!

elParaguayo:
Ooh. Coveralls status messages are here.

m-col:
Interesting the coverage isn't the same value across all python versions. Any idea why that might be? Do you know if it's possible to get the status for only one i.e. 3.10?

elParaguayo:
We can get it for one by only submitting test results for python 3.10.

m-col:
Cool, is that all we need, or is it helpful to have them all?

We also got some comments, like on https://github.com/qtile/qtile/pull/3119. Were these you experimenting, or should we be getting these too?

elParaguayo:
The comments were a result of the coveralls team checking the connection with our account. We shouldn't get any more as we don't want comments for every push. 

I'd leave it as it is. There  shouldn't be massive differences between the versions so I suspect they'll either all pass or all fail but we might catch some edge cases which would be interesting to see.

elParaguayo:
You could add the python 3.10 check to being required for automerge.

m-col:
OK sounds good.

> You could add the python 3.10 check to being required for automerge.

Oh yeah, that'll need updating won't it.

elParaguayo:
I was going mad, couldn't work out why I couldn't make a screenshot for this widget. Eventually tracked it down to this issue!

m-col:
Weird, mypy CI failed following the merge to master.

elParaguayo:
They just pushed a new release of mypy 30 mins ago.

m-col:
Brilliant, well that certainly explains it!

m-col:
I really think we should pin mypy in the CI. I'm also concerned that downstream packagers might run tests during packaging and it would fail if mypy updates between us releasing and them packaging.

m-col:
I'm happy to post a PR to update mypy and pin it

elParaguayo:
Yeah. It's a bit silly at the moment.

You want to take care of it? Do we pint to 0.900 and then go up with every 0.100 release?

PR #3128, opened 2021-12-22
Blackify the code

m-col:
These are the changes made:

1. Add `flake8-black` to the `pep8` tox environment, which uses
   `black --check` for checks. This enforces black formatting in the CI.
2. Add a `pyproject.toml` file which black uses for configuration.
3. Disable flake8's E203,W503 as these not pep8-compliant and therefore
   imcompatible with black. Enable W504 as black suggests. The other
   codes added to setup.cfg are the remaining flake8 defaults.
4. Run black on libqtile/ and test/


elParaguayo:
237 files changed :laughing:

m-col:
WIth auto-formatting I think we should consider lowering the line-length to something like 98. You'll see in the diff that now we have a lot of lines that are hitting that super long length and it's hurting readability IMO. Thoughts?

> 237 files changed

Yeah :)

elParaguayo:
Agreed. Some of those lines are a bit long for my liking.

m-col:
I'm code getting codestyle errors, but interestingly, I'm also getting them on master. Am I being an idiot? It passed on CI on master...

elParaguayo:
No, you're not going mad.

black has made some changes that codestyle doesn't like. e.g. in prompt.py
``` python
 txt3 = self.text[self.cursor_position + 1 :]
```
black has added a space before the colon but codestyle doesn't like it.

m-col:
I was a little bit - I was seeing the pylint output on master and thinking that was new output. Guess I'd never run the codestyle tox env locally before. I've disabled pylint because it wasn't doing anything, and ignored those error codes to make codestyle compatible with black.

m-col:
It passed!

@elParaguayo How would you prefer to proceed re how this will affect open PRs?

elParaguayo:
Hmm. No easy answer here.

Do we bite the bullet, merge and then ask people to rebase?

We probably need to give some instructions on how people should proceed here. Maybe set up a discussions page first with the relevant steps and then point people to that.

m-col:
It's inevitable that some people are going to have to rebase whenever this is merged. I think it should be enough to just tell people as part of the PR review process ("oh btw the CI failed because you need to run black on the code"). I'm more concerned about how much work this might cause for your decorators PR. FWIW I am happy to rebase for you if it becomes a trouble, because I don't want this shelved and to go stale!

elParaguayo:
Maybe I should do a test run: pull the black PR and then rebase a copy of the decorators onto it and see what happens!

m-col:
smart

elParaguayo:
Actually may not be too bad. I should be able to overwrite the changes from this PR with the decorator code and then just run black again.

m-col:
Cool, so we merge this then?

elParaguayo:
Yeah, do it. 

If you haven't already, we should add something to the docs in the hacking/contributing parts about running `black` on any code that gets submitted.

m-col:
Good idea - I've appended a commit mentioning black in the contributing docs.

ramnes:
Eh, so the last bastion that Qtile was happened to fall. :P

m-col:
Don't pretend like you didn't like it!

PR #3136, opened 2021-12-25
X11: only warp cursor when focusing a new window

### LINKED ISSUE #3135: cursor_warp makes impossible to put back a browser tab on chrome window (maybe any other browse)
# Issue description
 
When using multi monitors, I find it very useful to change cursor position when changing focus of workspaces and windows.

But as you can see in the video, that's an undesired behavior. As the title already says, I can't put it back.

# Other minor issues

When I click, holding CTRL, on a link to open a new tab inside a browser, for example, the cursor goes back to the center of that window.

If i click on someone's profile picture inside telegram-desktop, the cursor also goes back to the center of the window (in this case is indeed a new window because of the way telegram handles this situation specifically).

Thee might be a lot of other examples, but that's the general idea.

# Wish list

It would be very nice (at least for me) if we could change this behavior on specifics windows, or if I could invoke a function (with lazy?) that centers the cursor manually when switch to a window or workspace.

# Other

If there is documentation for the issue that I'm pointing here, I apologize, but I did a little bit of research before posting this issue.

I'm not demanding the devs to solve my problem, and by the fact that english is not my native language, I'm afraid I'm being a little rude.

https://user-images.githubusercontent.com/50176483/147363025-07b54054-bd96-4fe3-8024-ea7803bb35a2.mp4

# Configuration

My config is very long and I don't want to pollute this post, but I can confirm that `cursor_warp = True`

elParaguayo:
There is a bug in X11 where an active window sends a focus event (e.g. opening a link in a new tab) which results in the cursor
being warped if `cursor_warp` is set to `True` in the config.

This PR addresses that by only warping when the focus event applies to a new window.

Fixes #3135

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -57,6 +57,7 @@ class _Group(CommandObject):
         self.focus_history = []
         self.screen = None
         self.current_layout = None
+        self.last_focused = None
 
     def _configure(self, layouts, floating_layout, qtile):
         self.screen = None
@@ -153,6 +154,10 @@ class _Group(CommandObject):
                     self.floating_layout.layout(floating, screen_rect)
                 if self.current_window and self.screen == self.qtile.current_screen:
                     self.current_window.focus(warp)
+                else:
+                    # Screen has lost focus so we reset record of focused window so
+                    # focus will warp when screen is focused again
+                    self.last_focused = None
 
     def set_screen(self, screen, warp=True):
         """Set this group's screen to screen"""
@@ -200,7 +205,13 @@ class _Group(CommandObject):
         if win:
             if win not in self.windows:
                 return
+
+            # ignore focus events if window is the current window
+            if win is self.last_focused:
+                warp = False
+
             self.current_window = win
+            self.last_focused = self.current_window
             if win.floating:
                 for layout in self.layouts:
                     layout.blur()


m-col:
(why) is it not enough to check `qtile.current_window == self`?

elParaguayo:
If you do it in that function, `current_window` is already `self`

m-col:
Does it have to be in that function - could we not block this function altogether in the calling function?

elParaguayo:
Almost certainly - but I wasn't 100% sure which was the calling function here!

m-col:
Ah I see - you can use the `traceback` module to log the stack from that function, if you can reliably recreate the problem.

elParaguayo:
Nice. Will take a look tomorrow. About to head off on a longish drive.

m-col:
A good day for one!

libqtile/backend/x11/window.py
```diff
@@ -997,7 +1000,6 @@ def focus(self, warp: bool) -> None:
                 self.window.set_property("_NET_WM_STATE", state)
 
         # re-grab button events on the previously focussed window
-        old = self.qtile.core._root.get_property("_NET_ACTIVE_WINDOW", "WINDOW", unpack=int)
```
m-col, line 1000: Why did this need to be moved?

elParaguayo: Ha. It doesn't any more. I moved it because I was using `old` in a test to limit warping but removed that test. I'll put it back.

m-col:
Thanks!

PR #3138, opened 2021-12-26
Added multi window scratchpad support.

falamous:
Hacked in multi-window scratchpad support (for example for tabbed, which is often used to add tabbing support to terminals like st or alacritty by spawning tabs as new windows). Currently if a `Match` is provided to `config.DropDown`, the corresponding dropdown will only catch a single matching window, but in my opinion catching all matching windows is far more intuitive, so this it what was changed. The dropdown_info was also changed to return info for all windows instead of a single one.

### DIFF
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -203,6 +203,64 @@ class DropDownToggler(WindowVisibilityToggler):
             WindowVisibilityToggler.show(self)
 
 
+class DropDownTogglerSet:
+    """
+    DropDownTogglerSet is wrapper that stores DropDownToggler objects and allows
+    to toggle, show, hide the entire set.
+    """
+
+    def __init__(self, scratchpad_name, ddconfig):
+        self.scratchpad_name = scratchpad_name
+        self.ddconfig = ddconfig
+        self._windows = {}
+
+    def reconfigure(self, ddconfig):
+        self.ddconfig = ddconfig
+
+    def add_window(self, window):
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for submitting this.

Can you give us a bit more details on how you see this being used and the benefits over the current system?

If you've got multiple windows that have been matched, how do you navigate between them?

falamous:
As I said it would be used for applications that spawn multiple windows and manage them independent of the window manager (through window embedding for example), perhaps it could also be used for applications with popups. In both cases no way to navigate between windows is required from the window manager. It is a very small change both to the current user experience and the code base. I could add a allow_multiple_windows (defaulting to false) or max_windows (defaulting to 1 and if set to a value < 0 allowing for any amount of windows) to the scratchpad object and make it so if a scratchpad can capture only one window it will only capture a window, only if was spawned by it, so there would absolutely zero change to the user experience, however I didn't think it was necessary the change is already small as is. If it helps to understand my idea, here is my config, which I made the changes to accommodate:
```python
DropDown(
    "kek",
    f"tabbed -c -n {scratchpad_id} alacritty --class {scratchpad_id} --embed",
    match=Match(wm_class=[scratchpad_id]),
    opacity=1,
    y=0.05,
    x=0.05,
    height=0.9,
    width=0.9,
    on_focus_lost_hide=True,
    warp_pointer=False,
)
```

m-col:
It looks like all of the matched windows take the same geometry. How do you navigate between the windows in a set?

falamous:
It seems that recent changes in the tools I use render this hack unnecessary. Sorry for bothering.

PR #3142, opened 2021-12-27
Update widgets "how to" in docs

elParaguayo:
Added detail on mouse events (button presses and enter/leave).

### DIFF
--- a/docs/manual/howto/widget.rst
+++ b/docs/manual/howto/widget.rst
@@ -433,6 +433,76 @@ For example, the Mpris2 widget uses the following code:
 interfaces. Refer to the `dbus-next documentation <https://python-dbus-next.readthedocs.io/en/latest/>`_
 for more information on how to use the module.
 
+Mouse events
+============
+
+By default, widgets handle button presses and will call any function that is bound to the button in the
+``mouse_callbacks`` dictionary. The dictionary keys are as follows:
+
+ - ``Button1``: Left click
+ - ``Button2``: Middle click
+ - ``Button3``: Right click
+ - ``Button4``: Scroll up
+ - ``Button5``: Scroll down
+ - ``Button6``: Scroll left
+ - ``Button7``: Scroll right
+
+You can then define your button bindings in your widget (e.g. in ``__init__``):
+
+.. code:: python
+
+    class MyWidget(widget.TextBox)
+
+        def __init__(self, *args, **config):
+            widget.TextBox.__init__(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


docs/manual/howto/widget.rst
```diff
@@ -433,6 +433,63 @@ For example, the Mpris2 widget uses the following code:
...
+            self.short_format = self.format
+
+        def mouse_enter(self, *args, **kwargs):
+            self.format = self.long_format
+            self.bar.draw()
+
+        def mouse_leave(self, *args, **kwargs):
+            self.format = self.short_format
+            self.bar.draw()
+
```
m-col, line 492: Perhaps add a comment to say that `self.bar.draw()` is needed only if the size of the widget changes, otherwise `self.draw()` is sufficient?

docs/manual/howto/widget.rst
```diff
@@ -433,6 +433,63 @@ For example, the Mpris2 widget uses the following code:
...
+.. code:: python
+
+    class MyWidget(widget.TextBox)
+
+        def __init__(self, *args, **config):
+            widget.TextBox.__init__(self, *args, **kwargs)
+            self.add_callbacks(
+                {
+                    "Button1": self.left_click_method,
+                    "Button3": self.right_click_method
```
m-col, line 461: Lazy objects can't be used here, can they? I remember once this was asked and there have been changes to them so wondering if that is now possible. Reading the code it doesn't look like they could be - calling a lazy object is only useful for providing arguments, but not executing the lazy call. Is that right?

m-col (commented):
Awesome, thanks for adding this. Just a couple comments added.

docs/manual/howto/widget.rst
elParaguayo: That's already included further up the page.

docs/manual/howto/widget.rst
elParaguayo: Lazy calls can be added. This is different to the changes to lazy calls to pass args.

I'll add a bit to show that lazy calls cab be bound to buttons too.

m-col: Ah, the handling code is in `widget.base`, gotcha. Neat.

docs/manual/howto/widget.rst
m-col: Ah, it's too early for my brain

m-col (approved):
Great, LGTM!

PR #3143, opened 2021-12-27
Add wm_class to Window.info()

m-col:
This can be useful for scripting.

A test is added to ensure backends implement required cmd_info() values.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -215,7 +215,20 @@ class _Window(CommandObject, metaclass=ABCMeta):
 
     @abstractmethod
     def info(self) -> Dict[str, Any]:
-        """Return information on this window."""
+        """
+        Return information on this window.
+
+        Mimimum required keys are:
+        - name
+        - x
+        - y
+        - width
+        - height
+        - group
+        - id
+        - wm_class
+
+        """
         return {}
 
     def cmd_info(self) -> Dict:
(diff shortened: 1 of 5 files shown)


elParaguayo (approved):

elParaguayo:
LGTM

elParaguayo:
Why is this stalling the tests?

elParaguayo:
I don't know if this is related but I came across a weird issue when looking at the `Systray` icon ordering that may or may not be related. Basically, my first attempt had `get_wm_name` in the `key` function when I was trying to sort the icons. This didn't seem to work and, instead, I had to set the name when the window was created.

Is there a way the `wm_class` could be loaded when the window is created, rather than having to call `get_wm_class` each time there's an `info` call?

It may make no difference as the tests are timing out, whereas my subsequent calls to get the window name just ended up returning `None`.

m-col:
Interesting, that does sound like the same thing.

> Is there a way the wm_class could be loaded when the window is created, rather than having to call get_wm_class each time there's an info call?

I think this might be possible, yes. The window class/app_id can change at any point but I believe we can listen for changes via `PropertyNotify`. The wayland class already does listen for changes. I'll add a change for this.

m-col:
I've added a new commit to do the above. The new test failure is a legit failure which I need to fix, so the hanging is gone :D

elParaguayo:
Great. Although I have no idea what's causing that issue... Good that we can get round it.

elParaguayo:
Ha. This breaks `Systray. I *love* that widget!

elParaguayo:
It's working!

m-col:
About time!

PR #2606, opened 2021-07-05
New layout: MonadThreeCol

mbenford:
This PR introduces a new layout: `MonadThreeCol`. It's based on [XMonad's ThreeColumns](https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Layout-ThreeColumns.html) layout and it's useful for people using ultra wide displays. It behaves similarly to MonadTall with two key differences: the main window can be placed at the center of the screen and the secondary windows are divided into two columns:

image

Windows are distributed following XMonad's ThreeColumns behavior:

image
image

This is a working in progress and shouldn't be merged yet. It's functional - I've been using it myself - but there are some rough edges that need smoothing out. I wanted to share it as it stands to get some early feedback on both the code (Python isn't my strong suit) and the layout design. I tried to use as much as I could from `MonadTall` and only rewrote the parts that were too coupled to the way the tall layout works. 

To-do list
- [x] Port XMonad's ThreeCol layout.
- [x] Port XMonad's ThreeColMid layout. 
- [x] Fix double margins.
- [x] Fix resize of secondary windows.
- [x] Write tests.

PR #3155, opened 2021-12-29
Draft: ThreeCol layout

tim-kilian:
Hej, 

I used xmonad before, where there was I layout called [XMonad.Layout.ThreeColumns](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Layout-ThreeColumns.html). It's a layout similar to tall but with three columns. Mainly the ThreeColMid layout interest me, because the other one is somewhat achievable with the column layout. 

The spawning is a bit different from the one of Xmonad: 
```
|-----|-------|-----|
|  3  |       |  2  | 
|     |       |-----|
|-----|   1   |  4  | 
|  5  |       |-----|
|     |       |  6  | 
|-----|-------|-----|
```

For me it is more intuitive. 

Todo:
- [x] Layout configuration
- [x] Layout icon
- [ ] Add commands for window management
- [ ] Add tests
- [ ] Maybe increase count of main windows?


jwijenbergh:
I like this layout a lot, good work. Did you see https://github.com/qtile/qtile/pull/2606 though? I believe that is the same layout

tim-kilian:
Hej @jwijenbergh, 
no I have not seen that. Thanks for the reference. I believe it is the same intention! I've tested it and it works just like my draft. I also thought that the new_client_position should be configurable tho. Because I do not like the default xmonad spawning.

tim-kilian:
I thought about this again. And maybe my implementation is more like the qtile [Columns](https://docs.qtile.org/en/latest/manual/ref/layouts.html#columns) layout and not the xmonad [ThreeColumns](https://docs.qtile.org/en/latest/manual/ref/layouts.html#columns). I'll rework this to fit as a extension of the Columns layout. Like I said, the current Columns layout does not fit my needs. The main window is not centered there. You can move the windows around, so it is correct, but that is too much work and the orientation should be automated.

zys5945:
Workflow failing at unrelated changes:
`libqtile/layout/three.py:117:34: BLK100 Black would make changes.`

This PR is ready for review / merge

elParaguayo:
Not sure if any work is still happening on this one but, if it is, we may need a new icon as the one in the PR is very like the icon that we've got for `MonadThreeCol`. 

It there's no active work on this then should we close the PR?

PR #3156, opened 2021-12-31
Add command to centre floating windows

elParaguayo:
This is from a discussion in the qtile-dev group: add a new `cmd_center` to position a floating window in the middle of the screen.

Draft for now:
 - [x] needs a test
 - [ ] to discuss how this should behave on non-floating windows

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,6 +17,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           managers to request the name.
         - Fix bug where `Battery` widget did not retrieve `background` from `widget_defaults`.
         - Fix bug where widgets in a `WidgetBox` are rendered on top of bar borders.
+        - Add `lazy.window.center()` command to center a floating window on the screen.
 
 Qtile 0.19.0, released 2021-12-22:
     * features
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -492,6 +492,30 @@ class Window(_Window, metaclass=ABCMeta):
         """
         self.defunct = True
 
+    def cmd_center(self) -> None:
+        """Centers a floating window on the screen."""
+        if not self.floating:
+            return
+
+        if not (self.group and self.group.screen):
+            return
+
+        screen = self.group.screen
+
+        x = (screen.width - self.width) // 2  # type: ignore
+        y = (screen.height - self.height) // 2  # type: ignore
+
+        self.place(
+            x,
+            y,
+            self.width,  # type: ignore
+            self.height,  # type: ignore
+            self.borderwidth,
+            self.bordercolor,  # type: ignore
+            above=True,
+            respect_hints=True,
+        )
+
 
 class Internal(_Window, metaclass=ABCMeta):
     """An Internal window belonging to Qtile."""
--- a/test/test_window.py
+++ b/test/test_window.py
@@ -127,3 +127,24 @@ def test_bring_front_click(manager, bring_front_click):
         assert wins.index(wids[2]) < wins.index(wids[1]) < wins.index(wids[0])
     else:
         assert wins.index(wids[0]) < wins.index(wids[1]) < wins.index(wids[2])
+
+
+@bare_config
+def test_center_window(manager):
+    """Check that floating windows are centered correctly."""
+    manager.test_window("one")
+
+    manager.c.window.set_position_floating(50, 50)
+    manager.c.window.set_size_floating(200, 100)
+    info = manager.c.window.info()
+    assert info["x"] == 50
+    assert info["y"] == 50
+    assert info["width"] == 200
+    assert info["height"] == 100
+
+    manager.c.window.center()
+    info = manager.c.window.info()
+    assert info["x"] == (800 - 200) / 2  # (screen width - window width) / 2
+    assert info["y"] == (600 - 100) / 2  # (screen height - window height) / 2
+    assert info["width"] == 200
+    assert info["height"] == 100


m-col:
It might make more sense as part of the `Floating` layout?

m-col:
rLooks like the floating layout can't be reached in the command graph. So then `Qtile` might be the simplest. I don't think it needs to touch the backend.

re: non-floating windows, the most flexible approach would be to only execute when the current window is floating. Users can always enable floating first if they want it to apply to tiled windows too:
```py
lazy.window.enable_floating(), lazy.centre_window()
```

elParaguayo:
You're supposed to be on holiday!

Agreed, the code is probably in the wrong place, but it can wait until you're back!

elParaguayo:
Thinking about this some more: `lazy.window` does seem to be the appropriate place to call this. I don't think we do any other window positioning from `lazy`.

elParaguayo:
@m-col why are do these lines need to be properties:
https://github.com/qtile/qtile/blob/803dc06fc1f8b121a1d8fe047f26a43812cd427f/libqtile/backend/wayland/window.py#L125-L139

There's no special parsing going on here.

If I try to fix the mypy error here by adding type hinting for `width` and `height` in `libqtile.backend.base.Window` then mypy complains that is clashes with the Wayland typing even though those methods return ints.

m-col:
mypy doesn't like parent and child classes mixing attributes and properties for the same name, even if the functionality is the same. So the base class has properties for these, much like the x11 implementation, but making them attributes in the wayland implementation annoys mypy because it's inconsistent.

elParaguayo:
> mypy doesn't like parent and child classes mixing attributes and properties for the same name, even if the functionality is the same. So the base class has properties for these, much like the x11 implementation, but making them attributes in the wayland implementation annoys mypy because it's inconsistent.

I'm being dumb. How do I fix this?

m-col:
Nah, that was me, I assumed the base classes had width and height properties. Both x11 and wayland implementations independently define properties for these. mypy then sees `self.width` and `self.height` in the base class method and concludes that these are attributes (because if they were properties, they'd be written in the class definition). This is actually a bit of a mistake; the window implementations are required to have `window.width` and `window.height` due to other logic in things like layouts, for example, so the base should define abstract properties for those. _Then_ the wayland implementation needs boring implementations that just proxy to `self._width` and `self._height` because if they were just attributes, mypy will complain about the consistency there (and in fact if they are abstract properties then the implementation can't even be instantiated).

elParaguayo:
> Nah, that was me, I assumed the base classes had width and height properties. Both x11 and wayland implementations independently define properties for these. mypy then sees `self.width` and `self.height` in the base class method and concludes that these are attributes (because if they were properties, they'd be written in the class definition). This is actually a bit of a mistake; the window implementations are required to have `window.width` and `window.height` due to other logic in things like layouts, for example, so the base should define abstract properties for those. _Then_ the wayland implementation needs boring implementations that just proxy to `self._width` and `self._height` because if they were just attributes, mypy will complain about the consistency there (and in fact if they are abstract properties then the implementation can't even be instantiated).

Might be quicker for you to fix that as I'm clearly not understanding the issue as everything I try gives more and more mypy errors!

m-col:
I think this should do it:
```diff
diff --git a/libqtile/backend/base.py b/libqtile/backend/base.py
index 465fbfb8..5b5ae903 100644
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -150,6 +150,16 @@ class _Window(CommandObject, metaclass=ABCMeta):
     def wid(self) -> int:
         """The unique window ID"""
 
+    @property
+    @abstractmethod
+    def width(self) -> int:
+        """Get the window's current width."""
+
+    @property
+    @abstractmethod
+    def height(self) -> int:
+        """Get the window's current height."""
+
     @abstractmethod
     def hide(self) -> None:
         """Hide the window"""

```

elParaguayo:
Yeah, that's what I had tried based on your earlier comments. Gets weird:
```
libqtile/backend/x11/window.py:988: error: Subclass of "_Window" and "Internal" cannot exist: would have incompatible method signatures
libqtile/backend/x11/window.py:990: error: Statement is unreachable
libqtile/backend/x11/window.py:1081: error: Definition of "height" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/x11/window.py:1081: error: Definition of "width" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/x11/window.py:1149: error: Definition of "height" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/x11/window.py:1149: error: Definition of "width" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/x11/window.py:1246: error: Definition of "height" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/x11/window.py:1246: error: Definition of "width" in base class "_Window" is incompatible with definition in base class "_Window"
libqtile/backend/wayland/drawer.py:45: error: Unused "type: ignore" comment
libqtile/backend/wayland/drawer.py:61: error: Unused "type: ignore" comment
libqtile/backend/wayland/drawer.py:62: error: Unused "type: ignore" comment
libqtile/backend/wayland/drawer.py:65: error: Unused "type: ignore" comment
libqtile/backend/wayland/drawer.py:66: error: Unused "type: ignore" comment
```

m-col:
I think this is the best approach currently:
```diff
diff --git a/libqtile/backend/base.py b/libqtile/backend/base.py
index e2b7bffc..6bfb0dd4 100644
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -502,16 +502,16 @@ class Window(_Window, metaclass=ABCMeta):
 
         screen = self.group.screen
 
-        x = (screen.width - self.width) // 2
-        y = (screen.height - self.height) // 2
+        x = (screen.width - self.width) // 2  # type: ignore
+        y = (screen.height - self.height) // 2  # type: ignore
 
         self.place(
             x,
             y,
-            self.width,
-            self.height,
+            self.width,  # type: ignore
+            self.height,  # type: ignore
             self.borderwidth,
-            self.bordercolor,
+            self.bordercolor,  # type: ignore
             above=True,
             respect_hints=True,
         )

```
Obviously ideally we would remove these and actually add these as properties to the base class. However this cascades down quite far into the x11 backend, which is mostly untyped, and I think typing the whole backend is out of scope for this PR :smile:

elParaguayo:
Thanks. I know we try to avoid `# type: ignore` but I agree it's probably the best approach for this PR.

Also, much to my own displeasure, I've changed this to `center` rather than `centre` :wink:

m-col:
> Also, much to my own displeasure, I've changed this to center rather than centre wink

:crying_cat_face: 

Looks great, thanks for this!

PR #3159, opened 2021-12-31
Hide non-commandable windows from IPC

### LINKED ISSUE #3157: Accessing Systray icons via IPC results in crash
From reddit: https://www.reddit.com/r/qtile/comments/rssnm8/conflict_between_system_tray_widget_and/

Posting here as a reminder to fix it.

elParaguayo:
Some windows are not instances of `CommandObject` and will trigger an error if a user attempts to access these via IPC. This PR hides non-commandable windows from the command graph.

Fixes #3157

The alternative approach is to make `libqtile.backend.x11.window._Window` inherit `CommandObject` but that would make Systray icons commandable which I don't think is desirable but happy to discuss here.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -770,7 +770,9 @@ class Qtile(CommandObject):
         elif name == "bar":
             return False, [x.position for x in self.current_screen.gaps]
         elif name == "window":
-            return True, list(self.windows_map.keys())
+            windows: List[Union[str, int]]
+            windows = [k for k, v in self.windows_map.items() if isinstance(v, CommandObject)]
+            return True, windows
         elif name == "screen":
             return True, list(range(len(self.screens)))
         elif name == "core":
@@ -796,7 +798,11 @@ class Qtile(CommandObject):
             if sel is None:
                 return self.current_window
             else:
-                return self.windows_map.get(sel)  # type: ignore
+                windows: Dict[Union[str, int], base._Window]
+                windows = {
+                    k: v for k, v in self.windows_map.items() if isinstance(v, CommandObject)
+                }
+                return windows.get(sel)
         elif name == "screen":
             if sel is None:
                 return self.current_screen
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 Qtile x.x.x, released xxxx-xx-xx:
     * features
         - Add `place_right` option in the TreeTab layout to place the tab panel on the right side
+    * bugfixes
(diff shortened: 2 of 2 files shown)


elParaguayo:
@m-col this is throwing up a mypy error.

If you know why this line:
``` python
windows = [k for k, v in self.windows_map.items() if isinstance(v, CommandObject)]
```
gives this error:
```
2021-12-31T18:08:49.6877550Z libqtile/core/manager.py:774: error: Incompatible return value type (got "Tuple[bool, List[int]]", expected "Optional[Tuple[bool, List[Union[str, int]]]]")
```
I'd be very interested! Either I'm very tired and/or not understanding how typing works but the return value type seems compatible to me.

m-col:
Ha, isn't that a weird one. I think this is because of this: https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance. At least in part, with the inner `List`. Perhaps a type annotation on that line or the line before is sufficient? I'll pull the changes down and have a go at fixing it.

m-col:
This changes fixes it:
```diff
diff --git a/libqtile/core/manager.py b/libqtile/core/manager.py
index eed2ca52..e16a5671 100644
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -770,6 +770,7 @@ class Qtile(CommandObject):
         elif name == "bar":
             return False, [x.position for x in self.current_screen.gaps]
         elif name == "window":
+            windows: List[Union[str, int]]
             windows = [k for k, v in self.windows_map.items() if isinstance(v, CommandObject)]
             return True, windows
         elif name == "screen":
```

libqtile/core/manager.py
```diff
@@ -796,7 +798,10 @@ def _select(self, name: str, sel: Optional[Union[str, int]]) -> Optional[Command
             if sel is None:
                 return self.current_window
             else:
-                return self.windows_map.get(sel)  # type: ignore
+                windows = {
+                    k: v for k, v in self.windows_map.items() if isinstance(v, CommandObject)
+                }
+                return windows.get(sel)  # type: ignore
```
m-col, line 804: Do you know which mypy error is raised here without the comment?

elParaguayo: Something than can be fixed if we add a type hinting line before line 801...

m-col: Nice. I fantasise about type annotating the whole codebase so this makes me happy.

elParaguayo: You should keep that kind of thing to yourself...

m-col:
Awesome, thanks!

PR #3160, opened 2021-12-31
Better error handling in CheckUpdates widget

### LINKED ISSUE #3158: CheckUpdates widget swallows crashes and shows as no updates
As per title, it's not clear if the check update command is working as any error in the command results in the widget treating it as no updates. 

This makes debugging impossible.

elParaguayo:
Currently, any error when checking updates is swallowed silently and displayed as no updates. This PR adds a log message and the ability to display text when an error occurs.

Fixes #3158

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -47,6 +47,7 @@ class CheckUpdates(base.ThreadPoolText):
         ("colour_have_updates", "ffffff", "Colour when there are updates."),
         ("restart_indicator", "", "Indicator to represent reboot is required. (Ubuntu only)"),
         ("no_update_string", "", "String to display if no updates available"),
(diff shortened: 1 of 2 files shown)


elParaguayo:
From IRC:
```
[13:27] <JustineSmit> | elParaguayo: With reference to the issues with the CheckUpdates widget and using custom lines like checkupdates;paru -Qua . I have been
                      informed by the dev that paru will always return an error 1 if it has no results. So running the line checkupdates;paru -Qua will show updates
                      if there are none in the aur even if there are updates in the standard repo. 
```

Is it common for the return code to be 1 if there are no updates? Was that the reason for the current "error" handling?

m-col:
Yeah, makes sense. Just leave it as is? I can see why it might be a bit more annoying to work out an update command, but handling error codes internally also sounds pretty tricky.

PR #3163, opened 2022-01-01
Add references to qtile-examples repo

elParaguayo:
Added links to qtile-examples in  README and FAQs.


m-col:
Good idea, thanks!

PR #3167, opened 2022-01-02
Clicking on bar should focus that screen

### LINKED ISSUE #3165: Clicking on groupbox on different screen does not focus screen
Per IRC:
```
[01:43] <@     tych0> it used to be that clicking on a group in a non-focused monitor would change focus to that monitor
[01:43] <@     tych0> that behavior seems to have changed "recently" (before a708c8cd7ce000a36bb03307b731f46af97251ad at least)
[01:43] <@     tych0> any ideas?
[08:18] < elParaguayo> | tych0 - Firstly, Happy New Year to you.
[08:19] < elParaguayo> | As for your issue, I assume you're talking about clicking on a group in the GroupBox widget
[08:21] < elParaguayo> | I'm not aware of any changes but can't test yet as I don't have access to multiple monitors until tomorrow.
[08:22] < elParaguayo> | Actually, just tried it with fake_screens and I get the same issue. Will log as an issue and take a look.
```

### LINKED ISSUE #3166: With multi monitors, clicking empty screen group does not fully focus screen?
Basically I ran into this with dmenu (additional case below) which I have a keybinding for..
To replicate, you need at least 2 monitors. 
On first monitor open a window, the second monitor should be an empty workspace, click the second monitor workspace to active it, open dmenu. I expected dmenu to open in the second monitor but it opens in the first monitor. If the second monitor has a window open then moving your mouse over will activate the workspace and dmenu will appear in the appropriate workspace.

This is also notable for example, I have a bar in both screens with the CurrentLayout widget on each, if you repeat the scenario above but instead of dmenu go to the second screen and click the CurrentLayout widget which by default cycles through the available layouts, in this case clicking the widget on the second screen actually cycles the layouts on the first monitor.

elParaguayo:
Currently, clicking on a widget on a different bar does not focus that screen. This results in undesirable consequences such as changing group but new windows open on focused screen (#3165) and changing layouts on focused screen (#3166).

This PR addresses that issue by checking the whether the bar's screen is the focused screen when the bar is clicked. If the bar is not on the focused screen then the screen is focused before the click event is passed to the widget.

Fixes #3165
Fixes #3166

Happy to discuss alternative approaches to fixing this (e.g. I'm aware there was a previous discussion about when another screen should be focused).

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -446,6 +446,11 @@ class Bar(Gap, configurable.Configurable):
         return None
 
     def process_button_click(self, x: int, y: int, button: int) -> None:
+        # If we're clicking on a bar that's not on the current screen, focus that screen
+        if self.screen is not self.qtile.current_screen:
+            index = self.qtile.screens.index(self.screen)
+            self.qtile.focus_screen(index, warp=False)
+
         widget = self.get_widget_in_position(x, y)
         if widget:
             widget.button_press(


m-col:
Do you know when the regression was introduced?

elParaguayo:
No. Good point. I could do some bisecting to find out.

elParaguayo:
Hmm. Well it wasn't working on v0.18.1.

m-col:
OK, sounds good. Thanks!

PR #3169, opened 2022-01-02
Quicker window switching

Perlten:
Switch window in the same manner as switching group. Increasing the speed and consistency when working with multiple windows in a group. Also altered the Tasklist for better navigation with this feature.

Config example
```

groups = [Group(i) for i in "123456789"]

for index, i in enumerate(groups, 1):
    keys.extend(
        [
            Key(
                ["mod1"],
                i.name,
                lazy.switch_window(index)
            ),
            Key(
                ["mod1", "shift"],
                i.name,
                lazy.change_window_order(index)
            ),
        ]
    )

```

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -526,6 +526,33 @@ class _Group(CommandObject):
                 return win.info()
 
     @expose_command()
+    def focus_by_index(self, index: int) -> None:
+        """
+        Change to the window at the specified index in the current group.
+        """
+        windows = self.windows
+        if index < 0 or index > len(windows) - 1:
+            return
+
+        self.focus(windows[index])
+
+    @expose_command()
+    def swap_window_order(self, new_location: int) -> None:
+        """
+        Change the order of the current window within the current group.
+        """
+        if new_location < 0 or new_location > len(self.windows) - 1:
+            return
+
+        windows = self.windows
+        current_window_index = windows.index(self.current_window)
+
+        windows[current_window_index], windows[new_location] = (
+            windows[new_location],
+            windows[current_window_index],
+        )
+
+    @expose_command()
     def switch_groups(self, name):
         """Switch position of current group with name"""
         self.qtile.switch_groups(self.name, name)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -29,6 +29,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - New widget: GenPollCommand
         - Add ability to add a group at a specified index
         - Add ability to open the `WidgetBox` widgets at startup.
+        - Add ability to swap focused window based on index, and change the order of windows inside current group
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
@@ -117,6 +118,7 @@ Qtile 0.20.0, released 2022-01-24:
         - Fix bug where `Battery` widget did not retrieve `background` from `widget_defaults`.
         - Fix bug where widgets in a `WidgetBox` are rendered on top of bar borders.
 
+        - Add ability to swap focused window based on index, and change the order of windows inside current group
 Qtile 0.19.0, released 2021-12-22:
     * features
         - Add ability to draw borders to the Bar. Can customise size and colour per edge.
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1384,6 +1384,32 @@ class Qtile(CommandObject):
         mb.start_input(prompt, self.find_window, "window", strict_completer=True)
 
     @expose_command()
+    def switch_window(self, location: int) -> None:
+        """
+        Change to the window at the specified index in the current group.
+        """
+        windows = self.current_group.windows
+        if location < 1 or location > len(windows):
+            return
+
+        self.current_group.focus(windows[location - 1])
+
+    @expose_command()
+    def change_window_order(self, new_location: int) -> None:
+        """
+        Change the order of the current window within the current group.
+        """
+        if new_location < 1 or new_location > len(self.current_group.windows):
+            return
+
+        windows = self.current_group.windows
+        current_window_index = windows.index(self.current_window)
+
+        temp = windows[current_window_index]
+        windows[current_window_index] = windows[new_location - 1]
+        windows[new_location - 1] = temp
+
+    @expose_command()
     def next_urgent(self) -> None:
         """Focus next window with urgent hint"""
         try:
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -164,6 +164,16 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             None,
             "Path to icon theme to be used by pyxdg for icons. ``None`` will use default icon theme.",
         ),
+        (
+            "window_name_location",
+            False,
+            "Whether to show the location of the window in the title.",
+        ),
+        (
+            "window_name_location_offset",
+            0,
+            "The offset given to window loction",
+        ),
     ]
 
     def __init__(self, **config):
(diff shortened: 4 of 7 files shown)


m-col:
Thanks for the submission, and neat code!

There are some similar commands on the `Group` object, such as `cmd_focus_by_name`. Perhaps `cmd_switch_window` should instead be at `Group.cmd_focus_by_index`. Also, I think it would be more intuitive for users if the indexes were 0-based rather than 1-based, for consistency. I realise it's more likely that users will start the keys at 1, skipping 0, but that can be achieved by just using `enumerate(groups)` in the config.

Regarding `cmd_change_window_order`, is the intent here to only move the window within the list held by `Group`? I wonder if this will cause unintended side effects, such as if other code (such as layouts) read windows from there.

Perlten:
I agree that `cmd_focus_by_index` is a better name for the function, and also acknowledge that it creates better consistency if starting at 0

`cmd_change_window_order` is only for changing the order of the windows, giving the user the ability to create habits in regards to window location. For my personal use, I mainly use the functionality with `max` and `columns` and have found no side effects. However, I cannot guarantee that other layouts will behave the same.  
I would imagine that I could mitigate potential side effects, while keeping the feature, by avoiding mutating `windows` yet this would require creating additional state.

m-col:
> cmd_change_window_order is only for changing the order of the windows, giving the user the ability to create habits in regards to window location. For my personal use, I mainly use the functionality with max and columns and have found no side effects. However, I cannot guarantee that other layouts will behave the same.
I would imagine that I could mitigate potential side effects, while keeping the feature, by avoiding mutating windows yet this would require creating additional state.

Looking at the code some more, I think it might work OK without any side-effects. Some good testing should be illuminating here. If you look in `test/layouts/test_common.py` there are some tests that iterate over all of the layouts and test each one. I suggest adding a test there which will use `cmd_change_window_order` when each layout is active and testing that the window order is expected.

Lastly, please could you add a line to the changelog? Then others can find your new feature and use it too!

Perlten:
Renamed functions to cmd_focus_by_index and cmd_swap_window_order. Both functions are now also part of the Group class. Also created tests for swap_window_order that runs on all layouts.

elParaguayo:
Can someone help me out here: what's the advantage of something like this over `lazy.layout.up()` etc. I'm still not clear what the use case is but I'm happy to be educated!

m-col:
> what's the advantage of something like this over lazy.layout.up() etc

My understanding is that this uses absolute rather than relative positions. So moving a window to index 1 cannot be reproduced by moving a window up or down in the layout.

But perhaps @Perlten could speak on this a bit more to clarify

Perlten:
> > what's the advantage of something like this over lazy.layout.up() etc
> 
> My understanding is that this uses absolute rather than relative positions. So moving a window to index 1 cannot be reproduced by moving a window up or down in the layout.
> 
> But perhaps @Perlten could speak on this a bit more to clarify

@m-col @elParaguayo 
Yes, that is correct. Instead of repeatedly pressing a key to get to the correct window, you can access it with a single key from its index position in the list. This means that by organizing your Window locations, you can very quickly change between them. A bit like you already can with Groups.

libqtile/group.py
```diff
@@ -519,6 +519,30 @@ def cmd_info_by_name(self, name):
...
+        """
+        if new_location < 0 or new_location > len(self.windows) - 1:
+            return
+
+        windows = self.windows
+        current_window_index = windows.index(self.current_window)
+
+        temp = windows[current_window_index]
+        windows[current_window_index] = windows[new_location]
+        windows[new_location] = temp
```
m-col, line 544: You can avoid the temp:
```suggestion
windows[current_window_index], windows[new_location] = windows[new_location], windows[current_window_index]
```

m-col:
Apologies for forgetting about this PR. I've added one suggestion but otherwise this LGTM.

@elParaguayo anything to add or are you happy with this?

elParaguayo:
I find this one a bit odd.

The example in [this comment](https://github.com/qtile/qtile/pull/3169#issuecomment-1051306741) confuses the hell out of me. Why are you using type conversions on group names to bind command to select a window?

Also, what's the point of the offset in the windowtabs update?

Perlten:
> The example in [this comment](https://github.com/qtile/qtile/pull/3169#issuecomment-1051306741) confuses the hell out of me. Why are you using type conversions on group names to bind command to select a window?

That example maps the windows in the group the alt+1 alt+2 ... alt+9. The reason for the offset is so that to get the first window in the list, you don't have to press alt+0 but alt+1.

> Also, what's the point of the offset in the windowtabs update?

So you can have the widget display the actual number, you must press to select that specific window. 

I hope that makes it clearer otherwise, let me know :)

elParaguayo:
> > The example in [this comment](https://github.com/qtile/qtile/pull/3169#issuecomment-1051306741) confuses the hell out of me. Why are you using type conversions on group names to bind command to select a window?
> 
> That example maps the windows in the group the alt+1 alt+2 ... alt+9. The reason for the offset is so that to get the first window in the list, you don't have to press alt+0 but alt+1.

But you're not mapping windows in the group. You're just using the group name as a way to get the numbers 1-9. From what I can tell, this has nothing to do with groups. If your config had group names "qwertyuiop" then that code wouldn't work.

Perlten:
> > > The example in [this comment](https://github.com/qtile/qtile/pull/3169#issuecomment-1051306741) confuses the hell out of me. Why are you using type conversions on group names to bind command to select a window?
> > 
> > 
> > That example maps the windows in the group the alt+1 alt+2 ... alt+9. The reason for the offset is so that to get the first window in the list, you don't have to press alt+0 but alt+1.
> 
> But you're not mapping windows in the group. You're just using the group name as a way to get the numbers 1-9. From what I can tell, this has nothing to do with groups. If your config had group names "qwertyuiop" then that code wouldn't work.
> 
> ...or I'm really not understanding this code!


Ahh looking at it propper I agree with you. The way it should have been done was to use the index instead of the name (in my case it just happened to line up)

A better approach would be 

`
groups = [Group(i) for i in "123456789"]

for index, g  in enumerate(groups):
    keys.extend(
        [
            Key(
                ["mod1"],
                g.name,
                lazy.group.focus_by_index(index - 1),
            ),
            Key(
                ["mod1", "shift"],
                g.name,
                lazy.group.swap_window_order(index- 1),
            )
        ]
    )

`

elParaguayo:
Ok, I see, you're binding these to the same keys you use for switching groups, just with different modifiers.

No other comments.

You'll need to fix the conflict in CHANGELOG (by fetching latest master and rebasing your branch) then we should be good to go.

Perlten:
Great I will fix the issues you both brought up tonight and let you know when it's done.

Perlten:
@elParaguayo @m-col 

Now everything should be in order let me know if otherwise 

Took the liberty to add .vimspector.json to gitignore hope that was okay

Perlten:
@m-col  @elParaguayo  Any other feedback or changes that needs to be made ? :)

m-col:
Apologies for the delay - code all looks good to me! I've enabled the CI and auto-merge, so it will be merged if it all passes. Thanks again!

elParaguayo:
There's a conflict in `CHANGELOG` so you'll need to fetch the latest master and rebase your branch. You also need to run `black` as there's a formatting change needed in `tasklist.py`.

Perlten:
@m-col  @elParaguayo 

Now everything should be in order. Let me know if that is not the case :)

elParaguayo:
There still seem to be conflicts (but github's not telling me what they are). Can you make sure your master branch is up to date with upstream and then rebase your branch.

Perlten:
Did that seem to fix the problem ?

elParaguayo:
No. I'll clone your repo and see if I can fix the conflicts.

elParaguayo:
Sorry for the delay on this.

I've rebased the PR so that should clear the conflicts. Let's see how the tests get on...

elParaguayo:
Tidied this all up. Good to merge now.

PR #3174, opened 2022-01-03
Force redrawing of bar borders on expose event

### LINKED ISSUE #3173: Bar border fails to redraw on some applications
I was trying out the [kate](https://kate-editor.org) editor and I noticed that the bar border fails to redraw after the application has been toggled fullscreen or just exited when fullscreen. Not sure how it's possible that it only happens on some applications, but so far only kate does this for me. Maybe several other KDE related apps do this? I mostly used terminal applications so I don't know.

This bug was introduced by #3067 so it happens at current master
borderglitch

The border will never redraw unless I restart qtile.

Happens with the default config that have bar border settings enabled.

elParaguayo:
Borders are currently only drawn once to save processing time. This causes an issue when the bar is hidden (e.g. by a fullscreen
window) and then uncovered again.

This PR fixes the issue by forcing a dedrawing of the border whenever the bar's `expose` event is triggered.

Fixes #3173

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -276,7 +276,7 @@ class Bar(Gap, configurable.Configurable):
             self.drawer = self.window.create_drawer(width, height)
             self.drawer.clear(self.background)
 
-            self.window.process_window_expose = self.draw
+            self.window.process_window_expose = self.process_window_expose
             self.window.process_button_click = self.process_button_click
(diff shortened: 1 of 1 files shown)


jwijenbergh:
Yep, fixed for all my problematic apps: lockscreen, kate and kdevelop

m-col:
Ah yeah, makes sense. Thanks for the fix!

PR #3176, opened 2022-01-03
Update cpu.py

IagoNicolas:
From psutil 5.9.0 changelog:
  - 1851_, [Linux]: `cpu_freq()`_ is slow on systems with many CPUs. Read current
  frequency values for all CPUs from ``/proc/cpuinfo`` instead of opening many
  files in ``/sys`` fs.  (patch by marxin)

Adding a check to make it compatible with older and newer versions of psutil, since we don't need to divide by 1000 anymore.

### DIFF
--- a/libqtile/widget/cpu.py
+++ b/libqtile/widget/cpu.py
@@ -50,7 +50,10 @@ class CPU(base.ThreadPoolText):
 
         variables["load_percent"] = round(psutil.cpu_percent(), 1)
         freq = psutil.cpu_freq()
-        variables["freq_current"] = round(freq.current / 1000, 1)
+        if(float(psutil.__version__[:-2]) <= 5.8):
+            variables["freq_current"] = round(freq.current / 1000, 1)
+        else:
+            variables["freq_current"] = round(freq.current, 1)
         variables["freq_max"] = round(freq.max / 1000, 1)
         variables["freq_min"] = round(freq.min / 1000, 1)
 


elParaguayo:
Thanks.

Are they really changing the units that are output by `psutil.cpu_freq()`? Seems like a very odd thing to do and it's not implied by the changelog entry.

If this is correct then this will break `test_cpu` as we mock `psutil` and the mock doesn't include a `__version__` attribute. The tests should test both versions.

IagoNicolas:
Seems like it, i ran the code below only changing psutil between 5.9.0 and 5.8.0:
`import psutil

print("psutil version: " + psutil.__version__)
freq = psutil.cpu_freq()
print(freq)`
Got the following output from 5.9.0:
psutil version: 5.9.0
scpufreq(current=0.871125, min=400.0, max=3600.0)
and from 5.8.0:
psutil version: 5.8.0
scpufreq(current=2851.37675, min=400.0, max=3600.0)

elParaguayo:
But `min` and `max` are the same so how can `current` be different units?

The docs for 5.8 and 5.9 both say
```
Return CPU frequency as a named tuple including current, min and max frequencies expressed in Mhz.
```

So I'm still confused!

IagoNicolas:
Seems [https://github.com/giampaolo/psutil/commit/f18438d135c12f7eb186f49622e0f6683c37f7f5](this commit) messed a bit with the way they get cpu data.

IagoNicolas:
Seems like they used to read data from /sys/devices/system/cpu/cpufreq/policy[0-n]/scaling_cur_freq, which needs to be divided by 10^6 to be used in the widget, while now they read from /proc/cpuinfo, which needs only to be divided by 10^3 to be used in the widget.

elParaguayo:
I don't disagree with what you're saying. My only point is that it shouldn't be the widget that's responsible for the conversion. Unless I'm missing something, there's an error in psutil.

The units for `current`, `min` and `max` should all be the same i.e. we should expect the same output from `cpu_freq` between versions unless they explicitly state there's a change in the units (for `current` - as you can see from your testing above, `min` and `max` still return the same values), which they haven't done.

I'd be inclined to check this with their developers first before making a change here.

elParaguayo:
It's been submitted as a bug in psutil: https://github.com/giampaolo/psutil/issues/2049

elParaguayo:
I think we can close this. `pustil` developers are working on a fix.

PR #3179, opened 2022-01-05
Add toggle argument to Window.togroup

### LINKED ISSUE #3175: [Feature Request] Window.togroup with toggle
The `toscreen` method of `Group` has a `toggle` flag that allows toggling to the previous group if the target group is already shown on screen.

It would be nice to have this flag available on the `togroup` method of `Window` as well (and behaving correctly with the `switch_group` flag).

Use case: I use `mod4 + number` to navigate in groups, with toggling activated. I use `shift + mod4 + number` to move the active window in groups, with group switching activated. I would like window movement to behave consistently with group navigation w.r.t. toggling.

qwenger:
Fix #3175

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Qtile x.x.x, released xxxx-xx-xx:
         - Add `lazy.screen.set_wallpaper` command.
         - Added ability to scale the battery icon's size
         - Add Spiral layout
+        - Add `toggle` argument to `Window.togroup` with the same functionality as in `Group.toscreen`.
     * bugfixes
         - Fix `Systray` crash on `reconfigure_screens`.
         - Fix bug where widgets can't be mirrored in same bar.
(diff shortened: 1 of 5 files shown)


elParaguayo:
This looks ok to me.

I've rebased locally but can't push to your fork as you haven't given write access to maintainers.

Can you either give me access or do the rebasing yourself?

qwenger:
Do you still consider this pull request? From what I understood from your colleague @m-col (https://github.com/qtile/qtile/issues/3175#issuecomment-1013914505) they prefer not to integrate it...

elParaguayo:
We've been discussing it some more. While it's not something that I'd want to use, I do see understand your logic for wanting it. So, we're ok to include it.

qwenger:
Alright then.

qwenger:
Thanks!

m-col:
Yeah sorry about that, I should have followed up here following @elParaguayo and I's offline discussion. Happy to have it :)

PR #3184, opened 2022-01-09
Fix finalisation of layouts

### LINKED ISSUE #3171: If TreeTab layout is active when removed from the config and reloading causes qtile to enter into invalid state.
I rant into this whilst experimenting with my config. if TreeTab layout is active while being removed from the config and reloading, it causes the TreeTab panel to stay on permanently, requiring a qtile restart.

elParaguayo:
Previously, when finalising layouts, the manager loops over `self.config.layouts`. However, layouts are cloned per group and so the config layouts do not represent the layouts that are actually in use.

This is particularly problematic for the `TreeTab` layout which creates an `Internal` window for the panel which needs to be killed when the layout is finalised.

This PR fixes this issue by looping over layouts in each group to finalise them. It also ensures that the `TreeTab._panel` is killed.

Fixes #3171

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -281,8 +281,10 @@ class Qtile(CommandObject):
                 widget.finalize()
             self.widgets_map.clear()
 
-            for layout in self.config.layouts:
-                layout.finalize()
+            # For layouts we need to finalize each clone of a layout in each group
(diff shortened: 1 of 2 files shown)


m-col:
Of course! Thanks!

PR #3187, opened 2022-01-11
Place upper bound limit of pywlroots version

flacjacket:
The latest version of pywlroots has some breaking changes that will need
to be accounted for here before we can support that version.


elParaguayo:
Thanks for the heads-up!

PR #3198, opened 2022-01-15
Fix icon order in Systray

elParaguayo:
To prevent icon order changing after restart/reloading config, icons are now ordered by application name.

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -42,7 +42,14 @@ class Icon(window._Window):
     def __init__(self, win, qtile, systray):
         window._Window.__init__(self, win, qtile)
         self.systray = systray
+        self.name = win.get_name()
         self.update_size()
+
+    def __eq__(self, other):
+        if not isinstance(other, Icon):
+            return False
+
+        return self.window.wid == other.window.wid
 
     def update_size(self):
         icon_size = self.systray.icon_size
@@ -73,8 +80,9 @@ class Icon(window._Window):
 
     def handle_DestroyNotify(self, event):  # noqa: N802
         wid = event.window
+        icon = self.qtile.windows_map[wid]
+        self.systray.tray_icons.remove(icon)
         del self.qtile.windows_map[wid]
-        del self.systray.icons[wid]
         self.systray.bar.draw()
         return False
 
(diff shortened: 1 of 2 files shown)


m-col:
It might be a bit nicer to order them at the type level so that the sorting doesn't need to happen upon every draw. I see the `wid`s used for keys in `self.icons` aren't all that useful; they can be accessed at `icon.wid` (or `icon.window.wid`) during `finalize()`. Then the keys can be the window names, which would be used for ordering in an `OrderedDict` IIRC.

elParaguayo:
`OrderedDict`s remember the order items were added to the dict, rather than sorting them. You'd still need to sort each time you add an item to the dictionary (but that's probably still better than sorting on each draw).

elParaguayo:
Also looks like one of my new tests is playing up...

m-col:
> OrderedDicts remember the order items were added to the dict, rather than sorting them. 

Oh indeed, TIL, OrderedDicts are mostly redundant.

> You'd still need to sort each time you add an item to the dictionary (but that's probably still better than sorting on each draw).

Yeah this seems like the way to go, just running `sorted` on the dict when it an item is added.

libqtile/widget/systray.py
```diff
@@ -107,16 +109,16 @@ class Systray(window._Window, base._Widget):
     def __init__(self, **config):
         base._Widget.__init__(self, bar.CALCULATED, **config)
         self.add_defaults(Systray.defaults)
-        self.icons = {}
+        self.tray_icons = []
```
m-col, line 112: A list doesn't feel right here because it isn't fully type safe. If a misbehaving client sends 2 ClientMessages the list will end up with 2 icons for the same `wid`, which will then cause problems. I'm thinking that that was actually a benefit to using the wids as dictionary keys. So perhaps it could be `Dict[int, Icon]` with wid keys, and then we can sort it with:
```py
self.tray_icons = dict(sorted(self.tray_icons.items(), key=lambda item: item[1].name))
```
It's kinda ugly... but it's a bit more safe I think.

elParaguayo: If you're assuming that multiple messages have the same `wid` then the Icon object should be the same so we could add an `__eq__` method to `Icon` and then just do:
``` python
if icon not in self.tray_icons:
```
That would look a bit neater than your line (but is more code)!

m-col: Interesting. Yeah that sounds neat.

m-col:
LGTM, thanks! Auto merge enabled :)

PR #3203, opened 2022-01-16
Change Volume widget's handling of theme_path

elParaguayo:
By checking `theme_path` and loading images in `__init__`, the widget is not able to access any global defaults set in the config as the widget is initialised before the defaults are added.

This PR addresses this by delaying the checking until the widget is configured.


m-col:
Perhaps I'm being dim, but I don't see why this change would fix it. Could you possibly explain?

elParaguayo:
> Perhaps I'm being dim, but I don't see why this change would fix it. Could you possibly explain?

Not being dim at all. Took me a while to understand why this was happening!

In short, global defaults are defined in `config.py` but this is where widgets are also initialised. As a result, widgets are initialised before Qtile can set the global defaults in every widget and extension. That all makes sense.

The problem with the `Volume` widget is that it tries to retrieve the `theme_path` attribute in `__init__`. At this point, only the widget's own `defaults` are available and this is `None` so this becomes the value of that attribute. Setting the global defaults later does not overwrite any existing value, it just becomes an extra value that can be retrieved when `__getattr__` is first called for that attribute.

By moving the check for the `theme_path` attribute to `_configure`, the global default is now available to the widget.

Hopefully that made sense.

m-col:
What I don't quite understand is why setting the length to 0 during _configure changes anything

elParaguayo:
> What I don't quite understand is why setting the length to 0 during _configure changes anything

That's not the relevant change.

The important bit is this line:
``` python
if self.theme_path:
```
That's what calls `__getattr__` for the `Configurable`. If `theme_path` is not set in the parameters when creating the widget then it falls back to the widget's default or the global default _if it's set_. During `__init__`, the global default isn't available yet, as described above, so it uses the widget default. After this line, `Volume.theme_path` is now set and so the lookup in defaults and global defaults no longer happens.

m-col:
I see! Very magical. LGTM, thanks!

PR #3204, opened 2022-01-16
Prevent multiple instances of Systray

### LINKED ISSUE #2336: multiple systray widgets are racy
The Distrotube Guy reports via https://www.youtube.com/watch?v=9NO2DJARbU0 that multiple systrays are racy.

The spec says: https://specifications.freedesktop.org/systemtray-spec/systemtray-spec-0.1.html#locating

that we should use `_NET_SYSETM_TRAY_Sn` for the nth system tray widget, but I think our code calculating this is wrong:

https://github.com/qtile/qtile/blob/master/libqtile/widget/systray.py#L129

This code always uses qtile's current screen, instead of the screen that the bar is on. It should at least use the right value for `n`.

Perhaps instead, we could make multiple Systray() objects be an error, and just have people specify one and let it mirror everywhere.

elParaguayo:
The System Tray specification does not appear to support multiple trays. This can lead to issues where multiple trays are created.

This PR addresses the issue by introducing a mechanism to specify the maximum number of a type of widget. Where the number of instances exceed that limit, the widget is removed from the bar.

Fixes #2336

There may be better ways to deal with this so we can discuss here.

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -33,6 +33,7 @@ from xcffib.xproto import ClientMessageData, ClientMessageEvent, EventMask, SetM
 
 from libqtile import bar
 from libqtile.backend.x11 import window
+from libqtile.confreader import ConfigError
 from libqtile.widget import base
 
 if TYPE_CHECKING:
@@ -99,6 +100,9 @@ class Systray(window._Window, base._Widget):
     """
     A widget that manages system tray.
 
+    Only one Systray widget is allowed. Adding additional Systray
+    widgets will result in a ConfigError.
+
     .. note::
         Icons will not render correctly where the bar/widget is
         drawn with a semi-transparent background. Instead, icons
@@ -108,6 +112,8 @@ class Systray(window._Window, base._Widget):
         a fully opaque background colour or a fully transparent
         one.
     """
+
+    _instances = 0
 
     _window_mask = EventMask.StructureNotify | EventMask.Exposure
 
@@ -136,6 +142,9 @@ class Systray(window._Window, base._Widget):
 
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
+
+        if Systray._instances > 0:
+            raise ConfigError("Only one Systray can be used.")
 
         if self.configured:
             return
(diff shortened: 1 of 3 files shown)


m-col:
Is it likely that any other widget is going to need to limit its instances? It seems like general infrastructure and widget-writing API that is written specifically for one widget's limitation. An alternative approach might be something like:
```diff
diff --git a/libqtile/widget/systray.py b/libqtile/widget/systray.py
index a32a2869..78c3db5d 100644
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -29,7 +29,7 @@
 import xcffib
 from xcffib.xproto import ClientMessageData, ClientMessageEvent, EventMask, SetMode
 
-from libqtile import bar
+from libqtile import bar, confreader
 from libqtile.backend.x11 import window
 from libqtile.widget import base
 
@@ -99,6 +99,8 @@ class Systray(window._Window, base._Widget):
 
     orientations = base.ORIENTATION_BOTH
 
+    _instances = 0
+
     defaults = [
         ("icon_size", 20, "Icon width"),
         ("padding", 5, "Padding between icons"),
@@ -120,6 +122,9 @@ class Systray(window._Window, base._Widget):
         return length
 
     def _configure(self, qtile, bar):
+        if Systray._instances > 0:
+            raise confreader.ConfigError("Only one Systray can used at a time.")
+
         base._Widget._configure(self, qtile, bar)
 
         if self.configured:
```
I realise this also means that a `ConfigErrorWidget` will be created, but I think that might be a good thing - "Why is the systray on my second monitor not showing any icons?". This way users know something is wrong, the error message is informative, and it calls the user to do something. Thoughts?

elParaguayo:
That's exactly how I did it the first time!

I don't think that catches Mirrors though which is one of the reasons I changed my approach.

m-col:
Maybe override `create_mirror` on that widget to raise the error, or to just return itself?

Did we already have this conversation?

libqtile/widget/systray.py
```diff
@@ -170,6 +178,21 @@ def _configure(self, qtile, bar):
...
+        Systray cannot be mirrored as we do not use a Drawer object to render icons.
+
+        Return a fake widget which will crash during `_configure` so user will get a
+        ConfigErrorWidget in their bar instead.
+        """
+        class BrokenMirror(base._Widget):
+            def _configure(self, qtile, bar):
+                raise ConfigError("Systray widget cannot be mirrored.")
+
+        return BrokenMirror(0)
```
elParaguayo, line 194: 7 years' bad luck! :wink:

m-col: Is this necessary? I'm wondering, if you did just:
`return self`
Then it would run `_configure` on itself again and `ConfigError` that way. Might be slightly cleaner? Also consider when this happens, somebody might be like "OK, guess I'll make a second instance instead of passing the same one to two bars", then they hit a new config error. Thoughts?

elParaguayo: Oh, yeah. That's much neater!

m-col:
Cool, looks good. Perhaps this warrants a changelog entry?

m-col:
Great thanks!

Issue #3188, opened 2022-01-11
Restarting Qtile while browsers running sometimes results in audio control to break.


agoodfellow123:
Hello there, while I was configuring my Qtile I realized my sound widget was showing "M" for mute. I tried to unmute it but it did not work after a while I opened a video and realized that sound plays perfectly fine despite the widget showing "M", the only broken thing is the widget and volume control hotkeys. After another while, I closed one of my browsers and realized that my widget started to work again. However, that's a fifty-fifty solution because sometimes closing all the windows and even restarting after closing them doesn't solve this issue. Also running pulsemixer in the terminal when this bug occurs gives the "Failed to connect to pulseaudio" error.

elParaguayo:
Thanks. Are you using the `PulseVolume` widget?

agoodfellow123:
No, I'm using the `volume` widget.

elParaguayo:
Are there any error messages in your logs.

agoodfellow123:
Yeah there are numerous errors but I don't know which one is related to this: http://ix.io/3LV6

elParaguayo:
That link just gives me the path to your log, rather than the contents of the log.

agoodfellow123:
I am sorry about that here is the right one: http://ix.io/3LV9

elParaguayo:
Are you the same person who posted about the clock on reddit?

agoodfellow123:
No.

elParaguayo:
OK. Well you've got the same function and variable names from their code in your error...

Anyway, can you share your config line for the volume widget. The error suggests there's a bad mouse callback configured there.

agoodfellow123:
Ok, you caught me I'm the Reddit guy, so much for privacy :/
Here is the volume widget: `widget.Volume(step=5, mouse_callbacks={'Button3': lambda: qtile.cmd_spawn(terminal + ' -e pulsemixer')}),`
I should add though, that error did happened before adding mouse_callbacks too.

elParaguayo:
I've tidied up the thread a bit.

The callback errors say that spawn couldn't find `exec` so I'm not sure what you've got set as `terminal` in your config. 

Anyway, if the errors happened before then I wonder if it's to do with the rapid polling that's done by this widget. What happens if you change `update_interval`?

It's also possible that it's not qtile that's broken here and that your browser is locking the audio device when it restarts so the widget can no longer poll the volume command to get current levels.

agoodfellow123:
[...]

PR #3208, opened 2022-01-16
Fix rounding error in PulseVolume's reported volume

m-col:
See: https://github.com/qtile/qtile/issues/3188#issuecomment-1013926781


elParaguayo:
LGTM

PR #3209, opened 2022-01-17
WL: Bump wlroots to 0.15.0

m-col:
This updates the wlroots version that Qtile's wayland backend requires
to 0.15.x. Older versions are incompatible. pywlroots also needs to be
version 0.15.x.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,11 +9,11 @@ jobs:
         runs-on: ubuntu-20.04
         name: "python ${{ matrix.python-version }}"
         env:
-          WAYLAND: 1.19.0
-          WAYLAND_PROTOCOLS: 1.21
-          WLROOTS: 0.14.0
-          SEATD: 0.5.0
-          LIBDRM: 2.4.105
+          WAYLAND: 1.20.0
+          WAYLAND_PROTOCOLS: 1.24
+          WLROOTS: 0.15.0
+          SEATD: 0.6.3
+          LIBDRM: 2.4.109
         strategy:
             matrix:
                 # if you change one of these, be sure to update
@@ -69,8 +69,8 @@ jobs:
                 sudo ninja -C build install
             - name: Build wlroots
               run: |
-                wget -q --no-check-certificate https://github.com/swaywm/wlroots/archive/$WLROOTS.tar.gz
-                tar -xzf $WLROOTS.tar.gz
+                wget -q --no-check-certificate https://gitlab.freedesktop.org/wlroots/wlroots/-/archive/$WLROOTS/wlroots-$WLROOTS.tar.gz
+                tar -xzf wlroots-$WLROOTS.tar.gz
                 cd wlroots-$WLROOTS
                 meson build -Dexamples=false --prefix=/usr
                 ninja -C build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,9 @@ jobs:
                   libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz \
                   imagemagick libpulse-dev git xserver-xephyr xterm xvfb ninja-build libegl1-mesa-dev \
                   libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev libpixman-1-dev libpciaccess-dev \
-                  dbus-x11 libnotify-bin
+                  dbus-x11 libnotify-bin \
+                  xwayland libxcb-composite0-dev libxcb-icccm4-dev libxcb-res0-dev \
+                  libxcb-render0-dev libxcb-res0-dev libxcb-xfixes0-dev 
                 sudo pip -q install meson PyGObject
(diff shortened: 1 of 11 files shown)


m-col:
Note that this will require updating wlroots too. I'm not really sure when the best time to merge this would be, but perhaps right before a release - which we should do soon? We've just fixed a few bugs too which is nice. So if there are any other bug fixes that'd be good to get out then we can merge those, this, then release again.

m-col:
It had a small bug but I just pushed a fix. Working perfectly now.

m-col:
Do you mind if we merge this then push a new release so our package can be upgraded with the wlroots dep?

elParaguayo:
> Do you mind if we merge this then push a new release so our package can be upgraded with the wlroots dep?

Can I fix `WidgetBox`\`SysTray` issue first (and probably the notification issue too)?

m-col:
> Can I fix WidgetBox`SysTray` issue first (and probably the notification issue too)?

Yeah it'll be good to get that one in!

m-col:
The tests are going crazy and I am not sure why. Every wayland process seems to be crashing upon exit. When I run it locally using `./scripts/wephyr` I see I am getting a "double free or corruption" so I suspect `multiprocess` is capturing that are reporting the traceback. Will investigate.

m-col:
Sorry for the noise, this kept having tests problems, I should have fixed it all on my fork :smile: This is ready for review now.

elParaguayo:
No problem.

I can see `wlroots` is already at 0.15 in Arch repos. `pywlroots` is still 0.14.12 so I've flagged it as being out of date.

In terms of testing, what should I be looking out for?

m-col:
Interesting.... then I wonder if that means qtile wayland in the repos is currently non-functional.

> In terms of testing, what should I be looking out for?

Mainly just that things work as expected, the update don't change much, but maybe also check xwayland is working with both regular windows and override-redirect windows?

elParaguayo:
Ah, sorry. `wlroots` 0.15 is in `community-staging` not main repo.

m-col:
`pywlroots` already has the update for this, so I expect that when we release, they'll update qtile and pywlroots at the same time in their repos. I see sway just updated for wlroots 0.15.0 today too.

justinesmithies:
Both wlroots and python-wlroots for Arch have updated versions sitting in the community-staging area right now. So I guess they are maybe waiting for Qtile ?

m-col:
Yep, possibly. Gentle ping @elParaguayo when you get a moment pls could you review this :smile:

elParaguayo:
I don't know if they know that Qtile won't work with those versions so they may well release them before Qtile.

@m-col how did you install wlroots 0.15 if it's not in repos?

Can try to test quickly later today.

m-col:
I run wlroots from a local build checked out at that version. I think you might be able to give pacman a list of packages to fetch from a specific repository, and get wlroots from community-staging, though I'm not sure about that. Otherwise building locally is pretty painless. Just install the wlroots package to get the deps, then clone, checkout, and build with meson. Then to run I think you need to use e.g. `export LD_LIBRARY_PATH=~/git/wlroots/build` so that wlroots is used from there.

elParaguayo:
OK - will take a look at local build.

Arch wiki says don't add staging as a repo unless you want to break your system!

m-col:
Ha, yes, seems reasonable.

arjan-s:
I haven't tried it, but I would expect the command below to work as well, bypassing the need to build wlroots yourself. However, it will replace your global wlroots, possibly making your system unusable.
```
sudo pacman -U https://mirror.rackspace.com/archlinux/community-staging/os/x86_64/wlroots-0.15.0-4-x86_64.pkg.tar.zst
```

elParaguayo:
Yeah, I was thinking about doing that.

I've built wlroots locally from git but I can't get `pywlroots>=0.15` to build.

arjan-s:
It's in community-staging as well, so I just added it to the command in my previous comment

m-col:
> I've built wlroots locally from git but I can't get pywlroots>=0.15 to build.

What happens when you try?

elParaguayo:
`wlr/render/allocator.h: No such file or directory`

elParaguayo:
separate to this, I didn't build wlroots with xwayland support. Does that matter? and, if so, what if Arch doen't build it with with xwayland support?

m-col:
Arch do build wlroots with xwayland support, and pywlroots builds with xwayland support if wlroots was. @arjan-s's command above would certainly be the easiest way unless you also want the git version of wlroots around. You may need these for building:
```sh
export CFLAGS="-I $HOME/git/wlroots/include -I $HOME/git/wlroots/build/include"
export LDFLAGS="-L $HOME/git/wlroots/build"
```

elParaguayo:
@m-col Good news, it's compiled ok. Still have some comments/questions though.

Rather than spamming this thread with basic questions, I might post in our chat.

elParaguayo:
@m-col as discussed, I think we can merge this PR as is (as it maintains core Wayland functionality) but we flag that Xwayland support is "experimental" as there are some issues that need to ironed out. Depending on how easy those fixes are, we can push a point release once they're ready.

PR #3214, opened 2022-01-17
Fix Systray icon ordering - v2

elParaguayo:
This PR tries to fix the order of icons in Systray by sorting by application name. However, in certain circumstances, no window
name is available so we use the window's WID as a name in order to be able to sort the list of icons.

@elcaven - could you check that this fixes the issue you and I discussed in IRC?

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -47,7 +47,16 @@ class Icon(window._Window):
     def __init__(self, win, qtile, systray):
         window._Window.__init__(self, win, qtile)
         self.systray = systray
+        # win.get_name() may return None when apps provide a temporary window before the icon window
+        # we need something in self.name in order to sort icons so we use the window's WID.
+        self.name = win.get_name() or str(win.wid)
         self.update_size()
+
+    def __eq__(self, other):
+        if not isinstance(other, Icon):
+            return False
+
+        return self.window.wid == other.window.wid
 
     def update_size(self):
         icon_size = self.systray.icon_size
@@ -78,8 +87,8 @@ class Icon(window._Window):
 
     def handle_DestroyNotify(self, event):  # noqa: N802
         wid = event.window
-        del self.qtile.windows_map[wid]
-        del self.systray.icons[wid]
+        icon = self.qtile.windows_map.pop(wid)
+        self.systray.tray_icons.remove(icon)
         self.systray.bar.draw()
         return False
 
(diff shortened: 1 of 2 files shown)


libqtile/widget/systray.py
```diff
@@ -47,8 +47,17 @@ class Icon(window._Window):
     def __init__(self, win, qtile, systray):
         window._Window.__init__(self, win, qtile)
         self.systray = systray
+        # win.get_name() may return None when apps provide a temporary window before the icon window
+        # we need something in self.name in order to sort icons so we use the window's WID.
+        self.name = win.get_name() or str(win.wid)
```
m-col, line 52: Is it a 'temporary window', or have they just not set any appropriate hints yet?

libqtile/widget/systray.py
```diff
@@ -78,8 +87,9 @@ def handle_PropertyNotify(self, e):  # noqa: N802
 
     def handle_DestroyNotify(self, event):  # noqa: N802
         wid = event.window
+        icon = self.qtile.windows_map[wid]
+        self.systray.tray_icons.remove(icon)
         del self.qtile.windows_map[wid]
```
m-col, line 92:
```suggestion
        icon = self.qtile.windows_map.pop(wid)
        self.systray.tray_icons.remove(icon)
```

m-col (approved):
LGTM

libqtile/widget/systray.py
elParaguayo: It has a different WID so I'm assuming temporary.

libqtile/widget/systray.py
elParaguayo: Nice!

libqtile/widget/systray.py
m-col: Hmm, I wonder what they are up to...

elParaguayo: Not sure but, when I launch `blueman-applet`, I first get a window with the class `blueman-tray` then it's destroyed and replaced by one with a `blueman` class. Who knows what's going on but they are different windows.

elcaven:
> This PR tries to fix the order of icons in Systray by sorting by application name. However, in certain circumstances, no window name is available so we use the window's WID as a name in order to be able to sort the list of icons.
> 
> @elcaven - could you check that this fixes the issue you and I discussed in IRC?

Looks good to me, after a restart the systray loads up fine and it stays good after opening apps that previously made it crash.

elParaguayo:
Awesome. Thanks for confirming.

PR #3218, opened 2022-01-18
WL: Implement XWayland support

m-col:
This is adds support for X-only clients through XWayland.

XWayland must be installed on the system. It's lazy to startup, so will
only be started when the first X connection is made. This adds a bit of
delay to that first client's startup, but it's better there than at
Qtile's startup. Users could start it up themselves in a startup hook to
avoid the delay.

Override-redirect windows (such as rofi, dmenu) are managed as `Static`
windows, which seems to work fine. Everything else is a `Window`.

Support is provided by wlroots, via pywlroots, which must have been
built with XWayland support (i.e. if XCB was present on the system when
it was build).

--

Currently this requires this PR: https://github.com/flacjacket/pywlroots/pull/69

Any feedback and live testing is greatly appreciated. I do not use any X-only clients myself, but it would be good to test many different clients to check they all work as expected.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -61,6 +61,7 @@ from wlroots.wlr_types.server_decoration import (
 )
 from wlroots.wlr_types.virtual_keyboard_v1 import VirtualKeyboardManagerV1, VirtualKeyboardV1
 from wlroots.wlr_types.xdg_shell import XdgShell, XdgSurface, XdgSurfaceRole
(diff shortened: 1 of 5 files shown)


m-col:
xterm and rofi working as expected:
20220118_23h55m26s_grim

m-col:
I've moved the feature over to https://github.com/qtile/qtile/pull/3209 to make it easier for myself (both need updated and specific versions of pywlroots),

PR #3222, opened 2022-01-19
Remove docs test from CI

elParaguayo:
We now have a hook to ReadTheDocs so that docs are built for pull requests and a status is shown for successful/failed builds.

The `docs` test has been removed from CI but is still available in `tox.ini` so users can still run `tox -e docs` to test locally.


m-col:
Very nice, thanks for setting up the neat feature!

PR #3225, opened 2022-01-20
Fix Systray bug in WidgetBox

### LINKED ISSUE #3224: widget.box does not close
Hi,

  I'm running qtile on arch and after the latest update the widget.box does not close.  It was working fine before the update.  

qtilewidgetbox1

qtilewidgetbox2

elParaguayo:
9069a0e broke WidgetBox if there was a Systray widget in the box as tray icons were not removed correctly. This was due to renaming the list that contains the icons.

This PR fixes the issue by ensuring that icons are correctly hidden when closing the box.

Fixes #3224

### DIFF
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -120,7 +120,7 @@ class WidgetBox(base._Widget):
                 # are separate _Window instances.
                 # Systray unhides icons when it draws so we only need to hide them.
                 if isinstance(widget, Systray):
-                    for icon in widget.icons.values():
+                    for icon in widget.tray_icons:
                         icon.hide()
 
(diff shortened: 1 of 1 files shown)


m-col:
Cool. A bit surprised mypy didn't catch this but I think it didn't because it only analyses functions with annotations. We need to increase that coverage!

elParaguayo:
mypy was part of the problem. As `Systray` inherits from `_Window` it has an `icons` attribute already which was a `dict`. As I changed this to a list, I needed a different name to stop mypy complaining about different types.

m-col:
> mypy was part of the problem.

You mean mypy identified a parent and child class using the same attribute for different things ;)

PR #3226, opened 2022-01-20
Release Notification service from dbus on finalize

### LINKED ISSUE #3223: Notify widget should un-register from dbus when finalized
Without doing so, reloading the config when a `Notify` widget was previously used but now is not, the dbus interface remains in a registered state. This prevents other processes from registering on that interface.

elParaguayo:
Reloading the config when a `Notify` widget was previously used but now is not, the dbus interface remains in a registered state. This prevents other processes from registering on that interface.

This PR adds a `finalize` method to the `Notify` widget which releases the name from dbus.

Fixes #3223

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -22,10 +22,11 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
+import asyncio
 from typing import Any
 
 try:
+    from dbus_next import ReleaseNameReply
     from dbus_next.aio import MessageBus
     from dbus_next.service import ServiceInterface, method, signal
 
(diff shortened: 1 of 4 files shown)


elParaguayo:
This is very weird. I wrote a short test to check the unregistering of the service. The test uses `@pytest.mark.asyncio` but these seems to cause random breakages in different tests that are run after that test.

elParaguayo:
> Perhaps the unregistering should mirror the logic in `register`: a 'client' of the notification service (e.g. the widget) registers a callback with the service. If the service hasn't started yet it is started. The tear-down logic would then be for the client to unregister its callback, then `NotificationManager` would remove it from `self.callbacks`, and only if that is empty stop the service.

Yeah. I skipped that as I didn't expect there to be multiple clients for the manager but I suppose there could be. If you had two separate widgets but only removed one, you wouldn't want to stop the service.

libqtile/notify.py
```diff
@@ -155,6 +156,35 @@ async def register(self, callback, capabilities=None, on_close=None):
...
 
+        def unregister(self, callback, on_close=None):
+            try:
+                self.callbacks.remove(callback)
+            except ValueError:
+                logger.error("Unable to remove notify callback. Unknown callback.")
+
+            if on_close:
+                try:
+                    self.close_callbacks.remoce(on_close)
```
m-col, line 167: Typo

m-col (commented):
Looks great, thanks for fixing this! There is a typo, but feel free to enable auto-merge when it's fixed!

libqtile/notify.py
elParaguayo: dagnammit.

PR #3229, opened 2022-01-20
Fix global_defaults for Battery widget

### LINKED ISSUE #3228: Battery widget doesn't respect extension_defaults
OS: Arch linux
Qtile version: 0.19.0

I have defined `extension_defaults` for all my widgets (see part of  my config file below). This seems to be respected by all the widgets I use except the Battery widget, which seems to respect the foreground color, but not the background color. Manually adding `background=bg` to the Battery widget works.

This line gives a black background (wrong) and the correct foreground color:
`widget.Battery(battery=0, format='{percent:2.0%}', low_percentage=0, show_short_text=False),`

This line gives the correct background color and the correct foreground color:
`widget.Battery(battery=0, format='{percent:2.0%}', low_percentage=0, show_short_text=False, background=bg),`

Part of my config file:
```
# colors
bg = "#282828" 
fg = "#ebdbb2"
gray = "#928374"
bright_gray = "#a89984"
red = "#cc241d"
bright_red = "#fb4934"
green = "#98971a"
bright_green = "#b8bb26"
yellow= "#d79921"
bright_yellow = "#fabd2f"
blue = "#458588"
bright_blue = "#83a598"
purple = "#b16286"
bright_purple = "#d3869b"
aqua = "#689d6a"
faded_aqua = "#8ec07c"

# theme defaults
widget_defaults = dict(
    font='RobotoMono Nerd Font',
    fontsize=16,
    padding=3,
    background=bg,
    foreground=fg
)
extension_defaults = widget_defaults.copy()
```

elParaguayo:
Similar to #3203, the Battery widget reads a property in `__init__` which prevents it being set by global defaults.

This PR fixes the issue by moving the property read to `_configure`.

Fixes #3228

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           ConfigError.
         - Release notification name from dbus when finalising `Notify` widget. This allows other notification
           managers to request the name.
+        - Fix bug where `Battery` widget did not retrieve `background` from `widget_defaults`.
 
 Qtile 0.19.0, released 2021-12-22:
     * features
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -339,9 +339,12 @@ class Battery(base.ThreadPoolText):
         self._battery = self._load_battery(**config)
         self._has_notified = False
 
+    def _configure(self, qtile, bar):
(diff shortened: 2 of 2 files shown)


m-col (approved):
LGTM

PR #3232, opened 2022-01-21
Add _wm_class to Icon in Systray

m-col:
87ac116 added this attribute to the X11 window classes, for use in the
`info()` command to report the wm_class of windows. This was left off
the `Icon` class in systray.py, leading to AttributeErrors when there
are icons.

### DIFF
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -1198,3 +1198,21 @@ def test_cmd_reload_config(manager_nospawn):
         assert manager_nospawn.c.eval("self.core.wmname") == (True, "LG3D")
     assert "dd" in manager_nospawn.c.groups()["S"]["windows"]  # First dropdown persists
     assert "dd" in manager_nospawn.c.groups()["1"]["windows"]  # Second orphans to group
+
+
+class CommandsConfig(Config):
+    screens = [
+        libqtile.config.Screen(
+            bottom=libqtile.bar.Bar([libqtile.widget.Systray()], 20),
+        )
+    ]
+
+
+@pytest.mark.parametrize("manager", [CommandsConfig], indirect=True)
+def test_windows_from_commands(manager):
+    manager.test_window("one")
+    assert len(manager.c.items("window")) == 2  # This command returns windows including bars
+    windows = manager.c.windows()  # Whereas this one is just regular windows
+    assert len(windows) == 1
+    # And the Systray is absent
+    assert "TestWindow" in windows[0]["wm_class"]
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -777,7 +777,11 @@ class Qtile(CommandObject):
             return False, [x.position for x in self.current_screen.gaps]
         elif name == "window":
             windows: list[str | int]
-            windows = [k for k, v in self.windows_map.items() if isinstance(v, CommandObject)]
+            windows = [
+                k
+                for k, v in self.windows_map.items()
+                if isinstance(v, CommandObject) and not isinstance(v, _Widget)
+            ]
             return True, windows
         elif name == "screen":
             return True, list(range(len(self.screens)))
@@ -806,7 +810,9 @@ class Qtile(CommandObject):
             else:
                 windows: dict[str | int, base._Window]
                 windows = {
-                    k: v for k, v in self.windows_map.items() if isinstance(v, CommandObject)
+                    k: v
+                    for k, v in self.windows_map.items()
+                    if isinstance(v, CommandObject) and not isinstance(v, _Widget)
                 }
                 return windows.get(sel)
         elif name == "screen":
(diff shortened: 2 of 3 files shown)


elParaguayo:
Why is this needed? The icons aren't `CommandObject`s so it shouldn't be possible to call `info()` on them.

elParaguayo:
Ah, did #3159 not fix this? That should have hidden icons from `windows()`?

elParaguayo:
OK. So the issue with my previous fix was that I didn't change these lines:
https://github.com/qtile/qtile/blob/a89939366245aafbaee6e22b84c4205f1e5a908a/libqtile/core/manager.py#L1228-L1230

My fix stops you selecting a window via IPC (which I think is right), but if we still want windows to be shown then your fix is necessary.

m-col:
I think your way is best - there is no reason to expose `Icon` at all via IPC. I wonder if it's really necessary to include it in `Qtile.windows_map`. Excluding it from there, and instead just tracking them on `Systray`, might be sufficient.

elParaguayo:
The tray and the icons are both added to `windows_map`. We should probably exclude both. 

The tray **is** a `CommandObject` as it inherits `base._Widget` but it's not a commandable window so that could also create issues if people loop over the map and expect everything to have window commands.

Do you want to update this PR to remove the `windows_map` parts and I'll do a separate PR for the bit I highlighted in `cmd_windows` which I think is still worth updating?

m-col:
> Do you want to update this PR to remove the windows_map parts and I'll do a separate PR for the bit I highlighted in cmd_windows which I think is still worth updating?

Yep sounds good!

m-col:
Hmm, maybe it isn't so simple. The X11 core event loop uses `window = self.qtile.windows_map.get(event.window)` to find the objects involved in a given event. Icons use those for the property notify and destroy notify events.

Another approach might be to filter out subclasses of `widget._Widget` from IPC calls that get windows, and add that base class to `Icon`.

elParaguayo:
Why don't we do this:
 1. keep the tray and icons in wind
 2. exclude non-commandable **and** anything that inherits `base._Widget`

I don't think we need icons to inherit `base._Widget`.

m-col:
Smart, sounds good

elParaguayo:
I'll just push as a separate commit to this PR if ok as it's all linked to the same issue.

m-col:
Yep!

elParaguayo:
Permission denied.

You can just add this diff yourself:
``` diff
index ea048b05..573435c8 100644
@@ -776,7 +776,11 @@ class Qtile(CommandObject):
             return False, [x.position for x in self.current_screen.gaps]
         elif name == "window":
             windows: List[Union[str, int]]
-            windows = [k for k, v in self.windows_map.items() if isinstance(v, CommandObject)]
+            windows = [
+                k
+                for k, v in self.windows_map.items()
+                if isinstance(v, CommandObject) and not isinstance(v, _Widget)
+            ]
             return True, windows
         elif name == "screen":
             return True, list(range(len(self.screens)))
@@ -805,7 +809,9 @@ class Qtile(CommandObject):
             else:
                 windows: Dict[Union[str, int], base._Window]
                 windows = {
-                    k: v for k, v in self.windows_map.items() if isinstance(v, CommandObject)
+                    k: v
+                    for k, v in self.windows_map.items()
+                    if isinstance(v, CommandObject) and not isinstance(v, _Widget)
                 }
                 return windows.get(sel)
         elif name == "screen":
@@ -1227,7 +1233,13 @@ class Qtile(CommandObject):

     def cmd_windows(self) -> List[Dict[str, Any]]:
         """Return info for each client window"""
-        return [i.info() for i in self.windows_map.values() if not isinstance(i, base.Internal)]
+        return [
+            i.info()
+            for i in self.windows_map.values()
+            if not isinstance(i, base.Internal)
+            and not isinstance(i, _Widget)
+            and isinstance(i, CommandObject)
+        ]
```

m-col:
Sure you tried the push correctly? 
image

`git apply` is telling me that that diff is corrupt unfortunately.

test/test_manager.py
```diff
@@ -1198,3 +1198,18 @@ def assert_dd_appeared():
...
+            bottom=libqtile.bar.Bar([libqtile.widget.Systray()], 20),
+        )
+    ]
+
+
+@pytest.mark.parametrize("manager", [CommandsConfig], indirect=True)
+def test_windows_from_commands(manager):
+    manager.test_window("one")
+    assert len(manager.c.items("window")) == 2
+    assert len(manager.c.windows()) == 1
```
elParaguayo, line 1215: Is this a bit confusing? The difference between getting 2 and 1 here is because the `Bar` is excluded from the second.

It might be good to assert that `Systray` is not in either of `items("window")` and `windows()`.

m-col:
Alright updated this one

test/test_manager.py
```diff
@@ -1198,3 +1198,21 @@ def assert_dd_appeared():
...
+
+
+@pytest.mark.parametrize("manager", [CommandsConfig], indirect=True)
+def test_windows_from_commands(manager):
+    manager.test_window("one")
+    assert len(manager.c.items("window")) == 2  # This command returns windows including bars
+    assert len(manager.c.windows()) == 1  # Whereas this one is just regular windows
+
+    # And the Systray is not present
+    assert "Systray" not in [c for w in manager.c.windows() for c in w["wm_class"]]
```
elParaguayo, line 1218: Is the `Systray`'s `wm_class` ever set?

If I run `qtile cmd-obj -o widget systray -f info`, `wm_class` is `None` this assertion isn't really confirming that `Systray` is not in that list.

m-col: Ah yeah, fixed

m-col:
Gah, I kept going back and forth with the tests and kept getting confused. It was because on Wayland the test window's wm_class is `["TestWindow"]` whereas on X it's `["TestWindow", "TestWindow"]`. Sigh. I've used an `in` not so that should be it sorted.

Issue #3196, opened 2022-01-14
Keychords behave incorrectly


alethiophile:
Currently, the behavior of the KeyChord mapping style is to, on receiving the keychord entry keystroke, unmap all other keys except those which are mapped in the keychord; then remain in this keychord mode until a key that is mapped in the keychord is pressed (or longer, if the keychord sets a mode), passing all unmapped keys through to the application. This behavior is inconsistent with all other software that uses keychord mappings, and is basically incorrect.

The correct behavior for a keychord (that does not set a mode) is to, on receiving the keychord opening keystroke, _unconditionally_ grab the next key; then, if that key is not mapped in the keychord, either display an error or do nothing, but in any case, revert immediately back to the root keymap. This is the behavior of all other software I can think of that maps keychords, including StumpWM, Emacs and screen/tmux. It's also fairly clearly the correct behavior: no one ever actually wants to first open a keychord, then press other keys that go through to the application, then press a keychord-mapped key to complete the keychord action. Functionally speaking, keychords are always pressed in direct succession, with no other keys between. Thus, opening a keychord and then pressing a key that is unmapped in that keychord can be presumed to be an error, and so should be called out to the user and should not leave the system in an unexpected mode (here, with the keychord still open, thus ignoring keys in the root keymap). (This all applies to keychords that do not set a mode; modal activation is rarer and has fewer clear precedents, so I don't know exactly what the behavior should be there.)

I intend to submit a pull request regarding this issue sometime in the next week or so.

m-col:
This seems reasonable, but the design of the keychords seems as though this was intential. I am 90% sure this specific behaviour has been discussed before, however unfortunately searching on github isn't helping me find the discussion. I'd like to hear the thoughts of others who were more involved with the keychord code, @ep12, perhaps @tych0, do you have thoughts on this?

alethiophile:
I see some prior discussion along these lines in https://github.com/qtile/qtile/pull/2121. Someone there had the objection that the behavior shouldn't be changed on compatibility grounds.

As far as I'm concerned, this is just a bug, and preserving a bug just because that's the existing behavior is not sound.

ep12:
I think you're talking about my comment there. That wasn't really a compatibility issue but rather my personal use case that would make use of the current behaviour (and the reason for me preferring an option). Actually, even with the suggested behaviour I could still use the default bindings of xournalpp by adding (a lot of) dummy bindings to the mode that just send the stroke through to the application.

[...]

PR #3235, opened 2022-01-22
Update chord functionality for consistent interface

alethiophile:
An initial pass at fixes to address the issue in #3196.

This is not ready to merge yet, for at least the following reasons:

1. This code only handles the x11 backend. I don't understand Wayland keyboard handling, nor have a Wayland setup to test on, so I can't write that immediately. Any advice on doing Wayland-backend development from within an x11 environment is welcome.
2. This patch has `grab_chord` in the manager deal directly with internals of the x11 backend core. I don't quite understand the purpose of `_valid_mask` in that core, so I don't know what arrangement would be best to provide the necessary information here.
3. This code currently treats modal chords the same as non-modal chords: i.e. any keypresses that are not mapped in the chord are silently consumed and do not pass through to the application. (However, unlike non-modal chords, modal chords still remain active until Esc is pressed.) I don't know if this is actually the best policy. Comments welcome.

Despite this, I figured I'd get the pull request open in order to solicit advice.

### DIFF
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -122,8 +122,8 @@ shrink it etc. as many times as needed. To exit the mode, press <escape>.
 Keyboard grabbing
 -----------------
 
-A KeyChord may specify the optional parameter `grab_keyboard`, which defaults to
-True:
+A KeyChord may specify the optional parameter ``grab_keyboard``, which defaults
+to True:
 
 ::
 
@@ -143,7 +143,7 @@ useful for modes, in order to allow you to interact with an application normally
 while also issuing commands from the mode.
 
 If it is set to True (the default), then any key pressed while a KeyChord is
-active will be consumed by QTile and will not be passed to the application.
+active will be consumed by Qtile and will not be passed to the application.
 Also, for chords which do not set a mode, any key (even one not bound in the
 chord) will cause the chord to deactivate and return to the main keymap.
 
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -89,12 +89,6 @@ the config file.
 
 The above code will launch xterm when the user presses Mod + z, followed by x.
 
-.. warning::
-    Users should note that key chords are aborted by pressing <escape>. In the
-    above example, if the user presses Mod + z, any following key presses will
-    still be sent to the currently focussed window. If <escape> has not been
-    pressed, the next press of x will launch xterm.
-
 Modes
 -----
 
@@ -124,6 +118,34 @@ shrink it etc. as many times as needed. To exit the mode, press <escape>.
     If using modes, users may also wish to use the Chord widget
     (:class:`libqtile.widget.chord.Chord`) as this will display the name of the
     currently active mode on the bar.
+
+Keyboard grabbing
+-----------------
+
+A KeyChord may specify the optional parameter `grab_keyboard`, which defaults to
+True:
+
+::
+
+    from libqtile.config import Key, KeyChord
+
+    keys = [
+        KeyChord([mod], "z", [
+            Key([], "g", lazy.spawn("xterm")),
+        ],
+        grab_keyboard=False,
+        mode="spawn")
+    ]
+
+If this is set to False, then keys not bound in the KeyChord will be passed
+through to the focused application and the chord will remain open. This can be
+useful for modes, in order to allow you to interact with an application normally
+while also issuing commands from the mode.
+
+If it is set to True (the default), then any key pressed while a KeyChord is
+active will be consumed by QTile and will not be passed to the application.
+Also, for chords which do not set a mode, any key (even one not bound in the
+chord) will cause the chord to deactivate and return to the main keymap.
 
 Chains
 ------
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -444,6 +444,18 @@ class Core(base.Core):
 
         return keysym, modmask
 
+    def grab_keyboard(self) -> None:
+        self.conn.conn.core.GrabKeyboard(
+            True,
+            self._root.wid,
+            xcffib.xproto.Time.CurrentTime,
+            xcffib.xproto.GrabMode.Async,
+            xcffib.xproto.GrabMode.Async,
+        )
+
+    def ungrab_keyboard(self) -> None:
+        self.conn.conn.core.UngrabKeyboard(xcffib.xproto.Time.CurrentTime)
+
     def grab_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
         """Map the key to receive events on it"""
         keysym, modmask = self.lookup_key(key)
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -95,6 +95,7 @@ class KeyChord:
         key: str,
         submappings: List[Union[Key, KeyChord]],
         mode: str = "",
+        grab_keyboard: bool = True,
     ):
         self.modifiers = modifiers
         self.key = key
@@ -102,6 +103,7 @@ class KeyChord:
         submappings.append(Key([], "Escape"))
         self.submappings = submappings
         self.mode = mode
+        self.grab_keyboard = grab_keyboard
 
     def __repr__(self):
         return "<KeyChord (%s, %s)>" % (self.modifiers, self.key)
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -87,6 +87,11 @@ class KeyChord:
         A string with vim like mode name. If it's set, the chord mode will
         not be left after a keystroke (except for Esc which always leaves the
         current chord/mode).
+    grab_keyboard:
+        A boolean (default True) controlling whether Qtile grabs exclusive
+        control of the keyboard while this keychord is active. If False, keys
+        not bound in the keychord will pass through to the focused application
+        and leave the chord active.
     """
 
     def __init__(
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -424,13 +424,10 @@ class Qtile(CommandObject):
             hook.fire("enter_chord", chord.mode)
 
         self.ungrab_keys()
-        self.core.grab_keyboard()
+        if chord.grab_keyboard:
+            self.core.grab_keyboard()
         for key in chord.submappings:
-            # TODO figure out how to do this without reaching into backend
-            # internals
-            keysym, mask_key = self.core.lookup_key(key)
-            mask_key &= self.core._valid_mask
-            self.keys_map[(keysym, mask_key)] = key
+            self.grab_key(key)
 
     def cmd_ungrab_chord(self) -> None:
         """Leave a chord mode"""
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -376,6 +376,8 @@ class Qtile(CommandObject):
             logger.debug(
                 "Ignoring unknown keysym: {keysym}, mask: {mask}".format(keysym=keysym, mask=mask)
             )
+            if self.chord_stack and self.chord_stack[-1].mode == "":
+                self.cmd_ungrab_chord()
             return
 
         if isinstance(key, KeyChord):
@@ -422,13 +424,19 @@ class Qtile(CommandObject):
             hook.fire("enter_chord", chord.mode)
 
         self.ungrab_keys()
+        self.core.grab_keyboard()
         for key in chord.submappings:
-            self.grab_key(key)
+            # TODO figure out how to do this without reaching into backend
+            # internals
+            keysym, mask_key = self.core.lookup_key(key)
+            mask_key &= self.core._valid_mask
+            self.keys_map[(keysym, mask_key)] = key
 
     def cmd_ungrab_chord(self) -> None:
         """Leave a chord mode"""
         hook.fire("leave_chord")
 
+        self.core.ungrab_keyboard()
         self.ungrab_keys()
         if not self.chord_stack:
             logger.debug("cmd_ungrab_chord was called when no chord mode was active")
@@ -449,6 +457,7 @@ class Qtile(CommandObject):
     def cmd_ungrab_all_chords(self) -> None:
         """Leave all chord modes and grab the root bindings"""
         hook.fire("leave_chord")
+        self.core.ungrab_keyboard()
         self.ungrab_keys()
         self.chord_stack.clear()
         for key in self.config.keys:
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -62,6 +62,14 @@ class Core(CommandObject, metaclass=ABCMeta):
     def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
         """Get the screen information"""
 
+    def grab_keyboard(self) -> None:
+        """Grab the entire keyboard, passing on all key events"""
+        raise NotImplementedError()
+
+    def ungrab_keyboard(self) -> None:
+        """Release the keyboard and process key events normally"""
+        raise NotImplementedError()
+
     @abstractmethod
     def grab_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
         """Configure the backend to grab the key event"""
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -110,6 +110,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.keyboards: List[keyboard.Keyboard] = []
         self.grabbed_keys: List[Tuple[int, int]] = []
         self.grabbed_buttons: List[Tuple[int, int]] = []
+        self.keyboard_grabbed: bool = False
         DataDeviceManager(self.display)
         self.live_dnd: Optional[wlrq.Dnd] = None
         DataControlManagerV1(self.display)
@@ -709,6 +710,12 @@ class Core(base.Core, wlrq.HasListeners):
     def get_screen_info(self) -> List[Tuple[int, int, int, int]]:
         """Get the screen information"""
         return [screen.get_geometry() for screen in self.outputs if screen.wlr_output.enabled]
+
+    def grab_keyboard(self) -> None:
+        self.keyboard_grabbed = True
+
+    def ungrab_keyboard(self) -> None:
+        self.keyboard_grabbed = False
 
     def grab_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]:
         """Configure the backend to grab the key event"""
--- a/libqtile/backend/wayland/keyboard.py
+++ b/libqtile/backend/wayland/keyboard.py
@@ -111,12 +111,15 @@ class Keyboard(HasListeners):
             keysyms = [xkb_keysym[0][i] for i in range(nsyms)]
             mods = self.keyboard.modifier
             for keysym in keysyms:
-                if (keysym, mods) in self.grabbed_keys:
+                if (keysym, mods) in self.grabbed_keys or self.core.keyboard_grabbed:
                     self.qtile.process_key_event(keysym, mods)
                     return
 
             if self.core.focused_internal:
                 self.core.focused_internal.process_key_press(keysym)
                 return
+
+        if self.core.keyboard_grabbed:
+            return
 
         self.seat.keyboard_notify_key(event)


libqtile/backend/x11/core.py
```diff
@@ -444,6 +444,20 @@ def lookup_key(self, key: Union[config.Key, config.KeyChord]) -> Tuple[int, int]
...
+            self._root.wid,
+            xcffib.xproto.Time.CurrentTime,
+            xcffib.xproto.GrabMode.Async,
+            xcffib.xproto.GrabMode.Async,
+        )
+
+    def ungrab_keyboard(self) -> None:
+        self.conn.conn.core.UngrabKeyboard(
+            xcffib.xproto.Time.CurrentTime
+        )
```
ep12, line 459: The `pep8` CI probably wants this as a one-liner.

libqtile/core/manager.py
```diff
@@ -422,13 +424,19 @@ def grab_chord(self, chord: KeyChord) -> None:
...
+            keysym, mask_key = self.core.lookup_key(key)
+            mask_key &= self.core._valid_mask
+            self.keys_map[(keysym, mask_key)] = key
 
     def cmd_ungrab_chord(self) -> None:
         """Leave a chord mode"""
         hook.fire("leave_chord")
 
+        self.core.ungrab_keyboard()
         self.ungrab_keys()
```
ep12, line 440: It would probably be best to add `ungrab_keyboard()` and `grab_keyboard()` to the `Qtile` manager class. Then `ungrab_keyboard` could call the `self.core` implementation and do a `self.keys_map.clear()`. That would be the only part of `self.ungrab_keys()` that is necessary. Thus the two lines would become just the one 
```suggestion
        self.ungrab_keyboard()
```

libqtile/core/manager.py
```diff
@@ -449,6 +457,7 @@ def cmd_ungrab_chord(self) -> None:
     def cmd_ungrab_all_chords(self) -> None:
         """Leave all chord modes and grab the root bindings"""
         hook.fire("leave_chord")
+        self.core.ungrab_keyboard()
         self.ungrab_keys()
```
ep12, line 461: Same thing here.

ep12:
1. wayland. I think you can run qtile on wayland within a qtile x11 session (although you should probably be careful to specify a different `--config`), see http://docs.qtile.org/en/stable/manual/install/index.html#wayland
2. The internal stuff should probably be in the backend core, not the manager (see my suggestion in the comments above). (And the wayland core doesn't seem to have `lookup_key`) Also, the `_valid_mask` just makes sure that things like `numlock` and other special modifiers don't end up in the mod mask. It is used e.g. in `x11.core.handle_KeyPress`, that calls the manager `process_key_event`: https://github.com/qtile/qtile/blob/a89939366245aafbaee6e22b84c4205f1e5a908a/libqtile/backend/x11/core.py#L603-L607
I am not sure that it is really needed in the case of this pr, although `core.lookup_key` doesn't `& self._valid_mask` it.
3. Hm. What about a (possibly per-chord) flag for modal chords? Not a suggestion but really a question.

m-col (commented):
> I don't quite understand the purpose of _valid_mask in that core, so I don't know what arrangement would be best to provide the necessary information here.

It's there to account for num lock, which acts as a modifier but we want to ignore it as a modifier. The call to `self._auto_modmasks()` in the X11 core lets grabbing and ungrabbing to grab/ungrab all combinations of the desired keys with and without num lock enabled on. I see this is actually something that is absent from the Wayland backend. This omission is unintentional... it should probably be there. My keyboard does not have a numlock, but if anybody has it then confirming presence/absence of an issue (such as keybindings not working when num lock is on) would be helpful.

re Wayland: the Core is always responsible for clients getting key events, whether we have bound them to keys or not, so 'grabbing the entire keyboard' would only require that we continue to process keybindings, while no longer passing other keys to clients. Something like this might do (untested):
```diff
diff --git a/libqtile/backend/wayland/core.py b/libqtile/backend/wayland/core.py
index 69f39e99..b7df92b4 100644
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -117,6 +117,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.keyboards: List[inputs.Keyboard] = []
         self.grabbed_keys: List[Tuple[int, int]] = []
         self.grabbed_buttons: List[Tuple[int, int]] = []
+        self.keyboard_grabbed: bool = False
         DataDeviceManager(self.display)
         self.live_dnd: Optional[wlrq.Dnd] = None
         DataControlManagerV1(self.display)
@@ -785,6 +786,14 @@ class Core(base.Core, wlrq.HasListeners):
         """Release the grabbed key events"""
         self.grabbed_keys.clear()
 
+    def grab_keyboard(self) -> None:
+        """Grab the whole keyboard, swallowing all key events."""
+        self.keyboard_grabbed = True
+
+    def ungrab_keyboard(self) -> None:
+        """Ungrab the whole keyboard, passing events to clients as usual."""
+        self.keyboard_grabbed = False
+
     def grab_button(self, mouse: config.Mouse) -> int:
         """Configure the backend to grab the mouse event"""
         keysym = wlrq.buttons[mouse.button_code]
diff --git a/libqtile/backend/wayland/keyboard.py b/libqtile/backend/wayland/keyboard.py
index a338387e..47aeb968 100644
--- a/libqtile/backend/wayland/keyboard.py
+++ b/libqtile/backend/wayland/keyboard.py
@@ -119,4 +119,7 @@ class Keyboard(HasListeners):
                 self.core.focused_internal.process_key_press(keysym)
                 return
 
+        if self.core.keyboard_grabbed:
+            return
+
         self.seat.keyboard_notify_key(event)
```

Note that as these new core methods are used by `Qtile` we should add them to the base `Core` class too, maybe not as abstract methods but raising `NotImplementedError` (then a backend that doesn't implement it can still otherwise work without chords).

libqtile/core/manager.py
```diff
@@ -422,13 +424,19 @@ def grab_chord(self, chord: KeyChord) -> None:
...
 
         self.ungrab_keys()
+        self.core.grab_keyboard()
         for key in chord.submappings:
-            self.grab_key(key)
+            # TODO figure out how to do this without reaching into backend
+            # internals
+            keysym, mask_key = self.core.lookup_key(key)
+            mask_key &= self.core._valid_mask
+            self.keys_map[(keysym, mask_key)] = key
```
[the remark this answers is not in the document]
alethiophile: I haven't tried it. I switched this out just on general principles, but I don't know what the interaction between GrabKey and GrabKeyboard actually is in X11 protocol.

If that works without problems, it would certainly make the code organization simpler. One potential issue is if the key grabs from GrabKey persist in the X11 state past sending UngrabKeyboard; in that case we would have to either ensure we always called UngrabKey at the same time, or else have problems after the chord was unmapped.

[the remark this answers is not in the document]
m-col: I imagine that every time `grab_keyboard` is needed we can safely ungrab and regrab all keys before and after.

alethiophile:
Regarding the _valid_mask issue, it currently seems like 1. `lookup_key()` is not a function in the core definition, only the x11 implementation; 2. `lookup_key()` is only currently called in contexts where its output is either &ed with `_valid_mask`, or |ed with the members of `_auto_modmasks`. Thus, I think it would probably work to make `lookup_key()` always return its result & `_valid_mask`, at which point it could be promoted to the base API and called from manager. (For this to work, analogous code would have to be added to the Wayland backend.)

Does this seem reasonable?

m-col:
Hmm, I'm not sure. All of that stuff is a

m-col:
Hmm, I'm not sure. All of that stuff is an implementation detail that currently only exists in the Wayland backend. The Core API should be only those methods that are necessary for a functioning backend, and we should avoid adding anything additional if we can. Let's look top-down from the problem we are trying to solve: we need to temporarily absorb all key events so that the chords can get them all. The minimum required additions would be a method to turn it on, and one to turn it off. If we can simply add `grab_keyboard` and `ungrab_keyboard` to `Core` then that would be ideal. If they also require ungrabbing and regrabbing all keys before and after, those are just one call of an existing method.

alethiophile:
Just pushed an attempt at Wayland support based on @m-col's suggestion. I don't have the Wayland test suite working, so I'm just relying on the CI runs to see if it works.

alethiophile:
> Hmm, I'm not sure. All of that stuff is an implementation detail that currently only exists in the Wayland backend. The Core API should be only those methods that are necessary for a functioning backend, and we should avoid adding anything additional if we can.

The thing is, the functionality in the x11 core that the manager depends on is not _just_ grabbing and ungrabbing the keys. It's also doing the lookup to translate the Key object into a keysym and modmask. Currently this function is only exposed through the `grab_key` and `ungrab_key` APIs. This function also does depend on the backend connection, at least under x11, since x11 calls are made in order to construct `_valid_mask`.

If we do require backend functionality for the key lookup, it seems better for that to be a separate function, rather than lumped into `grab_key`. This provides better orthogonality, as we can already see from this case where we need to look up keysyms but prefer not to issue the x11 GrabKey call. Alternatively, if we can provide lookups without relying on the backend, perhaps that functionality can be moved out into a separate utility function. But the current arrangement where this necessary lookup functionality is only available by calling a separate function with a side effect seems bad.

ep12:
> Regarding the _valid_mask issue, it currently seems like 1. `lookup_key()` is not a function in the core definition, only the x11 implementation; 2. `lookup_key()` is only currently called in contexts where its output is either &ed with `_valid_mask`, or |ed with the members of `_auto_modmasks`. Thus, I think it would probably work to make `lookup_key()` always return its result & `_valid_mask`, at which point it could be promoted to the base API and called from manager. (For this to work, analogous code would have to be added to the Wayland backend.)
> 
> Does this seem reasonable?

Actually, I think having a `lookup_key` in the abstract core might be reasonable. @m-col said that the wayland backend uses `xkbcommon` key symbols (thus the keysyms of both backends are compatible) which might not be the preferred system of possible future backends (I think a macOS backend is wanted but not worked on). See also https://github.com/qtile/qtile/pull/2785#issuecomment-950180583.
I'd be interested in hearing more opinions about that general topic.

libqtile/core/manager.py
alethiophile: In light of my last commit, I don't think this is true -- since `grab_chord` now grabs keys as well as grabbing the keyboard, I think calling `ungrab_keys` is necessary as well.

Of course, we could still create an `ungrab_keyboard` on the manager class and make it call `ungrab_keys` as well.

alethiophile:
I just pushed code to make this option configurable per-chord.

As mentioned in the issue, I don't see a use case for suspending grab_keyboard on non-modal chords. However, since the two follow the same code path here, supporting the option for modal chords also gives it to non-modal chords for free. 

It may be worth adding a new test case to explicitly check whether keys are correctly being passed through or not according to the grab_keyboard option. I'm not sure how you would test that programmatically, however. Here I've just quickly verified in a Xephyr run that it behaves as I expect in both cases.

This commit might also make the Wayland version work, since the manager code is now no longer calling `lookup_key`.

libqtile/core/manager.py
ep12: Yeah, I think technically it would be sufficient to populate the `keys_map` with the right values without additional single-key grabs in the backend. But then `Qtile.grab_key` populates the `keys_map` from the output of `core.grab_key`.
One could add a parameter to the `core.grab_key` to just return the dictionary key without doing the actual backend call, but I'm fine with the way you did it. It seems a bit unintuitive to grab the whole keyboard and some keys at first sight but then it's really not worth the extra hassle.
Can you mark this as resolved?

libqtile/core/manager.py
ep12: ...and this as resolved.

alethiophile:
Anyone have any comment on the grab_keyboard configuration? It seems this is the most straightforward way to do it, but user-visible changes should be considered carefully.

docs/manual/config/keys.rst
```diff
@@ -125,6 +119,34 @@ shrink it etc. as many times as needed. To exit the mode, press <escape>.
     (:class:`libqtile.widget.chord.Chord`) as this will display the name of the
     currently active mode on the bar.
 
+Keyboard grabbing
+-----------------
+
+A KeyChord may specify the optional parameter `grab_keyboard`, which defaults to
```
m-col, line 125: This needs double backticks

docs/manual/config/keys.rst
```diff
@@ -125,6 +119,34 @@ shrink it etc. as many times as needed. To exit the mode, press <escape>.
...
+        mode="spawn")
+    ]
+
+If this is set to False, then keys not bound in the KeyChord will be passed
+through to the focused application and the chord will remain open. This can be
+useful for modes, in order to allow you to interact with an application normally
+while also issuing commands from the mode.
+
+If it is set to True (the default), then any key pressed while a KeyChord is
+active will be consumed by QTile and will not be passed to the application.
```
m-col, line 146: We stylise the name as "Qtile"

libqtile/config.py
```diff
@@ -87,6 +87,11 @@ class KeyChord:
         A string with vim like mode name. If it's set, the chord mode will
         not be left after a keystroke (except for Esc which always leaves the
         current chord/mode).
+    grab_keyboard:
+        A boolean (default True) controlling whether Qtile grabs exclusive
```
m-col, line 91: I see you actually did write it as "Qtile" here :smile:

m-col (commented):
That all looks great, the docs are great, and your code is clean, so thanks!

I agree that we need test(s) for all combinations of backend/grab_keyboard/modal etc to ensure it all works as expected and stays that way. Unfortunately there isn't a handler (yet) for simulating keypresses in a way that also correctly tests grabbing. There is only `Core.cmd_simulate_keypress`, but the event is internal only (i.e. it isn't a proper X/xkbcommon event).

I think in the meantime adding a test just for X would be fine. If you look at `test/backend/x11/conftest.py` the class at the bottom has a `fake_click` method - we'd basically need an equivalent method that also uses XTest to generate a fake key press. This would appear to Qtile has a completely native key event. For testing, note that the test windows (used by most tests) close themselves when they receive a key event, which can be useful for testing whether a key press was received by them.

alethiophile:
I'm currently trying to write a fake_keypress test simulator, and it's being difficult.

My diff:

```diff
diff --git a/test/backend/x11/conftest.py b/test/backend/x11/conftest.py
index edca0c7d..8d42dd9d 100644
--- a/test/backend/x11/conftest.py
+++ b/test/backend/x11/conftest.py
@@ -9,7 +9,7 @@ import xcffib.xproto
 import xcffib.xtest
 
 from libqtile.backend.x11.core import Core
-from libqtile.backend.x11.xcbq import Connection
+from libqtile.backend.x11.xcbq import Connection, keysyms
 from test.helpers import (
     HEIGHT,
     SECOND_HEIGHT,
@@ -21,6 +21,7 @@ from test.helpers import (
     TestManager,
 )
 
+from typing import List
 
 @Retry(ignore_exceptions=(xcffib.ConnectionException,), return_on_fail=True)
 def can_connect_x11(disp=":0", *, ok=None):
@@ -170,6 +171,15 @@ def xmanager(request, xephyr):
         yield manager
 
 
+# these constants are defined in the X protocol definition for the XTest
+# extension (I don't know why they aren't in xcffib)
+class FAKE_EVENT_TYPE:
+    KeyPress = 2
+    KeyRelease = 3
+    ButtonPress = 4
+    ButtonRelease = 5
+    MotionNotify = 6
+
 class XBackend(Backend):
     name = "x11"
 
@@ -184,9 +194,72 @@ class XBackend(Backend):
         conn = Connection(self.env["DISPLAY"])
         root = conn.default_screen.root.wid
         xtest = conn.conn(xcffib.xtest.key)
-        xtest.FakeInput(6, 0, xcffib.xproto.Time.CurrentTime, root, x, y, 0)
-        xtest.FakeInput(4, 1, xcffib.xproto.Time.CurrentTime, root, 0, 0, 0)
-        xtest.FakeInput(5, 1, xcffib.xproto.Time.CurrentTime, root, 0, 0, 0)
+        xtest.FakeInput(
+            FAKE_EVENT_TYPE.MotionNotify, 0, xcffib.xproto.Time.CurrentTime,
+            root, x, y, 0
+        )
+        xtest.FakeInput(
+            FAKE_EVENT_TYPE.ButtonPress, 1, xcffib.xproto.Time.CurrentTime,
+            root, 0, 0, 0
+        )
+        xtest.FakeInput(
+            FAKE_EVENT_TYPE.ButtonRelease, 1, xcffib.xproto.Time.CurrentTime,
+            root, 0, 0, 0
+        )
+        conn.conn.flush()
+        self.manager.c.sync()
+        conn.finalize()
+
+    def fake_keypress(self, mods: List[str], key: str) -> None:
+        """Send the specified keypress, with modifiers as necessary."""
+        conn = Connection(self.env["DISPLAY"])
+        root = conn.default_screen.root.wid
+        xtest = conn.conn(xcffib.xtest.key)
+
+        mc = conn.conn.core.GetModifierMapping()
+        mr = mc.reply()
+        mod_list = ["shift", "lock", "control", "mod1", "mod2", "mod3",
+                    "mod4", "mod5"]
+        mod_map: Dict[str, int] = {}
+        nk = mr.keycodes_per_modifier
+        for n, i in enumerate(mod_list):
+            mod_map[i] = [i for i in mr.keycodes[n * nk:(n+1) * nk] if i != 0]
+
+        for i in mods:
+            if i not in mod_map or len(mod_map[i]) == 0:
+                raise ValueError(f"invalid or unmapped mod key '{i}'")
+
+        for i in mods:
+            xtest.FakeInput(
+                FAKE_EVENT_TYPE.KeyPress,
+                mod_map[i][0],
+                xcffib.xproto.Time.CurrentTime,
+                root, 0, 0, 0
+            )
+
+        keysym = keysyms[key]
+        keycode = conn.keysym_to_keycode(keysym)[0]
+        xtest.FakeInput(
+            FAKE_EVENT_TYPE.KeyPress,
+            keycode,
+            xcffib.xproto.Time.CurrentTime,
+            root, 0, 0, 0
+        )
+        xtest.FakeInput(
+            FAKE_EVENT_TYPE.KeyRelease,
+            keycode,
+            xcffib.xproto.Time.CurrentTime,
+            root, 0, 0, 0
+        )
+
+        for i in mods:
+            xtest.FakeInput(
+                FAKE_EVENT_TYPE.KeyRelease,
+                mod_map[i][0],
+                xcffib.xproto.Time.CurrentTime,
+                root, 0, 0, 0
+            )
+
         conn.conn.flush()
         self.manager.c.sync()
         conn.finalize()
diff --git a/test/test_manager.py b/test/test_manager.py
index 0bd1c1c4..091e3ff7 100644
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -226,6 +226,25 @@ def test_keypress(manager):
     manager.c.simulate_keypress(["control"], "j")
     assert manager.c.groups()["a"]["focus"] == "one"
 
+# This is mostly a test of the fake_keypress code in conftest
+@manager_config
+def test_fake_keypress(manager):
+    manager.c.group.setlayout("max")
+
+    manager.test_window("one")
+    manager.test_window("two")
+
+    assert manager.c.groups()["a"]["focus"] == "two"
+
+    manager.backend.fake_keypress(["control"], "k")
+
+    assert manager.c.groups()["a"]["focus"] == "one"
+
+    manager.backend.fake_keypress(["control"], "j")
+
+    assert manager.c.groups()["a"]["focus"] == "two"
 
 class TooFewGroupsConfig(ManagerConfig):
     groups = []
```

When I run this test, it usually fails on the second assertion, but intermittently passes. I don't understand what's going on -- I tried adding a sleep between the fake_keypress calls, in case of a race condition, but it didn't change anything.

The test above references a test that often fails because of some Xephyr keymap issue, but if it were that I would expect it to fail on the first assertion rather than the second.

jwijenbergh:
Maybe this is easier to implement if I finish #3113?

OlMon:
@m-col Is there anything that can be done, to include this into Qtile?
I asked in issue #3196 about the state of this PR and it seemed to be ready. Only some tests made problems.

I am just curious because in #3196 it seemed everyone agreed that this was something that should be added.

Thanks and Cheers!

alethiophile:
I've just refreshed my keychords branch to apply correctly onto current HEAD. As far as I'm concerned this is ready to apply; however, it still lacks automated tests, which most recently I had trouble getting working using xtest.

I could probably put together some tests using simulate_keypress pretty quickly; this could exercise the chord logic fine, but would not test whether the keypress events went to the underlying window or not.

PR #3239, opened 2022-01-23
Fix render issue for WidgetBox with bar borders

elParaguayo:
Widgets in the WidgetBox are not being offset vertically and so draw over the bar's border.

This PR adds the `offsety` attribute for widgets in the widget box.


m-col:
LGTM, thanks!

PR #3248, opened 2022-01-25
fix changelog date typo

### LINKED ISSUE #3247: Typo on https://github.com/qtile/qtile/blob/master/CHANGELOG
Currently wrong year:
Qtile 0.20.0, released 2021-01-24:

Should be:
Qtile 0.20.0, released 2022-01-24:

m-col:
Doesn't time fly?

Closes #3247


elParaguayo:
I think my window centering PR may have been caught up in the features of the latest release rather than starting a new section for the next release.

If that's the case, do you want to fix it here too?

PR #3261, opened 2022-01-27
Fix Systray crash when screens are reconfigured

### LINKED ISSUE #3260: Systray widget crash
2022-01-27 09:48:40,366 ERROR libqtile bar.py:_configure_widget():L324 Systray widget crashed during _configure with error: ConfigError('Only one Systray can be used.')
[config.txt](https://github.com/qtile/qtile/files/7948239/config.txt)

elParaguayo:
`cmd_reconfigure_screens` doesn't finalise widgets so `Systray` is configured twice which results in a `ConfigError` (which was designed to stop multiple instances of the widget).

Fixes #3260

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 Qtile x.x.x, released xxxx-xx-xx:
     * features
         - Add `lazy.window.center()` command to center a floating window on the screen.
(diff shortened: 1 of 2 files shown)


m-col:
LGTM

PR #3262, opened 2022-01-27
Fix more issues with Systray in Widgetbox

elParaguayo:
WidgetBox tries to reconfigure widgets when it is, itself, reconfigured.

There is a bug where widgets in the box weren't being marked as configured since we moved the setting of this flag to Bar._configure.

This is particularly problematic when Systray is in the box for two reasons:
1. If the box is open when the bar is reconfigured, the bar sees an unconfigured Systray and tries to configure it and so creates a config error.
2. If the box is closed, widgetbox still tries to reconfigure the Systray and so also triggers a config error.

This PR addresses both of these scenarios by:
1. Ensuring the configured flag is set by the widgetbox
2. Only configuring widgets in the box when the box is configured for the first time.
3. Adding tests for systray and widgetbox for `screens_reconfigured` events.

### DIFF
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -79,13 +79,16 @@ class WidgetBox(base._Widget):
         base._Widget._configure(self, qtile, bar)
 
         self.layout = self.drawer.textlayout(
-            self.text_closed,
+            self.text_open if self.box_is_open else self.text_closed,
             self.foreground,
             self.font,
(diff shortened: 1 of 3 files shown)


elParaguayo:
@m-col - do you want to quickly check that the tests look ok to you before I merge this.

The code does fix the issue.

m-col:
Looks great, thanks!

PR #3271, opened 2022-01-29
Fix hard-coded ratio in Monad layouts

### LINKED ISSUE #3270: MonadTall cmd_reset should respect the configured ratio
If I set the MonadTall master/other ratio in my config to something other than 0.5, say 0.6:
```python
layouts = [
    layout.MonadTall(ratio=0.6),
]
```
all is good, the master window takes up 0.6 of my screen. But if I mess around with widening/thinning the master and decide to reset it with `cmd_reset` (or a keybinding to `lazy.layout.reset`) it goes back to being 0.5 of my screen, not 0.6 like I have configured.

I believe this is happening because of this line in `layout.MonadTall.cmd_reset`:
```python
def cmd_reset(self, redraw=True):
    "Reset Layout."
    self.ratio = self._med_ratio # <- this is hard-coded to 0.5
    # ...
```
This can be easily fixed by setting the `self._med_ratio` to what the user has provided (in `__init__`). This is how I have fixed it in my personal configuration with some inheritance:
```python
class FixedMonadTall(layout.MonadTall):
    def __init__(self, **config):
        super().__init__(**config)
        self._med_ratio = self.ratio
```
The only other place I see `_med_ratio` used is `layout.MonadTall._maximize_main` and I believe the same behavior would be expected there - if the master is wider than I have configured the default to be (not "wider than 0.5") minimize it, and maximize otherwise.

The same issue happens with MonadWide, but the above change will fix that too because of inheritance.

Thank you for making this amazing WM, it's a joy to use.

piotr-machura:
The ratio after using `layout.MonadTall.cmd_reset` was hard-coded to
0.5. Now it uses the ratio provided in layout's configuration, or
in the optional keyword argument (set it to 0.5 for old behavior).

The `_maximize_main` threshold is also affected - master will be
minimized if it's larger than the ratio configured by the user, not the
hard-coded 0.5.

The changes also affect MonadWide because of inheritance.

Fixes #3270

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -47,7 +47,8 @@ class MonadTall(_SimpleLayoutBase):
     A main pane that contains a single window takes up a vertical portion of
     the screen_rect based on the ratio setting. This ratio can be adjusted with
     the ``cmd_grow_main`` and ``cmd_shrink_main`` or, while the main pane is in
-    focus, ``cmd_grow`` and ``cmd_shrink``.
+    focus, ``cmd_grow`` and ``cmd_shrink``. You may also set the ratio directly
+    with ``cmd_set_ratio``.
 
     ::
 
@@ -143,7 +144,6 @@ class MonadTall(_SimpleLayoutBase):
 
     _left = 0
     _right = 1
-    _med_ratio = 0.5
 
     defaults = [
         ("border_focus", "#ff0000", "Border colour(s) for the focused window."),
@@ -196,6 +196,7 @@ class MonadTall(_SimpleLayoutBase):
             self.single_margin = self.margin
         self.relative_sizes = []
         self.screen_rect = None
+        self.default_ratio = self.ratio
 
     @property
     def focused(self):
@@ -227,6 +228,12 @@ class MonadTall(_SimpleLayoutBase):
         self.do_normalize = True
         return self.clients.remove(client)
 
+    def cmd_set_ratio(self, ratio):
+        "Directly set the main pane ratio"
+        ratio = min(self.max_ratio, ratio)
+        self.ratio = max(self.min_ratio, ratio)
+        self.group.layout_all()
+
     def cmd_normalize(self, redraw=True):
         "Evenly distribute screen-space among secondary clients"
         n = len(self.clients) - 1  # exclude main client, 0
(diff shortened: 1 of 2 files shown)


libqtile/layout/xmonad.py
```diff
@@ -237,16 +237,16 @@ def cmd_normalize(self, redraw=True):
...
-        self.ratio = self._med_ratio
+        self.ratio = ratio or self.default_ratio
         if self.align == self._right:
             self.align = self._left
         self.cmd_normalize(redraw)
 
     def _maximize_main(self):
         "Toggle the main pane between min and max size"
-        if self.ratio <= self._med_ratio:
+        if self.ratio <= self.default_ratio:
```
elParaguayo, line 249: Wouldn't it make sense for this number just to be halfway between `max_ratio` and `min_ratio` and not linked to `default_ratio` at all?

libqtile/layout/xmonad.py
```diff
@@ -237,16 +237,16 @@ def cmd_normalize(self, redraw=True):
             self.group.layout_all()
         self.do_normalize = False
 
-    def cmd_reset(self, redraw=True):
+    def cmd_reset(self, ratio=None, redraw=True):
```
elParaguayo, line 240: Can you add a `cmd_set_ratio(ratio)` method which could call `self.cmd_reset(ratio=ratio)`.

elParaguayo, line 240: Can you add a `cmd_set_ratio(ratio)` method which could call `self.cmd_reset(ratio=ratio)`.

elParaguayo (changes requested):

libqtile/layout/xmonad.py
piotr-machura: I am not sure we should use `cmd_reset` for that, since it also normalizes secondary windows (which may be counter-intuitive). I can write a method that will just set the ratio without touching the secondaries though, should not be too hard.

libqtile/layout/xmonad.py
piotr-machura: I think this way makes sense, if I have a "fresh layout" and use `_maximize_main` I expect it to always maximize the master (because of the way it's named).

But if I have say `ratio = 0.6`, `max_ratio = 0.85` and `min_ratio = 0.15` in my config it would actually minimize since the half-way threshold is 0.5.

elParaguayo: Well, if you're going by the name then it should never set it to the min ratio...

I don't feel particularly strongly about it so if you think it's best as you've got it, I won't argue!

piotr-machura: Hmm you are right, I got cought up in the name and forgot it is a "toggle point" between what's considered "maximizable" and "minimizable". It does make sense to have `default_ratio` to be the toggle point, but it also makes sense to just have it in the middle.

I think I'll change it to the midpoint like you suggested, and the user can always inherit from the layout and modify the method (as it is a very short one).

elParaguayo:
This is looking good now. Any chance we can add some tests to check the functionality of your changes.

Issue #3273, opened 2022-01-29
[ Xwayland ] Spotify doesn't respect Qtiles layouts ?


justinesmithies:
Using the latest git version of Qtile if I run Spotify from the aur it opens in a small floating window as pictured. If I press mod+f to fullscreen it then do the same again it starts to respect the monadtall layout and fills a window correctly.
![2022-01-29_21-01-1643492899](https://user-images.githubusercontent.com/1571498/151678680-8139e5ee-ec52-48e8-b67a-5bd19db899e6.jpg)
![2022-01-29_21-01-1643492916](https://user-images.githubusercontent.com/1571498/151678682-901db8e0-a093-4b95-8bde-5c442b9f8154.jpg)

justinesmithies:
I'm not actually sure if it's an Xwayland issue or just a Wayland issue as I know Spotify runs native in Wayland but for some reason still requires Xwayland support to run ?

cr1ogen:
I've tried other applications that use xwayland and they don't respect qtile layouts either, for example emacs 
![20220129_19h54m13s_grim](https://user-images.githubusercontent.com/9537788/151680201-e1ff2773-726b-4fd7-98af-cd5d31f436cf.png)

m-col:
Hmm, very weird. Any chance they could be triggering float rules or something? Previously `wm_type` wasn't used by matches when the wayland backend was running but now that is available on xwayland windows. Otherwise my other idea is that the logic that determines whether or not they want to become static windows is wrong. Do these windows behave like regular floating windows (draggable, tilable), or are they completely static?

justinesmithies:
I'll look into wether they are triggering float rules in the morning. But I can confirm they are floating , draggable and you can time them too. I can time Spotify by doing the following mod+f which obvs full screens the app then mod+f again and it's tilled correctly.

cr1ogen:
In my case it doesn't have any float rules. And the window can be dragged and enlarged with the mouse perfectly

m-col:
OK thanks guys, I can reproduce and think I've understood what's going 
on. Will post a fix soon.

cr1ogen:
It's happening again that spotify doesn't respect qtile layouts. Could someone confirm please? Thanks

PR #3274, opened 2022-01-30
WL: XWayland window request_configure should just give them the geometry

m-col:
And not also float the window - this was based on the incorrect
assumption that they would be requesting a new geometry via this
request, rather than using it to be configured with the new geometry
we've just given them. I dont really understand this but it seems to
work.

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -1254,7 +1254,6 @@ class XWindow(Window):
             # Add event listeners
             self.add_listener(self.surface.surface.commit_event, self._on_commit)
             self.add_listener(self.surface.request_fullscreen_event, self._on_request_fullscreen)
-            self.add_listener(self.surface.request_configure_event, self._on_request_configure)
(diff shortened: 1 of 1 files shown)


m-col:
Fixes https://github.com/qtile/qtile/issues/3273

m-col:
Test failures are unrelated

elParaguayo:
Spotify still ignores the layout when I try this.

m-col:
If you disable that signal altogether, the request configure one, does it fix it?

elParaguayo:
> If you disable that signal altogether, the request configure one, does it fix it?

Yes.

m-col:
OK, I've just removed it completely. I don't think we need it.

cr1ogen:
This bug is happening to me again after this last update of the Git version. Could someone confirm it? Thanks

cr1ogen:
> Fixes #3273

This bug is happening to me again after this last update of the Git version. Could someone confirm it? Thanks

PR #3279, opened 2022-01-30
Modernise type annotations and update black

m-col:
5 commits:

- Turn `Optional[X]` into `X | None`
- Turn `Union[X, Y]` into `X | Y`
- Move all type check-time imports behind `typing.TYPE_CHECKING` guard
- Update black version
- Use built-ins for types e.g. `List` -> `list`, `Dict` -> `dict` now that they are supported directly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,10 @@ jobs:
           LIBDRM: 2.4.109
         strategy:
             matrix:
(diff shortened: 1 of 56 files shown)


elParaguayo:
Nice!

elParaguayo:
I assume this is based on my IRC comment. By "resubmit", I meant just do something that changes the last commit hash so all tests are re-run.

PR #3284, opened 2022-01-31
Fix typo in SysTray

elParaguayo:
4c98020e introduced a typo in Systray which results in icons being displayed incorrectly.

This wasn't caught by tests because we don't really test the Systray widget...

See https://www.reddit.com/r/qtile/comments/sh9dtc/my_system_tray_has_detached_and_is_at_the_top_of/ for more.

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -221,7 +221,7 @@ class Systray(window._Window, base._Widget):
                 self.tray_icons.sort(key=lambda icon: icon.name)
                 self.qtile.windows_map[wid] = icon
 
-            self.conn.conn.core.ChangeSaveset(SetMode.Insert, wid)
+            self.conn.conn.core.ChangeSaveSet(SetMode.Insert, wid)
             self.conn.conn.core.ReparentWindow(wid, parent.wid, 0, 0)
             self.conn.conn.flush()
 


m-col:
Ah, apologies. Thanks for the fix.

elParaguayo:
> Ah, apologies. Thanks for the fix.

No problem at all. I think this just goes to show that we do need a proper `Systray` test. To do that, we need some app that we can use to create an app in the tray for us. Not sure what the best option is there. May test a few things but open to suggestions too.

m-col:
I went to find this: https://gnome.pages.gitlab.gnome.org/gtk/gtk3/class.StatusIcon.html in case it was only a few more lines to the test window script, but it seems to be deprecated. Maybe we can check for 1 of 2-3 known programs that use the systray, like vlc or whatever, and if none of them are present just skip the test.

elParaguayo:
> I went to find this: https://gnome.pages.gitlab.gnome.org/gtk/gtk3/class.StatusIcon.html in case it was only a few more lines to the test window script, but it seems to be deprecated. Maybe we can check for 1 of 2-3 known programs that use the systray, like vlc or whatever, and if none of them are present just skip the test.

That's exactly what I was looking at too. The deprecation is annoying.

PR #3285, opened 2022-01-31
Wayland: Added support for output-power-management-v1

anriha:
This adds basic support for output-power-management protocol. It is dependant on flacjacket/pywlroots#73

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -60,7 +60,11 @@ from wlroots.wlr_types.server_decoration import (
     ServerDecorationManager,
     ServerDecorationManagerMode,
 )
-from wlroots.wlr_types.output_power_management_v1 import OutputPowerManagerV1, OutputPowerV1SetModeEvent, OutputPowerManagementV1Mode
+from wlroots.wlr_types.output_power_management_v1 import (
+    OutputPowerManagerV1,
+    OutputPowerV1SetModeEvent,
+    OutputPowerManagementV1Mode,
+)
 from wlroots.wlr_types.virtual_keyboard_v1 import VirtualKeyboardManagerV1, VirtualKeyboardV1
 from wlroots.wlr_types.xdg_shell import XdgShell, XdgSurface, XdgSurfaceRole
 from xkbcommon import xkb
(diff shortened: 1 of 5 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -155,6 +156,8 @@ def __init__(self):
         XdgOutputManagerV1(self.display, self.output_layout)
         ScreencopyManagerV1(self.display)
         GammaControlManagerV1(self.display)
+        self._output_power_manager = OutputPowerManagerV1(self.display)
+        self.add_listener(self._output_power_manager.set_mode_event, self._on_output_power_manager_set_mode)
```
m-col, line 160: It looks like we don't need to keep around a handle to the manager as an attribute here.
```suggestion
        output_power_manager = OutputPowerManagerV1(self.display)
        self.add_listener(output_power_manager.set_mode_event, self._on_output_power_manager_set_mode)
```

m-col (commented):
Just one comment but overall this looks great, thank you very much!

One or two of the lines look quite long, so the CI might complain. You can just run `black` on the code and it'll fix it up.

Please could you also add a line to the changelog?

m-col:
It looks like black has reformatted a couple of unrelated lines, which I suspect is because it didn't find the `pyproject.toml`. Could you re-run it - you might need to be in the root folder.

anriha:
Now it should be OK, is the line in CHANGELOG ok like this? https://github.com/anriha/qtile/blob/c8280a30a757ad74297ca692aecd15434edd53d2/CHANGELOG#L4

m-col:
Yep it all looks perfect, thanks! I'll wait until the PR on pywlroots is merged to start the CI otherwise it will just fail. Then if they are all good then we can merge!

anriha:
flacjacket/pywlroots#73 was merged this should now work

m-col:
Looks like the docs aren't building. I think you need to add `wlroots.wlr_types.output_power_management_v1` to `MOCK_MODULES` in `docs/conf.py`

m-col:
Perfect. It'll merge if/when the tests pass. Thanks again!

PR #3288, opened 2022-02-01
Add test for systray icons

elParaguayo:
Adds a new test which loads two apps and checks:
a) That icons are created and placed
b) That icons are ordered alphabetically

Posting as draft for now in case @m-col has any ideas for additional things to test (e.g. if icons' positions are relative to the bar then we should probably test that the icons' parent is actually the bar and not the root window).

### DIFF
--- a/test/widgets/test_systray.py
+++ b/test/widgets/test_systray.py
@@ -17,11 +17,14 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import shutil
+
 import pytest
 
 import libqtile.bar
 import libqtile.config
 from libqtile import widget
+from test.helpers import Retry
 
 
 def test_no_duplicates_multiple_instances(manager_nospawn, minimal_conf_noscreen, backend_name):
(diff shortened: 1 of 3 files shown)


test/widgets/test_systray.py
```diff
@@ -91,3 +94,47 @@ def test_systray_reconfigure_screens(manager_nospawn, minimal_conf_noscreen, bac
...
+    def wait_for_icons():
+        _, icons = manager_nospawn.c.widget["systray"].eval("len(self.tray_icons)")
+        assert int(icons) == 2
+
+    if backend_name == "wayland":
+        pytest.skip("Skipping test on Wayland.")
+
+    for prog in ("volumeicon", "vlc"):
+        if shutil.which(prog) is None:
+            pytest.skip(f"{prog} must be installed. Skipping test.")
```
m-col, line 112: I wonder if we can have some redundancy here - by checking that the test suite has access to, say, 2 of a possible 3 or 4 options, then we increase the number of machines that will test this locally.

m-col (approved):
Looks good overall. I think this will be pretty useful in the future and will hopefully catch the bugs we inevitably introduce! :laughing:

test/widgets/test_systray.py
elParaguayo: Yeah, that's true. I was focusing more on CI (hence the additions to `ci.yml`).

Ok - I'll add some more e.g. `nm-applet` is a likely candidate.

elParaguayo: Thinking about this some more, I'm not sure I like the idea of a test launching other processes. The most important thing is that this passes on CI. If it's skipped locally, I'm ok with that.

m-col: Yeah, I think that's fair. It is a bit weird.

PR #3289, opened 2022-02-02
Update instructions for installation on fedora

modestcap:
I tried to install qtile with the current  instructions but dnf didn't found it. Currently fedora project lists it as unmaintained in it's official website https://src.fedoraproject.org/rpms/qtile

### DIFF
--- a/docs/manual/install/fedora.rst
+++ b/docs/manual/install/fedora.rst
@@ -2,9 +2,4 @@
 Installing on Fedora
 ====================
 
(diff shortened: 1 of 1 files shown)


m-col:
Thank you for this!

elParaguayo:
NB there are efforts to get the package maintained again: https://bugzilla.redhat.com/show_bug.cgi?id=2025124

So we can merge this for now but we may need to update it again if/when that issue is resolved.

PR #3290, opened 2022-02-02
Wayland: Added support for idle protocol

anriha:
This adds support for KDE idle protocol as implemented in wlroots. ~~It is dependant on flacjacket/pywlroots#74~~ 
It seems to work fine but I am not sure what actions should be considered an activity. Currently only mouse activity and keyboard activity is considered, but it should probably consider things like playing videos or games. But I am not really sure how to implement that. This should also help with #2852

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -49,6 +49,7 @@ from wlroots.wlr_types import (
     xdg_decoration_v1,
 )
 from wlroots.wlr_types.cursor import WarpMode
+from wlroots.wlr_types.idle import Idle
 from wlroots.wlr_types.layer_shell_v1 import LayerShellV1, LayerShellV1Layer, LayerSurfaceV1
 from wlroots.wlr_types.output_management_v1 import (
     OutputConfigurationHeadV1,
@@ -164,6 +165,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.add_listener(
             output_power_manager.set_mode_event, self._on_output_power_manager_set_mode
         )
+        self.idle = Idle(self.display)
         PrimarySelectionV1DeviceManager(self.display)
         self._virtual_keyboard_manager_v1 = VirtualKeyboardManagerV1(self.display)
         self.add_listener(
@@ -339,6 +341,7 @@ class Core(base.Core, wlrq.HasListeners):
 
     def _on_cursor_button(self, _listener, event: pointer.PointerEventButton):
         assert self.qtile is not None
+        self.idle.notify_activity(self.seat)
         pressed = event.button_state == input_device.ButtonState.PRESSED
         if pressed:
             self._focus_by_click()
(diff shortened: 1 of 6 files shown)


m-col:
Thank you very much for this!

> It seems to work fine but I am not sure what actions should be considered an activity. Currently only mouse activity and keyboard activity is considered, but it should probably consider things like playing videos or games. But I am not really sure how to implement that.

I think those might be sufficient. If a video or some other fullscreen client is running and wants to suppress responses to activity (powering down outputs), they can use the idle-inhibit protocol: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_idle_inhibit_v1.h

anriha:
> I think those might be sufficient. If a video or some other fullscreen client is running and wants to suppress responses to activity (powering down outputs), they can use the idle-inhibit protocol: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_idle_inhibit_v1.h

Right, that should be the correct way to do this, I may look into it.

anriha:
Pywlroots merged this should now work, not really sure why the pipeline is failing.

libqtile/backend/wayland/core.py
```diff
@@ -164,6 +165,7 @@ def __init__(self):
         self.add_listener(
             output_power_manager.set_mode_event, self._on_output_power_manager_set_mode
         )
+        self.idle = Idle(self.display)
```
m-col, line 168:
```suggestion
        self._idle = Idle(self.display)
```

anriha: That wouldn't work as it is also called from https://github.com/qtile/qtile/blob/eafc306f12faf6013ecd32e1f6add5a1923f111a/libqtile/backend/wayland/keyboard.py#L118
should i make it into a property?

m-col: Oh sorry I misread this bit -- your first approach is best

m-col:
Thanks again for this, auto merge enabled!

m-col:
Key presses do not seem to be preventing the idle state. Have you noticed that?

anriha:
Hmm, interesting for me they do. Only reason I can think of is that function returns before it gets to this line. https://github.com/anriha/qtile/blob/eafc306f12faf6013ecd32e1f6add5a1923f111a/libqtile/backend/wayland/keyboard.py#L118

m-col:
This is happening for me even when a client window receives input, so I know it is at least getting there. I think it would be good to notify activity before that return statement actually, so that Qtile keybindings also count. I'll investigate why presses aren't working for me and see if I can find out anything.

m-col:
OK I figured it out, it was my fault. I did a bad rebase with https://github.com/qtile/qtile/pull/2548 and the notify activity wasn't in the right place for it. All good now.

PR #3291, opened 2022-02-03
Wayland: added basic support for idle inhibit protocol

anriha:
This adds support for idle-inhibit-unstable-v1 protocol. It is dependant on flacjacket/pywlroots#75 
It works, but only for applications that only have one surface e.g. mpv for others like Firefox it does not. 

The problem is that the new idle inhibitor event only has reference to the surface it is bound to and currently I can not find any way to associate it with subsurfaces. For that it would require somehing like every window having references all the pointers of subsurfaces but that does not seem like good solution.
The other way would be to ignore if window is mapped and only care about if there is idle inhibitor. This would work but it wouldn't work as described in the protocol.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -50,6 +50,7 @@ from wlroots.wlr_types import (
 )
 from wlroots.wlr_types.cursor import WarpMode
 from wlroots.wlr_types.idle import Idle
+from wlroots.wlr_types.idle_inhibit_v1 import IdleInhibitorManagerV1, IdleInhibitorV1
 from wlroots.wlr_types.layer_shell_v1 import LayerShellV1, LayerShellV1Layer, LayerSurfaceV1
 from wlroots.wlr_types.output_management_v1 import (
     OutputConfigurationHeadV1,
@@ -166,6 +167,8 @@ class Core(base.Core, wlrq.HasListeners):
             output_power_manager.set_mode_event, self._on_output_power_manager_set_mode
         )
         self.idle = Idle(self.display)
+        _idle_ihibitor_manager = IdleInhibitorManagerV1(self.display)
+        self.add_listener(_idle_ihibitor_manager.new_inhibitor_event, self._on_new_inhibitor)
         PrimarySelectionV1DeviceManager(self.display)
         self._virtual_keyboard_manager_v1 = VirtualKeyboardManagerV1(self.display)
         self.add_listener(
@@ -405,6 +408,19 @@ class Core(base.Core, wlrq.HasListeners):
 
     def _on_new_virtual_keyboard(self, _listener, virtual_keyboard: VirtualKeyboardV1):
         self._add_new_keyboard(virtual_keyboard.input_device)
+
+    def _on_new_inhibitor(self, _listener, idle_inhibitor: IdleInhibitorV1):
+        logger.debug("Signal: idle_inhibitor new_inhibitor")
+
+        for win in self.qtile.windows_map.values():
+            if isinstance(win, (window.Internal, window.Static)):
+                continue
+            assert isinstance(win, window.Window)
+
+            if win.surface.surface == idle_inhibitor.surface:
+                logger.error(str(win))
+                win.add_idle_inhibitor(idle_inhibitor)
+                break
 
     def _on_output_power_manager_set_mode(self, _listener, mode: OutputPowerV1SetModeEvent):
         logger.debug("Signal: output_power_manager set_mode_event")
@@ -773,6 +789,19 @@ class Core(base.Core, wlrq.HasListeners):
             )
         else:
             self.stacked_windows = self.mapped_windows
+
+    def check_idle_inhibitor(self) -> None:
+        """
+        Checks if any window that is currently mapped has idle inhibitor
+        and if so inhibits idle
+        """
+
+        idle_inhibited = any(
+            win.is_idle_inhibited
+            for win in self.mapped_windows
+            if isinstance(win, window.Window) and not isinstance(win, (window.Internal, window.Static))
+        )
+        self.idle.set_enabled(self.seat, not idle_inhibited)
 
     def get_screen_info(self) -> list[tuple[int, int, int, int]]:
         """Get the screen information"""
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -167,8 +167,8 @@ class Core(base.Core, wlrq.HasListeners):
             output_power_manager.set_mode_event, self._on_output_power_manager_set_mode
         )
         self.idle = Idle(self.display)
-        _idle_ihibitor_manager = IdleInhibitorManagerV1(self.display)
-        self.add_listener(_idle_ihibitor_manager.new_inhibitor_event, self._on_new_inhibitor)
+        idle_ihibitor_manager = IdleInhibitorManagerV1(self.display)
+        self.add_listener(idle_ihibitor_manager.new_inhibitor_event, self._on_new_inhibitor)
         PrimarySelectionV1DeviceManager(self.display)
         self._virtual_keyboard_manager_v1 = VirtualKeyboardManagerV1(self.display)
         self.add_listener(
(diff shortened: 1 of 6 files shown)


libqtile/backend/wayland/window.py
```diff
@@ -290,6 +295,14 @@ def _on_foreign_request_close(self, _listener, _event):
         logger.debug("Signal: foreign_toplevel_management request_close")
         self.kill()
 
+    def _on_inhibitor_destroy(self, _listener, surface: surface.Surface):
+        # We don't have reference to the inhibitor, but it doesn't really
+        # matter we only need to keep count of how many inhibitors there are
+        self._idle_inhibitors.pop()
+        _listener.remove()
```
m-col, line 302: Could you rename `_listener` to `listener` and add a type annotation to it? The reasoning for "_listener" was because it is usually an unused variable that we want to ignore, whereas he it's being used.

libqtile/backend/wayland/core.py
```diff
@@ -406,6 +409,19 @@ def _on_new_pointer_constraint(self, _listener, wlr_constraint: PointerConstrain
...
+
+        for win in self.qtile.windows_map.values():
+            if isinstance(win, (window.Internal, window.Static)):
+                continue
+            assert isinstance(win, window.Window)
+
+            if win.surface.surface == idle_inhibitor.surface:
+                logger.error(str(win))
+                win.add_idle_inhibitor(idle_inhibitor)
+                break
```
m-col, line 423: You can iterate through all of a windows surfaces with a callback like this:

```suggestion
        for win in self.qtile.windows_map.values():
            if isinstance(win, window.Window):
		win.surface.for_each_surface(win.add_idle_inhibitor, idle_inhibitor)
		if idle_inhibitor.data:
		    break
```
In this approach, you also need to adjust `Window.add_idle_inhibitor`:
```py
    def add_idle_inhibitor(self, surface: surface.Surface, _x, _y, inhibitor: IdleInhibitorV1) -> None:
        if surface == inhibitor.surface:
            inhibitor.data = self
            self._idle_inhibitors.append(inhibitor)
            self.add_listener(inhibitor.destroy_event, self._on_inhibitor_destroy)
            if len(self._idle_inhibitors) == 1:
                self.core.check_idle_inhibitor()
```
Then we are using `inhibitor.data` to know if the corresponding surface and window has been found.

libqtile/backend/wayland/window.py
```diff
@@ -77,6 +79,7 @@ def __init__(self, core: Core, qtile: Qtile, surface: SurfaceType):
         self.surface = surface
         self._group: _Group | None = None
         self.popups: list[XdgPopupWindow] = []
+        self._idle_inhibitors: list[IdleInhibitorV1] = []
```
m-col, line 82: If we don't need the objects themselves, maybe we can get away with just using an `int` as a counter, and increment/decrement it?

libqtile/backend/wayland/window.py
```diff
@@ -161,6 +164,8 @@ def mapped(self, mapped: bool) -> None:
         else:
             self.core.mapped_windows.remove(self)
         self.core.stack_windows()
+        if hasattr(self, "_idle_inhibitors") and len(self._idle_inhibitors) > 0:
+            self.core.check_idle_inhibitor()
```
m-col, line 168: Is this because of another class inheriting this method, which might not have this attribute? If so, this hack will fortunately not be necessary soon by having more well-defined types: https://github.com/qtile/qtile/pull/3293

libqtile/backend/wayland/core.py
```diff
@@ -166,6 +167,8 @@ def __init__(self):
             output_power_manager.set_mode_event, self._on_output_power_manager_set_mode
         )
         self.idle = Idle(self.display)
+        _idle_ihibitor_manager = IdleInhibitorManagerV1(self.display)
+        self.add_listener(_idle_ihibitor_manager.new_inhibitor_event, self._on_new_inhibitor)
```
m-col, line 171: ` _idle_ihibitor_manager` -> ` idle_ihibitor_manager` just to be consistent with other local-only variables

libqtile/backend/wayland/core.py
```diff
@@ -774,6 +790,19 @@ def stack_windows(self) -> None:
...
+        Checks if any window that is currently mapped has idle inhibitor
+        and if so inhibits idle
+        """
+
+        idle_inhibited = any(
+            win.is_idle_inhibited
+            for win in self.mapped_windows
+            if isinstance(win, window.Window) and not isinstance(win, (window.Internal, window.Static))
+        )
+        self.idle.set_enabled(self.seat, not idle_inhibited)
```
m-col, line 804: I think the generator for loop hurts readability:
```suggestion
        for win in self.mapped_windows:
	    if isinstance(win, window.Window) and win.is_idle_inhibited:
		self.idle.set_enabled(self.seat, True)
		return

	self.idle.set_enabled(self.seat, False)
```

m-col (changes requested):
Thanks again for working on improving the Wayland backend. Overall this is looking great - I've left some review comments, which I think will also address your comment about not acccounting for subsurfaces.

libqtile/backend/wayland/core.py
anriha: Nice, that works!

libqtile/backend/wayland/window.py
anriha: Yep that is probably easier, commited.

libqtile/backend/wayland/window.py
anriha: Hmm that would solve this, it is present on several places in this PR, should I wait until your change is merged, or should I leave it as is and change it later?

m-col: I'm happy to merge this first and rebase that other PR on top of this. However, we do need to account for the different window classes, specifically we need to ensure that if `Window.cmd_static` is called to turn an XDG shell window into a `Static` window, that window's `_idle_inhibitors` also needs to be copied to the new object. Similarly we need to register idle inhibitors that are created for windows represented by `Static` and handle those just the same if the window is part of the layer shell or xdg shell. If it makes things easier to see how these things are represnted then waiting until that other PR is merged might make sense. Up to you.

anriha: Ok, I've added idle_inhibitors_count to other window classes, and to cmd_static. It seems to work fine.

anriha:
Pywlroots was merged so this should now work.

libqtile/backend/wayland/window.py
```diff
@@ -372,6 +385,22 @@ def paint_borders(self, color: ColorsType, width) -> None:
                 self.bordercolor = [_rgb(color)]
         self.borderwidth = width
 
+    def add_idle_inhibitor(
+        self, surface: surface.Surface, _x: int, _y: int, inhibitor: IdleInhibitorV1 | None
```
m-col, line 389: How come the type of inhibitor is `IdleInhibitorV1 | None` and not just `IdleInhibitorV1`?

libqtile/backend/wayland/core.py
```diff
@@ -774,6 +789,18 @@ def stack_windows(self) -> None:
         else:
             self.stacked_windows = self.mapped_windows
 
+    def check_idle_inhibitor(self) -> None:
+        """
+        Checks if any window that is currently mapped has idle inhibitor
+        and if so inhibits idle
+        """
+        for win in self.mapped_windows:
+            if isinstance(win, window.Window) and win.is_idle_inhibited:
```
m-col, line 798: The type of `self.mapped_windows` is `window.WindowType` so I think this `isinstance` test is redundant.

libqtile/backend/wayland/window.py
anriha: Because the type signature for all implementations of for_each_surface is `data: T | None` so without that mypy complains.

m-col: OK cool, thanks

libqtile/backend/wayland/core.py
anriha: You are right that is redundant, fixed.

m-col (approved):

m-col:
Thanks again for this! Auto merge enabled.

PR #3293, opened 2022-02-03
WL: Typing of wayland backend

m-col:
As more shells were being supported by the Wayland backend, I had
optimistically tried to keep their respective representations together
in one `Window` class and one `Static` class. This has become
increasingly spaghetti-like and prone to introduction of bugs and
confusion.

This change uses a generic base with a dedicated class per shell such
that we have:

- XdgWindow: regular windows of the XDG shell
- XWindow: regular X11 clients connecting via XWayland
- XdgStatic: static windows of th XDG shell, made static by the user
- XStatic: static X11 clients, made static by the user
- LayerStatic: static layer-shell windows that are always static

This also makes it more clear which methods were being passed around by
inheritance by splitting those common to all classes into a new `_Base`.

It also removes a whole bunch of the assert statements and 'type:
ignore' comments that were piling up just to appease mypy.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,7 @@ MOCK_MODULES = [
     'psutil',
     'pywayland',
     'pywayland.protocol.wayland',
+    'pywayland.protocol.wayland.wl_output',
     'pywayland.server',
     'wlroots',
     'wlroots.helper',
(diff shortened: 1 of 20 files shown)


m-col:
I increased the scope of the PR a bit and fully typed the entire backend, and added enforcement of typing of the backend to the mypy tox env so it'll stay that way. In the process I found a bunch of potential bugs that are now fixed too which is nice.

elParaguayo (approved):
LGTM. 

Typing looks clear and no issues starting up wayland backend.

PR #3304, opened 2022-02-05
x11: Only relayout upon hint update when increments update when floating

### LINKED ISSUE #3292: Reaper freezes Qtile
Creating a thread here so I can track issues/progress.

Issue seems to be that reaper freezes qtile. When I tried it this morning mouse clicks didn't work and qtile keybindings didn't respond initially but it seemed like they were queued as they were suddenly all executed in a burst.

Needs some investigating.

Copying @JustineSmithies for any additional info...

m-col:
When `WM_NORMAL_HINTS` are updated we are uncondtionally re-laying out
the group. The client 'reaper' is getting into a never ending loop
because when it is placed it updates its hints. The `group.layout_all()`
call was added in 07e658a so that updates to width and height increment
hints are respect. This change removes the call except for specifically
when these hints are updated while the window is floating, as we do not
respect these hints in tiled windows anyway.

Fixes #3292

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -619,6 +619,9 @@ class _Window:
         except (xcffib.xproto.WindowError, xcffib.xproto.AccessError):
             return
 
+        width_inc = self.hints["width_inc"]
+        height_inc = self.hints["height_inc"]
+
         if normh:
             self.hints.update(normh)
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Awesome work.

Tested and it kills the event loop for reaper as you say. Great summary of the bug too.

LGTM.

elParaguayo:
Although ...
```
[16:03] < JustineSmi> Hmm just noticed if I try run Reaper under Wayland it says it's running but no window ??
[16:13] < JustineSmi> mcol: logfiles from trying to run reaper with just Wayland but get no window
                      https://paste.sr.ht/~justinesmithies/673719a7275f7e909ad77cb86089f8d631a5bf6d
```

PR #3310, opened 2022-02-07
[Docs] Fix layouts page in docs

elParaguayo:
(Reason for this is the new `MonadThreeCol` layout is not showing up in the docs)

The layouts reference page does not automatically update for new layouts.

This PR fixes the issue by displaying information for all layouts that subclass `libqtile.layout.base.Layout`, rather than adding layouts individually.


m-col:
Very nice!

Issue #3308, opened 2022-02-06
Ghost window with KDE Connect 


MithicSpirit:
How to reproduce:
1. Install/start/configure KDE Connect
2. Open "Slideshow remote" (on phone)
3. Enable the pointer (tap & hold)
4. Besides the overlay window used to draw the pointer (which is floating), there appears to be another "ghost" window allocated in the layout; disabling the pointer makes it go away
5. Enabling the pointer again, move the overlay window (super+drag) so it remains floating but is no longer maximized; the ghost window remains
6. Disabling the pointer now makes ghost window stay there despite overlay window being gone
7. Restarting qtile (e.g. `qtile cmd-obj -o cmd -f restart`) makes the ghost window go away again

Note: this seems to happen on all layouts but it is easiest to observe in the `monadtall` layout with one other window open.

The expected behavior is that the ghost window never exist in the first place (step 4), although it is also unexpected that it should linger behind after disabling the pointer (step 6).

Additional info:
- `qtile --version`: `0.20.0`
- `python --version`: `Python 3.10.2`
- OS: Arch Linux
- Window system: X.org (with `picom` as compositor)

m-col:
If you `xprop` on the ghost window does it give you information about anything other than the root?

Could you get the window ID from `qtile cmd-obj -o cmd -f windows` and then get its info from `qtile cmd-obj -o window $WID if info`, before and after you disable the pointer?

Does `wmctrl -l` give you any information about it?

MithicSpirit:
`xprop` just gives info about the root window when taken on the ghost window, but it properly shows the information of the overlay (icons omitted as no meaningful information could be shown): 
```
_VARIABLE_REFRESH(CARDINAL) = 1
_NET_FRAME_EXTENTS(CARDINAL) = 0, 0, 0, 0
_NET_WM_DESKTOP(CARDINAL) = 5
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_USER_TIME(CARDINAL) = 3809284
_NET_WM_STATE(ATOM) = _NET_WM_STATE_ABOVE, _NET_WM_STATE_FULLSCREEN, _NET_WM_STATE_STAYS_ON_TOP
_NET_WM_ICON(CARDINAL) =
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) =
_NET_WM_NAME(UTF8_STRING) = "KDE Connect Daemon"
[...]

PR #3324, opened 2022-02-10
Fullscreen: Fix issues with floating windows

### LINKED ISSUE #3323: Moving a fullscreened window to a different group makes it disappear from all groups
When I use `lazy.window.togroup(group_name)` to send a window to another group and the window is fullscreened, it does not appear on the target group.

Some logs:
```
2022-02-10 22:33:15,589 ERROR libqtile manager.py:process_key_event():L391 KB command error togroup: Traceback (most recent call last):
  File "/home/mcol/git/qtile/libqtile/command/interface.py", line 312, in call
  File "/home/mcol/git/qtile/libqtile/backend/base.py", line 422, in cmd_togroup
    self.togroup(group_name, switch_group=switch_group)
  File "/home/mcol/git/qtile/libqtile/backend/wayland/window.py", line 370, in togroup
    self.group.remove(self)
  File "/home/mcol/git/qtile/libqtile/group.py", line 295, in remove
    self.tiled_windows.remove(win)
KeyError: Window(name='imv - [1/1] [1920x1080] [100%] /home/mcol/20220210_22h32m45s_grim.png [shrink to fit] ', wid=13)
```

This suggests that, as the window is removed from the current group but then the action errors out, the fullscreened window just remains in an unmapped state.

I think this is a relatively new bug. Pinging @jwijenbergh - I suspect it's to do with your recent fullscreen changes.

jwijenbergh:
Fullscreen windows still had some rough corners.

- Explicitly check for windows to remove in the tiled window
set. Sometimes windows were not guaranteed to be added, e.g. floating
windows newly added to a group. Fixes https://github.com/qtile/qtile/issues/3323. This makes a lot of sense and was suggested by @m-col
- Make sure that the _NET_WM_STATE is correctly updated when floating
a fullscreen windows. Fixes an issue in #3308
- Make sure that windows are only added to the floating layout if they
are not fullscreen. Fixes an issue in #3308

Still needs some testing on the wayland backend. I have not updated my wayland dependencies just yet so I will get to that soon. In the meantime lots of manual testing needed as I want to make sure we iron out these issues once and for all

Some thoughts:

The Qtile code for fullscreening and floating is weird to work with. IMO we should look into how we can completely separate the floating code and fullscreen code. Fullscreen windows should not be floating windows but should be a completely different state. This is something for a future PR though

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1330,6 +1330,11 @@ class Window(_Window, base.Window):
             self.window.set_property("_NET_WM_STATE", list(new_state))
 
     def update_fullscreen_wm_state(self, do_full):
+        # already done updating previously
+        if do_full == self.fullscreen:
+            return
+
+        # update fullscreen _NET_WM_STATE
         atom = set([self.qtile.core.conn.atoms["_NET_WM_STATE_FULLSCREEN"]])
         prev_state = set(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
 
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1305,6 +1305,7 @@ class Window(_Window, base.Window):
                 # if we are setting floating early, e.g. from a hook, we don't have a screen yet
                 self._float_state = FloatStates.FLOATING
         elif (not do_float) and self._float_state != FloatStates.NOT_FLOATING:
+            self.update_fullscreen_wm_state(False)
             if self._float_state == FloatStates.FLOATING:
                 # store last size
                 self._float_width = self.width
@@ -1324,19 +1325,25 @@ class Window(_Window, base.Window):
     def toggle_floating(self):
         self.floating = not self.floating
 
+    def set_wm_state(self, old_state, new_state):
+        if new_state != old_state:
+            self.window.set_property("_NET_WM_STATE", list(new_state))
+
+    def update_fullscreen_wm_state(self, do_full):
+        atom = set([self.qtile.core.conn.atoms["_NET_WM_STATE_FULLSCREEN"]])
+        prev_state = set(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
+
+        if do_full:
+            self.set_wm_state(prev_state, prev_state | atom)
+        else:
+            self.set_wm_state(prev_state, prev_state - atom)
+
     @property
     def fullscreen(self):
         return self._float_state == FloatStates.FULLSCREEN
 
     @fullscreen.setter
     def fullscreen(self, do_full):
-        atom = set([self.qtile.core.conn.atoms["_NET_WM_STATE_FULLSCREEN"]])
-        prev_state = set(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
-
-        def set_state(old_state, new_state):
-            if new_state != old_state:
-                self.window.set_property("_NET_WM_STATE", list(new_state))
-
         if do_full:
             screen = self.group.screen or self.qtile.find_closest_screen(self.x, self.y)
 
@@ -1348,13 +1355,9 @@ class Window(_Window, base.Window):
                 screen.height - 2 * bw,
                 new_float_state=FloatStates.FULLSCREEN,
             )
-            set_state(prev_state, prev_state | atom)
             return
 
         if self._float_state == FloatStates.FULLSCREEN:
-            # The order of calling set_state() and then
-            # setting self.floating = False is important
-            set_state(prev_state, prev_state - atom)
             self.floating = False
             return
 
@@ -1467,6 +1470,7 @@ class Window(_Window, base.Window):
         return (self.x, self.y)
 
     def _reconfigure_floating(self, new_float_state=FloatStates.FLOATING):
+        self.update_fullscreen_wm_state(new_float_state == FloatStates.FULLSCREEN)
         if new_float_state == FloatStates.MINIMIZED:
             self.state = IconicState
             self.hide()
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Qtile x.x.x, released xxxx-xx-xx:
     * bugfixes
         - Fix `Systray` crash on `reconfigure_screens`.
         - Fix bug where widgets can't be mirrored in same bar.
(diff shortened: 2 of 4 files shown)


m-col:
Amazing as always, thank you for the fix (and so quick)!

I agree with all that you said. I'll append this to my local branch so I can run it over the next few days, and I'll update if there are any problems (or if it's fixed everything AFAICT).

libqtile/backend/x11/window.py
```diff
@@ -1429,6 +1429,10 @@ def cmd_static(
         hook.fire("client_managed", s)
 
     def tweak_float(self, x=None, y=None, dx=0, dy=0, w=None, h=None, dw=0, dh=0):
+        # Floating is tweaked so the window
+        # cannot be fullscreen
+        if self.fullscreen:
```
jwijenbergh, line 1434: I'm not sure if this is needed but I wanted to reduce the fullscreen setter calls.

jwijenbergh:
> Amazing as always, thank you for the fix (and so quick)!
> 
> I agree with all that you said. I'll append this to my local branch so I can run it over the next few days, and I'll update if there are any problems (or if it's fixed everything AFAICT).

No problem! I will do the same, it's a bit hard to find every corner case

MithicSpirit:
Can confirm this closes #3308

m-col:
I've noticed a couple of small issue with this: when I send a fullscreen window to another group and that window matches something in `float_rules`, then it is no longer fullscreen on the new group, and is instead floating. Other windows are correctly fullscreen on the target group, however if the target group already fullscreen window, it remains fullscreen and should instead un-fullscreen.

jwijenbergh:
> I've noticed a couple of small issue with this: when I send a fullscreen window to another group and that window matches something in `float_rules`, then it is no longer fullscreen on the new group, and is instead floating. 

Im trying to reproduce this with

```python
    float_rules=[
        Match(wm_class="st"),
    ]
```

Note that I added st here as that is my terminal so it's easy to test.

I spawn st, it is floating. set to fullscreen, move to another group and it is still fullscreen for me.

> however if the target group already fullscreen window, it remains fullscreen and should instead un-fullscreen.

I'm not sure I understand, could you elaborate a bit more?

jwijenbergh:
> Can confirm this closes #3308

Thanks for testing! good to hear

m-col:
>     however if the target group already fullscreen window, it remains fullscreen and should instead un-fullscreen.
> 
> I'm not sure I understand, could you elaborate a bit more?

Try this:

1. Open a single tiled window on group 1 and fullscreen it
2. Go to group 2, open a single tiled window and fullscreen it
3. Send the current window to group 1
4. Drag the current window (which was the second created window and is currently fullscreen) to the side
5. The first window appears fullscreened underneath
6. Drag the first window out of the fullscreen position
7. Then kill both windows
8. The tiled layout (on the current group, group 1) continues to reserve space for 2 windows which are never coming back

(will follow up with more information on the first issue involving `float_rules`).

jwijenbergh:
> > ```
> > however if the target group already fullscreen window, it remains fullscreen and should instead un-fullscreen.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure I understand, could you elaborate a bit more?
> 
> Try this:
> 
>     1. Open a single tiled window on group 1 and fullscreen it
> 
>     2. Go to group 2, open a single tiled window and fullscreen it
> 
>     3. Send the current window to group 1
> 
>     4. Drag the current window (which was the second created window and is currently fullscreen) to the side
> 
>     5. The first window appears fullscreened underneath
> 
>     6. Drag the first window out of the fullscreen position
> 
>     7. Then kill both windows
> 
>     8. The tiled layout (on the current group, group 1) continues to reserve space for 2 windows which are never coming back
> 
> 
> (will follow up with more information on the first issue involving `float_rules`).

Thanks for this anyways, it's handy for me to test

jwijenbergh:
I have removed the weird test that sometimes failed. I still cannot figure out why it is sometimes failing. I'm confident that it has something to do with the testing code and not the code itself as I'm not using any random or anything.

elParaguayo:
Getting a lot of failures recently from that test so I've cherry-picked it into a separate PR so we can merge that commit now.

m-col:
I'm finding that floating windows, when made to fullscreen and then un-fullscreen, become tiled. This is also the case on master so it doesn't have to be part of this PR, but this seems like a related issue.

I'm going to do a full test of everything and then hopefully we can merge.

m-col:
I'm also think about what tests we should add. One that would be good is testing dragging windows from fullscreen i.e. `lazy.window.set_position_floating()`

m-col:
Everything seems to be working great. I'm happy to merge this as is.

I've mentioned testing fullscreening on https://github.com/qtile/qtile/discussions/2134. The obvious test (the original one) being flaky might make things complicated for testing but it would be good to at some point ensure all of the fullscreening behaviours are what they should be are kept that way. As currently there is the outstanding issue of floating->fullscreen->tiled, we can leave the fullscreen tests until that's ironed out too.

jwijenbergh:
> I'm finding that floating windows, when made to fullscreen and then un-fullscreen, become tiled. This is also the case on master so it doesn't have to be part of this PR, but this seems like a related issue.
> 
> I'm going to do a full test of everything and then hopefully we can merge.

This is a separate really annoying issue indeed. I have been meaning to take a look at this for the last few months. In short we need to remember the previous state the window was in before it is set fullscreen.

jwijenbergh:
> I'm also think about what tests we should add. One that would be good is testing dragging windows from fullscreen i.e. `lazy.window.set_position_floating()`

Yeah interesting! I am a bit reluctant to test with that tiled window set though, no idea what was going wrong in the flaky test

jwijenbergh:
> Everything seems to be working great. I'm happy to merge this as is.
> 
> I've mentioned testing fullscreening on #2134. The obvious test (the original one) being flaky might make things complicated for testing but it would be good to at some point ensure all of the fullscreening behaviours are what they should be are kept that way. As currently there is the outstanding issue of floating->fullscreen->tiled, we can leave the fullscreen tests until that's ironed out too.

Do you want me to rebase or no need?

elParaguayo:
Probably needs a rebase as I merged the fix for the flaky test.

jwijenbergh:
> Probably needs a rebase as I merged the fix for the flaky test.

Done!

m-col:
Thanks again!

PR #3331, opened 2022-02-12
Add Spiral layout

elParaguayo:
Adds a new, mathematical layout. Windows are arranged in a spiral according to a user-defined ratio.

Draft for now as a bit more to do but comments most welcome.

Looks like this:

To Do:
 - [x] Better icon
 - [x] Tests


m-col:
Very cool! A couple of initial comments from reading the diff:

- `SpiralInfo` is used once and the instance is used exclusively for its `get_spiral()` method. I didn't quite figure out what that returned yet (a list of something), but it looks like `SpiralInfo` is fundamentally a function that has taken the form of a class. I think it would be a bit more intuitive to just make it a function that takes what you're passing to `SpiralInfo` and returns what you're getting from `si.get_spiral()`.
- I think for non-tiny additions like this (well, in fact, also tiny additions) we should be adding type-hints to everything. This makes it much easier to understand what functions do and what class attributes are (and by extension their purpose), and makes it easier to identify places where improvements can be made. Also, mypy's type checking can sometimes be a headache for us but it has revealed a crazy amount of bugs in qtile's code over the years that would have otherwise been hard to track!

Some point soon I'll pull it down and try it out, I remember playing around with this layout in bspwm along time ago and loving it!

elParaguayo:
> Very cool! A couple of initial comments from reading the diff:
> 
> * `SpiralInfo` is used once and the instance is used exclusively for its `get_spiral()` method. I didn't quite figure out what that returned yet (a list of something), but it looks like `SpiralInfo` is fundamentally a function that has taken the form of a class. I think it would be a bit more intuitive to just make it a function that takes what you're passing to `SpiralInfo` and returns what you're getting from `si.get_spiral()`.

100% agreed. This is copied from `RatioTile` but, as I was doing it, I did think it was unnecessary. Will re-write.

> * I think for non-tiny additions like this (well, in fact, also tiny additions) we should be adding type-hints to everything. This makes it much easier to understand what functions do and what class attributes are (and by extension their purpose), and makes it easier to identify places where improvements can be made. Also, mypy's type checking can sometimes be a headache for us but it has revealed a crazy amount of bugs in qtile's code over the years that would have otherwise been hard to track!

Yup. Also agreed. 

Would appreciate thoughts on how/where windows are added and what commands may be useful to expose here.

m-col:
> Would appreciate thoughts on how/where windows are added

You mean where a new window should be placed? If so - interesting question, I'm not sure! I guess the most obvious use cases would be in the main window slot, or the current window slot (pushing the current window and all smaller windows inwards).

> and what commands may be useful to expose here

Hmm... not sure. I can imagine it might be useful to resize the main window, e.g. if your main window is left or right, you might want it wider temporarily if you're reading/watching something that would benefit from it. Just like resizing horizontally in `Columns`. But I guess it would propagate through the whole spiral, so it would be more like increasing or decreasing the ratio in increments? Besides that, there is the focus and move next/previous which would correspond to inwards/outwards.

P.S. your terminal looks super nice!

elParaguayo:
Added a test. Need to fix mypy/typing issue.

elParaguayo:
Hmm. Is the icon too much like Debian's icon now?

m-col:
Ha, neat. You could make it angular maybe?

elParaguayo:
Or rotate it...
rotated

elParaguayo:
I've also realised we could do a different version of this layout where you can just pick which way you want to split the space, rather than having a set order.

m-col:
> I've also realised we could do a different version of this layout where you can just pick which way you want to split the space, rather than having a set order.

We _could_ but _should_ we? :laughing:

elParaguayo:
It would be like that "plasma" layout that somebody else made.

Would be nice to have a layout that you could split in any direction and size windows individually.

m-col:
It does sound pretty cool!

elParaguayo:
OK - will take a look.

If this one passes all its tests, I'll squash it into a single commit.

elParaguayo:
Actually, thinking about it some more, it might be better to make that layout first as `Spiral` is just a special case where splits are set in a pre-determined order.

m-col:
Kinda reminds me of i3/sway where you can toggle between horizontal and vertical splits for newly created tiled windows. Here you could toggle between two options with command and that would determine which direction the next tile would push the spiral. `Spiral` as-is is just the default state without the command ever being used. (This is all assuming we're imagining the same thing :laughing: )

elParaguayo:
https://github.com/numirias/qtile-plasma is (or was, haven't used it in a long time) was a pretty nice implementation of that.

m-col (approved):
All looks good to me :)

Especially the type annotated parts :wink:

Issue #2616, opened 2021-07-10
Volume widget control with pamixer fails


leybodv:
# Issue description

I try to use pamixer for displaying/controlling volume level in Volume widget. In config.py I add 
`widget.Volume(volume_app = "pamixer", get_volume_command = "pamixer --get-volume-human", volume_down_command = "pamixer --decrease 1", volume_up_command = "pamixer --increase 1", mute_command = "pamixer --toggle-mute")`
However widget displays "0" instead of actual volume level.

The possible problem is in `widget.base.call_process` method which invokes `subprocess.check_output` without `shell=True` parameter.
call_process method is invoked in volume.py on 174th line to get volume from environment.

Invoking `subprocess.check_output("pamixer --get-volume-human")` in python shell causes:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/d/git/qtile/libqtile/widget/volume.py", line 174, in get_volume
    mixer_out = self.call_process(get_volume_cmd)
  File "/usr/lib/python3.9/site-packages/libqtile/widget/base.py", line 292, in call_process
    output = subprocess.check_output(command, **kwargs)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'pamixer --get-volume-human'

# Qtile version
0.17.0

# Stack traces

2021-07-10 17:41:18,935 ERROR libqtile base.py:_wrapper():L300 got exception from widget timer
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/libqtile/widget/base.py", line 298, in _wrapper
    method(*method_args)
  File "/usr/lib/python3.9/site-packages/libqtile/widget/volume.py", line 112, in update
    vol = self.get_volume()
  File "/usr/lib/python3.9/site-packages/libqtile/widget/volume.py", line 174, in get_volume
    mixer_out = self.call_process(get_volume_cmd)
[...]

Issue #2614, opened 2021-07-08
Volume widget showing mute only and not doing anything on input


debarkamondal:
I am using pipewire when i use pavucontrol i can access all the devices but the volume widget is stuck on mute and is not showing any value or doing any function like volume on scoll

m-col:
Please post a config, version, and relevant log output as per the issue template

debarkamondal:
Config: [Qtile config](https://github.com/debarkamondal/.dotfiles/blob/main/.config/qtile/config.py)
Version: 0.17.0-1
Log: Sorry to say but where to do I get a log file

m-col:
The log is at ~/.local/share/qtile/qtile.log, check to see if anything 
is logged during startup and when you hit the keybinding / mouse button 
when you expect the actions to happen.

Probably worth updating too - 0.18.0 was released recently and might 
have fixed it.

debarkamondal:
I think maybe the problem is somewhere else because alsamixer is no longer showing my sound card but it appears and works normally in pavucontrol

Fabien-Bernard:
If you use pulseaudio, it seems interesting to emphasize the existence of the PulseVolume widget (as mentioned at the end of this thread : https://groups.google.com/g/qtile-dev/c/1ykdOEpsziw ).

debarkamondal:
Sorry for the problem caused. But it is now working fine now that i have both pulseaudio and pipewire installed

PR #3332, opened 2022-02-12
Fix Volume widget compatibility with different commands and PipeWire

adolfgatonegro:
Fixes broken functionality of the Volume widget when trying to use commands other than `amixer` (e.g.  #2616) .

This also addresses the incompatibility of the widget with PipeWire (e.g. #2614), possibly making the `PulseVolume` widget unnecessary.

Added two additional configuration parameters, `check_mute_command` and `check_mute_string`, since different audio utilities show the mute status in different ways, and they will fail the hardcoded search for `[off]` in `mixer_out`. This way the user can specify the output that Qtile should look for in their custom command.

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -75,7 +75,7 @@ class Volume(base._TextBox):
         (
             "get_volume_command",
             None,
-            "Command to get the current volume."
+            "Command to get the current volume. "
             "The expected output should include 1-3 numbers and a ``%`` sign.",
         ),
         ("check_mute_command", None, "Command to check mute status"),
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -41,8 +41,7 @@ __all__ = [
     "Volume",
 ]
 
-re_vol = re.compile(r"\[(\d?\d?\d?)%\]")
-
+re_vol = re.compile(r"(\d?\d?\d?)%")
 
 class Volume(base._TextBox):
     """Widget that display and change volume
@@ -73,6 +72,8 @@ class Volume(base._TextBox):
         ("volume_up_command", None, "Volume up command"),
         ("volume_down_command", None, "Volume down command"),
         ("get_volume_command", None, "Command to get the current volume"),
+        ("check_mute_command", None, "Command to check mute status"),
+        ("check_mute_string", "[off]", "String to look for to check muted status in the output of ``check_mute_command``."),
         (
             "step",
             2,
@@ -186,16 +187,22 @@ class Volume(base._TextBox):
             if self.get_volume_command:
                 get_volume_cmd = self.get_volume_command
 
-            mixer_out = self.call_process(get_volume_cmd)
+            mixer_out = subprocess.getoutput(get_volume_cmd)
         except subprocess.CalledProcessError:
             return -1
 
-        if "[off]" in mixer_out:
+        check_mute = mixer_out
+        if self.check_mute_command:
+            check_mute = subprocess.getoutput(self.check_mute_command)
+        
+        if self.check_mute_string in check_mute:
             return -1
 
         volgroups = re_vol.search(mixer_out)
         if volgroups:
             return int(volgroups.groups()[0])
+        elif "pamixer" in self.get_volume_command:
+            return mixer_out
         else:
             # this shouldn't happen
             return -1
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -73,7 +73,11 @@ class Volume(base._TextBox):
         ("volume_down_command", None, "Volume down command"),
         ("get_volume_command", None, "Command to get the current volume"),
         ("check_mute_command", None, "Command to check mute status"),
-        ("check_mute_string", "[off]", "String to look for to check muted status in the output of ``check_mute_command``."),
+        (
+            "check_mute_string",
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for submitting this. I'll take a look when I get a moment.

libqtile/widget/volume.py
```diff
@@ -185,16 +190,22 @@ def get_volume(self):
...
+            check_mute = subprocess.getoutput(self.check_mute_command)
+
+        if self.check_mute_string in check_mute:
             return -1
 
         volgroups = re_vol.search(mixer_out)
         if volgroups:
             return int(volgroups.groups()[0])
+        elif "pamixer" in self.get_volume_command:
+            return mixer_out
```
elParaguayo, line 208: I'm not sure about these lines.

Firstly, the method is mean to return an integer but `subprocess.getoutput` returns a string.
Secondly, even if you convert the output to an integer, I suspect that there are other `pamixer` commands that wouldn't give the right output.
Thirdly, what's special about `pamixer` that means we only do this check for that application?

elParaguayo:
These changes should get rid of the test failures.

``` diff
diff --git a/libqtile/widget/volume.py b/libqtile/widget/volume.py
index f198c34e..8b080ace 100644
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -42,6 +42,7 @@ __all__ = [

 re_vol = re.compile(r"(\d?\d?\d?)%")

+
 class Volume(base._TextBox):
     """Widget that display and change volume

@@ -75,7 +76,7 @@ class Volume(base._TextBox):
         (
             "check_mute_string",
             "[off]",
-            "String to look for to check muted status in the output of ``check_mute_command``."
+            "String to look for to check muted status in the output of ``check_mute_command``.",
         ),
         (
             "step",
```

AdolfGatonegro:
Fixed. Let's see how it goes. 👌

AdolfGatonegro:
After some more testing, the `pamixer` volume commands are parsed correctly by the regex even if the `%` is present. The special check for `pamixer` was not needed after all. These changes work correctly with the default `amixer`, `pamixer`, and `pactl` as intended.

AdolfGatonegro:
Alright, supporting different commands with their various outputs is a bit of a mess. The current implementation should work with any audio tool that outputs the volume level as 1-3 digits plus a percentage sign: `ddd%`. This means that `pamixer` should be used with the `--get-volume-human` instead of just `--get-volume`. 

I think this would be the simplest way to handle it, and let the user know what the widget expects as volume output. Trying to cover for all the idiosyncrasies of the various Linux audio tools would require additional code to check for exceptions and weird edge cases.

libqtile/widget/volume.py
```diff
@@ -185,11 +196,15 @@ def get_volume(self):
             if self.get_volume_command:
                 get_volume_cmd = self.get_volume_command
 
-            mixer_out = self.call_process(get_volume_cmd)
+            mixer_out = subprocess.getoutput(get_volume_cmd)
```
elParaguayo, line 199: What's the reason for this change?

libqtile/widget/volume.py
```diff
@@ -71,7 +71,18 @@ class Volume(base._TextBox):
...
+            "get_volume_command",
+            None,
+            "Command to get the current volume."
+            "The expected output should include 1-3 numbers and a ``%`` sign.",
+        ),
+        ("check_mute_command", None, "Command to check mute status"),
+        (
+            "check_mute_string",
+            "[off]",
+            "String to look for to check muted status in the output of ``check_mute_command``.",
```
elParaguayo, line 84: Can you make it clear that, if this string is matched, then the volume is treated as unmuted (assuming I've got that right)?

libqtile/widget/volume.py
```diff
@@ -71,7 +71,18 @@ class Volume(base._TextBox):
         ("volume_app", None, "App to control volume"),
         ("volume_up_command", None, "Volume up command"),
         ("volume_down_command", None, "Volume down command"),
-        ("get_volume_command", None, "Command to get the current volume"),
+        (
+            "get_volume_command",
+            None,
+            "Command to get the current volume."
```
elParaguayo, line 77: You need a space after the `.` as the next line will be concatenated.

elParaguayo (changes requested):

m-col:
Apologies for this getting forgotten about. I've just tested it and it works as expected, changes LGTM. I've just triggered the tests, hopefully they're all happy.

adolfgatonegro:
No worries, after the last review I kind of forgot to commit the changes elParaguayo requested, and the PR just sat here for a bit. Looks like all the tests except Python 3.11 were successful, not sure what that's about.

libqtile/widget/volume.py
elParaguayo: Thanks for sharing this.

It's an interesting scenario as `getoutput` doesn't return an error when it's passed a list but it seems that it just runs the first item in the list.

Do you want to submit a fix for this?

inmaldrerah: I think just falling back to `self.call_process` would work. By the way, the current version "happens to work", because the output of the command intended to call, by default `amixer -D default sget Master`, is just part of the output of `amixer`, and the first `\d?\d?\d?%` happens to be the volume of `Master`.

PR #3335, opened 2022-02-13
WL: logger call is using wrong string format

m-col:
This will raise an exception when called.

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -340,4 +340,4 @@ def configure_device(device: InputDevice, configs: dict[str, InputConfig]) -> No
     elif device.device_type == input_device.InputDeviceType.KEYBOARD:
         _configure_keyboard(device, conf)
     else:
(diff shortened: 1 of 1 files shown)


elParaguayo:
Should we be more consistent between that style and f-strings?

m-col:
I would certainly love for most string formatting in the codebase to be consistent, but the case with `logger` is a bit different. See here for a bit more discussion: https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/ but to summarise, if you pass to `logger` a pre-formatted string then you've done the runtime work to format the string 100% of the time, whereas if you pass the arguments in as the `*args` to logger calls then the string formatting is _only_ done if that logger call is included in the current logger mode i.e. `logger.debug("text and %s", some_object)` will not execute the string formatting unless the logger is in debug mode. So it's a slight optimisation.

elParaguayo:
Thanks. Hadn't considered that.

I've almost certainly used some f-strings in `logger` messages but maybe it's less likely to be an issue for `warning` levels ad above as they're more likely to be logged all of the time whereas `debug` messages wouldn't be.

PR #3340, opened 2022-02-14
Move some layout/widget constructor args to defaults

elParaguayo:
By having `columns` as a kwarg in `Matrix.__init__` the argument was excluded from the configuration table in the docs.

Moving it to `defaults` should make configuration a bit clearer for users.

### DIFF
--- a/libqtile/layout/matrix.py
+++ b/libqtile/layout/matrix.py
@@ -24,9 +24,12 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+from __future__ import annotations
+
 import math
 
 from libqtile.layout.base import _SimpleLayoutBase
+from libqtile.log_utils import logger
 
 
 class Matrix(_SimpleLayoutBase):
@@ -40,13 +43,19 @@ class Matrix(_SimpleLayoutBase):
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
         ("border_width", 1, "Border width."),
+        ("columns", 2, "Number of columns"),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
     ]
 
-    def __init__(self, columns=2, **config):
+    def __init__(self, _columns: int | None = None, **config):
         _SimpleLayoutBase.__init__(self, **config)
         self.add_defaults(Matrix.defaults)
-        self.columns = columns
+        if _columns:
+            logger.warning(
+                "The use of a positional argument in Matrix is deprecated. "
+                "Please update your config to use columns=..."
+            )
+            self.columns = _columns
 
     @property
     def rows(self):
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -17,12 +17,16 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
 from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from libqtile import bar
 from libqtile.log_utils import logger
 from libqtile.widget import Systray, base
+
+if TYPE_CHECKING:
+    from typing import Any
 
 
 class WidgetBox(base._Widget):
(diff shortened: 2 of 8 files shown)


m-col:
Nice. Do we have others like it?

elParaguayo:
Found a few more!

elParaguayo:
Ah, interesting issue here.

Take `WidgetBox` as an example. The old contructor was `WidgetBox(widgets=[])`. This means you could initialise it with `WidgetBox([widget.Textbox("...")])`. That would fail following these changes.

Three options:
1) Do a `migrate` script to change a positional argument to a keyword argument
2) Allow the positional argument but add a deprecation warning
3) 1 and 2!

3 is probably the best approach I'm just a bit sceptical as to how many people actually use `migrate`. Ideally we'd want packagers to include a note advising users to run migrate on their configs after upgrading.

Thoughts?

elParaguayo:
The desire is to keep all widget/layout considerations in the `defaults` value so users can just look in one place (i.e. the table) to understand how to configure the object.

`WidgetBox` has a keyword argument `widgets` and the docstring includes an example showing how to configure the widget:
``` python
widget.WidgetBox(widgets=[
    widget.TextBox(text="This widget is in the box"),
    widget.Memory()
    ]
),
```

However, users could also configure the widget by passing the list as a positional argument:
``` python
widget.WidgetBox([
    widget.TextBox(text="This widget is in the box"),
    widget.Memory()
    ]
),
```

The former will work with my change but the second (which is currently used in the tests) will break. I'd like to protect against breaking people's configs.

m-col:
For could support both for a time to allow for migrating:
```diff
diff --git a/libqtile/widget/widgetbox.py b/libqtile/widget/widgetbox.py
index 46032219..51518e79 100644
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -61,13 +61,17 @@ class WidgetBox(base._Widget):
         ("text_open", "[>]", "Text when box is open"),
     ]
 
-    def __init__(self, widgets: list = None, **config):
+    def __init__(self, _widgets: list = None, **config):
         base._Widget.__init__(self, bar.CALCULATED, **config)
         self.add_defaults(WidgetBox.defaults)
         self.box_is_open = False
         self.widgets = widgets if widgets is not None else []
         self.add_callbacks({"Button1": self.cmd_toggle})
 
+        if _widgets:
+            logger.warning...
+            self.widgets = _widgets
+
         self.close_button_location: str
         if self.close_button_location not in ["left", "right"]:
             val = self.close_button_location

```

elParaguayo:
Yes. I think that's the best approach. Will do a `migrate` script too.

elParaguayo:
What do you think? Do we need a migrate script for this?

libqtile/layout/matrix.py
```diff
@@ -40,13 +43,19 @@ class Matrix(_SimpleLayoutBase):
...
 
-    def __init__(self, columns=2, **config):
+    def __init__(self, _columns: int | None = None, **config):
         _SimpleLayoutBase.__init__(self, **config)
         self.add_defaults(Matrix.defaults)
-        self.columns = columns
+        if _columns:
+            logger.warning(
+                "The use of a positional argument in Matrix is deprecated. "
+                "Please use update your config to use columns=..."
```
m-col, line 56: There is a typo here

elParaguayo: Oops.

elParaguayo:
I'm still very sceptical about `migrate`. I don't think some of the migrations works as well as we expect.

For example the `BitcoinTicker`/`CryptoTicker` will only work if the widget imports `BitcoinTicker` directly, not when it's `widget.BicoinTicker` but that's how we import widgets in the default config.

m-col:
Oh, really? Interesting. What do you propose?

elParaguayo:
I don't know what the best solution is. To catch both `BitcoinTicker` and `widget.BitcoinTicker` I think you need a custom selector which then means that `rename` doesn't work. It's messy.

My main concern is that we spend a good amount of time working out how on earth to write these migrate scripts (writing one for this PR has taken me a good few hours so far) but:

a) they don't work as we want them to and
b) we've no idea if anyone is even using it (because we don't advertise that users need to run `qtile migrate`).

It feels like it would just be easier to deprecate features and put messages in logs.

elParaguayo:
Have added a single migrate script (for `WIdgetBox`) to show you what I mean. 

If you think it's worth the effort, I'll refactor so we can quickly add `Matrix` and `LaunchBar`.

libqtile/scripts/migrate.py
```diff
@@ -125,6 +145,45 @@ def windowtogroup_groupName_argument(funcname, query):  # noqa: N802
     return query.select_method(funcname).modify_argument("groupName", "group_name")
 
 
+def modify(node, capture, filename):
```
elParaguayo, line 148: I know I need to rename this...

elParaguayo:
So `bowler` has said they can't support past python 3.8 (in their readme) so I'm still not inclined to do the migrate scripts, I'd prefer the deprecation warning messages.

m-col:
Yeah, fair enough. We tried...

elParaguayo:
Have removed the migrate script so assume we're ok to merge this one then?

libqtile/widget/widgetbox.py
```diff
@@ -17,13 +17,17 @@
...
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from libqtile import bar
 from libqtile.log_utils import logger
 from libqtile.widget import Systray, base
 
+if TYPE_CHECKING:
+    from typing import Any
```
m-col, line 29: It looks like `Any` isn't used anywhere - or I missed it?

m-col (approved):
Other than that comment I just made, this looks great :)

libqtile/widget/widgetbox.py
elParaguayo: Will check after work but I suspect you're right.

elParaguayo: It's used here: https://github.com/qtile/qtile/pull/3340/files#diff-0455ac9a3773d4a03dd70d6870c17c0e213481c9794da34b01cd201d3f08be83R70

I think you still need the import even though the typing is in a `# type` line.

m-col: Ah yes - gotcha

PR #3343, opened 2022-02-14
Use system python version for mypy in qtile check

### LINKED ISSUE #3342: Use current python version for qtile check
When running `qtile check`, the version of python used is hardcoded to 3.10 (on current master), as can be seen here:
https://github.com/qtile/qtile/blob/c1f5073935a61f2bb9a36e198791590410ec07a1/libqtile/scripts/check.py#L101

Since qtile will be run with the system python, which does not necessarily correspond to 3.10, it would make more sense to use that to decide the version. This can either be done by removing the `--python-version` argument to `mypy` (since by default it "assumes your code will be run using the same […] Python version as you are using to run mypy itself"), or by utilizing `sys.version_info` to generate the version string that is passed to `mypy`.

If you would like, I can create a pull request for this.

MithicSpirit:
Instead of hardcoding the python version, this instead lets mypy decide it based on the system python version. The version restriction was originally put in place due to `Literal` being used, which is only available in 3.8+, so the mypy check is skipped if the current python version is older than that.

Closes #3342.

### DIFF
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -98,7 +98,7 @@ def type_check_config_args(config_file):
     try:
         # we want to use Literal, which is in 3.8. If people have a mypy that
         # is too old, they can upgrade; this is an optional check anyways.
-        subprocess.check_call(["mypy", "--python-version=3.10", config_file])
+        subprocess.check_call(["mypy", config_file])
         print("config file type checking succeeded")
     except subprocess.CalledProcessError as e:
         print("config file type checking failed: {}".format(e))
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -95,8 +95,7 @@ def type_check_config_args(config_file):
     if shutil.which("mypy") is None:
         print("mypy not found, can't type check config file" "install it and try again")
         return
-    ver = sys.version_info
-    if (ver.major < 3) or (ver.major == 3 and ver.minor < 8):
+    if  sys.version_info.minor < 8:  # < 3.8
         print(
             "mypy check not supported for the current version of python, "
             + "please update python to at least 3.8 and try again"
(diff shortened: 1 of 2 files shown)


libqtile/scripts/check.py
```diff
@@ -95,9 +95,14 @@ def type_check_config_args(config_file):
     if shutil.which("mypy") is None:
         print("mypy not found, can't type check config file" "install it and try again")
         return
+    ver = sys.version_info
+    if (ver.major < 3) or (ver.major == 3 and ver.minor < 8):
```
m-col, line 99: I think we can skip the `ver.major` check and just check `sys.version_info.minor`.

m-col:
Great, thanks for this.

Could you also enable the mypy tox env for github actions? It includes the `qtile check` tests. Here is a diff:

```diff
diff --git a/tox.ini b/tox.ini
index 728c1241..66c94de4 100644
--- a/tox.ini
+++ b/tox.ini
@@ -113,6 +113,6 @@ commands =
 python =
     pypy-3.7: pypy3
     3.7: py37
-    3.8: py38
-    3.9: py39
-    3.10: py310, packaging, pep8, codestyle, docstyle, mypy, vulture
+    3.8: py38, mypy
+    3.9: py39, mypy
+    3.10: py310, mypy, packaging, pep8, codestyle, docstyle, vulture
```

libqtile/scripts/check.py
MithicSpirit: Fair enough. The script will probably fail earlier on python2, and python4 probably won't release for a _long_ time.

m-col:
CI is failing due to formatting. Looks like you need to run `black` over the code:
```
  libqtile/scripts/check.py:98:7: E271 multiple spaces after keyword
  libqtile/scripts/check.py:98:8: BLK100 Black would make changes.
```

MithicSpirit:
Oh wow that 3.8 check really is taking a while huh. Regardless, I believe the issue is fixed now; sorry about that.

PR #3347, opened 2022-02-17
Make LaunchBar font configurable

elParaguayo:
When an icon can't be found, `LaunchBar` defaults to displaying text. The text is formatted according to `base._TextBox`'s defaults. This will pick up values set by `widget_defaults` but doesn't allow individual customisation for this widget.

This PR fixes this issue by allowing users to configure the appearance of text.


m-col:
Makes sense. Thanks!

PR #3354, opened 2022-02-18
Make Battery notification timeout configurable

### LINKED ISSUE #3353: Battery low notifications go away with `dunst`
yesterday, thanks to @m-col, I manage to have beautiful popup notifications for my battery from within `qtile`'s bar and even sound and brightness bars as I change them with the keyboard:
notifications dunst
However it is still not behaving quite as I expected.

I expected the notifications for a low battery to be of *urgent*/*critical* priority and it is red indeed.
I configured my [dunst](https://github.com/a2n-s/dotfiles/blob/main/.config/dunst/dunstrc#L292) so that *low* and *normal* notifications go away quite quicly, but *critical*/red ones never go away!

This is what I observe when I run either
```
notify-send "ERROR" "this is an error" -u critical
notify-send "NORMAL" "this is a normal message" -u normal
notify-send "LOW" "this is nothing important" -u low
```
or
```
dunstify "ERROR" "this is an error" -u critical
dunstify "NORMAL" "this is a normal message" -u normal
dunstify "LOW" "this is nothing important" -u low
```

But this is **not the case with the *critical* battery low notifications** from `qtile`.

I do not really understand the code of the `libqtile.widget.Battery` class, could anyone help fix that?

elParaguayo:
Users may want a different timeout other than the default 10 seconds.

Closes #3353

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -330,6 +330,7 @@ class Battery(base.ThreadPoolText):
         ("update_interval", 60, "Seconds between status updates"),
         ("battery", 0, "Which battery should be monitored (battery number or name)"),
         ("notify_below", None, "Send a notification below this battery level."),
+        ("notification_timeout", 10, "Time in seconds to display notification. 0 for no expiry."),
     ]
 
     def __init__(self, **config) -> None:
@@ -345,6 +346,7 @@ class Battery(base.ThreadPoolText):
 
         self._battery = self._load_battery(**config)
         self._has_notified = False
+        self.timeout = int(self.notification_timeout * 1000)
 
     def _configure(self, qtile, bar):
         if not self.low_background:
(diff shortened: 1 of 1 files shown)


m-col:
Changes itself LGTM. The test failure is unrelated but cancelled the py310 tests so I've run them again.

libqtile/widget/battery.py
```diff
@@ -345,6 +346,7 @@ def __init__(self, **config) -> None:
 
         self._battery = self._load_battery(**config)
         self._has_notified = False
+        self.timeout = int(self.notification_timeout * 1000)
```
m-col, line 349: Maybe we can do away with this line and accept a millisecond argument

m-col: Oh nevermind, I merged :laughing:

elParaguayo: I thought about that. Kept it as seconds as that's consistent with `update_interval` for other widgets.

PR #3356, opened 2022-02-19
x11: move x11 connection used by many tests into fixture

m-col:
Many x11-backend tests use a manually created connection to the X server
and finalize it after use. Let's move that out into a fixture for them
to use to ensure connections are finalized and clean up the code a bit.


elParaguayo:
Nice and neat. LGTM.

PR #3359, opened 2022-02-19
Add ability to test logging messages

elParaguayo:
By default pytest's `caplog` fixture does not work when we run tests in a `manager` process.

This PR adds a `logger` fixture which uses a `QueueHandler` to pass messages from a separate process to the main thread which `caplog` can then read.

The PR also includes an example of its use with `test_images.py`.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -20,6 +20,8 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import logging
+import multiprocessing
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


m-col:
Neat!

test/conftest.py
```diff
@@ -103,8 +105,30 @@ def backend(request, backend_name, xephyr, wayland_session):
...
+
+    yield caplog
+
+    listener.stop()
+
+
+@pytest.fixture(scope="function")
+def manager_nospawn(request, backend, log_queue):
     with TestManager(backend, request.config.getoption("--debuglog")) as manager:
+        manager.log_queue = log_queue
```
m-col, line 131: Is there a reason it's only enabled for `manager_nospawn` and not `manager` too?

elParaguayo: Because `manager` is just `manage_nospawn` that is started before the fixture is yielded.

m-col: doh! :laughing: In that case, `TestManager` instances always have the log_queue attribute, making the `hasattr` check elsewhere redundant, right?

elParaguayo: Not quite. The `xmanager` fixture also creates a `TestManager` instance but I haven't added the log queue there as that seemed to be a different use case. Without the `hasattr` check, any tests using that fixture would fail.

m-col: I see - OK no problem

m-col (approved):
LGTM!

PR #3362, opened 2022-02-19
Add JSON encoder to IPC for set types

### LINKED ISSUE #3360: Groups cannot be listed using JSON IPC
Using the JSON serialization format in the IPC and trying to list groups yields exception 
```
Object of type set is not JSON serializable
NoneType: None
```
in the qtile logs and causes a `IPCError: Unable to decode json data` in the client end.

Reproduce with:
```python
from libqtile.command.client import (
    InteractiveCommandClient,
    Client,
    find_sockfile,
    IPCCommandInterface,
)
c = InteractiveCommandClient(
    command=IPCCommandInterface(
        Client(find_sockfile(), is_json=True)
    )
)
c.groups()
```

Jaakkonen:
Closes #3360. Fixes bug in which groups couldn't be listed over IPC when using JSON
encoding.

### DIFF
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -129,12 +129,19 @@ class _IPC:
     def pack(msg: Any, *, is_json: bool = False) -> bytes:
         """Pack the object into a message to pass"""
         if is_json:
-            json_obj = json.dumps(msg)
+            json_obj = json.dumps(msg, default=_IPC._json_encoder)
             return json_obj.encode()
 
         msg_bytes = marshal.dumps(msg)
         size = struct.pack(HDRFORMAT, len(msg_bytes))
         return size + msg_bytes
+
+    @staticmethod
+    def _json_encoder(field: Any) -> Any:
+        """Convert non-serializable types to ones understood by stdlib json module"""
+        if isinstance(field, set):
+            return list(field)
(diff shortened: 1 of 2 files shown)


elParaguayo:
Nice. Can we include a quick test here?

Jaakkonen:
> Nice. Can we include a quick test here?

Sure! I'll see how the tests are set up. Couple minutes :)

libqtile/ipc.py
```diff
@@ -129,13 +129,20 @@ def unpack(data: bytes, *, is_json: bool | None = None) -> tuple[Any, bool]:
...
-            json_obj = json.dumps(msg)
+            json_obj = json.dumps(msg, default=_IPC._json_encoder)
             return json_obj.encode()
 
         msg_bytes = marshal.dumps(msg)
         size = struct.pack(HDRFORMAT, len(msg_bytes))
         return size + msg_bytes
 
+    @staticmethod
+    def _json_encoder(field: Any):
```
elParaguayo, line 140: Can you set the return type to `Any` as well.

elParaguayo:
FYI - we're happy with force pushes on PRs to keep the number of commits down so, if possible, please amend your original commit rather than adding new ones.

elParaguayo:
Great. Looks good to me.

PR #3368, opened 2022-02-21
Add lazy.screen.set_wallpaper

elParaguayo:
Exposes a command to allow users to change wallpapers via the command interface. Includes an example in the docs.


m-col:
LGTM!

PR #3373, opened 2022-02-22
Fix resizing of wallpaper when screen scaled

### LINKED ISSUE #3370: Wallpaper not resizing when display scaled
Hello, first of all thank you for the amazing job you are doing. Qtile is amazing!

I have an issue I am not able to solve. Not sure if it is a bug in qtile or just an issue related to my config. 

From time to time, I connect a secondary monitor to my laptop. I want to set a refresh rate for this monitor of 120 and scale it by 10%. When I am on Gnome, I do not have any issues. When on Qtile, I cannot seem to make it work. I have tried the following:

1) make qtile run an autostart file with an xrandr command in it:
 
```
@hook.subscribe.startup
def autostart():
    home = os.path.expanduser('~/.config/qtile/autostart.sh')
    subprocess.run([home])
```
```
#!/usr/bin/sh
sh /home/eugenio/.local/bin/auto-screen-rotate.sh eDP-1 "Wacom HID 4956 Finger touch" &
picom &
xrandr --output eDP-1 --primary --output DP-1 --scale 1.1x1.1 --right-of eDP-1 --mode 1920x1280 --rate 120 &
sleep 1 &&
feh --no-fehbg --bg-scale /home/eugenio/Sync/Sfondi/wallhaven-vg6dqp.jpg --bg-scale /home/eugenio/Sync/Sfondi/1245752.jpg &
sleep 10 && xinput map-to-output 'Wacom HID 4956 Finger touch' eDP-1 &

```
While the other commands are working correctly, the xrandr command is ineffective.

2) start qtile from an xsession:
```
#!/usr/bin/bash

xsetroot -solid gray 
xrandr --output eDP-1 --primary --output DP-1 --scale 1.1x1.1 --right-of eDP-1 --mode 1920x1280 --rate 120 
qtile start
```
As in the previous case, qtile starts but the xrandr command is not working.

What I am doing wrong? Or is this a problem with qtile setting xrandr on its own and overriding my configuration?

Thank you very much.

elParaguayo:
Wallpaper was drawn based on the screen's `width/height_in_pixels` but this doesn't take account of scaling.

This PR draws wallpapers based on `screen.width/height` instead.

Closes #3370

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           on focused windows when none is focused. By default we do not match on focused windows,
           to change this set `if_no_focused` to True.
         - Widget with duplicate names will be automatically renamed by appending numeric suffixes
+        - Fix resizing of wallpaper when screen scale changes (X11)
 
 Qtile 0.20.0, released 2022-01-24:
     * features
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -686,6 +686,8 @@ class Painter:
         self.default_screen = self.screens[self.conn.pref_screen]
         self.conn.core.SetCloseDownMode(xcffib.xproto.CloseDown.RetainPermanent)
         self.atoms = AtomCache(self)
+        self.width = -1
+        self.height = -1
 
     def paint(self, screen, image_path, mode=None):
         try:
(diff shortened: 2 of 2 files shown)


m-col (approved):
LGTM

elParaguayo:
Any chance you could give it a test and try some scaling with `xrandr`?

elParaguayo:
OK - so it doesn't work on multiple monitors. Pretty sure I know why (we need to create a root pixmap that covers all monitors - my changes just resize to a single screen).

Tricky bit is getting the full size when scaling is involved as this isn't reflected in any of the available attributes (as far as I can tell). Doing `Connection.randr.query_crtcs(root)` get me the coords and (scaled) dimensions of the screens but I can't seem to access this from `Painter` but it could just be because I'm tired!

elParaguayo:
Should be working now: https://github.com/qtile/qtile/issues/3370#issuecomment-1048269942

PR #3374, opened 2022-02-22
Suppress log messages when exiting

### LINKED ISSUE #3351: Qtile loses connection with X when shutting down
Nothing concerning here, but it seems like Qtile is not shutting down fast enough whenever I "shutdown now" and throws errors right before computer shut down. Is there a way to make qtile end before X ?

Again, nothing concerning, but it's a bit messy. (Maybe my system is too slow ?)

```
2022-02-17 21:13:16,855 ERROR libqtile core.py:_xpoll():L323 Shutting down due to X connection error XCB_CONN_ERROR (1)
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 571, in wrapper
    return f(*args)
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 610, in poll_for_event
    self.invalid()
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 559, in invalid
xcffib.ConnectionException: xcb connection errors because of socket, pipe and other stream errors.

  File "/usr/lib/python3.10/site-packages/libqtile/backend/x11/core.py", line 285, in _xpoll
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 573, in wrapper
    self.invalid()
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 559, in invalid
xcffib.ConnectionException: xcb connection errors because of socket, pipe and other stream errors.
2022-02-17 21:13:16,856 ERROR libqtile manager.py:_finalize_configurables():L293 exception during finalize
  File "/usr/lib/python3.10/site-packages/libqtile/core/manager.py", line 281, in _finalize_configurables
    widget.finalize()
  File "/usr/lib/python3.10/site-packages/libqtile/widget/base.py", line 232, in finalize
    self.drawer.finalize()
  File "/usr/lib/python3.10/site-packages/libqtile/backend/x11/drawer.py", line 37, in finalize
    self._free_pixmap()
  File "/usr/lib/python3.10/site-packages/libqtile/backend/x11/drawer.py", line 117, in _free_pixmap
    self.qtile.core.conn.conn.core.FreePixmap(self._pixmap)
  File "/usr/lib/python3.10/site-packages/xcffib/xproto.py", line 2863, in FreePixmap
    return self.send_request(54, buf, is_checked=is_checked)
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 391, in send_request
    seq = self.conn.send_request(flags, xcb_parts + 2, xcb_req)
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 569, in wrapper
    self.invalid()
  File "/usr/lib/python3.10/site-packages/xcffib/__init__.py", line 559, in invalid
xcffib.ConnectionException: xcb connection errors because of socket, pipe and other stream errors.
```

elParaguayo:
Logs are occasionally filled with `ConnectionException` messages before exiting.

This PR suppresses those messages when we know qtile is already exiting and where losing a connection to the X server is
expected.

Might close #3351 (if my hunch is correct)

NOTE: this is untested as I'm not sure how to trigger the error messages reliably.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ Qtile x.x.x, released xxxx-xx-xx:
           on focused windows when none is focused. By default we do not match on focused windows,
           to change this set `if_no_focused` to True.
         - Widget with duplicate names will be automatically renamed by appending numeric suffixes
+        - Suppress `ConnectionException` messages when shutting down (X11).
 
 Qtile 0.20.0, released 2022-01-24:
     * features
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -37,6 +37,7 @@ from libqtile import config, hook, utils
 from libqtile.backend import base
 from libqtile.backend.x11 import window, xcbq
 from libqtile.backend.x11.xkeysyms import keysyms
+from libqtile.core.lifecycle import lifecycle
 from libqtile.log_utils import logger
 from libqtile.utils import QtileError
 
@@ -322,7 +323,11 @@ class Core(base.Core):
                 pass
             except Exception:
                 error_code = self.conn.conn.has_error()
-                if error_code:
+                if lifecycle.behavior == lifecycle.behavior.TERMINATE:
+                    # We're already shutting down so we don't need to do anything here
+                    # and flooding the log with ConnectionExceptions isn't helpful
+                    return
+                elif error_code:
                     error_string = xcbq.XCB_CONN_ERRORS[error_code]
                     logger.exception(
                         "Shutting down due to X connection error {error_string} ({error_code})".format(


m-col:
I can trigger it by using the xephyr script and Ctrl-C-ing the process from the terminal. Presumably this is equivalent to Qtile's signal handler triggering `Qtile.stop`.

I want to take a look at prevent the messages rather than hiding them before merging this.

jwijenbergh:
hmmm so is this only a 'fix' for when we expect it to close? Because when you shutdown your pc I imagine that the x server can get closed before qtile (not sure about that though, probs depends on set up). But when I run this in a xephyr session and then pkill xephyr I still get these errors.  For what I understand from the original issue, we want to never show these? Maybe instead of the full exception we can show a warning in the log such as "qtile shutting down because it lost connection to the x server". That way the log file does not get filled too much

jwijenbergh:
> I can trigger it by using the xephyr script and Ctrl-C-ing the process from the terminal. Presumably this is equivalent to Qtile's signal handler triggering `Qtile.stop`.
> 
> I want to take a look at prevent the messages rather than hiding them before merging this.

You beat me to it, what a coincidence lol

elParaguayo:
> hmmm so is this only a 'fix' for when we expect it to close? Because when you shutdown your pc I imagine that the x server can get closed before qtile (not sure about that though, probs depends on set up). 

I don't love this PR either but that's the scenario I was envisaging here. 

If we can isolate the real cause then I'm more than happy to scrap this one!

jwijenbergh:
Yeah it's hard to figure out,

> > hmmm so is this only a 'fix' for when we expect it to close? Because when you shutdown your pc I imagine that the x server can get closed before qtile (not sure about that though, probs depends on set up).
> 
> I don't love this PR either but that's the scenario I was envisaging here.
> 
> If we can isolate the real cause then I'm more than happy to scrap this one!

Yeah it's really hard to figure out where exactly a check for this needs to be added, I've tried to completely comment out the logging in that section you changed but I still get ConnectionExceptions all around the code. For example just getting window properties here https://github.com/qtile/qtile/blob/f056a7b417fedd3573abd96ad06fd986f5ddbc06/libqtile/backend/x11/window.py#L374

elParaguayo:
Shall we just close this then?

jwijenbergh:
Maybe it fixes it for some scenarios? I don't know. The changes you made do make sense to me though

jwijenbergh:
What about this? This works for me, it simply just checks if its an XCB connection error, and if so it logs on debug. We have to disable logging after that as otherwise other qtile functions will log exceptions

```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index bc331fdd..5623da50 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -324,11 +324,20 @@ class Core(base.Core):
                 error_code = self.conn.conn.has_error()
                 if error_code:
                     error_string = xcbq.XCB_CONN_ERRORS[error_code]
-                    logger.exception(
-                        "Shutting down due to X connection error {error_string} ({error_code})".format(
-                            error_string=error_string, error_code=error_code
+
+                    # This is mostly a normal error that happens
+                    # when we lose connection to the X server
+                    # just log as debug, an exception confuses users
+                    if error_string == "XCB_CONN_ERROR":
+                        logger.debug("Shutting down due to losing connection to the X server")
+                        # Disable further logging to not spam the log file
+                        logger.disabled = True
+                    else:
+                        logger.exception(
+                            "Shutting down due to X connection error {error_string} ({error_code})".format(
+                                error_string=error_string, error_code=error_code
+                            )
                         )
-                    )
                     self.remove_listener()
                     self.qtile.stop()
                     return
```

jwijenbergh:
> What about this? This works for me, it simply just checks if its an XCB connection error, and if so it logs on debug. We have to disable logging after that as otherwise other qtile functions will log exceptions
> 
> ```diff
> diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
> index bc331fdd..5623da50 100644
> --- a/libqtile/backend/x11/core.py
> +++ b/libqtile/backend/x11/core.py
> @@ -324,11 +324,20 @@ class Core(base.Core):
>                  error_code = self.conn.conn.has_error()
>                  if error_code:
>                      error_string = xcbq.XCB_CONN_ERRORS[error_code]
> -                    logger.exception(
> -                        "Shutting down due to X connection error {error_string} ({error_code})".format(
> -                            error_string=error_string, error_code=error_code
> +
> +                    # This is mostly a normal error that happens
> +                    # when we lose connection to the X server
> +                    # just log as debug, an exception confuses users
> +                    if error_string == "XCB_CONN_ERROR":
> +                        logger.debug("Shutting down due to losing connection to the X server")
> +                        # Disable further logging to not spam the log file
> +                        logger.disabled = True
> +                    else:
> +                        logger.exception(
> +                            "Shutting down due to X connection error {error_string} ({error_code})".format(
> +                                error_string=error_string, error_code=error_code
> +                            )
>                          )
> -                    )
>                      self.remove_listener()
>                      self.qtile.stop()
>                      return
> ```

@elParaguayo @m-col sorry for pinging but does this patch look good? If so I can make a PR

elParaguayo:
Thanks for looking at this again.

I thought our main concern with my original PR was that we wanted to find the root cause of the messages. My approach was to suppress messages only when we have initiated the shutdown because, at that point, I don't care about any connection errors because I'm shutting down. However, if I get connection errors when I'm not exiting then something's gone wrong and we shouldn't hide those.

Based on that, I'm not sure what your change does that mine didn't.

jwijenbergh:
> Thanks for looking at this again.
> 
> I thought our main concern with my original PR was that we wanted to find the root cause of the messages. My approach was to suppress messages only when we have initiated the shutdown because, at that point, I don't care about any connection errors because I'm shutting down. However, if I get connection errors when I'm not exiting then something's gone wrong and we shouldn't hide those.
> 
> Based on that, I'm not sure what your change does that mine didn't.

Ah gotcha. Well afaict the original issue (#3551, he used shutdown now, not a qtile shutdown command) was more about the fact that these messages are a bit confusing. When an user shuts down the PC, the x server might close before qtile does which will then spam the log file. My change simply logs it once in debug so that the log file doesn't get cluttered. 

I think when you shutdown qtile (e.g. with a keybind) these errors do not occur in the first place.

PR #3384, opened 2022-02-26
Reduce excessive logging for duplicate widgets

elParaguayo:
Having seen some live examples, the changes in #3010 result in a large number of warning messages in the log.

This PR changes that by collating all the messages into one message.

The message level has also been reduced to "info" as that earlier PR won't have broken anything previous behaviour for users. We are now just providing some additional information.


m-col:
Nice. I do wonder if we even need to inform users of *which* widgets were renamed. Maybe it'd be sufficient to just log that 1 or more widgets were renamed. Then we don't have to maintain a list. IDK maybe that's a needless change.

elParaguayo:
I'd be inclined to keep it as it might be helpful for some debugging later on.

libqtile/bar.py
```diff
@@ -309,6 +310,14 @@ def _configure(self, qtile, screen, reconfigure=False):
...
                     qtile.register_widget(i)
 
+        # Alert the user that we've renamed some widgets
+        if self.qtile.renamed_widgets:
+            logger.info(
+                "The following widgets were renamed in qtile.widgets_map: %s "
+                "To bind commands, rename the widget or use lazy.widget[new_name].",
+                ", ".join(self.qtile.renamed_widgets),
+            )
+
```
m-col, line 320: Could we add a `self.qtile.renamed_widgets.clear()` here?

elParaguayo: Ha. You beat me to it. I was just adding it!

PR #3387, opened 2022-02-26
Re-enable issue templates

m-col:
Github has updated its template feature which supports disabling blank
templates, and adding additional links to the template selection page.
This re-enables the template, and adds links to the template page to
direct feature requests and general questions to more appropriate
channels.

--

It might be worth discussing where we enable blank templates. I've disabled them here. Is there any reason we should allow them?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,17 +1,11 @@
-name: "Issue template"
-description: Required
+name: "Report an issue"
+description: "Our tracker for bugs and other issues with the core project"
 labels: ["unconfirmed"]
 body:
   - type: markdown
     attributes:
      value: |
        Please note that this tracker is only for bugs or other issues with the core project.
-
-       Have a **general question**?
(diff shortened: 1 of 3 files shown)


m-col:
Here is an example of what these links looks like: https://github.com/bchavez/RethinkDb.Driver/issues/new/choose

elParaguayo:
I'd prefer not to have blank templates. I can't think of a good use case. If it's a bug, use the template, anything else, go to Discussions.

m-col:
OK, fair enough!

m-col:
I've merged to get the template and selection menu live. Of course any further comments feedback now that it's up are very welcome.

m-col:
Looks like they aren't available. Maybe something is up with the yaml? Setting up the templates through github's UI generates a markdown template: https://github.com/m-col/qtile/pull/2/files

Maybe we should switch to markdown.

elParaguayo:
I think it's meant to be `config.yml` not `config.yaml`. Don't know if that makes a difference.

m-col:
OK, got it fixed: https://github.com/qtile/qtile/issues/new/choose

What are we thinking now that we've got it up? Good? Bad?

elParaguayo:
Looks really good. I like it!

m-col:
Great! Me too. We can always adjust things as needed if there are problems.

PR #3391, opened 2022-02-27
Reintroduce Issue templates

ura43:
I found you had an old and broken issue template lying around and spent a few minutes making it compatible again.

Markdown templates are the default by github and generally more readable and harder to break.
Unless you have a strong preference for yml, I urge you to keep it with that format.
As for the contents, I tried to stay true to the original while doing some formatting.

Some additional things you may want to consider:

- A wayland checkbox under the metadata section
- If applicable, version numbers of important dependencies that can break things
- (semi-related) Track feature requests via issues. Discussions have terrible visibility and UX.
- (semi-related) Use more labels. They really help a lot with maintenance. Look at [qutebrowser](https://github.com/qutebrowser/qutebrowser/issues) and [ani-cli](https://github.com/pystardust/ani-cli/issues) for reference


ura43:
> Thanks. Guess you didn't see #3387...

yup, sry

m-col:
Nice, great timing as the other was only submitted yesterday :)

Just to address your comments:

> A wayland checkbox under the metadata section
> If applicable, version numbers of important dependencies that can break things

These are useful certainly, mostly with wayland users (the x11 deps never change) but fortunately current wayland users are aware of the 'experimental' state of the wayland backend and usually provide this information. Of course, this won't always be the case, so we should revisit adding these after we've had to ask a couple times.

> (semi-related) Track feature requests via issues. Discussions have terrible visibility and UX.

We used to do this and IMO it was a great thing moving them to discussions. The main devs are more focussed on fixing issues and maintaining things than on implementing new features, and having the features mixed with the issues makes this task a bit more difficult. We've been at around 150 open issues for some time, and many of those are old feature requests that have been open for years (some we haven't migrated to discussions because it disrupts the comment thread). The problem is "discussions have terrible visibility and UX", so we can fix that by adding links to the template selection UI. This is done in #3387 so we can see how that pans out.

> (semi-related) Use more labels. They really help a lot with maintenance. Look at [qutebrowser](https://github.com/qutebrowser/qutebrowser/issues) and [ani-cli](https://github.com/pystardust/ani-cli/issues) for reference

Yeah this sounds like a good idea. We have a bunch defined but (speaking for myself) I only use a small number of them - the ones that I would use to indicate to myself to work on. Maybe we should remove any that are unused and add some more that might make things more easy to handle.

PR #3394, opened 2022-02-28
Ability to scale battery icon

JoaoDDuarte:
Very simple change to allow the user to scale the battery icon, which is especially useful when using custom icons or if you simply want a smaller icon.

This is achieved by introducing a new argument to BatteryIcon, `scale`, whose functionality is basically copied from `CurrentLayoutIcon`.

To be honest, this behaviour can be applied to other widgets, such as Image.


m-col:
Cool, thank you for this. Could you perhaps add a line to the changelog, so that others might see it who would use it :)

PR #3409, opened 2022-03-06
Another go at implementing z-layers for X11

### LINKED ISSUE #753: Z axis

### LINKED ISSUE #1145: Tint2 is covered by other apps

### LINKED ISSUE #1260: Always on top floating windows

### LINKED ISSUE #1403: Feature request: window.send_to_back
Inversely to `window.bring_to_front`, I'd like a command to send a window to the back, be it floating or not.

The use-case is that sometimes a floating window is hidden behind another window. I'd like to be able to send a particular window to the back.

Currently I'm using this workaround from another issue:

```python
@lazy.function
def float_to_front(qtile):
    logging.info("bring floating windows to front")
    for window in qtile.current_group.windows:
        if window.floating:
            window.cmd_bring_to_front()
```

...but it could still happen that one of the floating window is still hidden behind another one right? Being a tiling window manager, I'd like Qtile to allow to use a keyboard shortcut to make the desired window visible by sending the one on top of it to the back (or one layer below) instead of using the mouse to drag the offensive window somewhere else.

My problem arises from the fact that I'm not sure to understand how floating windows work in the current layout/group: when trying to move focus onto the next window, the floating ones never get focused. Could someone explain to me the behavior of moving focus to next window in a group/layout with both tiled and floating window? Because if I could just bring floating windows on top by focusing them I would not need to sned other windows to the back.

Anyway, I think the implementation for `send_to_back` would be something like

```python
    def cmd_send_to_back(self):
        self.window.configure(stackmode=StackMode.Below)
```

...but maybe I'm missing some cases here.

elParaguayo:
This is based on #2194 (without which this would have taken much, much longer).

The code here is not necessarily as neat as that other PR earlier but I've tried to make it a bit easier to follow.

The changes I've made here implement the following rules:
 - when pinning a layer by `cmd_keep_above/below` the window will be added to the top (for `above`) or bottom (for `below`) of that layer
 - `cmd_move_above/below` moves the window above or below the next client in that layer. e.g. if you have 3 windows in the same 'layer', if you do `cmd_move_above` on the bottom window, it will then appear between the other two windows.
 - Add an option to default config for floating windows to be automatically be pinned to the above layer (pretty sure it's not working yet)

Issues to be fixed:
- [x] Moving up/down seems to include an extra layer (i.e. needs more calls than necessary)
- [ ] When a window is `kept_above`, clicking on another window puts it on top of the `kept_above` window.
- [ ] ... there will be more as we test this

Things to do:
 - [x] more testing
 - [x] do we implement `cmd_to_top/bottom` to prevent need for multiple `move_above/below` calls (and rename the latter to `move_up/down`?
 - [x] fix default option to pin new floating windows
 - [ ] Wayland implementation?
 - [x] Fix CI failures (because they're going to happen!)
 
Known issues:
- [x] Child windows not spawning above parent (e.g. gimp export options)

Would be good if some people can test this and provide input. Would also like to hear suggestions for other changes.

Closes #753
Closes #1145
Closes #1260
Closes #1403

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -53,6 +53,12 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Add ability to set custom "Undefined" status key value to `Mpd2Widget`.
         - `Mpd2Widget` now searches for artist name in all similar keys (i.e `albumartist`, `performer`, etc.).
         - Add svg support to `CustomLayoutIcon`
+        - added layering controls for X11 (Wayland support coming soon!):
+          - `lazy.window.keep_above()/keep_below()` marks windows to be kept above/below other windows permanently.
+             Calling the functions with no arguments toggles the state, otherwise pass `enable=True` or `enable=False`.
+          - `lazy.window.move_up()/move_down()` moves windows up and down the z axis.
+          - added `only_focused` setting to Max layout, allowing to draw multiple clients on top of each other when
+            set to False
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
--- a/docs/manual/config/lazy.rst
+++ b/docs/manual/config/lazy.rst
@@ -124,6 +124,23 @@ Window functions
       - Put the focused window to/from floating mode
     * - ``lazy.window.toggle_fullscreen()``
       - Put the focused window to/from fullscreen mode
+    * - ``lazy.window.move_up()``
+      - Move the window above the next window in the stack.
+    * - ``lazy.window.move_down()``
+      - Move the window below the previous window in the stack.
+    * - ``lazy.window.move_to_top()``
+      - Move the window above all other windows with similar priority
+        (i.e. a "normal" window will not be moved above a ``kept_above`` window).
+    * - ``lazy.window.move_to_bottom()``
+      - Move the window below all other windows with similar priority
+        (i.e. a "normal" window will not be moved below a ``kept_below`` window).
+    * - ``lazy.window.keep_above()``
+      - Keep window above other windows.
+    * - ``lazy.window.keep_below()``
+      - Keep window below other windows.
+    * - ``lazy.window.bring_to_front()``
+      - Bring window above all other windows. Ignores ``kept_above`` priority.
+
 
 Screen functions
 ----------------
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -95,9 +95,6 @@ class Core(CommandObject, metaclass=ABCMeta):
     def warp_pointer(self, x: int, y: int) -> None:
         """Warp the pointer to the given coordinates relative."""
 
-    def update_client_list(self, windows_map: dict[int, WindowType]) -> None:
-        """Update the list of windows being managed"""
-
     @contextlib.contextmanager
     def masked(self):
         """A context manager to suppress window events while operating on many windows."""
@@ -252,6 +249,60 @@ class _Window(CommandObject, metaclass=ABCMeta):
 
         """
         return {}
+
+    @expose_command()
+    def keep_above(self, enable: bool | None = None):
+        """Keep this window above all others"""
+
+    @expose_command()
+    def keep_below(self, enable: bool | None = None):
+        """Keep this window below all others"""
+
+    @expose_command()
+    def move_up(self, force: bool = False) -> None:
+        """
+        Move this window above the next window along the z axis.
+
+        Will not raise a "normal" window (i.e. one that is not "kept_above/below")
+        above a window that is marked as "kept_above".
+
+        Will not raise a window where "keep_below" is True unless
+        force is set to True.
+        """
+
+    @expose_command()
+    def move_down(self, force: bool = False) -> None:
+        """
+        Move this window below the previous window along the z axis.
+
+        Will not lower a "normal" window (i.e. one that is not "kept_above/below")
+        below a window that is marked as "kept_below".
+
+        Will not lower a window where "keep_above" is True unless
+        force is set to True.
+        """
+
+    @expose_command()
+    def move_to_top(self) -> None:
+        """
+        Move this window above all windows in the current layer
+        e.g. if you have 3 windows all with "keep_above" set, calling
+        this method will move the window to the top of those three windows.
+
+        Calling this on a "normal" window will not raise it above a "kept_above"
+        window.
+        """
+
+    @expose_command()
+    def move_to_bottom(self) -> None:
+        """
+        Move this window below all windows in the current layer
+        e.g. if you have 3 windows all with "keep_above" set, calling
+        this method will move the window to the bottom of those three windows.
+
+        Calling this on a "normal" window will not raise it below a "kept_below"
+        window.
+        """
 
 
 class Window(_Window, metaclass=ABCMeta):
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -32,7 +32,7 @@ import xcffib
 import xcffib.render
 import xcffib.xproto
 import xcffib.xtest
-from xcffib.xproto import EventMask, StackMode
+from xcffib.xproto import EventMask
 
 from libqtile import config, hook, utils
 from libqtile.backend import base
@@ -167,6 +167,8 @@ class Core(base.Core):
         # The last motion notify event that we still need to handle
         self._motion_notify: xcffib.Event | None = None
 
+        self.last_focused: base.Window | None = None
+
     @property
     def name(self):
         return "x11"
@@ -232,6 +234,9 @@ class Core(base.Core):
         """Assign windows to groups"""
         assert self.qtile is not None
 
+        # Ensure that properties are initialised at startup
+        self.update_client_lists()
+
         if not initial:
             # We are just reloading config
             for win in self.qtile.windows_map.values():
@@ -240,8 +245,8 @@ class Core(base.Core):
             return
 
         # Qtile just started - scan for clients
-        _, _, children = self._root.query_tree()
-        for item in children:
+        for wid in self._root.query_tree():
+            item = window.XWindow(self.conn, wid)
             try:
                 attrs = item.get_attributes()
                 state = item.get_wm_state()
@@ -275,6 +280,9 @@ class Core(base.Core):
                     continue
 
             self.qtile.manage(win)
+
+            self.update_client_lists()
+            win.change_layer()
 
     def warp_pointer(self, x, y):
         self._root.warp_pointer(x, y)
@@ -441,17 +449,28 @@ class Core(base.Core):
         """The name of the connected display"""
         return self._display_name
 
-    def update_client_list(self, windows_map: dict[int, base.WindowType]) -> None:
-        """Updates the client stack list
+    def update_client_lists(self) -> None:
+        """Updates the _NET_CLIENT_LIST and _NET_CLIENT_LIST_STACKING properties
 
         This is needed for third party tasklists and drag and drop of tabs in
         chrome
         """
+        assert self.qtile
         # Regular top-level managed windows, i.e. excluding Static, Internal and Systray Icons
-        wids = [wid for wid, c in windows_map.items() if isinstance(c, window.Window)]
+        wids = [wid for wid, c in self.qtile.windows_map.items() if isinstance(c, window.Window)]
         self._root.set_property("_NET_CLIENT_LIST", wids)
-        # TODO: check stack order
-        self._root.set_property("_NET_CLIENT_LIST_STACKING", wids)
+
+        # We rely on the stacking order from the X server
+        stacked_wids = []
+        for wid in self._root.query_tree():
+            win = self.qtile.windows_map.get(wid)
+            if not win:
+                continue
+
+            if isinstance(win, window.Window) and win.group:
+                stacked_wids.append(wid)
+
+        self._root.set_property("_NET_CLIENT_LIST_STACKING", stacked_wids)
 
     def update_desktops(self, groups, index: int) -> None:
         """Set the current desktops of the window manager
@@ -738,11 +757,14 @@ class Core(base.Core):
             if not win.group or not win.group.screen:
                 return
             win.unhide()
+            self.update_client_lists()
+            win.change_layer()
 
     def handle_DestroyNotify(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
 
         self.qtile.unmanage(event.window)
+        self.update_client_lists()
         if self.qtile.current_window is None:
             self.conn.fixup_focus()
 
@@ -769,6 +791,7 @@ class Core(base.Core):
                 win.wid, win.window.conn.atoms["_NET_WM_DESKTOP"]
             )
         self.qtile.unmanage(event.window)
+        self.update_client_lists()
         if self.qtile.current_window is None:
             self.conn.fixup_focus()
 
@@ -837,9 +860,7 @@ class Core(base.Core):
                 qtile.config.bring_front_click != "floating_only"
                 or getattr(window, "floating", False)
             ):
-                self.conn.conn.core.ConfigureWindow(
-                    window.wid, xcffib.xproto.ConfigWindow.StackMode, [StackMode.Above]
-                )
+                window.bring_to_front()
 
             try:
                 if window.group.screen is not qtile.current_screen:
@@ -908,3 +929,13 @@ class Core(base.Core):
     def keysym_from_name(self, name: str) -> int:
         """Get the keysym for a key from its name"""
         return keysyms[name.lower()]
+
+    def check_stacking(self, win: base.Window) -> None:
+        """Triggers restacking if a fullscreen window loses focus."""
+        if win is self.last_focused:
+            return
+
+        if self.last_focused and self.last_focused.fullscreen:
+            self.last_focused.change_layer()
+
+        self.last_focused = win
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -10,15 +10,16 @@ from typing import TYPE_CHECKING
 import xcffib
 import xcffib.xproto
 from xcffib.wrappers import GContextID, PixmapID
-from xcffib.xproto import EventMask, SetMode, StackMode
+from xcffib.xproto import EventMask, SetMode
 
-from libqtile import hook, utils
+from libqtile import bar, hook, utils
 from libqtile.backend import base
 from libqtile.backend.base import FloatStates
 from libqtile.backend.x11 import xcbq
 from libqtile.backend.x11.drawer import Drawer
 from libqtile.command.base import CommandError, expose_command
 from libqtile.log_utils import logger
+from libqtile.scratchpad import ScratchPad
 
 if TYPE_CHECKING:
     from libqtile.command.base import ItemT
@@ -412,14 +413,7 @@ class XWindow:
         return self.conn.conn.core.GetWindowAttributes(self.wid).reply()
 
     def query_tree(self):
-        q = self.conn.conn.core.QueryTree(self.wid).reply()
-        root = None
-        parent = None
-        if q.root:
-            root = XWindow(self.conn, q.root)
-        if q.parent:
-            parent = XWindow(self.conn, q.parent)
-        return root, parent, [XWindow(self.conn, i) for i in q.children]
+        return self.conn.conn.core.QueryTree(self.wid).reply().children
 
     def paint_borders(self, depth, colors, borderwidth, width, height):
         """
@@ -506,6 +500,13 @@ class _Window:
         self.float_y: int | None = None
         self._float_width: int = self._width
         self._float_height: int = self._height
+
+        # We use `previous_layer` to see if a window has moved up or down a "layer"
+        # The layers are defined in the spec:
+        # https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
+        # We assume a window starts off in the layer for "normal" windows, i.e. ones that
+        # don't match the requirements to be in any of the other layers.
+        self.previous_layer = (False, False, True, False, False, False)
 
         self.bordercolor = None
         self.state = NormalState
@@ -662,7 +663,10 @@ class _Window:
             self._reconfigure_floating(new_float_state=FloatStates.FULLSCREEN)
 
         for s in triggered:
-            setattr(self, s, (s in state))
+            attr = s
+            val = s in state
+            if getattr(self, attr) != val:
+                setattr(self, attr, val)
 
     @property
     def urgent(self):
@@ -903,20 +907,255 @@ class _Window:
         self.width = width
         self.height = height
 
-        kwarg = dict(
-            x=x,
-            y=y,
-            width=width,
-            height=height,
-        )
-        if above:
-            kwarg["stackmode"] = StackMode.Above
+        self.window.configure(x=x, y=y, width=width, height=height)
 
-        self.window.configure(**kwarg)
+        if above:
+            self.change_layer(up=True)
+
         self.paint_borders(bordercolor, borderwidth)
 
         if send_notify:
             self.send_configure_notify(x, y, width, height)
+
+    def get_layering_information(self) -> tuple[bool, bool, bool, bool, bool, bool]:
+        """
+        Get layer-related EMWH-flags
+        https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
+
+        Copied here:
+
+         To obtain good interoperability between different Desktop Environments,
+         the following layered stacking order is recommended, from the bottom:
+
+            - windows of type _NET_WM_TYPE_DESKTOP
+            - windows having state _NET_WM_STATE_BELOW
+            - windows not belonging in any other layer
+            - windows of type _NET_WM_TYPE_DOCK (unless they have state
+              _NET_WM_TYPE_BELOW) and windows having state _NET_WM_STATE_ABOVE
+            - focused windows having state _NET_WM_STATE_FULLSCREEN
+
+        Windows that are transient for another window should be kept above this window.
+
+        The window manager may choose to put some windows in different stacking positions,
+        for example to allow the user to bring currently a active window to the top and return
+        it back when the window loses focus. To this end, qtile adds an additional layer so that
+        scratchpad windows are placed above all others, always.
+        """
+        state = self.window.get_net_wm_state()
+        _type = self.window.get_wm_type() or ""
+
+        # Check if this window is focused
+        active_window = self.qtile.core._root.get_property(
+            "_NET_ACTIVE_WINDOW", "WINDOW", unpack=int
+        )
+        if active_window and active_window[0] == self.window.wid:
+            focus = True
+        else:
+            focus = False
+
+        desktop = _type == "desktop"
+        below = "_NET_WM_STATE_BELOW" in state
+        dock = _type == "dock"
+        above = "_NET_WM_STATE_ABOVE" in state
+        full = (
+            "fullscreen" in state
+        )  # get_net_wm_state translates this state so we don't use _NET_WM name
+        is_scratchpad = isinstance(self.qtile.groups_map.get(self.group), ScratchPad)
+
+        # sort the flags from bottom to top, True meaning further below than False at each step
+        states = [desktop, below, above or (dock and not below), full and focus, is_scratchpad]
+        other = not any(states)
+        states.insert(2, other)
+
+        # If we're a desktop, this should always be the lowest layer...
+        if desktop:
+            # mypy can't work out that this gives us tuple[bool, bool, bool, bool, bool, bool]...
+            # (True, False, False, False, False, False)
+            return tuple(not i for i in range(6))  # type: ignore
+
+        # ...otherwise, we set to the highest matching layer.
+        # Look for the highest matching level and then set all other levels to False
+        highest = max(i for i, state in enumerate(states) if state)
+
+        # mypy can't work out that this gives us tuple[bool, bool, bool, bool, bool, bool]...
+        return tuple(i == highest for i in range(6))  # type: ignore
+
+    def change_layer(self, up=True, top_bottom=False):
+        """Raise a window above its peers or move it below them, depending on 'up'.
+        Raising a normal window will not lift it above pinned windows etc.
+
+        There are a few important things to take note of when relaying windows:
+        1. If a window has a defined parent, it should not be moved underneath it.
+           In case children are blocking, this could leave an application in an unusable state.
+        2. If a window has children, they should be moved along with it.
+        3. If a window has a defined parent, either move the parent or do nothing at all.
+        4. EMWH-flags follow strict layering rules:
+           https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
+        """
+        if len(self.qtile.windows_map) < 2:
+            return
+
+        if self.group is None:
+            return
+
+        parent = self.window.get_wm_transient_for()
+        if parent is not None and not up:
+            return
+
+        layering = self.get_layering_information()
+
+        # Comparison of layer states: -1 if window is now in a lower state group,
+        # 0 if it's in the same group and 1 if it's in a higher group
+        moved = (self.previous_layer > layering) - (layering > self.previous_layer)
+        self.previous_layer = layering
+
+        stack = list(self.qtile.core._root.query_tree())
+        if self.wid not in stack or len(stack) < 2:
+            return
+
+        group_windows = self.group.windows
+        if self.group.screen is not None:
+            group_bars = [gap for gap in self.group.screen.gaps if isinstance(gap, bar.Bar)]
+        else:
+            group_bars = []
+
+        # Get list of windows that are in the stack and managed by qtile
+        # List of tuples (XWindow object, transient_for, layering_information)
+        windows = list(
+            map(
+                lambda w: (
+                    w.window,
+                    w.window.get_wm_transient_for(),
+                    w.get_layering_information(),
+                ),
+                group_windows,
+            )
+        )
+
+        # Sort this list to match stacking order reported by server
+        windows.sort(key=lambda w: stack.index(w[0].wid))
+
+        # Get lists of windows on lower, higher or same "layer" as window
+        lower = [w[0].wid for w in windows if w[1] is None and w[2] > layering]
+        higher = [w[0].wid for w in windows if w[1] is None and w[2] < layering]
+        same = [w[0].wid for w in windows if w[1] is None and w[2] == layering]
+
+        # We now need to identify the new position in the stack
+
+        # If the window has a parent, the window should just be put above it
+        if parent:
+            sibling = parent
+            above = True
+
+        # Now we just check whether the window has changed layer.
+
+        # If we're forcing to top or bottom of current layer...
+        elif top_bottom:
+            if up:
+                sibling = same[-1]
+                above = True
+            else:
+                sibling = same[0]
+                above = False
+
+        # There are no windows in the desired layer (should never happen) or
+        # we've moved to a new layer and are the only window in that layer
+        elif not same or (len(same) == 1 and moved != 0):
+            # Try to put it above the last window in the lower layers
+            if lower:
+                sibling = lower[-1]
+                above = True
+
+            # Or below the first window in the higher layers
+            elif higher:
+                sibling = higher[0]
+                above = False
+
+            # Don't think we should end up here but, if we do...
+            else:
+                # Put the window above the highest window if we're raising it
+                if up:
+                    sibling = stack[-1]
+                    above = True
+
+                # or below the lowest window if we're lowering the window
+                else:
+                    sibling = stack[0]
+                    above = False
+
+        else:
+            # Window has moved to a lower layer state
+            if moved < 0:
+                if self.kept_below:
+                    sibling = same[0]
+                    above = False
+                else:
+                    sibling = same[-1]
+                    above = True
+
+            # Window is in same layer state
+            elif moved == 0:
+                try:
+                    pos = same.index(self.wid)
+                except ValueError:
+                    pos = len(same) if up else 0
+                if not up:
+                    pos = max(0, pos - 1)
+                else:
+                    pos = min(pos + 1, len(same) - 1)
+                sibling = same[pos]
+                above = up
+
+            # Window is in a higher layer
+            else:
+                if self.kept_above:
+                    sibling = same[-1]
+                    above = True
+                else:
+                    sibling = same[0]
+                    above = False
+
+        # If the sibling is the current window then we just check if any windows in lower/higher layers are
+        # stacked incorrectly and, if so, restack them. However, we don't need to configure stacking for this
+        # window
+        if sibling == self.wid:
+            index = stack.index(self.wid)
+
+            # We need to make sure the bars are included so add them now
+            if group_bars:
+                for group_bar in group_bars:
+                    bar_layer = group_bar.window.get_layering_information()
+                    if bar_layer > layering:
+                        lower.append(group_bar.window.wid)
+                    elif bar_layer < layering:
+                        higher.append(group_bar.window.wid)
+
+                # Sort the list to match the server's stacking order
+                lower.sort(key=lambda wid: stack.index(wid))
+                higher.sort(key=lambda wid: stack.index(wid))
+
+            for wid in [w for w in lower if stack.index(w) > index]:
+                self.qtile.windows_map[wid].window.configure(
+                    stackmode=xcffib.xproto.StackMode.Below, sibling=same[0]
+                )
+
+            # We reverse higher as each window will be placed above the last item in the current layer
+            # this means the last item we stack will be just above the current layer.
+            for wid in [w for w in higher[::-1] if stack.index(w) < index]:
+                self.qtile.windows_map[wid].window.configure(
+                    stackmode=xcffib.xproto.StackMode.Above, sibling=same[-1]
+                )
+
+            return
+
+        # Window needs new stacking info. We tell the server to stack the window
+        # above or below a given "sibling"
+        self.window.configure(
+            stackmode=xcffib.xproto.StackMode.Above if above else xcffib.xproto.StackMode.Below,
+            sibling=sibling,
+        )
+        # TODO: also move our children if we were moved upwards
+        self.qtile.core.update_client_lists()
 
     def paint_borders(self, color, width):
         self.borderwidth = width
@@ -1004,9 +1243,6 @@ class _Window:
         did_focus = self._do_focus()
         if not did_focus:
             return
-        if isinstance(self, base.Internal):
-            # self._do_focus is enough for internal windows
-            return
 
         # now, do all the other WM stuff since the focus actually changed
         if warp and self.qtile.config.cursor_warp:
@@ -1041,6 +1277,15 @@ class _Window:
 
         if self.group:
             self.group.current_window = self
+
+        # See https://github.com/qtile/qtile/pull/3409#discussion_r1117952134 for discussion
+        # on mypy error here
+        if self.fullscreen and not self.previous_layer[4]:  # type: ignore
+            self.change_layer()
+
+        # Check if we need to restack a previously focused fullscreen window
+        self.qtile.core.check_stacking(self)
+
         hook.fire("client_focus", self)
 
     @expose_command()
@@ -1101,6 +1346,98 @@ class _Window:
             state=state,
             float_info=float_info,
         )
+
+    @expose_command()
+    def keep_above(self, enable: bool | None = None):
+        if enable is None:
+            self.kept_above = not self.kept_above
+        else:
+            self.kept_above = enable
+
+        self.change_layer(top_bottom=True, up=True)
+
+    @expose_command()
+    def keep_below(self, enable: bool | None = None):
+        if enable is None:
+            self.kept_below = not self.kept_below
+        else:
+            self.kept_below = enable
+
+        self.change_layer(top_bottom=True, up=False)
+
+    @expose_command()
+    def move_up(self, force=False):
+        if self.kept_below and force:
+            self.kept_below = False
+        with self.qtile.core.masked():
+            # Disable masks so that moving windows along the Z axis doesn't trigger
+            # focus change events (i.e. due to `follow_mouse_focus`)
+            self.change_layer()
+
+    @expose_command()
+    def move_down(self, force=False):
+        if self.kept_above and force:
+            self.kept_above = False
+        with self.qtile.core.masked():
+            self.change_layer(up=False)
+
+    @expose_command()
+    def move_to_top(self, force=False):
+        if self.kept_below and force:
+            self.kept_below = False
+        with self.qtile.core.masked():
+            self.change_layer(top_bottom=True)
+
+    @expose_command()
+    def move_to_bottom(self, force=False):
+        if self.kept_above and force:
+            self.kept_above = False
+        with self.qtile.core.masked():
+            self.change_layer(up=False, top_bottom=True)
+
+    @property
+    def kept_above(self):
+        reply = list(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_ABOVE"]
+        return atom in reply
+
+    @kept_above.setter
+    def kept_above(self, value):
+        reply = list(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_ABOVE"]
+        if value and atom not in reply:
+            reply.append(atom)
+        elif not value and atom in reply:
+            reply.remove(atom)
+        else:
+            return
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_BELOW"]
+        if atom in reply:
+            reply.remove(atom)
+        self.window.set_property("_NET_WM_STATE", reply)
+        self.change_layer()
+
+    @property
+    def kept_below(self):
+        reply = list(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_BELOW"]
+        return atom in reply
+
+    @kept_below.setter
+    def kept_below(self, value):
+        reply = list(self.window.get_property("_NET_WM_STATE", "ATOM", unpack=int))
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_BELOW"]
+        if value and atom not in reply:
+            reply.append(atom)
+        elif not value and atom in reply:
+            reply.remove(atom)
+        else:
+            return
+        atom = self.qtile.core.conn.atoms["_NET_WM_STATE_ABOVE"]
+        if atom in reply:
+            reply.remove(atom)
+        self.window.set_property("_NET_WM_STATE", reply)
+        self.change_layer(up=False)
 
 
 class Internal(_Window, base.Internal):
@@ -1170,6 +1507,11 @@ class Internal(_Window, base.Internal):
             height=self.height,
             id=self.window.wid,
         )
+
+    @expose_command()
+    def focus(self, warp: bool = True) -> None:
+        """Focuses the window."""
+        self._do_focus()
 
 
 class Static(_Window, base.Static):
@@ -1270,7 +1612,8 @@ class Static(_Window, base.Static):
 
     @expose_command()
     def bring_to_front(self):
-        self.window.configure(stackmode=StackMode.Above)
+        if self.get_wm_type() != "desktop":
+            self.window.configure(stackmode=xcffib.xproto.StackMode.Above)
 
 
 class Window(_Window, base.Window):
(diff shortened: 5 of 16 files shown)


jwijenbergh:
Great work! I will try this later.

> do we implement cmd_to_top/bottom to prevent need for multiple move_above/below calls (and rename the latter to move_up/down?

Imo yes, it should just be a few lines of code extra right?

elParaguayo:
@m-col @jwijenbergh would appreciate thoughts here.

According to spec, we manage stacking in different layer priorities:
```
To obtain good interoperability between different Desktop Environments, the following layered stacking order is recommended, from the bottom:

- windows of type _NET_WM_TYPE_DESKTOP
- windows having state _NET_WM_STATE_BELOW
- windows not belonging in any other layer
- windows of type _NET_WM_TYPE_DOCK (unless they have state _NET_WM_TYPE_BELOW) and windows having state _NET_WM_STATE_ABOVE
- focused windows having state _NET_WM_STATE_FULLSCREEN
```
Most of our windows will default to being in that third group.

I've implemented `move_up/down` and `move_to_top/bottom` which moves windows inside those respective layers.

Two questions:
1) what should the behaviour of `cmd_bring_to_front()` be? We can use it override the spec and make it the topmost window. If we don't do that, it's just the same as `move_to_top`.
2) scratchpads rely on this. Toggling a scratchpad currently will still place it below any window in those last two layers. We could either make scratchpad windows `kept_above` or, we use the overriden option for `bring_to_front`.

FWIW, my thoughts are:
- I'd like to keep to spec as much as possible but...
- It makes sense to have a `bring_to_front` option which does what it says
- Scratchpads should be `kept_above`

jwijenbergh:
> @m-col @jwijenbergh would appreciate thoughts here.
> 
> According to spec, we manage stacking in different layer priorities:
> 
> ```
> To obtain good interoperability between different Desktop Environments, the following layered stacking order is recommended, from the bottom:
> 
> - windows of type _NET_WM_TYPE_DESKTOP
> - windows having state _NET_WM_STATE_BELOW
> - windows not belonging in any other layer
> - windows of type _NET_WM_TYPE_DOCK (unless they have state _NET_WM_TYPE_BELOW) and windows having state _NET_WM_STATE_ABOVE
> - focused windows having state _NET_WM_STATE_FULLSCREEN
> ```
> 
> Most of our windows will default to being in that third group.
> 
> I've implemented `move_up/down` and `move_to_top/bottom` which moves windows inside those respective layers.
> 
> Two questions:
> 
>     1. what should the behaviour of `cmd_bring_to_front()` be? We can use it override the spec and make it the topmost window. If we don't do that, it's just the same as `move_to_top`.
> 
>     2. scratchpads rely on this. Toggling a scratchpad currently will still place it below any window in those last two layers. We could either make scratchpad windows `kept_above` or, we use the overriden option for `bring_to_front`.
> 
> 
> FWIW, my thoughts are:
> 
>     * I'd like to keep to spec as much as possible but...
> 
>     * It makes sense to have a `bring_to_front` option which does what it says
> 
>     * Scratchpads should be `kept_above`

I just got the chance to test this quickly and it so far works as expected. I haven't played much with the cmd_ functions but I will test them. 

> I'd like to keep to spec as much as possible but...
> It makes sense to have a bring_to_front option which does what it says
> Scratchpads should be kept_above

I agree. Bring to front should 'ignore' the spec. IMO scratchpads should indeed have an option to be kept on top. Sometimes I have a fullscreen window and I want to do something in a terminal really quickly so I open a scratchpad

libqtile/resources/default_config.py
```diff
@@ -160,6 +160,7 @@
 dgroups_app_rules = []  # type: list
 follow_mouse_focus = True
 bring_front_click = False
+floats_kept_above = False
```
jwijenbergh, line 163: should we set this to True by default maybe?

elParaguayo: I didn't do that because it would change current behaviour (which often gets a negative reaction) however, I don't feel particularly strongly about it as I rarely use floating windows!

jwijenbergh: Yeah makes sense. It's really the fact that I've seen multiple people annoyed by the default behaviour. But it's an easy toggle so it doesn't matter that much

elParaguayo:
@m-col the new commands (`cmd_move_up/down` etc) don't have wayland implementations yet. I don't propose including them here but do you see anything problematic with what's been added here in terms of compatibility between backends?

libqtile/core/manager.py
```diff
@@ -173,6 +175,14 @@ def load_config(self, initial=False) -> None:
         if self.config.reconfigure_screens:
             hook.subscribe.screen_change(self.cmd_reconfigure_screens)
 
+        # To ensure compliance with X11's window layering spec
+        # We need to restack fullscreen windows if they lose focus
+        hook.subscribe.client_focus(self.check_stacking)
```
m-col, line 180: focus_change might be more appropriate, as it also fires when focus moves to nothing?

If this is X11-specific, could this be moved into the x11 core? It may not need to be hooked - there might be a good place to put it near the code that actually does the focussing.

I'm also a bit concerned that this may have unexpected side effects (and become difficult to debug). We don't have any core core code (only configurables code) that depends on hooks. Note that cmd_reload_config clears hooks, so currently hooks are a more user-oriented functionality. This hook here would get unsubscribed upon config reload and wouldn't get rehooked.

libqtile/backend/x11/core.py
```diff
@@ -418,11 +418,24 @@ def update_client_list(self, windows_map: dict[int, base.WindowType]) -> None:
...
+        stack = list(self._root.query_tree())
+        wids = [
+            wid
+            for wid, c in self.qtile.windows_map.items()
+            if isinstance(c, window.Window) and c.group and wid in stack
+        ]
+        # We need to keep the order in the tree so pick the items from that list
+        # but only keep the ones filtered out by wids
+        current_stack = [s for s in stack if s in wids]
+        self._root.set_property("_NET_CLIENT_LIST_STACKING", current_stack)
```
m-col, line 438: I think passing `wids` to `set_property` would have the same effect, right? `wids` is filtered by `stack` in the list comprehension, and we ignore those items in stack that aren't also in `self.qtile.windows_map`, so `wids` is already the intersection between `stack` and `self.qtile.windows_map.keys()`. So `current_stack == wids`.

libqtile/backend/x11/window.py
```diff
@@ -892,21 +886,232 @@ def place(
...
             width=width,
             height=height,
         )
         if above:
-            kwarg["stackmode"] = StackMode.Above
-
-        self.window.configure(**kwarg)
+            self.change_layer(kwargs=kwargs)
+        else:
+            self.window.configure(**kwargs)
```
m-col, line 898: There is confusing overlap in the role of `self.change_layer` and `self.window.configure` and I think we should segregate those two things. This would also remove `kwargs` from `change_layer` which AFAICT is an x11-specific detail.
```suggestion
        self.window.configure(
            x=x,
            y=y,
            width=width,
            height=height,
	)
        if above:
            self.change_layer(up=True)
```
```

File: libqtile/backend/x11/window.py [hunk 5]
```diff
@@ -892,21 +886,232 @@ def place(
...
+                lambda w: (
+                    w.window,
+                    w.window.get_wm_transient_for(),
+                    w.get_layering_information(),
+                ),
+                [
+                    self.qtile.windows_map[i]
+                    for i in stack
+                    if i in self.qtile.windows_map
+                    and isinstance(self.qtile.windows_map[i], _Window)
```
m-col, line 994: Aren't all values in `self.qtile.windows_map` going to be instances of `_Window`?

libqtile/backend/x11/window.py
```diff
@@ -1028,6 +1233,9 @@ def focus(self, warp: bool) -> None:
 
         if self.group:
             self.group.current_window = self
+
+        if getattr(self, "fullscreen", False) and not self.previous_layer[4]:
```
m-col, line 1237: Why `getattr(self, "fullscreen", False)` and not `self.fullscreen`? Do we ever have a situation where that attribute does not exist?

libqtile/backend/x11/window.py
```diff
@@ -1315,6 +1523,9 @@ def floating(self, do_float):
             else:
                 # if we are setting floating early, e.g. from a hook, we don't have a screen yet
                 self._float_state = FloatStates.FLOATING
+            if not self.kept_above and self.qtile.config.floats_kept_above:
+                print(f"1503 - Setting keep above: {self.name}")
```
m-col, line 1527: Oops :)

libqtile/backend/x11/window.py
```diff
@@ -1498,13 +1755,20 @@ def _reconfigure_floating(self, new_float_state=FloatStates.FLOATING):
...
         if self._float_state != new_float_state:
             self._float_state = new_float_state
             if self.group:  # may be not, if it's called from hook
                 self.group.mark_floating(self, True)
+            if (
+                new_float_state == FloatStates.FLOATING
+                and not self.kept_above
+                and self.qtile.config.floats_kept_above
+            ):
+                print(f"1742 - Setting keep above: {self.name}")
```
m-col, line 1770: Oops 2 :)

libqtile/core/manager.py
```diff
@@ -634,6 +646,17 @@ def unmanage(self, wid: int) -> None:
                     c.group.remove(c)
             del self.windows_map[wid]
             self.core.update_client_list(self.windows_map)
+            self.core.update_client_stack()
+
+    def check_stacking(self, win):
+        """Triggers restacking if a fullscreen window loses focus."""
```
m-col, line 652: Is this just used for fullscreen windows? It seems to be used more. Could you elaborate on the purpose of this function?

m-col (changes requested):
Looks awesome, we're so close.

I've mostly be focussing my review on the interface and encapsulation. We have to ensure that we aren't letting backend-specific details spill out into the main core. I think that is mostly the case which is great.

We may need to deprecate `cmd_bring_to_front` in favour of the new feature-complete set of commands that deal with Z axis on the user side.

I'll be thinking about how to implement this on the Wayland side, with the aim of providing a commit or 3 to append to this PR for merge.

libqtile/core/manager.py
elParaguayo: It's X11 specific in that there are rules/guides on priority for stacking windows. Any reason why Wayland wouldn't apply similar guides? i.e. presumably you'll still want the ability to keep windows above others and have fullscreen windows on top?

libqtile/backend/x11/window.py
elParaguayo: Yes. Much neater.

libqtile/backend/x11/window.py
elParaguayo: This is why I have a test in `qtile-extras` that fails if it finds print statements in my PRs!!

libqtile/core/manager.py
m-col: You would still want those things but clients are not aware of their relative positions, so there is no informing them of their position, and therefore not really anything to do upon focus change. It's also possible I've misunderstood a bit about what this part is doing because I wasn't so sure what `self.check_stacking` was for. Because the wayland backend is what determines Z axis at every level (by choosing the order to render windows in, and where input events get sent), it doesn't need a hook to respond to its own actions. So it's more the implementation detail that is x11 specific, rather than the concept

docs/manual/config/lazy.rst
```diff
@@ -109,6 +109,23 @@ Window functions
...
+        (i.e. a "normal" window will not be moved above a ``kept_above`` window).
+    * - ``lazy.window.move_to_bottom()``
+      - Move the window below all other windows with similar priority
+        (i.e. a "normal" window will not be moved below a ``kept_below`` window).
+    * - ``lazy.window.keep_above()``
+      - Keep window above other windows.
+    * - ``lazy.window.keep_below()``
+      - Keep window below other windows.
+    * - ``lazy.window.bring_to_front()``
+      - Bring window above all other windows. Ignores ``kept_above`` priority.
```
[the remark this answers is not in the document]
elParaguayo: I think the question we need to consider is whether there'd ever be a scenario where someone wants to call `move_to_top` but still have it appear below things that are tagged as `kept_above`. If there isn't then we don't need the `respectful` arg.

Presumably, if you're moving to top you want to see the window so you want to ignore the `kept_above` tag for this purpose.

However, I think this causes a new issue: as the layer priorities have not been respected, you've now got a window that's out of place according to the spec. When you open a new window, it's going to open above the one with the same layer priority which means it will also open above windows that are `kept_above`. I guess we could solve that by adding additional logic which looks for `kept_above` windows and ensures new windows open below them.

Thoughts?

libqtile/backend/x11/window.py
elParaguayo: That should be right if we've stopped `Systray` icons from being included in the map. They're `XWindow`s which don't inherit from `_Window`.

libqtile/backend/x11/window.py
elParaguayo: I think it must have failed at some point for me to write this. Let me test again!

libqtile/core/manager.py
elParaguayo: Fullscreen windows have top priority for stacking *when focused*. So, if a full screen window loses focus, we need to reposition it in the stack.

libqtile/resources/default_config.py
m-col: I think we should enable it by default. This won't affect existing users because they already have configs. Those that complain are often new users who are frustrated that they can't move focus to windows hidden at the back. This way that doesn't even happen, and when they reach that point in the config (i expect many new users will start qtile with the defaults before ever even opening the config!) they'll see that it can be disabled.

libqtile/backend/x11/core.py
m-col: Did you make this change and not yet push? just wondering if it was in fact resolved

docs/manual/config/lazy.rst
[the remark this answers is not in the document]
m-col: Isn't the point of `kept_above` that you wouldn't want windows to get above it? A primary use-case I'm thinking of is some kind of monitor or video playing in the corner of the screen which you'd always want visible, while you perform otherwise normal operations on your windows including raising them.

Ultimately it's pretty difficult to guess what people would want! I'd maybe err on the side of not having the respectful flag, implementing whichever behaviour is easier (respecting or not), and then seeing what people want after that.

m-col:
This is great. Looks like you've resolved some minor things but I haven't seen those updates. I've replied to a couple of open questions. In general I'm thinking let's not hold this back for the sake of the wayland backend. I can deal with that without this being hindered. Therefore we can just resolve the couple of open discussions above and then merge.

elParaguayo:
I think I need to make this one a priority - need to get it done.

I did resolve those minor comments but haven't pushed yet as there are still other, open items to address

3nprob:
If I get this right, addressing this for WL should solve #3675 / #3697? Considering those, would it make sense to bump doing so?

jwijenbergh:
> If I get this right, addressing this for WL should solve #3675 / #3697? Considering those, would it make sense to bump doing so?

Those issues seem unrelated

elParaguayo:
Right - I think we really need to pick this one up again and get it done!

I'm testing again and will update issues in the first post as I find them!

jwijenbergh:
Re wayland backend: When we eventually implement it using wlr_scene then z-ordering becomes pretty trivial for that

libqtile/group.py
```diff
@@ -256,6 +256,8 @@ def add(self, win, focus=True, force=False):
             win._float_state = FloatStates.FULLSCREEN
         elif self.floating_layout.match(win) and not win.fullscreen:
             win._float_state = FloatStates.FLOATING
+            if not win.kept_above and self.qtile.config.floats_kept_above:
+                win.cmd_keep_above()
```
m-col, line 260: We can change this to `keep_above()` now ;). Might be worth grepping, I saw another use of `cmd_` too.

libqtile/core/manager.py
m-col: Happy to leave this is as. If there is an issue with the implementation it will come up pretty soon as a bug, and we can fix it then

libqtile/backend/x11/window.py
```diff
@@ -1844,10 +2121,38 @@ def disable_fullscreen(self):
...
+            self.kept_above = False
+        self.change_layer(up=False)
+
+    @expose_command()
+    def move_to_top(self, force=False):
+        if self.kept_below and force:
+            self.kept_below = False
+        self.change_layer(top_bottom=True)
+
+    def move_to_bottom(self, force=False):
```
m-col, line 2152: Should this command be exposed?

libqtile/backend/base.py
```diff
@@ -253,6 +256,65 @@ def info(self) -> dict[str, Any]:
...
+        Move this window below all windows in the current layer
+        e.g. if you have 3 windows all with "keep_above" set, calling
+        this method will move the window to the bottom of those three windows.
+
+        Calling this on a "normal" window will not raise it below a "kept_below"
+        window.
+        """
+
+    def change_layer(self, up: bool = True, top_bottom: bool = False, kwargs=None) -> None:
+        """Trigger repositioning of window in layer stack."""
```
m-col, line 316: This appears to be an implementation detail of the X backend. Could we move it to that backend instead of here?

libqtile/backend/base.py
```diff
@@ -253,6 +256,65 @@ def info(self) -> dict[str, Any]:
...
         return {}
 
+    @property
+    def kept_above(self) -> bool:
+        return False
+
+    @kept_above.setter
+    def kept_above(self, value: bool) -> None:
+        pass
+
```
m-col, line 266: Is this something we're happy for users to use directly, perhaps from within layouts? I think it's good to expose this information via a read-only property, if logic like those in layouts can benefit from it, but by having a setter we're providing two ways for users to `keep_above()` a window (or at least, it looks like that).

If we don't want users to use this property's setter, it's probably more of an implementation detail of how the X11 backend handles Z stacking, and we should move it there. It's not clear yet if it would also be used by windows in the wayland backend.

Also, how does one undo a `keep_above()` using a command?

m-col (commented):
Left a couple more comments but one or two of the existing comments aren't resolved yet (though they are all relatively minor)

libqtile/backend/base.py
elParaguayo: I don't have any particular concerns with users being able to access it directly.

Undoing `keep_above()` is done by calling it again. The command just runs `self.kept_above = not self.kept_above`.

libqtile/backend/x11/window.py
elParaguayo: It causes a mypy failure `libqtile/backend/x11/window.py:1257: error: "_Window" has no attribute "fullscreen"`

libqtile/backend/x11/window.py
```diff
@@ -896,21 +890,245 @@ def place(
...
+        other = not any(states)
+        states.insert(2, other)
+
+        # This next bit could have been done by adding more conditions when setting the
+        # states flags but it's easier to look for the highest matching level and then
+        # set all other levels to False
+        highest = max(i for i, state in enumerate(states) if state)
+        win_states = [True if i == highest else False for i in range(5)]
+
+        return (win_states[0], win_states[1], win_states[2], win_states[3], win_states[4])
```
jwijenbergh, line 962:
```suggestion
        return tuple(i == highest for i in range(len(states))
```

libqtile/backend/x11/window.py
```diff
@@ -896,21 +890,245 @@ def place(
...
+                    self.qtile.windows_map[i]
+                    for i in stack
+                    if i in self.qtile.windows_map
+                    and not isinstance(self.qtile.windows_map[i], Internal)
+                    and self.qtile.windows_map[i].group == self.group
+                ],
+            )
+        )
+
+        layer_windows = [w[0].wid for w in windows if w[1] is None and w[2] == layering]
```
jwijenbergh, line 1017: layer_windows is the same as "same"? Can one of them be removed?

```suggestion
        layer_windows, lower, higher, same = []
        for w in windows:
            wid = w[0].wid    
            if w[1] is not None:
                continue
            if w[2] == layering:
                layer_windows.append(wid)
                same.append(wid)
            elif w[2] > layering:
                lower.append(wid)
            elif w[2] < layering:
                higher.append(wid)
```

libqtile/backend/x11/window.py
jwijenbergh: Still needs to be fixed :)

libqtile/backend/x11/window.py
jwijenbergh: Still there :)

libqtile/backend/base.py
m-col: In that case I think we should move this property into the X11 backend's Window implementation, as an implementation detail of how it manages Z stacking, and ideally as a private method (if it's only used by that class). I just want to avoid this becoming part of the 'user API' when we have `keep_above()`.

libqtile/backend/x11/window.py
m-col: > This is why I have a test in qtile-extras that fails if it finds print statements in my PRs!!

Fancy copy-pasting it into this repo? ;)

libqtile/backend/x11/window.py
m-col: I see. I guess what we should do is have `focus()` methods defined on each of the window classes, rather than one big one with if statements checking which class self is. Happy to type:ignore it for now, we can improve things later

libqtile/backend/base.py
elParaguayo: Ok. We also have things like `check_stacking` in `manager.py` and `update_client_list/stack` in `base/backend.py`. Does that all need to get moved to the x11 core?

`change_layer` is also called from `manage` in `manager.py` so, if we move it, we need to rethink how it's triggered when windows are managed.

libqtile/group.py
```diff
@@ -256,6 +256,8 @@ def add(self, win, focus=True, force=False):
             win._float_state = FloatStates.FULLSCREEN
         elif self.floating_layout.match(win) and not win.fullscreen:
             win._float_state = FloatStates.FLOATING
+            if not win.kept_above and self.qtile.config.floats_kept_above:
+                win.keep_above()
```
m-col, line 260:
```suggestion
            if self.qtile.config.floats_kept_above:
                win.keep_above(enable=True)
```

libqtile/backend/base.py
```diff
@@ -253,6 +256,57 @@ def info(self) -> dict[str, Any]:
...
         return {}
 
+    @expose_command()
+    def keep_above(self) -> None:
+        """Keep this window above all others"""
+
+    @expose_command()
+    def keep_below(self) -> None:
+        """Keep this window below all others"""
+
```
m-col, line 266: I saw that `group.py` checks `win.kept_above`, requiring the property. Perhaps this would work instead?
```suggestion
    @expose_command()
    def keep_above(self, enable: bool | None = None) -> None:
        """Keep this window above all others"""

    @expose_command()
    def keep_below(self, enable: bool | None = None) -> None:
        """Keep this window below all others"""
```
By default, `enable = None` can mean 'toggle', otherwise it's specifying whether to keep above/below or not.

libqtile/core/manager.py
```diff
@@ -179,6 +181,15 @@ def load_config(self, initial: bool = False) -> None:
         if "resume" in hook.subscriptions:
             subscribe_for_resume_events()
 
+        # To ensure compliance with X11's window layering spec
+        # We need to restack fullscreen windows if they lose focus
+        if self.core.name == "x11":
+            hook.subscribe.client_focus(self.check_stacking)
+
```
m-col, line 188: In line 167 just above we inform the core that the config is being loaded, so could we move this block into `Core.on_config_loaded` in the X backend?

Similarly, as `self.check_stacking` is doing something for the benefit of the X window layer spec, we can move that to the Core too. This hides the X11-specific `change_layer` window method inthe backend, so we can drop that from the base Window class.

libqtile/core/manager.py
```diff
@@ -660,6 +671,8 @@ def manage(self, win: base.WindowType) -> None:
             if not win.group and self.current_screen.group:
                 self.current_screen.group.add(win, focus=win.can_steal_focus)
         self.core.update_client_list(self.windows_map)
+        self.core.update_client_stack()
```
m-col, line 674: I'm thinking it might be good to move these into the X Core as well, and it seems like that would be straight forward, and I *think* without unexpected side effects.

In this case, `Qtile.manage()` is called by the Core, and we can move these two `upate_client_X` calls to after `manage()` returns to the Core. Functionally the only difference is that the client_managed hook is moved to before the client list and stack are updated. I suspect that might be fine?

Then only other place these functions are used is the inverse: in `Qtile.unmanage()`. In that case we could either move it to after the call returns, moving the client_killed hook to before the client list/stack update, which might be fine. Or alternatively the Core can copy `self.windows_map`, `del self.windows_map[wid]` from the copy, and update the list/stick before calling `Qtile.unmanage`. That one might be safer because who knows what users are doing in client_killed hooks.

If we make these changes we can also drop `update_client_list` and `update_client_stack` from the base `Core` class.

m-col (changes requested):

m-col:
One of the changes I mentioned fixes the test failure you mentioned on discord. This is the diff:
```diff
diff --git a/libqtile/backend/base.py b/libqtile/backend/base.py
index abdc02fc..9f396306 100644
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -257,11 +257,11 @@ class _Window(CommandObject, metaclass=ABCMeta):
         return {}
 
     @expose_command()
-    def keep_above(self) -> None:
+    def keep_above(self, enable: bool | None = None):
         """Keep this window above all others"""
 
     @expose_command()
-    def keep_below(self) -> None:
+    def keep_below(self, enable: bool | None = None):
         """Keep this window below all others"""
 
     @expose_command()
diff --git a/libqtile/backend/x11/window.py b/libqtile/backend/x11/window.py
index 8ef7e87c..8e2aaa25 100644
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -2125,12 +2125,18 @@ class Window(_Window, base.Window):
         self.window.configure(stackmode=xcffib.xproto.StackMode.Above)
 
     @expose_command()
-    def keep_above(self):
-        self.kept_above = not self.kept_above
+    def keep_above(self, enable: bool | None = None):
+        if enable is None:
+            self.kept_above = not self.kept_above
+        else:
+            self.kept_above = enable
 
     @expose_command()
-    def keep_below(self):
-        self.kept_below = not self.kept_below
+    def keep_below(self, enable: bool | None = None):
+        if enable is None:
+            self.kept_below = not self.kept_below
+        else:
+            self.kept_below = enable
 
     @expose_command()
     def move_up(self, force=False):
diff --git a/libqtile/group.py b/libqtile/group.py
index dd2df7e9..03415b30 100644
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -256,8 +256,8 @@ class _Group(CommandObject):
             win._float_state = FloatStates.FULLSCREEN
         elif self.floating_layout.match(win) and not win.fullscreen:
             win._float_state = FloatStates.FLOATING
-            if not win.kept_above and self.qtile.config.floats_kept_above:
-                win.keep_above()
+            if self.qtile.config.floats_kept_above:
+                win.keep_above(enable=True)
         if win.floating and not win.fullscreen:
             self.floating_layout.add_client(win)
         if not win.floating or win.fullscreen:

```

libqtile/core/manager.py
elParaguayo: I've moved these to the x11 core and subscribed to `client_managed` and `client_killed` hooks to call the methods.

m-col:
Unfortunately `bring_front_click = True` isn't working, but actually this seems to be the case on master as well, bizarrely. Was that known?

elParaguayo:
Works for me on master.

m-col:
I pushed a commit that stops the focus from moving between windows when moving them along the stack. The main issue that was happening was if you have the cursor over a window, then you move that window back behind a lower window, and if you have `follow_mouse_focus = True` then focus moves to that second window. I think this is unintuitive because you might expect hitting your keybind for `move_down()` a bunch of times would move the same window back incremently, but actually in that case it just swaps those two windows because the focus swaps between them each time.

> Works for me on master.

I'll re-test, I confused myself on some other testing because I thought I was restarting but I was only config reloading, and I could have done that here too.

m-col:
> I'll re-test, I confused myself on some other testing because I thought I was restarting but I was only config reloading, and I could have done that here too.

It works if `follow_mouse_focus = False` - these options seem to be interacting

elParaguayo:
> > I'll re-test, I confused myself on some other testing because I thought I was restarting but I was only config reloading, and I could have done that here too.
> 
> It works if `follow_mouse_focus = False` - these options seem to be interacting

It is indeed `False` in my config.

elParaguayo:
If that bring_to_front issue is an existing one. We could solve it separately. 

If you agree then I think we're close here. README needs fixing and we can squash it all down to a single commit.

Anything else you think we need here?

m-col:
> If that bring_to_front issue is an existing one. We could solve it separately.

Sorry, forgot to update on my re-test. It _is_ working on master, but on the PR branch, clicked windows aren't brought to the front if `follow_mouse_focus = True`, though I have no idea why they would interact!

libqtile/backend/x11/window.py
```diff
@@ -1118,20 +1114,15 @@ def change_layer(self, up=True, top_bottom=False, kwargs=None):
...
-            kwargs = {}
-        kwargs["stackmode"] = (
-            xcffib.xproto.StackMode.Above if above else xcffib.xproto.StackMode.Below
-        )
-        kwargs["sibling"] = sibling
+        kwargs = {
+            "stackmode": xcffib.xproto.StackMode.Above if above else xcffib.xproto.StackMode.Below,
+            "sibling": sibling
+        }
         self.window.configure(**kwargs)
```
m-col, line 1126: Maybe with only two static kwargs we can just do it inline, save the dictionary constructor:
```suggestion
        self.window.configure(
            stackmode=xcffib.xproto.StackMode.Above if above else xcffib.xproto.StackMode.Below,
            sibling=sibling,
	)
```

libqtile/backend/x11/window.py
```diff
@@ -1561,6 +1552,14 @@ def floating(self, do_float):
                 self._float_height = self.height
             self._float_state = FloatStates.NOT_FLOATING
             self.group.mark_floating(self, False)
+            stack = list(self.qtile.core._root.query_tree())
+            tiled = [win.window.wid for win in self.group.tiled_windows]
+            tiled_stack = [win for win in stack if win in tiled and win != self.window.wid]
```
m-col, line 1557: Nice! We can avoid the list constructor as we're iterating over the generator right away. Small nit but it helps me follow the X11 backend code if we use `wid` when we're only talking about the window ids:
```suggestion
            stack = self.qtile.core._root.query_tree()
            tiled = [win.window.wid for win in self.group.tiled_windows]
            tiled_stack = [wid for wid in stack if wid in tiled and wid != self.window.wid]
```

m-col (commented):
Nice this is great. I think we also need to do the inverse: when we make a tiled window not tiled, we should raise it at least above all tiled windows. We have two options there:

1. do what we're doing when maximizing and raise it to the top of the stack (respectfully) or
2. raise it above the tiled windows only. If it's visible when tiled, then it will remain visible when made floating. I guess the code here would be similar as that which you've just added: find the stack of `self.group.tiled_windows` and raise it right above the topmost of those.

CHANGELOG
```diff
@@ -43,6 +43,12 @@ Qtile x.xx.x, released XXXX-XX-XX:
           By default the keys are not passed (swallowed), so this argument is set to `True`.
           When set to `False`, the keys are passed to the focused client. A key is never swallowed if the
           function is not executed, e.g. due to failing the `.when()` check.
+        - added layering controls for X11:
+          - `lazy.window.keep_above()/keep_below()` marks windows to be kept above/below other windows permanently.
+             Calling the functions a further time removes the flag,
```
m-col, line 48:
```suggestion
             Calling the functions with no arguments toggles the state, otherwise pass `enable=True` or `enable=False`.
```

libqtile/backend/x11/core.py
```diff
@@ -276,6 +286,15 @@ def on_config_load(self, initial) -> None:
...
             self.qtile.manage(win)
 
+    def _on_client_managed(self, window):
+        self.update_client_list(self.qtile.windows_map)
+        self.update_client_stack()
+        window.change_layer()
+
+    def _on_client_killed(self, window):
+        self.update_client_list(self.qtile.windows_map)
+        self.update_client_stack()
```
m-col, line 296: Just noticed that these two functions are always called together. Should we just merge them into one `self.update_client_lists()`?

elParaguayo: Yes. We can. However, I don't think the logic in `_NET_CLIENT_LIST_STACKING` is correct. I need to look at that again.

libqtile/backend/x11/core.py
```diff
@@ -445,12 +445,16 @@ def update_client_lists(self) -> None:
...
-            for wid, c in self.qtile.windows_map.items()
-            if isinstance(c, window.Window) and c.group and wid in self._root.query_tree()
-        ]
+        # We rely on the stacking order from the X server
+        stacked_wids = []
+        for wid in self._root.query_tree():
+            if wid not in self.qtile.windows_map:
+                continue
+
+            win = self.qtile.windows_map[wid]
```
m-col, line 454: By using `dict.get` only one dictionary lookup is needed, saving a bit of runtime work:
```suggestion
            win = self.qtile.windos_map.get(wid)
            if not win:
                continue
```

m-col:
> we also need to do the inverse: when we make a tiled window not tiled, we should raise it at least above all tiled windows.

Was this addressed on those last couple of commits? Specifically I'm referring to making tiled windows 'regular' floating windows, not maximized windows. They will need to be risen up the stack, because it's unclear when you have only non-overlapping tiled windows which window is at the bottom, and making it floating can result in it hidden behind all tiled windows if it's not risen at least above the tiled windows.

libqtile/backend/x11/core.py
elParaguayo: This is why coding when you're tired is not the best idea.

elParaguayo:
> > we also need to do the inverse: when we make a tiled window not tiled, we should raise it at least above all tiled windows.
> 
> Was this addressed on those last couple of commits? Specifically I'm referring to making tiled windows 'regular' floating windows, not maximized windows. They will need to be risen up the stack, because it's unclear when you have only non-overlapping tiled windows which window is at the bottom, and making it floating can result in it hidden behind all tiled windows if it's not risen at least above the tiled windows.

Apologies, misunderstood the point. Yes, this does need fixing. We're doing this explicitly for maximised windows but that won't catch the scenario where a window is just made to float.

libqtile/backend/x11/window.py
```diff
@@ -1011,8 +1011,8 @@ def change_layer(self, up=True, top_bottom=False):
                     self.qtile.windows_map[i]
                     for i in stack
                     if i in self.qtile.windows_map
-                    and not isinstance(self.qtile.windows_map[i], Internal)
-                    and self.qtile.windows_map[i].group == self.group
+                    and (include_internal or not isinstance(self.qtile.windows_map[i], Internal))
+                    and ((include_internal and self.group is None) or self.qtile.windows_map[i].group == self.group)
```
m-col, line 1015: This big comprehension is getting a bit unwieldy. When we call this function, do we only operate on windows that below to this window's group? If so, what is the difference between the windows that are caught by `self.qtile.windows_map[i].group == self.group` and those found at `self.group.windows`?

I'm thinking if we're only concerned with this group's windows, we can get our windows of interest from there + `self.group.screen.gaps`.

Also while I'm here we replace the square brackets on this comprehension with round brackets to avoid the list constructor and iterating twice :)

libqtile/bar.py
```diff
@@ -308,6 +308,10 @@ def _configure(self, qtile, screen, reconfigure=False):
             )
             self.qtile.renamed_widgets.clear()
 
+        if self.qtile.core.name == "x11":
+            hook.subscribe.setgroup(self.keep_below)
+            hook.subscribe.startup_complete(self.keep_below)
```
m-col, line 313: How come we need to hook to these (i.e. versus doing `self.keep_below` once here only)?

libqtile/bar.py
```diff
@@ -678,5 +682,9 @@ def fake_button_press(self, screen, position, x, y, button=1):
         """
         self.process_button_click(x, y, button)
 
+    def keep_below(self):
+        self.window.keep_below(enable=True)
+        self.window.change_layer(include_internal=True, top_bottom=True, up=False)
```
m-col, line 687: The `change_layer` method here is an X11-specific implementation deal leaking from the backend. Can we avoid that? Perhaps conceptually it would make sense for `self.window.keep_below()` to be responsible for putting it in its place if it isn't already there?

m-col (changes requested):
I don't think we should always make the bar `keep_below` as a rule. Can we make this configurable and opt-in i.e. with a `keep_below: bool = False` option for `Bar`?

elParaguayo:
This last commit is ugly as hell. I'm mainly interested in seeing if this fixes the bug mentioned on Discord. If it does, I'll tidy it up!

libqtile/backend/x11/core.py
```diff
@@ -232,15 +234,22 @@ def on_config_load(self, initial) -> None:
...
             return
 
+        hook.subscribe.client_focus(self.check_stacking)
+        hook.subscribe.client_managed(self._on_client_managed)
+        hook.subscribe.client_killed(self._on_client_killed)
+
         # Qtile just started - scan for clients
-        _, _, children = self._root.query_tree()
+        children = (window.XWindow(self.conn, i) for i in self._root.query_tree())
         for item in children:
```
m-col, line 253:
```suggestion
        for wid in self._root.query_tree():
	    item = window.XWindow(self.conn, wid)

**Comment by elParaguayo** [comment]
@m-col save your time. I don't think this is ready for review yet.

**Comment by fdev31** [comment]
I made some quick experiment to add support for "desktop" windows (eg: xfdesktop, livewallpaper, ...).
It required some kind of sticky state (I have some custom one in my config based on a well-known example), and the following attributes set in a hook:
```py
        _sticky_windows.add(window)
        window.maximized = True
        window.floating = True
```
Here I use "maximized" instead of "fullscreen" to avoid some problems, but ideally it should be some kind of fullscreen...

I had to patch Qtile itself too:

```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index fc9ae169..ed6ab666 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -846,7 +849,7 @@ class Core(base.Core):
             if qtile.config.bring_front_click and (
                 qtile.config.bring_front_click != "floating_only"
                 or getattr(window, "floating", False)
-            ):
+            ) and window.get_wm_type() != "desktop":
                 self.conn.conn.core.ConfigureWindow(
                     window.wid, xcffib.xproto.ConfigWindow.StackMode, [StackMode.Above]
                 )
diff --git a/libqtile/backend/x11/window.py b/libqtile/backend/x11/window.py
index 3a3de0f7..1ab721ee 100644
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1571,7 +1571,7 @@ class Window(_Window, base.Window):
             else:
                 # if we are setting floating early, e.g. from a hook, we don't have a screen yet
                 self._float_state = FloatStates.FLOATING
-            if not self.kept_above and self.qtile.config.floats_kept_above:
+            if not self.kept_above and self.qtile.config.floats_kept_above and self.window.get_mw_type() != "desktop":
                 self.keep_above(enable=True)
         elif (not do_float) and self._float_state != FloatStates.NOT_FLOATING:
             self.update_fullscreen_wm_state(False)
```

elParaguayo:
@fdev31 Should a `desktop` window be below everything, without the ability to move it up?

libqtile/bar.py
m-col: Right, but doesn't `keep_below` set a state that will always make it get stacked below visible windows until the flag is removed? Wouldn't the issue you describe only happen if we were moving the window to the bottom as a one-shot change (rather than setting a flag)?

m-col:
> @fdev31 Should a desktop window be below everything, without the ability to move it up?

(discussed offline but replying for posterity) yes, a desktop window should be below everything. These windows are like the desktop metaphor utilities that contain file/app launcher icons in more sophisticated DEs, so the only thing behind them is the wallpaper (which in X isn't really a window)

libqtile/backend/base.py
```diff
@@ -253,6 +250,54 @@ def info(self) -> dict[str, Any]:
         """
         return {}
 
+    @expose_command()
+    def keep_above(self, enable: bool | None = None, include_internal: bool = False):
+        """Keep this window above all others"""
+
+    @expose_command()
+    def keep_below(self, enable: bool | None = None, include_internal: bool = False):
+        """Keep this window below all others"""
```
m-col, line 259: What does `include_internal` do? Perhaps worth adding a line to describe that, like with the methods below?

libqtile/backend/x11/window.py
```diff
@@ -903,21 +901,278 @@ def place(
...
+        #         sibling = lower[-1]
+        #         above = True
+
+        # # If we're already in the right position in the current layer, check if there's
+        # # anything higher and, if so, place ourselves below it
+        # # This is needed to ensure the window is restacked if there are windows in a
+        # # higher layer
+        # if sibling == self.wid and higher:
+        #     sibling = higher[0]
+        #     above = False
```
m-col, line 1128: Just posting here in case you left these comment blocks in by accident

m-col (commented):
code itself looks good, just a couple of minor Qs. Though one or two of the older comment threads aren't resolved yet

libqtile/bar.py
elParaguayo: The flag does persist and it will mean that the bar is in a lower group than "normal" windows. That bit is True. However, when a window is stacked, the code isn't checking _every_ window each time. It tries to position itself in the correct order within its own layer. It only checks that that layer is correctly placed if there are no other windows in the layer (this is what happens at line 1130 onwards). The theory here is that, when there's only one window in that layer then it's important to make sure that layer is positioned correctly. Any new windows added to the layer don't need the same process as the layer should have already been positioned correctly.

m-col: OK, gotcha. Then no worries, if having that here makes things easier then +1 from here. I shouldn't let my needless perfectionism get in the way of actual progress :smile:

libqtile/backend/x11/window.py
elParaguayo: Yup - these will get removed.

libqtile/backend/base.py
elParaguayo: We stack windows in the current group but, as the bar's group is `None`, we don't stack it directly. Using `include_internal` adds the screen's bars to the list of windows to stack. Typically, we don't want them included because otherwise things like `move_up/down` will include the bar.

However, it's only the bar that calls `change_layer` with `include_internal=True` but I added some code that exits `change_layer` when the window's `group` is `None`. 

We can probably strip out some of that code.

The bars are still included when doing that check I mentioned earlier (checking layers are in the correct order)

m-col: ok makes sense :+1:

docs/manual/config/lazy.rst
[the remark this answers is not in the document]
elParaguayo: Behaviour is as follows:
- `bring_to_top` puts window on top of everything
- `move_to_top` puts window at top of current layer
- passing `force=True` will remove `kept_below` flag if applied

elParaguayo:
> deskt

Don't know if it's one for this PR but I would also think that DESKTOP windows shouldn't be tiled.

libqtile/backend/x11/window.py
```diff
@@ -903,21 +901,259 @@ def place(
...
+        if desktop:
+            # mypy can't work out that this gives us tuple[bool, bool, bool, bool, bool]...
+            # (True, False, False, False, False, False)
+            return tuple(not i for i in range(6))  # type: ignore
+
+        # ...otherwise, we set to the highest matching layer.
+        # Look for the highest matching level and then set all other levels to False
+        highest = max(i for i, state in enumerate(states) if state)
+
+        # mypy can't work out that this gives us tuple[bool, bool, bool, bool, bool]...
```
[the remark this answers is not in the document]
elParaguayo: Yes, 6 is correct. I'll check if that's why `mypy` said there was an error.

libqtile/backend/x11/window.py
```diff
@@ -506,6 +500,7 @@ def __init__(self, window, qtile):
         self.float_y: int | None = None
         self._float_width: int = self._width
         self._float_height: int = self._height
+        self.previous_layer = (False, False, True, False, False)
```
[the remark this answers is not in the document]
elParaguayo: Probably doesn't impact the result but, yes, it should be 6 to be consistent.

libqtile/backend/x11/window.py
```diff
@@ -903,21 +901,259 @@ def place(
...
+                lower += [
+                    bar.window.wid
+                    for bar in group_bars
+                    if bar.window.get_layering_information() > layering
+                ]
+                higher += [
+                    bar.window.wid
+                    for bar in group_bars
+                    if bar.window.get_layering_information() < layering
+                ]
```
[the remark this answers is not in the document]
elParaguayo: Yes - that would be better.

jwijenbergh (approved):

libqtile/backend/x11/core.py
```diff
@@ -232,16 +234,23 @@ def on_config_load(self, initial) -> None:
...
         if not initial:
             # We are just reloading config
             for win in self.qtile.windows_map.values():
                 if type(win) is window.Window:
                     win.set_group()
             return
 
+        hook.subscribe.client_focus(self.check_stacking)
+        hook.subscribe.client_managed(self._on_client_managed)
+        hook.subscribe.client_killed(self._on_client_killed)
```
m-col, line 249: (Happy to iterate on this in a future PR, ideally adding a TODO here for the time being). As the backends listen for events from clients and hardware/X server, they are ultimately responsible for firing hooks, so it feels unnatural to subscribe to hooks in the backend too. E.g. currently the X backend subscribes to none, only fires them. The wayland backend subscribes only to `startup_complete` though for the same reason I think that should actually use an asyncio `call_soon` handler instead.

The idea is similar to a producer-consumer relationship, where the backend emits events and other python code within the qtile process listens for them as a consumer, and ideally we'd keep that relationship clean.

Concrete suggestions:

- drop 247, instead do `self.core.check_stacking()` here: https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/window.py#L1044
- drop 248 and `Core._on_client_managed`, instead do `self.core.update_client_lists(); self.change_layer()` here: https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/window.py#L1462
- drop 249 and `Core._on_client_killed`, instead do `self.update_client_lists()` after `unmanage()` in these two places:
  - https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/core.py#L728
  - https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/core.py#L754

libqtile/backend/x11/window.py
```diff
@@ -506,6 +500,7 @@ def __init__(self, window, qtile):
         self.float_y: int | None = None
         self._float_width: int = self._width
         self._float_height: int = self._height
+        self.previous_layer = (False, False, True, False, False, False)
```
m-col, line 503: Would it be helpful to our future selves to add a comment here listing what these are?

libqtile/bar.py
m-col, line 313: TBH this is general user logic that I guess isn't X-specific? Probs fine to drop the backend check:
```suggestion
        hook.subscribe.setgroup(self.keep_below)
        hook.subscribe.startup_complete(self.keep_below)
```

m-col (changes requested):
Left a couple Qs and suggestions mostly about API and docs, but this is really looking fantastic.

re making the bar `kept_below` configurable - let's leave it for now, and make it configurable if/when a user requests it to be configurable.

Lastly I want to reiterate how much of an amazing piece of work this is, you truly are an absolute hero.

libqtile/backend/base.py
```diff
@@ -253,6 +250,54 @@ def info(self) -> dict[str, Any]:
...
+    def keep_below(self, enable: bool | None = None):
+        """Keep this window below all others"""
+
+    @expose_command()
+    def move_up(self, force: bool = False) -> None:
+        """
+        Move this window above the next window along the z axis.
+
+        Will not raise a window where "keep_below" is True unless
+        force is set to True.
```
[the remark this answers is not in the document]
elParaguayo: If window is normal and only other window is `kept_above`, this will *not* move above it, even if `force=True` so it's the first one. I'll make this clearer in the code.

docs/manual/config/lazy.rst
```diff
@@ -124,6 +124,23 @@ Window functions
...
+    * - ``lazy.window.move_up()``
+      - Move the window above the next window in the stack.
+    * - ``lazy.window.move_down()``
+      - Move the window below the previous window in the stack.
+    * - ``lazy.window.move_to_top()``
+      - Move the window above all other windows with similar priority
+        (i.e. a "normal" window will not be moved above a ``kept_above`` window).
+    * - ``lazy.window.move_to_bottom()``
+      - Move the window below all other windows with similar priority
+        (i.e. a "normal" window will not be moved below a ``kept_below`` window).
```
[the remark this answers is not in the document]
elParaguayo: The docstrings will be used to populate the API pages automatically but you're right that this page isn't done that way.

The idea of the `lazy` page was to show some useful bindings to help in configs. We could probably do something to populate this automatically but I think it's one for later.

libqtile/backend/x11/core.py
elParaguayo: I think we should do this now. However, the `client_managed` bit you've linked to is in the `Static` class. I think the hook is fired for other windows by the manager here: https://github.com/qtile/qtile/blob/4ecf1bdef46d68da56145a654701ed355dcad64b/libqtile/core/manager.py#L677

elParaguayo: `unmanage` is more complicated as we need the window that's passed when the hook is fired. We could have `self.qtile.unmanage` return the window so it can be used by the calling method.

m-col (approved):

maplepy:
Hello! Was this implemented and if so, how can I use it? Thanks :)

elParaguayo:
Yes, it's in version 0.23. Available commands are shown here: https://docs.qtile.org/en/stable/manual/commands/api/windows.html

What do you mean by "how can I use it?"

maplepy:
How can I make floating windows sticky like in bspwm for instance? Like bind a keybind to make any window floating and sticky ? Sorry if my question is dumb, I am trying to learn

PR #3416, opened 2022-03-10
border_on_single option for Bsp Layout

K1llf0rce:
True/False option for drawing the border when there is only a single window. (like in the Columns Layout f.E.)


elParaguayo:
Thanks. Code looks good. Will try to test soon.

elParaguayo:
This runs as expected. Will merge once the tests pass.

Issue #3256, opened 2022-01-26
[feature-req] save positioning of windows


philiparvidsson:
I would like a way to save the position of windows so that after a reboot, when I autostart my apps, they are positioned as I want them. Right now, I have to begin every session by resizing windows.

Can this be done today? What can I do to achieve this? If I can't, then I'd like to request it as a feature.

elParaguayo:
I assume we're talking floating windows here...

I think you'd need to set up a function in your config which is called on the `client_managed` hook. You then compare that client's name or class and resize it (`window.cmd_set_size_floating(w, h)`) and position (`window.cmd_set_position_floating(x, y)`) based on some predetermined values.

philiparvidsson:
Sorry I was being unclear! This is for NON-floatign windows. I want windows to be positionend/sized the same time after reboot and autostarting all my apps.

elParaguayo:
OK. Interesting.

You'd need to run some code at the end of a startup hook that runs commands on the relevant groups/layouts to position and size them. It seems quite fragile though as this would depend on the right number of windows being open in each group at the time.

zordsdavini:
I had idea that I could „save“ the session and reboot. Then „load“ the session and get all opened apps back in same layout with sizes and positions

zordsdavini:
https://github.com/qtile/qtile/pull/3424 was a perfect solution. The argument that other layouts doesn't have is not critical. Under the given base other implement would be much easier if needed. Sadly we don't have feedback what layouts are used more. Sorry, I didn't noted this PR

PR #3424, opened 2022-03-13
Add save and load layout feature to Columns

zys5945:
Hi all!

### What is this?
I came from i3, and one of the features I used in i3 on a daily basis is [layout saving](https://i3wm.org/docs/layout-saving.html). It makes me a bit sad that qtile doesn't currently support it. That's why I decided it to implement something similar for the `Columns` layout.

This is a draft PR. Docs and unit tests are not included. Just want to know if this goes against any design philosophies, or if there's any major flaws in my implementation.

### Feature explanation

Two cmds are added for `Columns`: `cmd_save_layout` and `cmd_load_layout`.

`cmd_save_layout` will serialize some fields of current windows in the `Columns` object into a json file.

`cmd_load_layout` will read the json file and try to recreate the layout. Windows are matched using the `Match` object, created from rule definitions in the json file. Window position, column widths as well as window focus will be restored. Note this command does not create any of the windows, it performs layout only.

`serialize.py` is added to extract some common functionality for reuse, in case other layouts may want to implement save / load in the future.

Two utility methods are also added to the `Match` class.

### Example

```
# Saves current layout to ~/.config/qtile/layouts/test.json
qtile cmd-obj -o layout -f save_layout -a test

# Saves current layout to /tmp/test.json
qtile cmd-obj -o layout -f save_layout -a /tmp/test.json

# Saves group 1's layout to ~/.config/qtile/layouts/test.json
# not sure what is the best way to get a specific group
qtile cmd-obj -o group -f eval -a 'self.qtile.groups_map["1"].layout.cmd_save_layout("test")'
```

```
# load layout (note, assuming all the windows are created)

# load ~/.config/qtile/layouts/test.json to the current layout
qtile cmd-obj -o layout -f load_layout -a test

# load /tmp/test.json to the current layout
qtile cmd-obj -o layout -f load_layout -a /tmp/test.json

# load ~/.config/qtile/layouts/test.json to the group 1's layout:
qtile cmd-obj -o group -f eval -a 'self.qtile.groups_map["1"].layout.cmd_load_layout("test")'
```

Also closes https://github.com/qtile/qtile/issues/3256

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -63,5 +63,6 @@ Tips & Tricks
 
     manual/howto/widget
     manual/howto/git
+    manual/howto/save-load-layout
 
 * :ref:`genindex`
--- a/docs/manual/howto/save-load-layout.rst
+++ b/docs/manual/howto/save-load-layout.rst
@@ -1,0 +1,119 @@
+.. _save-load-layout:
+
+====================
+Save and load layout
+====================
+
+.. note:: Currently only ``Columns`` support layout saving and loading.
+
+Layout can be saved to and restored from json files using ``save_layout`` and ``load_layout`` commands on the ``Layout`` object.
+
+Window positions, sizes, and in general things that are controlled by ``Layout`` will be saved and restored.
+
+Note ``load-layout`` does not create any windows. Instead, it uses the window matching rules defined in the json file to match against open windows, and apply saved stylings to those matched windows. See `Layout files`_ for more info.
+
+By default qtile will look for layout files in ``CONFIG_DIRECTORY/layouts``, e.g. ``~/.config/qtile/layouts``
+
+Typical workflow
+================
+
+To save layout:
+
+- Create some windows
+
+- Save layout using ``save-layout`` command
+
+- Edit the generated json file with more accurate window matching rules
+
+Then to load layout:
+
+- Create the same windows
+
+- Load layout using ``load-layout`` command
+
+``save-layout`` examples
+========================
+
+.. code-block:: bash
+
+    # Save current layout to default location (i.e. ~/.config/qtile/layouts/my_layout.json)
+    qtile cmd-obj -o layout -f save_layout -a my_layout
+
+    # Save current layout to a specified place
+    qtile cmd-obj -o layout -f save_layout -a "/tmp/my_layout.json"
+
+``load-layout`` examples
+========================
+
+.. code-block:: bash
+
+    # Load ~/.config/qtile/layouts/my_layout.json
+    qtile cmd-obj -o layout -f save_layout -a my_layout
+
+    # Load /tmp/my_layout.json
+    qtile cmd-obj -o layout -f load_layout -a "/tmp/my_layout.json"
+
+Layout files
+============
+
+Using ``Columns`` as example, the json file produced by ``save_layout`` will look something like:
+
+.. code-block:: json
+
+    {
+        "name": "columns",
+        "columns": [
+            {
+                "split": true,
+                "insert_position": 0,
+                "width": 100,
+                "windows": [
+                    {
+                        "title": "my_window",
+                        "wm_type": "...",
+                        "wm_class": "...",
+                        "role": "...",
+                        "focus": true
+                    },
+                    {
+                        "title": "my_other_window",
+                        "wm_type": "...",
+                        "wm_class": "...",
+                        "role": "...",
+                        "focus": true
+                    }
+                ],
+                "heights": [
+                    100
+                ],
+                "current": 0
+            }
+        ]
+    }
+
+Most fields such as ``split``, ``insert_position`` are taken directly from the ``Columns`` object. These fields will be used to configure the ``Columns`` object when the layout is loaded.
+
+Each item in ``windows`` is used to create a ``Match`` object. These match objects are used to match against open windows in the current layout. Styling is then applied to matched windows.
+
+E.g. the above layout file will place the window with title "my_window" at the top of first column. Then place "my_other_window" below "my_window" in the same column.
+
+All constructor arguments for ``Match`` are supported:
+
+.. qtile_class:: libqtile.config.Match
+    :no-commands:
+    :noindex:
+
+Using regex to match windows
+============================
+
+Suffix field name with ``_regex`` to use regex matching instead. All fields except ``net_wm_pid`` support ``_regex``.
+
+When both the ``_regex`` and non-regex fields are given, ``_regex`` field takes precedence.
+
+E.g. the following example matches any window with a title that contains "foobar"
+
+.. code-block:: json
+
+    {
+        "title_regex": ".*foobar.*"
+    }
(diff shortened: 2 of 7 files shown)


m-col:
With a read through the code, the implementation looks pretty good - thanks for submitting this!

> Just want to know if this goes against any design philosophies, or if there's any major flaws in my implementation.

Overall I think your approach is good, and I do think this would be valuable to have in Qtile. This is asked about quite often.

As it's a non-trivial addition with new logic, please could you add full type hints? This will make it easier to read and review.

Could you describe how you imagine this feature (when fully implemented) would work from a user's perspective? It might be intuitive to have a way to have all of your layouts restore their previous state upon startup. Would this be possible? And by extension, maybe there would be a way to save all layouts? How do the serialised layouts relate with groups?

zys5945:
> could you add full type hints?

Sure. That will require a bit refactoring around typing the dicts, but I can definitely do that.

> It might be intuitive to have a way to have all of your layouts restore their previous state upon startup. Would this be possible?

The problem is window managers like qtile just doesn't understand the context of how to create a process and save / load their state. E.g. what environment variables each process was created with, which file handles they have open, how they interact with hardware, etc. AFAIK only hibernation or sleep can do full process state save and reload (even then they are not perfect), which is something far beyond the scope of a window manager.

> Could you describe how you imagine this feature (when fully implemented) would work from a user's perspective?

So this feature only deals with the saving and restoring window position, size and window focus. The creation of the windows themselves need to be handled by the user.

E.g. my routine is: create a bunch of windows, have them nicely laid out, then use something similar to `qtile cmd-obj -o layout -f save_layout -a my_layout`

Then I will have a separate script that creates all the windows, and calls qtile to restore the layout, e.g.:

```
import subprocess

subprocess.run(['chromium', 'http://my.favorite.website])

zys5945:
Added tests and docs.

Tests passed locally. Need approval on running workflow

zys5945:
Think I need some help with this doc build failure. Only added `typing_extensions` requirement to the docs tox environment, don't think it's causing the build failure?

elParaguayo:
I think you've got a couple of options:

1) Try to import `TypedDict` from `typing` and, if that gives an `ImportError`, fall back to `typing_extensions`

2) Add `typing_extensions` to `docs/requirements.txt`.

Personally, I prefer option 1 as, otherwise, it looks like everyone needs to install a non-standard library which we should avoid.

zys5945:
Wasn't sure if conditional check on import will break mypy or not. Turns out it didn't 🍻

I thought `typing_extensions` would be available during docs building since `libqtile.core` already uses it. I guess it's just not part of the docs.

I went with the version checks and all seems to be working. Docs building is on python 3.10 so `TypedDict` is available natively. Didn't need to add extra dependencies. Thanks for the help!

Also can someone approve running workflow?

zys5945:
Updated code to cover certain fields missing from mocks in 3.7. Please start the workflow again.

zys5945:
Failure: `test_systray_icons` seems unrelated?

elParaguayo:
> Failure: `test_systray_icons` seems unrelated?

Yes. Unrelated and not sure why it failed. Rerunning tests to see if it passes this time.

zys5945:
Tests passed. This CR is ready for review / merge.

zys5945:
That's some very good points. I thought about some of them during the refactoring process, and I think I didn't make some of my points very clear in my previous reply as well.

> You have limited the scope of state restoration such that we can match windows, and we can configure layouts, and this is good. That's probably enough. (To be honest we may be able to [reconstruct command lines](https://psutil.readthedocs.io/en/latest/#psutil.Process.cmdline) because we can get PIDs for each window, and re-launch the processes upon startup. Presumably that's what big DEs do. Perhaps that makes a lot of assumptions about what people want starting up on their systems tho.).

I think we will have very limited success. It is possible to extract cli arguments from PID, but things like environment variables or file / hardware configurations on startup are impossible to fully recreate without user configuration. Also any secondary windows created by user actions that are not part of process startup is also quite unlikely to be recreated. I'd prefer to give users full power to create the processes, then call the commands to layout the windows, rather than having them supplying startup configurations.

> 1. A startup_once hook defined in the default config (probably commented out but with some notes) that will look for layout states stored on disk and load them.
> 2. A shutdown hook also defined and commented out that will persist the state of all current layouts in those files.

It's definitely possible to have startup / shutdown hooks that automatically create and close / save layouts. I didn't want to force that on the users since I think it's also a valid use case to start "sessions" on an adhoc basis. E.g. I may have one set of windows specifically for studying, then another set for working, etc, and choosing which layout to use depending on what I'm trying to do on that day. I believe the current design supports both use cases quite well.

It will definitely be used by less users if not all layouts support it. I'm just assuming that most users are probably using columns and floating, or variants of that (I could be very wrong here since I do have an i3 bias).

> I think we need to provide a simple and clean API for layout classes to define their own serialization/deserialisation logic.

This is definitely a concern. I've spent some time thinking about it and here's my current design (and plans for potential future layout onboardings): 

- All things that should be serialized/deserialized have 2 classes.
  One named `Serialized<X>` which is a pure `TypedDict` representing the serialized data (e.g. `SerializedWindow`).
  Another name `<X>SerDes` for serdes logic (e.g. `WindowSerDes`).
  This structure plays nicely with mypy and is also easy to follow IMO. It avoids adding more clump to the layout class itself.

- Common functionalities are placed in `serialize.py`, which right now contains data and serdes logic of `Window`, as well as file operation logic. I don't see this file getting too many other things, since all that's left is layout specific settings.

- Each layout that supports serdes should have 2 methods: `serialize` and `deserialize`, and two commands: `cmd_save_layout` and `cmd_load_layout`. Only difference being `serialize` and `deserialize` deals directly with `Serialized<X>` objects, whereas the commands would save to / load from file. It is quite trivial to have a default implementation through mixin or base class for all layouts since they are just oneline functions that call the respective serdes functions.

Group saving and reloading is also quite trivial with this design. We can simply iterate through the list of layouts and call `serialize` and `deserialize`. I do think there's value in supporting both save/load group and save/load individual layout. However if it feels too bloated then we can definitely just remove the commands from the layout.

zys5945:
FWIW, if this PR goes through then I do plan to work on adding this support to floating layout, and eventually Group to save / load all layouts in one cmd, but I don't see myself working on other layouts.

elParaguayo:
It feels like this may be better suited to [qtile-examples](https://github.com/qtile/qtile-examples) rather than the main repo. As m-col says, I don't think we'd want to implement a change for just one layout (and I understand why you wouldn't want to work on other layouts).

However, if someone else sees it on qtile-examples and decides to broaden its scope to other layouts then we could reconsider a PR. As it is, I don't see this getting merged in its current state.

If you need some help to implement this as standalone code then please start a discussion and we'll help where we can.

@m-col if you agree with the above then I think we can close this for now.

zys5945:
I'd argue it's unlikely for someone to implement this feature for all layouts at once, especially considering there are still new layouts being added.

m-col:
> @m-col if you agree with the above then I think we can close this for now.

I agree. This is a great feature and hopefully in the future we can have it in qtile proper, but it would be a headache having it in a limited form and we don't have the man power to implement support for all the other layouts, and people _will_ be asking for them.

PR #3428, opened 2022-03-16
Add test for vertical bars

elParaguayo:
This was one of the suggestions in #2134.

We just replicate the init and configure test for widgets that support vertical bars.


m-col:
Nice, thanks!

PR #3429, opened 2022-03-16
Add `margin_on_single` option to Bsp layout

### LINKED ISSUE #3193: Support for single_margin in bsp layout
The MonadTall layout supports `single_margin` for configuring a specific margin (i.e. `0`) when there is only one window. And for the Columns layout there's the same option but it's called `margin_on_single`. I'd like to request the same configuration option for the bsp layout.

elParaguayo:
Closes #3193

### DIFF
--- a/libqtile/layout/bsp.py
+++ b/libqtile/layout/bsp.py
@@ -153,6 +153,11 @@ class Bsp(Layout):
         ("border_normal", "#220000", "Border colour(s) for un-focused windows."),
         ("border_width", 2, "Border width."),
         ("border_on_single", False, "Draw border when there is only one window."),
+        (
+            "margin_on_single",
+            None,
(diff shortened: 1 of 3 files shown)


libqtile/layout/bsp.py
```diff
@@ -209,6 +214,11 @@ def configure(self, client, screen_rect):
         node = self.get_node(client)
         color = self.border_focus if client.has_focus else self.border_normal
         border = 0 if node is self.root and not self.border_on_single else self.border_width
+        margin = (
+            self.margin_on_single
+            if (node is self.root and self.margin_on_single is not None)
+            else self.margin
+        )
```
m-col, line 221: Could this be simplified by doing this in `__init__`?:
```py
    if self.margin_on_single is None:
        self.margin_on_single = self.margin
```

m-col (approved):
Small question but otherwise looks great :)

PR #3431, opened 2022-03-17
Add instructions on running with kde

simrat39:
Pretty much title. This works for me after a lot of messing around, would love if some people could test it and we could have it in the docs so others don't struggle as much

### DIFF
--- a/docs/manual/config/kde.rst
+++ b/docs/manual/config/kde.rst
@@ -1,0 +1,32 @@
+====================
+Running Inside KDE
+====================
+
+Add the following file to the following location. This prevents kwin from starting which allows qtile to startup later on.
+::
+  $ cat $HOME/.config/plasma-workspace/env/wm.sh
+  
+  #!/bin/sh
+  export KDEWM=" "
+  
+Now, Add the following session file
+::
+  $ cat /usr/share/xsessions/plasma-qtile.desktop
+  
+  [Desktop Entry]
+  Type=XSession
+  Exec=/path/to/script
+  TryExec=/usr/bin/startplasma-x11
+  DesktopNames=KDE
+  Name=Plasma (Qtile)
+
+Change the Exec to a new script which you can make wherever you want
+::
+  $ cat /path/to/script
+  
+  qtile start &
+  startplasma-x11
+
+Make sure that this script is executable
+::
+  chmod +x /path/to/script


elParaguayo:
Thanks for submitting.

I would like to see confirmation from other users that this does work before we add to our docs.

hanschen:
I had some trouble with this when I tested Qtile, so having it in the documentation would be nice. Especially since you can't just use the standard method of specifying `export KDEWM="/usr/bin/qtile start"`.

Some suggestions:

1. Just a nitpick, I would suggest changing the title to "Running with KDE Plasma" or something similar. Plasma is the name of the desktop.
2. For Plasma 5.25 and later, I believe systemd is used to start Plasma. So you need to either create a new systemd service and 
mask the old one, or run the following command once to disable the new systemd-based startup:
``` sh
kwriteconfig5 --file startkderc --group General --key systemdBoot false
```
3. I had to add the following to my Qtile config to automatically kill the Plasma desktop window(s) (maybe there's a better way?):
``` python
@hook.subscribe.client_new
def auto_kill_plasma(window):
    wm_class = window.get_wm_class()
    if wm_class and "plasmashell" in wm_class:
        window.kill()

```

4. It could be good to add a note to make sure to disable overlapping keyboard shortcuts in System Settings to avoid conflicts between Plasma/KDE apps and Qtile.
5. A tip for Plasma users who still want to use the Plasma system tray is to put it in a scratchpad, something like this:
``` python
groups = [
    ...,
    ScratchPad("scratch", [
        DropDown(
            name="systray",
            cmd="plasmawindowed org.kde.plasma.systemtray",
            height=0.1,
            width=0.1,
            x=0.87,
            opacity=1,
            on_focus_lost_hide=True,
        ),
    ]),
]

keys = [
    ....,
    Key([mod], "s", lazy.group['scratch'].dropdown_toggle('systray')),
]

```

I make the system tray widget "autostart" using the following, which isn't very elegant...

```python
@hook.subscribe.startup_complete
def autostart_scratchpad():
    qtile.cmd_simulate_keypress([mod], "s")  # adjust based on your keybinding
    qtile.cmd_simulate_keypress([mod], "s")
```

Unfortunately the systemtray scratchpad isn't closed the first time Qtile is started.

Here are some other nice resources for running KDE Plasma with other WMs (mostly i3) which could serve as inspiration:
- https://userbase.kde.org/Tutorials/Using_Other_Window_Managers_with_Plasma (a bit out-of-date)
- https://maxnatt.gitlab.io/posts/kde-plasma-with-i3wm/
- https://github.com/heckelson/i3-and-kde-plasma

mocksoul:
you can run qtile with kde in a different and much more accurate form.

disable plasma-kwin_x11 user service

```
systemctl --user disable plasma-kwin_x11.service
```

1. set KDE_FULL_SESSION=true
1. run qtile as a first citizen
1. run any KDE services you need (if you need :-D). e.g.:

```
systemctl --user start plasma-powerdevil.service
systemctl --user start plasma-krunner.service
systemctl --user start kglobalaccel.service  # this is KDE global shortcut daemon and it will eat any defined shortcuts =)
```

4. semi-background apps (controllable via dbus) can be launched too:

```
systemctl --user start app-org.kde.spectacle.service
```

This way you have **waaaaaaay** better control of everything.

KDE's plasma runs completely in user's systemd stack with dbus messaging. Thus it is easy to control that way.

Also I cant see any reason for running plasma-plasmashell, since this is KDE whole workspace and it will be rendered as window in qtile which is useless IMO.

jwijenbergh:
> you can run qtile with kde in a different and much more accurate form.
> 
> disable plasma-kwin_x11 user service
> 
> ```
> systemctl --user disable plasma-kwin_x11.service
> ```
> 
>     1. set KDE_FULL_SESSION=true (without that big apps like kmail or kate will end up in busyloop waiting for KDE startup lol)
> 
>     2. run qtile as a first citizen
> 
>     3. run any KDE services you need (if you need :-D). e.g.:
> 
> 
> ```
> systemctl --user start plasma-powerdevil.service
> systemctl --user start plasma-krunner.service
> systemctl --user start kglobalaccel.service  # this is KDE global shortcut daemon and it will eat any defined shortcuts =)
> ```
> 
>     4. semi-background apps (controllable via dbus) can be launched too:
> 
> 
> ```
> systemctl --user start app-org.kde.spectacle.service
> ```
> 
> This way you have **waaaaaaay** better control of everything.
> 
> KDE's plasma runs completely in user's systemd stack with dbus messaging. Thus it is easy to control that way.
> 
> Also I cant see any reason for running plasma-plasmashell, since this is KDE whole workspace and it will be rendered as window in qtile which is useless IMO.
> 
> tip: dont make ugly oldschool startup bash scripts. Link user services to graphical-session.target instead making them `PartOf` it. E.g. for kbdd:
> 
> ```
> [Install]
> WantedBy=graphical-session.target
> 
> [Unit]
> Description=KBDD
> PartOf=graphical-session.target
> 
> [Service]
> ExecStart=/usr/bin/kbdd --nodaemon
> Restart=on-failure
> Slice=background.slice
> ```
> 
> and bind qtile itself to it in ~/.xsession (if using sddm) or any other startup script
> 
> ```
> systemd-run \
>     --user \
>     --unit wm \
>     --scope \
>     --slice-inherit \
>     --property=BindsTo=graphical-session.target \
>     qtile start --backend x11
> ```
> 
> voila!
> 
> p.s. @hanschen nice tip to run plasmawindowed tray btw, thnx!

do you want to make a PR?

hanschen:
Just an update, using the new systemd mechanism (Plasma 5.25 and later), I found it really easy to run a KDE Plasma desktop environment with Qtile instead of KWin:

1. Disable KWin and Plasma Shell:
    ```sh
   systemctl --user mask plasma-kwin_x11.service
   systemctl --user mask plasma-plasmashell.service
   ```

2. Add a Qtile systemd service file to `$HOME/.config/systemd/user/plasma-qtile.service`:
   ```ini
   [Install]
   WantedBy=plasma-workspace.target
   
   [Unit]
   Description=Plasma with Qtile window manager
   Before=plasma-workspace.target
   
   [Service]
   ExecStart=/usr/bin/qtile start
   Slice=session.slice
   Restart=on-failure
   ```
3. Reload systemd manager configuration:
   ```sh
   systemctl --user daemon-reload
   ```
4. Enable the new `plasma-qtile` service:
   ```sh
   systemctl --user enable plasma-qtile.service
   ```

Reference: https://wiki.archlinux.org/title/KDE#Replacing_KWin_service

PR #3436, opened 2022-03-18
WL: output power management: only unblank outputs that blanked

m-col:
Currently the wayland core indiscriminately blanks and unblanks outputs
(via the output power management protocol) without considering which
should be enabled or disabled. Outputs may be disabled through user
configuration of the output layout. We should only unblank outputs if it
was the power management protocol that blanked them in the first place.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -145,6 +145,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.output_manager = OutputManagerV1(self.display)
         self.add_listener(self.output_manager.apply_event, self._on_output_manager_apply)
         self.add_listener(self.output_manager.test_event, self._on_output_manager_test)
+        self._blanked_outputs: set[Output] = set()
 
         # set up cursor
         self.cursor = Cursor(self.output_layout)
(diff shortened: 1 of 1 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -444,10 +445,28 @@ def _on_new_inhibitor(self, _listener: Listener, idle_inhibitor: IdleInhibitorV1
...
     ) -> None:
-        logger.debug("Signal: output_power_manager set_mode_event")
+        """
+        Blank/unblank outputs via the output power management protocol.
+
+        `_blanked_outputs` keeps track of those that were blanked because we don't want
+        to unblank outputs that were already disabled due to not being part of the
+        user-configured layout.
+        """
+        logger.warning("Signal: output_power_manager set_mode_event")
```
elParaguayo, line 455: Why change from `debug` to `warning`?

elParaguayo:
LGTM other than one very little question.

libqtile/backend/wayland/core.py
m-col: Nice catch - I switched it to `warning` so I could debug without setting logging to debug mode locally. It should be `debug`.

PR #3439, opened 2022-03-20
Update clone url in docs

gelbermann:
Seems that due to recent changes by Github, cloning as described in the docs (with `git clone git://github.com/qtile/qtile.git`) doesn't work anymore. It produces this output:
```
Cloning into 'qtile'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

I updated the docs to to reflect a possible solution. 
I hope this is an acceptable contribution :)

### DIFF
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -95,7 +95,7 @@ With the dependencies in place, you can now install qtile:
 
 .. code-block:: bash
 
-    git clone git://github.com/qtile/qtile.git
+    git clone https://github.com/qtile/qtile.git
     cd qtile
     pip install .
 


m-col:
Nice catch :) thank you for this

elParaguayo:
Sorry. Before we merge this, can I check my understanding:

Cloning via SSH does still work but only if you're already authenticated with GitHub, otherwise it fails.

If that's the case, I don't think we should be telling everyone to use Https instead but, rather, if they can't use SSH.

Thoughts?

m-col:
> If that's the case, I don't think we should be telling everyone to use Https instead but, rather, if they can't use SSH.

Why would SSH be preferable?

elParaguayo:
Fair question.

My understanding was that GitHub recommended SSH over Https (but I believe they may have changed their position over time).

Go ahead and merge this then.

m-col:
Maybe it's my cynicism but the only reason I can think of as to why they might prefer authenticated access over SSH is to identify (and therefore track) clones.

Either way, the https option works for everyone, so it's simpler to document.

elParaguayo:
Works for me. Thanks.

PR #3441, opened 2022-03-21
Fix Launchbar crash when no `progs`

elParaguayo:
Didn't rebase #3340 before merging. #3428 reveals that there's an issue in `Launchbar`'s init where no `progs` are passed (which
shouldn't happen as it defeats the point of the widget...).

Regardless, the test doesn't fail on CI because, due to missing dependencies, the test is skipped for `LaunchBar`.

This PR fixes that issue by providing an empty list as a fallback.

### DIFF
--- a/libqtile/widget/launchbar.py
+++ b/libqtile/widget/launchbar.py
@@ -111,7 +111,7 @@ class LaunchBar(base._Widget):
                         "cmd": prog[1],
                         "comment": prog[2] if len(prog) > 2 else None,
                     }
-                    for prog in config["progs"]
+                    for prog in config.get("progs", list())
                 ]
             )
         )


elParaguayo:
Funny thing is, there was something bugging me about that line when I changed it and I just couldn't work out what it was. Now I know!

m-col:
I thought adding the type hint to `_progs` would trigger mypy for that method, but I think maybe it would need a return annotation. Or for some other reason mypy missed it, but I think mypy would identify that `dict.get(X)` does not always return an iterable.

m-col:
Ah wait, it's because of the square bracket access `config["progs"]` - this is basically asserting that the item _is_ there, because if it isn't a key error will be raised.

PR #3446, opened 2022-03-25
More typing

m-col:
2 commits to add typing to `libqtile.core` and `libqtile.utils`. A few potential bugs and some redundancies were identified and now removed.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -813,7 +813,7 @@ class Qtile(CommandObject):
             if sel is None:
                 return self.current_group.layout
             else:
-                return utils.lget(self.current_group.layouts, sel)
+                return lget(self.current_group.layouts, int(sel))
         elif name == "widget":
             return self.widgets_map.get(sel)  # type: ignore
         elif name == "bar":
@@ -833,7 +833,7 @@ class Qtile(CommandObject):
             if sel is None:
                 return self.current_screen
             else:
-                return utils.lget(self.screens, sel)
+                return lget(self.screens, int(sel))
         elif name == "core":
             return self.core
         return None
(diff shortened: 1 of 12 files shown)


libqtile/utils.py
```diff
@@ -113,50 +101,33 @@ def rgb(x: ColorType) -> tuple[float, float, float, float]:
...
-
     if isinstance(colour, (str, tuple)):
-        return remove_alpha(colour)
-
-    elif isinstance(colour, list):
-        return [remove_transparency(c) for c in colour]
-
-    return (0, 0, 0)
+        return tuple(x * 255.0 for x in rgb(colour)[:3])
+    return [remove_transparency(c) for c in colour]
```
elParaguayo, line 127: Nice. Makes my code look stupid...

libqtile/core/manager.py
```diff
@@ -917,15 +918,15 @@ def cmd_groups(self) -> dict[str, dict[str, Any]]:
         """
         return {i.name: i.info() for i in self.groups}
 
-    def cmd_display_kb(self, *args) -> str:
+    def cmd_display_kb(self) -> str:
```
elParaguayo, line 921: I'm not clear how `args` may have been used in this method but, if it's no longer required, should we still accept it and just provide a deprecation warning rather than having a breaking change?

m-col: I don't see why anybody would have ever used it as it never did anything. TBH I'd like it if CI caught this when it was added, because it's an unused variable. `vulture` would pick it up but we disabled it in CI because it is very verbose, I think `pylint` was possibly also disabled for the same reason, so that's probs why it passed. I was playing around with vulture again recently (trying to find dead code) and it I stand by keeping it disabled in CI, but perhaps we should reassess using pylint in CI - maybe we can configure it to pick up just the little things we might want to avoid (like these unused arguments). Anyway, that's unrelated to this PR, but I think it'd be fine to not bother with deprecating this.

elParaguayo: I'm fine for this to be merged but it might be worth rerunning CI as I think something's gone wrong somewhere so should check this is still ok

PR #3449, opened 2022-03-25
StatusNotifier: better handling of Ayatana

elParaguayo:
Ayatana indicators don't provide their service when registering an item. Instead they provide their object path. The SNI widget deals with this by emplying a message handler to listen for register messages and identifies the service that sent the request. However, the invalid call is still also logged by the widget.

This PR fixes this by checking for InvalidBusName exceptions that are caused when trying to introspect the object path rather than the service and, where this happens, removing the invalid item from the watcher.

This has no impact on the operation of the widget but will reduce error messages in the log.

### DIFF
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -105,6 +105,10 @@ class StatusNotifierItem:  # noqa: E303
             try:
                 introspection = await self.bus.introspect(self.service, self.path)
                 found_path = True
+            except InvalidBusNameError:
+                # This is probably an Ayatana indicator which doesn't provide the service name.
+                # We'll pick it up via the message handler so we can ignore this.
+                return False
             except InvalidObjectPathError:
                 logger.info(f"Cannot find {self.path} path on {self.service}.")
                 if self.path == STATUSNOTIFIER_PATH:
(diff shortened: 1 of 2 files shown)


libqtile/widget/statusnotifier.py
```diff
@@ -508,6 +512,14 @@ def item_added(self, item, service, future):
             if self.on_item_added:
                 self.on_item_added(item)
 
+        # It's an invalid item so let's remove it from the watchers
+        else:
+            for w in self.watchers:
+                try:
+                    w._items.remove(service)
+                except ValueError:
+                    pass
```
m-col, line 521: AFAICT `StatusNotifierHost` has a `watchers` attribute, but `StatusNotifierItem` here doesn't. Is this right?

elParaguayo: This is in `StatusNotifierHost`.

m-col: Oh, lol. github and its code folding. Nevermind!

m-col (approved):
LGTM!

libqtile/widget/statusnotifier.py
elParaguayo: That's caught me out more times than I'd like to admit!

PR #3450, opened 2022-03-26
Fix CI failures

elParaguayo:
EDIT: CI tests are now passing. Updated commit message:

CI is failing with 2 errors:
1) cairocffi is not found
2) mypy errors

It looks like this might be related to an update to setuptools.

While we could pin setuptools, it seems we have been using deprecated methods (e.g. `python setup.py install`) so I prefer to fix the issues.

This PR fixes the CI failures by:
1) Adding cairocffi installation to tox.ini
2) Installing qtile via `pip install .` in tox mypy environment
3) Fixing 2 mypy errors (unreachable code in utils.py)

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -135,8 +135,6 @@ def has_transparency(colour: ColorsType):
     elif isinstance(colour, list):
         return any([has_transparency(c) for c in colour])
 
-    return False
-
 
 def remove_transparency(colour: ColorsType):
     """
@@ -152,8 +150,6 @@ def remove_transparency(colour: ColorsType):
 
     elif isinstance(colour, list):
         return [remove_transparency(c) for c in colour]
-
-    return (0, 0, 0)
 
 
 def scrub_to_utf8(text):
--- a/tox.ini
+++ b/tox.ini
@@ -36,6 +36,7 @@ deps =
     PyGObject
 # pywayland has to be installed before pywlroots
 commands =
+    pip install cairocffi
     pip install pywlroots>=0.15.7,<0.16.0
     python3 setup.py -q install
     {toxinidir}/scripts/ffibuild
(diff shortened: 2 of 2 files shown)


elParaguayo:
Hmmm. `mypy` tests failing?!

m-col:
The mypy version hasn't changed in the env, it still says `mypy == 0.931`. Perhaps it is tox that has changed? I don't think we pin that.

m-col:
I guess not, latest release was Dec 29, 2021

elParaguayo:
The `mypy` failures are these:
```
2022-03-26T08:43:42.2678966Z error: not checking stubs due to mypy build errors:
2022-03-26T08:43:42.2679464Z default_config.pyi:4: error: Skipping analyzing "libqtile.config": module is installed, but missing library stubs or py.typed marker
2022-03-26T08:43:42.2680150Z default_config.pyi:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
2022-03-26T08:43:42.2680737Z default_config.pyi:5: error: Skipping analyzing "libqtile.layout.base": module is installed, but missing library stubs or py.typed marker
2022-03-26T08:43:42.2681276Z checking qtile config file /home/runner/work/qtile/qtile/test/../libqtile/resources/default_config.py
```
But nothing's changed. It's so weird.

elParaguayo:
Installing python 3.9 locally and I am getting the same `mypy` test failures too now.

elParaguayo:
WTF? Same errors on 3.10 but not locally. What's going on in CI?

m-col (approved):
LGTM! Thanks for figuring this out. I'm still not sure I fully understand so presumably setuptools has changed a bit.

elParaguayo:
Yeah - not sure I get it 100% but this works for now so I suggest we merge so we can commit other fixes and then we can look at pyproject.toml etc.

PR #3452, opened 2022-03-26
Add scrolling to `_TextBox` widgets

### LINKED ISSUE #3445: Need to implement better text scrolling
### The issue:

This is to remind me that I need to:
* Allow mpris2 widget to keep scrolling
* Implement pixel scrolling

### Required:

Need to revist #2039 and #1927. 

I think pixel scrolling could be a mixin so, in theory, anything can be scrolled. We can think whether it's added to the base `_TextBox` class or whether there's a separate `ScrollingTextBox` class.

elParaguayo:
**Draft for now - testers wanted**

I've implemented scrolling in `_TextBox` widgets and would be keen for people to test.

The gif is a bit jerky (and loops in the wrong place) but you can see we're doing pixel scrolling.
out

Scrolling also works on vertical bars.

Things to do:
- [x] Testing (how?)
- [x] Replace scrolling in mpris2 widget with this
- [x] Fix bugs people find

Fixes #3445

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -388,6 +388,29 @@ class _TextBox(_Widget):
         ("markup", True, "Whether or not to use pango markup"),
         ("fmt", "{}", "How to format the text"),
         ("max_chars", 0, "Maximum number of characters to display in widget."),
+        (
+            "scroll",
+            False,
+            "Whether text should be scrolled. When True, you must set the widget's ``width``.",
+        ),
+        (
+            "scroll_repeat",
+            True,
+            "Whether text should restart scrolling once the text has ended",
+        ),
+        (
+            "scroll_delay",
+            2,
+            "Number of seconds to pause before starting scrolling and restarting/clearing text at end",
+        ),
+        ("scroll_step", 1, "Number of pixels to scroll with each step"),
+        ("scroll_interval", 0.1, "Time in seconds before next scrolling step"),
+        (
+            "scroll_clear",
+            False,
+            "Whether text should scroll completely away (True) or stop when the end of the text is shown (False)",
+        ),
+        ("scroll_hide", False, "Whether the widget should hide when scrolling has finished"),
     ]  # type: list[tuple[str, Any, str]]
 
     def __init__(self, text=" ", width=bar.CALCULATED, **config):
@@ -395,6 +418,12 @@ class _TextBox(_Widget):
         _Widget.__init__(self, width, **config)
         self.add_defaults(_TextBox.defaults)
         self.text = text
+        self._is_scrolling = False
+        self._should_scroll = False
+        self._scroll_offset = 0
+        self._scroll_queued = False
+        self._scroll_timer = None
+        self._scroll_width = width
 
     @property
     def text(self):
@@ -407,6 +436,9 @@ class _TextBox(_Widget):
         self._text = value
         if self.layout:
             self.layout.text = self.formatted_text
+            if self.scroll:
+                self.check_width()
+                self.reset_scroll()
 
     @property
     def formatted_text(self):
@@ -461,6 +493,26 @@ class _TextBox(_Widget):
             self.fontshadow,
             markup=self.markup,
         )
+        if not isinstance(self._scroll_width, int) and self.scroll:
+            logger.warning("%s: You must specify a width when enabling scrolling.", self.name)
+            self.scroll = False
+
+        if self.scroll:
+            self.check_width()
+
+    def check_width(self):
+        """
+        Check whether the widget needs to have calculated or fixed width
+        and whether the text should be scrolled.
+        """
+        if self.layout.width > self._scroll_width:
+            self.length_type = bar.STATIC
+            self.length = self._scroll_width
+            self._is_scrolling = True
+            self._should_scroll = True
+        else:
+            self.length_type = bar.CALCULATED
+            self._should_scroll = False
 
     def calculate_length(self):
         if self.text:
@@ -481,15 +533,11 @@ class _TextBox(_Widget):
         if not self.can_draw():
             return
         self.drawer.clear(self.background or self.bar.background)
-        if self.bar.horizontal:
-            self.layout.draw(
-                self.actual_padding or 0,
-                int(self.bar.height / 2.0 - self.layout.height / 2.0) + 1,
-            )
-        else:
-            # We need to do some transformations for vertical bars.
-            self.drawer.ctx.save()
 
+        # size = self.bar.height if self.bar.horizontal else self.bar.width
+        self.drawer.ctx.save()
+
+        if not self.bar.horizontal:
             # Left bar reads bottom to top
             if self.bar.screen.left is self.bar:
                 self.drawer.ctx.rotate(-90 * math.pi / 180.0)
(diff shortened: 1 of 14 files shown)


libqtile/widget/base.py
```diff
@@ -388,13 +388,43 @@ class _TextBox(_Widget):
...
+        ("scroll_step", 1, "Number of pixels to scroll with each step"),
+        ("scroll_interval", 0.1, "Time in seconds before next scrolling step"),
+        ("scroll_width", 200, "Width of widget. Text wider than this will scroll."),
+        (
+            "scroll_clear",
+            False,
+            "Whether text should scroll completely away (True) or stop when the end of the text is shown (False)",
+        ),
+        ("reset_scroll_on_update", True, "When new text is provided, the scroll is restarted"),
+        ("hide_text_after_scroll", False, "Whether the widget should hide when scrolling has finised")
```
m-col, line 411: It would help with clarity if all scroll-related settings had the same prefix, like the ones with `scroll_X`. Maybe `enabled_scrolling` can just be `scroll`, then `scroll_repeat`, `scroll_reset_on_update`, `scroll_hide_when_done` or something.

re the settings themselves, are these all settings that you think would be immediately useful (like in test widgets you may have configured), or are you predicting what might be useful? It's just I'm a bit skeptical about the usefulness of e.g. `repeat_scrolling` (wouldn't everybody want `True`?), `reset_scroll_on_update` (same). The role of `scroll_clear` and `hide_text_after_scroll` also seems similar - scroll once, then disappear, just that ones scrolls away rather than vanishes? How do these interact with `repeat_scrolling`, and do the constraints on how they can be used together suggest that the configuration options can be made more concise?

Lastly, is `scroll_width` not redundant with `width`, as only one can be used, but conceptually they mean something similar? Could `width` be used the way `scroll_width` is being used when `enable_scrolling = True`?

m-col (commented):
Awesome, thanks for working on this! Looks cool. Gonna have to give it a proper out.

libqtile/widget/base.py
elParaguayo: Thanks. All good points. 

This is what I started with to get things working but I agree we can probably tidy up a bit.

I'm just playing with mpris2 widget but can refactor after that.

elParaguayo: `scroll_clear`: when `True` this keeps scrolling until all the text has scrolled out of the widget. When `False` the scrolling stops as soon as the end of the text is visible.
`hide_text_after_scroll`: This scrolls once and then hides the text.

elParaguayo: Made some changes. Haven't changed `width` yet. One reason for this is that it doesn't appear in `defaults` and I wanted to make it clearer but I definitely see the value in your approach too.

elParaguayo:
I think I'll split this into 2 PRs to keep the mpris widget changes separate.

elParaguayo:
> Could you make the default value for `scroll_repeat` True?

Yes.

elParaguayo:
Added some tests. Hopefully good to get this one merged now.

libqtile/widget/base.py
```diff
@@ -388,13 +388,42 @@ class _TextBox(_Widget):
         ("markup", True, "Whether or not to use pango markup"),
         ("fmt", "{}", "How to format the text"),
         ("max_chars", 0, "Maximum number of characters to display in widget."),
+        (
+            "scroll",
+            False,
+            "Whether text should be scrolled. When True, you must set the widget's ``width``.",
```
m-col, line 394: "When True, you must set the widget's ``width``" - so if this won't work with flexible widgets. Is that a technical limitation, or does it not make sense conceptually?

elParaguayo: To scroll, you've got to have a fixed/max width, otherwise the widget won't scroll. This is why I split out `width` and `scroll_width` originally.

m-col: > To scroll, you've got to have a fixed/max width, otherwise the widget won't scroll.

But what's the reason for this limitation?

m-col:
I pulled down the PR and tested it with the `Mpd2` widget, and got a configuration erorr complaining that `width` had been passed multiple times. I'm just debugging that now but I wonder what other widgets won't immediately work with this.

libqtile/widget/base.py
```diff
@@ -461,6 +493,26 @@ def _configure(self, qtile, bar):
             self.fontshadow,
             markup=self.markup,
         )
+        if not isinstance(self.scroll_width, int) and self.scroll:
+            logger.warning(f"{self.name}: You must specify a width when enabling scrolling.")
```
m-col, line 497:
```suggestion
            logger.warning("%s: You must specify a width when enabling scrolling.", self.name)
```

m-col:
My error is because `width=bar.CALCULATED` is hardcoded into `ThreadPoolText`, which means this widget can't scroll currently. I suspect this isn't necessary though, maybe we can update `ThreadPoolText` to let it scroll?

libqtile/widget/base.py
elParaguayo: If the text box resizes to the width of the text then there's no point in scrolling. You scroll when the text is wider than the container.

libqtile/widget/base.py
elParaguayo: Yes. Need to remember this is not string formatting, this is logger formatting so it's ok to use `%s`!

libqtile/widget/base.py
m-col: Huh, oh yeah :laughing:

elParaguayo:
> I pulled down the PR and tested it with the `Mpd2` widget, and got a configuration erorr complaining that `width` had been passed multiple times. I'm just debugging that now but I wonder what other widgets won't immediately work with this.

Ah. Interesting. I would hope the tests picked this up but I know `test_widget_init_configure` won't as it just ignores `ConfigErrors` (that's a separate issue that we may want to deal with...)

elParaguayo:
> My error is because `width=bar.CALCULATED` is hardcoded into `ThreadPoolText`, which means this widget can't scroll currently. I suspect this isn't necessary though, maybe we can update `ThreadPoolText` to let it scroll?

We just need to remove `width=bar.CALCULATED`  from `ThreadPoolText`. It actually shouldn't be there as it would always give an error if anyone tried to set the width, regardless of the changes here.

elParaguayo:
Ok. Thing these changes should fix that issue. Would be great if you can check too.

libqtile/widget/base.py
```diff
@@ -615,8 +729,10 @@ class ThreadPoolText(_TextBox):
         ),
     ]  # type: list[tuple[str, Any, str]]
 
-    def __init__(self, text, **config):
-        super().__init__(text, width=bar.CALCULATED, **config)
+    def __init__(self, text, width=bar.CALCULATED, **config):
+        if "width" in config:
+            width = config["width"]
+        super().__init__(text, width=width, **config)
```
m-col, line 735: The width kwarg will take the value of the width key in the `config` dict already:
```py
class Widget:
    def __init__(self, width=10, **config):
        print("width", width, "config", config)


Widget()
Widget(width=0)
Widget(width=0, var=0)
```
prints:
```
width 10 config {}
width 0 config {}
width 0 config {'var': 0}
```
so i dont think we should need anything like 733-4 besides the default argument, that you have added

elParaguayo: Well, what you really need to show is a separate widget inheriting `Widget` without a `width` argument but I think you're still right.

elParaguayo:
```
>>> class Widget:
...   def __init__(self, width=10, **config):
...     print(width, config)
...
>>> class GoodChild(Widget):
...   def __init__(self, **config):
...     Widget.__init__(self, **config)
...
>>> class BadChild(Widget):
...   def __init__(self, **config):
...     Widget.__init__(self, width=10, **config)
...
>>> Widget(width=15)
15 {}
<__main__.Widget object at 0x7fd22a474c70>
>>> GoodChild(width=15)
15 {}
<__main__.GoodChild object at 0x7fd22a4741c0>
>>> BadChild(width=15)
  File "<stdin>", line 3, in __init__
TypeError: __main__.Widget.__init__() got multiple values for keyword argument 'width'
>>>
```

m-col: Ah, gotcha, the duplicated `width=` is only acceptable if it's in the signature, not in any calling code, where the duplicate has to be resolved. Cool. So maybe this needs to do something like `config.setdefault("width", bar.CALCULATED)`

elParaguayo: Don't think we even need that. We just need to stop the child classes (`ThreadPoolText` etc) from setting a default value in their call to `_TextBox.__init__`. If it's just in config then `_TextBox` should handle it ok. That's the `GoodChild` example above.

elParaguayo: Some other widgets have this same issue e.g. `GroupBox`. While these shouldn't have fixed widths, I'm inclined to remove the code that causes the issue.

e.g.
https://github.com/qtile/qtile/blob/5c28cbb283c4397fb001951d6f12d0c5058988ae/libqtile/widget/groupbox.py#L47-L48

elParaguayo:
> Playing around with `Mpris2` I discovered that it isn't getting cleaned up because of its scrolling timer. We have to make sure they are e.g. when config is reloaded

I've actually made some other changes to the Mpris2 widget off the back of this PR. It was originally part of the PR but I thought it may be better to separate it. Part of that change is to clear the display when the widget is finalized and remove timers so that may fix your issue.

Shall I add the commit back to this PR?

elParaguayo:
When you say "always scrolls", do you mean that it repeats when it gets to the end? If so, that's because we set `scroll_repeat` to `True` by default (https://github.com/qtile/qtile/pull/3452#issuecomment-1090315419)

m-col:
I meant that, based on its current (pre this PR) documentation, you cannot configure it to never scroll its text. From a look at the code, setting `scroll_interval` to 0 or None may disable it, but it isn't clear.

elParaguayo:
Pushed the Mpris2 changes.

libqtile/widget/base.py
```diff
@@ -481,15 +533,11 @@ def draw(self):
...
-        if self.bar.horizontal:
-            self.layout.draw(
-                self.actual_padding or 0,
-                int(self.bar.height / 2.0 - self.layout.height / 2.0) + 1,
-            )
-        else:
-            # We need to do some transformations for vertical bars.
-            self.drawer.ctx.save()
 
+        size = self.bar.height if self.bar.horizontal else self.bar.width
```
m-col, line 537:
```suggestion
        size = self.bar.size
```
Looks like `size` is only used in place so may as well use `self.bar.size` inline.

libqtile/widget/windowtabs.py
```diff
@@ -52,7 +52,10 @@ class WindowTabs(base._TextBox):
     ]
 
     def __init__(self, **config):
-        base._TextBox.__init__(self, width=bar.STRETCH, **config)
+        width = config.get("width", bar.STRETCH)
+        if "width" in config:
+            del config["width"]
```
m-col, line 57:
```suggestion
        width = config.pop("width", bar.STRETCH)
```

libqtile/widget/base.py
```diff
@@ -388,13 +388,42 @@ class _TextBox(_Widget):
...
         self.layout = None
         _Widget.__init__(self, width, **config)
         self.add_defaults(_TextBox.defaults)
         self.text = text
+        self.is_scrolling = False
+        self.should_scroll = False
+        self.scroll_offset = 0
+        self.scroll_queued = False
+        self.scroll_timer = None
+        self.scroll_width = width
```
m-col, line 426: It looks like these are only used by `self`, so could we make them private?

m-col (approved):
Brilliant as always! I've put in some tiny suggestions but otherwise LGTM when you're ready!

libqtile/widget/mpris2widget.py
```diff
@@ -86,97 +145,144 @@ async def _config_async(self):
         )
 
         if not subscribe:
-            logger.warning("Unable to add signal receiver for %s.", self.objname)
+            msg = "Unable to add signal receiver for {}.".format(self.objname)
+            logger.warning(msg)
```
m-col, line 149: I think the original form would keep consistency with the rest of the logging. Actually I think flake8 will catch this!

libqtile/widget/windowtabs.py
elParaguayo: Nice. Didn't know that one.

libqtile/widget/mpris2widget.py
elParaguayo: I'm confused. I thought we needed the `%s` style for logging.

m-col: Yep we do. That's what it was before, and this change replaced it with a `"{}".format()` call.

elParaguayo: Ah. Sorry. I thought this was your suggested change but it's actually because I copied some code from an old commit!

elParaguayo:
We can now use this widget to listen to message from all compatible players. The current player's name is also retrieved but, currently, I do nothing with it, other than make it available via `info`.

We could expose it as a field (e.g. `{player}`) that can be added via the string formatting parameters but I left it out as this could be messy when we don't have the player name.

Or we could try to find the player's icon and show that (but that might cause some issue with the scrolling...)

m-col:
Ideally the widget would behave like other text widgets where there is a dictionary of strings to use to for formatting the displayed text. From what I can tell the widget doesn't allow that customisation at all. I don't think any icon would be necessary (though certainly cool... maybe it would be easier to implement as a separate widget in the same module). Either way, probably out of scope for this PR!

elParaguayo:
> Ideally the widget would behave like other text widgets where there is a dictionary of strings to use to for formatting the displayed text. From what I can tell the widget doesn't allow that customisation at all. I don't think any icon would be necessary (though certainly cool... maybe it would be easier to implement as a separate widget in the same module). Either way, probably out of scope for this PR!

Yeah - I did think about having a dictionary for the various metadata fields.

Will leave for another PR!

m-col:
Looks amazing, thanks again! LGTM if you'd like to merge.

PR #3459, opened 2022-03-29
fucus windows with buttons 1 - 3 instead of only button1 in X11

5ila5:
For me it is quite annoying that I cannot fucus windows with button 1 or 2 (https://github.com/qtile/qtile/discussions/3415). This would be my solution.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -772,22 +772,23 @@ class _Window:
         self.window.set_attribute(eventmask=self._window_mask)
 
     def _grab_click(self):
-        # Grab button 1 to focus upon click when unfocussed
+        # Grab buttons 1 - 3  to focus upon click when unfocussed
         for amask in self.qtile.core._auto_modmasks():
-            self.qtile.core.conn.conn.core.GrabButton(
-                True,
-                self.window.wid,
-                EventMask.ButtonPress,
-                xcffib.xproto.GrabMode.Sync,
-                xcffib.xproto.GrabMode.Async,
-                xcffib.xproto.Atom._None,
-                xcffib.xproto.Atom._None,
-                1,
-                amask,
-            )
+            for i in range(1, 4):
+                self.qtile.core.conn.conn.core.GrabButton(
+                    True,
+                    self.window.wid,
+                    EventMask.ButtonPress,
+                    xcffib.xproto.GrabMode.Sync,
+                    xcffib.xproto.GrabMode.Async,
+                    xcffib.xproto.Atom._None,
+                    xcffib.xproto.Atom._None,
+                    i,
+                    amask,
+                )
 
     def _ungrab_click(self):
-        # Ungrab button 1 when focussed
+        # Ungrab buttons 1 - 3 when focussed
         self.qtile.core.conn.conn.core.UngrabButton(
             xcffib.xproto.Atom.Any,
             self.window.wid,


libqtile/backend/x11/window.py
```diff
@@ -772,19 +772,20 @@ def _reset_mask(self):
...
-                self.window.wid,
-                EventMask.ButtonPress,
-                xcffib.xproto.GrabMode.Sync,
-                xcffib.xproto.GrabMode.Async,
-                xcffib.xproto.Atom._None,
-                xcffib.xproto.Atom._None,
-                1,
-                amask,
-            )
+            for i in range(1,3):
```
elParaguayo, line 777: This only does 1 and 2.

libqtile/backend/x11/window.py
```diff
@@ -772,19 +772,20 @@ def _reset_mask(self):
         self.window.set_attribute(eventmask=self._window_mask)
 
     def _grab_click(self):
-        # Grab button 1 to focus upon click when unfocussed
+        # Grab buttons 1 - 3  to focus upon click when unfocussed
```
elParaguayo, line 775: Can you change comment in `_ungrab_click` too.

elParaguayo (changes requested):

m-col:
I was going to ask for a Wayland version too but turns out this is already the case in the Wayland backend. Any cursor buttons focus an unfocussed window below the cursor. Scrolls don't count as buttons unlike in X.

Regardless I think this change makes sense. The X focus logic is quite delicate because it's difficult to test reliably so it would be good to get a couple confirmations and perhaps people using it for a day or two to triple check everything works as expected.

Also - do we have to call GrabButton multiple times or can we grab a bunch of buttons in a single request?

libqtile/backend/x11/window.py
5ila5: done

elParaguayo:
I think we have to grab multiple times. There is an `Any` option but that grabs scroll events too which probably isn't what we want.

5ila5:
> I was going to ask for a Wayland version too but turns out this is already the case in the Wayland backend. Any cursor buttons focus an unfocussed window below the cursor. Scrolls don't count as buttons unlike in X.
> 
> Regardless I think this change makes sense. The X focus logic is quite delicate because it's difficult to test reliably so it would be good to get a couple confirmations and perhaps people using it for a day or two to triple check everything works as expected.
> 
> Also - do we have to call GrabButton multiple times or can we grab a bunch of buttons in a single request?

Used it like this for a few days and did not have any problems

elParaguayo (approved):

elParaguayo:
This is works nicely for me.

I've squashed your commits into a single one to keep our commit history clean.

m-col (approved):
Cool, thanks for this!

FF-AntiK:
Since this is in the latest release now, I have the following question:
Will this also work with modifier keys pressed? For example in VSCode you can "go to definition" by pressing `ctrl` and left click.

elParaguayo:
That should still work. Are you seeing an issue?

5ila5:
> Since this is in the latest release now, I have the following question: Will this also work with modifier keys pressed? For example in VSCode you can "go to definition" by pressing `ctrl` and left click.

If you try to focus an unfocused window in while pressing a modifier key like `ctrl`,`alt`,... the clicked windows does not focus. This can also be a bit annoying.

FF-AntiK:
> > Since this is in the latest release now, I have the following question: Will this also work with modifier keys pressed? For example in VSCode you can "go to definition" by pressing `ctrl` and left click.
> 
> If you try to focus an unfocused window while pressing a modifier key like `ctrl`,`alt`,... the clicked windows does not focus. This can also be a bit annoying.

Hmmh, maybe there is a way to setup the desired behaviour in the config? I could imagine there are users who may not want the window focused on any mouse click no matter what... I guess a lot of users are using `follow_mouse_focus` anyway...

Personally, I find `follow_mouse_focus` annoying because I never type into the window I'm expecting. ;-)

PR #1423, opened 2019-09-05
Automate screenshots

pawamoy:
For issue #1405.

PR #3460, opened 2022-03-29
Add layout screenshots

elParaguayo:
Similar to the widget screenshots, this PR adds (animated) screenshots for our layouts. It also adds a very simple framework for generating more screenshots for new layouts and configuration examples.

Draft for now as there are some bugs that need ironing out (and I need to see if it builds on ReadTheDocs!).

Here's an example of the `MonadThreeCol` layout:

There's also #1423 which was merged but we never really used.

This PR (currently) focuses on the layouts rather than the impacts of their associated commands.

Would it be useful to include animations showing the impact of commands (maybe displaying the command on the screen)? I'm conscious that we could end up with too many images which is probably not ideal.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,6 +21,9 @@ build:
     - xvfb
     - dbus-x11
     - libnotify-bin
+    - scrot
+    - imagemagick
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -169,8 +169,7 @@ genkeyimg:
 	@echo "Keybinding images have been generated."
 
 genwidgetscreenshots:
-	rm -rf screenshots/widgets
-	@echo "Generating screenshots for widgets using pytest fixtures."
+	@echo "Generating screenshots for widgets and layouts using pytest fixtures."
 	pytest -o python_files="ss_*.py" -o python_functions="ss_*" --backend x11 ../test
 	@echo
-	@echo "Generated widget screenshots"
+	@echo "Generated widget and layout screenshots"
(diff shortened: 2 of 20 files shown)


elParaguayo:
OK. So, the docs build (yay) but... there are some issues.
- Font colour - no idea why it's inconsistent here when it's fine locally
- Issues with floating windows (`Floating` layout and maximised window on `VerticalTile`). I have the same issues locally.

Built version is here: https://qtile--3460.org.readthedocs.build/en/3460/ (not sure how long that link remains valid)

0xUnicorn:
It might not be finished, but looking through the build link, as it is, it would help out new users a lot for getting an insight into what the different layouts looks like.

Consider getting it merged. Nice job.

elParaguayo:
We should still do this - just not top priority at the moment.

ervinpopescu:
Hey @elParaguayo, can we still do this? It looks like a nice feature. I rebased on the latest master [here](https://github.com/ervinpopescu/qtile/tree/layout-screenshots), if you wanna take a look and see if you can cherry-pick that one directly. Comparing https://qtile--3460.org.readthedocs.build/en/3460/ with my local docs, there is a weird fontsize issue for the numbers inside the `widcards`. Will open another PR to check if the docs build on RTD and the behavior there.

pawamoy:
I'd be interested in merging my old #1423 stuff into this PR (once the move to MkDocs is validated and merged). I would change my multicolored windows to blue windows with a big number, and transform the screenshot specifications into tests.

elParaguayo:
@pawamoy that would be great. I do want to have layout screenshots in our docs as it will really help new users understand what each layout does. I may also have a commit which I didn't push which displays command names when showing the effect of the commands.

PR #3464, opened 2022-03-31
Reset colours in Chord widget

### LINKED ISSUE #3461: Chord widget doesn't have default background colour
### The issue:

The background colour is set by looking up the chord name in a dictionary. If the name is not in the dictionary then the chord will be displayed with whatever the current background colour is.

The widget needs to be updated to restore colours to defaults.

### Required:

elParaguayo:
Chord widget sets colours based on entry in `chord_colors` dict. If the chord is not in the dict then the colours are not changed. This can lead to undesirable location if the widget needs to occupy a fixed space in the bar but have a default colour when not showing a chord name.

This PR resets the chord's background and foreground colours when a chord is left or when a chord is not in the `chord_colors` dict.

Fixes #3461


m-col (approved):
LGTM

PR #3465, opened 2022-04-01
Better handling of `Notify` widget when dbus-next not installed

elParaguayo:
If a user tries to use the `Notify` widget without `dbus-next` installed, a `FakeManager` is provided. On a subsequent shutdown of qtile, an exception is raised as the `FakeManager` does not have an `unregister` method which is called when the widget is finalised.

This PR adds the `unregister` method to prevent that exception from arising.

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -209,16 +209,3 @@ if has_dbus:
                     logger.exception("Exception in notifier close callback")
 
     notifier = NotificationManager()
-
-else:
-
-    class FakeManager:
-        def __init__(self):
-            logger.warning(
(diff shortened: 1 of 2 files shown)


m-col:
This just extends the plaster pasted over our handling of not having dbus-next installed, and we can probably do better. Perhaps we should remove `FakeManager` completely and set `notifier` to `None`? Clients of `notifier` can then check that themselves. That would be a more 'correct' mechanism.

elParaguayo:
Yeah, that would work too. Leave it with me.

m-col (approved):
Cool yeah this looks more robust and type safe. Thanks. LGTM.

PR #3466, opened 2022-04-01
Delete .coveragerc

elParaguayo:
We're using `coveralls` now so we can lose this.


m-col (approved):
Makes sense, LGTM!

elParaguayo:
So, actually, `coveralls` also uses this file. Deleting it causes our coverage to drop so let's keep it.

PR #3468, opened 2022-04-02
Fix `has_keyboard` AttributeError

elParaguayo:
`Bar` has no default value for `has_keyboard` so, if `Bar.process_key_press` is called, an AttributeError may be raised.

This PR adds a default value which should prevent this error.


m-col (approved):
Cool, makes sense. Does this ever happen? I don't think it should, because it's set before it's accessed, but having a default value is definitely more correct so +1

elParaguayo:
Yeah, I actually see it fairly often in my log and I can't work out why!

PR #3469, opened 2022-04-03
Battery widget show as full when `Not charging`

### LINKED ISSUE #3411: battery widget displays the unknown char when battery is full
### The issue:

When the battery is full, the unknown char, i.e. "?" as I let the default, is displayed instead of the full battery char, which is a full battery emoji in my config.
The source code of that part of `qtile` is quite intricate so I'm not sure how `widget.Battery` gets the battery status, however, to help you:
`acpi` gives me `Discharging` when not charging, `Charging` when charging and `Not charging` when charging with a full battery. Maybe the `Not charging` status is unknown to `qtile`!

`qtile --version`: 0.20.0

### Required:

- [x] I have searched past issues to see if this bug has already been reported.

elParaguayo:
As per #3411, some batteries report their state as `Not charging` when they are full. The widget doesn't recognise this and shows the battery being in an "unknown" state.

While the issue can be solved by using the `unknown_char` parameter, that's undesirable in case a different, unknown state were to arise.

This PR adds a config parameter `full_not_charging` which, when set, will mark the battery as full when the state is `Not charging`.

Fixes #3411

The obvious alternative is to not have a parameter and just check for either wording but I wasn't sure if the `Not charging` status might arise at other times on systems where a full battery is reported as 'Full'.

I'd also like the poster in #3411 to check this (hence draft for now).

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -190,6 +190,12 @@ class _LinuxBattery(_Battery, configurable.Configurable):
             None,
             "Name of file with the current power draw in /sys/class/power_supply/battery_name",
         ),
+        (
+            "full_not_charging",
+            False,
+            "Some batteries report their status as 'Not charging' when full, rather than 'Full'. "
+            "If the widget reports an unknown state when full, try changing this to 'True'.",
+        ),
     ]
 
     filenames = {}  # type: dict
@@ -269,7 +275,7 @@ class _LinuxBattery(_Battery, configurable.Configurable):
     def update_status(self) -> BatteryStatus:
         stat = self._get_param("status_file")[0]
 
-        if stat == "Full":
+        if stat == "Full" or (self.full_not_charging and stat == "Not charging"):
             state = BatteryState.FULL
         elif stat == "Charging":
             state = BatteryState.CHARGING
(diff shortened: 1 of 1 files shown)


libqtile/widget/battery.py
```diff
@@ -190,6 +190,12 @@ class _LinuxBattery(_Battery, configurable.Configurable):
             None,
             "Name of file with the current power draw in /sys/class/power_supply/battery_name",
         ),
+        (
+            "full_not_charging",
+            False,
+            "Some batteries report their status as 'Not charging' when full, rather than 'Full'. "
+            "If the widget reports an unknown state when full, try changing this to 'True'.",
```
m-col, line 197: > If the widget reports an unknown state when full, try changing this to 'True'.

Can we do this for them instead of providing a new option?

elParaguayo: Well, we *could*... My concern is if there's a scenario that the battery reports "Not charging" when it's not actually full. I agree, it would be nice not to have this as an option.

The spec says this:
```
What:		/sys/class/power_supply/<supply_name>/status
Date:		May 2007
Contact:	linux-pm@vger.kernel.org
Description:
		Represents the charging status of the battery. Normally this
		is read-only reporting although for some supplies this can be
		used to enable/disable charging to the battery.

		Access: Read, Write

		Valid values:
			      "Unknown", "Charging", "Discharging",
			      "Not charging", "Full"
```
So "Not charging" should be separate to a "Full" state. I wonder if the "Full" state is returned if the `capacity` is less than a certain level so it's not actually full even though it will no longer accept a charge.

Let me go back to the OP and get them to check the capacity when the charge stops.

PR #3479, opened 2022-04-06
Notify: fix release of dbus service in `finalize`

### LINKED ISSUE #3475: notify-send stops working when Qtile config is reload instead restart, or if widget.Notify is disabled
### The issue:

notify-send does not work anymore if Qtile config is reload, until I restart Qtile. It also will not work anymore and wait if I disable the "widget.Notify". I am not 100% sure if this is a problem caused by Qtile.

1. Whenever I reload Qtile config instead of restarting, the next run of the program `notify-send` will hang and wait. I have to Ctrl+c to end the process. When I then restart Qtile, then the program `notify-send` works as expected.
2. And i just tested to disable the widget.Notify and try again. Now The program always hangs, reload or not. `notify-send` seems not to find anything to send to. When I enable the widget, then it works like described at point 1.

Some system information:

- Your Qtile version (qtile --version): `0.21.1.dev8+gea4462b0`
- Relevant logs from ~/.local/share/qtile/qtile.log: 
`2022-04-05 21:48:41,482 ERROR libqtile notify.py:_release():L183 Could not release org.freedesktop.Notifications.`

```
OS: Manjaro Linux [x86_64]
Kernel: 5.16.14-1-MANJARO
Uptime: 1 day, 17 hours, 36 mins
Packages: 1896 (pacman)[stable], 1 (flatpak)
Shell: zsh 5.8.1
Resolution: 2560x1440 @ 144Hz
DE: KDE Plasma 5.24.3
WM: Qtile 0.21.1.dev8+gea4462b0
Theme: breeze (GruvboxColors) [Plasma], Klaus [GTK2/3/4]
Icons: breeze [Plasma], Gruvbox [GTK2/3/4]
Font: Noto Sans (12pt) [Plasma], Noto Sans (12pt) [GTK2/3/4]
Cursor: default (24px)
Terminal: kitty
CPU: Intel Xeon E3-1230 v3 (8) @ 3.7GHz
GPU: Nvidia GeForce GTX 1070
Memory: 3225MiB / 15947MiB (20%)
Disk (/): 98GB / 219GB (44%)
Locale: de_DE.UTF-8
```

### Required:

elParaguayo:
There is a bug where a user has the `Notify` widget installed and tries to do `reload_config`. The result of the bug is that the notification server is unavailable fter reload.

As the release of the service name during `Notify.finalize()` was an async call, it may not complete until after the config has been reloaded. However, because the server was in place when the config was reloaded, a new server is not started. The `finalize` method then removes the server and the user is left with no server.

This PR fixes the issue by ensuring that the release of the service name is completed synchronously as part of `finalize`.

Fixes #3475

### DIFF
--- a/libqtile/widget/notify.py
+++ b/libqtile/widget/notify.py
@@ -26,7 +26,6 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-import asyncio
 from os import path
 
 from libqtile import bar, pangocffi, utils
@@ -89,6 +88,7 @@ class Notify(base._TextBox):
         )
 
     async def _config_async(self):
+        logger.debug("Registering notify widget")
         await notifier.register(self.update, self.capabilities, on_close=self.on_close)
 
     def set_notif_text(self, notif):
@@ -191,7 +191,14 @@ class Notify(base._TextBox):
             self.invoke()
 
     def finalize(self):
-        asyncio.create_task(self._finalize())
+        # We may need some async calls as part of the finalize call
+        # We run this with `call_soon_threadsafe` as this waits for
+        # the job to finish before continuing. This is important as,
+        # if the config is just reloading, we need to finish deregistering
+        # the notification server before the new Notify widget instance
+        # registers and creates a new server.
+        self.qtile.call_soon_threadsafe(self._finalize)
+        base._TextBox.finalize(self)
 
     async def _finalize(self):
         task = notifier.unregister(self.update)
@@ -202,4 +209,3 @@ class Notify(base._TextBox):
         # complete the finalisation of this widget.
         if task:
             await task
-        base._TextBox.finalize(self)


libqtile/widget/notify.py
```diff
@@ -89,6 +88,7 @@ def _configure(self, qtile, bar):
         )
 
     async def _config_async(self):
+        logger.warning("Registering notify widget")
```
m-col, line 91: Maybe this should be info or debug?

libqtile/widget/notify.py
```diff
@@ -191,7 +191,14 @@ def cmd_invoke(self):
...
 
     def finalize(self):
-        asyncio.create_task(self._finalize())
+        # We may need some async calls as part of the finalize call
+        # We run this with `call_soon_threadsafe` as this waits for
+        # the job to finish before continuing. This is important as,
+        # if the config is just reloading, we need to finish deregistering
+        # the notification server before the new Notify widget instance
+        # registers and creates a new server.
+        self.qtile.call_soon_threadsafe(self._finalize)
```
m-col, line 200: I'm a little surprised that this fixes it. Why/how does this ensure that the `self._finalize` coroutine is executed before config reload?

libqtile/widget/notify.py
elParaguayo: Yeah. That's debug at most!

libqtile/widget/notify.py
elParaguayo: I'm also surprised! From what I could tell when looking into this, this call is synchronous and so blocks until the coroutine has completed.

I tested with lots of log messages to check the sequencing and it does indeed seem to work.

m-col: OK, TIL! Thanks.

m-col (approved):
LGTM

PR #3480, opened 2022-04-06
Allow `qtile cmd-obj` to pass keyword arguments

elParaguayo:
Some exposed commands take keyword arguments so `qtile cmd-obj` should be able to send these.

This would benefit from a solution to #2435...

### DIFF
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -40,6 +40,15 @@ from libqtile.command.interface import IPCCommandInterface
 from libqtile.ipc import Client, find_sockfile
 
 
+def kwarg(value):
+    """Checks if argument is a key=value pair and returns (key, value)."""
+    key, _, val = value.partition("=")
+    if key and val:
+        return (key, val)
+
+    raise argparse.ArgumentTypeError("kwarg should be in key=value format.")
+
+
 def get_formated_info(obj: CommandClient, cmd: str, args=True, short=True) -> str:
     """Get documentation for command/function and format it.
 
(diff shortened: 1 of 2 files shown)


elParaguayo:
I have an idea for type conversion which I'm testing here. Keeping as draft for now.

m-col:
How would this be used? I think we have to make this interface perfect as changing it in the future will be difficult for users to change (e.g. shell scripts). Considering that `-a args` is basically broken, we have some freedom here to make it good.

elParaguayo:
The keyword args is useful for things like `base._TextBox.cmd_set_font` which takes three keyword args: `font`, `fontsize` and `fontshadow`.

The idea here is to be able to do something like `qtile cmd-obj -o widget textbox -f set_font -k fontsize=12`.

I'm also looking at doing type handling by having an `introspect` type method which asks for the signature of the function to be called and will then try to convert the args to the matching types. I need to come back and look at this again but I think having typing like `str | None` could be problematic here.

m-col:
I wonder if we can just use "everything else" as the args, e.g. `qtile cmd-obj -o widget textbox -f set_font -- fontsize=12` (the `--` is not necessary). The full spec would be like:
```py
qtile cmd-obj -o object -f func string_argument
qtile cmd-obj -o object -f func "string_argument"  # equivalent
qtile cmd-obj -o object -f func "string argument"   # quotes should group words as strings
qtile cmd-obj -o object -f func "string argument" "second arg"
qtile cmd-obj -o object -f func "string argument" 3  # We would need a cast for numbers of course
qtile cmd-obj -o object -f func "string argument" something=3  # Then this indicates a keyword arg
```

Thoughts? Open to other ideas of course. I just find the `-x` flags cumbersome.

elParaguayo:
I understand where you're coming from and would love to keep it simple.

My initial concerns with your approach are:
- Edge cases e.g. what if you've got something that looks like a `key=value` pair but is actually meant to be a string?
- Handling other types e.g. bools

I haven't given a lot of thought to this - that's just what sprung to mind.

m-col:
> Edge cases e.g. what if you've got something that looks like a key=value pair but is actually meant to be a string?

I considered this but figured that is unlikely. That's an assumption with no basis though. And it's certainly a limitation. Hmm. Escape the =?

> Handling other types e.g. bools

We know which functions and arguments require/accept bools so that's something we only need to check for those functions, and it may be as simple as checking for the words true/false. Possibly.

elParaguayo:
So we still need to do introspection? If we do that then we can probably be a bit smarter here e.g. if we get a string that looks like `key=val` then we can see if the function expects an arg called `key` and, if so, we assume it's a keyword argument otherwise we treat as a string.

m-col:
Ooh good point. I think that is the best way to get

PR #3487, opened 2022-04-08
Removed hardcoded formatting in the Net widget

SweepingBishops:
Fix for  #3486
        The formatting for the speeds in the Net widget is hard coded.
        This PR fixes that.
        `down`, which was previously a string (eg:'123B'), is split into
        `down` and `down_suffix` (same for `up`).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,8 @@ Qtile x.xx.x, released XXXX-XX-XX:
       - `Window.getsize` has been renamed `Window.get_size` (i.e. merged with the get_size command).
       - `Window.getposition` has been renamed `Window.get_position` (i.e. merged with the get_position command).
       - The `StockTicker` widget `function` option is being deprecated: rename it to `func`.
+      - The formatting of `NetWidget` has changed, if you use the `format` parameter in your config include 
+        `up_suffix`, `total_suffix` and `down_suffix` to display the respective units.
     * features
         - Add ability to set icon size in `LaunchBar` widget.
         - Add 'warp_pointer' option to `Drag` that when set will warp the pointer to the bottom right of
@@ -31,6 +33,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Add ability to spawn the `WidgetBox` widget opened.
         - Add ability to swap focused window based on index, and change the order of windows inside current group
         - Add `move_to_slice` command to move current window to single layout in `Slice` layout
+        - Made the `NetWidget` text formattable.
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -41,7 +41,7 @@ class Net(base.ThreadPoolText):
     defaults = [
         (
             "format",
-            "{interface}: {down} \u2193\u2191 {up}",
+            "{interface}: {down:6.2f}{down_suffix:<2}\u2193\u2191{up:6.2f}{up_suffix:<2}",
             "Display format of down/upload/total speed of given interfaces",
         ),
         (
@@ -123,15 +123,6 @@ class Net(base.ThreadPoolText):
                 }
             return interfaces
 
-    def _format(self, down, down_letter, up, up_letter, total, total_letter):
-        max_len_down = 7 - len(down_letter)
-        max_len_up = 7 - len(up_letter)
-        max_len_total = 7 - len(total_letter)
-        down = "{val:{max_len}.2f}".format(val=down, max_len=max_len_down)
-        up = "{val:{max_len}.2f}".format(val=up, max_len=max_len_up)
-        total = "{val:{max_len}.2f}".format(val=total, max_len=max_len_total)
-        return down[:max_len_down], up[:max_len_up], total[:max_len_total]
-
     def poll(self):
         ret_stat = []
         try:
@@ -144,21 +135,19 @@ class Net(base.ThreadPoolText):
                 down = down / self.update_interval
                 up = up / self.update_interval
                 total = total / self.update_interval
-                down, down_letter = self.convert_b(down)
-                up, up_letter = self.convert_b(up)
-                total, total_letter = self.convert_b(total)
-                down, up, total = self._format(
-                    down, down_letter, up, up_letter, total, total_letter
-                )
+                down, down_suffix = self.convert_b(down)
+                up, up_suffix = self.convert_b(up)
+                total, total_suffix = self.convert_b(total)
                 self.stats[intf] = new_stats[intf]
                 ret_stat.append(
                     self.format.format(
-                        **{
-                            "interface": intf,
-                            "down": down + down_letter,
-                            "up": up + up_letter,
-                            "total": total + total_letter,
-                        }
+                        interface=intf,
+                        down=down,
+                        down_suffix=down_suffix,
+                        up=up,
+                        up_suffix=up_suffix,
+                        total=total,
+                        total_suffix=total_suffix,
                     )
                 )
 
--- a/test/widgets/docs_screenshots/ss_net.py
+++ b/test/widgets/docs_screenshots/ss_net.py
@@ -39,6 +39,10 @@ def widget(monkeypatch):
     [
         {},
         {"format": "{interface}: U {up} D {down} T {total}"},
+        {
+            "format": "{interface}: U {up}{up_suffix} D {down}{down_suffix} T {total}{total_suffix}"
+        },
+        {"format": "{down:.0f}{down_suffix} \u2193\u2191 {up:.0f}{up_suffix}"},
         {"interface": "wlp58s0"},
         {"prefix": "M"},
     ],
--- a/test/widgets/test_net.py
+++ b/test/widgets/test_net.py
@@ -61,7 +61,10 @@ def patch_net(fake_qtile, monkeypatch, fake_window):
 
         # Reload fixes cases where psutil may have been imported previously
         reload(net)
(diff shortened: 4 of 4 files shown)


SweepingBishops:
How do you change this in the docs?

elParaguayo:
It's done here:
https://github.com/qtile/qtile/blob/634a470a12e787a1a00fe7747059fc6a9de0201c/test/widgets/docs_screenshots/ss_net.py#L37-L48

elParaguayo:
Now that you can do the sting formatting to remove decimal places, you could show that in the screenshot.

SweepingBishops:
> Now that you can do the string formatting to remove decimal places, you could show that in the screenshot.

Will do.

SweepingBishops:
` Error: ERROR: InvocationError for command /home/runner/work/qtile/qtile/.tox/pep8/bin/flake8 libqtile bin test '--exclude=test/configs/syntaxerr.py,**/_ffi*.py' (exited with code 1)`
What does this mean?

elParaguayo:
That just means the command failed. The relevant errors were the two lines before that:
```
  test/widgets/test_net.py:64:26: BLK100 Black would make changes.
  test/widgets/docs_screenshots/ss_net.py:41:10: BLK100 Black would make changes.
```

You need to use `black` to format your code. http://docs.qtile.org/en/stable/manual/contributing.html?highlight=black#submit-a-pull-request

SweepingBishops:
Can someone please review this?

elParaguayo:
There's no need to bump this. We review the PRs regularly and will get to this.

SweepingBishops:
> There's no need to bump this. We review the PRs regularly and will get to this.

Oh, ok.

m-col:
Please can you explain a bit how these format values work? What do you put in `down_suffix` or `up_suffix`?

SweepingBishops:
> Please can you explain a bit how these format values work? What do you put in `down_suffix` or `up_suffix`?

Previously, the `up` variable contained the complete value of upload speed, including the unit.
 `up = "{val:{max_len}.2f}".format(val=up, max_len=max_len_up)` and `"up": up + up_letter,` basically hardcode the formatting, so that the widget always shows 2 decimal places, something like `12.50kB`. 
What I've done is separate the numerical value of the upload speed and its unit, so `up` becomes `12.5` and `up_suffix` becomes `kB`. Now, we can change the number of decimal places (or any other formatting) in the `format` parameter.

Eg: If we want we can remove the decimals by `format="{interface}: Up={up:.0f}{up_suffix}| Down={down:.0f}{down_suffix}"`

m-col:
In that example, why would you pass `up_suffix` rather than putting it in the format string?:
```py
format="{interface}: Up={up:.0f}kB | Down={down:.0f}kB"
```
Is this so you can use the default `format` value but just pass in `up_suffix` and `down_suffix`?

elParaguayo:
The suffix can change so you don't want to hard code it in format.

elParaguayo:
The widget has a `_format` method which can probably be removed now. It hardcoded the formatting for these values to 2 decimal places.

https://github.com/qtile/qtile/blob/634a470a12e787a1a00fe7747059fc6a9de0201c/libqtile/widget/net.py#L126-L133

You can remove that but you probably then need to change the default format to include some precision to limit the number of decimal places.

m-col:
Oh sorry I totally misunderstood - I was thinking they were parameters to the widget, not values for the `format` value. Right, gotcha.

OK, yes this all looks good.

m-col (approved):
LGTM but I'll let @elParaguayo double check in case I've missed anything as i'm not super familiar with this widget

libqtile/widget/net.py
```diff
@@ -41,7 +41,7 @@ class Net(base.ThreadPoolText):
     defaults = [
         (
             "format",
-            "{interface}: {down} \u2193\u2191 {up}",
+            "{interface}: {down}{down_suffix} \u2193\u2191 {up}{up_suffix}",
```
elParaguayo, line 44: You need to add some default precision for `down` and `up` as, otherwise, you can get really long floats with 5 or 6 decimal places.

The value you set shouldn't cause the tests to fail.

elParaguayo (changes requested):

libqtile/widget/net.py
SweepingBishops: I've rounded the default to two decimal places, so that the default format doesn't actually change from what it is now.
Is that okay?

m-col: Does that mean that uses of the widget won't have anything change for them at all? because that would be ideal

SweepingBishops: > Does that mean that users of the widget won't have anything change for them at all? because that would be ideal

Yeah, except they can now change it if they want to.

m-col: Perfect!

elParaguayo: > Does that mean that users of the widget won't have anything change for them at all? because that would be ideal

Not *quite*...

The previous version was worded as it was as it fixed the length of the `up` and `down` text to stop the widget resizing, even if the units changed.

That's not being done here. To do it, we'd need to do two things:
- make the speed a fixed length e.g. `{down:5.2f}`
- make the units a fixed length e.g. `{down_unit:<2}` (left aligning)

SweepingBishops: https://user-images.githubusercontent.com/89494022/162620104-983a4532-ce94-41df-b963-2ae5eca6166c.mp4

I don't think it worked like that :thinking:. Your suggested changes also didn't do anything. (The video is from the master branch.)

elParaguayo: Maybe I misremembered. Will check when I'm back home.

elParaguayo: Looking at the code, I'm sure I am right. I think the issue is in the example I gave you. 5 was too few characters as `100.00` is 6 characters. So try `{down:7.2f}`

SweepingBishops: Nope, doesn't work.

elParaguayo: Ah. It will only fix the width if you're using a monospace font...

SweepingBishops: My bad.

elParaguayo: > My bad.

Not at all.

Monospace fonts aren't the default so most people are going to see changing width. However, latest change keeps this consistent with previous defaults.

elParaguayo:
Thanks. This is looking good.

Do you know how to squash all this into a single commit (with a clear title and commit message)? Have a look here: http://docs.qtile.org/en/stable/manual/howto/git.html#i-was-told-to-squash-some-commits

SweepingBishops:
Thanks! Will get to it by tomorrow.

libqtile/widget/net.py
SweepingBishops: Done

CHANGELOG
```diff
@@ -2,6 +2,8 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
+        - The format argument for widget.Net() takes `down` and `down_suffix` instead of just `down`.
+          This facilitates formatting of `down` (the same for `up`).
```
elParaguayo, line 6: I think we need to add a message here that the precision now needs to be stated explicitly in `format`. People who have set their own `format` string just using `{up}` and `{down}` are going to see multiple decimal places appear where they didn't have them before.

I know no one is really going to bother reading the changelog but, when we get the inevitable bug reports, we can point them to this!

SweepingBishops: Done!

CHANGELOG
```diff
@@ -2,6 +2,10 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
+        - The format argument for widget.Net() takes `down` and `down_suffix` instead of just `down`.
+          This facilitates formatting of `down` (the same for `up`). If you are using the `format parameter`,
+          consider setting the precision to avoid seeing a large number of decimals on the widget.
+          Eg: something like {down:.2f} for two decimal points.
```
elParaguayo, line 8:
```suggestion
          This facilitates formatting of `down` (the same for `up`). If you are using the `format` parameter,
          consider setting the precision to avoid seeing a large number of decimals on the widget.
          Eg: `{down:.2f}` for two decimal points.
```

m-col (approved):
LGTM!

elParaguayo:
Wow. Did we miss that this one never merged?

@SweepingBishops can you rebase  onto the current master and push it again?

elParaguayo:
@SweepingBishops can you squash this into a single commit?

SweepingBishops:
Will do sometime this week, am a bit busy now.

SweepingBishops:
@elParaguayo a little help please. I don't understand what's happening. :cry:

elParaguayo:
OK. Some weird things have happened there! 

I'm not sure how the screenshot code has ended up in `tkwindow` nor why the test seems to be showing as an entirely new file.

I'll take a look later on but it may be easiest if I push changes to your branch directly.

elParaguayo:
OK - that should all be tidied up now.

SweepingBishops:
Thanks!

>  push changes to your branch directly.

Just to understand, this means you rearranged the lines to where they were supposed to go and force pushed the commit right?

elParaguayo:
Correct.

elParaguayo (approved):

SweepingBishops:
Tried :cry:

elParaguayo:
👀 Did you pull my forced changes first?

Never mind, I can fix later today.

PR #3503, opened 2022-04-13
Fix psutil current freq value is already in correct precision

jagguli:


SweepingBishops:
Was the widget showing the wrong data this whole time, or did psutil update something?

SweepingBishops (approved):

elParaguayo:
There's a bug in psutil which they are fixing. As far as I'm aware, the distros haven't picked up the update yet.

m-col:
The `psutil` fix is merged but not released. Looking at their tags, they release very infrequently, so maybe instead we can do a version check in the widget ourselves? This will likely be required anyway because it's very possible (likely) users will not have the most recent psutil installed. Qtile has no version bounds for psutil.

elParaguayo:
Nice. Thanks for updating this.

You need to fix the test for this change too as we're only mocking `psutil` in the test environment (this is also why the docs don't build). Can you try changing these lines:
https://github.com/qtile/qtile/blob/53bfe9235f4c67520d7f49ccbcb27312adfcd17e/test/widgets/test_cpu.py#L33-L36

to:
``` python
class MockPsutil(ModuleType):
    __version__ = "5.8.0"

    @classmethod
    def cpu_percent(cls):
        return 2.6
```
i.e. just inserting the `__version__` value.

elParaguayo:
> The `psutil` fix is merged but not released. Looking at their tags, they release very infrequently, so maybe instead we can do a version check in the widget ourselves? This will likely be required anyway because it's very possible (likely) users will not have the most recent psutil installed. Qtile has no version bounds for psutil.

5.9.1 was released on 20 May so should hit Arch soon but other repos may still need this fix.

elParaguayo:
@jagguli Can you do one of the following
 - rebase this onto the latest master and squash into a single commit; or
 - give me access to do this (tick "Allow edits and access to secrets by maintainers" in the PR)

elParaguayo:
Closed in favour of #4059

PR #3509, opened 2022-04-16
WL: set default XWayland cursor image to left_ptr

m-col:
This is equivalent to running `xsetroot -cursor_name left_ptr` at
startup. The X11 backend does this via
`self._root.set_cursor("left_ptr")`. Here we set it too, otherwise the
default cursor that appears when the pointer is above XWayland clients
is an ugly cross.


m-col:
Requires https://github.com/flacjacket/pywlroots/pull/84

m-col:
pywlroots has released with the change so this is now usable (and reviewable)

cr1ogen:
Nice!!!

PR #3511, opened 2022-04-16
A bunch of improvements to logging

m-col:
A few miscellaneous changes to logging:

- log_utils typing
- simplifying where logs go: either to file (regular runtime) or stdout (tests/xephyr) but not both
- remove some redundant code from init_log
- catch and log exceptions raised within wayland callbacks - currently requires https://github.com/flacjacket/pywayland/pull/38
- make all log calls the same format, i.e. don't do string interpolation at the call site, let the logger do it, to avoid unnecessary work

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
+        - If you have a custom startup Python script that you use instead of `qtile start` and run init_log
+          manually, the signature has changed. Please check the source for the updated arguments.
     * features
         - Add ability to draw borders and add margins to the `Max` layout.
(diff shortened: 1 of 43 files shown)


elParaguayo:
Nice.

If you've changed log output to stdout during tests, does that mean we can revert #3359?

I don't know if that's enough for the `caplog` fixture but it would be great if your changes were enough.

m-col:
I don't think it changes the way the logging in the tests works, as that uses a `QueueHandler` which is independent of the `StreamHandler` that prints to stdout. They both will continue to receive log messages. I think using the `QueueHandler` is a good approach. Were you hoping to avoid it if possible?

elParaguayo:
Don't need to avoid it. It's more a case of wondering whether your change would make that one unnecessary but I suspect it doesn't and we do need to keep the `QueueHandler`.

m-col:
That pywayland PR was merged so this is ready

elParaguayo:
LGTM (other than the test failure...)

m-col:
Great thanks. Those were some orphan imports left in some test files so I've removed those and auto-enabled merge.

PR #3518, opened 2022-04-22
Add missing comma without which AllLayoutsMultipleBorders test fails

Blarse:


elParaguayo:
Thanks.

I am a bit confused, though. Why was the test passing previously without this change? I imagine python would have joined those two strings but they would have been an invalid hex colour at that point.

Where did you see the test failing?

Blarse:
Hi,

I am confused as well.
Right now I'm trying to build qtile 0.21.0 for Alt Linux.
The tests were disabled in previous builds, so I only discovered this error now.
We have python 3.10.4

I got the following error for each layout (I can send you the full build log if you want):
```
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/group.py", line 158, in layout_all
    self.layout.layout(normal, screen_rect)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/layout/base.py", line 56, in layout
    self.configure(i, screen_rect)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/layout/bsp.py", line 222, in configure
    client.place(
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/window.py", line 905, in place
    self.paint_borders(bordercolor, borderwidth)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/window.py", line 914, in paint_borders
    self.window.paint_borders(self.depth, color, width, self.width, self.height)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/window.py", line 454, in paint_borders
    gc, xcffib.xproto.GC.Foreground, [self.conn.color_pixel(colors[i])]
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/xcbq.py", line 677, in color_pixel
    pixel = self.screens[0].default_colormap.alloc_color(name).pixel
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/xcbq.py", line 393, in alloc_color
    color = hex(color)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/utils.py", line 117, in hex
    r, g, b, _ = rgb(x)
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/utils.py", line 107, in rgb
    raise ValueError("RGB specifier must be 6 or 8 characters long.")
ValueError: RGB specifier must be 6 or 8 characters long.
2022-04-21 16:39:24,976 ERROR libqtile group.py:layout_all():L160 Exception in layout bsp
  File "/usr/src/RPM/BUILD/qtile-0.21.0/libqtile/backend/x11/xcbq.py", line 387, in alloc_color
    return self.conn.conn.core.AllocNamedColor(self.cid, len(color), color).reply()
  File "/usr/lib/python3/site-packages/xcffib/__init__.py", line 332, in reply
    data = self.conn.wait_for_reply(self.sequence)
  File "/usr/lib/python3/site-packages/xcffib/__init__.py", line 571, in wrapper
    return f(*args)
  File "/usr/lib/python3/site-packages/xcffib/__init__.py", line 658, in wait_for_reply
    self._process_error(error_p[0])
  File "/usr/lib/python3/site-packages/xcffib/__init__.py", line 649, in _process_error
    raise error(buf)
xcffib.xproto.NameError
```

---
Also all of test/test_fakescreen.py and some other tests fail(ERROR) with 'AssertionError: Error launching qtile'
I suspect it is a build environment problem, so I haven't opened an issue yet.

elParaguayo:
This line:
```
ValueError: RGB specifier must be 6 or 8 characters long.
```
is the error I would have expected to see but it's passed on previous CI builds.

Looking at this more closely, this is an error in the test itself. If I add `assert False` to the end so I can see all the output, all those errors are actually still there. The test doesn't have any assertions in it so it keeps going but it does exit slowly which is probably because of the crash that the invalid colour triggers.

Lastly, our colour handling code doesn't accept 3 digit hex values so these all need to be converted to 6 digits i.e. `#000000` etc.

Blarse:
Oh, I see.

I could try to address this issue, right after I figure out how to pass other tests.

elParaguayo:
#3544 (if it gets merged) will mean we can keep the 3 character hex colour values.

m-col:
I've appended a commit to catch the error, but it isn't a general solution to catch backend errors, it's specific to this. When colours are allocated in the x11 colormap, if the passed string is invalid it just gets replaced by bright red.

Ultimately the test is just testing that all layouts can be configured and enabled/disabled with multiple border options without failing. Not sure how helpful that is tbh. A better test would be to screen shot the border, extract the colours that get rendered and compare them to what we expect. Needless to say we don't quite have a way to do that.

Should we just leave it as is?

elParaguayo:
Do you want to delete your commit then and just merge the original fix?

elParaguayo:
No idea why docs are failing here now.

m-col:
We can keep the other commit as it will draw users attention to bad colours and prevent execution from stopping. tbh i think the PR is fine as is

m-col:
Thanks @Blarse!

PR #3528, opened 2022-04-26
window: add an is_visible() method

tych0:
motivation for this is wanting the traverse.py implementation in
qtile-examples to not traverse hidden windows. With this patch, the
following diff does not traverse windows that have been hidden by the
layout. (I'll send a patch to traverse.py when we land whatever the final
API is here.)

diff --git a/dotfiles/qtile/traverse.py b/dotfiles/qtile/traverse.py
index 5b693af..298bc1a 100644
--- a/dotfiles/qtile/traverse.py
+++ b/dotfiles/qtile/traverse.py
@@ -70,7 +70,7 @@ def _focus_window(qtile, dir, axis):
         windows.remove(cur)

     for w in windows:
-        if isinstance(w, Screen) or not w.minimized:
+        if isinstance(w, Screen) or (not w.minimized and w.is_visible()):
             pos = getattr(w, axis) + getattr(w, dim) / 2
             gap = dir * (pos - cur_pos)
             if gap > 5:

A couple of questions:

1. Should we include floating minimization in is_visible()? Seems like
   maybe yes?
2. (from mcol on irc) can we switch these layouts to using minimization
   instead?


m-col:
> Should we include floating minimization in is_visible()? Seems like
> maybe yes?

I think so - I can't yet see any reason why not, and it is the 'obvious' behaviour.

> (from mcol on irc) can we switch these layouts to using minimization
> instead?

This would change the behaviour of these layouts in a way that is obvious to the user. Consider `Max`. Any windows that are not the current window are hidden. If they become necessarily minimized when hidden, some changes will be apparent, such as:

- They will appear in `TaskList` and other 1st or 3rd party pagers as minimized. Presumably they already do appear in these pagers, so maybe this is really only a cosmetic change? Unless a pager exclusively displays minimized windows (seems like a reasonable setup for a bar), then they will start to appear there. Again I'm not 100% how much this is desired/not desired/matters.
- The client is aware of it is minimised vs simply unmapped, and it may behave differently by e.g. displaying a systray icon or something. This might be a minor thing really. It's just the same as if they really were minimised.

I'd lean towards just using minimization in the layouts but I almost expect some people to wish we hadn't made that change.

tych0:
> I think so - I can't yet see any reason why not, and it is the 'obvious' behaviour.

Agreed, I'll make the change.

> The client is aware of it is minimised vs simply unmapped

This is maybe the case I can see where it matters: clients might free buffers and resources when minimized, but not when unmapped. I really have no idea what applications usually do here, but I can imagine it would be slower.

tych0:
So the way I read https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#changing_window_state suggests that our concepts of "minimized" and "hidden" *are* the same. In particular,

"""
    Advice to Implementors

    Clients can select for StructureNotify on their top-level windows to track transitions between Normal and Iconic states. Receipt of a MapNotify event will indicate a transition to the Normal state, and receipt of an UnmapNotify event will indicate a transition to the Iconic state. 
"""

suggests to me they are conceptually the same. There's some other stuff in there about "additional" states being left up to the WM, which is probably what we've done with the "hidden" concept. But it does feel like we could get rid of that.

Seems like we should maybe do that in a different patch, though.

m-col:
On the wayland side there is a difference between being minimized and just unmapped. In both cases they are informed of the change but I think, like you mention, that being minimized implies 'you can clear your buffers now'. I suppose this is akin to x11 where minimization triggers more than just the StructureNotify event, but also an update of `WM_STATE`.

For the sake of caution, perhaps we could find an alternative approach for this that lets us keep these two things distinct? They could be related, e.g. minimized windows are a subset of hidden windows.

tych0:
> For the sake of caution, perhaps we could find an alternative approach for this that lets us keep these two things distinct? They could be related, e.g. minimized windows are a subset of hidden windows.

Isn't that what this implements? Or am I misunderstanding you? The wayland implementation is:

```
self.mapped and not self.minimized
```

which means that minimized windows are a subset of hidden windows.

libqtile/backend/base.py
```diff
@@ -158,6 +158,13 @@ def hide(self) -> None:
     def unhide(self) -> None:
         """Unhide the window"""
 
+    def is_visible(self) -> bool:
+        """Is this window visible (i.e. not hidden)?"""
+        return False
+
+    def cmd_is_visible(self) -> bool:
+        return self.is_visible()
```
m-col, line 166: These will need to be combined when commands get decorated, so maybe we should just combine them here?

libqtile/backend/wayland/window.py
```diff
@@ -745,6 +745,9 @@ def unhide(self) -> None:
         if not self.mapped:
             self.surface.map_event.emit()
 
+    def is_visible(self) -> bool:
+        return self.mapped and not self.minimized
```
m-col, line 749: All minimized windows are not mapped by definition, so `self.mapped` is sufficient in this case. At least that's supposed to be the case so if it isn't then I should fix that.

And as we're in the same class may as well skip the property?

```suggestion
        return self._mapped
```

m-col (commented):
Ah sorry yes, I followed along with the PR without ever reading the patch, my bad :')

With `not self.hidden and not self.minimized` actually I think, if minimized windows are a subset of hidden windows, then if a window is not hidden, it must not be minimized. Right? Same applies to the wayland implementation with s/hidden/mapped

libqtile/backend/base.py
tych0: Yep, done, thanks.

tych0:
> With not self.hidden and not self.minimized actually I think, if minimized windows are a subset of hidden windows, then if a window is not hidden, it must not be minimized. Right?

For x11 at least, they're not the same. You can hide a window, or you can minimize it, or both. Layouts like max hide windows, for example.

tych0:
Ping :-)

m-col:
> For x11 at least, they're not the same. You can hide a window, or you can minimize it, or both. Layouts like max hide windows, for example.

Can you minimize a window without hiding it?

tych0:
> Can you minimize a window without hiding it?

For x11, minimization means setting WM_STATE to the iconic state. Hiding means setting `self.hidden = true` and unmapping it. You could certainly argue that maybe these are the same thing, but in terms of implementation they are different today, and it's not obvious to me that we could unify them.

e.g. an application which goes to the systray in iconic state and frees its window buffers, but when it's unmapped it keeps those buffers alive and doesn't show itself into the systray.

m-col (approved):
OK, all LGTM :)

tych0:
Cool, thanks!

PR #3529, opened 2022-04-26
Makefile: fix 'lint' target

tych0:
The "format" toxenv no longer exists, and seems like it has mostly migrated
to the "codestyle" env. Let's update the lint target to reflect that.


elParaguayo:
Thanks.

PR #3530, opened 2022-04-27
refactor(libqtile): rename border(width) to border_{color,width}

### LINKED ISSUE #2825: Inconsistent use of border/border_color and borderwidth/border_width
# Issue description

Title - some of the widgets are not consistent with other configurables with these config options. See [this comment](https://github.com/qtile/qtile/pull/2675#issuecomment-927275743) for more details.

It would be nice to make these consistent with other configurables, perhaps adding a migration and deprecation notice for a version or two.

# Qtile version

0.18.1

ayntgl:
- The `migrate` script is not tested; ideally, the `test/scripts` directory should have tests for migrate script.

Closes #2825

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
+        - To maintain consistency with other widgets, all of the instances of `border` parameter are renamed to `border_color` and `borderwidth` parameter to `border_width`. To easily to replace the deprecated parameter names, `qtile migrate` may be ran to replace instances of both of these parameters.
     * features
         - Add ability to draw borders and add margins to the `Max` layout.
     * bugfixes
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,10 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
-        - To maintain consistency with other widgets, all of the instances of `border` parameter are renamed to `border_color` and `borderwidth` parameter to `border_width`. To easily to replace the deprecated parameter names, `qtile migrate` may be ran to replace instances of both of these parameters.
+        - To maintain consistency with other widgets, all of the instances of `border`
+          parameter are renamed to `border_color` and `borderwidth` parameter to
+          `border_width`. To easily to replace the deprecated parameter names, 
+          `qtile migrate` may be ran to replace instances of both of these parameters.
     * features
         - Add ability to draw borders and add margins to the `Max` layout.
     * bugfixes
--- a/libqtile/popup.py
+++ b/libqtile/popup.py
@@ -43,7 +43,7 @@ class Popup(configurable.Configurable):
         ("opacity", 1.0, "Opacity of notifications."),
         ("foreground", "#ffffff", "Colour of text."),
         ("background", "#111111", "Background colour."),
-        ("border", "#111111", "Border colour."),
+        ("border_color", "#111111", "Border colour."),
         ("border_width", 0, "Line width of drawn borders."),
         ("font", "sans", "Font used in notifications."),
         ("font_size", 14, "Size of font."),
@@ -87,8 +87,8 @@ class Popup(configurable.Configurable):
         )
         self.layout.layout.set_alignment(pangocffi.ALIGNMENTS[self.text_alignment])
 
-        if self.border_width and self.border:
-            self.win.paint_borders(self.border, self.border_width)
+        if self.border_width and self.border_color:
+            self.win.paint_borders(self.border_color, self.border_width)
 
         self.x = self.win.x
         self.y = self.win.y
@@ -150,7 +150,13 @@ class Popup(configurable.Configurable):
 
     def place(self) -> None:
         self.win.place(
-            self.x, self.y, self.width, self.height, self.border_width, self.border, above=True
+            self.x,
+            self.y,
+            self.width,
+            self.height,
+            self.border_width,
+            self.border_color,
+            above=True,
         )
 
     def unhide(self) -> None:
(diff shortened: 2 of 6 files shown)


m-col:
This is fantastic, migration and all, thank you for this!

My first thought is does this cover every single instance of `borderwidth` and `bordercolor`? A grep would be helpful for catching all of them, e.g. `backend/wayland/window.py` also uses these.

I'm also wondering how disruptive this might be. While some people hopefuly use `qtile migrate`, some don't and may get problems. I wonder how doable a phased deprecation might be, e.g. by using a keyword argument to any `Configurable` class options that catch the old versions and log a deprecation notice, and maybe keeping old versions in any `cmd_` signatures to do the same. Maybe that's overkill/unnecessary? What do you think @elParaguayo?

Lastly, please could you add a note to the changelog under the config breakage section

elParaguayo:
Nice!

My personal feeling on `migrate` scripts is that I don't particularly like them. I'm all in favour of trying to prevent breakages for users but:
- I'm sceptical as to how many users actually use the script
- I have reservations about how effective our scripts are (i.e. do we actually cover all ways a change may be required?)
- Depending on the change, the tests can be complicated to write which is frustrating given the above two points
- It looks like `bowler` is not supported past python 3.8 so may be incompatible with some users' configs.

For those reasons, I prefer the phased deprecation as per @m-col 's post above. We've got examples of this already elsewhere in the codebase.

m-col:
> EDIT: I also appreciate that my comment in the issue that this closes (https://github.com/qtile/qtile/issues/2825) said we needed a migrate script. My views have evolved since then...!

Ha, nice.

> It looks like bowler is not supported past python 3.8 so may be incompatible with some users' configs.

To be precise, bowler is no longer adding language features introduced after 3.8's walrus operator. It does work on more recent python versions and will probably continue to work for quite some time, but won't consider e.g. the newer `match` syntax. But yes, fair point. Agreed - phased deprecation would be good.

elParaguayo:
> Thank you for the feedback, @m-col @elParaguayo! Do you mind if you can point me where exactly example of phased deprecation is in the codebase?

Here's an example where we moved from a positional argument to a new parameter in `defaults`:
https://github.com/qtile/qtile/blob/b7b458073941264b2f5a84c0b49f4ed316679a9b/libqtile/widget/launchbar.py#L88-L103

elParaguayo:
> > It looks like bowler is not supported past python 3.8 so may be incompatible with some users' configs.
> 
> To be precise, bowler is no longer adding language features introduced after 3.8's walrus operator. It does work on more recent python versions and will probably continue to work for quite some time, but won't consider e.g. the newer `match` syntax. But yes, fair point. Agreed - phased deprecation would be good.

Yes - that's fair. It's not that the code doesn't work on later versions but that they won't/can't update the grammar parsing for new features.

libqtile/popup.py
```diff
@@ -67,6 +68,13 @@ def __init__(
         self.add_defaults(Popup.defaults)
         self.qtile = qtile
 
+        if "border" in config:
+            logger.warning(
```
elParaguayo, line 72: Make sure `logger` is imported:
``` python
```

libqtile/widget/groupbox.py
```diff
@@ -50,17 +50,26 @@ def __init__(self, **config):
         self.add_defaults(base.PaddingMixin.defaults)
         self.add_defaults(base.MarginMixin.defaults)
 
+        if "borderwidth" in config:
+            logger.warning(
```
elParaguayo, line 54: Make sure `logger` is imported:
``` python
```
You may need to add this to the other files you changed.

libqtile/widget/groupbox.py
```diff
@@ -36,11 +36,12 @@
 
 from libqtile import bar, hook
 from libqtile.widget import base
+from libqtile.log_utils import logger
```
elParaguayo, line 39:
```suggestion
from libqtile.log_utils import logger
from libqtile.widget import base
```

elParaguayo:
OK - I've just tried this patch and there are a few bugs. I'll post some comments later.

libqtile/popup.py
```diff
@@ -67,6 +68,13 @@ def __init__(
...
+                'The "border" argument is obsolete and deprecated.',
+                'As a substitute, you may use "border_color" argument instead.',
+            )
```
elParaguayo, line 75:
```suggestion
            logger.warning(
                'The "border" argument is obsolete and deprecated. '
                'As a substitute, you should use "border_color" argument instead.'
            )
```
The old `,` at the end of the first line meant you were passing two arguments to `logger.warning` when actually it should be a single string.

Also "may" -> "should".

This should be updated for the other messages.

elParaguayo: We should do tests too to check that the log messages are triggered when using deprecated configs.

An example is here:
https://github.com/qtile/qtile/blob/d827edd7b04c2fa95827c33aad3146794dfc39b2/test/widgets/test_launchbar.py#L31-L38

I don't think you'll need the monkeypatching, just initialise the widget with the deprecated parameters and check `caplog.records` for the messages.

I can help if you get stuck with these.

libqtile/scripts/migrate.py
```diff
@@ -130,6 +138,8 @@ def windowtogroup_groupName_argument(funcname, query):  # noqa: N802
     tile_master_windows_rename,
     threaded_poll_text_rename,
     pacman_to_checkupdates,
+    border_to_border_color,
+    borderwidth_to_border_width,
```
elParaguayo, line 142: If we're adding migrate scripts, we should test them.

Have a look at https://github.com/qtile/qtile/blob/master/test/test_migrate.py for examples.

CHANGELOG
```diff
@@ -2,6 +2,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
     !!! Config breakage !!!
         - lazy.qtile.display_kb() no longer receives any arguments. If you passed it any arguments
           (which were ignored previously), remove them.
+        - To maintain consistency with other widgets, all of the instances of `border` parameter are renamed to `border_color` and `borderwidth` parameter to `border_width`. To easily to replace the deprecated parameter names, `qtile migrate` may be ran to replace instances of both of these parameters.
```
elParaguayo, line 5:
```suggestion
        - To maintain consistency with other widgets, all of the instances of `border`
          parameter are renamed to `border_color` and `borderwidth` parameter to
          `border_width`. To easily to replace the deprecated parameter names, 
          `qtile migrate` may be ran to replace instances of both of these parameters.
```

Splitting up long line.

elParaguayo (changes requested):

Issue #3534, opened 2022-04-29
Libinput configuration with qtile


JoaoDDuarte:
### The issue:

Hi,
I am trying to configure my touchpad in the following way:

```python
from libqtile.log_utils import logger

try:
    from libqtile.backend.wayland import InputConfig

    wl_input_rules = {"type:pointer": InputConfig(tap=True), "type:touchpad": InputConfig(tap=True)}
    logger.warning("worked!")
except ImportError as e:
    logger.exception(e)
    wl_input_rules = None
```
However, when I look at the qtile log, I get the following:

```
2022-04-29 23:24:56,761 WARNING libqtile wayland.py:<module>():L7 worked!
2022-04-29 23:24:57,783 ERROR libqtile hook.py:fire():L391 Error in hook startup_complete
Traceback (most recent call last):
  File "/home/joao/.local/lib/python3.10/site-packages/libqtile/hook.py", line 389, in fire
    i(*args, **kwargs)
  File "/home/joao/.local/lib/python3.10/site-packages/libqtile/backend/wayland/core.py", line 668, in _configure_pending_inputs
    inputs.configure_device(device, self.qtile.config.wl_input_rules)
  File "/home/joao/.local/lib/python3.10/site-packages/libqtile/backend/wayland/inputs.py", line 335,in configure_device
    if handle and libinput.libinput_device_config_tap_get_finger_count(handle) > 0:
AttributeError: 'NoneType' object has no attribute 'libinput_device_config_tap_get_finger_count'
```

I am not quite sure why this is happening, considering that the wl_input_rules are in fact being correctly set.

A sample of my libinput list-devices:
```
Device:           SynPS/2 Synaptics TouchPad
Kernel:           /dev/input/event15
Group:            8
Seat:             seat0, default
Size:             100x68mm
Capabilities:     pointer gesture
Tap-to-click:     disabled
Tap-and-drag:     enabled
Tap drag lock:    disabled
Left-handed:      disabled
Nat.scrolling:    disabled
Middle emulation: disabled
Calibration:      n/a
Scroll methods:   *two-finger edge
Click methods:    *button-areas clickfinger
Disable-w-typing: enabled
Accel profiles:   flat *adaptive
Rotation:         n/a
```

I am using the latest dev version of qtile straight from github.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
[...]

PR #3539, opened 2022-05-01
WL: Check libinput support was built before accessing it

m-col:
This is mostly checked already, where `_configure_pointer` checks
libinput has been imported correctly, but there is another use of the
libinput module before this point which should also check. Without doing
so, confusing AttributeErrors appear because `libinput is None`.

https://github.com/qtile/qtile/issues/3534


elParaguayo:
LGTM

PR #3541, opened 2022-05-01
Add `resume` hook

elParaguayo:
Adds a new hook that is fired when the system wakes from suspend or sleep. The hook relies on systemd's inhibit dbus interface.

The hook can be used to trigger re-polling of widgets etc in users' configs.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -50,7 +50,7 @@ from libqtile.extension.base import _Extension
 from libqtile.group import _Group
 from libqtile.log_utils import logger
 from libqtile.scratchpad import ScratchPad
-from libqtile.utils import get_cache_dir, lget, send_notification
+from libqtile.utils import get_cache_dir, lget, send_notification, subscribe_for_resume_events
 from libqtile.widget.base import _Widget
 
 if TYPE_CHECKING:
@@ -173,6 +173,10 @@ class Qtile(CommandObject):
 
         if self.config.reconfigure_screens:
             hook.subscribe.screen_change(self.cmd_reconfigure_screens)
+
+        # If user wants resume hooks we need to add a dbus rule
+        if "resume" in hook.subscriptions:
+            subscribe_for_resume_events()
 
         if initial:
             hook.fire("startup_complete")
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -49,6 +49,7 @@ try:
 except ImportError:
     has_dbus = False
 
+from libqtile import hook
 from libqtile.log_utils import logger
 
 
@@ -471,8 +472,35 @@ async def find_dbus_service(service: str, session_bus: bool) -> bool:
         return False
 
     bus.disconnect()
-    del bus
 
     names = msg.body[0]
 
     return service in names
+
+
+def subscribe_for_resume_events() -> None:
+    task = asyncio.create_task(
+        add_signal_receiver(
+            on_resume,
+            session_bus=False,
+            dbus_interface="org.freedesktop.login1.Manager",
+            bus_name="org.freedesktop.login1",
+            signal_name="PrepareForSleep",
+            check_service=True,
+        )
+    )
+    task.add_done_callback(_resume_callback)
+
+
+def _resume_callback(task: asyncio.Task) -> None:
+    if not task.result():
+        logger.warning("Unable to subscribe to system resume events")
+
+
+# We need to ignore typing here. msg is a dbus_next.Message object but dbus_next may not be installed
+def on_resume(msg):  # type: ignore
+    sleeping = msg.body[0]
+
+    # Value is False when the event is over i.e. the machine has woken up
+    if not sleeping:
+        hook.fire("resume")
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
     T = TypeVar("T")
 
 try:
-    from dbus_next import Message, Variant
+    from dbus_next import AuthError, Message, Variant
     from dbus_next.aio import MessageBus
     from dbus_next.constants import BusType, MessageType
 
@@ -278,7 +278,8 @@ async def _notify(
 
     # a new bus connection is made each time a notification is sent so
     # we disconnect when the notification is done
-    bus.disconnect()
+    if bus:
+        bus.disconnect()
 
 
 def guess_terminal(preference: str | Sequence | None = None) -> str | None:
@@ -353,7 +354,7 @@ async def _send_dbus_message(
     member: str,
     signature: str,
     body: Any,
-) -> tuple[MessageBus, Message | None]:
+) -> tuple[MessageBus | None, Message | None]:
     """
     Private method to send messages to dbus via dbus_next.
 
@@ -367,7 +368,11 @@ async def _send_dbus_message(
     if isinstance(body, str):
         body = [body]
 
-    bus = await MessageBus(bus_type=bus_type).connect()
+    try:
+        bus = await MessageBus(bus_type=bus_type).connect()
+    except (AuthError, Exception):
+        logger.warning("Unable to connect to dbus.")
+        return None, None
 
     msg = await bus.call(
         Message(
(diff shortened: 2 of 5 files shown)


SweepingBishops:
A feature I didn't know I wanted till I saw it :grin:.
I noticed that this hook also runs on hibernate.
I actually prefer it that way. Anything I would want to run after a suspend, I would want to run after a hibernate too.
Maybe just add hibernate to the docstring?

ramnes:
Shouldn't you check that systemd and dbus are available? What if they aren't?

elParaguayo:
> Shouldn't you check that systemd and dbus are available? What if they aren't?

The call to `add_signal_receiver` should fail in that case. It will then return `False` which will trigger these lines:
``` python
    def _wake_callback(self, task):
        if not task.result():
            logger.warning("Unable to subscribe to wake events")
```

elParaguayo:
**Note to self:** I need to add some tidying up steps here. If a user regularly restarts/reloads their config we will end up with multiple `AddMatch` rules being registered with dbus.

It's not specific to this PR so I'll deal with it separately.

elParaguayo:
Should we have the hook pass an argument to the callback so we can differentiate between sleep and suspend/hibernate?

libqtile/core/manager.py
```diff
@@ -174,9 +174,38 @@ def load_config(self, initial: bool = False) -> None:
...
+    def on_wake(self, msg):  # type: ignore
+        # We only want the PrepareForSleep or PrepareForSuspend signals
+        if not msg.member.startswith("PrepareFor"):
+            return
+
+        sleeping = msg.body[0]
+
+        # Value is False when the event is over i.e. the machine has woken up
+        if not sleeping:
+            hook.fire("wake_up")
```
m-col, line 207: There isn't much use of `self` here. Could you move this all into `utils`?

libqtile/core/manager.py
```diff
@@ -174,9 +174,38 @@ def load_config(self, initial: bool = False) -> None:
...
+                    session_bus=False,
+                    dbus_interface="org.freedesktop.login1.Manager",
+                    bus_name="org.freedesktop.login1",
+                    check_service=True,
+                )
+            )
+            task.add_done_callback(self._wake_callback)
+
         if initial:
             hook.fire("startup_complete")
```
m-col, line 191: Is there any reason we must create the task before the startup_complete hook fires, or could you move it after? That way, if anybody is binding hooks within startup_complete, they will work.

libqtile/utils.py
```diff
@@ -424,9 +442,37 @@ async def add_signal_receiver(
...
+        "",
+        [],
+    )
+
+    if bus is None or msg is None or (msg and msg.message_type != MessageType.METHOD_RETURN):
+        logger.warning("Unable to send lookup call to dbus.")
+        return False
+
+    bus.disconnect()
+    del bus
```
m-col, line 474: why do we need to del bus?

libqtile/core/manager.py
elParaguayo: No reason at all.

m-col: One reason might be if there is some state detectable from within the startup_complete hook that relates to whether or not that task has been sent. Not that I can think why anybody might be digging into that though!

libqtile/utils.py
elParaguayo: I thought dbus-next had a `__del__` method for disconnecting so had that in first but realised I was wrong and I should really `disconnect()` instead. We can remove the `del` statement.

m-col:
> Should we have the hook pass an argument to the callback so we can differentiate between sleep and suspend/hibernate?

Yeah makes sense. Also going into and coming out of? I wonder if there is some kind of system for this signalling already. IIRC systemd wakeup scripts are passed 'pre' or 'post' as argumants. That's not super elegant, though maybe makes sense for a shell script. Can we do better? Or maybe that's enough? i.e. pass in one of `["sleep", "suspend", "hibernate"]` as arg 1 and as arg 2 one of `["pre", "post"]`. Or concatenate those strings and just pass in one e.g. `"pre_suspend"`. String arguments feel wrong to me but I do wonder if they are more user-friendly than using an enum.

libqtile/core/manager.py
elParaguayo: Yes - that would be neater. Let me redo it.

elParaguayo:
> > Should we have the hook pass an argument to the callback so we can differentiate between sleep and suspend/hibernate?
> 
> Yeah makes sense. Also going into and coming out of? I wonder if there is some kind of system for this signalling already. IIRC systemd wakeup scripts are passed 'pre' or 'post' as argumants. That's not super elegant, though maybe makes sense for a shell script. Can we do better? Or maybe that's enough? i.e. pass in one of `["sleep", "suspend", "hibernate"]` as arg 1 and as arg 2 one of `["pre", "post"]`. Or concatenate those strings and just pass in one e.g. `"pre_suspend"`. String arguments feel wrong to me but I do wonder if they are more user-friendly than using an enum.

We haven't implemented the `pre` events here as I've not added any of the inhibitor code. Currently, we're only looking at the waking up side. We could have a `pre` hook and set an inhibitor in that scenario that's released once the callback has finished. Worth implementing?

Also, looking at the docs, the signals don't differentiate between sleep, suspend and hibernate. You get shutdown events (reboot, power off) and sleep (suspend, hibernate).

libqtile/core/manager.py
elParaguayo: My logic was simpler: `startup_complete` should be called once everything has been set up. Therefore, any steps for adding hooks, which are part of starting up, should be done before that hook is fired.

m-col: That makes a lot of sense, but puts a lot of faith in users :')

m-col:
> We haven't implemented the pre events here as I've not added any of the inhibitor code. Currently, we're only looking at the waking up side. We could have a pre hook and set an inhibitor in that scenario that's released once the callback has finished. Worth implementing?

Hmm, I guess it isn't worth implementing if nobody wants to use it. Maybe leave it for now? The risk however might be changing the signature of this hook to pass that information. Though we can solve that issue by having a completely separate hook for pre, and keep `wake_up` for post. So maybe leave it just as is with post?

> Also, looking at the docs, the signals don't differentiate between sleep, suspend and hibernate. You get shutdown events (reboot, power off) and sleep (suspend, hibernate).

OK good to know - hopefully not an issue!

elParaguayo:
Also a post-reboot/restart hook is pointless as it's the same as `startup_complete/once`!

SweepingBishops (commented):
Maybe change the `CHANGELOG` to make it clearer that it is a user subscribable hook rather than  
something internal? Also wouldn't it be better to separate the hook and the parts that check for a dbus service into separate PRs? Looks fine to me otherwise. :+1:

m-col (approved):
I wonder if we can make the log warnings more actionable for the user. If someone doesn't have dbus-next installed but tries to use this, they'll see "dbus-next is not installed. Unable to subscribe to signals". I think this can only be notifications or this hook, so with only 2 options we might be able to get away with making it say "dbus-next is not installed. Unable to subscribe suspend/notification to signals". With it installed, and it fails, the warning message is more explicit (though not actionable, however idk why it would fail so maybe raising an issue is preferable in that case). What do you think?

Otherwise code LGTM :)

elParaguayo:
> Maybe change the `CHANGELOG` to make it clearer that it is a user subscribable hook rather than something internal? 

All hooks are user subscribable already but I could make it a bit clearer.

> Also wouldn't it be better to separate the hook and the parts that check for a dbus service into separate PRs (maybe move to #3543)?

I think it's fine here as this PR needs it. The commit history is more important than the PRs in any case and I've kept that change as a separate commit so any breakages that it may cause can be reverted more easily!

elParaguayo:
> I wonder if we can make the log warnings more actionable for the user. If someone doesn't have dbus-next installed but tries to use this, they'll see "dbus-next is not installed. Unable to subscribe to signals". I think this can only be notifications or this hook, so with only 2 options we might be able to get away with making it say "dbus-next is not installed. Unable to subscribe suspend/notification to signals". 

Good point.

> With it installed, and it fails, the warning message is more explicit (though not actionable, however idk why it would fail so maybe raising an issue is preferable in that case). What do you think?

It really shouldn't fail and the message is really there to say something's not going to work the way you want it to.

> Otherwise code LGTM :)

👍

elParaguayo:
Should we change the hook name? `resume`?

m-col:
I'm happy with `wake_up` but up to you!

elParaguayo:
Actually, the "dbus-next is not installed. Unable to subscribe to signals" message is part of `add_signal_receiver`. That code is used elsewhere (`StatusNotifier`, `Mpris2` etc.) so we can't really change that code. However, when that message is provided, the code returns `False` which will result in users seeing the "Unable to subscribe to wake events" message.

elParaguayo:
> I'm happy with `wake_up` but up to you!

I changed it to `resume`. I think that's more consistent with how people think of it.

PR #3543, opened 2022-05-02
Remove dbus match rules on shutdown/restart/reload

elParaguayo:
Various object create match rules on the dbus server (e.g. SNI and Mpris2 widgets). These rules should be removed when qtile shuts down, restarts or reloads the config to eliminate redundant messages being sent by the dbus server.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -55,7 +55,7 @@ from libqtile.log_utils import logger
 from libqtile.resources.sleep import inhibitor
 from libqtile.scratchpad import ScratchPad
 from libqtile.scripts.main import VERSION
-from libqtile.utils import cancel_tasks, get_cache_dir, lget, send_notification
+from libqtile.utils import cancel_tasks, get_cache_dir, lget, remove_dbus_rules, send_notification
 from libqtile.widget.base import _Widget
 
 if TYPE_CHECKING:
@@ -298,6 +298,7 @@ class Qtile(CommandObject):
         self.groups_map.clear()
         self.groups.clear()
         self.screens.clear()
+        remove_dbus_rules()
         self.load_config()
 
     def _finalize_configurables(self) -> None:
@@ -324,6 +325,7 @@ class Qtile(CommandObject):
 
     def finalize(self) -> None:
         self._finalize_configurables()
+        remove_dbus_rules()
         inhibitor.stop()
         cancel_tasks()
         self.core.finalize()
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -52,6 +52,8 @@ except ImportError:
 
 from libqtile.log_utils import logger
 
+dbus_bus_connections = set()
+
 # Create a list to collect references to tasks so they're not garbage collected
 # before they've run
 TASKS: list[asyncio.Task[None]] = []
@@ -424,25 +426,32 @@ async def _send_dbus_message(
     signature: str,
     body: Any,
     negotiate_unix_fd: bool = False,
+    bus: MessageBus | None = None,
 ) -> tuple[MessageBus | None, Message | None]:
     """
     Private method to send messages to dbus via dbus_next.
 
+    An existing bus connection can be passed, if left empty, a new
+    bus connection will be created.
+
     Returns a tuple of the bus object and message response.
     """
-    if session_bus:
-        bus_type = BusType.SESSION
-    else:
-        bus_type = BusType.SYSTEM
+    if bus is None:
+        if session_bus:
+            bus_type = BusType.SESSION
+        else:
+            bus_type = BusType.SYSTEM
+
+        try:
+            bus = await MessageBus(
+                bus_type=bus_type, negotiate_unix_fd=negotiate_unix_fd
+            ).connect()
+        except (AuthError, Exception):
+            logger.warning("Unable to connect to dbus.")
+            return None, None
 
     if isinstance(body, str):
         body = [body]
-
-    try:
-        bus = await MessageBus(bus_type=bus_type, negotiate_unix_fd=negotiate_unix_fd).connect()
-    except (AuthError, Exception):
-        logger.warning("Unable to connect to dbus.")
-        return None, None
 
     # Ignore types here: dbus-next has default values of `None` for certain
     # parameters but the signature is `str` so passing `None` results in an
@@ -458,6 +467,11 @@ async def _send_dbus_message(
             body=body,
         )
     )
+
+    # Keep details of bus connections so we can close them on exit
+    # dbus_bus_connetions is a set so we don't need to worry about
+    # duplicates
+    dbus_bus_connections.add(bus)
 
     return bus, msg
 
@@ -503,6 +517,8 @@ async def add_signal_receiver(
 
     rule = ",".join("{}='{}'".format(k, v) for k, v in match_args.items() if v)
 
+    logger.debug("Adding dbus match rule: %s", rule)
+
     bus, msg = await _send_dbus_message(
         session_bus,
         MessageType.METHOD_CALL,
@@ -511,7 +527,7 @@ async def add_signal_receiver(
         "/org/freedesktop/DBus",
         "AddMatch",
         "s",
-        rule,
+        [rule],
     )
 
     # Check if message sent successfully
@@ -548,3 +564,18 @@ async def find_dbus_service(service: str, session_bus: bool) -> bool:
     names = msg.body[0]
 
     return service in names
+
+
+def remove_dbus_rules() -> None:
+    # Disconnecting the bus connections is enough to remove the match rules.
+    while dbus_bus_connections:
+        bus = dbus_bus_connections.pop()
+        try:
+            bus.disconnect()
+        except OSError:
+            # Socket has already shut down
+            pass
+
+        # We need to manually close the socket until https://github.com/altdesktop/python-dbus-next/pull/148
+        # gets merged. There's no error on multiple calls to 'close()'.
+        bus._sock.close()
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -508,14 +508,14 @@ async def add_signal_receiver(
             return False
 
     match_args = {
-        "type": "signal",
         "sender": bus_name,
         "member": signal_name,
         "path": path,
         "interface": dbus_interface,
     }
 
-    rule = ",".join("{}='{}'".format(k, v) for k, v in match_args.items() if v)
+    rule = "type='signal',"
+    rule += ",".join("{}='{}'".format(k, v) for k, v in match_args.items() if v)
 
     logger.debug("Adding dbus match rule: %s", rule)
 
@@ -532,7 +532,15 @@ async def add_signal_receiver(
 
     # Check if message sent successfully
     if bus and msg and msg.message_type == MessageType.METHOD_RETURN:
-        bus.add_message_handler(callback)
+
+        def signal_callback_wrapper(msg: Message) -> None:
+            """Custom wrapper to only run callback if message matches our rule."""
+            if msg.message_type == MessageType.SIGNAL and msg._matches(
+                **{k: v for k, v in match_args.items() if v}
+            ):
+                callback(msg)
+
+        bus.add_message_handler(signal_callback_wrapper)
         return True
 
     else:
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -29,6 +29,7 @@ import string
 from typing import TYPE_CHECKING
 
 from dbus_next import Message, Variant
+from dbus_next.aio import MessageBus
 from dbus_next.constants import MessageType
 
 from libqtile.command.base import expose_command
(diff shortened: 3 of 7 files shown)


libqtile/utils.py
```diff
@@ -420,13 +424,37 @@ async def add_signal_receiver(
...
     if msg and msg.message_type == MessageType.METHOD_RETURN:
         bus.add_message_handler(callback)
+        dbus_match_calls.append((bus, rule))
         return True
 
     else:
         return False
+
+
+async def remove_signal_receiver(bus: MessageBus, rule: str) -> bool:
```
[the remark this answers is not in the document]
m-col: Hmm, good question. Probably dbus-next type stubs in the `stubs` folder. Check out stubgen (`stubgen -m dbus_next`) which can generate them for you.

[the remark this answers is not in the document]
elParaguayo: Need to do some reading about stubs. This generated lots of mypy errors:
```
2022-05-02T20:39:18.2849235Z stubs/dbus_next/_private/constants.pyi:1: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.5896114Z stubs/dbus_next/_private/marshaller.pyi:2: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.5943587Z stubs/dbus_next/signature.pyi:3: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.5944085Z stubs/dbus_next/message.pyi:7: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.5944566Z stubs/dbus_next/errors.pyi:4: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.6802682Z stubs/dbus_next/_private/unmarshaller.pyi:6: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.6829495Z stubs/dbus_next/_private/address.pyi:3: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.7056791Z stubs/dbus_next/_private/util.pyi:5: error: Name "Tuple" is not defined
2022-05-02T20:39:18.7057337Z stubs/dbus_next/_private/util.pyi:5: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Tuple")
2022-05-02T20:39:18.7113578Z stubs/dbus_next/introspection.pyi:6: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:18.7114473Z stubs/dbus_next/introspection.pyi:15: error: Self argument missing for a non-static method (or an invalid type for self)
2022-05-02T20:39:18.7115086Z stubs/dbus_next/introspection.pyi:33: error: Self argument missing for a non-static method (or an invalid type for self)
2022-05-02T20:39:19.0616446Z stubs/dbus_next/service.pyi:6: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:19.0616975Z stubs/dbus_next/proxy_object.pyi:8: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:19.0617455Z stubs/dbus_next/message_bus.pyi:12: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:19.0618096Z stubs/dbus_next/glib/proxy_object.pyi:16: error: Return type "List[ProxyObject]" of "get_children" incompatible with return type "List[BaseProxyObject]" in supertype "BaseProxyObject"
2022-05-02T20:39:19.0618830Z stubs/dbus_next/aio/proxy_object.pyi:17: error: Return type "List[ProxyObject]" of "get_children" incompatible with return type "List[BaseProxyObject]" in supertype "BaseProxyObject"
2022-05-02T20:39:19.0619433Z stubs/dbus_next/glib/message_bus.pyi:9: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:19.0620078Z stubs/dbus_next/glib/message_bus.pyi:53: error: Argument 3 of "get_proxy_object" is incompatible with supertype "BaseMessageBus"; supertype defines the argument type as "Union[Node, str, Element]"
2022-05-02T20:39:19.0620683Z stubs/dbus_next/glib/message_bus.pyi:53: note: This violates the Liskov substitution principle
2022-05-02T20:39:19.0621652Z stubs/dbus_next/glib/message_bus.pyi:53: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
2022-05-02T20:39:19.0622217Z stubs/dbus_next/aio/message_bus.pyi:10: error: Module "_typeshed" has no attribute "Incomplete"
2022-05-02T20:39:19.0622747Z stubs/dbus_next/aio/message_bus.pyi:33: error: Signature of "introspect" incompatible with supertype "BaseMessageBus"
2022-05-02T20:39:19.0623216Z stubs/dbus_next/aio/message_bus.pyi:33: note:      Superclass:
2022-05-02T20:39:19.0623944Z stubs/dbus_next/aio/message_bus.pyi:33: note:          def introspect(self, bus_name: str, path: str, callback: Callable[[Optional[Node], Optional[Exception]], None]) -> Any
2022-05-02T20:39:19.0624742Z stubs/dbus_next/aio/message_bus.pyi:33: note:      Subclass:
2022-05-02T20:39:19.0625421Z stubs/dbus_next/aio/message_bus.pyi:33: note:          def introspect(self, bus_name: str, path: str, timeout: float = ...) -> Coroutine[Any, Any, Node]
2022-05-02T20:39:19.0626001Z stubs/dbus_next/aio/message_bus.pyi:34: error: Signature of "request_name" incompatible with supertype "BaseMessageBus"
2022-05-02T20:39:19.0626443Z stubs/dbus_next/aio/message_bus.pyi:34: note:      Superclass:
2022-05-02T20:39:19.0627641Z stubs/dbus_next/aio/message_bus.pyi:34: note:          def request_name(self, name: str, flags: NameFlag = ..., callback: Optional[Callable[[Optional[RequestNameReply], Optional[Exception]], None]] = ...) -> Any
2022-05-02T20:39:19.0628211Z stubs/dbus_next/aio/message_bus.pyi:34: note:      Subclass:
2022-05-02T20:39:19.0628893Z stubs/dbus_next/aio/message_bus.pyi:34: note:          def request_name(self, name: str, flags: NameFlag = ...) -> Coroutine[Any, Any, RequestNameReply]
2022-05-02T20:39:19.0629460Z stubs/dbus_next/aio/message_bus.pyi:35: error: Signature of "release_name" incompatible with supertype "BaseMessageBus"
2022-05-02T20:39:19.0629924Z stubs/dbus_next/aio/message_bus.pyi:35: note:      Superclass:
2022-05-02T20:39:19.0630699Z stubs/dbus_next/aio/message_bus.pyi:35: note:          def release_name(self, name: str, callback: Optional[Callable[[Optional[ReleaseNameReply], Optional[Exception]], None]] = ...) -> Any
2022-05-02T20:39:19.0631224Z stubs/dbus_next/aio/message_bus.pyi:35: note:      Subclass:
2022-05-02T20:39:19.0631827Z stubs/dbus_next/aio/message_bus.pyi:35: note:          def release_name(self, name: str) -> Coroutine[Any, Any, ReleaseNameReply]
2022-05-02T20:39:19.0632481Z stubs/dbus_next/aio/message_bus.pyi:38: error: Argument 3 of "get_proxy_object" is incompatible with supertype "BaseMessageBus"; supertype defines the argument type as "Union[Node, str, Element]"
2022-05-02T20:39:19.0633083Z stubs/dbus_next/aio/message_bus.pyi:38: note: This violates the Liskov substitution principle
2022-05-02T20:39:19.0633798Z stubs/dbus_next/aio/message_bus.pyi:38: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

I generated stubs by doing `stubgen -p dbus_next -o stubs` but maybe it's quicker to do this manually for the bits we need.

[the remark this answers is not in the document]
elParaguayo: Maybe we don't need stubs here. `dubs-next` is a dependency of the `mypy` environment in tox which is how people should be running tests locally anyway.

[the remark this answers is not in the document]
m-col: Just rethinking this... we only need stubs if the library's source isn't typed with a [py.typed](https://github.com/altdesktop/python-dbus-next/blob/master/dbus_next/py.typed) file. Without the library installed, `mypy` will emit a message saying that type information cannot be found and that the user should installed the library or get some stubs. In this case, the better approach is to just install the library because that's where the true type information is. So back to your question:

> What's the right approach in these scenarios?

mypy will handle telling the user what to do - they should install the library.

libqtile/utils.py
```diff
@@ -420,13 +424,37 @@ async def add_signal_receiver(
...
+            destination="org.freedesktop.DBus",
+            interface="org.freedesktop.DBus",
+            path="/org/freedesktop/DBus",
+            member="RemoveMatch",
+            signature="s",
+            body=[rule],
+        )
+    )
+
+    bus.disconnect()
```
elParaguayo, line 454: This may be a bad idea as there could be multiple match rules per bus. So we either want a dictionary with the bus as the key and a list of rules. Or... is disconnecting the bus sufficient to delete the remove the rules?

elParaguayo:
We do need to implement this fix. Currently, dbus connections are not being tidied up when configs are reloaded so we end up with more and more bus connections.

However, I've also simplified the approach as I think all that's needed is to close the bus connections rather than sending a `RemoveMatch` request.

libqtile/core/manager.py
```diff
@@ -382,6 +390,12 @@ def _process_screens(self, reloading: bool = False) -> None:
 
         self.screens = screens
 
+    def remove_dbus_rules(self) -> None:
+        # Disconnecting the bus connections is enough to remove the match rules.
+        while dbus_bus_connections:
+            bus = dbus_bus_connections.pop()
+            bus.disconnect()
```
m-col, line 397: As this method doesn't use `self`, could we move it into `utils` as a standalone function?

m-col (commented):
Does this solution fix it? From what I can tell reading dbus-next, `disconnect()` shuts down the socket but does nothing else. Is the idea that any pending tasks in the asyncio queue that are planning to use that socket will instead gracefully fail without contacting the dbus server?

libqtile/core/manager.py
elParaguayo: Yes

elParaguayo:
> Does this solution fix it? From what I can tell reading dbus-next, `disconnect()` shuts down the socket but does nothing else. Is the idea that any pending tasks in the asyncio queue that are planning to use that socket will instead gracefully fail without contacting the dbus server?

This is only being called when we're finalising widgets. When they're reconfigured, they'll create a new bus connection with a new listening rule.

There shouldn't be any tasks in the queue that use this socket. When we create a signal listener, it creates a new bus connection and messages that meet our rule on that bus are sent to the appropriate callback. However, it is a one way connection, we're not actively sending messages on that bus. Where widgets do need to send message on a bus, they create a new connection.

m-col:
OK, so by shutting down the socket on our side, the dbus server is getting a socket closed signal and will stop sending any future messages on that socket, thus preventing qtile from receiving redundant messages?

elParaguayo:
I actually think there's a bug in dbus-next as the socket isn't closed on a `disconnect` call. Someone else already posted an issue about that and I've submitted a PR. However, more worrying is the fact that the dev has not been at all active recently.

I also want to add some other tidying up to this (e.g. `_send_dbus_message` should be able to accept a bus object rather than creating a new one every time).

libqtile/widget/mpris2widget.py
```diff
@@ -261,14 +267,31 @@ async def process_message(self, message):
 
         self.parse_message(*message.body)
 
+    async def send_message(self, destination, interface, path, member, signature, body):
```
m-col, line 270: If appropriate
```suggestion
    async def _send_message(self, destination, interface, path, member, signature, body):
```

m-col (approved):
LGTM. I wonder if we should encapsulate the dbus-related code in `utils` into a dedicated `dbus.py` module, or perhaps a dbus management class considering we maintain some state (`dbus_bus_connections`).

elParaguayo:
> LGTM. I wonder if we should encapsulate the dbus-related code in `utils` into a dedicated `dbus.py` module, or perhaps a dbus management class considering we maintain some state (`dbus_bus_connections`).

We could move it out of `utils.py` if that file's becoming too big. However, I'm not sure we need a management class as widgets create their own bus connections. The alternative would be to create a single bus connection for qtile and have that subscribe to signals etc but it would then need the logic to trigger the appropriate callbacks. I'm not sure if that's worth the effort nor if it would bring any measurable benefits.

m-col:
>  I'm not sure if that's worth the effort nor if it would bring any measurable benefits.

Yep fair enough - good as is!

elParaguayo:
Rebasing this as we should get this merged.

libqtile/utils.py
```diff
@@ -523,6 +537,29 @@ async def add_signal_receiver(
         return False
 
 
+async def remove_signal_receiver(bus: MessageBus, rule: str) -> bool:
```
elParaguayo, line 540: We can probably delete this function entirely. Don't think anything is using it.

elParaguayo:
@m-col @jwijenbergh I added some more tidying up of dbus bits. Would be good to get these merged as, without them, users who reload their config frequently will keep adding dbus connections while not closing old ones and there'll be some spamming in the logs too from the bluetooth widget.

PR #3544, opened 2022-05-02
Allow 3 digit hex colour values

elParaguayo:
A common short-hand version for hex colours is to use `RGB` to represent `RRGGBB`.

This PR amends `libqtile.utils` to allow 3 digit hex values to be treated as valid colours.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -91,9 +91,13 @@ def rgb(x: ColorType) -> tuple[float, float, float, float]:
             alpha = float("0." + alpha_str)
         else:
             alpha = 1.0
-        if len(x) not in (6, 8):
-            raise ValueError("RGB specifier must be 6 or 8 characters long.")
-        vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
+        if len(x) not in (3, 6, 8):
+            raise ValueError("RGB specifier must be 3, 6 or 8 characters long.")
+        if len(x) == 3:
+            # Multiplying by 17: 0xA * 17 = 0xAA etc.
+            vals = tuple(int(i, 16) * 17 for i in x)
+        else:
+            vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
         if len(x) == 8:
             alpha = int(x[6:8], 16) / 255.0
         vals += (alpha,)  # type: ignore
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -63,6 +63,22 @@ def test_rgb_from_base10_tuple_with_alpha():
     assert utils.rgb([255, 255, 0, 0.5]) == (1, 1, 0, 0.5)
 
 
+def test_rgb_from_3_digit_hex_number():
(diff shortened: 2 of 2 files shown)


libqtile/utils.py
```diff
@@ -91,8 +91,10 @@ def rgb(x: ColorType) -> tuple[float, float, float, float]:
             alpha = float("0." + alpha_str)
         else:
             alpha = 1.0
-        if len(x) not in (6, 8):
-            raise ValueError("RGB specifier must be 6 or 8 characters long.")
+        if len(x) not in (3, 6, 8):
+            raise ValueError("RGB specifier must be 3, 6 or 8 characters long.")
+        if len(x) == 3:
+            x = "".join(f"{col}{col}" for col in x)
```
m-col, line 97: This conversion looks like an interesting challenge. Is this the fastest way?

`"".join(col * 2 for col in x)` seems very slightly faster but I bet we can do better. Maybe something to do with `bytes`...

elParaguayo: I couldn't think of anything brilliant here. It is just a string after all.

To get to the double digit version of a single hex number you multiply by 17 but you don't want to split the string, convert to an int, multiply by 17, convert back to hex, join into a string. It must be quicker to stick with string manipulation.

elParaguayo: I suspect your way is the best.

m-col: Is that faster?

m-col:
LGTM!

libqtile/utils.py
elParaguayo: There's virtually nothing in it, from what I can tell (I'm not an expert on timing code though...)

```
>>> import time
>>>
>>> def timer(func, args, reps=10000):
...   times = []
...   for _ in range(reps):
...     start = time.time()
...     func(args)
...     times.append(time.time() - start)
...   return sum(times)
...
>>> def elp(x):
...     x = "".join(f"{col}{col}" for col in x)
...     vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
...
>>> def mcol(x):
...     x = "".join(col * 2 for col in x)
...     vals = tuple(int(i, 16) for i in (x[0:2], x[2:4], x[4:6]))
...
>>> def elpmultiply(x):
...     vals = tuple(int(i, 16) * 17 for i in x)
...
>>> val = "abc"
>>>
>>> print(f"ElParaguayo: {timer(elp, val):.10f}")
ElParaguayo: 0.0359008312
>>> print(f"m-col: {timer(mcol, val):.10f}")
m-col: 0.0216133595
>>> print(f"ElParaguayo #2: {timer(elpmultiply, val):.10f}")
ElParaguayo #2: 0.0117256641
```

m-col: You might find `timeit.timeit` useful, it's basically your `timer` function here.

"There's virtually nothing in it" - you managed to cut it down by 2/3! Well played.

elParaguayo: Never used `timeit`!

I also looked at:
``` python
vals = tuple(int(i * 2, 16) for i in x)
```
And that's slower than my last one too.

PR #3546, opened 2022-05-02
Enforce proper logging style

m-col:
This usess the flake8 plugin 'flake8-logging-format' to enforce proper
style of logging calls. The main misuse in libqtile is manually passing
exceptions, when the logging module gets exceptions and appends them to
'logging.exception' calls, and using 'logger.error' in place of
'logger.exception' when passing exceptions.


elParaguayo:
Very nice. LGTM.

m-col:
Woohoo!

ramnes:
Yay, more linting!

PR #3547, opened 2022-05-02
add debugging in wayland section to docs

### LINKED ISSUE #3298: Wayland Qtile current git master crashes when the first window is opened
I'm usually running the latest git master on Wayland to test the latest additions/fixes and to be able to merge in patches. I know this is not necessarily a good idea, so this issue is just to notify you of a new problem.
The current git master crashes silently when the first window is opened. There is no relevant error in Qtile.log, but journalctl shows the following:
```
feb 04 08:40:23 laptop systemd-coredump[31713]: [🡕] Process 31418 (qtile) of user 1000 dumped core.

                                                Module /usr/bin/python3.10 with build-id 12c63a5e2c0b1214b6384a2631e9515614ef42a3
                                                Stack trace of thread 31418:
                                                #0  0x00007f9ab92b1ca4 n/a (n/a + 0x0)
                                                ELF object binary architecture: AMD x86-64
```
I'd like to be able to debug these kind of problems better, but [the docs](https://docs.qtile.org/en/stable/manual/troubleshooting.html) don't go into low level debugging with Wayland, so I really don't know what my options are.

m-col:
closes #3298

### DIFF
--- a/docs/manual/troubleshooting.rst
+++ b/docs/manual/troubleshooting.rst
@@ -24,8 +24,8 @@ within libqtile by using the ``logger``:
 
    try:
        # some changes here that might error
-   raise Exception as e:
-       logger.exception(e)
+   except Exception:
+       logger.exception("Uh oh!")
 
 ``logger.warning`` is convenient because its messages will always be visibile
 in the log. ``logger.exception`` is helpful because it will print the full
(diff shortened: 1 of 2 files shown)


ramnes:
:)

PR #3556, opened 2022-05-14
WL: support for primary selection (clibpoard)

anriha:
This adds proper support for primary selection so it works the same way as with X server.
It is dependant on https://github.com/flacjacket/pywlroots/pull/85

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -137,6 +137,7 @@ class Core(base.Core, wlrq.HasListeners):
         DataControlManagerV1(self.display)
         self.seat = seat.Seat(self.display, "seat0")
         self.add_listener(self.seat.request_set_selection_event, self._on_request_set_selection)
+        self.add_listener(self.seat.request_set_primary_selection_event, self._on_request_set_primary_selection)
         self.add_listener(self.seat.request_start_drag_event, self._on_request_start_drag)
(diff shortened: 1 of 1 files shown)


anriha:
Pywlroots is merged so this should be ready now.

m-col:
Thanks for this. I haven't noticed any programs that didn't have working clipboard. I'm not sure I understand the difference (in practice) between this protocol and those already supported. Please could you give me an example? Maybe I just don't use any programs that use this one.

anriha:
Atleast for me primary clipboard under wayland basically didn't work at all (just by selecting something). Only exception was firefox and there it only worked when copying from firefox to firefox so I assume that there is another clipboard directly in firefox. If I want to for example copy something from firefox to terminal it doesn't work for me.

anriha:
I use alacritty but I also tested it with foot. When testing with sway everything works as it should. (or after applying this PR)

m-col:
Interesting! I use foot and it works fine. I have firefox running natively in wayland, rather than via XWayland. Is yours running via XWayland?

anriha:
Nope it is running in wayland.

m-col:
foot has this configuration setting:
```
selection-target=both
```
It can be set to primary, which I guess yours is. Mine is both. OK - I'm happy with that.

Thanks again for this!

PR #3512, opened 2022-04-17
Rewriting the docs.

SweepingBishops:
Exactly what the title says :)

PR #3562, opened 2022-05-18
Add option to build docs without screenshots

elParaguayo:
Added a `make quick` target to build docs without screenshots.

Speeds up build times which is useful when debugging docs locally.

### DIFF
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -77,6 +77,16 @@ Build the documentation with ``make html``.
 
 Check the result by opening ``_build/html/index.html`` in your browser.
 
+.. note::
+
+  To speed up local testing, screenshots are not generated each time the documentation
+  is built.
+
+  You can enable screenshots by setting the ``QTILE_BUILD_SCREENSHOTS`` environmental
+  variable at build time e.g. ``QTILE_BUILD_SCREENSHOTS=1 make html``. You can also
+  export the variable so it will apply to all local builds ``export QTILE_BUILD_SCREENSHOTS=1``
+  (but remember to unset it if you want to skip building screenshots).
+
 Development and testing
 =======================
 
--- a/docs/manual/troubleshooting.rst
+++ b/docs/manual/troubleshooting.rst
@@ -71,7 +71,7 @@ Qtile, so that the debug logs printed by Qtile are only the server's.
 
 If you suspect a client may be responsible for a bug, it can be helpful to look
 at the issue trackers for other compositors, such as `sway
-<https://github.com/swaywm/sway/issues`_. Similarly if you're hacking on
+<https://github.com/swaywm/sway/issues>`_. Similarly if you're hacking on
 Qtile's internals and think you've found an unexpected quirk it may be helpful
 to search the issue tracker for `wlroots
 <https://gitlab.freedesktop.org/wlroots/wlroots/-/issues>`_.
--- a/docs/sphinx_qtile.py
+++ b/docs/sphinx_qtile.py
@@ -256,7 +256,12 @@ def generate_widget_screenshots():
 
 def setup(app):
     generate_keybinding_images()
-    generate_widget_screenshots()
+    # screenshots will be skipped unless QTILE_BUILD_SCREENSHOTS environment variable is set
+    # Variable is set for ReadTheDocs at https://readthedocs.org/dashboard/qtile/environmentvariables/
+    if os.getenv("QTILE_BUILD_SCREENSHOTS", False):
+        generate_widget_screenshots()
+    else:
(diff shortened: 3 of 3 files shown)


SweepingBishops (approved):
Thanks!

m-col:
For the makefile, how about something like this?:
```diff
diff --git a/docs/Makefile b/docs/Makefile
index 4d793021..330c1a82 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,8 +2,9 @@
 #
 
 # You can set these variables from the command line.
+DOCS_NO_SCREENSHOTS ?= 0
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build -E -W
+SPHINXBUILD   = DOCS_NO_SCREENSHOTS=${DOCS_NO_SCREENSHOTS} sphinx-build -E -W
 PAPER         =
 BUILDDIR      = _build
 
```
This lets one disable screenshots for all commands by exporting the env var in their shell or sticking it before the make command.

elParaguayo:
> Thanks! Could you also remove the commit fixing the documentation bug, I've fixed it here #3512.

Let's see which one gets merged first! :wink:

elParaguayo:
> For the makefile, how about something like this?:
> 
> ```diff
> diff --git a/docs/Makefile b/docs/Makefile
> index 4d793021..330c1a82 100644
> --- a/docs/Makefile
> +++ b/docs/Makefile
> @@ -2,8 +2,9 @@
>  #
>  
>  # You can set these variables from the command line.
> +DOCS_NO_SCREENSHOTS ?= 0
>  SPHINXOPTS    =
> -SPHINXBUILD   = sphinx-build -E -W
> +SPHINXBUILD   = DOCS_NO_SCREENSHOTS=${DOCS_NO_SCREENSHOTS} sphinx-build -E -W
>  PAPER         =
>  BUILDDIR      = _build
>  
> ```
> 
> This lets one disable screenshots for all commands by exporting the env var in their shell or sticking it before the make command.

I see what you're saying. It's more work for the user though.

elParaguayo:
> It looks like there is already a command to generate screenshots, so maybe we just shouldn't do that in the main build? Similarly the make command `genkeyimg` which is currently a dependency of the `html` command but there is no need for that -- we'll just need to run make a few times during deployment.

Not entirely sure what you're suggesting here. I think I tried making `genwidgetscreenshots` as a dependency of `make html` but it didn't stop the build on an error which I wanted it to do (so readthedocs would show an error).

m-col:
I'm suggesting we separate out `genwidgetscreenshots`, `genkeyimg` and `html` as completely different make commands. Then for CI and RTD, we just call `make` multiple times. Is that possible?

elParaguayo:
> I'm suggesting we separate out `genwidgetscreenshots`, `genkeyimg` and `html` as completely different make commands. Then for CI and RTD, we just call `make` multiple times. Is that possible?

Ah. Understood. Let me investigate...

elParaguayo:
Readthedocs does not use the makefile. They rely on extensions to customise the build so we do need to trigger the build of those images from within our `sphinx_qtile.py` extension.

m-col:
Maybe the extension can detect that it's in a RTD environment and act accordingly? They might export an environmental variable we can detect

elParaguayo:
I think I can set environmental variables as part of the build process so that would do it.

elParaguayo:
Funnily enough, I think my original approach was ok. I've removed the `make quick` target but that was just a shortcut to setting the environmental variable which is what determines whether screenshots are built.

The variable is unset by default which will result in screenshots being built. Setting the variable to any value will disable the docs.

I've added a note to the hacking page to clarify.

m-col:
> The variable is unset by default which will result in screenshots being built. Setting the variable to any value will disable the docs.

Can we invert this logic so that they don't get built by default, then set the variable in the RTD build process so that it does? Sorry if I've misunderstood something above!

elParaguayo:
> > The variable is unset by default which will result in screenshots being built. Setting the variable to any value will disable the docs.
> 
> Can we invert this logic so that they don't get built by default, then set the variable in the RTD build process so that it does? Sorry if I've misunderstood something above!

I think the answer to that is "yes" but it's a bit messy. We can't set the environmental variable in the configuration file. We set it in ReadTheDocs settings.

Happy to give that a try if you'd prefer it that way round.

m-col:
I think that would be preferable if it's not too much trouble. Anything to minimise friction for users who might want to contribute (especially to the docs!)

elParaguayo:
This should be working as you want it now. Screenshots are generated on readthedocs (you can check by clicking on the details link next the docs status message).

m-col:
Perfect! Thanks again.

PR #3563, opened 2022-05-18
Added widget for displaying load averages.

SweepingBishops:
For qtile#3561


SweepingBishops:
I could use some help in writing tests for this. I looked through a few of the current tests in the `test/widgets` directory but I still don't quite understand.

elParaguayo:
I will help with tests later. Let's review the widget first.

Unfortunately, I may not have much time until the weekend to do that but I'm happy to look then.

libqtile/widget/load.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def next_load(self):
+        self.time = next(self.cycled_times)
+
+    def poll(self):
+        loads = {}
+        uptime = subprocess.check_output("uptime").decode("utf-8").strip()
+        loads["1m"], loads["5m"], loads["15m"] = eval(
+            uptime[uptime.index("load") + 13 :]
+        )  # Gets the load averages as a dictionary.
```
elParaguayo, line 49: I don't love the string parsing and use of `eval`. Why wouldn't you use `psutil.getloadavg()`? Is it just to avoid the third party library?

The CPU widget already has this dependency so I would think it's fine to use here.

SweepingBishops: > This gets asked a lot. I wonder if we should add a `force_update` command to `ThreadPoolText` widgets so you could just do:

Yes, that would be ideal.

elParaguayo:
I've added a test for you (and a couple of small changes to the widget).

You'll need to fetch the updates from the remote and merge into your local branch. Or I can tidy it all up into a single commit and force push it to your branch and then get it merged.

SweepingBishops:
Thanks elParaguayo. :smile:

elParaguayo:
Added screenshots for docs too.

m-col:
LGTM!

elParaguayo:
Great. Can you clear your review so we can (squash and) merge?

3nprob:
Looks like this should cycle between the different periods right?

Constantly displaying `Load(1m)` with format `Load({time}):{load:.2f}`. The actual load value is updating, though.

elParaguayo:
You have to click on it to select which load you want to view.

3nprob:
oh, cool! thanks :)

PR #3564, opened 2022-05-18
Disable stalebot action

m-col:
This was useful in the past but at this stage we've removed a lot of
truly stale issues and we're left with important issues getting made
stale. Coupled with slightly less dev activity the past few months, this
has become troublesome. In the future when we are up to top speed again
we can re-enable it if we find it useful!

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,7 +15,8 @@ jobs:
         exempt-issue-labels: 'task,bug,enhancement'
         stale-pr-message: 'This PR is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 30 days.'
         stale-pr-label: 'stale'
-        days-before-stale: 90
+        exempt-all-pr-assignees: true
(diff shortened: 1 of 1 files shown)


elParaguayo:
We could be better at assigning issues/PRs amongst us to manage our workflow.

My one slight reservation about this is that it's quite useful to have genuinely stale issue automatically closed. If we remove this, we'll need to go through the list ourselves to do this - but maybe that's not a bad thing!

m-col:
> We could be better at assigning issues/PRs amongst us to manage our workflow.

Yeah for sure. Maybe we try to do better at this, and double the stale countdown for issues? I think we should disable it for PRs completely though.

elParaguayo:
Perfect.

m-col:
OK I've doubled the countdown duration, but for PRs rather than disabling it completely it will ignore PRs that are assigned, encouraging us to assign them!

elParaguayo:
LGTM.

PR #3565, opened 2022-05-19
Rewrite config.py docstrings, typing and docs

m-col:
This does a few things:
- Adds the --keep-going flag to sphinx-build so we can see all errors
  that would happen rather than it exiting when the first is hit.
- Rewrites all libqtile.config docstrings in a consistent format.
- Documents the Ez classes
- Makes all references in docstrings and the docs link to where they
  reference, improving navigability of the docs.
- Adds type hints to all of config.py and enforces this. They also
  improve the docs by showing valid arguments to class inits and
  methods.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build -E -W
+SPHINXBUILD   = sphinx-build -E -W --keep-going
 PAPER         =
 BUILDDIR      = _build
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -169,7 +169,7 @@ pygments_style = 'sphinx'
 todo_include_todos = True
 
 
-# -- Options for HTML output --------fautod-------------------------------------------
+# -- Options for HTML output ---------------------------------------------------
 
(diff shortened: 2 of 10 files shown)


SweepingBishops (changes requested):
> * Makes all references in docstrings and the docs link to where they
>   reference, improving navigability of the docs.

Can you make them look more 'clickable'? Like a link or something. Right now they look similar to the inline code format that we use.

m-col:
Do you know how I could? This is the standard RTD theme internal reference style AFAICT.

m-col:
Note that the inline code blocks with black text are all meant to be links. We just have a lot of broken ones that need fixing!

elParaguayo:
> Can you make them look more 'clickable'? Like a link or something. Right now they look similar to the inline code format that we use.

Not sure I follow what you mean. Can you give an example of what you think is wrong in this PR?

SweepingBishops:
The internal links (clickable):


The inline code blocks (not clickable):


I was suggesting making the above look more clickable, like a link:

m-col:
I see what you're saying, but I think deviating from the RTD theme can itself be confusing because of how popular the theme is online. On well-formatted docs all of those internal links are clickable (some of ours are not but should be).

PR #3569, opened 2022-05-20
Add `force_update` to `ThreadPoolText` widgets

elParaguayo:
This is a convenience command to help users trigger updates of polling widgets from key bindings, hooks etc.

This gets requested fairly often and we often end up telling people to do something like:
``` python
lazy.widget["some_polling_widget].eval("self.update(self.poll())")
```

This would be replaced by
``` python
lazy.widget["some_polling_widget"].force_update()
```


m-col:
LGTM

PR #3574, opened 2022-05-22
ThermalSensor: make text formattable

DCsunset:
Currently, the text for the ThermalSensor widget is hardcoded. This PR adds formattable text and fixes this problem.

Now three variables are available in the format string: `temp`, `tag`, and `unit`.

### DIFF
--- a/libqtile/widget/sensors.py
+++ b/libqtile/widget/sensors.py
@@ -43,8 +43,8 @@ class ThermalSensor(base.InLoopPollText):
     """
 
     defaults = [
+        ("format", "{temp:.1f}{unit}", "Display string formatting"),
         ("metric", True, "True to use metric/C, False to use imperial/F"),
-        ("show_tag", False, "Show tag sensor"),
         ("update_interval", 2, "Update interval in seconds"),
         ("tag_sensor", None, 'Tag of the temperature sensor. For example: "temp1" or "Core 0"'),
         (
@@ -79,7 +79,6 @@ class ThermalSensor(base.InLoopPollText):
 
         temperature_list = {}
         temps = psutil.sensors_temperatures(fahrenheit=not self.metric)
-        unit = "°C" if self.metric else "°F"
         empty_index = 0
         for kernel_module in temps:
             for sensor in temps[kernel_module]:
@@ -89,7 +88,7 @@ class ThermalSensor(base.InLoopPollText):
                         kernel_module if kernel_module else "UNKNOWN", str(empty_index)
                     )
                     empty_index += 1
-                temperature_list[label] = (str(round(sensor.current, 1)), unit)
+                temperature_list[label] = sensor.current
 
         return temperature_list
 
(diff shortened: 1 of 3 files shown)


libqtile/widget/sensors.py
```diff
@@ -89,21 +88,24 @@ def get_temp_sensors(self):
...
     def poll(self):
         temp_values = self.get_temp_sensors()
         if temp_values is None:
             return False
-        text = ""
-        if self.show_tag and self.tag_sensor is not None:
-            text = self.tag_sensor + ": "
-        text += "".join(temp_values.get(self.tag_sensor, ["N/A"]))
-        temp_value = float(temp_values.get(self.tag_sensor, [0])[0])
+        temp_value = temp_values.get(self.tag_sensor, float("nan"))
```
SweepingBishops, line 99:
```suggestion
        temp_value = temp_values.get(self.tag_sensor, float("N/A"))
```
I believe this would be more understandable.

DCsunset: But in Python, `N/A` cannot be converted to a float number. I'm using `nan` here because it's a float number instead of text now.

elParaguayo: Users may be a bit confused by the "nan" display.

If we can't get a temperature value then we could just display an error message instead.

libqtile/widget/sensors.py
```diff
@@ -43,8 +43,8 @@ class ThermalSensor(base.InLoopPollText):
     """
 
     defaults = [
+        ("format", "{temp:.1f}{unit}", "Display string formatting"),
```
elParaguayo, line 46: Can you add which fields are available here? e.g.  I see that `tag` is also an option but users won't know they can use this unless they look at the source.

libqtile/widget/sensors.py
```diff
@@ -89,21 +88,24 @@ def get_temp_sensors(self):
...
-            text = self.tag_sensor + ": "
-        text += "".join(temp_values.get(self.tag_sensor, ["N/A"]))
-        temp_value = float(temp_values.get(self.tag_sensor, [0])[0])
+        temp_value = temp_values.get(self.tag_sensor, float("nan"))
         if temp_value > self.threshold:
             self.layout.colour = self.foreground_alert
         else:
             self.layout.colour = self.foreground_normal
-        return text
+        unit = "°C" if self.metric else "°F"
```
elParaguayo, line 104: We don't need to set this every time the widget is polled. Can we set once (e.g. during `_configure`) instead?

elParaguayo (changes requested):

elParaguayo:
Thanks for this.

I've made a couple of comments but can you also add some more tests to check the formatting works?

DCsunset:
I've made the changes to fix the above issues. As for tests, I couldn't find previous tests for this widget. Should I add a new test for it?

elParaguayo:
Don't worry about the tests for now. I can pick that up later (I was getting it confused with `ThermalZone`).

One other question for you: does the default `format` produce the same output as the default for the old version?

DCsunset:
Yes. I think so. One change is that I remove the `show_tag` option because the user now can use the `tag` field in the format string.

PR #3580, opened 2022-05-23
clipboard widget: escape text

tych0:
If we don't, text with markup in it can generate:

2022-05-19 07:13:46,651 ERROR libqtile hook.py:fire():L391 Error in hook selection_change
  File "/home/tycho/.local/lib/python3.10/site-packages/libqtile/hook.py", line 389, in fire
    i(*args, **kwargs)
  File "/home/tycho/.local/lib/python3.10/site-packages/libqtile/widget/clipboard.py", line 95, in hook_change
    self.text = text
  File "/home/tycho/.local/lib/python3.10/site-packages/libqtile/widget/base.py", line 409, in text
  File "/home/tycho/.local/lib/python3.10/site-packages/libqtile/drawer.py", line 72, in text
  File "/home/tycho/.local/lib/python3.10/site-packages/libqtile/pangocffi.py", line 186, in parse_markup
    raise Exception("parse_markup() failed for %s" % value)

### DIFF
--- a/libqtile/widget/clipboard.py
+++ b/libqtile/widget/clipboard.py
@@ -21,7 +21,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from libqtile import bar, hook
+from libqtile import bar, hook, pangocffi
 from libqtile.backend.x11 import xcbq
 from libqtile.widget import base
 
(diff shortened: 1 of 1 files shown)


m-col:
Only some text needs escaping, whereas markup is actually desired at some points (i.e. at the point of formatting before rendering). If the textbox widgets have a common step between getting the content (then escaping that), and then applying markup, then it might be neat doing it that way yeah.

tych0:
Yeah, seems like we'd need to do some more advanced decoupling. But it also seems like it might be work worth doing, as I suspect there are other widgets out there with this same bug that nobody has noticed. In any case, I'll merge this for now.

PR #3586, opened 2022-05-26
groupbox: add `invert_line` parameter

jx11r:
**Description:**
This PR allows the "line" highlight method to toggle up/down when it's highlighted.

**Example:** *Tested with Xephyr simulating two monitors.*

```python
widget.GroupBox(
    highlight_method = 'line',
    invert_line = False,
)
```
image

```python
widget.GroupBox(
    highlight_method = 'line',
    invert_line = True,
)
```
image

**Video:**

https://user-images.githubusercontent.com/78786202/170595021-1d87b3f8-6c65-491d-a408-1a06668ef0a5.mp4

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -91,6 +91,7 @@ class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
         rounded=False,
         block=False,
         line=False,
+        invert_line=False,
         highlighted=False,
     ):
         self.layout.text = self.fmt.format(text)
@@ -104,6 +105,8 @@ class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
                 (self.bar.height - self.layout.height - self.borderwidth) / 2,
                 (self.bar.height - self.layout.height + self.borderwidth) / 2,
             ]
+            if highlighted:
+                invert_line = False
         else:
             pad_y = self.padding_y
 
@@ -127,7 +130,7 @@ class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
         if block and bordercolor is not None:
             framed.draw_fill(offset, y, rounded)
         elif line:
-            framed.draw_line(offset, y, highlighted)
+            framed.draw_line(offset, y, invert_line, highlighted)
         else:
             framed.draw(offset, y, rounded)
 
@@ -224,6 +227,7 @@ class GroupBox(_GroupBase):
             "Hide groups that have no windows and that are not displayed on any screen.",
         ),
         ("spacing", None, "Spacing between groups" "(if set to None, will be equal to margin_x)"),
+        ("invert_line", False, "Invert visibility when 'line' highlight method is not highlighted."),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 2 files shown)


elParaguayo:
Why do you need to edit `drawer.py`? From what I can see, the `toggle` parameter just results in the `y` value effectively being changed. If that's the case, why don't you just pass the modified `y` value in the first place?

jx11r:
Because the `y` value is the same, but is only applied if `highlight` and `toggle` are true. If there is another way to do it, any change is welcome.

libqtile/widget/groupbox.py
```diff
@@ -127,7 +128,7 @@ def drawbox(
         if block and bordercolor is not None:
             framed.draw_fill(offset, y, rounded)
         elif line:
-            framed.draw_line(offset, y, highlighted)
+            framed.draw_line(offset, y, highlighted, toggle=toggle)
```
elParaguayo, line 131: Couldn't you just check the state of `highlighted` and `toggle` here and set the `y` value accordingly, rather than adding code to `drawer.py`?

The `toggle_line` name is a bit confusing too. You seem to be changing the position of the line rather than the visibility of the line.

jx11r: I think I could do it, but I'm not sure if it will apply when highlighted, since `highlight` is inside drawer.py

jx11r: And yes, the name is a bit confusing, how would you name it?

elParaguayo: > I think I could do it, but I'm not sure if it will apply when highlighted, since `highlight` is inside drawer.py

Isn't the `highlight` in `drawer.py` just the `highlighted` argument in the `draw_line` call above? If so, you can do all the condition checking here.

jx11r: Yes, that's right, I'll try it.

jx11r: Ok, I can't set a value to `y` from `groupbox.py` because in `drawer.py` the value of `y` is always `self.height - self.border_width`.

jx11r: @elParaguayo so?

elParaguayo: So you can use `framed.height` and `framed.border_width`.

I will confess, I'm finding this whole PR a bit confusing. The `invert` parameter is just to put the line on the top for groups on the inactive screen. Is that right? 

Conceivably, somebody might want that the other way round. So, maybe the text frame needs to be more flexible as to where the line is drawn (and this widget should also be updated to work on vertical bars so we would need lines on left and right).

If that's sounding like too much work then you may want to keep your changes within your config and wait until I get round to doing the other bits.

jx11r: > So you can use `framed.height` and `framed.border_width`.

But if I modify those values, won't it affect the border width and bar size?

> The invert parameter is just to put the line on the top for groups on the inactive screen. Is that right?

Yes.

> Conceivably, somebody might want that the other way round.

That's easy to fix, I just need to create another argument or create just one that contains 3 possible values: `'None'`, `'Up'` and `'Down'`.

> this widget should also be updated to work on vertical bars

Huh?, the docs says that it only supports horizontal orientation.

jx11r: Never mind, closed for now while the rest is completed.

elParaguayo: > Huh?, the docs says that it only supports horizontal orientation.

Yes, it does. My point was that, there's no reason this widget couldn't work on a vertical bar too and, once that change has been made, that would have implications for this PR too.

jx11r: Yup, I understand it :)

PR #3590, opened 2022-05-28
Replace subprocess.run with subprocess.Popen in docs

PreciousWarrior:
`subprocess.run` is blocking, so in case the script has errors or a missing `&`, the window manager will be stuck on a black screen. `subprocess.Popen` is not blocking, so it is a better choice to recommend to new users for autostarting applications.

### DIFF
--- a/docs/manual/config/hooks.rst
+++ b/docs/manual/config/hooks.rst
@@ -64,7 +64,7 @@ We can then subscribe to ``startup_once`` to run this script:
     @hook.subscribe.startup_once
     def autostart():
         home = os.path.expanduser('~/.config/qtile/autostart.sh')
-        subprocess.run([home])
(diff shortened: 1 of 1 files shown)


elParaguayo:
Don't most autostart scripts exit almost immediately as they push processes to the background?

We've not had any bug reports where this has been the issue.

PreciousWarrior:
Yes, but silly me forgot an `&` at the last line, so I spent an hour debugging and messing around with Xorg, since when starting the WM all I saw was a blank screen. I think its just better to not block the WM entirely if there is a problem with the autostart script.

elParaguayo:
Fair enough. 

You can also use an async function with the hook and use an `asyncio` method to run the script but that's even less friendly for beginners!

m-col (approved):

PR #3592, opened 2022-05-28
Fire client_killed hook after client is removed from group

### LINKED ISSUE #3513: [Wayland] Crash on client_killed hook when calling cmd_to_layout_index
### The issue:

Hi,

I've stumbled upon an issue that appears on Wayland. I'm using Qtile version 0.21.1.dev9+g634a470a.

My use case is that I want to automatically fall back to a "default" layout whenever there is only a single window left on the given group.

To achieve this I utilize the client_killed hook:
```python
@hook.subscribe.client_killed
def fallback_default_layout(client: Window) -> None:
    """Reset a group to default layout when theres is only one window left"""
    assert qtile is not None, "This should never be None"

    if client.group is None:
        return

    try:
        win_count = len(client.group.windows)
    except AttributeError:
        win_count = 0

    if win_count > 2:
        return

    group_name: str = client.group.name
    default_layout_index: int = 0

    qtile.cmd_to_layout_index(default_layout_index, group_name)
```

The above works fine on X11, but I just now started testing out Wayland and noticed that this unfortunately causes Qtile to silently crash when triggered.
I managed to verify (using logger) that the culprit indeed seems to be the cmd_to_layout_index function on the last row.

I'm not getting anything out of the logs when the crash happens though, despite also trying the debug level. If you have any pointers on how I could go about and dig into this further I'm happy to do so.

Thank you for developing Qtile!

### Required:

m-col:
I'm not sure why this isn't the case already but let's keep an eye out
for issues that arise due to this. I suspect most uses of this hook
don't involve the group but those that do may see different
behaviour.

This fixes a Wayland bug where laying out the group's layout with a
killed window raises a wlroots assertion.

fixes #3513

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -628,7 +628,6 @@ class Qtile(CommandObject):
     def unmanage(self, wid: int) -> None:
         c = self.windows_map.get(wid)
         if c:
-            hook.fire("client_killed", c)
             if isinstance(c, base.Static):
                 if c.reserved_space:
                     self.free_reserved_space(c.reserved_space, c.screen)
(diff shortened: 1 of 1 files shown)


m-col:
This isn't an appropriate fix. I'll submit a new PR with attempt number 2.

Issue #3601, opened 2022-05-30
Can't login to the qtile wayland session


deepanshu188:
### The issue:

Qtile x11 is working fine but wayland session does not open.

I have tried **qtile start -b wayland** command on terminal but it is throwing following errors

```
Traceback (most recent call last):
  File "/sbin/qtile", line 33, in <module>
    sys.exit(load_entry_point('qtile==0.21.0', 'console_scripts', 'qtile')())
  File "/usr/lib/python3.10/site-packages/libqtile/scripts/main.py", line 66, in main
    options.func(options)
  File "/usr/lib/python3.10/site-packages/libqtile/scripts/start.py", line 90, in start
    q = make_qtile(options)
  File "/usr/lib/python3.10/site-packages/libqtile/scripts/start.py", line 51, in make_qtile
    kore = libqtile.backend.get_core(options.backend)
  File "/usr/lib/python3.10/site-packages/libqtile/backend/__init__.py", line 15, in get_core
    return importlib.import_module(f"libqtile.backend.{backend}.core").Core(*args)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/libqtile/backend/wayland/__init__.py", line 2, in <module>
    from libqtile.backend.wayland.inputs import InputConfig  # noqa: F401
  File "/usr/lib/python3.10/site-packages/libqtile/backend/wayland/inputs.py", line 26, in <module>
    from wlroots import ffi, lib
ModuleNotFoundError: No module named 'wlroots'

```

[...]

PR #3602, opened 2022-05-30
WL: Print useful message when backend deps are missing

m-col:
If pywayland or pywlroots is missing, `qtile start -b wayland` will
output this message:

"Backend dependencies not found. Please install pywayland and pywlroots."

Rather than printing a traceback mentioning 'wlroots' (can be confused
with the main C library).

See #3601

### DIFF
--- a/libqtile/backend/__init__.py
+++ b/libqtile/backend/__init__.py
@@ -1,14 +1,45 @@
+from __future__ import annotations
+
 import importlib
+import importlib.util
+from typing import TYPE_CHECKING
 
 from libqtile.utils import QtileError
 
-CORES = [
-    "wayland",
-    "x11",
-]
+if TYPE_CHECKING:
+    from typing import Any
+
+    from libqtile.backend.base import Core
 
 
-def get_core(backend, *args):
+CORES = {
+    "wayland": ("wlroots",),
+    "x11": ("xcffib",),
+}
+
+
+def has_deps(backend: str) -> list[str]:
+    """
+    Check if the backend has all its dependencies installed.
+
+    Args:
+        backend: The backend to check.
+
+    Returns:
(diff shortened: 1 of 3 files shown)


elParaguayo:
Hmm. Docs are failing because the screenshots use pytest which, in turn, reads every `conftest.py` file.

m-col:
Why didn't it fail before? This doesn't add a new exception, it just replaces the previous (uninformative) one with a different one

elParaguayo:
You've got new imports in `libqtile/backend/wayland/__init__.py` that weren't there before.

m-col:
`InputConfig` was imported before, and that module imports `pywayland` and `wlroots`, so it did before, just indirectly. Or am I missing something?

elParaguayo:
> `InputConfig` was imported before, and that module imports `pywayland` and `wlroots`, so it did before, just indirectly. Or am I missing something?

Not missing anything. That is weird!

elParaguayo:
Ah `conftest` catches an `ImportError` but not the `QtileError`.

m-col:
Oops, didn't mean to request reviews from everyone. Hey all :)

ramnes (approved):
Love it

libqtile/backend/__init__.py
```diff
@@ -1,14 +1,45 @@
...
-    "x11",
-]
+if TYPE_CHECKING:
+    from typing import Any
+
+    from libqtile.backend.base import Core
+
+
+CORES = {
+    "wayland": ("wlroots",),
```
ramnes, line 16: Why not listing pywayland and pywlroots here?

m-col: `wlroots` is the name of the `pywlroots` module (and we test by importing), and pywlroots pulls in `pywayland` anyway

PR #3604, opened 2022-05-30
invert_mouse_wheel option on backlight widget

fdev31:
Allows reversing the mouse wheel actions so moving up increases the backlight and down decreases it.

### DIFF
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -84,6 +84,7 @@ class Backlight(base.InLoopPollText):
         ("step", 10, "Percent of backlight every scroll changed"),
         ("format", "{percent:2.0%}", "Display format"),
         ("change_command", "xbacklight -set {0}", "Execute command to change value"),
+        ("invert_mouse_wheel", False, "invert mouse wheel actions"),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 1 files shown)


libqtile/widget/backlight.py
```diff
@@ -102,12 +103,17 @@ def __init__(self, **config):
...
+                "Button4": partial(self.cmd_change_backlight, ChangeDirection.DOWN),
+                "Button5": partial(self.cmd_change_backlight, ChangeDirection.UP),
+            }
+        else:
+            callbacks = {
                 "Button4": partial(self.cmd_change_backlight, ChangeDirection.UP),
                 "Button5": partial(self.cmd_change_backlight, ChangeDirection.DOWN),
             }
-        )
+        self.add_callbacks(callbacks)
```
elParaguayo, line 116: I have a slight preference for writing it like this but don't feel particularly strongly about it!
``` python
self.add_callbacks(
    {
        "Button4": partial(self.cmd_change_backlight, ChangeDirection.DOWN if self.invert_mouse_wheel else ChangeDirection.UP),
        "Button5": partial(self.cmd_change_backlight, ChangeDirection.UP if self.invert_mouse_wheel else ChangeDirection.DOWN),
    }
)
```

What do you think?

`black` may also format my lines if they're too long.

elParaguayo:
Thanks for this.

Seems like a reasonable change.

Are there any other widgets that may also benefit from having this change?

SweepingBishops:
> Are there any other widgets that may also benefit from having this change?

The volume widgets, mpd2, cmus etc. Basically anything that uses mouse scroll.

m-col:
Maybe it would be handy to just have in `_Widget`? Any widget could have mouse callbacks - of course it only applies to _default_ mouse callbacks, and we should ensure the docs make that clear.

elParaguayo:
I think that's actually quite hard to implement.

The base widget init code is run before other widgets add their own callbacks. We could move the flipping code into configure but that would then pick up user callbacks too.

Could have this as a mixing so we can just add to widgets that define callbacks on scroll events.

I'd also prefer to keep it being added to the defaults for all widgets if it has no effect for lots of them.

m-col:
I see - makes sense. I'm not sure a mixin would be able to append the option to the defaults, though, unless we just used the mixin for the mouse_callbacks setup, though it's a bit weird having part of the functionality of the mixin in the mixin, and part in the widget.

Another approach (which breaks away from how we have been doing things!) could be a decorator, as a decorator would have the ability to append to the defaults and append logic to `__init__`. It might also be a bit weird though.

elParaguayo:
Let me have a play with a mixin option and I'll post it as a draft PR so people can take a look.

PR #3608, opened 2022-06-04
WL: keep XWayland Z stack in sync with backend's stack

### LINKED ISSUE #3250: [ xwayland ] Cursor display and input sketchy running Ardour
If I run Ardour from the arch repo using Qtile with wayland obvs the new session window opens with the cursor as a cross but it will not let me select any options. I can tab through to select them on the keyboard. When i get to the main window the cursor is still a cross and menus are operational but if I then exit I get dumped out of Qtile to the tty login.
I'm using qtile-git and the latest wayland python-pywlroots , python-pywalyand from the normal repos.

m-col:
The X(Wayland) server keeps its own Z stack of clients but this isn't
always in sync with the wayland core's stack. This causes some weird
input issues, because windows look like they have one stacking order but
input handling uses a different order. This improves the situation and
attempts to keep them more in sync.

Fixes #3250

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -772,7 +772,6 @@ class Core(base.Core, wlrq.HasListeners):
 
         if isinstance(win, (window.XWindow, window.XStatic)):
             if not win.surface.or_surface_wants_focus():
-                win.surface.restack(None, 0)  # XCB_STACK_MODE_ABOVE
                 return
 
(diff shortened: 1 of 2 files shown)


justinesmithies:
@m-col This fixes all issues with Ardour now. So it's good to be merged when you're ready.

m-col:
Awesome thanks for confirming :)

PR #3609, opened 2022-06-04
Add InvertScrollMixin

elParaguayo:
Short "proof of concept" following discussion in #3604.

Comments welcome.

To confirm that the mixin's default is added to the widget's:

### DIFF
--- a/libqtile/widget/backlight.py
+++ b/libqtile/widget/backlight.py
@@ -38,7 +38,7 @@ class ChangeDirection(enum.Enum):
     DOWN = 1
 
 
-class Backlight(base.InLoopPollText):
+class Backlight(base.InLoopPollText, base.InvertScrollMixin):
     """A simple widget to show the current brightness of a monitor.
 
     If the change_command parameter is set to None, the widget will attempt to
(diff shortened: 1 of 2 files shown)


m-col:
Oh yeah nicely done!

How do you think this should work WRT default or custom mouse bindings? Only apply to the defaults, or just globally swap Button4 and Button5?

elParaguayo:
My view is that it's done globally. The default is not to invert so if a user defines their own callbacks and also sets `invert_scroll` then that's on them!

Also, this mixin would only be added to widgets that have hard coded scroll callbacks where we might not expect users to override so the risk feels low to me.

m-col:
If someone provides button 4 and 5 bindings to a widget with this option, and one without, but passes the option to both (or in widget_defaults), the bindings are going to flip relative to each other, are they not?

elParaguayo:
Yes - but if a user is passing mouse_callbacks in their config, they really shouldn't be setting the invert option because they can set the scroll directions directly.

m-col:
Hmm, yes, that sounds about right. I can't shake off the feeling we're missing something here, logically speaking. But OK, I'm happy to go with it!

PR #3614, opened 2022-06-05
Fix text positioning when Bar has margin

elParaguayo:
The scrolling text commit introduced a bug where text was positioned incorrectly if the Bar had a margin.

This PR removes that bug. Hopefully.


m-col:
LGTM

justinesmithies:
That certainly fixes the issues I had.

Issue #3594, opened 2022-05-28
XWayland: Firefox context menus always open in their very first position


m-col:
### The issue:

Version: 32a89c5ec067d9e96b086cb2501aaaf16949f486

This is specific to XWayland, so firefox needs to be run using that.

This likely applies to all override-redirect windows.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

Shinyzenith:
Hi @m-col would you mind sharing a screenshot? I'd like to try my hands at fixing some wayland bugs and I don't quite understand "open in their very first position"

Sydiepus:
Not sure if a screenshot would help but i'll try to explain "open in their very first position".

Imagine firefox as a 2D canvas (i think it's called) with x axis as abscissa and y axis as coordinate like this : 
![image](https://user-images.githubusercontent.com/49863809/171504517-21e7be0e-9392-41ec-bc96-441ce4e83a3d.png)
photo / info from [here](https://www.cuemath.com/geometry/x-and-y-graph/).

If the cursor is at (50,80) and right click is clicked then the context menu / right click menu will open at (50,80)
Now if the cursor moved to (70,100) and  right click is clicked then the context menu will open at (50,80) since the first right click was at (50, 80)

This also applies to tab names, the menu that appears when you hover on a tab.

I tried to record it but obs is crashing, and none of (kooha, peek) worked.
If you can recommend anything else i would be happy to record it.

Shinyzenith:
> I tried to record it but obs is crashing, and none of (kooha, peek) worked. If you can recommend anything else i would be happy to record it.

That's fine, your explanation is crystal clear 😁. I'll try my hand at fixing this.

m-col:
Yes exactly as  described, thank you @Sydiepus.

@Shinyzenith, thank you for trying to tackle this. For a bit of background on how the backend currently handles X11 override-redirect windows like these context menus, this is what happens when they are created (for the first time):

- `Core._on_xwayland_new_surface` is called with the new XWayland surface, and a new `XWindow` is added to `Core.pending_windows`, where it will wait until mapped for the first time.
- When it is mapped, `XWindow._on_map` is called
- The window is in `self.core.pending_windows` so we create a new WID for the window and we start properly managing it.
- In the case of these windows, `self.surface.override_redirect` is `True`, so we convert the window to a `XStatic` via `self.cmd_static`, and manage it through that interface instead.

[...]

PR #3618, opened 2022-06-06
WL: XStatic on_set_geometry listener added.

Shinyzenith:
* Closes https://github.com/qtile/qtile/issues/3594.

This bug genuinely taught me stuff and I think I'll be deploying a similar `*Static*` and `*Window*` structure for my compositor! Thank you Matt!

Signed-off-by: Shinyzenith <aakashsensharma@gmail.com>

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -1370,6 +1370,11 @@ class XStatic(Static[xwayland.Surface]):
         self.add_listener(surface.set_title_event, self._on_set_title)
         self.add_listener(surface.set_class_event, self._on_set_class)
 
+        # Checks to see if the user manually created the XStatic surface.
+        # In which case override_redirect would be false.
+        if surface.override_redirect:
+            self.add_listener(surface.set_geometry_event, self._on_set_geometry)
+
     def kill(self) -> None:
         self.surface.close()
 
@@ -1414,6 +1419,12 @@ class XStatic(Static[xwayland.Surface]):
         logger.debug("Signal: xstatic set_class")
         self._wm_class = self.surface.wm_class
         self.ftm_handle.set_app_id(self._wm_class or "")
+
+    def _on_set_geometry(self, _listener: Listener, _data: Any) -> None:
+        logger.debug("Signal: xstatic set_geometry")
+        self.place(
+            self.surface.x, self.surface.y, self.surface.width, self.surface.height, 0, None
+        )
 
 
 class LayerStatic(Static[LayerSurfaceV1]):


libqtile/backend/wayland/window.py
```diff
@@ -1415,6 +1420,13 @@ def _on_set_class(self, _listener: Listener, _data: Any) -> None:
         self._wm_class = self.surface.wm_class
         self.ftm_handle.set_app_id(self._wm_class or "")
 
+    def _on_set_geometry(self, _listener: Listener, _data: Any) -> None:
+        logger.debug("Signal: xstatic set_geometry")
+        self.x = self.surface.x
```
m-col, line 1425: It looks like we aren't respecting a new width or height. Should we also do that?

It would also be good to use `self.place` to reset the position rather than having some lines duplicated here - `self._find_outputs` is needed too.

Shinyzenith:
@m-col Updated!

Shinyzenith:
I did notice that not setting self.x = self.surface.x would lead to the same behavior although place() already sets self.x to

libqtile/backend/wayland/window.py
m-col: Ah sorry I wasn't clear - lines 1426-1429 aren't needed because the values passed into `place` will be used. You also need to pass in `0` as the borderwidth.

Shinyzenith: As stated earlier, the bug still persists if I don't explicitly mention self.x = self.surface.x for some odd reason that I don't currently understand.

m-col: Ah I see. It's because your place() call fails because it needs the borderwidth argument, which is required. I think if you pass that then you won't need the manual setters for `self.x` and `self.y`.

Shinyzenith: > Ah I see. It's because your place() call fails because it needs the borderwidth argument, which is required. I think if you pass that then you won't need the manual setters for `self.x` and `self.y`.

already thought of that and created the following line:
```py
self.place(self.surface.x, self.surface.y, geometry.width, geometry.height, 0, None)
```
But oh well it still doesn't work. I'll look into this when i get some time.

Shinyzenith: > logger.exception

Hm I didn't try a `try/except` block before. But sad news, still doesn't work ahaha just silently fails.

I'll push my progress so maybe you can take a look?

m-col:
Just testing it and I'm getting this:
```
  File "/home/mcol/git/pywayland/pywayland/server/listener.py", line 40, in notify_func
    callback(listener, data)
  File "/home/mcol/git/qtile/libqtile/backend/wayland/window.py", line 1444, in _on_set_geometry
    geometry = self.surface.get_geometry()
AttributeError: 'Surface' object has no attribute 'get_geometry'
```
This instead seems to work:
```diff
diff --git a/libqtile/backend/wayland/window.py b/libqtile/backend/wayland/window.py
index d2988f35..1b125c24 100644
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -1441,8 +1441,7 @@ class XStatic(Static[xwayland.Surface]):
 
     def _on_set_geometry(self, _listener: Listener, _data: Any) -> None:
         logger.debug("Signal: xstatic set_geometry")
-        geometry = self.surface.get_geometry()
-        self.place(self.surface.x, self.surface.y, geometry.width, geometry.height, 0, None)
+        self.place(self.surface.x, self.surface.y, self.surface.width, self.surface.height, 0, None)
 
 
 class LayerStatic(Static[LayerSurfaceV1]):
```

Shinyzenith:
> Just testing it and I'm getting this:
> 
> ```
> Traceback (most recent call last):
>   File "/home/mcol/git/pywayland/pywayland/server/listener.py", line 40, in notify_func
>     callback(listener, data)
>   File "/home/mcol/git/qtile/libqtile/backend/wayland/window.py", line 1444, in _on_set_geometry
>     geometry = self.surface.get_geometry()
> AttributeError: 'Surface' object has no attribute 'get_geometry'
> ```
> 
> This instead seems to work:
> 
> ```diff
> diff --git a/libqtile/backend/wayland/window.py b/libqtile/backend/wayland/window.py
> index d2988f35..1b125c24 100644
> --- a/libqtile/backend/wayland/window.py
> +++ b/libqtile/backend/wayland/window.py
> @@ -1441,8 +1441,7 @@ class XStatic(Static[xwayland.Surface]):
>  
>      def _on_set_geometry(self, _listener: Listener, _data: Any) -> None:
>          logger.debug("Signal: xstatic set_geometry")
> -        geometry = self.surface.get_geometry()
> -        self.place(self.surface.x, self.surface.y, geometry.width, geometry.height, 0, None)
> +        self.place(self.surface.x, self.surface.y, self.surface.width, self.surface.height, 0, None)
>  
>  
>  class LayerStatic(Static[LayerSurfaceV1]):
> ```

huh, not sure why I didn't get the error.. eitherway I will get it fixed and pushed

m-col:
Please could you format the code with black? Looks like it's complaining in the CI. Then good to go i think

m-col:
Amazing, thanks again for this fix!

PR #3620, opened 2022-06-06
Do not spam logs when we have to try multiple files for battery info

ramnes:
Here's what I've got otherwise:

```
022-06-06 20:50:45,564 ERROR libqtile battery.py:_load_file():L239 Failed to read '/sys/class/power_supply/BAT0/current_now':
  File "libqtile/widget/battery.py", line 237, in _load_file
    return f.read().strip(), value_type
OSError: [Errno 19] No such device
```

The code tries multiple files on purpose, let's not log an error when one of these can't be found.


elParaguayo:
Nice. I just noticed some of those messages too.

LGTM.

PR #3631, opened 2022-06-15
feat(volume/widget) select which emojis show up

LamprosPitsillos:
I simply wanted to choose symbols from my font instead of the emojis, so i added that feature.

I think the only thing missing is to add to the documentation.

Should also work with plain text : `...,emoji_list = ["mute","low","high","max"],...` but that is not my indented feature, i'd say its a plus tho. Maybe change the name to something more generic?


elParaguayo:
Thanks.

Can you add some error checking as it will crash if the list doesn't have enough items.

elParaguayo:
If it were me, I'd probably check the list length (only if user was using emojis) and, if there weren't enough items, I'd log a message and fill the list with default icons.

LamprosPitsillos:
> If it were me, I'd probably check the list length (only if user was using emojis) and, if there weren't enough items, I'd log a message and fill the list with default icons.

I was asking about the logging part mostly, the check is simple enough. How do I log this error properly?

m-col:
> How do I log this error properly?

Like this (if that's what you were asking):

```py

...

logger.warning("message here")
```

KrolixW:
Have you considered using dict instead of a list?

LamprosPitsillos:
> Have you considered using dict instead of a list?

Could you elaborate a bit more? Why would i use dictionaries?

KrolixW:
> Could you elaborate a bit more? Why would i use dictionaries?
1) you won't need to check the length
2) it will be self explanatory which icon goes were instead of relying on order
Also you could fallback to default icon only for those that haven't been set (but that's probably easier if you just make every icon a separate option)

LamprosPitsillos:
> > Could you elaborate a bit more? Why would i use dictionaries?
> 
> 1. you won't need to check the length
> 2. it will be self explanatory which icon goes were instead of relying on order
>    Also you could fallback to default icon only for those that haven't been set (but that's probably easier if you just make every icon a separate option)

I believe lists are a bit better suited for this kind of use, no need for the extra complexity of a dictionary.
Placement of the icons is pretty easy to understand,from lower to higher, a dictionary approach would be something like

 `{"mute":icon , "low": icon , "mid": icon ,"high": icon}`

, which like you said, makes order obsolete but _why_ would you mix the order?

I see the small benefits you are talking about but i don't think their necessary nor worth the change.

libqtile/widget/volume.py
```diff
@@ -35,6 +35,7 @@
 
 from libqtile import bar
 from libqtile.widget import base
+from libqtile.log_utils import logger
```
m-col, line 38:
```suggestion
from libqtile.log_utils import logger
from libqtile.widget import base
```

libqtile/widget/volume.py
```diff
@@ -147,14 +154,18 @@ def _update_drawer(self):
             self.drawer.ctx.set_source(self.surfaces[img_name])
             self.drawer.ctx.paint()
         elif self.emoji:
+            if len(self.emoji_list) < 4:
+                self.emoji_list = ["\U0001f507", "\U0001f508", "\U0001f509", "\U0001f50a"]
+                logger.warning("Emoji list given has less than 4 items. Falling back to default emojis.")
```
m-col, line 159: I'm not 100% certain what the CI was complaining about, but if you run `black` on your code it will fix it for you. My guess is the line is too long:

```suggestion
                logger.warning(
                    "Emoji list given has less than 4 items. Falling back to default emojis."
                )
```

libqtile/widget/volume.py
```diff
@@ -67,6 +68,12 @@ class Volume(base._TextBox):
             "Use emoji to display volume states, only if ``theme_path`` is not set."
             "The specified font needs to contain the correct unicode characters.",
         ),
+        (
+            "emoji_list",
+            ["\U0001f507", "\U0001f508", "\U0001f509", "\U0001f50a"],
+            "List of emojis/font-symbols to display volume states, only if ``emoji`` is set."
+            "List contains 4 symbols, from lowest volume to highest.",
```
m-col, line 75: The concatenated string of these two lines needs a space:
```suggestion
            " List contains 4 symbols, from lowest volume to highest.",
```

Though I notice that the options `emoji` and `step` also have this issue: http://docs.qtile.org/en/stable/manual/ref/widgets.html#libqtile.widget.Volume. Maybe you could put spaces on those lines too while you're there? :)

m-col (changes requested):
LGTM, you just need a couple of formatting changes to make the CI happy - I've commented inline what needs to be done :)

FreibergVlad:
Thanks for that. It's a nice feature I was looking for too.

@m-col @elParaguayo @LamprosPitsillos  are there any chances it will be merged in the nearest future? :)

LamprosPitsillos:
I know this is not a substantial PR ,but if it's missing anything for it to be pushed ,by all means , please tell me

m-col:
Looks good, apologies for the delay. I've rebased and enabled auto merge :)

m-col (approved):

PR #3632, opened 2022-06-15
Fix typo in docstring

CharString:
And change the value to show how multiple xkb options are to be passed.


m-col:
Excellent, thanks for this!

PR #3639, opened 2022-06-19
Fix `next` and `previous` commands for Bsp layout

elParaguayo:
The `next` and `previous` commands were broken as the layout looks for `self.current` in the client list. However, `self.current` is a Node, not a client and so there is never a match.

This PR fixes this by looking up the current node's `client` value.

I've also added looping so, when you call `next()` on the last client, you move to the first one.

### DIFF
--- a/libqtile/layout/bsp.py
+++ b/libqtile/layout/bsp.py
@@ -163,6 +163,11 @@ class Bsp(Layout):
         ("grow_amount", 10, "Amount by which to grow a window/column."),
         ("lower_right", True, "New client occupies lower or right subspace."),
         ("fair", True, "New clients are inserted in the shortest branch."),
+        (
+            "wrap_clients",
+            False,
+            "Whether client list should be wrapped when using ``next`` and ``previous`` commands.",
+        ),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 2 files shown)


m-col:
Looks like you need to update test_bsp_window_focus_cycle and test_focus_cycle. FWIW the looping you introduced is configurable in Columns via "wrap_focus_columns". Maybe it's worth making it configurable here too? Or just waiting until someone asks for it :smile:  otherwise LGTM

elParaguayo:
Was so bogged down in the `xcffib` stuff that I forgot about this! Thanks for reminding me.

elParaguayo:
Interesting. Adding looping breaks the behaviour of `lazy.group.next_window()`. That command works by firstly looping through the windows in the layout, when it reaches the end, it then loops through floating windows.

However, with looping in place, you never reach the end of layout windows and so you'll never hit the floating windows.

Looking at this, I wonder if the code for `cmd_next_window` should be changed so that it takes a list of all the windows in a group and just cycles through that rather than this approach:
``` python
            nxt = (
                self.layout.focus_next(self.current_window)
                or self.floating_layout.focus_first(group=self)
                or self.layout.focus_first()
            )
        self.focus(nxt, True)
```
(with looping, focus_next will always provide a client)

PR #3647, opened 2022-06-20
Accept lowercase keybinds

oddlama:
Currently, the wayland backend accepts lowercase keybinds, as it uses `case_insensitive=True` on all symbol lookups (see `grab_key()`). Both the X11 backend and the config validator only accept symbols as reported by X11. In reality, no two keys exist that share a name but have different capitalization, which is why `case_insensitive=True` even exists in wayland. Modifier keys on the other hand require all-lowercase in qtile. This PR fixes this inconsistent behavior across all backends, allowing any capitalization.

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -62,15 +62,6 @@ class XCBQError(QtileError):
     pass
 
 
-def rdict(d):
-    r = {}
-    for k, v in d.items():
-        r.setdefault(v, []).append(k)
-    return r
-
-
-rkeysyms = rdict(keysyms)
-
 # Keyboard modifiers bitmask values from X Protocol
 ModMasks = {
     "shift": 1 << 0,
@@ -790,7 +781,7 @@ class Painter:
 
 
 def get_keysym(key: str) -> int:
-    keysym = keysyms.get(key)
+    keysym = keysyms.get(key.lower())
     if not keysym:
         raise XCBQError("Unknown key: %s" % key)
     return keysym
@@ -812,7 +803,7 @@ def translate_masks(modifiers: list[str]) -> int:
     masks = []
     for i in modifiers:
         try:
-            masks.append(ModMasks[i])
+            masks.append(ModMasks[i.lower()])
         except KeyError as e:
             raise XCBQError("Unknown modifier: %s" % i) from e
     if masks:
--- a/libqtile/backend/wayland/wlrq.py
+++ b/libqtile/backend/wayland/wlrq.py
@@ -96,7 +96,7 @@ def translate_masks(modifiers: list[str]) -> int:
     masks = []
     for i in modifiers:
         try:
-            masks.append(ModMasks[i])
+            masks.append(ModMasks[i.lower()])
         except KeyError as e:
             raise WlrQError("Unknown modifier: %s" % i) from e
     if masks:
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -153,7 +153,7 @@ class Core(base.Core):
         self.qtile = None  # type: Qtile | None
         self._painter = None
 
-        numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["Num_Lock"])[0]
+        numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["num_lock"])[0]
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
         self._valid_mask = ~(
             self._numlock_mask
@@ -751,7 +751,7 @@ class Core(base.Core):
         """Simulates a keypress on the focused window."""
         # FIXME: This needs to be done with sendevent, once we have that fixed.
         modmasks = xcbq.translate_masks(modifiers)
-        keysym = xcbq.keysyms.get(key)
+        keysym = xcbq.keysyms.get(key.lower())
 
         class DummyEv:
             pass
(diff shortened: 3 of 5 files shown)


m-col:
Nice, thank you for this. I have a couple Qs just so I understand fully.

> Modifier keys on the other hand require all-lowercase in qtile.

This is imposed by us, which is the reasoning behind the `masks.append(ModMasks[i.lower()])` change. Is that right?

Then the remaining changes make the x11 backend not care about the case of keys (including modifiers) that users specify in their configs.

Sounds good. I think there are a few other places where the keysyms are used that would need updating. I also wanted to see if we should just make the dict in `xkeysyms.py` lower case in the file, and doing so with this diff seems to be sufficient to stop the tests from failing, so I think we can do that, rather than keeping two copies of the dict around:
```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index 4087e333..609e6689 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -153,7 +153,7 @@ class Core(base.Core):
         self.qtile = None  # type: Qtile | None
         self._painter = None
 
-        numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["Num_Lock"])[0]
+        numlock_code = self.conn.keysym_to_keycode(xcbq.keysyms["num_lock"])[0]
         self._numlock_mask = xcbq.ModMasks.get(self.conn.get_modifier(numlock_code), 0)
         self._valid_mask = ~(
             self._numlock_mask
@@ -751,7 +751,7 @@ class Core(base.Core):
         """Simulates a keypress on the focused window."""
         # FIXME: This needs to be done with sendevent, once we have that fixed.
         modmasks = xcbq.translate_masks(modifiers)
-        keysym = xcbq.keysyms.get(key)
+        keysym = xcbq.keysyms.get(key.lower())
 
         class DummyEv:
             pass
@@ -851,4 +851,4 @@ class Core(base.Core):
 
     def keysym_from_name(self, name: str) -> int:
         """Get the keysym for a key from its name"""
-        return keysyms[name]
+        return keysyms[name.lower()]
diff --git a/libqtile/backend/x11/xcbq.py b/libqtile/backend/x11/xcbq.py
index ec65c2f2..16b9ef43 100644
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -790,7 +790,7 @@ class Painter:
 
 
 def get_keysym(key: str) -> int:
-    keysym = keysyms.get(key)
+    keysym = keysyms.get(key.lower())
     if not keysym:
         raise XCBQError("Unknown key: %s" % key)
     return keysym
@@ -812,7 +812,7 @@ def translate_masks(modifiers: list[str]) -> int:
     masks = []
     for i in modifiers:
         try:
-            masks.append(ModMasks[i])
+            masks.append(ModMasks[i.lower()])
         except KeyError as e:
             raise XCBQError("Unknown modifier: %s" % i) from e
     if masks:
diff --git a/libqtile/backend/x11/xkeysyms.py b/libqtile/backend/x11/xkeysyms.py
index 5da3f3cd..282de659 100644
--- a/libqtile/backend/x11/xkeysyms.py
+++ b/libqtile/backend/x11/xkeysyms.py
@@ -2205,3 +2205,4 @@ keysyms = {
     "braille_dots_2345678": 0x10028FE,
     "braille_dots_12345678": 0x10028FF,
 }
+keysyms = {k.lower(): v for k,v in keysyms.items()}
```

I think that would be preferable. What do you think?

I also noticed that the `rkeysyms` and `rdict` from `xcbq` seem to be old crusty code that is no longer needed, so we might as well rip that out if we can while we're here.

oddlama:
> > Modifier keys on the other hand require all-lowercase in qtile.
> 
> This is imposed by us, which is the reasoning behind the `masks.append(ModMasks[i.lower()])` change. Is that right?

Yes. All modifier keys seem to be defined by qtile, so I thought it would only be right to keep it consistent and make them ignore the case too.

> Then the remaining changes make the x11 backend not care about the case of keys (including modifiers) that users specify in their configs.

Exactly.

> Sounds good. I think there are a few other places where the keysyms are used that would need updating. I also wanted to see if we should just make the dict in `xkeysyms.py` lower case in the file [...]

Even better. I did a quick skim over the files last week but didn't find anything else at the time. As I wasn't sure whether changing the source definition would be acceptable, I just fixed everything related to the keybinds.
 
> [diff]. I think that would be preferable. What do you think?

Yes, but I think this is missing the `key.lower() in valid_keys` change in the testing file. So it should now fail a `qtile check` when an uppercase character is used. So that change should also be included.

> I also noticed that the `rkeysyms` and `rdict` from `xcbq` seem to be old crusty code that is no longer needed, so we might as well rip that out if we can while we're here.

Great idea. I'll update the PR to include your diff and this change right away if you don't mind.

libqtile/backend/x11/xcbq.py
```diff
@@ -69,8 +69,6 @@ def rdict(d):
     return r
 
 
-rkeysyms = rdict(keysyms)
```
m-col, line 72: We can remove the definition of `rdict` as well, as it was only used here.

m-col:
Excellent thanks, auto-merge enabled!

elParaguayo:
I think this may have inadvertently broken qtile-extras 😉 

The popup toolkit has keyboard navigation and the default keymap for that used some uppercase names e.g. `Left`.

Couldn't work out why so many tests were suddenly failing!

oddlama:
Oh no! Sorry about that. I though everything will be fine when the qtile tests succeed. Is extras another project?

Also shouldn't `Left` be allowed with case-insensitive logic? Or was the code directly accessing the keysyms dict?

elParaguayo:
Don't worry about it!

Yes, extras is a separate project so all that matters is for the qtile tests to pass. If extras breaks then I'll need to fix it.

And, yes, the code directly accessed the dict. Will look to fix later today.

PR #3648, opened 2022-06-20
CryptoTicker widget: support to Binance API

fedemahf:
I need to use the Binance API because Coinbase do not list the BNB cryptocurrency. The best approach I found is to add and configure manually all the required APIs in the widget.

Let me know if I am missing something!

### DIFF
--- a/libqtile/widget/crypto_ticker.py
+++ b/libqtile/widget/crypto_ticker.py
@@ -26,6 +26,7 @@
 
 import locale
 
+from libqtile.confreader import ConfigError
 from libqtile.log_utils import logger
 from libqtile.widget.generic_poll_text import GenPollUrl
 
@@ -35,8 +36,8 @@ _DEFAULT_SYMBOL = str(locale.localeconv()["currency_symbol"])
 
 class CryptoTicker(GenPollUrl):
     """
-    A cryptocurrency ticker widget, data provided by the coinbase.com API. Defaults to
-    displaying currency in whatever the current locale is. Examples:
+    A cryptocurrency ticker widget, data provided by the coinbase.com or the binance.com
+    API. Defaults to displaying currency in whatever the current locale is. Examples:
 
         # display the average price of bitcoin in local currency
         widget.CryptoTicker()
@@ -55,8 +56,14 @@ class CryptoTicker(GenPollUrl):
     """
 
     QUERY_URL_DICT = {
-        "coinbase": "https://api.coinbase.com/v2/prices/{}-{}/spot",
-        "binance": "https://api.binance.com/api/v3/ticker/price?symbol={}{}",
+        "coinbase": (
+            "https://api.coinbase.com/v2/prices/{}-{}/spot",
+            lambda x: float(x["data"]["amount"]),
+        ),
+        "binance": (
+            "https://api.binance.com/api/v3/ticker/price?symbol={}{}",
+            lambda x: float(x["price"]),
+        ),
     }
 
     defaults = [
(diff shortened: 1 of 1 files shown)


libqtile/widget/crypto_ticker.py
```diff
@@ -74,14 +82,36 @@ def __init__(self, **config):
...
             self.symbol = "$"
 
+        if self.api in self.QUERY_URL_DICT:
+            self.query_url = self.QUERY_URL_DICT[self.api]
+        else:
+            apis = sorted(self.QUERY_URL_DICT.keys())
+            logger.error(
+                "%s is not a valid API. Use one of the list: %s.",
+                self.api,
+                str(apis),
```
m-col, line 92: I think this `str` call is redundant because the `"%s"` in the format string will do it.

libqtile/widget/crypto_ticker.py
```diff
@@ -74,14 +82,36 @@ def __init__(self, **config):
...
+
         variables = dict()
         variables["crypto"] = self.crypto
         variables["symbol"] = self.symbol
-        variables["amount"] = float(body["data"]["amount"])
+        variables["amount"] = self.getAmountFromBodyResponse(body)
 
         return self.format.format(**variables)
+
+    def getAmountFromBodyResponse(self, body):
```
m-col, line 110: Please could you convert this to snake_case, just to be consistent with the code base? thanks

libqtile/widget/crypto_ticker.py
```diff
@@ -74,14 +82,36 @@ def __init__(self, **config):
...
+        if (self.api == "binance"):
+            return float(body["price"])
+        
+        if (self.api == "coinbase"):
+            return float(body["data"]["amount"])
+
+        return 0.0
```
m-col, line 117: I guess this 0.0 will be reached in the case where the `api` parameter that was passed from the config wasn't found in `QUERY_URL_DICT`. When that happens it might be best to raise a `ConfigError` (from `libqtile.confreader`) from within the `_configure` method (rather than `__init__`). This causes the widget to be replaced in the bar by a text widget that tells the user that the widget was invalid, and the original widget doesn't get set up. Then we can assume from this method that if `self.api == "binance"` is False, that it _must_ be `"coinbase"`. The `_configure` method would look something like this:
```py
        if self.api not in self.QUERY_URL_DICT:
            raise ConfigError("Unknown provider passed as 'api' to Cryptoticker")

        GenPollUrl._configure(self, qtile, bar)
        self.query_url = self.QUERY_URL_DICT[self.api]
```
(untested)

m-col (changes requested):
This looks great, thank you for this! I've added some minor review comments. Could you also update the top line of the widget docstring, to mention that it supports binance in addition to coinbase?

elParaguayo:
Argh. Did my review comments on this not get submitted?

I had one extra comment/suggestion: rather than using an `if` statement to lookup the relevant api name and have the code to parse the data accordingly in the body of the widget, should we have some dictionary of api providers with the url and a data parsing method? That way, users could provide a custom api more easily as it's not hard coded.

This is how we do it in the `CheckUpdates` widget for example.

libqtile/widget/crypto_ticker.py
fedemahf: The function was removed. Commit: 25096b15e9602a690d0649e79e195f7ecdce0ccc

libqtile/widget/crypto_ticker.py
fedemahf: That's correct. The `str` call was removed. Commit: 25096b15e9602a690d0649e79e195f7ecdce0ccc

libqtile/widget/crypto_ticker.py
fedemahf: I made some modifications to your code to adapt it to the new dictionary. Take a look: 25096b15e9602a690d0649e79e195f7ecdce0ccc

fedemahf:
> Argh. Did my review comments on this not get submitted?
> 
> I had one extra comment/suggestion: rather than using an `if` statement to lookup the relevant api name and have the code to parse the data accordingly in the body of the widget, should we have some dictionary of api providers with the url and a data parsing method? That way, users could provide a custom api more easily as it's not hard coded.
> 
> This is how we do it in the `CheckUpdates` widget for example.

I wanted to do this, but my Python skills are a little rusty. I did a quick search for lambda functions and implemented it. I'm not sure if I did it right. Commit: 25096b15e9602a690d0649e79e195f7ecdce0ccc

m-col (approved):
Awesome, thanks!

m-col:
CI looks to have stalled or something, but they passed on your branch so I suspect they are all fine. I'll merge anyway, and tests will re-run on master. If things fail there I'll submit a quick fix to update it. Thanks again, and sorry for taking ages to review again :)

PR #3649, opened 2022-06-21
Add graceful call to unbind keys in dgroups simple key binder

okgarces:
Hi all,

I am a newbie user and have a weird behavior with DGroups keybindings. When I close emacs the Dgroups keybindings stop working until I restart Qtile.
Looking at the logs I got this:

image

Not sure if this is useful for the community but it solved my issues. Please let me know if it doesn't make sense and close this PR.

Many thanks.

Cheers,

### DIFF
--- a/libqtile/dgroups.py
+++ b/libqtile/dgroups.py
@@ -37,11 +37,17 @@ from libqtile.log_utils import logger
 def simple_key_binder(mod, keynames=None):
     """Bind keys to mod+group position or to the keys specified as second argument"""
 
+    def _graceful_fail_call(func, param):
+        try:
+            func(param)
+        except KeyError:
+            logger.warning('Key not found in dgroup keys: "%s"', str(param))
+
     def func(dgroup):
         # unbind all
         for key in dgroup.keys[:]:
-            dgroup.qtile.ungrab_key(key)
-            dgroup.keys.remove(key)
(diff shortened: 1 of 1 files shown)


m-col:
Hi there! and thanks for the contribution!

I do wonder though - why is the issue even happening? I think i it would be best to work out the source of the problem rather than patch over it. How is your `key_binder` function defined?

okgarces:
Hi! Thanks for the quick answer.
You are right, and I do believe it's much better to solve a problem from the source instead of creating patches.

```
from libqtile.dgroups import simple_key_binder
dgroups_key_binder = simple_key_binder("mod4")
```

That's how my `key_binder` is defined. I haven't changed much, I am using the default `simple_key_binder`.

The behaviour occurs when I quit emacs and sometimes a browser or electron-based apps, the groups' key binder stops working. Debugging it seems it's a race condition issue (sorry for the ignorance in Qtile if this is not a problem). 

The `simple_key_binder` function's first step is to make sure the keys are being removed from Qtile but at some point, it is trying to delete inexistent keys. 
Maybe instead of adding this patch, the keys could be safely removed from Qtile. 

Let me know, what do you reckon? I want to learn if it's possible.

okgarces:
Closing to reduce clutter in PRs.

PR #3653, opened 2022-06-22
Fix dependency in the documentation.

CharString:
xdg is a different package that doesn't provide xdg.IconTheme


CharString:
@elParaguayo No problem. It was a bit puzzling because both packages use the same namespace.
Is it right that it doesn't show the nm-applet icon? It seems missing, but it will show the redshit-gtk one (wihtout the menu, but I need to checkout your other repo for that).

PR #3655, opened 2022-06-24
Avoid raising windows to the top of the stack.

### LINKED ISSUE #676: Windows spawned after certain screen lockers wrongly drawn on top

QBobWatson:
As discussed in [this email](https://www.mail-archive.com/qtile-dev@googlegroups.com/msg00629.html) from 2017, ordinarily window managers reorder managed windows amongst themselves (e.g. using XRestackWindows) but never bring them to the top of the X server's stack (above override-redirect windows) using e.g. XRaiseWindow.  qtile uses the xcb equivalent of XRaiseWindow (ConfigureWindow with StackMode.Above and no sibling), which causes undesirable effects:

- Long-lived notification popups (e.g. lxqt-notificationd) get covered by new windows.
- New windows can appear on top of a lock screen (#676).

It seems silly for a tiling window manager to keep its own window stack, so this PR has a simpler solution.  In backend.x11.core.Core.distribute_windows, an offscreen (but mapped) window called stacking_window is created and pushed to the bottom of the stack.  Then all instances of ConfigureWindow with StackMode.Above and no sibling are replaced by ConfigureWindow with StackMode.Below and sibling=stack_window.  In this way qtile can use the X server to keep track of stacking order directly (as it did before), but it ensures that managed windows stay below override-redirect windows, as is conventional.

This fixes #676.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -162,6 +162,8 @@ class Core(base.Core):
             | xcbq.PointerMotionHintMask
         )
 
+        self.stack_window = None
+
     @property
     def name(self):
         return "x11"
@@ -233,9 +235,16 @@ class Core(base.Core):
                     win.set_group()
             return
 
+        if not self.stack_window:
+            self.stack_window = self.conn.create_window(-1, -1, 1, 1)
+        self.stack_window.map()
+        self.stack_window.configure(stackmode=StackMode.Below)
+
         # Qtile just started - scan for clients
         _, _, children = self._root.query_tree()
         for item in children:
+            if item.wid == self.stack_window.wid:
+                continue
             try:
                 attrs = item.get_attributes()
                 state = item.get_wm_state()
@@ -256,6 +265,9 @@ class Core(base.Core):
             if item.wid in self.qtile.windows_map:
                 win = self.qtile.windows_map[item.wid]
                 win.unhide()
+                # This window was created before distribute_windows() was
+                # called (e.g. a bar); push it below self.stack_window
+                self.raise_window(win.window)
                 return
 
             if internal:
@@ -761,6 +773,10 @@ class Core(base.Core):
         d.state = modmasks
         self.handle_KeyPress(d)
 
+    def raise_window(self, window):
+        if self.stack_window:
+            window.configure(sibling=self.stack_window.wid, stackmode=StackMode.Below)
+
     def focus_by_click(self, e, window=None):
         """Bring a window to the front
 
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks for looking at this.

There's a broader PR for window layering (#3409) so I'd prefer to solve the issue within that PR rather than add a new layering solution here.

PR #3661, opened 2022-06-25
WL: add commands to hide and unhide the cursor

### LINKED ISSUE #3651: [ Feature ] Allow users to toggle cursor on and off under both X and Wayland
### The issue:

I think a missing feature of qtile that would be very useful to most users would be an option to be able to toggle the cursor on and off under both X and wayland. Something similar to unclutter but allowing the user to toggle it via scripts or keybindings.

### Required:

m-col:
This functionality is needed for feature parity with X11, but
unfortunately is not (yet) available through a protocol, and so third
party utilities cannot do this for us.

This adds two commands:

- core.hide_cursor: hides the cursor until it is unhidden again
- core.unhide_cursor: unhides it, if it was hidden

Example usage would be via a tool such as swayidle to hide the cursor
when input has idled for a while, and then unhide the cursor when input
starts again.

Closes #3651


justinesmithies:
This works perfectly for me thank you ;)

m-col:
no probs, let me know if you think it isn't working as it should

Issue #3634, opened 2022-06-16
Qtile check breaks


Malex:
### The issue:

Qtile check breaks on some parameter passed to shutil.copytree on python 3.7

`foijeaf@localhost ~ $ qtile check
checking qtile config file /home/foijeaf/.config/qtile/config.py
Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.7/qtile", line 33, in <module>
    sys.exit(load_entry_point('qtile==0.21.0', 'console_scripts', 'qtile')())
  File "/usr/lib/python3.7/site-packages/libqtile/scripts/main.py", line 66, in main
    options.func(options)
  File "/usr/lib/python3.7/site-packages/libqtile/scripts/check.py", line 118, in check_config
    shutil.copytree(path.dirname(args.configfile), tempdir, dirs_exist_ok=True)
TypeError: copytree() got an unexpected keyword argument 'dirs_exist_ok'`

Qtile version is 0.21.0

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

LordXerus:
dirs_exist_ok is a parameter added in Python 3.8

m-col:
`qtile check` does not support Python 3.7 as `Typing` is used in the codebase, which was added in 3.8. Had you not hit that error, qtile check would have printed an error saying this message. That's pretty ironic if the message can't even be reached on py37, so that should definitely be fixed. Regardless, we typically only support the most recent 3 major of versions... 3.7 is pretty old. It may not be a very satisfying answer, sorry, but I suggest upgrading.

Malex:
> `qtile check` does not support Python 3.7 as `Typing` is used in the codebase, which was added in 3.8. Had you not hit that error, qtile check would have printed an error saying this message. That's pretty ironic if the message can't even be reached on py37, so that should definitely be fixed. Regardless, we typically only support the most recent 3 major of versions... 3.7 is pretty old. It may not be a very satisfying answer, sorry, but I suggest upgrading.

I perfectly understand your reasoning, I myself would use a more recent version of python, but python is not considered stable on my distro, for whatever reason (I'm on Funtoo). Not your problem, it's just it was not clear which versions are supported: the README does not state it, the setup.py had 3.7 as one of the version.
I think though that qtile check should (or could) fail more "elegantly", which a message clearly stating check is not supported for <3.8, and/or clearly state what qtile supports and what qtile "should mostly work on but some stuff will fail" (3.7)

[...]

PR #3662, opened 2022-06-26
small improvements to qtile check

m-col:
The main change here is that dependencies for the script are all checked
at the very beginning. Currently there is an issue on python 3.7 where
the error message that tells the user that 3.7 isn't supported doesn't
even get reached! This fixes that.

Also some minor cosmetic changes.

See #3634


elParaguayo:
LGTM.

PR #3663, opened 2022-06-26
WL: implement input inhibitor protocol

m-col:
This is an attempt at implementing the input inhibitor protocol, built
upon the work done by @Graeme22 in #2852 (thanks @Graeme22!).

The main additional step made by this draft over #2852 is blocking
pointer and keyboard focus from going to surfaces that belong to other
clients, while still (in theory) allowing focus to move to other
surfaces belonging to the same client. All bindings (button/scroll/key)
become disabled.

One outstanding issue that prevents this from being a complete and
precise implementation of the protocol is that qtile commands executed
in the background proceed as normal. E.g. you can change group from a
script if you start the script, then start the input inhibitor, and then
the script changes group. On the plus side, input _is_ inhibited, so
this doesn't do much. In the case where a screenlocker is used via the
layer shell protocol, it covers the screen anyway, and so
layer_shell+input_inhibit screen locking seems usable. Unless a sneaky
process ran `qtile cmd-obj -o window -f kill` in the background, killing
it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Add `force_update` command to `ThreadPoolText` widgets to simplify updating from key bindings
         - Add scrolling ability to `_TextBox`-based widgets.
         - Add player controls (via mouse callbacks) to `Mpris2` widget.
+        - Wayland: input inhibitor protocol support added (pywayland>=0.4.14 & pywlroots>=0.15.19)
     * bugfixes
         - Widgets that are incompatible with a backend (e.g. Systray on Wayland) will no longer show 
           as a ConfigError in the bar. Instead the widget is silently removed from the bar and a message
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -39,6 +39,7 @@ from wlroots.wlr_types import (
     ExportDmabufManagerV1,
     ForeignToplevelManagerV1,
     GammaControlManagerV1,
+    InputInhibitManager,
     OutputLayout,
     PrimarySelectionV1DeviceManager,
     RelativePointerManagerV1,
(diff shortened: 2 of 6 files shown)


m-col:
Anyone who is interested in this, please could pull it down and give it a test out? With some testing here it seems to be usable in general, so it may not need any changes for screen locking. Any feedback and bug reporting is appreciated.

m-col:
Totally forgot, I made these changes to pywlroots, which are needed:
```diff
diff --git a/wlroots/wlr_types/input_inhibit.py b/wlroots/wlr_types/input_inhibit.py
index 82c6472..e9e5600 100644
--- a/wlroots/wlr_types/input_inhibit.py
+++ b/wlroots/wlr_types/input_inhibit.py
@@ -10,6 +10,8 @@ class InputInhibitManager(Ptr):
         """Creates a wlr_input_inhibit_manager"""
         self._ptr = lib.wlr_input_inhibit_manager_create(display._ptr)
 
+        self.activate_event = Signal(ptr=ffi.addressof(self._ptr.events.activate))
+        self.deactivate_event = Signal(ptr=ffi.addressof(self._ptr.events.deactivate))
         self.destroy_event = Signal(ptr=ffi.addressof(self._ptr.events.destroy))
 
     def is_inactive(self) -> bool:

```

m-col:
... which is at this PR: https://github.com/flacjacket/pywlroots/pull/92

justinesmithies:
Tested this PR and I can indeed run swaylock -F -i ~/.cache/wallpaper --effect-blur 10x5 --clock --indicator or just plain swaylock but it will not give me any keyboard input to unlock it again. I did find this in my [qtile.log](https://paste.sr.ht/~justinesmithies/0dd2bcf7cf774a498d8e1c7840a7d51b89cacd2c) though if it helps ? I'm using all the latest pywlroots and pywayland and so on.

m-col:
Oh sorry, it looks like I made a change to pywayland for which I never made a PR. Will do that now.

arjan-s:
I'm running with these patches now as well, and everything works as expected. Much better than the previous implementation!

About the qtile commands in the background: I would actually expect that they keep running normally by default. But perhaps you'd want to block them when a certain config option is set, so that users with more strict requirements can make sure they are blocked as well (I know I'd enable such an option).

m-col:
Requires https://github.com/flacjacket/pywlroots/pull/98 in addition to https://github.com/flacjacket/pywayland/pull/39

PR #3664, opened 2022-06-27

karthink:
when_floating is None by default
when set to True, apply lazy call to floating windows only
when set to False, apply lazy call to tiled windows only

Mention when_floating change in CHANGELOG

Add example of new when_floating usage.

---

This is a redo of PR #2900 (see also issue #2846). The previous PR was closed because I couldn't get to some formatting issues with the CHANGELOG in time.

The idea is to enable commands to trigger under any of these conditions:
- _only_ for floating windows,
- _only_ for tiled windows, and
-  for _all_ windows irrespecitve of floating state.

Example:

``` python
    Key([mod, mod3], "h",
        lazy.layout.shuffle_left().when(when_floating=False),
        lazy.window.move_floating(-32, 0).when(when_floating=True),
        desc="Move window to the left"),
    Key([mod, mod3], "l",
        lazy.layout.shuffle_right().when(when_floating=False),
        lazy.window.move_floating(32, 0).when(when_floating=True),
        desc="Move window to the right"),
    Key([mod, mod3], "j",
        lazy.layout.shuffle_down().when(when_floating=False),
        lazy.window.move_floating(0, 32).when(when_floating=True),
        desc="Move window down"),
    Key([mod, mod3], "k",
        lazy.layout.shuffle_up().when(when_floating=False),
        lazy.window.move_floating(0, -32).when(when_floating=True),
        desc="Move window up"),
```

So you can do more with the same keys.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -67,6 +67,9 @@ Qtile 0.22.0, released 2022-09-22:
         - `KeyChord`'s signature has changed. ``mode`` is now a boolean to indicate whether the mode should persist.
           The ``name`` parameter should be used to name the chord (e.g. for the ``Chord`` widget).
     * features
+        - lazy.<object>.when(when_floating=X) now behaves differently: the lazy call will be executed
+          independently of the current window's float state by default, and can be limited to when it
+          is floating or tiled by passing when_floating as True or False respectively.
         - Add ability to draw borders and add margins to the `Max` layout.
         - The default XWayland cursor is now set at startup to left_ptr, so an xsetroot call is not needed to
           avoid the ugly X cursor.
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -97,8 +97,11 @@ conditions are supported:
             lazy.layout.grow_down().when(layout='columns')
         ),
 
-        # Limit action to when the current window is not floating (default True)
+        # Limit action to when the current window is not floating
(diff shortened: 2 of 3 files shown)


jwijenbergh:
Hey sorry for not looking at this earlier. Do you think we can add a test for this?

karthink:
Sure. Is there a guide for adding tests to qtile? I'll rebase the PR in the meantime.

jwijenbergh:
There is https://docs.qtile.org/en/stable/manual/contributing.html?highlight=writing%20tests#unit-testing but it does not contain any detailed info. Maybe you can add some tests to the existing when check tests:

https://github.com/qtile/qtile/blob/master/test/test_when.py

jwijenbergh:
The way these tests work is that we have the config where keys are defined. We then create test windows with the `manager.test_window` methods and do operations on them with the keybindings. We then query the properties of these windows.

I think for your tests it might make sense to e.g. do something more like the last two keybindings in the test where we switch layouts and then test if we have actually switched layout. You could then define 3 keybindings based on True, False, None for when_floating and test it on different floating states of a test window.

To float a window you can use `manager.c.window.enable_floating()`. 

You can then push a test and I will approve so that we can see if anything passes. If it doesn't I will help to fix it

jwijenbergh:
@karthink should I make a follow up pr with the test added?

karthink:
> @karthink should I make a follow up pr with the test added?

Please do -- I haven't used qtile in a while now.

PR #3670, opened 2022-07-01
Fix libffi version issue in CI

elParaguayo:
New release of `cffi` (1.15.1) seems to break cairocffi so pin at version 1.15.0 for now.


elParaguayo:
Breakage seems to be on Ubuntu 20.04 as that is the environment for both CI and readthedocs. No failures when building in a clean Arch linux nspawn container.

PR #3684, opened 2022-07-08
Remove `Gap` objects from Command Graph

### LINKED ISSUE #3668: Qtile shell should better handle exceptions
### The issue:

I was playing around with the qtile shell, specifically I was exploring the folder structure.
When I tried to ls a folder defined as None, the shell broke down and dropped me back to the terminal.

I think the exception should be catched, the error message shown but to keep being in the the shell.

Here's the example:
`bar[bottom] > ls widget/
  File "/usr/lib/python-exec/python3.8/qtile", line 33, in <module>
    sys.exit(load_entry_point('qtile==0.21.0', 'console_scripts', 'qtile')())
  File "/usr/lib/python3.8/site-packages/libqtile/scripts/main.py", line 66, in main
    options.func(options)
  File "/usr/lib/python3.8/site-packages/libqtile/scripts/shell.py", line 36, in qshell
    qsh.loop()
  File "/usr/lib/python3.8/site-packages/libqtile/sh.py", line 360, in loop
    val = self.process_line(line)
  File "/usr/lib/python3.8/site-packages/libqtile/sh.py", line 317, in process_line
    val = builtin(args)
  File "/usr/lib/python3.8/site-packages/libqtile/sh.py", line 236, in do_ls
    objects, items = self._ls(node, rest_path)
  File "/usr/lib/python3.8/site-packages/libqtile/sh.py", line 127, in _ls
    str_items = [str(i) for i in items]
TypeError: 'NoneType' object is not iterable
`
I dont have a bottom bar (it's on the top)

Qtile version is 0.21.0

### Required:

elParaguayo:
`Gap` objects have no real use on the graph and can cause issues in `qtile shell` as they don't expose the same commands as bars.

Fixes #3668

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -496,7 +496,7 @@ class Screen(CommandObject):
         elif name == "window" and self.group is not None:
             return True, [i.wid for i in self.group.windows]
         elif name == "bar":
-            return False, [x.position for x in self.gaps]
+            return False, [x.position for x in self.gaps if isinstance(x, Bar)]
         elif name == "widget":
             bars = (g for g in self.gaps if isinstance(g, Bar))
             return False, [w.name for b in bars for w in b.widgets]
@@ -520,7 +520,10 @@ class Screen(CommandObject):
                         return i
         elif name == "bar":
             assert isinstance(sel, str)
-            return getattr(self, sel)
+            bar = getattr(self, sel)
+            if isinstance(bar, Bar):
+                return bar
+
         elif name == "widget":
             for gap in self.gaps:
                 if not isinstance(gap, Bar):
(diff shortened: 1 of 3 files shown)


libqtile/config.py
```diff
@@ -496,7 +496,7 @@ def _items(self, name: str) -> ItemT:
         elif name == "window" and self.group is not None:
             return True, [i.wid for i in self.group.windows]
         elif name == "bar":
-            return False, [x.position for x in self.gaps]
+            return False, [x.position for x in self.gaps if isinstance(x, CommandObject)]
```
m-col, line 499: I think it would be more readable to test here and at the other 3 `isinstance` calls added below that the variable is a `Bar` directly rather than `CommandObject`, because it makes it clear to the reader that we're choosing bars and ignoring gaps

m-col (approved):
one comment but otherwise LGTM

PR #3685, opened 2022-07-08
Fix pyxdg refs in docs

elParaguayo:
LaunchBar and StatusNotifier docs both reference pyxdg dependency but rst gets confused when you have links with the same name.

One fix is to anonymise the links by using a double trailing underscore.


m-col (approved):
LGTM

PR #3694, opened 2022-07-11
Make Cmus widget use filename as fallback

MNandor:
Hi! I use Qtile, Cmus, and the Cmus widget inside Qtile. My music collection exists as local files. I noticed something that could be improved.

I've brought some screenshots to justify my changes.

If I'm playing music in the Cmus music player, then the Qtile widget will display its artist and title (seen bottom of screen in green).
image

This is done by getting the metadata from the music files. If there is no metadata though, we get empty strings, as seen below.
image

Adding this code makes it so the widget uses the file's name as a fallback text to display instead of just " - ".
image

Additional considerations:
- use fallback even if Artist is set but Title is not?
- remove file extension from fallback (.mp3 in this case)

### DIFF
--- a/libqtile/widget/cmus.py
+++ b/libqtile/widget/cmus.py
@@ -99,6 +99,10 @@ class Cmus(base.ThreadPoolText):
             if self.local:
                 artist = info["artist"]
                 now_playing = "{0} - {1}".format(artist, title)
+                if now_playing == " - ":
+                    file_path = info["file"]
+                    file_path = file_path.split("/")[-1]
+                    now_playing = file_path
             else:
                 if info["stream"]:
                     now_playing = info["stream"]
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for this.

Ideally, could you add a test in `test/widgets/test_cmus.py` too?

elParaguayo:
Test looks good. The CI failure is because you need two blank lines between functions in the test file so please fix that.

Can you then squash this all into a single commit and force push it again.

MNandor:
@elParaguayo Thanks for pointing that out. Added the extra line!

MNandor:
Ah, sorry! Just caught the squash "this all". Done so, how it's fine like this.

elParaguayo:
Great. Thanks.

PR #3704, opened 2022-07-14
Wayland: Separate window types, fix app_id propagation 

### LINKED ISSUE #3705: Wayland: Alacritty app_id not recognized
### The issue:

Seeing how different terminals behave in a dropdown:

* kitty: Works normally
* foot: Works normally
* Alacritty: Gets treated as a normal window, not matched.

So, attempting to match via app_id:

```
DropDown('term', 'alacritty --class foo,foo', match=Match(wm_instance_class='foo'))
```

This does not work either - same behavior. A closer look in `qtile shell` confirms that qtile returns the window `wm_class` as `None` for Alacritty, but correctly for the other terminals.

### Required:

3nprob:
This does two things:

* Refactor `backend/wayland/window.py`, breaking out implementations into individual files
  * As a newcomer to the codebase, this made the code a lot more approachable
* Detect changed app_id in _on_map of XdgWindow. This fixes #3705


3nprob:
The CI errors seem unrelated to these changes.

3nprob:
Temporarily basing on #3670 to unbreak CI

elParaguayo:
> Temporarily basing on #3670 to unbreak CI

FYI that PR is unlikely to get merged - the CI issues seem to be linked to building on Ubuntu 20.04 (but I'm still looking into it).

I'll leave the review of the Wayland bits to @m-col...

3nprob:
In the meantime, would be great to make sure the python workflows pass

elParaguayo:
> In the meantime, would be great to make sure the python workflows pass

Agreed - need to get this fixed as we won't merge any PRs until that's solved. However, we're reluctant to push a fix that is only needed on some distros.

3nprob:
> > In the meantime, would be great to make sure the python workflows pass
> 
> Agreed - need to get this fixed as we won't merge any PRs until that's solved. However, we're reluctant to push a fix that is only needed on some distros.

Oh, I was just referring to triggering a run as-is here: 
image

m-col:
It will be a little while before I will have time to give this a proper review but as a first glance it all looks great, so thank you for working on this! I agree the structure was a bit intimidating and if this improves how approachable it is then that is fantastic.

I also noticed that `SubSurface` now exclusively belongs to `layer.py`, whereas before subsurfaces were also integrated in the code into `XdgSurface`. Is there a reason for this change? Likewise are there other functional changes that should be noted?

elParaguayo:
> Oh, I was just referring to triggering a run as-is here: 

Ha. Ok. Pretty sure it will fail (and not because of your changes) - but let's see!

3nprob:
> I also noticed that `SubSurface` now exclusively belongs to `layer.py`, whereas before subsurfaces were also integrated in the code into `XdgSurface`. Is there a reason for this change?

Not sure exactly what you're referring to here - `XdgSurface` comes from the external `wlroots` package. As for the references to `SubSurface` in the classes which are now in `xdgwindow.py`, it's imported there. So there should be no change there.

> Likewise are there other functional changes that should be noted?

The only functional change (aside from a [minor type guard](https://github.com/qtile/qtile/pull/3704/files#diff-be6027be0d0dd2f4835fb869cbae1f6f894f6359dfe7e9f849101a467b14e375R815) and [logging change](https://github.com/qtile/qtile/pull/3704/files#diff-be6027be0d0dd2f4835fb869cbae1f6f894f6359dfe7e9f849101a467b14e375R819)) is in 06d22ba98b18d15a606397335d2b8b5d0a13370e.

3nprob:
> Ha. Ok. Pretty sure it will fail (and not because of your changes) - but let's see!
> 
> EDIT: forgot you included my "fix" so the failures may be yours after all wink

Thanks!

```
libqtile/backend/wayland/xdgwindow.py:198:41: BLK100 Black would make changes.
```

Avast, foiled by BLK! :skull_and_crossbones:

elParaguayo:
Great. Tests pass so we just need to find the proper fix to the CI errors, rather than my hack.

3nprob:
Rebased on master, CI commit removed.

libqtile/backend/wayland/core.py
```diff
@@ -833,7 +833,7 @@ def _focus_by_click(self) -> None:
         found = self._under_pointer()
 
         if found:
-            win, surface, _, _ = found
+            win, _, _, _ = found
```
m-col, line 836: Huh, I wonder why this wasn't detected as an unused variable before.

m-col (approved):
All looks excellent to me, thanks again for this, it was definitely needed. It'll auto-merge if the CI passes.

PR #3707, opened 2022-07-15
Add dev requirements

3nprob:
Add some missing dependencies and adds a section in the docs on addressing common linting errors.

### DIFF
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -116,11 +116,11 @@ request that git commit messages follow the
 Some common linting errors and how to resolve them:
 
 ==================================================== =======================================================
-Lint Error                                           Verbose Diff Command
+Lint Error                                           Fix
 ==================================================== =======================================================
-``I001 isort found an import in the wrong position`` ``python -m flake8 ./libqtile/ --isort-show-traceback``
-``I005 isort found an unexpected missing import``    ``python -m flake8 ./libqtile/ --isort-show-traceback``
-``BLK100 Black would make changes.``                 ``python -m black --diff ./libqtile/``
+``I001 isort found an import in the wrong position`` ``python -m isort ./libqtile/``
+``I005 isort found an unexpected missing import``    ``python -m isort ./libqtile/``
+``BLK100 Black would make changes.``                 ``python -m black ./libqtile/``
 ==================================================== =======================================================
 
 Logging
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -113,6 +113,16 @@ to ensure your patch complies with reasonable formatting constraints. We also
 request that git commit messages follow the
 `standard format <https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`_.
 
+Some common linting errors and how to resolve them:
+
+==================================================== =======================================================
+Lint Error                                           Verbose Diff Command
+==================================================== =======================================================
+``I001 isort found an import in the wrong position`` ``python -m flake8 ./libqtile/ --isort-show-traceback``
+``I005 isort found an unexpected missing import``    ``python -m flake8 ./libqtile/ --isort-show-traceback``
+``BLK100 Black would make changes.``                 ``python -m black --diff ./libqtile/``
+==================================================== =======================================================
+
 Logging
 =======
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,7 @@
+bowler
+dbus-next
 flake8
+flake8-black
 flake8-isort
 flake8-tidy-imports
 mypy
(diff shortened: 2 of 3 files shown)


docs/manual/hacking.rst
```diff
@@ -113,6 +113,16 @@ to ensure your patch complies with reasonable formatting constraints. We also
...
 `standard format <https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`_.
 
+Some common linting errors and how to resolve them:
+
+==================================================== =======================================================
+Lint Error                                           Verbose Diff Command
+==================================================== =======================================================
+``I001 isort found an import in the wrong position`` ``python -m flake8 ./libqtile/ --isort-show-traceback``
+``I005 isort found an unexpected missing import``    ``python -m flake8 ./libqtile/ --isort-show-traceback``
+``BLK100 Black would make changes.``                 ``python -m black --diff ./libqtile/``
```
elParaguayo, line 123: Any reason why you wouldn't just let `isort` and `black` make the changes directly? You can view the changes with `git diff` afterwards.

requirements-dev.txt
```diff
@@ -1,10 +1,14 @@
+bowler
 flake8
+flake8-black
 flake8-isort
 flake8-tidy-imports
 mypy
+sphinx
```
elParaguayo, line 7: Isn't this just for building the docs? It's already included in `docs/requirements.txt` so shouldn't be needed here.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 cairocffi >= 0.9.0
 cffi >= 1.1.0
+dbus-next
```
elParaguayo, line 3: `dbus-next` is an optional requirement so shouldn't be in the main `requirements.txt` file. It can be in `requirements-dev.txt` though.

elParaguayo (changes requested):

m-col:
Nice, I was recently thinking we could do with overhauling the requirements files so thanks for looking at these. Like @elParaguayo, we have a specific one for the docs. I think we should then have a folder to contain them because there are a few sets of dependencies with different roles. Open for debate is how best to structure those, but for example we have a 'base' set of deps, we have x11 deps, we have wayland deps, and we have 'additional' deps (e.g. dbus-next, widget-specific deps), and we have dev deps. Thoughts?

docs/manual/hacking.rst
3nprob: Mainly that I don't know if/how there exists a command for flake8/isort for that so consistency. All ears, though.

elParaguayo: I just run `black libqtile` or `isort libqtile`...

elParaguayo:
To run test suite, you should use `tox` as that installs the relevant dependencies in a separate environment.

docs/manual/hacking.rst
3nprob: So you would rather see the destructive project-wide auto-fix than the diff for this to be merged? Can change if so

3nprob: Changed.

3nprob:
> To run test suite, you should use `tox` as that installs the relevant dependencies in a separate environment.

Just running `$ TOXENV=codestyle,pep8,vulture,mypy tox` as triggered by `make lint` or anything else worth writing mentioning explicitly?

elParaguayo:

m-col:
Agreed @3nprob. All sounds good.

The widget deps are separate to the backend deps (e.g. you need `psutil` for the `CPU` widget) but some of their deps are actually non-Python (`Cmus` widget needs the `cmus` software installed), so maybe having a widget deps requirements file is overkill. They document their deps individually anyway. As for the backends and base deps I can do that in another PR. My main motivation for that is pinning wayland dep minimum versions which is currently done in 3 places in tox.ini and setup.cfg so it would be nice to have that in one place. Again, I dont think any of this needs to be done for this PR.

m-col:
LGTM so I've merged master onto it and enabled auto-merge. Thanks.

3nprob:
Anything still outstanding on this one?

m-col:
> Anything still outstanding on this one?

I don't think so. The docs failed to build, and it looks like it's unrelated to this PR and to do with ubuntu cairo-related packages. @elParaguayo is this what you mentioned recently, and perhaps what https://github.com/qtile/qtile/pull/4203 aimed to address? If so, should we hold on this or are you happy to merge PRs in the meantime until the RTD build gets fixed?

elParaguayo:
I'm ok to merge this one. `requirements-dev.txt` isn't used by ReadTheDocs and there are no meaningful changes to `requirements.txt`.

elParaguayo:
Although... that is a different build error to the last one I saw.

PR #3711, opened 2022-07-16
Add `single_border_width` and `single_margin` to VerticalTile layout

luisalcarasr:
Add `single_border_width` and `single_margin` to the `VerticalTile` layout to be used when there is only one window.

### DIFF
--- a/libqtile/layout/verticaltile.py
+++ b/libqtile/layout/verticaltile.py
@@ -136,10 +136,7 @@ class VerticalTile(_SimpleLayoutBase):
                 border_color = self.border_normal
 
             # width
-            if n > 1:
-                width = screen_rect.width - self.border_width * 2
-            else:
-                width = screen_rect.width
+            width = screen_rect.width - border_width * 2
 
             # height
             if n > 1:
@@ -158,7 +155,7 @@ class VerticalTile(_SimpleLayoutBase):
                 else:
                     height = normal_pane_height
             else:
-                height = screen_rect.height
+                height = screen_rect.height - border_width * 2
 
             # y
             y = screen_rect.y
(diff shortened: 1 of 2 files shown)


libqtile/layout/verticaltile.py
```diff
@@ -120,7 +122,16 @@ def configure(self, window, screen_rect):
...
+                border_width = self.single_border_width
+
+            # margin
+            if n > 1:
+                margin = self.margin
+            else:
+                if self.single_margin != None:
+                    margin = self.single_margin
+                else:
+                    margin = self.margin
```
elParaguayo, line 134:
```suggestion
            if n == 1 and self.single_margin is not None:
                margin = self.single_margin
            else:
                margin = self.margin
```

elParaguayo (changes requested):
Looks good. Thanks for doing this.

I've suggested a change to tidy up the code a bit.

Can you also add a test for this?

libqtile/layout/verticaltile.py
luisalcarasr: That looks better, Python is great. Thanks!

elParaguayo:

Are you able to write a test or would you want some help with that?

luisalcarasr:
@elParaguayo, I will try to write some tests for this, if I have problems I would appreciate the help.

luisalcarasr:
Writing the tests I have detected an error with the `single_border_width`, the dimensions of the window are not recalculated. I'm working on the fix.

luisalcarasr:
@elParaguayo, I've written the tests and fixed the bugs, let me know any other comments.

luisalcarasr:
Should I add something to the `CHANGELOG`?
Is the name of the parameters correct? Some layouts use other names.

luisalcarasr:
No longer needed, #3990 solved the requirement.

PR #3717, opened 2022-07-23
cmd_spawn: replace os.fork with Popen

elParaguayo:
Just an idea...

As far as I know, `subprocess.Popen` uses `os.fork` and `os.execvp` so we could simplify our code here.

Thoughts?

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1141,86 +1141,26 @@ class Qtile(CommandObject):
             cmd = subprocess.list2cmdline(args)
 
         to_lookup = args[0]
-        if shell:
-            args = ["/bin/sh", "-c", cmd]
 
         if shutil.which(to_lookup) is None:
             logger.error("couldn't find `%s`", to_lookup)
             return -1
 
-        r, w = os.pipe()
-        pid = os.fork()
-        if pid < 0:
-            os.close(r)
-            os.close(w)
-            return pid
+        env = os.environ.copy()
+        try:
+            env.pop("VIRTUAL_ENV")
+        except KeyError:
+            pass
 
-        if pid == 0:
-            os.close(r)
-
-            # close qtile's stdin, stdout, stderr so the called process doesn't
-            # pollute our xsession-errors.
-            os.close(0)
-            os.close(1)
-            os.close(2)
-
-            pid2 = os.fork()
-            if pid2 == 0:
-                os.close(w)
-                try:
-                    # if qtile was installed in a virutal env, we don't
-                    # necessarily want to propagate that to children
-                    # applications, since it may change e.g. the behavior
-                    # of shells that spawn python applications
-                    del os.environ["VIRTUAL_ENV"]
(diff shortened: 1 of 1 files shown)


libqtile/core/manager.py
```diff
@@ -1141,86 +1141,26 @@ def cmd_spawn(self, cmd: str | list[str], shell: bool = False) -> int:
...
-
-            # 1024 bytes should be enough for any pid. :)
-            pid = int(os.read(r, 1024))
-            os.close(r)
-            return pid
+        env = os.environ.copy()
+        try:
+            env.pop("VIRTUAL_ENV")
+        except KeyError:
+            pass
```
m-col, line 1153:
```suggestion
        env.pop("VIRTUAL_ENV", None)
```

m-col:
This seems so obvious that I can't help but think there must be a reason why it wasn't like this before :thinking: but +1 from me. I'll pull it down and run it on my daily driver branch and see if there are any side effects but AFAICT it is functionally equivalent.

elParaguayo:
> This seems so obvious that I can't help but think there must be a reason why it wasn't like this before thinking but +1 from me. I'll pull it down and run it on my daily driver branch and see if there are any side effects but AFAICT it is functionally equivalent.

Ha. Yeah. That was exactly my thinking too!

I'm going to run it for a while and see if I find any issues.

libqtile/core/manager.py
elParaguayo: Keep forgetting you can do that.

libqtile/core/manager.py
```diff
@@ -1141,86 +1141,26 @@ def cmd_spawn(self, cmd: str | list[str], shell: bool = False) -> int:
...
+
+        proc = subprocess.Popen(
+            args,
+            stdout=subprocess.DEVNULL,
+            stdin=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            shell=shell,
+            env=env,
+        )
```
tych0, line 1162: This doesn't double fork, so qtile will remain the parent of these processes, and you'll need to wait() on them as well, or you'll leak lots of zombies. That's the reason for all the rigamaroll above, is to daemonize things so that we don't have to implement wait().

elParaguayo: Ha! There it is!

Thanks tych0. Let's kill this then.

elParaguayo: Feels like there should be a way to do this with `subprocess` though... Might be worth a bit of digging.

tych0: Yeah, it does feel like there should be some library which does this out of the box. Seems like there's still some simplification to be had here with the inner thread at least? But maybe it's not that much.

elParaguayo: It should be possible to do this with `asyncio` e.g. `asyncio.create_subprocess_exec` as we can then `wait()` asynchronously.

PR #3725, opened 2022-07-27
Fixed FFBuild Script

ChetanSagta:
This is a very small change that just fixes the ffbuild script which had the file path wrong..instead of putting '..' for parent directory, it uses '.' for current directory.


elParaguayo:
I'm curious - why do you think this is necessary? The build script works fine without this.

ChetanSagta:
It wasn't working for me, therefore I updated the paths. I checked the commit for the file and it was old, so I thought maybe the file structure was changed and someone forgot to update the paths in the script. :)

elParaguayo:
The commit is old because it's worked for so long :wink:

Can you explain more about how it "wasn't working"? How were you triggering the script? What messages did you get?

You can see that this change has already broken the building of our docs so this PR won't get merged.

ChetanSagta:
I just ran the specific files.. I wasn't aware of the builds.. maybe I'll read in the docs and update the PR.

elParaguayo:
I don't think a PR is needed at all. We're not getting any errors with the script as it is.

That's why I asked what you did to trigger the error.

PR #3732, opened 2022-07-28
CheckUpdates: add `initial_text` parameter

jx11r:
Allows the user to display an initial text while `ThreadPoolText` is completed.

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -81,6 +81,12 @@ class CheckUpdates(base.ThreadPoolText):
             (lambda x: x),
             "Lambda function to modify line count from custom_command",
         ),
+        (
+            "initial_text",
+            "",
+            "Draw the widget immediately with an initial text, "
+            "useful if it takes time to check system updates.",
+        ),
         ("update_interval", 60, "Update interval in seconds."),
         ("execute", None, "Command to execute on click"),
         ("display_format", "Updates: {updates}", "Display format if updates available"),
@@ -91,7 +97,7 @@ class CheckUpdates(base.ThreadPoolText):
     ]
 
     def __init__(self, **config):
-        base.ThreadPoolText.__init__(self, "", **config)
+        base.ThreadPoolText.__init__(self, config.pop("initial_text", ""), **config)
         self.add_defaults(CheckUpdates.defaults)
 
         # Helpful to have this as a variable as we can shorten it for testing


libqtile/widget/check_updates.py
```diff
@@ -63,8 +63,8 @@ class CheckUpdates(base.ThreadPoolText):
         ("no_update_string", "", "String to display if no updates available"),
     ]
 
-    def __init__(self, **config):
-        base.ThreadPoolText.__init__(self, "", **config)
+    def __init__(self, initial_text="", **config):
+        base.ThreadPoolText.__init__(self, initial_text, **config)
```
elParaguayo, line 67: If you do it this way, it won't be very clear from the docs that you're able to set `initial_text`.

An alternative would be to add `initial_text` to `defaults` and then do this:
```suggestion
    def __init__(self, **config):
        base.ThreadPoolText.__init__(self, config.pop("initial_text", ""), **config)
```

elParaguayo:
How many other widgets could benefit from this?

I wonder if there's a better way of doing this within the `base._TextBox` class.

jx11r:
> How many other widgets could benefit from this?

Hmm... well, there are too many widgets that use the `base.ThreadTextPool`.

> I wonder if there's a better way of doing this within the `base.ThreadPoolText` class.

Yup, maybe something like this:
```diff
diff --git a/libqtile/widget/base.py b/libqtile/widget/base.py
index c6cf834e..0ebbbef4 100644
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -737,9 +737,12 @@ class ThreadPoolText(_TextBox):
             600,
             "Update interval in seconds, if none, the " "widget updates whenever it's done.",
         ),
+        ("initial_text", "", ""),
     ]  # type: list[tuple[str, Any, str]]
 
     def __init__(self, text, **config):
+        if "initial_text" in config:
+            text = config.pop("initial_text")
         super().__init__(text, **config)
         self.add_defaults(ThreadPoolText.defaults)
```

Any suggestions for the key description?

jx11r:
Not sure if more widgets need this, most of them draw the widget immediately, so I don't know...

m-col:
> Not sure if more widgets need this, most of them draw the widget immediately, so I don't know...

Yeah I thought the same. Initially it does look like it would be sensible to just stick it in `ThreadPoolText` but actually most of the other widgets do handle initial text themselves. Therefore I'd be worried about confusing users of those in the docs.

So @elParaguayo if you're happy with that - I'm happy to merge this as-is :)

elParaguayo:
Fine by me.

m-col:
Cool. The tests seemt to have stalled or something, and I can't restart them. @jx11r please could you rebase and push? That should trigger them

jx11r:
@m-col done :)

m-col:
Awesome, thanks! The PR will auto-merge when the tests pass :)

PR #3749, opened 2022-08-07
Better support for multiple StatusNotifier widgets

### LINKED ISSUE #3738: StatusNotifier doesn't handle multiple instances correctly
### The issue:

The StatusNotifierHost assumes there's only one widget running at any time and so cannot handle multiple callbacks to different widgets.

This could cause some rendering issues (`self.bar.draw()` may not be called when new items are added) but it also results in menus being rendered in the wrong location when using the `qtile-extras` modded version.

### Required:

elParaguayo:
The StatusNotifierHost only allowed a single callback to be registered. This prevents different widgets from registering their own callbacks. This is not particularly problematic for the stock widget but does cause issues for the qtile-extras version which  needs to draw context menus as the menu could be rendered against the wrong instance.

This PR addresses the issue by appending registered callbacks to a list and calling each whenever the triggering event arises.

Fixes #3738

### DIFF
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -238,7 +238,7 @@ class StatusNotifierItem:  # noqa: E303
         """Method to invalidate icon cache and redraw icons."""
         self._invalidate_icons()
         if self.on_icon_changed is not None:
-            self.on_icon_changed()
+            self.on_icon_changed(self)
 
     def _invalidate_icons(self):
         self.surfaces = {}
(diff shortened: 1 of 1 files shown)


libqtile/widget/statusnotifier.py
```diff
@@ -484,17 +484,35 @@ def __init__(self):
         self.items: List[StatusNotifierItem] = []
         self.name = "qtile"
         self.icon_theme: str = None
+        self.started = False
+        self.on_item_added: List[Callable] = []
+        self.on_item_removed: List[Callable] = []
+        self.on_icon_changed: List[Callable] = []
```
m-col, line 491: If these are private attributes (not accessed by your code in qtile-extras) could we prefix with underscores so that's clear in the future?

m-col (approved):
LGTM and thanks :)

Issue #3337, opened 2022-02-14
Double margins in some layouts


jx11r:
**Layouts with this issue:**
>  - Columns
>  - ~Spiral~ (Fixed: #3331)
>  - Stack
>  - Bsp
>  - Matrix
>  - RatioTile
>  - Tile
>  - VerticalTile
>  - Zoomy

**Layouts without this issue:**
>  - MonadTall
>  - MonadWide
>  - MonadThreeCol

**Additional context:**
```python
config = {
    'margin': 30,
    'single_margin': 30,
}

layouts = [
    layout.MonadTall(**config),
    layout.Columns(**config),
]
```
<details>
  <summary>Screenshot</summary>
  <h3 align="center">MonadTall</h3>
  <img src="https://user-images.githubusercontent.com/78786202/163640333-405fb8e9-fb01-400b-a323-b9f6149abf80.png" />
  <h3 align="center">Columns</h3>
  <img src="https://user-images.githubusercontent.com/78786202/163640491-59d73a5a-0b10-4aa4-9c9d-4f56042d6c6e.png" />
</details>

<details>
  <summary>Screenshot (with grids)</summary>
  <h3 align="center">MonadTall</h3>
  <img src="https://user-images.githubusercontent.com/78786202/163648947-3820d1c4-4c35-454c-aea1-2751a7578b3c.png" />
  <h3 align="center">Columns</h3>
  <img src="https://user-images.githubusercontent.com/78786202/163648957-98aabc85-03f2-454f-8d35-d653e1112d6a.png" />
</details>

elParaguayo:
Thanks. I noticed this too. 

I think what's happening is that we get doubling of margins when windows are next to each other in some layouts.

Will take a look.

elParaguayo:
OK - the xmonad layout does have additional handling for this:
https://github.com/qtile/qtile/blob/c1f5073935a61f2bb9a36e198791590410ec07a1/libqtile/layout/xmonad.py#L358

Will check how this works and then see if we can add to the other layouts in a consistent way.

jx11r:
Yup, it looks exactly like the margin is doubled on the other layouts.

Thanks for checking it out.

arjan-s:
Similar but more specific issue here: https://github.com/qtile/qtile/issues/3193

elParaguayo:
@arjan-s that's a different issue. This is about doubling margins when windows are placed next to each other. The `single_margin` option is about setting a different margin when there's only one window.

[...]

PR #3750, opened 2022-08-08
Fix margin list for xmonad layouts

ylpoonlg:
Add support for margin list for xmonad layouts as previously ignored in [#2041](https://github.com/qtile/qtile/pull/2041).
The middle margin is calculated as the maximum of the left and right (or top and bottom) margin so that it is not double counted.
I'm not sure if this should be the expected behavior but I think it makes the most sense to make it the union of two adjacent margins.

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -151,7 +151,7 @@ class MonadTall(_SimpleLayoutBase):
         ("border_width", 2, "Border width."),
         ("single_border_width", None, "Border width for single window"),
         ("single_margin", None, "Margin size for single window"),
-        ("margin", 0, "Margin of the layout"),
+        ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
         (
             "ratio",
             0.5,
@@ -336,6 +336,11 @@ class MonadTall(_SimpleLayoutBase):
         width_main = int(self.screen_rect.width * self.ratio)
         width_shared = self.screen_rect.width - width_main
 
+        if isinstance(self.margin, int):
+            margin = [self.margin] * 4
+        else:
+            margin = self.margin
+
         # calculate client's x offset
         if self.align == self._left:  # left or up orientation
             if cidx == 0:
@@ -347,7 +352,7 @@ class MonadTall(_SimpleLayoutBase):
         else:  # right or down orientation
             if cidx == 0:
                 # main client
-                xpos = self.screen_rect.x + width_shared - self.margin
+                xpos = self.screen_rect.x + width_shared - 2 * self.border_width
             else:
                 # secondary client
                 xpos = self.screen_rect.x
@@ -364,8 +369,25 @@ class MonadTall(_SimpleLayoutBase):
             height = self._get_absolute_size_from_relative(self.relative_sizes[cidx - 1])
             # fix double margin
             if cidx > 1:
-                ypos -= self.margin
-                height += self.margin
+                ypos -= min(margin[0], margin[2])
+                height += min(margin[0], margin[2])
+
+            # margin list
+            if self.align == self._left:
+                margin = [
+                    margin[0],
+                    margin[1],
+                    margin[2],
+                    max(margin[1], margin[3]),
+                ]
+            else:
+                margin = [
+                    margin[0],
+                    max(margin[1], margin[3]),
+                    margin[2],
+                    margin[3],
+                ]
+
             # place client based on calculated dimensions
             client.place(
                 xpos,
@@ -374,10 +396,27 @@ class MonadTall(_SimpleLayoutBase):
                 height - 2 * self.border_width,
                 self.border_width,
                 px,
-                margin=self.margin,
+                margin=margin,
             )
         else:
             # main client
+
+            # margin list
+            if self.align == self._left:
+                margin = [
+                    margin[0],
+                    2 * self.border_width,
+                    margin[2] + 2 * self.border_width,
+                    margin[3],
+                ]
+            else:
+                margin = [
+                    margin[0],
+                    margin[1],
+                    margin[2] + 2 * self.border_width,
+                    2 * self.border_width,
+                ]
+
             client.place(
                 xpos,
                 self.screen_rect.y,
@@ -385,12 +424,7 @@ class MonadTall(_SimpleLayoutBase):
                 self.screen_rect.height,
                 self.border_width,
                 px,
-                margin=[
-                    self.margin,
-                    2 * self.border_width,
-                    self.margin + 2 * self.border_width,
-                    self.margin,
-                ],
+                margin=margin,
             )
 
     def info(self):
@@ -888,6 +922,11 @@ class MonadWide(MonadTall):
         height_main = int(self.screen_rect.height * self.ratio)
         height_shared = self.screen_rect.height - height_main
 
+        if isinstance(self.margin, int):
+            margin = [self.margin] * 4
+        else:
+            margin = self.margin
+
         # calculate client's x offset
         if self.align == self._up:  # up orientation
             if cidx == 0:
@@ -899,7 +938,7 @@ class MonadWide(MonadTall):
         else:  # right or down orientation
             if cidx == 0:
                 # main client
-                ypos = self.screen_rect.y + height_shared - self.margin
+                ypos = self.screen_rect.y + height_shared - 2 * self.border_width
             else:
                 # secondary client
                 ypos = self.screen_rect.y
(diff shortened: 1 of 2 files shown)


elParaguayo:
According to https://github.com/qtile/qtile/issues/3337, there is no double counting of margins in the xmonad layouts.

I'll look at your code properly later on but I've started the tests in the meantime.

elParaguayo:
Re double counting: if a user sets a margin of `5`, I would not expect to have a gap of `10` between windows.

ylpoonlg:
Sorry for the unclear context, but what I mean is that the xmonad layouts were ignored because it didn't have a strategy to combine the left and right margin into one. This PR will not break any configs, it will still work the same way when given an integer.
Example: if margin = [10, 20, 30, 40], the middle margin will be 40 (max of 20 and 40), and the horizontal margins on the secondary pane will be 30 (max of 10 and 30).

libqtile/layout/xmonad.py
```diff
@@ -364,8 +369,23 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
+                margin[2],
+                max(margin[1], margin[3]),
+            ]
+            margin_right = [
+                margin[0],
+                max(margin[1], margin[3]),
+                margin[2],
+                margin[3],
+            ]
+
```
elParaguayo, line 388:
```suggestion
            # margin list
            if self.align == self._left:
                margin = [
                    margin[0],
                    margin[1],
                    margin[2],
                    max(margin[1], margin[3]),
                ]
            else:
                margin = [
                    margin[0],
                    max(margin[1], margin[3]),
                    margin[2],
                    margin[3],
                ]

```

libqtile/layout/xmonad.py
```diff
@@ -374,23 +394,33 @@ def _configure_specific(self, client, screen_rect, px, cidx):
                 height - 2 * self.border_width,
                 self.border_width,
                 px,
-                margin=self.margin,
+                margin=margin_left if self.align == self._left else margin_right,
```
elParaguayo, line 397:
```suggestion
                margin=margin,
```

libqtile/layout/xmonad.py
```diff
@@ -374,23 +394,33 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
+                margin[2] + 2 * self.border_width,
+                margin[3],
+            ]
+            margin_right = [
+                margin[0],
+                margin[1],
+                margin[2] + 2 * self.border_width,
+                2 * self.border_width,
+            ]
+
```
elParaguayo, line 415:
```suggestion

            # margin list
            if self.align == self._left:
                margin = [
                    margin[0],
                    2 * self.border_width,
                    margin[2] + 2 * self.border_width,
                    margin[3],
                ]
            else:
                margin = [
                    margin[0],
                    margin[1],
                    margin[2] + 2 * self.border_width,
                    2 * self.border_width,
                ]

```

libqtile/layout/xmonad.py
```diff
@@ -374,23 +394,33 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
                 self.screen_rect.height,
                 self.border_width,
                 px,
-                margin=[
-                    self.margin,
-                    2 * self.border_width,
-                    self.margin + 2 * self.border_width,
-                    self.margin,
-                ],
+                margin=margin_left if self.align == self._left else margin_right,
```
elParaguayo, line 423:
```suggestion
                margin=margin,
```

libqtile/layout/xmonad.py
```diff
@@ -916,8 +951,23 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
+                margin[2],
+                margin[3],
+            ]
+            margin_down = [
+                margin[0],
+                margin[1],
+                max(margin[0], margin[2]),
+                margin[3],
+            ]
+
```
elParaguayo, line 970:
```suggestion
            # margin list
            if self.align == self._up:
                margin = [
                    max(margin[0], margin[2]),
                    margin[1],
                    margin[2],
                    margin[3],
                ]
            else:
                margin = [
                    margin[0],
                    margin[1],
                    max(margin[0], margin[2]),
                    margin[3],
                ]

```

libqtile/layout/xmonad.py
```diff
@@ -926,23 +976,33 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
+                2 * self.border_width,
+                margin[3],
+            ]
+            margin_down = [
+                2 * self.border_width,
+                margin[1] + 2 * self.border_width,
+                margin[2],
+                margin[3],
+            ]
+
```
elParaguayo, line 997:
```suggestion
            # margin list
            if self.align == self._up:
                margin = [
                    margin[0],
                    margin[1] + 2 * self.border_width,
                    2 * self.border_width,
                    margin[3],
                ]
            else:
                margin = [
                    2 * self.border_width,
                    margin[1] + 2 * self.border_width,
                    margin[2],
                    margin[3],
                ]

```

libqtile/layout/xmonad.py
```diff
@@ -926,23 +976,33 @@ def _configure_specific(self, client, screen_rect, px, cidx):
...
                 height_main,
                 self.border_width,
                 px,
-                margin=[
-                    self.margin,
-                    self.margin + 2 * self.border_width,
-                    2 * self.border_width,
-                    self.margin,
-                ],
+                margin=margin_up if self.align == self._up else margin_down,
```
elParaguayo, line 1005:
```suggestion
                margin=margin,
```

elParaguayo (changes requested):

elParaguayo:
I've made some suggestions which would remove some unnecessary calculations.

I haven't tried this code yet though but will run the rest suite to see if the current tests still pass with your changes.

elParaguayo:
@ylpoonlg One question for you on this: if you've changed the margin calculations here why aren't the tests (`test_wide_margins` and `test_tall_margins`) now failing? Is this because they only open 2 windows?

elParaguayo:
OK - are you able to add a test for this or do you need me to do it?

ylpoonlg:
I think I should be able to do it, thanks

elParaguayo:
Ignore the failed test. That's unrelated to your test.

Could you just add some comments to your test to explain why the dimensions/positions are equal to the numbers you're checking.

ylpoonlg:
I have added some comments for one of the cases, not sure if this is enough to make it more understandable. There are too many different cases for the main pane and client panes.

test/layouts/test_xmonad.py
```diff
@@ -255,8 +266,30 @@ def test_tall_margins(manager):
     assert_focused(manager, "one")
     assert_dimensions(manager, 4, 4, 392, 588)
 
+    manager.test_window("three")
+    assert_focused(manager, "three")
+    assert_dimensions(manager, 404, 4, 388, 288)
+
+    manager.c.layout.next()
+    assert_focused(manager, "two")
+    assert_dimensions(manager, 404, 300, 388, 292)
```
elParaguayo, line 275: `two` and `three` have different heights. That doesn't seem right. Is that the issue you were trying to fix?

test/layouts/test_xmonad.py
```diff
@@ -255,8 +266,30 @@ def test_tall_margins(manager):
...
+    # y = screen_height / 2
+    # width = screen_width * 0.5 - margin[1] - margin[3] - 2 * border
+    # height = screen_height / 2 - margin[0] - margin[2] - 2 * border
+    #            + min(margin[0], margin[2]) # Combine top margin
+    assert_focused(manager, "two")
+    assert_dimensions(manager, 408, 300, 384, 290)
+
+    manager.c.layout.previous()
+    assert_focused(manager, "three")
+    assert_dimensions(manager, 408, 2, 384, 288)
```
elParaguayo, line 289: `two` and `three` are still different heights (but are closer now). Is this a rounding issue?

test/layouts/test_xmonad.py
```diff
@@ -255,8 +266,30 @@ def test_tall_margins(manager):
...
+    assert_dimensions(manager, 404, 4, 388, 288)
+
+    manager.c.layout.next()
+    assert_focused(manager, "two")
+    assert_dimensions(manager, 404, 300, 388, 292)
+
+    manager.c.next_layout()
+
+    # x = screen_width * 0.5 + margin[3]
+    # y = screen_height / 2
```
elParaguayo, line 280: Shouldn't there be a an adjustment for the margin here too? Or is the whole margin applied to the window above it?

(Thinking out loud here!) I guess you can't distribute the layout evenly between windows.

test/layouts/test_xmonad.py
```diff
@@ -269,6 +302,29 @@ def test_wide_margins(manager):
...
+    assert_dimensions(manager, 4, 304, 388, 288)
+
+    manager.c.layout.next()
+    assert_focused(manager, "two")
+    assert_dimensions(manager, 400, 304, 392, 288)
+
+    manager.c.next_layout()
+
+    # x = screen_width / 2 + margin[3] - min(margin[1], margin[3])
+    # y = screen_height * 0.5 + max(margin[0], margin[2])
```
elParaguayo, line 316: Isn't this too big? Top window finishes at y=296, you're now at y=306.

elParaguayo (commented):
Thanks for updating the test.

I'm still a bit confused by the logic so have added some questions for you.

ylpoonlg:
Sorry, I know it is quite confusing. I actually didn't change how it handles combining the margins. I have just changed the values of the margins. If you take a look at the original source code of the layout, you can see how the dimensions were calculated.

I've put together a simple diagram for a test case:

qtile_xmonad_layout

PR #3757, opened 2022-08-10
Volume widget display change

Ahmed-Jarir:
this is a minor change, where u'd see both the volume and if it's muted or not as (volume [on] or [off]) instead of the current behavior. [on] or [off] can also be changed by supplying values to unmutetext and mutetext in widgets. any changes to variable names or the default on off would make it better I think, but I cannot come up with other names.

### DIFF
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -50,6 +50,7 @@ class PulseVolume(Volume):
         self.default_sink = None
         self.handle = ffi.new_handle(self)
         self.client_name = ffi.new("char[]", b"Qtile-pulse")
+        self._old_mute = None
 
         self.connect()
 
@@ -221,23 +222,24 @@ class PulseVolume(Volume):
         manually re-schedule update
         """
         vol = self.get_volume()
-        if vol != self.volume:
+        if vol != self.volume or self.mute is not self._old_mute:
             self.volume = vol
+            self._old_mute = self.mute
             # Update the underlying canvas size before actually attempting
             # to figure out how big it is and draw it.
             self._update_drawer()
             self.bar.draw()
 
     def get_volume(self):
-        self.mute = self.mute_text
+        self.mute = self.unmute_text
         if self.default_sink:
             if self.default_sink["muted"]:
-                return -1
+                self.mute = self.mute_text
+                return 0
             base = self.default_sink["base_volume"]
             if not base:
                 return -1
             current = max(self.default_sink["values"])
-            self.mute = self.unmute_text
             return round(current * 100 / base)
         return -1
 
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -229,6 +229,7 @@ class PulseVolume(Volume):
             self.bar.draw()
 
     def get_volume(self):
+        self.mute = self.mute_text
         if self.default_sink:
             if self.default_sink["muted"]:
                 return -1
@@ -236,6 +237,7 @@ class PulseVolume(Volume):
             if not base:
                 return -1
             current = max(self.default_sink["values"])
+            self.mute = self.unmute_text
             return round(current * 100 / base)
         return -1
 
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -231,10 +231,10 @@ class PulseVolume(Volume):
             self.bar.draw()
 
     def get_volume(self):
-        self.mute = self.unmute_text
+        self.mute = False
         if self.default_sink:
             if self.default_sink["muted"]:
-                self.mute = self.mute_text
+                self.mute = True
                 return 0
             base = self.default_sink["base_volume"]
             if not base:
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -235,7 +235,6 @@ class PulseVolume(Volume):
         if self.default_sink:
             if self.default_sink["muted"]:
                 self.mute = True
-                return 0
             base = self.default_sink["base_volume"]
             if not base:
                 return -1
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -85,7 +85,7 @@ class Volume(base._TextBox):
         ),
         (
             "mute_format",
-            "{volume}% M",
+            "M",
             "Format of text to display when the volume is muted. Available fields: 'volume'",
         ),
     ]
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -32,11 +32,9 @@
 
 import re
 import subprocess
-from copy import deepcopy
 
 from libqtile import bar
 from libqtile.widget import base
-from libqtile.log_utils import logger
 
 __all__ = [
     "Volume",
@@ -92,10 +90,10 @@ class Volume(base._TextBox):
         self.mutedtext = "[off]"
         self.unmutedtext = "[on]"
 
-        if "muteicon" in config:
+        if "mutetext" in config:
             self.mutedtext = config["mutetext"]
 
-        if "unmuteicon" in config:
+        if "unmutetext" in config:
             self.unmutedtext = config["unmutetext"]
 
 
@@ -137,11 +135,11 @@ class Volume(base._TextBox):
 
     def update(self):
         vol = self.get_volume()
-        prevMute = deepcopy(self.mute)
-    
-        self.check_mute()
-        if vol != self.volume or self.mute != prevMute:
+        nMute = self.get_mute()
+
+        if vol != self.volume or self.mute != nMute:
             self.volume = vol
+            self.mute = nMute
             # Update the underlying canvas size before actually attempting
             # to figure out how big it is and draw it.
             self._update_drawer()
@@ -209,8 +207,8 @@ class Volume(base._TextBox):
             # this shouldn't happen
             return -1
 
-    def check_mute(self):
-        self.mute = self.mutedtext if "[off]" in self.mixer_out else self.unmutedtext
+    def get_mute(self):
+        return self.mutedtext if "[off]" in self.mixer_out else self.unmutedtext
 
     def draw(self):
         if self.theme_path:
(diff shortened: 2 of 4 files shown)


SweepingBishops (commented):
For variable names: `variable_names_follow_this_format`.
So `mutetext` would be `mute_text`,
`nMute` would be `next_mute` (or `not_mute`?)

libqtile/widget/volume.py
```diff
@@ -85,6 +85,17 @@ def __init__(self, **config):
         self.add_defaults(Volume.defaults)
         self.surfaces = {}
         self.volume = None
+        self.mute = ""
+        self.mixer_out = ""
+        self.mutedtext = "[off]"
+        self.unmutedtext = "[on]"
+
+        if "mutetext" in config:
```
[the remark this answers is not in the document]
Ahmed-Jarir: Yes, I should have done that my bad, I'll change it in the next commit.

libqtile/widget/volume.py
```diff
@@ -178,27 +187,29 @@ def setup_images(self):
...
 
-        volgroups = re_vol.search(mixer_out)
         if volgroups:
             return int(volgroups.groups()[0])
         else:
             # this shouldn't happen
             return -1
 
+    def get_mute(self):
+        return self.mutedtext if "[off]" in self.mixer_out else self.unmutedtext
```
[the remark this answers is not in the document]
Ahmed-Jarir: actually this does what you're saying it should, what it's doing is basically 
`if "[off]" in self.mixer_out:
    return self.mutedtext
 else:
     return self.unmutedtext`
and self.mixer_out is set in get_volume(), where self.mixer_out is the output of the volume command. also I'm using this config currently and it works with media keys and the command.

elParaguayo:
Thanks for this.

I'll post some review comments on this later as I'm a bit tied up today.

libqtile/widget/volume.py
```diff
@@ -178,27 +180,29 @@ def setup_images(self):
...
                 get_volume_cmd = self.get_volume_command
 
-            mixer_out = self.call_process(get_volume_cmd)
+            self.mixer_out = self.call_process(get_volume_cmd)
         except subprocess.CalledProcessError:
             return -1
 
-        if "[off]" in mixer_out:
-            return -1
+        volgroups = re.compile(r"\[(\d?\d?\d?)%\]").search(self.mixer_out)
```
elParaguayo, line 195: Not sure why you changed this. You don't want to `re.compile` this every time `get_volume` is called. This has already been compiled earlier in the code.

```suggestion
        volgroups = re_vol.search(self.mixer_out)
```

elParaguayo (changes requested):

elParaguayo:
Added some comments.

Can you also add a test to `test_volume.py` to check the mute display works as intended.

Ahmed-Jarir:
I changed the condition because the behavior of `get_volume()` was changed, but I kept it as `<=0` just in case the subprocess error occurs

libqtile/widget/volume.py
Ahmed-Jarir: yes, I forgot to change that sorry. I used it when I was experimenting with overriding the function, but forgot to change it after copying it.

libqtile/widget/volume.py
```diff
@@ -156,10 +164,7 @@ def _update_drawer(self):
             elif self.volume >= 80:
                 self.text = "\U0001f50a"
         else:
-            if self.volume == -1:
-                self.text = "M"
-            else:
-                self.text = "{}%".format(self.volume)
+            self.text = self.format.format(volume=self.volume, mute=self.mute)
```
elParaguayo, line 167: Oh, hang on... the Pulse volume widget inherits the `Volume` widget so I wonder if this breaks it. Let's see what happens in the tests!

test/widgets/test_volume.py
```diff
@@ -51,14 +51,21 @@ def test_emoji():
...
+
 
 def test_text():
     fmt = "Volume: {}"
     vol = Volume(fmt=fmt)
-    vol.volume = -1
+    vol.volume = 80
     vol._update_drawer()
-    assert vol.text == "M"
+    assert vol.text == "80%"
```
elParaguayo, line 64: I'd leave this as `-1` and correct the output for that scenario. `-1` now represents an error so maybe we should tell the user.

test/widgets/test_volume.py
```diff
@@ -51,14 +51,21 @@ def test_emoji():
...
 
+    format = "Volume: {volume}% {mute}"
```
elParaguayo, line 66: Have this test as a separate one with a new function name.

test/widgets/test_volume.py
elParaguayo: Actually, as per comment above, `-1` also means muted for the Pulse volume widget so we should continue to treat that way.

elParaguayo:
I think this would make the Pulse widget compatible:
``` diff
diff --git a/libqtile/widget/pulse_volume.py b/libqtile/widget/pulse_volume.py
index 592ac408..992d8a35 100644
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -229,6 +229,7 @@ class PulseVolume(Volume):
             self.bar.draw()

     def get_volume(self):
+        self.mute = self.mute_text
         if self.default_sink:
             if self.default_sink["muted"]:
                 return -1
@@ -236,6 +237,7 @@ class PulseVolume(Volume):
             if not base:
                 return -1
             current = max(self.default_sink["values"])
+            self.mute = self.unmute_text
             return round(current * 100 / base)
         return -1
```

Ahmed-Jarir:
@elParaguayo I don't have pulse audio to check, but isn't it supposed to be the same behavior as the normal volume widget? if so am I not supposed to alter the update function of it to match the amixer volume widget?

elParaguayo:
The issue is that we changed the formatting in the `Volume` widget which means we need both `self.volume` and `self.mute` attributes. The pulse widget wasn't setting the mute status so I *think* my diff fixes that!

Ahmed-Jarir:
hmm ok then, I made that change in the last commit can we test it now?

elParaguayo:
There's no test for the Pulse widget. I'll need to pull this PR onto my machine and test locally.

However, we can still run the CI test suite here.

Ahmed-Jarir:
@elParaguayo  can you please inform me when you test it?

elParaguayo:
Small change. This makes the volume `0` when muted and also enables the right mute text.

``` diff
diff --git a/libqtile/widget/pulse_volume.py b/libqtile/widget/pulse_volume.py
index 992d8a35..8ada5edb 100644
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -50,6 +50,7 @@ class PulseVolume(Volume):
         self.default_sink = None
         self.handle = ffi.new_handle(self)
         self.client_name = ffi.new("char[]", b"Qtile-pulse")
+        self._old_mute = None

         self.connect()

@@ -221,23 +222,24 @@ class PulseVolume(Volume):
         manually re-schedule update
         """
         vol = self.get_volume()
-        if vol != self.volume:
+        if vol != self.volume or self.mute is not self._old_mute:
             self.volume = vol
+            self._old_mute = self.mute
             # Update the underlying canvas size before actually attempting
             # to figure out how big it is and draw it.
             self._update_drawer()
             self.bar.draw()

     def get_volume(self):
-        self.mute = self.mute_text
+        self.mute = self.unmute_text
         if self.default_sink:
             if self.default_sink["muted"]:
-                return -1
+                self.mute = self.mute_text
+                return 0
             base = self.default_sink["base_volume"]
             if not base:
                 return -1
             current = max(self.default_sink["values"])
-            self.mute = self.unmute_text
             return round(current * 100 / base)
         return -1
```

elParaguayo:
I'll be honest - I'm still not 100% on this one. I wonder if there's a slightly better approach.

We've got a `format` string but we could also have a `muted_format` string too which is used when the audio is muted. I think this would be slightly neater than the `format`, `mute_text` and `unmute_text` that we have at the moment.

Ahmed-Jarir:
I'm not quite sure that I understand this, do you mean it should be something like "50% M" when muted and "50%" when its not?

elParaguayo:
If you just have the current `format` string and set it to `{volume}% {mute}` then you get `60% [on]` and `60% [off]`. The problem is that you always have the volume. There's no way to hide it.

However, if you had `format="{volume}%"` you could also have `muted_format="{volume}% muted"` or something just like `muted_format="M"` which replicates the current behaviour.

Ahmed-Jarir:
I changed format to unmute_format, because it doesn't work as format for some reason, is that ok?

libqtile/widget/pulse_volume.py
```diff
@@ -221,17 +222,20 @@ def update(self):
...
             self._update_drawer()
             self.bar.draw()
 
     def get_volume(self):
+        self.mute = False
         if self.default_sink:
             if self.default_sink["muted"]:
-                return -1
+                self.mute = True
+                return 0
```
elParaguayo, line 238: We can remove the `return 0` line. It should be fine to return the actual volume level as we've set the `mute` flag.

libqtile/widget/volume.py
```diff
@@ -78,13 +78,25 @@ class Volume(base._TextBox):
...
             "Only used if ``volume_up_command`` and ``volume_down_command`` are not set.",
         ),
+        (
+            "unmute_format",
+            "{volume}%",
+            "Format of text to display. Available fields: 'volume'",
+        ),
+        (
+            "mute_format",
+            "{volume}% M",
```
elParaguayo, line 88: Could this just be "M" for now? That would replicate the current behaviour.

elParaguayo (changes requested):

Ahmed-Jarir:
@elParaguayo any updates?

m-col:
From what I can tell this is good to go and the review comments are all resolved. Could you add an entry to CHANGELOG so that other users can see your new options :) might be worth mentioning that these options apply to both `Volume` and `PulseVolume` (if I'm not mistaken?).

Also just to check I've understood correctly - this doesn't introduce any changes to the behaviour for people's existing setups?

Ahmed-Jarir:
Yes it applies for both `Volume` and `PulseVolume`.
And yes, there will not be any changes for people's existing setups, since default settings reproduce the current behavior.

PR #3758, opened 2022-08-11
fix: log python errors in config.py also on startup

oddlama:
Previously, any syntax errors in `config.py` were silently ignored on startup due to the exception of the module import being caught and converted into a simple notification message. This makes it hard to debug errors in the config file. On reload, this already is logged as expected, just not on the initial start.

This PR adds the exception log on startup to make the behavior consistent in both cases.


m-col:
Good stuff, thanks

PR #3772, opened 2022-08-17
docs: Add note on xwayland focus issue

3nprob:
Until resolved, it seems prudent to note this.


m-col:
Sorry fot the delay, thanks for noting this! Hopefully we can squash this bug soon.

PR #3773, opened 2022-08-17
Drop hard xcffib requirement, document pip install options

m-col:
This drops xcffib as a requirement for all users and prevents it from
being installed when performing `pip install qtile`. Instead, x11 or
wayland dependencies can be installed alongside qtile using install
options as defined in setup.cfg, currently already the case for wayland
(though undocumented). Specifically, these commands can be used:

```
pip install qtile[x11]
pip install qtile[wayland]
```

Not specifying an option will only install actual hard dependencies i.e.
cffi and cairocffi.

This is documented front and centre at the top of the installation page
as a quick summary which I suspect will be sufficient for many users who
probably don't want to read the whole page!


docs/manual/install/index.rst
```diff
@@ -2,6 +2,15 @@
 Installing Qtile
 ================
 
+TL;DR:
+
+.. code-block:: sh
```
SweepingBishops, line 7:
```suggestion
.. code-block:: sh

```

SweepingBishops: Um... The TL;DR is empty. I think you need a space here.

m-col: Ah oops, thanks!

elParaguayo:
This is a great idea.

For X11, does `xcffib` still get installed before `cairocffi`?

m-col:
If they aren't already installed, they will be installed in order. This is ensured by the (already present) `install_requires` dependency in setup.cfg for `cairocffi[xcb]`, which tells cairocffi to install with xcffib support.

If cairocffi is already installed but isn't built with xcffib support, then installation will print out a warning message and advice from setup.py. This already is the case and when it happens installation is blocked. This PR does change this so that installation continues (warning message still prints). The reasoning for that change is because that advice only needs to be followed if xcffib is needed i.e. the user wants to use the x11 backend. The warning message is updated to clarify this so that users who are only after using wayland can ignore it.

Reading this again I actually should update `install_requires` so that `cairocffi[xcb]` is just `cairocffi`, then add `cairocffi[xcb]` to the x11-only requirements

m-col:
Ah, more changes are required:
```
(.venv)  ~/git/qtile ▋▌▍ ./scripts/wephyr
  File "/home/mcol/git/qtile/bin/qtile", line 33, in <module>
    from libqtile.scripts.main import main  # noqa: E402
  File "/home/mcol/git/qtile/libqtile/scripts/main.py", line 6, in <module>
    from libqtile.scripts import check, cmd_obj, migrate, run_cmd, shell, start, top
  File "/home/mcol/git/qtile/libqtile/scripts/check.py", line 27, in <module>
    from libqtile import confreader
  File "/home/mcol/git/qtile/libqtile/confreader.py", line 31, in <module>
    from libqtile.backend.x11 import core
  File "/home/mcol/git/qtile/libqtile/backend/x11/core.py", line 31, in <module>
    import xcffib
ModuleNotFoundError: No module named 'xcffib'
```
There are a couple of dependency-related changes that are needed really, similarly https://github.com/qtile/qtile/pull/3602

libqtile/confreader.py
```diff
@@ -140,8 +140,17 @@ def validate(self) -> None:
...
+        valid_mods = [
+            "shift",
+            "lock",
+            "control",
+            "mod1",
+            "mod2",
+            "mod3",
+            "mod4",
+            "mod5",
+        ]
```
ramnes, line 153: I don't think we want to hard code this!

m-col: Yeah, I agree. One thing to consider is that we validate the config in `qtile check`, so in theory we shouldn't need a running core to check. This works fine with the X backend (presumably why the functions were standalone functions in the core.py module). A possible approach might be to set `get_keys` and `get_modifiers` as abstract static methods on the base `Core` as part of the 'backend protocol'. The wayland backend should be able to ask xkbcommon for the data.

ramnes: > A possible approach might be to set `get_keys` and `get_modifiers` as abstract static methods on the base `Core` as part of the 'backend protocol'. The wayland backend should be able to ask xkbcommon for the data.

:+1:

m-col: https://github.com/qtile/qtile/pull/4180 does this

ramnes:
@m-col if you want to keep it open you need to add a `kind: ...` label

PR #3778, opened 2022-08-20
WIP: Add a xwayland_ready hook.

CharString:
When run with the wayland backend, this will fire when `DISPLAY` becomes
available in the os.environ, such that X clients can be launched in the
config.py

I'm running qtile on a system with no X.org, by using lightdm with wayland greeter.
Starting my ssh-agent when `startup_once` is fired wasn't enough to support [`ssh-add -c`](https://spencerkrum.com/posts/ssh-askpass-wayland/) as `DISPLAY` wasn't set yet in the env where the agent is spun up. I tried a little async co-routine that would sleep until `DISPLAY` would become available, but haven't had any luck.With this new hook I can start it like this:

```python
@hook.subscribe.xwayland_ready
def ssh_agent():
    # start agent
    agent = subprocess.Popen(
        "ssh-agent",
        stdout=subprocess.PIPE,
        text=True,
    )
    # cleanup on shutdown
    hook.subscribe.shutdown(agent.kill)

    # set the env
    output, err = agent.communicate()
    for var, val in (ln.split("=") for ln in output.splitlines() if "=" in ln):
        os.environ[var] = val.split(";")[0].strip()
```

TODO:

- [ ] discussion
- [ ] test
- [ ] documentation

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -581,6 +581,7 @@ class Core(base.Core, wlrq.HasListeners):
                     image._ptr.hotspot_x,
                     image._ptr.hotspot_y,
                 )
+        hook.fire("xwayland_ready")
 
     def _on_xwayland_new_surface(self, _listener: Listener, surface: xwayland.Surface) -> None:
         logger.debug("Signal: xwayland new_surface")
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -355,6 +355,18 @@ class Subscribe:
         """
         return self._subscribe("resume", func)
 
+    def xwayland_ready(self, func):
+        """
+        Called when xwayland is started and ready and "DISPLAY" is set in os.environ.
+
+        Note: only the wayland backend fires this.
+
+        **Arguments**
+
+        None
+        """
+        return self._subscribe("xwayland_ready", func)
+
 
 subscribe = Subscribe()
 


m-col:
While the `_on_xwayland_ready` signal is emitted when XWayland is ready to host clients, if it is available on the system when Qtile starts up, it starts in a somewhat dormant state where the `DISPLAY` is known. Indeed, `os.environ["DISPLAY"]` is set during `Core.__init__`, not in `_on_xwayland_ready`, so you should be able to start X11 clients from a regular startup script without this hook. I'm not really sure why you were having issues though. Hmm.

Can you confirm that when Qtile starts up you get this log message?:
```
2022-08-20 16:40:47,690 libqtile core.py:__init__():L234  Set up XWayland with DISPLAY=:0
```

CharString:
I was starting up things in the `startup_once` hook. I tracked it down and IIRC when that gets fired, the wayland backend hasn't setup xwayland yet.

-------- Original Message --------

m-col:
Does ssh-agent output anything of note? I can add `xterm` to the `startup_once` hook and it reliably opens up then.

CharString:
Weird I can't reproduce the issue anymore. Though I'm having problems with my greeter too (lightdm with elephant-greeter). But I don't suspect starting from a tty login would change this behaviour... Closing the PR.

On a side note `hook.subscribe.shutdown(agent.kill)` doesn't kill the agent, `hook.subscribe.shutdown(lamda: subprocess.run(['ssh-agent', '-k'])` does.

elParaguayo:
> Weird I can't reproduce the issue anymore. Though I'm having problems with my greeter too (lightdm with elephant-greeter). But I don't suspect starting from a tty login would change this behaviour... Closing the PR.
> 
> On a side note `hook.subscribe.shutdown(agent.kill)` doesn't kill the agent, `hook.subscribe.shutdown(lamda: subprocess.run(['ssh-agent', '-k'])` does.

`agent` is not in the scope of your hooked function.

CharString:
> `agent` is not in the scope of your hooked function.

`agent.kill` is the function I'm hooking within  `ssh_agent`. I'm missing the score issue. What function do you mean with "hooked function"? `agent.kill` or `ssh_agent`?

elParaguayo:
Sorry. Misread your code. Yeah, that is a bit odd.

Do you get anything in the log to suggest what the error is?

elParaguayo:
Does `ssh-agent` fork when it's started? If so, `agent` would not contain the running process.

CharString:
> Does `ssh-agent` fork when it's started? If so, `agent` would not contain the running process.

That's it! From the ssh-agent man page:

    -D      Foreground mode.  When this option is specified ssh-agent will not fork.

PR #3785, opened 2022-08-21
Improve CheckUpdates docs

elParaguayo:
This PR includes the available configuration options in the docs so users can see which option to use without looking at the source code.

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -24,10 +24,35 @@ from subprocess import CalledProcessError, Popen
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
+# format: "Distro": ("cmd", "number of lines to subtract from output")
+CMD_DICT = {
+    "Arch": ("pacman -Qu", 0),
+    "Arch_checkupdates": ("checkupdates", 0),
+    "Arch_Sup": ("pacman -Sup", 0),
+    "Arch_paru": ("paru -Qu", 0),
+    "Arch_paru_Sup": ("paru -Sup", 0),
+    "Arch_yay": ("yay -Qu", 0),
+    "Debian": ("apt-show-versions -u -b", 0),
+    "Gentoo_eix": ("EIX_LIMIT=0 eix -u# --world", 0),
+    "Ubuntu": ("aptitude search ~U", 0),
+    "Fedora": ("dnf list updates -q", 1),
+    "FreeBSD": ("pkg_version -I -l '<'", 0),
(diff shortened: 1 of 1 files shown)


m-col:
Perhaps we should move the dictionary to out of the class (or to a class attribute) and use its data to complete the docstring. This way, future changes to the dictionary will definitely propagate to the docs.

elParaguayo:
> Perhaps we should move the dictionary to out of the class (or to a class attribute) and use its data to complete the docstring. This way, future changes to the dictionary will definitely propagate to the docs.

Yes. That would be nice.

See also #3789

m-col:
I wasn't sure if this was ready for review but LGTM :)

Neo-29:
I might be off topic but... Is there a way to update the information ( number of updates) after having updated?
The only way to.. I found is re spawn ( mod shift R) to update the widget.
If any tips... I'm all ears 😊. Tks.

elParaguayo:
How are you triggering the update? If you do it by setting the `execute` parameter and then clicking on the widget, the widget will update after the update process is complete.

If not, you could run `qtile cmd-obj -o widget checkupdates -f force_update`.

Neo-29:
> How are you triggering the update? If you do it by setting the `execute` parameter and then clicking on the widget, the widget will update after the update process is complete.
> 
> If not, you could run `qtile cmd-obj -o widget checkupdates -f force_update`.
Tks for the answer, 
Where should I put this command? ( sorry I'm quite new on qtile)

elParaguayo:
You can just run it in a terminal.

Neo-29:
> You can just run it in a terminal.

So can be used in `execute` command?

elParaguayo:
No. The idea of the execute command is to put in the command that actually runs the system update for you. After that has run, the widget would update the text.

Neo-29:
> No. The idea of the execute command is to put in the command that actually runs the system update for you. After that has run, the widget would update the text.

So.... How can I use this command?
Spawn?

Neo-29:
> No. The idea of the execute command is to put in the command that actually runs the system update for you. After that has run, the widget would update the text.

So.... How can I use this command?
Spawn?

elParaguayo:
You've got plenty of options:
- If you've opened a terminal to update your system, type that command into the terminal when you're finished.
- Bind a key to `lazy.widget["checkupdates"].force_update()`
- Use the `execute` command as it's intended and use the widget to run the system update command. That way you don't need a separate command to update the widget, it's done automatically.

PR #3788, opened 2022-08-22
Tidying up Mirrors

### LINKED ISSUE #3739: Mirror widgets do not stretch
### The issue:

I want do share the bar on all screens of the system. Because Its a laptop, the number of screens is variable.

Qtile Version: 0.21.0
Qtile Log: does not show a new error.

I began with the following attempt:

``` python

common_bar = [widget.Sep(), widget.Image(), widget.GroupBox(),  ..., widget.Promt(), widget.Spacer(), widget.WindowName(), ..., widget.CurrentLayoutIcon(), widget.Systray(),...]

screens = [
    Screen(top=bar.Bar(common_bar), 26, background=background),
    Screen(top=bar.Bar(common_bar), 26, background=background),
]

```
The bar is shown, but:
- not correct (stuff is missing)
- systray crashed
- some widget are on Screen 1 and two half widgets are on display 0.

### Required:

elParaguayo:
Some of my earlier code seemed a bit over-engineered and there were a few annoying bugs:

1) Mirror's contents only updated when the reflection called `draw`.
2) Stretch widgets would become fixed length.

This PR removes some redundant code and simplifies the reflection by storing a copy of the last drawn surface in a separate `RecordingSurface` object. Mirrors can paint that surface at any time.

Fixes #3739

@m-col - would appreciate your thoughts on this too (only when you have a moment, no rush!)

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -637,10 +637,11 @@ class Drawer:
         self._height = height
 
         self.surface: cairocffi.RecordingSurface
+        self.last_surface: cairocffi.RecordingSurface
         self.ctx: cairocffi.Context
         self._reset_surface()
 
-        self.mirrors: dict[Drawer, bool] = {}
+        self.mirrors: set[Drawer] = set()
 
         self.current_rect = (0, 0, 0, 0)
         self.previous_rect = (-1, -1, -1, -1)
@@ -653,16 +654,7 @@ class Drawer:
 
     def add_mirror(self, mirror: Drawer):
         """Keep details of other drawers that are mirroring this one."""
-        self.mirrors[mirror] = False
-
-    def reset_mirrors(self):
-        """Reset the drawn status of mirrors."""
-        self.mirrors = {m: False for m in self.mirrors}
-
-    @property
-    def mirrors_drawn(self) -> bool:
-        """Returns True if all mirrors have been drawn with the current surface."""
-        return all(v for v in self.mirrors.values())
+        self.mirrors.add(mirror)
 
     @property
     def width(self) -> int:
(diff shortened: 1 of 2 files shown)


m-col:
I think it all looks good! :)

m-col:
I got one of these when I added a second monitor, enabling my second (previously non-configured) `Screen`:
```
2022-09-15 19:40:48,995 ERROR libqtile loop.py:_handle_exception():L63 'Drawer' object has no attribute 'last_surface'
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
  File "/home/mcol/git/qtile/libqtile/core/manager.py", line 851, in f
    func(*args)
  File "/home/mcol/git/qtile/libqtile/bar.py", line 612, in _actual_draw
    i.draw()
  File "/home/mcol/git/qtile/libqtile/widget/base.py", line 883, in draw
    self.reflects.drawer.paint_to(self.drawer)
  File "/home/mcol/git/qtile/libqtile/backend/base.py", line 699, in paint_to
    drawer.ctx.set_source_surface(self.last_surface)
AttributeError: 'Drawer' object has no attribute 'last_surface'
```
As a guess, a widget that previously had no mirrors was surprised when it suddenly had a mirror, on the bar that just got set up on the second screen. The user experience is unaffected - it looks like all widgets (including mirrors) started working as expected sometime later without any further input (while I typed this).

m-col:
Actually this looks like it prevents mirror widgets from updating themselves or something, perhaps until done so manually (e.g. when clicked). It isn't clear to me exactly the mechanism of the symptom, but the cause does appear to be having mirrors in a `Bar` that doesn't get used until later i.e. when an output is connected.

elParaguayo:
Thanks for the info. I'll try to take a look when I can.

Just so I understand the issue: are you seeing this when you start qtile and then plug in a second monitor after that?

m-col:
Yep exactly

PR #3793, opened 2022-08-25
Add ScreenSplit layout

elParaguayo:
Adds a new layout which allows users to split screen into sub areas and assign a layout to each area, while retaining the ability for fullscreen windows to occupy the whole screen.

This is draft for now as I'm interested in people's thoughts.

This could easily go in qtile-extras if it's not wanted here.

Also, I've not looked at tests yet...

Screenshot (`Max` on top, `Columns` on bottom):


elParaguayo:
@nazar-pc - copying your question from the discussion here:

> Looks interesting and... quite simple? Would it be possible to allow setting a list of layouts for each split or it is expected that multiple ScreenSplits can be defined with a unique combination of layouts in each?

At the moment, each split takes a single layout object. It should be possible to allow a list of layouts if that's useful (but I'd need to block nested `ScreenSplit` layouts).

nazar-pc:
Another thing that could be useful is to be able to resize after instantiation like in Column and other layouts.

But even without those changes should be useful and usable, I'll try to test it over the weekend.

elParaguayo:
> Another thing that could be useful is to be able to resize after instantiation like in Column and other layouts.

Unlikely to happen. The logic would get pretty complicated if you had more than 2 splits. The idea is really to break the screen down into smaller, fixed-size screen areas.

m-col:
Another difference this has compared to `fake_screens` is that with fake screens the regions also correspond to groups, allowing you to change groups within one region without changing others. Would this be desirable behaviour?

re fullscreening, we could add a command to fullscreen a window across the entire output layout rather than a single screen (could also useful even when using multiple physical outputs).

And for bars, it appears you can give bars a large negative margin in one direction and they will extend beyond the screen it 'belongs' to. E.g. if you wanted to vertically split a 3840 pixel wide window into two 1920 wide virtual screens, giving the left 'screen' a bar with margin `[0, -1920, 0, 0` actually makes it become 1920 pixels wider, running along the top of the full 3840 pixels :laughing:

nazar-pc:
I don't think lack of groups is an issue for me personally, the way it is done with static config for one group is perfectly fine for what I have in mind, but moving windows across nested layouts is generally desirable (though may be non-critical for the very initial implementation.

elParaguayo:
> Another difference this has compared to `fake_screens` is that with fake screens the regions also correspond to groups, allowing you to change groups within one region without changing others. Would this be desirable behaviour?

I had assumed "no" to this question as, otherwise, people should use `fake_screens`

> re fullscreening, we could add a command to fullscreen a window across the entire output layout rather than a single screen (could also useful even when using multiple physical outputs).

A combination of `fake_screens` and this pretty much negates the need for this layout!

elParaguayo:
> I don't think lack of groups is an issue for me personally, the way it is done with static config for one group is perfectly fine for what I have in mind, but moving windows across nested layouts is generally desirable (though may be non-critical for the very initial implementation.

I haven't implemented moving windows between nested layouts yet.

elParaguayo:
@nazar-pc I think this is in fairly good shape now. Would be great to get some feedback.

nazar-pc:
Will do once I am a bit less busy, thank you!

nazar-pc:
Initial feedback:
* I really need borders to understand what split is active
* I wasn't able to move windows across multiple splits, it seems to be stuck and only moves between some two splits instead
* On every layout I try, I miss `master_match` from `Tile` layout to predefine positions for certain apps

I see why resizing is extremely challenging to implement. Maybe one day we'll just have ability to nest any layout in any layout and this will be solved :roll_eyes:

elParaguayo:
> Initial feedback:
> 
> * I really need borders to understand what split is active

Pretty hard to do. That's why I added the `ScreenSplit` widget (but may not have mentioned in this thread...). It tells you the name of the active split.

> * I wasn't able to move windows across multiple splits, it seems to be stuck and only moves between some two splits instead

Interesting. Let me check this.

> * On every layout I try, I miss `master_match` from `Tile` layout to predefine positions for certain apps

Are you saying you'd like a match rule for each split to put windows in specific splits straightaway? I can look at that.

> 
> I see why resizing is extremely challenging to implement. Maybe one day we'll just have ability to nest any layout in any layout and this will be solved roll_eyes

It's good to have dreams :wink:

nazar-pc:
> Pretty hard to do. That's why I added the ScreenSplit widget (but may not have mentioned in this thread...). It tells you the name of the active split.

I'm not using any widgets though, I just have tint2 that is hidden 95.5% of the time and no other panels anywhere.

> Are you saying you'd like a match rule for each split to put windows in specific splits straightaway?

Exactly. Or else it takes quite a bit of effort to move things around manually every time.

elParaguayo:
> > Pretty hard to do. That's why I added the ScreenSplit widget (but may not have mentioned in this thread...). It tells you the name of the active split.
> 
> I'm not using any widgets though, I just have tint2 that is hidden 95.5% of the time and no other panels anywhere.

Another idea would be to write a function that's hooked to `layout_change` and displays a notification (if you've got notifications enabled). The hook is fired every time the active split changes.

> 
> > Are you saying you'd like a match rule for each split to put windows in specific splits straightaway?
> 
> Exactly. Or else it takes quite a bit of effort to move things around manually every time.

Ok. Can add this.

elParaguayo:
Added matches - see the docs for how to use it (basically just add `"matches": [Match(wm_class="vlc")]` to the relevant split).

nazar-pc:
One slightly annoying thing: `matches` locks windows permanently, I can't move them to other splits even if I try. Can this behavior be only applied to new windows as they are, but not affecting existing windows? Other layouts like Tile allow moving, but unfortunately re-sort once any other window is opened (which is also annoying).

I'm going to use this layout with Plasm inside of splits (https://github.com/numirias/qtile-plasma/pull/25 is necessary if you're planning to do the same) as the main setup going forward.

elParaguayo:
Sorry - that was a silly mistake by me. Yes, we can fix it so matches are only applied when windows are opened. Not when we're moving them.

nazar-pc:
Hm, last commit doesn't seem to match windows anymore

elParaguayo:
> Hm, last commit doesn't seem to match windows anymore

And docs have failed... I must have mucked up the commit!

nazar-pc:
Please DM me with USDC wallet address, I'd like to send you a few bucks as a thank you :slightly_smiling_face:

nazar-pc:
Unless I'm doing something wrong I don't think prev/next_split methods actually work. Moving does work, just switching doesn't seem to. I have `follow_mouse_focus = True` in my layout if that affects it.

elParaguayo:
Hmm. I suspect `follow_mouse_focus` will probably affect as the active split will be whichever one has the focussed window.

nazar-pc:
Running this layout for a month now, except above nuance works great. Is this waiting for someone's review before being merged?

elParaguayo:
We've got a big PR to merge first and then we'll look at getting others merged.

I'm also dealing with an issue with our test suite at the moment too.

elParaguayo:
Can you recreate this without using the Plasma layout?

nazar-pc:
Please don't close, I'm using it and looking forward for it to be merged :pray:

nazar-pc:
Can someone reopen this? I'd hate it to be forgotten, this is an extremely useful layout!

elParaguayo:
I wonder if i need to put this in qtile-extras...

Nebucatnetzer:
I sounds quite interesting for large monitors.
I'm currently looking at fake_screens because I would like to have one quarter or my monitor use TreeTab because applications like Remmina open new windows and mess up the layout and I still would like to have the main window at hand.

nazar-pc:
> I wonder if i need to put this in qtile-extras...

Regardless of where it goes, I just hope it gets merged and maintained. I don't think Qtile will be usable for me on 8k screen without this.

elParaguayo:
I think we'd like to include this in the main repo. I'll put this back on my to do list so I can tidy up the code and get the tests passing.

jwijenbergh:
Yes +1 this sounds like a must have for ultrawide monitor users

m-col:
+1

elParaguayo:
OK. I've rebased but I suspect there'll be some test failures for me to work on ;)

elParaguayo:
Everything's passing now so squashed into a single commit.

m-col:
Tests are pretty happy ;)

PR #3796, opened 2022-08-26
Add option to specify custom log path

elParaguayo:
This adds a new command line option ("-p", "--log-path") to define a custom path for the log file.

A bit niche but would be useful for my development work. I test development in a separate X session and have noticed some issues with my primary qtile session not writing to the log after the development session has written to that file. Writing to a separate file should solve that.

### DIFF
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -1,6 +1,7 @@
 import argparse
 import logging
 import sys
+from pathlib import Path
 
 from libqtile.log_utils import get_default_log, init_log
 from libqtile.scripts import check, cmd_obj, migrate, run_cmd, shell, start, top
@@ -20,6 +21,14 @@ except ModuleNotFoundError:
         VERSION = "dev"
 
 
+def check_folder(value):
+    path = Path(value)
+    if not path.parent.is_dir():
+        raise argparse.ArgumentTypeError("Log path destination folder does not exist.")
+    # init_log expects a Path object so return `path` not `value`
+    return path
+
+
 def main():
     parent_parser = argparse.ArgumentParser(add_help=False)
     parent_parser.add_argument(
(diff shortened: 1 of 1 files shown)


m-col:
Perhaps we can automate this in some way such that separate Qtile sessions always have unique log files? The obvious idea would be to put the value of DISPLAY/WAYLAND_DISPLAY into the log file name, like with the IPC sockets. Currently `init_log` is called right before whichever CLI command is called (`options.func(options)`), but we could move it to inside the function. If this might be a preferable approach (feel free to say 'nah, the command line argument is more easy to use'!) I'd suggest that:

- `start`: the log is initialised by `Core.__init__`, which is what knows what the display value is and currently is the first thing to log anything, so no behaviour changes
- `shell`, `cmd_obj`, `run_cmd`: i don't think any of these log, but they are all clients and can determine the log file from the display variable, like they do with the socket file
- `top`, `migrate`, `check`: these don't look like they log anything, instead printing to stdout

Thoughts?

elParaguayo:
I've nothing against some sort of automation here (kind of like what Xorg does with its `Xorg.0.log`) but I'd be reluctant to do anything that changes the default path.

m-col:
Ah yes - very good point. Perhaps we can check for the existence of socket files and if there aren't any (i.e. there are no other qtile processes running) we symlink the log path we're about to use to the current default path?

elParaguayo:
Is that overcomplicating things? I'm still leaning towards keeping the default path as is and only overriding this if it's explicitly set.

I'm not expecting many people to use this but it will be helpful for my debugging.

m-col:
yeah, I suppose that is overcomplicating things :) do you use the xephyr/wephyr scripts or have you put together your own debugging system?

elParaguayo:
I just use a different `.xinitrc` file on a separate TTY which loads a virtual environment. I would also test Wayland the same way but I can't get it running in the virtual environment at the moment.

elParaguayo:
Any objection to merging this?

m-col:
Nope, all good with me.

elParaguayo:
Have made a slight tweak (script now passes a `Path` object to `init_log` instead of a string).

m-col (approved):
Cool, even better

jwijenbergh:
LGTM?

elParaguayo:
The only thing I'm hesitant about now is whether we even need this!

Having been converted the benefits of xephy/wephyr for testing, log messages are just printed to stdout so the issue with multiple instances trying to access the same file doesn't seem to apply.

That said, I don't really see a downside to having this as it doesn't impact default behaviour.

PR #3799, opened 2022-08-27
Add xdg support to TaskList

### LINKED ISSUE #3637: No icon for some apps in Tasklist widget
### The issue:

```console
$ qtile --version
    0.21.1.dev67+g2b2cd60b
```

As far as I've checked, this only happens with apps that don't "publish" an X property for icon (_NET_WM_ICON, WM_ICON_NAME). Can the WM_CLASS property be used for this, possibly?

`I know what I have to do, but I don't know if I have the strength to do it.`

image
`That little smudge of color is supposed to be the icon :)`

image
`Normal behaviour`

```console
$ xprop
XdndProxy(WINDOW): window id # 0x4600004
WM_CLASS(STRING) = "spotify", "Spotify"
WM_NAME(STRING) = "Spotify"
_NET_WM_NAME(UTF8_STRING) = "Spotify"
_NET_WM_STATE(ATOM) = _NET_WM_STATE_FOCUSED
_NET_FRAME_EXTENTS(CARDINAL) = 0, 0, 0, 0
_NET_WM_DESKTOP(CARDINAL) = 5
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
		program specified location: 0, 0
_NET_WM_PID(CARDINAL) = 34927
WM_CLIENT_MACHINE(STRING) = "arch"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING
```

```python
widget.TaskList(
    parse_text=no_text,
    highlight_method="block",
    icon_size=48,
    border=colors[0],
    rounded=True,
    padding=0,
    padding_y=-3,
    margin_x=2,
    margin_y=45,
    txt_floating="🗗 ",
    txt_maximized="🗖 ",
    txt_minimized="🗕 "
),
```

### Required:

elParaguayo:
Adds local icon support to TaskList widget. This enables users to select themes for icons as well as providing icon support for Wayland backend.

Fixes #3637

Needs some testing!

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -27,7 +27,15 @@ import re
 
 import cairocffi
 
+try:
+    from xdg.IconTheme import getIconPath
+
+    has_xdg = True
+except ImportError:
+    has_xdg = False
+
 from libqtile import bar, hook, pangocffi
+from libqtile.images import Img
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
@@ -37,6 +45,9 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
 
     Contrary to WindowTabs this is an interactive widget.  The window that
     currently has focus is highlighted.
+
+    Optional requirements: `pyxdg <https://pypi.org/project/pyxdg/>`__ is needed
+    to use theme icons and to display icons on Wayland.
     """
 
     orientations = base.ORIENTATION_HORIZONTAL
@@ -140,6 +151,18 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             "icon_size",
             None,
             "Icon size. " "(Calculated if set to None. Icons are hidden if set to 0.)",
+        ),
+        (
+            "theme_mode",
+            None,
+            "When to use theme icons. `None` = never, `preferred` = use if available, "
+            "`fallback` = use if app does not provide icon directly. "
+            "`preferred` and `fallback` have identical behaviour on Wayland.",
+        ),
+        (
+            "theme_path",
+            None,
+            "Path to icon theme to be used by pyxdg for icons. ``None`` will use default icon theme.",
         ),
     ]
 
@@ -288,10 +311,19 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
 
-        if qtile.core.name == "wayland" and self.icon_size != 0:
+        if not has_xdg and self.theme_mode is not None:
+            logger.warning("You must install pyxdg to use theme icons.")
+            self.theme_mode = None
+
+        if self.theme_mode and self.theme_mode not in ["preferred", "fallback"]:
+            logger.warning(
+                "Unexpected theme_mode (%s). Theme icons will be disabled.", self.theme_mode
+            )
+            self.theme_mode = None
+
+        if qtile.core.name == "wayland" and self.theme_mode is None and self.icon_size != 0:
             # Disable icons
             self.icon_size = 0
-            logger.warning("TaskList icons not supported in Wayland.")
 
         if self.icon_size is None:
             self.icon_size = self.bar.height - 2 * (self.borderwidth + self.margin_y)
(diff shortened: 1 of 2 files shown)


libqtile/confreader.py
```diff
@@ -122,7 +122,7 @@ def load(self):
         sys.path.insert(0, path.parent.as_posix())
 
         if name in sys.modules:
-            self._reload_config_submodules(path)
+            # self._reload_config_submodules(path)
```
ervinpopescu, line 125: This borks my config :( I need to always restart, just like those good old times :)

Otherwise, this works perfectly. I tried adding the line and I saw some weird flickers and remains of what had already been drawn on the last reload. Is that why you commented it out?

elParaguayo: Argh. That's not meant to be in there! Sorry.

ervinpopescu:
Well for some reason, having just tried to copy and paste the latest code and having modified my config accordingly before that, a live restart does the job, i.e. reloading works fine. I'm going to test this out more, with a logout and a shutdown, which should be fine, and then I'll remove the `reload` from my config, which should be redundant now, right?

ervinpopescu:
The code works perfectly, and I fixed the cursor issues, by editing `/usr/share/icons/default/index.theme` instead of `$HOME/.icons/default/index.theme`. Any idea why this was necessary?

ervinpopescu:
> I fixed the cursor issues, by editing `/usr/share/icons/default/index.theme` instead of `$HOME/.icons/default/index.theme`. Any idea why this was necessary?

Nevermind, this was unrelated to qtile...

libqtile/widget/tasklist.py
```diff
@@ -141,6 +149,14 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
...
             "Icon size. " "(Calculated if set to None. Icons are hidden if set to 0.)",
         ),
+        (
+            "theme_mode",
+            None,
+            "When to use theme icons. `None` = never, `preferred` = use if available, "
+            "`fallback` = use if app does not provide icon directly. "
+            "`preferred` and `fallback` have identical behaviour on Wayland.",
+        ),
+        ("theme_path", None, ""),
```
m-col, line 159: Perhaps a doc here? Just reading through https://pyxdg.readthedocs.io/en/latest/icontheme.html and it looks like `None` is a good default, so it might be worth mentioning that `None` will use the default icon theme in the doc (if I've understood that page correctly)

libqtile/widget/tasklist.py
```diff
@@ -288,10 +304,19 @@ def calc_box_widths(self):
     def _configure(self, qtile, bar):
         base._Widget._configure(self, qtile, bar)
 
-        if qtile.core.name == "wayland" and self.icon_size != 0:
+        if not has_xdg and self.theme_mode is not None:
+            logger.warning("You must install pyxdg to use theme icons.")
+            self.theme_mode = None
+
+        if self.theme_mode not in ["preferred", "fallback"]:
```
m-col, line 311: `None` seems to be a valid value so in that case the warning may be unnecessary?

```suggestion
        if self.theme_mode and self.theme_mode not in ["preferred", "fallback"]:
```

libqtile/widget/tasklist.py
```diff
@@ -417,16 +438,63 @@ def get_window_icon(self, window):
...
+
+    def _get_theme_icon(self, window):
+        classes = window.get_wm_class()
+
+        if not classes:
+            return None
+
+        icon = None
+
+        for cl in window.get_wm_class():
```
m-col, line 451:
```suggestion
        for cl in classes:
```

m-col (commented):
Looks good - I've made a couple of comments. Could you add a note about the optional pyxdg dependency in the widget's main docstring?

I wonder if it would be worth testing the combinations of backend vs `theme_mode` values, though I'm unsure how straight forward setting up system icon themes or window icons would be in the test environment so nw if not.

elParaguayo:
Nice. Thanks for the comments. Will update soon.

libqtile/widget/tasklist.py
elParaguayo: Agree with the change but not sure I follow your comment. If `theme_mode` is `None` then we don't use xdg icons. So `None` is valid but it disables icons. If someone's entered a string here then they want icons but have made a mistake with the value.

elParaguayo:
Have made those changes. Left one comment open for you. If you're happy (and the tests pass) then we can merge this.

libqtile/widget/tasklist.py
m-col: I was just thinking about the warning message. If it was `None` then `"Unexpected theme_mode None. Theme icons will be disabled."` would have been logged, which may have been confusing to users because it isn't unexpected.

LGTM!

elParaguayo: Ah, of course.

PR #3804, opened 2022-08-29
Fix focus on click after clicking on root window

### LINKED ISSUE #3800: Clicking blank screen then previous window doesn't focus clicked window
### The issue:

I have two screens, and `follow_mouse_focus = False`.

Suppose I have two windows on my first monitor, A and B, A is currently focused, and I have no windows on my second monitor.

If I click my second monitor, the focused screen is now my second monitor. 
Interestingly, I can still type stuff in A afterwards, but it doesn't have the focus outline anymore.

Now if I click A, not only does qtile not shift the focus to A (as in no focus outline), my second monitor also remains the focused screen. 
**I expect that if I click A, A should be refocused and that my first monitor is also now focused.**

However, if I click B, everything works as expected. 

However, if I click B, everything works as expected.

I don't see any relevant logs, and I'm running qtile version 0.21.0 on Archlinux. 

### Required:

elParaguayo:
Fixes a bug where opening a window on screen A, clicking on Screen B (with no windows) and clicking back on the window does not focus the window.

Fixes #3800

@m-col - I don't love this. It works, but I can't see why it's needed when there's only a single window open. The bug is not there when you have multiple windows open on the first screen and click on a different window.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -888,6 +888,8 @@ class Core(base.Core):
             # clicked on root window
             screen = qtile.find_screen(e.root_x, e.root_y)
             if screen:
+                if qtile.current_window:
+                    qtile.current_window._grab_click()
                 qtile.focus_screen(screen.index, warp=False)
 
     def flush(self):
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1986,12 +1986,7 @@ class Window(_Window, base.Window):
                 self.group.focus(self, False)
             if self.group.screen and self.qtile.current_screen != self.group.screen:
                 self.qtile.focus_screen(self.group.screen.index, False)
-        else:
-            self._grab_click()
         return True
-
-    def handle_LeaveNotify(self, e):  # noqa: N802
-        self._ungrab_click()
 
     def handle_ButtonPress(self, e):  # noqa: N802
         self.qtile.core.focus_by_click(e, window=self)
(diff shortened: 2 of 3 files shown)


m-col:
I think your solution is in the right direction, but I suspect the issue is a quite specific edge case where our current logic bugs a bit. It's because the way we 'focus' screens isn't quite the same as how we focus windows. Currently all non-focussed windows have their buttons grabbed, and only the focussed doesn't, and I think compared to using Enter/LeaveNotify this would reduce the number of events when moving the mouse around (particularly with focus_follow_mouse as False). So I think we just have to `_grab_click` the focussed window if we're focussing a screen. Indeed, this seems to fix it too:

```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index 337989bf..e052b407 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -796,6 +796,8 @@ class Core(base.Core):
             # clicked on root window
             screen = qtile.find_screen(e.root_x, e.root_y)
             if screen:
+                if qtile.current_window:
+                    qtile.current_window._grab_click()
                 qtile.focus_screen(screen.index, warp=False)
 
     def flush(self):
```

tych0:
Just driving by: does this happen because of weird `RevertPointerTo` settings? Something like 7b43c9f8af839831306a99d717a9afb0280a7189 maybe?

elParaguayo:
> Just driving by: does this happen because of weird `RevertPointerTo` settings? Something like [7b43c9f](https://github.com/qtile/qtile/commit/7b43c9f8af839831306a99d717a9afb0280a7189) maybe?

Yes. That's my assumption.

Need to revisit this.

elParaguayo:
Tidying up branches on my laptop and found some old issues that weren't resolved!

@m-col I've implemented your suggestion here. However, there's another case we need to handle: when the bar is clicked on a different screen that also focuses the screen. We need similar code in the bar to make sure we grab clicks for that window. However, as the bar is not in `libqtile/backend/x11` I need to check the backend name. It's a bit ugly but it does work.

ramnes (approved):
Yeah it's a bit ugly but it works. Let's go. :)

PR #3809, opened 2022-09-07
Expose Pomodoro widget toggle break/active methods as commands

m-col:
They are already used for mouse callbacks but this lets them be controlled by keybindings/IPC etc


elParaguayo:
LGTM

Issue #3440, opened 2022-03-21
Error caused by screen change hook persists after deletion


xtalax:
### The issue:

I am getting the following error with [this](https://gist.github.com/xtalax/bd2d1a4237fb2181f436e41467c9f358) config:
```
2022-03-21 19:06:13,176 ERROR libqtile hook.py:fire():L392 Error in hook screen_change
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/libqtile/hook.py", line 390, in fire
    i(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/libqtile/core/manager.py", line 359, in cmd_reconfigure_screens
    self._process_screens()
  File "/usr/lib/python3.10/site-packages/libqtile/core/manager.py", line 340, in _process_screens
    scr._configure(self, i, x, y, w, h, grp, reconfigure_gaps=reconfigure_gaps)
  File "/usr/lib/python3.10/site-packages/libqtile/config.py", line 302, in _configure
    self.set_group(group)
  File "/usr/lib/python3.10/site-packages/libqtile/config.py", line 364, in set_group
    g1 = self.group
AttributeError: 'Screen' object has no attribute 'group'
```

I did have a `screen_change` hook defined previously:
```
@hook.subscribe.screen_change
def set_screens(event):
    subprocess.run(["autorandr", "--change"])
    # lazy.spawn("mydock")
    qtile.restart()

```
The error persists after deleting the `screen_change` hook and reloading the config

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
Can I check: are you saying you get the error if you remove that hooked function and start qtile?

elParaguayo:
OK. Next question: are you on v0.20 or the latest git version? We've made a few changes to reconfiguring screens since v0.20 so this issue may be fixed (and no one else seems to have reported this recently either).

Are you able to install the git version?

xtalax:
I've been having trouble with the AUR package for the git version, 
`error: failed retrieving file 'python-pytest-7.0.1-1-any.pkg.tar.zst' from mirrors.celianvdb.fr : The requested URL returned error: 404` etc.

Do you have a link to manual install instructions?

elParaguayo:
That suggests your mirrors are out of date. When did you last do a full update? 

[...]

PR #3810, opened 2022-09-07
Ensure Screens always have an Group associated group

m-col:
Currently sometimes AttributeErrors are raised:

```
AttributeError: 'Screen' object has no attribute 'group'
```

This avoids that by ensuring `Screen`s get a group the first time they are configured, and this removes any `hasattr` tests and instead checks things directly.

Possibly related: https://github.com/qtile/qtile/issues/3440

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -27,7 +27,6 @@
 #
 from __future__ import annotations
 
-import contextlib
 import os.path
 import sys
 from typing import TYPE_CHECKING
@@ -38,7 +37,7 @@ from libqtile.command.base import CommandObject
 from libqtile.log_utils import logger
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, ContextManager, Iterable
+    from typing import Any, Callable, Iterable
 
     from libqtile.backend import base
     from libqtile.bar import BarType
@@ -359,7 +358,6 @@ class Screen(CommandObject):
     """
 
     group: _Group
-    previous_group: _Group
     index: int
 
     def __init__(
@@ -389,6 +387,7 @@ class Screen(CommandObject):
         self.y = y if y is not None else 0
         self.width = width if width is not None else 0
         self.height = height if height is not None else 0
+        self.previous_group: _Group | None = None
 
     def _configure(
         self,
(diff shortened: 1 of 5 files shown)


libqtile/config.py
```diff
@@ -392,6 +391,22 @@ def _configure(
...
+
+        if not hasattr(self, "group"):
+            # Ensure that this screen actually *has* a group. It will get this the first
+            # time this method is called, and until then accessing self.group will raise
+            # an AttributeError. The 'right' approach would be typing `self.group` as
+            # `_Group | None` and setting it to `None` upon init, but conceptually no
+            # screens should ever not have a group and this helps us avoid having to
+            # check for `None`s everywhere.
+            for grp in qtile.groups:
+                if grp.screen is None or grp.screen is self:
```
elParaguayo, line 403: Why do we check if the group's `screen` is None? Isn't the `screen` value only set when the group is visible? If that's the case, this check will put the first available group on the screen and ignore the `group` parameter in this method call.

I could have it completely wrong but it just looks odd to me.

m-col: It is odd, so let's definitely make the comment more clear so that future readers of the code understand the logic, and let's make sure the logic actually does make sense :D 

Basically the idea is that this ensures that when `Qtile._process_screens` runs (config loading, reloading or `cmd_reconfigure_screens`), every screen has a group so that `screen.group` never raises an `AttributeError` (as no group as assigned during `__init__`). _Then_, the desired group (the `group` parameter to this method) is set as the screen's actual group. Thus, any additional operations within `self.set_group(group)` can do their thing. Often, this is a no-op because it'll be the very same screen, but sometimes it won't be, and importantly this is the case that raises the origin `AttributeError` that is sometimes reported, because `self.set_group(group)` tries to access `self.group` before it's every been set to something.

Looking at this again I think it might be better to move this block out of this method and into `Qtile._process_screens` directly to run right before `scr._configure` is called, so it's more like `Qtile` is setting up the screen rather than the screen bootstrapping itself, but maybe that's unimportant.

m-col: I've moved this block out into `Qtile._process_screens`, reworded the comment a bit, and used the group that is about to get passed into `scr._configure` so hopefully it's more clear what's going on

PR #3811, opened 2022-09-07
repr Window objects using the name of their class

m-col:
There should be no noticable changes in x11 because there is only 1 `Window` and `Static` class, but in wayland we have one of each for all of xwayland, xdg shell and layer shell.

This is more helpful for debugging than anything else, e.g.:
```
>>> qtile cmd-obj -o cmd -f eval -a "self.windows_map"
(True,
 "{16: XdgWindow(name='[tmux] zsh:~', wid=16), 17: Internal(wid=17), 18: "
 "XdgWindow(name='zsh:.../libqtile/backend', wid=18), 19: XdgWindow(name='vim "
 "[1] ~/git/qtile/.git/COMMIT_EDITMSG', wid=19)}")
 ```
More informative debugging info 😄

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -266,7 +266,7 @@ class Window(_Window, metaclass=ABCMeta):
     float_y: int | None
 
     def __repr__(self):
-        return "Window(name=%r, wid=%i)" % (self.name, self.wid)
+        return "%s(name=%r, wid=%i)" % (self.__class__.__name__, self.name, self.wid)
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
> The test failure is unrelated: `test_statusnotifier_left_click_vertical_bar`

Yeah - this test has been quite flaky recently. Weird as it's always that one.

elParaguayo:
LGTM - happy for this to be merged if you are.

Issue #3403, opened 2022-03-04
Wayland: disabled outputs don't seem to be removed from the output layout's geometry


m-col:
### The issue:

If I have 3 monitors placed in a row and use kanshi to disable the far end one, the mouse can still move off the side of the second monitor into where the 3rd would be. Moving the disabled output's position (e.g. using wdisplays) fixes this.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
I have a fix for this, just testing out and I'll submit the PR in a few days when I've confirmed there are no unexpected side effects.

m-col:
I believe this was fixed by https://github.com/qtile/qtile/pull/3812

Issue #3402, opened 2022-03-03
[Wayland] Qtile misses some display initialisation code


arjan-s:
### The issue:

It seems that Qtile (current git master) is missing some display initialization code. When starting Qtile right after system boot, it doesn't initialize all my monitors automatically. `kanshi`, `wlr-randr` and `wdisplays` don't even see all the outputs. However, if I first start Sway, which does intialize and recognize all outputs, then exit it and then start Qtile, it correctly sees all outputs as well.

Relevant logging:
Directly after boot:
```
2022-03-03 08:58:26,614 DEBUG libqtile core.py:_on_new_output():L281 Signal: backend new_output_event
2022-03-03 08:58:26,637 DEBUG libqtile core.py:_on_output_layout_change():L303 Signal: output_layout change_event
2022-03-03 08:58:26,638 DEBUG libqtile core.py:_on_new_output():L281 Signal: backend new_output_event
2022-03-03 08:58:26,644 DEBUG libqtile core.py:_on_output_layout_change():L303 Signal: output_layout change_event
```
After starting Sway first:
```
2022-03-03 09:03:31,177 DEBUG libqtile core.py:_on_new_output():L281 Signal: backend new_output_event
2022-03-03 09:03:31,192 DEBUG libqtile core.py:_on_output_layout_change():L303 Signal: output_layout change_event
2022-03-03 09:03:31,193 DEBUG libqtile core.py:_on_new_output():L281 Signal: backend new_output_event
2022-03-03 09:03:31,200 DEBUG libqtile core.py:_on_output_layout_change():L303 Signal: output_layout change_event
2022-03-03 09:03:31,200 DEBUG libqtile core.py:_on_new_output():L281 Signal: backend new_output_event
2022-03-03 09:03:31,206 DEBUG libqtile core.py:_on_output_layout_change():L303 Signal: output_layout change_event
```

This is probably the same issue as I encountered when running Qtile in a VirtualBox VM: after starting Qtile it simply keeps using the KMS (kernel modesetting) screen resolution, instead of correctly initializing the output with the maximum available resolution. Sway does this correctly. Also, when resizing the VirtualBox display, Qtile does not change the output resolution accordingly, while Sway does.

It may also be related to #3389.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
Hey @arjan-s, would you be able to test to see if this was fixed by https://github.com/qtile/qtile/pull/3812? It's merged, and in the latest release.

[...]

PR #3812, opened 2022-09-07
Various Wayland quality of life improvements

m-col:
- fix bug where windows can't be `__repr__`'d before mapping for the first time due to having no WID
- remove some unused attributes on `Core`
- fix adding/removing of outputs from output layout, fixing a swathe of output-related issues
- remove a very noisy and not very helpful debug log message

Requires https://github.com/flacjacket/pywlroots/pull/102


m-col:
Fixes https://github.com/qtile/qtile/issues/3403

m-col:
I suspect this will fix https://github.com/qtile/qtile/issues/3402 too

m-col:
The status notifier test is again failing here, so the CI otherwise passes.

We should do a release very soon I think :)

PR #3813, opened 2022-09-08
Evaluate cmd_eval code with globals of command object's module

m-col:
Currently code executed with `cmd_eval` calls is called within the definition of `cmd_eval` i.e. inside `libqtile.command.base`. Sometimes when I'm debugging stuff from the command line with evals I wish they were evaluated within the namespace of the module that defined the command object. So, this PR does just that!

### DIFF
--- a/libqtile/command/base.py
+++ b/libqtile/command/base.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import abc
 import inspect
+import sys
 import traceback
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


elParaguayo:
Nice! This has bugged me for a while.

m-col:
The test failure is due to a change in pywlroots. Qtile is updated accordingly in https://github.com/qtile/qtile/pull/3812.

elParaguayo:
LGTM - happy for this to be merged if you are.

PR #3820, opened 2022-09-10
Mpd2 Widget: status currentsong value

mrtnvgr:


elParaguayo:
Thanks for posting this. I'll try to take a look tomorrow and provide some comments.

libqtile/widget/mpd2widget.py
```diff
@@ -257,6 +257,8 @@ def formatter(self, status, current_song):
         song_info = defaultdict(lambda: default)
         song_info["play_status"] = self.play_states[status["state"]]
 
+        status["currentsong"] = str(int(status["song"])+1)
```
elParaguayo, line 260: I don't think you need the `str()` conversion as this is just used for string formatting so an integer is fine.

Is `song` always available and always an integer? If not, this line is going to fail.

You'll also need to update the `test_mpd2widget.py` file as the field is not available there.

elParaguayo: Can you also add a test that shows this field being used.

elParaguayo:
Don't worry - I haven't forgotten this. Just very busy at the moment!

elParaguayo:
@mrtnvgr I've tidied up the test for this (I've pushed an extra commit to your branch).

A few questions for you:
- Is the output of `song_info["song"]` always a string? If it's an integer, this will cause a crash when `html.escape` is called on it.
- Is the song index zero-indexed? You're adding one to it in the code.

If you're happy with those then we need to squash this into a single commit and force push back to your branch.

mrtnvgr:
> - Is the output of `song_info["song"]` always a string? If it's an integer, this will cause a crash when `html.escape` is called on it.
Yes

> - Is the song index zero-indexed? You're adding one to it in the code.
Yes

elParaguayo:
OK. Have you been using this version of the widget in your own setup? is it working ok?

mrtnvgr:
> OK. Have you been using this version of the widget in your own setup? is it working ok?

Works fine

PR #3821, opened 2022-09-11
Update hooks.rst

jlcoulter:
Suggestion for people like me. I tried cd into ~/.config/qtile/ and running "chmod +x autostart.sh". Which didn't execute on start. After running with the full path it is working as intended. This change would have helped me.

### DIFF
--- a/docs/manual/config/hooks.rst
+++ b/docs/manual/config/hooks.rst
@@ -43,7 +43,7 @@ emitted every time Qtile starts (including restarts), whereas ``startup_once``
 is only emitted on the very first startup.
 
 Let's create an executable file ``~/.config/qtile/autostart.sh`` that will
-start a few programs when Qtile first runs. Remember to `chmod +x` this file so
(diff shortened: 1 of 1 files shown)


docs/manual/config/hooks.rst
```diff
@@ -43,7 +43,7 @@ emitted every time Qtile starts (including restarts), whereas ``startup_once``
 is only emitted on the very first startup.
 
 Let's create an executable file ``~/.config/qtile/autostart.sh`` that will
-start a few programs when Qtile first runs. Remember to `chmod +x` this file so
+start a few programs when Qtile first runs. Remember to `chmod +x ~/.config/qtile/autostart.sh` this file so
```
elParaguayo, line 46:
```suggestion
start a few programs when Qtile first runs. Remember to `chmod +x ~/.config/qtile/autostart.sh` so
```

m-col:
Great, thanks! Will merge when tests pass :)

Issue #3825, opened 2022-09-11
[Docs] FAQ bug


ervinpopescu:
### The issue:

![image](https://user-images.githubusercontent.com/84532402/189532481-58906133-9dab-40fa-be56-2e7e468bc950.png)

I hope this is not intended behavior. :)

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
Ha. No. Will fix.

PR #3826, opened 2022-09-11
Tidy up cairo FAQ in docs

elParaguayo:
This is in relation to #3825. Added some css to break `code` lines if they're too long.  I've also reformatted the FAQ to tidy it up a bit.

Looks like this on RTD:


m-col (approved):
Nice, LGTM

PR #3827, opened 2022-09-12
added xmodmap support

ShiNoNeko47:
Loading ~/.Xmodmap when changing keyboard layout in X11.


libqtile/widget/keyboardlayout.py
```diff
@@ -94,6 +95,8 @@ def set_keyboard(self, layout: str, options: str | None) -> None:
             logger.error("Can not change the keyboard layout:")
         except OSError:
             logger.error("Please, check that setxkbmap is available:")
+        else:
+            os.system("xmodmap $HOME/.Xmodmap")
```
m-col, line 99: Perhaps we should test that the file exists first before calling it?

It might also be good to use the same `try` `check_output` approach used right before it, and catch and log any errors that arise.

m-col (commented):
Is there any reason why somebody would _not_ want this to be called - i.e. should it be configurable, or should this always be the expected behaviour?

ShiNoNeko47:
Just added logging.

I think it should be the expected behaviour. If the user doesn't want it called they can just remove or rename the file if they have it.

m-col (commented):
In the docstring for the widget there is a line that mentions setxkbmap - could you also add a bit saying that xmodmap will be used too if it's available. And ive suggested another change in line :)

PR #3833, opened 2022-09-14
Fix Systray inheritance order

elParaguayo:
I'll be honest, I'm confused by this one!

Switching the inheritance order of the base classes fixes a bug where users using RectDecorations from qtile-extras have colours inverted (i.e. the decoration's fill colour is the widget background colour and the background colour is the widet's fill colour).

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -96,7 +96,9 @@ class Icon(window._Window):
     handle_UnmapNotify = handle_DestroyNotify  # noqa: N815
 
 
-class Systray(window._Window, base._Widget):
+# Mypy doesn't like the inheritance of height and width as _Widget's
+# properties are read only but _Window's have a getter and setter.
+class Systray(base._Widget, window._Window):  # type: ignore[misc]
     """
     A widget that manages system tray.
 


m-col:
Interesting indeed. Any ideas on what the cause is? What shared attributes do those two classes have? I see `self.group` at least

elParaguayo:
I'll post some screenshots to show you the issue but I do want to understand why there is an issue before merging.

elParaguayo:
Here is current behaviour with a decoration:
2022-09-15-190828_150x28_scrot
(Notice the decoration colours are inverted)

With this PR:
2022-09-15-190905_150x28_scrot
Colours now the right way round.

I'm doing a separate PR in qtile-extras to fix the icon background as that's an issue caused by the decoration but I need this PR as well.

However, let me see if I can track down the offending attribute first!

elParaguayo:
Ok. Think I've worked this one out.

The shared attributes are `info`, `width` and `height`. By having the `_Window` inheritance first, those definitions take priority. This mean the widget has a `height` of `1` and a `width` of `1` irrespective of the number of icons being displayed.

This causes an issue with decorations as they need to look at the parent widget's dimensions to calculate how to draw themselves.

m-col (approved):
Nice, LGTM

PR #3836, opened 2022-09-15
WL: emit screen_change hook every time layout changes

m-col:
Currently it's only emitted when the output layout is changed via an external tool, but actually it's enough for a physical monitor to be plugged/unplugged to trigger the change.


m-col:
This isn't quite right, and is causing some segfaults for me. I've recently got my hands on a physical monitor that doesn't play by the rules and it's annoying but good for identifying potential issues that need patching :) setting to draft for now.

PR #3842, opened 2022-09-17
FAQ: add solution for matching bars with picom

m-col:


m-col:
Here's the render: https://qtile--3842.org.readthedocs.build/en/3842/manual/faq.html#how-can-i-match-the-bar-with-picom

ramnes:
nice

PR #3843, opened 2022-09-18
Fix various mirror bugs

elParaguayo:
There are a few issues where mirrors don't tidy themselves up if they're removed from a running session (e.g. they're on a bar on a screen that is disabled).

This PR fixes this by:
- Introducing better handling of mirrors through the reflected widget rather than directly in the `Drawer` object
- Ensuring mirrors finalize themselves
- Removing dead widgets from `qtile.widgets_map`
- Unsetting hooks in various widgets so `bar.draw()` is not called on dead screens.

### DIFF
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -641,7 +641,7 @@ class Drawer:
         self.ctx: cairocffi.Context
         self._reset_surface()
 
-        self.mirrors: set[Drawer] = set()
+        self._has_mirrors = False
 
         self.current_rect = (0, 0, 0, 0)
         self.previous_rect = (-1, -1, -1, -1)
@@ -652,9 +652,16 @@ class Drawer:
         self.surface = None
         self.ctx = None
 
-    def add_mirror(self, mirror: Drawer):
-        """Keep details of other drawers that are mirroring this one."""
-        self.mirrors.add(mirror)
+    @property
+    def has_mirrors(self):
+        return self._has_mirrors
+
+    @has_mirrors.setter
+    def has_mirrors(self, value):
+        if value and not self._has_mirrors:
+            self._create_last_surface()
+
+        self._has_mirrors = value
 
     @property
     def width(self) -> int:
@@ -679,6 +686,10 @@ class Drawer:
             None,
         )
         self.ctx = self.new_ctx()
+
+    def _create_last_surface(self):
+        """Creates a separate RecordingSurface for mirrors to access."""
+        self.last_surface = cairocffi.RecordingSurface(cairocffi.CONTENT_COLOR_ALPHA, None)
 
     @property
     def needs_update(self) -> bool:
@@ -769,9 +780,8 @@ class Drawer:
         """
         if self._enabled:
             self._draw(offsetx, offsety, width, height)
-            if self.mirrors:
-                # mypy is tripping over CONTENT_COLOR_ALPHA here despite it working earlier in this file!
-                self.last_surface = cairocffi.RecordingSurface(cairocffi.CONTENT_COLOR_ALPHA, None)  # type: ignore
+            if self.has_mirrors:
+                self._create_last_surface()
                 ctx = cairocffi.Context(self.last_surface)
                 ctx.set_source_surface(self.surface)
                 ctx.paint()
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -383,6 +383,10 @@ class Bar(Gap, configurable.Configurable, CommandObject):
 
     def kill_window(self):
         """Kill the window when the bar's screen is no longer being used."""
+        for name, w in self.qtile.widgets_map.copy().items():
+            if w in self.widgets:
+                w.finalize()
+                del self.qtile.widgets_map[name]
         self.drawer.finalize()
         self.window.kill()
         self.window = None
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -170,6 +170,7 @@ class _Widget(CommandObject, configurable.Configurable):
 
         self.configured = False
         self._futures: list[asyncio.TimerHandle] = []
+        self._mirrors: set[_Widget] = set()
 
     @property
     def length(self):
@@ -368,6 +369,43 @@ class _Widget(CommandObject, configurable.Configurable):
 
     def mouse_leave(self, x, y):
         pass
+
+    def _draw_with_mirrors(self) -> None:
+        self._old_draw()
+        for mirror in self._mirrors:
+            if not mirror.configured:
+                continue
+
+            # If the widget and mirror are on the same bar then we could have an
+            # infinite loop when we call bar.draw(). mirror.draw() will trigger a resize
+            # if it's the wrong size.
+            if mirror.length_type == bar.CALCULATED and mirror.bar is not self.bar:
+                mirror.bar.draw()
+            else:
+                mirror.draw()
+
+    def add_mirror(self, widget: _Widget):
+        if not self._mirrors:
+            self._old_draw = self.draw
+            self.draw = self._draw_with_mirrors  # type: ignore
+
+        self._mirrors.add(widget)
+        if not self.drawer.has_mirrors:
+            self.drawer.has_mirrors = True
+
+    def remove_mirror(self, widget: _Widget):
+        try:
+            self._mirrors.remove(widget)
+        except KeyError:
+            pass
+
+        if not self._mirrors:
+            self.drawer.has_mirrors = False
+
+            if hasattr(self, "_old_draw"):
+                # Deletes the reference to draw and falls back to the original
+                del self.draw
+                del self._old_draw
 
 
 UNSPECIFIED = bar.Obj("UNSPECIFIED")
@@ -841,7 +879,6 @@ class Mirror(_Widget):
 
     def __init__(self, reflection, **config):
         _Widget.__init__(self, reflection.length, **config)
-        reflection.draw = self.hook(reflection.draw)
         self.reflects = reflection
         self._length = 0
         if self.reflects.length_type == bar.STRETCH:
@@ -849,7 +886,7 @@ class Mirror(_Widget):
 
     def _configure(self, qtile, bar):
         _Widget._configure(self, qtile, bar)
-        self.reflects.drawer.add_mirror(self.drawer)
+        self.reflects.add_mirror(self)
         # We need to fill the background once before `draw` is called so, if
         # there's no reflection, the mirror matches its parent bar.
         self.drawer.clear(self.background or self.bar.background)
@@ -863,16 +900,6 @@ class Mirror(_Widget):
     @length.setter
     def length(self, value):
         self._length = value
-
-    def hook(self, draw):
-        def _():
-            draw()
-            if self.length_type == bar.STRETCH:
-                self.bar.draw()
-            else:
-                self.draw()
-
-        return _
 
     def draw(self):
         if self.length_type != bar.STRETCH and self._length != self.reflects.length:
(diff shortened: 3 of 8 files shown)


libqtile/bar.py
```diff
@@ -383,6 +383,11 @@ def finalize(self):
 
     def kill_window(self):
         """Kill the window when the bar's screen is no longer being used."""
+        for name in list(self.qtile.widgets_map.keys()):
+            w = self.qtile.widgets_map[name]
+            if w in self.widgets:
+                w.finalize()
+                del self.qtile.widgets_map[name]
```
m-col, line 390:
```suggestion
        for w, name in self.qtile.widgets_map.copy().items():
            if w in self.widgets:
                w.finalize()
                del self.qtile.widgets_map[name]
```

libqtile/widget/base.py
```diff
@@ -369,6 +370,35 @@ def mouse_enter(self, x, y):
...
+
+            if mirror.length_type == bar.CALCULATED:
+                mirror.bar.draw()
+            else:
+                mirror.draw()
+
+    def add_mirror(self, widget: _Widget):
+        if not self.mirrors:
+            self._old_draw = self.draw
+            self.draw = self._draw_with_mirrors  # type: ignore
```
m-col, line 387: What type error does this line create?

libqtile/backend/base.py
```diff
@@ -680,6 +691,10 @@ def _reset_surface(self):
         )
         self.ctx = self.new_ctx()
 
+    def _create_last_surface(self):
+        """Creates a separate RecordingSurface for mirrors to access."""
+        self.last_surface = cairocffi.RecordingSurface(cairocffi.CONTENT_COLOR_ALPHA, None)
+
```
m-col, line 697: Why is this called the last surface?

libqtile/backend/base.py
```diff
@@ -654,8 +654,19 @@ def finalize(self):
...
+        if not self.mirrors:
+            self._create_last_surface()
         self.mirrors.add(mirror)
 
+    def remove_mirror(self, mirror: Drawer):
+        try:
+            self.mirrors.remove(mirror)
+        except KeyError:
+            # This may get called multiple times but only the first call will work
+            pass
```
m-col, line 668: From what I can tell, each mirror would be removed exactly once. In what situation can this method be called more than once for the same widget and get this key error? Same with `_Widget.remove_mirror`

libqtile/widget/base.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, length, **config):
 
         self.configured = False
         self._futures: list[asyncio.TimerHandle] = []
+        self.mirrors: set[_Widget] = set()
```
m-col, line 173: It looks like `self.mirrors` is only accessed by `self` - if that's the case could we make it `_mirrors`?

libqtile/widget/base.py
```diff
@@ -369,6 +370,35 @@ def mouse_enter(self, x, y):
     def mouse_leave(self, x, y):
         pass
 
+    def _draw_with_mirrors(self):
```
m-col, line 373: mypy ignores methods that don't have at least one annotation so we can get it to check this one too with this:
```suggestion
    def _draw_with_mirrors(self) -> None:
```
(You know how much I love my typing! ;))

libqtile/widget/base.py
```diff
@@ -369,6 +370,35 @@ def mouse_enter(self, x, y):
...
+
+    def remove_mirror(self, widget: _Widget):
+        try:
+            self.mirrors.remove(widget)
+            self.drawer.remove_mirror(widget.drawer)
+        except KeyError:
+            pass
+
+        if not self.mirrors and hasattr(self, "_old_draw"):
+            self.draw = self._old_draw  # type: ignore
```
m-col, line 400: We might be able to get away with not saving `self._old_draw` and getting it from `self.__class__.draw`.

(2 mins later after some testing...) it looks like restoring it with `self.__class__.draw` causes a `missing 1 required positional argument: 'self'`, but `types.MethodType` can fix this: `self.draw = types.MethodType(self.__class__.draw, self)`. See: https://stackoverflow.com/questions/47797661/python-types-methodtype

But actually that's also unnecessary. Methods are looked up in the MRO when needed using the class definitions, so it's as straightforward as doing `del self.draw`, then later references to `self.draw` should get the original method :D

libqtile/widget/base.py
```diff
@@ -369,6 +370,35 @@ def mouse_enter(self, x, y):
...
         pass
 
+    def _draw_with_mirrors(self):
+        self._old_draw()
+        for mirror in self.mirrors:
+            if not mirror.configured:
+                continue
+
+            if mirror.length_type == bar.CALCULATED:
+                mirror.bar.draw()
```
m-col, line 380: Could this not result in an infinite loop between bar.draw and mirror.draw?

m-col (commented):
I've made a bunch of comments but it's mostly small things, I got a bit carried away because this part of the code is interesting :) Overall this solution to the problem is super nice, and thanks again for looking into it!

I do have one reservation and that's that both a reflected widget and it's drawer keep track of an equivalent set of mirrors. This seems a bit delicate to future changes because they need to be kept in sync. But I can see why it's needed - the reflected widget needs to know the _widgets_ that are mirroring it, but drawers need to know which _drawers_ are mirroring it, and not all drawers have widgets (e.g. popups) so there isn't always a 1-1 relationship there. Any thoughts on whether this could be improved?

I checked out other widgets to see if any needed to be updated so they could be set up properly aftering being finalized and I did find some. I made those changes so here is a diff:
```diff
diff --git a/libqtile/widget/mpd2widget.py b/libqtile/widget/mpd2widget.py
index 0ba5a68d..22d7ad63 100644
--- a/libqtile/widget/mpd2widget.py
+++ b/libqtile/widget/mpd2widget.py
@@ -177,11 +177,14 @@ class Mpd2(base.ThreadPoolText):
         super().__init__("", **config)
 
         self.add_defaults(Mpd2.defaults)
+        if self.color_progress:
+            self.color_progress = utils.hex(self.color_progress)
+
+    def _configure(self, qtile, bar):
+        super()._configure(qtile, bar)
         self.client = MPDClient()
         self.client.timeout = self.timeout
         self.client.idletimeout = self.idletimeout
-        if self.color_progress:
-            self.color_progress = utils.hex(self.color_progress)
 
     @property
     def connected(self):
diff --git a/libqtile/widget/notify.py b/libqtile/widget/notify.py
index 67ec18e8..a4643c73 100644
--- a/libqtile/widget/notify.py
+++ b/libqtile/widget/notify.py
@@ -223,4 +223,3 @@ class Notify(base._TextBox):
             # complete the finalisation of this widget.
             if task:
                 await task
-        base._TextBox.finalize(self)
diff --git a/libqtile/widget/pulse_volume.py b/libqtile/widget/pulse_volume.py
index 592ac408..65a56a65 100644
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -51,6 +51,8 @@ class PulseVolume(Volume):
         self.handle = ffi.new_handle(self)
         self.client_name = ffi.new("char[]", b"Qtile-pulse")
 
+    def _configure(self, qtile, bar):
+        super()._configure(qtile, bar)
         self.connect()
 
     def finalize(self):

```

libqtile/bar.py
elParaguayo: Think you've got `w` and `name` the wrong way round there but, otherwise, yes!

libqtile/widget/base.py
elParaguayo: `Cannot assign to a method`

libqtile/backend/base.py
elParaguayo: It's the last drawn surface. Once the drawer is drawn, the recording surface is cleared. The idea is to keep a copy of that RecordingSurface so mirrors can access it.

libqtile/backend/base.py
elParaguayo: That was my expectation to but I hit the `KeyError` in practice.

libqtile/widget/base.py
elParaguayo: Sure.

libqtile/widget/base.py
elParaguayo: I'd looked at `MethodType` and wanted to ignore it. `del self.draw` is neat!

libqtile/widget/base.py
elParaguayo: Hmm. It could if they were on the same bar...

elParaguayo:
Can you explain why the changes in your diff (re finalize) are needed?

elParaguayo:
> I do have one reservation and that's that both a reflected widget and it's drawer keep track of an equivalent set of mirrors. This seems a bit delicate to future changes because they need to be kept in sync. But I can see why it's needed - the reflected widget needs to know the _widgets_ that are mirroring it, but drawers need to know which _drawers_ are mirroring it, and not all drawers have widgets (e.g. popups) so there isn't always a 1-1 relationship there. Any thoughts on whether this could be improved?

This is a good question. I don't think the `Drawer` needs to know about all its mirrors, only if it has any. I wonder if we could simplify it e.g. have the widget set a `is_mirrored` flag on its `Drawer`. I'll have a play later on.

libqtile/widget/base.py
m-col: Strange edge case but what if there were 2 widgets, and each was passed to 2 bars, but the mirrors are on different bars. Then they will induce recursion in each other via their mirrors. Maybe this is impossible because one bar is guaranteed to be setup before the other and get all of the non-mirrored widgets, and that also prevents this happening when playing around with adding/removing screens etc. That sounds about right, right?

elParaguayo: Correct. Bars are initialised one at a time so I can't see a situation where Bar A has a mirror of a widget on Bar B, and Bar B has a mirror of a widget on Bar A.

m-col:
| Can you explain why the changes in your diff (re finalize) are needed?

Ah the change to the `Notify` widget? That was unrelated and I wasn't meant to paste that - I made the change while I was there and forgot about it, but that `base._TextBox.finalize(self)` is actually redundant because it's called via `async def _finalize(self):` which is scheduled by the regular `finalize` which also calls `base._TextBox.finalize(self)`.

elParaguayo:
> | Can you explain why the changes in your diff (re finalize) are needed?
> 
> Ah the change to the `Notify` widget? That was unrelated and I wasn't meant to paste that - I made the change while I was there and forgot about it, but that `base._TextBox.finalize(self)` is actually redundant because it's called via `async def _finalize(self):` which is scheduled by the regular `finalize` which also calls `base._TextBox.finalize(self)`.

OK. Were the mpd2 and pulse changes just needed to move some configurations out of `__init__`?

elParaguayo:
Think I've closed off your comments so let me know if I should merge this.

libqtile/widget/base.py
```diff
@@ -369,6 +370,42 @@ def mouse_enter(self, x, y):
...
+            self._mirrors.remove(widget)
+        except KeyError:
+            pass
+
+        if not self._mirrors:
+            self.drawer.has_mirrors = False
+
+            if hasattr(self, "_old_draw"):
+                # Deletes the reference to draw and falls back to the original
+                del self.draw
```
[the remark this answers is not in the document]
m-col: Ah my bad we do need it in `_draw_with_mirrors`.

Yeah whole PR LGTM :)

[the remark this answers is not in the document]
elParaguayo: We do need `_old_draw` but I should probably delete the reference to it here as well if there are no mirrors. What do you think?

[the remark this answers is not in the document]
m-col: Yeah, may as well remove it too

m-col:
Yep all looks good, happy to merge :)

PR #3857, opened 2022-09-22
Launchbar render fix

elParaguayo:
The `draw` method calls `clear` multiple times, unnecessarily.

This makes it impossible to render widget decorations correctly.

This PR makes changes to `draw` so that `clear` is only called once.

Fixes: https://github.com/elParaguayo/qtile-extras/issues/117


m-col:
Nice, LGTM

PR #3858, opened 2022-09-23
Add option to `Drag` for warping the cursor at the start of dragging

SpyrosRoum:
This PR adds an extra argument to `Drag` which allows the user to set a point where the cursor should be warped at when the dragging begins.

This allows the user to (for example) warp the cursor to the bottom right of the window when resizing very easily in the config:
```py
# In config.py
@lazy.function
def calc_bottom_right(qtile):
    w = qtile.current_window
    pos = w.cmd_get_position()
    size = w.cmd_get_size()
    x = pos[0] + size[0]
    y = pos[1] + size[1]
    return x, y


# Drag floating layouts.
mouse = [
    Drag(
        [mod], "Button3", lazy.window.set_size_floating(), start=lazy.window.get_size(), warp_pointer=calc_bottom_right()
    ),
]
```

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -757,7 +757,7 @@ class Qtile(CommandObject):
                 else:
                     val = (0, 0)
 
-                if m.warp_cursor:
+                if m.warp_pointer:
                     win_size = self.current_window.cmd_get_size()
                     win_pos = self.current_window.cmd_get_position()
                     x = win_size[0] + win_pos[0]
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -756,6 +756,14 @@ class Qtile(CommandObject):
                         continue
                 else:
                     val = (0, 0)
+
+                if m.warp_cursor:
+                    win_size = self.current_window.cmd_get_size()
+                    win_pos = self.current_window.cmd_get_position()
+                    x = win_size[0] + win_pos[0]
+                    y = win_size[1] + win_pos[1]
+                    self.core.warp_pointer(x, y)
+
                 self._drag = (x, y, val[0], val[1], m.commands)
                 self.core.grab_pointer()
                 handled = True
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -757,7 +757,7 @@ class Qtile(CommandObject):
                 else:
                     val = (0, 0)
 
-                if m.warp_pointer:
+                if m.warp_pointer and self.current_window is not None:
                     win_size = self.current_window.cmd_get_size()
                     win_pos = self.current_window.cmd_get_position()
                     x = win_size[0] + win_pos[0]
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,8 @@ Qtile x.xx.x, released XXXX-XX-XX:
       - The `StockTicker` widget `function` option is being deprecated: rename it to `func`.
     * features
         - Add ability to set icon size in `LaunchBar` widget.
+        - Add 'warp_cursor' option to `Drag` that when set will warp the cursor to the bottom right of
+          the window when dragging begins.
     * bugfixes
 
 Qtile 0.22.0, released 2022-09-22:
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,7 +20,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
       - The `StockTicker` widget `function` option is being deprecated: rename it to `func`.
     * features
         - Add ability to set icon size in `LaunchBar` widget.
-        - Add 'warp_cursor' option to `Drag` that when set will warp the cursor to the bottom right of
+        - Add 'warp_pointer' option to `Drag` that when set will warp the pointer to the bottom right of
           the window when dragging begins.
     * bugfixes
 
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -160,8 +160,8 @@ class Drag(Mouse):
         A list :class:`LazyCall` objects to evaluate in sequence upon drag.
     start:
         A :class:`LazyCall` object to be evaluated when dragging begins. (Optional)
-    warp_cursor:
-        A :class:`bool` indicating if the cursor should be warped to the bottom right of the window
+    warp_pointer:
+        A :class:`bool` indicating if the pointer should be warped to the bottom right of the window
         at the start of dragging. (Default: `False`)
 
     """
@@ -172,11 +172,11 @@ class Drag(Mouse):
         button: str,
         *commands: LazyCall,
         start: LazyCall | None = None,
-        warp_cursor: bool = False,
+        warp_pointer: bool = False,
     ) -> None:
         super().__init__(modifiers, button, *commands)
         self.start = start
-        self.warp_cursor = warp_cursor
+        self.warp_pointer = warp_pointer
 
     def __repr__(self) -> str:
         return "<Drag (%s, %s)>" % (self.modifiers, self.button)
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -160,6 +160,9 @@ class Drag(Mouse):
         A list :class:`LazyCall` objects to evaluate in sequence upon drag.
     start:
         A :class:`LazyCall` object to be evaluated when dragging begins. (Optional)
+    warp_cursor:
+        A :class:`bool` indicating if the cursor should be warped to the bottom right of the window
+        at the start of dragging. (Default: `False`)
 
     """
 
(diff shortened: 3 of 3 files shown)


libqtile/command/base.py
```diff
@@ -207,10 +207,10 @@ def cmd_eval(self, code: str) -> tuple[bool, str | None]:
             error = traceback.format_exc().strip().split("\n")[-1]
             return False, error
 
-    def cmd_function(self, function, *args, **kwargs) -> None:
+    def cmd_function(self, function, *args, **kwargs) -> "Any":
         """Call a function with current object as argument"""
         try:
-            function(self, *args, **kwargs)
+            return function(self, *args, **kwargs)
```
elParaguayo, line 213: I'm interested as to why this is needed. `Drag` has a `start` parameter which takes a lazy function and returns a value so why do we need it for your warp parameter?

libqtile/core/manager.py
```diff
@@ -738,6 +738,16 @@ def process_button_click(self, button_code: int, modmask: int, x: int, y: int) -
...
                     val = (0, 0)
+
+                if m.warp_pointer:
+                    i = m.warp_pointer
+                    status, warp_to = self.server.call((i.selectors, i.name, i.args, i.kwargs))
+                    if status in (interface.ERROR, interface.EXCEPTION):
+                        logger.error("Mouse command error %s: %s", i.name, val)
+                        continue
+                    x, y = warp_to[0], warp_to[1]
+                    self.core.warp_pointer(x, y)
```
elParaguayo, line 749:
```suggestion
                    self.core.warp_pointer(*warp_pointer)
```

elParaguayo (dismissed):

elParaguayo:
You might be better doing this (`lazy.window.function` passes the current window as an argument)

``` python
# In config.py
@lazy.window.function
    pos = window.cmd_get_position()
    size = window.cmd_get_size()
    x = pos[0] + size[0]
    y = pos[1] + size[1]
    return x, y
```

libqtile/core/manager.py
ervinpopescu: This should be

```
self.core.warp_pointer(*warp_to)
``` 

:)

elParaguayo: Oops. Yes!

SpyrosRoum: We need to overshadow `x, y` that come in as method parameters so then `self._drag = (x, y, val[0], val[1], m.commands)` a few lines later has the correct values

libqtile/command/base.py
SpyrosRoum: This took me a bit to figure out so I'll try to explain as best as I can but essentially build-in commands run from a different place, this only affects user defined commands.

All user commands are named `cmd_function`, which is the method I changed, where `function` is the actual user defined function. Meanwhile build-in commands have their own `cmd_<something>` name and they can return whatever they want.

[Here](https://github.com/qtile/qtile/blob/master/libqtile/command/base.py#L147) is the line that chooses the command, `name` is always `function` for user functions which evaluates to the method I changed.

Hope that makes sense

SpyrosRoum:
> You might be better doing this (`lazy.window.function` passes the current window as an argument)
> 
> ```python
> # In config.py
> @lazy.window.function
>     pos = window.cmd_get_position()
>     size = window.cmd_get_size()
>     x = pos[0] + size[0]
>     y = pos[1] + size[1]
>     return x, y
> ```

I didn't know that was a thing, thank you!

libqtile/command/base.py
elParaguayo: I know exactly how they work. I'm just wondering _why_ the change is necessary.

The `start` parameter takes a lazy function but this already works without your change so I'm curious why you need this for your `warp_pointer` argument.

libqtile/core/manager.py
elParaguayo: Cool. That makes sense.

libqtile/command/base.py
SpyrosRoum: It's not specific to `warp_pointer`, it's just so it enables user defined functions to return things.
If you pass the same function as I had in the example to `start` it won't work either without my chanbge

elParaguayo:
> > You might be better doing this (`lazy.window.function` passes the current window as an argument)
> > ```python
> > # In config.py
> > @lazy.window.function
> >     pos = window.cmd_get_position()
> >     size = window.cmd_get_size()
> >     x = pos[0] + size[0]
> >     y = pos[1] + size[1]
> >     return x, y
> > ```
> 
> I didn't know that was a thing, thank you!

Hmm. My comment list the function definition line. You need the same definition that you had but it takes `window` rather than `qtile` as an argument.

libqtile/command/base.py
elParaguayo: But `start` _does_ work without your change from what I can see.

I will test this later though as I may be missing something.

SpyrosRoum: Hmm are you sure? I get `TypeError: 'NoneType' object is not subscriptable` in the logs if I give it my function and remove that `return` from `cmd_function`

elParaguayo: That's why I'm confused.

What I'm trying to say is that, in the default config, Drag has a `start` parameter that takes a lazy function which returns a value. That already works without your change.

SpyrosRoum: I was stuck on this too, the difference is that build-in lazy functions get evaluated differently from user-defined functions.

User functions go through `cmd_function` which executes the function but doesn't return anything. Did you try passing a custom function to `start`?

elParaguayo: Nice. That makes sense. There's no reason not to return the value other than it's not needed 99% of the time!

m-col:
Is it sufficient to warp within the `start` function itself?
```py
@lazy.function
def start(qtile):
    w = qtile.current_window
    pos = w.cmd_get_position()
    size = w.cmd_get_size()
    x = pos[0] + size[0]
    y = pos[1] + size[1]
    qtile.core.warp_pointer(x, y)
    return w.cmd_get_size()


# Drag floating layouts.
mouse = [
    Drag(
        [mod], "Button3", lazy.window.set_size_floating(), start=start,
    ),
]
```
Of course this would also require that `lazy.function` returns the return values of the decorated function if it doesn't currently.

elParaguayo:
We don't need to change `start`. It works because it's a built-in method that already returns a value. The issue is using custom functions as `lazy.function` doesn't return the result of the function.

I think it's ok for us to change it so the result is returned.

SpyrosRoum:
> Is it sufficient to warp within the `start` function itself?
> 
> ```python
> @lazy.function
> def start(qtile):
>     w = qtile.current_window
>     pos = w.cmd_get_position()
>     size = w.cmd_get_size()
>     x = pos[0] + size[0]
>     y = pos[1] + size[1]
>     qtile.core.warp_pointer(x, y)
>     return w.cmd_get_size()
> 
> 
> # Drag floating layouts.
> mouse = [
>     Drag(
>         [mod], "Button3", lazy.window.set_size_floating(), start=start,
>     ),
> ]
> ```
> 
> Of course this would also require that `lazy.function` returns the return values of the decorated function if it doesn't currently.

I hadn't thought about that. It *almost* works.
The reason it doesn't work is because inside of `process_button_click` we will need to always update the cursor position if there is a `start` function in case the user warped it.

So if there is a way to query for the current position of the cursor then yes, this will work without the need for the extra parameter

m-col:
Is your intended usage only to warp the cursor to the bottom right corner of the window so that resizing is less weird? If so, I'd be in favour of having that specific functionality baked in (e.g. `warp=True`), because that feature has been discussed before (https://github.com/qtile/qtile/pull/1788) which suggests it isn't quite a niche feature. Whereas warping to other positions likely would be.

SpyrosRoum:
This is my intended use, but I tried to make it general enough. Tbh I don't see any other possible uses right now.

m-col:
Cool - I also don't see any other use cases. I think in this case being less general is better because: 1) more flexibility means more chances of bugs (both in internal code and user config code) and 2) for the desired functionality all users would need to duplicate your `calc_bottom_right` function (more work for them, also more chance of bugs).

If you'd be happy with simply hard-coding warping to the bottom-right as in https://github.com/qtile/qtile/pull/1788, then that would be great :)

SpyrosRoum:
I left in the commit that enables user defined lazy functions to return things even though it's not needed after all. Should I remove it?

m-col:
Yeah, may as well remove that part. Could you add the new feature to the changelog? This way people will be able to use your contribution :)

m-col:
That looks perfect, thanks again for the submission :) I've started the CI.

We may hold off on merging for a few days so that we can get https://github.com/qtile/qtile/pull/2727 merged first. It's a big PR that touches almost every file, and in the past we ended up in a situation where rebasing it on top of updates to master became difficult, so we're hoping to merge that before more changes to master are made.

m-col:
Thanks again for this, and thanks for rebasing. I've enabled the CI and it will auto-merge if everything passes :)

Issue #3718, opened 2022-07-24
[Wayland] Ardour issues


justinesmithies:
### The issue:

I'm running the latest Qtile git master under Wayland. When I run Ardour if i go to the Session menu and click Import or use the shortcut Ctrl + I it opens the import window no problem. But if I close that window and try to reopen it either from the menu or shortcut then no window ever appears again. You then have to quit Ardour and go back in again.
I tested this with sway and plasma and it works perfectly every time.
I don't see anything relevant in my log files either ??

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
I believe this is fixed now but please re-open if not.

Issue #3678, opened 2022-07-04
XWayland: Android Studio context menus flash for a split second and disappear


arjan-s:
### The issue:

Context menus (and for instance the file selection menu opened by double shift press) appear and immediately disappear again.

Qtile version: current master

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

TinyAtoms:
I want to add that this is also present in clion and pycharm in the current master version of qtile.

m-col:
Related sway issues: https://github.com/swaywm/sway/issues/4370, https://github.com/swaywm/sway/issues/5414, https://github.com/swaywm/sway/issues/3007

m-col:
I believe this is fixed now but please re-open if not.

Issue #3847, opened 2022-09-20
Xwayland in Gnome supports more resolutions.


m-col:
Xwayland in Gnome supports more resolutions.

In gnome I use the monitor at 4k scaled at 200% and in Qtile using wlr-randr.

If I write to a console in Gnome the xrandr command, it gives me the following output
```Screen 0: minimum 16 x 16, current 3840 x 2160, maximum 32767 x 32767
XWAYLAND0 connected primary 3840x2160+0+0 (normal left inverted right x axis y axis) 620mm x 340mm
   3840x2160     59.98*+
   2048x1536     59.95  
   1920x1440     59.90  
   1600x1200     59.87  
   1440x1080     59.99  
   1400x1050     59.98  
   1280x1024     59.89  
   1280x960      59.94  
   1152x864      59.96  
   1024x768      59.92  
   800x600       59.86  
   640x480       59.38  
   320x240       59.52  
   2560x1600     59.94  
   1920x1200     59.88  
   1680x1050     59.95  
   1440x900      59.89  
   1280x800      59.81  
   720x480       59.71  
   640x400       59.95  
   320x200       58.96  
   3200x1800     59.96  
   2880x1620     59.96  
   2560x1440     59.96  
   2048x1152     59.90  
   1920x1080     59.96  
   1600x900      59.95  
   1368x768      59.88  
   1280x720      59.86  
   1024x576      59.90  
   864x486       59.92  
   720x400       59.55  
   640x350       59.77 
 ```

and if I write it to Qtile, only this

```Screen 0: minimum 16 x 16, current 1920 x 1080, maximum 32767 x 32767
XWAYLAND0 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 620mm x 340mm
   1920x1080     59.96*+
```
This makes steam games in qtile the maximum resolution is 1920x1080, while in gnome games can use resolutions up to 4k (my monitor is 4k), could something be done so that this can be implemented in qtile? Thank you

_Originally posted by @cr1ogen in https://github.com/qtile/qtile/discussions/2409#discussioncomment-3676479_

m-col:
@cr1ogen would you be able to test on sway to see if they are reporting all expected output modes?

cr1ogen:
> @cr1ogen would you be able to test on sway to see if they are reporting all expected output modes?

Yes.  When I get home I'll try it.

cr1ogen:
[...]

Issue #3831, opened 2022-09-13
[ Wayland ] Ardour import window issue


justinesmithies:
### The issue:

I'm running the latest Qtile git master .21.1.dev106+g5161618c under Wayland 
When I run Ardour and open the Import menu all is fine until i close it with my window kill keybind mod + w
The window disapears but can no longer be reopened from the menu or the ctrl + I key.
I noticed though using  qtile cmd-obj -o cmd -f windows that it did show the floating window to start with but once killed the output of  qtile cmd-obj -o cmd -f windows looked wierd and seems to show an invalid window id still there. So maybe that's stopping the menu reopening ???
You can see the difference if you run up ardour and open the import menu then run  qtile cmd-obj -o cmd -f windows
Next close that window and run  qtile cmd-obj -o cmd -f windows again. You'll see a remaining id that doesn't look right.

Also one other Ardour related issue. When it is first run and the new session window appears once you click on new session the next menu will not appear until you move the cursor out of that window.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
I believe this is fixed by https://github.com/qtile/qtile/pull/3862, please re-open if not

m-col:
Just saw your messages on IRC that this isn't fixed. Copying here to track:
```
JustineSmithies: mcol: That wl_pointer fix ( misc wayland ) partially works for Ardour.
  When it opens i still have to click new session more than once to get the next window of options.
JustineSmithies: Also once in i can ctrl + I to get the import window up which works .
  Then close it with mod + w and open it again. But after that I cannot close it as i get quit Ardour
  instead then that window vanishes behind the main once and i cannot move it at all or close it
  yet the options in that window still work.
```

PR #3862, opened 2022-09-24
Misc wayland window fixes

m-col:
This makes the wayland core track which window of any type is under the pointer, rather than just internal windows. This lets us detect when the pointer moves out of a Window/Static window, so we can reset the cursor image (to the default 'left_ptr' icon) when that happens. Previously the icon was set upon all cursor movements if no non-internal window was under the pointer, doing needless work.

This also prevents windows that don't have pointer focus from changing the icon (firefox does this), and improves handling of focus and icon when transitioning between windows and borders.

Editted - added a few more commits that improve windows

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -41,6 +41,7 @@ from wlroots.wlr_types import (
     GammaControlManagerV1,
     InputInhibitManager,
     OutputLayout,
+    PointerGesturesV1,
     PrimarySelectionV1DeviceManager,
     RelativePointerManagerV1,
     ScreencopyManagerV1,
@@ -175,12 +176,21 @@ class Core(base.Core, wlrq.HasListeners):
         # set up cursor
         self.cursor = Cursor(self.output_layout)
         self.cursor_manager = XCursorManager(24)
+        self._gestures = PointerGesturesV1(self.display)
         self.add_listener(self.seat.request_set_cursor_event, self._on_request_cursor)
         self.add_listener(self.cursor.axis_event, self._on_cursor_axis)
         self.add_listener(self.cursor.frame_event, self._on_cursor_frame)
         self.add_listener(self.cursor.button_event, self._on_cursor_button)
         self.add_listener(self.cursor.motion_event, self._on_cursor_motion)
         self.add_listener(self.cursor.motion_absolute_event, self._on_cursor_motion_absolute)
+        self.add_listener(self.cursor.pinch_begin, self._on_cursor_pinch_begin)
+        self.add_listener(self.cursor.pinch_update, self._on_cursor_pinch_update)
+        self.add_listener(self.cursor.pinch_end, self._on_cursor_pinch_end)
+        self.add_listener(self.cursor.swipe_begin, self._on_cursor_swipe_begin)
+        self.add_listener(self.cursor.swipe_update, self._on_cursor_swipe_update)
+        self.add_listener(self.cursor.swipe_end, self._on_cursor_swipe_end)
+        self.add_listener(self.cursor.hold_begin, self._on_cursor_hold_begin)
+        self.add_listener(self.cursor.hold_end, self._on_cursor_hold_end)
         self._cursor_state = wlrq.CursorState()
 
         # set up shell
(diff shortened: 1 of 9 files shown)


m-col:
Fixes #3718, #3831

m-col:
This PR is ready for review. It grew arms and legs a bit and now it covers a number of wayland issues that touched the same code so this was easier for working on things simultaneously instead of having to merge everything after :)

Also fixes https://github.com/qtile/qtile/issues/3678 and https://github.com/qtile/qtile/issues/3697.

IchikaZou:
Will these fixes get merge into next version of qtile?

m-col:
Yes should be! :)

cr1ogen:
#3847 any notice? sorry xD

libqtile/backend/wayland/core.py
```diff
@@ -672,44 +761,69 @@ def _process_cursor_motion(self, time_msec: int, cx: float, cy: float) -> None:
...
+                        if isinstance(self._hovered_window, window.Internal):
+                            if motion is not None:
+                                # moved from an Internal to a different Internal
+                                self._hovered_window.process_pointer_leave(
+                                    cx - self._hovered_window.x,
+                                    cy - self._hovered_window.y,
+                                )
+                        else:
+                            # moved from a Window or Static to an Internal
+                            if self.seat.pointer_state.focused_surface:
```
elParaguayo, line 793: Could this be a single `elif`?

libqtile/backend/wayland/core.py
```diff
@@ -981,8 +1104,16 @@ def _under_pointer(self) -> tuple[window.WindowType, Surface | None, float, floa
...
-    def stack_windows(self) -> None:
-        """Put all windows of all types in a Z-ordered list."""
+    def stack_windows(self, restack: window.WindowType | None = None) -> None:
+        """
+        Put all windows of all types in a Z-ordered list.
+
+        A (non-layer_shell) window passed as 'restack' will bubble up the Z order.
+        """
+        if restack:
+            self.mapped_windows.remove(restack)
```
elParaguayo, line 1114: Will `restack` always be in `self.mapped_windows`? There's no error handling here.

test/backend/wayland/conftest.py
```diff
@@ -27,6 +31,22 @@ def wayland_environment(outputs):
     yield env
 
 
+@pytest.fixture(scope="function")
+def wmanager(request, wayland_session):
+    """
+    This replicates the `manager` fixture except that the Wayland backend is hard-coded.
+    We cannot parametrize the `backend_name` fixture module-wide because it gets
+    parametrized by `pytest_generate_tests` in test/conftest.py and only one of these
+    parametrize calls can be used.
```
elParaguayo, line 40: Not quite following why this is needed. Can you explain for me?

elParaguayo (commented):
Some minor comments questions but, overall, code looks really nice.

libqtile/backend/wayland/core.py
```diff
@@ -240,6 +253,10 @@ def __init__(self) -> None:
         # Start
         self.backend.start()
 
+        # Place cursor
+        self.warp_pointer(0, 0)
```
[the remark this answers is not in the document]
m-col: Good idea - just added.

libqtile/backend/wayland/core.py
m-col: nice catch, yep

libqtile/backend/wayland/core.py
m-col: Yes it always will be. It's not particularly pretty code, this stuff with `mapped_windows`. I'm hoping to rewrite some of this in the not so distant future (using some nice helpers from wlroots called wlr_scene that should handle this for us to some extent)

test/backend/wayland/conftest.py
m-col: Ha yes, this is an annoying quirk of pytest. You'll see the same thing in `test/backend/x11/conftest.py` with an `xmanager` fixture. There is a good reason for it, to do with parametrising pytest via CLI (`--backend=X`), but I'm struggling to identify what the reason was. Let me look into this.

m-col: If you make them all use `manager` then the x11 tests get attempted using the wayland backend, and vice-versa, via the `backend` fixture used by `manager`. Removing `xmanager` and `wmanager` and using `manager` causes this:
```
test/backend/test_backend.py::test_get_core_bad PASSED                                                 [  1%]
test/backend/wayland/test_window.py::test_info[1-wayland-BareConfig] PASSED                            [  3%]
test/backend/wayland/test_window.py::test_info[1-x11-BareConfig] FAILED                                [  5%]
```
Note the failed test is a wayland test parameterised by the x11 fixtures.

elParaguayo: Ok. Can't we mark the tests to skip in those scenarios?

m-col: The only way (i think) that we can do that is something like this:
```py
@bare_config
def test_info(backend_name, manager):
    if backend_name != "wayland":
        raise pytest.skip()
    ...
```
We might be able to use some kind of decorator to automate this, perhaps, but it would need adding to all tests (perhaps not that difficult given they'll fail otherwise). However an undesirable side effect is that the output of pytest logs SKIPPED messages:
```
test/backend/test_backend.py::test_get_core_bad PASSED                                                                 [  3%]
test/backend/wayland/test_window.py::test_info[1-wayland-BareConfig] PASSED                                            [  6%]
test/backend/wayland/test_window.py::test_info[1-x11-BareConfig] SKIPPED                                               [  9%]
```
The current solution avoids this, but it does seem hacky and a bit strange (and is ultimately duplicate code that we ought to avoid). If you had any other ideas for how to handle this please do let me know!

elParaguayo: Stick with the current approach for now!

PR #3863, opened 2022-09-24
Fix Bluetooth widget name bug

### LINKED ISSUE #3816: Bluetooth widget displays adapter name instead of name of connected device
### The issue:

version: 0.21.0
log: no relevant log

I configured the bluetooth-widget.
When a device is connected, it shows the adapter name, instead of the device name.

### Required:

elParaguayo:
Fixes an edge case where, if the bluetooth adapter emits a PropertiesChanged signal with the name of the adapter, the widget displays the adapter name rather than the connected device name. This is achieved by having separate handlers for the device and adapter signals.

Fixes #3816

### DIFF
--- a/libqtile/widget/bluetooth.py
+++ b/libqtile/widget/bluetooth.py
@@ -72,7 +72,7 @@ class Bluetooth(base._TextBox):
 
         powered = await iface.get_powered()
         # subscribe receiver to property changed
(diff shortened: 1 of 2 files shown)


m-col (approved):
Nice, LGTM :)

PR #3870, opened 2022-10-01
Move from pytest-cov to coverage

elParaguayo:
pytest-cov 4.0.0 dropped support for `multiprocessing` which massively impacts our coverage scores. Pinning to 3.0.0 would retain that support but is not a good long term solution.

pytest-cov recommends moving to coverage but I'm currently seeing a few drops in coverage on a small number of tests. Needs debugging.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,8 @@ whitelist_externals=convert
 # xcffib has to be installed before cairocffi
 deps =
     pytest >= 6.2.1
-    pytest-cov <= 3.0.0
+    ; pytest-cov <= 3.0.0
+    coverage
     setuptools >= 40.5.0
     bowler
     xkbcommon >= 0.3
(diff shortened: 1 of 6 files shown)


elParaguayo:
Asked for some support in debugging the differences here: https://github.com/nedbat/coveragepy/issues/1458

tox.ini
```diff
@@ -48,9 +48,16 @@ commands =
     {toxinidir}/scripts/ffibuild
 # py310 currently fails with -W error, see: https://gitlab.gnome.org/GNOME/pygobject/-/issues/476
```
m-col, line 49: This gnome issue has been resolved now so we no longer need a separate `py310` run. Might be worth updating that while we're here?

m-col (approved):

tox.ini
elParaguayo: Still got an error.

elParaguayo:
Coverage actually looks ok.

elParaguayo:
Most of the tests that have reduced coverage are ones that inherit `base.ThreadPoolText`. It looks like the `poll` method is no longer being covered in the `init_configure` test, perhaps because this is called by `qtile.run_in_executor`.

PR #3872, opened 2022-10-03
Pin pytest-cov to 3.0.0

elParaguayo:
pytest-cov 4.0.0 dropped multiprocessing support which makes all our `manager` tests report as uncovered.

This PR pins pytest-cov to 3.0.0. This is a temporary fix while we look for a better solution in #3870.


m-col:
LGTM

PR #3884, opened 2022-10-08
Border color for stacked windows in stack layout

Jack8472:
This adds a new feature in **stack layout**. Now it is possible to change border color if the windows are in 'split' or 'stacked' state. This feature is present in column layout and is very useful to quickly see if there are windows "hiding" below. The feature will be as useful in stack layout.

### DIFF
--- a/libqtile/layout/stack.py
+++ b/libqtile/layout/stack.py
@@ -55,6 +55,10 @@ class Stack(Layout):
     defaults = [
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
+        ("border_focus_stack", "#881111",
+         "Border colour(s) for the focused stacked window."),
+        ("border_normal_stack", "#220000",
+         "Border colour(s) for un-focused stacked windows."),
         ("border_width", 1, "Border width."),
         ("autosplit", False, "Auto split all new stacks."),
         ("num_stacks", 2, "Number of stacks."),
@@ -206,9 +210,9 @@ class Stack(Layout):
             return
 
         if client.has_focus:
-            px = self.border_focus
+            px = self.border_focus if s.split else self.border_focus_stack
         else:
-            px = self.border_normal
+            px = self.border_normal if s.split else self.border_normal_stack
 
         column_width = int(screen_rect.width / len(self.stacks))
         xoffset = screen_rect.x + i * column_width
--- a/libqtile/layout/stack.py
+++ b/libqtile/layout/stack.py
@@ -58,9 +58,11 @@ class Stack(Layout):
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
         ("border_focus_stack", None,
-         "Border colour(s) for the focused stacked window. If 'None' will default to border_focus."),
+         "Border colour(s) for the focused stacked window. If 'None' will \
+         default to border_focus."),
         ("border_normal_stack", None,
-         "Border colour(s) for un-focused stacked windows. If 'None' will default to border_normal."),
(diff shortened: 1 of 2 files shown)


elParaguayo:
Can you post a screenshot to explain what this does?

Also, when adding new options, we should try not to change the default behaviour/appearance i.e. users who upgrade will not expect to see something different.

Jack8472:
I made three screenshots with extra wide borders which hopefully convey the behavior of the functionality. This behavior is already present in the column layout so maybe it is not a big change in behavior.

Screenshot-1-windows-are-split png
Screenshot-2-windows-are-stacked-focused
Screenshot-3-windows-are-stacked-unfocused png

Jack8472:
Similar functionality in columns layout I was refering to above is in [columns.py](https://github.com/qtile/qtile/blob/master/libqtile/layout/columns.py):

Lines 121-130:

```
        (
            "border_focus_stack",
            "#881111",
            "Border colour(s) for the focused window in stacked columns.",
        ),
        (
            "border_normal_stack",
            "#220000",
            "Border colour(s) for un-focused windows in stacked columns.",
        ),
```

Lines 250-253:

```
        if client.has_focus:
            color = self.border_focus if col.split else self.border_focus_stack
        else:
            color = self.border_normal if col.split else self.border_normal_stack
```

libqtile/layout/stack.py
```diff
@@ -55,6 +55,10 @@ class Stack(Layout):
     defaults = [
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
+        ("border_focus_stack", "#881111",
+         "Border colour(s) for the focused stacked window."),
+        ("border_normal_stack", "#220000",
+         "Border colour(s) for un-focused stacked windows."),
```
m-col, line 61: Could you make the default values `None`? That way the feature is opt-in and users won't have things suddenly change when they update.

Then below when the colors are accessed it could test for `None`s.

m-col (changes requested):
Neat! I've made a small comment inline but otherwise it's a great feature, thanks for the submission!

libqtile/layout/stack.py
Jack8472: I tried to do what you suggested. Please let me know if I was successful.

m-col (commented):
Thanks!

I'm also wondering about the docstrings for the options: "Border colour(s) for the focused stacked window." It might be worth making it clear that if not provided, it will fall back to the `border_focus`?

Also - if you add an entry to the changelog, users will learn about this option and can use it :)

libqtile/layout/stack.py
```diff
@@ -206,9 +210,9 @@ def configure(self, client, screen_rect):
             return
 
         if client.has_focus:
-            px = self.border_focus
+            px = self.border_focus if s.split else self.border_focus_stack
```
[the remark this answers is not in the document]
Jack8472: Thank you. Did my best to follow the suggestions.

Jack8472:
Had to do another committ to fix a bug (colors were reversed). Now the thing works as intended.

m-col (approved):
Perfect :) I've triggered the CI and will enable auto-merge so that if all the tests pass it will get merged.

Thanks for the PR!

m-col:
Not very clear error message:
```
  libqtile/layout/stack.py:60:10: BLK100 Black would make changes.
```
I suspect it doesn't like the way the multi-line doc strings are formatted in some way. Might be worth running `black` on the libqtile folder and let it figure it out?

Jack8472:
> I suspect it doesn't like the way the multi-line doc strings are formatted in some way. Might be worth running `black` on the libqtile folder and let it figure it out?

Yes. Black found bad formatting on my strings. Sorry for that. Committed a correction.

m-col:
Fantastic, thanks again!

Jack8472:
> Fantastic, thanks again!

Thank YOU very much!

PR #3888, opened 2022-10-10
Fix typing for config.Group.layouts

elParaguayo:
libqtile.config.Group.layouts should take a list of layout instances (unlike the `layout` parameter which takes the name of a layout).


m-col:
Good catch!

elParaguayo:
> Good catch!

I take no credit - it was spotted by a user on Discord who was getting confused by some error messages.

PR #3890, opened 2022-10-11
Add spotify widget 

ThanosApollo:
### It ain't much but it's honest work, hope I didn't mess anything up
- Created a spotify widget which I've been using the past week, *inspired by the moc widget*
- I tried to make it using dbus_next *but* I was only able to create bugs, so i settled with dbus-python, hope its good enough
- "Buttons" are using **spotify-control** [AUR](https://aur.archlinux.org/packages/spotify-control) [Github](https://github.com/J00LZ/spotify-control)
  - I was already using it so I thought it was good enough, and there was no need to recreate play/pause etc. functions 
    - +Easy to install via the AUR 
    - +Less code to maintain
- I'm new here so any comments/fixes are welcomed.

### DIFF
--- a/libqtile/widget/__init__.py
+++ b/libqtile/widget/__init__.py
@@ -77,6 +77,7 @@ widgets = {
     "Sep": "sep",
     "She": "she",
     "Spacer": "spacer",
+    "Spotify": "spotify",
     "StatusNotifier": "statusnotifier",
     "StockTicker": "stock_ticker",
     "SwapGraph": "graph",
--- a/libqtile/widget/spotify.py
+++ b/libqtile/widget/spotify.py
@@ -1,0 +1,70 @@
+# Copyright (C) 2022, ThanosApollo
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+import dbus
+import subprocess
+from functools import partial
+from libqtile.widget import base
+
+class Spotify(base.ThreadPoolText):
+    """A simple Spotify widget, made using dbus-python.
+    Show the song and artist of now listening song and allow basic mouse
+    control from the bar using spotify-control(
+    Github: https://github.com/J00LZ/spotify-control
+    AUR:    https://aur.archlinux.org/packages/spotify-control
+    ):
+    - toggle pause (or play if stopped) on left click;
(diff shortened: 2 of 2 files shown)


elParaguayo:
Hi there. Thanks for wanting to contribute to qtile. 

What does this add over the Mpris2 widget? I think that can do everything that this widget seems to do.

ThanosApollo:
> Hi there. Thanks for wanting to contribute to qtile.
> 
> What does this add over the Mpris2 widget? I think that can do everything that this widget seems to do.

Hello! 
- I was using Mpris2 widget but if I'd open a second media player(with spotify paused) such as vlc it would crush or give "No metadata for current track"(I guess it was picking metadata from vlc, where i usually watch movies or stream youtube videos).
- I just wanted a widget that picks metadata **only** from spotify, so it will be easy to jump on and off listening to music. Without my widget crushing or pick metadata from other programs/apps. Similar functionality to moc widget, but for spotify.

elParaguayo:
You can do that with the Mpris2 widget. Just set `objname="org.mpris.MediaPlayer2.spotify"` when configuring the widget and it will only listen to Spotify messages. If that doesn't work then it's a bug that should be fixed!

ThanosApollo:
- I tried to recreate some errors and post screenshots [here](https://github.com/ThanosApollo/dotfiles/tree/master/Pictures/Screenshots) I will delete them tomorrow tho
  - Not picking metadata etc. Hope it helps
- One easy to recreate annoying bug is that it "crushes" if you reload qtile while using spotify, 
- Spotify is quite popular, so having a dedicated widget for it wouldn't be a bad idea, since we have for other music playerst. 
- Mpris2 is an amazing widget, but it's frustrating when something goes wrong that 5% of the time.
Anyways, there is not much else to it. Thanks for looking at my pull request that fast tho xD!

ThanosApollo:
Open spotify, reload qtile, the widget disappears

elParaguayo:
> Open spotify, reload qtile, the widget disappears

Ah ok. So it doesn't query a player when it first starts. Presumably if you change track the text appears?

elParaguayo:
I've got a fix for the lost metadata on reload.

As for your screenshots, this widget is relying on signals being broadcast over dbus. The advantage of this was that we're not constantly polling the player. However, if these are missing then we get those messages that you see. We could add a secondary polling method that runs on a larger interval so that it can tidy up those issues but I wouldn't have it as the primary method of getting metadata.

Thoughts?

ThanosApollo:
I like it, I'd also suggest that it displays somekind of message when it's "missing" metadata signals, such as "No metadata to display" at least,  + a straight forward explanation on the docs on how to configure Mpris2 with spotify, if someone is not familiar with dbus he probably will have a hard time understanding the docs, some basic examples would be a nice addition.
Personally, I will be using the custom widget I've made cause it's simpler and more straight forward with dbus-python and I don't get the errors I do with mpris2.

elParaguayo:
I think there's a general theme about making the docs more user friendly as they do assume a certain amount of knowledge which can make configuration harder for newer users.

As for using your own widget - makes perfect sense. One of the things that I really like about qtile is how easy it is to make your own widgets etc to fit your own needs.

ThanosApollo:
Yeah, you and the rest of the team have done an amazing job!

PR #3891, opened 2022-10-11
Use logger.error in loop exception handler

elParaguayo:
Using logger.exception results in `NoneType: None` messages being written to the log. This is because the method is not called from an exception handler. Using logger.error removes this issue.


m-col:
Nice!

PR #3892, opened 2022-10-11
Two StatusNotifier fixes

elParaguayo:
These two fixes are needed to close the issue posted on qtile-extras: https://github.com/elParaguayo/qtile-extras/issues/130
* Fixes an edge case where, when a user switches from `libqtile.widget.StatusNotifier` to `qtile_extras.widget.StatusNotifier` via a `reload_config` call, menus are not available
* Some apps don't have an `Activate` method exposed on dbus resulting in an error message when the icon is clicked.

Closes: https://github.com/elParaguayo/qtile-extras/issues/130

### DIFF
--- a/libqtile/widget/statusnotifier.py
+++ b/libqtile/widget/statusnotifier.py
@@ -326,7 +326,8 @@ class StatusNotifierItem:  # noqa: E303
         return icon
 
     def activate(self):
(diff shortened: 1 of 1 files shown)


m-col (approved):
Nice thanks

PR #3900, opened 2022-10-15
Add `desc` attribute to `KeyChord`

elParaguayo:
Users who create scripts to show configured keybindings will view the `desc` attribute of `Key` objects but `KeyChord` objects do not have this attribute.


m-col:
Nice, LGTM.

PR #3901, opened 2022-10-15
Add ability to disabling toggling in GroupBox

elParaguayo:
Adds a new parameter to disabling toggling of group when clicking on the same group name.


m-col:
LGTM. Looks like black would make changes:
```
  libqtile/widget/groupbox.py:240:86: BLK100 Black would make changes.
```

PR #3903, opened 2022-10-16
Update Match typing to include re.Pattern

m-col:
Some `Match` properties support `str | re.Pattern | None` so let's type accordingly. Also updates the docstring to clarify this a bit.


elParaguayo:
LGTM

PR #3914, opened 2022-10-20
Fix Window.center() bug

### LINKED ISSUE #3912: `centre` doesn't centre window on correct screen.
Separate issue created from https://github.com/qtile/qtile/issues/2163#issuecomment-1218821651_:

> However, it has another issue - `cmd_center` doesn't seem to work correctly for my second monitor (the window always becomes centered on my left monitor). This is on X11. I can file a separate issue for this if it's not the intended behavior (unless I made a mistake in my code).

elParaguayo:
Window.center() did not take account of screen offsets so windows would appear on the wrong screen when using multi-monitor setups.

Fixes #3912

### DIFF
--- a/test/test_window.py
+++ b/test/test_window.py
@@ -1,6 +1,6 @@
 import pytest
 
-from test.conftest import BareConfig
+from test.conftest import BareConfig, dualmonitor
 from test.layouts.layout_utils import assert_focused
 from test.test_manager import ManagerConfig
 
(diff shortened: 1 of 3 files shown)


test/test_window.py
```diff
@@ -1,12 +1,48 @@
...
+            x=0,
+            y=300,
+            width=800,
+            height=300,
+        ),
+    ]
+    screens = []
+
+
+fakescreen_config = pytest.mark.parametrize("manager", [FakeScreenConfig], indirect=True)
```
m-col, line 43: You could use this to avoid creating a new config:
```
from test.conftest import dualmonitor
```

m-col (approved):
LGTM

PR #3916, opened 2022-10-21
Remove confusion in Key documentation

elParaguayo:
Previously, the `commands` argument said users could provide a list of commands. This is incorrect and providing a list will result in an error.


m-col (approved):
LGTM

PR #3927, opened 2022-10-25
Simplify running lazy command from config

elParaguayo:
We often see users trying to run lazy commands from within a function in their config. This is because there is some confusion about what lazy commands actually do. In addition, while users are familiar with the lazy command syntax, they are less familiar with the underlying python objects.

This PR addresses these issues by adding a `qtile.run_lazy` method which allows users to pass a lazy command as an argument and this will be run.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -61,6 +61,7 @@ if TYPE_CHECKING:
     from libqtile.command.base import ItemT
     from libqtile.confreader import Config
     from libqtile.layout.base import Layout
+    from libqtile.lazy import LazyCall
 
 
 class Qtile(CommandObject):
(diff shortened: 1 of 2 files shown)


m-col:
With this, wouldn't we just be replacing "don't use lazy commands directly, access the commands on the object itself" with "don't use lazy commands directly, pass them to qtlle.run_lazy()"? I'm not sure this is a huge improvement. If it's a documentation issue, as part of https://github.com/qtile/qtile/pull/3540 we could provide code snippets for each object showing how they can be accessed directly by traversing the graph by e.g. `qtile.current_window`, `qtile.current_screen` etc

elParaguayo:
I'm not wedded to this. It's mainly because we see so many examples of people using lazy calls within methods. 

Happy to add better documentation instead.

elParaguayo:
The only other thing that this code does do is remove the duplication of the server call in `manager.py` and `widget.__init__.py`, so the code is a bit neater :wink:

m-col:
Well you've got me there... I suppose this _and_ documentation would be a double whammy that might doubly help?

PR #3928, opened 2022-10-25
WL: correctly check failed xwayland startup

m-col:
If XWayland can't be started up it raises a RuntimeError, it doesn't return None, so this needs to be handled with a catch.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -228,14 +228,16 @@ class Core(base.Core, wlrq.HasListeners):
         self.foreign_toplevel_manager_v1 = ForeignToplevelManagerV1.create(self.display)
 
         # Set up XWayland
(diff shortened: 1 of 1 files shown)


inmaldrerah:
This solution seems to work like a charm. It does work without most xorg support (except xorg-setxkbmap). So does this mean that XWayland should no longer be a required dependency for building/installing Qtile (once this PR is merged)? (Has it ever been designed to be one?)

m-col:
Nice, thanks for testing this. In theory yes this would allow one to run wayland without any X dependencies, which is definitely something we're aiming for. If I remember correctly, Qtile always needs xcffib around as it imports it (even though under wayland it isn't used), but this is something I'm working on ATM.

inmaldrerah:
Could you please also update the AUR repo qtile-wayland-git such that xorg-xwayland would become an optional dependency? (You are the AUR package maintainer, right?)

m-col:
Yes and yes, done :)

PR #3930, opened 2022-10-25
Add latest python and pypy versions to CI

elParaguayo:
I may have got the tox syntax wrong...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
                 # - /libqtile/scripts/check.py mypy argument
                 # If adding new python versions, consider also updating
                 # python version in .readthedocs.yaml
-                python-version: [pypy-3.7, 3.7, 3.8, 3.9, '3.10']
+                python-version: [pypy-3.9, 3.8, 3.9, '3.10', '3.11']
         steps:
             - uses: actions/checkout@v2
             - name: Set up python ${{ matrix.python-version }}
@@ -86,8 +86,8 @@ jobs:
                 [ "$(grep -c -P '\t' CHANGELOG)" = "0" ]
                 tox
             - name: Upload coverage data to coveralls.io
-              # For speed purposes, we don't do coverage reporting on pypy 3.7
-              if: ${{ matrix.python-version != 'pypy-3.7' }}
+              # For speed purposes, we don't do coverage reporting on pypy 3.9
+              if: ${{ matrix.python-version != 'pypy-3.9' }}
               run: |
                 pip -q install coveralls
                 coveralls --service=github
--- a/tox.ini
+++ b/tox.ini
@@ -4,10 +4,10 @@ skipsdist=True
 minversion = 1.8
 envlist =
     pypy3,
-    py37,
     py38,
     py39,
     py310,
+    py311,
     docs,
     pep8,
     codestyle,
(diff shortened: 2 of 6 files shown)


m-col:
Should we also drop 3.8 (and pypy-3.8)?

elParaguayo:
This looks ok to merge now. We just need to update the branch protection rules after this to reflect the new version (current PRs would need to be rebased too).

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
                 # - /libqtile/scripts/check.py mypy argument
                 # If adding new python versions, consider also updating
                 # python version in .readthedocs.yaml
-                python-version: [pypy-3.7, 3.7, 3.8, 3.9, '3.10']
+                python-version: [pypy-3.9, 3.8, 3.9, '3.10', '3.11']
```
m-col, line 25: Per the comment, should we bump the version in .readthedocs.yaml to 3.11 too?

tox.ini
```diff
@@ -48,10 +48,15 @@ commands =
     {toxinidir}/scripts/ffibuild
     # py310 currently fails with -W error, see: https://gitlab.gnome.org/GNOME/pygobject/-/issues/476
     # pypy3 is very slow when running coverage reports so we skip it (coverage is reported on py37 to py310)
```
m-col, line 50:
```suggestion
    # pypy3 is very slow when running coverage reports so we skip it
```

tox.ini
```diff
@@ -129,8 +134,8 @@ commands =
 
 [gh-actions]
 python =
-    pypy-3.7: pypy3
-    3.7: py37
+    pypy-3.9: pypy3
     3.8: py38, mypy
     3.9: py39, mypy
     3.10: py310, mypy, packaging, pep8, codestyle, docstyle, vulture
```
m-col, line 140:
```suggestion
    3.10: py310, mypy
```

.github/workflows/ci.yml
elParaguayo: Yes. We can do that. It may break a couple of the screenshots while dbus-next isn't fixed, though.

m-col: I suppose for the docs it doesn't matter which version is used and the result would be the same? If so then yeah let's leave it for now

m-col:
There was a conflict and I tried out github's UI to resolve it but it added a merge commit which isn't ideal. Pls could you rebase and drop that last commit? There was a conflict in the changelog.

m-col (approved):
Awesome LGTM :)

m-col:
The two remaining github actions that are hanging are those 3.7 ones that have been removed. I wonder if a previous run of the CI (befoire updating the commits) told github that it wanted those, and it's cached those runs, expecting them to complete. The rest are done, so we could merge and hope that the CI on master does the right thing?

elParaguayo:
> The two remaining github actions that are hanging are those 3.7 ones that have been removed. I wonder if a previous run of the CI (befoire updating the commits) told github that it wanted those, and it's cached those runs, expecting them to complete. The rest are done, so we could merge and hope that the CI on master does the right thing?

I think they're just showing as they're "required" in the branch protections. I think merging this an updating the protections would fix this.

m-col:
I've updated the protections to drop the 3.7s and add the pypy3.9 and 3.11

PR #3934, opened 2022-10-28
Drop deprecated license_file from setup.cfg

m-col:
From setuptools 56.0.0 (Apr 2021) license_file is deprecated. Instead we could use license_files, but by default it finds LICENSE so it's redundant. This just removes it.


elParaguayo (approved):
LGTM

PR #3935, opened 2022-10-28
WL: don't segfault when clients don't provide drag icons

### LINKED ISSUE #3894: [wayland] qtile crashes (segfaults) on chrome tab drag n drop
### The issue:

When I start dragging a tab qtile immediately crashes. (newest qtile and chrome versions). No log output, direct segfault in qtile (dmesg). I would expect - even if this is a chrome issue - qtile to not crash :D)

I expected this to be a common wayland bug, since chrome (and electectron) is notoriously annoying there :P However after scouring the internet for potential fixes all bug reports I found have been fixed (for e.g. swaywm). Also the qtile issues relating this never describe this particular bug and are solved for the most part (#408 #1257).

How is drag-n-drop with other programs:
firefox: works, but only shows drag hints, does not actually drag the tab, same for vscode

Possibly related: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/46

Thus: did I miss some magic setting or is this a kown bug? (I get the behavior with the default config as well)
The way I start chrome:  `google-chrome --ozone-platform-hint=auto --enable-features=WebRTCPipeWireCapturer`

Chrome: Version 106.0.5249.119 (Official Build) (64-bit)
Qtile: 0.22.2.dev9+g946908b9

### Required:

m-col:
Sometimes they don't provide icons when dragging (e.g. chrome when dragging tabs). Thus `wlr_drag.icon` can be `None`.

Requires https://github.com/flacjacket/pywlroots/pull/107

Fixes #3894

### DIFF
--- a/libqtile/backend/wayland/output.py
+++ b/libqtile/backend/wayland/output.py
@@ -227,8 +227,8 @@ class Output(HasListeners):
         self, dnd: Dnd, now: Timespec, scale: float, transform_matrix: Matrix
     ) -> None:
         """Render the drag-n-drop icon if there is one."""
-        icon = dnd.wlr_drag.icon
-        if icon.mapped:
+        icon = dnd.icon
+        if icon is not None and icon.mapped:
             texture = icon.surface.get_texture()
             if texture:
                 box = Box(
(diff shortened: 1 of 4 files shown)


elParaguayo (approved):
LGTM - I see that `pywlroots` PR was merged too so this should be good to go.

PR #3938, opened 2022-10-29
Fix CryptoTicker test

m-col:
83388a6 made some changes to this widget but CI was failing at the time, so this follow-up PR fixes the test for it.

### DIFF
--- a/test/widgets/test_crypto_ticker.py
+++ b/test/widgets/test_crypto_ticker.py
@@ -22,6 +22,7 @@
 # We just need to test parsing here.
 
 from libqtile import widget
+from test.widgets.conftest import FakeBar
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Not sure where all those mypy errors are coming from!

m-col:
Ah yeah don't worry about those, it was an update in pywlroots. They'll be fixed by https://github.com/qtile/qtile/pull/3935, which I can rebase on top of this after merging and hopefully they'll all pass.

m-col:
I merged the other PR that should fix the type checking so rebased this so now CI should pass completely here. If not, I can append more fixes so that when merging this, other PRs will start passing again

m-col:
Nice this has fixed it all :)

elParaguayo:
Thanks.

PR #3939, opened 2022-10-30
Fire client_killed hook after client is removed from group

### LINKED ISSUE #3513: [Wayland] Crash on client_killed hook when calling cmd_to_layout_index

m-col:
I'm not sure why this isn't the case already but let's keep an eye out for issues that arise due to this. We want a killed window to be removed from its group so that hooked functions that cause the group to lay out and configure windows don't cause errors (e.g. a wlroots assertion is raised when configuring a killed window). However we want `c.group` to still point to the group because it's useful for hooked functions, and code already exists that use it so this stops that from breaking.

fixes #3513

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -658,15 +658,20 @@ class Qtile(CommandObject):
     def unmanage(self, wid: int) -> None:
         c = self.windows_map.get(wid)
         if c:
-            hook.fire("client_killed", c)
+            group = None
             if isinstance(c, base.Static):
                 if c.reserved_space:
                     self.free_reserved_space(c.reserved_space, c.screen)
             elif isinstance(c, base.Window):
                 if c.group:
+                    group = c.group
                     c.group.remove(c)
             del self.windows_map[wid]
             self.core.update_client_list(self.windows_map)
+            if isinstance(c, base.Window):
+                # Put the group back on the window so hooked functions can access it.
+                c.group = group
+            hook.fire("client_killed", c)
 
     def find_screen(self, x: int, y: int) -> Screen | None:
         """Find a screen based on the x and y offset"""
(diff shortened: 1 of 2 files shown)


elParaguayo (approved):
It feels a bit weird to reset the group after the window has been removed from that group but I suppose it's really just a reference to the group that it was on as it has actually been removed from the group.

So... LGTM.

m-col:
Agreed, a nicer way might be to pass the window _and_ the group to hooks but that would be a breaking change for little benefit. After the hook, the window object has no more references and will get GCed so at least the group attribute won't last much longer.

PR #3945, opened 2022-10-31
Fix KeyChord stack bug

### LINKED ISSUE #3937: Outer Non-modal KeyChord in Nested Chain Not Aborted After Inner Key Is Pressed
### The issue:

In previous versions of Qtile, when I press a key in the innermost layer of a nested keychord chain, all of the outer keychords in the chain would be exited. However recently (I believe it started when I updated to 0.22.1, but I didn't check at the beginning as I assumed I mispressed something). I have not changed my config recently (until today when I was trying to figure out what was going on).

Version: 0.22.1 installed with Pacman on Arch

## Minimal Config
An example minimal config that has a problematic KeyChord chain in it can be created by inserting the following into the default config file (https://docs.qtile.org/en/latest/manual/config/default.html#default-config):
In the imports section:
```python
import KeyChord
```
In the `keys` list:
```python
    KeyChord([mod], 'x', [
        Key([], 't', lazy.spawn(terminal)),
        KeyChord([], 'w', [
            Key([], 'w', lazy.spawn('firefox')),
            Key([], 'q', lazy.spawn('firefox https://docs.qtile.org/en/latest/manual/config/default.html#default-config')),
        ]),
    ]),
```
### Expected Behavior
Using this configuration I would expect `Super-x, w, w, t` to run `firefox` and send a 't' to the focused application (probably typing a 't' into the address bar of Firefox).
### Actual Behavior
Firefox is started, but rather than the `t` being sent to Firefox, a terminal window is launched.
### Expected Behavior
`Super-x, w, w, t, t` should launch Firefox and send 'tt' to it.
### Actual Behavior
Firefox is started, a terminal is started, and a 't' is typed into the terminal.

### Xephyr Note
When running this in Xephyr for me, firefox opens outside, but if the mouse and keyboard are grabbed you can still see that the terminal gets launched with the next press of `t`, I would have used another program that acts less weird, but I can't think of any other ubiquitous GUI programs.

### Required:

elParaguayo:
Changing `KeyChord.mode` to a boolean broke some functionality when using a stack of chords where non-modal chords were not exited on a keypress.

Fixes #3937

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -473,7 +473,7 @@ class Qtile(CommandObject):
         # Find another named mode or load the root keybindings:
         while self.chord_stack:
             chord = self.chord_stack.pop()
-            if chord.mode != "":
+            if chord.mode:
                 self.grab_chord(chord)
                 break
         else:


m-col (approved):
Nice catch, it's a shame it wasn't caught earlier by mypy. One day the whole codebase will be typed and a whole class of bugs will be eliminated :)

LGTM

PR #3949, opened 2022-11-02
 Add GenPollCommand() widget

thingsiplay:
Configuring GenPollText() to output scripts and shell command output can be tricky for new users. This new widget should make this task more easy and straight forward.

### DIFF
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -106,7 +106,7 @@ class GenPollCommand(base.ThreadPoolText):
 
     def __init__(self, **config):
         base.ThreadPoolText.__init__(self, "", **config)
-        self.add_defaults(Output.defaults)
+        self.add_defaults(GenPollCommand.defaults)
 
     def _configure(self, qtile, bar):
         base.ThreadPoolText._configure(self, qtile, bar)
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -92,3 +92,47 @@ class GenPollUrl(base.ThreadPoolText):
             text = "Can't parse"
 
         return text
+
+    
+class GenPollCommand(base.ThreadPoolText):
+    """A generic text widget to display output from scripts or shell commands"""
+
+    defaults = [
+        ("update_interval", 60,"update time in seconds"),
+        ("cmd", "date", "command line as a string or list of arguments to execute"),
+        ("shell", True, "run command through a shell to enable piping and shell expansion"),
+        ("text_before", "", "additional text to display before command output"),
+        ("text_after", "", "additional text to display after command output"),
+    ]
+
+    def __init__(self, **config):
+        base.ThreadPoolText.__init__(self, "", **config)
+        self.add_defaults(Output.defaults)
+
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
+
+        self.text_before = str(self.text_before)
+        self.text_after = str(self.text_after)
+
+        self.add_callbacks({
+            "Button1": self.cmd_force_update,
+        })
+
+    def poll(self):
+        self.run_process()
+        return self.get_output()
+
+    def run_process(self):
+        self.completed_process = subprocess.run(
+            self.cmd,
+            capture_output=True,
+            text=True,
+            shell=self.shell,
+        )
+        return self.completed_process
+
+    def get_output(self):
+        output = self.completed_process.stdout.strip()
+        output = self.text_before + output + self.text_after
+        return str(output)
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -1,4 +1,5 @@
 import json
+import subprocess
 from typing import Any
 from urllib.error import URLError
 from urllib.request import Request, urlopen
@@ -93,16 +94,14 @@ class GenPollUrl(base.ThreadPoolText):
 
         return text
 
-    
+
 class GenPollCommand(base.ThreadPoolText):
     """A generic text widget to display output from scripts or shell commands"""
 
     defaults = [
-        ("update_interval", 60,"update time in seconds"),
-        ("cmd", "date", "command line as a string or list of arguments to execute"),
-        ("shell", True, "run command through a shell to enable piping and shell expansion"),
-        ("text_before", "", "additional text to display before command output"),
-        ("text_after", "", "additional text to display after command output"),
+        ("update_interval", 60, "update time in seconds"),
+        ("cmd", None, "command line as a string or list of arguments to execute"),
+        ("shell", False, "run command through shell to enable piping and shell expansion"),
     ]
 
     def __init__(self, **config):
@@ -112,16 +111,12 @@ class GenPollCommand(base.ThreadPoolText):
     def _configure(self, qtile, bar):
         base.ThreadPoolText._configure(self, qtile, bar)
 
-        self.text_before = str(self.text_before)
-        self.text_after = str(self.text_after)
-
         self.add_callbacks({
-            "Button1": self.cmd_force_update,
+            "Button1": self.force_update,
         })
 
     def poll(self):
-        self.run_process()
-        return self.get_output()
+        return self.run_process().stdout.strip()
 
     def run_process(self):
         self.completed_process = subprocess.run(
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks for submitting this. As mentioned on Reddit, I've got some comments on this which I'll submit later today.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
 
         return text
+
+    
+class GenPollCommand(base.ThreadPoolText):
+    """A generic text widget to display output from scripts or shell commands"""
+
+    defaults = [
+        ("update_interval", 60,"update time in seconds"),
+        ("cmd", "date", "command line as a string or list of arguments to execute"),
```
elParaguayo, line 102:
```suggestion
        ("cmd", None, "command line as a string or list of arguments to execute"),
```
I wouldn't have a default value here.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+        ("shell", True, "run command through a shell to enable piping and shell expansion"),
```
elParaguayo, line 103:
```suggestion
        ("shell", False, "run command through a shell to enable piping and shell expansion"),
```
I'd be inclined to set this to `False`. Usage of `shell=True` is typically discouraged unless required (for the reasons you say).

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+        ("text_before", "", "additional text to display before command output"),
+        ("text_after", "", "additional text to display after command output"),
```
elParaguayo, line 105: Don't think you need these lines. The `fmt` parameter is available to `ThreadPoolText` widgets and this will be automatically applied by default.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+        self.add_callbacks({
+            "Button1": self.cmd_force_update,
+        })
+
+    def poll(self):
+        self.run_process()
+        return self.get_output()
+
+    def run_process(self):
+        self.completed_process = subprocess.run(
```
elParaguayo, line 127: You need to `import subprocess`

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+
+        self.text_before = str(self.text_before)
+        self.text_after = str(self.text_after)
+
+        self.add_callbacks({
+            "Button1": self.cmd_force_update,
+        })
+
+    def poll(self):
+        self.run_process()
```
elParaguayo, line 123: `run_process` and `get_output` don't need to be in separate methods. You could just have it all in `poll`.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+        ("update_interval", 60,"update time in seconds"),
+        ("cmd", "date", "command line as a string or list of arguments to execute"),
+        ("shell", True, "run command through a shell to enable piping and shell expansion"),
+        ("text_before", "", "additional text to display before command output"),
+        ("text_after", "", "additional text to display after command output"),
+    ]
+
+    def __init__(self, **config):
+        base.ThreadPoolText.__init__(self, "", **config)
+        self.add_defaults(Output.defaults)
```
elParaguayo, line 110:
```suggestion
        self.add_defaults(GenPollCommand.defaults)
```

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
             text = "Can't parse"
 
         return text
+
+
```
elParaguayo, line 96:
```suggestion

```
Line has whitespace.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
             text = "Can't parse"
 
         return text
+
+    
+class GenPollCommand(base.ThreadPoolText):
+    """A generic text widget to display output from scripts or shell commands"""
+
+    defaults = [
+        ("update_interval", 60,"update time in seconds"),
```
elParaguayo, line 101:
```suggestion
        ("update_interval", 60, "update time in seconds"),
```

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+            self.cmd,
+            capture_output=True,
+            text=True,
+            shell=self.shell,
+        )
+        return self.completed_process
+
+    def get_output(self):
+        output = self.completed_process.stdout.strip()
+        output = self.text_before + output + self.text_after
```
elParaguayo, line 137: This line is not needed (as per comment above).

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+        self.add_defaults(Output.defaults)
+
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
+
+        self.text_before = str(self.text_before)
+        self.text_after = str(self.text_after)
+
+        self.add_callbacks({
+            "Button1": self.cmd_force_update,
```
elParaguayo, line 119:
```suggestion
            "Button1": self.force_update,
```

In the latest git version, we don't use the `cmd_` prefix any more.

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +92,47 @@ def poll(self):
...
+            capture_output=True,
+            text=True,
+            shell=self.shell,
+        )
+        return self.completed_process
+
+    def get_output(self):
+        output = self.completed_process.stdout.strip()
+        output = self.text_before + output + self.text_after
+        return str(output)
```
elParaguayo, line 138: Doesn't `text=True` mean you get a string object already? If so, you could just `return output`.

elParaguayo (changes requested):
Thanks for this. I've made some comments and suggested some changes (some are just formatting changes which are needed for the tests to pass).

You also need to add an entry to the `widgets` dict in `libqtile/base/widget/__init__.py`. The entry is `WidgetClassName: module` so, would be `"GenPollCommand":  "generic_poll_text"`.

Lastly, can you add an entry to the CHANGELOG to say that we've added a new widget.

libqtile/widget/generic_poll_text.py
thingsiplay: I had this with my first try of the PR and forgot to add it again when submitting.

thingsiplay:
> You also need to add an entry to the widgets dict in libqtile/base/widget/__init__.py.

Do you mean at https://github.com/qtile/qtile/blob/master/libqtile/widget/__init__.py ? Should I create a new PR for this change or is it done in this PR somewhere?

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +93,36 @@ def poll(self):
...
+        return self.run_process().stdout.strip()
+
+    def run_process(self):
+        self.completed_process = subprocess.run(
+            self.cmd,
+            capture_output=True,
+            text=True,
+            shell=self.shell,
+        )
+        return self.completed_process
```
elParaguayo, line 128:
```suggestion
    def poll(self):
        process = subprocess.run(
            self.cmd,
            capture_output=True,
            text=True,
            shell=self.shell,
        )
        return process.stdout.strip()
```

There's no need for `run_process` to be a separate method. I'd keep it all in `poll`.

elParaguayo (changes requested):

libqtile/widget/generic_poll_text.py
```diff
@@ -92,3 +93,36 @@ def poll(self):
...
+    def __init__(self, **config):
+        base.ThreadPoolText.__init__(self, "", **config)
+        self.add_defaults(GenPollCommand.defaults)
+
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
+
+        self.add_callbacks({
+            "Button1": self.force_update,
+        })
```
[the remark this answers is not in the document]
thingsiplay: Maybe it was the last comma. I also removed the empty line above it.

elParaguayo (approved):

PR #3959, opened 2022-11-06
Fix scratchpad spawn issue

elParaguayo:
0b028019 introduced a bug that, when you kill the dropdown, the `hide` method of the dropdown is being called before the `on_client_killed` method. This is an issue as `hide` unsubscribes some hooks (if `on_focus_lost_hide` is set to `True`) but `on_client_killed` also tries to unsubscribe these hooks and so raises an exception.


m-col:
Noice thanks

PR #3960, opened 2022-11-06
Add command to do interactive mac-like sliding between groups

m-col:
Submitting to start discussion and get feedback and any eager testers :)

Note: the first 13 commits are from PR https://github.com/qtile/qtile/pull/3862 as the changes in this feature are based on that git history. That PR is mostly bug fixes and is ready for merge, but this one may need more discussion or work before merging, so I've kept them separate.

The remaining commits (starting from 89724d3728a31a3e278c4ff67b682acb5b202f08) do:

- add `Swipe` and `Pinch` configurables, subclasses of `Drag`, that similarly can be used for interactive actions. Unfortunately these are Wayland-exclusive. I don't know how they could be implemented in the X backend without a good amount of libinput wrapping C, which would essentially write `libinput-gestures` into our own codebase. The key difference between something like `libinput-gestures` and this feature is the maintained interactivity.
    - I'm unsure whether pinch bindings should be included, or just included yet. Swiping is essentially just a coordinate translation, like dragging, so shares much of the same code within the `Qtile` class. Pinching OTOH has a scaling component which would likely need to be relayed to any bound commands.
- Adds a command (with corresponding `start` command) to `Screen` that can be bound to `Drag` or `Swipe` to start the sequence.
- Adds an implementation for both X11 and Wayland backends.

See commit messages and in-code comments and the command docstrings for more info. If anyone tests this out and finds issues or has comments please feel free to post!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Add ability to have different border color when windows are stacked in Stack layout. Requires 
           setting `border_focus_stack` and `border_normal_stack` variables.
         - New widget: GenPollCommand
+        - Add `Swipe` and `Pinch` actions, similar to `Drag`, but for touchpad gestures.
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
--- a/libqtile/backend/base.py
+++ b/libqtile/backend/base.py
@@ -87,6 +87,18 @@ class Core(CommandObject, metaclass=ABCMeta):
     def ungrab_pointer(self) -> None:
         """Release grabbed pointer events"""
 
+    def start_slide_into_group(
+        self,
+        screen: config.Screen,
+        next_group: _Group,
+        prev_group: _Group,
+        scale: float,
+    ) -> None:
+        """Start a slide between groups, if supported."""
+
+    def slide_to_group(self, dx: int, dy: int) -> None:
+        """Continue a slide between groups, if supported."""
+
     def on_config_load(self, initial: bool) -> None:
         """
         Respond to config loading. `initial` will be `True` if Qtile just started.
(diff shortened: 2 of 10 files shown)


m-col:
Here's a video for demonstration, in wayland with `Swipe`: https://f.mcol.xyz/1667759509.mp4 (github limits video size so ive hosted it)

elParaguayo:
Really impressive.

m-col:
A couple things worth mentioning:

- the wallpaper doesn't move in the X implementation unfortunately. That's a To Do, assuming it doesn't require us sending a big stream of data to the X server at a high frequency.
- there are a couple of things that can cause a crash or create a weird state, typically when an action such as changing groups happens while the sequence is non-interactively finishing up (i.e. when you let go, it smoothly finishes the sequence, but some things need to be accounted for during that period).

m-col:
Converting this to draft as there are a number of improvements that are needed before consideration. The changes I just pushed work to encapsulate the feature a bit more inside a helper for the wayland backend. I'll likely to the same for the X backend too. This makes things more easy to reason about. I also need to account for multiple monitors, and make it possible to continue interactivity when a swipe is non-interactively finishing up.

This kind of feature I'd much prefer to have inside qtile-examples as a standalone 'plugin' but unsurprisingly it needs pretty tight integration with each backend. This is something I'm keeping in mind as I work on it though.

jwijenbergh:
This works really well! Very cool :). Tested on the wayland backend

clotodex:
Btw regarding `Pinch`, something that could be really cool is a "compositor-level-zoom".
This would allow to interact with a fully layouted group in a completely new way, instead of maximizing a window, which is quite binary, one could zoom in and when lifting the fingers it snaps back (or stays with a different keybind).
This might require more changes in backend though (Might be easier with wayland than X though).

I just wanted to drop this as inspiration of what "pinch" could be good for, but likely way out of scope for this PR.

jgarte:
Sounds like this might be useful for the PinePhone

m-col:
Sounds cool indeed! I need to rebase this PR onto the wlroots 0.16 PR once that's merged. I was getting a segfault occasionally and never managed to track down the cause, but I think the 0.16 changes will help prevent it, as we hand off more of the rendering code to wlroots.

PR #3971, opened 2022-11-08
Tidying up Mirrors some more

elParaguayo:
This removes the untidy handling of the mirror length by having the mirror inherit the reflection's `length_type` and then having the `length` getter and setter behave appropriately.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -881,8 +881,7 @@ class Mirror(_Widget):
         _Widget.__init__(self, reflection.length, **config)
         self.reflects = reflection
         self._length = 0
-        if self.reflects.length_type == bar.STRETCH:
-            self.length_type = bar.STRETCH
+        self.length_type = self.reflects.length_type
 
     def _configure(self, qtile, bar):
         _Widget._configure(self, qtile, bar)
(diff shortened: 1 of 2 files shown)


libqtile/widget/base.py
```diff
@@ -891,6 +890,9 @@ def _configure(self, qtile, bar):
         # there's no reflection, the mirror matches its parent bar.
         self.drawer.clear(self.background or self.bar.background)
 
+    def calculate_length(self):
+        return self.reflects.calculate_length()
+
```
m-col, line 895: If here it's always using the length of the widget it reflects, how does it stretch according to its own bar?

m-col: Ah of course! :smile: 

Are there any weird or unexpected effects that could arise by having mirrored widgets be different sizes? I suppose the only risk is that the content of a larger original widget can be cut off when drawing to a smaller mirror? Is drawing a larger buffer to a smaller buffer allowed?

elParaguayo: The only time the widget would be a different size is when it's a stretch widget. In that scenario, contents could be cut off if the widget has less space. This is fine and is no different to the default behaviour of a stretch widget.

m-col (approved):
Sounds good :+1: LGTM

Issue #3976, opened 2022-11-10
Wayland: cursor shouldn't warp to layer shell surfaces


m-col:
### The issue:

Cursor warping using `cursor_warp = True` should only apply to regular windows, not layer surface windows like notifications.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

Issue #3764, opened 2022-08-14
Chromium and Telegram windows are shifted in Qtile Wayland


TinyAtoms:
# Problem
Chromium and Telegram windows are shifted towards the bottom-right in qtile. This happens in all layouts.
The problem is not present when logging in Qtile for xorg, and also not present in sway, so I believe it to be a problem specific to Qtile wayland.

## Expected behaviour
This is what I expect normal windows to behave like
![image](https://user-images.githubusercontent.com/44430688/184521192-7b51bb8e-a12e-4ca6-8d95-3f8a35c7a9f4.png)
## Resulting behaviour
This is how Telegram and Chromium behave
![image](https://user-images.githubusercontent.com/44430688/184521207-8bd1b68b-5c50-483f-b056-2df0707e1a4a.png)
It is clearer that they're shifting bottom-right in the image below, where I placed the bar at the bottom, and chromium is overlapping with the bar.
![image](https://user-images.githubusercontent.com/44430688/184521221-7d8d30e8-b6e7-4a6a-abd7-15311e1bf3a0.png)

# Environment
I don't know what information is relevant, and in the interest of being not overly verbose I've only added these. There is no relevant qtile log, since there are no warnings or errors produced for this issue. 

Distro: Arch linux
qtile 0.21.0 
wlroots 0.15.1-6
Dunno if you need these, but here they are
wayland 1.21.0-1 
wayland-protocols 1.26-1
xorg-xwayland 22.1.3-1 
python-pywayland 0.4.14-1 

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
Do any of the appearance settings within chromium make a difference? Such as the theme or 'Use system title bar and borders'

TinyAtoms:
I can't believe I didn't think of that! That fixed the issue for both chromium and telegram.
The relevant setting in telegram is located in `Settings > advanced > use system window frame`
Thanks!

clotodex:
@TinyAtoms I have the same issue. Asking Telegram to use the system frame just fills the gap with a really ugly white frame. Can that be configure as well? (like in qtile?).

clotodex:
@TinyAtoms @m-col Relaunching Telegram fixed the offset (duh! sorry). Your advice fixed all issues for me. Feel free to close this issue.

TinyAtoms:
Do we know what's causing the issue? What is the common factor between telegram and chromium?

clotodex:
[...]

PR #3985, opened 2022-11-13
Wlroots 0.16.0 and migration to wlr scene graph API

m-col:
Draft for now because internal windows aren't supported yet (struggling to figure out the best way to manage them).

Also changes are needed in pywlroots too of course. Posting just to get the ball rolling. 

TODO:

- [x] kitty isn't working completely
- [x] xwayland pointer focus is jammy, see https://github.com/qtile/qtile/pull/3985#issuecomment-1619110269
- [x] ensure client-side and user-side fullscreening works as expected
- [x] ensure output layout is getting reconfigured at the right place
- [x] get Internal windows working
- [x] get wallpapers working
- [x] update MOCK_MODULES
- [x] can we remove Drawer intermediate surface?
- [x] window opacity
- [x] remove `Internal` window artifact visible at startup
- [x] Is under_pointer working correctly i.e. across window types, borders and the background?
- [x] check gamma control works
- [x] check `window.static()` isn't completely broken
- [x] ensure windows are able to start up hidden (when matched to groups)
- [x] can we drop SubSurface completely?
- [x] fix floating window borders during drag

also closes https://github.com/qtile/qtile/issues/3976


elParaguayo:
You probably need to add `wlroots.wlr_types.scene` to `MOCK_MODULES` in `docs/conf.py`. There may be other modules too.

m-col:
Just some updates on this for those that are interested:

- This needs corresponding changes in pywlroots available at PR https://github.com/flacjacket/pywlroots/pull/109
- I migrated the backend to use wlroots' scene graph API, which was work that was going to be done anyway but doing that work when updating for the new wlroots release means that less work would be needed in total, particularly when it comes to updating `Internal` windows (whose current implementation within Qtile has been impacted hard by the wlroots changes)
- XDG shell, layer shell, XWayland, drag icons etc all generally work
- Internal windows and Qtile-managed wallpapers do not yet work
- I haven't done anything specific for XWayland override-redirect windows but i suspect they might work
    - similarly i haven't tested xdg popups but they might work

The scene graph design is similar to `dwl`'s (and indeed much credit goes to them for many of these changes), where we have a persistent list of subtrees that represent Z axis layers; these are created in `Core.__init__`, hopefully they are clear. Layer shell windows are inserted into the subtree corresponding to their layer. Regular windows have a dedicated subtree (for both XDG shell and xwayland clients) at `Core.window_tree`. Regular windows are each allocated a dedicated subtree within the `window_tree`, which holds their surface, borders and any other popups and subsurfaces etc that belong to them. Drag icons are also given a dedicated subtree at the top.

m-col:
Pushed an update that starts to replace the wlr_textures used for `Internal` windows and wallpapers with a cairo-backed wlr_buffer implementation, similar to the idea discussed at https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3354. As this added some more C code, I replaced the libinput-specific ffi builder with a general backend one. The wlr_buffer implementation compiles by itself (if you comment out the line including cairocffi's ffi builder in the new ffi builder), but creating a wlr_buffer using a cairo surface fails as the surface's pointer comes from a different ffi instance:
```
TypeError: initializer for ctype 'struct _cairo_surface *' appears indeed to be 'struct _cairo_surface *', but the types are different (check that you are not e.g. mixing up different ffi instances)
```
The solution here is to `ffi.include()` the ffi instance from cairocffi in the build step. However, due to duplicate definitions of various XCB types in xcffib (included by cairocffi if available) and pywlroots (for XWayland), we get this error (likely for a large number of types, not just `xcb_generic_event_t`:
```
cffi.FFIError: multiple declarations of anonymous xcb_generic_event_t (for interactive usage, try cdef(xx, override=True))
```
Posting this as I'm a bit of a brick wall with this... not sure what the 'correct' approach is for handling this, though even a hacky approach might do. If anybody has any ideas for merging multiple FFIs together that may duplicate some definitions, any pointers would be helpful :)

m-col:
Marked ready for review as this is now working... for the most part. It's at a mostly usable state without any obvious bugs, but there are occasional crashes that will require regular usage to identify properly. The CI is failing because the pywlroots used in CI is the last release and this PR depends on https://github.com/flacjacket/pywlroots/pull/109.

So if anyone has some time for code review or just test-driving the changes then that would be appreciated :)

m-col:
Also fixes https://github.com/qtile/qtile/issues/3764,

m-col:
I fixed the issue @jwijenbergh mentioned on discord about reserved space. I think there was an existing bug with that code which is now fixed. While working on this, I made a grab bag of other changes to bar.py, most of which hopefully aren't controversial, but some experimentation with adding static typing to the configuables Gap and Bar is a bit messy. 

The "solution" implemented in the last commit is not nice, though it *does* enforce correct typing of bar.py. I definitely won't make equivalent changes to other configurables without first redesigning how to nicely add typing to all configurables in a way that doesn't make development more difficult. Obviously typing exists to make development easier. I do have some ideas on how to make it a nice thing, such as replacing the `defaults` options lists with class attributes that take advantage of `typing.Annotated` to specify accepted types, default values and documentation all wrapped up together. This would also allow complete runtime type checking of options passed to configurables. I plan to use `Bar` as my experimental subject to get something working, but if I can't get something nice then I will probably undo the ugly typing implemented in the last commit I just pushed.

Anyway - see the commit message for the changes made in that commit. I think that's the only commit that made significant changes outside of `libqtile/backend/wayland/`.

m-col:
To make that last typing stuff seem less nebulous, here is a working prototype for how this `Configurable` system would work:
```py
#!/usr/bin/env python3

from inspect import get_annotations
from typing import Annotated, get_origin, get_args


class _ConfigurableMeta(type):
    def __init__(self, clsname, bases, attrs):
        # Valid types for options
        opt_types = {}

        for name, value in get_annotations(self, eval_str=True).items():
            if get_origin(value) is Annotated:
                valid_type, _doc = get_args(value)
                opt_types[name] = valid_type

        self._opt_types = opt_types

        return super().__init__(clsname, bases, attrs)


class Configurable(metaclass=_ConfigurableMeta):
    def __init__(self, **config):
        for name, value in config.items():
            if name in self._opt_types:
                if isinstance(value, self._opt_types[name]):
                    # Usual Configurable logic happens here where the passed value is
                    # stored for later use in __getattr__. Simplified here to setattr.
                    # Note this does not happen if the passed value is an invalid type,
                    # or for passed options that aren't annotated.
                    setattr(self, name, value)
                else:
                    print(f"Option '{name}' passed invalid type '{type(value)}'")
            else:
                clsname = self.__class__.__name__
                print(f"{clsname} received invalid option '{name}'")


class SomethingUseful(Configurable):
    # `Annotated` class attributes replace the current `defaults` list of tuples
    # The structure is:
    # <name>: Annotated[<valid type>, <documentation>] = <default value>
    opt: Annotated[int, "this is an int"] = 1

    def __init__(self, **config):
        super().__init__(**config)


print(" - Default:")
obj = SomethingUseful()
print(obj.opt)

print("\n - Valid value:")
obj = SomethingUseful(opt=2)
print(obj.opt)

print("\n - Invalid value:")
obj = SomethingUseful(opt=None)
print(obj.opt)

print("\n - Unknown opt:")
obj = SomethingUseful(var=None)
print(hasattr(obj, "var"))
```

Which when run, outputs this (python 3.8+):
```
 - Default:
1

 - Valid value:
2

 - Invalid value:
Option 'opt' passed invalid type '<class 'NoneType'>'
1

 - Unknown opt:
SomethingUseful received invalid option 'var'
False
```
This has the following functional changes:

- runtime type checking, so we can log when options have invalid types and replace them with the default value
- logging of and disregarding any unknown arguments
- types of options are clearly documented and can be pulled out for the sphinx documentation
- mypy and other static type checkers are meant to treat `Annotated[t, ...]` type annotations as `t`, so the `opt` option in the above example is typed as `int`, thus adding static type checking to all configurables.
  - so we won't shoot ourselves in the foot and can catch more bugs
  - potentially makes `qtile check` extremely powerful for identifying config issues, because the type checking of configuables penetrates completely into their implementation

Note that this would be a gradual conversion, not an all-in-one change, so if people supported this we would retain `defaults` support until we have fully converted them to this approach, and then some further time.

Thoughts on this?

Note that this is completely tangential to this wayland PR, I'm just excited for this and I eagerly modified `bar.py` in a drive-by while working on the wayland stuff, so figured I'd give more explanation.

Sydiepus:
Hello, I wanted to test the PR out but I am unable to get the `ffibuild` script working.
The `ffibuild` script is working fine on the main branch.
It also fails to install because of `libqtile` is not found though `export PYTHONPATH=$(pwd)` fixes the issue, and there's the `_ffi` not being found probably because the `ffibuild` script not working.

I'm using [my fork](https://github.com/Sydiepus/qtile) of the PR and [my fork](https://github.com/Sydiepus/pywlroots) of pywlroots.

jwijenbergh:
> Hello, I wanted to test the PR out but I am unable to get the `ffibuild` script working. The `ffibuild` script is working fine on the main branch. It also fails to install because of `libqtile` is not found though `export PYTHONPATH=$(pwd)` fixes the issue, and there's the `_ffi` not being found probably because the `ffibuild` script not working.
> 
> I'm using [my fork](https://github.com/Sydiepus/qtile) of the PR and [my fork](https://github.com/Sydiepus/pywlroots) of pywlroots.

what error does the ffibuild script give?

m-col:
You can get full output from `ffibuild` with `ffibuild -v`. Make sure you have wlroots 0.16.0 available. For example, I have wlroots cloned, checked out to 0.16.0, and compiled. Then I set these environmental variables:
```sh
# This one is needed for qtile to run, so I set it in my startup script before starting the qtile process
export LD_LIBRARY_PATH=~/git/wlroots/build:$LD_LIBRARY_PATH

# These are only needed for compilation, so it's in my zshrc:
export CFLAGS="-I $HOME/git/wlroots/include -I $HOME/git/wlroots/build/include"
export LDFLAGS="-L $HOME/git/wlroots/build"
```

---

There are a couple of remaining bugs that I don't think should block this, as they may not be Qtile bugs. Specifically:

- XWayland popup menus seem to stop becoming visible after a few re-uses. E.g. if I move the cursor over the dropdown menus in gimp 2-3 times so that the same menu opens and closes once or twice (the number doesn't seem to be consistent), it stops appearing. It still receives pointer input, so submenus and documentation tooltips can still open. However, this appears to be a bug present in wlroots master: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3570
- With `autofullscreen` enabled, un-fullscreening a fullscreened Firefox makes it send a request to fullscreen itself, which we respect. Therefore, firefox can't be un-fullscreened with `autofullscreen` enabled. I'm unsure if this is a Qtile, Firefox, or wlroots bug. Qtile does the same as `dwl` here, which I've used to guide much of the wlr_scene code, however I can't get Firefox to open in dwl at all.

I have a couple pending tasks before this is ready to go:

- [ ] go over the review comments on the pywlroots PR
- [ ] check other XWayland stuff works - some initial testing with `ardour` (a particularly annoying X11 client) suggests a few things are broken.

Sydiepus:
I'm using wlroots package provided by arch repos.
This is output of `ffibuild`
```
[dev1@DEV1 qtile]$ ./scripts/ffibuild -v
Building Pango interface
Building xcursors
Building PulseAudio interface
Building Wayland CFFI
    Failed!
    This is optional and is needed for: Wayland backend
Installing wlroots v0.15.24 requires wlroots v0.15.x, found v0.16.1
  File "/home/dev1/repos/qtile/./libqtile/backend/wayland/cffi/build.py", line 4, in <module>
    from libqtile.backend.wayland.cffi import cairo_buffer, libinput
ModuleNotFoundError: No module named 'libqtile'

[dev1@DEV1 qtile]$ export PYTHONPATH=$(pwd)
[dev1@DEV1 qtile]$ ./scripts/ffibuild -v
Building Pango interface
Building xcursors
Building PulseAudio interface
Building Wayland CFFI
    Failed!
    This is optional and is needed for: Wayland backend
Installing wlroots v0.15.24 requires wlroots v0.15.x, found v0.16.1
  File "/home/dev1/repos/qtile/./libqtile/backend/wayland/cffi/build.py", line 4, in <module>
    from libqtile.backend.wayland.cffi import cairo_buffer, libinput
  File "/home/dev1/repos/qtile/libqtile/backend/wayland/__init__.py", line 2, in <module>
    from libqtile.backend.wayland.inputs import InputConfig  # noqa: F401
  File "/home/dev1/repos/qtile/libqtile/backend/wayland/inputs.py", line 31, in <module>
    from libqtile.backend.wayland.wlrq import HasListeners, buttons
  File "/home/dev1/repos/qtile/libqtile/backend/wayland/wlrq.py", line 33, in <module>
    from libqtile.backend.wayland._ffi import ffi, lib
ModuleNotFoundError: No module named 'libqtile.backend.wayland._ffi'

[dev1@DEV1 qtile]$ cat /usr/include/wlr/version.h 
#ifndef WLR_VERSION_H
#define WLR_VERSION_H

#define WLR_VERSION_STR "0.16.1"

#define WLR_VERSION_MAJOR 0
#define WLR_VERSION_MINOR 16
#define WLR_VERSION_MICRO 1

#define WLR_VERSION_NUM ((WLR_VERSION_MAJOR << 16) | (WLR_VERSION_MINOR << 8) | WLR_VERSION_MICRO)

#endif
[dev1@DEV1 qtile]$ 

```

m-col:
@Sydiepus I haven't forgotte your message, I'm just ensuring I fully understand the changes in the PR so I don't post a misleading or misguided answer.

The bunch of commits I just pushed improve things a bunch, which is nice. Multi-monitor is largely working, with a couple of (non-segfaulting) quirks. Layer, XDG and Wayland surfaces are generally. Working. I tore out much of the xdg-popup management that seems to have reduce the area of possible problems. I still want to test more complicated XWayland clients, ones known to cause issues, because ideally we'd kill bad clients rather than let them kill us.

m-col:
@Sydiepus maybe building will work with this in the shell:
```
  export CFLAGS="$CFLAGS -I/usr/include/wlroots0.15"
  export LDFLAGS="$LDFLAGS -L/usr/lib/wlroots0.15"
```
Just to tie us over to the new release.

I've just pushed a bunch of work and pretty much everything is working as expected, and them some. No more segfaults fortunately, even Ardour is working perfectly (an X client that causes lots of issues). No more popup troubles either. I think this is ready for real-life testing, if anybody is interested in giving it a go. Remember, you need wlroots 0.16, and therefore _not_ the variables above.

Any issues we get whilst testing we can add to the first post for tracking.

m-col:
I forgot to mention that I've made further changes the the pywlroots PR mentioned above - anyone testing this branch will need those changes.

Sydiepus:
@m-col Thank you for your amazing work, I can't wait to test the PR.

It seems that i'm too stupid to setup a working dev environment even with the `CFLAGS` and `LDFLAGS`  exported i still get the same error.

I'm setting up another VM hopefully it'll work this time.

m-col:
74 commits lmao, so sorry @elParaguayo, @jwijenbergh. There are very few code changes outside of the wayland backend fortunately, but a number of changes outside of `libqtile/`. Through much of this work and testing it I did a few drivebys too, which I probably should have done as separate PRs on the side. Happy to help review synchronously, or have only a light review and fix bugs from master as they come - in which case it might be worth focussing your review on files outside of `libqtile/backend/wayland/`.

The commits I'm pushing at the minute are iterating on test and CI fixes and such, but the main body of the PR is "finished", so I've requested reviews :)

m-col:
The tests pass :tada:

I'm struggling a bit with setting up the RTD env though, perhaps resident expert @elParaguayo you might have some insight into the problems here? Essentially I added the script `scripts/ubuntu_wayland_setup` to set up wayland dependencies on Ubuntu with the aim of using it for github actions (successfully :heavy_check_mark:) as well as RTD, though it seems the way their environments are set up means we are a regular user without root access, I presume. Do you know how they set it up? I'm thinking it might be enough to only use `sudo` on github, and without it otherwise, instead sticking the `apt` dependencies into the RTD config rather than `apt install`ing them. Assuming that `ninja` also works without sudo...

elParaguayo:
I'll take a look at RTD later on but, for my own benefit, can I check why we need to build Wayland for the docs? Is this because we can't document the relevant classes without it? I didn't think we did this before so curious why it's needed now.

elParaguayo:
If we can't get RTD to work (I think they lock down ability to run commands as root) then an alternative may be to use github pages and point the docs.qtile.org address to that.

m-col:
>  I didn't think we did this before so curious why it's needed now.

Previously most of  `libqtile/backend/wayland/inputs.py` was under a huge if statement conditional upon the ffi within the backend being built. I removed it under the naive presumption that setting up RTD might be preferable but actually I'm concluding that it ain't. I'll stick the being if statement back. TBH I think less needs to be under it now anyway, only a couple of dicts at the top level that have `lib.<something>` as their values. I'll nuke all of this new RTD stuff, as long as the classes can be inspected by sphinx in the absence of the built ffi then it's all good.

libqtile/backend/wayland/drawer.py
```diff
@@ -64,25 +54,22 @@ def _draw(
...
-            width,
-            height,
-            cairocffi.cairo.cairo_image_surface_get_data(self._source._pointer),
-            dst_x=offsetx,
-            dst_y=offsety,
-        )
-        self._win.damage()
+        # Paint recorded operations to our window's underlying ImageSurface
+        with cairocffi.Context(self._win.surface) as context:
+            context.set_operator(cairocffi.OPERATOR_SOURCE)
```
jwijenbergh, line 59: do we still want to remove this line?

elParaguayo: I made a separate PR to rework the `Drawer` class. Ideally, that should be rebased onto this.

bodograumann:
I tried testing this (wrt #4258), but only got a Segmentation Fault. Probably because I am doing it totally wrong.
Is there a documentation for how to test such things that I haven't found?
Arch only has pywlroots 0.15, so I had to use the virtual environment. Here is what I did:

```
gh repo clone qtile/qtile
cd qtile
gh pr checkout 3985
./dev.sh
. ./venv/bin/activate
pip install pywlroots
CFLAGS=" -I/usr/include/wlroots" \
  LDFLAGS=" -L/usr/lib/wlroots" \
  PYTHONPATH=./ \
  ./scripts/ffibuild -v
./bin/qtile start -b wayland -c ./libqtile/resources/default_config.py
```

m-col:
re XWayland pointer focus being messed up:
```
[22:43]Jeroen: getting quirks where xwayland windows keep having focus
[22:44]Jeroen: I think some of it is fixed with 
diff --git a/libqtile/backend/wayland/xwindow.py b/libqtile/backend/wayland/xwindow.py
index 909cff25..af3ba59b 100644
--- a/libqtile/backend/wayland/xwindow.py
+++ b/libqtile/backend/wayland/xwindow.py
@@ -158,7 +158,6 @@ class XWindow(Window[xwayland.Surface]):
                     self.tree.node.set_position(self.borderwidth, self.borderwidth)
 
                 self.container.node.set_enabled(enabled=True)
-                self.core.focus_window(self)
                 return
 
         # This is the first time this window has mapped, so we need to do some initial
[22:44]Jeroen: but im not sure if that is correct
[22:47]Jeroen: Okay weird steps to reproduce but here they are so I don't forget:
[22:48]Jeroen:
Open firefox with the wayland backend
Open emacs (that uses xwayland by default)
Focus firefox with the mouse
The right click menu in firefox -> take a screenshot does nothing, you can observe in emacs that it tries to focus the emacs window by looking at the cursor inside the editor
 
[22:48]Jeroen: that patch fixes it
[22:49]Jeroen: unhiding whilst not mapping for the first time should not focus I think so maybe that patch makes sense
[22:51]Jeroen: hmmm we also do this focussing in xdgwindow, any reason why? (The focus within the if self not in self.core.pending_windows) 
```

m-col:
Think I fixed the XWayland issue, with it the firefox/emacs repro works as expected.

> Jeroen: hmmm we also do this focussing in xdgwindow, any reason why? (The focus within the if self not in self.core.pending_windows) 

To focus the window when it is unhidden. `unhide()` can be called from e.g. layouts or user code, so we focus the window when this happens. Though maybe I misunderstood which part you're asking about?

m-col:
@bodograumann sorry to miss your message. For this you'll need more than just the headers (which are sufficient for compiling pywlroots, for example). You need the compiled 0.16 wlroots library, then set `export LD_LIBRARY_PATH=$HOME/git/wlroots/build:` to the build folder containing it

jwijenbergh:
I unticked the top two TODOs as they're not quite working already. Xwayland pointer focus fix is on discord @m-col. Investigating kitty....

libqtile/backend/wayland/xdgwindow.py
```diff
@@ -176,16 +106,63 @@ def _on_set_app_id(self, _listener: Listener, _data: Any) -> None:
...
+
+        self.qtile.manage(self)
+
+        # Send a frame done event to provide an opportunity to redraw even if we aren't
+        # going to map (i.e. because the window was opened on a hidden group)
+        surface.surface.send_frame_done(Timespec.get_monotonic_time())
 
     @expose_command()
     def kill(self) -> None:
+        self.hide()
```
jwijenbergh, line 165: any reason why this is added? Removing this solves the kitty issue i'm having

libqtile/backend/wayland/xwindow.py
```diff
@@ -168,13 +139,92 @@ def _on_set_class(self, _listener: Listener, _data: Any) -> None:
...
+                # listen for commits on and create a tree for.
+                self.add_listener(self.surface.surface.commit_event, self._on_commit)
+                if not self.tree:
+                    self.tree = SceneTree.subsurface_tree_create(
+                        self.container, self.surface.surface
+                    )
+                    self.tree.node.set_position(self.borderwidth, self.borderwidth)
+
+                self.container.node.set_enabled(enabled=True)
+                self.core.focus_window(self)
```
jwijenbergh, line 165: I still think this needs to be removed. This gives me buggy pointer focus (the emacs + firefox issue) and hide/unhide still works after this

jwijenbergh:
Added my fixes to both as comments. Will spin this with those patches

libqtile/backend/wayland/xdgwindow.py
m-col: I think this can be removed. How has testing been? Any side effects with killing other clients?

libqtile/backend/wayland/xwindow.py
m-col: I'm not sure I fully understand 1) how it works without this and 2) why anything bad should happen with it, as it's basically a no-op when focussing the currently focussed window.

ftpd:
I'm testing this since yesterday and I can see two issues:

foot 1.15.x doesn't work (1.14 works fine) with the following:
```
warn: wayland.c:1493: fractional scaling not available
warn: wayland.c:1501: no server-side cursors available, falling back to client-side cursors
warn: wayland.c:1512: text input interface not implemented by compositor; IME will be disabled  <--- this line is also there when running working foot 1.14
zsh: segmentation fault  foot
```

Activation protocol seems not to work. When I run `sleep 3; tput bel` and change workspace, I don't have the urgency notification on the workspaces list (works fine on X11, so qtile itself is able to do it).

ftpd:
> foot 1.15.x doesn't work (1.14 works fine) with the following:

It seems it's because https://codeberg.org/dnkl/foot/pulls/1445 - so we don't have viewporert interface in qtile (whatever it is :P)

jwijenbergh:
> > foot 1.15.x doesn't work (1.14 works fine) with the following:
> 
> It seems it's because https://codeberg.org/dnkl/foot/pulls/1445 - so we don't have viewporert interface in qtile (whatever it is :P)

The following patch fixes this:

```diff
diff --git a/libqtile/backend/wayland/core.py b/libqtile/backend/wayland/core.py
index 119fe2b1..95a06701 100644
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -50,6 +50,7 @@ from wlroots.wlr_types import (
     RelativePointerManagerV1,
     ScreencopyManagerV1,
     Surface,
+    Viewporter,
     XCursorManager,
     XdgOutputManagerV1,
     input_device,
@@ -142,6 +143,7 @@ class Core(base.Core, wlrq.HasListeners):
             self.backend,
             self._subcompositor,
         ) = wlroots_helper.build_compositor(self.display)
+        self.viewporter = Viewporter(self.display)
         self.socket = self.display.add_socket()
         os.environ["WAYLAND_DISPLAY"] = self.socket.decode()
         logger.info("Starting core with WAYLAND_DISPLAY=%s", self.socket.decode())
```

cc @m-col

jwijenbergh:
@ftpd for ugency thingy, please try. Works for me with alacritty now, but not with foot (sway has the same issue, maybe it's a setting)

```diff
diff --git a/libqtile/backend/wayland/xdgwindow.py b/libqtile/backend/wayland/xdgwindow.py
index a1edddb3..8f110333 100644
--- a/libqtile/backend/wayland/xdgwindow.py
+++ b/libqtile/backend/wayland/xdgwindow.py
@@ -196,9 +196,7 @@ class XdgWindow(Window[XdgSurface]):
             self.group.focus(self)
 
         elif focus_on_window_activation == "smart":
-            if not self.group.screen:
-                logger.debug("Ignoring focus request (focus_on_window_activation='smart')")
-            elif self.group.screen == self.qtile.current_screen:
+            if self.group.screen == self.qtile.current_screen:
                 logger.debug("Focusing window (focus_on_window_activation='smart')")
                 self.qtile.current_screen.set_group(self.group)
                 self.group.focus(self)
```

not sure if this is correct @m-col

elParaguayo:
If this is working well enough now we should merge and then do incremental fixes separately.

jwijenbergh:
> If this is working well enough now we should merge and then do incremental fixes separately.

We can rebase and merge tbh, I can add those fixes in a follow up pr quickly

libqtile/backend/wayland/drawer.py
elParaguayo: @jwijenbergh that operator is actually really important. Was this the line that caused issues for you when testing this branch? Might be worth you testing the drawer refactor once it's in a slightly better state.

m-col:
> @ftpd for ugency thingy, please try. Works for me with alacritty now, but not with foot (sway has the same issue, maybe it's a setting)
> 
> ```diff
> diff --git a/libqtile/backend/wayland/xdgwindow.py b/libqtile/backend/wayland/xdgwindow.py
> index a1edddb3..8f110333 100644
> --- a/libqtile/backend/wayland/xdgwindow.py
> +++ b/libqtile/backend/wayland/xdgwindow.py
> @@ -196,9 +196,7 @@ class XdgWindow(Window[XdgSurface]):
>              self.group.focus(self)
>  
>          elif focus_on_window_activation == "smart":
> -            if not self.group.screen:
> -                logger.debug("Ignoring focus request (focus_on_window_activation='smart')")
> -            elif self.group.screen == self.qtile.current_screen:
> +            if self.group.screen == self.qtile.current_screen:
>                  logger.debug("Focusing window (focus_on_window_activation='smart')")
>                  self.qtile.current_screen.set_group(self.group)
>                  self.group.focus(self)
> ```
> 
> not sure if this is correct @m-col

The logic was likely copied from the X11 version. I don't think the diff actually changes anything though. If `self.group.screen == None` then it's a no-op in either case, right?

jwijenbergh:
> > @ftpd for ugency thingy, please try. Works for me with alacritty now, but not with foot (sway has the same issue, maybe it's a setting)
> > ```diff
> > diff --git a/libqtile/backend/wayland/xdgwindow.py b/libqtile/backend/wayland/xdgwindow.py
> > index a1edddb3..8f110333 100644
> > --- a/libqtile/backend/wayland/xdgwindow.py
> > +++ b/libqtile/backend/wayland/xdgwindow.py
> > @@ -196,9 +196,7 @@ class XdgWindow(Window[XdgSurface]):
> >              self.group.focus(self)
> >  
> >          elif focus_on_window_activation == "smart":
> > -            if not self.group.screen:
> > -                logger.debug("Ignoring focus request (focus_on_window_activation='smart')")
> > -            elif self.group.screen == self.qtile.current_screen:
> > +            if self.group.screen == self.qtile.current_screen:
> >                  logger.debug("Focusing window (focus_on_window_activation='smart')")
> >                  self.qtile.current_screen.set_group(self.group)
> >                  self.group.focus(self)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > not sure if this is correct @m-col
> 
> The logic was likely copied from the X11 version. I don't think the diff actually changes anything though. If `self.group.screen == None` then it's a no-op in either case, right?

It does change something because now we fallthrough here which is what we need https://github.com/m-col/qtile/blob/wlroots-0.16.0/libqtile/backend/wayland/xdgwindow.py#L206-L207

PR #3986, opened 2022-11-13
Added support for single margin and border for VerticalTile layout

matthewsia98:
added single_margin and single_border_width keys for VerticalTile layout similar to MonadTall

### DIFF
--- a/libqtile/backend/wayland/layer.py
+++ b/libqtile/backend/wayland/layer.py
@@ -104,7 +104,7 @@ class LayerStatic(Static[LayerSurfaceV1]):
         logger.debug("Signal: layerstatic map")
         self.mapped = True
         self.output.organise_layers()
-        self.focus(True)
+        self.focus(False)
 
     def _on_unmap(self, _listener: Listener, _data: Any) -> None:
         logger.debug("Signal: layerstatic unmap")


m-col:
Nice thanks! It looks like within that method we're checking `n > 1` a few times - perhaps that could simplified to just one if statement? Also it might be good at add the new option to the changelog so that others can see the change :)

elParaguayo:
Are we being consistent with names across layouts?

matthewsia98:
Names were taken from the MonadTall layout

image

elParaguayo:
Thanks. Could remember if we used `single_margin` or `margin_on_single`.

elParaguayo:
We should add a test for this.

matthewsia98:
> Nice thanks! It looks like within that method we're checking `n > 1` a few times - perhaps that could simplified to just one if statement? Also it might be good at add the new option to the changelog so that others can see the change :)

Done and done.

matthewsia98:
> Thanks. Could remember if we used `single_margin` or `margin_on_single`.

That's probably because the Columns layout uses margin_on_single.

image

libqtile/layout/verticaltile.py
```diff
@@ -118,21 +124,21 @@ def configure(self, window, screen_rect):
             index = self.clients.index(window)
 
             # border
-            if n > 1:
-                border_width = self.border_width
-            else:
-                border_width = 0
+            border_width = self.border_width if n > 1 else self.single_border_width
```
m-col, line 127: Might also be worth collapsing these if statements, border, margin, width and height

matthewsia98: done

matthewsia98: If you're talking about the ```__init__``` function, I was trying to keep it same as MonadTall

image

PR #4001, opened 2022-11-17
Add grow left and grow right commands to xmonadtall

martanman:


elParaguayo:
Thanks for submitting this and including a test.

The code all looks neat but I'm interested what's the reason for including this.

`grow_left/right` are just calling existing commands `grow/shrink_main`. Is this just for convenience or is it because other layouts have the same commands and so you want to use the same key binding?

Also, if you're adding these, wouldn't you also add `grow_up/down` to `MonadWide`?

martanman:
Yeah, it's so that it can be consistent with the corresponding commands of other layouts.

Sure, I don't mind adding a `grow_up/down` to `MonadWide` as well.

jx11r:
What's wrong with `lazy.layout.grow()` and `lazy.layout.shrink()`?

martanman:
The behavior of `grow/shrink` is contingent on the focused window, and as for `grow_main/shrink_main`, it does the same regardless of alignment (i.e. left/right hand side) of the main pane.

jx11r:
Oh, I see, just like `Columns` does, and that's great, but now there's no point in keeping `grow_main/shrink_main`.
I've always thought there should be standardization in some layout methods.

martanman:
I still think there is point in keeping `grow/shrink_main` because `grow_left/right` is alignment sensitive whereas the former is not.

jx11r:
But they do the same thing, i.e. they only grow/shrink the main panel...

elParaguayo:
Yes. And I use those commands and won't use the new ones. Removing those commands would break my config (and the config of other users).

libqtile/layout/xmonad.py
```diff
@@ -1024,6 +1042,19 @@ def swap_main(self):
         elif self.align == self._down:
             self.swap_left()
 
+    @expose_command()
+    def grow_up(self):
+        """Grow/shrink main pane by moving it up
+
+        Logic corresponds to `MonadTall.grow_left()`
+        """
+        self.grow_left()
```
elParaguayo, line 1051: Does the logic correspond? Aren't the values of `self.align` different in `MonadWide`?

elParaguayo:
`MonadThreeCol` inherits `MonadTall` so will get the `grow_left/right` methods. Do they work for that layout or do they need to be overriden?

libqtile/layout/xmonad.py
martanman: The logic works, yes; I've tested while changing alignment and the behavior holds.

MonadTall:

```
_left = 0
_right = 1
```

MonadWide:

```
_up = 0
_down = 1
```

That is how the two classes define their alignment so they will correspond to each other.

I could rewrite the functions in `MonadWide` using it's own alignment variables for clarity if that's preferable but it would be functionally identical.

martanman:
The functions don't exactly have an equivalence for `MonadThreeCol` so I've just overridden with pass functions to avoid confusing behavior.

martanman:
Would it make more sense if the three classes all derived from a `MonadBase` class? I've noticed that the `.flip()` method on `MonadThreeCol` don't appear to do anything either as it is.

PR #4007, opened 2022-11-18
Added ability to specify text to show when muted for Volume widget

matthewsia98:


elParaguayo:
Thanks for submitting this. However, I think this functionality is covered by #3757

PR #4017, opened 2022-11-22
Add decorator to hide commands

elParaguayo:
By default, when an object inherits from a CommandObject, any methods that were exposed on the command graph will also be exposed for the child, regardless of whether the method is explicitly decorated or not.

Sometimes, we may want methods to be hidden (e.g. when they are not appropriate for the child object). This PR adds a decorator that will hide methods from the graph.

@m-col the context for this is #4001 where the `grow_left/right` commands don't make sense for the `MonadWide` layout.

This is more of a "nice to have" so don't mind if you don't like it!

### DIFF
--- a/libqtile/command/base.py
+++ b/libqtile/command/base.py
@@ -86,6 +86,20 @@ def expose_command(name: Callable | str | list[str] | None = None) -> Callable:
     return wrapper
 
 
+def hide_command(func: Callable) -> Callable:
+    """
+    Decorator to hide a previously exposed command.
+
+    This should be used when an object inherits from an existing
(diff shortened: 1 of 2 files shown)


m-col:
I don't think this is the right solution for the issue because it makes objects and their commands on the command interface misalign with the usual expectations of subclassing and inheritance, and also from the actual methods they have available i.e. within configs. `MonadWide.grow_left` would presumably still be available and callable from within a custom function, so just not exposing it via the command interface doesn't fully resolve the issue. Perhaps a mix-in class would get the job done?

PR #4023, opened 2022-11-25
Add configurable precision and min- and max-length to Net widget

noggynoggy:
This change just touches the `_format()` function and the `defaults` list. It adds 5 new config options: `min_length`, `max_length`, `precision_down`, `precision_up` and `precision_total`. Defaults are set so that the same (old) config will produce the same (old) output.

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -53,6 +53,31 @@ class Net(base.ThreadPoolText):
         ("update_interval", 1, "The update interval."),
         ("use_bits", False, "Use bits instead of bytes per second?"),
         ("prefix", None, "Use a specific prefix for the unit of the speed."),
+        (
+            "min_length",
+            5,
+            "Minimum length of down-, up- and total-digits (decinal point included).",
+        ),
+        (
+            "max_length",
+            99999,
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for this. Does your PR do anything different from #3487 ?

noggynoggy:
If i see it correctly https://github.com/qtile/qtile/pull/3487 originated out of the same problem (the 2 decimal being hard-coded). This commit also adds min- and max-length on top which are """dynamic""" with `.rjsut()`. So basically you can do more with mine, but both should fix the core issue.

elParaguayo:
But you can do all of that with python's string formatting so I'm not sure they need to be extra defaults. 

I need to take a look at both PRs to think what's the best approach.

Also, please don't take my comments as a criticism. I'm grateful that you've taken time to want to improve qtile.

noggynoggy:
Yes you are right! I thought about doing it similarly to https://github.com/qtile/qtile/pull/3487. I wanted to make a change that doesn't affect any "old" configs but could get to work properly (now I see why.. `up`, `down` and `total` are strings in the end. That's why `:.0f` doesn't just work in the config). I am fairly new to this so no criticism taken :D

PR #4026, opened 2022-11-27
Better logging of errors in eventloop

elParaguayo:
The current exception handler just prints the exception but this does not include any reference to the function that raised the error.

Adding the `message` provides some additional information.

As an example:

Before the change:
`2022-11-27 17:52:06,647 libqtile loop.py:_handle_exception():L63  Expecting value: line 1 column 1 (char 0).`

No idea which object is causing this, or what type of error it is.

After the change:
`2022-11-27 18:00:09,051 libqtile loop.py:_handle_exception():L62  Expecting value: line 1 column 1 (char 0). Exception in callback TVHWidget._read_data(<Future finis... 1 (char 0)')>).`

However, we've still lost the fact that this is a `JSONDecodeError` which is annoying as the handler's `context` contains this:
```{'message': "Exception in callback TVHWidget._read_data(<Future finis... 1 (char 0)')>)", 'exception': JSONDecodeError('Expecting value: line 1 column 1 (char 0)'), 'handle': <Handle TVHWidget._read_data(<Future finis... 1 (char 0)')>)>}```

### DIFF
--- a/libqtile/core/loop.py
+++ b/libqtile/core/loop.py
@@ -54,13 +54,12 @@ class LoopContext(contextlib.AbstractAsyncContextManager):
         loop: asyncio.AbstractEventLoop,
         context: dict,
     ) -> None:
-        # message is always present, but we'd prefer the exception if available
         if "exception" in context:
             exc = context["exception"]
             # CancelledErrors happen when we simply cancel the main task during
             # a normal restart procedure
             if not isinstance(exc, asyncio.CancelledError):
-                logger.error(exc)
+                logger.exception("Exception in event loop:")
         else:
             logger.error("unhandled error in event loop: %s", context["msg"])
 


m-col:
nice! LGTM

elParaguayo:
Any idea how we get the exception type? That would be pretty useful too.

ervinpopescu:
Isn't it just `exc.__class__.__name__` ? :)

m-col:
@ervinpopescu put me on the right path I think (thanks!), we can certainly use `exc` but if we want to log the full traceback like we do with `logger.exception()` it turns out we can actually pass the exception for it to use manually: `logger.exception("Exception in event loop:", exc_info=exc)`. This doesn't seem to be documented, I only discovered this reading the source of the logging module :thinking:

elParaguayo:
Does that prevent the weird `NoneType: None` message that we had previously when using logger.exception?

m-col:
Yep! I think that comes from `sys.exc_info` which is `None` when not inside an exception context. Passing in `exc_info=exc` makes the logger instead use that context, so we get the same log format we do with the usual `logger.exception()` calls.

m-col:
Are you also seeing the desired exception logging for the `JSONDecodeError`? If so LGTM

elParaguayo:
It gives you the full traceback.

m-col:
Nice, definitely a huge improvement!

elParaguayo:
I'm really confused by this!

It's failing the test because flake8 says we don't need to use `exc_info`. Omitting that was what gave use the weird "NoneType: None` lines in our log but, having just tested it now, it seems to work fine.

tych0:
I think because you switched from `logger.error()` to `logger.exception()`?

elParaguayo:
We changed to `logger.error` from `logger.exception` because we were getting spurious lines in the log and no traceback (see #3891). From what I could tell from various google searches, this happens when `logger.exception` is not called from within a `try... except...` block. Setting the `exc_info` as suggested above is an obvious fix to that but, given the pep8 failure, I tried removing the `exc_info` and it still works.

m-col:
Just to I understand, we never needed to pass `exc_info` because we were always within an exception context? But `sys.exc_info` is `None` at that point, no? Why does it work as is?

m-col:
> It's failing the test because flake8 says we don't need to use exc_info.

Passing `exc_info` wasn't documented, I only discovered that from reading the source here: https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L1558-L1574. It sounds a bit like flake8 bug, as from what I can tell we are doing the "right" thing, as we want to log a full exception with traceback outside of an exception context - but maybe not because it's undocumented. I wonder what the "right" thing is if we want to log these asyncio exceptions... I can't find answers online. Regardless, happy to suppress flake8 here and move on.

elParaguayo:
To compound my frustration with this, I didn't realise that automerge was still enabled when I pushed the update!

ervinpopescu:
> I'm really confused by this!
> 
> It's failing the test because flake8 says we don't need to use `exc_info`. Omitting that was what gave use the weird "NoneType: None` lines in our log but, having just tested it now, it seems to work fine.

Mpris2 widget pollutes logs with `NoneType: None`  on pause/play. Just to be sure, you're saying here that it's not supposed to be doing that?

elParaguayo:
No. I was expecting that NoneTypr error to resurface.

Can you open a new issue for Mpris2.

PR #4035, opened 2022-12-02
Add Ability to Add Group at a specified index

Yoppler:
My use case necessitated the ability to add a group at a specific position in order for the GroupBox widget to update correctly.


elParaguayo:
Thanks for this.

Can you explain the use case for us please?

Yoppler:
What I was attempting to make "subgroups" from a key press. So if I were in group "a", and I wanted another group that would be placed after "a" as "a1". Before the change any added group just got appended to the end of the groups list so when the GroupBox updated it would be "a" "b" "a1". With it I was able to configure it to be "a" "a1" "b" in the GroupBox.

elParaguayo:
No. That's ok, I was just curious. I'll try to take a look at this over the weekend but I've got the tests running already.

PR #4037, opened 2022-12-03
Add get_mouse_position method in Wayland

elParaguayo:


m-col:
What's this used for? I see the X backend has one, but it isn't used anywhere in libqtile or the tests.

ervinpopescu:
> What's this used for? I see the X backend has one, but it isn't used anywhere in libqtile or the tests.

I used it in my custom widget once ;)

elParaguayo:
I wanted to have consistency across backends so that if users wanted to access the mouse position, they can do it in a consistent way.

I ran into this in qtile-extras where I'm trying to have a popup menu open wherever the cursor is and having to check the backend and add wayland-specific code felt unnecessary.

PR #4041, opened 2022-12-06
Warn when config duplicates Key mod+key spec

m-col:
An alternative would be to ignore the duplicate (keep the first), or support specifying multiple `Key`s and keep lists in `self.keys_map` (which is already what `self.mouse_map` does, so one can actually specify multiple `Mouse` instances which get sequentially triggered).

However, as is this does not change behaviour.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -436,8 +436,10 @@ class Qtile(CommandObject):
 
     def grab_key(self, key: Key | KeyChord) -> None:
         """Grab the given key event"""
-        keysym, mask_key = self.core.grab_key(key)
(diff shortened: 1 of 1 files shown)


elParaguayo:
> An alternative would be to ignore the duplicate (keep the first), or support specifying multiple `Key`s and keep lists in `self.keys_map` (which is already what `self.mouse_map` does, so one can actually specify multiple `Mouse` instances which get sequentially triggered).

I don't think we should support specifying multiple instances of the same key and adding to a list. Personally, if someone's defined two separate `Key`s with the same combination, I think that's a mistake. We already support binding multiple commands to the same `Key` so there's no reason to combine `Key`s.

I would there support merging this as it is.

PR #4042, opened 2022-12-07
Add 'root' as default `qtile cmd-obj` object and undocument 'cmd'.

m-col:
'cmd' is left to nothing breaks for anything, this is just a cosmetic change that aims to reduce confusion with naming things multiple ways.

I also added an example usage showing how the graph can be recursively traversed.

### DIFF
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -192,12 +192,12 @@ def add_subcommand(subparsers, parents):
         """\
     Examples:
      qtile cmd-obj
-     qtile cmd-obj -o root
+     qtile cmd-obj -o root # same as above
      qtile cmd-obj -o root -f prev_layout -a 3 # prev_layout on group 3
      qtile cmd-obj -o group 3 -f focus_back
      qtile cmd-obj -o root -f restart # restart qtile
     The graph traversal recurses:
-     qtile cmd-obj -o screen 0 -o bar bottom -f root -o screen -o group -o window -o root -f status
+     qtile cmd-obj -o screen 0 bar bottom screen group window -f info
      """
     )
     description = "Access the command interface from a shell."
@@ -213,10 +213,10 @@ def add_subcommand(subparsers, parents):
         "-o",
         dest="obj_spec",
         nargs="+",
-        default="root",
+        default=["root"],
         help="Specify path to object (space separated).  "
         "If no --function flag display available commands.  "
-        "Use `root` to specify the root node.",
+        "The root node is selected by default or you can pass `root` explicitly.",
     )
     parser.add_argument("--function", "-f", default="help", help="Select function to execute.")
     parser.add_argument(
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -93,7 +93,7 @@ def get_object(client: CommandClient, argv: list[str]) -> CommandClient:
     """
     Constructs a path to object and returns given object (if it exists).
     """
-    if argv[0] == "cmd":
+    if argv[0] in ("cmd", "root"):
         argv = argv[1:]
 
     # flag noting if we have consumed arg1 as the selector, eg screen[0]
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks. Will take a proper look tonight.

libqtile/scripts/cmd_obj.py
```diff
@@ -192,11 +192,12 @@ def add_subcommand(subparsers, parents):
...
-     qtile cmd-obj -o cmd
-     qtile cmd-obj -o cmd -f prev_layout -i
-     qtile cmd-obj -o cmd -f prev_layout -a 3 # prev_layout on group 3
+     qtile cmd-obj -o root
+     qtile cmd-obj -o root -f prev_layout -a 3 # prev_layout on group 3
      qtile cmd-obj -o group 3 -f focus_back
-     qtile cmd-obj -o cmd -f restart # restart qtile
+     qtile cmd-obj -o root -f restart # restart qtile
+    The graph traversal recurses:
+     qtile cmd-obj -o screen 0 -o bar bottom -f root -o screen -o group -o window -o root -f status
```
elParaguayo, line 200: This line look wrong to me. To traverse the graph I'd just do one `-o` argument e.g. `qtile cmd-obj -o screen 0 bar bottom`.

There's no way to return to the root node from any other object so, if you actually get an output of `"OK"` here it's because it's running `status` on just the `-o root` part and it's ignoring everything else.

libqtile/scripts/cmd_obj.py
```diff
@@ -212,9 +213,10 @@ def add_subcommand(subparsers, parents):
         "-o",
         dest="obj_spec",
         nargs="+",
+        default="root",
         help="Specify path to object (space separated).  "
         "If no --function flag display available commands.  "
-        "Use `cmd` to specify root command.",
+        "Use `root` to specify the root node.",
```
elParaguayo, line 219: Do you want to mention that, by omitting it, the root node is selected?

elParaguayo:
Might need to update `test/test_qtile_cmd.py` too.

elParaguayo:
I'd make these changes (can push them directly if you'd prefer):
``` diff
diff --git a/libqtile/scripts/cmd_obj.py b/libqtile/scripts/cmd_obj.py
index 022a034d..b3916248 100644
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -192,12 +192,12 @@ def add_subcommand(subparsers, parents):
         """\
     Examples:
      qtile cmd-obj
-     qtile cmd-obj -o root
+     qtile cmd-obj -o root # same as above
      qtile cmd-obj -o root -f prev_layout -a 3 # prev_layout on group 3
      qtile cmd-obj -o group 3 -f focus_back
      qtile cmd-obj -o root -f restart # restart qtile
     The graph traversal recurses:
-     qtile cmd-obj -o screen 0 -o bar bottom -f root -o screen -o group -o window -o root -f status
+     qtile cmd-obj -o screen 0 bar bottom screen group window -f info
      """
     )
     description = "qtile.command functionality exposed to the shell."
@@ -213,10 +213,10 @@ def add_subcommand(subparsers, parents):
         "-o",
         dest="obj_spec",
         nargs="+",
-        default="root",
+        default=["root"],
         help="Specify path to object (space separated).  "
         "If no --function flag display available commands.  "
-        "Use `root` to specify the root node.",
+        "The root node is selected by default or you can pass `root` explicitly.",
     )
     parser.add_argument("--function", "-f", default="help", help="Select function to execute.")
     parser.add_argument(
diff --git a/test/test_qtile_cmd.py b/test/test_qtile_cmd.py
index e10d5c98..0a513bfc 100644
--- a/test/test_qtile_cmd.py
+++ b/test/test_qtile_cmd.py
@@ -89,13 +89,16 @@ class ServerConfig(Config):
 server_config = pytest.mark.parametrize("manager", [ServerConfig], indirect=True)


-def run_qtile_cmd(args):
+def run_qtile_cmd(args, no_eval=False):
     cmd = os.path.join(os.path.dirname(__file__), "..", "bin", "qtile")
     argv = [cmd, "cmd-obj"]
     argv.extend(args.split())
     pipe = subprocess.Popen(argv, stdout=subprocess.PIPE)
     output, _ = pipe.communicate()
-    return eval(output.decode())  # as returned by pprint.pprint
+    output = output.decode()
+    if no_eval:
+        return output
+    return eval(output)  # as returned by pprint.pprint


 @server_config
@@ -141,7 +144,7 @@ def test_qtile_cmd(manager):
 def test_display_kb(manager):
     from pprint import pprint

-    cmd = "-s {} -o cmd -f display_kb".format(manager.sockfile)
+    cmd = "-s {} -o root -f display_kb".format(manager.sockfile)
     table = run_qtile_cmd(cmd)
     print(table)
     pprint(table)
@@ -156,3 +159,12 @@ def test_display_kb(manager):
     assert re.search(r"(?m)^named\s{3,}b\s{9,}togroup\('b'\)\s*$", table)
     assert re.search(r"(?m)^named>_\s{3,}a\s{9,}togroup\('a'\)\s*$", table)
     assert re.search(r"(?m)^<root>\s{3,}y\s{9,}\s*$", table) is None
+
+
+@server_config
+def test_cmd_obj_root_node(manager):
+    base = f"-s {manager.sockfile} -f ok"
+    cmd_no_root = base
+    cmd_with_root = f"{base} -o root"
+
+    assert run_qtile_cmd(cmd_no_root, no_eval=True) == run_qtile_cmd(cmd_with_root, no_eval=True)
```

NB by having root as the default node, you now need to explicitly call `-h` if you want to see the help.

elParaguayo:
@m-col this is lower priority than other things you're looking at but I'm still in favour of merging this once we've addressed the above points.

elParaguayo:
This one really should get merged so I've rebased and pushed my recommended changes. If tests pass I'll squash down to a single commit.

PR #4044, opened 2022-12-08
Added toggle for icon in Cmus widget

2e7:
Default behavior unchanged. User can set `icon = False` to disable the music note icon.


elParaguayo:
Thanks for submitting this.

Instead of having a flag to toggle that icon on and off, I think a better way to fix this would be to create a `format` parameter which describes what the output should look like and users can edit that via their config. This would give users much more flexibility over the appearance of the widget.

elParaguayo:
I don't love those changes. I would prefer to have a `format` parameter as this is consistent with other widgets. The default value would be something like `{play_icon}{artist} - {title}` (`play_icon` would include the extra space if users wanted a gap).

I also don't understand the space in the dictionary key names. Can you explain why that's needed?

2e7:
Re: Spaces - Introducing `albumartist` led to errors in scraping, but I should have thought to just add one space to the if statement instead of all of the spaces to the keys...

The specific direction is greatly appreciated, I'll take another crack at it from the top with a proper eye to other widgets.

2e7:
It looks like I made all of this way, way more complicated than it needed to be.

m-col:
The changes look good (from what I can tell there aren't any outstanding changes you want to make?). You also need to update the Cmus widget tests. Specifically, these ones are failing:
```
  FAILED test/widgets/test_cmus.py::test_cmus_buttons[1-x11] - AssertionError: ...
  FAILED test/widgets/test_cmus.py::test_cmus_buttons[1-wayland] - AssertionErr...
  FAILED test/widgets/test_cmus.py::test_cmus_play_stopped - AssertionError: as...
  FAILED test/widgets/test_cmus.py::test_missing_metadata - AssertionError: ass...
```

2e7:
Went back and modified the widget so the icon appears as the default was before.

m-col:
It might also be worth adding the new `format` option to the changelog, this way others will see your new feature :)

PR #4046, opened 2022-12-10
doc: Fix some typos on Group documentation

hiagofranco:
Fix some errors on config.py Group documentation.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -640,7 +640,7 @@ class Group:
     exclusive:
         When other apps are started in this group, should we allow them here or not?
     spawn:
-        This will be ``exec()``uted when the group is created. You can pass either a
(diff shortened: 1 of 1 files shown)


libqtile/config.py
```diff
@@ -640,8 +640,8 @@ class Group:
     exclusive:
         When other apps are started in this group, should we allow them here or not?
     spawn:
-        This will be ``exec()`` d when the group is created. Tou can pass either a
-        program name or a list of programs to ``exec()``
+        This will be ``exec()``uted when the group is created. You can pass either a
```
elParaguayo, line 643:
```suggestion
        This will be executed (via ``qtile.spawn()``) when the group is created. You can pass either a
```

elParaguayo:
Thanks. One minor change suggested (as I don't like `"exec()"d` and `"exec()"uted`!)

PR #4050, opened 2022-12-11
WidgetBox: add box_is_open default

mrtnvgr:
This Pull Request adds a `box_is_open` "default" in widgetbox, so widgetbox can be opened by default.
My setup:
```
widget.WidgetBox(widgets=[
                    ...
                ], foreground=foreground, ..., box_is_open=True),
```
When I launch qtile, half of the widgets in the box are not displayed.

Can anyone help me solve this bug? I can't figure out on my own.

### DIFF
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -67,13 +67,13 @@ class WidgetBox(base._Widget):
         ),
         ("text_closed", "[<]", "Text when box is closed"),
         ("text_open", "[>]", "Text when box is open"),
+        ("box_is_open", False, "Default box state"),
         ("widgets", list(), "A list of widgets to include in the box"),
     ]
 
     def __init__(self, _widgets: list[base._Widget] | None = None, **config):
         base._Widget.__init__(self, bar.CALCULATED, **config)
         self.add_defaults(WidgetBox.defaults)
-        self.box_is_open = False
         self.add_callbacks({"Button1": self.toggle})
 
         if _widgets:
@@ -124,6 +124,10 @@ class WidgetBox(base._Widget):
         # Disable drawing of the widget's contents
         for w in self.widgets:
             w.drawer.disable()
+
+        if self.box_is_open:
+            self.box_is_open = False
+            self.toggle()
 
     def calculate_length(self):
         return self.layout.width
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -179,6 +179,3 @@ class WidgetBox(base._Widget):
         self.toggle_widgets()
         self.set_box_label()
         self.bar.draw()
-
-    def finalize(self):
-        base._Widget.finalize(self)
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from libqtile import bar
+from libqtile import bar, hook
 from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
 from libqtile.widget import Systray, base
@@ -68,6 +68,7 @@ class WidgetBox(base._Widget):
         ("text_closed", "[<]", "Text when box is closed"),
         ("text_open", "[>]", "Text when box is open"),
         ("widgets", list(), "A list of widgets to include in the box"),
+        ("start_opened", False, "Open the box at startup"),
     ]
 
     def __init__(self, _widgets: list[base._Widget] | None = None, **config):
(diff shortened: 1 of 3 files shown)


elParaguayo:
It's going to be a bit messy to have the box opened when configuring the widget because the widget has to insert its contents into the bar.

I wonder if an easier approach would be something like this:
``` diff
diff --git a/libqtile/widget/widgetbox.py b/libqtile/widget/widgetbox.py
index a5e88e07..2264688f 100644
--- a/libqtile/widget/widgetbox.py
+++ b/libqtile/widget/widgetbox.py
@@ -21,7 +21,7 @@ from __future__ import annotations

-from libqtile import bar
+from libqtile import bar, hook
 from libqtile.command.base import expose_command
 from libqtile.widget import Systray, base
@@ -68,6 +68,7 @@ class WidgetBox(base._Widget):
         ("text_closed", "[<]", "Text when box is closed"),
         ("text_open", "[>]", "Text when box is open"),
         ("widgets", list(), "A list of widgets to include in the box"),
+        ("start_opened", False, "Open the box at startup"),
     ]

     def __init__(self, _widgets: list[base._Widget] | None = None, **config):
@@ -89,6 +90,9 @@ class WidgetBox(base._Widget):
             logger.warning("Invalid value for 'close_button_location': %s", val)
             self.close_button_location = "left"

+        if self.start_opened:
+            hook.subscribe.startup(self.toggle)
+

@@ -175,3 +179,8 @@ class WidgetBox(base._Widget):
         self.toggle_widgets()
         self.set_box_label()
         self.bar.draw()
+
+    def finalize(self):
+        if self.start_opened:
+            hook.unsubscribe.startup(self.toggle)
+        base._Widget.finalize(self)
```

This will just call the `toggle` function when qtile has started and so takes care of all the logic about inserting widgets etc.

mrtnvgr:
@elParaguayo Can I use ```hook.subscribe.startup_once``` instead of ```hook.subscribe.startup```? If I understood right, this function won't require unsubscribe?

elParaguayo:
> @elParaguayo Can I use `hook.subscribe.startup_once` instead of `hook.subscribe.startup`? If I understood correctly, this function won't require unsubscribe?

Yes. You can do that.

elParaguayo:
You've got a conflict in CHANGELOG which is because I recently merged something that added a line to that file so you'll need to rebase your branch on the latest master.

elParaguayo:
Looks like you've got some tabs in CHANGELOG.

Also, did you rebase your branch or did you just copy the existing CHANGELOG line into your commit?

libqtile/widget/widgetbox.py
```diff
@@ -175,3 +179,6 @@ def toggle(self):
         self.toggle_widgets()
         self.set_box_label()
         self.bar.draw()
+
+    def finalize(self):
+        base._Widget.finalize(self)
```
elParaguayo, line 184: You can delete these lines if we're no longer unsubscribing the hook.

mrtnvgr:
> Also, did you rebase your branch or did you just copy the existing CHANGELOG line into your commit?

I rebased it

elParaguayo:
Nice. This looks good. I assume the startup hook does work as expected here?

mrtnvgr:
> Nice. This looks good. I assume the startup hook does work as expected here?

Yep

m-col:
Does this setting work when reloading the config without a restart? I'm wondering if instead of using the hook, `self.toggle()` needs to be called within `_configure()`.

elParaguayo:
I wouldn't think you want the box to open when reloading config, just the first time qtile starts.

m-col:
If it's a default state, wouldn't reloading want it to return to that default state? Reloading replaces the existing widget with a new instantiation, so in the case where the widget's configuration didn't change, it will appear to have closed.

elParaguayo:
Ah, ok. I see your point.

Ideally we'd preserve state on a reload rather than revert to default. I doubt we're going to do that so handling this in configure is probably a better solution...

m-col:
I don't think we should preserve state at this level of granularity, it _is_ a new widget instance after all, and the benefit is small compared to the added complexity.

elParaguayo:
I agree, far too complicated!

I'll make a PR to make the change.

mrtnvgr:
> I'll make a PR to make the change.

I would like the current behaviour. Maybe we can keep current implementation as `start_opened_at_startup`?

elParaguayo:
@mrtnvgr - if you only want it opened the very first time at startup you would still be able to do that by setting `start_opened` to False and then using a hook in your config:
``` python
from libqtile import qtile

@hook.subscribe.startup
def open_widgetbox():
    qtile.widgets_map["widgetbox"].toggle()  # or .cmd_toggle() if you're on 0.22.1
```

PR #4053, opened 2022-12-11
Add qtile/qtile#3820 to changelog

mrtnvgr:
This Pull Request adds a note about qtile/qtile#3820 to CHANGELOG file.


elParaguayo:
Why add this now? I've no problem with it, just curious!

mrtnvgr:
> Why add this now? I've no problem with it, just curious!

I forgot to do this then :)

PR #4054, opened 2022-12-11
Add new widget to get Ethernet status

hiagofranco:
The goal of this widget is to show ethernet status on Linux using only sysfs. It will show if the cable is connected or not or if the link is up or not. Based on Wlan widget.

### DIFF
--- a/libqtile/widget/eth.py
+++ b/libqtile/widget/eth.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2012 Sebastian Bechtel
+# Copyright (c) 2013 Tao Sauvage
+# Copyright (c) 2014 Sebastian Kricner
+# Copyright (c) 2014 Sean Vig
+# Copyright (c) 2014 Tycho Andersen
+# Copyright (c) 2014 Craig Barnes
+# Copyright (c) 2015 farebord
+# Copyright (c) 2015 Jörg Thalheim (Mic92)
+# Copyright (c) 2016 Juhani Imberg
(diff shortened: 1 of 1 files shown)


hiagofranco:
I was missing this widget in the main repo, since I often used the Ethernet cable and I'm also connected to the WiFi. I'm not a python expert, so I'm sorry if something is wrong. Hope this PR helps!

elParaguayo:
It really feels like we should have a combined wifi and ethernet widget rather than two separate ones.

Thoughts?

hiagofranco:
thanks for the comment @elParaguayo. 

I this case, we could do something similar to the other DE's like Gnome or Cinnamon: If you're using WiFi only, the WiFi icon appears, otherwise if you connect the Ethernet cable too, the cable connection indication appears.

In this case, Wlan widget would receive both interfaces names (as they are not default) and check first for the cable connection, as I did in this PR. If it's true, the cable connection icon/string will show. If something is wrong, like the cable is not connected or the interface name is wrong, we will show only the WiFi information. This will also prevent this new feature breaking some configs that the user didn't pass the Ethernet interface name correctly.

Let me know if this makes sense, so I can close this PR and open a new one with these modifications.

elParaguayo:
I do think a combined widget makes sense. However, there may well be some people who just want to see wifi or ethernet so, ideally, configuration of the combined widget could allow that possibility too.

hiagofranco:
OK, got it. It makes sense to me. Therefore, I'm closing this PR and I'll open a new one when I have something done, so we can continue this discussion. Thanks!

PR #4055, opened 2022-12-11
Add ability to move current window to Slice

elParaguayo:
Currently, only windows matching the defined rule can be added to the single layout in `Slice`. This PR adds a new `move_to_slice` command to move the current window to the slice.


m-col:
Neat. LGTM

PR #4059, opened 2022-12-12
Fix CPU widget for psutil bug

elParaguayo:
This is to replace #3503. I didn't rebase that PR as it's on that user's `master` branch and I'm not comfortable force pushing to that.

This addresses the issue in psutil 5.9.0 where the cpu frequency is provided with the wrong precision.


m-col:
Nice

PR #4061, opened 2022-12-13
add YandexDisk widget

mrtnvgr:
This PR adds a widget for [Yandex.Disk](https://yandex.com/support/disk-desktop-linux/) client.

I have a question. How do I make fake status files in tests?

### DIFF
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -24,7 +24,7 @@ from libqtile.widget import base
 
 
 class YandexDisk(base.InLoopPollText):
-    """A simple widget to show YandexDisk folder sync status.
+    """A simple widget to show YandexDisk client folder sync status.
 
     Yandex.Disk_ is a service that lets you store files on Yandex servers.
 
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -23,6 +23,13 @@ import os
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
+status_mapping = {
+    "idle": "IDLE",
+    "stopped": "STOPPED",
+    "busy": "BUSY",
+    "index": "INDEX",
+}
+
 
 class YandexDisk(base.InLoopPollText):
     """A simple widget to show YandexDisk client folder sync status.
@@ -35,11 +42,7 @@ class YandexDisk(base.InLoopPollText):
 
     defaults = [
         ("sync_folder", "~/Yandex.Disk/", "Yandex.Disk folder path"),
-        (
-            "daemon_stopped",
-            "Error: daemon not started",
-            "Text when Yandex.Disk daemon is stopped",
-        ),
+        ("status_mapping", status_mapping, "Sync status mapping"),
         ("update_interval", 10, "The delay in seconds between updates"),
         ("format", "{status}", "Display format"),
     ]
@@ -58,15 +61,18 @@ class YandexDisk(base.InLoopPollText):
 
     def _get_status(self):
         with open(self.sync_folder, "r") as file:
-            return file.read().split("\n")[-1]
+            status = file.read().split("\n")[-1]
+            return status
 
     def poll(self):
         try:
             status = self._get_status()
         except FileNotFoundError:
-            return self.daemon_stopped
+            status = "stopped"
         except Exception:
             logger.exception("Error getting status for yandex.disk")
             return "Error"
+
+        status = self.status_mapping.get(status, status)
 
         return self.format.format(status=status)
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -58,6 +58,8 @@ class YandexDisk(base.InLoopPollText):
     def poll(self):
         try:
             status = self._get_status()
+        except FileNotFoundError:
+            return "Error: daemon not started"
         except Exception:
             logger.exception("Error getting status for yandex.disk")
             return "Error"
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -58,6 +58,7 @@ class YandexDisk(base.InLoopPollText):
         try:
             status = self._get_status()
         except Exception as e:
-            return f"Error: {e}"
+            logger.exception("Error getting status for yandex.disk")
+            return "Error"
 
         return self.format.format(status=status)
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -110,13 +110,13 @@ class YandexDisk(base.InLoopPollText):
 
     @staticmethod
     def _get_progress_log_dict(log_data):
-        file_size = int(log_data.pop(-1))
+        total_size = int(log_data.pop(-1))
         log_data.pop(-1)
         synced_size = int(log_data.pop(-1))
         filename = " ".join(log_data)
         return {
             "filename": filename,
             "synced_size": synced_size,
-            "file_size": file_size,
-            "percentage": int((synced_size / file_size) * 100),
+            "total_size": total_size,
+            "percentage": int((synced_size / total_size) * 100),
         }
--- a/libqtile/widget/yandexdisk.py
+++ b/libqtile/widget/yandexdisk.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 
 import os
+import shlex
 
 from libqtile.log_utils import logger
 from libqtile.widget import base
@@ -48,7 +49,7 @@ class YandexDisk(base.InLoopPollText):
         ("status_mapping", status_mapping, "Sync status mapping"),
         ("update_interval", 5, "The delay in seconds between updates"),
         ("format", "{status}{progress}", "Display format"),
-        ("progress_format", " ({filename} {percentage:.1%}%)", "Progress format"),
+        ("progress_format", " ({filename} {percentage:.1%})", "Progress format"),
     ]
 
     def __init__(self, **config):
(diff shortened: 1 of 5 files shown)


mrtnvgr:
Why the docs/readthedocs tests failed?

mrtnvgr:
> Why the docs/readthedocs tests failed?

@elParaguayo 

```
[rtd-command-info] start-time: 2022-12-13T15:28:36.746084Z, end-time: 2022-12-13T15:28:36.831556Z, duration: 0, exit-code: 128
git checkout --force aaf8c5553b61156f11a3a2c4f77c821f5d27f03f
fatal: reference is not a tree: aaf8c5553b61156f11a3a2c4f77c821f5d27f03f
```

test/widgets/test_yandexdisk.py
```diff
@@ -0,0 +1,43 @@
...
+    tmp = "/var/tmp/qtile/test/widgets/yandexdisk"
+    sync_folder = os.path.join(tmp, "Yandex.Disk")
+
+    status_file = os.path.join(sync_folder, ".sync", "status")
+    os.makedirs(os.path.dirname(status_file), exist_ok=True)
+
+    with open(status_file, "w") as f:
+        f.write("99999\nidle")
+
+    yield sync_folder
```
elParaguayo, line 38:
```suggestion
@pytest.fixture
def yandexdisk_folder():
    with tempfile.TemporaryDirectory() as tmp:
        sync_folder = os.path.join(tmp, "Yandex.Disk")

        status_file = os.path.join(sync_folder, ".sync", "status")
        os.makedirs(os.path.dirname(status_file), exist_ok=True)

        with open(status_file, "w") as f:
            f.write("99999\nidle")

        yield sync_folder
```

Slight tweak. Using `tempfile` should ensure that the path exists (this is better than hardcoding an assumed path). Also, by using the context manager (`with ...`) the folder will be deleted after the test.

You'll also need to `import tempfile`

libqtile/widget/yandexdisk.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _get_status(self):
+        with open(self.sync_folder, "r") as file:
+            return file.read().split("\n")[-1]
+
+    def poll(self):
+        try:
+            status = self._get_status()
+        except Exception as e:
+            return f"Error: {e}"
```
elParaguayo, line 61:
```suggestion
            logger.exception("Error getting status for yandex.disk")
            return "Error"
```
You don't want to put the whole exception in the widget contents.

You'll need to do `from libqtile.log_utils import logger` too.

elParaguayo (commented):
I've added a couple of comments/suggestions.

However, my main question is whether this is something that should be included in the main repo. I've no idea how popular yandex.disk is. This is something that could be achieved very easily with the `GenPollText` widget and there's nothing here that really means this needs to be a separate widget.

mrtnvgr:
> However, my main question is whether this is something that should be included in the main repo. I've no idea how popular yandex.disk is. This is something that could be achieved very easily with the `GenPollText` widget and there's nothing here that really means this needs to be a separate widget.

Almost half of my country uses Yandex and its services as a replacements for Google's ones. Such as Search Engine, Google Maps(Yandex Maps), Google Drive(Yandex.Disk), GMail (Yandex Mail)  etc...

Search engine usage statistics:

mrtnvgr:
I think I will rewrite this widget. Instead of checking in status file, I will use core.log file events.

elParaguayo:
Nice. That's looking pretty good.

Any chance you could add some screenshots here? see https://qtile--4061.org.readthedocs.build/en/4061/manual/howto/widget.html#screenshots

Should be pretty quick as you can import the fixture from your test.

mrtnvgr:
> Any chance you could add some screenshots here? see http://docs.qtile.org/en/stable/manual/howto/widget.html#screenshots

Can you help me please?

test/widgets/docs_screenshots/ss_yandexdisk.py
```diff
@@ -0,0 +1,34 @@
...
+from test.widgets.test_yandexdisk import yandexdisk_folder
+
+
+@pytest.fixture
+def widget(monkeypatch):
+    monkeypatch.setattr("libqtile.widget.yandexdisk.YandexDisk.sync_folder", yandexdisk_folder)
+    yield yandexdisk.YandexDisk
+
+
+@pytest.mark.parametrize("screenshot_manager", indirect=True)
```
elParaguayo, line 32:
```suggestion
@pytest.mark.parametrize("screenshot_manager", [{}], indirect=True)
```
You have to pass the configuration for the widget, even if it's the default i.e. `{}`.

elParaguayo:
Not sure that quite worked:
image

Don't think we meant to get `N/A` here.

elParaguayo:
That is very odd but almost certainly not linked to anything in your widget. I'll take a look later.

elParaguayo:
I'll also take a look at that screenshot later. Pretty sure it's the monkeypatching that's the issue as your path will be overwritten by the default.

I've fixed this on other tests but don't remember how I did it but can check later.

elParaguayo:
Don't worry about the screenshot for now. Let me try to fix it later tonight if I've got time.

mocksoul:
dont get me rude, but adding such a niche widget to the main repo is not good idea i think.

you can sync yandex disk with rclone (as well as a **ton** of other clouds https://rclone.org/#providers !sic). RClone widget could be much more universal in that case. But even rclone widget should not be in the main repo in my (humble) opinion.

elParaguayo:
Being niche is not a reason for it not to be included. However, a broader cloud storage solution would be preferable.

mrtnvgr:
> dont get me rude, but adding such a niche widget to the main repo is not good idea i think.

I understand your point but don't we already have niche widgets in main repo? For example, GmailChecker. And, in my opinion, support for official client is better but if @elParaguayo wants I can close this PR and start working on rclone widget.

mrtnvgr:
I think that rclone is useful when you use a lot of clouds, but when I have only one cloud, I want the most minimalistic software without support for other clouds. YandexDisk cli client is convenient and works very well

mrtnvgr:
When I change some value in `status_mapping` and reload qtile config, I get this error:
```
2022-12-15 15:20:15,491 ERROR libqtile loop.py:_handle_exception():L62 Exception in event loop:
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/core/manager.py", line 898, in f
    func(*args)
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/widget/base.py", line 734, in timer_setup
    update_interval = self.tick()
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/widget/base.py", line 761, in tick
    self.update(text)
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/widget/base.py", line 702, in update
    self.text = text
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/widget/base.py", line 484, in text
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/drawer.py", line 72, in text
  File "/home/user/.local/lib/python3.10/site-packages/libqtile/pangocffi.py", line 186, in parse_markup
    raise Exception("parse_markup() failed for %s" % value)
Exception: parse_markup() failed for b'\xe2\x98\x81\xef\xb8\x8f: >.<'
```

mrtnvgr:
> When I change some value in `status_mapping` and reload qtile config, I get this error:

I solved it, I escaped "<" and ">" symbols. In Widgetbox widget, you don't need to escape anything (at least in `text_open` and `text_closed` parameters), so I thought that escaping won't solve the issue.

mocksoul:
> I think that rclone is useful when you use a lot of different clouds, but when I use only one, I want the most minimalistic software without support for other clouds. YandexDisk cli client is convenient and works very well for me

yandex disk is closed source, rclone is open source. rclone is much more reliable in my setup, I did used yandex disk before and had bad experience watching how yandex disk rereads file hashes again and again. Rclone allows both unidirectional and bidirectional sync (latter is kind of experimental, but it works without problems).

you should give it a try :)

mocksoul:
> I understand your point but don't we already have niche widgets in main repo?

if something done badly before that does not mean everyone should continue to do the same :). The main difference is there support is coming from. If something sits in the main repo - qtile team should support it. If it sits as 3rd party widget - widget author himself supports it and qtile team provides a list of 3rd party widgets if they supported well and still work.

im not affiliated with qtile team, this is just my 5 cents.

elParaguayo:
> > I understand your point but don't we already have niche widgets in main repo?
> 
> if something done badly before that does not mean everyone should continue to do the same :). 

Niche is not the same as bad and can actually be the opposite.

Anyway, back to the main point, if there is a way to have widget that supports multiple cloud storage providers then that would definitely be preferable to including a single provider.

@mrtnvgr - you've done a lot of work on this and I'd hate to see that effort go to waste so, if you want, you could submit this to qtile-examples instead. That repo is a place for users to share things they've made but there's no support from the development team.

elParaguayo:
> > When I change some value in `status_mapping` and reload qtile config, I get this error:
> 
> I solved it, I escaped "<" and ">" symbols. In Widgetbox widget, you don't need to escape anything (at least in `text_open` and `text_closed` parameters), so I thought that escaping won't solve the issue.

markup is enabled by default but you can disable it by passing `markup=False` in your widget's configuration in your config file. I wonder if that's also related to the `-.-` issue too.

mrtnvgr:
> markup is enabled by default but you can disable it by passing `markup=False` in your widget's configuration in your config file. I wonder if that's also related to the `-.-` issue too.

This does not solve the issue.

elParaguayo:
> > markup is enabled by default but you can disable it by passing `markup=False` in your widget's configuration in your config file. I wonder if that's also related to the `-.-` issue too.
> 
> This does not solve the issue.

Which one? Or both of them?

mrtnvgr:
> you should give it a try :)

Thanks for the suggestion, I will definitely try it!

mrtnvgr:
> > > markup is enabled by default but you can disable it by passing `markup=False` in your widget's configuration in your config file. I wonder if that's also related to the `-.-` issue too.
> > 
> > This does not solve the issue.
> 
> Which one? Or both of them?

The `-.-` issue

mrtnvgr:
> The `-.-` issue

Now I will close this PR and open a new one in qtile-examples as you said. Let's discuss this issue in #4066

PR #4065, opened 2022-12-14
libqtile: configreader: ensure config is not evaluated twice during reload fixes #4064

### LINKED ISSUE #4064: All hooks in config will be subscribed and fired twice.
### The issue:

~~This is probably due to configuration testing syntax step.~~

```python
'startup_complete': [<function xstartup_complete at 0x7f2005fc49d0>,
                      <function xstartup_complete at 0x7f2005fc5510>]}
```

Code to reproduce is simple:

in config:

```python
@hook.subscribe.startup_complete
def xstartup_complete():
    ...

logger.warn(pprint.pformat(hook.subscriptions))
```

All hooks are actually being fired twice, not only startup but all for each event.

### Required:

mocksoul:
fixes #4064

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -108,6 +108,14 @@ class Config:
             if hasattr(module, "__file__") and module.__file__ is not None:
                 subpath = Path(module.__file__)
 
+                if subpath == path:
+                    # do not reevaluate config itself here, we want only
+                    # reload all submodules. Also we cant reevaluate config
+                    # here, because it will cache all current modules before they
+                    # are reloaded. Thus, config file should be reloaded after
+                    # this routine.
+                    continue
+
                 # Check if the module is in the config folder or subfolder
                 # if so, reload it
                 if folder in subpath.parents:
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -34,6 +34,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
         - Fix `CPU` precision bug with specific version of `psutil`
+        - Fix config being reevaluated twice during reload (e.g. all hooks from config were doubled)
     * python version support
         - We have added support for python 3.11 and pypy 3.9.
         - python 3.7, 3.8 and pypy 3.7 are not longer supported.
--- a/test/configs/reloading.py
+++ b/test/configs/reloading.py
@@ -88,4 +88,6 @@ if hasattr(qtile, "test_data"):
     floating_layout = layout.Floating()
     wmname = "TEST"
 
+    qtile.test_data_config_evaluations += 1
+
 groups.append(ScratchPad("S", dropdowns))
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -1167,7 +1167,10 @@ def test_reload_config(manager_nospawn):
 
     # Reload #1 - with libqtile.qtile.test_data
     manager_nospawn.c.eval("self.test_data = 1")
+    manager_nospawn.c.eval("self.test_data_config_evaluations = 0")
     manager_nospawn.c.reload_config()
+    # should be readed twice (check+read), but no more
+    assert manager_nospawn.c.eval("self.test_data_config_evaluations") == (True, "2")
     assert manager_nospawn.c.eval("len(self.keys_map)") == (True, "2")
     assert manager_nospawn.c.eval("len(self.mouse_map)") == (True, "2")
     assert "".join(manager_nospawn.c.get_groups().keys()) == "123456789S"
@@ -1192,6 +1195,7 @@ def test_reload_config(manager_nospawn):
 
     # Reload #2 - back to without libqtile.qtile.test_data
     manager_nospawn.c.eval("del self.test_data")
+    manager_nospawn.c.eval("del self.test_data_config_evaluations")
     manager_nospawn.c.reload_config()
     assert manager_nospawn.c.eval("len(self.keys_map)") == (True, "1")
     assert manager_nospawn.c.eval("len(self.mouse_map)") == (True, "1")


mocksoul:
I'll also put test here

mocksoul:
looks like its done.
but I cant run workflows here ("first-time contributor")

libqtile/confreader.py
```diff
@@ -108,6 +108,14 @@ def _reload_config_submodules(self, path: Path) -> None:
             if hasattr(module, "__file__") and module.__file__ is not None:
                 subpath = Path(module.__file__)
 
+                if subpath == path:
+                    # do not reevaluate config itself here, we want only
+                    # reload all submodules. Also we cant reevaluate config
+                    # here, because it will cache all current modules before they
+                    # are reloaded. Thus, config file should be reloaded after
+                    # this routine.
```
m-col, line 116: As far as I can tell, this issue still happens with other submodules. For example, let's say I have these 3 files, which contain these imports:

- b.py
- a.py <- imports b.py
- config.py <- imports a.py

If we we first reload a.py before b.py, it will import outdated data from the cached b.py. Then we reload b.py, but that new data isn't used anywhere because a.py was already reloaded.

Reading the importlib docs I can't see any helpers we can use to recursively reload the entire folder. From some reading through, it looks like we might be best invalidating the caches recursively in the config folder, and then reloading just the main config file. However, I think this would require defining a custom Loader... I don't know how simple or complex that would be though. Thoughts?

mocksoul: in submodules, especially in 3rd-party like json/etc, there should be no logic in global namespace. So, reimporting them 100500 times does nothing bad.

in contrast in config.py all logic sits in global namespace, thats why evaluating in >1 times is so harmful.

personally I dont even think reloading anything but config.py itself is good idea. Or maybe everything in conf folder but nothing else. That can be done easily but just removing all modules from config folder and just reimport/reevaluate config itself. But I'd consider full restart in that case anyway, especially knowing the fact it does not require to stop/restart X server.

for this fix alone - it is concentrated on config.py quick reloading and probably will be the same in any complex logic you will try to implement :)) (please, dont :-))

elParaguayo: Sorry, not sure I completely follow what you're saying but I am tired! What are you saying "please don't" to?

We are only reloading files that are in the config folder (or a subfolder of that). We do need to support this as many users are now splitting their configs into subfiles. Restarting qtile would be simple but seems excessive.

m-col: Note restarting isn't supported at all when using the Wayland backend.

To be clear, I think this change is fine because it reduces a bit of unnecessary work but functionally it doesn't change anything. The issue I describe is an existing bug that remains after this PR, which is fine, but we should definitely fix that.

elParaguayo: Hmm. Aren't there also going to be scenarios where we need to reload submodules before we test whether we can reload the config? e.g. if a new variable had been defined in a submodule and imported into the top config, wouldn't this crash without reloading the submodule first?

m-col: This change ensures that submodules within the config dir are reloaded before the main config.py, just not that submodules are reloaded in whatever order they import each other. Users who split their configs are fine _only_ if submodules don't import other submodules.

elParaguayo: Ignore my last comment, it was wrong.

m-col:
There is another place we're loading the config (as we test the config is valid at the top of `reload_config`) but I have some changes I'm going to do that will remove that as well. Will submit soon.

Thanks for this!

libqtile/confreader.py
mocksoul: > Note restarting isn't supported at all when using the Wayland backend.

this does not work with wayland backend even replacing process by itself (via execve) ? Didnt knew that =(

ervinpopescu: > This change ensures that submodules within the config dir are reloaded before the main config.py, just not that submodules are reloaded in whatever order they import each other. Users who split their configs are fine _only_ if submodules don't import other submodules.

This is my case :) What will this introduce for me?

elParaguayo: :bomb: -> :boom:

:smile:

elParaguayo: Joking aside, the issues are:
1) Changed variables in submodules may not propagate correctly
2) Hooks in submodules could be subscribed multiple times
3) Probably some other unintended consequences (m-col's evidently thought about this more than I have!)

m-col: I don't think the hooks issue will persist fortunately because each submodule is only reloaded once. It's mainly just that the order of submodules getting reloaded isn't necessarily the order that they need to get reloaded to ensure the imports between them are pulling from yet-to-be-reloaded cached submodules. But I'm hoping there is a canonical way to handle this using python's own machinery rather than hand-rolling a solution.

ervinpopescu: >     1. Changed variables in submodules may not propagate correctly

>     2. Hooks in submodules could be subscribed multiple times
> 
>     3. Probably some other unintended consequences (m-col's evidently thought about this more than I have!)

ervinpopescu: Yeah, this messes things up for me... 
```
2022-12-19 21:52:34,175 ERROR libqtile manager.py:load_config():L112 Configuration error:
  File "/usr/lib/python3.10/site-packages/libqtile/core/manager.py", line 109, in load_config
    self.config.load()
  File "/usr/lib/python3.10/site-packages/libqtile/confreader.py", line 136, in load
    config = importlib.import_module(name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/ervin/.config/qtile-x11/config.py", line 1, in <module>
    from modules import (
  File "/home/ervin/.config/qtile-x11/modules/__init__.py", line 1, in <module>
    from modules import hooks
  File "/home/ervin/.config/qtile-x11/modules/hooks.py", line 11, in <module>
    from modules.functions import check_if_process_running
  File "/home/ervin/.config/qtile-x11/modules/functions.py", line 10, in <module>
    from modules.settings import group_layouts as def_group_layouts
  File "/home/ervin/.config/qtile-x11/modules/settings.py", line 3, in <module>
    from modules.theme import colors
  File "/home/ervin/.config/qtile-x11/modules/theme.py", line 3, in <module>
    from modules.settings import qtile_path
ImportError: cannot import name 'qtile_path' from partially initialized module 'modules.settings' (most likely due to a circular import) (/home/ervin/.config/qtile-x11/modules/settings.py)
```
The only "chain" I see happening is this one, so I suppose not importing in `theme.py` is as good a hack I can get.

m-col: I'm a bit surprised you weren't getting that error before. If I'm reading that traceback right, you have these imports:

functions.py -> imports from settings.py -> imports from theme.py -> imports from modules.settings

ervinpopescu: Yeah, it's a good old circular import :)

PR #4068, opened 2022-12-15
Improve `qtile check` error message on missing dep

yejianfengblue:
As an unexperienced Python user, I get from the original error message to install the dependencies with `pip install xxx`, but not know it's necessary to add to PATH. See my original question https://github.com/qtile/qtile/discussions/4067


elParaguayo:
Thanks for submitting this. We'll take a look when we can.

tych0 (approved):

m-col:
Thanks for this! LGTM :)

PR #4072, opened 2022-12-17
tox: Fix passenv error

hiagofranco:
Fix whitespace error related to tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 skip_missing_interpreters = True
 skipsdist=True
-minversion = 1.8
+minversion = 4.0.12
 envlist =
     pypy3,
     py39,
@@ -20,7 +20,7 @@ envlist =
 # that our tests use.
 setenv = LC_CTYPE = en_US.UTF-8
 # Pass Display down to have it for the tests available
-passenv = DISPLAY WAYLAND_DISPLAY
+passenv = DISPLAY,WAYLAND_DISPLAY
 whitelist_externals=convert
 # xcffib has to be installed before cairocffi
 deps =


hiagofranco:
Message error:

```
py39: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'DISPLAY WAYLAND_DISPLAY'
  py39: FAIL code 1 (0.00 seconds)
  evaluation failed :( (0.04 seconds)
```

m-col:
Thanks. Do you know why this hasn't caused issues in the past? Did something change in tox?

hiagofranco:
This change was merged last week and the last release is quite new: https://github.com/tox-dev/tox/pull/2671

m-col:
Excellent, thanks. In that case, we should probably try to force updates or at least avoid confusing problems for devs using older tox versions. Maybe `minversion` in tox.ini: https://tox.wiki/en/latest/config.html#minversion

hiagofranco:
Updated the `minversion` to 4.0.12.

m-col:
I suspect you'll need to update `.github/workflows/ci.yml`, though it looks like it already pins tox: `pip -q install "tox<4" tox-gh-actions`. See 6c8112cf3ad48c092f6e6819b9e6ed82c385edde. Perhaps tox v4 now supports tox-gh-actions?

hiagofranco:
Thanks for the information @m-col ! 
Checkig the [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) repo, it appears that support to tox v4 is still in beta:
image
So I think it's not a good idea to migrate this now...

m-col:
We could give it a go. If you update the `pip install` line to install the new tox version, the CI on the PR will run with the new version and we can see if it works :)

ramnes:
Already fixed / merged in 5bb4d6498f556419e5649af10ba5bac436d19c19

PR #4073, opened 2022-12-17
x11: Catch ConnectionExceptions when finalising following server disconnect

m-col:
This follows on from #3374 which aimed to suppress the noisy log messages. This approach instead catches and handles the exceptions (usually ignoring them and continuing), letting Qtile continue to shut down in a more graceful way.

Thanks @elParaguayo and @jwijenbergh for looking into this on #3374 (and apologies for missing the ping). I wanted to avoid globally blocking all logging so here instead just accounted for some specific points in the code where the exception is expected to be raised in the absence of an X connection.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -810,11 +810,11 @@ class Core(base.Core):
         """Try to close windows gracefully before exiting"""
 
         try:
-            pids = [
-                pid
-                for win in self.qtile.windows_map.values()
-                if not isinstance(win, base.Internal) and (pid := win.get_pid())
-            ]
+            pids = []
+            for win in self.qtile.windows_map.values():
+                if not isinstance(win, base.Internal):
+                    if pid := win.get_pid():
+                        pids.append(pid)
         except xcffib.ConnectionException:
             logger.warning("Server disconnected, couldn't close windows gracefully.")
             return
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -170,10 +170,13 @@ class Core(base.Core):
         return "x11"
 
     def finalize(self) -> None:
-        self.conn.conn.core.DeletePropertyChecked(
-            self._root.wid,
-            self.conn.atoms["_NET_SUPPORTING_WM_CHECK"],
-        ).check()
+        try:
+            self.conn.conn.core.DeletePropertyChecked(
+                self._root.wid,
+                self.conn.atoms["_NET_SUPPORTING_WM_CHECK"],
+            ).check()
+        except xcffib.ConnectionException:
+            pass
         self.qtile = None
         self.conn.finalize()
 
@@ -331,12 +334,7 @@ class Core(base.Core):
             except Exception:
                 error_code = self.conn.conn.has_error()
                 if error_code:
-                    error_string = xcbq.XCB_CONN_ERRORS[error_code]
-                    logger.exception(
-                        "Shutting down due to X connection error %s (%s)",
-                        error_string,
-                        error_code,
-                    )
+                    logger.warning("Shutting down due to disconnection from X server")
                     self.remove_listener()
                     self.qtile.stop()
                     return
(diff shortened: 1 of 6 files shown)


libqtile/backend/x11/core.py
```diff
@@ -328,12 +331,7 @@ def _xpoll(self) -> None:
             except Exception:
                 error_code = self.conn.conn.has_error()
                 if error_code:
-                    error_string = xcbq.XCB_CONN_ERRORS[error_code]
-                    logger.exception(
-                        "Shutting down due to X connection error %s (%s)",
-                        error_string,
-                        error_code,
-                    )
+                    logger.warning("Shutting down due to disconnection from X server")
```
elParaguayo, line 334: How do we know it's disconnection-related error?

libqtile/backend/x11/core.py
```diff
@@ -804,19 +802,15 @@ def flush(self):
...
-                logger.exception("Got an exception in getting the window pid")
-                return None
-
-        pids = map(get_interesting_pid, self.qtile.windows_map.values())
-        pids = list(filter(lambda x: x is not None, pids))
+        try:
+            pids = [
+                pid
+                for win in self.qtile.windows_map.values()
+                if not isinstance(win, base.Internal) and (pid := win.get_pid())
```
elParaguayo, line 809: First walrus operator?

libqtile/backend/x11/core.py
```diff
@@ -804,19 +802,15 @@ def flush(self):
...
+            ]
+        except xcffib.ConnectionException:
+            logger.warning("Server disconnected, couldn't close windows gracefully.")
+            return
```
elParaguayo, line 813: Would `pids` be empty at this point? If there are some processes, shouldn't we still try to close them rather than returning here?

libqtile/backend/x11/drawer.py
```diff
@@ -83,7 +83,11 @@ def _create_gc(self):
 
     def _free_gc(self):
         if self._gc is not None:
-            self.qtile.core.conn.conn.core.FreeGC(self._gc)
+            try:
+                self.qtile.core.conn.conn.core.FreeGC(self._gc)
+            except xcffib.ConnectionException:
+                # Qtile got disconnected from the X server
+                pass
```
elParaguayo, line 90: Might be overkill but you use this `try... except...` block multiple times. We could create a context manager wrapper so you could do something like:
``` python
with skip_on_connection_exception():
    self.qtile.core.conn.conn.core.FreeGC(self._gc)
```

elParaguayo (commented):
Looks good and is a bit more considered than my crude attempt!

Have made some comments but they're all pretty minor.

libqtile/backend/x11/core.py
m-col: I assumed that was the case based on the log message that I replaced

libqtile/backend/x11/core.py
m-col: I think so! This isn't the prettiest use of it but it means we don't have to do `win.get_pid()` twice. Alternatively we could just do
```py
            pids = []
                if not isinstance(win, base.Internal):
                    if pid := win.get_pid():
                        pids.append(pid)
```
Perhaps easier to read?

libqtile/backend/x11/core.py
m-col: It's the `win.get_pid()` that tries to communication with the X server and fails, so `pids` doesn't actually get set. This is also why there are two for loops - get all pids (if we can), then kill them below. Either they will all fail or none (because it's same connection). If they fail, we don't know which pids were X clients so we can't kill any processes. From what I can tell, this situation comes about if the X server closes connections before us, and the same thing will happen to those clients, so they're all getting connection errors too, and will close themselves.

libqtile/backend/x11/drawer.py
m-col: Yeah I like it. DRY. Will do.

libqtile/backend/x11/core.py
elParaguayo: Yes - that's easier to follow.

libqtile/backend/x11/drawer.py
elParaguayo: Nice. Didn't know about `contextlib.suppress`. Exactly what's needed.

jwijenbergh:
My 'patch' was a big hack, so this looks a lot better to me.

PR #4075, opened 2022-12-18
Don't try to update Qtile.keys_map in Qtile.grab_keys

m-col:
e577389 added log warnings when a key that is already present in `Qtile.keys_map` gets grabbed. However, the X Core calls `Qtile.grab_keys` when it detects a keyboard (at least in Xephyr), re-grabbing all keys and emitting lots of warnings. This avoids those warnings and instead just forwards the key grabs to the core.


jwijenbergh:
LGTM

Issue #3966, opened 2022-11-08
High CPU Usage


pranavsetpal:
### The issue:

Qtile version: 0.22.1

Whenever I am using Qtile, it takes about 25+% CPU usage, even when it's freshly booted. I faced this issue on 0.21.0 as well

When I used any DE or xmonad I did not face this issue, it usually used to float 3-5%

top:
![image](https://user-images.githubusercontent.com/68775426/200514044-aa28b0c4-19d5-47e7-99ea-3c5cec2327a6.png)

P.S. I did find another issue regrading this same problem, but it was marked unconfirmed and closed. In case you need me to post it there instead kindly inform me.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
Does the same thing happen with the default config? If not, post your config please.

pranavsetpal:
The default config works properly. Should have tested this beforehand.
I'll just mark this as closed and search for the issue. Thanks!

elParaguayo:
No problem. If you do find any issues feel free to reopen this.

pranavsetpal:
P.S. I found out what was causing it, I had set the update interval for the PulseVolume widget to 0, in order to have the widget refresh immediately when I change the volume, Not quite sure how to do it, so if anyone knows, that would much appreciated!

m-col:
Does this change to the widget (bottom of the file) fix it?:
```diff
diff --git a/libqtile/widget/pulse_volume.py b/libqtile/widget/pulse_volume.py
index 8f7385c7..7e4e6bad 100644
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -247,7 +247,7 @@ class PulseVolume(Volume):
 
     def timer_setup(self):
         self.poll()
-        if self.update_interval is not None:
+        if self.update_interval:
             self.timeout_add(self.update_interval, self.timer_setup)
         if self.theme_path:
             self.setup_images()

```
I see that other widgets skip `self.timeout_add` if `self.update_interval` is falsey so this might need to here. An update interval of 0 should be prevented from ramming the CPU.

pranavsetpal:
Hello, this should the fix the issue.
I am not sure how to test it though, I have edited the file in site_packages, but not sure how to install it as pip just says everything is already satisfied.

m-col:
Modifying the file and restarting should be fine

pranavsetpal:
[...]

Issue #4011, opened 2022-11-20
PulseVolume widget should update on volume change event


dannycolin:
### The issue:

Qtile: 0.22.1
Widget config: 
`widget.PulseVolume(fmt="V: {}")` OR `widget.PulseVolume(fmt="V=: {}", update_interval=0)` 

**STR**

1. Add PulseVolume widget
2. Add a keybind to change your volume
3. Change volume

**Actual Behavior**

By default, PulseVolume has a `update_interval` of `0.2` and it creates a delay between the moment you change the volume and the widget redraw.

To avoid this, I set the `update_interval` to `0`. However, this workaround creates another issue making Qtile uses a lot of CPU (> 50%) at all time.

**Expected Behavior**

PulseVolume shouldn't have an `update_interval` by default and instead redraw its content when an event from PulseAudio API is received. This would eliminate the delay in the redrawing and be less taxing on the CPU than aggressively redrawing the widget content.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

Crystal4276:
Hello,

I think i saw this problem reported before.
Probably this one :
https://github.com/qtile/qtile/issues/3966

dannycolin:
@Crystal4276 Yes, that's the issue with `update_interval=0`. However, I opened this one after talking with mcol in #qtile about the fact that the widget should be event based instead of using an update interval. So it isn't a bug report but a feature request ;).

mrtnvgr:
This widget is already event based. If you want to turn off updates on interval, you can set `update_interval` to None.

mrtnvgr:
> This widget is already event based. If you want to turn off updates on interval, you can set `update_interval` to None.

```python
if self.update_interval:
    self.timeout_add(self.update_interval, self.timer_setup)
```

mrtnvgr:
> ... If you want to turn off updates on interval, you can set `update_interval` to `None`.

I think we can make this as default. I will make a PR.

dannycolin:
> This widget is already event based. If you want to turn off updates on interval, you can set `update_interval` to `None`.

It doesn't work for me. When I set it to `None`, the widget never returns the volume information and permanently shows `M`.

mrtnvgr:
> > This widget is already event based. If you want to turn off updates on interval, you can set `update_interval` to `None`.
> 
> It doesn't work for me. When I set it to `None`, the widget never returns the volume information and permanently shows `M`.

[...]

PR #4081, opened 2022-12-23
PulseVolume widget: fix high CPU usage

mrtnvgr:
[pranavsetpal](https://github.com/pranavsetpal):
> Sorry for the late reply, was a bit a busy, but yes,
> 
> > * ```
> >      if self.update_interval is not None:
> >   ```
> > 
> > 
> > * ```
> >      if self.update_interval:
> >          self.timeout_add(self.update_interval, self.timer_setup)
> >      if self.theme_path:
> >          self.setup_images()
> >   ```
> 
> This change fixes the issue for when update_interval is 0.

fixes: #3966

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,6 +35,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Fix `Notify` bug when apps close notifications.
         - Fix `CPU` precision bug with specific version of `psutil`
         - Fix config being reevaluated twice during reload (e.g. all hooks from config were doubled)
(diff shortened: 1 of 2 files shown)


mrtnvgr:
This is a temporary workaround until `PulseVolume` widget will be rewritten to update only when volume change events are received (#4011).

mrtnvgr:
I would like to take #4011 issue, but, unfortunately I won't have time to finish it before my holidays. So instead, I made this PR based on @m-col's solution.

m-col:
Nice, thanks for confirming this fixes it and for posting the PR. LGTM.

PR #4084, opened 2022-12-23
Add separate "lint" and "test" extras

carlwgeorge:
Distribution [packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_linters) usually recommends running the upstream test suite during the package build, but not the upstream linters.  Being able to install the corresponding dependencies for each separately would be useful.


m-col:
Excellent, thanks for this!

PR #4089, opened 2022-12-28
Fix Volume widget volume-getting command execution

inmaldrerah:
Pull request #3332 changed `self.call_process` to `subprocess.getoutput` in the Volume widget. This PR reverts that change.

PR #3332 shouldn't have changed that because the `create_amixer_command` function, which is used to generate command for amixer in the widget, outputs a command in the form of a list of command arguments. `self.call_process` accepts commands in the argument list form while `subproces.getoutput` instead treats the list given as `[command, *args]` where `command` is the actual command executed and `args` are treated as _shell_ parameters, according to [this discussion](https://github.com/qtile/qtile/pull/3332#discussion_r1057588716).

Meanwhile, as I pointed out [here](https://github.com/qtile/qtile/pull/3332#discussion_r1057727841), PR #3332 didn't really break the Volume widget because the first `(\d?\d?\d?)%` in the output of `amixer` happened to be the master volume. So this PR does not introduce any user-visible change either.

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -228,7 +228,9 @@ class Volume(base._TextBox):
         if self.volume_up_command is not None:
             volume_up_cmd = self.volume_up_command
         else:
-            volume_up_cmd = self.create_amixer_command("-q", "sset", self.channel, "{}%+".format(self.step))
+            volume_up_cmd = self.create_amixer_command(
+                "-q", "sset", self.channel, "{}%+".format(self.step)
+            )
 
         subprocess.call(volume_up_cmd, shell=True)
 
@@ -237,7 +239,9 @@ class Volume(base._TextBox):
         if self.volume_down_command is not None:
             volume_down_cmd = self.volume_down_command
         else:
-            volume_down_cmd = self.create_amixer_command("-q", "sset", self.channel, "{}%-".format(self.step))
+            volume_down_cmd = self.create_amixer_command(
+                "-q", "sset", self.channel, "{}%-".format(self.step)
+            )
 
         subprocess.call(volume_down_cmd, shell=True)
 
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -130,7 +130,7 @@ class Volume(base._TextBox):
             cmd.extend(["-D", str(self.device)])
 
         cmd.extend([x for x in args])
-        # Escape existing quotation marks and construct the command with adequate qutation marks
+        # Escape existing quotation marks and construct the command with adequate quotation marks
         return " ".join(map(lambda x: r'"' + x.replace(r'"', r'\"') + r'"', cmd))
 
     def button_press(self, x, y, button):
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -199,13 +199,13 @@ class Volume(base._TextBox):
             if self.get_volume_command:
                 get_volume_cmd = self.get_volume_command
 
-            mixer_out = subprocess.getoutput(get_volume_cmd)
+            mixer_out = self.call_process(get_volume_cmd)
         except subprocess.CalledProcessError:
             return -1
 
         check_mute = mixer_out
         if self.check_mute_command:
-            check_mute = subprocess.getoutput(self.check_mute_command)
+            check_mute = self.call_process(self.check_mute_command)
 
         if self.check_mute_string in check_mute:
             return -1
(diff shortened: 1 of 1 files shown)


adolfgatonegro:
I've looked into this a bit more and I see what the problem is. However, reverting to `self.call_process` breaks the widget when using audio tools other than `amixer`, which is precisely why I made the change in the first place.

Since the problem is that `subprocess.getoutput` doesn't handle `create_amixer_command` correctly as a list, I believe we could solve this with a change to the `create_amixer_command` function itself to output the  command as a string instead of a list using `join()`. This way `subprocess.getoutput` doesn't need to be reverted, and other audio tools will continue to work.

inmaldrerah:
> I've looked into this a bit more and I see what the problem is. However, reverting to `self.call_process` breaks the widget when using audio tools other than `amixer`, which is precisely why I made the change in the first place.

Why won't they work? Just pass a list rather than a shell command, for me `["sh", "-c", "echo $(pw-volume status | jq .volume)%"]`, and it would work. I do think this needs some sort of documentation though.

**Or** maybe we can change all these user-customizable commands to string form. Then `create_amixer_command` should also be redesigned. This can't be as simple as joining the commands: we have to escape the string parameters/variables of `create_amixer_command` correctly. (Personally I do not think this is quite worthy)

BTW, the `subprocess.call` for mute/volume up/volume down also accepts only command lists. If using a shell-style command is really a goal, these `subprocess.call`s should also be dealt with.

There is also a simple, compatible but (a bit) ugly workaround: check the custom command parameters, and if they are provided as strings, use `["sh", "-c", command]`; if they are provided as lists, just use the list.

elParaguayo:
Can't we just have it so that it takes a list by default but, if you pass a string, we just add `shell=True` to `call_process`  (as this will get passed to `subprocess.check_output`?

adolfgatonegro:
> Can't we just have it so that it takes a list by default but, if you pass a string, we just add `shell=True` to `call_process` (as this will get passed to `subprocess.check_output`?

That would be an option, I guess. But, unless I'm missing something, list vs string only matters when creating the `amixer` command since every other command (eg. volume up/down, mute, etc.) is passed as a string anyway. Having `create_amixer_command` return a string would solve the issue without having to rework other functions or adding checks for lists vs strings.

elParaguayo:
You can pass custom commands too, can't you? That would give us the flexibility to handle both cases.

The important thing is that we don't break existing configs so if users could have entered a list or string we need to make sure that we handle both cases.

adolfgatonegro:
Every instance of `subprocess.call()` already has `shell=True`, so passing a custom command as a list would not work, as only the first item of the list would be processed. As far as I can tell, lists are only relevant for `create_amixer_command`, and tweaking that function to return a string doesn't cause any issues. Existing user configs would still work as they did before.

elParaguayo:
Ah, sorry. I thought the commands were being executed by `call_process`.

So, if custom commands are always strings already then your approach makes sense.

inmaldrerah:
Then what about `get_volume_command`? Existing users might have already passed lists for it. Are special treatments required here?

adolfgatonegro:
> Then what about `get_volume_command`? Existing users might have already passed lists for it. Are special treatments required here?

`get_volume_command` and other custom commands can only be strings, passing a list doesn't work because only the first item is processed. So there is no special treatment required if `create_amixer_command` returns a string instead of a list. Every custom command is still processed the same way, and no existing configs should break.

inmaldrerah:
Users could pass a list to `get_volume_command` (only), because it was passed to `self.call_process` before #3332, although I don't think that really matters much.

Okay to use `subprocess.getoutput` then.

m-col:
I'm not sure if this is helpful at all because I'm not sure I fully understand the discussion, but the flexibility of accepting both lists and strings for command lines while ensuring the command doesn't get corrupted by the conversion might be achieved using `shlex.split(cmd)` or `subprocess.list2cmdline(cmd)` depending on context, e.g.: https://github.com/qtile/qtile/blob/master/libqtile/core/manager.py#L1201-L1205

Either way, if you're happy with the current changes, that also works. Just triggered the CI to run so let us know if it's all good and we can go ahead and merge

inmaldrerah:
@m-col Could you please trigger the test again, as the previous one failed with pep8 violations?

inmaldrerah:
I don't quite understand why it failed the CI here and passed the ones in my repo after I executed `black` for `volume.py`... I rebased my fork against the master branch, trying to solve the problem (so a force push is done).

inmaldrerah:
Okay since it passed the checks I think it is good to go now. Please select "squash & merge" as there are quite many commits here.

m-col:
Great, thanks for this

Issue #3680, opened 2022-07-05
Laggy mouse window movement with mouse polling rate set to 1000 Hz


thingsiplay:
### The issue:

I can't say if this is a specific issue with Qtile, but I never had this issue when I was still using Ubuntu with Gnome. Nowadays I am on Manjaro and use Qtile (and KDE as secondary login) and have a new mouse. So that's the backstory.

I purchased a new gaming mouse that is capable of setting the USB polling rate to 1000 Hz, which is double of standard USB rate (actually a regular mouse has a setting of 125 Hz, some gaming mices have higher rates such as 250 or 500 Hz and some even 1k like mine). When I set a window in Qtile to windowed mode and want to move it with mouse, then it is extremely laggy and the window is moving slowing to its final position where I dragged it with the mouse, following the path. When I set the mouse polling rate to 500 Hz, then the it is faster and not laggy that much. But it still is laggy. BTW setting the window size is the same problem.

* Your Qtile version

```
$ qtile --version
0.21.1.dev46+g2a0b9179
```

* Relevant **logs**

Nothing relevant (I think).

* If relevant, the problematic part of your config.

```
mouse = [

    Drag([mod, numlock], "Button1",
        lazy.window.set_position_floating(),
        start=lazy.window.get_position(),
    ),

    Drag([mod, numlock], "Button3",
        lazy.window.set_size_floating(),
        start=lazy.window.get_size(),
    ),

]
```

* Some system information as additional info:

```
OS: Manjaro Linux [x86_64]
Kernel: 5.18.7-1-MANJARO
Uptime: 22 hours, 51 mins
Packages: 2063 (pacman)[stable], 3 (flatpak)
Shell: zsh 5.9
Resolution: 2560x1440 @ 144Hz
DE: KDE Plasma 5.24.5
WM: Qtile 0.21.1.dev46+g2a0b9179 (X11)
Theme: breeze (GruvboxColors) [Plasma], Klaus [GTK2/3/4]
Icons: breeze [Plasma], Gruvbox [GTK2/3/4]
Font: Noto Sans (12pt) [Plasma], Noto Sans (12pt) [GTK2/3/4]
Cursor: default (24px)
Terminal: kitty
CPU: Intel Xeon E3-1230 v3 (8) @ 3.7GHz
GPU: Nvidia GeForce GTX 1070
Memory: 3924MiB / 15949MiB (24%)
Disk (/): 164GiB / 219GiB (74%)
Locale: de_DE.UTF-8
```

[...]

PR #4091, opened 2022-12-28
X11 Core: Fix laggy resize by doing MotionNotifies with default 120 fps

### LINKED ISSUE #1584: Floating Windows Stutter/Lag when Moved/Resized
When I move/resize windows they lag/stutter really badly; sometimes if you really quickly either resize or move a floating window the whole wm will freeze for a couple seconds. This is especially prevalent with heavier programs for some reason. Also; somewhat unrelated; when resizing if the resized by the corner closest to your mouse instead of bottom right all the time it would feel a lot more natural.

jwijenbergh:
Inspired by this 2014 commit from dwm:

https://git.suckless.org/dwm/commit/3d1090ba896319368c4771b88d325fcee368a608.html

The fps is configurable per screen

Fixes #1584 #3680

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -369,7 +369,10 @@ class Screen(CommandObject):
     resized to fill it. If the mode is ``"stretch"``, the image is stretched to fit all
     of it into the screen.
 
-    The ``x11_drag_polling_rate`` parameter specifies the rate for drag events in the X11 backend. By default this is set to 60, but if you prefer it you can set it to your monitor's refresh rate. 60 would mean that we handle a drag event 60 times per second.
+    The ``x11_drag_polling_rate`` parameter specifies the rate for drag events in the X11
+    backend. By default this is set to 120, but if you prefer it you can set it lower for
+    better performance or higher if you have a high refresh rate monitor. 120 would mean
+    that we handle a drag event 120 times per second.
 
     """
 
@@ -384,7 +387,7 @@ class Screen(CommandObject):
         right: BarType | None = None,
         wallpaper: str | None = None,
         wallpaper_mode: str | None = None,
-        x11_drag_polling_rate: int = 60,
+        x11_drag_polling_rate: int = 120,
         x: int | None = None,
         y: int | None = None,
         width: int | None = None,
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -369,6 +369,8 @@ class Screen(CommandObject):
     resized to fill it. If the mode is ``"stretch"``, the image is stretched to fit all
     of it into the screen.
 
+    The ``x11_drag_polling_rate`` parameter specifies the rate for drag events in the X11 backend. By default this is set to 60, but if you prefer it you can set it to your monitor's refresh rate. 60 would mean that we handle a drag event 60 times per second.
+
     """
 
     group: _Group
@@ -382,6 +384,7 @@ class Screen(CommandObject):
         right: BarType | None = None,
         wallpaper: str | None = None,
         wallpaper_mode: str | None = None,
+        x11_drag_polling_rate: int = 60,
         x: int | None = None,
         y: int | None = None,
         width: int | None = None,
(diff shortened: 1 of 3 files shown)


jwijenbergh:
With a 1000hz polling mouse:

Before:

https://user-images.githubusercontent.com/46386452/209800205-451d92cb-ca52-47c3-90f0-f909e8542ee9.mp4

After: 

https://user-images.githubusercontent.com/46386452/209800224-1a1cf9af-5b9e-4370-b83c-55357144e645.mp4

elParaguayo:
This looks nice and very similar to the approach I tried earlier but my attempt was a bit cruder. 

Would be great if we could get the users who reported the bugs to test this.

Wampa842:
Logitech G502 polling at 1000Hz. It works.

Moving and resizing used to lag terribly. Not experiencing any lag after installing the fork, only minimal when running picom with vsync.

jwijenbergh:
> Logitech G502 polling at 1000Hz. It works.
> 
> Moving and resizing used to lag terribly. Not experiencing any lag after installing the fork, only minimal when running picom with vsync.

Thanks for testing, that is great to hear!

m-col:
Excellent work, this looks great. Could we name the parameter more generally, as it applies to all `Drag` events, and also prefix it with `x11_` to make it clear it only applies to that backend? Perhaps `x11_drag_polling_rate`?

jwijenbergh:
> Excellent work, this looks great. Could we name the parameter more generally, as it applies to all `Drag` events, and also prefix it with `x11_` to make it clear it only applies to that backend? Perhaps `x11_drag_polling_rate`?

Great point! I have implemented this

m-col:
Just a thought - if we hard-coded the value to 120, could we get away with it not being configurable or documented? Is there a perceptible difference in the lagginess between 60 and 120?

jwijenbergh:
> Just a thought - if we hard-coded the value to 120, could we get away with it not being configurable or documented? Is there a perceptible difference in the lagginess between 60 and 120?

I think we could pick a value and stick with it e.g. 120, but I guess that some people only would want 60 on lower end systems whereas people with high refresh rate monitors and a good cpu would prefer 120 or above. We could also set it to the refresh rate or something? But maybe that gives problems with really high refresh rates like 360, so I thought making it configurable was best here.

What do you think?

ramnes (approved):
excellent

ramnes:
As an ex-gamer, there definitely is a perception difference between 60 and 120 FPS on any animation! I'm fine with how the PR is as-is (i.e. leaving it documented and configurable), although I'd put 120 as a default value. I think we most likely have more users with computers that can handle 120 correctly than users with computers that would lag at 120.

m-col:
Makes sense! Sounds good

jwijenbergh:
Good point. I have set the default to 120! Also rebased with master.

libqtile/config.py
```diff
@@ -369,6 +369,8 @@ class Screen(CommandObject):
     resized to fill it. If the mode is ``"stretch"``, the image is stretched to fit all
     of it into the screen.
 
+    The ``x11_drag_polling_rate`` parameter specifies the rate for drag events in the X11 backend. By default this is set to 120, but if you prefer it you can set it lower for better performance or higher if you have a high refresh rate monitor. 120 would mean that we handle a drag event 120 times per second.
```
tych0, line 372: Can you wrap this line and the line in the changelog? Other than that this patch looks awesome, thank you for doing it!

jwijenbergh: Thanks!

thingsiplay:
I am little late, but want to confirm this fixes for me with the default values in Qtile on my 144Hz monitor and 1000Hz polling mouse (also G502). Thank you for this improvement!

jwijenbergh:
> I am little late, but want to confirm this fixes for me with the default values in Qtile on my 144Hz monitor and 1000Hz polling mouse (also G502). Thank you for this improvement!

Thanks for testing! Great to hear

PR #4092, opened 2022-12-28
Fixes #4090 toggle bar visibility

### LINKED ISSUE #4090: hide_show_bar("all") errors when there are bar.Gap objects on any side
### The issue:

```
❯ qtile --version
0.22.1
```
If there are Gap objects on any side, and use the "all" option to toggle bar visibility with `hide_show_bar()` function, it throws an error warning.
For my case, I have a bar only on one side, bottom, and rest are `Gap`s, the `hide_show_bar()` function's default option "all".

logs from `~/.local/share/qtile/qtile.log`
```
2022-12-28 06:08:54,637 ERROR libqtile manager.py:process_key_event():L407 KB command error hide_show_bar: Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/libqtile/command/interface.py", line 312, in call
  File "/usr/lib/python3.10/site-packages/libqtile/core/manager.py", line 1565, in cmd_hide_show_bar
    if is_show is None:
AttributeError: 'Gap' object has no attribute 'is_show'
```
For the key press event, I'm using this in my config, since "all" is default
`Key([mod], "b", lazy.hide_show_bar(), desc="Toggle bar visibility")`
Instead of using the `if bar:` condition, checking if the bar is an actual bar object or not resolves the issue.
```
for bar in [screen.left, screen.right, screen.top, screen.bottom]:
#if bar:                                ## the original condition
if isinstance(bar, libqtile.bar.Bar):
    if is_show is None:
        is_show = not bar.is_show()
    bar.show(is_show)
```
I don't know if this is meant to be as is for bars on all sides only, but I think this way, you wouldn't have to toggle each bar one at a time either.

### Required:

ugly-bastard:
Fixes #4090 toggle bar visibility issue

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1663,7 +1663,7 @@ class Qtile(CommandObject):
             screen = self.current_screen
             is_show = None
             for bar in [screen.left, screen.right, screen.top, screen.bottom]:
-                if bar:
+                if isinstance(bar, libqtile.bar.Bar):
                     if is_show is None:
                         is_show = not bar.is_show()
                     bar.show(is_show)


elParaguayo:
Excellent. Thanks for submitting this. I've started the tests running.

m-col:
Huh, strange that mypy didn't pick up on the bug. If I get mypy to print the inferred types of the variables involved:
```py
            for bar in [screen.left, screen.right, screen.top, screen.bottom]:
                reveal_type(bar)
                reveal_type(screen.left)
                reveal_type(screen.right)
                reveal_type(screen.top)
                reveal_type(screen.bottom)
```
It reveals:
```
libqtile/core/manager.py:1691: note: Revealed type is "Any"
libqtile/core/manager.py:1692: note: Revealed type is "Union[Union[libqtile.bar.Bar, libqtile.bar.Gap], None]"
libqtile/core/manager.py:1693: note: Revealed type is "Union[Union[libqtile.bar.Bar, libqtile.bar.Gap], None]"
libqtile/core/manager.py:1694: note: Revealed type is "Union[Union[libqtile.bar.Bar, libqtile.bar.Gap], None]"
libqtile/core/manager.py:1695: note: Revealed type is "Union[Union[libqtile.bar.Bar, libqtile.bar.Gap], None]"
```
The last 4 are correct: `Union[Union[libqtile.bar.Bar, libqtile.bar.Gap], None]` is an alias for `BarType | None` which is how they are annotated on `Screen`. But despite the line on the `for` line containing 4 items of the same type, mypy infers `bar` as `Any`, and therefore doesn't complain about the `is_show()` method potentially being absent in the case of `Gap`. IDK why this is happening.

Anyway, that's a thought for another day. Fix looks good and the tests are happy, thank you for this.

PR #4093, opened 2022-12-28
Wayland: Fix get_inputs not returning touchpad

### LINKED ISSUE #3978: Wayland: `get_inputs` command reports type:pointer instead of type:trackpad for trackpads
### The issue:

Perhaps other input device types are being incorrectly reported.

### Required:

jwijenbergh:
Fixes #3978 

Sway still gives more types though in a nested session (host running x11) sway gives X11 touch (touch, not touchpad there is a difference), X11 pointer, X11 keyboard. But with this it still gives only pointer and keyboard. However, just running it from the tty reports touchpad correctly so this does fix something. To fully fix it, we need some additional code here https://github.com/qtile/qtile/blob/3225a562e0b0d573294dc140631399076a018d1f/libqtile/backend/wayland/core.py#L327. That is handled in #3931

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -25,6 +25,7 @@ import contextlib
 import os
 import time
 import typing
+from collections import defaultdict
 
 import pywayland
 import pywayland.server
@@ -1286,26 +1287,18 @@ class Core(base.Core, wlrq.HasListeners):
     @expose_command()
     def get_inputs(self) -> dict[str, list[dict[str, str]]]:
         """Get information on all input devices."""
-        info = {}
-        device_lists: dict[str, list[inputs._Device]] = {
-            "type:keyboard": self.keyboards,  # type: ignore
-            "type:pointer": self._pointers,  # type: ignore
-        }
+        info: defaultdict[str, list[dict]] = defaultdict(list)
+        devices: list[inputs._Device] = self.keyboards + self._pointers  # type: ignore
 
-        for type_key, devices in device_lists.items():
-            type_info = []
+        for dev in devices:
+            type_key, identifier = dev.get_info()
+            type_info = dict(
+                name=dev.wlr_device.name,
+                identifier=identifier,
+            )
+            info[type_key].append(type_info)
 
-            for dev in devices:
-                type_info.append(
-                    dict(
-                        name=dev.wlr_device.name,
-                        identifier=dev.get_info()[1],
-                    )
-                )
-
-            info[type_key] = type_info
-
-        return info
+        return dict(info)
 
     def get_mouse_position(self) -> tuple[int, int]:
         """Get mouse coordinates."""


libqtile/backend/wayland/core.py
```diff
@@ -1286,26 +1287,19 @@ def unhide_cursor(self) -> None:
     @expose_command()
     def get_inputs(self) -> dict[str, list[dict[str, str]]]:
         """Get information on all input devices."""
-        info = {}
-        device_lists: dict[str, list[inputs._Device]] = {
-            "type:keyboard": self.keyboards,  # type: ignore
-            "type:pointer": self._pointers,  # type: ignore
```
jwijenbergh, line 1292: This pr fixes something because pointer was hardcoded here, but we still need to do additional checking if it's a pointer or a touchpad. See https://github.com/qtile/qtile/blob/3225a562e0b0d573294dc140631399076a018d1f/libqtile/backend/wayland/inputs.py#L172

libqtile/backend/wayland/core.py
```diff
@@ -1286,26 +1287,19 @@ def unhide_cursor(self) -> None:
...
+        devices: list[inputs._Device] = self.keyboards + self._pointers  # type: ignore
+
+        for dev in devices:
+            dev_info = dev.get_info()
+            type_key = dev_info[0]
+            type_info = dict(
+                name=dev.wlr_device.name,
+                identifier=dev_info[1],
+            )
+            info[type_key] += [type_info]
```
m-col, line 1300: We can avoid creating the throwaway list by just appending:
```suggestion
            type_key, identifier = dev.get_info()
            type_info = dict(
                name=dev.wlr_device.name,
                identifier=identifier,
            )
            info[type_key].append(type_info)
```

libqtile/backend/wayland/core.py
```diff
@@ -1286,26 +1287,19 @@ def unhide_cursor(self) -> None:
...
-            for dev in devices:
-                type_info.append(
-                    dict(
-                        name=dev.wlr_device.name,
-                        identifier=dev.get_info()[1],
-                    )
-                )
-
-            info[type_key] = type_info
+        info: DefaultDict[str, list[dict]] = defaultdict(lambda: [])
```
m-col, line 1290: We no longer need to use typing.DefaultDict for annotations, and I think we can avoid creating a new object (the lambda) by just using list:
```suggestion
        info: defaultdict[str, list[dict]] = defaultdict(list)
```

libqtile/backend/wayland/core.py
m-col: I'm not sure I understand what would be left to fix - from what I can tell, your changes fix the issue.

Perhaps you mean that we don't report all possible input devices, such as 'touch' devices, as you mention in the OP re https://github.com/qtile/qtile/pull/3931? For a bit more context, we don't support those devices yet, but when we implement support we will have to ensure `get_inputs()` is reporting things as expected. I think your solution here wouldn't need changing as it is indifferent, but we'll probably end up needing a new `_Device` subclass like `Touch` that reports the correct info from `get_info()`.

jwijenbergh: Yes that's what I was saying indeed.  The only thing we would then need to do is also add the touch devices to the list that we loop through in get_inputs.

My PR description was more of a thinking out loud as I was comparing the output to sway's output

libqtile/backend/wayland/core.py
jwijenbergh: And that makes a lot of sense

libqtile/backend/wayland/core.py
jwijenbergh: Ah I need to read up on the updated annotations. Thanks!

jwijenbergh:
Steps done:
- Squashed commits
- Implemented suggestions

m-col:
Just tested and confirmed, thanks for this! :)

PR #4094, opened 2022-12-28
Move `GenPollCommand`'s default Button1 callback to `ThreadPoolText` class.

mrtnvgr:
This PR adds default `force_update` callback to `GenPollUrl` class just like in the `GenPollCommand` class.

### DIFF
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -61,10 +61,6 @@ class GenPollUrl(base.ThreadPoolText):
         if self.data and not isinstance(self.data, str):
             self.data = json.dumps(self.data).encode()
 
-    def _configure(self, qtile, bar):
-        base.ThreadPoolText._configure(self, qtile, bar)
-        self.add_callbacks({"Button1": self.force_update})
-
     def fetch(self):
         req = Request(self.url, self.data, self.headers)
         res = urlopen(req)
@@ -111,10 +107,6 @@ class GenPollCommand(base.ThreadPoolText):
     def __init__(self, **config):
         base.ThreadPoolText.__init__(self, "", **config)
         self.add_defaults(GenPollCommand.defaults)
-
-    def _configure(self, qtile, bar):
-        base.ThreadPoolText._configure(self, qtile, bar)
-        self.add_callbacks({"Button1": self.force_update})
 
     def poll(self):
         process = subprocess.run(
--- a/libqtile/widget/generic_poll_text.py
+++ b/libqtile/widget/generic_poll_text.py
@@ -61,6 +61,10 @@ class GenPollUrl(base.ThreadPoolText):
         if self.data and not isinstance(self.data, str):
             self.data = json.dumps(self.data).encode()
 
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
+        self.add_callbacks({"Button1": self.force_update})
+
     def fetch(self):
         req = Request(self.url, self.data, self.headers)
         res = urlopen(req)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Add ability to spawn the `WidgetBox` widget opened.
         - Add ability to swap focused window based on index, and change the order of windows inside current group
         - Add `move_to_slice` command to move current window to single layout in `Slice` layout
+        - Add ability to force update a `OpenWeather` widget on `Button1` mouse callback.
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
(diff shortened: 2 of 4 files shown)


elParaguayo:
I'm not sure this needs to be hard coded in the widget. You could just do this in your config by setting `mouse_callbacks={"Button1": lazy.widget["openweather"].force_update()}`

mrtnvgr:
> I'm not sure this needs to be hard coded in the widget. You could just do this in your config by setting `mouse_callbacks={"Button1": lazy.widget["openweather"].force_update()}`

I think it might be useful not only for me, but in any case you decide. Other widgets do similar approach (e.g. `do_execute` in `CheckUpdates` widget)

mrtnvgr:
I noticed that `force_update` callback sets to every widget based on `GenPollCommand`, maybe we do this with `GenPollUrl` too?

libqtile/widget/generic_poll_text.py
```diff
@@ -61,6 +61,10 @@ def __init__(self, **config):
         if self.data and not isinstance(self.data, str):
             self.data = json.dumps(self.data).encode()
 
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
+        self.add_callbacks({"Button1": self.force_update})
```
elParaguayo, line 66: This should not be done in `_configure`. Widgets typically set their callbacks during `__init__` so, if we change it here, any custom settings in widgets will be overriden when the widget is configured.

libqtile/widget/generic_poll_text.py
```diff
@@ -61,6 +61,10 @@ def __init__(self, **config):
         if self.data and not isinstance(self.data, str):
             self.data = json.dumps(self.data).encode()
 
+    def _configure(self, qtile, bar):
+        base.ThreadPoolText._configure(self, qtile, bar)
```
elParaguayo, line 65: Why not just do this in `ThreadPoolText`? That's the class that defines the `force_update` method.

elParaguayo (changes requested):
I've added a couple of comments but, to be honest, I'm still not 100% convinced of the need for this.

libqtile/widget/generic_poll_text.py
mrtnvgr: Oh, but `GenPollCommand` does it in `_configure`

mrtnvgr:
> I've added a couple of comments but, to be honest, I'm still not 100% convinced of the need for this.

With this and #4085, user can make a widget which updates only on mouse presses.

elParaguayo:
> > I've added a couple of comments but, to be honest, I'm still not 100% convinced of the need for this.
> 
> With this and #4085, user can make a widget which updates only on mouse presses.

It doesn't _need_ this PR (the button configuration can be done in users' own config) but it's a convenience.

mrtnvgr:
> > > I've added a couple of comments but, to be honest, I'm still not 100% convinced of the need for this.
> > 
> > With this and #4085, user can make a widget which updates only on mouse presses.
> 
> It doesn't _need_ this PR (the button configuration can be done in users' own config) but it's a convenience.

Isn't convenience a good thing?

mrtnvgr:
> This should not be done in `_configure`. Widgets typically set their callbacks during `__init__` so, if we change it here, any custom settings in widgets will be overriden when the widget is configured.

 I do this in `__init__`, some tests fail.

elParaguayo:
> > > > I've added a couple of comments but, to be honest, I'm still not 100% convinced of the need for this.
> > > 
> > > 
> > > With this and #4085, user can make a widget which updates only on mouse presses.
> > 
> > 
> > It doesn't _need_ this PR (the button configuration can be done in users' own config) but it's a convenience.
> 
> Isn't convenience a good thing?

Potentially. But not if it causes unexpected behaviour for users. Some users may not want their widgets updated when they click on them but this change forces that situation on them. To restore their normal behaviour, they now need to make extra config changes. That's not "convenient" for them.

This isn't solving an issue that impacts everybody. That's why I'm hesitant here.

elParaguayo:
> > This should not be done in `_configure`. Widgets typically set their callbacks during `__init__` so, if we change it here, any custom settings in widgets will be overriden when the widget is configured.
> 
> If I do this in `__init__`, some tests fail.

Understanding _why_ they fail would be useful.

m-col:
+1 for not applying this at the level of `ThreadPoolText` to not change things for most existing configs. The reality of qtile is that some things the perfectionist programmer would want to change have to be lived with as users prefer stability over consistency or convenience :laughing:

That said, it does seem intuitive to me for this to be the default for the generic polling widgets in `generic_poll_text.py`, though it's only the case for one of them, and that fact is not documented and one would only discover this through testing or reading the code. To avoid the disruption perhaps it would be best just to add a line to the docstrings of the widgets saying how they can achieve this (i.e. @elParaguayo's first comment), and for `GenPollCommand` that it will update when clicked.

Issue #3933, opened 2022-10-26
Suspended/minimized window disappears and it is difficult to restore (reproducible with Emacs)


polaris64:
### The issue:

I have noticed that sometimes windows get minimized/suspended due to something that the owning application does and it's then quite difficult to recover them. This is easily reproducible for me using Emacs.

## Steps to reproduce
1. Open Emacs
2. Type `M-x suspend-frame RET` to minimize/suspend the Emacs frame (window)

## Outcome
The window's title is still visible in the Qtile bar and it appears that the window still has input focus, but the window is completely invisible.

### Restoring the window state
The only way I have found to restore the window is to toggle the window's floating state. Setting it to floating after suspending makes it visible again but not floating (two successive requests to toggle floating are required to make it float).

This is more complicated if the window exists on a desktop with one or more other windows because as soon as another visible window is focused there seems to be no way of focusing the invisible window again except by moving all visible windows to another desktop so that the invisible window is the only window remaining.

## Technical information
 - Window system: x11
 - Qtile version: 0.22.1

Nothing appears in `~/.local/share/qtile/qtile.log` when this happens so I have not included it.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

m-col:
It does sound minimized. In that case, it shouldn't get any input. What makes you think it has input focus? There is a `lazy.window.toggle_minimize()` command if that helps. IIRC the only convenient way to unminimize a window is using the `TaskList` widget. Also, note you can stop clients from minimizing themselves by setting `auto_minimize = False`.

polaris64:
> It does sound minimized. In that case, it shouldn't get any input. What makes you think it has input focus?

I've just double checked and you're right, it doesn't have input focus after all.

> There is a lazy.window.toggle_minimize() command if that helps.

That works perfectly, thank you!

> IIRC the only convenient way to unminimize a window is using the TaskList widget. Also, note you can stop clients from minimizing themselves by setting auto_minimize = False.

These are both good suggestions too, thank you very much. The `TaskList` widget is very useful if there's more than one window on the desktop at the point at which one gets minimized so I have started using that. Un-setting `auto_minimize` also prevents windows from being minimized in the first place as you say, but in the case of Emacs it just causes the window to be "frozen" without any obvious way to restore it other than switching desktops so I decided against using that.

[...]

PR #4098, opened 2022-12-31
more auto_minimize hackery

tych0:
when steam games lose focus, they try to:

1. set IconicState
2. remove _NET_WM_FULLSCREEN

previously with the auto_minimize behavior we ignored the first part, which was fine. If a user only had one monitor, when _NET_WM_FULLSCREEN is removed, we toggle_floating() and re-place() the window according to where the bar was. Except that people with one monitor didn't see this behavior.

If users have multiple monitors, this removal of _NET_WM_FULLSCREEN results in a toggle_floating() which moves the fullscreen window to not being fullscreen, which makes it pop into the current layout, which is weird.

Instead, let's not respect _NET_WM_FULLSCREEN removals like we're not respecting setting to IconicState, and get rid of both of these problems.

Maybe this is a hack-on-hack, and we should find some other way to implement this. I can't really tell myself, but I don't want things to auto minimize just because I switch screens.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1276,6 +1276,7 @@ class Window(_Window, base.Window):
     def __init__(self, window, qtile):
         _Window.__init__(self, window, qtile)
         self._wm_class: list[str] | None = None
+        self.ignored_minimize = False
         self.update_wm_class()
         self.update_name()
         self.set_group()
@@ -1379,7 +1380,7 @@ class Window(_Window, base.Window):
             )
             return
 
-        if self._float_state == FloatStates.FULLSCREEN and self.qtile.config.auto_minimize:
+        if self._float_state == FloatStates.FULLSCREEN and self.ignored_minimize:
             self.floating = False
             return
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks. Was this targeting an existing issue? Feels like something we should get some more people to test.

tych0:
Only something that had bothered me for a long time; steam games on multiple screens were kind of shaky when you alt-tabbed. In general I don't understand how people use minimizing windows with qtile, e.g. the "difficult to restore" part of this bug: https://github.com/qtile/qtile/issues/3933

The behavior makes sense: with `auto_minimize` we're essentially ignoring the applications request to unmap. It just so happens that steam apps continue to function when we've ignored this request, while based on that thread emacs doesn't. But I'm not sure of a better way to solve this problem.

tych0:
Arg, this has some unfortunate side effects... fullscreen browser windows (e.g. firefox) no longer get re-inserted into the layout with this.

tych0:
Yeah, I dunno. Given how much of a hack this is, and how it screws with other things that are more important, I'm going to close this for now. The last WIP commit here I thought should fix it, but didn't seem to. I'll let someone who this bothers more look into it :)

PR #4100, opened 2023-01-01
Relax regex on qtile shell command arg parsing

elParaguayo:
Qtile shell currently uses a regular expression to split the command name from the arguments. However, the arguments only match a limited subset of characters; characters like `.`, `[]` etc are not matched resulting in an `InvalidCommand`. This limits the usefulness of the `eval` command in particular.

This PR fixes this by allowing any characters to be passed as args.

### DIFF
--- a/libqtile/sh.py
+++ b/libqtile/sh.py
@@ -319,7 +319,7 @@ class QSh:
             else:
                 return "Invalid builtin: {}".format(cmd)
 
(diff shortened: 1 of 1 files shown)


tych0:
LGTM, though I wonder if we can steal shlex or something to parse this? I guess you thought about that though based on IRC discussion.

tych0 (approved):

elParaguayo:
I stuck with the regex as it handles extracting the relevant parts from the `command_name(args)` string that it receives.

PR #4110, opened 2023-01-10
Fix width issue for scrolling text

### LINKED ISSUE #4101: Get Mpd2 widget fixed width
### The issue:

- qtile version: 0.22.1

Mpd2 widget has only the width attribute to control the maximize widget width, it won't always get the same width when the different lengths of song/artist names are.

### Required:

elParaguayo:
Currently, when widgets have `scroll=True` set, the `width` parameter acts as maximum width. Where the width of the text is less than this value, the widget will shrink accordingly. This may be undesirable for users wishing to have a fixed width at all times.

This PR adds a new parameter, `scroll_fixed_width`, which, when set to `True`, forces the text widget to have a fixed width at all times.

Fixes #4101


m-col:
Nice, lgtm

PR #4118, opened 2023-01-16
`widget.Wttr`: Allow automatic location

maximilian-franz:
[wttr.in](https://wttr.in) uses your IP address if no location is provided. I made a minor change to make `widget.Wttr` also support this.


maximilian-franz:
I added an explanation note. I don't think that it is possible to query the current location alongside other locations in wttr.in, so it wouldn't be possible here either.

Issue #1634, opened 2020-04-18
Make tests parallelizable


pawamoy:
I looked at the test suite code, and it *looks* like tests should be able to run in parallel, though as soon as I use 2 workers with `pytest-xdist`, tests start failing randomly.

In `conftest.Xephyr.start_xephyr()`:

```python
self.display = ":{}".format(xcffib.testing.find_display())
```

`xcffib.find_display()`:

```python
def find_display():
    display = 10
    while True:
        if not os.path.exists(lock_path(display)):
            return display
        display += 1
```

Doesn't look very thread-safe to me? I mean, is `os.path.exists` atomic?

Also, not sure if 10 is a hard-limit or an arbitrary value.

tych0:
Yes, it is. I just copied this code from lightdm, note the comment: https://github.com/canonical/lightdm/blob/master/src/dm-tool.c#L389

I suppose we could change xcffib to use flock, so at least it wouldn't race with itself?

pawamoy:
Something like

```python
import fnctl

def find_display():
    display = 10
    while True:
        lockfile = lockpath(display)
        if not os.path.exists(lockfile):
            os.system("touch %s" % lockfile)
            f = open(lockfile, "r")
            try:
                fcntl.flock(f.fileno(), fcntl.LOCK_EX|fcntl.LOCK_NB)
                # successfully got a lock
                return display
            except Exception:
                # failed to get a lock
                pass  # try next one
        display += 1
```

Bits found at https://www.programcreek.com/python/example/1605/fcntl.flock

tych0:
Yeah, exactly. I'd definitely take a patch for xcffib that looked like that.

pawamoy:
I can't get it to work.

```python
def lock_path(display):
    return '/tmp/.X%d-lock' % display


def get_lock(display):
    lockfile = lock_path(display)
    if os.path.exists(lockfile):
        print("file exists")
        return False
    try:
        os.system("touch %s" % lockfile)
        f = open(lockfile, "r")
        fcntl.flock(f.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
        return True  # successfully got a lock
    except Exception:
        return False  # failed to get a lock


def release_lock(display):
    lockfile = lock_path(display)
    if not os.path.exists(lockfile):
        print("file does not exist")
        return True
    try:
        f = open(lockfile, "r")
        # the unlock operation doesn't raise any exception
        # so instead we try to re-acquire the lock, then release it
[...]

PR #4122, opened 2023-01-18
Tox: Use pytest-xdist for parallel tests

jwijenbergh:
Testing if this works now. Probably not

Context:

#1634
#2107


elParaguayo:
The changes to the wttr tests shouldn't be needed. They're not making any external requests. It's checking that the widget generates the appropriate URL and that it can parse the provided json response.

PR #4126, opened 2023-01-22
Adding pseudo-transparency

elParaguayo:
Adds pseudotransparency. Benefits are transparency without compositor and fixing issue with systray where you can only have fully transparent or opaque.

Things to do:
- [ ] Maybe some more tidying up?
- [x] Update docs
- [ ] Think how to test this
- [x] Make sure compatible with widget decorations
- [x] Ensure bar/widgets redrawn if wallpaper changed

### DIFF
--- a/docs/manual/config/screens.rst
+++ b/docs/manual/config/screens.rst
@@ -44,6 +44,9 @@ background="#000000")` will give you a black back ground (the default), while
 :code:`bar.Bar(..., background=["#000000", "#FFFFFF"])` will give you a
 background that fades from black to white.
 
+Bar and Widget Transparency
+---------------------------
+
 Bars (and widgets) also support transparency by adding an alpha value to the
 desired color. For example, :code:`bar.Bar(..., background="#00000000")` will
 result in a fully transparent bar. Widget contents will not be impacted i.e.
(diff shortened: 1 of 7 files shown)


elParaguayo:
I can ensure widgets are updated if someone updates the wallpaper with `lazy.screen.set_wallpaper` but, if they use something like `feh`, I don't think there's an easy way to do it. Might be worth adding a command to force a refresh.

elParaguayo:
@m-col @jwijenbergh It's a bit ugly but worth a quick look to see if there's anything obviously wrong here.

Would also welcome thoughts on how any of this can be tested.

docs/manual/config/screens.rst
```diff
@@ -52,7 +55,18 @@ entire window.
 
 .. note::
     In X11 backends, transparency will be disabled in a bar if the ``background``
-    color is fully opaque.
+    color is fully opaque. This is primarily to prevent a rendering issue with
+    the ``Systray`` widget where icons will be displayed with a fully transparent
+    background, regardless of the colour of the bar/widget.
+
+    X11 users can also use "pseudotransparency" by setting ``fake_transparency=True``
```
m-col, line 62: Could we name the option `x11_fake_transparency`?

elParaguayo: Yes.

elParaguayo:
I'm going to rework this one. If we add the ability to use a background image for the bar then we can use the same functionality for pseudo-transparency.

libqtile/backend/x11/drawer.py
```diff
@@ -88,20 +112,27 @@ def _free_gc(self):
...
+    def _free_surfaces(self):
+        for surface in [
+            self._xcb_surface,
+            self.surface,
+            self._pseudo_surface,
+            self._root_surface,
+        ]:
+            if surface is not None:
+                surface.finish()
+                surface = None
```
m-col, line 135: `surface = None` here and `pixmap = None` on line 153 are redundant - the variables get rebound immediately after at the start of the next loop. If you mean to reset the attributes you might need to make it loop over attribute names and use `getattr`/`setattr`

m-col (approved):
LGTM, pretty cool. It would be really nice if we could completely contain the x11-specific logic within the backend -- ideally a bar or widget wouldn't need to know if pseudotransparency is being used -- but I can see why it makes things easier to code.

elParaguayo:
Close in favour of #4496

PR #4133, opened 2023-01-24
Fix ThermalSensors widgets_defaults bug

### LINKED ISSUE #4132: `ThermalSensor` doesn't respect `foreground` in `widget_defaults`
### The issue:

(reported on Discord/IRC)

ThermalSensor widget doesn't respect foreground colour in widget defaults. This is because the foreground value is copied to `foreground_normal` during `__init__` but this happens before `widget_defaults` are copied to the widget (which happens in `_configure`

### Required:

elParaguayo:
The ThermalSensors widget doesn't respect `foreground` set in `widget_defaults` as the foreground value is copied to `foreground_normal` during `__init__` but `widget_defaults` aren't loaded until `_configure`.

Fixes #4132

### DIFF
--- a/libqtile/widget/sensors.py
+++ b/libqtile/widget/sensors.py
@@ -67,7 +67,6 @@ class ThermalSensor(base.InLoopPollText):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(ThermalSensor.defaults)
         temp_values = self.get_temp_sensors()
-        self.foreground_normal = self.foreground
 
         if temp_values is None:
             self.data = "sensors command not found"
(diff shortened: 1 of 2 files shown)


jwijenbergh:
LGTM, no real issues with this. One small comment but not blocking

libqtile/widget/sensors.py
```diff
@@ -81,6 +80,7 @@ def __init__(self, **config):
     def _configure(self, qtile, bar):
         self.unit = "°C" if self.metric else "°F"
         base.InLoopPollText._configure(self, qtile, bar)
+        self.foreground_normal = self.foreground
```
jwijenbergh, line 83: Any reason why we need to make this a variable anyways? I saw only one use for foreground_normal

elParaguayo: It's certainly not necessary to have it but it does, potentially, make the code a little clearer as it differentiates the different foreground values.

jwijenbergh: Yeah I was looking for something to nitpick here, whatever works for me honestly it doesn't matter much

PR #4140, opened 2023-01-27
Fix Mpris2 metadata format bug

elParaguayo:
The `Mpris2` widget now uses a `format` string to display metadata. Previously, desired fields were passed as a list to the `display_metadata` parameter. The widget allows this old parameter by converting it to a format string. However, there was a bug in the conversion that meant that, instead of generating a valid format string (with field names inside curly brackets), the widget just generated a literal string.

This PR fixes the error and generates the correct format string. This will also fix a rendering bug in the widget's screenshot on our docs!


m-col (approved):
Nice

Issue #4119, opened 2023-01-16
Nvidia drivers cause lag / slowdown when resizing or dragging windows


Dolvur:
### The issue:

I am using EndeavourOS's community Qtile config and have only changed some keybindings currently, but I had the same issue on Fedora with another config, so I am pretty sure this is an issue with either Qtile or the nvidia drivers / how I installed the drivers.

Qtile version: ``0.22.1``

So far I tried installing the nvidia drivers in two different ways:
1) ``pacman -S nvidia nvidia-utils nvidia-settings xorg-server-devel opencl-nvidia``
2) From [Lutris recommendation](https://github.com/lutris/docs/blob/master/InstallingDrivers.md) for Arch: ``sudo pacman -S --needed nvidia-dkms nvidia-utils lib32-nvidia-utils nvidia-settings vulkan-icd-loader lib32-vulkan-icd-loader``

The issue is **highly** pronounced when for example watching a video on the second monitor (with mpv) or having discord open. Disabling hardware acceleration on discord removes the issue with discord.

Video example with nvidia drivers:
https://youtu.be/4nPN730OHIg

Example with nouveau drivers:
https://youtu.be/eGI8tQhGBDQ

I would be fine with using nouveau drivers except that high refresh rate does not seem to be working as smoothly.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

gabor-motko:
That issue is fixed by #4091, but it's not included in 22.1. You can mitigate the issue by setting your mouse's polling rate to a lower frequency; 125 Hz worked perfectly for me.

Dolvur:
> That issue is fixed by #4091, but it's not included in 22.1. You can mitigate the issue by setting your mouse's polling rate to a lower frequency; 125 Hz worked perfectly for me.

Thanks for the help! Sadly it only fixed a small part of the problem though. I didn't manage to decrease the polling rate of my g305, but I upgraded to the latest qtile ``0.22.2.dev115+g653c52c2`` with the git AUR. Moving a floating window is now very snappy **except** when I run a video with mpv (guessing it is because of some hardware acceleration stuff, but I am not sure). 

The resizing still seems laggy and I also got another bug during the recording where the tiling broke a little, which might be unrelated since I am not using the release version.

https://youtu.be/LSQFDdl3CuU

Jareng:
You can try to disable picom if it's enabled and/or disable composition pipeline in nvidia-settings

Dolvur:
[...]

PR #4141, opened 2023-01-28
X11 Core: Handle motion notify events later

jwijenbergh:
This is a different approach to limiting motion notifies by handling
the events later. Inspired by AwesomeWM. See
https://github.com/awesomeWM/awesome/blob/master/awesome.c#L396. Hopefully this fixes the nvidia laggy behaviour.

See #4119

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -65,6 +65,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Fix unfullscreening bug in conjunction with Chromium based clients when auto_fullscreen is set to `False`.
         - Ensure `CurrentLayoutIcon` expands paths for custom folders.
         - Fix vertical alignment of icons in `TaskList` widget
+        - Fix laggy resize/positioning of floating windows in X11 by handling motion notify events later
     * python version support
         - We have added support for python 3.11 and pypy 3.9.
         - python 3.7, 3.8 and pypy 3.7 are not longer supported.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -60,8 +60,6 @@ Qtile x.xx.x, released XXXX-XX-XX:
         - Fix config being reevaluated twice during reload (e.g. all hooks from config were doubled)
         - Fix `PulseVolume` high CPU usage when update_interval set to 0.
         - Fix `Battery` widget on FreeBSD without explicit `battery` index given.
-        - Fix laggy resize in X11 by limiting the MotionNotify fps to a default of 120. This is configurable
-          per screen with the x11_drag_polling_rate variable.
         - Fix XMonad layout faulty call to nonexistent _shrink_up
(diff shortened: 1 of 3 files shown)


m-col (approved):
Change itself LGTM, hopefully it fixes the issues :)

purplebar0:
Having a bad experience with this. The removal of `x11_drag_polling_rate` causes coil whine whenever I move a window now, and it broke the config until I commented that out.

jwijenbergh:
Coil whine as your PC? Do you run picom? Also the config giving an error because we removed it is correct. Breaking changes are expected if you run the git version

purplebar0:
> Coil whine as your PC? Do you run picom?
Yes to both.

> How is the resize performance itself?
The performance is good, even with a game running (I haven't done extensive testing yet). Still, I'd prefer if the ability to cap the rate was possible, because the coil whine is pretty annoying and a lot of wasted frames are generated.

jwijenbergh:
Could you let me know if it improves without picom?

Capping events is a bit of a hack that I'd still like to avoid if possible. Worst case we need to do a hybrid approach though (this pr + old cap)

purplebar0:
> Could you let me know if it improves without picom?
> 
> Capping events is a bit of a hack that I'd still like to avoid if possible. Worst case we need to do a hybrid approach though (this pr + old cap)

Still occurs without picom running.

jwijenbergh:
> > Could you let me know if it improves without picom?
> > Capping events is a bit of a hack that I'd still like to avoid if possible. Worst case we need to do a hybrid approach though (this pr + old cap)
> 
> Still occurs without picom running.

Thanks. What is the value you used before and what is your refresh rate?

purplebar0:
> > > Could you let me know if it improves without picom?
> > > Capping events is a bit of a hack that I'd still like to avoid if possible. Worst case we need to do a hybrid approach though (this pr + old cap)
> > 
> > 
> > Still occurs without picom running.
> 
> Thanks. What is the cap value you used before and what is your refresh rate?

I had it set to 144 as per my monitor's refresh rate.

jwijenbergh:
Cool maybe I should reintroduce the cap but have it be off by default what do you think? As the original issue was regarding performance which should be fixed already. Or I do something fancy by looking at your refresh rate but that would have to be dynamic as well then :/

purplebar0:
> Cool maybe I should reintroduce the cap but have it be off by default what do you think? As the original issue was regarding performance which should be fixed already. Or I do something fancy by looking at your refresh rate but that would have to be dynamic as well then :/

I think bringing back the custom cap would be perfect. Just to avoid the hacky nature of the cap, perhaps the best tradeoff would be to simply add it to the default configuration file as a commented line, and have it do nothing by default.

jwijenbergh:
Thanks for the insightful discussion. Could you test #4344

PR #4142, opened 2023-01-29
Correct the description of opacity arg in dropdown

K1ngst0m:
In practice, when opacity is set to one the window is opaque.


elParaguayo:
Nice. Thank you.

PR #4147, opened 2023-02-04
added last_screen attribute to group to allow clamping groups to specific windows

7aske:
Useful when we want to show the group on the last screen it was visible on.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -69,6 +69,7 @@ class _Group(CommandObject):
 
     def _configure(self, layouts, floating_layout, qtile):
         self.screen = None
+        self.last_screen = None
         self.current_layout = 0
         self.focus_history = []
         self.windows = []
@@ -172,6 +173,7 @@ class _Group(CommandObject):
         if screen == self.screen:
             return
         self.screen = screen
+        self.last_screen = screen
         if self.screen:
             # move all floating guys offset to new screen
             self.floating_layout.to_screen(self, self.screen)
@@ -183,6 +185,7 @@ class _Group(CommandObject):
             self.hide()
 
     def hide(self):
+        self.last_screen = self.screen
         self.screen = None
         with self.qtile.core.masked():
             for i in self.windows:


jwijenbergh:
Thanks for the pr! Not sure I understand what this is for. How are you using it in your config?

7aske:
It allows calling `toscreen` of a given grouo with the last screen where that group was seen on. Lets say I have a group with social media on one of my screens and I put something else on that screen. Later when switching to the social media group it will be focused on that same screen where I "left" it. Allows me to move a group to a screen and keep it there. Example from my config would be:

```py
def focus_group_by_name(name):
    """
    Factory method for a focus group by name function that preserves the
    groups current screen if present.
    """

    def inner(q):
        """
        Inner function to be passed to the qtile command object. Captures the
        name of the group to focus.
        """
        for group in qtile.groups:
            if group.name != name:
                continue

            current_screen_index = q.screens.index(q.current_screen)
            if group.screen:
                idx = group.screen.index
                # If current screen is not that of the group we want to focus
                # we first need to switch to that screen.
                if current_screen_index != idx:
                    q.cmd_to_screen(idx)

                return group.cmd_toscreen(screen=idx)

            # If the group has no screen that means that it is not visible
            # on any of the screens.
            idx = q.screens.index(q.current_screen)
            if group.last_screen:
                idx = group.last_screen.index
            return group.cmd_toscreen(screen=idx)

        # No-op if group not found.
        # Shouldn't reach this anyway.
        return lambda: None

    return inner
```

Its a niche use case but the change was so small that I felt like making a PR.

m-col:
I don't think we should have things in the core codebase that aren't used in the codebase, as is, -1 from me.

However, the use case has obvious value and is reminscent of http://docs.qtile.org/en/stable/manual/faq.html#how-can-i-get-my-groups-to-stick-to-screens, specifically the first snippet. I wonder if this behaviour could be achieved simply by tracking the `last_screen` on each group and then updating `Group.toscreen` to make the screen index optional, falling back to the group's last screen. Perhaps that's worth exporing?

PR #4155, opened 2023-02-12
Remove Heisenbugs

elParaguayo:
As mentioned by @tych0, there are potential "Heisenbugs" in our code base where we use `asyncio.create_task` but don't retain a reference to the created task. In those scenarios, the garbage collector could collect the task before it's executed.

This PR stores created tasks in a list and periodically removed completed tasks from that list to limit its size.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -33,7 +33,7 @@ from shutil import which
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, TypeVar, Union
+    from typing import Any, Callable, Coroutine, TypeVar, Union
 
     ColorType = Union[str, tuple[int, int, int], tuple[int, int, int, float]]
     ColorsType = Union[ColorType, list[ColorType]]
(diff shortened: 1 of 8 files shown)


libqtile/utils.py
```diff
@@ -52,6 +52,33 @@
...
 
+# Create a list to collect references to tasks so they're not garbage collected
+# before they've run
+TASKS: list[asyncio.Task[None]] = []
+
+# Interval for removing completed tasks from list
+TASKS_INTERVAL = 10
+
+
+async def clear_tasks():
```
tych0, line 63: Is it conceptually simpler to have our own create_task() wrapper that adds things to a list (dict?) with an additional add_done_callback() to do this removal? Then you only have to do work when it's really necessary.

tych0:
Maybe worth citing the blog post (perhaps this comment is enough) that turned me on to this: https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/

libqtile/utils.py
elParaguayo: Well, yes. All of that is a much neater idea!

Will rework!

elParaguayo:
Test failures are unrelated and should be fixed by #4156

tych0 (approved):

tych0:
This looks great, thanks! Who knows if we've ever hit any of this kind of bug, but now we won't.

PR #4163, opened 2023-02-17
Don't depend on libxcb when not using X11

ramnes:


m-col:
I had a similar PR drafted: https://github.com/qtile/qtile/pull/3773 I think that one is slightly better as it still uses the X backend's keysyms to validate under wayland (which are the same), but does it in a way that avoids importing xcffib. Really the "proper" way would be to delegate to the backend and let it deal with validation, but that's a problem for another day. Perhaps you could take a look at that PR?

m-col:
Actually nvm, there are some things that need ironing out in that PR, specifically changes to `setup.cfg` to ensure that installing works as expected.

Happy to merge this PR in the meantime, LGTM!

tych0 (approved):

ramnes:
I'll merge this one in the meantime by pure convenience then!

> Really the "proper" way would be to delegate to the backend and let it deal with validation, but that's a problem for another day.

I think we just need to make sure that `core` targets the current core in these two lines:

```python
        valid_keys = core.get_keys()
        valid_mods = core.get_modifiers()
```

PR #4164, opened 2023-02-18
Layout XMonad: Fix shrink_up call typo

jwijenbergh:
`Shrink_up` gave an attributeError and then broke `shrink_drown` by leaving gaps

image


elParaguayo:
LGTM

Issue #855, opened 2016-01-28
Resize and move tiled windows with mouse


Sqozz:
Hi all,

i'm currently writing my config script for qtile. As for now, qtile seems to do his job quiet well. But i miss one feature from awesome and that is, that i can resize windows that do **not** float. I'd also appreciate if i could rearrange my tiled windows by - lets lay mod4+mouse-drag. Is that somehow possible in qtile? There a quiet a bunch undocumented features so maybe it's an easy task?

flacjacket:
There currently isn't any functionality in place to move around tiled windows with a mouse binding, it would have to be added as a new feature

Sqozz:
Thanks for your answer! Is resizing somehow possible? Maybe by hooking mouse events and resizing it frame by frame or is resizing in tiled layouts not possible?

ramnes:
I definitely support that issue. Every layout need to have a normalized way of resizing and moving windows, which is not the case currently. That could be done by moving windows size logic to the core of Qtile and only letting layouts mess with placement. There's a lot of work to do for that, though.

andering:
Isn't [https://github.com/qtile/qtile/blob/develop/libqtile/window.py#L1368](https://github.com/qtile/qtile/blob/develop/libqtile/window.py#L1368) method for moving tiled windows by mouse ?

example of use:
```
mouse = [
    Drag([mod], "Button1", lazy.window.set_position(),
         start=lazy.window.get_position())
]
```

flacjacket:
That will set the position of floating windows, this issue is to be able to reconfigure the tiled layout with the mouse.

andering:
Sorry but it's still not clear for me,
is not this issue about :

- [ ] Be able to change positions of windows in tiled layout by mouse

- [ ] Be able to resize windows in tiled layout by mouse, 

If yes,
i really believe first issue has been implemented, where

set_position_floating() - works with floating windows
set_position() - works with tiled layouts

i believe it's this pull #410

Sqozz:
Sorry for the late answer. I think, the first request is actually fixed since you can move a tiled window in its position. Resizing is still a thing, which would be great.

ramnes:
The first request has been implemented, yes, but it needs much more work. See #1334.

pawamoy:
If the first request is "being able to move tiled windows with mouse", then how is it done? I tried the snippet above:

```python
mouse = [
    Drag([mod], "Button1", lazy.window.set_position(),
         start=lazy.window.get_position())
[...]

PR #4166, opened 2023-02-19
Backend + Group: Fix setting tiled position by mouse

jwijenbergh:
It gave an AttributeError:

```python
    self.group.layout.focused = index2
AttributeError: can't set attribute 'focused'
```

Let's just use the layout swap method instead of the hacky .clients (which would only work for clientlist). To avoid duplicated code let's make a swap method in group. This checks if the layout also supports a swap method to avoid erroring out

Fixes part of https://github.com/qtile/qtile/issues/855#issuecomment-540725219 as it now works, however not for any layout of course. For a layout to support this just simply implement a "swap" method. I have added support for `_SimpleLayoutBase` by moving the implementation from the `XMonad` layout

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -265,12 +265,6 @@ class _Group(CommandObject):
         if focus:
             self.focus(win, warp=True, force=force)
 
-    def swap(self, win, win_other):
-        try:
-            self.layout.swap(win, win_other)
-        except AttributeError:
-            logger.warning("This layout does not support swapping tiled windows")
-
     def remove(self, win, force=False):
         self.windows.remove(win)
         hadfocus = self._remove_from_focus_history(win)
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -265,6 +265,12 @@ class _Group(CommandObject):
         if focus:
             self.focus(win, warp=True, force=force)
 
+    def swap(self, win, win_other):
+        try:
+            self.layout.swap(win, win_other)
+        except AttributeError:
+            logger.warning("This layout does not support swapping tiled windows")
+
     def remove(self, win, force=False):
         self.windows.remove(win)
         hadfocus = self._remove_from_focus_history(win)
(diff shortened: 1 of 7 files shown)


elParaguayo:
How would I go about testing this?

jwijenbergh:
You probably have this in your config:

```python
```

Change it to

```python
   Drag([mod], "Button1", lazy.window.set_position(),
```

And drag windows in a tiled layout (I tested xmonadtall)

elParaguayo:
Can we add a test for this?

jwijenbergh:
> Can we add a test for this?

Great idea! added.

elParaguayo:
Swapping works nicely! This will be popular.

jwijenbergh:
> Swapping works nicely! This will be popular.

Great! It was added a while ago but it was broken unfortunately

libqtile/group.py
```diff
@@ -265,6 +265,12 @@ def add(self, win, focus=True, force=False):
         if focus:
             self.focus(win, warp=True, force=force)
 
+    def swap(self, win, win_other):
+        try:
+            self.layout.swap(win, win_other)
+        except AttributeError:
+            logger.warning("This layout does not support swapping tiled windows")
```
m-col, line 272: This will catch unexpected attribute errors that are raised from within the `swap` method, so it might be better explicitly testing for the existence of the method with `hasattr`. That said, for the sake of self-documenting code, you could add the swap method to the base `Layout` class that does nothing expect error, then you don't need to check whether the method is on the current layout:
```py
    def swap(win, win_other) -> None:
        raise NotImplementedError()
```
If it's always used within a command (here it's nested within `Window.set_position()`, then maybe raising the CommandError is more appropriate, and I think it would then get caught by the command machinery which would log a message saying the command couldn't be executed.

m-col (approved):
Neat! just one small comment but otherwise LGTM

m-col (approved):

PR #4168, opened 2023-02-20
Fix Notify widget finalisation

### LINKED ISSUE #4153: Widget `Notify` throws error in logs every time it's used (even if it works fine).
### The issue:

Every time I feed my Notify widget with something (using, for example, `notify-send`), it throws the following into the log file:

````
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run    
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 909, in f    
    func(*args)    
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 132, in real_update    
    self.set_notif_text(notif)    
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 104, in set_notif_text    
    self.text = pangocffi.markup_escape_text(notif.summary)    
    ^^^^^^^^^    
  File "/usr/lib/python3.11/site-packages/libqtile/widget/base.py", line 490, in text    
    ^^^^^^^^^^^^^^^^    
  File "/usr/lib/python3.11/site-packages/libqtile/drawer.py", line 73, in text    
  File "/usr/lib/python3.11/site-packages/libqtile/pangocffi.py", line 118, in set_attributes    
    pango.pango_layout_set_attributes(self._pointer, attrs)    
````

This error is repeated 16 times (always).

I've tried it on a default settings for the widget, so I don't think my configuration is relevant here.

````
~ ❯ qtile --version                                                                                                                                                                                                                                                                                                                                                26s
0.22.2.dev127+gffbf518e
````

### Required:

### PR DESCRIPTION
**Pull request by elParaguayo**
The `Notify` widget was failing to remove callbacks when the widget was finalised. This meant that callbacks were triggered on widgets that no longer had an active `Drawer` resulting in error messages.

This PR simplifies the finalisation process by removing the call for the notifier to release the service name from bus. The name will be released automatically when qtile shuts down so there is no need to explicitly call the method.

Fixes #4153

Note: If a user starts a notification server like `dunst` this should replace qtile's server. However, if `dunst` is subsequently stopped, qtile's server will not be restarted.

### DIFF
--- a/libqtile/notify.py
+++ b/libqtile/notify.py
@@ -25,8 +25,8 @@
 from typing import Any
 
 try:
-    from dbus_next import ReleaseNameReply
     from dbus_next.aio import MessageBus
+    from dbus_next.constants import NameFlag
     from dbus_next.service import ServiceInterface, method, signal
 
     has_dbus = True
@@ -34,7 +34,6 @@ except ImportError:
     has_dbus = False
 
 from libqtile.log_utils import logger
-from libqtile.utils import create_task
 
 BUS_NAME = "org.freedesktop.Notifications"
 SERVICE_PATH = "/org/freedesktop/Notifications"
@@ -137,7 +136,12 @@ if has_dbus:
                     self.bus = await MessageBus().connect()
                     self._service = NotificationService(self)
                     self.bus.export(SERVICE_PATH, self._service)
-                    await self.bus.request_name(BUS_NAME)
+                    await self.bus.request_name(
+                        BUS_NAME,
+                        flags=NameFlag.ALLOW_REPLACEMENT
+                        | NameFlag.REPLACE_EXISTING
+                        | NameFlag.DO_NOT_QUEUE,
+                    )
                 except Exception:
                     logger.exception("Dbus connection failed")
                     self._service = None
@@ -167,23 +171,6 @@ if has_dbus:
                     self.close_callbacks.remove(on_close)
                 except ValueError:
                     logger.error("Unable to remove notify on_close callback. Unknown callback.")
-
-            if not self.callbacks:
-                return create_task(self._release())
-
-        async def _release(self):
-            """
-            If the manager has no more callbacks then we need to release the service name
-            from dbus and reset _service to None (to force subsequent calls to `register` to
-            re-register the name on dbus.)
-            """
-            reply = await self.bus.release_name(BUS_NAME)
-
-            if reply != ReleaseNameReply.RELEASED:
-                logger.error("Could not release %s", BUS_NAME)
-                return
-
-            self._service = None
 
         def add(self, notif):
             self.notifications.append(notif)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by elParaguayo** [comment]
@m-col @jwijenbergh Appreciate your thoughts here: This PR has been confirmed as fixing the linked issue but I'm curious whether you think we should do anything about the last sentence in the first post here. To be honest, we can address it in a separate PR and still merge this (e.g. we could listen for a notification server disappearing from dbus and start the qtile server at that point).

**Comment by m-col** [comment]
> Note: If a user starts a notification server like dunst this should replace qtile's server. However, if dunst is subsequently stopped, qtile's server will not be restarted.

Maybe it would be sufficient to expose a command on the widget that will attempt to register it as the notification service? Then anyone who jumps between notification servers for whatever reason can manually bring it back up again.

What's odd is... why can't we have two notification servers running simultanously? I use `swaync` for mine, and if I start `dunst` it prints:
```
CRITICAL: [dbus_cb_name_lost:1044] Cannot acquire 'org.freedesktop.Notifications': Name is acquired by 'SwayNotificationCenter' with PID '39153'.
```
So I guess it must be part of the spec. I don't quite get why that would be the case. Anyway, that's a tangent.

In regular usage, do we expect people to frequently jump between notification servers? If the config is reloaded, would that be enough to connect the notification server?

elParaguayo:
> What's odd is... why can't we have two notification servers running simultanously? I use `swaync` for mine, and if I start `dunst` it prints:
> 
> ```
> CRITICAL: [dbus_cb_name_lost:1044] Cannot acquire 'org.freedesktop.Notifications': Name is acquired by 'SwayNotificationCenter' with PID '39153'.
> ```
> 
> So I guess it must be part of the spec. I don't quite get why that would be the case. Anyway, that's a tangent.

Correct. The spec just allows for one server. I suspect this helps client as they need to call a method on the server so they need to know which server to connect to.

NB we're being friendly here and we set a flag which says our notification server can be replaced by another. We could just remove that flag and have it behave like Sway instead.

> In regular usage, do we expect people to frequently jump between notification servers?

Probably not!

> If the config is reloaded, would that be enough to connect the notification server?

Not yet but we could make it do that (or at least try to connect subject to any other running servers).

m-col:
> NB we're being friendly here and we set a flag which says our notification server can be replaced by another. We could just remove that flag and have it behave like Sway instead.

I think it's good we set that flag - if the user runs another server, they must want that one to become the current server. Then if they're just testing out servers but actually want the Notify widget running in the meantime, reloading the config or restart qtile are likely the first things they try. So...

> Not yet but we could make it do that (or at least try to connect subject to any other running servers).

This sounds like a good idea

elParaguayo:
That last commit should do that. It tries to restart and, if it can't take ownership of the name on the bus, it logs a warning message.

m-col (approved):
Awesome. LGTM :)

PR #4171, opened 2023-02-23
Add foot to list of terminals.

ftpd:


jwijenbergh (approved):

m-col:
This will cause problems with anybody who has foot installed but runs qtile in X.

It might be enough to check whether `WAYLAND_DISPLAY` is in `os.environ`, then if so, prepend foot to the list of terminals. `DISPLAY` is always going to be there (unless somebody runs `qtile check` from a TTY, which case we're already only considering X terminals). If `WAYLAND_DISPLAY` is, then either `qtile check` was run from within a wayland session or the function was run from the config of a wayland session, so we should consider wayland terminals.

Also, for the sake of maintainence it might be good to stick wayland-exclusive terminals into a separate list:
```py
    if "WAYLAND_DISPLAY" in os.environ:
        # Wayland-only terminals
         test_terminals += [
            "foot"
        ]
    ... # continue to append main list of terminals
```

ftpd:
@m-col sure, it makes sense, thanks!

m-col:
Excellent, thanks @ftpd! I've triggered CI and enabled auto-merge for when the tests are happy :)

ftpd:
@m-col I've run my code through linter (thanks @jwijenbergh for the suggestion). Please restart the pipeline ;-)

jwijenbergh:
> @m-col I've run my code through linter (thanks @jwijenbergh for the suggestion). Please restart the pipeline ;-)

No problem, approved!

PR #4178, opened 2023-02-25
 Don't mask AttributeErrors that arise during startup 

### LINKED ISSUE #4152: qtile gives wrong error message when wayland backend cannot start.
### The issue:

When I start qtile with wayland backend without Xwayland installed, it doesn't start because the backend isn't initialized properly. This is probably som bug in pywlroots (as @m-col suggested on Discord), but it's not even relevant for this bug report.

The problem is: qtile gives very strange and misleading error message when it happens:

```
~ ❯ /usr/bin/qtile --version
0.22.2.dev127+gffbf518e
~ ❯ /usr/bin/qtile start -b wayland
usage: qtile [-h] [-v] {start,shell,top,run-cmd,cmd-obj,check,migrate,help} ...

Did you mean:
/usr/lib/python-exec/python3.11/qtile start -b wayland start
```

It should detect, that the backend has failed and inform user about that, instead of this very generic error message.

Relevant links to the code (also thanks to @m-col for them):
- https://github.com/qtile/qtile/blob/master/libqtile/scripts/main.py#L66
- https://github.com/qtile/qtile/blob/master/libqtile/scripts/main.py#L63-L72

### Required:

m-col:
Commit 1:
```
Proofread qtile command messages

- Normalises sentence case and adds periods
- Uses same message for all `--socket` arguments (a few subcommands use
  it)
- Update command description to drop "pure python", which isn't quite
  true these days
```
Commit 2:
```
 Don't mask AttributeErrors that arise during startup

When our argparsing runs the target subcommand, it does so inside a try
block which appears to have been added to catch the absence of
`options.func` when someone just runs `qtile`, rather than `qtile
start`. This has the side effect of hiding any AttributeErrors that
arise early on in startup e.g. when the core is being created. This
doesn't affect users so much (except those really living on the edge)
but it can be annoying during development. Instead, this does the
attribute check explicitly.
```

Closes #4152

### DIFF
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -60,11 +60,11 @@ def main():
     help_.set_defaults(func=print_help)
 
     options = main_parser.parse_args()
-    try:
+    if func := getattr(options, "func", None):
         log_level = getattr(logging, options.log_level)
         init_log(log_level, log_path=get_default_log())
-        options.func(options)
-    except AttributeError:
+        func(options)
+    else:
         main_parser.print_usage()
         print("")
         print("Did you mean:")
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -34,7 +34,7 @@ def main():
 
     main_parser = argparse.ArgumentParser(
         prog="qtile",
-        description="A full-featured, pure-Python tiling window manager.",
+        description="A full-featured tiling window manager for X11 and Wayland.",
     )
     main_parser.add_argument(
         "-v",
@@ -56,7 +56,7 @@ def main():
     def print_help(options):
         main_parser.print_help()
 
-    help_ = subparsers.add_parser("help", help="Print help information and exit")
+    help_ = subparsers.add_parser("help", help="Print help message and exit.")
     help_.set_defaults(func=print_help)
 
     options = main_parser.parse_args()
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -133,7 +133,7 @@ def check_config(args):
 
 def add_subcommand(subparsers, parents):
     parser = subparsers.add_parser(
-        "check", parents=parents, help="Check a configuration file for errors"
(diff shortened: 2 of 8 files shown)


libqtile/scripts/main.py
```diff
@@ -60,11 +60,11 @@ def print_help(options):
     help_.set_defaults(func=print_help)
 
     options = main_parser.parse_args()
-    try:
+    if func := getattr(options, "func", None):
```
jwijenbergh, line 63: Gotta love the walrus

m-col: I'm a big fan

jwijenbergh:
Note that this does not give a warning when you pass arguments:

```bash
$ bin/qtile -b wayland  
usage: qtile [-h] [-v] {start,shell,top,run-cmd,cmd-obj,check,migrate,help} ...
qtile: error: argument {start,shell,top,run-cmd,cmd-obj,check,migrate,help}: invalid choice: 'wayland' (choose from 'start', 'shell', 'top', 'run-cmd', 'cmd-obj', 'check', 'migrate', 'help')
```

but on master it didn't as well, so this is probably fine

jwijenbergh (approved):

m-col:
In that case, isn't that printed usage message the 'right't thing to print?

jwijenbergh:
> In that case, isn't that printed usage message the 'right' thing to print?

It's debatable, I think that that message was added for backwards compatibility (?). For backend it maybe is not so useful as that came later, but for example if someone has an old config where he does:

```qtile -c pathtoconfig.py``` it would immediately notify him that he needs to add "start" between qtile and -c

m-col:
Ah yep. If it's the same on master I guess the backwards compatible setup was already dropped, right?

jwijenbergh:
> Ah yep. If it's the same on master I guess the backwards compatible setup was already dropped, right?

Yeah this is perfectly fine as is

PR #4180, opened 2023-02-26
Validate config types when loading

m-col:
This largely rewrites the `Config` class to clean up some unused code and validate specified options in user configs using the class's type annotations.

This validation extends the existing `Config.validate` functionality to do basic runtime type checking.

After the type-checking, it carries out the existing check on the values of keys and modifiers.

Only if both stages pass successfully does the loaded config change in any way.

The base Core class is given static methods used to provide `Config` with valid keys and modifiers for value checking. These methods can return a list of valid values, or alternatively (by default), `None`, which lets validation continue without checking. Backends then can then override these with implementations that provide their set of valid key/mod names.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -24,17 +24,21 @@
 from __future__ import annotations
 
 import importlib
+import inspect
+import os
 import sys
+import types
+import typing
+from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import Any, Literal, Optional, Union
 
-if TYPE_CHECKING:
-    from typing import Any
-
-    from typing_extensions import Literal
-
-    from libqtile.config import Group, Key, Mouse, Rule, Screen
-    from libqtile.layout.base import Layout
+import libqtile
+from libqtile.backend import get_core
+from libqtile.config import Group, Key, Mouse, Rule, Screen
+from libqtile.layout.base import Layout
+from libqtile.log_utils import logger
+from libqtile.resources import default_config
 
 
 class ConfigError(Exception):
@@ -52,6 +56,8 @@ from libqtile.layout.base import Layout
 
 class Config:
     # All configuration options
+    # While we support Python 3.9 we can't use the | operator in types as annotations
+    # are evaluated at runtime for config validation.
     keys: list[Key]
     mouse: list[Mouse]
     groups: list[Group]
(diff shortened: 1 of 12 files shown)


libqtile/confreader.py
```diff
@@ -118,43 +120,118 @@ def _reload_config_submodules(self, path: Path) -> None:
...
+            value = settings[key]
+
+            if self._check_type(valid_type, value):
+                new_config[key] = value
+            else:
+                errors.add(key)
+
+        if errors:
+            keys = ", ".join(errors)
+            raise ConfigError(f"Invalid types: {keys}")
```
elParaguayo, line 167: Are we sure our typing is robust enough for us to raise a `ConfigError` on an invalid type? Great if the answer is yes!

libqtile/resources/default_config.py
```diff
@@ -186,6 +194,10 @@
 # When using the Wayland backend, this can be used to configure input devices.
 wl_input_rules = None
 
+# This can be used instead of `screens` to split a physical monitor into multiple
+# 'screens': http://docs.qtile.org/en/stable/manual/config/screens.html#fake-screens
```
elParaguayo, line 198: One for another PR but we should add much more commentary in our default config.

libqtile/scripts/check.py
```diff
@@ -119,16 +120,22 @@ def check_config(args):
 
         # are the top level config variables the right type?
         module_name = path.splitext(path.basename(args.configfile))[0]
-        type_check_config_vars(tempdir, module_name)
+        ok &= type_check_config_vars(tempdir, module_name)
```
elParaguayo, line 123: Don't see that operator too often! Is simplicity better than readability?

elParaguayo (commented):
Looks nice. Some very minor comments but I suspect none require a change.

libqtile/confreader.py
m-col: I believe so. Some config options are typed as `Any` so will always pass the validation test - these are the ones where typing isn't fully set up in the most precise way, but that isn't really an issue, except that e.g. `dgroups_key_binder = "not a function"` will get loaded but then later used as a function and raise an error at runtime that won't appear to be related with the users config. For a future PR.

libqtile/resources/default_config.py
m-col: 100% agree

libqtile/scripts/check.py
m-col: The thing here is that we want to all of these checks but store whether any have failed to print a message and return with a non-0 error code. Is there a more readable way to do this? Happy to change it.

Something like this?:
```diff
diff --git a/libqtile/scripts/check.py b/libqtile/scripts/check.py
index 2761cb61..defde918 100644
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -109,7 +109,7 @@ def check_deps() -> None:
 def check_config(args) -> None:
     check_deps()
 
-    ok = True
+    failed = False
     print("Checking Qtile config at:", args.configfile)
 
     # need to do all the checking in a tempdir because we need to write stuff
@@ -120,10 +120,12 @@ def check_config(args) -> None:
 
         # are the top level config variables the right type?
         module_name = path.splitext(path.basename(args.configfile))[0]
-        ok &= type_check_config_vars(tempdir, module_name)
+        if not type_check_config_vars(tempdir, module_name):
+            failed = True
 
         # are arguments passed to qtile APIs correct?
-        ok &= type_check_config_args(tmp_path)
+        if not type_check_config_args(tmp_path):
+            failed = True
 
     # can we load the config?
     config = confreader.Config(args.configfile)
@@ -134,7 +136,7 @@ def check_config(args) -> None:
         print(e)
         print("\nYour config cannot be loaded by Qtile.")
 
-    if not ok:
+    if failed:
         sys.exit(1)
 
 
```

libqtile/confreader.py
```diff
@@ -118,43 +120,118 @@ def _reload_config_submodules(self, path: Path) -> None:
...
+                if k.key.lower() not in valid_keys:
+                    raise ConfigError("No such key: %s" % k.key)
+                for m in k.modifiers:
+                    if m.lower() not in valid_mods:
+                        raise ConfigError("No such modifier: %s" % m)
+        if valid_mods is not None:
+            for ms in self.mouse:
+                for m in ms.modifiers:
+                    if m.lower() not in valid_mods:
+                        raise ConfigError("No such modifier: %s" % m)
```
jwijenbergh, line 237: Might as well use convert this to an f string if we're at it :)

jwijenbergh:
Just a general comment, would it be possible to skip over invalid values instead? So if we have an invalid key in the config it will warn in the log file/notify the user through a notification but it would still load the valid ones

libqtile/scripts/check.py
elParaguayo: That's certainly more readable but I'm happy to stick with your original.

m-col:
So load as many keys as possible rather than failing entirely in an all or nothing approach? Yeah that sounds great

jwijenbergh:
> So load as many keys as possible rather than failing entirely in an all or nothing approach? Yeah that sounds great

Exactly. Same for other things ideally, widgets, bars, screens

m-col:
Going to close this for now, and maybe revisit it in the future. There are some things that would need ironing out (circular imports) and the value isn't huge, plus some good testing would be good to have, and i want to move on to other things.

Issue #4176, opened 2023-02-25
Wrong window position and shifted mouse input when leaving fullscreen


fdev31:
### The issue:

- Xorg / Archlinux / Nvidia
- Qtile 0.22.2.dev121+g4f6ff94d
- No relevant log

### Pre-conditions

- Youtube opened in Google chrome or Brave
- Happens at least with bsp, colums & max layouts
- Bottom bar on screen
- ```auto_fullscreen = False```

### Scenario

- maximize video from youtube's icon
- maximize the window (mod+f)
- un-maximize video from youtube's icon

*Expected result:* the window is displayed according to layout and mouse pointer works as expected, something like this:
![image](https://user-images.githubusercontent.com/238622/221352575-3278a892-cea2-4048-9ec7-f352bc3bb02a.png)

*Current result:* the window is displayed shifted (a margin appears at the top) and the mouse pointer actions seem to be shifted by the same number of pixels (actions don't match visual representation):

![image](https://user-images.githubusercontent.com/238622/221352667-a33b2345-1954-4972-92b7-047859122ee5.png)

Note: some actions (eg: moving cursor to another screen in multi-screen setup) will fix it.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

jwijenbergh:
seems like a chromium bug, but let's not assume that. I will see if I can implement a workaround

jwijenbergh:
@fdev31 great bug report. Does #4181 fix it?

fdev31:
Yes it does !

PR #4181, opened 2023-02-26
X11 Window: Workaround unfullscreen redrawing

jwijenbergh:
See the comment in the code. It's a workaround and I'm not entirely sure yet if it's maybe something we do wrong in Qtile.

Fixes https://github.com/qtile/qtile/issues/4176

I don't think this can be tested properly

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -649,10 +649,17 @@ class _Window:
     def update_state(self):
         triggered = ["urgent"]
 
+        state = self.window.get_net_wm_state()
+
         if self.qtile.config.auto_fullscreen:
             triggered.append("fullscreen")
-
-        state = self.window.get_net_wm_state()
+        # This might seem a bit weird but it's to workaround a bug in chromium based clients not properly redrawing
+        # The bug is described in https://github.com/qtile/qtile/issues/4176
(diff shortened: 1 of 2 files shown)


jwijenbergh:
Patching dwm to allow to fullscreen only by keybinding (so auto fullscreen False basically) gives the same bug, so this workaround will stay for now

libqtile/backend/x11/window.py
```diff
@@ -649,10 +649,17 @@ def update_hints(self):
...
+
         if self.qtile.config.auto_fullscreen:
             triggered.append("fullscreen")
-
-        state = self.window.get_net_wm_state()
+        # This might seem a bit weird but it's to workaround a bug in certain clients not properly redrawing
+        # The bug is described in https://github.com/qtile/qtile/issues/4176
+        # This only happens when auto fullscreen is set to false because we then do not obey the disable fullscreen state
+        # So here we simply re-place the window at the coordinates which will magically solve the issue
+        # This only seems to be an issue with unfullscreening, thus we check if we're fullscreen and the window wants to unfullscreen
```
m-col, line 660: Should we change "certain clients" to specifically name chromium? That way we can be more confident in the future about undoing this if we want to for whatever reason, because we'll only need to test the one client

jwijenbergh: good point. Done!

m-col (approved):

PR #4192, opened 2023-03-13
TaskList: add `markup_focused_floating` option

Akmadan23:
This fixes the mutual exclusion of `markup_focused` and `markup_floating` for windows both focused and floating.


Akmadan23:
I noticed that `window is window.group.current_window` as first check marks minimized windows as floating and focused, so I'm closing this PR and will reopen a new correct one

PR #4194, opened 2023-03-14
add ability to set custom "Undefined" status key value to `Mpd2Widget`

mrtnvgr:
This PR implements "undefined_value" option to `Mpd2Widget`.

Since it is very opinionated, we can allow user to change this value (like a `idle_message`).

_In my opinion, "Undefined" looks bad in the status bar, changing it to something like "[-]" would be better.
I see it regularly, because a lot of my music doesn't have any tags._

### DIFF
--- a/test/widgets/test_mpd2widget.py
+++ b/test/widgets/test_mpd2widget.py
@@ -18,7 +18,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import sys
-from collections import defaultdict
 from types import ModuleType
 
 import pytest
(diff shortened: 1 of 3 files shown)


test/widgets/test_mpd2widget.py
```diff
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import sys
+from collections import defaultdict
```
elParaguayo, line 21: Why is this needed? It doesn't look like your test calls this.

elParaguayo (commented):
Thanks for this. Looks ok to me other than one minor comment.

test/widgets/test_mpd2widget.py
mrtnvgr: > Why is this needed? It doesn't look like your test calls this.

Oh, my bad! I was testing some stuff and forgot to remove this.

mrtnvgr:
@elParaguayo

PR #4203, opened 2023-03-18
Update ReadTheDocs dependencies

elParaguayo:
Use packaged version of cairocffi and xcffib to prevent mismatches with pip versions.

EDIT: Hmm more to it than this.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -22,6 +22,10 @@ build:
     - xvfb
     - dbus-x11
     - libnotify-bin
+  jobs:
+    pre_install:
+      - pip install xcffib wheel
+      - pip install --no-cache --upgrade --no-build-isolation cairocffi
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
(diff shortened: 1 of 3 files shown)


tych0:
I see:

> Get:376 http://archive.ubuntu.com/ubuntu focal/universe amd64 python3-cairocffi all 0.9.0-4 [67.1 kB]

in the logs, so it's using a pretty old version of cairocffi that does satisfy your version constraints. The logs have:

> ./scripts/ffibuild
> Building Pango interface
>    Failed!

so it's not obvious why it failed, but my guess is that cairocffi is too old and incompatible. Perhaps you need `=1.5.0` or something?

elParaguayo:
Sorry - ignore the RTD logs. I'd tried fixing the error by using the packaged cairocffi version rather than the pip package.

I'm talking about the failures in the main CI test. That's installing 1.5.0 with pip and is failing as it says there's no xcb support.

elParaguayo:
Ok - this is apparently a pip issue with the pip version in Ubuntu 20.04 being too old to support latest type of packaging. https://github.com/Kozea/cairocffi/issues/211#issuecomment-1479011927

@mcol - should we move to 22.04 now?

m-col:
> @m-col - should we move to 22.04 now?

Absolutely!

m-col:
Existing changes LGTM, also happy for you to bump the ubuntu version for CI.

PR #4207, opened 2023-03-23
Hide notifications & taskbars from tasklist

fdev31:
Notifications & other "technical" windows make the taskbar blink with no added value.
This patch solves this issue.

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -263,6 +263,12 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
 
     @property
     def windows(self):
+        if self.qtile.core.name == "x11":
+            return [
+                w
+                for w in self.bar.screen.group.windows
+                if w.window.get_wm_type() in ("normal", None)
+            ]
         return self.bar.screen.group.windows
 
     def calc_box_widths(self):


elParaguayo:
Thanks. Is checking that hint enough? The spec suggests that hint shouldn't be set if there's a window type hint that makes it obvious that the window should be excluded from taskbar.

fdev31:
Oh ok, I didn't search very deep... and for me it seems to be enough. I should probably check for notifications too explicitly... can you share a ref, or what you think is "obvious" taskbar items ?

elParaguayo:
From http://standards.freedesktop.org/wm-spec/latest/ar01s05.html#idm139870829988448
```
_NET_WM_STATE_SKIP_TASKBAR indicates that the window should not be included on a taskbar. This hint should be requested by the application, i.e. it indicates that the window by nature is never in the taskbar. Applications should not set this hint if _NET_WM_WINDOW_TYPE already conveys the exact nature of the window. 
```

The window types are listed on that same page. One approach could be to exclude anything that's not `_NET_WM_WINDOW_TYPE_NORMAL`.

elParaguayo:
Sorry - I've just realised this will break on Wayland as `get_wm_type()` is only for X11.

fdev31:
> Sorry - I've just realised this will break on Wayland as `get_wm_type()` is only for X11. We'd need to check the backend before applying this filter.

Ok, so for wayland I would do something similar to the first version / commit ? Or skip it ?

Also, it looked like tests are failing but not because of this PR...

elParaguayo:
I think, for Wayland, you're sticking with `return self.bar.screen.group.windows` (but @m-col may know if there's a way to filter out certain window types).

As for test failures, if you fetch the latest upstream changes and then rebase your PR onto the updated master then those failures should disappear.

libqtile/widget/tasklist.py
```diff
@@ -263,7 +263,14 @@ def get_taskname(self, window):
...
     def windows(self):
-        return self.bar.screen.group.windows
+        if self.qtile.core.name == "wayland":
+            return self.bar.screen.group.windows
+        else:
+            return [
+                w
+                for w in self.bar.screen.group.windows
+                if w.window.get_wm_type() in ("normal", None)
+            ]
```
m-col, line 273: Maybe reverse the if statement because X11 is the special case here (if/when a new backend comes along, the if statement would use the same code path here as wayland):
```suggestion
        if self.qtile.core.name == "x11":
            return [
                w
                for w in self.bar.screen.group.windows
                if w.window.get_wm_type() in ("normal", None)
            ]
	return self.bar.screen.group.windows
```

m-col (approved):
small suggestion but lgtm :)

PR #4218, opened 2023-03-31
Add TaskList widget test

elParaguayo:


Akmadan23:
Isn't this missing all tests regarding `markup_*` options? And now that I look at it aren't `txt_*` options somewhat redundant since they can all be declared as `markup_*`? If so we could completely remove all `txt_*`s and change default markups from `None` to the various `"V {}"`, `"[] {}"`, `"_ {}"`, etc. It would be a little breaking change but also very easy to fix for those using `txt_*` options in their configs.

elParaguayo:
> Isn't this missing all tests regarding `markup_*` options? 

Yes. I probably should add those.

>And now that I look at it aren't `txt_*` options somewhat redundant since they can all be declared as `markup_*`? If so we could completely remove all `txt_*`s and change default markups from `None` to the various `"V {}"`, `"[] {}"`, `"_ {}"`, etc. It would be a little breaking change but also very easy to fix for those using `txt_*` options in their configs.

Maybe... I assume the idea of separating them was to make this a bit easier for users. I think "somewhat redundant" is probably a good description here.

Ideally, this widget should be more consistent with other text messages which use a `format` parameter to set the text and `markup` to set whether markup code needs to be parsed (this is `True` by default for text widgets).

Akmadan23:
> Ideally, this widget should be more consistent with other text messages which use a `format` parameter to set the text and `markup` to set whether markup code needs to be parsed (this is `True` by default for text widgets).

Yeah, maybe instead of both `txt_*` and `markup_*` something like `fmt_*` would be clearer and more consistent with other widgets, and although it would be even more breaking I think it's worth it

elParaguayo:
The `text_` fields do seem to be redundant as they're disregarded when using the equivalent `markup_` field.

I'm still inclined to push on with this test so we can then get your PR merged.

Akmadan23:
> I'm still inclined to push on with this test so we can then get your PR merged.

Yes, once this will be merged I'll add tests to my PR and then I'll maybe open a new one with all the changes necessary to make the transition and we'll discuss that

test/widgets/test_tasklist.py
```diff
@@ -0,0 +1,179 @@
...
+from libqtile.config import Screen
+from libqtile.confreader import Config
+from libqtile.widget.tasklist import TaskList, has_xdg
+
+
+@pytest.fixture
+def override_xdg(request):
+    return getattr(request, "param", False)
+
+
```
m-col, line 33: Could we avoid having attributes at normal runtime that only exist for tests? We could monkey patch the widget here and uses that for tests to get the data that we want, perhaps something like:
```python
class TestTaskList(TaskList):
    def __init__(self, *args, **kwargs):
        TaskList.__init__(self, *args, **kwargs)
        self._text = ""

    def calc_box_widths(self):
        ret_val = TaskList.calc_box_widths(self)
        self._text = "|".join(self.get_taskname(w) for w in windows)
        return ret_val

    def info(self):
        info = base._Widget.info(self)
        info["text"] = self._text
        return info
```

m-col (approved):
Made a small suggestion, happy to merge with that if the tests are valid (even if there are more that could be added) :)

test/widgets/test_tasklist.py
elParaguayo: Absolutely. Nice idea.

elParaguayo:
Two of the tests are a bit flaky so I'm going to take a closer look at them.

elParaguayo:
Don't understand why those 2 tests are consistently failing here.  Will mark with `xfail` so they don't cause an overall failure.

m-col:
expected failure? Or just pytest.skip?

elParaguayo:
Not expected (which, I know, is what `xfail` is for). I've run the tests 40 times on my laptop and they pass every time. They've also worked on the other runners. So I'm using `xfail` to mean "flaky"!

m-col:
Fair :)

elParaguayo:
Still need to make your suggested change. Will need to be tomorrow now!

PR #4228, opened 2023-04-08
Fix #4123 by changing keyboard layout regex on X11 backend

### LINKED ISSUE #4123: Configured keyboards limited to 3
### The issue:

In my config I've defined 4 keyboard layouts, however in the widget on the bar, it never reaches the fourth layout when left-clicking. When the third layout is removed (moving the fourth into the third position) it is suddenly accessible so it's not a problem with the layout itself. I don't notice any logs that would apply to this.

```
widget.KeyboardLayout(
    font = defaults['font'],
    fontsize = defaults['fontsize'],
    configured_keyboards = ['us', 'es', 'semimak-jq', 'mtgap'],
    display_map = { #makes everything lowercase
        'us': 'us',
        'es': 'es',
        'workman': 'wm',
        'semimak': 'sm',
        'mtgap': 'mt',
    }
),
```

From my config, I defined us, es, semimak-jq, and mtgap, but I can never rotate onto mtgap unless I remove semimak-jq. When I manually set the layout with setxkbmap, it is correctly displayed as 'mt' in the widget, I just can't rotate onto it via left-click on the widget.

Qtile Version: 0.22.1, X11

### Required:

Malex:


elParaguayo:
Sorry for the delay.

Thanks for submitting this. Regex change looks good.

Were the other changes caused by `black` formatting?

Malex:
looking around my editor settings, it seems it defaulted to yapf for this project instead. I'll take a look on how to fix the formatting defaults on my editor and switch to black for this project only

Malex:
@elParaguayo fixed

PR #4233, opened 2023-04-13
fix typo

jgarte:


elParaguayo:
Thanks for this. 

The PR looks fine. Docs seem to have failed to build but I think that's because your local git version hasn't been updated to include some recent commits. There's no way your PR is breaking the docs so I'm ok merging this now.

PR #4234, opened 2023-04-13
save/restore window geometry for fullscreen and maximized modes

Yonnji:
It's useful in floating only workspaces. When you open YouTube video in fullscreen mode in browser and then exit it, the browser's window geometry will be restored back. The same applies to other apps and maximized windows.

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -448,10 +448,7 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
             )
 
             if self._float_state not in (FloatStates.MAXIMIZED, FloatStates.FULLSCREEN):
-                self.base_x = self.x
-                self.base_y = self.y
-                self.base_width = self.width
-                self.base_height = self.height
+                self._save_geometry()
 
             bw = self.group.floating_layout.fullscreen_border_width if self.group else 0
             self._reconfigure_floating(
@@ -462,14 +459,7 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
                 new_float_state=FloatStates.FULLSCREEN,
             )
         elif self._float_state == FloatStates.FULLSCREEN:
-            if self.base_x is not None:
-                self.x = self.base_x
-            if self.base_y is not None:
-                self.y = self.base_y
-            if self.base_width is not None:
-                self.width = self.base_width
-            if self.base_height is not None:
-                self.height = self.base_height
+            self._restore_geometry()
             self.floating = False
 
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


tych0:
Hi, can you rebase and write a test for this?

Yonnji:
@tych0 I did it and added the test.

tych0:
Thanks, look like maybe some copypasta errors causing the test to fail?

https://github.com/qtile/qtile/actions/runs/5936865839/job/16098357165?pr=4234#step:5:225

elParaguayo:
The backends share a lot of coffee so, ideally, this code would be backend-agnostic.

Yonnji:
Added support for wayland, but haven't tested it yet.

>The backends share a lot of coffee so, ideally, this code would be backend-agnostic.

Window-related stuff is separated for some reason, but the most of the code is similar.

elParaguayo:
None of that code looks backend specific so I think we should be trying to get as much of it into the `base.Window` definition. The code to set/use the new variables could be put in new methods in that class and then you just call those methods from the wayland and x11 code. That should reduce unnecessary code duplication here.

(I also edited my and your previous posts to fix the mistake from predictive text on my phone. I meant "code" and didn't notice it typed "coffee"!)

Yonnji:
Moved some code into base.

>I meant "code" and didn't notice it typed "coffee"!

I thought it was some kind of slang, and it was still understandable. xD

libqtile/backend/wayland/window.py
```diff
@@ -153,6 +153,11 @@ def __init__(self, core: Core, qtile: Qtile, surface: S):
         # attribute optional to avoid the destroy().
         self.ftm_handle: ftm.ForeignToplevelHandleV1 | None = None
 
+        self.base_x: int | None = None
+        self.base_y: int | None = None
+        self.base_width: int | None = None
+        self.base_height: int | None = None
```
elParaguayo, line 159: I think you can include this is `base.Window.__init__` as that's called at the start of both the Wayland and X11 window definitions.

elParaguayo (changes requested):
Thanks. This is looking neater. Have made one more suggestion to tidy up a bit further.

libqtile/backend/wayland/window.py
Yonnji: Done.

elParaguayo:
Thanks.

You can ignore the docs failure in the checks. It's unrelated to your PR.

jwijenbergh:
@Yonnji can you rebase?

ramnes (approved):
LGTM

ramnes:
That's neat, thanks!

PR #4246, opened 2023-04-30
Ensure icons are centered in TaskList widget

### LINKED ISSUE #3873: Tasklist Widget Icons not vertically centered
### The issue:

Depending on the icon size, the placement of the icon relative to the text shifts. Here is an example with icon size = 45 and text size = 30.

And here it is with icon size = 30 as well:

A simple fix would be to add a user-defined variable to the `y` calculation for `draw_icon`. I subclassed tasklist and manually modified the `y` value to match my icon size, and that works okay. A more flexible solution would be to modify the y value so it is always drawn such that it is centered. 

#1714 discussed the same issue but was never fixed. 

### Required:

elParaguayo:
The vertical alignment of the icons ignored the `icon_size` property resulting in misaligned icons.

Fixes #3873


elParaguayo:
Confirmed as working here: https://github.com/qtile/qtile/issues/3873#issuecomment-1529119879

To be merged once coveralls and docs are fixed.

PR #4253, opened 2023-05-05
Added default apps in groups

Bennett-Wendorf:
This project adds functionality to groups that allows the user to set a default application for each group, and then spawn the default app for the current group with a single key bind (such as mod+t).

Since my implementation in the `qtile-examples` repo wasn't clean enough for my liking, this is the more integrated version. That alternative implementation can be found [here](https://github.com/qtile/qtile-examples/pull/85). My intention is to present both solutions, and let the maintainers decide which implementation they would prefer.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -717,6 +717,9 @@ class Group:
     label:
         The display name of the group. Use this to define a display name other than name
         of the group. If set to ``None``, the display name is set to the name.
+    default_app:
+        The default application to spawn when the group is selected and ``group.spawn_default_app()``
+        is called. If not specified, nothing will happen when ``group.spawn_default_app()`` is called.
 
     """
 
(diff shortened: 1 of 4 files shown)


Bennett-Wendorf:
Hi, I noticed that one of the workflows is failing on this project, but it looks unrelated to what I changed, and I'm not really sure what's going on. Would someone more knowledgeable be able to take a look and help me figure out how to get the docs building correctly?

elParaguayo:
You probably need to rebase your branch onto the latest master. We fixed a bug in the workflows a few days ago.

Bennett-Wendorf:
I had tried something similar to that initially, but I found that it didn't really abstract out to a module particularly well. While I know that many Qtile users have enough experience programming to make something like this work, my hope with both pull requests was to make it as simple for the end user as possible. If you and the other maintainers think that this functionality doesn't make much sense in core, that's OK too!

jwijenbergh:
> I had tried something similar to that initially, but I found that it didn't really abstract out to a module particularly well. While I know that many Qtile users have enough experience programming to make something like this work, my hope with both pull requests was to make it as simple for the end user as possible. If you and the other maintainers think that this functionality doesn't make much sense in core, that's OK too!

Fair enough.

What do you think about this? @elParaguayo

elParaguayo:
I don't see a need for it in the core code.

jwijenbergh:
Let's work towards a nice example in the qtile-examples repo then. Thanks for the PR @Bennett-Wendorf

PR #4255, opened 2023-05-06
Replacing spawn's double fork

elParaguayo:
@jwijenbergh Here's what I had for replacing the fork.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1222,86 +1222,26 @@ class Qtile(CommandObject):
             cmd = subprocess.list2cmdline(args)
 
         to_lookup = args[0]
-        if shell:
-            args = ["/bin/sh", "-c", cmd]
 
         if shutil.which(to_lookup) is None:
             logger.error("couldn't find `%s`", to_lookup)
             return -1
 
-        r, w = os.pipe()
-        pid = os.fork()
-        if pid < 0:
-            os.close(r)
-            os.close(w)
-            return pid
+        env = os.environ.copy()
+        try:
+            env.pop("VIRTUAL_ENV")
+        except KeyError:
+            pass
 
-        if pid == 0:
-            os.close(r)
-
-            # close qtile's stdin, stdout, stderr so the called process doesn't
-            # pollute our xsession-errors.
-            os.close(0)
-            os.close(1)
-            os.close(2)
-
-            pid2 = os.fork()
-            if pid2 == 0:
-                os.close(w)
-                try:
-                    # if qtile was installed in a virutal env, we don't
-                    # necessarily want to propagate that to children
-                    # applications, since it may change e.g. the behavior
-                    # of shells that spawn python applications
-                    del os.environ["VIRTUAL_ENV"]
-                except KeyError:
-                    pass
-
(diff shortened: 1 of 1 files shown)


libqtile/core/manager.py
```diff
@@ -1219,89 +1223,32 @@ def spawn(self, cmd: str | list[str], shell: bool = False) -> int:
...
-        else:
-            os.close(w)
-            os.waitpid(pid, 0)
+        return proc.pid
 
-            # 1024 bytes should be enough for any pid. :)
-            pid = int(os.read(r, 1024))
-            os.close(r)
-            return pid
+    def _kill_zombies(self):
```
ei-grad, line 1248: "kill" is a bit misleading, there is no `kill` syscall involved and they are already dead after all

maybe "reap"? "relieve"? "get_rid_of"?

elParaguayo: Ha. It was named as a joke - like killing zombies in the movies.

On a more serious note, I suspect this PR won't ever get merged - it was to address a conversation that @jwijenbergh and I had been having.

libqtile/core/manager.py
```diff
@@ -1219,89 +1223,32 @@ def spawn(self, cmd: str | list[str], shell: bool = False) -> int:
...
-                    os.dup2(fd, 2)
+        if not self._spawned_processes:
+            self._zombie_killer = self.call_later(10, self._kill_zombies)
 
-                try:
-                    os.execvp(args[0], args)
-                except OSError:
-                    # can't log here since we forked :(
-                    pass
+        self._spawned_processes.append(proc)
```
ei-grad, line 1244: this would leak `proc` if called in parallel with _kill_zombies

what is qtile threading policy, can it be relied that Qtile methods are only called from a single thread?

libqtile/core/manager.py
```diff
@@ -1219,89 +1223,32 @@ def spawn(self, cmd: str | list[str], shell: bool = False) -> int:
...
 
-                    # Again, this shouldn't happen, but we should just check
-                    if fd > 0:
-                        os.dup2(fd, 0)
+        proc = subprocess.Popen(args, shell=shell, env=env)
 
-                    os.dup2(fd, 1)
-                    os.dup2(fd, 2)
+        if not self._spawned_processes:
+            self._zombie_killer = self.call_later(10, self._kill_zombies)
```
ei-grad, line 1242: probably should be called every time, after adding proc to `_spawining_processes`

otherwize if there is only one `spawn` call (or a single more that 10 seconds after the last process spawned earlier had exit), then it would become alone zombie forever

elParaguayo:
@ei-grad while I appreciate you looking at this, I would suggest you don't waste your time at this stage. This PR is not meant as a serious request at this stage, it was to test an idea that another dev and I had been discussing to see if it addressed a particular issue.

ramnes:
you can't just use Popen as is because we actually want the double fork behavior (i.e. we don't want Qtile to be the parent and we don't want the children to access the TTY), but there may be something to try with the newer options of Popen such as `creationflags=CREATE_NEW_CONSOLE` and/or `start_new_session=True`

elParaguayo:
Thanks. Lots I don't know about here!

Closing this.

PR #4266, opened 2023-05-15
send_notification use default timeout from notification daemon

btwe:
see https://specifications.freedesktop.org/notification-spec/

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -253,7 +253,7 @@ def send_notification(
     title: str,
     message: str,
     urgent: bool = False,
-    timeout: int = 10000,
+    timeout: int = -1,
     id_: int | None = None,
 ) -> int:
     """


elParaguayo:
Thanks for submitting this.

I don't have an issue with your proposed change but we should also add something in our own notification server (https://github.com/qtile/qtile/blob/master/libqtile/notify.py) that checks if the timeout is `-1` and, if it is, set a default timeout for the notification.

Can you add that to this PR?

btwe:
> I don't have an issue with your proposed change but we should also add something in our own notification server (https://github.com/qtile/qtile/blob/master/libqtile/notify.py) that checks if the timeout is `-1` and, if it is, set a default timeout for the notification.
> 
> Can you add that to this PR?

I could do that too. AFAIK, notify.py also sets `timeout=-1` as default.

What is the reason there should be a default timeout in the management layer here? I expect the notification daemon of the system (e.g. dunst) to set default timeouts based on own rules (e.g. urgent flag, content, ...)

elParaguayo:
Qtile provides its own notification server (i.e. you don't need dunst) and so that server needs the default timeout set.

btwe:
It turned out that the notification service is _primarily_ used by qtile's notify-widget. The widget provides already a configuration option [default_timeout](https://github.com/qtile/qtile/blob/6c20b89487522e00d09a747a42862c4389b6246b/libqtile/widget/notify.py#L53). This should be used if users want a default timeout. So no further changes needed here. 

Are you ok with that?

elParaguayo:
I think that's ok.

elParaguayo:
We can merge this and I'll deal with the timeouts in #4280

PR #4268, opened 2023-05-18
Fix #4115 by looking through PPIDs in addition to PIDs if proper config

### LINKED ISSUE #4115: app spawning logic flawed
### The issue:

qtile --version  = 0.22.1

my config part is ; 
`Group("3", spawn="code", matches=[Match(wm_class=["Code"])]),`
 `Group("4", spawn="code", matches=[Match(wm_class=["Code"])]),`

**problem 1-** so what I want is using mutiple vs-code with different gruops so,  if I do this it wont work it spawns both of them in group "3" 

**problem 2-** if i want to run vs-code at group 5, it spawns at group "3" so I had to goto group "3" and shift window to group "5" , i mean what I want is vs-code should spawn at group "3" but after that I should be able to run it directly at another groups 

I think it would be better having 2 kind of "Match()" one is for standart "hard match" meaning it always spawn at specific group the specific app
the other "Match2()" should spawn the app at start and shift that window to a intended group , **but only that** , so that we can run that program at any group later without stucking with that specific group

I hope you understand what I mean 

### Required:

Malex:
Fix https://github.com/qtile/qtile/issues/4115 by looking through PPIDs in addition to PIDs if proper config is set
(I added config because this could lead to a performance hit, also it changes the behavior of additional child window).

To test it I just had to put any program that spawns a window inside a bash script and spawn that one: with property missing (or set to False) it goes to the group 1, with the setting it goes to the proper group


jwijenbergh:
You can maybe also get it from /proc/pid/status

Malex:
> Thanks for this. You can maybe also get it from /proc/pid/status

Sure, from the proc stat file I can get that, that's essentially what I meant with "most likely just copying psutils "Linux" implementation"
Basically have to replicate this for the one field 
https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L1686

I wondered if adding it as a dependency was an option to avoid duplicating the code (and given it's already a dev/test dependency)

Let me know what solution is better for you

libqtile/dgroups.py
```diff
@@ -64,6 +65,19 @@ def func(dgroup):
 
     return func
 
+def _match_ppid(pid: int):
+    """Utility function to match pid againts window.pid or any of its ppid"""
+    
+    def func(w: Window) -> bool:
+        cid = w.get_pid()
+        while cid not in (pid, 0, 1): # if 0 or 1 are reached, we cannot really find parents that are not qtile
```
tych0, line 73: Can we check the cmdline here or something too? This seems like it could accidentally be used to match other windows if e.g. i spawn something from alacritty or whatever.

tych0: > I dont think checking the cmdline would work properly all the times either (name in /stat file generally is the same, but there's no guarantee either).

Sure, it's all heuristics at the end of the day. If it's default-off then maybe that's ok along with a note that it is potentially too aggressive when things are spawned from a terminal.

Malex:
I'll work on a solution without psutils (as suggested above) and with more
straightforward documentation on "don't use this unless needed"

Il lun 5 giu 2023, 18:52 Tycho Andersen ***@***.***> ha
scritto:

libqtile/dgroups.py
Malex: > > I dont think checking the cmdline would work properly all the times either (name in /stat file generally is the same, but there's no guarantee either).
> 
> Sure, it's all heuristics at the end of the day. If it's default-off then maybe that's ok along with a note that it is potentially too aggressive when things are spawned from a terminal.

I'll work on a solution without psutils (as suggested above) and with more
straightforward documentation on "don't use this unless needed"

Malex:
Removed psutil dependency, rewritten the docs

Malex:
Rebased to fix the installation, I dont understand why py3.11 fails now

libqtile/utils.py
```diff
@@ -548,3 +548,20 @@ def on_resume(msg):  # type: ignore
...
+def get_ppid(pid: int) -> int | None:
+    """
+    Gets the parent process id (ppid) of a given process (represented by the pid in input)
+    """
+    try:
+        # Buffering is taken by the psutil default
+        with open("/proc/{0}/stat".format(pid), "rb", buffering=32 * 1024) as f:
+            data = f.read()
+    except OSError:  # Very unlikely we are not able to read the /proc stat file
+        logger.warning("Unable to read stat file for pid={0}".format(pid))
```
elParaguayo, line 562:
```suggestion
        logger.warning("Unable to read stat file for pid=%s", pid)
```
That should fix your failure. The reason for this format is that the string substitution is only performed if the log level is set at warning or higher.

Malex: this is a little embarassing, I completly mis-read the pep8 command 👍

jwijenbergh:
Could you rebase again? I will then give it for a spin, but the code looks good now

Malex:
> Could you rebase again? I will then give it for a spin, but the code looks good now

done

Malex:
(Just occasionally rebasing when a conflict arises, no change is actually being made)

libqtile/config.py
```diff
@@ -714,7 +714,9 @@ class Group:
     label:
         The display name of the group. Use this to define a display name other than name
         of the group. If set to ``None``, the display name is set to the name.
-
+    grab_child_window:
+        Child windows generated by the spanwed program(s) (see ``spawn`` parameter). Defaults to False.
```
m-col, line 718: nit
```suggestion
        Match windows opened by child processes of the ``spawn`` command.
```

m-col (commented):
Do we really need to check all ppids recursively or could we get away with only checking 1 or 2 parents up?

Malex:
> Do we really need to check all ppids recursively or could we get away with only checking 1 or 2 parents up?

2 is certanly needed for the case in the Issue (VSCode opens 2 child before getting to the actual window).
I went for full recursive because I did not know what an educated guess could be. Or we could add it as a configuration option.

Generally speaking I noticed on my instance (looking in htop) it's very rare any process has more than 5 parents anyway.
I'm open to discuss this, not sure what's the best way around this.

PR #4271, opened 2023-05-20
Fix a typo in the TaskList widget documentation

thomasave:
The documentation for the `window_name_location_offset` parameter in the TaskList widget contained `loction` instead of `location`.
There should also be a `the` before `window` in this sentence.
This simple PR fixes those typos.


elParaguayo:
Thanks for this.

PR #4275, opened 2023-05-22
Add items to `qtile_info` root command

### LINKED ISSUE #4274: `qtile_info` root command returns empty dict
### The issue:

This is maybe not that big of a deal, as there's plenty of other things currently on track to be solved, but I'm not one for throwing this under the bed and calling it a day 😄  As a starting point, maybe we could have the path of the currently loaded config? What do you think?

```console
$ qtile cmd-obj -o cmd -f qtile_info
{}
```

### Required:

ervinpopescu:
Items are: `config_path`, `version`, `log_level`. Closes #4274.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -31,6 +31,7 @@ import signal
 import subprocess
 import tempfile
 from collections import defaultdict
+from logging.handlers import RotatingFileHandler
 from typing import TYPE_CHECKING
 
 import libqtile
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for doing this. Could we get the log path in here too? I haven't looked into how easy that would be as, sometimes, qtile won't write to a log file (e.g. if you're using the `xephyr` script, log is to stdout). So, we'd need to check if there's a file handler in the log and, if so, see where it's writing the log.

elParaguayo:
Thanks. Tested and this works nicelyl

PR #4280, opened 2023-05-27
Add different timeouts for notifications

elParaguayo:
Allows users to set different default timeouts for low, normal and urgent priority notifications.

### DIFF
--- a/libqtile/widget/notify.py
+++ b/libqtile/widget/notify.py
@@ -51,7 +51,9 @@ class Notify(base._TextBox):
     defaults = [
         ("foreground_urgent", "ff0000", "Foreground urgent priority colour"),
         ("foreground_low", "dddddd", "Foreground low priority  colour"),
-        ("default_timeout", None, "Default timeout (seconds) for notifications"),
+        ("default_timeout_low", 5, "Default timeout (seconds) for low urgency notifications."),
+        ("default_timeout", 10, "Default timeout (seconds) for normal notifications"),
+        ("default_timeout_urgent", None, "Default timeout (seconds) for urgent notifications"),
         ("audiofile", None, "Audiofile played during notifications"),
         ("action", True, "Enable handling of default action upon right click"),
         (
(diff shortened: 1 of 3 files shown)


libqtile/widget/notify.py
```diff
@@ -51,7 +51,9 @@ class Notify(base._TextBox):
     defaults = [
         ("foreground_urgent", "ff0000", "Foreground urgent priority colour"),
         ("foreground_low", "dddddd", "Foreground low priority  colour"),
-        ("default_timeout", None, "Default timeout (seconds) for notifications"),
+        ("default_timeout_low", 5, "Default timeout (seconds) for low urgency notifications."),
+        ("default_timeout", 10, "Default timeout (seconds) for normal notifications"),
```
jwijenbergh, line 55: This means that it now has 10 seconds instead of `None` right? Could you add a change to the changelog stating this?

jwijenbergh:
Just some small comments, code looks good :)

libqtile/widget/notify.py
```diff
@@ -94,6 +96,14 @@ def _configure(self, qtile, bar):
         if notifier is None:
             logger.warning("You must install dbus-next to use the Notify widget.")
 
+        # Create a tuple of our default timeouts. Urgency is an integer of 0-2 so they will work
```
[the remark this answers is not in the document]
elParaguayo: I used the `urgent` name as this was consistent with the `foreground/background_urgent` properties. I'd prefer not to change this.

libqtile/widget/notify.py
elParaguayo: That's correct. I'll put a not in the changelog.

elParaguayo:
We also need to do better handling of timeouts. The issue, as I see it, is that when a notification arrives it sets a timeout to call the `clear` method but there's no notification ID attached to that callback. So, if a user, scrolls to another notification, the callback is not cancelled and a different notification could end up being cleared.

I think that should be a separate PR and not something that should stop this one.

jwijenbergh (approved):

Issue #4284, opened 2023-05-31
Yet another libcairo.so.2 issue


otval229:
### The issue:

I can't start qtile

Last error log:

---
```
2023-05-31 18:15:06,780 ERROR libqtile start.py:start():L92 Qtile crashed
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/libqtile/scripts/start.py", line 90, in start
    q.loop()
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 198, in loop
    asyncio.run(self.async_loop())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 223, in async_loop
    self.load_config(initial=True)
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 136, in load_config
    self._process_screens(reloading=not initial)
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 359, in _process_screens
    scr._configure(self, i, x, y, w, h, grp, reconfigure_gaps=reconfigure_gaps)
  File "/usr/lib/python3.11/site-packages/libqtile/config.py", line 412, in _configure
    i._configure(qtile, self, reconfigure=reconfigure_gaps)
  File "/usr/lib/python3.11/site-packages/libqtile/bar.py", line 277, in _configure
    self.drawer.clear(self.background)
  File "/usr/lib/python3.11/site-packages/libqtile/backend/x11/drawer.py", line 177, in clear
    self._check_xcb()
  File "/usr/lib/python3.11/site-packages/libqtile/backend/x11/drawer.py", line 122, in _check_xcb
    self._xcb_surface = self._create_xcb_surface()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/backend/x11/drawer.py", line 88, in _create_xcb_surface
    surface = cairocffi.XCBSurface(
[...]

PR #4285, opened 2023-05-31
Add missing --no-build-isolation flags in the docs and dev.sh

ramnes:
Rationale: https://github.com/qtile/qtile/issues/4284#issuecomment-1570668454


elParaguayo:
I'd been putting this off as I was hoping the cairocffi packagers would fix this but they haven't been able to do that yet.

One point: if you're installing qtile with `pip`, I think that also needs the `--no-build-isolation` option as, otherwise, it also fails to see the `cairocffi` installation. We've have had some users saying it's needed.

ramnes:
Let's follow here: https://github.com/Kozea/cairocffi/issues/214

tych0:
idk, I just hit this myself a few days ago. I think it might be worth merging this PR while we wait for the packaging situation to be resolved. It was very non-obvious to me that pip was doing this kind of build isolation all of the sudden.

PR #4289, opened 2023-06-04
In line

tych0:
derp derp, should have sent these when I released the new xcffib!

(context: https://github.com/Kozea/cairocffi/issues/214)


tych0:
Erg. I have no idea what that CI failure is. I'll try to look at it this week.

tych0:
```
cffi.VerificationError: ffi object <cffi.api.FFI object at 0x7fbd1c481c10> includes <cffi.api.FFI object at 0x7fbd1c4754c0>, but the latter has not been prepared with set_source()
```

I guess we need to port our xcursors code to the in-line mode as well. I'll do that if nobody beats me to it (i'm afk after this for the day).

dev.sh
```diff
@@ -15,7 +15,7 @@ pip install -U pip setuptools wheel
 
 # https://github.com/qtile/qtile/issues/994#issuecomment-497984551
 echo "Installing xcffib then cairocffi..."
-pip install 'xcffib >= 0.10.1' && pip install --no-cache-dir 'cairocffi >= 0.9.0'
+pip install 'xcffib >= '1.4.0' && pip install --no-cache-dir 'cairocffi >= 0.9.0'
```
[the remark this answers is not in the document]
tych0: yes, good point, thanks.

ramnes:
It still won't pass if pip decides to install qtile before xcffib, and I've come to no longer trust pip on this matter.

tych0:
Surely it will install dependencies before the actual thing it's trying to install right? (right?)

ramnes:
Well as you know it installs cairocffi before xcffib with `pip install cairocffi[xcb]` so I'm quite paranoid now. :D 

Plus with the multiple backends of Qtile we may end up going the extra dependencies route as well with something like `qtile[x11,wayland]`...

ramnes:
Alright so the problem lies here: https://github.com/qtile/qtile/blob/72f6b637efd6e28e024929f6f36dbec26ecc3da5/libqtile/backend/x11/xcursors_ffi_build.py#L30

xcffib must call `set_source` but it doesn't: https://github.com/tych0/xcffib/blob/517169a2fb6dcc127059f581d3890e0861e2e950/module/ffi.py#L246

tych0:
Those are the docs for out-of-line, though. We're using the in-line API now, so I think we just have to port everything downstream of xcffib to use in-line as well?

tych0:
Can you send a patch? I'm not sure I understand how it's supposed to work.

tych0:
Oh, well. Don't we want to port it anyway so people don't have to have a C compiler installed?

ramnes:
Yes, definitely, it probably makes sense to go in-line all the way up anyway since we've made the change on both xcffib and cairocffi already.

(I did start to tinker with cffi and Qtile to see if I was talking nonsense or if it actually worked, but my Gentoo laptop has weird behaviors and is really out of date so I'm currently recompiling my kernel instead...)

tych0:
Ok. I *think* this is the CI failure I expect to see: we're getting an ImportError not because of the missing symbol, but because the `in-line` branch of cffi moved ffi_build.py to ffi.py. Once they do a release and fix that, we can pick it up here with version constraints and I think it all should work.

tych0:
Oh, no. That would only explain our import error. This is cairocffi directly. So I'm not sure what's going on...

libqtile/backend/x11/xcursors.py
```diff
@@ -105,7 +105,7 @@ def finalize(self):
     def _setup_xcursor_binding(self):
         try:
             xcursor = ffi.dlopen("libxcb-cursor.so.0")
-        except OSError:
+        except Exception:
```
tych0, line 108: Can you log the contents of the exception to see what it actually is?

ramnes: would we see the log in the CI tests logs? IIRC pytest catches everything, but maybe not warnings yeah

ramnes:
Yeehaw, the tests pass! :tada: 

We just need to fix mypy, get @liZe to release cairocffi, and I think we're good for a merge / release.

ramnes:
Everything passes! We just need a cairocffi release now. :)

elParaguayo:
Amazing work guys. I'll try to take a proper look later.

Also, not one for this PR but I'll update the readthedocs build to see if can reflect these changes there too. There are some other docs dependencies which I want to look at which is why it can be separate from this.

tych0:
Thank you @ramnes for finishing this off. LGTM.

elParaguayo:
@ramnes this all looks fine to me.

The only issue I see is that most distros are unlikely to have packaged the latest `xcffib` and `cairocffi` releases which will cause some issues for those people who have been relying on packages, rather than `pip`, for those modules. However, I don't think that's a good enough reason for us to delay this.

elParaguayo (approved):

tych0:
More recent pips get a lot more aggressive with their `--break-system-packages` arguments, so I guess people on "modern" distros are probably using virtualenvs unless they know what they're doing. Hopefully (famous last words) it won't be too big a deal.

ramnes:
Let's go then!

PR #4292, opened 2023-06-05
Fix two bugs in MonadThreeCol

### LINKED ISSUE #4288: Can't set the margin as [N, E, S, W] on MonadThreeCol
### The issue:

While other layouts (e.g. Colums) accept the margin set to a list of margins [N, E, S, W], MonadThreeCol seems to have a problem with it. When I try it everything about the layout breaks
image
The photo clearly shows that qtile doesn't understand the layout we set. I think it should work like any other layout

### Required:

### LINKED ISSUE #4311: Gap on second window monadthreecol
### The issue:

I wanted to toggle the bar off with two windows open in MonadThreeCol layout.
It successfully hid the bar but left a gap on only the secondary window above and below but not the main window. It works as expected in the MonadTall layout.
I expected both windows to successfully span the full screen from top to bottom.

### Required:

elParaguayo:
Fixes two bugs in`MonadThreeCol` layout 

- bug where it would crash if `margin` was set as a list (fixes #4288)
- bug where `screen_rect` changes but secondary windows are not resized (fixes #4311)

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -196,8 +196,18 @@ class MonadTall(_SimpleLayoutBase):
         if self.single_margin is None:
             self.single_margin = self.margin
         self.relative_sizes = []
-        self.screen_rect = None
+        self._screen_rect = None
         self.default_ratio = self.ratio
+
+    # screen_rect is a property as the MonadThreeCol layout needs to perform
+    # additional actions when the attribute is modified
+    @property
+    def screen_rect(self):
+        return self._screen_rect
+
+    @screen_rect.setter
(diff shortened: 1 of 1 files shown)


elParaguayo:
Confession time... The second commit is a bit of a hack. The issue is that the `MonadThreeCol` stores pixel heights in `self.relative_sizes` whereas `MonadTall/Wide` store a ratio. If `MonadThreeCol` did the same then I don't think this fix is needed. However, I don't know the workings of these layouts well enough to be able to make that conversion!

jwijenbergh:
I think this is fine

jwijenbergh (approved):

Issue #4287, opened 2023-06-04
pip installs wrong version of qtile (0.19.0)


elParaguayo:
### The issue:

Trying to install qtile in a clean virtual environment gives the following:
```
Collecting qtile
  Using cached qtile-0.22.1.tar.gz (488 kB)
  Preparing metadata (pyproject.toml) ... done
Discarding https://files.pythonhosted.org/packages/c5/6d/662b20c4f19e4c68e44564e4c5b51e41d477722cfacfaa178acb548c40f2/qtile-0.22.1.tar.gz (from https://pypi.org/simple/qtile/): Requested qtile from https://files.pythonhosted.org/packages/c5/6d/662b20c4f19e4c68e44564e4c5b51e41d477722cfacfaa178acb548c40f2/qtile-0.22.1.tar.gz has inconsistent version: expected '0.22.1', but metadata has '0.0.0'
  Using cached qtile-0.22.0.tar.gz (488 kB)
  Preparing metadata (pyproject.toml) ... done
Discarding https://files.pythonhosted.org/packages/31/26/60f423a0221d18fd37dac826a1070765a58e65133e23d69855f1695c9de0/qtile-0.22.0.tar.gz (from https://pypi.org/simple/qtile/): Requested qtile from https://files.pythonhosted.org/packages/31/26/60f423a0221d18fd37dac826a1070765a58e65133e23d69855f1695c9de0/qtile-0.22.0.tar.gz has inconsistent version: expected '0.22.0', but metadata has '0.0.0'
  Using cached qtile-0.21.0.tar.gz (475 kB)
  Preparing metadata (pyproject.toml) ... done
Discarding https://files.pythonhosted.org/packages/a4/7e/4f3921eda7e3912aa7487b0455e1993a638d99cf6075205de3cf8a13f291/qtile-0.21.0.tar.gz (from https://pypi.org/simple/qtile/): Requested qtile from https://files.pythonhosted.org/packages/a4/7e/4f3921eda7e3912aa7487b0455e1993a638d99cf6075205de3cf8a13f291/qtile-0.21.0.tar.gz has inconsistent version: expected '0.21.0', but metadata has '0.0.0'
  Using cached qtile-0.20.0.tar.gz (439 kB)
  Preparing metadata (pyproject.toml) ... done
[...]

PR #4307, opened 2023-06-15
Fix the release workflow

ramnes:
* manylinux_2_24 is EOL and broken, so we upgrade to manylinux_2_28
* manylinux_2_28 uses AlmaLinux 8, that's dnf-based, so we update install accordingly
* xcffib isn't needed at build time anymore since we moved to inline (#4289), but we still need cffi to build PulseAudio stuff, so we just install cffi
* `auditwheel repair --plat manylinux2014_x86_64` now fails with the following error, so I guess we have no other choice than removing it and only build for the latest Linux versions?

```
auditwheel: error: cannot repair "dist/qtile-0.1.dev1+gc02152d-cp310-cp310-linux_x86_64.whl" to "manylinux2014_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
```

Context: I was trying to release a beta version, but the release action is completely broken: https://github.com/qtile/qtile/actions/runs/5275820671

@flacjacket that would be awesome if you could take a quick look :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,17 +9,13 @@ jobs:
   build-wheel:
     name: Build wheels
     runs-on: ubuntu-latest
-    container: quay.io/pypa/manylinux_2_24_x86_64
+    container: quay.io/pypa/manylinux_2_28_x86_64
     strategy:
       matrix:
         python-version: ["cp3.9", "cp3.10", "cp3.11", "pp3.9"]
     steps:
       - name: Install dependencies
-        run: |
-          apt update
-          apt-get install -y --no-install-recommends \
-            libcairo-dev \
-            libffi-dev
+        run: dnf install -y cairo-devel libffi-devel
       - uses: actions/checkout@v2
       - name: Set environment variables
         run: |
@@ -30,14 +26,13 @@ jobs:
           PYTHON_VERSION: ${{ matrix.python-version }}
       - name: Build wheels
         run: |
-          pip install xcffib
+          python -m pip install cffi
           python setup.py bdist_wheel
-          auditwheel repair --plat manylinux2014_x86_64 -w output_wheels dist/qtile-*.whl
       - name: Upload wheels
         uses: actions/upload-artifact@v2
         with:
           name: wheels-${{ matrix.python-version }}
-          path: output_wheels/*.whl
+          path: dist/*.whl
   test-wheel:
     name: Test wheels
     runs-on: ubuntu-latest
@@ -70,15 +65,12 @@ jobs:
   build-source:
     name: Build source
     runs-on: ubuntu-latest
-    container: quay.io/pypa/manylinux_2_24_x86_64
+    container: quay.io/pypa/manylinux_2_28_x86_64
     env:
       python-version: "cp3.11"
     steps:
       - name: Install dependencies
-        run: |
-          apt update
-          apt-get install -y --no-install-recommends \
-            libcairo-dev
+        run: dnf install -y cairo-devel libffi-devel
       - uses: actions/checkout@v2
       - name: Set environment variables
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = [ "setuptools>=41", "wheel", "setuptools-git-versioning<2", ]
+build-backend = "setuptools.build_meta"
+
 [tool.black]
 line-length = 98
 exclude = 'libqtile/_ffi.*\.py|libqtile/backend/x11/_ffi.*\.py|test/configs/syntaxerr.py'
@@ -8,3 +12,11 @@ testpaths = ["test"]
 filterwarnings = [
     "error:::libqtile",
 ]
+
+[tool.setuptools-git-versioning]
+enabled = true
+
+[project]
+name = "qtile"
+description = "A pure-Python tiling window manager."
+dynamic = ["version", "readme"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,13 @@ filterwarnings = [
 [tool.setuptools-git-versioning]
 enabled = true
 
+[tool.setuptools.dynamic]
+readme = {file = "README.md"}
+
 [project]
 name = "qtile"
 description = "A pure-Python tiling window manager."
-dynamic = ["version", "readme"]
+dynamic = ["version"]
+
+[project.scripts]
+qtile = "libqtile.scripts.main:main"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,36 +1,3 @@
-[metadata]
-url = https://qtile.org
-license = MIT
-long_description = file: README.rst
-long_description_content_type = text/x-rst
-author = Aldo Cortesi
-author_email = aldo@nullcube.com
-maintainer = Sean Vig
-maintainer_email = sean.v.775@gmail.com
-keywords =
-  qtile
-  tiling
-  window
-  manager
-classifiers =
-  Development Status :: 3 - Alpha
-  Intended Audience :: End Users/Desktop
-  License :: OSI Approved :: MIT License
-  Operating System :: POSIX :: BSD :: FreeBSD
-  Operating System :: POSIX :: Linux
-  Programming Language :: Python :: 3 :: Only
-  Programming Language :: Python :: 3.9
-  Programming Language :: Python :: 3.10
-  Programming Language :: Python :: 3.11
-  Programming Language :: Python :: Implementation :: CPython
-  Programming Language :: Python :: Implementation :: PyPy
-  Topic :: Desktop Environment :: Window Managers
-project_urls =
-  Documentation = https://docs.qtile.org/
-  Code = https://github.com/qtile/qtile/
-  Issue tracker = https://github.com/qtile/qtile/issues
-  Contributing = https://docs.qtile.org/en/latest/manual/contributing.html
-
 [options]
 packages = find_namespace:
 python_requires >= 3.9
@@ -48,10 +15,6 @@ tests_require =
   pep8-naming
   psutil
   coverage
-
-[options.entry_points]
-console_scripts =
-  qtile = libqtile.scripts.main:main
 
 [options.extras_require]
 doc =
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,6 @@
 [metadata]
-name = qtile
 url = https://qtile.org
 license = MIT
-description = A pure-Python tiling window manager.
 long_description = file: README.rst
 long_description_content_type = text/x-rst
 author = Aldo Cortesi


elParaguayo:
Should we be moving away from `setup.cfg` and using `pyproject.toml` as per PEP517? This would fix #4287 but it also looks like there's a simplified way of submitting packages to PyPI using [flit](https://flit.pypa.io/en/stable/).

ramnes:
Yes definitely, but what would that change as to how we build the wheels?

elParaguayo:
I have no idea how any of the packaging/uploading to PyPI works but, from the flit docs, it looks like building/publish is done differently to how we do things now so I assumed that might impact the automatic build process we run at the moment.

You can ignore me if I'm wrong!

ramnes:
Packaging and uploading to PyPI is usually very simple, but when you want to compile stuff and include it in the wheels you provide, it gets a bit harder, and this is mostly unrelated to setup.cfg or pyproject.toml.

My understanding is that we compile upfront so that people without a compiling toolchain or development headers can still install our package, but it means that we have to compile for different targets with such a CI step (the one I try to fix here).

Since we use CFFI extensively to communicate with C libraries, and we need to compile these CFFI interfaces. Unless we use their so called ["in-line mode"](https://cffi.readthedocs.io/en/latest/overview.html#simple-example-abi-level-in-line), which is what we've done for xcffib and cairocffi. But we didn't do it for the PulseAudio interface, so we still need that step before we release.

What I don't understand though is why we don't compile our Wayland libinput interface in this workflow and how that works for Wayland users. Do people actually use it? Because AFAICT it's not available in our wheels.

That being said, maybe we should just move to the in-line mode for PulseAudio as well. Or just ignore PA entirely and just let users figure it out (just like the libinput interface for Wayland?) In both cases, we could then remove that wheel compilation step entirely and make the release process much simpler.

@flacjacket @tych0 @m-col @krya: insights welcome since you're the ones who implemented our CFFI stuff

tych0:
I'm +1 on moving to in-line for PA (and wayland) as well. I don't think there's any reason to have both ways in our code base.

tych0:
Here's an attempt https://github.com/qtile/qtile/pull/4314

I have to run for now, but I'll poke at it more later unless someone gets to it first.

.github/workflows/release.yml
```diff
@@ -31,14 +26,13 @@ jobs:
           PYTHON_VERSION: ${{ matrix.python-version }}
       - name: Build wheels
         run: |
-          pip install xcffib
+          python -m pip install cffi
           python setup.py bdist_wheel
-          auditwheel repair --plat manylinux2014_x86_64 -w output_wheels dist/qtile-*.whl
```
flacjacket, line 36: The error you posted seems like it is saying the platform here is too old. At the time this was made, we were able to have it generate the `manylinux2014` wheels, but this may need to be bumped to `manylinux_2_28_x86_64`. IIRC, if an older version would work, the logs would say so at that point (and I think it would generate the new requested version and the older more compatible wheels). It may also be possible to drop the `--plat` argument entirely here.

flacjacket:
> What I don't understand though is why we don't compile our Wayland libinput interface in this workflow and how that works for Wayland users. Do people actually use it? Because AFAICT it's not available in our wheels.

This is very interesting. My guess would be that people using wayland are also using latest git or some packaged version from a distro that handles the pywayland setup correctly.

One of the challenges that would need to be tackled to change that over to in-line ABI as well would be transitioning pywayland and pywlroots over to in-line ABI as well. I remember we had problems when we picked up cairocffi from pycairo because that project used the ABI mode while I think at the time we had been using the API mode, and it was not possible to extend an ffi object of one type using one of the other type. I think we had asked them if they could change, but the argument there is that project also serves Windows users, which typically don't have a compiler available and so wouldn't be able to compile the API mode extension (I wonder how much the improvements since then with CFFI packaging, the addition of the out-of-line modes, and wheel generation would change the equation there today...). The arguments for keeping API mode are generally outlined in the [cffi docs](https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api) and I think give some strong arguments to keep API mode where possible. In addition to what is outlined there, one of the key things for pywayland/pywlroots is the ability to automatically check that structs are defined with the correct sets of members, and functions are correctly defined, as noted [here](https://cffi.readthedocs.io/en/latest/cdef.html#letting-the-c-compiler-fill-the-gaps). Especially with wlroots, version changes bring lots of API changes, and using the API mode makes all of the breakages apparent at build time.

flacjacket:
Just to check my understanding here, the key benefit to changing ABI mode ffi's over to the in-line mode is that we can see if optional dependencies are present at import time rather than needing to have them available at install time? It seems like from the cffi docs there would be some import time slow downs with in-line mode compared to out-of-line mode, but there aren't the same challenges that in-line API mode had pre-cffi 1.0. Thanks for figuring out that change could be made with ultimately nominal downside impacts while eliminating the painful challenges of installing all of the dependencies there.

tych0:
Whoops. Cross posting this here: https://github.com/qtile/qtile/pull/4314#issuecomment-1656910252

> tldr: I think we need a release plan where we don't use in-line mode for pulse.

tych0:
Ping, it looks like there's some test errors here (segfaults everywhere). Does anyone know why? I don't know much about how this is all supposed to work.

I propose we do a release ASAP, and then perhaps go about refactoring our packaging. Unless of course nobody understands this and it's easiest to rewrite it all :)

elParaguayo:
We definitely need to release asap. 

I have no idea how any of this packaging stuff works so I'm not sure how helpful I can be. If it's possible to release without it then that could be an option.

We do need to fix metadata in pyproject.toml so pip can read the version number. It thinks we're 0.0.0 at the moment.

tych0:
Ok. It looks to me like we're almost maintaining two separate builds: one in ci.yaml, and one in release.yaml that only gets run once a year. I wonder if we can switch to using ci.yaml's build somehow, and just doing a few extra release actions once the tests pass?

I'll admit I don't know anything about release.yaml, but I'm happy to look at doing something automatic here. Otherwise we could drop this for now and someone can release from their laptop.

ramnes:
Thank you for taking over @tych0! IIRC the PR was functional as is but I was interested in a review, from Sean in particular. We have it now so I guess we should just merge this and move forward anyway.

ramnes:
Also I think removing that build step isn't a good idea without more work on the PulseAudio side. I'm afraid we'd end up with people having install/build issues again.

Honestly if we can't make that PA widget inline, I think we should just rewrite it with calls to pactl rather than cffi, it would probably be much much simpler and work just as well.

tych0:
Yeah, though the tests didn't pass due to a segfault from your original run (I guess that's probably a bug that's been fixed). I'm gonna try to get the versioning working too before merging.

Which build step are you referring to here? I was thinking of just releasing PA as-is, since we pretty desperately need a release. People who don't use it will be able to consume qtile without all the xcffib/cairocffi confusion, which seems like a big win. People who do use it will have to do the same dance they always did, but overall it seems like a big win to just cut a release.

elParaguayo:
@ramnes - maybe rewriting the widget with something like [pulsectl-asyncio](https://github.com/mhthies/pulsectl-asyncio) could work.

However, in the interests of time, I agree with taking the quicker option and releasing as-is. I assume packagers will build with pulse support so it's users installing with pip who will still need to build as they did before.

@m-col @jwijenbergh any differing views?

tych0:
Any reason not to land this and try to release?

elParaguayo:
Yeah. We need to do this asap.

There are always bugs to fix (eg #4425) but we can always do bug bugfix release later.

tych0:
I agree, more releases is better. I'm going to land this and will try to tag one this weekend if nobody tags one before me.

elParaguayo:
I think this has introduced a bug.

qtile-extras (and its docs) builds against the current git version of qtile. The docs are now giving me this error:
```
WARNING: Discarding git+https://github.com/qtile/qtile.git@master#egg=qtile. Requested unknown from git+https://github.com/qtile/qtile.git@master#egg=qtile has inconsistent name: filename has 'qtile', but metadata has 'UNKNOWN'
```

elParaguayo:
Ignore that - I think the `egg` part is this issue here.

elParaguayo:
Still can't get qtile-extras tests to build successfully (failing on building wlroots ffi). Ideally, I'd like to have that working before we do qtile release as I do a release of qtile-extras at the same time.

ramnes:
Are you sure it's due to that PR? Isn't it just the same issue as #4287?

elParaguayo:
Tests always worked before. They've just started failing since this was merged.

elParaguayo:
Just checked: reverting these commits makes the wayland tests pass (i.e. ffibuild runs correctly).

elParaguayo:
<details>
<summary>Here's a longer error message from when I try to install this in a virtual environment</summary>
<blockquote>pip install --no-build-isolation "git+https://github.com/qtile/qtile.git"
Collecting git+https://github.com/qtile/qtile.git
  Cloning https://github.com/qtile/qtile.git to /tmp/pip-req-build-p5rnmj9h
  Running command git clone --filter=blob:none --quiet https://github.com/qtile/qtile.git /tmp/pip-req-build-p5rnmj9h
  Resolved https://github.com/qtile/qtile.git to commit 04e390e356dd76fc48d4065503a5af3a970ad976
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [42 lines of output]
      /home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/config/pyprojecttoml.py:108: _BetaConfiguration: Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.
        warnings.warn(msg, _BetaConfiguration)
      /home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/_distutils/dist.py:264: UserWarning: Unknown distribution option: 'use_scm_version'
        warnings.warn(msg)
      Built with support for XWayland helpers.
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 149, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/build_meta.py", line 377, in prepare_metadata_for_build_wheel
          self.run_setup()
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/build_meta.py", line 335, in run_setup
          exec(code, locals())
        File "<string>", line 98, in <module>
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/__init__.py", line 87, in setup
          return distutils.core.setup(**attrs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 147, in setup
          _setup_distribution = dist = klass(attrs)
                                       ^^^^^^^^^^^^
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/dist.py", line 476, in __init__
          _Distribution.__init__(
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 282, in __init__
          self.finalize_options()
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/dist.py", line 900, in finalize_options
          ep(self)
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/setuptools/dist.py", line 920, in _finalize_setup_keywords
          ep.load()(self, ep.name, value)
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/cffi/setuptools_ext.py", line 219, in cffi_modules
          add_cffi_module(dist, cffi_module)
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/cffi/setuptools_ext.py", line 49, in add_cffi_module
          execfile(build_file_name, mod_vars)
        File "/home/elp/Documents/scripts/piptest/venv/lib/python3.11/site-packages/cffi/setuptools_ext.py", line 25, in execfile
          exec(code, glob, glob)
        File "libqtile/backend/wayland/cffi/build.py", line 4, in <module>
          from libqtile.backend.wayland.cffi import cairo_buffer, libinput
      ModuleNotFoundError: No module named 'libqtile'
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
</blockquote>
</details>

So, it looks like the ffibuild script can't find the `libqtile` module.

If I remove "--no build isolation" there's no error but the cffi modules aren't built for wayland.

elParaguayo:
Well I'm really reaaly confused!

The wayland ffibuild is running ok with this PR. However, with this PR, I get these errors:
```
  E             File "/home/runner/work/qtile-extras/qtile-extras/.tox/py310/lib/python3.10/site-packages/libqtile/backend/wayland/core.py", line 1104, in setup_listener
  E               self.fd = lib.wl_event_loop_get_fd(self.event_loop._ptr)
  E           NameError: name 'lib' is not defined
```
Wayland's `core.py` has this:
```python
try:
    # Continue if ffi not built, so that docs can be built without wayland deps.
    from libqtile.backend.wayland._ffi import lib
except ModuleNotFoundError:
    pass
```
So the file loads but the name is undefined.

But the tests were ok when I reverted this PR but now I'm doubting myself!

elParaguayo:
Oh, FFS. I don't know any more. Reverting this PR is now giving me the same errors. So it's not this. Sorry for the noise.

However, installing from github via pip does show version number still being 0.0.0. Not sure if that's expected.

elParaguayo:
Right - I've hidden my old posts as they're too confusing. Here's a summary:

Prior to this PR being merged, qtile-extras's test suite ran this command to install qtile:
`pip install --no-build-isolation git+https://github.com/qtile/qtile.git`

This now fails with the following error:
```
        File "libqtile/backend/wayland/cffi/build.py", line 4, in <module>
          from libqtile.backend.wayland.cffi import cairo_buffer, libinput
      ModuleNotFoundError: No module named 'libqtile'
```

If you build current master without `--no-build-isolation`, the build succeeds but there is an issue with the Wayland ffi build as shown by the python shell:
```
Python 3.11.5 (main, Aug 28 2023, 20:02:58) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from libqtile.backend.wayland import core
>>> core.lib
AttributeError: module 'libqtile.backend.wayland.core' has no attribute 'lib'
```

If you revert this PR (as I've done in a branch):
`pip install --no-build-isolation "git+https://github.com/elParaguayo/qtile.git@revert-pyproject-toml"`
qtile builds successfully.

You can verify that the Wayland ffi build was successful in a python shell
```
Python 3.11.5 (main, Aug 28 2023, 20:02:58) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from libqtile.backend.wayland import core
>>> core.lib
<Lib object for 'libqtile.backend.wayland._ffi'>
```

So, it does feel like something broke with this PR.

elParaguayo:
Think I've got this working via https://github.com/qtile/qtile/pull/4437

Issue #4316, opened 2023-06-22
Can't set icon theme in Launchbar widget


elParaguayo:
### The issue:

Other widgets using `pyxdg` for icons have the ability to set the `theme_path`. This isn't available for `LaunchBar` but could be added.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

PR #4319, opened 2023-06-27
Added theme_path option and pyxdg error handling to LaunchBar

kritdass:
This solves #4316.

### DIFF
--- a/libqtile/widget/launchbar.py
+++ b/libqtile/widget/launchbar.py
@@ -45,7 +45,11 @@ from __future__ import annotations
 import os.path
 
 import cairocffi
-from xdg.IconTheme import getIconPath
+try:
+    from xdg.IconTheme import getIconPath
+    has_xdg = True
+except ImportError:
+    has_xdg = False
 
 from libqtile import bar
 from libqtile.images import Img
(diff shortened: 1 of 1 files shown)


libqtile/widget/launchbar.py
```diff
@@ -122,6 +127,11 @@ def __init__(
 
     def _configure(self, qtile, pbar):
         base._Widget._configure(self, qtile, pbar)
+
+        if not (has_xdg or self.text_only):
```
elParaguayo, line 131: You don't need `pyxdg` to have icons. Users can still use icons by passing a path as the first value. All `pyxdg` does is find the icon, it's not relevant to displaying it.

elParaguayo (changes requested):
Thanks for doing this.

In addition to the one specific comment, could you also update the widget's docstring to explain more about how/when `pyxdg` is used. i.e. it is used to lookup an icon path if the user has not provided an icon path as the first value in the list of `progs`.

libqtile/widget/launchbar.py
kritdass: Done! Let me know if anything else needs to be done.

kritdass:
Done! Let me know if anything else needs to be done.

elParaguayo:
Thanks. The only bit I don't like is the example icon path (and it's just me being fussy!).

It's good that it's an absolute path as that's what the widget expects however I'd rather just use something a bit more explanatory e.g. `/path/to/icon.png`.

There are actually some issues with the icon code that should also be improved:
- We should allow expansion of `~`
- We should allow other image formats beyond `png`.

However, we can keep those separate from this PR.

kritdass:
Understandable, I changed that path. I'll open a new PR for the other issues.

Issue #1946, opened 2020-10-30
Inconsistent margin sizes in columns layout


risingprismtv:
# Issue description

When in column layout, the outer margins are always what you set them to, in my case 5px. But the part where the margins touch each other ,such as down the center when the first two windows are open, or in between items on the stack is doubled, leading to inconsistent margin sizes around a single window. I have seen this behavior on two different machines on qtile 0.16.1 . If this is the intended behavior, could it be configurable so that margins would 'overlap' instead of doubling up at the intersections, leading to a single margin thickness around all windows?

# Qtile version
0.16.1

hash 04be6f6

# Stack traces
Nothing seems relevant. Happy to provide if needed

# Configuration
https://gist.github.com/risingprismtv/bef3e4d451a1775a7a71fd2703766b65

risingprismtv:
![qtile_columns_spacing](https://user-images.githubusercontent.com/65463190/97757346-e9a08e80-1ad2-11eb-99f5-b4519061d783.png)

m-col:
There is an "inner" gap around the windows that is created by the Columns layout, but the "outer" gaps around the inside edge of the screen are created using `Gap` objects. To make the gap between the columns (which is 2x the "inner" gap) equal to the gap between a column and the screen (1x "inner" gap beside the column and 1x "outer" gap beside the screen edge), these need to be the same value. For example, I have this behaviour using this:
```
layouts = [layout.Columns(margin=6)]
screens = [
    Screen(
        top=mybar,
        bottom=bar.Gap(6),
        left=bar.Gap(6),
        right=bar.Gap(6),
    )
]
```

tych0:
Gonna close this since it's answered.

grtcdr:
Hi guys!

I'm having a hard time configuring Qtile to have even margins outside and between windows using @m-col's solution.

1. It's still not even.
2. Setting gaps for every other edge besides the bar leaves gaps when using layout.Max, which I want to avoid.

m-col:
@grtcdr what are you trying, that results in it not being even?

If you want to change the size of the `Gap`s when you switch layout, try using the `layout_change` hook. E.g (untested):
```py
from libqtile import qtile, hook, bar

@hook.subscribe.layout_change
def _(lay, grp):
    if lay.name == "max":
        gap_size = 0
    else:
        gap_size = 6

    for gap in qtile.current_screen.gaps:
        if isinstance(gap, bar.Gap):
            gap.size = gap_size
    grp.layout_all()
[...]

PR #4326, opened 2023-07-04
fixed double margin problem in the columns layout

KonradGrande:
In order to avoid the issue of duplicated margins, I had to remove the option to set distinct margins for different sides of the window. I don't really see what the benefit of exposing this to the user would be anyway.

With this patch, the user will be able to set the margin to a number and have that be the gap between the windows and between the outermost windows and the edge of the screen. I think this is the behaviour the majority of the users expect the margin argument to have.

I also updated and added some tests for the columns layout.

### DIFF
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -301,13 +301,14 @@ class Columns(Layout):
     def configure(self, client: Window, screen_rect: ScreenRect) -> None:
         "Position client based on order and sizes"
 
-        # percentage_of_space: percent of a clients fair share of the available space
-        # available_space: the size of the space in pixels
+        # percentage_of_space: percentage of the available space when
+        #                      divided equally among the clients
+        # available_space: the size of the available space in pixels
         # clients_in_space: number of clients that are going to share the space
         def get_pixels_from_relative_size(percentage_of_space, available_space, clients_in_space):
             return round(0.01 * percentage_of_space * available_space / clients_in_space)
 
-        # percent of space occupied by the window horizontally
+        # percentage of the space occupied by the windows to the left of the client
         pos_x = 0
         for col in self.columns:
             if client in col:
@@ -317,7 +318,7 @@ class Columns(Layout):
             client.hide()
             return
 
-        # percent of space occupied by the window vertically
+        # percentage of the space occupied by the windows above the client
         pos_y = 0
         for c in col:
             if client == c:
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -146,11 +146,11 @@ class Columns(Layout):
         ("border_width", 2, "Border width."),
         ("single_border_width", None, "Border width for single window."),
         ("border_on_single", False, "Draw a border when there is one only window."),
-        ("margin", 0, "Margin of the layout (int or list of ints [N E S W])."),
+        ("margin", 0, "Margin of the layout."),
         (
             "margin_on_single",
             None,
-            "Margin when only one window. (int or list of ints [N E S W])",
+            "Margin when only one window.",
         ),
         ("split", True, "New columns presentation mode."),
         ("num_columns", 2, "Preferred number of columns."),
@@ -291,6 +291,16 @@ class Columns(Layout):
         return self.columns[self.current].cw
 
     def configure(self, client: Window, screen_rect: ScreenRect) -> None:
+        "Position client based on order and sizes"
+
+        # percentage_of_space: percent of a clients fair share of the available space
+        # available_space: the size of the space in pixels
+        # clients_in_space: number of clients that are going to share the space
+        def get_absolute_size_from_relative(
+            percentage_of_space, available_space, clients_in_space
+        ):
+            return round(0.01 * percentage_of_space * available_space / clients_in_space)
+
         pos = 0
         for col in self.columns:
             if client in col:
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for looking at this.

I'm curious why you had to remove the option for different margin sizes per side. I don't disagree that most users probably don't use that feature but I'm not keen on removing it unnecessarily, particularly when it is implemented in other layouts.

I would prefer a solution that maintained this functionality.

KonradGrande:
Yes, I understand. I don't like that it might break someones config either.

The double margins is not really a bug, it is just a consequence of the user
being able to set the margins.

There are two margins:
- actual_margin: the margin given as an argument to client.place() by the layout
- user_margin  : the margin supplied to the layout by the user

Currently, these are the same. When user_margin is a list, actual_margin is set
to user_margin, while it is set to [user_margin] * 4 when user_margin is an
integer.
They could be described as the number of pixels that are going to be taken away
from the sides of the window and turned into padding. This amount is
also added to x and y to recenter the window.

In my opinion, user_margin should be the gap you want between the windows and
from the outermost windows to the edge of the screen. I think that is what
most people want to achieve and what people who has not looked at the code expect
to happen when they set the margin.

Accepting user_margin as a list, does not work with this interpretation
of what user_margin should be. You would run into conflicts when windows are next
to each other. The closest you could get is to give the user the option to set one
margin horizontally and another margin vertically.

In this fix, the user is restricted to setting the margin to an integer. We can then
set actual_margin to [0, user_margin, user_margin, 0] and add a offset of user_margin
to x and y. That way you get a gap of user_margin pixels everywhere.

Initially, I tried to fix the problem of double margins between two windows by
dividing the margin by two on both windows where they meet. However, this makes
corner windows smaller than windows that are in between two others. The corner
windows would have to have full margins between itself and the edge of the
screen, whereas the windows that are between two other ones only need to have a
margin of 1/2 on each side. So the window in the middle would for instance be
(height - margin) tall, while the corner window would be (height - 3/2 * margin)
tall. The margins subtract from the window size, so for the windows to have the
same size with the same width and height, they also need to have the same amount
of margins.

elParaguayo:
I'd need to look at this more closely (it's not a layout that I use) but the issue feels similar to the one that #3750 is looking to solve and that one does preserve the ability to have different size margins. Is that something that could work here?

KonradGrande:
I think you guys have to decide if you want the margin parameter to be interpreted as a padding around the window or the gap between the windows. The current implementation sees the margin as a padding around the window. That is why you get a gap of margin * 2 between the windows right now. 

In #3750, he interprets the margin parameter to be the gap between the windows (same as my solution), but he accepts a list. The problem here, as I said earlier, is that you get conflicts. With two windows open and a margin of [0, 10, 30, 40], the window to the left wants a gap of 40 between itself and the rightmost window ,and the rightmost window wants a gap of 10. He resolves the conflict by picking the biggest one. The benefit here is that no configs are broken and there are no double gaps. However, unless you desperately want to have a different margin between the rightmost window and the right edge of the screen, and the leftmost window and the left edge of the screen, it is equivalent to being able to set a horizontal and a vertical gap. So, we could instead just transition to have a parameter called "gap" that can be set to a list of the form [x, y], where x is the horizontal gap and y is the vertical one. That makes it explicit that it is interpreted as the gap between the windows, fixes the problem and i'd argue that it allows for the same amount of flexibility. 

That is if setting a separate column horizontally and vertically is important. If that is not the case, this patch should be sufficient.

elParaguayo:
Thanks again for taking the time to set out your thoughts here.

Let me discuss with the other devs and we'll see if we can agree an approach.

m-col:
If you want equal margins everywhere, like this (2px):
```
┌──────────────┐
│              │
│  ┌──┐  ┌──┐  │
│  │  │  │  │  │
│  │  │  │  │  │
│  └──┘  └──┘  │
│              │
└──────────────┘
```
this would be achieved (currently) with 1px margins on the screen and 1px margins on the layout.

Is your suggestion that this should be possible with margins only on the layout?

What would the expected window positioning be if the layout is configured with a margin of 1px (or indeed any odd number)?

KonradGrande:
Yeah, I saw your fix using screens here #1946, but yes i'd like for it to be possible by just configuring the layout. Especially since people seem want to have different margins for different layouts. I'd also like to have a way to change the margins(including the outside margins) using keybindings as described here #2375. Making the layout responsible for the margin would simplify this too. 

The simplest way to deal with odd margins is probably to just do a +1. The margin would be divided by 2 where windows meet, so even numbers are preferable.

KonradGrande:
What do you guys think?

jwijenbergh:
Any way we can have only double margins when a list is used? I think completely removing the list option is a problem as I guess a fair amount of people use that

KonradGrande:
Yes, that could definitely be done.

KonradGrande:
This should do the trick.
```Diff
diff --git a/libqtile/layout/columns.py b/libqtile/layout/columns.py
index 4005a128..36cd79f6 100644
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -276,12 +276,17 @@ class Columns(Layout):
             if self.margin_on_single is not None:
                 margin = self.margin_on_single
 
-        # fix double margins by just having margins on the east and south side of each window
-        x_offset = screen_rect.x + margin
-        y_offset = screen_rect.y + margin
-        usable_width = screen_rect.width - margin
-        usable_height = screen_rect.height - margin
-        margin = [0, margin, margin, 0]  # [N, E, S, W]
+        x_offset = screen_rect.x
+        y_offset = screen_rect.y
+        usable_width = screen_rect.width
+        usable_height = screen_rect.height
+        if isinstance(margin, int):
+            # fix double margins by just having margins on the east and south side of each window
+            x_offset += margin
+            y_offset += margin
+            usable_width -= margin
+            usable_height -= margin
+            margin = [0, margin, margin, 0]  # [N, E, S, W]
 
         width = get_absolute_size_from_relative(col.width, usable_width, len(self.columns))
         x = x_offset + get_absolute_size_from_relative(pos, usable_width, len(self.columns))

```

Should I add it?

ramnes:
Currently running the branch and I love the result, thanks!

What's the problem with `math.ceil(margin / 2)` pixels around all clients and then the other half all around the screen? I read your comments but I really don't understand the maths here, please ELI5

KonradGrande:
I'm glad you like it @ramnes! The approach you suggest would also work. The benefit in doing it the way it is implemented, is that odd number are also allowed. That is not the case with this approach, as `2 * math.ceil(margin / 2) = margin + 1` for all odd natural numbers.  

In my approach, if the user has set the margin to 5 for instance, each window is made 5 pixels shorter and 5 pixels slimmer. This gives you a gap between all the windows, but the gap between the edge of the screen and the windows is not even; there is no gap between the windows and the edge of the screen on the northern or western side of the screen, and there is twice as much of a gap on the southern and eastern side as there is supposed to be. So, to center the windows on the screen, making the gaps around the edges of the screen even, I also shift the windows 5 pixels south and 5 pixels east. That is basically it. Apart from that you just have to calculate the size of the windows with the offset in mind, as it does affect how much of the screen you can paint windows on. I'm not sure if this explanation would be suitable for a 5-year-old, but it's the best I can do

I have now separated this from the margin parameter by adding a new parameter called 'gap'. That way people can choose to use this approach to spacing if they want it, and we don't break any configs. I think mixing it could be confusing as it is not the same thing conceptually. If you set the margin to 10, 10 pixels will be removed from all sides of each window. If you set the gap to 10, you will get 10 pixels between the windows and 10 pixels between the window and the edge of the screen.

Issue #1714, opened 2020-05-09
Tasklist icon placement 


am01k01h3:
I can't seem to get the icons & text placed in the center vertically in the tasklist.

If the text is placed vertically center (by setting padding = 8), then the icon is chopped off like in the picture. 
![Screenshot from 2020-05-08 22-42-28_2](https://user-images.githubusercontent.com/62161236/81462010-a85f9f80-917d-11ea-86ca-cb2b6845884c.png)

if I adjust padding to 0, so that the icon is centerd, the text goes all the way up.

![Screenshot from 2020-05-08 22-49-50](https://user-images.githubusercontent.com/62161236/81462152-68e58300-917e-11ea-8d35-fe96a8c5741d.png)

Is this a bug?

qtile version 0.15.0
[config.py.txt](https://github.com/qtile/qtile/files/4602520/config.py.txt)

am01k01h3:
Some observations:
I'm really just trying to make text and icon vertically centered in the bar and it seems like this is the default behavior for other widgets. But for tasklist, icon and text are vertically aligned to top edge.

With padding=0, the rectangle box around text becomes small/tight and vertically aligned to top.
Icon placement works correctly with padding=0.

With padding=8, the text becomes centered correctly. But it adds padding to the icons on all directions, and pushed it down (especially if I set the icons to a large size over 24 pixels).

Groupbox or Graph widgets seem to not have this issue, text or graphs are always vertically aligned to center, unless overridden with margin_y.

M05QU170:
did you try to change icon size in widget.tasklist section? And try to play with font size if this will not workout

am01k01h3:
> did you try to change icon size in widget.tasklist section? And try to play with font size if this will not workout

Yes, my qtile config is posted above.
My current workaround is to use font size 10 which is same as all other widgets, icon size 20, padding 8 and everything is sort of centered correctly. The issue I've raised is that if we use icon size larger than 24 it starts to show the behaviour I mentioend above.

DeeeeLAN:
This is still a valid issue, the tasklist top-aligns the icon and text which is not very flexible. I think either center-align both, or have separate padding values for the icon and the text would work fine.

ezntek:
the issue still exists.

this is what it looks like with the following:
```py
# snip
margin_x=12,
margin_y=8,
padding=3,
icon_size=16,
```

[...]

PR #4008, opened 2022-11-18
Two fixes for CurrentLayoutIcon

elParaguayo:
Two fixes for the widget:
- Ensure `os.path.expanduser` is called on `custom_paths`
- Include svg support

Closes #4003
Closes #4004

PR #4329, opened 2023-07-08
fix TaskList icons' vertical centering issue

ezntek:
One line fix that tries to address #1714 and #3873, as the fixes still do not work on my local system running on the main dev branch ([described here](https://github.com/qtile/qtile/issues/1714#issuecomment-1627198379)).

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -531,7 +531,7 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             return
 
         x = offset + self.borderwidth + self.padding_x
-        y = (self.height - self.icon_size) // 2
+        y = ((self.height // 2) - (self.icon_size // 2)) // 2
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for this.

I'm not convinced your solution is the correct one, however. If the icon's size was equal to `icon_size` then the existing code should work so there must be something going wrong elsewhere.

I suspect the issue is with the icon resizing code. Let me take a look.

elParaguayo:
Yeah - I removed similar code in `CurrentLayoutIcon` to fix a similar issue (#4008) so I would think the same should be done here.

So, you could either change this or close it and I'll do it.

ezntek:
I guess I'll leave it up to you, then.

ezntek:
not to rush you or anything, but any progress @elParaguayo?

elParaguayo:
@ezntek See #4349

PR #4336, opened 2023-07-10
Added active_label parameter to GroupBox

kritdass:
If you want me to do additional changes to GroupBox, I can help.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -238,6 +238,11 @@ class GroupBox(_GroupBase):
         ),
         ("spacing", None, "Spacing between groups" "(if set to None, will be equal to margin_x)"),
         ("toggle", True, "Enable toggling of group when clicking on same group name"),
+        (
+            "active_label",
+            None,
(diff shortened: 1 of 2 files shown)


elParaguayo:
Is this needed? Groups have a `label` attribute already.

kritdass:
It is specifically for the current group and will provide another way for highlighting it. I think `focused_label` would be a better name after looking at some of the wording.

kritdass:
Actually, I suppose that `lazy.group.set_label` can be used in a keybind. Do you think this is worth it to add for ease?

elParaguayo:
I don't like the idea of getting a group label in the widget. You could use a hook to change the group label when focus changes.

kritdass:
It doesn't change the group label itself, just what is displayed in the GroupBox if that is what you're thinking. I do agree this is possible otherwise, should I close this PR?

elParaguayo:
I don't think it makes sense for the `Group` to set the `label` but have the widget set an `active_label` value. Also, having it in the widget means it's the same for every group. I'm inclined to close this.

kritdass:
I'll close it, thanks for clarifying!

PR #4343, opened 2023-07-15
Expose next/prev_group commands of GroupBox

superpupervlad:
I found these commands very useful if you want to move only between active groups with opened windows.

Also why don't just _somehow_ expose every function of every object  but with some `force` flag only. This will allow users to  use any command without specifically expose it but only if they know what they are doing.

### DIFF
--- a/libqtile/widget/groupbox.py
+++ b/libqtile/widget/groupbox.py
@@ -35,6 +35,7 @@ from functools import partial
 from typing import Any
 
 from libqtile import hook
+from libqtile.command.base import expose_command
 from libqtile.widget import base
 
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks.  I don't have any real objections to this.  There are exposed commands already for moving to the next group but the groupbox ones have the advantage of limiting this to the groups shown by the widget. 

As for exposing everything: the exposed commands represent the "stable" API (ie we'll try not to change that too much). Users do have full access to all the other commands though. If you do `lazy.widget["groupbox"].function(myfunc)`, you can then define a function called `myfunc`which should take one argument which will be the widget instance.  You then have access to all the widget's internals.

superpupervlad:
Oh, don't know about `lazy.widget["groupbox"].function(myfunc)`. This really solve the problem.

Well, I guess then you can close PR or if this functions won't change then why not merge.

elParaguayo:
OK. I'll close as I don't think this really needs to be exposed.

PR #4344, opened 2023-07-17
X11 Core: Bring back motion notify cap that is off by default

jwijenbergh:
Previously we introduced this but Nvidia users complained that floating resize/move was still laggy. Now we handle MOTION NOTIFY events later. In this commit we bring back the cap for users that want to have a cap set for resource reasons or that they simply see that the "handling later" fix is not good enough.

It is off by default. I also added a comment in the default config.

Reported by @purplebar0 at https://github.com/qtile/qtile/pull/4141#issuecomment-1636754538


elParaguayo:
LGTM

PR #4345, opened 2023-07-17
Default config: Add toggle floating, fullscreen keybinds

jwijenbergh:
title 

I missed these


elParaguayo:
Nice. These are frequently requested.

LGTM

PR #4348, opened 2023-07-20
KeyboardLayout: Remove unnecessary logging

### LINKED ISSUE #4342: Logs pollution - keyboardlayout.py
### The issue:

Bunch of these:
```
2023-07-15 15:28:06,304 ERROR libqtile keyboardlayout.py:set_keyboard():L102 Can not load ~/.Xmodmap:
```
My humble opinion is that if I don't use something optional, I shouldn't have an error about it in the logs. Oh and it's supposed to be **"Cannot"**. I don't have another idea on how this should be done atm.

### Required:

elParaguayo:
The widget will log an error if the user does not have a .Xmodmap file. As this is optional for users, we should check the file exists before trying to load it.

Fixes #4342

### DIFF
--- a/libqtile/widget/keyboardlayout.py
+++ b/libqtile/widget/keyboardlayout.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import re
 from abc import ABCMeta, abstractmethod
+from pathlib import Path
 from subprocess import CalledProcessError, check_output
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


ervinpopescu:
This does its job. Thanks, @elParaguayo, will try to give this kind of "easy" issues a try in the future, but I'm currently overwhelmed with work.

PR #4349, opened 2023-07-21
TaskList: Use qtile Img object for all icons

elParaguayo:
Doing this makes drawing/scaling the icons easier.

Also fixes issue with vertical positioning of the icon and the text.

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -24,6 +24,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import re
+from io import BytesIO
 
 import cairocffi
 
@@ -418,10 +419,12 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
             framecolor = bordercolor
 
         framed = self.layout.framed(border_width, framecolor, padding_x, self.padding_y)
+
+        y_pos = max(self.margin_y, (self.bar.height - self.layout.height) // 2)
         if block and bordercolor is not None:
-            framed.draw_fill(offset, self.margin_y, rounded)
+            framed.draw_fill(offset, y_pos, rounded)
         else:
-            framed.draw(offset, self.margin_y, rounded)
+            framed.draw(offset, y_pos, rounded)
 
         if icon:
             self.draw_icon(icon, offset)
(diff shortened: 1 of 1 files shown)


ezntek:
just tested it, its a bit _wonky_.

Screenshot_2023-07-22_13-42-50
Here's with `margin_y=0`,

Screenshot_2023-07-22_13-44-35
But here's with `margin_y=10`.

The icon's size magically changes when the margin_y does, oh and the highlight area (border in qtile jargon iirc) is now off-centered and is not affected by margin x or y. 

Screenshot_2023-07-22_13-45-44

i think this may need further fixes.

elParaguayo:
Thanks for testing. Will investigate further!

elParaguayo:
> just tested it, its a bit _wonky_.
> 
> Screenshot_2023-07-22_13-42-50 Here's with `margin_y=0`,
> 
> Screenshot_2023-07-22_13-44-35 But here's with `margin_y=10`.
> 
> The icon's size magically changes when the margin_y does

If you don't set the `icon_size` then the widget calculates the icon size required to match the bar height less any border and margin so changing `margin_y` should change the icon size in your case.

> oh and the highlight area (border in qtile jargon iirc) is now off-centered and is not affected by margin x or y.
> 
> Screenshot_2023-07-22_13-45-44
> 
> i think this may need further fixes.
> 
> Relevant config:
> 
> ```python
> widget.TaskList(
>             highlight_method="block",
>             border="#313244",
>             urgent_border="f38ba8",
>             margin_x=12,
>             margin_y=8,
>             padding=3,
>             #theme_path="/usr/share/icons/Papirus-dark",
>             txt_floating="🗗",
>             txt_maximized="🗖",
>             txt_minimized="🗕", 
>             **DECORATIONS_LR
>         ),
> ```

Can you post the `DECORATIONS_LR` definition please.

ezntek:
Here's the decorations constant (using qtile-extras btw)

```py
DECORATIONS_LR = {
    "decorations": [
        RectDecoration(colour="#11111e", radius=16, filled=True, padding=3),
        #RectDecoration(line_width=3, colour="#cba6f7", padding=3, radius=17),
     ],
    "foreground": "#cdd6f4",
}
```

elParaguayo:
In my opinion, text and icons should be centered vertically by default. `margin_y` should be used for fine tuning up or down.

elParaguayo:
@ezntek - sorry, forgot about this one. Based on the explanations above, is this one ok to merge or is it still "wonky"?

ezntek:
It's okay, I also forgot about this one. I don't use qtile anymore, and I have no way to see if this still works so I think it is okay to leave it be.

when someone else reports this issue then they can try this fix.

Issue #4353, opened 2023-07-23
Built-in Clock Widget doesnt respect the LC_TIME variable


parham-dev27:
### The issue:

The LC_TIME variable which defines the time locale, is not used by the clock widget. So if for example $LC_TIME="de_DE.UTF-8", but the widget still shows the month/day names in english. 

I am able to fix this, if you want let me know. I can add a pr for it. Thank you

### Required:

- [x] I have searched past issues to see if this bug has already been reported.

elParaguayo:
Shouldn't the locale be systemwide? You can just add these lines to your config:
```python
import locale
locale.setlocale(locale.LC_TIME, "de_DE.UTF-8")
```
And that should work.

I'm surprised python's `datetime` module isn't picking it up be default.

parham-dev27:
Thank you. I also created a pull request fixing this.

elParaguayo:
Did some digging. This is from the python docs:
```
Initially, when a program is started, the locale is the C locale, no matter what the user’s preferred locale is. There is one exception: the [LC_CTYPE](https://docs.python.org/3/library/locale.html#locale.LC_CTYPE) category is changed at startup to set the current locale encoding to the user’s preferred locale encoding. The program must explicitly say that it wants the user’s preferred locale settings for other categories by calling setlocale(LC_ALL, '').
```
So, this is why `LC_TIME` is ignored and we do need to set it.

I'm still on the fence as to whether this needs to be in the clock widget or whether users should just set it in their configs.

parham-dev27:
Well, you are right that users could add it to their configs, but the average user (at least for me) expects it to work out of the box. So if it doesn't work, they won't do too much digging. Alternatively, we could add this to the configuration options.

elParaguayo:
I agree with your point about the average user and so it feels like it should be enabled by default. Let me look at the PR later.

parham-dev27:
Fixed by #4356

PR #4354, opened 2023-07-23
Fix Clock Widget Respecting LC_TIME

parham-dev27:
Check Issue #4353 and fixed a little typo

### DIFF
--- a/libqtile/widget/clock.py
+++ b/libqtile/widget/clock.py
@@ -24,6 +24,10 @@
 import sys
 import time
 from datetime import datetime, timedelta, timezone
+from os import getenv
+
+from locale import setlocale, LC_TIME
+from locale import Error as LocaleError
 
 from libqtile.log_utils import logger
 from libqtile.widget import base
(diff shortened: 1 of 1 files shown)


elParaguayo:
We actually set the locale during `qtile start`: 
https://github.com/qtile/qtile/blob/142dc80aae835afc1767d327560311f3ba648a58/libqtile/scripts/start.py#L83

Should we add `LC_TIME` here?

parham-dev27:
The problem is `locale.getdefaultlocale()` returns the LANG variable from /etc/locale.conf. So the problem occurs here. 
` locale.setlocale(locale.LC_ALL, locale.getdefaultlocale())` would also fix it, if the LC_TIME variable is returned and not the LANG variable. 

For example if someone lives in Germany and has LC_TIME set to de_DE but has the system language (LANG) set to en_US, the problems occur because everything is going to be set to en_US. 

The only way to fix this is by getting the environment variable using  `os.getenv("LC_TIME")` and only setting the LC_TIME to the returned value.

elParaguayo:
What I meant was adding the line to set LC_TIME in the same place,  rather than in the widget.

parham-dev27:
My bad for not understanding what you meant. We can do that. 

Additionally we can set all other locals that the python locale module supports so no other problems occurs because of the wrong locale.

Supported locale categories by the python-locale module: 
LC_ALL
LC_COLLATE
LC_CTYPE
LC_MESSAGES
LC_MONETARY
LC_NUMERIC
LC_TIME 


P.S. Working on a PR, if you have any suggestions let me know :)

parham-dev27:
@elParaguayo Closing in favor of #4356

PR #4356, opened 2023-07-24
Fix Locale Setting

parham-dev27:
Check Issue #4353 and PR #4354

Note: Used try/except every time, so if one fails, the other can continue.

### DIFF
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -80,7 +80,7 @@ def make_qtile(options):
 
 def start(options):
     try:
-        locale.setlocale(locale.LC_ALL, '')
+        locale.setlocale(locale.LC_ALL, "")
     except locale.Error:
         pass
 
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -83,7 +83,30 @@ def start(options):
         locale.setlocale(locale.LC_ALL, locale.getdefaultlocale())
     except locale.Error:
         pass
-
+    try: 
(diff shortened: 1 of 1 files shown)


parham-dev27:
@elParaguayo will you please check this out :) thank you

ramnes (approved):
LGTM

elParaguayo:
Do we need all of this?

Looking at python's docs, they say:

> Applications typically start with a call of

``` python
import locale
locale.setlocale(locale.LC_ALL, '')
```
> This sets the locale for all categories to the user’s default setting (typically specified in the LANG environment variable). If the locale is not changed thereafter, using multithreading should not cause problems.

Would it be sufficient to change the existing `setlocale` line with this?

parham-dev27:
I read the docs for the locale module and did some testing. You can achieve the same with what the docs recommend. So I changed the code. Thank you for your suggestion!

elParaguayo:
I'll run the tests but I expect them to fail as there seems to be an unrelated issue at the moment which we may need to look into.

elParaguayo:
Unrelated to your PR. If you rebase onto latest master they should pass OK.

parham-dev27:
still failed

elParaguayo:
You need to run `black` on the code to format it.  I'm pretty sure it will change `''` to `""`.

parham-dev27:
yeah I read why the tests were failing and saw that is was because of PEP8 Standards. I went to the docs and figured out that was the problem. Thanks now the tests pass

PR #4360, opened 2023-07-29
Pin pywayland

elParaguayo:
Trying some fixes to see if we can get CI passing:
- Test 1: pin `pywayland` to 0.4.15


jwijenbergh:
LGTM

PR #4364, opened 2023-07-31
Update CI to tox 4

elParaguayo:
tox-gh-actions now supports tox 4.

Also fix `passenv` error with whitespace.

Let's see if this passes.


elParaguayo:
This needs investigating further.

Despite saying that tox 4 is supported, the tests are using the wrong python version (e.g. pypy 3.9 is running `py310`). This could be my fault...

We're getting this error:
```
failed with /home/runner/work/qtile/qtile/scripts/ffibuild (resolves to /home/runner/work/qtile/qtile/scripts/ffibuild)
is not allowed, use allowlist_externals to allow it
```

I've tried various iterations of `allowlist_externals` and they all failed. Again, this one's probably my fault!

elParaguayo:
Do not be fooled - these tests have not passed successfully!

elParaguayo:
OK - this is working now. It needs some tidying up and then we should merge.

PR #4368, opened 2023-07-31
Config: Set group after configuring bar

jwijenbergh:
When we do not do this then the layout.show() method gets a screen_rect that does not account for the bar borders. So we first configure the gaps (bars) such that the screen gets the correct screen_rect when used in set_group that further calls group.set_screen, which calls layout.show()

You can observe this issue with the treetab layout. Start qtile with a bar that has borders and have the treetab be the first layout to be shown. You can see that the heading is not set correctly, when you spawn a window or switch to and back to the group it fixes itself.

Got on startup:
gotonstartup

Want on startup (this is after I switched groups and back):
wantonstartup

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -469,9 +469,9 @@ class Screen(CommandObject):
         self.width = width
         self.height = height
 
-        self.set_group(group)
         for i in self.gaps:
             i._configure(qtile, self, reconfigure=reconfigure_gaps)
+        self.set_group(group)
         if self.wallpaper:
             self.wallpaper = os.path.expanduser(self.wallpaper)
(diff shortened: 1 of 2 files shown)


elParaguayo:
Why does this PR impact that test?

jwijenbergh:
> Why does this PR impact that test?

Discussed on discord but it's the fact that the focus_change hook is now called after the bar is initialized. Checking on empty string is IMO wanted behaviour for this widget because it's hacky that there are multiple variants of the text when there are no windows (space initially and empty string after hooks have been called)

elParaguayo (approved):
LGTM

PR #4370, opened 2023-08-01
Added void and VoidLinux in the distribution list for the CheckUpdate…

ebiscaia:
- Added Void Linux as distribution for Check Update widget.

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -39,6 +39,8 @@ CMD_DICT = {
     "Fedora": ("dnf list updates -q", 1),
     "FreeBSD": ("pkg_version -I -l '<'", 0),
     "Mandriva": ("urpmq --auto-select", 0),
+    "void": ("xbps-install -nuMS", 0),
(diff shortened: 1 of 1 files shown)


libqtile/widget/check_updates.py
```diff
@@ -39,6 +39,8 @@
     "Fedora": ("dnf list updates -q", 1),
     "FreeBSD": ("pkg_version -I -l '<'", 0),
     "Mandriva": ("urpmq --auto-select", 0),
+    "void": ("xbps-install -nuMS", 0),
+    "VoidLinux": ("xbps-install -nuMS", 0),
```
elParaguayo, line 43: Why do we need 2 entries here? The command seems to be the same.

ebiscaia: It was just because the people could get the values from different piping commands. But I will leave just one.

jwijenbergh: Can we just call it "Void"?

ebiscaia: I think so. I could set an environment variable in case I change distributions in the future.

PR #4373, opened 2023-08-02
Fix type annotations

fabianfranzelin:
Hi guys,

I am having an issue with recent master with regards to erroneous type annotations in various files. 

The changes in this PR fixed the problem for me and qtile is back to running like a charm :)

### DIFF
--- a/libqtile/extension/base.py
+++ b/libqtile/extension/base.py
@@ -21,7 +21,7 @@
 import re
 import shlex
 from subprocess import PIPE, Popen
-from typing import Any
+from typing import Any, List, Tuple
 
 from libqtile import configurable
 from libqtile.log_utils import logger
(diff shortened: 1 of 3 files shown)


elParaguayo:
What python version are you running? `List` and `Tuple` were deprecated in python 3.9.

fabianfranzelin:
> What python version are you running? `List` and `Tuple` were deprecated in python 3.9.

I am running 3.8.10. I was not aware of this change with 3.9. I will update the version and check again. Thanks.

elParaguayo:
Ah, that makes sense then.

FYI qtile aims to support the last 3 versions of python (i.e. 3.9-11). Our test suite includes type checking so I'd be confident in saying that you won't have any issues with annotations after you've upgraded.

I'll close this too as the changes are not required.

fabianfranzelin:
> Ah, that makes sense then.
> 
> FYI qtile aims to support the last 3 versions of python (i.e. 3.9-11). Our test suite includes type checking so I'd be confident in saying that you won't have any issues with annotations after you've upgraded.
> 
> I'll close this too as the changes are not required.

It looks good. Thanks for your help.

PR #4357, opened 2023-07-24
[x11] Fix two stacking issues

elParaguayo:
1) window.bring_to_front ignored stacking rules. This has been amended so that stacking rules are respected.
2) Static windows were ignored by stacking rules. This has also been amended to ensure that these windows are included.

Fixes #4355

@ervinpopescu - would be great if you could test this.

PR #4375, opened 2023-08-03
 X11 Window: Return true for focus success after sending TAKE_FOCUS protocol event

### LINKED ISSUE #4372: Key input going to wrong window on recent focus change
### The issue:

I have this issue, it started approximately 8-12 months ago but I was unable to figure out reliable steps to reproduce, apologies for the following instructions:

 -   Set follow_mouse_focus = False, aside from that the config I am using is entirely default
 -   Install BurpSuite Community Edition, https://portswigger.net/burp/communitydownload
 -   Open BurpSuite (it is in path pointing to /usr/local/bin/BurpSuiteCommunity at the top of the display there is a tab called 'Repeater', within the main panel, click the 'X' to show the request/response input fields
 -   Open Firefox
 -   Open a third window (this is my current workaround)
 -   In firefox, I click the address bar, type 'aaaa'
 -   In BurpSuite, I click the 'request' area and type 'bbbb'
 -   In firefox, I cilck the address bar and type 'cccc' - the 'cccc' is inputted into burpsuite, despite firefox having focus
-    If I reproduce the same steps, but after typing 'bbbb', I click the terminal, and then click the firefox address bar, it works fine. Focus seems to work fine with the terminal (e.g., if instead of Firefox I reproduce with the terminal). I am able to reproduce if I use a HTML text input in firefox (it is not specific to the the address bar).

BurpSuite is (from my understanding) a Java/SWING application. I have no idea if this is relevant but I have not seen the issue manifest elsewhere.

Tested Ubuntu 23.04 using Qtile git master (and previous versions) using X11. As I said above, this issue appears to have been introduced in the last 8-12 months (maybe a bit earlier, idk), I'm unsure if this is a Qtile bug or an underlying lib, but either way the issue does not seem to be widespread to other Window Managers.

focus-bug

### Required:

jwijenbergh:
Fixes #4372 

First commit was cherry picked from #4357 and will be removed once that is merged

See for some hilarious X11 rant about this https://github.com/sminez/penrose/issues/120#issuecomment-766362138

Logic is from i3: https://github.com/i3/i3/blob/next/src/manage.c#L642-L644

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1243,6 +1243,7 @@ class _Window:
             )
 
             self.window.send_event(e)
+            return True
 
         # we didn't focus this time. but now the window knows if it wants
         # focus, it should SetFocus() itself; we'll get another notification


m-col:
Good stuff, lgtm

PR #4377, opened 2023-08-04
Move base backend into folder

m-col:
This re-organises the base backend a bit to more closely align with the implementations.

Also avoids disruption due to changes in import paths due to the recent Drawer refector


elParaguayo:
Looks great.

elParaguayo (approved):
This is a nice improvement to tidying up the code.

I'll let you figure out the mypy failures ;)

ramnes:
Nice

PR #4381, opened 2023-08-06
[x11] Revert `Drawer.clear` method

elParaguayo:
There is an apparent issue whereby a `RecordingSurface` cannot clear an `XCBSurface`. This leads to issue with transparency where artefacts are rendered in incorrect locations.

Using `CLEAR` and `SOURCE` operators in `Drawer._paint` are also insufficient to fix the issue.

This PR addresses this by ensuring that `clear` is drawn directly to the `XCBSurface`.

This should be looked at in the future as this feels like an unnecessary piece of code!


m-col (approved):
Works well :)

PR #4388, opened 2023-08-09
Add Suspend hook

elParaguayo:
Draft for now as there's a slight issue...

The way this code works is that, if a `suspend` hook is set then we create an inhibitor lock which delays the suspend operation. We wrap the hooked function in a wrapper that should release the lock once the code has run.

According to the [spec](https://www.freedesktop.org/wiki/Software/systemd/inhibit/), calling `Inhibit` returns a file descriptor which should be closed to release the lock. That's not working (you can verify with `watch -n 1 -- systemd-inhibit --list` and see that the inhibitor is not removed when suspending). I'm not sure what I'm doing wrong!

The pseudocode on that spec page makes it look very straightforward.

I tried looking at some other implementations:
- [upower](https://gitlab.freedesktop.org/upower/upower/-/blob/master/src/linux/up-backend.c#L670) - just closes the descriptor that's returned
- [swayidle](https://github.com/swaywm/swayidle/blob/master/main.c#L218) - they duplicate the descriptor. I tried that with `os.dup`, didn't work.

@m-col @jwijenbergh would love a second pair of eyes on this when you get a chance.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -61,6 +61,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
           - `lazy.window.move_up()/move_down()` moves windows up and down the z axis.
           - added `only_focused` setting to Max layout, allowing to draw multiple clients on top of each other when
             set to False
+        - Add `suspend` hook to run functions before system goes to sleep.
     * bugfixes
         - Fix bug where Window.center() centers window on the wrong screen when using multiple monitors.
         - Fix `Notify` bug when apps close notifications.
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -51,15 +51,10 @@ from libqtile.dgroups import DGroups
 from libqtile.extension.base import _Extension
 from libqtile.group import _Group
 from libqtile.log_utils import logger
+from libqtile.resources.sleep import inhibitor
 from libqtile.scratchpad import ScratchPad
 from libqtile.scripts.main import VERSION
-from libqtile.utils import (
-    cancel_tasks,
-    get_cache_dir,
-    lget,
-    send_notification,
-    subscribe_for_resume_events,
-)
+from libqtile.utils import cancel_tasks, get_cache_dir, lget, send_notification
 from libqtile.widget.base import _Widget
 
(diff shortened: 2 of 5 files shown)


elParaguayo:
It's very weird - the Inhibitor code works when run outside of qtile...

elParaguayo:
Can't see the error. I've confirmed the file descriptor is created and successfully closed but, for some reason, after closing the descriptor, the inhibitor remains.

If I run the same code outside of qtile then it works correctly so it seems to be some interaction with our event loop.

tych0:
Have you/the tests/anyone made a copy of qtile's process (in particular, via our spawn code) with this fd in the fd table, and does it have O_CLOEXEC? I see:

https://github.com/qtile/qtile/blob/f46450eef6eb2fab5e6a1037fe46893a8c8cc0ce/libqtile/core/manager.py#L1246-L1250

that we have a bug where we only close the stdfds, and not *all* open fds, before execing. This would leave anything without O_CLOEXEC in the spawned task's fd table, and as long as that task doesn't exit (which it wouldn't if it was e.g. a terminal or something), it'll keep a copy of the fd pinned.

Maybe we should try to close everything there? But perhaps a simpler solution would be to just to set CLOEXEC on your systemd inhibit fd.

elParaguayo:
@tych0 Thank you!

I don't know much (well, virtually nothing, really) about file descriptors but I did see that `swayidle` set `CLOEXEC`. I won't pretend that I understand what it does but I thought I read somewhere that this was set automatically by python. Clearly that wasn't the case as adding
```python
fcntl.fcntl(self.fd, fcntl.F_SETFD, fcntl.FD_CLOEXEC)
```
seems to make it work perfectly.

tych0:
I think it's probably only set on fds opened/etc. by the python runtime itself. I haven't read the dbus-next implementation, but I'd guess maybe it comes from some C library somewhere? Anyway, glad it works.

tych0 (approved):

elParaguayo:
Yes, that could well be right.

There are a couple more things I need to do on this one (checking some error handling scenarios) and then we can get this merged.

PR #4396, opened 2023-08-13
Drawer refactor take3

elParaguayo:
There is an apparent issue/limitation with the cairo library where setting the `OPERATOR_CLEAR` library in a `RecordingSurface` will not clear the background when painted to an `XCBSurface`. This means that, for our x11 backend code, we need to clear the underlying `XCBSurface`
directly.

To simplify this, our `Drawer.clear` method (which actually clears and paints a background colour) has been updated to include a call to a new method, `clear_rect`, which does the erasing of the background. For X11, this new method clears the `XCBSurface` rather than the recording surface.

The new method also accepts optional parameters to allow us to clear a subset of the `Drawer`. This is currently only used by third_party libraries (i.e. qtile-extras) and should help them avoid the need for writing backend-specific code.

### DIFF
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -264,15 +264,38 @@ class Drawer:
         else:
             ctx.set_source_rgba(*utils.rgb(colour))
 
-    def clear(self, colour):
+    def clear_rect(self, x=0, y=0, width=0, height=0):
+        """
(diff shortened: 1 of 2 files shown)


m-col (approved):
LGTM. Can confirm everything works as expected in x11, wayland, xephyr, wephyr. Incl. with qtile-extras widgets at c503f10 (master as i post this).

PR #4398, opened 2023-08-13
Wayland XWindow: Only restack the surface on unhide

jwijenbergh:
Previously the bring_to_front fix was made such that xwayland dialogs had proper focus. However, the only part of bring_to_front we need is the surface restacking. If we do a full bring_to_front then xwayland toplevels will come on top of fullscreen toplevels.

Let's do this 'hacky' patch up until we fix it properly with z-layering.

To give some explanation in proper english, bug on master which this fixes:

Spawn xterm (anything that uses xwayland), spawn alacritty/foot (anything that runs wayland). Fullscreen the wayland terminal, xterm will still be on top.

### DIFF
--- a/libqtile/backend/wayland/xwindow.py
+++ b/libqtile/backend/wayland/xwindow.py
@@ -162,7 +162,12 @@ class XWindow(Window[xwayland.Surface]):
                     self.tree.node.set_position(self.borderwidth, self.borderwidth)
 
                 self.container.node.set_enabled(enabled=True)
-                self.bring_to_front()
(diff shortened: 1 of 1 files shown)


dr-carlos:
Just wondering if there is any progress on this PR?

jwijenbergh:
> Just wondering if there is any progress on this PR?

This pr should improve things quite a lot (at least it did for me). It's a bit hacky but I think this should be fine to be merged. Thought? @m-col @elParaguayo

elParaguayo:
If it's an improvement on where we are currently then I'm ok with it as we're coming up for a release.

PR #4401, opened 2023-08-14
[docs] Update hooks page

elParaguayo:
One regular bit of feedback is that our docs should be more user friendly for beginners. This PR adds a bit more detail on hooks and code examples for each hook.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -68,97 +68,290 @@ class Subscribe:
         **Arguments**
 
         None
+
+        Example:
+
+        .. code:: python
+
+          import os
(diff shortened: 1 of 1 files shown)


tych0 (approved):
This looks great, thank you very much!

One other thing I have always thought is that the left sidebar is kind of a mess. I hadn't looked at it in a while, but it looks like you also cleaned that up earlier this year. Thank you for that as well.

elParaguayo:
Ha, yes. That was me. The sidebar was very ugly so I decided to tidy it up a bit.

Thanks for taking a look at this.

PR #4404, opened 2023-08-16
Fix empty space bug in bar draw

elParaguayo:
Where widgets do not completely fill the bar (which happens if there is no `SPACER`-type widget), the bar should be filled with the background colour. However, this is not currently happening as the bar is only cleared/filled when drawing borders. This is particularly noticeable on Wayland.

This PR addresses this by ensuring the bar background is always painted whenever `Bar.draw()` is called.


m-col (approved):
lgtm

PR #4405, opened 2023-08-18
Split Wayland and X11 tests in CI

elParaguayo:
This PR runs the X11 and Wayland backends in parallel runners which shortens our run time. Tests completed in 22 mins on my fork.

The `tox.ini` file is pretty messy and needs some love but we can do that separately.

One issue I have with this PR is that it makes it a bit harder for users to run`tox` locally. This is because, to save duplicating lines, I set the `--backend` option by using an environmental variable (which is set in Github Actions). To run tests, users would need to run `BACKEND=wayland tox -e py311`. If we add more lines to `tox.ini`, we could have users run `tox -e py311-wayland` (I'm assuming most users aren't going to run full test suite locally on both backends).

Branch protections would need updating too as I've changed job names.

Would welcome thoughts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 jobs:
     build:
         runs-on: ubuntu-22.04
-        name: "python ${{ matrix.python-version }}"
+        name: "python ${{ matrix.python-version }} on ${{ matrix.backend }}"
         strategy:
             matrix:
                 # If you change one of these, be sure to update:
@@ -17,6 +17,7 @@ jobs:
                 # If adding new python versions, consider also updating
                 # python version in .readthedocs.yaml
                 python-version: [pypy-3.9, 3.9, '3.10', '3.11']
+                backend: ['x11', 'wayland']
         steps:
             - uses: actions/checkout@v3
             - name: Set up python ${{ matrix.python-version }}
@@ -31,14 +32,18 @@ jobs:
                   libdbus-1-dev libgirepository1.0-dev gir1.2-gtk-3.0 gir1.2-notify-0.7 gir1.2-gudev-1.0 graphviz \
                   imagemagick git xserver-xephyr xterm xvfb dbus-x11 libnotify-bin \
                   libxcb-composite0-dev libxcb-icccm4-dev libxcb-res0-dev libxcb-render0-dev libxcb-res0-dev \
-                  libxcb-xfixes0-dev vlc volumeicon-alsa
+                  libxcb-xfixes0-dev vlc volumeicon-alsa libxkbcommon-dev
                 sudo pip -q install meson PyGObject
-                pip -q install tox tox-gh-actions 
-                bash -x ./scripts/ubuntu_wayland_setup
+                pip -q install tox tox-gh-actions
+            - name: Install wayland
+              if: ${{ matrix.backend == 'wayland' }}
+              run: bash -x ./scripts/ubuntu_wayland_setup
             - name: Run Tests
               run: |
                 [ "$(grep -c -P '\t' CHANGELOG)" = "0" ]
                 tox
+              env:
+                BACKEND: ${{ matrix.backend }}
             - name: Upload coverage data to coveralls.io
               # For speed purposes, we don't do coverage reporting on pypy 3.9
               if: ${{ matrix.python-version != 'pypy-3.9' }}
@@ -47,7 +52,7 @@ jobs:
                 coveralls --service=github
               env:
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-                COVERALLS_FLAG_NAME: ${{ matrix.python-version }}
+                COVERALLS_FLAG_NAME: ${{ format('{0}-{1}', matrix.python-version, matrix.backend) }}
                 COVERALLS_PARALLEL: true
                 COVERALLS_SERVICE_JOB_ID: ${{ github.run_id }}
 
--- a/tox.ini
+++ b/tox.ini
@@ -3,19 +3,43 @@ skip_missing_interpreters = True
 skipsdist=True
 minversion = 4.0.12
 envlist =
-    pypy3,
-    py39,
-    py310,
-    py311,
+    # Python environments with specific backend
+    py{py3,39,310,311}-{x11,wayland}
     docs,
     pep8,
     codestyle,
     docstyle,
     mypy,
     packaging,
-    vulture
+    vulture,
+    # For running pytest locally
+    test-{x11,wayland,both}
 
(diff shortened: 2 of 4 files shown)


tych0:
It looks like test coverage goes "down" since you've split things, but IMO coverage as a metric is a bit of a red herring. I'd be fine with just deleting it entirely.

Otherwise, lgtm.

elParaguayo:
It goes down on individual workers because they're running fewer tests. Howver, coverage merges all the results together at the end so, really, the important result is this one:
`Coverage remained the same at 74.516%`

elParaguayo:
I should rebase this onto the lastest master to see if anything broke.

tych0:
Ah, I missed that there's an aggregate thanks. So these will fail on the first merge, and then pass the rest of the time I guess? Not really obvious to me why the individual split runs failed.

elParaguayo:
The individual ones failed because we've told `coveralls` to fail if there's a coverage drop of >= 1% (I think). There's no prior commit here with these worker names so it can't workout the coverage change. I guess it reports that as a failure but it should be ok on subsequent PRs after this merged and they're rebased onto this.

tych0:
Agreed. Should we force-merge this then and change the branch protections?

elParaguayo:
I'd like to rebase this and tidy up the commit history first. Can do tonight.

smurfix:
Gentle question about "tonight" being two weeks off by now? ;-)

NB, thanks for doing this work!

elParaguayo:
It is actually rebased already. However, we're really just trying to sort packaging so we can the next release out as it's been one year (today) since the last!

This isn't urgent for that work but will certainly be nice to have for future PRs.

.github/workflows/ci.yml
```diff
@@ -31,14 +32,18 @@ jobs:
...
                   libxcb-composite0-dev libxcb-icccm4-dev libxcb-res0-dev libxcb-render0-dev libxcb-res0-dev \
-                  libxcb-xfixes0-dev vlc volumeicon-alsa
+                  libxcb-xfixes0-dev vlc volumeicon-alsa libxkbcommon-dev
                 sudo pip -q install meson PyGObject
-                pip -q install tox tox-gh-actions 
-                bash -x ./scripts/ubuntu_wayland_setup
+                pip -q install tox tox-gh-actions
+            - name: Install wayland
+              if: ${{ matrix.backend == 'wayland' }}
+              run: bash -x ./scripts/ubuntu_wayland_setup
```
m-col, line 40: Nice :fire:

scripts/ubuntu_wayland_setup
```diff
@@ -17,7 +17,7 @@ LIBDRM=2.4.114
 # Packaged dependencies
 sudo apt update
 sudo apt install --no-install-recommends \
-    ninja-build libegl1-mesa-dev libgles2-mesa-dev libgbm-dev libinput-dev libxkbcommon-dev \
+    ninja-build libegl1-mesa-dev libgles2-mesa-dev libgbm-dev libinput-dev \
```
m-col, line 20: libxkbcommon-dev isn't needed? huh

tox.ini
```diff
@@ -3,66 +3,110 @@ skip_missing_interpreters = True
...
+    {[base]commands}
+
+    # GithubActions will trigger one of these three lines as the
+    # backend environment is set.
     # pypy3 is very slow when running coverage reports so we skip it
-    pypy3: python3 -m pytest --backend=x11 --backend=wayland {posargs}
-    py39: coverage run -m pytest --backend=x11 --backend=wayland {posargs}
-    py3{10,11}: coverage run -m pytest --backend=x11 --backend=wayland {posargs}
+    pypy3-{wayland,x11}: python3 -m pytest --backend={env:BACKEND} {posargs}
+    py39-{wayland,x11}: coverage run -m pytest --backend={env:BACKEND} {posargs}
```
m-col, line 62: Would something along these lines allow for running just the tox command without the env var?:
```suggestion
    pypy3-x11: python3 -m pytest --backend=x11 {posargs}
    pypy3-wayland: python3 -m pytest --backend=wayland {posargs}
    py39-x11: coverage run -m pytest --backend=x11 {posargs}
    py39-wayland: coverage run -m pytest --backend=wayland {posargs}
```

m-col (approved):
Nice

scripts/ubuntu_wayland_setup
elParaguayo: This is because it's installed in the main ci build. If it's only needed for Wayland then it should be changed back to how it was.

tox.ini
elParaguayo: You do still need the env var in ci.yml but your change above is definitely neater.

m-col: Does this remove the need for users to set the env var locally?

elParaguayo: I don't think they ever did. The env var is set by CI. What have I missed?

m-col: From the original PR message:

> One issue I have with this PR is that it makes it a bit harder for users to runtox locally. This is because, to save duplicating lines, I set the --backend option by using an environmental variable (which is set in Github Actions). To run tests, users would need to run BACKEND=wayland tox -e py311.

elParaguayo: Ah, ok! Let me check but I don't think that's needed. I created some helper environment names and referred to these in the docs (e.g. `tox -e test-both`).

However, thinking about this some more, the split leads to some inefficiency if these are run in sequence as both the Wayland and x11 tests will run the same backend-agnostic tests.

m-col: Presumably someone would always run the both env, and not the x11 and then wayland ones separately? At least hopefully - in the former case the non-backend related tests would only be run once fortunately

elParaguayo:
Ah, yes, indeed.

scripts/ubuntu_wayland_setup
elParaguayo: I think I remember the issue: `mypy` need it to install `xkbcommon`. If I split out the backends, when `mypy` runs on an x11 backend, installing that package will fail. This wouldn't have been an issue before because the wayland backend was always installed.

m-col: Mupy doesnt install packages does it? Do you mean tox?

elParaguayo: Yes, the `mypy` environment run by `tox`. Doesn't that need the wayland stuff to be built?

elParaguayo:
Cool - this looks ok now so I'll tidy it all up and then I think we can merge and reset the branch protections.

scripts/ubuntu_wayland_setup
m-col: Yes i think so

PR #4406, opened 2023-08-18
Wayland: Support paint_borders for internal windows

jwijenbergh:
This makes popups with borders work on wayland, because the SceneTree of these windows closely mimic actual windows we should try & refactor this to avoid the duplicated code and isinstance check.


elParaguayo:
Can confirm that borders paint on popup windows correctly with this patch.

elParaguayo:
Fine by me

PR #4413, opened 2023-08-24
[Docs] Fix RTD build error

elParaguayo:
Looks like `html_theme` needs to be set in config.py.


elParaguayo:
Closing this as commit added to #4420

PR #4417, opened 2023-08-25
[x11 layering] Fix bug with multiple transients

elParaguayo:
Google Earth identified a couple of issues with the layering code:
- The `transient_for` property for some windows returned the ID of a window that was not being managed by qtile. As a result, the parent window would not have been stacked and placing the new window above this would therefore put the new window in the wrong stack position.
- Transient windows were excluded from the list of windows in different layers and so, if an app has multiple transient windows, a new window would not be stacked after the last transient.

Draft for now as I'd like some feedback.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1062,7 +1062,7 @@ class _Window:
                     for k, v in self.qtile.windows_map.items()
                     if v.window.get_wm_transient_for() == parent
                 )
-                window_group.sort(key=lambda w: stack.index(w))
+                window_group.sort(key=stack.index)
 
                 # Make sure we're above the last window in that group
                 sibling = window_group[-1]
@@ -1183,8 +1183,30 @@ class _Window:
             stackmode=xcffib.xproto.StackMode.Above if above else xcffib.xproto.StackMode.Below,
             sibling=sibling,
         )
-        # TODO: also move our children if we were moved upwards
+
+        # Move window's children if we were moved upwards
+        if above:
+            self.raise_children(stack=stack)
+
         self.qtile.core.update_client_lists()
+
+    def raise_children(self, stack=None):
+        """Ensure any transient windows are moved up with the parent."""
+        children = [
+            k
+            for k, v in self.qtile.windows_map.items()
+            if v.window.get_wm_transient_for() == self.window.wid
+        ]
+        if children:
+            if stack is None:
+                stack = list(self.qtile.core._root.query_tree())
+            parent = self.window.wid
+            children.sort(key=stack.index)
+            for child in children:
+                self.qtile.windows_map[child].window.configure(
+                    stackmode=xcffib.xproto.StackMode.Above, sibling=parent
+                )
+                parent = child
 
     def paint_borders(self, color, width):
         self.borderwidth = width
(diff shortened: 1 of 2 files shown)


libqtile/backend/x11/window.py
```diff
@@ -1044,14 +1044,15 @@ def change_layer(self, up=True, top_bottom=False):
         windows.sort(key=lambda w: stack.index(w[0].wid))
 
         # Get lists of windows on lower, higher or same "layer" as window
-        lower = [w[0].wid for w in windows if w[1] is None and w[2] > layering]
-        higher = [w[0].wid for w in windows if w[1] is None and w[2] < layering]
-        same = [w[0].wid for w in windows if w[1] is None and w[2] == layering]
```
elParaguayo, line 1049: Is there any reason why we should exclude transient windows when calculating the list of windows in lower, same and higher layers? I can't think of any.

If we don't need to exclude them then I think we can remove the transient field from the windows tuple.

libqtile/backend/x11/window.py
```diff
@@ -1044,14 +1044,15 @@ def change_layer(self, up=True, top_bottom=False):
...
+        lower = [w[0].wid for w in windows if w[2] > layering]
+        higher = [w[0].wid for w in windows if w[2] < layering]
+        same = [w[0].wid for w in windows if w[2] == layering]
 
         # We now need to identify the new position in the stack
 
         # If the window has a parent, the window should just be put above it
-        if parent:
+        # If the parent isn't being managed by qtile then it may not be stacked correctly
+        if parent and parent in self.qtile.windows_map:
```
elParaguayo, line 1055: I've no idea why Google Earth says that some dialogs are transient windows for a window that we're not managing (and not displaying). Is there a better way to handle that scenario>

m-col: If we aren't managing the parent window, could we ignore the transient hint and treat it as a normal window?

libqtile/backend/x11/window.py
```diff
@@ -1044,15 +1044,32 @@ def change_layer(self, up=True, top_bottom=False):
...
+            # If the window is modal then it should be placed above every other window that is in that window group
+            # e.g. the parent of the dialog and any other window that is also transient for the same parent.
+            if "_NET_WM_STATE_MODAL" in self.window.get_net_wm_state():
+                window_group = [parent]
+                window_group.extend(
+                    k
+                    for k, v in self.qtile.windows_map.items()
+                    if v.window.get_wm_transient_for() == parent
+                )
+                window_group.sort(key=lambda w: stack.index(w))
```
m-col, line 1065:
```suggestion
                window_group.sort(key=stack.index)
```

libqtile/backend/x11/window.py
m-col: The only thing that matters is they are above their parent, but that doesn't require excluding them here does it?

libqtile/backend/x11/window.py
elParaguayo: That's essentially what that line does. i.e. if we're not managing it, this condition isn't met and we'll fall back to normal rules.

elParaguayo:
I probably need to add something that moves transients back about the parent if the parent is brought to the front.

m-col (approved):

PR #4419, opened 2023-08-27
PulseVolume: remove FFI dependency

elParaguayo:
This PR rewrites the `PulseVolume` widget to rely on an external library, `pulsectl_asyncio`. The advantage of this is that we can remove our FFI bindings which will, in turn, simplify building and packaging.


ramnes (approved):
Can confirm that it works very well, even better than the previous version (handles sink changes). Only one little regression: when updating the volume, the bar flickers. That's particularly visible with the scroll wheel. I really don't think that's a blocker though.

elParaguayo:
Let me check that. I suspect we're getting multiple callbacks from the server, triggering a draw on each one whereas we only need one.

tych0 (approved):
Thanks for handling this!

elParaguayo:
@ramnes I have _hopefully_ fixed the flickering (only calling `bar.draw()` if the length of the widget has changed.

ramnes:
Can confirm, that works!

PR #4423, opened 2023-08-29
docs: drop note about old ubuntus

tych0:
17.04 (and 18.04) are now end of life, so this is no longer relevant.

Reported-by: Christian Reis <kiko@luizalabs.com>


tych0:
I wonder if we need a `--no-build-isolation` arg here too with more modern pips? thoughts?

PR #4433, opened 2023-09-01
Fix number of bugs in the bar

elParaguayo:
c840df2 introduced a change to Bar.draw to ensure that a bar was completely filled even when there were insufficient widgets to fill the bar.

However, this introduced a bug:

1) When the bar is drawn, it clears the entire bar and then redraws
   every widget. However, the clearing is unnecessary as widgets also
   clear the surface before they are drawn. The bar's clearing therefore
   results in flickering of the bar as there may be a gap between
   clearing the bar and the widgets drawing themselves, particularly
   when there are a large number of widgets in the bar.

That bug revealed another:

2) Drawing the borders also cleared the entire bar so flickering would
   also occur when a user used borders in their config.

Which, in turn, revealed another:

3) Bar borders were not being rendered correctly for vertical bars
   (surprised this wasn't flagged earlier...)

This PR addresses all of these issues by:
- Removing the need to clear the bar's surface when it's drawn. Borders will now only clear the part of the bar covered by those borders.
- Similarly, filling any bar space no occupied by widgets will only draw in that space.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -264,9 +264,9 @@ class Bar(Gap, configurable.Configurable, CommandObject):
                     self._length = self.height
                     self._size += margin[1] + margin[3]
                     if screen.left is self:
-                        self.x += margin[3]
+                        self.x += margin[3] - self.border_width[3]
                     else:
-                        self.x -= margin[1]
+                        self.x -= margin[1] + self.border_width[1]
 
             self._reserved_space_updated = False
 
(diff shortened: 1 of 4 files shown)


libqtile/bar.py
```diff
@@ -628,42 +623,79 @@ def _actual_draw(self) -> None:
...
-                self.drawer.ctx.set_line_width(border_width)
-                self.drawer.ctx.move_to(*move_to)
-                self.drawer.ctx.line_to(*line_to)
-                self.drawer.ctx.stroke()
-
-        self.drawer.draw()
+                self.drawer.ctx.fill()
+                src_x, src_y, width, height = rect
+
+                # We#re
```
ramnes, line 654: # Probably#unnecessary :P

ramnes (approved):
LGTM

jwijenbergh (approved):
Very tidy and the added comments are helpful

tych0:
I'm going to merge this, it seems like the type of thing we'd want in a release.

Issue #4421, opened 2023-08-28
[Wayland] Segmentation faults after the upgrade to wlroots 0.16


arjan-s:
### The issue:

After the upgrade to wlroots 0.16 was merged, Qtile started crashing regularly on both my systems, at least once or twice a day. The crashes happen on various actions, like closing or opening windows, or even just clicking a link in qutebrowser.

In my system journal I found that it appears to be a segmentation fault in wlroots. So I modified my wlroots to include debug symbols, and I got the below stacktraces (with `coredumpctl debug` and then `bt full` in gdb) from a few of those crashes.

```
#0  0x00007fb35dea15c3 in scene_node_get_root (node=node@entry=0x30f8c20) at ../types/scene/wlr_scene.c:50
        tree = 0xa1
#1  0x00007fb35dea2e97 in wlr_scene_node_destroy (node=0x30f8c20) at ../types/scene/wlr_scene.c:94
        scene = <optimized out>
        __PRETTY_FUNCTION__ = "wlr_scene_node_destroy"
#2  0x00007fb35df5caef in _cffi_f_wlr_scene_node_destroy () from /nix/store/vwqgwq7wg6kxijlknpj4fa925zinmdgs-python3-3.10.12-env/lib/python3.10/site-packages/wlroots/_ffi.abi3.so
No symbol table info available.
#3  0x00007fb360bd7b35 in cfunction_vectorcall_O (func=<optimized out>, args=<optimized out>, nargsf=<optimized out>, kwnames=<optimized out>) at Objects/methodobject.c:516
        tstate = 0x1c0bd10
        nargs = 1
        meth = <optimized out>
        result = <optimized out>



#0  0x00007f4a0ecbd5c3 in scene_node_get_root (node=node@entry=0x2418500) at ../types/scene/wlr_scene.c:50
        tree = 0x81
#1  0x00007f4a0ecbee97 in wlr_scene_node_destroy (node=0x2418500) at ../types/scene/wlr_scene.c:94
        scene = <optimized out>
        __PRETTY_FUNCTION__ = "wlr_scene_node_destroy"
#2  0x00007f4a0ed78aef in _cffi_f_wlr_scene_node_destroy () from /nix/store/vwqgwq7wg6kxijlknpj4fa925zinmdgs-python3-3.10.12-env/lib/python3.10/site-packages/wlroots/_ffi.abi3.so
No symbol table info available.
[...]

PR #4435, opened 2023-09-02
WLRQ: Setup the dnd destroy listener using self.icon

jwijenbergh:
As far as I can tell we're not actually destroying it double but just setting up the signal wrong

It seems like if we use wlr_drag.destroy_event we can get segfaults when destroying. Using self.icon for the destroy event fixes this. Fixes one crash of #4421


arjan-s:
I can confirm this patch fixes that crash, thanks!

PR #4437, opened 2023-09-04
Fix Wayland CFFI build

elParaguayo:
This PR changes how the Wayland CFFI modules are built when the package is installed. It addresses the `ModuleNotFound` error by adding the build environment to the python path to ensure modules can be found before they're installed.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -1,3 +1,9 @@
+import sys
+from pathlib import Path
+
+path = Path(__file__).parent / ".." / ".." / ".." / ".."
+sys.path.insert(0, path.absolute().as_posix())
+
 import wlroots.ffi_build as wlr
 from cffi import FFI
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ include LICENSE
 include README.rst
 include MANIFEST.in
 include CONTRIBUTING.md
+include libqtile/backend/wayland/_ffi*.so
 
 exclude .coveragerc
 exclude .pylintrc


elParaguayo:
NB. Installing from git using pip (e.g. in qtile-extras test suite) shows the version number is not being calculated correctly:
```
2023-09-04T06:21:09.4682760Z Successfully installed qtile-0.0.0
```

libqtile/backend/wayland/cffi/build.py
```diff
@@ -1,3 +1,9 @@
+import sys
+from pathlib import Path
+
+path = Path(__file__).parent / ".." / ".." / ".." / ".."
```
ramnes, line 4: wut?

elParaguayo: Ha!

So, as far as I can tell, there was an issue when trying to build this as I got a modulenotfound error for qtile.

This line is just working out the path for the root folder (i.e. the folder containing libqtile). `Path(__file__)` gives the path to `build.py`, `.parent` is used to get the folder and then you can use the `/ ".."` syntax as if it were a file path to move up levels.

elParaguayo:
I don't know if this is the correct solution but it might help identify the real issue.

tych0:
I'm confused. Don't we want to have people either: 1. pass `--no-build-isolation` themselves explicitly or 2. have the right libs/headers/etc. in their virtualenv?

elParaguayo:
My understanding is that pyproject.toml doesn't allow _optional_ build dependencies (which pywlroots would be). As such, if you don't pass `--no-build-isolation` the installer can't see the user's existing pywlroots installation.

I haven't tested this theory by adding pywlroots in the build section though...

Issue #4346, opened 2023-07-18
`pip install qtile` in a fresh environment fails


3301n:
### The issue:

I wanted to install qtile using the command pip install qtile, but it showed a lot of error messages, and I expected it do a clean install . I think this is a bug. I am not in root but as a user installation. 
I am using python version 3.11.4 and pip version 23.2 in a debian based operating system.
![Screenshot_2023-07-18_04_55_55](https://github.com/qtile/qtile/assets/69240615/df85225d-5081-490c-935b-b641daeaaa32)

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
What version of `cairocffi` and `xcffib` do you have installed?

elParaguayo:
(separately, what window manager are you using in that screenshot?)

ramnes:
This is what happens when you try to run `pip install qtile` in a fresh environment:

```
$ virtualenv .venv
created virtual environment CPython3.11.4.final.0-64 in 824ms
  creator CPython3Posix(dest=/home/ramnes/tmp2/.venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/ramnes/.local/share/virtualenv)
    added seed packages: pip==23.1.2, setuptools==68.0.0, wheel==0.40.0
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
$ source .venv/bin/activate
(.venv) $ pip install qtile
Collecting qtile
  Downloading qtile-0.22.1.tar.gz (488 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 489.0/489.0 kB 3.7 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [41 lines of output]
      Failed to find pywlroots. Wayland backend libinput configuration will be unavailable.
      Traceback (most recent call last):
[...]

PR #4440, opened 2023-09-04
pyproject: more (hopefully the last?) porting

tych0:
Here's the last bits of what I think we should port to pyproject.toml before releasing.

* I dropped the "Alpha" Classifier, we're beyond alpha, I've been using qtile as a daily driver for nearly 15 years
* I dropped the point-release python version Classifiers (i.e. 3.8, 3.9, etc.). These are not that interesting and a pain to maintain.
* I dropped Sean Vig as the listed maintainer. We list all the maintainers at the bottom of the readme, which ends up in the long_description. No reason to maintain two maintainer lists.
* I dropped Aldo as the Author:. There are lots of authors of qtile, this info is available from the git log.

My hope is that this will not give a syntax error when uploading source now during the release workflow. I've inspected the output with:

    python3 setup.py bdist_wheel --keep-temp && cat build/bdist.linux-x86_64/wheel/qtile-*.dist-info/METADATA

and it looks pretty similar to the METADATA from the current release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,25 @@ readme = {file = "README.rst"}
 [project]
 name = "qtile"
 description = "A pure-Python tiling window manager."
-dynamic = ["version"]
+dynamic = ["version", "readme"]
+license = {text = "MIT"}
+classifiers = [
+    "Intended Audience :: End Users/Desktop",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: POSIX :: BSD :: FreeBSD",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Desktop Environment :: Window Managers",
+]
+
+[project.urls]
+"Homepage" = "https://qtile.org"
+"Documentation" = "https://docs.qtile.org/"
+"Code" = "https://github.com/qtile/qtile/"
+"Issue tracker" = "https://github.com/qtile/qtile/issues"
+"Contributing" = "https://docs.qtile.org/en/latest/manual/contributing.html"
 
 [project.scripts]
 qtile = "libqtile.scripts.main:main"


elParaguayo:
Ha. Didn't want my review then?

elParaguayo (commented):
This looks good. Two comments/questions:

1) Do we need something like:
```
[tool.setuptools.dynamic]
readme = {file = ["README.rst"]}
```
or does it automatically look for a file with that name?

2) I'm not certain if the dynamic version numbering is working. I've seen some 0.0.0 versions following your previous commit (as part of trying to get the Wayland FFI build to work) but I've not really looked too closely at it.

tych0:
1. [yep, we have that](https://github.com/qtile/qtile/blob/f45dc910ada928dae63b9db7ae89bf4c285909a8/pyproject.toml#L19-L20), we just weren't using it.
2. very possible; I didn't actually build the wayland stuff myself. But there is a Version: in your METADATA file right? IIUC that's the important bit for e.g. pypa or our dynamic version imports or whatever.

tych0:
I wonder if we should fix this too:

```
2023-09-04T15:45:13.2477480Z packaging: commands[1]> python3 setup.py check -m -s
2023-09-04T15:45:13.5507424Z /home/runner/work/qtile/qtile/.tox/packaging/lib/python3.11/site-packages/setuptools/__init__.py:84: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
2023-09-04T15:45:13.5508951Z !!
2023-09-04T15:45:13.5509066Z 
2023-09-04T15:45:13.5509200Z         ********************************************************************************
2023-09-04T15:45:13.5509517Z         Requirements should be satisfied by a PEP 517 installer.
2023-09-04T15:45:13.5509974Z         If you are using pip, you can try `pip install --use-pep517`.
2023-09-04T15:45:13.5510286Z         ********************************************************************************
2023-09-04T15:45:13.5510426Z 
2023-09-04T15:45:13.5510493Z !!
```

elParaguayo:
Let me check versioning again tonight. If that's ok then I think this one's fine.

Agreed on the other point but that can be tidied up separately.

tych0:
Yeah, I'm out of momentum for figuring out python packaging stuff right now :)

elParaguayo:
So versioning does work if you don't use `--no-build-isolation` but the Wayland stuff doesn't build correctly. If you use that flag then you get a `ModuleNotFoundError: No module named 'libqtile'` error. Fixing that error (e.g. #4437) gives a version number of 0.0.0.

tych0:
Can you paste the output? I have never gotten the wayland stuff to build right on my laptop.

ramnes (approved):
LGTM

Since we're here, can't we port and remove setup.cfg and setup.py as well? 

I think there's a chance it would fix the version problem (just guessing)

tych0:
> Since we're here, can't we port and remove setup.cfg and setup.py as well?

Someone could :). I really just want a minimal release right now.

elParaguayo:
> LGTM
> 
> Since we're here, can't we port and remove setup.cfg and setup.py as well?
> 
> I think there's a chance it would fix the version problem (just guessing)

I don't think pyproject.toml supports `cffi_modules` so I think we're stuck with setup.py in some form.

elParaguayo:
Done some more reading.

We may be able to do this by using a custom backend builder (which wraps `setuptools.build_meta`). We can inject build dependencies at that stage and we can specify option requirements too. So, if there's a way for the builder to check if we've run something like `pip install qtile[wayland]` then we can add the necessary build requirements.

Injecting the build requirement allowed me to successfully build wayland. I was still using `setup.py` but I think that code could also be moved into the builder script.

It's late now but I'll look into this more tomorrow.

elParaguayo:
The version info fails with `--no-build-isolation` as I think this just ignores the build system section entirely.

As for Wayland, building works if you add `pywlroots` to the `requirements` section in `[build-system]`. That feels like a bad idea as I'm assuming `pywlroots` will not install without Wayland headers on the system (but I'm just guessing).

My idea above of injecting a build requirement has some issues too. I can't find a way to detect that a user has added `[wayland]` to the package name. We can pass config options (e.g. `-C backend=wayland`) and inject the build requirement if that is set. However, this is a build requirement, not package requirement so `pywlroots` is removed after installing qtile. This is solveable if we do `pip install -C backend=Wayland qtile[Wayland]` but that is clunky. It may be possible to inject package dependencies with the same `-C` flag but I haven't solved that yet.

So... would it make more sense to have 2 packages: `qtile` and `qtile-wayland` with separate `pyproject.toml` files for each? Or package the qtile Wayland ffi bindings separately and then make that an optional dependency.

I'm really just figuring out how pyproject.toml works but would be great to get @m-col and @flacjacket thoughts here.

We do need a quick solution though as we're way past due for a release.

m-col:
For the optional dependency sets we can use https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dependencies-optional-dependencies

From what I can tell this would continue to allow `pip install qtile[wayland]` :)

elParaguayo:
Yes - but you can't specify additional _build_ dependencies that way. Doing this only means that `pywlroots` will be installed at some point with qtile. However, it's not available in the build environment so the ffi build fails.

ramnes:
We're already back to the optional build dependencies problem, awesome. :)

Looks like a `qtile-wayland` package would be the right thing™ to do, see the end of the discussion here: https://github.com/pypa/pipenv/issues/4482#issuecomment-706572894

tych0:
Isn't the idea of migrating to cffi's in-line API and porting the PA widget that we /wouldn't/ need these build deps? Since we're not compiling anything, we don't need any -dev headers or whatever, and it will all work at import time? Or am I misunderstanding something.

tych0:
Also, I think I will merge this for now, since there don't really seem to be any objections to this patch. I'll keep pushing on a release. (We should continue this discussion though, and I'm happy to continue it here.)

elParaguayo:
> Isn't the idea of migrating to cffi's in-line API and porting the PA widget that we /wouldn't/ need these build deps? Since we're not compiling anything, we don't need any -dev headers or whatever, and it will all work at import time? Or am I misunderstanding something.

We haven't migrated the wayland ffi build and I don't think we intend to make that in-line as it would have a slight performance impact.



> We're already back to the optional build dependencies problem, awesome. :)
> 
> Looks like a `qtile-wayland` package would be the right thing™ to do, see the end of the discussion here: [pypa/pipenv#4482 (comment)](https://github.com/pypa/pipenv/issues/4482#issuecomment-706572894)

Yes - that's a good summary of the issue.

I prefer the separate package to a hacky solution requiring some custom command to install the wayland backend correctly.

elParaguayo:
Interesting CI error when this was merged:
```
2023-09-06T02:35:28.2934227Z ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
2023-09-06T02:35:28.2935595Z     pywlroots<0.17.0,>=0.16.4 from https://files.pythonhosted.org/packages/55/c7/fdaa9084fa6d77d6d60c33dcc7d776ea119682e0f7849ec0e0b54507f213/pywlroots-0.16.4-cp39-cp39-manylinux_2_28_x86_64.whl:
2023-09-06T02:35:28.2936236Z         Expected sha256 82972b9d6ca9f5b4d3388fc444b53b6a796f44f3265afc9d42fd6e26d3bb2419
2023-09-06T02:35:28.2979012Z              Got        793bcece515bafe969df98456f9d9140835cf2e409a25c1e48e4a81b4aacc7d8
```
Not seen that one before! Have rerun the tests to see if it was just a glitch.

elParaguayo:
I don't know enough about CFFI to comment but that was my original understanding from when we were discussing xcffib and cairocffib changes.

If that is correct then it doesn't sound like two packages is the right approach as we wouldn't need a custom build environment and could just have the optional import.

However @m-col and @flacjacket should be the ones to comment on this.

elParaguayo:
If that's the only implication then that would get my vote. I've spent far more time than I care to admit looking at ways we could get packaging to work!

tych0:
Yep, same here. And for what it's worth, I haven't noticed any real slowdown with xcffib imports. I'm sure it's there, but computers are fast, and it takes firefox and spotify much longer to start when I log in than qtile itself :)

flacjacket:
Yeah, I think the main difference between in-line and out-of-line ABI mode is the import time, which makes sense to be minor. I think this summarizes where moving to ABI-mode would be hard for pywayland/pywlroots: https://github.com/qtile/qtile/pull/4307#issuecomment-1597233946. There is the in-line API mode, but that was a huge pain back when that was the only way to distribute the API modules (pre-cffi 0.9, I think) since you might need to compile the library at import time. The cffi docs seem to tilt heavily towards favoring API mode, and do note that there is a (large?) speed up at run time for API mode over ABI mode. https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api It may be less of a concern for xcffib, since a lot of the heavy lifting there is being done by the x server, where a wayland server needs to be doing a lot more work, I think.

flacjacket:
Is the main issue now that there are optional components in qtile that need to be compiled against the pywayland/wlroots cffi builder? We could see if there are things there that make sense to push over to one of those two projects, or a different module. The libinput stuff could stand alone (pylibinput?) but the cairo buffer stuff seems like it might be a bit harder to find a good place for if it is called "cairo buffer", but the functionality there seems like it is a general purpose wlroots buffer. It _might_ be something that we can reconstruct in Python using some of the pywayland callback functionality, but if not, we could think about moving that to pywlroots. If those optional built dependencies are gone from qtile, then qtile can dynamically check to see if the required deps are installed or not at run time, which I think should get rid of the biggest pain point for the compiled modules here. Does that make sense, or is there something else I'm missing here?

flacjacket:
Ah, and maybe the need for the cairo buffer cffi implementation is the need to define the `handle_begin_data_ptr_access` and `handle_end_data_ptr_access` callbacks (check me on that @m-col ?). That could be something we add to pywlroots.

tych0:
>  The cffi docs seem to tilt heavily towards favoring API mode, and do note that there is a (large?) speed up at run time for API mode over ABI mode. https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api

Ah, this is interesting. I'm remain unconvinced that performance is a problem for us, but maybe it is for wayland. Qtile should mostly be doing nothing, the interesting thing is the contents of the windows, not the windows/decorations themselves.

jwijenbergh:
> > The cffi docs seem to tilt heavily towards favoring API mode, and do note that there is a (large?) speed up at run time for API mode over ABI mode. https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api
> 
> Ah, this is interesting. I'm remain unconvinced that performance is a problem for us, but maybe it is for wayland. Qtile should mostly be doing nothing, the interesting thing is the contents of the windows, not the windows/decorations themselves.

This is true for x11, not for wayland. We need to also e.g. do input handling, output handling, *render* everything (sure mostly handled by wlroots now we're using a higher level API), etc.

tych0:
> input handling, output handling, render everything

Help me understand here, since I'm mostly wayland ignorant. It looks to me like there are similar events for mouse enter/leave for example. wayland seems like it needs to do a little extra cursor configuration here, but what output handling or rendering is there to do?

That said, I don't mind if you want to maintain wayland with a glue package. I'm skeptical that it's worth it, but I also don't maintain that part, so I won't argue with you :).

Are there any objections to attempting another release this weekend? I realize there are at least two things that are broken:
1. wayland packaging
2. qtile --version when installed with --no-build-isolation

I will figure out how to fix the first one (though maybe after a release).

m-col:
> Ah, and maybe the need for the cairo buffer cffi implementation is the need to define the handle_begin_data_ptr_access and handle_end_data_ptr_access callbacks (check me on that @m-col ?). That could be something we add to pywlroots.

Yep.

> Are there any objections to attempting another release this weekend? I realize there are at least two things that are broken:
>
>    wayland packaging

Just to confirm exactly the implications - this will only not work for others that try to package in an environment without the build deps, but our own pypi packaging will work and install fine because we have control over that env... is that correct?

m-col:
> Looks like a qtile-wayland package would be the right thing™ to do, see the end of the discussion here: https://github.com/pypa/pipenv/issues/4482#issuecomment-706572894

Reading through that thread I'm actually warming up to the idea of a `qtile-wayland` package. Would that let us avoid having to do proper profiling of ABI-mode? :)

> Help me understand here, since I'm mostly wayland ignorant. It looks to me like there are similar events for mouse enter/leave for example. wayland seems like it needs to do a little extra cursor configuration here, but what output handling or rendering is there to do?

The qtile process does everything that the X server process would do. Rather than subscribing to events emitted by the server as in X, the qtile process is actually the process that dictates what is going on and fires events (though events are mostly fired within wlroots, then subscribed to from within the qtile python code). Of course, all the heavy lifting is done within wlroots C code - but we do a _lot_ of jumping between python and c(ffi) contexts.

tych0:
> Just to confirm exactly the implications - this will only not work for others that try to package in an environment without the build deps, but our own pypi packaging will work and install fine because we have control over that env... is that correct?

I'm not sure; based on [this](https://github.com/qtile/qtile/pull/4307#issuecomment-1595345971) comment I don't understand how it ever worked in the past, so I was thinking doing a release now would be business as usual.

I guess the regression here is with `pip install --no-build-isolation` as @elParaguayo notes above?

flacjacket:
> Questions for @m-col and @flacjacket: If we add `pywlroots` to the build requirements I'm assuming this is installed with `pip` as well:
> 
>     * Does `pywlroots` have built wheels for this or do users need to have all the dev libraries on their machine.

Yes, pywayland and pywlroots both have wheels

> 
>     * If there are wheels, can we just have `pywlroots` as a permanent build requirement? or, does the build environment version need to be the same as the package that the user has on their system (e.g. if they installed `python-pywlroots` for their distro)?

This is something I've wondered about myself, but haven't gotten around to testing. I'm not sure what happens when there is a version mismatch between the pywlroots/pywayland wheel libraries and the system libraries.

tych0:
I think for releases it'll be fine to pin versions and then revert immediately. I think it would be useful to have `qtile --version` work for bug reports during development though, so I do think it's useful to fix if we're going to continue to require `--no-build-isolation`.

tych0:
(and I volunteer to fix it post release.)

elParaguayo:
There's another potential issue to flag:
Using this new build method results in wheels being built. Wheels ignore `Manifest.in` so the docs, tests etc all get packaged. However, because of our folder structure, these folders don't sit within `libqtile` but as sister folders to it so users would end up with a `test` folder in `site-packages`. That doesn't seem right to me.

elParaguayo:
> I think for releases it'll be fine to pin versions and then revert immediately. I think it would be useful to have `qtile --version` work for bug reports during development though, so I do think it's useful to fix if we're going to continue to require `--no-build-isolation`.

In which case, adding `pywlroots` to the build requirements _should_ be sufficient but we need to make sure we then build wheels against the same python versions as `pywlroots`.

tych0:
> Using this new build method results in wheels being built. Wheels ignore Manifest.in so the docs, tests etc all get packaged. However, because of our folder structure, these folders don't sit within libqtile but as sister folders to it so users would end up with a test folder in site-packages. That doesn't seem right to me.

Ah, damn. I suppose there's some new equivalent of Manifest.in. I love rewriting all this for the third or fourth time :)

elParaguayo:
Is there a way to do test releases to check we've got everything working? Feels like there are a lot of changes!

tych0:
Yeah, @ramnes was tagging things like `v0.23.0-betaN`, so I started doing the same (and deleting the tags when they didn't work, though I think I left the last one up.) It's quite slow iteration though, i found it faster to cat the METADATA file and see if it looked reasonable.

m-col:
I guess worst case re wayland is someone tries to install and it fails. Their previous WM will work fine even  if it was the current qtile version. No harm no foul. Feeling reassured, +1 to release :)

tych0:
Yeah, I'll fix this MANIFEST.in thing before I tag anything real. We definitely don't want to pollute peoples' site-packages.

elParaguayo:
Might be as simple as:
```
[tool.setuptools.packages.find]
include = ["libqtile*"]
```

tych0:
Sadly, that doesn't work :(. There is this https://github.com/pypa/setuptools/issues/3341 which suggests that perhaps it isn't implemented at all?

But it does look like poetry has a nice implementation. This has actually happened several times to me throughout this process (poetry has nice docs, setuptools doesn't even have an implementation), which makes me wonder if we can switch? I guess the cffi stuff might be a problem.

elParaguayo:
Really? When I did a `pip install` with that change it only installed the `libqtile` folder, no docs, no test. Might be an issue if we need to get more granular.

tych0:
Ah ha, it does work! Maybe I just had leftover stuff in my build dir. Anyway, by cleaning it manually (why doesn't setuptools do this?):

```
~/packages/qtile/build/bdist.linux-x86_64/wheel master ls
libqtile/  qtile-0.23.0b3.dev0+gf45dc910.d20230911.dist-info/
```

clotodex:
Just an FYI. Installing qtile is getting a bit wild, multiple issues have different answers to you being unable to install qtile. I was following five different issues until I found this random discussion of build isolation not working on a merge pull request. (while other issues pointed out to use it to fix some other issues). My [qtile-installer](https://github.com/clotodex/qtile-install-script) is working again now :D
It might make sense to have a dedicated "install command" listed in the readme, or a section that clarifies what is the *current* way. Because the "installation" part of the qtile doc was not updated for a while.
Also see https://github.com/qtile/qtile/issues/4346

I love that it is getting better and easier to install qtile though. Just hard to keep track is all :D

ramnes:
It's complicated to install because our packaging is broken, and it's broken because Python packaging authorities decided to break multiple things in the way we currently package Qtile. That's why we're trying to fix all the mess ASAP rather than documenting the current (broken) way. With the next release it will be installable with a mere `pip install qtile`. It's taking quite a bit of time because we're all busy in life and Qtile is not our top 1 priority. PR welcome.

elParaguayo:
To add to what ramnes said, I think we're virtually there on the packaging. #4459 aims to fix the remaining problems. We just need to do a test build/upload to pypi but I'm not familiar with how to do that!

PR #4441, opened 2023-09-07
README: drop www from website url

tych0:
The SSL cert github serves is not valid for www.qtile.org, just qtile.org. Let's drop the www here.


elParaguayo (approved):

jwijenbergh:
Could we also get a valid cert for www or is this not possible?

tych0:
It probably is, but we still do not have real control of the DNS, so it will be a while to make any changes there.

PR #4444, opened 2023-09-07
Fix incorrect typing with Dropdown config

Sigmanificient:
Remove the `Dict` typing from the `config` parameter of dropdown `__init__` method as kwargs are annotated only by there content.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -1046,7 +1046,7 @@ class DropDown(configurable.Configurable):
         ),
     )
 
-    def __init__(self, name: str, cmd: str, **config: dict[str, Any]) -> None:
+    def __init__(self, name: str, cmd: str, **config: Any) -> None:
         """
         Initialize :class:`DropDown` window wrapper.
 


elParaguayo:
Thanks for taking the time to submit this.

I'm curious, though. Why do you think this is incorrect?

I'm not an expert on type hinting but the mypy tests have no issue with this.

Sigmanificient:
> I'm not an expert on type hinting but the mypy tests have no issue with this.

I think this is due to the typing being used on the user config side
image

From python [pep 0484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values):

> 
> *Arbitrary argument lists can as well be type annotated, so that the definition:*
> 
> ```py
> def foo(*args: str, **kwds: int): ...
> ```
> *is acceptable and it means that, e.g., all of the following represent function calls with valid types of arguments:*
> ```
> foo('a', 'b', 'c')
> foo(x=1, y=2)
> foo('', z=0)
> ```
>
> *In the body of function foo, the type of variable `args` is deduced as `Tuple[str, ...]` and the type of variable `kwds` is `Dict[str, int]`.*

Thus, `dict[str, Any]` is resolved as `dict[str, Any]`

Sigmanificient:
Btw i think using `typing.overload` could be a nice options for later on to have more precise type hinting.

Here is little example from a old project of mine:

```py

    @overload
    async def create_guild(
        self,
        *,
        name: str,
        region: Optional[str] = None,
        icon: Optional[str] = None,
        roles: Optional[List[Role]] = None,
        channels: Optional[List[Channel]] = None,
        afk_channel_id: Optional[Snowflake] = None,
        afk_timeout: Optional[int] = None,
    ) -> Guild:
        ...

    async def create_guild(self, name: str, **kwargs) -> Guild:
        g = await self.http.post("guilds", data={"name": name, **kwargs})
        return await self.get_guild(g["id"])
```

elParaguayo:
Thanks for the info. I phrased my previous post badly - I wasn't saying you were wrong, I was curious how you identified this.

The change looks fine to me.

ramnes (approved):

ramnes:
Thanks!

PR #4446, opened 2023-09-09
Pulse widget - refactor to use single server connection

elParaguayo:
Previously, each widget instance created its own connection to the pulse server. This is unnecessary given all widgets are retrieving the same data.

This PR addresses this by creating a single connection and allowing widgets to subscribe to that one connection.


m-col (approved):
Good idea :)

PR #4447, opened 2023-09-10
Fix two bugs when adding/removing screens

elParaguayo:
There are two bugs that can arise when a screen is disconnected and plugged back in:

1) Disconnecting the screen calls `kill_window()` on the gaps on that screen. However, that method does not include a call to cancel any scheduled timers (e.g. `self.draw()`) so the bar my try to draw finalised widgets resulting in a PangoError when the bar tries to get the size of the widgets in `self._resize()`.

2) Reconnecting a screen could result in widgets on that screen not updating. This was because the screen uses previously initialised/configured widgets and so `self.configured` and `self.finalized` were both `True`. This prevented widgets from being reconfigured correctly as well as blocking any timers.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -432,6 +432,9 @@ class Bar(Gap, configurable.Configurable, CommandObject):
         """Kill the window when the bar's screen is no longer being used."""
         assert self.qtile is not None
 
+        if self.future:
+            self.future.cancel()
+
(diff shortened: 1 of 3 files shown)


jwijenbergh (approved):
LGTM

m-col (approved):
Nice

PR #4448, opened 2023-09-10
Rewrite Bluetooth widget

elParaguayo:
This widget provides additional functionality above the `Bluetooth` widget. It allows users to pair/connect/disconnect devices, show battery level, toggle power status of adapters.

In addition, configuration is easier as users do not need to know the dbus object path of their devices.

Draft for now in case anyone wants to test it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
       for more information on how to run tests locally.
       - Add `ScreenSplit` layout which allows multiple layouts per screen. Also adds `ScreenSplit`
         widget to display name of active split.
+      - Updated `Bluetooth` widget which allows users to manage multiple devices in a single widget
     * bugfixes
 
 Qtile 0.23.0, released 2023-09-24:
--- a/libqtile/scripts/migrations/change_bluetooth_args.py
+++ b/libqtile/scripts/migrations/change_bluetooth_args.py
@@ -1,0 +1,103 @@
+# Copyright (c) 2023, elParaguayo. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import libcst as cst
+import libcst.matchers as m
+
+from libqtile.scripts.migrations._base import (
+    Check,
+    MigrationTransformer,
+    _QtileMigrator,
+    add_migration,
+)
+
+
+class BluetoothArgsTransformer(MigrationTransformer):
+    @m.call_if_inside(
+        m.Call(func=m.Name("Bluetooth")) | m.Call(func=m.Attribute(attr=m.Name("Bluetooth")))
+    )
+    @m.leave(m.Arg(keyword=m.Name("hci")))
+    def update_bluetooth_args(self, original_node, updated_node) -> cst.Arg:
+        """Changes positional  argumentto 'widgets' kwargs."""
+        self.lint(
+            original_node,
+            "The 'hci' argument is deprecated and should be replaced with 'device'.",
+        )
+        return updated_node.with_changes(keyword=cst.Name("device"))
+
+
+class BluetoothArgs(_QtileMigrator):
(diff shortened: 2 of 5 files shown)


elParaguayo:
Any better name suggestions `BluetoothExtra` is a bit rubbish!

m-col:
Obvious question I almost hate to ask - why not just have these features integrated(/replacing) the existing bluetooth widget?

elParaguayo:
It's a fair question. My thinking was that its behaviour is quite different to the current widget but I also think the current widget's behaviour is pretty useless!

m-col:
Yeah this one looks far superior, I think it would be fine to drop the old one. It only has the one parameter too, so maybe a small effort to consider that argument in `__init__` and use it to configure this one might be straight forward. Presumably everyone using that one would be happy to migrate :)

elParaguayo:
I could make it so, if that parameter is set then it defaults to showing that device (rather than the other text).

m-col:
I don't think it matters if the appearance changes for them. Would it be feasible to use it for `adapter_paths` and then continue as normal with the way this widget works (if I've understood the implementation correctly)?

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+        except RuntimeError:
+            pass
+
+    def properties_changed(self, _interface_name, _changed_properties, _invalidated_properties):
+        """Handler for properties_changed signal."""
+        create_task(self.update_props())
+
+    async def update_props(self):
+        raise NotImplementedError
+
```
m-col, line 83: Is this method necessary? It seems it's only used where we know it won't error, so it might be fine just being defined on `BluetoothDevice`?

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+BLUEZ_SERVICE = "org.bluez"
+BLUEZ_DEVICE = "org.bluez.Device1"
+BLUEZ_ADAPTER = "org.bluez.Adapter1"
+BLUEZ_BATTERY = "org.bluez.Battery1"
+OBJECT_MANAGER_INTERFACE = "org.freedesktop.DBus.ObjectManager"
+PROPERTIES_INTERFACE = "org.freedesktop.DBus.Properties"
+
+
+def catch_error(msg):
```
m-col, line 37: nit: could this be `_catch_error`?

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+
+    Exposes basic properties/methods and also listens to signals to
+    update properties as needed.
+    """
+
+    def __init__(self, path, interface, properties_interface, widget):
+        _BluetoothBase.__init__(self, path, interface, properties_interface, widget)
+        self._connected = False
+        self._paired = False
+        self._status = 2  # 0: connected, 1: paired & unconnected, 2: unpaired
```
m-col, line 97: Maybe worth an enum?

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+
+    def _wrapper(func):
+        async def f(self):
+            try:
+                await func(self)
+            except DBusError:
+                # flake8 doesn't like formatting of log messages unless
+                # message will be inlcude in log.
+                # Mute the error here as it keeps code near.
+                logger.warning(msg.format(self=self))  # noqa: G001
```
m-col, line 48: I think flake8 is just wanting this:
```py
msg = "Unable to connect to device: %s."

... # msg passed as arg to decorator

logger.warning(msg, self.name)
```

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+    def remove_battery(self):
+        self.battery_device = None
+        self._battery = 0
+
+    async def get_battery(self):
+        proxy = await self.widget.get_proxy(self.path)
+        try:
+            self.battery_device = proxy.get_interface(BLUEZ_BATTERY)
+        except InterfaceNotFoundError:
+            return
```
m-col, line 174: There are a few ignored exceptions. Might be bit cleaner to use `contextlib.suppress`?

libqtile/widget/bluetooth_extra.py
```diff
@@ -0,0 +1,634 @@
...
+        try:
+            self._name = await self.interface.get_name()
+            self.has_name = True
+        except (AttributeError, DBusError):
+            self._name = await self.interface.get_address()
+            self.has_name = False
+
+        self._connected = await self.interface.get_connected()
+        self._paired = await self.interface.get_paired()
+        self._adapter = await self.interface.get_adapter()
```
m-col, line 188: Do these need to happen in series? If not, the three awaitables could be awaited simultaneously

elParaguayo:
Thanks for comments - I'll pick those up later.

libqtile/widget/bluetooth_extra.py
elParaguayo: Why?

m-col: To indicate to others (users, contributors) that they shouldn't import from this widget module just because they want a similar function. If they do, we should factor it out into something that can be shared by both widgets

libqtile/widget/bluetooth_extra.py
elParaguayo: It's just there in case someone wants to extend the widget in the future.

libqtile/widget/bluetooth_extra.py
elParaguayo: How would you propose doing that? I doubt there's any meaningful performance benefit of doing this.

m-col: > How would you propose doing that?

Like this:
```py
self._connected, self._paired, self._adapter = asyncio.gather(
    self.interface.get_connected(),
    self.interface.get_paired(),
    self.interface.get_adapter(),
)
```

>  I doubt there's any meaningful performance benefit of doing this.

You're probably right :) unless for some reason your dbus connection was slow (is remote dbus a thing?)

elParaguayo: Well that looks neat. I'll make that change (although I suspect I need to `await` the `.gather` object.

And, yes, remote dbus is a thing but I can't imagine people would be using it much (and we don't provide the option of specifying a different dbus connection path).

elParaguayo:
Right - this is rebased. Should be good to go.

m-col (approved):
lgtm :)

PR #4449, opened 2023-09-10
Fix layout issue with killed scratchpads

elParaguayo:
When a scratchpad is killed (rather than toggled) it can prevent layouts from setting layout correctly when tiled windows are closed.

I'm not entirely sure why this happens but the bug is fixed by having `layout._ClientList.remove` return the next window which allows the layout to focus the window correctly.

If you want to test this, first verify the bug exists in current master:
- Use `MonadTall` layout
- Open two windows (they open side by side)
- Close left-hand window, remaining window fills layouts
- Open another window (so you have two windows again)
- Open a scratchpad
- Kill scratchpad window (not hide)
- Close currently focused window
- Layout does not correctly adjust layout

### DIFF
--- a/libqtile/layout/base.py
+++ b/libqtile/layout/base.py
@@ -345,6 +345,8 @@ class _ClientList:
         elif idx <= self._current_idx:
             self._current_idx = max(0, self._current_idx - 1)
 
+        return self[self._current_idx]
+
     def rotate_up(self, maintain_index=True):
         """
         Rotate the list. The first client is moved to last position.


m-col:
I was trying to understand what was going on here and adding the type info helped (https://github.com/qtile/qtile/pull/4461). Some layouts were suggesting that `remove()` always returned `None` but sometimes it returned a `Window`. The type hints in that PR make that clear, so code that calls that method know to handle both cases. In the case of this bug, I suspect that's the floating layout, which has `remove()` called in group.py.

Accordingly, the change in this PR is also part of https://github.com/qtile/qtile/pull/4461 so if we're happy to, we can supercede this one with that PR.

elParaguayo:
Closing in favour of #4461

PR #4452, opened 2023-09-11
only include libqtile in our wheel

### LINKED ISSUE #4443: Dependency conflict on Arch
### Issue description

I tried to update my packages this morning and saw it failed with a conflict between python-pipenv and qtile-git (installed through AUR). 

```bash
error: failed to commit transaction (conflicting files)
python-pipenv: /usr/lib/python3.11/site-packages/docs/__pycache__/conf.cpython-311.opt-1.pyc exists in filesystem (owned by qtile-git)
python-pipenv: /usr/lib/python3.11/site-packages/docs/__pycache__/conf.cpython-311.pyc exists in filesystem (owned by qtile-git)
python-pipenv: /usr/lib/python3.11/site-packages/docs/conf.py exists in filesystem (owned by qtile-git)
```

I uninstalled qtile temporarily to resolve the conflict and update the rest of the packages, but attempting to reinstall qtile fails with the same errors:

```bash
error: failed to commit transaction (conflicting files)
qtile-git: /usr/lib/python3.11/site-packages/docs/__pycache__/conf.cpython-311.opt-1.pyc exists in filesystem (owned by python-pipenv)
qtile-git: /usr/lib/python3.11/site-packages/docs/__pycache__/conf.cpython-311.pyc exists in filesystem (owned by python-pipenv)
qtile-git: /usr/lib/python3.11/site-packages/docs/conf.py exists in filesystem (owned by python-pipenv)
Errors occurred, no packages were upgraded.
```

Was there a change in qtile or in python-pipenv that is leading to this conflict? What can I do to resolve it?

### Version

qtile-git-0.23.0.beta2.r4.g2c955718-1

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

tych0:
This makes me wonder about two other things people have asked us to distribute:

1. the tests/, so maintainers can run our automated tests on their builds
2. resources/ (for qtile.desktop, admittedly small)

But hey, at least we aren't dumping stuff in peoples' site-packages/ with this commit.

Fixes #4443

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,9 @@ enabled = true
 [tool.setuptools.dynamic]
 readme = {file = "README.rst"}
 
+[tool.setuptools.packages.find]
+include = ["libqtile*"]
+
 [project]
 name = "qtile"
 description = "A pure-Python tiling window manager."


elParaguayo:
> 1. the tests/, so maintainers can run our automated tests on their builds

Would this need to be packaged? I'd expect maintainers to rely on keeping their git fork up to date with the main branch.

> 2. resources/ (for qtile.desktop, admittedly small)

Yes - I was thinking about this too. We could make these files very clear in our docs so people can copy and paste but we do also want packagers to be able to include these files (e.g. https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/blob/main/PKGBUILD?ref_type=heads#L118-L119)

Do we just move them into `libqtile/resources` instead?

Also, this approach doesn't package the license, readme etc.

Why does this have to be so difficult?!

tych0:
> Would this need to be packaged? I'd expect maintainers to rely on keeping their git fork up to date with the main branch.

Not all distros run git master :).  At least some of them package releases themselves, and I think it's those folks who wanted the tests included. I don't remember where we were asked for this, but I can dig it up.

> Also, this approach doesn't package the license, readme etc.

Oh, good point. Maybe we need some clever use of `include-package-data` https://setuptools.pypa.io/en/latest/userguide/datafiles.html

tych0:
Here is what I was thinking of: https://github.com/tych0/xcffib/issues/59

So not qtile, but I guess the same logic applies.

jwijenbergh:
So this doesn't include the docs?

tych0:
Correct.

tych0:
In particular, it looks like that report was from a gentoo dev. But it seems there's more work to do here :(.

It also seems like this segfault is happening on lots of PRs, maybe there's some new flakiness in some of the wayland bindings?

elParaguayo:
I think 3 or 4 of my last PRs had only failed on pypy39 and it was just the statusnotifier tests (both x11 and wayland) but it seems to be worse now.

The new pywlroots release must be a good bet.

tych0:
I have been playing around with this more, and I don't understand at all how it is supposed to work. It seems like it includes some stuff from MANIFEST.in, but not the `graft` stuff?

elParaguayo:
> Here is what I was thinking of: [tych0/xcffib#59](https://github.com/tych0/xcffib/issues/59)
> 
> So not qtile, but I guess the same logic applies.

If we include tests in the wheel, doesn't that mean we're creating a `tests` folder in `dist-packages`? That doesn't seem right.

Also, surely we're not the only project with this issue? It's frustrating that this seems to be so complicated.

tych0:
I think (?) that if it's called `test/`, there's some special casing. I had to rename from tests -> test in xcffib: https://github.com/tych0/xcffib/commit/541abb843c27026aca37184a488e255341467d7d and it doesn't install one.

tych0:
Heh. This is ridiculous. A wheel is just a tar file. What if we write our own minimal packaging solution that just generates the thing in the right format? It is crazy that we have spent this long on this problem trying to get setuptools/pyproject/etc. to cooperate together.

elParaguayo:
Should tests actually be included in a wheel? I would have thought that a wheel is just to install the package in your system in a way that it can just be run. If you want to run tests etc, wouldn't you download the sdist instead so you can then build and test?

elParaguayo:
Is there a reason why we can't just package `libqtile` in the wheel and just have the full suite with `docs`, `tests`, `*.desktop\LICENCE\README` files etc in the `sdist`? Most users aren't going to want the `docs`, `tests` folders. They'll have to go through a bit of effort to install a desktop file but they'd have to do that anyway if installing directly with pip. The licence is already on the website so we can include the desktop files easily enough too.

Packagers can download the source, build locally and run tests etc. They can then package other files as they see fit.

All of that feels achievable to me.

Thoughts? (pinging @ramnes @jwijenbergh @m-col too)

ramnes:
Yeah I do agree and I'd vouch for this. Now I'm not a package maintainer so I probably don't know all the details... I'd say let's do it that way and if someone is unhappy we'll see why?

elParaguayo:
Great. Then we probably want to add something like this:
```diff
diff --git a/docs/manual/install/index.rst b/docs/manual/install/index.rst
index ce54099a..1b2e9006 100644
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -196,3 +196,8 @@ window:

 See the :ref:`Wayland <wayland>` page for more information on running Qtile as
 a Wayland compositor.
+
+Similar to the xsession example above, a wayland session file can be used to start qtile
+from a login manager. To use this, you should create a `qtile-wayland.desktop
+<https://github.com/qtile/qtile/blob/master/resources/qtile-wayland.desktop>`_ file in
+``/usr/share/wayland-sessions``.
```

Then there's the question of whether we add `pywlroots` to the build dependencies as that would remove the need for ``--no-build-isolation`` which breaks the versioning code.

m-col:
> Then there's the question of whether we add pywlroots to the build dependencies as that would remove the need for --no-build-isolation which breaks the versioning code.

Seems reasonable

jwijenbergh:
> Great. Then we probably want to add something like this:
> 
> ```diff
> diff --git a/docs/manual/install/index.rst b/docs/manual/install/index.rst
> index ce54099a..1b2e9006 100644
> --- a/docs/manual/install/index.rst
> +++ b/docs/manual/install/index.rst
> @@ -196,3 +196,8 @@ window:
> 
>  See the :ref:`Wayland <wayland>` page for more information on running Qtile as
>  a Wayland compositor.
> +
> +Similar to the xsession example above, a wayland session file can be used to start qtile
> +from a login manager. To use this, you should create a `qtile-wayland.desktop
> +<https://github.com/qtile/qtile/blob/master/resources/qtile-wayland.desktop>`_ file in
> +``/usr/share/wayland-sessions``.
> ```
> 
> Then there's the question of whether we add `pywlroots` to the build dependencies as that would remove the need for `--no-build-isolation` which breaks the versioning code.

Looks good

tych0:
Closing in favor of #4459

PR #4453, opened 2023-09-11
Pin pywlroots to 0.16.4 and skip StatusNotifier tests

jwijenbergh:
title

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ wayland =
   xkbcommon>=0.3
   # NOTE: When updating the major or minor version of pywlroots here, also update:
   # tox.ini, docs/manual/install/index.rst, docs/manual/wayland.rst
-  pywlroots>=0.16.4,<0.17.0
+  pywlroots==0.16.4
 
 [options.package_data]
 libqtile.resources =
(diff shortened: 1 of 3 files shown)


elParaguayo:
@m-col @flacjacket Do either of you know why we're getting segfaults with latest pywlroots update?

elParaguayo:
If this passes then this should be squashed and merged.

@jwijenbergh if the Statusnotifier tests keep failing then just disable them in tox and I'll fix them later (if we get a coverage failure because coverage has dropped too much as a result then that can be ignored too.)

elParaguayo:
Ok Statusnotifier failed again. Let's just disable that test.

I've been working on a replacement test but let's not hold things up for that.

jwijenbergh:
I have also pinned it in setup.cfg but maybe that's not needed. But maybe good that we're consistent

jwijenbergh:
> Ok Statusnotifier failed again. Let's just disable that test.
> 
> I've been working on a replacement test but let's not hold things up for that.

I removed the whole file

elParaguayo:
That will break the docs as the screenshot imports some parts from the test.

You might be better skipping the test.

jwijenbergh:
> That will break the docs as the screenshot imports some parts from the test.
> 
> You might be better skipping the test.

Oh I now removed the wait for icon, not good like this?

jwijenbergh:
Okay nvm they're now skipped, thanks

m-col:
Probably worth updating the docs too, where we have the supported pywlroots version range listed?

elParaguayo:
Ok. So I need to unpin this in the packaging PR.

PR #4458, opened 2023-09-16
[#4457] Introduce guess_browser() function

nikitabugrovsky:
Implement `guess_browser()` function

### DIFF
--- a/docs/manual/config/keys.rst
+++ b/docs/manual/config/keys.rst
@@ -22,6 +22,7 @@ key. The basic operation is:
 * ``mod + <group name>``: switch to that group
 * ``mod + <shift> + <group name>``: send a window to that group
 * ``mod + <enter>``: start terminal guessed by ``libqtile.utils.guess_terminal``
+* ``mod + b``: start browser guessed by ``libqtile.utils.guess_browser``
 * ``mod + r``: start a little prompt in the bar so users can run arbitrary
   commands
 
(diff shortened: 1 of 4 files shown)


m-col:
Hey, thanks for this. I think this is a bit out of scope as the slippery slope is that we do the same for every kind of app. The difference with a terminal is that, without out, new users can get stuck not knowing how to do anything, and it gives them an escape hatch through which they can edit their config, open other apps from the terminal etc to help them out of that situation. @elParaguayo @jwijenbergh thoughts?

jwijenbergh:
> Hey, thanks for this. I think this is a bit out of scope as the slippery slope is that we do the same for every kind of app. The difference with a terminal is that, without out, new users can get stuck not knowing how to do anything, and it gives them an escape hatch through which they can edit their config, open other apps from the terminal etc to help them out of that situation. @elParaguayo @jwijenbergh thoughts?

I was about to comment the same thing. I think a terminal makes sense because it's really essential in just navigating with a tiling window manager, a browser less. So I prefer to not have this in core, maybe [qtile-examples](https://github.com/qtile/qtile-examples/)?

m-col:
> I prefer to not have this in core, maybe [qtile-examples](https://github.com/qtile/qtile-examples/)?

Sounds good :)

m-col:
@nikitabugrovsky if you'd be happy submitting it to qtile-examples, it would make a cool addition there :)

PR #4459, opened 2023-09-16
Another go at packaging

elParaguayo:
I've had another go at this.

What I've done here is include `pywlroots` in the build requirements and then use a custom build backend to try to build the cffi bindings. This will fail if wayland headers etc aren't available (e.g. on readthedocs) but it allows the build to continue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,28 +1,16 @@
 [build-system]
-requires = [ "setuptools>=41", "wheel", "setuptools-git-versioning<2", ]
-build-backend = "setuptools.build_meta"
-
-[tool.black]
-line-length = 98
-exclude = 'libqtile/_ffi.*\.py|libqtile/backend/x11/_ffi.*\.py|test/configs/syntaxerr.py'
-
-[tool.pytest.ini_options]
-python_files = "test_*.py"
-testpaths = ["test"]
-filterwarnings = [
-    "error:::libqtile",
+requires = [
+    "setuptools>=60",
+    "setuptools-scm>=7.0",
+    "wheel",
+    "pywlroots>=0.16.4,<0.17.0"
 ]
-
-[tool.setuptools-git-versioning]
-enabled = true
-
-[tool.setuptools.dynamic]
-readme = {file = "README.rst"}
+build-backend = "setuptools.build_meta"
 
 [project]
 name = "qtile"
 description = "A pure-Python tiling window manager."
-dynamic = ["version", "readme"]
+dynamic = ["version", "readme", "dependencies", "optional-dependencies"]
 license = {text = "MIT"}
 classifiers = [
     "Intended Audience :: End Users/Desktop",
@@ -30,6 +18,9 @@ classifiers = [
     "Operating System :: POSIX :: BSD :: FreeBSD",
     "Operating System :: POSIX :: Linux",
     "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
     "Topic :: Desktop Environment :: Window Managers",
@@ -44,3 +35,22 @@ classifiers = [
 
 [project.scripts]
 qtile = "libqtile.scripts.main:main"
+
+[tool.setuptools.packages.find]
+include = ["libqtile*"]
+
+[tool.setuptools_scm]
+
+[tool.setuptools.dynamic]
+readme = {file = "README.rst"}
+
+[tool.black]
+line-length = 98
+exclude = 'libqtile/_ffi.*\.py|libqtile/backend/x11/_ffi.*\.py|test/configs/syntaxerr.py'
+
+[tool.pytest.ini_options]
+python_files = "test_*.py"
+testpaths = ["test"]
+filterwarnings = [
+    "error:::libqtile",
+]
--- a/tox.ini
+++ b/tox.ini
@@ -68,10 +68,10 @@ commands =
 deps =
     check-manifest
     twine
+    build
 commands =
     check-manifest
-    python3 setup.py check -m -s
-    python3 setup.py sdist
+    python3 -m build --sdist .
     twine check dist/*
 
 [testenv:pep8]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,16 +6,150 @@ on:
   release:
     types: [published]
 jobs:
+  build-wayland:
+    name: Build Wayland libraries
+    runs-on: ubuntu-latest
+    container: quay.io/pypa/manylinux_2_28_x86_64
+    env:
+      libdrm-version: "2.4.114"
+      seatd-version: "0.6.3"
+      pixman-version: "0.42.0"
+      wayland-protocols-version: "1.31"
+      wayland-version: "1.22.0"
+      wlroots-version: "0.16.2"
+    steps:
+      - name: Install dependencies
+        run: |
+          yum -y install \
+            hwdata \
+            python3-pip \
+            libffi-devel \
+            libinput-devel \
+            libpciaccess-devel \
+            libudev-devel \
+            libxkbcommon-x11-devel \
+            libxml2-devel \
+            mesa-libEGL-devel \
+            mesa-libgbm-devel \
+            xcb-util-devel \
+            xcb-util-image-devel \
+            xcb-util-keysyms-devel \
+            xcb-util-renderutil-devel \
+            xcb-util-wm-devel \
+            xorg-x11-server-Xwayland \
+            ninja-build \
+            wget
+      - name: Set environment variables
+        run: |
+          echo "CPATH=${HOME}/wayland/usr/include" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=${HOME}/wayland/usr/lib" >> $GITHUB_ENV
+          echo "LIBRARY_PATH=${HOME}/wayland/usr/lib" >> $GITHUB_ENV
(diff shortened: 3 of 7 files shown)


qtile_builder.py
```diff
@@ -0,0 +1,73 @@
...
+    python path.
+    """
+    print("Building wayland backend cffi bindings...")
+    sys.path.insert(0, ROOT.as_posix())
+    python_path = ":".join(sys.path)
+    try:
+        subprocess.run(
+            [sys.executable, "libqtile/backend/wayland/cffi/build.py"],
+            check=True,
+            env=os.environ.update({"PYTHONPATH": python_path}),
```
m-col, line 43: `os.environ.update({"PYTHONPATH": python_path})` returns `None`. You need could do:
```suggestion
            env=os.environ | {"PYTHONPATH": python_path},
```

m-col (approved):
Is this held back due to the pywlroots bump? If so, should we just pin pywlroots in the meantime while we bisect that bug?

elParaguayo:
Yes - we should pin pywlroots for now, we need to get the test suite passing as it's holding us up.

qtile_builder.py
elParaguayo: Interesting. Maybe this isn't needed then as it built ok for me locally. Need to test again.

Also, should we just have this run `ffibuild` instead that way we can just manage building in a single script?

pyproject.toml
```diff
@@ -13,34 +19,21 @@ filterwarnings = [
...
-license = {text = "MIT"}
-classifiers = [
-    "Intended Audience :: End Users/Desktop",
-    "License :: OSI Approved :: MIT License",
-    "Operating System :: POSIX :: BSD :: FreeBSD",
-    "Operating System :: POSIX :: Linux",
-    "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: Implementation :: CPython",
-    "Programming Language :: Python :: Implementation :: PyPy",
-    "Topic :: Desktop Environment :: Window Managers",
```
tych0, line 35: Don't we want to keep most of this metadata? In particular, this is what it uses to render the sidebar here IIUC: https://pypi.org/project/qtile/

(You could probably drop all the classifiers, but at least the license and links to docs/homepage/code seem useful)

tych0:
Thanks for picking this up! I was clearly going in circles :)

pyproject.toml
elParaguayo: Yes - I'm not sure how that got deleted here!

pyproject.toml
```diff
@@ -13,34 +19,21 @@ filterwarnings = [
     "error:::libqtile",
 ]
 
-[tool.setuptools-git-versioning]
-enabled = true
+[tool.setuptools_scm]
```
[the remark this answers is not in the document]
elParaguayo: Apparently the header is enough. Agree it looks weird.

elParaguayo:
Have made some changes. wheel and sdist both seem to build ok with `python -m build .`

One question for the wayland folk, these are the compiled files packaged after running `ffibuild`:
```
   169160  2023-02-24 18:52   libqtile/backend/wayland/_ffi.cpython-310-x86_64-linux-gnu.so
   170128  2023-09-18 19:53   libqtile/backend/wayland/_ffi.cpython-311-x86_64-linux-gnu.so
   385776  2023-09-18 19:53   libqtile/backend/wayland/_ffi.o
    79240  2023-09-13 05:48   libqtile/backend/wayland/_ffi.pypy310-pp73-x86_64-linux-gnu.so
   153472  2023-04-21 16:56   libqtile/backend/wayland/_libinput.cpython-310-x86_64-linux-gnu.so
   153968  2023-07-31 20:13   libqtile/backend/wayland/_libinput.cpython-311-x86_64-linux-gnu.so
   132368  2022-07-05 20:03   libqtile/backend/wayland/_libinput.cpython-39-x86_64-linux-gnu.so
   348688  2023-07-31 20:13   libqtile/backend/wayland/_libinput.o
    73800  2023-04-21 18:36   libqtile/backend/wayland/_libinput.pypy39-pp73-x86_64-linux-gnu.so
```
`libinput` seems to have been built for different python versions compared to _ffi. I'm guessing this is to do with what's included in the pywlroots wheel but let me know if this looks odd to you.

elParaguayo:
This _should_ now pass all tests.

It would be great if some of the devs could test this locally making sure that it behaves as expected (i.e. wheel builds our wayland backend bindings, wheel excludes tests, sdist is complete etc.)

elParaguayo:
Looks like a need to fix manifest file!

tych0:
At the very least it looks like it'll build /all/ our dependent wheels:
```
~/packages/qtile fixing-packaging ls *whl
cairocffi-1.6.1-py3-none-any.whl
cffi-1.15.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
pycparser-2.21-py2.py3-none-any.whl
qtile-0.23.0b3.dev13+g9525a921-py3-none-any.whl
xcffib-1.5.0-py3-none-any.whl
```
which is maybe ok, but we should probably change the upload regex:
```
diff --git a/.github/workflows/release.yml b/.github/workflows/release.yml
index d6abb50b..021c00a7 100644
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: wheels-${{ matrix.python-version }}
-          path: dist/*.whl
+          path: qtile*.whl
   test-wheel:
     name: Test wheels
     runs-on: ubuntu-latest
```

is there a way we can avoid building the other wheels if we're not going to use them? also, what's "none" there? I assume "any" is the arch (which should be fine with all of our CFFI fixes/pulse widget ports, etc. at least on x11).

elParaguayo:
That's odd. Running `python -m build .` only gives me `qtile-0.23.0b3.dev13+g9525a921-py3-none-any.whl` and  `qtile-0.23.0b3.dev13+g9525a921.tar.gz`.

As for uploading to pypi, presumably we also need to upload the sdist

tych0:
I ran `pip3 wheel .`, so maybe that's the difference? Of course it is not at all obvious to me which one of these is the right entry point :)

tox.ini
```diff
@@ -68,10 +68,10 @@ commands =
 deps =
     check-manifest
     twine
+    build
 commands =
     check-manifest
-    python3 setup.py check -m -s
```
elParaguayo, line 73: I'm not aware of any full replacement for `python setup.py check` - `twine check` is the best we've got but it's not a drop-in solution.

qtile_builder.py
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2033, elParaguayo
```
elParaguayo, line 1: Ha. I'm a time traveller.

elParaguayo:
Now that tests pass I'll tidy this up tonight and we can get it merged.

How do we go about testing the release workflow and checking things are actually built and uploaded to pypi correctly?

tych0:
I've just been tagging `v0.23.0-betaN` style releases and running it. I guess if it succeeds it'll upload something to pypi, but hopefully pypi will only render things without beta? :) It would be a great problem to have that we had an accidental upload.

elParaguayo:
Are you pushing that tag to the upstream `master` branch after merging? I assume it won't work if I push to my fork.

tych0:
Not to master, I'm just pushing tags.

elParaguayo:
I'm obviously doing something wrong. I've added a tag to this branch and pushed it but it's not running.

tych0:
Are you sure you pushed it? I don't see anything other than v0.23.0-beta2 on github, which is a tag I created. I think (?) they need to be annotated tags too, which might be the difference.

elParaguayo:
@tych0 - you need to treat me like a complete beginner here!

I've added an annotated tag but the workflow is not running. The tag is on my fork if that makes a difference here. I don't see how I can add a tag to the main repo while this is just a PR. Do I need to add this to a new branch in the main repo to run?

tych0:
You won't need a new branch or anything, you should be able to `git push origin v0.23.0-beta3` or whatever. git is smart enough to upload any commits missing from that tag. Then to delete it (I should probably delete my broken beta2 tag, but I was leaving it so we could keep incrementing them in case I accidentally released something), you can do `git push origin :v0.23.0-beta3`.

elParaguayo:
I had to do a different branch but it seems to have worked.

However, is it correct that there's only one wheel? https://test.pypi.org/project/qtile/0.23.0b3/#files

tych0:
> However, is it correct that there's only one wheel?

Heh, I have no idea :). I can make a conainer and install it and see later today.

elParaguayo:
0.22.1 had multiple wheels, one for each python version but we're just getting the one here.

elParaguayo:
I think we're getting a single wheel because it's not building the wayland bindings :(

m-col:
> libinput seems to have been built for different python versions compared to _ffi. I'm guessing this is to do with what's included in the pywlroots wheel but let me know if this looks odd to you.

I think this is right - the cffi builder saves the file using the filename including the current python version. The files would all exist together if they were built in the same environment - but I guess they aren't?

elParaguayo:
Right. After too many hours on this, I think it's working as it should.

See: https://test.pypi.org/project/qtile/0.23.0b19/#files

m-col:
Can confirm that in a fresh venv I can install qtile with `pip install -i https://test.pypi.org/simple/ qtile==0.23.0b19 --extra-index-url https://pypi.org/simple` and X11 works as expected, Wayland works out of the box without our own build script needing to be run locally. The nested wayland session doesn't run, but the proper session works as intended. No need for build/install-time pywlroots too :) excellent work!

tych0:
I agree, it looks good from my perspective as well. Excellent (dare I say heroic?) work here. I'm going to merge this now, but please @elParaguayo do the hono(u)rs of tagging a new release. You earned it :)

elParaguayo:
@tych0 someone else will need to tag the release as I'm traveling with work until Friday now and don't have personal laptop.

Rest assured, I will still take great satisfaction in seeing this done!

tych0:
Ok, if someone doesn't beat me to it I'll do it tonight. Thank you!

PR #4460, opened 2023-09-16
Testing some stuff out on the CI to get more info from the segfaults

m-col:


elParaguayo:
`RuntimeError: Gtk couldn't be initialized. Use Gtk.init_check() if you want to handle this case.`
Not sure this is the cause but it's a weird error to pop up.

PR #4461, opened 2023-09-17
Add type information to base layout module and inherited methods

m-col:
This adds some type annotations to our layouts, focussing on `libqtile/layouts/base.py` which now has type annotations enforced, and then annotating any methods in other layouts that are inherited from the base layout.

This was motivated by some PRs I've read recently that relate to layouts and the lack of self-documentation that untyped methods are guilty of.

A few unused attributes and potential errors were also revealed by mypy while adding these :)


elParaguayo:
Amazing work as always.

PR #4450, opened 2023-09-11
feat: Add options to set initial caps lock and num lock state

Sydiepus:
This adds the ability  set an initial state for the caps lock and num lock using kb_capslock and kb_numlock options.

Closes:
https://github.com/qtile/qtile/issues/4225

PR #4467, opened 2023-09-23
WL: print message at wayland core.py ffi module no import

m-col:
Adds a slightly better message when this file is imported from but `lib` couldn't be loaded. We allow this so that the docs env doesn't need pywlroots. But sometimes if the core is used a `NameError` is raised when accessing `lib` at qtile startup, and it isn't clear why.

E.g. see #4450


jwijenbergh (approved):
LGTM

PR #4473, opened 2023-09-25
load icons for group-specific layouts

### LINKED ISSUE #4472: CurrentLayoutIcon widget does not load icons for group-specific layouts
### Issue description

When using the `CurrentLayoutIcon` widget with layouts that are defined group-only,
(For example: 
```py
layouts = [
    # Some layouts here
]
groups = [
    Group("0", layouts=[
        # Any layouts not included in `layouts` above
    ])
]
```
)
the icon does not properly load and display for layouts that are only defined for the group specifically.

This issue occurs because in the setup function for loading icons https://github.com/qtile/qtile/blob/fd93e74c110f23846af5f721195b6c1f90b873f8/libqtile/widget/currentlayout.py#L207-L235
 it only loads icons for layouts found by `_get_layout_names`, which only includes layouts defined under `layouts` in the main configuration https://github.com/qtile/qtile/blob/fd93e74c110f23846af5f721195b6c1f90b873f8/libqtile/widget/currentlayout.py#L174-L181

This means that layouts are accessible that do not have icons that are loaded, leading to errors in the form
```
2023-09-24 23:31:50,483 ERROR libqtile currentlayout.py:draw():L158 No icon for layout <layout name>
```
Indicating that the issue isn't that an icon file wasn't found (I have them present in `~/.icons`, and no error message saying they aren't found), but rather that an icon file was never loaded.

A potential solution would be to also go through each group in `_get_layout_names` to also include those layouts, deduplicating as necessary.

### Version

0.22.1

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

Infinidoge:
Fixes #4472

Technically `list(set(...))` is unnecessary because sets are iterable, but I kept it as a list to match the docstring.

Can remove the unnecessary list and fix the doc string, though.

### DIFF
--- a/libqtile/widget/currentlayout.py
+++ b/libqtile/widget/currentlayout.py
@@ -25,6 +25,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import itertools
 import os
 
 from libqtile import bar, hook
@@ -160,7 +161,8 @@ class CurrentLayoutIcon(base._TextBox):
                 self.drawer.clear(self.background or self.bar.background)
                 self.drawer.ctx.save()
                 self.drawer.ctx.translate(
-                    (self.width - surface.width) / 2, (self.bar.height - surface.height) / 2
+                    (self.width - surface.width) / 2,
+                    (self.bar.height - surface.height) / 2,
                 )
                 self.drawer.ctx.set_source(surface.pattern)
                 self.drawer.ctx.paint()
@@ -173,12 +175,16 @@ class CurrentLayoutIcon(base._TextBox):
 
     def _get_layout_names(self):
         """
-        Returns a list of tuples of lowercased layout name and class name strings for each available layout.
+        Returns a sequence of tuples of layout name and lowercased class name
+        strings for each available layout.
         """
-        return [
-            (layout.name, layout.__class__.__name__.lower())
-            for layout in self.qtile.config.layouts
-        ]
+
+        layouts = itertools.chain(
+            self.qtile.config.layouts,
+            (layout for group in self.qtile.config.groups for layout in group.layouts),
+        )
+
+        return set((layout.name, layout.__class__.__name__.lower()) for layout in layouts)
 
     def _update_icon_paths(self):
         self.icon_paths = []


Infinidoge:
I have a feeling it's going to yell at me for the black line length, because my usual line length limit is 120 instead of 80.

libqtile/widget/currentlayout.py
```diff
@@ -175,10 +176,13 @@ def _get_layout_names(self):
         """
         Returns a list of tuples of lowercased layout name and class name strings for each available layout.
```
m-col, line 177: Could change it :)
```suggestion
	Returns a sequence of tuples of lowercased layout name and class name
	strings for each available layout.
```

libqtile/widget/currentlayout.py
```diff
@@ -175,10 +176,13 @@ def _get_layout_names(self):
...
-            (layout.name, layout.__class__.__name__.lower())
-            for layout in self.qtile.config.layouts
-        ]
+
+        layouts = itertools.chain(
+            self.qtile.config.layouts,
+            (layout for group in self.qtile.config.groups for layout in group.layouts),
+        )
+
+        return list(set((layout.name, layout.__class__.__name__.lower()) for layout in layouts))
```
m-col, line 185:
```suggestion
        return set((layout.name, layout.__class__.__name__.lower()) for layout in layouts)
```

m-col (changes requested):
Excellent, thanks! :) Couple minor suggestions. I'll enable the CI

Infinidoge:
Something I noticed, `layout.name` isn't lowercased, when the docstring of the function indicates it should. Should that be changed?

m-col:
Good spot - reading the rest of the code it looks like it doesn't need to be lower case. Please could you drop that bit from the docstring?

Infinidoge:
Tests failed because I had
```py
import os
import itertools
```

Instead of
```py
import itertools
import os
```

Took a hot minute to come to that conclusion.

It would probably be a good idea if the style checks ran as early as possible so it could fail out early for small problems.

elParaguayo:
> It would probably be a good idea if the style checks ran as early as possible so it could fail out early for small problems.

That's not a bad idea! I agree it's frustrating when you wait 40 mins for main tests to pass and then get a failure for something like this!

Infinidoge:
I really need to fix my editor setup, because it seems that it can't decide on a line length to use and so it formats inconsistently.

libqtile/widget/currentlayout.py
```diff
@@ -160,7 +161,8 @@ def draw(self):
                 self.drawer.clear(self.background or self.bar.background)
                 self.drawer.ctx.save()
                 self.drawer.ctx.translate(
-                    (self.width - surface.width) / 2, (self.bar.height - surface.height) / 2
+                    (self.width - surface.width) / 2,
+                    (self.bar.height - surface.height) / 2,
```
Infinidoge, line 165: While not very relevant to the changes of this PR, black wanted it done `¯\_(ツ)_/¯`

PR #4474, opened 2023-09-25
Add `pip install[X]` for all deps and document

m-col:


elParaguayo:
Can all the stuff in `setup.cfg` be moved to `pyproject.toml`?

elParaguayo:
Oh, and great idea too!

m-col:
> Can all the stuff in setup.cfg be moved to pyproject.toml?

Most of it I think so yeah :) one thing at a time!

setup.cfg
```diff
@@ -39,6 +39,27 @@ wayland =
...
+  pytz
+  xdg
+  xmltodict
+all =
+  # All user dependencies i.e. excluding dev deps.
+  %(ipython)s
+  %(wayland)s
+  %(widgets)s
+  bowler
+  dbus-next
```
elParaguayo, line 61: This is included in `widgets`

m-col: Ah nice

m-col: Ooh the widgets one has an underscore and the test one has a dash. I guess pip normalises those when installing things

docs/manual/install/index.rst
```diff
@@ -116,6 +116,13 @@ With the dependencies in place, you can now install the stable version of qtile
 
    pip install qtile
 
+Or with sets of dependencies:
+
+.. code-block:: bash
+
+   pip install qtile[wayland]  # for Wayland dependencies
+   pip install qtile[widgets]  # for all widget dependencies
+   pip install qtile[all]      # for all user dependencies
```
jwijenbergh, line 125: Great idea!

From an outside view maybe "user" dependencies is a bit vague. Is this documented somewhere?

jwijenbergh: If it's not documented a small sentence somewhere above or in the comment itself will do

m-col: Good point, maybe just "all dependencies"?

setup.cfg
```diff
@@ -39,6 +39,26 @@ wayland =
...
   pywlroots==0.16.4
+widgets =
+  dbus-next
+  imaplib2
+  iwlib
+  keyring
+  mailbox
+  psutil
+  pulsectl
+  pulsectl_asyncio
```
elParaguayo, line 50: I think `pulsectl_asyncio` pulls in `pulsectl`

docs/manual/install/index.rst
jwijenbergh: Oh that is much less confusing

setup.cfg
jwijenbergh: It does indeed

m-col: We import pulsectl so figured it is a dep in its own right. Im not so familiar with the widget or those libs though so maybe they are always together?

Issue #4407, opened 2023-08-19
[Wayland] Screen order reversed using git master


justinesmithies:
### The issue:

I'm running on Void Linux and have both `wlroots0.16` and `wlroots0.16-devel` installed along with all the required deps installed using pip.

Qtile version: ```0.22.2.dev335+gbd507e89```

I've been running Qtile stable for a while on my dual screen setup and I obviously use kanshi too. I have it setup to make my external monitor HDMI screen 0 on the left and the laptop e-DP screen 1 on the right. Also this means Qtile draws my primary bar on screen 0 my external HDMI screen and the secondary bar on my laptop screen which is to the right of the main. This all works perfectly so I was suprised to find when I unistalled Qtile stable and installed the latest Qtile from the Git repo that the screen orders are reversed even when using exactly the same configs for both Qtile and kanshi.
The latest version makes my external HDMI screen 1 and the laptop e-DP screen 0.
IF I switch back to stable everything is all ok and when I reinstall the latest git version the screens are reversed.

My kanshi config is here:

```
profile Laptop {
	output eDP-1 position 0,0 mode 1920x1080
	#exec sh -c "$HOME/.config/yambar/scripts/yambar-start.sh"
}

profile Dock {
	output HDMI-A-1 position 0,0 mode 1920x1080
	output eDP-1 position 1920,0 mode 1920x1080
	#exec sh -c "$HOME/.config/yambar/scripts/yambar-start.sh"
}
```
My dotfiles are here but they are a little outdated due to changes but I have already rectified the changes according to the changelog locally. I've just not pushed my updated configs yet until I'm 100% happy.

 https://codeberg.org/JustineSmithies/qtile-wayland-dotfiles/src/branch/master/.config/qtile

The only errors I see in my logs are here:

```
2023-08-19 08:40:28,737 WARNING libqtile screens.py:<module>():L20 Found 2 monitor(s)
2023-08-19 08:40:29,369 WARNING libqtile hooks.py:outputs_changed():L28 Screens reconfigured
2023-08-19 08:40:29,369 WARNING libqtile hooks.py:outputs_changed():L28 Screens reconfigured
2023-08-19 08:40:30,371 WARNING libqtile hooks.py:outputs_changed():L30 Reloading config...
2023-08-19 08:40:30,388 WARNING libqtile screens.py:<module>():L20 Found 2 monitor(s)
[...]

PR #4475, opened 2023-09-25
Wayland Core: Set initial cursor and output together

jwijenbergh:
Fixes where the cursor location does not match the output focus. We now set the cursor location when we set the current initial output as well.

F i x e s a part of (sorry I have to do this due to GitHub otherwise closing issues) #4407


m-col:
Interesting! thanks

m-col (approved):

justinesmithies:
LGTM

PR #4485, opened 2023-09-29
Log backend name with incompatible widget

m-col:
rather than class object repr


elParaguayo:
Nice. Have enable automerge.

PR #4486, opened 2023-09-29
WL: set urgency hint and respond to activation requests properly

### LINKED ISSUE #4380: [wayland] Urgency doesn't work.
### The issue:

Like the title says - tested in foot and alacritty.

How to replicate:
1. open a terminal in one group
2. run `sleep 3; tput bel` in it
3. switch group

Expected result: urgency notification in GroupBox widget.

This can be solved via this patch: https://github.com/qtile/qtile/pull/3985#issuecomment-1657244951 but I've decided to create an issue instead of PR, as it was @jwijenbergh who found it and I don't want to take credit.

### Required:

m-col:
This rearranges this logic a bit for clarity. Focus functionality (per `focus_on_window_activation` is unchanges but the urgency flag is handled slightly differently, making groupbox show urgency as expected.

Fixes #4380

### DIFF
--- a/libqtile/backend/wayland/xdgwindow.py
+++ b/libqtile/backend/wayland/xdgwindow.py
@@ -194,25 +194,36 @@ class XdgWindow(Window[XdgSurface]):
             # Likely still pending, ignore this request.
             return
 
+        # For "focus", always focus the window.
         if focus_on_window_activation == "focus":
-            logger.debug("Focusing window (focus_on_window_activation='focus')")
+            logger.debug(
+                "Focusing window (focus_on_window_activation='%s')", focus_on_window_activation
+            )
             self.qtile.current_screen.set_group(self.group)
             self.group.focus(self)
+            return
 
-        elif focus_on_window_activation == "smart":
-            if not self.group.screen:
-                logger.debug("Ignoring focus request (focus_on_window_activation='smart')")
-            elif self.group.screen == self.qtile.current_screen:
-                logger.debug("Focusing window (focus_on_window_activation='smart')")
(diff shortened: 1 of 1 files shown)


ftpd:
Why "focus" and "urgent" settings are the same now? What's the reason behind this?

Note: if it's going to stay, the docs must also be updated [here](https://github.com/qtile/qtile/blob/master/docs/manual/config/index.rst?plain=1#L116-L126).

m-col:
> Why "focus" and "urgent" settings are the same now? What's the reason behind this?

m-col:
I was mistaken. Updated now to align with X11 code. @ftpd does this look good to you? :)

libqtile/backend/wayland/xdgwindow.py
```diff
@@ -194,25 +194,36 @@ def handle_activation_request(self, focus_on_window_activation: str) -> None:
...
-                logger.debug("Ignoring focus request (focus_on_window_activation='smart')")
-            elif self.group.screen == self.qtile.current_screen:
-                logger.debug("Focusing window (focus_on_window_activation='smart')")
-                self.qtile.current_screen.set_group(self.group)
-                self.group.focus(self)
-            else:
-                self._urgent = True
-                hook.fire("client_urgent_hint_changed", self)
+        # For "never", do nothing.
+        if focus_on_window_activation == "never":
```
ftpd, line 207: Isn't it against the whole idea of this PR? You started it after we talked that the urgency should be triggered every time, despite setting of `focus_on_window_activation`. In this version, when the variable is set to `never`, the function will exit and nothing will happen with the urgency - which is basically the same state we have now.

PR #4490, opened 2023-10-02
support XDG_CONFIG_DIRS

c0rydoras:
PR for #4483 

1. Checks in `${XDG_CONFIG_HOME:-~/.config}` for `qtile/config.py`
2. Checks in $XDG_CONFIG_DIRS for `qtile/config.py`
3. If not a single one was wound it returns `${XDG_CONFIG_HOME:-~/.config}/qtile/config.py` where `make_qtile` will then place the default configuration file

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -202,13 +202,15 @@ def get_cache_dir() -> str:
 
 
 def get_config_file():
-    config_home = os.getenv("XDG_CONFIG_HOME", "~/.config")
-    if (config_file := Path(f"{config_home}/qtile/config.py").expanduser()).exists():
+    config_home = Path(os.getenv("XDG_CONFIG_HOME", "~/.config")).expanduser()
+    config_file = config_home.joinpath("qtile/config.py")
+    if config_file.exists():
         return config_file
 
     xdg_config_dirs = os.getenv("XDG_CONFIG_DIRS", "/etc/xdg/").split(":")
     for config_dir in xdg_config_dirs:
-        if (system_wide_config := Path(f"{config_dir}/qtile/config.py")).expanduser().exists():
+        system_wide_config = Path(config_dir).expanduser().joinpath("qtile/config.py")
+        if system_wide_config.exists():
             return system_wide_config
 
     return config_file
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -31,6 +31,7 @@ from collections.abc import Sequence
 from random import randint
 from shutil import which
 from typing import TYPE_CHECKING
+from pathlib import Path
 
 if TYPE_CHECKING:
     from typing import Any, Callable, Coroutine, TypeVar, Union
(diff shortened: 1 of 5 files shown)


libqtile/utils.py
```diff
@@ -199,6 +200,21 @@ def get_cache_dir() -> str:
         os.makedirs(cache_directory)
     return cache_directory
 
+def _get_config_file():
+    config_home = os.getenv("XDG_CONFIG_HOME", "~/.config")
+    if (config_file:= Path(f"{config_home}/qtile/config.py")).exists():
```
m-col, line 205: Does this need the tilde resolved before the `exists`?
```suggestion
    config_home = Path(os.getenv("XDG_CONFIG_HOME", "~/.config")).expanduser()
    if (config_file:= config_home / "qtile/config.py").exists():
```

libqtile/utils.py
```diff
@@ -199,6 +200,21 @@ def get_cache_dir() -> str:
...
+        return config_file
+
+    xdg_config_dirs = os.getenv("XDG_CONFIG_DIRS", "/etc/xdg/").split(":")
+    for config_dir in xdg_config_dirs:
+        if (system_wide_config := Path(f"{config_dir}/qtile/config.py")).exists():
```
m-col, line 210: These also might need the tildes resolved in each loop. I'm basing this only on i3's code though so if you've tested already and it works then it's fine to leave it

m-col (commented):
Couple comments but this looks great, thanks!

Also make sure to update `docs/manual/config/index.rst`

libqtile/utils.py
c0rydoras: Yes it needs the tilde resolved before the `exists`

fixed here: https://github.com/qtile/qtile/pull/4490/commits/8b38ef38099d77b3bb96fb7d76e9832c7d724b4e

libqtile/utils.py
c0rydoras: This would in fact break for any directory in `$XDG_CONFIG_DIRS` with a `~`

Fixed here: https://github.com/qtile/qtile/pull/4490/commits/8b38ef38099d77b3bb96fb7d76e9832c7d724b4e

libqtile/utils.py
```diff
@@ -200,6 +201,19 @@ def get_cache_dir() -> str:
...
 
+def get_config_file():
+    config_home = os.getenv("XDG_CONFIG_HOME", "~/.config")
+    if (config_file := Path(f"{config_home}/qtile/config.py").expanduser()).exists():
+        return config_file
+
+    xdg_config_dirs = os.getenv("XDG_CONFIG_DIRS", "/etc/xdg/").split(":")
+    for config_dir in xdg_config_dirs:
+        if (system_wide_config := Path(f"{config_dir}/qtile/config.py")).expanduser().exists():
+            return system_wide_config
```
m-col, line 212: This is getting hard to read with all the parentheses, could you split out the assignments both here and on line 206? In fact, you've misplaced them, and here we could be returning a path without the resolved tilde ;)
```suggestion
    for config_dir in xdg_config_dirs:
        system_wide_config = Path(f"{config_dir}/qtile/config.py")).expanduser()
        if system_wide_config.exists():
            return system_wide_config
```

c0rydoras: done here: https://github.com/qtile/qtile/pull/4490/commits/e3a9eafe47703a71fdfb3bc0c6a60596dc64d5bf

m-col (approved):
Nice, thanks!

elParaguayo:
@c0rydoras Can you rebase this onto the latest master so we can run the test suite before merging this? Thanks.

PR #4496, opened 2023-10-06
[x11] Add pseudo-transparency

elParaguayo:
Adds a new config option, `x11_fake_transparency`, to emulate transparency in bars/widgets by copying the root window contents to the background of the bar/widget before rendering the contents on top.

Primary benefit of this is to prevent issues with the `Systray` widget which cannot currently be displayed with semi-transparent backgrounds.

Unlike the previous version, I've (largely) managed to contain the logic to the x11 backend.

Draft for now as some things to check:
- [x] Check behaviour when widgets don't fill bar
- [x] Check multiple screen behaviour (e.g. positioning)
- [x] Fix `Systray` behaviour
- [x] Handle wallpaper changes (both by `set_wallpaper` and by external commands)

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -112,7 +112,6 @@ class Core(base.Core):
             | EventMask.EnterWindow
             | EventMask.LeaveWindow
             | EventMask.ButtonPress
-            | EventMask.PropertyChange
         )
         self._root.set_attribute(eventmask=self.eventmask)
 
@@ -240,6 +239,20 @@ class Core(base.Core):
         """Assign windows to groups"""
         assert self.qtile is not None
 
+        # Check whether we need to subscribe for PropertyChange events
+        # (we use these to tell if the root pixmap has been updated which is
+        # required to update bars relying on pseudotransparency)
+        if self.qtile.config.x11_fake_transparency:
+            # Add PropertyChange to root window's event mask if it's not already there
+            if not self.eventmask & xcffib.xproto.EventMask.PropertyChange:
+                self.eventmask |= xcffib.xproto.EventMask.PropertyChange
+                self._root.set_attribute(eventmask=self.eventmask)
+        else:
+            # Remove PropertyChange from root window's event mask if it's there
+            if self.eventmask & xcffib.xproto.EventMask.PropertyChange:
+                self.eventmask &= ~xcffib.xproto.EventMask.PropertyChange
+                self._root.set_attribute(eventmask=self.eventmask)
+
         # Ensure that properties are initialised at startup
         self.update_client_lists()
 
@@ -289,7 +302,7 @@ class Core(base.Core):
 
             self.update_client_lists()
             win.change_layer()
-
+        
     def warp_pointer(self, x, y):
         self._root.warp_pointer(x, y)
         self._root.set_input_focus()
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -35,6 +35,7 @@ import xcffib.xtest
 from xcffib.xproto import EventMask
 
 from libqtile import config, hook, utils
+from libqtile import bar
 from libqtile.backend import base
 from libqtile.backend.x11 import window, xcbq
 from libqtile.backend.x11.xkeysyms import keysyms
@@ -111,6 +112,7 @@ class Core(base.Core):
             | EventMask.EnterWindow
             | EventMask.LeaveWindow
             | EventMask.ButtonPress
+            | EventMask.PropertyChange
         )
         self._root.set_attribute(eventmask=self.eventmask)
 
@@ -170,6 +172,8 @@ class Core(base.Core):
         self._last_motion_time = 0
 
         self.last_focused: base.Window | None = None
+
+        self.root_pixmap = None
 
     @property
     def name(self):
@@ -638,6 +642,20 @@ class Core(base.Core):
             self._selection[name]["selection"] = value
             hook.fire("selection_change", name, self._selection[name])
 
+        elif all(
+            (
+                name in ("ESETROOT_PMAP_ID", "_XROOTPMAP_ID"),
+                event.window == self._root.wid,
+                self.qtile.config.x11_fake_transparency,
+            )
+        ):
+            self.root_pixmap = self._get_root_pixmap()
+
+            for screen in self.qtile.screens:
+                for gap in screen.gaps:
+                    if isinstance(gap, bar.Bar):
+                        gap.draw()
+
     def handle_ClientMessage(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
 
@@ -952,3 +970,20 @@ class Core(base.Core):
             self.last_focused.change_layer()
 
         self.last_focused = win
+
+    def _get_root_pixmap(self):
+        root_win = self.conn.default_screen.root
+
+        try:
+            root_pixmap = root_win.get_property("_XROOTPMAP_ID", xcffib.xproto.Atom.PIXMAP, int)
+        except xcffib.ConnectionException:
+            root_pixmap = None
+
+        if not root_pixmap:
+            root_pixmap = root_win.get_property(
+                "ESETROOT_PMAP_ID", xcffib.xproto.Atom.PIXMAP, int
+            )
+        if root_pixmap:
+            self.root_pixmap = root_pixmap[0]
+        else:
+            self.root_pixmap = None
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -662,7 +662,7 @@ class Core(base.Core):
                 self.qtile.config.x11_fake_transparency,
             )
         ):
-            self.root_pixmap = self._get_root_pixmap()
+            self._get_root_pixmap()
 
             for screen in self.qtile.screens:
                 for gap in screen.gaps:
--- a/libqtile/backend/x11/drawer.py
+++ b/libqtile/backend/x11/drawer.py
@@ -40,6 +40,7 @@ import xcffib.xproto
 
 from libqtile import utils
 from libqtile.backend.base import drawer
+from libqtile.log_utils import logger
 
 if TYPE_CHECKING:
     from libqtile.backend.base import Internal
@@ -65,6 +66,10 @@ class Drawer(drawer.Drawer):
         )
         # Create an XCBSurface and pixmap
         self._check_xcb()
+        self.pseudotransparent = self._depth == 24 and self.qtile.config.x11_fake_transparency
+
+        self.root_pixmap = None
+        self.pseudo_pixmap = None
 
     def finalize(self):
         self._free_xcb_surface()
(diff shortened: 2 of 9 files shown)


elParaguayo:
Getting some weird behaviour with a second screen:
- Widget background is blank until redrawn
- Systray background does not work on secondary monitor

ervinpopescu:
So far, this does not fix #2935, as can be seen in the screenshot below:
image
Expected behavior:
image
Is it a problem with the decoration, as in qtile-extras should be updated too?

elParaguayo:
I'll need to fix qtile-extras separately. Is that just the systray in a rectdecoration? Can you share the config for that decoration?

ervinpopescu:
Hmm, that branch still doesn't solve it... Currently my bar bg is `#ffffff00` and the systray decoration config is:
```py
RectDecoration(
    colour=decor_bg,
    radius=10,
    extrawidth=10,
    line_width=8,
    line_colour=decor_bg,
    padding_y=-3,
    padding_x=3,
    filled=True,
)
```

elParaguayo:
Are you talking about the overlapping icon issue or an issue with rendering the decoration? I can't recreate this. Can you try disabling your compositor?

As for the bar not updating, have you pulled the latest update to this PR? I think that issue should be fixed now.

ervinpopescu:
qtile-extras on latest version (built using `makepkg -sic` in `~/.cache/yay/qtile-extras-git`)
```console
$ yay -Qi qtile-extras-git | grep Version
Version         : r291.3c00b08-1
```
```console
$ qtile --version
0.23.0b3.dev64+g8ce36674
```
The decoration renders fine, however the icons are still overlapping and I can't figure out ATM what to change for this to work.

elParaguayo:
I assume you have `x11_fake_transparency = True` in your config.

Which apps are giving the overlapping widgets?

ervinpopescu:
> I assume you have `x11_fake_transparency = True` in your config.

Yes I do.

> Which apps are giving the overlapping widgets?

All which are updating: `nm-applet`, `blueman-applet`, `mictray`, `pa-applet`
Of course `flameshot` has no problems.

> Are you able to post a link to your config so I can try the whole thing?

https://github.com/ervinpopescu/dots/tree/qtile-pseudotransparency

I need to add something about the bar not being redrawn, I was changing the wallpapers when I noticed that, so I suppose the "fake/pseudo" stands for the bar copying the part of the wallpaper which is under it to its background, correct?

elParaguayo:
Thanks for the info.

You're correct about how pseudotransparency works. We copy the wallpaper to the background of the widget.

Re changing wallpaper. That would make sense as I haven't added any code to redraw the bar in that scenario yet. Are you using the built-in `set_wallpaper` command or an external command?

I'll take a look again and see if I can recreate the overlapping icons.

ervinpopescu:
I'm running the wallpaper update externally. Is there too much overhead if the bar diffs the current root background every 5 seconds against its own background?

I suppose I should migrate to using `set_wallpaper`, I've been postponing `qtile`-related config updates for a while now. I still think the wallpaper change should be handled.

> I'll take a look again and see if I can recreate the overlapping icons.
 
Thanks!

Can't figure out how to check if the imports can fail for you, so here is a trusty command for you to check them yourself:
```bash
grep --binary-files=without-match --color=always -r --no-filename -w import | sort -h | uniq
```

elParaguayo:
You're right that the wallpaper change needs to be handled. It's trivial if using the `set_wallpaper` command (I just haven't implemented it yet) but I need to think about how to do it when it's changed by an external app. I'm reluctant to continuously poll and would prefer to see if there's some event that I can catch.

elParaguayo:
OK - the wallpaper change issue should be fixed (the code needs some tidying but it should work).

libqtile/backend/x11/core.py
```diff
@@ -111,6 +112,7 @@ def __init__(self, display_name: str | None = None) -> None:
             | EventMask.EnterWindow
             | EventMask.LeaveWindow
             | EventMask.ButtonPress
+            | EventMask.PropertyChange
```
elParaguayo, line 115: I don't love this as it means we get all `PropertyNotify` events on the root window. I wanted to only enable this if `x11_fake_transparency` had been set by `Core.setup_listener` is called before the config is loaded.

One approach could be to use a hook which is fired when the config is loaded and subscribe to that.

ervinpopescu: I dig that hook, I don't have any other ideas right now which involve subscribing to it, but it sounds cool.

elParaguayo: It's not actually necessary here after all.

m-col: You can use `Core.on_config_load` (ignore this if you found it)

libqtile/backend/x11/core.py
```diff
@@ -236,6 +239,20 @@ def on_config_load(self, initial) -> None:
...
+        if self.qtile.config.x11_fake_transparency:
+            # Add PropertyChange to root window's event mask if it's not already there
+            if not self.eventmask & xcffib.xproto.EventMask.PropertyChange:
+                self.eventmask |= xcffib.xproto.EventMask.PropertyChange
+                self._root.set_attribute(eventmask=self.eventmask)
+        else:
+            # Remove PropertyChange from root window's event mask if it's there
+            if self.eventmask & xcffib.xproto.EventMask.PropertyChange:
+                self.eventmask &= ~xcffib.xproto.EventMask.PropertyChange
+                self._root.set_attribute(eventmask=self.eventmask)
```
m-col, line 254: Is this what other projects use when they use pseudo transparency? Polybar perhaps?

libqtile/backend/x11/core.py
```diff
@@ -638,6 +655,20 @@ def handle_PropertyNotify(self, event) -> None:  # noqa: N802
...
             hook.fire("selection_change", name, self._selection[name])
 
+        elif all(
+            (
+                name in ("ESETROOT_PMAP_ID", "_XROOTPMAP_ID"),
+                event.window == self._root.wid,
+                self.qtile.config.x11_fake_transparency,
+            )
+        ):
+            self.root_pixmap = self._get_root_pixmap()
```
m-col, line 665: If only we'd typed the X backend... the return type of `self._get_root_pixmap()` is `None` ;)

m-col (commented):
Nice work :) couple comments

libqtile/backend/x11/core.py
elParaguayo: Yes - they listen for the same event. I didn't know that before I wrote the code but just checked!

elParaguayo: interestingly, polybar suggests that systray always uses pseudotransparency, even if the bar doesn't. I didn't think that was possible as you couldn't reparent a window with 24-bit depth into one with a 32-bit depth. I should look into that.

libqtile/backend/x11/core.py
elParaguayo: Interesting. I wonder why it still worked!

libqtile/backend/x11/drawer.py
```diff
@@ -235,3 +260,32 @@ def clear_rect(self, x=0, y=0, width=0, height=0):
...
+            self._win.x + offsetx,
+            self._win.y + offsety,
+        )
+
+        with self.qtile.core.masked():
+            self.qtile.core.conn.conn.core.CopyArea(
+                self.qtile.core.root_pixmap,
+                self.pseudo_pixmap,
+                self._gc,
+                *pos,
```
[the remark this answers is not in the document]
elParaguayo: That made more sense when the logic was a bit more convoluted. Definitely neater your way.

libqtile/backend/x11/core.py
m-col: Nice :) happy to have that in this PR or a follow up

libqtile/backend/x11/core.py
m-col: Honestly me too, surely it's always None?

elParaguayo: Ah. It's because the drawers check this (correctly) if the value is None.

libqtile/backend/x11/core.py
elParaguayo: I might have a play around and see if I can add it here.

elParaguayo:
No. Because I can't recreate it!

libqtile/backend/x11/core.py
elParaguayo: Had a play. Will leave this for a follow-up!

elParaguayo:
@ervinpopescu Can you confirm that you get the same overlapping issue without using widget decorations? That should be the case but wanted to check.

ervinpopescu:
Yes, running e2fc6dfd without using decorations gives the same result.

ervinpopescu:
[This minimal config](https://raw.githubusercontent.com/ervinpopescu/dots/qtile-pseudotransparency/configs/.config/qtile/xephyr_config.py) (no decorations) still gives the same overlapping icons. What should I do to troubleshoot this, given that you are unable to reproduce?

elParaguayo:
Let me try that config. Does the issue arise when you just have a single icon in the tray or does it only happen with multiple icons?

Can you also explain what you do to trigger the issue i.e. how are you making the tray icon change?

ervinpopescu:
One icon triggers it and I am changing the icon based on the state of the application I guess. For example, pressing on the `mictray` icon is muting the microphone so the icon changes to a "cut microphone".

elParaguayo:
Interesting. I get that behaviour with your config but not mine. Needs further investigating.

elParaguayo:
OK - so the first issue with that config is that it has no wallpaper and so it disables pseudotransparency as it can't find a pixmap for the root window (lots of lines in the log). If you add a wallpaper and run your config, the icons behave correctly.

However, the issue returns if you run picom...

elParaguayo:
So, the issue is that `picom` is swallowing events which means a redraw of the bar is not being called. If you disable picom, you should find that this works ok.

ervinpopescu:
I assume I need to wait for you to push the changes, right? For me, with `picom` not running, this is still appearing.

elParaguayo:
That's odd. I can reliably trigger this when running picom but it works fine without it for me but maybe that's a xephyr thing.

ervinpopescu:
Oh, okay, I misunderstood, my bad. What other application can be swallowing events? And how did you figure out that is what's happening?

elParaguayo:
I put some print statements in the parts of the widget code that trigger the bar to be redrawn. This happens when certain events are handled by the icon. Those print statements were not called once picom was running.

elParaguayo:
Can you try with the `blueman-applet` with and without picom:
1) toggle the powered state with `busctl set-property  org.bluez /org/bluez/hci0 org.bluez.Adapter1 Powered b true` and `false`
2) (exit and restart applet) toggle powered state with context menu from the icon.

Do you get any different behaviours?

ervinpopescu:
Unfortunately, with or without `picom`, the 2 situations behave the same.

elParaguayo:
Then there's still something I'm missing...

Does `qtile cmd-obj -o bar top - f eval -a "self.draw()"` clear the issue (albeit temporarily)?

ervinpopescu:
Nope, nothing is changed.

elParaguayo:
Really?! Then I've really got no idea what's causing this.

ervinpopescu:
Glad we're on the same page! 😄

jooooscha:
I am still experiencing this issue. Is there any new progress?

I would be willing to work on this, if anyone could summarize the current status.

elParaguayo:
I'm not actively working on this at the moment.

PR #4499, opened 2023-10-07
wlroots-next: chase wlroots master

m-col:
This updates the `wlroots-next` branch to catch up to wlroots' master branch as of writing.

See https://github.com/flacjacket/pywlroots/pull/129 for pywlroots `wlroots-next` branch that this depends on (generally with same commit messages).

Commits are in chronological order (of merge to wlroots master) with the wlroots commit hash and message in their message. Checking out one of the wlroots commits, as well as pywlroots and qtile wlroots-next commits with the same hash mentioned, should compile and run. Obviously sometimes there are bugs (in any of the 3 projects).

Couple things to note as current issues:

- The KDE `idle` protocol was dropped, which we used, so currently there is no idle protocol implementated in qtile.
- Commit [(wlroots 26676c8) xwm: use unified map logic](https://github.com/qtile/qtile/commit/e9d5193135dcf8a65a519ebf0aec0a90e96ba4af) broke XWayland with a segfault that raises when an X client connects. I'm not sure if this is a wlroots bug but I suggest not running this branch if this is an issue.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -60,8 +60,6 @@ from wlroots.wlr_types import (
     xdg_decoration_v1,
 )
 from wlroots.wlr_types.cursor import Cursor, WarpMode
-from wlroots.wlr_types.idle import Idle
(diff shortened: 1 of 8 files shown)


ramnes:
closing in favor of #4752

PR #4509, opened 2023-10-09
Battery widget: recognize 'Not charging' state (on linux)

TheoVanderkooy:
Linux has a ["Not charging"](https://github.com/torvalds/linux/blob/94f6f0550c625fab1f373bb86a6669b45e9748b3/drivers/power/supply/power_supply_sysfs.c#L79) battery state, which is currently not understood by the Battery widget. This change adds an extra configurable character for this case.

- As far as I can tell, `acpiconf` on FreeBSD doesn't have an equivalent state so I didn't touch the FreeBSD implementation.
- I am not entirely sure what causes the battery state to be "Not charging". On my machine I've only seen it when the battery is full or almost full. (I don't know how it decides between "Full" and "Not charging" though, I have seen both)
- I'm happy to tweak the behaviour of this case as long as I can configure this case to be distinguishable from "Unknown". `"*"` was an arbitrary choice for the default char, and I considered displaying as "Full" with short text if the battery is nearly full but decided to keep the initial PR simple.

## Testing
I've manually verified it works how I expect on my machine. I added a test case but had difficulty setting up a dev environment, so I haven't actually run that test. I assume it will be run automatically before merging.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -59,7 +59,8 @@ class BatteryState(Enum):
     DISCHARGING = 2
     FULL = 3
     EMPTY = 4
-    UNKNOWN = 5
+    NOT_CHARGING = 5
+    UNKNOWN = 6
 
 
 BatteryStatus = NamedTuple(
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks. I'd started looking at this in #3469 but we should get this fixed!

ramnes (approved):
LGTM

elParaguayo:
Ignore the failures for now - there's a broader issue with the test suite which we need to fix (pypy3 is segfaulting). Once that's done you can rebase this and run the suite.

ramnes:
Thanks!

Issue #4502, opened 2023-10-08
Segfault in CI on pypy3


elParaguayo:
### Issue description

Pinging @tych0 for his input here!

We're getting consistent test failures on pypy3. The test failures immediately after the call to run pytest.

I've tried locally and get a segfault too. From the coredump (below) it seems to be related to cairo but working this out is a bit beyond my skills so was curious to see if you had any ideas (NB I've got pypy 3.10 installed locally while CI is using 3.9. I've confirmed the same issue applies with 3.10: https://github.com/elParaguayo/qtile/actions/runs/6433434609/job/17500512776).

### Version

Latest

### Backend

X11 (default), Wayland (experimental)

### Config

```python
n/a
```

### Logs

```text
Stack trace:

Stack trace of thread 107186:
#0  0x00007efd9c0ac83c n/a (libc.so.6 + 0x8e83c)
#1  0x00007efd9c05c668 raise (libc.so.6 + 0x3e668)
#2  0x00007efd9c0444b8 abort (libc.so.6 + 0x264b8)
#3  0x00007efd9e4e536f n/a (libpypy3.10-c.so + 0x22e536f)
#4  0x00007efd9dac15d5 n/a (libpypy3.10-c.so + 0x18c15d5)
#5  0x00007efd9da443d7 n/a (libpypy3.10-c.so + 0x18443d7)
#6  0x00007efd9912bea0 n/a (/home/elp/Documents/scripts/github/qtile-dev/qtile/.tox/pypy3-x11/lib/pypy3.10/site-packages/cairo/_cairo.pypy310-pp73-x86_64-linux-gnu.so + 0xfea0)
```

Debug info
```
ELF object binary architecture: AMD x86-64

Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44	      return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
#1  0x00007efd9c0ac8a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
78	  return __pthread_kill_implementation (threadid, signo, 0);
#2  0x00007efd9c05c668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
26	  int ret = __pthread_kill (__pthread_self (), sig);
[...]

PR #4514, opened 2023-10-12
CI: allow pypy3.9 to fail

elParaguayo:
Pypy3 is segfaulting in CI (see #4502). Let's allow pypy3.9 to fail for now.

If this works, we'll need to update branch protections to remove pypy3.9 requirement too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ jobs:
     build:
         runs-on: ubuntu-22.04
         name: "python ${{ matrix.python-version }} on ${{ matrix.backend }}"
+        # pypy is segfaulting at the moment so let's allow it to fail.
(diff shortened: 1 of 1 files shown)


elParaguayo:
Hmm - so `continue-on-error` still gives you a green tick. You can only tell if it failed by looking at the detail.

ramnes:
let's just mark it optional in the branch protections?

ramnes:
(done)

elParaguayo:
Don't think that's enough. The other tests will still be cancelled without this change (or setting `fail-fast` to `false`).

ramnes:
right, then I think we should put the continue on error at the job level

ramnes:
I said "at the job level" in opposition to the step level in which it currently is :P

elParaguayo:
My mistake!

ramnes:
Thanks!

PR #4521, opened 2023-10-14
Fullscreen windows shouldn't be draggable

### LINKED ISSUE #2387: Make fullscreen windows non-draggable or restore previous state before dragging
It doesn't make sense to drag a window that's taking the whole screen size.

Thank you @drishal for the idea.

### LINKED ISSUE #2415: Previously floating fullscreen windows tiles after leaving fullscreen
# Issue description

When opening a floating window, entering fullscreen with and leaving fullscreen the window tiles instead of floating again.

For example floating window playing youtube. double click playing video and go fullscreen. after leaving fullscreen, it will be tiled

# Qtile version

0.17.0-1

# Configuration

https://github.com/strangr/dotfiles/tree/main/.config/qtile

ramnes:
Closes #2387

@m-col In that issue you said you'd prefer if we could drag fullscreen windows but restoring the window original format while doing so. I'd be happy to see that later, I'm just trying to move a little forward. I think the logical order would be to 1/ merge this 2/ to fix issue #2415 and 3/ to implement that said feature. Would you be able to live with it in the meantime? :)

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -796,7 +796,7 @@ class Qtile(CommandObject):
                         if status in (interface.ERROR, interface.EXCEPTION):
                             logger.error("Mouse command error %s: %s", i.name, val)
                         handled = True
-            elif isinstance(m, Drag):
+            elif isinstance(m, Drag) and not self.current_window.fullscreen:
                 if m.start:
                     i = m.start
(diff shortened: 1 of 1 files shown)


m-col (approved):
Yeah sounds good :)

PR #4525, opened 2023-10-17
update wayland issue label link

cjbayliss:
* update the wayland issue label link
* also don't use 'here' as the link, instead use a longer section of words for users with accesibilty needs.


ramnes:
Thanks!

PR #4527, opened 2023-10-17
Add custom hooks

elParaguayo:
Adds ability for users to create custom hooks which can be fired from external scripts.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -49,7 +49,7 @@ class Subscribe:
     def __init__(self):
         hooks = set([])
         for i in dir(self):
-            if not i.startswith("_"):
+            if not i.startswith("_") and i != "custom":
                 hooks.add(i)
         self.hooks = hooks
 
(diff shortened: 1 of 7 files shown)


elParaguayo:
I'll need to fix the docs here as it doesn't build the docstring correctly which is probably because this hook decorator takes an argument.

ramnes (approved):
nice

libqtile/hook.py
```diff
@@ -850,6 +850,53 @@ def lock_on_sleep():
...
+
+        .. code:: python
+
+          from libqtile import hook
+          from libqtile.log_utils import logger
+
+          @hook.subscribe.custom("my_custom_hook")
+          def hooked_function():
+            logger.warning("Custom hook received.")
+
```
m-col, line 872: that's a super nice api

m-col (approved):

libqtile/hook.py
elParaguayo: I actually wonder whether it would be neater to do:
```python
@hook.subscribe.custom.my_custom_hook
def hooked_function():
   ...
```

elParaguayo: Would `user` be better than `custom`?

ramnes: Both work really. Do we really need a prefix anyway?

elParaguayo: No. You don't need the prefix. I just thought it made things a bit clearer and prevented any clashes with existing hooks.

m-col: If it's all the same to you guys, I do think using a string is clearer here. Then it's clear that it's an arbitrary piece of text (and supports punctation, spaces etc) rather than an attribute, which could be interpreted as being one of a predefined set of names. Bit less magical

elParaguayo: Killing the magic :(

I'm fine to change it back!

libqtile/hook.py
```diff
@@ -45,11 +45,36 @@ def clear():
     subscriptions.clear()
 
 
+class UserHook:
```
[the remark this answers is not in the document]
elParaguayo: Not needed if we're not using dotted syntax.

libqtile/hook.py
```diff
@@ -954,6 +936,11 @@ def _fire_async_event(co):
 
 def fire(event, *args, **kwargs):
     if event not in subscribe.hooks:
+        # Suppress error for custom events. Event is only in subscribe.hooks
+        # if a custom hook has been subscribed but it should be possible to fire custom
+        # hooks even if there are no subscriptions.
```
ramnes, line 941: Why should it be possible to fire a custom event if there are no subscriptions? What's the point?

libqtile/hook.py
```diff
@@ -914,7 +887,16 @@ def hooked_function():
           However, the same socket will need to be passed wherever you run ``qtile cmd-obj`` or ``qtile shell``.
 
         """
-        return self._user
+        if hook_name in dir(self):
+            raise ValueError(f"Do not use name of existing hook for custom hooks: {hook_name}.")
+
+        def _wrapper(func):
+            name = f"user_{hook_name}"
```
ramnes, line 894: Just nitpicking: maybe we could store these hooks somewhere else (.e.g `self.user_hooks`) to avoid that little trick?

libqtile/hook.py
ramnes: sound like another topic to me, i.e. how we handle third party hooks; I feel like they shouldn't enter in the "user hooks" category

elParaguayo: ok. I'll remove for now but it is something that I'd think about adding (unless I just do some code injection with qtile-extras...)

libqtile/hook.py
elParaguayo: Not sure about that. We just have the one entry point for firing events (`hook.fire`) so you could still have a name clash even if the hooks were stored in different sets. Also, `hook.subscriptions` is a single dict with the hook name as the key. Without the prefix you could have clashes.

m-col: Won't there be a single `hook.fire` call in the codebase for user hooks, the one in `Core.fire_user_hook`? Unless there are other potential use cases that might make it confusing, we could always add a `hook.fire_user` or something to select that other set?

np either way though :)

m-col (approved):
This is neat

elParaguayo:
I'd prefer to drop this in favour of #4551 which has been updated to include this functionality.

libqtile/hook.py
elParaguayo: My expectation with the user hooks is that they're fired by external apps and so need to be fired via the command interface.

PR #4531, opened 2023-10-20
Choose column for new windows in Columns layout

elParaguayo:
Adds ability to set which column receives new windows in `Columns` layout.

Based on this post: https://www.reddit.com/r/qtile/comments/17bice5/is_there_a_way_to_get_layouts_for_lefties/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
       - Add `ScreenSplit` layout which allows multiple layouts per screen. Also adds `ScreenSplit`
         widget to display name of active split.
       - Updated `Bluetooth` widget which allows users to manage multiple devices in a single widget
+      - Add `align` option to `Columns` layout so new windows can be added to left or right column.
     * bugfixes
 
 Qtile 0.23.0, released 2023-09-24:
--- a/libqtile/layout/columns.py
+++ b/libqtile/layout/columns.py
@@ -17,6 +17,7 @@
 # SOFTWARE.
 from __future__ import annotations
 
+import contextlib
 from typing import TYPE_CHECKING
 
 from libqtile.command.base import expose_command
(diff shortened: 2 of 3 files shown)


ramnes:
Why specify a column number in particular rather than keeping consistent with Monad layouts and using an `align` parameter to give a direction? It seems to me that the latter would be give a better experience, especially when `num_columns` is more than 2.

elParaguayo:
How would that work if you had, say, 4 columns and wanted to add windows to the second one? It's a stupid example but I don't think align addresses it.

Happy to be proved wrong.

elParaguayo:
I guess aligning allows us to do first and last columns which may be enough here.

ramnes:
> How would that work if you had, say, 4 columns and wanted to add windows to the second one? It's a stupid example but I don't think align addresses it.

Why would you want that?

elParaguayo:
> > How would that work if you had, say, 4 columns and wanted to add windows to the second one? It's a stupid example but I don't think align addresses it.
> 
> Why would you want that?

Never underestimate your users :D

Joking aside, it is very unlikely you'd ever want that. I'll switch this to `align`.

ramnes:
Just tried the branch and I think the behavior is... weird?
* Window 1 opens fullscreen
* Window 2 opens on its right and gets focus
* Window 3 opens on the left column, even if the focus is on the right column

If we want a "mirrored" layout, the correct behavior would be that:
* Window 1 opens fullscreen
* Window 2 opens on its left and gets focus
* Window 3 opens on the current column (left)

ramnes:
I think a simpler diff like this works better :)

```diff
@@ -239,7 +254,8 @@ class Columns(Layout):
     def add_client(self, client: Window) -> None:
         c = self.cc
         if len(c) > 0 and len(self.columns) < self.num_columns:
-            c = self.add_column()
+            prepend = (self.align is Columns._left)
+            c = self.add_column(prepend=prepend)
         if self.fair:
             least = min(self.columns, key=len)
             if len(least) < len(c):

```

elParaguayo:
Yes. All makes sense. I'll update when I get a chance.

ramnes:
I can push the changes if you want

elParaguayo:
Sure. Go ahead. I'm traveling today so won't have access to my laptop until tomorrow and I may be too tired at that point as well!

PR #4532, opened 2023-10-20
add default config for switching to VTs under wayland

cjbayliss:


cjbayliss:
Things I've tried to get this PR to pass CI:

 * using `==` instead of `is` (still failed)
 * doing a `if qtile.core is not None` check (still failed)

I barely know python, so can't figure this out,

bbb651:
Comparing strings with `is` is definitely bogus, `is` compares the identity of the object, basically the location in memory, rather than `==` which checks equality, which for complicated python string literal internal reasons™ happens to return `True` and make the X11 test fail since `lazy.core.change_vt` only exists on wayland. I think changing is to `==` the code should work. Also you should use `list.append` rather than `list.extend` with a new list, it's both more efficient and more readable.

libqtile/resources/default_config.py
```diff
@@ -80,6 +80,12 @@
     Key([mod], "r", lazy.spawncmd(), desc="Spawn a command using a prompt widget"),
 ]
 
+if qtile.core.name is "wayland":
+    for i in range(1,7):
+        keys.extend([
+            Key(["control", "mod1"], f"f{i}", lazy.core.change_vt(i), desc=f"Switch to VT{i}")
+        ])
```
m-col, line 87: Sadly when the default config is imported, `libqtile.qtile` is `None`, so we can't have backend checks in the default config.

One option would be to comment it out, but I suspect we specifically want this one to be enabled.

Another option would be to enable the binding even on X, as it would be a no-op anyway. In which case, we'd need to be careful not to override the existing VT-switching bindings of the X server.

I suggest we do add bindings to any backend, but we change the keys:
```suggestion
for vt in range(1, 7):
    keys.append(
	Key(["mod1"], f"f{vt}", lazy.core.change_vt(vt), desc=f"Switch to VT{i}")
    )
```

elParaguayo: Could we extend the `when()` method for lazy calls to only run on specific backends?

m-col: Interesting. Yeah, that could be cool, and might nicely give when() some more visibility. +1 for that

elParaguayo: It wouldn't stop the key being bound unless we added code to check if the backend had been specified and remove the binding when the backend doesn't match.

m-col: Isn't the key replayed if the when() failed?

ramnes: This shouldn't be a problem anymore since we merged #4544, right? If so, let's merge it as is to relieve @cjbayliss and open another PR if we want to change the code to something we consider better (e.g. using the new `when()` arguments).

ramnes:
@cjbayliss I tried to push to your branch to see if the tests now pass but didn't have the permission. So... please rebase on master. :)

cjbayliss:
I’m not doing well health wise, so can only use my phone right now. If you need more, please close this PR and create a new one. Feel free to base it on the changes I made here. I don’t need accreditation for validation. 🙂

elParaguayo:
Sorry to hear that. Look after yourself and I hope you're better soon.

Issue #856, opened 2016-01-28
toggle_minimize in widget.TaskList


Sqozz:
Hi all,

is it possible to catch a click event from an item in the TaskList so I can call toggle_minimize() on that window? I'd like to minimize and restore unneeded windows by a simple click onto their TaskList entries.
My current approach is a client_focus event, which gets called just fine if I activate a "hidden" window. But if I click on an already active window in my TaskList I only get a client_focus event from a strange target (Internal event: client_focus((Internal(<no name>, 2097231),), {}))

flacjacket:
I haven't looked into this too deeply, but it might be possible to subclass TaskList and override some functions to do what you want

Sqozz:
I'll look into that and report if I have any success.

andering:
Hello i wanted address this by creating hook but than i realized that i don't really understand hook/event concept in the qtile, i would expect that there would be event, and its default behavior, and when i hook something on the event default behavior would get disabled or i would have ability to disable it.

So i only add toggle_minimalize() as behaviour in case i press on active window.
#923

flacjacket:
This is closed by #923 as the default behavior

PR #4536, opened 2023-10-21
feature: kill window via tasklist widget by using middle mouse click

graves501:
Hi, 

since I'm fairly new to qtile and not that experienced in Python, I couldn't figure out how to kill a window via the tasklist widget without directly changing the `tasklist.py` file itself.

I'm using the middle mouse button to kill the clicked window, since I was able to implement that quite easily in AwesomeWM.

Thanks for your efforts, so far I'm really liking qtile!

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -188,7 +188,7 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
         if self.spacing is None:
             self.spacing = self.margin_x
 
-        self.add_callbacks({"Button1": self.select_window})
+        self.add_callbacks({"Button1": self.select_window, "Button2": self.kill_window})
 
     def box_width(self, text):
         """
@@ -450,6 +450,11 @@ class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
                     window.bring_to_front()
             else:
                 window.toggle_minimize()
+
+    def kill_window(self):
+        if self.clicked:
+            window = self.clicked
+            window.kill()
 
     def _get_class_icon(self, window):
         if not getattr(window, "icons", False):


ramnes:
Hey there,

This PR is not needed for what you want to achieve, you can give `mouse_callbacks={...}` to any widget. :)

Thanks for the kind words!

graves501:
@ramnes

Hi, thanks for the response.

I am aware of the `mouse_callbacks`, but without changing the widget's implementation, how are you supposed to pass the clicked element of the widget to the callback? 

If you look into the code, I did pass a function to the `mouse_callbacks`, but from within the `tasklist` widget.

Why?

`get_clicked` is NOT exposed. 

In the callback function passed to the widget I would need to be able to do something like:

```python
c = CommandClient()
clicked = c.navigate("widget", "tasklist").call("get_clicked")
clicked.kill()
```

E.g. I want to close the Chromium window, but the focused window is Firefox.

image

There is even a [Pull Request](https://github.com/qtile/qtile/issues/856) from 2016 regarding these click events, here it was about using `window.toggle_minimize()`, in my case it would simply be `window.kill()`. 

The solution for the 2016 issue was to modify the internals of `tasklist`.

elParaguayo:
```python
def tasklist_kill(widget):
    window = widget.clicked
    if window is not None:
        window.kill()

# Then add this to your mouse callbacks:
mouse_callbacks =  {"Button2": lazy.widget["tasklist"].function(tasklist_kill)}
```

graves501:
@elParaguayo 

Thank you so much for taking the time to reply! Very much appreciated.

I agree, it doesn't make sense to make this kind of behaviour the default, since this is just my personal preference.

I apologize for opening this "discussion" - this should've been a question on Discord/IRC instead. I did try to find a solution in the documentation, but clearly failed - I'm still new to the whole qtile "eco-system".

Thanks once again - at least there is a chance that this PR might serve/help someone in the future.

elParaguayo:
No need to apologise. Thanks for taking the time to learn and want to make qtile better. 

Feel free to keep asking questions etc as needed.

PR #4537, opened 2023-10-23
Fix extra client_mouse_enter events

QBobWatson:
The Wayland core was emitting client_mouse_enter event on every mouse motion over a non-focused client.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -984,21 +984,15 @@ class Core(base.Core, wlrq.HasListeners):
                     self.seat.pointer_notify_clear_focus()
 
             if win is not self.qtile.current_window:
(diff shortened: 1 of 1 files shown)


jwijenbergh (changes requested):
Thanks for this! I would prefer just having this code moved above the if/else check https://github.com/qtile/qtile/blob/2347e18663fff2c2160272591d0a8d1838a957f8/libqtile/backend/wayland/core.py#L988-L992

QBobWatson:
Here you go.  I think it's more readable like this.

m-col (approved):
Awesome thanks! I really got test more with `follow_focus_mouse = False`

jwijenbergh (approved):
perfect, so clean now!

PR #4539, opened 2023-10-24
Allow bar to float above windows

elParaguayo:
Adds `floating` parameter to `Bar` to allow the bar to float above tiled windows. Useful when users want to have bar visibility toggled with a keypress but don't want their windows resized.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -184,6 +184,11 @@ class Bar(Gap, configurable.Configurable, CommandObject):
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
         ("border_color", "#000000", "Border colour as str or list of str [N E S W]"),
         ("border_width", 0, "Width of border as int of list of ints [N E S W]"),
+        (
+            "reserve",
+            True,
+            "Reserve screen space (when set to 'False', bar will be drawn above windows).",
+        ),
     ]
 
     def __init__(self, widgets: list[_Widget], size: int, **config: Any) -> None:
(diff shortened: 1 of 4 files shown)


libqtile/config.py
```diff
@@ -490,28 +490,34 @@ def gaps(self) -> Iterable[BarType]:
 
     @property
     def dx(self) -> int:
-        return self.x + self.left.size if self.left else self.x
+        return (
+            self.x + self.left.size
+            if (self.left and not getattr(self.left, "floating"))
+            else self.x
+        )
```
ramnes, line 497: Why using getattr here?

Also it's most likely easier to read without a ternary:

```suggestion
        if self.left and not self.left.floating:
            return self.x + self.left.size
        return self.x
```

ramnes (approved):
nice

libqtile/config.py
elParaguayo: Used getattr because the attribute doesn't exist for `Gap` objects.

Agreed on tidier code.

libqtile/bar.py
```diff
@@ -184,6 +184,7 @@ class Bar(Gap, configurable.Configurable, CommandObject):
         ("margin", 0, "Space around bar as int or list of ints [N E S W]."),
         ("border_color", "#000000", "Border colour as str or list of str [N E S W]"),
         ("border_width", 0, "Width of border as int of list of ints [N E S W]"),
+        ("floating", False, "Don't reserve screen space (bar will be drawn above windows)."),
```
m-col, line 187: I think the words "floating" and "above" here are confusing because they already have specific uses in qtile. Could we use something like `"dont_reserve"` and "Don't reserve screen space for the bar in tiled layouts"?

m-col (approved):

libqtile/bar.py
elParaguayo: Ha. That's exactly what I want for with my first version of this and for the reasons you said. I went with floating because it felt neater but happy to change it.

ramnes:
Thanks!

Issue #4096, opened 2022-12-28
Terminal doesn't scroll up automatically


MelihDarcanxyz:
### The issue:

```
> qtile --version
> 0.22.1

> cat ~/.local/share/qtile/qtile.log
> 2022-12-28 23:44:55,609 WARNING libqtile bar.py:_configure_widget():L315 Widget removed: Systray does not support <libqtile.backend.wayland.core.Core object at 0x7f358b702140>.
> 2022-12-28 23:44:55,721 WARNING libqtile bar.py:_configure_widget():L315 Widget removed: Systray does not support <libqtile.backend.wayland.core.Core object at 0x7f358b702140>.
```

Tried to copy some text on ```foot``` terminal. The text was long enough so I need the terminal to scroll up when my cursor was near enough to the top of the terminal (or above it) while highlighting the text. It doesn't happen.

I'm on wayland.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

MelihDarcanxyz:
Update: It scrolls up while it's on the top border. If I go above it, I'm not able to scroll up if I put the cursor back on the top border.

MelihDarcanxyz:
Another update: Brave works just fine.

jwijenbergh:
> Another update: Brave works just fine.

Is this for #4095 instead?

MelihDarcanxyz:
> > Another update: Brave works just fine.
> 
> Is this for #4095 instead?

Yes, sorry for that.

jwijenbergh:
> > > Another update: Brave works just fine.
> > 
> > 
> > Is this for #4095 instead?
> 
> Yes, sorry for that.

No worries, just checking :)

m-col:
Thanks for the report. Just to make sure we're talking about the same thing is the issue that, unlike in X, if you hold down a mouse button over a window then drag it out, the window should continue to receive movement input? Many clients use this functionality e.g. terminals for scrolling, text editors and browsers for highlighting content (button 1) or even an auto-scroll feature like firefox's middle click. Is that right?

jwijenbergh:
What about #3985?

jwijenbergh:
> This issue is stale because it has been open 90 days with no activity. Remove the `status: stale` label or comment, or this will be closed in 30 days.

I hate you with a passion

MelihDarcanxyz:
It's my turn to hate this bot with passion.

jwijenbergh:
[...]

PR #4540, opened 2023-10-24
Add implicit pointer grabs in Wayland.

QBobWatson:
A Wayland client expects to receive pointer motion events from the time a button is pressed until it is released, even if the pointer leaves the client's surface while the button is held.  This is crucial for scrollbar functionality: a scrollbar should continue to drag even if the pointer leaves the window.  This does not currently work in qtile.  For instance, in Chrome, if you drag the scrollbar, move the mouse off the Chrome window, release the button, then move the pointer back onto the Chrome window, Chrome will resume scrolling without any button press.

In X11, this functionality is implemented on the client side using XGrabPointer or XIGrabDevice.  Wayland provides no such mechanism: this behavior must be implemented in the compositor.  This does not seem to be well documented.  However, if you look in gdk_wayland_device_grab in the gtk source, you'll see that grabbing the pointer in GDK is basically a no-op: it does some internal accounting and changes the cursor, then simply assumes the pointer is already grabbed.

Implicit grabs are implemented in Sway (see
sway/sway/input/seatop_{default,down}.c), although that's not what they're called.  In short, pressing a button on a client surface switches the cursor handling code into "down" mode, which short-cuts most of the WM pointer processing logic and sends events directly to the client, until the button is released.  (This is part of a more general mechanism in Sway: pointer events can be processed in several modes, e.g. for moving floating windows.)

This commit implements basic implicit grabs in qtile Wayland, following what Sway does (functionally, not structurally).  Pressing a button creates an ImplicitGrab object in the Core; releasing the button destroys it.  While the ImplicitGrab object is alive, all pointer events (button, {absolute_,}motion, axis) are sent directly to wlroots, bypassing qtile's event processing.  This seems to produce the desired UI behavior.

Implicit grabs should also be implemented for touch and tablet tool events, if and when those are added to qtile.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -108,7 +108,7 @@ if TYPE_CHECKING:
     from libqtile.core.manager import Qtile
 
 
-class ImplicitGrab:
+class ImplicitGrab(wlrq.HasListeners):
     """Keep track of an implicit pointer grab.
 
     A Wayland client expects to receive pointer events from the moment a
@@ -127,15 +127,12 @@ class ImplicitGrab:
     ) -> None:
         self.core = core
         self.surface = surface
-        self.start_x = start_x
-        self.start_y = start_y
-        self.start_sx = start_sx
-        self.start_sy = start_sy
-        self._listener = pywayland.server.Listener(self._on_destroy)
-        surface.destroy_event.add(self._listener)
+        self.start_dx = start_sx - start_x
+        self.start_dy = start_sy - start_y
+        self.add_listener(surface.destroy_event, self._on_destroy)
 
     def finalize(self) -> None:
-        self._listener.remove()
+        self.finalize_listeners()
 
     def _on_destroy(self, _listener: Listener, _data: Any) -> None:
         self.core._release_implicit_grab()
@@ -648,17 +645,14 @@ class Core(base.Core, wlrq.HasListeners):
 
         if not handled:
             if self._pressed_button_count == 1 and found:
-                assert self._implicit_grab is None
                 win, surface, sx, sy = found
                 if surface:
                     self._create_implicit_grab(event.time_msec, surface, sx, sy)
             self.seat.pointer_notify_button(event.time_msec, event.button, event.button_state)
 
-    def _implicit_grab_motion(self, time: int):
-        moved_x = self.cursor.x - self._implicit_grab.start_x
-        moved_y = self.cursor.y - self._implicit_grab.start_y
-        sx = moved_x + self._implicit_grab.start_sx
-        sy = moved_y + self._implicit_grab.start_sy
+    def _implicit_grab_motion(self, time: int) -> None:
+        sx = self.cursor.x + self._implicit_grab.start_dx
+        sy = self.cursor.y + self._implicit_grab.start_dy
         self.seat.pointer_notify_motion(time, sx, sy)
 
     def _on_cursor_motion(self, _listener: Listener, event: pointer.PointerMotionEvent) -> None:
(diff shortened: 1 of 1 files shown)


QBobWatson:
> Thanks a lot for this and the detailed explanation, I will test and review this soon. So I guess this will fix #4096?

Yes, I believe so -- at least, now I can select text as described in the issue.

libqtile/backend/wayland/core.py
```diff
@@ -587,8 +647,20 @@ def _on_cursor_button(self, _listener: Listener, event: pointer.PointerButtonEve
             handled = self._process_cursor_button(button, pressed)
 
         if not handled:
+            if self._pressed_button_count == 1 and found:
+                assert self._implicit_grab is None
```
m-col, line 651: Was this needed? Im surprised mypy wouldn't be able to infer this must be the case beyond the if statement on line 636

libqtile/backend/wayland/core.py
```diff
@@ -108,6 +108,39 @@
     from libqtile.core.manager import Qtile
 
 
+class ImplicitGrab:
```
m-col, line 111: Might be worth using `HasListeners` just to follow the style of other classes that uses listeners?

libqtile/backend/wayland/core.py
```diff
@@ -108,6 +108,39 @@
...
+        start_y: int,
+        start_sx: int,
+        start_sy: int,
+    ) -> None:
+        self.core = core
+        self.surface = surface
+        self.start_x = start_x
+        self.start_y = start_y
+        self.start_sx = start_sx
+        self.start_sy = start_sy
```
m-col, line 133: It looks like the two x values and two y values are only used together, so maybe we could collapse them. E.g.:
```suggestion
        self.start_dx = start_sx - start_x
        self.start_dy = start_sy - start_y
```
Then just use that one below in `_implicit_grab_motion`

libqtile/backend/wayland/core.py
```diff
@@ -587,8 +647,20 @@ def _on_cursor_button(self, _listener: Listener, event: pointer.PointerButtonEve
...
 
         if not handled:
+            if self._pressed_button_count == 1 and found:
+                assert self._implicit_grab is None
+                win, surface, sx, sy = found
+                if surface:
+                    self._create_implicit_grab(event.time_msec, surface, sx, sy)
             self.seat.pointer_notify_button(event.time_msec, event.button, event.button_state)
 
+    def _implicit_grab_motion(self, time: int):
```
m-col, line 657: This may need a `-> None` return annotation

m-col (commented):
This is really excellent work, thank you for this!

I've made a couple comments but overall this looks great, super clear and understandable.

libqtile/backend/wayland/core.py
QBobWatson: Good point.  I originally had that line above line 636, but now it's automatic.

libqtile/backend/wayland/core.py
QBobWatson: Sure, if you prefer.

libqtile/backend/wayland/core.py
QBobWatson: Good point!  I was just copying the computation from Sway.

libqtile/backend/wayland/core.py
QBobWatson: Done.  I'll push another commit shortly.

m-col (approved):
Works like a charm. Much appreciated :)

ftpd:
Can please someone re-check that? It broke the ability to move tabs right/left in Firefox using the mouse, I'm also unable to move tickets on kanban board inside Jira ;-)
@m-col I'm tagging you directly as you were the one who merged this. Should I open a new issue?

PR #4541, opened 2023-10-24
Allow XStatic configure requests.

QBobWatson:
In the X11 backend, static windows are allowed to change their position and size, unless those parameters were set explicitly in Window.static().  This commit implements the same behavior for XStatic windows in the Wayland backend.  This fixes gkrellm.

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -732,6 +732,12 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
         self.defunct = True
         if self.group:
             self.group.remove(self)
+
+        conf_x = x
+        conf_y = y
+        conf_width = width
+        conf_height = height
+
         if x is None:
             x = self.x + self.borderwidth
         if y is None:
@@ -748,7 +754,7 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
             for rect in self._borders.pop():
                 rect.node.destroy()
 
-        win = self._to_static()
+        win = self._to_static(conf_x, conf_y, conf_width, conf_height)
 
         # Pass ownership of the foreign toplevel handle to the static window.
         if self.ftm_handle:
@@ -767,7 +773,7 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
         return self.container.node.enabled
 
     @abc.abstractmethod
-    def _to_static(self) -> Static:
+    def _to_static(self, x: int | None, y: int | None, width: int | None, height: int | None) -> Static:
         # This must return a new `Static` subclass instance
         pass
 
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -733,6 +733,8 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
         if self.group:
             self.group.remove(self)
 
+        # Keep track of user-specified geometry to support X11.
+        # Respect configure requests only if these are `None` here.
         conf_x = x
         conf_y = y
         conf_width = width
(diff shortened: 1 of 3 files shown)


libqtile/backend/wayland/xwindow.py
```diff
@@ -393,10 +406,20 @@ def _on_unmap(self, _listener: Listener, _data: Any) -> None:
         self.core.pending_windows.add(win)
 
     def _on_commit(self, _listener: Listener, _data: Any) -> None:
-        logger.debug("Signal: xstatic commit")
+        pass
+        # logger.debug("Signal: xstatic commit")
```
m-col, line 410: Should we just drop this method and its listener?

libqtile/backend/wayland/xwindow.py
```diff
@@ -393,10 +406,20 @@ def _on_unmap(self, _listener: Listener, _data: Any) -> None:
...
 
     def _on_request_configure(self, _listener: Listener, event: SurfaceConfigureEvent) -> None:
         logger.debug("Signal: xstatic request_configure")
+        cw = ConfigWindow
```
m-col, line 414: The wayland backend does not depend on xcffib, and we shouldn't make it. If we just need these 4 constants, could we just lift them and stick them into this file?

libqtile/backend/wayland/xwindow.py
```diff
@@ -351,13 +355,22 @@ def __init__(
...
         surface = win.surface
         Static.__init__(
             self, core, qtile, surface, win.wid, idle_inhibitor_count=idle_inhibitor_count
         )
         self._wm_class = surface.wm_class
 
+        self.conf_x = x
+        self.conf_y = y
+        self.conf_width = width
+        self.conf_height = height
```
m-col, line 372: If these are only used by `self`, could we make them private attributes?

libqtile/backend/wayland/window.py
```diff
@@ -732,6 +732,12 @@ def static(
         self.defunct = True
         if self.group:
             self.group.remove(self)
+
+        conf_x = x
```
m-col, line 736: just to help our future selves
```suggestion
        # Keep track of user-specified geometry to support X11
        # configure requests only if these are `None` here
        conf_x = x
```

m-col (changes requested):
Nice, thanks! Couple suggestions

libqtile/backend/wayland/xwindow.py
QBobWatson: Fine with me.  Each instance of gkrellm was producing 10 lines per second in the log, which is why I had it commented.

libqtile/backend/wayland/xwindow.py
QBobWatson: Yes, that's easy enough.  They're just 1, 2, 4, and 8, respectively :)

libqtile/backend/wayland/xwindow.py
QBobWatson: We could, but they're public in x11.window.Static.  It's up to you.

m-col: Yep let's go private. Really the x11 ones ought to be too.

libqtile/backend/wayland/xwindow.py
```diff
@@ -336,8 +336,21 @@ def static(
...
+
+class ConfigWindow:
+    """The XCB_CONFIG_WINDOW_* constants.
+
+    Reproduced here to remove a dependency on xcffib.
+    """
+    X = 1
+    Y = 2
+    Width = 4
+    Height = 8
```
m-col, line 353: Nice, thanks :+1:

libqtile/backend/wayland/xwindow.py
QBobWatson: Done!

m-col (approved):
Thanks

PR #4544, opened 2023-10-25
Add boolean and func options to .when()

elParaguayo:
We can't use `qtile.core.name` in the default config so one way to have backend specific keybindings is to add a `backend=` option to the `.when` method for lazy calls.

### DIFF
--- a/libqtile/__init__.py
+++ b/libqtile/__init__.py
@@ -20,15 +20,15 @@
 # SOFTWARE.
 
 
-class UndefinedCore:
+class _UndefinedCore:
     name = None
 
 
-class UndefinedQtile:
-    core = UndefinedCore()
+class _UndefinedQtile:
+    core = _UndefinedCore()
 
 
-qtile = UndefinedQtile()
+qtile = _UndefinedQtile()
 
 
 def init(q):
--- a/libqtile/__init__.py
+++ b/libqtile/__init__.py
@@ -20,7 +20,15 @@
 # SOFTWARE.
 
 
-qtile = None
+class UndefinedCore:
+    name = None
+
+
+class UndefinedQtile:
+    core = UndefinedCore()
+
+
+qtile = UndefinedQtile()
 
 
 def init(q):
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -24,9 +24,10 @@ from typing import TYPE_CHECKING
 from libqtile.command.client import InteractiveCommandClient
 from libqtile.command.graph import CommandGraphCall, CommandGraphNode
 from libqtile.command.interface import CommandInterface
+from libqtile.log_utils import logger
 
 if TYPE_CHECKING:
-    from typing import Iterable
+    from typing import Callable, Iterable
 
     from libqtile.command.graph import SelectorType
     from libqtile.config import Match
@@ -53,6 +54,8 @@ class LazyCall:
         self._if_no_focused: bool = False
         self._layouts: set[str] = set()
         self._when_floating = True
+        self._check: bool | None = None
+        self._func: Callable | None = None
 
     def __call__(self, *args, **kwargs):
         """Convenience method to allow users to pass arguments to
@@ -93,14 +96,22 @@ class LazyCall:
 
     def when(
         self,
+        check: bool | None = None,
+        *,
         focused: Match | None = None,
         if_no_focused: bool = False,
         layout: Iterable[str] | str | None = None,
         when_floating: bool = True,
+        func: Callable | None = None
     ) -> "LazyCall":
-        """Enable call only for given layout(s) and floating state
+        """Enable call only for matching criteria.
 
-        Parameters
+        Positional parameters
+        ---------------------
+        check: a boolean value to determine whether the lazy object should
+            be run.
+
+        Keyword parameters
         ----------
         focused: Match or None
             Match criteria to enable call for the current window.
@@ -117,10 +128,15 @@ class LazyCall:
             If None, enable the call for all layouts.
         when_floating: bool
             Enable call when the current window is floating.
+        func: callable
+            Enable call when the result of the callable evaluates to True
         """
         self._focused = focused
 
         self._if_no_focused = if_no_focused
+
+        self._check = check
+        self._func = func
 
         if layout is not None:
             self._layouts = {layout} if isinstance(layout, str) else set(layout)
@@ -130,6 +146,9 @@ class LazyCall:
 
     def check(self, q) -> bool:
         cur_win_floating = q.current_window and q.current_window.floating
+
+        if self._check is False:
+            return False
 
         if self._focused:
             if q.current_window and not self._focused.compare(q.current_window):
@@ -143,6 +162,16 @@ class LazyCall:
 
         if self._layouts and q.current_layout.name not in self._layouts:
             return False
+
+        if self._func:
+            try:
+                result = self._func()
+            except Exception:
+                logger.exception("Error when running function in lazy call. Ignoring.")
+                result = True
+
+            if not result:
+                return False
 
         return True
 
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -21,9 +21,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from libqtile.backend import CORES
 from libqtile.command.client import InteractiveCommandClient
 from libqtile.command.graph import CommandGraphCall, CommandGraphNode
 from libqtile.command.interface import CommandInterface
+from libqtile.confreader import ConfigError
 
 if TYPE_CHECKING:
     from typing import Iterable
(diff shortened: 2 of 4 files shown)


m-col (approved):
Lovely stuff

ramnes:
> We can't use `qtile.core.name` in the default config

Can't we fix this instead? Core is defined by the CLI, we probably can find a way to have it available when the config gets evaluated I guess?

elParaguayo:
It's fine for users' configs but there's an issue with the default config being loaded early. @m-col would have a better idea about whether this can be changed.

My other issue with the `qtile.core.name == "wayland"` test is that, if people check their config with `python config.py`, it's going to fail as `qtile` will be `None`.

elParaguayo:
I don't _think_ that would solve the issue. All you're doing is making `qtile.core.name` equal to `None` rather than `qtile`. 

My understanding of the problem is that the default config is loaded before the backend so, if you're trying to extend `keys` with some wayland-specific bindings, this doesn't help as `qtile.core.name` will be `None`.

I'll give it a try to check my understanding.

elParaguayo:
So my assumption is correct. The backend name is not available when the default config is loaded.

So, we have a couple of options:
* Keep this PR
* Find a way to load default config after the backend has started. I don't know this part of the codebase as well.

jwijenbergh:
> So my assumption is correct. The backend name is not available when the default config is loaded.
> 
> So, we have a couple of options:
> 
> * Keep this PR
> * Find a way to load default config after the backend has started. I don't know this part of the codebase as well.

Imo I think this is an amazing indeed regardless the NoneType issue. It makes it easier to make some keybindings backend specific without going to the if, extend keys dance

jwijenbergh (approved):

ramnes:
@elParaguayo my idea for your issue regarding `qtile.core.name` still being `None` in the default config was that we should now be able to do stuff like this:

```diff
diff --git a/libqtile/scripts/start.py b/libqtile/scripts/start.py
index 886d32a7..5a3f2248 100644
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -29,7 +29,7 @@ from sys import exit
 from typing import TYPE_CHECKING

 import libqtile.backend
-from libqtile import confreader
+from libqtile import confreader, qtile
 from libqtile.utils import get_config_file
 from libqtile.log_utils import logger

@@ -55,6 +55,8 @@ def rename_process():


 def make_qtile(options) -> Qtile | None:
+    qtile.core.name = options.backend
+
     if missing_deps := libqtile.backend.has_deps(options.backend):
         print(f"Backend '{options.backend}' missing required Python dependencies:")
         for dep in missing_deps:
```

Didn't try it though.

I don't mind if we merge it as is, feel free to ignore me; I'm just trying to push for the generic solution that IMO should already work in the first place, rather than adding yet another very niche functionality and its very specific extra lines of codes. As you may have gathered from our recent discussions, I'm in the mood to delete code, not to add more. :)

m-col:
> What about something in those lines? Would that fix both?
> 
> ```diff
> diff --git a/libqtile/__init__.py b/libqtile/__init__.py
> index 63a61df0..8d9210cc 100644
> --- a/libqtile/__init__.py
> +++ b/libqtile/__init__.py
> @@ -19,8 +19,15 @@
>  # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
>  # SOFTWARE.
> 
> +class UndefinedCore:
> +    name = None
> 
> -qtile = None
> +
> +class UndefinedQtile:
> +    core = UndefinedCore()
> +
> +
> +qtile = UndefinedQtile()
> ```

Looks like this might work without setting it to not None. In the loading of the default config, the keys arent defined but that might be good actually. These bindings are for new users who get stuck, in which case theyve likely just had the default config copied into place and loaded as their user config, at which point the keys are added. The default config's just no-op which is good because they are redundant.

elParaguayo:
@ramnes - yes, that works. I wonder if there's actually an argument to include both solutions in the code. Your fix is nice and aligns with what we say in the docs but the `.when()` option allows us to adjust bindings directly rather than conditionally append to lists.

e.g. compare:
```python
keys = [
    Key(
        [mod],
        "x",
        lazy.spawn("some wayland app").when(backend="wayland"),
        lazy.spawn("x11 equivalent").when(backend="x11"))
]
```

```python
if qtile.core.name == "wayland":
    keys.append(Key([mod], "x", lazy.spawn("some wayland app"))
else:
    keys.append(Key([mod], "x", lazy.spawn("x11 equivalent"))
```

Of course, people are unlikely to be switching regularly between backends...

elParaguayo:
@m-col - do we just want to go with ramnes's idea? I don't really care but it would be good to get a solution merged.

ramnes:
`when` could just take any function, that would be both generic and elegant in that scenario

ramnes:
e.g.

```python
is_wayland = functools.partial(operator.eq, qtile.core.name, "wayland")
is_x11 = functools.partial(operator.eq, qtile.core.name, "x11")

keys = [
    Key(
        [mod],
        "x",
        lazy.spawn("some wayland app").when(is_wayland),
        lazy.spawn("x11 equivalent").when(is_x11)
    )
]
```

and as we can see the function is not even useful here, so we probably want to also support just a boolean value as well, which makes sense since the config is written in Python and can execute stuff

```python
keys = [
    Key(
        [mod],
        "x",
        lazy.spawn("some wayland app").when(qtile.core.name == "wayland"),
        lazy.spawn("x11 equivalent").when(qtile.core.name == "x11")
    )
]
```

actually we could probably refactor all the current checks as individual functions as well...

elParaguayo:
@ramnes thanks for continuing to push on this. You're absolutely right, of course, accepting anything that is boolean or a callable function would be a good way of extending `.when()`'s functionality. There's a slight ease-of-use benefit to putting some of these behind keywords but I agree that it's not a good long-term approach.

Let me refactor this so lazy can accept a boolean or a callable.

elParaguayo:
Oops. Looks like I forgot to remove your suggestions for the `qtile` object!

If you're ok with the changes to `.when()` here, I can tidy this up and split that part our into a separate commit.

libqtile/__init__.py
```diff
@@ -20,7 +20,15 @@
...
 
-qtile = None
+class UndefinedCore:
+    name = None
+
+
+class UndefinedQtile:
+    core = UndefinedCore()
+
+
```
m-col, line 30: Please could we make these private to the module, assuming we don't need them elsewhere?

libqtile/lazy.py
```diff
@@ -53,6 +54,8 @@ def __init__(self, call: CommandGraphCall, args: tuple, kwargs: dict) -> None:
         self._if_no_focused: bool = False
         self._layouts: set[str] = set()
         self._when_floating = True
+        self._check: bool | None = None
+        self._func: Callable | None = None
```
m-col, line 58: Could we fully type Callable? It should be `Callable[[], bool]`.

We could probably do something like this as a default value rather than `None`, what do you think?:
```py
        self._func: Callable[[], bool] = lambda: True
```

libqtile/lazy.py
```diff
@@ -93,14 +96,22 @@ def kwargs(self) -> dict:
 
     def when(
         self,
+        check: bool | None = None,
+        *,
```
m-col, line 100: Nice.

Is this a potentially config breaking change? I guess only for the case where people pass an argument to `focused` but without the key word?

libqtile/lazy.py
```diff
@@ -144,6 +163,16 @@ def check(self, q) -> bool:
...
+            try:
+                result = self._func()
+            except Exception:
+                logger.exception("Error when running function in lazy call. Ignoring.")
+                result = True
+
+            if not result:
+                return False
+
         return True
```
m-col, line 176: With the typing change above to `self._func`, this whole bit could just be:
```suggestion
        return self._func()
```
Assuming of course that people's configs are type-safe, which would give us some confidence that an error won't be raised. Ha. Yeah, good to catch an exception :)

m-col (approved):
:fire:

libqtile/lazy.py
elParaguayo: That's a good point.

I wouldn't expect it to be a config breakage as I would have thought that users would always be using keyword args here. However, that's just an expectation!

A migrate script would be no good here either. 

Perhaps, the best idea is to use a keyword for the Boolean test too. Maybe `condition`?

libqtile/lazy.py
elParaguayo: As useful as typing is, I don't see it as a replacement for error handling!

libqtile/__init__.py
elParaguayo: I'll change this to a single underscore prefix unless you want full on name mangling too.

ramnes (approved):
LGTM, thank you for bearing with me to the end :D

PR #4545, opened 2023-10-25
Columns: black formatting

elParaguayo:
I'm getting these changes every time I run `black` on a branch so let's merge them!


ramnes:
Oops, my fault. We probably want to add pre-commit hooks and run that in the CI as well.

PR #4546, opened 2023-10-25
Allow third party code to register hooks

elParaguayo:
Adds a new `hook.register_hook` function to allow third party layouts, widgets etc to create new hooks.


elParaguayo:
I'm going to close this for now. I'm not sure this is the right approach.

PR #4547, opened 2023-10-25
Fix static positioning in Wayland.

QBobWatson:
The static was being positioned as if it still had a border.


jwijenbergh (approved):
LGTM

m-col:
Nice thanks!

PR #4551, opened 2023-10-27
Hook registries

elParaguayo:
@ramnes here's an early attempt at using a registry for hooks.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
       - Add ability to have bar drawns over windows by adding `reserve=False` to bar's config to
         stop the bar reserving screen space.
       - Add ability for third-party code (widgets, layouts) to create hooks
+      - Add ability to create user-defined hooks which can be fired from external scripts
     * bugfixes
       - Fix two bugs in stacking transient windows in X11
       - Checking configs containing `qtile.core.name` with `python config.py` don't fail anymore (but `qtile.core.name`
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
           condition is evaluated once when the config file is first loaded.
       - Add ability to have bar drawns over windows by adding `reserve=False` to bar's config to
         stop the bar reserving screen space.
+      - Add ability for third-party code (widgets, layouts) to create hooks
     * bugfixes
       - Fix two bugs in stacking transient windows in X11
       - Checking configs containing `qtile.core.name` with `python config.py` don't fail anymore (but `qtile.core.name`
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1768,3 +1768,8 @@ class Qtile(CommandObject):
     def run_extension(self, extension: _Extension) -> None:
         """Run extensions"""
         extension.run()
+
+    @expose_command()
+    def fire_user_hook(self, hook_name: str, *args: Any) -> None:
+        """Fire a custom hook."""
+        hook.fire(f"user_{hook_name}", *args)
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -29,12 +29,18 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+from __future__ import annotations
+
 import asyncio
 import contextlib
+from typing import TYPE_CHECKING
 
 from libqtile import utils
 from libqtile.log_utils import logger
 from libqtile.resources.sleep import inhibitor
+
+if TYPE_CHECKING:
+    from typing import Callable
 
 subscriptions = {}  # type: dict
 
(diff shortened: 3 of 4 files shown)


ramnes:
Excellent, thanks for trying out the idea! How do you like it?

I think I'd have done it pretty much the same way, except maybe the `Subscribe` and `Unsubscribe` classes; why do we still need them as classes rather than having them as methods of `Registry`?

You say it's an early attempt; what's left to do?

elParaguayo:
> Excellent, thanks for trying out the idea! How do you like it?

I do like it as it allows me to do some fun stuff with qtile extras (have widgets firing hooks opens up a number of possibilities).

> I think I'd have done it pretty much the same way, except maybe the `Subscribe` and `Unsubscribe` classes; why do we still need them as classes rather than having them as methods of `Registry`?

Because I didn't want to break existing syntax of `hook.unsubscribe.hook_name(func)`. `subscribe` and `unsubscribe` aren't methods.

> You say it's an early attempt; what's left to do?
- Mainly get your reaction!
- Also, fix test failures and then see if it works for qtile extras.
- I would also add the user-defined hooks to this if it's a route we want to go down.

ramnes:
> Because I didn't want to break existing syntax of `hook.unsubscribe.hook_name(func)`. `subscribe` and `unsubscribe` aren't methods.

but, but... you're already doing `subscribe = qtile_hooks.subscribe`, so what would be the difference if `subscribe` was a method of `Registry` rather than an instance of `Subscribe`? am I missing something?

elParaguayo:
> > Because I didn't want to break existing syntax of `hook.unsubscribe.hook_name(func)`. `subscribe` and `unsubscribe` aren't methods.
> 
> but, but... you're already doing `subscribe = qtile_hooks.subscribe`, so what would be the difference if `subscribe` was a method of `Registry` rather than an instance of `Subscribe`? am I missing something?

Yes, you're missing something!

We do `hook.subscribe` but `subscribe` is an attribute of the registry, not `hook`. So, by adding `subscribe = qtile_hooks.subscribe` I don't have to change any existing subscriptions. Without that, I'd have to do `hook.qtile_hooks.subscribe`.

Now you're going to tell me there's simpler way of doing it...

ramnes:
Ahah no I got that, we do need the `subscribe = qtile_hooks.subscribe` indeed, what I'm saying is that `qtile_hooks.subscribe` could be a method rather than a variable of the registry instance. :)

elParaguayo:
How can I do that and preserve the existing syntax?

I don't disagree, there's no need for two classes here but we cannot change the hook syntax as it would impact every single user.

ramnes:
Haha yeah so, what I was missing is that the current API is `subscribe.hook_name`, not `subscribe("hook_name")`, sorry for overlooking that! That being said, functions are also object in Python, so we can assign on them directly as well... :D

Didn't try it at all, but here is what I have in mind overall:

```diff
diff --git a/libqtile/hook.py b/libqtile/hook.py
index bea92685..dec5502f 100644
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -33,6 +33,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import functools
 from typing import TYPE_CHECKING
 
 from libqtile import utils
@@ -42,12 +43,6 @@ from libqtile.resources.sleep import inhibitor
 if TYPE_CHECKING:
     from typing import Callable
 
-subscriptions = {}  # type: dict
-
-
-def clear():
-    subscriptions.clear()
-
 
 def _fire_async_event(co):
     from libqtile.utils import create_task
@@ -62,18 +57,18 @@ def _fire_async_event(co):
         create_task(co)
 
 
-def _resume_func(self):
+def _resume_func(registry):
     def f(func):
         inhibitor.want_resume()
-        return self._subscribe("resume", func)
+        return registry.subscribe("resume", func)
 
     return f
 
 
-def _suspend_func(self):
+def _suspend_func(registry):
     def f(func):
         inhibitor.want_sleep()
-        return self._subscribe("suspend", func)
+        return registry.subscribe("suspend", func)
 
     return f
 
@@ -85,66 +80,40 @@ class Hook:
         self.func = func
 
 
-class Subscribe:
-    def __init__(self, registry_name: str, check_name=True):
-        self.hooks = set([])
-        if check_name and registry_name in subscriptions:
-            raise NameError("A hook registry already exists with that name: {registry_name}")
-        self.registry_name = registry_name
+class Registry:
+    def __init__(self, hooks: list[Hook]) -> None:
+        self.subscriptions = {}
+        for hook in hooks:
+            self.register(hook)
 
-    def _subscribe(self, event: str, func: Callable) -> Callable:
-        if self.registry_name not in subscriptions:
-            subscriptions[self.registry_name] = {}
-        lst = subscriptions[self.registry_name].setdefault(event, [])
-        if func not in lst:
-            lst.append(func)
+    def subscribe(self, hook_name: str, func: Callable) -> Callable:
+        self.subscriptions[hook_name].add(func)
         return func
 
-    def _register(self, hook: Hook) -> None:
-        def _hook_func(func):
-            return self._subscribe(hook.name, func)
-
-        self.hooks.add(hook.name)
-        setattr(self, hook.name, _hook_func if hook.func is None else hook.func(self))
-        setattr(getattr(self, hook.name), "__doc__", hook.doc)
+    def unsubscribe(self, hook_name: str, func: Callable) -> None:
+        self.subscriptions[hook_name].remove(func)
 
-
-class Unsubscribe(Subscribe):
-    """
-    This class mirrors subscribe, except the _subscribe member has been
-    overridden to remove calls from hooks.
-    """
-
-    def _subscribe(self, event: str, func: Callable) -> None:
-        lst = subscriptions[self.registry_name].setdefault(event, [])
-        try:
-            lst.remove(func)
-        except ValueError:
-            raise utils.QtileError(
-                "Tried to unsubscribe a hook that was not currently subscribed"
-            )
-
-
-class Registry:
-    def __init__(self, name: str) -> None:
-        self.name = name
-        self.subscribe = Subscribe(name)
-        self.unsubscribe = Unsubscribe(name, check_name=False)
-
-    def register_hook(self, hook: Hook) -> None:
-        if hook.name in dir(self.subscribe):
+    def register(self, hook: Hook) -> None:
+        if hook.name in self.subscriptions:
             raise utils.QtileError(
                 f"Unable to register hook. A hook with that name already exists: {hook.name}"
             )
 
         logger.debug("Registered new hook: '%s'.", hook.name)
-        self.subscribe._register(hook)
-        self.unsubscribe._register(hook)
+        self.subscriptions[hook.name] = set()
+
+        if hook.func:
+            subscriber = hook.func(self)
+        else:
+            subscriber = functools.partial(self.subscribe, hook.name)
+
+        subscriber.__doc__ = hook.doc
+        setattr(self.subscribe, hook.name, subscriber)
 
     def fire(self, event, *args, **kwargs):
-        if event not in self.subscribe.hooks:
+        if event not in self.subscriptions:
             raise utils.QtileError("Unknown event: %s" % event)
-        for i in subscriptions[self.name].get(event, []):
+        for i in self.subscriptions.get(event, []):
             try:
                 if asyncio.iscoroutinefunction(i):
                     _fire_async_event(i(*args, **kwargs))
@@ -950,12 +919,7 @@ hooks: list[Hook] = [
     ),
 ]
 
-
-qtile_hooks = Registry("qtile")
-
-for hook in hooks:
-    qtile_hooks.register_hook(hook)
-
-subscribe = qtile_hooks.subscribe
-unsubscribe = qtile_hooks.unsubscribe
-fire = qtile_hooks.fire
+registry = Registry(hooks=hooks)
+subscribe = registry.subscribe
+unsubscribe = registry.unsubscribe
+fire = registry.fire
```

ramnes:
My diff doesn't work. :)

> That being said, functions are also object in Python, so we can assign on them directly as well... :D

TIL: this is wrong, we can't assign to bound methods apparently. This is discussed here: https://stackoverflow.com/questions/7891277/why-does-setattr-fail-on-a-bound-method

So we do need to keep the classes. :)

elParaguayo:
Thanks for taking a look at this.

I'm going to see if I can get the user-defined hooks working in this PR too.

ramnes (approved):

ramnes:
Looks nice, is it missing something? We probably want to merge this early otherwise. :)

elParaguayo:
Don't think anything's missing here.

ramnes:
Thanks for this!

PR #4556, opened 2023-11-01
Floating isn't meant to be used as a normal layout

ramnes:
Didn't try to build the docs yet, just throwing this for now so that I don't forget.


jwijenbergh:
FWIW I have a local branch that aims to fix most issues with floating/fullscreen/maximized/minimized windows that we have. Including that the floating layout can be used in your layouts variable

Issue #1413, opened 2019-08-29
drag: wrong window is dragged


pawamoy:
See this screen recording:

![r](https://user-images.githubusercontent.com/3999221/63952715-57d96380-ca80-11e9-9bc9-abed139e8476.gif)

If the window I try to drag is on both screens, then it drags the one that is *only* on this screen. When it is entirely contained within the screen, it works correctly.

ramnes:
Yes, I've already encountered this. Thanks for the great bug report!

ervinpopescu:
TODO: Need to test #4557 after rebase. Don't mind me, just updating this to get it down the stack in the updated-asc sort :)

lolorc:
also reproduced here on 4 screens setup only when cursor_warp=True, #4557 makes it worse

tych0:
Hmm, can you guys share some config snippets? I guess these are `Drag` keybinds? The relevant code is here:

https://github.com/qtile/qtile/blob/2e57a4a9ebb9600109d41cc185b7ec720a3de691/libqtile/core/manager.py#L864-L884

...it's a little crazy that it does an IPC call if you supply `start`, but let's ignore that. It looks like `_focus_hovered_window()` eventually calls this:

https://github.com/qtile/qtile/blob/2e57a4a9ebb9600109d41cc185b7ec720a3de691/libqtile/backend/x11/core.py#L924-L927

which doesn't seem all that wrong, but maybe there's some nuance here?

tych0:
Does https://github.com/qtile/qtile/commit/d798efaca758cd6e76b658386597e89e60b768ea fix it?

Issue #890, opened 2016-03-05
Can't drag floating window from right-most screen when straddling screens


flacjacket:
The assignment of a floating window to a group/screen will follow the upper left corner of the window. When the window straddles 2 screens, Click/Drag events will only work on the window when done from the same screen that this upper left corner is in.

ramnes:
Can confirm, this is still an issue.

PR #4557, opened 2023-11-01
x11: Add pointer barriers for screen focus

elParaguayo:
This is draft for now and comments very welcome.

This is an attempt to fix the issue where screens are not focused when the mouse moves to a different screen.

This PR works by creating pointer barriers at the edges where two `Screen`s touch. When the mouse hits one these barriers, an event is fired which allows us to move the mouse to the next screen and focus that screen.

At the moment, this is automatic and not behind a config option.

This probably needs some tests too...

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -822,7 +822,7 @@ class Core(base.Core):
         mouse_y = event.root_y >> 16
 
         if x is not None:
-            if mouse_x >> 16 < x:
+            if mouse_x < x:
                 # Focus screen 2
                 mouse_x += 1
                 screen = screen2
@@ -832,7 +832,7 @@ class Core(base.Core):
                 screen = screen1
 
         else:
-            if event.root_y >> 16 < y:
+            if mouse_y < y:
                 # Focus screen 2
                 mouse_y += 1
                 screen = screen2
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -44,6 +44,7 @@ from libqtile.utils import QtileError
 if TYPE_CHECKING:
     from typing import Callable, Iterator
 
+    from libqtile.config import Screen
     from libqtile.core.manager import Qtile
 
 _IGNORED_EVENTS = {
@@ -809,6 +810,40 @@ class Core(base.Core):
     def handle_ScreenChangeNotify(self, event) -> None:  # noqa: N802
         hook.fire("screen_change", event)
 
+    def handle_BarrierHit(self, event):
+        """Warps pointer to other side of barrier and focuses the relevant screen."""
+        barrier = self.conn.xfixes.barriers.get(event.barrier)
+        if barrier is None:
+            logger.warning("Unknown barrier hit event.")
+            return
+
+        x, y, screen1, screen2 = barrier
+        mouse_x = event.root_x >> 16
+        mouse_y = event.root_y >> 16
+
+        if x is not None:
+            if mouse_x >> 16 < x:
+                # Focus screen 2
+                mouse_x += 1
+                screen = screen2
+            else:
+                # Focus screen 1
+                mouse_x -= 1
+                screen = screen1
+
+        else:
+            if event.root_y >> 16 < y:
+                # Focus screen 2
+                mouse_y += 1
+                screen = screen2
+            else:
+                # Focus screen 1
+                mouse_y -= 1
+                screen = screen1
+
+        self.warp_pointer(mouse_x, mouse_y)
+        self.qtile.focus_screen(screen.index, warp=False)
+
     def _fake_input(self, input_type, detail, x=0, y=0) -> None:
         self._xtest.FakeInput(
             input_type,
@@ -952,3 +987,36 @@ class Core(base.Core):
             self.last_focused.change_layer()
 
         self.last_focused = win
+
+    def setup_barriers(self, screens: list[Screen]) -> None:
+        if not (hasattr(self.conn, "xinputextension") and hasattr(self.conn, "xfixes")):
+            logger.warning("Unable to set pointer barriers. Missing extensions.")
+            return
+        self.conn.xinputextension.select_events(self._root.wid)
+
+        if len(screens) < 2:
+            return
+
+        borders = set()
+
+        for screen in screens:
+            for other in [s for s in screens if s is not screen]:
+                border = screen.get_joining_edge(screen, other)
+                if border is None:
+                    continue
+
+                # Vertical border (x coordinate is the same)
+                if border[0] == border[2]:
+                    # We need the leftmost screen first
+                    first = screen if screen.x < other.x else other
+                # Horizontal border
+                else:
+                    # We need higher screen first
+                    first = screen if screen.y < other.y else other
+
+                touching = other if first is screen else screen
+
+                borders.add((*border, first, touching))
+
+        for border in borders:
+            self.conn.xfixes.add_pointer_barrier(self._root.wid, *border)
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2019 Aldo Cortesi
 # Copyright (c) 2019 Sean Vig
+# Copyright (c) 2023 elParaguayo
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -1001,7 +1002,7 @@ class Core(base.Core):
 
         for screen in screens:
             for other in [s for s in screens if s is not screen]:
-                border = screen.get_joining_edge(screen, other)
+                border = screen.get_joining_edge(other)
                 if border is None:
                     continue
 
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -811,7 +811,7 @@ class Core(base.Core):
     def handle_ScreenChangeNotify(self, event) -> None:  # noqa: N802
         hook.fire("screen_change", event)
 
-    def handle_BarrierHit(self, event):
+    def handle_BarrierHit(self, event) -> None:  # noqa: N802
         """Warps pointer to other side of barrier and focuses the relevant screen."""
         barrier = self.conn.xfixes.barriers.get(event.barrier)
         if barrier is None:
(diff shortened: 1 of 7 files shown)


ramnes:
How does other WM handle this?

elParaguayo:
Good question. Don't know the answer. Does it matter? ;)

ramnes:
I'm simply wondering if this is "the right way™️".

Yeah I broke the CI but it's fixed on master now, you can simply rebase.

elParaguayo:
I don't even know if other window managers have the ability to focus screens by moving mouse cursor to that screen. However, I didn't see any reference to pointer barriers in i3, dwm or bspwm.

If you don't want it the main repo then I can probably find a way to make it available via qtile-extras.

Etn40ff:
> I don't even know if other window managers have the ability to focus screens by moving mouse cursor to that screen. However, I didn't see any reference to pointer barriers in i3, dwm or bspwm.

Not a i3 user here so I can't test but the discussion in this issue https://github.com/i3/i3/issues/3639 seems to suggest that i3 does.

jwijenbergh:
Dwm and i3 both handle it. Dwm handles it by finding the monitor under the cursor on each motion notify event https://git.suckless.org/dwm/file/dwm.c.html#l1137

@elParaguayo's solution is more complicated but has the advantage of being more performant/doing less work per mouse move on the root window. No clue what i3 does but I cba to look

libqtile/backend/x11/core.py
```diff
@@ -952,3 +987,36 @@ def check_stacking(self, win: base.Window) -> None:
...
+        self.conn.xinputextension.select_events(self._root.wid)
+
+        if len(screens) < 2:
+            return
+
+        borders = set()
+
+        for screen in screens:
+            for other in [s for s in screens if s is not screen]:
+                border = screen.get_joining_edge(screen, other)
```
ramnes, line 1004: I'm one of those who think that static methods are a code smell... Do you mind updating `get_joining_edge` to a real method? :)

```suggestion
                border = screen.get_joining_edge(other)
```

ramnes:
@jwijenbergh i3 also does it with motion notify events ([source](https://github.com/i3/i3/blob/next/src/handlers.c#L206)), but we're indeed limited by CPython here. :)

ramnes (changes requested):
LGTM for the technique (where did you hear about barrier hits?), that's very nice.

That being said, going from a left screen to a right screen, I can't go back to the left screen with the mouse. :)

libqtile/backend/x11/core.py
elParaguayo: There was a reason I did this but I don't remember now... It does feel like an ordinary method should work!

elParaguayo:
> LGTM for the technique (where did you hear about barrier hits?), that's very nice.

Heard about it from @m-col 

> That being said, going from a left screen to a right screen, I can't go back to the left screen with the mouse. :)

Well that's not good...

There is a way to allow the mouse to pass through the barrier but the events are very erratic when I do that so we can't rely on them. However, when the barrier is solid, the events are fired perfectly and we then move the mouse to the other side.

Also, is this something that we need to put behind a config option?

ramnes:
I don't see the point for a config option; I think all users would want that behavior, assuming it's fully functional. (and if it's not, I don't think we want it merged)

ramnes:
> That being said, going from a left screen to a right screen, I can't go back to the left screen with the mouse. :)

fixed :))

ramnes (approved):

elParaguayo:
Thanks for the fix, I wonder if that will make my test work. I hadn't pushed it as the event was only firing when the mouse moved in one direction...!

elParaguayo:
I think there may be a problem with this PR in its current form: I think it breaks dragging across screen boundaries...

ramnes:
Dragging across screens never worked correctly

elParaguayo:
> Dragging across screens never worked correctly

Are you sure? I can only test with `fake_screens` at the moment and dragging between those works fine before this PR and is broken with it.

ramnes:
Yes, on my phone right now but there's likely multiple issues opened on this matter. I have two screens and never try to drag a window from a screen to another because it messes things up.

jwijenbergh:
> Yes, on my phone right now but there's likely multiple issues opened on this matter. I have two screens and never try to drag a window from a screen to another because it messes things up.

I never had issues with dragging across screens

elParaguayo:
@jwijenbergh are you able to test on x11? This shouldn't get merged if it breaks existing (working) behaviour.

jwijenbergh:
> @jwijenbergh i3 also does it with motion notify events ([source](https://github.com/i3/i3/blob/next/src/handlers.c#L206)), but we're indeed limited by CPython here. :)

I doubt python couldn't handle it to be fair, we already handle motion notifies in a delayed manner + with a limit (this is also what c wms do). I think @elParaguayo's solution is just neater

jwijenbergh:
> @jwijenbergh are you able to test on x11? This shouldn't get merged if it breaks existing (working) behaviour.

Will have a look

elParaguayo:
I think it should be possible to fix the issue here. The problem is that, when the new screen is focused `current_window` may change to `None` or another window so we need to check if we're dragging when the focus changes and react accordingly.

ramnes:
I just checked on X11 and the problems I have with dragging windows across screens seem exaggerated with this PR, although it's really not usable anyway. Existing issues: #890 #1413

dmolin:
> Yes, on my phone right now but there's likely multiple issues opened on this matter. I have two screens and never try to drag a window from a screen to another because it messes things up.

I can confirm this is my experience too. if I try dragging a floating window across screen boundaries all hell breaks loose!

elParaguayo:
Thanks for testing. I'm still working on a fix for that. Will update here as soon as I've got something workable.

jwijenbergh:
> > Yes, on my phone right now but there's likely multiple issues opened on this matter. I have two screens and never try to drag a window from a screen to another because it messes things up.
> 
> I can confirm this is my experience too. if I try dragging a floating window across screen boundaries all hell breaks loose!

But are you saying that's also on master or only this pr?

elParaguayo:
It's possible there's an issue anyway as the `current_window` which is the target of the `Drag` could get reset during the drag in certain scenarios. The barriers definitely have this issue but it's possible it happens without it too. I've not tested it much.

dmolin:
Basically, what happens with the current Qtile 0.23.0 is that dragging windows between screens only works if those screens are completely empty OR if they ONLY contain floating windows.

jwijenbergh:
> > Thanks for testing. I'm still working on a fix for that. Will update here as soon as I've got something workable.
> 
> just wanted to clarify that this has been a long-standing issue with Qtile under X11. Yes @jwijenbergh , I'm using Qtile stable 0.23.0 and this is my current experience. I never drag floating windows between screens since I know it's not working at all most of the times

Ah thanks. I will try to fix this, I don't use X11 much anymore but I am currently refactoring floating windows inside of qtile

elParaguayo:
I've pushed a commit which may help dragging window between screens but would be good if someone can test.

dmolin:
I would be more than glad to test this but I'm not sure how to do this.. could someone drives me through the steps, considering that I've already installed the standard version of Qtile (that I want to keep since I'm using my laptop for work so I don't want to mess up my main WM)?

elParaguayo:
> It seems even more broken to me

Ha. Wow. Ok. Back to the drawing board...

ramnes:
> I would be more than glad to test this but I'm not sure how to do this.. could someone drives me through the steps, considering that I've already installed the standard version of Qtile (that I want to keep since I'm using my laptop for work so I don't want to mess up my main WM)?

What you want it to run a second instance of Qtile that's from a virtualenv and uses this branch. Basically, something like this:

```
git clone git@github.com:qtile/qtile.git
cd qtile
./dev.sh  # setup the virtualenv and deps
gh pr checkout 4557  # checkout the branch for this PR
. venv/bin/activate
which qtile  # make sure it's going to use our virtualenv version of qtile
startx qtile start  # run the second X session
```

dmolin:
Is anyone able to give this a test-run? I cannot do it right now since I cannot risk destabilizing my work environment and I've no other laptops to test with

elParaguayo:
Very hard for me to test this one regularly as I rarely have a second screen hooked up to my laptop.

Hate9:
poking this because github actions apparently closed it and it clearly just needs testing

ervinpopescu:
I'll give testing this a go, subscribing to this thread so I get reminded about it.

elParaguayo:
I'll reopen this but it almost certainly needs rebasing onto the latest git version. I'll see if I can do that at some point today.

elParaguayo:
> poking this because github actions apparently closed it and it clearly just needs testing

"just needs testing" may be a slight understatement. I think this triggered some unexpected behaviours too. e.g. When you do test, can you look at dragging a window between screens.

I would still be keen to merge this if we can as it's quite a neat solution.

elParaguayo:
This has been rebased so testers welcome.

elParaguayo:
I've rebased this and made a small fix to dragging windows between screens.

From a quick test, this is working well for me but would like more feedback from users.

JohanTiniusSkjelberg:
I just tried testing it and got some bugs. The barrier worked with a touchpad but not with my Logictech mouse. When using a mouse and crossing screens the mouse warped to the middle of the main monitor.

I also got some applications frozen until clicked when using the mouse.

tych0:
Yeah, I'm happy to test too, as I have multiple monitors. Honestly, the code looks okay to me. I think we should land this and get wider testing.

ervinpopescu:
Ok, I can see why dragging across screens was not great, I rarely did it myself as well and it's bad, I have no idea if it was better or worse before, but at least I can finish dragging it from the other screen if and when it gets stuck.

The main point of this PR is focus across screens, which works perfectly for me, so LGTM for the merge.

(QQ: does wayland need this?)

elParaguayo:
@tych0 Have you tested this? Wondering if we should get it merged.

elParaguayo:
@tych0 should I clean up the commit history here so we can merge?

lolorc:
I've just been testing this PR on my 3 screens setup, the screen focus only works with 2 screens and the dragging issue is still there.
I also have a 4 screens setup to try it on but at the moment it's not usable with 3. (even with cursor_warp=False)

elParaguayo:
Thanks for the feedback. We'll hold off merging for now.

I'll see if I can get round to having another look at this at some point but it's not a priority issue for me as I don't use multiple monitors or x11!

PR #4558, opened 2023-11-03
Black fixes for earlier commits

elParaguayo:
Not sure why I'm still seeing `black` want to make changes after I've pulled the latest commits but here we are...


ramnes:
And you got me working on #4559 :))

PR #4559, opened 2023-11-04
Move linters from tox to pre-commit

ramnes:
Tox makes a lot of sense for running our tests against multiple Python versions but not much for linters. Let's replace it with pre-commit that's 1. more modern 2. makes linting reproducible and 3. will let us avoid pushing code that didn't pass linters.

Plus, it seems to me that the linters are not running at all currently. :)


ramnes:
@m-col I changed a few things I'm not sure about in the Wayland backend, you might want to take a look

PR #4567, opened 2023-11-10
`qtile_info`: `config_path` => `PosixPath` to `str`

ervinpopescu:
`self.config.file_path` is by default a `Path` (since we are using `python` on `Linux`, a `PosixPath`, see [`get_config_file`](https://github.com/qtile/qtile/blob/master/libqtile/utils.py#L204)), if no config is provided.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -32,6 +32,7 @@ import subprocess
 import tempfile
 from collections import defaultdict
 from logging.handlers import RotatingFileHandler
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 import libqtile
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for looking at this. Is `self.config.file_path` ever `None` or is it always a `Path`? If so, we could use `.as_posix()` rather than the `str()` conversion. It feels a bit neater.

What's the value when the default config is used?

ervinpopescu:
It's either a `Path` by default, as the `default` for `options.configfile` in [`scripts/start.py`](https://github.com/qtile/qtile/libqtile/scripts/start.py) is returned by [`get_config_file`](https://github.com/qtile/qtile/blob/master/libqtile/utils.py#L204), or a `str`, as the type is not specified in its corresponding [`parser.add_argument`](https://github.com/qtile/qtile/libqtile/scripts/start.py#L118-125).

`options.configfile` is then passed to `confreader.Config` in [`make_qtile`](https://github.com/qtile/qtile/libqtile/scripts/start.py#L118-125). Maybe we could pass it directly as `Path` and not convert it in [`Config.load`](https://github.com/qtile/qtile/libqtile/confreader.py#L126)? It's not used anywhere else as far as I can see.

elParaguayo:
Docs are failing to build because it looks like `file_path` is `None` and so `Path(None)` fails.

ervinpopescu:
Oh, I see, that's sad. I will revert these changes and add config_path based on its type, I guess.

ervinpopescu:
Well, `as_posix` is just `str(self).replace(f.sep, '/')` sooo, maybe just `str` is fine? :)

ervinpopescu:
I guess I should look at the `hacking` page again on the Qtile docs. :)

elParaguayo:
Can you rebase this and force push please. We've added python 3.12 to CI to need to re-run it.

ervinpopescu:
🥳

Issue #4565, opened 2023-11-08
[wayland] missing handle_activation_request() in SceneTree


pg83:
### Issue description

Sometimes have such errors in qtile.log, do not quite understand how to reproduce it, and when it happens:

```
2023-11-08 16:12:51,658 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
Traceback (most recent call last):
  File "/ix/store/UCa9W8mB9JMNDogm-lib-py-wayland-py/lib/pywayland/server/listener.py", line 40, in notify_func
    callback(listener, data)
  File "/ix/store/SL8asQZCgLklM1pM-lib-qtile-py/lib/libqtile/backend/wayland/core.py", line 853, in _on_xdg_activation_v1_request_activate
    win.handle_activation_request(focus_on_window_activation)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SceneTree' object has no attribute 'handle_activation_request'
2023-11-08 16:12:53,149 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
```

### Version

dev

### Backend

Wayland (experimental)

### Config

```python
https://github.com/pg83/bayanwm/blob/main/config.py
```

### Logs

```text
2023-11-08 16:12:51,658 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
Traceback (most recent call last):
  File "/ix/store/UCa9W8mB9JMNDogm-lib-py-wayland-py/lib/pywayland/server/listener.py", line 40, in notify_func
    callback(listener, data)
  File "/ix/store/SL8asQZCgLklM1pM-lib-qtile-py/lib/libqtile/backend/wayland/core.py", line 853, in _on_xdg_activation_v1_request_activate
    win.handle_activation_request(focus_on_window_activation)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SceneTree' object has no attribute 'handle_activation_request'
2023-11-08 16:12:53,149 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
```

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

pg83:
Managed to fix it in https://github.com/pg83/ix/blob/main/pkgs/bin/qtile/py/core.py#L852-L853, will send PR later

[...]

PR #4568, opened 2023-11-10
proper popups handling

pg83:
Fix for https://github.com/qtile/qtile/issues/4565

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -923,6 +923,8 @@ class Core(base.Core, wlrq.HasListeners):
         surface = event.surface
         if surface and surface.is_xdg_surface:
             xdg_surface = XdgSurface.from_surface(surface)
+            while xdg_surface.role == XdgSurfaceRole.POPUP:
+                xdg_surface = XdgSurface.from_surface(xdg_surface.popup.parent)
             if win := xdg_surface.data:
                 win.handle_activation_request(focus_on_window_activation)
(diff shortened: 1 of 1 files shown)


jwijenbergh:
Thanks for the PR. Maybe we also need to check if there is no more parent or of the role is None? See e.g. https://github.com/djpohly/dwl/blob/d6fabe3a150d2e464078c3269410d56e8bbeb54f/client.h#L91-L104.

libqtile/backend/wayland/core.py
```diff
@@ -923,6 +923,8 @@ def _on_xdg_activation_v1_request_activate(
         surface = event.surface
         if surface and surface.is_xdg_surface:
             xdg_surface = XdgSurface.from_surface(surface)
```
jwijenbergh, line 925: Maybe we also have to check if this is None? Other compositors seem to do it

jwijenbergh: I done that in #4571, you can rebase if that is merged

pg83:
> Thanks for the PR. Maybe we also need to check if there is no more parent or of the role is None? See e.g. https://github.com/djpohly/dwl/blob/d6fabe3a150d2e464078c3269410d56e8bbeb54f/client.h#L91-L104.

I would prefer to keep the original version, because this is what I have tested, and for which I can be responsible.

If I have any more errors on this topic in my log, I will come and fix them again!

PR #4569, opened 2023-11-11
Ethernet functionality on the Wlan widget

Sprinter05:
Using `os.popen()` and a simple `cat` command it is possible to check the status of the ethernet interface that can be defined using the `ethint` parameter and then output a message that can also be changed by modifying the `ethernet_message` parameter.
I did not include this check in the case the wlan device is not detected because if the user does not have a wlan device why would they be using the `Wlan` widget in the first place? I also did not include an `ssid` parameter or `quality` parameter since those do not apply to ethernet.

### DIFF
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -73,12 +73,11 @@ class Wlan(base.InLoopPollText):
             essid, quality = get_status(self.interface)
             disconnected = essid is None
             if disconnected:
-                statfile = open("/sys/class/net/"+self.ethint+"/operstate", "r")
-                if (statfile.read().strip() == 'up'):
-                    return self.ethernet_message
-                else:
-                    return self.disconnected_message
-
+                with open(f"/sys/class/net/{self.ethint}/operstate", "r") as statfile:
+                    if (statfile.read().strip() == 'up'):
+                        return self.ethernet_message
+                    else:
+                        return self.disconnected_message
             return self.format.format(essid=essid, quality=quality, percent=(quality / 70))
         except EnvironmentError:
             logger.error(
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -68,6 +68,7 @@ class Wlan(base.InLoopPollText):
     def __init__(self, **config):
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Wlan.defaults)
+        self.ethernetInterfaceNotFound = False
 
     def poll(self):
         try:
@@ -82,17 +83,17 @@ class Wlan(base.InLoopPollText):
                             else:
                                 return self.disconnected_message
                     except FileNotFoundError:
-                        logger.error(
-                            "%s: Ethernet interface has not been found!",
-                            self.__class__.__name__,
-                        )
+                        if not self.ethernetInterfaceNotFound:
+                            logger.error(
+                                "Ethernet interface has not been found!"
+                            )
+                            self.ethernetInterfaceNotFound = True
                         return self.disconnected_message
                 else:
                     return self.disconnected_message
             return self.format.format(essid=essid, quality=quality, percent=(quality / 70))
         except EnvironmentError:
             logger.error(
-                "%s: Probably your wlan device is switched off or "
-                " otherwise not present in your system.",
-                self.__class__.__name__,
+                "Probably your wlan device is switched off or "
+                " otherwise not present in your system."
             )
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -73,11 +73,18 @@ class Wlan(base.InLoopPollText):
             essid, quality = get_status(self.interface)
             disconnected = essid is None
             if disconnected:
-                with open(f"/sys/class/net/{self.ethernet_interface}/operstate", "r") as statfile:
-                    if (statfile.read().strip() == 'up'):
-                        return self.ethernet_message
-                    else:
-                        return self.disconnected_message
+                try:
+                    with open(f"/sys/class/net/{self.ethernet_interface}/operstate", "r") as statfile:
+                        if (statfile.read().strip() == 'up'):
+                            return self.ethernet_message
+                        else:
+                            return self.disconnected_message
+                except FileNotFoundError:
+                    logger.error(
+                        "%s: Ethernet interface has not been found!",
+                        self.__class__.__name__,
+                    )
+                    return self.disconnected_message
             return self.format.format(essid=essid, quality=quality, percent=(quality / 70))
         except EnvironmentError:
             logger.error(
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -27,6 +27,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import iwlib
+import os
 
 from libqtile.log_utils import logger
 from libqtile.widget import base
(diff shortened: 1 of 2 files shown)


jwijenbergh:
Hi! Thanks for the pr, but why not just read the file with python?

Sprinter05:
> Hi! Thanks for the pr, but why not just read the file with python?

Oh true I forgot python has the `open()` function, I can modify it if that's better.

jwijenbergh:
> > Hi! Thanks for the pr, but why not just read the file with python?
> 
> Oh true I forgot python has the `open()` function, I can modify it if that's better.

Yes that would be much better!

Sprinter05:
> > > Hi! Thanks for the pr, but why not just read the file with python?
> > 
> > 
> > Oh true I forgot python has the `open()` function, I can modify it if that's better.
> 
> Yes that would be much better!

Done! Sorry it took a bit but its now using `open()` instead of `os` library

libqtile/widget/wlan.py
```diff
@@ -71,7 +73,11 @@ def poll(self):
             essid, quality = get_status(self.interface)
             disconnected = essid is None
             if disconnected:
-                return self.disconnected_message
+                statfile = open("/sys/class/net/"+self.ethint+"/operstate", "r")
```
jwijenbergh, line 76: Could you use 
```python
with open(f"/sys/class/net/{self.ethint}/operstate", "r")) as statfile:
    ... rest of code indented here ...
```

here? That makes sure the file is closed at the end and makes sure f strings are used

libqtile/widget/wlan.py
```diff
@@ -53,8 +53,10 @@ class Wlan(base.InLoopPollText):
     orientations = base.ORIENTATION_HORIZONTAL
     defaults = [
         ("interface", "wlan0", "The interface to monitor"),
+        ("ethint", "eth0", "The ethernet interface to monitor"),
```
jwijenbergh, line 56: I think eth_interface would be a more descriptive name

jwijenbergh (changes requested):
Thanks for this! Some small requests

libqtile/widget/wlan.py
jwijenbergh: Could you also document that this is only used if there is no wireless connection?

libqtile/widget/wlan.py
Sprinter05: Done!

libqtile/widget/wlan.py
Sprinter05: Also done!

Sprinter05:
> Thanks for this! Some small requests

I have added your requested changes and also added some try and error checks to prevent undesired results and a variable to use or not this ethernet functionality, you have more info in the commits, let me know if there is anything you would prefer me to change!

Sprinter05:
> > Thanks for this! Some small requests
> 
> I have added your requested changes and also added some try and error checks to prevent undesired results and a variable to use or not this ethernet functionality, you have more info in the commits, let me know if there is anything you would prefer me to change!

Finally I was thinking of maybe adding a priority variable for whether the user prefers to check first if Wlan is connected or if Ethernet is connected due to some situations where the network service might be detecting both ethernet connected and also wlan and even if the ethernet is the one being used the widget will display Wlan unless the wireless connection is deactivated (happens to me using NetworkManager, i dont know how it might behave with other network services)

libqtile/widget/wlan.py
```diff
@@ -71,8 +74,21 @@ def poll(self):
...
+                        with open(f"/sys/class/net/{self.ethernet_interface}/operstate", "r") as statfile:
+                            if (statfile.read().strip() == 'up'):
+                                return self.ethernet_message
+                            else:
+                                return self.disconnected_message
+                    except FileNotFoundError:
+                        logger.error(
+                            "%s: Ethernet interface has not been found!",
+                            self.__class__.__name__,
+                        )
```
elParaguayo, line 88: Two things:
1) You don't need `self.__class__.__name__`. The log includes the name of the file already.
2) Can you include some sort of flag so this message doesn't get logged every time the widget polls?

elParaguayo (changes requested):
Have left a suggestion regarding some changes.

I would also like to include a test for this but let me know if you can't do it and I'll send you the diff for one when I get the chance.

libqtile/widget/wlan.py
Sprinter05: Sure I can try to get this worked out tomorrow. Also are the other changes requested from some weeks ago okay? I have had no notice of that yet.

jwijenbergh: You mean my comments? You fixed them nicely, thanks!

Sprinter05: Oh okay, I was a bit confused since Github still says youre requesting changes

jwijenbergh (approved):

libqtile/widget/wlan.py
jwijenbergh: Sorry! Approved now

Sprinter05: Appreciated! I will try to get the other changes done once I have some time

Sprinter05:
> Have left a suggestion regarding some changes.
> 
> I would also like to include a test for this but let me know if you can't do it and I'll send you the diff for one when I get the chance.

All the changes have been now commited, sorry for talking so long. Is there anything else you want me to do?

elParaguayo:
Here's a test:
```diff
diff --git a/test/widgets/test_wlan.py b/test/widgets/test_wlan.py
index 2095d9bf..f910ffa8 100644
--- a/test/widgets/test_wlan.py
+++ b/test/widgets/test_wlan.py
@@ -19,7 +19,7 @@
 # SOFTWARE.

 # Widget specific tests
-
+import builtins
 import sys
 from importlib import reload
 from types import ModuleType
@@ -34,6 +34,23 @@ def no_op(*args, **kwargs):
     pass


+def mock_open(output):
+    class MockOpen:
+        def __init__(self, *args):
+            self.output = output
+
+        def __enter__(self):
+            return self
+
+        def __exit__(*exc):
+            return None
+
+        def read(self):
+            return self.output
+
+    return MockOpen
+
+
 class MockIwlib(ModuleType):
     DATA = {
         "wlan0": {
@@ -44,7 +61,10 @@ class MockIwlib(ModuleType):
             "ESSID": b"QtileNet",
             "Mode": b"Managed",
             "stats": {"quality": 49, "level": 190, "noise": 0, "updated": 75},
-        }
+        },
+        "wlan1": {
+            "ESSID": None,
+        },
     }

     @classmethod
@@ -80,3 +100,23 @@ def test_wlan_display(minimal_conf_noscreen, manager_nospawn, patched_wlan, kwar

     text = manager_nospawn.c.bar["top"].info()["widgets"][0]["text"]
     assert text == expected
+
+
+@pytest.mark.parametrize(
+    "kwargs,state,expected",
+    [
+        ({"interface": "wlan1", "use_ethernet": True}, "up", "eth"),
+        ({"interface": "wlan1", "use_ethernet": True}, "down", "Disconnected"),
+        (
+            {"interface": "wlan1", "use_ethernet": True, "ethernet_message": "Wired"},
+            "up",
+            "Wired",
+        ),
+    ],
+)
+def test_ethernet(
+    minimal_conf_noscreen, manager_nospawn, patched_wlan, kwargs, state, expected, monkeypatch
+):
+    monkeypatch.setattr(builtins, "open", mock_open(state))
+    widget = patched_wlan.Wlan(**kwargs)
+    assert widget.poll() == expected
```

Sprinter05:
Got it, I suppose I should change the interface values to the proper ones in my system when running the test right?

Sprinter05:
`pytest --backend x11 test/widgets/test_wlan.py`
```
=================== test session starts ===================
platform linux -- Python 3.11.8, pytest-7.4.4, pluggy-1.4.0
rootdir: /home/sprinter/Programming/qtile
configfile: pyproject.toml
plugins: anyio-4.3.0
collected 3 items                                         

test/widgets/test_wlan.py ...                       [100%]

==================== 3 passed in 0.79s ====================
```
Any other test required?

elParaguayo (approved):

PR #4570, opened 2023-11-12
Fix bar positioning when not reserving space

elParaguayo:
Where a bottom/right bar is used, this was positioned to the bottom/right of the available screen space which had been adjusted to leave space for the bar. However, with the introduction of the `reserve=False` configuration option, this logic fails as the available screen space is not adjusted so the bar ends up being positioned off screen. This PR corrects that logic.


jwijenbergh (approved):
LGTM

PR #4572, opened 2023-11-12
bump python versions

tych0:
here's a somewhat science experiment to see if we can move to pypy 3.10 and python 3.12.

As far as I can tell, we don't need (or at least, we don't) pass mypy's --python-version flag any more, so I dropped that from the list of places to update.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,9 @@ jobs:
                 # If you change one of these, be sure to update:
                 # - /tox.ini:[gh-actions]
                 # - /setup.cfg:[mypy]
-                # - /libqtile/scripts/check.py mypy argument
                 # If adding new python versions, consider also updating
                 # python version in .readthedocs.yaml
-                python-version: [pypy-3.9, 3.9, '3.10', '3.11']
+                python-version: [pypy-3.10, 3.9, '3.10', '3.11', '3.12']
                 backend: ['x11', 'wayland']
         steps:
             - uses: actions/checkout@v3
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ skipsdist=True
 minversion = 4.0.12
 envlist =
     # Python environments with specific backend
-    py{py3,39,310,311}-{x11,wayland}
+    py{py3,39,310,311,312}-{x11,wayland}
     docs,
     packaging,
     # For running pytest locally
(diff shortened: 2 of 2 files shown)


elParaguayo:
Thanks for this. Are any of the main distros still on 3.9?

We'll also need to change branch protections.

tych0:
Ubuntu 20.04 is 3.8, but we've already dropped support for that; alpine linux has 3.11 in its tree, F39 has 3.12, the previous few have 3.11, and Arch is arch, so IMO I think it's safe to drop.

I wonder about the test failure though... looks like maybe I forgot to do 3.9 -> 3.10 in one place for pypy. Let me push a fix...

m-col (approved):

carlwgeorge:
> Are any of the main distros still on 3.9?

CentOS 9 and its downstream distributions (RHEL 9, AlmaLinux 9, etc.) default to Python 3.9.  I've been working with @FrostyX on bringing qtile back to Fedora's official repos, and if there is demand it could be shipped in EPEL 9 for the EL9 distro family as well.

Debian 11 also still uses Python 3.9.

ramnes:
If we add support for 3.12 and that maintaining support 3.9 is cumbersome, sorry for the "stable" distros but let's just drop it. We (maintainers) discussed this already and ended up to the conclusion that we'd only put effort in supporting the last three Python versions. That being said, if we can also have 3.9 support "for free" and that it doesn't add any inconvenience, there's no reason in dropping it, so I'd be in favor of keeping 3.9 in the CI as long as we don't really need a feature that's not in 3.9. :)

@tych0 Did you drop it for a particular reason other than "we support the last three versions"?

elParaguayo:
Agree with ramnes. I'm not aware of any restrictions we're currently hitting with 3.9 so I'd be ok keeping it in for now.

FrostyX:
Thank you all, @ramnes point of view sounds reasonable to me. For RHEL9 and clones, we would appreciate keeping Qtile support for Python 3.9 as long as you can before it starts to complicate your life too much. We are going to be stuck with 3.9 on those distros for years. However, we won't care about many consequent python versions after that, so if you keep 3.9 longer and then cut support for multiple python versions at once, that would be very much okay with us.

Of course do whatever is the best option for you, just providing feedback from the maintenance perspective.

tych0:
Yep, no reason other than "last three versions". I'll re add it shortly...

ramnes:
Looks like the experiment was a success. :)

Thanks!

tych0:
Thanks, I just updated the branch protections, but it looks like that will also force everyone to rebase :(

elParaguayo:
Do we need to bump the release to build a wheel for these new versions?

PR #4575, opened 2023-11-13
Fix packaging test

elParaguayo:
The packaging test has not been running since splitting x11 and wayland backends in CI. This PR should address that by adding the backends to test in `tox.ini`.


elParaguayo:
Excellent. It works now and fails as I expected (`.pre-commit-config.yaml` is not part of the manifest yet). Will fix this when I get home.

ramnes (approved):

ramnes:
Thanks!

PR #4577, opened 2023-11-14
Add default switch VT keybinding for Wayland

elParaguayo:
Building on the work in #4532 and the additional options in `lazy.when()`, this PR adds default keybindings to allow switching of VTs in Wayland.


elParaguayo:
@jwijenbergh Would be great if you could test this.

jwijenbergh (approved):
We agreed on discord that tty 7 should also be useful, this has been added now so approved. works as expected

cjbayliss:
This looks good to me 👍 I love the lispy `when`, wish it and `unless` were part of python. I’m still not well enough to get up and test it at my computer, sorry.

PR #4579, opened 2023-11-16
Prevent KeyError in third party hook registries

elParaguayo:
Reloading the config can cause third party hook registries to raise a KeyError when firing hooks. This is because the `subscriptions` dict is cleared but the regitry key is only created when a hook is subscribed (as the hook module is not reloaded). Therefore, if a user does not subcribe to any hooks from the third party library, this will cause the hook to fail.

Fixes: https://github.com/elParaguayo/qtile-extras/issues/301

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -175,6 +175,12 @@ class Registry:
     def fire(self, event, *args, **kwargs):
         if event not in self.subscribe.hooks:
             raise utils.QtileError("Unknown event: %s" % event)
+        # We should check if the registry name is in the subscriptions dict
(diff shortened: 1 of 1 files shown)


elParaguayo:
Confirmed it works here: https://github.com/elParaguayo/qtile-extras/issues/301#issuecomment-1816559358

ramnes:
I think I have a branch that does something like this already, let me check later – but we can probably merge this in the meantime if I'm too slow

elParaguayo:
No problem. Have a check first and then we can merge later if needed.

PR #4580, opened 2023-11-17
Renderer API for Wayland.

QBobWatson:
This PR is a simple framework for installing a custom renderer in Wayland, as discussed here: https://github.com/qtile/qtile/discussions/4553 .  There's a new abstract class called `Renderer` and an implementation called `WlrootsRenderer` that retains the current behavior.

There are only a few substantive changes:
- Now `wayland.output.Output` calls `core.renderer.render(self)` instead of `scene_output.commit()`.
- Now `wayland.window.Window.paint_borders` calls `core.renderer.create_borders(self)` to allow for custom borders.
- I moved the logic from `Window.paint_borders` to `WlrootsRenderer`.
- There's a new config option `wl_renderer`, which can optionally be set to install a custom renderer. 

I also deleted `wayland.window.Internal.paint_borders`, since it seems never to be used (??).

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -87,6 +87,7 @@ from libqtile import hook, log_utils
 from libqtile.backend import base
 from libqtile.backend.wayland import inputs, layer, window, wlrq, xdgwindow, xwindow
 from libqtile.backend.wayland.output import Output
+from libqtile.backend.wayland.renderer import Renderer, WlrootsRenderer
 from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
 
@@ -169,7 +170,7 @@ class Core(base.Core, wlrq.HasListeners):
         (
             self.compositor,
             self.allocator,
-            self.renderer,
+            self.wlr_renderer,
             self.backend,
             self._subcompositor,
         ) = wlroots_helper.build_compositor(self.display)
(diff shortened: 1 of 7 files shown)


jwijenbergh:
Thanks for the pr! Just a general comment from my phone, borders for internal Windows are used in libqtile/popup.py see https://github.com/qtile/qtile/pull/4406

QBobWatson:
> borders for internal Windows are used in libqtile/popup.py

Thanks -- now I see.  I've put `Internal.paint_borders` back.  It should be cleaner this way.

jwijenbergh:
I think you need to add pywayland.utils to the docs MOCK_MODULES for the CI to work. Add it under here https://github.com/qtile/qtile/blob/551269802f4fedcbe130f139528afa2fdd3967f4/docs/conf.py#L52

QBobWatson:
Thanks—I was going to ask about that.

libqtile/backend/wayland/core.py
```diff
@@ -1231,6 +1245,9 @@ def on_config_load(self, initial: bool) -> None:
             for device in [*self.keyboards, *self._pointers]:
                 device.configure(self.qtile.config.wl_input_rules)
 
+        # Apply wl_renderer option
+        self._configure_renderer()
```
jwijenbergh, line 1249: is this needed if we already do this when subscribing to startup complete?

QBobWatson: I was imagining that the renderer could be changed by reloading config.  Does that also fire startup complete?

Shinyzenith:
Hi, gentle ping @jwijenbergh Can we get any updates on this? Apologies if I am disturbing.

jwijenbergh:
Hi! Sorry for not working on qtile for so long. I have been meaning to pickup the wlroots 0.17.0 PR but haven't had time yet. I feel like it makes more sense to get wlroots 0.17 support before we merge this.

Shinyzenith:
Thank you for the brisk response! No need to apologize for not doing free work. Hope you're doing well and I look forward to the wlroots chase!

PR #4583, opened 2023-11-18
[x11] Better stacking unmanaged windows

### LINKED ISSUE #4466: New Window Will Block Rofi
### Issue description

I tried to use Rofi while a program was loading. The new program window blocked Rofi, even though Rofi was the Window taking input.

I have a few programs that take a bit to load. A browser is a good example, but this is better seen with something that can be partially transparent, like a terminal.

Here are the steps to reproduce:

1. Make sure you use `picom` or Compton to get transparency.
2. Put the following in a script:

```
#! /bin/sh

sleep 5

exec alacritty # or some other terminal with transparency
```

3. Run `rofi -show run` and then type the full path of the script from 1.
4. While the script is in its sleep, run `rofi -show run` again. Do not type anything.
5. Wait until Alacritty or the terminal appears. You will see Rofi behind the terminal.
6. Type stuff; it will appear in Rofi. Try to run a command; it will succeed even though Rofi is behind the terminal. You can also press escape to leave Rofi, but you cannot input to the window that just appeared until Rofi is gone.

This appears to be a bug in the stacking code documented [here](https://docs.qtile.org/en/latest/manual/stacking.html) because the stacking code says that windows with `_NET_WM_STATE_ABOVE` set should be above new windows.

Rofi does have that set; its `xprop` output is the following:

```
WM_CLIENT_MACHINE(STRING) = "<redacted>"
_NET_WM_PID(CARDINAL) = 125634
WM_CLASS(STRING) = "rofi", "Rofi"
WM_NAME(STRING) = "rofi - run"
_NET_WM_NAME(UTF8_STRING) = "rofi - run"
WM_PROTOCOLS(ATOM): protocols  WM_TAKE_FOCUS
_NET_WM_STATE(ATOM) = _NET_WM_STATE_ABOVE
```

The logs below are of an exception that happens whenever a new window is opened.

### Version

0.22.1

### Backend

X11 (default)

### Config

```python
# Here is the relevant parts of my config:

dgroups_key_binder = None
dgroups_app_rules = []  # type: list
follow_mouse_focus = True
bring_front_click = False
cursor_warp = True
floating_layout = layout.Floating(
    border_focus=colors[8],
    border_width=2,
    float_rules=[
        Match(wm_class="confirmreset"),   # gitk
        Match(wm_class="dialog"),         # dialog boxes
        Match(wm_class="download"),       # downloads
        Match(wm_class="error"),          # error msgs
        Match(wm_class="file_progress"),  # file progress boxes
        Match(wm_class='kdenlive'),       # kdenlive
        Match(wm_class="makebranch"),     # gitk
        Match(wm_class="maketag"),        # gitk
        Match(wm_class="notification"),   # notifications
        Match(wm_class='pinentry-gtk-2'), # GPG key password entry
        Match(wm_class="ssh-askpass"),    # ssh-askpass
        Match(wm_class="toolbar"),        # toolbars
        Match(wm_class="Yad"),            # yad boxes
        Match(title="branchdialog"),      # gitk
        Match(title='Confirmation'),      # tastyworks exit box
        Match(title='Qalculate!'),        # qalculate-gtk
        Match(title="pinentry"),          # GPG key password entry
        Match(title="tastycharts"),       # tastytrade pop-out charts
        Match(title="tastytrade"),        # tastytrade pop-out side gutter
        Match(title="tastytrade - Portfolio Report"), # tastytrade pop-out allocation
        Match(wm_class="tasty.javafx.launcher.LauncherFxApp"), # tastytrade settings
        Match(wm_class="rofi"),
        Match(wm_class="Rofi"),
    ]
)
auto_fullscreen = True
focus_on_window_activation = "smart"
reconfigure_screens = True

# If things like steam games want to auto-minimize themselves when losing
# focus, should we respect this or not?
auto_minimize = True

# When using the Wayland backend, this can be used to configure input devices.
wl_input_rules = None
```

### Logs

```text
2023-09-23 16:34:43,279 ERROR libqtile loop.py:_handle_exception():L63 initializer for ctype 'PangoLayout *' must be a cdata pointer, not NoneType
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 871, in f
    func(*args)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 123, in real_update
    self.set_notif_text(notif)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 99, in set_notif_text
    self.text = pangocffi.markup_escape_text(notif.summary)
    ^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/widget/base.py", line 483, in text
    ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/drawer.py", line 73, in text
  File "/usr/lib/python3.11/site-packages/libqtile/pangocffi.py", line 118, in set_attributes
    pango.pango_layout_set_attributes(self._pointer, attrs)
2023-09-23 16:34:43,280 ERROR libqtile loop.py:_handle_exception():L63 initializer for ctype 'PangoLayout *' must be a cdata pointer, not NoneType
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 871, in f
    func(*args)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 123, in real_update
    self.set_notif_text(notif)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 99, in set_notif_text
    self.text = pangocffi.markup_escape_text(notif.summary)
    ^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/widget/base.py", line 483, in text
    ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/drawer.py", line 73, in text
  File "/usr/lib/python3.11/site-packages/libqtile/pangocffi.py", line 118, in set_attributes
    pango.pango_layout_set_attributes(self._pointer, attrs)
2023-09-23 16:34:43,280 ERROR libqtile loop.py:_handle_exception():L63 initializer for ctype 'PangoLayout *' must be a cdata pointer, not NoneType
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 871, in f
    func(*args)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 123, in real_update
    self.set_notif_text(notif)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 99, in set_notif_text
    self.text = pangocffi.markup_escape_text(notif.summary)
    ^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/widget/base.py", line 483, in text
    ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/drawer.py", line 73, in text
  File "/usr/lib/python3.11/site-packages/libqtile/pangocffi.py", line 118, in set_attributes
    pango.pango_layout_set_attributes(self._pointer, attrs)
2023-09-23 16:34:43,281 ERROR libqtile loop.py:_handle_exception():L63 initializer for ctype 'PangoLayout *' must be a cdata pointer, not NoneType
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
  File "/usr/lib/python3.11/site-packages/libqtile/core/manager.py", line 871, in f
    func(*args)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 123, in real_update
    self.set_notif_text(notif)
  File "/usr/lib/python3.11/site-packages/libqtile/widget/notify.py", line 99, in set_notif_text
    self.text = pangocffi.markup_escape_text(notif.summary)
    ^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/widget/base.py", line 483, in text
    ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/libqtile/drawer.py", line 73, in text
  File "/usr/lib/python3.11/site-packages/libqtile/pangocffi.py", line 118, in set_attributes
    pango.pango_layout_set_attributes(self._pointer, attrs)
```

### Required

elParaguayo:
DRAFT FOR NOW - will add some comments.

Where qtile has unmanaged windows (e.g. override redirect) then these windows were ignored for stacking purposes which meant they could appear in the wrong place.

This PR adds the ability for qtile to stack these windows correctly while still not fully managing the windows.

Fixes #4466

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -46,7 +46,6 @@ if TYPE_CHECKING:
     from typing import Callable, Iterator
 
 _IGNORED_EVENTS = {
-    # xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
     xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
@@ -591,10 +590,11 @@ class Core(base.Core):
         self, x: int, y: int, width: int, height: int, desired_depth: int | None = 32
     ) -> base.Internal:
         assert self.qtile is not None
-
         win = self.conn.create_window(x, y, width, height, desired_depth)
         internal = window.Internal(win, self.qtile, desired_depth)
         internal.place(x, y, width, height, 0, None)
+        if win.wid in self.override_redirect_map:
+            del self.override_redirect_map[win.wid]
         self.qtile.manage(internal)
         return internal
 
@@ -765,6 +765,14 @@ class Core(base.Core):
             win.change_layer()
 
     def handle_CreateNotify(self, event) -> None:  # noqa: N802
+        # Wait until qtile has started (so we don't capture root windows)
+        if self.qtile is None or not self.qtile.no_spawn:
+            return
+
+        # Internal windows are added to the windows_map before this method is run
+        if event.window in self.qtile.windows_map:
+            return
+
         xwin = window.XWindow(self.conn, event.window)
         win = window._Window(xwin, self.qtile)
         self.override_redirect_map[event.window] = win
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
     from typing import Callable, Iterator
 
 _IGNORED_EVENTS = {
-    xcffib.xproto.CreateNotifyEvent,
+    # xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
     xcffib.xproto.KeyReleaseEvent,
     # DWM handles this to help "broken focusing windows".
@@ -169,6 +169,8 @@ class Core(base.Core):
         self._last_motion_time = 0
 
         self.last_focused: base.Window | None = None
+
+        self.override_redirect_map: dict[int, window._Window] = {}
 
     @property
     def name(self):
@@ -311,7 +313,6 @@ class Core(base.Core):
                 event = self.conn.conn.poll_for_event()
                 if not event:
                     break
-
                 if event.__class__ in _IGNORED_EVENTS:
                     continue
 
@@ -725,6 +726,9 @@ class Core(base.Core):
     def handle_MapRequest(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
 
+        if event.window in self.override_redirect_map:
+            del self.override_redirect_map[event.window]
+
         xwin = window.XWindow(self.conn, event.window)
         try:
             attrs = xwin.get_attributes()
@@ -760,8 +764,16 @@ class Core(base.Core):
             self.update_client_lists()
             win.change_layer()
 
+    def handle_CreateNotify(self, event) -> None:  # noqa: N802
+        xwin = window.XWindow(self.conn, event.window)
+        win = window._Window(xwin, self.qtile)
+        self.override_redirect_map[event.window] = win
+
     def handle_DestroyNotify(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
+
+        if event.window in self.override_redirect_map:
+            del self.override_redirect_map[event.window]
 
         self.qtile.unmanage(event.window)
         self.update_client_lists()
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1168,14 +1168,22 @@ class _Window:
                 higher.sort(key=lambda wid: stack.index(wid))
 
             for wid in [w for w in lower if stack.index(w) > index]:
-                win = self.qtile.windows_map.get(wid, self.qtile.core.override_redirect_map[wid])
-                win.window.configure(stackmode=xcffib.xproto.StackMode.Below, sibling=same[0])
+                win = self.qtile.windows_map.get(
+                    wid, self.qtile.core.override_redirect_map.get(wid, None)
+                )
+                if win:
+                    win.window.configure(stackmode=xcffib.xproto.StackMode.Below, sibling=same[0])
 
             # We reverse higher as each window will be placed above the last item in the current layer
             # this means the last item we stack will be just above the current layer.
             for wid in [w for w in higher[::-1] if stack.index(w) < index]:
-                win = self.qtile.windows_map.get(wid, self.qtile.core.override_redirect_map[wid])
-                win.window.configure(stackmode=xcffib.xproto.StackMode.Above, sibling=same[-1])
+                win = self.qtile.windows_map.get(
+                    wid, self.qtile.core.override_redirect_map.get(wid, None)
+                )
+                if win:
+                    win.window.configure(
+                        stackmode=xcffib.xproto.StackMode.Above, sibling=same[-1]
+                    )
 
             return
 
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1021,6 +1021,7 @@ class _Window:
         group_windows = group.windows.copy()
         statics = [win for win in self.qtile.windows_map.values() if isinstance(win, Static)]
         group_windows.extend(statics)
+        group_windows.extend(v for w, v in self.qtile.core.override_redirect_map.items())
 
         if group.screen is not None:
             group_bars = [gap for gap in group.screen.gaps if isinstance(gap, bar.Bar)]
@@ -1050,7 +1051,6 @@ class _Window:
         lower = [w[0].wid for w in windows if w[2] > layering]
         higher = [w[0].wid for w in windows if w[2] < layering]
         same = [w[0].wid for w in windows if w[2] == layering]
-
         # We now need to identify the new position in the stack
 
         # If the window has a parent, the window should just be put above it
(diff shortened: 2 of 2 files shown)


libqtile/backend/x11/core.py
```diff
@@ -171,6 +171,8 @@ def __init__(self, display_name: str | None = None) -> None:
 
         self.last_focused: base.Window | None = None
 
+        self.override_redirect_map: dict[int, window._Window] = {}
```
elParaguayo, line 174: May be badly named as rofi (as an example) does not set the override redirect window.

libqtile/backend/x11/core.py
```diff
@@ -736,6 +737,9 @@ def handle_MappingNotify(self, event):  # noqa: N802
     def handle_MapRequest(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
 
+        if event.window in self.override_redirect_map:
```
elParaguayo, line 740: If we're mapping it then the window will be in windows_map so we should remove it here.

libqtile/backend/x11/core.py
```diff
@@ -771,9 +775,17 @@ def handle_MapRequest(self, event) -> None:  # noqa: N802
             self.update_client_lists()
             win.change_layer()
 
+    def handle_CreateNotify(self, event) -> None:  # noqa: N802
+        xwin = window.XWindow(self.conn, event.window)
+        win = window._Window(xwin, self.qtile)
```
elParaguayo, line 780: We use the `_Window` class as `Window` sets various attributes etc on `__init__` which results in errors for some windows.

libqtile/backend/x11/window.py
```diff
@@ -1164,16 +1164,14 @@ def change_layer(self, up=True, top_bottom=False):
                 higher.sort(key=lambda wid: stack.index(wid))
 
             for wid in [w for w in lower if stack.index(w) > index]:
-                self.qtile.windows_map[wid].window.configure(
-                    stackmode=xcffib.xproto.StackMode.Below, sibling=same[0]
-                )
+                win = self.qtile.windows_map.get(wid, self.qtile.core.override_redirect_map[wid])
```
elParaguayo, line 1167: Looks up the wid in windows_map and, if it's not there then it must be in the override_redirect map.

elParaguayo:
Need to check the tests too!

libqtile/backend/x11/core.py
```diff
@@ -47,7 +47,7 @@
     from libqtile.core.manager import Qtile
 
 _IGNORED_EVENTS = {
-    xcffib.xproto.CreateNotifyEvent,
+    # xcffib.xproto.CreateNotifyEvent,
```
jwijenbergh, line 50: remove it instead

jwijenbergh:
Tbf I don't really understand the details here still, but the code to me looks pretty clean

elParaguayo:
This didn't work for the OP of the issue. Back to the drawing board...

libqtile/backend/x11/core.py
tych0: I wonder if we even need a _Window? Can we keep a cache of wids here we're not supposed to touch? Seems like we should also check for the OverrideRedirect bitmask right?

tych0: Or even better, no cache at all, and if we see events for wids we don't know about, we could GetWindowAttributes for those and check the OverrideRedirect flag?

elParaguayo: OK. Help me out a bit here. How did you imagine this working?

If we listen for messages in `handle_ClientMessage`, these arrive before windows are mapped and I suspect we don't really want to get attributes every time we get a message.

Also, from a quick look, some override redirect windows (like VLC's menu) are not transients so I guess we just bring these to the front as they're closed if they lose focus. However, we'll need to track transients so that they can be stacked with their parent if other windows are stacked above them.

tych0: > I suspect we don't really want to get attributes every time we get a message

I was thinking we would, if only to avoid the two-sources-of-truth problem. Once we keep a cache, we have to remember to update it everywhere that it could be invalidated, and I suspect that will be hard.

But it's not really every time we get a message is it? We "just" want to skip over replies from QueryTree in change_layer() that are managed on their own; given that we're already doing a request for every window in the QueryTree reply to potentially adjust its stacking, I don't think it's that expensive?

We don't get any other events for these windows, so we won't accidentally touch them anywhere but here, IIUC.

> Also, from a quick look, some override redirect windows (like VLC's menu) are not transients so I guess we just bring these to the front as they're closed if they lose focus.

I was hoping (heh, famous last words) that just not modifying their stacking would be enough here. But maybe I'm wrong, I haven't played with any of this in actuality :)

elParaguayo: OK. You're going to need to spell this out for me. Are we listening for client messages or are we listening for something else? If it's not a managed window, check if it's transient and then move to top/above its parent as necessary?

tych0: I was thinking of this: https://github.com/qtile/qtile/commit/cc122253f5d7e76047869f213113b5963e79ffa4

But I don't know if that will really work. I guess the question is: if an OverrideRedirect window for an application that has been modified with StackMode.Below some other window pops up, will the application map that window at the same stacking level, or above the other window? And really, shouldn't that OverrideRedirect window probably be on top regardless of what window is actually on top, since it's presumably a dialog or whatever?

But maybe this is all my perspective. I don't use floating much, so any window that's floating for me is probably some kind of modal thing that I want to know about.

elParaguayo:
@tych0 I'm very tempted to rewrite the x11 stacking code as I think it's overly complicated at the moment.

A couple of questions for you:
- Should it be safe to assume that apps will stack override redirect windows appropriately? If not, what's the "right" way to deal with them?
- When I do `QueryTree` on the root window, I get lots of children that don't seem to reflected as actual windows. Do you have any idea what these could be?

PR #4585, opened 2023-11-19
[x11] Fix focus event on `Window.focus()`

### LINKED ISSUE #3751: window.cmd_focus() and hook.fire("focus_change") do not update border
qtile version: `0.21.0`

I am trying to activate a window through a script (the relevant part of which can be viewed [here](https://codeberg.org/surrealcereal/Qmin/src/branch/main/Qmin.py#L164)), but cannot get the window to focus with both the border and the `WindowName` widget updating.

The related keybind in the config below is pressed at the beginning of each video.

Without `hook.fire("focus_change")`, the `WindowName` does not update: [video](https://i.imgur.com/Oo9tzWE.mp4).
With it, it does; but the border does not: [video](https://i.imgur.com/wDwKfgl.mp4).

The barebones config used in this demonstration, with `hook.fire("focus_change")` being commented out for the relevant demo:
```py
from libqtile.config import Screen, Key
from libqtile.bar import Bar
from libqtile import widget, qtile
from libqtile import hook
from libqtile.command import lazy

ID = 27263032

@lazy.function
def demo(qtile):
    window = qtile.windows_map[ID]
    window.cmd_focus()
    hook.fire("focus_change")

if __name__ in ["config", "__main__"]:
    keys = [
        Key(["mod4"], "u",
            demo(),
        ),
        Key(["mod4"], "r",
            lazy.restart(),
        ),
        Key(["mod4"], "Return",
            lazy.spawn("alacritty -e /bin/zsh"),
        ),
        Key(["mod4"], "w",
            lazy.window.kill(),
        ),
    ]
    screens = [
        Screen(top=Bar(widgets=[widget.WindowName()], size=20)),
        Screen(top=Bar(widgets=[widget.WindowName()], size=20))
    ]
```
[The related reddit post and thread with elParaguayo, who instructed me to create this issue.](https://www.reddit.com/r/qtile/comments/wirt0i/move_focus_in_between_windows_gracefully_using_a/)

Edits are me fleshing out the issue.

### Required:

elParaguayo:
When calling `Window.focus()` the borders did not update on the focused window and the `WindowName` widget did not reflect the change.

Fixes #3751

I believe this works because the existing code set the current window in the group but, by doing that, it bypassed all the logic updating the layouts. Then, when the hook was called, because the window was already the `current_window` no changes were required. So, by removing the setting line, the hook then triggers all the relevant updates.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1327,11 +1327,11 @@ class _Window:
         self.qtile.core._root.set_property("_NET_ACTIVE_WINDOW", self.window.wid)
         self._ungrab_click()
 
-        if self.group:
-            self.group.current_window = self
-
         # Check if we need to restack a previously focused fullscreen window
         self.qtile.core.check_stacking(self)
+
+        if self.group and self.group.current_window is not self:
+            self.group.focus(self)
 
         hook.fire("client_focus", self)
 
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -267,8 +267,8 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
                 self.y + self._height / 2,
             )
 
-        if self.group:
-            self.group.current_window = self
+        if self.group and self.group.current_window is not self:
+            self.group.focus(self)
 
         hook.fire("client_focus", self)
 
--- a/test/test_window.py
+++ b/test/test_window.py
@@ -1,6 +1,6 @@
 import pytest
 
-from libqtile import config, layout, resources
+from libqtile import bar, config, layout, resources, widget
 from libqtile.confreader import Config
 from test.conftest import BareConfig, dualmonitor
 from test.layouts.layout_utils import assert_focused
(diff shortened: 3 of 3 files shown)


libqtile/backend/x11/window.py
```diff
@@ -1327,9 +1327,6 @@ def focus(self, warp: bool = True) -> None:
         self.qtile.core._root.set_property("_NET_ACTIVE_WINDOW", self.window.wid)
         self._ungrab_click()
 
-        if self.group:
```
jwijenbergh, line 1330: I don't really have a good mental picture right now, but it feels like they focus the window directly to avoid recursion with the group focus method. Can't we do something like this:
```python
        if self.group and self.group.current_window is not self:
            self.group.focus(self)
```

elParaguayo: Yes. I think my theory is wrong. I don't see anything subscribing to that hook.

elParaguayo: Ok - this is not the right fix. It's "working" because, after the focus changes, we get an `EnterNotify` event which then sorts out the group focus point (if `follow_mouse_focus` is set). I could use that code instead, after I've tried yours.

elParaguayo:
One small issue with `focus` is that, if the window is on a group that's not shown, the window/group will not be focused. This is because the window is `hidden` and so `self._do_focus()` won't let you focus the window.

elParaguayo:
Have added a test for this. Interestingly, it fails locally on Wayland for me so might need to fix that too if it fails here.

elParaguayo:
@jwijenbergh Look ok for a squash and merge?

jwijenbergh:
LGTM, nice!

PR #4592, opened 2023-11-26
Fix launchbar icon issue on Wayland

### LINKED ISSUE #4552: Wayland LaunchBar not rendering image
### Issue description

Using the LaunchBar widget configured like this: 
```python
widget.LaunchBar(progs=[("/home/silas/Nextcloud/uni/stundenplan.png", "mupdf /home/silas/Nextcloud/uni/stundenplan.png","stundenplan")]),
```
leads to not showing the image but a hole in my bar when using wayland (X11 works fine).

I tested with older versions of this widget. This widget works fine in Wayland before the changes in #3857 by @elParaguayo .

After revoking some changes by this pr it wokred again after replacing https://github.com/qtile/qtile/blob/1e0f09f45a4396d563039bf711ed82bdbeae8b36/libqtile/widget/launchbar.py#L291
with
```python
self.drawer.draw(offsetx=self.offset + xoffset, offsety=self.offsety, width=self.width)
```
Seems to work fine in X11 as well but I'm not sure if this may lead to other problems or if this is an appropriate fix for this bug as the X11 _version_ seems to render this differently than the wayland _version_

### Version

0.23.0

### Backend

Wayland (experimental)

### Config

_No response_

### Logs

_No response_

### Required

elParaguayo:
Icons did not render correctly on Wayland when supplying path to the icon.

Fixes #4552

### DIFF
--- a/libqtile/widget/launchbar.py
+++ b/libqtile/widget/launchbar.py
@@ -272,19 +272,12 @@ class LaunchBar(base._Widget):
                 )
             else:
                 # display an icon
-                self.drawer.ctx.save()
+                # Translate to vertically centre the icon
                 self.drawer.ctx.translate(0, self._icon_padding + self.padding_y)
                 self.drawer.ctx.set_source(self.surfaces[name])
                 self.drawer.ctx.paint()
-                self.drawer.ctx.restore()
 
             self.drawer.ctx.restore()
-
-            self.drawer.draw(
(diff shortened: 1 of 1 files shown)


elParaguayo:
Confirmed as working here: https://github.com/qtile/qtile/issues/4552#issuecomment-1829470780

PR #4593, opened 2023-11-26
Docs: Move to furo theme

jwijenbergh:
Let's try this

also mentioned in https://github.com/qtile/qtile/discussions/4548


jwijenbergh:
Closed in favor of #4612

PR #4598, opened 2023-11-27
Update FAQ section on sticking groups to screens

00000vish:
Updating the sticking groups to screen code section so it can be copy pasted directly into the default config without errors and changes.


ramnes (approved):
why not :)

elParaguayo:
Should we be encouraging users to use typing? If so, should we just add the relevant imports instead?

Issue #4602, opened 2023-11-27
Window without a class opens on the first group


nazar-pc:
### Issue description

I just noticed that when window doesn't have a class, it opens on the first group instead of current:
```
_NET_WM_STATE(ATOM) = 
_NET_FRAME_EXTENTS(CARDINAL) = 1, 1, 1, 1
_VARIABLE_REFRESH(CARDINAL) = 1
_NET_WM_DESKTOP(CARDINAL) = 0
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified size: 1280 by 960
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
WM_CLIENT_MACHINE(STRING) = "nazar-pc"
_NET_WM_PID(CARDINAL) = 1424202
WM_CLASS(STRING) = "", ""
XdndAware(ATOM) = BITMAP
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x2, 0x0, 0x1, 0x0, 0x0
_NET_WM_NAME(UTF8_STRING) = "Welcome - Iced"
WM_NAME(STRING) = "Welcome - Iced"
```

I do have window matching rules in the config, but nothing that matches empty string. This is an interesting edge-case I have not seen before.

### Version

0.21.0

### Backend

X11 (default)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

jwijenbergh:
I assume you're using iced-rs, can you share the code?

nazar-pc:
Yes, I just tried this example: https://github.com/iced-rs/iced/tree/master/examples/tour

Setting window class with this helped to open window on current group:
```rust
    let mut settings = Settings::default();
    settings.window.platform_specific.application_id = "test".to_string();
    Tour::run(settings)
```

jwijenbergh:
Please give https://github.com/qtile/qtile/pull/4603 a spin, it seems like we do some very weird matching that I fail to understand

Issue #2670, opened 2021-07-31
Match on WM_CLASS matches substrings


Sawrz:
# Issue description

It seems the `Match` class checks if a sub-string exists in in a given `wm-class` string. For example:

Sublime Text's `WM_CLASS` is `subl`. Now, I want to have a intrusive float match for `Ssh-askpass-sublime`. However adding that match makes Sublime Text an intrusive floating windows as well. 

I tried different combinations, like `sublome` but the result is the same. If I understood the source code correctly you use a lambda function to compare two strings with the `in` keyword. However, I didn't dig to much into the code so I  might be wrong. 

Is that on purpose? 

# Qtile version

`0.18.1.dev0+g8e7ecc0a.d20210719 `
Seems to be Version 0.18.0 according to releases tab

# Configuration

[My configuration file](https://gitlab.com/lazylinuxuser/dotfiles/-/blob/main/qtile/.config/qtile/config.py)

yobleck:
Yes the Match class does appear to be using substring matching [here](https://github.com/qtile/qtile/blob/a3dcd5db984f3ab08ef3f89eff86e014dd367ee1/libqtile/config.py#L605)

This is just a temporary hack but you could try:
```
@hook.subscribe.client_managed
    def my_func(window)
        window.window.get_wm_class() #or get name or something
        #some logic to parse the correct window the way you want
        qtile.cmd_spawn("xdotool search --name "Title of App" set_window --class "New WM Class" --classname "New WM Class Name")
        #requires xdotool from your package manager
```

([xdotool man page](https://manpages.ubuntu.com/manpages/trusty/man1/xdotool.1.html))

Sawrz:
Thank you @yobleck!

m-col:
The bit from the docs about that "func" argument: "delegate the match to the given function, which receives the tested client as argument and must return True if it matches, False otherwise". In the function you provide you can specify the logic yourself, so for example something like this might testing your askpass window and only that window (untested):
```py
mymatch = Match(func=lambda w: w.get_wm_class() == "Ssh-askpass-sublime")
```

Sawrz:
@m-col I saw that argument as well. I didn't used it for two reasons:
1. I wasn't sure how to use it
2. I don't know if this behavior, as implemented, is intended

Anyway, no harm in trying it out. However, it doesn't seem to work. To check it's validity I tried to force `subl` in my match to see what happens. The result is: Nothing. No intrusive or floating behavior, so we seem to miss something here.

[...]

PR #4603, opened 2023-11-28
Config: Do not match on substrings

jwijenbergh:
For some reason in the Match class we match on substrings. So e.g. for wm_class if value = ["test1", "bogus"] and our candidate is "test1what" we would loop through the list and determine that the test passes because test1 is a substring of test1what. This leads to confusion and bugs such as #4602 and #2670.

In the current situation even the default config is pretty broke on X11 with the float rules (e.g. wm_class dialogpleasedontfloat would match dialog and thus result in a floating window).


nazar-pc:
Hm, I can see how both substrings and full matches would be useful.

For https://github.com/qtile/qtile/issues/4602 the fix is to only do exact match in case of an empty string.

Otherwise I think having both `wm_class` and `wm_class_exact` would be a solution to all edge-cases.

elParaguayo:
My understanding was that you can use regexes in situations where substring matching was undesired.

jwijenbergh:
> Hm, I can see how both substrings and full matches would be useful.
> 
> For #4602 the fix is to only do exact match in case of an empty string.
> 
> Otherwise I think having both `wm_class` and `wm_class_exact` would be a solution to all edge-cases.

It's either we remove this or we change the default config (with ^dialog$) but many people think matching actually matches without a substring, I think it's not very intuitive at all and we should have people use regexes

elParaguayo:
It's definitely not intuitive. I've helped on several issues related to this over the years!

jwijenbergh:
Note btw it's not substring matching on what you specify, it's the other way around

Setting  Match(wm_type="dialog"), matches wm_types "dial"

nazar-pc (approved):
Okay, considering `func` exists, I think it is fine to use exact match for class and other things by default

jwijenbergh:
> Okay, considering `func` exists, I think it is fine to use exact match for class and other things by default

Yeah indeed I have added this note to the changelog too. I could indeed fix it only for the empty case but I don't think it makes sense if I specify in my config
```python3
# default float rules
        Match(wm_class="firefoxsomegarbage"), 
```

firefox would float, I guess I could have more explained it like superstring matching. I was getting a bit confused by how it works

ramnes:
much better indeed, thanks

PR #4614, opened 2023-12-03
Docs: small changed in suggested keybindings in MonadTall layout

holykremowka2137:
i think its good idea to also add 'reset' option by default to suggested keybindings

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -147,7 +147,8 @@ class MonadTall(_SimpleLayoutBase):
         Key([modkey, "shift"], "k", lazy.layout.shuffle_up()),
         Key([modkey], "i", lazy.layout.grow()),
         Key([modkey], "m", lazy.layout.shrink()),
-        Key([modkey], "n", lazy.layout.normalize()),
+        Key([modkey], "n", lazy.layout.reset()),
+        Key([modkey, "shift"], "n", lazy.layout.normalize()),
         Key([modkey], "o", lazy.layout.maximize()),
         Key([modkey, "shift"], "space", lazy.layout.flip()),
     """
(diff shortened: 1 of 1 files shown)


libqtile/layout/xmonad.py
```diff
@@ -147,7 +147,8 @@ class MonadTall(_SimpleLayoutBase):
         Key([modkey, "shift"], "k", lazy.layout.shuffle_up()),
         Key([modkey], "i", lazy.layout.grow()),
         Key([modkey], "m", lazy.layout.shrink()),
-        Key([modkey], "n", lazy.layout.normalize()),
+        Key([modkey], "n", lazy.layout.reset()),
+        Key([modkey, "shift"], "n", lazy.layout.normalize()),
```
elParaguayo, line 151: Why change the previous binding? Couldn't you add reset on shift+n, rather than changing the normalize binding?

holykremowka2137: 'n' just feel more natural for this, instead of Shift + n. I consulted several other ppl abt this and they agreed with me, so thats why i did it like this. But if changed previous keybinding is out of question, how abt smth like this? 

```python
Key([modkey], "r", lazy.layout.reset()),
Key([modkey], "n", lazy.layout.normalize()),
```

elParaguayo: `mod + r` is already a binding in the default config so stick with your original idea.

However, this should be added to `MonadWide` as well.

holykremowka2137: Done. Dont mind these "Merge branch [...]" commits. I had several problems pushing this and from what I tried and read that was the only solution. Hopefully I didnt effed up something. Git is confusing

elParaguayo:
Thanks. I've set this to merge once the tests complete.

PR #4621, opened 2023-12-11
Let the GPU rescale the wallpaper on Wayland

QBobWatson:
This PR changes the behavior of `backend.wayland.wlrq.Painter`.  The wallpaper image is uploaded directly to the GPU.  Any scaling or stretching is then performed by the GPU's sampling hardware.  This has two advantages:

- It's way faster than scaling in software using cairo.
- More importantly, wallpapers now work on hidpi displays.

As for the second point: if the user sets a scale factor on a monitor (using `wlr_output.set_scale(2)`, for instance), then any wallpaper will probably have twice the resolution of the Screen associated to the Output.  Without this PR, the background image would be downsampled and then upsampled again, resulting in a blurry, low-res version.

With this PR, if you use `wallpaper_mode` equal to `fill` or `stretch`, then the wallpaper will be sampled in its original resolution and will fill the screen.  This seems to be the simplest solution, as it's not clear to me how to get a `wayland.output.Output` from a Screen object.  This has the effect, however, that if you set `wallpaper_mode = None` then the image size is used for the wallpaper size _in layout coordinates_, which doesn't take into account any scale factor.

### DIFF
--- a/libqtile/backend/wayland/wlrq.py
+++ b/libqtile/backend/wayland/wlrq.py
@@ -27,6 +27,7 @@ from typing import TYPE_CHECKING, cast
 
 import cairocffi
 from pywayland.server import Listener
+from wlroots import ffi as wlr_ffi, lib as wlr_lib
 from wlroots.wlr_types import Buffer, SceneBuffer, SceneTree, data_device_manager
 from wlroots.wlr_types.keyboard import KeyboardModifier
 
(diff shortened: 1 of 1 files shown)


jwijenbergh:
LGTM but will actually test it later

tych0:
ping @jwijenbergh any chance you can give this a whirl?

jwijenbergh:
Whoops, sorry. Works fine for me, let's merge

PR #4624, opened 2023-12-14
Load xcursor scale factors.

QBobWatson:
The XCursorManager needs to be told to load different scale factors of the cursors when outputs are scaled.  Otherwise, `cursor_manager.set_cursor_image()` has no effect.


jwijenbergh:
LGTM

Issue #3646, opened 2022-06-20
Solving internal wayland scaling problems.


oddlama:
### The issue:

I've been investigating the [blurry `Bar`](https://github.com/qtile/qtile/discussions/3400#discussioncomment-2290380) issue on wayland, as it started bothered me. As always I'm happy to provide a PR for all of what I'm proposing. But first, I require some details and your input on how to solve the deeper underlying problem.

### Blurry Bar

So, why exactly is the Bar blurry when using scaled wayland outputs? Because the internal window (that is used to render the bar) is just a cairo texture based on the _effective_ resolution. On 2x scale, the effective resolution of the window is 1920x24 instead of 3480x48. Halfed due to the scaling. But the display area, to which the window is transformed, is still 3480x48 which results in upscaling the texture and therefore making it blurry.

This is an easy fix though: Just for testing I've changed the code to create a proper full-size texture and increased the resolution of the corresponding `Drawer`. But this solution is _conceptually wrong_ in my opinion.

### The Underlying Problem

It all boils down to qtile always operating on the effective resolution of the output. If the output is scaled by a factor of `2` on a UHD display, qtile regards it as a Full-HD 1920x1080 display instead. All internal window positions and sizes are interpreted in terms of effective resoultion. I think this is wrong. as qtile knows the native resolution of any output and should have the ability to discern between the two.

The current behavior has some bad side-effects:
- Blurry bar. If qtile knew the actual resolution, the internal windows would automatically use a texture of correct size.
- Windows can only be positioned on the "effective pixel grid". If I have a big beamer with 5x scaling, the windows can currently only be positioned and resized in increments of 5 pixels for no valid reason.
- Gaps and borders are currently specified in "effective pixels". I really want a way to have "single-native-pixel" borders.

### Possible Solution: Units

So instead, qtile should always look at outputs in terms of their native resolution (the resolution provided by the output's current mode). Generally I think the correct solution to this problem is to have units for all spatial values. Internally, everything is converted to native pixels to ensure everything stays "pixel-perfect". I'm thinking about supporting various units such as `px` (native pixel), `dp` (density independent pixel), `dpi` (dots per inch), `mm` (millimeters, actual screen size is known on wayland), `cm` (centimeters), `in` (inches). If no unit is given, we assume effective pixels, to ensure backwards compatibility.

[...]

PR #4625, opened 2023-12-14
Blurry bar: quick fix

QBobWatson:
As noted in #3646, qtile currently renders internal windows (e.g. the Bar) at layout resolution and not output resolution, which makes them blurry on hidpi monitors.  That issue suggests an elegant but complex solution.  This PR implements a dumb solution: a config option that determines what scale the internal windows are rendered at.

I'm not claiming this is the best solution, or even a good one.  But it's trivial to implement and will satisfy most use cases.


jwijenbergh:
Could we make the scale factor an output variable? Or better yet, can't we just get the scale from the wlr_output pointer?

QBobWatson:
> Could we make the scale factor a config.Screen variable? Or better yet, can't we just get the scale from the wlr_output pointer?

The relationship between Screens and outputs can be complicated with fake screens and all.  This makes per-output configuration kind of problematic.

I think we’re supposed to be listening for the `output_enter` event of the underlying `wlr_scene_buffer` in order to keep track of which outputs the window is displayed on.  Then we’re supposed to render to a surface scaled up by the largest of the outputs’ scale factors (or maybe the least common multiple).  This is more or less what Wayland clients do.

Although that might be the correct thing to do, it would be a pain, and the advantage is minimal.  It would just mean that if you had more than one monitor, with one scaled up and one not, and you had a bar on each monitor, then you don’t have to render a scaled-up version of the bar on the low dpi monitor.  I’m not so interested in writing that PR at the moment, since I’m just trying to get qtile/Wayland working on my laptop.

frosklis:
I fixed the pre-commit hook failures here: https://github.com/qtile/qtile/pull/4744

richcarni:
@elParaguayo I would like to see if I can get this PR merged. I am not the author of this or #4744, but I have contacted @QBobWatson and have access to the relevant repo/branch

I have fixed the pre-commit failures and also made a small modification so that it also supports fractional scaling

I'm not sure if you still want to try and merge this PR (it is some 300 commits behind master now and I can't get the CI to pass without re-basing off master) or if I should create a fresh PR?

Happy to be guided by you (or any of the maintainers)

Thanks,
Richard

jwijenbergh:
> @elParaguayo I would like to see if I can get this PR merged. I am not the author of this or #4744, but I have contacted @QBobWatson and have access to the relevant repo/branch
> 
> I have fixed the pre-commit failures and also made a small modification so that it also supports fractional scaling
> 
> I'm not sure if you still want to try and merge this PR (it is some 300 commits behind master now and I can't get the CI to pass without re-basing off master) or if I should create a fresh PR?
> 
> Happy to be guided by you (or any of the maintainers)
> 
> Thanks, Richard

I have reopened it for you

richcarni:
I've been doing some more thinking/experimenting around getting the scale from wlr_output (as suggested by @jwijenbergh)

I have a working proof of concept but would be great to get some feedback before I go too far down a wrong path...

Starting in `Qtile._process_screens()`,  I've modified `Core.get_screen_info()` to also get `wlr_output.scale`.  I've added a `scale` parameter to `Screen`, which `Qtile._process_screens()` sets/updates for each screen

`Bar` now it has access to `screen.scale`

Then I pass `screen.scale` to `Internal` when creating the bar window

And the rest is basically as per the existing PR

I've only done some rudimentary testing so far. It works on my single monitor setup. It also updates scaling dynamically, when I change it with wlr-randr, which I thought was nice

My test branch: https://github.com/richcarni/qtile/commit/99eb9236326b993ed2e1cd7beff774d6f73fc6d2

richcarni:
> > I've been doing some more thinking/experimenting around getting the scale from wlr_output (as suggested by @jwijenbergh)
> > I have a working proof of concept but would be great to get some feedback before I go too far down a wrong path...
> > Starting in `Qtile._process_screens()`, I've modified `Core.get_screen_info()` to also get `wlr_output.scale`. I've added a `scale` parameter to `Screen`, which `Qtile._process_screens()` sets/updates for each screen
> > `Bar` now it has access to `screen.scale`
> > Then I pass `screen.scale` to `Internal` when creating the bar window
> > And the rest is basically as per the existing PR
> > I've only done some rudimentary testing so far. It works on my single monitor setup. It also updates scaling dynamically, when I change it with wlr-randr, which I thought was nice
> > Relevant commits on my dev branch (https://github.com/richcarni/qtile/commits/wayland-hidpi-dev/): [richcarni@1fe94f8](https://github.com/richcarni/qtile/commit/1fe94f84f5a3063794fc544191f1158a34c06b88) [richcarni@a13b3f6](https://github.com/richcarni/qtile/commit/a13b3f689a4c46e650c6483fd074740be76e3b33) [richcarni@35f02f7](https://github.com/richcarni/qtile/commit/35f02f78e156e0cedfbe98f5a898cfc8ba771501)
> > (Popups and tree layout appear to be the other use of internal windows... I haven't given these much thought yet - I imagine they should work the same as the bar?)
> 
> Can you make a pr?

Sure! You mean as a clean alternative (rather than adding to) this one?

richcarni:
> > * It didn't look exactly the same as if i'd simply scaled the whole bar - maybe not a big deal
> 
> What does that mean? What was different?

3way_compare_2

Top is using Cairo to scale everything as per this PR. Middle is me multiplying bar height, font size and padding by 1.5 - I realise you can also adjust GroupBox margins and spacing, and increase StatusNotifier icon_size, but this is what I was referring to. Bottom, for reference, is what unpatched qtile currently looks like for me with Wayland scaling of 1.5

> > * I think full control means implementing and applying scale() to all/relevant elements. Is this something authors or individual widgets will also need to think about?
> 
> I'm ok with this as the idea would be to have the scaling code happen "behind the scenes" so widget authors should not notice any change. There will be some unavoidable issues (e.g. where widgets override methods) so we'd need documentation to be very clear on how the scaling works.
> 

> > Cons:
> > 
> > * Pixmaps don't scale nicely, i.e. they are blurry - we should be able to find a workaround
> > * May or may not be difficult to also apply in X11
> 
> I think these cons are significant. As above, I don't mind different solutions for X11 and Wayland as long as there's some consistent interface for writing the code. In my opinion, blurry pixmaps should be avoided at all costs.
 
> > Honestly, I don't know if there is any right approach!
> 
> It's a balance between complexity of implementation vs result. I'd prefer to put the effort in to get the best output we can.

In my view, this PR (or similar) is still worthwhile. A couple of blurry icons is a big improvement over blurry everything

But if this just serves as a stand in, and there is enthusiasm around a more comprehensive solution, that would be cool...

I was actually thinking about the practicality of extending the Drawer class to essentially wrap all Cairo calls with scale-aware versions. It seems to me something like this would come closest to a no compromise solution

elParaguayo:
Thanks for the screenshots. It's clear that this PR is an improvement on the current position.

If we merge it, I suspect it's a temporary fix until we can implement something that would work for both backends but I have no objections to merging it now.

However, I'll defer to @jwijenbergh as he's the wayland expert here. If he's ok with this then we should merge it.

richcarni:
> > I've been doing some more thinking/experimenting around getting the scale from wlr_output (as suggested by @jwijenbergh)
> > I have a working proof of concept but would be great to get some feedback before I go too far down a wrong path...
> > Starting in `Qtile._process_screens()`, I've modified `Core.get_screen_info()` to also get `wlr_output.scale`. I've added a `scale` parameter to `Screen`, which `Qtile._process_screens()` sets/updates for each screen
> > `Bar` now it has access to `screen.scale`
> > Then I pass `screen.scale` to `Internal` when creating the bar window
> > And the rest is basically as per the existing PR
> > I've only done some rudimentary testing so far. It works on my single monitor setup. It also updates scaling dynamically, when I change it with wlr-randr, which I thought was nice
> > Relevant commits on my dev branch (https://github.com/richcarni/qtile/commits/wayland-hidpi-dev/): [richcarni@1fe94f8](https://github.com/richcarni/qtile/commit/1fe94f84f5a3063794fc544191f1158a34c06b88) [richcarni@a13b3f6](https://github.com/richcarni/qtile/commit/a13b3f689a4c46e650c6483fd074740be76e3b33) [richcarni@35f02f7](https://github.com/richcarni/qtile/commit/35f02f78e156e0cedfbe98f5a898cfc8ba771501)
> > (Popups and tree layout appear to be the other use of internal windows... I haven't given these much thought yet - I imagine they should work the same as the bar?)
> 
> Can you make a pr?

@jwijenbergh Created PR #5159

Made the commits quite small. Not sure if that is annoying - I thought it made it easier to follow - can squash if that is preferred or maybe you already do that on merge

PR #4626, opened 2023-12-15
revised/new patch for improve ratio calculation

alexander-n8hgeg5e:
Here is the description about why a change can be considered:
[description right at the beginning of this closed pull request](https://github.com/qtile/qtile/pull/4601)

Now, I tested a function that actually works and made screenshots.
It works with upright and landscape ratios and
also with my "ratio-1.0-square" use case.
Ratios > 1 are "landscape" and ratios < 1 are "upright".

It works like this:
[the code from below more readable](https://gist.github.com/alexander-n8hgeg5e/29d27fecf7f78ec7fc7c0cc589b82b25)
`
def is_more_golden_than( a, b , gs=1):                                                          
    goldstandard = gs                                                                           
    ao=a                                                                                        
    bo=b                                                                                        
    # TODO: use math to refactor the next 6 lines                                               
    a = a / goldstandard                                                                        
    b = b / goldstandard                                                                        
    wa =  a if a < 1 else -1/a                                                                  
    wb =  b if b < 1 else -1/b                                                                  
    wa = abs(abs(wa)-1)                                                                         
    wb = abs(abs(wb)-1)                                                                         
    judgement = True if wa < wb else False                                                      
    print(f'{ao:>5.3} vs {bo:<5.3} > {wa:+.3f}, {wb:+.3f}  golden = {gs:0.2f}   {judgement}'  ) 
    return judgement                                                                            
                                                                                                
is_more_golden_than(1.5 , 0.66666   , gs=1  )                                                   
is_more_golden_than(2.0 , 2.1       , gs=2  )                                                   
is_more_golden_than(1.9 , 2.0       , gs=2  )                                                   
is_more_golden_than(1.9 , 2.1       , gs=2  )                                                   
is_more_golden_than(0.25, 0.75      , gs=0.5)                                                   
                                                                                                
  output:                                                                                       
    1.5 vs 0.667 > +0.333, +0.333  golden = 1.00   True                                         
    2.0 vs 2.1   > +0.000, +0.048  golden = 2.00   True                                         
    1.9 vs 2.0   > +0.050, +0.000  golden = 2.00   False                                        
    1.9 vs 2.1   > +0.050, +0.048  golden = 2.00   False                                        
   0.25 vs 0.75  > +0.500, +0.333  golden = 0.50   False                                        
`
update here: pictures removed, newer pictures below


alexander-n8hgeg5e:
For the failing test_ratiotile.py  test,
maybe I should make a script that pops up windows in the ratios from the test and the new ones to check
if it makes sense. Could be viewed "zoomed out" side-by side.
Can I update pull requests if they are fixed or do I need to close it ?

elParaguayo:
Just keep adding commits to your branch.

alexander-n8hgeg5e:
I got the data out and generated some side by side pictures which are only 5k each.
Unfortunately I don't know the ratio that is used in this test. I guess it is 0.5 in the "add windows" test because the next test named "test_ratiotile_add_windows_golden_ratio"
does call manager.c.next_layout() and this layout is the second one in a list made above inside 
"class RatioTileConfig(Config)".
The resize test is maybe 1.618 (golden ratio),...


Here is the code I used to generate it:
[branch_in_my_repo](https://github.com/alexander-n8hgeg5e/qtile/tree/hacks/ratio_tile_test_data_extract)
I temporary disabled all the other test because I believe they passed.

The test "add windows" seems to produce reasonable results. 0.5 means upright double the high as the width. Hm, now I notice I don't know whether there is averaging in case there are different ratios in the grid (the case with 5 windows). It should consider all the windows and not switch because of one good one.
 
The other failed test "resizing", probably needs more
resize steps/commands for the switching to happen.
When I tested it with real windows it did switch.
I'll update this here.
On the left is the v0.23.0 and on the right is 
my patch. I added a divider in between the screens.
The test screen size seems to be 800x600.

# failed test: "add windows"
img_add_windows_0
img_add_windows_1
img_add_windows_2
img_add_windows_3
img_add_windows_4
img_add_windows_5
img_add_windows_6
img_add_windows_7
img_add_windows_8
img_add_windows_9
img_add_windows_10
img_add_windows_11

# failed test: "resizing"
It does not switch with this test.
img_resizing_0
img_resizing_1
img_resizing_2

alexander-n8hgeg5e:
I rebased and force pushed it to clean it up.
It needs probably a more cleanup after the code analysis tools
inspected it.

alexander-n8hgeg5e:
This here is everything summarized and more organized.
This is a new set of pictures based on the rebased commits of the pull request.
For the pull request I had to adjust the "add_windows" test and the "resizing" test.
Here are 2 branches in my repo that explain the generation of the pictures and show the way they are derived from the pull request. For the pictures of the "add windows" test I had to remove a commit from the pull request because, to make the test pass, the commit changes the values of two "layout-infos" and the original values are needed for the side-by-side view.
# For the "add_windows" test:
[branch that made the pictures for the "add_windows" test](https://github.com/alexander-n8hgeg5e/qtile/tree/pull_request_1702648090.output_pictures.test_add_windows)
The commit "output window data from test_add_windows" shows that the previous expected values remain in the code and are used for the generation of the side-by-side pictures.
## side-by-side pictures for the "add_windows" test
img_add_windows_0
img_add_windows_1
img_add_windows_2
img_add_windows_3
img_add_windows_4
img_add_windows_5
img_add_windows_6
img_add_windows_7
img_add_windows_8
img_add_windows_9
img_add_windows_10
img_add_windows_11


# For the resizing test:
[branch that made the pictures for the resizing test](https://github.com/alexander-n8hgeg5e/qtile/tree/pull_request_1702648090.output_pictures.test_resizing)

## pictures for the "resizing" test
img_resizing_0
img_resizing_1
img_resizing_2
img_resizing_3
img_resizing_4
img_resizing_5
img_resizing_6

elParaguayo:
Can you explain, in as few words as possible, what the issue is, what your PR does and what the images above show? I'm finding this quite hard to follow but I appreciate you've put a lot of effort in here.

alexander-n8hgeg5e:
> Can you explain, in as few words as possible, what the issue is, what your PR does and what the images above show? I'm finding this quite hard to follow but I appreciate you've put a lot of effort in here.

The issue is, the layout calculation does not make the windows as close to the desired ratio as possible.
This PR would improve it.
Why it works? I'm not sure I know it myself but it turned out it works.
The pictures of the "add_windows" test show that it did not change that much in this test.
The pictures of the other test (resizing) are intended to show that the test, which I completely changed, makes sense (hopefully) and are meant to read together with this commit (one picture is taken at every "assert sizes()" statement):
[commit](https://github.com/qtile/qtile/commit/3d38991ade1bacff7e97054fd45c0254a19e2e62)

### The issue in detail:
If the desired ratio=1 and there are 3 windows on a screen that is 1.125 higher than wide,
then the result is 3 windows 1.125 tall and 0.333 wide. (screen width=1)
Then the ratio is 1.125/(1/3)=3.375
But it would be better to have 3 "landscape" windows because they would be 1.125/3 in height and 1 wide, which is better because if you see the ratio as longer_edge_of_window/shorter_edge_of_window then it is 1/(1.125/3)= 2.666 (closer to 1)
(note: the ratios in this description are inverse(1/x) to the ones in class ratioTile)

I know that this is not the most important issue and it requires some time to supervise something like this.

PR #4628, opened 2023-12-15
Don't bring static windows to front.

QBobWatson:
The Wayland backend raises a Static window when `config.bring_front_click == "floating_only"`.  The x11 backend does not (as far as I can tell), and this doesn't seem like it would be the desired behavior.


m-col (approved):
Good spot, thanks!

PR #4629, opened 2023-12-15
Implement move_to_{top,bottom} in Wayland

QBobWatson:
This adds `move_to_top()` and `move_to_bottom()` to Wayland windows.  The `bring_to_front()` command had already been implemented, and these do essentially the same thing.  (Before there was no command to lower a Wayland window.)

### DIFF
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -197,7 +197,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
         """
 
     @expose_command()
-    def move_to_top(self) -> None:
+    def move_to_top(self, force: bool = False) -> None:
         """
         Move this window above all windows in the current layer
         e.g. if you have 3 windows all with "keep_above" set, calling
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for all the pull requests!

I'm not getting involved in these, though, as they're Wayland focused so I'll leave it to the experts.

m-col:
Thanks for the PR. @elParaguayo what are the differences in semantics between `bring_to_front` and `move_to_top`? Their implementations differ in the X backend.

elParaguayo:
The definitions in `base.window` have the same signature as these methods. However, in X11 we have an additional, optional keyword argument `force`. This is is used where a window is tagged as `kept_above/below`. Forcing disables that tag and then moves the window.

Can Wayland implement `keep_above/below` too?

QBobWatson:
> Can Wayland implement keep_above/below too?

The analogous functionality is provided by the layer shell.  This is implemented in the sense that clients can request to be placed on different layers using the layer shell protocol, but currently qtile can't place windows on a specified layer.  I'd have to look through the protocol to figure out how that's supposed to work.  In any case, it's beyond the scope of this PR.

elParaguayo:
That's fine. However, given the x11 backend allows the `force` argument to the `move_to_top/bottom` methods, we should allow them here so there's no config breakages between backends. We can then add the other functionality later. 

@m-col @jwijenbergh any concerns with that approach?

PR #4631, opened 2023-12-17
Update build process - make Wayland build optional

elParaguayo:
This is an attempt to fix some issues that users are having with building qtile and not wanting the Wayland backend.

The PR removes `pywlroots` from the build dependencies and uses a custom build backend to inject the requirement if the wayland backend is specifically requested.

The builder will also raise an error if the backend is requested but `pywroots` is not installed. This is useful for `--no-build-isolation` flag as that builds against pre-installed libraries.

Comments welcome.

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -1,0 +1,54 @@
+# Copyright (c) 2023, elParaguayo. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import sys
+
+from setuptools import build_meta as _orig
+from setuptools.build_meta import *  # noqa: F401,F403
+
+WAYLAND_DEPENDENCIES = ["pywlroots>=0.16.4,<0.17.0"]
+
+
+def wants_wayland(config_settings):
+    if config_settings:
+        for key in ["Backend", "backend"]:
+            if config_settings.get(key, "").lower() == "wayland":
+                return True
+
+    return False
+
+
+def get_requires_for_build_wheel(config_settings=None):
+    """Inject pywlroots into build dependencies if wayland requested."""
+    reqs = _orig.get_requires_for_build_wheel(config_settings)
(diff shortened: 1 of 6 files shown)


builder.py
```diff
@@ -0,0 +1,54 @@
...
+from setuptools import build_meta as _orig
+from setuptools.build_meta import *  # noqa: F401,F403
+
+WAYLAND_DEPENDENCIES = ["pywlroots>=0.16.4,<0.17.0"]
+
+
+def wants_wayland(config_settings):
+    if config_settings:
+        for key in ["Backend", "backend"]:
+            if config_settings.get(key, "") in ("Wayland", "wayland"):
```
jwijenbergh, line 31: config_settings.get(key, "").lower() == "wayland"

jwijenbergh: same for backend maybe?

jwijenbergh: or maybe we should just make the key --wayland?

elParaguayo: Not possible. There are different builders (`pip`, `python -m build` etc) so the only standard way to pass arguments is the config-settings as that's part of the spec.

elParaguayo: > or: config_settings.get(key, "").lower() == "wayland"

Ha. I thought about this and went with the other option.

jwijenbergh: > Not possible. There are different builders (`pip`, `python -m build` etc) so the only standard way to pass arguments is the config-settings as that's part of the spec.

ah yeah I meant --config-setting wayland or --config-setting --wayland (but that would probably not work). Let's leave it like this and maybe do the lower thing but that's optional

jwijenbergh:
if only there was a neater way to have optional build requires. But this seems like the way to go https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-time-dependencies

if CI passes LGTM

jwijenbergh (approved):

cr1ogen:
If you set the installation of wayland as optional, how should Qtile be compiled or installed for those who use wayland as a backend?

elParaguayo:
> If you set the installation of wayland as optional, how should Qtile be compiled or installed for those who use wayland as a backend?

Good question. This is my main concern with this PR as it complicates things a little bit.

Ideally, you should have all the dependencies installed already. In that case, passing the `--no-build-isolation` flag should be enough to let the wayland backend be compiled against your current libraries.

If you don't do that, you can pass `--config-option backend=wayland` which will inject `pywlroots` into the dependencies and it will trigger a failure if it can't build the backend.

cr1ogen:
Couldn't you make an installation script that asks if you want to install the wayland backend?

elParaguayo:
Short answer: no, I don't think so.

What's your main concern? To have Wayland working, you need all dependencies installed beyond just python packages. This change doesn't change anything to do with that but it does remove an unnecessary dependency being installed for people using the x11 backend. 

We can list what packages need to be installed on a user's system but I don't think we should be providing any form of installation script.

jwijenbergh:
I don't really see the problem here, it's just a single flag and would only apply when building from source. Easy to fix in packaging

cr1ogen:
> Short answer: no, I don't think so.
> 
> What's your main concern? To have Wayland working, you need all dependencies installed beyond just python packages. This change doesn't change anything to do with that but it does remove an unnecessary dependency being installed for people using the x11 backend. 
> 
> We can list what packages need to be installed on a user's system but I don't think we should be providing any form of installation script.

I don't have any problem because I know how to do it, but it would be good for the less experienced, at least who wanted to be a window manager only for developers

jwijenbergh:
> > Short answer: no, I don't think so.
> > What's your main concern? To have Wayland working, you need all dependencies installed beyond just python packages. This change doesn't change anything to do with that but it does remove an unnecessary dependency being installed for people using the x11 backend.
> > We can list what packages need to be installed on a user's system but I don't think we should be providing any form of installation script.
> 
> I don't have any problem because I know how to do it, but it would be good for the less experienced, at least who wanted to be a window manager only for developers

Maybe an install script makes sense that just follows the steps that we document.

elParaguayo:
How do you cater for all the different distros/package managers?

jwijenbergh:
> How do you cater for all the different distros/package managers

I mean a script that runs the necessary pip commands, nothing more

cr1ogen:
> > How do you cater for all the different distros/package managers
> 
> I mean a script that runs the necessary pip commands, nothing more

That's what I meant, sorry if I couldn't explain it.

tych0 (approved):

tych0:
Any reason not to land this?

PR #4632, opened 2023-12-17
Fix Bluetooth log spamming

elParaguayo:
Bluetooth widget can spam the logs when device awakens from sleep and resources are not immediately available.

### DIFF
--- a/libqtile/widget/bluetooth.py
+++ b/libqtile/widget/bluetooth.py
@@ -42,11 +42,14 @@ def _catch_dbus_error(msg):
     """Decorator to catch DBusErrors and log a message."""
 
     def _wrapper(func):
-        async def f(self):
+        async def f(self, *args, **kwargs):
             try:
-                await func(self)
+                await func(self, *args, **kwargs)
             except DBusError:
                 logger.warning(msg, self.name)
+            except BlockingIOError:
+                logger.debug("DBus resource is temporarily unavailable.")
+                pass
 
         return f
 
@@ -118,6 +121,7 @@ class BluetoothDevice(_BluetoothBase):
         await self.interface.call_pair()
         await self.connect()
 
+    @_catch_dbus_error("Unable to call action: %s.")
     async def action(self):
         """Helper method to call appropriate method based on device status."""
         if self.connected:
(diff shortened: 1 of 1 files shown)


libqtile/widget/bluetooth.py
```diff
@@ -42,11 +42,14 @@ def _catch_dbus_error(msg):
...
     def _wrapper(func):
-        async def f(self):
+        async def f(self, *args, **kwargs):
             try:
-                await func(self)
+                await func(self, *args, **kwargs)
             except DBusError:
                 logger.warning(msg, self.name)
+            except BlockingIOError:
+                # Resource is temporarily unavailable
```
ramnes, line 51: Let's use a `logger.debug` here?

elParaguayo: Yes.

There are still some scenarios that aren't being caught (usually when laptop has been asleep and then resumed) so I need to catch those too and test some more.

tych0 (approved):

ramnes:
Is this the kind of spam that will get fixed?

```
2024-01-18 10:52:56,347 ERROR libqtile loop.py:_handle_exception():L62 Exception in event loop:
  File "libqtile/widget/bluetooth.py", line 277, in update_props
    self.widget.refresh()
  File "libqtile/widget/bluetooth.py", line 582, in refresh
    self.show_line()
  File "libqtile/widget/bluetooth.py", line 617, in show_line
    self.update(self.format_object(obj))
  File "libqtile/widget/base.py", line 754, in update
    old_width = self.layout.width
                ^^^^^^^^^^^^^^^^^
  File "libqtile/backend/base/drawer.py", line 421, in width
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libqtile/pangocffi.py", line 137, in get_pixel_size
    pango.pango_layout_get_pixel_size(self._pointer, width, height)
```

elParaguayo:
@ramnes in a word: no.

That's an issue with textbox based widgets and it drives me insane. I thought I'd fixed it many times but there are some cases still getting through. I'll look at that again.

The bluetooth spamming is to do with the bluez interface not immediately being ready after resume (more annoyingly, some interfaces take longer than other) and so some of the calls fail.

I'm not sure I'll have time to find and patch all the cases before the weekend (which is when @tych0 wants to do the next release). However, I can make the `logger.debug` change that you suggested and then we can merge this one as it still has some improvements over the current one.

PR #4636, opened 2023-12-23
Fix TreeTab draw resolution issue

### LINKED ISSUE #4634: TreeTab not drawing menu correctly on multimonitor setups
### Issue description

I have a two monitor setup with different resolutions. 
One of my groups uses the TreeTab layout. Depending on which monitor it is first created on this layout has a gap in the menu region as you can see in this screenshot (the higher resolution monitor is on the left)

### Version

0.23.0

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

elParaguayo:
When a TreeTab layout moves between screens with different resolutions, the Drawer object is not updated to reflect the new size. This means that rendering issues can arise when moving from a lower resolution screen to a larger resolution.

Fixes #4634

### DIFF
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -733,6 +733,12 @@ class TreeTab(Layout):
         self.group.layout_all()
 
     def _create_drawer(self, screen_rect):
+        # Create a new drawer object if the screen is a different height
+        # e.g. if moving between screens with different resolutions
+        if self._drawer is not None and self._drawer.height != screen_rect.height:
+            self._drawer.finalize()
+            self._drawer = None
+
         if self._drawer is None:
             self._drawer = self._panel.create_drawer(
                 self.panel_width,


libqtile/layout/tree.py
```diff
@@ -752,6 +758,7 @@ def layout(self, windows: Sequence[base.Window], screen_rect: ScreenRect) -> Non
         Layout.layout(self, windows, body)
 
     def _resize_panel(self, screen_rect):
+        print("Resize Panel")
```
jwijenbergh, line 761: Remove this, otherwise looks good

elParaguayo: I need to add my print statement checker as a pre-commit hook. I have it on qtile-extras too prevent annoying things like this!

jwijenbergh:
LGTM

Etn40ff:
Sorry for the late reply, thanks for the patch.
LGTM

PR #4638, opened 2023-12-29
Fix code error in the Command Graph docs page

VelocityREST:
Fix a minor typo where the function `toscreen` is named `to_screen`


elParaguayo (approved):
Thanks. Looks good.

PR #4641, opened 2024-01-06
Allow DnD to override implicit grabs.

QBobWatson:
This fixes a bug (that I introduced) in https://github.com/qtile/qtile/pull/4540.

A client does not expect to retain an implicit grab when it initiates a drag-and-drop.


ftpd:
Fixes https://github.com/qtile/qtile/pull/4540

jwijenbergh:
Thanks!

PR #4645, opened 2024-01-14
[x11 stacking] filter unknown windows

elParaguayo:
There have been two recent reports of users seeing an error where stacking fails because a window cannot be found in the X server's stack. This shouldn't happen but let's filter out unknown windows.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1040,6 +1040,9 @@ class _Window:
             )
         )
 
+        # Remove any windows that aren't in the server's stack
+        windows = list(filter(lambda w: w[0].wid in stack, windows))
+
         # Sort this list to match stacking order reported by server
         windows.sort(key=lambda w: stack.index(w[0].wid))
 


libqtile/core/manager.py
```diff
@@ -436,6 +436,7 @@ def paint_screen(self, screen: Screen, image_path: str, mode: str | None = None)
 
     def process_key_event(self, keysym: int, mask: int) -> tuple[Key | KeyChord | None, bool]:
         key = self.keys_map.get((keysym, mask), None)
+        print(key)
```
jwijenbergh, line 439: remove

elParaguayo: Dammit...

jwijenbergh:
Nice. LGTM. The only problem I have with this is that the stacking code loops through the windows a bunch of times using list comprehensions. We could probably combine a couple of those into a normal for loop. But maybe that is something for later

elParaguayo:
Agreed. I'll merge for now and look at refactoring later.

jwijenbergh:
No worries, thanks for the quick fix :)

PR #4650, opened 2024-01-19
Docs: Small fixes in documentation

serweryn617:
Doc fixes:
- Properly escape role in config.py
- Fix decorator name in client_mouse_enter hook
- Fix invisible code block in Backlight widget

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -404,10 +404,10 @@ class ScreenRect:
 
 
 class Screen(CommandObject):
-    """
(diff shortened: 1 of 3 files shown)


elParaguayo:
Nice. Thanks for this. 

I'll take a quick look over the weekend but expect it's fine to merge

serweryn617:
I also expected it to be fine, but pre-commit failed because of the escape sequence I used.

PR #4652, opened 2024-01-20
Add Touchpad widget

serweryn617:
A widget that shows touchpad state and allows to toggle it with mouse click or keybind.

What are your thoughts?


libqtile/widget/touchpad.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if self.device_id is None:
+            device_id = self._get_touchpad_device_name()
+        if self.get_state_func is None:
+            get_state_func = self._get_touchpad_enabled
+        if self.set_state_func is None:
+            set_state_func = self._set_touchpad_enabled
+
+        new_state = not get_state_func(device_id)
+        set_state_func(device_id, new_state)
```
elParaguayo, line 119: You check `self.get_state_func` twice. Once here, and once in `poll`. You could just do it once in `_configure`.

Also, if `self.get_state_func` is not `None` then `get_state_func` is undefined which means line 118 would fail.

Same issue for `set_state_func`.

libqtile/widget/touchpad.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def _get_touchpad_device_name() -> str:
+        properties = subprocess.check_output(["xinput", "list", "--name-only"], text=True)
+
+        for line in properties.splitlines():
+            if "touchpad" in line.lower():
+                return str(line)
+
+        raise RuntimeError("Could not determine touchpad name automatically!")
+
```
elParaguayo, line 131: Do these all need to be `staticmethod`s?

These could just be functions outside of the class

elParaguayo (changes requested):
Thanks for this. I don't have any particular objections to including in the main repo but have made some comments.

I would like to include a test for this widget, particularly to test the use of custom `get/set_state_func` functions.

libqtile/widget/touchpad.py
serweryn617: Right, thanks for spotting that

libqtile/widget/touchpad.py
serweryn617: I'm fine with that

libqtile/widget/touchpad.py
serweryn617: Done. I also had to add some type asserts to keep mypy happy

serweryn617:
I've added unit tests, updated everything and tested if it works. Everything looks good.

elParaguayo:
You need to ignore the widget in this test:
https://github.com/qtile/qtile/blob/2b5ad3c5976052199adf154555384919c6d0aba2/test/widgets/test_widget_init_configure.py#L60-L62

serweryn617:
Looks like it's only a `subprocess` call that fails because it requires `xinput`.
I can also just add an override for the `device_id` parameter so this call is never executed.
I don't know what would be a better option.

elParaguayo:
If you can get it to work with the override, that's better.

serweryn617:
Should be good now. Are there any other tests that run on all widgets? I only checked the ones I've added, not the entire suite

elParaguayo:
One other thought, does this work in Wayland?

serweryn617:
The default functions probably not. I don't really have a way to test it.

elParaguayo:
Ok. Pinging @jwijenbergh to see if he knows if it can be done on Wayland. 

If not, you can set `supported_backends = {"x11"}` as a class attribute of your widget.

serweryn617:
Is there anything else I should add/change @elParaguayo ?

PR #4653, opened 2024-01-21
Add :meth:`Spiral.left` and :meth:`Spiral.right` to allow moving left/right on Spiral Layout

JasonGrace2282:
The title explains it pretty well what I tried to accomplish. 
Hopefully I can pull out the pieces to reduce the code redundency, and if there are any (better) ways to get things like the layout or the currently focused coordinates please let me know! It's my first time with the codebase so I'm not 100% familiar with everything.

## Algorithm Explanation
I'll explain the left side only, as the right side is almost the exact same.
Consider a spiral like the following
A spiral layout
The way we decide what the window to the left of it is is by ranking the distances of the x coordinate of each window to the x coordinate of the current window. For example for the green window the shortest distance is $1440-960=480$.
Now there are three windows with an x distance of $480$, so we sort them by the *absolute value* of the differences in the $y$ coordinates. The window with the shortest distance in both the $x$ and $y$ directions is the one that gets chosen. In this case the shortest distance in the $y$ direction is obviously black, as they have the same $y$ value.

## Future PRs
Maybe in future PRs a similar idea could be applied to up/down (instead of the current `next`/`previous` renames we seem to have) and there could be some configuration for the way windows are selected based on their $y$ value.

### DIFF
--- a/libqtile/layout/spiral.py
+++ b/libqtile/layout/spiral.py
@@ -378,7 +378,10 @@ class Spiral(_SimpleLayoutBase):
         coords = info["layout_info"]
         focused_x, focused_y, *_ = coords[info["current"]]
 
-        min_x_diff = min(x - focused_x for x, *_ in coords if x > focused_x)
+        diffs = tuple(x - focused_x for x, *_ in coords if x > focused_x)
+        if not diffs:
+            return
+        min_x_diff = min(diffs)
         # find the windows that are a minimal x distance away
         min_x = [
             (idx, x, y) for idx, (x, y, *_) in enumerate(coords) if x - focused_x == min_x_diff
--- a/libqtile/layout/spiral.py
+++ b/libqtile/layout/spiral.py
@@ -343,6 +343,54 @@ class Spiral(_SimpleLayoutBase):
         d["clockwise"] = self.clockwise
         return d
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
I'll take a proper look later but my initial reaction is that the code looks overly complicated. There's a lot of duplicate code so you may be able to break some of that out into common methods.

However, more importantly, I'm not sure how good a solution this is. The reason for my hesitation is that `Spiral` already has `up` and `down` methods (they're not listed in the docs which is a different issue). These move up and down the windows but following the order of the spiral. As a result, `left` and `right` really serve no purpose here.

There is also a more generic `up`, `down`, `left`, `right` solution available for all layouts here: https://github.com/qtile/qtile-examples/blob/master/traverse.py

JasonGrace2282:
Aw man it would've been amazing if I knew about that example before xD

To some extent I do agree with you about `up` and `down`, but I personally found it easier to think of the screen in vertical/horizontal components rather then spiral so I will keep using it locally (after, of course, trying out that example). If you don't think it's particularly needed, feel free to close this PR.

elParaguayo:
Sorry about that. I completely forgot about the `traverse` script when this came up on Discord earlier.

Issue #4656, opened 2024-01-23
On a config reload, a custom layout that manages internal windows can face a 'dict changed size during iteration' error


aravinda0:
### Issue description

I'm building a custom layout that allows for nested tabs, and hence nested/dynamic tab bars. These tab bars are created dynamically as needed (eg. when the user asks for a subtab), and are basically invocations of `qtile.core.create_internal()`.

I've provided a minimal demo of such a layout below in the `config` section.

Now, when a config-reload happens, the code goes to `Core.on_config_load()` [here on x11](https://github.com/qtile/qtile/blob/2b5ad3c5976052199adf154555384919c6d0aba2/libqtile/backend/x11/core.py#L244) and [here on wayland](https://github.com/qtile/qtile/blob/2b5ad3c5976052199adf154555384919c6d0aba2/libqtile/backend/wayland/core.py#L1206). In either case, there is an iteration over all windows under qtile:

```
            for win in self.qtile.windows_map.values():
                if type(win) is window.Window:
                    win.set_group()
```

But work is only done for 'managed' windows. As part of that work, the managed windows are added to the various groups and their layout instances. At that point, if there is some dynamic internal window creation in the layout, we face:

`RuntimeError: dictionary changed size during iteration`

since internal window gets added to `qtile.windows_map`, which is presently being iterated on.

I figure a simple solution would be to simply iterate over prefetched list of managed windows:
```
        managed_wins = [
            w for w in self.qtile.windows_map.values() if isinstance(w, window.Window)
        ]
        for win in managed_wins:
            win.set_group()
```

I'm creating a PR to go with this.

### Version

0.24

### Backend

X11 (default), Wayland (experimental)

### Config

```python
from libqtile.backend.base.window import Window
from libqtile.config import ScreenRect
from libqtile.layout.base import Layout


class BorkedTabLayout(Layout):
    def __init__(self):
        super().__init__()

        self._wins = []
        self._focused_win = None
        self._bar_win = None
        self._bar_drawer = None
        self._bar_h = 50

    def add_client(self, client: Window) -> None:
        self._wins.append(client)

        # Creating bar here just to demonstrate possible dynamic creation of internal
        # windows.
        self._maybe_make_dynamic_bar()

    def remove(self, client: Window) -> Window | None:
        self._wins.remove(client)

    def configure(self, client: Window, screen_rect: ScreenRect):
        client.place(
            screen_rect.x,
            screen_rect.y + self._bar_h,
            screen_rect.width,
[...]

PR #4657, opened 2024-01-23
fix window iteration on config reload to only look at managed windows

aravinda0:
Fixes https://github.com/qtile/qtile/issues/4656

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -1203,10 +1203,10 @@ class Core(base.Core, wlrq.HasListeners):
 
         assert self.qtile is not None
 
-        for win in self.qtile.windows_map.values():
-            if not isinstance(win, window.Window):
-                continue
-
+        managed_wins = [
+            w for w in self.qtile.windows_map.values() if isinstance(w, window.Window)
+        ]
+        for win in managed_wins:
             group = None
             if win.group:
                 if win.group.name in self.qtile.groups_map:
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for this.

Just to make sure I completely understand, is the issue that your layout creates new internal windows during that loop over the windows map?

aravinda0:
> Thanks for this.
> 
> Just to make sure I completely understand, is the issue that your layout creates new internal windows during that loop over the windows map?

Yup, that's right.
I create internal windows under certain conditions inside `Layout.add_client()`. 
During config reload, qtile loops over the windows map and a call chain goes something like `win.set_group() --> group.add(self) --> layout.add_client(win)`,  which creates an internal window, modifying windows map, leading to an exception.

elParaguayo:
Great. There are a couple of tiny fixes you need to resolve the `check` failure here:
```diff
diff --git a/libqtile/backend/x11/core.py b/libqtile/backend/x11/core.py
index 88d52042..391a5215 100644
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -242,12 +242,10 @@ class Core(base.Core):
         if not initial:
             # We are just reloading config
             managed_wins = [
-                w
-                for w in self.qtile.windows_map.values()
-                if isinstance(w, window.Window)
+                w for w in self.qtile.windows_map.values() if isinstance(w, window.Window)
             ]
-            for win in managed_wins:
-                win.set_group()
+            for managed_win in managed_wins:
+                managed_win.set_group()
             return
```
If you make those changes, I'm happy to merge this.

aravinda0:
I've made the fixes. My bad - hadn't noticed the formatter failure.

PR #4663, opened 2024-01-25
Delete non-persistent groups when windows are moved to another group

### LINKED ISSUE #4654: Non-persistent Groups are still alive when its last window is moved to another group
### Issue description

# Issue
When a group is created with the option `persist=False` and its last window is moved to another group (instead of being kill inside that group), the non-persistent group is still alive.

## Expected behavior
A non-persistent group should be removed when it no longer contains any window and _not_ when its last window is **specifically killed inside the group**. 

## Steps to reproduce:

1. Create a group with the following configuration:
    ```python
    groups = [
        # Some property to match
        Group("a", matches=[Match(wm_class="foo")], exclusive=True),
        # other groups
    ]
    ```
2. Open a window in group `"a"` that doesn't match any of the group's matches and doesn't match any other group.
3. A new group will be created with the name of the window as the last group.
4. Move this new group to the current screen.
5. Send the (only) window to any other group.
6. The new group is still alive, even though it doesn't have any windows and it's not persistent.

The group can still be remove either through the commands API or, as a workaround, by opening a new window in that group and then killing it while the window is still in the group.

The problem seems to be that the `DGroups._del()` method at `/libqtile/dgroups.py` it's only run when the `client_kill` event is fired, so groups are only ever consider for removal when windows are killed/closed but not when windows are only removed from them, which matches the previously described behavior.

> NOTE: I've only test it X11

Is this intended behavior?

### Version

0.24.0

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

ValdezFOmar:
Fix #4654

Save a reference to the original group before removing it from the client and set it back before calling `DGroups._del(client)` so it can grab the reference and schedule it for deletion.

Check if the client has already been scheduled for deletion in order to avoid scheduling the same client multiple times and generating a `KeyError`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Qtile x.xx.x, released XXXX-XX-XX:
     * features
     * bugfixes
+      - Fix groups marked with `persist=False` not being deleted when their last window is moved to another group.
 
 Qtile 0.24.0, released 2024-01-20:
     !!! config breakage/changes !!!
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -301,7 +301,17 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
             if self.group.screen:
                 # for floats remove window offset
                 self.x -= self.group.screen.x
+            group_ref = self.group
             self.group.remove(self)
+            # delete groups with `persist=False`
+            if (
+                not self.qtile.dgroups.groups_map[group_ref.name].persist
+                and len(group_ref.windows) <= 1
+            ):
+                # set back original group so _del() can grab it
+                self.group = group_ref
+                self.qtile.dgroups._del(self)
+                self.group = None
 
         if group.screen and self.x < group.screen.x:
             self.x += group.screen.x
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -2009,7 +2009,16 @@ class Window(_Window, base.Window):
             if self.group.screen:
                 # for floats remove window offset
                 self.x -= self.group.screen.x
+            group_ref = self.group
             self.group.remove(self)
+            if (
+                not self.qtile.dgroups.groups_map[group_ref.name].persist
+                and len(group_ref.windows) <= 1
+            ):
+                # set back original group so _del() can grab it
+                self.group = group_ref
+                self.qtile.dgroups._del(self)
+                self.group = None
 
         if group.screen and self.x < group.screen.x:
             self.x += group.screen.x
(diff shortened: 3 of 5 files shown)


elParaguayo:
Thanks for doing this. I'd really like to include a test in the PR to check it's working as intended. Are you able to add that?

CHANGELOG
```diff
@@ -1,6 +1,7 @@
 Qtile x.xx.x, released XXXX-XX-XX:
     * features
     * bugfixes
+      - Fix groups marked with `persist=False` not being deleted when their last window it's moved to another group.
```
elParaguayo, line 4:
```suggestion
      - Fix groups marked with `persist=False` not being deleted when their last window is moved to another group.
```

elParaguayo (changes requested):

ValdezFOmar:
Sure, no problem, I can add a test. Where would it me more appropriate to add the test? I think under `test/backend/<backend>/test_window.py`, but I'm not sure.

elParaguayo:
I'd think `test/test_group.py`. This isn't backend specific and we're checking group functionality.

elParaguayo:
We try not to use `time.sleep` in our tests but there is a helper you can use:
```diff
diff --git a/test/test_group.py b/test/test_group.py
index aae57ecc..ff360a10 100644
--- a/test/test_group.py
+++ b/test/test_group.py
@@ -22,13 +22,13 @@
 # SOFTWARE.

 import random
-import time

 import pytest

 import libqtile.config
 from libqtile import layout
 from libqtile.confreader import Config
+from test.helpers import Retry


 class GroupConfig(Config):
@@ -115,6 +115,10 @@ class NoPersistGroupConfig(GroupConfig):

 @pytest.mark.parametrize("manager", [NoPersistGroupConfig], indirect=True)
 def test_non_persistent_groups(manager):
+    @Retry(ignore_exceptions=(AssertionError,))
+    def wait_for_removed(group_name):
+        assert group_name not in manager.c.get_groups()
+
     window_name = "no_match"
     manager.c.group["b"].toscreen()
     manager.test_window(window_name)
@@ -124,13 +128,11 @@ def test_non_persistent_groups(manager):
     manager.c.group[group_name].toscreen()
     assert manager.c.window.info()["name"] == window_name
     manager.c.window.togroup("a")
-    time.sleep(1)  # Wait for the group to be removed
-    assert window_name not in manager.c.get_groups()
+    wait_for_removed(window_name)

     window_name = "bar"
     manager.c.group["c"].toscreen()
     manager.test_window(window_name)
     assert manager.c.window.info()["name"] == window_name
     manager.c.window.togroup("a")
-    time.sleep(1)  # Wait for the group to be removed
-    assert "c" not in manager.c.get_groups()
+    wait_for_removed("c")
```

elParaguayo (approved):

PR #4667, opened 2024-01-27
Different format for pomodoro text output

pepbutler:
Change the pomodoro time text to always include two digits for the second and minute because it's easier to read and follows convention.


elParaguayo:
Nice. Looks good to me.

PR #4669, opened 2024-01-27
Use deepcopy to clone widgets

### LINKED ISSUE #4668: Icons used in cloned LaunchBars do not dynamically change their size when `icon_size` isn't set
### Issue description

I cloned a LaunchBar for use on my multi-monitor setup, and drew bars for each monitor, with the main bar a different size from the secondary bars.

Resulting behavior: the icons on the secondary bar are sized appropriately, and the icons on the main bar match the size of the icons on the secondary bar.

Secondary bar:

Main bar (whose height is 2x the height of the secondary bar):

Expected behavior: the size of the icons fit to the size of their respective bars.

Steps to reproduce:
1. Create a qtile environment with at least two available screens.
2. Create a LaunchBar widget that loads an image for the icon.  For example, a simple Discord launch button like `LB_test = widget.LaunchBar(progs=['discord', 'discord', ''])` should suffice.
3. Draw two bars in `screens` where one bar has a different size from the second one, and a clone of the above widget is inserted in place of the original.  For example, `screens = [Screen(bottom=bar.Bar([LB_test], 56)), Screen(bottom=bar.Bar([LB_test.clone()], 28))]`

If needed, I can include a version of my configuration file where the issue is produced.

### Version

0.23.0

### Backend

X11 (default)

### Logs

_No response_

### Required

elParaguayo:
Widgets' `clone` method used `copy.copy`. This causes an issue where widgets defined list and dicts during `__init__` as these are not copied and, instead, the same object is shared between widget instances. Using `copy.deepcopy` should mean these objects are also copied.

Fixes #4668

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -400,7 +400,7 @@ class _Widget(CommandObject, configurable.Configurable):
         return Mirror(self, background=self.background)
 
     def clone(self):
-        return copy.copy(self)
+        return copy.deepcopy(self)
 
     def mouse_enter(self, x, y):
         pass


tych0:
>  defined list and dicts during __init__ 

I think the first time I hit this bug using python was more than 20 years ago. Glad to know it's still around :)

PR #4677, opened 2024-02-01
Move bring_to_front to _Window

elParaguayo:
Moving `bring_to_front` to the `base._Window` class. Two reasons:
1) Avoid code duplication
2) Allow Popups (which are Internal windows) to utilise `bring_to_front`.

@jwijenbergh I haven't touched Wayland code. Let me know if we should change it too.


elParaguayo:
Any reason not to merge this one?

tych0:
I don't think so -- seems like the wayland implementation already had this?

PR #4678, opened 2024-02-02
stop accepting sponsorships

m-col:
Unfortunately I've had limited time to commit to Qtile lately so I'm going to take my name down from the sponsor button.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,0 @@
-github: ["m-col"]
-custom: ["https://liberapay.com/mcol"]


elParaguayo:
Ridiculous. I'm not in favour of this.

m-col:
It's only fair, there are other maintainers (yourself included) more deserving, and really it's been a few months since I've made any substantial contribution. But eventually I'll get back on the horse :)

elParaguayo:
There are so many people benefiting from the efforts you put in. The fact that you're currently less active now is irrelevant.

m-col:
Hmm. Maybe you're right. I have some guilt about accepting money when I'm not contributing, but the code I made is still there and still working. At least let me buy you a fancy cheese or something :)

PR #4680, opened 2024-02-03
Parse mpris2 text before display

elParaguayo:


tych0:
I'm beginning to wonder if we should have a more systematic way to do this...

elParaguayo:
> I'm beginning to wonder if we should have a more systematic way to do this...

We should!

The problem is that we can't escape all characters as there may be pango markup there too which needs to be parsed correctly. So we need to escape the text before it's formatted by the widget's format parameter. Very messy!

tych0:
I wonder if python can help us here somehow with a custom formatter? I'll poke around.

elParaguayo:
Probably, yes. I used a custom formatter for the Mpris2 widget already but for a different reason.

PR #4681, opened 2024-02-04
[wayland] Add environment variable to configure cursor size

plusgut:
At X11 the way to configure the cursor_size is done by configuring XCURSOR_SIZE https://man.archlinux.org/man/Xcursor.3#XCURSOR_SIZE

There sadly is no standardized way to do that in wayland, the weston-example compositor also does it by XCURSOR_SIZE https://man.archlinux.org/man/weston.1.en#XCURSOR_SIZE
https://gitlab.freedesktop.org/wayland/weston/-/blob/master/clients/window.c#L1338

I added the same approach to qtile with this pr

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -108,6 +108,8 @@ if TYPE_CHECKING:
     from libqtile.core.manager import Qtile
 
 
+DEFAULT_CURSOR_SIZE = 24
+
 class ImplicitGrab(wlrq.HasListeners):
     """Keep track of an implicit pointer grab.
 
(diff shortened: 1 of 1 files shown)


jwijenbergh:
Sure, makes sense. But what about making it a config option instead?

plusgut:
@jwijenbergh sure I can do that. Should it be a wayland-specific option? How to tell the X11 backened programmatically the cursor size is not obvious to me, and I would need a couple pointers.

jwijenbergh:
Yes I think prefixing it with wl_ is a good idea

jwijenbergh:
This has been added in the wlr 0.17 pr with a wl_xcursor_size variable in the config

PR #4685, opened 2024-02-11
battery: add "fancy" charge control

tych0:
Inspired by this talk at FOSDEM:

https://fosdem.org/2024/schedule/event/fosdem-2024-2123-from-kernel-api-to-desktop-integration-how-do-we-integrate-battery-charge-limiting-in-the-desktop/

I wrote some "smart charge" logic for qtile. It probably needs better detection of other kinds of docking stations, since all I have is a thunderbolt one. But at least this is a start.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -42,6 +42,7 @@ from subprocess import CalledProcessError, check_output
 from typing import TYPE_CHECKING, NamedTuple
 
 from libqtile import bar, configurable, images
+from libqtile.command.base import expose_command
 from libqtile.images import Img
 from libqtile.log_utils import logger
 from libqtile.utils import send_notification
@@ -70,6 +71,8 @@ BatteryStatus = NamedTuple(
         ("percent", float),
         ("power", float),
         ("time", int),
+        ("charge_start_threshold", int),
+        ("charge_end_threshold", int),
     ],
 )
 
@@ -170,7 +173,39 @@ class _FreeBSDBattery(_Battery):
         else:
             raise RuntimeError("Could not get remaining battery time!")
 
-        return BatteryStatus(state, percent=percent, power=power, time=time)
+        return BatteryStatus(
+            state,
+            percent=percent,
+            power=power,
+            time=time,
+            charge_start_threshold=0,
+            charge_end_threshold=100,
+        )
+
+
+def connected_to_thunderbolt():
+    try:
+        sysfs = "/sys/bus/thunderbolt/devices"
+        entries = os.listdir(sysfs)
+        for e in entries:
+            try:
+                name = Path(sysfs, e, "device_name").read_text()
+            except FileNotFoundError:
+                continue
+            else:
+                logger.debug("found dock %s", name)
+                return True
+    except OSError:
+        logger.debug("failed to detect thunderbot %s", exc_info=True)
+    return False
+
+
+def thunderbolt_smart_charge() -> tuple[int, int]:
+    # if we are thunderbolt docked, set the thresholds to 40/50, per
+    # https://support.lenovo.com/us/en/solutions/ht078208-how-can-i-increase-battery-life-thinkpad-and-lenovo-vbke-series-notebooks
+    if connected_to_thunderbolt():
+        return (40, 50)
+    return (0, 90)
 
 
 class _LinuxBattery(_Battery, configurable.Configurable):
@@ -190,6 +225,20 @@ class _LinuxBattery(_Battery, configurable.Configurable):
             "power_now_file",
             None,
             "Name of file with the current power draw in /sys/class/power_supply/battery_name",
+        ),
+        (
+            "charge_controller",
+            None,
+            """
+            A function that takes no arguments and returns (start, end) charge
+            thresholds, e.g. ``lambda: (0, 90)``; set to None to disable smart
+            charging.
+            """,
+        ),
+        (
+            "force_charge",
+            False,
+            "Whether or not to ignore the result of charge_controller() and charge to 100%",
         ),
     ]
 
@@ -214,6 +263,7 @@ class _LinuxBattery(_Battery, configurable.Configurable):
         self.add_defaults(_LinuxBattery.defaults)
         if isinstance(self.battery, int):
             self.battery = "BAT{}".format(self.battery)
+        self.charge_threshold_supported = True
 
     def _get_battery_name(self):
         if os.path.isdir(self.BAT_DIR):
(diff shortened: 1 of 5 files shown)


tych0:
One problem here is I'm not really sure how to test any of this. I suppose we could write a fake sysfs? But even that won't catch the udev rules not being set up.

elParaguayo:
Nice!

I'd wanted to do this but my laptop doesn't have `charge_control_{start,end}_threshold`.

tych0:
Do you have the (legacy, IIUC) `charge_{start,stop}_threshold`? I think those are roughly the same files.

elParaguayo:
No. Don't even have those. It's a pretty old laptop now. Dell XPS13 9350.

tych0:
ah, yeah, looks like nobody has implemented it for dell laptops:
```
~/packages/linux/drivers/platform/x86 (HEAD detached at vfs/vfs.pidfd) git grep -l charge_control
asus-wmi.c
huawei-wmi.c
lg-laptop.c
msi-ec.c
msi-ec.h
system76_acpi.c
thinkpad_acpi.c
toshiba_acpi.c
```

I think (?) it could be made to work if you're interested. We would have to look for the datasheet of the battery and see if that supports it first, then write all the ACPI controls.

elParaguayo:
Well, I'm interested but there's no way I have the skills to code that!

elParaguayo:
Battery is not in a great state and Dell don't seem to make it any more...

```
        Manufacture Date: 11/16/2015
        Serial Number: 0B68
        Name: DELL JHXPY53
```

tych0:
Nice, and can you put the output of `sudo acpidump -b` somewhere? (Or just e-mail it to me.)

elParaguayo:
I just get an error with that ("could not open output file")

elParaguayo:
Ignore that. Will email now.

libqtile/widget/battery.py
```diff
@@ -267,7 +277,64 @@ def _get_param(self, name) -> tuple[str, str]:
...
+            with open(path, "w+") as f:
+                f.write(str(end))
+        except FileNotFoundError:
+            self.charge_threshold_supported = False
+        except OSError:
+            self.charge_threshold_supported = False
+            logger.exception("Failed to write %s (did you set up udev rules?)", path)
+
     def update_status(self) -> BatteryStatus:
+        self.set_battery_charge_thresholds()
```
elParaguayo, line 337: Should there be an option to toggle whether we want smart charging enabled or not, rather than just with the `qtile cmd-obj` commands?

elParaguayo (commented):
Even though I can't test this, it doesn't mean we shouldn't include it.

Have added some questions for you.

elParaguayo:
Also, should there be some indicator to show what charge mode we're in?

libqtile/widget/battery.py
```diff
@@ -191,6 +192,12 @@ class _LinuxBattery(_Battery, configurable.Configurable):
             None,
             "Name of file with the current power draw in /sys/class/power_supply/battery_name",
         ),
+        (
+            "default_max_charge",
```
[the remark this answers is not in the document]
tych0: Yeah, I thought about it. I wonder if we shouldn't make the charging algorithm itself pluggable, in light of your questions below. Then we can do something by default here (probably 90/0 like it defaults to now), and people would plug in their own custom thing.

libqtile/widget/battery.py
```diff
@@ -267,7 +277,64 @@ def _get_param(self, name) -> tuple[str, str]:
...
+    def set_battery_charge_thresholds(self):
+        if not self.charge_threshold_supported:
+            return
+
+        start = 0
+        end = self.default_max_charge
+
+        # if we are thunderbolt docked, set the thresholds to 40/50, per
+        # https://support.lenovo.com/us/en/solutions/ht078208-how-can-i-increase-battery-life-thinkpad-and-lenovo-vbke-series-notebooks
+        if self.connected_to_thunderbolt():
```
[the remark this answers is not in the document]
tych0: I spend most of my normal workday plugged into a thunderbolt dock, so I mostly don't need to charge there, and want to do the optimal thing for battery health.

Generally, if I'm plugged into an AC adapter I'm traveling, and there I definitely *do* want to fully charge. So (for me, at least) it's a fairly automatic distinction of when I want to force charge vs not.

libqtile/widget/battery.py
tych0: Perhaps just letting people set the default value of force_charge?

tych0:
> Also, should there be some indicator to show what charge mode we're in?

I wondered about this too, but couldn't decide a good rendering for it. Maybe `49% (m40% M50%)`?

tych0:
> I wondered about this too, but couldn't decide a good rendering for it. Maybe 49% (m40% M50%)?

Ok, answering my own question, I put them in BatteryStatus, so now people can use `{charge_[start,end]_threshold}` in their format strings if they want.

Also, I think I will make the charging algorithm programmable, and we can do the default as `lambda: (0, 90)`, which will let people define their own.

Then I wonder if my code for thunderbolt dock detection etc. should live in the qtile tree, or in my config tree? Feels like nobody will ever find it if it's just in my config tree, but you have qtile-extras that's popular, maybe it could live there?

tych0:
Let me know what you think of the way it looks now. I left the thunderbolt stuff here because I didn't want to lose it, but I'm happy to move it to qtile-extras if that's what makes sense to you.

libqtile/widget/battery.py
```diff
@@ -191,6 +226,17 @@ class _LinuxBattery(_Battery, configurable.Configurable):
             None,
             "Name of file with the current power draw in /sys/class/power_supply/battery_name",
         ),
+        (
+            "charge_controller",
```
elParaguayo, line 230: I think you need to reference this in the docstring i.e. give examples of how to use it (including the `thunderbolt_smart_charge` function which I would keep in this module).

You may also want to be clear that the default behaviour (where this is supported) is to charge to 90% - given that this is a change to current behavior.

libqtile/widget/battery.py
```diff
@@ -267,7 +314,37 @@ def _get_param(self, name) -> tuple[str, str]:
...
+        except OSError:
+            logger.debug("Failed to write %s", path, exc_info=True)
+        return (start, end)
+
     def update_status(self) -> BatteryStatus:
+        (charge_start_threshold, charge_end_threshold) = self.charge_controller()
+        if self.force_charge:
+            charge_start_threshold = 0
+            charge_end_threshold = 100
+        self.set_battery_charge_thresholds(charge_start_threshold, charge_end_threshold)
```
elParaguayo, line 347: This will always set thresholds but what happens if a user has another script/app that's already setting charge limits? Should we have an option to disable smart charge completely?

elParaguayo (commented):
This is looking very nice - some comments/questions for you.

libqtile/widget/battery.py
tych0: Yep, good point. Now you can set charge_controller to None and it shouldn't do anything.

libqtile/widget/battery.py
```diff
@@ -267,7 +314,37 @@ def _get_param(self, name) -> tuple[str, str]:
...
+                f.write(str(end))
+        except FileNotFoundError:
+            self.charge_threshold_supported = False
+        except OSError:
+            logger.debug("Failed to write %s", path, exc_info=True)
+        return (start, end)
+
     def update_status(self) -> BatteryStatus:
+        (charge_start_threshold, charge_end_threshold) = self.charge_controller()
+        if self.force_charge:
```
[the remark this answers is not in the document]
tych0: Done.

libqtile/widget/battery.py
tych0: Sounds good. I updated the doc string, switched the default back to 100% (i.e. `charge_controller is None`), and added a comment in the changelog.

tych0:
Two other things: 1. I will add some tests (e.g. to the formatter to make sure that the charge control bits work); the rest of it I don't have a good handle on how to test really. 2. (I will do this in a separate PR), I think we should start shipping udev rules for qtile. We now have at least two (backlight + battery) widgets that can manipulate sysfs, and the battery one in particular will be hairy for people to figure out with other kinds of laptops.

elParaguayo (approved):
This looks great. Nothing further from me.

tych0:
Cool, thank you!

PR #4686, opened 2024-02-11
Add `initial_ratio` to `Columns` layout

elParaguayo:
Adds ability to set an initial ratio so that columns do not have even widths when creating a new column.


tych0 (approved):
any reason not to land this?

ThomasAH:
> any reason not to land this?

Not for me. I'm using it daily and it works great!

PR #4687, opened 2024-02-11
Don't attempt to reload config if file has been removed

brianm78:
Currently, if a previously imported file has been deleted in a user's config (eg. due to reorganization), this will break config reloading until qtile is restarted.

This adds a check that the file still exists before attempting to reload.


elParaguayo:
Nice. This sounds like a good addition.

PR #4688, opened 2024-02-11
Fire client_urgent_hint_changed when focus_on_window_activation is "urgent" or "smart"

### LINKED ISSUE #4074: client_urgent_hint_changed hook not firing
### The issue:

I'm trying to use the client_urgent_hint_changed hook however it doesn't seem to be firing ever. The group box and task list widgets update and the next_urgent command works 
```
@hook.subscribe.client_urgent_hint_changed
def urgent_hint_changed(window):
    logger.warning(f"urgent hint changed for {window.name}")
```
I set up a key-bind to look at the window's attributes and saw that `window.urgent` and `window._demands_attention` were True but `window.hints["urgent"]` was False. Running `window.update_hints()` manually does cause the hook to fire

I found #1847 which might be related

### Required:

ava5627:
fixes #4074

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -2164,9 +2164,11 @@ class Window(_Window, base.Window):
                     else:  # self.group.screen != self.qtile.current_screen:
                         logger.debug("Setting urgent flag for window")
                         self.urgent = True
+                        hook.fire("client_urgent_hint_changed", self)
                 elif focus_behavior == "urgent":
                     logger.debug("Setting urgent flag for window")
                     self.urgent = True
+                    hook.fire("client_urgent_hint_changed", self)
                 elif focus_behavior == "never":
                     logger.debug("Ignoring focus request (focus_on_window_activation='never')")
                 else:
--- a/test/backend/x11/test_window.py
+++ b/test/backend/x11/test_window.py
@@ -1,14 +1,16 @@
 import os
 import subprocess
 import tempfile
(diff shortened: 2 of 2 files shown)


tych0:
Patch itself looks good to me, any chance you can add some tests so we don't break this behavior in the future?

ava5627:
```python
class UrgentConfig(ManagerConfig):
    focus_on_window_activation = "urgent"


@pytest.mark.parametrize("xmanager", [UrgentConfig], indirect=True)
def test_urgent_hook_fire(xmanager, conn):
    hook_fired = False

    def _hook(window):
        nonlocal hook_fired
        hook_fired = True

    hook.subscribe.client_urgent_hint_changed(_hook)

    xmanager.test_window("one")
    window_info = xmanager.c.window.info()
    data = xcffib.xproto.ClientMessageData.synthetic([0, 0, 0, 0, 0], "IIIII")
    ev = xcffib.xproto.ClientMessageEvent.synthetic(
        32, window_info["id"], conn.atoms["_NET_ACTIVE_WINDOW"], data
    )
    conn.default_screen.root.send_event(ev, mask=xcffib.xproto.EventMask.StructureNotify)
    conn.xsync()

    assert hook_fired
```
do you know why the hook isn't firing here. I was able to confirm that `hook.fire` is being called from window and calling `hook.fire` within the test does set `hook_fired` to `true`

tych0:
Looks like you figured it out? Test LGTM, thank you.

PR #4690, opened 2024-02-12
Fix error if an wayland xwindow has unknown wm_type

dariogoetz:
This re-applies a past fix (see #3612) that apparently got lost during a refactoring (see #3704).

If an xwayland window has an unknown type (in my case a slack huddle window), the error appears and the window will not be visible.


tych0:
Thank you for writing a great PR description. Can you also put it in the commit message (`git commit --amend` and force push is fine) so we have it in the git logs as well?

tych0:
Thanks!

PR #4694, opened 2024-02-13
Allow providing a callback to make when a drag operation finishes.

brianm78:
This adds a callback that will be invoked when a drag window operation finishes.

The default configuration lets you drag windows between screens / groups as floating windows, but I wanted it to revert the floating status at the end (ie. when the button is released), so it'd snap into the other screens layout rather than needing to re-tile it as a second step.  Adding a callback here allows this to be configurable by something like:

            start=lazy.window.get_position(), end=lazy.window.toggle_floating() ),

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -184,7 +184,8 @@ class Drag(Mouse):
     warp_pointer:
         A :class:`bool` indicating if the pointer should be warped to the bottom right of the window
         at the start of dragging. (Default: `False`)
-
+    end:
+        A :class:`LazyCall` object to be evaluated when dragging finishes. (Optional)
     """
 
     def __init__(
@@ -194,9 +195,11 @@ class Drag(Mouse):
         *commands: LazyCall,
         start: LazyCall | None = None,
         warp_pointer: bool = False,
+        end: LazyCall | None = None,
     ) -> None:
         super().__init__(modifiers, button, *commands)
         self.start = start
+        self.end = end
         self.warp_pointer = warp_pointer
 
     def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


elParaguayo:
It's an interesting idea but I'm curious if there are any other use cases for this. The reason I ask is that, for your scenario, it seems it would be simpler to just use a key binding to move the window to the other screen's group without dragging.

brianm78:
I think there are probably a few other potential usecases.  Potentially, you could also support rearranging within a layout, if combined with something that determines the best position within the layout based on the final mouse / window position (though that's also obviously also something you can use the keyboard for, and may even be doable with just the regular drag events).

Personally, I find it handy to use just the mouse for this (I have my thumb button bound to drag without a modifier, so don't need a hand on the keyboard)

elParaguayo:
Thanks. I don't have any particular objections this. I'll take a proper look at the code later.

brianm78:
Incidentally, is the CI failure above something this change broke?  I tried rerunning tox locally, and it passes here (eventually), and looks like it may be unrelated (Timeout error starting the server on pypy?), but figured I should check.

elParaguayo:
I suspect that failure is unrelated. Am re-running the tests.

elParaguayo:
Although `start` can apparently take any lazy call, it seems designed specifically to take a function that returns a tuple of the window's position. If that's right, the description for `start` is misleading and `end` is therefore doing something a bit different.

brianm78:
Would it be better for these to allow arbitrary Callables as well as LazyCalls, similar to what widget callbacks do?  Ie if isinstance(action, LazyCall): invoke it, else: action()  (or action(x,y) in the case of commands?)

There seem reasons you might want to do something in start (eg. with the above, set a flag for windows that started floating so you don't reset it).  However, it looks like it's not actually possible to pass a start value that's actually a Lazy.function, since there's no way to return a value from that (it gets swallowed), meaning it'll end up returning `None` which breaks things.  Its less of an issue for commands, or the added end, since those don't need to return anything, but it'd probably be more consistent to do the same.

elParaguayo:
Can you explain this part:
>  However, it looks like it's not actually possible to pass a start value that's actually a Lazy.function, since there's no way to return a value from that (it gets swallowed),

The current functions assigned to `start` do return a value.

brianm78:
I mean if you wanted to do something like:

    @lazy.function
    def my_alternative_start(qtile):
        do_something_extra()
        return qtile.current_window.get_position()

It won't actually work as a start= argument, as the return value from a lazy function is swallowed - there's no way to use anything else for start other than something like lazy.window.get_postiion / lazy.window.get_size, or some other existing function that returns a tuple of values.

PR #4697, opened 2024-02-16
ci: coveralls master -> main

tych0:
It looks like the coveralls action  went from master -> main here, do the same in our actions code.


elParaguayo:
Nice.

Could you do pre_commit.yml too?

tych0:
Happy to do it, but I'm not sure I understand what you mean?

elParaguayo:
There were some similar messages on that workflow too.

Screenshot_20240216_223356_Chrome

tych0:
Ok... I don't understand how to fix these last two, so I'll leave them for now.

PR #4702, opened 2024-02-19
docs: move Scripting below Configuration on the nav bar

tych0:
I think people will probably configure before they script, let's rearrange these.


elParaguayo (approved):
Agreed. Will set to auto merge.

PR #4704, opened 2024-02-19
add udev rules

tych0:
Now that we support managing three different kinds of hardware, I think it's time for us to ship our own udev rules + script, instead of asking people to paste a random script and udev rules in each of the widget docs. Hopefully this means things will Just Work for people installing via distro packaging, but because of virtualenvs, people installing from source may have to do a bit of extra fiddling.

(Is there a better way than adding another script? Like a virtualenv one liner that will set up the virtual env and invoke qtile?)

### DIFF
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -188,3 +188,41 @@ Similar to the xsession example above, a wayland session file can be used to sta
 from a login manager. To use this, you should create a `qtile-wayland.desktop
 <https://github.com/qtile/qtile/blob/master/resources/qtile-wayland.desktop>`_ file in
 ``/usr/share/wayland-sessions``.
+
+udev rules
+==========
+
+Qtile has widgets that support managing various kinds of hardware (LCD
+backlight, keyboard backlight, battery charge thresholds) via the kernel's
+exposed sysfs endpoints. However, to make this work, Qtile needs permission to
+write to these files. There is a udev rules file at
+``/resources/99-qtile.rules`` in the tree, which users installing from source
+will want to install at ``/etc/udev/rules.d/`` on their system. By default,
+this rules file changes the group of the relevant files to the ``sudo`` group,
+and changes the file mode to be g+w (i.e. writable by all members of the sudo
+group). The theory here is that most systems qtile is installed on will also
+have the primary user in the ``sudo`` group. However, you can change this to
+whatever you like with the ``--group`` argument; see the sample udev rules.
+
+Note that this file invokes Qtile's hidden ``udev`` from udevd, so udevd will
+need ``qtile`` in its ``$PATH``. For distro packaging this shouldn't be a
+problem, since /usr/bin is typically in udev's path. However, for users that
+installed from source, you may need to modify the udev script to be one that
+sources your virtualenv and then invokes qtile (or just invoke it via its
+hardcoded path if you installed it with ``--break-system-packages``), e.g.:
+
+.. code-block:: bash
+
+    # create a wrapper script that loads the right stuff from our home directory; since
+    # udev will run this script as root, it has no idea about how we've installed qtile
+    mkdir -p $HOME/.local/bin
+    tee $HOME/.local/bin/qtile-udev-wrapper <<- EOF
+    #!/bin/sh
+
+    export PYTHONPATH=$HOME/.local/lib/python$(python3 --version | awk -F '[ .]' '{print $2 "." $3}')/site-packages
+    $HOME/.local/bin/qtile $@
+    EOF
+
+    # copy the in-tree udev rules file to the right place to make udev see it,
+    # and change the rules to point at our wrapper script above.
+    sed "s,qtile,$HOME/.local/bin/qtile-udev-wrapper,g" ./resources/99-qtile.rules | sudo tee /etc/udev/rules.d/99-qtile.rules
--- a/libqtile/scripts/udev.py
+++ b/libqtile/scripts/udev.py
@@ -1,0 +1,43 @@
+import argparse
+import glob
+import os
+import shutil
+import stat
+
+MODE = stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP | stat.S_IROTH
+
+
+def set_file_perms(p, options):
+    try:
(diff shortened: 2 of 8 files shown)


libqtile/scripts/udev.py
```diff
@@ -0,0 +1,43 @@
...
+        do_battery_setup(options)
+    else:
+        raise "Unknown udev option {}".format(options.kind)
+
+
+def add_subcommand(subparsers, parents):
+    parser = subparsers.add_parser("udev", parents=parents, help=argparse.SUPPRESS)
+    parser.add_argument("kind", choices=["backlight", "battery"])
+    parser.add_argument("--device")
+    parser.add_argument("--group", default="sudo")
```
elParaguayo, line 42: This assumes user is in the `sudo` group. Might be worth highlighting that in the docs in case someone needs to change it.

libqtile/widget/backlight.py
```diff
@@ -45,10 +45,8 @@ class Backlight(base.InLoopPollText):
     If the change_command parameter is set to None, the widget will attempt to
     use the interface at /sys/class to change brightness. Depending on the
     setup, the user may need to be added to the video group to have permission
```
elParaguayo, line 47: We refer to `video` group here but we're not setting the group ownership to `video`.

resources/99-qtile.rules
```diff
@@ -0,0 +1,26 @@
+# for controlling LCD backlight
+ACTION=="add", SUBSYSTEM=="backlight", RUN+="qtile udev backlight --device %k"
+
+# keyboard backlight
+ACTION=="add", SUBSYSTEM=="leds",      RUN+="qtile udev backlight --device %k"
```
elParaguayo, line 5:
```suggestion
ACTION=="add", SUBSYSTEM=="backlight", RUN+="qtile udev backlight --device %k  --group video"

# keyboard backlight
ACTION=="add", SUBSYSTEM=="leds",      RUN+="qtile udev backlight --device %k --group video"
```
Just an idea...

elParaguayo (commented):
Nice. Have added some comments here.

docs/manual/install/index.rst
```diff
@@ -188,3 +188,30 @@ Similar to the xsession example above, a wayland session file can be used to sta
...
+
+.. code-block:: bash
+
+    #!/bin/sh
+
+    export PYTHONPATH=/home/tycho/.local/lib/python$(python3 --version | awk -F '[ .]' '{print $2 "." $3}')/site-packages
+    /home/tycho/.local/bin/qtile $@
+
+or similar, and then ``sed -i 's,qtile,/home/tycho/config/bin/qtile-udev,g'
+99-qtile.rules``.
```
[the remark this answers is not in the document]
tych0: Yep, maybe I should turn it into a heredoc so people can copy and paste it and automatically sed it. I will do that and see how it looks.

libqtile/scripts/udev.py
tych0: Thank you, I meant to do this and totally forgot.

libqtile/widget/backlight.py
tych0: Thanks, need to drop this as well.

resources/99-qtile.rules
tych0: IMO, we should use the same group (sudo, video, whatever) for everything. I picked sudo because I assumed that most single user systems have the user qtile runs as in the sudo group. We could add a new `qtile` group, but that would introduce even more setup than already exists for people installing by source, so I didn't opt for that.

elParaguayo: I'm fine with using sudo group as long as it's clear to users. Agreed, a single group would be preferable.

tych0:
Ok, check that out and see it if it reads more nicely.

docs/manual/install/index.rst
```diff
@@ -188,3 +188,42 @@ Similar to the xsession example above, a wayland session file can be used to sta
...
+backlight, keyboard backlight, battery charge thresholds) via the kernel's
+exposed sysfs endpoints. However, to make this work, Qtile needs permission to
+write to these files. There is a udev rules file at
+``/resources/99-qtile.rules`` in the tree, which users installing from source
+will want to install at ``/etc/udev/rules.d/`` on their system. By default,
+this rules file changes the group of the relevant files to the ``sudo`` group,
+and changes the file mode to be g+w (i.e. writable by all members of the sudo
+group). The theory here is that most systems qtile is installed on will also
+have the primary user in the ``sudo`` group. However, you can change this to
+whatever you like with the ``-group`` argument; see the sample udev rules.
```
elParaguayo, line 205:
```suggestion
whatever you like with the ``--group`` argument; see the sample udev rules.
```

docs/manual/install/index.rst
```diff
@@ -188,3 +188,42 @@ Similar to the xsession example above, a wayland session file can be used to sta
...
+    #!/bin/sh
+
+    export PYTHONPATH=$HOME/.local/lib/python$(python3 --version | awk -F '[ .]' '{print $2 "." $3}')/site-packages
+    $HOME/.local/bin/qtile $@
+    EOF
+
+    # copy the in-tree udev rules file to the right place to make udev see it
+    cp ./resources/99-qtile.rules /etc/udev/rules.d
+    # change it from pointing to qtile to our wrapper script
+    sed -i "s,qtile,$HOME/.local/bin/qtile-udev-wrapper,g" /etc/udev/rules.d/99-qtile.rules
```
elParaguayo, line 229: Would it be easier to change the file before moving it to `/etc/udev/rules.d`?

elParaguayo (commented):
Looking really nice.

One typo and one potential change but, other than that, this looks good.

docs/manual/install/index.rst
tych0: Then you end up modifying it in your work tree, and you have to reset. There's no reason to do it in two commands though, we can just do it as part of the copy. This version has neither of those problems :)

elParaguayo: Good point!

elParaguayo (approved):
Thanks for this. No further comments.

tych0:
Thank you!

PR #4706, opened 2024-02-20
Use fallback icon in StatusNotifier widget

dariogoetz:
StatusNotifier widget items do not render (at all), when a corresponding icon can not be found.

In my case, the `nm-applet` does not render at all, if there is no network connection, but a wireguard connection is configured (and active). The corresponding StatusNotifier item does not render at all, thus not allowing me to activate a network connection to resolve this state.

This commit allows a StatusNotifier item to fallback to some default icon in that case.

The fallback icon is copied from `libqtile/resources/layout-icons/layout-unknown.png`.


tych0:
Patch itself lgtm. Any chance you can add a test for this so we don't break it?

elParaguayo:
I have mixed feelings about this one.

Firstly, are we 100% sure that the missing icon is not due to a bug in the code? I make mistakes so I'm wondering if there's some scenario that the widget hasn't catered for. There's not a huge amount of logging from the widget with regards to how it finds icons.

With regards to the patch, with no tooltips, it's impossible to know which app is responsible for the fallback icon (e.g. if you have more than one app in this scenario). The core qtile widget only provides the activate action when the icon is clicked which isn't always helpful. However, the qtile-extras version (which builds off this one) will benefit as it will allow access to the context menus.

I think we should increase debugging too i.e. if it's looking for a local icon but can't find it then we should add more logging to show which icon name it couldn't find so a user knows what file they would need to install to fix this.

dariogoetz:
> I have mixed feelings about this one.
> 
> Firstly, are we 100% sure that the missing icon is not due to a bug in the code? I make mistakes so I'm wondering if there's some scenario that the widget hasn't catered for. There's not a huge amount of logging from the widget with regards to how it finds icons.

From what I have researched, this is not due to a bug in the code. At least in my case, the `nm-applet` item specified an icon called `nm-no-connection-secure`, for which there is no icon in my iconset.

The code for loading the icon first tries finding a "custom" icon by concatenating the DBus-object's `get_icon_theme_path` with the icon's name (in my case, the theme path is always an empty string). When it can not find image files at these locations, it next tries using `pyxdg`'s `getIconPath` function for looking for the icon's image file. If the icon does not exist, this function returns `None`. In particular, it does not tell, where it tried to look for the icon. This is what happened in my case.

> With regards to the patch, with no tooltips, it's impossible to know which app is responsible for the fallback icon (e.g. if you have more than one app in this scenario). The core qtile widget only provides the activate action when the icon is clicked which isn't always helpful. However, the qtile-extras version (which builds off this one) will benefit as it will allow access to the context menus.
> 
> I think we should increase debugging too i.e. if it's looking for a local icon but can't find it then we should add more logging to show which icon name it couldn't find so a user knows what file they would need to install to fix this.

I agree, it is an issue that by the icon alone, a user can not tell which app is responsible without `qtile-extras`. The warning message that I added gives the name of the icon to look for, but not the path where the icon is looked for. This is certainly possible for the "custom" icon path (but in my case would not help much as `get_icon_theme_path` gives an empty string). As far as I can see, `pydg` does not provide any information in that regard in the case no icon could be found. I am therefore not sure, how logging/debugging could be improved here.

elParaguayo:
Thanks for the detailed answer here. Based on that, I think this is ok too. I'd just ask for one change: can you add something in the widget's docstring that explains that a fallback icon is used if the widget can't find the necessary icon and, in that situation, the user should check their log to see the name of the missing icon.

As for adding a test, that's not entirely straightforward. Have a go if you want, if not, I'll have a think as to how we could do that.

dariogoetz:
I added a note in the docstring of the widget.

More importantly, I realized that the previous solution was not correct. It would fallback to the default icon before checking app-provided icons (`Icon`, `Attention`, or `Overlay`), which is not what we want.
I have implemented a fix for this (first a "local" icon is tried to be found without fallback, then the app-provided ones are checked, and only if those are not available, the fallback is used and "local" icons will be registered for update events). 

Currently, the tests for the `StatusNotifier` widget are skipped due to apparently being "broken". I'll have a look, if I can get them to work.

elParaguayo:
Don't worry about the tests. I'll fix them later. I just need to test your code locally before approving this.

dariogoetz:
It works well for me, but I was not able to successfully get the tests working, sorry.

elParaguayo:
Thanks. I've not had time to look at the tests for this yet but I don't think that should hold up us merging this.

Code looks ok to me.

Issue #933, opened 2016-07-01
create Key by keycode


danmilon:
I'm often switching between US and Greek keyboard layouts, and when I do that, the registered keyboard shortcuts stop working, which is reasonable because `Key('M-f')` is different than `Key('M-φ')`.

But the interesting thing is that both `f`, and `φ`, which are the same physical key, have the same keycode. So I thought I should be able to create a shortcut by specifying the keycode instead of the symbol. [Awesome WM does that](https://awesome.naquadah.org/wiki/Global_Keybindings) through notation such as `#48` (indicating key with keycode 48). It would be very helpful for me to have similar functionality in QTile.

If you agree we need this, I could jump in and try to implement it but would need to know where in the code I should do that.

As a workaround I can create both `Key('M-f')` and `Key('M-φ')` but thats redundant.

andering:
So, keys are being compared with (i think Unicode Hexadecimal) [xkeysyms.py](https://github.com/qtile/qtile/blob/develop/libqtile/xkeysyms.py)
and check is being done here [config.py](https://github.com/qtile/qtile/blob/develop/libqtile/config.py#L58)

btw.: f = 0x03d5 and φ = 0x0066 , that's reason it doesn't work way you expect.
I would personally create **'else if'** with check if format is hexadecimal, in case key is not found in xkeysyms.

But i have no clue if that's ok with developers :).

zordsdavini:
it should be ok, because there are many users what are not using latin
basic or even latin alphabet. For example my number line is "ąčęėįšų", so I
can't bind any keybinding on numbers (ex. dynamic groups)

On 1 July 2016 at 12:16, Andrej Kouřil <notifications@github.com> wrote:

> So, keys are being compared with (i think Unicode Hexadecimal) xkeysyms.py
> <https://github.com/qtile/qtile/blob/develop/libqtile/xkeysyms.py>
> and check is being done here config.py
> <https://github.com/qtile/qtile/blob/develop/libqtile/config.py#L58>
>
> btw.: f = 0x03d5 and φ = 0x0066 , that's reason it doesn't work way you
> expect.
> I would personally create *'else if'* with check if format is
> hexadecimal, in case key is not found in xkeysyms.
>
> But i have no clue if that's ok with developers :).
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/qtile/qtile/issues/933#issuecomment-229898971>, or mute
> the thread
[...]

PR #4707, opened 2024-02-20
Bind keys with keycodes

elParaguayo:
Draft for now...

Based on the discussion in #933 this is an attempt to allow users to bind keys via keycode rather than keysym.

Keycodes can be bound with:
```python
Key([mod], key_code, lazy...)
```
Where `key_code` is an integer value (compared to keysyms being strings).

Ideally, I need users to test this and report any bugs.

@jwijenbergh @m-col would welcome thoughts on the wayland implementation. I've hardcoded this to match against `self.keyboards[0]` but not sure if that's appropriate or not.

CI will almost certainly fail for the time being too!

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -99,6 +99,7 @@ from libqtile.backend.wayland.output import Output
 from libqtile.command.base import expose_command
 from libqtile.config import ScreenRect
 from libqtile.log_utils import logger
+from libqtile.utils import QtileError
 
 try:
     # Continue if ffi not built, so that docs can be built without wayland deps.
@@ -392,6 +393,16 @@ class Core(base.Core, wlrq.HasListeners):
     @property
     def name(self) -> str:
         return "wayland"
+
+    @property
+    def active_keyboard(self) -> inputs.Keyboard | None:
+        keyboard = self.seat.get_keyboard()
+        if keyboard is not None:
+            for kb in self.keyboards:
+                if kb.keyboard._ptr == keyboard._ptr:
+                    return kb
+
+        return None
 
     def finalize(self) -> None:
         self.finalize_listeners()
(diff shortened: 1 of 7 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -1449,14 +1449,20 @@ def get_screen_info(self) -> list[ScreenRect]:
 
     def grab_key(self, key: config.Key | config.KeyChord) -> tuple[int, int]:
         """Configure the backend to grab the key event"""
-        keysym = xkb.keysym_from_name(key.key, case_insensitive=True)
+        if isinstance(key.key, str):
+            keysym = xkb.keysym_from_name(key.key, case_insensitive=True)
+        else:
+            keysym = self.keyboards[0]._state.key_get_one_sym(key.key)
```
jwijenbergh, line 1455: You can get the active keyboard using the seat. Use self.seat.get_keyboard() and null check it. _state is nowhere defined anyways right?

elParaguayo: How does that help? the `Keyboard` returned by this doesn't have the keymap (as far as I can tell) so I can't get a keysym from the code.

`self._state` is defined in this PR as part of our `Keyboard` class.

jwijenbergh: Loop through keyboards and check if the wlr_keyboard is the right one. getting the first keyboard is wrong

jwijenbergh (approved):
Wayland looks good now. Thanks!

libqtile/backend/wayland/core.py
```diff
@@ -393,6 +394,15 @@ def get_enabled_outputs(self) -> list[Output]:
     def name(self) -> str:
         return "wayland"
 
+    @property
+    def active_keyboard(self) -> inputs.Keyboard | None:
+        keyboard = self.seat.get_keyboard()
```
jwijenbergh, line 399: Oh wait can you check if this is None before the for loop?

jwijenbergh: Otherwise we would get NoneType has no attribute ._ptr inside the for loop

elParaguayo: Yeah. Just thought of that after I pushed this.

jwijenbergh: Fine now thanks

jwijenbergh (commented):
Typo layours -> layouts

elParaguayo:
Other than my typo (which I'll fix tonight), @tych0 are you ok with this?

Shringe:
If this still needs testing, how can I download and test this commit locally?

tych0 (approved):

tych0:
> If this still needs testing, how can I download and test this commit locally?

you should be able to add elParaguayo's github remote and fetch the `config-allow-keycodes` branch and install that. the docs have info on installing from source generally.

elParaguayo:
Right. I've fixed the typo and squashed into a single commit. Will set this to automerge.

PR #4712, opened 2024-02-28
Fix WindowTabs use of parse_text

heitzmann:
This fixes the issue raised in #3741 by using the parser function for each window title before any state or markup is applied, similarly to what is done for the TaskList widget.

Additionally, instead of checking if the value is callable, it uses a default callable value and also skips the `try...except` block. I believe any raised  errors will be logged anyway, won't they? And with better debug information than only `"parse_text function failed:"`

### DIFF
--- a/libqtile/widget/windowtabs.py
+++ b/libqtile/widget/windowtabs.py
@@ -23,6 +23,7 @@
 # SOFTWARE.
 
 from libqtile import bar, hook, pangocffi
+from libqtile.log_utils import logger
 from libqtile.widget import base
 
 
@@ -62,7 +63,11 @@ class WindowTabs(base._TextBox):
     def update(self, *args):
         names = []
         for w in self.bar.screen.group.windows:
-            name = self.parse_text(w.name if w and w.name else " ")
+            try:
+                name = self.parse_text(w.name if w and w.name else "")
+            except:  # noqa: E722
+                logger.exception("parse_text function failed:")
+                name = w.name if w and w.name else "(unnamed)"
             state = ""
             if w.maximized:
                 state = "[] "
--- a/libqtile/widget/windowtabs.py
+++ b/libqtile/widget/windowtabs.py
@@ -23,7 +23,6 @@
 # SOFTWARE.
 
 from libqtile import bar, hook, pangocffi
-from libqtile.log_utils import logger
 from libqtile.widget import base
 
 
--- a/libqtile/widget/windowtabs.py
+++ b/libqtile/widget/windowtabs.py
@@ -39,15 +39,10 @@ class WindowTabs(base._TextBox):
         ("selected", ("<b>", "</b>"), "Selected task indicator"),
         (
             "parse_text",
-            None,
-            "Function to parse and modify window names. "
-            "e.g. function in config that removes excess "
-            "strings from window name: "
-            "def my_func(text)"
-            '    for string in [" - Chromium", " - Firefox"]:'
-            '        text = text.replace(string, "")'
-            "   return text"
-            "then set option parse_text=my_func",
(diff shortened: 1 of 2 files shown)


libqtile/widget/windowtabs.py
```diff
@@ -68,22 +62,17 @@ def _configure(self, qtile, bar):
...
+            task = pangocffi.markup_escape_text(state + name)
             if w is self.bar.screen.group.current_window:
                 task = task.join(self.selected)
             names.append(task)
         self.text = self.separator.join(names)
-        if callable(self.parse_text):
-            try:
-                self.text = self.parse_text(self.text)
-            except:  # noqa: E722
-                logger.exception("parse_text function failed:")
```
elParaguayo, line 88: Please keep the `try... except...` here. `logger.exception` prints that line but then also prints the error traceback. Removing the logger line still logs the error but it's good to have that additional line first.

You'd want to move this to where you're now calling `parse_text`. You could also default to the window name if the function fails too.

elParaguayo (changes requested):
Thanks for this. I've made a request for one change.

libqtile/widget/windowtabs.py
heitzmann: Got it! I've added the `try` block back and set a default value for the name in case of error.

libqtile/widget/windowtabs.py
```diff
@@ -68,22 +63,21 @@ def _configure(self, qtile, bar):
     def update(self, *args):
         names = []
         for w in self.bar.screen.group.windows:
+            try:
+                name = self.parse_text(w.name if w and w.name else "")
+            except:  # noqa: E722
+                logger.exception("parse_text function failed:")
+                name = w.name if w and w.name else "(unnamed)"
```
elParaguayo, line 70: Do you need the `if w` part of the check here? If there any scenario where `w` is not a window object?

elParaguayo (commented):
Thanks for the updates. I've added one question.

Also, would be good to add a test for this. This should do it:
```diff
diff --git a/test/widgets/test_windowtabs.py b/test/widgets/test_windowtabs.py
index 424044b7..f582b8cd 100644
--- a/test/widgets/test_windowtabs.py
+++ b/test/widgets/test_windowtabs.py
@@ -28,6 +28,10 @@ from libqtile.config import Screen
 from libqtile.confreader import Config


+def custom_text_parser(name):
+    return f"TEST-{name}-TEST"
+
+
 class WindowTabsConfig(Config):
     auto_fullscreen = True
     groups = [libqtile.config.Group("a"), libqtile.config.Group("b")]
@@ -40,6 +44,7 @@ class WindowTabsConfig(Config):
             top=bar.Bar(
                 [
                     widget.WindowTabs(),
+                    widget.WindowTabs(name="customparse", parse_text=custom_text_parser)
                 ],
                 24,
             ),
@@ -136,3 +141,14 @@ def test_escaping_text(manager):
     """
     manager.test_window("Text & Text")
     assert manager.c.widget["windowtabs"].info()["text"] == "<b>Text &amp; Text</b>"
+
+
+@windowtabs_config
+def test_custom_text_parser(manager):
+    """Test the custom text parser function."""
+    manager.test_window("one")
+    assert manager.c.widget["customparse"].info()["text"] == "<b>TEST-one-TEST</b>"
```

libqtile/widget/windowtabs.py
heitzmann: That I'm not sure. It comes from the original implementation, so I kept it.

elParaguayo (approved):

elParaguayo:
Great. This looks ok now. I'll set it to merge once tests pass.

PR #4716, opened 2024-03-06
Better handling of timezones in Clock widget

### LINKED ISSUE #3746: Clock and tzupdate
### The issue:

qtile version:
0.21.0

These days I need to travel with my system and have the need to update my timezone in order to work with my calendar. I'm using `tzupdate` which easily updates my timezone with one command. 

I'm using the Clock widget in the qtile bar as so:
``` python 
        widget.Clock(format="%A %d %b %Y %H:%M:%S %z"),
```

Updating the timezone with `tzupdate` however does not change the timezone on the Clock widget. It requires restarting qtile in order to get this done. I would expect qtile to poll the current system timezone at each tik.

### Required:

### LINKED ISSUE #3789: [CheckUpdates, doc, Arch] mention `pacman -Sy` is required for `pacman -Qu` to show updated info
### The issue:

As per title. Maybe this is common knowledge among arch users. I had to spend some debugging time to get the CheckUpdate widget to show updated info, and ended up here

https://www.reddit.com/r/qtile/comments/ur8mz3/comment/i90sa7d/

### Required:

elParaguayo:
This PR adds two new commands to the Clock widget:
1) update_timezone: allows users to specify a new timezone or just
   refresh the clock based on changes to the system timezone
2) use_system_timezone: clears user-specified timezone so widget uses
   system time

Fixes #3746

### DIFF
--- a/libqtile/widget/clock.py
+++ b/libqtile/widget/clock.py
@@ -20,11 +20,13 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+from __future__ import annotations
 
 import sys
 import time
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta, timezone, tzinfo
 
+from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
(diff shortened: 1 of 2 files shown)


test/widgets/test_clock.py
```diff
@@ -43,10 +43,10 @@ class MockDatetime(datetime.datetime):
     def now(cls, *args, **kwargs):
         return cls(2021, 1, 1, 10, 20, 30)
 
-    def astimezone(self, tzone=None):
-        if tzone is None:
-            return self
-        return self + datetime.timedelta(hours=tzone)
+    # def astimezone(self, tzone=None):
```
tych0, line 46: Can we drop this?

libqtile/widget/clock.py
```diff
@@ -88,3 +101,30 @@ def poll(self):
...
+        If the method is called with no arguments then the widget will reload
+        the timzeone set on the computer (e.g. via ``timedatectl set-timezone ..``).
+        This will have no effect if you have previously set a ``timezone`` value.
+
+        Alternatively, you can pass a timezone string (e.g. ``"Europe/Lisbon"``) to change
+        the specified timezone. Setting this to an empty string will cause the clock
+        to rely on the system timezone.
+        """
+        if timezone:
+            self._check_timezone(timezone)
```
tych0, line 119: minor, but i wonder if this would look cleaner if timezone returned the right value, so these four lines would be `self.timezone = _lift_timezone(timezone)` or similar.

libqtile/widget/clock.py
```diff
@@ -88,3 +101,30 @@ def poll(self):
...
+        elif timezone is not None:
+            self.timezone = None
+
+        # Force python to update timezone info
+        time.tzset()
```
tych0, line 124: I'm confused about why we need this line. IIUC, either we 1. have an explicit non-none `self.timezone` in which case we pass that in `poll()` when doing our rendering, or 2. the timezone is none, in which case we want to use the system time. Since our rendering code never fiddles with `TZ`, we shouldn't ever need to reset it should we?

Or is the idea that we need this call after a system time change in order to kick python into rendering relative to the new system timezone?

elParaguayo: Correct. It's the latter.

test/widgets/test_clock.py
elParaguayo: No. 

Oh, okay then.

libqtile/widget/clock.py
elParaguayo: Yes.

PR #4717, opened 2024-03-07
Remove rogue print statements from tests

elParaguayo:
There are some tests that still have print statements but these look intentional. However, the ones removed by this PR were clearly left over from when the tests were being written!


m-col (approved):
Nice :) (hello)

PR #4730, opened 2024-03-17
Added available parameter to memory widget

sohanglal:


tych0:
Seems like this is causing some test failure, can you take a look?

elParaguayo:
We mock the psutil output so you just need to add a line to include a value for the available amount. Ideally, you should add a test to check this output works too.

sohanglal:
I accidently disabled auto merge

tych0:
CI failure is not related to your PR (it's a hang we're investigating, but it looks like my recent code didn't work...). I will send a PR for that and re-run.

PR #4731, opened 2024-03-17
drop deprecated command_* things

tych0:
These were deprecated before 0.17.0, which was released two years ago. Seems fine to drop these now.


elParaguayo:
Nice.

You may need to delete this migration too: https://github.com/qtile/qtile/blob/master/libqtile/scripts/migrations/module_renames.py

tych0:
I was thinking of keeping the migration, but maybe not. It does feel like we shouldn't keep this old cruft around forever.

Two years seems like a reasonable date after which to delete migrations, should I write that down in the documentation?

Worth noting that this was from before we had migrations IIRC, so hopefully we won't have lots of code like in this PR sprinkled around the tree in the future, so it will be mostly self contained in migrations.

elParaguayo:
Happy to go with whichever option you prefer.

elParaguayo:
Thanks for this.

Thinking about the migrations, we could leave these for longer as `qtile migrate` has an `--after-version` option to only apply changes introduced after a specific version.

PR #4732, opened 2024-03-18
tests: reorder SIGWINCH to be after sleep

tych0:
It looks like in a recent run:
https://github.com/qtile/qtile/actions/runs/8321862376/job/22786549176 which had my previous patch we didn't get the stack traces. Presumably that's because we sent SIGWINCH and then immediately sent SIGKILL. I meant to put this sleep *after* SIGWINCH to give python time to walk its stack, but put it before. derp.


elParaguayo (approved):
LGTM

PR #4734, opened 2024-03-20
Add HDD widget

florian-lnx:
I made a HDD widget that displays the HDD utilization in percent based on io-ticks (The same principle used by the HDDBusyGraph widget). I also wrote a test and screenshot routine and ran the pre-commit hooks.

### DIFF
--- a/libqtile/widget/hdd.py
+++ b/libqtile/widget/hdd.py
@@ -1,0 +1,61 @@
+# Copyright (c) 2024 Florian G. Hechler
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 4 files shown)


libqtile/widget/hdd.py
```diff
@@ -0,0 +1,61 @@
...
+            with open(self.path) as f:
+                io_ticks = int(f.read().split()[9])
+
+            variables["HDDPercent"] = round(
+                max(min(((io_ticks - self._prev) / self.update_interval) / 10, 100.0), 0.0), 1
+            )
+
+            self._prev = io_ticks
+
+        except IOError:
```
tych0, line 58: Here I would just let it crash. Qtile itself will handle the exception and log it, which should make it more obvious to people why their widget is rendering 0 instead of a reasonable number if e.g. they got the path wrong, udev labeled their drive with a different name this boot, etc.

tych0:
Hmm. Somehow github actions never ran for this? Let me rebase it. Sorry for the delay.

PR #4737, opened 2024-03-22
Handle string type in `xmonad.set_ratio`

fiorematteo:
`set_ratio` can be called from IPC which will pass ratio as a string.
This matches implementation found in spiral layout.

I'm not sure how I can add a test for this use case or if it's even possible.

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -40,6 +40,7 @@ from typing import TYPE_CHECKING
 
 from libqtile.command.base import expose_command
 from libqtile.layout.base import _SimpleLayoutBase
+from libqtile.log_utils import logger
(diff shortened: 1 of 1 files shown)


tych0:
I would prefer a more general solution here for all argument types and all functions that are callable via IPC. We have been discussing it in https://github.com/qtile/qtile/pull/2435

I have it somewhat working in https://github.com/tych0/qtile/commit/7f8aef7a432b11b83bd321feed49e54994dd1a1c but not well enough to land unfortunately. If you have time to get it over the finish line that would be great, otherwise I will keep working on it as time permits.

jwijenbergh:
I agree. Would be neat to not have to worry about this in every expose_command

PR #4738, opened 2024-03-22
widget: drop UNSPECIFIED arg to set_font()

tych0:
It's a little weird to have what is essentially our own version of None as a default argument for one function. Let's drop it, and just use none the real None.

Hopefully nobody uses this, since it's the default arg, but I wrote a migration anyway, mostly to understand how the new migrations work.

The motivation here is that I want to use set_font() as an example of the type hoisting, since it has interesting types (int, str, ColorType which is itself a union, etc.).

### DIFF
--- a/libqtile/scripts/migrations/rename_unspecified.py
+++ b/libqtile/scripts/migrations/rename_unspecified.py
@@ -1,0 +1,112 @@
+# Copyright (c) 2024, Tycho Andersen. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+from itertools import filterfalse
+
+import libcst as cst
+import libcst.matchers as m
+
+from libqtile.scripts.migrations._base import (
+    Check,
+    MigrationTransformer,
+    _QtileMigrator,
+    add_migration,
+)
+
+
+class RenameUnspecifiedTransformer(MigrationTransformer):
+    @m.leave(m.Call(func=m.Attribute(attr=m.Name("set_font"))))
+    def update_unspecified_arg(self, original_node, updated_node) -> cst.Call:
+        args = original_node.args
+
+        def is_interesting_arg(arg):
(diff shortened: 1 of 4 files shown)


elParaguayo:
Is `None` every a valid value for any of those properties? If so, we'd need to be able to differentiate the times when the user is explicitly setting a value to `None` and when they haven't set it at all.

tych0:
Yes, you're right. This is really a tristate, which is kind of annoying. For `font` itself, None is not a valid value, but for font size it is (it's automatically calculated if the font size is None), and for fontshadow the shadowing is disabled when the shadow is None.

So I suppose to be strictly correct, we need to keep `'UNSPECIFIED'` around for the fontsize and fontshadow arguments. ugh.

tych0:
One thought I had: we could move to the zero-value as unset here. The migration would be from 'UNSPECIFIED' to `fontsize=0` and `fontshadow=""` to get the calculated values here. Or we could just abandon this outside of `font=`.

elParaguayo:
I like the `fontsize` approach but I'd be inclined to drop `fontshadow`.

tych0:
meaning you'd like to keep the special string `UNSPECIFIED` for fontshadow?

elParaguayo:
I may have misinterpreted what you were suggesting. I thought you meant dropping the ability to set `fontshadow`.

tych0:
Oh, I was thinking allowing to set it, but changing the sentinel from `UNSPECIFIED -> ""`

elParaguayo:
Sure. That works.

tych0:
Ok, take a look at it now. It also uncovered a bug for lifting Union types, which I added a fix for.

libqtile/scripts/migrations/rename_unspecified.py
```diff
@@ -0,0 +1,112 @@
...
+    HELP = """
+    The UNSPECIFIED object was removed in favor of using the zero values (or
+    None) to leave behavior unspecified. That is:
+
+        font=UNSPECIFIED -> font=None
+        fontsize=UNSPECIFIED -> fontsize=0
+        fontshadow=UNSPECIFIED -> fontshadow=""
+    """
+
+    AFTER_VERSION = "0.24.0"
```
elParaguayo, line 80: This should be `0.25.0` because that tag has been released and so this change was added _after_ `0.25.0`.

libqtile/scripts/migrations/rename_unspecified.py
```diff
@@ -0,0 +1,112 @@
...
+
+        args = list(filterfalse(is_interesting_arg, original_node.args))
+
+        if len(args) == len(original_node.args):
+            return original_node
+
+        return updated_node.with_changes(args=args)
+
+    def strip_unspecified_imports(self, original_node, updated_node) -> cst.Import:
+        self.lint(original_node, "'UNSPECIFIED' can be dropped")
```
elParaguayo, line 49: Isn't this going to be added even if `UNSPECIFIED` isn't in the import line? We only want the linting if it's found.

The tests only check if linting is present if there's a change. Should we change that so it also checks that there's no linting if there's no change? In that case, a `NoChange` test here with a standard import line would need to provide no linting.

elParaguayo (commented):
Thanks. Have left a couple of comments but this is almost ready to merge.

libqtile/scripts/migrations/rename_unspecified.py
tych0: derp, thank you!

libqtile/scripts/migrations/rename_unspecified.py
tych0: Yes, good point. I moved the `self.lint` call, I can add a no-lint-if-no-changes test in another PR.

libqtile/scripts/migrations/rename_unspecified.py
```diff
@@ -0,0 +1,112 @@
+# Copyright (c) 2023, elParaguayo. All rights reserved.
```
elParaguayo, line 1: Minor point but this is yours and this year...

Other than that, this all looks good to me.

tych0: urgh, yep. I wonder if we can switch to SPDX headers vs. having this giant blurb everywhere. MIT makes it hard i guess.

tych0: Oh, it does have an SPDX header: https://spdx.org/licenses/MIT.html

It's just debian that won't define MIT. So maybe we can switch and get rid of all of this crap!

PR #4740, opened 2024-03-24
move launchbar module docstring

tych0:
this docstring has examples of additional functionality in the widget, but doesn't get rendered in our docs because it's at the module level instead of the class level. let's drop the module docstring entirely and put that string in the class string, so we get these examples in our docs.


elParaguayo (approved):

jwijenbergh:
Thanks

PR #4741, opened 2024-03-26
Type conversions

tych0:
Ok, I finally managed to get this working in a reasonable way. See the patch notes for details.

The core shift was from using `inspect.signature()` like in my samples from #2435 to `typing.get_type_hints()`, which simplified a lot of things.


tych0:
oof, looks like there's a bug here. will fix.

tych0:
Ok, I think this is finally ready for a look. Some strange things, in no particular order:

1. we have to drop 3.9 support, since the typing module's evaluator doesn't understand the "foo | bar" type syntax in 3.9
2. we have to declare every type that we could possibly want in interface.py, which is a bit ugly
3. we have sort of a laundry list of special type handling

I still think it's worth doing it this way (at least giving it a try), vs. having this code sprinkled all over the tree. However, I'd like to do a release soon, and I will probably wait until that to actually merge this, since it is a somewhat more delicate change than I was anticipating.

Anyway, thoughts welcome.

ramnes:
> 1. we have to drop 3.9 support, since the typing module's evaluator doesn't understand the "foo | bar" type syntax in 3.9

No big deal since we said we'd focus on supporting the last three versions and 3.12 is out. Worst case scenario I guess we can replace these with `Union` and that would work.

.github/workflows/release.yml
```diff
@@ -125,7 +125,7 @@ jobs:
     container: quay.io/pypa/manylinux_2_28_x86_64
     strategy:
       matrix:
-        python-version: ["cp3.9", "cp3.10", "cp3.11", "pp3.9"]
+        python-version: ["cp3.10", "cp3.11", "pp3.9"]
```
ramnes, line 128: why no 3.12 here?

tox.ini
```diff
@@ -35,7 +35,7 @@ commands =
     !x11: {toxinidir}/scripts/ffibuild
 
 # These are the environments that should be triggered by Github Actions
-[testenv:py{py3,39,310,311}-{wayland,x11}]
+[testenv:py{py3,39,310,311,312}-{wayland,x11}]
```
ramnes, line 38: Let's remove 39 here?

ramnes (approved):
didn't try but LGTM code-wise

elParaguayo:
Thanks for persisting with this.

Some comments:
- Is `lift_arg` being run even when those commands are called internally? We should limit this to calls over IPC only.
- I still feel that the better solution would be to restrict the potential types that exposed commands can take rather than trying to build in more and more special cases in our code. For example, `lazy.run_extension` takes an extension. However, if we defined `extensions = [...]` in the config then we can just pass a string name to the `run_extension` (or we make extensions command objects and do `lazy.extension["extension_name"].run()`.

.github/workflows/release.yml
tych0: you're right here and below, just an oversight that I left them out.

tych0:
> Is lift_arg being run even when those commands are called internally? We should limit this to calls over IPC only.

I hope so? I'm Pretty Sure that interface.py is the right place to put it so that it only runs on IPC calls, but I'm not completely sure.

> I still feel that the better solution would be to restrict the potential types that exposed commands can take rather than trying to build in more and more special cases in our code. For example, lazy.run_extension takes an extension. However, if we defined extensions = [...] in the config then we can just pass a string name to the run_extension (or we make extensions command objects and do lazy.extension["extension_name"].run().

Right now the way I've done it passing an extension by lazy should Just Work (since it's only run on IPC calls, not in-process calls via lazy.*), but it definitely will not work correctly if we do an IPC version of the same `client.run_extension()` call. Perhaps we should send back IPC failures for the cases where we know we aren't correctly hoisting types?

tych0:
Hmm. Looks like interface.py isn't the right place:

```
2024-04-01 18:54:00,060 ERROR libqtile core.py:_xpoll():L372 Got an exception in poll loop
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 347, in _xpoll
    self.handle_event(event)
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 314, in handle_event
    ret = target(event)
          ^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 663, in handle_KeyPress
    key, handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/core/manager.py", line 452, in process_key_event
    status, val = self.server.call(
                  ^^^^^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/command/interface.py", line 390, in call
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/command/interface.py", line 329, in lift_arg
  ```
  
  I will try and figure out the right place.

elParaguayo:
> ... it definitely will not work correctly if we do an IPC version of the same `client.run_extension()` call. Perhaps we should send back IPC failures for the cases where we know we aren't correctly hoisting types?

Personally, I think if we're exposing commands as an external API then we should have consistent behaviour regardless of whether the command is run in-process or via IPC.

An alternative could be to disable IPC calls for those commands with "difficult" types. This is essentially the same as your last point.

I'm inclined to look at all the exposed commands and see how many have those difficult types and see if we can rewrite them in a way that means they can be called via IPC (this would need migrations too). I've already got extensions working as command objects (well, almost) so you can do `qtile cmd-obj -o extension extension_name -f run`.

elParaguayo:
I might push on with the `Extension` conversion as I've also fixed the issue where the extension can cause IPC to crash due to a timeout.

tych0:
> Personally, I think if we're exposing commands as an external API then we should have consistent behaviour regardless of whether the command is run in-process or via IPC.

yeah, i agree in principle. but people want to use these things over the CLI. that essentially restricts us to types which can be constructed from a single string when using extensions, which IMO is not really necessary. If people want to use complex types in `lazy.*`, they should be allowed to. but maybe we strive to design all the APIs in such a way that they don't have to?

anyway, not something we have to agree on to land this now, since the vast majority of things are indeed base types, but perhaps worth writing down somewhere once we decide.

tych0:
Hmm... actually it [looks like](https://github.com/qtile/qtile/blob/master/libqtile/ipc.py#L124) we probably already support passing real python objects via IPC. So I wonder if this code should live in a different place: in the `cmd-obj` parser.

We'd have to resolve the method in qtile/libqtile (basically we'd copy/paste `lift_arg` somewhere else). But then that means the only people who are doing this lifting are CLI users. Everyone else would have to pass the right thing in their python code or their lazy.*, but that seems fine to me.

Thoughts?

tych0:
Ok, I believe this is hopefully worth another look. Quoting the commit message for the appraoch:

> This works by adding a new parameter to the IPC protocol (perhaps we should
> document this somewhere? but it doesn't need to be stable IMO, since
> everyone will presumably use libqtile to implement it) that's a boolean
> flag. If true, qtile will attempt to invoke the type lifting code, and it
> won't otherwise.

> This means that users can choose whether or not they want to invoke the
> type lifting code, or set it to false and use the struct marshalling as
> before. Throughout the tree we have a few callers, I have hard coded them
> all to false with the notable exception of the `cmd-obj` subcommand, which
> is the one place I believe we (currently) want to do this.

tych0:
Ping - any thoughts here? I'd like to merge this (I've been running it for a while with no badness), but it is a reasonably large change, so would be good to have some more eyes.

.github/workflows/release.yml
elParaguayo: We do need 3.12 in the release script but this isn't the right PR for that.

elParaguayo (approved):
LGTM - I don't see a problem with merging this. The lifting code isn't called when doing lazy calls etc so there shouldn't be any impact on core behaviour.

Thanks for your work on this.

tych0:
Cool, thank you. I realized I need a changelog entry for dropping 3.9 support, so I will edit the branch protections, push that change log, and then enable automerge.

PR #4744, opened 2024-03-30
Blurry bar: quick fix

frosklis:
This fixes the mypy issues at https://github.com/qtile/qtile/pull/4625

### DIFF
--- a/libqtile/backend/wayland/drawer.py
+++ b/libqtile/backend/wayland/drawer.py
@@ -74,10 +74,12 @@ class Drawer(drawer.Drawer):
             height = self.height
         if height > self._win.height - offsety:
             height = self._win.height - offsety
+        scale = self.qtile.config.wl_scale_factor
 
         # Paint recorded operations to our window's underlying ImageSurface
         with cairocffi.Context(self._win.surface) as context:
             context.set_operator(cairocffi.OPERATOR_SOURCE)
+            context.scale(scale, scale)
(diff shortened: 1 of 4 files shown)


elParaguayo:
Thanks. I'm not going to merge this as the fix needs to be in the same PR.

When I run `pre-commit` on that other PR I only get two failures:
```
libqtile/backend/wayland/drawer.py:90: error: Unsupported operand types for * ("None" and "int")  [operator]
libqtile/backend/wayland/drawer.py:90: note: Left operand is of type "int | None"
```
That's happening because mypy hasn't noticed that we've excluded the `None` possibility in the lines above. We could just fix that with something like this:
```diff
diff --git a/libqtile/backend/wayland/drawer.py b/libqtile/backend/wayland/drawer.py
index 70ab88ad..9bd4d4e6 100644
--- a/libqtile/backend/wayland/drawer.py
+++ b/libqtile/backend/wayland/drawer.py
@@ -76,6 +76,9 @@ class Drawer(drawer.Drawer):
             height = self._win.height - offsety
         scale = self.qtile.config.wl_scale_factor

+        assert width is not None
+        assert height is not None
+
         # Paint recorded operations to our window's underlying ImageSurface
         with cairocffi.Context(self._win.surface) as context:
             context.set_operator(cairocffi.OPERATOR_SOURCE)
```
However, it's these kind of things that annoy me about mypy as it doesn't feel like they should be necessary.

PR #4750, opened 2024-04-04
Updates to Extensions

elParaguayo:
Draft for now.

This PR makes a few changes to extensions:
- extensions are now `CommandObject`s i.e. they can be accessed via qtile's command graph so we can do `lazy.extension["extensionname"].run()`
- extensions are run asynchronously which prevents the IPC server from timing out
- docs are updated to show extensions on the command graph
- tests are updated

Outstanding:
- Need to think of best way to have extensions defined in the config.
- Needs migrations to fix property name clashes

### DIFF
--- a/docs/qtile_docs/graph.py
+++ b/docs/qtile_docs/graph.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 from dataclasses import dataclass
+from math import cos, pi, sin
 
 from docutils.parsers.rst import Directive, directives
 from qtile_docs.base import SimpleDirectiveMixin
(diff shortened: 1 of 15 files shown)


tych0:
> Need to think of best way to have extensions defined in the config.

```
extensions = {
    "foo": FooExtension(...),
}
```

in the top level config maybe?

docs/qtile_docs/graph.py
```diff
@@ -58,54 +59,80 @@ def node_args(self, enabled=True, highlight=False, relative_url=str()):
         }
 
 
+def calc_vertex_position(number_nodes, include_origin=True, radius=2, offset=0, rounding=2):
+    """
+    Generator to calculate x, y coordinates for vertices on a regular polygon.
```
tych0, line 64: can you make this comment `regular inscribed polygon`? i had to stare at your math for a while to understand what was happening.

libqtile/extension/base.py
```diff
@@ -100,7 +110,7 @@ class RunCommand(_Extension):
         #       among all the objects inheriting this class, and if one of them
         #       modified it, all the other objects would see the modified list;
         #       use a string or a tuple instead, which are immutable
-        ("command", None, "the command to be launched (string or list with arguments)"),
+        ("cmd", None, "the command to be launched (string or list with arguments)"),
```
tych0, line 113: is there a motivation for this change?

libqtile/extension/command_set.py
```diff
@@ -62,24 +62,29 @@ class CommandSet(Dmenu):
     """
 
     defaults = [
-        ("commands", None, "dictionary of commands where key is runable command"),
+        ("commandset", None, "dictionary of commands where key is runable command"),
```
tych0, line 65: commanddict? but can we just leave this as commands, or is there a technical reason not to?

elParaguayo: Yes - it's an annoying clash.

And, yes, it's a dict. Will fix.

PR #4757, opened 2024-04-06
fix race in starup/restart logic

### LINKED ISSUE #4755: has no attribute current_window
### Issue description

I'm getting "has no attribute current_window" in logs and not sure why

### Version

0.25.0

### Backend

X11 (default)

### Config

_No response_

### Logs

```text
2024-04-06 20:40:51,224 ERROR libqtile core.py:_xpoll():L372 Got an exception in poll loop
  File "/home/nazar-pc/.local/lib/python3.10/site-packages/libqtile/backend/x11/core.py", line 347, in _xpoll
    self.handle_event(event)
  File "/home/nazar-pc/.local/lib/python3.10/site-packages/libqtile/backend/x11/core.py", line 314, in handle_event
    ret = target(event)
  File "/home/nazar-pc/.local/lib/python3.10/site-packages/libqtile/backend/x11/core.py", line 781, in handle_DestroyNotify
    if self.qtile.current_window is None:
  File "/home/nazar-pc/.local/lib/python3.10/site-packages/libqtile/command/base.py", line 281, in __getattr__
    raise AttributeError(f"{self.__class__} has no attribute {name}")
AttributeError: <class 'libqtile.core.manager.Qtile'> has no attribute current_window
```

### Required

tych0:
I am getting:

2024-04-06 13:38:10,669 ERROR libqtile core.py:_xpoll():L372 Got an exception in poll loop Traceback (most recent call last):
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 347, in _xpoll
    self.handle_event(event)
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 314, in handle_event
    ret = target(event)
          ^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/backend/x11/core.py", line 663, in handle_KeyPress
    key, handled = self.qtile.process_key_event(keysym, event.state & self._valid_mask)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/core/manager.py", line 449, in process_key_event
    if cmd.check(self):
       ^^^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/lazy.py", line 148, in check
    cur_win_floating = q.current_window and q.current_window.floating
                       ^^^^^^^^^^^^^^^^
  File "/home/tycho/.local/lib/python3.11/site-packages/libqtile/command/base.py", line 281, in __getattr__
    raise AttributeError(f"{self.__class__} has no attribute {name}")
AttributeError: <class 'libqtile.core.manager.Qtile'> has no attribute current_window

on restart.

current_screen is first set in:

loop
  -> async_loop
    -> load_config()
      -> _process_screens()

but in async_loop(), we setup_listener() before we load the config, i.e.:

loop
  -> async_loop
    -> setup_listener() # whoops
    ...
    -> load_config()
      -> _process_screens()

This means that we could potentially race and start receiving events before we were completely configured, as above. This is especially likely if we are restarting, but the race exists as well on startup and could be present if e.g. people were starting things in their xsessions before exec()ing qtile.

To fix this, we load the config before we setup the listener, so that everything will be present. But there's a bit of a circular dependency we need to break here: setup_listener() is the one that initialize's the Core's copy of the qtile object. Let's just set that explicitly in Core's async_loop(), and remove the parameter of setup_listener(). So then we have:

1. initialize qtile object
2. load_config()
3. setup_listener()

and hopefully we won't see more races like this. Indeed, what clued me in that it was a race vs just a bug is that I saw other missing attributes sometimes on restarts e.g.:

    AttributeError: <class 'libqtile.backend.x11.window.Window'> has no attribute has_focus

This is an interesting bug: it seems like it has been around for quite a long time, but perhaps the new asyncio and python speed improvements are making it such that now we can actually see it?

Fixes #4755

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -44,8 +44,6 @@ from libqtile.utils import QtileError
 if TYPE_CHECKING:
     from typing import Callable, Iterator
 
-    from libqtile.core.manager import Qtile
-
 _IGNORED_EVENTS = {
     xcffib.xproto.CreateNotifyEvent,
     xcffib.xproto.FocusInEvent,
@@ -151,7 +149,6 @@ class Core(base.Core):
         # setup the default cursor
         self._root.set_cursor("left_ptr")
 
-        self.qtile = None  # type: Qtile | None
         self._painter = None
         self._xtest = self.conn.conn(xcffib.xtest.key)
 
@@ -181,7 +178,8 @@ class Core(base.Core):
                 self._root.wid,
                 self.conn.atoms["_NET_SUPPORTING_WM_CHECK"],
             ).check()
-        self.qtile = None
+        if hasattr(self, "qtile"):
+            delattr(self, "qtile")
         self.conn.finalize()
 
     def get_screen_info(self) -> list[tuple[int, int, int, int]]:
@@ -211,7 +209,7 @@ class Core(base.Core):
         self._wmname = wmname
         self._supporting_wm_check_window.set_property("_NET_WM_NAME", wmname)
 
-    def setup_listener(self, qtile: "Qtile") -> None:
+    def setup_listener(self) -> None:
         """Setup a listener for the given qtile instance
 
         :param qtile:
@@ -220,7 +218,6 @@ class Core(base.Core):
             The eventloop to use to listen to the file descriptor.
         """
         logger.debug("Adding io watch")
-        self.qtile = qtile
         self.fd = self.conn.conn.get_file_descriptor()
         asyncio.get_running_loop().add_reader(self.fd, self._xpoll)
 
--- a/libqtile/backend/base/core.py
+++ b/libqtile/backend/base/core.py
@@ -19,6 +19,7 @@ if typing.TYPE_CHECKING:
 class Core(CommandObject, metaclass=ABCMeta):
     painter: Any
     supports_restarting: bool = True
+    qtile: Qtile
 
     @property
     @abstractmethod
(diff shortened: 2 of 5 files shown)


tych0:
Arg, mypy really doesn't like:
```
libqtile/core/manager.py:213: error: "Core" has no attribute "qtile"  [attr-defined]
```

tych0:
Ok, I think I finally got it. But wayland people take a look; I think it simplifies the code greatly to just use `Qtile` instead of `Qtile | None` here, but perhaps there's some dependency I don't understand.

libqtile/backend/x11/core.py
```diff
@@ -211,7 +211,10 @@ def wmname(self, wmname):
         self._wmname = wmname
         self._supporting_wm_check_window.set_property("_NET_WM_NAME", wmname)
 
-    def setup_listener(self, qtile: "Qtile") -> None:
+    def set_qtile(self, qtile: "Qtile") -> None:
+        self.qtile
```
elParaguayo, line 215: This doesn't look right.

tych0: Yep, thanks. leftover from when I was considering using a setter; I deleted it.

tych0:
I'm gonna merge this since it fixes a real bug. Hopefully it doesn't break the wayland stuff too hard :)

jwijenbergh:
> I'm gonna merge this since it fixes a real bug. Hopefully it doesn't break the wayland stuff too hard :)

:( #4767 seems to be related to this

PR #4758, opened 2024-04-07
Fix window matching on ScreenSplit layout

elParaguayo:
There was a bug where the matching rules for nested layouts were never actually being checked. This PR fixes that and adds a test for it.


nazar-pc (approved):
This works for me, thanks!

elParaguayo:
Thanks for confirming.

PR #4760, opened 2024-04-11
Add Plasma layout

elParaguayo:
Adds `Plasma` layout from https://github.com/numirias/qtile-plasma.

DRAFT FOR NOW:
This is a beast of a layout but it is pretty clever (kudos to @numirias). I broadly understand how it works (the magic is in the `Node` objects) but I'm still checking some bits.

I think I've got this working ok (specific tests should all pass) but there's more I'd like to do before merging:
- Fix the inevitable test failures
- Better documentation (particularly explaining how methods work)
- Look at configuration options
- Add type hinting
- Probably some other stuff too


libqtile/layout/plasma.py
```diff
@@ -0,0 +1,1008 @@
...
+        if self.payload is payload:
+            return self
+        for child in self:
+            needle = child.find_payload(payload)
+            if needle is not None:
+                return needle
+        return None
+
+
+class Plasma(Layout):
```
jwijenbergh, line 730: Maybe add the fact that this was forked and upstreames from the original qtile plasma layout with a link to it. That way some better credit is given apart from the copyright header on the top.

elParaguayo: Agreed. I want to add much more on documentation, docstrings, comments etc before merging

jwijenbergh: Thanks! I was shocked at the line count at first but 1.6k in tests is really nice! I would say 1k for the layout is manageable. So I think this is indeed a good candidate for up streaming, especially because a bunch of users started with i3

elParaguayo: Yeah. The amount of tests here is really impressive. We've got to 100% coverage of the layout.

elParaguayo:
@jrwrigh How do you tell where your next window is going to open? There's no indicator to show if it will be horizontal or vertical. 

Would it be helpful to have a means to do this? either a widget (that's invisible on other layouts) or a different coloured border?

libqtile/layout/plasma.py
```diff
@@ -0,0 +1,1008 @@
...
+
+    @property
+    def is_root(self):
+        return self.parent is None
+
+    @property
+    def is_leaf(self):
+        # self is True if self.children is not an empty list (due to __len__)
+        # self.is_leaf therefore returns True if the node has no children
+        return not self
```
tych0, line 165: This is a little weird; why not just test len(self) instead of having a comment here to explain that we're testing len(self)? :)

tych0:
Overall looks reasonable to me, modulo my one small nit. Can you add a CHANGELOG entry too?

libqtile/layout/plasma.py
elParaguayo: By having the `__len__` dunder method, the class can be tested for truthiness directly. I just added that note to explain why the existing code worked. But, you're right, we could explicitly check `len(self)`!

jrwrigh:
> @jrwrigh How do you tell where your next window is going to open? There's no indicator to show if it will be horizontal or vertical. 
> 
> Would it be helpful to have a means to do this? either a widget (that's invisible on other layouts) or a different coloured border?


I don't, I've just gotten used to repositioning windows if they end up in a unexpected location. 🙃

I think I'd prefer an indicator widget over a change in the border color. I think i3 will color the border of a specific side differently, which would be pretty nice too, but I'd imagine more tricky.

elParaguayo:
Thanks. I might see if I can do a quick widget too here then.

The border idea would be the nicest. It is possible but it's too much work at this stage!

nazar-pc:
> @jrwrigh How do you tell where your next window is going to open? There's no indicator to show if it will be horizontal or vertical.

I believe I addressed this in https://github.com/numirias/qtile-plasma/pull/26 that checks proportions and splits tall windows horizontally and wide windows vertically.

elParaguayo:
I can add that as an option.

elParaguayo:
Right. I'd suggest we squash this as is to get it merged. There are some enhancements that can be added but they might take me longer to get round to.

jwijenbergh:
> @jrwrigh How do you tell where your next window is going to open? There's no indicator to show if it will be horizontal or vertical.
> 
> Would it be helpful to have a means to do this? either a widget (that's invisible on other layouts) or a different coloured border?

I would be a big fan of both, but especially the coloured border. That is actually pretty similar to how i3/sway do it

elParaguayo:
Borders would be good but we need to change the border rendering to allow more complex borders. I'm happy to look at this but let's have it as a separate PR.

nazar-pc:
That seems to be too complex to me. https://github.com/numirias/qtile-plasma/pull/26 add a natural rule that doesn't require additional visual indication (for me at least).

elParaguayo:
I would make your PR configurable. I agree it's nice.

I think a visual cue for the next direction would be nice and I want more options around borders anyway.

jwijenbergh:
> Borders would be good but we need to change the border rendering to allow more complex borders. I'm happy to look at this but let's have it as a separate PR.

at least for wayland it's trivial, not sure about x11. E.g. on wayland every border is an individual rectangle so we can just change color for one side

elParaguayo:
It's slightly less trivial on x11 but I've done it before as a proof of concept (doing a gradient border).

jwijenbergh:
> It's slightly less trivial on x11 but I've done it before as a proof of concept (doing a gradient border).

we can do this in a follow up pr anyways, if you have the x11 implementation later I will add the wayland one

libqtile/layout/plasma.py
```diff
@@ -0,0 +1,1085 @@
...
+    def __setitem__(self, key, value):
+        self.children[key] = value
+
+    def __len__(self):
+        return len(self.children)
+
+    @property
+    def root(self):
+        try:
+            # Walk way down treet until we find root
```
jwijenbergh, line 154: treet -> the tree? Also doesn't this walk 'up' the tree once?

libqtile/layout/plasma.py
```diff
@@ -0,0 +1,1085 @@
...
+    def top_right(self):
+        return Point(self.x + self.width, self.y)
+
+    @property
+    def bottom_left(self):
+        return Point(self.x, self.y + self.height)
+
+    @property
+    def bottom_right(self):
+        return Point(self.x + self.width, self.y + self.height)
```
jwijenbergh, line 339: these are only here for testing code (and probs other properties too), we can just e.g. self.x and self.y + self.height in the test itself

libqtile/layout/plasma.py
elParaguayo: Well, yes, it does go up but I'd expect the "root" to be at the bottom of a "tree" which is why I used the word down.

"treet" is a typo though.

libqtile/layout/plasma.py
elParaguayo: Agreed.

tych0 (approved):

Issue #4762, opened 2024-04-13
0.25.0: Five failing tests


dvzrv:
### Issue description

Hi :wave: 

I'm currently upgrading the package on Arch Linux to 0.25.0 and ran into failing tests:

```
=========================== short test summary info ============================
FAILED test/backend/x11/test_window.py::test_urgent_hook_fire[wayland-2] - Ke...
FAILED test/widgets/test_clock.py::test_clock_datetime_timezone - AssertionEr...
FAILED test/widgets/test_clock.py::test_clock_pytz_timezone - AssertionError:...
FAILED test/widgets/test_clock.py::test_clock_dateutil_timezone - AssertionEr...
FAILED test/widgets/test_clock.py::test_clock_change_timezones - AssertionErr...
= 5 failed, 1726 passed, 71 skipped, 4 xpassed, 2 warnings in 1017.94s (0:16:57) =
```

[qtile-0.25.0-1-x86_64-check.log](https://github.com/qtile/qtile/files/14965804/qtile-0.25.0-1-x86_64-check.log)
[qtile-0.25.0-1-x86_64-build.log](https://github.com/qtile/qtile/files/14965805/qtile-0.25.0-1-x86_64-build.log)

In regards to how tests are run, nothing has been changed since last upgrade:
https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/blob/c9582a5362b017c90a6f3c06b57003dd1326901b/PKGBUILD

### Version

0.25.0

### Backend

X11 (default), Wayland (experimental)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

tych0:
Hmm, first one looks a little strange to me, can you try the above PR and see if that fixes it? The rest all look like "the same" bug, I will take a look at those in a bit.

elParaguayo:
Fair enough. Let me take a look.

tych0:
The math looks like it is pretty clearly ignoring the timezone delta, but I really don't understand why that would happen.

elParaguayo:
If I run `pkgctl build --testing` to build this in a chroot, all those clock tests pass successfully for me.

jwijenbergh:
Fwiw the clock tests also fail for me on my machine

jwijenbergh:
[...]

PR #4763, opened 2024-04-13
tests: make $DISPLAY available for x11

tych0:
I am not really sure why this passes in our CI ($DISPLAY wouldn't be set there I wouldn't think?), but hopefully this fixes it.

This is part of #4762


tych0:
I think I figured out what the problem is, and it's not this.

PR #4768, opened 2024-04-15
Wayland: Fix self.qtile startup logic

### LINKED ISSUE #4767: [wayland] Keyboard stopped working on master
### Issue description

I updated to the newest version and found myself not being able to use the keyboard (using wayland). The mouse still worked.
I created the issue for awareness and will help debug this in the next few days.
Not working on: 0.25.1.dev5+ga890670d
Reverted back to: 0.23.1.dev65+g2e89b757 (working here, with the rest of the system identical)

Gentoo Kernel: 6.8.6-gentoo-dist #1 SMP PREEMPT_DYNAMIC Sat Apr 13 15:30:53 -00 2024 x86_64 12th Gen Intel(R) Core(TM) i9-12900H GenuineIntel GNU/Linux

Thank you for your hard work on this!

### Version

master -  0.25.1.dev5+ga890670d

### Backend

Wayland (experimental)

### Config

_No response_

### Logs

_No response_

### Required

jwijenbergh:
In #4757 we have fixed some of the logic with regards to setting self.qtile. This means that in certain scenarios the self.qtile attribute does not exist, e.g. when we're adding keyboards before the config is loaded.

We already mark these devices as "pending" if
self.qtile is `None`, but this now does not work as the attribute is not set in the first place. Let's check with hasattr and remove the unneeded self.qtile setting inside of the keyboard constructor.

This fixes #4767

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -446,7 +446,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.seat.set_capabilities(capabilities)
 
         logger.info("New device: %s %s", *device.get_info())
-        if self.qtile:
+        if hasattr(self, "qtile"):
             if self.qtile.config.wl_input_rules:
                 device.configure(self.qtile.config.wl_input_rules)
         else:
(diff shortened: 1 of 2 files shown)


tych0:
oops, thanks :)

PR #4775, opened 2024-04-21
don't use wayland backend in x11 tests

tych0:
We were previously using `manager_nospawn`, which is a fixture that enables the test for both backends, even though this test is x11 specific. You can see this with the failed test name+params here:

    FAILED test/backend/x11/test_window.py::test_urgent_hook_fire[wayland-2]

instead, let's introduce our own xmanager_nospawn, and use that.

This is the other half of #4762

### DIFF
--- a/test/backend/x11/conftest.py
+++ b/test/backend/x11/conftest.py
@@ -170,6 +170,21 @@ def xmanager(request, xephyr):
 
 
 @pytest.fixture(scope="function")
+def xmanager_nospawn(request, xephyr):
+    """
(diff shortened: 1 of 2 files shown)


elParaguayo:
Am I right in thinking this works in CI because of the environmental variables that we set but running locally has an issue?

tych0:
Yeah, I think that's exactly it. It works in CI because we only manually use a subset of the fixtures, but if you try to run all the tests with all the dependencies locally, I think it will fail. (Clearly nobody ever does that on our dev team :)

PR #4776, opened 2024-04-22
Upgrade actions/stale version from v2.0.0 to v9

hawl1:
No one wants their Github Actions to not work, especially if its in a big project.

### DIFF
--- a/libqtile/layout/screensplit.py
+++ b/libqtile/layout/screensplit.py
@@ -120,7 +120,7 @@ class ScreenSplit(Layout):
         self._split_index = 0
         self.layouts = {}
         self._move_win = None
-        self._has_matches = False
+        self._has_matches = None
         print(self.splits)
         splits = []
         for s in self.splits:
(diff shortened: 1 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         stale-pr-label: 'status: stale'
         days-before-stale: 90
         days-before-close: 30
-    - uses: actions/stale@v2.0.0
+    - uses: actions/stale@v9.0.0
```
tych0, line 20: Can we use v9 here to get any point releases? (i.e. without the .0.0?)

hawl1: yes, we can. let me change it real quick...

hawl1: Alright, done.

tych0:
Hmm. I tried to squash this and it closed the PR?

tych0:
Ok, I dunno what happened there, but I manually pushed it to master: 3e7284b0514e47eb700900fcc037812b86b8481e

Thanks!

hawl1:
github being github lmao

PR #4151, opened 2023-02-07
Volume widget display change

Ahmed-Jarir:
changed volume widget to be able to specify the format when the volume is muted and unmuted

PR #4779, opened 2024-04-26
Add mute and unmute formats for Volume and PulseVolume widgets

shyguyCreate:
Rewrite of #4151

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -108,7 +108,7 @@ class Volume(base._TextBox):
         self.add_defaults(Volume.defaults)
         self.surfaces = {}
         self.volume = None
-        self.mute = 0
+        self.mute = False
 
         self.add_callbacks(
             {
@@ -217,13 +217,13 @@ class Volume(base._TextBox):
 
             mixer_out = subprocess.getoutput(get_volume_cmd)
         except subprocess.CalledProcessError:
-            return -1, 0
+            return -1, False
 
         check_mute = mixer_out
         if self.check_mute_command:
             check_mute = subprocess.getoutput(self.check_mute_command)
 
-        mute = 1 if self.check_mute_string in check_mute else 0
+        mute = self.check_mute_string in check_mute
 
         volgroups = re_vol.search(mixer_out)
         if volgroups:
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -76,6 +76,8 @@ class Volume(base._TextBox):
             " List contains 4 symbols, from lowest volume to highest.",
         ),
         ("mute_command", None, "Mute command"),
+        ("mute_format", "M", "Format to display when volume is muted."),
+        ("unmute_format", "{volume}%", "Format of text to display when volume is not muted."),
         ("volume_app", None, "App to control volume"),
         ("volume_up_command", None, "Volume up command"),
         ("volume_down_command", None, "Volume down command"),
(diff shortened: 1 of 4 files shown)


shyguyCreate:
Help. Can someone point me out how to get the PulseAudio volume value when mute, it always returns `-1`.

I can't find anything in [pulse_volume.py](https://github.com/qtile/qtile/blob/master/libqtile/widget/pulse_volume.py).

libqtile/widget/volume.py
```diff
@@ -221,7 +224,9 @@ def get_volume(self):
             check_mute = subprocess.getoutput(self.check_mute_command)
 
         if self.check_mute_string in check_mute:
-            return -1
+            self.mute = 1
+        else:
+            self.mute = 0
```
tych0, line 229: Can you use python's True + False for these literals?

tych0: Better yet, can you make `get_volume()` return a tuple that's the volume and whether or not the widget is muted? You can override it in your test with some monkey patching.

elParaguayo:
Agree with @tych0 - this needs some work. Having `get_volume()` return a tuple of the volume level and whether the device is muted or not seems like the correct approach.

elParaguayo (changes requested):
I think this would be a nice addition to the code. However, I'm marking this one as "request changes` for now because:
a) it needs changes as suggested by tych0
b) the volume widgets are likely to be used by a large portion of the userbase so I want to take a close look before this gets merged.

elParaguayo:
Why close this?

shyguyCreate:
Sorry I panic, I think I messed up the commit history.

shyguyCreate:
For the record, I did messed up the history right? Because I see 10+ commits that are not mine.

elParaguayo:
If you look at this PR page you'll see it says it's only adding one commit. The other ones were probably because you did a merge to bring the branch in line with the current master. However, github knows those commits aren't needed here.

shyguyCreate:
Next time I have to update the master I'll ask before undoing everything.

elParaguayo:
Just rebase onto master rather than try to merge changes into the branch.

shyguyCreate:
Question: I have heard bad things about using `--force` on push, and after playing with git rebase and getting my commit on top, I really need to use `--force` with git push?

elParaguayo:
There are definitely times when you shouldn't use force (eg when pushing to a branch that other users are tracking and making changes to). However, for a PR here, we're happy for you to use force.

shyguyCreate:
All right, thanks

libqtile/widget/volume.py
```diff
@@ -214,21 +217,20 @@ def get_volume(self):
 
             mixer_out = subprocess.getoutput(get_volume_cmd)
         except subprocess.CalledProcessError:
-            return -1
+            return -1, 0
```
tych0, line 220: Can you switch to True/False here? Other than that looks good to me.

shyguyCreate: What do you mean exactly? Do you want to use True/False in self.mute instead of 1/0?

tych0: yep, exactly

shyguyCreate: Done

tych0:
Thanks!

elParaguayo:
This broke the Volume widget. We had a method called `mute` but this is overwritten by defining a variable called `mute` during `__init__`. 

We need better tests!

shyguyCreate:
It should now be fixed in the new PR.

PR #4780, opened 2024-04-27
move stack dump handler

### LINKED ISSUE #4778: flaky tests do not report anything interesting
### Issue description

See e.g.:

https://github.com/qtile/qtile/actions/runs/8801434359/job/24154761197

reproduced here for posterity: https://gist.github.com/tych0/8653a01f34005c30b563e08fe0f10c0f

I have tried via c4d3c6e602f9ed619086cc3942bd1c47ebabedbf (and fixing it in 6f2c2ade8d4ad0283848249ce354d00a79b63791) to dump the stack and see where we're hanging in this case, but it seems to have not worked.

I am mostly filing this so I remember to keep working on it, but any other help is welcome. Flaky tests annoy me :)

### Version

git master

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

tych0:
We set this up in main, but our test suite never calls qtile's main method, so my attempts at logging why tests were hanging would never succeed.

Instead, let's set this up in async_loop() where we set up the other signal handling, which is called by the test suite and hopefully as a result will yield stack traces when our test suite hangs.

Fixes #4778

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import asyncio
+import faulthandler
 import io
 import logging
 import os
@@ -213,6 +214,12 @@ class Qtile(CommandObject):
         self.core.qtile = self
         self.load_config(initial=True)
         self.core.setup_listener()
+
+        faulthandler.enable(all_threads=True)
+        # This is a bit unfortunate. We use SIGUSR1&2 for reloading config &
+        # restarting qtile, so we overload SIGWINCH here to dump threads.
+        faulthandler.register(signal.SIGWINCH, all_threads=True)
+
         try:
             async with LoopContext(
                 {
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks. This also seems to fix the weird issue I had with errors appearing in `qtile shell`.

tych0:
> This also seems to fix the weird issue I had with errors appearing in qtile shell.

oh, duh. because when you resize your terminal, that sends a SIGWINCH which we were also using to dump stacks. I wonder: should we use SIGUSR2 instead? We were previously using it for reloading, but I bet that nobody was actually using it for reloading (probably everyone just picked SIGUSR1), so I don't think it would break anything. And it wouldn't have any strange unintended consequences like SIGWINCH does.

elParaguayo:
SIGUSR2 would make more sense. There's no real reason for using it to reload the config as that can be triggered via IPC anyway.

tych0:
Ok, I will send a patch if you don't beat me to it :)

elParaguayo:
All yours, I'm working on something else!

PR #4782, opened 2024-04-28
Add ability to override floating rules

elParaguayo:
Some float rules (e.g. fixed size/aspect ratio) can be too broad and users may want exceptions to these rules. It is possible to do this with hooks but they can become unwieldy with multiple exceptions.

This PR adds an `override_rules` argument to the floating layout. Where a window matches one of these rules then window will not be floated.

Question for @tych0 - any idea where the test fails if the windows are added in the other order? i.e. if I open the `nofloat` window second, the focus remains on the first one so `manager.c.window` points to the wrong window.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Qtile x.xx.x, released XXXX-XX-XX:
         the `qtile cmd-obj` command line
       - Add `Plasma` layout. The original layout (https://github.com/numirias/qtile-plasma) appears to be unmaintained
         so we have added this to the main codebase.
+      - Add ability to define exceptions to rules for floating windows.
     * bugfixes
 
 Qtile 0.25.0, released 2024-04-06:
--- a/libqtile/layout/floating.py
+++ b/libqtile/layout/floating.py
@@ -65,6 +65,8 @@ class Floating(Layout):
         Match(func=lambda c: c.has_fixed_ratio()),
     ]
 
+    default_override_rules: list[Match] = []
+
     defaults = [
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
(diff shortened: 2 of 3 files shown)


tych0:
I think it's "intentional", i.e. when we add a window underneath a floating window, the floating window remains focused. But,

> +         return matches and not overrides

this whole patch seems to be because our Match rules can't express this `and not` pattern. Can we make it possible for matches to express that instead somehow? Otherwise potentially everywhere we use match will need a patch like this.

elParaguayo:
Good idea. Let me rework this.

elParaguayo:
Thinking about it some more, do we need exceptions for other uses of `Match`? The float rules have some very broad rules which may benefit from some exceptions. But, for other times when users provide Match rules, they're typically very focused already.

elParaguayo:
The alternative would be to allow a Match object to take a list of Match objects to treat as exceptions to the parent.

tych0:
> Thinking about it some more, do we need exceptions for other uses of Match? The float rules have some very broad rules which may benefit from some exceptions. But, for other times when users provide Match rules, they're typically very focused already.

IMO having exceptions for float rules is also weird, but I don't use floating at all, so maybe I don't know. What specifically are the exceptions that people are interested in? Is it because the default rules are too broad?

elParaguayo:
The float rules are deliberately broad eg they capture all windows requesting a fixed aspect ratio but you may want to carve out some specific apps.

You could do this with a user defined function but then you're asking users to use non documented methods to retrieve properties of windows. 

This was just an idea to make it a bit easier for users.

elParaguayo:
A "fun" way to do it would be to use bitwise operators e.g. `(Match(wm_class="notification")) | Match(func=lambda c: c.has_fixed_size())) ~ Match(wm_class="mpv")` but that's far less intuitive for users and is probably pretty complicated to code!

tych0:
Honestly, I like it. A DSL for window matching. I guess you mean `(Match(wm_class="notification")) | Match(func=lambda c: c.has_fixed_size())) & ~Match(wm_class="mpv")` (additional `&`) in your example?

I think this would be great :)

elParaguayo:
Yes - missed the `&` sign.

OK - I might have a play at implementing this.

elParaguayo:
I've actually managed to get this working. I still need to write some tests and fix the mypy stuff so will upload later this week.

PR #4787, opened 2024-04-28
drop libqtile.command_* migration

tych0:
in accordance with the new policy, we can drop this migration as well. People who still want this code can use 0.25.0 to migrate old config files if they want.


tych0:
Uh oh. This is the same seg fault as in #4785

jwijenbergh:
> Uh oh. This is the same seg fault as in #4785

Yes very strange. But #4786 does pass CI, maybe it's flaky somehow?

tych0:
Looks like it is https://github.com/pypy/pypy/issues/3959

Seems like it should be safe (if annoying) to re-run until we get green in that case.

PR #4789, opened 2024-04-29
Add Plasma widget

elParaguayo:
This PR adds a `Plasma` widget to display direction for next window in the `Plasma` layout.

I've moved the border customisation stuff to qtile-extras so there's a modified version of the layout there which allows for border highlighting (see https://github.com/elParaguayo/qtile-extras/pull/341).

### DIFF
--- a/libqtile/layout/plasma.py
+++ b/libqtile/layout/plasma.py
@@ -24,8 +24,22 @@ from enum import Enum, Flag, auto
 from math import isclose
 from typing import NamedTuple
 
+from libqtile import hook
 from libqtile.command.base import expose_command
+from libqtile.hook import Hook, qtile_hooks
 from libqtile.layout.base import Layout
+
+plasma_hook = Hook(
+    "plasma_add_mode",
+    """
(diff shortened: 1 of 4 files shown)


libqtile/hook.py
```diff
@@ -1024,6 +1024,14 @@ def hooked_function():
         """,
         _user_hook_func,
     ),
+    Hook(
+        "plasma_add_mode",
```
tych0, line 1028: can we do this from inside plasma.py, something like:

```
from libqtile.hook import hooks
...
hooks.register_hook(Hook(...))
```
?

tych0: that is, layouts/plasma.py

tych0 (approved):

libqtile/hook.py
elParaguayo: Yes - that would be neat. I don't think it will appear in the docs but that's fine as there's no real reason for users to need the hook.

elParaguayo:
@tych0 I've made your suggested change and will set this to automerge.

PR #4793, opened 2024-04-30
Add boolean logic to Match objects

elParaguayo:
This PR adds the ability to use bitwise operators on `Match` objects in order to build more complex rules.

Match objects now support `|`, `&`, `~` and `^` operators.

This allows you to define more complex rules. For example, the following rule would match all windows with a fixed aspect ratio *except* an `mpv` instance:
```python
Match(func=lambda c: c.has_fixed_ratio()) & ~Match(wm_class="mpv")
```

I should update the docs for this but would welcome thoughts.

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -832,7 +832,85 @@ def convert_deprecated_list(vals: list[str], name: str) -> re.Pattern:
     return re.compile(regex_input)
 
 
-class Match:
+class _Match:
+    """Base class to implement bitwise logic methods for Match objects."""
+
+    def compare(self, client: base.Window) -> bool:
+        return True
+
+    def __invert__(self) -> InvertMatch:
+        return InvertMatch(self)
+
+    def __and__(self, other: _Match) -> MatchAll:
+        if not isinstance(other, _Match):
+            raise TypeError
+
+        return MatchAll(self, other)
+
+    def __or__(self, other: _Match) -> MatchAny:
+        if not isinstance(other, _Match):
+            raise TypeError
+
+        return MatchAny(self, other)
+
+    def __xor__(self, other: _Match) -> MatchOnlyOne:
+        if not isinstance(other, _Match):
+            raise TypeError
+
+        return MatchOnlyOne(self, other)
+
+
+class InvertMatch(_Match):
+    """Wrapper to invert the result of the comparison."""
+
+    def __init__(self, match: _Match):
+        self.match = match
+
+    def compare(self, client: base.Window) -> bool:
+        return not self.match.compare(client)
+
+    def __repr__(self) -> str:
+        return "<InvertMatch(%r)>" % self.match
+
+
+class MatchAll(_Match):
+    """Wrapper to check if all comparisons return True."""
+
+    def __init__(self, *matches: _Match):
+        self.matches = matches
+
+    def compare(self, client: base.Window) -> bool:
+        return all(m.compare(client) for m in self.matches)
+
+    def __repr__(self) -> str:
+        return "<MatchAll(%r)>" % (self.matches,)
+
+
+class MatchAny(MatchAll):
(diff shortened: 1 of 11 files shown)


elParaguayo:
This is hanging tests on wayland but not x11. Very odd. I also had erratic behaviour with tests locally on x11. Seems to be related to opening windows but I'm not sure why.

libqtile/config.py
```diff
@@ -832,7 +832,82 @@ def convert_deprecated_list(vals: list[str], name: str) -> re.Pattern:
...
+    """Wrapper to invert the result of the comparison."""
+
+    def __init__(self, match: _Match):
+        self.match = match
+
+    def compare(self, client: base.Window) -> bool:
+        return not self.match.compare(client)
+
+    def __repr__(self) -> str:
+        return "<_InvertMatch(%r)>" % self.match
```
tych0, line 873: you don't need the `<>` here right?

tych0:
idk, IMO this looks pretty neat, I think we should land it (maybe with some docs of course).

the tests failing make me think this series of tests exercises whatever our sadness is that has been flaking randomly the last few months that I was trying to catch with the whole signal -> stack trace business. If one of them is hung, can you get a set of stacks to see what it's actually doing?

libqtile/config.py
elParaguayo: No. I've always done that out of habit if nothing else!

I was also thinking that we could make those wrapper classes public in case people wanted to use them instead of the operators.

elParaguayo:
> idk, IMO this looks pretty neat, I think we should land it (maybe with some docs of course).

I really like this. I think it's a good addition.

> the tests failing make me think this series of tests exercises whatever our sadness is that has been flaking randomly the last few months that I was trying to catch with the whole signal -> stack trace business. If one of them is hung, can you get a set of stacks to see what it's actually doing?

I'll have to investigate more but it's just this test that's hanging so my guess is I've done something odd somewhere.

tych0:
> I'll have to investigate more but it's just this test that's hanging so my guess is I've done something odd somewhere.

Ok, I'm happy to look too when I have time. The occasional hangs we see are quite annoying!

libqtile/config.py
tych0: Yep, agreed on making them public. If we drop the `<>`, I think you can copy paste them into a python interpreter, which is my understanding of what repr() is supposed to do. With `<>`, you'd have to manually delete them all.

jwijenbergh:
Love this!

elParaguayo:
Don't know how much it will get used but I think it's a really good addition.

libqtile/config.py
elParaguayo: All the other `repr`s in that file have the `<>`!

tych0: Ah, reading the actual docs, it's either or: https://docs.python.org/3/library/functions.html#repr

my read is: if you can't copy and paste it into eval() it should have <>, but otherwise it should not.

elParaguayo:
OK - have made wrappers public and think I've fixed the tests. 

Just need to do the docs now.

tych0:
Hm. we finally got our first CI flake with a stack trace: https://github.com/qtile/qtile/actions/runs/8901596066/job/24445775719

maybe a race in cffi?

elParaguayo:
Docs are done so should be good to go.

tych0:
I wonder: can we rewrite `float_rules` to be one match with `|` instead of a list? I guess we can, but should we?

elParaguayo:
> I wonder: can we rewrite `float_rules` to be one match with `|` instead of a list? I guess we can, but should we?
> 
> EDIT: in the default config, that is.

You can but I'm not sure it improves readibility. I suspect this PR is better suited at fine tuning specific rules, rather than as a means to rewrite existing ones.

PR #4794, opened 2024-04-30
Removed instructions for running inside Gnome

wingej0:
The instructions for running inside of Gnome no longer work with newer versions of Gnome, so I have removed them.


tych0:
I have no problem believing these are broken, but can you talk a little bit about why they're broken?

elParaguayo:
You'd also need to remove `gnome` from the toctree in `docs/manual/install/index.rst`.

wingej0:
I'm not sure which version of Gnome broke this configuration, but it doesn't work with any of the new versions.  I have tried on several occasions in the last few years to make this work because it sounds like the ultimate configuration to me, but I've run into several errors.  To my knowledge, swapping out Mutter for another WM isn't possible.  I'd be happy to write some docs on how to run Qtile in XFCE, which works great, but in Gnome doesn't work. 

I wish I had the technical details to add...

wingej0:
> You'd also need to remove `gnome` from the toctree in `docs/manual/install/index.rst`.

Done.

PR #4798, opened 2024-05-01
Better X11 window focus handling for scratchpads.

amosbird:
Prior to this commit, the behavior regarding window focus within a scratchpad was unexpected. When a window in the scratchpad was focused, the entire scratchpad group would also be focused, based on the settings of focus_on_window_activation being set to "focus" or "smart".

With this commit, the focus behavior has been refined. Now, only the activated window within the scratchpad will be focused and brought to the front, mirroring the expected toggle behavior.


libqtile/backend/x11/window.py
```diff
@@ -2149,8 +2149,15 @@ def handle_ClientMessage(self, event):  # noqa: N802
                 focus_behavior = self.qtile.config.focus_on_window_activation
                 if focus_behavior == "focus":
                     logger.debug("Focusing window")
-                    self.qtile.current_screen.set_group(self.group)
-                    self.group.focus(self)
+                    # Windows belonging to a scratchpad need to be toggled properly
+                    if isinstance(self.group, ScratchPad):
+                        for dropdown in self.group.dropdowns.values():
+                            if dropdown.window == self:
```
elParaguayo, line 2155:
```suggestion
                            if dropdown.window is self:
```
This would confirm that the window is exactly the same object.

libqtile/backend/x11/window.py
```diff
@@ -2159,8 +2166,15 @@ def handle_ClientMessage(self, event):  # noqa: N802
                         return
                     if self.group.screen == self.qtile.current_screen:
                         logger.debug("Focusing window")
-                        self.qtile.current_screen.set_group(self.group)
-                        self.group.focus(self)
+                        # Windows belonging to a scratchpad need to be toggled properly
+                        if isinstance(self.group, ScratchPad):
+                            for dropdown in self.group.dropdowns.values():
+                                if dropdown.window == self:
```
elParaguayo, line 2172:
```suggestion
                                if dropdown.window is self:
```
As above.

elParaguayo (changes requested):
Thanks for this. 

I've suggested a couple of small changes but the rest looks good.

elParaguayo:
Thanks for making the change. I assume it still works correctly!

elParaguayo (approved):

PR #4799, opened 2024-05-01
Add mute foreground for Volume and PulseVolume

shyguyCreate:
Kind of sure this is how you should change foreground colors.


elParaguayo:
Thanks.

PR #4800, opened 2024-05-01

shyguyCreate:
Follow-up of #3664 

Tests added

### DIFF
--- a/libqtile/lazy.py
+++ b/libqtile/lazy.py
@@ -53,7 +53,7 @@ class LazyCall:
         self._focused: Match | None = None
         self._if_no_focused: bool = False
         self._layouts: set[str] = set()
-        self._when_floating = True
+        self._when_floating: bool | None = None
         self._condition: bool | None = None
         self._func: Callable[[], bool] = lambda: True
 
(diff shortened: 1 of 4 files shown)


libqtile/lazy.py
```diff
@@ -99,7 +99,7 @@ def when(
         focused: Match | None = None,
         if_no_focused: bool = False,
         layout: Iterable[str] | str | None = None,
-        when_floating: bool = True,
+        when_floating: bool | None = None,
```
tych0, line 102: Can you keep the type narrowed to `bool` and just default to False here? or does that change semantics somehow?

shyguyCreate: https://github.com/qtile/qtile/pull/3664#issue-1285547709

> when_floating is None by default
> when set to True, apply lazy call to floating windows only
> when set to False, apply lazy call to tiled windows only

Yes it changes, the current default (when_floating=True) is to match all the windows regardless if it is floating or not.

tych0: I see, ok. Can you add all that to the commit message?

shyguyCreate: Commit message expanded

elParaguayo:
This looks fine to me. Thanks.

PR #4813, opened 2024-05-12
core: use ScreenRect everywhere we represent a screen

tych0:
We had at least three different classes (all dataclasses without the name) that represented a screen, and additionally open coded them with tuple unpacking in a few places as well. Let's consolidate on one class, use dataclass so we get free __init__ and __repr__, and use it everywhere.

For now this still just contains xywh, but it is where we will eventually add the serial number for the output, and could in the future carry information about DPI, etc.

I don't love the name ScreenRect for this, but this might (?) be user facing, and I don't have a better name, so whatever...

This does fail some layout tests, but I'm really not sure why. I'm also mostly just sending it so @elParaguayo can take a look, maybe we can share this as a place to return both serial and dpi info from get_screen_info().

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -45,7 +45,7 @@ from libqtile.command.client import InteractiveCommandClient
 from libqtile.command.interface import IPCCommandServer, QtileCommandInterface
 from libqtile.config import Click, Drag, Key, KeyChord, Match, Mouse, Rule
 from libqtile.config import ScratchPad as ScratchPadConfig
-from libqtile.config import Screen
+from libqtile.config import Screen, ScreenRect
 from libqtile.core.lifecycle import lifecycle
 from libqtile.core.loop import LoopContext, QtileEventLoopPolicy
 from libqtile.core.state import QtileState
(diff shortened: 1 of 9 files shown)


elParaguayo:
Nice. I think this probably makes sense to do.

I'd defer to @jwijenbergh on the wayland stuff.

There aren't that many test failures. I can take a look but it's not obvious why they are failing!

tych0:
I agree, I don't understand why they're failing at all. Seems like maybe an event is getting swallowed, but I don't see where or how :(

elParaguayo:
Test fixes (same issue for both - not sure why the change caused this):
```diff
index 4056883e..1d6716b2 100644
--- a/libqtile/layout/ratiotile.py
+++ b/libqtile/layout/ratiotile.py
@@ -312,9 +312,11 @@ class RatioTile(_SimpleLayoutBase):
         if new_ratio < 0:
             return
         self.ratio = new_ratio
+        self.dirty = True
         self.group.layout_all()

     @expose_command()
     def increase_ratio(self):
         self.ratio += self.ratio_increment
+        self.dirty = True
         self.group.layout_all()
diff --git a/libqtile/layout/spiral.py b/libqtile/layout/spiral.py
index f7122363..eeb85dbd 100644
--- a/libqtile/layout/spiral.py
+++ b/libqtile/layout/spiral.py
@@ -359,6 +359,8 @@ class Spiral(_SimpleLayoutBase):
             return

         setattr(self, prop, value)
+        # Force layout to be recalculated
+        self.dirty = True
         self.group.layout_all()

     @expose_command()
@@ -414,4 +416,6 @@ class Spiral(_SimpleLayoutBase):
         """Reset ratios to values set in config."""
         self.ratio = self.initial_ratio
         self.main_pane_ratio = self.initial_main_pane_ratio
+        # Force layout to be recalculated
+        self.dirty = True
         self.group.layout_all()
```

elParaguayo:
Looking at the `do_not_disturb` test failure. This is happening because the poll function is  being called twice during setup. The second time is triggered because this PR results in the screens being reconfigured. Not sure what's causing that but changing a type shouldn't have that impact.

libqtile/core/manager.py
```diff
@@ -392,9 +394,9 @@ def _process_screens(self, reloading: bool = False) -> None:
 
             # If the screen has changed position and/or size, or is a new screen then make sure that any gaps/bars
             # are reconfigured
-            reconfigure_gaps = ((x, y, w, h) != (scr.x, scr.y, scr.width, scr.height)) or (
-                i + 1 > len(self.screens)
-            )
+            reconfigure_gaps = (
+                (info.x, info.y, info.width, info.height) != (scr.x, scr.y, scr.width, scr.height)
+            ) or (i + 1 > len(self.screens))
```
tych0, line 399: something with this change perhaps? But i agree, it should be a no-op.

tych0:
you are right, it is this https://github.com/qtile/qtile/pull/4813/files#diff-71a17283e4aa06f2d83c5904b5e7f1689c7e4f6437178ead9d3e3c79a65f7280L199-L201

restoring that makes all the tests pass. it's a hack, but I don't really want to unwind it right now, so I will drop it from this PR and file a bug.

tych0:
hmm, so it seems like we maybe just need to remove pypy from the matrix for now, or CI still shows a big red X?

Issue #4815, opened 2024-05-15
CI: do not fail if we can't upload coverage to Coveralls


ramnes:
right?

ramnes:
Rationale if that's not clear enough: Coveralls gives 500 errors right now and I get notifications that our CI is failing because of that.

elParaguayo:
Fine by me.

If there are significant additions to codebase then we should expect tests to be added but we don't need CI to fail

PR #4817, opened 2024-05-18
don't fail on coveralls upload failure

tych0:
I don't know what other consequences this will have if coveralls fails (will it list our coverage as 0%?). I'm also a little weary of this, because if we break it somehow for real, it feels like it'll take a while for someone to notice.

Ideally, there would be some coveralls --fail-only-if-upload-500s. Maybe we should grep output for error strings? That also feels brittle, and it's not clear whether the coveralls error from [1] goes to stdout or stderr, etc. etc.

I don't know. Maybe we shouldn't merge this? coveralls doesn't fail that often.

Fixes: #4815

[1]: https://github.com/qtile/qtile/actions/runs/9091975351/job/24992594703

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ jobs:
     build:
         runs-on: ubuntu-22.04
         name: "python ${{ matrix.python-version }} on ${{ matrix.backend }}"
+        # https://github.com/pypy/pypy/issues/4956
+        continue-on-error: "${{ matrix.python-version == 'pypy-3.10' }}"
         strategy:
             matrix:
                 # If you change one of these, be sure to update:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
               env:
                 BACKEND: ${{ matrix.backend }}
             - name: Upload coverage data to coveralls.io
+              continue-on-error: true
               run: |
                 pip -q install coveralls >=3.3.0
                 coveralls --service=github


jwijenbergh:
Nice! Do you have any idea about the pypy failure? It happens to my wlr 0.17 pr too

elParaguayo:
That will always show the job as passing. I'm not sure we want that.

How about setting `continue-on-error: true` in the step instead?

tych0:
I think it's a pip bug actually, cairocffi >= 1.7.0 should satisfy cairocffi[xcb] >= 1.7.0 IMO. Probably we can just require cairocffi >= 1.7.0 for now, let me see if that will fix it.

tych0:
Hmm. that did not work: https://github.com/tych0/qtile/actions/runs/9141382753/job/25135722849

tych0:
> How about setting continue-on-error: true in the step instead?

oh, i googled for this but didn't find it. I agree that's better. but it still has the same problem, right? we won't notice that the job breaks because it will show as a green check mark in the UI?

tych0:
> Hmm. that did not work

oh, because I didn't change it in setup.cfg. grep fail. but when i change it there, I get a different error...

https://github.com/tych0/qtile/actions/runs/9141421585/job/25135804952

xcffib has no such requirement of a -dev version of cffi. I think this is some version of the pypy-embeds-cffi problem again. I am very tempted to drop pypy support and see if anyone complains, given that we regularly do these kinds of hacks in order to support it and they're annoying.

tych0:
idk, i filed: https://github.com/pypy/pypy/issues/4956 and added it to our skip list with `continue-on-error:` as well similar to what we had in e410b9d0d7c0ed0d3bfcc97cccc478adadad8715

ramnes:
thanks pizza boy

PR #4822, opened 2024-05-20
Wayland: Configure window opacity

jwijenbergh:
This is now possible with wlroots 0.17

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -33,6 +33,7 @@ import wlroots.helper as wlroots_helper
 import wlroots.wlr_types.virtual_keyboard_v1 as vkeyboard
 import wlroots.wlr_types.virtual_pointer_v1 as vpointer
 from pywayland.protocol.wayland import WlSeat
+from pywayland.utils import wl_container_of
 from wlroots import xwayland
 from wlroots.util import log as wlr_log
 from wlroots.util.box import Box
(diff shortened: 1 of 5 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -391,6 +392,31 @@ def finalize(self) -> None:
         if hasattr(self, "qtile"):
             delattr(self, "qtile")
 
+    def win_from_node(self, node) -> Window | None:
```
tych0, line 395: is it always true that a node must have a window?

elParaguayo:
I suspect the docs are failing to build because you need to update the mock modules list.

libqtile/backend/wayland/core.py
jwijenbergh: this code is only called from areas where it is guaranteed we are getting a window from a node. For example when we are iterating through the window tree

jwijenbergh: you make a good point, it can also return internal windows. I changed the signature

tych0: I guess what I'm really wondering is, if we expect this to always return a window, can we drop the `| None` from the signature and just do the assert in this function, instead of making all the callers do it?

jwijenbergh: I think asserting there is fine. I have added this.

PR #4823, opened 2024-05-20
Wayland: Refactor output handling

jwijenbergh:
Output handling was pretty broken still. Custom modes were not working correctly and we were not handling putting output off/on. This patch makes it work well enough in my testing. Unfortunately we have to add a wlroots function in our ffi as this has not been added to pywlroots yet.

The biggest diff in the patch is making sure we use output state everywhere instead of setting wlr_output properties.

Additionally, this also adds adaptive sync.

Note: Most of the logic from this is from DWL

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -46,6 +46,7 @@ from wlroots.wlr_types import (
     GammaControlManagerV1,
     InputInhibitManager,
     OutputLayout,
+    OutputState,
     PointerGesturesV1,
     Presentation,
     PrimarySelectionV1DeviceManager,
@@ -219,7 +220,7 @@ class Core(base.Core, wlrq.HasListeners):
         self.exclusive_client: pywayland.server.Client | None = None
 
         # Set up outputs
-        self.outputs: list[Output] = []
+        self._outputs: list[Output] = []
         self._current_output: Output | None = None
         self.add_listener(self.backend.new_output_event, self._on_new_output)
         self.output_layout = OutputLayout()
@@ -375,6 +376,16 @@ class Core(base.Core, wlrq.HasListeners):
         # Start
         self.backend.start()
 
+    def get_enabled_outputs(self) -> list[Output]:
+        ret = []
+        for output in self._outputs:
+            if not output.wlr_output:
+                continue
+            if not output.wlr_output.enabled:
+                continue
+            ret.append(output)
+        return ret
+
     @property
     def name(self) -> str:
         return "wayland"
@@ -386,7 +397,7 @@ class Core(base.Core, wlrq.HasListeners):
             kb.finalize()
         for pt in self._pointers.copy():
             pt.finalize()
-        for out in self.outputs.copy():
+        for out in self._outputs.copy():
             out.finalize()
 
         if self._xwayland:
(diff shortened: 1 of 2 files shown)


jwijenbergh:
Fine now ^:)

jwijenbergh:
Getting some nonsense test failures that do not fail locally

libqtile/backend/wayland/core.py
```diff
@@ -471,22 +482,39 @@ def _on_new_output(self, _listener: Listener, wlr_output: wlrOutput) -> None:
...
+            head.state.enabled = False
+            self.output_layout.remove(output.wlr_output)
+
+        for output in self._outputs:
+            if not output.wlr_output.enabled:
+                continue
+            # TODO: move this function to pywlroots
+            if (
+                lib.wlr_output_layout_get(self.output_layout._ptr, output.wlr_output._ptr)
+                == ffi.NULL
```
tych0, line 503: man black is ugly sometimes.

tych0 (approved):

libqtile/backend/wayland/core.py
jwijenbergh: Yeah haha

tych0:
sorry, i just merged and made you have to rebase :(

jwijenbergh:
> sorry, i just merged and made you have to rebase :(

it's okay, for some reason CI is having a nonsense fail currently

libqtile/backend/wayland/core.py
```diff
@@ -503,22 +514,39 @@ def _on_new_output(self, _listener: Listener, wlr_output: wlrOutput) -> None:
...
+            head = OutputConfigurationHeadV1.create(config, output.wlr_output)
+            head.state.enabled = False
+            self.output_layout.remove(output.wlr_output)
+
+        for output in self._outputs:
+            if not output.wlr_output.enabled:
+                continue
+            # TODO: move this function to pywlroots
+            if (
+                lib.wlr_output_layout_get(self.output_layout._ptr, output.wlr_output._ptr)
```
tych0, line 534: this seems like it will allocate/increment some kind of refcount? do we need to decrement that refcount to avoid leaks?

jwijenbergh: I don't think this uses refcounting, the output layout is manually destroyed here: https://github.com/qtile/qtile/blob/ca7b28ef46721536d3e5b1ebf59b16d2ac93398b/libqtile/backend/wayland/core.py#L396

jwijenbergh:
draft until weird CI is fixed

libqtile/backend/wayland/core.py
jwijenbergh: I double checked and no we shouldn't free/release this. if you do that all sorts of funny stuff happens with the output layout. I also don't see that malloc or calloc is called in the wlroots source code, it just gives you the pointer to the existing output_layout_output

jwijenbergh:
@tych0 any idea about the ci failures here? I don't see what they have to do with my changes

tych0:
I'm very confused. I don't see the actual recursion in our test output? Does this reproduce for you locally?

jwijenbergh:
> I'm very confused. I don't see the actual recursion in our test output? Does this reproduce for you locally?

I'm also confused, it is is caused by something in this PR but the test pass for me locally. Only seems to happen with tests that use `@dualmonitor`

jwijenbergh:
green! It was recursion in the output layout change wayland event. The code inside there triggered a new event each time it seems. I have fixed it by making better checks whether we need to add the output to the layout or not

PR #4825, opened 2024-05-21
bar: unify finalize() and kill_window()

tych0:
...by just deleting kill_window() and using finalize in the one place it was called.

Their implementations are (I think?) "the same", although kill_window()'s was much more complicated to achieve the same goal, so let's switch to finalize().


tych0:
Even though this didn't fix the bug, I'm going to merge this since it's a nice cleanup.

PR #4829, opened 2024-05-22
document release steps

tych0:
I realized a lot of this is just in my bash history and stuff I've stumbled over. I'm sure other people know this, I'm mostly just writing it down for reference next time I need to tag a release.

I didn't add it to the table of contents or anything since I don't think any users really care, but I can if people think it's useful.

### DIFF
--- a/docs/manual/releasing.rst
+++ b/docs/manual/releasing.rst
@@ -1,0 +1,47 @@
+===============
+Releasing Qtile
+===============
+
+Here is a short document about how to tag Qtile releases.
+
+1. create a "Release vX.XX.XX" commit. I like to thank all the contributors; I
+   find that list with something like:
+
+.. bash::
+
+   git log --oneline --no-merges --format=%an $(git describe --tags --abbrev=0)..HEAD | sort -u
+
+note that this can sometimes generate duplicates if people commit with slightly
+different names; I typically clean it up manually.
+
+Be sure that you GPG-sign (i.e. the ``-S`` argument to ``git commit``) this commit.
+
+2. Create a GPG-signed annotated tag (``git tag -a -s vX.XX.XX``); I usually just use
+   exactly the same commit message as the actual release commit above.
+
+3. Push your tag to qtile/qtile directly:
+
+.. bash::
+
+   git push origin vX.XX.XX
+
+4. Check the `Release action
+   <https://github.com/qtile/qtile/actions/workflows/release.yml>`_; the "Test
+   PyPI upload" action should build and upload the wheels to the test
+   environment. If this step breaks, you can delete your tag (``git push ...
+   :vX.XX.XX``), fix the bug, re-tag, and hopefully that will work.
+
+5. Create a new `Github release
+   <https://github.com/qtile/qtile/releases/new>`_; this is what will trigger
+   the actions in ``.github/workflows/release.yml`` to actually do the real pypi
+   upload.
+
+6. Make sure all of these actions complete as green. The release should show up
+   in a few minutes after completion here: https://pypi.org/project/qtile/
+
+7. send a mail to qtile-dev@googlegroups.com; I sometimes just use
+   git-send-email with the release commit, but a manual copy/paste of the
+   release notes into an e-mail is fine as well. Additionally, drop a message
+   in IRC/Discord.
+
+8. Relax and enjoy a $beverage. Thanks for releasing!


docs/manual/releasing.rst
```diff
@@ -0,0 +1,34 @@
...
+.. bash::
+
+   git log --oneline --no-merges --format=%an $(git describe --tags --abbrev=0)..HEAD | sort -u
+
+note that this can sometimes generate duplicates if people commit with slightly
+different names; I typically clean it up manually.
+
+Be sure that you GPG-sign (i.e. the ``-S`` argument to ``git commit``) this commit.
+
+2. Create a GPG-signed annotated tag (``git commit -a -s``); I usually just use
```
elParaguayo, line 19: Shouldn't this be:
```suggestion
2. Create a GPG-signed annotated tag (``git tag -a -s vX.XX.XX``); I usually just use
```

docs/manual/releasing.rst
```diff
@@ -0,0 +1,34 @@
...
+   exactly the same commit message as the actual release commit above.
+
+3. Push your release commit to git master and your tag.
```
elParaguayo, line 22: Do you need to say you're pushing straight to upstream rather than your fork? Would be good to give command line here too.

docs/manual/releasing.rst
```diff
@@ -0,0 +1,34 @@
...
+
```
elParaguayo, line 23: Before creating the release, we should check the action that builds the wheels and pushes these to test.pypi. This is triggered by the tag being pushed. If that's green then we can create the release.

elParaguayo (commented):
Thanks for this. I've added some suggestions.

docs/manual/releasing.rst
tych0: Oh! I think I knew this, but I totally forgot to do it for the most recent release. I will make a note.

docs/manual/releasing.rst
```diff
@@ -0,0 +1,47 @@
...
+Be sure that you GPG-sign (i.e. the ``-S`` argument to ``git commit``) this commit.
+
+2. Create a GPG-signed annotated tag (``git tag -a -s vX.XX.XX``); I usually just use
+   exactly the same commit message as the actual release commit above.
+
+3. Push your tag to qtile/qtile directly:
+
+.. bash::
+
+   git push origin vX.XX.XX
```
elParaguayo, line 26: For me, `origin` is my fork, `upstream` is the main qtile repo...

docs/manual/releasing.rst
```diff
@@ -0,0 +1,47 @@
...
+
+4. Check the `Release action
+   <https://github.com/qtile/qtile/actions/workflows/release.yml>`_; the "Test
+   PyPI upload" action should build and upload the wheels to the test
+   environment. If this step breaks, you can delete your tag (``git push origin
+   :vX.XX.XX``), fix the bug, re-tag, and hopefully that will work.
+
+5. If the release worked, push to qtile/qtile master directly (``git push
+   origin master``).
```
elParaguayo, line 35: What are you pushing here? I don't think you want any commits being added after the tag. So, once you've fixed the bug, re-tagged the commit, test release builds ok (all step 4) then you go to step 6.

docs/manual/releasing.rst
```diff
@@ -0,0 +1,47 @@
...
+   <https://github.com/qtile/qtile/releases/new>`_; this is what will trigger
+   the actions in ``.github/workflows/release.yml`` to actually do the real pypi
+   upload.
+
+7. Make sure all of these actions complete as green. The release should show up
+   in a few minutes after completion here: https://pypi.org/project/qtile/
+
+8. send a mail to qtile-dev@googlegroups.com; I sometimes just use
+   git-send-email with the release commit, but a manual copy/paste of the
+   release notes into an e-mail is fine as well.
```
elParaguayo, line 47: A message via IRC/Discord would be good too.

docs/manual/releasing.rst
tych0: This is the risk of writing explicit git commands :). Should I drop them? Or what do you prefer?

docs/manual/releasing.rst
tych0: I want the release commits on the git master branch. All the release fixing commits should really go *before* this release commit. Perhaps it's better to delete all of this language, as it's too confusing.

elParaguayo: Yes. I don't think you need step 5.

docs/manual/releasing.rst
elParaguayo: Maybe remove the example and be more explicit in line 22 e.g. adding something like `... as opposed to your fork`

Issue #4225, opened 2023-04-06
[wayland] Enable NumLock on login


arjan-s:
### The issue:

Question, how can I enable NumLock by default when using the Wayland backend? It doesn't seem to be implemented in Qtile.

On X, people usually use an external tool like `numlockx`, but this doesn't work on wayland.

For example, Sway does it through their input settings:
https://wiki.archlinux.org/title/Sway#Initially_enable_CapsLock/NumLock
https://github.com/swaywm/sway/blob/master/sway/commands/input/xkb_numlock.c

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

Sherry65-code:
To enable NumLock on login in qtile on Wayland, you can use the following steps:

- Install the numlockx package by running the following command in your terminal:

  ```bash
  sudo apt install numlockx
  ```

- Add the following line to your qtile configuration file (~/.config/qtile/config.py):

  ```python
  import subprocess
  subprocess.call(["numlockx", "on"])
  ```

  This will turn on the NumLock key when qtile starts up.

- Save the configuration file and restart qtile.

  ```bash
  qtile cmd-obj -o cmd -f restart
  ```

After following these steps, the NumLock key should be enabled automatically when you log in to qtile on Wayland.

arjan-s:
Thanks for taking the time to reply to my question, but sadly it isn't very useful. As I said in my original question, numlockx works on X, not on Wayland.

Sherry65-code:
Enabling **NumLock** by default when using the Wayland backend can be tricky, but there are a few things you can try. 

First, you can check if your Wayland compositor (the program that manages your windows and displays) has a setting to enable NumLock by default. If you're using the Sway compositor, you can add a line to your configuration file to enable NumLock for your keyboard.

If your compositor doesn't have this setting, you can create a custom script that runs when you log in to enable NumLock. This involves creating a file with some commands in it that will run automatically when you log in. 

Finally, if neither of those options work, you can try switching to a different keyboard layout that has NumLock enabled by default. The "us" layout is one example. You can switch to this layout with a command in the terminal.

Keep in mind that these solutions may not work for everyone, so you may need to experiment a bit to find a solution that works for you. 😅

elParaguayo:
Qtile is the Wayland compositor.

[...]

PR #4831, opened 2024-05-23
feat: Add options to set initial state for caps lock and num lock

Sydiepus:
This adds the ability set an initial state for
caps lock and num lock using kb_capslock and
kb_numlock InputConfig options.

Closes:
https://github.com/qtile/qtile/issues/4225

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -24,6 +24,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 from pywayland.protocol.wayland import WlKeyboard
+from wlroots.wlr_types.keyboard import ModifiersMask
 from xkbcommon import xkb
 
 from libqtile import configurable
@@ -143,6 +144,8 @@ class InputConfig(configurable.Configurable):
         ("kb_variant", None, "Keyboard variant i.e. ``XKB_DEFAULT_VARIANT``"),
         ("kb_repeat_rate", 25, "Keyboard key repeats made per second"),
         ("kb_repeat_delay", 600, "Keyboard delay in milliseconds before repeating"),
+        ("kb_numlock", None, "``True`` or ``False`` Enable numlock on startup"),
+        ("kb_capslock", None, "``True`` or ``False`` Enable capslock on startup"),
     ]
 
     def __init__(self, **config: Any) -> None:
@@ -313,6 +316,15 @@ class Keyboard(_Device):
 
         self.seat.keyboard_notify_key(event)
 
+    def _set_num_caps_lock(self, numlock: bool | None, capslock: bool | None) -> None:
+        """Set the initial num/caps lock state of the keyboard."""
+        mask = ModifiersMask(self.keyboard)
+        if numlock:
+            mask.add("Mod2")
+        if capslock:
+            mask.add("Lock")
+        self.keyboard.notify_modifiers(mask)
+
     def configure(self, configs: dict[str, InputConfig]) -> None:
         """Applies ``InputConfig`` rules to this keyboard device."""
         config = self._match_config(configs)
@@ -320,6 +332,7 @@ class Keyboard(_Device):
         if config:
             self.keyboard.set_repeat_info(config.kb_repeat_rate, config.kb_repeat_delay)
             self.set_keymap(config.kb_layout, config.kb_options, config.kb_variant)
+            self._set_num_caps_lock(config.kb_numlock, config.kb_capslock)
 
 
 class Pointer(_Device):
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -144,8 +144,8 @@ class InputConfig(configurable.Configurable):
         ("kb_variant", None, "Keyboard variant i.e. ``XKB_DEFAULT_VARIANT``"),
         ("kb_repeat_rate", 25, "Keyboard key repeats made per second"),
         ("kb_repeat_delay", 600, "Keyboard delay in milliseconds before repeating"),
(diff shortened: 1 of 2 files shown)


libqtile/backend/wayland/inputs.py
```diff
@@ -142,6 +143,8 @@ class InputConfig(configurable.Configurable):
         ("kb_variant", None, "Keyboard variant i.e. ``XKB_DEFAULT_VARIANT``"),
         ("kb_repeat_rate", 25, "Keyboard key repeats made per second"),
         ("kb_repeat_delay", 600, "Keyboard delay in milliseconds before repeating"),
+        ("kb_numlock", None, "``True`` or ``False`` Enable numlock on startup"),
+        ("kb_capslock", None, "``True`` or ``False`` Enable capslock on startup"),
```
tych0, line 147: Can you make the default `False` here instead of `None`?

libqtile/backend/wayland/inputs.py
```diff
@@ -272,13 +275,23 @@ def _on_key(self, _listener: Listener, event: KeyboardKeyEvent) -> None:
 
         self.seat.keyboard_notify_key(event)
 
+    def _set_num_caps_lock(self, numlock: bool | None, capslock: bool | None) -> None:
+        """Set the initial num/caps lock state of the keyboard."""
+        mask = ModifiersMask(self.keyboard)
+        if numlock:
+            mask.add("Mod2")
```
tych0, line 282: is it true that `Mod2` is always numlock? At least on x11, you'd need to query the modmap to see which thing is numlock, and add that one.

libqtile/backend/wayland/inputs.py
Sydiepus: sure.

libqtile/backend/wayland/inputs.py
Sydiepus: I am not sure, maybe if the keys where remapped it will be different.
I don't really understand how xkb work but there's a modifier named `NumLock` does it guarantee that it's always numlock?

libqtile/backend/wayland/inputs.py
jwijenbergh: I would argue that num lock should be true by default

jwijenbergh:
To be complete we should also call wlr_keyboard_led_update.

Sydiepus:
> To be complete we should also call wlr_keyboard_led_update.

the pywlroots implementation of `notify_modifiers` calls `wlr_keyboard_notify_modifiers` which in turn calls `keyboard_led_update` see [here](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_keyboard.c#L96)

jwijenbergh:
> > To be complete we should also call wlr_keyboard_led_update.
> 
> the pywlroots implementation of `notify_modifiers` calls `wlr_keyboard_notify_modifiers` which in turn calls `keyboard_led_update` see [here](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_keyboard.c#L96)

Oh I assumed it was needed because sway did it

libqtile/backend/wayland/inputs.py
Sydiepus: In that case wouldn't it be better to set the numlock to turn on in the default config instead?

jwijenbergh: that is a good idea

libqtile/resources/default_config.py
```diff
@@ -25,6 +25,7 @@
 # SOFTWARE.
 
 from libqtile import bar, layout, qtile, widget
+from libqtile.backend.wayland.inputs import InputConfig
```
Sydiepus, line 28: This would fail if the Wayland dependencies are not present.

elParaguayo: You could include these lines in the default config but comment them out. You can also add some comments to explain what they're for.

@jwijenbergh may have a different idea.

jwijenbergh: You can also do the import in an if backend is Wayland block. But at this point it's better to just make NumLock the default and have a false version commented?

libqtile/backend/wayland/inputs.py
tych0: I don't think it'll work if you put `NumLock` in there, but I honestly don't know. I think you probably have to do a xkb GetMap request to really get this to work right, which will require some changes in xcffib.

Sydiepus:
I was finally able to resolve the issues I was having and I was able to test the changes.
I tried using a rule on an external keyboard and it worked.

arjan-s (approved):
I'm using this PR to automatically enable numlock when logging in. It works perfectly, so I vote for merging it.

arjan-s:
Still using this PR, would still love to see it merged. Please don't close.

tych0:
I still think we can't hardcode Mod2 here. We need something (could be just xmodmap -q, but xcffib now has support for xkb, so that code could be written via actual library calls). I'll see if i can whip up a sample, but in the meantime, `xtrace xmodmap -q` is where people can get started if they are curious.

tych0:
I just released v1.9.0, which has some additional xkb fixes. Here is some sample code for querying the (mod) map: https://github.com/tych0/xcffib/blob/1dfc1da397b5eebedd3a68133c14f856d303e799/test/test_xkb.py#L23-L49

Unfortunately, I haven't had time to sift through the binary format. It should be possible to parse in a similar manner to libxkbcommon, whose source I read to get the above code working.

Once we have that, we can figure out which key is the real mod key dynamically, and I think we can land this.

arjan-s:
I can't remove the `stale` label, but I think this should remain open.

samunemeth:
I would also like to use this in my config, if someone has the time to work on it

PR #4836, opened 2024-05-25
Make default `Plasma` add mode dynamic

nazar-pc:
Backport of https://github.com/numirias/qtile-plasma/pull/26

Basically when window is tall it is split vertically by default and when it is wide it is split horizontally by default. This is in contrast to always splitting windows horizontally by default.


nazar-pc:
Updated previously failing test, conveniently it does test the actual behavior that has changed in this PR

PR #4837, opened 2024-05-26
Add `background` parameter to `Screen`

elParaguayo:
Adds a new `background` parameter which will paint a solid colour the screen.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -65,6 +65,7 @@ if TYPE_CHECKING:
     from libqtile.command.base import ItemT
     from libqtile.confreader import Config
     from libqtile.layout.base import Layout
+    from libqtile.utils import ColorType
 
 
 class Qtile(CommandObject):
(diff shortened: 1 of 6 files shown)


libqtile/core/manager.py
```diff
@@ -450,6 +460,15 @@ def reconfigure_screens(self, *_: list[Any], **__: dict[Any, Any]) -> None:
...
         self.core.painter.paint(screen, image_path, mode)
 
+    def fill_screen(self, screen: Screen, background: ColorType) -> None:
+        with tempfile.NamedTemporaryFile() as temp:
+            surface = cairocffi.ImageSurface(cairocffi.FORMAT_ARGB32, screen.width, screen.height)
+            with cairocffi.Context(surface) as ctx:
+                ctx.set_source_rgba(*rgb(background))
+                ctx.paint()
+            surface.write_to_png(temp.name)
+            self.core.painter.paint(screen, temp.name)
```
tych0, line 470: Can we add a method to get the background surface and do this painting via:

```
ctx.set_source_rgba(0, 0, 0, 0.5)
ctx.rectangle(0, 0, WIDTH, HEIGHT)
ctx.fill()
```
?

it seems a little ugly to me to write a temp file to do this.

elParaguayo: Sure. I didn't want to create a temporary file. I just did this as it allows us to use the `paint` method. The advantage of that was to use the existing code for updating the relevant atoms in x11.

I was being a bit lazy so let me rework this and see if I can refactor those methods.

elParaguayo:
OK - I've refactored this so we're not using a temporary file.

elParaguayo:
Don't merge this yet - I suspect there's a more efficient way to do this on Wayland.

elParaguayo:
@jwijenbergh Can you have a look at the wayland implementation here please.

jwijenbergh (approved):
No comments here for wayland, works perfectly and code is clean

elParaguayo:
Thanks. I'll rebase this into a single commit then.

Issue #4838, opened 2024-05-26
Unmaximising a window using ```lazy.window.toggle_maximize()``` causes window to glitch through its borders (and go to bottom of stack)


theshatterstone:
### Issue description

Issue Description:
When a window is un-maximised (using ```lazy.window.toggle_maximize()``` ), it will either:

- Go to the bottom of the stack, if it wasn't there already
- Glitch through the borders, if it was originally the last window in the stack

- In my config, I have it set so that a maximised window will have no borders and no gaps around it.
- BUT a regular, tiled window will have borders
- So when a window is un-maximised, the above happens
- This only happens with Wayland windows (Foot, Kitty, Floorp (Firefox-based, Wayland backend), Thunar, etc.)
- The issue does NOT occur in the X11 backend
- The issue does NOT apply to XWayland Windows (tested with ``` MOZ_ENABLE_WAYLAND=0 firefox ```)
- I used ```xlsclients``` to confirm that the example windows I tested it with, as described above, were in fact, Wayland windows or XWayland windows respectively
- Also, this issue did not happen before (I was on 0.23.0 before and did not have this issue then)

Relevant Information:
Setup: Qtile Wayland with Qtile-extras (both 0.26.0) on Fedora 40 (AMD CPU & GPU if relevant)

Note: The config should be irrelevant for the most part but I have included it just in case.
If you need any more information or clarifications, please let me know.

### Version

0.26.0

### Backend

Wayland (experimental)

### Config

```python
import os
import re
import socket
import subprocess
from libqtile import qtile
from libqtile.config import Click, Drag, Group, KeyChord, Key, Match, Screen
#from libqtile.command import lazy
from libqtile import layout, bar, hook
from libqtile import widget
#from libqtile.widget import StatusNotifier
from libqtile import config
from libqtile.layout import  base
from libqtile.lazy import lazy
from libqtile.utils import guess_terminal
from typing import List  # noqa: F401
from libqtile.backend.wayland import InputConfig

from qtile_extras import widget
from qtile_extras.widget.decorations import BorderDecoration
from qtile_extras.widget.decorations import PowerLineDecoration

import os
os.environ["XCURSOR_SIZE"] = "24"
os.environ["XDG_CURRENT_DESKTOP"] = "qtile"


home = os.path.expanduser('~') # defining home directory for the hardware configs   

if qtile.core.name == "x11":
    #browser = "vivaldi"
    myTerm = "kitty"
[...]

PR #4839, opened 2024-05-26
Wayland {X,Xdg}Window: Remove all position checks in place

jwijenbergh:
Previously I checked if border needs to change or the window needs to be repositioned, instead of calling configure and setting sizes every time `place()` is called. However, this has too many edge cases. Let's not do premature optimization here because it leads to issues such as #4838


jwijenbergh:
Without this, fullscreening and maximizing is pretty broken. I should have tested that more

PR #4840, opened 2024-05-26
remember to finish all surfaces when finalizing

tych0:
We need to .finalize() all surfaces when we release them, or they may keep resources allocated in their respective backends.

### DIFF
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -75,9 +75,12 @@ class Drawer:
 
     def finalize(self):
         """Destructor/Clean up resources"""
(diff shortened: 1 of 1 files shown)


elParaguayo (approved):
It would nice if the `cairocffi.Surface` class called `finish()` when the object is deleted.

However, this looks fine.

tych0:
> It would nice if the cairocffi.Surface class called finish() when the object is deleted.

Yeah, although I expect they don't want to do that for the same reason we don't: `__del__` isn't guaranteed to be called, so you can leak stuff if you don't explicitly manage lifetimes :(

PR #4841, opened 2024-05-26
remember to finalize widgets

tych0:
when finalizing the bar, we need to explicitly finalize each widget. maybe part of #4821.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -439,6 +439,8 @@ class Bar(Gap, configurable.Configurable, CommandObject):
         if self.window:
             self.window.kill()
             self.window = None
+        for widget in self.widgets:
+            widget.finalize()
         self.widgets.clear()
 
     def _resize(self, length: int, widgets: list[_Widget]) -> None:
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -472,6 +472,10 @@ class Screen(CommandObject):
             self.wallpaper = os.path.expanduser(self.wallpaper)
             self.paint(self.wallpaper, self.wallpaper_mode)
 
+    def finalize(self) -> None:
+        for gap in self.gaps:
+            gap.finalize()
+
     def paint(self, path: str, mode: str | None = None) -> None:
         if self.qtile:
             self.qtile.paint_screen(self, path, mode)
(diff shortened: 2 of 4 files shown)


elParaguayo:
Is this right?

When we do things like `reload_config()`, we call `_finalize_configurables()` which will finalize all widgets. Why do we need to do it again when the bar is finalized?

Also, looping over `self.widgets` won't work if there's a `WidgetBox` in your bar as the widgets in that box don't appear in the bar's widgets list. Adding a `WidgetBox.finalize()` method to finalize the nested widgets might not work either as it would clash with the `_finalize_configurables()` method which loops over `widgets_map` which *does* include the widgets inside the `WidgetBox`.

tych0:
> When we do things like reload_config(), we call _finalize_configurables() which will finalize all widgets. Why do we need to do it again when the bar is finalized?

We should not have two ways to finalize the bar (heh, really three if you consider `kill_window()` that I deleted). IMO, we should change `_finalize_configurables()` to just finalize the screens, instead of open coding all of screen, bar, widget finalization there again.

> Also, looping over self.widgets won't work if there's a WidgetBox in your bar as the widgets in that box don't appear in the bar's widgets list. Adding a WidgetBox.finalize() method to finalize the nested widgets might not work either as it would clash with the _finalize_configurables() method which loops over widgets_map which does include the widgets inside the WidgetBox.

I think we should fix this in the bar's finalize(), then. It sounds like we're probably leaking stuff there too.

tych0:
What do you think of this?

elParaguayo:
> What do you think of this?

I think it looks pretty neat!

tych0:
One thing that's kind of annoying here is that we have to keep this `self.widgets_map.clear()`, viz the test failure, even if we're holding references from the screen as well, which kind of confuses the ownership.

Ideally, the core (qtile) object wouldn't need to know about the widgets at all, and they would only be known about by the Screen that contains them. Unfortunately, we have APIs like `c.widgets["..."]`, which means without this centralized widget map, we would need to walk each Screen every time someone did something like this. Maybe that's not a problem (I guess people are not iterating over widgets in their bar in a tight loop), but it does make cleaning it up and maintaining that API slightly annoying.

tych0:
In any case, I think for our internal idea of ownership, we can stick with Screen -> Bar -> widget for finalization, rather than having it open coded in two places, so IMO this is good to land.

elParaguayo:
Agreed. I'm ok to merge this. Not seeing an obvious fix to the API question but will give it some thought.

PR #4842, opened 2024-05-26
bar: finalize drawer before recreating it

tych0:
If the drawer previously existed, we need to .finalize() it, or we'll leak its resources.

The comment about the stated reason for this is a little strange: we have width and height setters here, can we not just set them and not re-create the drawer at all? Thoughts welcome.

This is another leak that I found while looking for the source of #4821.

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -32,7 +32,7 @@ if typing.TYPE_CHECKING:
     import asyncio
     from typing import Any
 
-    from libqtile.backend.base import Internal, WindowType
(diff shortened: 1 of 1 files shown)


elParaguayo:
If there is a `drawer` already then resizing it should work. The X11 backend finishes the existing `XCBSurface` when dimensions are changed so, I agree, just resetting dimensions may be neater.

tych0:
Ok, I switched it to just set width+height if it already exists. Hopefully it looks reasonable now?

elParaguayo (approved):
Looks good

PR #4845, opened 2024-05-27
Add idle-notify-v1 protocol support

### LINKED ISSUE #4820: Wayland: Implement Idle notify v1 protocol
### Issue description

Is this good enough? Which applications support it? Need to investigate, for now qtile has no idle protocol implementation as the KDE one was deprecated

### Version

git

### Backend

Wayland (experimental)

### Config

_No response_

### Logs

_No response_

### Required

arjanoosting:
As I upgraded to the latest version of qtile on wayland I saw that swayidle did not work any more as the KDE idle protocol support is dropped and the idle notify v1 protocol was not added.

I think I have added the necessary code by reverting the  "(wlroots a289f81) drop KDE idle protocol support"  change and then adjusting the code to use to the IdleNotifierV1 of the idle notify v1 protocol.

At least swayidle will start for me, still need to test whether it will now lock my screen again if I am away from my keyboard :-)

I hope this fixes #4820


arjanoosting:
btw, my screen got locked when I went away from my keyboard for some time :)

tych0:
cc @jwijenbergh do you have any thoughts on this?

jwijenbergh:
Hi, sorry. Will take a look later this week

arjanoosting:
Noticed that er was still in issue with the set_inhited call thus fixed that and did another force push

phragment:
I'm very interested in this, because my use of swayidle with wlr-dpms to lock and then turn off the displays on my main desktop broke in the last qtile release.
I just tested @arjanoosting version d0e6412 (adapted aur/qtile-git to use their branch) and it works for me.
Thank you for your efforts!

jwijenbergh:
Works fine for me and code looks good. Can you add this feature to the changelog? Then ready to merge from my side

arjanoosting:
Added the requested Changelog message

jwijenbergh:
Thanks! Enabled auto merge

PR #4846, opened 2024-05-29
Fix help subcommand

bdabelow:
The help subcommand on the CLI currently throwas an exception because the log_level property is not available on the options objects.

This PR fixes it by adding the parent parser to the help subparser.

Let me know if you feel the test is too much or need an issues first.


tych0:
Thanks for the fix!

PR #4847, opened 2024-05-29
Fix dev requirements

bdabelow:
Fixes two small dependency issues I bumped into while setting up a local dev environment.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,10 @@ repos:
   rev: 6.1.0
   hooks:
   - id: flake8
-    additional_dependencies: ["flake8-pyproject", "flake8-logging-format", "pep8-naming"]
+    # setuptools is required by flake8-logging-format, but not declared
+    # as dependency.
+    # Can be removed when this is fixed: https://github.com/globality-corp/flake8-logging-format/issues/68
+    additional_dependencies: ["flake8-pyproject", "flake8-logging-format", "pep8-naming", "setuptools"]
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v1.6.1
   hooks:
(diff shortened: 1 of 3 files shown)


bdabelow:
Just noticed that the setuptools dependency was already fixed for CI in https://github.com/qtile/qtile/commit/5dd579325718a6a023f2cf7a7aa14e055cc33abe.

Should I revert the separate installation of setuptools in the workflow in this PR as it should be done by pre-commit now?

.pre-commit-config.yaml
```diff
@@ -13,7 +13,9 @@ repos:
   rev: 6.1.0
   hooks:
   - id: flake8
-    additional_dependencies: ["flake8-pyproject", "flake8-logging-format", "pep8-naming"]
+    # setuptools is required by flake8-logging-format, but not declared
+    # as dependency.
```
tych0, line 17: can you link the issue in this comment too, so that we know if we can ever drop the explicit dependency?

tych0:
> Just noticed that the setuptools dependency was already fixed for CI in 5dd5793.

well, "fixed" in CI, but it sounds like you were experiencing it locally? I think it's fine to land these patches (modulo my one nit above).

elParaguayo:
I've hit this locally too. Never got round to pushing the fix though, so thank you.

.pre-commit-config.yaml
bdabelow: Extended the comment.

bdabelow:
> > Just noticed that the setuptools dependency was already fixed for CI in [5dd5793](https://github.com/qtile/qtile/commit/5dd579325718a6a023f2cf7a7aa14e055cc33abe).
> 
> well, "fixed" in CI, but it sounds like you were experiencing it locally? I think it's fine to land these patches (modulo my one nit above).

Yes, happened locally with Python 3.12. I have now also removed the explicit installation of setuptools in the workflow, as the dependecy will take care of that.

PR #4853, opened 2024-05-30
Fix volume clickable event not working

shyguyCreate:
Solves comment https://github.com/qtile/qtile/pull/4779#issuecomment-2140295077

### DIFF
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -202,7 +202,7 @@ class PulseVolume(Volume):
         Volume.__init__(self, **config)
         self.add_defaults(PulseVolume.defaults)
         self.volume = 0
-        self.mute = 0
+        self.is_mute = 0
         self._previous_state = (-1.0, -1)
 
     def _configure(self, qtile, bar):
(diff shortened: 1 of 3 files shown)


shyguyCreate:
Help, how can I make the test pass? I am getting this error message `/bin/sh: 1: amixer: not found` in the checks.

tych0:
the test environment needs amixer, you'll have to add it to the packages list in `.github/workflows/ci.yaml`

shyguyCreate:
Thanks, just wanted to know if added a new dependency was okay.

elParaguayo:
I would assume that the CI environment doesn't have any sound devices so this would probably fail. I'd be inclined to monkeypatch `subprocess.call` and store the command that's called.

shyguyCreate:
Like this `monkeypatch.setattr(vol, "mute", "amixer -q sset Master toggle")`? (first time with monkeypatch)

elParaguayo:
Not quite. I can write a test later for this so don't worry about it for now.

However, I've also noticed that `PulseVolume` has the same issue with the name clashes (that was my fault!) - could you fix that too?

elParaguayo:
Thanks for this. Seems to work fine.

PR #4854, opened 2024-05-30
More cleanup

tych0:
Here are some more leaks I found while strolling through the code for #4821

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -630,6 +630,7 @@ class Painter:
         self.atoms = AtomCache(self)
         self.width = -1
         self.height = -1
+        self.root_pixmap_id = None
 
     def _get_root_pixmap_and_surface(self, screen):
         # Querying the screen dimensions via the xcffib connection does not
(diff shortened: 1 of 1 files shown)


libqtile/backend/x11/xcbq.py
```diff
@@ -702,6 +703,11 @@ def _update_root_pixmap(self, root_pixmap):
         self.conn.core.ClearArea(0, self.default_screen.root.wid, 0, 0, self.width, self.height)
         self.conn.flush()
 
+        # now that we have drawn the new pixmap, free the old one
+        if self.root_pixmap_id is not None and self.root_pixmap_id != root_pixmap:
+            self.conn.core.FreePixmap(self.root_pixmap_id)
+            self.root_pixmap_id = root_pixmap
```
pawamoy, line 709: Didn't try that yet, but it looks like we will never enter this condition, since `self.root_pixmap_id` is given a value only when we enter it, but we need it not to be `None` to enter it. I suppose you just need to dedent the last line assigning `root_pixmap` to it.

tych0: You are totally right. Can you send a patch? Otherwise I can.

PR #4855, opened 2024-05-31
[Wayland] add ability to load xkb keymap files

elParaguayo:
Draft for now. This is based on a request on Discord. 

Would welcome @jwijenbergh thoughts on this.

I think we also need to update the keyboard layout widget for this.

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from pywayland.protocol.wayland import WlKeyboard
@@ -28,6 +29,7 @@ from xkbcommon import xkb
 
 from libqtile import configurable
 from libqtile.backend.wayland.wlrq import HasListeners, buttons
+from libqtile.confreader import ConfigError
 from libqtile.log_utils import logger
 
 try:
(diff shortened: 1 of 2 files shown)


tych0:
would it be useful to have xkb support in xcffib? based on one of the other PRs I looked at adding it recently and I don't think it would be super hard.

elParaguayo:
Yes, would make sense if it's easy to do.

jwijenbergh:
I haven't tested it yet but this looks fine to me, any keymap file I can test with?

libqtile/backend/wayland/inputs.py
```diff
@@ -228,17 +233,33 @@ def finalize(self) -> None:
...
+            keymap = self._keymaps[(layout, options, variant, xkb_file)]
+        elif xkb_file is not None:
+            xkbf = Path(xkb_file).expanduser()
+            if xkbf.is_file():
+                xkbf_name = xkbf.as_posix()
+            else:
+                raise ConfigError(f"kb_file {xkb_file} is not found.")
+
+            with open(xkbf_name, "r") as f:
+                keymap = self.xkb_context.keymap_new_from_file(f)
```
jwijenbergh, line 256: move this to a function maybe?

elParaguayo: If we could use this in x11 then, yes, makes sense.

elParaguayo:
Slight change, `keymap_new_from_file` gave me a `PermissionError` but using `keymap_new_from_buffer` works fine.

I tested by dumping my keymap and then editing it to swap a couple of keys round. Loading the keymap gave the expected behaviour.

tych0:
> Yes, would make sense if it's easy to do.

ok, I spent some time messing around with it today and couldn't get my haskell-fu right, but I'll work on it.

tych0:
Ok, I have finally released xcffib 1.6.0, which has support for xkb (although the xcb-proto xmls are questionable based on my testing, so it probably needs some work). But at least now people can start playing with it for stuff like this if they want.

elParaguayo:
How would we go about implementing this? Not everyone's going to have `xcffib>=1.6.0` - it's not on Arch yet.

tych0:
I just released 1.9.0, which has some additional xkb fixes. There is some sample code here: https://github.com/tych0/xcffib/blob/1dfc1da397b5eebedd3a68133c14f856d303e799/test/test_xkb.py#L23-L49

unfortunately, the response from GetMap() is essentially a binary format, and needs additional parsing once we have it in hand. I was looking at libxkbcommon to try and reverse engineer it, but I haven't gotten that far yet.

As for how to implement it, I think we just increase the xcffib requirement as part of this PR. Maybe that means we hold a release for a while to let the version of xcffib percolate through the distros, but that doesn't seem like a huge deal to me.

PR #4857, opened 2024-06-01
WIP: Wayland z layering

jwijenbergh:
Some TODOs here and there, just want this PRed now so people can test.

Also tests still TODO of course

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -278,6 +278,15 @@ class Core(base.Core, wlrq.HasListeners):
         #     │   │   ├── LayerStatic.tree
         #     │   │   └── ...
         #     │   │
+        #     │   ├── self.keep_below_window_tree
+        #     │   │   ├── XdgWindow.container
+        #     │   │   │   ├── XdgWindow.tree
+        #     │   │   │   └── XdgWindow._borders
+        #     │   │   ├── XWindow.container
+        #     │   │   │   ├── XWindow.tree
+        #     │   │   │   └── XWindow._borders
+        #     │   │   └── ... (further keep below windows)
+        #     │   │
         #     │   ├── self.mid_window_tree
         #     │   │   ├── XdgWindow.container
         #     │   │   │   ├── XdgWindow.tree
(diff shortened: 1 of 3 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -308,11 +353,25 @@ def __init__(self) -> None:
         self.layer_trees = [
             SceneTree.create(self.windows_tree),  # Background
             SceneTree.create(self.windows_tree),  # Bottom
+            SceneTree.create(self.windows_tree),  # Keep below windows
             SceneTree.create(self.windows_tree),  # Regular windows
+            SceneTree.create(
+                self.windows_tree
+            ),  # Keep above windows (floats go here if floats_kept_above
```
jwijenbergh, line 360: holy crap black is ugly

libqtile/backend/wayland/window.py
```diff
@@ -952,7 +1021,7 @@ def place(
         respect_hints: bool = False,
     ) -> None:
         if above:
-            self.bring_to_front()
+            self.move_up()
```
jwijenbergh, line 1024: @elParaguayo is this correct?

libqtile/backend/wayland/window.py
```diff
@@ -686,10 +698,67 @@ def enable_fullscreen(self) -> None:
...
+        self.reparent(self.core.keep_below_window_tree)
+
+    @expose_command()
+    def keep_above(self) -> None:
+        self.reparent(self.core.keep_above_window_tree)
+
+    @expose_command()
+    def move_up(self, force: bool = False) -> None:
+        if force and self.tree_layer is self.core.keep_below_window_tree:
+            self.reparent(self.get_new_layer(self._float_state))
```
jwijenbergh, line 719: @elParaguayo does move_up or down with force also remove it from bring_to_front layer?

libqtile/backend/wayland/xwindow.py
```diff
@@ -478,6 +475,7 @@ def unhide(self) -> None:
                 self.tree.node.set_position(self.borderwidth, self.borderwidth)
 
             self.container.node.set_enabled(enabled=True)
+            # TODO: is this correct
             self.bring_to_front()
```
jwijenbergh, line 479: I have to check all bring to front calls in the wayland code base

libqtile/backend/wayland/window.py
jwijenbergh: I am confused what needs to happen when above is set in `place`

libqtile/backend/wayland/window.py
elParaguayo: x11 doesn't have a `bring_to_front` layer. In x11, `bring_to_front` ignores layers and brings the window to the front. Once it loses focus, it should be restacked according to its layer.

The behaviour of `force` is as follows. If this is `True`, `move_up` will move the window out of the `kept_below` layer group, and `move_down` will move the window out of the `kept_above` layer group. If `False`, the windows can only move within their respective layer groups.

libqtile/backend/wayland/window.py
```diff
@@ -686,10 +698,71 @@ def enable_fullscreen(self) -> None:
...
+        self.container.node.reparent(layer)
+
     @expose_command()
-    def bring_to_front(self) -> None:
+    def keep_below(self) -> None:
+        self.reparent(self.core.keep_below_window_tree)
+
+    @expose_command()
+    def keep_above(self) -> None:
+        self.reparent(self.core.keep_above_window_tree)
```
elParaguayo, line 714: In X11, we have a `kept_above/below` property. We should add that here. 

In addition, the X11 method signature includes an `enable` argument. If this is set, then `kept_above/below` is set to that value. If it's `None` then the state is toggled.

libqtile/backend/wayland/window.py
elParaguayo: I think `move_up` is ok here - but let's see what happens in the tests ;)

elParaguayo (commented):
Have started having a go with this. Some thoughts below.

Also noticed that rofi opened behind floating windows. I then had a coredump...

PR #4861, opened 2024-06-03
x11: ignore initial screen change notifications

### LINKED ISSUE #4818: drop silly hack in x11's `get_screen_info()`
### Issue description

We have the following code in `get_screen_info()`:

```
         if self.qtile:
             self._xpoll()
```

This is a pretty odd place to put this code, and it would be nice to get rid of it. Unfortunately, when we drop it, some tests fail. This is because it has the effect of swallowing the initial ScreenChangeNotify events from the x server, which some of our tests depend on.

### Version

git master

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

tych0:
...and drop our _xpoll() hack. Fixes #4818.

Instead of swallowing these manually by invoking an _xpoll() before the on_screen_change hook is registered, let's just leave that event masked until we are ready to handle it.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -185,10 +185,7 @@ class Core(base.Core):
         self.conn.finalize()
 
     def get_screen_info(self) -> list[ScreenRect]:
-        ps = self.conn.pseudoscreens
-        if self.qtile:
-            self._xpoll()
-        return ps
+        return self.conn.pseudoscreens
 
     @property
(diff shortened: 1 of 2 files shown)


elParaguayo (approved):
Nice. This looks much neater.

tych0:
Yeah, it's a little awkward that we have all these wrappers for individual extensions, but leaving the event masked is definitely the way to go.

PR #4867, opened 2024-06-07
Fix code for freeing old pixmaps

pawamoy:
Asked in https://github.com/qtile/qtile/pull/4854


tych0:
Thank you, good catch.

Issue #4868, opened 2024-06-07
pytest errors: openSUSE Packaging 0.26.0


uncomfyhalomacro:
### Issue description

Running pytest results in errors specifically the migration tests.

### Version

0.26.0

### Backend

X11 (default), Wayland (experimental)

### Logs

```text
[ 1844s] =================================== FAILURES ===================================
[ 1844s] ____________ test_check_all_migrations[ModuleRenames-no-check-test] ____________
[ 1844s] 
[ 1844s] migration_tester = <test.migrate.conftest.MigrationTester object at 0x7fc717d9df50>
[ 1844s] 
[ 1844s]     @pytest.mark.parametrize("migration_tester", migration_tests, indirect=True, ids=migration_ids)
[ 1844s]     def test_check_all_migrations(migration_tester):
[ 1844s] >       migration_tester.assert_check()
[ 1844s] 
[ 1844s] test/migrate/test_check_migrations.py:51: 
[ 1844s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[ 1844s] 
[ 1844s] self = <test.migrate.conftest.MigrationTester object at 0x7fc717d9df50>
[ 1844s] 
[ 1844s]     def assert_check(self):
[ 1844s]         if not self.test.source:
[ 1844s] >           assert False, f"{self.test_id} has no Check test."
[ 1844s] E           AssertionError: ModuleRenames has no Check test.
[ 1844s] E           assert False
[ 1844s] 
[ 1844s] test/migrate/conftest.py:76: AssertionError
[ 1844s] ________________ test_check_all_migrations[RenameUnspecified-0] ________________
[ 1844s] 
[ 1844s] migration_tester = <test.migrate.conftest.MigrationTester object at 0x7fc7192ddb10>
[ 1844s] 
[ 1844s]     @pytest.mark.parametrize("migration_tester", migration_tests, indirect=True, ids=migration_ids)
[ 1844s]     def test_check_all_migrations(migration_tester):
[ 1844s] >       migration_tester.assert_check()
[ 1844s] 
[ 1844s] test/migrate/test_check_migrations.py:51: 
[ 1844s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[...]

PR #4869, opened 2024-06-07
test: do not actually call set_font() in migration

tych0:
We have reports of the stack trace:

[ 1845s] Checking Qtile config at: /tmp/tmpmddmh3ov/config.py [ 1845s] Checking if config is valid python...
[ 1845s] Traceback (most recent call last):
[ 1845s]   File "/home/abuild/rpmbuild/BUILD/qtile-0.26.0/libqtile/scripts/check.py", line 122, in check_config
[ 1845s]     config.load()
[ 1845s]   File "/home/abuild/rpmbuild/BUILD/qtile-0.26.0/libqtile/confreader.py", line 134, in load
[ 1845s]     config = importlib.import_module(name)
[ 1845s]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ 1845s]   File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
[ 1845s]     return _bootstrap._gcd_import(name[level:], package, level)
[ 1845s]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ 1845s]   File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
[ 1845s]   File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
[ 1845s]   File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
[ 1845s]   File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
[ 1845s]   File "<frozen importlib._bootstrap_external>", line 940, in exec_module
[ 1845s]   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
[ 1845s]   File "/tmp/tmpmddmh3ov/config.py", line 10, in <module>
[ 1845s]     tb.set_font(fontsize=12)
[ 1845s]   File "/home/abuild/rpmbuild/BUILD/qtile-0.26.0/libqtile/widget/base.py", line 733, in set_font
[ 1845s]     self.bar.draw()
[ 1845s]     ^^^^^^^^
[ 1845s]   File "/home/abuild/rpmbuild/BUILD/qtile-0.26.0/libqtile/command/base.py", line 281, in __getattr__
[ 1845s]     raise AttributeError(f"{self.__class__} has no attribute {name}")
[ 1845s] AttributeError: <class 'libqtile.widget.textbox.TextBox'> has no attribute bar

the problem here is that the test actually calls set_font() when imported, which we can't do since we haven't actually configured a bar yet.

This is part of #4868 and suggests that we are somehow not running the migration tests correctly in our CI.

### DIFF
--- a/libqtile/scripts/migrations/rename_unspecified.py
+++ b/libqtile/scripts/migrations/rename_unspecified.py
@@ -89,8 +89,9 @@ class RenameUnspecified(_QtileMigrator):
             tb = TextBox(text="hello")
             # just to use ORIENTATION_BOTH and force us to delete only the
             # right thing
-            tb.orientations = ORIENTATION_BOTH
-            tb.set_font(font=UNSPECIFIED, fontsize=12)
+            if False:
+                tb.orientations = ORIENTATION_BOTH
+                tb.set_font(font=UNSPECIFIED, fontsize=12)
             """,
             """
             from libqtile.widget.base import ORIENTATION_BOTH
@@ -100,8 +101,9 @@ class RenameUnspecified(_QtileMigrator):
             tb = TextBox(text="hello")
             # just to use ORIENTATION_BOTH and force us to delete only the
             # right thing
-            tb.orientations = ORIENTATION_BOTH
-            tb.set_font(fontsize=12)
+            if False:
+                tb.orientations = ORIENTATION_BOTH
+                tb.set_font(fontsize=12)
             """,
         )
     ]
--- a/tox.ini
+++ b/tox.ini
@@ -29,6 +29,7 @@ testdeps =
     libcst >= 1.0.0
     PyGObject   
     isort
+    mypy
 commands =
     !x11: pip install pywlroots==0.17.0
     pip install .


tych0:
This failure is expected, it is the same as the unfixed one in #4868 when we actually run the migration test in CI. I played around with fixes and I couldn't figure out anything. I'll leave this here for now, and will try to pick it up later unless someone else gets to it first.

uncomfyhalomacro:
is this error unique to just mypy? it seems that adding mypy causes the second commit to fail?

uncomfyhalomacro:
I found some issues with mypy from years ago and it still plagues other packages when mypy is done on the CI
- https://github.com/python/mypy/issues/10428
And an issue that (although misdiagnosed) seems to describe what mypy does https://github.com/python/mypy/issues/1293#issuecomment-208962575

but again i am no expert in python. i just found some references that may or may not be related to the current situation here hmm

elParaguayo:
So... we used to have a separate `mypy` environment in our `tox.ini` file. We only ran this for the latest version of mypy. The reasoning was that the `mypy` checks are very slow and it was unnecessary to run in every environment. In addition, packagers wanted to separate type checking tests from functionality tests as they didn't care about the type hinting.

Tests requiring `mypy` should be skipped if it's not installed. I wouldn't have it as a main test dependency but we should re-enable a `mypy` environment.

uncomfyhalomacro:
I confirmed that builds are successful removing mypy but i am not sure how I can help in this PR. because it looks like a weird cat and mouse chase of what's causing the builds to fail if mypy was added 🥲

--
Sincerely,
Soc Virnyl Estela

tych0:
> So... we used to have a separate mypy environment in our tox.ini file. We only ran this for the latest version of mypy. 

Do you know why we deleted it?

elParaguayo:
No idea but #4782 is an attempt to restore this. There are a number of errors that need to be addressed but they're wayland related.

tych0:
> No idea but https://github.com/qtile/qtile/pull/4782 is an attempt to restore this. There are a number of errors that need to be addressed but they're wayland related.

Hm? I don't see that restoring mypy (but there are indeed tests that fail, one of them I fixed here, the other one I need to do more research on how to fix exactly).

elParaguayo:
That PR adds a new CI workflow for type checking (so it can run in parallel to the various tests). There's a new environment in `tox.ini` which will run `mypy` on the entire codebase. If that passes (which it curently doesn't - that's another issue that needs to be fixed) it will run the tests that require `mypy`.

tych0:
Oh! Yes. I can push my fix there and close this one if you like? But we'll still have one more to land.

elParaguayo:
Sure. Happy for you to add to that PR.

Issue #4865, opened 2024-06-06
CPU load increased to 95 % since commit b2220d27cf6e8858c2e83dad9d0c46f6c83c2349


arjanoosting:
### Issue description

Today I tried out git master to see the fix for #4811 and noticed that laptop fan continued to make noise.
I discovered some error messages in the log (#4864) and high cpu load with qtile continuously using 95% of an CPU in the htop output.

I git bisected this issue to commit b2220d27cf6e8858c2e83dad9d0c46f6c83c2349 which mentions removing an premature optimization. It seems the premature optimization was needed on my system

> git bisect bad
> b2220d27cf6e8858c2e83dad9d0c46f6c83c2349 is the first bad commit
> commit b2220d27cf6e8858c2e83dad9d0c46f6c83c2349
> Author: Jeroen Wijenbergh <qtile@jwijenbergh.com>
> Date:   Sun May 26 13:56:34 2024 +0200
> 
>     Wayland {X,Xdg}Window: Remove all position checks in place
>     
>     Previously I checked if border needs to change or the window needs to be repositioned,
>     instead of calling configure and setting sizes every time `place()` is called. However,
>     this has too many edge cases. Let's not do premature optimization here because it leads to issues
>     such as #4838
> 
>  CHANGELOG                             |  3 ++-
>  libqtile/backend/wayland/xdgwindow.py | 34 ++++++----------------------------
>  libqtile/backend/wayland/xwindow.py   | 22 ++--------------------
>  3 files changed, 10 insertions(+), 49 deletions(-)
>  

### Version

0.26.1.dev4+gb2220d27

### Backend

Wayland (experimental)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

arjanoosting:
After reverting that commit the CPU usage dropped and fluctuates between 3.9 and 25%

jwijenbergh:
Interesting as we never did this optimization in 0.25.0. can you comment the clipping inside of the place methods?

arjanoosting:
Hi, 

I just tried with commented clipping calls, but the CPU load is still high:

> git diff
> diff --git a/libqtile/backend/wayland/xdgwindow.py b/libqtile/backend/wayland/xdgwindow.py
> index 9364f24f..1acc0537 100644
> --- a/libqtile/backend/wayland/xdgwindow.py
[...]

Issue #4878, opened 2024-06-12
Wev  crashes "Error marshalling request: Too many open files"


bendi-github:
### Issue description

When running wev to debug input events, the log gets filled up with 
```
[09:     xdg_surface] configure: serial: 32353
[10:    xdg_toplevel] configure: width: 848; height: 330
                      activated tiled-left tiled-right tiled-top tiled-bottom
```
and eventually crashes
```
error marshalling arguments for create_pool: dup failed: Too many open files
Error marshalling request: Too many open files
```
Input events are still logged while the log is being spammed.

### Version

0.23.0b3.dev350+g6ee02e5c

### Backend

Wayland (experimental)

### Config

```python
Default, I just changed the modkey to `mod1` for testing
```

### Logs

Nothing shows up, besides the usual warnings the default config throws on wayland

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

jwijenbergh:
`0.23.0b3.dev350+g6ee02e5c`

that is a really old version, please test on the latest

bendi-github:
```bash
git clone https://github.com/qtile/qtile.git
cd qtile
pip install .
pip install --config-setting backend=wayland . 
```
`qtile-0.26.1.dev25+g29695c71`

The issue persists. I tested a nested qtile instance in additon to starting it directly via sddm. I noticed when I launched wev in a nested session, it crashed faster when compared to a non-nested session.

tych0:
I wonder if this is really a dupe of #4821, that we're leaking stuff when certain events happen, and in the wev model these things end up as file descriptors?

jwijenbergh:
#4880 fixes this for me

jwijenbergh:
> I wonder if this is really a dupe of #4821, that we're leaking stuff when certain events happen, and in the wev model these things end up as file descriptors?

I doubt it, this seems like a recent regression in git

PR #4880, opened 2024-06-15
Fix wayland cpu usage

jwijenbergh:
This patch reverts an optimization that was removed, so in english: it adds back an optimization :^)

Previously this optimization had issues with fullscreen/max window toggle but this patch fixes that by checking the toplevel width/height.

This optimization is needed in the wlr 0.17 codebase because we `place()` on each commit. Which is what other compositors also do and they also check for geometry changes

Fixes: #4865 #4878

### DIFF
--- a/libqtile/backend/wayland/xdgwindow.py
+++ b/libqtile/backend/wayland/xdgwindow.py
@@ -255,8 +255,8 @@ class XdgWindow(Window[XdgSurface]):
             width -= margin[1] + margin[3]
             height -= margin[0] + margin[2]
 
+        state = self.surface.toplevel._ptr.current
         if respect_hints:
-            state = self.surface.toplevel._ptr.current
             width = max(width, state.min_width)
             height = max(height, state.min_height)
             if state.max_width:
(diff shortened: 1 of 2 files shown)


jwijenbergh:
@elParaguayo can you daily drive this for a bit?

elParaguayo:
Sure. Happy to give it a go.

tych0 (approved):

libqtile/backend/wayland/xdgwindow.py
```diff
@@ -287,6 +284,17 @@ def place(
...
         )
+        state = self.surface.toplevel._ptr.current
+        place_changed = any(
+            [
+                self.x != x,
+                self.y != y,
+                self._width != width,
+                self._height != height,
+                state.width != width,
+                state.height != height,
```
elParaguayo, line 295: Why do we need to check `width/height` and `_width/_height`?

jwijenbergh: the state's width and height is the width and height what the window reports using wlroots APIs, whilst the self._width and self._height is the width and height of the window. If any of those are mismatched we need to resize the window. the check against `self._width` and `self._height` is potentially unneeded but maybe there is some edge case to consider

elParaguayo:
This has been running fine for me.

libqtile/backend/wayland/xdgwindow.py
elParaguayo: OK. It's just not clear to me how they can be different. Doesn't `self.width` just retrieve `self._width`?

jwijenbergh: It says `state.width` not `self.width`

elParaguayo: Ha ha. Wow. I obviously need to get my eyes tested!

jwijenbergh: Haha it's fine, thanks for the review

PR #4884, opened 2024-06-15
Implemented Layout.Bsp.rotate()

vhalragnarok-ffxiv:
Per a user request on Discord, I implemented rotate()

Logic is simple: flip_up(), flip_down(), and flip_right() until the window moves, emulating a 90 degree Clockwise rotation.


tych0:
hi - any chance you can write a test or two for this logic so we don't break it?

vhalragnarok-ffxiv:
I am going to close this as the more I think on it, this is not A: Good Code and B: Not really testable code.

PR #4887, opened 2024-06-17
Fix inconsistent battery level in notification from battery widget

### LINKED ISSUE #4876: Battery widget, "wrong" value on low battery notification
### Issue description

For the Battery widget, the value shown in the low level notification differs (by 1%) from the one shown in the widget itself.
This is not a bug per se, but something that is slightly annoying and should be fairly easy to solve.

After looking at the code, I think this is caused by the way the percentage is printed on the notification and on the widget. i.e. `percent = int(status.percent * 100)` is what goes to the notification, while something like `'{percent:2.0%}'.format(percent=status.percent)` is what goes to the widget.
One method always rounds down while the other rounds up and down.

I think this could be solved by changing one line in the poll function of the Battery class:
```
percent = int(status.percent * 100)
if percent < self.notify_below:
  if not self._has_notified:
    send_notification(
      "Warning",
      "Battery at {0}%".format(percent), <-- this line
      urgent=True,
      timeout=self.timeout,
    )
```
should change to:
```
"Battery at {0:2.0%}".format(status.percent),
```

### Version

0.26.0

### Backend

Wayland (experimental)

### Config

```python
Not relevant, imo
```

### Logs

```text
None
```

### Required

omulh:
This should hopefully fix #4876

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -515,7 +515,7 @@ class Battery(base.ThreadPoolText):
                 if not self._has_notified:
                     send_notification(
                         "Warning",
-                        "Battery at {0}%".format(percent),
+                        "Battery at {0:2.0%}".format(status.percent),
                         urgent=True,
                         timeout=self.timeout,
                     )


tych0:
Thanks!

PR #4892, opened 2024-06-20
Multiple batteries support

fjpavon:
The widget can manage all the batteries in the computer: BAT0, BAT1,... and calculate the global information: % charge, time...
It's available for FreeBSD and Linux
By default all the batteries are checked. I modified the battery number default value to -1 because all the batteries are numbered with values equal or grater than 0 and if you want to get all the batteries you have to use negative numbers.
I changed the test_battery because i use the battery number to get the path to the info files.

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -159,12 +159,12 @@ class Img:
         self.path = path
 
     def _reset(self):
-        attrs = ("surface", "pattern")
-        for attr in attrs:
-            try:
-                delattr(self, attr)
-            except AttributeError:
-                pass
+        if hasattr(self, "surface"):
+            self.surface.finish()
+            del self.surface
+        if hasattr(self, "pattern"):
+            # patterns do not need to be finish()ed, only surfaces do
+            del self.pattern
 
     @classmethod
     def from_path(cls, image_path):


tych0:
Hi - thanks for the patches. Can you rebase these onto current git master so all the double-counted commits go away?

jwijenbergh:
Hmmm is this really needed? I have two batteries and just use two battery widgets

Sniki:
> Hmmm is this really needed? I have two batteries and just use two battery widgets

I would really like this to be merged please.

This was a request of mine months ago, i have many thinkpads with two batteries, yes i have been using two widgets for that but it makes space problems on the bar and it’s a bit confusing with two to know the exact eta for time left on battery, with this, i can get a single widget that shows me the time left from both batteries and have one extra space for another widget that i could really use.

- extra space for another widget
- Proper eta of time left on battery
- A more tidy bar

Thanks !

fjpavon:
> Hi - thanks for the patches. Can you rebase these onto current git master so all the double-counted commits go away?

Hi Tych0
Sorry for the late reply. Sorry again, but it's my first time contributing into a project and i don't known to much about git. I use it only for get my configuration files and password-store saved in a "safe" place, you know...git clone, push and pull. 
I'm reading a git book trying to get a deeper knowledge about and i'm not sure if i did it's right or not, so excuse me if i make mistakes.
I run these commandsgit rebase
**git checkout master**
**git rebase**
**git rebase multiple_batteries**
multiple_batteries is my dev branch
Sorry but i don't know if it's what you expect to be done or if i need to run any other command.

Thanks and kind regards

tych0:
Right now your branch has multiple commits that already exist on master. I don't know how you got into this state, but I think you can fix it via the following:

```
git checkout multiple_batteries
git checkout -b multiple_batteries-pending
git checkout multiple_batteries
git reset --hard origin/master
git cherry-pick <the sha of your git patch from multiple_batteries>
```

What we need to land this is a clean git history on the branch.

Can you also address @jwijenbergh's comment? In particular, can this be addressed with adding some margin tuning vs. having this for loop over batteries?

fjpavon:
Hi tych0
I run the git commands as you suggested and i hope that everything is fixed now.
The widget add the option to show the overall info in one place. I changed the default value because is the way that it works in others projects: kde, gnome, zsh... but If you want 2 widgets as @jwijenbergh you can get them or if you want only one as @Sniki and me you can get it too.
In this image I have two widgets, in the left side showing the overall status and the right one showing the first battery, but you can configure it as you wish.
image
```
widget.Battery(battery=-1),
powerline('color2', 'color3'),
widget.Battery(battery=0),
```

I loop over the batteries but i only get the info of the selected battery. In the LinuxBattery class
```
for bat in bats:
    if int(batnum) < 0 or self.battery == bat:
```
In the FreeBSD class:

```
for bat in range(numbat):
    if bat == int(self.battery) or int(self.battery) < 0:

```

elParaguayo:
I have some comments on this too and will post them when I get a chance.

libqtile/widget/battery.py
```diff
@@ -460,7 +536,11 @@ class Battery(base.ThreadPoolText):
         ("low_foreground", "FF0000", "Font color on low battery"),
         ("low_background", None, "Background color on low battery"),
         ("update_interval", 60, "Seconds between status updates"),
-        ("battery", 0, "Which battery should be monitored (battery number or name)"),
+        (
+            "battery",
```
elParaguayo, line 540: If we're changing the default behaviour, can you add something to the changelog to make this very clear.

libqtile/widget/battery.py
```diff
@@ -344,57 +362,115 @@ def set_battery_charge_thresholds(self, start, end):
...
+                elif power_unit == "uW":
+                    power = power / 1e6
+
+                power_acc += power
 
         return BatteryStatus(
             state=state,
-            percent=percent,
-            power=power,
+            percent=percent / len(bats),
```
elParaguayo, line 472: This assumes batteries have equal capacity. The percentage should be calculated as the sum of `now` over the sum of `full`.

elParaguayo (changes requested):
Sorry for the delay. A couple of comments here.

I can't comment on appropriateness of FreeBSD changes.

fjpavon:
Be patient. I have to install a linux distribution in my laptop and i have to be sure that when i come back to freebsd i have all my conf files...;)

fjpavon:
Hi All,
After reinstall the laptop I have changed the percentage calculation and the result is the same as in gnome. So if i'm doing wrong the gnome guys too. I'm continue running the code waiting some difference in the numbers so i'll commit the modification tomorrow. Meanwhile i'm enjoy with the debian experience.

libqtile/widget/battery.py
fjpavon: Hi, Sorry for the very late reply
I fixed the formula in linux and freebsd. The value of **now** in freebsd can not be got so easy as in linux, the only way to do it was usingpercentage and the full value of each battery. The results are consistent using other ways to check the total percentage, gnome battery widget in linux or apm command in freebsd

libqtile/widget/battery.py
fjpavon: Sorry for the noob question beacuse its my first contribution to an open source project, i don't use git,... . Do i have to modify the changelog file in the root of the repository?

jwijenbergh: Yes, that's the one!

elParaguayo:
Did you mean to close this?

fjpavon:
Hi, 
I thought that i was resolving your comments....sorry but works with git and all this stuff is very difficult to me. 

Thanks and kind regards.

Sniki:
What is left for this PR to be ready for merge ?
As long as the accuracy of battery reporting is somewhat close, it should be fine. 

CC @fjpavon

fjpavon:
> What is left for this PR to be ready for merge ? As long as the accuracy of battery reporting is somewhat close, it should be fine.
> 
> CC @fjpavon

Sorry i was focused in other stuff and i forgot that it was not finisthed.
During the weekend i was working trying to fix an error in the test_battery and i disconvered that it's not working on FreeBSD...but it's resolved, now the problem is the rufus-format .
I think that the os check in the tests can be the next issue to resolve.

elParaguayo:
Why did you close this?

fjpavon:
> Why did you close this?

Because the unknowledge is the worst weapon. I tried to fix some merge issues but i forget to request help before to run any kind of git command

PR #4896, opened 2024-06-23
allow dgroups to persist

phragment:
- allow specifying if added group should be persistent

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -578,9 +578,12 @@ class Qtile(CommandObject):
         label: str | None = None,
         index: int | None = None,
         screen_affinity: int | None = None,
+        persist: bool | None = False,
     ) -> bool:
         if name not in self.groups_map.keys():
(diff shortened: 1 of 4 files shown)


phragment:
Thank you for considering this, I'm open to improvements (especially for the test cases).

tych0:
Thanks!

elParaguayo:
@phragment This is throwing up a minor bug - see #4933

It's weird because `client.group` seems to be `None` but the `_del` method can only be called if the client group is not None.

phragment:
@elParaguayo I will take a look, maybe this is an x11/wayland issue.

elParaguayo:
@phragment note, there's some discussion in #4933

phragment:
I went through the mentioned issues discussions, it seems only to happen on x11 with systray (xembed windows).
At last that explains why I completely missed this potential issue. (I'm daily driving this qtile setup, but with the wlroots backend.)
I'm relieved to see the issue solved with https://github.com/qtile/qtile/pull/4941

PR #4898, opened 2024-06-24
Wayland Inputs: Allow keybinds to repeat when holding keys

jwijenbergh:
This is a patch to enable keybinds to repeat according to the keyboard's repeat rate and delay. It does this by using asyncio timer's. Other compositors use wayland timers, but using asyncio timers is more qtile-y and seems to work well


elParaguayo (approved):
Repeat works nicely for me.

jwijenbergh:
Thanks for testing!

PR #4905, opened 2024-07-01
Fix playback status text color issue in Cmus widget

### LINKED ISSUE #4904: TextBox widget ignores layout changes if text didn't change
### Issue description

Due to [this](https://github.com/qtile/qtile/blob/master/libqtile/widget/base.py#L751) code in the TextBox widget's update function

```python
        if self.text == text:
            return
```

layout changes (like text color) are not applied unless the text changes.

See the discussion in #4905.

### Version

0.26.1.dev0+ge4b4bb8b.d20240523

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

saviola777:
This fixes an issue in the Cmus widget where the text color would not change to reflect the current playback status unless the song was changed (#4904).

The reason for the issue is that the TextBox widget only redraws the widget if the text changes without checking if layout information like text color have changed.

We use the `_force_redraw` flag of the `TextWidget` to force a redraw if the playback status changes to fix this.

Closes #4904.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -509,6 +509,7 @@ class _TextBox(_Widget):
         self._scroll_queued = False
         self._scroll_timer = None
         self._scroll_width = width
+        self._force_redraw = False
 
     @property
     def text(self):
@@ -748,8 +749,10 @@ class _TextBox(_Widget):
         if not self.can_draw():
             return
 
-        if self.text == text:
+        if self.text == text and not self._force_redraw:
             return
+        self._force_redraw = False
+
         if text is None:
             text = ""
 
--- a/libqtile/widget/cmus.py
+++ b/libqtile/widget/cmus.py
@@ -100,6 +100,7 @@ class Cmus(base.ThreadPoolText):
         if info:
             status = info["status"]
             if self.status != status:
+                self._force_redraw = True
                 self.status = status
                 if self.status == "playing":
                     self.layout.colour = self.play_color


tych0:
Should we be doing this in TextBox itself? Seems like something all widgets would need?

elParaguayo:
I thought about that too. I think the approach here is correct.

The default TextBox behaviour is to only redraw if the text value changes. The issue here is that the text isn't changing but another variable is and that means we need to redraw. However, because we're using a polling widget, there's no other place to trigger a redraw other than the way it's done here.

elParaguayo:
The alternative is to change how `ThreadPoolText` works. Have some variable like `self._force_redraw` and have that checked here: https://github.com/qtile/qtile/blob/dfb7186eed9253fff9958877bb3fe1e5e0ffcf32/libqtile/widget/base.py#L850-L852

saviola777:
Sounds like a cleaner solution to me. But why not just pull that logic into the `TextBox` widget, i.e. change

https://github.com/qtile/qtile/blob/dfb7186eed9253fff9958877bb3fe1e5e0ffcf32/libqtile/widget/base.py#L751-L752

to 

```python
if self.text == text and not self._force_redraw:
    return
# […]
self._force_redraw = False
```

What is the usual procedure now, do I close this PR and open a new one with the other approach, or do I force-push the new approach into this PR?

elParaguayo:
> Sounds like a cleaner solution to me. But why not just pull that logic into the `TextBox` widget, i.e. change
> 
> https://github.com/qtile/qtile/blob/dfb7186eed9253fff9958877bb3fe1e5e0ffcf32/libqtile/widget/base.py#L751-L752
> 
> to
> 
> ```python
> if self.text == text and not self._force_redraw:
>     return
> # […]
> self._force_redraw = False
> ```

Good question. I had assumed this was more relevant to polling widgets but your suggestion seems fine too.

> What is the usual procedure now, do I close this PR and open a new one with the other approach, or do I force-push the new approach into this PR?

Force pushing here is fine.

saviola777:
I've updated the approach as discussed. Let me know if/where I should add tests/documentation. I just noticed that I forgot to create a feature branch on my fork, sorry about that and thanks for not yelling at me. 🙈

tych0:
I wonder, since we have `backend/base/drawer.py:TextLayout` that's a wrapper here, can we make it so that it keeps track of its dirty status? That way we don't have to annotate every place in the code where we want to force a refresh because the colors change. Maybe via https://pycairo.readthedocs.io/en/latest/reference/context.html#cairo.Context.get_source ?

saviola777:
You mean automatically keeping track of changes to colors etc.? That does sound like the best solution, but implementing that currently exceeds my knowledge of the codebase.

I suggest you create a new issue for this, and I close this PR so someone more experienced can try their hand at the proper solution?

elParaguayo:
I'm not sure that checking the context is the right approach. Couldn't we use the property setters in the text layout class to set a dirty flag if they're changed?

tych0:
> I'm not sure that checking the context is the right approach. Couldn't we use the property setters in the text layout class to set a dirty flag if they're changed?

we can, but I think our use of property setters is kind of confusing, since it is hard to know what assignments will have side effects beyond just setting the variable and what won't. It's not a strongly held belief, though, I'd be fine if we went that route.

saviola777:
I will close this PR and re-open a new one (from a feature branch this time!) once I've implemented the approach with the property setters. I have to admit that I don't quite understand tych0's approach with the context, still haven't delved deep enough into the codebase for that I guess.

elParaguayo:
I could be missing something too but I don't think the context works as it will only change when we write to the drawer object but we're only going to do that if we need to.

PR #4912, opened 2024-07-04
Fix volume widget ignoring default foreground

saviola777:
This fixes an issue in the volume widget where the foreground color in widget_defaults would be ignored because unmute_foreground is initialized in `__init__()`.

This is fixed by moving the initialization of unmute_foreground into `_configure()`.

Let me know if you want me to add a test. (Also, I hope it is fine to create PRs without related issues if the problem and solution are properly explained? If not I can create an issue of course.)

### DIFF
--- a/libqtile/widget/volume.py
+++ b/libqtile/widget/volume.py
@@ -110,7 +110,6 @@ class Volume(base._TextBox):
         self.surfaces = {}
         self.volume = None
         self.is_mute = False
-        self.unmute_foreground = self.foreground
 
         self.add_callbacks(
             {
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks. No need for a test. I'm familiar with the issue and this is the appropriate solution.

saviola777:
Oops, I totally did not submit this PR without running the tests, nobody would be that stupid!

… I'll see about getting the tests to pass, sorry. :smile:

saviola777:
It seems like `_configure()` is not called in the tests, so `unmute_foreground` is not defined in `_update_drawer()`.

What's the "right" way to improve these tests (assuming that duplicating the initialization of `unmute_foreground` in `__init()` is not the right way to fix this test failure)?

elParaguayo:
There are a number of tests that call `_configure` with a `fake_qtile` fixture and a `FakeBar` instance. Have a look at `test_clock.py` for example.

saviola777:
Updated the test, is this what you had in mind?

elParaguayo:
> Updated the test, is this what you had in mind?

Yes!

Issue #4907, opened 2024-07-02
check_update widget uses a FreeBSD deprecated command 


fjpavon:
### Issue description

The pkg_version command is deprecated since FreeBSD 10.0.
In the [Release Notes](https://www.freebsd.org/releases/10.0R/relnotes/): _The pkg_add, pkg_create, pkg_delete, pkg_info, pkg_updating, and pkg_version utilities have been removed. [pkg(7)](http://www.freebsd.org/cgi/man.cgi?query=pkg&sektion=7) must now be used to install binary packages. [pkg(7)](http://www.freebsd.org/cgi/man.cgi?query=pkg&sektion=7) is the next generation FreeBSD package manager
To check the updates you have to run the command _pkg upgrade_

### Version

all versions

### Backend

X11 (default), Wayland (experimental)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

fjpavon:
Hi tych0
I want to create a pr with the new command, but i think that i have to fix my fork before. Anyway this is the output of the command in my laptop:
```
 fjp@claire  ~  pkg upgrade --dry-run                                                                                                                                                                          80%
Checking for upgrades (74 candidates): 100%
Processing candidates (74 candidates): 100%
The following 19 package(s) will be affected (of 0 checked):

Installed packages to be UPGRADED:
	aom: 3.8.2 -> 3.8.3
	emacs: 29.3,3 -> 29.3_3,3
	firefox: 127.0_1,2 -> 127.0.2,2
	glib: 2.80.2,2 -> 2.80.3,2
	qt5-assistant: 5.15.13p4 -> 5.15.14p4
	qt5-core: 5.15.13p142 -> 5.15.14p140
	qt5-dbus: 5.15.13p142 -> 5.15.14p140
	qt5-gui: 5.15.13p142 -> 5.15.14p140
	qt5-help: 5.15.13p4 -> 5.15.14p4
	qt5-linguist: 5.15.13p4 -> 5.15.14p4
	qt5-linguisttools: 5.15.13p4 -> 5.15.14p4
	qt5-network: 5.15.13p142 -> 5.15.14p140
	qt5-printsupport: 5.15.13p142 -> 5.15.14p140
	qt5-sql: 5.15.13p142 -> 5.15.14p140
[...]

PR #4914, opened 2024-07-07
check_update widget uses a FreeBSD deprecated command

fjpavon:
To fix the issue #4907 the check command it's changed
the widget use the new command and the output is filtered using awk to get a line for each packates. This lines begin with a tab

### DIFF
--- a/libqtile/widget/check_updates.py
+++ b/libqtile/widget/check_updates.py
@@ -37,7 +37,7 @@ CMD_DICT = {
     "Guix": ("guix upgrade --dry-run", 0),
     "Ubuntu": ("aptitude search ~U", 0),
     "Fedora": ("dnf list updates -q", 1),
-    "FreeBSD": ("pkg_version -I -l '<'", 0),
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks.

Will this break anything for users on an earlier version of FreeBSD or has this command been available for a while?

fjpavon:
Yes, we are changing the command used to check the updates but the previous command was deprecated in FreeBSD 10.0 and it was published in January 20, 2014 [FreeBSD Releases](https://www.freebsd.org/releases/).

tych0:
These xkbcommon build failures are not your fault, I will try to take a look soon. Looks like maybe a new constant, and our CI version of libxkb is too old.

elParaguayo:
Need to pin xkbcommon version in CI.

https://github.com/sde1000/python-xkbcommon/issues/23

tych0:
It looks like we can also move to 24.04 and make it work? I am trying that here: https://github.com/tych0/qtile/tree/ci-ubuntu-2404

elParaguayo:
That would be better!

PR #4916, opened 2024-07-08
Improve and extend Cmus widget

saviola777:
This improves and extends the Cmus widget by

- replacing "play_icon" with "playing_text", "paused_text", and "stopped_text" which can be included depending on the current playback status using the "status_text" format variable,
- making "position", "duration", and "remaining" track information available in "[h:]mm:ss" format in the format string (as well as "position_percent" and "remaining_percent"),
- adding information about all available format variables to the docstring,
- adding two additional format strings for streams and when no artist is available,
- replacing "play_color" and "noplay_color" with "playing_color", "paused_color", "stopped_color", and
- streamlining tests by replacing FakeBar approach with cmus_manager.

It provides a workaround for #4904 because the text can now change depending on the playback status.

It intends to keep full backwards compatibility with the exception of showing the artist for non-local songs if available.

Any feedback (including "No thank you") is welcome, especially for the tests. I didn't really know what I was doing and when I wanted to inject a different configuration for the test of duration etc., I looked at other tests and found this `_manager` approach, which I then tried to adapt. One thing led to another, and things quickly escalated to the point where I found myself desperately writing things like

```python
widget.eval("subprocess.index = 1")
```

because I found that directly interating with `MockCmusRemoteProcess` did not have any effect. 😔 

Let me know what you think. 😄  As I wrote above, this does not solve the underlying issue of #4904 (and I'm still not sure what the best way to fix that is), but it provides are useful workaround for me personally (I just copied the cmus status line for my config which displays ">" when playing and "|" when paused), and it adds some hopefully useful features while (hopefully!) not breaking any existing configs.

image
image


saviola777:
Added handling for the case where cmus is stopped and no file is selected.

tych0 (approved):

saviola777:
I realized I reverted to some of my bad naming habits when I named the new format variables `format_stream` and `format_no_artist`. Want me to rename them to `stream_format` and `no_artist_format` to be consistent with the rest?

tych0:
yeah, that would be great, thank you

saviola777:
Done :+1: 

Also fixed a small typo in the docstring.

PR #4919, opened 2024-07-10
OpenWeather widget: use https query URL

saviola777:
This changes the OpenWeather widget query URL to use https instead of http.

If preferred, I could make this a widget option.


jwijenbergh:
Thanks

Issue #4897, opened 2024-06-23
`reload_config` causes systray (GDK) windows to die


ervinpopescu:
### Issue description

Example of `blueman-tray` output:
```bash
$ blueman-tray

(blueman-tray:58207): Gdk-WARNING **: 15:02:59.366: GdkWindow 0x2a00003 unexpectedly destroyed

(blueman-tray:58207): Gdk-WARNING **: 15:02:59.366: The program 'blueman-tray' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 281 error_code 3 request_code 18 (core protocol) minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

Similar behavior with:
```
mictray
nm-applet
blueman-applet
systray_profile.py (custom GDK app)
pa-applet
```

With the default config:
```
$ blueman-tray

(blueman-tray:73413): Gdk-WARNING **: 15:13:19.501: GdkWindow 0x2a00002 unexpectedly destroyed

(blueman-tray:73413): Gdk-WARNING **: 15:13:19.502: The program 'blueman-tray' received an X Window System error.
This probably reflects a bug in the program.
The error was 'RenderBadPicture (invalid Picture parameter)'.
  (Details: serial 281 error_code 143 request_code 139 (RENDER) minor_code 7)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

Not sure why the difference in errors. `RenderBadPicture (invalid Picture parameter)` vs `BadWindow (invalid Window parameter)`

`flameshot` (`qt`) is fine, so this is why I ruled out `qt`.

I would like some pointers for further debugging in addition to the current info returned by the error, if that's possible. Thanks in advance!

### Version

0.26.1.dev30+ge2158512

### Backend

X11 (default)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
[...]

PR #4921, opened 2024-07-11
Revert "remember to finalize widgets"

tych0:
This reverts commit 7c3fbef756c451031e7d822dafed0b54eec0f2ac.

We have a more serious bug (systray windows dying) reported with this commit than the simple memory leak it fixes. We want to do a release to capture other fixes, so let's revert this for now.

See #4897


jwijenbergh:
Merci!

PR #4922, opened 2024-07-14
[feat(tasklist)]: skip windows with _NET_WM_STATE_SKIP_TASKBAR

ervinpopescu:
Also fixup a little bug in scratchpad from #4770. :)


elParaguayo:
Thanks. This all looks good to me.

PR #4935, opened 2024-07-27
Better finalisation of objects

elParaguayo:
Tries to add some sensible hierarchy to finalisation of objects: Screen -> Gap -> Widgets

Also fixes bug by finalising wallpaper buffer and surface objects on Wayland backend.

### DIFF
--- a/libqtile/backend/base/core.py
+++ b/libqtile/backend/base/core.py
@@ -38,6 +38,9 @@ class Core(CommandObject, metaclass=ABCMeta):
     def finalize(self):
         """Destructor/Clean up resources"""
 
+    def _finalize_wallpapers(self):
+        """Destroy wallpaper objects"""
+
     @property
     @abstractmethod
     def display_name(self) -> str:
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -421,6 +421,7 @@ class Core(base.Core, wlrq.HasListeners):
 
         if self._xwayland:
             self._xwayland.destroy()
+        self._finalize_wallpapers()
         self._cursor_manager.destroy()
         self.cursor.destroy()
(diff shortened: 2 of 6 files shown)


tych0:
This seems like an attempt to resurrect 2be6b7eace1b3309c740239c05a3159f2e8a65eb  ; have you confirmed that wallpapers and widgets in the systray work after a restart?

elParaguayo:
> This seems like an attempt to resurrect [2be6b7e](https://github.com/qtile/qtile/commit/2be6b7eace1b3309c740239c05a3159f2e8a65eb) ; have you confirmed that wallpapers and widgets in the systray work after a restart?

Not yet. This was primarily driven by an issue reported on discord where a wayland user noted increasing memory usage when reloading three config. This does fix that issue.

tych0:
I see, ok, thanks. We should definitely fix this, I just want to make sure we don't reintroduce that bug. I'm on vacation for another week here, but should be back to "normal" levels of focus soon and can test this out.

elParaguayo:
I've flipped this to being draft for now so I can look at those other points and we can test carefully before merging.

ervinpopescu:
> This seems like an attempt to resurrect [2be6b7e](https://github.com/qtile/qtile/commit/2be6b7eace1b3309c740239c05a3159f2e8a65eb) ; have you confirmed that wallpapers and widgets in the systray work after a restart?

Just checked this now and they don't :)

elParaguayo:
I haven't tried to fix the Systray yet.

ervinpopescu:
Gotcha, didn't take a good look at the code, just glanced over it and I wasn't sure what you wanted to fix here. Let me know if and how I can help.

elParaguayo:
I want to fix the Systray issue too however the main goal was to fix a memory leak with wallpapers first.

PR #4937, opened 2024-07-28
Add rhythmbox widget

theincompletedetective:
To submit a pull request for Rhythmbox widget in official Qtile repo

### DIFF
--- a/libqtile/widget/rhythmbox.py
+++ b/libqtile/widget/rhythmbox.py
@@ -52,6 +52,8 @@ class Rhythmbox(base.ThreadPoolText):
     - ``%te``: track elapsed time
 
     Rhythmbox (https://www.rhythmbox.org/) and Playerctl (https://github.com/altdesktop/playerctl) should be installed.
+
+    This widget supports the following bar orientations: horizontal and vertical.
     """
 
     defaults = [
--- a/libqtile/widget/rhythmbox.py
+++ b/libqtile/widget/rhythmbox.py
@@ -1,0 +1,116 @@
+# Copyright (C) 2015, Juan Riquelme González
+# Copyright (C) 2023, Elijah Smith
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with this program. If not, see
+# <http://www.gnu.org/licenses/>.
+
+import subprocess
+
+from functools import partial
+
+from libqtile import pangocffi
+from libqtile.widget import base
+
+
+class Rhythmbox(base.ThreadPoolText):
+    """
+    A simple Rhythmbox widget that is based on the Cmus widget.
+
+    Show details about the paused or playing song, and allow basic mouse control from the bar:
+        - Toggle pause (if playing) or play (if paused) on left click
+        - Skip forward in playlist on scroll up
+        - Skip backward in playlist on scroll down
+
+    Rhythmbox (https://www.rhythmbox.org/) and Playerctl (https://github.com/altdesktop/playerctl) should be installed.
+    
+    Format Options
+    --------------
+    %at - album title
+    %aa - album artist
+    %aA - album artist (lowercase)
+    %as - album artist sortname
+    %aS - album artist sortname (lowercase)
+    %ay - album year
+    %ag - album genre
+    %aG - album genre (lowercase)
+    %an - album disc number
+    %aN - album disc number, zero padded
+    %st - stream title
+    %tn - track number (i.e., 8)
+    %tN - track number, zero padded (i.e., 08)
+    %tt - track title
+    %ta - track artist
+    %tA - track artist (lowercase)
+    %ts - track artist sortname
+    %tS - track artist sortname (lowercase)
+    %td - track duration
+    %te - track elapsed time
+
+    Variables can be combined using quotes. For example, "%tn %aa %tt", will print the track number followed
+    by the artist and the title of the track.
+
+    The available options can also be found here: https://man.archlinux.org/man/rhythmbox-client.1.en#FORMAT_OPTIONS.
+    """
+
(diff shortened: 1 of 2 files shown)


libqtile/widget/rhythmbox.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with this program. If not, see
+# <http://www.gnu.org/licenses/>.
```
elParaguayo, line 12: No idea why `Cmus` has a different licence to the rest of qtile...

libqtile/widget/rhythmbox.py
```diff
@@ -0,0 +1,117 @@
...
+class Rhythmbox(base.ThreadPoolText):
+    """
+    A simple Rhythmbox widget that is based on the Cmus widget.
+
+    Show metadata about the currently paused or playing song, and allow basic mouse control from the bar:
+        - Toggle pause (if playing) or play (if paused) on left click
+        - Skip forward in playlist on scroll up
+        - Skip backward in playlist on scroll down
+
+    The following options (extracted from ``man rhythmbox-client``) are available for the `format` string:
```
elParaguayo, line 31:
```suggestion
    The following options (extracted from ``man rhythmbox-client``) are available for the ``format`` string:
```

libqtile/widget/rhythmbox.py
```diff
@@ -0,0 +1,117 @@
...
+    - ``%ta``: track artist
+    - ``%tA``: track artist (lowercase)
+    - ``%ts``: track artist sortname
+    - ``%tS``: track artist sortname (lowercase)
+    - ``%td``: track duration
+    - ``%te``: track elapsed time
+
+    Rhythmbox (https://www.rhythmbox.org/) and Playerctl (https://github.com/altdesktop/playerctl) should be installed.
+
+    This widget supports the following bar orientations: horizontal and vertical.
```
elParaguayo, line 56:
```suggestion
    `Rhythmbox`_ and `Playerctl`_ should be installed.

    This widget supports the following bar orientations: horizontal and vertical.
    
    .. _Rhythmbox: https://www.rhythmbox.org/
    .. _Playerctl: https://github.com/altdesktop/playerctl

```

elParaguayo:
Is this widget needed?

If you're using `playerctl` then that means it should have an mpris interface which then means you can use the `Mpris2` widget. I'd be reluctant to add a new widget if an existing one already works.

libqtile/widget/rhythmbox.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, **config):
+        base.ThreadPoolText.__init__(self, "", **config)
+        self.add_defaults(Rhythmbox.defaults)
+        self.status = ""
+
+        self.add_callbacks(
+            {
+                "Button1": self.play,
+                "Button4": partial(subprocess.Popen, ["rhythmbox-client", "--next"]),
+                "Button5": partial(subprocess.Popen, ["rhythmbox-client", "--previous"]),
```
elParaguayo, line 78: Using `Popen` and not `wait()`ing for the process to exit will create zombie processes.

```suggestion
                "Button4": lazy.spawn("rhythmbox-client --next"),
                "Button5": lazy.spawn("rhythmbox-client --previous"),
```

libqtile/widget/rhythmbox.py
```diff
@@ -0,0 +1,117 @@
...
+            }
+        )
+
+    def play(self):
+
+        if self.status == "Stopped":
+            subprocess.Popen(["rhythmbox-client", "--play"])
+        elif self.status in ("Paused", "Playing"):
+            subprocess.Popen(["rhythmbox-client", "--play-pause"])
```
elParaguayo, line 87:
```suggestion
        if self.status == "Stopped":
            self.qtile.spawn("rhythmbox-client --play")
        elif self.status in ("Paused", "Playing"):
            self.qtile.spawn("rhythmbox-client --play-pause")
```

esmith1412:
> Is this widget needed?
> 
> If you're using `playerctl` then that means it should have an mpris interface which then means you can use the `Mpris2` widget. I'd be reluctant to add a new widget if an existing one already works.

`playerctl` also supports `cmus`, so I'm not sure why there would need to be a widget for it either.

elParaguayo:
Agreed. I'm not sure why we need the cmus widget either in that case.

esmith1412:
According to the MPRIS v2 metadata guidelines used by the `Mpris2` widget linked [here](http://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index5h3), there isn't a way that I can see to display the track's position _and_ length: only the length is available, but both `playerctl` and `rhythmbox-client` allow for both to be displayed.

elParaguayo:
The position is not part of the metadata, it's a separate value but it's still accessible over dbus. It should hopefully be fairly trivial to add support to the mpris2 widget for this. 

The mpris2 widget in qtile-extras accesses the position info but only uses it to display on a popup.

elParaguayo:
Ah, I remember the issue with the position...

The mpris2 widget works by listening for signals when the playback status or metadata change. However, the position doesn't send a signal when it changes (as it's always changing during playback) so you have to poll the data instead. Should still be possible to add support for it.

esmith1412:
> Ah, I remember the issue with the position...
> 
> The mpris2 widget works by listening for signals when the playback status or metadata change. However, the position doesn't send a signal when it changes (as it's always changing during playback) so you have to poll the data instead. Should still be possible to add support for it.

Would this require changing the base widget, or can it be done without doing that?

elParaguayo:
I'd probably edit the Mpris2 widget to include polling the position _if_ the format string included the position field.

I can look at implementing this later.

esmith1412:
> I'd probably edit the Mpris2 widget to include polling the position _if_ the format string included the position field.
> 
> I can look at implementing this later.

I'd be more than happy to work on it, if you don't have time for it now.

elParaguayo:
I'm working on something. Will let you know when it's ready to test.

elParaguayo:
You can take a look at #4944. There are some bugs in there still but feedback would be great.

PR #4938, opened 2024-07-29
Add secondary window auto-maximization to xmonad.

marcopaganini:
- Add the `auto_maximize` parameter to MonadTall. When set, this will cause the secondary window in focus to automatically maximize, and keep the non-focused secondary windows minimized to `min_secondary_size`. This produces an effect somewhat similar to i3's "stacked" layout on the secondary window.

- Add a new lazy method: `toggle_auto_maximize`, which allows the feature to be dynamically toggled.

- Calling reset will also turn the feature off (default state.)

- This PR is a port of the qtile-monadstack layout introduced in https://github.com/marcopaganini/qtile-monadstack into the main tree.

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -14,6 +14,7 @@
 # Copyright (c) 2014 dequis
 # Copyright (c) 2014 Florian Scherf
 # Copyright (c) 2017 Dirk Hartmann
+# Copyright (c) 2024 Marco Paganini (auto_maximization code).
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -108,6 +109,21 @@ class MonadTall(_SimpleLayoutBase):
     methods. As mentioned the main pane is considered the top of the stack;
     moving up is counter-clockwise and moving down is clockwise.
 
+    ::
+
+        ---------------------          ---------------------
+        |            |      |          |            |      |
+        |            |______|          |            |Focus |
+        |            |      |          |            |      |
+        |            |______|          |            |______|
+        |            |      |          |            |______|
+        |            |      |          |            |      |
+        ---------------------          ---------------------
+
+    Setting ``auto_maximize`` will cause the focused secondary pane to be
+    automatically maximized on focus. The non-maximized panes will shrink to
+    the height specified by ``min_secondary_size``.
+
     The opposite is true if the layout is "flipped".
 
     ::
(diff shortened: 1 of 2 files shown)


marcopaganini:
I've fixed the problem with `test_xmonad.py` but now I'm getting an error in `test_check_migrations.py`. In my machine, even pulling from upstream/master I still get an error. Could you please approve the workflow so I can see if this also happens in the CI?

marcopaganini:
Strangely, one of xmonad tests runs OK locally, but fails when running on CI. I added an explicit assertion on the window name. Let's see if it works.

marcopaganini:
I finally managed to install all the wayland components to test this. Something interesting happens:

```
for i in {1..100};
  do tox -e test-wayland -- -x test/layouts/test_xmonad.py::test_tall_stacked_window_kill
  echo $? >>/tmp/results.txt
done

$ sort /tmp/results.txt | uniq -c
     69 0
     31 1
 ```
 
So, the same test fails 31% of the time on my computer. This is strange as the test is very simple. @elParaguayo any tips on how to proceed here? I'm a bit stuck, unfortunately.

**Comment by elParaguayo** [comment]
I suspect that's a race condition. The `kill()` command will return immediately and the test will move on to the assertion in the next line but, in reality, the window may not have closed yet.

The way to deal with this is to check the window count before moving on to the assertion.

Try something like this:
```python
from test.helpers import Retry

@monadtallstacked_config
def test_tall_stacked_window_kill(manager):

    @Retry(ignore_exceptions=(AssertionError))
    def assert_window_count(num):
        assert len(manager.c.windows()) == num

    manager.test_window("one")
    assert_focused(manager, "one")

    manager.test_window("two")
    assert_focused(manager, "two")

    manager.test_window("three")
    assert_focused(manager, "three")

    manager.c.layout.previous()
    assert_focused(manager, "two")
    assert_dimensions(manager, 400, 0, 396, 511)

    manager.c.window.kill()
    assert_window_count(2)
    assert_focused(manager, "one")
    assert_dimensions(manager, 0, 0, 396, 596)

    manager.c.layout.next()
    assert_focused(manager, "three")
    assert_dimensions(manager, 400, 0, 396, 596)
```

The `Retry` decorator will handle multiple attempts with delays in between retries. If it fails 5 times it will then raise an assertionerror.

marcopaganini:
Fantastic! This seems to have done the trick! Let's see how CI will behave.

marcopaganini:
Ping? Could someone run the workflows? I found a small bug and fixed it, so I had to re-push it.

elParaguayo:
This looks good. Anything else you want to do before we merge this?

marcopaganini:
I've been using it for a while without any problems. I think it's good to
merge.

Thanks!

elParaguayo:
Done.

PR #4944, opened 2024-08-01
Add length and position support to Mpris2 widget

elParaguayo:
Draft for now as some bugs to iron out...

This PR adds support for track position and track length display e.g.:
```python
widget.Mpris2(format="{xesam:title} - {xesam:album} - {xesam:artist} - {position}/{mpris:length}"),
```

How it works:
- The widget subscribes to the `Seeked` signal which reports the track position when it changes in an unexpected way
- The widget sets a time to increment the position every second (i.e. it doesn't poll the player for the position)

Issues I've found so far:
- I'm guessing some players don't emit the `Seeked` signal at the start of a new track as the [spec](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Signal:Seeked) doesn't require it
- The display timer is not synced to the actual timer on the player
- The position increments on a timer without polling so there may be some drift over time if the increments are delayed due to qtile doing other things.
- `vlc` reports a track length that's twice as long as it should be (but corrects this when the track isn't playing?!)
- `ncspot` doesn't implement the `Seeked` signal on the mpris interface

I would welcome people testing this and reporting bugs to me.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -477,6 +477,29 @@ async def _send_dbus_message(
     return bus, msg
 
 
+async def add_match_rule(
+    bus: MessageBus, preserve: bool = False, **kwargs: str | None
+) -> tuple[MessageBus, Message]:
+    rule = ",".join("{}='{}'".format(k, v) for k, v in kwargs.items() if v)
+
+    logger.debug("Adding dbus match rule: %s", rule)
+
+    bus, msg = await _send_dbus_message(
+        False,
+        MessageType.METHOD_CALL,
+        "org.freedesktop.DBus",
+        "org.freedesktop.DBus",
+        "/org/freedesktop/DBus",
+        "AddMatch",
+        "s",
+        [rule],
+        bus=bus,
+        preserve=preserve,
+    )
+
+    return bus, msg
+
+
 async def add_signal_receiver(
     callback: Callable,
     session_bus: bool = False,
@@ -510,36 +533,35 @@ async def add_signal_receiver(
             )
             return False
 
+    if use_bus is None:
+        if session_bus:
+            bus_type = BusType.SESSION
+        else:
+            bus_type = BusType.SYSTEM
+
+        try:
+            bus = await MessageBus(bus_type=bus_type).connect()
+        except (AuthError, Exception):
+            logger.warning("Unable to connect to dbus.")
+            return False
+    else:
+        bus = use_bus
+
     match_args = {
-        "sender": bus_name,
-        "member": signal_name,
-        "path": path,
+        "type": "signal",
         "interface": dbus_interface,
+        "path": path,
+        "member": signal_name,
+        "sender": bus_name,
     }
 
-    rule = "type='signal',"
-    rule += ",".join("{}='{}'".format(k, v) for k, v in match_args.items() if v)
-
-    logger.debug("Adding dbus match rule: %s", rule)
-
-    bus, msg = await _send_dbus_message(
-        session_bus,
-        MessageType.METHOD_CALL,
-        "org.freedesktop.DBus",
-        "org.freedesktop.DBus",
-        "/org/freedesktop/DBus",
-        "AddMatch",
-        "s",
-        [rule],
-        bus=use_bus,
-        preserve=preserve,
-    )
+    bus, msg = await add_match_rule(bus, preserve=preserve, **match_args)
 
     # Check if message sent successfully
     if bus and msg and msg.message_type == MessageType.METHOD_RETURN:
 
         def match_message(msg: Message, match_args: dict[str, str | None]) -> bool:
-            return msg._matches(**{k: v for k, v in match_args.items() if v})
+            return msg._matches(**{k: v for k, v in match_args.items() if (v and k != "type")})
 
         async def resolve_sender(signal_msg: Message) -> tuple[str, Message]:
             """Looks up a pretty bus name to retrieve the unique name."""
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -304,14 +304,15 @@ class Mpris2(base._TextBox):
         if not self._tick:
             _, rem = divmod(self.position, 1e6)
             rem /= 1e6
+            rem = 1 - rem
             if rem > 0:
                 delay = rem
 
         return delay
 
-    def _set_position(self, message, seeked=True):
-        # if not self.needs_position:
-        #     return
+    def _set_position(self, message):
+        if not self.needs_position:
+            return
 
         self.position = message.body[0] // 1000000
         self._tick = 0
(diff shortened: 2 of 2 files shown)


esmith1412:
I'd love to test this, but I'm unclear on how I could do so within an actual Qtile environment, rather than through unit tests, which seems like the best way. I've read the Contributing and Hacking on Qtile sections of the wiki, but this seems to only deal with this from the perspective of unit tests.

elParaguayo:
Best way to test is to install qtile from the relevant branch (https://github.com/elParaguayo/qtile/tree/mpris2-position)

esmith1412:
> Best way to test is to install qtile from the relevant branch (https://github.com/elParaguayo/qtile/tree/mpris2-position)

Will do; I'll get on that, and provide any useful feedback that I can!

elParaguayo:
Thanks. NB this widget will break if you try to import it via qtile-extras. I'll need to fix that separately.

esmith1412:
I noticed that when playing videos on YouTube, the widget looks like this:
2024-08-04-16:52:57-screenshot. Notice how the end time doesn't show up at all. Is this a YouTube issue, or a widget issue?

elParaguayo:
Looks like youtube doesn't report track length.

esmith1412:
CMUS and Rhythmbox report the position and length correctly, but I noticed that when watching a video at 2x speed using VLC, the position shown in the widget is sometimes ahead of what the player reports, and sometimes behind; it never seems to be exactly right.

elParaguayo:
Interesting. That's because the widget is not checking the `Rate` property. Should be easy enough to add that.

elParaguayo:
Latest commit should fix that problem.

esmith1412:
When watching a video on YouTube at 2x speed, the position shows as <= half of the actual position; is this a byproduct of YouTube not providing the length, or something else?

elParaguayo:
It's probably not setting the `Rate` property.

esmith1412:
Your last commit seems to have fixed the 2x speed issue with VLC.

elParaguayo:
Yes, because VLC provides the Rate property so we can listen for changes. Can't do anything about it if a player doesn't announce it's playing at a different speed.

esmith1412:
When you reload the config while a track is playing, it doesn't reload the metadata of the currently-playing track, but shows nothing instead; you have to select another track, or wait until it automatically moves to the next one in the playlist.

esmith1412:
Also, it would be nice if a feature could be added to this widget to allow you to change the color of the text, if the song is playing or paused.

elParaguayo:
> > Yes, because VLC provides the Rate property so we can listen for changes. Can't do anything about it if a player doesn't announce it's playing at a different speed.
> 
> So, the specification doesn't enforce providing rate or length?

Well, it does. This just seems to be a non-compliant implementation of it. However, it is the browser that creates the Mpris interface and there will also be some communication between the embedded player and the browser so I don't know where the fault lies.

> When you reload the config while a track is playing, it doesn't reload the metadata of the currently-playing track, but shows nothing instead; you have to select another track, or wait until it automatically moves to the next one in the playlist.

Yes. The widget works by listening for signals from the player. However, it should be possible to look for an active player when the widget starts and polling it.

> Also, it would be nice if a feature could be added to this widget to allow you to change the color of the text, if the song is playing or paused.

You can do this with qtile-extras as that version of the widget includes hooks for when the playback status changes so you can hook a function that will set the colour of the text.

esmith1412:
I've noticed that when I watch several YouTube videos in a row, without using any other MPRIS player in between, the widget doesn't reset the position, but treats it as if I'm watching one extremely long video.

elParaguayo:
Yes. We've established that the mpris implementation by browsers is not good!

However, it should be possible to cater for this.

elParaguayo:
```
❯ busctl -j --no-pager --user call org.mpris.MediaPlayer2.firefox.instance_1_9 /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties GetAll s org.mpris.MediaPlayer2.Player
{
	"type" : "a{sv}",
	"data" : [
		{
			"PlaybackStatus" : {
				"type" : "s",
				"data" : "Playing"
			},
			"Metadata" : {
				"type" : "a{sv}",
				"data" : {
					"mpris:trackid" : {
						"type" : "o",
						"data" : "/org/mpris/MediaPlayer2/firefox"
					},
					"xesam:title" : {
						"type" : "s",
						"data" : "Day 14: BBC TV Coverage"
					},
					"xesam:album" : {
						"type" : "s",
						"data" : ""
					},
					"xesam:artist" : {
						"type" : "as",
						"data" : [
							""
						]
					},
					"mpris:artUrl" : {
						"type" : "s",
						"data" : "file:///home/elp/.mozilla/firefox/firefox-mpris/822_81.png"
					}
				}
			},
			"CanGoNext" : {
				"type" : "b",
				"data" : false
			},
			"CanGoPrevious" : {
				"type" : "b",
				"data" : false
			},
			"CanPlay" : {
				"type" : "b",
				"data" : true
			},
			"CanPause" : {
				"type" : "b",
				"data" : true
			},
			"CanSeek" : {
				"type" : "b",
				"data" : false
			},
			"CanControl" : {
				"type" : "b",
				"data" : true
			}
		}
	]
}
```
You can see the browser doesn't provide position or length data.

theincompletedetective:
When watching a video on YouTube, if the video stops playing because the part I'm about to watch hasn't loaded, the position keeps incrementing, until I press the pause button.

elParaguayo:
Again, this is a failing of the implementation of the player/browser. If it doesn't signal that it's paused because it's buffering then the widget can't react. If the player doesn't expose the position date, the widget can't react.

I guess we need to check if position/length is supported and, if not, hide that data.

theincompletedetective:
I haven't been able to find any more issues with this, so as far as I can tell, it seems like this will work well once merged.

elParaguayo:
I'm not sure. It feels like this could be buggy and I haven't implemented anything to hide those fields if the data isn't available.

theincompletedetective:
@elParaguayo, are you in any better of a position to implement this feature now?

elParaguayo:
I don't particularly like this PR. I think it could be implemented a bit better.

theincompletedetective:
Please let me know if there's anything I can do to help you implement it.

elParaguayo:
The issue I've got with this is that clients are not expected to poll the player for the track position. Instead, they're meant to increment the time based on the `Rate` property and adjust when `PlaybackStatus` changes or if there's a `Seeked` signal. We can, obviously, poll intermittently too to make sure there's no drifting.

However, we still can't guarantee correct behaviour (particularly if players don't implement the spec correctly) and so there may be scenarios where we're showing the playback time increasing when it shouldn't be.

We also need to think about how we hide playback time if the player doesn't implement it. I think we do this with new `playback_time_format` and `no_playback_time_format` options.

I can have another go at implementing it. It could always end up in `qtile-extras` if I don't think the changes work well enough to be included in the main repo.

elParaguayo:
@theincompletedetective I've rebased this and pushed to #5911.

Would be great if you could test and add feedback/suggestions.

PR #4963, opened 2024-08-14
Allow setting the exitcode of qtile via shutdown command

phragment:
I would like to have the possibility to set the exitcode of the Qtile process by specifying it (optionally) using the shutdown command.
This would allow for a communication of poweroff or reboot intents to a login manager.

I'm currently using greetd with gtkgreet as login manager. This feature would make it possible to implement a proper shutdown of the whole system.

### DIFF
--- a/libqtile/core/lifecycle.py
+++ b/libqtile/core/lifecycle.py
@@ -19,6 +19,7 @@ class LifeCycle:
     # referenced here when atexit fires will NOT be finalized properly.
     def __init__(self) -> None:
         self.behavior = Behavior.NONE
+        self.exitcode: int = 0
         self.state_file: str | None = None
         atexit.register(self._atexit)
 
@@ -35,6 +36,10 @@ class LifeCycle:
             os.execv(sys.executable, argv)
         elif self.behavior is Behavior.TERMINATE:
             logger.warning("Qtile will now terminate")
+            # the if statement should be unnecessary, but keeps code coverage working
+            # calling os._exit prevents later injected atexit handlers from running
+            if self.exitcode:
+                os._exit(self.exitcode)
         elif self.behavior is Behavior.NONE:
             pass
 
--- a/test/core/test_exitcode.py
+++ b/test/core/test_exitcode.py
@@ -1,0 +1,101 @@
+# Copyright (c) 2024 Thomas Krug
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import os
+import subprocess
+import tempfile
+import threading
+import time
+
+import test
+
+repo_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+socket_path = tempfile.NamedTemporaryFile().name
+
+
+def run_qtile(backend):
+    cmd = os.path.join(repo_path, "bin/qtile")
+    args = [cmd, "start", "-s", socket_path]
+    args.extend(["-b", backend.name])
+
+    env = os.environ.copy()
+    if backend.name == "wayland":
+        env.update(backend.env)
+
+    proc = subprocess.Popen(args, env=env, stdout=subprocess.PIPE)
+    out, err = proc.communicate(timeout=10)
+    exitcode = proc.returncode
+    return exitcode
+
+
+def stop_qtile(code):
+    cmd = os.path.join(repo_path, "bin/qtile")
+    args = [cmd, "cmd-obj", "-o", "cmd", "-s", socket_path, "-f", "shutdown"]
+    if code:
+        args.extend(["-a", str(code)])
+    proc = subprocess.Popen(args, stdout=subprocess.PIPE)
+    proc.communicate(timeout=10)
+    exitcode = proc.returncode
+    return exitcode
+
+
+def deferred_stop(code=0):
+    # wait for qtile process to start
+    wait = 10
+    while not test.helpers.can_connect_qtile(socket_path):
+        time.sleep(1)
+        if not wait:
+            raise Exception("timeout waiting for qtile process")
+        wait -= 1
+
+    stop_qtile(code)
+
+
+# in these testcases there are two qtile instances
+# one started by the helpers.TestManager,
+# which is used to evaluate code coverage
+# and a second instance started by run_qtile,
+# which is used to test the actual exit code behavior
+def test_exitcode_default(manager):
+    thread = threading.Thread(target=deferred_stop)
+    thread.daemon = False
+    thread.start()
+
+    exitcode = run_qtile(manager.backend)
+
+    # shutdown qtile instance started by testmanager
+    manager.c.shutdown()
+
+    assert exitcode == 0
+
+
+def test_exitcode_explicit(manager):
+    code = 23
+
+    thread = threading.Thread(target=deferred_stop, args=(code,))
+    thread.daemon = False
+    thread.start()
+
+    exitcode = run_qtile(manager.backend)
+
+    # shutdown qtile instance started by testmanager
+    manager.c.shutdown()
+
+    assert exitcode == code
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -236,9 +236,10 @@ class Qtile(CommandObject):
             self.finalize()
             self.core.remove_listener()
 
-    def stop(self) -> None:
+    def stop(self, exitcode: int = 0) -> None:
         hook.fire("shutdown")
         lifecycle.behavior = lifecycle.behavior.TERMINATE
+        lifecycle.exitcode = exitcode
         self.core.graceful_shutdown()
         self._stop()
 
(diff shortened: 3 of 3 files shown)


libqtile/core/lifecycle.py
```diff
@@ -19,6 +19,7 @@ class LifeCycle:
     # referenced here when atexit fires will NOT be finalized properly.
     def __init__(self) -> None:
         self.behavior = Behavior.NONE
+        self.exitcode: int = 0
```
tych0, line 22: This seems unused? Also, can you add a test?

tych0: Oh, no, it is used :). But can you add a test?

phragment: It is not stricly needed to specify the member here, because python will just add the variable dynamically (from stop method). This is used to save the exitcode between termination of the shutdown command method and the real shutdown in LifeCycle class. Of course the member variable is only needed for a rather short amount of time. But I did not see another way to communicate the exitcode to "the last line" of qtile in its execution.
Should I try to write a test starting qtile, issuing the shutdown command and checking the exitcode?

tych0: > Should I try to write a test starting qtile, issuing the shutdown command and checking the exitcode?

Yep, that sounds great, thank you.

>  But I did not see another way to communicate the exitcode to "the last line" of qtile in its execution.

Yep, no worries, it's fine as is, I was just confused.

phragment:
I'm not too sure about the test implementation. It should be a good end to end test, but I don't know how this affects the test coverage. (And I'm at a loss why the test coverage decreases.)

Is it possible to check for the selected backend e.g. using the manager instance? (see TODO: https://github.com/qtile/qtile/commit/bafecba2e3fac0c19af51d8d5a05ca720e1b7925)

tych0:
The test looks good to me, The coverage dropping by 3% seems like a bug: you aren't adding 3% more code, so it's not possible that it can drop that far. I am personally not very concerned about test coverage, I think it's kind of a Goodhart's law thing.

That said: I can force merge this, but I don't understand how to get the tests to run again. Seems like it's blocked from running by a previously failed coverage report?

tych0:
Oh: maybe it's because you're fork+exec-ing another copy of qtile, and the coverage stuff is getting injected, and it's *that* copy of qtile that's writing out its coverage? That's a bit annoying...

elParaguayo:
> Oh: maybe it's because you're fork+exec-ing another copy of qtile, and the coverage stuff is getting injected, and it's _that_ copy of qtile that's writing out its coverage? That's a bit annoying...

I'm not mad about reducing the test coverage as it increases the risk of us introducing bugs. So I would prefer to solve this if we can.

elParaguayo:
> > Oh: maybe it's because you're fork+exec-ing another copy of qtile, and the coverage stuff is getting injected, and it's _that_ copy of qtile that's writing out its coverage? That's a bit annoying...
> 
> I'm not mad about reducing the test coverage as it increases the risk of us introducing bugs. So I would prefer to solve this if we can.

Actually, my comment is a bit stupid as the tests still pass. However, we're now just not getting accurate details about how much of the code is covered by the tests. That being said, I know the coverage wasn't 100% accurate to begin with.

phragment:
Thank you both for the review and input, I'll work on it this weekend.
Regarding the test coverage. I think tych0 is right and coverage metrics are taken from "the wrong Qtile instance".
My current problem is either how to get the exitcode of the "manager qtile instance" including the injected coverage code or keep this test out of coverage metrics? I think the test by itself is a real end-to-end test which should be good to check the feature.

test/core/test_exitcode.py
```diff
@@ -0,0 +1,99 @@
...
+        args.extend(["-a", str(code)])
+    proc = subprocess.Popen(args, stdout=subprocess.PIPE)
+    out, err = proc.communicate(timeout=10)
+    exitcode = proc.returncode
+    return exitcode
+
+
+def deferred_stop(code=0):
+    # wait for qtile process to start
+    time.sleep(3)
```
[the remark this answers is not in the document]
phragment: I'll check what I can do. I understand this could be an issue with the test system and I try to keep it in mind. (e.g. I already limited the test runtime by specifying a timeout for the subprocesses.
Is there an easy, maybe builtin way to check for the socket?

tych0:
> My current problem is either how to get the exitcode of the "manager qtile instance" including the injected coverage code or keep this test out of coverage metrics?

I think it's fine to keep it out of the metrics, but I googled around for a bit and couldn't find an obvious method to actually accomplish that :(

phragment:
As I get a better understanding of the test system, I was able to implement a lot shorter test. But it isn't working anymore, probably because the Qtile test instance started by the helpers.TestManager does not use/instanciate the LifeCycle object?
I'm not sure on how to proceed.

test/core/test_exitcode.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@minimal_config
+def test_exitcode_explicit(manager):
+    code = 23
+
+    manager.c.shutdown(code)
+
+    manager.proc.join(10)
+    exitcode = manager.proc.exitcode
```
tych0, line 45: My guess is that the "same thing" is happening: because qtile exits here without doing that much, it writes down "i didn't have much coverage", and then somehow that gets interpreted as coverage for the whole thing.

Unfortunately, I don't know much about how the coverage statistics actually work, so I'm not sure how to advise you to proceed. The best thing I can think is to try and unset some env vars or unload some injected modules from this copy of the manager... maybe via manager.c.eval()? But I don't know what those env vars / injected modules would be, and I suspect they would be subject to change. Maybe coveralls has a special env var you can set for "don't report coverage"?

phragment: After some testing I believe the issue is os._exit() leads to the installed coverage atexit handler not being run.
But I have to use the _exit variant because otherwise we would get a loop on shutdown and I need to explicitly call an exit method to set the exitcode.

I would have to mock the handler method removing os._exit which would reduce the test to absurdity.
My suggestion is to remove the test case or just test the command setting the exitcode within the lifecycle intsance.

phragment: Ok first up sorry for the noise, I will squash the commits.
What I came up with is that always calling os._exit lead to not executing the atexit handlers (pytest, cov) and thereby the coverage reports were missing for a lot of testcases.
This is fixed by the if statement in lifecycle.

On the test side I went back to starting own Qtile instances, which seems to work fine.
I just have a problem when creating Qtile with the wayland backend, which is why I currently skip the test for wayland.

phragment:
Ok first up sorry for the noise, I will squash the commits.
What I came up with is that always calling os._exit lead to not executing the atexit handlers (pytest, cov) and thereby the coverage reports were missing for a lot of testcases.
This is fixed by the if statement in lifecycle.

On the test side I went back to starting own Qtile instances, which seems to work fine.
I just have a problem when creating Qtile with the wayland backend, which is why I currently skip the test for wayland.

elParaguayo:
Nice work! I was trying to play with that `os._exit` line too as I figured that was probably causing an issue with the handlers as you say. Great to see tests passing now.

What's the issue with the Wayland backend?

phragment:
Thank you for your kind words.
My issue with the Wayland backend ist that starting Qtile using "qtile start -b wayland" from within the test environment results in the following error:

```
$ pytest --backend=wayland test/core/test_exitcode.py 
=============================================== test session starts ================================================
platform linux -- Python 3.12.4, pytest-8.2.2, pluggy-1.5.0
rootdir: /home/thomas/qtile
configfile: pyproject.toml
plugins: cov-5.0.0
collected 2 items                                                                                                  

test/core/test_exitcode.py EE                                                                                [100%]

====================================================== ERRORS ======================================================
________________________________ ERROR at setup of test_exitcode_default[1-wayland] ________________________________

request = <SubRequest 'backend' for <Function test_exitcode_default[1-wayland]>>, backend_name = 'wayland'
xephyr = None
wayland_session = {'WLR_BACKENDS': 'headless', 'WLR_HEADLESS_OUTPUTS': '1', 'WLR_LIBINPUT_NO_DEVICES': '1', 'WLR_RENDERER': 'pixman', ...}

    def backend(request, backend_name, xephyr, wayland_session):
        if backend_name == "x11":
            from test.backend.x11.conftest import XBackend
    
            yield XBackend({"DISPLAY": xephyr.display}, args=[xephyr.display])
        elif backend_name == "wayland":
            from test.backend.wayland.conftest import WaylandBackend
    
>           yield WaylandBackend(wayland_session)

test/conftest.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <test.backend.wayland.conftest.WaylandBackend object at 0x74d66d95d400>
env = {'WLR_BACKENDS': 'headless', 'WLR_HEADLESS_OUTPUTS': '1', 'WLR_LIBINPUT_NO_DEVICES': '1', 'WLR_RENDERER': 'pixman', ...}
args = ()

    def __init__(self, env, args=()):
        self.env = env
        self.args = args
>       self.core = Core
E       NameError: name 'Core' is not defined

test/backend/wayland/conftest.py:56: NameError
```

I'm probably missing the correct environment variables. Any hints would be appreciated.

tych0:
that's probably because the import https://github.com/qtile/qtile/blob/master/test/backend/wayland/conftest.py#L9-L14 failed, my guess is you haven't built installed the wayland backend correctly. What does `python3 -c 'from libqtile.backend.wayland.core import Core'` show?

tych0:
you could also just try to push your test to CI here. if it works in x11 it'll probably work in wayland since it's backend agnostic.

phragment:
> What does `python3 -c 'from libqtile.backend.wayland.core import Core'` show?

```
$ python3 -c 'from libqtile.backend.wayland.core import Core'
  File "<string>", line 1, in <module>
  File "/home/thomas/qtile/libqtile/backend/wayland/__init__.py", line 23, in <module>
    from libqtile.backend.wayland.inputs import InputConfig  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomas/qtile/libqtile/backend/wayland/inputs.py", line 30, in <module>
    from libqtile.backend.wayland.wlrq import HasListeners, buttons
  File "/home/thomas/qtile/libqtile/backend/wayland/wlrq.py", line 30, in <module>
    from wlroots import ffi as wlr_ffi
ImportError: cannot import name 'ffi' from 'wlroots' (unknown location)
```

I use a dedicated test system which is Arch Linux and has no Qtile package installed system wide (not my productive Qtile setup).

phragment:
I deleted my current venv and recreated it using dev.sh
Afterwards I ran as Sydiepus suggested (with venv activated):

```
pip install pywlroots
# arch issue
export CFLAGS="$CFLAGS -I/usr/include/wlroots0.17"
export LDFLAGS="$LDFLAGS -L/usr/lib/wlroots0.17"
# unclear issue
export PYTHONPATH=$(pwd):$PYTHONPATH
./scripts/ffibuild -v
```

Now `python3 -c 'from libqtile.backend.wayland.core import Core'` works fine, but the prior error still remains:

```
$ pytest --backend=wayland test/core/test_exitcode.py 
[...]
RuntimeError: Failed to create wlroots backend
```

phragment:
Ok, so the issue was missing environmental variables for wlroots to start up without screen etc.

For future reference, if somebody stumbles onto this pullrequest (these are defined in backend.env):
{'WLR_BACKENDS': 'headless', 'WLR_LIBINPUT_NO_DEVICES': '1', 'WLR_RENDERER_ALLOW_SOFTWARE': '1', 'WLR_RENDERER': 'pixman', 'XDG_RUNTIME_DIR': '/tmp', 'WLR_HEADLESS_OUTPUTS': '1', 'WAYLAND_DISPLAY': 'wayland-0'}

phragment:
I'm pretty confident this is now in a good state.
Thank you for your support!

elParaguayo:
Excellent. Thanks for tracking this down.

PR #4965, opened 2024-08-17
Change focus_on_window_activation var to allow functions

MoonCubes:
When the _NET_ACTIVE_WINDOW message is send the focus_on_window_activation config variable is checked.
If it is a function (that takes the window as an argument) and returns True the window is focused. If it is a string then the old behaviors are used.
This could for example be used if you only want to focus the browser when you open a link, and it's in a different group.

This is how you could set that up:
```py
def focus_behavior(client: Window):
    return "Firefox-esr" in client.get_wm_class() or client.group.screen == qtile.current_screen
focus_on_window_activation = focus_behavior
```

### DIFF
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -118,7 +118,7 @@ configuration variables that control specific aspects of Qtile's behavior:
       - Floating windows are kept above tiled windows (Currently x11 only. Wayland support coming soon.)
     * - ``focus_on_window_activation``
       - ``'smart'``
-      - Behavior of the _NET_ACTIVATE_WINDOW message sent by applications
+      - Behavior of the _NET_ACTIVE_WINDOW message sent by applications
 
         - urgent: urgent flag is set for the window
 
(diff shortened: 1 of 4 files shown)


docs/manual/commands/keybindings.rst
```diff
@@ -11,6 +11,7 @@ Default configuration
 .. LS_PNG
 .. image:: /_static/keybindings/mod4.png
 .. image:: /_static/keybindings/mod4-shift.png
+.. image:: /_static/keybindings/control-mod1.png
```
tych0, line 14: This seems like a separate fix, right? can you put it in a separate patch?

docs/manual/config/index.rst
```diff
@@ -127,6 +127,9 @@ configuration variables that control specific aspects of Qtile's behavior:
         - smart: automatically focus if the window is in the current group
 
         - never: never automatically focus any window that requests it
+    * - ``window_activation_rules``
+      - ``[]``
+      - A list of Match objects of windows that will always be focused when the _NET_ACTIVE_WINDOW message is sent by applications.
```
tych0, line 132: I wonder: can we make `focus_on_window_activation` take a function, so that people can define this stuff on their own?

MoonCubes: @tych0 Should `focus_on_window_activation` only take a function or either a function or a string with the old behaviors? (so old configs don't break)

tych0: Yeah, we'll want old configs to stay working, so I think a string or a function should be allowable.

tych0:
Thanks, looks reasonable to me. Can you add a test to make sure we don't break this?

MoonCubes:
Done. I added it to the test_urgent_hook_fire test

PR #4974, opened 2024-08-25
Nix test fixes

jwijenbergh:
- Do not assume python interpreter path, but use sys.executable
- Make MPD mocking work

If this is green in the CI here I propose these patches are wanted because they also make our tests work with Nix


elParaguayo (approved):
LGTM

PR #4334, opened 2023-07-09
sourcery refactor

ervinpopescu:
Using `sourcery review --fix libqtile/`.

PR #4975, opened 2024-08-25
Apply automated sourcery refactoring (`sourcery review . --fix`)

Sigmanificient:
- inline return statements
- use `yield from` pattern
- replaced `.format` with f-string
- use `_` for loop variable when it is unused
- bound `except` to `Exception` instead of anything
- rewrite some block with list/set expression
- simplify some if conditions
- use `contextmanager.suppress` when skipping the exception

### DIFF
--- a/docs/screenshots/take_all.py
+++ b/docs/screenshots/take_all.py
@@ -414,9 +414,7 @@ def take(name, layout, spec):
             try:
                 client.run_layout_command(command)
             except Exception:
-                errors.append(
-                    "While running command {}:\n{}".format(command, traceback.format_exc())
-                )
+                errors.append(f"While running command {command}:\n{traceback.format_exc()}")
                 break
             time.sleep(0.05)
             screen.shoot()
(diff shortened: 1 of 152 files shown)


docs/screenshots/take_all.py
```diff
@@ -428,9 +426,7 @@ def take(name, layout, spec):
     except Exception:
         errors.append("While cleaning layout:\n" + traceback.format_exc())
 
-    if errors:
-        return False, "\n\n".join(errors)
-    return True, ""
+    return (False, "\n\n".join(errors)) if errors else (True, "")
```
jwijenbergh, line 429: don't like this one

elParaguayo: Yeah. That's nasty.

elParaguayo:
We looked at this before (#4334) and decided against it.

libqtile/backend/x11/core.py
```diff
@@ -333,13 +332,6 @@ def _xpoll(self) -> None:
                         self._motion_notify = None
                     self.handle_event(event)
 
-            # Catch some bad X exceptions. Since X is event based, race
```
jwijenbergh, line 336: it still removes comments :(

Sigmanificient: Oh that bad!

Sigmanificient:
> I can't see this getting merged.

Hi, this pr is still in work in progress, and have already been quite lighten/improved since this comment.
I hope you'll accept to give it another look when it will be ready

tych0:
The same is true of the last time we tried this: if we're going to merge it, it needs to be enforced in CI, otherwise we'll just diverge again. But I don't love patches like this to begin with since they make traversing git history painful, so I'd like to see some arguments for why listening to sourcery in addition to black is necessary.

PR #4981, opened 2024-08-29
Ruff

c0rydoras:


libqtile/utils.py
```diff
@@ -266,10 +266,10 @@ def lazify_imports(
     """
     __all__ = tuple(registry.keys())
 
-    def __dir__() -> tuple[str, ...]:
+    def __dir__() -> tuple[str, ...]:  # noqa: N807
```
elParaguayo, line 269: Why is ruff adding a `noqa` comment when we didn't need it before?

elParaguayo (commented):
Most of these changes are pretty benign (which is a good thing).

libqtile/utils.py
c0rydoras: also i added that one manually

libqtile/scripts/run_cmd.py
```diff
@@ -81,7 +81,7 @@ def add_subcommand(subparsers, parents):
         help="Do not break on match (keep applying rules).",
     )
     parser.add_argument("-g", "--group", help="Set the window group.")
-    parser.add_argument("cmd", help="Command to execute."),
+    (parser.add_argument("cmd", help="Command to execute."),)
```
c0rydoras, line 84: this should be
```suggestion
    parser.add_argument("cmd", help="Command to execute."
```
right?

libqtile/utils.py
c0rydoras: seems to be a mistake from ruff

https://github.com/PyCQA/pep8-naming/blob/f0edf20bd88cbd0950240ebeeea4fe6f4d90a094/CHANGELOG.rst?plain=1#L136-L137

elParaguayo: Yes - it's a valid use of those names.

c0rydoras: i did some digging and
- `pep8-naming`s N807 just ignores anything called `__dir__` and `__getattr__`
- `ruff`s N807 only ignores top level `__dir__` and `__getattr__`

tych0:
Can you squash the last commit with the lint commit? Other than that, looks good to me.

tych0:
I'm going to land this now since I didn't hear anyone's strong objection, and this way we can keep from having to rebase it as we land other things.

PR #4985, opened 2024-09-02
enable more rules (ruff)

c0rydoras:
- set `target-version` in `pyproject.toml` to `python >= 3.10` as thats the lowest we support (afaict)
- enabled [`flake8-pie`](https://docs.astral.sh/ruff/rules/#flake8-pie-pie)
- enabled [`pyupgrade`](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
  - applied available fixes
    - https://github.com/qtile/qtile/commit/be6489c478d691abecf14c23736b022536ec9a8b
    - https://github.com/qtile/qtile/commit/c07d8d972bc8e4724595344ef4ac7507803b56ce
  - fixed the remaining issues manually (https://github.com/qtile/qtile/commit/ecfba8356f8fe532a8b3cf3ab7e1d86ea0ddada6)
- formatted


c0rydoras:
Given this is somewhat similar to #4975

> The same is true of the last time we tried this: if we're going to merge it, it needs to be enforced in CI, otherwise we'll just diverge again. But I don't love patches like this to begin with since they make traversing git history painful, so I'd like to see some arguments for why listening to sourcery in addition to black is necessary.

(source: https://github.com/qtile/qtile/pull/4975#issuecomment-2308981108)

1. this would be enforced in CI by `ruff`
2. we could add these commits to a `.git-blame-ignore-revs` (https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) to make the git history a little better

tych0:
Woah, TIL about .git-blame-ignore-revs, that sounds great!

c0rydoras:
should I squash the commits?

elParaguayo:
Sorry, I may be being dumb here: how does `.git-blame-ignore-revs` work if the commit hash changes when the PR is merged?

tych0:
Yeah, I think we want it to pass bisect, so i would at the very least squash the fixes into the commit that enables them.

c0rydoras:
> Sorry, I may be being dumb here: how does `.git-blame-ignore-revs` work if the commit hash changes when the PR is merged?

Either merge them first, then do the `.git-blame-ignore-revs` in a separate PR, or idk maybe rebase merge doesn't change the commit hashes, haven't really thought about this

elParaguayo:
> > Sorry, I may be being dumb here: how does `.git-blame-ignore-revs` work if the commit hash changes when the PR is merged?
> 
> Either merge them first, then do the `.git-blame-ignore-revs` in a separate PR, or idk maybe rebase merge doesn't change the commit hashes, haven't really thought about this

I think the "rebase and merge" (which is what we typically do) changes the hash. Maybe the "Create a merge commit" preserves the hash. I'm not a git pro!

tych0:
I think it won't change the hashes if it's already on current git master, but it will if the branch is not.

c0rydoras:
merging this, then doing the `.git-blame-ignore-revs` in a seperate PR should work

libqtile/backend/wayland/wlrq.py
```diff
@@ -136,7 +137,7 @@ def paint(self, screen: Screen, image_path: str, mode: str | None = None) -> Non
         try:
             with open(image_path, "rb") as f:
                 image, _ = cairocffi.pixbuf.decode_to_image_surface(f.read())
-        except IOError:
+        except OSError:
```
tych0, line 140: I'm confused about both of these, seems like they're both wrong?
```
>>> open("/tmp/foo/bar")
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/foo/bar'
>>> isinstance(FileNotFoundError, IOError)
False
>>> isinstance(FileNotFoundError, OSError)
False
```

c0rydoras:
```py
>>> issubclass(FileNotFoundError, IOError)
True
>>> issubclass(FileNotFoundError, OSError)
True
```

c0rydoras: they're not wrong, but catching a `FileNotFoundError` would probably be better either way

tych0: Derp :), thank you! Not firing on all cylinders today apparently.

PR #4988, opened 2024-09-03
chore: add .git-blame-ignore-revs

c0rydoras:


elParaguayo:
Nice. Just need to update `MANIFEST.in` to exclude this file.

PR #4997, opened 2024-09-11
Completed test for all hooks

shyguyCreate:
**Inconveniences**
- [ ] In `test_client_mouse_enter` a possible problem may arise since the way to fire the hook in X11 and Wayland are different.
- [ ] In `test_client_urgent_hint_changed`  for Wayland the urgent hint is not detected.

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -319,7 +319,15 @@ class TestManager:
         if not success():
             raise AssertionError("Window could not be killed...")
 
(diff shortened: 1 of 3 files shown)


elParaguayo:
This looks good but can you tidy up the commit history please. Multiple commits with the same name isn't ideal. I'd probably put all the tests in one commit and keep the urgent hint separate.

PR #4998, opened 2024-09-11
[todo-cleanup]: bar.Bar.fake_button_press

ervinpopescu:
`bar.Bar.fake_button_press` had extra arguments `screen` and `position` which were unused. 

I also updated relevant tests.

### DIFF
--- a/docs/manual/howto/widget.rst
+++ b/docs/manual/howto/widget.rst
@@ -433,6 +433,8 @@ For example, the Mpris2 widget uses the following code:
 interfaces. Refer to the `dbus-next documentation <https://python-dbus-next.readthedocs.io/en/latest/>`_
 for more information on how to use the module.
 
+.. _mouse-events:
+
 Mouse events
 ============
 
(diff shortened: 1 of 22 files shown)


elParaguayo:
Amazing. That had been annoying me for a while but I never got round to changing it.

ervinpopescu:
> Amazing. That had been annoying me for a while but I never got round to changing it.

Heh. I still need to update docs to point to the [mouse events](https://docs.qtile.org/en/stable/manual/howto/widget.html#mouse-events) 😁

ervinpopescu:
Whelp, let's stop there. The rest of the TODO's are either too niche or on `wayland` territory.

elParaguayo:
Oh. Do you think people are using this code in the wild? I'd assumed it's only in tests.

ervinpopescu:
> Oh. Do you think people are using this code in the wild? I'd assumed it's only in tests.

Same. TBH, I would have done it myself, if I knew that was the case. 

@elParaguayo I can write one possibly today, most likely tomorrow, if you wanna do it, go ahead. It'd be my first migration so it's gonna take some time to look at the existing examples.

PR #5004, opened 2024-09-21
Nix: Run make update-flake

jwijenbergh:


jwijenbergh:
this test failure is odd @elParaguayo. Does it have to do with the image decoding changes?

elParaguayo:
Looks like we've lost svg support. That's not due to the image decoding change as svg files would have been handled the same way. I think, in the past, we've told users to make sure they've got `librsvg` installed. So you may need to add that (or whatever it's called in ubuntu) to our build environment.

tych0:
No luck with librsvg2 :(

elParaguayo:
It would have been weird to suddenly need a new dependency.

Has something happened to the SVG files?

tych0:
I agree... and nothing has changed in tree. :ghost:

jwijenbergh:
Maybe this flake.lock is cursed even though the CI doesn't use it 😂

elParaguayo:
This is failing on other PRs so it's not related to this PR.

PR #5005, opened 2024-09-22
Pre-commit: Add pyright

jwijenbergh:
This patch adds pyright to pre-commit. I am testing this as I find that pyright catches *much* more errors than mypy. Maybe useful to have.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,7 @@ repos:
     rev: v2.10
     hooks:
       - id: vulture
+  - repo: https://github.com/RobertCraigie/pyright-python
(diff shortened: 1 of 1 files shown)


jwijenbergh:
WIP here, it also reports missing imports which are incorrect

tych0:
Do you happen to know if it's a superset, and we could maybe get rid of mypy?

jwijenbergh:
> Do you happen to know if it's a superset, and we could maybe get rid of mypy?

That's a good question. Skimming this https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md it seems like it is a superset. But I'm not confident to say it definitely is

PR #5008, opened 2024-09-25
Fix TreeTab panel bug

### LINKED ISSUE #4979: TreeTab Layout, using `increase_ratio` results in weird window names
### Issue description

I tried to `increase_ratio` for the `TreeTab` layout, it does work but the window names dont scale (it does weird things)

image
After calling `increase_ratio` a couple times:
image

I expected it to not look this cursed.

How to reproduce:

1. Add tree tab layout to configuration
2. Add keybinding for `lazy.layout.increase_ratio()`  
3. Reload config
4. Open window in tree tab layout
6. Use Keybinding for increasing ratio

observe issue

### Version

0.28.2.dev0+gf1ed49bc.d20240813

### Backend

X11 (default)

### Config

```python
-
```

### Logs

```text
-
```

### Required

elParaguayo:
The TreeTab layout does not create a new Drawer object when the panel ratio is changed. This results in rendering issues.

Fixes #4979

### DIFF
--- a/libqtile/layout/tree.py
+++ b/libqtile/layout/tree.py
@@ -735,7 +735,9 @@ class TreeTab(Layout):
     def _create_drawer(self, screen_rect):
         # Create a new drawer object if the screen is a different height
         # e.g. if moving between screens with different resolutions
-        if self._drawer is not None and self._drawer.height != screen_rect.height:
+        if self._drawer is not None and (
+            self._drawer.height != screen_rect.height or self.panel_width != self._drawer.width
+        ):
             self._drawer.finalize()
             self._drawer = None
 


tych0:
Arg, patch looks fine to me. We really need to get to the bottom of the CI failure. I tried ltracing locally, but I didn't see any cairo things actually invoked, which I wonder if it's an artifact of how we use cffi. Anyway, I would like to see what's happening in the .so somehow...

tych0:
I think it would also be fine to disable the test_images* tests until we figure out what's going on, and ban all changes to the images code until CI is resurrected.

elParaguayo:
> I think it would also be fine to disable the test_images* tests until we figure out what's going on, and ban all changes to the images code until CI is resurrected.

That works for me.

PR #5010, opened 2024-09-27
Fix image test failures

elParaguayo:
For some reason, CI now needs gdkpixbuf and svg libraries for our image tests to pass.


jwijenbergh:
Very quick, thanks

PR #5017, opened 2024-10-02
Hook 'client_mouse_enter' consistent in wayland and x11

shyguyCreate:
Hook `client_mouse_enter` now is fired regardless of the focus window.


tych0:
CI failure looks unrelated:
```
sha256sum: release/coveralls-linux.tar.gz: No such file or directory
30
release/coveralls-linux.tar.gz: FAILED open or read
31
```

i will give it a bit and retry. patch itself looks fine to me, I'll leave it here in case @elParaguayo has any objections.

tych0 (approved):

PR #5019, opened 2024-10-04
wl: keysym -> keycode

### LINKED ISSUE #4983: default prompt can't type capital letter
### Issue description

I use default prompt. I can't enter capital letter or column, or anything need shift key. On xorg, it works ok. \
It output error when I launch the prompt, and any subsequences input. I use `tail -f .local/share/qtile/qtile.log` to check.

### Version

0.28.2.dev0+gf1ed49bc.d20240813

### Backend

Wayland (experimental)

### Config

[My config on github](https://github.com/Jeff-WuYo/Dot_file/blob/main/dot_config/qtile/config.py)

### Logs

```text
2024-08-30 15:22:53,705 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
  File "/usr/lib/python3.12/site-packages/pywayland/server/listener.py", line 40, in notify_func
    callback(listener, data)
  File "/usr/lib/python3.12/site-packages/libqtile/backend/wayland/inputs.py", line 311, in _on_key
    self.core.focused_internal.process_key_press(keysym)
                                                 ^^^^^^
UnboundLocalError: cannot access local variable 'keysym' where it is not associated with a value
2024-08-30 15:22:53,759 ERROR pywayland.server listener.py:notify_func():L42 Exception in callback function
  File "/usr/lib/python3.12/site-packages/pywayland/server/listener.py", line 40, in notify_func
    callback(listener, data)
  File "/usr/lib/python3.12/site-packages/libqtile/backend/wayland/inputs.py", line 311, in _on_key
    self.core.focused_internal.process_key_press(keysym)
                                                 ^^^^^^
UnboundLocalError: cannot access local variable 'keysym' where it is not associated with a value
```

### Required

tych0:
I don't really know this code that well, but the name of the argument to process_key_press() is keycode, so it stands to reason that this is just a typo, instead of a more serious logic bug...

Fixes #4983

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -307,7 +307,7 @@ class Keyboard(_Device):
             if handled:
                 return
             if self.core.focused_internal:
-                self.core.focused_internal.process_key_press(keysym)
+                self.core.focused_internal.process_key_press(keycode)
                 return
 
         self.seat.keyboard_notify_key(event)


elParaguayo:
What's prompted this? I'm not sure this change is correct.

tych0:
just looking at #4983 and a wild guess based on the function parameter's name, happy to have a real fix if this isn't it :)

elParaguayo:
I would think that change would break a few things but haven't testedit.

PR #5020, opened 2024-10-04
Screen change notification fix

tych0:
Here is another attempt, which works for me in testing, to fix our ugly hacks about screen change notifications and stack traces and such. We'll need people from both of the mentioned bugs from the commit messages to test, I will ping them shortly.

### DIFF
--- a/test/test_scratchpad.py
+++ b/test/test_scratchpad.py
@@ -37,7 +37,7 @@ def spawn_cmd(title):
     return cmd
 
 
-class ScratchPadBaseConfic(Config):
+class ScratchPadBaseConfig(Config):
     auto_fullscreen = True
     screens = []
     groups = [
@@ -73,7 +73,7 @@ class ScratchPadBaseConfic(Config):
     mouse = []
 
 
-scratchpad_config = pytest.mark.parametrize("manager", [ScratchPadBaseConfic], indirect=True)
+scratchpad_config = pytest.mark.parametrize("manager", [ScratchPadBaseConfig], indirect=True)
 
 
 @Retry(ignore_exceptions=(KeyError,))
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -185,10 +185,7 @@ class Core(base.Core):
         self.conn.finalize()
 
     def get_screen_info(self) -> list[ScreenRect]:
-        ps = self.conn.pseudoscreens
-        if self.qtile:
-            self._xpoll()
-        return ps
+        return self.conn.pseudoscreens
 
     @property
     def wmname(self):
@@ -277,6 +274,14 @@ class Core(base.Core):
 
             self.update_client_lists()
             win.change_layer()
+
+        self.conn.enable_screen_change_notifications()
+        # previous to the enable_screen_change_notifications() refactoring,
+        # we triggered a screen change notification on every boot,
+        # regardless of whether the screen had actually changed. so, we do
+        # that here, since we had tests that enforced that behavior so
+        # maybe someone depended on it.
+        hook.fire("screen_change", None)
 
     def warp_pointer(self, x, y):
         self._root.warp_pointer(x, y)
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -402,7 +402,6 @@ class Xinerama:
 class RandR:
     def __init__(self, conn):
         self.ext = conn.conn(xcffib.randr.key)
-        self.ext.SelectInput(conn.default_screen.root.wid, xcffib.randr.NotifyMask.ScreenChange)
 
     def query_crtcs(self, root):
         infos = []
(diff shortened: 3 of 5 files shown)


tych0:
arg, I vaguely remember this scratchpad test failing before, but i forget why. cc @elParaguayo I think you fixed it before? But I don't see that in the git logs anywhere...

elParaguayo:
It's something to do with the screens being reconfigured. If you set `reconfigure_screens = False` in `ScratchPadBaseConfig` then the test passes. 

The `on_focus_change` method is not being called which is why the scratchpad still appears in the list of windows. 

I'll keep investigating.

tych0:
Ah, ok. So maybe we're not reconfiguring the screens at all, that would manifest this?

tych0:
Because I suspect we aren't any more: I have disabled randr, so that wouldn't trigger at all.

elParaguayo:
The `client_focus`  hook is not being fired when a new window opens in that test.

tych0:
Hmm, very strange. Ok, ty.

test/test_scratchpad.py
```diff
@@ -71,6 +71,7 @@ class ScratchPadBaseConfig(Config):
     floating_layout = libqtile.resources.default_config.floating_layout
     keys = []
     mouse = []
+    reconfigure_screens = False
```
elParaguayo, line 74: While this makes the test pass, this can't be the solution. However, I just don't understand why this PR stops the `client_focus` hook from firing here.

Bizarrely, `test/test_hook.py::test_client_focus` passes, even when `reconfigure_screens = True`, so there's something else going on too.

tych0: Yeah, absolutely agreed. Sorry, I didn't intend to merge this, I was just experimenting. I think I do understand what is going on here, but I couldn't manage to "fix" it with sleeps, so I'm not completely convinced.

We are getting a ScreenChangeNotify event (not the fake one I added, but a real one from the X11 server) during *every* test now with this change. I found this doc: https://www.x.org/archive/X11R6.8.1/doc/Xrandr.3.html#toc4 which says:

> Additionally, to eliminate a potential race condition, this event may be generated immediately upon selecting for notification if the screen has changed since the client of Xrandr connected to the X server, to enable reliable screen resolution changing when a user may log in and change the configuration while one or many clients are starting up.

So my theory is: we are unmasking ScreenChangeNotify much later than previously, right before we hand off control to the test suite. So this means that the test suite sees this ScreenChangeNotify, which it previously did not. Most tests must be agnostic to this, but it seems this particular Scratchpad one is not? If that's the case, you'd think you could put a `time.sleep(5)` at the beginning of the failing test and get it to pass, but that didn't work for me. I tried flushing the X11 connection and a few other things, no luck.

Another option I had considered was that something about the scratchpad code was generating this ScreenChangeNotify, but I didn't see anything suspicious there. Plus, given that it happens on *other* tests as well (e.g. I tried with the dgroups ones and see the same behavior), I suspect it's unrelated to the actual scratchpad test.

The final piece of weirdness in all of this: I explicitly disable Xephyr's randr here, so in principle we shouldn't be getting any ScreenChangeNotify events at all, because the randr extension shouldn't exist. But I think Xephyr's implementation of randr is pretty poor based on my previous experimentation, so I haven't read a lot into this.

We could try and switch all of the testing to the x11's "dummy" driver instead of these xvfb/xephyr hacks, which would enable us to actually test my randr stuff in this branch. I'm still trying to validate my theory, hopefully I can get there by adding some extra synchronization points to the tests.

elParaguayo: So, is the scratchpad test failing because it's the only test that has `reconfigure_screens = True`? I find it a bit concerning that that apparently stops the hook being fired.

tych0: > the hook being fired.

Here you meant the `client_new` hook, not `screen_changed`? I agree it is weird and shouldn't do that, there is more weirdness going on on our side. But I suspect that weirdness has always been there, and it is only the moving of the ScreenChangeNotify that is triggering it.

tych0: I am still looking at this, and have no real leads.

I did get the xtrace stuff working, though, which has been helpful to understand what's going on. Seems like maybe I should pull that and the other spelling fix out to a separate branch to land them in case we need this in the future.

tych0:
Ok, I'm finally back looking at this because I really want to land it, and I tracked down our bug. The problematic thing that causes the focus hook to not be fired is this: https://github.com/qtile/qtile/blob/master/libqtile/group.py#L167-L168 coupled with the code from `_process_screens()` which creates new screens and deletes the old ones: https://github.com/qtile/qtile/blob/master/libqtile/core/manager.py#L384-L437

The problem is that since we get a reconfigure event, that focus check fails because the screen objects are not equal, even though they do in fact represent the same screen. So qtile thinks the current screen doesn't have focus, and it doesn't trigger the focus hook.

This is actually a bug in upstream as well, but probably most people don't notice because they don't reconfigure screens very often. I can think of two ways to solve this:

1. don't re-create Screen() objects, and mutate the existing ones. this seems the cleanest, but I don't know what other implications that might have.
2. define an `__eq__` for `Screen`s such that objects that represent the same offset and geometry are considered equal. That will fix our immediate problem, and leave everything the way it is, but might be ugly. (Eventually we could do this via serial number, which might be cleaner...)

elParaguayo:
Nice work. I like option 2.

PR #5023, opened 2024-10-06
chore: explicitly re-export layouts using `__all__`

WieeRd:
I use Pyright to lint my Qtile config, and these diagnostics kept appearing on each layout.

image

After some research, it turns out this is the intended behavior of Pyright.

- Related issue: microsoft/pyright#2639
- Recommended way to define library interface: <https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface>

**TL;DR** - By default, an imported symbol is considered private (not re-exported).
If the intent is to re-export the imported symbol, it needs to be either:

- Be included in an `__all__` list
- Use a redundant aliased import form `from .foo import Bar as Bar`.

I thought the former would look cleaner and added `__all__` to explicitly and publically re-export layouts from submodules.

### DIFF
--- a/libqtile/layout/__init__.py
+++ b/libqtile/layout/__init__.py
@@ -24,6 +24,27 @@
 # is annoying, so we ignore libqtile/layout/__init__.py completely
 # flake8: noqa
 
+__all__ = [
+    "Bsp",
+    "Columns",
+    "Floating",
+    "Matrix",
+    "Max",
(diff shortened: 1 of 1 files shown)


ervinpopescu:
Nice, been getting these as well with both pyright and ruff, good to see someone took the time to fix this!

Thanks @WieeRd!

PR #5024, opened 2024-10-07
Add urgency to xwayland windows

shyguyCreate:
This PR started as a way to fire the `client_urgent_hint_changed` hook from wayland, and I had trouble with the script `test/scripts/window.py` because GTK remove urgency from the wayland backend so I had to force the Xwayland window.

Now I have accomplished that but the hook is fired in different occasions that do not involve urgency. So what kinds restrictions are needed [here](https://github.com/shyguyCreate/qtile/blob/a8ddd6aecb7f0b1d6f7889ecf52ddc35d5e64f1b/libqtile/backend/wayland/xwindow.py#L172) to make the hook work properly?


shyguyCreate:
I have been digging and I think it has to do with this C function `xcb_icccm_wm_hints_get_urgency` but I don't now if there is an equivalent for python. (btw using [sway](https://github.com/swaywm/sway/blob/c90cb37b2a0861548461daa9b75d75317e01b679/sway/desktop/xwayland.c#L746) as a reference)

shyguyCreate:
Correct me if I am wrong but this may not be possible to fully implement since `pywayland` does not support the hints necessary to specifically determine urgency.

What I mean by that is that in the `wlroots/xwayland.py` [file](https://github.com/flacjacket/pywlroots/blob/main/wlroots/xwayland.py) there should be a class that handles a pointer to the required hints like this `self._ptr = ffi.cast("xcb_icccm_wm_hints_t *", ptr)` but there isn't.

tych0:
I think that is right, but it should be possible to add that code to pywlroots, and when we upgrade we can land this.

tych0:
Does it fall back in a reasonable way if the functionality in pywlroots isn't there? If so, and you have a PR landed for the next version of pywlroots, that seems reasonable to mark it as ready.

RFCreate:
I just open a [PR in the pywlroots repo](https://github.com/flacjacket/pywlroots/pull/203).

Once it is merged this should be ready with the following changes:

```diff
diff --git a/libqtile/backend/wayland/xwindow.py b/libqtile/backend/wayland/xwindow.py
index ea230976..f2cae91f 100644
--- a/libqtile/backend/wayland/xwindow.py
+++ b/libqtile/backend/wayland/xwindow.py
@@ -30,6 +30,7 @@ from libqtile import hook
 from libqtile.backend import base
 from libqtile.backend.base import FloatStates
 from libqtile.backend.wayland.window import Static, Window
+from libqtile.backend.x11.window import UrgencyHint
 from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
 
@@ -170,9 +171,10 @@ class XWindow(Window[xwayland.Surface]):
             self.ftm_handle.set_app_id(self._wm_class or "")
 
     def _on_set_hints(self, _listener: Listener, _data: Any) -> None:
-        if self.qtile.current_window != self:
-            self._urgent = True
-            hook.fire("client_urgent_hint_changed", self)
+        if UrgencyHint & self.surface.hints.flags:
+            if self.qtile.current_window != self:
+                self._urgent = True
+                hook.fire("client_urgent_hint_changed", self)
 
     def hide(self) -> None:
         super().hide()
```

shyguyCreate:
It is expected to fail, waiting for activity in the [PR](https://github.com/flacjacket/pywlroots/pull/203).

RFCreate:
https://github.com/flacjacket/pywlroots/pull/203 is now merged, but a new release would be needed.

RFCreate:
Same old.

RFCreate:
Hoping the day will come.

RFCreate:
Close in favor of #5602.

PR #5032, opened 2024-10-14
Accept modifiers from distinct keyboards under the wayland environment

RollMan:
This commit tries to apply modifiers from a keyboard to not only the keyboard itself but also the other keyboards under the wayland backend environment.

Background: I use two keyboards at the same time as a primitive manner of split keyboards. That really helps with my posture, and relaxes my shoulders, chest, etc. I noticed that the current master (d13eb3d851fec52d6dc84a5de2de41d2f139d283) as a wayland compositor does not accept key bindings with combination of the super key and any key from distinct keyboards. For example, Super (keyboard 1) + 'w' (keyboard 2) results in only propagation of 'w' key-press into a focused window (showing 'w' if the window is of a text editor) while I expect close of the window with the default key bindings. 

I think this is a natural behavior of WMs. Key bindings outside qtile (e.g. Ctrl-C) from distinct keyboards work well with qtile ( wayland), and additionally, tty, X and Windows to the best of my knowledge.

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -21,6 +21,8 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from functools import reduce
+from operator import or_
 from typing import TYPE_CHECKING
 
 from pywayland.protocol.wayland import WlKeyboard
@@ -279,7 +281,7 @@ class Keyboard(_Device):
                 xkb_keysym,
             )
             keysyms = [xkb_keysym[0][i] for i in range(nsyms)]
-            mods = self.keyboard.modifier
+            mods = reduce(or_, [k.keyboard.modifier for k in self.core.keyboards])
             handled = False
             should_repeat = False
             if event.state == KEY_PRESSED:


libqtile/backend/wayland/inputs.py
```diff
@@ -279,7 +281,7 @@ def _on_key(self, _listener: Listener, event: KeyboardKeyEvent) -> None:
                 xkb_keysym,
             )
             keysyms = [xkb_keysym[0][i] for i in range(nsyms)]
-            mods = self.keyboard.modifier
+            mods = reduce(or_, [k.keyboard.modifier for k in self.core.keyboards])
```
tych0, line 284: Can you write this as `any()`?

RollMan: I am not very sure how the later program treats `mods`, but I don't think `any()` works; [`wlroots.wlr_types.keyboard.Keyboard.modifier`](https://github.com/flacjacket/pywlroots/blob/ebafb5aeb65821134980b9e939d406bfeb222f55/wlroots/wlr_types/keyboard.py#L148) returns [`KeyboardModifier`](https://github.com/flacjacket/pywlroots/blob/ebafb5aeb65821134980b9e939d406bfeb222f55/wlroots/wlr_types/keyboard.py#L25), which is a [bitwise flag](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/3b3ed21e61b871abdad83f0ee769effdb7d43e56/include/wlr/types/wlr_keyboard.h#L30), not a boolean. `mods` must retain its bitwise representation to distinguish which modifier key is pressed and that's why I use bitwise-or `operator.or_`, not logical-or (`or`) which is compatible with `any()`.

tych0:
I'll leave this up for a few days for the wayland people to take a look, but seems reasonable to me.

jwijenbergh:
I will try to review this on thursday or so

PR #5036, opened 2024-10-15
Change the focus to the window under the mouse when click or drag

shyguyCreate:
This is useful when in the config you have `follow_mouse_focus = False` and would like to use the window under the mouse when clicking or dragging a window.

P.S. Maybe this should be a global option in the config but I am not sure. If that is the case what name would be good? Maybe `follow_mouse_click_or_drag` or maybe better a new value like `follow_mouse_focus = "click_or_drag_only"`?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
     * features
         - Add `SwayNC` widget to interact with Sway Notification Centre (wayland only)
         - Add `swap` method to Plasma layout
+        - New `click_or_drag_only` option for follow_mouse_focus to change the focus to the window under the mouse when click or drag
     * bugfixes
         - Make MonadWide layout up/down focus navigation behave like MonadTall's left/right
 
--- a/docs/manual/config/index.rst
+++ b/docs/manual/config/index.rst
@@ -135,7 +135,9 @@ configuration variables that control specific aspects of Qtile's behavior:
     * - ``follow_mouse_focus``
       - ``True``
       - Controls whether or not focus follows the mouse around as it moves
-        across windows in a layout.
+        across windows in a layout. Otherwise set this to ``"click_or_drag_only"``
+        to change focus only when doing a :class:`~libqtile.config.Click` or
+        :class:`~libqtile.config.Drag` action.
     * - ``widget_defaults``
       - ``dict(font='sans', fontsize=12, padding=3)``
       - Default settings for bar widgets.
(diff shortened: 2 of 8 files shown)


shyguyCreate:
I have decided to go for the `follow_mouse_focus = "click_or_drag_only"` since IMO is what best represents what I am trying to accomplish.

elParaguayo:
> I have decided to go for the `follow_mouse_focus = "click_or_drag_only"` since IMO is what best represents what I am trying to accomplish.

Agree it makes sense to use that variable. However, what happens if you want the behaviour of `True` and `click_or_drag_only`? Are they mutually exclusive?

shyguyCreate:
`True` is always changing focus based on the mouse so in a click or drag action you don't need to change focus again to the hovered window because it is already in focus.

elParaguayo:
This looks ok to me. Can you rebase to clear the conflict?

nazar-pc:
The description of the option is super confusing to me, I don't understand what this does. IIUC before it was True/False, which made sense, but how is `click_or_drag_only` different from them?

nazar-pc:
Didn't `False` focus a window on clicking or dragging already?

saviola777:
You are half-right, clicked windows are focussed even with `follow_mouse_focus = False`, dragging however drags the currently focussed window no matter where the mouse cursor is, while it first focusses and then drags the window under the cursor with `click_or_drag_only`.

So I think the change

```python
            if isinstance(m, Click):
                if self.config.follow_mouse_focus == "click_or_drag_only":
                    self._focus_hovered_window()
```

does not have any effect.

nazar-pc:
I see, that makes more sense now, thanks for explaining!
Now I'm wondering why would anyone want `follow_mouse_focus = False` then, I'd expect `False` to do what this PR does instead of having a third option :thinking:

saviola777:
I agree! I guess there's a subtle point about dragging here. I think it is fair to assume that if I have a binding like `Drag([mod], "Button1", lazy.window.set_position_floating(), start=lazy.window.get_position()),` that it would affect the currently focussed window. This is the case for `False`, while `click_or_drag_only` changes the focus before dragging. Essentially `click_or_drag_only` saves you one click when wanting to drag the window under the cursor!

RFCreate:
> You are half-right, clicked windows are focussed even with `follow_mouse_focus = False` ...

There is a misunderstanding between a normal click (which yes focuses the window) and a `Click` which can perform a lazy action the same way `Drag` does but when set to `False` will not set the focus to the window under the mouse.

hanschen:
This solves one of my minor nitpicks with Qtile, so many thanks for that!

I had the same thought as @nazar-pc above - is there any scenario where one would prefer the behavior of `follow_mouse_focus = False` over `click_or_drag_only`? I think it would be nice if this was the default behavior with `follow_mouse_focus = False`, and to get rid of the `click_or_drag_only` option. This would both improve discoverability of the feature and simplify the documentation.

Issue #5034, opened 2024-10-15
Newly opened windows appear to be floating until group has been switched back and forth


tcld:
### Issue description

I'm having some issues with the most recent Qtile-master, in particular [this PR](https://github.com/qtile/qtile/pull/4426) seems to be where those issues originate (just did a bisect):

When I open a new terminal (terminator) or browser (firefox), the windows open up unmaximized at the center of the screen (I've tested Bsp- and Treetab-layouts), almost as if they were floating windows. They take up about a quarter of the space available on the screen for me.
After switching to another group and back the windows are displayed the way the layouts should display them. Any new windows suffer from the same issue, though.

I haven't found anything interesting in the Qtile-log.

@elParaguayo fyi

### Version

dev 284e6dc16be6f8a9ba6e94b597f498078371c397

### Backend

X11 (default)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

elParaguayo:
Very odd. I can't recreate this. Windows open normally for me.

jwijenbergh:
Did you try the default config?

elParaguayo:
Yes

ervinpopescu:
Huh, I get the same issue with my config but not with the default one. Interesting. I only tested default before...

elParaguayo:
Hmm. If multiple users are saying this then I'm inclined to revert for now.

ervinpopescu:
If it's config generated, I would say revert is not needed. @tcld can you please confirm if default config generates this?

tcld:
I'm using my setup for work right now, and I don't know how to quickly test this (and restart from the default config afterwards) without disrupting things too much. Maybe I can try in about four or five hours.

jwijenbergh:
Do you have the floating layout in your layouts?

tcld:
No, I don't.

tcld:
I checked the default configuration, specifically at https://github.com/qtile/qtile/commit/284e6dc16be6f8a9ba6e94b597f498078371c397 . I see the issue there as well.

For completeness, I modified the default config slightly by adding these keys (otherwise I just copied the "default_config.py" into a fresh ".config/qtile"-folder and renamed it to "config.py"):
```python
    Key([mod, "shift", "control"], "r", lazy.restart(), desc="Restart Qtile."),
[...]

PR #5042, opened 2024-10-19
Revert "maximize windows that wanted to maximize"

elParaguayo:
This reverts commit 284e6dc16be6f8a9ba6e94b597f498078371c397.

It's causing a few issues so reverting for now while we investigate these (#5034 and #5043).


MoonCubes:
for me this only happened, when newly opened windows wanted to instantly maximize, so maybe we can prevent that but still allow it after that?

elParaguayo:
> for me this only happened, when newly opened windows wanted to instantly maximize, so maybe we can prevent that but still allow it after that?

Ok. That may be the same issue as the other one then.

tych0:
Thanks, I'll start on a release now.

PR #5044, opened 2024-10-19
doc: update release process

tych0:
I haven't been sending out e-mails or notifying IRC for the last few releases, and nobody has complained. Let's drop this, since it's one extra step.

I have, however, been updating the website, and I had forgotten to mention that initially, so note that down.

### DIFF
--- a/docs/manual/releasing.rst
+++ b/docs/manual/releasing.rst
@@ -39,9 +39,10 @@ Be sure that you GPG-sign (i.e. the ``-S`` argument to ``git commit``) this comm
 6. Make sure all of these actions complete as green. The release should show up
    in a few minutes after completion here: https://pypi.org/project/qtile/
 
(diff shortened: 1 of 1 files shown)


ervinpopescu:
QQ: is releasing meant to not be shown somewhere in the TOC (or whatever the thing on the left is called by RTD/you guys)?

tych0:
Not secret at all, I just didn't put it in the TOC because I didn't want the noise for all the people who don't release. I'm happy to add it if you think it's useful, I just figured it wouldn't be.

elParaguayo:
While you're doing this, can you fix line 17. I think that `-S` should be lowercase (as per the example in the following line).

tych0:
No, it really is `-S`, `-s` is "add the Signed-off-by for the author's e-mail" flag. Hooray for consistency.

PR #5045, opened 2024-10-19
Add ability to hide crashed widgets from the bar

elParaguayo:
Currently, if a widget crashes when it's configured, it's replaced by an instance of a `ConfigErrorWidget`. This PR adds the ability to allow widgets to be removed from the bar silently if they crash (although error messages will still be in the log).

This was requested in #5040 to particularly address the limitation of `Systray` which only allows one instance at any time. Configs using the same widget list initialiser across multiple screens would get a Config Error unless they removed the Systray widget first.

Closes #5040


tych0 (approved):

elParaguayo:
Thanks for merging. Just noticed I didn't change the name in the copyright I copied from another file ;)

Don't think he'll mind!

PR #5046, opened 2024-10-19
Add SwayNC widget

elParaguayo:
Adds a basic widget for interacting with the Sway Notification Centre.

Draft for now as we get comments/feedback from users.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Qtile x.xx.x, released xxxx-xx-xx:
     * features
+        - Add `SwayNC` widget to interact with Sway Notification Centre (wayland only)
     * bugfixes
 
 Qtile 0.29.0, released 2024-10-19:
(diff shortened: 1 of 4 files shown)


tych0:
Code looks reasonable to me, not that I know anything about sway's NC :). How is this different than libnotify?

elParaguayo:
> Code looks reasonable to me, not that I know anything about sway's NC :). How is this different than libnotify?

This does not display the notifications itself. It provides a counter showing how many notifications there are. `swaync` provides the notification server but also provides a side panel with media info etc. This widget can also toggle displaying that panel.

tych0:
I see, ok thanks!

arjan-s (approved):
This works exactly as I hoped! Thank you so much!

elParaguayo:
> This works exactly as I hoped! Thank you so much!

Thanks. I'll tidy up the commits, fix the test failure and get this merged.

PR #5047, opened 2024-10-19
Add window swapping to Plasma layout

MoonCubes:
This commit adds the 'swap' method to the Plasma layout. It works by checking all the leafs of the root Node and swapping the two windows.

**Note**: this is still a little buggy and I'm trying to implement some tests for this.

### DIFF
--- a/libqtile/layout/plasma.py
+++ b/libqtile/layout/plasma.py
@@ -25,6 +25,7 @@ from math import isclose
 from typing import NamedTuple
 
 from libqtile import hook
+from libqtile.backend.base import Window
 from libqtile.command.base import expose_command
 from libqtile.hook import Hook, qtile_hooks
 from libqtile.layout.base import Layout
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for this.

What bugs are you running into?

MoonCubes:
The windows only sometimes swap. and I have to try multiple times.
https://github.com/user-attachments/assets/1c65cd19-e1ec-408e-8976-fc5a091e0099

elParaguayo:
But is that a problem with your code or something else? Is your method being called?

MoonCubes:
My method is being called, but I haven't fully tested yet if it has a bug, or the bug is Qtile related

libqtile/layout/plasma.py
```diff
@@ -836,6 +837,18 @@ def __init__(self, **config):
...
+        node_c1 = node_c2 = None
+        for leaf in self.root.all_leafs:
+            if leaf.payload is not None:
+                if c1 == leaf.payload:
+                    node_c1 = leaf
+                elif c2 == leaf.payload:
+                    node_c2 = leaf
+            if node_c1 is not None and node_c2 is not None:
+                node_c1.payload, node_c2.payload = node_c2.payload, node_c1.payload
+                return
```
elParaguayo, line 850:
```suggestion
                node_c1.payload, node_c2.payload = node_c2.payload, node_c1.payload
                self.group.layout_all()
                self.group.focus(c1)
                return
```
This should fix it for you. The windows were swapping but you weren't redrawing the layout.

MoonCubes: That works, thanks. Can this be merged now?

MoonCubes:
sory, I am apparently too stupid for git. The change has been added.

PR #5050, opened 2024-10-20
Make MonadWide up/down focus navigation behave like MonadTall left/right

holocronweaver:
MonadTall exposes left/right commands to focus the closest window to the left/right, allowing easy navigation between columns. MonadWide had no similar override for up/down focus commands, instead it inherited the next/previous = down/up overrides of MonadTall, making directional navigation between rows more cumbersome by requiring cycling amongst all windows. The result is inconsistent window navigation between MonadTall and MonadWide.

This change retains the original navigation commands and behavior of MonadTall (including left/right), while adding a symmetric up/down command to MonadWide.

---

Besides manually testing the Qtile directional focus navigation behavior for both MonadTall and Wide layouts, I added unit tests, and have verified the full test suite passes locally (except for one related to migration that seems unrelated to my changes). I also verified that my unit tests for MonadTall pass without my layout code changes (implying no behavior change), while those for MonadWide fail (implying I fixed the inconsistency).

### DIFF
--- a/libqtile/layout/xmonad.py
+++ b/libqtile/layout/xmonad.py
@@ -756,11 +756,11 @@ class MonadTall(_SimpleLayoutBase):
         # shrink client by total change
         self.relative_sizes[self.focused - 1] -= self._get_relative_size_from_absolute(change)
 
-    @expose_command("down")
+    @expose_command()
     def next(self) -> None:
         _SimpleLayoutBase.next(self)
 
-    @expose_command("up")
+    @expose_command()
     def previous(self) -> None:
         _SimpleLayoutBase.previous(self)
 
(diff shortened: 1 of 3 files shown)


libqtile/layout/xmonad.py
```diff
@@ -858,6 +858,16 @@ def right(self):
...
+    @expose_command()
+    def up(self):
+        """Focus on the closest window above the current window"""
+        self.previous()
+
+    @expose_command()
+    def down(self):
+        """Focus on the closest window below the current window"""
+        self.next()
+
```
elParaguayo, line 870: You're probably correct, but why is this needed. The previous decorators should have had the same effect as this. Does it cause issues for `MonadWide` on inheritance?

holocronweaver: Exactly, the decorators caused issues with `MonadWide` on inheritance. I tried overriding in `MonadWide` and leaving the decorators as is, but the overrides were ignored and the decorators in `MonadTall` took precedence (confirmed by adding some log lines). Avoiding decorators allowed inheritance and overrides to work as usual.

PR #5055, opened 2024-10-24
Add key map to gen-img to fix #5053

### LINKED ISSUE #5053: gen-keybind-img doesn't register all keys
### Issue description

gen-keybind-img fails on keys that are registered, at least for me with ezkey, with `<keyname>`. Such as `<grave>`, `<slash>`, etc

Potential solution (and what I did with my copy of script) is to make a key name map for KInfo. It does properly work after that

```py
class KInfo:
    NAME_MAP = {
        "togroup": "to group",
        "toscreen": "to screen",
    }

    KEY_MAP = {
        "grave": "`",
        "semicolon": ";",
        "slash": "/",
        "backslash": "\\",
        "comma": ",",
        "period": ".",
        "bracketleft": "[",
        "bracketright": "]",
        "quote": "'",
        "minus": "-",
        "equals": "=",
    }

    def __init__(self, key):
        if key.key in self.KEY_MAP:
            self.key = self.KEY_MAP[key.key]
        else:
            self.key = key.key
        self.command = self.get_command(key)
        self.scope = self.get_scope(key)
```

Another issue with this script is lack of escape key and function row

### Version

latest from github

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

al1-ce:


tych0:
Thanks!

PR #5060, opened 2024-10-26
fix(_get_group): group not in list when reloading

shyguyCreate:
I have this hook enabled in my config.

```
@hook.subscribe.group_window_remove
def group_window_remove(group, window):
    if not group.windows and group.screen:
        if group.screen.previous_group and group.screen.previous_group.windows:
            group.toscreen(toggle=True)
        else:
            group.screen.next_group(skip_empty=True, skip_managed=False)
```

That started throwing errors when I reload the config having a window in the current group.

```
ERROR libqtile hook.py:fire():L196 Error in hook group_window_remove
  File "/usr/lib/python3.12/site-packages/libqtile/hook.py", line 194, in fire
    i(*args, **kwargs)
  File "/home/shyguy/.config/qtile/config.py", line 342, in group_window_remove
    group.screen.next_group(skip_empty=True, skip_managed=False)
  File "/usr/lib/python3.12/site-packages/libqtile/config.py", line 661, in next_group
    n = self.group.get_next_group(skip_empty, skip_managed)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/libqtile/group.py", line 436, in get_next_group
    return self._get_group(1, skip_empty, skip_managed)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/libqtile/group.py", line 429, in _get_group
    index = (groups.index(self) + direction) % len(groups)
             ^^^^^^^^^^^^^^^^^^
ValueError: <group.Group ('2')> is not in list
```

What I found is that when reloading, all groups are changed for new ones but when the hook is fire the group associated to the window is still the old one and is not in `qtile.groups` which throws the error. This PR should fix this issue.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -1286,13 +1286,12 @@ class Core(base.Core, wlrq.HasListeners):
                             name = old_group[0]
                             if win.group.name == name:
                                 group = self.qtile.groups[i]
+                if win in win.group.windows:
+                    # Remove window from old group
+                    win.group.remove(win)
             if group is None:
                 # Falling back to current group if none found
                 group = self.qtile.current_group
-            if win.group and win in win.group.windows:
-                # It might not be in win.group.windows depending on how group state
-                # changed across a config reload
-                win.group.remove(win)
             group.add(win)
             if group == self.qtile.current_group:
                 win.unhide()
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -232,6 +232,9 @@ class Core(base.Core):
                 w for w in self.qtile.windows_map.values() if isinstance(w, window.Window)
             ]
             for managed_win in managed_wins:
+                if managed_win.group and managed_win in managed_win.group.windows:
+                    # Remove window from old group
+                    managed_win.group.remove(managed_win)
                 managed_win.set_group()
             return
 
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -656,20 +656,22 @@ class Screen(CommandObject):
         return dict(index=self.index, width=self.width, height=self.height, x=self.x, y=self.y)
 
     @expose_command()
-    def next_group(self, skip_empty: bool = False, skip_managed: bool = False) -> None:
+    def next_group(
+        self, skip_empty: bool = False, skip_managed: bool = False, warp: bool = True
+    ) -> None:
         """Switch to the next group"""
-        n = self.group.get_next_group(skip_empty, skip_managed)
-        self.set_group(n)
-        return n.name
+        group = self.group.get_next_group(skip_empty, skip_managed)
+        self.set_group(group, warp=warp)
+        return group.name if group is not None else None
 
     @expose_command()
     def prev_group(
         self, skip_empty: bool = False, skip_managed: bool = False, warp: bool = True
     ) -> None:
         """Switch to the previous group"""
-        n = self.group.get_previous_group(skip_empty, skip_managed)
-        self.set_group(n, warp=warp)
-        return n.name
+        group = self.group.get_previous_group(skip_empty, skip_managed)
+        self.set_group(group, warp=warp)
+        return group.name if group is not None else None
 
     @expose_command()
     def toggle_group(self, group_name: str | None = None, warp: bool = True) -> None:
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -425,9 +425,13 @@ class _Group(CommandObject):
                 return False
             return True
 
-        groups = [group for group in self.qtile.groups if match(group)]
-        index = (groups.index(self) + direction) % len(groups)
-        return groups[index]
+        try:
+            groups = [group for group in self.qtile.groups if match(group)]
+            index = (groups.index(self) + direction) % len(groups)
+            return groups[index]
+        except ValueError:
+            # group is not managed
+            return None
 
     def get_previous_group(self, skip_empty=False, skip_managed=False):
         return self._get_group(-1, skip_empty, skip_managed)
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -548,6 +548,23 @@ def test_nextprevgroup(manager):
     assert manager.c.group.info()["name"] == start
 
 
+def test_nextprevgroup_reload(manager_nospawn):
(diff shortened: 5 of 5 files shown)


tych0:
Can you add a test for this failing case? These bootstrapping issues are tricky, I have some code in my patch queue that reorders them myself, more testing here would be much appreciated.

shyguyCreate:
@tych0 hooks seem to behave different when testing because they are not fired while reloading as opposed to my machine where they are.

I have been trying this without success (the counter is just to make it more visual)

```python
from libqtile import hook
from multiprocessing import Value
def test_reload_groups(manager_nospawn):
    class GroupConfig(BareConfig):
        @hook.subscribe.group_window_add
        def group_window_add(group, window):
            assert group in group.qtile.groups

        @hook.subscribe.group_window_remove
        def group_window_remove(group, window):
            assert group in group.qtile.groups

    manager_nospawn.add = Value("i", 0)
    manager_nospawn.remove = Value("i", 0)

    def inc_add(group, window):
        manager_nospawn.add.value += 1

    def inc_remove(group, window):
        manager_nospawn.remove.value += 1

    hook.subscribe.group_window_add(inc_add)
    hook.subscribe.group_window_remove(inc_remove)

    manager_nospawn.start(GroupConfig)
    _, group = manager_nospawn.c.eval("self.groups")
    manager_nospawn.test_window("Test Window")
    manager_nospawn.c.reload_config()
    assert group == manager_nospawn.c.eval("self.groups")[1]
    assert manager_nospawn.add.value == 2
    assert manager_nospawn.remove.value == 1
```

Error from counter:

```python
>       assert manager_nospawn.add.value == 2
E       assert 1 == 2
E        +  where 1 = <Synchronized wrapper for c_int(1)>.value
E        +    where <Synchronized wrapper for c_int(1)> = <test.helpers.TestManager object at 0x7f348ff95280>.add

test/test_manager.py:1337: AssertionError
```

shyguyCreate:
You might think that the window is not removed nor added but if I do this:

```diff
diff --git a/libqtile/group.py b/libqtile/group.py
index 22b2b884..dc08d058 100644
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -249,6 +249,7 @@ class _Group(CommandObject):
         )
 
     def add(self, win, focus=True, force=False):
+        print("add")
         hook.fire("group_window_add", self, win)
         if win not in self.windows:
             self.windows.append(win)
@@ -270,6 +271,7 @@ class _Group(CommandObject):
 
     def remove(self, win, force=False):
         self.windows.remove(win)
+        print("remove")
         hook.fire("group_window_remove", self, win)
         hadfocus = self._remove_from_focus_history(win)
         win.group = None
```

I get the following, which means that the hook indeed should be fired.

```
---------------------------------------- Captured stdout call -----------------------------------------
add
remove
add
-------------------------------------- Captured stdout teardown ---------------------------------------
remove
```

(Ignoring teardown, not sure what that means)

shyguyCreate:
I could just check if the group is still in `groups` but I can not get the Group object into the variables, if I use `eval` I get a string, and if I use something like `manager_nospawn.c.current_group` I get an error **libqtile.command.base.SelectError: Not valid child or command, name: current_group, selectors: []**

shyguyCreate:
@tych0 what is the objective of the test? Maybe I should just change the PR to avoid the extra conditional if that is the bother.

```diff
diff --git a/libqtile/group.py b/libqtile/group.py
index 99c043e6..2f891277 100644
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -412,9 +412,6 @@ class _Group(CommandObject):
             skips the groups that have a screen
         """
 
-        if self not in self.qtile.groups:
-            return None
-
         def match(group):
             from libqtile import scratchpad
 
@@ -428,9 +425,12 @@ class _Group(CommandObject):
                 return False
             return True
 
-        groups = [group for group in self.qtile.groups if match(group)]
-        index = (groups.index(self) + direction) % len(groups)
-        return groups[index]
+        try:
+            groups = [group for group in self.qtile.groups if match(group)]
+            index = (groups.index(self) + direction) % len(groups)
+            return groups[index]
+        except ValueError:
+            pass
 
     def get_previous_group(self, skip_empty=False, skip_managed=False):
         return self._get_group(-1, skip_empty, skip_managed)
```

tych0:
IMO, the objective of the test would be to enforce that the reboot bootstrapping order is "sane", whatever "sane" means in this context (e.g. that groups are initialized before any user hooks are called, which seems like what is breaking in this case). Sorry, I haven't had much time this week to look at your PR in detail.

Given that you got a test to work, so long as it fails before your patch and succeeds after, I think that's a good thing to add.

shyguyCreate:
@tych0 after a clarification on the test, I tried something new but more useful. (ignore the last 3 lines is a WIP)

```python
def test_nextprevgroup_reload(manager_nospawn):
    manager_nospawn.start(BareConfig)
    # Current group will become unmanaged after reloading
    manager_nospawn.c.eval("self.test_group = self.current_group")
    manager_nospawn.c.reload_config()
    # Not managed group should not change the group in the screen
    success, message = manager_nospawn.c.eval("self.test_group.screen.next_group()")
    assert success, message
    assert success and message == "False"
    success, message = manager_nospawn.c.eval("self.test_group.screen.prev_group()")
    assert success and message == "False"
```

But got an error different than if I test it locally.

```python
>       assert success, message
E       AssertionError: AttributeError: 'NoneType' object has no attribute 'next_group'
E       assert False

test/test_manager.py:558: AssertionError
```

It seems that the screen is now None, but in my machine I would get that the screen is still set in the test_group.

```sh
> qtile cmd-obj -f eval -a 'self.test_group = self.current_group'
(True, None)
> qtile cmd-obj -f reload_config
> qtile cmd-obj -f eval -a 'self.test_group.screen'
(True, '<libqtile.config.Screen object at 0x716a515b41a0>')
> qtile cmd-obj -f eval -a 'self.test_group.screen.next_group()'
(False, "ValueError: <group.Group ('2')> is not in list")
```

shyguyCreate:
After your clarification, I decided to make a more useful test than the last one.

tych0:
Great, thank you!

PR #5061, opened 2024-10-27
Add redshift widget

saths008:
Adds a basic widget for redshift (sets the temperature, brightness and gamma RGB that the user provides when enabled).

### DIFF
--- a/libqtile/widget/redshift.py
+++ b/libqtile/widget/redshift.py
@@ -1,0 +1,316 @@
+# Copyright (c) 2024 Saath Satheeshkumar(saths008)
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import subprocess
+
+from libqtile.command.base import expose_command
+from libqtile.widget import base
+from libqtile.widget.base import _TextBox
+
+
+class GammaGroup:
+    """
+    GammaGroup is a helper class to group redshift gamma settings.
+    """
+
+    def __init__(self, red, green, blue):
+        self.gamma_red = red
+        self.gamma_green = green
+        self.gamma_blue = blue
+
+    def _format_gamma(self, format_txt: str) -> str:
+        return format_txt.format(
+            gamma_red=self.gamma_red,
+            gamma_green=self.gamma_green,
+            gamma_blue=self.gamma_blue,
+        )
+
+    def _redshift_fmt(self) -> str:
+        return f"{self.gamma_red}:{self.gamma_green}:{self.gamma_blue}"
+
+    def __repr__(self) -> str:
+        return f"Gamma: {self._redshift_fmt()}"
+
+    def __str__(self) -> str:
+        return (
+            f"GammaGroup(red={self.gamma_red}, green={self.gamma_green}, blue={self.gamma_blue})"
+        )
+
+
+class RedshiftDriver:
+    """
+    RedshiftDriver is a helper class
+    that executes all of the interactions with redshift.
+    """
+
+    def enable(self, temperature, gamma: GammaGroup, brightness):
+        """
+        Run redshift command with given parameters
+        """
+        subprocess.run(
+            [
+                "redshift",
+                "-P",
+                "-O",
+                str(temperature),
+                "-b",
+                str(brightness),
+                "-g",
+                gamma._redshift_fmt(),
+            ],
+            check=True,
+        )
+
+    def reset(self):
+        """
+        Call reset on redshift to reset to default settings.
+        """
+        subprocess.run(["redshift", "-x"], check=True)
+
+
+class Redshift(_TextBox, base.PaddingMixin):
+    """
+    Redshift widget provides the following functionality:
+      - Call redshift with a specific brightness, temperature and gamma config without using your location.
+
+    The redshift command can be called by just left-clicking the widget and disabling it the same way.
+
+    Scrolling through the widget will show the settings,
+    ie. the brightness, gamma and temperature that redshift
+    will use if the widget is enabled.
+
+    Widget requirements: redshift_
+
+    .. _redshift: https://github.com/jonls/redshift
+    """
+
+    defaults = [
+        ("brightness", 1.0, "Redshift brightness"),
+        (
+            "disabled_txt",
+            "󱠃",
+            "Text to show when redshift is disabled. NOTE: by default, a nerd icon is used."
+            "Available fields: 'brightness' brightness to set when redshift is enabled, "
+            "'is_enabled' boolean to state whether the widget is enabled or not, "
+            "'gamma_blue' gamma blue value to set when redshift is enabled, "
+            "'gamma_green' gamma green value to set when redshift is enabled, "
+            "'gamma_red' gamma red value to set when redshift is enabled, "
+            "'temperature' temperature to set when redshift is enabled, ",
+        ),
+        (
+            "enabled_txt",
+            "󰛨",
+            "Text to show when redshift is disabled. NOTE: by default, a nerd icon is used."
+            "Available fields: see disabled_txt's available fields for all of them",
+        ),
+        ("gamma_red", 1.0, "Redshift gamma red setting"),
+        ("gamma_blue", 1.0, "Redshift gamma blue"),
+        ("gamma_green", 1.0, "Redshift gamma green"),
+        ("font", "sans", "Default font"),
+        ("fontsize", 20, "Font size"),
+        ("foreground", "ffffff", "Font colour for information text"),
+        ("temperature", 1700, "Redshift temperature to set when enabled"),
+        (
+            "temperature_fmt",
+            "Temperature: {temperature}",
+            "Text to display when showing temperature text"
+            "Available fields: 'temperature' temperature to set when redshift is enabled. ",
+        ),
+        (
+            "brightness_fmt",
+            "Brightness: {brightness}",
+            "Text to display when showing brightness text"
+            "Available fields: 'brightness' brightness to set when redshift is enabled. ",
+        ),
+        (
+            "gamma_fmt",
+            "Gamma: {gamma_red}:{gamma_green}:{gamma_blue}",
+            "Text to display when showing gamma text"
+            "Available fields: 'gamma_red' gamma red value to set when redshift is enabled. "
+            "'gamma_blue' gamma blue value to set when redshift is enabled. "
+            "'gamma_green' gamma green value to set when redshift is enabled. ",
+        ),
+    ]
+
+    supported_backends = {"x11"}
+
+    def __init__(self, **config):
+        _TextBox.__init__(self, **config)
+        self.add_defaults(Redshift.defaults)
+        self.add_defaults(base.PaddingMixin.defaults)
+        self.is_enabled = False
+        self._line_index = 0
+        self._lines = []
+        self.redshift_driver = RedshiftDriver()
+        self.add_callbacks(
+            {
+                "Button1": self.click,
+                "Button4": self.scroll_up,
+                "Button5": self.scroll_down,
+            }
+        )
+
+    def _configure(self, qtile, bar):
+        _TextBox._configure(self, qtile, bar)
+        # disable redshift so we have a known initial state
+        self.reset_redshift()
+        self._set_lines()
+        # set text after bar configuration is done
+        self.qtile.call_soon(self._set_text)
+
+    @expose_command
+    def scroll_up(self):
+        """
+        Scroll up to next item.
+        """
+        self._scroll(1)
+
+    @expose_command
+    def scroll_down(self):
+        """
+        Scroll down to next item.
+        """
+        self._scroll(-1)
+
+    def show_line(self):
+        """
+        Update the text with the the current index in lines.
+        """
+        if not self._lines:
+            return
+
+        line = self._lines[self._line_index]
+
+        self.update(line)
+
+    @expose_command
+    def enable_redshift(self):
+        """
+        Enable redshift with the temp, gamma and brightness params.
+        """
+        gamma_val = GammaGroup(self.gamma_red, self.gamma_green, self.gamma_blue)
+        try:
+            self._get_rs_driver().enable(self.temperature, gamma_val, self.brightness)
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"redshift: could not enable redshift: {e}")
+
+    @expose_command
+    def click(self):
+        """
+        Either enable or disable the widget ONLY if we are currently on the
+        first line ie. the enabled/disabled text.
+        """
+        if self._line_index != 0:
+            return
+        if self.is_enabled:
+            self.reset_redshift()
+        else:
+            self.enable_redshift()
+        self.is_enabled = not self.is_enabled
+        self._set_text()
+
+    @expose_command
+    def reset_redshift(self):
+        """
+        Make a call to redshift to reset everything.
+        """
+        try:
+            self._get_rs_driver().reset()
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"redshift: could not reset redshift: {e}")
+
+    def _scroll(self, step):
+        """
+        Scroll up/down dictated by 'step' the items
+        and then display that item.
+        """
+        # Make sure that the lines var is set, maybe
+        # some of the values have been updated?
+        self._set_lines()
+        self._line_index = (self._line_index + step) % len(self._lines)
+        self.show_line()
+
+    def _set_lines(self):
+        """
+        Set the list of formatted text items to scroll through.
+        """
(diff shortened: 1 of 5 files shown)


libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+    def __repr__(self) -> str:
+        return f"Gamma: {self._redshift_fmt()}"
+
+    def __str__(self) -> str:
+        return (
+            f"GammaGroup(red={self.gamma_red}, green={self.gamma_green}, blue={self.gamma_blue})"
+        )
+
+
+class RedshiftDriver:
```
elParaguayo, line 57: I don't see any reason for this to be a separate class. There's no preservation of any class variables etc. I think you could just include these methods in the widget itself.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+                "redshift",
+                "-P",
+                "-O",
+                str(temperature),
+                "-b",
+                str(brightness),
+                "-g",
+                gamma._redshift_fmt(),
+            ],
+            check=True,
```
elParaguayo, line 78: If you're using `check=True` then you should wrap the call in `try... except subprocess.CalledProcessError` and do something if the command fails.

EDIT: ok - I see you've done this below.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        )
+
+    def reset(self):
+        """
+        Call reset on redshift to reset to default settings.
+        """
+        subprocess.run(["redshift", "-x"], check=True)
```
elParaguayo, line 85: As above.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+      - Call redshift with a specific brightness, temperature and gamma config without using your location.
+
+    The redshift command can be called by just left-clicking the widget and disabling it the same way.
+
+    Scrolling through the widget will show the settings,
+    ie. the brightness, gamma and temperature that redshift
+    will use if the widget is enabled.
+    """
+
+    redshift_driver = RedshiftDriver()
```
elParaguayo, line 100: As above.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+    orientations = base.ORIENTATION_HORIZONTAL
```
elParaguayo, line 101: I think this can be removed. Text boxes should display fine both horizontally and vertically.
```suggestion

```

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+            "gamma_fmt",
+            "Gamma: {gamma_red}:{gamma_green}:{gamma_blue}",
+            "Text to display when showing gamma text"
+            "Available fields: 'gamma_red' gamma red value to set when redshift is enabled. "
+            "'gamma_blue' gamma blue value to set when redshift is enabled. "
+            "'gamma_green' gamma green value to set when redshift is enabled. ",
+        ),
+    ]
+
+    _dependencies = ["redshift"]
```
elParaguayo, line 151:
```suggestion

```
Ha. This is something that's used in qtile-extras docs but not in qtile.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+
```
elParaguayo, line 152:
```suggestion

    supported_backends = {"x11"}

```
My understanding is that `redshift` only works on x11 so we need to make sure it's not loaded in wayland systems.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        )
+
+    def _configure(self, qtile, bar):
+        _TextBox._configure(self, qtile, bar)
+        # disable redshift so we have a known initial state
+        self.reset_redshift()
+        self._set_lines()
+        self._set_text()
+
+    def draw(self):
```
elParaguayo, line 175: Is this method needed? The text should already be set.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        Set the widget text and redraw the widget.
+        """
+        self.drawer.clear(self.background or self.bar.background)
+        self._set_text()
+        _TextBox.draw(self)
+
+    def format_object(self, obj):
```
elParaguayo, line 183: This method seems odd. I'm not clear why you need it if you're only checking if it's a string or not.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        obj = self._lines[self._line_index]
+
+        self.update(self.format_object(obj))
+
+    @expose_command
+    def enable_redshift(self):
+        """
+        Enable redshift with the temp, gamma and brightness params.
+        """
+        gamma_val = GammaGroup(self.gamma_red, self.gamma_green, self.gamma_blue)
```
elParaguayo, line 225: If these values never change then you can create `gamma_val` once (e.g. during `_configure`).

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        try:
+            self._get_rs_driver().enable(self.temperature, gamma_val, self.brightness)
+        except Exception as e:
```
elParaguayo, line 228:
```suggestion
        except subprocess.CalledProcessError as e:
```
It's bad practice to catch `Exception`. If some other error arises here then there's an unexpected bug which should be addressed rather than caught.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+            self._format_brightness_text(),
+            self._format_temp_text(),
+            self._format_gamma_text(
+                GammaGroup(self.gamma_red, self.gamma_green, self.gamma_blue)
+            ),
+        ]
+
+    def _set_text(self):
+        """
+        Set text (not update) to the current index of lines.
```
elParaguayo, line 291: Why `not update`? Is this because you call the line during `_configure` and so don't want to trigger a draw of the bar? That's easy enough to fix - see my suggestion in `_configure`.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+                "Button5": self.scroll_down,
+            }
+        )
+
+    def _configure(self, qtile, bar):
+        _TextBox._configure(self, qtile, bar)
+        # disable redshift so we have a known initial state
+        self.reset_redshift()
+        self._set_lines()
+        self._set_text()
```
elParaguayo, line 173:
```suggestion
        self.qtile.call_soon(self._set_text)
```
This will run `self._set_text()` once the bar's configuration has completed.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        Set text (not update) to the current index of lines.
+        """
+        if self._line_index == 0:
+            if self.is_enabled:
+                self.text = self._format_first_text(str(self.enabled_txt))
+            else:
+                self.text = self._format_first_text(str(self.disabled_txt))
+        else:
+            self.text = self.format_object(self._lines[self._line_index])
```
elParaguayo, line 299:
```suggestion
            if self.is_enabled:
                text = self._format_first_text(str(self.enabled_txt))
            else:
                text = self._format_first_text(str(self.disabled_txt))
        else:
            text = self.format_object(self._lines[self._line_index])

        self.update(text)
```

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def _format_temp_text(self) -> str:
+        """
+        Format the temperature text.
+        """
+        return str(self.temperature_fmt).format(temperature=self.temperature)
```
elParaguayo, line 305: Does this need to be wrapped in `str`?

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def _format_brightness_text(self) -> str:
+        """
+        Format the brightness text.
+        """
+        return str(self.brightness_fmt).format(brightness=self.brightness)
```
elParaguayo, line 311: As above.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def _format_gamma_text(self, gamma: GammaGroup) -> str:
+        """
+        Format the gamma text.
+        """
+        return gamma._format_gamma(str(self.gamma_fmt))
```
elParaguayo, line 317: Why wouldn't `self.gamma_fmt` be a string?

test/widgets/test_redshift.py
```diff
@@ -0,0 +1,229 @@
...
+@pytest.mark.parametrize(
+    "redshift_manager",
+    [
+        {
+            "disabled_txt": "Redshift disabled",
+            "enabled_txt": "Redshift enabled",
+        }
+    ],
+    indirect=True,
+)
```
elParaguayo, line 133:
```suggestion
@config(disabled_txt="Redshift disabled", enabled_txt="Redshift enabled")
```
The `config` decorator simplifies the code here.

test/widgets/test_redshift.py
```diff
@@ -0,0 +1,229 @@
...
+            "disabled_txt": "brightness{brightness}, temp{temperature}, r{gamma_red}, g{gamma_green}, b{gamma_blue}, is_enabled{is_enabled}",
+            "enabled_txt": "brightness{brightness}, temp{temperature}, r{gamma_red}, g{gamma_green}, b{gamma_blue}, is_enabled{is_enabled}",
+            "gamma_red": 0.2,
+            "gamma_green": 0.2,
+            "gamma_blue": 0.2,
+            "temperature": 1200,
+        }
+    ],
+    indirect=True,
+)
```
elParaguayo, line 185:
```suggestion
@config(
    brightness=0.4,
    disabled_txt="brightness{brightness}, temp{temperature}, r{gamma_red}, g{gamma_green}, b{gamma_blue}, is_enabled{is_enabled}",
    enabled_txt="brightness{brightness}, temp{temperature}, r{gamma_red}, g{gamma_green}, b{gamma_blue}, is_enabled{is_enabled}",
    gamma_red=0.2,
    gamma_green=0.2,
    gamma_blue=0.2,
    temperature=1200,
)
```

elParaguayo (changes requested):
Thanks for this. I've gone through and made some suggestions. Let me know if you have any queries.

libqtile/widget/redshift.py
saths008: The main reason I did this was to make mocking easier (redshift isn't available in the test env so I could just mock out `redshift_driver`), is this reasonable or is there a better way to go about this?

libqtile/widget/redshift.py
saths008: I've decided to include redshift as part of the `widget_requirements` instead now like the `canto` widget has done

libqtile/widget/redshift.py
saths008: I was thinking about the situation where `self._lines` would contain things other than `str`. Eg. I was going to have `GammaGroup` but you're  right that's just confusing- removing the method

libqtile/widget/redshift.py
saths008: So I could get away without forcing it to be a string( we know it is a string) but I think python doesn't because we're declaring it in the `defaults` rather than through `self.` so I have a bunch of LSP errors: 
```
Diagnostics:
1. "format" is not a known attribute of "None" [reportOptionalMemberAccess]
2. Cannot access attribute "format" for class "partial[Unknown]"
     Attribute "format" is unknown [reportAttributeAccessIssue]

```

This doesn't stop the code from functioning but I thought it would be better to silence it with `str()`

libqtile/widget/redshift.py
saths008: Format comment above

libqtile/widget/redshift.py
saths008: Format comment above

libqtile/widget/redshift.py
saths008: Yep, thanks that explains the failed `wayland` test too. I've now skipped the tests if we are on `wayland`

saths008:
Note: I am adding an increase/decrease (brightness/temp) feat now too

libqtile/widget/redshift.py
elParaguayo: I thought that might be the reason. You can probably just mock the `subprocess.call` method instead. There are probably some examples of that in the test suite already.

saths008: Done, removed.

libqtile/widget/redshift.py
saths008: `draw` has now been removed

libqtile/widget/redshift.py
saths008: Thanks, done

saths008:
@elParaguayo  I also have `test_check_migrations.py` failing (even checked out to the commit `f53ded` and it still fails), could you shed some light on that?

elParaguayo:
> @elParaguayo I also have `test_check_migrations.py` failing (even checked out to the commit `f53ded` and it still fails), could you shed some light on that?

Failing locally? You probably need `isort` installed.

saths008:
Actually I'm sure it's fine, seeing as the CI passed here

saths008:
> > @elParaguayo I also have `test_check_migrations.py` failing (even checked out to the commit `f53ded` and it still fails), could you shed some light on that?
> 
> Failing locally? You probably need `isort` installed.

There was an assertion error, but I'm sure it's fine, seeing as the CI passed here

saths008:
> For your test failure, you'll need to add the widget to this line:
> 
> https://github.com/qtile/qtile/blob/f53ded59c1f05587ed92a2b31dc4741c8c36d1fc/test/widgets/test_widget_init_configure.py#L64
> 
> The test (currently) look at the `supported_backends` attribute of widgets.

Just putting this here so it's all in one place. The redshift widget currently also uses `reset_redshift` in `configure`- this crashes as a result of redshift not being in the test environment (regardless of wayland/x11), which is why I have skipped the test for `redshift` for now

libqtile/widget/redshift.py
saths008: @elParaguayo Am I ok to resolve this? The rest of the open issues are based on this

elParaguayo: I don't like the current code. Feels unnecessary but I've not had time to look at a solution.

saths008: Fair enough, don't worry I'll just fix the forcing of `str` then

saths008: Ok no worries

saths008: @elParaguayo If you let me know your reasons why I can take a look myself

elParaguayo: I don't like forcing the type just to make the linter happy.

Would it work by adding `temperature_format: str` after the class definition?

saths008: Yh you're right, let me give that a shot. What are your feelings on something like `assert isinstance(x, str)`?

elParaguayo: We've used those elsewhere to keep mypy happy so we could do it here but I still find that ugly. I'd prefer to see if we can set the type for that attribute separately and only use asset if the alternatives don't work.

saths008: Accidentally deleted my previous comment but I have declared the defaults like: `temperature_format: str` and that seems to work

saths008:
Recent push: Declared defaults with types so no more LSP complaints and removed previous type forcing

saths008:
@elParaguayo What else needs to be done for this to be merged, I'm pretty sure all your comments have been resolved. Thanks

elParaguayo:
I need time to do a final review. I'll do that when I can.

saths008:
> I need time to do a final review. I'll do that when I can.

Thanks, just a reminder

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,476 @@
...
+    def __repr__(self) -> str:
+        return f"Gamma: {self._redshift_fmt()}"
+
+    def __str__(self) -> str:
+        return (
+            f"GammaGroup(red={self.gamma_red}, green={self.gamma_green}, blue={self.gamma_blue})"
+        )
+
+
+class Redshift(_TextBox, base.PaddingMixin):
```
elParaguayo, line 56: You're not using the `PaddingMixin` at all so you could remove it. `_TextBox` has a `padding` parameter already.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,476 @@
...
+
+    @expose_command
+    def run_redshift(self):
+        """
+        Run redshift command with defined parameters.
+        """
+        try:
+            subprocess.run(
+                [
+                    "redshift",
```
elParaguayo, line 464: This is hardcoded. Can you either: i) make this a parameter (default value would be `shutil.which("redshift")` or ii) update the docs to say that `redshift` needs to be in the user's path.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,476 @@
...
+                    str(self.temperature),
+                    "-b",
+                    str(self.brightness),
+                    "-g",
+                    self.gamma_val._redshift_fmt(),
+                ],
+                check=True,
+            )
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"redshift: could not enable redshift: {e}")
```
elParaguayo, line 476: This should be a log message and the widget should display an error message.

libqtile/widget/redshift.py
```diff
@@ -0,0 +1,476 @@
...
+
+    @expose_command
+    def reset_redshift(self):
+        """
+        Call reset on redshift to reset to default settings.
+        """
+        try:
+            subprocess.run(["redshift", "-x"], check=True)
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"redshift: could not enable redshift: {e}")
```
elParaguayo, line 454: We should log the error and display an error message rather than raising an error.

test/widgets/test_widget_init_configure.py
```diff
@@ -59,6 +59,7 @@
 # To skip a test entirely, list the widget class here
 no_test = [widgets.Mirror, widgets.PulseVolume]  # Mirror requires a reflection object
 no_test += [widgets.ImapWidget]  # Requires a configured username
+no_test += [widgets.Redshift]  # Initialisation will reset redshift
```
elParaguayo, line 62: What does `initialisation will reset redshift` mean? This should be fine to test if we make the changes so the widget displays an error if it can't run `redshift`.

elParaguayo (changes requested):

test/widgets/test_widget_init_configure.py
saths008: The widget now displays a crash error, I've also allowed the initialisation test to take place which should now be fine.

saths008:
- Allow user to specify path to redshift executable
- Add widget crash message (in case, `redshift_path` doesn't exist or is invalid)
- Allow `initialisation` test in `widget_init_configure` for redshift
- Remove `PaddingMixin`

elParaguayo (approved):

PR #5062, opened 2024-10-27
Allow customizing battery widget short text

holocronweaver:
Previously the battery widget showed hard-coded short text (`Full` and `Empty`) when the battery is full or empty and "show_short_text" is enabled.

This change allows customizing the full and empty short text using the new `full_short_text` and `empty_short_text` parameters, while retaining the original hard-coded values as defaults.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -450,7 +450,21 @@ class Battery(base.ThreadPoolText):
         ("unknown_char", "?", "Character to indicate the battery status is unknown"),
         ("format", "{char} {percent:2.0%} {hour:d}:{min:02d} {watt:.2f} W", "Display format"),
         ("hide_threshold", None, "Hide the text when there is enough energy 0 <= x < 1"),
-        ("show_short_text", True, 'Show "Full" or "Empty" rather than formated text'),
+        (
+            "full_short_text",
+            "Full",
+            "Short text to indicate battery is full; see `show_short_text`",
+        ),
+        (
+            "empty_short_text",
+            "Empty",
+            "Short text to indicate battery is empty; see `show_short_text`",
+        ),
+        (
+            "show_short_text",
+            True,
+            "Show only characters rather than formatted text when battery is full or empty",
+        ),
         ("low_percentage", 0.10, "Indicates when to use the low_foreground color 0 < x < 1"),
         ("low_foreground", "FF0000", "Font color on low battery"),
         ("low_background", None, "Background color on low battery"),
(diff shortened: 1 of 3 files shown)


libqtile/widget/battery.py
```diff
@@ -550,13 +553,13 @@ def build_string(self, status: BatteryStatus) -> str:
...
             if self.show_short_text:
-                return "Full"
+                return self.full_short_text
             char = self.full_char
         elif status.state == BatteryState.EMPTY or (
             status.state == BatteryState.UNKNOWN and status.percent == 0
         ):
             if self.show_short_text:
-                return "Empty"
+                return self.full_empty_text
```
jwijenbergh, line 562: This should be `empty_short_text` right?

holocronweaver: Yep, fixed in latest revision.

elParaguayo:
CI is failing because you've used a tab in the changelog rather than spaces.

Also, it would be good if you could update `test/widgets/test_battery.py` to test this new code.

libqtile/widget/battery.py
jwijenbergh: Thanks!

holocronweaver:
Added unit tests and switched tabs to spaces in CHANGELOG. Updated my text editor to use spaces by default, so hopefully won't happen again. :sweat_smile:

jwijenbergh:
Thanks you just need to format the files. With black

holocronweaver:
@jwijenbergh I avoided using ruff in this case because the formatting doesn't match the rest of the file. I think keeping format consistent is better for readability.

I could do a formatting-only followup PR to format the whole file, but the original formatting is OK so could also just keep as is. I don't have a strong preference either way.

What do you think?

elParaguayo:
The entire codebase should have been formatted by ruff already. Are you saying it's making changes beyond the lines you added?

holocronweaver:
Hm, I thought Ruff was only being applied to diff (to allow formatting overrides when the autoformatter produces undesirable output), but I verified it is being applied to the whole file. It introduces inconsistent formatting with rest of file due to small line length overage from description strings:
```diff
         ("unknown_char", "?", "Character to indicate the battery status is unknown"),
         ("format", "{char} {percent:2.0%} {hour:d}:{min:02d} {watt:.2f} W", "Display format"),
         ("hide_threshold", None, "Hide the text when there is enough energy 0 <= x < 1"),
-        ("full_short_text", "Full", "Short text to indicate battery is full; see `show_short_text`"),
-        ("empty_short_text", "Empty", "Short text to indicate battery is empty; see `show_short_text`"),
-        ("show_short_text", True, "Show only characters rather than formatted text when"
-         " battery is full or empty"),
+        (
+            "full_short_text",
+            "Full",
+            "Short text to indicate battery is full; see `show_short_text`",
+        ),
+        (
+            "empty_short_text",
+            "Empty",
+            "Short text to indicate battery is empty; see `show_short_text`",
+        ),
+        (
+            "show_short_text",
+            True,
+            "Show only characters rather than formatted text when" " battery is full or empty",
+        ),
         ("low_percentage", 0.10, "Indicates when to use the low_foreground color 0 < x < 1"),
         ("low_foreground", "FF0000", "Font color on low battery"),
         ("low_background", None, "Background color on low battery"),
```

Will update PR.

jwijenbergh (approved):

jwijenbergh:
Thanks! @holocronweaver

PR #5076, opened 2024-11-09
x11: look at all the wm type atoms

### LINKED ISSUE #5065: Zoom screenshare leaves behind zombie windows
### Issue description

 After a zoom call with screenshare some fake window remains in qtile stack of managed windows even though the window does not exists anymore. This messes up both with navigation commands and with appearence. In the following example the right side of the screen is unusable.

```
[ins] In [1]: from libqtile.command.client import InteractiveCommandClient
          ...: c = InteractiveCommandClient()
          ...: c .group.info()
Out[1]:
{'name': '5',
 'label': '5',
 'focus': 'VulK@bunion: ipython',
 'tiled_windows': {'VulK@bunion: ipython', 'zoom'},
 'windows': ['VulK@bunion: ipython'],
 'focus_history': ['VulK@bunion: ipython'],
 'layout': 'columns',
 'layouts': ['max', 'columns'],
 'floating_info': {'name': 'floating', 'clients': []},
 'screen': 1}
```
image

### Version

0.29.0

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

tych0:
_NET_WM_WINDOW_TYPE is really a *list*, not a single atom. We were previously using just the first element to distinguish types, but at least one application (Zoom, see #5065) set multiple _NET_WM_WINDOW_TYPE atoms, and the first one is not one that we understand natively:

    _NET_WM_WINDOW_TYPE(ATOM) = _KDE_NET_WM_WINDOW_TYPE_OVERRIDE, _NET_WM_WINDOW_TYPE_NORMAL

however, the second one is. Let's look through this entire list and see if we understand any of the types natively. If it doesn't exist, we still return the raw first atom, to preserve previous behavior.

Unclear if this fixes #5065, but it is certainly a bug that is easily fixed, so let's fix it :)

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -281,8 +281,16 @@ class XWindow:
         """
         r = self.get_property("_NET_WM_WINDOW_TYPE", "ATOM", unpack=int)
         if r:
-            name = self.conn.atoms.get_name(r[0])
-            return xcbq.WindowTypes.get(name, name)
+            first_name = None
+            for i, a in enumerate(r):
+                name = self.conn.atoms.get_name(a)
+                if i == 0:
+                    first_name = name
+                qtile_type = xcbq.WindowTypes.get(name, None)
+                if qtile_type is not None:
+                    return qtile_type
+            return first_name
+        return None
 
     def get_net_wm_state(self):
         r = self.get_property("_NET_WM_STATE", "ATOM", unpack=int)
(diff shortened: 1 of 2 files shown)


tych0:
I will try to write a test for this in a bit; our existing `test/scripts/window.py` can't actually replicate this state because GDK's API only allows one wm type, so it'll have to be raw xcffib calls.

elParaguayo:
Nice. Happy for you to merge this.

PR #5077, opened 2024-11-09
`qtile migrate` - defer libcst import

elParaguayo:
libcst import can be slow so we only import it when we're actually executing a `qtile migrate` command.

### DIFF
--- a/libqtile/scripts/migrate.py
+++ b/libqtile/scripts/migrate.py
@@ -33,13 +33,9 @@ from libqtile.utils import get_config_file
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
-BACKUP_SUFFIX = ".migrate.bak"
-
-
-try:
     import libcst
-except ImportError:
-    pass
+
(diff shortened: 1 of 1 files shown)


tych0:
LGTM (tests notwithstanding of course :)

tych0:
can you squash this so it doesn't break bisection? but then i'd say let's merge it.

elParaguayo:
Squashed. But I would like @ervinpopescu to confirm this fixes the issue.

ervinpopescu:
Confirmed as fixed.

```bash
$ time ./bin/qtile cmd-obj -f windows &>/dev/null

real    0.96s
user    0.85s
sys     0.10s
cpu     98%
```

I would have hoped it was a bit faster to compare with qticc, but it is what it is.
```bash
$ time qticc cmd-obj -f windows &>/dev/null

real    0.01s
user    0.00s
sys     0.00s
cpu     44%
```

I'll open a PR if I manage to find areas of further improvement.

PR #5078, opened 2024-11-10
Add `VerticalClock` widget.

elParaguayo:
The `VerticalClock` widget displays text horizontally in vertical bars.

Draft for now. Would be good to have some people test this. I also need to write some tests.

### DIFF
--- a/docs/manual/howto/widget.rst
+++ b/docs/manual/howto/widget.rst
@@ -626,6 +626,7 @@ An example screenshot file is below:
     import pytest
 
     from libqtile.widget import wttr
+    from test.widgets.docs_screenshots.conftest import vertical_bar, widget_config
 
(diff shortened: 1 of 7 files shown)


elParaguayo:
I'd like to add a screenshot for the docs but that needs a little more work as all screenshots are with horizontal bars.

tych0 (approved):

elParaguayo:
Screenshot is working but I need to rebase this to split out the screenshot manager changes from the widget.

ervinpopescu:
Looks fine to me.

elParaguayo:
> Although, you sure default needs to be `%H %M %S`?
> 
> I for one only use `HH:MM` in my bar.

Good point. Will change that.

PR #5082, opened 2024-11-13
widget/sensors: switch to ThreadPoolText

### LINKED ISSUE #5080: Switching workspaces sometimes takes seconds!
### Issue description

This problem started recently. I tried without picom, it didn't stop the issue. 

### Version

### Backend

X11 (default)

### Config

_No response_

### Logs

_No response_

### Required

tych0:
We have reports of this widget causing the whole qtile UI to hang, which makes sense since it's InLoopPollText, which as that base's docstring says can block the event loop.

I think it's likely that it's the psutil.sensors_temperature() call that's blocking; looking at the implementation [1], it walks sysfs, which may take a while if some particular hardware device is slow to respond (or if there are lots of sysfs devices to be queried).

Instead, let's switch this to be a ThreadPoolText based widget, so that updates do not block the main UI event loop.

Fixes #5080

[1]: https://github.com/giampaolo/psutil/blob/687695c289a84585b20f2037eee84fa11ca49b81/psutil/_pslinux.py#L1377-L1492


elParaguayo:
I think you need to change a couple more lines in the widget which still refer to `InLoopPollText`

tych0:
Oh, yes, of course. Funny that we didn't crash in the tests...

PR #5084, opened 2024-11-13
widget: move update_interval to the base widget

tych0:
This one is a little uglier, and not a strongly held opinion, but maybe it's worth merging.

"Almost" all widgets (see below)  poll things instead of listen for events in the event loop. So they all have some kind of update_interval, which lets us get rid of these custom __init__ methods. Since _Widget has the logic to call setup_timer, the concept of widget timers is already implied by _Widget, so I think this probably makes sense.

We could commonize these to a mixin, but that's almost more code than just leaving them copypasta'd.

Widgets like pulse_volume do not actually need a timer (which it notes in its docs) because it installs an fd listener directly in the event loop. However, it installs one anyway by virtue of inheriting from Volume :)

In principle the mpris2widget also does not need a timer doing the same eventloop trick, but it does have additional custom code to install a timer based on a configuration setting called poll_interval instead, so it does *both*. Perhaps we should s/poll_interval/update_interval and delete some of that at the very least.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -495,6 +495,11 @@ class _TextBox(_Widget):
             "When ``scroll=True`` the ``width`` parameter is a maximum width and, when text is shorter than this, the widget will resize. "
             "Setting ``scroll_fixed_width=True`` will force the widget to have a fixed width, regardless of the size of the text.",
         ),
+        (
+            "update_interval",
+            600,
+            "Update interval in seconds, if none, the widget updates only once.",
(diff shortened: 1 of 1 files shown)


elParaguayo:
I'm a bit on the fence with this one. I'm all for simplifying but this will pollute the docs slightly because, as you note, some widgets like `WindowName` etc don't need `update_interval`.

However, that issue extends beyond `update_interval` and I had been thinking about a PR to hide unused variables from the docs.

PR #5090, opened 2024-11-19
Fire window_remove hook earlier

mingmingrr:
Moving group_window_remove hook to first statement, so hooks can know what index the removed window was at.

Not sure if a migration is needed for this. Seems like the hook is new enough and the use-case for this is rare enough to get by without one.

### DIFF
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -269,8 +269,8 @@ class _Group(CommandObject):
             self.focus(win, warp=True, force=force)
 
     def remove(self, win, force=False):
-        self.windows.remove(win)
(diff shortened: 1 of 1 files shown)


elParaguayo:
What is the issue you're trying to solve here? i.e. what's the "index" you're referring to?

mingmingrr:
I would like the window to still be in _Group.windows so I have its index in that list.

elParaguayo:
OK. I think this is probably fine and it's consistent with the `group_window_add` hook which is also fired before the window has been added.

PR #5095, opened 2024-11-23
WIP: Refactor and improve extensions 

sargon:
Refactor the extension classes and improve them to be more hackable.

 * Cleanup the structure towards type safety 
 * Move code into functional methods.
   * Allowing usage of functionality of superclasses in deeper subclasses
 * Allow function as command arguments in CommandSet extension
   * Allows faster response time for qtile actions, instead of calling qtile cmd via spawn
   * Improve extension to a hackable config prototype shell for new extension ideas. 

Sadly the refactoring towards type safety of the extension classes breaks API compatibility
for repo external extensions.

Need some more review and patch cleanup.


libqtile/extension/base.py
```diff
@@ -18,66 +18,37 @@
...
 from subprocess import PIPE, Popen
-from typing import Any
+from typing import TYPE_CHECKING
 
 from libqtile import configurable
 from libqtile.log_utils import logger
 
-RGB = re.compile(r"^#?([a-fA-F0-9]{3}|[a-fA-F0-9]{6})$")
+if TYPE_CHECKING:
+    from typing import Any
```
tych0, line 32: This seems like a step backwards; can we get rid of this and just use Optional directly if we drop support for e.g. 3.10?

tych0:
> Sadly the refactoring towards type safety of the extension classes breaks API compatibility for repo external extensions.

is the breakage something you can write a migration for? I haven't really read the code closely, but e.g. duplicating _check_colors() seems not ideal.

sargon:
The API break arise from the return value of the run function.
Before it was either Popen or str, now it is None.
The functionality for the sub-classes have been moved into separate function.

I think the color checking function is more of a general utility function and should be available via a utility module. I just moved it to the first using sub-class to get a more logical structure.

libqtile/extension/base.py
sargon: That is how it is done. In this case the TYPE_CHECKING block seems redundant, because it is only used for typing module stuff. But you have those blocks anyway in the whole project.

I feel deprecating the python version is a topic that should be discussed on a broader level.

tych0: > But you have those blocks anyway in the whole project.

Yes, and I have been deleting them as we can; I really don't want to add more.

tych0:
> The API break arise from the return value of the run function.

Can you say a little bit about why this API break is valuable? I don't see it in the commit logs. But either way, we'll need a migration for it.

elParaguayo:
Is that return value only accessed internally by the extensions themselves, or can users access it in their config a?

If it's the former then wouldn't this only be an issue for anyone who's written a custom extension?

sargon:
Moving the default settings is relevant for of repo extensions (sub-classes of _Extension). 
The  API break (introduced in https://github.com/qtile/qtile/pull/5095/commits/af83cb79fdbaf6feed80eb83ca1fdceee03118b6 ) is relevant for of repo extensions (sub-classes of RunCommand or Dmenu).

From config view these changes are irrelevant.

elParaguayo:
Exactly. Migrations are designed to update people's configs for breaking changes. I'm not sure this qualifies. 

There needs to be a nite in the changeling but I don't think we can rewrite custom extensions that users may have created.

libqtile/extension/base.py
sargon: In that case we can replace the typing.Optional with `x | None`, but where does `typing.Any` come from?
You could argue that we can get rid of the block in this case by default importing `typing.Any`, but getting rid
of all the `if TYPE_CHECKING` blocks is going to be very hard, iff you want to be able to strictly type check the code - my experience.

elParaguayo: `Any` was typically used to typing the `defaults` section.

tych0:
> Exactly. Migrations are designed to update people's configs for breaking changes. I'm not sure this qualifies.

What is the difference between an extension and stuff in a user's config?

elParaguayo:
> > Exactly. Migrations are designed to update people's configs for breaking changes. I'm not sure this qualifies.
> 
> What is the difference between an extension and stuff in a user's config?

If someone wants to use the `WindowList` extension, they would do something like this:
```python
from libqtile import extension
keys = [
    Key(['mod4'], 'r', lazy.run_extension(extension.WindowList())),
    ...
]
```
Any API changes to the base extension class are invisible to the config. There should be nothing to migrate (i.e. that command would still work).

However, if a user creates their own extension in their config...
```python
from libqtile.extension.dmenu import Dmenu


class CustomExtension(Dmenu):

    def run(self):
        ...
        out = super().run(items=self.item_to_win.keys())
```
`out` is now a different object type than before so, without changes, the custom extension would break.

tych0:
> out is now a different object type than before so, without changes, the custom extension would break.

Right, that is the case I'm concerned about. My understanding of "extensions" is that they are just ways for us to carry bits of code upstream that interact with specific applications, so maybe nobody has out-of-tree ones?

Should we make it explicit that we do not guarantee out-of-tree extensions will be forwards compatible? But if that's the case, I wonder if "extensions" is really the wrong name.

PR #5097, opened 2024-11-25
Dont fork tests

tych0:
Here is the first patches in what I hope will not be a super long slog to get rid of our fork warnings in the tests. Then we should enable `-W error` so we can catch these sooner :)


elParaguayo:
Big fan of handling the logging better in tests! Thanks for fixing that.

tych0:
I don't know about "better", but certainly "different" :). Thanks for the review.

PR #5103, opened 2024-11-27
tests, bluetooth: use a process instead of a thread

tych0:
...for our fake daemon.

The in our fixture setup, the bluetooth daemon thread necessarily must start before the manager, because the widget wants to connect on startup. Of course, we fork() the manager (via multiprocessing), so spawning this Thread before then will cause a warning. Let's just use a process instead.

This is in service of getting rid of all multithreading so that we can enable -W error. It seems to (?) eliminate all the warnings besides those in test_window_list.py, although 3.13 of course introduces more warnings inside dependencies :(


elParaguayo:
Thanks. We also use threads in `test/core/test_exitcode.py`. Are those ones OK?

tych0:
You are quite right, I have patches for those too. I just didn't want to send too many patches all at once. In fact they would have been fine with some slight adjustments, but I converted them to multiprocessing anyway. Let me land this and send that series...

PR #5106, opened 2024-11-29
Pre-commit: update dependencies to latest version

RFCreate:


tych0:
Hmm, this seems like an API break. Is there a reason for this change?

shyguyCreate:
It isn't really used in the codebase and it doesn't comply with the return type annotation.

tych0:
What if we change the return type to be `str | None` instead? Then we don't need to worry about API breakage.

elParaguayo:
Do we ever use the return value?

Surprised this hasn't been flagged by our type checkers before.

tych0:
> Surprised this hasn't been flagged by our type checkers before.

I wonder if the type gets erased during expose_command()?

shyguyCreate:
> I wonder if the type gets erased during expose_command()?

I guess that it does not know the type `group.name` will return and treats that as `None`.

> Do we ever use the return value?

Not really, it should throw a error in the tests if it was.

RFCreate:
> Surprised this hasn't been flagged by our type checkers before.

@tych0 @elParaguayo, I updated pre-commit dependencies to latest version, which now with a new version of mypy shows the error that you where concern regarding the type checker.

Issue #4658, opened 2024-01-23
0.24.0: test_all_migrations[MatchListRegex-0] is failing


dvzrv:
### Issue description

Hi! :wave: 

I'm currently upgrading the Arch Linux package to 0.24.0 and ran into the failing test `test_all_migrations[MatchListRegex-0]`.
Here is the build script: https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/blob/c543d02ffd76b4b4cb98c615a5c5fa6648733380/PKGBUILD

Full build and check logs:
[qtile-0.24.0-1-x86_64-build.log](https://github.com/qtile/qtile/files/14027824/qtile-0.24.0-1-x86_64-build.log)
[qtile-0.24.0-1-x86_64-check.log](https://github.com/qtile/qtile/files/14027822/qtile-0.24.0-1-x86_64-check.log)

As I currently do not have the time for further debugging this and don't know how to fix the issue at hand, I am disabling the test for now.

### Version

24.0.0

### Backend

X11 (default), Wayland (experimental)

### Logs

```text
=================================== FAILURES ===================================
____________________ test_all_migrations[MatchListRegex-0] _____________________

migration_tester = <test.migrate.conftest.MigrationTester object at 0x7333b9f4bfd0>

    @pytest.mark.parametrize("migration_tester", migration_tests, indirect=True, ids=migration_ids)
    def test_all_migrations(migration_tester):
        # If the migration is expected to result in a change then it should also provide linting
        if migration_tester.test.needs_lint:
            migration_tester.assert_lint(), "No linting provided"

        # Tests whether the expected output is returned by the migration
>       migration_tester.assert_migrate()

test/migrate/test_migrations.py:51:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test.migrate.conftest.MigrationTester object at 0x7333b9f4bfd0>

    def assert_migrate(self):
        if not self.test.source:
            assert False, f"{self.test_id} has no tests."
        argv = [self.cmd, "migrate", "--yes", "-r", self.test_id, "-c", self.test.source_path]
        try:
            subprocess.check_call(argv)
        except subprocess.CalledProcessError:
            assert False

        updated = Path(f"{self.test.source_path}").read_text()
>       assert updated == self.test.expected
[...]

PR #5108, opened 2024-12-01
tests: fix failing tests on NixOS

mingmingrr:
In test/backend/x11/test_window.py: `subprocess.run(["import", ...])` fails with:
`FileNotFoundError: [Errno 2] No such file or directory: 'import'`
For some reason, this errors even if imagemagick is installed hence the `shutil.which`.

In libqtile/scripts/migrations/match_list_regex.py, a missing `isort` import causes test_all_migrations[MatchListRegex-0] to fail. Adding it to common-python-deps fixes this. Related issue: #4658

### DIFF
--- a/test/widgets/docs_screenshots/ss_bluetooth.py
+++ b/test/widgets/docs_screenshots/ss_bluetooth.py
@@ -19,7 +19,7 @@
 # SOFTWARE.
 import pytest
 
-from test.widgets.test_bluetooth import dbus_thread, wait_for_text, widget  # noqa: F401
+from test.widgets.test_bluetooth import fake_dbus_daemon, wait_for_text, widget  # noqa: F401
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tych0:
> For some reason, this errors even if imagemagick is installed hence the shutil.which.

Hmm, that seems odd. Can you post the `$PATH` when `shutil.which` works, and where the binary is finally located when it does?

> In libqtile/scripts/migrations/match_list_regex.py, a missing isort import causes test_all_migrations[MatchListRegex-0] to fail. Adding it to common-python-deps fixes this.

this part looks good to me; you could separate it out as a separate patch and I'd be happy to land that.

mingmingrr:
Here's my `$PATH`: [path.txt](https://github.com/user-attachments/files/17970641/path.txt)
`print(shutil.which("import"))` shows it's at `/nix/store/irkcg78qfr4q5hfmjw8k6w92f8bgb39f-imagemagick-7.1.1-38/bin/import`

I did some more digging with a modified subprocess module. Turns out when `env=` is set in `subprocess.run()`, it gets passed along to `os.get_exec_path`, and the current `$PATH` is not automatically inserted, in which case it uses a fallback of `PATH=/usr/bin:/bin`. So either `$PATH` would need to be added to `env=`, or a filepath would need to be passed as the executable.

TLDR: to make that test work on NixOS, we would either need `shutil.which`, or add `$PATH` to `subprocess.run(env=...)`

tych0:
Nice, that makes sense. Can you add that second blurb to your commit message? Then I think it looks good to me!

tych0:
Arg. I tried to update with the commit log, and it force closed the PR and I can't reopen it.

tych0:
idk, I reopened as #5113; I will land it from there. Thanks for the patch!

PR #5128, opened 2024-12-21
StatusNotifier - check for dbus icon

### LINKED ISSUE #5122: Status Notifier doesn't show fcitx5 icons for languages
### Issue description

Normally on Systray fcitx5 would show the icons of any languages/input methods that use icons instead of text.
Such as MOZC for japanese.
However, on StatusNotifier the icon isn't rendered at all.

Fcitx5 on text based language icons
2024-12-15-141514_3840x1080_scrot
Fcitx5 on icon based language icons
20241215_14h12m58s_grim
On Systray:
2024-12-15-141427_3840x1080_scrot

Maybe the icons have some kind of X11 path. All other icons behave normally

### Version

### Backend

X11 (default), Wayland (experimental)

### Config

_No response_

### Logs

_No response_

### Required

elParaguayo:
Some apps use a combination of icons from users' systems and also provided over the dbus interface. The widget has a bug where, if the first icon is a local icon, it will not check for icons provided over dbus. This results in the "fallback" icon being displayed instead of the correct icon.

Fixes #5122

### DIFF
--- a/libqtile/widget/helpers/status_notifier/statusnotifier.py
+++ b/libqtile/widget/helpers/status_notifier/statusnotifier.py
@@ -259,7 +259,11 @@ class StatusNotifierItem:  # noqa: E303
             if not self.icon:
                 self.icon = self._get_xdg_icon(icon_name)
 
-        if not self.icon and fallback:
+        if fallback:
+            for icon in ["Icon", "Attention", "Overlay"]:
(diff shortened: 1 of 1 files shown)


elParaguayo:
Verified working here: https://github.com/qtile/qtile/issues/5122#issuecomment-2558185697

PR #5129, opened 2024-12-21
Nix flake dependencies fix

Gurjaka:
Last commit about flake tried to migrate Qtile flake from dbus-next to dbus-fast, however I didn't take correct approach which resulted in having both dbus packages and unexpected behaviors. This commit tries to fix all that, before opening this pr I checked if everything worked as expected, even checked Qtile dependencies after built and seems like everything should be fixed, however I still demand careful review by @jwijenbergh and @Sigmanificient !!!

image

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -26,16 +26,10 @@ self: final: prev: {
             # use the source of the git repo
             src = ./..;
             # for qtile migrate, not in nixpkgs yet
-            propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [
-              pprev.libcst
-              pprev.dbus-fast
-            ];
-            dependencies = prev.lib.filter (x: x != pprev.dbus-next) (old.propagatedBuildInputs or [ ]);
+            propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pprev.libcst ];
           }
         )).override
-          {
-            wlroots = prev.wlroots_0_17;
-          };
+          { wlroots = prev.wlroots_0_17; };
     })
   ];
   python3 =
--- a/docs/manual/install/nixos.rst
+++ b/docs/manual/install/nixos.rst
@@ -124,7 +124,3 @@ In the development shell, there are a few useful things:
 
 - `qtile-run-tests-wayland`: Run all Wayland tests
 - `qtile-run-tests-x11`: Run all X11 tests
-
-Furthermore nix build can also be used to run NixOS VM tests:
-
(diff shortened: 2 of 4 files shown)


nix/overlays.nix
```diff
@@ -28,13 +28,12 @@ self: final: prev: {
...
             propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [
               pprev.libcst
-              pprev.dbus-fast
             ];
-            dependencies = prev.lib.filter (x: x != pprev.dbus-next) (old.propagatedBuildInputs or [ ]);
           }
         )).override
           {
             wlroots = prev.wlroots_0_17;
+            dbus-next = prev.python3Packages.dbus-fast;
```
elParaguayo, line 36: I'm not good on nix flakes but assigning the `dbus-fast` package to a variable called `dbus-next` doesn't seem like a great idea.

Gurjaka: I can't find any other way
If anyone knows I am happy to accept

jwijenbergh: Can't we not do it like you did previously with `old.dependencies`?

Gurjaka: > Can't we do it like you did previously but with `old.dependencies`?

That doesn't work, we don't have influence on `old.dependencies`, tried dozens of times

jwijenbergh: Maybe better to upstream the dbus-fast changes and update the flake?

jwijenbergh: Maybe just copy over the whole dependencies list? It's only up until we make a new release anyways

Gurjaka: > Maybe just copy over the whole dependencies list? It's only up until we make a new release anyways

But what's the point if it can be done this way?

jwijenbergh: Because overriding dbus-next is hacky

Gurjaka: Well it's for you to decide

Gurjaka: But keep in mind, doesn't matter if you set dependencies in overlay, since it gets ignored for some reason and still uses default package dependencies

Gurjaka: I tried setting dependencies list to empty list, but it still built with all the dependencies listed in Qtile package

Gurjaka: So far this is only way I actually got rid of dbus-next, if you have ideas you can try on my fork, since you are added as collaborator, I'll check changes in the morning

Gurjaka:
@jwijenbergh @Sigmanificient

PR #5132, opened 2024-12-22
Use float state setters in group.add()

### LINKED ISSUE #5130: X11: Initially fullscreen mpv is not resized/positioned correctly
### Issue description

When using mpv with the configuration option `fullscreen=yes` and the (default) qtile configuration setting of `auto_fullscreen = True`, the mpv window is initially rendered centered on the screen in the original video resolution, i.e. the window is not resized or repositioned to appear fullscreen. This should be true for other applications that want to start in fullscreen mode, but it might not and I don't have any other applications to test at the moment.
I have a bar at the top and at the bottom. If the video resolution is the same as the screen resolution the video is rendered starting below the top bar and exceeding the screen at the bottom by the same margin (hiding the bottom bar).
Pressing `f` to toggle (leave) fullscreen correctly maximizes the window between the two bars, pressing `f` to toggle fullscreen again correctly goes into fullscreen, hiding both bars.

What I have identified as (one of) the areas in the code that cause this issue is

https://github.com/qtile/qtile/blob/5fe89de8343d0e7f6b88468b712b1dbfaa83efc5/libqtile/group.py#L255-L275

In particular, the line

```python
win._float_state = FloatStates.FULLSCREEN
```

This directly changes the internal `_float_state`, skipping the relevant [setter code](https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/window.py#L1803). The setter code is also not called from [update_state()](https://github.com/qtile/qtile/blob/master/libqtile/backend/x11/window.py#L659) because of the check `if getattr(self, attr) != val` – the getter for `fullscreen` already returns `True`.

Simply changing the above-mentioned line to

```python
win.fullscreen = True
```

fixes the issue for me, mpv is correctly spawned in fullscreen mode covering the whole screen.

The question is: why are we directly setting the `_float_state` in `group.add()` instead of calling the setter? The same logic should apply to floating as well, is there a reson we do not simply use `win.floating = True`? That would also render the lines

https://github.com/qtile/qtile/blob/5fe89de8343d0e7f6b88468b712b1dbfaa83efc5/libqtile/group.py#L264-L265

unnecessary, because they are essentially replicated in the setter:

https://github.com/qtile/qtile/blob/5fe89de8343d0e7f6b88468b712b1dbfaa83efc5/libqtile/backend/x11/window.py#L1751-L1752

I can create a PR with these two changes (or at least with the fullscreen one), but I first wanted to hear your opinions and thoughts because for now I'm assuming there is a good reason why we skip the setters here.

Possibly related to a lot of other issues like #5043, #5034, #4058, #4176, #3814, #2415, #2163

### Version

0.29.0+fd6b2cd.flake

### Backend

X11 (default)

### Config

```python
auto_fullscreen = True
```

### Logs

_No response_

### Required

saviola777:
This replaces direct assignments to `win._float_state` with usage of the appropriate property setters in `group.add()`.

This required a change in the window.floating setter to avoid trying to
make windows floating which have not yet been placed.

I could not resist adding a small refactoring by replacing `if not win.floating or win.fullscreen:` with `else:`, but I can (re)move that (to a separate commit/PR) if you prefer.

Fixes #5130.

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -409,8 +409,8 @@ class Window(typing.Generic[S], _Base, base.Window, HasListeners):
     @floating.setter
     def floating(self, do_float: bool) -> None:
         if do_float and self._float_state == FloatStates.NOT_FLOATING:
-            if self.group and self.group.screen:
-                screen = self.group.screen
+            if self.is_placed():
+                screen = self.group.screen  # type: ignore[union-attr] # see is_placed()
                 if not self._float_width:  # These might start as 0
                     self._float_width = self._width
                     self._float_height = self._height
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -246,7 +246,7 @@ class Window(_Window, metaclass=ABCMeta):
 
     qtile: Qtile
 
-    # If float_x or float_y are None, the window has never floated
+    # If float_x or float_y are None, the window has never been placed
     float_x: int | None
     float_y: int | None
 
@@ -319,6 +319,15 @@ class Window(_Window, metaclass=ABCMeta):
     def has_user_set_position(self) -> bool:
         """Whether this window has user-defined geometry"""
         return False
+
+    def is_placed(self) -> bool:
+        """Whether this window has been placed, i.e. both float offsets are not None."""
+        return (
+            self.group is not None
+            and self.group.screen is not None
+            and self.float_x is not None
+            and self.float_y is not None
+        )
 
     def is_transient_for(self) -> WindowType | None:
         """What window is this window a transient window for?"""
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1728,7 +1728,7 @@ class Window(_Window, base.Window):
         tiled = [win.window.wid for win in (self.group.tiled_windows if self.group else [])]
         tiled_stack = [wid for wid in stack if wid in tiled and wid != self.window.wid]
         if do_float and self._float_state == FloatStates.NOT_FLOATING:
-            if self.group and self.group.screen:
+            if self.is_placed():
                 screen = self.group.screen
                 self._enablefloating(
                     screen.x + self.float_x,
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from libqtile import hook, utils
-from libqtile.backend.base import FloatStates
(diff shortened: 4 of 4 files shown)


saviola777:
Okay, calling the `floating` setter seems to break things, not sure why. Let me set up a local test environment (at least for X11) and see if I can understand what's happening.

saviola777:
I have now (locally) managed to get the tests to pass.

An issue that appeared after using the `floating` setter was that it assumed `float_x` and `float_y` to not be `None`, which is the case for newly created windows. I have added checks for `float_x` and `float_y` values.

Let me know what you think. Even though the tests pass I will first try to use this PR myself for a while to see if any issues arise.

tych0 (approved):

tych0:
Looks reasonable to me.

libqtile/backend/wayland/window.py
```diff
@@ -409,7 +409,12 @@ def floating(self) -> bool:
     @floating.setter
     def floating(self, do_float: bool) -> None:
         if do_float and self._float_state == FloatStates.NOT_FLOATING:
-            if self.group and self.group.screen:
+            if (
+                self.group
+                and self.group.screen
+                and self.float_x is not None
+                and self.float_y is not None
+            ):
```
tych0, line 417: Can you make this check into a function that returns a boolean since we use it twice? That will also force us to give it a reasonable name.

saviola777: Sure, I just checked around a little and found that `self.group and self.group.screen` is repeated quite often so might also be a candidate for a function.

But for this function you are suggesting (I would call it `has_float_offsets`?): should it include only `self.float_x is not None and self.float_y is not None` or also `self.group and self.group.screen`?

The only place where these float offsets are set is in implementations of `_Window.place`, so `is_placed` would be another possible name for the function, though there is one implementation of `place` (`LayerStatic`) where this is not true.

Should this function be added to `_Window` or `Window`?

tych0: Seems like Window, since that's the only place it's used. 

I think starting with one that includes all four of the tests in your if is good. We can refactor beyond that in future patches.

saviola777: I have updated it, but mypy is complaining now because it doesn't see that we have made sure that screen can't be `None`.

tych0:
Arg, that seems to have broken type inference somehow. Maybe chain the booleans? But that's annoying...

saviola777:
I have used a `# type: ignore` comment for now (I have no idea why it is only necessary in one of the two places), let me know what you think.

tych0:
Looks reasonable to me, thanks!

PR #5138, opened 2024-12-26
Add ability for horizontal text in vertical bar

elParaguayo:
Adds a new option `rotate=False` to text based widgets to prevent text from being rotated.


elParaguayo:
@tych0 there's a small bug in this (related to padding). I'm hoping to push a fix today.

PR #5139, opened 2024-12-27
Make cmd_obj print proper JSON

kaistian:
The output look like json but it's not valid JSON and make JSON tools like jq and gron fail.

This will possible break script people already have created.
But I think it would be worth it so we can use JSON tools to parse the output.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 Qtile x.xx.x, released xxxx-xx-xx:
+    !!! breaking changes !!!
+      - Output of `qtile cmd-obj` is now in JSON.
     * features
     * bugfixes
 
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -29,7 +29,7 @@ from __future__ import annotations
 
 import argparse
 import itertools
-import pprint
+import json
 import sys
 import textwrap
 
@@ -38,6 +38,12 @@ from libqtile.command.client import CommandClient
 from libqtile.command.graph import CommandGraphRoot
 from libqtile.command.interface import IPCCommandInterface
 from libqtile.ipc import Client, find_sockfile
+
+
+def set_to_list(obj):
+    if isinstance(obj, set):
+        return list(obj)
+    raise TypeError
 
 
 def get_formated_info(obj: CommandClient, cmd: str, args=True, short=True) -> str:
@@ -177,7 +183,7 @@ def cmd_obj(args) -> None:
         else:
             ret = run_function(obj, args.function, args.args)
             if ret is not None:
-                pprint.pprint(ret)
+                print(json.dumps(ret, indent=2, default=set_to_list))
     else:
         print_base_objects()
         sys.exit(1)
--- a/test/test_qtile_cmd.py
+++ b/test/test_qtile_cmd.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import json
 import os
 import re
 import subprocess
@@ -89,16 +90,18 @@ class ServerConfig(Config):
 server_config = pytest.mark.parametrize("manager", [ServerConfig], indirect=True)
 
 
-def run_qtile_cmd(args, no_eval=False):
+def run_qtile_cmd(args, no_json_loads=False):
     cmd = os.path.join(os.path.dirname(__file__), "..", "bin", "qtile")
     argv = [cmd, "cmd-obj"]
     argv.extend(args.split())
     pipe = subprocess.Popen(argv, stdout=subprocess.PIPE)
     output, _ = pipe.communicate()
     output = output.decode()
-    if no_eval:
+    if not output:
+        return False
+    if no_json_loads:
         return output
-    return eval(output)  # as returned by pprint.pprint
+    return json.loads(output)
 
 
 @server_config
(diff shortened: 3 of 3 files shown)


ervinpopescu:
Even before a maintainer enables the CI run, I can say with 100% confidence that this will fail in some cases. Take `eval` which can return objects (`Group`,`Bar` etc).

That said, maybe a more clever approach can help this be done, but I would rather have the JSON parsing out of the codebase.

I have a shell function for converting the output to some extent which I will share here later.

kaistian:
For it to be JSON it must have double quote and not single quote, and `True`/`False` most lower case `true`/`false`

So if I run this I get closer to JSON
`qtile cmd-obj -o cmd -f windows | tr "'" '"' | sed -e 's/True/true/g' -e 's/False/false/g'`

But then you have this
``` {'float_info': {'height': 2348, 'width': 3490, 'x': 2560, 'y': 1184},
  'floating': False,
  'fullscreen': False,
  'group': 'www',
  'height': 1150,
  'id': 94372033,
  'maximized': False,
  'minimized': False,
  'name': '[nftables] 6.4.\xa0Using sets in nftables commands | Red Hat '
          'Product Documentation — Mozilla Firefox',
  'width': 412,
  'wm_class': ['Navigator', 'Firefox'],
  'x': 2560,
  'y': 1184},
  ```
the `name:` is on several lines and the escape `\xa0` is not valid JSON escape.
This is not so easy to find a reliable work around for.

**Comment by elParaguayo** [comment]
`eval(...)` should probably just return a string. I'm not sure it's reasonable to expect a valid json response to that.

As for providing json formatted output, could this just be put behind a flag so as not to break current behaviour?

**Comment by tych0** [comment]
Can we parse the output of eval with the json library in the tests? It seems reasonable to me to want the output in json, but we want the tests to pass of course.

**Comment by kaistian** [comment]
> As for providing json formatted output, could this just be put behind a flag so as not to break current behaviour?

Locally I have added `-j / --json` with this patch
```patch
diff --git a/libqtile/scripts/cmd_obj.py b/libqtile/scripts/cmd_obj.py
index ece09c08..13144200 100644
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 import argparse
 import itertools
 import json
+import pprint
 import sys
 import textwrap

@@ -177,7 +178,10 @@ def cmd_obj(args) -> None:
         else:
             ret = run_function(obj, args.function, args.args)
             if ret is not None:
-                print(json.dumps(ret, indent=2))
+                if args.json:
+                    print(json.dumps(ret, indent=2))
+                else:
+                    pprint.pprint(ret)
     else:
         print_base_objects()
         sys.exit(1)
@@ -225,4 +229,5 @@ def add_subcommand(subparsers, parents):
         help="With both --object and --function args prints documentation for function.",
     )
     parser.add_argument("--socket", "-s", help="Use specified socket for IPC.")
+    parser.add_argument("--json", "-j", action='store_true', help="Print output as JSON.")
     parser.set_defaults(func=cmd_obj)
```

tych0:
IMO we don't need the flag and we can just change the output, but we'll need to fix the tests. Probably worth adding a note to the change log as well.

elParaguayo:
> IMO we don't need the flag and we can just change the output, but we'll need to fix the tests. Probably worth adding a note to the change log as well.

Yes, my suggestion was a bad one, there's no reason to have text output by default.

kaistian:
I have tried to change the test, had problem to test `-o group` locally because the result/output was empty.
But when I commented group out all other test passed.

kaistian:
A little back and fourth with all the `push --force`, but this is a cleaner solution.

kaistian:
I see the test fails, I'm not a developer so bear with me, I think this is the same as I see local
it's failing with the `group`, but `window` and `screen` is OK.
When running group it's output is a empty sting.
```python
     for obj in ["window", "group", "screen"]:
         assert run_qtile_cmd(f"-s {manager.sockfile} -o {obj} -f info")
```
It's not easy to see where because of JSON, and need some additional print to see what is happening.

I'll try a more in-depth debugging with print to find out what is happening.

elParaguayo:
OK. The test error is actually hiding the true source of the problem. The test fails because it can't do `json.loads` and it can't do `json.loads` because it's actually receiving an empty string.

So, why is it receiving an empty string? That's because some of the values we send include `set`s but a `set` is not a serializable object for JSON so we've got to convert them to a `list` first.

This diff should do it:
```diff
diff --git a/libqtile/scripts/cmd_obj.py b/libqtile/scripts/cmd_obj.py
index ece09c08..d4c443bf 100644
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -40,6 +40,12 @@ from libqtile.command.interface import IPCCommandInterface
 from libqtile.ipc import Client, find_sockfile


+def set_to_list(obj):
+    if isinstance(obj, set):
+        return list(obj)
+    raise TypeError
+
+
 def get_formated_info(obj: CommandClient, cmd: str, args=True, short=True) -> str:
     """Get documentation for command/function and format it.

@@ -177,7 +183,7 @@ def cmd_obj(args) -> None:
         else:
             ret = run_function(obj, args.function, args.args)
             if ret is not None:
-                print(json.dumps(ret, indent=2))
+                print(json.dumps(ret, indent=2, default=set_to_list))
     else:
         print_base_objects()
         sys.exit(1)
diff --git a/test/test_qtile_cmd.py b/test/test_qtile_cmd.py
index 7a0a69f2..f04074c3 100644
--- a/test/test_qtile_cmd.py
+++ b/test/test_qtile_cmd.py
@@ -97,6 +97,8 @@ def run_qtile_cmd(args, no_json_loads=False):
     pipe = subprocess.Popen(argv, stdout=subprocess.PIPE)
     output, _ = pipe.communicate()
     output = output.decode()
+    if not output:
+        return False
     if no_json_loads:
         return output
     return json.loads(output)
```

kaistian:
Ah I understand the problem know now, I have update the PR to include your changes.
Thank you for the help @elParaguayo

PR #5142, opened 2024-12-28
Bugfix for non-rotated text in vertical bar

elParaguayo:
Non-rotated text in vertical bars was being positioned incorrectly.

Fixed 5141


elParaguayo:
@tych0 I've asked someone on Discord to confirm this works for them. I'll merge it once they've confirmed it works (and CI passes).

PR #5146, opened 2025-01-01
Add border_on_single option for spiral layout.

rauaap:
Adds "border_on_single" option for the spiral layout, similar to what other layouts have.

### DIFF
--- a/libqtile/layout/spiral.py
+++ b/libqtile/layout/spiral.py
@@ -94,6 +94,7 @@ class Spiral(_SimpleLayoutBase):
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
         ("border_width", 1, "Border width."),
+        ("border_on_single", True, "Draw border when there is only one window."),
         ("margin", 0, "Margin of the layout (int or list of ints [N E S W])"),
         ("ratio", 1 / GOLDEN_RATIO, "Ratio of the tiles"),
         (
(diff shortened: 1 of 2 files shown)


libqtile/layout/spiral.py
```diff
@@ -94,6 +94,7 @@ class Spiral(_SimpleLayoutBase):
         ("border_focus", "#0000ff", "Border colour(s) for the focused window."),
         ("border_normal", "#000000", "Border colour(s) for un-focused windows."),
         ("border_width", 1, "Border width."),
+        ("border_on_single", False, "Draw border when there is only one window."),
```
elParaguayo, line 97: Should we set this as `True`? By setting it to `False`, the default behaviour will change for people using this layout as the border is currently drawn for all windows.

libqtile/layout/spiral.py
```diff
@@ -188,13 +189,19 @@ def configure(self, win, screen):
             bc = self.border_normal
 
         (x, y, w, h), margins = self._fix_double_margins(x, y, w, h)
+        num_windows = len(self.clients)
+
+        if num_windows == 1 and not self.border_on_single:
```
elParaguayo, line 194:
```suggestion
        if len(self.clients) == 1 and not self.border_on_single:
```
Don't think we need a separate variable as you're only checking it once.

elParaguayo (commented):
Thanks for this.

Have left a couple of comments.

We should update the tests too.

libqtile/layout/spiral.py
rauaap: Change request implemented.

libqtile/layout/spiral.py
rauaap: Change request implemented.

tych0:
Is it possible to commonize this border_on_single layout logic at all?

rauaap:
> Is it possible to commonize this border_on_single layout logic at all?

From what I can see the layouts that have implemented this (Bsp, Colums, Tile) use different ways to determine if there's only a single client or not. If there was a unified way for layouts to determine this, it shouldn't be too difficult.

tych0:
Yeah, makes sense. It has always annoyed me that the layouts all do this differently, but not enough to fix it :)

tych0 (approved):

PR #5149, opened 2025-01-04
Enhancements to Minimize, Unminimize and Unhide Functionality

mahdi739:
This pull request addresses several issues related to the minimizing, unminimizing, and unhiding of applications when invoked from system tray icons or via web browsers:

### Handling Configure Requests:
When an application (e.g., Telegram) is launched from a browser, a configure request is generated. Previously, this was only applied to windows in the current group, potentially leaving applications in other groups unhandled. This PR ensures that such requests are processed regardless of the group, and in stack mode, ensures hidden windows are unhidden.

### Systray Application Group Handling:
When opening an application from a systray icon, the application should open in the current group rather than switching to the group where the application was last active. The `set_group()` method has been replaced with `togroup()` to ensure new windows open in the current group.

### Focus Management Post-Minimization:
Minimizing an application should automatically shift focus to the last unminimized window to prevent unexpected behavior, especially in stack mode. This PR corrects the focus not changing from the minimized window.

### State Change on Reinvocation:
If an application is minimized and then called again from another program (e.g., `rofi`, `wmctrl`, `alttab`), it should both become visible and have its minimized state reverted. This PR ensures that the application's state is updated to unminimized when it receives focus.

**Note:** Some of these changes might cause existing tests to fail. I believe this indicates that the current test behavior needs adjustment to reflect these improvements.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -1694,7 +1694,7 @@ class Window(_Window, base.Window):
         self._wm_class: list[str] | None = None
         self.update_wm_class()
         self.update_name()
-        self.set_group()
+        self.togroup()
 
         # add window to the save-set, so it gets mapped when qtile dies
         qtile.core.conn.conn.core.ChangeSaveSet(SetMode.Insert, self.window.wid)
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -2080,15 +2080,14 @@ class Window(_Window, base.Window):
         else:
             width, height, x, y = self.width, self.height, self.x, self.y
 
-        if self.group and self.group.screen:
-            self.place(
-                x,
-                y,
-                width,
-                height,
-                self.borderwidth,
-                self.bordercolor,
-            )
+        self.place(
+            x,
+            y,
+            width,
+            height,
+            self.borderwidth,
+            self.bordercolor,
+        )
         self.update_state()
         return False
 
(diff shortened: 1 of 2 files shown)


libqtile/backend/x11/window.py
```diff
@@ -2080,16 +2084,16 @@ def handle_ConfigureRequest(self, e):  # noqa: N802
...
-        if self.group and self.group.screen:
-            self.place(
-                x,
-                y,
-                width,
-                height,
-                self.borderwidth,
-                self.bordercolor,
-            )
+        self.place(
```
elParaguayo, line 2087: Couldn't this move windows between groups as you're now ignoring whether or not the window is in a visible group? That can't be desirable behaviour/

libqtile/backend/x11/window.py
```diff
@@ -1694,7 +1694,7 @@ def __init__(self, window, qtile):
         self._wm_class: list[str] | None = None
         self.update_wm_class()
         self.update_name()
-        self.set_group()
+        self.togroup()
```
elParaguayo, line 1697: I appreciate you're doing this for `Systray` but this will impact every window created this way. Every window would end up ignoring requests from the server to be placed on a different group or, when it's a transient, displaying with its parent.

libqtile/group.py
```diff
@@ -232,6 +232,7 @@ def focus(self, win, warp=True, force=False):
                 self.floating_layout.blur()
                 for layout in self.layouts:
                     layout.focus(win)
+            win.minimized = False
```
elParaguayo, line 235: Why is this needed?

elParaguayo (commented):
I'm not convinced by some of the changes here as they could have a large impact on qtile's behaviour.

@tych0 any thoughts?

PR #5159, opened 2025-01-16
[Wayland] Improve blurry bar on HiDPI displays

richcarni:
Relevant issue: #3646

Same solution as proposed in #4625, except instead of relying on a single config scale value, attempts to obtain Wayland scale factor for each screen. Also allows fractional scaling

This fix works by using Cairo to scale up the bar (and other internal windows) instead of Wayland. This works for text and Cairo (vector) drawing

The difference without and with this fix:

comparison_small

Limitations: Doesn't fix scaling of images such as icons

### DIFF
--- a/libqtile/backend/base/core.py
+++ b/libqtile/backend/base/core.py
@@ -95,7 +95,15 @@ class Core(CommandObject, metaclass=ABCMeta):
         """A context manager to suppress window events while operating on many windows."""
         yield
 
-    def create_internal(self, x: int, y: int, width: int, height: int) -> Internal:
+    def create_internal(
+        self,
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        *,
+        scale: float | None = None,
+    ) -> Internal:
         """Create an internal window controlled by Qtile."""
         raise NotImplementedError  # Only error when called, not when instantiating class
 
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -123,6 +123,7 @@ class _Window(CommandObject, metaclass=ABCMeta):
         above=False,
         margin=None,
         respect_hints=False,
+        scale=None,
     ):
         """Place the window in the given position."""
 
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -1557,9 +1557,17 @@ class Core(base.Core, wlrq.HasListeners):
     def flush(self) -> None:
         self._poll()
 
-    def create_internal(self, x: int, y: int, width: int, height: int) -> base.Internal:
+    def create_internal(
+        self,
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        scale: float | None = 1.0,
+    ) -> base.Internal:
         assert self.qtile is not None
-        internal = window.Internal(self, self.qtile, x, y, width, height)
+        scale = scale if scale is not None else 1.0
+        internal = window.Internal(self, self.qtile, x, y, width, height, scale)
         self.qtile.manage(internal)
         return internal
 
--- a/libqtile/backend/wayland/drawer.py
+++ b/libqtile/backend/wayland/drawer.py
@@ -75,9 +75,16 @@ class Drawer(drawer.Drawer):
         if height > self._win.height - offsety:
             height = self._win.height - offsety
 
+        # Despite checks for None above, mypy requires these aserts
+        assert width is not None
+        assert height is not None
+
+        scale = self._win.scale
+
         # Paint recorded operations to our window's underlying ImageSurface
         with cairocffi.Context(self._win.surface) as context:
             context.set_operator(cairocffi.OPERATOR_SOURCE)
+            context.scale(scale, scale)
             # Adjust the source surface position by src_x and src_y e.g. if we want
             # to render part of the surface in a different position
             context.set_source_surface(self.surface, offsetx - src_x, offsety - src_y)
@@ -85,7 +92,9 @@ class Drawer(drawer.Drawer):
             context.fill()
 
         damage = PixmanRegion32()
-        damage.init_rect(offsetx, offsety, width, height)
+        damage.init_rect(
+            int(offsetx * scale), int(offsety * scale), int(width * scale), int(height * scale)
+        )
         # TODO: do we really need to `set_buffer` here? would be good to just set damage
         self._win._scene_buffer.set_buffer_with_damage(self._win.wlr_buffer, damage)
         damage.fini()
--- a/libqtile/backend/wayland/output.py
+++ b/libqtile/backend/wayland/output.py
@@ -135,7 +135,8 @@ class Output(HasListeners):
 
     def get_screen_info(self) -> ScreenRect:
         width, height = self.wlr_output.effective_resolution()
-        return ScreenRect(int(self.x), int(self.y), width, height)
+        scale = self.wlr_output.scale
+        return ScreenRect(int(self.x), int(self.y), width, height, scale)
 
     def organise_layers(self) -> None:
         """Organise the positioning of layer shell surfaces."""
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -927,7 +927,16 @@ class Internal(_Base, base.Internal):
     Internal windows are simply textures controlled by the compositor.
     """
 
-    def __init__(self, core: Core, qtile: Qtile, x: int, y: int, width: int, height: int):
+    def __init__(
+        self,
+        core: Core,
+        qtile: Qtile,
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        scale: float,
+    ):
         self.core = core
         self.qtile = qtile
         self._wid: int = self.core.new_wid()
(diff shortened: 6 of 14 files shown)


richcarni:
@tych0 I haven't tried extending this to X11 but certainly have thought about it. Assuming it's not too difficult to get the display scale from xrandr (I think that's where it would come from?), I think it would just be a matter of scaling `ScreenRect` so that it reports the effective/logical screen resolution. And also mapping mouse click locations from physical to logical screens

I'm guessing the reason you wouldn't want this in X11 yet is that currently the up-scaling blurs images. For me, this affects icons in the StatusNotifier widget. For Wayland I feel this is tolerable (for now) because the alternative (current situation) is that everything is blurry when using screen scaling!

At the moment I'm experimenting with solutions for the blurry icons in StatusNotifier, which I think will boil down to using librsvg for SVGs and somehow being scale aware when working with pixmap icons

elParaguayo:
We load SVGs via the `cairocffi.ImageSurface` class. This uses `librsvg` for svg support. You probably need to trick the widget into thinking it wants a bigger icon.

richcarni:
> You probably need to trick the widget into thinking it wants a bigger icon.

Yes, this! I successfully tested this in the logic where it picks the best icon size from the list of available sizes

I think it could work for SVGs too, but I've been playing around with rsvg handles and the nice thing is you don't have to worry about display scaling (I'm not sure if it converts the SVG to Cairo primitives or just delays rasterisation). With a helper class the syntax could look something like this:
```
rsvg_handle = rsvgcffi.RsvgHandle.new_from_file(path)
rsvg_handle.render_document(self.drawer.ctx, (xoffset, yoffset, self.icon_size, self.icon_size))
```

elParaguayo:
I'd be very reluctant to include a new dependency for this. Really feels like it should be doable with cairo.

richcarni:
The dependency is still librsvg (https://gnome.pages.gitlab.gnome.org/librsvg/Rsvg-2.0/method.Handle.render_cairo.html) unless you were also referring to having another helper class

elParaguayo:
Your code example has `rsvgcffi` in it. I'm not familiar with that library.

richcarni:
Sorry, my bad! That's the helper/wrapper class (in the style of `libqtile.pangocffi`)

elParaguayo:
Ok, but we're trying to limit the number of those classes as much as possible. So, if we can fix this in cairocffi, rather than have a new helper class, that would be preferable.

richcarni:
Maybe I created some confusion with talk of solutions to some of the issues this PR doesn't address - namely blurry icons, and tiny gaps on the bar between widgets (fractional scaling) - but I (possibly incorrectly) assumed it would be better to deal with those in subsequent PRs

Same for extending scaling to also work on the X11 backend

I certainly appreciate this isn't a trivial PR in the sense that once we choose a path for Wayland hidpi support, this will probably also affect X11 hidpi support, how widgets will need to be modified to support hidpi, etc.

Would it be helpful to increase the scope so everyone can see what the complete implementation would look like? Or for me to draw up some kind of roadmap?

Please let me know if there is more I can do

(I use Qtile with Wayland as my daily driver - I'm incredibly impressed by how complete and stable it is. For me hidpi is an essential feature and to that end, I would like to continue helping in whatever way I can)

tpaniaki:
For blurry icons, can't we easily apply the `scalefactor` variable inside StatusNotifier icons settings ?

richcarni:
> For blurry icons, can't we easily apply the `scalefactor` variable inside StatusNotifier icons settings ?

True, this is easy enough. But then this also affects qtile_extras.widget.StatusNotifier. And I don't even know how many other official, let alone custom widgets might also need updating

Ideally, I would imagine the author of a widget shouldn't have to think about scale factors. But I can't see how to do this without some sort of abstraction on top of Cairo. Ultimately, I suspect (understandably) the core developers may feel a little uneasy about implementing this feature piece by piece, without a clear understanding of exactly how it will all work when complete

tych0:
> So, if we can fix this in cairocffi, rather than have a new helper class, that would be preferable.

I agree it *feels* like it should be possible using cairo, since cairo has svg rendering built in. It seems like maybe [this](https://github.com/Kozea/CairoSVG/commit/1a0eba6d92a599546d042dd0cabeaebc7481b770#diff-6df618583a245d89c3677cc1602472143bbd999778b40dcdf2704cb99ab7acfeR156-R215) is the magic incantation to do it in cairo directly?


If that doesn't work, maybe if we only import/dlopen librsvg when users actually want scaling, and use cairo's rendering by default? That way people only need to install librsvg if they want scaling. But ideally we could just do it with cairo.

richcarni:
> > So, if we can fix this in cairocffi, rather than have a new helper class, that would be preferable.
> 
> I agree it _feels_ like it should be possible using cairo, since cairo has svg rendering built in. It seems like maybe [this](https://github.com/Kozea/CairoSVG/commit/1a0eba6d92a599546d042dd0cabeaebc7481b770#diff-6df618583a245d89c3677cc1602472143bbd999778b40dcdf2704cb99ab7acfeR156-R215) is the magic incantation to do it in cairo directly?
> 
> If that doesn't work, maybe if we only import/dlopen librsvg when users actually want scaling, and use cairo's rendering by default? That way people only need to install librsvg if they want scaling. But ideally we could just do it with cairo.

I think cairo only has support for outputting to svg files built in? But CairoSVG looks interesting. It appears to be an svg converter, written in python, that parses svg to cairo and then uses cairo to output to different formats? I guess using it's core svg conversion might work. That would add a dependency on a python package rather than librsvg

I should just add that the StatusNotifier widget uses libqtile/images.py, which in effect uses GDK-PixBuf to decode images, including svg icons. However, GDK-PixBuf requries librsvg to read svg images - admittedly it's an optional dependency

Perhaps to clarify, is the issue adding a hard dependency on librsvg or adding another pangocffi helper file? Or both?! :)

richcarni:
> > Ideally, I would imagine the author of a widget shouldn't have to think about scale factors.
> 
> I agree - I think users and widget implementers should get DPI scaling "for free", short of manually being able to scale things as they want perhaps.
> 
> I have some raw xcffib code to do the x11 scaling, but it's quite ugly. The actual scaling call is here: https://gist.github.com/tych0/a0e2d802b4bde6476997051c576d0ea6#file-monitors-py-L151
> 
> I had intended to try to implement DPI scaling in qtile via the Screen() object, allowing people to add a scale_factor there, and it would do the right thing for folks, but that has long been blocked on stuff like #5020. The idea being that users know which monitor they want to scale by what factor, and qtile should do the rest for widgets, bars, images, etc.

Let me preface this comment by admitting I have near zero knowledge around x11s inner workings

I'm guessing Wayland makes some of this stuff easier. You set a scale for each display (using something like wlr-randr) and qtile can see this and scale it's internal windows appropriately. I've stored the scale factor in `ScreenRect` (it seemed to make sense). Because of the way Wayland does scaling, I don't think it makes sense to set this is in config. For example, if my global Wayland scale factor (wlr-randr) is 2.0 and I set my qtile scale factor to 10.0, qtile will render the bar at 10x resolution and then Wayland will downsample it to the dimensions given in the bar config (relative to the effective screen resolution, which is actual screen resolution / 2.0 (global scale factor) )

Kind of confusing trying to explain but from what I can gather seems quite different to x11

I'm guessing I'm currently in the minority of qtile users who are also on Wayland with a hidpi display. I'll keep working on my hidpi branch, maybe some users will start testing it and it gets some momentum. I've pretty much solved all the issues that bother me on a day to day basis - some more hackishly than other. Or maybe someone else comes up with a better solution that gets implemented - and that would be great too!

richcarni:
Apologies, amateur hour over here! 

Didn't mean to close this (was just attempting some housekeeping on my github)

tpaniaki:
> I'm guessing I'm currently in the minority of qtile users who are also on Wayland with a hidpi display. I'll keep working on my hidpi branch, maybe some users will start testing it and it gets some momentum.

I daily-drive your branch without any issue. Call me superficial but this blurry bar really is something I can't bear on the master. It is my understanding that any progress on wayland is somehow dependent on jwijenbergh's insights. Perhaps PRs are going to be merged more quickly once the backend will have been rewritten in C.

richcarni:
@tpaniaki I assume you're referring to the fix-blurry-bar branch. I also have: https://github.com/richcarni/qtile/tree/wayland-hidpi-dev
which includes a fix for wallpaper scaling, blurry statusnotifier icons and gaps between widgets (I think this last issue is only with fractional scaling)

(if you're using qtile-extras/statusnotifier, you'll also need https://github.com/richcarni/qtile-extras)

Consider it a work in progress... I'd welcome any feedback

tpaniaki:
>  I'd welcome any feedback

Brillant I hadn't followed your last commits. 

Icons are greatly improved but I cannot say why it's still slightly suboptimal. The ouput of papirus-dark:

swappy-20250322_201316

richcarni:
> > I'd welcome any feedback
> 
> Brillant I hadn't followed your last commits.
> 
> Icons are greatly improved but I cannot say why it's still slightly suboptimal. The ouput of papirus-dark:
> 
> swappy-20250322_201316

@tpaniaki, let's move this conversation to https://github.com/qtile/qtile/discussions/5215

PR #5160, opened 2025-01-18
allow for vertical text direction to be overriden

TheHighGroundWins:
Change keeps default behavior unless specified, won't affect other users negatively.
Allows for user to override vertical text direction if wanted.
Especially important for some writing systems like Mongolian Script, which must be rotated 90 degrees and read top to bottom.
As currently the text appears upside down in left bar.

Previously mentioned:
#5127

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -497,10 +497,12 @@ class _TextBox(_Widget):
         ),
         ("rotate", True, "Rotate text in vertical bar."),
         (
-            "vertical_text_direction",
-            "default",
-            "Override direction the text would read in, in a vertical bar"
-            "'default', 'top_to_bottom', 'bottom_to_top'",
+            "direction", 
+            "default", 
+            "Override the text direction in vertical bar, has no effect on text in horizontal bar."
+            "default: text displayed based on vertical bar position (left/right)"
+            "ttb: text read from top to bottom, btt: text read from bottom to top."
+            "'default', 'ttb', 'btt'",
         ),
     ]  # type: list[tuple[str, Any, str]]
 
@@ -583,6 +585,11 @@ class _TextBox(_Widget):
         _Widget._configure(self, qtile, bar)
         if self.fontsize is None:
             self.fontsize = self.bar.height - self.bar.height / 5
+        if self.direction != "default" and self.direction != "ttb" and self.direction != "btt":
+            logger.warning("""Invalid value set for direction: %s
+                               Valid values are: default, ttb, btt
+                               direction set to default""", self.direction)
+            self.direction = "default"
         self.layout = self.drawer.textlayout(
             self.formatted_text,
             self.foreground,
@@ -656,20 +663,20 @@ class _TextBox(_Widget):
             # Left bar reads bottom to top
             # Can be overriden to read bottom to top all the time with vertical_text_direction
             if (
-                self.bar.screen.left is self.bar
-                and self.vertical_text_direction == "default"
-                or self.vertical_text_direction == "bottom_to_top"
-            ):
+                (self.bar.screen.left is self.bar
+                and self.direction == "default")
+                or self.direction == "btt"
+                ):
                 self.drawer.ctx.rotate(-90 * math.pi / 180.0)
                 self.drawer.ctx.translate(-self.length, 0)
 
             # Right bar is top to bottom
             # Can be overriden to read top to bottom all the time with vertical_text_direction
             elif (
-                self.bar.screen.right is self.bar
-                and self.vertical_text_direction == "default"
-                or self.vertical_text_direction == "top_to_bottom"
-            ):
+                (self.bar.screen.right is self.bar
+                and self.direction == "default")
+                or self.direction == "ttb"
+                ):
                 self.drawer.ctx.translate(self.bar.width, 0)
                 self.drawer.ctx.rotate(90 * math.pi / 180.0)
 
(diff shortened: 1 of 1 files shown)


libqtile/widget/base.py
```diff
@@ -496,6 +496,12 @@ class _TextBox(_Widget):
             "Setting ``scroll_fixed_width=True`` will force the widget to have a fixed width, regardless of the size of the text.",
         ),
         ("rotate", True, "Rotate text in vertical bar."),
+        (
+            "vertical_text_direction", 
+            "default",
```
elParaguayo, line 501:
```suggestion
            "vertical_text_direction",
            "default",
```
Need to remove trailing spaces.

libqtile/widget/base.py
```diff
@@ -648,12 +654,22 @@ def draw(self):
 
         if not self.bar.horizontal and self.rotate:
             # Left bar reads bottom to top
-            if self.bar.screen.left is self.bar:
+            # Can be overriden to read bottom to top all the time with vertical_text_direction
+            if (
+                self.bar.screen.left is self.bar
+                and self.vertical_text_direction == "default"
+                or self.vertical_text_direction == "bottom_to_top"
```
elParaguayo, line 661:
```suggestion
                (self.bar.screen.left is self.bar
                and self.vertical_text_direction == "default")
                or self.vertical_text_direction == "bottom_to_top"
```
I think some parentheses might help readability here.

libqtile/widget/base.py
```diff
@@ -648,12 +654,22 @@ def draw(self):
...
                 self.drawer.ctx.rotate(-90 * math.pi / 180.0)
                 self.drawer.ctx.translate(-self.length, 0)
 
             # Right bar is top to bottom
-            else:
+            # Can be overriden to read top to bottom all the time with vertical_text_direction
+            elif (
+                self.bar.screen.right is self.bar
+                and self.vertical_text_direction == "default"
+                or self.vertical_text_direction == "top_to_bottom"
```
elParaguayo, line 671:
```suggestion
                (self.bar.screen.right is self.bar
                and self.vertical_text_direction == "default")
                or self.vertical_text_direction == "top_to_bottom"
```
As above.

elParaguayo (commented):
Can you add some code in `_configure` so that the `vertical_text_direction` is set to `default` if an unexpected value is provided. You should also include a log message saying what you've done so the user can see why they're getting unexpected behaviour.

TheHighGroundWins:
@elParaguayo Before I continue, is there anyway to make specifying the text direction shorter to type as a parameter?
I've used them in my own config, and while they work its a little long to write if you want different styles per widget unlikely but possible.

a shorter name of vertical_text_direction, like vert_text_dir for example
a shorter string for "top_to_bottom" could be "t_to_b". what do you think.

I don't want to decrease readability as everything else seems to be in this style for this project, but its a little impractical to have everything in long words.

elParaguayo:
I'd suggest just using `direction`. The description can make it clear that this has no impact in horizontal bars.

As for the strings, I think you could also accept "ttb" for top-to-bottom, and "btt" for bottom-to-top.

Just make the docs very clear on whatever you decide.

libqtile/widget/base.py
```diff
@@ -577,6 +585,11 @@ def _configure(self, qtile, bar):
         _Widget._configure(self, qtile, bar)
         if self.fontsize is None:
             self.fontsize = self.bar.height - self.bar.height / 5
+        if self.direction != "default" and self.direction != "ttb" and self.direction != "btt":
```
elParaguayo, line 588:
```suggestion
        if self.direction not in ("default", "ttb", "btt"):
```

elParaguayo (commented):
Thanks. This is looking good. One last recommended change...

TheHighGroundWins:
> Thanks. This is looking good. One last recommended change...
> 
> EDIT: Oh, I've just seen that the pre-commit checks are also failing. Ideally, you should install pre-commit on your system so the checks are run (and, where possible, automatically fixed) before committing. If you can't do that, please look at the error messages [here](https://github.com/qtile/qtile/actions/runs/13382851508/job/37382170757?pr=5160) and apply the fixes).

I see, I wasn't aware I could run these locally. Next time I make another pull request which I'm planning to do not anytime soon but eventually, I do that.

libqtile/widget/base.py
elParaguayo: You've marked this as resolved but the change hasn't been committed.

libqtile/widget/base.py
```diff
@@ -577,6 +585,14 @@ def _configure(self, qtile, bar):
...
+            logger.warning(
+                """Invalid value set for direction: %s
+                               Valid values are: default, ttb, btt
+                               direction set to default""",
+                self.direction,
+            )
```
elParaguayo, line 594:
```suggestion
            logger.warning(
                "Invalid value set for direction: %s. Valid values are: 'default', 'ttb', 'btt'. "
                "direction has been set to 'default'",
                self.direction,
            )
```
Don't think we need the log message to span over 3 lines.

elParaguayo (commented):
I've just tested this and it works well. Last few minor points and then we can merge.

If possible, please squash everything into a single commit (with a clear commit message) and force push. If not, I can squash the commits here when we merge.

TheHighGroundWins:
> I've just tested this and it works well. Last few minor points and then we can merge.
> 
> If possible, please squash everything into a single commit (with a clear commit message) and force push. If not, I can squash the commits here when we merge.

Yeah can you do that. I put the final commit message to be the overall commit message. And I don't know how to do this without messing up my git branches and everything.

PR #5162, opened 2025-01-21
docs: add a note about using uv for udev rules

tych0:
I finally gave uv a try, and it solves the udev sourcing problem quite nicely of course, among many other things. Let's move at least this small piece of docs toward uv.

Perhaps we should start recommending it elsewhere too, but I'll leave it at this for now.


elParaguayo:
Sorry - bit snowed with other stuff at the moment. Thanks for doing this.

tych0:
No worries!

PR #5166, opened 2025-01-24
update ruff

c0rydoras:


RFCreate:
Same as #5106 

I'm not 100% sure, but I think the problem is in:
https://github.com/qtile/qtile/blob/cc473155d81d376f80cf917ec17eeae2d475f734/libqtile/backend/base/window.py#L253-L258

PR #5170, opened 2025-01-29
Add tuned-manager widget

BladeOfEnat:
A widget to interact with the tuned power daemon.

It can:
- Show the  currently active profile
- Cycle through a list of user-defined profiles, with defaults that are typical to find on most linux distributions that bundle tuned.


libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,103 @@
...
+# SOFTWARE.
+import re
+import subprocess
+from os import system
+
+from libqtile import bar
+from libqtile.widget import base
+
+
+class TunedManager(base._TextBox):
```
tych0, line 28: Should this poll for updates as our other widgets do, in case something updates the configuration out of band of this widget?

BladeOfEnat: It might make sense for it to behave like that. Making the `tuned-adm active` call and stripping the name from the output is not the most light-weight operation though. It's not very ressource heavy either, but since this is my first commit, I decided to be very conservative with the amount of ressources the widget should need. It would probably make the widget a bit more useful.

Not polling for updates would rely on users not changing the power state without using the widget.

I can make changes to the widget to use a polling widget base class, too. I was simply not sure about this, as it is my first commit to qtile.

tych0: Yeah, no worries. When you say it's "not the most light-weight" operation, presumably it is a conversation over a unix socket to a local daemon? We have other widgets that do that (e.g. the mpris widgets talk over dbus, etc.), so I think there is precedent.

If it's more expensive than that, maybe we should re-evaluate.

BladeOfEnat:
I made some changes so the widget polls for updates now.

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+# SOFTWARE.
+import re
+import subprocess
+from os import system
+
+from libqtile import bar
+from libqtile.widget import base
+
+
+class TunedManager(base.InLoopPollText):
```
tych0, line 28: Can you make this threaded since it does external IO?

tych0:
Overall it looks pretty good to me. One minor change, can you clean up the git history as well?

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def update_bar(self):
+        self.current_mode = self.find_mode()
+        self.text = self.current_mode
+        self.bar.draw()
+
+    def execute_command(self, index: int):
+        argument = self.modes[index]  # pyright: ignore
+        # NOTE: For some reason, this does not work through subprocesses. It does work with os though. TODO: Fix.
+        result = system(f"tuned-adm profile {argument}")
```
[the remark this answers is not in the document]
elParaguayo: Couldn't we use `self.qtile.spawn` here, rather than calling `os.system` directly?

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+        base.InLoopPollText._configure(self, qtile, bar)
+        self.text = TunedManager.find_mode(self)
+
+    def poll(self):
+        return self.find_mode()
+
+    def find_mode(self):
+        result = subprocess.run("tuned-adm active", shell=True, capture_output=True, text=True)
+        output = result.stdout
+        regex = re.compile(r"Current active profile:\s+(\S+)")
```
elParaguayo, line 80: We don't need to compile this with every loop.

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+        if not regex.findall(output):
+            return ""
+        return regex.findall(output)[0]
```
elParaguayo, line 83: We don't need to call `findall` twice. Use a walruss expression or save the value in a variable before the if statement.

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def execute_command(self, index: int):
+        argument = self.modes[index]  # pyright: ignore
+        # NOTE: For some reason, this does not work through subprocesses. It does work with os though. TODO: Fix.
+        result = system(f"tuned-adm profile {argument}")
+        if result == 0:
+            self.update_bar()
+        else:
+            self.text = "Error setting mode"
+            self.bar.draw()
```
elParaguayo, line 98:
```suggestion
            self.update("Error setting mode")
```
The `update` method will handle redrawing calls.

libqtile/widget/tuned_manager.py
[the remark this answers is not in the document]
BladeOfEnat: > Can you elaborate on how it doesn't work?

Right, I forgot to look into this in more detail. I meant to fix this before pushing. This was more of a thing to test if the widget functions like that at all. I will look more into it.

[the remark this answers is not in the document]
BladeOfEnat: > Couldn't we use `self.qtile.spawn` here, rather than calling `os.system` directly?
> 
> EDIT: no, we can't. We need the return code...
> 
> EDIT 2: This feels like it should work with `subprocess.run(f"tuned-adm profile {argument}", check=True)` and then we just catch the `subprocess.CalledProcessError`.

Will try that

libqtile/widget/tuned_manager.py
BladeOfEnat: Willl make it threaded, yes

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def next_mode(self):
+        next_index: int = (self.modes.index(self.current_mode) + 1) % len(self.modes)  # pyright: ignore
+        self.execute_command(next_index)
+
+    def previous_mode(self):
+        prev_index: int = (self.modes.index(self.current_mode) - 1) % len(self.modes)  # pyright: ignore
+        self.execute_command(prev_index)
```
elParaguayo, line 106:
```suggestion
    def _change_mode(self, step=1);
        next_index: int = (self.modes.index(self.current_mode) + step) % len(self.modes)  # pyright: ignore
        self.execute_command(next_index)

    def next_mode(self):
        self._change_mode()

    def previous_mode(self):
        self._change_mode(step=-1)
```

test/widgets/test_tuned_widget.py
```diff
@@ -0,0 +1,75 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Widget specific tests
+
+from unittest.mock import MagicMock, patch
```
elParaguayo, line 23: It would be nice to have some tests that result in the qtile manager starting up and using the widget. We do this with other (patched) widgets so it would be good to do the same here.

And a widget screenshot too if possible: https://docs.qtile.org/en/stable/manual/howto/widget.html#screenshots

elParaguayo (commented):
Thanks for this.

I've added some suggestions but I don't think any of them are "must haves".

BladeOfEnat:
Huh. I'll look into it. Strange.

elParaguayo:
I think this was fixed in a recent commit so rebasing the PR should work.

BladeOfEnat:
Okay, let's see.

BladeOfEnat:
This should be good to merge then?

libqtile/widget/tuned_manager.py
```diff
@@ -0,0 +1,106 @@
...
+            ["powersave", "balanced-battery", "throughput-performance"],
+            "The modes to cycle through",
+        )
+    ]
+
+    def __init__(self, **config):
+        base.ThreadPoolText.__init__(self, "", **config)
+        self.add_defaults(TunedManager.defaults)
+        self.length_type = bar.CALCULATED
+        self.regex = re.compile(r"Current active profile:\s+(\S+)")
```
elParaguayo, line 58: Is the output always in English? If not, we should try to expose this to allow users to change it.

elParaguayo (commented):
Can you add a note to the Changelog to advertise the new feature.

BladeOfEnat:
I just verified it, it seems like TuneD does not provide any localization (The docs only exist in English and switching locale always still produced English results).

BladeOfEnat:
I edited the Changelog.

elParaguayo:
Looks good to me.

@tych0 Any other changes you want before we merge this?

tych0:
Looks like the fit history still needs some fixing, but the code itself looks good to me.

BladeOfEnat:
Aight, I'll get to rebasing once I get up tomorrow then.

BladeOfEnat:
Huh, I thought I fixed that conflict while rebasing...

BladeOfEnat:
Give me a moment

BladeOfEnat:
Ah, shit, I see what happened

Issue #5161, opened 2025-01-20
StatusNotifier widget failing to show some nm-applet icons


richcarni:
### Issue description

Quite frequently after resuming from sleep I notice that the wifi icon (nm-applet) stops somewhere in the middle of it's connecting animation

I'm pretty sure I've only seen this after a sleep/resume. It looks like this (if not clear, only the 3rd bar is illuminated), although it is random what icon it stops on:

![Image](https://github.com/user-attachments/assets/8b817f7d-8445-482b-bfa2-cd353b8b831e)

Been trying to debug, but don't really understand the inner workings of this widget

What I can see are successive calls to `_update_local_icon()` and then in `_get_local_icon()`:
```
            if not self.icon:
                self.icon = self._get_xdg_icon(icon_name)
```

`icon_name` is correct, but`self.icon` appears to still be set, so it never calls `_get_xdg_icon()`

I can't see any relevant errors or warnings in the log

### Version

0.30.1.dev1+gfecc37d8

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

richcarni:
Thanks @elParaguayo. On further examination, I think it's just updating the icon too quickly

I'm testing a fix, which cancels the running `_get_local_icon` task if `_update_local_icon` is called before it has completed

So far so good. I'll create a PR it stays that way

TheHighGroundWins:
I've had a similar problem, where the wifi icon gets stuck at disconnected even though it's connected.

And fixes itself if you change or try to connect to another network.

TheHighGroundWins:
> [@TheHighGroundWins](https://github.com/TheHighGroundWins) If you're comfortable installing from source or patching, you could test my branch: https://github.com/richcarni/qtile/tree/fix-statusnotifier-icon-update

I have my own separate pull request. So I'll try to merge your patch or clone it separately.

elParaguayo:
Can we close this now that we've merged a separate fix?

TheHighGroundWins:
> Can we close this now that we've merged a separate fix?
Agreed it's the same issue under a different name.

PR #5174, opened 2025-02-02
Fix statusnotifier when icons rapidly change

richcarni:
Certain apps (e.g. nm-applet) can rapidly change their status icon (e.g. when connecting to a WiFi) causing statusnotifier to display the wrong icon - #5161

To remedy: if _update_local_icon is called before a previous task to update the icon has completed, then cancel this running task and create a new one to process the latest update

### DIFF
--- a/libqtile/widget/helpers/status_notifier/statusnotifier.py
+++ b/libqtile/widget/helpers/status_notifier/statusnotifier.py
@@ -277,6 +277,7 @@ class StatusNotifierItem:  # noqa: E303
     def _update_local_icon(self):
         if self._get_local_icon_task is not None and not self._get_local_icon_task.done():
             self._get_local_icon_task.cancel()
+            self._get_local_icon_task.remove_done_callback(self._redraw)
             self._get_local_icon_task.add_done_callback(lambda task: self._update_local_icon())
         else:
             self.icon = None
@@ -365,10 +366,7 @@ class StatusNotifierItem:  # noqa: E303
 
         return arr
 
-    def _redraw(self, task):
-        if task.cancelled():
-            return
-
+    def _redraw(self, result):
         """Method to invalidate icon cache and redraw icons."""
         self._invalidate_icons()
         if self.on_icon_changed is not None:
--- a/libqtile/widget/helpers/status_notifier/statusnotifier.py
+++ b/libqtile/widget/helpers/status_notifier/statusnotifier.py
@@ -132,6 +132,7 @@ class StatusNotifierItem:  # noqa: E303
         self.icon_theme = icon_theme
         self.icon = None
         self.path = path if path else STATUSNOTIFIER_PATH
+        self._get_local_icon_task = None
 
     def __eq__(self, other):
         # Convenience method to find Item in list by service path
@@ -270,12 +271,16 @@ class StatusNotifierItem:  # noqa: E303
             self.icon = Img.from_path(path.resolve().as_posix())
 
     def _create_task_and_draw(self, coro):
-        task = create_task(coro)
-        task.add_done_callback(self._redraw)
+        self._get_local_icon_task = create_task(coro)
+        self._get_local_icon_task.add_done_callback(self._redraw)
 
     def _update_local_icon(self):
-        self.icon = None
-        self._create_task_and_draw(self._get_local_icon())
+        if self._get_local_icon_task is not None and not self._get_local_icon_task.done():
+            self._get_local_icon_task.cancel()
+            self._get_local_icon_task.add_done_callback(lambda task: self._update_local_icon())
+        else:
+            self.icon = None
+            self._create_task_and_draw(self._get_local_icon())
 
     def _new_icon(self):
         self._create_task_and_draw(self._get_icon("Icon"))
@@ -360,7 +365,10 @@ class StatusNotifierItem:  # noqa: E303
 
         return arr
 
-    def _redraw(self, result):
+    def _redraw(self, task):
+        if task.cancelled():
+            return
+
         """Method to invalidate icon cache and redraw icons."""
         self._invalidate_icons()
         if self.on_icon_changed is not None:
--- a/libqtile/widget/helpers/status_notifier/statusnotifier.py
+++ b/libqtile/widget/helpers/status_notifier/statusnotifier.py
@@ -132,7 +132,6 @@ class StatusNotifierItem:  # noqa: E303
         self.icon_theme = icon_theme
         self.icon = None
         self.path = path if path else STATUSNOTIFIER_PATH
-        self._get_local_icon_task = None
 
     def __eq__(self, other):
         # Convenience method to find Item in list by service path
@@ -254,10 +253,13 @@ class StatusNotifierItem:  # noqa: E303
             except (AttributeError, DBusError):
                 icon_path = None
 
+            icon = None
             if icon_path:
-                self.icon = self._get_custom_icon(icon_name, Path(icon_path))
+                icon = self._get_custom_icon(icon_name, Path(icon_path))
 
-            if not self.icon:
+            if icon:
+                self.icon = icon
+            else:
                 self.icon = self._get_xdg_icon(icon_name)
 
         if fallback:
@@ -271,17 +273,12 @@ class StatusNotifierItem:  # noqa: E303
             self.icon = Img.from_path(path.resolve().as_posix())
 
     def _create_task_and_draw(self, coro):
-        self._get_local_icon_task = create_task(coro)
-        self._get_local_icon_task.add_done_callback(self._redraw)
+        task = create_task(coro)
+        task.add_done_callback(self._redraw)
 
     def _update_local_icon(self):
-        if self._get_local_icon_task is not None and not self._get_local_icon_task.done():
-            self._get_local_icon_task.cancel()
-            self._get_local_icon_task.remove_done_callback(self._redraw)
-            self._get_local_icon_task.add_done_callback(lambda task: self._update_local_icon())
-        else:
-            self.icon = None
-            self._create_task_and_draw(self._get_local_icon())
+        self.icon = None
+        self._create_task_and_draw(self._get_local_icon())
 
     def _new_icon(self):
         self._create_task_and_draw(self._get_icon("Icon"))
--- a/libqtile/widget/helpers/status_notifier/statusnotifier.py
+++ b/libqtile/widget/helpers/status_notifier/statusnotifier.py
@@ -17,6 +17,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+from asyncio import current_task
 from collections.abc import Callable
 from contextlib import suppress
 from functools import partial
@@ -132,6 +133,7 @@ class StatusNotifierItem:  # noqa: E303
         self.icon_theme = icon_theme
         self.icon = None
         self.path = path if path else STATUSNOTIFIER_PATH
+        self.last_icon_name = None
 
     def __eq__(self, other):
         # Convenience method to find Item in list by service path
@@ -248,6 +250,13 @@ class StatusNotifierItem:  # noqa: E303
         # the app. We also don't want to do the recursive lookup with an empty
         # icon name as, otherwise, the glob will match things that are not images.
         if icon_name:
+            if icon_name == self.last_icon_name:
+                current_task().remove_done_callback(self._redraw)
+                return
+
+            self.last_icon_name = icon_name
+            self.icon = None
+
             try:
                 icon_path = await self.item.get_icon_theme_path()
             except (AttributeError, DBusError):
@@ -261,6 +270,8 @@ class StatusNotifierItem:  # noqa: E303
                 self.icon = icon
             else:
                 self.icon = self._get_xdg_icon(icon_name)
+        else:
+            self.icon = None
 
         if fallback:
             for icon in ["Icon", "Attention", "Overlay"]:
@@ -277,7 +288,6 @@ class StatusNotifierItem:  # noqa: E303
         task.add_done_callback(self._redraw)
 
     def _update_local_icon(self):
-        self.icon = None
         self._create_task_and_draw(self._get_local_icon())
 
     def _new_icon(self):
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
         - Spiral layout: `shuffle_up` and `shuffle_down` commands now shuffle the focused window instead of rotating all windows.
     * features
     * bugfixes
+        - Fix StatusNotifier displaying wrong icon when icon changes too quickly (e.g. nm-applet)
 
 Qtile 0.30.0, released 2025-01-06:
     !!! breaking changes !!!


elParaguayo:
Thanks. This looks good and makes sense to me.

libqtile/widget/helpers/status_notifier/statusnotifier.py
```diff
@@ -270,12 +271,16 @@ async def _get_local_icon(self, fallback=True):
...
-        task.add_done_callback(self._redraw)
+        self._get_local_icon_task = create_task(coro)
+        self._get_local_icon_task.add_done_callback(self._redraw)
 
     def _update_local_icon(self):
-        self.icon = None
-        self._create_task_and_draw(self._get_local_icon())
+        if self._get_local_icon_task is not None and not self._get_local_icon_task.done():
+            self._get_local_icon_task.cancel()
+            self._get_local_icon_task.add_done_callback(lambda task: self._update_local_icon())
```
elParaguayo, line 280: What's the purpose of adding this callback? Can't we carry on in the same call without a new one?

libqtile/widget/helpers/status_notifier/statusnotifier.py
```diff
@@ -360,7 +365,10 @@ def _reorder_bytes(self, icon_bytes):
 
         return arr
 
-    def _redraw(self, result):
+    def _redraw(self, task):
+        if task.cancelled():
```
elParaguayo, line 369: I agree this is probably sensible to include here but, presumably, you could also do `.remove_done_callback(self._redraw)` when you're cancelling the task.

elParaguayo (commented):
Thanks for this. Couple of comment/questions.

Could you also add a note to the changelog to say what bug you've fixed.

libqtile/widget/helpers/status_notifier/statusnotifier.py
richcarni: Yes, good question - hope this makes sense

The short answer is the purpose is to wait for the running task to complete (cancelled or otherwise) before creating the next

Why is that necessary? Because `cancel()` doesn't actually cancel the task but "arranges for a [CancelledError](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) exception to be thrown into the wrapped coroutine on the next cycle of the event loop."

Now, having gone through the documentation again, i think I've made a mistake and actually need to try/catch that exception like below (otherwise it's just waiting for the task to complete)

Which kind of leads to an alternative solution - just use a lock in `_get_local_icon()` so that only one task can be running it at a time - no need to keep track of tasks, it's really simple. While I think cancelling the running task, waiting for it to be cancelled, then starting the next is probably the more efficient approach, I'm not sure anyone would notice some millisecond delay in rendering an icon! (As long as the last icon is correct!!) Curious what you think

```
async def _get_local_icon(self, fallback=True):
    try:
        # Default to XDG icon
        # Some implementations don't provide an IconName property so we
        # need to catch an error if we can't read it.
        # We can't use hasattr to check this as the method will be created
        # where we've used the default XML spec to provide the object introspection
        icon_name = ""
        try:
            icon_name = await self.item.get_icon_name()
        except DBusError:
            return

        # We only need to do these searches if there's an icon name provided by
        # the app. We also don't want to do the recursive lookup with an empty
        # icon name as, otherwise, the glob will match things that are not images.
        if icon_name:
            try:
                icon_path = await self.item.get_icon_theme_path()
            except (AttributeError, DBusError):
                icon_path = None

            if icon_path:
                self.icon = self._get_custom_icon(icon_name, Path(icon_path))

            if not self.icon:
                self.icon = self._get_xdg_icon(icon_name)

        if fallback:
            for icon in ["Icon", "Attention", "Overlay"]:
                await self._get_icon(icon)
    
    except asyncio.CancelledError:
        return

        if not self.has_icons and fallback:
            # Use fallback icon libqtile/resources/status_notifier/fallback_icon.png
            logger.warning("Could not find icon for '%s'. Using fallback icon.", icon_name)
            path = Path(__file__).parent / "fallback_icon.png"
            self.icon = Img.from_path(path.resolve().as_posix())
```

libqtile/widget/helpers/status_notifier/statusnotifier.py
richcarni: 100%

BTW, in terms of making changes, I assume I should add a new commit? I read something on the contributing page about force pushing but wasn't entirely clear to me in what circumstances that is preferable

elParaguayo: When we merge this we'll want everything in a single commit. So, it's fine to make changes in a new commit now but, if you do, you'll need to squash everything into a single commit before merging. Or, if you're happy with the changes, force pushing an amended commit is also fine.

In this instance, I'd suggest a new commit so we can try your other proposal (above).

libqtile/widget/helpers/status_notifier/statusnotifier.py
elParaguayo: Might be worth a try.

richcarni:
After observing some odd results while trying to implement a lock, I spent some time looking at this more carefully and I think I've gotten to the root cause of this behaviour

I believe the problem is this bit of logic in `_get_local_icon()`:

```
if icon_path:
    self.icon = self._get_custom_icon(icon_name, Path(icon_path))

    if not self.icon:
    self.icon = self._get_xdg_icon(icon_name)
```

After carefully tracking each task, I found that when icon updates arrive too quickly, there is an issue around sharing `self.icon`

If we consider 2 icons in a sequence, icon 1 and icon 2, with the icon 2 update arriving shortly after icon 1

Icon 2 starts by setting `self.icon = None`. Icon 1, which is only a little ahead then updates self.icon. Now, when icon 2 gets to the logic above, `self.icon` is no not None, so it skips calling '_get_xdg_icon()'. Instead it goes ahead and effectively draws icon 1 again

Simplest solution is just to replace this logic with something not dependent on the state of `self.icon`:
```
icon = None
    if icon_path:
        icon = self._get_custom_icon(icon_name, Path(icon_path))

    if icon:
        self.icon = icon
    else:
        self.icon = self._get_xdg_icon(icon_name)
```

I have been running this just this change for a few days now without issue (no locks or task cancelling)


BTW, to give you some idea of the timing of updates for nm-applet:

```
Timestamp 	| Message serial # 	| Icon name
11:51:08.055 	| 781 			| nm-stage02-connecting05-secure
11:51:08.055	| 782 			| nm-stage02-connecting05-secure
11:51:08.112	| 791 			| nm-stage02-connecting06-secure
11:51:08.132	| 792 			| nm-stage02-connecting06-secure
11:51:08.156	| 801 			| nm-stage02-connecting07-secure
11:51:08.156 	| 802 			| nm-stage02-connecting07-secure
11:51:08.193 	| 809 			| nm-stage03-connecting08-secure
11:51:08.215 	| 810 			| nm-no-connection-secure
11:51:08.234 	| 816 			| nm-no-connection-secure
11:51:08.234 	| 817 			| nm-no-connection-secure
11:51:08.245 	| 827 			| nm-no-connection-secure
11:51:08.259 	| 828 			| nm-signal-75-secure
11:51:08.270 	| 833 			| nm-signal-75-secure
11:51:08.270 	| 834 			| nm-signal-75-secure
11:51:08.299 	| 846 			| nm-signal-75-secure
11:51:08.310 	| 847 			| nm-signal-75-secure
```

Most of the duplicates seem to be related to having a vpn connected. Might be a good idea to ignore duplicates, but if they aren't causing an issue I thought that might be outside the scope of this PR

elParaguayo:
I'm happy for you to implement whatever works!

With regards to those messages, are they all `NewIcon` signals or are they something else (like `PropertiesChanged`)?

richcarni:
Yeah, they are all `NewIcon`. Maybe it's emitting for both the WiFi and VPN connections? But it's not causing any issues as far as I can tell

Having cycled through a number of solutions now, I'm happy enough with the latest changes I pushed - it's the keep it simple solution

So next I need to squash into a single commit?

elParaguayo:
Can you make it so we only update if the icon has actually changed. We shouldn't respond to every signal if there's no change.

Don't worry about squashing yet - let's just get a working solution. We can then tidy up the commit history.

richcarni:
To verify repeat/duplicate NewIcon updates are being ignored I logged tasks (by id) at start of `_get_local_icon()` and in the corresponding `redraw()` callback

You can see there is now only one redraw for each unique change of icon

If you're wondering why I moved `self.icon = None`. I wanted to avoid setting it to None if that task is going to be terminated. Hypothetically, if tasks are running concurrently I could see that clearing self.icon set by the preceding task

I have been running with this for a couple of days without issue

Before the latest commit:

```
2025-02-12 11:00:39,438 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-131 | icon_name: nm-stage02-connecting08-secure
2025-02-12 11:00:39,439 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-132 | icon_name: nm-stage02-connecting08-secure
2025-02-12 11:00:39,439 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-131
2025-02-12 11:00:39,446 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-132
2025-02-12 11:00:39,499 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-133 | icon_name: nm-stage03-connecting09-secure
2025-02-12 11:00:39,500 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-134 | icon_name: nm-stage03-connecting09-secure
2025-02-12 11:00:39,582 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-135 | icon_name: nm-no-connection-secure
2025-02-12 11:00:39,582 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-136 | icon_name: nm-no-connection-secure
2025-02-12 11:00:39,582 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-133
2025-02-12 11:00:39,582 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-134
2025-02-12 11:00:39,595 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-135
2025-02-12 11:00:39,595 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-136
2025-02-12 11:00:39,595 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-137 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,596 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-138 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,596 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-139 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,596 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-140 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,596 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-141 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,596 WARNING libqtile statusnotifier.py:_get_local_icon():L252 task: Task-142 | icon_name: nm-signal-75-secure
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-137
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-138
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-139
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-140
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-141
2025-02-12 11:00:39,605 WARNING libqtile statusnotifier.py:_redraw():L369 task: Task-142
```

With commit:
```
2025-02-12 11:28:31,409 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-139 | icon_name: nm-stage02-connecting08-secure
2025-02-12 11:28:31,409 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-140 | icon_name: nm-stage02-connecting08-secure
2025-02-12 11:28:31,409 WARNING libqtile statusnotifier.py:_redraw():L378 task: Task-139
2025-02-12 11:28:31,455 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-141 | icon_name: nm-stage03-connecting09-secure
2025-02-12 11:28:31,455 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-142 | icon_name: nm-stage03-connecting09-secure
2025-02-12 11:28:31,482 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-143 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,482 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-144 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,482 WARNING libqtile statusnotifier.py:_redraw():L378 task: Task-141
2025-02-12 11:28:31,497 WARNING libqtile statusnotifier.py:_redraw():L378 task: Task-143
2025-02-12 11:28:31,512 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-145 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,513 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-146 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,513 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-147 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,513 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-148 | icon_name: nm-no-connection-secure
2025-02-12 11:28:31,514 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-149 | icon_name: nm-signal-100-secure
2025-02-12 11:28:31,514 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-150 | icon_name: nm-signal-100-secure
2025-02-12 11:28:31,521 WARNING libqtile statusnotifier.py:_redraw():L378 task: Task-149
2025-02-12 11:28:31,556 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-151 | icon_name: nm-signal-100-secure
2025-02-12 11:28:31,556 WARNING libqtile statusnotifier.py:_get_local_icon():L253 task: Task-152 | icon_name: nm-signal-100-secure
```

elParaguayo:
Thanks. It still looks like we're getting multiple calls for the same icon. Am I reading that right?

Also, you'll need to rebase and clear the conflict in changelog.

However, looks like we're getting close to merging this too.

richcarni:
You are reading it right. Because it has to fetch the icon_name before it can check whether it's the same icon. If it is, it removes the redraw callback and ends that task

TheHighGroundWins:
Any updates on this pull requests?
Doesn't seem like there are any merge issues, are there still some unresolved issues?

richcarni:
Sorry if I've messed something up - I can't see anything about merge conflicts on my end

I think I need to squash the commits though, if we're happy with where this is at?

elParaguayo:
Easier than I thought: The conflict was in changelog in an earlier commit. You fixed this in a later commit so, while this blocked rebasing, squashing was fine.

Thanks for your work on this.

richcarni:
No worries! Thanks for merging

Not sure if it's worth worrying about or if we can do anything about it now, but the commit message isn't quite correct

```
Fix statusnotifier when icons rapidly change (https://github.com/qtile/qtile/pull/5174)
Certain apps (e.g. nm-applet) can rapidly change their status icon (e.g. when connecting to a WiFi) causing statusnotifier to display the wrong icon - https://github.com/qtile/qtile/issues/5161

To remedy: if _update_local_icon is called before a previous task to update the icon has completed, then cancel this running task and create a new one to process the latest update.

We also only update the icon if a new icon name is requested.
```

The "To remedy" paragraph would more accurately be something like
```
Fix: Modify _get_local_icon logic. Specifically, avoid checking against self.icon, because it may be modified by another concurrent task
```

elParaguayo:
Can't change the message now.

PR #5180, opened 2025-02-13
Fix PusleVolume widget breaking after reload config

### LINKED ISSUE #5111: PulseVolume widget shows 0% volume when reloading Qtile config
### Issue description

Hello,
The PulseVolume widget doesn't seem to be able to update the volume % and is stuck at 0% when reloading the Qtile configuration. No changes are required to the configuration simply reloading it will cause the volume to display at 0% and not change.
The below error is shown in the logs:
`WARNING libqtile pulse_volume.py:get_sink_info():L116 Could not get info for default sink`
The volume control actually works though, I can use my scrollwheel to change the volume.

### Version

### Backend

X11 (default)

### Config

```python
widget.PulseVolume(
                    fmt="  {}",
                    padding = 3,
                    mouse_callbacks={'Button1': lazy.group['scratchpad'].dropdown_toggle('pavucontrol')}
                    ),
```

### Logs

```text
WARNING libqtile pulse_volume.py:get_sink_info():L116 Could not get info for default sink`
```

### Required

richcarni:
Fixes #5111. Using xrandr/wlr-randr causes the widget to reconfigure resulting in duplicate subscriptions, which aren't correctly unsubscribed from when reloading config

To prevent duplicate subscriptions, callbacks list is replaced by a set

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
     * features
         - Add TunedManager Widget to show the currently active power profile and cycle through a list of configurable power profiles
     * bugfixes
+        - PulseVolume widget: fix breakage with xrandr/wlr-randr + reload config #5111
 
 Qtile 0.30.0, released 2025-01-06:
     !!! breaking changes !!!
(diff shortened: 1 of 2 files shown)


elParaguayo:
Thanks for this. I assume changing the list of callbacks to a set has fixed the problem then?

CI is failing but that is unrelated to your changes.

richcarni:
Yep, working without issue for me now

tych0:
Hmm... I re-ran again, let's see...

PR #5182, opened 2025-02-14
Fix PusleVolume widget not following changes of the default sink

CharString:
The default sink might have changed since we setup the connection e.g. after (dis)connection of a Bluetooth headphone.
This resets the default sink before each action.

### DIFF
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -96,6 +96,10 @@ class PulseConnection:
         async for event in self.pulse.subscribe_events("sink", "server"):
             # Sink events will signify volume changes
             if event.facility == "sink":
+                # There's been a change to available sinks
+                # Update default sink details before querying the sink
+                if event.t in ("new", "remove"):
+                    await self.get_server_info()
                 await self.get_sink_info()
             # Server events include when the default sink changes
             elif event.facility == "server":


elParaguayo:
It's interesting that this is required as the widget is subscribed to server events and so should update whenever the default sink changes.

I've never actually experienced an issue either. Can you explain a bit more about what the issue looks like?

CharString:
> I've never actually experienced an issue either. Can you explain a bit more about what the issue looks like?

@elParaguayo Well, the widget became unresponsive and stay at a %. It appeared to happen after using a bluetooth headphone or connecting to a docking station/monitor via usb-c. But maybe it's sleep/hibernation related. It's on my laptop.

To be honest, I've just updated my qtile and this patch I had still open on my checkout against the v0.28.1 tag. But I doubt going from dbus-next to dbus-fast, changed anything.

> It's interesting that this is required as the widget is subscribed to server events and so should update whenever the default sink changes.

Hmmm. But when I change the volume using pavucontrol, the widget doesn't change along.

Feel free to close the PR. I had an intermittent issue, and this looked like it fixed it. But it looks like I need to look into this a bit further.

elParaguayo:
@CharString have you been able to test the above diff?

CharString:
Yes and it's a lot more responsive. Thanks!

PR #5186, opened 2025-02-15
Fix screensplit hooks

elParaguayo:
The ScreenSplit layout unsubscribes from hooks when it is hidden.
However, the `hide` method can be called again after the layout is
hidden.

Also gets rid of a rogue `print` statement...

### DIFF
--- a/libqtile/layout/screensplit.py
+++ b/libqtile/layout/screensplit.py
@@ -122,7 +122,6 @@ class ScreenSplit(Layout):
         self.layouts = {}
         self._move_win = None
         self._has_matches = None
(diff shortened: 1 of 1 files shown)


nazar-pc:
Applied on top of 0.30.0, restarted and appears to be working fine, thanks!

elParaguayo:
Thanks. Have set this to merge one CI passes.

nazar-pc:
This PR wasn't mentioned in the release notes for some reason, but seems like an important bugfix (at least to me :slightly_smiling_face:)

PR #5191, opened 2025-02-20
core: close /dev/null during spawn()

### LINKED ISSUE #5189: `lazy.spawn()` leaks file handles to `/dev/null`
### Issue description

Starting with qtile `0.30`, it leaks one file descriptor every time `lazy.spawn()` is called. This does not reproduce with qtile `0.29`.

Eg. add something like this to your `config.py`:

```py
...
Key([mod], "Return", lazy.spawn(terminal), desc="Launch terminal"),
...
```

And every time you trigger that `lazy.spawn()` call, an extra file handle to `/dev/null` is held in memory and never released.

Here's a diff of `lsof` output before and after triggering the above `lazy.spawn()`:

```diff
 COMMAND  PID   USER  FD      TYPE             DEVICE SIZE/OFF    NODE NAME
 ...
+qtile   1339 asdf    91r      CHR                1,3      0t0       4 /dev/null
```

This eventually causes qtile to fail to run `lazy.spawn()` since it will run out of file handles. Restarting qtile is required.

### Version

0.30.1.dev0+g6e31cf27.d20250210

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

tych0:
d'oh, rookie mistake: we open but do not close /dev/null during spawn, thus leaking it.

Closes #5189

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1324,25 +1324,25 @@ class Qtile(CommandObject):
             env.pop("VIRTUAL_ENV", None)
 
         # std{in,out,err} should be /dev/null
-        null = os.open("/dev/null", os.O_RDONLY)
-        file_actions: list[tuple] = [
-            (os.POSIX_SPAWN_DUP2, 0, null),
-            (os.POSIX_SPAWN_DUP2, 1, null),
-            (os.POSIX_SPAWN_DUP2, 2, null),
-        ]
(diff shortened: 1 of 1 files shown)


elParaguayo:
LGTM

tych0:
This one is bad enough that it might even be worth a release...

PR #5195, opened 2025-02-28
Swap groups between screens

kovasap:
A cleaned up version of https://github.com/qtile/qtile/pull/2381


libqtile/core/manager.py
```diff
@@ -1440,6 +1440,18 @@ def switch_groups(self, namea: str, nameb: str) -> None:
             for w in group.windows:
                 w.group = group
 
+    @expose_command()
+    def swap_screens(self, reverse = False):
```
tych0, line 1444: `swap` here is really not correct, it's more like `shift`? And we're shifting groups, not screens, IIUC. (We already have `toggle_group()`, which sounds similar to this function name.)

I'm wondering if we can add this as an option to `toggle_group(shift=True)`? But boolean parameters are bad, so maybe it makes sense to have this as a separate function.

Issue #5201, opened 2025-03-07
CheckUpdates not showing available updates on Fedora


e-devnull:
Hi,

On Fedora 41 **dnf list updates** returns nothing even if updates are available. Fedora 41uses dnf5 instead of dnf4, the fix is to replace **updates** with **--upgrades**. It works fine on both dnf4 and dnf5.

PR #5202, opened 2025-03-07
fix for CheckUpdates widget on fedora #5201

e-devnull:
This PR fix updates not showing on Fedora 41. Maintains compatibility between dnf4 and dnf5.


elParaguayo:
Docs build failure is unrelated to this PR.

tych0:
> Just so I understand, was the --upgrades" syntax already available in dnf4and so the only change indnf5is to discontinue support forupdates`?

Yes, looks like that's the case:

```
~/packages/qtile master dnf4 list --upgrades
Last metadata expiration check: 0:00:04 ago on Fri 07 Mar 2025 01:59:55 PM MST.
```

PR #5205, opened 2025-03-08
feat: support single pixel buffer protocol (wayland)

otaj:
Hi, this adds support for `SinglePixelBuffer` protocol. It is not a very widely known or used protocol, however, it is used by a lovely tool called `chayang` (https://git.sr.ht/~emersion/chayang).

Since the support for the protocol is already in `pywlroots`, the change is very simple.

I tested this locally and `chayang` now works (which it didn't before). I don't know if there's any other way how to properly test a support for a particular protocol.

Also, I'm sorry I didn't check before opening the PR whether it would be welcome, but I figured, that this is such a trivial change, that it should be non-contentious.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Qtile x.xx.x, released xxxx-xx-xx:
     * features
+        - Add support for `single-pixel-buffer` protocol for wayland 
     * bugfixes
 
 Qtile 0.31.0, released 2025-03-07:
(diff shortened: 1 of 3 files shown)


otaj:
Thanks for the pointer on how to fix the docs build!

tych0:
I'm going to go ahead and land this. If it turns out to be problematic we can always revert before a release, but I agree with @elParaguayo, I don't think it'll cause a problem. Thanks!

jwijenbergh:
Yeah this was fine to be merged, thank you. Sorry for the late reply

PR #5208, opened 2025-03-10
Fix: PyGObject dependecy requieres libgirepository-2.0-dev

RFCreate:


elParaguayo:
Can you provide some context for needing this change? Did CI and the docs start failing? Did you just want to upgrade CI to 24.04?

I don't mind the change but would to be good to understand the reason.

RFCreate:
I have another open PR #5106 and after pushing the latest changes, I started to get this error in the CI.

```
Build type: native build
Project name: pygobject
Project version: 3.52.1
C compiler for the host machine: cc (gcc 13.3.0 "cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0")
C linker for the host machine: cc ld.bfd 2.42
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program python3 found: YES (/home/runner/work/qtile/qtile/.tox/py310-x11/bin/python)
Found pkg-config: YES (/usr/bin/pkg-config) 1.8.1
Run-time dependency python found: YES 3.10
Found CMake: /usr/local/bin/cmake (3.31.6)
Run-time dependency girepository-2.0 found: NO (tried pkgconfig and cmake)
Not looking for a fallback subproject for the dependency girepository-2.0 because:
Use of fallback dependencies is disabled.

../meson.build:31:9: ERROR: Dependency 'girepository-2.0' is required but not found.
```

elParaguayo:
Don't think that's directly related to your PR. Looks like a new version of `PyGObject` was released yesterday so it's probably that.

tych0 (approved):

PR #5209, opened 2025-03-11
Pre-commit: update dependencies to latest version

RFCreate:
Update formatter and type checker to latest available versions. All of the changes where made for ruff and mypy pre-commit to pass.


tych0:
Thanks!

Issue #5210, opened 2025-03-13
Windows not spawning on musl after update to Qtile version 0.31.0


thomas-jacobson:
### Issue description

After updating from version 0.29.0 to version 0.31.0 on Void linux (musl), I cannot get windows to spawn. This is apparently unrelated to anything in my config file, as the behavior is the same even if I force Qtile to revert to the built-in default config.

One of the Void maintainers [suggested](https://www.reddit.com/r/voidlinux/comments/1ja2qwm/comment/mhltgbk/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) that this is probably due to `posix_spawn_file_actions_addclosefrom_np` only being available on glibc. Relevant log file output is pasted below.

Version 0.29.0 still seems to work fine for now. This must be a new issue starting from version 0.30.0 or 0.31.0. 

Just wanted to bring this to the Qtile devs' attention in case you all were not already aware of this issue.

### Version

0.31.0

### Backend

X11 (default)

### Config

```python

```

### Logs

```text
2025-03-13 09:20:03,598 ERROR libqtile core.py:_xpoll():L362 Got an exception in poll loop
Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 337, in _xpoll
    self.handle_event(event)
    ~~~~~~~~~~~~~~~~~^^^^^^^
  File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 304, in handle_event
    ret = target(event)
  File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/window.py", line 1570, in handle_KeyPress
    self.process_key_press(keysym)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.13/site-packages/libqtile/bar.py", line 585, in process_key_press
    self._has_keyboard.process_key_press(keycode)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/libqtile/widget/prompt.py", line 699, in process_key_press
    handle_key()
    ~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/libqtile/widget/prompt.py", line 621, in _send_cmd
    self.callback(self.user_input)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
[...]

PR #5211, opened 2025-03-15
fix(core): check if os.POSIX_SPAWN_CLOSEFROM is supported

classabbyamp:
Instead of checking if the API *may* exist based on Python version heuristics, check if the API *does* exist, as some platforms (namely musl on Linux) do not support `os.POSIX_SPAWN_CLOSEFROM` in their `pthread` API.

fixes: #5210


tych0 (approved):

tych0:
Thanks!

PR #5212, opened 2025-03-17
core: drop sys import

tych0:
Not sure how this skipped past our PR checks, maybe a new version of ruff or something? Anyway, we can drop this import.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -30,7 +30,6 @@ import shlex
 import shutil
 import signal
 import subprocess
-import sys
 import tempfile
 from collections import defaultdict
 from logging.handlers import RotatingFileHandler


elParaguayo:
Thanks.

CI is failing because Wayland gitlab downloads are offline until 22 March.

jwijenbergh:
> > Wayland gitlab downloads are offline until 22 March.
> 
> ha, I wonder if we should point at their releases page? But it looks like there's a bunch of stuff we download from there, probably it's okay to just wait.

Can't we cache these artifacts? I'm on phone so haven't checked if we do this already

tych0:
We don't, but we probably could once they're back online.

tych0:
Seems like these might be back early, let me try this again...

Issue #5194, opened 2025-02-25
Qtile.current_screen and group not updated when screen is removed


saviola777:
### Issue description

I ran an xrandr command which removed the screen that was focussed. On the one remaining other screen, I then tried to open a ScratchPad (see config below) and then spawn a terminal using `lazy.spawn(guess_terminal())` both of which apparently opened in the group that was last focused on the removed screen, which I only realized later. To me it looked like both had no effect and so I assumed the screen wasn't focused.

So I tried to change the focus using `lazy.next_screen()`, which did yield some logs (see below) but did not help. Then I tried to focus the screen (which showed an empty group) by clicking with the mouse, which had no effect (I guess because click-focus only works if there's a window?).

What worked in the end was spawning rofi (which is configured to spawn on the focussed screen) with `lazy.spawn()` which seemed to focus the screen, and from there spawn a terminal window.

Maybe this is a simple fix for someone who knows the code, if not I don't mind digging in and preparing a PR, for now I don't know the code well enough to be confident I could do it quickly (by myself).

Maybe somewhat related to #3993.

### Version

0.30.0+e34b2ac.flake

### Backend

X11 (default)

### Config

```python
follow_mouse_focus = False

# ScratchPads
groups.append(
    ScratchPad("terminal", [
        DropDown("scratch", "kitty sh -c \"tmux new-session -A -s s\"", height=0.65),
        ],
    )
)

keys.append(
    Key(
        'M-<Space>', lazy.group['terminal'].dropdown_toggle('scratch')
    )
)

#[…]
    Key("M-<Comma>", lazy.next_screen(), desc="Switch to next screen"),
    Key("M-<Return>", lazy.spawn(terminal), desc="Launch terminal"),
    Key(
        "A-<Space>",
        lazy.spawn("rofi -disable-history -show run -modes run")
    ),

#[…]
```

### Logs

```text
2025-02-25 08:41:24,668 ERROR libqtile manager.py:process_key_event():L482 KB command error next_screen: Traceback (most recent call last):
  File "/nix/store/sj5qskny340bv7pxwsq3b752bjxhfxsv-python3.12-qtile-0.30.0+e34b2ac.flake/lib/python3.12/site-packages/libqtile/command/interface.py", line 431, in call
    return SUCCESS, cmd(*args, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^
[...]

Issue #5197, opened 2025-03-03
Bar does not exist after disconnecting X11 screen


holocronweaver:
### Issue description

When I disconnect an X11 display, the bar object ceases to exist until I reconnect the display.

1. Add a bar to your config and reload config.
1. Note bar appears on solo display as expected.
1. Connect a second display.
1. Note that the bar is added to the new screen as expected.
1. Disconnect the second display.
1. Note the bar ceases to exist on the remaining display (takes up no space on screen).
1. Reconnect the second display.
1. Note the bar is created on both screens as expected.

I confirmed the bar object is destroyed after display disconnect by adding a keybinding to `lazy.hide_show_bar(position='all', screen='all')` and seeing this log warning:
```
2025-03-03 10:41:54,325 WARNING libqtile manager.py:hide_show_bar_screen():L1767 Not found bar for hide/show.
```

I tried changing which display is set as primary prior to disconnect and it seems to make no difference.

I don't believe this has ever worked correctly for me, but I could be mistaken.

* Workaround: exit Qtile after disconnecting second display, then start it again. `restart()` Qtile command is not sufficient.
* OS: Kubuntu 24.04.2
* Possibly related: #5165 (fairly different setup from me where only one display is on at a time whereas mine uses both displays at once; unclear if bar object is destroyed as in my case; also reconnecting the display does not resolve their issue but it does for me)

### Version

0.29.1.dev159+gf3fa1204 (commit e34b2acb5bb2cab7)

### Backend

X11 (default)

### Config
I removed my `@hook.subscribe.screens_reconfigured` hook and the behavior persists.

```python
@hook.subscribe.screens_reconfigured
def screens_reconfigured():
    logger.warning(f'Display count after reconfigure: {hw.system.displays.count}')
    qtile.reload_config()
```
### Logs

```text
2025-03-03 10:39:54,707 WARNING libqtile config.py:screens_reconfigured():L56 Display count after reconfigure: 1
2025-03-03 10:39:55,181 ERROR libqtile loop.py:_handle_exception():L62 Exception in event loop:
Traceback (most recent call last):
  File "/usr/lib/python3.12/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/home/jesse/apps/qtile/libqtile/core/manager.py", line 991, in f
    func(*args)
  File "/home/jesse/apps/qtile/libqtile/widget/base.py", line 702, in draw
[...]

PR #5183, opened 2025-02-15
qtile has no attribute `current_window`

elParaguayo:
This is a quick fix for an issue where qtile complains that there's no `current_window` attribute.

Fixes #5135

Issue #5135, opened 2024-12-24
AttributeError: 'libqtile.core.manager.Qtile' has no attribute current_window


sashah86:
### Issue description

# Firstly, thank you for your work and contributions.

I have been using Qtile for a few years and am very happy with it. From time to time, I have encountered some issues, but I was usually able to fix them myself.

Recently, I encountered a minor issue:

In my logs, I saw the following error:
 ## AttributeError: <class 'libqtile.core.manager.Qtile'> has no attribute current_window 

_In my humble opinion, the fix for this issue is to add a try/except block or hasattr check._

**Location of the Issue:**
libqtile/backend/x11/core.py

**Current Code:**
```
771     def handle_DestroyNotify(self, event) -> None:  # noqa: N802
772         assert self.qtile is not None
773
774         self.qtile.unmanage(event.window)
775         self.update_client_lists()
776         if self.qtile.current_window is None:
777             self.conn.fixup_focus()
```
**Proposed code:**
```
771     def handle_DestroyNotify(self, event) -> None:  # noqa: N802
772         assert self.qtile is not None
773
774         self.qtile.unmanage(event.window)
775         self.update_client_lists()
776         try:
777             if self.qtile.current_window is None:
778                 self.conn.fixup_focus()
779         except AttributeError:
780             self.conn.fixup_focus()
```

**Alternative proposed code:**
```
771     def handle_DestroyNotify(self, event) -> None:  # noqa: N802
772         assert self.qtile is not None
773
774         self.qtile.unmanage(event.window)
775         self.update_client_lists()
776         if not hasattr(self.qtile, 'current_window') or self.qtile.current_window is None:
777             self.conn.fixup_focus()
```

### Version

0.29.1.dev57+g6dedf4dc

### Backend

X11 (default)

### Logs

```text
16 2024-12-24 00:18:37,278 ERROR libqtile core.py:_xpoll():L362 Got an exception in poll loop
  17 Traceback (most recent call last):
  18   File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 337, in _xpoll
  19     self.handle_event(event)
  20     ~~~~~~~~~~~~~~~~~^^^^^^^
  21   File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 304, in handle_event
  22     ret = target(event)
  23   File "/usr/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 776, in handle_DestroyNotify
[...]

PR #5218, opened 2025-04-06
Screen change notification fix v2

tych0:
This is a repeat of #5020, since I can't seem to re-open that one after it's been closed and I force pushed to it (what? why add this restriction...)

It has an additional patch to fix the one failing test, which it turns out has been a bug for a long time.

Finally, let me tag the bugs I think this fixes, so people can test: #4967 #5135 #5183 #5197 and possibly others.


tych0:
hmm. coveralls seems busted?

ervinpopescu:
@tych0 maybe #4673 is helped by this? I can give it a test, but I'd rather let you chime in first.

tych0:
Yeah, looks like it could be related, I'd be curious what your testing shows.

tych0:
And it looks like coveralls is down for scheduled maintenance: https://status.coveralls.io/?utm_source=embed

libqtile/config.py
```diff
@@ -457,6 +457,16 @@ def __init__(
         self.height = height if height is not None else 0
         self.previous_group: _Group | None = None
 
+    def __eq__(self, other: object) -> bool:
```
elParaguayo, line 460: Where do we actually test for equivalence?

ervinpopescu:
Unfortunately it looks like https://github.com/qtile/qtile/issues/4673 is not ameliorated by this.

That said, just staying on X11 for the 10-15 minutes it took to plug/unplug my monitors and try to reproduce, I got this sudden feeling I need to get back on Wayland asap...

Please don't take this personally, I love your work on X11, but the simple fact that my 2 monitors with 60 fps cannot live together with my 90 fps without feeling sluggish and is crazy territory to me, so it does look like I can't go back to X :)

saviola777:
Looks like this also fixes #5194

libqtile/config.py
tych0: I put the link in the commit message, but this whole function probably deserves a comment noting where, I'll move the discussion to a comment here.

tych0:
> Please don't take this personally, I love your work on X11

Ha, no worries. Some day I will switch to wayland :)

saviola777:
This seems to introduce a new bug. When I have two screens (laptop internal screen + external monitor) with different resolutions, and I disable the first screen (laptop screen), which basially moves the qtile screen 1 from the laptop screen to the external monitor, then qtile seems to still use the laptop screen resolution for that screen (xrandr shows the correct resolution, so it's not that the external monitor has a wrong mode). It looks like somewhere in qtile the resolution might not be updated? This is not the case when I do the same thing on git master.

tych0:
> It looks like somewhere in qtile the screen dimensions might not be updated? This is not the case when I do the same thing on git master.

Hmm. Can you post logs from `query_crtcs()` to show the geometry, and output from `xrandr -q`?

In principle disabled monitors should be skipped by this code: https://github.com/tych0/qtile/blob/87267ff585b1bc70ec7c835f4c3c37a9ccc64185/libqtile/backend/x11/xcbq.py#L414-L415

But I can imagine that they're not somehow, which would generate the behavior you describe. You can log `info.name.to_string()` there to get the same names that xrandr uses.

tych0:
One other question, when you disable your laptop monitor, do you make the external one `--primary` as well? IIRC that had some bugs. Qtile doesn't do that automatically, but it could.

(My end goal is to make all of this stuff configurable via qtile, so you don't need xrandr at all, but of course that takes a lot more code :)

tych0:
ping @saviola777 , can you comment on the `--primary` bit? I suspect that'll fix your issue, and I can write the code in qtile to do it, but I want to make sure it really fixes things before I do.

saviola777:
Sorry for the delay. Yeah I did not explicitly make the monitor `--primary`, I thought that would be done automatically when there's only one monitor left.

I noticed the problem with my laptop in the office and have been at home since then where I don't use a laptop… that's why I haven't gotten around to testing it again.

I tried to make the monitor primary, but it had no effect (i.e. I still saw the same behavior described above):

```bash
# First separately
❯ xrandr --output eDP-1 --off
❯ xrandr --output HDMI-1 --primary
# Then both at the same time; no difference
❯ xrandr --output eDP-1 --left-of HDMI-1 --preferred
❯ xrandr --output eDP-1 --off --output HDMI-1 --preferred --primary
```

saviola777:
> Hmm. Can you post logs from `query_crtcs()` to show the geometry […]

Can I do this using the qtile CLI, or do I need to modify the code to include logging?

tych0:
> Can I do this using the qtile CLI, or do I need to modify the code to include logging?

changing the source is probably the easiest; but better yet, you can just run this script when your display is in the "strange" mode:

```
~ cat tester.py
#!/usr/bin/env -S uv run --with xcffib --script
import os

from pprint import pprint

import xcffib
import xcffib.xproto
import xcffib.randr

conn = xcffib.connect(os.environ["DISPLAY"])
randr = conn(xcffib.randr.key)

roots = conn.get_setup().roots
if len(roots) > 1:
    log.warning("more than one root window, assuming first is root")
rootid = roots[0].root

sr = randr.GetScreenResources(rootid).reply()

for output in sr.outputs:
    info = randr.GetOutputInfo(output, xcffib.CurrentTime).reply()
    if info.connection != xcffib.randr.Connection.Connected:
        continue
    if not info.crtc:
        continue
    crtc_info = randr.GetCrtcInfo(info.crtc, xcffib.CurrentTime).reply()
    pprint(info.name.to_string())
    pprint(info.__dict__)
    pprint(crtc_info.__dict__)
```

> I'll try to work on your other questions regarding the logs later today… or later this week, more likely! Sorry.

Yep, no worries. Thanks for reporting the bug and helping debug!

saviola777:
Okay I ran `xrandr -q` and your script both before and after triggering the strangeness, I tried to get your script to actually print something useful for the xcffib.List objects, but I failed.

Two things I noticed is that the mouse is restricted to the visible area, and calling `lazy.next_screen()` seems to at least change the `qtile.current_screen.index` variable (I have a widget showing marking the active screen using this variable), so there seems to be a second screen object? Keyboard and mouse focus remain on the visible screen.
 Maybe this is normal since I have two screens defined in my config, I have never tried calling `lazy.next_screen()` on a single monitor setup before.

[tester_output_after.txt](https://github.com/user-attachments/files/19772563/tester_output_after.txt)
[tester_output_before.txt](https://github.com/user-attachments/files/19772564/tester_output_before.txt)
[xrandr_q_output_after.txt](https://github.com/user-attachments/files/19772565/xrandr_q_output_after.txt)
[xrandr_q_output_before.txt](https://github.com/user-attachments/files/19772566/xrandr_q_output_before.txt)

tych0:
> Two things I noticed is that the mouse is restricted to the visible area, and calling lazy.next_screen() seems to at least change the qtile.current_screen.index variable (I have a widget marking the active screen using this variable), so there seems to be a second screen object? Keyboard and mouse focus remain on the visible screen.

Hmm, that is odd. We should only be using the screens that physically exist based on this code:
https://github.com/tych0/qtile/blob/87267ff585b1bc70ec7c835f4c3c37a9ccc64185/libqtile/core/manager.py#L386-L390

So I wonder if it's that the get_screen_info() call really sees two screens when the event is triggered, and that's why qtile reasons about it incorrectly. Can you add some logging there and see?

My guess is that we are reconfiguring "too early", so we still see the old screen configuration when we query it. I'll have to think about how to fix that (I would have expected xrandr to Grab the root window...).

> I tried to get your script to actually print something useful for the xcffib.List objects, but I failed.

yeah, I noticed that too, which is annoying. this should help, if you want to play with it:

https://github.com/tych0/xcffib/pull/183

saviola777:
Okay, well I'm confused. I've added the logging and the result is: `reconfigure_screens` is called exactly once, when qtile starts, and then never again, i.e. the `screen_change` hook is not fired when changing stuff with xrandr.

This explains the issues I'm seeing. But looking at your code changes, I don't really know what could have introduced this change in behavior. But maybe you do. 😅

tych0:
Oh, do you have the hook subscription in your config like we note here:

https://docs.qtile.org/en/stable/manual/ref/hooks.html#libqtile.hook.subscribe.screen_change

? (I wonder if I should just do that subscription automatically in this PR.)

saviola777:
Yeah, I have `reconfigure_screens = True` in my config.

tych0:
Hmmm. If you add the hook explicitly does it work? Do you get the event on git master?

tych0:
Yeah, exactly. But this must be something I broke with this branch, let me play around with the it.

libqtile/backend/x11/core.py
```diff
@@ -185,10 +185,7 @@ def finalize(self) -> None:
         self.conn.finalize()
 
     def get_screen_info(self) -> list[ScreenRect]:
-        ps = self.conn.pseudoscreens
-        if self.qtile:
-            self._xpoll()
```
saviola777, line 190: Is the removal of this call to _xpoll perhaps a problem?

tych0: No, that's one of the goals of the patch series though.

We call get_screen_info() *after* we get the ScreenChangeNotify event, as a result of it. I suspect that the code here is wrong: https://github.com/qtile/qtile/pull/5218/files#diff-ec6963b4884cc405393d63da145f2ed420915dd373a5b6689a1fc726ed237487R423 or not getting called or something. But it's copy pasted from the original code, so it should be the same...

tych0:
Ok, I did find an ordering problem with the selection not getting set up when reloading the config, vs. a completely fresh start. It should be fixed now, try this version.

saviola777:
Looks good! The screens always seem to be reconfigured twice, but that's much better than not at all.


> 2025-04-23 08:52:15,033 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> 2025-04-23 08:52:15,079 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> 2025-04-23 08:52:53,940 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> 2025-04-23 08:52:54,089 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.

ervinpopescu:
> Looks good! The screens always seem to be reconfigured twice, but that's much better than not at all.
> 
> > 2025-04-23 08:52:15,033 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> > 2025-04-23 08:52:15,079 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> > 2025-04-23 08:52:53,940 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.
> > 2025-04-23 08:52:54,089 INFO libqtile manager.py:reconfigure_screens():L449 Reconfiguring screens.

I think this was always an "issue", as far as I remember at least :)

tych0:
> I think this was always an "issue", as far as I remember at least :)

Yeah, some kind of weird x11 nonsense. It seems to only happen the first time (here i'm just playing around making different monitors `--primary`):

```
~ xev -event randr
Outer window is 0x3800001, inner window is 0x3800002

RRScreenChangeNotify event, serial 25, synthetic NO, window 0x3800001,
    root 0x542, timestamp 214207, config_timestamp 205138
    size_index 65535, subpixel_order SubPixelUnknown
    rotation RR_Rotate_0
    width 7680, height 2160, mwidth 1194, mheight 336

RRScreenChangeNotify event, serial 25, synthetic NO, window 0x3800001,
    root 0x542, timestamp 48756293, config_timestamp 205138
    size_index 65535, subpixel_order SubPixelUnknown
    rotation RR_Rotate_0
    width 7680, height 2160, mwidth 1194, mheight 336

RRNotify event, serial 25, synthetic NO, window 0x3800001,
    subtype XRROutputChangeNotifyEvent
    output eDP-1, crtc None, mode None
    rotation RR_Rotate_0
    connection RR_Connected, subpixel_order SubPixelUnknown

RRNotify event, serial 25, synthetic NO, window 0x3800001,
    subtype XRROutputChangeNotifyEvent
    output DP-1-8, crtc 64, mode 3840x2160 (3840x2160)
    rotation RR_Rotate_0
    connection RR_Connected, subpixel_order SubPixelUnknown

RRScreenChangeNotify event, serial 32, synthetic NO, window 0x3800001,
    root 0x542, timestamp 48778232, config_timestamp 205138
    size_index 65535, subpixel_order SubPixelUnknown
    rotation RR_Rotate_0
    width 7680, height 2160, mwidth 1194, mheight 336

RRNotify event, serial 32, synthetic NO, window 0x3800001,
    subtype XRROutputChangeNotifyEvent
    output DP-2-2, crtc 63, mode 1920x1200 (1920x1200)
    rotation RR_Rotate_0
    connection RR_Connected, subpixel_order SubPixelUnknown

RRNotify event, serial 32, synthetic NO, window 0x3800001,
    subtype XRROutputChangeNotifyEvent
    output DP-1-8, crtc 64, mode 3840x2160 (3840x2160)
    rotation RR_Rotate_0
    connection RR_Connected, subpixel_order SubPixelUnknown
```

The timestamp on the first one looks bad. I suppose what we could do is query the current time from the x11 server, and if the event is not "close" to that time, we can discard it? But I agree,

> I think this was always an "issue", as far as I remember at least :)

this has been a problem for a long time.

tych0:
Gonna paste the failure here for posterity, I don't *think* it's related to my PR:

```
  =================================== FAILURES ===================================
  _______________________________ test_cmus[1-x11] _______________________________

  cmus_manager = <test.helpers.TestManager object at 0x7fdf567744d0>

      def test_cmus(cmus_manager):
          widget = cmus_manager.c.widget["cmus"]

          widget.eval("self.update(self.poll())")
          assert widget.info()["text"] == "♫ Rick Astley - Never Gonna Give You Up"
  >       assert widget.eval("self.layout.colour") == widget.eval("self.playing_color")
  E       AssertionError: assert (True, 'ffffff') == (True, '00ff00')
  E
  E         At index 1 diff: 'ffffff' != '00ff00'
  E         Use -v to get more diff

  test/widgets/test_cmus.py:166: AssertionError
  ---------------------------- Captured stdout setup -----------------------------
  2025-04-22 20:54:14,805 libqtile manager.py:reconfigure_screens():L449  Reconfiguring screens.
  --------------------------- Captured stderr teardown ---------------------------
  The XKEYBOARD keymap compiler (xkbcomp) reports:
  > Warning:          Could not resolve keysym XF86CameraAccessEnable
  > Warning:          Could not resolve keysym XF86CameraAccessDisable
  > Warning:          Could not resolve keysym XF86CameraAccessToggle
  > Warning:          Could not resolve keysym XF86NextElement
  > Warning:          Could not resolve keysym XF86PreviousElement
  > Warning:          Could not resolve keysym XF86AutopilotEngageToggle
  > Warning:          Could not resolve keysym XF86MarkWaypoint
  > Warning:          Could not resolve keysym XF86Sos
  > Warning:          Could not resolve keysym XF86NavChart
  > Warning:          Could not resolve keysym XF86FishingChart
  > Warning:          Could not resolve keysym XF86SingleRangeRadar
  > Warning:          Could not resolve keysym XF86DualRangeRadar
  > Warning:          Could not resolve keysym XF86RadarOverlay
  > Warning:          Could not resolve keysym XF86TraditionalSonar
  > Warning:          Could not resolve keysym XF86ClearvuSonar
  > Warning:          Could not resolve keysym XF86SidevuSonar
  > Warning:          Could not resolve keysym XF86NavInfo
  Errors from xkbcomp are not fatal to the X server
  ___________ test_thermal_sensor_colour_alert[1-x11-sensors_manager0] ___________

  sensors_manager = <test.helpers.TestManager object at 0x7fdf564d1810>

      @pytest.mark.parametrize("sensors_manager", [{"threshold": 30}], indirect=True)
      def test_thermal_sensor_colour_alert(sensors_manager):
          _, temp = sensors_manager.c.widget["thermalsensor"].eval("self.layout.colour")
  >       assert temp == "ff0000"
  E       AssertionError: assert 'ffffff' == 'ff0000'
  E
  E         - ff0000
  E         + ffffff

  test/widgets/test_sensors.py:97: AssertionError
  ---------------------------- Captured stdout setup -----------------------------
  2025-04-22 20:54:43,875 libqtile manager.py:reconfigure_screens():L449  Reconfiguring screens.
  --------------------------- Captured stderr teardown ---------------------------
  The XKEYBOARD keymap compiler (xkbcomp) reports:
  > Warning:          Could not resolve keysym XF86CameraAccessEnable
  > Warning:          Could not resolve keysym XF86CameraAccessDisable
  > Warning:          Could not resolve keysym XF86CameraAccessToggle
  > Warning:          Could not resolve keysym XF86NextElement
  > Warning:          Could not resolve keysym XF86PreviousElement
  > Warning:          Could not resolve keysym XF86AutopilotEngageToggle
  > Warning:          Could not resolve keysym XF86MarkWaypoint
  > Warning:          Could not resolve keysym XF86Sos
  > Warning:          Could not resolve keysym XF86NavChart
  > Warning:          Could not resolve keysym XF86FishingChart
  > Warning:          Could not resolve keysym XF86SingleRangeRadar
  > Warning:          Could not resolve keysym XF86DualRangeRadar
  > Warning:          Could not resolve keysym XF86RadarOverlay
  > Warning:          Could not resolve keysym XF86TraditionalSonar
  > Warning:          Could not resolve keysym XF86ClearvuSonar
  > Warning:          Could not resolve keysym XF86SidevuSonar
  > Warning:          Could not resolve keysym XF86NavInfo
  Errors from xkbcomp are not fatal to the X server
  FAILED test/widgets/test_cmus.py::test_cmus[1-x11] - AssertionError: assert (...
  FAILED test/widgets/test_sensors.py::test_thermal_sensor_colour_alert[1-x11-sensors_manager0]
  ====== 2 failed, 1312 passed, 3 skipped, 2 xpassed in 1536.11s (0:25:36) =======
  ```
  
  I re-ran CI, we'll see...

tych0:
Arg. Same two tests fail...

elParaguayo:
> Arg. Same two tests fail...

Let me take a look.

tych0:
Ok, I think what's happening here is that:

1. the tests correctly `poll()` and render the correct alerting foreground
2. when we `reconfigure_screens()`, we create all new screen/bar/widget objects, but do not actually `poll()` them. This means that their foreground is reset to the default
3. we do the assert from the tests and see this new reconfigured bar, instead of the one that actually polled.

IMO, the fix here is to not subscribe to `reconfigure_screens()` from (at least) these two tests, but perhaps we should do it everywhere just to avoid confusion.

The breakage is introduced because we're moving the reconfigure event much later, but that is somewhat unavoidable with this series (and is e.g. the thing that will fix the missing `current_window` attributes, etc.).

tych0:
I'm happy to push that patch, or if you see a different way @elParaguayo happy to look at that as well.

elParaguayo:
> I'm happy to push that patch, or if you see a different way @elParaguayo happy to look at that as well.

Your fix sounds correct to me. I had assumed this was related to the timing of the `poll` call.

tych0:
Ok, that seemed to work. I'm wondering if we shouldn't merge this and let it soak for a bit before any release.

tych0:
ok, here goes take two...

Issue #5155, opened 2025-01-12
Popup windows not working on GIMP on xwayland


TheHighGroundWins:
### Issue description

Popups in Xwayland GIMP not appearing, possibly true to all xwayland programs.

Tested on an empty config, and issue happens regardless.

The popups work just after a file is made or opened, but after any changes they stop appearing.

I tested this on sway another tilling manager albiet more suited to wayland and it doesn't have this issue, so its not a window tilling manager specific problem.

### Version

0.30.1.dev1+gfecc37d8

### Backend

Wayland (experimental)

### Logs

```text
2025-01-12 13:47:04,079 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1163] Failed to get window property
2025-01-12 13:47:04,234 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1163] Failed to get window property
2025-01-12 13:47:05,013 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1163] Failed to get window property
2025-01-12 13:47:06,065 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1163] Failed to get window property
2025-01-12 13:47:06,915 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1619] xcb error: op ChangeProperty (no minor), code Window (no
extension), sequence 1690, value 25165830
2025-01-12 13:47:06,915 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1619] xcb error: op DeleteProperty (no minor), code Window (no
extension), sequence 1691, value 25165830
2025-01-12 13:47:06,916 ERROR wlroots log.py:log_func_callback():L15 [xwayland/xwm.c:1619] xcb error: op DeleteProperty (no minor), code Window (no
extension), sequence 1693, value 25165830
2025-01-12 13:48:17,801 WARNING libqtile wlrq.py:finalize_listener():L234 Failed to remove listener for event: <pywayland.server.listener.Signal object at 0x76b11e52ca70>
2025-01-12 13:48:17,801 WARNING libqtile wlrq.py:finalize_listener():L234 Failed to remove listener for event: <pywayland.server.listener.Signal object at 0x76b11e52c950>
2025-01-12 13:48:18,467 WARNING libqtile wlrq.py:finalize_listener():L234 Failed to remove listener for event: <pywayland.server.listener.Signal object at 0x76b11e6b7530>
[...]

PR #5219, opened 2025-04-07
Fix Xwayland popup behaviour

richcarni:
In some Xwayland apps, popups will appear the first time, but after being closed, cannot be opened again. Issue #5155

At least in some cases this appears to be caused by the app unmapping the popup when it is closing, but not destroying it. Then when it is needed it is re-mapped

This currently doesn't work in qtile, because when an Xwayland window is unmapped/dissociated it loses it associate_event listener

This fix consists of:
- Preserving the associate_event listener
- Removing the window from the group it may be in

This second modification appears necessary for the window to be managed correctly when it is re-mapped


richcarni:
I'm not confident that removing the window from the group is the correct way to go about ensuring the window is re-managed...

Perhaps someone with a deeper understanding of the window management system would care to comment

libqtile/backend/wayland/xwindow.py
```diff
@@ -79,6 +79,12 @@ def _on_dissociate(self, _listener: Listener, _data: Any) -> None:
         if wlr_surface := self.surface.surface:
             self.finalize_listener(wlr_surface.map_event)
             self.finalize_listener(wlr_surface.unmap_event)
+        # For some X11 pop-up windows that unmap but are not destroyed, we need to unset their
+        # group, so they can be correctly re-managed
+        if self.group:
+            if self in self.group.windows:
+                self.group.windows.remove(self)
+            self.group = None
```
tych0, line 87: IIUC, what we want here should be the equivalent of X11's UnmapNotify event, for which the code is here:

https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/backend/x11/core.py#L779-L804

If we ignore the X11 property stuff, it looks basically like you want to `.hide()` the window, then call qtile's `.unmanage()`.

tych0:
Also, seems like coveralls is back online, so you can ignore the CI failure, it'll be fixed on the next push.

libqtile/backend/wayland/xwindow.py
richcarni: So I think the xwayland equivalent is `_on_unmap()`, which calls `unmanage()`
https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/backend/wayland/xwindow.py#L116-L138
but at the end of `unmanage()` it puts the group back on the window (to preserve hooks)
https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/core/manager.py#L742-L758
finally in `manage()` it doesn't get to `self.current_screen.group.add(win)` because `win.group` is not None
https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/core/manager.py#L720-L740

So one solution would be to not put the group back on the window in `unmanage()` - presumably breaking hooks

My solution was to remove in the xwindow `_on_dissociate()` event handler. I guess this would still break hooks? But only for xwayland windows??

tych0: I'm probably out of my depth here on the wayland side, but what do you mean by `hooks` here? This is the wayland callbacks, not qtile hooks proper?

richcarni: Was referring to these lines with comment from `qtile.unmanage()`
https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/core/manager.py#L755-L758

I assume this is referring to the built-in hooks? (e.g. `subscribe.client_killed()`)

The problem I'm seeing is that preserving `group` is preventing `qtile.manage()` from reaching the final line here: (which seems to be necessary to see the window again)

https://github.com/qtile/qtile/blob/42e03cdb6bbb6d88f0fd58927c6ec6258f625961/libqtile/core/manager.py#L735-L738

tych0: Yeah, that clarifies, thanks. I wonder if the bug is here:

https://github.com/tych0/qtile/blob/87267ff585b1bc70ec7c835f4c3c37a9ccc64185/libqtile/core/manager.py#L757-L759

That we should delete the group again after running `client_killed`?

richcarni: I think your last suggestion is much tidier than what I was doing!

It wasn't clear to me that it's literally just the client_killed hook that we were concerned about. So the reason we want to preserve the group is that somebody might be using the group for something in their client_killed hook? After that we don't care - group will be reassigned when/if the window is managed again? So this "should" work fine?

tych0: "should", does it work like you expect in your testing?

richcarni: Yeah, sorry, that was probably unclear

It still fixes the xwayland popups not reappearing issue in my testing

tych0: Ok, cool. Can you drop the first patch? LGTM otherwise.

richcarni: Should I add a comment to the changelog, squash all the commits and rebase?

tych0: Yep, then I'm good to land. Thanks for the fix!

richcarni: Thanks for reviewing!

PR #5223, opened 2025-04-09
feat(wlan): add `{ipaddr}` to display local IP address

trimclain:
Hey there,

first of all, thank you for your amazing work on `qtile`!

With this PR I wanted to add the ability to display the local IP in the qtile bar, so that I can have
```
ethernet_message="🖧 {ipaddr}"
```
in my config, similar to how [i3status](https://i3wm.org/docs/i3status.html#_ethernet) and [waybar](https://github.com/Alexays/Waybar/wiki/Module:-Network#format-replacements) allow to do it.

What do you think?

### DIFF
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -25,6 +25,8 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import subprocess
+
 import iwlib
 
 from libqtile.log_utils import logger
@@ -39,6 +41,28 @@ def get_status(interface_name):
     quality = interface["stats"]["quality"]
     essid = bytes(interface["ESSID"]).decode()
     return essid, quality
+
+
+def get_private_ip(interface_name):
+    try:
+        result = subprocess.run(
+            ["ip", "-brief", "addr", "show", "dev", interface_name],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except (subprocess.CalledProcessError, OSError):
+        logger.exception(f"Couldn't get the IP for {interface_name}:")
+        return "N/A"
+
+    output = result.stdout.strip()
+    parts = output.split()
+    if len(parts) > 2 and parts[1] == "UP":
+        ip_address = parts[2].split("/")[0]
+        if ":" not in ip_address:
+            return ip_address
+
+    return "N/A"
 
 
 class Wlan(base.InLoopPollText):
@@ -60,7 +84,11 @@ class Wlan(base.InLoopPollText):
         ),
         ("update_interval", 1, "The update interval."),
         ("disconnected_message", "Disconnected", "String to show when the wlan is diconnected."),
-        ("ethernet_message", "eth", "String to show when ethernet is being used"),
+        (
+            "ethernet_message_format",
+            "eth",
+            "String to show when ethernet is being used. For IP of ethernet interface use {ipaddr}.",
+        ),
         (
             "use_ethernet",
             False,
@@ -69,11 +97,16 @@ class Wlan(base.InLoopPollText):
         (
             "format",
             "{essid} {quality}/70",
-            'Display format. For percents you can use "{essid} {percent:2.0%}"',
+            'Display format. For percents you can use "{essid} {percent:2.0%}". For IP of wlan interface use {ipaddr}.',
         ),
     ]
 
     def __init__(self, **config):
+        if "ethernet_message" in config:
+            logger.warning(
+                "`ethernet_message` parameter is deprecated. Please rename to `ethernet_message_format`"
+            )
+            config["ethernet_message_format"] = config.pop("ethernet_message")
         base.InLoopPollText.__init__(self, **config)
         self.add_defaults(Wlan.defaults)
         self.ethernetInterfaceNotFound = False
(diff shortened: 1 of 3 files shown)


libqtile/widget/wlan.py
```diff
@@ -41,6 +43,20 @@ def get_status(interface_name):
     return essid, quality
 
 
+def get_local_ip():
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        s.connect(("8.8.8.8", 80))  # Google's public DNS server
```
tych0, line 49: This will get the IP address for the interface that's the default route, not necessarily the wlan. 

I guess we probably need some netlink stuff to figure out the interface name :(

trimclain: There were some talks about a unified network widget for wlan and ethernet in #4054 and #4071. I would gladly continue that discussion in this PR, since `{ip_addr}` (at least in its current form) is much more fitting there. I think the [Net](https://docs.qtile.org/en/latest/manual/ref/widgets.html#net) widget could very well be extended, if you don't want a create a new one.
As a reference for how it could be done, I like the [Waybar Network Module](https://github.com/Alexays/Waybar/wiki/Module:-Network) implementation.

tych0: Yeah, I think that's fine if you want to do a larger change; all this information should be available via netlink, and then you won't have to make any connections to the internet to do it.

trimclain: I've pushed the update. Let me know if this is what you meant.

libqtile/widget/wlan.py
```diff
@@ -41,6 +43,26 @@ def get_status(interface_name):
...
+        output = result.stdout.strip()
+        parts = output.split()
+        if len(parts) > 2 and parts[1] == "UP":
+            ip_address = parts[2].split("/")[0]
+            if ":" not in ip_address:
+                return ip_address
+    except Exception:
+        logger.exception(f"Couldn't get the IP for {interface_name}:")
+
+    return "N/A"
```
elParaguayo, line 63: This `try... except...` block is too big and it's generally bad practice to catch `Exception` as you would swallow unexpected scenarios.

The only thing that would throw an exception here is the `subprocess.run` call. Everything else should be handled by your logic.

```suggestion
def get_private_ip(interface_name):
    try:
        result = subprocess.run(
            ["ip", "-brief", "addr", "show", "dev", interface_name],
            capture_output=True,
            text=True,
            check=True,
        )
    except subprocess.CalledProcessError:
        logger.exception(f"Couldn't get the IP for {interface_name}:")
        return "N/A"

    output = result.stdout.strip()
    parts = output.split()
    if len(parts) > 2 and parts[1] == "UP":
        ip_address = parts[2].split("/")[0]
        if ":" not in ip_address:
            return ip_address

    return "N/A"
```

libqtile/widget/wlan.py
```diff
@@ -82,8 +104,10 @@ def poll(self):
         try:
             essid, quality = get_status(self.interface)
             disconnected = essid is None
+            ipaddr = get_private_ip(self.interface)
             if disconnected:
                 if self.use_ethernet:
+                    ipaddr = get_private_ip(self.ethernet_interface)
```
elParaguayo, line 110: I don't think we need to get the ip addr is we're not connected to wifi.

```suggestion
            ipaddr = ""
            if not disconnected:
                ipaddr = get_private_ip(self.interface)
            else:
                if self.use_ethernet:
                    ipaddr = get_private_ip(self.ethernet_interface)
```

elParaguayo (changes requested):
Added a couple of suggestions.

libqtile/widget/wlan.py
trimclain: Thanks, added your suggestions, except I did
```
ipaddr = "N/A"
```
instead of 
```
ipaddr = ""
```
to be consistent.

tych0:
Cool, thanks for switching to iproute2, that will give the right answer. LGTM assuming CI passes.

trimclain:
I should probably add the information about `{ipaddr}` to docs, right? If yes, in what form? The info about `'{essid}` and `{quality}` [is put](https://docs.qtile.org/en/stable/manual/ref/widgets.html#wlan) in the `format` key. Should I maybe write it as a Note below or above the "Configuration options" table with a message similar to:

Following replacement options can be used in messages: `{essid}`, `{quality}`, `{percent}` and `{ipaddr}`

tych0:
I think usually we put them in line with the documentation tuple, there are a few examples in some of the other widgets.

libqtile/widget/wlan.py
```diff
@@ -41,6 +43,28 @@ def get_status(interface_name):
...
 
+def get_private_ip(interface_name):
+    try:
+        result = subprocess.run(
+            ["ip", "-brief", "addr", "show", "dev", interface_name],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
```
elParaguayo, line 54:
```suggestion
    except (subprocess.CalledProcessError, OSError):
```
Weirdly, the docs built fine on my fork of this. However, given that `subprocess.run` can raise an `OSError` too, we should catch that here.

trimclain:
Since we find the IP of both interfaces, I changed the `ethernet_message` to be a format string, which accepts the `{ipaddr}`. However I know that you usually name keys, which are format strings, as `..._format` ([cmus widget](https://docs.qtile.org/en/latest/manual/ref/widgets.html#cmus) is a good example with multiple format keys). So how do I move forward without creating a breaking change?

tych0:
You can do a "migration" to change people's configs automatically. There are some examples in-tree of renaming things.

trimclain:
Any idea, why the CI fails?

tych0:
Yeah, probably freedesktop.org is down again. Annoying that it's so flaky, maybe we should do the caching assets thing now...

trimclain:
If you don't have any more feedback, I think this PR is ready to merge.

tych0:
Yeah, cc @elParaguayo for his approval, but lgtm.

libqtile/widget/wlan.py
```diff
@@ -60,7 +84,11 @@ class Wlan(base.InLoopPollText):
         ),
         ("update_interval", 1, "The update interval."),
         ("disconnected_message", "Disconnected", "String to show when the wlan is diconnected."),
-        ("ethernet_message", "eth", "String to show when ethernet is being used"),
+        (
+            "ethernet_message_format",
+            "eth",
+            "String to show when ethernet is being used. For IP of ethernet interface use {ipaddr}.",
+        ),
```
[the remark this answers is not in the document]
trimclain: Added :)

PR #5224, opened 2025-04-10
accept direct password and add possibility to hide empty ImapWidget

zordsdavini:
I don't use Gnome Keyring and wanted to use directly given password. Also I wanted to hide empty widget (when Unseen == 0)

### DIFF
--- a/libqtile/widget/imapwidget.py
+++ b/libqtile/widget/imapwidget.py
@@ -33,7 +33,7 @@ class ImapWidget(base.ThreadPoolText):
 
     This widget will scan one of your imap email boxes and report the number of
     unseen messages present.  I've configured it to only work with imap with
-    ssl. Your password is obtained from the Gnome Keyring.
+    ssl. Your password can be obtained from the Gnome Keyring.
 
     Writing your password to the keyring initially is as simple as (changing
     out <userid> and <password> for your userid and password):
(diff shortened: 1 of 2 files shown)


libqtile/widget/imapwidget.py
```diff
@@ -61,37 +61,48 @@ class ImapWidget(base.ThreadPoolText):
...
+
     Widget requirements: keyring_.
 
     .. _keyring: https://pypi.org/project/keyring/
     """
 
     defaults = [
         ("mbox", '"INBOX"', "mailbox to fetch"),
-        ("label", "INBOX", "label for display"),
+        ("label", "INBOX: ", "label for display"),
```
tych0, line 74: This seems unnecessary?

tych0 (commented):
One small bit, looks good to me otherwise.

libqtile/widget/imapwidget.py
zordsdavini: the `text` was cosntructed with hardcoded `:`. I removed that, so wanted to get same functionality

tych0: I see, ok. seems like the tests  need some fixing too.

zordsdavini: I'm trying to understand them. Not good at writing tests with python

zordsdavini: @tych0 it looks like I fixed tests. But docs are failing. Should them be fixed?

tych0: yeah, everything needs to build correctly.

elParaguayo:
You need to rebase this onto the latest got master. That should fix the docs build.

tych0:
I went ahead and rebased this, assuming the tests pass lgtm.

PR #5228, opened 2025-04-14
udev: add a --debug-log-file option

tych0:
systemd-udevd (and probably others) do not capture stdout, only exit codes, so when there is a problem with the udev script, currently all output is lost. Instead, let's add an option to log to an explicit file, so people can do things like `qtile udev ... --debug-log-file=/tmp/qtile_udev` and debug.

Note that you will need a:

    udevadm control --reload-rules

once you change your rules for this to take effect.


tych0:
This is slightly broken: if udev calls `qtile udev` multiple times, it'll overwrite and garble the log file. I suppose we could/should switch it to log to separate files, but that also seems ugly. Really I just want this around for testing, and unfortunately I haven't seen things fail yet.

PR #5232, opened 2025-04-20
Add ability to spawn process in different group

elParaguayo:
Adds a `group` argument to `spawn()` to try to spawn processes in different groups.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -29,6 +29,7 @@ import pickle
 import shlex
 import shutil
 import signal
+import socket
 import subprocess
 import tempfile
 from collections import defaultdict
@@ -1277,7 +1278,11 @@ class Qtile(CommandObject):
 
     @expose_command()
     def spawn(
-        self, cmd: list[str] | str, shell: bool = False, env: dict[str, str] = dict()
+        self,
+        cmd: list[str] | str,
+        shell: bool = False,
+        env: dict[str, str] = dict(),
+        group: str | None = None,
     ) -> int:
         """
         Spawn a new process.
@@ -1291,6 +1296,8 @@ class Qtile(CommandObject):
             -c". This enables the use of shell syntax within the command (e.g. pipes).
         env:
             Dictionary of environmental variables to pass with command.
+        group:
+            Name of group in which to spawn process.
 
         Examples
         ========
@@ -1331,9 +1338,24 @@ class Qtile(CommandObject):
                 (os.POSIX_SPAWN_DUP2, 2, null.fileno()),
             ]
 
+            # To create a rule to move process to a specific group and manage race conditions
+            # we need to do the following:
+            # 1) Create a socket pair so we can communicate to child process
+            # 2) Spawn a "qtile launch" process and pass the file descriptor
+            # 3) Have "qtile launch" notify qtile that it's ready
+            # 4) Create a rule using the pid from step 2
+            # 5) Send a message over the socket to "qtile launch"
+            # 6) Once the message is received, "qtile launch" execs the desired process
+            if group is not None:
+                parent_sock, child_sock = socket.socketpair()
+                # socket pair fds are non-inheritable by default but we need this to survice
+                os.set_inheritable(child_sock.fileno(), True)
+
             # this API is only available on python 3.13 or above, and only on platforms
             # where posix_spawn_file_actions_addclosefrom_np() exists (only glibc on Linux).
-            if hasattr(os, "POSIX_SPAWN_CLOSEFROM"):
+            # where we're creating a rule to manage the group, we'll handle closing fds in
+            # `qtile launch`
+            if hasattr(os, "POSIX_SPAWN_CLOSEFROM") and group is None:
                 # we should close all fds so that child processes don't
                 # accidentally write to our x11 event loop or whatever; we never
                 # used to do this, so it seems fine to only do it where this nice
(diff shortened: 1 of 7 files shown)


elParaguayo:
Looks like this is a bit hit and miss. Passed on wayland with python 3.10 but failed on 3.11.

libqtile/core/manager.py
```diff
@@ -1341,7 +1347,12 @@ def spawn(
                 file_actions.append((os.POSIX_SPAWN_CLOSEFROM, 3))
 
             try:
-                return os.posix_spawnp(args[0], args, env, file_actions=file_actions)
+                pid = os.posix_spawnp(args[0], args, env, file_actions=file_actions)
+                if group is not None:
```
tych0, line 1351: This is always going to be slightly racy: if the pid launches and creates a window before we add the pid to the rules list, we won't catch it.

To fix that, I think the following would work:

1. add a `qtile launch --fd N ...` which waits for a character to be sent on FD N, and then `exec()s` the process
2. changes this to `posix_spawnp()` that doesn't close all the fds, and prepends `qtile launch --fd N` to that
3. in this code, we can create a socket pair, launch the `qtile launch` process, add that pid rule, send the notification on the socket pair

Since the `exec()` from `qtile launch` will keep the same pid, the rule match should work.

elParaguayo: We've got `qtile run-cmd` which is where this `Rule`-based approach came from. I was already suspicious about the race condition there but it seemed to work well and this code also worked well in (admittedly, limited) practice but I wasn't surprised to see it fail in CI.

I *think* I understand your steps above (I'm still relatively new to sockets/file descriptors) and will give it a try.

tych0: Oh yeah, I wonder if we have tests for that. But in any case, that's racy too.

libqtile/scripts/launch.py
```diff
@@ -0,0 +1,59 @@
...
+import os
+import select
+
+
+def launch(opts) -> None:
+    fd = opts.fd
+
+    if fd > 0:
+        # Check if there's anything to read on the file descriptor
+        # but don't block
```
tych0, line 35: Why not block? Isn't the point of this to block?

elParaguayo: Bad wording. It's a timeout so it blocks for a second. 

I included a timeout so that, if for some reason the message doesn't get sent, we still launch the required process.

tych0: But if the message doesn't get sent, isn't that a bug? So with the timeout, we'll just be masking the bug, IIUC.

elParaguayo: You could look at it like that!

My thinking was that spawning the process should happen, regardless of whether we're create the rule or not. If we just try to read the socket and there's no data then read will block, the process will never spawn and we'll still have a `qtile launch` process.

I'm open to suggestions.

tych0: Perhaps I have PTSD from work, but I think introducing timeouts just masks real bugs. Personally, I'd just do a raw .read() and if that hangs, then we have something to fix.

libqtile/scripts/launch.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+
+
+def launch(opts) -> None:
+    fd = opts.fd
+
+    if fd > 0:
+        # Read the socket but we don't care what the contents is as it
+        # will be empty until qtile writes to it
+        os.read(fd, 4)
```
tych0, line 35: It feels like we could check for your sentinel string here. I'm wondering if the source of flakiness is that this is somehow a nonblocking socket by default, so this is just giving 0?

elParaguayo: From my tests, `os.read` blocks when reading from a socket created by `socket.socketpair()`.

tych0: hm, ok. unclear why it's flaky then...

elParaguayo: Let me add some debugging and see if we can get some clues.

elParaguayo: Ha. So everything passed...

elParaguayo: Yup. Looks like window is launching before rule has been added.

tych0: Sometimes I find inserting manual sleep()s to trigger things in a certain order works for debugging. I can't look now, but hopefully I can look later this week.

elParaguayo: Latest failure showed that the `client_new` hook wasn't called to run the rules on the new window.

elParaguayo: @tych0 I think what's happening is that qtile is sending the sentinel but `qtile launch` isn't ready to read it yet. 

Am I right in thinking that we need to be listening for data before it's sent? The data doesn't persist until it's been read?

libqtile/scripts/launch.py
```diff
@@ -0,0 +1,66 @@
...
+    if fd > 0:
+        # Create a socket from the file descriptor
+        sock = socket.fromfd(fd, socket.AF_INET, socket.SOCK_STREAM)
+
+        # Tell qtile we're ready to listen for their message
+        sock.sendall(b"READY")
+
+        # Receive OK message (but we're not checking the contents, we just
+        # block until we receive the data)
+        _ = sock.recv(2)
```
tych0, line 42: I would print out what you get back here. The fact that it's not working suggests something strange is happening.

libqtile/scripts/launch.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import socket
+
+
+def launch(opts) -> None:
+    fd = opts.fd
+
+    if fd > 0:
+        # Create a socket from the file descriptor
+        sock = socket.fromfd(fd, socket.AF_INET, socket.SOCK_STREAM)
```
tych0, line 35: I wonder if this is doing something strange. We can just read() from the fd directly, I don't think we need to promote it like this.

libqtile/scripts/launch.py
elParaguayo: Where are we printing it? stdout and stderr have been redirected to `/dev/null`. I suppose we could not do that redirect until just before we `exec`.

elParaguayo:
It's not quite ready as I would like to move the `Group` `spawn` parameter to this but it requires a bit more refactoring to make it work.

PR #5241, opened 2025-04-25
fix: display hook registry name in error message

emmanuel-ferdman:
# PR Summary
Small PR - evaluates and displays the hook registry name in error message.


elParaguayo:
Thanks. This looks ok to me.

PR #5242, opened 2025-04-27
FAQ typo

ervinpopescu:
Small note: Opening the file in my editor ran some prettifier on it. If it doesn't break docs, I think it's fine to leave it there?


tych0:
In general I don't like auto-formatting stuff in diffs unless we decide to enforce it via CI, because everyone has their editor configured differently and to auto-format things differently.

That said, I don't think it's so important in docs, so it's fine in this patch. Thanks!

PR #5243, opened 2025-05-01
Handle uppercase letters and special characters for internal windows

meistermo:
### Summary
This PR fixes an issue with the key input handling on the wayland backend where uppercase letters and special characters could not be typed into internal windows, like the prompt widget on the bar. On X11, this is not an issue.

### Solution
- Moved the `handled` check earlier in the `_on_key` function.
- Adjusted keysyms based on modifiers and passed the adjusted keysym to `process_key_press` for internal windows, because just passing modifiers alongside default keysyms like with `process_key_event` was not an option.

### Testing
- Verified that uppercase letters and special characters can now be typed in internal windows.


tych0:
Your write-up in the PR message is great, can you include that in the commit message?

PR #5246, opened 2025-05-08
Add ability to pass kwargs to `qtile cmd-obj`

elParaguayo:


libqtile/scripts/cmd_obj.py
```diff
@@ -40,6 +40,23 @@
...
+            kwargs[k] = v
+
+        setattr(args, self.dest, kwargs)
+
+
+def keyval(value):
+    k, _, v = value.partition("=")
+    if not k and v:
+        raise argparse.ArgumentTypeError(f"{value} is not a valid key=value pair.")
+    return value
```
elParaguayo, line 57: Actually, this is probably redundant as we can deal with this during the custom action parsing.

tych0 (approved):

Issue #5256, opened 2025-05-20
Qtile eat my key on dropdown toggle


QiangF:
### Issue description

I have this in my config file.

    groups.append(
        ScratchPad("scratchpad", [
            DropDown("term", "/home/my_usr/bin/st", on_focus_lost_hide=False, height=1.0, width=0.5, x=0, y=0,)
        ]), )

    groups.append(
        ScratchPad("scratchpad", [
            DropDown("emacs", "/usr/bin/emacs --name=editor", on_focus_lost_hide=False, height=1.0, width=0.8, x=0, y=0,),
            DropDown("term", "/usr/local/bin/st", on_focus_lost_hide=False, height=1.0, width=0.5, x=0, y=0,)
        ],  single=False, ), )

    EzKey("M-t", lazy.group['scratchpad'].dropdown_toggle('term'))
    EzKey("M-e", lazy.group['scratchpad'].dropdown_toggle('emacs'))
    
to reproduce:
1. show the smaller window dropdown, here is the term dropdown
2. show the bigger window dropdown, here is emacs, now the term dropdown is blocked, under emacs dropdown, invisible
3. press M-t once, the term dropdown does not pop up
4. press M-t once more, the term dropdown pops up

### Version

latest

### Backend

X11 (default)

tych0:
Feels like some kind of bug in this logic:

https://github.com/qtile/qtile/blob/6e31cf27f95fcd41e22322442d9ecb3eae249ffb/libqtile/scratchpad.py#L101-L139

on the first press of your hotkey, what are the values of `self.visible` and `self.shown` in `show()`?

PR #5260, opened 2025-05-22
fix dropdown_toggle when on_focus_lost_hide=False

QiangF:
When multiple dropdowns are shown, press toggle keybinding for unfocused dropdown should focus it and bring it to the front, instead of needing to press it twice to get focused.

Fixes https://github.com/qtile/qtile/issues/5256


tych0:
CI failure doesn't look like your fault, seems gitlab started appending the hash to the directory names inside of their tarballs, which is... annoying. I'll try to fix this in #5261

tych0:
Blah! I didn't want to close this...

PR #5269, opened 2025-05-30
fix: libdrm repository name change

RFCreate:
Using a hash is not really needed if you use the current repository name. This should be more maintainable than using the hash.


tych0:
Thanks!

PR #5271, opened 2025-06-04
`meson build ...` deprecation since 0.64

ervinpopescu:
Switch `/scripts/ubuntu_wayland_setup` to `meson setup build ...` instead.

https://wiki.archlinux.org/title/Meson_package_guidelines#Using_meson_binary_directly - good old Arch Wiki :)

https://stackoverflow.com/questions/63549328/how-do-i-set-basic-options-with-meson - SO


tych0:
I noticed this too, thanks for fixing it!

PR #5273, opened 2025-06-08
add netup widget

e-devnull:
A simple network connectivity checker. I wrote this widget because I don't use nm-applet (and sometimes my internet connection is unreliable), so it's useful to see, at a glance, if the connection is up or down.

### DIFF
--- a/libqtile/widget/netup.py
+++ b/libqtile/widget/netup.py
@@ -1,0 +1,111 @@
+# Copyright (c) 2025 e-devnull
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


libqtile/widget/netup.py
```diff
@@ -0,0 +1,49 @@
...
+from libqtile.widget import base
+
+
+class NetUP(base.ThreadPoolText):
+    """
+    A simple network connectivity tester
+    """
+
+    defaults = [
+        ("host", "localhost", "Host to ping."),
```
tych0: Defaulting to localhost seems kind of weird, although I understand why you do it. Maybe default to `None`, and then print an error telling people to pick a host?

e-devnull: I made the change and set host to None.

tych0:
Oh, I almost misssed @elParaguayo's comment here: https://github.com/qtile/qtile/commit/f92221da412efa272d5f49289746c7ef88a42746#r159574439

e-devnull:
I added the option to perform a check via tcp connection.

PR #5274, opened 2025-06-10
Serial number ordering

tych0:
Here is another step towards a randr style API where users can specify Screen(..., serial="...") as it's printed on the back of their monitor, and have their bar info bound to that exact Screen object.

I have more patches that will directly do randr style stuff of setting resolution etc, but I have no real way to do the wayland side, so I'll save those for now. Hopefully the wayland side of this is trival to test :)


ervinpopescu:
I think I had some issues with this on Wayland, but I can't remember what... I need to test this again...

tych0:
Any thoughts on this?

ervinpopescu:
I'm really busy with work, I might have some time on Saturday, but I'm fine with you landing it, I'll report issues afterwards (if any). Sorry for the delay!

LE: Ah, I don't even have a way to test, since I'm away from home and with no external monitors available :)

zordsdavini:
Interesting solution and waiting for landing. Maybe there are script to output serial numbers of current monitors? I don't see s/n on my laptop. 
Other thing: if I have connected 1 monitor I want to have bottom bar on and systray and other widgets on laptop but by connecting bigger monitor I want to move all this stuff to bigger monitor and on laptop only leave layout and windowTitle. Will this solve it?
And if I go to other random place and want to use bigger monitor - will I have to duplicate config parts to add another s/n?

tych0:
>  Other thing: if I have connected 1 monitor I want to have bottom bar on and systray and other widgets on laptop but by connecting bigger monitor I want to move all this stuff to bigger monitor and on laptop only leave layout and windowTitle. Will this solve it?

No, it won't, but that's an interesting point. It's almost like what we really want is a function which, on initial boot or screen configuration, passes the list of ScreenRect()s with their serial numbers, and then people can return a list of config.Screen objects that they want to bind. That would work.

>  And if I go to other random place and want to use bigger monitor - will I have to duplicate config parts to add another s/n?

Yes, you would. (I mostly don't go anywhere, which is why all the static binding.)

But if it was a function, you could do this dynamic binding. Does that make sense? Something like:

`def configure_screens(monitors: ScreenRect) -> config.Screen`?

zordsdavini:
> But if it was a function, you could do this dynamic binding. Does that make sense? Something like:
> 
> `def configure_screens(monitors: ScreenRect) -> config.Screen`?

I just imagine simple meet-up and connecting to projector. Ex., I'll have s/n for laptop and prepared "empty" screen for such cases. It should setup laptop screen by s/n and "empty" screen attached to projector.

Another example would be for this dynamic binding. Lonely laptop with known s/n can use one screen setup but if big monitor attached with known s/n - it should be possible to set laptop screen to other setup

Am I thinking right?

Speaking the true. For me primary monitor as first screen and second screen setup later repeated for each attached was the best solution and it worked for 10 years ( wow - 10 years of using Qtile - time to celebrate :D ). I use `autorandr` and it sets right primary monitor on different layouts. In attached projector case I just used `arandr` to setup quick layout and Qtile all did right

tych0:
> I just imagine simple meet-up and connecting to projector. Ex., I'll have s/n for laptop and prepared "empty" screen for such cases. It should setup laptop screen by s/n and "empty" screen attached to projector.

Yeah, that is what happens with this code right now, but you said,

> And if I go to other random place and want to use bigger monitor - will I have to duplicate config parts to add another s/n?

I was thinking here you wanted bar widgets and such, which would require another screen. As-is, this branch preserves existing behavior, where if an addition screen is attached, qtile maps it with no bars or gaps.

richcarni:
From what I have gathered, this would be incredibly useful to many users with multiple displays! Let me know what I can do to help

My only comment would be to wonder if there is any value in also being able to bind by display name? I think, at least in wayland, this seems to reliably be available in the same place. But I do see the advantage of using serial

Anyway, I've tried to refer some actual wayland users over to try this out

tych0:
At least in x11 the display name is not stable across boots, so the serial number is the only option. I'm certainly happy to add the ability to bind by display, it's not very hard.

Note that I have not actually tested the wayland version of this since I rebased it on wayc, I'm just doing the minimum to keep it alive until I have time to work on it again :). Feel free to take it over if you want.

tych0:
One thing I was thinking would be helpful that was suggested by @zordsdavini that I totally missed is:
> Maybe there are script to output serial numbers of current monitors? I don't see s/n on my laptop.
We could do this pretty simply with a subcommand `qtile monitor-info`. Or maybe even `qtile monitor info`, suggesting that at some point we would allow xrandr style things via `qtile monitor config`? But perhaps that's too ambitious.

richcarni:
>At least in x11 the display name is not stable across boots

Ah, I have not tested extensively tbh. It's likely that they're also not 100% stable for wayland too then

re `qtile monitor-info`, sounds like this could be useful if it's a hassle to find the serial on x11. On wayland, `wlr-randr` seems to display the serial if available

Ironically, I don't actually use multiple screens. For dev/testing I use a vm with two virtual displays - unfortunately these don't appear to have serials so I can't currently test this PR

If we did `qtile monitor config` I imagine we could also setup monitors in config. Sway has something like this

libqtile/config.py
```diff
@@ -356,19 +356,22 @@ class ScreenRect:
     y: int
     width: int
     height: int
+    serial: str | None
```
richcarni: Could this have a default = None, so we don't have to pass None every time we use it without serial?

tych0:
> re qtile monitor-info, sounds like this could be useful if it's a hassle to find the serial on x11. On wayland, wlr-randr seems to display the serial if available

It's not exactly a hassle, but you do have to do some edid decoding, which is annoying. I wish xrandr would have done this for us, but here we are...

Maybe not worth adding, then, if we're just going to move to wayland?

In any case, I'm not sure `qtile monitor config` is that valuable. For monitors that people always have, they probably want to specify things like Screen(left_of=other_screen), rather than invoking the command line every time. For monitors that they plug into (e.g. at a conf giving a talk or whatever), just adding an "empty" screen seems like enough, it doesn't really matter where it is.

acidkewpie:
This worked for me, however in general this feels a bit XY the more I read the code. get_screen_info() was presumably intended for providing the visual representation of a screen. Now though, as features are added for managing the screens more fundamentally (awesome!), relevant info about a screen expands significantly, and I think there's a minor paradigm shift required, rather than searching through the code and tagging on a serial field in each case where a ScreenRect is passed. ScreenRect should remain untouched, and instead Screen should be expanded.

I'm not quite sure how constantly aware of the screen layout qtile is, but all the data about all screens (including disabled ones?) should be available within the screens list, rather than being grabbed on demand from the backend. Maybe. Or if not, the grabbing of that data should be done through the Screen class itself?

I mentioned disabled screens as, I think, when a screen is disabled it is destroyed in code? By keeping disabled screens we can, when re-enabled, remember what group it had on it and restore it quite easily.

libqtile/backend/wayland/qw/server.c
```diff
@@ -90,7 +90,7 @@ void qw_server_loop_output_dims(struct qw_server *server, output_dims_cb_t cb) {
         }
         int width, height;
         wlr_output_effective_resolution(o->wlr_output, &width, &height);
-        cb(o->x, o->y, width, height);
+        cb(o->x, o->y, width, height, o->wlr_output->serial);
     }
```
acidkewpie, line 94: Pass the whole o->wlr_output instead?

libqtile/backend/wayland/core.py
```diff
@@ -514,9 +514,10 @@ def create_internal(self, x: int, y: int, width: int, height: int) -> base.Inter
     def get_screen_info(self) -> list[ScreenRect]:
         rects = []
 
-        @ffi.callback("void(int, int, int, int)")
-        def loop(x: int, y: int, width: int, height: int) -> None:
-            rects.append(ScreenRect(x, y, width, height))
+        @ffi.callback("void(int, int, int, int, char *)")
+        def loop(x: int, y: int, width: int, height: int, serial: ffi.CData) -> None:
+            serial_str = ffi.string(serial).decode() if serial != ffi.NULL else None
+            rects.append(ScreenRect(x, y, width, height, serial_str))
```
acidkewpie: Is the serial number really any business of the ScreenRect? From what I understand, it should be changing to appending the actual Screen here if we're expanding the scope of what "screen info" means?

So I'd think the callback should be updating an existing Screen, rather than making a new ScreenRect?

libqtile/core/manager.py
```diff
@@ -385,25 +386,66 @@ def _process_screens(self, reloading: bool = False) -> None:
...
-                scr = config[i]
+            scr = Screen(serial=info.serial)
+            fresh_screen = True
+
+            # first, try to find a screen that matches this one by serial
+            # number
+            for screen in config:
+                if screen.serial is not None:
+                    if not have_serials_from_hardware:
+                        # if no hardware provided a serial and people provided
```
acidkewpie, line 432: Here, for Wayland at least, we could also match on combinations of description, make and model, week and year as would be trivially found by having the entire wlr_output data, in the case of having no serial, which also could just be user preference.

libqtile/backend/wayland/core.py
tych0: Screen is user specified, it's how people tell us what config they want to bind to a particular screen. 

ScreenRect is Qtile's representation of the underlying hardware. However, we could rename it to Output in this series since we're adding just more than xywh.

libqtile/core/manager.py
tych0: Yeah, we could do all that, but it seems like overkill to me. If you really want to bind that way, maybe we can do some wayland specific thing.

libqtile/backend/wayland/core.py
richcarni: > ScreenRect is Qtile's representation of the underlying hardware. However, we could rename it to Output in this series since we're adding just more than xywh.

I was thinking something similar. Maybe Output contains a ScreenRect to describe geometry plus additional properties (name, serial, state, etc.)

richcarni:
>I mentioned disabled screens as, I think, when a screen is disabled it is destroyed in code? By keeping disabled screens we can, when re-enabled, remember what group it had on it and restore it quite easily.

I think it's important to be clear about what problem(s) we're trying to solve here...

I would expect that if you're putting monitors into a power saving mode or even putting your computer to sleep, that when you resume, everything will come back how it was before. That is what I see with my single display - I don't have to explicitly save/load state. I take it that this is not what you are seeing with multiple displays?

If I unplug my monitor then I lose my wlr-randr settings and qtile state. So I suspect the solution to at least some of your problems is to do with power management

If we're talking about disconnecting and reconnecting displays (like docking/undocking a laptop), I think that is a slightly different problem

tych0:
> If I unplug my monitor then I lose my wlr-randr settings and qtile state. So I suspect the solution to at least some of your problems is to do with power management

For sure. I guess the wlr idle work will help us here? And for x11, we can probably hook into systemd to do a similar state save + restore. Something like @acidkewpie has, but in the qtile codebase proper.

> If we're talking about disconnecting and reconnecting displays (like docking/undocking a laptop), I think that is a slightly different problem

Yeah, that's what I'm trying to solve with this. I agree that the other problem is orthogonal.

tych0:
> And for x11, we can probably hook into systemd to do a similar state save + restore.

But actually it's interesting, I use x11 DPMS pretty constantly for a dual monitor setup, and I never lose state. So maybe it's just a wayland problem after all?

libqtile/backend/wayland/core.py
tych0: Yeah, that makes sense. I will make that change next time I'm looking at this code, unless someone gets to it before me.

richcarni:
>So maybe it's just a wayland problem after all?

Yeah, I think we just haven't implemented wlr output power management yet!

elParaguayo:
> > So maybe it's just a wayland problem after all?
> 
> Yeah, I think we just haven't implemented wlr output power management yet!

This was next on my list (before you suggested changes to the idle PR!)

libqtile/core/manager.py
acidkewpie: It could be sure, but a generic way for a user to match on whatever fields they choose to provide would be nice, no need to specifically look for a serial, just the provided attribute(s). If remaining hardcoded, I think the description string is worth it at the minimum, I recall hyprland can match on description or serial.

libqtile/backend/wayland/core.py
acidkewpie: I see that Screen contains a get_rect(), which to me made it feel that a ScreenRect is far more superficial. And actually Screen is described as "A physical screen, and its associated paraphernalia."..?

acidkewpie:
> Yeah, that's what I'm trying to solve with this. I agree that the other problem is orthogonal.

Agree totally, ultimately it just looks to me that there are a few core steps that could be taken exactly at this stage to broaden the scope of data available that will have significant knock on effects. Whilst pulling / decoding more data can be added later, much of it would be replacing the serial specific code here, so doesn't seem out of place to hit that now.

libqtile/backend/wayland/core.py
tych0: Yeah, that pydoc is all written to be user facing in the autogenerated documentation. Maybe we should clarify there, but I think it would only add confusion. Perhaps we should add a pydoc to ScreenRect saying that it's an underlying physical representation? (although that is also not strictly true viz. split layouts or pseudo screens and stuff. )

tych0:
Ok, based on feedback in this thread I made the following changes:

1. add an Output class, which contains the serial number, name, and ScreenRect for an output. There are some other things I didn't include, but maybe I should (e.g. manufacturer, model number). Happy to add those if people think they're useful.
2. add a `qtile identify-output` command, which prints all this stuff so people can easily edit their configs.

In general, I don't want to add x/wrandr style config as a qtile subcommand; I'd rather encourage people to do it in their configs, and we should develop combinators to support that. I have some very old patches that do this (as I mentioned in the initial PR), but they probably need rethought.

Anyway, let me know what you all think.

richcarni:
Changes look good to me!

>In general, I don't want to add x/wrandr style config as a qtile subcommand; I'd rather encourage people to do it in their configs, and we should develop combinators to support that. I have some very old patches that do this (as I mentioned in the initial PR), but they probably need rethought.

I think you're referring to being able to set things like output position, resolution and scale in config. That would be a very nice convenience

tych0:
> I think you're referring to being able to set things like output position, resolution and scale in config. That would be a very nice convenience

Yep, exactly. And I actually have all the x11 randr stuff figured out, I just got distracted by other stuff for the last year and didn't push on upstreaming it :). Maybe if we land this, I'll see about sending a rebased version so that we can start working on the wayland bits too.

richcarni:
@tych0 Is there anything else you want to add/change with this? Are you happy for me to merge?

tych0:
I realized I didn't address this comment: https://github.com/qtile/qtile/pull/5274#pullrequestreview-3544447404 I can try to do that tonight.

One bigger problem is this one that @zordsdavini pointed out: https://github.com/qtile/qtile/pull/5274#issuecomment-3095246206

> Other thing: if I have connected 1 monitor I want to have bottom bar on and systray and other widgets on laptop but by connecting bigger monitor I want to move all this stuff to bigger monitor and on laptop only leave layout and windowTitle. Will this solve it?

In order to do this, we'd have to have some configuration input to qtile, which currently we don't. I proposed a function for this earlier in this thread, but that's not particularly general. What if instead, we had:

```
from libqtile.config import system_state
```

then, in peoples' configs at startup time, they could do stuff like:

```
if len(system_state.outputs) == 2: ...
```

or

```
if "XYZ" in [output.serial for output in system_state.outputs]: ...
```

etc.

That would be "forwards" compatible with everything I've written in this branch, and then we could potentially add other things too (`system_state.backend`, `system_state.qtile_version` come to mind immediately, but there are probably others that are interesting).

That would also let people reason about the *new* number of screens in e.g. their reconfigure_screens hook, provided we update it before calling the hook.

Thoughts?

richcarni:
> ```
> from libqtile.config import system_state
> ```
> 
> 
>     
>   
> 
> then, in peoples' configs at startup time, they could do stuff like:
> 
> 
> ```
> if len(system_state.outputs) == 2: ...
> ```
> 
> 
>     
>   
> 
> or
> 
> 
> ```
> if "XYZ" in [output.serial for output in system_state.outputs]: ...
> ```

I had been thinking about being able to specify a list of serials, so that a Screen could bind to one output if it's present and another if it isn't. But truthfully I like what you're proposing better - I think it's both more flexible and easier to use/understand

I'm probably missing something but can't we do this currently (in this branch) using `get_output_info()`, like:
```
if len(qtile.core.get_output_info()) == 2: ...
```

tych0:
> I'm probably missing something but can't we do this currently (in this branch) using get_output_info(), like:
> if len(qtile.core.get_output_info()) == 2: ...

This is a bit of an ugliness in our config approach.

You can, once the qtile object exists. But qtile looks at e.g. the `screens` variable in your config before it ever instantiates the qtile object. (Well, not really, but that's the way it's currently organized.) So if you want to decide what to export as `screens` before we populate `libqtile.qtile`, you can't.

Ideally you'd be able to access some nontrivial subset of the API to figure stuff like this out, but we'd have to be very careful about what we export, since we wouldn't yet have a list of screens in the config or whatever.

`system_state` is one way to export all of this, but it is a bit ugly since we have to make a list of all the interesting things. Maybe there's another way?

acidkewpie:
Wayland - Are the CDEF updates for wlr_output not missing from build.py? Including modes for future things, this looks reasonable?

```
struct wl_list {
	struct wl_list *prev;
	struct wl_list *next;
};

struct wl_signal {
	struct wl_list listener_list;
};

struct wlr_output {
	char *name;
	char *description;
	char *make, *model, *serial;
	int32_t phys_width, phys_height;
	struct wl_list modes;
	struct wlr_output_mode *current_mode;
	int32_t width, height;
	int32_t refresh;
	bool enabled;
	float scale;
	int subpixel;
	int transform;
	int adaptive_sync_status;
	uint32_t render_format;
	bool adaptive_sync_supported;
    ...;
};

struct wlr_output_mode {
	int32_t width, height;
	int32_t refresh;
	bool preferred;
	int picture_aspect_ratio;
	struct wl_list link;
};
```

acidkewpie:
Running `qtile identify-output -b wayland` I just get a new WL-1 display, not my existing ones, looks like a new core is created? calling get_output_info() inside config.py does work (with the CDEF update)

tych0:
> Are the CDEF updates for wlr_output not missing from build.py?

Oh, they sure are. I guess pytest is not smart enough to rebuild the ffi bindings, so it probably just used the old version on my disk. Thanks.

> Running qtile identify-output -b wayland I just get a new WL-1 display, not my existing ones, looks like a new core is created? calling get_output_info() inside config.py does work (with the CDEF update)

So now I'm going to have to show my wayland ignorance here: how should this work? Why are the displays for the second core not the same as the displays for the first core?

richcarni:
> So now I'm going to have to show my wayland ignorance here: how should this work? Why are the displays for the second core not the same as the displays for the first core?

It looks like nested wayland servers don't get access to the underlying physical outputs

If you can't use, or want to avoid using the first core, I think the way to get the display info is via a client interface. Like a simplified, integrated version of wlr-randr. Any thoughts @jwijenbergh?

richcarni:
>calling get_output_info() inside config.py does work (with the CDEF update)

Yes, for me too! But it doesn't work on master - @tych0, have you modified something for this to work? (I'm curious for my own learning)

tych0:
> Yes, for me too! But it doesn't work on master - @tych0, have you modified something for this to work? (I'm curious for my own learning)

No, I was relying on the test suite. But the test suite doesn't actually spawn a qtile by virtue of the `manager_nospawn` fixture it uses, so it doesn't see this problem.

We could of course connect to the qtile for the `WAYLAND_DISPLAY`, but what happens if someone is running a *different* wayland wm? How do we connect to that in a general way? Or maybe we just don't, because this is a new wayland security thing?

tych0:
Hmm, just playing with it a bit, we probably can do it, but we'd have to link against `wayland-client` in order to make it work. That seems like a lot for just `identify-monitor`.

Perhaps I'll rework it to be an IPC call on wayland? That sort of limits the usefulness.

Maybe there is a command that shows all this stuff on wayland? The problem for x11 is that `xrandr` doesn't do all the EDID decoding like wayland does, so we need qtile's decoding to make it work.

EDIT: yes, it looks like wlr-randr *does* print all the EDID information.

I don't have a strong opinion here, I'm happy to drop the `identify-monitor` patch or not.

richcarni:
> Hmm, just playing with it a bit, we probably can do it, but we'd have to link against `wayland-client` in order to make it work. That seems like a lot for just `identify-monitor`.

I think you're right. At that point the x11 and wayland implementations have probably diverged enough to be their own scripts

> The problem for x11 is that `xrandr` doesn't do all the EDID decoding like wayland does, so we need qtile's decoding to make it work.
> 
> EDIT: yes, it looks like wlr-randr _does_ print all the EDID information.
> 
> I don't have a strong opinion here, I'm happy to drop the `identify-monitor` patch or not.

Sounds like x11-identify-monitor would be useful, and Wayland users can use wlr-randr?

tych0:
oh yeah, that's a good way to structure it. then we can delete it all when we delete x11. i'll change it to that tonight.

acidkewpie:
> Sounds like x11-identify-monitor would be useful, and Wayland users can use wlr-randr?

Is the scope of this just for getting this information outside of the main qtile process? I'm hopeful that we'll be able to get to a place where under wayland we can obtain all the information that wlr-randr currently provides, and then have simplistic function calls to arrange the screen as desired in config.py without needing wlr-randr at all. Lots of different angles on this though, so I'm thinking this ISN'T what is being discussed here, and that's all fine.

tych0:
> Is the scope of this just for getting this information outside of the main qtile process? 

Well, the rub is that there might not be a qtile process. e.g. if you are running `qtile identify-monitor` while using some other compositor to set up your config. IIUC we could make `qtile identify-monitor` handle this case, but we'd have to link against wayland-client, which is a lot just for functionality that `wlr-randr` already provides. So if we make this x11 only, we don't have the problem, and there is no real loss of functionality since `wlr-randr` can already do it.

tych0:
How do people feel about this PR now?

jwijenbergh:
> How do people feel about this PR now?

I will take a look soon

acidkewpie:
Purely on Wayland,  I'm happy with the progress, however I would still rather the Output class contained everything passed back to it rather than just holding the dimensions, serial and name. Totally get that that isn't / wasn't the fundamental objective of this PR, but assuming we move towards managing screen layout via config.py then I would need to know the transform values, and it seems like a waste to get the information half way and then drop it.

I still suck at git, esp as this is on a PR itself, so...

[moar_data.diff.txt](https://github.com/user-attachments/files/24208961/moar_data.diff.txt)

Again aware this is all past the core PR objective, but thought there could be breaking changes in providing the extra info. Doesn't actually look like it after all, but there's the (untested) code anyway.

If accepted, there's also then the fact that we'd be exposing the enabled state and as such would then be able to also include disabled outputs (totally for future here of course...) which I *think* would just be a case of removing the enabled check in qw_server_loop_output_dims(). It could also be an idea to shift the effective_resolution and such to the python side, and just give it the wlr_outputs and nothing else at all for clarity, as part of a natural progression given the swap already done on the serial string for the object... Anyway.

This all may make more sense as it's own PR, your call. If so, happy with everything as is.

libqtile/backend/wayland/cffi/build.py
```diff
@@ -108,6 +100,42 @@
...
+};
+
+struct wlr_output_mode {
+	int32_t width, height;
+	int32_t refresh;
+	bool preferred;
+	int picture_aspect_ratio;
+	struct wl_list link;
+};
+
```
richcarni: Would be nice to minimise this. Either just pass back serial and name or
```
struct wlr_output {
	char *name;
	char *description;
	char *make, *model, *serial;
    ...;
};

acidkewpie: Not from my perspective. The intent may not have been interpreted as such, but this was deliberately added to be able to reach this data in user land as part of the goal of replacing wlr-randr etc, for screen information. As above, maybe it's out of the scope for this PR, but I feel it's a shame to drop this potentially useful information on the cutting room floor when it's so easy to pass onwards.

tych0: > As above, maybe it's out of the scope for this PR, but I feel it's a shame to drop this potentially useful information on the cutting room floor when it's so easy to pass onwards.

I don't have a problem with exposing this information eventually, once someone has done the work of figuring out what the config file API should actually look like. But for now, I tend to agree with @richcarni, we should only expose information we want people to use.

I have some patches with combinators in my private tree for doing randr style config, but they are 1.5+ years old, as are a lot of the patches in this series. The whole thing could probably use a fresh set of eyes :)

richcarni: > As above, maybe it's out of the scope for this PR, but I feel it's a shame to drop this potentially useful information on the cutting room floor when it's so easy to pass onwards.

I could have explained better. It is partly that we're not using this information yet but also that we have tried to avoid cdef'ing wl_ and wlr_ structs. Iterating a wl_list in python is messy so we have used other patterns - iterate in c with a python callback, for example

As tych0 suggested, let's start with what the user interface, the config API, would look like and work form there - maybe create a discussion?

tych0:
Ok, hearing no major objections to this particular series, I'm going to land it as a step 0 toward implementing randr style binding.

I would love for us to expose all the info in a sane way, so that people can do what they want. This is the best idea I think I have had for that: https://github.com/qtile/qtile/pull/5274#issuecomment-3629228884 but probably someone else has an even better one. Looking forward to that discussion :)

PR #5277, opened 2025-06-12
Add `qtile repl`

elParaguayo:
Draft for now

Implements a basic repl to interact with qtile objects.
```
❯ ./bin/qtile repl
Connected to Qtile REPL
Press Ctrl+C to exit.
>>> qtile
<libqtile.core.manager.Qtile object at 0x78d11cf2e900>
>>> qtile.spawn("vlc")
401577
>>> ^C
Exiting REPL...
```

There are some bugs to check e.g. multiline functions don't render correctly. There's no history, tab completion. I'd like to make the port configurable.

Feedback would be appreciated.


tych0:
Neat! Can you use readline for the history and tab completion? Then people will get a bunch of other things for free I think.

elParaguayo:
> Neat! Can you use readline for the history and tab completion? Then people will get a bunch of other things for free I think.

Maybe... Need to look at that as this is using asyncio StreamReader/Writer.

elParaguayo:
I had to use a third party library `prompt_toolkit` to make this work nicely but it's looking nice:

https://github.com/user-attachments/assets/518e7663-245c-4b79-a449-597f28c7e899

libqtile/core/manager.py
```diff
@@ -1862,3 +1870,15 @@ def run_extension(self, extension: _Extension) -> None:
     def fire_user_hook(self, hook_name: str, *args: Any) -> None:
         """Fire a custom hook."""
         hook.fire(f"user_{hook_name}", *args)
+
+    @expose_command()
+    def start_repl_server(self) -> None:
+        create_task(repl_server.start(locals_dict={"qtile": self}, port=self.config.repl_port))
```
elParaguayo: @tych0 Should we give access to any other objects here, other than the qtile instance?

elParaguayo:
I think this may be one of my favourite PRs that I've ever submitted. I've wanted to implement this for so long.

libqtile/core/manager.py
tych0: IIUC, the qtile object implies access to pretty much everything else. e.g. `qtile.windows` etc., so I think that's pretty much the meat of it.

docs/manual/config/index.rst
```diff
@@ -145,6 +145,15 @@ configuration variables that control specific aspects of Qtile's behavior:
       - ``True``
       - Controls whether or not to automatically reconfigure screens when there
         are changes in randr output configuration.
+    * - ``auto_minimize``
+      - ``True``
+      - If things like steam games want to auto-minimize themselves when losing
+        focus, should we respect this or not?
+    * - ``repl_port``
+      - ``41414``
```
tych0: Is there a reason to bind this on tcp and not just a (different) unix socket?

tych0:
Overall I like it, pretty neat, especially with the completion library. I wish it was a unix socket interface instead of a tcp one, but it's a (relatively) minor point.

I wonder: could it even execute over the same unix socket as a new message type somehow?

libqtile/core/manager.py
elParaguayo: Great. That was my thinking too.

docs/manual/config/index.rst
elParaguayo: Only because that was the example code I found.

elParaguayo:
I can look at changing this to working over a unix socket but I'd prefer it to be a separate socket so we can avoid the (rare) situation where the client consumes messages that weren't meant for it.

Admittedly, socket programming is not my forte!

elParaguayo:
Tried converting it but I get `BrokenPipeErrors` every time I try to send a message from the client to the server!

Happy to accept a patch if you can make it work!

We could then remove changes to config file etc.

tych0:
Yeah, I can give it a try. I think we'll probably need to change the wire format, but I don't think that'll be a problem.

I've long thought we should wrap our wire format in some kind of length+message type buffer, so that we could do things like long running connections as well. This will be a good excuse to do that.

tych0:
Ok, I think what I'm going to do is going to be a little more... invasive than you want to wait for, so here's what I propose:

1. we drop all the port numbers etc. from the config file in this PR. Hard code everything, and I'll come along and rework it to work with messages over a unix socket in due time.
2. we say that, in general, the qtile wire format is not stable. What this means is that (1) you have to use it with python, and (2) with the same version of libqtile that you currently have installed. That likely means that external non-python tools cannot interact with us.

However, I don't think writing this down is really a big change: the wire format already relies on `import marshal`, which has the following disclaimer in its docs:

> Therefore, the Python maintainers reserve the right to modify the marshal format in backward incompatible ways should the need arise.

so essentially these restrictions are already true, we just need to enshrine them in the docs.

tldr: if you are comfortable dropping the port from the config, this is good for me now. I can come along and rework the messaging later.

tych0:
All that is to say: should we make a PR enshrining that we reserve the right to change the wire format? Maybe it's not necessary, since it's already implied and has not yet been a problem. I've just not thought about it before right now.

ervinpopescu:
> All that is to say: should we make a PR enshrining that we reserve the right to change the wire format? Maybe it's not necessary, since it's already implied and has not yet been a problem. I've just not thought about it before right now.

I for one think that it would be best. There were people around here (at least 2 😆) that use the wire format externally. I use it for https://github.com/ervinpopescu/qtile-cmd-client and https://github.com/ervinpopescu/qalttab (not a shameless plug, just an example).

elParaguayo:
Seems pretty shameless to me... 😉 

I'll update PR tonight if I can.

elParaguayo:
@tych0 were you thinking of something like protobuf for the rewrite e.g. so we can define a schema for messages?

tych0:
I wasn't, but maybe we should. 

I hadn't realized anyone else was using the wire format! It looks like you're using the json output, so maybe there's a way we can do json and everything else without bringing protobuf into it?

My initial work had just been to add a message type to the header in addition to length, and make it so that you could pass more than one message per connection.

elParaguayo:
I'll defer to you on the best approach. I was just wondering if it's worth having some form of defined schema in case people want to make their own tools.

tych0:
This turned into a longer brain dump than I was thinking, proceed with caution :)

Possibly, let me draw a distinction between a few things:

1. the contents of the messages, e.g. for a particular layout:
```
layout > info()
{'clients': ['Alacritty'],
 'current_stack': 0,
 'group': 'a',
 'name': 'stack',
 'stacks': [{'clients': ['Alacritty'], 'current': 0, 'split': False},
            {'clients': [], 'current': 0, 'split': False}]}
```
so a layout may have exposed stuff like `current_stack`, or a bar widget may have custom fields, etc.

2. the format that this content is transmitted over the wire, of which we have two now: raw json and the format used by the `marshal` library, which is not really documented other than that it can change at any time :). IIUC, the way the server works today is the following: if it can decode the message someone sent as json, it will speak json back to them, otherwise it will fall back to `marshal`

While I can see how a schema for (1) would be useful, but I would like to try and avoid defining one, because it would be yet another maintenance headache as people refactor widgets and layouts. We'd have to test it, etc. and I just don't think we have that many external consumers.

What I can see enshrining is the wire format in (2). I didn't realize anyone was speaking to us externally via the json engine, but in retrospect it seems obvious and obviously useful: why add anything besides the `marshal` format if you're not going to use it from external tooling? It also seems obviously useful to have a json format to talk to qtile: every language has json serializers and parsers close to their standard library, so you can write code to interface with qtile in exactly the way you want using whatever language, so long as you're willing to pattern match all the json gunk.

3. we really have only supported one mechanism for interacting with qtile up to this point: the command tree. This is pretty complex and intuitive IMO, and I sense that you agree based on your comments in the commit messages for this PR :). However, we do need to keep supporting it, because it's what people use, and have built tooling around. And yet: this PR adds a new "message type", the "repl type", which is distinct from the "command tree" type of message. So we need some sort of abstraction to make this distinction.

Given these three things, I draw two conclusions:

a. it is definitely useful to have an easily parseable message like json. We should keep this.
b. we necessarily need to make some change to the wire format to add the new "repl message type" distinction somehow.

So how should we proceed? Assuming you take as an axiom my point (1) above, that we don't want to do a protobuf layout/json schema for our whole object tree, here are three possibilities:

x. we define a protobuf message like this:
```
enum message_type {
    cmd_obj = 0;
    repl = 1;
}

message QtileMessage {
    message_type type = 1;
    string content = 2; // this is where we put json
}
```

y. we do something fancy with `MSG_OOB`; this means we have to move away from unix domain sockets, as they do not support `MSG_OOB`, to binding on a port. But you can imagine the server listening for `MSG_OOB`, and if it sees one, that means "speak this other message type"; the default would remain command-obj json. The advantage of this is that nobody has to change their existing parsers, but they do have to change their code to connect over TCP instead of unix sockets.

z. we make a one-time change to the message format: it is *always* json, and has a top level structure of `{"message_type": "command", "content": <the old json tree>}`; and for your new thing, we can add a new message type of "repl", and stuff all your messages above in the "content" section. This of course requires people to change their parsers, but it is a hopefully simple change, and it will only be one time.

Of these, I prefer (z).

(One final point: I would like to make it so that we can do more than one message per connection. If we go with (z), we'll have to teach python's json parser to stop when it encounters a final `}` for the top level message, but I don't think that should be too hard. Or we could do some delimiter (null byte?), or length-prefix, or something. But I suspect most language parsers can be configured to stop reading when exactly one object is parsed, and so that would be the easiest thing.)

elParaguayo:
OK. Can we merge this as is and carry on the discussion (although, we're hopefully near a conclusion)?

elParaguayo:
I'm going to show my ignorance now, but bear with me...

Your logic above makes sense to me and so I can see why (z) appeals. The questions I've got are as follows:

1) What happens when you have multiple clients connected to a socket that are both trying to read at the same time? e.g. if there's a `repl` session but another app runs `qtile cmd-obj`, could the `repl` client consume the message intended for the `cmd-obj`? That needs to be avoided.
2) Slightly related to the above, the `repl` client needs to preserve state between messages. i.e. a user can define a new variable in the session which would be local to that session. However, once the client has exited those local variables shouldn't be preserved. That seemed to work well the tcp connection and the `handle_client` method in this PR and I wouldn't want to lose this if we end up using the same socket for all messaging.

tych0:
> could the repl client consume the message intended for the cmd-obj?

No, each connection has a separate fd associated, separate send/receive buffers, etc. so this won't happen. We'll have a separate "thread" for each connection in qtile (not really, due to asyncio magic, just more fds in the epoll loop) that will maintain state about who it is talking to and what kind of messages they've asked for.

> Slightly related to the above, the repl client needs to preserve state between messages. i.e. a user can define a new variable in the session which would be local to that session. However, once the client has exited those local variables shouldn't be preserved.

Yep, that should work as-above. We'll have one separate thread per repl connection, and all the state should be freed (bugs in the implementation not withstanding :) at the end.

> OK. Can we merge this as is and carry on the discussion (although, we're hopefully near a conclusion)?

Yeah, absolutely, I'll do it right after I post this message. Thanks for the quick config change.

elParaguayo:
Great. If that all works then I'm fine with that approach.

tych0:
Ok, I know I've just made an argument for (z), but I didn't realize something: json encoding escapes newlines:
```
>>> json.dumps("foo\nbar")
'"foo\\nbar"'
```
so I think for a variety of reasons, newline delimited will be easier, vs. just straight objects next to each other, so I'll go ahead and implement that assuming nobody strongly objects here.

PR #5280, opened 2025-06-14
nix: update flake, add disabled attribute to overlay

Gurjaka:
* Updates the `nix` flake input to the latest available version.
* Adds a `disabled = true;` attribute to the flake overlay to comply with recent overlay evaluation changes.

#### Motivation

As of [[nixpkgs#414609](https://github.com/NixOS/nixpkgs/issues/414609)](https://github.com/NixOS/nixpkgs/issues/414609), overlays without an explicit `disabled` attribute are evaluated unconditionally, which can break evaluation or cause side effects even when not used. This PR ensures the flake overlay is opt-in by default by setting `disabled = true`.

#### Details

* The flake input for `nix` has been updated to align with upstream improvements and bug fixes.
* The overlay now includes `disabled = true;` to prevent unintended evaluation when the overlay is unused.

#### Notes

* This change is backwards compatible.
* The overlay now includes `disabled = true;` to prevent unintended evaluation during flake output traversal. It will still work normally when explicitly added to `nixpkgs.overlays`

#### Related

* Closes [[#414609](https://github.com/NixOS/nixpkgs/issues/414609)](https://github.com/NixOS/nixpkgs/issues/414609)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1747327360,
-        "narHash": "sha256-LSmTbiq/nqZR9B2t4MRnWG7cb0KVNU70dB7RT4+wYK4=",
+        "lastModified": 1749794982,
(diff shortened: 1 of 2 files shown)


jwijenbergh:
Thanks!

PR #5287, opened 2025-06-16
Set default background logo

tych0:


elParaguayo:
Yes, there's a qtile reddit. It's unofficial but I try to monitor it when I can.

tych0:
Totally. It's a great source of low hanging obvious fails that we could fix it seems like. I'll try to drip feed patches. But for now I'm working on re-working IPC :)

elParaguayo:
Nice. Let me know if you want me to test anything.

PR #5292, opened 2025-06-19
qtile top: note command to start tracemalloc

tych0:
...I wonder if instead we shouldn't just default --force-start to true... if you typed `qtile top`, surely you want to see allocations?


elParaguayo:
Do you think anyone's using `qtile top`?

tych0:
No idea, I just fired it up to see if it worked, and then I had to figure out how to turn tracemalloc on :)

I did think the results were interesting, there's a bunch of stuff we can probably drop to make event handling faster. But it doesn't seem like a huge issue in practice, I haven't really noticed any lag.

elParaguayo:
In which case, let's go with your idea to start `tracemalloc` automatically when someone uses `qtile top`. Seems very reasonable.

PR #5293, opened 2025-06-19
TaskList: add support for reading Desktop files

elParaguayo:
Some apps provide a path to their icon in .desktop files. This PR adds support for reading that icon path.

### DIFF
--- a/libqtile/widget/tasklist.py
+++ b/libqtile/widget/tasklist.py
@@ -23,10 +23,13 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import os
+from pathlib import Path
 
 import cairocffi
 
 try:
+    from xdg.DesktopEntry import DesktopEntry
     from xdg.IconTheme import getIconPath
 
     has_xdg = True
@@ -38,6 +41,14 @@ from libqtile import hook, pangocffi
 from libqtile.images import Img
 from libqtile.log_utils import logger
 from libqtile.widget import base
+
+data_home = os.environ.get("XDG_DATA_HOME", "~/.local/share")
+
+DESKTOP_LOCATIONS = [
+    Path(data_home) / "applications",
+    Path("/usr/local/share/applications"),
+    Path("/usr/share/applications"),
+]
 
 
 class TaskList(base._Widget, base.PaddingMixin, base.MarginMixin):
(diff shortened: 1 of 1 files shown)


libqtile/widget/tasklist.py
```diff
@@ -506,6 +515,34 @@ def _get_class_icon(self, window):
 
         return img
 
+    def _read_desktop_file(self, desktopfile):
+        icon = DesktopEntry(desktopfile).getIcon()
+        if icon and Path(icon).expanduser().exists():
+            return icon
+        print(icon, Path(icon).expanduser().exists())
```
jwijenbergh: print here ;)

elParaguayo: :man_facepalming:

elParaguayo:
Feels like there's a good amount of duplicate code in this one which could be refactored.

libqtile/widget/tasklist.py
```diff
@@ -38,6 +41,12 @@
 from libqtile.log_utils import logger
 from libqtile.widget import base
 
+DESKTOP_LOCATIONS = [
+    Path("~/.local/share/applications."),
```
ervinpopescu: https://wiki.archlinux.org/title/XDG_Base_Directory advocate here :)

Please consider using `$XDG_DATA_HOME` instead of `~/.local/share`. That said, not even Archwiki seems to know if `$XDG_DATA_HOME` is read by hopefully most apps similarly to what I am asking and it uses `~/.local/share/applications` all over the place instead of `$XDG_DATA_HOME/applications`, but chatgpt answered yes to the below question 😃:

```
do most apps read from $XDG_DATA_HOME/applications, if $XDG_DATA_HOME is set?
``` 

AND https://wiki.archlinux.org/title/Desktop_entries#Modify_desktop_files references `XDG_DATA_HOME`, at least.

Thanks for reading through!

tych0: agreed, we already use XDG data and cache dirs elsewhere, no reason not to use one here.

elParaguayo: No problem.

For what it's worth, I'm not 100% convinced by this PR. I'm trying to see if it solves an issue reported by a user on discord.

elParaguayo: Interestingly, I have .desktop files in that folder but `$XDG_DATA_HOME` is not set for me.

ervinpopescu: Yeah, most apps use `~/.local/share` if the var is not set, is that what you mean? Sorry if I don't get it :)

elParaguayo: I thought your comment was to _only_ use `$XDG_DATA_HOME` for the user folder. My point was that wouldn't have worked for me but I realised I needed the fallback option anyway.

ervinpopescu: Ah yes, that looks fine to me as well.

    XDG_DATA_HOME
        Where user-specific data files should be written (analogous to /usr/share).
        Should default to $HOME/.local/share.

tych0 (approved):

ervinpopescu:
Anything holding this one back?

elParaguayo:
No. Looks like I just forgot about it!

Need to fix the conflict though. I can take a look later.

Issue #4805, opened 2024-05-05
Dialogs are rendered behind apps sometimes


nazar-pc:
### Issue description

Reproduction:
1) Open one project in IDE (IntelliJ)
2) Open another project in IDE in a separate window
3) Open search dialog in a second window

Expected:
Search dialog window opens above other windows

Actual:
Search dialog opens above second window, but behind the first window somehow. Clicking or re-focusing on it fixes the issue for that particular instance.

This wasn't the case in older verison of Qtile (0.21 at least). This also happens with other apps like Geany for example. In some cases the dialog is completely hidden, which is hard to work with.

### Version

0.25.0

### Backend

X11 (default)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

elParaguayo:
Can you see if #4583 fixes this?

nazar-pc:
Could you rebase it on latest `master`, please?

elParaguayo:
Done.

nazar-pc:
I'm getting some errors unfortunately:
<details>

```
2024-05-05 14:40:38,469 ERROR libqtile core.py:_xpoll():L372 Got an exception in poll loop
Traceback (most recent call last):
  File "/home/nazar-pc/.local/share/pipx/venvs/qtile/lib/python3.12/site-packages/libqtile/backend/x11/core.py", line 347, in _xpoll
    self.handle_event(event)
  File "/home/nazar-pc/.local/share/pipx/venvs/qtile/lib/python3.12/site-packages/libqtile/backend/x11/core.py", line 314, in handle_event
    ret = target(event)
          ^^^^^^^^^^^^^
  File "/home/nazar-pc/.local/share/pipx/venvs/qtile/lib/python3.12/site-packages/libqtile/backend/x11/core.py", line 781, in handle_DestroyNotify
    if self.qtile.current_window is None:
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nazar-pc/.local/share/pipx/venvs/qtile/lib/python3.12/site-packages/libqtile/command/base.py", line 281, in __getattr__
    raise AttributeError(f"{self.__class__} has no attribute {name}")
AttributeError: <class 'libqtile.core.manager.Qtile'> has no attribute current_window
```

</details>

I then tried main and my config failed to load:
<details>

```
[...]

PR #5297, opened 2025-06-21
Plasma - client restoration fix

### LINKED ISSUE #4915: A bug when combining ScreenSplit and Plasma
Finally tested and it is reproducible with Plasma layout, but not Tile for example.
Now that both are upstream it makes sense to create a proper issue.

Here is relevant part of the config:
<details>

```python
primary_match = [
    Match(wm_class='firefox'),
]

main_layout = ScreenSplit(
    name="Main",
    splits=[
        {
            "name": "left",
            "rect": (0, 0, 0.15, 1),
            "layout": Plasma(
                insert_position=1,
            ),
        },
        {
            "name": "top",
            "rect": (0.15, 0, 0.6, 0.4),
            "layout": Plasma(
                insert_position=1,
            ),
        },
        {
            "name": "bottom",
            "rect": (0.15, 0.4, 0.6, 0.6),
            "layout": Plasma(
                insert_position=1,
            ),
            "matches": primary_match,
        },
        {
            "name": "right",
            "rect": (0.75, 0, 0.25, 1),
            "layout": Plasma(
                insert_position=1,
            ),
        },
    ],
)

layouts = [
    main_layout,
]
```

</details>

Reproduction steps:
* Open Firefox
    * Opens in `bottom` split
    * Make sure this is the only app in this split
* Open anonymous window in Firefox
* Move main Firefox window to a different split (for example `top`)
    * Anonymous window remains in `bottom` and is the only window there
* Close anonymous window

Observe layout being broken:
1) I can no longer move Firefox window between splits, it doesn't get activated
2) 50% of `bottom` split is occupied by "nothing" like if there is an app there, but there isn't (even though private window before that occupied 100% of the split)

_Originally posted by @nazar-pc in https://github.com/qtile/qtile/issues/3793#issuecomment-1272122365_

elParaguayo:
The Plasma layout tries to restore clients to their original position when they're returned to the layout (e.g. from floating back to tiled). However, there is a bug where this restoration happens after any sibling clients in the same node have been removed. The layout tries to resize the sibling too and crashes because it can't find it.

Fixes #4915

### DIFF
--- a/libqtile/layout/plasma.py
+++ b/libqtile/layout/plasma.py
@@ -18,6 +18,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import contextlib
 import copy
 import time
 from enum import Enum, Flag, auto
@@ -135,6 +136,7 @@ class Node:
         self.last_accessed = 0
         self.parent = None
         self.restorables = {}
+        hook.subscribe.client_killed(self._clean_restorables)
 
     def __repr__(self):
         info = self.payload or ""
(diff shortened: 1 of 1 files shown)


nazar-pc:
Seems to work fine for me, thanks!

Any chance you'll be able to look into https://github.com/qtile/qtile/issues/4805 as well?

nazar-pc (approved):

elParaguayo:
Thanks. I've merged that. I'll look at the other one soon.

PR #5298, opened 2025-06-25
docs: move to recommending `uv`

### LINKED ISSUE #5294: Shipping broken in Debian (and wrong installation instructions)
### Issue description

Qtile is completely broken on debian. 

First the installation docs are wrong, according to this page https://docs.qtile.org/en/latest/manual/install/ubuntu.html

It says to do the following

```bash
sudo apt install xserver-xorg xinit
sudo apt install libpangocairo-1.0-0
sudo apt install python3-pip python3-xcffib python3-cairocffi
pip install qtile
```

This will give you an error about installing with an externally-managed-environment

You can work around this with pipx

```
pipx install qtile
```

And that got it working, but after all this a terminal won't even open, despite copying and pasting the exact default config and using mod4 + return with xterm installed. Nothing happens.

### Version

latest

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

tych0:
I have been using uv for a while, and it's great. It is much cleaner than managing venvs, passing --break-system-packages, or whatever other hacks people have been doing. We should recommend people use it.

Fixes #5294

### DIFF
--- a/docs/manual/howto/git.rst
+++ b/docs/manual/howto/git.rst
@@ -50,7 +50,7 @@ The numeric pull request id can be found in the url or next to the title
 
 .. note:: Having the feature branch checked out doesn't mean that it is
    installed and will be loaded when you restart qtile. You might still need to
-   install it with ``pip``.
+   install it with ``uv``.
 
 I committed changes and the tests failed
 ========================================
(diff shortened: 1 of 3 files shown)


tych0:
Hearing no major objections, I'll go ahead and land this. At some point I'll probably also work all of our CI to move away from tox etc, and "just" use `uv`, because I can never exactly understand how all of it works. But that is (very) future work :)

PR #5299, opened 2025-06-25
New widget CurrentLayoutIconOrText

RFCreate:
Either display the icon of the current layout or its name, imagine switching between `CurrentLayout` and `CurrentLayoutIcon` with a click.

I like to have the icon displayed with the `CurrentLayoutIcon` widget in my bar but sometimes I forget what the icon meant, and I end up adding the `CurrentLayout`. This would save space in the bar since I can have only display the icon and when I'd like to know more then simply right-click and I will get the current layout name.

### DIFF
--- a/libqtile/widget/currentlayout.py
+++ b/libqtile/widget/currentlayout.py
@@ -222,3 +222,48 @@ class CurrentLayoutIcon(CurrentLayout):
             self.surfaces[layout_name] = img
 
         self.icons_loaded = True
+
+
+class CurrentLayoutIconOrText(CurrentLayoutIcon):
+    """
+    Either display the icon of the current layout or its name, imagine switching
+    between CurrentLayout and CurrentLayoutIcon with a click.
+    """
+
+    defaults = [
+        ("draw_icon_first", True, "Should draw icon or text when bar is initialized."),
+    ]
+
+    def __init__(self, **config):
+        super().__init__(**config)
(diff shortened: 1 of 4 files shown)


libqtile/widget/currentlayout.py
```diff
@@ -239,6 +223,47 @@ def _setup_images(self):
...
-        base._TextBox.finalize(self)
+
+class CurrentLayoutIconOrText(CurrentLayoutIcon):
+    """
+    Either display the icon of the current layout or its name, imagine switching
+    between CurrentLayout and CurrentLayoutIcon with a click.
+    """
+
+    defaults = [
+        ("draw_icon_first", True, "Should draw icon or text when bar is initialized."),
```
tych0, line 234: Is it possible to add this config to `CurrentLayout` and collapse the existing `CurrentLayoutIcon` into it as well, instead of adding a whole new widget?

elParaguayo: Beat me to it. We definitely don't want/need 3 layout widgets. 

I've got similar things in qtile-extras where you can choose icon or text modes or have text appear when you click on the icon (and then disappear after a timeout).

That approach would be good here but we may need a migration too if we remove a widget.

RFCreate: I reworked it in a new PR #5302.

tych0:
Closing this in favor of #5302

PR #5300, opened 2025-06-26
rework logging path handling

### LINKED ISSUE #5286: default log dir `mkdir` fails with both `FileNotFoundError` and `FileExistsError`
### Issue description

While installing qtile in CI on my self-hosted k8s runners, I intermittently ran into the below errors.

[failing job on py311](https://github.com/ervinpopescu/qtile-cmd-client/actions/runs/15653356397/job/44101071386)

At first glance, it might look like there's some interference between the runners which are somehow accessing either `~/.local/share` (which is my `$XDG_DATA_HOME`) or `~/.local/share/qtile` and messing it up somehow (which is also quite impossible since the runners don't share `~` anyway...).

But if we look at the other runs [py312](https://github.com/ervinpopescu/qtile-cmd-client/actions/runs/15653356397/job/44101071375) and [py313](https://github.com/ervinpopescu/qtile-cmd-client/actions/runs/15653356397/job/44101071379), we notice they haven't reached that CI step yet.

Some extra logging did not really help much:
```diff
$ git diff HEAD~
diff --git a/libqtile/log_utils.py b/libqtile/log_utils.py
index 539a6241..cb7c3f30 100644
--- a/libqtile/log_utils.py
+++ b/libqtile/log_utils.py
@@ -85,8 +85,16 @@ def get_default_log() -> Path:
         data_directory = os.path.expanduser("~/.local/share")

     qtile_directory = Path(data_directory) / "qtile"
-    if not qtile_directory.exists():
-        qtile_directory.mkdir(parents=True)
+    if qtile_directory.exists():
+        print("Exists:", qtile_directory)
+        print("Is dir:", qtile_directory.is_dir())
+        print("Is file:", qtile_directory.is_file())
+        print("Is symlink:", qtile_directory.is_symlink())
+        print("os.stat:", os.stat(qtile_directory))
+    if qtile_directory.exists() and not qtile_directory.is_dir():
+        # Clean up the invalid file/symlink
+        qtile_directory.unlink()
+    qtile_directory.mkdir(parents=True, exist_ok=True)

     return qtile_directory / "qtile.log"

```
```bash
Exists: /home/runner/.local/share/qtile
Is dir: True
Is file: False
Is symlink: False
os.stat: os.stat_result(st_mode=16877, st_ino=12059637, st_dev=1048676, st_nlink=2, st_uid=1001, st_gid=1001, st_size=4096, st_atime=1750003804, st_mtime=1750003804, st_ctime=1750003804)
```

`exist_ok=True` + the unlinking seems to fix it, I'll open a PR after I figure out how reliable and/or dependent it is with and without or to the unlinking. 

I'm still pretty confused why this could come up. 

Also not that happy that even though, when users set `--logfile / -p`, `get_default_log` still gets executed and creates the dir. Maybe we can do something about that as well?

### Version

0.32.0

### Backend

X11 (default)

### Config

```python
default config
```

### Logs

```text
  File "/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/pathlib.py", line 1116, in mkdir
    os.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.local/share/qtile'

  File "/home/runner/qtile/./build_dir/home/runner/qtile/venv/bin/qtile", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/runner/qtile/libqtile/scripts/main.py", line 40, in main
    default=get_default_log(),
            ^^^^^^^^^^^^^^^^^
  File "/home/runner/qtile/libqtile/log_utils.py", line 89, in get_default_log
    qtile_directory.mkdir(parents=True)
  File "/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/pathlib.py", line 1121, in mkdir
    self.mkdir(mode, parents=False, exist_ok=exist_ok)
  File "/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/pathlib.py", line 1116, in mkdir
    os.mkdir(self, mode)
FileExistsError: [Errno 17] File exists: '/home/runner/.local/share/qtile'
```

### Required

tych0:
1. do not unconditionally mkdir the default log path, even if people have specified a different one
2. do allow people to specify a path that doesn't exist, we can just create the directories for them

Fixes #5286

### DIFF
--- a/libqtile/log_utils.py
+++ b/libqtile/log_utils.py
@@ -85,9 +85,6 @@ def get_default_log() -> Path:
         data_directory = os.path.expanduser("~/.local/share")
 
     qtile_directory = Path(data_directory) / "qtile"
-    if not qtile_directory.exists():
-        qtile_directory.mkdir(parents=True)
-
     return qtile_directory / "qtile.log"
 
 
@@ -111,6 +108,13 @@ def init_log(
 
     else:
         # Otherwise during normal usage, log to file.
+        try:
+            log_path.parent.mkdir(parents=True, exist_ok=True)
+        except PermissionError as e:
+            # ok, only one place to write: stderr
+            print(f"couldn't mkdir {log_path.parent}: {e}", file=sys.stderr)
+            os._exit(1)
+
         handler = RotatingFileHandler(
             log_path,
             maxBytes=log_size,
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -26,14 +26,6 @@ except PackageNotFoundError:
     VERSION = "dev"
 
 
-def check_folder(value):
-    path = Path(value)
-    if not path.parent.is_dir():
-        raise argparse.ArgumentTypeError("Log path destination folder does not exist.")
-    # init_log expects a Path object so return `path` not `value`
-    return path
-
-
 def main():
     parent_parser = argparse.ArgumentParser(add_help=False)
     parent_parser.add_argument(
@@ -50,7 +42,7 @@ def main():
         "--log-path",
         default=get_default_log(),
         dest="log_path",
-        type=check_folder,
+        type=Path,
         help="Set alternative qtile log path",
     )
     main_parser = argparse.ArgumentParser(


libqtile/log_utils.py
```diff
@@ -111,6 +108,7 @@ def init_log(
 
     else:
         # Otherwise during normal usage, log to file.
+        log_path.parent.mkdir(parents=True, exist_ok=True)
```
elParaguayo: Should we have some exception handling here in case the user has specified a path without write access?

ervinpopescu:
Thanks @tych0, really busy with work these days...

tych0:
Yeah, no worries! Thanks for the report.

libqtile/log_utils.py
tych0: yeah, good point, i'll make the change when i'm back at the keyboard.

ervinpopescu:
This works for my usecase FWIW.

PR #5302, opened 2025-06-26
Merge CurrentLayoutIcon into CurrentLayout

RFCreate:
Closes #5299

### DIFF
--- a/libqtile/widget/currentlayout.py
+++ b/libqtile/widget/currentlayout.py
@@ -35,46 +35,8 @@ from libqtile.widget import base
 
 class CurrentLayout(base._TextBox):
     """
-    Display the name of the current layout of the current group of the screen,
-    the bar containing the widget, is on.
-    """
-
-    def __init__(self, width=bar.CALCULATED, **config):
-        base._TextBox.__init__(self, "", width, **config)
-
-    def _configure(self, qtile, bar):
-        base._TextBox._configure(self, qtile, bar)
-        layout_id = self.bar.screen.group.current_layout
-        self.text = self.bar.screen.group.layouts[layout_id].name
-        self.setup_hooks()
-
-        self.add_callbacks(
-            {
-                "Button1": qtile.next_layout,
-                "Button2": qtile.prev_layout,
-            }
-        )
-
-    def hook_response(self, layout, group):
-        if group.screen is not None and group.screen == self.bar.screen:
-            self.text = layout.name
-            self.bar.draw()
-
-    def setup_hooks(self):
-        hook.subscribe.layout_change(self.hook_response)
-
-    def remove_hooks(self):
-        hook.unsubscribe.layout_change(self.hook_response)
-
-    def finalize(self):
-        self.remove_hooks()
-        base._TextBox.finalize(self)
-
-
-class CurrentLayoutIcon(base._TextBox):
-    """
-    Display the icon representing the current layout of the
-    current group of the screen on which the bar containing the widget is.
+    Display the icon or the name of the current layout of the current group
+    of the screen on which the bar containing the widget is.
 
     If you are using custom layouts, a default icon with question mark
     will be displayed for them. If you want to use custom icon for your own
@@ -95,6 +57,7 @@ class CurrentLayoutIcon(base._TextBox):
     orientations = base.ORIENTATION_HORIZONTAL
 
     defaults = [
+        ("icon_first", False, "Should draw icon or text when bar is initialized."),
         ("scale", 1, "Scale factor relative to the bar height. Defaults to 1"),
         (
             "custom_icon_paths",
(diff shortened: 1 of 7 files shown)


RFCreate:
The Nix error is caused by `qtile-extras` which should be outside the scope of this repo.

The X11 error, I'm still checking on that, but it may have to do with the previous `CurrentLayoutIcon` and not with the new `CurrentLayout` implementation. I can confirm this by getting the same error with this changes in the master branch.

```diff
diff --git a/test/configs/reloading.py b/test/configs/reloading.py
index c5ddf28e..1ed0b093 100644
--- a/test/configs/reloading.py
+++ b/test/configs/reloading.py
@@ -67,7 +67,7 @@ if hasattr(qtile, "test_data"):
     layouts.append(layout.Max())
 
     screens = [
-        Screen(top=bar.Bar([widget.CurrentLayout()], 32)),
+        Screen(top=bar.Bar([widget.CurrentLayoutIcon()], 32)),
     ]
 
     widget_defaults["background"] = "#ff0000"
diff --git a/test/test_manager.py b/test/test_manager.py
index 430a77b7..3fbb8048 100644
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -1211,7 +1211,7 @@ def test_reload_config(manager_nospawn):
     assert manager_nospawn.c.eval("len(self._mouse_map)") == (True, "2")
     assert "".join(manager_nospawn.c.get_groups().keys()) == "123456789S"
     assert len(manager_nospawn.c.group.info()["layouts"]) == 2
-    assert manager_nospawn.c.widget["currentlayout"].eval("self.background") == (True, "#ff0000")
+    assert manager_nospawn.c.widget["currentlayouticon"].eval("self.background") == (True, "#ff0000")
     screens = manager_nospawn.c.get_screens()[0]
     assert screens["gaps"]["top"][3] == 32 and not screens["gaps"]["bottom"]
     assert len(manager_nospawn.c.internal_windows()) == 1
```

tych0:
Agreed, we should not be failing CI on qtile-extras problems. @jwijenbergh or @Gurjaka I think know this code?

Gurjaka:
> Agreed, we should not be failing CI on qtile-extras problems. @jwijenbergh or @Gurjaka I think know this code?

Yup, I am the one who made `nix flake check` CI. The issue comes from nixpkgs where we have single test for both Qtile and Qtile-Extras, which is planned to be separated soon (I'll make PR today if I find time). I don't know about other CIs but Nix one can be ignored.

jwijenbergh:
If you rebase the PR, the nix test failure should be gone

Gurjaka:
> If you rebase the PR, the nix test failure should be gone

I doubt it, if you are coming from the PR you merged about Qtile-Extras, it wasn't fixing this issue.

tych0:
Wait - the `Nix Flake Check` failure should be gone right? Because you removed the `qtile-extras` overlay?

There is another problem (potentially pre-existing that this PR uncovered), but we can address that separately. IIUC the nix check thing should be fixed by 544887c36b639d67df7f123f42852d140716250d

Gurjaka:
> Wait - the `Nix Flake Check` failure should be gone right? Because you removed the `qtile-extras` overlay?
> 
> There is another problem (potentially pre-existing that this PR uncovered), but we can address that separately. IIUC the nix check thing should be fixed by [544887c](https://github.com/qtile/qtile/commit/544887c36b639d67df7f123f42852d140716250d)

No you understood it wrong, that commit is to get rid of old overlay for broken Qtile-Extras tests (we removed it because @elParaguayo fixed them in `0.32.0` release). In order to fix this specific `nix-flake-check` error, we got two options:
1) Make a PR on nixpkgs that separates Qtile-Extras and Qtile tests (ideal, I have already done it, but haven't made pr yet, because don't know what to name commit or PR, nixpkgs is sensitive to commit naming)
2) make a temporary overlay in Qtile's flake that will ignore current Qtile-Extras test related to this PR.

First one is ideal but will take time, because merging in nixpkgs is quite slow, but we could use help from @Sigmanificient since he has merge rights, I tried to contact him for commit name suggestion, but I think he is busy/not available for now.

tych0:
I see, can you send a PR for (2)? Alternatively, we can just drop the nix flake check until it is sorted out, but I don't want to block merging stuff on this.

Gurjaka:
> I see, can you send a PR for (2)? Alternatively, we can just drop the nix flake check until it is sorted out, but I don't want to block merging stuff on this.

I can send it, give me about 10 minutes and I will tag you on it

jwijenbergh:
> > If you rebase the PR, the nix test failure should be gone
> 
> I doubt it, if you are coming from the PR you merged about Qtile-Extras, it wasn't fixing this issue.

My bad

Gurjaka:
PR is now opened. If anything still breaks after this is merged and the other PR gets rebased, feel free to tag me — I’ll do my best to help with the Nix side of things.

RFCreate:
Even though it is fixed, that x11 issue didn't make sense, why is it that a drawer problem prevent the "bar window" from being **unmanaged** in reload? Testing I found out that wayland had the same _drawer-integer problem_ but it did **unmanaged** the "bar window".

EDIT: Solved!

libqtile/widget/currentlayout.py
```diff
@@ -102,7 +102,7 @@ def current_layout(self):
 
     @property
     def length(self):
-        return self.image_length if self.draw_icon_first else self.calculate_length()
+        return int(self.image_length if self.draw_icon_first else self.calculate_length())
```
tych0: Weird... which is it that doesn't return an int? calculate_length() or image_length?

RFCreate: Both actually, it depends on the size of the bar and the `actual_padding`.

tych0: Oh, that is annoying. Does this fix it?
```
index e49fe2c2..0b172cef 100644
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -577,7 +577,7 @@ class _TextBox(_Widget):
     @property
     def actual_padding(self):
         if self.padding is None:
-            return self.fontsize / 2
+            return int(self.fontsize / 2)
         else:
             return self.padding
```

or probably we should use `//`. But I'd like to fix it for all consumers of this if possible, not just here.

RFCreate: If I set `padding=1.5`, meaning a float, it will still break if the length property is not wrap in int(). Actually the base class has it that way (more or less).

https://github.com/qtile/qtile/blob/eb4872ae1ad86692ba638f136793fa4616b573f1/libqtile/widget/base.py#L178-L182

tych0: Oh, I see. I wonder if we should just do int() at config time somehow. We can't do subpixel padding anyway.

RFCreate:
I ended overwriting `calculate_length()` instead of messing with the `length`'s getter and setter which broke things in the first place.

> Oh, I see. I wonder if we should just do int() at config time somehow. We can't do subpixel padding anyway.

After thinking it through, you cannot save everyone from breaking their configs with wrong values, it is going to happen and it is up to them to fix them, `padding` as float will break as well as `fontsize` as float will break, most of the options set as a different value type will break and it cannot be helped.

tych0:
> After thinking it through, you cannot save everyone from breaking their configs with wrong values, it is going to happen and it is up to them to fix them, padding as float will break as well as fontsize as float will break, most of the options set as a different value type will break and it cannot be helped.

Yeah, it's a fair point. Thanks for adding the patch to use integer division, but otherwise it's all user config values. If we're going to type check user config values, we should do it for everything, not just this value, so we need a more generic solution.

This whole series looks good to me, I'll leave it around for @elParaguayo in case they have any more comments, but otherwise I will merge in a day or two. Thanks for the cleanups!

tych0 (approved):

elParaguayo:
@tych0 @RFCreate I know I'm late here but I didn't get round to commenting before this got merged.

While I think combining both widgets into one is a good idea, my main problem is that we've made this an icon or text widget. There are users who have both the icon and the name (using both widgets) so we should cater for those users and allow the widget to display both.

I also find the name `icon_first` very confusing. To me that is asking whether the icon should be displayed before or after the text. However, it seems to be toggling whether we display an icon or text. I would have thought we want to have an option for `mode` which accepts `"icon"`, `"text"` or `"both"`. `icon_first` is then applied when using `mode="both"` to determine where the icon is displayed.

elParaguayo:
This also breaks one of the widgets in qtile-extras. That's not a big issue in itself but I can't really fix that until this has been fixed.

I can do this myself but may not be until next weekend.

RFCreate:
> There are users who have both the icon and the name (using both widgets) so we should cater for those users and allow the widget to display both.

You could have the widget twice, but I see why doing it on the same widget is more convenient.

PR #5304, opened 2025-06-27
nix: disable qtile-extras current_layout_icon test

Gurjaka:
This PR disables the `current_layout_icon` test from `qtile-extras` when running under Nix Flakes. This prevents the `nix flake check` CI from failing on [qtile/qtile#5302](https://github.com/qtile/qtile/pull/5302).

**Context**
The `current_layout_icon` test currently causes issues in the flake CI pipeline. Temporarily disabling it ensures that the flake check passes cleanly while work on [qtile/qtile#5302](https://github.com/qtile/qtile/pull/5302) progresses.

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -28,6 +28,13 @@ self: final: prev: {
           }
         )).override
           { wlroots = prev.wlroots_0_17; };
+
+      qtile-extras = pprev.qtile-extras.overridePythonAttrs (oldAttrs: {
+        # disable currentlayouticon test for https://github.com/qtile/qtile/pull/5302
+        disabledTestPaths = oldAttrs.disabledTestPaths ++ [
+          "test/widget/test_currentlayouticon.py"
+        ];
+      });
     })
   ];
   python3 =


Gurjaka:
@tych0 @jwijenbergh this should do it

tych0:
Great, thanks for the quick response!

jwijenbergh:
Hmmm but why is this needed? Why does our CI fail on a test failure of qtile-extras? Qtile-extras is not defined in our flake right?

Gurjaka:
> Hmmm but why is this needed? Why does our CI fail on a test failure of qtile-extras?

The CI fails because `nix flake check` runs passthru tests defined in `nixpkgs`, and the Qtile passthru test currently includes `qtile-extras`. That test was originally written with the assumption that both Qtile and Qtile-Extras would always be bumped together, so the test checks that they work together.

We discussed this on Discord a while ago and agreed to separate the dependency — Qtile shouldn't depend on Qtile-Extras in passthru tests. The current unified test setup was added by @Sigmanificient with good intent, but in hindsight, it shouldn't have been this way.

I am planning to open PR to properly separate these tests. Until that PR gets merged and hits `nixos-unstable`, this temporary fix disables the problematic `current_layout_icon` test so `nix flake check` CI doesn't fail for [qtile/qtile#5302](https://github.com/qtile/qtile/pull/5302)

jwijenbergh:
> > Hmmm but why is this needed? Why does our CI fail on a test failure of qtile-extras?
> 
> The CI fails because `nix flake check` runs passthru tests defined in `nixpkgs`, and the Qtile passthru test currently includes `qtile-extras`. That test was originally written with the assumption that both Qtile and Qtile-Extras would always be bumped together, so the test checks that they work together.
> 
> We discussed this on Discord a while ago and agreed to separate the dependency — Qtile shouldn't depend on Qtile-Extras in passthru tests. The current unified test setup was added by @Sigmanificient with good intent, but in hindsight, it shouldn't have been this way.
> 
> I am planning to open PR to properly separate these tests. Until that PR gets merged and hits `nixos-unstable`, this temporary fix disables the problematic `current_layout_icon` test so `nix flake check` CI doesn't fail for [qtile/qtile#5302](https://github.com/qtile/qtile/pull/5302)

Ah! Thanks for the detailed explanation. Makes sense, this will do for now

PR #5309, opened 2025-06-30
Add missing TextLayout setters

RFCreate:
1. Add missing `textlayout` setters properties for `base._TextBox` class.
2. Remove found `self.layout` attribute assignment on values that remain constant.


libqtile/widget/base.py
```diff
@@ -562,7 +562,17 @@ def font(self):
     def font(self, value):
         self._font = value
         if self.layout:
-            self.layout.font = value
+            self.layout.font_family = value
```
tych0: Woah... this bug has been around for a long time, I guess.

libqtile/widget/base.py
```diff
@@ -549,10 +549,10 @@ def foreground(self):
         return self._foreground
 
     @foreground.setter
-    def foreground(self, fg):
-        self._foreground = fg
+    def foreground(self, value):
+        self._foreground = value
         if self.layout:
-            self.layout.colour = fg
+            self.layout.colour = value
```
tych0: /me blinks, this is a no-op, right?

libqtile/widget/base.py
```diff
@@ -574,6 +584,16 @@ def fontshadow(self, value):
         if self.layout:
             self.layout.font_shadow = value
 
+    @property
+    def markup(self):
+        return self._markup
+
+    @markup.setter
+    def markup(self, value):
+        self._markup = value
```
tych0: Do we even use _fontsize or _markup here? Is there any reason to save them on this object?

tych0:
Second patch in the series looks good to me, a few questions on the first one.

libqtile/widget/base.py
RFCreate: When I first saw it, I thought the same. :eyes:

libqtile/widget/base.py
RFCreate: Not sure what you mean, I just rename `fg` to `value` for consistency.

tych0: Yeah, I just wanted to make sure.

libqtile/widget/base.py
RFCreate: It needs to be `_*` otherwise when getting the value, i.e. `def markup(self): return self.markup`, it will be calling itself recursively.

tych0: Yep, what I'm wondering is if we need the _ attribute at all, or if we can get rid of it entirely and just have the setter be a proxy.

RFCreate: I hear you, but I am wondering what the getter will be returning if we remove the _attr.

tych0: Can't it return self.layout.markup?

Does anything actually use these getters and setters? Maybe we can just fix that code?

RFCreate:
> Can't it return self.layout.markup?

Done that in newest commit. https://github.com/qtile/qtile/pull/5309/commits/abf2e37b18e2f7d48d39b7af9fd337933c16f01b

> Does anything actually use these getters and setters? Maybe we can just fix that code?

Even though it can, doing testing locally I found out that cpu and memory spiked and even qtile terminated in timeout. I do not know if there is some recursion or is it just slowing down by the TextLayout getters and setters.

libqtile/widget/base.py
RFCreate: I had to keep `self._text` because some tests expect the text before it is formatted.

RFCreate:
@tych0 do you know if there is anything wrong with this function?
https://github.com/qtile/qtile/blob/313a79423b0d47b5150a39410a0faa941b989453/libqtile/pangocffi.py#L169-L170

I made this changes and suddenly pytest performance normalized.

```diff
diff --git a/libqtile/widget/base.py b/libqtile/widget/base.py
index a8ca70b8..fb0e2479 100644
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -566,10 +566,11 @@ class _TextBox(_Widget):
 
     @property
     def fontsize(self):
-        return self.layout.font_size
+        logger.warning(self.layout.font_size) # edit
+        return self._font_size
 
     @fontsize.setter
     def fontsize(self, value):
+        self._font_size = value
         self.layout.font_size = value
 
     @property
```

EDIT: Found something strange using the logger, the fontsize starts normal with `12` but then jumped to a big value, in my case `12288`.

RFCreate:
Need a way to transform the `font_size` absolute_size to its "normal" size for the getter.

tych0:
> do you know if there is anything wrong with this function?

`pango.pango_font_description_get_size()` is an FFI call, so it's not hard for me to imagine that adding calls to it will make things substantially slower.

But can you be a bit more specific about the problem? Maybe a test case that reproduces the issue?

I'll try to take a closer look at the PR today, there's a lot more here now than there originally was.

libqtile/widget/base.py
```diff
@@ -507,7 +507,10 @@ class _TextBox(_Widget):
     ]  # type: list[tuple[str, Any, str]]
 
     def __init__(self, text=" ", width=bar.CALCULATED, **config):
-        self.layout = None
+        class EmptyLayout:
+            pass
+
+        self.layout = EmptyLayout()
```
tych0: This feels like a hack; why is `None` no longer okay?

RFCreate: It needs a way to temporarily save the values before the real layout is set.

RFCreate:
> But can you be a bit more specific about the problem?

As I see it, the fontsize grow to large due to **using `pango_units_from_double` and not converting back in the fontsize getter** which complicates rendering and takes up memory and cpu processing. I found a way to fix it with `pango_units_to_double`.

RFCreate:
> Yep, what I'm wondering is if we need the _ attribute at all, or if we can get rid of it entirely and just have the setter be a proxy.

<details>
<summary>Said if it could remove the _attrs and I did, but I can see plenty of more examples in the code base of the same _attrs.</summary>

https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/base/drawer.py#L92-L106
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/base/window.py#L80-L88
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/base/window.py#L103-L111
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/wayland/window.py#L79-L97
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/wayland/window.py#L179-L193
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/wayland/window.py#L825-L831
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/backend/x11/window.py#L561-L563
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/bar.py#L125-L128
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/command/graph.py#L93-L96
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/command/graph.py#L103-L106
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/layout/xmonad.py#L231-L237
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/lazy.py#L87-L95
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/widget/bluetooth.py#L130-L144
https://github.com/qtile/qtile/blob/99c8a54b026d9108be238626f5e88880d2eb077a/libqtile/widget/bluetooth.py#L253-L263
</details>

Those are some of the examples that I could find, there might be more who knows, and I am not saying that everyone can or should be replaced, I am just pointing it out.

RFCreate:
Close in favor of #5325.

PR #5314, opened 2025-07-05
[wayc] Remove listeners in qw_server_finalize

richcarni:
With this change, now passes test/core/test_exitcode.py


jwijenbergh:
amazing!

PR #5316, opened 2025-07-05
Add vertical bar support to LaunchBar

elParaguayo:
Adding support for vertical bars.

Also changes bizarre method for displaying text label. Widget previously used `base._TextBox` widget for each label. This has been changed to just use a `TextLayout` object.

Do not merge until I've updated tests.

### DIFF
--- a/libqtile/widget/launchbar.py
+++ b/libqtile/widget/launchbar.py
@@ -23,6 +23,7 @@
 # SOFTWARE.
 from __future__ import annotations
 
+import math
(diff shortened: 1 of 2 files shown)


elParaguayo:
Confirmed working here: https://www.reddit.com/r/qtile/comments/1lj1bc5/comment/n1ivy9o/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

PR #5318, opened 2025-07-07
Add support for orientation vertical to CurrentLayout

RFCreate:
One thing that was lost when merging CurrentLayout and CurrentLayoutIcon was having suport for both orientations. This fixes it by updating the draw_icon function to support vertical as well.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
     !!! breaking changes !!!
       - Remove CurrentLayoutIcon, functionality merged with CurrentLayout
     * features
+      - Add support for orientation vertical to the new CurrentLayout widget
     * bugfixes
 
 Qtile 0.32.0, released 2025-06-09:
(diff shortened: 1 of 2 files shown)


RFCreate:
Unfortunately there is a problem where it is not drawing more icons if I repeat the widget vertically. Example:

```
left=bar.Bar(
            widgets=[
                widget.CurrentLayout(icon_first=True),
                widget.CurrentLayout(icon_first=True),
```

RFCreate:
Messed with the PR :sweat_smile:, vertical orientation now really fixed.

RFCreate:
I am currently working in https://github.com/qtile/qtile/pull/5302#issuecomment-3038846705 which goes after this PR due to sharing changes. I got the drawing working, later I will work on tests and migrations.

elParaguayo:
Thanks.

Am I right in thinking that this just resizes/positions the icon correctly in the bar but it doesn't rotate it?

RFCreate:
That is right for this one, but in the other PR #5320 I does rotate.

PR #5319, opened 2025-07-07
[wayc] Handle XDG window names

richcarni:
- Set name when qtile window created
- Update name when it changes


jwijenbergh:
merci

PR #5320, opened 2025-07-07
CurrentLayout mode kwarg

RFCreate:
Addresses comment https://github.com/qtile/qtile/pull/5302#issuecomment-3038846705 by @elParaguayo.

As I said in my other PR https://github.com/qtile/qtile/pull/5318#issuecomment-3043457978, this should go after that one because it adds all the logic for the vertical bar as well as the `mode="both"` to draw the icon and the text at the same time. When #5318 is merged, I will come back and solve the merge conflicts because these two PRs share some lines.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
       - Remove CurrentLayoutIcon, functionality merged with CurrentLayout
     * features
       - Add support for orientation vertical to the new CurrentLayout widget
+      - Draw icon and text together with the same CurrentLayout widget
     * bugfixes
 
 Qtile 0.32.0, released 2025-06-09:
--- a/docs/manual/howto/widget.rst
+++ b/docs/manual/howto/widget.rst
@@ -353,6 +353,17 @@ background. Usually this is done by including the following line at the start of
 The background can be a single colour or a list of colours which will result in a linear gradient
 from top to bottom.
 
+Vertical Orientation
+--------------------
+
+If you plan to support vertical orientation in your widget, after calling
+``self.drawer.clear`` and ``self.drawer.ctx.save`` place this function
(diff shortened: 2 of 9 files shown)


elParaguayo:
#5318 has been merged so would be good to fix this one. Once that's done, I'll sort out qtile-extras.

RFCreate:
I thought you would hold any changes https://github.com/elParaguayo/qtile-extras/commit/9569410301371f5b214baee7aea3a2b55e17eb35 until this was merged.

elParaguayo:
I wanted to stop any immediate breakages. I'll update again one we've made final changes here.

qtile-extras tries to maintain compatibility with qtile at all times.

elParaguayo:
Can we add widget screenshots to show all the different config options i.e. icon only, text only, icon then text, text then icon?

RFCreate:
Ready! It got a nice render.

elParaguayo:
@RFCreate Are you happy with this one now?

RFCreate:
Yes, I think is ready, unless you see anything that needs improvement.

RFCreate:
I have a fix that I didn't notice, give me one hour.

RFCreate:
It should be ready now, didn't realize the image was adding its own padding and fix the icon and text when bar is vertical and rotate=False.

elParaguayo:
Does the migration also need to cover users who've installed the previous version and so are using `icon_first` to display an icon?

RFCreate:
Do you think? It hasn't been out in an official release. But if you see it as necessary I'll try to write it.

elParaguayo:
No. That's fine. Leave as is as long as the migration moves from last release to this version of the widget.

RFCreate:
Then we are covered, the migration is in https://github.com/qtile/qtile/pull/5320/commits/641584ca90a2172b94d013b8579af4c01416de67.

PR #5322, opened 2025-07-09
[wayc] Fix geometry for outputs during tests

richcarni:
Call a qw_server function to set test_env flag

Count as outputs are added and set custom modes:
- Output 1: 800 x 600
- Output 2: 640 x 480

Related test: test/test_manager.py > test_screen_dim (but requires working Window.info())

Also changed insertion order of wl_list server->outputs, so they are iterated in order added

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -158,6 +158,9 @@ class Core(base.Core):
         self.qw.cursor_motion_cb = lib.cursor_motion_cb
         self.qw.cursor_button_cb = lib.cursor_button_cb
         self.qw.on_screen_change_cb = lib.on_screen_change_cb
+        # During tests, we want to fix the geometry of the 1 or 2 outputs
+        if "PYTEST_CURRENT_TEST" in os.environ:
+            lib.qw_server_enable_test_env(self.qw)
         lib.qw_server_start(self.qw)
 
     def new_wid(self) -> int:
--- a/libqtile/backend/wayland/qw/output.c
+++ b/libqtile/backend/wayland/qw/output.c
@@ -52,9 +52,18 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
     wlr_output_state_init(&state);
     wlr_output_state_set_enabled(&state, true);
 
-    struct wlr_output_mode *mode = wlr_output_preferred_mode(wlr_output);
-    if (mode) {
-        wlr_output_state_set_mode(&state, mode);
+    if (server->test_env) {
+        if (server->output_count == 0) {
+            wlr_output_state_set_custom_mode(&state, 800, 600, 0);
+        }
+        else {
+            wlr_output_state_set_custom_mode(&state, 640, 480, 0);
+        }
+    } else {
+        struct wlr_output_mode *mode = wlr_output_preferred_mode(wlr_output);
+        if (mode) {
+            wlr_output_state_set_mode(&state, mode);
+        }
     }
 
     wlr_output_commit_state(wlr_output, &state);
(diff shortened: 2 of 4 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -158,6 +158,9 @@ def __init__(self) -> None:
         self.qw.cursor_motion_cb = lib.cursor_motion_cb
         self.qw.cursor_button_cb = lib.cursor_button_cb
         self.qw.on_screen_change_cb = lib.on_screen_change_cb
+        # During tests, we want to fix the geometry of the 1 or 2 outputs
+        if "PYTEST_CURRENT_TEST" in os.environ:
+            lib.qw_server_enable_test_env(self.qw)
```
jwijenbergh: I guess checking the env in c using `getenv` doesn't work?

libqtile/backend/wayland/qw/output.c
```diff
@@ -74,7 +83,11 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
     output->destroy.notify = qw_output_handle_destroy;
     wl_signal_add(&wlr_output->events.destroy, &output->destroy);
 
-    wl_list_insert(&server->outputs, &output->link);
+    // Insert output at end of list
+    wl_list_insert(server->outputs.prev, &output->link);
+    if (server->test_env) {
+        server->output_count++;
```
jwijenbergh: any reason we can't just check `wl_list_length(server->outputs)`?

libqtile/backend/wayland/qw/server.c
```diff
@@ -244,6 +244,10 @@ struct qw_view *qw_server_view_at(struct qw_server *server, double lx, double ly
     return tree->node.data;
 }
 
+void qw_server_enable_test_env(struct qw_server *server) {
+    server->test_env = true;
+}
+
```
jwijenbergh: can't we just modify the struct directly in the python code?

libqtile/backend/wayland/core.py
richcarni: Have to admit I didn't think of that. Easy

libqtile/backend/wayland/qw/output.c
richcarni: Haha, thought I'd be able to tidy this up before you saw it!

libqtile/backend/wayland/qw/server.c
richcarni: Not an issue since using getenv now

But yes, turns out we can

I get a bit confused about what python can and can't see on the c side. I'm sure I've had it complain when trying to access members of a struct before, something about being opaque IIRC

richcarni:
I've made changes in a new commit - thought that made sense here rather than force push (but happy to force push if you prefer)

libqtile/backend/wayland/core.py
```diff
@@ -158,6 +158,9 @@ def __init__(self) -> None:
         self.qw.cursor_motion_cb = lib.cursor_motion_cb
         self.qw.cursor_button_cb = lib.cursor_button_cb
         self.qw.on_screen_change_cb = lib.on_screen_change_cb
+        # During tests, we want to fix the geometry of the 1 or 2 outputs
```
jwijenbergh: Can you remove this instead?

richcarni: Sorry about that

jwijenbergh:
Thanks!

libqtile/backend/wayland/core.py
jwijenbergh: No worries at all! Thanks for the contributions

PR #5310, opened 2025-07-01
nix: remove qtile-extras overlay

Gurjaka:
This PR removes the `qtile-extras` overlay that was previously used to disable the `test_currentlayouticon` test for [qtile/qtile#5302](https://github.com/qtile/qtile/pull/5302).

**Reason:**
The overlay is no longer necessary once [Nixpkgs#421427](https://nixpk.gs/pr-tracker.html?pr=421427) is merged and reaches `nixos-unstable`.
> Note: After that point, we should avoid modifying `qtile-extras` in this repository altogether.

**Do not merge this PR until** [Nixpkgs#421427](https://nixpk.gs/pr-tracker.html?pr=421427) has landed in `nixos-unstable`.

PR #5325, opened 2025-07-10
Remove setter proxies to self.layout attributes

RFCreate:
Write this as an alternative to #5309. I see this as an improvement and cleaner approach.

It removes the temporary placeholder class: https://github.com/qtile/qtile/pull/5309#pullrequestreview-2979617118 and removes the setters to the `self.layout` attributes.

Instead when an attribute needs to be set, it is placed directly to `self.layout`.

PS. I also added some `fix: ...` commits that I found along the way.

### DIFF
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -404,8 +404,8 @@ class TextLayout:
             try:
                 attrlist, value, accel_char = pangocffi.parse_markup(value)
                 self.layout.set_attributes(attrlist)
-            except pangocffi.BadMarkup:
-                logger.warning("parse_markup() failed for {value}")
+            except pangocffi.BadMarkup as e:
+                logger.warning(e)
         self.layout.set_text(utils.scrub_to_utf8(value))
 
     @property
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -376,8 +376,7 @@ class TextLayout:
         layout.set_alignment(pangocffi.ALIGN_CENTER)
         if not wrap:  # pango wraps by default
             layout.set_ellipsize(pangocffi.ELLIPSIZE_END)
-        desc = pangocffi.FontDescription.from_string(font_family)
-        desc.set_absolute_size(pangocffi.units_from_double(float(font_size)))
+        desc = pangocffi.FontDescription.from_string(f"{font_family} {font_size}px")
         layout.set_font_description(desc)
         self.font_shadow = font_shadow
         self.layout = layout
@@ -446,12 +445,11 @@ class TextLayout:
     @property
     def font_size(self):
         d = self.fontdescription()
-        return d.get_size()
+        return pangocffi.units_to_double(int(d.get_size()))
 
     @font_size.setter
     def font_size(self, size):
         d = self.fontdescription()
-        d.set_size(size)
         d.set_absolute_size(pangocffi.units_from_double(size))
         self.layout.set_font_description(d)
 
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -419,8 +419,7 @@ class TextLayout:
         self._width = value
         self.layout.set_width(pangocffi.units_from_double(value))
 
-    @width.deleter
-    def width(self):
+    def reset_width(self):
         self._width = None
         self.layout.set_width(-1)
 
--- a/libqtile/pangocffi.py
+++ b/libqtile/pangocffi.py
@@ -71,6 +71,7 @@ def patch_cairo_context(cairo_t):
 ALIGN_CENTER = pango.PANGO_ALIGN_CENTER
 ELLIPSIZE_END = pango.PANGO_ELLIPSIZE_END
 units_from_double = pango.pango_units_from_double
+units_to_double = pango.pango_units_to_double
 
 
 ALIGNMENTS = {
--- a/libqtile/pangocffi.py
+++ b/libqtile/pangocffi.py
@@ -178,12 +178,14 @@ def parse_markup(value, accel_marker=0):
     attr_list = ffi.new("PangoAttrList**")
     text = ffi.new("char**")
     error = ffi.new("GError**")
-    value = value.encode()
+    markup_text = value.encode()
 
-    ret = pango.pango_parse_markup(value, -1, accel_marker, attr_list, text, ffi.NULL, error)
+    ret = pango.pango_parse_markup(
+        markup_text, -1, accel_marker, attr_list, text, ffi.NULL, error
+    )
 
     if ret == 0:
-        raise BadMarkup(f"parse_markup() failed for {value}")
+        raise BadMarkup(f"parse_markup() failed for: {value}")
 
     return attr_list[0], ffi.string(text[0]), chr(accel_marker)
 
--- a/libqtile/popup.py
+++ b/libqtile/popup.py
@@ -46,7 +46,7 @@ class Popup(configurable.Configurable):
         ("border", "#111111", "Border colour."),
         ("border_width", 0, "Line width of drawn borders."),
         ("font", "sans", "Font used in notifications."),
-        ("font_size", 14, "Size of font."),
+        ("fontsize", 14, "Size of font."),
         ("fontshadow", None, "Colour for text shadows, or None for no shadows."),
         ("horizontal_padding", 0, "Padding at sides of text."),
         ("vertical_padding", 0, "Padding at top and bottom of text."),
@@ -80,7 +80,7 @@ class Popup(configurable.Configurable):
             text="",
             colour=self.foreground,
             font_family=self.font,
-            font_size=self.font_size,
+            font_size=self.fontsize,
             font_shadow=self.fontshadow,
             wrap=self.wrap,
             markup=True,
@@ -114,24 +114,6 @@ class Popup(configurable.Configurable):
     def height(self, value: int) -> None:
         self.win.height = value
         self.drawer.height = value
-
-    @property
-    def text(self) -> str:
-        return self.layout.text
-
-    @text.setter
-    def text(self, value: str) -> None:
-        self.layout.text = value
-
-    @property
-    def foreground(self) -> ColorType:
-        return self._foreground
-
-    @foreground.setter
-    def foreground(self, value: ColorType) -> None:
-        self._foreground = value
-        if hasattr(self, "layout"):
-            self.layout.colour = value
 
     def set_border(self, color: ColorType) -> None:
         self.win.paint_borders(color, self.border_width)
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -545,36 +545,6 @@ class _TextBox(_Widget):
         return self.fmt.format(self._text)
 
     @property
-    def foreground(self):
-        return self._foreground
-
-    @foreground.setter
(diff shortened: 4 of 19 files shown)


libqtile/pangocffi.py
```diff
@@ -178,12 +179,14 @@ def parse_markup(value, accel_marker=0):
...
+    markup_text = value.encode()
 
-    ret = pango.pango_parse_markup(value, -1, accel_marker, attr_list, text, ffi.NULL, error)
+    ret = pango.pango_parse_markup(
+        markup_text, -1, accel_marker, attr_list, text, ffi.NULL, error
+    )
 
     if ret == 0:
-        raise BadMarkup(f"parse_markup() failed for {value}")
+        raise BadMarkup(f"parse_markup() failed for: {value}")
```
tych0, line 189: here you need to change the variable name in the f string as well.

RFCreate: You mean changing to `{markup_text}`?, I leave it as `{value}` because when it is raised  the message would contain the same value as the input.

```python
# before
>>> pangocffi.parse_markup("♫ Above & Beyond - Always - Tinlicker Extended Mix")
  File "<python-input-1>", line 1, in <module>
    pangocffi.parse_markup("♫ Above & Beyond - Always - Tinlicker Extended Mix")
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/libqtile/pangocffi.py", line 189, in parse_markup
    raise BadMarkup(f"parse_markup() failed for: {markup_text}")
libqtile.pangocffi.BadMarkup: parse_markup() failed for: b'\xe2\x99\xab Above & Beyond - Always - Tinlicker Extended Mix'


# after
>>> pangocffi.parse_markup("♫ Above & Beyond - Always - Tinlicker Extended Mix")
  File "<python-input-3>", line 1, in <module>
    pangocffi.parse_markup("♫ Above & Beyond - Always - Tinlicker Extended Mix")
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/libqtile/pangocffi.py", line 189, in parse_markup
    raise BadMarkup(f"parse_markup() failed for: {value}")
libqtile.pangocffi.BadMarkup: parse_markup() failed for: ♫ Above & Beyond - Always - Tinlicker Extended Mix
```

tych0: ah, makes sense, thanks!

libqtile/widget/base.py
```diff
@@ -603,22 +567,29 @@ def _configure(self, qtile, bar):
...
-            self.layout.width = self.bar.width - 2 * self.actual_padding
-
     def check_width(self):
         """
         Check whether the widget needs to have calculated or fixed width
         and whether the text should be scrolled.
         """
+        # Reset the layout width to let the layout calculate
+        # the width based on the length of the text.
+        del self.layout.width
```
tych0: doesn't the very next line use this attribute?

RFCreate: Yep but is not really deleting the attribute, it is just to make sure that the width is not fixed. See https://github.com/qtile/qtile/blob/master/libqtile/backend/base/drawer.py:

```python
    @property
    def width(self):
        if self._width is not None:
            return self._width
        else:

    @width.deleter
    def width(self):
        self._width = None
        self.layout.set_width(-1)
```

elParaguayo:
How much of qtile-extras is this going to break? I guess it might be worth me creating a branch in qtile-extras that pulls in this branch and see how CI goes!

RFCreate:
> How much of qtile-extras is this going to break? I guess it might be worth me creating a branch in qtile-extras that pulls in this branch and see how CI goes!

@elParaguayo  Nothing was broken by this change, check the [workflow](https://github.com/RFCreate/qtile-extras/actions/runs/16225959965). You can see the [changes I made here](https://github.com/elParaguayo/qtile-extras/compare/main...RFCreate:qtile-extras:main), I just change the url in the tox.ini file and remove the currentlayouticon test that as you know that widget was removed.

libqtile/widget/base.py
tych0: Can we make this a function call somehow? The behavior is a little too magical for me.

RFCreate: You got it! (EDIT: Breaks qtile-extras)

RFCreate:
Latest change does break qtile-extras due to this [line](https://github.com/elParaguayo/qtile-extras/blob/3a7fc359e175f74144d630e072a73c37b7e6668d/qtile_extras/widget/groupbox2.py#L444) in `groupbox2.py`.

Should I revert?

elParaguayo:
> Latest change does break qtile-extras due to this [line](https://github.com/elParaguayo/qtile-extras/blob/3a7fc359e175f74144d630e072a73c37b7e6668d/qtile_extras/widget/groupbox2.py#L444) in `groupbox2.py`.
> 
> Should I revert?

No. No need to revert. We just need to update a line in qtile-extras to use your new method.

That `del` usage that tych0 didn't like was done by me so I'm happy to change qtile-extras.

libqtile/popup.py
```diff
@@ -115,24 +115,6 @@ def height(self, value: int) -> None:
...
-    @property
-    def foreground(self) -> ColorType:
-        return self._foreground
-
-    @foreground.setter
-    def foreground(self, value: ColorType) -> None:
-        self._foreground = value
-        if hasattr(self, "layout"):
-            self.layout.colour = value
-
```
elParaguayo, line 135: What's the reason for deleting these? I'm a bit wary of doing so because there are things like this (https://github.com/qtile/qtile-examples/blob/master/graphical_notifications.py) that rely on those methods.

libqtile/backend/base/drawer.py
```diff
@@ -404,8 +403,8 @@ def text(self, value):
             try:
                 attrlist, value, accel_char = pangocffi.parse_markup(value)
                 self.layout.set_attributes(attrlist)
-            except pangocffi.BadMarkup:
-                logger.warning("parse_markup() failed for {value}")
+            except pangocffi.BadMarkup as e:
+                logger.warning(e)
```
elParaguayo, line 407: Is the exception description clear enough? It's important to see what text caused the parser to fail.

libqtile/widget/base.py
```diff
@@ -528,52 +528,16 @@ def text(self, value):
             value = value[: self.max_chars] + "…"
         self._text = value
         if self.layout:
-            # Reset the layout width
-            # Reason is because, if we've manually set the layout width,
-            # adding longer text will result in wrapping which increases the height of the layout.
-            del self.layout.width
```
elParaguayo, line 534: Do you need the call to reset the width here?

libqtile/widget/base.py
```diff
@@ -528,52 +528,16 @@ def text(self, value):
...
             self.layout.text = self.formatted_text
+
             if self.scroll:
                 self.check_width()
                 self.reset_scroll()
 
-            elif not self.bar.horizontal and not self.rotate:
-                self.layout.width = self.bar.width - 2 * self.actual_padding
-
```
elParaguayo, line 542: Can you explain why this isn't needed?

libqtile/widget/base.py
```diff
@@ -528,52 +528,16 @@ def text(self, value):
...
-    @property
-    def fontshadow(self):
-        return self._fontshadow
-
-    @fontshadow.setter
-    def fontshadow(self, value):
-        self._fontshadow = value
-        if self.layout:
-            self.layout.font_shadow = value
```
elParaguayo, line 575: Never really thought about this? If these are in `_TextBox.defaults` do they replace the properties when the widget tries to access the attribute?

libqtile/widget/battery.py
```diff
@@ -492,13 +492,10 @@ def __init__(self, **config) -> None:
         self._has_notified = False
         self.timeout = int(self.notification_timeout * 1000)
 
-    def _configure(self, qtile, bar):
```
elParaguayo: We tend to avoid checking config values in `__init__` because `widget_defaults` have not been applied at this point. So, if a user has set `background` in `widget_defaults` I don't think that would be picked up.

libqtile/popup.py
RFCreate: Just consistency, nothing more. This are the only changes needed in the file you mention: lines [289](https://github.com/qtile/qtile-examples/blob/9e54d54ef78339137932b1ac0a5d0813cd4fe1f0/graphical_notifications.py#L289) and [301](https://github.com/qtile/qtile-examples/blob/9e54d54ef78339137932b1ac0a5d0813cd4fe1f0/graphical_notifications.py#L301).

libqtile/backend/base/drawer.py
RFCreate: It uses the same text as before, the difference is it uses the original message.

https://github.com/qtile/qtile/blob/a9b01bd9d53880449978fdb2b3259bb0482b31a2/libqtile/pangocffi.py#L188

libqtile/widget/base.py
RFCreate: It is the only place it makes sense to me, nowhere else in the file we are setting `self.layout.width`.

libqtile/widget/base.py
RFCreate: Because it is already set in `_configure` and nowhere else it will be changed so its "final", unless when scroll and that is why the `reset_width()` is only in scroll.

libqtile/widget/base.py
RFCreate: The magic of `__getattr__` in the Configurable class, for example `self._foreground` does not exists when trying to access `self.foreground`, so it raised an AttributeError which is catched by the getter which as a result now searches for  `self.foreground` in the `_variable_defaults` and when it is found calls the setter and creates `self._foreground` with the default value.

libqtile/widget/battery.py
RFCreate: I see, let me fix it.

libqtile/popup.py
elParaguayo: OK. It would be good if you could submit a PR for the changes to fix that code in `qtile-examples` as I believe there are users who use it.

elParaguayo:
This looks fine to me.

We need #5310 to be merged (which is, itself, waiting for `nixpkgs` to merge a PR). Once that's done and you rebase this the nix flake test should pass.

After that, I can fix `GroupBox2` in `qtile-extras`.

libqtile/popup.py
RFCreate: I open a PR https://github.com/qtile/qtile-examples/pull/91, I could not found much to fix other than the file that you already pointed out.

RFCreate:
I think #5310 will take some time, fixing the qtile-extras directly will be much quicker.

elParaguayo:
Yes, but it's chicken and egg here. This is failing because of qtile-extras but I can't fix qtile-extras without this being merged.

Alternative is to have the niz flake here disable the groupbox2 test which would then allow this to pass CI.

RFCreate:
Of course on will have to fail for the other to pass, but (I might be wrong) a failing PR can still be merge, right?

PR #5327, opened 2025-07-11
focus before bring to front

QiangF:
To make scratchpad toggle work when current window is some wine application.
If we do not focus the window first, bring_to_front won't work for some windows application in wine.

### DIFF
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -97,8 +97,8 @@ class WindowVisibilityToggler:
             self.hide()
         elif self.shown and self.visible:
             win = self.window
+            self.window.group.focus(win, True)
             win.bring_to_front()
-            win.focus(warp=True)
             self.shown = True
         else:
             self.show()
@@ -120,6 +120,7 @@ class WindowVisibilityToggler:
             win._float_state = FloatStates.TOP
             # add to group and bring it to front.
             win.togroup()
+            win.focus(warp=True)
             win.bring_to_front()
             # toggle internal flag of visibility
             self.shown = True


tych0:
Patch looks good, any chance you can add a test to enforce this ordering?

QiangF:
The other place that need this ordering is in focus_by_click, I have fixed it. Probably fixing these two places is enough.

I am not sure if there are any cases that a window need to be brought to front without focusing it, if we will always focus the window to be brought to front, we can enforce this ordering by adding focus to bring_to_front function, but that might be overdone and might affect other users.

Also I don't have wayland, perhaps it is not needed there. In any case this pull request just changes some line ordering and will do no harm.

tych0:
I agree it won't hurt, but can you look into a test to enforce this behavior so we don't break it in the future?

QiangF:
I don't know how to make that test. Maybe put in the definition of bring_to_front like this:


    @expose_command()
    def bring_to_front(self):
        if self.get_wm_type() != "desktop":
            if self.qtile.current_window != self:
                self.focus(warp = False)
            self.window.configure(stackmode=xcffib.xproto.StackMode.Above)
            self.raise_children()
            self.qtile.core.update_client_lists()

tych0:
For a test, I'm imagining that you would monkey patch Window somehow to save the order of focus / place events, to ensure that the window gets focused before it gets placed?

QiangF:
Sorry I don't know how to do that.

tych0:
I appreciate the fixes, but I really think we need some tests to enforce this behavior. Otherwise, we're likely to break it with some future refactoring, since all this focus logic is pretty brittle.

QiangF:
> I appreciate the fixes, but I really think we need some tests to enforce this behavior. Otherwise, we're likely to break it with some future refactoring, since all this focus logic is pretty brittle.

Can we just fixed it in the scratchpad.py file, and put a comment there? It is clear enough and why would anybody change the order of three lines in any subsequent refactoring? If that happens I will let him/her know. :-)

tych0:
Unfortunately, that doesn't work very well in practice. People move on, get busy, etc. and we lose the reason for the ordering. An automated test is the best way to enforce it.

QiangF:
> Unfortunately, that doesn't work very well in practice. People move on, get busy, etc. and we lose the reason for the ordering. An automated test is the best way to enforce it.

Do you mean using pytest to enforcing it? Please give me an example and I am glad to do it. 

Also I have limited the modification to scratchpad.py only. It seems that this is not an ordering issue. group.focus(win) is missing, and without that the wine application doesn't behave well. It is not an ordering issue. Sorry I am stilling learning in the process.

libqtile/scratchpad.py
```diff
@@ -97,8 +97,8 @@ def toggle(self):
             self.hide()
         elif self.shown and self.visible:
             win = self.window
+            self.window.group.focus(win, False)
```
elParaguayo: Why is `warp` `False` here when the original code had it as `True`?

QiangF: I thought the cursor should stay away from the view most of the time. Fixed in the last commit.

QiangF:
Just a comment to make this issue alive.

tych0:
Yeah, I would still really love a test for this since this kind of thing is so easy to break.

QiangF:
@tych0 Can you give a direction on how to make this test? ie. check if a window is in focus before bring to front.

tych0:
There are lots of tests that manipulate focus and then do `assert_focused()`, I would start by reading some of those.

PR #5328, opened 2025-07-11
[WayC] nix: get rid off pywayland and xkbcommon

Gurjaka:
Get rid of unnecessary dependencies: `pywayland`, `xkbcommon`


jwijenbergh:
merci

PR #5335, opened 2025-07-12
fix(docs): new path to logo.png

RFCreate:
Fix `logo.png` path and add a check for when the error happen in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ clean: ## Clean generated files
 
 .PHONY: run-ffibuild
 run-ffibuild: ## Build FFI modules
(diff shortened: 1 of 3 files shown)


elParaguayo:
Thanks for this. Changes look fine but I'm interested in what's driving the need for this.

RFCreate:
It is kind of obscure, see it for yourself. :eyes:

Before: https://docs.qtile.org/en/latest/manual/commands/keybindings.html

After: https://qtile--5335.org.readthedocs.build/en/5335/manual/commands/keybindings.html

elParaguayo:
Works for me. Thanks!

PR #5336, opened 2025-07-13
Prevent interactive help in `qtile repl`

elParaguayo:
`help()` starts an interactive help session in a standard python repl session but it will block qtile's event loop.

This PR patches help to tell users that interactive help is unavailable.

Any other commands we should repeat this approach? Anything blocking is going to be bad so maybe a generic warning?

### DIFF
--- a/libqtile/interactive/repl.py
+++ b/libqtile/interactive/repl.py
@@ -34,6 +34,13 @@ COMPLETION_REQUEST = "___COMPLETE___::"
 REPL_PORT = 41414
 
 
+def mark_unavailable(func):
+    def _wrapper(*args, **kwargs):
+        print(f"'{func.__name__}' is disabled in this REPL.")
+
+    return _wrapper
+
+
 def make_safer_env():
     """
     Returns a dict to be passed to the REPL's global environment.
@@ -53,6 +60,9 @@ def make_safer_env():
 
     # Store original help so we can still call it safely
     builtins.help = safe_help
+
+    # Mask other builtins
+    builtins.input = mark_unavailable(builtins.input)
 
     return {"__builtins__": builtins}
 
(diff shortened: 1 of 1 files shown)


tych0:
but wait, if we run it in a separate thread, it shouldn't block the event loop should it?

or alternatively, if we're gonna ban help, do we need the separate thread?

elParaguayo:
So... I started off just blocking `help()` as it made my laptop unresponsive to any key presses. 

Other things that could block like `time.sleep` are solved by running a thread but `help()` would still be an issue in the thread because it's waiting to read stdin (hence why I also blocked `input`).

It might be possible to solve that but it's completely unnecessary in my view when `help(object)` works fine.

tych0:
I guess what I'm asking is: if we're going to block stuff we don't want to do because it blocks the whole event loop, do we need the threaded part? It seems like either or, not both are necessary.

elParaguayo:
The stdin stuff needs to be blocked but, without the threaded part, you can't do things like `time.sleep`, waiting on subprocess etc either. So, unless I'm missing something, we should.have all of these.

tych0:
Ok, maybe I'm being thick, but: if time.sleep() runs in a separate thread, why does it block qtile's main event loop? I guess because it's not `await time.sleep()`, so we don't yield back to the asyncio engine?

Anyway, patches themselves look fine to me.

elParaguayo:
It works because we're using asyncio's `run_in_executor` and we await the result so the repl won't return the result until the code has executed but the eventloop is free to do other stuff.

Without the executor, `time.sleep` in synchronous and so doesn't yield back to asyncio to run other tasks. Sure, people could use `await asyncio.sleep` but that's not really the point here.

PR #5338, opened 2025-07-14
fix(x11): StateHint typo error

RFCreate:
Change `StateHing` to `StateHint`.

Also removed hint values from `window.py` since their use has been superseded by the **HintsFlags** dictionary in `xcbq.py`.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -34,26 +34,6 @@ HeightValue = 0x0008
 AllValues = 0x000F
 XNegative = 0x0010
 YNegative = 0x0020
-InputHint = 1 << 0
-StateHint = 1 << 1
-IconPixmapHint = 1 << 2
-IconWindowHint = 1 << 3
-IconPositionHint = 1 << 4
-IconMaskHint = 1 << 5
-WindowGroupHint = 1 << 6
-MessageHint = 1 << 7
-UrgencyHint = 1 << 8
-AllHints = (
-    InputHint
-    | StateHint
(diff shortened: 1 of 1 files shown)


libqtile/backend/x11/window.py
```diff
@@ -189,7 +169,7 @@ def get_wm_hints(self):
             return {
                 "flags": flags,
                 "input": atoms_list[1] if "InputHint" in flags else None,
-                "initial_state": atoms_list[2] if "StateHing" in flags else None,
+                "initial_state": atoms_list[2] if "StateHint" in flags else None,
```
tych0, line 172: hah, whoops :)

tych0: Hmm, actually, I wonder if this fixes stuff like all of our override-redirect weirdness?

RFCreate: I really don't know what it does, I just know it should not be misspelled.

elParaguayo: How did you even come across this?

RFCreate: I have another PR #5024 that has almost like a year and I was revisiting to move the import UrgencyHint to `xcbq` and there is where is stumble accross this.

RFCreate: Sad stuff the pywlroots situation.

elParaguayo: Well our wayland backend is being rewritten in C so we'll no longer be dependent on pywlroots.

PR #5342, opened 2025-07-15
Widgets: new draw_at_default_position function

RFCreate:
As I was working with in the widgets, I realized that there were many use cases of using the offset property for drawing i.e. `self.drawer.draw(offsetx=self.offset, ...`. I found out to be problematic since it is only there to work for horizontal bars and when trying to implement vertical it always needs fix.

I remove the property to prevent further bad use cases like the above, and as I was fixing all draw() methods, I realized that I was repeating to much, so I decided to create this little helper property `draw_kwargs` which automatically passes the correct values to the drawer.

Will cause breakage and that is why I have PRs ready for qtile-extras and qtile-examples.

### DIFF
--- a/docs/manual/howto/widget.rst
+++ b/docs/manual/howto/widget.rst
@@ -202,13 +202,13 @@ drawings are set out below.
 
 It is important to note that the bar controls the placing of the widget by
 assigning the ``offsetx`` value (for horizontal positioning) and ``offsety``
-value (for vertical positioning). Widgets should use this at the end of the
-``draw`` method. Both ``offsetx`` and ``offsety`` are required as both values will
-be set if the bar is drawing a border.
+value (for vertical positioning). While the widget controls its ``width`` and
+``height``. These four values should be use at the end of the ``draw`` method.
+It is recommended to call this helper function to do it automatically:
 
 .. code:: python
 
-    self.drawer.draw(offsetx=self.offsetx, offsety=self.offsety, width=self.width)
+    self.draw_at_default_position()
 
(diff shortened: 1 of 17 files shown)


elParaguayo:
Is this really necessary? This should work for vertical and horizontal bars:
```python
self.drawer.draw(offsetx=self.offsetx, offsety=self.offsety, width=self.width, height=self.height)
```
`width` and `height` are properties which should be set correctly based on the bar's orientation.

The above code is from `base._TextBox` which renders correctly in both orientations.

If anything, if the same arguments are used every time, we would just have a method in `base._Widget` that takes no arguments and calls the above command, and call that method from the widgets.

RFCreate:
Necessary no but it is convenient to avoid misspelling or misplaced arguments, of course `base._TextBox` has it correct, the issue is that many widgets do not use the same arguments.

If a function will be better, what kind of name should it have? I can't think of anything that conveys the idea.

As you know `draw()` already exists. One that is very direct could be `drawer_draw()` maybe to in the nose but who knows.

tych0:
Maybe `draw_at_default_position()`?

elParaguayo:
We should also update the docs for this: https://docs.qtile.org/en/stable/manual/howto/widget.html#the-draw-method

RFCreate:
That means we all agree in the method name `draw_at_default_position()`?

elParaguayo:
Yes. Works for me.

tych0:
looks good to me, thanks for another negative diff!

PR #5344, opened 2025-07-18
fix(images): do not fail on input zero

RFCreate:
In image resize and scale, instead of raising an error, do nothing when the width or height is zero.

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -199,6 +199,8 @@ class Img:
     height = _PixelSize("height")
 
     def resize(self, width=None, height=None):
+        if width == 0 or height == 0:
+            return
         width0, height0 = self.default_size
         width_factor, height_factor = None, None
         if width is not None:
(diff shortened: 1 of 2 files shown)


RFCreate:
Last minute change :sweat_smile:

tych0:
Rather than adding individual validations for parameters, I would like to look for a way to include type information in the defaults tuple, so we can apply validators at config load time. This would help users detect a few other problems you've pointed out recently too.

The simplest thing (IMO), would be to add an optional 4th tuple entry, which, if present, should cast to the type or raise an error that we can raise to the user at config load time.

But I can imagine more complex things. It would maybe be nice to move away from the `Configurable` interface all together, and try and leverage python's increasingly native type checking. That would be a bigger refactor though.

RFCreate:
I close this because I write this for the widgets that got images that resize, now realize that a crash is better than a giant image that does not explain what went wrong

tych0:
Yep, agreed.

PR #5345, opened 2025-07-18
Widgets: expose bar's 'size' and 'length'

RFCreate:
I'm proud of this one, is huge can could have been broken into multiple PRs but that would slow things down, I'll try to break it up:

1. It started because it tried to create two properties that return the bar's height and width but that were adjusted to orientation, found out that they are already implemented in the bar as `size` and `length` so I used those in the widgets.
2. Move `margin_side` and `margin_top` to their base class so that can be inherited.
3. Move `padding_side` and `padding_top` to their base class so that can be inherited.
4. Move Mixins add_defaults invocation to `_Widgets` so that other classes can override the values in their defaults.


elParaguayo:
Amazing. Really appreciate all these PRs tidying up our codebase.

I'll give this a test but one immediate question/suggestion: I really hate the current `padding|margin_x|y` bits. I like what you've done but can we replace `x|y` entirely with `top` and `side`? That way users can keep the same config for widgets when they are in horizontal or vertical bars.

It would need a migration but it seems like this might simplify things. @tych0 any views?

libqtile/widget/groupbox.py
```diff
@@ -128,10 +128,7 @@ def drawbox(
         framed = self.layout.framed(border_width, framecolor, 0, pad_y, highlight_color)
         y = self.margin_y
         if self.center_aligned:
-            for t in base.MarginMixin.defaults:
-                if t[0] == "margin":
-                    y += (self.bar.height - framed.height) / 2 - t[1]
-                    break
+            y += (self.bar.height - framed.height) / 2 - self.margin_y
```
tych0: Can you say a bit more in the commit message about why this change is necessary?

libqtile/bar.py
```diff
@@ -55,13 +55,9 @@ class Gap:
...
-        # 'size' corresponds to the height of a horizontal gap, or the width
-        # of a vertical gap
-        self._size = size
-        self._initial_size = size
-        # '_length' corresponds to the width of a horizontal gap, or the height
-        # of a vertical gap
-        self._length: int = 0
+        self.length: int = 0  # width of a horizontal gap or the height of a vertical gap
+        self.size: int = size  # height of a horizontal gap or the width of a vertical gap
+        self.fullsize: int = size  # sum of 'size' and margins
```
tych0, line 60: Can we make this a function call since it's computed? Or is the idea is that it stays mostly static, so we should only compute it once?

libqtile/widget/tasklist.py
```diff
@@ -461,7 +449,7 @@ def drawbox(
         self.drawtext(text, textcolor, width)
 
         icon_padding = (self.icon_size + self.padding_side) if icon else 0
-        padding_x = [self.padding_side + icon_padding, self.padding_side]
+        pad_x = [self.padding_side + icon_padding, self.padding_side]
```
tych0, line 449: Is this hunk and the one below necessary?

tych0 (commented):
LGTM, a few small questions. Thanks for another negative diff!

libqtile/widget/tasklist.py
RFCreate: I change it to `pad_x` so that when searching for `padding_x`, they do not appear in the results. The name still gives the idea of padding.

libqtile/bar.py
RFCreate: They are computed in the `_configure` functions, after that they stay mostly the same. I don't see a function as necessary.

libqtile/widget/groupbox.py
RFCreate: I hope I convey the message. https://github.com/qtile/qtile/pull/5345/commits/4fcdd708c991a473593b3cfc4bee4d0096d6dbe4 : "After the box has been displaced by margin_y, to center the box properly the same value of margin_y should be subtracted from the 'y' axis."

elParaguayo:
Any thoughts on removing `_x|y`? it's not ideal if you want to set this in `widget_defaults` but want to have both horizontal and vertical bars.

RFCreate:
It would need some rewrite on the margin logic for each widget. For example, I took the image widget as it is the easiest one and to make margin_x be always "sides" and margin_y be always "tops", it needed the following:

```diff
--- a/libqtile/widget/image.py
+++ b/libqtile/widget/image.py
@@ -65,7 +65,7 @@ class Image(base._Widget, base.MarginMixin):
         img.theta = self.rotate
         if not self.scale:
             return
-        new_height = self.bar.size - (self.margin_top * 2)
+        new_height = self.bar.size - (self.margin_y * 2)
         img.resize(height=new_height)
 
     def draw(self):
@@ -74,7 +74,7 @@ class Image(base._Widget, base.MarginMixin):
 
         self.drawer.clear(self.background or self.bar.background)
         self.drawer.ctx.save()
-        self.drawer.ctx.translate(self.margin_x, self.margin_y)
+        self.drawer.ctx.translate(self.margin_side, self.margin_top)
         self.drawer.ctx.set_source(self.img.pattern)
         self.drawer.ctx.paint()
         self.drawer.ctx.restore()
@@ -84,7 +84,7 @@ class Image(base._Widget, base.MarginMixin):
         if self.img is None:
             return 0
         img_length = self.img.width if self.bar.horizontal else self.img.height
-        return img_length + (self.margin_side * 2)
+        return img_length + (self.margin_x * 2)
 
     @expose_command()
     def update(self, filename):
```

What it demostrates is that `_x|_y` are still needed and not in every case can be replaced with `_top|_side` if what is expected is for `_x|_y` behave the same in both orientations.

> I like what you've done but can we replace `x|y` entirely with `top` and `side`?

We can rename it, but we would still need the functionality of the `_top|_side` properties.

elParaguayo:
What I was hoping to do was have `_top` be equivalent to `_y` on horizontal bars and `_x` on vertical bars.

My concern is that there's inconsistency across our widgets. e.g. if the widget is just rotating the `drawer` then it's just the same as a horizontal bar. However, if a widget is introducing custom logic for vertical bars then the widget creator might have swapped `_x` and `_y`.

This isn't a hill I'm going to die one but if we're going to change this I want it to be as good as possible.

I do appreciate that `_top` and `_side` may be as confusing as `_x` and `_y` depending on the widget!

docs/manual/howto/widget.rst
```diff
@@ -269,8 +256,8 @@ most commonly used to draw icons but the same method applies to other images.
 
         d_images = images.Loader(self.imagefolder)(*names)  # images.Loader can take more than one folder as an argument
 
+        new_height = self.bar.size - 2
```
elParaguayo, line 259: Why is this `- 2` whereas the line you deleted was only `- 1`?

RFCreate: I see it as taking one pixel up one pixel down, instead of half and half.

libqtile/bar.py
tych0: I guess I'm worried about two sources of truth: if someone makes a change to the length or size somewhere, they do not also get the implied change of fullsize without remembering to change that as well. It seems like a performance optimization to not have a function, which may or may not be necessary.

libqtile/widget/groupbox.py
tych0: Yep, that's great, thank you.

libqtile/bar.py
RFCreate: `length` does not affect size or fullsize. `size` is final/constant. `fullsize` is the one that can still change. Are you more worried by changes to the values inside or outside the `Gap|Bar` class?

tych0: Either from inside or outside, e.g. by hooks.

But maybe it's not worth worrying about; I haven't read much of the widget code recently, if they really do not change their size (I guess it's just STRETCH bars that do?), seems like it's fine as-is.

RFCreate: I write this function to prevent any modifications:

```python
def __setattr__(self, name, value):
    if name in ("length", "size", "x", "y", "width", "height", "horizontal"):
        if self._configured:
            raise AttributeError(f"Cannot set attribute '{name}' after _configured.")
        # Set size only once
        if name == "size" and hasattr(self, "size"):
            raise AttributeError(f"Cannot reset attribute '{name}'.")
    self.__dict__[name] = value
```

Do you want me to add it? @tych0

RFCreate:
Rewrite the commit messages as in https://github.com/qtile/qtile/pull/5355#issuecomment-3113966506 looks better now.

libqtile/bar.py
tych0: No, that's ok; if anything, i'd prefer we make the computed value (`fullsize` in this case) a function.

tych0:
Thanks!

PR #5349, opened 2025-07-21
drop mailing list

tych0:
...and clean up other bits by deleting some redundant docs and re-ordering things there to be the same as in the README.

The mailing list has been essentially unused (one or two "real" messages in the last year), and it is completely full of spam (sad that even Google can't figure out we don't want to buy DMT :). Given it's light usage and the maintenance burden it has caused (for a while, I was banned from qtile-dev!), I think it's best we put it out to pasture, in favor of IRC/discord and github discussions. If we land this, I will do some work to figure out how to properly "archive" the google group.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,6 @@ Qtile is supported by a dedicated group of users. If you need any help, please
 don't hesitate to fire off an email to our mailing list or join us on IRC. You
 can also ask questions on the discussions board.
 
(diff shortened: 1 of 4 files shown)


elParaguayo:
No objections here.

jwijenbergh:
totally agree with this!

tych0:
Ok, I've disabled posting by anyone who is not a group manager, but left the archive up for search results. The end of an era!

Issue #2620, opened 2021-07-13
Battery widget `notify_below` cannot send notification because Eventloop has not started


mrossinek:
# Issue description

I am trying to configure a Battery widget in combination with the `notify_below` setting. I think my config is fine, but whenever I expect a notification I instead get this message in `~/.local/share/qtile/qtile.log`:
```
WARNING libqtile utils.py:send_notification():L249 Eventloop has not started. Cannot send notification.
```

# Qtile version

Fresh build from `qtile-git` AUR package:
```sh
% qtile --version
0.18.1.dev39+g15e9f4cf
```

# Stack traces

<details>
<summary>
Here is the beginning of the logs from a fresh `qtile start -l DEBUG`. It gets rather spammy after a while but above message is contained twice (I have two batteries).
</summary>

```
2021-07-13 18:46:41,220 DEBUG libqtile log_utils.py:init_log():L126 Starting logging for Qtile
2021-07-13 18:46:41,224 INFO libqtile xcursors.py:_setup_xcursor_binding():L112 xcb-cursor not found, fallback to font pointer
2021-07-13 18:46:41,245 DEBUG libqtile utils.py:guess_terminal():L315 Guessing terminal: roxterm
2021-07-13 18:46:41,245 DEBUG libqtile utils.py:guess_terminal():L315 Guessing terminal: sakura
2021-07-13 18:46:41,246 DEBUG libqtile utils.py:guess_terminal():L315 Guessing terminal: hyper
2021-07-13 18:46:41,246 DEBUG libqtile utils.py:guess_terminal():L315 Guessing terminal: alacritty
2021-07-13 18:46:41,246 INFO libqtile utils.py:guess_terminal():L319 Terminal found: alacritty
2021-07-13 18:46:41,257 DEBUG libqtile clock.py:__init__():L70 Defaulting to the system local timezone.
[...]

PR #5353, opened 2025-07-22
drop DefaultEventLoopPolicy

### LINKED ISSUE #5346: Pytest fails under Python 3.14
### Issue description

Tests are failing on Python 3.14 while trying to package Qtile 0.32.0 on Fedora Rawhide. It's due to a deprecation warning with asyncio.DefaultEventLoopPolicy. Simply ignoring the warning results in tests erroring out. 

See:
* https://docs.python.org/3/library/asyncio-policy.html#asyncio.DefaultEventLoopPolicy.
* https://docs.python.org/dev/whatsnew/3.14.html#id3

```
ImportError while loading conftest '/builddir/build/BUILD/qtile-0.32.0-build/qtile-0.32.0/test/conftest.py'.
test/conftest.py:27: in <module>
    from test.helpers import BareConfig, TestManager
test/helpers.py:24: in <module>
libqtile/core/manager.py:50: in <module>
    from libqtile.core.loop import LoopContext, QtileEventLoopPolicy
libqtile/core/loop.py:67: in <module>
    class QtileEventLoopPolicy(asyncio.DefaultEventLoopPolicy):
/usr/lib64/python3.14/asyncio/__init__.py:61: in __getattr__
    warnings._deprecated(f"asyncio.{name}", remove=(3, 16))
/usr/lib64/python3.14/_py_warnings.py:830: in _deprecated
    _wm.warn(msg, DeprecationWarning, stacklevel=3)
E   DeprecationWarning: 'asyncio.DefaultEventLoopPolicy' is deprecated and slated for removal in Python 3.16
```

### Version

0.32.0

### Backend

Wayland (experimental)

### Config

```python
N/A
```

### Logs

```text
N/A
```

### Required

tych0:
python3.14 depreciates event loop policies entirely, in favor of just manually specifying the event loop as an argument to asyncio.run().

we were really using the event loop policy as a hack for not having a global variable, so let's switch to a global variable instead, and use it in the one place we need it.

we have to drop two verisons of python because they do not have the loop_factory= argument to asyncio:

    raise AssertionError(f"Error launching qtile, traceback:\n{error}")
    AssertionError: Error launching qtile, traceback:
      File "/home/runner/work/qtile/qtile/test/helpers.py", line 207, in run_qtile
        ).loop()
      File "/home/runner/work/qtile/qtile/libqtile/core/manager.py", line 216, in loop
        asyncio.run(self.async_loop(), loop_factory=lambda: libqtile.event_loop)
    TypeError: run() got an unexpected keyword argument 'loop_factory'

but maybe that's ok: 3.10 is outside of our last-three-versions policy, and 3.11 will be outside it in a few months when 3.14 is released.

There is another option here: since the motivation for this was really #2620, we could "just" switch the battery to running in-loop, which would mean that it is not run via run_in_executor() i.e. in a separate thread. That means that get_running_loop() would return the right loop, and we wouldn't need this hack at all.

(It also means that the battery could potentially block the UI while reading from e.g. procfs/sysfs files, but that seems unlikely. Perhaps it also suggests another widget type: something with an `async def poll()`, which could run in the main loop and avoid both of these problems.)

The only other widget which uses send_notification() is pomodoro, which doesn't read any files at all and just uses gettimeofday(), which is a VDSO i.e. memory access, that should also not block, so if we convert the battery widget, we should do this one as well.

Finally, we suggest use of send_notification() in the hooks documentation, but all hooks are fired from the main thread, so the get_running_loop() call should work fine there.

That does mean we'd have to be careful to always send_notification() from the main thread going forward, but it doesn't seem like that hard of a challenge.

Fixes #5346

### DIFF
--- a/libqtile/__init__.py
+++ b/libqtile/__init__.py
@@ -18,6 +18,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import asyncio
 
 
 class _UndefinedCore:
@@ -29,6 +30,7 @@ class _UndefinedQtile:
 
 
 qtile = _UndefinedQtile()
+event_loop: asyncio.AbstractEventLoop
 
 
 def init(q):
--- a/libqtile/core/loop.py
+++ b/libqtile/core/loop.py
@@ -10,8 +10,6 @@ from libqtile.log_utils import logger
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-    from libqtile.core.manager import Qtile
-
 
 class LoopContext(contextlib.AbstractAsyncContextManager):
     def __init__(
@@ -62,20 +60,3 @@ class LoopContext(contextlib.AbstractAsyncContextManager):
                 logger.exception("Exception in event loop:", exc_info=exc)  # noqa: G202
         else:
             logger.error("unhandled error in event loop: %s", context["msg"])
-
-
-class QtileEventLoopPolicy(asyncio.DefaultEventLoopPolicy):
-    """
-    Asyncio policy to ensure the main event loop is accessible
-    even if `get_event_loop()` is called from a different thread.
-    """
-
-    def __init__(self, qtile: Qtile) -> None:
-        asyncio.DefaultEventLoopPolicy.__init__(self)
-        self.qtile = qtile
-
-    def get_event_loop(self) -> asyncio.AbstractEventLoop:
-        if isinstance(self.qtile._eventloop, asyncio.AbstractEventLoop):
-            return self.qtile._eventloop
-
-        raise RuntimeError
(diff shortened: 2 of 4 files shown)


tych0:
I guess we have to change branch protections if we actually land this, I'm happy to do so pending what others think.

tych0:
Oh, the other thought is: we only break the last-three-versions policy if we release this code before python 3.14 is released. I have been trying to do one release per quarter, so maybe we can cut a release now, and then we'll keep with that cadence by doing one now and one in late october/november.

elParaguayo:
Thank you for doing this. I'll take a proper look when I'm home later today.

elParaguayo:
Are any major distros on python 3.11? If not, then I'm fine with this approach.

tych0:
Ubuntu 22.04 is still on 3.10 for system python, but I'm inclined to ignore it because there is 24.04 that is on 3.12, and 3.10 is so close to outside our window of last three anyway. 

Fedora is okay, as is Arch. I expect that the story is similar to 22.04 for RHEL and Debian stable: it is 3.10 or close to it, which we are about to EOL anyway.

elParaguayo:
OK. So is your plan to do a release now without this, then merge this and do next release after python 3.14 is released? If so, that works for me.

tych0:
> So is your plan to do a release now without this, then merge this and do next release after python 3.14 is released? If so, that works for me.

Yep, that is what I was thinking. I'll try to do a release in the next few days, and then do all the branch conditions to merge this.

PR #5355, opened 2025-07-22
Fix: finalize bar when it raises an exception in configure

RFCreate:
The problem is that when the screen is configured and the bar raises an exception, the screen will not be added to `self.screens` which means that the next time in `_process_screens()`, the bar will not finalize which will throw errors when trying to get `length` in CALCULATED _TextBox widgets since it depends on the call to the layout which already finalize.


libqtile/backend/base/drawer.py
```diff
@@ -388,7 +388,7 @@ def finalize(self):
         self.layout.finalize()
 
     def finalized(self):
-        self.layout.finalized()
+        return self.layout.finalized()
```
tych0: I don't see where this return value is used?

libqtile/config.py
```diff
@@ -514,8 +514,14 @@ def _configure(
...
 
-        for i in self.gaps:
-            i._configure(qtile, self, reconfigure=reconfigure_gaps)
+        for gap in self.gaps:
+            try:
+                gap._configure(qtile, self, reconfigure=reconfigure_gaps)
+            except Exception:
+                logger.exception(f"Error configuring {gap.position} gap/bar.")
+                # Call finalize to prevent future execution of the _actual_draw method
+                self.finalize_gap(gap.position)
```
tych0, line 523: Urgh, this is annoying. Ideally a gap would free all resources it has allocated if it is going to fail. I don't see a great way to factor the code to do that, though...

libqtile/backend/base/drawer.py
RFCreate: Right here:
https://github.com/qtile/qtile/blob/cbb896ed0b797caeab4fa65cc7f34ddbc74b1e8d/libqtile/widget/base.py#L639-L643

Before it basically did nothing because it always evaluated to `True`, we can remove it and in the _Widget finalize() method add `self.layout = None`. Might be less confusing this way.

libqtile/config.py
RFCreate: I don't see how it would know it is going to fail before running it?

tych0: Yeah, classic problem in languages with exceptions. It's just annoying.

libqtile/backend/base/drawer.py
tych0: Yeah, that sounds a bit nicer, do you mind changing it to that?

RFCreate: I change it as we said to only check layout None, also remove the check to `self.offsetx` since it is always int never None.

tych0:
Ok, I squashed the add-return and delete-finalized commits into one only delete one, and fixed up some commit messages (mostly to abide by https://cbea.ms/git-commit/#seven-rules), but it looks good to me, thanks for the cleanups!

PR #5357, opened 2025-07-23
[wayc] Add internal window info

richcarni:


jwijenbergh:
Thanks again and again

PR #5359, opened 2025-07-23
[wayc] Implement Static windows

richcarni:
Create a new window class: Static, and methods to convert a Window ->
Static


jwijenbergh:
wow, nice!

Issue #5305, opened 2025-06-28
Different screen order in v0.32


adolfocorreia:
After updating qtile from v0.31 to v0.32, the order of my two screens changed.  The index numbers associated with each monitor (0 and 1) are now reversed.

The culprit seems to be commit #50b8cf9, in which xrandr is priorized over xinerama.  Note that, in the old version of the `query_crtcs` method, the loop was over CRTC objects and now it loops over Output objects.

At least on my system, such objects are ordered differently as shown be the following script results:

Script:
```
from libqtile.backend.x11 import xcbq
import xcffib
conn = xcbq.Connection(":0")
wid = conn.screens[0].root.wid
ext = conn.conn(xcffib.randr.key)

print("CRTCs:")
for crtc in ext.GetScreenResources(wid).reply().crtcs:
    crtc_info = ext.GetCrtcInfo(crtc, xcffib.CurrentTime).reply()
    print(f"{crtc_info.x} {crtc_info.y} {crtc_info.width} {crtc_info.height}")

print("Outputs:")
for output in ext.GetScreenResources(wid).reply().outputs:
    output_info = ext.GetOutputInfo(output, xcffib.CurrentTime).reply()
    if not output_info.crtc:
        continue
    crtc_info = ext.GetCrtcInfo(output_info.crtc, xcffib.CurrentTime).reply()
    print(f"{crtc_info.x} {crtc_info.y} {crtc_info.width} {crtc_info.height}")
```

Result:
```
CRTCs:
0 0 2560 1440
2560 0 2560 1440
0 0 0 0

Outputs:
2560 0 2560 1440
0 0 2560 1440
```

Note also that xrandr orders my monitors the same way as that of the CRTCs objects:

```
> xrandr --listmonitors
Monitors: 2
 0: +*HDMI-2 2560/597x1440/336+0+0  HDMI-2
 1: +DP-1 2560/597x1440/336+2560+0  DP-1
```

I don't know what the expected behaviour is supposed to be here, but I suspect the `query_crtcs` method should respect the monitors order as listed by the `xrandr` command.

tych0:
Hmm, I would have expected them to be in the same order. Can you print the crtc name as well in your first output?

But in general, X11 doesn't guarantee stable ordering across reboots of these devices, so the real fix for this is #5274, any testing on that would be much appreciated.

tych0:
Also,
[...]

PR #5376, opened 2025-07-24
x11: prepend primary output when computing screens

tych0:
It is still not clear to me that this is quite the correct thing to do; xrandr -q does not sort the outputs as near as I can tell. Here is where it renders things:

https://gitlab.freedesktop.org/xorg/app/xrandr/-/blob/6f714830da6c8d74f024be6b0bb32c1ea39c1217/xrandr.c#L3629-3643

which it does in the order of `all_outputs`; `all_outputs` is populated here:

https://gitlab.freedesktop.org/xorg/app/xrandr/-/blob/6f714830da6c8d74f024be6b0bb32c1ea39c1217/xrandr.c#L1773-1810

Which seems to do it in the order of the outputs from ScreenResources, the same as we were doing before. And yet, it seems to fix the ordering issue for at least one person. Short of any better ideas as to how to sort things, let's land this for now as a band-aid for #5305. Perhaps it is the "complete fix", but it would be nice to understand why the ordering is different5305. Perhaps it is the "complete fix", but it would be nice to understand why the ordering is different for our queries vs. the ones that xrandr does...

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -405,6 +405,7 @@ class RandR:
 
     def query_crtcs(self, root):
         infos = []
+        primary = self.ext.GetOutputPrimary(root).reply().output
         for output in self.ext.GetScreenResources(root).reply().outputs:
             info = self.ext.GetOutputInfo(output, xcffib.CurrentTime).reply()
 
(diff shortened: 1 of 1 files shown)


elParaguayo:
Looks ok to me.

tych0:
Whoops, will fix the typo in the commit message :)

tych0:
> Looks ok to me.

Ok, thanks. I really do not understand why we need this, but sometimes that's how it goes...

PR #5377, opened 2025-07-24
[Wayc] save wlr_session and add change_vt support

elParaguayo:
Adds support for `change_vt` which requires retrieving the `wlr_session` object from the backend.

I'm not a C coder so you can be as rude as you like about this!


jwijenbergh:
This is perfectly good c code, you're even checking for null pointers, nice!

Gurjaka:
And he was saying he's new to C and wayland...

PR #5381, opened 2025-07-25
[wayc] Implement basic pointer focus

richcarni:


libqtile/backend/wayland/qw/cursor.c
```diff
@@ -143,6 +150,15 @@ static void qw_cursor_handle_frame(struct wl_listener *listener, void *data) {
     wlr_seat_pointer_notify_frame(cursor->server->seat);
 }
 
+static struct qw_position qw_cursor_get_pos(void *self) {
```
richcarni: Not sure about this. If think if we add wlr_cursor.h to cdefs we can access cursor members x,y directly and do away with this

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -30,6 +27,11 @@ static void qw_cursor_process_motion(struct qw_cursor *cursor, uint32_t time) {
     // Find the view under the cursor and get surface-local coordinates
     struct qw_view *view =
         qw_server_view_at(cursor->server, cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+    cursor->view = view;
```
richcarni, line 30: Thought about returning view in the callback. That's fine but there are other functions that need it and it was getting messy. Not sure if there is any performance hit when accessing c variables or functions from python? (ChatGPT gave me a very non-convincing answer!)

libqtile/backend/wayland/core.py
```diff
@@ -222,6 +228,66 @@ def focus_window(self, win: WindowType) -> None:
...
+        return view
+
+    def get_cursor_pos(self):
+        return self.qw_cursor.get_pos(self.qw_cursor)
+
+    def _focus_pointer(self, cx: int, cy: int) -> None:
+        assert self.qtile is not None
+        view = self.qw_cursor.view
+
+        if view != ffi.NULL:
```
jwijenbergh: early return on `view == ffi.NULL` so that the whole block is not indented

libqtile/backend/wayland/qw/cursor.c
jwijenbergh: I don't think there's any performance hit. Premature optimization is not needed anyways, if something ends up being too slow we can fix it later

libqtile/backend/wayland/qw/cursor.c
jwijenbergh: this is what i'm doing for `wlr_box.h` as well, but I had to duplicate definitions, lol!

libqtile/backend/wayland/qw/cursor.c
richcarni: Sounds fair

libqtile/backend/wayland/core.py
richcarni: Makes sense

libqtile/backend/wayland/qw/cursor.c
richcarni: I think you understood even though I wasn't super clear

I'm using a function pointer method to get the cursor position from the `wlr_cursor` struct in `qw_cursor` because `wlr_cursor` is opaque

Is this preferred because it would be considered bad practice to expose all of `wlr_cursor` on the python side?

richcarni: Ohhhhhhh, I understand what you mean about wlr_box now!

I'll follow what you've done

jwijenbergh:
Great work yet again!

Issue #5385, opened 2025-07-26
[Docs] version issues


elParaguayo:
@tych0 something weird is going on with our docs. We seem to have a v0.34.0 release.
<img width="1177" height="206" alt="Image" src="https://github.com/user-attachments/assets/75884fce-bd5c-4a6d-adb9-7b2755dba27e" />
(v0.33.0 is at the top of the list as I had to rebuild it as it failed (timeout) when we pushed the release).

Looking at the "stable" docs, there seem to be 3 commits from after the release but these aren't in the main repo:
<img width="1142" height="282" alt="Image" src="https://github.com/user-attachments/assets/20b3aa20-afcf-4ab0-963e-d17b5a091dcb" />

"v0.34.0" is linked here:
<img width="1138" height="80" alt="Image" src="https://github.com/user-attachments/assets/ec62c280-9760-41b9-bf34-6a2872b1d6f0" />
This is the commit: https://github.com/qtile/qtile/commit/7adfef813e627381f9849ff094e1ef959d70a605

Any idea what's happened?

I'd assume I can just delete the versions to these odd commits but let me know first.

tych0:
Oh, yes, you can just delete those. I was testing, sorry!

elParaguayo:
Thanks. Will delete shortly.

elParaguayo:
Looks like you can't delete builds so I've just hidden v0.34.0 for now and will need to remember to unhide it when we do the next release.

The `latest` version should be ok as we've now merged other commits.

tych0:
Thanks, sorry for the confusion. I should have used some other more bogus version number...

PR #5387, opened 2025-07-26
Modernize pythons

tych0:
Here's some patches that modernize our python versions, including in the release.yml. I also ported our release process to trusted publishers to get rid of some warnings there, and hopefully avoid some supply chain attacks as well.


elParaguayo:
Are these the cause of #5385 ?

tych0:
Yep, exactly. My bad, I forgot about the docs. feel free to delete!

PR #5390, opened 2025-07-27
[wayc]: add pywlroots copyright in core.py

Gurjaka:
Diff also has some formatting changes, but that's on ruff (python formatter that Qtile uses). But main Idea is to add pywlroots copyright at the top of the file, and remove TODOs.


jwijenbergh:
Thanks!

PR #5394, opened 2025-07-28
[wayc] Add window is_visible()

richcarni:


libqtile/backend/wayland/window.py
```diff
@@ -359,6 +359,10 @@ def get_wm_class(self) -> list | None:
             return [self._wm_class]
         return None
 
+    @expose_command()
```
elParaguayo, line 362: Any reason this needs to be exposed to the public API?

richcarni: Not that I'm aware of
It just is on both existing wayland and x11 backends

elParaguayo: Yes. I noticed that last night. We can leave for now and tidy it up across both backends later.

jwijenbergh:
Thanks!

PR #5398, opened 2025-07-30
move to uv

tych0:
This moves all of the non-docs infra to uv, deleting requirements.txt, requirements-dev.txt, setup.cfg, tox.ini, and dev.sh, each of which had a slightly different set of dependencies in them, and consolidates to pyproject.toml to manage all of this.

It also consolidates on using the makefile as the entry point for everything, and has CI call `make check check-packaging`. There are some new variables for supplying only a single backend or python version.

As a consequence of using `uv sync --all-extras` for CI, we now need to install the wayland deps in all environments even if we are only going to run the x11 backend tests, but that seems simpler than fragmenting these things again.

Finally, I organized the deps into four sections: dev, optional_core, widgets, and wayland. I wonder if optional_core is worth it, or if we should just start requiring all of those? But the other three seem reasonable to me.

It looks like the uv lock file is expected to be committed based on https://github.com/astral-sh/uv/issues/9797 , but I generally find these kinds of lock files annoying, and we haven't needed one in the past. Given that we should encode any real constraints/bugs we know about (e.g. API incompatibilities, bugs, etc.) in pyproject.toml directly, I don't see a ton of value in uv.lock. I think it will also help downstreams to not feel like they have to package *exactly* the versions we happened to test with, but can just respect our pyproject.toml.


ervinpopescu:
```
ERROR: setuptools==0.1.dev1+g62ae50e is used in combination with setuptools-scm>=8.x

Your build configuration is incomplete and previously worked by accident!
setuptools-scm requires setuptools>=61

Suggested workaround if applicable:
 - migrating from the deprecated setup_requires mechanism to pep517/518
   and using a pyproject.toml to declare build dependencies
   which are reliably pre-installed before running the build tools

  warnings.warn(
/tmp/build-env-hzmqk659/lib/python3.12/site-packages/setuptools_scm/git.py:167: UserWarning: "/home/runner/work/qtile/qtile" is shallow and may cause errors
  warnings.warn(f'"{wd.path}" is shallow and may cause errors')
```

Is this ok? I've seen it before in my CI when I was trying different dep versions. But for me it was failing, I don't get why here it doesn't fail.

tych0:
Heh, all the coverage seems to be busted, which is unfortunate...

> Is this ok?

The shallow warning is probably because you've done a shallow clone. But r.e. setuptools, does:

```
-    "setuptools>=60",
+    "setuptools>=61",
```

Make it go away? We should probably do that anyway.

elParaguayo:
Certainly much neater.

What's going on with coveralls though?

tych0:
I think I didn't do this:

```
-    py3{10,11,12,13}-x11: coverage run -m pytest --backend=x11 {posargs}
-    py3{10,11,12,13}-wayland: coverage run -m pytest --backend=wayland {posargs}
-
-    # Coverage is only run via GithubActions
-    # Coverage runs tests in parallel so we need to combine results into a single file
-    !pypy3-{wayland,x11}: coverage combine -q
-    # Include a text summary in the build log
-    !pypy3-{wayland,x11}: coverage report -m
-    # Create an xml summary to be submitted to coveralls.io
-    !pypy3-{wayland,x11}: coverage xml
```

I can work on fixing that, but... I personally think coverage is kind of a bogus metric. Am I the only one who thinks that way? If so, I'll fix it. If not, maybe we can just let it die.

pyproject.toml
```diff
@@ -37,6 +44,44 @@ classifiers = [
...
+[project.optional-dependencies]
+dev = [
+    "coverage",
+    "pytest-cov",
+    "pre-commit",
+    "PyGObject",
+    "isort",
+    "pytest >= 6.2.1",
+    "mypy",
+    "pytest-anyio",
```
dvzrv: This should be anyio

elParaguayo:
I like it as it can encourage users to write tests (although we should reduce the threshold that triggers a failure). However, I won't lose much sleep if we decide to ditch it.

pyproject.toml
```diff
@@ -37,6 +44,44 @@ classifiers = [
...
+    "keyring",
+    "mailbox",
+    "psutil",
+    "pulsectl",
+    "pulsectl_asyncio",
+    "python-mpd2",
+    "pytz",
+    "xdg",
+    "xmltodict",
+]
```
dvzrv, line 77: Would it be possible to be more precise here and specify which exact widget requires what dependency?

See e.g. the Arch Linux package for a list of (hopefully still correct) optional dependencies: https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/blob/b5acf08e06b2b27b0838efccc5b806477bda9532/PKGBUILD#L53-76

dvzrv:
Thanks! This generally looks great and simplifies things quite a bit!

FWIW, for downstream distribution purposes the uv environment will not be used (usually), so the lock file is not that relevant to e.g. Arch Linux per se (but may be relevant for developers and users installing from source).

pyproject.toml
tych0: Whoops, yep, thanks. I'll add a separate patch for that.

pyproject.toml
tych0: It is, I wonder if that makes it annoying to install from source though? Then you have to make a list of widgets you care about: `uv tool install qtile[imap,sensors,optional_core]`.

Maybe we should include pure-python dependencies in the default distribution, but split out the ones that link against separate C libraries, does that make it less cumbersome?

tych0:
> for downstream distribution purposes the uv environment will not be used (usually), so the lock file is not that relevant

ok, thanks, that's what I figured. Seems okay to leave it out for now, and just ensure that pyproject.toml reflects any versions where we know there are bugs / conflicts / whatever.

tych0:
Ok... seems like the coverage fixes worked, so I folded them in.

ervinpopescu:
Sorry for such an obvious question, but why is `/scripts/ffibuild` not needed anymore? Am I missing something in the diff?

tych0:
Sorry, I should have put that in the commit message. Over a series of commits (9770e8cba5306ad1393031ea3a8f6eafb9ed5e67 , 24435adec038ad88f46739433d0bde5b59fda698 , and 5736f4f10fb0f46621276a577635e363a2138fbf ) we removed the need to pre-build ffi stuff except for wayland, so the script ends up just being a wrapper around the single ffi build we still needed. There's no real need for it, we can just call that build script directly.

tych0:
I'm gonna go ahead and land this now, but I think it's worth having a larger conversation how to organize deps based on this thread: https://github.com/qtile/qtile/pull/5398#discussion_r2243792108

In particular, some of the asyncio stuff I have done recently could also potentially add some dependencies, and it would be nice to know how people feel about that.

elParaguayo:
One downside of this seems to be that `uv` doesn't have an easy way to add packages after install. So add-ons like qtile-extras, bonsai etc need to be installed at the same time. 

There's no equivalent of `pipx inject`. 

We don't want to have to add third party addons to our pyproject.toml file.

tych0:
Yes, although reinstalling should be "pretty fast" right? Or is that complicated for some reason? (I appreciate getting the command line exactly right is probably annoying the first time...)

elParaguayo:
Should be quick but it's just a bit stupid to say to people that they need to reinstall qtile just to install qtile-extras.

tych0:
Yeah, fair enough. I suppose another option is that people could muck with `sys.path` from their `config.py` to point at it directly. Neither are great options.

I can appreciate `uv`'s stance of wanting an install to be atomic. I'll try to think about a good way to do this.

PR #5399, opened 2025-07-30
Smarter library loading, fixes OpenBSD support

ekollof:
- Don't hardcode library file names
- Should be compatible with other operating systems.

NOTE: While this will fix loading of pango/gobject/cairo for systems like OpenBSD, there is still an issue with libxcb with OpenBSD's Xenocara. But this is in `xcffib`. I have found a solution for this, but that requires a patch for that.

Proof qtile now works in OpenBSD (at least in Xephyr):

2025-07-30-213156_1277x1017_scrot

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -564,17 +564,12 @@ class _TextBox(_Widget):
     def formatted_text(self):
         return self.fmt.format(self._text)
 
-    @property
-    def actual_padding(self):
-        if self.padding is None:
-            return self.fontsize // 2
-        else:
-            return self.padding
-
     def _configure(self, qtile, bar):
         _Widget._configure(self, qtile, bar)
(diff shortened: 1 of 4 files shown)


ekollof:
xcffib pull request is here: https://github.com/tych0/xcffib/pull/190

ekollof:
I don't know why ruff and mypy don't like these changes. The nix flake thing puzzles me too.

tych0:
Blah, I keep forgetting that you cannot force-push to branches, or github will close them for some reason. Sorry.

`make lint` will run all the linters, and ruff will fix things for you. But I think my fixes to your commit should work, let me re-open sorry.

tych0:
Foo

tych0:
Ugh, see #5400, looks like the authorship is okay, but now somehow github didn't detect you as the author? Blah.

PR #5400, opened 2025-07-31
Smarter library loading, fixes OpenBSD support

tych0:
- Should be compatible with other operating systems.

### DIFF
--- a/libqtile/pangocffi.py
+++ b/libqtile/pangocffi.py
@@ -46,12 +46,21 @@
 #
 # This is not intended to be a complete cffi-based pango binding.
 
+import ctypes.util
 
 from libqtile.pango_ffi import pango_ffi as ffi
 
-gobject = ffi.dlopen("libgobject-2.0.so.0")  # type: ignore
-pango = ffi.dlopen("libpango-1.0.so.0")  # type: ignore
-pangocairo = ffi.dlopen("libpangocairo-1.0.so.0")  # type: ignore
+
+def load_library(name):
+    lib_path = ctypes.util.find_library(name)
+    if lib_path is None:
+        raise ImportError(f"Could not find library: {name}")
+    return ffi.dlopen(lib_path)
+
+
+gobject = load_library("gobject-2.0")
+pango = load_library("pango-1.0")
+pangocairo = load_library("pangocairo-1.0")
 
 
 def patch_cairo_context(cairo_t):


tych0:
cc @jwijenbergh @Gurjaka looks like the nix flake carries patches, and if they don't apply it fails? Is it possible to upstream these?

alternatively, can we set the nix flake to "ignorable" or something so we don't block merging real fixes on it?

Gurjaka:
> cc @jwijenbergh @Gurjaka looks like the nix flake carries patches, and if they don't apply it fails? Is it possible to upstream these?
> 
> alternatively, can we set the nix flake to "ignorable" or something so we don't block merging real fixes on it?

As far as I am aware, our flake doesn't carry any patch, I'll check why it might fail

Gurjaka:
okay I found the issue @tych0 in nixpkgs Qtile actually does have postPatches:
```nix
  postPatch = ''
    substituteInPlace libqtile/pangocffi.py \
      --replace-fail libgobject-2.0.so.0 ${glib.out}/lib/libgobject-2.0.so.0 \
      --replace-fail libpangocairo-1.0.so.0 ${pango.out}/lib/libpangocairo-1.0.so.0 \
      --replace-fail libpango-1.0.so.0 ${pango.out}/lib/libpango-1.0.so.0
    substituteInPlace libqtile/backend/x11/xcursors.py \
      --replace-fail libxcb-cursor.so.0 ${xcbutilcursor.out}/lib/libxcb-cursor.so.0
    substituteInPlace libqtile/backend/wayland/cffi/build.py \
        --replace-fail /usr/include/pixman-1 ${lib.getDev pixman}/include \
        --replace-fail /usr/include/libdrm ${lib.getDev libdrm}/include/libdrm
  '';
```

I will tinker with Qtile overlay to remove them in flake

tych0:
Ok, thank you!

Gurjaka:
> Ok, thank you!

I think I did it, but before submitting PR, I will test on your branch as well

Gurjaka:
Hey @tych0,

I finished the fix I was working on, but I’m now hitting a different issue that’s unrelated to it.

The Qtile tests are currently failing with this error:

```
Invalid MIT-MAGIC-COOKIE-1 key  
xwininfo: error: unable to open display ":0.0"
```

The fix for this test failure has already been merged into `nixpkgs-unstable-small` (we either need it to reach `nixpkgs-unstable` or pin the flake input to a specific commit). However, the [PR that separates Qtile and Qtile-Extras tests](https://github.com/NixOS/nixpkgs/pull/431532)] hasn’t been merged yet. This [earlier PR](https://github.com/NixOS/nixpkgs/pull/421427) should have done it, but since we were focused on the VM test part at the time, the separation part accidentally got overwritten.

Right now, I’ve pinned the flake input to my `nixpkgs` fork, which includes both the test fix and the test separation, so the CI runs cleanly.

We’ve got three options:

1. **Wait** for the Qtile-Extras separation PR to be created, merged, and propagated to `nixos-unstable`, which would also let us remove the Qtile-Extras overlay from the flake — the cleanest solution.
2. **Use my nixpkgs fork** with both the test fix and the separation PR, giving us full control over the test setup — this lets us remove the Qtile-Extras overlay right away.
3. **Ignore the CI failure** caused by the missing separation PR and merge this as-is.

What do you think?

tych0:
I see that the two PRs just got merged, does that change the situation any?

In general, I do not like holding patches from qtile proper because of some out-of-tree thing. The NixOS stuff has been problematic because of this, so whatever it takes so that we have control over our own destiny is good. Maybe that's pointing at your fork, or decoupling from qtile-extras, removing patches, etc.; probably it's all of them.

It seems like (3) is the nuclear option, but if I'm going to ignore CI, I'll just remove it. I expect you don't want that :). So maybe the situation has been resolved with the above merges?

Gurjaka:
> I see that the two PRs just got merged, does that change the situation any?
> 
> In general, I do not like holding patches from qtile proper because of some out-of-tree thing. The NixOS stuff has been problematic because of this, so whatever it takes so that we have control over our own destiny is good. Maybe that's pointing at your fork, or decoupling from qtile-extras, removing patches, etc.; probably it's all of them.
> 
> It seems like (3) is the nuclear option, but if I'm going to ignore CI, I'll just remove it. I expect you don't want that :). So maybe the situation has been resolved with the above merges?

Yup they got merged, but take some time to land in `nixpkgs-unstable`, though I could change flake input to master branch instead of unstable if you'd like.

I really don't want to remove flake-check CI, because then other Flake users will have to deal with the problem jump scare :D

tych0:
Sure, if `master` means that we will be blocked less often, I'm all for it. As much as possible, I would like to peruse a design where the only commits that can break qtile's CI are qtile commits.

Gurjaka:
> Sure, if `master` means that we will be blocked less often, I'm all for it. As much as possible, I would like to peruse a design where the only commits that can break qtile's CI are qtile commits.

`master` doesn't really mean that, but for now we should use one of master or my fork, I would prefer using master. About only Qtile commits breaking Qtile CI, that's already done by https://github.com/NixOS/nixpkgs/pull/431532

jwijenbergh:
Tbh I think the best approach here is making nix CI optional to pass

tych0:
this was fixed as part of #5422, so i'll go ahead and close it.

PR #5397, opened 2025-07-29
[x11] New window stacking logic

elParaguayo:
Draft for now!

I've hated the window stacking logic in our x11 backend for a while. Especially because I wrote a lot of it.

Having looked at the new Wayland backend and seen the concept of sub-trees for different layer groups I felt that was a cleaner way to manager stacking in general. So this PR is an attempt to implement that sort of stacking.

The key idea is a new class `ZManager` which maintains a snapshot of the stacking order of windows across different layer groups. When a window requests to be moved in the stack, the manager updates the stack and then tells the window which other window it needs to be stacked above/below and the window then sends this request to the X server.

This will break a lot of things:
* ~It presumably breaks wayland entirely as there's no implementation of `ZManager` in the wayland backend. Will sort soon.~
* ~I need some logic to restack windows when they lose focus (e.g. fullscreen windows are elevated when they're focused so we need to know when they lose focus. Hooks?)~
* More stuff that I haven't figured out yet.

Also:
* Needs more tests
* Needs more documentation

PR #5401, opened 2025-07-31
[wayc] Implement core.query_tree()

richcarni:
I bundled the 2nd commit because both are required to pass tests: test/test_window.py -> test_bring_front_click()

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -579,4 +579,10 @@ class Core(base.Core):
     @expose_command()
     def query_tree(self) -> list[int]:
         """Get IDs of all mapped windows in ascending Z order."""
-        raise Exception("TODO: implement")
+        wids = []
+        @ffi.callback("void(int)")
+        def loop(wid):
+            wids.append(wid)
+
+        test = lib.qw_server_loop_visible_views(self.qw, loop)
+        return wids
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -512,3 +512,25 @@ bool qw_server_change_vt(struct qw_server *server, int vt) {
     }
     return wlr_session_change_vt(server->session, vt);
 }
+
+static void qw_server_query_iterator(struct wlr_scene_buffer *buffer, int sx, int sy, void *data) {
+    node_wid_cb_t cb = data;
+    // Walk back up tree until we find a window or run out of parents
+    struct wlr_scene_node *node = &buffer->node;
+    while (node) {
+        struct qw_view *view = node->data;
+        if (view && node->enabled) {
+            cb(view->wid);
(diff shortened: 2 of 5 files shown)


elParaguayo:
Looks like you've left lots of debugging logging lines in this.

richcarni:
> Looks like you've left lots of debugging logging lines in this.

Thanks! How embarrassing!!

elParaguayo:
Not at all. I do it all the time!

richcarni:
That looks better!

BTW, I also use a debugger but find it tricky with tests. I think because they launch multiple processes

elParaguayo:
I just put print statements in tests (which I forget to remove when committing) and then check the pytest output.

One point to note, I've been rewriting stacking logic for x11 in #5397 and am separately trying to see if I can use the same `ZManager` framework for Wayland i.e. just using the same method names to call the relevant methods in the wayc backend. Don't think this has any bearing on this PR but just wanted to let you know.

jwijenbergh:
Can you resolve the conflicts? Then LGTM

elParaguayo:
Comments below are general (minor) comments and not meant to hold up merging this.

Based on our separate conversations, I think we're agreed that we don't need zmanager in Wayland. That being said, I would like to have some alignment between backends to ensure consistent behaviour.

Does query_tree give you all windows, not just the visible ones? I can implement a query_tree in X11 and then have a common function that lists the windows in order but shows what group they're in.

In X11, I also have a debugging function that shows which layer group each window is in. Would be nice to do that here too.

Lastly, there's an enum for the layer group (e.g. `lib.LAYER_BRINGTOFRONT`). If we implement the ZManager code (in x11 only), would it be worth mapping the `LayerGroup` enum in that PR to the c enum so it's clear that we expect similar behaviour?

richcarni:
> Based on our separate conversations, I think we're agreed that we don't need zmanager in Wayland. That being said, I would like to have some alignment between backends to ensure consistent behaviour.

Makes a ton of sense to me!

> Does query_tree give you all windows, not just the visible ones? I can implement a query_tree in X11 and then have a common function that lists the windows in order but shows what group they're in.

Just visible windows

> In X11, I also have a debugging function that shows which layer group each window is in. Would be nice to do that here too.

Sounds similar to #5414? I've found this super useful

> Lastly, there's an enum for the layer group (e.g. lib.LAYER_BRINGTOFRONT). If we implement the ZManager code (in x11 only), would it be worth mapping the LayerGroup enum in that PR to the c enum so it's clear that we expect similar behaviour?

Just my opinion but yes, if this is the path we're going down, I think it makes sense to share a common interface (same layers, same layer operations even) and then let each backend take care of the implementation details

richcarni:
No net changes - just cleaned up the debugging comments properly

jwijenbergh:
Thanks!

PR #5402, opened 2025-07-31
remove funding yml

m-col:
There is so much great work happening by others so I think it would be better for someone else to have this :)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,0 @@
-github: ["m-col"]
-custom: ["https://liberapay.com/mcol"]


elParaguayo:
It should be @jwijenbergh for the way work.

jwijenbergh:
> It should be @jwijenbergh for the way work.

Thanks a lot, but I prefer it to go to someone else right now

m-col:
Doing great thank you :smile: tech companies love their macbooks so I sadly spend less time in qtile than I used to :cry: . Hope you guys are all doing well! I've been following along with the development and it seems there is plenty of interesting work going on!

elParaguayo:
> Doing great thank you 😄 tech companies love their macbooks so I sadly spend less time in qtile than I used to 😢 . Hope you guys are all doing well! I've been following along with the development and it seems there is plenty of interesting work going on!

So get working on the Mac backend for qtile!

m-col:
Ha maybe one day!

PR #5404, opened 2025-08-02
fix(images): do not fail on input zero

RFCreate:
This is a rewrite of #5344. Now instead of returning early on zero, attempt the resize/scale and let the width/height be set to one pixel.

Right now when the input is a negative number, it sets the width/height to one pixel, I don't know if that was intended, but using the same mechanism lets apply it to zero input as well instead of raising an error.

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -199,21 +199,20 @@ class Img:
     height = _PixelSize("height")
 
     def resize(self, width=None, height=None):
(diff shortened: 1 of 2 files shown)


tych0:
I think it's probably not intended, but keeping things the same is fine with me. I would really love to get a validation framework in place so that we can validate all of these configuration parameters in `qtile check` somehow.

PR #5407, opened 2025-08-03
[wayc] Add support for scene_buffer borders

elParaguayo:
This PR adds support in the wayc backend for borders to be rendered via wlr_scene_buffer objects. The immediate beneficiary of this is the qtile-extras library which provides different border styles instead of solid colours.

I needed a lot of help from ChatGPT on this one (and also had to correct its mistakes...).

I can't guarantee there aren't any memory leaks here.

### DIFF
--- a/libqtile/backend/wayland/qw/cairo-buffer.c
+++ b/libqtile/backend/wayland/qw/cairo-buffer.c
@@ -1,5 +1,5 @@
 #include "cairo-buffer.h"
-
+#include "wlr/util/log.h"
 #include <drm_fourcc.h>
 #include <wlr/interfaces/wlr_buffer.h>
 
@@ -50,4 +50,59 @@ struct wlr_buffer *cairo_buffer_create(int width, int height, size_t stride, voi
     cairo_buffer->data = data;
     cairo_buffer->stride = stride;
     return &cairo_buffer->base;
+}
+
+// Creates a wlr_buffer from a subregion of a cairo image surface
+struct wlr_buffer *cairo_buffer_from_surface_region(cairo_surface_t *surface, int x, int y,
+                                                    int width, int height) {
+    if (cairo_surface_get_type(surface) != CAIRO_SURFACE_TYPE_IMAGE) {
+        wlr_log(WLR_ERROR, "Cairo surface is not an image surface\n");
+        return NULL;
+    }
+
+    cairo_surface_flush(surface);
+
+    unsigned char *data = cairo_image_surface_get_data(surface);
+    int surface_width = cairo_image_surface_get_width(surface);
+    int surface_height = cairo_image_surface_get_height(surface);
+    int stride = cairo_image_surface_get_stride(surface);
+
+    if (x < 0 || y < 0 || x + width > surface_width || y + height > surface_height) {
+        wlr_log(WLR_ERROR, "Requested subregion (%d,%d %dx%d) is out of bounds (%dx%d)\n", x, y,
+                width, height, surface_width, surface_height);
+        return NULL;
+    }
+
+    // Compute offset pointer for the subregion
+    unsigned char *sub_data = data + y * stride + x * 4;
+
+    return cairo_buffer_create(width, height, stride, sub_data);
+}
+
+// Given a surface, slice it into 4 fixed regions and create scene buffers
+struct wlr_scene_buffer **create_scene_buffers_from_surface(struct wlr_scene_tree *scene,
+                                                            cairo_surface_t *surface,
+                                                            struct wlr_box *regions,
+                                                            int num_regions) {
+    struct wlr_scene_buffer **scene_buffers =
+        calloc(num_regions, sizeof(struct wlr_scene_buffer *));
+    if (!scene_buffers) {
+        wlr_log(WLR_ERROR, "Failed to allocate scene buffer array\n");
+        return NULL;
+    }
+
+    for (int i = 0; i < num_regions; ++i) {
+        struct wlr_buffer *buf = cairo_buffer_from_surface_region(
+            surface, regions[i].x, regions[i].y, regions[i].width, regions[i].height);
+
+        if (!buf) {
+            wlr_log(WLR_ERROR, "Failed to create buffer for region %d\n", i);
+            continue;
+        }
+
+        scene_buffers[i] = wlr_scene_buffer_create(scene, buf);
+        wlr_buffer_drop(buf); // Transfer ownership to the scene buffer
+    }
+
+    return scene_buffers;
 }
--- a/libqtile/backend/wayland/qw/view.c
+++ b/libqtile/backend/wayland/qw/view.c
@@ -1,4 +1,5 @@
 #include "view.h"
+#include "cairo-buffer.h"
 #include "server.h"
 #include <stdlib.h>
 #include <wlr/util/log.h>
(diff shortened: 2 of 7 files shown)


libqtile/backend/wayland/qw/cairo-buffer.c
```diff
@@ -3,6 +3,8 @@
 #include <drm_fourcc.h>
 #include <wlr/interfaces/wlr_buffer.h>
 
+#include <stdio.h>
```
elParaguayo: Ha. Left over from debugging...

libqtile/backend/wayland/qw/cairo-buffer.c
```diff
@@ -51,3 +53,68 @@ struct wlr_buffer *cairo_buffer_create(int width, int height, size_t stride, voi
...
     return &cairo_buffer->base;
 }
+
+// Creates a wlr_buffer from a subregion of a cairo image surface
+struct wlr_buffer *cairo_buffer_from_surface_region(
+    cairo_surface_t *surface,
+    int x, int y, int width, int height
+) {
+    if (cairo_surface_get_type(surface) != CAIRO_SURFACE_TYPE_IMAGE) {
+        fprintf(stderr, "Cairo surface is not an image surface\n");
```
jwijenbergh: `wlr_log(WLR_ERROR` instead. Same for all other fprintf

libqtile/backend/wayland/qw/view.h
```diff
@@ -29,6 +31,30 @@ typedef void (*set_app_id_cb_t)(char *app_id, void *userdata);
...
+            float color[4][4]; // RGBA per side (NESW)
+        } rect;
+
+        struct {
+            cairo_surface_t *surface; // Full border ring image
+        } buffer;
+    };
+};
+
+struct qw_border_region {
```
jwijenbergh: You can use wlr_box instead

libqtile/backend/wayland/qw/cairo-buffer.c
```diff
@@ -51,3 +51,66 @@ struct wlr_buffer *cairo_buffer_create(int width, int height, size_t stride, voi
...
+    }
+
+    cairo_surface_flush(surface);
+
+    unsigned char *data = cairo_image_surface_get_data(surface);
+    int surface_width = cairo_image_surface_get_width(surface);
+    int surface_height = cairo_image_surface_get_height(surface);
+    int stride = cairo_image_surface_get_stride(surface);
+
+    if (x < 0 || y < 0 || x + width > surface_width || y + height > surface_height) {
```
Sigmanificient, line 70: Shouldn't the bound check use `>=` if it starts at 0?
```suggestion
    if (x < 0 || y < 0 || x + width >= surface_width || y + height >= surface_height) {
```

libqtile/backend/wayland/qw/view.h
```diff
@@ -29,6 +31,26 @@ typedef void (*set_app_id_cb_t)(char *app_id, void *userdata);
...
 struct qw_server;
 
+enum qw_border_type {
+    QW_BORDER_RECT,
+    QW_BORDER_BUFFER,
+};
+
+struct qw_border {
+    enum qw_border_type type;
+    int width; // border thickness (for all sides)
```
Sigmanificient: I think this should be unsigned to avoid negative border size

libqtile/backend/wayland/qw/cairo-buffer.c
elParaguayo: No. I don't think so. x+width should equal surface_width for a region touching the right hand side.

If I make the change, it does trigger the error.

Sigmanificient: so x == width is a valid case? But the surface start at 0 right?
If the width is 1, then for x == 0 and x == y the check passes, is this intended?

elParaguayo: Yes, the surface starts at 0, 0 and has dimensions of `surface_width` and `surface_height`. I'm just trying to check that the requested region is contained by the surface.

I should check that `width` > 0 as a 0 width border shouldn't be rendered.

But, in your example, if (border) `width` is 1 and `x=0`, this should pass as long as the surface is at least one pixel wide. If it was exactly one pixel wide then `x + width` equals 1 which is the same as the surface width and so that's fine.

Apologies if I'm missing a different scenario that you had in mind.

libqtile/backend/wayland/qw/view.c
```diff
@@ -74,64 +92,77 @@ bool qw_view_is_visible(struct qw_view *view) {
...
+    int outer_h = view->height + total_width * 2;
+
+    int coord = 0;
+    for (int i = 0; i < border_count; i++) {
+        const struct qw_border *src = &borders[i];
+        view->borders[i].type = src->type;
+
+        int bw = src->width;
+
+        struct wlr_box sides[4] = {
```
Gurjaka, line 129: Hi, Just curious did you forget to re-add comment `// clang-format off` here or it is formatted correctly?

elParaguayo: I probably deleted them by mistake and didn't understand what they're for.

I've not been using any kind of formatter for my wayc code. Should I be?

Gurjaka: It's not necessary, but good practice so code stays consistent. That comment disables clang-format until you enable with another comment. It was there because formatter actually gives it really bad structure like this:
```c
        struct border_pairs pairs[4] = {
            {.x = coord, .y = coord, .w = outer_w - coord * 2, .h = bw}, // top border
            {.x = outer_w - bw - coord,
             .y = bw + coord,
             .w = bw,
             .h = outer_h - bw * 2 - coord * 2}, // right border
            {.x = coord,
             .y = outer_h - bw - coord,
             .w = outer_w - coord * 2,
             .h = bw}, // bottom border
            {.x = coord,
             .y = bw + coord,
             .w = bw,
             .h = outer_h - bw * 2 - coord * 2}, // left border
        };
```

but if you disable formatter, it keeps the style you leave e.g:

```c
        // clang-format off
        struct border_pairs pairs[4] = {
            { .x = coord, .y = coord, .w = outer_w - coord * 2, .h = bw },                          // top border
            { .x = outer_w - bw - coord, .y = bw + coord, .w = bw, .h = outer_h - bw * 2 - coord * 2 },  // right border
            { .x = coord, .y = outer_h - bw - coord, .w = outer_w - coord * 2, .h = bw },          // bottom border
            { .x = coord, .y = bw + coord, .w = bw, .h = outer_h - bw * 2 - coord * 2 },          // left border
        };
        // clang-format on
```

elParaguayo:
I'm inclined to add the desktop wallpaper code to the same PR as there will be some shared code (e.g. wallpaper could also be an image (using a scene buffer) or a colour (using a scene rect)).

jwijenbergh:
> I'm inclined to add the desktop wallpaper code to the same PR as there will be some shared code (e.g. wallpaper could also be an image (using a scene buffer) or a colour (using a scene rect)).

we can also merge already?

elParaguayo:
> > I'm inclined to add the desktop wallpaper code to the same PR as there will be some shared code (e.g. wallpaper could also be an image (using a scene buffer) or a colour (using a scene rect)).
> 
> we can also merge already?

Sure. If I tweak the border code, we can deal with that in the separate PR.

jwijenbergh:
Thanks!

PR #5409, opened 2025-08-04
[wayc] Scene tree: store a reference to internal views

richcarni:
Currently required for query_tree() to count wids correctly.

Internal views have default wid = -1. Unclear whether we need to set the wid, title or other attributes of these views via callback


richcarni:
As I mentioned in the comment, I wasn't sure whether we also either need to or should set the wid, titile, etc of internal-views?

If so, I assume this would be via a callback?

This commit is enough for query-tree() to at least count the window.

richcarni:
OK, yes we at least need the wid for pointer focus/buttons.

I'm going to close this PR and work on internal window focus/buttons

PR #5411, opened 2025-08-04
[wayc] Add methods to hide/show mouse cursor

elParaguayo:


elParaguayo:
Still learning about how c works. Do i need to add those definitions to the header file? I'm guessing not as they're not called from anywhere in the c code at the moment. If that's right, should I remove them?

Gurjaka:
> Still learning about how c works. Do i need to add those definitions to the header file? I'm guessing not as they're not called from anywhere in the c code at the moment. If that's right, should I remove them?

Which methods are you talking about? If `qw_cursor_{hide, show}` then no, they are right and should stay in header, since you are calling them in Python. If you ever implement function, that isn't supposed to go outside the file, then you can create static function e.g `static void function_name` and you don't define them in header files. I haven't tested the code, but looking at the diff, structure seems correct, though does it work? And how can we test it?

elParaguayo:
Yes. I was asking about those two methods I included in the header file.

As for testing, I ran it in a separate tty and it worked fine. Not sure about testing with pytest though.

Gurjaka:
> Yes. I was asking about those two methods I included in the header file.
> 
> As for testing, I ran it in a separate tty and it worked fine. Not sure about testing with pytest though.

Do you have modified config that you use for testing? If so could you share on discord, I'd like to test as well.

libqtile/backend/wayland/qw/cursor.h
```diff
@@ -11,6 +11,11 @@ struct qw_cursor {
     struct wlr_cursor *cursor;
     struct qw_view *view;
 
+    struct wlr_surface *saved_surface;
```
jwijenbergh: Can you move all of these under private data?

elParaguayo: Done.

jwijenbergh (approved):
Thanks!

PR #5413, opened 2025-08-05
[wayc] Handle cursor button for internal windows

richcarni:


elParaguayo:
Thanks for this. 

I've just tested by adding this in my config:
```python
class MouseButtonTest(widget.TextBox):
    def __init__(self, **config):
        super().__init__("Waiting for button press...", **config)

        self.update(f"{button} pressed.")
```
and adding the widget to the bar.

Button clicks are recognised but the scroll buttons aren't mapped to Buttons 4, 5, 6, and 7. They're giving 768-771 which is strange because the mappings in the `qw` code should have picked this up.
```c
enum BUTTON_SCROLL {
    BUTTON_SCROLL_UP = 0x300,
    BUTTON_SCROLL_DOWN = 0x301,
    BUTTON_SCROLL_LEFT = 0x302,
    BUTTON_SCROLL_RIGHT = 0x303,
};
```

richcarni:
Thanks for the mini test! I'll see if I can figure out what's going on

I get the same
0x300 HEX = 768 DEC
What is the expected output? 4, 5, 6, 7? Do we need to change the enum?

elParaguayo:
The button codes should have been mapped to 4-7 here:
```c
int qw_util_get_button_code(uint32_t button) {
    // Array of Linux input event button codes (from linux/input-event-codes.h)
    // These codes represent mouse buttons and scroll events
    uint32_t mappings[] = {
        // BTN_LEFT (left mouse button)
        0x110,
        // BTN_MIDDLE (middle mouse button)
        0x112,
        // BTN_RIGHT (right mouse button)
        0x111,
        BUTTON_SCROLL_UP,
        BUTTON_SCROLL_DOWN,
        BUTTON_SCROLL_LEFT,
        BUTTON_SCROLL_RIGHT,
        // BTN_SIDE (side mouse button)
        0x113,
        // BTN_EXTRA (extra mouse button)
        0x114,
    };

    // Calculate the number of elements in the mappings array
    int n = sizeof(mappings) / sizeof(mappings[0]);

    for (int i = 0; i < n; ++i) {
        if (button == mappings[i]) {
            // Return the 1-based index of the button mapping
            return i + 1;
        }
    }

    return 0;
}
```

richcarni:
Should work as expected now

elParaguayo:
Yes. The widget shows the correct output now.

Sigmanificient:
@elParaguayo I would suggest having a macro because this is something that tend to appear quite often in C
```c
#define LENGTH_OF(arr) (sizeof (arr) / sizeof *(arr))
```

Also, you might want to prefer `unsigned` types for iterating through array (they match `sizeof` type (which is `size_t`))

elParaguayo:
@Sigmanificient Thanks but none of this is my code (other than the widget to test this). The code in my comment above was lifted straight from the `wayc` branch.

jwijenbergh:
Thanks!

PR #5414, opened 2025-08-05
[wayc] Add scene graph dump tool

richcarni:
Draws the scene graph tree to log

### DIFF
--- a/libqtile/backend/wayland/qw/debug_util.c
+++ b/libqtile/backend/wayland/qw/debug_util.c
@@ -1,0 +1,93 @@
+#include "debug_util.h"
+
+static void print_scene_node(struct wlr_scene_node *node, int depth, bool *draw_verticals, int child_i, int child_count, bool hide_rects) {
+    if (hide_rects && node->type == WLR_SCENE_NODE_RECT) {
+        return;
+    }
+
+    bool is_last = false;
+    if (child_i == child_count-1) {
+        is_last = true;
+    }
+
+    const char *type = "unknown";
+    switch (node->type) {
+        case WLR_SCENE_NODE_TREE:   type = "tree"; break;
+        case WLR_SCENE_NODE_RECT:   type = "rect"; break;
+        case WLR_SCENE_NODE_BUFFER: type = "buffer"; break;
+    }
+
+    const char* layer_name = "unknown";
+    switch (child_i) {
+        case LAYER_BACKGROUND:   layer_name = "LAYER_BACKGROUND"; break;
+        case LAYER_BOTTOM:       layer_name = "LAYER_BOTTOM"; break;
+        case LAYER_KEEPBELOW:    layer_name = "LAYER_KEEPBELOW"; break;
+        case LAYER_LAYOUT:       layer_name = "LAYER_LAYOUT"; break;
+        case LAYER_KEEPABOVE:    layer_name = "LAYER_KEEPABOVE"; break;
+        case LAYER_MAX:          layer_name = "LAYER_MAX"; break;
+        case LAYER_FULLSCREEN:   layer_name = "LAYER_FULLSCREEN"; break;
+        case LAYER_BRINGTOFRONT: layer_name = "LAYER_BRINGTOFRONT"; break;
+        case LAYER_TOP:          layer_name = "LAYER_TOP"; break;
+        case LAYER_OVERLAY:      layer_name = "LAYER_OVERLAY"; break;
+        case LAYER_END:          layer_name = "LAYER_END"; break;
+    }
+
+    struct qw_view *view = node->data;
+
+    // Print visual indent tree structure
+    char line[512] = {0};
+    char *p = line;
+    for (int i = 1; i < depth; i++) {
+        if (draw_verticals[i]) {
+            p += sprintf(p, "│   ");
+        } else {
+            p += sprintf(p, "    ");
+        }
+    }
+    if (depth > 0) {
+        p += sprintf(p, is_last ? "└── " : "├── ");
+    }
+
+    // Append node info
+    if (view) {
+        if (view->wid == -1) {
+            sprintf(p, "%s (%sInternal)",
+                    type, node->enabled ? "" : "disabled, ");
+            wlr_log(WLR_ERROR, "%s", line);
+        } else {
+            sprintf(p, "%s (%sname=%s, wid=%d)",
+                    type, node->enabled ? "" : "disabled, ", view->title, view->wid);
+            wlr_log(WLR_ERROR, "%s", line);
+        }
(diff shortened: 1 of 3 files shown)


richcarni:
Not perfect but I've found it useful while debugging failing tests...

Why WLR_ERROR? Because it's the only level I can get to output for tests - I couldn't get output from INFO or DEBUG. I think it's a me issue

The other shortcoming is how to identify internal windows. I don't think internal windows have names? We could set the name to "Internal" or the shell variable to "Internal" and use that. Another option would be check if the buffer has a surface, and if not assume internal

libqtile/backend/wayland/qw/debug_util.c
```diff
@@ -0,0 +1,93 @@
...
+        case LAYER_KEEPBELOW:    layer_name = "LAYER_KEEPBELOW"; break;
+        case LAYER_LAYOUT:       layer_name = "LAYER_LAYOUT"; break;
+        case LAYER_KEEPABOVE:    layer_name = "LAYER_KEEPABOVE"; break;
+        case LAYER_MAX:          layer_name = "LAYER_MAX"; break;
+        case LAYER_FULLSCREEN:   layer_name = "LAYER_FULLSCREEN"; break;
+        case LAYER_BRINGTOFRONT: layer_name = "LAYER_BRINGTOFRONT"; break;
+        case LAYER_TOP:          layer_name = "LAYER_TOP"; break;
+        case LAYER_OVERLAY:      layer_name = "LAYER_OVERLAY"; break;
+        case LAYER_END:          layer_name = "LAYER_END"; break;
+    }
```
Sigmanificient: Since this is a simple enum, maybe consider using a Look-Up Table?
```c
// at the top of the file
char *LAYER_NAMES[] = {
  [LAYER_BACKGROUND] = "LAYER_BACKGROUND",
  [LAYER_BOTTOM] = "LAYER_BOTTOM",
  [LAYER_KEEPBELOW] = "LAYER_KEEPBELOW"
  // ...
};
```
and even forward it to the header
```c
extern char *LAYER_NAMES[LAYER_COUNT];
```

I would personally hand this with a X-macro pattern, but this might be a bit hard for other to read/maintain

libqtile/backend/wayland/qw/debug_util.c
```diff
@@ -0,0 +1,93 @@
...
+
+    const char *type = "unknown";
+    switch (node->type) {
+        case WLR_SCENE_NODE_TREE:   type = "tree"; break;
+        case WLR_SCENE_NODE_RECT:   type = "rect"; break;
+        case WLR_SCENE_NODE_BUFFER: type = "buffer"; break;
+    }
+
+    const char* layer_name = "unknown";
```
Sigmanificient: Since the layer types are known and limited, is there a need for a `unknown` one (maybe for some kind of extension logic)?

libqtile/backend/wayland/qw/debug_util.c
```diff
@@ -0,0 +1,93 @@
...
+        case LAYER_BRINGTOFRONT: layer_name = "LAYER_BRINGTOFRONT"; break;
+        case LAYER_TOP:          layer_name = "LAYER_TOP"; break;
+        case LAYER_OVERLAY:      layer_name = "LAYER_OVERLAY"; break;
+        case LAYER_END:          layer_name = "LAYER_END"; break;
+    }
+
+    struct qw_view *view = node->data;
+
+    // Print visual indent tree structure
+    char line[512] = {0};
```
Sigmanificient: how is this length computed? having random hard-coded number is typically something you want to avoid in string buffer, because it may cause very bad effects

libqtile/backend/wayland/qw/debug_util.c
```diff
@@ -0,0 +1,93 @@
...
+    for (int i = 1; i < depth; i++) {
+        if (draw_verticals[i]) {
+            p += sprintf(p, "│   ");
+        } else {
+            p += sprintf(p, "    ");
+        }
+    }
+    if (depth > 0) {
+        p += sprintf(p, is_last ? "└── " : "├── ");
+    }
```
Sigmanificient: Use `snprintf` rather than `sprintf` to have the size constraint respected

libqtile/backend/wayland/qw/debug_util.c
```diff
@@ -0,0 +1,93 @@
+#include "debug_util.h"
+
+static void print_scene_node(struct wlr_scene_node *node, int depth, bool *draw_verticals, int child_i, int child_count, bool hide_rects) {
```
Sigmanificient: Can the depth be negative? Otherwise I would suggest to restrict the type.

Judging by the later print logic, depth could also be limited to a `char` (4 character printed for each level of depth, with a 512 limit => 128 is enough to overflow)

Sigmanificient (changes requested):

richcarni:
@Sigmanificient thanks for suggestions

This is only intended for debugging purposes so yes, I have taken some shortcuts

If you'd like to help out, I'd be happy to take PRs to my branch!

libqtile/backend/wayland/qw/debug_util.c
richcarni: Made this change but made LAYER_NAMES static for now

libqtile/backend/wayland/qw/debug_util.c
richcarni: I suppose in the unlikely event that they add a node type

libqtile/backend/wayland/qw/debug_util.c
richcarni: Moved to a setting. Still hard coded but added checks for overflow (snprintf). Especially for debugging, unlikely to have trees with lines longer than this

libqtile/backend/wayland/qw/debug_util.c
richcarni: Thanks. Done

libqtile/backend/wayland/qw/debug_util.c
richcarni: Added a depth check to prevent overflow

richcarni:
Addressed comments. Rebased

richcarni:
Not that I want to spend too much time on this, but have an idea for an alternative implementation...

What if we just create a few c functions to allow traversing and extracting data from the scene graph on the python side. This would avoid all the messy c string manipulation and give us more flexibility to extract and view relevant data

With hooks, could even build a live view that updates as windows are created and destroyed

richcarni:
Actually, like query_tree, seems the best pattern here is also going to be traverse in c, calling back to python as it goes, building a representation of the tree in python

PR #5420, opened 2025-08-06
nix: lots of refactor.

Gurjaka:
Read comment messages, we tried hard xD

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -1,55 +1,58 @@
 self: final: prev:
 let
-  inherit (prev) lib pkgs;
+  inherit (prev) pkgs lib;
 
   build-config = import ./build-config.nix pkgs;
+
+  local-version =
+    with builtins;
+    let
+      flakever = self.shortRev or "dev";
+
+      symver = lib.pipe (readFile ../CHANGELOG) [
+        (split "\n")
+        (filter (x: !isList x && lib.strings.hasPrefix "Qtile" x))
+        (release: elemAt release 1) # the 0th element is the template
+        (match "Qtile ([0-9.]+), released ([0-9-]+):")
+        head
+      ];
+    in
+    "${symver}+${flakever}.flake";
+
+  qtile-override-func =
+    qtile-prev:
+    {
+      name = "${qtile-prev.pname}-${qtile-prev.version}";
+
+      version = local-version;
+
+      src = ./..; # use the source of the git repo
+    }
+    // {
+      env = build-config.resolved-env-vars;
+
+      pypaBuildFlags = build-config.resolved-config-settings;
+
+      postPatch = ""; # removes nixpkgs postPatch
+    };
 in
 {
   pythonPackagesOverlays = (prev.pythonPackagesOverlays or [ ]) ++ [
-    (_: pprev: {
-      qtile =
-        (pprev.qtile.overrideAttrs (
-          old:
-          let
-            flakever = self.shortRev or "dev";
-
-            releases = (
-              builtins.filter (x: !builtins.isList x && prev.lib.strings.hasPrefix "Qtile" x) (
-                builtins.split "\n" (builtins.readFile ../CHANGELOG)
-              )
-            );
-
-            # the 0th element is the template
-            current-release-title = builtins.elemAt releases 1;
-
-            symver = builtins.head (
-              builtins.match "Qtile ([0-9.]+), released ([0-9-]+):" current-release-title
-            );
-          in
-          {
-            version = "${symver}+${flakever}.flake";
-            # use the source of the git repo
-            src = ./..;
-            disabled = false;
-
-            env = build-config.resolved-env-vars;
-
-            pypaBuildFlags = build-config.resolved-config-settings;
-
-            # removes nixpkgs postPatch
-            postPatch = "";
-          }
-        )).override
-          { wlroots = prev.wlroots_0_17; };
-    })
+    (
+      _: py-prev:
+      let
+        qtile = py-prev.qtile.overrideAttrs qtile-override-func;
+      in
+      {
+        qtile = qtile.override { wlroots = pkgs.wlroots_0_17; };
+      }
+    )
   ];
-  python3 =
-    let
-      self = prev.python3.override {
-        inherit self;
-        packageOverrides = prev.lib.composeManyExtensions final.pythonPackagesOverlays;
-      };
-    in
-    self;
+
+  python3 = prev.python3.override {
+    self = final.python3;
+    packageOverrides = lib.composeManyExtensions final.pythonPackagesOverlays;
+  };
+
   python3Packages = final.python3.pkgs;
 }
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -32,8 +32,12 @@ let
       env = build-config.resolved-env-vars;
 
       pypaBuildFlags = build-config.resolved-config-settings;
+    }
+    // {
+      # removes nixpkgs patching, as we handle it locally
+      postPatch = "";
 
-      postPatch = ""; # removes nixpkgs postPatch
+      patches = [ ];
     };
 in
 {
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -28,14 +28,6 @@ self: final: prev: {
           }
         )).override
           { wlroots = prev.wlroots_0_17; };
-
-      qtile-extras = pprev.qtile-extras.overridePythonAttrs ({
-        # disable all tests in these directories
-        disabledTestPaths = [
-          "test/widget/*"
-          "test/popup/*"
-        ];
-      });
     })
   ];
   python3 =
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -1,4 +1,10 @@
-self: final: prev: {
+self: final: prev:
+let
+  inherit (prev) lib pkgs;
+
+  build-config = import ./build-config.nix pkgs;
+in
+{
   pythonPackagesOverlays = (prev.pythonPackagesOverlays or [ ]) ++ [
     (_: pprev: {
       qtile =
@@ -25,6 +31,13 @@ self: final: prev: {
             # use the source of the git repo
             src = ./..;
             disabled = false;
+
+            env = build-config.resolved-env-vars;
+
+            pypaBuildFlags = build-config.resolved-config-settings;
+
+            # removes nixpkgs postPatch
+            postPatch = "";
           }
         )).override
           { wlroots = prev.wlroots_0_17; };
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -30,7 +30,6 @@ prune bin
 prune docs
 prune scripts
 prune rpm
-include scripts/ffibuild
 
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
--- a/flake.lock
+++ b/flake.lock
@@ -2,17 +2,17 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1750776420,
-        "narHash": "sha256-/CG+w0o0oJ5itVklOoLbdn2dGB0wbZVOoDm4np6w09A=",
+        "lastModified": 1754511038,
+        "narHash": "sha256-fie5dSsWbHKtR1443lYwhUxeVk2dEQrq3L+FKOKPpGk=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "30a61f056ac492e3b7cdcb69c1e6abdcf00e39cf",
+        "rev": "a66fdcbf442afb2e4c3b8795c73a3c3ded92392a",
         "type": "github"
       },
       "original": {
         "owner": "nixos",
-        "ref": "nixos-unstable",
         "repo": "nixpkgs",
+        "rev": "a66fdcbf442afb2e4c3b8795c73a3c3ded92392a",
         "type": "github"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -2,7 +2,10 @@
   description = "Qtile's flake, full-featured, hackable tiling window manager written and configured in Python";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
+    # note: we are using a commit hash to a fix,
+    # this should be removed one the pr has landed into unstable, see:
+    # https://nixpk.gs/pr-tracker.html?pr=431532
+    nixpkgs.url = "github:nixos/nixpkgs/a66fdcbf442afb2e4c3b8795c73a3c3ded92392a";
   };
 
   outputs =
--- a/flake.nix
+++ b/flake.nix
@@ -1,13 +1,11 @@
 {
   description = "Qtile's flake, full-featured, hackable tiling window manager written and configured in Python";
-
   inputs = {
     # note: we are using a commit hash to a fix,
     # this should be removed one the pr has landed into unstable, see:
     # https://nixpk.gs/pr-tracker.html?pr=431532
     nixpkgs.url = "github:nixos/nixpkgs/a66fdcbf442afb2e4c3b8795c73a3c3ded92392a";
   };
-
   outputs =
     {
       self,
@@ -32,103 +30,94 @@
           in
           function (import nixpkgs nixpkgs-settings)
         );
+
+      flake-attributes = forAllSystems (pkgs: rec {
+        build-config = import ./nix/build-config.nix pkgs;
+
+        common-python-deps = with pkgs.python3Packages; [
+          python-dateutil
+
+          # docs building
+          numpydoc
+          sphinx
+          sphinx_rtd_theme
+
+          # tests
+          coverage
+          pytest
+          isort
+        ];
+
+        tests = {
+          wayland = pkgs.writeScriptBin "qtile-run-tests-wayland" ''
+            pytest -x --backend=wayland
+          '';
+          x11 = pkgs.writeScriptBin "qtile-run-tests-x11" ''
+            pytest -x --backend=x11
+          '';
+        };
+
+        common-system-deps = with pkgs; [
+          # Gdk namespaces
+          wrapGAppsHook
+          gobject-introspection
+
+          # docs graphs
+          graphviz
+
+          # x11 deps
+          xorg.xorgserver
+          xorg.libX11
+          wlroots_0_17
+
+          # test/backend/wayland/test_window.py
+          gtk-layer-shell
+          imagemagick
+
+          pre-commit
+        ];
+
+        shell-env = {
+          LD_LIBRARY_PATH =
+            with pkgs;
+            lib.makeLibraryPath [
+              glib
+              pango
+              xcb-util-cursor
+              pixman
+              libdrm.dev
+            ];
+        }
+        // build-config.resolved-env-vars;
+
+        pkgs-wrapped = pkgs.lib.lists.flatten [
+          common-python-deps
+          common-system-deps
+          (builtins.attrValues tests)
+        ];
+      });
     in
     {
       formatter = forAllSystems (pkgs: pkgs.nixfmt-rfc-style);
 
       checks = forAllSystems (pkgs: pkgs.python3Packages.qtile.passthru.tests);
 
-      overlays.default = import ./nix/overlays.nix self;
-
       packages = forAllSystems (pkgs: {
         inherit (pkgs.python3Packages) qtile;
-
         default = self.packages.${pkgs.system}.qtile;
       });
 
-      devShells = forAllSystems (
-        pkgs:
-        let
-          common-python-deps =
-            ps:
-            with ps;
-            [ python-dateutil ]
-            ++ [
-              # docs building
-              numpydoc
-              sphinx
-              sphinx_rtd_theme
-              # tests
-              coverage
-              pytest
-              isort
-            ];
+      devShells = forAllSystems (pkgs: {
+        default = pkgs.mkShell {
+          env = flake-attributes.${pkgs.system}.shell-env;
 
-          tests = {
-            wayland = pkgs.writeScriptBin "qtile-run-tests-wayland" ''
-              ./scripts/ffibuild -v
-              pytest -x --backend=wayland
-            '';
+          shellHook = ''
+            export PYTHONPATH=$(readlink -f .):$PYTHONPATH
+          '';
 
-            x11 = pkgs.writeScriptBin "qtile-run-tests-x11" ''
-              ./scripts/ffibuild -v
-              pytest -x --backend=x11
-            '';
-          };
-
-          common-system-deps =
-            with pkgs;
-            [
-              # Gdk namespaces
-              wrapGAppsHook
-              gobject-introspection
-
-              # docs graphs
-              graphviz
-
-              # x11 deps
-              xorg.xorgserver
-              xorg.libX11
-
-              wlroots_0_17
-              # test/backend/wayland/test_window.py
-              gtk-layer-shell
-              imagemagick
-            ]
-            ++ (builtins.attrValues tests);
-        in
-        {
-          default = pkgs.mkShell {
-            env = {
-              QTILE_PIXMAN_PATH = "${pkgs.pixman}/include/pixman-1";
-              QTILE_LIBDRM_PATH = "${pkgs.libdrm.dev}/include/libdrm";
-
-              LD_LIBRARY_PATH =
-                with pkgs;
-                lib.makeLibraryPath [
-                  glib
-                  pango
-                  xcb-util-cursor
-                  pixman
-                  libdrm.dev
-                ];
-            };
-
-            shellHook = ''
-              export PYTHONPATH=$(readlink -f .):$PYTHONPATH
-            '';
-
-            inputsFrom = [ self.packages.${pkgs.system}.qtile ];
-
-            packages =
-              with pkgs;
-              [
-                (python3.withPackages common-python-deps)
-                pre-commit
-              ]
-              ++ common-system-deps;
-          };
-        }
-      );
+          inputsFrom = [ self.packages.${pkgs.system}.qtile ];
+          packages = flake-attributes.${pkgs.system}.pkgs-wrapped;
+        };
+      });
     };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -40,20 +40,11 @@
 
       overlays.default = import ./nix/overlays.nix self;
 
-      packages = forAllSystems (
-        pkgs:
-        let
-          qtile' = pkgs.python3Packages.qtile;
-        in
-        {
-          default = self.packages.${pkgs.system}.qtile;
+      packages = forAllSystems (pkgs: {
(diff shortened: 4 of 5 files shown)


Gurjaka:
I would like to get review either from @Sigmanificient or @jwijenbergh before getting this merged.

Gurjaka:
The flake check CI here is going to fail, don't mind that, it's supposed to!

nix/overlays.nix
```diff
@@ -25,17 +25,21 @@ self: final: prev: {
             # use the source of the git repo
             src = ./..;
             disabled = false;
+            postPatch = ''
+              substituteInPlace libqtile/pangocffi.py \
+              	--replace 'load_library("gobject-2.0")' 'ffi.dlopen("${prev.glib.out}/lib/libgobject-2.0.so.0")' \
+              	--replace 'load_library("pango-1.0")' 'ffi.dlopen("${prev.pango.out}/lib/libpango-1.0.so.0")' \
+              	--replace 'load_library("pangocairo-1.0")' 'ffi.dlopen("${prev.pango.out}/lib/libpangocairo-1.0.so.0")'
```
Sigmanificient: `--replace` is deprecated, use either `--replace-fail`, `--replace-warning` or `--replace-silent` (I would prefer failing because this is critical for the wm to work)

nix/overlays.nix
```diff
@@ -25,17 +25,21 @@ self: final: prev: {
             # use the source of the git repo
             src = ./..;
             disabled = false;
+            postPatch = ''
+              substituteInPlace libqtile/pangocffi.py \
+              	--replace 'load_library("gobject-2.0")' 'ffi.dlopen("${prev.glib.out}/lib/libgobject-2.0.so.0")' \
```
Sigmanificient: Btw we could use `${lib.getLib pkg}/lib/libpkg.so` instead of `${pkgs.out}/lib/libpkg.so`

Sigmanificient (changes requested):

nix/overlays.nix
tych0: Can we do this as a patch to the library loading instead of this regex? Like, if we're on nix, we can do the special lookups? Or do we only know these paths at build time?

Sigmanificient: There is a way to apply a custom patch with variable substitution!

Gurjaka: Well this is the way qtile in nixpkgs used before, see https://github.com/NixOS/nixpkgs/blob/ce01daebf8489ba97bd1609d185ea276efdeb121/pkgs/development/python-modules/qtile/default.nix#L53

Sigmanificient:
@tych0 The goal of the nix builder is to achieve reproducibility by isolating every package into it own small nix store unit, so the path to the library are meant to be known at build time and patched in some way. 
Trying to find them dynamically is often discouraged if it can be done at build time. Technically we could do it by prefixing `LD_LIBRARY_PATH` via `makeWrapper` (i think at least).

Here is a proposal of a alternative solution to the `substituteInPlace`. It change the library at build time, with a clean patch. This approach (and the following example) is taken [from the cairocffi derivation on nixpkgs](https://github.com/NixOS/nixpkgs/blob/38384587c58dd1b58784e945916a8be79defcfa9/pkgs/development/python-modules/cairocffi/default.nix#L33), but applying it in this context would be surprisingly similar:

```nix
{
  # ...
  patches = [
    # OSError: dlopen() failed to load a library: gdk-pixbuf-2.0 / gdk-pixbuf-2.0-0
    (replaceVars ./dlopen-paths.patch {
      ext = stdenv.hostPlatform.extensions.sharedLibrary;
      cairo = cairo.out;
      glib = glib.out;
      gdk_pixbuf = gdk-pixbuf.out;
    })
  ];
}
```

Here what there patch look like:
```patch
--- a/cairocffi/__init__.py
+++ b/cairocffi/__init__.py
@@ -22,6 +22,14 @@ VERSION = __version__ = '1.7.1'
 version = '1.17.2'
 version_info = (1, 17, 2)
 
+# Use hardcoded soname, because ctypes.util use gcc/objdump which shouldn't be
+# required for runtime
+_LIBS = {
+    'cairo': '@cairo@/lib/libcairo@ext@',
+    'glib-2.0': '@glib@/lib/libglib-2.0@ext@',
+    'gobject-2.0': '@glib@/lib/libgobject-2.0@ext@',
+    'gdk_pixbuf-2.0': '@gdk_pixbuf@/lib/libgdk_pixbuf-2.0@ext@',
+}
(truncated for brivety)
```

tych0:
>  Trying to find them dynamically is often discouraged if it can be done at build time. Technically we could do it by prefixing LD_LIBRARY_PATH via makeWrapper (i think at least).

Ok, thanks for that clarification. I didn't realize the intent was to hard code everything.

> It change the library at build time, with a clean patch.

My problem is that the way things are now with the patch approach, if the patch suddenly fails to apply because we make some change (like the one in this PR), our CI suddenly fails, and we have to wait for nixpkgs to update before it passes again. This process typically takes several weeks, which is annoying.

I can think of a few resolutions:

1. the patch lives in the qtile tree, so people changing this code would have to update the nixpkgs patch as well. I expect that doesn't work so well given the architecture of nixpkgs, but I don't really know.
2. We do something like https://github.com/tych0/xcffib/blob/master/module/__init__.py#L27-L34 but then also add in a `platform.freedesktop_os_release()["NAME"] == "NixOS"` conditional, put some "@NIXOS_LIBPATH@" strings that can be replaced using a hopefully more robust sed or something instead of a patch apply. Alternatively, we could make the thing that decides this into a function, and nix could monkey patch it at runtime to return the right thing. This would still be a build time patch, but in the entrypoint, which we leave to setuptools or whatever to define anyway.
3. we remove the nix flake from our CI.

I don't have a problem carrying nix specific code, but I would like it to not break CI for weeks at a time requiring us to get patches merged elsewhere to fix things.

tych0:
Oh, or perhaps there's an option (4) that makes the most sense: we add a `--config-setting=PANGO_SONAME=...` `--config-setting=PANGOCAIRO_SONAME=...`. Then we can have the defaults as they are, but allow nix to override at install time. I think that fixes both sets of concerns?

Sigmanificient:
It is quite true that the library loading patching logic of nix can really be fragile. I am very thankful that you are open to the discussion, and hope this will bring a great long-term solution!

From experience, I don't see any issue from fetching a patch from qtile tree (solution 1)! We can even directly reference it using `"${src}/path/to/the.patch"`. Solution 2 and 4 are very reasonable, with maybe a preference for the later, as it is what many toolchain/project (configure, cmake, ...) do.

## 

However, it seems that these approaches are missing a slight issue: currently, the variable substitution is still done on nixpkgs side and inherited by the flake (which fetch nixpkgs qtile derivation and then override it). Thanksfully, I think we can setup a small layer to replace whatever substitution/providing logic would be done in the nixpkgs side to a small manageable attrset of the libraries on qtile side. Under the hood, it call `replaceVars`, or `--config-setting=${libname}_SONAME=` with a small utility function we wrote and control locally.

This would integrate well with solution 2 and 4 and reduce the friction a lot when there is a lib added, removed, renamed, etc...

## 

Lastly (lets think long term!) we need to consider that there is [another patch from nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/development/python-modules/qtile/fix-restart.patch) that theoretically could cause a CI failure one day. We can force it to a local copy by overriding it too, to point to a copy of the patch from the qtile repo (this would make nixpkgs future patches opt-in, so it would by default not cause trouble).

Sigmanificient:
Regardless of the abstraction approach, what need to be rewritten is the following part:
```nix
  postPatch = ''
    substituteInPlace libqtile/pangocffi.py \
      --replace-fail libgobject-2.0.so.0 ${glib.out}/lib/libgobject-2.0.so.0 \
      --replace-fail libpangocairo-1.0.so.0 ${pango.out}/lib/libpangocairo-1.0.so.0 \
      --replace-fail libpango-1.0.so.0 ${pango.out}/lib/libpango-1.0.so.0
    substituteInPlace libqtile/backend/x11/xcursors.py \
      --replace-fail libxcb-cursor.so.0 ${xcbutilcursor.out}/lib/libxcb-cursor.so.0
    substituteInPlace libqtile/backend/wayland/cffi/build.py \
        --replace-fail /usr/include/pixman-1 ${lib.getDev pixman}/include \
        --replace-fail /usr/include/libdrm ${lib.getDev libdrm}/include/libdrm
  '';
```

Here is a rough implementation of solution 4 on the nix side could look like:
```nix
let
# exposed:
  runtime-libs = {
    "gobject-2.0" = pkgs.glib;
    "pango-1.0" = pkgs.pango;
    "pangocairo-1.0" = pkgs.pango;
    "xcb-cursor" = pkgs.xcb-util-cursor;
  };

  headers = {
    "pixman-1" = pkgs.pixman;
    "libdrm" = pkgs.libdrm;
  };

# internal logic:
  inherit (lib.strings) concatStringsSep;
  inherit (lib.attrsets) mapAttrs attrValues;

  mk-flags = take: map': attrs: (attrValues
    (mapAttrs
      (name: value: "--config-setting:${name}=${map' name (take value)}")
      attrs));

in (
  (mk-flags lib.getLib (name: value: value + "/lib/${name}.so") runtime-libs)
  ++ (mk-flags lib.getDev (name: value: value + "/include/${name}") headers)
)
```

outputting the following:

```
[
  "--config-setting:gobject-2.0=/nix/store/8j6yqx8qfl0g1nwbshq3502hzv36gai6-glib-2.84.2/lib/gobject-2.0.so"
  "--config-setting:pango-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pango-1.0.so"
  "--config-setting:pangocairo-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pangocairo-1.0.so"
  "--config-setting:xcb-cursor=/nix/store/p3wm9m7w0bak9y3r3kyy28l5klcw0ngh-xcb-util-cursor-0.1.5/lib/xcb-cursor.so"
  "--config-setting:libdrm=/nix/store/sqlyy8hdm45y2y37g6qwzh8n9j8a7wbd-libdrm-2.4.124-dev/include/libdrm"
  "--config-setting:pixman-1=/nix/store/clyjdzxjmdvsk2bs7503i2qdvi9axcd8-pixman-0.46.0/include/pixman-1"
]
```

jwijenbergh:
> @tych0 The goal of the nix builder is to achieve reproducibility by isolating every package into it own small nix store unit, so the path to the library are meant to be known at build time and patched in some way. Trying to find them dynamically is often discouraged if it can be done at build time. Technically we could do it by prefixing `LD_LIBRARY_PATH` via `makeWrapper` (i think at least).
> 
> Here is a proposal of a alternative solution to the `substituteInPlace`. It change the library at build time, with a clean patch. This approach (and the following example) is taken [from the cairocffi derivation on nixpkgs](https://github.com/NixOS/nixpkgs/blob/38384587c58dd1b58784e945916a8be79defcfa9/pkgs/development/python-modules/cairocffi/default.nix#L33), but applying it in this context would be surprisingly similar:
> 
> ```nix
> {
>   # ...
>   patches = [
>     # OSError: dlopen() failed to load a library: gdk-pixbuf-2.0 / gdk-pixbuf-2.0-0
>     (replaceVars ./dlopen-paths.patch {
>       ext = stdenv.hostPlatform.extensions.sharedLibrary;
>       cairo = cairo.out;
>       glib = glib.out;
>       gdk_pixbuf = gdk-pixbuf.out;
>     })
>   ];
> }
> ```
> 
> Here what there patch look like:
> 
> ```diff
> --- a/cairocffi/__init__.py
> +++ b/cairocffi/__init__.py
> @@ -22,6 +22,14 @@ VERSION = __version__ = '1.7.1'
>  version = '1.17.2'
>  version_info = (1, 17, 2)
>  
> +# Use hardcoded soname, because ctypes.util use gcc/objdump which shouldn't be
> +# required for runtime
> +_LIBS = {
> +    'cairo': '@cairo@/lib/libcairo@ext@',
> +    'glib-2.0': '@glib@/lib/libglib-2.0@ext@',
> +    'gobject-2.0': '@glib@/lib/libgobject-2.0@ext@',
> +    'gdk_pixbuf-2.0': '@gdk_pixbuf@/lib/libgdk_pixbuf-2.0@ext@',
> +}
> (truncated for brivety)
> ```

I don't agree with this, we removed patches in the qtile nix derivation to make it easier to update

jwijenbergh:
> Regardless of the abstraction approach, what need to be rewritten is the following part:
> 
> ```nix
>   postPatch = ''
>     substituteInPlace libqtile/pangocffi.py \
>       --replace-fail libgobject-2.0.so.0 ${glib.out}/lib/libgobject-2.0.so.0 \
>       --replace-fail libpangocairo-1.0.so.0 ${pango.out}/lib/libpangocairo-1.0.so.0 \
>       --replace-fail libpango-1.0.so.0 ${pango.out}/lib/libpango-1.0.so.0
>     substituteInPlace libqtile/backend/x11/xcursors.py \
>       --replace-fail libxcb-cursor.so.0 ${xcbutilcursor.out}/lib/libxcb-cursor.so.0
>     substituteInPlace libqtile/backend/wayland/cffi/build.py \
>         --replace-fail /usr/include/pixman-1 ${lib.getDev pixman}/include \
>         --replace-fail /usr/include/libdrm ${lib.getDev libdrm}/include/libdrm
>   '';
> ```
> 
> Here is a rough implementation of solution 4 on the nix side could look like:
> 
> ```nix
> let
> # exposed:
>   runtime-libs = {
>     "gobject-2.0" = pkgs.glib;
>     "pango-1.0" = pkgs.pango;
>     "pangocairo-1.0" = pkgs.pango;
>     "xcb-cursor" = pkgs.xcb-util-cursor;
>   };
> 
>   headers = {
>     "pixman-1" = pkgs.pixman;
>     "libdrm" = pkgs.libdrm;
>   };
> 
> # internal logic:
>   inherit (lib.strings) concatStringsSep;
>   inherit (lib.attrsets) mapAttrs attrValues;
> 
>   mk-flags = take: map': attrs: (attrValues
>     (mapAttrs
>       (name: value: "--config-setting:${name}=${map' name (take value)}")
>       attrs));
> 
> in (
>   (mk-flags lib.getLib (name: value: value + "/lib/${name}.so") runtime-libs)
>   ++ (mk-flags lib.getDev (name: value: value + "/include/${name}") headers)
> )
> ```
> 
> outputting the following:
> 
> ```
> [
>   "--config-setting:gobject-2.0=/nix/store/8j6yqx8qfl0g1nwbshq3502hzv36gai6-glib-2.84.2/lib/gobject-2.0.so"
>   "--config-setting:pango-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pango-1.0.so"
>   "--config-setting:pangocairo-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pangocairo-1.0.so"
>   "--config-setting:xcb-cursor=/nix/store/p3wm9m7w0bak9y3r3kyy28l5klcw0ngh-xcb-util-cursor-0.1.5/lib/xcb-cursor.so"
>   "--config-setting:libdrm=/nix/store/sqlyy8hdm45y2y37g6qwzh8n9j8a7wbd-libdrm-2.4.124-dev/include/libdrm"
>   "--config-setting:pixman-1=/nix/store/clyjdzxjmdvsk2bs7503i2qdvi9axcd8-pixman-0.46.0/include/pixman-1"
> ]
> ```

This is better, but why don't we use env variables?

Gurjaka:
> Regardless of the abstraction approach, what need to be rewritten is the following part:
> 
> ```nix
>   postPatch = ''
>     substituteInPlace libqtile/pangocffi.py \
>       --replace-fail libgobject-2.0.so.0 ${glib.out}/lib/libgobject-2.0.so.0 \
>       --replace-fail libpangocairo-1.0.so.0 ${pango.out}/lib/libpangocairo-1.0.so.0 \
>       --replace-fail libpango-1.0.so.0 ${pango.out}/lib/libpango-1.0.so.0
>     substituteInPlace libqtile/backend/x11/xcursors.py \
>       --replace-fail libxcb-cursor.so.0 ${xcbutilcursor.out}/lib/libxcb-cursor.so.0
>     substituteInPlace libqtile/backend/wayland/cffi/build.py \
>         --replace-fail /usr/include/pixman-1 ${lib.getDev pixman}/include \
>         --replace-fail /usr/include/libdrm ${lib.getDev libdrm}/include/libdrm
>   '';
> ```
> 
> Here is a rough implementation of solution 4 on the nix side could look like:
> 
> ```nix
> let
> # exposed:
>   runtime-libs = {
>     "gobject-2.0" = pkgs.glib;
>     "pango-1.0" = pkgs.pango;
>     "pangocairo-1.0" = pkgs.pango;
>     "xcb-cursor" = pkgs.xcb-util-cursor;
>   };
> 
>   headers = {
>     "pixman-1" = pkgs.pixman;
>     "libdrm" = pkgs.libdrm;
>   };
> 
> # internal logic:
>   inherit (lib.strings) concatStringsSep;
>   inherit (lib.attrsets) mapAttrs attrValues;
> 
>   mk-flags = take: map': attrs: (attrValues
>     (mapAttrs
>       (name: value: "--config-setting:${name}=${map' name (take value)}")
>       attrs));
> 
> in (
>   (mk-flags lib.getLib (name: value: value + "/lib/${name}.so") runtime-libs)
>   ++ (mk-flags lib.getDev (name: value: value + "/include/${name}") headers)
> )
> ```
> 
> outputting the following:
> 
> ```
> [
>   "--config-setting:gobject-2.0=/nix/store/8j6yqx8qfl0g1nwbshq3502hzv36gai6-glib-2.84.2/lib/gobject-2.0.so"
>   "--config-setting:pango-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pango-1.0.so"
>   "--config-setting:pangocairo-1.0=/nix/store/0gl1az5qg24bs9wa9yb61fcva2n83ppa-pango-1.56.3/lib/pangocairo-1.0.so"
>   "--config-setting:xcb-cursor=/nix/store/p3wm9m7w0bak9y3r3kyy28l5klcw0ngh-xcb-util-cursor-0.1.5/lib/xcb-cursor.so"
>   "--config-setting:libdrm=/nix/store/sqlyy8hdm45y2y37g6qwzh8n9j8a7wbd-libdrm-2.4.124-dev/include/libdrm"
>   "--config-setting:pixman-1=/nix/store/clyjdzxjmdvsk2bs7503i2qdvi9axcd8-pixman-0.46.0/include/pixman-1"
> ]
> ```

This is really interesting approach, I would upvote this :)

tych0:
> This is better, but why don't we use env variables?

for what? as arguments to builder? I'd prefer actual arguments to the install command, but that's just me.

if you're talking about at runtime, that means that all Nix users have to set them before they exec qtile, which is also annoying.

> Here is a rough implementation of solution 4 on the nix side could look like:

And here is a rough implementation of the qtile side: https://github.com/qtile/qtile/commit/d03bdc24e6127e0f62c0ba3511bca8e83e368f3e

I might re-work the build_config so that the module itself is optional, since i run a lot of stuff directly from the tree. But it seems promising...

Gurjaka:
> > This is better, but why don't we use env variables?
> 
> for what? as arguments to builder? I'd prefer actual arguments to the install command, but that's just me.
> 
> if you're talking about at runtime, that means that all Nix users have to set them before they exec qtile, which is also annoying.
> 
> > Here is a rough implementation of solution 4 on the nix side could look like:
> 
> And here is a rough implementation of the qtile side: [d03bdc2](https://github.com/qtile/qtile/commit/d03bdc24e6127e0f62c0ba3511bca8e83e368f3e)
> 
> I might re-work the build_config so that the module itself is optional, since i run a lot of stuff directly from the tree. But it seems promising...

We could set environment variables during build. So the Qtile package would result with them built in. Though I am not really sure if that'll work as we expect it to.

tych0:
> We could set environment variables during build. So the Qtile package would result with them built in. Though I am not really sure if that'll work as we expect it to.

Yeah, I think that will work with my branch with just:
```
diff --git a/builder.py b/builder.py
index fc25570e..daa9fef1 100644
--- a/builder.py
+++ b/builder.py
@@ -57,6 +57,6 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
         f.write("# This file is generated at build time by builder.py\n")
         for lib in ["PANGO", "PANGOCAIRO", "GOBJECT", "XCBCURSOR"]:
             p = lib + "_PATH"
-            f.write(f"{p} = {config_settings.get(p)}\n")
+            f.write(f"{p} = {os.environ.get(p)}\n")

     return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)
```
but generally I prefer flags to environment variables, since it's more obvious what's getting used by what command. It looks like it's not too hard to do the flags viz. Sigmanificient's patch above, so I'd prefer that if possible.

Gurjaka:
> > We could set environment variables during build. So the Qtile package would result with them built in. Though I am not really sure if that'll work as we expect it to.
> 
> Yeah, I think that will work with my branch with just:
> 
> ```
> diff --git a/builder.py b/builder.py
> index fc25570e..daa9fef1 100644
> --- a/builder.py
> +++ b/builder.py
> @@ -57,6 +57,6 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
>          f.write("# This file is generated at build time by builder.py\n")
>          for lib in ["PANGO", "PANGOCAIRO", "GOBJECT", "XCBCURSOR"]:
>              p = lib + "_PATH"
> -            f.write(f"{p} = {config_settings.get(p)}\n")
> +            f.write(f"{p} = {os.environ.get(p)}\n")
> 
>      return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)
> ```
> 
> but generally I prefer flags to environment variables, since it's more obvious what's getting used by what command. It looks like it's not too hard to do the flags viz. Sigmanificient's patch above, so I'd prefer that if possible.

Yup sure, I will message him now

tych0:
> Lastly (lets think long term!) we need to consider that there is another patch from nixpkgs

Thanks, it seems like maybe we should just take that upstream? I think we were trying to defend against people running qtile via shebang vs. the interpreter directly, but it seems like that is probably unnecessary. That code is quite old, 9ed0da283348ed67ec27e9475dfe2754196b4c17 is where it was first introduced, and I suspect using the `argv[0]` directly as nix does is just fine these days.

Sigmanificient:
Having the `fix-restart` patch solved too would shave off even more friction from the build, so it can be only a good thing! If we manage to get a patch-less derivation in the end, it would be quite solid.
For the env based appraoch, this is possible by using `makeWrapper`:
```nix
{
  nativeBuildInputs = [ makeWrapper ];
  postInstall = ''
    makeWrapper $out/bin/qtile \
      --prefix PATH "..." \
      --set SOME_ENV_VAR "..."
  '';
```
However, i tend to discourage this when there is a cleaner solution that avoid adding more shell "glue". Here we are also in a situation where qtile is wrapped at the option level (when using `services.xserer.window-manager.qtile`) so i don't inow if adding an extra one would play nicely.

We could still provide env variable at build time tho using:
```nix
{
  env = {
    SOME_SETTINGS = "...";
    # ...
  };
}
```
but i think the flag based approach is preferable

tych0:
> Having the fix-restart patch solved too would shave off even more friction from the build, so it can be only a good thing!

Do you want to send it? Otherwise I can.

tych0:
Ok, check out the above PR. If it looks good, I will land it and set the nix flake to ignore temporarily while the nix side lands.

Gurjaka:
> Ok, check out the above PR. If it looks good, I will land it and set the nix flake to ignore temporarily while the nix side lands.

LGTM, but let's also wait for @Sigmanificient .

Also I assume tests supposed to fail?

tych0:
The flake is expected to fail, the test failure is just us overzealously monkey patching the stdlib. I will fix it before I land.

Sigmanificient:
@tych0 We implemented the nix override with flags based on your branch! Should we:
- re-target the pr to yours?
- rebase it after you merge your pr?

There is also a few python change (ruff check, remove a `script/ffibuild` mention, and a small fix in the builder). While working on this, we took a bit of time to refactor the `overlay.nix` file, because it was getting quite cursed (also i think i improved a lot in nix in the mean time).

tych0:
> rebase it after you merge your pr?

yeah, I think this is probably best. I guess we also want #5425 as well, so you guys can drop the other patch.

I fixed the CI for mine, so assuming you guys are happy with it, I think we can go ahead and land it.

Gurjaka:
I had some time to myself and decided to refactor `flake.nix.` This version should be cleaner, more readable, and easier to maintain. Sigma and I discussed refactoring yesterday, it wasn’t planned for this PR, but I figured since we’re already making significant changes, we might as well do it now. I think it’s an improvement over what we had, but I’m open to feedback and happy to discuss further.

@jwijenbergh @Sigmanificient @tych0

tych0:
Ok, I merged the other two PRs and rebased this one... let's see how CI goes :)

tych0:
And it passed! Thanks for everyone's help!

Sigmanificient:
Well if I knew that gurjaka would have removed the overlay in the end, I might have done a even more eager refactor!

Gurjaka:
> Well if I knew that gurjaka would have removed the overlay in the end, I might have done a even more eager refactor!

PRs welcome haha

tych0:
Seeing these cherry picks makes me wonder: what's the plan for getting `wayc` upstream to master? Is there going to be one big rebase? (is someone rebasing it now?) are we going to merge and do one big conflict resolution? I would prefer option (1) if possible, just to keep changes together.

Gurjaka:
> Seeing these cherry picks makes me wonder: what's the plan for getting `wayc` upstream to master? Is there going to be one big rebase? (is someone rebasing it now?) are we going to merge and do one big conflict resolution? I would prefer option (1) if possible, just to keep changes together.

I have rebased locally, haven't made a pr though

tych0:
I think a rebase probably needs a force push to `wayc`; I don't think you can do a hard reset via the github PR UI?

elParaguayo:
Agreed. I would think we should be rebasing `wayc` to keep it up to date with `master` but agree that we'd need to force push to that branch which isn't ideal given number of people working on it but that still seems preferable to the alternative.

Gurjaka:
> Agreed. I would think we should be rebasing `wayc` to keep it up to date with `master` but agree that we'd need to force push to that branch which isn't ideal given number of people working on it but that still seems preferable to the alternative.

I can push the branch to my fork, i resolved conflicts as well, maybe you can merge that?

jwijenbergh:
> Agreed. I would think we should be rebasing `wayc` to keep it up to date with `master` but agree that we'd need to force push to that branch which isn't ideal given number of people working on it but that still seems preferable to the alternative.

this is totally fine IMO and the only option, I will force push with lease

Sigmanificient:
I wasn't sure of the strategy that would be used, so what why i cherry-picked the commit into a draft pr

tych0:
yeah, makes sense. just wanted to be sure it was on someone's radar :)

PR #5422, opened 2025-08-08
Build config libs

tych0:
This is an alternative to #5420, allowing Nix to pass `--config-settings=PANGO_PATH=...` instead of relying on patching.

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -45,6 +45,9 @@ def get_requires_for_build_wheel(config_settings=None):
 
 def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     """Stop building if wayland requested but pywlroots is not installed."""
+    if config_settings is None:
+        config_settings = {}
+
     if wants_wayland(config_settings):
         try:
             import wlroots  # noqa: F401
@@ -57,6 +60,6 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
         f.write("# This file is generated at build time by builder.py\n")
         for lib in ["PANGO", "PANGOCAIRO", "GOBJECT", "XCBCURSOR"]:
             p = lib + "_PATH"
-            f.write(f"{p} = {config_settings.get(p)}\n")
+            f.write(f"{p} = {config_settings.get(p)!r}\n")
 
     return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)


elParaguayo:
Nice.

What do the changes to `Moc` have to do with this?

tych0:
Without it, xcursors won't load because `ctypes.util.find_library()` wants to `subprocess.Popen()` a call to ldconfig to figure out where the library is. I put the exact stack trace in the commit message.

But looking at the xcursors code closer, I don't think anything actually uses it, so probably we can just delete it entirely...

tych0:
Ah, no, just kidding, it's used here: https://github.com/tych0/qtile/blob/e4c328a7ec49f7458283662053b0795d189aefb4/libqtile/backend/x11/window.py#L302-L305

Let me just fix the test.

Sigmanificient:
Sorry, I have no clue how to use it, am i missing something :sob:

tych0:
> Sorry, I have no clue how to use it, am i missing something 😭

ha, seems like you got it figured out in #5420?

test/widgets/test_moc.py
```diff
@@ -25,7 +25,7 @@
 import pytest
 
 import libqtile.config
-from libqtile.widget import moc
+from libqtile.widget import base, moc
```
Sigmanificient, line 28: There is a unused import here which cause `ruff` to fail, see: https://github.com/qtile/qtile/pull/5420/commits/b1fa16710b49d8f6e018cd8ec0bd5f70debc00c9

builder.py
```diff
@@ -51,4 +51,12 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
         except ImportError:
             sys.exit("Wayland backend requested but pywlroots is not installed.")
 
+    # Write library paths to file, if they are specified at build time via
+    # --config-settings=PANGO_PATH=...
+    with open("libqtile/build_config.py", "w") as f:
+        f.write("# This file is generated at build time by builder.py\n")
+        for lib in ["PANGO", "PANGOCAIRO", "GOBJECT", "XCBCURSOR"]:
+            p = lib + "_PATH"
+            f.write(f"{p} = {config_settings.get(p)}\n")
```
Sigmanificient: Here the config path need to be a string, see https://github.com/qtile/qtile/pull/5420/commits/53b4e84c4b285b6884f147d7274dc8d79025f146

Sigmanificient (changes requested):
Yup, we finally figured it out, it did took some time to understand it :sweat_smile: 

While doing the nix change we had to do a few change on the python side, I think you may want to chery-pick them:
- [Remove deleted script/ffibuild from MANIFEST.in](https://github.com/qtile/qtile/pull/5420/commits/7db570fe224d4ed74599343aeec1d02a1fec0bce)
- [Quote generated variable in the build_config.py file](https://github.com/qtile/qtile/pull/5420/commits/53b4e84c4b285b6884f147d7274dc8d79025f146)
- [Remove unused import (ruff fix)](https://github.com/qtile/qtile/pull/5420/commits/b1fa16710b49d8f6e018cd8ec0bd5f70debc00c9)

builder.py
tych0: Yep, thank you! I cherry picked your patch to this branch.

test/widgets/test_moc.py
tych0: It's not any more, it was a bug in my fix for the tests. It should be used now, so I think you can drop that patch from your branch.

tych0:
> [Remove deleted script/ffibuild from MANIFEST.in](https://github.com/qtile/qtile/pull/5420/commits/7db570fe224d4ed74599343aeec1d02a1fec0bce)

This one looks good too, thank you! If you want to send it separately I'm happy to merge, or you can just keep it in the patch series for the nix flake reorg.

Sigmanificient (approved):
Great!

tych0:
Hmm. I think I saw this previously as well... I wonder if we have a flaky test all of the sudden because of some recent change?

```
FAILED test/test_dgroups.py::test_dgroup_spawn_in_group[1-wayland-DGroupsSpawnConfig] - assert 0 > 0
 +  where 0 = len([])
 +    where [] = <libqtile.command.client.InteractiveCommandClient object at 0x7fe31ef65a60>()
 +      where <libqtile.command.client.InteractiveCommandClient object at 0x7fe31ef65a60> = <libqtile.command.client.InteractiveCommandClient object at 0x7fe31ef7ce60>.windows
 +        where <libqtile.command.client.InteractiveCommandClient object at 0x7fe31ef7ce60> = <test.helpers.TestManager object at 0x7fe31c310f20>.c
====== 1 failed, 1281 passed, 59 skipped, 2 xpassed in 1037.49s (0:17:17) ======
```

tych0:
Ok, seems like we might have a test flake here. I'll keep an eye out, but in the meantime I want to rebase + merge this so we can get the nix flake PR landed.

builder.py
```diff
@@ -45,10 +45,21 @@ def get_requires_for_build_wheel(config_settings=None):
...
+
     if wants_wayland(config_settings):
         try:
             import wlroots  # noqa: F401
         except ImportError:
             sys.exit("Wayland backend requested but pywlroots is not installed.")
 
+    # Write library paths to file, if they are specified at build time via
+    # --config-settings=PANGO_PATH=...
+    with open("libqtile/build_config.py", "w") as f:
```
RFCreate, line 59: This filename is not the same as in `.gitignore`.

tych0: Derp, thank you. Can you send a patch to fix it?

PR #5423, opened 2025-08-08
CI: add libxcb-cursor0

tych0:
we have apparently never installed this library in CI, and thus never actually tested the xcursors code there. Let's add this library so we can test it :)


elParaguayo:
Do we already have tests or do we need to add those too?

tych0:
> Do we already have tests or do we need to add those too?

That's a good question; the code is not dead in the sense that it is called here: https://github.com/tych0/qtile/blob/a4553c20fbfd9e3726255ced375e243fe475708d/libqtile/backend/x11/core.py#L151-L152 assuming the library can be loaded, but we don't have any tests to make sure that call actually works. Probably we should.

tych0:
I'm going to land this for now since having it and running the code is better than not having it, but it seems like we should re-evaluate this, or even just use xcffib directly instead?

PR #5424, opened 2025-08-08
drop bin/qtile + setup.py

tych0:
With our pyproject.toml migration, we are using libqtile.scripts.main:main directly as the entrypoint, not this binary any more.

The only thing that looks important is the lifecycle import, which seems to happen early enough anyway (by virtue of termination still working correctly in this new model), so it's safe to drop.


tych0:
Oh, ha, the tests use it! I will fix.

PR #5425, opened 2025-08-09
core: use argv[0] when re-execing

tych0:
Rather than wrapping ourselves in sys.executable, let's just re-exec whatever was argv[0], and let the shebang or whatever be resolved again.


tych0:
This intentionally breaks the nix flake: they are applying this same patch, but it's something we can carry upstream, so let's do that.

Sigmanificient (approved):
No problem with the breaking the flake for now, we can fix that via #5420

PR #5426, opened 2025-08-09
[wayc] build via pyrpoject.toml

elParaguayo:
`setup.py` has been removed from main repo so, instead, we can incorporate the ffi build into a custom builder and put it behind a config setting.


elParaguayo:
Do we really need to build the wayland backend to build our docs?

builder.py
```diff
@@ -26,37 +17,33 @@
...
+def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
+    """If wayland backend is requested, build it!"""
+    if wants_wayland(config_settings):
+        try:
+            from libqtile.backend.wayland.cffi.build import ffi
 
-    return cffi_modules
+            ffi.compile(verbose=True)
+        except (distutils.errors.DistutilsError, subprocess.CalledProcessError) as e:
+            sys.exit(f"Wayland backend requested but backend could not be built: {e}")
```
elParaguayo, line 47: Alternatively, we could always try to build the backend but only stop qtile building if the user has explicitly asked for wayland and the backend build fails.

Gurjaka:
Does this mean we can remove `./scripts/ffibuild` from wayc as well?

elParaguayo:
> Does this mean we can remove `./scripts/ffibuild` from wayc as well?

I don't think so. We'll still want to run that command locally. The builder works for people installing with a package manager.

jwijenbergh:
didn't test, but looked go to me

builder.py
```diff
@@ -26,37 +17,33 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-import importlib
+import distutils.errors
```
Sigmanificient, line 20: I dont know why, but importing `distutils` break the build in nix :o

PR #5428, opened 2025-08-10
fix: match gitignore libqtile/_build_config.py filename

RFCreate:
Fixes #5422

PR #5429, opened 2025-08-10
fix(mypy): _build_config.py causes unused 'type: ignore'

RFCreate:
Addresses https://github.com/qtile/qtile/pull/5428#issuecomment-3172808030

More than a proposed change, it is more to give visibility to a problem with mypy that I explain in the above comment.


libqtile/__init__.py
```diff
@@ -58,13 +59,15 @@ def find_library(key: DynamicLibraries) -> str | None:
...
-        attr = key.name + "_PATH"
-        value = getattr(_build_config, attr, None)
-        if value is not None:
-            return value
-    except ImportError:
+        with open(f"{Path(__file__).parent}/_build_config.py") as f:
+            attr = key.name + "_PATH"
+            for line in f:
+                if line.startswith(attr):
+                    value = line.split("=")[1].strip().strip("'\"")
```
tych0: Thanks for pointing this out. I'd really like to keep this as a python import vs. ad-hoc parsing of python code. I assume you can't just drop the `type: ignore` here or something else typechecks incorrectly, so maybe we need to wrap things in isinstance(..., str) to fix the type bounds?

RFCreate: I also prefer the import, another option is this change for precommit:

```diff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       - id: mypy
         additional_dependencies: ["types-python-dateutil", "types-pytz"]
         files: "^libqtile\/.*"
+        exclude: "^libqtile\/__init__.py"
   - repo: https://github.com/jendrikseipp/vulture
     rev: v2.14
     hooks:
```

tych0: I think we should be able to get it to type check by doing some guards, right? That way we can still get other stuff in the file to type check.

RFCreate: Then maybe in the `pyproyect.toml` add an override to exclude the "import" check.

RFCreate: Like this: (i also tested adding the ignore comments in other files and it got trigger to no recursion i hope)

```toml
[[tool.mypy.overrides]]
module = "libqtile"
warn_unused_ignores = false
```

tych0: Hmm, can't we just delete these `type: ignore`? https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores suggests that it is really that mypy has gotten better, and that we don't need these any more. If I upgrade to mypy 1.17.1 in my tree and delete the ones it complains about, things type check.

RFCreate: Try forcing mypy to recheck the file by deleting the cache.

```
> rm libqtile/_build_config.py
> rm -rf .mypy_cache/
> pre-commit run mypy --files libqtile/__init__.py
mypy.....................................................................Passed
> touch libqtile/_build_config.py
> pre-commit run mypy --files libqtile/__init__.py
mypy.....................................................................Passed
> rm -rf .mypy_cache/
> pre-commit run mypy --files libqtile/__init__.py
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

libqtile/__init__.py:61: error: Unused "type: ignore" comment  [unused-ignore]
Found 1 error in 1 file (checked 1 source file)
```

RFCreate: As for removing `type: ignore`, this is the printed error:

```
> rm libqtile/_build_config.py
> rm -rf .mypy_cache/
> pre-commit run mypy --files libqtile/__init__.py
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

libqtile/__init__.py:61: error: Module "libqtile" has no attribute "_build_config"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

tych0: Oh, I think I finally understand, thanks for bearing with me. There are two cases: if the file is present, we get the error about `Unused "type: ignore"` with newer mypys. But by default the file is *not* present, so we need the `type: ignore` or mypy will complain.

So I wonder: should we just drop the `type: ignore` and commit an empty file to the tree? That way it's always there for local development, and we will overwrite it during wheel build with whatever options the user actually specified.

RFCreate: The only problem is that once the file is tracked, even if it is in `.gitginore`, git will track all modifications, I will write a big comment in the file and you tell me what you think.

tych0: Yep, looks reasonable to me. Thanks for bearing with me.

PR #5433, opened 2025-08-12
[wayc] builder update

elParaguayo:
Always tries to build wayland but install only fails if wayland backend explicitly requested.

Changes exception handling as ffi builds trigger different exceptions on different systems.


Sigmanificient (approved):
Well now we are so to catch anything for sure!

PR #5435, opened 2025-08-12
[wayc] Add wallpaper support for screens

elParaguayo:


jwijenbergh:
doesn't build with nix, please add

```diff
diff --git a/libqtile/backend/wayland/qw/output.c b/libqtile/backend/wayland/qw/output.c
index a5b93a81..a562aeea 100644
--- a/libqtile/backend/wayland/qw/output.c
+++ b/libqtile/backend/wayland/qw/output.c
@@ -1,3 +1,4 @@
+#include "cairo-buffer.h"
 #include "output.h"
 #include "layer-view.h"
 #include "server.h"
```

elParaguayo:
Sure. However, can you explain why it works locally but not on nix?

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
         wlr_output_layout_add_auto(server->output_layout, wlr_output);
     wlr_scene_output_layout_add_output(server->scene_layout, l_output, output->scene);
 }
+
+void qw_output_background_destroy(struct qw_output *output) {
+    // Destroy the color scene_rect if it exists
+    struct wlr_scene_rect *color_rect = output->background.color_rect;
+    if (color_rect) {
```
jwijenbergh: in discord we decided it might be better to do more explicit NULL checks

!= NULL

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+        wlr_scene_node_destroy(&color_rect->node);
+        output->background.color_rect = NULL;
+    }
+
+    // Destroy wallpaper resources if wallpaper exists
+    if (output->background.wallpaper) {
```
jwijenbergh: != NULL

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+        struct wlr_scene_buffer *buffer = output->background.wallpaper->buffer;
+        if (buffer) {
+            wlr_scene_node_destroy(&buffer->node);
+            output->background.wallpaper->buffer = NULL;
+        }
+
+        cairo_surface_t *surface = output->background.wallpaper->surface;
+        if (surface) {
```
jwijenbergh: != NULL

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+    cairo_set_source_surface(cr, source, 0, 0);
+    cairo_paint(cr);
+
+    // Get data and stride from scaled surface
+    unsigned char *data = cairo_image_surface_get_data(wallpaper_surface);
+    int stride = cairo_image_surface_get_stride(wallpaper_surface);
+
+    // Create wlr_buffer from scaled surface data
+    struct wlr_buffer *buffer = cairo_buffer_create(o_width, o_height, stride, data);
+    if (!buffer) {
```
jwijenbergh: == NULL

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+        wlr_log(WLR_ERROR, "Failed to create wlr_buffer from scaled surface");
+        cairo_surface_destroy(wallpaper_surface);
+        return;
+    }
+
+    struct wlr_scene_buffer *scene_buf =
+        wlr_scene_buffer_create(output->server->scene_wallpaper_tree, buffer);
+
+    if (!output->background.wallpaper) {
```
jwijenbergh: == NULL

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +156,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+    qw_output_background_destroy(output);
+
+    // Get screen dimensions
+    int o_width, o_height;
+    wlr_output_effective_resolution(output->wlr_output, &o_width, &o_height);
+
+    // Create the scene rect
+    struct wlr_scene_rect *rect =
+        wlr_scene_rect_create(output->server->scene_wallpaper_tree, o_width, o_height, color);
+    if (!rect) {
```
jwijenbergh: == NULL

libqtile/backend/wayland/qw/server.c
```diff
@@ -535,3 +538,34 @@ static void qw_server_query_iterator(struct wlr_scene_buffer *buffer, int sx, in
...
+            return o;
+        }
+    };
+    wlr_log(WLR_ERROR, "Cannot find output at %d,%d.", x, y);
+    return NULL;
+}
+
+void qw_server_paint_wallpaper(struct qw_server *server, int x, int y, cairo_surface_t *source,
+                               int mode) {
+    struct qw_output *o = qw_server_get_output_at_pos(server, x, y);
```
jwijenbergh: can't we use? https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.19/include/wlr/types/wlr_output_layout.h?ref_type=heads#L80

jwijenbergh: you can then use wlr_output->data to get the qw_output

jwijenbergh:
> Sure. However, can you explain why it works locally but not on nix?

I can't explain the other way around, why it works locally. The compiler has to know where `cairo_buffer_create` is

jwijenbergh:
some nits but overall looks really good, awesome work!

elParaguayo:
Have made all those changes.

libqtile/backend/wayland/qw/output.h
```diff
@@ -1,13 +1,25 @@
...
 struct qw_server;
 
+struct qw_output_background_wallpaper {
+    struct wlr_scene_buffer *buffer;
+    cairo_surface_t *surface;
+};
+
+struct qw_output_background {
+    struct wlr_scene_rect *color_rect;
+    struct qw_output_background_wallpaper *wallpaper;
```
Sigmanificient: Wouldn't it be better for this to be a discriminated union? Can you have both `color_rect` an `wallpaper`?

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +157,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+            output->background.wallpaper->surface = NULL;
+        }
+
+        free(output->background.wallpaper);
+        output->background.wallpaper = NULL;
+    }
+}
+
+
+void qw_output_paint_wallpaper(struct qw_output *output, cairo_surface_t *source, int mode) {
```
Sigmanificient: I think it would be better to type the modes with a `enum` rather than a `int`, so that the switch ensure a 1:1 mapping (`-Wswitch-enum`)

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +157,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+    wlr_output_effective_resolution(output->wlr_output, &o_width, &o_height);
+
+    // Calculate x and y scaling factors
+    double scale_x = (double)o_width / img_width;
+    double scale_y = (double)o_height / img_height;
+
+    // Create a new surface sized to output size
+    cairo_surface_t *wallpaper_surface =
+        cairo_image_surface_create(CAIRO_FORMAT_ARGB32, o_width, o_height);
+    cairo_t *cr = cairo_create(wallpaper_surface);
```
Sigmanificient: I think we should check for this call to fail, either with xwrapper or skip

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +157,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+    // Scale image to screen - don't preserve aspect ratio
+    case WALLPAPER_MODE_STRETCH:
+        cairo_scale(cr, scale_x, scale_y);
+        break;
+
+    // Center image on screen but don't scale image
+    case WALLPAPER_MODE_CENTER:
+        int t_x = (o_width - img_width) / 2;
+        int t_y = (o_height - img_height) / 2;
+        cairo_translate(cr, t_x, t_y);
```
Sigmanificient, line 262: Does this work if the wallpaper is bigger than the surface?

libqtile/backend/wayland/qw/output.c
```diff
@@ -156,3 +157,159 @@ void qw_server_output_new(struct qw_server *server, struct wlr_output *wlr_outpu
...
+        wlr_log(WLR_ERROR, "Failed to create wlr_buffer from scaled surface");
+        cairo_surface_destroy(wallpaper_surface);
+        return;
+    }
+
+    struct wlr_scene_buffer *scene_buf =
+        wlr_scene_buffer_create(output->server->scene_wallpaper_tree, buffer);
+
+    if (output->background.wallpaper == NULL) {
+        output->background.wallpaper = calloc(1, sizeof(struct qw_output_background_wallpaper));
```
Sigmanificient, line 292: Question for coding style:

I tend to prefer the following to avoid binding the type twice, what are you though on this?
```suggestion
        output->background.wallpaper = calloc(1, sizeof *output->background.wallpaper);
```

libqtile/backend/wayland/qw/output.h
elParaguayo: Originally, I had thought we could have both (e.g. if you had a small image which you wanted centred over a solid background). However, the core qtile logic (currently) draws one or the other. So, yes, a union is ok.

libqtile/backend/wayland/qw/output.c
elParaguayo: I don't know enough about c but that is how I'd typed it originally and it failed to build.

Sigmanificient: Oh the issue is that you made a anonymous enum, you need to give it a name to be able to reference it as a type:
```c
enum xxx { ...; };

void my_func(enum xxx mode);
```
or
```c
typedef enum { ...; } xxx;

void my_func(xxx mode);
```

libqtile/backend/wayland/qw/output.c
elParaguayo: Yes. It's still centred correctly but the wallpaper is shifted offscreen. (which reminds me, we probably need to apply clipping here too).

libqtile/backend/wayland/qw/output.c
elParaguayo: I'll defer to the more experience c coders but, to me, the first one makes it clear what type of object is expected there.

libqtile/backend/wayland/qw/output.c
Gurjaka: @Sigmanificient what do you think of using uint_16 instead here? I don't think any monitor can go higher than 65,535

libqtile/backend/wayland/qw/output.c
jwijenbergh: I actually prefer the explicit struct type in the size of, the size is on the type, not the instance

Sigmanificient:
I am missing the header when building with nix, but also when building manually :eyes: 
image

jwijenbergh:
Needs a rebase

elParaguayo:
> Needs a rebase

Done

jwijenbergh:
merci

Issue #5437, opened 2025-08-12
[wayc] build the backend without so many hacks


jwijenbergh:
### Issue description

currently:

- source files are concatenated
- header files are copied to cdef without the `private data` stuff

do this properly...

https://github.com/qtile/qtile/blob/wayc/libqtile/backend/wayland/cffi/build.py

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #5443, opened 2025-08-13
[wayc]  Move toward a python extension

Sigmanificient:
- Fix: #5437

> [!NOTE]
> The current cffi library as multiple issues, one of which is the fragile build method it impose. Declaration are really not straight forward, the `CDEF` string is a huge hack, and we dont really have that much control on the way it is built.
>
> This pr aims to introduce a python extension which allow to keep our work in progress C wayland backend (and continue iteration on it), while fixing the way it binds to python under the hood.

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -132,17 +132,18 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     from setuptools.command.build_ext import build_ext
 
     generate_build_config(config_settings or {})
-    build_cff_lib(config_settings.get("backend") == "wayland")
 
-    dist = Distribution()
-    dist.ext_modules = prepare_extensions()
+    wayland_needed = config_settings.get("backend") == "wayland"
+    build_cff_lib(wayland_needed)
 
-    print("Building Wayland backend extension")
+    if wayland_needed:
+        dist = Distribution()
+        dist.ext_modules = prepare_extensions()
 
-    cmd = build_ext(dist)
-    cmd.build_lib = WAYLAND_DIR
-    cmd.ensure_finalized()
-    cmd.run()
+        cmd = build_ext(dist)
+        cmd.build_lib = WAYLAND_DIR
+        cmd.ensure_finalized()
+        cmd.run()
 
     return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)
 
--- a/libqtile/backend/wayland/__init__.py
+++ b/libqtile/backend/wayland/__init__.py
@@ -30,5 +30,14 @@ try:
 
 except Exception as e:
     print("Fail to import Wayland Backend:", e)
+    from types import ModuleType
+    from typing import Any
+
+    class WaylandBackendStub(ModuleType):
+        def __getattr__(self, name: str) -> Any:
+            print("[wayland_backend stub] Called '%s' (no-op)", name)
+
+    wayland_backend = WaylandBackendStub("wayland_backend")
+
 else:
     wayland_backend.hello()
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -89,7 +89,7 @@ except ModuleNotFoundError:
 
     from libqtile.backend.wayland.ffi_stub import ffi, lib
 
-from libqtile.backend.wayland import wayland_backend
+from . import wayland_backend
 
 if TYPE_CHECKING:
     from libqtile import config


elParaguayo:
Do we need to use `setup.py`? We had been trying to move away from it and use the more modern building via `pyproject.toml` (plus the changes I'd included in `builder.py`).

To be honest, I don't understand the issue this PR is trying to solve but that's not a criticism of the PR, just a comment on my lack of understanding about c!

jwijenbergh:
> Do we need to use `setup.py`? We had been trying to move away from it and use the more modern building via `pyproject.toml` (plus the changes I'd included in `builder.py`).
> 
> To be honest, I don't understand the issue this PR is trying to solve but that's not a criticism of the PR, just a comment on my lack of understanding about c!

The building using cffi is currently very hacky. We have to concatenate all source and header files and also have to exclude certain data in the build (see the private data comments). On top of that, building currently is causing issues across environments. E.g. it builds locally, not on nix, or vice versa. Then also, there is currently a vague split between what is exposed in python and what is exposed for c land. Some of these problems can be fixed with cffi, but the proper way forward seems to be that we should make a proper python extension using `python.h`. This involves more boilerplate code, but seems worth it on the long run.

It would be neat to build the wayland backend like a proper c project, that exposes a python interface. This is what the `python.h`/python extension PR is trying to do

tych0:
I was wondering... can we use rust here instead of C? There is a bunch of work going on to make Rust linkage work a lot better with cpython. 

I agree, cffi seems nice in principle, but has turned out to be somewhat of a headache to distribute.

jwijenbergh:
> I was wondering... can we use rust here instead of C? There is a bunch of work going on to make Rust linkage work a lot better with cpython.
> 
> I agree, cffi seems nice in principle, but has turned out to be somewhat of a headache to distribute.

Sure, we can, but rust gives us more headache than solutions for wayland. See e.g. https://way-cooler.org/blog/2020/01/09/way-cooler-post-mortem.html

There is smithay, for a wlroots alternative, but it's not as production ready as wlroots. If anything, instead of c, I prefer to explore Zig if that hits 1.0

Sigmanificient:
@elParaguayo I did a bit of dark magic and got everything into `build.py`

elParaguayo:
> @elParaguayo I did a bit of dark magic and got everything into `build.py`

That's the kind of ugliness I'd expect in `qtile-extras`!

I'm going to sit out of this one and leave it to the people who know what they're talking about.

Sigmanificient:
> That's the kind of ugliness I'd expect in `qtile-extras`!

I am working on it!

builder.py
```diff
@@ -17,45 +17,115 @@
...
+
+    dist = Distribution()
+    dist.ext_modules = prepare_extensions()
+
+    print("Building Wayland backend extension")
+
+    cmd = build_ext(dist)
+    cmd.build_lib = WAYLAND_DIR
+    cmd.ensure_finalized()
+    cmd.run()
```
elParaguayo: What happens if the build fails? `build_wheel` will be run if people just want the x11 backend so we must allow this function to succeed.

My previous version only raised an error if the wayland backend didn't build when the user explicitly requests the backend via `--config-settings backend=wayland`. I think you need that here unless I'm missing something.

Sigmanificient: Done!

Gurjaka:
> > I was wondering... can we use rust here instead of C? There is a bunch of work going on to make Rust linkage work a lot better with cpython.
> > I agree, cffi seems nice in principle, but has turned out to be somewhat of a headache to distribute.
> 
> Sure, we can, but rust gives us more headache than solutions for wayland. See e.g. https://way-cooler.org/blog/2020/01/09/way-cooler-post-mortem.html
> 
> There is smithay, for a wlroots alternative, but it's not as production ready as wlroots. If anything, instead of c, I prefer to explore Zig if that hits 1.0

Oh yeah, Zig is modern C, that would be really interesting to work on

Sigmanificient:
@Gurjaka Zig still has a unstable API tho, and it would be harder to interface with python

Gurjaka:
> @Gurjaka Zig still has a unstable API tho

Haven't dug in yet, though it looks promising

jwijenbergh:
> > @Gurjaka Zig still has a unstable API tho
> 
> Haven't dug in yet, though it looks promising

maybe in the future, definitely not now, building would be a lot more annoying

Gurjaka:
> > > @Gurjaka Zig still has a unstable API tho
> > 
> > 
> > Haven't dug in yet, though it looks promising
> 
> maybe in the future, definitely not now, building would be a lot more annoying

For now C is good enough, and so much work anyway to refactor haha

PR #5444, opened 2025-08-13
Add window swapping to Bsp and Columns Layouts

MoonCubes:
Implements the  `swap` method for the `Bsp` and the `Columns` Layouts.
This is used by `lazy.window.set_position()`


elParaguayo:
Thanks for this.

Is it possible to add a test to show that these changes work as intended?

MoonCubes:
@elParaguayo Done!

PR #5447, opened 2025-08-14
Always log to stdout

Sigmanificient:


Sigmanificient:
Worth retargeting it `master`?

jwijenbergh:
> Worth retargeting it `master`?

IMO yes

Sigmanificient:
@jwijenbergh done

tych0:
I think this will end up generating two of the same logs on users' systems under x11, one in ~/.local/share/qtile/qtile.log from the python logging package, and one in ~/.xsession-errors from the stdout logging. 

We probably only want one of those, I'm fine with making it only xsession-errors and dropping the other log, but will that cause problems on wayland?

Sigmanificient:
@tych0 I dont have the `~/.xesssion-errors` generated on my config :thinking:

tych0:
Even under x11? How are you starting qtile?

Sigmanificient:
Yes, i'm starting it using `startx`

tych0:
Oh, I suppose that makes sense if you're not using xsession. 

You probably don't get logs to startx's stdout from qtile then? Can you check where qtile's stdout points to? is it /dev/null?

Sigmanificient:
I'm not sure how to check that, but the log are definitely not shown to `stdout`/`stderr` after quitting the session manually

tych0:
If you look at `ls -alh /proc/$(pidof qtile)/fd` stdout will be where 1 points to, stderr will be where 2 points to.

Sigmanificient:
```
dr-x------ 2 sigmanificient users 13 Aug 15 18:56 .
dr-xr-xr-x 9 sigmanificient users  0 Aug 15 16:53 ..
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 0 -> /dev/tty1
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 1 -> /dev/tty1
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 10 -> 'socket:[22884]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 11 -> 'socket:[4866]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 12 -> 'socket:[4867]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 2 -> /dev/tty1
l-wx------ 1 sigmanificient users 64 Aug 15 18:56 3 -> /home/sigmanificient/.local/share/qtile/qtile.log
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 4 -> 'socket:[14220]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 5 -> 'anon_inode:[eventpoll]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 6 -> 'socket:[9810]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 7 -> 'socket:[9811]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 8 -> 'socket:[4864]'
lrwx------ 1 sigmanificient users 64 Aug 15 18:56 9 -> 'socket:[3966]'
```
It does says `/dev/tty1`

tych0:
Yeah, makes sense. So you will not get two writes per log to your disk, since you're using /dev/tty1 for stdout. I assume you get these logs on tty1 then, which is presumably where you ran startx from?

One thing we could do is use `sys.stdout.isatty()` to decide where to put log output. Sometimes I find programs that do this are annoying, but I would find writing the same logs to two places more annoying :)

PR #5457, opened 2025-08-18
[wayc] Deactivate previous surfaces on focus

richcarni:
xwayland_view requires activation of a newly focused surface and deactivation of the previously focused service. The absence of which was causing issues with xwayland views reliably receiving keyboard input

Since the previous surface could be either an xdg_toplevel or xwayland_surface (or potentially other in future), deactivate_surface needs to work for all these different surfaces

xdg_view requires the same surface deactivation of the previously focused surface

### DIFF
--- a/libqtile/backend/wayland/qw/xwayland-view.c
+++ b/libqtile/backend/wayland/qw/xwayland-view.c
@@ -28,7 +28,7 @@ static void qw_xwayland_view_do_focus(struct qw_xwayland_view *xwayland_view,
     wlr_scene_node_raise_to_top(&xwayland_view->base.content_tree->node);
 
     //Deactivate previous surface if any
-    if (prev_surface) {
+    if (prev_surface != NULL) {
         qw_util_deactivate_surface(prev_surface);
     }
 


libqtile/backend/wayland/qw/util.c
```diff
@@ -70,3 +70,19 @@ int qw_util_get_modifier_code(const char *codestr) {
 xkb_keysym_t qwu_keysym_from_name(const char *name) {
 	return xkb_keysym_from_name(name, XKB_KEYSYM_CASE_INSENSITIVE);
 }
+
+void qw_util_deactivate_surface(struct wlr_surface *surface) {
```
richcarni, line 74: I wasn't sure where this function belonged. My first though was xdg_view but it doesn't actually operate on one

elParaguayo: Could the base view definition include a `deactivate` function and then we customise this for xdg and xwayland views?

richcarni: The problem is it needs to try all the different surface types to figure out what the surface is. I suppose this could be in a  base view function which then calls a view specific activation function.

But as it stands it doesn't actually operate on views. I thought about having it take a view and get the focused surface from it, but then it's duplicating what we've already done in `do_focus()`

libqtile/backend/wayland/qw/xwayland-view.c
```diff
@@ -26,6 +27,13 @@ static void qw_xwayland_view_do_focus(struct qw_xwayland_view *xwayland_view,
 
     wlr_scene_node_raise_to_top(&xwayland_view->base.content_tree->node);
 
+    //Deactivate previous surface if any
+    if (prev_surface) {
```
jwijenbergh: != NULL

richcarni: If this is our convention, OK. But, for example, `server.c` is littered with this style of check: `if (!new_allocator)`. Going forward should this instead be written `if (new_allocator == NULL)`?

Unless we do a global refactor, isn't this going to result in the code being a mish mash of styles?

elParaguayo: This was the convention that @jwijenbergh has asked us to follow so code should be consistent.

jwijenbergh: Indeed, this is the new convention we have so I still have to convert old code

PR #5460, opened 2025-08-20
Add transient hooks

elParaguayo:
Hooks that return ``True`` will be automatically unsubscribed.

The idea behind this was needing some code to restack windows that had been temporarily elevated to a higher layer without constantly needing to fire hooks.

Draft for now just to get feedback on the idea.

If people like it, I'll add some tests.

### DIFF
--- a/libqtile/hook.py
+++ b/libqtile/hook.py
@@ -51,17 +51,24 @@ def clear():
     subscriptions.clear()
 
 
-def _fire_async_event(co):
+def _fire_async_event(co, unsubscribe):
     from libqtile.utils import create_task
+
+    def finish_task(task):
+        if task.result() is True:
+            unsubscribe()
 
     loop = None
     with contextlib.suppress(RuntimeError):
         loop = asyncio.get_running_loop()
 
     if loop is None:
-        asyncio.run(co)
+        result = asyncio.run(co)
+        if result is True:
+            unsubscribe()
     else:
-        create_task(co)
+        task = create_task(co)
+        task.add_done_callback(finish_task)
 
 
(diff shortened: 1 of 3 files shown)


libqtile/hook.py
```diff
@@ -184,17 +191,33 @@ def fire(self, event, *args, **kwargs):
         # third party libraries will need this to prevent KeyErrors when firing hooks
         if self.name not in subscriptions:
             subscriptions[self.name] = dict()
+
+        # Handlers for transient hooks
+        def unsubscribe_func(event, func):
+            def _wrapper():
+                getattr(self.unsubscribe, event)(func)
```
tych0, line 198: I don't have a problem with the idea.

This code has always annoyed me though, (there's a lot of metaprogramming here for a what is basically a dict from `event name -> list[Callable]`). I think this just adds to it viz the closure, getattr invoking __getattr__ from HookHandlerCollection.

Given that `unsubscribe` ends up being a global variable in this file anyway, do we need to close over anything here? Can we just reference it directly from _fire_async_event?

But better yet, (or perhaps in addition) can we simplify this file somehow?

elParaguayo: Yes. This could probably be rewritten but it's not as high on my list of priorities!

The important point for me is the ability to create addition hook registries. qtile-extras uses this to create hooks for some of its features. Directly referring to `unsubscrive` in `hook.py` would bypass that as it's bound to the qtile hooks.

elParaguayo:
@tych0 I've added some tests for this which are all passing.

I'd like to get this merged if you're ok with it. We can refactor code later.

tych0:
yep, works for me.

PR #5463, opened 2025-08-25
[wayc] Expose wayland scene graph to python

richcarni:
Alternative to PR #5414

Traverse the scene graph tree in wayc with a callback to python to build the tree in a dict

Intended for debugging purposes


libqtile/backend/wayland/core.py
```diff
@@ -631,6 +631,38 @@ def loop(wid):
         lib.qw_server_loop_visible_views(self.qw, loop)
         return wids
 
+    @expose_command()
+    def get_scene_graph(self):
```
elParaguayo: My PR for stacking in X11 has a similar exposed method in the X11 `Core` called `show_stacking_order`. Given these are both doing effectively the same thing, can we find a name that works across both backends? `scene tree` is a bit wayland specific.

richcarni: Definitely! I'll checkout your PR

richcarni:
And I will correct NULL checks... and done

jwijenbergh:
has been manually rebased onto wayc

PR #5466, opened 2025-08-26
[do not merge] Test CI on python 3.14

elParaguayo:


tych0:
Thanks for working on this!

elParaguayo:
I _think_ the issue is that we're passing a `Backend` instance as an argument to `run_qtile` whereas we need to create the backend from within the `multiprocessing` process.

elParaguayo:
@FrostyX 
Ok - so the issue is summarised here: https://docs.python.org/3.14/whatsnew/3.14.html#incompatible-changes

`mulitprocessing`'s default start method was `fork` prior to 3.14 but is `forkserver` for 3.14 onwards. In fork server, every argument passed to the Process needs to be picklable.

In this PR, I've just added `multiprocessing.set_start_method("fork", force=True)` and that seems to work. https://github.com/qtile/qtile/pull/5466/files#diff-67f3b6f6186f6db6e27a630e3377b022c84859b0290734c5402364196e470698R211-R216

It feels like a bit of a hack but might buy us some time if we want to rewrite the tests in a way that's compliant with forkserver.

tych0:
> It feels like a bit of a hack but might buy us some time if we want to rewrite the tests in a way that's compliant with forkserver.

Yeah... I agree it needs some love. The whole framework was really done in 2008, and hasn't been touched much then. I have been wondering if we can't somehow run the tests in the context of the qtile process (maybe make them all asyncio methods?), so we don't have to fork at all. I haven't played around with it, though.

test/helpers.py
```diff
@@ -181,39 +208,34 @@ def get_log_buffer(self):
         return os.read(self.logspipe, 64 * 1024).decode("utf-8")
 
     def start(self, config_class, no_spawn=False, state=None):
+        try:
+            multiprocessing.set_start_method("fork", force=True)
+        except RuntimeError:
+            # It's already been set somewhere else in your program
```
jwijenbergh: What is your program?

elParaguayo: Ha - busted. I was working this through with chatgpt! In fairness, it identified the issue immediately.

jwijenbergh: I thought so ;)

elParaguayo:
Should we tidy this up and merge it? It probably makes sense to have 3.14 in CI but not a requirement for merging. It will help us identify issues earlier on.

tych0:
Yeah, that works for me!

elParaguayo:
OK - this should be good now.

PR #5472, opened 2025-08-27
[wayc] Implement input device configuration (incomplete)

richcarni:
Apply config from wl_input_rules

Need to reload config to take effect

This is a functional but incomplete implementation. Most pointer configuration options haven't been added and there are various todos marked in the code

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -1,0 +1,149 @@
+import enum
+from libqtile import configurable
+from typing import Any
+from libqtile.log_utils import logger
+
+try:
+    from libqtile.backend.wayland._ffi import ffi, lib
+
+except ModuleNotFoundError:
+    print("Warning: Wayland backend not built. Backend will not run.")
+
+    from libqtile.backend.wayland.ffi_stub import ffi, lib
+
+
+class InputConfig(configurable.Configurable):
+    """
+    This is used to configure input devices. An instance of this class represents one
+    set of settings that can be applied to an input device.
+
+    To use this, define a dictionary called ``wl_input_rules`` in your config. The keys
+    are used to match input devices, and the values are instances of this class with the
+    desired settings. For example:
+
+    .. code-block:: python
+
+        from libqtile.backend.wayland import InputConfig
+
+        wl_input_rules = {
+            "1267:12377:ELAN1300:00 04F3:3059 Touchpad": InputConfig(left_handed=True),
+            "*": InputConfig(left_handed=True, pointer_accel=True),
+            "type:keyboard": InputConfig(kb_options="ctrl:nocaps,compose:ralt"),
+        }
+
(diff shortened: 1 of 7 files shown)


libqtile/backend/wayland/qw/server.c
```diff
@@ -640,3 +645,20 @@ void qw_server_paint_background_color(struct qw_server *server, int x, int y, fl
...
 }
+
+void qw_server_loop_input_devices(struct qw_server *server, input_device_cb_t cb) {
+    struct qw_input_device *input_device;
+    wl_list_for_each(input_device, &server->input_devices, link) {
+        struct wlr_input_device *device = input_device->device;
+
+        int vendor = 0;
+        int product = 0;
+        if (wlr_input_device_is_libinput(device) == true) {
```
jwijenbergh: For booleans without == is fine

libqtile/backend/wayland/inputs.py
```diff
@@ -0,0 +1,149 @@
...
+            conf = None
+
+        if conf is not None:
+            if type == InputDeviceType.POINTER:
+                device = lib.qw_input_device_get_libinput_handle(input_device)
+                if device == ffi.NULL:
+                    return
+
+                if conf.accel_profile is not None:
+                    lib.qw_input_device_config_accel_set_profile(device, InputDeviceAccelProfile[conf.accel_profile].value)
```
richcarni: I'd prefer to be only dealing with the libinput device on the c-side, but that would mean having to do this (and associated check) for every config option

Toyed with the idea of packing the config options into a struct. Probably overall a bit more complicated than this approach

libqtile/backend/wayland/qw/input-device.c
```diff
@@ -0,0 +1,72 @@
...
+void qw_input_device_config_kbd_set_repeat_info(struct wlr_keyboard *keyboard, int kb_repeat_rate, int kb_repeat_delay) {
+    wlr_keyboard_set_repeat_info(keyboard, kb_repeat_rate, kb_repeat_delay);
+}
+
+void qw_input_device_config_kbd_set_keymap(struct wlr_keyboard *keyboard,
+                                           const char *layout,
+                                           const char *options,
+                                           const char *variant) {
+    qw_keyboard_set_keymap(keyboard, layout, options, variant);
+}
```
richcarni: As an example, I feel that there could be a cleaner way to integrate input_device and keyboard management

richcarni: Subsequently I have moved keyboard related functions to qw_keyboard and added a reference from input devices (that are keyboards) to their associated qw_keyboard via data pointer

libqtile/backend/wayland/inputs.py
richcarni: I suppose another option would be a function to check and store libinput handle - called before setting configs

jwijenbergh:
has been manually rebased onto wayc

PR #5474, opened 2025-08-29
Added more terminals

tych0:
Added more terminals to search for in test_terminals, 1 exclusively for wayland, and 7 for X11. There is a problem for zutty where the default font is set to 9x18, and it may not have a universal default that will be on every system.

One solution would be to supply a font for zutty, but that would add something most people wouldn't use.


libqtile/utils.py
```diff
@@ -358,30 +358,40 @@ def guess_terminal(preference: str | Sequence | None = None) -> str | None:
...
-        "eterm",
+        "mlterm",
+        "ptyxis",
+        "qterminal",
+        "roxterm",
+        "sakura",
         "st",
+        "terminator",
+        "terminology",
+        "termite",
```
jwijenbergh: Note that termite is archived https://github.com/thestinger/termite?tab=readme-ov-file#termite-is-obsoleted-by-alacritty

jwijenbergh: It was there before but maybe we should remove it

tych0:
Bah, httpbin 500. Maybe we should run a local copy...

libqtile/utils.py
tych0: Oh, sure, let me drop it.

PR #5478, opened 2025-08-31
add prompt_toolkit to optional core dependencies

tych0:
we use this for the repl, which I was just using to debug some stuff. let's add it so it gets installed.


elParaguayo:
Glad to see the repl getting some use!

PR #5482, opened 2025-09-01
core: only cancel tasks once

tych0:
While debugging something else in our exit path, I noticed we try to cancel asyncio tasks twice, once in libqtile/core/loop.py::_cancel_all_tasks(), and once in the finalizer.

The version in _cancel_all_tasks() is probably preferable, since it iterates over all asyncio tasks, so we will cancel stuff that's started by libraries or whatever. So, let's drop our own version.

Also, while I was here, I switched our task list to a set like the docs show, and removed some of the unnecessary callbacks by calling set.discard() directly.

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -58,7 +58,6 @@ from libqtile.resources.sleep import inhibitor
 from libqtile.scratchpad import ScratchPad
 from libqtile.scripts.main import VERSION
 from libqtile.utils import (
-    cancel_tasks,
     create_task,
     get_cache_dir,
     lget,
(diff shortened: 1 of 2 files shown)


elParaguayo (commented):
Interesting. With this change, I'm not sure there's even a need for for `TASKS` (and therefore not for our `create_task` wrapper either).

However, I'm pretty sure I added this code because tasks weren't being cancelled correctly at exit.

tych0:
IIUC we still need TASKS to hold a strong reference as the blog post you linked to says. I didn't spelunk which came first, the utils.py code or the loop.py code, but I think the utils.py cancel is mostly a no-op now...

elParaguayo:
Ah, that was it. We needed the reference.

In which case, your code looks good to me!

PR #5488, opened 2025-09-03
Log upgrades

tych0:
A couple of logging upgrades as a result of #5484


elParaguayo:
All these changes look good. Thanks.

Issue #5114, opened 2024-12-05
Group(spawn=[...]) has strange semantics when the argument is a list


tych0:
### Issue description

The [documentation](https://github.com/qtile/qtile/blob/a49ed89184965d4fd5389418ea4ad087cb627f54/libqtile/config.py#L705-L707) for this parameter says:

> This will be executed (via ``qtile.spawn()``) when the group is created. You can pass either a program name or a list of programs to ``exec()``.

What that means is that the semantics are not the same as `qtile.spawn()`: it runs each element in the list as a separate spawn():

https://github.com/qtile/qtile/blob/a49ed89184965d4fd5389418ea4ad087cb627f54/libqtile/dgroups.py#L130-L137

instead of `spawn()`, which runs the first element in the list and supplies the rest as argv.

I find this quite confusing, but I can see how it evolved: one might want to spawn multiple programs in a single group when it starts. However, it does mean that you can't spawn a single program that requires some arguments.

Of course, I could create a wrapper script, but that is ugly. I just want to run a program with arguments, as I would with `spawn()`, as the docs say!

I am tempted to just fix this to take a nested list, but that seems *especially* ugly. Maybe I should just leave it and add a wrapper script, but the semantics were confusing, and I had to read the code to figure it out, so I'm sure other people are confused as well. Thoughts?

### Version

git master

### Backend

X11 (default)

### Required

- [X] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #5489, opened 2025-09-04
unify graceful_shutdown()

tych0:
Wayland's implementation is better than x11's since it uses actual WM semantics, and also it's not generally very backend-specific, so let's adopt that approach for both backends. We can do this leveraging async so we don't need to block the rest of the event loop, potentially allowing for other shutdown code to happen.

There is one insidiously complex change hidden in here, which is the shift from:

    if not isinstance(win, base.Internal):

to:

    def normal_windows(self) -> list[base.WindowType]:
        return list(
            filter(lambda w: isinstance(w, base.Window), self.windows_map.copy().values())
        )

it turns out our systray is not a base.Internal window, but also does not set NET_WM_PID, so we were previously not killing it in the x11 version via get_pid().

However, if you .kill() (i.e. WM_DELETE_WINDOW instead of the kill(2) syscall) the systray widget's window, the X server kills the connection that window was created from, which just so happens to be qtile's main connection, which kills the whole WM.

So instead, we switch to only trying to kill "normal" windows, since those are the clients with state we want to notify that we're shutting down.

(The systray widget is finalized [or not, modulo bugs :)] via the normal widget finalize() path.)

### DIFF
--- a/libqtile/backend/base/core.py
+++ b/libqtile/backend/base/core.py
@@ -102,9 +102,6 @@ class Core(CommandObject, metaclass=ABCMeta):
     def flush(self) -> None:
         """If needed, flush the backend's event queue."""
 
-    def graceful_shutdown(self):
-        """Try to close windows gracefully before exiting"""
-
     def simulate_keypress(self, modifiers: list[str], key: str) -> None:
         """Simulate a keypress with given modifiers"""
 
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -24,7 +24,6 @@ import asyncio
 import contextlib
 import os
 import signal
-import time
 from collections import defaultdict
 from typing import TYPE_CHECKING, cast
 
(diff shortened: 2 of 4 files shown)


tych0:
One nice side effect of the exitcode test refactoring is that we're now a bit closer to running the tests in parallel at least on the x11 side. I have been running them in parallel locally and there's still some issues, but it's nice that the big negative diff gets us closer to that.

elParaguayo:
Damn coveralls...

tych0:
Yeah, their status page shows some badness, unclear if it's related though...

tych0:
Huh, wild. The wayland dgroups thing failed again...
```
FAILED test/test_dgroups.py::test_dgroup_spawn_in_group[1-wayland-DGroupsSpawnConfig] - assert 0 > 0
 +  where 0 = len([])
 +    where [] = <libqtile.command.client.InteractiveCommandClient object at 0x7f1c7d065fd0>()
 +      where <libqtile.command.client.InteractiveCommandClient object at 0x7f1c7d065fd0> = <libqtile.command.client.InteractiveCommandClient object at 0x7f1c7d0675f0>.windows
 +        where <libqtile.command.client.InteractiveCommandClient object at 0x7f1c7d0675f0> = <test.helpers.TestManager object at 0x7f1c7ceb2210>.c
====== 1 failed, 1260 passed, 59 skipped, 2 xpassed in 1025.04s (0:17:05) ======
```

I wonder if we shouldn't spawn some other window that knows how to speak wayland specifically for this test?

tych0:
I wonder if this is a race between xwayland being ready and the window spawning and trying to use xwayland?

elParaguayo:
That's very possible. Maybe we create a script to create a native Wayland window for those tests.

I've got a separate PR for layout screenshots and I've just used xcffib to create windows.

I don't see any issue with having backend specific window generation.

tych0:
I think we could use the `test/scripts/test_window.py`, which should do the right thing since it really uses gobject to do all the WM level stuff. But we can't for this specific test, because of #5114.

PR #5490, opened 2025-09-04
nix(flake): add `aiohttp` to propagatedBuildInputs

Gurjaka:
For https://github.com/qtile/qtile/commit/f2b68125a1801b4dfe748d851e4a75335f97cc34

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -29,6 +29,9 @@ let
       src = ./..; # use the source of the git repo
     }
     // {
+      propagatedBuildInputs =
+        (qtile-prev.propagatedBuildInputs or [ ]) ++ (with pkgs.python3Packages; [ aiohttp ]);
+
       env = build-config.resolved-env-vars;
 
       pypaBuildFlags = build-config.resolved-config-settings;


jwijenbergh:
I think we should run the tests under nix too for the future

Gurjaka:
> I think we should run the tests under nix too for the future

We definitely should! I found out this was missing from @saviola777 on discord!

tych0:
Is there a way nix can use the upstream dependency list? I have just spent a bunch of time deleting a lot of these disparate dependency lists that we forget to update and are half synchronized, I would love to avoid maintaining another one.

Gurjaka:
> Is there a way nix can use the upstream dependency list? I have just spent a bunch of time deleting a lot of these disparate dependency lists that we forget to update and are half synchronized, I would love to avoid maintaining another one.

It should be possible, but will be hacky, also that would require re-writing qtile derivation for the repo. 
@Sigmanificient what do you think, you know more about this stuff 😄

Sigmanificient:
Yeah it is possible but not very reliable due to names not being 1:1 between pypi <-> nixpkgs

Gurjaka:
> Yeah it is possible but not very reliable due to names not being 1:1 between pypi <-> nixpkgs

@tych0 In these case, I think best way to deal with this is to tag me when a PR is updating the dependency list I guess. That way I can always keep nix side in sync.

Gurjaka:
Or maybe we can write a small script that checks if dependencies have updated, then search the `nixpkgs` for similar packages, and automatically add to nix derivation? Before adding any changes tho, script should run the `nix flake check`, seems complicated but doable with help of [nh](https://github.com/nix-community/nh)

tych0:
> Yeah it is possible but not very reliable due to names not being 1:1 between pypi <-> nixpkgs

Yeah, fair enough. I guess mostly the distro packaging lives out-of-tree, so I don't see this bit, and nix is the outlier here.

(I also notice that we still have a separate dependency list in `.github/workflows/release.yml` that I did not hit in my cleanup... work for another day.)

PR #5492, opened 2025-09-05
[wayc] Deactivate previous surfaces on focus

richcarni:
xwayland_view requires activation of a newly focused surface and deactivation of the previously focused service. The absence of which was causing issues with xwayland views reliably receiving keyboard input

Since the previous surface could be either an xdg_toplevel or xwayland_surface (or potentially other in future), deactivate_surface needs to work for all these different surfaces

xdg_view requires the same surface deactivation of the previously focused surface


richcarni:
This is another one that was merged but seems to have been overwritten the other day (see #5457)

I arrived here planning to add a check for layer_view, but doesn't look like layer_view does activate/deactivate surface

jwijenbergh:
Thanks!

PR #5495, opened 2025-09-07
Less threads

tych0:
With the asyncio switch I am sort of on a rampage to use less threads in qtile, since they add complexity (viz. things like event loop policies, etc.), memory footprint, complexities due to fork(), etc.

A bunch of the widgets were just using threads where they didn't need to, so here's those simple conversions.

I am working on more porting to GenPollCommand for some other widgets, but that will come in a later branch.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -396,7 +396,7 @@ class _LinuxBattery(_Battery, configurable.Configurable):
         )
 
 
-class Battery(base.BackgroundPoll):
+class Battery(base.InLoopPollText):
     """
     A text-based battery monitoring widget supporting both Linux and FreeBSD.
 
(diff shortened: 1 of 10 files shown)


tych0:
A handful of these are monitoring proc and sysfs files. I have wanted to switch away from polling those files to some kind of fanotify or inotify-based API for a while, but there's nothing that's obviously great. Using asyncio-based stuff would add a dependency for a small set of widgets, or if we rolled our own we'd have to do another cffi binding which maybe isn't the end of the world. The inotify syscall API is relatively simple, so perhaps we could just use it directly?

But anyway, just doing these as read()s from the event loop now is good enough.

elParaguayo:
So these just do a straight read in the main loop? Makes sense if we're happy these can't hang/block.

tych0:
Yeah, exactly. In principle they shouldn't block, modulo kernel bugs. I have seen it happen of course, but it's not expected.

PR #5496, opened 2025-09-08
Drop string events

tych0:
I have been meaning to come back to `qtile top` for a while... at least two of the allocations from Qtile are easy to get rid of. I have also landed some fixes for the other ones in xcffib's tip of tree.

There are probably more allocations that can be avoided, but two of the top four from Qtile on my machine are totally eliminated by this patch series.

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -71,8 +71,10 @@ _NET_WM_STATE_TOGGLE = 2
 
 
 def _geometry_getter(attr):
+    attr_name = "_" + attr
+
     def get_attr(self):
-        if getattr(self, "_" + attr) is None:
+        if getattr(self, attr_name) is None:
             g = self.window.get_geometry()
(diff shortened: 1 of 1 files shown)


libqtile/backend/x11/core.py
```diff
@@ -384,21 +382,45 @@ def _get_target_chain(self, event_type: str, event) -> list[Callable]:
         """
         assert self.qtile is not None
 
-        handler = f"handle_{event_type}"
+        handler = {
```
elParaguayo: Is this dict getting recreated every time? Can't it be constant?

tych0: oh, yeah, I thought I had pushed that change! will do.

tych0: Oh, because this is the wrong branch...

tych0: Ok, now it should be more sane :)

PR #5497, opened 2025-09-12
Widget cleanups

tych0:
Here is a handful of cleanups/features I've been using locally for a bit.

The biggest one is our zombie reaping, which interferes with internal asyncio stuff (ugh...). We could have fixed this via custom event loop policies, but those are going away, so let's just keep a list of things not to actually wait() on.

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -636,16 +636,57 @@ def remove_dbus_rules() -> None:
         bus._sock.close()
 
 
+ASYNC_PIDS: set[int] = set()
+
+
+async def acall_process(command: str | list[str], shell: bool = False) -> str:
+    """
+    Like call_process, but the async version. Tracks PIDs in ASYNC_PIDS.
+    """
+    stdin = asyncio.subprocess.DEVNULL
+    stdout = asyncio.subprocess.PIPE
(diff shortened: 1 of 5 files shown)


tych0:
Ok, I think this is "just" another race:

```
2025-09-12T19:42:00.8636502Z test/widgets/test_systray.py ....                                        [ 49%]
2025-09-12T19:42:11.8332163Z test/widgets/test_tasklist.py F........XX..                              [ 50%]
```

systray is what spawns the volumeicon test, but the error is from the tasklist test:

```
tasklist_manager = <test.helpers.TestManager object at 0x7fd00391b4d0>

    @horizontal_and_vertical
    def test_tasklist_defaults(tasklist_manager):
        widget = tasklist_manager.c.widget["tasklist"]
    
        tasklist_manager.test_window("One")
        tasklist_manager.test_window("Two")
>       assert widget.info()["text"] == "One|Two"
E       AssertionError: assert 'volumeicon|One|Two' == 'One|Two'
E         
E         - One|Two
E         + volumeicon|One|Two

test/widgets/test_tasklist.py:104: AssertionError
```

I'm sure the root cause here is part of why we can't run the tests in parallel. I think we don't do a good enough job of cleaning things up once we've spawned them.

PR #5499, opened 2025-09-14
[wayc] Move stacking commands to window.Base

elParaguayo:
This enables Internal windows to access commands like keep_above/below etc.


jwijenbergh:
merci

PR #5500, opened 2025-09-16
[wayc] Move XDG toplevel request listeners

richcarni:
The issue is that clients can make requests like maximize for toplevel XDG surfaces before the surface is managed (i.e. callbacks are attached) and even before the surface is initialized (e.g. discord updater screen)

Following the way Sway handles this, we can delay attaching these event listeners until after the surface has been managed. Any requests pending will be handled then


jwijenbergh:
Thanks!

PR #5501, opened 2025-09-16
[wayc] Implement window has_fixed_sized

richcarni:
Required by float rules

Implemented and tested for xdg-view. Stub added for xwayland-view


jwijenbergh:
Thanks!

PR #5502, opened 2025-09-16
Docs build cleanups

tych0:


docs/sphinx_qtile.py
```diff
@@ -53,14 +53,8 @@ def generate_widget_screenshots():
...
 def setup(app):
-    run(["make", "-C", base_dir, "run-ffibuild"])
     generate_keybinding_images()
-    # screenshots will be skipped unless QTILE_BUILD_SCREENSHOTS environment variable is set
-    # Variable is set for ReadTheDocs at https://readthedocs.org/dashboard/qtile/environmentvariables/
-    if os.getenv("QTILE_BUILD_SCREENSHOTS", False):
-        generate_widget_screenshots()
-    else:
-        print("Skipping screenshot builds...")
+    generate_widget_screenshots()
```
elParaguayo: The idea was to make this optional so if people are making changes to docs and want to test locally, they don't have to wait for the screenshots to build.

tych0: Yeah, I understand the intent, but I tripped over it. I spent a while trying to figure out why the screenshots wouldn't build even though I had made changes to them on a branch. It turns out you also need to export this variable.

I can change this patch to default it to on, and export the variable if you want building the screenshots off. But then my thought process was: the only two people who are building the docs locally are probably you and I, and we both probably want to generate the screenshots. But maybe you don't :)

Let me know, I'm happy to change it to default-on, but I do think the sphinx build should build everything by default, and you disable things if you want to go faster.

Another approach would be to actually teach make about the screenshot dependencies, then it would only build exactly what it needed to. But given that it takes <1m to run on my laptop, I'm not too bothered by just doing it every time.

elParaguayo: Ha. Yes, it probably is just us. I'm ok with building by default but would like the ability to disable as sometimes I may need lots of builds to fix the stupid mistakes that I will invariably make!

tych0: Ok, I'll switch it to on-by-default.

elParaguayo (approved):

docs/sphinx_qtile.py
ervinpopescu: I built the docs once! (pick me behavior 😆) Retreating to my no-screenshots lair now...

tych0: Now if you want to build the widgets, you won't have to figure out how! :D

PR #5503, opened 2025-09-17
[wayc] Fix xwayland crashes

richcarni:
Xwayland still needs a lot of work - both cleanup and implementation of functionality - but this small commit at least fixes some causes of segfaults

Tested with xterm and discord app


jwijenbergh:
Thanks!

PR #5505, opened 2025-09-20
[wayc] Add fullscreen background

elParaguayo:
It's in the XDG spec but I've added to xwayland too.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -122,11 +122,11 @@ def log_cb(importance: int, formatted_str: ffi.CData) -> None:
     """Callback that logs the string at the given level"""
     log_str = ffi.string(formatted_str).decode()
     if importance == lib.WLR_ERROR:
-        qw_logger.error(log_str)
+        logger.error(log_str)
     elif importance == lib.WLR_INFO:
-        qw_logger.info(log_str)
+        logger.info(log_str)
     elif importance == lib.WLR_DEBUG:
-        qw_logger.debug(log_str)
+        logger.debug(log_str)
 
 
 @ffi.def_extern()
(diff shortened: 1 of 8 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -122,11 +122,11 @@ def log_cb(importance: int, formatted_str: ffi.CData) -> None:
     """Callback that logs the string at the given level"""
     log_str = ffi.string(formatted_str).decode()
     if importance == lib.WLR_ERROR:
-        qw_logger.error(log_str)
+        logger.error(log_str)
```
elParaguayo: This probably shouldn't be part of this PR but I don't understand why this isn't the default position.

jwijenbergh:
why did coveralls decrease by 10%

libqtile/backend/wayland/core.py
jwijenbergh: fine by me, let's tackle fixing qw_logger later

elParaguayo:
> why did coveralls decrease by 10%

Weird. As some of the checks say coverage increased! Could just be a coveralls bug as they've migrated us to a different process.

Let me rebase and fix the conflict when I get a chance and then we can see if this is still an issue.

libqtile/backend/wayland/core.py
elParaguayo: I'll remove this from the PR for now.

libqtile/backend/wayland/qw/xwayland-view.c
```diff
@@ -545,6 +545,21 @@ static bool qw_xwayland_view_has_fixed_size(void *self) {
     return false;
 }
 
+static void qw_xwayland_view_update_fullscreen(void *self, bool fullscreen) {
```
Gurjaka, line 548: unused function?

Gurjaka: nvm, found it

elParaguayo:
@jwijenbergh coveralls is ok on this one now? Should we merge it or do you want more changes?

jwijenbergh:
Fine to go, thanks!

PR #5514, opened 2025-09-24
[wayc] Build wayc with makefile

richcarni:
Build wayc as library with a makefile. Then build an interface to the library via cffi

This avoids creating a single _ffi.c source file making reading errors and debugging more natural

Maybe improves stability of build process?

### DIFF
--- a/libqtile/backend/wayland/cffi/Makefile
+++ b/libqtile/backend/wayland/cffi/Makefile
@@ -1,0 +1,29 @@
+# Paths
+QW_PATH      := ../qw
+QW_PROTO_OUT := ../qw/proto
+
+# Compiler/linker
+CC := gcc
+CFLAGS := -O2 -fPIC \
+          -DWLR_USE_UNSTABLE \
+          -I$(QW_PATH) \
+          -I$(QW_PROTO_OUT) \
+          $(shell pkg-config --cflags wlroots-0.19 wayland-server cairo pixman-1 libdrm)
+
+LDFLAGS := -shared $(shell pkg-config --libs wlroots-0.19 wayland-server cairo pixman-1 libdrm)
+
+TARGET := ../libwayc.so
+
+.PHONY: all
+all: $(TARGET)
+
+$(TARGET): $(wildcard $(QW_PATH)/*.c)
+	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
+
+.PHONY: clean
+clean:
+	rm -f $(OBJECTS) $(TARGET)
+
+.PHONY: rebuild
+rebuild: clean all
+
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -143,25 +143,21 @@ for file in cdef_files:
             CDEF += line
 
 SOURCE = ""
+for header in cdef_files:
+    SOURCE += f'#include "{header}"\n'
 
-for root, dirs, files in os.walk(QW_PATH):
-    for file in files:
-        if file.endswith(".c"):
-            with open(QW_PATH / file) as f:
-                SOURCE += f.read()
-
-
+#TODO: Is this used/needed?
 def get_include_path(lib: str) -> str:
     return subprocess.run(
         ["pkg-config", "--variable=includedir", lib], text=True, stdout=subprocess.PIPE
     ).stdout.strip()
 
+LIBWAYC = QW_PATH / ".." / "libmyffi.so"
 
 ffi = FFI()
 ffi.set_source(
     "libqtile.backend.wayland._ffi",
     SOURCE,
-    libraries=["wlroots-0.19", "wayland-server", "input", "cairo"],
     define_macros=[("WLR_USE_UNSTABLE", None)],
     include_dirs=[
         os.getenv("QTILE_CAIRO_PATH", "/usr/include/cairo"),
@@ -171,8 +167,11 @@ ffi.set_source(
         QW_PATH,
         QW_PROTO_OUT_PATH,
     ],
+    extra_link_args=[LIBWAYC],
 )
 
+def build_library():
+    subprocess.run(["make", "-C", ".", "rebuild"], check=True)
 
 def ffi_compile(verbose: bool = False) -> None:
     # The ffi source of "libqtile.backend.wayland._ffi" means that we'll compile the library file
@@ -191,4 +190,6 @@ def ffi_compile(verbose: bool = False) -> None:
 
 ffi.cdef(CDEF)
 if __name__ == "__main__":
+
+    build_library()
     ffi_compile()


Gurjaka:
@Sigmanificient you were working on build.py refactor right? This might be interesting

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+# Compiler/linker
+CC := gcc
+CFLAGS := -O2 -fPIC \
+          -DWLR_USE_UNSTABLE \
+          -I$(QW_PATH) \
+          -I$(QW_PROTO_OUT) \
+          $(shell pkg-config --cflags wlroots-0.19 wayland-server cairo pixman-1 libdrm)
+
+LDFLAGS := -shared $(shell pkg-config --libs wlroots-0.19 wayland-server cairo pixman-1 libdrm)
```
tych0: LDLIBS right? not LDFLAGS? Then you don't need to pass it explicitly below (same with CFLAGS)

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+TARGET := ../libwayc.so
+
+.PHONY: all
+all: $(TARGET)
+
+$(TARGET): $(wildcard $(QW_PATH)/*.c)
+	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
```
tych0: Is it worth setting `-g` somehow? I don't know how you guys are really debugging this.

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
+# Paths
+QW_PATH      := ../qw
+QW_PROTO_OUT := ../qw/proto
```
Sigmanificient: Maybe this is could be built from `QW_PATH` to make it more flexible
```suggestion
QW_PROTO_OUT := ../$(QW_PROTO)/proto
```

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+# Compiler/linker
+CC := gcc
```
Sigmanificient: Forcing `CC` is a bit of a bad practice, because it makes it harder to change
```suggestion
CC ?= gcc
```

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+CFLAGS := -O2 -fPIC \
+          -DWLR_USE_UNSTABLE \
+          -I$(QW_PATH) \
+          -I$(QW_PROTO_OUT) \
+          $(shell pkg-config --cflags wlroots-0.19 wayland-server cairo pixman-1 libdrm)
```
Sigmanificient:
```suggestion
CFLAGS := -O2 -fPIC
CFLAGS += -DWLR_USE_UNSTABLE
CFLAGS += -isystem $(QW_PATH) -isystem $(QW_PROTO_OUT)
CFLAGS += $(shell pkg-config --cflags wlroots-0.19 wayland-server cairo pixman-1 libdrm)
```

This avoid `\`+spaces combo (spaces are nasty in Makefiles too!)
I also changed `-I` to `-iquote`

libqtile/backend/wayland/cffi/Makefile
Sigmanificient: yup! `LDFLAGS` would be for something like `-flto`

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+LDFLAGS := -shared $(shell pkg-config --libs wlroots-0.19 wayland-server cairo pixman-1 libdrm)
+
+TARGET := ../libwayc.so
+
+.PHONY: all
+all: $(TARGET)
+
+$(TARGET): $(wildcard $(QW_PATH)/*.c)
```
Sigmanificient: I think this should be moved to its own variable, I would also suggest doing `.o` intermediates:

```make
BUILD_DIR := .build


SRC := $(shell find -type f -name "*.c")
OBJS := $(SRC:%.c=$(BUILD_DIR)/%.o)

$(BUILD_DIR)/%.o: %.c
    $(COMPILE.c) $@ $^
```

libqtile/backend/wayland/cffi/Makefile
Sigmanificient: Using `-g` will also you to get more info, such as line number in valgrind reports. Nowadays this is not that heavy to keep it, but you can also have 2 builds if you prefer (release/debug)

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
+
+.PHONY: clean
+clean:
+	rm -f $(OBJECTS) $(TARGET)
+
+.PHONY: rebuild
+rebuild: clean all
```
Sigmanificient:
```suggestion
.PHONY: rebuild
.NOTPARALLEL: rebuild
rebuild: clean all
```
otherwise it breaks with `make rebuild -j`.
I also suggest `re` shorthand

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+.PHONY: all
+all: $(TARGET)
+
+$(TARGET): $(wildcard $(QW_PATH)/*.c)
+	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
+
+.PHONY: clean
+clean:
+	rm -f $(OBJECTS) $(TARGET)
```
Sigmanificient:
```suggestion
	$(RM) $(OBJECTS) $(TARGET)
```

Sigmanificient: `$(OBJECT)` is not used here, did you plan to do `.o`?

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
+# Paths
+QW_PATH      := ../qw
+QW_PROTO_OUT := ../qw/proto
+
+# Compiler/linker
+CC := gcc
+CFLAGS := -O2 -fPIC \
```
Sigmanificient: Not settings `-Wall -Wextra`? Maybe `-pedantic` too.
It would also be nice to provide a standard to follow (not too recent too, for better portability)

libqtile/backend/wayland/cffi/Makefile
Sigmanificient:
```suggestion
$(TARGET): $(wildcard $(QW_PATH)/*.c)
	$(LINK.c) -o $@ $^ $(LDFLAGS) $(LDLIBS)
```

libqtile/backend/wayland/cffi/build.py
```diff
@@ -171,8 +167,11 @@ def get_include_path(lib: str) -> str:
         QW_PATH,
         QW_PROTO_OUT_PATH,
     ],
+    extra_link_args=[LIBWAYC],
 )
 
+def build_library():
+    subprocess.run(["make", "-C", ".", "rebuild"], check=True)
```
Sigmanificient: Probably worth adding a configurable `-j`!

libqtile/backend/wayland/cffi/Makefile
Sigmanificient: I also suggest the following:
```mk
/ := ..
QW_PATH := $/qw
QW_PROTO_OUT := $/qw/proto
```

libqtile/backend/wayland/cffi/Makefile
```diff
@@ -0,0 +1,29 @@
+# Paths
```
Sigmanificient: A small marker to indicate that we follow `POSIX`
```suggestion
.POSIX:
# Paths
```

libqtile/backend/wayland/cffi/build.py
richcarni, line 1: L166-173:
```
include_dirs=[
    os.getenv("QTILE_CAIRO_PATH", "/usr/include/cairo"),
    os.getenv("QTILE_PIXMAN_PATH", "/usr/include/pixman-1"),
    os.getenv("QTILE_LIBDRM_PATH", "/usr/include/libdrm"),
    os.getenv("QTILE_WLROOTS_PATH", "/usr/include/wlroots-0.19"),
    QW_PATH,
    QW_PROTO_OUT_PATH,
],
```

I was wondering how to synchronise these between here and the makefile?

Would it make sense to have the build file generate the makefile?

Also, while it's useful to be able to specify custom include paths, do we also need to be concerned about custom lib paths? On ubuntu, I have to build wlroots from source and because I like to keep it out of my /usr dirs, this is a pain point for me

Sigmanificient: I think it is definitely doable to have the build.py script emit a `libconfig.mk` that is included afterwards into the Makefile

libqtile/backend/wayland/cffi/Makefile
richcarni: I do use gdb so this would be useful. I currently manually edit `build.py` when debugging

libqtile/backend/wayland/cffi/build.py
richcarni: That would work

Although I'm wondering now whether there is much value in using a Makefile at all - maybe just easier to call the compile instruction directly in `build.py`?

tych0: Yeah, maybe just doing it directly in build.py is the best? execing a bunch of commands for e.g. pkgconfig will be kind of annoying, but not the end of the world.

libqtile/backend/wayland/cffi/Makefile
richcarni: That is neat but I think I opt for simple

libqtile/backend/wayland/cffi/Makefile
richcarni: This didn't work for me. Make kept defaulting to c99, which wouldn't build - I think because it doesn't support some GNU extensions

Anyway, I've made it settable from python

libqtile/backend/wayland/cffi/Makefile
richcarni: Added -Wall -Wextra
Figure can always add -pedantic as CFLAG now

Tried -std=c99 and had issues as mentioned above, maybe -std=gnu99 ??

libqtile/backend/wayland/cffi/Makefile
richcarni: Did you add `-isystem` backwards? Should be before wlroots, cairo include paths?

Is the main benefits suppressing warnings? It doesn't seem to be an issue at present so I've left as is for now

libqtile/backend/wayland/cffi/Makefile
richcarni: I think I understand and have more or less followed this

libqtile/backend/wayland/cffi/Makefile
richcarni: I think should be able to pass as a CFLAG now

libqtile/backend/wayland/cffi/Makefile
richcarni: Removed rebuild

libqtile/backend/wayland/cffi/build.py
richcarni: Decided to try with .mk file - can see pros and cons with either approach

libqtile/backend/wayland/cffi/build.py
richcarni: Yes, but not sure of an easy way to pass. Anyway, build is only a few seconds so probably not critical at the moment

PR #5516, opened 2025-09-26
[wayc] Update move up/down

elParaguayo:
Moving windows up and down now moves relative to windows that aren't hidden and are on the same output.

@jwijenbergh here's an attempt at the stacking logic that we discussed on discord.

### DIFF
--- a/libqtile/backend/wayland/qw/util.c
+++ b/libqtile/backend/wayland/qw/util.c
@@ -87,3 +87,20 @@ void qw_util_deactivate_surface(struct wlr_surface *surface) {
         }
     #endif
 }
+
+bool qw_surfaces_on_same_output(struct wlr_surface *surface_a, struct wlr_surface *surface_b) {
+    if (surface_a == NULL || surface_b == NULL) {
+        return false;
+    }
+
+    struct wlr_surface_output *output_a;
+    wl_list_for_each(output_a, &surface_a->current_outputs, link) {
+        struct wlr_surface_output *output_b;
+        wl_list_for_each(output_b, &surface_b->current_outputs, link) {
+            if (output_a->output == output_b->output) {
+                return true;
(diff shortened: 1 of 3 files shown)


libqtile/backend/wayland/qw/view.h
```diff
@@ -65,6 +65,7 @@ struct qw_view {
     char *title;
     char *app_id;
     struct wlr_scene_tree *content_tree; // Scene tree holding the view's content
+    struct wlr_output *current_output;   // used for stacking order with views on same output
```
jwijenbergh: hmmm wlr_surface has a list of outputs, can't we just use this? https://kennylevinsen.pages.freedesktop.org/wlroots/wlr/types/wlr_compositor.h.html#struct-wlr_surface

I don't like this bookkeeping

elParaguayo: No problem. I didn't realise there was a list of surfaces. That's even better.

elParaguayo: My idea had been to limit the stacking within the group that the view is on. If a view is floating and extends into another output, that would be ignored by my code. If we look at the list of outputs then I assume any output which contains part of the surface is in the list. In that case, our stacking code will take account of windows in all outputs containing part of the window being moved. I think that's probably correct but just wanted to explain my thinking.

elParaguayo:
@jwijenbergh I've updated this to find the wlr_surface object from the view's tree and then compare outputs in the surface's `current_outputs`. It's all server side now but there's more iterating over lists to do this.

jwijenbergh:
IMO this is fine and easy to change as it's a separate helper function

jwijenbergh:
Whoops conflicts now

jwijenbergh:
thanks

PR #5515, opened 2025-09-26
pyproject: fix xdg component

tych0:
As the pypi for xdg notes https://pypi.org/project/xdg/:

    xdg has been renamed to xdg-base-dirs due to an import collision with PyXDG. Therefore the xdg package is deprecated. Install xdg-base-dirs instead.

indeed, we really want pyxdg in the tasklist widget. So let's specify that.

PR #5517, opened 2025-09-27
Fix X11 backend test (and docs)

elParaguayo:
xcffib>1.11.0 has a breaking change. `xcffib.testing.find_display()` now returns a tuple of the display number and a file decriptor. Previously, the method returned a file object which we closed with `.close()`.

### DIFF
--- a/test/backend/x11/conftest.py
+++ b/test/backend/x11/conftest.py
@@ -72,7 +72,7 @@ def stop_x11(proc, display, display_file):
 
     # clean up the lock file for the display we allocated
     try:
-        display_file.close()
+        os.close(display_file)
         os.remove(xcffib.testing.lock_path(int(display[1:])))
     except OSError:
         pass


tych0:
Oh, oof. And this change doesn't even fix the problem. 

Maybe I'll take a 1.11.2 instead with this API change reverted, no sense in changing things if they don't actually fix the problem. 

Thoughts?

elParaguayo:
What problem were you trying to fix?

tych0:
That `find_display()` (or something about the xvfb testing code) does not actually isolate displays, so you can end up with failures like I'm seeing:

https://github.com/qtile/qtile/pull/5497#issuecomment-3287117115

But the pyxdg fix seems to be able to trigger it much more reliably:

https://github.com/qtile/qtile/pull/5515

tych0:
https://github.com/tych0/xcffib/commit/aae84636c50ac33f978613674fa0652356a3bcc2 was the attempted fix, but it does not seem to have helped at all, given that I got a failure in CI with that patch applied here: https://github.com/tych0/xcffib/actions/runs/18052535989/job/51377039923

tych0:
Ok, I just reverted and released 1.11.2; hopefully that will fix this.

Now that I have a reliable reproducer, hopefully I can finally track down this bug. This has really bothered me for years: qtile's test suite should be easy to run in parallel, but the fact that it's not means we're doing something wrong. I just have never been able to figure out exactly what :)

PR #5518, opened 2025-09-28
Disable flaky spawn_in_group test

elParaguayo:
Test is causing failures on unrelated PRs. Let's skip it for now.

### DIFF
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -428,7 +428,10 @@ def test_spawn_list(manager):
 
 
 @manager_config
-def test_spawn_in_group(manager):
+def test_spawn_in_group(manager, backend_name):
+    if backend_name == "wayland":
+        pytest.skip("TODO: X11 only for now.")
+
     @Retry(ignore_exceptions=(AssertionError,))
     def wait_for_window(empty=False):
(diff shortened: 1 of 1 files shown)


tych0:
Do you have a link to one of the failures?

elParaguayo:
https://github.com/qtile/qtile/actions/runs/18072869419/job/51426748583?pr=5494

tych0:
Oh yeah, I looked at this a couple of weeks ago. I think we should only skip it on wayland, I believe x11 should work fine.

I think the bug is in the wayland backend: the xwayland code starts up in `Core.__init__`, but is not necessarily ready. Once we do `on_config_load()`, we execute all the `startup_once` hooks, which is what dgroups uses to spawn these group arguments. If that happens before `on_xwayland_ready` fires, I suspect things are broken.

I was going to wait until `wayc` was landed to try and fix it, but maybe you guys can fix it in the process of doing `wayc`. It seems like we need to defer running `startup_once` hooks until everything is actually ready in wayland land. I don't know exactly what "everything" is, but it looks like that 1. there's an output present 2. xwayland is ready at the very least...

tych0:
Thanks!

PR #5525, opened 2025-09-30
[wayc] Fix xdg_popups not taking button presses

richcarni:
Fix: Treat popups as a type of view by adding qw_view base struct

Also fixed some unrelated bugs, updated parent type, and removed some unnecessary code


elParaguayo:
Thanks.

PR #5527, opened 2025-09-30
[wayc] Remove decorations from XDG windows

elParaguayo:
Our decoration request mode function was not working as the mode is only set when the surface is initiliased and the call to this function happened before then. Adding a call to set the mode once the surface has committed fixes this issue.

NB I've added a forward declaration to the function here to avoid reordering functions in the module but let me know if you'd like this changed.


libqtile/backend/wayland/qw/xdg-view.c
```diff
@@ -90,6 +92,11 @@ static void qw_xdg_view_handle_commit(struct wl_listener *listener, void *data)
         xdg_view->base.title = xdg_view->xdg_toplevel->title;
         xdg_view->base.app_id = xdg_view->xdg_toplevel->app_id;
     }
+
+    if (xdg_view->decoration != NULL) {
+        qw_xdg_view_handle_decoration_request_mode(&xdg_view->decoration_request_mode,
```
elParaguayo: One other question: Should this also be inside the `initial_commit` block?

richcarni: That makes sense to me, rather than checking every time focus is requested

richcarni:
Good question about forward declaration vs re-ordering functions. I'm always unsure where to put new functions! Personally, I would probably avoid forward declarations unless there's no other option

elParaguayo:
I've made both changes and I still get the desired behaviour.

PR #5528, opened 2025-09-30
[wayc] Destroy internal views on kill

elParaguayo:
Killing internal windows (e.g. Popups) just hides them. That doesn't seem right. This PR destroys the nodes instead.


richcarni:
I think we also need to free the view. When I tried this it failed the exit test so that’s where I got stuck. Maybe needs an on destroy listener on the buffer to do this?

elParaguayo:
Yes - that makes sense.

Separately, what on earth is going on with coveralls at the moment?!

richcarni:
Not sure about coveralls - I'm not really familiar with it (yet!)

About freeing the view - ignore what I said about listeners. The problem is kill is called twice, once by `core.graceful_shutdown()` and then again by `bar.finalize()`

That aside, I'm not 100% sure what needs to be freed but suspect it could be as much as:
```
    cairo_surface_destroy(view->image_surface);
    wlr_buffer_drop(view->buffer);
    wlr_scene_node_destroy(&view->base.content_tree->node);
    free(view);
```

elParaguayo:
Killing a popup window removes it from the window map so it shouldn't be killed again via `graceful_shutdown`. However, the bar *is* being killed twice which is a problem!

elParaguayo:
I added some code to stop windows being killed more than once so this should be good to go.

PR #5530, opened 2025-10-01
[Wayc] fullscreen fixes

### LINKED ISSUE #5529: [wayc] Problem full-screening windows
### Issue description

Using mpv video player:
- Made player fullscreen (using fullscreen button on app. note some of the focus border remained)
- Pressed fullscreen button again and was now fullscreen without any border

But then could not get out of fullscreen. Closing app, screen was all black (can see and move cursor). Tried switching to other groups but could only see a black screen + cursor

Needs further investigation

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

elParaguayo:
Two commits:
* Fixes issue with mpv not exiting fullscreen on `escape` (should also address issues with using fullscreen button)
* Fixes issue with fullscreen background not being disabled when a fullscreen window is killed

Closes #5529


richcarni:
So close!

Fullscreen toggle and esc key work perfectly! ✅

When I kill a full screen window the black background is disabled ✅

But there is still an issue with the black background remaining active across other groups. So if I open and fullscreen a window in group 1 and change to any other group, they are all black

elParaguayo:
Nice bug report. I've fixed that issue now (basically, I force a check of the background status whenever the group on a screen changes)

richcarni:
Working perfectly now for me!

Not sure what is going on with ci

elParaguayo:
That CI failure is my fault. Will fix later today.

PR #5534, opened 2025-10-03
Licensing + Maintainers changes

tych0:
cc @elParaguayo @jwijenbergh @flacjacket @cortesi @ramnes @m-col 

We talked about these offline, but here they are as an actual PR.


tych0:
heh, more x11 racing:
> E       AssertionError: assert 'volumeicon|One|Two' == 'One|Two'

I have been working pretty angrily on a better locking mechanism exporting from xcffib so we can use it here too... but I haven't quite gotten it working yet.

ramnes (approved):
Wonderful!

flacjacket (approved):

PR #5535, opened 2025-10-03
x11: clear _NET_ACTIVE_WINDOW when not focused

tych0:
EWMH says this should be None (0) if no window is focuse:

https://specifications.freedesktop.org/wm-spec/1.3/ar01s03.html#id-1.4.10


elParaguayo:
Very neat. I'm happy to merge this unless there's anything else you want to add here.

tych0:
Nah, maybe just wait a day or two to get confirmation that it really does fix the bug? But it fixes *a* bug at least...

PR #5536, opened 2025-10-03
[wayc] Improve wayland build

elParaguayo:
- Adds option to `make wayland[-force]` to qtile's makefile.
- Adds argument parsing to build.py
- Checks hash of header files and forces rebuild if these have changed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,3 +45,10 @@ clean: ## Clean generated files
 update-flake: ## Update the Nix flake.lock file, requires Nix installed with flake support, see: https://nixos.wiki/wiki/Flakes
 	nix flake update
 
+.PHONY: wayland
+wayland:  ## Build the wayland backend
+	@python libqtile/backend/wayland/cffi/build.py
+
+.PHONY: wayland-force
+wayland-force:  ## Force complete rebuild of the wayland backend
+	@python libqtile/backend/wayland/cffi/build.py --force
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -1,3 +1,5 @@
+import argparse
+import hashlib
 import os
 import subprocess
 import sysconfig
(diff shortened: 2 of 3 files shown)


richcarni:
I don't think this will solve the problem that only updating a header won't trigger make to rebuild libwayc.so

I think that all `build.py --force` needs to do is call `make clean` before calling `make` (or better, put back the `make rebuild` rule that I'd originally included and call that)

Shouldn't be any need to delete any _ffi files. I did check this: by modifying view.h, rebuilding, and checking all the _ffi files have fresh time stamps

elParaguayo:
Not sure I follow. If the headers have changed then we call `make -C libqtile/backend/wayland/cffi clean`. Isn't that what you meant?

richcarni:
Yep, exactly (is what I meant by make clean)

No makefiile guru, but this appears to be the automated way (rather than us checking if headers have changed)
https://stackoverflow.com/a/52036564

Gurjaka:
Perhaps @Sigmanificient (makefile guru) could help with this?

elParaguayo:
I'd welcome and and all suggestions!

The trouble is, I'm struggling to understand why the issue happened. I changed a header file AND some of the c files but there seemed to be a mismatch between libwayc and the FFI build.

richcarni:
I didn't have time to investigate this properly, but GPT spat out this (based on the link above)

```
.POSIX:

include libconfig.mk

BUILD_DIR := .build
SRC := $(wildcard $(QW_PATH)/*.c)
OBJS := $(patsubst $(QW_PATH)/%.c,$(BUILD_DIR)/%.o,$(SRC))
DEPENDS := $(OBJS:.o=.d)

.PHONY: all
all: $(TARGET)

$(BUILD_DIR)/%.o: $(QW_PATH)/%.c
	@mkdir -p $(dir $@)
	$(CC) -O2 -fPIC -DWLR_USE_UNSTABLE -Wall -Wextra $(CFLAGS) $(INCLUDES) -MMD -MP -c $< -o $@

$(TARGET): $(OBJS)
	$(LINK.c) -shared $(LDFLAGS) -o $@ $^ $(LDLIBS)

.PHONY: clean
clean:
	$(RM) $(OBJS) $(DEPENDS) $(TARGET)

-include $(DEPENDS)
```

Makefile
```diff
@@ -45,3 +45,10 @@ clean: ## Clean generated files
 update-flake: ## Update the Nix flake.lock file, requires Nix installed with flake support, see: https://nixos.wiki/wiki/Flakes
 	nix flake update
 
+.PHONY: wayland
+wayland:  ## Build the wayland backend
+	@python libqtile/backend/wayland/cffi/build.py
```
tych0, line 50: `uv run python`?

libqtile/backend/wayland/cffi/build.py
```diff
@@ -188,11 +190,23 @@ def get_include_path(lib: str) -> str:
...
+
+def build_library(clean: bool = False) -> None:
+    if clean:
+        subprocess.run(["make", "-C", QW_PATH / ".." / "cffi", "clean"], check=True)
+
     subprocess.run(["make", "-C", QW_PATH / ".." / "cffi"], check=True)
 
 
-def ffi_compile(verbose: bool = False) -> None:
+def ffi_compile(verbose: bool = False, force: bool = False) -> None:
```
tych0, line 209: `--force` here is weird to me, and suggests something else about the dependency calculations is incorrect. Can we just fix that instead?

tych0: Oh, I see, you guys are already discussing this :).

tych0:
> $(BUILD_DIR)/%.o: $(QW_PATH)/%.c

here, you need to include any .h files that .c includes. The pattern people usually use for that is generating .deps files with a list, and then catting those and depending on make.

But maybe we can get away for now with just depending on *.h for each .o? That would rebuild the world if any header file changes are made, but perhaps that's okay for now.

Alternatively, we could switch to something like meson, which would do all of this for us.

PR #5540, opened 2025-10-04
[wayc] Fix window title crash

### LINKED ISSUE #5532: [wayc] Setting custom window title
### Issue description

For example:
```
gnome-terminal --title=test
```
appears to create an unmanaged window

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

elParaguayo:
Some apps (e.g. gnome-terminal) seem to free the memory holding the window title after the view has been committed. We therefore need to copy the string rather than the pointer to it.

Fixes #5532

### DIFF
--- a/libqtile/backend/wayland/qw/xdg-view.c
+++ b/libqtile/backend/wayland/qw/xdg-view.c
@@ -103,8 +103,6 @@ static void qw_xdg_view_handle_commit(struct wl_listener *listener, void *data) 
     // On initial commit, set size and notify server to manage this view
     if (xdg_view->xdg_toplevel->base->initial_commit) {
         wlr_xdg_toplevel_set_size(xdg_view->xdg_toplevel, 0, 0);
-        xdg_view->base.title = xdg_view->xdg_toplevel->title;
-        xdg_view->base.app_id = xdg_view->xdg_toplevel->app_id;
         if (xdg_view->decoration != NULL) {
             qw_xdg_view_handle_decoration_request_mode(&xdg_view->decoration_request_mode,
                                                        xdg_view->decoration);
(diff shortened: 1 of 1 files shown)


elParaguayo:
@jwijenbergh @richcarni can `title` or `app_id` be `NULL`? If so, we might need to guard against that as I'm not sure how strdup behaves with NULL pointers. 

Also, I suspect I'd now need to free `title` and `app_id` when we destroy the view as we've now allocated memory to them. Is that right?

richcarni:
Probably safest to assume they could be NULL. And yes, `strdup` requires freeing memory

That said, something feels off here. Something freeing `xdg_view->xdg_toplevel->title` seems like a bug

BTW, I realise I keep putting issues in the todo column - I think I should move them back to deal with after merging, because I don't think issues like this one should prevent us from merging

richcarni:
Updating base.tile and base.app_id seems to fix the issue. Maybe no need to set them in initial commit at all?

```
static void qw_xdg_view_handle_map(struct wl_listener *listener, void *data) {
    struct qw_xdg_view *xdg_view = wl_container_of(listener, xdg_view, map);
    xdg_view->mapped = true;

    struct wlr_xdg_surface *surface = xdg_view->xdg_toplevel->base;
    struct wlr_box geom = surface->geometry;
    xdg_view->base.width = geom.width;
    xdg_view->base.height = geom.height;
    
    // Update these, as they may have changed since initial commit
    xdg_view->base.title = xdg_view->xdg_toplevel->title;
    xdg_view->base.app_id = xdg_view->xdg_toplevel->app_id;
```

elParaguayo:
OK. I've reverted my initial commit and pushed yours (also removing the title/app_id setting in initial commit).

If this works (seems to on my system) then I'll flatten the commits so we can merge.

elParaguayo:
> Probably safest to assume they could be NULL. And yes, `strdup` requires freeing memory
> 
> That said, something feels off here. Something freeing `xdg_view->xdg_toplevel->title` seems like a bug
> 
> BTW, I realise I keep putting issues in the todo column - I think I should move them back to deal with after merging, because I don't think issues like this one should prevent us from merging

I actually do think this issue is important enough to fix before releasing. We don't know how many apps could be affected.

richcarni:
Looks good on my system too!

I reckon what was happening was that particular client, gnome-terminal, was changing the title after commit - that could leave base.title referencing freed memory

PR #5542, opened 2025-10-06
[wayc] Complete xwayland implementation

richcarni:
Complete at least in comparison to xdg-view and old backend

This includes:
- Adding remaining event handlers, for maximize, minimize, etc
- Removing unused/commented code
- Moving set_tile and set_class listener creation to after view is managed


richcarni:
Not sure about the ci failure, can't reproduce on my system

elParaguayo:
I'll rerun the jobs and see if they fail in the same place again.

jwijenbergh:
Green now and seems like a big improvement

PR #5544, opened 2025-10-08
[wayc] Fix xdg-view clipping bug

richcarni:
Fixes windows (like discord) not being clipped


elParaguayo:
Thanks. Confirmed that this fixes the issue.

PR #5548, opened 2025-10-09
[wayc] Fix pointer focus when pointer hasn't moved

### LINKED ISSUE #5547: [wayc] Switching windows with keys not working correctly
### Issue description

With `follow_mouse_focus = True`

Open a couple of terminals side by side in Columns layout
`layout.left()` and `layout.right()` will switch between the two windows

Move the mouse
Now `layout.left()` and `layout.right()`fails to switch between the windows

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python
follow_mouse_focus = True
```

### Logs

```text

```

### Required

richcarni:
If the pointer hasn't moved but windows have changed, we need a different focus behaviour than for cursor motion. Both on wayc and python sides

To this end I have removed a function to only update pointer without motion from `qw_cursor_process_motion()` so it can be called independently. And added a motion parameter to `core._focus_pointer()` to modify its behaviour

Added a check to only call `wlr_seat_pointer_notify_enter()` if the surface under the pointer has changed. Maybe get a small performance gain here

And some refactoring

Fixes #5547

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -34,8 +34,10 @@ void qw_cursor_update_focus(struct qw_cursor *cursor, struct wlr_surface **surfa
         wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, "default");
         wlr_seat_pointer_clear_focus(seat);
     } else {
-        // TODO: May be a performance benefit by only calling notify_enter when the surface changes
-        wlr_seat_pointer_notify_enter(seat, tmp_surface, tmp_sx, tmp_sy);
+        struct wlr_surface *prev_surface = seat->pointer_state.focused_surface;
+        if (tmp_surface != prev_surface) {
+            wlr_seat_pointer_notify_enter(seat, tmp_surface, tmp_sx, tmp_sy);
+        }
     }
 
     // Return via output parameters, if provided


libqtile/backend/wayland/qw/cursor.c
```diff
@@ -19,6 +19,36 @@ void qw_cursor_destroy(struct qw_cursor *cursor) {
...
+    struct wlr_seat *seat = cursor->server->seat;
+
+    struct wlr_surface *tmp_surface = NULL;
+    double tmp_sx = 0.0, tmp_sy = 0.0;
+
+    cursor->view = qw_server_view_at(cursor->server, cursor->cursor->x, cursor->cursor->y,
+                                     &tmp_surface, &tmp_sx, &tmp_sy);
+
+    // Reset cursor if there's no view and we're not dragging
+    if ((!cursor->view || !tmp_surface) && cursor->server->seat->drag == NULL) {
```
richcarni, line 33: Can someone double check this logic with me? (I haven't changed it)

If either view or surface is NULL and drag is NULL then reset cursor and don't notify enter - OK

But what if surface is NULL and drag is not NULL? It's going to try and notify enter with a NULL surface

Am I interpreting this correctly

elParaguayo: I did it like this:
```diff
diff --git a/libqtile/backend/wayland/qw/cursor.c b/libqtile/backend/wayland/qw/cursor.c
index 88d20260..00a1de55 100644
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -63,8 +63,10 @@ static void qw_cursor_process_motion(struct qw_cursor *cursor, uint32_t time) {
                                 (int)cursor->cursor->y);

     // Reset cursor if there's no view and we're not dragging
-    if ((!view || !surface) && cursor->server->seat->drag == NULL) {
-        wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, "default");
+    if (surface != NULL) {
+        if (cursor->server->seat->drag == NULL) {
+            wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, "default");
+        }
         wlr_seat_pointer_clear_focus(seat);
         return;
     }
```

However, I've tested your code and I also don't get the crash when dragging.

Gurjaka:
fixes: https://github.com/qtile/qtile/issues/5561

PR #5550, opened 2025-10-11
Update python 3.14 version in CI

elParaguayo:
Python 3.14 has been released so let's use release version in CI.

@tych0 If we add branch protections for this now we'll need to rebase every open PR. How do you feel about that?


tych0:
Rebasing isn't a big deal IMO, I'd say feel free to add branch protections. Thanks for doing this!

PR #5551, opened 2025-10-11
[wayc] Stop screen_change hook on VT change

elParaguayo:
This fixes bug with the bar being killed when switching VT.

@jwijenbergh @richcarni are there any other bits of code that should be paused when switching to a different VT?

### DIFF
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -202,7 +202,13 @@ static void qw_server_handle_output_layout_change(struct wl_listener *listener, 
     }
 
     wlr_output_manager_v1_set_configuration(server->output_mgr, config);
-    server->on_screen_change_cb(server->cb_data);
+
+    // Only trigger screen change callback if the session is active
+    // or if session is NULL (i.e. in a nested or headless session)
+    // This prevents the hook firing on VT changes
+    if (server->session == NULL || server->session->active) {
+        server->on_screen_change_cb(server->cb_data);
+    }
 }
 
 // Reconfigure output(s) according to the provided configuration.


jwijenbergh:
What happens with other hooks like focus changes?

elParaguayo:
> What happens with other hooks like focus changes?

No idea! Haven't checked them.

However, a bigger issue seems to be that this stops qtile launching in CI..

elParaguayo:
OK - I know why this breaks: `server->session` seems to be `NULL` for nested sessions (same issue with `wephyr`) so `server->session->active` is undefined.

libqtile/backend/wayland/qw/server.c
```diff
@@ -202,7 +202,13 @@ static void qw_server_handle_output_layout_change(struct wl_listener *listener,
     }
 
     wlr_output_manager_v1_set_configuration(server->output_mgr, config);
-    server->on_screen_change_cb(server->cb_data);
+
+    // Only trigger screen change callback if the session is active
+    // or if session is NULL (i.e. in a nested or headless session)
+    // This prevents the hook firing on VT changes
+    if (server->session == NULL || (server->session != NULL && server->session->active)) {
```
richcarni: I think this could be simplified?
```
if (server->session == NULL || server->session->active)
```

richcarni:
I really can't think where else this might be an issue though. I guess something to be watchful for

libqtile/backend/wayland/qw/server.c
elParaguayo: Yeah. It's obvious now I see it.

elParaguayo:
OK - so let's merge this for now and we can make changes later if we need to.

Issue #5554, opened 2025-10-11
[wayc] Clipboard sync


Gurjaka:
Clipboard does not sync between Xwayland and native Wayland apps.

elParaguayo:
Can you see if #5568 fixes this please

Gurjaka:
> Can you see if [#5568](https://github.com/qtile/qtile/pull/5568) fixes this please

I am not sure if the PR actually fixed it, but seems to work now

PR #5568, opened 2025-10-12
[wayc] Fix Xwayland clipboard support

elParaguayo:
Might need to change the title of this as it might be a fix for other xwayland issues too.


Gurjaka:
fixes https://github.com/qtile/qtile/issues/5554

elParaguayo:
Ok - let me rename the commit as it's not clear what we're actually doing here.

I need to pop out for a bit so will do later.

PR #5569, opened 2025-10-12
[wayc] Set borders to NULL after free

elParaguayo:
This seems to be responsible for some crashes in xwayland windows and probably elsewhere.


richcarni:
Makes sense

Tried to reason how we could get here with `view->borders` freed. I suppose something like `place()` being called after `unmap()`? Doesn't really matter - that's what the guard is for

Issue #5557, opened 2025-10-11
[wayc] Conky window behavior


Gurjaka:
**Opened this issue because Conky is `Xwayland`, so something must be broken on `Xwayland` side**

Conky does not respect its set window parameters. See screenshots below for reference

Old backend:
<img width="2559" height="1439" alt="Image" src="https://github.com/user-attachments/assets/ddad8388-6277-4f74-9134-9e962d643386" />

New backend:
<img width="2560" height="1920" alt="Image" src="https://github.com/user-attachments/assets/a2646e9e-8caa-4f98-9d56-642f19975d7e" />

> Note: Had to take this pic from phone since `flameshot` kept crashing...

elParaguayo:
My guess is that we're tiling this because `has_fixed_size` is not implemented for xwayland so it doesn't satisfy any of our float rules.

Is that a floating window in your first screenshot? or is it rendered as part of the background?

elParaguayo:
Setting:
```
out_to_wayland yes
out_to_x no
```
renders the window in the background but it doesn't update...

elParaguayo:
From my further testing, conky does render correctly. The above settings will render conky to the `LAYER_BOTTOM` layer.

I'll leave this open until you have confirmed it's working for you.

Gurjaka:
> Setting:
> 
> ```
> out_to_wayland yes
> out_to_x no
> ```
> 
> renders the window in the background but it doesn't update...

adding these parameters displays conky like this:

<img width="477" height="1184" alt="Image" src="https://github.com/user-attachments/assets/043b20b8-5e3c-4cf3-9c53-d75388e5275e" />

conky output:
```bash
conky: FOUND: wayland
conky: FOUND: x11
conky: FOUND: file
conky: FOUND: ncurses
conky: FOUND: console
conky: 'unknown' wayland session running
```

elParaguayo:
OK - so we've lost the background colour? Not sure if that's a conky issue or a qtile one.

Gurjaka:
> OK - so we've lost the background colour? Not sure if that's a conky issue or a qtile one.

Might be conky issue, but still I didn't have those settings set before, and used to work as expected on old backend

elParaguayo:
Yes - I appreciate that. Not sure why that was the case. Are you able to run the old backend with your old conky settings and do an `xprop` on the conky window?

[...]

PR #5572, opened 2025-10-13
[wayc] Add xwayland fixed-size check

richcarni:
Should address part of #5557


elParaguayo:
Code looks good (I actually wrote the same code) but it doesn't fix #5557 as conky doesn't set size hints at all.

I'm still ok to merge this though.

PR #5576, opened 2025-10-14
fix: cleanup borders typo

Gurjaka:
If borders doesn't exist return, else clean up.


elParaguayo:
Thanks. Yes. This was a stupid mistake by me. I was tidying up `if (!view->borders)` and turned it into `if (view->borders != NULL)`.

Gurjaka:
> Thanks. Yes. This was a stupid mistake by me. I was tidying up `if (!view->borders)` and turned it into `if (view->borders != NULL)`.

Haha, happens :)

PR #5577, opened 2025-10-14
[wayc] Fix potential null pointer dereference

richcarni:
We should never call pointer_notify_enter if surface is null

Also tidied up some cruft

Completes #5548


richcarni:
@elParaguayo This was intended to be the last commit of #5548

jwijenbergh:
Thanks!

Issue #5549, opened 2025-10-10
[wayc] Implement implicit grab


richcarni:
### Issue description

> A Wayland client expects to receive pointer events from the moment a
    pointer button is pressed on its surface until the moment the button is
    released.  The Wayland protocol leaves this behavior to the compositor.

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #5578, opened 2025-10-15
[wayc] Implement implicit grab

richcarni:
Addresses #5549

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.h
+++ b/libqtile/backend/wayland/qw/cursor.h
@@ -6,10 +6,17 @@
 
 struct qw_server; // Forward declaration to avoid circular dependency
 
+struct qw_implicit_grab {
+    double start_dx;
+    double start_dy;
+    bool live;
+};
+
 // Cursor structure that holds cursor state and event listeners
 struct qw_cursor {
     struct wlr_cursor *cursor;
     struct qw_view *view;
+    struct qw_implicit_grab implicit_grab;
 
     // private data
     struct qw_server *server;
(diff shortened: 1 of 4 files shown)


libqtile/backend/wayland/qw/cursor.h
```diff
@@ -6,12 +6,18 @@
...
+    double start_dy;
+};
+
 // Cursor structure that holds cursor state and event listeners
 struct qw_cursor {
     struct wlr_cursor *cursor;
     struct qw_view *view;
 
     // private data
+    struct qw_implicit_grab *implicit_grab;
```
richcarni: Given such a small struct now, feels like better to pass by value with additional `bool active` or `bool live` member

richcarni: Made this change

richcarni:
@elParaguayo @jwijenbergh This is ready for review/merge

elParaguayo:
Is the purpose of this that motion events are sent to the same "grabbed" client, even if the cursor leaves the window?

It looks ok but I was just wondering if we need to do anything about the locked state (which we do check in `qw_cursor_process_motion`.

richcarni:
>Is the purpose of this that motion events are sent to the same "grabbed" client, even if the cursor leaves the window?

Yes, exactly! For example if you've grabbed a scroll bar and your cursor wanders off into the adjacent window, you want to keep focus and scrolling rather than switch focus to that window

>It looks ok but I was just wondering if we need to do anything about the locked state (which we do check in qw_cursor_process_motion.

Good question! Do you think we need a similar check in `void qw_cursor_implicit_grab_motion()`? Maybe just release the grab if the screen is locked?

I don't know if this is possible/likely/serious, but it looks like if you locked the screen during an implicit grab, that it will continue processing that motion

elParaguayo:
Agreed. I would just release the grab if there's a lock.

elParaguayo:
I'm going to merge this for now and we can add lockscreen bit later if we need to.

PR #5579, opened 2025-10-16
[wayc] Add more xwayland window properties

elParaguayo:
Adds "instance" and "role".

"instance" is used to extend `get_wm_class()` so it now returns the equivalent result to X11 being a list of `[instance, class]`.

"role" is a string such as `GtkFileChooserDialog` which can be used by users for floating rules.


elParaguayo:
This PR shouldn't be merged yet. I should add `wm_type` support...

elParaguayo:
I'm actually inclined to merge this now as it adds useful functionality. We can add wm_type etc as a todo.

jwijenbergh:

Issue #5519, opened 2025-09-28
[wayc] startup_once hook fires too soon


elParaguayo:
I think the bug is in the wayland backend: the xwayland code starts up in `Core.__init__`, but is not necessarily ready. Once we do `on_config_load()`, we execute all the `startup_once` hooks, which is what dgroups uses to spawn these group arguments. If that happens before `on_xwayland_ready` fires, I suspect things are broken.

I was going to wait until `wayc` was landed to try and fix it, but maybe you guys can fix it in the process of doing `wayc`. It seems like we need to defer running `startup_once` hooks until everything is actually ready in wayland land. I don't know exactly what "everything" is, but it looks like that 1. there's an output present 2. xwayland is ready at the very least...

_Originally posted by @tych0 in https://github.com/qtile/qtile/issues/5518#issuecomment-3343728865_

richcarni:
I don't know if it's related but I'm still getting flaky test failures after #5518

I can't quite figure out what is going on?

```
=================================== FAILURES ===================================
________________________ test_reload_config[1-wayland] _________________________

manager_nospawn = <test.helpers.TestManager object at 0x7f17c4a2d460>

    def test_reload_config(manager_nospawn):
        # The test config uses presence of Qtile.test_data to change config values
        # Here we just want to check configurables are being updated within the live Qtile
        manager_nospawn.start(lambda: BareConfig(file_path=configs_dir / "reloading.py"))
    
        @Retry(ignore_exceptions=(AssertionError,))
        def assert_dd_appeared():
            assert "dd" in manager_nospawn.c.group.info()["windows"]
    
        # Original config
        assert manager_nospawn.c.eval("len(self.keys_map)") == (True, "1")
        assert manager_nospawn.c.eval("len(self._mouse_map)") == (True, "1")
        assert "".join(manager_nospawn.c.get_groups().keys()) == "12345S"
        assert len(manager_nospawn.c.group.info()["layouts"]) == 1
        assert manager_nospawn.c.widget["clock"].eval("self.background") == (True, "None")
        screens = manager_nospawn.c.get_screens()[0]
        assert screens["gaps"]["bottom"][3] == 24 and not screens["gaps"]["top"]
        assert len(manager_nospawn.c.internal_windows()) == 1
        assert manager_nospawn.c.eval("self.dgroups.key_binder") == (True, "None")
        assert manager_nospawn.c.eval("len(self.dgroups.rules)") == (True, "6")
        manager_nospawn.test_window("one")
[...]

PR #5580, opened 2025-10-17
[wayc] Set WAYLAND_DISPLAY before loading config

richcarni:
Necessary for startup and startup_once hooks to launch wayland programs

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -248,6 +248,7 @@ class Core(base.Core):
         self.qw.focus_current_window_cb = lib.focus_current_window_cb
         self.qw.on_session_lock_cb = lib.on_session_lock_cb
         lib.qw_server_start(self.qw)
+        os.environ["WAYLAND_DISPLAY"] = self.display_name
         self.qw_cursor = lib.qw_server_get_cursor(self.qw)
 
         self.painter = Painter(self)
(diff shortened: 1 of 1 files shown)


richcarni:
May partly address issues with apps launched at startup silently falling back to X11 (I say partly because xwayland apps should also run without issues)

Possibly relates to #5519?

libqtile/backend/wayland/core.py
```diff
@@ -574,7 +575,6 @@ def setup_listener(self) -> None:
         self.fd = lib.qw_server_get_event_loop_fd(self.qw)
         if self.fd:
             asyncio.get_running_loop().add_reader(self.fd, self._poll)
-            os.environ["WAYLAND_DISPLAY"] = self.display_name
```
richcarni, line 577: probably a @jwijenbergh question, am I overlooking a critical reason the display is set here?

jwijenbergh: I honestly don't remember why I added it here. But if your patch fixes the issue and you don't notice anything broken then LGTM

Gurjaka:
Thanks for this! I had Discord and Firefox on startup, and they both used to start as Xwayland, this fixes it

tych0:
Yeah, my guess is that due to #5519 we will still race sometimes, though both firefox and discord are probably slow enough at starting that they don't.

elParaguayo:
Well, I'm not seeing any drawbacks to merging this...

PR #5581, opened 2025-10-17
[wayc] Limit callbacks when scrolling

### LINKED ISSUE #5565: [wayc] Scrolling behavior feels off
Scrolling starts slow but accelerates unnaturally over time.

elParaguayo:
Currently, the wayc compositor sends a button press callback for every single scroll axis event. This triggers far more callbacks than are needed, resulting in high CPU usage/laggy scroll behaviour.

This PR introduces a limitation on how often callbacks should be triggered.

Fixes #5565

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -195,27 +195,33 @@ static void qw_cursor_handle_axis(struct wl_listener *listener, void *data) {
     bool handled = false;
     // TODO: exclusive client and implicit grab
 
-    if (event->delta != 0) {
-        int dir = (event->delta > 0) ? 1 : -1;
-        enum wl_pointer_axis axis = event->orientation;
-        uint32_t elapsed = event->time_msec - last_scroll_time;
+    // Determine which button this corresponds to
+    uint32_t button = 0;
+    if (event->orientation == WL_POINTER_AXIS_VERTICAL_SCROLL) {
+        button = (event->delta > 0) ? BUTTON_SCROLL_DOWN : BUTTON_SCROLL_UP;
+    } else if (event->orientation == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
+        button = (event->delta > 0) ? BUTTON_SCROLL_RIGHT : BUTTON_SCROLL_LEFT;
+    }
 
-        // Trigger callback if enough time has passed or direction/axis changed
-        if (elapsed >= QW_SCROLL_CALLBACK_INTERVAL_MS || dir != last_scroll_dir ||
-            axis != last_scroll_axis) {
-            last_scroll_time = event->time_msec;
-            last_scroll_dir = dir;
-            last_scroll_axis = axis;
+    uint32_t button_mapped = qw_util_get_button_code(button);
 
-            uint32_t button = 0;
-            if (axis == WL_POINTER_AXIS_VERTICAL_SCROLL) {
-                button = (dir > 0) ? BUTTON_SCROLL_DOWN : BUTTON_SCROLL_UP;
-            } else {
-                button = (dir > 0) ? BUTTON_SCROLL_RIGHT : BUTTON_SCROLL_LEFT;
-            }
+    // If it's a physical wheel fire callbacl immediately if there is a discrete delta
+    if (event->source == WL_POINTER_AXIS_SOURCE_WHEEL && event->delta_discrete != 0.0) {
+        handled = qw_cursor_process_button(cursor, button_mapped, true);
+        // for anything else, we're using rate limiting
+    } else if (event->source != WL_POINTER_AXIS_SOURCE_WHEEL) {
+        // Touchpad or smooth scroll: throttle events to 50ms
+        int dir = (event->delta > 0) ? +1 : -1;
+        uint32_t now = event->time_msec;
 
-            uint32_t button_mapped = qw_util_get_button_code(button);
+        if ((now - last_scroll_time >= QW_SCROLL_CALLBACK_INTERVAL_MS) ||
+            (dir != last_scroll_dir) || (event->orientation != last_scroll_axis)) {
+
             handled = qw_cursor_process_button(cursor, button_mapped, true);
+
+            last_scroll_time = now;
+            last_scroll_dir = dir;
+            last_scroll_axis = event->orientation;
         }
     }
 


elParaguayo:
Question for @Gurjaka @richcarni @jwijenbergh should the interval be something that users can configure as opposed to being hardcoded?

At the moment I've just hardcoded a value that applies to all cursors but could/should we add it to `InputConfig` instead?

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -195,27 +195,33 @@ static void qw_cursor_handle_axis(struct wl_listener *listener, void *data) {
...
+    uint32_t button_mapped = qw_util_get_button_code(button);
 
-            uint32_t button = 0;
-            if (axis == WL_POINTER_AXIS_VERTICAL_SCROLL) {
-                button = (dir > 0) ? BUTTON_SCROLL_DOWN : BUTTON_SCROLL_UP;
-            } else {
-                button = (dir > 0) ? BUTTON_SCROLL_RIGHT : BUTTON_SCROLL_LEFT;
-            }
+    // If it's a physical wheel fire callbacl immediately if there is a discrete delta
+    if (event->source == WL_POINTER_AXIS_SOURCE_WHEEL && event->delta_discrete != 0.0) {
```
richcarni, line 209: just a niggle, event->delta_discrete is an int

elParaguayo: Yes. That was a hangover from a previous attempt where I thought it gave me fractions of a step.

elParaguayo:
@jwijenbergh @richcarni So, unfortunately, I'm on holiday somewhere where Discord is banned so I can't access the wayc development chat. I don't have any messages since the once I sent about widgets still receiving lots of callbacks.

@jwijenbergh I know you don't want an arbitrary rate limit and I understand that so I'm open to ideas (and happy to close this if you want to try something else) but, fundamentally, I don't see a need for us to ever need every single axis event via a callback.

ps if you need my input on anything, it will have to be via github comments this week!

richcarni:
You would have missed my reply then, which was basically:
1. I agree with Jeroen re: "Isn't it better if we solve the TODO? That c takes control of the key list". I'm not really working on anything at the moment so I might start looking into this (create an issue #5584)

2. However, if this making some systems unusable and thereby impacting testing, I'd support either a) merging this PR with the caveat that we come back and look at it more closely later or b) just disable axis callbacks altogether, can again come back and deal with it later

On the PR, my feeling is throwing away scroll events is probably not the correct way to handle a high resolution scroll device. But I also don't know or understand enough about scrolling, which certainly seems more complicated than I had thought, to say what we should be doing. For example, is the trackpad giving us a delta that says how far we have moved since the last event? If that's the case, should we be accumulating in c until we reach some threshold, at which point we fire a single callback? Essentially converting the continuous trackpad motion into discrete steps

richcarni:
Since I mentioned delta, did a quick experiment on my laptop trackpad to see what it's doing:
```
2025-10-20 10:50:14,228 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: 0.743203
2025-10-20 10:50:14,236 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: 0.457356
2025-10-20 10:50:14,244 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: 0.228678
2025-10-20 10:50:14,292 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: 0.000000
2025-10-20 10:50:14,292 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: 0.000000
2025-10-20 10:50:15,312 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: -2.000931
2025-10-20 10:50:15,319 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: -2.629795
2025-10-20 10:50:15,328 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: -1.772253
2025-10-20 10:50:15,335 ERROR qw core.py:log_cb():L125 [cursor.c:194] delta: -2.401117
```

In contrast, this is for my old school clicky wheel mouse:
```
2025-10-20 11:00:39,931 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: 15.000000
2025-10-20 11:00:39,977 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: 15.000000
2025-10-20 11:00:40,043 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: 15.000000
2025-10-20 11:00:40,089 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: 15.000000
2025-10-20 11:00:40,775 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: -15.000000
2025-10-20 11:00:40,820 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: -15.000000
2025-10-20 11:00:40,853 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: -15.000000
2025-10-20 11:00:40,886 ERROR qw core.py:log_cb():L125 [cursor.c:206] delta: -15.000000
```

richcarni:
#5585
I have a PR for testing that only calls `cursor_button_cb()` if either
- there is a binding for that button+modifier
- we are over an internal window
- we need to focus the window

If this is working correctly it should greatly reduce CPU usage while scrolling (in most circumstances)

elParaguayo:
Yes. But you'll flood widgets with scroll events which is bad for, say, volume control.

This issue doesn't arise on x11 so I don't think widget specific code should be used to manage this.

richcarni:
I think we can support both smooth and discrete (clicky) scrolling in widgets

It seems like widgets currently expect discrete button clicks for scrolling. To emulate this behaviour with a trackpad (or high res scroll wheel) I think we should accumulate deltas, and fire a button click when that passes some threshold (+-/15? but may be configurable)

On top of that we could have a separate scroll event that widgets could use to receive each delta - like a single axis pointer motion. This could be used to implement smooth scrolling

elParaguayo:
If we send any smooth scrolling events to the python backend we're going to hit the same issue. At the moment, I can't think of any reason why we'd even need to but let me know if you can.

elParaguayo:
Do we get axis events more frequently than cursor motion ones? That would seem odd...

Just thinking that we have a callback in `qw_cursor_process_motion` and, having checked this, moving cursor lots does push up the CPU too (about 45-50% on my laptop).

richcarni:
>Do we get axis events more frequently than cursor motion ones? That would seem odd...

Ah! This is the thing right. Presumably on a trackpad it's literally exactly the same thing so I'd expecting polling/events to be the same frequency

I've been looking at what handle button entails, with a view to what could be moved over to c. In short:
- `_focus_by_click()`
- `qtile.process_button_click()`
- if over a internal window -> call that windows `process_button_click()`

I'm sure there must be some optimisation here. Like if the view hasn't changed, no need to `_focus_by_click()`

I'm not sure moving much of this over to c is feasible, because of how many qtile functions and properties they interact with

richcarni:
https://github.com/richcarni/qtile/blob/72dfa82a049c1100cfff6d66e467e2e660886532/libqtile/backend/wayland/qw/cursor.c#L213-L227

I think for compatibility we still need to convert scroll motion into discrete button events. We can always add smooth scrolling support later if that is something people want

I wrote a little test for how I would go about this. Would need to track horizontal and vertical displacement separately, but I'm sure you get the idea

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -188,19 +188,41 @@ static void qw_cursor_handle_axis(struct wl_listener *listener, void *data) {
...
+    uint32_t button = 0;
+    if (event->orientation == WL_POINTER_AXIS_VERTICAL_SCROLL) {
+        button = (event->delta > 0) ? BUTTON_SCROLL_DOWN : BUTTON_SCROLL_UP;
+    } else if (event->orientation == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
+        button = (event->delta > 0) ? BUTTON_SCROLL_RIGHT : BUTTON_SCROLL_LEFT;
+    }
+
+    uint32_t button_mapped = qw_util_get_button_code(button);
+
+    // If it's a physical wheel fire callbacl immediately if there is a discrete delta
```
jwijenbergh, line 208: typo

libqtile/backend/wayland/qw/cursor.c
jwijenbergh: could this be fixed? If so let's just merge and merge the wayc branch

elParaguayo: Should we make the change that @richcarni suggested in his last comment here?

elParaguayo:
@jwijenbergh Do we want to make rich's suggestion above?

Let me know what you think and I'll try to sort today.

jwijenbergh:
to be honest, I think we can deal with all these enhancements after we merge. It's already in a usable-ish state

elParaguayo:
You want to merge then?

jwijenbergh:
the wayc branch sure, and this branch too if we rebase

elParaguayo:
Is there an easy way to rebase this onto wayc given that wayc has had forced pushes? I shouldn't need to resolve all the conflicts showing here.

Close this and cherry-pick commits onto a new branch? Feels like the should be some git commands to sort the existing branch.

jwijenbergh:
Maybe indeed just cherry-pick

richcarni:
We’re going to merge the wayc branch imminently? So we should withhold any new PRs to wayc?

jwijenbergh:
> We’re going to merge the wayc branch imminently? So we should withhold any new PRs to wayc?

It makes sense to me to merge it in the current state, a lot of the things left to do I am confident we can tackle it before the next release. You can still make PRs, no problem

Issue #5582, opened 2025-10-18
[wayc] Add pointer gesture support


elParaguayo:
Not needed for release but should be added

elParaguayo:
@richcarni @jwijenbergh I've got some basic code working in wayc that can track swipes (including sequences) and pinches (including rotations).

My plan would be to record the gesture and trigger the callback at the end.

The bit I'm struggling with is what to do when clients can also react to these gestures. There are methods for forwarding the beginning, update and end events. We can check if a gesture *may* be handled by qtile when it starts, but we won't *know* if it's handled until the gesture is complete. So, do we not forward events if it could be handled?

richcarni:
Interesting. I have run into this with IOS before. If I remember and have looked this up correctly, the way IOS handles this is to add a slight delay between the touch event and beginning the gesture. In that time, as soon as you know it's not a gesture, you go and process the event as normal, so every touch isn't laggy. Else you use the touch plus any motion to determine it is, say a three finger drag, and then forward to the correct handler

So it means you have to figure out what the gesture is before it's complete. That is probably doable for most common gestures?

elParaguayo:
On a trackpad, wlroots already has events for swipe and pinch so we don't need to do anything to decide if these gestures are being triggered or not.

I've created a PR and will do some more testing. There are also some bits of that PR that I don't like that I need to think about how to change.

elParaguayo:
Touch screens are, I think, different and we might need the approach that you suggest. My understanding is that it's a touch device (but a touchpad isn't?!). So, we need to track each touch and then decide if we're swiping, pinching etc. That would introduce a slight delay as you say.

I'm still looking into it but, as my laptop doesn't have a touch screen, I can't really do anything to develop/test it!

richcarni:
My comment was in relation to your problem:
>We can check if a gesture may be handled by qtile when it starts, but we won't know if it's handled until the gesture is complete. So, do we not forward events if it could be handled?

The only thing I can think of is to try and identify if the gesture is handled just after it has begun

elParaguayo:
The easiest idea is to do this:
* when a gesture starts, check to see whether the user has a binding for that gesture type (e.g. just a pinch or swipe) with any modifiers
* if they do, we swallow the event, even if the gesture ultimately ends up not matching (e.g. the user has only bound swipe left and it was a right swipe)
* if there's no match, we forward events so clients can listen for them

This allows us to do things like pinch zooming in google maps etc.

[...]

PR #5583, opened 2025-10-19
[wayc] Add touch device support and pointer gestures

elParaguayo:
Add support for swipe and pinch gestures.

Draft for now as I have some questions on this. See #5582.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -137,6 +137,7 @@ cdef_files = [
     "input-device.h",
     "keyboard.h",
     "pointer.h",
+    "touch.h",
 ]
 
 for file in cdef_files:
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -15,6 +15,7 @@
 #include "pointer.h"
 #include "server.h"
 #include "session-lock.h"
+#include "touch.h"
 #include "util.h"
 #include "view.h"
 #include "wayland-server-core.h"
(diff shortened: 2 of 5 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -122,11 +102,11 @@ def log_cb(importance: int, formatted_str: ffi.CData) -> None:
     """Callback that logs the string at the given level"""
     log_str = ffi.string(formatted_str).decode()
     if importance == lib.WLR_ERROR:
-        qw_logger.error(log_str)
+        logger.error(log_str)
```
elParaguayo, line 83: Ugh. I keep changing this to help with debugging. Really should just submit a PR once and for all.

libqtile/config.py
```diff
@@ -229,6 +229,94 @@ def __repr__(self) -> str:
         return f"<Click ({self.modifiers}, {self.button})>"
 
 
+class Gesture:
+    def __init__(self, modifiers: list[str], *commands: LazyCall) -> None:
+        self.modifiers = modifiers
+        self.commands = commands
+        self.modmask = 0
```
elParaguayo, line 209: I don't love how the `Swipe` and `Pinch` gestures are currently configured.

I'm trying to keep `commands` consistent with other config variables like `lazy` bindings as well as `Click` and `Drag`. In those objects, `commands` is the last argument and can be any number of commands.

However, for the gestures there could be optional arguments like number of fingers (to be added). Python doesn't allow optional positional arguments so, to do this, I have three options:
1) Make `commands` a keyword argument but multiple commands would need to be added as a list.
2) Force gesture config options to be keywords and keep `commands` as positional. This changes the ordering compared to those other config variables.
3) Set optional config values via methods. e.g. `Pinch([], lazy.sceen.next_group().shrink()`

I've gone for number 3 at the moment but would appreciate thoughts.

elParaguayo:
Was going to add touch device support too. Will keep in this PR as I might need to refactor gesture code so we can also use it for touch gestures.

elParaguayo:
The touch support is just proof of context for now. Opening something like Google Maps, touching the screen and moving your finger should now move the map.

There's lots still to add (drag'n'drop, explicit grab and gestures).

elParaguayo:
Making good progress on this but it's not needed before we merge wayc into main.

jwijenbergh:

Issue #5584, opened 2025-10-20
[wayc] Move input events "to grab" decision to c land


richcarni:
### Issue description

As it stands, every key press or mouse action, calls a python handler to determine if it should be grabbed by qtile rather than passed to the active client

We are seeing that this can unnecessarily spike qtile's CPU utilisation when, for example, while scrolling with a trackpad outside of a widget

As per the keyboard related TODO:
>// TODO: for efficiency maybe let c take control of the key list?

There is likely even more efficiency to gain by doing something similar for cursor events

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #5585, opened 2025-10-20
[wayc] Filter grab buttons in c

richcarni:
Filter grab button events in c and only call cursor_button_cb() if either:
- there is a binding for that button+modifier
- we are over an internal window
- we need to focus the window

This improves efficiency by minimising unnecessary python callbacks

issue #5584


jwijenbergh:

jwijenbergh:
Do you want to rebase this?

richcarni:
It was only 3 weeks ago but reading the code feels more like 3 years!

This was a bit of an experiment and I think I reached the conclusion that it added a lot of complexity for minimal gain. And we fixed the associated issue elsewhere

I might have another go at it at some stage - I think I can do better

PR #5590, opened 2025-10-23
docs(widgets): explain how to modify text attributes correctly

RFCreate:
This PR expands the documentation to explain why setting the widget's `font`, `fontsize`, `fontshadow`, `foreground` and `markup` parameters will not affect the text in the widget as of changes in #5325. And what users do instead to overcome this.

It also expands a currently implemented function (`set_font`) to accept all the parameters that modify the text layout to act as a setter for the layout attributes.

I write this as per discussion in #5467, so I would want to hear @saviola777 thoughts on this.


saviola777:
I like it, definitely an improvement!

elParaguayo:
I'm happy to merge this if there are no other changes needed.

RFCreate:
Not from my side.

PR #5598, opened 2025-10-25
Add option to load default config

elParaguayo:
When debugging issues, we often ask users to run the default config. This PR adds a command line option to load the default config instead of the user's normal config.

### DIFF
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -15,6 +15,9 @@ if TYPE_CHECKING:
     from libqtile.core.manager import Qtile
 
 
+LIBQTILE_PATH = Path(__file__).parent.parent
+
+
 def rename_process():
     """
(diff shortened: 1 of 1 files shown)


libqtile/scripts/start.py
```diff
@@ -95,14 +103,23 @@ def start(options):
...
         "-c",
         "--config",
         action="store",
         default=get_config_file(),
         dest="configfile",
         help="Use the specified configuration file.",
     )
+    configs.add_argument(
+        "-d",
+        "--default",
```
tych0: Does argparse render these config groups somehow? otherwise maybe `--use-default-config` instead, so it's clear what is defaulting to what?

elParaguayo: If you do `--help` it shows something like `[-c CONFIG | -d]`. However, `--use-default-config` is a better choice of wording.

Issue #5573, opened 2025-10-14
[wayc] xwayland app menus not being placed correctly


elParaguayo:
Open `leafpad` and click on a menu. The menu will be floated but placed in the center of the screen.

richcarni:
Is this by fixed transient_for support #5574?

elParaguayo:
Possibly. We'd need to detect that the menu is requesting a specific location and decide when we should respect that request (which could depend on the window type)

richcarni:
Ah, I think it's override_redirect that tells us it needs to be placed and is unmanaged. Got it in the right position but currently it crashes shortly after

<img width="1117" height="477" alt="Image" src="https://github.com/user-attachments/assets/a304b203-35ef-4b25-bff1-7e7104e56b4f" />

elParaguayo:
Well - that's still progress

elParaguayo:
@richcarni should this be closed now?

PR #5600, opened 2025-10-25
[wayland] Add xwayland override_redirect support

richcarni:
Xwayland surfaces where override_redirect is true should not be managed but instead placed at the given location. These surfaces are usually things like menus or popups

Follows sway's implementation pretty closely

Addresses #5573

Outstanding issues:
- Menu doesn't move if window is moved
- ~~If main xwayland window loses and regains focus, the menu goes beneath the main window~~
- Since these surfaces aren't  managed, they are visible on all groups

### DIFF
--- a/libqtile/backend/wayland/qw/xwayland-view.h
+++ b/libqtile/backend/wayland/qw/xwayland-view.h
@@ -10,7 +10,6 @@
 
 struct qw_xwayland_view {
     struct qw_view base;
-    struct qw_server *server;
 
     struct wlr_scene_tree *scene_tree;
     struct wlr_scene_surface *scene_surface;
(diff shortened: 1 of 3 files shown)


elParaguayo:
> Still broken:
> 
>     * Menu doesn't move if window is moved

Do we get any events for the surface when the parent is moved e.g. configure?

> 
>     * If main xwayland window loses and regains focus, the menu goes beneath the main window

Not surprised by this. As above, we never tell it go back above its parent. Also, focusing on a window can, I think, trigger `bring_to_front`.

I wonder if setting the `KEEP_ABOVE` property (part of my xwayland properties work) would help solve this.

richcarni:
>Do we get any events for the surface when the parent is moved e.g. configure?

I think it must be `events.set_geometry`

I'm looking at sway's implementation. They have separate functions/handlers for unmanaged xwayland views. I think I should follow this approach - a little bit of duplication but less conditionals make it easier to follow/reason (my opinion)

libqtile/backend/wayland/qw/xwayland-view.h
richcarni, line 13: Since base also holds a reference to server I think we should remove L13 and just use the base reference

We also have two references in layout-view

elParaguayo: Yes - I came across this in my window type PR. I kept getting segfaults when trying to access `xwayland_view->server` because it's never actually set!

elParaguayo:
Can confirm that xwayland menus display correctly with this.

@richcarni do you need to do anything else before we merge?

richcarni:
No it's ready, oops, forgot to make it not a draft

PR #5602, opened 2025-10-25
[wayc] Add urgency to xwayland windows

RFCreate:
This is a rewrite of an old PR of mine #5024, adapted to the new Wayland C backend.

### DIFF
--- a/test/test_hook.py
+++ b/test/test_hook.py
@@ -588,6 +588,9 @@ def test_client_urgent_hint_changed(manager_nospawn):
     manager_nospawn.start(ClientUrgentHintChangedConfig)
     manager_nospawn.test_window("Test Client", urgent_hint=True)
     assert_window(manager_nospawn, "Test Client")
+    # Get urgency of the window
+    _, urgent = manager_nospawn.c.eval("list(self.windows_map.values())[0].urgent")
+    assert urgent == "True"
 
 
 class CallLayoutGroup:
--- a/libqtile/backend/base/__init__.py
+++ b/libqtile/backend/base/__init__.py
@@ -1,7 +1,7 @@
 from libqtile.backend.base.core import Core  # noqa: F401
 from libqtile.backend.base.drawer import Drawer  # noqa: F401
+from libqtile.backend.base.float_states import FloatStates  # noqa: F401
 from libqtile.backend.base.window import (  # noqa: F401
-    FloatStates,
     Internal,
     Static,
     Window,
--- a/libqtile/backend/base/float_states.py
+++ b/libqtile/backend/base/float_states.py
@@ -1,0 +1,11 @@
+import enum
+
+
+@enum.unique
+class FloatStates(enum.Enum):
+    NOT_FLOATING = 1
+    FLOATING = 2
+    MAXIMIZED = 3
+    FULLSCREEN = 4
+    TOP = 5
+    MINIMIZED = 6
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
-import enum
 import typing
 from abc import ABCMeta, abstractmethod
+from types import FunctionType
 
+from libqtile import hook
 from libqtile.command.base import CommandError, CommandObject, expose_command
+from libqtile.log_utils import logger
+from libqtile.scratchpad import ScratchPad
 
 if typing.TYPE_CHECKING:
     from typing import Any
@@ -15,16 +18,6 @@ if typing.TYPE_CHECKING:
     from libqtile.core.manager import Qtile
     from libqtile.group import _Group
     from libqtile.utils import ColorsType
-
-
-@enum.unique
-class FloatStates(enum.Enum):
-    NOT_FLOATING = 1
-    FLOATING = 2
-    MAXIMIZED = 3
-    FULLSCREEN = 4
-    TOP = 5
-    MINIMIZED = 6
 
 
 class _Window(CommandObject, metaclass=ABCMeta):
@@ -96,9 +89,13 @@ class _Window(CommandObject, metaclass=ABCMeta):
         return False
 
     @property
-    def urgent(self):
+    def urgent(self) -> bool:
         """Whether this window urgently wants focus"""
         return False
+
+    @urgent.setter
+    def urgent(self, urgent: bool) -> None:
+        raise NotImplementedError
 
     @property
     def opacity(self) -> float:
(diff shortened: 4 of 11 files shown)


richcarni:
Thanks RF! I want to merge this before my PR to fix the menu position issue, since you got yours in first

Looks good, but because it touches some tests I'd like to get someone who knows the tests better than me to also review, maybe @tych0, who commented on the original PR - although that was 12 months ago! or @elParaguayo

I'm not sure I follow what is going on with GDK_BACKEND. Do we have to set it to x11 for this to pass tests?

test/test_hook.py
```diff
@@ -578,16 +576,12 @@ class ClientNameUpdatedConfig(BareConfig):
...
-    if backend_name == "wayland":
-        pytest.skip("Core not listening to XDG request_activate_event ?")
-
+def test_client_urgent_hint_changed(manager_nospawn):
     class ClientUrgentHintChangedConfig(BareConfig):
         groups = [
             config.Group("a"),
             config.Group("b", matches=[Match(title="Test Client")]),
         ]
-        focus_on_window_activation = "urgent"
```
tych0, line 590: why is this hunk needed?

elParaguayo:
Thanks for this. I was halfway through adding this but very happy to cross off my list!

elParaguayo:
As I understand it, the `GDK_BACKEND` changes are just to allow us to force the test window to be handled via xwayland. However, the changes mean that this only happens if we want to set the urgent status. All other test windows using the wayland backend will still be launched as xdg windows.

Code looks great to me but I'd like to see an answer to tych0's question before we merge.

test/test_hook.py
RFCreate: The wayland backend does not have it implemented yet, but in X11 if the window is the current window or in the current screen, it will not trigger.

https://github.com/qtile/qtile/blob/34300ce4d3012ed2ba40cd4c73aad61b2bdc6ca4/libqtile/backend/x11/window.py#L623-L625
https://github.com/qtile/qtile/blob/34300ce4d3012ed2ba40cd4c73aad61b2bdc6ca4/libqtile/backend/x11/window.py#L2184-L2187

elParaguayo: Shouldn't we add it to wayland then, as part of this backend?

RFCreate:
The `GDK_BACKEND=x11` is there because gtk windows in wayland do nothing when `set_urgency_hint` is called, so the hook will never trigger in the tests.

You can check the source code to probe it: https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gdk/wayland/gdkwindow-wayland.c

Maybe I should add the url reference in the changes, idk.

test/test_hook.py
RFCreate: Most likely.

tych0: Can we hoist this logic to the base backend somehow?

elParaguayo: Apologies - I'm missing something here. The hook is fired when urgency is set for xwayland windows as we trigger the callback. What's the actual issue we want to address with this bit?

RFCreate: > Apologies - I'm missing something here. The hook is fired when urgency is set for xwayland windows as we trigger the callback. What's the actual issue we want to address with this bit?

I think is to have consistency with x11 hooks.

> Can we hoist this logic to the base backend somehow?

Sure, but should it compare with current window or current screen? I am confused in that regard.

elParaguayo: Ok. This should be done in the python code. We just need to update this method:
```
    def handle_view_urgent(self, view: ffi.CData) -> None:
        """Handle view urgency notification"""
        assert self.qtile is not None
        wid = view.wid
        win = self.qtile.windows_map.get(wid)

        if win:
            # Mark window as urgent in Qtile
            # TODO: Fix the following line, probably supposed to be: win._urgent
            # win.urgent = True
            hook.fire("client_urgent_hint_changed", win)

            if self.qtile.config.focus_on_window_activation == "smart":
                win.focus(False)
```

RFCreate: True, in which way we update it?

elParaguayo: To be consistent with the X11 backend behaviour.

RFCreate: I know but would it be comparing to current window or current screen?

elParaguayo: You're allowed your own opinion here! i.e. what would be most useful to you? I don't deal with `urgency` at all so I'm largely indifferent.

If it were me, I'd probably compare to `self.qtile.current_window` - that gets the current window on the current screen. If you've got multiple screens but the urgent window is on an inactive screen, I'd expect the focus to change.

richcarni:
> The `GDK_BACKEND=x11` is there because gtk windows in wayland do nothing when `set_urgency_hint` is called, so the hook will never trigger in the tests.
> 
> You can check the source code to probe it: https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gdk/wayland/gdkwindow-wayland.c
> 
> Maybe I should add the url reference in the changes, idk.

Thanks for the extra explanantion. I see now, we're just forcing this to be an xwayland test

Interesting that GDK wayland seems to support some hints and not others

RFCreate:
It took way more time than estimated, but as far as I know, it is a complete implementation of the urgent status so far. I am open to any improvements you might see.

test/test_hook.py
RFCreate: The new function in base/window shoud do the trick. Let me know your thoughts.

https://github.com/qtile/qtile/blob/9de421f525b5e1e0b93ef67409725135312f59c4/libqtile/backend/base/window.py#L534

richcarni: In `server.c` we also have `qw_handle_activation_request()`. Is that the wayland version of what we're discussing here?
If so, I wonder if that should also be calling the function you've quoted above (we can do that in a separate PR)

RFCreate: Inside `qw_handle_activation_request()` there is this C function call that in python later calls the function I talked about.

https://github.com/qtile/qtile/blob/9de421f525b5e1e0b93ef67409725135312f59c4/libqtile/backend/wayland/qw/server.c#L517

elParaguayo: There's overlap with #5611 here. I've started a thread on discord so we can discuss the approach in a single place: https://discord.com/channels/955163559086665728/1432276759512092742

elParaguayo:
I'm going to merge this now as it adds some useful improvements. 

We can think about broader changes to window activation requests separately.

@RFCreate many thanks for bearing with us on this and thanks for the help.

PR #5603, opened 2025-10-26
Fix stdio redirection to /dev/null

### LINKED ISSUE #5595: Lunching certain processes freezes qtile
### Issue description

lunching some programs like `/bin/cat` through `lazy.spawncmd()` completely freezes qtile, and killing the spawn process from the tty results in a zombie process and qtile doesn't unfreeze. 

Steps to reproduce: 
1. trigger `lazy.spawncmd()` through it's assign keybinding
2. run `/bin/cat` using the spawn prompt 

other programs that cause the same problem: `pc1toppm` , `machinectl`
I often end up accidentally lunching them when trying to tab-complete `pcmanfm` and `materialgram` 

one common factor between `/bin/cat` and `pc1toppm` is that they both wait for input from stdin, however that's not the case with `machinectl`, though it might be a separate issue 

### Version

0.33.1.dev0+g9c51897f.d20250729

### Backend

X11 (default)

### Config

```python

```

### Logs

```text
the only messages in the logs are about the moc widget, nothing related to this issue.
```

### Required

yui-915:
Closes #5595

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1344,9 +1344,9 @@ class Qtile(CommandObject):
         # std{in,out,err} should be /dev/null
         with open("/dev/null") as null:
             file_actions: list[tuple] = [
-                (os.POSIX_SPAWN_DUP2, 0, null.fileno()),
-                (os.POSIX_SPAWN_DUP2, 1, null.fileno()),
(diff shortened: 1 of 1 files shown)


tych0:
Can you add your analysis from you comment to the git commit message?

Thank you!

PR #5604, opened 2025-10-26
[wayland] Add support for `get_wm_type`

elParaguayo:
Would welcome thoughts of @jwijenbergh and @richcarni here.

The frustrating part of this is in xwayland. The xwayland_surface has the window_type atoms but there's no public header that translates these atoms for us. We can connect to an xserver (using xcb) to query this. Again, the xwayland server's connection isn't exposed so we need to create our own, temporary one. Having eventually worked this out, I've seen that sway does the same thing. Would have saved me a lot of time!

@richcarni I've also added a very similar function for xdg windows which returns "popup" for xdg_popups and "normal" for others. Not sure if this could/should be expanded or not.

### DIFF
--- a/libqtile/backend/wayland/qw/view.h
+++ b/libqtile/backend/wayland/qw/view.h
@@ -43,6 +43,7 @@ struct qw_server;
 
 enum qw_view_type {
     QW_VIEW_XDG,
+    QW_VIEW_XDG_POPUP,
     QW_VIEW_XWAYLAND,
     QW_VIEW_INTERNAL,
 };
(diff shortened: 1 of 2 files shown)


elParaguayo:
There is `wlr_xwayland_surface_has_window_type`: https://sergiogdr.pages.freedesktop.org/wlroots/wlr/xwayland/xwayland.h.html#func-wlr_xwayland_surface_has_window_type

We could use that instead but you'd need to call that function lots of times to get the type - doesn't seem the best way to do it here.

jwijenbergh:
Thanks!

PR #5607, opened 2025-10-26
[wayc] Add more xwayland window properties

elParaguayo:
Adds "instance" and "role".

"instance" is used to extend `get_wm_class()` so it now returns the equivalent result to X11 being a list of `[instance, class]`.

"role" is a string such as `GtkFileChooserDialog` which can be used by users for floating rules.


jwijenbergh:
Needs a rebase

elParaguayo:
@jwijenbergh @richcarni can this get merged?

jwijenbergh:
Yes. Thanks!

PR #5608, opened 2025-10-26
[wayc] Add option to build with debugging symbols

elParaguayo:
Adds ability to build with debug symbols and address sanitisation support which may be helpful for tracking down errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,3 +49,17 @@ clean: ## Clean generated files
 update-flake: ## Update the Nix flake.lock file, requires Nix installed with flake support, see: https://nixos.wiki/wiki/Flakes
 	nix flake update
 
+.PHONY: build-wayland
+build-wayland: ## Build wayland backend
+	@python libqtile/backend/wayland/cffi/build.py
+
+.PHONY: build-wayland-debug
+build-wayland-debug: ## Build wayland backend with debug symbols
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -49,3 +49,17 @@ clean: ## Clean generated files
 update-flake: ## Update the Nix flake.lock file, requires Nix installed with flake support, see: https://nixos.wiki/wiki/Flakes
 	nix flake update
 
+.PHONY: build-wayland
+build-wayland: ## Build wayland backend
+	python libqtile/backend/wayland/cffi/build.py
```
tych0: Dumb question: does this happen automatically when doing something like `uv tool install ...` somehow?

elParaguayo: Yes. When we do install then we trigger the build process which invokes `builder.py` which, in turn, calls the above command.

ervinpopescu: Another dumb q(maybe): did you miss an "@" before `python`?

elParaguayo: It would appear so! I've got it on all the others...

jwijenbergh:
Is this OK to be merged?

elParaguayo:
> Is this OK to be merged?

As far as I'm concerned, yes.

jwijenbergh:
Thanks!

PR #5609, opened 2025-10-26
Xwayland - set layer if window has set above or below

elParaguayo:
xwayland surfaces have `above` and `below` attributes which correspond to `_NET_WM_STATE_ABOVE|BELOW`. On mapping the view we now check if either is set and reparent the view accordingly. Event handlers also check in case the state is changed by the client.


jwijenbergh:
Thanks!

PR #5610, opened 2025-10-26
wayland - add `skip_taskbar` property

elParaguayo:
This is set to `False` by default for XDG windows but can be set manually (e.g. scratchpads set this atom on X11).

For xwayland views, we set this depending on the the `skip_taskbar` attribute (which corresponds to the `NET_WM_STATE_SKIP_TASKBAR` atom) and subscribe to signals advertising a change of status.


elParaguayo:
@jwijenbergh this has been rebased

elParaguayo:
Any reason we can't merge this?

jwijenbergh:
I'm not sure why it's needed?

elParaguayo:
It's for the very important reason to exclude the app from our TaskList widget if this is set. This is the behaviour on x11 so we're just replicating here.

jwijenbergh:
ah now I understand

jwijenbergh:
Thanks!

PR #5613, opened 2025-10-27
feat: add iw support to libqtile.widget.wlan to allow use on Fedora

nickeisenberg:
`iwlib` is not easily installable on fedora as  [`wireless-toos-devel`](https://fedoraproject.org/wiki/Changes/RemoveWirelessExtensions) has been dropped and therefor the `Wlan` widget cant be easily used on Fedora as it is now. 

However, `iw` is available and easily installable with `dnf install iw`. This PR adds `iw` as a fall back for `libqtile.widget.Wlan` to use if `iwlib` is not importable. This causes no breaking changes and now allows the widget to be used on Fedora.

### DIFF
--- a/libqtile/widget/wlan.py
+++ b/libqtile/widget/wlan.py
@@ -63,8 +63,9 @@ def _get_status_from_iw(interface_name: str):
         elif line.startswith("signal:"):
             signal_match = re.search(r"signal:\s*(-?\d+)", line)
             if signal_match:
-                quality = int(signal_match.group(1))
-                quality *= -1 if quality < 0 else 1
+                signal = int(signal_match.group(1))
+                # see for quality: https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate
+                quality = signal + 110
 
     if essid is None:
         logger.exception(f"SSID could not be determined from `iw dev {interface_name}` link")
--- a/test/widgets/test_wlan.py
+++ b/test/widgets/test_wlan.py
@@ -31,6 +31,11 @@ def mock_open(output):
     return MockOpen
 
 
+# see for quality: https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate
+# below are the calculations to relate signal from iw and level from iwlib as
+# well as quality from iwlib and signal from iw.
+# level = signal + 256
+# quality = signal + 110
 class MockIwlib(ModuleType):
     DATA = {
         "wlan0": {
@@ -40,7 +45,7 @@ class MockIwlib(ModuleType):
             "BitRate": b"650 Mb/s",
             "ESSID": b"QtileNet",
             "Mode": b"Managed",
-            "stats": {"quality": 49, "level": 190, "noise": 0, "updated": 75},
+            "stats": {"quality": 49, "level": 195, "noise": 0, "updated": 75},
         },
         "wlan1": {
             "ESSID": None,
@@ -63,7 +68,7 @@ class MockIWSubprocessRun:
                 freq: 5180.0
                 RX: 100109613 bytes (77173 packets)
                 TX: 3595242 bytes (19864 packets)
-                signal: -49 dBm
+                signal: -61 dBm
                 rx bitrate: 780.0 MBit/s VHT-MCS 8 80MHz short GI VHT-NSS 2
                 tx bitrate: 650.0 MBit/s VHT-MCS 7 80MHz short GI VHT-NSS 2
                 bss flags: short-slot-time


tych0:
I had this same problem, and wrote essentially the same patch, until I noticed:
```
~ iw help | tail -n2
Do NOT screenscrape this tool, we don't consider its output stable.
```
At the very least, we should add some tests and sample output so we can compare if/when they change it. But really it seems like we should do all the netlink nonsense...

nickeisenberg:
Dain, didn't realize the help message had that comment. Thanks for pointing that out!

tych0:
Yeah... maybe what we do is land something like this and see how stable it really is?

If we do, I'd really like some tests parsing the expected output, so we can compare if we land it.

elParaguayo:
Agreed.

Can you also update the widget's docstring to explain the fallback to `iw`?

nickeisenberg:
I can update the docstring. But are you ok with sticking with `iw` considering the message at the bottom of the help message?

tych0:
Yes, I think we can try it. If it's too noisy, then we can drop the code. 

But I think we need some sample output and tests.

libqtile/widget/wlan.py
```diff
@@ -16,6 +35,59 @@ def get_status(interface_name):
...
+        if line.startswith("SSID:"):
+            essid_match = re.search(r"SSID:\s*(.*)", line)
+            if essid_match:
+                essid = essid_match.group(1)
+
+        elif line.startswith("signal:"):
+            signal_match = re.search(r"signal:\s*(-?\d+)", line)
+            if signal_match:
+                quality = int(signal_match.group(1))
+                quality *= -1 if quality < 0 else 1
```
elParaguayo: Is this right? `iwlib` shows my quality as 41 while `iw` gives me `-70dBm`. This code would return `70` which gets converted to 100%. 

Isn't a higher number better than a lower number? i.e. -30dBm is better than -70dBm.

nickeisenberg: Youre right. Is signal from iwlib given in dBm? I can mess around with this later to see what the relation between iw and iwlib is.

nickeisenberg: So I am getting essentially the signal for both iw and iwlib 

```bash
nicholas@lenovo:~$ iw dev wlp0s20f3 link
Connected to a0:40:a0:80:fe:0f (on wlp0s20f3)
	SSID: Dr.Mike
	freq: 2437.0
	RX: 17710890 bytes (12371 packets)
	TX: 347169 bytes (1811 packets)
	signal: -54 dBm
	rx bitrate: 144.4 MBit/s MCS 15 short GI
	tx bitrate: 144.4 MBit/s MCS 15 short GI
	bss flags: short-slot-time
	dtim period: 2
	beacon int: 100
```

```python
>>> from pprint import pprint
>>> import iwlib
>>> pprint(iwlib.get_iwconfig("wlp0s20f3"))
{'Access Point': b'A0:40:A0:80:FE:0F',
 'BitRate': b'144.4 Mb/s',
 'ESSID': b'Dr.Mike',
 'Frequency': b'2.437 GHz',
 'Mode': b'Managed',
 'NWID': b'Auto',
 'stats': {'level': 202, 'noise': 0, 'quality': 56, 'updated': 75}}
>>>

```

elParaguayo: You might just be in a position where they are similar. As I said, my numbers were very different.

Can you try moving further away from your router and checking again?

elParaguayo: Right now, `iw` is giving me `signal: -39 dBm`. `iwlib`shows `'quality': 70`

nickeisenberg: Ok, so what I posted above what actually just a coincidence. The `quality` key from `iwlib` is not the same thing as the `signal` key from `iw`.

What I have found out is that in the `stats` dict, the `level` key is equal to the `signal` key from `iw` by the equation `signal=level - 256`. So if you see above, I have `level=202` and `202 - 256 = -54` which corresponds to the value for `signal` in my `iw` output above.

Also, according to [this stackexchange post](https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate), you can calculate quality with `(signal + 110)`. I tested this and it matches all up. Ie, from the `iw` output above, `-54 + 110= 56` and this matches with the `iwlib` output, `quality = 56`.

I made this correction in `libqtile.widet._get_status_from_iw` as well as
`test.widet.test_wlan.MockIWSubprocessRun`

test/widgets/test_wlan.py
```diff
@@ -51,6 +57,41 @@ def get_iwconfig(cls, interface):
         return cls.DATA.get(interface, dict())
 
 
+class MockIWSubprocessRun:
+    def __call__(self, *args, **kwargs):
+        assert len(args) == 1
+        cmd = args[0]
+        if "wlan0" in cmd:
+            _stdout = textwrap.dedent("""
+            Connected to 12:34:56:78:90:AB (on wlan0)
```
tych0, line 66: Instead of doing this mocking, can we pass this output to your _get_status_from_iw() directly, since that's what we're really trying to test?

nickeisenberg: As it is now, I cant do that. _get_status_from_iw() only takes the interface name as an input and it runs the subprocess command inside the function to get the output. Are you saying you would rather just have a iw parser and then the subprocess command would be run outside the parser and then the output would be passed to the parser?

tych0: > rather just have a iw parser and then the subprocess command would be run outside the parser and then the output would be passed to the parser?

Yes, perhaps it's best to split it to a whole separate widget, based on GenPollCommand, since that's designed for running commands and parsing the output explicitly. You'll see in the other tests for GenPollCommand widgets that we only have to test the `parse()` methods, because the actual command running is handled and tested in the superclass.

nickeisenberg: Just so to make sure, you are suggesting to rewrite the Wlan widget as a GenPollCommand widget? Or do you want to keep the original Wlan widget and the also create a new WlanIW widget?

tych0: I would create a WlanIW widget for now; otherwise we'll have to do a migration for people.

nickeisenberg: Why would there be a migration? The user would never know anything changed and the widget would still be used just as it is now. My PR just uses iw as a fall back but does not change how libqtile.widgets.Wlan works. Sorry if I am missing something!

tych0: > The user would never know anything changed and the widget would still be used just as it is now.

If you do your rewrite of Wlan to use `iw`, the dependencies will change out from under people, so we need to migrate them to the old WlanWithIwlib.

I understand yours falls back now, but we have a widget and test infrastructure designed for polling a command and parsing its output, so it would be great to avoid duplicating that somehow.

nickeisenberg: Oh gotcha, I didnt realize that you guys had a base class (GenPollCommand) specifically for querying shell commands. 

Should I just scrap this PR then and make a new one? I see you guys like having one widget per file. Is that how you want this done as well?

tych0: I think it's fine to force push here, but if you want a separate PR that's also fine.

One widget per file is good for me, I don't have a strong opinion there. battery.py is notably a mess :)

nickeisenberg: Ill make a new PR, I have a new branch now on my fork with the new widget and tests.

PR #5621, opened 2025-10-30
CI: Cache UV

jwijenbergh:


tych0:
why?

Gurjaka:
> why?

Maybe @jwijenbergh can give better answer, but I found this on discord
image

jwijenbergh:
It's not nice to constantly curl their script, even _if_ we don't get rate limited

tych0:
i see, makes sense. i wonder if we can use their action instead?

https://docs.astral.sh/uv/guides/integration/github/

tych0:
We can close this, I switched to the uv action in 2e57a4a9ebb9600109d41cc185b7ec720a3de691 which should cache the actual uv installation.

PR #5623, opened 2025-10-30
[WIP] layouts: implement scrolling layout

Gurjaka:
TODO:
* [x] Each workspace is treated as a **horizontally infinite strip**, where **columns (tiles)** are arranged side-by-side. The user can scroll or pan left and right to navigate between them.
* [ ] Each column contains **windows stacked vertically** (or sometimes tabbed). The overall layout is “columns of windows” laid out horizontally.
* [ ] **New windows** don’t resize or shift existing ones when inserted — meaning the layout is **stable under insertion**.
* [ ] The **focused window** never moves unless explicitly repositioned by the user. Its relative position stays consistent.
* [ ] **Floating windows** are separate from tiled ones — they float on top, do not scroll with the layout, and have their own per-workspace layer.
* [ ] Parameters such as “default column width,” “preset widths/heights,” and “center focused column” define spatial rules.

### DIFF
--- a/libqtile/layout/scroll.py
+++ b/libqtile/layout/scroll.py
@@ -15,7 +15,8 @@ class Scroll(base.Layout):
 
     defaults = [
         ("border_width", 5, "Window border width"),
-        ("border_colour", "00ff00", "Window border colour"),
+        ("border_focus", "ff0000", "Window focused border colour"),
+        ("border_normal", "333333", "Window normal border colour"),
         ("margin_focused", 5, "Margin width for focused windows"),
         ("margin_unfocused", 5, "Margin width for unfocused windows"),
     ]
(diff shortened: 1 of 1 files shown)


elParaguayo:
I started working on a scrolling layout a while ago. I had implemented the ability to crop windows if they were partially on our off the screen. That code won't work with the new backend but can be adapted. Can share the link if you're interested.

Gurjaka:
> I started working on a scrolling layout a while ago. I had implemented the ability to crop windows if they were partially on our off the screen. That code won't work with the new backend but can be adapted. Can share the link if you're interested.

Sure, first time writing layout so I would appreciate any kind of reference/help

elParaguayo:
There's a guide on writing layouts in our docs too.

Gurjaka:
> There's a guide on writing layouts in our docs too.

Yeah I know, I took the example code, and modified it slightly

Gurjaka:
Let's just hope I don't lose momentum and interest on this feature 😂

PR #5624, opened 2025-10-31
[wayland] Notify pointer motion when dragging

### LINKED ISSUE #5619: [wayland] Firefox can't customize toolbar
### Issue description

Firefox
menu > more tools > customize toolbar...

Can't reorder items in toolbar
Can't add/remove items to/from toolbar

Also can't reorder tabs by dragging

### Version

0.33.1.dev229+geb6c992fc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

richcarni:
When dragging in firefox, pointer focus is lost, so we must also notify pointer motion while dragging

Fixes #5619

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -86,7 +86,8 @@ static void qw_cursor_process_motion(struct qw_cursor *cursor, uint32_t time) {
     wlr_scene_node_set_position(&cursor->server->drag_icon->node, (int)cursor->cursor->x,
                                 (int)cursor->cursor->y);
 
-    if (seat->pointer_state.focused_surface != NULL) {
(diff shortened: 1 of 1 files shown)


elParaguayo:
Does this have any impact on implicit grabs?

richcarni:
Shouldn't affect implicit grab - in every circumstance I've tested IG maintains pointer focus so already correctly notifies pointer motion

This isn't even an issue in most dnd contexts. chrome, for example, seems to do it fine. Just firefox for some reason is giving up pointer focus when we initiate a drag and as a result wasn't updating pointer motion as you drag

elParaguayo:
Ok. So we should just merge this then?

richcarni:
Good to go!

Issue #5625, opened 2025-10-31
[wayland] lazy.spawn rofi failing to launch discord


richcarni:
### Issue description

Can launch discord from terminal
Can open rofi from terminal and launch discord
So seems to be related to `lazy.spawn()`

The issue appears after #5603
which fixes this part of the spawn command:
```
with open("/dev/null") as null:
    file_actions: list[tuple] = [
        (os.POSIX_SPAWN_DUP2, null.fileno(), 0),
        (os.POSIX_SPAWN_DUP2, null.fileno(), 1),
        (os.POSIX_SPAWN_DUP2, null.fileno(), 2),
    ]
```

my config:

```
rofi_cmd = "/opt/rofi-wayland/bin/rofi -config ~/.config/rofi/wayland-config.rasi"
...
Key([mod], "d", lazy.spawn(rofi_cmd + " -show drun")),
```

This is wayland discord with `ELECTRON_OZONE_PLATFORM_HINT=wayland`

### Version

0.0.0 ?

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

saviola777:
Same issue on X11 with `signal-desktop`.

TheCrazyGM:
It's every electron app, doesn't work with dmenu_run either.

EDIT:
Is that is the proper way to send null to in / out / err ?

Because: `discord < /dev/null > /dev/null 2>&1` works just fine from the terminal

EDIT AGAIN:

Because it feels like it should be something like:
```python
with open("/dev/null") as null:
    file_actions: list[tuple] = [
        (os.POSIX_SPAWN_DUP2, null.fileno(), 0),
        (os.POSIX_SPAWN_DUP2,1, null.fileno()),
        (os.POSIX_SPAWN_DUP2,2, null.fileno()),
    ]
```

TheCrazyGM:
Updating it to use this:

```python
with open("/dev/null") as null:
    file_actions: list[tuple] = [
        (os.POSIX_SPAWN_DUP2, null.fileno(), 0),
        (os.POSIX_SPAWN_DUP2, 1, null.fileno()),
        (os.POSIX_SPAWN_DUP2, 2, null.fileno()),
    ]
```

[...]

PR #5626, opened 2025-10-31
fix: correct file descriptor redirection in process spawning

TheCrazyGM:
- Fixed incorrect argument order in POSIX_SPAWN_DUP2 calls for stdout and stderr redirection
- Swapped source and destination file descriptors to properly redirect output to /dev/null
- This resolves issues with process output handling when spawning new processes

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1345,8 +1345,8 @@ class Qtile(CommandObject):
         with open("/dev/null") as null:
             file_actions: list[tuple] = [
                 (os.POSIX_SPAWN_DUP2, null.fileno(), 0),
-                (os.POSIX_SPAWN_DUP2, null.fileno(), 1),
-                (os.POSIX_SPAWN_DUP2, null.fileno(), 2),
+                (os.POSIX_SPAWN_DUP2, 1, null.fileno()),
+                (os.POSIX_SPAWN_DUP2, 2, null.fileno()),
             ]
 
             # To create a rule to move process to a specific group and manage race conditions


TheCrazyGM:
Fixes: #5625 and the original issue in #5595

tych0:
Hmm, are you sure? This seems like it goes back to the old way where we dup stdout to /dev/null, and then dup stderr to that same fd, which is now stdout. 


Can you look at /proc/pid/fd for some of these applications and make sure it looks like you expect?

Or maybe better yet, we add a test that fails if all its stdfds are not /dev/null...

TheCrazyGM:
Well isn't that the expected beahvior.
- null -> stdin
- stdout -> null
- stderr -> null

That's how it would work from the cli if you spawned a program and didn't want any input or output

I'm running it currently all the apps that wouldn't open do, and apps expecting input from the prompt just don't do anything.

Pretty sure it's working as intended

tych0:
But I am confused, https://docs.python.org/3/library/os.html#os.POSIX_SPAWN_DUP2 says:

> (os.POSIX_SPAWN_DUP2, fd, new_fd)

So doesn't this diff dup e.g. stdin (i.e. 1) to the /dev/null fd, which is 3 or whatever? What am I missing?

tych0:
Oh, am I reading this backwards? So #5595 was wrong somehow?

Because I would not expect the directions of the POSIX_SPAWN_DUP2 to differ based on whether its stdin vs. stdout. In both cases we want to replace that fd number with /dev/null. So it should all be the same, it's just a question of what direction...

TheCrazyGM:
Yes, that was the issue, the direction of the piping.

tych0:
So then don't we need to change stdin back as well?

TheCrazyGM:
no, that was the issue before, we dont' want to send stdin to null we want to send null  to stdin
e.g.
```bash
discord < /dev/null # No input

discord > /dev/null 2>&1 # No output, no error

discord < /dev/null > /dev/null 2>&1 # No input, no output, no stderr since it goes to same null
```

tych0:
Oh, oh, I see, thank you. So this statement of mine was incorrect:

> I would not expect the directions of the POSIX_SPAWN_DUP2 to differ based on whether its stdin vs. stdout

Sorry, thank you for being patient with me!

TheCrazyGM:
Anytime!

PR #5627, opened 2025-11-01
[wayland] change foreign toplevel activation callback

elParaguayo:
Foreign toplevel activation requests just focused the window via the compositor but didn't change the group. This meant input was being directed to a window that may not have been visible.

This PR changes the way we handle activation requests. Instead, we treat them like we do in X11 i.e. an activation request from a third party client (such as `rofi`) will depend on how `focus_on_window_activation` is set in the user's config.


jwijenbergh:
Thanks!

jwijenbergh:
Ah I just seen you requested a review from @richcarni . hope this was good to be merged

PR #5629, opened 2025-11-01
[wayland] Handle repeats for key bindings

elParaguayo:


jwijenbergh:
Is that X11 test flaky?

elParaguayo:
> Is that X11 test flaky?

Yes!

richcarni:
Would this still work if we start our timer for key repeats after we've checked if the keypress is handled?

I think this would simplify the `do_repeat()` function - it wouldn't have to deal with checking for handling and notifying seat - that seems like something we shouldn't need to be doing (after the initial key press)

elParaguayo:
Yes. That would seem like a very sensible improvement!

elParaguayo:
@richcarni I've made a change - let me know if that's what you had in mind.

Also, we should squash this into a single commit before merging.

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -18,6 +18,46 @@ static void qw_keyboard_handle_destroy(struct wl_listener *listener, void *data)
...
+
+    if (keyboard->repeat_handled) {
+        // Repeat handled key: send callback
+        uint32_t modifiers = wlr_keyboard_get_modifiers(keyboard->wlr_keyboard);
+        server->keyboard_key_cb(keyboard->repeat_keysym, modifiers, server->cb_data);
+    } else {
+        // Repeat unhandled key: forward to seat
+        wlr_seat_set_keyboard(seat, keyboard->wlr_keyboard);
+        wlr_seat_keyboard_notify_key(seat, time_msec, keycode - 8, WL_KEYBOARD_KEY_STATE_PRESSED);
+    }
```
richcarni: My understanding is we only need to implement repeats for handled keys

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -39,20 +79,56 @@ static void qw_keyboard_handle_key(struct wl_listener *listener, void *data) {
...
+        }
+
+        // Schedule the repeat timer
+        if (keyboard->repeat_source != NULL) {
+            struct wlr_keyboard *wlr_kbd = keyboard->wlr_keyboard;
+            int delay = wlr_kbd->repeat_info.delay;
+            if (delay <= 0)
+                delay = 400; // fallback default
+            wl_event_source_timer_update(keyboard->repeat_source, delay);
+        }
```
richcarni: if handled, I don't think we need to do this

richcarni:
>Would this still work if we start our timer for key repeats after we've checked if the keypress is handled?

What I meant was so we could do something like:
```
handled = server->keyboard_key_cb()
if (handled):
    start_repeat_timer:
else:
    // forward keypress to seat
```

Forgive the mixed language pseudo-code :joy:

libqtile/backend/wayland/qw/keyboard.c
elParaguayo: We do need that, but only if the key press is handled. That's the code to setup and start the timer.

elParaguayo:
@richcarni Thanks for the comments. Let's see if I've got it this time!

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -39,20 +64,52 @@ static void qw_keyboard_handle_key(struct wl_listener *listener, void *data) {
...
                 break;
             }
         }
-    }
 
-    // If not handled, forward the key event to the seat for default processing
-    if (!handled) {
+        if (handled) {
+            // Set up timer to repeat key press
+            if (!keyboard->repeat_source) {
```
jwijenbergh: == NULL

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -39,20 +64,52 @@ static void qw_keyboard_handle_key(struct wl_listener *listener, void *data) {
...
+        } else {
+            // If not handled, forward the key event to the seat for default processing
+            wlr_seat_set_keyboard(seat, keyboard->wlr_keyboard);
+            wlr_seat_keyboard_notify_key(seat, event->time_msec, event->keycode, event->state);
+        }
+
+        // Remove the timer if key is released
+    } else if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+        keyboard->key_pressed = false;
+        if (keyboard->repeat_source) {
```
jwijenbergh: != NULL

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -39,20 +64,52 @@ static void qw_keyboard_handle_key(struct wl_listener *listener, void *data) {
...
+                keyboard->repeat_source =
+                    wl_event_loop_add_timer(server->event_loop, qw_keyboard_do_repeat, keyboard);
+            }
+
+            // Schedule the repeat timer
+            if (keyboard->repeat_source != NULL) {
+                struct wlr_keyboard *wlr_kbd = keyboard->wlr_keyboard;
+                int delay = wlr_kbd->repeat_info.delay;
+                if (delay <= 0)
+                    delay = 400; // fallback default
```
jwijenbergh: where is this default defined?  I mean how did you come up with this value

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -39,20 +64,52 @@ static void qw_keyboard_handle_key(struct wl_listener *listener, void *data) {
...
+                wl_event_source_timer_update(keyboard->repeat_source, delay);
+            }
+        } else {
+            // If not handled, forward the key event to the seat for default processing
+            wlr_seat_set_keyboard(seat, keyboard->wlr_keyboard);
+            wlr_seat_keyboard_notify_key(seat, event->time_msec, event->keycode, event->state);
+        }
+
+        // Remove the timer if key is released
```
jwijenbergh, line 104: comment indentation

libqtile/backend/wayland/qw/keyboard.c
elParaguayo: I asked google for a sensible value in case the user submitted something that made no sense.

libqtile/backend/wayland/qw/keyboard.c
elParaguayo: Sorry - just very used to not doing these explicit checks in python.

libqtile/backend/wayland/qw/keyboard.c
jwijenbergh: maybe it's better to take 600? https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_keyboard.c#L131

libqtile/backend/wayland/qw/keyboard.c
elParaguayo: It's clang doing this.

richcarni:
My apologies, I think I dropped the ball on this one. 3 weeks ago! Looks good FWIW

libqtile/backend/wayland/qw/keyboard.c
jwijenbergh: Tip: you can do clang-format: off as comments and then on again

PR #5631, opened 2025-11-02
feat(manager): update wlr log level when changing qtile log level

RFCreate:
Since qtile allows to update the logging level after startup, it would be nice to do the same for wlr logging level when it is changed using the `manager.py` exposed commands.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -52,6 +52,7 @@ enum wlr_log_importance {
     WLR_DEBUG,
     ...
 };
+enum wlr_log_importance wlr_log_get_verbosity(void);
 extern "Python" void log_cb(enum wlr_log_importance importance,
                                        const char *log_str);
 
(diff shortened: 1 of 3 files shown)


libqtile/core/manager.py
```diff
@@ -1031,30 +1031,35 @@ def run_in_executor(self, func: Callable, *args: Any) -> asyncio.Future:
...
     def debug(self) -> None:
         """Set log level to DEBUG"""
+        self.core.update_backend_log_level()
         logger.setLevel(logging.DEBUG)
         logger.debug("Switching to DEBUG threshold")
 
     @expose_command()
     def info(self) -> None:
         """Set log level to INFO"""
+        self.core.update_backend_log_level()
```
tych0: Don't you want these all after the setLevel() call?

But ideally this wouldn't use side effects, and would take the new log level as an argument so the ordering wouldn't matter.

RFCreate: You are right, I messed with the order while local testing, now I am trying to integrate pytest but it fails as of now, I'll try later again.

RFCreate:
It is ready now! With tests implemented.

PR #5632, opened 2025-11-02
[wayland] update foreign toplevel activation callback

elParaguayo:
This reverts commit f51be7a2d2075b9e6bda4a4e4fe85034a86b3602 and updates it so that requests from pagers automatically focus the window. This in line with behaviour on x11.

The comments in the x11 backend are also updated.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -102,7 +102,7 @@ struct wlr_cursor {
 // main callbacks
 typedef uint32_t xkb_keysym_t;
 typedef struct _cairo_surface cairo_surface_t;
-extern "Python" void view_urgent_cb(struct qw_view *view, void *userdata);
+extern "Python" void view_activation_cb(struct qw_view *view, void *userdata);
 extern "Python" int keyboard_key_cb(xkb_keysym_t, uint32_t, void *userdata);
 extern "Python" void manage_view_cb(struct qw_view *view, void *userdata);
 extern "Python" void unmanage_view_cb(struct qw_view *view, void *userdata);
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -153,9 +153,9 @@ def on_screen_reserve_space_cb(output: ffi.CData, userdata: ffi.CData) -> None:
 
 
 @ffi.def_extern()
-def view_urgent_cb(view: ffi.CData, userdata: ffi.CData) -> None:
+def view_activation_cb(view: ffi.CData, userdata: ffi.CData) -> None:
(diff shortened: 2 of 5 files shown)


RFCreate:
How about sharing the logic in `base/window.py`?

```diff
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -531,6 +531,12 @@ class Window(_Window, metaclass=ABCMeta):
             respect_hints=True,
         )
 
+    def handle_external_focus(self) -> None:
+        logger.debug("Focusing window from external request")
+        self.qtile.current_screen.set_group(self.group)
+        self.group.focus(self)
+        self.bring_to_front()
+
     def handle_window_activation(self):
         focus_behavior = self.qtile.config.focus_on_window_activation
         if (
```

elParaguayo:
Yes - could rename the function as it's only called here. Similarly, we could rename the urgent_cb as it's not always urgent.

Will have to wait until tomorrow now.

elParaguayo:
I've left the focus function name unchanged but I've renamed the urgent one as I think it's misleading.

We should think about the `qw_handle_activation_request` function and see if there are times where we do want to focus the window immediately (i.e. bypassing the user's `focus_on_window_activation` setting).

RFCreate:
> We should think about the `qw_handle_activation_request` function and see if there are times where we do want to focus the window immediately (i.e. bypassing the user's `focus_on_window_activation` setting).

I can't think of a scenario, we'll have to wait til someone files an issue.

I have written a commit that should complement this: https://github.com/RFCreate/qtile/commit/eb160de0aa0e8a7b875c348a13764b31fcf582cc

@elParaguayo what do you think? it now handles the required logic from `focus()`.

elParaguayo:
@RFCreate I agree, I also can't think of a scenario as it shouldn't be possible for one app to try to activate another one via the activation interface. I thought that interface was for apps to try to activate themselves and the foreign toplevel interface was for third party apps (like `rofi` to do this).

I had the same idea about consolidating code into the base class to eliminate duplication so I think you're definitely on the right track there. 

I'd suggest merging this one and then you submit your code as a new PR.

richcarni:
>We should think about the qw_handle_activation_request function and see if there are times where we do want to focus the window immediately (i.e. bypassing the user's focus_on_window_activation setting).

I had gotten confused between this and the ftl handler and had assumed they should all work the same

`qw_handle_activation_request` function is the one where sway will give focus if the request comes from a view with focus else mark as urgent (or something to that effect). Not saying we should necessarily do the same - I think quite reasonable to wait for user feedback

>I'd suggest merging this one and then you submit your code as a new PR.

Happy to merge if there are no objections

PR #5637, opened 2025-11-06
feat: added a WlanIw widget that does not depend on iwlib

nickeisenberg:
A wlan wdiget for those who do not have access to iwlib.

### DIFF
--- a/libqtile/widget/wlaniw.py
+++ b/libqtile/widget/wlaniw.py
@@ -2,27 +2,8 @@ import re
 import subprocess
 
 from libqtile.log_utils import logger
+from libqtile.pangocffi import markup_escape_text
 from libqtile.widget.generic_poll_text import GenPollCommand
-
-
-def parse_iw_output(raw: str):
-    essid, quality = None, None
-    for line in raw.splitlines():
-        line = line.strip()
-
-        if line.startswith("SSID:"):
-            essid_match = re.search(r"SSID:\s*(.*)", line)
-            if essid_match:
-                essid = essid_match.group(1)
-
-        elif line.startswith("signal:"):
-            signal_match = re.search(r"signal:\s*(-?\d+)", line)
-            if signal_match:
-                signal = int(signal_match.group(1))
-                # see: https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate
-                quality = signal + 110
-
-    return essid, quality
 
 
 def get_private_ip(interface_name):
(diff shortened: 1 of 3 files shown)


libqtile/widget/wlaniw.py
```diff
@@ -0,0 +1,132 @@
...
+            signal_match = re.search(r"signal:\s*(-?\d+)", line)
+            if signal_match:
+                signal = int(signal_match.group(1))
+                # see: https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate
+                quality = signal + 110
+
+    return essid, quality
+
+
+def get_private_ip(interface_name):
```
tych0, line 9: Can we import this from wlan instead of copy/pasting it?

libqtile/widget/wlaniw.py
```diff
@@ -0,0 +1,132 @@
...
+            ipaddr = "N/A"
+            if not disconnected:
+                ipaddr = get_private_ip(self.interface)
+
+            else:
+                if self.use_ethernet:
+                    ipaddr = get_private_ip(self.ethernet_interface)
+                    try:
+                        with open(
+                            f"/sys/class/net/{self.ethernet_interface}/operstate"
```
tych0: Hmm, actually can we share all of this logic? (Or better yet, maybe get rid of it? I guess most wlans will not have an address unless they are up, contrary to wired NICs...)

test/widgets/test_wlaniw.py
```diff
@@ -0,0 +1,90 @@
...
+    assert out == "QtileNet 71%"
+
+    widget = patched_wlaniw.WlanIw(format="{ipaddr}")
+    out = widget.parse(IW_OUTPUT)
+    assert out == IP_ADDRESS
+
+
+def test_wifi_no_connection(patched_wlaniw):
+    widget = patched_wlaniw.WlanIw()
+    out = widget.parse(IW_OUTPUT_NO_CONN)
```
tych0, line 74: These all look great, thank you.

libqtile/widget/wlaniw.py
nickeisenberg: Yeah thats not an issue, but a try/except statement needs to be added to `import iwlib` from the `widget.wlan.py` file, other wise and error will be thrown when importing `get_private_ip` from within `wlaniw.py`.

libqtile/widget/wlaniw.py
nickeisenberg: What do you mean by share all the logic? My first PR that I closed essentially shared all of the logic so we could do that if you want.

libqtile/widget/wlaniw.py
tych0: Oh, you can just move it here and have wlan.py import this one, then. Just so long as there's only one copy.

libqtile/widget/wlaniw.py
tych0: Can we make it a function instead of a method and import it in the other file?

nickeisenberg: Yeah that should be fine. How do you want this structured? Make a wlan dir inside widgets and then place both wlan widgets in there and also place a utils file in there as well and then both widgets import from the common utils file?

tych0: I wouldn't bother with the directory and the utils file, I think wlan.py can import from wlaniw.py directly.

nickeisenberg: I broke up the logic into shared helpers now and the tests pass. Let me know what you think!

nickeisenberg: Hold up, theres going to be an issue with setting the ethernet interface not found attribute for both Wlan classes. I'll fix this later when I'm at my computer.

tych0: LGTM otherwise, let me know when you're happy with it and I'll land it. Thanks!

nickeisenberg: I think its all good now. The ethernet_interface_not_found attribute will be changed to True now if that part of the loop gets triggered. Hopefully I didnt miss anything. Thanks for all the feedback!

PR #5638, opened 2025-11-07
feat(window): share activation logic across backends

RFCreate:
As discussed in #5632, this puts the window activation logic into `base/window.py` so now it is the same across X11 and Wayland backends.


elParaguayo:
This is a nice improvement. Thank you.

PR #5643, opened 2025-11-10
wayland: internal popups return focus on kill

richcarni:


libqtile/backend/wayland/qw/internal-view.c
```diff
@@ -100,6 +100,11 @@ static void qw_internal_view_unhide(void *self) {
 
 static void qw_internal_view_kill(void *self) {
     struct qw_internal_view *view = (struct qw_internal_view *)self;
+    if (view->base.server->seat->keyboard_state.focused_surface == NULL) {
+        wlr_scene_node_set_enabled(&view->base.content_tree->node, false);
+        view->base.server->focus_current_window_cb(view->base.server->cb_data);
+    }
+
```
richcarni, line 107: Maybe should be in `qw_internal_view_hide()` and popup calls `hide()` as first step in `kill()`?

But I've avoided touching popup since there is no issue in x11

tych0:
Can you add some detail in the commit message about what exactly this fixes?

richcarni:
I should have just just shared this via my personal repo for testing

PR #5645, opened 2025-11-11
wayland: Give server key bindings precedence

### LINKED ISSUE #5605: [wayland] Server doesn't respond to keybindings after closing a `Popup` window that's grabbed the keyboard
`qtile-extras` uses qtile's `Popup` class in its popup toolkit which can be used to create context menus for widgets (e.g. `StatusNotifier`. If the popup is closed (e.g. by pressing escape) the server is not grabbing keybindings until I click somewhere on the screen.

richcarni:
Server key bindings are processed first and suppress internal window bindings when matched, consistent with the behaviour of the old backend

Fixes #5605

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -370,15 +370,14 @@ class Core(base.Core):
         self.check_screen_fullscreen_background()
 
     def handle_keyboard_key(self, keysym: int, mask: int) -> bool:
+        if (keysym, mask) in self.grabbed_keys:
+            assert self.qtile is not None
+            _, swallowed = self.qtile.process_key_event(keysym, mask)
+            if swallowed:
+                return True
+
         if self.focused_internal:
             self.focused_internal.process_key_press(keysym)
-            return True
-
-        if (keysym, mask) not in self.grabbed_keys:
-            return False
-
-        assert self.qtile is not None
-        if self.qtile.process_key_event(keysym, mask)[1]:
             return True
 
         return False


richcarni:
This PR doesn't fix focus not being restored to the previous window when an internal popup is closed. This is also an issue with the old backend

My suggestion would be that popups handle this like bar/widgets:
https://github.com/qtile/qtile/blob/7db14d21b499f22024aa71e12a3cd4d8ac9d87dc/libqtile/bar.py#L612-L631

jwijenbergh:
merci!

jwijenbergh:
> This PR doesn't fix focus not being restored to the previous window when an internal popup is closed. This is also an issue with the old backend
> 
> My suggestion would be that popups handle this like bar/widgets:
> 
> https://github.com/qtile/qtile/blob/7db14d21b499f22024aa71e12a3cd4d8ac9d87dc/libqtile/bar.py#L612-L631

fine to handle this later, sounds quite specific

elParaguayo:
I don't remember this being an issue at all on the old backend.

jwijenbergh:
> I don't remember this being an issue at all on the old backend.

aren't popups rarely used

elParaguayo:
> > I don't remember this being an issue at all on the old backend.
> 
> aren't popups rarely used

Not at all. They're used for context menus etc in qtile extras. I have quite a few widgets that rely on it.

jwijenbergh:
> > > I don't remember this being an issue at all on the old backend.
> > 
> > 
> > aren't popups rarely used
> 
> Not at all. They're used for context menus etc in qtile extras. I have quite a few widgets that rely on it.

I doubt most users use this ;)

elParaguayo:
Very good point!

richcarni:
Doesn’t everyone on wayland use statusnotifier from qtile-extras?!

I need to start a petition to have that replace the built in statusnotifier

elParaguayo:
Won't happen. Supporting the popup toolkit in qtile is a lot of extra code.

Issue #5641, opened 2025-11-09
[wayc] rofi spawns only on screen 0


ervinpopescu:
### Issue description

When trying to spawn rofi on screen 1/2, it only spawned on screen 0. Will take a look at the code.

### Version

0.33.1.dev309+g7db14d21b

### Backend

Wayland (experimental)

### Config

```python
I don't know what exactly might be relevant here
```

### Logs

```text
>INFO logs show nothing of interest.
```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

richcarni:
Thanks for the report! This was one I was aware of but good to have a real world example

Also good to see a multi-screen user on the new backend!

tdivis:
I switched to qtile@master and wofi also spawns only on main notebook monitor even though I have HDMI monitor configured as `position 0,0` in kanshi and I'm focused on the HDMI monitor.

ervinpopescu:
More examples:
- firefox window which is used for `Ctrl+Tab`-ing between windows
- `nwgbar`

@richcarni of course i did not get to this before you did :P thanks! will test your patch.

EDIT: I see that we have some possible performance implications and you'll do some further changes. I'll test after those as I have some "performance-critical" workflow these days.

richcarni:
@ervinpopescu Actually it occurred to me, performance implications aside, that this probably doesn't work the way you would expect

My understanding is that you need to click on an empty screen to give it focus. With #5646 rofi appears on whatever screen the pointer is on, but (assuming you're using it as a launcher) the app will open on the screen which qtile thinks has focus, and these may not be the same screen

Thinking maybe we need to use a hook to synchronise the current screen or a callback to check qtile's screen 

Unfortunately not going to be able to work on it for a few days - feel free have a go yourself if you feel inclined

ervinpopescu:
> My understanding is that you need to click on an empty screen to give it focus. With https://github.com/qtile/qtile/pull/5646 rofi appears on whatever screen the pointer is on, but (assuming you're using it as a launcher) the app will open on the screen which qtile thinks has focus, and these may not be the same screen

Agree.

> Thinking maybe we need to use a hook to synchronise the current screen or a callback to check qtile's screen

That does sound like the right thing to do... More likely, the callback is better.

[...]

Issue #3381, opened 2022-02-24
Mouseover on an empty monitor doesn't bring focus to it.


SweepingBishops:
In a double monitor setup click on  monitor 1 (both the monitors are currently empty,i.e have no open windows). This causes the keyboard focus to shift to that monitor.  
Move the mouse to monitor 2 and open a window. I would expect it to open where the mouse is(monitor 2), but it does not.  
What I infer from this is that the focus is never shifted.  
Clicking on monitor 2 before opening the window opens it as expected.  
If a window is open on monitor 2 a mouse over shifts focus as expected.  
Also of interest is that when I open rofi, it opens where the mouse is (also dunst notifications but I have set it to follow my mouse in its config). So do floating windows behave differently?  
I use Arch, and qtile is from its repos.

elParaguayo:
I think the behaviour you describe is expected:
- Windows (tiled and floating) open in the currently focused group/screen
- Screens are not focused by a mouse entering that screen
- Clicking a screen focuses it
- Rofi has an option to open on the monitor with the mouse and that is supported by Qtile

This issue was discussed a bit in #2657 too.

m-col:
>  Also of interest is that when I open rofi, it opens where the mouse is (also dunst notifications but I have set it to follow my mouse in its config). So do floating windows behave differently?

Focussing an output is a Qtile concept, rather than X11 concept, so rofi doesn't know which one is focussed. That's why these programs will can decide to: 1. open in a set output, 2. open on the currently focussed window's output, or 3. open on the output containing the mouse.

> This issue was discussed a bit in https://github.com/qtile/qtile/issues/2657 too.

Reading through this again it looks like we (or I) was pro tracking cursor motion across outputs, but one thing missing from that issue (maybe it's in another) is that, while this is simple when moving across the root window, moving across the screen boundaries while over a window and toggling that would _also_ require tracking cursor motion for each window. This is also possible, just, you know even more cursor tracking.

Another caveat: consider dragging a window from near its left edge rightward over a screen boundary. Currently the screen focus switches when the centre of that window crosses the screen boundary, while the mouse is still on the initial screen. I'm not sure what the best approach there would be as both options might seem intuitive.

arjan-s:
Actually this is something I've missed as well after switching from sway.
[...]

PR #5646, opened 2025-11-11
wayland: Update current screen on cursor motion

### LINKED ISSUE #5511: [wayc] xdg popup appears on wrong display
### Issue description

If you open an xdg popup on a second display, it appears in the wrong position on the primary display

I think the issue is with `output->x` and `output->y` in `qw_xdg_popup_unconstrain()`

Image

### Version

wayc

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

richcarni:
Ensure the compositor updates the current output when the cursor moves, so that layer-shell views and XDG popups appear on the correct screen

Fixes #5511, #5641

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -176,6 +176,12 @@ def on_session_lock_cb(locked: bool, userdata: ffi.CData) -> None:
     core.set_locked(locked)
 
 
+@ffi.def_extern()
+def get_current_output_dims_cb(userdata: ffi.CData) -> ffi.CData:
+    core = ffi.from_handle(userdata)
+    return core.handle_get_current_output_dims()
+
+
 def get_wlr_log_level() -> int:
     if logger.level <= logging.DEBUG:
         return lib.WLR_DEBUG
@@ -221,6 +227,7 @@ class Core(base.Core):
         self.qw.on_input_device_added_cb = lib.on_input_device_added_cb
         self.qw.focus_current_window_cb = lib.focus_current_window_cb
         self.qw.on_session_lock_cb = lib.on_session_lock_cb
+        self.qw.get_current_output_dims_cb = lib.get_current_output_dims_cb
         lib.qw_server_start(self.qw)
         os.environ["WAYLAND_DISPLAY"] = self.display_name
         self.qw_cursor = lib.qw_server_get_cursor(self.qw)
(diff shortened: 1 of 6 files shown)


jwijenbergh:
what is the performance impact of this?

richcarni:
> what is the performance impact of this?

Great question! Short answer is I don't know, but I was already debating whether we should/need to store the state of the current output at all

Since we not (currently) doing anything when the current_output changes, if would be much more efficient to just do this check in the couple of places where it's needed, when it's needed

I'll make that change

richcarni:
It's a little academic now given this is the wrong approach (because it will not be synchronised with `qtile.current_screen`), but I was curious re performance. 

This is a flame graph from `py-spy --native` (20 seconds, 100 sample/s, while randomly moving the mouse)

You have to dig a bit to find it but `qw_cursor_process_motion()` appears to be entirely taken up by `cursor_motion_cb()` and `wlr_seat_pointer_send_motion()`

So my conclusion is updating the current screen here has negligible overhead

screen_update

image

richcarni:
As I see it, the 2 options here are to get the current screen from qtile via a callback (when required) or use the `subscribe.current_screen_change()` hook to update server->current_output

I went with the callback. My reasoning is it avoids keeping extra state. One could argue the alternative is better for being simpler - I don't feel strongly either way

@jwijenbergh If you're happy with this I'll remove the original commit to clean up the history

ervinpopescu:
Tested with c7f0cb268.

One more nitpick: when I'm switching to another screen which has an empty group,`rofi` still gets opened on the previous screen AND the app is opened on the previous screen. I have `follow_mouse_focus = True`. Idk if anything else matters config-wise.

richcarni:
> One more nitpick: when I'm switching to another screen which has an empty group,`rofi` still gets opened on the previous screen AND the app is opened on the previous screen. I have `follow_mouse_focus = True`. Idk if anything else matters config-wise.

I was a bit confused by this behaviour too. As far as I can tell `follow_mouse_focus` doesn't focus an empty screen - although this may be the wrong behaviour! (I don't usually use multiple screens)

When I test on the old backend, I get rofi opening on the screen where the pointer is, BUT the app opens on the other screen

So I think this implementation is correct - it just uses qtile's focused screen. If the focused screen should follow the mouse is maybe a separate issue? Does anyone in this conversation know or can test what x11 does?

Update: Appears to be a few existing issues around this: #3381, #2567

jwijenbergh:
is this ready?

richcarni:
> is this ready?

Yep! Thanks! I think the behaviour is correct (or at least consistent)

jwijenbergh:
Thanks! Smart to get the current output from qtile

libqtile/backend/wayland/core.py
```diff
@@ -300,6 +307,18 @@ def get_screen_for_output(self, output: ffi.CData) -> Screen:
 
         return self.qtile.current_screen
 
+    def handle_get_current_output_dims(self) -> ffi.CData:
+        assert self.qtile is not None
+
+        output_dims = ffi.new("struct wlr_box *")
```
elParaguayo, line 313: @jwijenbergh @richcarni For my own education, how does this data get tidied up?

jwijenbergh: by the python GC/refcounter. This function creates a struct and returns it (so the data is copied over). You can see when it is cleaned up using (untested):

```python
output_dims = ffi.new("struct wlr_box *")
output_dims = ffi.gc(output_dims, lambda p: print("output dims are cleaned up"))
```

PR #5647, opened 2025-11-11
Fix release workflow

jwijenbergh:
See commit descriptions


jwijenbergh:
This gives a green CI when building the wheels in my fork with a test release

tych0:
I think you can also push a tag like `v0.33.5-bogus` and it'll not do an upload to `test.pypi.org` (i.e. not the prod pypi). I can give you creds there as well, if you want to do any more vetting. But none of that is necessary, I suspect your testing is enough. Just for if you want to test more now or in the future.

PR #5649, opened 2025-11-15
backlight: look for a sane default

tych0:
Often machines only have one backlight, but it can vary by GPU driver: acpi_video0, intel_backlight, amdgpu1, etc.

Let's just look in the backlight directory and see which one users have, and default to that.


elParaguayo (approved):
Very nice change.

elParaguayo (approved):
Very nice change.

PR #5650, opened 2025-11-15
ci: use the astral provided uv action

tych0:
We have seen 429s because we are hammering uv's install script too hard. Let's use their action and enable caching so that we don't hammer their server or pypi so hard.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,12 @@ jobs:
               uses: actions/setup-python@v5
               with:
                   python-version: ${{ matrix.python-version }}
-            - name: Install uv
-              run: curl -LsSf https://astral.sh/uv/install.sh | sh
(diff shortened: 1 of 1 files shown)


tych0:
This is a nicer version of #5621 IMO, thoughts welcome.

tych0:
Worth noting that this switches us to uv managed pythons from github actions managed pythons. Maybe that's a bad thing? uv-managed pythons are typically built against a minimal libc, which e.g. doesn't support the fancy os.POSIX_SPAWN_DUP2 arguments and stuff.

tych0:
Ok, on second thought, I don't like using `uv`'s pythons for exactly that reason. So let me drop that part, but I'll keep the uv action and caching and stuff.

tych0:
Ok, hearing no objections, I'm going to land this so we stop banging on their servers :)

PR #5652, opened 2025-11-16
docs: order config variables alphabetically

ervinpopescu:
Nitpick, but hopefully merge-able :)


elParaguayo:
@tych0 Do we ditch coveralls entirely or just make that check optional?

This PR is clearly fine to merge!

tych0:
> Do we ditch coveralls entirely or just make that check optional?

I'd vote for ditching entirely, but I don't have a strong opinion and could be convinced otherwise.

tych0:
I just re-logged in to coveralls since I think somehow it depends on either yours (@elParaguayo) or my oauth tokens. If this doesn't work, I'll send a patch to nuke it.

tych0:
Arg, seems like that worked. coveralls lives another day :)

PR #5656, opened 2025-11-17
fix(xdg_activation): remove optional valid_surface check

RFCreate:
I have been trying to develop a window that can send the `xdg_activation` activate request to qtile but the request always get caught in this lines before it can pass to python.

https://github.com/qtile/qtile/blob/3a641a207a32be196b1756acaac7e81b2107023c/libqtile/backend/wayland/qw/server.c#L560-L563

Since we have in the config an option (`focus_on_window_activation`) that handles what to do when a window request activation, the config should be one that decides whether to let the activation through or ignore it.

### DIFF
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -554,11 +554,9 @@ static void qw_handle_activation_request(struct wl_listener *listener, void *dat
     }
 
     if (server->view_activation_cb != NULL) {
+        wlr_log(WLR_DEBUG, "Activation token valid, focusing view");
         server->view_activation_cb((struct qw_view *)view, server->view_activation_cb_data);
     }
-
-    wlr_log(WLR_INFO, "Activation token valid, focusing view");
-    qw_xdg_view_focus(view, 1);
 }
 
 void qw_server_set_keymap(struct qw_server *server, const char *layout, const char *options,


elParaguayo:
Apologies if I'm missing something here but, if the window is already focused, why do we need to do anything with the activation request?

RFCreate:
Before the rewrite to C, the activation request would have passed even with the window focused, but I think you are right, maybe a clearer log message will be enough.

RFCreate:
I am trying to send the activation after unfocusing the window but now got an "invalid_surface", is it that when the token is sent and the window is not focused it will always be `NULL`? In that case the activation will never be reached.

https://github.com/qtile/qtile/blob/3a641a207a32be196b1756acaac7e81b2107023c/libqtile/backend/wayland/qw/xdg-view.c#L531-L533

elParaguayo:
Are you calling `set_surface` on the token to set the requestor's surface? (https://wayland.app/protocols/xdg-activation-v1#xdg_activation_token_v1:request:set_surface)

RFCreate:
No, but when I have the window in focus, I do not have that problem.

EDIT: I think is a gdk thing. [reference](https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gdk/wayland/gdkwindow-wayland.c#L3989-3993)

elParaguayo:
So it only sets the surface if its focused when making the request? I guess that implies that the request is from user interaction and so should be treated as a valid request.

RFCreate:
So should we remove the "valid_surface" check? In the wayland docs it says that `set_surface` is optional.

elParaguayo:
I think so. If it's optional, it seems crazy to reject a request that hasn't set it. @richcarni let us know if you disagree.

richcarni:
Sounds like it is up to us if it's optional. I would tend to go with what the old backend did - unless there is a good argument for change - so no, I don't disagree

richcarni:
While we're here, quick question on this same function...

Do either of you know if the final `qw_xdg_view_focus()` is necessary? I'd assume focusing would be handled in the callback?
https://github.com/qtile/qtile/blob/4c92f35b80fc843ef2930856a43d4da82a26b6f2/libqtile/backend/wayland/qw/server.c#L554-L560

elParaguayo:
> Sounds like it is up to us if it's optional.

Yes, that's fair. The protocol does say we may reject requests if this isn't set.

I can't see that we checked the token at all in the old backend.

So, if we know clients aren't setting this but we still want to allow the request then we need to skip this check.

RFCreate:
> Do either of you know if the final `qw_xdg_view_focus()` is necessary? I'd assume focusing would be handled in the callback?

Thanks for pointing that out, after having a successful activation that function caused weird behavior, like I was in another group but the window that made the request had keyboard focus. So I remove that step to prevent that.

libqtile/backend/wayland/qw/server.c
```diff
@@ -546,28 +546,21 @@ static void qw_handle_activation_request(struct wl_listener *listener, void *dat
...
     }
 
     if (focused == event->surface) {
-        wlr_log(WLR_INFO, "Activation token invalid for surface");
+        wlr_log(WLR_INFO, "Ignoring activation request, surface is focused");
         return;
     }
 
     if (server->view_activation_cb != NULL) {
+        wlr_log(WLR_INFO, "Activation token valid, focusing view");
```
elParaguayo: Should this be debug level instead? Behaviour would seem normal to a user so no need to tell them what's happened. I appreciate this is just moved from further below but this would seem a good time to change this too.

RFCreate: Sure, will do. Since we are talking about messages, should we remove this one? It does not serve much purpose.

https://github.com/qtile/qtile/blob/b57ad5c49de462159d39671a310cc0d2c9e1b65d/libqtile/backend/wayland/qw/server.c#L556-L558

elParaguayo: Yes - that can go.

RFCreate: I believe this to be ready.

elParaguayo:
Looks good. I've enabled merge once CI is finished.

PR #5660, opened 2025-11-20
bar: preserve bar and widget references during finalization

### LINKED ISSUE #5436: Bar disappears when the monitor is turned off
### Issue description

Those issues and changes look relevant

- https://github.com/qtile/qtile/issues/4821
- https://github.com/qtile/qtile/issues/5313
- https://github.com/qtile/qtile/pull/5311 (available in v0.33.0)

Similar to #5313 my Qtile bar disappears after turning off my monitor (pressing the power button). Everything else in Qtile seems to remain functional. Reloading the configuration with `lazy.reload_config()` does revert the bar

### Version

0.33.1.dev0+g9c51897f.d20250729

### Backend

X11 (default)

### Config

https://github.com/angrybacon/dot/blob/master/config/qtile.org#bar

### Logs

```text
I too get no logs in `~/.local/share/qtile/qtile.log`
```

### Required

tych0:
The core insight here is: finalization is for the resources allocated by a bar or widget, not deleting the bar or widgets themselves. If we do that, then when people turn their monitors back on, we will have mutilated their configs, deleting the bars.

Instead, we should finalize these python objects, since they know how to re-allocate the resources as necessary on _configure() again if someone re-attaches a screen. However, we do not do silly things like widgets.clear() (deleting all the widgets from a bar), or screen.top = None, removing the widget from the Screen.

This lets us revert 99c8a54b026d ("x11: don't ignore powered off monitors"), since we really do want to ignore powered off monitors, but we couldn't before because we were deleting widgets and bars from configs when the monitors were disconnected.

Fixes #5436

### DIFF
--- a/libqtile/backend/x11/xcbq.py
+++ b/libqtile/backend/x11/xcbq.py
@@ -377,6 +377,10 @@ class RandR:
         for output in self.ext.GetScreenResources(root).reply().outputs:
             info = self.ext.GetOutputInfo(output, xcffib.CurrentTime).reply()
 
+            # ignore disconnected monitors
+            if info.connection != xcffib.randr.Connection.Connected:
+                continue
+
             # ignore outputs with no monitor plugged in
             if not info.crtc:
                 continue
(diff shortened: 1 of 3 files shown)


tych0:
Ok, I didn't want to land this before the release, but I am Pretty Sure this is the correct fix. I've been running it for ~week and it seems to work, hopefully getting it into git will help more people test it.

PR #5661, opened 2025-11-20
wayland: Update XDG surfaces if geometry changed

### LINKED ISSUE #5654: [wayland] Files/nautilus incomplete rendering on open
### Issue description

Nautilus (v46.4) is the only app I'm seeing this for currently. When I open it, it's missing strips along the left and top edges. Once something triggers a re-render (changing windows is enough) everything renders correctly

I don't see this on the old backend. I can reproduce on a second machine

Image

### Version

#d5576a2

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

richcarni:
Some clients update their surface geometry after being mapped. When this occurs we should reposition the surface

Sway and old backend do/did something similar

Fixes #5654

### DIFF
--- a/libqtile/backend/wayland/qw/xdg-view.c
+++ b/libqtile/backend/wayland/qw/xdg-view.c
@@ -101,19 +101,41 @@ static void qw_xdg_view_handle_destroy(struct wl_listener *listener, void *data)
     free(xdg_view);
 }
 
+static void qw_xdg_view_clip(struct qw_xdg_view *xdg_view);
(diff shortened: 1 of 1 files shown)


elParaguayo:
Looks ok to me.

jwijenbergh:
Maybe later we can add a geom changed inline function, as we do the same thing in two places

jwijenbergh:
Oh and thanks!

PR #5665, opened 2025-11-23
Support building without wayland backend

mgorny:
Support passing `backend=x11` config setting symmetric to `backend=wayland` that can be used to disable building the Wayland backend entirely, rather than trying to build it anyway. This is important for redistributors who need to ensure reproducible build irrespective of whether Wayland dependencies are installed in the build environment or not. The automatic behavior is preserved if neither value is passed.

While at it, I've made the automatic path output the exception message as well -- it's not exactly helpful when you see a bunch of successful compilations followed by "Warning: Wayland backend not built. Backend will not run."

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -9,10 +9,13 @@ WAYLAND_FFI_BUILD = "./libqtile/backend/wayland/cffi/build.py"
 def wants_wayland(config_settings):
     if config_settings:
         for key in ["Backend", "backend"]:
-            if config_settings.get(key, "").lower() == "wayland":
+            value = config_settings.get(key, "").lower()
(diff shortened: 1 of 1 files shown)


elParaguayo:
Thanks for this. Seems like a reasonable request to tidy up the build process.

Could you add something to the docs to explain this? If you don't have time, I'll add in a separate PR later.

elParaguayo:
I've tested this and it works fine. I appreciate you've probably got a lot else on looking after a ton of gentoo packages so I'll sort the docs.

mgorny:
Thanks a lot!

PR #5667, opened 2025-11-24
XWayland: clear keyboard focus on hide

RFCreate:
- fix xwayland `request_activate` signal notify
- fix `util.c` was not compile with WLR_HAS_XWAYLAND defined
- clear keyboard focus on hide for Xwayland
- deactivate surface on hide for Xwayland and XDG


richcarni:
> fix util.c was not compile with WLR_HAS_XWAYLAND defined

Not sure I understand - I'm probably missing something - but when/why wasn't it compiling and how does changing the order of includes fix that?

Rest looks good!

RFCreate:
`#include <wlr/config.h>` header is needed.

richcarni:
> `#include <wlr/config.h>` header is needed.

Oh yeah 😳

Thanks for this!

PR #5672, opened 2025-11-26
unify hovered_window calculation logic

### LINKED ISSUE #1413: drag: wrong window is dragged

tych0:
X11's logic seems flawed when the window straddles two monitors. Rather than try and fix this up, let's just do the same thing wayland does: keep track of the last window we entered, because that's the one the mouse is currently over.

Since we're doing the same thing wayland is, we can do it on the core object, and get rid of some backend specific code for x11 here.

This hopefully:

Fixes #1413

...I'm making this PR just because I don't want to lose the change. I think it's a nice improvement, but I haven't been able to verify yet, nor has anyone on that bug.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -190,8 +190,6 @@ class Core(base.Core):
     supports_restarting: bool = False
 
     def __init__(self) -> None:
-        # This is the window under the pointer
-        self._hovered_window: base.WindowType | None = None
         # this Internal window receives keyboard input, e.g. via the Prompt widget.
         self.focused_internal: base.Internal | None = None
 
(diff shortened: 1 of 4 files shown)


RFCreate:
+1. It may fix this:

```py
Exception ignored from cffi callback <function cursor_button_cb at 0x7f4a33aa8c20>:
  File "/home/user/github/qtile/venv/lib/python3.13/site-packages/libqtile/backend/wayland/core.py", line 138, in cursor_button_cb
    if core.handle_cursor_button(button, mask, pressed, x, y):
  File "/home/user/github/qtile/venv/lib/python3.13/site-packages/libqtile/backend/wayland/core.py", line 353, in handle_cursor_button
    handled = self.qtile.process_button_click(int(button), int(mask), x, y)
  File "/home/user/github/qtile/venv/lib/python3.13/site-packages/libqtile/core/manager.py", line 852, in process_button_click
    self._focus_hovered_window()
  File "/home/user/github/qtile/venv/lib/python3.13/site-packages/libqtile/core/manager.py", line 839, in _focus_hovered_window
    window = self.core.hovered_window
  File "/home/user/github/qtile/venv/lib/python3.13/site-packages/libqtile/command/base.py", line 272, in __getattr__
    raise AttributeError(f"{self.__class__} has no attribute {name}")
AttributeError: <class 'libqtile.backend.wayland.core.Core'> has no attribute hovered_window
```

RFCreate:
Nitpicking here. Should `_hovered_window` be without the leading underscore?

tych0:
Yeah, I think you're right. We have a lot of unnecessary underscore spam in the codebase, let me not add more. Thank you!

RFCreate:
Can we get this merge? I am writing some tests that are blocked by the `AttributeError` above https://github.com/qtile/qtile/pull/5672#issuecomment-3581273733

elParaguayo:
Are you happy that this PR fixes the issue? I've not had a chance to run this but am happy to merge if it's working for you.

tych0:
I'm Pretty Sure it works, let's see 🤞

PR #5673, opened 2025-11-26
wayland: Add core.get_mouse_position()

### LINKED ISSUE #5669: [wayland] core.get_mouse_position() not implemented
### Issue description

the title

### Version

latest

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

richcarni:
Fixes #5669

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -636,6 +636,10 @@ class Core(base.Core):
                 hook.fire("unlocked")
         self._locked = locked
 
+    def get_mouse_position(self) -> tuple[int, int]:
+        """Get mouse coordinates."""
+        return int(self.qw_cursor.cursor.x), int(self.qw_cursor.cursor.y)
+
     @expose_command()
     def set_keymap(
         self,


jwijenbergh:
nit: there in theory could be no cursor right?

richcarni:
> nit: there in theory could be no cursor right?

Yes, good point!

I will add a check, and then I think return (0, 0) if no cursor? I believe this would be consistent with x11 (XQueryPointer)

richcarni:
Looking at this more closely it appears there is always a cursor (unless the server is torn down)

I added this check to `server.c`, before there are any input devices:
```
server->cursor = qw_server_cursor_create(server);
wlr_log(WLR_ERROR, "input devices empty: %i", wl_list_empty(&server->input_devices));
wlr_log(WLR_ERROR, "cursor x, y: %f %f", server->cursor->cursor->x, server->cursor->cursor->y);
```

Which outputs:
```
[server.c:699] input devices empty: 1
[server.c:700] cursor x, y: 100.000000 100.000000
```

I guess the cursor position initially defaults to 100, 100? Presumably the last known position if all input devices are removed (untested)

PR #5674, opened 2025-11-27
[wayland] bump layer shell to v5

elParaguayo:
wlroots 0.19 supports version 5 of the layer shell protocol so we should tell clients that we support it.

~We can also remove the unstable protocol builder as it's part of wlroots.~ No we can't.

### DIFF
--- a/libqtile/backend/wayland/qw/output.c
+++ b/libqtile/backend/wayland/qw/output.c
@@ -1,7 +1,6 @@
 #include "output.h"
 #include "cairo-buffer.h"
 #include "layer-view.h"
-#include "proto/wlr-layer-shell-unstable-v1-protocol.h"
 #include "server.h"
 #include "session-lock.h"
 #include "util.h"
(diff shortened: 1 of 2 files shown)


elParaguayo:
This should probably be 4, not 5. The xml protocol file has this:
```
<interface name="zwlr_layer_shell_v1" version="4">
```

richcarni:
So v5 doesn't work?

elParaguayo:
Tests complete successfully with 5 but I suspect if a client actually asked for it and washed features from v5, it would fail.

richcarni:
> Tests complete successfully with 5 but I suspect if a client actually asked for it and washed features from v5, it would fail.

Ah, I see. Perusing wayland explorer it looks like 
- v4 adds zwlr_layer_surface_v1::keyboard_interactivity { on_demand }
- v5 adds zwlr_layer_surface_v1::set_exclusive_edge

Oh, so now looking at the issue flagged on discord, the warning makes sense
```
fuzzel says - warn: wayland.c:2133: keyboard-focus=on-demand requires zwlr_layer_shell_v1 version: 4, found version: 3
```

As far as I can tell, we haven't implemented on-demand keyboard focus though

elParaguayo:
OK. That's going to be a bit of a heavier lift to implement!

PR #5675, opened 2025-11-28
Implement save and load state commands

acidkewpie:
Added methods to save and load the qtile state from a file.

As I understand it we have the get_state exposed method yet I see nothing to do with it once we have it. so I created the load_state method to restore a layout from an arbitrary point in time. Given get_state output, when executed via an eval on cli needs to be processed anyway, it also seemed reasonable to add a save_state function so that its output is directly compatible with the load_state input.

For reference this came from finding idle code is not implemented in the new wayland backend and exploring workarounds I found when I directly turn my displays off and on again with wlr-randr, groups are no longer on screens so wanted to restore from before.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -31,6 +31,10 @@ PROTOS = [
         f"{QW_PROTO_IN_PATH}/wlr-layer-shell-unstable-v1.xml",
     ],
     ["xdg-shell-protocol.h", f"{WAYLAND_PROTOCOLS}/stable/xdg-shell/xdg-shell.xml"],
+    [
+        "pointer-constraints-unstable-v1-protocol.h",
+        f"{WAYLAND_PROTOCOLS}/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml",
+    ],
 ]
 
 QW_PROTO_OUT_PATH = QW_PATH / "proto"
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -607,7 +607,7 @@ class Core(base.Core):
     def masked(self) -> Generator:
         yield
         # Update pointer focus without cursor motion
-        lib.qw_cursor_update_focus(self.qw_cursor, ffi.NULL, ffi.NULL)
+        lib.qw_cursor_update_pointer_focus(self.qw_cursor)
         self._focus_pointer(motion=False)
 
     @property
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -1,4 +1,6 @@
 #include <stdlib.h>
+#include <wlr/types/wlr_relative_pointer_v1.h>
+#include <wlr/util/region.h>
 
 #include "cursor.h"
 #include "output.h"
@@ -19,14 +21,10 @@ void qw_cursor_destroy(struct qw_cursor *cursor) {
     free(cursor);
 }
 
-void qw_cursor_update_focus(struct qw_cursor *cursor, double *sx, double *sy) {
+// Pointer focus helper function
+static void update_pointer_focus(struct qw_cursor *cursor, struct wlr_surface *surface, double sx,
+                                 double sy) {
     struct wlr_seat *seat = cursor->server->seat;
-
-    struct wlr_surface *surface = NULL;
-    double tmp_sx = 0.0, tmp_sy = 0.0;
-
-    cursor->view = qw_server_view_at(cursor->server, cursor->cursor->x, cursor->cursor->y, &surface,
-                                     &tmp_sx, &tmp_sy);
 
     if (surface == NULL) {
         wlr_seat_pointer_clear_focus(seat);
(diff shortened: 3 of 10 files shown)


elParaguayo:
Thanks. Idle code for the wayland backend is coming soon.

tych0:
Can you describe your use case for this? Right now the format is not really a "public" API, but this would make it one.

acidkewpie:
Whilst as @elParaguayo says, idle code is on the way for the wayland backend, I'm looking at this today as being able to get the state out allows me to save the state before a script turns off my monitors. when it wakes up again, the state can be restored to put the right groups back on the right screens as before.

What felt more strange is that there's a function called "get_state". I'd ask what that exposed function is for? It doesn't seem to have any function as getting the pickled state out of qtile isn't seemingly any use at all as there's no way to get it back in again. So I created the load state, but then also felt that spitting the get_state output through jq wasn't really appropriate so then also felt a save_state might be useful.

The use case is currentlly relevant to me, but could well become obsolete. However I'd still think there's plenty of potential for it to be useful for others, certainly more so than get_state if I'm not missing something significant.

tych0:
> I'm looking at this today as being able to get the state out allows me to save the state before a script turns off my monitors. when it wakes up again, the state can be restored to put the right groups back on the right screens as before.

So the problem is something like: when reconfigure_screens() is called because you turn both monitors off, we "corrupt" the state of things because we delete one (or both) of the Screen objects, and then when you turn them back on, you want it to go back to the way it was.

I wonder: can your script do some hook unsubscription so reconfigure_screens() isn't called?

> What felt more strange is that there's a function called "get_state". I'd ask what that exposed function is for? It doesn't seem to have any function as getting the pickled state out of qtile isn't seemingly any use at all as there's no way to get it back in again.

Yeah, the only way right now is via `--with-state` when restarting qtile. I'd really like to keep it non-public (i.e. both producers and consumers are in our tree).

But I agree, it is weird to have this command exposed, and we probably should not have done that.

So I'd like to understand the use case better if it's not what I wrote above, so we can potentially design something more sane vs. exposing this pickled object mess.

acidkewpie:
So I'd cobbled together this script for something resembling idle support whilst it's being implemented:
```
#!/usr/bin/env bash

IDLE_SECONDS=$((15 * 60))   # 15 minutes
WLR_RANDR_STATE="/tmp/wlr-randr-state"
QTILE_STATE="/tmp/qtile-state"

while true; do
    # Watch input devices with timeout
    # Exit status: 0 = event occurred, 2 = timeout, 1 = error
    inotifywait -t "$IDLE_SECONDS" -qq -e access /dev/input/event* 2>/dev/null
    exit_code=$?
    
    if [ "$exit_code" -eq 2 ]; then
        # Timeout - idle period exceeded
        # Save current qtile state
        qtile cmd-obj -o cmd -f save_state -a "$QTILE_STATE"

        # Save current display state and turn all outputs off
        wlr-randr --json | tee "$WLR_RANDR_STATE" | jq -r '.[] | select(.enabled == true) | .name' | while read -r output; do
            wlr-randr --output "$output" --off
        done
        
        # Wait indefinitely for any input device activity
        inotifywait -qq -e access /dev/input/event* 2>/dev/null
        
        # Input detected, restore displays from saved state
        if [ -f "$WLR_RANDR_STATE" ]; then
            # First, turn on all previously enabled outputs
            jq -r '.[] | select(.enabled == true) | .name' "$WLR_RANDR_STATE" | while read -r output; do
                wlr-randr --output "$output" --on
            done
            
            # Then configure their modes, positions, and transforms
            jq -r '.[] | select(.enabled == true) | 
                "--output \(.name)" +
                (if .current_mode then " --mode \(.current_mode.width)x\(.current_mode.height)" else "" end) +
                (if .position then " --pos \(.position.x),\(.position.y)" else "" end) +
                (if .transform then " --transform \(.transform)" else "" end)
            ' "$WLR_RANDR_STATE" | while read -r cmd; do
                eval "wlr-randr $cmd"
            done
        fi
        
        qtile cmd-obj -o cmd -f load_state -a "$QTILE_STATE"
    else
        # Input detected while screens are on - sleep for a bit to avoid constant restarts
        sleep 60
    fi
done
```
I don't have another use case for the changes at present, but can imagine there might well be other reasons someone else does have one. I guess with full idle control is implemented, if you're sent a suitable event you can bypass the reconfigure, as you say, which would presumably eliminate my use case. 

Previously I was using wlopm to turn off my displays, but this didn't actually disable to the outputs like wlr-randr does, which I assume is why previously the screens never got messed around like I'm currently finding they are being.

tych0:
Does ec7fce6a3b6d466689736349c61ff60e21327904 help? Since we don't delete the bar and widgets from the screen objects, presumably they come back?

acidkewpie:
Not for me, I'm purely Wayland and don't use any bar or widgets at all. certainly "# ignore disconnected monitors" is a good comment to see though!

tych0:
The patch isn't wayland specific. 

But I suppose it won't help with saving state like you do here because it still removes the screen. 

I'm not opposed to merging this functionality but I think we should put a big warning that state restoration is not supported across versions of Qtile, and not considered a publicly consumable API.

PR #5678, opened 2025-11-29
Fix wayland follow_mouse_focus

RFCreate:
Since `follow_mouse_focus allows` allows for a string value (i.e. `click_or_drag_only`), then an explicit True value is needed when cursor enters a window

This PR solves this issue and adds a test to prevent future breaks in this config option.


tych0:
Thanks for the test, too!

PR #5679, opened 2025-11-30
[wayland] add idle notify/inhibit support

elParaguayo:
Adds support for wayland's idle-notify and idle-inhibit extensions.

Inhibitors can be created in a number of ways:
  - By the application itself
  - In users' configs (via a new `wl_idle_inhibitors` variable)
  - Dynamically (`lazy.window.add_idle_inhibitor()`)
  - Globally (`lazy.core.set_idle_inhibitor()`)

Inhibitors can be set to be active when a window is open, focused, visible or fullscreen.

This is draft for now as I need people to test it.

Known issues (to be updated as people report bugs):
- [ ] ~fd errors when exiting qtile (not 100% certain that's linked to this PR...)~ Not this PR.
- [x] formatting, mypy checks etc
- [x] missing tests
- [ ] missing documentation (could be in a later PR)
- [ ] add a widget to show inhibitor status

### DIFF
--- a/libqtile/backend/wayland/idle_inhibit.py
+++ b/libqtile/backend/wayland/idle_inhibit.py
@@ -1,0 +1,215 @@
+from __future__ import annotations
+
+from enum import IntEnum, auto
+from typing import TYPE_CHECKING
+
+from libqtile import hook
+from libqtile.log_utils import logger
+
+try:
+    from libqtile.backend.wayland._ffi import ffi, lib
+except ModuleNotFoundError:
+    from libqtile.backend.wayland.ffi_stub import ffi, lib
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Any
+
+    from libqtile.backend.base.window import Window
+    from libqtile.backend.wayland.core import Core
+    from libqtile.core.manager import Qtile
+
+
+class InhibitorType(IntEnum):
+    GLOBAL = auto()
+    OPEN = auto()
+    APPLICATION = auto()
+    VISIBLE = auto()
+    FOCUS = auto()
+    FULLSCREEN = auto()
+
+
+inhibitor_map = {
+    "open": InhibitorType.OPEN,
+    "visible": InhibitorType.VISIBLE,
+    "focus": InhibitorType.FOCUS,
+    "fullscreen": InhibitorType.FULLSCREEN,
+}
+
+
+class Inhibitor:
+    def __init__(
+        self,
+        qtile: Qtile,
+        window: Window | None = None,
+        handle: ffi.CData | None = None,
+        inhibitor_type: InhibitorType | None = None,
+        is_layer_surface: bool = False,
+        is_session_lock: bool = False,
+    ):
+        if inhibitor_type != InhibitorType.GLOBAL and (
+            (window is None and handle is None) or inhibitor_type is None
+        ):
+            raise ValueError("Inhibitor created with invalid arguments.")
+
+        self.qtile = qtile
+        self.window = window
+        self.handle = handle
+        self.inhibitor_type = inhibitor_type
+        self.is_layer_surface = is_layer_surface
+        self.is_session_lock = is_session_lock
+
+    def check(self) -> bool:
+        # If a session lock is active we only allow inhibitors from the session lock
+        if self.qtile.locked:
+            return self.is_session_lock and lib.qw_server_inhibitor_surface_visible(
+                self.handle, ffi.NULL
+            )
+
+        match self.inhibitor_type:
+            # Global inhibitor is always active
+            case InhibitorType.GLOBAL:
+                active = True
+            # Application-set inhibitors should apply when the client is visible
+            case InhibitorType.APPLICATION:
+                if self.window is not None:
+                    active = self.window.visible
+                elif self.is_session_lock or self.is_layer_surface:
+                    active = lib.qw_server_inhibitor_surface_visible(self.handle, ffi.NULL)
+                # We shouldn't really get here but, if we do, treat it as active
+                else:
+                    active = True
+            # Inhibitor is removed when client is killed so if it exists in the list then the client
+            # must be open
+            case InhibitorType.OPEN:
+                active = True
+            case InhibitorType.FOCUS:
+                active = bool(
+                    self.qtile.current_window and self.qtile.current_window == self.window
+                )
+            case InhibitorType.FULLSCREEN:
+                active = bool(self.window and self.window.fullscreen)
+            case InhibitorType.VISIBLE:
+                active = bool(self.window and self.window.visible)
+
+        return active
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Inhibitor):
+            raise NotImplementedError
+
+        return (
+            self.window,
+            self.handle,
+            self.inhibitor_type,
+            self.is_layer_surface,
+            self.is_session_lock,
+        ) == (
+            other.window,
+            other.handle,
+            other.inhibitor_type,
+            other.is_layer_surface,
+            other.is_session_lock,
+        )
+
+    def __repr__(self) -> str:
+        name = self.window.name if self.window else "no window"
+        mode = self.inhibitor_type.name.lower()
+        active = "ACTIVE" if self.check() else "inactive"
+        return f"<window={name} type={mode} status={active}>"
+
+
+class InhibitorManager:
+    def __init__(self, core: Core) -> None:
+        self.core = core
+        self.inhibitors: list[Inhibitor] = []
+
+    def set_hooks(self) -> None:
+        hook.subscribe.focus_change(self.check)
+        hook.subscribe.startup(self.update_user_inhibitors)
+
+    def sort(self) -> None:
+        self.inhibitors.sort(key=lambda x: x.inhibitor_type)
+
+    def add_inhibitor_safe(self, inhibitor: Inhibitor) -> bool:
+        if inhibitor not in self.inhibitors:
+            self.inhibitors.append(inhibitor)
+            self.sort()
+            self.check()
+            return True
+
+        return False
+
+    def remove_inhibitor(
+        self, remove: Callable[[Inhibitor], bool], skip_check: bool = False
+    ) -> bool:
+        old = len(self.inhibitors)
+        self.inhibitors = [o for o in self.inhibitors if not remove(o)]
+        removed = len(self.inhibitors) < old
+        if removed and not skip_check:
+            self.check()
+
+        return removed
+
+    def add_window_inhibitor(self, window: Window, inhibitor_type: str) -> None:
+        itype = inhibitor_map.get(inhibitor_type)
+        if itype is None:
+            logger.error("Unexpected inhibitor type {inhibitor_type}.")
+            return
+
+        inhibitor = Inhibitor(qtile=self.core.qtile, window=window, inhibitor_type=itype)
+        self.add_inhibitor_safe(inhibitor)
+
+    def remove_window_inhibitor(self, window: Window) -> None:
+        self.remove_inhibitor(lambda o: o.window == window)
+
+    def remove_window_inhibitor_by_wid(self, wid: int) -> None:
+        def match_window_by_wid(inhibitor: Inhibitor) -> bool:
+            win = inhibitor.qtile.windows_map.get(wid)
+            if not win:
+                return False
+            return win.wid == wid
+
+        self.remove_inhibitor(lambda o: match_window_by_wid(o))
+
+    def add_extension_inhibitor(
+        self,
+        handle: ffi.CData,
+        window: Window | None,
+        is_layer_surface: bool,
+        is_session_lock: bool,
+    ) -> bool:
+        inhibitor = Inhibitor(
+            qtile=self.core.qtile,
+            window=window,
+            handle=handle,
+            inhibitor_type=InhibitorType.APPLICATION,
+        )
+        return self.add_inhibitor_safe(inhibitor)
+
+    def remove_extension_inhibitor(self, pointer: ffi.CData) -> bool:
+        return self.remove_inhibitor(lambda o: o.handle == pointer)
+
+    def add_global_inhibitor(self) -> None:
+        inhibitor = Inhibitor(qtile=self.core.qtile, inhibitor_type=InhibitorType.GLOBAL)
+        self.add_inhibitor_safe(inhibitor)
+
+    def remove_global_inhibitor(self) -> None:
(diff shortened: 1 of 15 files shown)


tych0:
Can you paste the fd errors? I'm happy to eyeball those.

elParaguayo:
> Can you paste the fd errors? I'm happy to eyeball those.

```
 err: render.c:4363: TIOCSWINSZ: Bad file descriptor
 err: fdm.c:228: no such FD: 10
 err: fdm.c:228: no such FD: 13
 err: fdm.c:228: no such FD: 12
 err: fdm.c:228: no such FD: 11
 err: fdm.c:228: no such FD: 8
 err: fdm.c:228: no such FD: 9
 err: fdm.c:228: no such FD: 7
 err: spawn.c:113: /usr/lib/utempter/utempter: failed to spawn: Bad file descriptor
 err: fdm.c:228: no such FD: 6
warn: fdm.c:124: FD list not empty
```

EDIT: I'm actually seeing this on master branch too (I'm testing with `wephyr`) so it's not related to this PR.

libqtile/backend/wayland/qw/util.c
```diff
@@ -122,3 +122,45 @@ bool qw_surfaces_on_same_output(struct wlr_surface *surface_a, struct wlr_surfac
     }
     return false;
 }
+
+struct qw_view *qw_view_from_wlr_surface(struct wlr_surface *surface, bool *is_layer_surface, bool *is_session_lock_surface) {
+	struct wlr_xdg_surface *xdg_surface;
```
richcarni, line 131: I think `is_layer_surface` and `is_session_lock_surface` should be set false at beginning of function?

elParaguayo: Agreed. However, I suspect we'll merge your pointer constraints PR before this one so we can use the function from there!

libqtile/backend/wayland/qw/server.c
```diff
@@ -913,3 +972,37 @@ struct wlr_output *qw_server_get_current_output(struct qw_server *server) {
 
     return NULL;
 }
+
+void qw_server_idle_notify(struct qw_server *server) {
```
elParaguayo: Need to rename this as it's misleading. We're notifying *activity*.

elParaguayo:
@richcarni I've rebased this on to master but I've no idea what's going on with `util.c` it's saying there are changes but they look identical!

If tests pass here, I'll flatten the commits before we merge (assuming you're happy with this).

richcarni:
>I've rebased this on to master but I've no idea what's going on with util.c it's saying there are changes but they look identical!

git is so dumb sometimes! The other day I was trying to swap the order of two commits... what a drama!

richcarni:
Just been doing some basic testing. Everything seems to work great! The rules are very neat!

Only issue I experienced: it seems that if I add/change inhibitor rules, I need to restart qtile for them to take effect, reloading config isn't working for me

My inhibitors look like this:
```
wl_idle_inhibitors = [
    IdleInhibitor(
        match=Match(title="Alacritty"),
        when="focus"
    )
]
```

libqtile/hook.py
```diff
@@ -1083,6 +1083,26 @@ def play_resume_sound():
 
         """,
     ),
+    Hook(
+        "idle_inhibitor_change",
+        """
+        Called when the wayland backend's idle inhibitor state changes.
```
richcarni, line 1089: Nothing wrong, just wondering what this would be used for

idle_timeout and idle_resume seem like the most useful hooks - I know you were looking into those. If pretty involved maybe add in a separate PR?

elParaguayo:
> Just been doing some basic testing. Everything seems to work great! The rules are very neat!
> 
> Only issue I experienced: it seems that if I add/change inhibitor rules, I need to restart qtile for them to take effect, reloading config isn't working for me
> 
> My inhibitors look like this:
> 
> ```
> wl_idle_inhibitors = [
>     IdleInhibitor(
>         match=Match(title="Alacritty"),
>         when="focus"
>     )
> ]
> ```

Good point! Definitely need to look at this. I guess user-defined inhibitors should be removed and recreated but application and global ones are preserved.

libqtile/hook.py
elParaguayo: I was thinking it would be useful if you wanted some sort of notification that an inhibitor was active (eg a widget).

elParaguayo: I'm also not 100% convinced by adding the timeout and resume hooks. We can't neatly create a timeout subscription because the backend can't access the client functions exposed by wlroots so we'd need to have our own timers which just feels a bit messy. 

However, the alternative is to use an external client with user defined hooks but this is also a bit messy. 

I think the internal timer is probably the right way to go.

richcarni: I agree that both options sound messy. I would of thought that an external client would handle the majority of use cases - everything except calling custom qtile functions (though I guess this is part of the point of qtile :)

Is there definitely no way the backend could use the relevant protocol/interface as a client?

elParaguayo: We'd need to set up a connection to the compositor as a client and connect to the extension (i.e. write a minimal client) to do this. I can look into that but it might be more complicated than the alternative.

elParaguayo:
> Just been doing some basic testing. Everything seems to work great! The rules are very neat!
> 
> Only issue I experienced: it seems that if I add/change inhibitor rules, I need to restart qtile for them to take effect, reloading config isn't working for me
> 
> My inhibitors look like this:
> 
> ```
> wl_idle_inhibitors = [
>     IdleInhibitor(
>         match=Match(title="Alacritty"),
>         when="focus"
>     )
> ]
> ```

This should now be fixed.

I've also added support for adding/removing internal timers which fill fire hooks.

richcarni:
I was able to change from focus to visible, but when i tried to change back to focus it seems to still be on visible?

I think I'll let Chris from discord test the internal timer/hooks

elParaguayo:
> I was able to change from focus to visible, but when i tried to change back to focus it seems to still be on visible?
> 
> I think I'll let Chris from discord test the internal timer/hooks

Do you mean you changed this in the config and reloaded it?

EDIT: hmm. yes. It's not clearing old inhibitors.

elParaguayo:
Should be working now.

richcarni:
> Should be working now.

Confirmed, working great now!

elParaguayo:
Think this is good to go now.

richcarni:
I had an idea, a bit late, I'm sorry...

If I'm following correctly, the timeout/resume hooks are essentially using their own idle notifier, running in parallel to the external idle notifier provided by wlroots. One downside of this is that the hooks won't be called when using an external client like swayidle

What if we take wlroots' implemenation of ext-idle-notify-v1 (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_idle_notify_v1.c?ref_type=heads), rename it qw_idle_notify_v1, add events for timeout and resume (that we can use to fire our hooks), and use that in place of wlr_idle_notify_v1?

Ideally we could even upstream those modifications

elParaguayo:
> I had an idea, a bit late, I'm sorry...

Not a problem.

> If I'm following correctly, the timeout/resume hooks are essentially using their own idle notifier, running in parallel to the external idle notifier provided by wlroots. One downside of this is that the hooks won't be called when using an external client like swayidle

Your understanding is correct. However, why would you want hooks to fire on timers requested by clients like swayidle? The whole point is that swayidle world handle these.

> What if we take wlroots' implemenation of ext-idle-notify-v1 (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_idle_notify_v1.c?ref_type=heads), rename it qw_idle_notify_v1, add events for timeout and resume (that we can use to fire our hooks), and use that in place of wlr_idle_notify_v1?

I'm not averse to your idea as I don't like the separate timers. I would need to look through the wlr code to understand it better though 

> Ideally we could even upstream those modifications

Feels like we're adding qtile-specific code but, yes, if we think we can improve the implementation, we should consider upstreaming.

richcarni:
>Your understanding is correct. However, why would you want hooks to fire on timers requested by clients like swayidle? The whole point is that swayidle world handle these.

But swayidle can't call custom qtile code - I thought that's why we're adding the timers and hooks? We have similar hooks for suspend, resume, shutdown (events coming from outside qtile)

I think what we're implementing here currently is a completely separate idle notifier - it could actually be a generic (x11 + wayland) feature, as far as I can tell

>I'm not averse to your idea as I don't like the separate timers. I would need to look through the wlr code to understand it better though
>
>Feels like we're adding qtile-specific code but, yes, if we think we can improve the implementation, we should consider upstreaming.

It doesn't seem that complicated. The only change I'm proposing is to add a couple of events to expose timeout and resume. No qtile specific changes

Current signature:

```
struct wlr_idle_notifier_v1 {
	struct wl_global *global;

	struct {
		bool inhibited;
		struct wl_list notifications;

		struct wl_listener display_destroy;
	} WLR_PRIVATE;
};
```

Proposing:
```
struct wlr_idle_notifier_v1 {
	struct wl_global *global;

	struct {
			struct wl_signal timeout;
			struct wl_signal resume;
	} events;

	struct {
		bool inhibited;
		struct wl_list notifications;

		struct wl_listener display_destroy;
	} WLR_PRIVATE;
};
```

Actually, thinking on it some more I can see at least one complication: since it's possible to have multiple clients with different timers, I'm not sure how qtile would be able to differentiate which timer has gone off. I suppose we could return the swayidle arguments to help... starting to get messy again :(

elParaguayo:
As agreed with @richcarni on discord, I've removed custom idle timer support for now while we think about the best way to implement (plus implementing in x11)

elParaguayo:
@tych0 @richcarni  I've also been looking at adding support for equivalent behaviour in x11. The issue is detecting user activity in order to fire `resume` notifications. 

Currently, we only get mouse and key events to the extent that we've bound them. Anything else is not passed to qtile.

As I see it, we have 2 options:
1) Get raw events from XInput. I've tested this and it works well as notifications are instant but this comes at the cost of increased CPU as we're now receiving a load more events than before.
2) Use the ScreenSaver extension. This has the ability to query the `ms_since_user_input`. The advantage here is that the tracking is done in the xserver, not in our python loop. The disadvantage is we'd have to to query the server regularly but the notifications wouldn't be instant (I believe `xidlehook` uses this approach).

Thoughts?

richcarni:
I was initially going to say we can rule out option 1. right away, given the numbers you showed me (8% -> 30% CPU load)

But if it's only for to detect activity for resume, is it possible to only use this mode (getting raw events from XInput) while the system is idle, and then return to normal mode after resuming?

libqtile/backend/wayland/idle_inhibit.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class InhibitorManager:
+    def __init__(self, core: Core) -> None:
+        self.core = core
+        self.inhibitors: list[Inhibitor] = []
+
+    def set_hooks(self) -> None:
+        hook.subscribe.focus_change(self.check)
+        hook.subscribe.config_reloaded(self.update_user_inhibitors)
```
richcarni: To save creating a new hook, would `startup` or `startup_complete` hooks also work here?
Although I like that `config_reloaded` better expresses when we want to call this

richcarni:
Besides my comment above, with which I'm happy whatever you choose to do, is this one ready to go? I have tested together with output power management and that combo is working for me!

libqtile/backend/wayland/idle_inhibit.py
elParaguayo: The `idle_` hooks are different.

As for `startup`, that's a good point. I suppose it's very similar to the reload one.

elParaguayo: Yes - maybe the "startup" hook is enough. It may even be better than `config_reloaded` as it's fired every time the config is loaded.

elParaguayo:
> I was initially going to say we can rule out option 1. right away, given the numbers you showed me (8% -> 30% CPU load)
> 
> But if it's only for to detect activity for resume, is it possible to only use this mode (getting raw events from XInput) while the system is idle, and then return to normal mode after resuming?

This is a great idea. I assume by "normal mode" you mean polling the status via the screensaver extension? If so, yes, we can absolutely do this.

libqtile/backend/wayland/idle_inhibit.py
elParaguayo: I've deleted the `config_reloaded` hook and used `startup` instead.

elParaguayo:
I think this is good to go as is.

The user idle hook code will refactor this code a bit but it's still worth merging this for the time being.

richcarni:
> > I was initially going to say we can rule out option 1. right away, given the numbers you showed me (8% -> 30% CPU load)
> > But if it's only for to detect activity for resume, is it possible to only use this mode (getting raw events from XInput) while the system is idle, and then return to normal mode after resuming?
> 
> This is a great idea. I assume by "normal mode" you mean polling the status via the screensaver extension? If so, yes, we can absolutely do this.

I'm not really sure what I meant - I don't think I fully understood the problem

So now I'm looking at the screen saver extension (and I'm not at all familiar with x11 at that level) and wondering if we can get ScreenSaverNotify events without having to poll?

elParaguayo:
> I'm not really sure what I meant - I don't think I fully understood the problem

The problem is monitoring idle time and user activity. Easy in wayland but harder in x11 as we're not the server and we only see a subset of the total events handled by the server.

The screensaver has an idle time value that can be queried but I don't see a way to set a timeout and receive an event so we need to poll instead. This introduces some delay.

The alternative is we ask to receive all input events but this floods our event queue handling in python, pushing up CPU usage.

I think the solution is, to poll for the idle time and, as soon as we're idle (i.e. past the first requested timeout interval), we start polling for raw events. We can then disable this polling once the user is active again.

> So now I'm looking at the screen saver extension (and I'm not at all familiar with x11 at that level) and wondering if we can get ScreenSaverNotify events without having to poll?

The docs refer to some "core SetScreenSaver" method. But I can't find any other reference to it.

richcarni:
>The screensaver has an idle time value that can be queried but I don't see a way to set a timeout and receive an event so we need to poll instead. This introduces some delay.

I can't see a small delay before going into idle state being an issue. I'd be more concerned with polling overhead but assuming you're not polling more than once a second, I expect this is pretty small

>The docs refer to some "core SetScreenSaver" method. But I can't find any other reference to it.

Same!

Only other option I can think of is to write something that lives in c-land, so it can process xinput events more efficienctly, and send start_idle and resume events back to qtile

elParaguayo:
Agreed. A small deejay before entering idle should be ok.

Also, I don't think writing a c helper is the answer there. Might as well use xidlehook at that point.

PR #5681, opened 2025-12-02
[wayland] Add image-copy-capture and image-capture-source support

elParaguayo:
Looking at sway's implementation (https://github.com/swaywm/sway/pull/7976/files), this is all that's needed!

Will see if I can do some testing.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -34,6 +34,12 @@ PROTOS = [
     [
         "pointer-constraints-unstable-v1-protocol.h",
         f"{WAYLAND_PROTOCOLS}/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml",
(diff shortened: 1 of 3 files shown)


wineee:
This is sufficient for screen recording. Window recording still needs to wait for the release of wlroots 0.20.

It can be tested using [xdg-desktop-portal-wlr >= 0.8.0](https://github.com/emersion/xdg-desktop-portal-wlr/pull/334).

ervinpopescu:
I already had (full) screen recording on OBS? Am I missing something here? 

Does this allow to select a particular part of the screen?

elParaguayo:
OBS can do screen capture because it uses some xdg-desktop-portal. I'll be honest, I have no idea how that portal captures the screen without these extensions but I'm assuming wlroots exposes something to allow it to do so.

These extensions allow screen capture directly without going through the portal. I don't know how many clients currently use the extension but I don't see any downside to including in our compositor now.

PR #5682, opened 2025-12-03
feat: Add screen-output bindings

richcarni:
Add a config option to Screen to bind it to a specified output (display/monitor)

More than one output can be specified in a list

Any unmatched screens/outputs will be bound with no guarantee as to their order

### **This is a currently a super crude implementation for the purposes of getting early feedback**

The final implementation would try to match outputs to screens based on output preference, using screen order if multiple screen share the same preference.
Finally any unmatched outputs would be bound as they currently are - by screen order

Config looks like:
```
screens = [
    Screen(
        output="DP-2",
        bottom=bar.Bar(
            [
                widget.CurrentLayout(),
```

My first question is should this be a general (x11 + wayland) feature? I have assumed yes

I don't like that I have extended `ScreenRect` to "carry" output name. Maybe there should be a `ScreenInfo` that could include a `ScreenRect`? Maybe I am overthinking it?

Not sure what to do with aliased outputs? Not sure what aliased outputs are tbh

Other thoughts and suggestions?

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -512,9 +512,10 @@ class Core(base.Core):
     def get_screen_info(self) -> list[ScreenRect]:
         rects = []
 
-        @ffi.callback("void(int, int, int, int)")
-        def loop(x: int, y: int, width: int, height: int) -> None:
-            rects.append(ScreenRect(x, y, width, height))
+        @ffi.callback("void(int, int, int, int, char*)")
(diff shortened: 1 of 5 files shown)


elParaguayo:
Isn't this similar to #5274 ?

richcarni:
> Isn't this similar to #5274 ?

Yep, don't know how I missed that. Going to close this and see if I can help land that one

PR #5683, opened 2025-12-03
nix: update dependency names

Gurjaka:


tych0:
I don't really know how to review this -- can you add some links about why the dependency names have changed? Or were they just typos before?

Gurjaka:
> I don't really know how to review this -- can you add some links about why the dependency names have changed? Or were they just typos before?

I couldn’t find an upstream explanation in nixpkgs docs, so here’s my explanation why these changes are necessary.

• **`sphinx_rtd_theme` -> `sphinx-rtd-theme`**
This package was renamed in nixpkgs back in 2022. It didn’t emit warnings before, but `nix flake check` recently started flagging the old name. Commit reference:
[https://github.com/NixOS/nixpkgs/commit/4af35d92a0ceee11b8082befccbba2b036f5a536](https://github.com/NixOS/nixpkgs/commit/4af35d92a0ceee11b8082befccbba2b036f5a536)

• **`wrapGAppsHook`**
The old name was just a symlink. Using the real package name is the recommended pattern now.

• **`pkgs.system` removal**
`pkgs.system` used to be a symlink for `pkgs.stdenv.hostPlatform.system`, and that alias was removed. Relevant commit:
[https://github.com/NixOS/nixpkgs/pull/456527/commits/90cb7876446bf1684ffe40ab7832de0ec1b92991](https://github.com/NixOS/nixpkgs/pull/456527/commits/90cb7876446bf1684ffe40ab7832de0ec1b92991)

The goal of my PR is to apply these upstream changes consistently inside Qtile.

jwijenbergh:
Thanks! Looked good to me

tych0:
In the future, all of that writeup would be good in a commit message :)

Gurjaka:
> In the future, all of that writeup would be good in a commit message :)

Will consider doing so in future!

PR #5684, opened 2025-12-04
Pytest: throw error when using eval

RFCreate:
Right now, in the test suite when getting and error in an `eval()` function, it is silently ignored which make it harder to troubleshoot the issue, this adds a check to see if the test suite is active and rethrows the exception for `pytest` to stop the test and show the error.

Also rewritten some tests as there is no need to save and assert the first argument that is returned by `eval()` which just contains the status of the call, just kept the assertions to values in the second arguments.

### DIFF
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -549,14 +549,12 @@ def test_nextprevgroup_reload(manager_nospawn):
     manager_nospawn.c.reload_config()
     # Check that group has become unmanaged
     manager_nospawn.c.eval("self.new_group = self.current_group")
-    assert "True" == manager_nospawn.c.eval("self.old_group != self.new_group")[1]
+    assert manager_nospawn.c.eval("self.old_group != self.new_group") == "True"
     # Unmanaged group should not change the group in the screen
-    success, message = manager_nospawn.c.eval("self.old_group.screen.next_group()")
-    assert "True" == manager_nospawn.c.eval("self.new_group == self.current_group")[1]
-    assert success, message
-    success, message = manager_nospawn.c.eval("self.old_group.screen.prev_group()")
-    assert "True" == manager_nospawn.c.eval("self.new_group == self.current_group")[1]
-    assert success, message
+    manager_nospawn.c.eval("self.old_group.screen.next_group()")
+    assert manager_nospawn.c.eval("self.new_group == self.current_group") == "True"
+    manager_nospawn.c.eval("self.old_group.screen.prev_group()")
+    assert manager_nospawn.c.eval("self.new_group == self.current_group") == "True"
 
 
 @manager_config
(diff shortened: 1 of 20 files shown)


libqtile/command/base.py
```diff
@@ -319,7 +320,10 @@ def eval(self, code: str) -> tuple[bool, str | None]:
             except SyntaxError:
                 exec(code, globals_, locals())
                 return True, None
-        except Exception:
+        except Exception as e:
+            if os.environ.get("PYTEST_VERSION") is not None:
```
tych0: I like the idea, but this seems a little brittle to me, since we might lose this env var somehow and silently start failing again.

Given that all the code in the test suite essentially ignores the first parameter, can we change the function signature to return str|None, and always raise an exception instead?

RFCreate: How so? Can you provide an example.

tych0: Something like https://github.com/qtile/qtile/commit/5aa12f8374c4747d56bd4f16fcee17d75c284da4

I guess we probably would need some try/excepts in `qtile repl`, since `command.eval()` used to be doing that for us (maybe cmd-obj too?). But it seems nicer than starting to silently fail at some point in the future.

tych0: Oh, I guess we probably don't propagate exceptions across the IPC wire, so maybe the command client should do the exception catching? But then I guess we have to special case the selector in call() somewhere...

RFCreate: I don't know how common is to use `eval()` in the config, but changing the return type would mean a breaking change.

I could write a test that checks for a raise exception in `eval()`, so that if it starts to fail, it will alert the team and it won't go silently.

tych0: I bet it's not super common, presumably in configs people can just actually execute code they care about. But across the wire it seems like it would be more common, just like the test suite does. 

We can write a migration to fix people's configs though, and the wire format is "not stable", although I know there are some external users of it.

elParaguayo: I'm struggling to think of any reason why someone would need this in their config!

RFCreate: I am not familiar with the IPC system to tell if something would be affected other than the return type.

ervinpopescu: I do this in my config:
```
@lazy.function
def set_layout_current(qtile: Qtile):
    """toggle layout for current group between max and default"""
    groups = list(qtile.groups_map.keys())
    groups.pop()
    current_layout = qtile.current_screen.group.eval("self.layout.name")[1]
    current_group = qtile.current_screen.group.info()["name"]
    current_group_index = groups.index(current_group)  # type: ignore
    if current_layout == "max":
        qtile.current_screen.group.setlayout(group_layouts[current_group_index])  # type: ignore
    else:
        qtile.current_screen.group.setlayout("max")
```

The proposed change will break the indexing, right?

TBH I never liked the indexing caused by `(bool, str | None)` and I much more like the `Option` type (`str | None`).

PS: If I'm stupid and I can access .layout on the group, call me out (but I think I'm not, otherwise I wouldnt have this in my config :)).

elParaguayo: WTF?
`self.current_screen.group.layout.name`?

RFCreate: So should I go ahead with `return str | None`?

test/test_manager.py
```diff
@@ -1193,23 +1193,23 @@ def assert_dd_appeared():
     manager_nospawn.c.eval("self.test_data_config_evaluations = 0")
     manager_nospawn.c.reload_config()
     # should be readed twice (check+read), but no more
-    assert manager_nospawn.c.eval("self.test_data_config_evaluations") == (True, "2")
-    assert manager_nospawn.c.eval("len(self.keys_map)") == (True, "2")
-    assert manager_nospawn.c.eval("len(self._mouse_map)") == (True, "2")
+    assert manager_nospawn.c.eval("self.test_data_config_evaluations")[1] == "2"
+    assert manager_nospawn.c.eval("len(self.keys_map)")[1] == "2"
+    assert manager_nospawn.c.eval("len(self._mouse_map)")[1] == "2"
```
tych0: Do we still need the indexing here? I thought we'd be able to get rid of this?

RFCreate: I have not implemented it yet.

tych0: oh sorry, somehow I thought this was the updated branch.

RFCreate:
`eval()` now returns only the string value instead of a tuple.

tych0:
Thanks, I think this is a great cleanup!

PR #5640, opened 2025-11-09
wayland hidpi: Improve blurriness of internal windows

richcarni:
Scale cairo surfaces by their output's scale factor instead of having wayland upscale them. Since cairo largely uses vector drawing primitives, the result is much shaper

Fractional scale factors supported

Changes to statusnotifier widget break (oversized icons) qtile-extras statusnotifier without https://github.com/elParaguayo/qtile-extras/pull/459

PR #5685, opened 2025-12-05
wayland: Fix blurry widget icons on hidpi displays

richcarni:
Add some convenience functions to Drawer class to support painting images that are display scale aware

The technique is:
- Upscale images to their final size
- Downscale images to their logical size for compositing
- When the entire window is finally upscaled for display, since these last two scaling operations are on a recording surface they effectively cancel out

Widgets updated:
- [x] Statusnotifer (#5640)
- [x] Image
- [x] Volume
- [x] BatteryIcon
- [x] TaskList

I'm not sure it's obvious from the code what is going on. I find it difficult to explain but this is an attempt
[hidpi+icon_summary.pdf](https://github.com/user-attachments/files/24120961/hidpi%2Bicon_summary.pdf)

### DIFF
--- a/libqtile/backend/base/drawer.py
+++ b/libqtile/backend/base/drawer.py
@@ -423,11 +423,11 @@ class Dimension:
 
     @property
     def physical_size(self):
-        return self._logical_size * self._drawer._win.scale
+        return self._logical_size * getattr(self._drawer._win, "scale", 1)
 
     @property
     def physical_size_int(self):
-        return int(self._logical_size * self._drawer._win.scale)
+        return int(self._logical_size * getattr(self._drawer._win, "scale", 1))
 
 
 class TextLayout:
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -34,7 +34,7 @@ def get_cairo_surface_for_data(image):
     surf = cairocffi.ImageSurface.create_for_data(
         image.data, image.format, image.width, image.height
     )
-    return _SurfaceInfo(surf, None)
+    return _SurfaceInfo(surf, image.format)
 
 
 def get_cairo_pattern(surface, width=None, height=None, theta=0.0):
@@ -184,7 +184,10 @@ class Img:
         try:
             return self._default_surface
         except AttributeError:
-            surf, fmt = get_cairo_surface(self.bytes_img)
+            if self.bytes_img:
+                surf, fmt = get_cairo_surface(self.bytes_img)
+            elif self.image_data:
+                surf, fmt = get_cairo_surface_for_data(self.image_data)
             self._default_surface = surf
             return surf
 
@@ -252,7 +255,7 @@ class Img:
             if self.bytes_img:
                 surf, fmt = get_cairo_surface(self.bytes_img, self.width, self.height)
             elif self.image_data:
-                surf = get_cairo_surface_for_data(self.image_data)
+                surf, fmt = get_cairo_surface_for_data(self.image_data)
             self._surface = surf
             return surf
 


richcarni:
I want to try and update as many widgets as possible here, using as generic an approach as possible

elParaguayo:
Does this have to be done widget by widget? Is there no way to adapt the Img class?

richcarni:
> Does this have to be done widget by widget? Is there no way to adapt the Img class?

I don't think so, but I'm open to suggestions

What I've been trying to do is add these functions to Drawer that hide the output scale factor so callers don’t need to handle it directly. In simple cases, like volume and batteryicon, this seems fairly clean

But yeah, kind of an experiment, as I said open to suggestions. Was going to write up a summary of what I've actually done, that might be helpful, maybe tomorrow

elParaguayo:
Can you make the whole bar's drawer have this dpi aware scaling?

richcarni:
> Can you make the whole bar's drawer have this dpi aware scaling?

I'm not sure I understand what you mean. Maybe some concrete examples would be illustrative (also wrote up a short document with additional explanation, attached in initial comment). 

This is a hypothetical widget draw function:
```python
def draw(self):
	img = Img.from_path(self.filename)
	
	new_height = self.bar.size - (self.margin_top * 2)
	img.resize(height=new_height)

    self.drawer.ctx.save()
    self.drawer.ctx.translate(self.margin_x, self.margin_y)
    self.drawer.ctx.set_source(self.img.pattern)
    self.drawer.ctx.paint()
    self.drawer.ctx.restore()
    self.draw_at_default_position()
```

Then the most basic modifications to support hidpi:
```python
def draw(self):
	scale = getattr(self._win, "scale", 1) # Retrieve scale

	img = Img.from_path(self.filename)
	
	new_height = (self.bar.size - (self.margin_top * 2))
	img.resize(height=new_height * scale) # Up-scale

    self.drawer.ctx.save()
    self.drawer.ctx.translate(self.margin_x, self.margin_y)
    self.ctx.scale(1 / scale, 1 / scale) # Down-scale
    self.drawer.ctx.set_source(self.img.pattern)
    self.drawer.ctx.paint()
    self.drawer.ctx.restore()
    self.draw_at_default_position()
```

Instead with my Drawer functions (note no references to scale)
```python
def draw(self):
	img = Img.from_path(self.filename)
	
	new_height = self.bar.size - (self.margin_top * 2)
	pattern = self.drawer.dpi_aware_get_pattern(img, height=new_height)

    self.drawer.dpi_aware_paint_pattern(pattern, self.margin_x, self.margin_y)
    self.draw_at_default_position()
```

or
```python
def draw(self):
	img = Img.from_path(self.filename)
	
	new_height = self.bar.size - (self.margin_top * 2)
	img.resize(height=new_height)

    self.drawer.dpi_aware_paint_img(img, self.margin_x, self.margin_y)
    self.draw_at_default_position()
```

richcarni:
@elParaguayo, any suggestions on how to progress this?

Compared to what I've done here, do you prefer a few lines in each widget to 1. get scale, 2. up-scale and 3. down-scale? In some ways it's a smaller change and more explicit vs an extra layer of abstraction to hide the scale factor. And it's what I've already done for statusnotifer

Unfortunately, I don't believe there is any way to implement this without touching each widget that uses images in some way

elParaguayo:
I never particularly liked the way we draw an Img object in the first place.

Can we create a draw method in the Img class (which may need to receive the drawer instance as an argument) and handle the extra code there? Widgets then just call img.draw?

richcarni:
Hmm, so instead of
```python
img = Img.from_path(self.filename)
self.drawer.dpi_aware_paint_image(self.img, self.margin_x, self.margin_y)
```

something like
```python
img = Img.from_path(self.filename)
img.draw(self.drawer, xoffset=self.margin_x, yoffset=self.margin_y)
```
or as a one-liner
```python
img = Img.from_path(self.filename).draw(self.drawer, xoffset=self.margin_x, yoffset=self.margin_y)
```

I don't see any reason it couldn't be done that way. Let me try and implement something - I think in a clean branch - and we can see what that looks like

Thanks for the feedback

elParaguayo:
Yes. Probably not in a one liner as you only need to create the Img object once for each image. But I do think `img.draw` is a better api. It's clear what it's doing and the dpi stuff is hidden from users and handled automatically behind the scenes.

richcarni:
OK, begun a new attempt in a fresh PR: #5795 😅

I chose to start with the tasklist widget because it is one of the more complicated ones (it and statusnotifier)

richcarni:
Closed in favour of #5795

PR #5692, opened 2025-12-10
wayland: Don't mark non-internal windows killed

### LINKED ISSUE #5689: Killing transient window prevents from closing parent window
### Issue description

Whenever I kill a transient popup the parent associated to it fails to close afterwards if I try to kill it.

### Version

0.34.1.dev8+gf6b8f6c59

### Backend

Wayland (experimental)

### Config

```python
I make the transient float:

floating_layout = layout.Floating(border_focus="1d5663",
    float_rules=[
    	Match(func=lambda c: c.is_transient_for() is not None)
]
```

### Logs

```text

```

### Required

richcarni:
Clients may ignore a close request on a parent toplevel when it has an
open child. Since we can't detect success when closing a window, avoid
marking the window as killed, because this will prevent the parent from
being closed later

We will still track killed state for internal windows, which don't have
this issue, to prevent use after free

Fixes #5689

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -42,7 +42,6 @@ class Base(base._Window):
         # TODO: what is this?
         self.defunct = False
         self.group: _Group | None = None
-        self._killed = False
         self.visible = True
 
     def reparent(self, layer: int) -> None:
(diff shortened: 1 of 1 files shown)


tych0 (approved):

elParaguayo:
I had added those `_killed` lines as part of a PR about killing internal windows. I assume you're not seeing any issues here.

richcarni:
> I had added those `_killed` lines as part of a PR about killing internal windows. I assume you're not seeing any issues here.

Thanks, yes, this absolutely breaks killing internal windows

I suppose an obvious solution is to ignore `_killed` for everything other than internal windows. I'll think on it some

Issue #5694, opened 2025-12-10
[wayland] amixer repeatedly forked


richcarni:
### Issue description

I noticed my pids increasing faster than expected, even when the system is seemingly idle

Then running: `sudo bpftrace -e 'tracepoint:sched:sched_process_fork {printf("%llu ms %d -> %d %s\n", elapsed / 1000000, args->parent_pid, args->child_pid, comm);}'`

I see:
```
Attaching 1 probe...
39 ms 33303 -> 43286 python3
39 ms 43286 -> 43287 sh
42 ms 43287 -> 43288 amixer
43 ms 43287 -> 43289 amixer
246 ms 33303 -> 43290 python3
247 ms 43290 -> 43291 sh
250 ms 43291 -> 43292 amixer
251 ms 43291 -> 43293 amixer
455 ms 33303 -> 43294 python3
456 ms 43294 -> 43295 sh
459 ms 43295 -> 43296 amixer
460 ms 43295 -> 43297 amixer
663 ms 33303 -> 43298 python3
664 ms 43298 -> 43299 sh
667 ms 43299 -> 43300 amixer
668 ms 43299 -> 43301 amixer
871 ms 33303 -> 43302 python3
872 ms 43302 -> 43303 sh
875 ms 43303 -> 43304 amixer
876 ms 43303 -> 43305 amixer
1079 ms 33303 -> 43306 python3
...
```

With the same config, I see none of these amixer forks on the old backend

Will attempt to bisect

### Version

#8f4f545

### Backend

Wayland (experimental)

### Config

```python
widget.PulseVolume(limit_max_volume=True),
```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

richcarni:
Closed by #5696

PR #5696, opened 2025-12-11
widgets: separate Volume and PulseVolume

tych0:
Because PulseVolume inherits from Volume, it also inherits all of the timers, which means it still runs commands to find the volume, even if it can get them over the wire from pulse.

Instead, let's hoist the common code (mostly icon resolution, formatting, and rendering) to VolumeBase, and have each inherit from it with their own implementation for actually getting the volume.

### DIFF
--- a/libqtile/widget/pulse_volume.py
+++ b/libqtile/widget/pulse_volume.py
@@ -7,7 +7,7 @@ from libqtile import qtile
 from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
 from libqtile.utils import create_task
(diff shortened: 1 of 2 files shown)


richcarni:
Tested and fixes my issue in #5694

I still find it odd that this wan't previously an issue (for me) but see no reason not to proceed with the change

@elParaguayo, any objections?

elParaguayo:
I need to check `qtile-extras` as I inherit those widgets. Interestingly, I was having some issues with my widgets and had some ugly fixes. I wonder if this was the cause of them. 

EDIT: All good.

PR #5699, opened 2025-12-13
wayland: Move float above when floats_kept_above=True

### LINKED ISSUE #5688: [wayland] Firefox dialog for 'save image as' appears behind the window
### Issue description

Open firefox, click 'save image as' (possibly for other dialogs).

Dialog appears behind the window.

even with `floats_kept_above = True`

### Version

master

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

richcarni:
Mirroring the same checks in the x11 backend

Fixes #5688

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -657,6 +657,8 @@ class Window(Base, base.Window):
             else:
                 # if we are setting floating early, e.g. from a hook, we don't have a screen yet
                 self._float_state = FloatStates.FLOATING
+            if self.layer != lib.LAYER_KEEPABOVE and self.qtile.config.floats_kept_above:
+                self.keep_above(enable=True)
         elif (not do_float) and self._float_state != FloatStates.NOT_FLOATING:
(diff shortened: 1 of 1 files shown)


richcarni:
Question... What if the app is fullscreen? LAYER_KEEPABOVE is below LAYER_FULLSCREEN so it doesn't actually keep above in this instance

elParaguayo:
> Question... What if the app is fullscreen? LAYER_KEEPABOVE is below LAYER_FULLSCREEN so it doesn't actually keep above in this instance

I've wondered about that too. "Keep above" really just means keep above tiled windows. I think it makes sense for a fullscreen window to have priority. There's always "bring to front" which would put a window above a fullscreen.

elParaguayo:
Do you want to make any more changes here or are you happy for me to merge it?

richcarni:
I have a few including this one that I was planning to merge myself, but going to be offline for a few days over xmas so figured I'd hold off. Feel free to merge anything that's not in draft!

PR #5701, opened 2025-12-13
[wayland] convert keycodes to keysyms

### LINKED ISSUE #5700: [Wayland] Using keycodes in config breaks qtile
### Issue description

The keycode to keysym code is missing from the backend so using a keycode in the config results in qtile failing to start with the following error:
`AttributeError: cffi library 'libqtile.backend.wayland._ffi' has no function, constant or global variable named 'qwu_get_sym_from_code'. Did you mean: 'qwu_keysym_from_name'?`

### Version

Still present on latest commit (`9e6f1d6d1`)

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

elParaguayo:
Fixes #5700

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -560,12 +560,11 @@ class Core(base.Core):
 
         return rects
 
-    def _get_sym_from_code(self, keycode: int) -> str:
-        # TODO: test keycodes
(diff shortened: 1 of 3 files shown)


tych0:
Can you add your write up from the bug report to the commit message?

PR #5702, opened 2025-12-13
[Wayland] Compiler warnings

Gurjaka:
- Fix variable shadowing in cursor.c by reusing existing sx/sy variables
  instead of redeclaring them in inner scope
- Add missing includes in cursor.h for wayland-server-core.h, pixman.h,
  and stdint.h to resolve incomplete type errors
- Move qw_wallpaper_mode enum definition to server.h and include it in
  output.h instead of forward declaring (ISO C forbids enum forward refs)
- Remove unused wayland-server-protocol.h include from server.c
- Add void parameter to qw_server_create() for proper C function prototype

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -481,8 +481,8 @@ static void check_constraint_region(struct qw_cursor *cursor) {
             int nboxes;
             pixman_box32_t *boxes = pixman_region32_rectangles(region, &nboxes);
             if (nboxes > 0) {
-                double sx = (boxes[0].x1 + boxes[0].x2) / 2.;
-                double sy = (boxes[0].y1 + boxes[0].y2) / 2.;
+                sx = (boxes[0].x1 + boxes[0].x2) / 2.;
+                sy = (boxes[0].y1 + boxes[0].y2) / 2.;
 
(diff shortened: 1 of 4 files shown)


tych0:
Can you add whatever compiler flags you were using to the build so that we don't reintroduce these things?

Gurjaka:
> Can you add whatever compiler flags you were using to the build so that we don't reintroduce these things?

I had a commit that added the flags I was using, but I dropped it to avoid enforcing additional warnings/errors project-wide. I kept only the fixes.

For reference, these are the flags I used while working on this:
```
-Wall
-Wextra
-Wunused
-Wpedantic
-Wno-unused-parameter
-Wunused-result
-Wcast-align
-Wcast-qual
-Wduplicated-branches
-Wduplicated-cond
-Wformat=2
-Wstrict-prototypes
-Wunreachable-code
-Wno-discarded-qualifiers
-Werror=format-nonliteral
-Werror=int-conversion
-Werror=return-type
-Werror=vla-larger-than=0
-Werror=write-strings
-Werror=shadow
-Wno-unknown-pragmas
```

If you’d prefer these to be part of the build (or gated behind an option), I can re-add the commit.

cc @richcarni @jwijenbergh @elParaguayo

tych0:
I think they should be part of a CI build at the very least.

PR #5709, opened 2025-12-14
Also check wheel builds

tych0:


tych0:
Ok... this is getting substantially fancier with our github actions builds, but it does let us test the release process during CI, which would hopefully avoid last-minute failures like we had for 0.34.1. Thoughts?

tych0:
heh, this is making me wonder... what if we ran our test suite against the built wheels somehow? that would let us get rid of the ubuntu install stuff from the tree.

builder.py
```diff
@@ -30,6 +31,7 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
 
             ffi_compile(verbose=bool(wayland_requested))
         except Exception as e:
+            traceback.print_exc()
```
elParaguayo: Do we want this here? The idea is that the wayland build should be allowed to fail silently. We should definitely ignore the errors if the build only requested the x11 backend.

tych0: Yeah, I went ahead and moved it inside the wayland_requested check. I did find it was necessary to debug the issue we were seeing during the release, though.

PR #5712, opened 2025-12-15
fix prompt widget: exposed start input command

samunemeth:
Exposed `start_input()` command for the Prompt widget, as it is mentioned in the documentation, but not available.

It is mentioned in the documentation [here](https://docs.qtile.org/en/stable/manual/ref/widgets.html#prompt).
[Here](https://docs.qtile.org/en/stable/manual/commands/api/widgets.html#libqtile.widget.prompt.Prompt) it is nor present. (But I suspect that this is automatically generated.)
The documentation comment of this function also suggests that it is supposed to be exposed.

Without this command, the prompt widget is only usable for spawning applications with `lazy.spawncmd()`.
Please correct me if I'm misunderstanding the intended usage.

I wonder also: *Is `process_key_press()` also supposed to be exposed?*
(Also there is a small typo there, where the old name *minibuffer* is not changed to prompt, but I do not want to make this commit messy.)

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -413,6 +413,7 @@ class Prompt(base._TextBox):
         if self.bell_style == "visual":
             self.original_background = self.background
 
+    @expose_command()
     def start_input(
         self,
         prompt,


elParaguayo:
Thanks for this. What's your use case for exposing the command? It's really designed to be called internally. 

`start_input` just starts the prompt widget reading input and then calls its callback with the text as an argument. You can certainly use this for your own purposes but exposing via the API wouldn't be the way to do this.

Similarly, `process_key_press()` is not meant to be exposed either.

samunemeth:
Thanks for your quick reply.

I want to have a keybinding for power actions. I want it to focus this prompt, and run a function after that does actions with the input, in this case shutdown, reboot, suspend the computer based on what matches the input string. I did not find anything that accomplishes this. I can also imagine using it for switching keyboard layouts, or running bash commands that run in the terminal, not in their own window (`spawncmd()` does not work with them to the best of my knowledge). I can also imagine a small python calculator with this.

Should a take a different approach to implementing this, maybe based on the `lazy` object?

samunemeth:
(Sorry I messed with the commit a bit by accident, I'm still new to git pull requests.)

elParaguayo:
> Thanks for your quick reply.
> 
> I want to have a keybinding for power actions. I want it to focus this prompt, and run a function after that does actions with the input, in this case shutdown, reboot, suspend the computer based on what matches the input string. I did not find anything that accomplishes this. I can also imagine using it for switching keyboard layouts, or running bash commands that run in the terminal, not in their own window (`spawncmd()` does not work with them to the best of my knowledge). I can also imagine a small python calculator with this.
> 
> Should a take a different approach to implementing this, maybe based on the `lazy` object?

Yes - I don't think you want to be doing this by exposing the command. I think you'd do something like this:

```python

def prompt_callback(text):
    # Your code to do something based on the text input here

@lazy.function
def power_prompt(qtile):
    prompt = qtile.widgets_map["prompt"]
    prompt.start_input("power command", prompt_callback)

keys = [
...
Key([mod], "p", power_prompt),  # change binding as needed
...
]
```

samunemeth:
This seems to work, thank you very much!

I am unsure however why. What is inherently different from accessing the object in this way? I also can't really find documentation on this. After messing with this for a bit, and making what I wanted, and getting a better understanding of it, I will definitely look into updating the documentation, or adding some examples.

Thank you again for your quick help!

tych0:
I have often thought our prompt API is a little... obtuse. I think changing the docs from `Input should be started using the .start_input() method on this class.` to @elParaguayo's example at the very least would be an improvement.

ou need elParaguayo's `prompt_callback` to do the actual stuff, that seems required.

The rest of it IMO the ugly part:

```
@lazy.function
def power_prompt(qtile):
    prompt = qtile.widgets_map["prompt"]
    prompt.start_input("power command", prompt_callback)

keys = [
...
Key([mod], "p", power_prompt),  # change binding as needed
```
ideally this would look something like

```
Key([mod], "p", lazy.prompt(..., power_prompt))
```

Maybe we should add a helper that does this too? (And unifies the complaint about lack of a `prompt` widget in peoples' configs if it doesn't exist?). Looking through manager.py, there is a bunch of things that call the prompt widget, and all do ad-hoc checking for a prompt widget.

samunemeth:
A helper fuction seems like a descent idea, I will look into it.

I also found the solution to the completers strange, as they just look the objects up from a dictionary. This is annoying as I can't really create a completer without modifying the source code. But this is probably a bit of a stretch for me at the moment, I will get myself a bit more familiar before attempting something like that.

samunemeth:
Also, I'm not sure if this is the right place for this discussion. Feel free to close this. I will think about it, and open an issue or a new request if I have something to add.

tych0:
> I also found the solution to the completers strange, as they just look the objects up from a dictionary. This is annoying as I can't really create a completer without modifying the source code. But this is probably a bit of a stretch for me at the moment, I will get myself a bit more familiar before attempting something like that.

Yeah, there's no real reason to have those organized that way.

Honestly, this could all use reworked: it would be great to have one class representing each thing, with the function to actually perform the action and the completer living inside the class.

You'd probably have to think a bit about how exactly to organize it, since some completers want access to the qtile object, some of the do-the-thing functions close over the qtile object, etc., but it would be nice to have all the logic in the same place.

PR #5713, opened 2025-12-16
New cursor types for the prompt widget.

samunemeth:
I have added a new option `cursor_type` to the prompt widget. Depending on the value, the visual of the cursor is different. There are 4 options:

  - `line`: The original underline, and colored character.
  - `block`: A block cursor, that works similarly to the underline. It sets the background to the desired cursor color, and makes the foreground transparent, basically inverting the colors in this way. It supports blinking with `cursorblink`.
  - `none`: A cursor with no visual, except the color. This might be useful for a minimalist setup. Just the color is maybe enough if you only occasionally go back to edit stuff. It does not blink, respecting that the color did not blink in the original scenario. (Maybe it would be more feature full to make it blink, but I think it just makes it more confusing.)
  - `bar`: A vertical cursor bar simulated with a Unicode character. It looks nice if you have it at the end of the input, but the illusion break if you move it inside the text. Supports blinking, but the color only applies to the cursor itself and not to characters.

I think the `block` and `none` options are nice. I'm honestly a bit conflicted on the `bar` version. Please let me know what you think. I tried to execute some black magic with the `letter_spacing` argument I found in the [Pango markup refrence](https://docs.gtk.org/Pango/pango_markup.html) by setting it to a negative value. This does not cause any errors, and does shift the next character left, but getting the spacing right is probably impossible, as the width of a single character is not easily calculated from the height of said character. If it is possible, it would probably be really hacky.

I have not squashed the commits yet on purpose, so that I can separate them easily if I need to.

I have a few question:
  - Is the documentation style that I used acceptable?
  - Should I have some fallback value in the *if ... elif* statement? (I did not see any in similar cases, that is why I did it this way, but I'm happy to change it.)
  - Is the naming of the options alright? This seemed logical to me, but I guess it might be a bit confusing...

I have tested these changes in a virtual machine first, then on my main system with a lot of different setting combinations. I have not found any issues with it. Please mess around with it a bit and see if I have missed something.

### DIFF
--- a/libqtile/widget/prompt.py
+++ b/libqtile/widget/prompt.py
@@ -511,18 +511,19 @@ class Prompt(base._TextBox):
     def _update(self) -> None:
         if self.active:
             self.text = self.archived_input or self.user_input
-            cursor = pangocffi.markup_escape_text(" ")
             if self.cursor_position < len(self.text):
+                # Escaping after slicing to preserve the cursor position
                 txt1 = self.text[: self.cursor_position]
+                txt1 = pangocffi.markup_escape_text(txt1)
                 txt2 = self.text[self.cursor_position]
-                txt3 = self.text[self.cursor_position + 1 :]
-                for text in (txt1, txt2, txt3):
-                    text = pangocffi.markup_escape_text(text)
+                txt2 = pangocffi.markup_escape_text(txt2)
                 txt2 = self._highlight_text(txt2)
-                self.text = f"{txt1}{txt2}{txt3}{cursor}"
+                txt3 = self.text[self.cursor_position + 1 :]
+                txt3 = pangocffi.markup_escape_text(txt3)
+                self.text = f"{txt1}{txt2}{txt3}"
             else:
                 self.text = pangocffi.markup_escape_text(self.text)
-                self.text += self._highlight_text(cursor)
+                self.text += self._highlight_text(" ")
             self.text = self.display + self.text
         else:
             self.text = ""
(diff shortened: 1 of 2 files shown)


tych0:
Historically the different cursor types have indicated insert / delete modes, do you have plans to implement that?

samunemeth:
Good point, I was just thinking about the different visuals and was not planning the implement different editing modes. The underline that is now the default represents replace mode (in vim at least) but it works as insert mode in the current version. I'm not even sure that different editing modes would make sense for a short input field like this one. I hope that this is not an issue.

I have found an issue however, that if you have a "<" in the input and you move your cursor next to it, it messes up the Pango markup rendering. I will try to fix this, and the formatting issue that came up during the check in the following days. 

I have marked this as a draft for now, and will convert it back after the fixes.
If you have any more input please tell me, I'm still new to contributing.

samunemeth:
I have found an fixed the issue with the input character escaping. The `update` method included the following:
```python
for text in (txt1, txt2, txt3):
    text = pangocffi.markup_escape_text(text)
```
This does not modify the original variables. I have changed this, and also removed some redundant lines.

tych0:
Yeah, I was mostly just curious. I don't have a strong opinion about the editing modes.

Can you add a test for this as well?

samunemeth:
I will try my best, but it might take me some time to figure that out.

samunemeth:
Hello, I have added tests are requested. I have not created tests for Qtile yet, so feedback on them is greatly appreciated. If you think they look good, I will open the request again.

samunemeth:
Hi, sorry for dragging this out. I restructured the commits to be clearer and more concise. I also fixed the issues with formatting and validated it with pre-commit and ruff locally.

PR #5716, opened 2025-12-17
wayland: fix compiler warnings

Gurjaka:
- Fix variable shadowing in cursor.c by reusing existing sx/sy variables instead of redeclaring them in inner scope
- Add missing includes in cursor.h for wayland-server-core.h, pixman.h, and stdint.h to resolve incomplete type errors
- Move qw_wallpaper_mode enum definition to server.h and include it in output.h instead of forward declaring (ISO C forbids enum forward refs)
- Remove unused wayland-server-protocol.h include from server.c
- Add void parameter to qw_server_create() for proper C function prototype

Had to reopen #5702 since I migrated to codeberg, and accidentally deleted repo on github :D

### DIFF
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -481,8 +481,8 @@ static void check_constraint_region(struct qw_cursor *cursor) {
             int nboxes;
             pixman_box32_t *boxes = pixman_region32_rectangles(region, &nboxes);
             if (nboxes > 0) {
-                double sx = (boxes[0].x1 + boxes[0].x2) / 2.;
-                double sy = (boxes[0].y1 + boxes[0].y2) / 2.;
+                sx = (boxes[0].x1 + boxes[0].x2) / 2.;
+                sy = (boxes[0].y1 + boxes[0].y2) / 2.;
 
                 wlr_cursor_warp_closest(cursor->cursor, NULL, sx - view->x, sy - view->y);
 
--- a/libqtile/backend/wayland/qw/cursor.h
+++ b/libqtile/backend/wayland/qw/cursor.h
@@ -1,9 +1,13 @@
 #ifndef CURSOR_H
 #define CURSOR_H
 
+#include <pixman.h>
+#include <wayland-server-core.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_pointer_constraints_v1.h>
 #include <wlr/types/wlr_xcursor_manager.h>
+
+#include <stdint.h>
 
 struct qw_server; // Forward declaration to avoid circular dependency
 
--- a/libqtile/backend/wayland/qw/output.h
+++ b/libqtile/backend/wayland/qw/output.h
@@ -9,9 +9,7 @@
 #include <wlr/types/wlr_scene.h>
 #include <wlr/types/wlr_session_lock_v1.h>
 
-struct qw_server;
-
-enum qw_wallpaper_mode;
+#include "server.h"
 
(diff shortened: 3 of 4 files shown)


tych0:
Can you add the flags you were using to CI?

Gurjaka:
> Can you add the flags you were using to CI?

Not sure how to do so, I could share the flags I used though, if anyone can guide me with updating CI

tych0:
How did you compile with these to notice them? I would start by adding the flags there.

Gurjaka:
> How did you compile with these to notice them? I would start by adding the flags there.

I modified `build.py` file, but do we want that?

Gurjaka:
@tych0

I noticed these warnings while compiling the Wayland backend locally with a stricter build (extra warning flags enabled).

I agree that if we want to enforce warnings in CI, they should be added explicitly and intentionally.

Some of the flags I used are fairly aggressive and may not be appropriate for CI given generated code and `wlroots` headers, but I’m happy to:

1) Share the full set of strict flags I used locally
2) Work with maintainers to decide which subset (if any) makes sense to add to CI or the default build

Here are all the flags that I used:
```python
    WARNING_FLAGS = [
        "-Wall",
        "-Wextra",
        "-Wunused",
        "-Wno-unused-parameter",
        "-Wunused-result",
        "-Wcast-align",
        "-Wduplicated-branches",
        "-Wduplicated-cond",
        "-Wformat=2",
        "-Wstrict-prototypes",
        "-Wunreachable-code",
        "-Wno-discarded-qualifiers",
        "-Werror=format-nonliteral",
        "-Werror=int-conversion",
        "-Werror=return-type",
        "-Werror=vla-larger-than=0",
        "-Werror=write-strings",
        "-Werror=shadow",
        "-Wno-unknown-pragmas",
    ]
```

Sigmanificient:
Here is my take on this (for context, this is a set of flags that i use along my C projects, including one i stat with gurjaka):

Theses are mostly soft flags to follow strict practices (cast, conversion, const usage) and some of them i have actually never proc (like duplicated-branches, due to its niche nature).

I have also hand picked a few warning which i promoted to error (the flags that are prefixed with `-Werror=`) which can emit new build failures. I recommend taking the time to look into the detailed description from the gcc has they may not fit with your workflow (https://gcc.gnu.org/onlinedocs/gcc-15.1.0/gcc/Warning-Options.html). For instance, i like to ban the usage of VLAs, but this might not be the case for you (because it usually only matter within embedded code)

Some other flags that may be interesting as well:
- `-Wpedantic`
- `-Wstrict-aliasing=0`
- `-Wmissing-prototypes`

I think having a standard set is a must go, and that is should be old enough to have good portability (like c99). You may go with `gnu99` if you want to rely on some extensions. Or directly go with `_GNU_SOURCE` or setting `POSIX_STANDARD`

Lastly, although it is not a warning flag, you could also be consider `-DNEBUG` which i noticed to be missing (in release mode).

tych0:
> I think having a standard set is a must go, and that is should be old enough to have good portability (like c99). You may go with gnu99 if you want to rely on some extensions.

I agree on the standard set, and I probably care less about portability than most :).

I think that if we are going to patch warnings generated by compilers, they need to be errors in CI. Either we think they are important enough to land fixes and respect the warnings going forward, or we don't, and so we should not bother fixing them. This list seems like a fine start to that. Since CI runs builder.py, having them there is fine.

Maybe we want to have a --release flag to builder.py that disables all of these, and enables NDEBUG.

tych0:
Bump, it would be nice to have a standard set of warnings enabled in CI, anyone have time to look at this?

PR #5720, opened 2025-12-18
wayland: Fix formatting in util.c

richcarni:
Fix `// clandg-format on` directive so clang-format formats remainder of file


tych0:
Can we enforce clang-format as part of CI?

richcarni:
> Can we enforce clang-format as part of CI?

Do you mean clang-tidy? Actually I think my commit message isn't very clear

tych0:
I mean whatever autoformatter generated this diff 😃

richcarni:
clang-format is part of CI, unless you explicitly disable it for blocks of code within
```
// clang-format off

// clang-format on
```

The issue in util.c was a missing space between the `//` and `clang-format on`. So it wasn't turning back on for the rest of the file

*Updated commit message

tych0:
I see, ok, thanks!

PR #5722, opened 2025-12-19
User idle timers

elParaguayo:
~Very draft for now!!~

Adds custom idle timers for both X11 and Wayland. Timers can be created in the config as follows:

```python
from libqtile.config import IdleTimer

idle_timers = [
    # IdleTimer(timeout_in_seconds, action=None, resume=None, respect_inhibitor=True)
    # action and resume are optional (but you need at least one of them) and are any callable function (including async)
    # or a lazy command
    IdleTimer(300, lazy.spawn("my_screen_dimmer.sh"))
]
```

Things to do:
- [x] Fix tests
- [x] Fix formatting and mypy
- [x] Fix docs
- [x] Draft docs
- [ ] Feedback from users


libqtile/backend/x11/core.py
```diff
@@ -168,6 +171,11 @@ def __init__(self, display_name: str | None = None) -> None:
 
         self.last_focused: base.Window | None = None
 
+        self.create_idle_notifier()
+        self.create_idle_inhibitor_manager()
+
+        self._inhibited = False
```
tych0: This code all looks common to me between x11 and wayland; The base.Core doesn't have an `__init__` right now, but perhaps we should add one and call it from both places in this patch? There's probably some more stuff we could commonize.

libqtile/backend/x11/core.py
```diff
@@ -702,8 +713,10 @@ def handle_ButtonRelease(self, event) -> None:  # noqa: N802
         self.qtile.process_button_release(button_code, state)
 
     def handle_MotionNotify(self, event) -> None:  # noqa: N802
+        print("MOTION")
         assert self.qtile is not None
 
+
```
tych0: noise here, and a debug statement above.

libqtile/backend/wayland/core.py
```diff
@@ -785,48 +789,42 @@ def check_screen_fullscreen_background(self, screen: Screen | None = None) -> No
...
-        self.inhibitor_manager.add_global_inhibitor()
+        self.idle_inhibitor_manager.check()
 
-    @expose_command()
-    def remove_idle_inhibitor(self) -> None:
-        """Remove global idle inhibitor."""
-        self.inhibitor_manager.remove_global_inhibitor()
+    @base.Core.inhibited.setter
+    def inhibited(self, value: bool) -> None:
+        print(f"{value=} {self.inhibited=}")
```
tych0: some debug here?

libqtile/backend/x11/idle_notify.py
```diff
@@ -0,0 +1,80 @@
...
+        self.root = self.core._root.wid
+        timeout_ms = [t * 1000 for t in self.timeouts]
+
+        def get_idle():
+            return self.core.conn.mit_screen_saver.get_idle_time(self.root)
+
+        idx = 0
+        last_idle = get_idle()
+
+        while self.do_poll:
```
tych0: This means we are essentially outside xscreensaver's purview, so things like disabling dpms dynamically won't work. 

Can we instead do this without the loop, instead using `SelectInput(xscreensaver.Event.NotifyMask)`? That should generate a `xscreensaver.NotifyEvent` through the normal fd when we want to enable (or turn off) the screensaver, and we can start / kill it as appropriate.

As a bonus, I think it'll let us drop this extra event loop.

libqtile/backend/x11/xcbq.py
```diff
@@ -431,12 +490,13 @@ def __init__(self, display):
         self.cursors = Cursors(self)
         self.setup = self.conn.get_setup()
         extensions = self.extensions()
+        print(extensions)
```
tych0: some more debug?

libqtile/backend/x11/idle_notify.py
tych0: I have an LLM-generated translation of xss-lock into xcffib-speak that seems to work, let me know if you want that code. Though presumably you can do the same :)

elParaguayo: Would be quicker if you just share yours!

libqtile/backend/x11/core.py
elParaguayo: Yes - I was dealing without a common ``__init__``. Am focusing on getting things working and will tidy up after.

libqtile/backend/x11/idle_notify.py
elParaguayo: I wasn't convinced by xscreensaver (e.g. how do we set the timeout? can we have multiple timeouts?) but I'm not at all familiar with it.

tych0: > I wasn't convinced by xscreensaver (e.g. how do we set the timeout? can we have multiple timeouts?) but I'm not at all familiar with it.

You are right that we can't set the timeout via xcffib, since it doesn't look like that stuff is exported via xcb. There's an xlib example that we could just exec here: https://gitlab.freedesktop.org/xorg/app/xset/-/blob/master/xset.c?ref_type=heads#L1069 , i.e. `xset s 0`.

I can add the xcb-proto+xffib stuff for `{Set,Get}Screensaver` so we can control it without an exec, but for now it's probably fine to just exec xset.

You are right that there's only one timer. Maybe I'm old, but I can't imagine a use case for more than one timer. Perhaps we just raise an exception on the x11 backend if users specify more than one?

https://gist.github.com/tych0/b7263cb5d6c8e42cce902fb2132f3c89 is my LLM sample, you can test it with `xset s activate`, which should give you:

```
~/packages/xcffib master python3 idle_monitor.py
MIT-SCREEN-SAVER extension version: 1.1
Registered for screensaver events. Waiting for idle/active state changes...

Screensaver event: state=ON, kind=External, forced, time=134241235
  → User is now IDLE (screensaver activated)
Screensaver event: state=OFF, kind=External, not forced, time=134241284
  → User is now ACTIVE (screensaver deactivated)
```

tych0: Oh, they're there, they're just not in the screensaver extension, they're in `xproto` proper. So you should be able to do `conn.core.{Get,Set}ScreenSaver()` as xset does.

elParaguayo: The `swayidle` example shows multiple timeouts:
```
swayidle -w \\
	timeout 300 'swaylock -f -c 000000' \\
	timeout 600 'swaymsg "output * power off"' \\
		resume 'swaymsg "output * power on"' \\
	before-sleep 'swaylock -f -c 000000'
```
Admittedly, it's not the best example but I do think it's reasonable for users to want to have more than one. e.g. screen dim first then screen off/lock.

I wonder if we can`SetScreenSaver` and then set it again once the first timeout has been reached... (EDIT: or we count cycles after the screensaver is activated).

elParaguayo: If we do use `SetScreenSaver`, we will break the screensaver for users who also want idle timers. There's probably very very few users who may fall into that bracket but I'm still leaning towards maintaining our own timers but I accept it feels hacky.

tych0: > I accept it feels hacky.

It's not just about hacky: as we move away from the GIL in python, these multiple event loops will be free-threaded as real OS threads, so you'll have to coordinate reads/writes to objects across both of them. (This is one of the reasons I have been trying to get rid of our threads: so we don't have to deal with these kinds of bugs when free-threaded becomes a reality.)

> If we do use SetScreenSaver, we will break the screensaver for users who also want idle timers.

The flip side is that googling "how do I temporarily disable my screensaver" will give a working answer in all X11 DEs except Qtile, since we will be explicitly working around the screensaver extension.

I suppose it's a fine tradeoff if you want to make it to support these very very few users, but I am dubious :)

richcarni:
Had a chance to test on wayland and the following worked great
```
idle_timers = [
    IdleTimer(15, lazy.spawn("/home/rich/.cargo/bin/dpms-off")),
    IdleTimer(30, lazy.spawn("systemctl suspend")),
]
```

Only issue - I can't turn them off in the way I would expect to. If I remove idle_timers altogether or set it to `= []` and reload config they still fire. If I set different timers that seems to work, either with different times or setting action to None

elParaguayo:
> Had a chance to test on wayland and the following worked great
> 
> ```
> idle_timers = [
>     IdleTimer(15, lazy.spawn("/home/rich/.cargo/bin/dpms-off")),
>     IdleTimer(30, lazy.spawn("systemctl suspend")),
> ]
> ```
> 
> Only issue - I can't turn them off in the way I would expect to. If I remove idle_timers altogether or set it to `= []` and reload config they still fire. If I set different timers that seems to work, either with different times or setting action to None

Yes - haven't fixed that bit yet. Will do soon.

libqtile/backend/x11/idle_notify.py
elParaguayo: I'm not sure I follow the GIL point. We're not creating a new eventloop, theses are just happening in qtile's main loop.

Also, if a user sets idle timers, we can disable screensaver with `SetScreenSaver(0, 0, ...)` (I assume that's what `xset s off` is doing). We can add documentation to say that xscreensaver is disabled, or we can handle a special timer with a timeout of 0 to explicitly disable it. Do we also need to call `DPMSDisable`?

For what it's worth, `xidlehook` uses the same approach with multiple timers that I'm using here i.e. getting the amount of idle time and firing timers based on that.
```
xidlehook \
  `# Don't lock when there's a fullscreen application` \
  --not-when-fullscreen \
  `# Don't lock when there's audio playing` \
  --not-when-audio \
  `# Dim the screen after 60 seconds, undim if user becomes active` \
  --timer 60 \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness .1' \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness 1' \
  `# Undim & lock after 10 more seconds` \
  --timer 10 \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness 1; i3lock' \
    '' \
  `# Finally, suspend an hour after it locks` \
  --timer 3600 \
    'systemctl suspend' \
    ''
```

I'd prefer not to have differences between the x11 and wayland backends so configs just work the same, regardless of the backend.

Ultimately, if we decide to go down the route you want (a single timer and listening for screensaver events instead of polling idle timer), I'm ok with that. However, I want to be clear on the reasons I've gone with the approach that I have.

tych0: > I'm not sure I follow the GIL point. We're not creating a new eventloop, theses are just happening in qtile's main loop.

Oh duh, you are right of course. Reading too fast.

> Also, if a user sets idle timers, we can disable screensaver with SetScreenSaver(0, 0, ...) (I assume that's what xset s off is doing). We can add documentation to say that xscreensaver is disabled, or we can handle a special timer with a timeout of 0 to explicitly disable it. Do we also need to call DPMSDisable?

IMO, this is the wrong direction. I think we should integrate with x11's idle stuff, rather than writing our own polling model. But yes, you'd probably want to disable their DPMS stuff too if you're going to manage that yourself.

> For what it's worth, xidlehook uses the same approach with multiple timers that I'm using here i.e. getting the amount of idle time and firing timers based on that.

Yeah, I think that's reasonable. It just seemed complex to me for not a lot of users, but so long as it's not polling and using the idle timeout from X11, I think it's fine.

elParaguayo: OK. If we want to integrate with X11's idle stuff then we should just do the single timer and listen for the event. We can't do multiple timers based on just listening for events (setting a new screensaver interval after the first event is fired doesn't create a new activated event, but it does delay the cycle events).

tych0: huh, how does xidlehook work around that?

But yeah, I think it's fine to just error out on more than one.

tych0: I guess I don't even know what Cycle events really represent... maybe I should really go read the code.

tych0: >     If this bit is set, ScreenSaverNotify events are generated whenever the screen saver cycle interval passes.

Oh, duh. So this is the thing we actually care about. 

> (setting a new screensaver interval after the first event is fired doesn't create a new activated event, but it does delay the cycle events).

So isn't this the "right" behavior? In the xidlehook example,
```
  --timer 60 \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness .1' \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness 1' \
  `# Undim & lock after 10 more seconds` \
  --timer 10 \
    'xrandr --output "$PRIMARY_DISPLAY" --brightness 1; i3lock' \
    '' \
  `# Finally, suspend an hour after it locks` \
  --timer 3600 \
    'systemctl suspend' \
    ''
```

I think we would probably (?) specify this as:
```
IdleTimer(60, ...)
IdleTimer(70, ...)
```

And then we'd have to do some math to figure out when the next event is after the current idle timeout has fired?

elParaguayo: > huh, how does xidlehook work around that?

It doesn't, it polls the idle time like this PR does at the moment. 

> But yeah, I think it's fine to just error out on more than one.

elParaguayo: Nice. Let me test!

elParaguayo: Thanks for your patience here. I think this works: First timer sets `On` state. We can then set a new timeout which triggers a `Cycle` event. I'll go and rewrite this.

tych0: No worries, thanks for your patience with me too!

libqtile/backend/x11/core.py
elParaguayo: Have removed these lines as they made type checking messier than it needed to be

elParaguayo:
@richcarni can you test again and see if reloading config works now.

tych0 (approved):

tych0:
whoops, meant to comment with my review. This looks good to me, thanks for working on it. 

I will probably not have time to test until after Jan 1, but feel free to merge before then if you're happy with it.

tych0:
Somehow I can't comment on the lines directly but HAS_SCREENSAVER stuff in xcbq seems unnecessary to me. 

Are there builds that don't ship xcffib.screensaver? If so, I should fix that in xcffib. It feels like we shouldn't need this check in qtile.

elParaguayo:
> Somehow I can't comment on the lines directly but HAS_SCREENSAVER stuff in xcbq seems unnecessary to me.
> 
> Are there builds that don't ship xcffib.screensaver? If so, I should fix that in xcffib. It feels like we shouldn't need this check in qtile.

Yes - I've made a bit of a mistake in my code here. `xcffib.screensaver` should always exist as it's built from `xcbproto`. Checking for `xcffib.screensaver` therefore makes no sense.

However, we do still need to test if the extension is available (and, although unlikely, I suspect it's possible for the x server to be compiled without every extension) so I'll change the code.

ervinpopescu:
>> If I remove idle_timers altogether or set it to = [] and reload config they still fire.

> @richcarni can you test again and see if reloading config works now.

I've tested this and it's not working...

elParaguayo:
> > > If I remove idle_timers altogether or set it to = [] and reload config they still fire.
> 
> > @richcarni can you test again and see if reloading config works now.
> 
> I've tested this and it's not working...

Thanks - can you check on the latest commit?

libqtile/backend/x11/xcbq.py
```diff
@@ -477,6 +494,7 @@ class Connection:
         "xinerama": Xinerama,
         "randr": RandR,
         "xfixes": XFixes,
+        "mit-screen-saver": ScreenSaver,
```
tych0: Any reason not to just make this `mit_screen_saver` and not have to do the replace below?

richcarni:
> @richcarni can you test again and see if reloading config works now.

Tested on #35ba781

If I set `idle_timers = []` after having previously set some timers, they are disabled ✅

However, if I remove `idle_timers = ...`  completely from config, the timers still go off

elParaguayo:
That seems like a bug with how we reload the config. Will investigate.

elParaguayo:
So, as a test I made a file called `a.py`:
```python
a = 1
b = 2
```
Opened a terminal:
```
>>> from importlib import reload
>>> import a
>>> a.a
1
>>> a.b
2
```
I then edited `a.py` and removed the last line (so the file is just `a = 1`.
```
>>> reload(a)
<module 'a' from '/tmp/a.py'>
>>> a.a
1
>>> a.b
2
```
The variable still exists. I guess because it's in globals and not overwritten when reloading.

elParaguayo:
@richcarni As that scenario you identified is a separate issue, I would propose that we merge this one now.

PR #5724, opened 2025-12-21
docs: Update outdated Wayland discussion link

kamperemu:
Someone on discord told me the discussion link is outdated.
I think @m-col should also update the To Do List to link to the [Wayland project board](https://github.com/orgs/qtile/projects/2) to avoid any confusion.

### DIFF
--- a/docs/manual/wayland.rst
+++ b/docs/manual/wayland.rst
@@ -5,8 +5,8 @@ Running Qtile as a Wayland Compositor
 .. _wayland:
 
 Some functionality may not yet be implemented in the Wayland compositor.
-Please see the `Wayland To Do List <https://github.com/qtile/qtile/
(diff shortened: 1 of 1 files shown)


tych0:
Hmm, can you link to where the todo list is? I don't see it on that page somehow. But otherwise LGTM...

kamperemu:
> Hmm, can you link to where the todo list is? I don't see it on that page somehow. But otherwise LGTM...

Here's the link https://github.com/qtile/qtile/discussions/2409. It's the link I updated from in the commit. 
Just in case here's the new link as well https://github.com/orgs/qtile/projects/2.

tych0:
Oh, I missed that it was the old one. I closed it with a comment, thanks for the fix!

PR #5729, opened 2025-12-26
Add auto_group config variable [Draft]

MoonCubes:
This PR adds the `auto_group` variable to
the config.  Usually, programs like Brave
remember the workspaces of each of it's windows
after restarting. If this variable is set to
`False`, this is ignored and instead all windows
are opened in the current workspace.

This was discussed in the [discord](https://discord.com/channels/955163559086665728/1166312212223250482/1454161866342400216).

It's a draft for now, since I want to get feedback and possibly add some tests, to make sure it works as intended and doesn't break any existing setups that use hooks to manually change the group.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Qtile x.xx.x, released xxxx-xx-xx:
     * features
+    - Add `auto_group` config variable
     * bugfixes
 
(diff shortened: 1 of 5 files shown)


MoonCubes:
I also haven't tested it on Wayland yet.

PR #5739, opened 2025-12-29
Fix `reload_config` bugs

### LINKED ISSUE #5736: Two issues with `reload_config`
Posting this here so I don't forget (although I've started work on a fix already)...

There are two issues with our code for reloading the config:
1) Deleting a previously defined variable from the config and reloading does not remove the variable. This is a feature of how python's `reload` logic work - the module is reexecuted in its namespace so variables are overwritten with new values but are left unchanged if there's no new value.
2) Once this is fixed, our code looks for values in the user's config and, if it doesn't find them, it uses this `value = getattr(self, key, default[key])`. The effect of this is that it falls back to the previously defined value if there was one, rather than falling back to the default config (which should be the correct behaviour).

I'll post a fix when I have a chance to finish it.

elParaguayo:
There are two issues with our code for reloading the config:

1) Deleting a previously defined variable from the config and reloading does not remove the variable. This is a feature of how python's reload logic work - the module is reexecuted in its namespace so variables are overwritten with new values but are left unchanged if there's no new value.

2) Our code looks then looks for values in the user's config and, if it doesn't find them, it uses this value for a fallback `getattr(self, key, default[key])`. The effect of this is that it falls back to the previously defined value if there was one, rather than falling back to the default config (which should be the correct behaviour).

Fixes #5736

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -75,7 +75,8 @@ class Config:
             try:
                 value = settings[key]
             except KeyError:
-                value = getattr(self, key, default[key])
+                # Always fall back to default config value
+                value = default[key]
             setattr(self, key, value)
 
     def _reload_config_submodules(self, path: Path) -> None:
@@ -109,11 +110,18 @@ class Config:
 
         if name in sys.modules:
             self._reload_config_submodules(path)
-            config = importlib.reload(sys.modules[name])
-        else:
-            config = importlib.import_module(name)
 
-        self.update(**vars(config))
+            # Using reload updates the existing module namespace with new values.
+            # If a value is removed from the config, the old value is not removed.
+            # To fix this, we exec the config and store the values in a globals dict
+            config = {}
+            with open(self.file_path) as f:
+                code = compile(f.read(), self.file_path, "exec")
+            exec(code, config)
+        else:
+            config = vars(importlib.import_module(name))
+
+        self.update(**config)
 
     def validate(self) -> None:
         """


libqtile/confreader.py
```diff
@@ -109,11 +110,18 @@ def load(self):
...
             self._reload_config_submodules(path)
-            config = importlib.reload(sys.modules[name])
+
+            # Using reload updates the existing module namespace with new values.
+            # If a value is removed from the config, the old value is not removed.
+            # To fix this, we exec the config and store the values in a globals dict
+            config = {}
+            with open(self.file_path) as f:
+                code = compile(f.read(), self.file_path, "exec")
+            exec(code, config)
```
tych0, line 120: Is there a reason not to just delete all the members explicitly (perhaps by creating a whole new Config object)?

elParaguayo: That's what this is doing. If we do `importlib.reload` we reuse the existing module namespace. This puts it in a new one.

I'm open to alternative approaches as we agree on the principle.

elParaguayo:
Wow - this breaks a lot of stuff. Wasn't expecting that.

libqtile/confreader.py
tych0: I'm wondering if we can instead do something like `del sys.modules[name]`; I'm just weary of having two totally different codepaths for booting and reloading configs.

I guess we need something a bit fancier than that, because of `_reload_config_submodules()`; we can hopefully write a similar for loop and reload everything.

tych0: Or maybe we should just delete this codepath entirely, and reload configs would always re-exec?

PR #5740, opened 2025-12-30
wayland: plasma layout check payload when removing window

### LINKED ISSUE #5734: [wayland] plasma layout: black screen after killing fullscreened floating window
### Issue description

To reproduce:
- Open a window (e.g. alacritty) in _**plasma layout**_
- Make the window floating
- Make the window fullscreen
- Kill the window

I'm guessing this is something to do with the black background behind fullscreen windows

If I just fullscreen and kill I don't see this. Seems to be something to do with the combo of float -> fullscreen

**UPDATE:** Looks like this might be a me problem. I only see it on one of my three machines 🤔

**UPDATE 2:** Don't see it with default config - definitely a me problem

**UPDATE 3:** It's a plasma layout problem 😅

### Version

#553845b

### Backend

Wayland (experimental)

### Config

```python
floats_kept_above = True
```

### Logs

```text

```

### Required

richcarni:
When a plasma window is made floating, it is removed from the plasma layout. When this window is killed, qtile.unmanage() will call the layouts's remove function, but since the window has already been removed, it ends up calling remove on None

This issue manifests on wayland when the floating window has been fullscreened, because the fullscreen background is subsequently not removed

Fix by checking for None before calling remove

Fixes #5734

### DIFF
--- a/libqtile/layout/plasma.py
+++ b/libqtile/layout/plasma.py
@@ -944,7 +944,9 @@ class Plasma(Layout):
         self.add_mode = None
 
     def remove(self, client):
-        self.root.find_payload(client).remove()
+        payload = self.root.find_payload(client)
+        if payload is not None:
+            payload.remove()
 
     def configure(self, client, screen_rect):
         self.root.x = screen_rect.x


tych0:
Is this different semantics than the old wayland or x11 backends, or has this bug always been here?

elParaguayo:
This has always been there. Nothing to do with the new backend.

PR #5742, opened 2025-12-30
layouts: add scrollable column-based layout

Gurjaka:
Introduce a new Scroll layout with horizontally scrollable columns and vertically stacked windows. The layout is stable under insertion, keeps focused windows stationary unless explicitly moved, and separates floating windows from the tiled layer.

Includes commands for column navigation, window shuffling, and per-column width control.

### DIFF
--- a/libqtile/layout/__init__.py
+++ b/libqtile/layout/__init__.py
@@ -17,6 +17,7 @@ __all__ = [
     "MonadThreeCol",
     "MonadWide",
     "Zoomy",
+    "Scroll",
 ]
 
 from libqtile.layout.bsp import Bsp
@@ -27,6 +28,7 @@ from libqtile.layout.max import Max
 from libqtile.layout.plasma import Plasma
 from libqtile.layout.ratiotile import RatioTile
 from libqtile.layout.screensplit import ScreenSplit
+from libqtile.layout.scroll import Scroll
 from libqtile.layout.slice import Slice
 from libqtile.layout.spiral import Spiral
 from libqtile.layout.stack import Stack
--- a/libqtile/layout/scroll.py
+++ b/libqtile/layout/scroll.py
@@ -1,0 +1,674 @@
+from libqtile.backend.base import Window
+from libqtile.command.base import expose_command
+from libqtile.config import ScreenRect
+from libqtile.layout import base
+from libqtile.utils import ColorsType
+
+
+def flatten(lst):
+    for item in lst:
+        if isinstance(item, list):
+            yield from flatten(item)
+        else:
+            yield item
+
+
+class Scroll(base.Layout):
+    """
+    Scrollable 2D tiling system, niri-style.
+
+    Windows are arranged in columns that scroll horizontally. Within a
+    column, windows keep a natural height and the column scrolls
+    vertically when its windows overflow the screen, rather than always
+    shrinking every window to fit.
+    """
+
+    border_width: int
+    border_focus: ColorsType
+    border_normal: ColorsType
+    margin: int
+    column_width: int | None
+    columns_per_screen: int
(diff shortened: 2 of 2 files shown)


Gurjaka:
I don't understand why CI is failing ;-; Could I get some help?

elParaguayo:
CI runs a series of tests of every layout as behaviour should be identical.

Let me take a look for you.

elParaguayo:
I haven't really tested the layout yet but this should be enough for CI to pass:
```diff
diff --git a/libqtile/layout/scroll.py b/libqtile/layout/scroll.py
index 16a95e46..c4c22bd5 100644
--- a/libqtile/layout/scroll.py
+++ b/libqtile/layout/scroll.py
@@ -4,6 +4,14 @@ from libqtile.config import ScreenRect
 from libqtile.layout import base


+def flatten(lst):
+    for item in lst:
+        if isinstance(item, list):
+            yield from flatten(item)
+        else:
+            yield item
+
+
 class Scroll(base.Layout):
     """
     Scrollable 2D tiling system.
@@ -68,8 +76,8 @@ class Scroll(base.Layout):
             win.place(
                 screen_rect.x,
                 screen_rect.y,
-                screen_rect.width,
-                screen_rect.height,
+                screen_rect.width - self.border_width * 2,
+                screen_rect.height - self.border_width * 2,
                 self.border_width,
                 self.border_focus if win.has_focus else self.border_normal,
                 margin=self.margin,
@@ -250,7 +258,7 @@ class Scroll(base.Layout):
         elif col_idx < len(self.clients) - 1:
             return self.clients[col_idx + 1][0]
         else:
-            return self.focus_first()
+            return None

     def focus_previous(self, win: Window) -> Window | None:
         pos = self._get_window_position(win)
@@ -263,7 +271,7 @@ class Scroll(base.Layout):
         elif col_idx > 0:
             return self.clients[col_idx - 1][-1]
         else:
-            return self.focus_last()
+            return None

     @expose_command()
     def up(self) -> None:
@@ -509,3 +517,9 @@ class Scroll(base.Layout):
         if col_idx in self.client_column_widths:
             del self.client_column_widths[col_idx]
             self.group.layout_all()
+
+    @expose_command()
+    def info(self):
+        info = super().info()
+        info["clients"] = [c.name for c in flatten(self.clients)]
+        return info
```

To explain why this was needed:
- The configure change is unrelated to CI - I just noticed a single window displayed wrong
- Our tests look at the client list to check the layouts are working as expected. This is done via a call to `info()` so that must return the client list.
- `test_focus_cycle` then failed because `focus_next` and `focus_previous` were wrapping. They must return `None` when they reach the end as this is what tells `layout.group.next_window()` to move on to floating windows (https://docs.qtile.org/en/stable/manual/howto/layout.html#focus-next).

elParaguayo:
I'm pretty sure this layout is going to fail in a scenario where a user has a vertical bar and a window is scrolled off the side of the screen with the bar. Unless you crop the window, it's going to be drawn over the bar which is undesirable. This is why I'd been working on adding window clipping in my own scrolling layout.

EDIT: Just checked - it does cause the issue above.

Gurjaka:
> I'm pretty sure this layout is going to fail in a scenario where a user has a vertical bar and a window is scrolled off the side of the screen with the bar. Unless you crop the window, it's going to be drawn over the bar which is undesirable. This is why I'd been working on adding window clipping in my own scrolling layout.
> 
> EDIT: Just checked - it does cause the issue above.

maybe we could collab on this PR? :)

elParaguayo:
I've got clipping working in the new wayland backend and I should be able to dig out my old code for doing this in x11. Just need to work out the best way for windows to request clipping...

PR #5744, opened 2025-12-31
Fix test fd leak

elParaguayo:
This was a patch in Arch Linux's packaging for 0.34.1 (https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/commit/b84e3f071e63a07feba5548f51d86d12ca6a8bce)

@tych0 can you take a look at this one to see if it makes sense. I've just copied the patch verbatim.

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -187,6 +187,7 @@ class TestManager:
 
         def run_qtile():
             try:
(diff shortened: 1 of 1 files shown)


tych0:
Yep, thanks for this. Derp!

PR #5746, opened 2026-01-01
widgets/canto: fix lint

tych0:
it seems that pre-commit is not a required check for automerge, so this went in even with some formatting issues.

### DIFF
--- a/libqtile/widget/canto.py
+++ b/libqtile/widget/canto.py
@@ -1,10 +1,5 @@
-from subprocess import call
-
-from libqtile.widget import base
-
-from libqtile.widget.generic_poll_text import GenPollCommand
-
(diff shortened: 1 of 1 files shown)


tych0:
Huh, somehow I can't add pre-commit or nix flake to the list of required statuses? But I did just add python 3.14, so hopefully we won't have this problem there...

elParaguayo:
> Huh, somehow I can't add pre-commit or nix flake to the list of required statuses? But I did just add python 3.14, so hopefully we won't have this problem there...

I've added both `check` and `flake-check` to be required.

tych0:
Thank you! Not sure why I couldn't find those, I assume because of the crying baby 😂

PR #5750, opened 2026-01-01
Allow idle_inhibitors to run functions

elParaguayo:
A function returning `True` will treat the inhibitor as being active. The function will only be run immediately before an idle timer is due to be triggered.

### DIFF
--- a/libqtile/backend/base/idle_inhibit.py
+++ b/libqtile/backend/base/idle_inhibit.py
@@ -4,6 +4,7 @@ from enum import IntEnum, auto
 from typing import TYPE_CHECKING, Generic, TypeVar
 
 from libqtile import hook
+from libqtile.backend.base.window import Window
 from libqtile.log_utils import logger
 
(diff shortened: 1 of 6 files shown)


libqtile/backend/base/idle_inhibit.py
```diff
@@ -69,14 +75,24 @@ def check(self) -> bool:
                 active = bool(self.window and self.window.fullscreen)
             case InhibitorType.VISIBLE:
                 active = bool(self.window and self.window.visible)
+            case InhibitorType.FUNCTION:
+                if callable(self.function):
```
tych0, line 79: Do we need this check? Seems like we could let the try/catch below handle it? Otherwise, we should probably log an error. But: doesn't the type itself enforce that this is a Callable?

elParaguayo: Typing doesn't enforce anything but, you're right, the `try... except...` would catch any error.

PR #5758, opened 2026-01-06
bar: protect all widgets from individual draw failures

tych0:
If a single widget fails to draw, we should not abort this entire loop, otherwise all the other widgets will fail to draw.


tych0:
Yeah, I wonder how many times I've seen this myself and just fixed the widget crash and not even thought about this. Derp!

PR #5763, opened 2026-01-09
Net widget: fix crash on missing interface

### LINKED ISSUE #5737: [wayland] widget.Net doesn't grab interface at first boot
### Issue description

My wireguard interface is no longer identifed at first boot. Need to reload qtile. Probably a timing issue.

The problem has come up following wayc implementation. 

Similar behavior in the main branch.

### Version

0.34.2.dev17+g8eaf39dfe

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text
2025-12-29 17:22:11,396 ERROR libqtile net.py:poll():L171 Net widget errored while polling:
  File "/usr/lib/python3.13/site-packages/libqtile/widget/net.py", line 131, in poll
    down = new_stats[intf]["down"] - self.stats[intf]["down"]
           ~~~~~~~~~^^^^^^
KeyError: '8700g'
```

### Required

elParaguayo:
Net widget previously crashed if an inteface was unavailable. We now handle missing interface with a configurable option to warn user that the interace is not found.

Fixes #5737

### DIFF
--- a/libqtile/widget/net.py
+++ b/libqtile/widget/net.py
@@ -4,7 +4,6 @@ from math import log
 
 import psutil
 
-from libqtile.log_utils import logger
 from libqtile.widget import base
 
 
@@ -38,6 +37,11 @@ class Net(base.InLoopPollText):
             "format",
             "{interface}: {down:6.2f}{down_suffix:<2}\u2193\u2191{up:6.2f}{up_suffix:<2}",
             "Display format of down/upload/total speed of given interfaces",
+        ),
+        (
+            "missing_interface",
+            "{interface} not found",
+            "Display when interface is missing/unavailable.",
         ),
         (
             "interface",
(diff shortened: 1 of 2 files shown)


libqtile/widget/net.py
```diff
@@ -125,47 +129,54 @@ def get_stats(self):
...
+                    up_cumulative_suffix=up_cumulative_suffix,
+                    total=total,
+                    total_suffix=total_suffix,
+                    total_cumulative=total_cumulative,
+                    total_cumulative_suffix=total_cumulative_suffix,
                 )
+            )
 
-            return " ".join(ret_stat)
-        except Exception:
```
tych0, line 170: Nice, I wonder if there are a bunch of these that we can get rid of by doing this at the top level instead...

elParaguayo: Yes. By not returning a value after the exception, the widget stops polling. There will be plenty of scenarios where that's not desirable.

PR #5767, opened 2026-01-11
Add Screen name matching

richcarni:
Optionally match qtile Screens to physical outputs by name. For example DP-1, HDMI-1, etc.

Add some extra debug logging

Currently, we can match by monitor serial but not all monitors report this. Although output name may not be as stable (e.g. will change if ports are swapped) it seems to be stable enough in many cases to still be useful

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -116,13 +116,13 @@ def test_screen_serial_ordering_the_order(manager_nospawn, minimal_conf_noscreen
     assert manager_nospawn.c.screen[1].info()["serial"] == "b"
 
 
-def make_screen(name: str) -> Screen:
-    return Screen(serial=name, top=Bar([TextBox(name)], 10))
+def make_screen(name: str | None = None, serial: str | None = None, text: str = "") -> Screen:
+    return Screen(name=name, serial=serial, top=Bar([TextBox(text)], 10))
 
 
 def test_screen_serial_ordering_one_serial(manager_nospawn, minimal_conf_noscreen, monkeypatch):
     # one serial number is allowed, serial re-use overwrites to avoid confusion
-    minimal_conf_noscreen.screens = [Screen(), make_screen("one")]
+    minimal_conf_noscreen.screens = [Screen(), make_screen(serial="one", text="one")]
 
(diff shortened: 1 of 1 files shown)


libqtile/config.py
```diff
@@ -406,6 +406,11 @@ class Screen(CommandObject):
     config should be bound to. You can find this via ``get-edid -b $BUS |
     parse-edid``, or by looking at the sticker on the back of your monitor :).
     This is mostly useful for people with multi-monitor configs.
+
+    ``name`` is optionally the output name of the monitor this Screen's config
+    should be bound to. You can find this using utilities like ``xrandr`` and
+    ``wlr-randr``. This is mostly useful for people with multi-monitor configs.
+    If both serial and name are specified, serial will take precendence.
```
tych0: I would add a note here that at least for x11, names are not stable across boots and serial is probably preferred for that reason.

richcarni: Amended

I thought when you said this previously you meant "may" not be stable. It seemed odd that it would only be an X11 issue because I had assumed that these names were assigned at the kernel/driver level. Apparently it's more complicated than that

Not exactly a definite answer but from [StackExchange](https://unix.stackexchange.com/a/692721):
>It also appears that Wayland, which requires KMS capable drivers, will always show the same IDs, with the caveat that if you use Xwayland and xrandr, xrandr will show XWAYLAND0 ... type port IDs

tych0:
Can you add a test for this too?

libqtile/config.py
tych0: Yeah, my experience has been that it varies by driver the most. Some drivers are good about stable ordering, some are not. And of course, the same monitor plugged into different hardware will not be named the same thing. 

All this is to say: for most people the name will probably work most of the time, so I think this is an entirely reasonable PR.

richcarni:
Working on the tests and it turns out the serial over name precedence doesn't work like I expected/described - guess that's one reason we do tests

richcarni:
I decided to drop guarantees about precedence of matching. It gets complicated really quickly and I can only see it being an issue if you were to provide both a serial and output name in config where one of these is incorrect

PR #5768, opened 2026-01-11
start command: autodetect backend

tych0:
The presence of the WAYLAND_DISPLAY environment variable indicates that the user expects the wayland backend to be running. Let's just look for this, so people stop having to explicitly pass `--backend wayland`.

### DIFF
--- a/libqtile/scripts/start.py
+++ b/libqtile/scripts/start.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import locale
+import os
 from os import makedirs, path
 from pathlib import Path
 from sys import exit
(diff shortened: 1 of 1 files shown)


jwijenbergh:
hmmm but wouldn't this break people's configs? the display manager GDM e.g. runs on wayland so I think that WAYLAND_DISPLAY is set. But that doesn't necessarily mean that they want to run qtile in wayland mode. Haven't verified any of this myself though

tych0:
> the display manager GDM e.g. runs on wayland so I think that WAYLAND_DISPLAY is set.

That's not my understanding; my understanding was that the display manger was responsible for setting this variable for clients, and that whatever it uses internally is not exported. But I'm happy to be wrong if you test it and that's how it works :)

PR #5769, opened 2026-01-11
Don't finalise widgets twice

### LINKED ISSUE #5753: Excessive log noise when reloading config or exiting
I've been seeing these entries a lot in the log:
```
2026-01-01 16:51:20,053 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (layout_change) that was not currently subscribed.
2026-01-01 16:51:20,054 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (client_managed) that was not currently subscribed.
2026-01-01 16:51:20,054 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (client_urgent_hint_changed) that was not currently subscribed.
2026-01-01 16:51:20,054 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (client_killed) that was not currently subscribed.
2026-01-01 16:51:20,055 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (setgroup) that was not currently subscribed.
2026-01-01 16:51:20,055 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (group_window_add) that was not currently subscribed.
2026-01-01 16:51:20,055 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (current_screen_change) that was not currently subscribed.
2026-01-01 16:51:20,055 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (changegroup) that was not currently subscribed.
2026-01-01 16:51:20,056 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (client_name_updated) that was not currently subscribed.
2026-01-01 16:51:20,056 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (focus_change) that was not currently subscribed.
2026-01-01 16:51:20,056 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (float_change) that was not currently subscribed.
2026-01-01 16:51:20,056 libqtile hook.py:_subscribe():L129  Tried to unsubscribe a hook (current_screen_change) that was not currently subscribed.
```

Looks like we're trying to finalise widgets twice.

elParaguayo:
`Qtile._finalize_configurables` finalises every widget in `widgets_map`. It then finalises every `Bar` and `Gap` instance. `Bar.finalize` will try to finalise every widget in the bar. This can result in log noise where widgets have subscribed to hooks and tries to unsubscribe having already unsubscribed.

Widgets have a `finalized` flag so let's just check the widget hasn't been finalised before calling `finalize()`.

Fixes #5753

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -349,7 +349,8 @@ class Qtile(CommandObject):
         """
         try:
             for widget in self.widgets_map.values():
-                widget.finalize()
+                if not widget.finalized:
+                    widget.finalize()
             self.widgets_map.clear()
 
             # For layouts we need to finalize each clone of a layout in each group
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -410,7 +410,8 @@ class Bar(Gap, configurable.Configurable, CommandObject):
         if self.future:
             self.future.cancel()
         for widget in self.widgets:
-            widget.finalize()
+            if not widget.finalized:
+                widget.finalize()
         if hasattr(self, "drawer"):
             self.drawer.finalize()
             del self.drawer


jwijenbergh:
Just noticed the log on my machine too, thanks for the fix

libqtile/core/manager.py
```diff
@@ -349,7 +349,8 @@ def _finalize_configurables(self) -> None:
         """
         try:
             for widget in self.widgets_map.values():
-                widget.finalize()
+                if not widget.finalized:
```
tych0, line 352: Can we just delete this code now, since the bars are going to do it for us? Then we don't need to check the flag, because it should only happen once?

elParaguayo: I thought about that and may have got it wrong.

I had originally thought we needed both because the `WidgetBox` widget adds and removes widgets from the bar so we can't rely on the bar to tidy up. However, we can just make sure the widgetbox finalises the widgets in the box if they're not in the bar.

I can do a separate PR.

elParaguayo: ...although, in qtile-extras, it's possible to add widgets to a pop-up window. They aren't finalised with the bar but would be finalised if we iterate over the widgets_map dict. It's probably preferable for them _not_ to be finalised at the same time and only finalised when the popup is killed.

tych0: It seems like qtile should finalize popups, and popups should finalize whatever they have created, vs. directly iterating over the widgets map in manager.py. The idea being that each piece of code that creates a resource should have a corresponding release of that resource close by.

elParaguayo: Yes, completely agree. #5773 is a go at doing this.

PR #5770, opened 2026-01-11
Wayland: Implement core.get_inputs

jwijenbergh:
The returned dict has the same format as the old backend


jwijenbergh:
Thanks Rich!

PR #5771, opened 2026-01-12
Some command object typechecking

tych0:
I would really like to drop our `cmd_` legacy wrapper: https://github.com/tych0/qtile/commit/adad8d1411b5ed854d89638938b0a6fff04b7003. When we defined `__getattr__`, we lost a lot of type checking there all throughout the tree, viz. all the fixes in this branch.

There are still ~60 more errors produced by `make lint`; most of them just inheritance issues or whatever, but some of them are real bugs too I think.

Anyway, here's the first set that I worked on tonight.

### DIFF
--- a/libqtile/backend/x11/core.py
+++ b/libqtile/backend/x11/core.py
@@ -796,7 +796,7 @@ class Core(base.Core):
     def handle_UnmapNotify(self, event) -> None:  # noqa: N802
         assert self.qtile is not None
 
-        win = self.qtile.windows_map.get(event.window)
+        win = self.qtile.lookup_client(event.window)
 
         if win and getattr(win, "group", None):
             try:
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -815,8 +815,8 @@ class Qtile(CommandObject):
             if not win.group and self.current_screen.group:
                 self.current_screen.group.add(win)
 
-        # Check if any user-defined inhibitor rules match the window
-        win.add_config_inhibitors()
+            # Check if any user-defined inhibitor rules match the window
+            win.add_config_inhibitors()
 
         hook.fire("client_managed", win)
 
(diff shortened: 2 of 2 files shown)


libqtile/backend/base/idle_inhibit.py
```diff
@@ -154,5 +154,5 @@ def update_user_inhibitors(self) -> None:
         self.remove_inhibitor(
             lambda o: o.inhibitor_type != InhibitorType.APPLICATION, skip_check=True
         )
-        for win in self.core.qtile.windows_map.values():
+        for win in self.core.qtile.windows.values():
```
elParaguayo: I don't think this is right. `Qtile.windows` is a method which provides a dict of strings.

tych0: Oof, you are quite right. I wonder why I didn't get a type error here. (Maybe I did and I just missed it.) I'll fix that, thanks.

tych0:
> However, how is any of this related to the cmd_ prefix point in the PR description?

Dropping the `cmd_` wrapper allows us to delete the `__getattr__` method in CommandObject. That deletion triggers additional type checking, which does not pass currently because we have never been enforcing it. If you cherry pick that patch and do a `make lint`, you should see a bunch of errors.

I didn't actually realize implementing `__getattr__` caused mypy to drop the object from it's type checking. Of course it makes sense in hindsight, but it does suggest that we should avoid overriding `__getattr__` wherever possible.

As an aside, I have been thinking about how we could do more type checking for peoples' configs to catch user errors earlier, and it seems like getting rid of the `__getattr__` override in Configurable is probably at the top of the todo list, but it looks like there are several others in the tree that would be fairly useful to get rid of as well.

libqtile/backend/base/idle_inhibit.py
elParaguayo: This is also quite confusing when you have things like `Group.windows` which is not an exposed method but is just an attribute that does return a list of the window instances in that group.

elParaguayo:
I think some of the layouts override that method too. `ScreenSplit` and `Slice`?

tych0:
Yeah, looks like there are a handful. I haven't investigated so far beyond the patches in this series.

libqtile/backend/base/idle_inhibit.py
tych0: > This is also quite confusing when you have things like Group.windows which is not an exposed method but is just an attribute that does return a list of the window instances in that group.

Yeah... probably some API cleanup to do here.

PR #5773, opened 2026-01-12
Better finalisation of widgets

elParaguayo:
Widgets should be finalised by the object that created them so we should never iterate over widgets_map blindly and finalise everything.

### DIFF
--- a/libqtile/widget/base.py
+++ b/libqtile/widget/base.py
@@ -232,6 +232,13 @@ class _Widget(CommandObject, configurable.Configurable):
             self.layout.finalize()
             self.layout = None
         self.drawer.finalize()
+
+        # Remove finalised widget from the widgets_map
+        for name in list(self.qtile.widgets_map.keys()):
+            if self.qtile.widgets_map[name] is self:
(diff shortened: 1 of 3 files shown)


libqtile/widget/base.py
```diff
@@ -232,6 +232,13 @@ def finalize(self):
             self.layout.finalize()
             self.layout = None
         self.drawer.finalize()
+
+        # Remove finalised widget from the widgets_map
```
tych0, line 236: Are you sure we want this part? It looks suspiciously similar to what I dropped in ec7fce6a3b6d466689736349c61ff60e21327904

elParaguayo: I removed `self.widgets_map.clear()` so I'm now just removing a widget when it's finalised.

What issue did you have?

Note: I've not looked at the test errors yet.

tych0: Yeah, sorry, just looked similar. I think it won't actually be an issue: you're not dropping things from the bar, but from the core's widgets_map (which, IMO, should proabably be removed... but that's a different patch series).

PR #5777, opened 2026-01-13
Update default_config.py

Sinaaaa:
Added a helpful reminder so that users would know that to achieve consistent results, they also need to set XCURSOR_SIZE & XCURSOR_THEME env vars when theming their cursors in qtile wayland.

I thought the default config is a good place to put this, since changing just the settings already in the default config may lead to inconsistent results & bug reports.

### DIFF
--- a/libqtile/resources/default_config.py
+++ b/libqtile/resources/default_config.py
@@ -195,6 +195,8 @@ auto_minimize = True
 wl_input_rules = None
 
 # xcursor theme (string or None) and size (integer) for Wayland backend
+# If deviating from the default values it's recommended to also set
+# the XCURSOR_THEME & XCURSOR_SIZE env vars before running qtile.
 wl_xcursor_theme = None
 wl_xcursor_size = 24
 


libqtile/resources/default_config.py
```diff
@@ -195,6 +195,8 @@
 wl_input_rules = None
 
 # xcursor theme (string or None) and size (integer) for Wayland backend
+# If deviating from the default values it's recommended to also set
+# the XCURSOR_THEME & XCURSOR_SIZE env vars before running qtile.
```
elParaguayo, line 199: Which takes precedence? The config values or the environmental variables?

Sinaaaa: > Which takes precedence? The config values or the environmental variables?

A clear precedence cannot be established. I have done some testing on the minimal config. There are some things for which the qtile settings don't work, cursor in rofi for example & alternatively env vars alone don't work for everything, for example hovering the cursor over window borders or just over the wallpaper are not affected by env vars. 

I've been experimenting with wlroots WMs on and off for over 3 years & cursor theming always has been a mess, not even labwc & sway has this quite right. (there are additional overriding factors for cursor sizes & themes  like GTK theming and flatpak sandboxes as well)

jwijenbergh: the config values take precedence because qtile sets the env variables

jwijenbergh: so I think this comment is not needed? see https://github.com/qtile/qtile/blob/b07efcb0073876eee0ba4a4b7b98b7429f783494/libqtile/backend/wayland/qw/cursor.c#L589-L592

Sinaaaa: > so I think this comment is not needed? see
> 
> https://github.com/qtile/qtile/blob/b07efcb0073876eee0ba4a4b7b98b7429f783494/libqtile/backend/wayland/qw/cursor.c#L589-L592

Yes the comment is not needed, but for some reason this doesn't work quite right then..

jwijenbergh: Do you see a difference when you set the env vars manually? Maybe make an issue?

Sinaaaa: > Do you see a difference when you set the env vars manually? Maybe make an issue?

Yes there is a difference and I should make an issue indeed. I regret not starting with that.

PR #5780, opened 2026-01-13
Made test independent of Xterm.

samunemeth:
I am on a mission to make the tests work on NixOS using the provided flake. These changes were made to help with this goal, but are also good practice in other cases.

Some tests were dependent on Xterm being available. It was used as a dummy window. I changed them to use the `guess_terminal()` function instead. (This was discussed in Discord with @Gurjaka and @jwijenbergh)

One small change was made to a test launching an external python script with extra environment variables. It was not adding new environment variables, but overriding all of the existing ones. Therefore it would work if no `env` was provided to the function but would not work if `env` was provided. Now it adds the `env` dictionary to the default ones, with the parameter being the stronger one, and therefore overriding the existing ones if present.

I tested all of these changes locally.

I have not squashed the commits on purpose, as I believe they form better small change units this way, but I am happy to merge them if you find that better.

### DIFF
--- a/test/test_dgroups.py
+++ b/test/test_dgroups.py
@@ -16,10 +16,13 @@ class DGroupsConfig(libqtile.confreader.Config):
     screens = []
 
 
+terminal = libqtile.utils.guess_terminal()
+
+
 class DGroupsSpawnConfig(DGroupsConfig):
     groups = [
         libqtile.config.Group("a"),
-        libqtile.config.Group("b", spawn=["xterm"]),
+        libqtile.config.Group("b", spawn=[terminal]),
     ]
 
 
--- a/test/test_manager.py
+++ b/test/test_manager.py
@@ -15,11 +15,14 @@ from libqtile.config import Match
 from libqtile.confreader import Config
 from libqtile.group import _Group
 from libqtile.lazy import lazy
+from libqtile.utils import guess_terminal
 from test.conftest import dualmonitor, multimonitor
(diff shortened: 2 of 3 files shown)


test/backend/x11/test_x11_identify_output.py
```diff
@@ -4,14 +4,13 @@
 import sys
 
 
-def run_identify_output(env=None):
+def run_identify_output(env={}):
```
Sigmanificient: Setting a container as default argument as unexpected side effects, as they are evaluated at the function declaration.
See: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

(Note: probably fine for a test cass, but I think this avoid this pattern completely is better)

Sigmanificient (changes requested):

test/backend/x11/test_x11_identify_output.py
samunemeth: Tanks for pointing this out! I have changed back the default value to None.

samunemeth:
After that small change, I think this is ready to merge.

Sigmanificient:
@samunemeth I am not a maintainer sorry 😅

samunemeth:
Oh, I could ask you for review because you pointed out that issue I guess. Thx anyways!

samunemeth:
Seems like half of this pr that changed environment variable passing was implemented/fixed by #5887. If have removed that half of this pr.

PR #5786, opened 2026-01-14
Fix confusion/typo in DF widget format description

octogun89mm:
### Fix confusion/typo in DF widget format description:
"uf" should be "used free space", not "user free space", to prevent confusion.

### DIFF
--- a/libqtile/widget/df.py
+++ b/libqtile/widget/df.py
@@ -19,7 +19,7 @@ class DF(base.InLoopPollText):
             "format",
             "{p} ({uf}{m}|{r:.0f}%)",
             "String format (p: partition, s: size, "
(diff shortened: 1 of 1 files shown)


elParaguayo:
Isn't it just used space? "Used free" sounds like a contradiction to me.

jlacours:
> Isn't it just used space? "Used free" sounds like a contradiction to me. 

You're absolutely right, I will correct it.

This is my first pull request ever by the way 😅

elParaguayo:
I've just had a look at this widget. I think "user free space" is correct as it's looking at how much space is available for unprivileged users, which may be different to the total available space.

As such, the existing code is correct and doesn't need to be changed.

I do understand the confusion and appreciate you wanting to contribute to qtile. Please don't let this put you off submitting more pull requests if you find other issues.

PR #5787, opened 2026-01-17
wayland: Update pointer focus when view under pointer changes

### LINKED ISSUE #5775: [wayland] Weird bug with mpv, cursor needs to move 1 pixel to allow for back and forth toggling fullscreen.
### Issue description

1. Have this in `~/.config/mpv/input.conf`: `MBTN_RIGHT  cycle fullscreen `
2. Open a video file with mpv.
3. Use the right button to fullscreen toggle twice in a row without a single pixel of cursor movement. 

If I don't move the cursor at all mpv won't be reacting to right clicks. I think the bug is related to the fullscreening action itself losing the pointer somehow, mpv does not lose keyboard focus.
(tested this in sway current and it's fine there) 

### Version

0.34.2.dev35+g9bd5e8fbb

### Backend

Wayland (experimental)

### Config

```python
from libqtile import qtile

wl_xcursor_theme = "Adwaita"
wl_xcursor_size = 48

wl_input_rules = {
    "*": InputConfig(
        left_handed=False,
        pointer_accel=False,
        accel_profile="flat",
    ),

    "type:keyboard": InputConfig(
        kb_layout="hu",				
    ),
}

mod = "mod4"


keys = [
]


for i in groups:

layouts = [
    layout.MonadTall(),
    layout.Max(),
]

screens = [
    Screen(
        top=bar.Bar(
            [
                widget.GroupBox(),
                widget.Prompt(),
                widget.WindowName(),
            ],
            24,
        ),
    ),
]

import os
@hook.subscribe.startup_once
def autostart():
    qtile.cmd_spawn(os.path.expanduser("~/.config/qtile/autostart.sh")) # minimal config or not, I literally cannot use my computer without wlr-randr setting the displays.

```

### Logs

```text
-
```

### Required

richcarni:
Currently, in most circumstances, pointer focus only changes when the
pointer is moved

This commit adds checks to update pointer focus when views are
created, moved/resized and destroyed

Fixes #5775

Also cleaned up some code in xwayland's commit function while it was staring me in the face

### DIFF
--- a/libqtile/backend/wayland/qw/xwayland-view.c
+++ b/libqtile/backend/wayland/qw/xwayland-view.c
@@ -459,8 +459,22 @@ static void qw_xwayland_view_handle_commit(struct wl_listener *listener, void *d
     // This commit handler can be used for other purposes like updating geometry
     // or handling surface state changes after the view is already managed
 
-    // Update clipping if geometry changed
-    qw_xwayland_view_clip(xwayland_view);
+    struct wlr_xwayland_surface *surface = xwayland_view->xwayland_surface;
+    struct wlr_surface_state *state = &surface->surface->current;
+    struct wlr_box geom = {};
+    geom.width = state->width;
+    geom.height = state->height;
+
+    bool size_changed =
+        xwayland_view->geom.width != geom.width || xwayland_view->geom.height != geom.height;
+
+    // Update clipping if surface size changed
+    if (size_changed) {
+        qw_xwayland_view_clip(xwayland_view);
+
+        // View under the cursor may have changed
+        qw_cursor_update_pointer_focus(xwayland_view->base.server->cursor);
+    }
 
     if (xwayland_view->initial_commit) {
         // The view may have received pointer focus before it was able to update


Sinaaaa:
Seems to be working well for mpv. Though the issue persists where if I open rofi without moving the cursor at all I cannot click the menu item under the cursor until there is 1px movement. Is that a separate issue?

richcarni:
Yeah, same issue. rofi uses layer shell - in theory should be straightforward to update pointer focus there too, but for whatever reason it's refusing to cooperate...

richcarni:
Hmm... Really struggling to get the pointer to update when rofi window appears behind it. Checked sway and it has the same behaviour (pointer not updating) for rofi and gtk4-layer-demo

Anyway still trying to figure out what's going on and how to make this work

Sinaaaa:
> Hmm... Really struggling to get the pointer to update when rofi window appears behind it. Checked sway and it has the same behaviour (pointer not updating) for rofi and gtk4-layer-demo
> 
> Anyway still trying to figure out what's going on and how to make this work

Don't stress over it too much this specific rofi issue is an almost negligible problem, it's straight up nitpicky compared to the mpv one, which can be extremely annoying.

tych0:
Sounds like we should land this?

richcarni:
@Sinaaaa, can you test the latest commit with rofi?

richcarni:
> Sounds like we should land this?

Yeah, maybe worth separating into 2 PRs. The rofi fix is a bit ugly and I wouldn't mind a second opinion on whether this is the right thing to do and the right way to do it!

But starting with a more general query... does `follow_mouse_focus` only apply to keyboard focus and not pointer focus? I've assumed the answer is yes

And if I've assumed correctly, are there or should there be any exceptions?

Sinaaaa:
> @Sinaaaa, can you test the latest commit with rofi?

I tested it, this workaround indeed fixes the niche cursor needs to move 1px thing in rofi, however it also breaks standard rofi behavior if `-hover-select` is used. 
It's typical for rofi users to press a keybind to trigger rofi & then press enter to launch the top item on the list, hover select + this = surprise. 

Thinking about this now with all the information, the fact that this is the same on Sway my not even be unintentional. I'm really uncertain what is the best here. It's easy to have the best of both ways, if the user is recognizes what's going on & is not using hover select in their keybinds, but if the keybind is not rofi, but a script that manipulates rofi in some ways, then they would need to maintain 2 instances of that...

richcarni:
> It's typical for rofi users to press a keybind to trigger rofi & then press enter to launch the top item on the list, hover select + this = surprise.

Hmm, I see what you mean! We might need to investigate what the expected rofi behaviour is in these circumstances - it's quite possible I am fighting a feature rather than a bug

When I get a chance I'll split this PR so that this layer shell (rofi) stuff doesn't hold up the fix for mpv and other desktop apps

Sinaaaa:
Just noticed the "minor" issue that -hopefully- the rofi edition of this PR breaks virtual keyboards, or at least the one I'm using for couch computing. It's `wvkbd-mobintl`, the only game in town outside of big DE accessibility options. 

edit: Booted into a snapshot that has the mpv fix & not the rofi fix & the keyboard is not broken.

richcarni:
OK, walked back the rofi related changes

@Sinaaaa, I have tested but can you also re-test:
- mpv
- rofi
- wvkbd

If everything looks OK, I think we can land this

Sinaaaa:
> OK, walked back the rofi related changes
> 
> @Sinaaaa, I have tested but can you also re-test:
> 
>     * mpv
> 
>     * rofi
> 
>     * wvkbd
> 
> 
> If everything looks OK, I think we can land this

Tested, I think it's good to go.

* mpv ✅
* rofi 🟡
* wvkbd ✅

PR #5793, opened 2026-01-20
Avoid mutating screen matching parameters

### LINKED ISSUE #5792: x11 multi-monitor config without names/serials no longer working
### Issue description

I'm using a simple multi-monitor setup with something like

```python
screens = []

screen.append(Screen(…))
screen.append(Screen(…))
```

without specifying names or serials, and this was working fine on the version I built on the 13th (from master). I'm suspecting it is related to 38efe81. ~~I saw no relevant entries in the log~~ Logs added below. Here is the xrandr output:

```
Screen 0: minimum 320 x 200, current 3840 x 1080, maximum 16384 x 16384
DP-1 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 528mm x 297mm
   1920x1080     60.00*+  59.94    59.93  
   1680x1050     59.95    59.88  
   1400x1050     59.98  
   1600x900      59.95    60.00    59.82  
   1280x1024     60.02  
   1440x900      60.00  
   1400x900      59.96    59.88  
   1280x960      60.00  
   1440x810      59.97  
   1368x768      59.88    59.85  
   1280x800      59.97    59.81    59.91  
   1280x720      60.00    59.99    59.86    60.00    59.94    59.74  
   1024x768      60.04    70.07    60.00  
   960x720       60.00  
   928x696       60.05  
   896x672       60.01  
   1024x576      59.95    59.96    59.90    59.82  
   960x600       59.93    60.00  
   960x540       59.96    59.99    59.63    59.82  
   800x600       70.00    65.00    60.00    60.32    56.25  
   840x525       60.01    59.88  
   864x486       59.92    59.57  
   700x525       59.98  
   800x450       59.95    59.82  
   720x480       60.00    59.94  
   640x512       60.02  
   700x450       59.96    59.88  
   640x480       60.00    60.00    59.94  
   720x405       59.51    58.99  
   720x400       70.08  
   684x384       59.88    59.85  
   640x400       59.88    59.98  
   640x360       59.86    59.83    59.84    59.32  
   512x384       70.07    60.00  
   512x288       60.00    59.92  
   480x270       59.63    59.82  
   400x300       60.32    56.34  
   432x243       59.92    59.57  
   320x240       60.05  
   360x202       59.51    59.13  
   320x180       59.84    59.32  
DP-2 connected 1920x1080+1920+0 (normal left inverted right x axis y axis) 528mm x 297mm
   1920x1080     60.00*+  59.94    59.93  
   1680x1050     59.95    59.88  
   1400x1050     59.98  
   1600x900      59.95    60.00    59.82  
   1280x1024     60.02  
   1440x900      60.00  
   1400x900      59.96    59.88  
   1280x960      60.00  
   1440x810      59.97  
   1368x768      59.88    59.85  
   1280x800      59.97    59.81    59.91  
   1280x720      60.00    59.99    59.86    60.00    59.94    59.74  
   1024x768      60.04    70.07    60.00  
   960x720       60.00  
   928x696       60.05  
   896x672       60.01  
   1024x576      59.95    59.96    59.90    59.82  
   960x600       59.93    60.00  
   960x540       59.96    59.99    59.63    59.82  
   800x600       70.00    65.00    60.00    60.32    56.25  
   840x525       60.01    59.88  
   864x486       59.92    59.57  
   700x525       59.98  
   800x450       59.95    59.82  
   720x480       60.00    59.94  
   640x512       60.02  
   700x450       59.96    59.88  
   640x480       60.00    60.00    59.94  
   720x405       59.51    58.99  
   720x400       70.08  
   684x384       59.88    59.85  
   640x400       59.88    59.98  
   640x360       59.86    59.83    59.84    59.32  
   512x384       70.07    60.00  
   512x288       60.00    59.92  
   480x270       59.63    59.82  
   400x300       60.32    56.34  
   432x243       59.92    59.57  
   320x240       60.05  
   360x202       59.51    59.13  
   320x180       59.84    59.32  
DP-3 disconnected (normal left inverted right x axis y axis)
HDMI-1 disconnected (normal left inverted right x axis y axis)
```

Only DP-2 is managed by qtile.

### Version

0.34.1+58b3fbd.flake

### Backend

X11 (default)

### Config

```python
screens = []

screen.append(Screen(…))
screen.append(Screen(…))
```

### Logs

```text
2026-01-20 10:47:51,422 INFO libqtile manager.py:reconfigure_screens():L546 Reconfiguring screens.
2026-01-20 10:47:51,427 DEBUG libqtile manager.py:_process_screens():L450 using config serial 22806129 for output ;=D
2026-01-20 10:47:51,430 DEBUG libqtile manager.py:_process_screens():L459 using config name ;=D for output ;=D
```

### Required

richcarni:
Currently we match against Screen.name and Screen.serial but also update these with the matched outputs's name and serial

This can cause subtle issues - for example when two outputs have the same name (which shouldn't happen but can), the second output will match against the first matched screen, which will have been updated with this name

Fixes #5792

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -430,8 +430,7 @@ class Screen(CommandObject):
         y: int | None = None,
         width: int | None = None,
         height: int | None = None,
-        serial: str | None = None,
-        name: str | None = None,
+        match: dict[str, str | None] | None = None,
     ) -> None:
         self.top = top
         self.bottom = bottom
@@ -449,8 +448,9 @@ class Screen(CommandObject):
         self.width = width if width is not None else 0
         self.height = height if height is not None else 0
         self.previous_group: _Group | None = None
-        self.serial = serial
-        self.name = name
+        self.serial: str | None = None
+        self.name: str | None = None
+        self.match = match if match is not None else {"serial": None, "name": None}
 
     def __eq__(self, other: object) -> bool:
         # When we trigger a reconfigure_screens(), _process_screens()
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -429,13 +429,13 @@ class Qtile(CommandObject):
         )
 
         for i, info in enumerate(output_info):
-            scr = Screen(serial=info.serial, name=info.name)
+            scr = Screen()
             fresh_screen = True
 
             # first, try to find a screen that matches this one by serial
             # number
             for screen in config:
-                if screen.serial is not None:
+                if screen.match["serial"] is not None:
                     if not have_serials_from_hardware:
                         # if no hardware provided a serial and people provided
                         # hardware, maybe the hardware didn't have one (e.g.
@@ -444,38 +444,43 @@ class Qtile(CommandObject):
                             "serial (%s) specified in config, none found from hardware.",
                             screen.serial,
                         )
-                    if screen.serial == info.serial:
+                    if screen.match["serial"] == info.serial:
                         scr = screen
                         fresh_screen = False
                         logger.debug(
-                            f"using config serial {screen.serial} for output {info.name}"
+                            f"using config serial {screen.match['serial']} for output {info.name}"
                         )
                         break
 
-                if screen.name is not None:
-                    if screen.name == info.name:
+                if screen.match["name"] is not None:
+                    if screen.match["name"] == info.name:
                         scr = screen
                         fresh_screen = False
-                        logger.debug(f"using config name {screen.name} for output {info.name}")
+                        logger.debug(
+                            f"using config name {screen.match['name']} for output {info.name}"
+                        )
                         break
 
             # if we didn't find one by serial number, take the ith screen
             # assuming it exists, ignoring its serial number
             if fresh_screen and i < len(config):
-                if config[i].serial is not None and config[i].serial != info.serial:
+                if (
+                    config[i].match["serial"] is not None
+                    and config[i].match["serial"] != info.serial
+                ):
                     logger.warning(
                         "using config serial %s for output %s with physical serial %s",
-                        config[i].serial,
+                        config[i].match["serial"],
                         info.name,
                         info.serial,
                     )
                     # we need a copy here in case the ith window was a
                     # previously used serial number
                     scr = copy.copy(config[i])
-                elif config[i].name is not None and config[i].name != info.name:
+                elif config[i].match["name"] is not None and config[i].match["name"] != info.name:
                     logger.warning(
                         "using config name %s for output %s with physical name %s",
-                        config[i].name,
+                        config[i].match["name"],
                         info.name,
                         info.name,
                     )
@@ -484,8 +489,8 @@ class Qtile(CommandObject):
                     scr = config[i]
                     logger.debug(f"using config at index {i} for output {info.name}")
 
-                scr.serial = info.serial
-                scr.name = info.name
+            scr.serial = info.serial
+            scr.name = info.name
 
             if not hasattr(self, "current_screen") or reloading:
                 self.current_screen = scr
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -117,7 +117,7 @@ def test_screen_serial_ordering_the_order(manager_nospawn, minimal_conf_noscreen
 
 
 def make_screen(name: str | None = None, serial: str | None = None, text: str = "") -> Screen:
-    return Screen(name=name, serial=serial, top=Bar([TextBox(text)], 10))
+    return Screen(match={"name": name, "serial": serial}, top=Bar([TextBox(text)], 10))
 
 
 def test_screen_serial_ordering_one_serial(manager_nospawn, minimal_conf_noscreen, monkeypatch):


richcarni:
This PR does break config - at least in that the old screen matching syntax won't work
old:
`Screen(serial="1234")`
new:
`Screen(match={"serial": "1234"})`

Since screen matching hasn't been in a release yet, is this OK?

`match` vs `match_output` or other name?

And need to update docs

libqtile/config.py
```diff
@@ -432,6 +432,7 @@ def __init__(
         height: int | None = None,
         serial: str | None = None,
         name: str | None = None,
+        match: dict = {"serial": None, "name": None},
```
tych0: This is a bit ugly... now we have keyword parameters and a dictionary that takes additional kwargs it looks like? Which one takes precedence?

But really, it seems like we should come up with a better solution. Does it work if we only mutate when we're using the "wrong" name or serial, i.e.:

```
index 44e31e10..c3b058f8 100644
@@ -472,6 +472,7 @@ class Qtile(CommandObject):
                     # we need a copy here in case the ith window was a
                     # previously used serial number
                     scr = copy.copy(config[i])
+                    scr.serial = info.serial
                 elif config[i].name is not None and config[i].name != info.name:
                     logger.warning(
                         "using config name %s for output %s with physical name %s",
@@ -480,12 +481,11 @@ class Qtile(CommandObject):
                         info.name,
                     )
                     scr = copy.copy(config[i])
+                    scr.name = info.name
                 else:
                     scr = config[i]
                     logger.debug(f"using config at index {i} for output {info.name}")

-                scr.serial = info.serial
-                scr.name = info.name

             if not hasattr(self, "current_screen") or reloading:
                 self.current_screen = scr
```

Seems like we also probably need a test since this behavior is so complicated.

tych0:
> Since screen matching hasn't been in a release yet, is this OK?

In principle I agree that it's fine: the code is unreleased, so no migration is necessary. But I think we should come up with a better API.

libqtile/config.py
richcarni: > This is a bit ugly... now we have keyword parameters and a dictionary that takes additional kwargs it looks like? Which one takes precedence?

I should have removed name and serial as init arguments. I was trying to separate what the user would like to bind with (match dict) from info about the output that it was then bound too (name, serial). It seems unnecessarily complicated to do both in the same variables

Another way to achieve the same would be to keep the same input parameters (name, serial) and store info about the bound output in a new Output property?

>But really, it seems like we should come up with a better solution. Does it work if we only mutate when we're using the "wrong" name or serial, i.e.:

Agree - I don't have any strong opinions on this. API is important. But it would be nice if the logic was somewhat easier to reason about

I think your code might work. But I'll write a test and then we can know for sure 😀

**UPDATE:**
Code above passed the test I added but failed `test_screen_serial_ordering_the_order`:
```
>       assert manager_nospawn.c.screen[0].info()["serial"] == "a"
E       AssertionError: assert None == 'a'
```

richcarni:
Possibly adding too much complexity but (following floating layout rules) dict could be replaced by something like:

```python
Screen(output_rules=[OutputMatch(serial="1234")])
```
which adds the ability to create custom matching functions:
```python
Screen(output_rules=[OutputMatch(func=lambda o: o.x > 1920)])
Screen(output_rules=[OutputMatch(func=lambda o: o.name.startswith("DP-"))])
```

libqtile/config.py
tych0: >  Code above passed the test I added but failed test_screen_serial_ordering_the_order:

Oh, yeah, I suppose that makes sense because we make a new Screen regardless of the user's config. That has always seemed like kind of a hack to me.

But I don't understand how this really fixes the bug reported. Shouldn't we be essentially ignoring all of this logic if `serial=None` and `name=None`?

richcarni: >But I don't understand how this really fixes the bug reported. Shouldn't we be essentially ignoring all of this logic if serial=None and name=None?

This is my understanding of this bug, in so much as why we aren't ignoring all this logic when serial and name are None (please forgive the inconsistent code/pseudocode)

Say we have 2 outputs, and for some reason they have the following info:
Output_1: serial: 1234, name: @#$%
Output_2: serial: None, name: @#$%

And we have no serials, or names specified for our `Screens` in config
`screens = [Screen(id=1), Screen(id=2)]`

In `_process_screens()`:
- Try and match Output_1:
  - Matching by serial and name fails, so we fallback to binding to screens[0]. When we do this we update screens[0], now
  `screens = [Screen(id=1, serial=1234, name="@#$%"), Screen(id=2)]`

- Try and match Output_2:
  - Matching by serial fails, but matching by name works - we match screens[0] by name (since both outputs have the same name), now 
  `screens = [Screen(id=1, serial=1234, name="@#$%"), Screen(id=1, serial=1234, name="@#$%")]]`
  
So both outputs have matched/bound screens[0] !

Arguably the underlying issue is that we haven't parsed the output name correctly

tych0: > Matching by serial fails, but matching by name works - we match screens[0] by name (since both outputs have the same name), now

Oh, I see. That is very unfortunate... but not unlikely. I own two of the same monitors, which both have the same name. It really feels like we shouldn't create a separate Screen() object each time in _process_screens(). We should just use what the user gave us.

(I really don't like the dict-as-arg API: that makes it harder to add type checking for users.)

> Arguably the underlying issue is that we haven't parsed the output name correctly

Yeah. I wonder if we fix that instead? I commented on the original issue.

richcarni: > Oh, I see. That is very unfortunate... but not unlikely. I own two of the same monitors, which both have the same name.

Does name mean something different between wayland/x11? On wayland names are the port identifier: "HDMI-A-1", "DP-1", etc. So I think they have to be unique

> It really feels like we shouldn't create a separate Screen() object each time in _process_screens(). We should just use what the user gave us.

Sounds reasonable. I would probably go further and say we also shouldn't bind to a Screen with name/serial that doesn't match an output. Maybe a two pass process: first match by any names/serials specified, then bind any remaining outputs to any remaining Screens without names/serials

> (I really don't like the dict-as-arg API: that makes it harder to add type checking for users.)

Fair! Not sure if you saw the alternative I posted below. Anyway, besides fixing the related issue, this isn't something I was really planning to progress

```python
Screen(output_rules=[OutputMatch(serial="1234")])

```

tych0: > Does name mean something different between wayland/x11? On wayland names are the port identifier: "HDMI-A-1", "DP-1", etc. So I think they have to be unique

On x11 it's the name of the monitor, like, Dell XYZ, or whatever. We should probably make this the same :). I don't have a strong opinion on which it is.

I posed a fix for the EDID name parsing in #5806, but it looks like we probably need to sync on DP-1 vs. Dell XYZ. Maybe `.port == DP-1` and `.model == Dell XYZ` to avoid confusion? "name" is kind of generic.

>  Maybe a two pass process: first match by any names/serials specified, then bind any remaining outputs to any remaining Screens without names/serials

Yeah, I thought about that. But then if you plug into a new display that's not your usual one, you have to change your config instead of just getting the screen config you usually get for free.

richcarni: >On x11 it's the name of the monitor, like, Dell XYZ, or whatever. We should probably make this the same :). I don't have a strong opinion on which it is.
>
>I posed a fix for the EDID name parsing in https://github.com/qtile/qtile/pull/5806, but it looks like we probably need to sync on DP-1 vs. Dell XYZ. Maybe .port == DP-1 and .model == Dell XYZ to avoid confusion? "name" is kind of generic.

On wayland, for example my laptop: wlr_output.name="eDP-1", .model="Color LCD"
But no objections to calling eDP-1 a port. So map wlr-output.name -> screen_info.port and replace name binding with port binding (or in addition to model binding, but on wayland I think port binding is what most people will use)

>Yeah, I thought about that. But then if you plug into a new display that's not your usual one, you have to change your config instead of just getting the screen config you usually get for free

That behaviour makes sense. And it does throw a warning in case it wasn't intentional (typo in serial or something). Although, I suspect it might not behave as expected if screens are re-ordered - we could add a test for that, but not sure it is a major issue

richcarni: I think I'll close this PR since it looks like we'll be going in a different direction

tych0:
So I think https://github.com/qtile/qtile/commit/93aa4ec997e725d5684fdb6b3e33ca87a9edbc7d will fix the mismatch between the wayland semantics and the x11 semantics, and expose a few more things.

I left `Screen.name` as is, we need to figure out what the API is there. Perhaps we should do OutputMatch as you suggested @richcarni; I'm also wondering if we can do some more general thing where, at config time, we export a `libqtile.system_state` or similar. We could have `libqtile.system_state.outputs`, and then people can look at that and decide what to export as their `screens`.

But then what happens when people attach a new screen? So then I was thinking about letting users define `process_screens()` in their configs, and we would call that to match things up?

I don't know, maybe `OutputMatch` is easiest. But if you could write actual python code yourself, that feels easier than reading our documentation and trying to understand our matching semantics.

tych0:
Anyway, let me know what you think of that patch. I can send it as a PR to fix the immediate issues if it looks reasonable.

richcarni:
> Anyway, let me know what you think of that patch. I can send it as a PR to fix the immediate issues if it looks reasonable.

Looks good!

>I left Screen.name as is, we need to figure out what the API is there. Perhaps we should do OutputMatch as you suggested @richcarni; I'm also wondering if we can do some more general thing where, at config time, we export a libqtile.system_state or similar. We could have libqtile.system_state.outputs, and then people can look at that and decide what to export as their screens.
>
>But then what happens when people attach a new screen? So then I was thinking about letting users define process_screens() in their configs, and we would call that to match things up?
>
>I don't know, maybe OutputMatch is easiest. But if you could write actual python code yourself, that feels easier than reading our documentation and trying to understand our matching semantics.

Could an OutputMatch rule take a function that uses the exposed system_state to apply different matching logic depending on how many or what specific outputs are available (even when that changes)? 

If that's doable, I feel like this should be powerful enough without having to go all the way and write your own process_screens. And even if we want custom process_screens, I think OutputMatch + process_screens would be better than basic match + process_screens. My two cents anyway

tych0:
> Could an OutputMatch rule take a function that uses the exposed system_state to apply different matching logic depending on how many or what specific outputs are available (even when that changes)?

I don't understand how it would work with a function, since the OutputMatch list is supplied as an argument to the Screen? I was thinking of a new top level thing `generate_screens(outputs: list[Output]) -> list[Screen]`, that users would provide instead of an explicit list of Screen objects.

I'll play around with it a bit. I wasn't thinking people would have to write their own equivalent of _process_screens(), but more like, something that takes an argument and generates a list of Screens, which we would then feed to the existing process screens.

richcarni:
> I don't understand how it would work with a function, since the OutputMatch list is supplied as an argument to the Screen?

Yeah, I'm not sure either. Probably didn't think it through thoroughly

> I wasn't thinking people would have to write their own equivalent of _process_screens(), but more like, something that takes an argument and generates a list of Screens, which we would then feed to the existing process screens.

Sounds like that would offer the most flexibility. I guess I'm struggling to think of concrete examples where one might require such sophisticated screen matching capabilities... probably because I don't have a complicated multi-screen setup myself!

PR #5795, opened 2026-01-22
wayland: Fix blurry icons on hidpi displays [redo]

richcarni:
The technique is:
- Rasterize images at their final physical size (upscale if required)
- Downscale images to their logical size for compositing
- When the entire window is finally upscaled for display, since these
  last two scaling operations are on a recording surface they
  effectively cancel out
  
Attempting to implement this in Img class, rather than Drawer (#5685)
e.g.
```python
img = Img.from_path(icon_path)
img.resize(height=self.icon_size)
self.drawer.draw_image(img, offsetx=x, offsety=y)
```

So this looks like it will require using Img's instead of cairo surfaces and pattern as the primary data type before rendering

Widgets updated:
- [x] TaskList
- [x] Statusnotifer
- [x] Image
- [x] Volume
- [x] BatteryIcon

Additional todos:
- [x] tests

### DIFF
--- a/libqtile/images.py
+++ b/libqtile/images.py
@@ -1,11 +1,15 @@
+from __future__ import annotations
+
 import os
 from collections import namedtuple
+from copy import copy
 from math import pi
 
 import cairocffi
 import cairocffi.pixbuf
 
-from libqtile.utils import scan_files
+from libqtile import utils
+from libqtile.utils import ColorsType, scan_files
 
 
 class LoadingError(Exception):
@@ -29,6 +33,11 @@ def get_cairo_surface(bytes_img, width=None, height=None):
         )
         surf, fmt = cairocffi.pixbuf.decode_to_image_surface(bytes_img)
         return _SurfaceInfo(surf, fmt)
+
+
+def get_cairo_surface_for_data(data, format, width, height):
+    surf = cairocffi.ImageSurface.create_for_data(data, format, width, height)
+    return _SurfaceInfo(surf, format)
 
 
 def get_cairo_pattern(surface, width=None, height=None, theta=0.0):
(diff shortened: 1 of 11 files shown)


richcarni:
@elParaguayo I'd like to finish off this aging PR (start clearing my PR backlog!) If this looks OK I'll go ahead and update the other widgets

Currently works like (or have a look at the tasklist widget):

```python
img = Img.from_path(icon_path)
img.resize(height=self.icon_size, output_scale=self.drawer.output_scale)
img.draw(drawer=self.drawer, offsetx=x, offsety=y)
```

Personally, I think I prefer the following for that final line, but it's a quibble
```python
self.drawer.draw(img, offsetx=x, offsety=y)
```

elParaguayo:
Sorry for the delay.

I agree that that the latter code is neater but we can't use `self.drawer.draw` as that's the main function for rendering the surface but something like `self.drawer.draw_img` would still be an improvement.

Do you need anything else from me on this?

richcarni:
> Do you need anything else from me on this?

Don't think so! Thanks! It's a pain going through and updating all the widgets. Was just looking for (even minimal) reassurance this is going in the right direction

elParaguayo:
This is looking pretty good to me. I may have to update qtile-extras once this is merged but I don't think this breaks anything.

richcarni:
> This is looking pretty good to me. I may have to update qtile-extras once this is merged but I don't think this breaks anything.

Cool. I'll run it locally for a while and I think some tests would be worthwhile

I haven't tested if it actually breaks qtile-extras yet but I think only statusnotifier should be affected? I could even make that PR. I just added a color option to draw_image, specifically for qtile-extras' statusnotifier

Oh, i just saw you already have a separate ImgMask class. Oh well, we'll figure it out!

richcarni:
Can confirm this at least breaks qtile-extras statusnotifier. Have created a complementary PR over there

Not sure if anything else will be broken

libqtile/images.py
```diff
@@ -220,12 +301,51 @@ def _scale_free(width_factor, height_factor, initial_size):
...
+            ctx.set_source(self.pattern)
+            ctx.paint()
+            ctx.translate(offsetx, offsety)
+            ctx.set_source(overlay.pattern)
+            ctx.paint()
+        data = bytearray(surface.get_data())
+        fmt = surface.get_format()
+        return Img.from_data(data, fmt, self.width, self.height)
+
+    def paint_mask(self, colour: ColorsType):
```
elParaguayo:
```suggestion
    def paint_mask(self, colour: ColorsType) -> Img:
```

richcarni: I thought it would be good to add type annotation for new functions in images.py, but they don't seem to be checked, so I'm not sure if there is any point? 

I don't really understand why they aren't checked. My best guess is that the type stubs for cariocffi are full of Any and somehow this impacts everything that touch these? TBH I'm not all that familiar with how type annotations work in python

elParaguayo: Regardless of actual checking, they're helpful for people browsing the source code.

libqtile/images.py
```diff
@@ -313,39 +313,43 @@ def paste(self, overlay: Img, offsetx: int = 0, offsety: int = 0) -> Img:
...
-            ctx.fill()
-        data = bytearray(surface.get_data())
-        fmt = surface.get_format()
-        return Img.from_data(data, fmt, self.width, self.height)
+                ctx.set_operator(cairocffi.OPERATOR_SOURCE)
+                ctx.mask(cairocffi.SurfacePattern(surface))
+                ctx.fill()
+
+        self._operations.append(func)
+        return self
```
elParaguayo, line 360: Why return `self` here? I'm not sure you need to return anything.

test/backend/wayland/test_drawer.py
```diff
@@ -82,3 +82,18 @@ def test_hidpi_pixel_data_scaling(rgba_pixel_data, drawer):  # noqa:F811
...
+
+def test_hidpi_img_paint_mask(svg_img, drawer):
+    assert svg_img.width == 24
+    assert svg_img.height == 24
+    # Snapshot svg_img before any scaling
+    img0 = copy(svg_img)
+    red_img0 = img0.paint_mask("#ff0000")
+    # Set icon image size
+    svg_img.resize(height=16)
+    red_img = svg_img.paint_mask("#ff0000")
```
elParaguayo: `red_img` isn't a new object. `red_img` and `svg_img` are the same because `paint_mask` returns `self`.

elParaguayo (commented):
I like the idea of adding operations to an image but think there are a couple of very small issues with your current code.

libqtile/images.py
richcarni: The thinking was that you could potentially chain operations, like:
```
img.paste(img2).paint_mask("#FF0000")
```

Obviously not all that useful when there are only a couple of operations! (I still need to update paste)

elParaguayo: Ah. Ok. Nice.

test/backend/wayland/test_drawer.py
richcarni: Ah, that was a leftover from before I updated paint_mask. I think it's still valid though. But should really just be:
```python
svg_img.resize(height=16)
svg_img.paint_mask("#ff0000")
d, image_surface = drawer
d.draw_image(svn_img)
```

tpaniaki:
Thanks and sorry for the delay. 

Currently, icons are properly scaled but 1) as mentioned no icons from StatusNotifier and 2) the bar loses its background:

swappy-20260515_202112

richcarni:
> Currently, icons are properly scaled but 1) as mentioned no icons from StatusNotifier and 2) the bar loses its background:

Thanks for picking up on that! I accidentally removed the line that draws the background

tpaniaki:
Hello, my icons in StatusNotifier still won't display.

elParaguayo:
If you're using qtile-extras, that version of the widget hasn't been updated yet.

PR #5804, opened 2026-01-25
tests: assume gtk is present

tych0:
The import of GtkLayerShell triggers GTK's initialization, which creates 4 threads, which can deadlock my local pytest when doing a run of all the tests, with the cryptic output below.

Instead, don't test this at all, just assume that gtk is present. Since the actual gtk import is done via the test/scripts/window.py in a subprocess, that will not mess with pytest's forking. And if gtk is not present, the script will fail with a reasonable error. It is listed as dev dependency in pyproject.toml, so it will probably always work assuming people have installed the dev dependencies.

    =================================== ERRORS ====================================
    ________ ERROR at setup of test_inhibitor_open[1-x11-InhibitorConfig] _________

    fixturedef = <FixtureDef argname='manager' scope='function' baseid='test'>
    request = <SubRequest 'manager' for <Function test_inhibitor_open[1-x11-InhibitorConfig]>>

        @pytest.hookimpl(wrapper=True)
        def pytest_fixture_setup(fixturedef: FixtureDef, request) -> object | None:
            asyncio_mode = _get_asyncio_mode(request.config)
            if not _is_asyncio_fixture_function(fixturedef.func):
                if asyncio_mode == Mode.STRICT:
                    # Ignore async fixtures without explicit asyncio mark in strict mode
                    # This applies to pytest_trio fixtures, for example
    >               return (yield)
                            ^^^^^

    .venv/lib/python3.13/site-packages/pytest_asyncio/plugin.py:728:
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
    test/conftest.py:93: in manager
        manager_nospawn.start(config)
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    self = <test.helpers.TestManager object at 0x7fd093565160>
    config_class = <class 'test_idle_inhibit.InhibitorConfig'>, no_spawn = False
    state = None

        def start(self, config_class, no_spawn=False, state=None):
            multiprocessing.set_start_method("fork", force=True)
            readlogs, writelogs = os.pipe()
            rpipe, wpipe = multiprocessing.Pipe()

            def run_qtile():
                try:
                    rpipe.close()
                    os.environ.pop("DISPLAY", None)
[...]

PR #5810, opened 2026-01-27
wayland: implement wp-cursor-shape-v1 protocol

Gurjaka:
Add support for the staging cursor-shape-v1 protocol. This allows clients to request standard cursor shapes by enum (e.g., pointer, text, crosshair) rather than providing their own pixel buffers.

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -432,6 +432,18 @@ class Core(base.Core):
         else:
             return self.qtile.process_button_release(button, mask)
 
+    @expose_command
+    def get_cursor_shape_v1(self) -> str | None:
+        """
+        Get the current cursor shape name from cursor-shape-v1 protocol.
+
+        Returns None if the cursor has no shape name set.
+        """
+        cursor = self.qw_cursor
+        if cursor.current_shape_name == ffi.NULL:
+            return None
+        return ffi.string(cursor.current_shape_name).decode("utf-8")
+
     def handle_manage_view(self, view: ffi.CData) -> None:
         wid = self.new_wid()
         view.wid = wid
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -375,6 +375,7 @@ struct qw_cursor *qw_server_cursor_create(struct qw_server *server) {
     cursor->cursor = wlr_cursor_create();
     wlr_cursor_attach_output_layout(cursor->cursor, server->output_layout);
     cursor->mgr = wlr_xcursor_manager_create(NULL, 24);
+    cursor->current_shape_name = NULL;
 
     // Setup listeners for various pointer events
     cursor->request_set.notify = qw_cursor_handle_seat_request_set;
@@ -686,6 +687,7 @@ static void qw_handle_request_set_cursor_shape(struct wl_listener *listener, voi
     }
 
     const char *name = wlr_cursor_shape_v1_name(event->shape);
+    cursor->current_shape_name = name;
 
     wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, name);
 }
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -360,6 +360,9 @@ static void qw_cursor_handle_frame(struct wl_listener *listener, void *data) {
     wlr_seat_pointer_notify_frame(cursor->server->seat);
 }
 
+// forward declaration
+static void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data);
+
 struct qw_cursor *qw_server_cursor_create(struct qw_server *server) {
     // Allocate memory for qw_cursor
     struct qw_cursor *cursor = calloc(1, sizeof(*cursor));
@@ -673,7 +676,7 @@ void qw_cursor_fake_click(struct qw_cursor *cursor) {
     qw_cursor_handle_button(&cursor->button, &event);
 }
 
-void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
+static void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
     struct qw_cursor *cursor = wl_container_of(listener, cursor, request_set_cursor_shape);
     struct wlr_cursor_shape_manager_v1_request_set_shape_event *event = data;
 
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -15,6 +15,7 @@ void qw_cursor_destroy(struct qw_cursor *cursor) {
     wl_list_remove(&cursor->motion_absolute.link);
     wl_list_remove(&cursor->frame.link);
     wl_list_remove(&cursor->button.link);
+    wl_list_remove(&cursor->request_set_cursor_shape.link);
 
     wlr_xcursor_manager_destroy(cursor->mgr);
 
@@ -391,6 +392,11 @@ struct qw_cursor *qw_server_cursor_create(struct qw_server *server) {
     cursor->button.notify = qw_cursor_handle_button;
     wl_signal_add(&cursor->cursor->events.button, &cursor->button);
 
+    cursor->cursor_shape_mgr = wlr_cursor_shape_manager_v1_create(server->display, 1);
+    cursor->request_set_cursor_shape.notify = qw_handle_request_set_cursor_shape;
+    wl_signal_add(&cursor->cursor_shape_mgr->events.request_set_shape,
+                  &cursor->request_set_cursor_shape);
+
     wl_list_init(&cursor->constraint_commit.link);
 
     return cursor;
@@ -665,4 +671,18 @@ void qw_cursor_fake_click(struct qw_cursor *cursor) {
     // Simulate release
     event.state = WL_POINTER_BUTTON_STATE_RELEASED;
     qw_cursor_handle_button(&cursor->button, &event);
+}
+
+void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
+    struct qw_cursor *cursor = wl_container_of(listener, cursor, request_set_cursor_shape);
+    struct wlr_cursor_shape_manager_v1_request_set_shape_event *event = data;
+
+    struct wlr_seat_client *focused_client = cursor->server->seat->pointer_state.focused_client;
+    if (focused_client != event->seat_client) {
+        return;
+    }
+
+    const char *name = wlr_cursor_shape_v1_name(event->shape);
+
+    wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, name);
 }
--- a/libqtile/backend/wayland/qw/cursor.h
+++ b/libqtile/backend/wayland/qw/cursor.h
@@ -2,6 +2,7 @@
 #define CURSOR_H
 
 #include <wlr/types/wlr_cursor.h>
+#include <wlr/types/wlr_cursor_shape_v1.h>
 #include <wlr/types/wlr_pointer_constraints_v1.h>
 #include <wlr/types/wlr_xcursor_manager.h>
 
@@ -28,8 +29,10 @@ struct qw_cursor {
     struct wl_listener frame;
     struct wl_listener button;
     struct wl_listener constraint_commit;
+    struct wl_listener request_set_cursor_shape;
     struct wlr_xcursor_manager *mgr;
     struct wlr_xcursor_manager *xwayland_mgr;
+    struct wlr_cursor_shape_manager_v1 *cursor_shape_mgr;
     struct wlr_surface *saved_surface;
     uint32_t saved_hotspot_x;
     uint32_t saved_hotspot_y;
@@ -67,6 +70,7 @@ void qw_cursor_pointer_constraint_new(struct qw_cursor *cursor,
                                       struct wlr_pointer_constraint_v1 *constraint);
 
 void qw_cursor_configure_xcursor(struct qw_cursor *cursor);
+void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data);
 
 void qw_cursor_constrain_cursor(struct qw_cursor *cursor,
                                 struct wlr_pointer_constraint_v1 *constraint);
--- a/libqtile/backend/wayland/qw/cursor.h
+++ b/libqtile/backend/wayland/qw/cursor.h
@@ -19,6 +19,7 @@ struct qw_cursor {
     struct wlr_cursor *cursor;
     struct qw_view *view;
     struct qw_implicit_grab implicit_grab;
+    struct wlr_cursor_shape_manager_v1 *cursor_shape_mgr;
 
     // private data
     struct qw_server *server;
@@ -32,7 +33,6 @@ struct qw_cursor {
     struct wl_listener request_set_cursor_shape;
     struct wlr_xcursor_manager *mgr;
     struct wlr_xcursor_manager *xwayland_mgr;
-    struct wlr_cursor_shape_manager_v1 *cursor_shape_mgr;
     struct wlr_surface *saved_surface;
     uint32_t saved_hotspot_x;
     uint32_t saved_hotspot_y;
@@ -70,7 +70,6 @@ void qw_cursor_pointer_constraint_new(struct qw_cursor *cursor,
                                       struct wlr_pointer_constraint_v1 *constraint);
 
 void qw_cursor_configure_xcursor(struct qw_cursor *cursor);
-void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data);
 
 void qw_cursor_constrain_cursor(struct qw_cursor *cursor,
                                 struct wlr_pointer_constraint_v1 *constraint);
--- a/libqtile/backend/wayland/qw/cursor.h
+++ b/libqtile/backend/wayland/qw/cursor.h
@@ -20,6 +20,7 @@ struct qw_cursor {
     struct qw_view *view;
     struct qw_implicit_grab implicit_grab;
     struct wlr_cursor_shape_manager_v1 *cursor_shape_mgr;
+    const char *current_shape_name;
 
     // private data
     struct qw_server *server;
--- a/test/backend/wayland/test_cursor.py
+++ b/test/backend/wayland/test_cursor.py
@@ -1,0 +1,33 @@
+from pathlib import Path
+
+import pytest
+
+pytest.importorskip("libqtile.backend.wayland.core")
+
+BIN_PATH = Path(__file__).parent.parent / "test/wayland_clients/bin/cursor-shape-v1"
+
+
+@pytest.mark.parametrize("shape", ["text", "crosshair", "wait", "help"])
+def test_cursor_shape_protocol(wmanager, shape):
+    """Test that the C client can successfully request shapes via the protocol."""
+
+    try:
+        wmanager.test_window("cursor-shape-v1")
+
+        wmanager.c.core.warp_pointer(150, 150)
+
+        cursor_name = wmanager.c.eval("""
+from libqtile.backend.wayland._ffi import ffi
+cursor = self.core.qw_cursor
+ffi.string(cursor.current_shape_name).decode('utf-8') if cursor.current_shape_name != ffi.NULL else None
+        """)
+
+        if shape == "wait":
+            assert cursor_name in ["wait", "watch"]
+        elif shape == "help":
+            assert cursor_name in ["help", "whats_this", "question_arrow"]
+        else:
+            assert cursor_name == shape
+
+    except Exception:
+        return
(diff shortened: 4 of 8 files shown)


tych0:
Can you add a test for this?

Gurjaka:
> Can you add a test for this?

Have never touched tests before, so I really got no idea how to write one ;-; especially for this kind of protocol tbh...

Perhaps I could get some guidance?

tych0:
Seems like it would need to spawn an application that uses this protocol to set the cursor, and then confirm that the cursor is set. It might be useful to add a core API for testing purposes that exposes the cursor or similar.

Gurjaka:
> Seems like it would need to spawn an application that uses this protocol to set the cursor, and then confirm that the cursor is set. It might be useful to add a core API for testing purposes that exposes the cursor or similar.

Oh I think I know what to do! Just running low on time now...

Gurjaka:
I can't seem to run qtile tests locally for some reason, tried playing around a bit, but still no luck.

Maybe that should be fixed first?

elParaguayo:
What error are you getting?

Gurjaka:
> What error are you getting?

```bash
gurami on desktop qtile on  master [⇡15] via 🐍 python v3.13.9 via  impure (nix-shell-env) took 0s 
➜ qtile-run-tests-wayland
================================================================================================================== test session starts ==================================================================================================================
platform linux -- Python 3.13.9, pytest-8.4.2, pluggy-1.6.0
rootdir: /home/gurami/Dev/qtile
configfile: pyproject.toml
testpaths: test
plugins: asyncio-1.2.0, anyio-4.11.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 1372 items                                                                                                                                                                                                                                    

test/backend/test_backend.py .                                                                                                                                                                                                                    [  0%]
test/backend/test_idle_inhibit.py .E

======================================================================================================================== ERRORS =========================================================================================================================
___________________________________________________________________________________________ ERROR at setup of test_inhibitor_open[1-wayland-InhibitorConfig] ____________________________________________________________________________________________

fixturedef = <FixtureDef argname='manager' scope='function' baseid='test'>, request = <SubRequest 'manager' for <Function test_inhibitor_open[1-wayland-InhibitorConfig]>>

    @pytest.hookimpl(wrapper=True)
    def pytest_fixture_setup(fixturedef: FixtureDef, request) -> object | None:
        asyncio_mode = _get_asyncio_mode(request.config)
        if not _is_asyncio_fixture_function(fixturedef.func):
            if asyncio_mode == Mode.STRICT:
                # Ignore async fixtures without explicit asyncio mark in strict mode
                # This applies to pytest_trio fixtures, for example
>               return (yield)
                        ^^^^^

/nix/store/fm3575308n2s8n6vzvavx8sh8cp97pcb-python3.13-pytest-asyncio-1.2.0/lib/python3.13/site-packages/pytest_asyncio/plugin.py:733: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test/conftest.py:93: in manager
    manager_nospawn.start(config)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test.helpers.TestManager object at 0x7f4e151af770>, config_class = <class 'test_idle_inhibit.InhibitorConfig'>, no_spawn = False, state = None

    def start(self, config_class, no_spawn=False, state=None):
        multiprocessing.set_start_method("fork", force=True)
        readlogs, writelogs = os.pipe()
        rpipe, wpipe = multiprocessing.Pipe()
    
        def run_qtile():
            try:
                rpipe.close()
                os.environ.pop("DISPLAY", None)
                os.environ.pop("WAYLAND_DISPLAY", None)
                init_log(self.log_level)
                kore = self.backend.create()
                os.environ.update(self.backend.env)
                from libqtile.core.lifecycle import lifecycle
    
                os.close(readlogs)
                formatter = logging.Formatter("%(levelname)s - %(message)s")
                handler = logging.StreamHandler(os.fdopen(writelogs, "w"))
                handler.setFormatter(formatter)
                logger.addHandler(handler)
    
                Qtile(
                    kore,
                    config_class(),
                    socket_path=self.sockfile,
                    no_spawn=no_spawn,
                    state=state,
                ).loop()
                lifecycle._atexit()
            except Exception:
                wpipe.send(traceback.format_exc())
            finally:
                wpipe.close()
    
        self.proc = multiprocessing.Process(target=run_qtile)
        self.proc.start()
        wpipe.close()
        os.close(writelogs)
        self.logspipe = readlogs
    
        # First, wait for socket to appear
        try:
            if can_connect_qtile(self.sockfile, ok=lambda: not rpipe.poll()):
                ipc_client = ipc.Client(self.sockfile)
                ipc_command = command.interface.IPCCommandInterface(ipc_client)
                self.c = command.client.InteractiveCommandClient(ipc_command)
                self.backend.configure(self)
                return
            if rpipe.poll(0.1):
                error = rpipe.recv()
                raise AssertionError(f"Error launching qtile, traceback:\n{error}")
>           raise AssertionError("Error launching qtile")
E           AssertionError: Error launching qtile

test/helpers.py:234: AssertionError
----------------------------------------------------------------------------------------------------------------- Captured stdout setup -----------------------------------------------------------------------------------------------------------------
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [backend/backend.c:342] Loading user-specified backends due to WLR_BACKENDS: headless
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [backend/headless/backend.c:60] Creating headless backend
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [util/env.c:9] Loading WLR_NO_HARDWARE_CURSORS option: 1
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [util/env.c:25] Loading WLR_RENDERER option: pixman
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [render/pixman/renderer.c:328] Creating pixman renderer
2026-02-03 16:26:55,342 libqtile core.py:log_cb():L106  [backend/headless/backend.c:17] Starting headless backend
2026-02-03 16:26:55,343 libqtile core.py:log_cb():L106  [server.c:188] Updating: 0,0 (800x600)
2026-02-03 16:26:55,344 libqtile core.py:log_cb():L106  [server.c:116] Running Wayland compositor on WAYLAND_DISPLAY=wayland-0
2026-02-03 16:26:55,380 libqtile prompt.py:_configure():L412  Prompt widget only supports audible bell under X11
2026-02-03 16:26:55,381 libqtile bar.py:_configure_widget():L344  Widget removed: Systray does not support wayland.
=================================================================================================================== warnings summary ====================================================================================================================
test/backend/test_idle_inhibit.py::test_inhibitor_open[1-wayland-InhibitorConfig]
  /nix/store/3lll9y925zz9393sa59h653xik66srjb-python3-3.13.9/lib/python3.13/multiprocessing/popen_fork.py:67: DeprecationWarning: This process (pid=1133230) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

================================================================================================================ short test summary info ================================================================================================================
ERROR test/backend/test_idle_inhibit.py::test_inhibitor_open[1-wayland-InhibitorConfig] - AssertionError: Error launching qtile
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================================================================= 2 passed, 1 warning, 1 error in 8.42s =========================================================================================================

```

elParaguayo:
Same error as me.

Are you on latest git version? I had hoped https://github.com/qtile/qtile/pull/5804 would fix it! Note: I've not had time to check whether that fix works for me.

tych0:
I believe xwayland forks as well, so there are probably more sources of deadlocking than in the X11 world. I think we need to get rid of threading entirely in our test setup code. I've been working on it, but it's not totally done.

Gurjaka:
> Same error as me.
> 
> Are you on latest git version? I had hoped #5804 would fix it! Note: I've not had time to check whether that fix works for me.

I am running latest flake version on my system, but the tests are running from local fork of Qtile (rebased from master)

richcarni:
> Seems like it would need to spawn an application that uses this protocol to set the cursor, and then confirm that the cursor is set. It might be useful to add a core API for testing purposes that exposes the cursor or similar.

I've been thinking for a while (every time there is some sort of regression! Looking at you pointer constraints!!) that the wayland backend would benefit from more tests. TBH, I don't think there are really any tests of wayland specific features

I guess we need to start somewhere, and ideally start adding tests for existing functionality too

But I'm not sure we want to add a bunch of dependencies on external applications to use for testing? So that would mean writing our own minimal testing apps - which I think is a good idea, but a lot of work!

tych0:
> But I'm not sure we want to add a bunch of dependencies on external applications to use for testing? 

As long as they're available relatively simply for CI (apt-get install or similar), I don't see the problem. We already depend on a bunch of things, xterm, xeyes, xephyr, etc.

Gurjaka:
> > But I'm not sure we want to add a bunch of dependencies on external applications to use for testing?
> 
> As long as they're available relatively simply for CI (apt-get install or similar), I don't see the problem. We already depend on a bunch of things, xterm, xeyes, xephyr, etc.

Nix could also easily be used to manage dependencies in CI if other package managers get complicated

Gurjaka:
Can this be merged? As tests are broken, I really got no idea how to write a new one for this + It's really annoying to have default cursor in gtk4 applications ;-;

tych0:
The tests are not broken, they just have some bugs. You can try to run your individual test file as a workaround.

I would really appreciate a test for this code.

Gurjaka:
> The tests are not broken, they just have some bugs. You can try to run your individual test file as a workaround.
> 
> I would really appreciate a test for this code.

Yeah I get it, but couldn't that be different PR? I'll try my best during weekend to write something.

tych0:
My experience with different PRs that add tests is that they never materialize :(

Gurjaka:
> My experience with different PRs that add tests is that they never materialize :(

Alright I've added test for the cursor shape protocol in `test/backend/wayland/test_cursor.py`. The tests verify that cursor shapes can be set using the wp-cursor-shape-v1 protocol.

Tests pass with `pytest --backend=wayland test/backend/wayland/test_cursor.py`

Gurjaka:
Umm, I don't think CI error is related to this PR is it?

Gurjaka:
Ready for review ig

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -644,3 +653,36 @@ void qw_cursor_configure_xcursor(struct qw_cursor *cursor) {
...
+    if (focused_client != event->seat_client) {
+        return;
+    }
+
+    const char *name = wlr_cursor_shape_v1_name(event->shape);
+
+    wlr_cursor_set_xcursor(cursor->cursor, cursor->mgr, name);
+}
+
+void qw_cursor_set_shape(struct qw_cursor *cursor, uint32_t shape) {
```
richcarni: This function is just used for testing?

richcarni: I think shape should be have type: enum wp_cursor_shape_device_v1_shape

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -644,3 +653,36 @@ void qw_cursor_configure_xcursor(struct qw_cursor *cursor) {
...
+    if (!cursor)
+        return;
+
+    const char *name = wlr_cursor_shape_v1_name(shape);
+
+    free(cursor->current_shape_name);
+    cursor->current_shape_name = name ? strdup(name) : NULL;
```
richcarni: No need to copy string and free memory here. Can either just store the pointer `cursor->current_shape_name = name`, or consider just storing the enum value instead and getting the name when needed

libqtile/backend/wayland/qw/cursor.c
```diff
@@ -644,3 +653,36 @@ void qw_cursor_configure_xcursor(struct qw_cursor *cursor) {
         wlr_cursor_warp(cursor->cursor, NULL, cursor->cursor->x, cursor->cursor->y);
     }
 }
+
+void qw_handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
```
richcarni: Can this function be static?

libqtile/backend/wayland/core.py
```diff
@@ -282,6 +282,7 @@ def __init__(self) -> None:
         lib.qw_server_start(self.qw)
         os.environ["WAYLAND_DISPLAY"] = self.display_name
         self.qw_cursor = lib.qw_server_get_cursor(self.qw)
+        self.cursor_shape_manager = self.qw_cursor.cursor_shape_mgr
```
richcarni: Is this used anywhere?

richcarni:
Thanks for working on this @Gurjaka!

I'm not sure these tests add much though. I wonder how hard it would be to write a tiny gtk app that does nothing except set the cursor shape using this protocol?

We could run this and then the test could verify that the cursor was indeed set

I'm thinking going forward, we could create a whole library of these little tests for testing specific wayland functionality?

elParaguayo:
I would think we want to create a (single) client to test support for most extensions.

libqtile/backend/wayland/qw/cursor.c
Gurjaka: of course, looks like I forgot about it 😄

libqtile/backend/wayland/qw/cursor.c
Gurjaka: yes, though might not be necessary if we consider another way of testing

Gurjaka: correct

libqtile/backend/wayland/qw/cursor.c
Gurjaka: Removed the copying logic, and converted `current_shape_name` to const pointer

libqtile/backend/wayland/core.py
Gurjaka: looks like it's leftover from playing around with py code

Gurjaka:
@richcarni I resolved your comments and force pushed, not really sure what I should do about tests though

richcarni:
I was thinking for the test(s) we have a small gtk app (client) that uses this protocol to request a certain cursor shape, then we would need a function (presumably from python) to get the current pointer shape and assert that it is the same

weston has a bunch of small clients like this and wlroots used to (still does?) have example clients for various protocols - something like this. AI could probably produce something like this?

I was thinking multiple simple clients for the various protocols, but it could be combined in a single client - I'm not sure how much it matters

richcarni:
I wrote a small test client for cursor-shape-v1: https://github.com/richcarni/qtile/commit/f4298dcc70e2bb79ad89ba2b50143350410fd70a

@Gurjaka you should be able to cherry-pick it into your branch

No additional dependencies outside of wayland client protocols

I'm not sure if this is the way to go - whether this could be expanded into something that supports multiple protocols or if it's better to have a client per protocol (or something in between)? Just hoping to help keep this moving :)

And for this PR, still needs some way of checking that the clients request has been handled correctly

[Disclaimer: some AI assistance used but mostly based on https://github.com/emersion/hello-wayland/blob/master/main.c]

Gurjaka:
> I wrote a small test client for cursor-shape-v1: [richcarni@f4298dc](https://github.com/richcarni/qtile/commit/f4298dcc70e2bb79ad89ba2b50143350410fd70a)
> 
> @Gurjaka you should be able to cherry-pick it into your branch
> 
> No additional dependencies outside of wayland client protocols
> 
> I'm not sure if this is the way to go - whether this could be expanded into something that supports multiple protocols or if it's better to have a client per protocol (or something in between)? Just hoping to help keep this moving :)
> 
> And for this PR, still needs some way of checking that the clients request has been handled correctly
> 
> [Disclaimer: some AI assistance used but mostly based on https://github.com/emersion/hello-wayland/blob/master/main.c]

Sorry for the delay, life's been rough past few days.

Getting to work rn!

Gurjaka:
@richcarni just did a tiny change to your client:
result value of `ftruncate` check is necessary, if it == -1 then we should log and exit the client (also reformatted with clang-format)

elParaguayo:
> Sorry for the delay, life's been rough past few days.

No need to apologise. Hope you're ok.

Gurjaka:
@elParaguayo Any idea why tests are cancelled?

elParaguayo:
They timed out. Looks like they hung when trying to install dependencies. I'll retrigger them.

Gurjaka:
> They timed out. Looks like they hung when trying to install dependencies. I'll retrigger them.

looks like everything is passing.

EDIT: looks like I forgot to remove unused function

Another EDIT: removed it

libqtile/backend/wayland/qw/cursor.h
```diff
@@ -18,6 +19,8 @@ struct qw_cursor {
     struct wlr_cursor *cursor;
     struct qw_view *view;
     struct qw_implicit_grab implicit_grab;
+    struct wlr_cursor_shape_manager_v1 *cursor_shape_mgr;
+    const char *current_shape_name;
```
richcarni, line 23: I can't see that this is being set anywhere?

Instead, would it be possible to get this from cursor_shape_manager and avoid having to store this extra state?

test/backend/wayland/test_cursor.py
```diff
@@ -0,0 +1,33 @@
...
+    try:
+        wmanager.test_window("cursor-shape-v1")
+
+        wmanager.c.core.warp_pointer(150, 150)
+
+        cursor_name = wmanager.c.eval("""
+from libqtile.backend.wayland._ffi import ffi
+cursor = self.core.qw_cursor
+ffi.string(cursor.current_shape_name).decode('utf-8') if cursor.current_shape_name != ffi.NULL else None
+        """)
```
richcarni: This might be what @tych0 was referring to about adding a test API - to avoid dealing with ffi in the tests? And probably doesn't belong in core.py

libqtile/backend/wayland/qw/cursor.h
Gurjaka: Looks like it's leftover from previous attempts of writing tests. Will remove it later today

test/backend/wayland/test_cursor.py
Gurjaka: I don't really understand what's wrong with this, or what we really should be doing instead. Would appreciate some help :smile:

richcarni: Maybe others will disagree but feels like too much code to put into the eval call

I think it would be nice to be able to call something like:
`cursor_shape = wmanager.c.core.get_cursor_shape_v1()`

Except that function probably doesn't belong in core? (Although not sure how else to make it work like that!)

Maybe `wmanager.c.core.cursor_info()["shape_v1"]`

richcarni:
> Seems like it would need to spawn an application that uses this protocol to set the cursor, and then confirm that the cursor is set. It might be useful to add a core API for testing purposes that exposes the cursor or similar.

Curious what you were thinking of in terms of a core API here @tych0 ?

test/backend/wayland/test_cursor.py
elParaguayo: Either of those two seems ok. I'd only use the latter if you wanted to expose more cursor info.

test/backend/wayland/test_cursor.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if shape == "wait":
+            assert cursor_name in ["wait", "watch"]
+        elif shape == "help":
+            assert cursor_name in ["help", "whats_this", "question_arrow"]
+        else:
+            assert cursor_name == shape
+
+    except Exception:
+        return
```
richcarni: um, this test is only passing because the try/except is catching any assert exceptions!

Gurjaka:
I hate every second of dealing with tests xd

Issue #5760, opened 2026-01-08
[wayland] Bar/wallpaper failing to draw with multiple displays


richcarni:
### Issue description

Migrating this issue from discord: https://discord.com/channels/955163559086665728/955650645833162822/1455238037616460034

@tpaniaki wrote:
>Last git build on Arch (0.34.2.dev17+g8eaf39dfe) needs a reload to display the bar and wallpaper. Can someone confirm?
>
> OK just tried a single-monitor boot and this works. Definitely an issue with multiple displays support.
Edit. Here is what seems to happen: as usual qtile sets screens to displays before shikane (kanshi); then shikane default profile is applied. However, now it seems that qtile doesn't properly responds to shikane profile loading resulting in screen0 (where the bar is) seemingly "staying" in my "third" monitor which is shut down by shikane (as my default profile requires). But... spawning app on screen0 works nevertheless, so I do have screen0 on my main monitor but without the bar. 

### Version

0.34.2.dev17+g8eaf39dfe

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

richcarni:
@richcarni:
>Not sure if this would be any help, but you can now fix qtile "Screens" to physical monitors by serial number:
> ```
>screens = [
>    Screen(
>        serial = <monitor_serial_no>,
>        bottom=bar.Bar(
>```

richcarni:
@tpaniaki:
>Hi rich, thanks for sneaking in. The problem with this is that shikane handles the scale factor as well. So I now have my bar correctly popping up on my main display but the scale (2x) is not applied as long as I don't reload the config.
>I guess it's easily reproducible on your end.

richcarni:
@richcarni:
>For me the issue seems to be the way displays without a serial (or a null serial) behave
>
>I added the option to match by name (DP-1, HDMI-1...) and that's fixed it for me
>
>You can test this with https://github.com/richcarni/qtile/tree/match-screen-name
>
>So just use name = instead of serial = (you can serial for one Screen, and name for another. not sure what will happen if you define both for a Screen - if this helps we can worry about that later)

richcarni:
@tpaniaki:
[...]

PR #5812, opened 2026-01-28
wayland: Fix output enable+position crash

### LINKED ISSUE #5807: [Wayland] Segmentation fault on disabling & then re-enabling a screen, possibly related to #5782
### Issue description

We talked about this on discord before, I'm not the only one that had this problem, but decided now to post an issue, even though it may not be potentially very useful & for that I'm sorry. Anyway I'm willing to do more troubleshooting steps if pointed in the right direction. 

1. Have 3 (or potentially just 2) displays, all enabled. 
2. *Disable display X 
3. *Enable display X
Depending on which display X is either nothing happens or qtile crashes with segfault, see my cropped photo below. It's worth noting that in my case it loves crashing if i disable & re-enable my side monitor, while my other 2 mirrored displays are untouched, but there are other crashing permutations. Tested this on Sway & it's fine over there. 

*I've tried wlr-randr & wdisplays the tool does not seem to matter.

This is what I see if I crash qtile while having started it from the TTY:
Image

### Version

0.34.2.dev55+g590b7663b

### Backend

Wayland (experimental)

### Config

```python
from libqtile import qtile

wl_xcursor_theme = "Adwaita"
wl_xcursor_size = 48

wl_input_rules = {
    "*": InputConfig(
        left_handed=False,
        pointer_accel=False,
        accel_profile="flat",
    ),

    "type:keyboard": InputConfig(
        kb_layout="hu",				
    ),
}

mod = "mod4"


keys = [
]


for i in groups:

layouts = [
    layout.MonadTall(),
    layout.Max(),
]

screens = [
    Screen(
        top=bar.Bar(
            [
                widget.GroupBox(),
                widget.Prompt(),
                widget.WindowName(),
            ],
            24,
        ),
    ),
]

import os
@hook.subscribe.startup_once
def autostart():
    qtile.cmd_spawn(os.path.expanduser("~/.config/qtile/autostart.sh")) # minimal config or not, I literally cannot use my computer without wlr-randr setting the displays.

```

### Logs

Other than my crappy screenshot above that I got from starting qtile in the shell, I've got nothing. Nothing in the qtile log, nothing in dmesg/journal, cannot find the coredump there.

Edit: systemd-coredump creates very large dump files such as this: `core.qtile.1000.e2654fd8b0d9416abfdf0a93da57983d.223272.1769353346000000.zst`  (13mb compressed, over 400 uncompressed)  Can send one as an email attachment if needed. 

### Required

richcarni:
Ensure output is committed before adding to layout

Fixes #5807

### DIFF
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -240,12 +240,6 @@ static void qw_server_output_manager_reconfigure(struct qw_server *server,
             wlr_output_state_set_transform(&state, head->state.transform);
             wlr_output_state_set_scale(&state, head->state.scale);
             wlr_output_state_set_adaptive_sync_enabled(&state, head->state.adaptive_sync_enabled);
-            struct wlr_box box;
-            wlr_output_layout_get_box(server->output_layout, head->state.output, &box);
-            if (box.x != head->state.x || box.y != head->state.y) {
-                wlr_output_layout_add(server->output_layout, head->state.output, head->state.x,
-                                      head->state.y);
-            }
             // TODO rescale the cursor if necessary
             // TODO: cursor_manager.load
             // TODO: set_xcursor if no surface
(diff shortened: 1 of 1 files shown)


Sinaaaa:
The crashes are gone. Cannot wait for this to be merged, thank you!

One thing I noticed though is that after re-enabling the side monitor only the clock will render from the bar, but this is just a variation on issue #5760, not anything directly related to this PR. (config reload still fixes it)

PR #5815, opened 2026-01-29
wayland: layers rejig [experimental!!]

richcarni:
An attempt to improve/simplify layering behaviour

Ideally this effort will be guided by user feedback (other than my own!) - so all feedback and ideas very much appreciated

Implemented so far:
- Group xwayland transients with their parent (on the scene graph). In English: A transient/child cannot go behind its parent and will move with its parent
- Group XDG children with their parent


elParaguayo:
Are you adding transients to a tree that's owned by the parent?

Personally, I like that approach. This is what I did in my PR for the X11 stacking. However, feedback n that was that transients should open above everything else, not just above their parent (which could be below other windows). I'm not convinced that's right but would like your thoughts.

richcarni:
> Are you adding transients to a tree that's owned by the parent?

Yes. Exactly that

I went and read through all the feedback for the X11 stacking PR - quite helpful - looks like you've done a lot of work/thinking on this already. Not all that surprised to learn this kind of transient behaviour is problematic. Turns out that managing windows is really complicated!

Anyway, picking up Tycho's example:
1. A is mapped
2. B is mapped
3. Ac is mapped
4. Bc is mapped


I think the top illustration is the current behaviour?

But, personally, I prefer the two below, which are compatible with child above parent - just need to bring in focus parent to top. 

So that raises the question, should focus raise a window (to the top of it's layer, not to keep-above layer) - this feels like the correct behaviour for floating windows and in most cases won't affect tiled windows?

(Also, ideally child would be centred over parent? I realise this isn't always possible and isn't a solution to this issue)

transient_dialog

Sinaaaa:
It is my understanding that currently the most annoying layering behavior, meaning the child going behind the floating parent occurs after 1px of pointer movement pushes the parent to the top. Meaning the user instinctively wants to move the cursor to be above the child, but the travel itself will hide it.

Personally I really like the idea that if I move my cursor whatever under it is gaining focus. It would be a bit sad if at one point this had to be sacrificed for more consistent layering. I actually have code that used to do this poorly on qtile xorg & a bit better but still poorly on awesomeWM. 

>Also, ideally child would be centered over parent?

Sure, that would be a decent default behavior, I just hope there would be a way to adjust this on the user end without needing to subclass window?.py.  For example imagine if we had window rule options for children, like offset x y % & attach to top bottom left right, or even simultaneously top & left for sticking to the window corner. Though I recognize that without me being able to contribute a PR for this, it's probably a bit too much to suggest such things :-P

richcarni:
> It is my understanding that currently the most annoying layering behavior, meaning the child going behind the floating parent occurs after 1px of pointer movement pushes the parent to the top.

This is a good argument for focus to not raise windows and leave it to `bring_front_click`. But I feel this is a bit weird with `follow_mouse_focus` and floating windows where you can give focus to a window that will remain partly obscured by another window

Actually, I think it should be possible to extend this child above parent behaviour to managed dialogs like save dialog and others. I mean if other desktops can...

elParaguayo:
Your bottom two images are what I had in mind but I understand why people may want/expect the top one. Whatever we do, I want to ensure the same behaviour in both X11 and wayland.

tpaniaki:
> This is a good argument for focus to not raise windows and leave it to bring_front_click.

`bring_front_click = True` doesn't prevent the child from losing focus and getting behind the parent without any click (if this is what you meant). 

But there're exceptions - typically libreoffice popups which stay on the foreground even while losing focus.

3hhh:
> Your bottom two images are what I had in mind but I understand why people may want/expect the top one.

IIRC the main point was that transient windows are usually dialog windows and only shown after direct user interaction (e.g. right click for dropdown or menu click) until they are hidden by the next user click.
So the topmost situation (two windows with transient children, both mapped) is unlikely to happen.
The layering code of the bottom two however may lead to a transient window being behind another tiled window which is not its parent (if the parent is unfocused, but was e.g. right clicked without focus change to show a dialog). This is counter-intuitive for users as they usually want to interact with the dialog window which they just opened.

So the issue mostly comes from a transient child window being opened for a parent that does not have focus.

One solution is to always put child windows on top of all others, another may be to transfer focus to them on creation. The latter however is problematic wrt focus steal protection as it enables all windows to steal focus by creating a child.

So I still vote for the first option.

I guess one could also make it configurable...

richcarni:
> The layering code of the bottom two however may lead to a transient window being behind another tiled window which is not its parent (if the parent is unfocused, but was e.g. right clicked without focus change to show a dialog). This is counter-intuitive for users as they usually want to interact with the dialog window which they just opened.

Thanks for pointing this out - I hadn't considered someone might want to access a dialog without focusing the parent. My solution for this would be to have a bring_top_click option/behaviour - raise the parent above it's siblings (windows in the same layer) when it receives the click

The thing is there are some big advantages (I think) of keeping the parent and transient "attached":
- If I grab and move a tiled window (i.e. convert to float), which has a menu open, the menu will move with the parent
- If I send a window to another screen while it has a menu open, both the parent and menu will be moved
- In wayland, at least, all this costs us very little because it fits well with how the scene graph works

And I suspect you may well be right that we will have to make this configurable - unless there is overwhelming support, I don't think we can modify an existing behaviour (unless it is clearly wrong/a bug)

Sinaaaa:
I really like the idea of strongly attaching the parent to its transients. One annoying example I can immediately think of that happens to me a lot is the following:

1. Have a rule to open App  Y on group Z.
2. open App Y
3. Move App Y to group C.
4. Open a dialog within App Y

Result: The dialog will open on group Z, while the app is locked out of interaction on group C. Admittedly this has to be  solvable within config.py, but still it's really not great.

3hhh:
On 2/10/26 7:12 AM, Richard Carnibella wrote:
> richcarni left a comment (qtile/qtile#5815)
> 
>> The layering code of the bottom two however may lead to a transient window being behind another tiled window which is not its parent (if the parent is unfocused, but was e.g. right clicked without focus change to show a dialog). This is counter-intuitive for users as they usually want to interact with the dialog window which they just opened.
> 
> Thanks for pointing this out - I hadn't considered someone might want to access a dialog without focusing the parent. My solution for this would be to have a bring_top_click option/behaviour - raise the parent above it's siblings (windows in the same layer) when it receives the click

This indeed sounds like an even more optimal solution.

richcarni:
Playing with the idea of not allowing a parent to steal focus from it's children (wayland only for now, i.e. not xwayland)

I suspect this makes sense for dialogs but probably not all children? (Seems like a lot of children are dialogs though!)

Not sure there is a definitive way to determine if a surface should behave like a modal dialog in wayland. So maybe there should be a way for qtile to specify (rules) which windows should have this behaviour. Similar to float_rules - modal_rules?

And I think it would be nice if child received focus again when app loses and regains focus (even if pointer is only over parent)

https://github.com/user-attachments/assets/6c4dd41d-373a-4f71-bbdb-c540ecb28498

richcarni:
Here is an issue with a menu going below the adjacent tiled window :(
PXL_20260623_104949632

elParaguayo:
Yup - this was the issue in the X11 PR too.

Instead of having the child in the parent's tree, should the parent just retain a list of children and, whenever the parent's moved, we make sure the children are displayed above it? That way, a child window can still open above other windows.

richcarni:
> Instead of having the child in the parent's tree, should the parent just retain a list of children and, whenever the parent's moved, we make sure the children are displayed above it? That way, a child window can still open above other windows.

Yeah, I mean maybe we're trying too hard to conform to the scene graph api when it's not really flexible enough to handle everything the way we want. I was thinking about a keep_above layer per layer (so LAYOUT_LAYER, FULLSCREEN_LAYER, etc. each have their own keep_above layer) could be combined with your suggestion of storing parent/child relations so that we can still have things like parent/children moving together, relative placement and no focus stealing

That said, a quick workaround for the hidden menu issue above is to use the client_focus_by_click hook:
```
@hook.subscribe.client_focus_by_click
def client_focus_by_click(client):
    client.move_to_top()
```

PXL_20260624_001202758

elParaguayo:
My sense is we should do the following:
* When a parent spawns a child, the child is moved to the top of the layer group containing the parent
* We store a reference between parents and children
* If a child is moved down, it cannot be moved below its parent - if this would happen then the parent must be moved down too
* Similarly, a parent should not be above its child - if that would happen then the child must be moved up too

Should parents and children always be in the same layer group?

richcarni:
> My sense is we should do the following:
> 
> * When a parent spawns a child, the child is moved to the top of the layer group containing the parent
> * We store a reference between parents and children
> * If a child is moved down, it cannot be moved below its parent - if this would happen then the parent must be moved down too
> * Similarly, a parent should not be above its child - if that would happen then the child must be moved up too

This doesn't sound unreasonable. I think it's going to be complicated, but if the rules are clear (like above), I think it stands a chance

I like the scene graph api because it's simple (though limited/inflexible) and we already have it. Of course since it's wayland only we're going to have to (at least) implement our own in python for x11. So yeah, maybe we should just build what we need

> Should parents and children always be in the same layer group?

I would say yes but I'm sure someone will have a use case that this doesn't work for

richcarni:
Actually the current issue I'm seeing with this PR is that because children are positioned relative to their parent, when i have a floating child dialog for a tiled parent, qtile is calculating centre of the screen, but this position, relative to it's parent is pushing it offscreen

Probably a timely reminder that qtile/python still needs to be responsible for managing windows!

(I think for the most part I do like/prefer that the dialog is centred over it's parent than in the middle of the screen)

elParaguayo:
Yeah. I was going to say that the stacking logic should be in the python backend so we can be consistent across x11 and wayland.

PR #5816, opened 2026-01-29
Drop cmd_ prefix fallback

tych0:
...and do a bunch of additional type checking.

This is a lot of small changes, probably easiest to review patch by patch. No Real Bugs here I don't think, but it does leave the code factored in a nicer way than it was originally (e.g. lots of x11 specific stuff that was in core is now pushed into backends/).

### DIFF
--- a/libqtile/config.py
+++ b/libqtile/config.py
@@ -1063,14 +1063,14 @@ class Match(_Match):
             return lambda other: other == value
         elif name == "wm_class":
 
-            def predicate(other) -> bool:  # type: ignore
+            def predicate(other) -> bool:
                 match = getattr(other, "match", lambda v: v == other)
                 return value and any(match(v) for v in value)
 
             return predicate
         else:
 
-            def predicate(other) -> bool:  # type: ignore
+            def predicate(other) -> bool:
                 match = getattr(other, "match", lambda v: v == other)
                 return match(value)
 
--- a/libqtile/core/loop.py
+++ b/libqtile/core/loop.py
@@ -29,7 +29,7 @@ class LoopContext(contextlib.AbstractAsyncContextManager):
 
         return self
 
-    async def __aexit__(self, *args) -> None:  # type: ignore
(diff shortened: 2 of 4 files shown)


tacaswell:
I was not watching the logs (because things were working) so I was a (the?) remaining user bit by this 😉 .

elParaguayo:
Do check the changelogs as well. The `cmd_` prefix was deprecated in v0.23.0 (September 2023) so you've had a couple of years of warnings in your logs!

tych0:
I wonder if we should do https://en.wikipedia.org/wiki/Brownout_(software_engineering) and fail starts in the future?

tacaswell:
I was checking nothing, it was happily working and I was letting Arch update itself as it went.  The first I learned about any of this is yesterday when some of my keybindings stopped working (took maybe 30-45 min to track it down, most of which was re-discovering that `qtile repl` existed).

With my maintainer-looking-at-complaining-user hat on: I completely agree that this is 100% on me for YOLO upgrading, not watching the CHANGELOG, and not checking my logs.

However with my user hat on, it would have been nice to have gotten a "push" that this was going to break rather than something I had to "pull" to find out about (I discovered the log location as part of debugging this).  Like dump something into the notification system (if one is running) to say you are using a deprecated command or on startup give a notification than the error/warning logs from the last session were over 1k lines or something (also fair if crossing over to user-space notification is Not A Thing, my maintainer instincts are more on the library side of things).

elParaguayo:
Thanks @tacaswell - I think that's a very fair position to take. I think it's very clear that most users don't look at a changelog and will only look at logs if something is broken. 

@tych0 - I wonder if we could inject a widget into the bar if a user has deprecated features in their config.

tych0:
Yeah, we could do that. Although I think this failure mode is hopefully mostly gone now with the invention of migrations? The `cmd_` stuff was from before that IIRC. But anyway, injecting something into the bar in red is reasonable, IMO

Issue #5818, opened 2026-02-01
segfault in CI on wayland backend


tych0:
### Issue description

```
2026-02-01T02:09:22.9013801Z ==================================== ERRORS ====================================
2026-02-01T02:09:22.9014558Z ___________ ERROR at setup of test_tasklist_defaults[1-wayland-top] ____________
2026-02-01T02:09:22.9014873Z 
2026-02-01T02:09:22.9015181Z request = <SubRequest 'tasklist_manager' for <Function test_tasklist_defaults[1-wayland-top]>>
2026-02-01T02:09:22.9016362Z manager_nospawn = <test.helpers.TestManager object at 0x7f1af20c2d50>
2026-02-01T02:09:22.9016867Z override_xdg = False
2026-02-01T02:09:22.9017254Z monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f1af1f8f620>
2026-02-01T02:09:22.9017663Z position = 'top'
2026-02-01T02:09:22.9017924Z 
2026-02-01T02:09:22.9018018Z     @pytest.fixture
2026-02-01T02:09:22.9018427Z     def tasklist_manager(request, manager_nospawn, override_xdg, monkeypatch, position):
2026-02-01T02:09:22.9019024Z         monkeypatch.setattr("libqtile.widget.tasklist.has_xdg", override_xdg)
2026-02-01T02:09:22.9019430Z     
2026-02-01T02:09:22.9019681Z         config = getattr(request, "param", dict())
2026-02-01T02:09:22.9019974Z     
2026-02-01T02:09:22.9020194Z         class TasklistConfig(Config):
2026-02-01T02:09:22.9020498Z             auto_fullscreen = True
2026-02-01T02:09:22.9020779Z             groups = [
2026-02-01T02:09:22.9021046Z                 libqtile.config.ScratchPad(
2026-02-01T02:09:22.9021358Z                     "SCRATCHPAD",
2026-02-01T02:09:22.9021632Z                     dropdowns=[
2026-02-01T02:09:22.9022076Z                         libqtile.config.DropDown("dd-a", spawn_cmd("dd-a"), on_focus_lost_hide=False),
2026-02-01T02:09:22.9022529Z                     ],
2026-02-01T02:09:22.9022762Z                 ),
2026-02-01T02:09:22.9023009Z                 libqtile.config.Group("a"),
2026-02-01T02:09:22.9023333Z                 libqtile.config.Group("b"),
[...]

PR #5823, opened 2026-02-03
Add section_linewidth option to TreeTab layout

florian-lnx:
Added a linewidth parameter to the TreeTab layout that controls the width of the separator line between sections. The parameter works with integer values to set the width or falsy values (0, None, False) to disable the line completely.


tych0:
This CI failure also looks suspiciously like #5818?

Issue #5827, opened 2026-02-05
Update the `#setting-up-the-environment` section in the documentation to reflect current status


Im-shamo:
https://docs.qtile.org/en/latest/manual/hacking.html#setting-up-the-environment

Is out-of-date as `./dev.sh` does not exist.

It should be something like `uv venv` in the root of the project. Active the venv. Then `make deps` to get the all dependencies.

I can update it later as I have time.

tych0:
A patch would be great, thanks!

PR #5828, opened 2026-02-06
docs: update `Setting up the environment`

Im-shamo:
#5827 
It was very out of date as `./dev.sh` does not exist.

### DIFF
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -48,11 +48,21 @@ dependencies.
 Setting up the environment
 ==========================
 
-In the root of the project, run ``./dev.sh``.
-It will create a virtualenv called ``venv``.
+In the root of the project, run ``uv venv``.
+It will create a virtualenv called ``.venv``.
 
-Activate this virtualenv with ``. venv/bin/activate``.
+Activate this virtualenv with
+
+* bash, zsh: ``source .venv/bin/activate``
+* fish: ``source .venv/bin/activate.fish``
+
 Deactivate it with the ``deactivate`` command.
+
+To Install every python dependencies that Qtile needs with ``make deps``.
+
+To build wayland backend with ``make build-wayland``.
+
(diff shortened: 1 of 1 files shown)


docs/manual/hacking.rst
```diff
@@ -48,12 +48,22 @@ dependencies.
...
+* fish: ``source .venv/bin/activate.fish``
 
-Activate this virtualenv with ``. venv/bin/activate``.
 Deactivate it with the ``deactivate`` command.
 
+To Install every python dependencies that Qtile needs with ``make deps``.
+
+To build wayland backend with ``make build-wayland``.
+
+Install locally into the virtualenv with ``uv pip install .``.
```
tych0, line 65: I wonder if it's worth a mention of the extra dependencies? I usually test by running `uv tool install .[dev,widgets,optional-core]`

Im-shamo: I don't know enough of `uv` to have an informed opinion lol. Cause I only used `pip` and `pipx`. But if you say so I guess it may be a good option.

Im-shamo: Also does't `uv tool` install into `~/.local/bin`? ie the user `$PATH`. So it just install from source? Like `pipx` ?

Im-shamo: I believe mentioning local install to `$PATH` in [Installation](https://docs.qtile.org/en/latest/manual/install/index.html#qtile) make more sense. Also we could place a link from that section to here for people who want to test it in a venv.

tych0: > Also does't uv tool install into ~/.local/bin? ie the user $PATH. So it just install from source? Like pipx ?

Yeah, exactly. But sorry, the `uv tool` was mostly a distraction, sorry. I was thinking of also adding the optional deps to the venv. But now I see you have `make deps` in there, which does `--all-extras`. So ignore me, I'm just being noisy.

PR #5829, opened 2026-02-06
docs: update install qtile-git with dependencies

Im-shamo:
Install optional dependencies when installing from git source.


tych0:
Thanks!

PR #5832, opened 2026-02-08
Revert "wayland: Fix output enable+position crash"

tych0:
This reverts commit 020a87db1d3ccc8f5cd44cef158ea9f3d4e030ad.

We have started seeing crashes after this commit on the wayland backend in CI. It is hard to see how this is related, but such is the life of memory corruption bugs.

Related to (and maybe Fixes:, but not sure...) #5818


richcarni:
Has it "resolved" itself? (stale bot appears to be a different issue)

image

tych0:
I don't think so, it's a race condition, so we've just been lucky recently.

PR #5833, opened 2026-02-08
ci: drop the 'stale' bot

tych0:
I waffle on whether or not this is useful.

We introduced it before github Discussions were a thing, so people were often opening issues for things that weren't real issues. Now that we have discussions, people have mostly been using those for that purpose, so we don't get as many non-issue Issues that are annoying to manage.

I also know that a bunch of the current maintainer team does not really like this bot :). So, let's talk about dropping it.


jwijenbergh:
The day has finally come

Issue #5820, opened 2026-02-01
JSON IPC is broken due to a type error in `IPCCommandServer`


Mikadore:
### Issue description

I was experimenting around with writing an IPC client in Rust for qtile. As part of that, 
since the protocol isn't documented, I looked at the code and found out the protocol 
is written so support both binary (`marshal`) and json encoding for messages. As far 
as I can tell, json isn't used by qtile anywhere. While trying to reverse engineer the 
protocol, I found out that using json will always produce a type error in the server handler.

## Reproduce 

I had already cloned the repo, I checked out the same version that's running on my desktop (`v0.34.1`).
I wanted to just experiment around with `qtile cmd-obj` and see what the communication looks like.

I used socat to create a relay socket to the qtile IPC socket, printing out all the traffic going through.
For binary, everything works fine. In order to use json, I made a very minor edit to my local repo:

```
diff --git a/libqtile/scripts/cmd_obj.py b/libqtile/scripts/cmd_obj.py
index fed57e38..8dafae6f 100644
--- a/libqtile/scripts/cmd_obj.py
+++ b/libqtile/scripts/cmd_obj.py
@@ -156,7 +156,7 @@ def cmd_obj(args) -> None:
 
     if args.obj_spec:
         sock_file = args.socket or find_sockfile()
-        ipc_client = Client(sock_file)
+        ipc_client = Client(sock_file, is_json=True)
         cmd_object = IPCCommandInterface(ipc_client)
         cmd_client = CommandClient(cmd_object)
         obj = get_object(cmd_client, args.obj_spec)
diff --git a/libqtile/scripts/main.py b/libqtile/scripts/main.py
index ad647edf..236fe31e 100644
--- a/libqtile/scripts/main.py
+++ b/libqtile/scripts/main.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import sys
 from pathlib import Path
+sys.path.append(str(Path(__file__).resolve().parent.parent.parent))
 
 from libqtile.log_utils import get_default_log, init_log
```
(the second is just so that the scripts use the local repo, using an empty venv without libqtile installed)

Now, running any command using `python libqtile/scripts/main.py cmd-obj -s /tmp/qtile.socket` will fail.

Providing any or no arguments doesn't matter, but for example:
```
$ python libqtile/scripts/main.py cmd-obj -s /tmp/qtile.socket -o window 123 -f info
Traceback (most recent call last):
[...]

PR #5835, opened 2026-02-09
Rework (JSON) IPC

Mikadore:
This for #5820

To get started with, I extracted IPC messages into their own types. I would welcome any feedback as to whether this goes into the right direction. Since `IPCCommandInterface` now returns `IPCReplyMessage` and takes in an `IPCCommandMessage`, some assumptions for users could be broken. However, most tests passed, (except IPC obviously). I got some weird regressions:

```
FAILED test/backend/test_idle_notify.py::test_idle_timer[1-x11-IdleConfig] - AssertionError: assert 'unset' == 'fired'
FAILED test/backend/test_idle_notify.py::test_idle_timer_inhibited[1-x11-IdleConfig] - AssertionError: assert '1' == '2'
FAILED test/widgets/test_mouse_callback.py::test_lazy_callback[1-x11] - libqtile.command.base.CommandException: ['Traceback (most recent call last):\n', '  File "/d/code/qtile_fork/libqtile/command/interface.py", line 425, in call\n    return ipc.IPCReplyMessage.success(cmd(*args, **kwargs))\n                                       ^^^^...
```
I'll have to look into that. The main advantage of my current approach, is it doesn't break backwards compat since old json format still works, there could be some attributes/warning logs to deprecate it though. I didn't actually add a nicer JSON wire format yet, but with this scaffolding that should be more or less trivial (just convert `IPCMessage` into a dict and `dumps`).

Let me know your thoughts!

### DIFF
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -124,10 +124,7 @@ class IPCCommandMessage(IPCMessage):
 
     @classmethod
     def from_json(cls, json: dict) -> IPCCommandMessage:
-        msg = IPCCommandMessage(**json)
-        # Must always lift a message deserialized from JSON
-        msg.lifted = True
-        return msg
+        return IPCCommandMessage(**json)
 
     def __iter__(self):
         return iter((self.selectors, self.name, self.args, self.kwargs, self.lifted))
@@ -199,7 +196,7 @@ class _IPC:
             The unpacked message
         """
         try:
-            obj = json.loads(data.decode())
+            obj = json.loads(data.decode(), object_hook=_IPC._json_tuple_object_hook)
             match obj:
                 case {"message_type": MessageType.COMMAND, "content": content}:
                     return IPCCommandMessage.from_json(content)
@@ -225,15 +222,31 @@ class _IPC:
             "message_type": msg.message_type,
             "content": msg.to_json(),
         }
-        json_obj = json.dumps(tagged_dict, default=_IPC._json_encoder)
+        json_obj = _IPC._HintTuplesJsonEncoder().encode(tagged_dict)
         return json_obj.encode()
 
+    class _HintTuplesJsonEncoder(json.JSONEncoder):
+        def encode(self, o):
+            def hint_tuple(o):
+                if isinstance(o, tuple):
+                    return {"$tuple": list(o)}
+                if isinstance(o, list):
+                    return [hint_tuple(i) for i in o]
+                if isinstance(o, dict):
+                    return {key: hint_tuple(val) for key, val in o.items()}
+                # This is to retain the old `_json_encoder` behavior
+                # of converting a set to a list for serialization
+                if isinstance(o, set):
+                    return hint_tuple(list(o))
+                return o
+
+            return json.JSONEncoder.encode(self, hint_tuple(o))
+
     @staticmethod
-    def _json_encoder(field: Any) -> Any:
-        """Convert non-serializable types to ones understood by stdlib json module"""
-        if isinstance(field, set):
-            return list(field)
-        raise ValueError(f"Tried to JSON serialize unsupported type {type(field)}: {field}")
+    def _json_tuple_object_hook(o):
+        if "$tuple" in o and len(o) == 1:
+            return tuple(o["$tuple"])
+        return o
 
 
 class Client:
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -13,9 +13,10 @@ import json
 import os.path
 import socket
 import struct
+import threading
 import traceback
 from abc import ABCMeta, abstractmethod
-from collections.abc import Callable, Iterator
+from collections.abc import Callable, Coroutine, Iterator
 from dataclasses import asdict, dataclass
 from typing import TYPE_CHECKING, Any, Self
 
@@ -89,7 +90,7 @@ class IPCMessage(metaclass=ABCMeta):
     @property
     @abstractmethod
     def message_type(self) -> MessageType:
-        """Discrimantor for the message type of the instance"""
+        """Discriminator for the message type of the instance"""
 
     @abstractmethod
     def to_json(self) -> dict:
@@ -299,7 +300,7 @@ class IPCStreamIO:
         await self.writer.wait_closed()
 
 
-class Client:
+class ReconnectingClient:
     def __init__(self, socket_path: str) -> None:
         """Create a new IPC client
 
@@ -317,7 +318,7 @@ class Client:
     def send(self, msg: tuple) -> IPCReplyMessage:
         """Send the message and return the response from the server
 
-        If any exception is raised by the server, that will propogate out of
+        If any exception is raised by the server, that will propagate out of
         this call.
         """
         return asyncio.run(self.async_send(msg))
@@ -330,6 +331,50 @@ class Client:
         """
         async with AsyncClient(self.socket_path) as c:
             return await c.send(IPCCommandMessage(*msg))
+
+
+class PersistentClient:
+    def __init__(self, socket_path: str) -> None:
+        self._loop = asyncio.new_event_loop()
+        self._thread = threading.Thread(target=self._loop.run_forever, daemon=True)
+        self._thread.start()
+        self._client = AsyncClient(socket_path)
+
+    def _run[T](self, coro: Coroutine[Any, Any, T]) -> T:
+        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
+        return future.result()
+
+    def connect(self) -> None:
+        self._run(self._client.connect())
+
+    def close(self) -> None:
+        self._run(self._client.close())
+
+    def send(self, msg: tuple) -> IPCReplyMessage:
+        # Since the PersistentClient retains the connection,
+        # the user should be aware of the need to close it.
+        # This can be done manually via `self.close()`, or
+        # using a with block. However, to remain compatible
+        # with the old `Client` implementation, we're gonna
+        # call `connect` ourselves here
+        if not self._client.is_connected():
+            # This should probably be a warning, but since none of the code
+            # using an IPC client is aware of closing right now, that would
+            # clutter the logs enormously
+            logger.debug("send called on PersistentClient that's not connected")
+            self._run(self._client.connect())
+
+        message = IPCCommandMessage(*msg)
+        return self._run(self._client.send(message))
+
+    def __enter__(self):
+        self.connect()
+
+    def __exit__(self, _exc_type, _exc, _tb):
+        self.close()
+
+
+Client = ReconnectingClient
 
 
 class AsyncClient:
@@ -356,6 +401,9 @@ class AsyncClient:
             raise IPCError(f"Could not open {self.socket_path}")
         except asyncio.TimeoutError:
             raise IPCError("Connection to server timed out")
+
+    def is_connected(self) -> bool:
+        return self.stream is not None
 
     async def send(self, message: IPCCommandMessage) -> IPCReplyMessage:
         """Send the message to the server using the existing connection.
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -269,27 +269,29 @@ class IPCStreamIO:
         self.writer.write(data)
         await self.writer.drain()
 
-    async def read_frame(self) -> bytes:
+    async def read_frame(self) -> bytes | None:
         """Reads the the frame header (length) and then reads and returns the expected
-        number of bytes"""
+        number of bytes. Returns None if underlying stream has closed"""
         try:
             frame_header = await self.reader.readexactly(self.FRAME_HEADER_LENGTH)
             frame_length = struct.unpack(self.FRAME_HEADER_FORMAT, frame_header)[0]
             data = await self.reader.readexactly(frame_length)
             return data
-        except asyncio.IncompleteReadError:
-            raise IPCError("Invalid message framing, couldn't read the data")
+        except asyncio.IncompleteReadError as e:
+            if self.reader.at_eof():
+                return None
+            else:
+                logger.debug(f"Received partial bytes: {e.partial!r}")
+                raise IPCError("Invalid message framing, couldn't read the data")
 
     async def write_message(self, message: IPCMessage):
         await self.write_frame(_IPC.pack(message))
 
-    async def read_message(self, *, timeout: float | None = None) -> IPCMessage:
+    async def read_message(self, *, timeout: float | None = None) -> IPCMessage | None:
         message_bytes = await asyncio.wait_for(self.read_frame(), timeout=timeout)
+        if message_bytes is None:
+            return None
         return _IPC.unpack(message_bytes)
-
-    def at_eof(self) -> bool:
-        """Returns `reader.at_eof()`"""
-        return self.reader.at_eof()
 
     async def close(self):
         """Closes the connection"""
@@ -431,13 +433,18 @@ class Server:
         logger.debug("Connection made to server")
 
         try:
-            while not stream.at_eof():
+            while True:
                 # There is no timeout here to enable long lived connections
                 # by clients, without having to implement a heartbeat protocol
                 # which would impose a huge burden on the current implementation.
                 # Clients are assumed to be trusted, although there is no actual
                 # verification mechanism for this
                 req = await stream.read_message()
+                # EOF
+                if req is None:
+                    logger.debug("Client disconnected")
+                    break
+
                 if not isinstance(req, IPCCommandMessage):
                     logger.error("Expected command message from client")
                     break
--- a/libqtile/ipc.py
+++ b/libqtile/ipc.py
@@ -12,6 +12,7 @@ import fcntl
 import json
 import os.path
 import socket
+import struct
 import traceback
 from abc import ABCMeta, abstractmethod
 from collections.abc import Callable, Iterator
@@ -249,6 +250,53 @@ class _IPC:
         return o
 
 
+class IPCStreamIO:
+    """Wraps an asyncio `StreamReader` and `StreamWriter` and implements
+    a simple framing protocol to handle sending and receiving IPCMessages
+    """
+
+    FRAME_HEADER_FORMAT = "!L"
+    FRAME_HEADER_LENGTH = struct.calcsize(FRAME_HEADER_FORMAT)
+
+    def __init__(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter):
+        self.reader = reader
+        self.writer = writer
+
+    async def write_frame(self, data: bytes):
+        """Prepends the data with the frame header (length) and writes it to the writer"""
+        frame_header = struct.pack(self.FRAME_HEADER_FORMAT, len(data))
+        self.writer.write(frame_header)
+        self.writer.write(data)
+        await self.writer.drain()
+
+    async def read_frame(self) -> bytes:
+        """Reads the the frame header (length) and then reads and returns the expected
+        number of bytes"""
+        try:
+            frame_header = await self.reader.readexactly(self.FRAME_HEADER_LENGTH)
+            frame_length = struct.unpack(self.FRAME_HEADER_FORMAT, frame_header)[0]
+            data = await self.reader.readexactly(frame_length)
+            return data
+        except asyncio.IncompleteReadError:
+            raise IPCError("Invalid message framing, couldn't read the data")
+
+    async def write_message(self, message: IPCMessage):
+        await self.write_frame(_IPC.pack(message))
+
+    async def read_message(self, *, timeout: float | None = None) -> IPCMessage:
+        message_bytes = await asyncio.wait_for(self.read_frame(), timeout=timeout)
+        return _IPC.unpack(message_bytes)
+
+    def at_eof(self) -> bool:
+        """Returns `reader.at_eof()`"""
+        return self.reader.at_eof()
+
+    async def close(self):
+        """Closes the connection"""
+        self.writer.close()
+        await self.writer.wait_closed()
+
+
 class Client:
     def __init__(self, socket_path: str) -> None:
         """Create a new IPC client
@@ -282,26 +330,21 @@ class Client:
             reader, writer = await asyncio.wait_for(
                 asyncio.open_unix_connection(path=self.socket_path), timeout=3
             )
+            stream = IPCStreamIO(reader, writer)
         except (ConnectionRefusedError, FileNotFoundError):
             raise IPCError(f"Could not open {self.socket_path}")
 
         try:
-            send_data = _IPC.pack(IPCCommandMessage(*msg))
-            writer.write(send_data)
-            writer.write_eof()
-
-            read_data = await asyncio.wait_for(reader.read(), timeout=10)
-        except TimeoutError:
+            await stream.write_message(IPCCommandMessage(*msg))
+            response = await stream.read_message(timeout=10.0)
+        except asyncio.TimeoutError:
             raise IPCError("Server not responding")
         finally:
-            # see the note in Server._server_callback()
-            writer.close()
-            await writer.wait_closed()
+            await stream.close()
 
-        data = _IPC.unpack(read_data)
-        if not isinstance(data, IPCReplyMessage):
+        if not isinstance(response, IPCReplyMessage):
             raise IPCError("Expected a reply message from the server")
-        return data
+        return response
 
 
 class Server:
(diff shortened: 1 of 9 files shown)


libqtile/ipc.py
```diff
@@ -69,53 +77,232 @@ def find_sockfile(display: str | None = None):
     raise IPCError("Could not find socket file.")
 
 
+class IPCWireFormat(enum.StrEnum):
+    JSON_RAW = "json_raw"
+    """Formats the message data directly as a tuple/list in json"""
+
+    JSON_TAGGED = "json_tagged"
+    """Formats the message as a dict, contained within a top level
```
tych0: I think we only need to support this, unless you have a strong argument for why we would support the others. (I'm ignoring backwards compatibility as one of the strong arguments :)

libqtile/ipc.py
```diff
@@ -69,53 +77,232 @@ def find_sockfile(display: str | None = None):
...
+
+    @staticmethod
+    def exception(exception: Exception) -> IPCReplyMessage:
+        """Construct a reply message from an exception
+        with status EXCEPTION. The exception is formatted to
+        provide useful information to the recipient"""
+        # DEV: The original code only returned the last line
+        # of the traceback, whereas this will return a list
+        # containing the whole traceback
+        data = traceback.format_exception(exception)
```
tych0, line 187: This is great, we don't have this now and it's annoying.

libqtile/ipc.py
```diff
@@ -69,53 +77,232 @@ def find_sockfile(display: str | None = None):
...
+                raise IPCError(
+                    f"Malformed list message. Expected 2 or 5 elements, got: {len(data)}"
+                )
+
+    @staticmethod
+    def _from_tagged_dict(message_type: str, content: dict) -> IPCMessage:
+        # There could be more type validation done here
+        # before trying to unpack content
+        match message_type:
+            case "command":
```
tych0: Can we match against IPCCommandMessage.message_type() somehow, rather than writing the string twice? Perhaps by making `message_type()` a class method and setting up a dict of the types doing dynamic dispatch that way? It would just be nice to only have to add an option in one place when introducing a new message type.

tych0: Or maybe:
```
class MessageType(StrEnum):
    Command = "command"
    REPL = "repl"

    def parse(...)

    def unparse(...)
```

and then everyone can pattern match on that?

tych0 (approved):
Looks like a nice start, though I think you can drop all the compat stuff. Have you thought about how to make a socket handle more than one message, so it can deal with stateful things like the repl connections will be?

libqtile/ipc.py
Mikadore: `JSON_RAW` is really only for backwards compatibility. My current impl for `BYTES` is the same as the old one, just with the scaffolding. You could potentially make an argument for supporting `marshall` serialization because of performance, but unless we have benchmarks, and care a lot (for IPC, probably not, at least not now), I don't think that's 

THE thing `BYTES` has going for it, is that `marshall` preserves basic python types, whereas JSON will convert tuples into lists, and doesn't support things like set, etc. This is something that has to be solved anyway for working JSON IPC though, so at that point, why not just ditch `BYTES`, it definitely makes the code a lot simpler. 

TL;DR: I'm happy to ditch `BYTES` and `JSON_RAW`

Preserving tuples <-> lists opens another discussion related to type lifting, and how to approach it. There's a bug in my implementation for `JSON_TAGGED` I'm hunting down currently, and as part of it I wanna write some tests for (un)packing, and optimally some tests for `cmd-obj` using JSON too (or just using `Client` perhaps). If everything works fine with the current approach of calling `lift_args` in `IPCCommandServer.call`, that's great, but optimally I think that lifting should be tied to the IPC-specific code, since it's only relevant if a message has been serialized->deserialized over a socket (I left a comment in #5820 similar to what I'm describing here)

I think something like:
```py
    def __post_init__(self):
        if isinstance(self.args, list):
            self.args = tuple(self.args)
            # same for selectors
```
Would be fine just fix-up the basic tuple <-> list issue though

Mikadore: Oh, I'm just stupid. Somewhere along the refactoring process I messed up and didn't actually tag the messages anywhere. Even worse, for debugging I was sending commands to my main systems socket, not the instance rendering to xephyr :facepalm: So, with a simple fix it works (always setting `lifted = True` when unpacking). My main point still stands, but it could be considered out of scope for this PR

https://github.com/qtile/qtile/pull/5835/changes/9d73d6eb1bbe382be5b8db26be9a3d21d05fd121

libqtile/ipc.py
Mikadore: Good point, this was an oversight. I added a module-level `MessageType` string enum in [026477d](https://github.com/qtile/qtile/pull/5835/commits/026477d9d662df1d986b94129a52b72d8cf72855), since unfortunately python doesn't have anonymous member enums. The name could be changed to `IPCMessageType` though, or just moved into `IPCMessage`, which would make things more verbose. Since I can't really think of a good use for the enum outside the `ipc` module's internals, I think it's fine as is

Mikadore:
I'll probably go ahead with dropping the other formats then, this eliminates the need for `WireFormat` and `is_json` in `Client` for example. It should be easy enough to re-add them (`BYTES`, for example) if desired. Once tests are added, I think that fulfills the original goal of fixing json ipc/making the format more readable (+ some restructuring of the code)  

> Looks like a nice start, though I think you can drop all the compat stuff. Have you thought about how to make a socket handle more than one message, so it can deal with stateful things like the repl connections will be?

Thanks!  I'm a bit confused on what exactly you mean by 'repl', since the `qtile repl` lives in `interactive/repl.py` and uses it's own socket/server implementation. My original comment about handling multiple messages with a single connection was more of an observation, that currently the `Client` will `connect` to the unix domain socket for each message, and the server will close the connection after the message is handled. So basically, it looks to me you could more or less wrap that code in a `while not is_eof():`
```py
    async def _server_callback(
        self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter
    ) -> None:
        """Callback when a connection is made to the server

        Read the data sent from the client, execute the requested command, and
        send the reply back to the client.
        """
        try:
            logger.debug("Connection made to server")
            data = await reader.read()
            logger.debug("EOF received by server")

            req, wire_format = _IPC.unpack(data)
            if not isinstance(req, IPCCommandMessage):
                logger.error("Expected command message from client")
                return
        except IPCError as e:
            logger.warning("Invalid data received, closing connection")
            logger.debug(e)
        else:
            # Don't handle requests when session is locked
            if self.locked.is_set():
                rep = IPCReplyMessage.error({"error": "Session locked."})
            else:
                rep = self.handler(req)

            result = _IPC.pack(rep, wire_format=wire_format)

            logger.debug("Sending result on receive EOF")
            writer.write(result)
            logger.debug("Closing connection on receive EOF")
            writer.write_eof()
        finally:
            writer.close()
            await writer.wait_closed()
```
Although I'm not familiar with `asyncio` in python in particular

Mikadore:
I went ahead and removed the other formats. Everything seems to work fine, except tests will fail when they expect a tuple, since they get a list. This is seemingly because the `TestManager`'s `self.c` is an `IPCCommandInterface` over an actual `ipc.Client`. I don't know if this is intentional, since I'd expect the tests to not necessarily care about IPC, so they could just use a `QtileCommandInterface`. This goes back to the [discussion](https://github.com/qtile/qtile/pull/5835#discussion_r2794749274) about `BYTES`'s advantage being preserving tuples. Qtile itself runs seemingly without issues (in xephyr), so I'm not sure how to fix this exactly. I found a very hacky solution on [stack overflow](https://stackoverflow.com/questions/15721363/preserve-python-tuples-with-json), which I really don't like, but should work. Alternatively, we can change the tests, by either not going over actual IPC, or just assuming arrays everywhere. The distinction betwen a tuple and an array in IPC is questionable in itself imo, does it really matter whether something is `[x, y, z]` or `(x, y, z)` conceptually? (It does in python, of course). If we were to tie type lifting to `IPC*Message`, we could just lift the reply as well, although. 

Failed tests (+ipc, but I've yet to write ones up to date):
```
FAILED test/layouts/test_matrix.py::test_matrix_navigation[1-x11-MatrixConfig] - assert [0, 2] == (0, 2)
FAILED test/layouts/test_ratiotile.py::test_ratiotile_add_windows[1-x11-RatioTileConfig] - assert [[0, 0, 800, 600]] == [(0, 0, 800, 600)]
FAILED test/layouts/test_ratiotile.py::test_ratiotile_add_windows_golden_ratio[1-x11-RatioTileConfig] - assert [[0, 0, 800, 600]] == [(0, 0, 800, 600)]
FAILED test/layouts/test_ratiotile.py::test_ratiotile_alternative_calculation[1-x11-RatioTileConfig] - assert [[0, 0, 800, 600]] == [(0, 0, 800, 600)]
FAILED test/layouts/test_ratiotile.py::test_resizing[1-x11-RatioTileConfig] - assert [[0, 0, 160, ... 0, 160, 600]] == [(0, 0, 160, ... 0, 160, 600)]
FAILED test/test_bar.py::test_geometry[1-x11-GeomConf] - assert [0, 0, 800, 10] == (0, 0, 800, 10)
FAILED test/test_command.py::test_items_qtile[x11-ServerConfig-2] - assert [True, [0, 1, 2]] == (True, [0, 1, 2])
FAILED test/test_command.py::test_items_group[1-x11-ServerConfig] - assert [True, [4194306]] == (True, [4194306])
FAILED test/test_command.py::test_items_screen[1-x11-ServerConfig] - assert [True, [0, 1, 2]] == (True, [0, 1, 2])
FAILED test/test_command.py::test_items_bar[1-x11-ServerConfig] - assert [True, []] == (True, [])
FAILED test/test_command.py::test_items_layout[1-x11-ServerConfig] - assert [True, []] == (True, [])
FAILED test/test_command.py::test_items_window[x11-ServerConfig-2] - assert [True, []] == (True, [])
FAILED test/test_command.py::test_items_widget[1-x11-ServerConfig] - assert [True, []] == (True, [])
FAILED test/test_fakescreen.py::test_gaps[1-x11-FakeScreenConfig] - assert [0, 316, 500, 24] == (0, 316, 500, 24)
```

tych0:
> Thanks! I'm a bit confused on what exactly you mean by 'repl', since the qtile repl lives in interactive/repl.py and uses it's own socket/server implementation.

Right, one of my goals is that we could unify that separate socket/server into the IPC implementation here with distinct message types. I think we could get rid of a lot of code that way.

> So basically, it looks to me you could more or less wrap that code in a while not is_eof()

agreed, I would like to do that as part of this work as well (though with `async while`, not just `while`; IIRC the code is not structured super well to do that today).

> I don't know if this is intentional, since I'd expect the tests to not necessarily care about IPC, so they could just use a QtileCommandInterface.

They can't do that though, because the qtile that's running in the tests is a separate process. The IPC is how we talk to that process. (In principle we could run the tests in the *same* process via async which would fix many problems for us, but I haven't actually tried it.)

> except tests will fail when they expect a tuple, since they get a list.

I wonder if we can/should do type lifting on the client end as well, and it could look at return types?

Mikadore:
> > Thanks! I'm a bit confused on what exactly you mean by 'repl', since the qtile repl lives in interactive/repl.py and uses it's own socket/server implementation.
> 
> Right, one of my goals is that we could unify that separate socket/server into the IPC implementation here with distinct message types. I think we could get rid of a lot of code that way.

That should be doable, I haven't looked into the repl code in too much depth, however it definitely requires a stateful connection, but we want to be able to do that anyway. I'm not sure if it'll actually reduce a lot duplication, since it would probably require a bunch of new message types (`repl_start`, `repl_eval`, `repl_end`?). It would definitely be nice not to have two separate servers/sockets running to talk to Qtile though.  

> > So basically, it looks to me you could more or less wrap that code in a while not is_eof()
> 
> agreed, I would like to do that as part of this work as well (though with `async while`, not just `while`; IIRC the code is not structured super well to do that today).

Sure, I'll look into that.
 
> > I don't know if this is intentional, since I'd expect the tests to not necessarily care about IPC, so they could just use a QtileCommandInterface.
> 
> They can't do that though, because the qtile that's running in the tests is a separate process. The IPC is how we talk to that process. (In principle we could run the tests in the _same_ process via async which would fix many problems for us, but I haven't actually tried it.)

This is fixed now

> > except tests will fail when they expect a tuple, since they get a list.
> 
> I wonder if we can/should do type lifting on the client end as well, and it could look at return types?

On the last point, I have thought about that quite extensively as well. My idea was to add a top level `type_hint` (or similarly named) field, that would contain a dict with the same structure as content, but the values would be the types of each field in `content`. This is rather painful though, and adds a lot of code overhead.  After talking to @elParaguayo on discord, I went ahead with converting tuples to a special object in `pack`, and converting back to tuples in `unpack`. I searched through the code base, and I couldn't find an instance of a command taking, or returning any custom object, the only issue is the tuples, which should be solved by this (although this solution is perhaps a bit unpleasant for custom IPC client implementors, it works and is rather simple)

@elParaguayo said we don't want to send/receive custom types over IPC anyway, which makes sense to me. I think that type lifting in `interface.py` shouldn't be necessary at all now, although I haven't tried it yet.

Other than the repl and stateful/connection reusing stuff, I think the JSON/IPC part itself is complete, except for tests of course. If y'all like the code/json format (and tuple handling) as it is, I can go ahead and write the tests, and look into the aforementioned extras

Mikadore:
Re: removing lifting

I removed lift_args and ignored `lifted` within `IPCCommandServer.call` to see if everything works.
All tests passed, except `cmd-obj` tests that specifically tested the lifting logic. So, my understanding
is that currently lifting is necessary, because `cmd-obj` will send all args and kwargs as strings (straight
from the CLI), and relies on them being lifted to int/float by the server. So I'll just leave that untouched
for now. I'm not sure how, or if it's even a good idea, but there's a possibility of leaving the conversion
of str -> number to the cli itself (perhaps just check if the str is all alphanumeric, and use a special syntax for numeric strings?). This would make the IPC server code cleaner, and remove a lot of unnecessary  `False` literals in the code base when invoking a command within qtile.

This isn't a priority at all, but something to consider perhaps

Mikadore:
@tych0 I'm pretty satisfied with the core changes to the IPC system. The repl requires a stateful connection. All the pieces are there to support that, however, for reasons outlined in `2a333d4`, I had left the `Client` to reopen a connection for each request. 

> Originally, I wanted to add the same logic to `Client`,
reusing an underlying connection. However, since `asyncio.run`
will create a new event loop, no streams (connections) created
within that loop can be reused. I tried having separate logic
in the synchronous and asynchronous send methods, where the async_send
method would reuse `self.stream`, and synchronous `send` would always
create a new connection. This lead to pretty messy code, however,
so I decided to just add a new `AsyncClient`.

> `Client` is implemented by creating an `AsyncClient`, and sending one
message. With `async with`, connecting and closing the underlying
streams happens automatically.

I figured out how to properly wrap the `AsyncClient`, that is by running an event loop on a separate thread, and blocking on the results. There is *some* overhead because of the need to create a thread (more below), but the main reason I haven't just replaced the original `Client`, is that using this approach requires the user to close the connection manually (or use a `with` block).

So for now, the `ipc` module has a `ReconnectingClient` (the original implementation), and a `PersistentClient` (sync wrapper for `AsyncClient`), with `Client` being an alias for `ReconnectingClient`. The `connect`/`disconnect` handling (optimally, users should be forced to call `connect` manually or use a `with` block, but right now it just logs a warning with level debug) is imo the biggest criterion for deciding which to keep (or both?), so I'm keen on your input. Below, there are some more observations on the performance characteristics of each: 

### Performance differences

I did make a (rather simple)  [benchmark](https://github.com/Mikadore/qtile/blob/ipc_client_bench/bench_client.py) comparing the two on a separate branch of my fork. Here are the results:
```
Running: Single use with setup: Reconnecting
Min:    0.6611ms Max:    0.7202ms Mean:    0.6831ms Median:    0.6772ms
Running: Single use with setup: Persistent
Min:    0.6488ms Max:    1.2031ms Mean:    0.7560ms Median:    0.6860ms
Running: Single use with existing client: Reconnecting
Min:    0.6522ms Max:    0.7124ms Mean:    0.6651ms Median:    0.6609ms
Running: Single use with existing client: Persistent
Min:    0.3590ms Max:    0.4420ms Mean:    0.3828ms Median:    0.3761ms
Running: Multiple use: Reconnecting
Min: 1350.2837ms Max: 1350.2837ms Mean: 1350.2837ms Median: 1350.2837ms
Running: Multiple use: Persistent
Min: 1222.8516ms Max: 1222.8516ms Mean: 1222.8516ms Median: 1222.8516ms
```
On average, the two approaches are about the same for creating a client and running a single command, although the `PersistentClient` has much worse worst-case performance. However, in my (very imperfect) `multi_use` example (opening and then killing 10 windows), there's a ~140ms (varies, it runs only one measurement) performance gain by using the persistent client, which makes sense (also, most of the test is probably made up by running `xdotool` with `subprocess` to get the window id from the given `pid` that `spawn` returns)

tych0:
Sorry, I haven't had time to look at this carefully recently. I will try to take a look this weekend.

libqtile/ipc.py
```diff
@@ -69,121 +73,379 @@ def find_sockfile(display: str | None = None):
...
+    REPLY = "reply"
+
+
+class IPCMessage(metaclass=ABCMeta):
+    """Abstract base class for all IPC messages"""
+
+    @property
+    @abstractmethod
+    def message_type(self) -> MessageType:
+        """Discrimantor for the message type of the instance"""
```
ervinpopescu: Typo :) -> Discriminator

ervinpopescu: I'll also take a look with my client and see if there is anything else to spot here other than the format change.

Mikadore: @ervinpopescu Nice, my idea is to basically just move all the networking out of the `repl` module, and leave a `QtileREPL` (or a differently named object) that handles all the repl logic. Then we'd have IPC message types like `repl_start`, `repl_completion_(request|reply)`, `repl_code`, `repl_result` (or response, whatever), and `repl_end`. All of which correspond to some method/constructor of `QtileREPL`. That means you can just open an IPC connection (via `PersistentClient`) in the repl CLI, and don't have to start and wait for a separate repl server. I'll start working on it today/tomorrow probably, but I welcome any input! If you're active on discord, feel free to HMU any time, my nick is `mikdore`, I'm in the qtile server (or just message me here, whatever you prefer). 

Thanks for pointing out a typo, made a commit to fix it, I'll push it out with some other work soon

Mikadore:
I pushed a proof of concept merger of the repl and the IPC code. I'm not 100% happy with the code as it is, there are some things like passing `qtile` to the server as well as the handler that could probably be improved, but I think this illustrates how I would approach this pretty well. I removed the repl client/server tests, since they are heavily based on the custom socket protocol, but they should be rather easy to add again if we settle on an interface/approach for merging the two. Manually testing the repl works well.

libqtile/ipc.py
ervinpopescu: Hi again,

I tested the current status of the code, and it looks pretty easy to integrate with (well, i did use gemini to scaffold most of the code) -- PR https://github.com/ervinpopescu/qtile-cmd-client/pull/135 if you wanna look.

https://github.com/qtile/qtile/pull/5835/changes#diff-7234a0e155e3b26310bb112c87469b08ae25b4bf7463d66036a158fa1bbe7c88R179 did help a lot to integrate :)

libqtile/ipc.py
```diff
@@ -69,121 +73,379 @@ def find_sockfile(display: str | None = None):
...
         then wait for and return the response from the server.
         """
+        async with AsyncClient(self.socket_path) as c:
+            return await c.send(IPCCommandMessage(*msg))
+
+
+class PersistentClient:
+    def __init__(self, socket_path: str) -> None:
+        self._loop = asyncio.new_event_loop()
+        self._thread = threading.Thread(target=self._loop.run_forever, daemon=True)
```
tych0, line 397: Do we need a thread here? I was really hoping we could have an `async def` and run all of these things in the main loop.

tych0 (commented):
Ok, I finally looked at this and it looks reasonable. I see that you did a length-prefixed framing protocol, which I think makes sense, though it will mean people have to do a bit of scaffolding to write clients, but such is life. 

In addition to my comment about threading, can you fix up the git history and squash out the commits that removed the backwards compat stuff?

libqtile/ipc.py
Mikadore: It is necessary imo. `PersistentClient` is a direct wrapper of `AsyncClient`, but it makes the interface synchronous by offloading to the thread and blocking on the `result()` of the future given by `asyncio.run_coroutine_threadsafe`. 

> I was really hoping we could have an `async def` and run all of these things in the main loop.

Afaik `Client` (alias for `ReconnectingClient`) is only used in scripts, or the library to provide `CommandClient`/`InteractiveCommandClient`/the `lazy` stuff, so the main event loop qtile runs isn't available to them (it could maybe be done for `lazy` stuff in user configs, but I'd have to look into how that works in detail, which I haven't). If the process running the `Qtile` instance would want to use an ipc client (what for?), it can always use `AsyncClient` if it's within the main loop's context. 

I mentioned it before somewhere, but currently there's two IPC clients. `PersistentClient` is required for the repl to work using the IPC server, but I didn't want to replace the old client (now `ReconnectingClient`, aliased to `Client`) because `PersistentClient` requires the user to manage disconnecting either manually or with a `with` statement (sidenote: I just noticed `PersistentClient` leaks the thread in the `close` method, oops.)

If you can think of a nice solution to this problem (maybe refactoring uses of `Client` to properly close is very easy?), I'd be keen to hear them.

Mikadore:
Thanks for reviewing. I've been out of the country the past few days, I hope to get this PR mergeable this week. That means mainly re-adding the REPL tests, maybe expanding on the IPC tests, and cleaning up the history as you said

libqtile/ipc.py
tych0: > Afaik Client (alias for ReconnectingClient) is only used in scripts, or the library to provide CommandClient/InteractiveCommandClient/the lazy stuff, so the main event loop qtile runs isn't available to them (it could maybe be done for lazy stuff in user configs, but I'd have to look into how that works in detail, which I haven't). 

Sorry, you're right. I was thinking this was on the server side. Still, I wonder if people using the client code can do their own `asyncio.run` so we don't have to? I guess that would mean things like `qtile cmd-obj` would need this `asyncio.run`, instead of having it in the library. It just feels impolite to create threads in libraries to me, I guess.

> If you can think of a nice solution to this problem (maybe refactoring uses of Client to properly close is very easy?), I'd be keen to hear them.

Yeah, I think that's it. Probably a lot of try/finally or using context managers like you have. The old code was pretty lazy about these things, so any adjustment will necessarily require us to introduce a bit more rigor.

Mikadore:
Sorry this took so long, I was feeling off. I've managed to overcome my poor git skills and clean up the commit history. I think this is good as is, arguably [c086b10](https://github.com/qtile/qtile/pull/5835/commits/c086b101300dd4a21d399476c5ba5aefa3f1301b) could be squashes as well, but when I last tried it got a little messy. I'm pretty happy with the state of the PR, I think tests could be improved, and I would like to add docs eventually (especially for the IPC format, directed potential external implementors), but that can be done in a separate PR just as well. 

LMK how you want to proceed @tych0

ervinpopescu:
@tych0 just to keep the ball rolling, could you enable the tests here? I am curious how many of the tests pass :)

elParaguayo:
You can ping me too if you want the tests run!

tych0:
Sorry, I had dental surgery a couple of days ago, still catching up :)

tych0:
Overall I think this looks pretty good. A couple of complaints, these may or may not be things we should fix:

* I don't think we should do the alias `Client = ReconnectingClient`; I think we really shouldn't have a reconnecting mode at all. Let's just make `Client` the `PersistentClient`.
* I think `PersistentClient` shouldn't spawn a thread, but require users to do `asyncio.run()` somewhere themselves; that means that we'll need to do it in `qtile repl`, `qtile shell`, etc.


I guess it needs a rebase and to fix the linter complaints. But overall I think it looks pretty good, thank you for doing this.

tych0:
Oh, and I guess we need some repl tests too?

elParaguayo:
I think there are some (basic) repl tests already. What did you have in mind?

tych0:
This code deletes two of them which seem important. Unless I'm misreading the diff?

elParaguayo:
That;s what I get for looking on my phone! I'll let mikadore update for your other comments but can help with tests after that.

PR #5845, opened 2026-02-16
wayland: Add z-layer stacking tests with fixes

richcarni:
Move these tests from backend/x11/test_window to common test_window

Currently all tests pass on wayland except `test_fullscreen_on_top`, which is skipped for the time being

### DIFF
--- a/test/test_window.py
+++ b/test/test_window.py
@@ -567,3 +567,61 @@ def test_floats_kept_above(manager):
     # Open a different floating window. This should be above the first floating one.
     manager.test_window("three", floating=True)
     assert _clients() == ["two", "one", "three"]
+
+
+@manager_config
+def test_fullscreen_on_top(manager):
+    """Test fullscreen, focused windows are on top."""
+
+    if manager.backend.name == "wayland":
+        pytest.skip("TODO: Fix Wayland failing")
+
+    def _clients():
+        stack = manager.backend.get_all_windows()
+        wins = [(w["name"], stack.index(w["id"])) for w in manager.c.windows()]
+        wins.sort(key=lambda x: x[1])
+        return [x[0] for x in wins]
+
+    manager.test_window("one", floating=True)
+    manager.test_window("two")
+
+    # window "one" is kept_above, "two" is norm
+    assert _clients() == ["two", "one"]
+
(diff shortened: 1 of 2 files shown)


libqtile/backend/wayland/window.py
```diff
@@ -70,11 +70,12 @@ def keep_below(self, enable: bool | None = None) -> None:
         is_enabled = self.layer == lib.LAYER_KEEPBELOW
         if enable is None:
             enable = not is_enabled
-
         if enable:
             self.reparent(lib.LAYER_KEEPBELOW)
+            self.move_to_bottom()
         else:
             self.reparent(lib.LAYER_LAYOUT)
+            self.move_to_bottom()
```
richcarni, line 78: Is this correct/make sense?

test/backend/wayland/test_window.py
```diff
@@ -18,6 +74,7 @@
 pytestmark = pytest.mark.skipif(not has_layer_shell, reason="GtkLayerShell not available")
```
richcarni, line 18: Do we need gtk layer shell? We can add it as a test dependency or if we don't actually need it currently, avoid importing and remove this skip condition

richcarni: Actually ManagerConfig with widgets works if we don't import GtkLayerShell (I don't know why!), so if I make the stacking test a common test, it should be fine

And we currently don't need layershell for tests because they currently aren't managed by qtile, so can't test info on it anyway

test/test_window.py
```diff
@@ -431,3 +432,196 @@ def test_follow_mouse_focus(manager, follow_mouse_focus):
...
+    assert _clients() == ["two", "one"]
+
+    # A fullscreen, focused window should display above windows that are "kept above"
+    window_by_name(manager.c, "two").enable_fullscreen()
+    window_by_name(manager.c, "two").focus()
+    assert _clients() == ["one", "two"]
+
+    # Focusing the other window should cause the fullscreen window to drop from the highest layer
+    window_by_name(manager.c, "one").focus()
+    assert _clients() == ["two", "one"]
```
richcarni, line 598: This is the one that is failing on wayland

Issue #5317, opened 2025-07-07
popups of xwayland apps do not pop up


cr1ogen:
### Issue description

Popup menus stopped showing in Creality Print in somes popups , same problem in Orca Slicer
Edit: It doesn't work on Steam either.

https://github.com/user-attachments/assets/cdfe270f-13cd-4b64-8936-5507ce7126e0

### Version

0.32.1.dev28+gfe45cd78 - Same in 0.32

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

- [x] I have searched past issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

rpenido:
I have the same problem here.

elParaguayo:
Just a quick note to say we're not ignoring you!

There's a lot of work at the moment to rewrite our Wayland backend and that's where the majority of the Wayland dev time is being spent. I suspect the preference will be too finish the new backend (which should hopefully remove this issue) rather than work on a fix only to see that code removed by the new backend.

Happy to discuss further if you have any questions/comments.

cr1ogen:
> Just a quick note to say we're not ignoring you!
> 
> There's a lot of work at the moment to rewrite our Wayland backend and that's where the majority of the Wayland dev time is being spent. I suspect the preference will be too finish the new backend (which should hopefully remove this issue) rather than work on a fix only to see that code removed by the new backend.
> 
> Happy to discuss further if you have any questions/comments.

Sounds good to me. I'm looking forward to the new backend.

ervinpopescu:
@cr1ogen is this still an issue?

cr1ogen:
> [@cr1ogen](https://github.com/cr1ogen) is this still an issue?

Sorry, I was traveling. It's fixed now.
Edit: It was fixed in an update, but I updated again today and it failed again.

cr1ogen:
Pop-up windows don't work on Steam either.

Qtile version 0.34.1.dev18+g3ab6c3535

cr1ogen:
Please re open the issue
`qtile -version 0.34.2.dev1+gb219d2742`

cr1ogen:
Could you please take another look? It's not working in apps that use xwayland, for example, Creality Print, Steam, Winecfg... They fixed it in one update, but it happened again in another.
qtile --version 0.34.2.dev42+g1e13874d2

[...]

PR #5846, opened 2026-02-16
wayland: fix layer-shell popups and dialogs appearing behind windows

### LINKED ISSUE #5776: [wayland] Some apps unexpectedly moving to BRINGTOFRONT layer
### Issue description

This creates an issue with some dialogs that go to LAYERKEEPABOVE appearing behind the main app

For me GIMP and FreeCAD are both exhibiting this behaviour:

- Open GIMP (v3.0.6)
- Run `qtile cmd-obj -o core -f stacking_info` and find GIMP under LAYER_LAYOUT
- Open a file to edit
- Run `qtile cmd-obj -o core -f stacking_info` and now find GIMP under LAYER_BRINGTOFRONT

Unsure if this is related to other open "bring to front issues" on X11: #4805, #5725

### Version

#8badb08

### Backend

Wayland (experimental)

### Config

```python
bring_front_click = False
floats_kept_above = True
floating_layout = layout.Floating(
    border_focus="#308280",
    border_width=4,
    float_rules=[
        Match(wm_class="confirmreset"),  # gitk
        Match(wm_class="makebranch"),  # gitk
        Match(wm_class="maketag"),  # gitk
        Match(wm_class="ssh-askpass"),  # ssh-askpass
        Match(title="branchdialog"),  # gitk
        Match(title="pinentry"),  # GPG key password entry
        Match(wm_class="keepassxc"),  # org.keepassxc.KeePassXC for Wayland
        Match(wm_class="org.keepassxc.KeePassXC"),
        Match(wm_class="LANDrop"),
        Match(func=lambda c: c.is_transient_for() is not None),
    ],
)
```

### Logs

```text

```

### Required

Gurjaka:
Fixes #5776

Problem: XDG popups (context menus) and dialog windows were appearing behind their parent windows due to incorrect scene graph layer placement.

Root cause: Nested popups were inheriting their parent's scene tree instead of being placed in a top layer. Dialog windows were being moved to `LAYER_BRINGTOFRONT` before Python's manage_view_cb ran, causing the Python layer assignment to override the C code's intent.

Solution:
- Place all XDG popups in `LAYER_BRINGTOFRONT` to ensure they appear above regular windows
- Move dialog layer override to after manage_view_cb so Python's config runs first, then C code enforces dialogs stay on top

This ensures popups and dialogs always appear above their parent windows regardless of the parent's layer or user configuration.

### DIFF
--- a/libqtile/backend/wayland/qw/xdg-view.c
+++ b/libqtile/backend/wayland/qw/xdg-view.c
@@ -448,7 +448,6 @@ static void qw_xdg_view_handle_new_popup(struct wl_listener *listener, void *dat
 // Handle map event: when the xdg_view becomes visible/mapped
 static void qw_xdg_view_handle_map(struct wl_listener *listener, void *data) {
     UNUSED(data);
-
     struct qw_xdg_view *xdg_view = wl_container_of(listener, xdg_view, map);
     xdg_view->mapped = true;
 
@@ -460,6 +459,14 @@ static void qw_xdg_view_handle_map(struct wl_listener *listener, void *data) {
     xdg_view->base.app_id = xdg_view->xdg_toplevel->app_id;
 
     struct wlr_xdg_toplevel *xdg_toplevel = xdg_view->xdg_toplevel;
+
+    // check if this window has a parent (i.e., it's a dialog)
+    bool is_dialog = (xdg_toplevel->parent != NULL);
+    struct qw_xdg_view *parent_view = NULL;
+
+    if (is_dialog) {
+        parent_view = xdg_toplevel->parent->base->data;
+    }
 
     // Create foreign toplevel manager and listeners
     if (xdg_view->base.ftl_handle == NULL) {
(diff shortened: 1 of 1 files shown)


cr1ogen:
See this too please https://github.com/qtile/qtile/issues/5317

richcarni:
I think #5776 has a different root cause - that one is where activation requests are bringing tiled windows to the front

It does sound like #5137 though!

I am in favour of what I was experimenting with in #5815, which is to place transient popup/dialogs in the scene graph as children of their parent app - this means they can never go below their parent and also move with their parent (for example to a different screen). The only issue is, technically they can go below an adjacent tiled window and/or its popups/dialogs. So some more thinking/work required there

That said, I'm not necessarily against this as a temporary solution. I have a questin though, why is python's layer assignment not bringing these to front or keepabove, since the are floating and we have `floats_kept_above = True`?

cr1ogen:
> I think It does sound like #5137 though!
> 
>
>

Is #5317 xD

Issue #4519, opened 2023-10-14
[Bug / X11] Fullscreen windows not being resized properly when moved to screen with a different resolution


Crystal4276:
### Issue description

Fullscreen windows not being resized properly when moved to screen with a different resolution

Here a video showing the problem:

https://github.com/qtile/qtile/assets/27446196/ceb2a4cb-a8f2-4276-a194-699fe800fdce

### Version

0.23.0

### Backend

X11 (default)

### Config

```python
import os
import subprocess
import asyncio
import psutil

from libqtile import bar, layout, hook, qtile
from libqtile.log_utils import logger
from libqtile.config import Click, Drag, Group, Key, KeyChord, Match, Screen, Rule, DropDown
from libqtile.lazy import lazy
#from libqtile.widget import Spacer

from qtile_extras import widget
from qtile_extras.widget.decorations import RectDecoration, PowerLineDecoration, BorderDecoration
from qtile_extras.widget import modify

from popups.power_menu import show_power_menu
from widgets.padgroupbox import padGroupBox


# Parsing : remove all text.
def txt_remove(text):
    return ""

# Send a window within a group to group displayed on left or right screen. Three monitors configuration, monitor 0 is the central monitor -> Screens: [1,0,2] 
def window_to_previous_screen(qtile, switch_group=False, switch_screen=False):
    i = qtile.screens.index(qtile.current_screen)
    if i == 0:
        group = qtile.screens[i + 1].group.name
        qtile.current_window.togroup(group, switch_group=switch_group)
        if switch_screen == True:
            qtile.to_screen(i + 1)
    if i == 2:
        group = qtile.screens[i - 2].group.name
        qtile.current_window.togroup(group, switch_group=switch_group)
        if switch_screen == True:
            qtile.to_screen(i - 2)

def window_to_next_screen(qtile, switch_group=False, switch_screen=False):
    i = qtile.screens.index(qtile.current_screen)
    if i == 1:
        group = qtile.screens[i - 1].group.name
        qtile.current_window.togroup(group, switch_group=switch_group)
        if switch_screen == True:
            qtile.to_screen(i - 1)
    if i == 0:
        group = qtile.screens[i + 2].group.name
        qtile.current_window.togroup(group, switch_group=switch_group)
        if switch_screen == True:
            qtile.to_screen(i + 2)

# Key bindings for group stick to screen
def go_to_group(name: str) -> callable:
    def _inner(qtile) -> None:
        if len(qtile.screens) == 1:
[...]

Issue #4219, opened 2023-03-31
group.next_window and group.prev_window behave unintuitively with fullscreen windows


austin5627:
### The issue:

Fullscreen windows always have the `floating` attribute set to True but remain in the layout they were in before being fullscreened. If a window was in the normal layout before being fullscreened it has the `floating` attribute true while still being in the normal layout.
If the current window is fullscreen, this section of code from `next_window` will always set `nxt` to the last window in the normal layout, vice versa with `prev_window`
```  python    
if self.current_window.floating:
    nxt = (
        self.floating_layout.focus_previous(self.current_window)
        or self.layout.focus_last()
        or self.floating_layout.focus_last(group=self)
    )
else:
    nxt = (
        self.layout.focus_previous(self.current_window)
        or self.floating_layout.focus_last(group=self)
        or self.layout.focus_last()
    )

```
This makes it impossible to switch to floating windows if there are no other windows in the normal layout as next/prev will always switch to the current window.
Similarly if there are two fullscreen windows it is impossible to switch to any windows between them in the layout

qtile version: 0.22.1

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
Thanks. Yes, I see the problem.

So, rather than checking the `floating` attribute, we should check whether the window is tiled or not.

Issue #4058, opened 2022-12-12
Full screen toggle forgets floating windows size and location


justinesmithies:
### The issue:

Qtile version: 0.22.2.dev80+g9e16cb5b

I have a hook that when mpv opens it floats, resizes and positions the mpv window. 

```
@hook.subscribe.client_managed
def repos(window):
    if window.get_wm_class() and 'mpv' in window.get_wm_class():
        window.floating = True
        window.place(1200, 650, 640, 360, 2, "#ffffff")
```

When i use my key bind to full screen toggle the window it works as expected but when I then toggle it again I'd expect it to return to the exact size and location plus floating. It doesn't it just tiles the mpv window.

```
    # Toggle Fullscreen
    Key([mod], "f",
        lazy.window.toggle_fullscreen(),
        lazy.hide_show_bar(position='all'),
        desc='Toggle fullscreen and the bars'
        ),
```

Is this normal behaviour ? As if I use my key bind to move the mpv window to another workspace or monitor works as expected.

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

elParaguayo:
I think that's normal. I don't think we save position/state when toggling fullscreen.

justinesmithies:
Oh OK in that case could it be added as an enhancement as I for one would find it incredibly useful and I'm sure others would too ?

m-col:
I think this also happens when windows are minimised, or at least they forget whether they were floating. We currently track this float state where each window has exactly one of these flags:
```py
@enum.unique
class FloatStates(enum.Enum):
    NOT_FLOATING = 1
    FLOATING = 2
    MAXIMIZED = 3
    FULLSCREEN = 4
    TOP = 5
    MINIMIZED = 6
```
Perhaps this could be improved by instead tracking floating vs tiled as one flag and then the other states as another.

jwijenbergh:
I agree that we need to do a refactor of how fullscreen windows work. Just getting fullscreen windows to properly restore the layout was a hassle to figure out

mitchdost:
Hi,

think I'm also having a reproducable issue that belongs to the same topic. Until 0.22.1 a virtual box machine opens in fullscreen and non-floating mode.

In version 0.23.0 it opens in floating mode and the upper left corner of the window is positioned in the middle of the screen.
May it be possible, that the actual problem is the floating mode?

I attached two files with output from xprop. One for 0.22.1 and one for 0.23.0

[...]

Issue #3517, opened 2022-04-22
Colum order not retained when maximize using lazy.window.toggle_maximize() 


druidamix:
### The issue:

Using qtile version 0.21.0
While using column layout, calling lazy.window.toggle_maximize() and toggling back, the original window position is lost. 

On post  [https://github.com/qtile/qtile/issues/1098]  was mentioned toggle_maximize() and toggle_fullscreen(), then on
([https://github.com/qtile/qtile/issues/1482],  was resolved toggle_fullscreen(). Can you do the same for toggle_maximize()?

I'm using the sample config.py with:
  Key([mod], "o", lazy.window.toggle_maximize()),
and layout:
layout.Columns(border_focus_stack=["#d75f5f", "#8f3d3d"], border_width=4),

https://gfycat.com/rawbarejunco gif showing the problem.

Thanks

### Required:

- [X] I have searched past issues to see if this bug has already been reported.

jwijenbergh:
It makes sense to also add this functionality for maximized windows. I tried to create a more general fix in https://github.com/qtile/qtile/pull/2724 but that ended up being too many code changes

ramnes:
Should be fixed now: https://github.com/qtile/qtile/pull/3057

ramnes:
Ah sorry, it only works for fullscreen windows but not maximized ones, right?

mauromotion:
Any updates on this? I just switched from Awesome to Qtile and I love it but I miss the maximize functionality without the windows swapping all the time (in Columns). I'm trying to go around it by binding a toggle to the Max layout but it's getting ridiculous (remembering the prev layout requires storing layout's state somehow, I think?).

PR #5849, opened 2026-02-24
Refactor floating

richcarni:
My attempt to revive @jwijenbergh's floating refactor branch

I believe this was an effort to address a number of issues around floating, maximized and fullscreen windows, specifically but probably not limited to: https://github.com/qtile/qtile/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22floating%20refactor%22

The gist of it is to have multiple secondary layouts for floating, maximized and fullscreen windows, decoupling floating from having to provide all those functions

It's a pretty big one and is still a work in progress, but seems to be stable enough to be test

As always, feedback is greatly appreciated!

**Current tests status**
wayland ✅
=== 1320 passed, 79 skipped, 5 warnings ===
x11 ✅
=== 1415 passed, 19 skipped ===

TODOS:

- [x] x11 support
- [x] existing tests
- [ ] new tests
- [ ] clean up code!
- [ ] migrations
- [ ] documentation

Verify fixes:

- [ ] #4519
- [ ] #4219
- [ ] #4058
- [ ] #3517
- [ ] #2512
- [ ] #2415
- [ ] #2088 
- [ ] #1444 
- [ ] #1201

### DIFF
--- a/libqtile/backend/x11/window.py
+++ b/libqtile/backend/x11/window.py
@@ -493,6 +493,8 @@ class _Window:
         self._float_width: int = self._width
         self._float_height: int = self._height
 
+        self._win_state_follows = True
+        self._prev_win_state: WindowStates | None = None
         # We use `previous_layer` to see if a window has moved up or down a "layer"
         # The layers are defined in the spec:
         # https://specifications.freedesktop.org/wm-spec/1.3/ar01s07.html#STACKINGORDER
@@ -652,7 +654,16 @@ class _Window:
         # So here we simply re-place the window at the coordinates which will magically solve the issue
         # This only seems to be an issue with unfullscreening, thus we check if we're fullscreen and the window wants to unfullscreen
         elif self.fullscreen and "fullscreen" not in state:
-            self._reconfigure_floating(new_win_state=WindowStates.FULLSCREEN)
+            self.place(
+                self.x,
+                self.y,
+                self.width,
+                self.height,
+                self.borderwidth,
+                self.bordercolor,
+                above=False,
+                respect_hints=True,
+            )
 
         for s in triggered:
             attr = s
@@ -692,7 +703,7 @@ class _Window:
             group=group,
             id=self.window.wid,
             wm_class=self.get_wm_class(),
-            floating=self._win_state != WindowStates.TILED,
+            floating=self._win_state == WindowStates.FLOATING,
             float_info=float_info,
             maximized=self._win_state == WindowStates.MAXIMIZED,
             minimized=self._win_state == WindowStates.MINIMIZED,
@@ -894,9 +905,10 @@ class _Window:
                     y += int(height_adjustment / 2)
 
         # save x and y float offset
-        if self.group is not None and self.group.screen is not None:
-            self.float_x = x - self.group.screen.x
-            self.float_y = y - self.group.screen.y
+        if self.float_x is None or self.float_y is None or self.floating:
+            if self.group is not None and self.group.screen is not None:
+                self.float_x = x - self.group.screen.x
+                self.float_y = y - self.group.screen.y
 
         self.x = x
         self.y = y
@@ -1713,24 +1725,97 @@ class Window(_Window, base.Window):
     def edges(self):
         return (self.x, self.y, self.x + self.width, self.y + self.height)
 
+    def set_property_by_state(self, state: WindowStates) -> None:
+        if state == WindowStates.TILED:
+            self.tiling = True
+        if state == WindowStates.FLOATING:
+            self.floating = True
+        if state == WindowStates.FULLSCREEN:
+            self.fullscreen = True
+        if state == WindowStates.MAXIMIZED:
+            self.maximized = True
+        if state == WindowStates.MINIMIZED:
+            self.maximized = True
+
+    def set_property_from_prev_win_state(self, default: WindowStates) -> None:
+        # If a floating window is toggled it should return to the main layout
+        # If a fullscreen window is toggled it should return to the main layout
+        assert self.group is not None and self.group.layout is not None
+        if self.floating and self._win_state == WindowStates.FLOATING:
+            state = self.group.layout._manages_win_state
+        elif self.fullscreen and self._win_state == WindowStates.FULLSCREEN:
+            state = self.group.layout._manages_win_state
+        elif self.maximized and self._win_state == WindowStates.MAXIMIZED:
+            state = self.group.layout._manages_win_state
+        else:
+            state = self._prev_win_state
+
+        # if state == self._win_state or state is None:
+        if state is None:
+            state = default
+
+        # If we want to set tiling but the current layout doesn't support tiling windows
+        # We set the state to the default layout state
+        if (
+            state == WindowStates.TILED
+            and self.group is not None
+            and self.group.layout._manages_win_state != WindowStates.TILED
+        ):
+            state = self.group.layout._manages_win_state
+            # Already set, set floating instead
+            if self._win_state == self.group.layout._manages_win_state:
+                state = WindowStates.FLOATING
+        # If the state is equal
+        self.set_property_by_state(state)
+
+    @property
+    def tiling(self) -> bool:
+        return self._win_state == WindowStates.TILED
+
+    @tiling.setter
+    def tiling(self, do_tiled: bool) -> None:
+        if do_tiled:
+            self._prev_win_state = self._win_state
+            if self.group and self.group.screen:
+                self.group.mark_tiling(self)
+        else:
+            self.set_property_from_prev_win_state(default=WindowStates.FLOATING)
+
     @property
     def floating(self):
-        return self._win_state != WindowStates.TILED
+        return self._win_state == WindowStates.FLOATING or self._win_state == WindowStates.TOP
 
     @floating.setter
     def floating(self, do_float):
         stack = self.qtile.core._root.query_tree()
         tiled = [win.window.wid for win in (self.group.tiled_windows if self.group else [])]
         tiled_stack = [wid for wid in stack if wid in tiled and wid != self.window.wid]
-        if do_float and self._win_state == WindowStates.TILED:
+        if do_float:
+            if self.floating:
+                return
+
             if self.is_placed():
                 screen = self.group.screen
-                self._enablefloating(
-                    screen.x + self.float_x,
-                    screen.y + self.float_y,
-                    self._float_width,
-                    self._float_height,
-                )
+                if not self._float_width:  # These might start as 0
+                    self._float_width = self.width
+                    self._float_height = self.height
+
+                if self.group and self.group.layout:
+                    if self._win_state == self.group.layout._manages_win_state:
+                        self.float_x = self.x - screen.x
+                        self.float_y = self.y - screen.y
+
+                self._prev_win_state = self._win_state
+                # if we are setting floating early, e.g. from a hook, we don't have a screen yet
+                if self.group and self.group.screen:
+                    self.group.mark_floating(self)
+
+                # self._enablefloating(
+                #     screen.x + self.float_x,
+                #     screen.y + self.float_y,
+                #     self._float_width,
+                #     self._float_height,
+                # )
 
                 # Make sure floating window is placed above tiled windows
                 if tiled_stack and (not self.kept_above or self.qtile.config.floats_kept_above):
@@ -1746,21 +1831,25 @@ class Window(_Window, base.Window):
             if not self.kept_above and self.qtile.config.floats_kept_above:
                 self.keep_above(enable=True)
 
-        elif (not do_float) and self._win_state != WindowStates.TILED:
-            self.update_fullscreen_wm_state(False)
-            if self._win_state == WindowStates.FLOATING:
-                # store last size
-                self._float_width = self.width
-                self._float_height = self.height
-            self._win_state = WindowStates.TILED
-            self.group.mark_floating(self, False)
+        else:
+            if not self.floating:
+                return
+            self.set_property_from_prev_win_state(default=WindowStates.TILED)
+
+            # self.update_fullscreen_wm_state(False)
+            # if self._win_state == WindowStates.FLOATING:
+            #     # store last size
+            #     self._float_width = self.width
+            #     self._float_height = self.height
+            # self._win_state = WindowStates.TILED
+            # self.group.mark_floating(self, False)
             if self.kept_above and self.qtile.config.floats_kept_above:
                 self.keep_above(enable=False)
             if tiled_stack:
                 self.window.configure(
                     stackmode=xcffib.xproto.StackMode.Above, sibling=tiled_stack[-1]
                 )
-            hook.fire("float_change")
+        hook.fire("window_state_change")
 
     @property
     def wants_to_fullscreen(self):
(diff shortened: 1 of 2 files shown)


Sinaaaa:
Thank you so much for attempting this!

I have a dumb monkey patch to do a great many things, this breaking that is expected, but there is one -I hope- interesting tidbit.
Basically in `manager.py` have "replaced" `process_button_motion` & I'm running `win.place(win.x, win.y, win.width, win.height, 11, "00000000")`  in the replacement. Before this PR this prevented qtile from calculating the border at all. Now though It's doing all the border math. How do I know? Well, my super fancy qtile-extras derivative borders cannot be rendered in motion, 3fps is the best my computer can do.

I will do a bit more testing to make this hopefully easier to reproduce & understand & edit this comment.

Edit: Ok, so deleted my patch & tried what happens if I set 00000000 - fully transparent black for the -fancy- active border. The results are that with this PR qtile is working really hard to render the border, without this PR it's not drawing it. 

Edit2: With this PR, using the minimal config cannot reproduce https://github.com/qtile/qtile/issues/5843 anymore. (though still could be a fluke, I have tried a few different display configurations)

tpaniaki:
Thanks Richard.

A quick feedback.

- Scratchpads are now sometimes getting under the layer of a newly spawned window. But not always.
- In the Max layout, my windows are all spawned as floating windows maximized. 
- Maximization still don't work on scratchpad. This is probably desired as it would switch group but it might be interesting to have a universal toggle.
- If I send a fullscreen window to another screen, the background of the initial screen still is blank (black). 
- I have a bunch of functions regarding my floatings that don't work anymore. For instance:

```
@hook.subscribe.client_new
def set_floating_windows(client):
    if client.is_transient_for():
        return

    if any(rule.compare(client) for rule in floating_windows_square):
        float_center(client, 800, 600)
        return

    if any(rule.compare(client) for rule in floating_windows_rectangular):
        float_center(client, 900, 950)
        return
```

richcarni:
@tpaniaki, thanks for testing and taking the time to leave feedback

Obviously this will need to support all kinds of user customizations (with zero to minimal modifications) but it may also be worth testing with the default config to help isolate the cause of some of the behaviour you're seeing

>Scratchpads are now sometimes getting under the layer of a newly spawned window. But not always.

Still need to address layering but I think we can do that separately (and hopefully speedily)

>In the Max layout, my windows are all spawned as floating windows maximized.

If you mean window info says they are both maximized and floating, that is wrong. Fixed

>Maximization still don't work on scratchpad. This is probably desired as it would switch group but it might be interesting to have a universal toggle.

Hadn't considered this as something to support. It kind of works for me though. What are you seeing?

>If I send a fullscreen window to another screen, the background of the initial screen still is blank (black).

I don't see this behaviour, using default `lazy.window.togroup(i.name, switch_group=True)` keybind. Are you doing something different?


>```
>@hook.subscribe.client_new
>def set_floating_windows(client):
>    if client.is_transient_for():
>        return
>
>   if any(rule.compare(client) for rule in floating_windows_square):
>        float_center(client, 800, 600)
>        return
>```

Is it the matching or the float_center function that is failing?

richcarni:
Thanks for the feedback @Sinaaaa 

> I have a dumb monkey patch to do a great many things, this breaking that is expected, but there is one -I hope- interesting tidbit.
> Basically in `manager.py` have "replaced" `process_button_motion` & I'm running `win.place(win.x, win.y, win.width, win.height, 11, "00000000")` in the replacement. Before this PR this prevented qtile from calculating the border at all. Now though It's doing all the border math. How do I know? Well, my super fancy qtile-extras derivative borders cannot be rendered in motion, 3fps is the best my computer can do.

>Edit: Ok, so deleted my patch & tried what happens if I set 00000000 - fully transparent black for the -fancy- active border. The results are that with this PR qtile is working really hard to render the border, without this PR it's not drawing it.

It could well be that something got replaced/modified when rebasing the original code (I think it was something like 600+ commits behind master!) I'll take a look

>Edit2: With this PR, using the minimal config cannot reproduce https://github.com/qtile/qtile/issues/5843 anymore. (though still could be a fluke, I have tried a few different display configurations) 

Not sure what might be going on there but happy to take any win!

tpaniaki:
> I don't see this behaviour, using default lazy.window.togroup(i.name, switch_group=True) keybind. Are you doing something different?

No I use `lazy.window.toscreen()`.

> Is it the matching or the float_center function that is failing?

I logged this:

```
@hook.subscribe.client_new
def set_floating_windows(client):
    logger.warning("client_new: START")
    logger.warning(f"client_new: name={client.name!r}")
    logger.warning(f"client_new: wm_class={(client.get_wm_class() or [])!r}")
    logger.warning(f"client_new: app_id={getattr(client, 'app_id', None)!r}")
    logger.warning(f"client_new: wid={getattr(client, 'wid', None)!r}")
    logger.warning(f"client_new: floating={getattr(client, 'floating', None)!r}")
    logger.warning(f"client_new: fullscreen={getattr(client, 'fullscreen', None)!r}")
    logger.warning(f"client_new: maximized={getattr(client, 'maximized', None)!r}")
    logger.warning(f"client_new: minimized={getattr(client, 'minimized', None)!r}")

    try:
        transient = bool(client.is_transient_for())
    except Exception as e:
        logger.warning(f"client_new: is_transient_for() ERROR: {e!r}")
        transient = False

    logger.warning(f"client_new: transient={transient!r}")
    if transient:
        logger.warning("client_new: transient window -> RETURN")
        return

    square_match = False
    rectangular_match = False

    try:
        square_match = any(rule.compare(client) for rule in floating_windows_square)
    except Exception as e:
        logger.warning(f"client_new: square rules compare ERROR: {e!r}")

    logger.warning(f"client_new: square_match={square_match!r}")
    if square_match:
        logger.warning("client_new: calling float_center(client, 800, 600)")
        try:
            float_center(client, 800, 600)
            logger.warning("client_new: float_center DONE (800x600)")
        except Exception as e:
            logger.warning(f"client_new: float_center ERROR (800x600): {e!r}")
        return

    try:
        rectangular_match = any(rule.compare(client) for rule in floating_windows_rectangular)
    except Exception as e:
        logger.warning(f"client_new: rectangular rules compare ERROR: {e!r}")

    logger.warning(f"client_new: rectangular_match={rectangular_match!r}")
    if rectangular_match:
        logger.warning("client_new: calling float_center(client, 900, 950)")
        try:
            float_center(client, 900, 950)
            logger.warning("client_new: float_center DONE (900x950)")
        except Exception as e:
            logger.warning(f"client_new: float_center ERROR (900x950): {e!r}")
        return

    logger.warning("client_new: no rule matched -> END")
```

And got:

```
2026-02-25 14:38:36,513 WARNING libqtile config.py:set_floating_windows():L311 client_new: START
2026-02-25 14:38:36,513 WARNING libqtile config.py:set_floating_windows():L312 client_new: name='Rédaction du message'
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L313 client_new: wm_class=['org.mozilla.Thunderbird']
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L314 client_new: app_id=None
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L315 client_new: wid=9
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L316 client_new: floating=False
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L317 client_new: fullscreen=False
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L318 client_new: maximized=False
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L319 client_new: minimized=False
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L328 client_new: transient=False
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L342 client_new: square_match=True
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L344 client_new: calling float_center(client, 800, 600)
2026-02-25 14:38:36,514 WARNING libqtile config.py:set_floating_windows():L347 client_new: float_center DONE (800x600)
```

To sum up:  the logs show that the rule matches and float_center() is executed successfully, so the logic is working.
However, it runs too early in client_new, before the window is fully managed and set to floating, so the resize gets ignored.

richcarni:
> No I use `lazy.window.toscreen()`.

Can confirm this wasn't working - but I believe it is fixed now

As for the other issue with your client_new hook, can you test it again with latest updates? I have reverted some changes that were probably overwriting your float sizing

tpaniaki:
Thanks here is what I get:

- Windows in the Max layout still appear as floating. 
- Scratchpads maximization still not work (I think I saw a commit about that that's why I tested).
- Fullscreen to_screen() is solved.
- Regarding the floating, I still can't resized. I run those hooks:

```
@hook.subscribe.client_new
def log_window_name_timer(client): Timer(0.2, lambda: logger.warning(f"client.name_timer: {client.name} | floating={client.floating}")).start()

@hook.subscribe.client_new
def log_window_name_immediate(client): logger.warning(f"client.name: {client.name} | floating={client.floating}")
```

In the current git, I get:

```
2026-02-26 12:45:33,429 WARNING libqtile config.py:log_window_name_immediate():L1137 client.name: Rédaction du message | floating=True
2026-02-26 12:45:33,629 WARNING libqtile config.py:<lambda>():L1129 client.name_timer: Rédaction : (pas de sujet) - Thunderbird | floating=True

2026-02-26 12:46:26,305 WARNING libqtile config.py:log_window_name_immediate():L1137 client.name: foot | floating=False
2026-02-26 12:46:26,505 WARNING libqtile config.py:<lambda>():L1129 client.name_timer: foot | floating=True
```

In your branch:

```
2026-02-26 12:42:30,029 WARNING libqtile config.py:log_window_name_immediate():L1137 client.name: Rédaction du message | floating=False
2026-02-26 12:42:30,230 WARNING libqtile config.py:<lambda>():L1129 client.name_timer: Rédaction : (pas de sujet) - Thunderbird | floating=False

2026-02-26 12:42:56,562 WARNING libqtile config.py:log_window_name_immediate():L1137 client.name: foot | floating=False
2026-02-26 12:42:56,762 WARNING libqtile config.py:<lambda>():L1129 client.name_timer: foot | floating=True
```

Here foot is a scratchpad and is in both cases properly resized; Thunderbird is the window you get from "New message" and never switch to floating. I should have mentioned the whole block in my config:

```
def float_center(client, width, height, borderwidth=5, bordercolor="000000"):
    screen = (
        getattr(client, "screen", None)
        or getattr(getattr(client, "group", None), "screen", None)
        or qtile.current_screen
    )
    if not screen:
        return

    x = screen.dx + (screen.dwidth - width) // 2
    y = screen.dy + (screen.dheight - height) // 2

    client.floating = True
    client.place(x=x, y=y, width=width, height=height, borderwidth=borderwidth, bordercolor=bordercolor)
    client.bring_to_front()


@hook.subscribe.client_new
def set_floating_windows(client):
    if client.is_transient_for():
        return

    if any(rule.compare(client) for rule in floating_windows_square):
        float_center(client, 800, 600)
        return

    if any(rule.compare(client) for rule in floating_windows_rectangular):
        float_center(client, 900, 950)
        return
```

The culprit probably lies somewhere here:

```
    x = screen.dx + (screen.dwidth - width) // 2
    y = screen.dy + (screen.dheight - height) // 2

    client.floating = True
    client.place(x=x, y=y, width=width, height=height, borderwidth=borderwidth, bordercolor=bordercolor)
    client.bring_to_front()
```

richcarni:
> * Regarding the floating, I still can't resized. I run those hooks:

Yep, same here. Figured out what it is, missing this bit from the floating setter:

```
else:
    # if we are setting floating early, e.g. from a hook, we don't have a screen yet
    self._float_state = FloatStates.FLOATING
```

I suppose I need to put it back in for compatibility or maybe it is needed for something else I'm missing

While I'm thinking about that, if you change your hook to `@hook.subscribe.client_managed`, I think you'll find it works as expected

jwijenbergh:
This is great, thanks for picking this up. Note that I had a lot of ideas for this that I never put into writing. So here we go:

- all settings related to floating, max, fullscreen should be in their layout. E.g. `floats_kept_above`, `auto_fullscreen` etc. I made a beginning with the border widths for max and fullscreen. `auto_fullscreen` should just be similar to `float_rules` but now called `rules` in the fullscreen layout. 
- We have to know how to deal with the following scenario: Cycle through layouts, from tiling to floating to tiling. In the scenario where we are in the layout 'floating' we thus have two active floating layouts (the floating layer and the floating layout). I added some very complex code for dealing with this, not sure if that is still in this branch. E.g. what should toggle_tiling do when the floating layout is active? Maybe `toggle_tiling` or the `tiling` layer should be renamed to the `layout` layer
-  All these issues should be fixed: https://github.com/qtile/qtile/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22floating%20refactor%22

Now that I think about it, the best idea, but involves a lot of changes, IMO:

The qtile core defines 'no' layers, this is all defined by the config. The default config could be:
```
layers = [
.. minimized layout(rules=...)
.. tiling layouts(rules=all windows by default)
.. floating layout(rules=...)
.. maximized layout(rules=...)
.. fullscreen layout(rules=...)
]
```
using these layers, qtile dynamically creates functions for toggling these layers and creates e.g. for wayland these layers with z-layering in the scene tree. Then let's say a user wants qtile without maximized, floating, tiling or minimized (KIOSK style) then it could define:
```
layers = [
... fullscreen layout
]
```
Now you have made an alternative to cage in qtile

Let's look at another use case, a floating window manager/compositor. Just define:
```
layers = [
.. floating layout
]
```
Now that's something similar to tinywl and other simple WMs

This `layers` variable is ordered by z-layer. So qtile actually becomes hackable, flexible and the base would be KISS. E.g. see what `group.py` does, it's almost not related to groups anymore but just handles all the layers in a messy base

richcarni:
Hey @jwijenbergh, great to hear from you!

> * We have to know how to deal with the following scenario: Cycle through layouts, from tiling to floating to tiling. In the scenario where we are in the layout 'floating' we thus have two active floating layouts (the floating layer and the floating layout). I added some very complex code for dealing with this, not sure if that is still in this branch.

I believe it is. For example floating windows will be managed in the floating_layout and if the primary layout is floating it also manages the floating_layout, so unlike previously you don't have (for lack of better wording) tiling floating windows and floating floating windows. And there is a win_state_follows flag to differentiate between a window that belongs in the primary layout and one that has been moved to a another layout, so you can move back and forth without breaking your layout

I have to admit, the first few reads I was struggling to understand a lot of your ideas. Eventually it dawned on me that what I've been thinking of as tiling/main/primary vs other/alt/specialised/secondary layouts, you've started thinking of as layers. I think that actually makes a lot of sense if, like you suggested, it corresponds to z-layering. I don't know what to do about it but yeah, the language can definitely be confusing

If we were to rename tiling to layout layer, would that mean we have the following? 
> layers = [
> .. minimized layout(rules=...)
> .. layout layouts(rules=all windows by default)

Sorry, couldn't resist! 😭

> * all settings related to floating, max, fullscreen should be in their layout. E.g. `floats_kept_above`, `auto_fullscreen` etc. I made a beginning with the border widths for max and fullscreen. `auto_fullscreen` should just be similar to `float_rules` but now called `rules` in the fullscreen layout.

Makes sense for things like border width. Not sure about floats_kept_above - maybe we can keep a global option and add per layout overrides. And I do like the idea of having rules for each, um, layer

I think there is one semi-related issue/shortcoming with the PR in it's current state, which is you can currently only specify options for one type of each layout. For example, if you have a floating layout (in tiled layouts), those options: border width, color, etc will be applied to all floats. And you can't have 2 floating layouts with different options, they will both share the options of the first defined in config. Similar applies to max and fullscreen layouts. I'm not sure how big of an issue this is, do users really have differently styled versions of the same layout in their layouts?

I also have some thoughts about hiding/suppressing windows when we move them from a tiled layout to one of the speciality (float, max, fullscreen) layouts. Currently we just blur them for max and fullscreen, so layout order is maintained when we un-max/fullscreen, and this works reasonably well because they cover the tiled layout. For floats we have to remove them and then they can end up out of order when returned. Was thinking of adding an optional window flag like hide_in_layout, which a layout can use to skip this window when composing its layout. By making it optional, existing/custom layouts would not break and we could go about upgrading the built-in layouts to support this functionality. But I'm thinking of leaving it to a separate PR 

Overall, my main concern is the more we add to this PR, the harder it is going to be to get it merged (especially if they are breaking changes or required migrations). But this might just be me, as the new guy, still not having a good handle on what is required to make these kinds of changes

tych0:
Some of these commits just seem like bug fixes, so you could probably split those off and land them independently. 

I haven't read the whole thing, but I know people don't like the way our floating support works, so if this fixes a lot of those complaints, I think we should try to land it, regardless of how big it is. 

In terms of API breakage, as long as we can fix up the configs via migration I think it's fine. If there are behavior changes, those are the bug fixes people want...

richcarni:
> Some of these commits just seem like bug fixes, so you could probably split those off and land them independently.

Unfortunately I think they are just fixing bugs introduced in the initial effort to get this code working. A lot of these (all?) can be squashed before merging - can deal with that later

> I haven't read the whole thing, but I know people don't like the way our floating support works, so if this fixes a lot of those complaints, I think we should try to land it, regardless of how big it is.

Cool! I'm going to try and verify that it fixes all the complaints it claims to. And the code's pretty rough in places - I need to go over it myself and tidy it up before asking you or anyone else to review it. Taking a mini break from it at the moment - once again underestimated how much work it can be to get something passing all tests

> In terms of API breakage, as long as we can fix up the configs via migration I think it's fine. If there are behavior changes, those are the bug fixes people want...

I think API breakage will be minimal, so the migration should be straightforward

PR #5850, opened 2026-02-25
Fix up our screen matching API

tych0:
Since our last release, we've added a couple of patches allowing people to match the serial and output name (or so I thought :) ). However, the `.name` member didn't mean the same thing across x11 and wayland.

Since the API is not yet released, we can still change it. Here is my attempt to make it better: instead of adding a complex matching API, we can just add a function that users can define in order to do whatever complex matching they want, returning the list of Screens at the end. In some future world, we could even get rid of the reconfigure_screens hook, and just invoke this function again.

I think this is a release blocker, since we don't want to release with an API that we don't like.

Two tests in this PR fail on wayland, and I have not had time to look at them. But I just want to put this up now in hopes that someone can, since we only have a month left before we violate our "one release per quarter" rule.

### DIFF
--- a/libqtile/backend/base/core.py
+++ b/libqtile/backend/base/core.py
@@ -3,11 +3,10 @@ from __future__ import annotations
 import contextlib
 import typing
 from abc import ABCMeta, abstractmethod
-from dataclasses import dataclass
 
 from libqtile import hook
 from libqtile.command.base import CommandObject, expose_command
-from libqtile.config import Screen, ScreenRect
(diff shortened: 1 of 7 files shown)


tych0:
I'm gonna land these two commits for now. This still does not have `generate_screens()`, but I cannot for the life of me get claude to debug why that commit was broken ;). These two are good to go and fix up the semantic mismatch between x11 and wayland, though.

PR #5854, opened 2026-02-25
fix: add missing imports to config_pyi_header in confreader

### LINKED ISSUE #5853: qtile check: stub type checking fails due to missing imports in config_pyi_header
## Bug Description

Running `qtile check` fails the stub type checking step with:

```
error: not checking stubs due to mypy build errors:
config.pyi:13: error: Name "FunctionType" is not defined  [name-defined]
config.pyi:30: error: Name "IdleInhibitor" is not defined  [name-defined]
```

## Root Cause

In `libqtile/confreader.py`, the `config_pyi_header` string (used to generate a temporary `.pyi` stub) is missing two imports that the `Config` class annotations reference:

- `FunctionType` — used by `focus_on_window_activation`
- `IdleInhibitor` — used by `idle_inhibitors`

The `TYPE_CHECKING` block at the top of `confreader.py` correctly imports both, but the generated stub header was not updated to match.

## Fix

```python
config_pyi_header = """
from typing import Any
from typing import Literal
from types import FunctionType                                               # add this
from libqtile.config import Group, IdleInhibitor, Key, Mouse, Rule, Screen  # add IdleInhibitor
from libqtile.layout.base import Layout
"""
```

## Environment

- Qtile version: 0.34.1
- Python version: 3.14.3
- OS: CachyOS Linux (kernel 6.19.3)

idunkoncode:
## Summary

- `FunctionType` and `IdleInhibitor` are referenced in `Config.__annotations__` but were absent from the `config_pyi_header` string used to generate the temporary `.pyi` stub during `qtile check`
- This caused mypy to bail out with build errors before stub checking could run

## Test plan

- [ ] Run `qtile check` on a config — no more `Name "FunctionType" is not defined` or `Name "IdleInhibitor" is not defined` errors
- [ ] Stub type checking proceeds normally

Fixes #5853

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -20,7 +20,8 @@ class ConfigError(Exception):
 config_pyi_header = """
 from typing import Any
 from typing import Literal
-from libqtile.config import Group, Key, Mouse, Rule, Screen
+from types import FunctionType
+from libqtile.config import Group, IdleInhibitor, Key, Mouse, Rule, Screen
 from libqtile.layout.base import Layout
 
 """


tych0:
Hmm. Can you see why our existing qtile check tests didn't catch this?

ervinpopescu:
BTW I noticed this as well when using `qtile check`.

The reason the existing `qtile check` tests failed to catch the missing imports is due to a "silent failure" in the command's exit logic and the test's reliance on exit codes.

  1. `qtile check` returns an exit code of 0 on type errors
  In the `check_config` implementation, when type checking fails (either via `stubtest` or `mypy`), the valid flag is set to False. However, the function simply prints a warning message and continues to the end of the function without
  calling `sys.exit()` or raising an unhandled exception. As a result, the process exits successfully (0) even when it detects errors.

https://github.com/qtile/qtile/blob/2211ae67be428284ba41acd855eb769281ca591c/libqtile/scripts/check.py#L123-L145

  2. Tests only fail on non-zero exit codes
  The run_qtile_check helper in the test suite uses `subprocess.check_call()`. This function only raises a `CalledProcessError` (which the test catches to return `False`) if the process returns a non-zero exit code. Because of the behavior described in Point 1, the tests perceive the type-check failure as a success.

https://github.com/qtile/qtile/blob/2211ae67be428284ba41acd855eb769281ca591c/test/test_check.py#L26-L36

  3. Incomplete Fix in PR #5854
  While the PR correctly identifies that `FunctionType` and `IdleInhibitor` were missing from the generated stub header, it still omits `IdleTimer`. This means that even with this PR applied, qtile check will still report type errors for the default configuration because IdleTimer is used in the Config annotations but not imported in the stub.

https://github.com/qtile/qtile/blob/2211ae67be428284ba41acd855eb769281ca591c/libqtile/confreader.py#L50-L60

libqtile/confreader.py
```diff
@@ -20,7 +20,8 @@ class ConfigError(Exception):
 config_pyi_header = """
 from typing import Any
 from typing import Literal
-from libqtile.config import Group, Key, Mouse, Rule, Screen
+from types import FunctionType
+from libqtile.config import Group, IdleInhibitor, Key, Mouse, Rule, Screen
```
ervinpopescu, line 24:
```suggestion
from libqtile.config import Group, IdleInhibitor, IdleTimer, Key, Mouse, Rule, Screen
```

ervinpopescu: sorry for the commit spam :_)

tych0: IdleTimer without the space, right? Can you send a fixed version to your branch?

ervinpopescu: Uhm, yeah, oopsie. My branch should already have it, if you're talking to me.

tych0:
I believe this was fixed by #5859, thanks for the report!

PR #5860, opened 2026-03-04
nix: update nixpkgs, fix deprecations, and align qtile deps with upstream

Gurjaka:
- Update nixpkgs input
- Replace deprecated `nixfmt-rfc-style` with `pkgs.nixfmt`
- Replace deprecated `xorg.xcbutilcursor` with `libxcb-cursor`
- Add missing runtime dependency `python3Packages.cairocffi` (fixes: `ERROR Missing dependencies: cairocffi[xcb]>=1.6.0`)

Qtile packaging:
- Drop manual filtering of pywlroots, pywayland, and xkbcommon (pywlroots removed from nixpkgs; deps no longer propagated upstream)
- Simplify propagatedBuildInputs to match current qtile derivation

Also clean up minor formatting in nix expressions.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -87,7 +87,7 @@
       });
     in
     {
-      formatter = forAllSystems (pkgs: pkgs.nixfmt-rfc-style);
+      formatter = forAllSystems (pkgs: pkgs.nixfmt-tree);
 
       checks = forAllSystems (pkgs: pkgs.python3Packages.qtile.passthru.tests);
 
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1764517877,
-        "narHash": "sha256-pp3uT4hHijIC8JUK5MEqeAWmParJrgBVzHLNfJDZxg4=",
+        "lastModified": 1772542754,
+        "narHash": "sha256-WGV2hy+VIeQsYXpsLjdr4GvHv5eECMISX1zKLTedhdg=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "2d293cbfa5a793b4c50d17c05ef9e385b90edf6c",
+        "rev": "8c809a146a140c5c8806f13399592dbcb1bb5dc4",
         "type": "github"
       },
       "original": {
--- a/nix/build-config.nix
+++ b/nix/build-config.nix
@@ -19,7 +19,7 @@ let
     }
     {
       varname = "XCBCURSOR";
-      pkg = pkgs.xorg.xcbutilcursor;
+      pkg = pkgs.libxcb-cursor;
       libname = "libxcb-cursor.so";
     }
   ];
--- a/nix/qtile.nix
+++ b/nix/qtile.nix
@@ -22,15 +22,6 @@ let
     in
     "${symver}+${flakever}.flake";
 
-  removeOldDeps =
-    dep:
-    !(pkgs.lib.hasAttr "pname" dep)
-    || (
-      dep.pname != pkgs.python3Packages.pywlroots.pname
-      && dep.pname != pkgs.python3Packages.pywayland.pname
-      && dep.pname != pkgs.python3Packages.xkbcommon.pname
-    );
-
   qtile-override-func =
     qtile-prev:
     {
@@ -42,16 +33,6 @@ let
     }
     // {
       env = build-config.resolved-env-vars;
-
-      propagatedBuildInputs =
-        (with pkgs; [
-          wayland-scanner
-          wayland-protocols
-          python3Packages.cffi
-          python3Packages.xcffib
-          python3Packages.aiohttp
-        ])
-        ++ (lib.filter removeOldDeps qtile-prev.propagatedBuildInputs);
 
       pypaBuildFlags = [ "--config-setting=backend=wayland" ] ++ build-config.resolved-config-settings;
     }


Gurjaka:
`Nix flake check` is failing due to formatting changes in `nix fmt`. Just mentioning that nothing is actually wrong. https://github.com/NixOS/nixfmt/releases/tag/v1.2.0

tych0:
> Nix flake check is failing due to formatting changes in nix fmt. Just mentioning that nothing is actually wrong.

Hmm, can we bump the version we use in CI so that the formatting matches this patch? The rest of it looks reasonable to me.

flake.nix
```diff
@@ -92,12 +82,12 @@
...
           common-python-deps
           common-system-deps
-          (builtins.attrValues scripts)
+          (builtins.attrValues tests)
         ];
       });
     in
     {
-      formatter = forAllSystems (pkgs: pkgs.nixfmt-rfc-style);
+      formatter = forAllSystems (pkgs: pkgs.nixfmt);
```
Sigmanificient: You should use `nixfmt-tree`, not `nixfmt` directly

> *Passing directories or non-Nix files (such as ".") is deprecated and will be unsupported soon. Please use the `pkgs.nixfmt-tree` wrapper instead, or https://github.com/numtide/treefmt-nix for more flexibility*

Sigmanificient (changes requested):

Gurjaka:
> > Nix flake check is failing due to formatting changes in nix fmt. Just mentioning that nothing is actually wrong.
> 
> Hmm, can we bump the version we use in CI so that the formatting matches this patch? The rest of it looks reasonable to me.

CI uses version from flake that's in repo. So after merging this PR it will use latest version

Edit: weird enough CI passes after using `nixfmt-tree` instead

tych0:
> CI uses version from flake that's in repo. So after merging this PR it will use latest version

hmm, it shouldn't. But I don't claim to understand how all this Nix stuff fits together. Green is good enough for me.

Gurjaka:
> > CI uses version from flake that's in repo. So after merging this PR it will use latest version
> 
> hmm, it shouldn't. But I don't claim to understand how all this Nix stuff fits together. Green is good enough for me.

I should've been clearer, the CI doesn't actually pull 'the latest' from the web. It’s strictly locked to the commit hash in flake.lock (which is updated manually by us). When I mentioned 'latest' before, I meant the version of `nixpkgs-unstable` I've manually capture in this PR. 

Think of it like package-lock.json or Cargo.lock. We pin our dependencies to a specific, tested version of nixpkgs. This ensures that if the build is green here, it stays green everywhere, and we don't get hit by breaking changes in Qtile or its dependencies unless we explicitly choose to update that lockfile.

tych0:
> Think of it like package-lock.json or Cargo.lock. We pin our dependencies to a specific, tested version of nixpkgs. This ensures that if the build is green here, it stays green everywhere, and we don't get hit by breaking changes in Qtile or its dependencies unless we explicitly choose to update that lockfile.

Sure, but then why did it fail? Shouldn't the CI take your branch, merge it into master, and then run? (i.e. wouldn't the result of that have your flake.lock updates?)

Maybe I'm about to learn something :)

Gurjaka:
> > Think of it like package-lock.json or Cargo.lock. We pin our dependencies to a specific, tested version of nixpkgs. This ensures that if the build is green here, it stays green everywhere, and we don't get hit by breaking changes in Qtile or its dependencies unless we explicitly choose to update that lockfile.
> 
> Sure, but then why did it fail? Shouldn't the CI take your branch, merge it into master, and then run? (i.e. wouldn't the result of that have your flake.lock updates?)
> 
> Maybe I'm about to learn something :)

I'm afraid I don't have an answer to this 😅
Though I am pretty sure I ran `nix fmt` locally :)

tych0:
Well, and it passed git master CI, so it's clearly "right" by some version of `flake check`. But I really don't understand why it failed here.

Gurjaka:
> Well, and it passed git master CI, so it's clearly "right" by some version of `flake check`. But I really don't understand why it failed here.

Perhaps because previously I used `pkgs.nixfmt` which no longer supports tree formatting as @Sigmanificient mentioned. I then changed to `pkgs.nixfmt-tree` and CI succeeded, so maybe my assumption that nothing was wrong with formatter itself was wrong?

tych0:
Yeah, I guess it must be that, since that's the only change really.

Gurjaka:
> Yeah, I guess it must be that, since that's the only change really.

Most importantly CI succeeded, fix is merged.

Sigmanificient:
Nixfmt does indeed does not support file argument any longer, which caused an issue with `nix fmt .` invocation in the CI

```
nixfmt: ./.direnv/flake-inputs/ln4j1iqnnzs2ynx2cr88bdh65fmds2aq-source/pkgs/top-level: openTempFileWithDefaultPermissions: permission denied (Read-only file system)
HasCallStack backtrace:
  collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
  toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
  throwIO, called at libraries/ghc-internal/src/GHC/Internal/Control/Exception/Base.hs:195:43 in ghc-internal:GHC.Internal.Control.Exception.Base
```

PR #5862, opened 2026-03-07
Add missing widget dependencies to qtile check

elParaguayo:
When widgets are missing a dependency, the widget is replaced with an `ImportError` mesage in the bar. However, the actual missing dependency is only shown in the log file. In addition, if there are multiple missing dependencies, the log file will only show the first.

This PR adds an additional check to `qtile check`. It reads the config and, where there is an ImportError widget in the bar, it parses the relevant widget's module and checks all import statements, displaying a list of all modules that it cannot import.

As an example, if I edit groupbox.py:
```python
import itertools
from functools import partial
from typing import Any

from libqtile import hook


import made_up_module
import this_doesnt_exist
import nope


class _GroupBase(base._TextBox, base.PaddingMixin, base.MarginMixin):
...
```

Running `qtile check` will output:
```
Checking Qtile config at: config/missing_deps.py
Checking if config is valid python...
Checking widget dependencies...
The following widgets have import errors:
 GroupBox: made_up_module, nope, this_doesnt_exist
Type checking config file...
```

### DIFF
--- a/libqtile/scripts/check.py
+++ b/libqtile/scripts/check.py
@@ -7,6 +7,7 @@ from os import environ, path
 
 from libqtile import confreader
 from libqtile.utils import get_config_file
+from libqtile.widget.import_error import ImportErrorWidget
 
 
 class CheckError(Exception):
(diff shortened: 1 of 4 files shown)


elParaguayo:
Actually, I wonder if it would be better to do the dependency check inside the ImportError widget, that way we can do a cleaner log message when the config is loaded and then just query the widget during `qtile check`.

tych0:
Either approach works for me, but it could use a test. Thanks for doing this!

elParaguayo:
I'm doing the second approach and will add a test for it too.

tych0:
I squashed the move into the first commit, lgtm, thanks!

PR #5863, opened 2026-03-08
wayland: Fix floating kept above behaviour

richcarni:
The idea is to have sensible defaults with the ability to easily
customize other behaviours

`floats_kept_above` moves floating windows to the KEEPABOVE layer
Fixes floating windows going behind tiled windows when
`floats_kept_above = True`

~~`bring_front_click` moves clicked window to the top of its layer~~

~~`bring_to_front()` moves window to the BRINGTOFRONT layer
When a window returns from BRINGTOFRONT, it returns to its natural/base
layer~~

Added `client_focus_by_click` hook, which can be used for custom behaviour like:
```python
follow_mouse_focus = True 
bring_front_click = False
floats_kept_above = True

@hook.subscribe.client_focus_by_click
def client_focus_by_click(client):
    client.move_to_top() 
```

TODO:

- [x] Add tests (wayland) verifying correct movement between layers

### DIFF
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -423,10 +423,10 @@ class Core(base.Core):
     def handle_cursor_button(self, button: int, mask: int, pressed: bool, x: int, y: int) -> bool:
         assert self.qtile is not None
         if pressed:
-            if not self.qw_cursor.implicit_grab.live:
-                self._focus_by_click()
-
             handled = self.qtile.process_button_click(int(button), int(mask), x, y)
+
+            if not handled and not self.qw_cursor.implicit_grab.live:
+                self._focus_by_click()
 
             if isinstance(self.qtile.hovered_window, Internal):
                 self.qtile.hovered_window.process_button_click(
@@ -501,14 +501,28 @@ class Core(base.Core):
         # TODO logic imcomplete
         win._ptr.focus(win._ptr, False)  # What is the second argument?
 
+    def _grab_click_on_current_window(self) -> None:
+        """Grab button events on the current window.
+
+        Called before switching screens to ensure clicks on the now-unfocused
+        window will be intercepted to refocus it.
+        """
+        win = self.qtile.current_window
+        if win:
+            # In wayland backend, current_window is always an wayland Window
+            assert isinstance(win, Window)
+            win._grab_click()
+
     def _focus_by_click(self) -> ffi.CData:
         assert self.qtile is not None
         view = self.qw_cursor.view
 
         if view != ffi.NULL:
             win = self.qtile.windows_map.get(view.wid)
+            if win is None:
+                return
 
-            if win is not None and self.qtile.config.bring_front_click is True:
+            if self.qtile.config.bring_front_click is True:
                 win.bring_to_front()
             elif self.qtile.config.bring_front_click == "floating_only":
                 if isinstance(win, base.Window) and win.floating:
@@ -518,16 +532,17 @@ class Core(base.Core):
                 if win.screen is not self.qtile.current_screen:
                     self.qtile.focus_screen(win.screen.index, warp=False)
                 win.focus(False)
-            elif isinstance(win, base.Window):
+            elif isinstance(win, Window):
                 if win.group and win.group.screen is not self.qtile.current_screen:
                     self.qtile.focus_screen(win.group.screen.index, warp=False)
                 self.qtile.current_group.focus(win, False)
-
+                win._ungrab_click()
         else:
             screen = self.qtile.find_screen(
                 int(self.qw_cursor.cursor.x), int(self.qw_cursor.cursor.y)
             )
             if screen:
+                self._grab_click_on_current_window()
                 self.qtile.focus_screen(screen.index, warp=False)
 
         return view
--- a/libqtile/backend/wayland/core.py
+++ b/libqtile/backend/wayland/core.py
@@ -522,6 +522,8 @@ class Core(base.Core):
             if win is None:
                 return
 
+            hook.fire("client_focus_by_click", win)
+
             if self.qtile.config.bring_front_click is True:
                 win.bring_to_front()
             elif self.qtile.config.bring_front_click == "floating_only":
@@ -907,6 +909,9 @@ class Core(base.Core):
     @expose_command
     def idle_notify_activity(self) -> None:
         lib.qw_server_idle_notify_activity(self.qw)
+
+    def fake_click(self) -> None:
+        lib.qw_cursor_fake_click(self.qw_cursor)
 
 
 class Painter:
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -46,17 +46,17 @@ class Base(base._Window):
         self.core: Core = typing.cast("Core", qtile.core)
 
     def reparent(self, layer: int) -> None:
-        if self.layer == layer:
+        if self.layer() == layer:
             return
         lib.qw_view_reparent(self._ptr, layer)
 
-    @property
+    @expose_command()
     def layer(self) -> int:
-        return self._ptr.layer
+        return lib.qw_view_get_layer(self._ptr)
 
     @expose_command()
     def keep_above(self, enable: bool | None = None) -> None:
-        is_enabled = self.layer == lib.LAYER_KEEPABOVE
+        is_enabled = self.layer() == lib.LAYER_KEEPABOVE
         if enable is None:
             enable = not is_enabled
 
@@ -67,7 +67,7 @@ class Base(base._Window):
 
     @expose_command()
     def keep_below(self, enable: bool | None = None) -> None:
-        is_enabled = self.layer == lib.LAYER_KEEPBELOW
+        is_enabled = self.layer() == lib.LAYER_KEEPBELOW
         if enable is None:
             enable = not is_enabled
         if enable:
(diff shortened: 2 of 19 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -507,12 +499,13 @@ def _focus_by_click(self) -> ffi.CData:
...
             win = self.qtile.windows_map.get(view.wid)
+            hook.fire("client_mouse_click", win)
 
             if win is not None and self.qtile.config.bring_front_click is True:
-                win.bring_to_front()
+                win.move_to_top()
             elif self.qtile.config.bring_front_click == "floating_only":
                 if isinstance(win, base.Window) and win.floating:
-                    win.bring_to_front()
+                    win.move_to_top()
```
richcarni: If we go with this change, probably should consider renaming `bring_front_click` to something like `move_to_top_click`, or `raise_on_click`

test/test_hook.py
```diff
@@ -559,6 +559,20 @@ class ClientMouseEnterConfig(BareConfig):
...
+def test_client_mouse_click(manager_nospawn):
+    class ClientMouseClickConfig(BareConfig):
+        test = CallWindow()
+        hook.subscribe.client_mouse_click(test)
+        layouts = [layout.columns.Columns()]
+
+    manager_nospawn.start(ClientMouseClickConfig)
+    manager_nospawn.test_window("Test Client")
+    manager_nospawn.test_window("one")
+    manager_nospawn.backend.fake_click(0, 0)
```
richcarni, line 578: Hmm, I needed to add a second window for x11 since it doesn't appear to call focus_by_click() on an already focused window

wayland calls it on every click, which isn't ideal

Sinaaaa:
This seems to work quite well for me on the default config. One bug I've noticed is that often `Click([mod], "Button2", lazy.window.bring_to_front()),` ignores the first click, so need to use it twice for the desired effect.

richcarni:
> This seems to work quite well for me on the default config. One bug I've noticed is that often `Click([mod], "Button2", lazy.window.bring_to_front()),` ignores the first click, so need to use it twice for the desired effect.

Thanks for testing! Good to hear it is almost problem free for you

I'm not seeing that behaviour with mod + button 2 click. Are you using `qtile cmd-obj -o core -f stacking_info` to check whether it is changing layers?

Also bring_to_front() is now a toggle behaviour, in theory if you click twice, it should go to front and then come back again

Sinaaaa:
> > Are you using `qtile cmd-obj -o core -f stacking_info` to check whether it is changing layers?

🤣 I'll try to take a look later..

Sinaaaa:
[stacking1.txt](https://github.com/user-attachments/files/26123034/stacking1.txt)
[stacking2.txt](https://github.com/user-attachments/files/26123035/stacking2.txt)


ok so here stackin1.txt is the base state & stacking2 is after the modkey+middle mouse button fail attempt state.

I asked an LLM what the difference is between these two files & I will just shamelessly paste the output here:

The primary changes between State 1 and State 2 are as follows:
Layer Migration

    Window wid: 7: Moved from LAYER_BRINGTOFRONT to LAYER_KEEPABOVE.

Layer Occupancy

    LAYER_KEEPABOVE: Changed from empty to containing one window (wid: 7).

    LAYER_BRINGTOFRONT: Changed from containing two windows (wid: 7 and wid: 6) to containing only one window (wid: 6).

Drag Icon Coordinates

    X-Coordinate: Shifted from 2852 to 2954.

    Y-Coordinate: Shifted from 368 to 112.

Sinaaaa:
[stacking3.txt](https://github.com/user-attachments/files/26123130/stacking3.txt)
[stacking4.txt](https://github.com/user-attachments/files/26123131/stacking4.txt)

Here, stacking3 is the failed to elicit a change via modkey+middle button state & stacking4 is the state after a repeated keybind, a successful activation.

LLM summary:



The differences between State 3 and State 4 are as follows:
Layer Migration

    Window wid: 6: Moved from LAYER_BRINGTOFRONT to LAYER_LAYOUT.

    Window wid: 7: Moved from LAYER_KEEPABOVE to LAYER_BRINGTOFRONT.

Window Geometry Changes

    Window wid: 6: Coordinates shifted from x: 27, y: 24 to x: 504, y: 197.

    Window wid: 7: Coordinates shifted from x: 504, y: 197 to x: 27, y: 24.

Layer Occupancy

    LAYER_KEEPABOVE: Changed from containing one window (wid: 7) to being empty.

    LAYER_LAYOUT: Increased from three windows to four, now including wid: 6.

Drag Icon Coordinates

    X-Coordinate: Shifted from 2954 to 2852.

    Y-Coordinate: Shifted from 112 to 368.

richcarni:
Just looking at stacking1/stacking2...

In stack1.txt, the window you are mod + button2 clicking (according to LLM) is already in BRINGTOFRONT layer

```
        {
          "name": "LAYER_BRINGTOFRONT",
          "enabled": true,
          "x": 0,
          "y": 0,
          "type": "tree",
          "wid": null,
          "children": [
            {
              "name": "REDACTED-pc:~",
              "enabled": true,
              "x": 504,
              "y": 197,
              "type": "tree",
              "wid": 7,
              "children": [
 ```
 
So when you click it it toggles it back to it's base layer (here KEEPABOVE layer)

Then when you click it again it moves back to BRINGTOFRONT

Possible remedies, depending on what you want to achieve:
- Modify mouse in config with: `Click([mod], "Button2", lazy.window.move_to_top())
- Modify mouse in config with: `Click([mod], "Button2", lazy.window.bring_to_front(enable=True))
- We could modify `bring_to_front()` to instead move_to_top like the old backend, but then there is no way to use the BRINGTOFRONT layer at all

**Comment by richcarni** [comment]
Also, if you want an easier way of visualising the layers, you could try out my scene graph utility: https://github.com/richcarni/qtile-scene-graph-viewer/tree/main

(requires the latest rebase of this branch)

image

**Comment by Sinaaaa** [comment]
Ok so basically your point is that nothing is wrong, it's just `bring_to_front()` is not behaving as I would expect now. 

Issues with this:

```
mouse = [
    Drag([mod], "Button1", lazy.window.set_position_floating(), start=lazy.window.get_position()),
    Drag([mod], "Button3", lazy.window.set_size_floating(), start=lazy.window.get_size()),
    Click([mod], "Button2", lazy.window.bring_to_front()),
]
```
This is what everyone has in the default config when they start their qtile journey & I'd imagine many still have it years later unmodified. Changing this now could cause some distress, I don't know. Certainly having a keybind like that is not great for mod+ mouse2.

I don't know how to solve this, like can migration deal with something like this?  Like just change that line in everyone's config. Though it wouldn't be enough, because if you remove the entire `mouse =` block qtile will still default to this behavior. 

Alternatively reacting to,
>We could modify bring_to_front() to instead move_to_top like the old backend, but then there is no way to use the BRINGTOFRONT layer at all

you could rename the layers so that the bring to front of old could become move to top for everyone. Assuming move to top works every time on the first click as I would expect it does, I have not tested this yet.

(or maybe rename the new bring to front to something else & just give users a deprecation warning about it & for a while let it be move to top, I don't even know)

**Comment by richcarni** [comment]
> This is what everyone has in the default config when they start their qtile journey & I'd imagine many still have it years later unmodified. Changing this now could cause some distress, I don't know. Certainly having a keybind like that is not great for mod+ mouse2.
> 
> I don't know how to solve this, like can migration deal with something like this? Like just change that line in everyone's config.

Haha, I've never used that keybinding. I wasn't even aware of it until this conversation even though it's in my config!

But yes, this would be a pretty significant change and would need to be coordinated with an equivalent change in the x11 backend. And yes, I imagine it would require a migration

> Though it wouldn't be enough, because if you remove the entire `mouse =` block qtile will still default to this behavior.

We'd have to also update the default behaviour

TBH, I'm tempted to drop any changes to bring_to_front altogether. With the new hook, we can make click to activate do whatever we want (that reminds me I need to rename the hook to something like click "activate_window_click")

@elParaguayo Do you have thoughts on bring_to_front? Just keep the same behaviour as X11? That was my original intention until it started becoming unclear to me exactly what actions should return the window from BRINGTOFRONT

**Comment by richcarni** [comment]
@Sinaaaa I've dropped the changes to bring_to_front in favour of customising with a hook, and I'm probably going to separate this into two PRs (for floats_kept_above and client_activate_by_click hook)

Would be interested to know if that hook allows what you were referring to about separating focus and activation?

Suggest using it like:
```python
follow_mouse_focus = True 
bring_front_click = False
floats_kept_above = True

@hook.subscribe.client_activate_by_click
def client_activate_by_click(client):
    client.move_to_top()
```

Then with floating windows, mouse over should focus windows but not raise them. If you click a window, then it is raised

Sinaaaa:
It's been a whole 5 minutes, but this seems pretty effin great so far.

richcarni:
Looks like I've broken clicking on the bar
... and fixed

tych0:
Any reason not to land this?

richcarni:
> Any reason not to land this?

I'm pretty happy with this now, after using it myself for the past week

It does introduce a subtle change to x11 behaviour - which is that a window only ungrabs click when clicked (before focus would also ungrab click). This way we can use follow_mouse_focus and then click a window to raise it. I can't see what this could break but that change/commit (the last one) probably warrants review

Still happy to split into 2 PRs if that is helpful (although I think it is sensibly broken up by its commits)

libqtile/backend/wayland/window.py
```diff
@@ -236,6 +242,15 @@ def place(
 
     @expose_command()
     def focus(self, warp: bool = True) -> None:
+        # re-grab button events on the previously focused window,
+        # but only un-grab them on focus by click
+        old = lib.qw_server_active_view(self.core.qw)
+        logger.warning(f"old: {old}")
```
elParaguayo: Leftover from debugging?

richcarni: Yep, thank you!

elParaguayo:
Anything stopping us merging this? Would be good to get this done and then workout what to do with the X11 stacking PR (@richcarni would love your thoughts on this).

richcarni:
> Anything stopping us merging this? Would be good to get this done and then workout what to do with the X11 stacking PR (@richcarni would love your thoughts on this).

Was just about the check for any objections before merging

I'm keen to try and unify the stacking behaviour!

PR #5864, opened 2026-03-09
wayland: add ext-data-control manager

wineee:
Create the ext_data_control_manager_v1 alongside the existing wlr_data_control_manager_v1. ext_data_control is the replacement for wlr_data_control: the interface is identical and only the protocol name differs; wlr_data_control is expected to be deprecated. Exposing both keeps compatibility with current clipboard clients while supporting the new protocol.


wineee:
Sway Implementation：https://github.com/swaywm/sway/pull/8597

tych0:
Thanks!

PR #5865, opened 2026-03-11
Reapply position crash

tych0:
Here's another (completely AI debugged, but it *looks* maybe feasible) attempt at fixing our strange CI crash in #5818. I also reapplied the fix on top so we'll hopefully see it again if Claude was wrong...

### DIFF
--- a/libqtile/backend/wayland/qw/server.c
+++ b/libqtile/backend/wayland/qw/server.c
@@ -240,12 +240,6 @@ static void qw_server_output_manager_reconfigure(struct qw_server *server,
             wlr_output_state_set_transform(&state, head->state.transform);
             wlr_output_state_set_scale(&state, head->state.scale);
             wlr_output_state_set_adaptive_sync_enabled(&state, head->state.adaptive_sync_enabled);
-            struct wlr_box box;
-            wlr_output_layout_get_box(server->output_layout, head->state.output, &box);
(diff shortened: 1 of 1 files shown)


libqtile/backend/wayland/core.py
```diff
@@ -319,6 +319,17 @@ def on_config_load(self, initial: bool) -> None:
...
+        # Set xcursor environment variables from Python before calling into C.
+        # This avoids calling setenv() from C code, which is not thread-safe with
+        # respect to getenv() calls that may happen concurrently (e.g. from
+        # fontconfig/pango initialization in a glib worker thread). See #5818.
+        os.environ["XCURSOR_SIZE"] = str(self.qtile.config.wl_xcursor_size)
+        theme = self.qtile.config.wl_xcursor_theme
+        if theme is not None:
+            os.environ["XCURSOR_THEME"] = theme
+        elif "XCURSOR_THEME" in os.environ:
+            del os.environ["XCURSOR_THEME"]
```
richcarni: I think I left this comment in a different version of this but should we be clearing an XCURSOR_THEME variable that the user may have set outside of qtile? Seems like a subtle change from the original code. Maybe an improvement, I'm not really sure?

tych0: Oh sorry, I missed your earlier comment. I agree, let's drop that part.

Issue #5844, opened 2026-02-15
[wayland] Runaway pointer when resizing windows across display boundaries is acting on & glitching with the wrong window.


Sinaaaa:
### Issue description

1. Have 2 displays logically side by side next to each other.
2. Start firefox on the left.
3. Start whatever on the right, it can be either floating or tiled, it doesn't matter
4. Start a speed test or a 4k youtube video to make the firefox window a bit fat
5. star resizing the firefox window on the left toward the right and go through the display boundary while grabbing & resize in a fast motion.

What often happens is that the runaway cursor will resize the other window there, this can happens in a one smooth motion.

Be warned the demonstration is mildly NSFW. 

https://github.com/user-attachments/assets/03314ff2-ceb7-4ac5-8bf9-25ef42d2c71b

### Version

0.34.2.dev66+g3505e0631

### Backend

Wayland (experimental)

### Config

```python
from libqtile import hook, layout, bar, widget
from libqtile.config import Screen, Group, Key
from libqtile import qtile
from libqtile.lazy import lazy 
from libqtile.backend.wayland import InputConfig

wl_xcursor_theme = "Adwaita"
wl_xcursor_size = 48

wl_input_rules = {
    "*": InputConfig(
        left_handed=False,
        pointer_accel=False,
        accel_profile="flat",
    ),

    "type:keyboard": InputConfig(
        kb_layout="hu",				
        kb_variant="102_qwerty_comma_nodead",
    ),
}

mod = "mod4"


keys = [
	Key(["mod1"], "Tab", lazy.spawn("rofi -show window")),
	Key([mod], "t", lazy.spawn("rofi -show drun")),
	Key([mod], "f", lazy.window.toggle_fullscreen(), desc="Toggle fullscreen"),
    Key([mod], "h", lazy.layout.left()),
    Key([mod], "l", lazy.layout.right()),
    Key([mod], "j", lazy.layout.down()),
    Key([mod], "k", lazy.layout.up()),
    Key([mod], "Return", lazy.spawn("bash -c 'SHELL=/usr/bin/fish exec sakura'")),
    Key([mod], "q", lazy.window.kill()),
    Key([mod, "control"], "r", lazy.restart()),
    Key([], "Print", lazy.spawn('grim -g "0,0 1920x1080"')),
]
groups = [Group(i) for i in "1234567"]


for i in groups:
    keys.append(Key([mod], i.name, lazy.group[i.name].toscreen()))
[...]

PR #5869, opened 2026-03-14
Fix window lost focus when dragged between screens

### LINKED ISSUE #5843: [wayc] Focus loss while dragging a floating window accross the boundary between screens.
### Issue description

https://github.com/user-attachments/assets/3b3f3240-1d61-4298-a0a7-7b758c445669

1. Have the default config with 2 displays set up to be touching. 
2. Start dragging a window across the screen boundary

After roughly half of the window is across the boundary line it loses focus, meanwhile I can still actively continue dragging.

### Version

0.34.2.dev66+g3505e0631

### Backend

Wayland (experimental)

### Config

```python
tested with minimal config
```

### Logs

```text
-
```

### Required

elParaguayo:
Floating windows lose their focus border when dragged between screens. This is because the floating layout draws the border if client.has_focus is True. This is the case if the window is equal to qtile.current_window. qtile.current_window looks at the active group on the current screen. However, when dragging windows, we only change screen focus after the window has been added to the other group so qtile.current_window is looking at the old screen.

Fixes #5843

### DIFF
--- a/libqtile/backend/wayland/window.py
+++ b/libqtile/backend/wayland/window.py
@@ -828,8 +828,8 @@ class Window(Base, base.Window):
         screen = self.qtile.find_closest_screen(x + w // 2, y + h // 2)
         if self.group and screen is not None and screen != self.group.screen:
             self.group.remove(self, force=True)
-            screen.group.add(self, force=True)
             self.qtile.focus_screen(screen.index)
+            screen.group.add(self, force=True)
 
         self._reconfigure_floating(x, y, w, h)
 


Sinaaaa:
I tested this PR & it seems to work at fixing #5843 , thank you!

Though what's really interesting that this PR makes reproducing #5844 very easy.  (though could be coincidental, since I have not tested this in a while)

On minimal config:
1. Open a terminal on display A, open another on display B.
2. Make them both floaters.
3. Start resizing the one on display A to be so large as to overlap with the one on display B.

Now even very slow, snail like resizing will produce results at the overlap.

Sinaaaa:
>(though could be coincidental, since I have not tested this in a while)

Not a coincidence, it's much harder without this PR, just tested that too.

elParaguayo:
> > (though could be coincidental, since I have not tested this in a while)
> 
> Not a coincidence, it's much harder without this PR, just tested that too.

Maybe that's why the code was done the way it was.

I'll take a look when I can.

tych0:
What's the status here, should we land this or?

elParaguayo:
> What's the status here, should we land this or?

I'm inclined to say yes and then look at #5844 separately.

PR #5871, opened 2026-03-15
Introduce generate screens

tych0:
Here is a new config API: generate_screens(), that people can use to do screen matching, instead of us writing matchers on `serial`, `name`, etc.

Hopefully this can also be the basis for us doing new config stuff like calling this when new screens are attached, etc.

This also includes my first completely-AI commit; the subtle fix with `__annotations__` had stumped me for several weeks, it was only when I realized running the whole test suite reproduces things that Claude could sleuth it out.

### DIFF
--- a/docs/manual/config/screens.rst
+++ b/docs/manual/config/screens.rst
@@ -82,6 +82,147 @@ Wayland
 The Wayland backend supports the wlr-output-management protocol for configuration of
 outputs by tools such as `Kanshi <https://github.com/emersion/kanshi>`_.
 
+Dynamic Screen Configuration
+============================
+
+Instead of defining ``screens`` as a static list, you can define a
+``generate_screens`` function that is called with the list of currently
+connected outputs and returns the appropriate list of ``Screen`` objects. This
+lets you adapt your bar layout to different hardware setups, e.g. a laptop on
+its own, docked with a second monitor, or at a desk with multiple displays.
+
+The function signature must be::
+
+    from libqtile.config import Output, Screen
+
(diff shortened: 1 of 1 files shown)


ervinpopescu:
Time for CLAUDE.md in the repo! Just kidding! (unless...) :)

Any chance you can document an example of `generate_screens`? I'm going to admit I did not investigate how `serial`,`name` and the rest used to work (also they were added pretty recently, right?), but I still think a proper example extra to the comments would help.

tych0:
> Any chance you can document an example of generate_screens? I'm going to admit I did not investigate how serial,name and the rest used to work (also they were added pretty recently, right?), but I still think a proper example extra to the comments would help.

Yeah, great point, thank you. I will clean up my config and add it to this PR.

tych0:
And yeah, they were added recently, but also have not been released, which is why it's important to do this refactoring now :)

richcarni:
So the previous method of specifying name or serial in Screen is gone and the equivalent "simple" match by name would look like the following?

It's a bit more work but much more powerful and yeah, I think you're right, a single interface for this is the way to go

```python
def generate_screens(outputs: list[Output]) -> list[Screen]:
    screens = []
    for output in outputs:
        if output.name == "HDMI-2":
            scr = Screen(top=bar.Bar([
                ...
            ], height))
        elif output.name == "eDP-1":
            scr = Screen(top=bar.Bar([
                ...
            ], height))
        screens.append(scr)
    return screens
```

tych0:
Yep, exactly. So when we add more stuff to Output like some folks were asking for, it will immediately be available to use for this case.

PR #5873, opened 2026-03-16
feat(backend): add macOS backend

ervinpopescu:
## Summary

This PR adds an initial macOS backend for qtile, enabling it to run as a floating window manager / bar host on macOS via the Accessibility API and CoreGraphics.

> **Note:** This code was developed collaboratively with [Claude](https://claude.ai) (Anthropic's AI assistant, specifically Claude Sonnet 4.6 via Claude Code CLI). All design decisions and review were done by a human; Claude was used as an accelerated pair-programmer.

### What's included

- **Native ObjC layer** (`libqtile/backend/macos/src/`) — window management via AXUIElement, event tap for key/mouse capture, display enumeration, idle inhibition via IOPMAssertion, and a Cairo-backed drawing surface using NSWindow overlays
- **CFFI bindings** (`libqtile/backend/macos/cffi/build.py`) — Python ↔ ObjC bridge; the generated `_ffi.c` is excluded from the repo
- **Python backend** (`libqtile/backend/macos/`) — `Core`, `Window`, `Drawer`, `InputManager`, `IdleNotifier`, `IdleInhibitorManager` implementing the same base-class interfaces as the X11/Wayland backends
- **Unit tests** (`test/backend/macos/test_unit_*.py`) — Linux-runnable tests using a `FakeFFI` stub; no macOS hardware needed to run them
- **CI matrix** — macOS jobs (Python 3.12 and 3.13 on `macos-latest`) added alongside the existing x11/wayland Linux matrix

### Known limitations / not yet implemented

- No window tiling of native macOS apps (AXUIElement only manages focus/raise; actual layout requires Accessibility permission and has OS-level constraints)
- Wayland and X11 features that have no macOS equivalent (e.g. `set_urgent`, some wm hints) are no-ops
- Integration tests require a real macOS machine; only unit tests run in CI

### TODO / follow-up work

**Core functionality**
- [ ] Window tiling: investigate AXUIElement layout constraints and implement where possible
- [ ] Screen change events: react dynamically to display connect/disconnect (currently static at startup)
- [ ] `set_urgent` / wm-hint equivalents: map to `NSApplication.requestUserAttention` where possible
- [ ] Window minimize/maximize: wire `AXMinimizedAttribute` and `AXZoomedAttribute` through to qtile layout hooks
- [ ] Window close: implement `cmd_kill` via `AXCloseButton` action
- [ ] Application spawning: validate `subprocess`-launched apps are correctly picked up by the AX observer
- [ ] Input: media keys, brightness, and volume key routing through the event tap
- [ ] Input: IME / input method passthrough for non-ASCII keyboard layouts
- [ ] Mouse: cursor theme / cursor shape changes (e.g. resize handles)
- [ ] Mouse: drag-and-drop passthrough while qtile holds the event tap

**Rendering / drawing**
- [ ] Retina / HiDPI: audit `CGDisplayBounds` logical-vs-physical pixel handling throughout the drawing layer
- [ ] Bar z-order: ensure bar windows always sit above all app windows after a Space switch or full-screen app
- [ ] Transparency: verify premultiplied-alpha compositing matches expected widget output

**Integration & ecosystem**
- [ ] Clipboard: implement `qtile.cmd_spawn` clipboard passthrough via `NSPasteboard`
- [ ] Systray / menu-bar widgets: investigate `NSStatusItem` integration
- [ ] Notifications: wire `libnotify`-style notifications through `NSUserNotificationCenter` / `UNUserNotificationCenter`
- [ ] Screen lock: integrate with `CGSessionCopyCurrentDictionary` for lock-screen detection
- [ ] Power: expose battery / AC state via `IOPSCopyPowerSourcesInfo` for widget use
- [ ] qtile-extras compatibility audit: identify which extras widgets work as-is and which need macOS shims

**Testing & CI**
- [ ] Integration tests: add macOS integration test scaffold (requires macOS runner or self-hosted)
- [ ] Memory / leak testing: run under Instruments / Valgrind equivalent to catch AXUIElementRef leaks
- [ ] Widget smoke tests: run standard bar widget suite against the macOS Drawer

**Packaging & distribution**
- [ ] Documentation: add macOS backend page to the sphinx docs (install, Accessibility permission, known limits)
- [ ] Packaging: ensure `pyproject.toml` extras and wheel build include the ObjC extension on macOS
- [ ] Entitlements: document required `com.apple.security.accessibility` entitlement for sandboxed distribution
- [ ] Homebrew / pipx install path: verify `cffi/build.py` finds SDK headers when installed via Homebrew Python

### How to build and test

```bash
# On macOS — build the native extension
python libqtile/backend/macos/cffi/build.py

# Unit tests (run anywhere, no macOS needed)
python -m pytest test/backend/macos/test_unit_*.py -x -q --tb=short
```

### DIFF
--- a/libqtile/backend/macos/cffi/build.py
+++ b/libqtile/backend/macos/cffi/build.py
@@ -14,6 +14,8 @@ MAC_SRC_PATH = (Path(__file__).parent / ".." / "src").resolve()
 
 CDEF = """
 // display.h
+void mac_init_app(void);
+
 struct mac_output {
     char *name;
     int x;
--- a/libqtile/backend/macos/core.py
+++ b/libqtile/backend/macos/core.py
@@ -19,6 +19,7 @@ class Core(base.Core):
 
         self._ffi = _ffi.ffi
         self._lib = _ffi.lib
+        self._lib.mac_init_app()
         self._callback_handle = None
         self._observer_handle = None
         self._poll_handle: typing.Any = None
--- a/libqtile/backend/macos/src/display.h
+++ b/libqtile/backend/macos/src/display.h
@@ -5,6 +5,11 @@
 #include <stddef.h>
 #include <stdint.h>
 
+// Must be called before any AppKit API (NSWindow, NSView, etc.).
+// Initialises [NSApplication sharedApplication] and sets the activation policy
+// to accessory (no dock icon, no menu bar).
+void mac_init_app(void);
+
 struct mac_output {
     char *name;
     int x;
--- a/libqtile/backend/macos/src/display.m
+++ b/libqtile/backend/macos/src/display.m
@@ -1,7 +1,16 @@
 #include "display.h"
+#import <AppKit/NSApplication.h>
 #import <ApplicationServices/ApplicationServices.h>
 #import <Foundation/Foundation.h>
 #import <IOKit/pwr_mgt/IOPMLib.h>
+
+void mac_init_app(void) {
+    // NSApplication must exist before creating any NSWindow.
+    // Accessory policy: no dock icon, no main menu bar — qtile is a background
+    // window-management process, not a regular GUI app.
+    [NSApplication sharedApplication];
+    [NSApp setActivationPolicy:NSApplicationActivationPolicyAccessory];
+}
 
 static IOPMAssertionID g_idle_assertion = kIOPMNullAssertionID;
 
--- a/test/backend/macos/fake_ffi.py
+++ b/test/backend/macos/fake_ffi.py
@@ -8,6 +8,9 @@ class FakeLib:
     explicit stub here returning a sensible default so tests never hit real C code."""
 
     def __init__(self):
+        # Initialisation
+        self.mac_init_app = mock.MagicMock(return_value=None)
+
         # Integer / bool stubs
         self.mac_get_idle_time = mock.MagicMock(return_value=0.0)
         self.mac_observer_start = mock.MagicMock(return_value=0)


test/backend/x11/conftest.py
```diff
@@ -78,6 +78,36 @@ def stop_x11(proc, display, display_file):
...
+                return display, f
+            except OSError:
+                f.close()
+        except OSError:
+            pass
+
+        display += 1
+
+
+xcffib.testing.find_display = safe_find_display
```
tych0, line 108: seems like a bugfix we could send to xcffib?

tych0:
I have been waiting for this kind of PR :)

I'll be honest, I don't think we can really support an OSX backend. But if we could, I would like it to come in the form of small digestible PRs that multiple people really understand, vs one big one like this.

elParaguayo:
Could we have a way to have third party backends which are not part of the main repo?

Also, I don't know how much of qtile-extras would be compatible with this but I wouldn't be able to fix any bits that don't work.

Despite that, this is very exciting to see!

ervinpopescu:
> I'll be honest, I don't think we can really support an OSX backend.

Support -- as in most maintainers don't want to run macOS right? :)

> But if we could, I would like it to come in the form of small digestible PRs that multiple people really understand, vs one big one like this.

I **tried** to split this into logical commits <1k lines. What would be the point of introducing a non-working backend in a PR and fix it in another one? However, I'm open to discussing this.

> Could we have a way to have third party backends which are not part of the main repo?

I would also second this, if there is no intention in merging this.

> Also, I don't know how much of qtile-extras would be compatible with this but I wouldn't be able to fix any bits that don't work.

Yeah, that is expected, but maybe we can just have the qtile-only stuff? No need for fancy borders and such.
 
> Despite that, this is very exciting to see!

Glad to hear it. If you want to talk further about it I am available on Discord (but don't open it very often so ping me here if I don't respond).

NOTE: I am not sure why wayland breaks. Will investigate. This is anyways in a very rough form as shown by todos above.

test/backend/x11/conftest.py
```diff
@@ -78,6 +78,36 @@ def stop_x11(proc, display, display_file):
...
+def safe_find_display():
+    display = 10
+    while True:
+        lock_path = f"/tmp/.X{display}-lock"
+        sock_path = f"/tmp/.X11-unix/X{display}"
+
+        # Check if socket already exists (stale or active)
+        if os.path.exists(sock_path):
+            display += 1
+            continue
```
ervinpopescu, line 93: @tych0 this is the only change. I don't think it's a good fix.

> That skip-if-socket-exists guard is actually weaker than xcffib's approach — a stale socket with no live lock would cause our version to needlessly skip a valid display, whereas
>   xcffib's version would claim it (no lock held → lock succeeds). The monkey-patch was likely added to fix what looked like a race but is already handled upstream.
> 
>   It can be removed — both the function and the xcffib.testing.find_display = safe_find_display assignment, and the import fcntl that was added solely for it.
>

ervinpopescu:
```
  /app/qtile-source/.venv/lib/python3.12/site-packages/_pytest/unraisableexception.py:67: PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO name='/tmp/.X10-lock' mode='rb+' closefd=True>

    File "/app/qtile-source/test/backend/x11/conftest.py", line 37, in xvfb
      with xcffib.testing.XvfbTest():
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/.X10-lock' mode='w+' encoding='utf-8'>

  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))

```

But this **should** be fixed in xcffib, right? :)

I think I saw this multiple times in CI.

tych0:
> Support -- as in most maintainers don't want to run macOS right? :)

Exactly. I don't have any hardware that can run it, I don't know anything about the UI frontend, etc.

I'm not opposed to merging something like this in principle, but I think we'd want to add you to the maintainer team, and hopefully even one other person who is interested in it as well.

> What would be the point of introducing a non-working backend in a PR and fix it in another one?

The idea is that it should be digestible, which 4k lines all at once is not. My experience with models is that it is very easy to produce a lot of code in a hurry, and very hard to determine whether or not it makes sense, especially in areas where I don't have a lot of expertise like the OSX UI libraries.

It is a useful practice to split things up into "atomic" digestible commits. I ask this of all PRs, whether they are vibe coded or not. Having monkey patching of xcffib in the middle of commits adding an OSX backend is definitely a code smell, and I have really only looked at this for 30 seconds.

I would also like to see documentation, e.g. consideration of support matrix. What is the lifecycle of OSX UI libraries? How will that integrate with qtile's historically supported last-three-python-versions? How do I install this thing in the first place?

But taking a step back. Just because we *can* do this, doesn't mean we *should*. Producing code is easy now, but long term maintenance of API surface is not. This discussion has been ongoing across the internet recently, but I personally am not especially interested in maintaining large amounts of AI output. That is not to say that models are not useful. Indeed, I have myself recently used them to debug various issues. But I would like to be sure that a human has reviewed every line of code that goes in, and I'm not convinced of that here.

This is just my opinion, though, and not the opinion of the other maintainers. Perhaps they see it as less of an issue. I am happy to be overruled if so. We have talked about having an AI policy, and now that we are getting more AI PRs, perhaps it is time to write one.

test/backend/x11/conftest.py
tych0: So which claude prompt was right, the first one, or the second one? :)

But yes, I agree this is a real issue. It should be fixed in xcffib, though. There is an unreleased patch in the top of the tree that might fix this, I haven't tested for this exact case.

richcarni:
I've been trying to think about this as if there was no AI involved (that said, I 100% agree with everything tycho said about it above)

My main concern would be whether it's actually feasible to run qtile on macos without significant compromises? I know there are some existing tiling window managers for macos, but I have personally never tested them. The fact that they are all built on the accessibility API leads me to suspect there must be some inherent limitations?

I'm not necessarily against the idea of adding a macos backend, but I think I need convincing that it wouldn't be a severely crippled version of qtile

And as mentioned, we would need a couple of maintainers (including you) who are interested and willing to devote time to it (because I 'm personally not planning on going back to macos, it's just not of much interest/value to me)

ervinpopescu:
I am not ignoring you tycho and rich btw. I am trying to get my thoughts straight as well, as I was only trying to gauge the interest into this and also get you guys' opinion on the AI part.

I will come back with a more extensive message on how exactly I tried to "generate" all of this and also how much time I took to make sure Claude wasn't going and chasing golden goose eggs etc etc.

That said, I am completely on board with you guys. AI slop is not ok + I mostly have not seen much that this particular backend can give over other similar window managers.

"Skeletor will return with more disturbing facts!"

test/backend/x11/conftest.py
ervinpopescu: > So which claude prompt was right, the first one, or the second one? :)

Trick question. ;)

> But yes, I agree this is a real issue. It should be fixed in xcffib, though. There is an unreleased patch in the top of the tree that might fix this, I haven't tested for this exact case.

Ok, I'm not going to try to dabble in xcffib, not even with an "AI". That is not at all something I know anything about. :)

ervinpopescu:
> I'm not opposed to merging something like this in principle, but I think we'd want to add you to the maintainer team, and hopefully even one other person who is interested in it as well.

I would not mind that at all, I think I might be able to manage it fine nowadays, I definitely have a lot more time on my hands that I had one year ago for example. I have tried to keep close with this project to which I was drawn to from the first few days when I found it, but I haven't particularly managed it...

> > What would be the point of introducing a non-working backend in a PR and fix it in another one?
> 
> The idea is that it should be digestible, which 4k lines all at once is not. My experience with models is that it is very easy to produce a lot of code in a hurry, and very hard to determine whether or not it makes sense, especially in areas where I don't have a lot of expertise like the OSX UI libraries.
> 
> It is a useful practice to split things up into "atomic" digestible commits. I ask this of all PRs, whether they are vibe coded or not. Having monkey patching of xcffib in the middle of commits adding an OSX backend is definitely a code smell, and I have really only looked at this for 30 seconds.
> 

Let me be upfront about the AI involvement: I used Claude to generate the bulk of the initial scaffolding, but I didn't just paste it in and open a PR. I spent a significant amount of time iterating on it: reading through the generated output, understanding what it was doing, questioning decisions, and making sure it wasn't just hallucinating its way through the Cocoa/AppKit APIs.  That said, I completely understand if that's not enough reassurance. Tycho, you're right that producing code is easy now; the hard part is verifying it, and you shouldn't have to take my word for it.

I also agree with both of you that 4k lines at once is not reviewable, and I take the point about xcffib monkey-patching being a code smell, I noticed it too and should have flagged it explicitly rather than leaving it in. That's on me. FWIW, it was due to a local test issue which Claude couldn't get around so I just allowed it at that time.

> I would also like to see documentation, e.g. consideration of support matrix. What is the lifecycle of OSX UI libraries? How will that integrate with qtile's historically supported last-three-python-versions? How do I install this thing in the first place?

Those are valid questions, here's some short-ish answers I have:
1. macOS API lifecycle: The backend uses the Accessibility API (AX) for window management, which has been stable since macOS 10.9 and is the same foundation that Yabai, Amethyst, and others rely on. It also uses AppKit, CoreGraphics, Carbon (keyboard events), and IOKit (idle inhibit) — all of which are first-party Apple frameworks with no deprecation signals. The main risk is Apple tightening Accessibility permissions further, which they've done incrementally since Catalina. That's a real concern worth documenting. Two callouts:
    - `_AXUIElementGetWindow` — private SPI, runtime-resolved with a documented fallback. This is the same approach [yabai](https://github.com/asmvik/yabai/blob/54728ce89a134d3ff83b356b1f0b86fba50b90b5/src/misc/extern.h#L7) uses. Real risk: Apple could remove it silently in any release.
    - `activateWithOptions:NSApplicationActivateIgnoringOtherApps` — deprecated macOS 14 (Sonoma). The code already suppresses the warning. The replacement (activate on NSRunningApplication, available 14+) would need a runtime version check to support older macOS.

2. Python version support: The backend currently targets Python 3.12+, which is more restrictive than qtile's general support matrix. That's something to reconcile, either relax it or explicitly document the divergence.

3. Installation: It requires granting Accessibility permissions in System Settings, the Xcode command-line tools (for the CFFI/Objective-C compilation step), and the same cairocffi/cffi dependencies qtile already uses. None of that is documented yet.

4. What's missing: A proper `docs/manual/install/macos.rst` covering prerequisites, the permission grant flow, known limitations, and the support matrix. I haven't written that yet.

> But taking a step back. Just because we _can_ do this, doesn't mean we _should_. Producing code is easy now, but long term maintenance of API surface is not. This discussion has been ongoing across the internet recently, but I personally am not especially interested in maintaining large amounts of AI output. That is not to say that models are not useful. Indeed, I have myself recently used them to debug various issues. But I would like to be sure that a human has reviewed every line of code that goes in, and I'm not convinced of that here. 

Without trying to contradict my above statement about the AI involvement, I cannot say "under oath" that I did review each and every line of code that the AI produced, especially in the ObjC case. That said, I have not opened this PR to have any sort of gains for myself. It was a WIP, I specifically marked it as draft for that reason, but perhaps I was too hasty and did not realize I should have worked a bit more on it to be acceptable.
 
>  We have talked about having an AI policy, and now that we are getting more AI PRs, perhaps it is time to write one.

An AI policy is surely something `qtile` would need if we are to live in the modern world, should we like it or not.

-------

So here's where I land: I think the right thing to do is step back, close this PR, and come back with:

1. A clear write-up of what's actually possible on macOS and what the limitations are
2. An honest (self-)assessment of whether it's worth the maintenance burden
3. If we proceed, proper atomic commits with documentation

I don't want to be the person who dumps a pile of AI output on a project I respect and says "figure it out". That's not the kind of contributor I want to be.

If you still want me to continue working on this PR as-is, let me know.

PR #5874, opened 2026-03-16
docs: add profiling section to hacking guide 

ervinpopescu:
Add a short "Profiling" section to the hacking guide covering the two tools that work on a live Qtile process with no code changes or restart required:
- py-spy for CPU flamegraphs (interactive SVGs) and live top-style view
- qtile top for memory via tracemalloc (already built in)

### DIFF
--- a/docs/manual/hacking.rst
+++ b/docs/manual/hacking.rst
@@ -271,6 +271,32 @@ Note that we used the same display, ``:1``, in both the terminal command
 and the VSCode configuration environment variables.  Then ``debug`` usually
 in VSCode. Feel free to change the screen size to fit your own screen.
 
+Profiling
+=========
+
+`py-spy <https://github.com/benfred/py-spy>`_ is a sampling profiler that
+attaches to any running Python process with no code changes and no restart
+required:
+
+.. code-block:: bash
+
+    pip install py-spy
+
+    # Record a 30-second flamegraph (interactive SVG)
+    py-spy record -o flamegraph.svg -d 30 --pid $(pidof qtile)
+
+Open ``flamegraph.svg`` in a browser to explore the call stack interactively.
+``py-spy top`` gives a live ``top``-style view of where time is being spent.
+
+For memory, use the built-in :doc:`qtile top </manual/commands/shell/qtile-top>`
+subcommand, which wraps :mod:`tracemalloc`:
+
+.. code-block:: bash
+
+    qtile top          # live curses view of top allocators
+    qtile top --raw    # one-shot snapshot
+
+
 Resources
 =========
 


libqtile/profiling.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @contextlib.contextmanager
+    def timed(self, category: str, key: str) -> Generator[None, None, None]:
+        yield
+
+    def flush(self) -> None:
+        pass
+
+
+class ProfileCollector:
```
tych0: This whole set of infra seems unnecessary. Can we just enable the python builtin profiling from https://docs.python.org/3/howto/perf_profiling.html and then use perf's infra to generate profiles and draw flamegraphs and stuff, vs. having our own json output?

ervinpopescu: I did add both as it seemed like an interesting idea to first do the `profile-report` then dig in to the `perf` part. Here's what Claude had to say about it as I did ask the same question at the beginning:

> These serve different purposes:
> 
>   **`perf record` / py-spy** — statistical CPU sampling across the whole process.
>   Great for flamegraphs and finding unexpected hotspots. Requires `perf` installed
>   (and often elevated privileges). Answers: *where is the CPU spending time?*
> 
>   **Custom JSON instrumentation** — explicit wall-clock timing of specific
>   categories (`widget_draw`, `event_dispatch`, `startup`). No extra tools needed,
>   works unprivileged. Answers: *which widget is slow? which event type? how long
>   did startup take?* — surfaced directly without digging through a flamegraph.
> 
>   `perf` can't easily tell you "the Clock widget takes 4ms mean to draw" without
>   manually filtering the call graph down the draw path. The instrumentation
>   surfaces that in a one-liner:
> 
>   ```sh
>   qtile profile-report ~/qtile_profile.json
>  ```
>   Both are opt-in and zero-overhead when inactive. They're complementary:
>   instrumentation for day-to-day triage, perf/py-spy for deep dives.

If you don't think this makes sense, I can strip it down.

tych0: I don't think it's necessary to find hotspots, no. Have you tried using it? Is there a particular optimization that you think it would reveal that classic perf would not?

> surfaced directly without digging through a flamegraph.

> without manually filtering the call graph down the draw path.

in my experience, digging through flamegraphs is not hard. It's just some clicking. But it's not really about having the profiling available, but what optimizations we can glean from it. I would try to play around with it and see what you think. IMO the model is just wrong here and it's a lot of extra code without any value.

ervinpopescu: I did try it, it does not help with optimization. It just helps with filtering where you would want to look (for example a particular widget), instead of mentally holding a tree of functions like you have to do with a flamegraph. 

But I ultimately agree with you, the "price" of the lengthy code does not translate into real value.

Another option would be using external scripts/full fledged apps to read the perf output and do something with it. Which really sounds like reinventing the wheel, someone must have done it before. That would be more or less "throwaway" code (for lack of a better word). 

Anyway, is the perf part good? Do you see any changes required for it?

richcarni:
My two cents...

Some of these like draw timing, reload timing are interesting

But as for perf, py-spy is pretty easy to use as a stand alone utility. Is there much benefit to integrating it?

Kinda same argument for memory if it's using qtile top anyway

I guess this is targeted at helping hackers and contributors. I could be wrong here but, for me, finding the cause of performance related issues isn't usually the hard part (once you know how!) So documentation (how to use py-spy, qtile top, etc.) is probably as helpful as anything here

ervinpopescu:
> But as for perf, py-spy is pretty easy to use as a stand alone utility. Is there much benefit to integrating it?

I thought so as well, but wanted to give this a try and seeing whether there is interest. I honestly like py-spy more than perf anyway for python :)

> So documentation (how to use py-spy, qtile top, etc.) is probably as helpful as anything here

+1

tych0:
heh, I forgot about py-spy. But I had it installed, and I have this in my history: `py-spy record -o profile.svg --pid $(pidof qtile)`, so clearly I've used it before :). The SVGs it generates are interactive, which is kind of awesome.

Yeah, I was mostly thinking of having a flag that would automatically do:

```
sys.activate_stack_trampoline("perf")
```

for us, but py-spy doesn't even need that. So maybe we don't need anything?

Just running py-spy now and at least with my config, I don't see any crazy hotspots. It's all stuff like pangocffi.get_pixel_size(), xcffib's request_check(), etc. which is essentially just ffi overhead. (cffi/api.py::new shows up as well...).

ervinpopescu:
So just adding a section in https://docs.qtile.org/en/stable/manual/hacking.html would be enough, right?

PR #5876, opened 2026-03-18
fix: skip __main__ module during config reload

sohanglal:
## Summary

Fixes `ModuleNotFoundError: spec not found for the module '__main__'` when calling `qtile cmd-obj -o cmd -f restart` or `reload_config`.

## Problem

When using `_reload_config_submodules()`, the function iterates over `sys.modules` and attempts to reload modules that are in the config folder. However, the `__main__` module may be present in `sys.modules` and has a `__file__` attribute pointing to a file in the config folder (since qtile is started from there), but it cannot be reloaded because it has no module spec.

This causes:
```
ModuleNotFoundError: spec not found for the module '__main__'
```

## Solution

Add an explicit check to skip `__main__` before attempting to reload modules.

## Reproduction

1. Have a config that imports a local module (e.g., `from topics import something`)
2. Run `qtile cmd-obj -o cmd -f restart`
3. Observe the error in qtile.log

## Testing

After this fix, `qtile cmd-obj -o cmd -f restart` works correctly when the config imports local modules.

### DIFF
--- a/libqtile/confreader.py
+++ b/libqtile/confreader.py
@@ -85,6 +85,10 @@ class Config:
         """Reloads python files from same folder as config file."""
         folder = path.parent
         for module in sys.modules.copy().values():
+            # Skip __main__ as it cannot be reloaded (has no module spec)
+            if getattr(module, "__name__", None) == "__main__":
+                continue
+
             # Skip built-ins and anything with no filepath.
             if hasattr(module, "__file__") and module.__file__ is not None:
                 subpath = Path(module.__file__)


ervinpopescu:
I never encountered this, even though I have a multi file config?

https://github.com/ervinpopescu/dots/blob/lenovo/configs/.config/qtile/config.py

https://github.com/ervinpopescu/dots/blob/lenovo/configs/.config/qtile/modules/decorations.py

sohanglal:
This only occurs when I use/call qtile from within a virtual environment installed by uv. 

I agree, this issue doesn't occur when I use qtile from my system package manager (nixpkgs - NixOS)

tych0:
Hmm. Can you paste the full stack trace?

tych0:
Or better yet: can you add a test demonstrating the issue so that we can ensure it stays fixed?

sohanglal:
stack trace:

```
2026-03-18 16:22:17,966 WARNING libqtile start.py:start():L89 Starting Qtile 0.34.1 from /home/sl/dotfiles/qtile/.venv/lib/python3.13/site-packages
2026-03-18 16:23:35,874 ERROR libqtile manager.py:reload_config():L322 Configuration error:
  File "/home/sl/data/pkms/attachments/dotfiles/qtile/.venv/lib/python3.13/site-packages/libqtile/core/manager.py", line 320, in reload_config
    self.config.load()
    ~~~~~~~~~~~~~~~~^^
  File "/home/sl/data/pkms/attachments/dotfiles/qtile/.venv/lib/python3.13/site-packages/libqtile/confreader.py", line 111, in load
    self._reload_config_submodules(path)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/sl/data/pkms/attachments/dotfiles/qtile/.venv/lib/python3.13/site-packages/libqtile/confreader.py", line 100, in _reload_config_submodules
    importlib.reload(module)
    ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/sl/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/lib/python3.13/importlib/__init__.py", line 128, in reload
    raise ModuleNotFoundError(f"spec not found for the module {name!r}", name=name)
ModuleNotFoundError: spec not found for the module '__main__'

```

Will try writing a test to catch this

sohanglal:
Please review @tych0

ervinpopescu:
I don't know if I am doing something wrong, but I am running into this when trying to `reload_config`:
```
$ uv run qtile cmd-obj -f reload_config
error: Sorry cannot run function 'reload_config' with arguments []: TypeError: initializer for ctype 'xcb_connection_t *' appears indeed to be 'xcb_connection_t *', but the types are different (check that you are not e.g. mixing up different ffi instances)
```
Full stacktrace:
```
2026-03-19 19:47:37,401 ERROR libqtile loop.py:_handle_exception():L60 Exception in event loop:
  File "$HOME/.config/qtile/.venv/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 327, in _xpoll
  File "$HOME/.config/qtile/.venv/lib/python3.13/site-packages/xcffib/__init__.py", line 596, in wrapper
    self.invalid()
    ~~~~~~~~~~~~^^
  File "$HOME/.config/qtile/.venv/lib/python3.13/site-packages/xcffib/__init__.py", line 583, in invalid
    err = lib.xcb_connection_has_error(self._conn)
TypeError: initializer for ctype 'xcb_connection_t *' appears indeed to be 'xcb_connection_t *', but the types are different (check that you are not e.g. mixing up different ffi instances)

  File "/usr/lib/python3.13/asyncio/events.py", line 89, in _run
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "$HOME/.config/qtile/.venv/lib/python3.13/site-packages/libqtile/backend/x11/core.py", line 372, in _xpoll
  File "$HOME/.config/qtile/.venv/lib/python3.13/site-packages/xcffib/__init__.py", line 645, in has_error
    return lib.xcb_connection_has_error(self._conn)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: initializer for ctype 'xcb_connection_t *' appears indeed to be 'xcb_connection_t *', but the types are different (check that you are not e.g. mixing up different ffi instances
```

Here's how I install qtile: https://pastebin.com/Xr3SfZC8
 
I also tried:
```
qtile = { git = "https://github.com/sohanglal/qtile", branch = "fix/reload-config-skip-main" }
```

```
$ uv run qtile --version
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.13`.
0.1.dev6233+g2408a1c44
```

test/test_config.py
```diff
@@ -209,3 +211,45 @@ def the_order(self) -> list[Output]:
...
+
+    subdir = tmp_path / "submodules"
+    subdir.mkdir()
+    submodule_file = subdir / "helper.py"
+    submodule_file.write_text("# helper module")
+
+    # Create a fake __main__ module with __file__ pointing to a file
+    # in a subfolder of the config directory
+    fake_main = ModuleType("__main__")
+    fake_main.__file__ = str(submodule_file)
```
tych0: Is it possible to have the test work without this? I feel like this is implementation specific vs. having a test that just does the thing and doesn't monkey patch.

tych0:
> TypeError: initializer for ctype 'xcb_connection_t *' appears indeed to be 'xcb_connection_t *', but the types are different (check that you are not e.g. mixing up different ffi instances)

I wonder if we can't restart across things where the ffi signature has changed?

sohanglal:
Please review @tych0

sohanglal:
I realized the bug was occurring because I had the Qtile .venv installed inside the config directory. Running qtile from a separate .venv fixed the issue. I apologize for the noise! Tycho, I appreciate you pushing back on the changes, as it forced me to dig deeper into the root cause. Going forward, I’ll be sure to open a detailed issue for discussion before submitting a PR. Thanks for all your work on Qtile!

ervinpopescu:
@tych0 what if I want to have a .venv and pyproject.toml in config dir to install all necessary deps I have in my config? Must it be separate to it? Sounds a bit weird to not be able to do this?

Where would my pyproject.toml or .venv "live" if not in the config dir? I can't place it in the qtile dir since obviously it would cause a conflict to qtile's pyproject.toml (and .venv if present).

PR #5881, opened 2026-03-21
Fix systray finalize when the window is already removed

sohanglal:
## Summary
- add a regression test for `Systray.finalize()` when the systray window has already been removed from `windows_map`
- make systray teardown tolerate missing window map entries so config reloads do not fail with `KeyError`
- preserve `_instances` cleanup so the systray can be configured again after reload

### DIFF
--- a/libqtile/widget/systray.py
+++ b/libqtile/widget/systray.py
@@ -265,7 +265,7 @@ class Systray(base._Widget, window._Window):
         except xcffib.ConnectionException:
             self.hidden = True  # Usually set in self.hide()
 
-        del self.qtile.windows_map[self.wid]
+        self.qtile.windows_map.pop(self.wid, None)
         Systray._instances -= 1
 
     def info(self):


test/widgets/test_systray.py
```diff
@@ -78,6 +79,37 @@ def test_systray_reconfigure_screens(manager_nospawn, minimal_conf_noscreen, bac
     assert manager_nospawn.c.bar["top"].info()["widgets"][0]["name"] == "systray"
 
 
+def test_systray_finalize_ignores_missing_window(monkeypatch):
+    tray = widget.Systray()
+
+    monkeypatch.setattr(widget.base._Widget, "finalize", lambda self: None)
+    monkeypatch.setattr(widget.Systray, "_instances", 1)
```
tych0: Instead of monkeypatch and SimpleNamespace, can you do the action that triggers the stack trace?

sohanglal: I have replicated the action that triggers the stack trace in the test

test/widgets/test_systray.py
```diff
@@ -78,6 +79,45 @@ def test_systray_reconfigure_screens(manager_nospawn, minimal_conf_noscreen, bac
...
+    if backend_name == "wayland":
+        pytest.skip("Skipping test on Wayland.")
+
+    config = minimal_conf_noscreen
+    config.screens = [libqtile.config.Screen(top=libqtile.bar.Bar([widget.Systray()], 10))]
+
+    manager_nospawn.start(config)
+
+    manager_nospawn.c.widget["systray"].eval(
+        "self.conn.conn.core.DestroyWindow(self.wid); self.conn.conn.flush()"
```
tych0, line 94: I'm a little confused here. Is there a way to use this using the qtile API, instead of evaling random x11 functions?

I would like to see the sequence of calls from qtile that trigger the failure.

sohanglal:
See: https://github.com/qtile/qtile/pull/5876#issuecomment-4169090920

PR #5606, opened 2025-10-26
Add touch screen support plus pointer gestures

elParaguayo:
Work in progress...

PR #5882, opened 2026-03-21
[wayland] Add basic touch device support

elParaguayo:
Adds support for touch devices. Currently supports touch and motion only.

This replaces #5606 which tried to do too much too soon by adding configurable gesture support. We can still add that but let's get the basics working first.

### DIFF
--- a/libqtile/backend/wayland/inputs.py
+++ b/libqtile/backend/wayland/inputs.py
@@ -75,6 +75,11 @@ class InputConfig(configurable.Configurable):
         ("kb_variant", None, "Keyboard variant i.e. ``XKB_DEFAULT_VARIANT``"),
         ("kb_repeat_rate", 25, "Keyboard key repeats made per second"),
         ("kb_repeat_delay", 600, "Keyboard delay in milliseconds before repeating"),
+        (
+            "map_output",
+            None,
+            "Map touch device to specified output. Use device name or '*' to map over the entire output areas",
+        ),
     ]
 
     def __init__(self, **config: Any) -> None:
@@ -231,5 +236,16 @@ def configure_input_devices(server: ffi.CData, configs: dict[str, Any]) -> None:
                     ffi.new("char[]", (conf.kb_options or "").encode()),
                     ffi.new("char[]", (conf.kb_variant or "").encode()),
                 )
+
+            elif type == InputDeviceType.TOUCH:
+                touch = lib.qw_input_device_get_touch(input_device)
+                if touch == ffi.NULL:
+                    return
+
+                if conf.map_output is not None:
+                    all_outputs = conf.map_output == "*"
+                    mapped = lib.qw_touch_map_output(touch, conf.map_output.encode(), all_outputs)
+                    if not mapped:
+                        pass
 
     lib.qw_server_loop_input_devices(server, input_device_cb)
--- a/libqtile/backend/wayland/qw/input-device.c
+++ b/libqtile/backend/wayland/qw/input-device.c
@@ -49,6 +49,7 @@ void qw_server_input_device_new(struct qw_server *server, struct wlr_input_devic
         wlr_cursor_attach_input_device(server->cursor->cursor, device);
         break;
     case WLR_INPUT_DEVICE_TOUCH:
+        wlr_cursor_attach_input_device(server->cursor->cursor, device);
         qw_touch_handle_new(server, device);
         break;
     default:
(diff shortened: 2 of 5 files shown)


pranavsetpal:
Hey, firstly thank you so much for this PR!

Here are the issues I have found:
1. Touch only seems if:
   a. I tap on the region where the window is actually visible (as intended), and
   b. it is the "first" client / window in my workspace. By this, I am referencing `qtile shell > cd layout > info()`. (NOT as intended)

2. Touch input does not take scaling into account. That is, if I run `wlr-randr --output eDP-1 --scale 2` and give in any touch inputs, the touch inputs are calculated based on the initial `--scale 1` positions of my windows.

3. If multiple screens are connected to a computer, each screen's touch input affects only the "main" screen (I am not sure what constitutes as main, perhaps `qtile shell > get_screens()` being index 0?). However, the inputs are scaled correctly here. That is, my secondary screen has a different size than my main screen, but top of secondary screen maps to top of main screen, midway of secondary screen maps to midway of main screen, and bottom of secondary screen maps to bottom of main screen.


Comment:
It seems some apps or UI elements simply don't support touch. For instance, touch
   a. works in Firefox, but if it does NOT work I try to click on any action in its Hamburger Menu
   b. does NOT work on Vesktop (Discord frontend alternative)
   c. does NOT work on kitty (terminal emulator).
Upon some searching, I can confirm Kitty does not support touch (https://github.com/kovidgoyal/kitty/issues/5432). Idk others. Maybe this is why some display managers choose to also move the cursor on touch. Regardless, if you do implement "move cursor on touch", I think it makes to have that be optional, as I found "move cursor on touch" extremely annoying during my extremely limited use of KDE (literally only used it to test if my computer's touch functionality indeed works)

elParaguayo:
Great feedback. Thank you.

I can take a look at these but it's hard to test when I don't have a touch screen device.

Will push updates and let you know if I need you to test again.

pranavsetpal:
Oh, also pinch zoom does work! Not sure how to test rotate tbh.

I can try to work on a PR since I can test it, but I am otherwise pretty low on time and have zero experience with qtile and wayland internals, so it may take a while.
I will definitely test any updates you push ASAP. Feel free to throw at me as many iterations as you wish :)

elParaguayo:
I know how to fix pinch and zoom (need to track separate touches).

I'm testing by writing a script to mimic touches.

elParaguayo:
Ok. I suspect most of your issues were a result of some bad logic in my positional calculations. I _think_ that's fixed in latest commit!

elParaguayo:
@pranavsetpal Are you able to test with the latest commit here?

pranavsetpal:
Oh I am so sorry I thought I had responded!

The latest commit was really good. The first two issues are entirely resolved, and the third one semi. I'll try to explain the issue as best I can.

Suppose I have two touch monitors, both 1920x1280 placed side by side.
In my display configuration, my first monitor is in coordinates (0, 0) to (1920, 1280) and my second monitor is in coordinates (1920, 0, 1920+1920, 1280) = (1920, 0, 3840, 1280). I will call my region (0, 0) to (3840, 1280) my "full display" and all these coordinates virtual coordinates. (idk real name for this).

The touch input on any monitor is mapping from my monitor's coordinates to my the full display coordinates.
For example,
on tapping center of 1st display, input goes from my monitor's (960, 640) [in virtual coordinates (960, 640)] to (1920, 640)
on tapping center of 2nd display, input goes from my monitor's (960, 640) [in virtual coordinates (2880, 640)] to (1920, 640)

So essentially each monitor is separately treated like the "full display" for touch purposes.

Does this make sense?


The Firefox issue I mentioned may have just been because of how the touch was mapped before, because that does not happen now as well! Discord and Kitty still don't work but that was expected, and afaik there is nothing to do about it.

elParaguayo:
Thanks. The issue is that my code maps each touch device to the full output area (e.g. the area covered by all monitors). I need to update this so each device is only attached to its respective output device. Not quite sure how to do that yet so meed to do some research!

EDIT: ok. Pretty sure I know what I need to do. Will push an update when I have time.

elParaguayo:
@pranavsetpal Ok - need you to do some more testing, please.

I have hopefully added the ability to map touch devices to outputs but these needs some configuration on your part.

Firstly, you need to know the device name of your touch device(s). Using `sudo libinput list-devices` should list all devices. Find the ones you want and note the device name e.g. something like:
```
Device:                  DLL0704:01 06CB:76AE Touchpad
```
You also need the names of the outputs.

Then, in your config, you need:
```python

...

wl_input_rules = {
    "[touch-device-name]": InputConfig(map_output="[output-name]"),
    ...
}
```

I have no idea if this will work! If it does,

nerothank:
Chiming in here. Thank you for your efforts!
NOTE: I rebased this PR to master before running.

Touch on clients is working reasonably well for me. However, I am facing instant segmentation faults when performing any swipes in a touch-enabled window. Not sure whether this is caused by the rebase, newer logic in other parts of qtile, or simply by the setup I am running.

Qtile: 0.37.1 (Commit 83c697a5), with this PR rebased above
OS: Debian Sid, wlroots-0.19.3-1 and wlroots-0.20.2-1 installed (not sure which one qtile actually uses)
Python: 3.14.7, separate venv for qtile
Hardware: Lenovo Yoga 260

@pranavsetpal you did not write about similar issues, so I guess for now this is isolated on my device?

elParaguayo:
I should pick this up again!

Thanks for the comments. I'll try to take a look when I can.

PR #5889, opened 2026-03-24
Fix test_net.py

elParaguayo:
The test uses a fake psutil module to get net statistics. However, values are increased between tests and so will fail if tests are run in a different order. We fix this by resetting stats between tests.


Sigmanificient (approved):
- This patch works for https://github.com/NixOS/nixpkgs/pull/502350

tych0:
This is another instance where I wonder if forking for each test would help us.

Anyway, I think I've seen the same failures when I've tried to run the tests with pytest-xdist before to try and parallelize them, so having it fixed will be great, thank you.

PR #5897, opened 2026-03-28
Restore solid background for systray widgets

### LINKED ISSUE #5856: Systray Icons background, not filled anymore
### Issue description

Hi everyone,
thk for the really nice job done here.

I have a cosmetic issue with the last update I made (0.34.1.r96.g25e3e9a-1).
The background of the systray icons are not filled anymore :
Image

Rolling back to the previous one (0.34.1.r64.g020a87d-1) solved the problem.
[2026-02-27T14:41:25+0100] [ALPM] downgraded qtile-git (0.34.1.r96.g25e3e9a-1 -> 0.34.1.r64.g020a87d-1)

Image

I didn't find anything relevant in the Changelog and I don't know how to investigate any further

Thk for your time

### Version

0.34.2.dev64+g020a87db1

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

elParaguayo:
eab276e forced all windows to be 32-bit depth if the monitor supports this. However, we need 24-bit depth to set the background colour (I've no idea why but I could never get it to work - this is why we have that warning in the docs about transparent bars).

This PR restores 24 bit depth for bars that do not request a (semi-)transparent background colour.

Fixes #5856

Before:

After:

### DIFF
--- a/libqtile/bar.py
+++ b/libqtile/bar.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 from libqtile import configurable, hook
 from libqtile.command.base import CommandObject, expose_command
 from libqtile.log_utils import logger
-from libqtile.utils import is_valid_colors
+from libqtile.utils import has_transparency, is_valid_colors
 
 if typing.TYPE_CHECKING:
     import asyncio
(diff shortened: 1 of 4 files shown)


extremepayne:
Tested on EndeavourOS, seems to be working.

qtile 0.35.0 (from Arch `extra/qtile`)
qtile-trans

qtile 0.35.1.dev5+g2f0c78db6 (commit 2f0c78db)
qtile-opaque

libqtile/bar.py
```diff
@@ -266,7 +266,15 @@ def _configure(self, qtile: Qtile, screen: Screen, reconfigure: bool = False) ->
             # Whereas we won't have a window if we're startup up for the first time or
             # the window has been killed by us no longer using the bar's screen
 
-            self.window = qtile.core.create_internal(self.x, self.y, width, height)
+            if qtile.core.name == "x11":
+                if has_transparency(self.background):
+                    depth = 32
+                else:
+                    depth = qtile.core.conn.default_screen.root_depth  # type: ignore[attr-defined]
```
elParaguayo, line 273: This should fix the test error but means we now need to make sure we're in the x11 backend (and mypy isn't clever enough to work out that `qtile.core` here is the x11 core).

PR #5898, opened 2026-03-29
Fix `qtile cmd-obj -f spawn` crash

### LINKED ISSUE #5893: qtile cmd-obj -o cmd -f spawn -a "command" throws an error
### Issue description

Running qtile 0.35.0-2 on arch.
Have a bunch of fzf menu scripts with qtile cmd-obj -o cmd -f spawn -a "command"
Used to work fine, now throws an error.

Steps to reproduce:
Type qtile cmd-obj -o cmd -f spawn -a "command" with any command

### Version

0.35.0

### Backend

X11 (default)

### Config

```python

```

### Logs

```text
  File "/usr/lib/python3.14/site-packages/libqtile/ipc.py", line 102, in unpack
    assert len(data) >= HDRLEN
           ^^^^^^^^^^^^^^^^^^^
AssertionError

The above exception was the direct cause of the following exception:

  File "/usr/bin/qtile", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib/python3.14/site-packages/libqtile/scripts/main.py", line 77, in main
    func(options)
    ~~~~^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/libqtile/scripts/cmd_obj.py", line 178, in cmd_obj
    ret = run_function(obj, args.function, args.args, args.kwargs)
  File "/usr/lib/python3.14/site-packages/libqtile/scripts/cmd_obj.py", line 132, in run_function
    ret = client.call(funcname, *args, **kwargs, lifted=True)
  File "/usr/lib/python3.14/site-packages/libqtile/command/client.py", line 111, in call
    return self._command.execute(call, args, kwargs)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/libqtile/command/interface.py", line 240, in execute
    status, result = self._client.send(
                     ~~~~~~~~~~~~~~~~~^
        (call.parent.selectors, call.name, args, kwargs, call.lifted)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.14/site-packages/libqtile/ipc.py", line 152, in send
    return asyncio.run(self.async_send(msg))
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/asyncio/runners.py", line 204, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/lib/python3.14/asyncio/runners.py", line 127, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/libqtile/ipc.py", line 180, in async_send
    data, _ = _IPC.unpack(read_data, is_json=self.is_json)
              ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/libqtile/ipc.py", line 107, in unpack
    raise IPCError("error reading reply! (probably the socket was disconnected)") from e
libqtile.ipc.IPCError: error reading reply! (probably the socket was disconnected)
```

### Required

elParaguayo:
4c37efd introduced additional types that could be passed to `qtile.spawn`. This caused IPC to crash as it was unable to lift string arguments to this new type.

This PR adds better handling of argument types.

Fixes #5893

### DIFF
--- a/libqtile/command/interface.py
+++ b/libqtile/command/interface.py
@@ -351,7 +351,10 @@ def lift_args(cmd, args, kwargs):
     converted_args = []
     converted_kwargs = dict()
 
-    params = typing.get_type_hints(cmd, globalns=globals())
+    # We use the globals from the command itself to ensure that all types are
+    # available. This means that structural types need to imported at run time
+    # and not just for typing.
+    params = typing.get_type_hints(cmd, globalns=cmd.__globals__)
 
     non_return_annotated_args = filter(lambda k: k != "return", params.keys())
     for param, arg in zip(non_return_annotated_args, args):


elParaguayo:
@tych0 you may want to look at this one as it's to do with the `lift_args` code (specifically, how it handles `Sequence` and/or `PathLike`).

tych0:
Oh yeah, that's annoying... thanks for the fix. Maybe the json rework will fix this for us?


But if not, I'll have to think about how to avoid this in the future.

libqtile/core/manager.py
```diff
@@ -13,9 +13,11 @@
 import tempfile
 import time
 from collections import defaultdict
+from collections.abc import Sequence
 from logging.handlers import RotatingFileHandler
+from os import PathLike
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
```
elParaguayo: @tych0 One of the side effects of the lifting code is that the types need to be imported all the time and not inside the `if TYPE_CHECKING` block.

elParaguayo: There's probably a whole load that I've missed too. Not sure the best way to deal with this - just take every module that contains a `CommandObject` and move all `typing` imports to the main import section?

tych0: Yeah... I never really loved this. I wonder if we can look in the globals for the function that's being lifted into somehow?

tych0: I wonder if this really depends on #5925 ?

tych0: Or probably really it depends on python 3.14's lazy loading, since everything needs to be imported...

PR #5911, opened 2026-04-05
Add position support for Mpris2 widget

elParaguayo:
Resurrecting #4944 so draft for now.

This adds playback position support for the Mpris2 widget. Ideally it needs testers and feedback!

### DIFF
--- a/libqtile/utils.py
+++ b/libqtile/utils.py
@@ -467,6 +467,29 @@ async def _send_dbus_message(
     return bus, msg
 
 
+async def add_match_rule(
+    bus: MessageBus, preserve: bool = False, **kwargs: str | None
+) -> tuple[MessageBus, Message]:
+    rule = ",".join(f"{k}='{v}'" for k, v in kwargs.items() if v)
+
+    logger.debug("Adding dbus match rule: %s", rule)
+
+    bus, msg = await _send_dbus_message(
+        False,
+        MessageType.METHOD_CALL,
+        "org.freedesktop.DBus",
+        "org.freedesktop.DBus",
+        "/org/freedesktop/DBus",
+        "AddMatch",
+        "s",
+        [rule],
+        bus=bus,
+        preserve=preserve,
+    )
+
+    return bus, msg
+
+
 async def add_signal_receiver(
     callback: Callable,
     session_bus: bool = False,
@@ -500,36 +523,35 @@ async def add_signal_receiver(
             )
             return False
 
+    if use_bus is None:
+        if session_bus:
+            bus_type = BusType.SESSION
+        else:
+            bus_type = BusType.SYSTEM
+
+        try:
+            bus = await MessageBus(bus_type=bus_type).connect()
+        except (AuthError, Exception):
+            logger.warning("Unable to connect to dbus.")
+            return False
+    else:
+        bus = use_bus
+
     match_args = {
-        "sender": bus_name,
-        "member": signal_name,
-        "path": path,
+        "type": "signal",
         "interface": dbus_interface,
+        "path": path,
+        "member": signal_name,
+        "sender": bus_name,
     }
 
-    rule = "type='signal',"
-    rule += ",".join(f"{k}='{v}'" for k, v in match_args.items() if v)
-
-    logger.debug("Adding dbus match rule: %s", rule)
-
-    bus, msg = await _send_dbus_message(
-        session_bus,
-        MessageType.METHOD_CALL,
-        "org.freedesktop.DBus",
-        "org.freedesktop.DBus",
-        "/org/freedesktop/DBus",
-        "AddMatch",
-        "s",
-        [rule],
-        bus=use_bus,
-        preserve=preserve,
-    )
+    bus, msg = await add_match_rule(bus, preserve=preserve, **match_args)
 
     # Check if message sent successfully
     if bus and msg and msg.message_type == MessageType.METHOD_RETURN:
 
         def match_message(msg: Message, match_args: dict[str, str | None]) -> bool:
-            return all(getattr(msg, k) == v for k, v in match_args.items() if v)
+            return msg._matches(**{k: v for k, v in match_args.items() if (v and k != "type")})
 
         async def resolve_sender(signal_msg: Message) -> tuple[str, Message]:
             """Looks up a pretty bus name to retrieve the unique name."""
--- a/libqtile/widget/mpris2widget.py
+++ b/libqtile/widget/mpris2widget.py
@@ -12,7 +12,7 @@ from dbus_fast.constants import MessageType
 from libqtile import pangocffi
 from libqtile.command.base import expose_command
 from libqtile.log_utils import logger
-from libqtile.utils import _send_dbus_message, add_signal_receiver, create_task
+from libqtile.utils import _send_dbus_message, add_match_rule, create_task
 from libqtile.widget import base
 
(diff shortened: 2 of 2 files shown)


theincompletedetective:
I'm using the same configuration options as before, but neither the position nor the length show up in the bar: `widget.Mpris2(format="{xesam:title} - {xesam:album} - {xesam:artist} - {position}/{mpris:length}"),`

elParaguayo:
Which player are you using?

theincompletedetective:
> Which player are you using?

I tried with VLC and Cmus, but since it looks like you made some commits yesterday, I think I might not have all the changes. When I tried to do a `git pull`, I get an error that there is no tracking information for the current branch. What's the correct command to set my local branch to track your remote one?

elParaguayo:
Have you just cloned my repo or added it as another remote to a different fork?

Assuming you cloned separately:
```
# Get latest version of the remote repository
git fetch origin
# Checkout the branch
git checkout mpris2-position
# Discard local changes and make sure local branch matches remote branch
git reset --hard origin/mpris2-position
```

theincompletedetective:
I cloned the main repo, instead of yours; now I have the right branch, and both the position and length are showing up. I will continue testing and provide feedback.

elParaguayo:
That's getting beyond my git skills! If you've just cloned my repo, yes, delete the folder and clone again.

theincompletedetective:
image

When I started watching this video, the time was correct, but I paused it for a while, and when I went back to play it, the time in the widget is wrong.

theincompletedetective:
- The widget shows the correct position for VLC videos at normal and 2x speeds.
- The widget shows the correct position after I pause a VLC video, change the position (both backwards and forwards), and then press play.
- The widget shows the correct position after I change the position of a VLC video (both backwards and forwards) while it's playing.
- The widget shows the correct position for YouTube video after it's loaded and I press play.
- The widget shows the correct position if a YouTube video is playing at 1x or 2x speed, and I change the position (both backwards and forwards).
- The widget displays the correct position when playing an audio track in Cmus.
- The widget displays the correct position after I fast-forward or rewind a track in Cmus.
- The widget _doesn't_ display the correct position if I pause a YouTube video, change the speed from 1x to 2x, and then press play; it still shows as if it's playing at 1x speed. If I change the position after that, then it shows the correct one.

elParaguayo:
Thanks. Let me look into Youtube.

elParaguayo:
From what I can tell, it's a bug in the youtube player. Pausing a video emits a `Seeked` signal with the position. Resuming playback does the same but the `Seeked` time has incremented while the video was paused. So, the widget is behaving correctly.

theincompletedetective:
> From what I can tell, it's a bug in the youtube player. Pausing a video emits a `Seeked` signal with the position. Resuming playback does the same but the `Seeked` time has incremented while the video was paused. So, the widget is behaving correctly.

That makes sense to me. Have you run into any other bugs with this widget?

elParaguayo:
Youtube also doesn't implement the `Position` property correctly so the fallback check via a background poll also doesn't work!

theincompletedetective:
> Youtube also doesn't implement the `Position` property correctly so the fallback check via a background poll also doesn't work!

Would it be possible to store the position when it's paused in a variable, and then when play is clicked you set it to that before sending the signal?

elParaguayo:
No. Because we've got to treat the `Seeked` signal as authoritative otherwise there's no way to differentiate a play/resume vs a play and a seek.

theincompletedetective:
If certain clients don't properly or consistently implement the Mpris2 specification, then it seems like it's not the widget's fault, and perhaps documentation should be added so that the end users know in which situations it might not work as expected. It doesn't seem to me to be something that should prevent it from being made available, unless there are other, show-stopping bugs.

elParaguayo:
I think that's right.

I wonder if it's worth having an option to toggle the position text so you can choose to hide it for bad players.

theincompletedetective:
> I think that's right.
> 
> I wonder if it's worth having an option to toggle the position text so you can choose to hide it for bad players.

Perhaps there could be an option to provide a list of players for which the position wouldn't be displayed; I'm wondering how you would be able to implement toggling a field on the fly.

elParaguayo:
> > I think that's right.
> > I wonder if it's worth having an option to toggle the position text so you can choose to hide it for bad players.
> 
> Perhaps there could be an option to provide a list of players for which the position wouldn't be displayed; I'm wondering how you would be able to implement toggling a field on the fly.

Well, the issue is that the browser is the name of the player e.g. `firefox` not `youtube` and `firefox` seems to work with some other players.

The toggling could just be a keybinding to `lazy.widget["mpris2"].toggle_position_display()` (or something like that).

theincompletedetective:
> > > I think that's right.
> > > I wonder if it's worth having an option to toggle the position text so you can choose to hide it for bad players.
> > 
> > 
> > Perhaps there could be an option to provide a list of players for which the position wouldn't be displayed; I'm wondering how you would be able to implement toggling a field on the fly.
> 
> Well, the issue is that the browser is the name of the player e.g. `firefox` not `youtube` and `firefox` seems to work with some other players.
> 
> The toggling could just be a keybinding to `lazy.widget["mpris2"].toggle_position_display()` (or something like that).

Understood. How difficult would it be to implement this?

elParaguayo:
Have pushed the update.

I've added ``position_text_format`` and ``no_position_text_format`` options. In the main `format` you need to add `{qtile:position_text}` and this will be replaced by those new options. You toggle between them with `lazy.widget["mpris2"].toggle_position_text()`.

theincompletedetective:
> Have pushed the update.
> 
> I've added `position_text_format` and `no_position_text_format` options. In the main `format` you need to add `{qtile:position_text}` and this will be replaced by those new options. You toggle between them with `lazy.widget["mpris2"].toggle_position_text()`.



> Have pushed the update.
> 
> I've added `position_text_format` and `no_position_text_format` options. In the main `format` you need to add `{qtile:position_text}` and this will be replaced by those new options. You toggle between them with `lazy.widget["mpris2"].toggle_position_text()`.

This works for me. Perhaps there should be a default value for the `no_position_text`.

elParaguayo:
There is. It's no text i.e. `{qtile:position_text}` either shows `(position/length)` or nothing.

theincompletedetective:
When I use the `{qtile:player}`, and start playing with one player, if I start another player the name doesn't change.
image

elParaguayo:
That's a different issue. There's no logic in the widget to deal with priority of multiple players.

theincompletedetective:
Is `lazy.widget["mpris2"].toggle_position_text()` supposed to work whether the track is playing or paused? It only works when it's playing.

elParaguayo:
Yes. Need to fix that.

theincompletedetective:
> Yes. Need to fix that.

This is working now.

theincompletedetective:
If I have a video open in VLC and then close the application, should the track information still show up on the widget?

elParaguayo:
> If I have a video open in VLC and then close the application, should the track information still show up on the widget?

No. The widget listens for a signal to sound that the player's mpris interface has been removed. Sometimes it doesn't work.

Are you seeing this consistently with VLC? I'm not.

theincompletedetective:
This is the first time I've encountered it; I'm not sure exactly what triggered it.

elParaguayo:
I doubt it's caused by this PR as I've seen that behaviour before.

PR #5915, opened 2026-04-06
Cancel BackgroundPoll widget tasks on reload

### LINKED ISSUE #5913: BackgroundPoll widgets keep running after a config reload
### Issue description

It seems BackgroundPoll widgets don't get destroyed on a config reload and instead they keep running in the background.

Add this custom widget to a bar:

```python

class Background_Custom(widget.base.BackgroundPoll):
    def poll(self):
        logger.info("I'm alive")
        return "-"
```

Check the logs:

```bash
tail -f ~/.local/share/qtile/qtile.log
```

After reloading the configuration the speed will increase, hinting that the previous widgets' background tasks are still alive.

### Version

0.35.0

### Backend

Wayland (experimental)

### Config

```python

```

### Logs

```text

```

### Required

elParaguayo:
`BackgroundPoll` widgets' tasks don't get destroyed on a config reload. Instead, they carry on running and can contain references to finalised objects.

Fixes #5913

~If this works, it will need a changelog entry before merging.~ Edit: confirmed as working. changelog updated.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Qtile x.xx.x, released xxxx-xx-xx:
     * features
     * bugfixes
+      - Fix timers on BackgrounPoll-based widgets not clearing on config reload
 
 Qtile 0.35.0, released 2026-03-20:
     !!! Breaking changes !!!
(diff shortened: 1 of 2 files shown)


elParaguayo:
@tych0 I think this one's good to go now.

tych0:
Thanks!

PR #5919, opened 2026-04-08
Improve battery name getter for Linux in the battery widget

T1tukll:
Added support for non-standard ACPI names like 'axp288_fuel_gauge' found in some budget notebooks.

### DIFF
--- a/libqtile/widget/battery.py
+++ b/libqtile/widget/battery.py
@@ -230,7 +230,13 @@ class _LinuxBattery(_Battery, configurable.Configurable):
 
     def _get_battery_name(self):
         if os.path.isdir(self.BAT_DIR):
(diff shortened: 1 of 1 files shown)


libqtile/widget/battery.py
```diff
@@ -230,7 +230,13 @@ def __init__(self, **config):
 
     def _get_battery_name(self):
         if os.path.isdir(self.BAT_DIR):
-            bats = [f for f in os.listdir(self.BAT_DIR) if f.startswith("BAT")]
+            bats = []
+            for f in os.listdir(self.BAT_DIR):
+                if f.startswith("BAT"):
+                    bats.append(f)
```
elParaguayo, line 236: Do you need a `continue` after this? If a folder starts with `BAT` and had a `capacity` subfolder then it's going to get added twice.

Or change the second check to `elif` rather than `if`...

PR #5920, opened 2026-04-08
build: upgrade wlroots 0.19 → 0.20

Gurjaka:
Update flake.lock nixpkgs revision to pull wlroots 0.20 and adjust all build references accordingly.

Changes:
- Bump WLROOTS_PATH and library link from 0.19 to 0.20
- Update cursor.c XWayland cursor handling to use new buffer-based API
- Replace direct buffer access (image->buffer, stride calculations) with wlr_xcursor_image_get_buffer()

Wlroots 0.20 release notes: https://gitlab.freedesktop.org/wlroots/wlroots/-/releases/0.20.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,10 @@ RUN dnf install -y \
     python3.12 python3.12-devel \
     python3.13 python3.13-devel \
     python3.14 python3.14-devel \
+    wayland-devel \
     wayland-protocols-devel \
-    wlroots0.19-devel \
+    wlroots \
+    wlroots-devel \
     xcb-util-cursor \
     xorg-x11-server-Xorg \
     xorg-x11-server-Xephyr \
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1772542754,
-        "narHash": "sha256-WGV2hy+VIeQsYXpsLjdr4GvHv5eECMISX1zKLTedhdg=",
+        "lastModified": 1775423009,
+        "narHash": "sha256-vPKLpjhIVWdDrfiUM8atW6YkIggCEKdSAlJPzzhkQlw=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "8c809a146a140c5c8806f13399592dbcb1bb5dc4",
+        "rev": "68d8aa3d661f0e6bd5862291b5bb263b2a6595c9",
         "type": "github"
       },
       "original": {
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -12,7 +12,7 @@ from setuptools import Distribution
 from setuptools.command.build_ext import build_ext
 
 QW_PATH = (Path(__file__).parent / ".." / "qw").resolve()
-WLROOTS_PATH = os.getenv("QTILE_WLROOTS_PATH", "/usr/include/wlroots-0.19")
+WLROOTS_PATH = os.getenv("QTILE_WLROOTS_PATH", "/usr/include/wlroots-0.20")
 
 PKG_CONFIG = os.environ.get("PKG_CONFIG", "pkg-config")
 WAYLAND_SCANNER = os.environ.get("QTILE_WAYLAND_SCANNER", shutil.which("wayland-scanner"))
@@ -230,7 +230,7 @@ INCLUDE_DIRS = [
     QW_PATH,
     QW_PROTO_OUT_PATH,
 ]
-LIBRARIES = ["wlroots-0.19", "wayland-server", "input", "cairo"]
+LIBRARIES = ["wlroots-0.20", "wayland-server", "input", "cairo"]
 
 # SOURCE_FILES loads all .c files...
 SOURCE_FILES = glob.glob(f"{QW_PATH}/*.c")
--- a/libqtile/backend/wayland/qw/cursor.c
+++ b/libqtile/backend/wayland/qw/cursor.c
@@ -619,8 +619,8 @@ void qw_cursor_configure_xcursor(struct qw_cursor *cursor) {
             wlr_xcursor_manager_get_xcursor(cursor->xwayland_mgr, "default", 1);
         if (xcursor != NULL) {
             struct wlr_xcursor_image *image = xcursor->images[0];
-            wlr_xwayland_set_cursor(server->xwayland, image->buffer, image->width * 4, image->width,
-                                    image->height, image->hotspot_x, image->hotspot_y);
+            struct wlr_buffer *buffer = wlr_xcursor_image_get_buffer(image);
+            wlr_xwayland_set_cursor(server->xwayland, buffer, image->hotspot_x, image->hotspot_y);
         }
     }
 #endif
--- a/nix/build-config.nix
+++ b/nix/build-config.nix
@@ -48,8 +48,8 @@ let
     }
     {
       varname = "WLROOTS";
-      pkg = pkgs.wlroots_0_19;
-      header-dir = "wlroots-0.19";
+      pkg = pkgs.wlroots_0_20;
+      header-dir = "wlroots-0.20";
     }
   ];
 
--- a/nix/qtile.nix
+++ b/nix/qtile.nix
@@ -44,5 +44,5 @@ let
     };
 in
 (pkgs.python3Packages.qtile.overrideAttrs qtile-override-func).override {
-  wlroots = pkgs.wlroots_0_19;
+  wlroots = pkgs.wlroots_0_20;
 }


elParaguayo:
This is a good test as to whether writing our own compositor in C was worth the effort!

Gurjaka:
> This is a good test as to whether writing our own compositor in C was worth the effort!

Fun fact: River already supports 0.20.0

*Investigating CI failure*

elParaguayo:
@richcarni What's our approach here? Do we hold back 0.20.0 support until the next release?

Gurjaka:
image

image

Should we wait for the apt repo update or just build it from source?


EDIT: Using nix is also viable for this

elParaguayo:
I'd build from source. Can't wait for repo updates.

tych0:
I wonder if this is an argument for doing CI in a container or in nix or something, so we can use a distro that's more up to date than the latest ubuntu LTS, and we don't have to build from source all the time?

Gurjaka:
> I wonder if this is an argument for doing CI in a container or in nix or something, so we can use a distro that's more up to date than the latest ubuntu LTS, and we don't have to build from source all the time?

Using nix would be a great option in my opinion, it can be installed on ubuntu runner as well, and our flake is always up-to-date with latest changes. Building Qtile would be really simple using it as well, and it would be much faster than building deps from source

elParaguayo:
If it's easy to do, can you do a PR to use nix so we can see how CI works? I don't know how much has to change.

Gurjaka:
> If it's easy to do, can you do a PR to use nix so we can see how CI works? I don't know how much has to change.

I'll try to make one later today, need to figure out how current CI works as well :)

Gurjaka:
> If it's easy to do, can you do a PR to use nix so we can see how CI works? I don't know how much has to change.

https://github.com/qtile/qtile/pull/5921

Here you go

richcarni:
> @richcarni What's our approach here? Do we hold back 0.20.0 support until the next release?

I'm in favour of doing this now. As far as I can tell only a single breaking change that affects us (and has been addressed in this PR). Are there other downsides I'm not considering? I guess updated dependencies is going to be an issue for people like me, stubbornly clinging to ubuntu LTS

> Fun fact: River already supports 0.20.0

👿

richcarni:
> I wonder if this is an argument for doing CI in a container or in nix or something, so we can use a distro that's more up to date than the latest ubuntu LTS, and we don't have to build from source all the time?

In the interests of evaluating all options, I assume "in a container" means docker? If considering that option, what would be a good base? Fedora? And update every 6 months?

richcarni:
@Gurjaka Now that we've moved CI to fedora, if you rebase, you should be able to modify in Dockerfile: wlroots0.19 -> wlroots, and this should work 🤞

Gurjaka:
> @Gurjaka Now that we've moved CI to fedora, if you rebase, you should be able to modify in Dockerfile: wlroots0.19 -> wlroots, and this should work 🤞

Alright, will do once I get home

richcarni:
> > @Gurjaka Now that we've moved CI to fedora, if you rebase, you should be able to modify in Dockerfile: wlroots0.19 -> wlroots, and this should work 🤞
> 
> Alright, will do once I get home

No stress. Whenever you feel like working on it!

Gurjaka:
Can anyone re-ran tests? looks like it got stuck

tych0:
I think it got stuck on git master too just now, seems like a result of the docker migration, probably unrelated to this PR.

tych0:
So it's green... any reason not to land?

scripts/ubuntu_wayland_setup
richcarni, line 1: With the new ci we decided to remove this script

Dockerfile
```diff
@@ -15,8 +15,10 @@ RUN dnf install -y \
     python3.12 python3.12-devel \
     python3.13 python3.13-devel \
     python3.14 python3.14-devel \
+    wayland-devel \
     wayland-protocols-devel \
-    wlroots0.19-devel \
+    wlroots-0.20.0-1.fc44 \
+    wlroots-devel \
```
richcarni: I think this makes sense to pin the wlroots version?
In which case should the dev package be: wlroots-devel-0.20.0-1.fc44?

tych0: I wouldn't pin the packaging version, isn't their stable version just the major?

Gurjaka: Honestly I'm not familiar with dnf packaging, will modify as you guys decide

scripts/ubuntu_wayland_setup
Gurjaka: Got no idea why it's still her, I rebased but somehow it got re-added.

Dockerfile
richcarni: I was under the (I think mistaken) belief that fedora might update wlroots version mid release

Suggest following Tycho's advice on this

richcarni (approved):

elParaguayo:
Do we need a new release if we're bumping wlroots version?

Gurjaka:
> Do we need a new release if we're bumping wlroots version?

Might not be necessary for now since this doesn't introduce any major changes in codebase

Dockerfile
tych0: From my Fedora experience, it's that 'wlroots' will always be the latest, 'wlroots-$ver' will always be that version, and 'wlroots-$ver-$package' will hard code to that specific package. I don't think we want the last one, because if they fix a bug we won't get it (like a missing shared object, or ABI fix or whatever).

The first one will be fine across stable releases of fedora, but will change from e.g. F44 to F45. The second one won't change. I don't think it matters between these two.

Dockerfile
```diff
@@ -15,8 +15,10 @@ RUN dnf install -y \
     python3.12 python3.12-devel \
     python3.13 python3.13-devel \
     python3.14 python3.14-devel \
+    wayland-devel \
     wayland-protocols-devel \
-    wlroots0.19-devel \
+    wlroots-0.20.0-1.fc44 \
+    wlroots-devel-0.20.0-1.fc44 \
```
richcarni: if wlroots-0.20 is valid here, I reckon do that (for devel too), otherwise just wlroots

Gurjaka: Last time I checked, just defining wlroots pulled v0.19, that's why I'm specifying 0.20 now. Also shouldn't CI follow you current supported version? What if Fedora updates to 0.21 while our implementation is still at 0.20?

ervinpopescu: Barging in, but wouldn't wlroots-0.20 work for that case like richcarni mentioned above?

Gurjaka: You mean just using `wlroots-0.20`? Didn't know it was possible , since I couldn't find in dnf package repos 😄 really not familiar with Fedora

ervinpopescu: I was just quoting Tycho actually, I'm as unfamiliar with Fedora as you probably, I looked at https://packages.fedoraproject.org/pkgs/wlroots/wlroots/ and I can't see `wlroots-$ver` as well that Tycho mentioned.

Buuut. It looks like we do have `wlroots-0.20.0` to use (not `wlroots-0.20`): https://gist.github.com/ervinpopescu/c302b53a05674790df114f112f7c8dd1. 

Let's say we take Fedora 43 instead as an exercise: https://gist.github.com/ervinpopescu/40e0a81e5d6a1fd8f9b5283bbda0965c.

As you can see, you need to use the exact version ([0.19.3](https://packages.fedoraproject.org/pkgs/wlroots/wlroots/fedora-43-updates.html), instead of `0.19.0` for example or `0.19`) otherwise the package won't be found. And obviously we don't have `0.20.0` available (or even more obviously `0.20`).

richcarni: > Last time I checked, just defining wlroots pulled v0.19

I dropped into an interactive fedora:44 container:
```
[root@812ca81a25a3 workspace]# dnf info wlroots | grep Version
Updating and loading repositories:
Repositories loaded.
Version        : 0.20.0
Version        : 0.20.0
```

If we're really worried about the wlroots version being updated within the fedora 44 release (unlikely?), I think we can do this before the dnf install step:
```
dnf versionlock add "wlroots-0.20.*"
```
(then just `dnf install wlroots wlroots-dev`)

Gurjaka: So am I supposed to change anything here?

tych0: I would change them both to `wlroots-0.20` and `wlroots-devel-0.20`; the point release, packaging version, etc. don't need to be hardcoded.

Gurjaka: done

Dockerfile
```diff
@@ -15,8 +15,10 @@ RUN dnf install -y \
     python3.12 python3.12-devel \
     python3.13 python3.13-devel \
     python3.14 python3.14-devel \
+    wayland-devel \
     wayland-protocols-devel \
-    wlroots0.19-devel \
+    wlroots-0.20.0 \
+    wlroots-devel-0.20.0 \
```
tych0: I think you can drop the .0 here on these two, but otherwise LGTM.

Gurjaka: What if `0.20.1` gets released, and tests break because we don't yet support it? That's why I left `.0` in here :)

Gurjaka: What if `0.20.1` gets released, and tests break because we don't yet support it? That's why I left `.0` in here :)

richcarni: Minor releases shouldn't have breaking changes - I think we can live with the small risk they may still break tests

Gurjaka: hm, that seems to fail for some reason.

tych0: Oh, yeah, because there is no wlroots-0.20; fedora does `wlroots` and `wlroots-devel` as the latest, and adds the hardcoded versions for previous versions of wlroots, e.g. 0.19, 0.18, etc.

I have never seen them (and I suspect they never will) bump a wlroots major version in the middle of a release, so I suspect we can just use `wlroots` and `wlroots-devel` here. We'll stay on fedora 44, and when we bump to fedora 45 that will necessitate a wlroots update.

tych0:
Ok, lgtm. Feel free to merge or I can in the next couple of days.

tych0:
@Gurjaka did you mean to force push over my fix here?

Gurjaka:
> @Gurjaka did you mean to force push over my fix here?

No? Maybe I mixed up the branches 😔


EDIT: applied your fix

PR #5925, opened 2026-04-16
*: drop some __future__ and TYPE_CHECKING imports

tych0:
Move imports out of `if TYPE_CHECKING:` blocks to real module-level imports where it doesn't cause circular imports, and drop the accompanying `from __future__ import annotations` where possible. Forward references to still-needed TYPE_CHECKING-only imports are preserved, and the few files that still require them (e.g. for self-referential annotations or known circular dependencies between backend, config, manager, etc.) keep both the guard and the future import.

We can remove all of these __future__ and imports and TYPE_CHECKING guards once our lowest supported python version is 3.14, which imports PEP 649, deferred evaluation of annotations: https://peps.python.org/pep-0649/

This gets rid of ~300 lines of boilerplate that I often find inscrutable: "when should I put this in TYPE_CHECKING? when should I not? why the difference?".

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,6 +28,7 @@ MOCK_MODULES = [
     "libqtile._ffi_pango",
     "libqtile.backend.wayland._ffi",
     "libqtile.backend.x11._ffi_xcursors",
+    "cairocffi",
     "cairocffi.ffi",
     "cairocffi.xcb",
     "cairocffi.pixbuf",
--- a/libqtile/backend/__init__.py
+++ b/libqtile/backend/__init__.py
@@ -2,13 +2,11 @@ from __future__ import annotations
 
 import importlib
 import importlib.util
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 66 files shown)


elParaguayo:
Thanks for this. I've always struggled with when things need to go in the type checking block and when they don't so this seems good to me.

The docs failure is a weird one, I'm not sure why RTD is now tripping over `cairocffi.xcb.XCBSurface` but I wonder if changing that to `cairocffi.XCBSurface` would work instead (even though they're exactly the same) or is this because `cairocffi` is now being built without `xcb` support for some reason?

tych0:
My guess is that the docs don't set TYPE_CHECKING, so we need to add `cairocffi` itself to the list of mock_modules to get them to build? I'll look at it tonight.

elParaguayo:
We use the x11 backend to render the widget screenshots so it should be able to find the module.

elParaguayo:
Should we merge this?

tych0:
Yeah, I think so. And probably open an issue for when we get to 3.14.

PR #5931, opened 2026-04-21
wayland: Update active keyboard when destroyed

### LINKED ISSUE #5927: [wayland] After closing a virtual keyboard, keyboard & window focus becomes disjointed.
### Issue description

1. Run two windows that can capture keyboard input.  
2. Open wvkbd (or I'm assuming any virtual keyboard, but it's the only game in wayland town) 
3. Type in the window on the left with `wvkbd`
4. Close wvkbd and focus the window on the right via follow mouse focus. (not sure why my minimal config without follow focus has follow focus working, but I digress)
5. Start typing with a physical keyboard. 

I can type into the left window, while the right one has window focus.

https://github.com/user-attachments/assets/0cffee10-41c6-40d7-95f5-929e0ad32b32

I've seen similar behavior before, not even sure for how long, but have not attributed it to the virtual keyboard until now. Unfortunately due to a recurring spine related problem I'm partially paralyzed right now (probably I'll be fine again in 2 weeks) & as such I used the virtual keyboard A LOT over the past couple of days. 

### Version

latest git

### Backend

Wayland (experimental)

### Config

```python
from libqtile import qtile

wl_xcursor_theme = "Adwaita"
wl_xcursor_size = 48

wl_input_rules = {
    "*": InputConfig(
        left_handed=False,
        pointer_accel=False,
        accel_profile="flat",
    ),

    "type:keyboard": InputConfig(
        kb_layout="hu",				
    ),
}

mod = "mod4"


keys = [
]


for i in groups:

layouts = [
    layout.MonadTall(),
    layout.Max(),
]

screens = [
    Screen(
        top=bar.Bar(
            [
                widget.GroupBox(),
                widget.Prompt(),
                widget.WindowName(),
            ],
            24,
        ),
    ),
]
#follow_mouse_focus = True
floats_kept_above = True
import os
@hook.subscribe.startup_once
def autostart():
    qtile.spawn(os.path.expanduser("~/.config/qtile/autostart.sh")) # minimal config or not, I literally cannot use my computer without wlr-randr setting the displays.

```

### Logs

```text
-
```

### Required

richcarni:
Fixes #5927

### DIFF
--- a/libqtile/backend/wayland/qw/keyboard.c
+++ b/libqtile/backend/wayland/qw/keyboard.c
@@ -11,10 +11,28 @@ static void qw_keyboard_handle_destroy(struct wl_listener *listener, void *data)
     UNUSED(data);
 
     struct qw_keyboard *keyboard = wl_container_of(listener, keyboard, destroy);
+
     wl_list_remove(&keyboard->modifiers.link);
     wl_list_remove(&keyboard->key.link);
     wl_list_remove(&keyboard->destroy.link);
     wl_list_remove(&keyboard->link);
+
+    // There appears to be a delay between when a keyboard is destroyed and when
+    // wlroots updates the seat's active keyboard. It is possible to miss a focus
+    // event during this delay (issue #5927)
+    // We can work around this by updating the active keyboard here
+    if (keyboard->wlr_keyboard == wlr_seat_get_keyboard(keyboard->server->seat)) {
+        struct qw_keyboard *new_keyboard = NULL;
+        wl_list_for_each(new_keyboard, &keyboard->server->keyboards, link) {
+            wlr_seat_set_keyboard(keyboard->server->seat, new_keyboard->wlr_keyboard);
+            break;
+        }
+        if (new_keyboard == NULL) {
+            wlr_log(WLR_DEBUG, "No keyboards remaining, seat has no active keyboard");
+            wlr_seat_set_keyboard(keyboard->server->seat, NULL);
+        }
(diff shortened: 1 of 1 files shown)


Sinaaaa:
This seems to work really well, what a find!

libqtile/backend/wayland/qw/keyboard.c
```diff
@@ -11,10 +11,21 @@ static void qw_keyboard_handle_destroy(struct wl_listener *listener, void *data)
...
     wl_list_remove(&keyboard->link);
+
+    // If this is the active keyboard, set a new active keyboard (if available)
+    if (keyboard->wlr_keyboard == wlr_seat_get_keyboard(keyboard->server->seat)) {
+        struct qw_keyboard *new_keyboard = NULL;
+        wl_list_for_each(new_keyboard, &keyboard->server->keyboards, link) {
+            wlr_seat_set_keyboard(keyboard->server->seat, new_keyboard->wlr_keyboard);
+            break;
+        }
+    }
```
elParaguayo, line 27: Do you want some log messages here if no other keyboard is found?

richcarni: Sure. And have also explicitly cleared the active keyboard in this case

And updated the comment to better explain the issue

Have been poring over the wlroots source code to figure out what is going on. From testing I can see that it does update the active keyboard, but there seems to be a tiny delay in which our window focus is being missed

PR #5935, opened 2026-05-02
wayland: dispatch pointer enter/leave/motion to bar widgets

akaWolf:
Like x11, bar widgets need mouse_enter / mouse_leave to drive things like tooltips. On x11 the EnterNotify / LeaveNotify / MotionNotify path already feeds Bar.process_pointer_* on each bar window. On wayland nothing was firing those, so hover callbacks (e.g. qtile-extras' TooltipMixin) never ran.

The qw_view returned by qw_server_view_at is already enough — when it points to an Internal view, fire pointer_internal_event_cb after the existing focus update. Track the previous wid in qw_cursor so each transition produces one ENTER / LEAVE / MOTION event. The python side just forwards into win.process_pointer_*, which is the same dispatch x11 uses.

No extra hit-testing on the hot path: the scene graph lookup was already running for ordinary pointer focus.

### DIFF
--- a/libqtile/backend/wayland/cffi/build.py
+++ b/libqtile/backend/wayland/cffi/build.py
@@ -143,6 +143,7 @@ extern "Python" void manage_view_cb(struct qw_view *view, void *userdata);
 extern "Python" void unmanage_view_cb(struct qw_view *view, void *userdata);
 extern "Python" void cursor_motion_cb(void *userdata);
 extern "Python" int cursor_button_cb(int button, uint32_t mask, bool pressed, int x, int y, void *userdata);
+extern "Python" void pointer_internal_event_cb(int wid, int sx, int sy, int event_type, void *userdata);
 extern "Python" void on_screen_change_cb(void *userdata);
 extern "Python" void on_screen_reserve_space_cb(struct qw_output *output, void *userdata);
 extern "Python" void on_input_device_added_cb(void *userdata);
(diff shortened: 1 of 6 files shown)


richcarni:
Thanks for this! I didn't realise this functionality was missing

It's probably fine as is - my only concern is performance

I think we should implement this via callback(s) from wayc (the c wayland backend). Probably from `update_pointer_focus()` in cursor.c. Do you feel like this is something you would be interested working on (I can provide some assistance)? Alternatively, I could probably add the callbacks and you would be able to use them in this PR

tych0:
I think we could also use a test.

akaWolf:
> Thanks for this! I didn't realise this functionality was missing
> 
> It's probably fine as is - my only concern is performance
> 
> I think we should implement this via callback(s) from wayc (the c wayland backend). Probably from `update_pointer_focus()` in cursor.c. Do you feel like this is something you would be interested working on (I can provide some assistance)? Alternatively, I could probably add the callbacks and you would be able to use them in this PR

The old commit was reverted and a new one was added to this branch, as you said that it was desirable to preserve the history of decision-making.

richcarni:
Nice!

Just needs a test :)

akaWolf:
@richcarni the test was added.

richcarni:
Cool, thanks for contributing!

Sinaaaa:
Thank you so much for adding this, amazing!

PR #5936, opened 2026-05-04
wayland: dispatch pointer enter/leave/motion to bar widgets

akaWolf:
Like x11, bar widgets need mouse_enter / mouse_leave to drive things like tooltips. On x11 the EnterNotify / LeaveNotify / MotionNotify path already feeds Bar.process_pointer_* on each bar window. On wayland nothing was firing those, so hover callbacks (e.g. qtile-extras' TooltipMixin) never ran.

The qw_view returned by qw_server_view_at is already enough — when it points to an Internal view, fire pointer_internal_event_cb after the existing focus update. Track the previous wid in qw_cursor so each transition produces one ENTER / LEAVE / MOTION event. The python side just forwards into win.process_pointer_*, which is the same dispatch x11 uses.

No extra hit-testing on the hot path: the scene graph lookup was already running for ordinary pointer focus.

See also https://github.com/qtile/qtile/pull/5935#issuecomment-4365886111


richcarni:
Instead of creating a new PR, would you be able to add this to #5935 so we have a single history of revisions/conversation

(We can always clean up or squash commits before merging)

PR #5937, opened 2026-05-05
Added the functionality to set opacity to windows

Arnob90:
There was a property called opacity that did nothing. This PR attempts to fix that. It has not been tested extensively, only on alacritty. But it should work.


libqtile/backend/wayland/qw/view.c
```diff
@@ -437,3 +438,26 @@ struct qw_output *qw_view_get_primary_output(struct qw_view *view) {
 void qw_view_grab_click(struct qw_view *view) { view->grabbed_click = true; }
 
 void qw_view_ungrab_click(struct qw_view *view) { view->grabbed_click = false; }
+void qw_set_node_opacity(struct wlr_scene_node *node, float opacity) {
```
Gurjaka: can't this be static?

Arnob90: Good catch. Let me try that.

Arnob90: Done, I made it static

Arnob90:
Huh, do I have to implement opacity for x11 too? But that should be handled by picom, right? Can I mark a test as wayland only?

elParaguayo:
Put the test in this fine: https://github.com/qtile/qtile/blob/master/test/backend/wayland/test_window.py

Arnob90:
Fixed it. Thanks.
Sorry for being late. Had some irl stuff.

Gurjaka:
Commits could really use squashing :smile:

elParaguayo:
> Commits could really use squashing 😄

We can do that when we merge if @Arnob90 isn't comfortable doing it themselves. Most important part is that it works!

Arnob90:
Here's a screenshot for convenience. The  browser has opacity 0.8 whereas the terminal has opacity 0.5.
image

elParaguayo:
@richcarni any reason not to merge this?

richcarni:
> @richcarni any reason not to merge this?

I haven't played with it, did have a quick look over the code, LGTM!

elParaguayo:
Thanks. I've tested and it worked nicely for me.

PR #5939, opened 2026-05-06
docs: add a note about xdg-portal setup

tych0:
...and include some symptoms of what might break if this isn't set up correctly.

---
I wonder: should we put these files as examples in /resources for packagers?

### DIFF
--- a/docs/manual/install/index.rst
+++ b/docs/manual/install/index.rst
@@ -149,59 +149,39 @@ be officially supported when using ``uv tool``. However, the following code shou
 Starting Qtile
 ==============
 
-There are several ways to start Qtile. The most common way is via an entry in
-your X session manager's menu. The default Qtile behavior can be invoked by
-creating a `qtile.desktop
-<https://github.com/qtile/qtile/blob/master/resources/qtile.desktop>`_ file in
-``/usr/share/xsessions``.
+The recommended way to start Qtile is as a systemd user service launched from
+your display manager (SDDM, GDM, LightDM, greetd, ...). Starting Qtile this way
+activates ``graphical-session.target``, so services such as xdg-desktop-portal
+(screen sharing, file pickers, ...) work correctly, and it restarts Qtile if it
+crashes. The same setup works whether you run Qtile on X11 or as a Wayland
+compositor: Qtile detects the backend from the session it is launched in, so a
+single session file serves both.
 
-A second way to start Qtile is a custom X session. This way allows you to
-invoke Qtile with custom arguments, and also allows you to do any setup you
-want (e.g. special keyboard bindings like mapping caps lock to control, setting
-your desktop background, etc.) before Qtile starts. If you're using an X
-session manager, you still may need to create a ``custom.desktop`` file similar
-to the ``qtile.desktop`` file above, but with ``Exec=/etc/X11/xsession``. Then,
-create your own ``~/.xsession``. There are several examples of user defined
-``xsession`` s in the `qtile-examples
-<https://github.com/qtile/qtile-examples>`_ repository.
+Qtile ships the required files in `resources/
+<https://github.com/qtile/qtile/tree/master/resources>`_:
 
-If there is no display manager such as SDDM, LightDM or other and there is need
-to start Qtile directly from ``~/.xinitrc`` do that by adding 
-``exec qtile start`` at the end.
+* `qtile.desktop
+  <https://github.com/qtile/qtile/blob/master/resources/qtile.desktop>`_ is the
+  session entry your display manager lists. Install it to
+  ``/usr/share/xsessions`` to offer an X11 session and/or
+  ``/usr/share/wayland-sessions`` to offer a Wayland session.
+* `qtile.service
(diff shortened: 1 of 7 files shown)


docs/manual/install/index.rst
```diff
@@ -178,6 +178,48 @@ be to start through a loop to save running applications:
...
+
+Then, in your ``~/.xsession`` (or equivalent startup script), set
+``XDG_CURRENT_DESKTOP`` so the desktop portal picks the right backend,
+propagate the environment to the user manager so child services inherit it,
+and start Qtile via ``systemctl`` instead of execing it directly:
+
+.. code-block:: bash
+
+    export XDG_CURRENT_DESKTOP=qtile
+    systemctl --user import-environment XDG_CURRENT_DESKTOP PATH
```
dannycolin: Can we set it in qtile.service or qtile-session.target. If it's possible and doesn't break any good pratice rule, I think that'd be a better solution especially when running Qtile under Wayland.

dannycolin: Arch wiki has some interesting recommendations https://wiki.archlinux.org/title/Systemd/User#Environment_variables

ervinpopescu: I've always set this in my autostart script, even on Wayland, thanks for the pointer!

tych0: Oh yeah, I actually set a bunch of stuff including GTK theme. I'll stick them all in there as examples.

jwijenbergh:
This seems not complete, https://gist.github.com/jwijenbergh/48da1a8f4c4a56d122407c4d009bc81f#screen-sharing--desktop-portals

see `~/.config/xdg-desktop-portal/qtile-portals.conf`

tych0:
> see ~/.config/xdg-desktop-portal/qtile-portals.conf

Isn't this covered by:

> BindsTo=graphical-session.target

?

In particular, things like xdg-desktop-portal-gtk and friends are started for me.

tych0:
(But it looks like you have some other useful env vars and such there. I'll read those along with the systemd link, and incorporate some of my own findings after running this setup. I'll ping everyone for a re-review when I think I have it all integrated.)

docs/manual/install/index.rst
richcarni: Should this be `systemctl --user import-environment` here, to import all environment variables?
Otherwise, it seems like we're losing a lot of them (e.g. XDG_SESSION_TYPE, which many apps use to determine whether to use wayland) (I'm using a display manager)

tych0: I think probably yes, just so whatever people set in their xsession / wayland equivalent works out of the box.

tych0:
Ok, I think this is finally ready for a re-reivew after arguing with claude for a long time :). The commit message has the details, but I ended up stripping out a bunch of stuff and switching this to the recommended approach.

This also depends on https://github.com/qtile/qtile/pull/5988

tych0:
Ok, the flake failure is "expected":

> install: cannot stat 'resources/qtile-wayland.desktop': No such file or directory

I'm not sure how to fix it at this point, I guess we need some nixpkgs fixes; I don't think we need to block on that.

flake.nix
```diff
@@ -104,7 +104,13 @@
       packages = forAllSystems (pkgs: {
         default = self.packages.${pkgs.stdenv.hostPlatform.system}.qtile;
 
-        qtile = import ./nix/qtile.nix { inherit pkgs self; };
+        qtile = (import ./nix/qtile.nix { inherit pkgs self; }).overrideAttrs (_: {
```
Gurjaka: Could you move `postinstall` part in `nix/qtile.nix` file? Inside l39 bracket

tych0: Yep, thanks. Hopefully it looks ok now?

nix/qtile.nix
```diff
@@ -41,6 +41,11 @@ let
       postPatch = "";
 
       patches = [ ];
+
+      postInstall = ''
+        install resources/qtile.desktop -Dt $out/share/xsessions
+        install resources/qtile.desktop -Dt $out/share/wayland-sessions
```
Gurjaka, line 47: We need to also install service files, and replace hard-coded paths with correct nix store locations:
```nix
postInstall = ''
  install resources/qtile.desktop -Dt $out/share/xsessions
  install resources/qtile.desktop -Dt $out/share/wayland-sessions
  install -Dm644 resources/qtile.service \
    $out/share/systemd/user/qtile.service
  install -Dm644 resources/qtile-session.target \
    $out/share/systemd/user/qtile-session.target

  substituteInPlace $out/share/systemd/user/qtile.service \
    --replace-fail '%h/.local/bin/qtile' '${placeholder "out"}/bin/qtile' \
    --replace-fail '/usr/bin/systemctl' '${pkgs.systemd}/bin/systemctl'
'';
```

tych0: Those should still be the same, IIUC.

Gurjaka: > Those should still be the same, IIUC.

The desktop installation yeah, but current `postInstall` doesn't handle service instalation

tych0: Oh, yep, makes sense thanks.

PR #5940, opened 2026-05-07
Remove redundant `group.layout_all()` calls

elParaguayo:
In a couple of places there are calls to `group.layout_all` before `group.focus`. As `group.focus` calls `layout_all` at the end of the function, we can remove the unnecessary calls.


tych0:
Nice, thank you!

PR #5941, opened 2026-05-11
DF Widget: Floating Point Values and Used Space

samunemeth:
I noticed that the DF widget used integer division when converting from bytes to kilo-, mega- or gigabytes. This way, when trying to display the free space in GiB to two decimals with `{f:.2f}` in the format parameter of the widget, the digits after the dot would always be zero.

I was also missing a used space option, something that I would want to add to my widget. I checked using `df -h /` and the value is the same as what is calculated here.

*I recognize that this change affects existing configs,* as it might produce long unwanted decimals if `:.0f`is not added after the parameter in the format string.

I would like to hear your input on this change!

### DIFF
--- a/libqtile/widget/df.py
+++ b/libqtile/widget/df.py
@@ -18,7 +18,7 @@ class DF(base.InLoopPollText):
         (
             "format",
             "{p} ({uf:.0f}{m}|{r:.0f}%)",
-            "String format (p: partition, s: size, "
+            "String format (p: partition, s: size, u: used space, "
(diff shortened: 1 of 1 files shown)


tych0:
Thanks, change seems reasonable to me. I would structure the commits slightly differently: 1. commit changing from integer division to float division and changing the default format to `.0f` to compensate and 2. adding the new parameter's code, definition, and docs.

If you can squash it like that, I'm happy to merge.

samunemeth:
Hi, thanks I agree the commits were a bit messy, should be good now.

PR #5943, opened 2026-05-13
docs/screenshots: use \x1b/\x07 instead of invalid \e/\007 escapes

tych0:
\e is not a valid Python escape sequence and triggers a SyntaxWarning on 3.12+ (will become a SyntaxError eventually). Also switch to hex from octal, since I can't read octal :). This lets us get rid of the noqa.


elParaguayo:
Is this file even used in our docs?

tych0:
oh, i don't know actually. let me see if we can just delete it...

tych0:
Nice, docs build seems to have worked. Good catch!

PR #5944, opened 2026-05-13
widgets: move to zoneinfo for timezone manipulation

### LINKED ISSUE #5798: Migrate from pytz to datetime
Upstream python had introduced zoneinfo/datetime to the stdlib in python 3.9 and now pytz is deprecated and can cause issues with some later versions of python.

It looks like the oldest version of python supported in qtile is 3.12 so migrating to datetime should make things a little more straightforward and drop an unmaintained dep.

tych0:
Now that we can use upstream zoneinfo to do lookup-by-string for timezone, switch to that so there's no longer an optional out of tree dependency required.

This also greatly simplifies the tests: we don't need any fake bar infrastructure, monkey patching of all the libraries, etc. We can just create the widget, set the timezone, and call poll() accordingly to see what time it renders.

Fixes #5798

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 Qtile x.xx.x, released xxxx-xx-xx:
     * features
+      - The Clock widget now uses the stdlib zoneinfo module for string
+        timezone lookup. pytz is no longer a (optional) dependency.
     * bugfixes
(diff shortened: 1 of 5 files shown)


elParaguayo (approved):
Nice. I had to do a lot of monkeypatching on this one so nice to get it all tidied up. Thank you!

PR #5945, opened 2026-05-13
CHANGELOG: add entries for changes since 0.35.0

tych0:
I used AI to do this, trying to ignore all internal development stuff like our switch to github actions, etc.

I'm happy to hear "this is not useful", but then someone will have to do it manually, and usually nobody does :)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,8 +2,34 @@ Qtile x.xx.x, released xxxx-xx-xx:
     * features
       - The Clock widget now uses the stdlib zoneinfo module for string
         timezone lookup. pytz is no longer a (optional) dependency.
+      - prompt widget: add options for block, invisible, and bar cursor
+      - df widget: add used space formatting parameter
+      - wayland: add ability to set opacity on windows
+      - wayland: add grab click support
+      - Add client_focus_by_click hook
+      - Allow WindowName to have calculated width
     * bugfixes
       - Fix timers on BackgrounPoll-based widgets not clearing on config reload
+      - prompt widget: fix sanitization issue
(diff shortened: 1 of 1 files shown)


CHANGELOG
```diff
@@ -1,7 +1,34 @@
...
+      - df widget: add used space formatting parameter
+      - wayland: add ability to set opacity on windows
+      - wayland: add grab click support
+      - Add client_focus_by_click hook
+      - Allow WindowName to have calculated width
     * bugfixes
       - Fix timers on BackgrounPoll-based widgets not clearing on config reload
+      - prompt widget: fix sanitization issue
+      - df widget: use floating point division for better precision
+      - widget: make ThermalSensor threaded again
```
elParaguayo:
```suggestion
      - Thermal sensor widget: make threaded again
```

CHANGELOG
```diff
@@ -1,7 +1,34 @@
...
+      - Remove redundant group.layout_all() calls
```
elParaguayo: I'd delete this one (about group layout). People are unlikely to notice anything here.

We only noticed it because it was breaking experimental animations on wayland.

elParaguayo (commented):
I think this is a great idea. People always forget to update the changelog. 

I actually have a CI test on qtile-extras that fails if the changelog is not updated. However, I usually just end up ignoring it!

CHANGELOG
tych0: I dropped it, but I disagree that people are unlikely to notice anything. I think these redraws sometimes cause flickering, so getting rid of them is very useful.

PR #5946, opened 2026-05-13
scripts: drop dead qtb script

tych0:
qtb has been broken for years. It does `from libqtile import command` and uses `command.Client(...)` / `command.CommandError`, but libqtile/command/__init__.py is empty — neither symbol exists there. The argparse setup also passes `required=False` to positional add_argument calls, which raises TypeError.


elParaguayo (approved):
Never even knew about this!

tych0:
I hadn't thought about it either, but claude found it, and indeed it does crash in both ways claude suggests!

PR #5949, opened 2026-05-14
[wayland] border colour fixes

### LINKED ISSUE #5942: [WL] Full transparency `border_normal="#00000000"` is now just a black border.
### Issue description

1. In a layout have `border_normal="#00000000".`

You'll get an ugly black border instead of full transparency like before.
I have not updated qtile for the past couple of weeks, so I'm not sure yet which commit is responsible. 

(I'm of the opinion that this method to set transparent borders is rather important, no bias I'm somewhat sure I won't actually cry if I have to figure out how to hide borders while moving a window some other way.)

### Version

latest git

### Backend

Wayland (experimental)

### Config

```python
from libqtile import qtile

wl_xcursor_theme = "Adwaita"
wl_xcursor_size = 48

wl_input_rules = {
    "*": InputConfig(
        left_handed=False,
        pointer_accel=False,
        accel_profile="flat",
    ),

    "type:keyboard": InputConfig(
        kb_layout="hu",				
    ),
}

mod = "mod4"


keys = [
]


for i in groups:

layouts = [
    layout.MonadTall(
    border_normal="#00000000",
    ),
    layout.Max(),
]

screens = [
    Screen(
        top=bar.Bar(
            [
                widget.GroupBox(),
                widget.Prompt(),
                widget.WindowName(),
            ],
            24,
        ),
    ),
]
#follow_mouse_focus = True
floats_kept_above = True
import os
@hook.subscribe.startup_once
def autostart():
    qtile.spawn(os.path.expanduser("~/.config/qtile/autostart.sh")) # minimal config or not, I literally cannot use my computer without wlr-randr setting the displays.

```

### Logs

```text
-
```

### Required

elParaguayo:
Two issues being solved here:
1) scene rects require pre-multiplied RGBA values
2) changing the view opacity set the border opacity to the view opacity,
   overwriting any alpha value in the border.

Fixes #5942

### DIFF
--- a/libqtile/backend/wayland/qw/util.c
+++ b/libqtile/backend/wayland/qw/util.c
@@ -174,3 +174,12 @@ struct qw_view *qw_view_from_wlr_surface(struct wlr_surface *surface, bool *is_l
 
     return NULL;
 }
+
+void qw_util_premultiply_rgba(const float rgba[4], float opacity, float out[4]) {
+    float alpha = rgba[3] * opacity;
+
+    out[0] = rgba[0] * alpha;
+    out[1] = rgba[1] * alpha;
+    out[2] = rgba[2] * alpha;
+    out[3] = alpha;
+}
--- a/libqtile/backend/wayland/qw/util.h
+++ b/libqtile/backend/wayland/qw/util.h
@@ -38,4 +38,7 @@ bool qw_surfaces_on_same_output(struct wlr_surface *surface_a, struct wlr_surfac
 struct qw_view *qw_view_from_wlr_surface(struct wlr_surface *surface, bool *is_layer_surface,
                                          bool *is_session_lock_surface);
 
+// wlr scene rects expect pre-multiplied RGBA values
+void qw_util_premultiply_rgba(const float rgba[4], float opacity, float out[4]);
+
 #endif /* UTIL_H */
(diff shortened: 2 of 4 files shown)


elParaguayo:
@richcarni would welcome your thoughts on this one.

Background is that the opacity PR overwrites the border rect's alpha value so we need to store the desired colour somewhere and apply alpha on top of that. In addition, scene rect's expect a pre-multiplied RGBA value.

I've dealt with this by storing the RGBA value in the scene rect node's data field. As we currently expect node->data to be empty or have a pointer to the view, I've added some additional checks to make sure we don't try to create a view from the RGBA data.

richcarni:
I think I'd avoid storing different types in node->data if possible

Sounds like the issue from the opacity PR is that opacity should only be applied to the window, not the borders?

I've been thinking for a while that we could structure the view node more hierarchically. For example (and this ties in with children/transients but we can ignore that for now), something like in #5815:
view-node

Maybe the first level could be borders and views, so that opacity could be applied to the borders and views independently

Could something like that work, or am I misunderstanding the problem?

elParaguayo:
I think your misunderstanding the problem but I've also thought of a different solution that avoids saving the colour in the rect node's data field.

Let me give it a try.

elParaguayo:
OK. I've updated with a much much neater approach!

richcarni:
I think I follow now - need to store the non-alpha colour somewhere to use when changing opacity requires re-computing premultiplied RGBA

Anyway, looks good!

elParaguayo:
Almost, we need to store the full RGBA value (ie including alpha) and pre multiply it. Eg if a user sets the border to be be 00000000 the alpha value should always be 0, regardless of the window's own opacity.

PR #5953, opened 2026-05-16
Make scratchpad geometry changes persist across hide/show cycles

### LINKED ISSUE #5951: Scratchpad dropdown does not remember geometry across hide/show cycles (with fix)
## Problem

When a scratchpad dropdown is resized or moved by the user, hiding and re-showing it resets the window back to its originally configured position and size.

## Root cause

`DropDownToggler.show()` in `libqtile/scratchpad.py` recalculates the window geometry from the configured fractional values (`self.x`, `self.y`, `self.width`, `self.height`) on every call and force-places the window at that position:

```python
def show(self):
    if (not self.visible) or (not self.shown):
        win = self.window
        screen = win.qtile.current_screen
        # calculate windows floating position and width/height
        # these may differ for screens, and thus always recalculated.
        x = int(screen.dx + self.x * screen.dwidth)
        y = int(screen.dy + self.y * screen.dheight)
        win.float_x = x
        win.float_y = y
        width = int(screen.dwidth * self.width) - 2 * self.border_width
        height = int(screen.dheight * self.height) - 2 * self.border_width
        win.place(x, y, width, height, self.border_width, win.bordercolor, respect_hints=True)
        WindowVisibilityToggler.show(self)
```

## Fix

The changes are confined to `libqtile/scratchpad.py`, inside `DropDownToggler`.

In the `__init__` method, add a geometry attribute `self._geometry` alongside the other instance variables, before the `WindowVisibilityToggler.__init__` call (which triggers the first `show()`):

```python
self._geometry: tuple[float, float, float, float] | None = None
```

This keeps configured values (`self.x`, `self.y`, `self.width`, `self.height`) intact and separate from runtime state.

### Add a `hide()` override

Add this method immediately before the existing `show()`:

```python
def hide(self):
    if self.visible or self.shown:
        win = self.window
        screen = win.group.screen if win.group else None
        if screen is not None:
            self._geometry = (
                (win.x - screen.dx) / screen.dwidth,
                (win.y - screen.dy) / screen.dheight,
                (win.width + 2 * self.border_width) / screen.dwidth,
                (win.height + 2 * self.border_width) / screen.dheight,
            )
    WindowVisibilityToggler.hide(self)
```

### Version

v0.35.0

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

medwatt:
Pull request for issue discussed [here](https://github.com/qtile/qtile/issues/5951).

### DIFF
--- a/libqtile/scratchpad.py
+++ b/libqtile/scratchpad.py
@@ -163,6 +163,7 @@ class DropDownToggler(WindowVisibilityToggler):
         self.y = ddconfig.y
         self.width = ddconfig.width
         self.height = ddconfig.height
+        self.geometry = (self.x, self.y, self.width, self.height)
         self.border_width = window.qtile.config.floating_layout.border_width
 
         # add "SKIP_TASKBAR" to _NET_WM_STATE atom (for X11)
@@ -192,23 +193,36 @@ class DropDownToggler(WindowVisibilityToggler):
         )
         return info
 
+    def hide(self):
+        if self.visible or self.shown:
+            win = self.window
+            screen = win.group.screen if win.group else None
+            if screen is not None:
+                self.geometry = (
+                    (win.x - screen.dx) / screen.dwidth,
+                    (win.y - screen.dy) / screen.dheight,
+                    (win.width + 2 * self.border_width) / screen.dwidth,
+                    (win.height + 2 * self.border_width) / screen.dheight,
+                )
+        WindowVisibilityToggler.hide(self)
+
     def show(self):
         """
         Like WindowVisibilityToggler.show, but before showing the window,
         its floating x, y, width and height is set.
         """
         if (not self.visible) or (not self.shown):
-            # SET GEOMETRY
             win = self.window
             screen = win.qtile.current_screen
-            # calculate windows floating position and width/height
+            x_frac, y_frac, w_frac, h_frac = self.geometry
+            # Calculate windows floating position and width/height
             # these may differ for screens, and thus always recalculated.
-            x = int(screen.dx + self.x * screen.dwidth)
-            y = int(screen.dy + self.y * screen.dheight)
+            x = int(screen.dx + x_frac * screen.dwidth)
+            y = int(screen.dy + y_frac * screen.dheight)
             win.float_x = x
             win.float_y = y
-            width = int(screen.dwidth * self.width) - 2 * self.border_width
-            height = int(screen.dheight * self.height) - 2 * self.border_width
+            width = int(screen.dwidth * w_frac) - 2 * self.border_width
+            height = int(screen.dheight * h_frac) - 2 * self.border_width
             win.place(x, y, width, height, self.border_width, win.bordercolor, respect_hints=True)
             # Toggle the dropdown
             WindowVisibilityToggler.show(self)
--- a/test/test_scratchpad.py
+++ b/test/test_scratchpad.py
@@ -44,6 +44,17 @@ class ScratchPadBaseConfig(Config):
             ],
             single=True,
         ),
+        libqtile.config.ScratchPad(
+            "GEOMETRY_SCRATCHPAD",
+            dropdowns=[
+                libqtile.config.DropDown(
+                    "dd-a", spawn_cmd("dd-a"), x=0.2, y=0.2, width=0.6, height=0.6
+                ),
+                libqtile.config.DropDown(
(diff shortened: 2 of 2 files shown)


elParaguayo:
A test here would be good: toggle scratchpad, resize it, hide it, toggle it and verify size is correct.

Let me know if you need help with this test or the one in your other PR.

medwatt:
Thanks. I'm not that comfortable with testing. Any help in this regard would be nice.

elParaguayo:
I can help with the test but I suspect some users may not want this behaviour so we could add an option like `reset_size_on_toggle` and set this as `True` as default (so current users see no difference).

medwatt:
Thanks for helping with the tests. Yes, I think it makes a lot of sense to have an opt-in option for such a change in behavior as users may have already gotten accustomed to it. However, of the 4 tiling window managers I've tried, qtile is the only one that does not restore changes made in the geometry in a hide/show cycle.

elParaguayo:
Ok - can you add the option. We can debate the default value separately!

medwatt:
@elParaguayo, done. Added `reset_geometry_on_toggle` with a default of `True` and cleaned up the implementation slighly. Have a look.

elParaguayo:
Thanks will try to look later/tomorrow.

elParaguayo:
This should do the test for you:
```diff
diff --git a/test/test_scratchpad.py b/test/test_scratchpad.py
index 7574af78..fc93465d 100644
--- a/test/test_scratchpad.py
+++ b/test/test_scratchpad.py
@@ -44,6 +44,23 @@ class ScratchPadBaseConfig(Config):
             ],
             single=True,
         ),
+        libqtile.config.ScratchPad(
+            "GEOMETRY_SCRATCHPAD",
+            dropdowns=[
+                libqtile.config.DropDown(
+                    "dd-a", spawn_cmd("dd-f"), x=0.2, y=0.2, width=0.6, height=0.6
+                ),
+                libqtile.config.DropDown(
+                    "dd-b",
+                    spawn_cmd("dd-f"),
+                    x=0.2,
+                    y=0.2,
+                    width=0.6,
+                    height=0.6,
+                    reset_geometry_on_toggle=False,
+                ),
+            ],
+        ),
         libqtile.config.Group("a"),
         libqtile.config.Group("b"),
     ]
@@ -323,3 +340,49 @@ def test_skip_taskbar(manager):
         # check that window's _NET_WM_STATE contains _NET_WM_STATE_SKIP_TASKBAR
         net_wm_state = manager.c.window.eval("self.window.get_net_wm_state()")
         assert "_NET_WM_STATE_SKIP_TASKBAR" in net_wm_state
+
+
+@scratchpad_config
+def test_geometry(manager):
+    def assert_geometry(x, y, w, h):
+        info = manager.c.window.info()
+        assert (x, y, w, h) == (info["x"], info["y"], info["width"], info["height"])
+
+    def move_and_resize():
+        manager.c.window.set_size_floating(200, 200)
+        manager.c.window.set_position_floating(10, 10)
+
+    scratch = manager.c.group["GEOMETRY_SCRATCHPAD"]
+
+    for dd in ["dd-a", "dd-b"]:
+        scratch.dropdown_reconfigure(dd)
+
+    # dd-a should reset size on toggle
+    scratch.dropdown_toggle("dd-a")
+    is_spawned(manager, "dd-a", "GEOMETRY_SCRATCHPAD")
+    assert_geometry(160, 120, 478, 358)
+    move_and_resize()
+    assert_geometry(10, 10, 200, 200)
+
+    # Hide and restore
+    scratch.dropdown_toggle("dd-a")
+    scratch.dropdown_toggle("dd-a")
+
+    # Geometry should be reset
+    assert_geometry(160, 120, 478, 358)
+
+    scratch.dropdown_toggle("dd-a")
+
+    # dd-b should preserve position and size
+    scratch.dropdown_toggle("dd-b")
+    is_spawned(manager, "dd-b", "GEOMETRY_SCRATCHPAD")
+    assert_geometry(160, 120, 478, 358)
+    move_and_resize()
+    assert_geometry(10, 10, 200, 200)
+
+    # Hide and restore
+    scratch.dropdown_toggle("dd-b")
+    scratch.dropdown_toggle("dd-b")
+
+    # Geometry should be preserved
+    assert_geometry(10, 10, 200, 200)
```

tych0:
I added the test with @elParaguayo and dropped the config value since this seems like reasonable behavior to me.

tych0:
Didn't pass, though? :(

PR #5954, opened 2026-05-16
Fixes the issue where background wallpaper flashes briefly when switching between windows

### LINKED ISSUE #5952: Background flashes briefly when switching between windows (with fix)
## Problem

When switching focus between windows in layouts such as Stack or Max, the desktop wallpaper flashes for a single frame before the newly focused window appears. I reported this issue sometime ago [here](https://github.com/qtile/qtile/discussions/5006), but is yet to be resolved.

## Root cause

`Layout.layout()` in `libqtile/layout/base.py` iterates over the window list and calls `configure()` on each window in order:

```python
def layout(self, windows, screen_rect):
    for i in windows:
        self.configure(i, screen_rect)
```

In the Stack layout (and similarly in Max), `configure()` calls `client.hide()` on windows that are not currently focused and `client.unhide()` on the focused one.

When focus switches from window A to window B, the list order determines which `configure()` call happens first. If A appears before B in the list, A is unmapped first. There is then a brief interval, one or more rendered frames, before B is mapped. During that interval nothing covers the screen area, so the desktop background is exposed.

## Fix

The fix is confined to `Layout.layout()` in `libqtile/layout/base.py`.

Instead of iterating windows in arbitrary list order, always configure the currently focused window first. This guarantees that the incoming window is mapped before any outgoing window is unmapped, eliminating the exposed background window entirely.

```python
def layout(self, windows: Sequence[Window], screen_rect: ScreenRect) -> None:
    # Configure the focused window first so it is shown before any other
    # window is hidden, preventing a one-frame background flash on switch.
    current: Window | None = (
        cast(Window, self._group.current_window) if self._group is not None else None
    )

    if current is not None and current in windows:
        self.configure(current, screen_rect)
    for i in windows:
        if i is not current:
            self.configure(i, screen_rect)
```

`cast` must be added to the existing `from typing import ...` line at the top of the file. No other files need to change.

## Scope

Only two layouts override `layout()`:

- `TreeTab` (`libqtile/layout/tree.py`): overrides it but immediately calls `Layout.layout(self, windows, body)`, so it inherits the fix directly.

- `Slice` (`libqtile/layout/slice.py`): delegates to sub-layouts that each call their own `layout()`, so they also inherit the fix.

Every other layout (Column, MonadTall, MonadWide, Bsp, Tile, Spiral, Floating, etc.) inherits `layout()` from the base class unchanged, so all are covered by this single change.

### Version

v0.35.0

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

medwatt:
Pull request for issue discussed [here](https://github.com/qtile/qtile/issues/5952).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,8 @@ Qtile x.xx.x, released xxxx-xx-xx:
         Layer-shell surfaces that commit after their output is destroyed, session
         lock surfaces on a destroyed output, and killed internal views no longer
         dereference freed compositor state.
+      - Fix window configuration order so the current window is configured
+        before other windows, reducing flickering
 
 Qtile 0.36.0, released 2026-05-22:
     * features
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -122,6 +122,10 @@ class _Group(CommandObject):
                 floating = [x for x in self.windows if x.floating and not x.minimized]
                 screen_rect = self.screen.get_rect()
                 if normal:
+                    # Send current window first to prevent flickering
+                    if self.current_window is not None and self.current_window in normal:
+                        normal.remove(self.current_window)
+                        normal.insert(0, self.current_window)
                     try:
                         self.layout.layout(normal, screen_rect)
                     except Exception:


tych0:
Can you edit your issue explanation into the commit message?

libqtile/layout/base.py
```diff
@@ -32,8 +32,15 @@ def __init__(self, **config: Any) -> None:
         self._group: _Group | None = None
 
     def layout(self, windows: Sequence[Window], screen_rect: ScreenRect) -> None:
+        # Configure the focused window first so it is shown before any other
+        # window is hidden, preventing a one-frame background flash on switch.
+        # See issue: https://github.com/qtile/qtile/issues/5952
```
tych0: I don't think we need this comment, if you put your description in the commit message git blame will be enough.

elParaguayo:
Can we do this in `Group` instead? e.g.
```diff
diff --git a/libqtile/group.py b/libqtile/group.py
index 35289e1c..5e0bc9e7 100644
--- a/libqtile/group.py
+++ b/libqtile/group.py
@@ -122,6 +122,10 @@ class _Group(CommandObject):
                 floating = [x for x in self.windows if x.floating and not x.minimized]
                 screen_rect = self.screen.get_rect()
                 if normal:
+                    # Send current window first to prevent flickering
+                    if self.current_window is not None and self.current_window in normal:
+                        normal.remove(self.current_window)
+                        normal.insert(0, self.current_window)
                     try:
                         self.layout.layout(normal, screen_rect)
                     except Exception:
```

medwatt:
@elParaguayo, I just tried your patch and solves the issue. Your fix is cleaner than mine. You can disregard my pull request and use yours instead.

elParaguayo:
I'm happy if you just want to edit your PR to include it.

medwatt:
Thanks. Do you prefer that I push the fixes as an additional commit, or should I amend/squash them into the existing commit and force-push the PR branch?

elParaguayo:
Amending this and a force push is absolutely fine. Thanks.

elParaguayo:
Can you add a note in the changelog bugfixes section. This one''s been around a while so nice to show it's fixed.

medwatt:
@elParaguayo, done. Can you also have a look at the two other PRs ? For example, the one where a floating window becomes tiled after going fullscreen is very annoying especially when you use scratchpads as much as I do.

CHANGELOG
```diff
@@ -109,7 +110,7 @@ Qtile 0.32.0, released 2025-06-09:
     !!! breaking changes !!!
       - Output of `qtile cmd-obj` is now in JSON.
     * features
-        - Add support for `single-pixel-buffer` protocol for wayland 
+        - Add support for `single-pixel-buffer` protocol for wayland
```
tych0: Can you drop this from the diff? but looks good to me otherwise

medwatt:
Yesterday, I was using a program on my computer that automatically spawns new windows. For example, I do a simulation and a plot window appears after the simulation ends. I noticed the background wallpaper flash before the new window appears. It's like the old window is hidden before the new window has been fully rendered. I reverted to the master to check if my PR was causing it, but it was actually worse. So, while my PR does handle some cases, but not all cases.

PR #5955, opened 2026-05-16
Fixes the issue where a floating window loses its float state after exiting fullscreen

### LINKED ISSUE #5950: Floating window loses float state after exiting fullscreen (with fix)
## Problem

When a floating window enters and then exits fullscreen mode, it becomes tiled instead of returning to its previous floating state. This affects ordinary floating windows as well as scratchpad dropdowns. This issue is reported [here](https://github.com/qtile/qtile/issues/2415) and hasn't been fixed since 2021.

**Steps to reproduce:**

1. Open a floating window.
2. Enter fullscreen.
3. Exit fullscreen.
4. The window is now tiled instead of floating.

The same issue occurs with scratchpad dropdowns: toggling fullscreen inside a dropdown causes it to become a regular tiled window when fullscreen is exited, and the scratchpad loses track of the window entirely.

### The design issue

`_float_state` is an enum that tracks how a window is being managed:

```python
class FloatStates(enum.Enum):
    NOT_FLOATING = 1   # tiled by a layout
    FLOATING     = 2   # freely positioned by the user
    MAXIMIZED    = 3   # covers the work area
    FULLSCREEN   = 4   # covers the entire screen
    TOP          = 5   # scratchpad / always-on-top
    MINIMIZED    = 6   # hidden
```

Fullscreen and floating are conceptually orthogonal: a window can be fullscreen *and* tiled, or fullscreen *and* floating. However, qtile models fullscreen as just another value in the same enum, alongside the actual float/tile management states. This means entering fullscreen unconditionally overwrites `_float_state` with `FloatStates.FULLSCREEN`, destroying the information about what the window was doing before.

### The bug in the setter

The `fullscreen` setter in `libqtile/backend/x11/window.py` saves the window geometry before entering fullscreen so it can be restored on exit:

```python
if self._float_state not in (FloatStates.MAXIMIZED, FloatStates.FULLSCREEN):
    self._save_geometry()
```

But it never saves `_float_state` itself. On exit, it calls `self.floating = False` unconditionally:

```python
if self._float_state == FloatStates.FULLSCREEN:
    self._restore_geometry()
    self.floating = False   # always tiles the window
    self.change_layer()
    return
```

This removes the window from the floating layout and inserts it into the tiled layout regardless of whether it was floating before going fullscreen.

### Why scratchpads are also affected

Scratchpad dropdowns set `_float_state = FloatStates.TOP` directly (see `scratchpad.py`), which is a distinct value from `FloatStates.FLOATING`. After exiting fullscreen, the window ends up as `NOT_FLOATING`, so `window.floating` returns `False`. The `ScratchPad` group subscribes to the `float_change` hook and detaches its dropdown whenever its window is no longer floating, so the scratchpad loses track of the window entirely; the next toggle spawns a fresh process.

## Fix

The fix mirrors the existing pattern for geometry: just as `_save_geometry()` captures `x`, `y`, `width`, and `height` before they are overwritten by fullscreen, we capture `_float_state` at the same moment. In the `__init__` method, declare the attribute alongside the other float-state variables:

```python
self._float_state = FloatStates.NOT_FLOATING
self._pre_fullscreen_float_state: FloatStates = FloatStates.NOT_FLOATING
```

### Entering fullscreen: Save the pre-fullscreen float state

```python
if self._float_state not in (FloatStates.MAXIMIZED, FloatStates.FULLSCREEN):
    self._save_geometry()
    self._pre_fullscreen_float_state = self._float_state
```

### Exiting fullscreen; restore the saved state

```python
if self._float_state == FloatStates.FULLSCREEN:
    self._restore_geometry()
    if self._pre_fullscreen_float_state != FloatStates.NOT_FLOATING:
        self._reconfigure_floating(new_float_state=self._pre_fullscreen_float_state)
    else:
        self.floating = False
    self.change_layer()
    return
```

## Wayland

The Wayland backend has the identical bug in `libqtile/backend/wayland/window.py`. The fix is the same in structure, with one additional step: Wayland's `_reconfigure_floating` sets `_float_state` before calling `_update_fullscreen`, which means the fullscreen-exit signal to the compositor would be swallowed (the state already appears to match). For the floating-restore path, `_update_fullscreen(False)` must therefore be called explicitly before `_reconfigure_floating` so the compositor is notified while `_float_state` is still `FULLSCREEN`.

### Version

v0.35.0

### Backend

X11 (default)

### Config

```python

```

### Logs

```text

```

### Required

medwatt:
Pull request for issue discussed [here](https://github.com/qtile/qtile/issues/5950).

### DIFF
--- a/libqtile/backend/base/window.py
+++ b/libqtile/backend/base/window.py
@@ -6,6 +6,7 @@ from typing import Any
 
 from libqtile import config, hook
 from libqtile.backend.base import Drawer
+from libqtile.backend.base.float_states import FloatStates
 from libqtile.command.base import CommandError, CommandObject, ItemT, expose_command
 from libqtile.group import _Group
 from libqtile.log_utils import logger
@@ -17,6 +18,11 @@ if typing.TYPE_CHECKING:
 
 
 class _Window(CommandObject, metaclass=ABCMeta):
+    x: int
+    y: int
+    width: int
+    height: int
+
     def __init__(self):
         self.borderwidth: int = 0
         self.name: str = "<no name>"
@@ -125,24 +131,6 @@ class _Window(CommandObject, metaclass=ABCMeta):
     def _select(self, name, sel):
         return None
 
-    def _save_geometry(self):
-        """Save current window geometry."""
-        self.base_x = self.x
-        self.base_y = self.y
-        self.base_width = self.width
-        self.base_height = self.height
-
-    def _restore_geometry(self):
-        """Restore previously saved window geometry."""
-        if self.base_x is not None:
-            self.x = self.base_x
-        if self.base_y is not None:
-            self.y = self.base_y
-        if self.base_width is not None:
-            self.width = self.base_width
-        if self.base_height is not None:
-            self.height = self.base_height
-
     @abstractmethod
     @expose_command()
     def info(self) -> dict[str, Any]:
@@ -244,6 +232,11 @@ class Window(_Window, metaclass=ABCMeta):
     float_y: int | None
     bordercolor: ColorsType | None
 
+    _float_state: FloatStates
+    # The float state saved by save_float_state(), restored when leaving
+    # fullscreen or maximized
+    saved_float_state: FloatStates = FloatStates.NOT_FLOATING
+
     def __repr__(self):
         return f"{self.__class__.__name__!s}(name={self.name!r}, wid={self.wid:d})"
 
@@ -268,11 +261,28 @@ class Window(_Window, metaclass=ABCMeta):
     @property
     def maximized(self) -> bool:
         """Whether this window is maximized."""
-        return False
+        return self._float_state == FloatStates.MAXIMIZED
 
     @maximized.setter
     def maximized(self, do_maximize: bool) -> None:
-        raise NotImplementedError
+        if do_maximize:
+            screen = (self.group and self.group.screen) or self.qtile.find_closest_screen(
+                self.x, self.y
+            )
+
+            if self._float_state not in (FloatStates.MAXIMIZED, FloatStates.FULLSCREEN):
+                self.save_float_state()
+
+            bw = self.group.floating_layout.max_border_width if self.group else 0
+            self._reconfigure_floating(
+                screen.dx,
+                screen.dy,
+                screen.dwidth - 2 * bw,
+                screen.dheight - 2 * bw,
+                new_float_state=FloatStates.MAXIMIZED,
+            )
+        elif self._float_state == FloatStates.MAXIMIZED:
+            self.restore_float_state()
 
     @property
     def minimized(self) -> bool:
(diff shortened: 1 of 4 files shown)


elParaguayo:
This needs a test i.e. take a tiled window, toggle fullscreen, exit fullscreen and confirm it's tiled. Do same with a floating window.

medwatt:
@elParaguayo, are you asking if I had test that it works or that I should write a test for that ? If it's the latter, is there no test for such a thing already ?

elParaguayo:
I'm asking for a test to be written for this. If there was one already then it would be failing - if it doesn't fail then there's no bug!

However, I'll take a look as I'd really expect there to be a test for this!!

medwatt:
Thanks. Can you also have a look at the other PR. It is more opinionated, but I feel it makes a lot of sense to restore a scratchpad's gemetry to the state it were before it got hidden.

elParaguayo:
Struggling with the test for this one. If I toggle fullscreen, the window says it's floating but the group info still shows it as tiled and not floating.

elParaguayo:
There was a test to show that a floating window that is made fullscreen returns to the correct size/position but no test for a tiled window.

Try this:
```diff
diff --git a/test/test_floating.py b/test/test_floating.py
index 809ea574..1e9b79bb 100644
--- a/test/test_floating.py
+++ b/test/test_floating.py
@@ -14,6 +14,7 @@ class FakeScreenConfig(Config):
             "a",
             layouts=[floating_layout],
         ),
+        libqtile.config.Group("b"),
     ]
     layouts = [
         layout.Tile(),
@@ -88,3 +89,52 @@ def test_fullscreen(manager):
     assert manager.c.window.info()["x"] == 50
     assert manager.c.window.info()["y"] == 20
     assert manager.c.window.info()["group"] == "a"
+
+
+@fakescreen_config
+def test_fullscreen_restore_tiling(manager):
+    """Ensure window is restored to tiling"""
+
+    def window(name):
+        for win in manager.c.windows():
+            if win["name"] == name:
+                return manager.c.window[win["id"]]
+        assert False, f"Cannot find window: {name}"
+
+    def is_floating(name):
+        return window(name).info()["floating"]
+
+    def is_tiled(name):
+        return not is_floating(name)
+
+    def is_fullscreen(name):
+        return window(name).info()["fullscreen"]
+
+    def assert_geometry(name, x, y, w, h):
+        win = window(name)
+        info = win.info()
+        assert (x, y, w, h) == (info["x"], info["y"], info["width"], info["height"])
+
+    manager.c.group["b"].toscreen()
+
+    manager.test_window("one")
+    manager.test_window("two")
+
+    for win in ["one", "two"]:
+        assert is_tiled(win)
+        assert not is_fullscreen(win)
+
+    # Check window one's tiled geometry
+    assert_geometry("one", 1186, 10, 732, 1068)
+
+    # Fullscreen
+    window("one").toggle_fullscreen()
+    assert is_fullscreen("one")
+    assert not is_tiled("one")
+    assert_geometry("one", 0, 0, 1920, 1080)
+
+    # Revert to tiled
+    window("one").toggle_fullscreen()
+    assert not is_fullscreen("one")
+    assert is_tiled("one")
+    assert_geometry("one", 1186, 10, 732, 1068)
```

medwatt:
Thanks. Unfortunately, I am travelling today. I'll have a look as soon as I am back in 2 days.

elParaguayo:
No rush at all.

tych0:
Ok, I went ahead and rewrote this, and a bunch of other things as well, since both of the backends had this bug, and they both shared a lot of code. So excluding tests the diff is net negative, even though it fixes a the bug. I added @elParaguayo again as the author of one test, and then added a couple more.

tych0:
I'll plan to leave this up for review for a bit, and then merge it if nobody complains.

PR #5957, opened 2026-05-17
Fix prompt hiding when window closes

### LINKED ISSUE #5894: prompt widget closes on window close
### Issue description

The prompt widget closes whenever a window closes. This can be rather annoying when one was just typing something in there, for whatever reason some window closes and one has to do it again.

To reproduce:
1. Open a terminal window and execute: `sleep 3 && exit`.
2. Open a prompt via Mod+R, type something.
3. Wait until the terminal window closes.
4. Notice that the prompt closed as well.

### Version

0.35

### Backend

X11 (default)

### Config

```python
https://github.com/3hhh/qubes-qtile
```

### Logs

```text
No relevant logs for this issue.
```

### Required

elParaguayo:
If a window closes when a user is using the Prompt widget, qtile will try to focus a different window. If there is a suitable target then the prompt widget will lose focus and the text entry will end. We need to check whether a widget is using the keyboard and, if so, prevent focus shifting to a new window.

Fixes #5894

### DIFF
--- a/libqtile/core/manager.py
+++ b/libqtile/core/manager.py
@@ -1964,3 +1964,10 @@ class Qtile(CommandObject):
     def stop_repl_server(self) -> None:
         """Stop the REPL server."""
         create_task(repl_server.stop())
+
+    def widget_has_keyboard(self) -> bool:
+        for screen in self.screens:
+            for gap in screen.gaps:
+                if gap.has_keyboard():
+                    return True
+        return False
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -30,6 +30,7 @@ Qtile x.xx.x, released xxxx-xx-xx:
       - Initialise fontconfig before starting qtile
       - nix: add fontconfig to runtime library configuration
       - Only pass button presses to bar when press is inside bar
+      - Stop Prompt widget cancelling text entry when window closes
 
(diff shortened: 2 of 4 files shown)


libqtile/core/manager.py
```diff
@@ -1964,3 +1964,12 @@ def start_repl_server(self, locals_dict: dict[str, Any] = dict()) -> None:
     def stop_repl_server(self) -> None:
         """Stop the REPL server."""
         create_task(repl_server.stop())
+
+    @property
+    def widget_has_keyboard(self) -> bool:
+        for screen in self.screens:
+            for gap in screen.gaps:
+                if isinstance(gap, bar.Bar):
+                    if gap._has_keyboard is not None:
```
tych0: Can we make this a method on Gap?

elParaguayo: Have pushed an update. Let me know if that's what you had in mind.

tych0: Yep, looks good to me. I would say I generally prefer methods to properties, but at least that pushes the abstraction into Gap.

elParaguayo: OK - I'll switch to a method and squash it (I've disabled the automerge)

